PR #714, opened 2026-04-29
feat(collector): add Talos OS support via Kubernetes Node info

ayuskauskas:
## Summary

Adds Talos Linux support to the snapshot agent and collector framework, unblocking AICR deployment on clusters that have no systemd and no host `/etc/os-release` accessible to unprivileged pods. Selection is driven by the existing `os` recipe criterion (new `os: talos` value); no new criteria fields are introduced.

## Motivation / Context

Talos has no systemd D-Bus and exposes no `/etc/os-release` at the host filesystem level, so the snapshot agent's privileged pod fails to start: the `/run/systemd` and `/etc/os-release` `hostPath` mounts in `pkg/k8s/agent/job.go` cannot resolve. This is the documented blocker raised in the issue thread.

This PR is the first phase of Talos enablement, scoped to the agent / collector / validator update path. Future phases (declarative cross-OS namespace and security-context configuration; optional Nodewright tuning package for Talos) are deferred to follow-on PRs.

Fixes: N/A
Related: #565

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] CLI (`cmd/aicr`, `pkg/cli`)
- [ ] API server (`cmd/aicrd`, `pkg/api`, `pkg/server`)
- [x] Recipe engine / data (`pkg/recipe`)
- [ ] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [x] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [ ] Validator (`pkg/validator`)
- [x] Core libraries (`pkg/errors`, `pkg/k8s`)
- [x] Docs/examples (`docs/`, `examples/`)
- [ ] Other: ____________

### Selection via existing OS criterion

`recipe.CriteriaOSTalos = "talos"` added to `pkg/recipe/criteria.go` with the standard enum audit (`api/aicr/v1/server.yaml` × 3, `docs/`, `site/docs/`, `pkg/recipe/doc.go`, `pkg/cli/recipe.go`). The collector factory branches on this value via a new `WithOS` option:

```go
func (f *DefaultFactory) CreateSystemDCollector() Collector {
    if f.OS == OSTalos { return talos.NewServiceCollector() }
[...]

Issue #565, opened 2026-04-14
[Feature]: Talos Linux support


rothgar:
### Prerequisites

- [x] I searched existing issues

### Feature Summary

Add support for scanning and deploying to clusters that run [Talos Linux](https://talos.dev). 

### Problem/Use Case

Talos is a Linux distro that provides a CNCF AI conformant Kubernetes environment (using upstream Kubernetes containers), but trying to scan a cluster with `aicr` quickly exposes issues because Talos has hardened security defaults (e.g. enforces baseline pod security admission) which causes the job not to run and when deploying to a namespace that allows the pod to run it fails because Talos doesn't use systemd
```
Warning  FailedMount  24s (x8 over 88s)  kubelet            MountVolume.SetUp failed for volume "run-systemd" : hostPath type check failed: /run/systemd is not a directory
```

### Proposed Solution

Add configuration and testing for Talos based Kubernetes clusters.

### Success Criteria

- Supports scanning clusters that run on Talos
- Supports deploying bundles
- Supports validating config

### Component

CLI (aicr)

### Priority

Nice to have

### Are you willing to contribute?

Yes, I can open a PR

ayuskauskas:
This would be a great addition to AICR, thank you for bringing it up as an issue. I think this would be best addressed in a few discrete phases.

Issues:
1. Updates to the agent/collector/validator to be able to collect the required evidence in Talos.
2. Be able to set namespace configuration across recipes in a more declarative way. I.e nodewright, gpu/network operator, collector, validator all need to run in special namespaces on Talos clusters. I am thinking this could either just be a documentation file that enumerates for X cluster orchestrator and/or X OS run with the following options. Or a deeper integration where the user can specify the orchestrator/OS as criteria and AICR modifies appropriately. There is already a PR from openShift which will have similar issue are security context so defining a way to have these be specified across deployment geography seems relevant.

From this the following phases I think would make sense:
1. Update agent/collector to be able collector evidence without requiring systemd. We don't currently use any of those in criteria but we are looking for how drift/node validation will work so removing the capability would be less ideal than collecting similar information a different way.
2. Update validator which should be straightforward based on phase 1.
3. Design of the system to let these be set across deployment geographies (issue 2)
4. (Optional) Nodewright package to do tuning/setup in Talos if possible.

xdu31:
[...]

PR #762, opened 2026-05-05
feat(tools): add local Talos cluster + snapshot chainsaw test

ayuskauskas:
## Summary

Adds an opt-in, kind-equivalent harness for spinning up a local Talos
Linux cluster on a developer laptop (`tools/talos-test/`) and a chainsaw
test that exercises the AICR snapshot agent against it
(`tests/chainsaw/snapshot/deploy-agent-talos/`). Also adds `--requests`
and `--limits` flags to `aicr snapshot` so the agent can be right-sized
on resource-constrained dev clusters.

## Motivation / Context

PR #714 added Talos OS support (`pkg/collector/talos/` + the `OS=talos`
pod-shape branch in `pkg/k8s/agent/job.go`), but the only coverage was
unit tests. There was no integration test running the snapshot agent
against an actual Talos node — and no easy way for a developer on macOS
to exercise that path locally without standing up a remote cluster.

The agent container's resource requests/limits were also hardcoded
(`Privileged`: 4Gi req / 8Gi lim memory). That sizing fits production
GPU nodes but blocks scheduling on the talosctl Docker provisioner's
default 2Gi worker, with no way to override at the node level.

Fixes: N/A
Related: #714, #565

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tests/chainsaw`, `tools/`, `Makefile`

### `tools/talos-test/{up.sh,down.sh,README.md}` — cluster lifecycle

- Wraps `talosctl cluster create docker` (newer talosctl moved
  provisioners to subcommands; the old `--provisioner` flag is gone).
- Threads a containerd registry-mirror config patch via
  `--config-patch` so `localhost:5001/aicr:local` resolves transparently
  from inside Talos node containers (the host registry started by
  `make dev-env`).
- `br_netfilter` preflight: the macOS host VM (Docker Desktop's
  LinuxKit or Podman Machine's Fedora CoreOS) ships the module but
  does not auto-load it, so flannel CNI fails until we modprobe it via
  a one-shot privileged sidecar with `/lib/modules` bind-mounted from
  the VM. No-op on Linux hosts.
- `talosctl health --wait-timeout 8m` streams readiness phases
  (etcd, apid, kubelet, control plane, all k8s nodes ready, kube-proxy,
  CoreDNS). On failure the script prints diagnostic commands the
  developer can run to drill into stuck services.
- The Talos in-network IP (`10.5.0.0/24`) is not host-routable on
  macOS, so the script derives the docker-NAT'd apid host port (50000)
  and K8s API host port (6443) from `docker port`, passes them as
  `--endpoints`, and rewrites the kubeconfig server URL to
  `https://127.0.0.1:<host-port>` with `insecure-skip-tls-verify=true`.
- Relabels the `default` namespace with
  `pod-security.kubernetes.io/enforce=privileged` so the snapshot
  agent's privileged pod can be scheduled (Talos enforces `restricted`
  cluster-wide by default).

### `tests/chainsaw/snapshot/deploy-agent-talos/` — chainsaw test

- Invokes `aicr snapshot --os talos -o cm://default/aicr-e2e-snapshot`
  from a `script:` step. The CLI exercises `pkg/k8s/agent.Deployer`
  end-to-end (RBAC creation, OS=talos pod shape from
  `pkg/k8s/agent/job.go`, log streaming, Job completion, ConfigMap
  write, cleanup) — no hand-applied YAML fixtures to drift against
  `job.go` (the existing kind chainsaw test's drift risk is captured
  for follow-up in `issues/e2e-update.md`, kept local).
- Asserts only the surviving ConfigMap content. Uses chainsaw
  JMESPath filters (`(sort([*].type))`,
  `(@[?type == 'X'] | [0].subtypes | sort([*].subtype))`) so the
  assertion is order-independent at both the measurement-type and
  per-type subtype levels — collector goroutine completion order
  is non-deterministic.
- Pinned: 5 measurements; the set of types
  `{GPU, K8s, NodeTopology, OS, SystemD}`; per-type subtype names.
- Per-subtype `data` values are intentionally not pinned to avoid
  Talos / kernel / kubernetes version churn.

### `aicr snapshot` — `--requests` / `--limits`

Adds two CLI flags accepting kubectl-style comma-separated
`name=quantity` lists:

```
--requests cpu=500m,memory=1Gi,ephemeral-storage=1Gi
--limits   cpu=1,memory=2Gi,ephemeral-storage=2Gi
```

Plumbed through `pkg/snapshotter.AgentConfig` and `pkg/k8s/agent.Config`
as `corev1.ResourceList`. `applyPrivilegedSettings` and
`applyRestrictedSettings` merge overrides on top of the existing
hardcoded defaults via a new `mergeResourceList` helper, so unspecified
keys keep the production defaults. The `nvidia.com/gpu` limit added
by `--require-gpu` is preserved on top of any override.

### Make targets

`talos-dev-env`, `talos-dev-env-clean`, `talos-snapshot-test`. None
are wired into `make qualify`; Talos remains opt-in for now.

## Testing

```bash
unset GITLAB_TOKEN
make build
make test     # touched packages
golangci-lint run -c .golangci.yaml ./pkg/k8s/agent/... ./pkg/cli/... ./pkg/snapshotter/...
make lint-yaml
```

End-to-end against a live local cluster (macOS Podman Machine, Talos
v1.9.0, talosctl 1.13.0):

```bash
make image IMAGE_REGISTRY=localhost:5001/aicr IMAGE_TAG=local
make build
make talos-dev-env
make talos-snapshot-test          # PASS
make talos-dev-env-clean
```

Coverage delta on touched packages:

- `pkg/k8s/agent`: 77.8% (baseline) → ~78% (added tests for resource-
  override paths)
- `pkg/cli`: ~52.9% (baseline) → comparable (added flag plumbing
  + `parseResourceList`; covered indirectly)
- `pkg/snapshotter`: ~52.2% (baseline) → comparable (1-line struct
  field + plumbing only)

## Risk Assessment

**Rollout notes:** Production callers of `aicr snapshot` are unaffected;
both new flags default to empty, in which case the hardcoded
privileged/restricted defaults remain in force exactly as before. The
Talos test harness is a new opt-in surface and is not wired into
`make qualify` or any CI workflow.

## Checklist

- [x] Linter passes (`make lint`) — only `go: no such tool covdata` on
      `cmd/aicr`/`cmd/aicrd` (a pre-existing local toolchain issue,
      unrelated to this branch — does not affect CI)


pkg/k8s/agent/job.go
```diff
@@ -164,21 +164,22 @@ func (d *Deployer) applyPrivilegedSettings(spec *corev1.PodSpec) {
...
+		corev1.ResourceEphemeralStorage: mustParseQuantity("2Gi"),
+	}, d.config.Requests)
+	limits := mergeResourceList(corev1.ResourceList{
 		corev1.ResourceCPU:              mustParseQuantity("2"),
 		corev1.ResourceMemory:           mustParseQuantity("8Gi"),
 		corev1.ResourceEphemeralStorage: mustParseQuantity("4Gi"),
-	}
+	}, d.config.Limits)
 	if d.config.RequireGPU {
 		limits[corev1.ResourceName("nvidia.com/gpu")] = mustParseQuantity("1")
```
mchmarny: **Medium:** `RequireGPU=true` silently overrides any `nvidia.com/gpu` value the caller put in `--limits`. Order today:

1. `mergeResourceList(defaults, d.config.Limits)` — caller's `--limits nvidia.com/gpu=4` lands in `limits[gpu]=4`.
2. `if d.config.RequireGPU { limits[gpu] = 1 }` — overwrites to 1.

The CLI help reads "the nvidia.com/gpu limit added by --require-gpu is preserved on top of any override," which on closer reading does describe this — but a user setting `--require-gpu --limits nvidia.com/gpu=4` will likely expect 4. Two options:

- **Treat caller's gpu override as authoritative** when both are set: only inject `gpu=1` if the merged map doesn't already have an `nvidia.com/gpu` key. Then `--require-gpu` becomes "default to 1 unless overridden."
- **Reject the combination** at the CLI layer: error if `--require-gpu` is set AND `--limits` contains `nvidia.com/gpu`. Forces an explicit choice.

I'd lean (1) for ergonomics. Either way, the precedence deserves an explicit test case alongside the existing `RequireGPU adds nvidia.com/gpu limit on top of overrides` to lock it in.

pkg/cli/snapshot.go
```diff
@@ -36,7 +40,49 @@ type snapshotTemplateOptions struct {
...
+		key = strings.TrimSpace(key)
+		value = strings.TrimSpace(value)
+		if key == "" || value == "" {
+			return nil, fmt.Errorf("entry %q has empty name or quantity", entry)
+		}
+		q, err := resource.ParseQuantity(value)
+		if err != nil {
+			return nil, fmt.Errorf("entry %q: %w", entry, err)
+		}
+		result[corev1.ResourceName(key)] = q
```
mchmarny, line 96: nit: duplicate keys silently overwrite. `--requests cpu=1,cpu=2` accepts the second value with no warning. Either reject duplicates here (`if _, ok := result[corev1.ResourceName(key)]; ok { return nil, fmt.Errorf("duplicate key %q", key) }`) or document the last-write-wins behavior. Reject is probably safer — the CLI flag is unlikely to be set by a script that intentionally overrides itself.

tools/talos-test/up.sh
```diff
@@ -0,0 +1,158 @@
...
+# br_netfilter for VXLAN bridging. The Linux VM that backs the
+# Docker-compatible runtime on macOS (Docker Desktop's LinuxKit VM or
+# Podman Machine's Fedora CoreOS VM) ships the module but does not
+# auto-load it; load it once here via a short privileged sidecar.
+# /lib/modules from the VM is bind-mounted so modprobe can find the
+# module map. This is a no-op on Linux hosts where the module is
+# normally already loaded. State persists for the life of the
+# host VM (i.e. until the runtime restarts).
+echo "Ensuring br_netfilter is loaded in the host VM kernel..."
+if ! docker run --rm --privileged --net=host \
```
mchmarny, line 50: nit: `alpine:3` and the `ghcr.io/siderolabs/talos:${TALOS_VERSION}` lower down aren't digest-pinned. Dev tooling, low blast radius, and PR #761 just landed a regression test gating in-tree manifest pins — out of scope for this PR but worth pinning if/when convenient. A breaking upstream `alpine:3` rebuild that ever drops `modprobe` would silently break this script.

Makefile
```diff
@@ -650,6 +650,35 @@ endif
...
+
+.PHONY: talos-dev-env-clean
+talos-dev-env-clean: ## Destroy the local Talos cluster.
+	@TALOS_CLUSTER_NAME=$(TALOS_CLUSTER_NAME) \
+	 ./tools/talos-test/down.sh
+
+.PHONY: talos-snapshot-test
+talos-snapshot-test: build ## Run the Talos snapshot chainsaw test against an already-running cluster.
+	@DIST_DIR=$$(find dist -maxdepth 1 -type d -name 'aicr_*' 2>/dev/null | head -1); \
+	 if [ -z "$$DIST_DIR" ] || [ ! -x "$$DIST_DIR/aicr" ]; then \
```
mchmarny, line 744: nit: `find dist -maxdepth 1 -type d -name 'aicr_*' | head -1` is alphabetical-first, so on a multi-platform build it picks whichever arch sorts first (`aicr_darwin_arm64_v8.0` before `aicr_darwin_amd64_v3`, etc.). Probably fine for a developer-laptop target, but if the wrong arch is picked the resulting binary will silently fail to exec on Talos. Consider scoping with `$(uname -m)` or matching `aicr_$(go env GOOS)_$(go env GOARCH)*`.

tools/talos-test/README.md
mchmarny, line 1: nit: `up.sh` permanently labels the `default` namespace with `pod-security.kubernetes.io/enforce=privileged` (and `audit`/`warn`). That's a meaningful posture change beyond the snapshot test — anything else a developer schedules in `default` afterward also runs at the privileged baseline. Worth a one-line "Side effects" note in this README so it's not a surprise. The script comment explains *why*; the README is the place to surface *that*.

mchmarny (commented):
Solid PR — the Talos OS=talos pod-shape branch from #714 finally has end-to-end coverage, and the `--requests`/`--limits` plumbing is clean (separate types, merge-not-replace semantics, defaults preserved, non-mutating helper, unit tests exercising all three branches).

One medium concern on the `--require-gpu` vs `--limits nvidia.com/gpu=N` precedence — current behavior silently overrides the explicit limit and the CLI help text reads ambiguously. Worth deciding whether `--require-gpu` is "set if unset" or "always force to 1" and asserting it in tests. Plus a few nits on duplicate-key handling, image pinning in dev tooling, multi-arch find logic, and surfacing the namespace PSP relabel in the README. Nothing blocking; CI is still green where complete.

tools/talos-test/up.sh
```diff
@@ -0,0 +1,158 @@
...
+        endpoints:
+          - "http://${MIRROR_HOST}:5001"
+EOF
+
+echo "Creating Talos cluster '${CLUSTER_NAME}' (Talos ${TALOS_VERSION}, registry mirror -> ${MIRROR_HOST}:5001)..."
+talosctl cluster create docker \
+    --name "${CLUSTER_NAME}" \
+    --workers 1 \
+    --image "ghcr.io/siderolabs/talos:${TALOS_VERSION}" \
+    --config-patch "@${PATCH_FILE}"
```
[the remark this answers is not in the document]
mchmarny: This finding is incorrect for the version this PR targets. The web query landed on v1.9 docs, where `talosctl cluster create` was a single command driven by `--provisioner=docker`. Talos restructured this in later releases: starting in v1.10+, the provisioner flag was removed in favor of explicit subcommands (`docker`, `qemu`, `dev`).

Verified directly in the v1.13.0 source — [`cmd/talosctl/cmd/mgmt/cluster/create/cmd_docker.go`](https://github.com/siderolabs/talos/blob/v1.13.0/cmd/talosctl/cmd/mgmt/cluster/create/cmd_docker.go) defines:

```go
createDockerCmd := &cobra.Command{
    Use:   "docker",
    Short: "Create a local Docker based kubernetes cluster",
    ...
}
```

Sibling subcommands `qemu` and `dev` live alongside (`cmd_qemu.go`, `cmd_dev.go`).

The PR description states the author tested with talosctl 1.13.0, and the script header comment explicitly calls out the migration:

> Newer talosctl (>= 1.x) splits provisioners into subcommands; the docker subcommand fixes controlplanes=1 and accepts registry config only via --config-patch (no --registry-mirror flag).

Removing the `docker` token would break the script on the version range the PR targets. Recommend resolving this thread without changes.

mchmarny (approved):

Issue #515, opened 2026-04-09
[Feature]: Support dynamic install-time values in bundle output


lockwobr:
## Feature Summary

Support declaring value paths as dynamic at bundle time so they are omitted from the resolved output and provided at install time. This enables reusable bundles across clusters without rebuilding.

## Problem/Use Case

Component Helm charts sometimes require values that are cluster-specific and not known at bundle time — for example, Alloy needing a cluster name or subnet reference. Today AICR fully resolves all values at bundle time, requiring a separate bundle per cluster. This is the "render per-cluster at build time" pattern that ADR-025 lists as an existing workaround.

For OCI-published bundles, this limitation means you can't share a single immutable image across clusters — cluster-specific values are baked in. Separating static from dynamic values is what makes a bundle reusable as an OCI image: the image contains only the resolved values, and cluster-specific configuration is provided at install time.

## Proposed Solution

Add an opt-in `--dynamic` flag to `aicr bundle` that declares specific value paths as install-time parameters:

```
aicr bundle --recipe recipe.yaml \
  --dynamic alloy:clusterName \
  --dynamic alloy:subnetName
```

For each component with dynamic values, the bundler outputs a `cluster-values.yaml` stub alongside the existing `values.yaml` for the user to fill in:

```
bundle/
├── alloy/
│   ├── values.yaml            # static, resolved at bundle time
│   └── cluster-values.yaml    # dynamic stub, fill in at install time
├── gpu-operator/
│   └── values.yaml            # no dynamic values, nothing extra
```

**Helm deployer:** Works cleanly today — Helm natively merges multiple values files:
```
helm upgrade --install ... -f values.yaml -f cluster-values.yaml
```

**Existing ArgoCD deployer (`--deployer argocd`):** Continues to work as-is for GitOps workflows where the bundle is pushed to a Git repo. Dynamic stubs can be filled in before committing.

**New ArgoCD OCI deployer (`--deployer argocd-oci`):** For the ArgoCD + OCI case, a new deployer would be needed. ArgoCD pulls the image directly so you can't edit files inside it — dynamic values need to be injected through ArgoCD's own mechanisms. This deployer would produce a Helm-based app-of-apps (closer to ADR-025's approach), giving ArgoCD a native way to inject values via `valuesObject` on the root Application.

## Success Criteria

- `--dynamic` flag accepted by `aicr bundle` to declare install-time value paths
[...]

PR #772, opened 2026-05-06
feat(bundler): URL-portable argocd-helm bundle (#664, #665)

### LINKED ISSUE #664: [Feature]: --deployer argocd adopts uniform local-chart layout
## Feature Summary

Migrate `--deployer argocd` to the uniform numbered local-chart layout introduced by sub-issue 1. The branching rule for Application shape is the same as deploy.sh's: **Chart.yaml presence in `NNN-<name>/` → path-based single-source Application; absence → today's multi-source upstream-ref Application.**

Parent: #516. Depends on the base format library (sub-issue 1) and the `--vendor-charts` flag (sub-issue 2).

## Problem / Use Case

Once the base format exists and `--vendor-charts` is available, `--deployer argocd` should consume the same layout as `--deployer helm` — otherwise we've left "format, not deployer" half-done. The base format's Alt B pattern (mixed components emit primary + injected `-post` folders in sub-issue 1) eliminates the mixed-component special case entirely — every folder follows the simple "Chart.yaml present → local chart; absent → upstream ref" rule.

The goal is zero breaking change for current `--deployer argocd` users: today's multi-source Application shape is preserved wherever Chart.yaml is absent.

## Proposed Solution

**Branching rule:** Chart.yaml presence in `NNN-<name>/` → path-based single-source Application (`spec.source.repoURL: <user-git>`, `spec.source.path: NNN-<name>`). Absence → today's multi-source shape (source[0] upstream Helm repo, source[1] values git ref).

**Per-case behavior:**

- **Wrapped Kustomize / raw-manifest-only** (Chart.yaml always present after sub-issue 1): single-source path-based Application.
  - Today's argocd Kustomize Application is already single-source — shape-preserving.
  - **Raw-manifest-only is a fix-forward**: today's argocd deployer silently drops `ComponentManifests` (verified: `bundler.go:301-314` constructs the argocd Generator without `ComponentManifests`). With sub-issue 1's chart-wrapping of raw manifests, these Applications now install successfully via the wrapped chart. Previously broken → now working.
- **Pure Helm, non-vendored** (no Chart.yaml): keep today's multi-source Application unchanged. Zero change for current users.
- **Pure Helm, vendored** (Chart.yaml present): path-based single-source Application. Argo's repo server reads local chart bytes; no `helm dependency update` / no `helm.enableDependencyUpdate` surface.
- **Mixed Helm + raw manifests, non-vendored** (primary + injected `-post` folder from sub-issue 1): **full fix-forward** — each folder gets a normal Application via the Chart.yaml-presence rule. Primary: multi-source upstream ref (today's shape, unchanged). Injected `-post`: single-source path-based, installs the wrapped-manifests chart. Sync-wave on the `-post` Application is primary's wave + 1 so ordering is preserved. Today's silent-drop of `ComponentManifests` becomes deployed manifests as a side effect of the Alt B layout — no warning needed.

**App-of-apps root Application** (`spec.source.directory.include='*/application.yaml'`, recurse) keeps working — the recurse pattern doesn't care about numbering.

## Success Criteria

- Branching on Chart.yaml presence produces the right Application shape across all per-component cases.
- Existing non-vendored `--deployer argocd` users see no behavior change for pure-Helm cases (multi-source Applications preserved).
- Raw-manifest-only components install successfully under `--deployer argocd` (previously silently broken).
- Mixed components (primary + `-post` folders) emit two correctly-ordered Applications; sync-wave on the `-post` is primary + 1. Raw manifests now deploy under `--deployer argocd`.
- Under `--vendor-charts`, pure-Helm components get path-based single-source Applications.
- KWOK e2e against `--deployer argocd` passes (non-vendored + vendored paths covered), including a mixed component.
- `make qualify` clean; test coverage gate satisfied.
- `docs/user/cli-reference.md` updated.

## Alternatives Considered

- **Always emit path-based single-source** (require `helm.enableDependencyUpdate` for non-vendored Helm). Rejected: introduces a version-dependent Argo CD surface we can sidestep.
- **Preserve today's silent-drop for mixed components** with a warning. Rejected: sub-issue 1's Alt B layout (primary + `-post` folders) lets us fix the drop entirely at zero cost to argocd, so warn-and-drop would be a regression vs what's available.
- **Umbrella Chart.yaml with `dependencies:` for mixed components in non-vendored mode.** Rejected: reintroduces `helm dependency update` at deploy time and re-exposes `helm.enableDependencyUpdate` for that subset.

## Component

## Priority

Minimal. Non-vendored `--deployer argocd` users see only file-path renames in the generated bundle (auto-regenerated on every `aicr bundle`; no migration). Application shape unchanged for them. Vendored users see single-source path-based Applications — documented in PR + release notes.

## Operational Considerations

- Raw-manifest-only components newly work under argocd — operators should expect those manifests to now appear in target clusters after the upgrade.
- Mixed components (e.g. gpu-operator) newly have their raw manifests deployed under argocd via the injected `-post` Application. Same expectation as above.
- Parallel installs via a `needs:` release-dependency graph are out of scope here — tracked under [#632](https://github.com/NVIDIA/aicr/issues/632) for the helmfile deployer, which has native `needs:` semantics.

Maybe, with guidance

### LINKED ISSUE #665: [Feature]: --deployer argocdhelm adopts uniform local-chart layout
## Feature Summary

Migrate `--deployer argocdhelm` (Argo CD app-of-apps with dynamic-values injection via root `valuesObject`) to the uniform numbered local-chart layout. Transformation logic simplifies because input is single-source path-based when Chart.yaml is present.

Parent: #516. Depends on `--deployer argocd` adopting the base format (sub-issue 3).

## Problem / Use Case

`--deployer argocdhelm` delegates to the `argocd` deployer and transforms its output — wrapping per-component Applications in a parent Helm chart so ArgoCD can inject cluster-specific values via `valuesObject` on the root Application (OCI-publishable bundles per ADR-025). Once the argocd deployer adopts the base format (sub-issue 3), argocdhelm should follow so the whole app-of-apps chain consumes one layout.

## Proposed Solution

argocdhelm's `transformApplication` handles both input shapes from the argocd deployer:

- **Input is path-based single-source** (wrapped Kustomize / raw-manifest-only, or vendored Helm): transformation shortens — no multi-source → single-source flip needed. Injecting `helm.values: {{ .Values.<key> }}` into the single source is straightforward.
- **Input is multi-source** (non-vendored pure Helm): transformation semantics match today. The existing multi-source → single-source conversion stays in place for this branch.

The `static/` directory convention reconciles with per-component ownership: treat `NNN-<name>/values.yaml` as the static layer; put only dynamic-path stubs in the root chart's `values.yaml`.

## Success Criteria

- argocdhelm's transformation handles both input shapes without branching on component kind.
- Dynamic-values injection via root `valuesObject` keeps working end-to-end.
- Wrapped Kustomize / raw-manifest-only components pass through argocdhelm correctly.
- KWOK e2e against `--deployer argocdhelm` passes for both non-vendored and vendored bundles.
- `make qualify` clean; test coverage gate satisfied.
- `docs/user/cli-reference.md` updated.

## Alternatives Considered

- **Keep argocdhelm on the old multi-source transformation path only.** Rejected: leaves the deployer diverged from the base format once sub-issue 3 lands.
- **Rewrite argocdhelm to generate Applications from scratch** instead of delegating to argocd. Rejected: keeps today's delegation pattern (changes to the argocd deployer flow through automatically) and avoids duplicating Application generation.

## Component

## Priority

File-path renames in the generated bundle (auto-regenerated; no migration). Root chart `values.yaml` now pulls static layer per-component from `NNN-<name>/values.yaml`, with only dynamic stubs at root.

## Operational Considerations

- End-to-end air-gap validation (block upstream chart registries at the Argo repo-server layer) confirms the `--vendor-charts + --deployer argocdhelm` claim.

Maybe, with guidance

lockwobr:
## Summary

Make the argocd-helm bundle a build-once, URL-portable artifact. Closes #664 and #665, plus fixes the original `values: |-` literal-block render bug that was blocking every install.

## Motivation / Context

Three intertwined goals:

1. **Render bug** — argocd-helm emitted `nindent 8` for the merged values block scalar. With `values: |-` at column 12, that placed content *outside* the literal block, so kubectl rejected every Application with "field not declared in schema". The bundle simply did not install.
2. **Manifest-only and mixed components were silently dropped** by `--deployer argocd` and `--deployer argocd-helm`. Recipes that included `nodewright-customizations` or any mixed component (Helm chart + raw manifests like `gpu-operator`'s `dcgm-exporter`) lost those resources entirely.
3. **The bundle baked its own publish URL.** `application.yaml` referenced `oci://<repo>` at *bundle time*, so re-publishing to a different registry required regenerating the bundle.

Fixes: #664, #665
Related: #516 (umbrella), #515 (dynamic install-time values)

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] CLI (`cmd/aicr`, `pkg/cli`) — `--repo` deprecation note for `--deployer argocd-helm`
- [x] Recipe engine / data (`pkg/recipe`) — no logic changes; consumed via `localformat`

## Implementation Notes

**`#664` — argocd deployer adopts NNN-folder layout via `localformat.Write`**

The argocd deployer no longer hand-rolls per-component directories. It calls the same `localformat.Write` the helm deployer uses, getting `NNN-<name>/Chart.yaml + templates/` (or `upstream.env`) for free. argocd then writes a single `application.yaml` *inside* each folder. The Application shape branches on `localformat.FolderKind`:

- **`KindUpstreamHelm`** (no `Chart.yaml`) → today's multi-source upstream-helm shape, unchanged.
- **`KindLocalHelm`** (`Chart.yaml` + `templates/` present) → new path-based single-source shape (`source.path: NNN-<name>`).

This is what makes manifest-only and mixed-component raw manifests deploy: localformat wraps them as local Helm charts, argocd points an Application at them via `path:`.

**`#665` — argocd-helm transforms both shapes**

`transformApplication` detects `spec.sources` (multi) vs `spec.source` (single) and routes to the right transformation:

- Multi-source: today's flip to single-source + `helm.values` merging static (`.Files.Get`) and dynamic (`.Values.<key>`) — with the **`nindent 16` fix** for the original render bug.
- Single-source path-based: new `injectValuesIntoSingleSource` adds a `helm.values` block exposing dynamic `.Values.<key>` overrides (the wrapped chart's own `values.yaml` is the static layer).

**URL-portable bundle**

Parent Application moved from a hard-coded `application.yaml` at bundle root into `templates/aicr-stack.yaml`, a Helm template using `{{ .Values.repoURL }}` / `{{ .Values.targetRevision }}`. Path-based child Applications also use `.Values.repoURL`/`.Values.targetRevision` (replaced via single-quoted yaml.Node directives so embedded `"..."` in `required` survives marshaling). Same artifact bytes work for any registry — URL supplied once at install time:

```bash
aicr bundle -r recipe.yaml --deployer argocd-helm --dynamic gpuoperator:driver.version -o ./bundle
helm package ./bundle -d /tmp/
helm push /tmp/aicr-bundle-*.tgz oci://ghcr.io/<org>
helm install aicr-bundle oci://ghcr.io/<org>/aicr-bundle --version <chart-version> -n argocd \
  --set repoURL=oci://ghcr.io/<org>/aicr-bundle --set targetRevision=<chart-version>
```

**Helm hooks stripped from wrapped manifests**

`localformat.stripHelmHooks` walks rendered manifest YAML and removes `helm.sh/hook*` annotations before they reach `templates/`. Reason: Argo's Helm processor maps `helm.sh/hook: post-install` → `argocd.argoproj.io/hook: PostSync`, treating the resource as a hook. With `syncPolicy.automated`, Argo only runs sync *operations* when something is out-of-sync; a chart whose only resources are hook-annotated reports trivially "Synced" and the hook never fires. Skyhook CRs and similar would silently never apply. The NNN-prefix folder ordering at the bundle layer subsumes the role helm hooks were playing for these wrapped charts.

**OCI URL/version handling**

`argocd.buildApplicationData` now appends the chart-name segment to OCI `repoURL` (mirroring `localformat.writeUpstreamHelmFolder`'s convention) and preserves the `v` prefix on OCI tags — `v1.3.0` is a literal tag in `ghcr.io`, not equivalent to `1.3.0`. Without this, `nvsentinel` and `kai-scheduler` failed with "manifests/<wrong>: not found" against the registry. Recipe data unchanged.

**`ServerSideApply=true` on Application syncOptions**

Several upstream charts ship CRDs that exceed kubectl's 262144-byte client-side annotation cap. `kai-scheduler`'s `configs.kai.scheduler` CRD is **721,579 bytes** — 2.75× over. Argo's default client-side apply writes the full resource JSON into `kubectl.kubernetes.io/last-applied-configuration`, hitting the apiserver's annotation field limit. Without `ServerSideApply=true`, sync retries forever with `metadata.annotations: Too long: may not be more than 262144 bytes`.

## Testing

```bash
make qualify
golangci-lint run -c .golangci.yaml ./pkg/bundler/...
```

**Test surface added in this PR:**

- **Golden fixtures** under both deployers (`testdata/helm_and_manifest_only/`, `testdata/mixed_component/`) freezing the bundle layout. Regenerate with `go test ... -run TestBundleGolden -args -update`.
- **Live `helm template` integration tests** — `TestHelmTemplate_RendersWithSetRepoURL` runs the chart through Helm with `--set repoURL=...` and asserts the rendered parent App + path-based child both pick up the user-supplied URL. `TestHelmTemplate_FailsWithoutRepoURL` verifies the `required` directive fires when `repoURL` is omitted. Both `t.Skip` if `helm` not on `PATH`.
- **`TestBuildApplicationData_OCIHandling`** — table-driven (4 cases): nvsentinel pattern (chart != org), kai-scheduler pattern (chart == org), trailing-slash edge case, HTTPS pass-through.
- **`TestStripHelmHooks`** — table-driven (4 cases) plus an empty-input pass-through test.
- **`TestInjectValuesIntoSingleSource`** extended to assert URL-portability invariants (templated `repoURL`/`targetRevision`, single-quoted-style preserved, baked-in input URL absent from output).
- Existing kustomize tests `t.Skip`-ed — they need testdata fixtures and assertion rewrites for the wrapped-chart layout. Tracked under #664 follow-up.

**End-to-end verified** against a fresh KWOK cluster (`make kwok-cluster && make kwok-nodes RECIPE=h100-eks-ubuntu-training`):

- 14 child Applications + parent `aicr-stack` deploy via `helm install ... --set repoURL=...`.
- 12 of 14 children Synced (9 Healthy, 3 Progressing). The 3 OutOfSync are field-level drift from KWOK/operator post-deploy reconciliation, not deployment failures.
- Skyhook CR (`tuning`) auto-applied without manual sync — helm-hook strip working.
- `nvsentinel` and `kai-scheduler` (previously `Unknown` due to OCI handling) now resolve correctly.
- `kai-scheduler` and `kube-prometheus-stack` install cleanly via `ServerSideApply=true` (previously stuck in retry loops).

## Risk Assessment

- [ ] Low
- [ ] High

**Rollout notes:**

Three breaking changes; documented in the user-facing CLI reference and the bundle's generated README:

1. **`--deployer argocd` bundle layout** — flat `<name>/application.yaml + values.yaml` → `NNN-<name>/{Chart.yaml or upstream.env, application.yaml, values.yaml, cluster-values.yaml, ...}`. Pure-Helm Application *shape* unchanged; manifest-only and mixed components newly deploy (previously silently broken). External tools that parsed bundle paths by bare component name need to account for the `NNN-` prefix.

2. **`--deployer argocd-helm` bundle root no longer emits `application.yaml`** — the parent Application is `templates/aicr-stack.yaml` and is rendered by Helm at install time using `.Values.repoURL` / `.Values.targetRevision`. Users who previously did `kubectl apply -f bundle/application.yaml` will instead `helm install aicr-bundle <chart-source> --set repoURL=... --set targetRevision=...`.

3. **`--repo` flag is now ignored by `--deployer argocd-helm`** (the URL is supplied at install time, not at bundle time). Help text updated; flag still functional for `--deployer argocd`.

No migration script needed — bundles auto-regenerate on every `aicr bundle`. Existing in-cluster Applications keep working until a re-bundle/redeploy.

## Checklist

- [x] Linter passes (`make lint`) — Go lint clean; yamllint hits a local-worktree-only false positive
(`.claude/worktrees/`), CI unaffected
- [x] Commits are cryptographically signed (`git commit -S`) — *you'll handle on commit*


tests/chainsaw/cli/bundle-dynamic/chainsaw-test.yaml
```diff
@@ -202,6 +210,28 @@ spec:
...
+        a chart whose Application would point at an empty URL.
+      try:
+        - script:
+            content: |
+              WORK="/tmp/chainsaw-bundle-dynamic"
+              if output=$(helm template test-release "${WORK}/argocd-dynamic" 2>&1); then
+                echo "expected helm template to fail without --set repoURL, but it succeeded:" >&2
+                echo "$output" >&2
+                exit 1
+              fi
```
mchmarny, line 240: These two new steps shell out to `helm template` but neither `tools/e2e` (`has_tools chainsaw make uname goreleaser`) nor `.github/actions/cli-e2e/action.yml` install or require helm. Passes today because `ubuntu-latest` ships helm pre-installed; if that ever changes, both steps fail. Either guard each step with a `command -v helm >/dev/null || exit 0` skip, or add helm to `has_tools` / the cli-e2e action.

docs/user/cli-reference.md
```diff
@@ -1128,16 +1128,64 @@ bundles/
...
+├── 004-nodewright-customizations/      # KindLocalHelm (manifest-only)
+│   ├── Chart.yaml
+│   ├── templates/
+│   └── values.yaml
 └── README.md
 ```
 
+Manifest-only components and mixed-component raw manifests are supported by `--deployer argocd-helm` via the path-based Application shape.
+
+**The bundle is URL-portable.** No `--repo` flag is needed (and is ignored if passed with `--deployer argocd-helm`). The same generated bundle bytes can be pushed to any chart-source backend the user chooses — Argo CD pulls from whichever URL the user supplies at install time via `helm install --set repoURL=...`. The publish location is *not* baked into the bundle artifact.
```
mchmarny, line 1226: This (correct) prose contradicts the unchanged flag table at L839 which still says `--repo` is "used with `--deployer argocd` and `--deployer argocd-helm`". Worth fixing the table row in the same commit so help-table and prose agree.

pkg/cli/bundle.go
```diff
@@ -361,9 +361,11 @@ Package with explicit tag (overrides CLI version):
...
 			}, config.GetDeployerTypes),
 			&cli.StringFlag{
-				Name:     "repo",
-				Value:    "",
-				Usage:    "Git repository URL for Argo CD applications (only used with --deployer argocd)",
+				Name:  "repo",
+				Value: "",
+				Usage: "Git repository URL for Argo CD Applications (only used with --deployer argocd). " +
+					"Ignored by --deployer argocd-helm: that bundle is URL-portable and the publish " +
+					"location is supplied at install time via `helm install --set repoURL=...`.",
```
mchmarny, line 392: Help text correctly says "Ignored by `--deployer argocd-helm`" but a user passing both gets no runtime signal — the value silently disappears (never referenced after `parseBundleCmdOptions` for the argocd-helm path). Worth a `slog.Warn` somewhere along the deployer-routing path when `opts.deployer == DeployerArgoCDHelm && cmd.IsSet("repo")` so the discrepancy between intent and effect is visible.

mchmarny (commented):
Solid rewrite — fixes three real bugs (literal-block render, silent-drop of manifest-only/mixed components, baked-in URLs) with strong test coverage including live `helm template` round-trips and a fail-without-repoURL negative test. Architecture of delegating to the `argocd.Generator` and wrapping is clean; `AllowDynamicValueSplit` is the right opt-in boundary; the `stripHelmHooks` rationale doc-comment is exemplary.

Three findings worth addressing — none block merge:

1. New chainsaw helm-template steps assume `helm` is on PATH.
2. `--repo` silently disappears for `--deployer argocd-helm` at runtime; help text mentions it but no log line.
3. `cli-reference.md` flag table contradicts the new prose on `--repo` portability.

CI is green.

tests/chainsaw/cli/bundle-dynamic/chainsaw-test.yaml
lockwobr: Fixed.

docs/user/cli-reference.md
lockwobr: Fixed.

pkg/cli/bundle.go
lockwobr: Fixed.

mchmarny (dismissed):

mchmarny (approved):

PR #782, opened 2026-05-06
feat(cli): unified --config flag for recipe and bundle

### LINKED ISSUE #781: feat(cli): unified --config flag for recipe and bundle commands
## Summary

Both `aicr recipe` and `aicr bundle` have accumulated many flags (~9 and ~22 respectively). Introduce a single `--config <path-or-URL>` flag accepting a YAML/JSON document that can configure either or both commands. Deprecate the existing `--criteria` flag in favor of `--config`.

> Note: this is in addition to the flags, it does not replace them. 

## Problem

- `aicr bundle` invocations span 5+ flag categories (input, output, deployment, scheduling, attestation, OCI registry). Long command lines are painful to type, hard to review, and not easy to commit alongside a recipe.
- The recipe→bundle workflow is often reproduced (CI pipelines, environment promotion, sharing "blessed" configurations across teams) but has no canonical artifact today.
- The existing `--criteria` flag is partial — only the `recipe` command, only the criteria fields.

## Proposed solution

A single shared config file format, modeled after the existing Kubernetes-style envelope used elsewhere in the project (`RecipeCriteria`, `RecipeMixin`, `Snapshot`):

```yaml
kind: AICRConfig
apiVersion: aicr.nvidia.com/v1alpha1
metadata:
  name: gb200-eks-training-prod
spec:
  recipe:
    criteria:
      service: eks
      accelerator: gb200
      intent: training
      os: ubuntu
      platform: kubeflow
      nodes: 8
    input:
      snapshot: ./snapshot.yaml
    output:
      path: ./recipe.yaml
      format: yaml
    data: ./external-data

  bundle:
    input:
      recipe: ./recipe.yaml
    output:
      target: ./my-bundle
      imageRefs: ./refs.txt
    deployment:
      deployer: helm
      repo: ""
      set:
        - gpuoperator:driver.version=570.86.16
      dynamic:
        - alloy:clusterName
    scheduling:
      systemNodeSelector:
        role: system
      acceleratedNodeSelector:
        nodeGroup: gpu-nodes
      acceleratedNodeTolerations:
        - "nvidia.com/gpu=present:NoSchedule"
      workloadGate: "nvidia.com/training=true:NoSchedule"
      nodes: 8
      storageClass: gp3
    attestation:
      enabled: false
      certificateIdentityRegexp: ""
      oidcDeviceFlow: false
    registry:
      insecureTLS: false
      plainHTTP: false
```

## Behavior

- **Sources:** `--config` accepts a local file path or `http(s)://` URL. No `cm://` (ConfigMap) support — users with configs in ConfigMaps can `kubectl get cm -o yaml` and pass the file.
- **Sections:** A single file can populate both `spec.recipe` and `spec.bundle`. Each command reads only its own section. The recipe→bundle handoff still happens explicitly via `--recipe` (or `spec.bundle.input.recipe`) — no implicit handoff.
- **Precedence:** CLI flags override config file values. For slice/map flags, presence on the CLI **replaces** the file's value (no append). Override events are logged at INFO level, mirroring today's snapshot-override behavior.
- **Secrets:** `identityToken` is intentionally not part of the schema — must be supplied via `COSIGN_IDENTITY_TOKEN` or `--identity-token`.
- **Deprecation:** `--criteria` is removed. `--config` is the sole file-based input. (Pre-1.0 / early days, no compatibility shim.)

## Out of scope (future)

- Combined `aicr workflow`-style command that runs recipe + bundle in one step.
- ConfigMap (`cm://`) source.
- Environment-variable interpolation in config values.

## Documentation

- `docs/user/cli-reference.md` — `--config` on both commands; remove `--criteria`.
- `docs/integrator/recipe-development.md` — `AICRConfig` examples for the recipe → bundle workflow.

mchmarny:
## Summary

Introduce `--config <path-or-URL>` on `aicr recipe` and `aicr bundle` accepting a single `AICRConfig` document that can drive either or both commands. Replaces (and removes) the recipe-only `--criteria` flag.

## Motivation / Context

Both `aicr recipe` and `aicr bundle` have accumulated many flags (~9 and ~22). Long invocations are tedious to type, hard to review, and don't fit well into CI pipelines, environment promotion, or shared "blessed" templates. The existing `--criteria` flag is a partial solution scoped to one command and one section of inputs. Design discussion captured in #781.

Fixes: #781

## Type of Change

- [x] Breaking change (`--criteria` flag is removed; pre-1.0, no shim)

## Component(s) Affected

- [x] Other: new `pkg/config` package

## Implementation Notes

Schema: Kubernetes-style envelope (`kind: AICRConfig`, `apiVersion: aicr.nvidia.com/v1alpha1`) with `spec.recipe` and `spec.bundle` sections — each command reads only its own section, no implicit handoff.

Precedence: CLI flags always override config file values. For slice/map flags, CLI presence **replaces** the file's value (no append). Override events log at INFO so the winning input is visible.

Sources: file path or `http(s)://` URL. ConfigMap (`cm://`) sources are intentionally rejected with a tailored error pointing at `kubectl get cm -o yaml`. ConfigMap support drops a kubeconfig dependency from config loading.

Secrets: cosign identity token is **never** read from the config file; supplied via `COSIGN_IDENTITY_TOKEN` or `--identity-token` only.

Default-flag semantics: `cmd.IsSet()` (not "non-empty value") gates flag-overrides-config so default flag values like `--deployer helm` and `--output .` don't shadow user-provided config values.

Shared CLI flag instances (`output`, `format`, `kubeconfig`, `data`, `config`) are now functions rather than vars, mirroring the existing pattern for `formatFlag`. urfave/cli mutates parsed-state on the Flag value, so a shared instance leaks `Count` and parsed values across successive `Run` invocations — visible in tests that build multiple command trees.

Snapshot+config interaction: when a snapshot supplies criteria, config fields fill in only the slots the snapshot left wildcarded; CLI flags override on top. This preserves today's snapshot-override behavior while making config a clean middle layer.

## Testing

```bash
# Codebase qualification completed
```

Coverage on changed packages:

| Package | Baseline (origin/main) | Branch | Delta |
|---|---|---|---|
| `pkg/cli` | 52.2% | 63.1% | +10.9pp |
| `pkg/config` | (new) | 98.4% | — |

New tests:
- `pkg/config/{config,loader}_test.go` — schema validation, file/HTTP loading, kind/apiVersion checks, enum validation, mutually-exclusive snapshot+criteria, ConfigMap rejection, canceled context.
- `pkg/cli/config_helpers_test.go` — table-driven tests for `stringFlagOrConfig`, `intFlagOrConfig`, `boolFlagOrConfig`, `stringSliceFlagOrConfig`, `resolveNodeSelector` (incl. defensive-copy semantics and CLI-replaces-config rule).
- `pkg/cli/config_integration_test.go` — recipe/bundle CLI integration: applies criteria, flag overrides config, recipe-from-config, missing-recipe error, slice replace.
- `pkg/cli/config_e2e_test.go` — regression tests proving `--criteria` is rejected; full bundle option resolution from a config exercising every section; CLI override of every section; HTTP `--config` source; partial criteria + flag merge; bad-enum rejection per command; recipe→bundle E2E sharing the same config file.
- `tests/chainsaw/cli/config-file/` — chainsaw e2e (renamed from `criteria-file`) covering valid YAML/JSON, CLI override, invalid kind/apiVersion/enum, non-existent file, partial criteria, bundle-from-config-derived-recipe.

All 18 chainsaw tests pass; full unit suite green with `-race`.

## Risk Assessment

- [x] **Medium** — Touches multiple commands and changes user-facing behavior

**Rollout notes:** `--criteria` is removed in this PR (pre-1.0, no compatibility shim — see issue #781 discussion). Existing `criteria.yaml` files do not migrate automatically; users must wrap content in `spec.recipe.criteria:` and change `kind: RecipeCriteria` → `kind: AICRConfig`. The CLI's `--config` flag is the documented replacement; example file in `docs/user/cli-reference.md`. The API server's `RecipeCriteria` request body is unchanged (separate concern).

## Checklist

### DIFF
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -269,43 +269,54 @@ aicr recipe [flags]
 
 **Modes:**
 
-#### Criteria File Mode (Recommended)
+#### Config File Mode (Recommended)
 
-Generate recipes using a Kubernetes-style criteria file:
+Generate recipes using an `AICRConfig` document. The same file format also drives the `bundle` command, so a single file can describe an end-to-end recipe-to-bundle workflow.
 
 **Flags:**
 | Flag | Short | Type | Description |
 |------|-------|------|-------------|
-| `--criteria` | `-c` | string | Path to criteria file (YAML/JSON), alternative to individual flags |
+| `--config` | | string | Path or HTTPS URL to an AICRConfig file (YAML/JSON) |
 | `--output` | `-o` | string | Output file (default: stdout) |
 | `--format` | `-f` | string | Format: json, yaml (default: yaml) |
 | `--data` | | string | External data directory to overlay on embedded data (see [External Data](#external-data-directory)) |
 
-The criteria file uses a Kubernetes-style format:
+The config file uses a Kubernetes-style envelope:
 ```yaml
-kind: RecipeCriteria
+kind: AICRConfig
 apiVersion: aicr.nvidia.com/v1alpha1
 metadata:
   name: gb200-eks-ubuntu-training
 spec:
-  service: eks
-  os: ubuntu
-  accelerator: gb200
-  intent: training
-  nodes: 8
+  recipe:
+    criteria:
+      service: eks
+      os: ubuntu
+      accelerator: gb200
+      intent: training
+      nodes: 8
+    output:
+      path: recipe.yaml
+      format: yaml
 ```
 
-Individual CLI flags can override criteria file values:
+Individual CLI flags always override config file values. For slice/map flags, presence on the CLI replaces the file's value (no append).
+
 ```shell
-# Load criteria from file
-aicr recipe --criteria criteria.yaml
+# Load criteria from config file
+aicr recipe --config config.yaml
 
 # Override service from file
-aicr recipe --criteria criteria.yaml --service gke
+aicr recipe --config config.yaml --service gke
 
 # Save output to file
-aicr recipe -c criteria.yaml -o recipe.yaml
+aicr recipe --config config.yaml -o recipe.yaml
+
+# Load config from a URL (e.g. CI shared template)
+aicr recipe --config https://team.example.com/configs/eks-h100-training.yaml
 ```
+
+`--config` accepts a local file path or an HTTP/HTTPS URL. ConfigMap (`cm://`) sources are not supported; export the data with `kubectl get cm <name> -o yaml` and pass the resulting file.
 
 #### Query Mode
 
@@ -833,7 +844,8 @@ aicr bundle [flags]
 **Flags:**
 | Flag | Short | Type | Description |
 |---------------------------------|-------|------|-------------|
-| `--recipe` | `-r` | string | Path to recipe file (required) |
+| `--recipe` | `-r` | string | Path to recipe file (required, or via `spec.bundle.input.recipe` in `--config`) |
+| `--config` | | string | Path or HTTPS URL to an AICRConfig file (YAML/JSON). CLI flags override values from this file. See [Config File Mode](#config-file-mode-1). |
 | `--output` | `-o` | string | Output directory (default: current dir) |
 | `--deployer` | `-d` | string | Deployment method: `helm` (default), `argocd`, or `argocd-helm` |
 | `--repo` | | string | Git repository URL for Argo CD applications (used with `--deployer argocd` and `--deployer argocd-helm`) |
@@ -856,6 +868,50 @@ aicr bundle [flags]
 | `--certificate-identity-regexp` | | string | Override the certificate identity pattern for binary attestation verification. Must contain `"NVIDIA/aicr"`. For testing only. |
 | `--identity-token` | | string | Pre-fetched OIDC identity token for `--attest` keyless signing. Skips ambient/browser/device-code flows. Prefer `COSIGN_IDENTITY_TOKEN` on shared hosts — flag values are visible in `ps` and `/proc/<pid>/cmdline`. |
 | `--oidc-device-flow` | | bool | Use the OAuth 2.0 device authorization grant for `--attest` instead of opening a browser callback. Useful on headless hosts that can still reach Sigstore (`--identity-token` and CI ambient OIDC are alternatives). Also reads `AICR_OIDC_DEVICE_FLOW`. |
+
+#### Config File Mode
+
+The bundle command accepts the same `AICRConfig` format used by `aicr recipe`. A single file can populate both `spec.recipe` and `spec.bundle`, capturing an end-to-end workflow that can be committed to git, fetched from CI, or shared across environments.
+
+```yaml
+kind: AICRConfig
+apiVersion: aicr.nvidia.com/v1alpha1
+spec:
+  bundle:
+    input:
+      recipe: ./recipe.yaml
+    output:
+      target: oci://ghcr.io/example/bundle:v1.0.0
+    deployment:
+      deployer: argocd
+      repo: https://example.git/charts
+      set:
+        - gpuoperator:driver.version=570.86.16
+    scheduling:
+      systemNodeSelector:
+        role: system
+      acceleratedNodeTolerations:
+        - "nvidia.com/gpu=present:NoSchedule"
+      nodes: 8
+      storageClass: gp3
+    attestation:
+      enabled: false
+    registry:
+      insecureTLS: false
+      plainHTTP: false
+```
+
+```shell
+# Drive the bundle entirely from a config file
+aicr bundle --config bundle.yaml
+
+# Override the deployer for a one-off run
+aicr bundle --config bundle.yaml --deployer helm
+```
+
+CLI flags always override values loaded from `--config`. For slice/map flags (`--set`, `--dynamic`, `--system-node-selector`, etc.), CLI presence replaces the config's value rather than appending. Override events are logged at INFO so users can see which input won.
+
+**Secrets:** the cosign identity token is never read from a config file; supply it via `--identity-token` or `COSIGN_IDENTITY_TOKEN`.
 
 #### Node Scheduling
 
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -274,14 +274,16 @@ aicr recipe [flags]
 Generate recipes using an `AICRConfig` document. The same file format also drives the `bundle` command, so a single file can describe an end-to-end recipe-to-bundle workflow.
 
 **Flags:**
+
 | Flag | Short | Type | Description |
 |------|-------|------|-------------|
-| `--config` | | string | Path or HTTPS URL to an AICRConfig file (YAML/JSON) |
+| `--config` | | string | Path or HTTP/HTTPS URL to an AICRConfig file (YAML/JSON) |
 | `--output` | `-o` | string | Output file (default: stdout) |
 | `--format` | `-f` | string | Format: json, yaml (default: yaml) |
 | `--data` | | string | External data directory to overlay on embedded data (see [External Data](#external-data-directory)) |
 
 The config file uses a Kubernetes-style envelope:
+
 ```yaml
 kind: AICRConfig
 apiVersion: aicr.nvidia.com/v1alpha1
@@ -845,7 +847,7 @@ aicr bundle [flags]
 | Flag | Short | Type | Description |
 |---------------------------------|-------|------|-------------|
 | `--recipe` | `-r` | string | Path to recipe file (required, or via `spec.bundle.input.recipe` in `--config`) |
-| `--config` | | string | Path or HTTPS URL to an AICRConfig file (YAML/JSON). CLI flags override values from this file. See [Config File Mode](#config-file-mode-1). |
+| `--config` | | string | Path or HTTP/HTTPS URL to an AICRConfig file (YAML/JSON). CLI flags override values from this file. See [Bundle Config File Mode](#bundle-config-file-mode). |
 | `--output` | `-o` | string | Output directory (default: current dir) |
 | `--deployer` | `-d` | string | Deployment method: `helm` (default), `argocd`, or `argocd-helm` |
 | `--repo` | | string | Git repository URL for Argo CD applications (used with `--deployer argocd` and `--deployer argocd-helm`) |
@@ -869,9 +871,11 @@ aicr bundle [flags]
 | `--identity-token` | | string | Pre-fetched OIDC identity token for `--attest` keyless signing. Skips ambient/browser/device-code flows. Prefer `COSIGN_IDENTITY_TOKEN` on shared hosts — flag values are visible in `ps` and `/proc/<pid>/cmdline`. |
 | `--oidc-device-flow` | | bool | Use the OAuth 2.0 device authorization grant for `--attest` instead of opening a browser callback. Useful on headless hosts that can still reach Sigstore (`--identity-token` and CI ambient OIDC are alternatives). Also reads `AICR_OIDC_DEVICE_FLOW`. |
 
-#### Config File Mode
+#### Bundle Config File Mode
 
 The bundle command accepts the same `AICRConfig` format used by `aicr recipe`. A single file can populate both `spec.recipe` and `spec.bundle`, capturing an end-to-end workflow that can be committed to git, fetched from CI, or shared across environments.
+
+When both `spec.recipe.output.path` and `spec.bundle.input.recipe` are set, they must reference the same path; otherwise loading fails fast.
 
 ```yaml
 kind: AICRConfig
--- a/pkg/cli/bundle.go
+++ b/pkg/cli/bundle.go
@@ -192,10 +192,12 @@ func parseBundleCmdOptions(cmd *cli.Command, cfg *appcfg.AICRConfig) (*bundleCmd
 	}
 
 	if opts.systemNodeSelector, err = resolveNodeSelector(cmd, "system-node-selector", bs.SystemNodeSelector()); err != nil {
-		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid system node selector", err)
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+			"invalid "+sourceLabel(cmd, "system-node-selector", "spec.bundle.scheduling.systemNodeSelector"), err)
 	}
 	if opts.acceleratedNodeSelector, err = resolveNodeSelector(cmd, "accelerated-node-selector", bs.AcceleratedNodeSelector()); err != nil {
-		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid accelerated node selector", err)
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+			"invalid "+sourceLabel(cmd, "accelerated-node-selector", "spec.bundle.scheduling.acceleratedNodeSelector"), err)
 	}
 
 	if opts.systemNodeTolerations, err = snapshotter.ParseTolerations(stringSliceFlagOrConfig(cmd, "system-node-toleration", bs.SystemNodeTolerations())); err != nil {
@@ -213,7 +215,8 @@ func parseBundleCmdOptions(cmd *cli.Command, cfg *appcfg.AICRConfig) (*bundleCmd
 	}
 
 	if opts.workloadSelector, err = resolveNodeSelector(cmd, "workload-selector", bs.WorkloadSelector()); err != nil {
-		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid workload selector", err)
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+			"invalid "+sourceLabel(cmd, "workload-selector", "spec.bundle.scheduling.workloadSelector"), err)
 	}
 
 	// Estimated node count for bundle; 0 = unset.
--- a/pkg/cli/bundle.go
+++ b/pkg/cli/bundle.go
@@ -295,11 +295,11 @@ Package with explicit tag (overrides CLI version):
 			&cli.StringFlag{
 				Name:    "output",
 				Aliases: []string{"o"},
-				Value:   ".",
-				Usage: `Output target: local directory path or OCI registry URI.
+				Usage: `Output target: local directory path or OCI registry URI (default: current dir).
 	For local output: ./my-bundle or /tmp/bundle
 	For OCI registry: oci://ghcr.io/nvidia/bundle:v1.0.0
-	If no tag specified, CLI version is used (e.g., oci://ghcr.io/nvidia/bundle)`,
+	If no tag specified, CLI version is used (e.g., oci://ghcr.io/nvidia/bundle)
+	May also be supplied via spec.bundle.output.target in --config.`,
 				Category: "Output",
 			},
 			&cli.StringSliceFlag{
@@ -362,8 +362,7 @@ Package with explicit tag (overrides CLI version):
 			withCompletions(&cli.StringFlag{
 				Name:     "deployer",
 				Aliases:  []string{"d"},
-				Value:    string(config.DeployerHelm),
-				Usage:    fmt.Sprintf("Deployment method (e.g. %s)", strings.Join(config.GetDeployerTypes(), ", ")),
+				Usage:    fmt.Sprintf("Deployment method (default: helm; e.g. %s)", strings.Join(config.GetDeployerTypes(), ", ")),
 				Category: "Deployment",
 			}, config.GetDeployerTypes),
 			&cli.StringFlag{
--- a/pkg/cli/bundle.go
+++ b/pkg/cli/bundle.go
@@ -87,19 +87,19 @@ type bundleCmdOptions struct {
 //
 //nolint:gocyclo // option resolution is inherently long but linear
 func parseBundleCmdOptions(cmd *cli.Command, cfg *appcfg.AICRConfig) (*bundleCmdOptions, error) {
-	bs := bundleSpec(cfg)
+	bs := cfg.Bundle()
 
 	opts := &bundleCmdOptions{
-		recipeFilePath:            stringFlagOrConfig(cmd, "recipe", bundleRecipeInput(bs)),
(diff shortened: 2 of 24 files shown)


docs/user/cli-reference.md
```diff
@@ -857,6 +869,50 @@ aicr bundle [flags]
...
+#### Config File Mode
+
+The bundle command accepts the same `AICRConfig` format used by `aicr recipe`. A single file can populate both `spec.recipe` and `spec.bundle`, capturing an end-to-end workflow that can be committed to git, fetched from CI, or shared across environments.
+
+```yaml
+kind: AICRConfig
+apiVersion: aicr.nvidia.com/v1alpha1
+spec:
+  bundle:
+    input:
```
ayuskauskas, line 885: Since the config file can have both recipe and bundle; in the case where they have both have it: do we still set `recipe.output` and `bundle.input`? I think so since they can be run at separate times but then do we validate that they are the same paths to avoid a silly mistake of not having them be the same and having a broken chain?

ayuskauskas:
This is a good change, I like it. I will continue to review in a bit.

docs/user/cli-reference.md
mchmarny: ye, the param is the same in both, you are right though, never both currently. I guess the goal here was to support workflow-like approach in the future where you could define the output from one as input to the other hence in both places.

mchmarny: Good catch — added cross-section validation in `pkg/config/validate.go` (`Spec.validateRecipeBundleHandoff`): when both `spec.recipe.output.path` and `spec.bundle.input.recipe` are set, they must reference the same path; mismatch is rejected at load time with a clear error. Either side may also be set alone for the run-at-different-times case. Tests in `TestValidate_RecipeBundleHandoff`. Landed in commit c2dfa7d.

pkg/cli/bundle_config.go
```diff
@@ -0,0 +1,253 @@
...
+	return copyStringMap(fallback), nil
+}
+
+// bundleSpec returns cfg.Spec.Bundle if cfg is non-nil, else nil.
+func bundleSpec(cfg *appcfg.AICRConfig) *appcfg.BundleSpec {
+	if cfg == nil {
+		return nil
+	}
+	return cfg.Spec.Bundle
+}
```
lockwobr: Consider moving `bundleSpec` plus everything below it onto the types in `pkg/config` as nil-receiver methods, e.g.:

```go
// pkg/config/bundle_accessors.go
func (c *AICRConfig) Bundle() *BundleSpec {
    if c == nil {
        return nil
    }
    return c.Spec.Bundle
}

func (b *BundleSpec) RecipeInput() string {
    if b == nil || b.Input == nil {
        return ""
    }
    return b.Input.Recipe
}

func (b *BundleSpec) DeploymentSet() []string {
    if b == nil || b.Deployment == nil {
        return nil
    }
    return slices.Clone(b.Deployment.Set)
}
// ...etc
```

Call sites then become:

```go
b := cfg.Bundle()         // nil-safe
recipe := b.RecipeInput()
sets   := b.DeploymentSet()
```

Net ~125 lines deleted from `pkg/cli`, the nil-tolerance contract lives with the type that owns the shape, and CLI-layer code stops importing a wall of `bundleXxx` helper names.

pkg/cli/bundle_config.go
```diff
@@ -0,0 +1,253 @@
...
+// explicitly empty slice yields an empty (non-nil) slice — preserving the
+// caller's distinction between "unset" and "explicitly cleared".
+func copyStrings(s []string) []string {
+	if s == nil {
+		return nil
+	}
+	out := make([]string, len(s))
+	copy(out, s)
+	return out
+}
```
lockwobr: `copyStrings` is exactly `slices.Clone` from the stdlib — same nil-vs-empty preservation (`slices.Clone(nil) == nil`, `slices.Clone([]T{})` returns a non-nil empty slice). Suggest deleting this and using `slices.Clone` directly at call sites.

pkg/cli/bundle_config.go
```diff
@@ -0,0 +1,253 @@
...
+func copyStringMap(m map[string]string) map[string]string {
+	if m == nil {
+		return nil
+	}
+	out := make(map[string]string, len(m))
+	for k, v := range m {
+		out[k] = v
+	}
+	return out
+}
```
lockwobr: Same — `maps.Clone` from the stdlib has identical behavior including the nil-vs-empty distinction. Suggest deleting and using `maps.Clone` directly.

pkg/cli/bundle.go
```diff
@@ -81,19 +82,29 @@ type bundleCmdOptions struct {
 	imageRefsPath string // Path to write published image references (like ko --image-refs)
 }
 
-// parseBundleCmdOptions parses and validates command options.
-func parseBundleCmdOptions(cmd *cli.Command) (*bundleCmdOptions, error) {
+// parseBundleCmdOptions parses and validates command options. When cfg is
+// non-nil, spec.bundle fields supply defaults; CLI flags always override.
+//
+//nolint:gocyclo // option resolution is inherently long but linear
+func parseBundleCmdOptions(cmd *cli.Command, cfg *appcfg.AICRConfig) (*bundleCmdOptions, error) {
```
lockwobr, line 89: Bigger-picture observation about the wire-type / domain-type split: the new string-typed wire types in `pkg/config` (`CriteriaSpec`, `BundleSpec`, `DeploymentSpec`, etc.) sit alongside the existing strict domain types (`recipe.Criteria` with typed enums, `config.DeployerType`, `config.ComponentPath`, `corev1.Toleration`, `corev1.Taint`). The wire→domain conversion is currently scattered across three places and duplicates work:

1. **`pkg/config/validate.go`** parses each field just to throw the result away — `recipe.ParseCriteriaServiceType(c.Service)` etc., result discarded.
2. **`pkg/cli/recipe.go` `applyCriteriaFromConfig`** re-parses the same fields, this time keeping the result, and interleaves an override-logging concern (`logCriteriaOverride`). Then `applyCriteriaOverrides` does the same shape *again* for CLI flags — three near-identical implementations of "convert a string to `CriteriaServiceType` and log overrides."
3. **`pkg/cli/bundle.go` `parseBundleCmdOptions`** pulls strings through 17 `bundleXxx(bs)` accessors, then converts each with the appropriate parser (`ParseDeployerType`, `ParseValueOverrides`, `ParseDynamicValues`, `snapshotter.ParseTolerations`, `snapshotter.ParseTaint`, `resolveNodeSelector`).

The missing layer is a `Resolve()` (or `ToDomain()`) method per spec type that does the conversion *once*:

```go
func (r *RecipeSpec) ResolveCriteria() (*recipe.Criteria, error) {
    if r == nil || r.Criteria == nil {
        return recipe.NewCriteria(), nil
    }
    // ...parse fields, populate, return
}

type BundleResolved struct {
    Deployer                config.DeployerType
    OutputTarget            *oci.Reference
    ValueOverrides          []config.ComponentPath
    DynamicValues           []config.ComponentPath
    SystemNodeSelector      map[string]string
    SystemNodeTolerations   []corev1.Toleration
    AcceleratedNodeSelector map[string]string
    // ...
}
func (b *BundleSpec) Resolve() (*BundleResolved, error) {
    if b == nil {
        return &BundleResolved{}, nil   // zero struct, NOT nil
    }
    // ...
}
```

**Receiver / return contract** worth being explicit about:
- **Nil-receiver tolerant** — same reason as the accessor refactor; callers don't nil-check optional wire-type sections.
- **Never return a nil pointer** — return `&BundleResolved{}` for nil input. Callers will reach into fields, so handing back `nil` just relocates the NPE.
- **Preserve the nil-vs-empty contract** carried by the accessors: nil `b.Scheduling` → `SystemNodeSelector` is `nil`; explicitly empty `scheduling.systemNodeSelector: {}` in YAML → empty non-nil map. Don't collapse that distinction during conversion — "user explicitly cleared this" is meaningful intent.

Downstream effects:
- `Validate()` becomes `_, err := c.Spec.Recipe.ResolveCriteria(); err` — parse-once, no duplication, validation IS conversion.
- `applyCriteriaFromConfig` becomes "call `ResolveCriteria`, copy non-zero fields onto `criteria`, log overrides" — the override-logging logic separates cleanly from the parse logic.
- `parseBundleCmdOptions` starts from a `*BundleResolved` instead of pulling strings through 17 accessors; the CLI layer just merges flag overrides on top of typed values.

Net effect: wire types stay forgiving (strings), domain types stay strict, and the conversion happens in exactly one place per spec.

pkg/config/validate.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil
+	}
+	if s.Bundle == nil || s.Bundle.Input == nil || s.Bundle.Input.Recipe == "" {
+		return nil
+	}
+	if s.Recipe.Output.Path != s.Bundle.Input.Recipe {
+		return errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("spec.recipe.output.path (%q) and spec.bundle.input.recipe (%q) must match when both are set",
+				s.Recipe.Output.Path, s.Bundle.Input.Recipe))
+	}
```
lockwobr, line 83: Strict string equality on these paths is brittle and produces false rejections. Consider:

```yaml
spec:
  recipe:
    output:
      path: ./recipe.yaml
  bundle:
    input:
      recipe: recipe.yaml      # same file, fails this check
```

These two values point at the same file but `s.Recipe.Output.Path != s.Bundle.Input.Recipe` is true. Same problem with trailing slashes on parent dirs, `foo/../recipe.yaml`, mixed separators, etc.

Minimum fix: `filepath.Clean` both sides before comparing. Better fix: `filepath.Abs` (relative to the config file's directory, not the process CWD — the CWD at validate time can differ from where the config was authored). Easiest: just `filepath.Clean(s.Recipe.Output.Path) == filepath.Clean(s.Bundle.Input.Recipe)` and document that absolute paths or matching relative forms are required for the cross-check.

Worth a test case: `{path: "./recipe.yaml", recipe: "recipe.yaml"}` should be accepted (or, if you choose to reject it, the error message should explain the equivalence requirement).

pkg/cli/bundle.go
```diff
@@ -166,55 +180,44 @@ func parseBundleCmdOptions(cmd *cli.Command) (*bundleCmdOptions, error) {
 		opts.targetRevision = opts.ociRef.Tag
 	}
 
-	// Parse value overrides from --set flags
-	opts.valueOverrides, err = config.ParseValueOverrides(cmd.StringSlice("set"))
+	// Parse value overrides from --set flags (config replaced by CLI when set)
+	opts.valueOverrides, err = config.ParseValueOverrides(stringSliceFlagOrConfig(cmd, "set", bundleDeploymentSet(bs)))
 	if err != nil {
 		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid --set flag", err)
 	}
```
lockwobr, line 187: Error attribution: when the offending value comes from `spec.bundle.deployment.set` in `--config` (not from a `--set` flag), this still says `"invalid --set flag"`, which sends users to the wrong place. Same on line 191 for `--dynamic`. Consider distinguishing the source so the message points at the right input:

```go
rawSet := stringSliceFlagOrConfig(cmd, "set", bundleDeploymentSet(bs))
opts.valueOverrides, err = config.ParseValueOverrides(rawSet)
if err != nil {
    src := "--set"
    if !cmd.IsSet("set") {
        src = "spec.bundle.deployment.set in --config"
    }
    return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid "+src, err)
}
```

Alternatively, do the parsing inside the proposed `BundleSpec.Resolve()` and let it produce a fully-attributed error like `spec.bundle.deployment.set[2]: invalid value override` at the source — which is one of the wins of the unified `Resolve()` layer.

pkg/cli/root.go
```diff
@@ -357,9 +389,58 @@ func initDataProvider(cmd *cli.Command) error {
...
+func intFlagOrConfig(cmd *cli.Command, flagName string, fallback int) int {
+	if !cmd.IsSet(flagName) {
+		return fallback
+	}
+	v := cmd.Int(flagName)
+	if fallback > 0 && fallback != v {
+		slog.Info("CLI flag overriding config value", "flag", flagName, "config", fallback, "override", v)
+	}
+	return v
+}
```
lockwobr, line 449: Asymmetry vs. `stringFlagOrConfig`: this only logs the override when `fallback > 0`, so a config value of `0` (or a negative — possible if the schema allows it) is silently overridden without an INFO line. That's inconsistent with the string version's `fallback != ""` guard, which does log on a CLI-supplied empty string vs. non-empty config (impossible for `intFlagOrConfig` only because nothing in the schema currently allows negatives or treats 0 as meaningful — but `Nodes` is the obvious candidate to grow that ambiguity).

Consider `if fallback != v` as the symmetric guard, matching the spirit of `stringFlagOrConfig`. Minor — the current behavior is harmless for today's call sites — but the asymmetry is the kind of thing that bites later when a new int-typed config field allows zero as a meaningful value.

pkg/cli/recipe.go
```diff
@@ -178,6 +178,136 @@ Override snapshot-detected criteria:
...
 }
 
+// configRecipeSnapshot returns the snapshot path from spec.recipe.input.snapshot
+// (or empty string when cfg or any intermediate path is nil).
+func configRecipeSnapshot(cfg *appcfg.AICRConfig) string {
+	if cfg == nil || cfg.Spec.Recipe == nil || cfg.Spec.Recipe.Input == nil {
+		return ""
+	}
+	return cfg.Spec.Recipe.Input.Snapshot
+}
```
lockwobr: This accessor escaped the `bundle_config.go` cluster but follows the exact same pattern. If you take the option-2 refactor (move accessors onto spec types as nil-receiver methods), this becomes:

```go
func (r *RecipeSpec) Snapshot() string {
    if r == nil || r.Input == nil {
        return ""
    }
    return r.Input.Snapshot
}
```

and `recipeOutputPath` / `parseRecipeOutputFormat` (lines 192–222) collapse similarly into `RecipeSpec.OutputPath()` / `RecipeSpec.OutputFormat()`. Worth doing as part of the same sweep so the pattern is uniform across both spec types.

pkg/config/loader.go
```diff
@@ -0,0 +1,196 @@
...
+// and the deployer string parses against pkg/bundler/config.
+func Load(ctx context.Context, source string) (*AICRConfig, error) {
+	if source == "" {
+		return nil, errors.New(errors.ErrCodeInvalidRequest, "config source is empty")
+	}
+	if strings.HasPrefix(source, configMapURIScheme) {
+		return nil, errors.New(errors.ErrCodeInvalidRequest,
+			"ConfigMap (cm://) sources are not supported by --config; "+
+				"export the ConfigMap data with `kubectl get cm <name> -o yaml` and pass the resulting file")
+	}
```
lockwobr, line 63: Nice rejection with concrete kubectl guidance — but consider also rejecting `file://` URIs explicitly. Today they fall through to `readFile` which then fails with a confusing `"config file not found: \"file:///abs/path\""` because `os.ReadFile` doesn't strip the scheme. Either strip and accept, or reject with a message like "file:// URIs are not supported; pass the path directly". Tiny nit, but the cm:// rejection sets a high bar for error UX in this code path.

lockwobr (changes requested):

pkg/cli/bundle.go
mchmarny: Strong observation. The accessor refactor (your `bundle_config.go:88` / `recipe.go:188` comments) landed in this branch — accessors are now nil-receiver methods on `pkg/config` types using `slices.Clone` / `maps.Clone`, dropping ~150 lines from `pkg/cli`. The full `Resolve()`/`ToDomain()` refactor is bigger surgery touching `Validate()`, `applyCriteriaFromConfig`, and `parseBundleCmdOptions` together — filed as a follow-up: #783. Keeping it out of this PR so the merge is bounded; the new accessor layer is the natural skeleton for `Resolve()` to land on next.

pkg/config/validate.go
mchmarny: Fixed in 6a8bc0c — `validateRecipeBundleHandoff` now uses `filepath.Clean` on both sides; the example case (`./recipe.yaml` vs `recipe.yaml`) is accepted, as is `dir//file` vs `dir/file`. Mismatched-path rejection wording also updated to "must reference the same file" so the cleaned-comparison contract is implicit. New test cases in `TestValidate_RecipeBundleHandoff`. Mixing absolute vs. relative still fails — agree with your read that proper resolution would need the config file's directory as a base, but I'd rather defer that until someone hits it in practice.

pkg/cli/bundle.go
mchmarny: Fixed in 6a8bc0c — added a `sourceLabel(cmd, flagName, configPath)` helper that returns the user-facing label of whichever input contributed the value (CLI flag if explicitly set, otherwise the spec path). Applied to `--set`, `--dynamic`, `--system-node-toleration`, `--accelerated-node-toleration`, and `--workload-gate`. So a parse failure on a config-supplied value now reads e.g. `invalid spec.bundle.deployment.set: ...` instead of `invalid --set flag: ...`. Your second suggestion — push parsing into `BundleSpec.Resolve()` for fully-attributed errors at the source — is the cleaner endpoint and is captured in #783.

pkg/cli/root.go
mchmarny: Fixed in 6a8bc0c — `intFlagOrConfig` guard switched from `fallback > 0 && fallback != v` to just `fallback != v`, matching `stringFlagOrConfig`'s symmetric guard. A future int field with meaningful zero / negative values won't be silently overridden.

pkg/config/loader.go
mchmarny: Fixed in 6a8bc0c — `file://` URIs now rejected up-front with `"file:// URIs are not supported by --config; pass the path directly"`. Mirrors the `cm://` rejection style. Test in `TestLoad_RejectsFileURI`.

pkg/cli/recipe.go
```diff
@@ -178,6 +178,119 @@ Override snapshot-detected criteria:
...
+// present and differ; when criteria is empty (no snapshot), config simply
+// populates it. CLI flags subsequently override both via applyCriteriaOverrides,
+// yielding precedence: CLI > config > snapshot.
+//
+// Override events are logged at INFO so the resolved value is auditable.
+func applyCriteriaFromConfig(criteria *recipe.Criteria, cfg *appcfg.AICRConfig) error {
+	if cfg == nil || cfg.Spec.Recipe == nil || cfg.Spec.Recipe.Criteria == nil {
+		return nil
+	}
+	c := cfg.Spec.Recipe.Criteria
```
lockwobr: This site escaped the accessor migration — still uses the raw nil-check chain `cfg.Spec.Recipe.Criteria`. The `CriteriaFields()` method on `*RecipeSpec` was added to `accessors.go` for exactly this caller but isn't referenced anywhere, so it's currently dead code.

Suggested change to close the loop:

```go
func applyCriteriaFromConfig(criteria *recipe.Criteria, cfg *appcfg.AICRConfig) error {
    c := cfg.Recipe().CriteriaFields()
    if c == nil {
        return nil
    }
    // ...
```

(Or, if you'd rather inline the nil-check than keep `CriteriaFields()` around, `r := cfg.Recipe(); if r == nil || r.Criteria == nil { return nil }; c := r.Criteria` — and delete `CriteriaFields()` from `accessors.go` since `c.Spec.Recipe.Criteria` is the only place it would ever be useful.)

pkg/cli/bundle.go
```diff
@@ -166,55 +180,44 @@ func parseBundleCmdOptions(cmd *cli.Command) (*bundleCmdOptions, error) {
...
-		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid --dynamic flag", err)
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid "+sourceLabel(cmd, "dynamic", "spec.bundle.deployment.dynamic"), err)
 	}
 
-	// Parse node selectors
-	opts.systemNodeSelector, err = snapshotter.ParseNodeSelectors(cmd.StringSlice("system-node-selector"))
-	if err != nil {
-		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid --system-node-selector", err)
+	if opts.systemNodeSelector, err = resolveNodeSelector(cmd, "system-node-selector", bs.SystemNodeSelector()); err != nil {
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid system node selector", err)
```
lockwobr: These three wraps (lines 195, 198, 216) lost the `--` prefix in the rewrite — was previously `"invalid --workload-selector"` / `"invalid --system-node-selector"` / `"invalid --accelerated-node-selector"`. `resolveNodeSelector` only ever errors on CLI input (the fallback path is `maps.Clone` with no parse), so `sourceLabel` would unambiguously return the flag form. Suggest:

```go
if opts.systemNodeSelector, err = resolveNodeSelector(cmd, "system-node-selector", bs.SystemNodeSelector()); err != nil {
    return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
        "invalid "+sourceLabel(cmd, "system-node-selector", "spec.bundle.scheduling.systemNodeSelector"), err)
}
```

Minor — error users were getting `"invalid --system-node-selector"` before this PR, the rewrite drops the marker. Worth restoring for consistency with the toleration / set / dynamic / workload-gate wraps that *do* use `sourceLabel`.

dims (dismissed):
LGTM

pkg/cli/recipe.go
mchmarny: Fixed in 6933f60 — `applyCriteriaFromConfig` now uses `cfg.Recipe().CriteriaFields()`, closing the loop. Good catch on the dead-code call-out.

pkg/cli/bundle.go
mchmarny: Fixed in 6933f60 — restored the `--flag` marker on all three selector wraps via `sourceLabel`. Errors now read e.g. `invalid --system-node-selector` (CLI) or `invalid spec.bundle.scheduling.systemNodeSelector` (config) instead of the markerless form.

lockwobr (approved):

Issue #754, opened 2026-05-05
Define evidence bundle format and emit via aicr validate --emit-evidence


mchmarny:
Parent: #750
Depends on: cluster fingerprint child issue (filed alongside), #739 (BOM)

## Summary

Define a versioned, signed evidence bundle format and add `aicr validate --emit-evidence <out>` that produces it. The bundle is what a community contributor attaches to a recipe PR; `aicr verify-evidence` (separate child issue) consumes it.

## Format — in-toto attestation with custom AICR predicate

The bundle is a single tar.gz archive whose payload is an [in-toto Statement](https://github.com/in-toto/attestation) carrying a custom predicate. The Statement is signed via **cosign keyless OIDC** (matching `aicr bundle --attest`); KMS support is deferred.

Predicate type: `https://aicr.nvidia.com/recipe-evidence/v1`

Predicate fields (illustrative):

```json
{
  "_type": "https://in-toto.io/Statement/v1",
  "predicateType": "https://aicr.nvidia.com/recipe-evidence/v1",
  "subject": [{
    "name": "recipe:<recipe-name>",
    "digest": { "sha256": "<canonical-recipe-yaml-hash>" }
  }],
  "predicate": {
    "schemaVersion": "1.0.0",
    "aicrVersion": "<v>",
    "validatorCatalogVersion": "<v>",
    "validatorImages": { "<name>": "<digest>" },
    "recipe": "<canonical YAML>",
    "snapshot": "<snapshot YAML>",
    "fingerprint": { ... },
    "fingerprintMatch": { "matched": true, "diff": [] },
    "phases": {
      "deployment": { "status": "passed", "ctrf": {...}, "logs": [...] },
      "performance": { "status": "passed", "ctrf": {...}, "logs": [...] },
      "conformance": { "status": "passed", "ctrf": {...}, "logs": [...] }
    },
    "bom": { "ref": "bom.cdx.json", "digest": "sha256:..." },
    "manifest": { "files": [...], "hashes": {...} }
  }
}
```

The tarball layout:

```
evidence-bundle.tar.gz
├── attestation.intoto.jsonl    # signed in-toto Statement
├── recipe.yaml                  # subject material
├── snapshot.yaml                # subject material
├── bom.cdx.json                 # CycloneDX from #739
├── phases/
│   ├── deployment/{ctrf.json, logs/*}
│   ├── performance/{ctrf.json, logs/*}
│   └── conformance/{ctrf.json, logs/*}
[...]

Issue #753, opened 2026-05-05
aicr verify-evidence command for reviewers and CI


mchmarny:
Parent: #750
Depends on: evidence bundle format child issue (filed alongside)

## Summary

New top-level CLI command `aicr verify-evidence <bundle>` that maintainers and CI use to verify a recipe-evidence bundle produced by `aicr validate --emit-evidence`.

## Why a separate command

Different audience from `aicr validate`:
- `aicr validate` is run by recipe authors / cluster operators against live infrastructure.
- `aicr verify-evidence` is run by **reviewers** and **CI** against an offline artifact, with no cluster required.

Keeping the verification surface isolated also makes the verification logic auditable and minimal — the maintainer's trust assumption is "this command does exactly the cryptographic + schema checks I need."

## Verification steps

1. **Signature.** Verify the in-toto attestation via `cosign verify-blob-attestation` with the expected OIDC issuer/identity. Record the signer identity.
2. **Schema.** Validate the predicate against the published JSON Schema (`recipe-evidence-v1`). Reject unknown predicate types or unsupported schema versions.
3. **Inventory.** Verify every file referenced in the manifest exists in the bundle and its content hash matches.
4. **Recipe ↔ snapshot binding.** Run the fingerprint matcher (from the snapshot issue) — confirm the snapshot's fingerprint satisfies every dimension of the recipe's `criteria`. Reject if it doesn't, with a clear diff.
5. **Inline constraints.** Replay the recipe's inline constraints against the snapshot (today's `aicr validate --no-cluster` path). These are deterministic and should still pass on the reviewer's machine.
6. **Phase results.** Read the CTRF outputs and surface per-check status. Don't re-run validators; that defeats the purpose.
7. **BOM cross-reference.** Confirm the bundled CycloneDX BOM matches the recipe's expected component set.
8. **Output.** Emit a structured JSON report **and** a Markdown summary suitable for embedding in a PR review comment.

## Markdown summary shape

```
## Evidence verification — <recipe-name>

**Signer:** alice@example.com (via GitHub OIDC, github-actions identity)
**AICR:** v0.13.0  •  **Schema:** recipe-evidence-v1.0.0  •  **Bundle SHA:** sha256:abc…

### Cluster fingerprint
✅ service=eks, accelerator=h100, os=ubuntu, k8s=1.33.4 — matches recipe criteria

### Phase results
✅ deployment   12/12 checks passed
[...]

Issue #752, opened 2026-05-05
Capture cluster fingerprint in snapshot for recipe-criteria binding


mchmarny:
Parent: #750

## Summary

Make `aicr snapshot` capture a structured **cluster fingerprint** — the detected service, accelerator, OS, Kubernetes version, and any other dimensions a recipe declares in its `criteria` block. The fingerprint is what evidence bundles use to prove "the cluster the recipe was tested on actually matched the recipe's declared criteria."

## Why

Without a deterministic fingerprint in the snapshot, an evidence bundle can claim a recipe was tested but cannot prove it ran on hardware/services matching the recipe's `criteria`. A reviewer reading the bundle has no way to confirm the H100 / EKS / Ubuntu / Kubeflow recipe was actually validated on H100 / EKS / Ubuntu / Kubeflow versus, say, a kind cluster with mocked GPUs.

This is the first link in the trust chain for #750.

## Scope — what the fingerprint must include

At minimum:

- **Service** — detected provider (eks / aks / gke / oci / coreweave / kind / on-prem)
- **Accelerator** — detected GPU SKU (h100 / gb200 / b200 / mi300 / none)
- **OS** — node OS distribution and version (ubuntu, rhcos, bottlerocket, etc.)
- **Kubernetes** — server version (e.g., 1.33.4)
- **Node count and topology** — total GPU nodes, per-zone breakdown if relevant
- **Detection provenance** — which signals each value came from (node label, instance metadata, kubelet, etc.) so a reviewer can audit the inference

Any field a recipe can put in `criteria.*` should be derivable from the snapshot fingerprint. Extending `criteria` later means extending the fingerprint in lockstep.

## Proposed approach

1. Audit `pkg/snapshotter` and `pkg/collector` for what's already captured. Most of these signals likely exist; the gap is structuring them as a top-level `fingerprint` block in the snapshot YAML rather than scattered through collector outputs.
2. Define a `Fingerprint` Go type in a portable location (likely `pkg/snapshotter` or a new `pkg/fingerprint`).
3. Add a `Match(recipe)` helper that compares a fingerprint against a recipe's `criteria` and returns a structured diff (matched / mismatched / unknown per dimension). This is the function `aicr verify-evidence` will call.
4. Surface the fingerprint as a top-level field in snapshot YAML for human review.
5. Add table-driven tests covering each detection path and the Match comparison.

## Success criteria

- `aicr snapshot` emits a `fingerprint:` block at a stable schema location.
[...]

Issue #751, opened 2026-05-05
Recipe contribution workflow with CI evidence gate


mchmarny:
Parent: #750
Depends on: `aicr verify-evidence` command (filed alongside)

## Summary

Document the end-to-end community recipe contribution workflow and add a CI gate that runs `aicr verify-evidence` on every PR touching `recipes/overlays/` (or other recipe-defining paths). The gate posts a Markdown verification summary as a structured review comment so a maintainer can approve a recipe for hardware they don't have.

## Workflow (contributor view)

1. Author a new overlay or modify an existing one.
2. Apply it to a real cluster matching the recipe's `criteria`.
3. Run `aicr snapshot` on that cluster.
4. Run `aicr validate --recipe <r> --snapshot <s> --emit-evidence ./evidence/`.
5. Attach the resulting `evidence-bundle.tar.gz` to the PR (uploaded artifact, release asset, or large-file attachment — exact mechanism TBD per repo policy).
6. Open the PR. CI verifies the bundle and posts a summary. Maintainers review the bundle alongside the diff.

## Workflow (maintainer / CI view)

1. CI detects a recipe-touching PR.
2. CI fetches the bundle (path TBD — likely `evidence/<recipe>.tar.gz` checked into the PR, or a separate signed-artifact link).
3. CI runs `aicr verify-evidence <bundle>` with expected issuer/identity constraints set to public OIDC providers AICR trusts (initially GitHub Actions OIDC).
4. CI posts the Markdown summary as a review comment. A failing bundle blocks merge via a required check.

## Deliverables

- **`.github/PULL_REQUEST_TEMPLATE/recipe.md`** — recipe-specific PR template requiring evidence bundle attachment, contributor declaration of test environment, and link to relevant `criteria` documentation.
- **`.github/workflows/recipe-evidence.yaml`** — workflow that runs on `pull_request` events touching `recipes/overlays/**` (and other recipe paths). Runs `aicr verify-evidence`, posts a sticky review comment, sets a required check status.
- **`CONTRIBUTING.md` update** — new section: "Contributing a recipe for hardware AICR maintainers can't access" — walks through the contributor workflow with a worked example and an FAQ (what if I don't have an OIDC identity, what if my test failed but I think the failure is environmental, etc.).
- **`docs/integrator/recipe-development.md` update** — link from the existing recipe authoring docs to the contribution workflow page.

## Acceptance criteria for the gate

- A PR adding a recipe **without** an evidence bundle fails the required check.
[...]

PR #823, opened 2026-05-08
docs(design): add ADR-007 verifiable recipe test evidence

njhensley:
## Summary

Adds ADR-007: Verifiable Recipe Test Evidence — a V1 design for a
signed, OCI-distributed evidence bundle so recipe contributions for
hardware AICR maintainers cannot reach can be reviewed against a
cryptographically verifiable artifact instead of trust.

## Motivation / Context

AICR ships recipes spanning AWS/OCI/GCP/Azure/CoreWeave/Forge with H100,
GB200, B200, MI300, and multiple OSes — no single team has hands-on
access to all of them. Today, contributions for unreachable hardware
either stall or merge on faith: the maintainer has no artifact tying
the recipe-as-merged to a real `aicr validate` run on real hardware,
and no durable lineage for re-cert or audit. The recipe already
self-defines its tests via `validation`, `componentRefs`, `criteria`,
and the `aicr validate` pipeline; this ADR packages the answer to a
question the recipe already poses.

The ADR explicitly defers tier policy, multi-instance pointers, signed
layered logs predicates, material-slice canonicalization, KMS signing,
and re-cert automation to follow-on work pulled in by demand.

Fixes: N/A
Related: #750, #751, #752, #753, #754, #739, #745, ADR-006

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

ADR only — no code, schema, or CI changes in this PR. Key decisions
captured for follow-on PRs:

- **V1 surface = five PRs.** Bundle format + pointer +
  `aicr validate --emit-evidence`; `aicr verify-evidence` CLI; CI gate
  on `recipes/**`; required `maintainers:` block on
  `RecipeMetadataSpec`.
- **OCI-native transport with in-tree pointer.** Bundle bytes live at
  `ghcr.io/<owner>/aicr-evidence:<digest>`; pointer file at
  `recipes/evidence/<recipe>.yaml` (schema 1.0) binds the repo to the
  bundle by content hash. Pointer is a side effect of `--emit-evidence`,
  not a separate command.
- **Single bundle class in V1.** Verifier records the OIDC identity;
  maintainer review classifies. Tier A/B/C/D policy file deferred.
- **CycloneDX BOM ships in V1** (hard dep on #739) so the bundle survives
  Renovate digest rotations without re-cert.
- **Logs are optional and unsigned-as-a-whole.** Per-file sha256 in the
  summary's `manifest.json` pre-commits any log file that would be in
  the logs bundle, making tampering detectable without a separate
  signature event. Signed-logs predicate variants deferred.
- **Any recipe edit triggers re-cert in V1.** Material-slice
  canonicalizer deferred until a real attested-recipe corpus exists.
- **`verify-evidence` accepts four input forms** (pointer, OCI ref,
  tarball, unpacked dir) auto-detected — same verification logic for
  all four; covers CI, canonical-artifact, air-gapped, and contributor
  self-debug workflows.
- **Verification is offline.** No validator re-runs — that would defeat
  the purpose and require hardware the maintainer doesn't have.

## Testing

```bash
make qualify
```

Doc-only change; no code paths touched. `make qualify` covers the
docs sidebar check and lychee anchor-link verification on the
Fern-rendered docs site for any PR touching `docs/**`.

## Risk Assessment

- [x] **Low** — ADR is a design document under `docs/design/`; no
  binding effect until follow-on PRs land. Easy to revise or supersede.

**Rollout notes:** None. Implementation rolls out as five separately
scoped follow-on PRs tracked under #750 and its children (#751–#754).
Schema 1.0 of the pointer file is forward-compatible with the deferred
multi-instance role schema (1.0 → 2.0 bump when roles arrive).

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — N/A, doc-only
- [x] I added/updated tests for new functionality — N/A, doc-only
- [x] I updated docs if user-facing behavior changed — this PR *is* the doc
- [x] Changes follow existing patterns in the codebase — uses the
  established `docs/design/NNN-<slug>.md` ADR layout

### DIFF
--- a/docs/design/007-recipe-evidence.md
+++ b/docs/design/007-recipe-evidence.md
@@ -79,7 +79,7 @@ Two design tensions drive what's V1 vs deferred:
 | Tension | Full design choice | V1 choice |
 |---|---|---|
 | Identity model | Tier A/B/C/D with signed policy file, freshness-bounded | One bundle class; verifier records the OIDC identity, maintainer review classifies |
-| Re-cert on non-material edits | Material-slice canonicalizer (RFC 8785-derived, append-only versioned) suppresses re-cert on cosmetic changes | Any recipe edit triggers re-cert; the cost is acceptable until a real attested-recipe corpus exists |
+| Re-cert on non-material edits | Material-slice canonicalizer (RFC 8785-derived, append-only versioned) suppresses re-cert on cosmetic changes | Material-slice canonicalizer ships in V1; the CI gate rolls out soft-fail first (warning-only), then hardens to required after a concrete corpus trigger fires |
 | Multi-cluster attestation per recipe | Multi-instance pointer schema with primary / supplementary / negative roles | Pointer carries a list with one entry; second cluster attests via an additive PR. Schema bump (1.0 → 2.0) when roles arrive |
 | Logs handling | Three signed layered predicate types (logs / redaction / augmentation) attached to the same OCI digest | Optional unsigned logs bundle as separate OCI artifact; per-file content hashes pre-committed in summary's manifest binds them to the signature |
 
@@ -183,14 +183,99 @@ contribution volume or partner relationships pull it in.
    self-debug without a packaging step. Same verification logic runs
    against any of the four.
 
-3. **CI gate workflow + PR template.** Required check on PRs touching
-   `recipes/**`. Reads pointer, runs verify, posts Markdown comment.
-   PR without a pointer file for a touched recipe fails the gate.
+3. **CI gate workflow + PR template.** Check on PRs touching
+   `recipes/**` whose material slice (see "Material-slice
+   canonicalization" below) differs from the slice the current pointer
+   was attested against. Reads pointer, runs verify, posts Markdown
+   comment. **Two-phase rollout**:
+
+   - **Phase 1 — soft-fail (informational).** The check runs on every
+     qualifying PR but emits warning-only annotations; merge is not
+     blocked. Phase 1 starts when PR-C lands and stays in effect until
+     the corpus trigger fires (≥ 3 distinct accelerator classes have
+     at least one attested recipe with a valid pointer, OR 12 weeks
+     after PR-A merges, whichever comes first).
+   - **Phase 2 — hard-fail (required).** The check becomes a required
+     status. Maintainers can apply an `evidence/exempt` label to bypass
+     the gate for an individual PR; bypass requires a justification
+     line in the PR description and is recorded for audit. Pure
+     non-material edits (comments, whitespace, `displayName`,
+     `description`, key-order) never reach Phase 2 because the
+     canonicalizer collapses them to the same material-slice digest.
+
+   Promotion from Phase 1 to Phase 2 is a one-line config change in
+   the workflow, paired with a CONTRIBUTING update documenting the
+   `evidence/exempt` policy.
 
 4. **`maintainers:` block on `RecipeMetadataSpec`.** Required field on
    every recipe in `recipes/overlays/`, listing GitHub handle, org, and
    a durable escalation contact (DL or shared mailbox). One-time backfill
    PR populates existing recipes via `git log` heuristics.
+
+### Material-slice canonicalization (proposed)
+
+The material slice is the projection of the recipe-resolution surface
+that affects runtime behavior. Two recipe states with the same material
+slice produce the same deployed cluster; differences outside the slice
+(comments, formatting, descriptive metadata) cannot. The bundle's
+subject digest is computed over the canonical form of this slice, not
+over the raw post-resolution YAML — so non-material edits do not
+invalidate the bundle.
+
+**What the slice includes** (any change re-certs):
+
+- Leaf overlay's `criteria`, `componentRefs`, `constraints`,
+  `validation`, `nodeScheduling` paths.
+- Each `componentRef`'s resolved `{type, source, version, valuesFile,
+  manifestFiles, overrides}`, recursively into included
+  `valuesFile`/`manifestFiles` content.
+- Mixin contents pulled in by `spec.mixins`, scoped to their
+  `constraints` and `componentRefs` (mixins carry no other material
+  fields by construction — see ADR-005).
+- Registry entries (`recipes/registry.yaml`) referenced by the
+  resolution chain: `helm.{defaultRepository, defaultChart}`,
+  `kustomize.{defaultSource, defaultPath, defaultTag}`,
+  `valueOverrideKeys`, `nodeScheduling`.
+- `spec.maintainers` (PR-D) — a maintainer change is durable
+  signal-routing metadata, not cosmetic.
+
+**What the slice excludes** (changes pass without re-cert):
+
+- All YAML comments, leading/trailing whitespace, blank lines,
+  key-order, quoting style.
+- `metadata.{displayName, description, labels, annotations}` and
+  any registry `displayName`.
+- Top-level `# yaml-language-server: $schema=...` directives and
+  similar editor hints.
+
+**Algorithm.** RFC 8785 (JSON Canonicalization Scheme) over the
+filtered subtree, after a type-preserving YAML→JSON load and NFC
+Unicode normalization on string scalars. The slice is built by
+walking the resolution chain (overlay → mixins → registry →
+component values → manifest files transitively) and emitting only
+the included paths above. Output is the SHA-256 of the JCS bytes;
+this is the value bound by the predicate's subject digest.
+
+**`materialSliceVersion`.** The predicate carries an integer
+`materialSliceVersion` (V1 = `1`). Bundles attest under the algorithm
+in effect at sign-time; the verifier loads the matching algorithm by
+version. New versions append; old versions never get rewritten. This
+isolates canonicalizer bug fixes (and future slice-set changes) from
+historical bundles — a v2 algorithm does not invalidate v1
+attestations, and the verifier carries both parsers.
+
+**Verifier check.** The subject-digest step (verifier step 6) recomputes
+the slice from the repo at HEAD using the bundle's
+`materialSliceVersion` and confirms match against the predicate. A
+mismatch is a hard fail; a match means the recipe is materially
+unchanged since attest-time, regardless of cosmetic edits.
+
+The CI gate (V1 surface item 3) leans on the same canonicalizer:
+"PR touches `recipes/**`" is the broad trigger, but the gate only
+*fails* when the material-slice digest computed from the PR head
+differs from the slice the current pointer was attested against.
+Comment-only PRs, Renovate-driven `displayName` rewrites, and
+formatter changes pass without new evidence.
 
 ### Bundle anatomy (proposed)
 
@@ -233,6 +318,7 @@ arrive when demand justifies — see "Future direction."
 ```yaml
 # https://aicr.nvidia.com/recipe-evidence/v1
 schemaVersion: 1.0.0
+materialSliceVersion: 1
 attestedAt: 2026-05-08T10:23:11Z
 aicrVersion: v0.13.0
 validatorCatalogVersion: v2.4.0
@@ -270,12 +356,20 @@ manifest:
   fileCount: 9
 ```
 
-`subject.digest` is `sha256(canonicalize(post-resolution recipe YAML))`
-where canonicalize is "sort map keys, strip comments, normalize line
-endings, UTF-8 encode" — a small helper, not the RFC 8785-derived
-canonicalizer in the full design. Any recipe edit triggers re-cert;
-the canonicalizer that would suppress re-cert on non-material edits
-is deferred (see "Future direction").
+`subject.digest` is `sha256(JCS(material-slice(post-resolution recipe)))`
+under the algorithm identified by the predicate's
+`materialSliceVersion` (see "Material-slice canonicalization" above).
+Non-material edits — comments, whitespace, `displayName`,
+`description`, key-order — produce the same digest as the original
+attestation and pass the verifier's subject-digest check without new
+evidence. Material edits (chart version, criteria, constraints,
+override values, manifest content) produce a different digest and
+require a fresh attestation.
+
+The predicate also carries the `materialSliceVersion` integer so the
+verifier loads the matching algorithm. V1 ships version `1`; future
+canonicalizer revisions append (`materialSliceVersion: 2`, …)
+without invalidating bundles signed under the previous version.
 
 The `manifest.digest` field binds the manifest to the signature, which
 in turn binds every supporting file (snapshot, BOM, CTRF) by the
@@ -349,10 +443,13 @@ auto-detected input form — OCI ref, tarball, unpacked directory):
 5. **Inventory check.** Verify every file in `manifest.json` exists
    in the bundle; recompute SHA-256 per file; confirm match. Confirm
    `manifest.digest` matches predicate.
-6. **Subject digest check.** Recompute
-   `sha256(canonicalize(post-resolution recipe in repo at HEAD))`;
+6. **Subject digest check.** Load the algorithm identified by the
+   predicate's `materialSliceVersion`; recompute
+   `sha256(JCS(material-slice(post-resolution recipe in repo at HEAD)))`;
    confirm match against the in-toto Statement's `subject.digest`.
-   Any recipe drift since attest-time is a hard fail.
+   Material drift since attest-time is a hard fail. Non-material
+   edits (comments, formatting, `displayName`, `description`,
+   key-order) produce the same digest and pass.
 7. **Per-dimension fingerprint match.** Run `Fingerprint.Match(recipe.criteria)`
    from #752; confirm `criteriaMatch.matched: true`; render per-dimension
    diff in Markdown so reviewers see exactly which dimensions matched.
@@ -396,10 +493,14 @@ message, satisfying #751 acceptance criterion 1.
 
 Three V1 choices preserve future evolution at near-zero cost:
 
-1. **Predicate type is `recipe-evidence/v1`.** When the material
-   classifier or other breaking-change V2 work ships, becomes a clean
-   `/v2` — no migration shim, no schema-version negotiation. Verifiers
-   carry both parsers (append-only).
+1. **Predicate type is `recipe-evidence/v1`; `materialSliceVersion`
+   handles canonicalizer revs without a predicate bump.** A breaking
+   change to the material-slice algorithm ships as
+   `materialSliceVersion: 2` under the same `/v1` predicate type;
+   verifiers carry both algorithm parsers (append-only). The predicate
+   type bumps to `/v2` only when other parts of the predicate body
+   (signer envelope, phase shape, fingerprint surface) change in a way
+   the algorithm version cannot express.
 2. **`pointer.attestations` is a list from day one.** Multi-instance
    in schema 2.0 is additive: more entries, plus a `role:` field
    defaulting to `primary`. No structural break for V1 pointers.
@@ -415,7 +516,6 @@ Three V1 choices preserve future evolution at near-zero cost:
 | Deferred | Pulled by |
 |---|---|
 | Tier A/B/C/D identity policy file | First partner relationship requests a non-community trust label, OR community contribution volume creates review fatigue that tier filtering would relieve. |
-| Material-slice canonicalization (RFC 8785-derived, append-only versioned) | Renovate-driven re-cert flood becomes a real complaint (≥ 5 attested recipes + weekly chart-version bumps). Until then, V1's "any edit triggers re-cert" is honest and cheap. |
 | Multi-instance pointer (schema 2.0) with primary / supplementary / negative roles | Two contributors attest the same recipe from different clusters, OR a "this didn't work for me" negative attestation needs to coexist with a passing primary. |
 | Signed layered predicate types (logs / redaction / augmentation) | Contributor asks to publish redacted logs, OR third party wants to add an independent re-run with its own signer. V1's manifest-pre-commit binding handles "publish logs later" already. |
 | Re-cert age cutoffs (24mo hard, 23mo bot) | First bundle ages past 12 months. Document the policy in CONTRIBUTING; defer the bot. |
@@ -474,10 +574,25 @@ see `## Future direction`.
   judgment surface as today's PR-only review, with a richer artifact
   attached. The Tier policy file (deferred) is what eventually closes
   this gap; per-signal provenance alone does not.
-- **Any recipe edit triggers re-cert.** Without the material-slice
-  canonicalizer, a comment-only edit invalidates the existing
-  bundle. Until the project has multiple attested recipes under
-  active Renovate maintenance, this is a non-issue.
+- **Material-slice changes trigger re-cert; the slice definition is
+  itself a maintenance surface.** Edits to the leaf overlay's
+  `criteria`/`componentRefs`/`constraints`/`validation`/`nodeScheduling`,
+  to `componentRefs`-resolved values/manifest files, to mixin
+  `constraints` or `componentRefs`, or to registry entries pulled
+  through the resolution chain re-cert every recipe whose slice
+  touches them. Comment-only and metadata-only edits pass without
+  re-cert. The slice-set (what's in vs. out of the material projection)
+  is part of the algorithm version: tightening or relaxing it requires
+  bumping `materialSliceVersion`. Initial slice-set is conservative
+  (errs toward "include" for safety); empirical pull-trigger to relax
+  is contributors flagging false positives.
+- **Soft-fail window can hide real evidence drift.** During Phase 1,
+  a PR that legitimately needs new evidence merges with a warning the
+  reviewer may overlook. This is the cost of not blocking on hardware
+  availability before the corpus exists; the trigger conditions
+  (≥ 3 accelerator classes attested, OR 12 weeks) are calibrated to
+  end Phase 1 before drift accumulates. If the trigger drags, the
+  corpus is too thin to enforce a gate against in the first place.
 - **No tier label distinguishes first-party from community evidence.**
   A bundle signed by NVIDIA's CI looks the same to the verifier as
   one signed by an unfamiliar fork. Maintainers eyeball the cosign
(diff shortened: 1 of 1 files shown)


docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+  surprise rewrites; expansion is additive.
+
+### Negative
+
+- **No defense against forged collector inputs.** A contributor
+  controlling their own cluster can lie to the snapshot collectors.
+  V1 records the resolved fingerprint values; per-signal provenance
+  (which collector signal contributed, with what confidence) is
+  deferred. Maintainer review compensates — same posture as today's
+  PR-only review.
```
mchmarny: The framing here understates the impact on the ADR's central claim. The Problem section opens with "no artifact the maintainer can verify came from a real run on real hardware" — but if a contributor's collector output can be forged, the bundle does not, in fact, prove "real run on real hardware." It proves "someone with this OIDC identity signed this snapshot+result pair."

That's still real value, but it shifts the trust anchor from "the evidence" to "the signer's identity claims" (corporate cert, trusted issuer, employer-bound subject). I'd make that explicit somewhere in **Decision** or **Consequences/Positive** — e.g. "the trust handoff in V1 is signer-identity-bound: maintainers verify *who* attested, not *what* the cluster physically was." Without that, the ADR sounds like it solves the trust problem when it actually relocates it to cosign cert review, which is the same maintainer-judgment surface we have today, just with a richer artifact.

This also tightens the case for the deferred Tier-policy work: tier classification is what eventually closes this gap, not per-signal fingerprint provenance alone.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+   Detection is by URL prefix, file extension, and directory
+   existence. The four forms map to distinct workflows: pointer for
+   CI, OCI for canonical artifact verification, tarball for
+   air-gapped or email transport, unpacked directory for contributor
+   self-debug without a packaging step. Same verification logic runs
+   against any of the four.
+
+3. **CI gate workflow + PR template.** Required check on PRs touching
+   `recipes/**`. Reads pointer, runs verify, posts Markdown comment.
+   PR without a pointer file for a touched recipe fails the gate.
```
mchmarny: **Required CI gate + "any recipe edit triggers re-cert" is a sharp combination for V1.** A typo fix, comment, displayName change, or Renovate-driven chart-version bump in `recipes/overlays/` will block on the contributor (or maintainer) having access to the matching hardware to regenerate evidence. That bar is higher than today's status quo for the recipe authors AICR maintainers *can* reach — i.e., the gate may stall maintainer-driven hygiene PRs, not just unreachable-hardware contributions.

A few options worth considering for V1:
- A maintainer-applied label (e.g. `evidence/skip` or `evidence/inherit-prior`) that bypasses the gate for non-material edits, with the trade-off documented.
- Soft-fail (warning) for the first N months / first M attested recipes, hard-fail thereafter.
- Pull material-slice canonicalization forward into V1 in a minimal form (even just "ignore comment-only diffs") instead of accepting full re-cert thrash.

Right now the negative-consequence bullet on this is honest, but pairing "any edit invalidates" with "gate is required" risks making the system feel like it punishes contribution rather than enabling it — exactly the opposite of the trust-handoff motivation.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+  digest: sha256:...
+  fileCount: 9
+```
+
+`subject.digest` is `sha256(canonicalize(post-resolution recipe YAML))`
+where canonicalize is "sort map keys, strip comments, normalize line
+endings, UTF-8 encode" — a small helper, not the RFC 8785-derived
+canonicalizer in the full design. Any recipe edit triggers re-cert;
+the canonicalizer that would suppress re-cert on non-material edits
+is deferred (see "Future direction").
```
mchmarny: Hashing the **post-resolution** recipe couples the bundle to resolver determinism across the entire overlay → mixin → registry → component-values chain. That has two implications worth surfacing:

1. ADR-005 is mid-flight. Any change in resolver semantics (overlay merge order, mixin composition, registry hydration) invalidates every prior bundle even when the leaf overlay is byte-identical. That's a tight coupling between two ADR surfaces.
2. A registry edit, a mixin tweak, or a `recipes/components/<name>/values.yaml` change re-certs *every* recipe whose resolution chain touches them. The CI-gate scope ("PRs touching `recipes/**`") is already that broad, but the user-visible blast radius isn't called out.

Worth either (a) noting this dependency explicitly under **Consequences/Negative** alongside "any recipe edit triggers re-cert," or (b) considering whether the subject digest should bind the *unresolved* leaf overlay plus a manifest of resolution-chain inputs (each with its own hash). The latter at least lets the verifier tell "leaf changed" from "shared mixin changed" without re-running validation.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
+# ADR-007: Verifiable Recipe Test Evidence
+
+> **Status:** Proposed (design-only; not implemented). This ADR specifies
+> the V1 contract. Implementation lands as five follow-on PRs tracked
+> under [#750](https://github.com/NVIDIA/aicr/issues/750) and its children
+> ([#751](https://github.com/NVIDIA/aicr/issues/751)–[#754](https://github.com/NVIDIA/aicr/issues/754)).
+> Bundle formats, CLI flags, schema fields, and verifier behavior
+> described below are future intent, not current behavior.
```
mchmarny: **ADR layout drifts from the established template in this repo.** The other ADRs (002–006) all carry these top-level sections:

- `## Status` (own heading, with date — not a quote-block at the top)
- `## Context`, `## Decision`, `## Consequences`, plus
- `## Alternatives Considered` — present in 002, 003, 004
- `## References` — present in 002, 003, 004

This ADR has a status quote block, no `## Alternatives Considered`, and no `## References`. The substance is here, but two gaps stand out:

1. **No alternatives section.** The decision jumps straight to "in-toto + cosign keyless + OCI artifact + in-tree pointer + CycloneDX BOM" without weighing the obvious alternates: PR-attached tarball vs OCI, sigstore bundle format vs in-toto Statement, single bundle vs split summary/logs, KMS vs keyless, git-LFS vs OCI for transport. Some are touched in **Non-Goals** ("KMS deferred"), but the design space isn't documented. Future maintainers reopening this question will not know which alternates were considered and rejected.
2. **No references.** Worth pointing to in-toto attestation framework, DSSE, cosign keyless / Fulcio / Rekor, CycloneDX 1.5, OCI artifact spec, and related issues (#739, #745, #750, ADR-006).

Aligning the structure makes this ADR easier to cross-reference and review against its peers.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+(`accelerator: { value: h100, signals: [kubelet, dcgm, imds, dra],
+confidence: high }`) plus a "what the fingerprint cannot prove"
+disclosure surfaced in the verifier output. Multi-signal corroboration
+makes forged-collector attacks harder to mount unnoticed.
+
+**Pull-trigger:** first time a community contribution gets pushback
+for "is this cluster real?" Until then, the resolved values plus
+maintainer review are sufficient. **Compatibility:** the predicate's
+`fingerprint.<dim>` field grows from `{value}` to
+`{value, signals[], confidence}`; verifiers that parse the V1 shape
```
mchmarny: **Future direction is doing too much work in this ADR.** This section is ~165 lines (lines 465–629) sketching V2/V3 designs in significant detail — RFC 8785-derived canonicalizer internals, tier-policy file layout, schema 2.0 fields, four predicate types, OSV advisory feed, mirror bot, per-signal provenance.

The deferred-features table at line 386 already names each item with its pull-trigger. That's the right level for an ADR. The long-form sketches read like proto-design docs and create implicit commitment to designs that may not survive the demand event that pulls them in.

Suggest:
- Keep the deferred-features table as-is (it's tight and useful).
- Move each long-form sketch into its tracking issue under #750, where it can evolve when its trigger fires.
- Replace the `## Future direction` section with a 1-paragraph pointer to those issues.

Net effect: ADR drops from 659 → ~450 lines, lands closer to the size of ADR-005 (the largest existing ADR at 371 lines), and stops pseudo-committing to V2 design choices.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+   `./out/pointer.yaml` to `recipes/evidence/<recipe>.yaml` and commit.
+
+2. **`aicr verify-evidence` CLI.** Full verification: signature, schema,
+   inventory (manifest hashes), recipe ↔ snapshot fingerprint match
+   (per-dimension diff), inline constraint replay, phase results, BOM
+   cross-reference. Markdown + JSON output. Single positional argument
+   accepts any of four input forms:
+
+   ```bash
+   aicr verify-evidence <input>
```
mchmarny: **CLI surface convention.** Existing AICR commands group under nouns: `aicr snapshot`, `aicr recipe`, `aicr validate`, `aicr bundle`, `aicr query`. Both proposed surfaces here break that:

- `aicr validate --emit-evidence ./out --push <reg>` overloads `validate` with output-side concerns (signing, OCI push, pointer write).
- `aicr verify-evidence` introduces a new top-level verb.

Consider `aicr evidence emit` and `aicr evidence verify` as a noun-grouped surface. Benefits:
- Keeps `validate`'s job narrow (the validate stage in the pipeline diagram).
- Leaves room for `aicr evidence list` / `aicr evidence diff` / `aicr evidence push` later without further top-level pollution.
- Matches `aicr query` precedent for a dedicated, read-only inspection command.

Implementation-wise, `evidence emit` can still consume `validate`'s outputs internally — it doesn't have to mean re-running validation, just that the user-facing entry point is grouped with the rest of the verbs that operate on artifacts.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+   PR populates existing recipes via `git log` heuristics.
+
+### Bundle anatomy (proposed)
+
+Summary bundle (always published):
+
+```text
+oci://ghcr.io/<owner>/aicr-evidence:<digest>
+└── (OCI artifact whose layers contain:)
+    ├── attestation.intoto.jsonl    # DSSE-wrapped, cosign keyless signed
```
mchmarny: nit: a single in-toto Statement should be `attestation.intoto.json`. The `.jsonl` extension is for newline-delimited multi-statement streams. Cosign's standard output for a single DSSE-wrapped Statement is `.intoto.json` (or the bundle format `.dsse.json`). Worth fixing here so the bundle layout matches what tooling will actually emit.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+   against any of the four.
+
+3. **CI gate workflow + PR template.** Required check on PRs touching
+   `recipes/**`. Reads pointer, runs verify, posts Markdown comment.
+   PR without a pointer file for a touched recipe fails the gate.
+
+4. **`maintainers:` block on `RecipeMetadataSpec`.** Required field on
+   every recipe in `recipes/overlays/`, listing GitHub handle, org, and
+   a durable escalation contact (DL or shared mailbox). One-time backfill
+   PR populates existing recipes via `git log` heuristics.
```
mchmarny: The Adoption Plan describes PR-D as "fully independent of A/B/C." If it's independent of evidence, why is it in this ADR?

Either:
- Make the evidence dependency explicit (e.g., the `maintainers:` block is the durable contact for re-cert events, advisory revocations, or signer-identity disputes — that's a real connection worth stating), or
- Split it out. A `maintainers:` field on `RecipeMetadataSpec` with a CI presence gate is a self-contained schema change worth its own short ADR or an issue under #751, not a sub-bullet here.

As written, it reads like opportunistic bundling, which makes the V1 surface feel larger and less coherent than it needs to.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+| Re-cert on non-material edits | Material-slice canonicalizer (RFC 8785-derived, append-only versioned) suppresses re-cert on cosmetic changes | Any recipe edit triggers re-cert; the cost is acceptable until a real attested-recipe corpus exists |
+| Multi-cluster attestation per recipe | Multi-instance pointer schema with primary / supplementary / negative roles | Pointer carries a list with one entry; second cluster attests via an additive PR. Schema bump (1.0 → 2.0) when roles arrive |
+| Logs handling | Three signed layered predicate types (logs / redaction / augmentation) attached to the same OCI digest | Optional unsigned logs bundle as separate OCI artifact; per-file content hashes pre-committed in summary's manifest binds them to the signature |
+
+Two surfaces from the broader design **do** ship in V1 because their
+cost-to-defer is high:
+
+- **OCI transport + in-tree pointer file.** Bundle bytes live in OCI
+  (`ghcr.io/<owner>/aicr-evidence:<digest>`); the pointer file at
+  `recipes/evidence/<recipe>.yaml` binds the repo to the bundle by
```
mchmarny, line 91: nit: the bundle URL is shown as `ghcr.io/<owner>/aicr-evidence:<digest>` in three places (here, line 173, line 268). Worth one sentence stating that this is an example only — any OCI-1.1-compliant registry is acceptable (GHCR, GitLab Container Registry, Harbor, ECR, Artifact Registry). Contributors on corporate GitLab Enterprise won't be able to push to ghcr.io and reading this as a hard constraint will confuse them.

mchmarny (commented):
Solid first AICR ADR, njhensley — the problem framing is clear, the V1 contract is concrete enough to implement against, and the deferred-features table with pull-triggers is a strong pattern that I'd like to see used in future ADRs. The choice to bring OCI transport + CycloneDX BOM forward into v1 makes right trade-offs as well. 

A few things to sharpen before this lands:

**Higher-impact:**
- The trust-model framing risks overpromising: forged collector inputs are acknowledged as a gap, but the ADR opens with "real run on real hardware" — v1 actually delivers signer-identity-anchored trust, not cluster-physicality proof. Worth making that explicit.
- The required-CI-gate + any-edit-triggers-re-cert combination needs an escape hatch for v1, or it will block routine maintainer hygiene PRs (e.g. comment edits, Renovate bumps) on hardware availability.
- Hashing the post-resolution recipe ties bundle validity to resolver determinism across overlays/mixins/registry/values — worth surfacing the coupling.

**Structural:**
- ADR layout drifts from the established repo template (002–006): mainly the lack of `## Alternatives Considered` and `## References` sections.
- ~165 lines of "Future direction" detail is doing V2 design work that could be just in tracking issues; dropping the long-form would keep the ADR short and lose and decision making content. 

**Smaller items:**
- CLI surface (`aicr verify-evidence`, `--emit-evidence`) breaks AICR's noun-grouped subcommand convention; consider `aicr evidence emit` / `aicr evidence verify`.
- `attestation.intoto.jsonl` should be `.intoto.json` for a single Statement.
- The `maintainers:` block (PR-D) is described as independent of evidence — either tie it in or split it out.
- The `ghcr.io/<owner>/...` URL pattern reads as a hard constraint; one line clarifying any OCI-1.1 registry is fine would help.

Tooling-wise the design is feasible: `oras-go/v2`, `in-toto/attestation`, `sigstore-go`, and `cyclonedx-go` are already direct dependencies, so PR-A doesn't add a new external dep surface.

Nothing here blocks merging IMO — design-only ADR, easy to revise via follow-up. Comments are aimed at sharpening V1 before the implementation PRs hang off it.

docs/design/007-recipe-evidence.md
```diff
@@ -0,0 +1,659 @@
...
+   #   ./out/pointer.yaml      (ready to copy to recipes/evidence/<recipe>.yaml)
+   ```
+
+   With an optional `--push <oci-registry>` that closes the loop:
+
+   ```bash
+   aicr validate --recipe r.yaml --snapshot s.yaml \
+     --emit-evidence ./out \
+     --push ghcr.io/myorg/aicr-evidence \
+     [--push-logs]
```
mchmarny: **The proposed `aicr validate --emit-evidence` surface should plug into the existing `AICRConfig` envelope rather than introducing more standalone flags.**

`pkg/config` (added recently) defines `AICRConfig` (`kind: AICRConfig`, `apiVersion: aicr.nvidia.com/v1alpha1`) as the canonical multi-input capture for `aicr recipe` and `aicr bundle`. Its `Spec.Bundle` already carries `AttestationSpec` (cosign / keyless inputs) and `RegistrySpec` (OCI push target / credentials). A user passes `--config aicr.yaml` instead of a long flag list, and CLI flags override file values.

The evidence-emit invocation in this ADR has exactly that multi-input shape:

```yaml
spec:
  validate:
    input:
      recipe: r.yaml
      snapshot: s.yaml
    evidence:
      emit: ./out
      includeLogs: true
    registry:               # reuse pkg/config.RegistrySpec
      push: ghcr.io/myorg/aicr-evidence
      pushLogs: true
    attestation:            # reuse pkg/config.AttestationSpec (cosign keyless, KMS later)
      mode: keyless
      issuer: ...
```

Two concrete asks for the ADR:

1. Add a `ValidateSpec` to the AICRConfig schema (sibling of `RecipeSpec` / `BundleSpec`), with `Input`, `Evidence`, `Registry`, `Attestation` sub-sections — explicitly **reusing** `pkg/config.RegistrySpec` and `pkg/config.AttestationSpec` so the OCI-push and signer-identity surfaces stay consistent across `bundle` and `validate`.
2. State that `aicr validate --config aicr.yaml --emit-evidence` is the supported invocation; the current ADR examples (long flag chains) become the expanded form, not the canonical one.

This also gives PR-A a clean schema-bump landing pattern (AICRConfig is `v1alpha1`, so adding a `Validate` section is additive) and keeps the contributor workflow consistent: same config file flows through `aicr recipe` → `aicr validate --emit-evidence` → `aicr bundle`.

mchmarny (commented):
Follow-up on the CLI surface: the recently added `AICRConfig` envelope in `pkg/config` already captures multi-input shapes for `aicr recipe` and `aicr bundle` (including `AttestationSpec` and `RegistrySpec` under `BundleSpec`). The evidence-emit invocation should extend this same envelope with a `ValidateSpec` rather than introduce a new flag-only surface — see inline.

njhensley:
Thanks for the thorough read, Mark. Took all three higher-impact items as substantive design changes (trust model, CI-gate sharpness, resolver coupling) and the structural / smaller items as inline fixes. The `--emit-evidence` vs existing `--evidence-dir` flag-name overlap surfaced during self-review and is also resolved.

Pushed 8 commits stacked on the existing two — each addresses one thread (or related set) and is individually revertable if any of the design calls turn out wrong. Replies on each thread below.

docs/design/007-recipe-evidence.md
njhensley: Agreed — V1 is signer-identity-bound, not cluster-physicality-bound, and the original opening overstated the artifact. New `### Trust model` subsection at the top of `## Decision` makes the relocation explicit ("the maintainer's trust judgment moves from 'did this PR really run?' to 'do I trust this signer's claim that it did?'"), with a parallel sharpened bullet in `## Consequences/Negative`. Problem-section opener tightened from "real run on real hardware" → "cryptographically tie back to the contributor who ran it." Tier policy noted as the eventual closer. (Commit 68814ebd.)

docs/design/007-recipe-evidence.md
njhensley: Took all three options you floated, in combination. Pulled material-slice canonicalization into V1 (RFC 8785 / JCS over a filtered subtree, with `materialSliceVersion` as the algorithm-revision integer), so non-material edits — comments, whitespace, `displayName`/`description`, key-order — produce the same digest and pass without re-cert. Layered a soft-fail rollout on top: Phase 1 emits warning-only annotations until ≥ 3 distinct accelerator classes are attested OR 12 weeks after PR-A merges (whichever first); Phase 2 hardens to required with `evidence/exempt` as the maintainer-applied bypass for individual PRs (audit-logged, requires justification line). Specified at `### Material-slice canonicalization` and the rewritten V1 surface item 3. Soft-fail-window risk also surfaced as its own Negative consequence. (Commit 802c7d69.)

docs/design/007-recipe-evidence.md
njhensley: Took option (b). The in-toto Statement's `subject[0].digest.sha256` is now the **material-slice digest** (re-cert trigger; what the signature binds to), and the predicate body carries a new `chainManifest` field with the unresolved leaf overlay plus an ordered list of per-input hashes (`{kind: mixin|registry|values|manifest, path, sha256}`) for forensic provenance. Verifier step 6 split into 6a (material-digest recompute, hard-fail on mismatch) and 6b (chain-manifest walk, informational in V1 — surfaces "input X changed since attest-time" in the Markdown summary, doesn't itself fail verification; V2 may promote). Cost is near-zero since the resolver already walks every input. Post-resolution-only digest moved into Alternatives Considered as the rejected design with rationale citing exactly the ADR-005 mid-flight coupling and "which input changed" diagnostic problem you raised. (Commit cbab5e5c.)

docs/design/007-recipe-evidence.md
njhensley: Realigned. Status promoted from quote-block to `## Status` heading with date. New `## Alternatives Considered` covering the design space you named — PR-attached tarball vs. OCI, sigstore-bundle vs. in-toto Statement, single combined vs. split summary/logs, KMS vs. cosign keyless, git-LFS vs. OCI, single-attestation vs. list-from-day-one — plus two extras that came up during the rework (CLI verb shape, post-resolution-only digest). New `## References` with standards (in-toto, DSSE, Sigstore, Fulcio, Rekor, CycloneDX 1.5, OCI 1.1, ORAS, RFC 8785, CTRF), peer ADRs (002, 005, 006), and tracking issues (#739, #745, #750–#754). One template variation: `## Problem` and `## Non-Goals` sit between Status and Context, matching ADR-005/006 convention. (Commit 5ac7de2e.)

docs/design/007-recipe-evidence.md
njhensley: Collapsed. Section is now ~13 lines pointing back to the deferred-features table under `## Decision` as the canonical V2 placeholder list. Long-form V2 sketches (canonicalizer internals — those moved into `## Decision` per the line-161 thread; tier-policy file layout; schema 2.0 fields; four predicate types; OSV advisory feed; mirror bot; per-signal provenance) all removed. Per-row design happens at the demand event when a pull-trigger fires, not pre-emptively. Dangling cross-references throughout the doc (Non-Goals, Consequences/Positive, the deferred-features paragraph, Trust model) retargeted to the table. The savings here funded the new Alternatives Considered, Trust model, Material-slice canonicalization, Evidence taxonomy, AICRConfig integration, and chainManifest sections. (Commit 6273eb32.)

docs/design/007-recipe-evidence.md
njhensley: Took the noun-grouping for consumption: `aicr evidence verify` (full verification, four input forms, auto-detect), `aicr evidence list`, `aicr evidence show`. Production stays at the origin (`aicr validate --emit-attestation`) — evidence is a side-effect of the pipeline that produced its inputs, and re-emitting via `aicr evidence emit` would either re-run validate (defeating offline verification) or quietly couple `evidence` to validate's stored output. Added an `### Evidence taxonomy` subsection framing the family with multiple kinds: `cncf-conformance` (already shipped in `pkg/evidence`) and `recipe-test-attestation` (this ADR). Future kinds (CycloneDX BOM-only, SLSA provenance, license attestation) plug into `aicr evidence verify` as additional dispatches. Rejected `aicr verify-evidence` standalone-verb shape documented in Alternatives Considered with explicit rationale.

Also: the `--emit-evidence` flag name itself overlapped with the existing CNCF surface's `--evidence-dir` / `--cncf-submission` / `--feature` (all `Category: "Evidence"` in `pkg/cli/validate.go`). Renamed to `--emit-attestation` to match the taxonomy kind, with a side-by-side flag-layout table in the Evidence taxonomy section. (Commits 7d195534, 8f6a272b.)

docs/design/007-recipe-evidence.md
njhensley: Fixed. (Commit 0382cd1d.)

docs/design/007-recipe-evidence.md
njhensley: Kept under ADR-007 with the dependency made explicit. New "Why this lives under ADR-007 rather than as a separate schema PR" subsection in V1 surface item 4 names the three triggers: re-cert prompts when a material-slice digest expires, advisory-revocation notifications when a deployed image is flagged post-merge (deferred — see the advisory-feed row in the deferred-features table), and signer-identity disputes. PR-D remains *schedule*-independent (the field is additive metadata and can land first if convenient), but its *motivation* is the evidence lifecycle this ADR establishes. Adoption-plan PR-D bullet rewritten to make this distinction. (Commit 0382cd1d.)

docs/design/007-recipe-evidence.md
njhensley: Added at first mention: **"GHCR is shown as the example throughout this ADR; any OCI-1.1-compliant registry is acceptable"** with the alternatives enumerated (GHCR, GitLab Container Registry, Harbor, JFrog Artifactory, AWS ECR, Google Artifact Registry, Azure Container Registry) and a note that the verifier reads the registry from the pointer file via standard ORAS paths. The remaining `ghcr.io/<owner>/...` references downstream now read as illustrative under that disclaimer. (Commit 0382cd1d.)

docs/design/007-recipe-evidence.md
njhensley: Took it. PR-A now extends `pkg/config.AICRConfig` (still `v1alpha1`, additive) with a `ValidateSpec` sibling of `RecipeSpec`/`BundleSpec`. Cosign-identity (`AttestationSpec`) and OCI transport (`RegistrySpec`) are reused as-is — they nest under `spec.validate.evidence.attestation` because only the attestation kind signs and pushes (CNCF doesn't). Sibling structure: `spec.validate.evidence.cncf` and `spec.validate.evidence.attestation` so a single config file can drive either kind or both without flag-name ambiguity. `aicr validate --config aicr.yaml` is now stated as the supported invocation; the flag form is the expanded equivalent and overrides config values, keeping the contributor flow `aicr recipe` → `aicr validate --emit-attestation` → `aicr bundle` consistent through one envelope. (Commits 7d195534, 8f6a272b.)

mchmarny (approved):

PR #582, opened 2026-04-15
feat(cli): add aicr diff for configuration drift detection

sanjeevrg89:
Adds `aicr diff` — compares a snapshot against a recipe to surface constraint violations and component version drift.

**What's included:**
- Constraint evaluation: flags mismatches between expected recipe values and actual snapshot state
- Component drift: detects version skew between recipe chart versions and deployed images
- Prometheus metrics for drift tracking (counters, gauges, histograms)
- Table output for human-readable reports
- Full test coverage (190 tests, race-clean)

Scoped to just the diff feature per @xdu31's feedback on #464 — node-validate and node-scoped constraints will follow as separate PRs.

Issue #464, opened 2026-03-25
RFC: Dynamic GPU Orchestration Layer — From Day-Zero Recipes to Day-N Runtime Intelligence


sanjeevrg89:
## Summary

AICR solves the day-zero problem brilliantly: reproducible, version-locked recipes that eliminate the tribal knowledge of GPU cluster configuration. But production GPU infrastructure is a **dynamic problem** — traffic is bursty, nodes get preempted, hardware is heterogeneous, and GPUs sitting idle burn money at $2-4/hr per device.

This RFC proposes a **Dynamic GPU Orchestration Layer** that extends AICR from point-in-time validation to continuous runtime intelligence. The goal: make AICR the single control plane for GPU infrastructure — from initial deployment through steady-state operations.

---

## Upstream Context & Prior Art

This proposal builds on and complements existing upstream work:

| Upstream Issue | Relationship |
|---|---|
| [NVIDIA/aicr#448](https://github.com/NVIDIA/aicr/issues/448) — Inference perf validation with AIPerf | Our **Runtime Observability** (Section 5) extends this from point-in-time inference validation to continuous performance baselining. AIPerf benchmarks become the initial data source for ongoing regression detection. |
| [NVIDIA/aicr#442](https://github.com/NVIDIA/aicr/issues/442) — AKS H100 NCCL All-Reduce performance runtime | Our **Continuous Drift Detection** (Section 1) depends on complete platform runtime coverage. AKS NCCL runtime templates are a prerequisite for drift detection on AKS clusters. |
| [NVIDIA/aicr#443](https://github.com/NVIDIA/aicr/issues/443) — Validator Job scheduling alignment | Our **Workload Placement** (Section 3) operates at a different layer (workload scheduling vs. validator scheduling), but both share the principle that scheduling config should be recipe-driven. |
| [ROADMAP.md P2](https://github.com/NVIDIA/aicr/blob/main/ROADMAP.md) — Configuration Drift Detection | This RFC provides a concrete design for the drift detection item currently listed as a backlog bullet point with no issue or design. |

**Overlap assessment:** No existing issue or active work covers the dynamic runtime layer proposed here. The four areas below (drift detection, autoscaling, workload placement, fleet federation) are net-new to AICR.

---

## Motivation

Today's AICR workflow is:

```
Recipe → Bundle → Deploy → Validate (point-in-time) ✅
```

What's missing is the **feedback loop**:

```
[...]

PR #824, opened 2026-05-09
feat(cli): add aicr diff for configuration drift detection

sanjeevrg89:
## Summary

Adds `aicr diff` — compares two snapshots field-by-field to surface configuration drift between cluster states.

## Motivation / Context

Supersedes #582. That branch's history was a tangle of merge commits from main that made rebase impractical, so per @lockwobr's [review feedback](https://github.com/NVIDIA/aicr/pull/582#issuecomment-4407820557) this is a fresh branch off current main with a single signed squash commit.

No functional change relative to #582 HEAD apart from adapting to current main's helper-flag and `initDataProvider` signatures, plus addressing all open CodeRabbit findings preemptively.

Fixes: N/A
Related: #582 (superseded), #464 (original scope-reduction request)

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] API server
- [ ] Recipe engine / data
- [ ] Bundlers
- [ ] Validator
- [ ] Core libraries
- [x] Docs/examples (`docs/`)
- [ ] Other

## Implementation Notes

**What's included:**
- Field-level comparison across all measurement types and subtypes
- Reports added, removed, and modified readings with deterministic ordering (sorted by path, reproducible across runs)
- Distinct nil-reading sentinel (`<nil>`) so nil↔`""` drift is never silently masked
- Table, JSON, and YAML output formats
- ConfigMap URI input support (`cm://namespace/name`) for both baseline and target — local file ↔ live cluster mixes work
- `--fail-on-drift` flag returns `ErrCodeConflict` (not `ErrCodeInvalidRequest` — drift is a state mismatch, not a malformed request)
- Race-clean unit + integration test coverage of the CLI write path, including Close-error promotion and ConfigMap-output rejection through the real `writeDiffResult`

**Scope:** snapshot-vs-snapshot only. Recipe-vs-snapshot diff and node-scoped constraints will land as separate PRs (per #464).

**Open CodeRabbit findings from #582 — all addressed in this PR:**
- `<nil>` sentinel in `safeReadingString` (was returning `""`, masking drift)
- nil-`Measurement` and nil-`Reading` guards
- Index-based subtype access in `addedMeasurement` / `removedMeasurement` (no per-iteration struct copies)
- `TestDiffCmd_ConfigMapTableOutputGuard` replaced — old test re-implemented `strings.HasPrefix` and tested its own expression; new `TestDiffCmd_TableConfigMapRejected` exercises the cm:// rejection through the actual CLI path
- New `TestWriteDiffResult_TableToFile` and `TestWriteDiffResult_CreateFails` directly cover the named-return Close-error promotion (previously uncovered from CLI tests)
- "fails mid-table" test renamed and parameterized so the case actually exercises the mid-stream path (`successes: 2`)
- `--fail-on-drift` switched from `ErrCodeInvalidRequest` to `ErrCodeConflict` (correct semantic)
- CLI docs added in `docs/user/cli-reference.md` (was missing in #582)

## Testing

```bash
make qualify
```

Result:
- 56 packages pass with `-race` (no flakes)
- Total coverage **75.5%** (threshold 70%)
- pkg/diff: **92.5%**, pkg/cli: 63.3%
- golangci-lint v2.11.3: **0 issues**
- license headers, AGENTS.md sync, docs sidebar, chart-version pins, BOM: ✅
- e2e step skipped locally (chainsaw not on dev machine); CI will exercise it

Coverage delta on touched packages (vs origin/main):
- `pkg/cli`: new file additions only; combined package coverage 63.3% (well above 70% per-package floor in absolute, no decrease since prior baseline)
- `pkg/diff`: new package, **92.5%** at first checkpoint

## Risk Assessment

- [x] **Low** — Isolated, additive change. `pkg/diff` is a new package; the only existing-file edit is a 1-line registration in `pkg/cli/root.go`. Easy to revert by removing the package and the registration line.

**Rollout notes:** N/A — no migrations, no feature flags, no infra changes.

## Checklist

- [x] I updated docs (`docs/user/cli-reference.md`) since user-facing behavior was added
- [x] Changes follow existing patterns in the codebase (functional flag instances, `pkg/errors` codes, structured logging via `slog`, named-return for Close-error promotion)


sanjeevrg89:
@coderabbitai addressed in 26522063:

**1. `pkg/diff/table.go` — errWriter early-return:** Fixed. Added an `if ew.err != nil { return wrapTableErr(ew.err) }` check immediately after the `CHANGES` header `printf` so a failed write short-circuits before the tabwriter is constructed. (Didn't switch the tabwriter to write through `errWriter` — that would require adding a `Write` method to `errWriter` and is a bigger surface change for the same observable behavior.)

**2. `docs/user/cli-reference.md` — markdownlint MD031 on the synopsis fence:** Skipping. The existing convention in this file (every other command — `snapshot`, `recipe`, `query`, `validate`, `bundle`, `verify`, `trust update`, `skill`) is `**Synopsis:**` immediately followed by the opening fence with no blank line between. Adding the blank line would make the `diff` section inconsistent with the rest. There's no `.markdownlint*` config in the repo gating this, and CI doesn't currently enforce MD031 on these blocks. If the project wants to flip this convention it should be a separate sweep across all command sections, not a one-off here.

sanjeevrg89:
@coderabbitai addressed in f2a1285e: switched the table-output stdout default from `os.Stdout` to `cmd.Root().Writer` for testability and consistency with `query.go`, `skill.go`, `bundle.go`, `trust.go`, and `root.go`. File-output path unchanged.

sanjeevrg89:
@coderabbitai triaged the latest review (4ec10982):

**Fixed (2):**
1. **`docs/user/cli-reference.md` — `--output` description:** Updated to make the cm:// boundary explicit — `cm://namespace/name` is valid for JSON/YAML, rejected only for `--format table` (structured format required for ConfigMap storage).
2. **`pkg/cli/diff.go:163,182` — Close errors merged with active error:** Both deferred closers now use `stderrors.Join(err, ...)` instead of `if err == nil { err = ... }`. If `diff.WriteTable` / `ser.Serialize` fails AND `Close()`/flush also fails, both errors surface together rather than the close error being silently dropped. Added `stderrors` alias to disambiguate from `pkg/errors`.

**Skipped (2) — repo-wide patterns, out of scope for this feature PR:**
3. **`pkg/cli/diff.go:116` — timeout context not propagated to `FromFileWithKubeconfig`:** Genuine architectural concern, but `FromFileWithKubeconfig`'s signature doesn't accept a context, and **every existing CLI command uses it the same way** (`pkg/cli/query.go:156`, `pkg/cli/validate.go:545`). Fixing it just for `diff` would require either a parallel `FromFileWithKubeconfigContext` API or refactoring every caller. That belongs in a dedicated serializer-context PR, not in a feature add.
4. **`pkg/cli/diff.go:182` — `NewFileWriterOrStdout` ignores kubeconfig:** Same situation. The read/write asymmetry exists in **every existing CLI command** that does kubeconfig-aware reads (`query`, `snapshot`, `recipe`, `validate` all call plain `NewFileWriterOrStdout` for output). Adding a kubeconfig-aware writer just for `diff` would be inconsistent with the rest of the codebase. Worth a dedicated PR adding `NewFileWriterOrStdoutWithKubeconfig` and updating every command in lockstep.

For visibility: the previous PR #582 review also flagged the timeout-vs-I/O issue (comment 3120616856) and the same conclusion was reached.

sanjeevrg89:
@coderabbitai all 4 addressed in 9c2ad711:

**1. `pkg/cli/diff_test.go` — `--fail-on-drift` CLI test:** Added `TestDiffCmd_FailOnDriftReturnsConflict`. Builds two snapshots that differ on `K8s.server.version`, asserts the CLI returns a non-nil error containing `drift detected` when `--fail-on-drift` is set, and asserts no error on the same inputs without the flag. Plus a small `writeSnapshotWithVersion` helper using the canonical `subtype:`/`data:` YAML shape (verified against `pkg/snapshotter/doc.go`).

**2. `pkg/diff/diff_test.go` — explicit `Str("")` regression:** Added `TestSnapshots_ExplicitEmptyStringPreserved`. Compares `Str("1.32.4")` baseline vs `Str("")` target and asserts `*c.Target == ""` (not nil, not absent). This is the load-bearing test for the schema fix below.

**3. `pkg/diff/diff.go` — `Baseline`/`Target` are now `*string`:** Done. Schema clarified — `nil` = side genuinely absent (Added has no baseline; Removed has no target), `&""` = explicitly-empty reading, `&"X"` = a value. With `omitempty` on the JSON/YAML tag, `nil` is the only thing that gets dropped on the wire; an explicit empty reading round-trips correctly. Added a small `strPtr` helper at the construction sites and updated `renderCell` in `table.go` to render `-` for nil, `\"\"` for an explicit empty reading, and the literal value otherwise. All existing tests updated to use `strPtr` and to handle `*string` in field assertions.

**4. `pkg/diff/table.go` — nil `result` guard:** Added an explicit nil check at the top of `WriteTable` returning `ErrCodeInvalidRequest` ("WriteTable requires a non-nil Result"). Better than panicking on `len(result.Changes)`, and unambiguous so callers don't get a misleading "NO CHANGES" output for a programming error. Test added.

Local: `go test -race ./pkg/cli/... ./pkg/diff/...` passes; `golangci-lint run -c .golangci.yaml` reports 0 issues. `pkg/diff` coverage stays high (the new tests bring it up further).

I'll take you up on the offer to open tracking issues for the two skipped `*WithKubeconfig`/`*WithContext` items from the previous round — those legitimately need a coordinated cross-CLI sweep that doesn't belong in this feature PR. Please go ahead and file them.

docs/user/cli-reference.md
```diff
@@ -834,6 +834,68 @@ Results are output in CTRF (Common Test Report Format) — an industry-standard
 
 ---
 
+### aicr diff
```
mchmarny, line 839: **Blocker — accidental docs regression.** Two unrelated hunks in this file silently delete documentation that exists on current `main`:

1. The `--requests` and `--limits` rows from the `aicr snapshot` flag table (around old-line 95–96, hunk `@@ -92,8 +92,6 @@`).
2. The `AICR_REQUESTS` and `AICR_LIMITS` env-var rows (around old-line 1843–1844, hunk `@@ -1840,8 +1900,6 @@`).

Neither relates to the diff feature.

Root cause: the branch is `mergeable_state: behind` — it forked off `2f22e2b8` and main has moved past it. The cli-reference table on this branch predates the `--requests`/`--limits` addition that landed on main after the fork.

Fix: rebase onto current `origin/main`. The conflict in `docs/user/cli-reference.md` will surface; resolve in favor of main's content (keep the `--requests`/`--limits` rows and matching env-var entries) and re-insert the new `aicr diff` section at the same place.

Worth eyeballing the rest of the diff post-rebase for similar drift in other docs/sidebar files (the docs sidebar check in `make qualify` would catch some of this, but only after CI actually runs — see top-level review note).

pkg/diff/diff.go
```diff
@@ -0,0 +1,288 @@
...
+	Modified ChangeKind = "modified"
+)
+
+// Severity classifies the impact of a detected change.
+type Severity string
+
+const (
+	// SeverityInfo indicates an informational change.
+	SeverityInfo Severity = "info"
+)
```
mchmarny, line 45: YAGNI: the `Severity` type has exactly one value (`SeverityInfo`) and every Change in the codebase is constructed with it (lines 217, 221, 228, 254, 262). It tells consumers nothing today, but it's a public schema field on `Change` (`json:"severity" yaml:"severity"` — no `omitempty`), so it's hard to remove later without breaking JSON consumers.

Either:
- Drop `Severity` from `Change` and the `Severity` type entirely until a real classifier exists, or
- Drive at least one alternate value from the comparison logic (e.g., `SeverityWarning` for measurement-type-level adds/removes vs. field-level modifies) so the field carries information.

Adding the field "for future use" pushes the cost onto every downstream consumer of the schema; landing it without classification means schema noise that's easier to add than to remove.

pkg/cli/diff.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+	defer func() {
+		if closer, ok := ser.(interface{ Close() error }); ok {
+			if closeErr := closer.Close(); closeErr != nil {
+				err = stderrors.Join(err, errors.Wrap(errors.ErrCodeInternal, "failed to close serializer", closeErr))
+			}
+		}
+	}()
+
+	return ser.Serialize(ctx, result)
```
mchmarny, line 183: **Coverage gap on the default format.** `--format yaml` is the default per `formatFlag()`, so this branch is the path most invocations exercise — but no test directly validates its output. The existing tests cover:

- `TestWriteDiffResult_TableToFile` — table to file
- `TestWriteDiffResult_CreateFails` — table create-failure
- `TestDiffCmd_TableConfigMapRejected` — table cm:// rejection
- `TestDiffCmd_FailOnDriftReturnsConflict` — runs `--format json` end-to-end but only asserts the error string, never reads the produced output file

Suggest one test that drives `writeDiffResult` with `serializer.FormatYAML` (and one with `FormatJSON`) to a tempdir file, then unmarshals the file and asserts the `Result` round-trips with the expected `Summary` counts. That also incidentally covers the `closer.Close()` named-return path on the JSON/YAML branch (lines 175–181), which today is dead from a coverage perspective.

pkg/cli/diff.go
```diff
@@ -0,0 +1,184 @@
...
+
+	baseline, err := serializer.FromFileWithKubeconfig[snapshotter.Snapshot](baselinePath, kubeconfig)
+	if err != nil {
+		return err
+	}
+
+	target, err := serializer.FromFileWithKubeconfig[snapshotter.Snapshot](targetPath, kubeconfig)
+	if err != nil {
+		return err
+	}
```
mchmarny, line 118: nit: bare `return err` here loses the call-site context. `serializer.FromFileWithKubeconfig` already returns a structured error with a code (so the AGENTS.md "don't double-wrap" rule is satisfied), but the sibling `pkg/cli/validate.go` wraps the same call:

```go
snap, err = serializer.FromFileWithKubeconfig[snapshotter.Snapshot](snapshotFilePath, kubeconfig)
if err != nil {
    return errors.Wrap(errors.ErrCodeInternal, fmt.Sprintf("failed to load snapshot from %q", snapshotFilePath), err)
}
```

The wrap there carries the path into the message, which is genuinely useful when you're loading two snapshots and need to know which side failed. Worth doing the same here for both baseline and target so the operator-facing error names the file:

```go
return errors.Wrap(errors.ErrCodeInternal, fmt.Sprintf("failed to load baseline snapshot from %q", baselinePath), err)
```

(and similarly for target on line 117).

pkg/cli/diff_test.go
```diff
@@ -0,0 +1,344 @@
...
+	err := app.Run(t.Context(), []string{
+		"aicr", "diff",
+		"--baseline", snap,
+		"--target", snap,
+		"--format", "table",
+	})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+}
```
mchmarny, line 166: nit: this test runs `app.Run` without overriding the writer, so the `cmd.Root().Writer` defaults to `os.Stdout` and the table output (`NO CHANGES\n`) leaks into test runs. Quick fix:

```go
var buf bytes.Buffer
app.Writer = &buf
// ...run...
// optionally: assert buf contains expected output
```

`buildDiffCommandWithOutput` has the same issue. Setting `app.Writer = io.Discard` (or a buffer + assertion) keeps `go test -v` output clean and incidentally lets the test verify the table actually rendered.

pkg/diff/diff.go
```diff
@@ -0,0 +1,288 @@
...
+	return r.Summary.Total > 0
+}
+
+// Snapshots compares two snapshots and returns a structured diff result.
+// The baseline is the reference state; the target is the current state.
+// If either baseline or target is nil, returns an empty Result (no drift).
+func Snapshots(baseline, target *snapshotter.Snapshot) *Result {
+	if baseline == nil || target == nil {
+		return &Result{Changes: make([]Change, 0)}
+	}
```
mchmarny, line 111: nit: silently returning an empty (no-drift) `Result` when either snapshot is nil makes upstream loader bugs invisible. The CLI flow today won't hit this — `serializer.FromFileWithKubeconfig` errors when it can't deserialize — but the contract here trades safety for an operational gotcha: if a future caller gets a `(nil, nil)` from any source, the diff reports "no drift" instead of surfacing the error.

Either return an error (`ErrCodeInvalidRequest`, "baseline snapshot is nil") or document on the function comment that nil inputs are intentionally treated as no-op for the no-snapshot-yet bootstrap case. Right now the doc says "returns an empty Result (no drift)" which describes the behavior but doesn't justify it.

mchmarny (changes requested):
Core feature implementation is solid. The PR description's own thoroughness about CodeRabbit-finding remediation from the prior #582 round is appreciated.

Two issues block merge:

1. **Docs regression.** The branch is `mergeable_state: behind` (forked from `2f22e2b8`, two commits behind current main). `docs/user/cli-reference.md` silently deletes the `--requests` / `--limits` flag rows and the `AICR_REQUESTS` / `AICR_LIMITS` env-var rows from `aicr snapshot` documentation — those exist on `main` and are unrelated to this feature. See inline.
2. **CI hasn't actually run.** Only auto-jobs (Label PRs, Auto-Merge) executed against this head SHA; tests, lint, e2e, security scan, license check, BOM check, AGENTS.md sync — none have run. The PR description's `make qualify` claim is unverified by automation, and the per-package coverage gate the project enforces hasn't been checked. A rebase will trigger CI.

Both root-cause to the same fix: rebase onto current `origin/main`, resolve the `cli-reference.md` conflict in favor of main's content, force-push the rebased branch, then verify CI goes green.

Other findings (medium / low):
- `Severity` is a single-valued enum on a public schema — drop it or actually drive a value from comparison logic before this lands.
- The default `--format yaml` writeDiffResult path has no direct test coverage; existing tests skip it.
- Bare `return err` after snapshot loads loses the path context that sibling `validate.go` includes — minor consistency.
- Two tests leak output to real stdout.
- `Snapshots(nil, nil)` silently returns no-drift instead of erroring — defensive but masks upstream bugs.

sanjeevrg89:
@coderabbitai both addressed in ecf2e252:

**1. `pkg/diff/diff.go` — `HasDrift()` derives from `len(Changes)`:** Now `func (r *Result) HasDrift() bool { if r == nil { return false }; return len(r.Changes) > 0 }`. Caller-constructed Results (where Summary may not be populated) now report correctly, and a nil receiver no longer panics. Added `TestHasDrift_DerivedFromChanges` table-driven test covering: nil receiver → false; empty Changes + zero Summary → false; populated Changes + zero Summary → true (the bug fix); empty Changes + nonzero Summary → false (Changes wins).

**2. `docs/user/cli-reference.md` — qualified the CI-gating note:** Removed the "parse the structured error JSON" advice (default logging is text; structured JSON only with `--log-json`). Replaced with format-stable guidance: write the result via `--output drift.json --format json` and branch on the file's presence and `summary.total` field. That signal works regardless of `--log-json`.

Note on commit shape: this came in as a fix-up commit on top of the merge with main rather than a force-push amend, since @mchmarny had merged main into the branch on the remote and overwriting that with a force-push would be impolite. Squash-on-merge will collapse to a single main commit anyway.

Thanks also for filing #825 and #826 for the architectural follow-ups.

sanjeevrg89:
FYI for triage — the `Preview Fern Docs: Comment` failure ([run 25670442388](https://github.com/NVIDIA/aicr/actions/runs/25670442388)) isn't from this PR. It's the post-build comment workflow running in `main`'s context (head `012900e8`), and it fails with `ENOENT: fern/versions/v0.12.1-content/README.md`. The directory is missing on main — only `v0.12.0.yml` and `v0.12.1.yml` are under `fern/versions/`. Will fail on any docs-touching PR until restored.

mchmarny (approved):

Issue #732, opened 2026-05-04
Extract Validation as standalone portable type from pkg/recipe


xdu31:
## Description

Extract `Validation` (currently `ValidationConfig` in `pkg/recipe/metadata.go`) as a standalone, portable Go type that can be used in multiple contexts: embedded in recipes, standalone YAML files, or embedded in external data structures.

## Context

External consumers want to use AICR's validation types programmatically. To enable this, validation types need to be extracted as portable Go structs that can be embedded or extended in consumer code while remaining usable in AICR's existing recipe system.

## Requirements

1. Extract `ValidationConfig` as a standalone type (rename to `Validation` or keep as-is)
2. Add fields useful for external orchestration:
   - `NodeSelector map[string]string` - for node targeting
   - `ExpectedNewReplicas int` - for batch processing
   - Status tracking fields (phase state, check results)
3. Maintain backward compatibility with existing recipe structure
4. Ensure JSON/YAML tags are present for serialization
5. No external dependencies (keep it portable)

## Acceptance Criteria

- [ ] `Validation` type exists as standalone Go struct
- [ ] Type includes `nodeSelector` and `expectedNewReplicas` fields
- [ ] Type includes status fields for tracking phase/check state
- [ ] Existing recipe loading continues to work unchanged
- [ ] Type can be embedded via `json:",inline"` tag
- [ ] Unit tests validate JSON/YAML marshaling/unmarshaling
- [ ] Documentation updated with usage examples

## Related

- Parent: #731

Issue #731, opened 2026-05-04
[Epic]: Support external validation controller integration


xdu31:
## Overview

Enable external consumers to use AICR as a validation planning engine by exposing public Go package exports and extracting portable validation types.

## Context

External consumers want to use AICR's validation logic programmatically without using the CLI. Use cases include:

1. Embedding AICR types in custom data structures or APIs
2. Getting Job specs from AICR and deploying them with custom orchestration
3. Building higher-level validation workflows using AICR as a library

AICR needs to expose:

1. Portable Go types that can be embedded or extended
2. Planning functions that return Job specs without deploying them
3. Result extraction that is sidecar-safe
4. Utility functions for common operations

## Architecture

- **AICR role**: Provide planning functions (build Job specs from validation configs) and result extraction
- **Consumer role**: Use AICR types, get Job specs, handle deployment and orchestration
- **Separation**: AICR stays focused on validation logic, consumers handle deployment

## Required Changes

This is a tracking issue for the following changes:

### Phase 1: Extract Portable Types (Parallel)
- [x] #732 Extract Validation as standalone portable Go type
- [x] #733 Extract ValidatorCatalog as standalone portable type

### Phase 2: Core Go Package Exports (Depends on Phase 1)
- [x] #734 Add public RenderJob() function returning Job specs (with Plan() for internal use)
- [x] #795 Make ExtractResult sidecar-safe by reading from 'validator' container explicitly
- [x] #736 Export FilterEntriesBySpec() function

### Deferred (Add Only When Second Consumer Demonstrates Need)
- [ ] #735 Add WithCatalog() and WithValidation() builder options (**DEFERRED** - orchestration pattern)

## Implementation Order

1. **Phase 1**: Extract portable types (#732, #733 - parallel)
2. **Phase 2**: Core Go package exports (#734, #795, #736 - parallel, depends on Phase 1)
3. **Deferred**: Helpers - only if second consumer needs them

## Benefits

- External consumers can use AICR as a library
- AICR types become reusable in other projects
- Planning functions enable custom deployment strategies
- Minimal public Go package surface keeps semver manageable
- Prevents consumer divergence from AICR renderer
- Internal testability through Plan() separation

mchmarny:
The epic LGTM. 2 comments: 

[...]

PR #831, opened 2026-05-09
feat(validator): apply hybrid resource pattern to ValidatorCatalog

### LINKED ISSUE #733: Extract ValidatorCatalog as standalone portable type
## Description

Extract the validator catalog structure as a standalone, portable Go type that can be used in multiple contexts: loaded from YAML files, embedded in recipes, or used by external consumers.

## Context

The catalog currently lives in `pkg/validator/catalog/catalog.go` as `ValidatorEntry`. External consumers want to use AICR's catalog system programmatically. We need to formalize this as a portable type.

## Requirements

1. Define `ValidatorCatalog` type with list of `ValidatorEntry`
2. Ensure `ValidatorEntry` has all necessary fields:
   - Name, Phase, Description, Image, Timeout
   - Args, Env, Resources
3. No external dependencies (keep it portable)
4. JSON/YAML serialization support
5. Maintain compatibility with existing catalog loading from `validators/catalog.yaml`

## Acceptance Criteria

- [ ] `ValidatorCatalog` type exists with `Validators []ValidatorEntry` field
- [ ] Type can be loaded from YAML files (existing catalog.yaml format)
- [ ] Type can be embedded in external code via `json:",inline"`
- [ ] Existing catalog loading (`catalog.Load()`) continues to work
- [ ] Unit tests validate JSON/YAML marshaling
- [ ] Documentation updated

## Related

- Parent: #731

xdu31:
## Summary
Moves `ValidatorCatalog` types from `pkg/validator/catalog` to `pkg/api/validator/v1` and updates the API group to `validator.nvidia.com/v1alpha1`, consistent with the refactoring done for `ValidationInput` in #830.

## Motivation/Context
Fixes #733
Related to #732 (ValidationInput refactoring)

Mark's review feedback requested:
1. Move ValidatorCatalog to `pkg/api/validator/v1` (same location as ValidationInput)
2. Change API group from `aicr.nvidia.com/v1` to `validator.nvidia.com/v1alpha1`
3. Add JSON `omitempty` tags (only YAML tags were present)
4. Relax validation for embedded usage (allow empty APIVersion/Kind when Metadata is nil)

## Type of Change
- [x] Refactoring (non-breaking change which improves code organization)

## Components Affected
- `pkg/api/validator/v1` - New location for ValidatorCatalog types
- `pkg/validator/catalog` - Re-exports types for backward compatibility
- `recipes/validators/catalog.yaml` - Updated API version

### New Structure
- Created `pkg/api/validator/v1/catalog.go` with all catalog types:
  - `ValidatorCatalog`
  - `CatalogMetadata`
  - `ValidatorEntry`
  - `ResourceRequirements`
  - `EnvVar`
- Added both `json` and `yaml` `omitempty` tags to all optional fields
- Moved `ForPhase` method to v1 package
- Changed API version constant from `aicr.nvidia.com/v1` to `validator.nvidia.com/v1alpha1`

### Backward Compatibility
- `pkg/validator/catalog` now re-exports types as aliases
- All existing code continues to work without changes
- Validation functions and image resolution remain in catalog package

### Validation Updates
- Updated `validate()` to allow embedded usage:
  - When `Metadata` is nil: APIVersion and Kind are optional (embedded usage)
  - When `Metadata` is present: APIVersion and Kind are required (standalone file)

### Test Updates
- Updated all test YAML strings to use new API version `validator.nvidia.com/v1alpha1`
- Updated test expectations to use `v1.CatalogAPIVersion` and `v1.CatalogKind` constants

## Testing
- ✅ All catalog tests passing (97.6% coverage)
- ✅ All validator tests passing
- ✅ Full test suite: 75.3% coverage (above 70% threshold)
- ✅ golangci-lint: 0 issues
- ✅ make qualify: PASS

## Risk Assessment
**Low Risk**
- Type aliases maintain full backward compatibility
- No functional behavior changes
- Follows same pattern as ValidationInput refactoring (#830)
- All existing tests passing

## Checklist
- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Code commented where needed
- [x] Documentation updated (godoc)
- [x] No new warnings generated
- [x] Tests updated for new API version
- [x] All tests passing locally
- [x] make qualify passed
- [x] License headers ensured


pkg/validator/catalog/catalog.go
```diff
@@ -39,10 +39,33 @@ const (
...
+//	  name: default
+//	  version: 1.0.0
+//	validators: [...]
+//
+// Embedded usage (in a CR):
+//
+//	spec:
+//	  catalog:
+//	    validators: [...]
 type ValidatorCatalog struct {
```
mchmarny: **(a) Package location.** Making the struct embeddable is mechanically fine, but `ValidatorCatalog` lives in `pkg/validator/catalog` alongside the loader, jobspec builder, deployer, CTRF emitter, and RBAC helpers — so any external Go consumer importing the type still pulls the whole orchestration package and its dependencies.

If this is meant to be a portable contract for external consumers (Epic #731), it should sit in a dedicated public-API package — same `pkg/api/validator/v1` (or similar) called out on #830. Keep the orchestration code internal; export only the data types. Otherwise the "embed-friendly" change gives downstream code an embed-ready type buried under a heavy internal surface, and we can't refactor the orchestration package without breaking importers.

This and #830 should align on the same public package — best to decide that before either merges.

pkg/validator/catalog/catalog.go
```diff
@@ -39,10 +39,33 @@ const (
 )
 
 // ValidatorCatalog is the top-level catalog document.
+// Supports both standalone file usage (with full metadata) and embedded usage in CRs (metadata omitted).
+//
+// Standalone usage (catalog.yaml):
+//
+//	apiVersion: aicr.nvidia.com/v1
```
mchmarny: **(b) API group + (d) stability tier.** Same concern as #830:

1. **`aicr.nvidia.com` claims a CRD group AICR doesn't ship.** No CRD, no controller, no schema registered. Stamping `aicr.nvidia.com` on a Go-only type implies an API group AICR commits to maintain. Downstream OSS consumers wrapping this in their own CRDs will use their own groups, and pinning `aicr.nvidia.com` here either forces a rename later or risks collision if AICR ever does ship a real CRD. Either drop the documented `apiVersion`/`kind` defaults or frame them clearly as non-binding documentation.

2. **`v1` is a semver promise this type isn't ready for.** Brand new, no external consumers yet, and the package boundary itself is still being decided (see other comment). `v1alpha1` plus an explicit stability/deprecation note in a package `doc.go` is the honest call. The Epic explicitly mentions keeping the public Go surface semver-manageable — `v1` works against that.

pkg/validator/catalog/catalog.go
```diff
@@ -39,10 +39,33 @@ const (
...
-	Kind       string           `yaml:"kind"`
-	Metadata   CatalogMetadata  `yaml:"metadata"`
+	// APIVersion is the API version (optional, for standalone resource usage).
+	APIVersion string `yaml:"apiVersion,omitempty"`
+
+	// Kind is always "ValidatorCatalog" (optional, for standalone resource usage).
+	Kind string `yaml:"kind,omitempty"`
+
+	// Metadata contains catalog metadata (optional, for standalone resource usage).
+	Metadata *CatalogMetadata `yaml:"metadata,omitempty"`
```
mchmarny: Honest question before locking the hybrid pattern in: is there a concrete external consumer planning to embed `ValidatorCatalog` in their own resource? If the AICR CLI is the only catalog consumer today, the embed affordance has no first user — but the change still adds permanent API surface (the `*CatalogMetadata` pointer-vs-nil distinction, two new tests asserting the empty-marshalling shape, the godoc commitment to support both modes).

For an OSS project, every public-API affordance is something we have to keep supporting. Worth confirming there's a real consumer driving this before we take on the maintenance cost. If the answer is "anticipated, but not yet," I'd lean toward landing only the package-location move (per #830) and deferring the pointer-Metadata change until a consumer needs it.

mchmarny (changes requested):
Mechanically the change is small and clean (`omitempty` tags + pointer `Metadata` + tests), and it mirrors the pattern from #830. Holding for the same shape-level concerns plus one use-case question:

- (a) Public type belongs in a dedicated `pkg/api/validator/v1` package, not co-located with the orchestration code in `pkg/validator/catalog`.
- (b) `apiVersion: aicr.nvidia.com/...` implies a CRD group AICR doesn't ship — drop or explicitly scope as a non-binding example.
- (d) Tag the package `v1alpha1` with stability docs, not `v1`.
- Confirm there's a real external consumer for the embed affordance before locking the pointer-Metadata change into the public API.

This PR and #830 set the public-API package shape together — recommend deciding both at once and re-landing as a single coherent Phase 1 change.

Unrelated nit: PR body has `🤖 Generated with [Claude Code]` — repo policy in CLAUDE.md prohibits that attribution.

mchmarny (approved):

PR #840, opened 2026-05-11
refactor(tests): use host aicr binary in snapshot deploy-agent test

### LINKED ISSUE #727: [Feature]: refactor snapshot tests to exercise go deployer directly
### Prerequisites

### Feature Summary

The existing kind chainsaw test at `tests/chainsaw/snapshot/deploy-agent/`
provisions the snapshot agent by hand-applying `rbac.yaml` and
`snapshot-job.yaml`. Both files are reproductions of what
`pkg/k8s/agent.Deployer.Deploy()` already produces when the customer runs
`aicr snapshot -o cm://...`. The test should call the binary instead

### Problem/Use Case

`tests/chainsaw/snapshot/deploy-agent/snapshot-job.yaml` says explicitly that
it was *"Extracted from `test_snapshot()` in `tests/e2e/run.sh` lines
621-674."* It is a snapshot of what `pkg/k8s/agent/job.go` happens to
generate for systemd-based hosts: hostPID, hostNetwork, `privileged: true`,
hostPath mounts for `/run/systemd` and `/usr/local/bin/nvidia-smi`, the
`AICR_LOG_PREFIX=agent` env var, the `aicr-e2e-snapshot` Job name, and so on.

This duplication has three concrete costs:

1. **Drift.** Any change to `pkg/k8s/agent/job.go` that the chainsaw fixture
   does not also receive will silently pass the chainsaw test while
   misrepresenting customer behavior. The Talos work
   (`pkg/k8s/agent/job.go` skipping `/run/systemd` and `/etc/os-release`
   mounts when `OS=talos`) is the most recent example: the fixture would
   need a parallel hand-edit to stay accurate.
2. **Coverage gap.** The test asserts the *result* (the snapshot ConfigMap)
   but not the *path that produced it*. `pkg/k8s/agent.Deployer` —
   including its RBAC creation, log streaming, Job-completion wait, and
   cleanup — is not exercised end-to-end by this test. Unit tests in
   `pkg/k8s/agent/` cover these in isolation; nothing covers them
   together against a real cluster from this entry point.
3. **Maintenance.** Two files (`rbac.yaml`, `snapshot-job.yaml`) plus the
   `apply` and `delete` steps in `chainsaw-test.yaml` are pure boilerplate
   that does not belong in this repo — `pkg/k8s/agent` is the source of
   truth for that boilerplate.

### Proposed Solution

Replace the apply-RBAC + apply-Job + assert-Job-complete + assert-ConfigMap
steps with a single chainsaw `script` step that invokes the host's `aicr`
binary, mirroring the structure landing in
`tests/chainsaw/snapshot/deploy-agent-talos/`:

```yaml
- name: run-snapshot
  description: Invoke `aicr snapshot` against the kind cluster.
  try:
    - script:
        content: |
          aicr snapshot \
            --kubeconfig "${KUBECONFIG}" \
            -o cm://default/aicr-e2e-snapshot
```

The binary handles RBAC creation, Job deployment with the correct pod
shape for this OS, log streaming, completion wait, and Job/RBAC cleanup.
Only the resulting ConfigMap survives — which is exactly what the
`assert-snapshot-content.yaml` step needs.

After the refactor, the directory shrinks to:

```
tests/chainsaw/snapshot/deploy-agent/
├── chainsaw-test.yaml          # one script step + one assert step
└── assert-snapshot-content.yaml
```

`rbac.yaml`, `snapshot-job.yaml`, `assert-job-complete.yaml`, and
`assert-configmap.yaml` are deleted. The `cleanup` block becomes a
single `kubectl delete cm aicr-e2e-snapshot --ignore-not-found`; the Job
and RBAC are gone before chainsaw's cleanup runs.

## Prerequisites carried over from the kind dev-env

- The host running chainsaw needs the `aicr` binary on `PATH`. Either
  `make build` precedes the chainsaw run (preferred — same path other
  e2e flows use), or the chainsaw target declares the dependency.
- The kind cluster still needs the fake `nvidia-smi` injected via
  `tools/fake-nvidia-smi/` so the GPU collector produces the
  asserted-on `gpu.model: NVIDIA B200` content. That setup is unrelated
  to this refactor and stays.
- `KUBECONFIG` must point at the kind cluster (today the chainsaw test
  inherits this from the developer's environment via `make
  dev-env`/`tools/e2e`; behavior is unchanged).

## Risks

- **Cleanup semantics.** Verify that `aicr snapshot -o cm://...` cleans up
  Job + RBAC by default and leaves the ConfigMap intact. If the flag
  defaults differ, the chainsaw script step gains an explicit
  `--cleanup` (or equivalent) flag — a one-line change, not a design
  change.
- **Log noise.** `aicr snapshot` streams agent logs to stderr. Chainsaw
  captures them, which is a behavior change from the current test (which
  shows only chainsaw-level events). This is an improvement on failure
  diagnostics, not a regression.
- **Existing CI.** `tests/e2e/run.sh` and any GitHub Actions workflow
  that wires this chainsaw test in must be re-checked — the test name
  and entrypoint don't change, but the assumption that `rbac.yaml` and
  `snapshot-job.yaml` exist as copyable templates may be relied on
  elsewhere. Grep before deletion.

## Out of scope
- Any change to `pkg/k8s/agent` itself.
- Any change to fake-nvidia-smi injection.

### Success Criteria

- [ ] `tests/chainsaw/snapshot/deploy-agent/snapshot-job.yaml` deleted.
- [ ] `tests/chainsaw/snapshot/deploy-agent/rbac.yaml` deleted.
- [ ] `tests/chainsaw/snapshot/deploy-agent/assert-job-complete.yaml` deleted.
- [ ] `tests/chainsaw/snapshot/deploy-agent/assert-configmap.yaml` deleted.
- [ ] `chainsaw-test.yaml` invokes `aicr snapshot -o cm://...` from a
      `script` step.
- [ ] No grep hits for the deleted filenames anywhere outside the
      removed files themselves.
- [ ] `make qualify` passes; the existing kind e2e workflow that runs
      this chainsaw test passes against an unchanged kind cluster.

### Alternatives Considered

_No response_

### Component

Multiple components

### Priority

Nice to have

_No response_

Just a testing change should not affect anything released.

Yes, I can open a PR

pdmack:
## Summary

Replace hand-crafted RBAC and Job fixtures in the `snapshot-deploy-agent` chainsaw test with a single `aicr snapshot` CLI invocation, eliminating fixture drift and exercising the real `pkg/k8s/agent.Deployer` end-to-end.

## Motivation / Context

Fixes: #727
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: tests/chainsaw

## Implementation Notes

The existing test applied hand-crafted `rbac.yaml`, `snapshot-job.yaml`, and assertion fixtures that duplicated (and drifted from) what `pkg/k8s/agent.Deployer` actually produces. The old test was already broken against current main — the agent binary now expects to self-deploy its own RBAC and Job, so the hand-crafted Job fails with permission errors.

This refactor follows the pattern established by the Talos snapshot test (`tests/chainsaw/snapshot/deploy-agent-talos/`):

1. Invoke `aicr snapshot --image localhost:5001/aicr:local -o cm://default/aicr-e2e-snapshot` from a `script` step
2. The CLI handles RBAC creation, Job deployment, log streaming, completion wait, and cleanup
3. Assert only the surviving ConfigMap's snapshot content

**Deleted:** `rbac.yaml`, `snapshot-job.yaml`, `assert-job-complete.yaml`, `assert-configmap.yaml`

**Updated:** `assert-snapshot-content.yaml` — matches the actual deployer output structure (includes `NodeTopology` type and `GPU.hardware` subtype that the old fixtures did not cover).

## Testing

```bash
# Local Kind cluster with fake nvidia-smi
chainsaw test --test-dir tests/chainsaw/snapshot/deploy-agent
```

Test passes locally in ~9 seconds. Verified on Kind cluster (`kind-aicr`) with `kindest/node:v1.35.0`, fake nvidia-smi injected, and `localhost:5001/aicr:local` image.

## Risk Assessment

**Rollout notes:** N/A — the old test was already broken against current main.

## Checklist


tests/chainsaw/snapshot/deploy-agent/assert-snapshot-content.yaml
```diff
@@ -31,22 +31,23 @@ measurements:
...
       - subtype: node
+  - type: GPU
+    subtypes:
+      - subtype: hardware
+      - subtype: smi
+  - type: SystemD
+    subtypes:
+      - subtype: containerd.service
+      - subtype: docker.service
+      - subtype: kubelet.service
```
mchmarny: PR description says this "follows the pattern established by the Talos snapshot test" — but the assertion shape is the one the Talos test explicitly rejected. From `deploy-agent-talos/assert-snapshot-content.yaml:19-23`:

> Chainsaw's positional-list matching is too brittle here: collector completion order is non-deterministic (errgroup), and adding a new collector would silently break unrelated tests. We use JMESPath expressions instead so the assertion is order-independent at both the measurement-type level and the per-type subtype level.

The snapshotter runs collectors via errgroup, so the order this file pins (`OS, NodeTopology, K8s, GPU, SystemD`) is whatever the scheduler happened to produce on the author's machine. An infra blip or Go runtime change can reshuffle this and break the test for reasons unrelated to the snapshot. Same risk on subtype lists.

Suggest mirroring the Talos JMESPath form:

```yaml
measurements:
  (length(@)): 5
  (sort([*].type)): [GPU, K8s, NodeTopology, OS, SystemD]
  (@[?type == 'GPU'] | [0].subtypes | sort([*].subtype)): [hardware, smi]
  (@[?type == 'OS'] | [0].subtypes | sort([*].subtype)): [grub, kmod, release, sysctl]
  (@[?type == 'NodeTopology'] | [0].subtypes | sort([*].subtype)): [label, summary, taint]
  (@[?type == 'K8s'] | [0].subtypes | sort([*].subtype)): [image, node, policy, server]
  (@[?type == 'SystemD'] | [0].subtypes | sort([*].subtype)): [containerd.service, docker.service, kubelet.service]
```

tests/chainsaw/snapshot/deploy-agent/assert-snapshot-content.yaml
```diff
@@ -31,22 +31,23 @@ measurements:
...
   - type: K8s
     subtypes:
       - subtype: server
       - subtype: image
       - subtype: policy
       - subtype: node
+  - type: GPU
+    subtypes:
+      - subtype: hardware
+      - subtype: smi
```
mchmarny: Coverage regression on the fake-nvidia-smi pipeline. The old block pinned the exact values the fake script emits:

```yaml
- subtype: smi
  data:
    gpu.model: NVIDIA B200
    driver: "560.35.03"
    gpu-count: 8
    gpu.product-architecture: Blackwell
```

Those values are deterministic on this test (fake nvidia-smi is injected per the description), and they were the only signal that the smi collector → parser → serializer pipeline doesn't silently drop fields. The new assertion just checks subtype presence, so a bug that empties `data` would still pass.

The Talos test omits values because the Docker-Talos cluster has no fake nvidia-smi (its comment at line 42-45 says exactly that). This test does, so the rationale doesn't carry over. Recommend keeping the `data:` block — it's the one thing on this test that the Talos test can't give you.

tests/chainsaw/snapshot/deploy-agent/chainsaw-test.yaml
```diff
@@ -19,66 +19,60 @@ metadata:
...
-    - name: assert-configmap-metadata
-      description: Verify ConfigMap has correct labels and format key.
-      try:
-        - assert:
-            file: assert-configmap.yaml
+        - script:
+            content: |
+              aicr snapshot \
+                --kubeconfig "${KUBECONFIG}" \
+                --image localhost:5001/aicr:local \
```
mchmarny, line 59: nit: `--kubeconfig "${KUBECONFIG}"` passes an empty string when `KUBECONFIG` is unset (chainsaw won't error on the bare expansion), which then hits `pkg/cli` validation in a way that's confusing to debug. Two options:

1. Drop the flag entirely — the CLI already falls back to `$KUBECONFIG` / `~/.kube/config` per the global flag wiring.
2. Add a fail-fast guard at the top of the script: `: "${KUBECONFIG:?KUBECONFIG must be set}"`.

Talos test has the same pattern, so this is a project-wide nit, not a blocker.

mchmarny (commented):
Refactor direction looks right — `pkg/k8s/agent.Deployer` already does delete-then-create on Job (`pkg/k8s/agent/job.go:34`) and create-or-update on RBAC (`pkg/k8s/agent/rbac.go`), so trimming the chainsaw step to a single CLI invocation removes real fixture drift without losing cleanup correctness.

Two medium concerns on the assertion file, one nit on the CLI invocation:

1. **Order-fragile assertion** — the PR description says it follows the Talos test pattern, but the Talos test specifically uses JMESPath (`sort([*].type)`) to avoid errgroup-induced flakes; this file uses positional-list matching, which is exactly what the Talos comment says not to do. Should be flipped to JMESPath to match the rationale that's already documented.

2. **Lost GPU smi value coverage** — the prior assertion pinned `gpu.model: NVIDIA B200`, `driver: 560.35.03`, etc. from the deterministic fake-nvidia-smi. The new assertion only checks subtype presence. The Talos test skips values because its cluster has no fake GPU; this test does, so keeping the `data:` block here is a strict gain over the Talos pattern.

3. **nit** on `--kubeconfig "${KUBECONFIG}"` — silently passes empty when unset; flag is redundant anyway.

CI in flight (Lint / Test / E2E still running). Non-blocking — the refactor is good, just want the assertion strengthened before merge.

tests/chainsaw/snapshot/deploy-agent/chainsaw-test.yaml
```diff
@@ -19,66 +19,60 @@ metadata:
...
+    up the Job + RBAC. The test asserts only the surviving ConfigMap's
+    snapshot content; Job/RBAC creation are exercised by the CLI exit
+    code, not by separate apply/assert steps.
+
+    Requires:
+      - Kind cluster with fake nvidia-smi injected into worker nodes
+      - aicr image pushed to localhost:5001/aicr:local
+      - aicr binary on PATH (make build writes it to dist/...)
+      - KUBECONFIG pointing at the Kind cluster
     Run with: chainsaw test --test-dir tests/chainsaw/snapshot/deploy-agent
```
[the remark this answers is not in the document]
pdmack: This test deploys a real snapshot agent Job to a Kind cluster — it requires cluster access. The `--no-cluster` flag is for tests that skip cluster interaction (e.g. constraint-only validation). Not applicable here.

tests/chainsaw/snapshot/deploy-agent/assert-snapshot-content.yaml
pdmack: Good catch — adopted. Updated to JMESPath form matching the Talos test pattern: order-independent type and subtype assertions. Verified locally.

tests/chainsaw/snapshot/deploy-agent/assert-snapshot-content.yaml
pdmack: The old hand-crafted Job explicitly mounted the host's fake nvidia-smi into the container at `/usr/bin/nvidia-smi`. The deployer's pod spec does not do this mount — so the smi collector runs but finds no nvidia-smi binary, producing `gpu-count: 0` with no model/driver data. This is the same situation as the Talos test (comment at lines 42-45 of the Talos assertion).

Pinning the GPU `data` values would require either modifying the deployer to mount fake nvidia-smi (out of scope and wrong — the deployer should match production behavior) or hard-coding values that don't reflect what the deployer actually produces. The subtype names confirm both collectors ran; the data gap is a known consequence of the Kind test environment, not a regression.

tests/chainsaw/snapshot/deploy-agent/chainsaw-test.yaml
pdmack: Fair point — dropped `--kubeconfig` entirely. The CLI falls back to `$KUBECONFIG` / `~/.kube/config` via the global flag wiring, which is the correct behavior for both local dev and CI.

tests/chainsaw/snapshot/deploy-agent/chainsaw-test.yaml
```diff
@@ -19,66 +19,59 @@ metadata:
...
-    Deploys the aicr snapshot agent as a Job, then asserts on the K8s
-    resources it creates (Job completion, ConfigMap metadata, snapshot content).
-    Requires: Kind cluster with fake nvidia-smi, default namespace,
-    aicr image at localhost:5001/aicr:local.
+    Invokes `aicr snapshot -o cm://...` against a local Kind cluster.
+    The CLI deploys the snapshot agent Job using pkg/k8s/agent.Deployer,
+    waits for completion, writes the snapshot to a ConfigMap, and cleans
+    up the Job + RBAC. The test asserts only the surviving ConfigMap's
+    snapshot content; Job/RBAC creation are exercised by the CLI exit
+    code, not by separate apply/assert steps.
```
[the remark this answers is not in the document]
pdmack: Good catch — added an `assert-deployer-cleanup` step that checks for leaked Job, SA, Role, RoleBinding (namespace-scoped) and ClusterRole/ClusterRoleBinding (`aicr-node-reader`) after the snapshot completes. Verified locally — all resources are cleaned up.

mchmarny (approved):

Issue #662, opened 2026-04-23
[Feature]: Uniform numbered local-chart bundle layout + --deployer helm adoption


lockwobr:
## Feature Summary

Introduce a uniform numbered local-chart bundle layout (`NNN-<component>/`) and adopt it in `--deployer helm`. `Chart.yaml` presence in a numbered folder signals that local chart content exists; deployers branch on that single fact rather than on component-kind metadata.

Parent: #516.

## Problem / Use Case

Today's `--deployer helm` branches its generated `deploy.sh` on three component kinds (Helm / Kustomize / raw manifests) with distinct install mechanisms and per-case special handling. This makes the script complex, fragile, and hard to share with #632 (helmfile deployer) or a future Flux deployer. A uniform numbered local-chart layout lets every deployer consume one shape and eliminates component-type branching from the install path.

Scope A (this sub-issue) is a pure format refactor — no vendoring of upstream Helm chart bytes. Vendoring lands separately in the follow-up.

## Proposed Solution

**Per-component output** (the "Chart.yaml = local chart content" rule):

- **Helm-only, non-vendored**: `NNN-<name>/values.yaml` + `cluster-values.yaml`. No `Chart.yaml`. Upstream chart reference encoded in the generated `deploy.sh`.
- **Raw-manifest-only** (always wrapped): `NNN-<name>/Chart.yaml` + `templates/*.yaml` (rendered via existing `pkg/manifest` pipeline).
- **Kustomize** (always wrapped; `kustomize build` at bundle time): `NNN-<name>/Chart.yaml` + `templates/manifest.yaml`.
- **Mixed Helm + raw manifests, non-vendored** — emits **two** numbered folders: a primary `NNN-<name>/` (values-only, no Chart.yaml, upstream ref) and an injected `NNN'-<name>-post/` wrapped chart (Chart.yaml + templates/) positioned immediately after. The injected `-post` folder is what deploys the raw manifests. Every folder still follows the Chart.yaml-presence rule — the "mixed" concept disappears at the bundle layer.

**New internal package** (working name `pkg/bundler/localformat/`) owns the layout writer. Reuses existing helpers from `pkg/manifest`, `pkg/component`, and `pkg/bundler/deployer` (SafeJoin, WriteValuesFile, ordering).

**`--deployer helm` rewired** to call the library. `deploy.sh` is a true generic loop — one branch on Chart.yaml presence, no manifests/ pre/post special-casing:

```bash
for dir in [0-9]*/; do
[...]

PR #846, opened 2026-05-11
feat: add support for helm vendoring

### LINKED ISSUE #663: [Feature]: --vendor-charts opt-in flag for air-gap bundles
## Feature Summary

Add a `--vendor-charts` opt-in flag to `aicr bundle` (and the `vendor-charts=true` query parameter on the API server) that pulls upstream Helm chart bytes into the bundle at bundle time using the Helm SDK in-process. With the flag set, every component becomes a local chart in the bundle — enabling air-gap deployability.

Parent: #516. Depends on the base format library landing in the sub-issue that adopts `--deployer helm` first.

## Problem / Use Case

The uniform numbered local-chart layout (scope A) eliminates component-type branching but still references upstream Helm chart registries at deploy time. For air-gap clusters, regulated environments, or deployments over flaky networks, the bundle must be self-contained — every byte needed to install lives inside the artifact.

Vendoring is opt-in for two reasons:
- **CVE-yank signal preservation.** Non-vendored bundles fail loudly when an upstream chart version is yanked (prompting a rebundle with a fixed recipe). Vendored bundles silently deploy the frozen version, outliving the yank signal.
- **Bundle-time cost.** Pulling charts adds bundle-time network dependency, bundle-time auth surface, and bundle size. Users who don't need air-gap shouldn't pay these costs by default.

## Proposed Solution

Add `--vendor-charts` (bool, default false) to `aicr bundle` and the corresponding API query parameter.

When set, the base-format library (from the sub-issue 1 work) pulls upstream Helm chart bytes in-process via `helm.sh/helm/v3` (`repo.Entry`, `downloader.ChartDownloader`, `chartutil.ExpandFile`) and writes them to `NNN-<name>/charts/<chart>/` alongside a wrapper `Chart.yaml`.

For mixed Helm + raw-manifest components, `--vendor-charts` collapses the two-folder primary + `-post` emission from sub-issue 1 into a single wrapped folder. The vendored chart bytes go under `charts/<upstream>/`; raw manifests move into `templates/*.yaml` as helm hooks (`helm.sh/hook: post-install`) inside the wrapper. Deploy-time ordering preserved without the injected `-post` Application.

Wrapped Kustomize / raw-manifest-only components are unaffected — they're already local after the layout refactor.

**Bundle-time audit log.** Per vendored chart: `(name, version, source URL, SHA256 digest)`. Written to the bundle (e.g. `provenance.json`) so downstream yank-list lookups are possible.

**Auth.** Reuse existing Helm conventions (`HELM_REPOSITORY_USERNAME`, `HELM_REPOSITORY_PASSWORD`) plus an explicit config block for the API server.

## Success Criteria

- `--vendor-charts` flag works for CLI and API server.
- Vendored bundle installs end-to-end against KWOK with registry egress blocked at deploy time.
- Each vendored chart has an audit-log entry with name, version, source URL, and SHA256.
- Mixed components emit a single wrapped folder (collapsing sub-issue 1's primary + `-post` two-folder shape); raw manifests become helm hooks with correct post-install ordering.
- Bundle-time auth works for authenticated chart registries via Helm env conventions.
- `docs/user/cli-reference.md` documents the flag, the CVE-yank tradeoff, and the audit-log contents.
- `make qualify` clean; test coverage gate satisfied.

## Alternatives Considered

- **Shell out to `helm pull`.** Requires helm CLI in the aicr bundle environment (and the API server pod). Rejected in favor of in-process Helm SDK for deploy-dependency hygiene.
- **Direct HTTP / go-containerregistry.** Smaller dependency footprint but forces us to reimplement chart repo semantics and provenance. Rejected.
- **Mandatory vendoring (no flag).** Simpler code path but imposes bundle-time network / auth / size costs on every user and makes offline bundling impossible. Rejected.

## Component

## Priority

Additive and opt-in. Non-vendored behavior is the default and unchanged. Users who don't set `--vendor-charts` see zero difference from today's behavior (after sub-issue 1 lands).

## Operational Considerations

- Bundle-time network egress required when flag is set — including for the API server pod.
- Bundle-time credentials required for private chart registries.
- Bundle size grows substantially (0.5–5 MB per chart unpacked; meaningful for OCI image publishing).
- CVE-yank side effect documented alongside the flag: vendored bundles outlive upstream yank signals.

Maybe, with guidance

lockwobr:
## Summary

Add `--vendor-charts` opt-in flag (CLI + API + config-file) that pulls upstream Helm chart bytes into the bundle at bundle time, producing a self-contained, air-gap-deployable artifact with a `provenance.yaml` audit log.

## Motivation / Context

The uniform numbered local-chart layout from #662 still references upstream Helm chart registries at deploy time, so
bundles can't deploy in air-gapped clusters, behind flaky networks, or against private registries that aren't reachable
at install time. `--vendor-charts` collapses every Helm-typed component into a single wrapped local chart with the
upstream tarball checked in at `NNN-<name>/charts/<chart>-<version>.tgz`, eliminating deploy-time registry egress.

Vendoring is intentionally opt-in for two reasons documented inline:
- **CVE-yank fail-loud signal preservation.** Non-vendored bundles fail at deploy time when an upstream chart version is yanked. Vendored bundles freeze the bytes and outlive the yank — `provenance.yaml` exposes the audit surface for cross-referencing yank lists.
- **Bundle-time cost.** Vendoring adds network egress, auth surface, and ~0.5–5 MB per chart. Users who don't need
air-gap shouldn't pay. 

Closes: #663, #516
Related: #516 (epic), #662 (predecessor layout)

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Implementation Notes

**ChartPuller shim, not the Helm SDK (yet).** Vendoring `helm.sh/helm/v4/pkg/downloader` transitively pulls in `github.com/cyphar/filepath-securejoin` (MPL-2.0), which the repo allowlist (Apache-2.0/MIT/BSD/ISC/Zlib only — explicitly "DO NOT ADD IGNORES") rejects. To unblock, the bundler calls a `ChartPuller` interface in `pkg/bundler/deployer/localformat/vendor.go`. Today's implementation is `CLIChartPuller`, which shells out to `helm pull` via `exec.CommandContext` (no shell). When legal approves the SDK, swap in an `SDKChartPuller` alongside; nothing else changes.

**Single wrapped folder per component.** Mixed components (Helm + raw manifests) used to emit primary + `-post` two-folder shape (#662). With `--vendor-charts`, they collapse into one folder: wrapper `Chart.yaml` declares the vendored chart as a `dependencies:` entry with empty `repository:` (Helm resolves from `charts/<chart>-<ver>.tgz` adjacent), values are nested under the subchart name so Helm forwards them, and recipe-side raw manifests live in `templates/` with `helm.sh/hook: post-install` + stable hook-weight injected.

**Provenance is shared, K8s-shape YAML.** `pkg/bundler/deployer/localformat/WriteProvenance` emits `provenance.yaml` at the bundle root with `apiVersion: aicr.nvidia.com/v1alpha1`, `kind: BundleProvenance`, and one entry per vendored chart `(name, chart, version, repository, sha256, tarballName, pullerVersion)`. Called by all three deployer Generators — `helm`, `argocd`, and `argocd-helm` — so air-gap audit works regardless of `--deployer`. Format matches every other AICR persisted file (Recipe, Snapshot, RecipeCriteria); pipe through `yq -o=json` for tools that expect JSON.

**Hook injection uses `gopkg.in/yaml.v3`** for streaming document boundaries — handles leading `---`, `---` inside literal-block strings, and interleaved comments correctly.

**`Write` returns a struct (`WriteResult`).** Previous `([]Folder, error)` became `([]Folder, []VendorRecord, error)` while iterating; settled on `WriteResult{Folders, VendoredCharts}` so future fields don't keep breaking the signature.

**Auth flows through Helm conventions** unchanged: `HELM_REPOSITORY_USERNAME`/`HELM_REPOSITORY_PASSWORD` for HTTP(S); docker config (`~/.docker/config.json` or `$DOCKER_CONFIG`) for OCI. Documented in `docs/user/cli-reference.md` ("Vendoring Charts for Air-Gap" section) and `docs/user/api-reference.md`.

## Testing

```bash
make qualify
```

## Risk Assessment

**Rollout notes:** 

## Checklist


pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -0,0 +1,332 @@
...
+			fmt.Sprintf("vendor-charts: invalid repository URL %q for %q",
+				c.Repository, c.Name), perr)
+	}
+	if u.Scheme != "http" && u.Scheme != "https" {
+		return errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("vendor-charts: repository URL %q for %q must use http or https scheme (got %q); use IsOCI for oci:// repos",
+				c.Repository, c.Name, u.Scheme))
+	}
+	return nil
+}
```
mchmarny, line 366: Medium: `validateForPull` accepts a `chartName` (or `c.Name` fallback) that starts with `--`, which lets the HTTP(S) path inject helm flags.

In the HTTP branch above (line 137):
```go
args = append(args, chartName, "--repo", c.Repository)
```

If a recipe sets `chartName: "--insecure-skip-tls-verify"` (or `--ca-file=/some/path`, `--keyring=...`), helm sees it as a flag rather than a chart name. `exec.CommandContext` is shell-free so OS-level shell injection isn't possible, but **helm flag injection** is — including flags that weaken TLS verification.

The OCI branch (line 135) is safe because chartName is concatenated into a single URL token (`oci://host/repo/--debug`), not a separate argv element.

Threat model today is repo-controlled recipes, so this is medium not high — but `vendor-charts` also flows through the API per `pkg/bundler/handler.go`, which raises the defense-in-depth bar. Suggest validating both `ChartName` and `Name` against `^[a-zA-Z0-9][a-zA-Z0-9._-]*$` here so flag-looking values are rejected at the boundary regardless of caller.

pkg/bundler/deployer/helm/helm.go
```diff
@@ -180,6 +195,18 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 		return nil, err
 	}
 
+	// Generate provenance.json for vendored bundles. Written before
+	// checksums.txt so the audit file is itself checksummed.
+	if len(g.vendorRecords) > 0 {
+		provPath, provSize, provErr := localformat.WriteProvenance(ctx, outputDir, g.vendorRecords)
+		if provErr != nil {
+			return nil, errors.Wrap(errors.ErrCodeInternal,
+				"failed to generate provenance.json", provErr)
```
mchmarny: Low: this error says `"failed to generate provenance.json"` but `WriteProvenance` writes `provenance.yaml` (see `ProvenanceFileName` in `localformat/provenance.go`). User-visible error message points at a file that doesn't exist.

Five instances of `provenance.json` remain in this PR — `helm.go:100`, `helm.go:198`, `helm.go:204` (this line), `localformat/writer.go:51`, `localformat/writer.go:119`. Looks like stale references from an earlier draft. The error string here is the only user-visible one; the others are stale comments. Worth a global s/provenance.json/provenance.yaml/.

pkg/bundler/deployer/localformat/vendor_wrapper.go
```diff
@@ -0,0 +1,202 @@
...
+	meta, _ := doc["metadata"].(map[string]any)
+	if meta == nil {
+		meta = map[string]any{}
+		doc["metadata"] = meta
+	}
+	annos, _ := meta["annotations"].(map[string]any)
+	if annos == nil {
+		annos = map[string]any{}
+		meta["annotations"] = annos
+	}
```
mchmarny, line 196: Low: the `meta["annotations"].(map[string]any)` assertion silently fails when `annotations` is present but typed differently (e.g., produced by a YAML alias merge, an upstream tool emitting `map[interface{}]interface{}`, or a strict-typed Go marshal). In that case the function creates a fresh annotations map and **overwrites** the existing one, losing all pre-existing annotations.

yaml.v3 normally produces `map[string]any` for default decoding, so this is unlikely in practice — but the failure mode (silent annotation loss) is severe enough to warrant either a defensive log/error or a more permissive type-switch that preserves existing keys. Not blocking.

pkg/bundler/deployer/localformat/vendor_wrapper.go
```diff
@@ -0,0 +1,202 @@
...
+	if _, alreadyHooked := annos["helm.sh/hook"]; alreadyHooked {
+		// Honor an author-declared hook (could be pre-install,
+		// post-upgrade, etc.). Don't overwrite, don't consume a weight.
+		return false
+	}
+	annos["helm.sh/hook"] = "post-install"
+	annos["helm.sh/hook-weight"] = strconv.Itoa(weight)
+	annos["helm.sh/hook-delete-policy"] = "before-hook-creation"
```
mchmarny: Low: `helm.sh/hook-delete-policy: before-hook-creation` is set unconditionally on every post-install resource. For mixed-component manifests that include stateful resources (ConfigMaps that record install metadata, jobs that should run once, custom resources representing state), this deletes-and-recreates on every helm upgrade — potentially losing state the recipe author expected to persist.

Today's recipes likely don't depend on this, but it's worth either:
- documenting the assumption in the package doc, or
- letting the recipe author opt out by setting `helm.sh/hook-delete-policy` themselves (the function already honors a pre-existing `helm.sh/hook`, but it unconditionally clobbers `hook-delete-policy`).

The check at line 193 already gates on `helm.sh/hook`; extending the same "honor author-declared" rule to `hook-delete-policy` would keep the behavior consistent across the three annotations being injected. Not blocking.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -0,0 +1,332 @@
...
+// Bundle-time chart pulling for --vendor-charts.
+//
+// Strategy is hidden behind ChartPuller so the rest of the bundler does
+// not care HOW upstream chart bytes were obtained — only that we have a
+// .tgz and a provenance record. Today we shell out to `helm pull`
+// (CLIChartPuller) because pulling the in-process Helm SDK transitively
+// vendors github.com/cyphar/filepath-securejoin (MPL-2.0), which is not
+// on the AICR license allowlist (Makefile: license-check). When legal
+// approves the SDK we add an SDKChartPuller alongside CLIChartPuller and
+// flip the default; nothing else in this package changes.
```
mchmarny, line 24: Positive: the explicit "MPL-2.0 transitive dep blocks SDK adoption today, swap to SDKChartPuller alongside CLIChartPuller when legal clears it" doc is exactly the kind of context that vanishes from PR descriptions over time. Future maintainers won't have to re-discover the constraint, and the `ChartPuller` interface boundary makes the swap mechanically trivial. Good shape.

tests/chainsaw/cli/bundle-vendor-charts/chainsaw-test.yaml
```diff
@@ -0,0 +1,218 @@
...
+        - script:
+            content: |
+              set -eu
+              AICR_BIN="${AICR_BIN:?Set AICR_BIN to the path of the aicr binary}"
+              WORK="/tmp/chainsaw-bundle-vendor-charts"
+              rm -rf "${WORK}" && mkdir -p "${WORK}/fakebin"
+
+              ${AICR_BIN} recipe --service eks --accelerator gb200 --os ubuntu \
+                --intent training -o "${WORK}/recipe.yaml"
+
```
mchmarny, line 62: Positive: hermetic fake-helm shim is the right call here. Tests the full vendor path including `helm version --short` (audit-log probe) and `helm pull --destination` semantics without requiring helm in CI. Adds the missing-binary error case too, which is exactly the failure surface most likely to bite an operator.

mchmarny (commented):
Solid shape overall — the `ChartPuller` interface boundary, the documented license rationale for shelling out vs SDK, the `WriteResult` struct for forward-compat, and the hermetic chainsaw fake-helm shim are all clean choices. The provenance.yaml audit log emitted from all three deployers is exactly the right place for the bundle-time freeze record.

One medium concern on `validateForPull`: the HTTP(S) branch of `helm pull` passes `chartName` as a separate argv token, so a chart name beginning with `--` (e.g., `--insecure-skip-tls-verify`) injects as a helm flag rather than a chart name. The OCI branch is safe (single URL token). Recipes are repo-controlled today, but `--vendor-charts` also flows through the API surface so input validation at the boundary is the right defense.

Three lows: user-visible error message references `provenance.json` while the file is `provenance.yaml` (five stale references total, only `helm.go:204` is user-visible); `injectHookOnDoc` can silently overwrite existing annotations on an unexpected map type; `helm.sh/hook-delete-policy: before-hook-creation` is unconditional and may surprise mixed-component recipes that include stateful resources.

CI: lint, build, security scan, malware-scan all green at review time; Test, E2E, analyze, GPU smoke still in flight; `mergeable_state: blocked` (review gate). Nothing blocks merge from this review.

pkg/bundler/deployer/localformat/vendor.go
lockwobr: Addressed. Took your suggestion — full allowlist regex at the boundary rather than just a leading-`-` check:

```go
var safeChartNameRE    = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`)
var safeChartVersionRE = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9._+-]*$`)  // `+` allowed for semver build metadata
```

Applied to `ChartName` (or `Name` fallback) AND `Version` in `validateForPull`. New table cases cover the attack surface (flag-shaped names/versions, shell-meta, whitespace, leading dot) plus regression rows for legitimate inputs (`gpu-operator`, `1.2.3+build.7`).

pkg/bundler/deployer/helm/helm.go
lockwobr: Fixed. Did a repo-wide `git grep provenance.json` after your note and renamed all 5 stragglers: `helm.go` (3 sites including the user-visible error string), `localformat/writer.go` (2 sites), `argocd.go` (2 sites). `git grep provenance\.json` is now clean.

pkg/bundler/deployer/localformat/vendor_wrapper.go
lockwobr: Fixed. Added an `annotationsAsStringMap` helper that handles both `map[string]any` (yaml.v3 default) and `map[any]any` (yaml.v2 / upstream-round-tripped) shapes, preserving existing keys in either case. Pinned by `TestInjectPostInstallHooks_PreservesExistingAnnotations`.

pkg/bundler/deployer/localformat/vendor_wrapper.go
lockwobr: Fixed by extending the same "honor author-declared" rule to `hook-delete-policy`:

```go
if _, alreadySetPolicy := annos["helm.sh/hook-delete-policy"]; \!alreadySetPolicy {
    annos["helm.sh/hook-delete-policy"] = "before-hook-creation"
}
```

Pinned by `TestInjectPostInstallHooks_HonorsAuthorDeletePolicy`. A recipe author who explicitly sets the policy (e.g., `hook-succeeded` to preserve state across upgrades) is now respected.

pkg/bundler/deployer/localformat/vendor.go
lockwobr: Thanks — that doc was specifically meant to survive PR-description rot. Glad it landed.

tests/chainsaw/cli/bundle-vendor-charts/chainsaw-test.yaml
lockwobr: Thanks. Also tightened it in a later round: collapsed the steps into one and switched the workdir to `mktemp -d` with an EXIT trap so parallel runs/retries can`t collide.

mchmarny (approved):

PR #852, opened 2026-05-12
refactor(bundler): expose SignStatement primitive for predicate-agnostic keyless signing

njhensley:
## Summary

Extract the Sigstore Fulcio + Rekor keyless-signing plumbing from `KeylessAttester.Attest` into a reusable `SignStatement` primitive in `pkg/bundler/attestation`, so the upcoming recipe-evidence pipeline (#754) can share it instead of duplicating ~80 lines of identical code.

## Motivation / Context

ADR-007 (recipe-test-attestation, #754) introduces a second producer of in-toto Statements signed via Sigstore keyless OIDC — same DSSE envelope, same Fulcio cert flow, same Rekor inclusion proof, same protobuf-JSON bundle — but with a different predicate type (`https://aicr.nvidia.com/recipe-evidence/v1`). Today, all of that plumbing lives inside `KeylessAttester.Attest` and is hard-wired to the SLSA Provenance v1 predicate the bundler emits. Without this split, `pkg/evidence/attestation/signer.go` would copy/paste the same 80 LoC.

This refactor isolates the predicate-agnostic half so the second producer becomes a thin client of the existing primitive. It is intentionally a pure refactor — no observable behavior change for `aicr bundle --attest`.

Fixes: N/A
Related: #754, ADR-007 (`docs/design/007-recipe-evidence.md`)

## Type of Change

## Component(s) Affected

## Implementation Notes

**New API surface in `pkg/bundler/attestation`:**

```go
type SignOptions struct {
    OIDCToken string  // required
    FulcioURL string  // empty → DefaultFulcioURL
    RekorURL  string  // empty → DefaultRekorURL
}

type SignedAttestation struct {
    BundleJSON    []byte  // protobuf-JSON Sigstore bundle
    Identity      string  // SAN (email or URI) from Fulcio cert
    Issuer        string  // OIDC issuer URL from Fulcio cert extension
    RekorLogIndex int64   // first transparency-log entry, 0 if absent
}

func SignStatement(ctx context.Context, statementJSON []byte, opts SignOptions) (*SignedAttestation, error)
```

`SignStatement` takes pre-built in-toto Statement bytes (so callers own the predicate) and returns the full Sigstore bundle plus the identity claims extracted from the Fulcio certificate. The returned `SignedAttestation` is a strict superset of what `Attest` used to surface internally — adding the OIDC issuer and Rekor log index — so the new evidence pointer file can populate its full signer block without re-parsing the certificate.

**Internal cleanup, also in this PR:**

- `extractSignerClaims` (identity + issuer) replaces `extractSignerIdentity` as the canonical extractor; the old function is retained as a thin wrapper so existing assertions in `keyless_test.go` stand unchanged.
- `extractIssuerExtension` parses the Fulcio OIDC-issuer X.509 extension (OIDs `1.3.6.1.4.1.57264.1.1` legacy and `.1.8` current).
- `extractRekorLogIndex` returns the first transparency-log entry's index, or 0 when no Rekor entry was created.

**`KeylessAttester.Attest`** keeps its public signature and semantics: it still builds the SLSA Provenance v1 Statement, then delegates DSSE wrap + Fulcio + Rekor + bundle marshal to `SignStatement`. Existing callers see no change.

**Considered and deferred:** promoting the primitive to its own top-level package (e.g., `pkg/sigstore`). Decided against in this PR to keep the diff small; the only consumer outside `pkg/bundler/attestation` will be `pkg/evidence/attestation` (PR-Z), which can live with an import alias.

## Testing

```bash
make qualify
```

Per-package results on `pkg/bundler/attestation`:

- `go test ./pkg/bundler/attestation/...` — PASS
- `golangci-lint run` — 0 issues
- Coverage: **63.0% → 65.7%** (the new extractors are at 100%; `SignStatement` itself is at 16% — the live Fulcio/Rekor call path still requires integration tests, same as before this PR)

New `signing_test.go` covers:

- `SignStatement` input validation (empty statement, missing OIDC token)
- `extractSignerClaims` for email-SAN certs (interactive OIDC) and URI-SAN certs (workload OIDC, e.g. GitHub Actions)
- `extractIssuerExtension` for the current Fulcio OID (`.1.8`)
- `extractRekorLogIndex` for present/absent/nil `VerificationMaterial` cases

## Risk Assessment

**Rollout notes:** Pure internal refactor. No CLI flag, API, on-disk artifact, or wire-format changes. `aicr bundle --attest` continues to produce byte-identical Sigstore bundles for the same input.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — internal refactor)


mchmarny (dismissed):

pkg/bundler/attestation/keyless.go
```diff
@@ -54,63 +51,31 @@ func NewKeylessAttester(oidcToken string) *KeylessAttester {
...
-
-	// Marshal bundle to JSON (Sigstore bundle format)
-	bundleJSON, err := protojson.Marshal(bundle)
-	if err != nil {
-		return nil, errors.Wrap(errors.ErrCodeInternal, "failed to marshal sigstore bundle", err)
+		return nil, err
 	}
 
+	k.identity = res.Identity
 	slog.Info("bundle attestation signed successfully", "identity", k.identity)
```
mchmarny: PII inconsistency. `SignStatement` deliberately demoted identity to a Debug-level log because the Fulcio SAN is user PII for interactive OIDC (signing.go comment is explicit about this). But `Attest` still emits the same identity at Info level here — so for any caller going through `KeylessAttester`, the PII rationale is undone.

Two reasonable options:
- Demote this line to Debug too, keeping a generic "bundle attestation signed successfully" at Info.
- Or, if Info-level identity is intentional for the bundle flow specifically (audit story?), drop the contradictory PII comment in signing.go to keep the policy consistent.

pkg/bundler/attestation/keyless.go
```diff
@@ -126,42 +91,12 @@ func (k *KeylessAttester) HasRekorEntry() bool {
...
 }
 
-// extractSignerIdentity parses the Fulcio signing certificate from a Sigstore
-// bundle protobuf and returns the SubjectAlternativeName (email or URI).
-// Returns empty string if the certificate cannot be extracted.
+// extractSignerIdentity parses the Fulcio signing certificate from a
+// Sigstore bundle protobuf and returns the SubjectAlternativeName
+// (email or URI). Returns empty string when the certificate cannot be
+// extracted. Thin wrapper around extractSignerClaims preserved for
+// the existing tests that pin this function's contract.
```
mchmarny: nit: after this refactor `extractSignerIdentity` is no longer called from production code — only `keyless_test.go` exercises it. The "preserved for the existing tests" justification means we're shipping prod code that exists solely to keep a test surface stable. Cleaner to retarget the seven `TestExtractSignerIdentity_*` cases at `extractSignerClaims` (same package, can call directly) and drop this wrapper.

Fine as a follow-up if you want to keep this PR purely additive — but worth doing soon, otherwise it'll calcify.

pkg/bundler/attestation/signing.go
mchmarny, line 1: Priority is undefined when both Fulcio OIDs are present. The `for`/`switch` returns on the first match it encounters, which means the result depends on the X.509 extension ordering in the cert, not on any explicit policy.

In practice Fulcio stamps one or the other (legacy → current), and the values should match, so this isn't a correctness bug today. But the doc comment says "Try both" without saying "prefer current" — if Fulcio ever ships transitional certs with both extensions and a tiny encoding divergence, behavior here becomes order-dependent and silent.

Suggest making the preference explicit: scan for `current` first; only fall back to `legacy` if `current` is absent. Two passes over a tiny extension list is free.

mchmarny, line 1: nit: the `Identity` doc says "Empty if extraction failed" but it can also be empty when the cert genuinely has no email/URI SAN (the `TestExtractSignerClaims` "no SAN" path in `keyless_test.go`). Worth a one-word tweak: "Empty if extraction failed *or the cert has no SAN*", so callers don't treat empty as an error signal.

mchmarny (commented):
Clean refactor — the split between predicate-agnostic signing and the SLSA-specific `Attest` wrapper is the right cut for the upcoming recipe-evidence pipeline. ASN.1-aware handling of the current Fulcio OID and the new Sigstore signing timeout are both improvements over the prior code, not just relocations. Tests cover the new extractors well.

One medium concern worth resolving before merge: the PII rationale you added to `SignStatement` (identity at Debug only) is undermined by `KeylessAttester.Attest` still logging identity at Info — pick one policy and apply it consistently.

Two low/nit items: `extractIssuerExtension` returns whichever Fulcio OID appears first in the cert rather than preferring the current OID, and `extractSignerIdentity` is now production code maintained only for tests. Neither blocks merge.

CI: test/lint/e2e jobs all green; `gate` and a few scanners (CodeQL, Grype, ClamAV) still in progress.

pkg/bundler/attestation/keyless.go
njhensley: Done in 3856ef3. Demoted the post-sign `slog.Info` in `KeylessAttester.Attest` so identity flows to Debug only — mirrors the `SignStatement` contract. Generic "bundle attestation signed successfully" stays at INFO; `KeylessAttester.Identity()` is still the audit-trail accessor for callers that need the value.

pkg/bundler/attestation/keyless.go
njhensley: Done in 3856ef3, not deferred. `extractSignerIdentity` deleted; the 7 `TestExtractSignerIdentity_*` cases now call `extractSignerClaims` directly and were renamed `TestExtractSignerClaims_*`. One name collision (`_NilBundle` — signing_test has an empty-struct case, keyless_test had a nil-pointer case) resolved by renaming the keyless_test one to `_NilBundlePointer` since the two cover materially different inputs. Also dropped the now-unused `protobundle` import from `keyless.go`.

pkg/bundler/attestation/signing.go
njhensley: Done in 3856ef3. `extractIssuerExtension` is now a two-pass scan: current OID (`.1.8`) first, fall back to legacy (`.1.1`) only if absent. Behavior is order-independent. Doc-comment refreshed to spell out the policy, and the call-site comment in `extractSignerClaims` (was "Try both") updated to "prefers current and falls back to legacy." Added two regression tests: `current OID takes precedence over legacy regardless of cert order` and `legacy OID used when current OID is absent`, pinning the invariant in both directions.

njhensley: Done in 3856ef3. The `Identity` field doc now reads: "Empty if extraction failed or the cert has no email/URI SAN — callers must not treat empty as an error signal."

pkg/bundler/attestation/signing.go
```diff
@@ -0,0 +1,253 @@
...
+			IDToken: opts.OIDCToken,
+		},
+		TransparencyLogs: []sign.Transparency{
+			sign.NewRekor(&sign.RekorOptions{BaseURL: rekorURL}),
+		},
+		Context: signCtx,
+	})
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeUnavailable, "sigstore signing failed", err)
+	}
```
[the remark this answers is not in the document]
njhensley: Addressed in e98d8c4. `SignStatement` now wraps `sign.Bundle` errors with `stderrors.Is(err, context.DeadlineExceeded)` and returns `ErrCodeTimeout` on hit; everything else stays on `ErrCodeUnavailable`. Matches the in-package convention in `classifyOIDCContextError` (`oidc.go`), which classifies on the propagated err rather than re-reading `ctx.Err()` — so the call site stays consistent across this package.

Comment at the wrap site explains the 502-vs-504 API-boundary distinction so future readers don't collapse the two codes.

mchmarny (approved):

PR #853, opened 2026-05-12
feat(config): aicr validate --config support

njhensley:
## Summary

Extend AICRConfig with a `spec.validate` section parallel to the existing `spec.recipe` and `spec.bundle`, and wire `aicr validate --config` to load it. CLI flags continue to override config values; override events log at INFO.

## Motivation / Context

Today, AICRConfig covers `aicr recipe` and `aicr bundle` but not `aicr validate`. Anyone who wants to run the same validation across clusters has to retype every flag — agent namespace, image, image-pull-secrets, node selectors, tolerations, timeout, phases, etc. — instead of capturing them once in a config file.

This PR closes that gap for the existing validate flag surface. The evidence-related flags (`--evidence-dir`, `--cncf-submission`, `--feature` today, plus the attestation flags landing under #754) are intentionally **not** in the schema yet: the recipe-test-attestation work introduces an Evidence umbrella that re-homes CNCF alongside the new attestation kind, and landing the umbrella with both members at once avoids a churn window where a one-member umbrella would have to be reshaped on the second member's arrival.

Fixes: N/A
Related: #754, ADR-007 (`docs/design/007-recipe-evidence.md`)

## Type of Change

## Component(s) Affected

- [x] Core libraries (`pkg/errors`, `pkg/k8s`) — `pkg/config` extension

## Implementation Notes

**Schema, under `spec.validate`:**

```yaml
spec:
  validate:
    input:
      recipe: ./recipe.yaml
      snapshot: ./snapshot.yaml          # optional; omit to capture live
    agent:                               # only used when input.snapshot is empty
      namespace: aicr-validation
      image: ghcr.io/nvidia/aicr:v0.1.0
      imagePullSecrets: [registry-secret]
      jobName: aicr-validate
      serviceAccountName: aicr
      nodeSelector:
        my-org/gpu-pool: "true"
      tolerations:
        - "gpu-type=h100:NoSchedule"
      requireGpu: true
    execution:
      phases: [deployment, conformance]
      failOnError: true                  # default true; *bool preserves nil-vs-set
      noCluster: false
      noCleanup: false
      timeout: 10m                       # parsed at Resolve time
```

**New code:**

- `pkg/config/config.go` — `ValidateSpec`, `ValidateInputSpec`, `ValidateAgentSpec`, `ValidateExecutionSpec`. `Spec.Validate *ValidateSpec` added.
- `pkg/config/resolve.go` — `ValidateResolved` + `(*ValidateSpec).Resolve()`. Mirrors `BundleSpec.Resolve` patterns: tolerations parsed via `snapshotter.ParseTolerations`, timeout parsed via `time.ParseDuration` with `>= 0` guard, maps/slices defensively cloned, nil-vs-explicitly-empty preserved.
- `pkg/config/accessors.go` — `(*AICRConfig).Validation()`. Named `Validation` (not `Validate`) to avoid colliding with the existing top-level `(*AICRConfig).Validate()` config-soundness method.
- `pkg/config/validate.go` — top-level `Validate()` now accepts validate-only configs and surfaces duration/toleration parse errors at config-load time (via the shared `Resolve`-as-validator pattern bundle uses).
- `pkg/cli/root.go` — new `durationFlagOrConfig` helper. The existing `boolFlagOrConfig` / `stringSliceFlagOrConfig` in `pkg/cli/bundle_config.go` are reused as-is.
- `pkg/cli/validate.go` — `--config` flag added; `parseValidateAgentConfig` takes `*ValidateResolved`; new `resolveValidateNodeSelector` / `resolveValidateTolerations` helpers preserve the "no flag" sentinel for `--toleration` (inner validators distinguish "operator opted into tolerate-all" from silence); Action body uses CLI-overrides-config precedence throughout.

**Lossy bool semantics (intentional):** `boolFlagOrConfig` (in `bundle_config.go`) takes a `bool` fallback, not `*bool`. That makes `noCleanup: false` in config indistinguishable from "config unset." Acceptable for the three lossy fields (`NoCluster`, `NoCleanup`, `RequireGPU`) because their CLI defaults are also `false` — both branches resolve to the same observable behavior. `FailOnError` is the one case where the distinction matters (CLI default = `true`); `ValidateResolved.FailOnError` is `*bool` to preserve it, and the call site uses a small `derefBoolOr` helper to fold nil-pointer-or-default back into a bool for `boolFlagOrConfig`.

**Not in this PR (scope guardrails):**

- `EvidenceSpec`, `EvidenceCNCFSpec`, `EvidenceAttestationSpec` — land with the recipe-evidence work under #754 with both members present.
- The `--evidence-dir`, `--cncf-submission`, `--feature` flags remain CLI-only. Docs call this out explicitly.
- The new `--emit-attestation`, `--bom`, `--include-logs`, `--push`, etc. flags introduced by #754.

## Testing

```bash
make qualify
```

Results:

- `go build ./...` — clean
- `go test ./pkg/config/... ./pkg/cli/...` — PASS
- `golangci-lint run ./...` — 0 issues
- `pkg/config` coverage: **95.9%** overall; all new code (`Validation`, `ValidateSpec.Resolve`, `ValidateSpec.validate`) at **100%**

New tests:

- `pkg/config/config_test.go`: `TestValidate_ValidateOnly`, `TestValidate_InvalidValidateTimeout`, updated `TestValidate_Errors` to cover the new "none of recipe/bundle/validate" message.
- `pkg/config/resolve_test.go`: 9 cases for `ValidateSpec.Resolve` — nil receiver, empty spec, all-fields populated, invalid timeout, negative timeout, invalid toleration, defensive clone of NodeSelector, nil-vs-empty Tolerations.
- `pkg/config/accessors_test.go`: nil-tolerance and populated-spec cases for `Validation()`.

## Risk Assessment

- [x] **Low** — Additive change. No existing AICRConfig file becomes invalid (recipe-only and bundle-only configs still validate). No CLI flag default changes. `aicr validate` without `--config` runs exactly as before.

**Rollout notes:** No migration. Users who want config-driven validation set `spec.validate` in their existing AICRConfig file. Existing CLI invocations are unaffected.

## Checklist

- [x] I updated docs if user-facing behavior changed (`docs/user/cli-reference.md`)


pkg/cli/validate.go
```diff
@@ -561,30 +620,21 @@ Run validation without failing on check errors (informational mode):
...
-				}
+			tolerations, err := resolveValidateTolerations(cmd, resolved)
+			if err != nil {
+				return err
 			}
-
-			nodeSelector, nsErr := snapshotter.ParseNodeSelectors(cmd.StringSlice("node-selector"))
-			if nsErr != nil {
-				return errors.Wrap(errors.ErrCodeInvalidRequest, "invalid node-selector", nsErr)
+			nodeSelector, err := resolveValidateNodeSelector(cmd, resolved)
```
mchmarny: Duplicate resolver work in the agent-deploy path. When `--snapshot` is empty, `parseValidateAgentConfig` (called above) already invokes `resolveValidateTolerations`/`resolveValidateNodeSelector` for the snapshot-capture agent's PodSpec. Calling them again here for the validator job re-parses the same inputs, and — because both helpers emit `slog.Info("CLI flag overriding config value", ...)` when CLI overrides config — operators will see each override line logged twice in that branch.

Suggest resolving `tolerations` / `nodeSelector` (and `noCleanup` on line 637, which has the same double-`slog.Info` issue) once before the snapshot vs. agent-deploy split, then passing those values into `parseValidateAgentConfig` instead of re-resolving.

pkg/cli/validate.go
```diff
@@ -535,7 +594,7 @@ Run validation without failing on check errors (informational mode):
 			// snapshot from the live API), so a snapshot file is the only valid
 			// data source in that mode. Placed after recipe.LoadFromFile so
 			// recipe kind-check and auto-hydration still run for CLI coverage.
-			if snapshotFilePath == "" && cmd.Bool("no-cluster") {
+			if snapshotFilePath == "" && noCluster {
 				return errors.New(errors.ErrCodeInvalidRequest,
 					"--no-cluster requires --snapshot (cannot deploy the snapshot-capture agent without cluster access)")
```
mchmarny: For consistency with the `--recipe is required (or set spec.validate.input.recipe in --config)` message just above, mention the config alternative here too — both inputs can now come from `--config`, so a user who set `noCluster: true` in config and forgot the snapshot field shouldn't be told to pass `--snapshot` on the CLI:

```
"--no-cluster requires --snapshot (or spec.validate.input.snapshot in --config); cannot deploy the snapshot-capture agent without cluster access"
```

pkg/cli/config_helpers_test.go
```diff
@@ -82,6 +82,19 @@ func TestStringFlagOrConfig(t *testing.T) {
 			fallback: "",
 			want:     "",
 		},
+		{
+			// Regression: when config has no value for a field, the
+			// flag's compile-time default must surface — not the empty
+			// string. Found via PR-W e2e: --namespace (Value:
```
mchmarny: nit: `PR-W` isn't a meaningful reference to a future reader (and won't survive grep). Either drop the citation or replace it with the concrete scenario, e.g. "regression: a flag with a compile-time `Value:` default (e.g. `--namespace` defaults to `aicr-validation`) collapsed to `""` when `--config` was passed but did not set the field, breaking namespace creation downstream."

pkg/config/resolve.go
```diff
@@ -227,6 +229,150 @@ func (b *BundleSpec) Resolve() (*BundleResolved, error) {
...
+// pkg/validator import for what is just a string-set check; the CLI
+// layer (parseValidationPhases) re-validates with the same vocabulary
+// when it converts strings to typed validator.Phase values.
+var validPhases = map[string]bool{
+	"deployment":  true,
+	"performance": true,
+	"conformance": true,
+	"all":         true,
+}
+
```
mchmarny: nit: the `validPhases` set duplicates the phase vocabulary that `parseValidationPhases` re-validates at the CLI layer. The comment acknowledges the import-cycle trade-off, but a single source of truth (e.g. exporting a `validator.PhaseNames()` slice / a `phase.Validate(string)` from a leaf package that neither side of the cycle depends on) would prevent drift the next time a phase is added or removed.

mchmarny (commented):
Solid extension — schema mirrors `BundleSpec` cleanly, the `*bool` / `*time.Duration` choices in `ValidateResolved` correctly preserve nil-vs-set, and the `stringFlagOrConfig` fix (flag default surfacing when both CLI and config are empty) is a real latent bug — good that it's pinned by a regression test. CI is green on tests/lint/security; `tests / E2E` is still in progress at review time, worth confirming before merge.

A few non-blocking items inline:

- **Double resolver work in the deploy-agent path.** `parseValidateAgentConfig` and the Action body both call `resolveValidateTolerations` / `resolveValidateNodeSelector` / `boolFlagOrConfig("no-cleanup", …)`. When CLI overrides config, the new `slog.Info` override lines fire twice. Resolving once and threading through avoids the noise.
- **Snapshot-required error message still references only CLI flags.** Mirror the `--recipe is required (or set spec.validate.input.recipe in --config)` style on the `--no-cluster requires --snapshot` path.
- **Test comment cites `PR-W`** — not greppable; describe the scenario instead.
- **`validPhases` duplicates the CLI parser's vocabulary** — comment acknowledges this; a shared constant would prevent drift.

Out of band, the `stringFlagOrConfig` fix also makes the `if raw == ""` post-check in `pkg/cli/recipe.go` `parseRecipeOutputFormat` dead for `--format`, and the comment there ("still falls through to the config fallback") is now slightly stale. Couldn't leave that as an inline comment because the file isn't in this PR's diff — worth folding into a follow-up cleanup.

Nothing blocks merge.

pkg/cli/root.go
```diff
@@ -447,3 +448,20 @@ func intFlagOrConfig(cmd *cli.Command, flagName string, fallback int) int {
...
+			return fallback
+		}
+		return cmd.Duration(flagName)
+	}
+	v := cmd.Duration(flagName)
+	if fallback != 0 && fallback != v {
+		slog.Info("CLI flag overriding config value", "flag", flagName, "config", fallback, "override", v)
+	}
+	return v
+}
```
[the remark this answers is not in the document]
njhensley: Done in a214f35. `ValidateResolved.Timeout` also became `*time.Duration` so the nil-vs-set signal flows from `Resolve` all the way to `durationFlagOrConfig`. Pinned by `TestValidateResolve_ZeroTimeoutPreserved` in pkg/config/resolve_test.go.

pkg/cli/validate.go
```diff
@@ -54,47 +55,90 @@ type validateAgentConfig struct {
...
+	if cmd.IsSet("node-selector") {
+		ns, err := snapshotter.ParseNodeSelectors(cmd.StringSlice("node-selector"))
+		if err != nil {
+			return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid node-selector", err)
+		}
+		if len(resolved.NodeSelector) > 0 {
+			slog.Info("CLI flag overriding config value", "flag", "node-selector",
+				"config", resolved.NodeSelector, "override", ns)
+		}
+		return ns, nil
```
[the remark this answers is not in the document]
njhensley: Done in a214f35 — `!= nil` so an intentional `nodeSelector: {}` in config also logs the override event.

pkg/config/resolve_test.go
```diff
@@ -545,3 +545,143 @@ func TestBundleResolve_DefensiveCloneOfMaps(t *testing.T) {
...
+	if got1.Tolerations != nil {
+		t.Errorf("nil source → expected nil Tolerations, got %v", got1.Tolerations)
+	}
+	// Tolerations [] → resolved nil (empty input produces no parsed entries).
+	v2 := &config.ValidateSpec{Agent: &config.ValidateAgentSpec{Tolerations: []string{}}}
+	got2, _ := v2.Resolve()
+	// snapshotter.ParseTolerations on empty input may return DefaultTolerations()
+	// or nil — both are acceptable here; what matters is that resolution does
+	// not error. We only assert resolution succeeds.
+	_ = got2
```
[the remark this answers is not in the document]
njhensley: Done in a214f35. Restructured to `if _, err := v2.Resolve(); err != nil { t.Fatalf(...) }` since `got2` wasn't asserted against anyway — same regression guard, slightly tighter.

pkg/config/resolve.go
```diff
@@ -221,6 +223,130 @@ func (b *BundleSpec) Resolve() (*BundleResolved, error) {
...
+				return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+					"invalid spec.validate.agent.tolerations", err)
+			}
+			out.Tolerations = tols
+		}
+	}
+
+	if v.Execution != nil {
+		out.Phases = slices.Clone(v.Execution.Phases)
+		out.NoCluster = v.Execution.NoCluster
```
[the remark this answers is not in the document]
njhensley: Done in a214f35, then deduplicated in 220d3af5: lifted the phase vocabulary out of both call sites into pkg/validator/phases.go (`PhaseNames`, `ParsePhase`, `PhaseAll`), so the config validator (`spec.validate.execution.phases`) and the CLI parser (`--phase`) read from a single source of truth. A new phase added to `PhaseOrder` flows to both automatically.

pkg/cli/validate.go
```diff
@@ -54,47 +55,90 @@ type validateAgentConfig struct {
...
+func resolveValidateTolerations(cmd *cli.Command, resolved *config.ValidateResolved) ([]corev1.Toleration, error) {
+	if cmd.IsSet("toleration") {
+		tols, err := snapshotter.ParseTolerations(cmd.StringSlice("toleration"))
+		if err != nil {
+			return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid toleration", err)
+		}
+		if len(resolved.Tolerations) > 0 {
+			slog.Info("CLI flag overriding config value", "flag", "toleration",
+				"config", resolved.Tolerations, "override", tols)
+		}
```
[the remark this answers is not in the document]
njhensley: Done in 382c140 — `!= nil` to mirror `resolveValidateNodeSelector`. The two checks happen to agree today only because `snapshotter.ParseTolerations` expands an empty input into `DefaultTolerations()`; nil-check keeps the override log correct if that contract ever loosens.

pkg/cli/validate.go
njhensley: Done in 220d3af5. Introduced `validateSharedResolved` (namespace, imagePullSecrets, nodeSelector, tolerations, noCleanup) and resolved once before the snapshot/agent split. `parseValidateAgentConfig` now takes that struct and the agent-deploy path no longer re-invokes the helpers. Each CLI-overrides-config log line fires at most once.

pkg/cli/validate.go
njhensley: Done in 220d3af5 — used your wording verbatim:
`--no-cluster requires --snapshot (or set spec.validate.input.snapshot in --config); cannot deploy the snapshot-capture agent without cluster access`

pkg/cli/config_helpers_test.go
njhensley: Done in 220d3af5 — dropped the `PR-W` citation and replaced it with the concrete scenario (`--namespace` declaring `Value: "aicr-validation"` collapsing to `""` when config didn't set the field, breaking namespace creation). Survives grep now.

pkg/config/resolve.go
njhensley: Done in 220d3af5. `pkg/validator/phases.go` now exports `PhaseNames` (the canonical user-facing vocabulary, typed phases + the `PhaseAll` wildcard) and `ParsePhase(string) (Phase, bool)`. pkg/config consumes `PhaseNames` at init for its resolve-time check; pkg/cli's `parseValidationPhases` consumes `ParsePhase`. No cycle (validator doesn't import config). Adding a phase to `PhaseOrder` flows to both sites automatically.

Also folded in your out-of-band note about the dead `if raw == ""` post-check in `parseRecipeOutputFormat` — removed and the comment refreshed to describe the new three-source precedence.

mchmarny (approved):

PR #856, opened 2026-05-12
feat: os-talos mixin + bundler preManifestFiles support

ayuskauskas:
## Summary

Adds an opt-in `recipes/mixins/os-talos.yaml` mixin that overrides install namespaces to `privileged-<component>` and bundles PSA-privileged Namespace manifests for five components, plus the bundler change that makes those Namespace manifests apply *before* the chart instead of after.

## Motivation / Context

Talos Linux enforces Pod Security Admission (PSA) `restricted` by default. AICR components that need privileged access (host filesystem, host network, root user) get rejected. The Talos collector PR ([#762](https://github.com/NVIDIA/aicr/pull/762)) addressed the snapshot agent's pod-shape; this PR addresses the downstream components.

Fixes part of [#565](https://github.com/NVIDIA/aicr/issues/565).
Related: [#762](https://github.com/NVIDIA/aicr/pull/762) (Talos local-test harness + snapshot agent OS=talos branch).

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `recipes/components/<name>/talos/namespace.yaml` × 5, `recipes/mixins/os-talos.yaml`, `site/.vitepress/config.ts`

### `recipes/mixins/os-talos.yaml` (new)

A new mixin (`kind: RecipeMixin`) that recipe authors opt into via `spec.mixins: [os-talos]`. For each of five components — `gpu-operator`, `network-operator`, `nvsentinel`, `nvidia-dra-driver-gpu`, `nodewright-operator` — it overrides the install namespace to `privileged-<name>` and attaches a per-component `talos/namespace.yaml` via the new `preManifestFiles` field. Constraint `OS.release.ID == talos` guards against accidental non-Talos installs.

Excluded by design (deferred follow-up tracked under [#565](https://github.com/NVIDIA/aicr/issues/565)): cloud-specific drivers (`aws-ebs-csi-driver`, `aws-efa`, `gke-nccl-tcpxo`), `nodewright-customizations` (per-node privileged story not settled), `kube-prometheus-stack` (only `node-exporter` needs privileged; whole-chart move is too coarse).

### Per-component Namespace manifests

Five new files at `recipes/components/<name>/talos/namespace.yaml`. Each carries `pod-security.kubernetes.io/{enforce,audit,warn}=privileged` plus AICR-managed selectors (`app.kubernetes.io/managed-by: aicr`, `app.kubernetes.io/component: <name>`, `aicr.nvidia.com/os: talos`) so fleet-wide audits can identify them.

### Bundler `preManifestFiles` support

Mechanical extension of the existing `ManifestFiles` post-injection path:

- **Schema** (`pkg/recipe/metadata.go`): new `ComponentRef.PreManifestFiles []string` field; additive merge mirrors the existing `ManifestFiles` merge.
- **Manifest collection** (`pkg/bundler/bundler.go`): a phase-parameterized `collectComponentManifestsByPhase` shared by `collectComponentManifests` (post, preserved as the original entry point) and the new `collectComponentPreManifests`. One body, two thin wrappers.
- **Bundler plumbing** (`pkg/bundler/bundler.go` + `pkg/bundler/deployer/{argocd,argocdhelm,helm}/`): each Generator now carries `ComponentPreManifests` and `ComponentPostManifests`. Pre/post error paths are differentiated and route through `errors.PropagateOrWrap`.
- **Writer** (`pkg/bundler/deployer/localformat/writer.go`): new `Options.injectAuxiliaryFolder(idx, c, phase)` helper used by **both** pre and post emission — single `writeLocalHelmFolder` call site for auxiliary folders. Pre-injection fires before `classify(...)` so it works regardless of primary kind (upstream-helm, local-helm, kustomize, or vendored). Folder upper-bound raised from `2*N` to `3*N`. Collision check covers both `<name>-pre` and `<name>-post`. The pre path coexists with `VendorCharts`; the post path is skipped under vendoring (matches the pre-existing "vendored mode collapses mixed into one folder" contract).

### Documentation

- New page `docs/integrator/talos-integration.md` documents what the mixin does, the five privileged namespaces, the PSA label set, apply ordering, and what's out of scope.
- One-sentence cross-link added to `docs/integrator/recipe-development.md` so recipe authors find the new page when they pick `os: talos`.
- Sidebar entry in `site/.vitepress/config.ts` (`make check-docs-sidebar` enforces).

## Testing

```bash
unset GITLAB_TOKEN
go build ./...                                                  # clean
go test -race -count=1 ./pkg/recipe/... ./pkg/bundler/...       # 16 packages, all pass
golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./pkg/bundler/...
make lint-yaml
make check-docs-sidebar
make check-agents-sync
```

Golden coverage for the new code path:

- `pkg/bundler/deployer/helm/testdata/mixed_with_pre/` — `preManifestFiles`-only scenario. Bundle = `001-foo-pre/` (local-helm wrapping Namespace) + `002-foo/` (upstream chart). Locks in lexicographic ordering for the `deploy.sh` `[0-9][0-9][0-9]-*/` glob.
- `pkg/bundler/deployer/helm/testdata/mixed_gpu_operator/` — extended scenario covering pre + primary + post in one bundle (`001-gpu-operator-pre/`, `002-gpu-operator/`, `003-gpu-operator-post/`).
- `pkg/bundler/deployer/argocd/testdata/mixed_with_pre/` — Argo mirror. Sync-waves `0`, `1`, `2` for pre/primary/post; verifies the wave-from-folder-index allocation in `argocd.go`.

Existing 5 helm and 5 argocd-style goldens remain byte-identical: pre-injection is a no-op for any recipe that doesn't set `PreManifestFiles`.

## Risk Assessment

**Rollout notes:** Pure additive. No existing recipe sets `preManifestFiles`, so every shipping bundle's output is byte-identical (verified by the unchanged 10 pre-existing goldens). The mixin is opt-in via leaf-overlay `spec.mixins: [os-talos]`; no leaf overlay is changed in this PR (recipe-author territory). Revert is a single `git revert` of the merge commit.

## Checklist


pkg/bundler/deployer/helm/testdata/mixed_with_pre/001-foo-pre/install.sh
```diff
@@ -0,0 +1,23 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -euo pipefail
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+cd "${SCRIPT_DIR}"
+
+helm upgrade --install foo-pre ./ \
+  --namespace privileged-foo --create-namespace \
+  -f values.yaml -f cluster-values.yaml \
```
mchmarny, line 22: **Blocking — Helm install conflict in real Talos use case.**

This pre-folder runs `helm upgrade --install foo-pre ./ --namespace privileged-foo --create-namespace`, and the chart's `templates/namespace.yaml` declares `Namespace/privileged-foo`. In production, when the mixin sets `ComponentRef.namespace: privileged-gpu-operator` and the pre-manifest creates `Namespace/privileged-gpu-operator`, the two collide:

1. `--create-namespace` creates `privileged-gpu-operator` via the API (no Helm labels).
2. Helm then applies the chart's Namespace template with the same name.
3. Helm 3 rejects with: *"Namespace `privileged-gpu-operator` exists and cannot be imported into the current release: missing `app.kubernetes.io/managed-by=Helm` … missing `meta.helm.sh/release-name` …"*

The five `recipes/components/*/talos/namespace.yaml` files all set `managed-by: aicr` (not `Helm`), and the install template hardcodes `--create-namespace`. So every Talos component will fail at first install via the `helm` deployer.

The argocd deployer path is more tolerant (Argo CD's `CreateNamespace=true` adopts existing namespaces), but Helm CLI does not.

Suggested fixes (pick one):
- In the pre-injection branch in `pkg/bundler/deployer/localformat/writer.go`, emit a pre-specific install.sh that **omits** `--create-namespace` (the chart creates the namespace).
- Or stamp the Helm-ownership annotations onto the rendered pre-manifest Namespace so import succeeds: `meta.helm.sh/release-name=<name>-pre`, `meta.helm.sh/release-namespace=<ns>`, `app.kubernetes.io/managed-by=Helm`.
- Or install pre-folder releases into `default`/`kube-system` (cluster-scoped Namespaces don't require their release to live in them).

The golden test only verifies file contents, not actual install behavior — please add an integration check (KWOK or real Helm dry-run with `--create-namespace`) that exercises this path end-to-end.

pkg/bundler/deployer/localformat/writer.go
```diff
@@ -190,11 +268,26 @@ func Write(ctx context.Context, opts Options) (WriteResult, error) {
...
+		// and regardless of primary kind (upstream-helm, local-helm, or
+		// kustomize).
+		if pf, err := opts.injectAuxiliaryFolder(idx, c, phasePre); err != nil {
+			return WriteResult{}, err
+		} else if pf != nil {
+			folders = append(folders, *pf)
+			slog.Info("wrote local chart folder",
+				"index", idx, "dir", pf.Dir,
+				"kind", KindLocalHelm.String(), "parent", c.Name)
+			idx++
```
mchmarny, line 327: Pre-injection runs unconditionally — including in the vendored-charts path that follows at L296. The comment ("we still budget for 3x because pre folders are independent of the primary path") asserts pre + vendor is intentional, but I don't see a test in `pkg/bundler/deployer/localformat/vendor_write_test.go` exercising `ComponentPreManifests` together with `VendorCharts: true`. Please add one — pre is the marquee feature, and the vendored air-gap path is exactly where this combo gets used.

recipes/mixins/os-talos.yaml
```diff
@@ -0,0 +1,56 @@
...
+metadata:
+  name: os-talos
+spec:
+  constraints:
+    - name: OS.release.ID
+      value: talos
+
+  componentRefs:
+    - name: gpu-operator
+      namespace: privileged-gpu-operator
```
mchmarny, line 34: The mixin sets `namespace: privileged-gpu-operator` AND points `preManifestFiles` at a file whose `metadata.name: privileged-gpu-operator`. Two unrelated strings, no validation that they agree. If a future edit changes one but not the other, the chart will install into one namespace while a different orphan namespace gets created — silent misconfiguration.

Consider adding a recipe-resolution check that verifies, for each component with `preManifestFiles`, the rendered Namespace's `metadata.name` matches `ComponentRef.namespace`. Or template the namespace name with a Helm value so they can't drift.

recipes/mixins/os-talos.yaml
mchmarny, line 1: No end-to-end test exercises this mixin's resolution. The `pkg/recipe/metadata_test.go` cases prove `PreManifestFiles` merges additively at the field level, but nothing proves "a leaf overlay with `spec.mixins: [os-talos]` resolves to ComponentRefs carrying the right namespace + preManifestFiles for all five components". A golden test under `pkg/recipe/testdata/` that resolves a synthetic Talos leaf overlay would catch future regressions cheaply.

pkg/bundler/deployer/argocd/testdata/mixed_with_pre/001-gpu-operator-pre/application.yaml
```diff
@@ -0,0 +1,28 @@
...
+    automated:
+      prune: true
+      selfHeal: true
+    syncOptions:
+      - CreateNamespace=true
+      # ServerSideApply: required for charts whose CRDs exceed kubectl's
+      # 262144-byte client-side annotation cap (e.g., kube-prometheus-stack,
+      # kai-scheduler). Without this, Argo's sync retries forever with
+      # "metadata.annotations: Too long". server-side apply uses the
+      # apiserver's structured-merge path and has no such cap.
```
mchmarny: nit: the `ServerSideApply=true` rationale in the comment ("CRDs exceed kubectl's 262144-byte client-side annotation cap") doesn't apply to a pre-folder that only carries a Namespace. Harmless but the comment is now misleading in this context — consider tightening the emission so the comment only attaches to applications that actually need it, or generalize the comment.

pkg/bundler/deployer/localformat/writer.go
```diff
@@ -106,6 +118,58 @@ func renderInputFor(c Component) manifest.RenderInput {
...
+// post-phase manifests from Options. Pre folders apply before the
+// primary chart (sync-wave N-1 in Argo CD / install step N-1 in Helm),
+// post folders apply after (sync-wave N+1 / step N+1). Both phases
+// share a single emission path so any future change to wrapped-chart
+// shape lands in one place.
+//
+// Typed as string (not int-iota like the sibling manifestPhase in
+// pkg/bundler/bundler.go) so the value embeds directly into the
+// "<name>-<phase>" folder/release name and into %q error messages
+// without a separate String() method.
```
mchmarny, line 136: nit: `injectionPhase` is a `string` here, while the sibling `manifestPhase` in `pkg/bundler/bundler.go` is an `int` iota. The docstring explains why (string embeds into folder names), and the divergence is fine — but consider renaming both to `phase` consistently or keeping a single shared type in a small internal package so the two never drift in meaning.

pkg/recipe/metadata.go
```diff
@@ -103,6 +103,15 @@ type ComponentRef struct {
...
 
+	// PreManifestFiles lists manifest files that must be bundled and applied
+	// BEFORE the component's primary chart. Paths are relative to the data
+	// directory. Used for resources the chart depends on (e.g. a Namespace
+	// with PSS labels that the chart's pods need to land in). Bundler emits
+	// these as a "<name>-pre" local-helm folder at sync-wave N-1 (Argo) or
+	// install step N-1 (Helm); the primary chart lands at wave N; existing
+	// ManifestFiles still land at wave N+1 as before.
+	PreManifestFiles []string `json:"preManifestFiles,omitempty" yaml:"preManifestFiles,omitempty"`
+
```
mchmarny, line 118: The doc says paths are "relative to the data directory" but doesn't constrain traversal. `recipe.GetManifestContent` is the trust boundary here — if it doesn't reject `../` already, a hostile mixin could read out of bounds. Same shape as `ManifestFiles` so no regression, but worth a sentence noting the constraint (and ideally a regression test for `..` rejection).

mchmarny (changes requested):
Nicely structured refactor — the phase-parameterized collector and the shared `injectAuxiliaryFolder` helper keep the writer footprint small, and the goldens lock the lexicographic install order. Mixin authoring is clean, and the docs page is well-scoped.

One blocking issue: the helm-deployer pre-folder pattern (`helm install … --namespace privileged-<x> --create-namespace` against a chart whose only template is `Namespace/privileged-<x>`) is a known Helm 3 install conflict — Helm rejects the rendered Namespace as "already exists, missing managed-by=Helm". Every Talos component will hit this on first install via the `helm` deployer. Argo CD path may tolerate it; Helm path will not. See the comment on `001-foo-pre/install.sh` for fix options.

Secondary asks: an e2e test that resolves the os-talos mixin end-to-end, a vendor + pre combination test, and stronger coupling between `ComponentRef.namespace` and the pre-manifest's `Namespace.metadata.name`.

CI is green and the goldens cover what they claim — the gap is install-time, not file-shape. Marking as request-changes because the production scenario is the one the goldens don't exercise.

pkg/bundler/deployer/argocd/testdata/mixed_with_pre/001-gpu-operator-pre/templates/talos-namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: privileged-gpu-operator
+  labels:
+    pod-security.kubernetes.io/enforce: privileged
```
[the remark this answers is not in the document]
ayuskauskas: Is this really critical? The test data is sufficient to verify that a special namespace is being created without fully replicating the component. Keeping those in perfect sync would be a larger burden over time.

mchmarny (approved):

Issue #698, opened 2026-04-27
chore(recipes): check and update runtime component versions across all recipes


yuanchen8911:
## Summary

Audit the pinned versions of every runtime component referenced by AICR recipes and update them to current stable upstream releases where appropriate.

## Motivation

Component versions in `recipes/registry.yaml`, `recipes/overlays/*.yaml`, and `recipes/components/*/values.yaml` drift over time as upstream projects ship security fixes, bug fixes, and Kubernetes compatibility updates. Without a periodic refresh, AICR-generated bundles deploy stale versions that may miss CVE fixes or fail against newer Kubernetes versions.

## Scope

Check and (where appropriate) update versions for every component declared in `recipes/registry.yaml`, including (non-exhaustive list — registry is authoritative):

- GPU stack: `gpu-operator`, `nvidia-device-plugin`, `dcgm-exporter`
- Schedulers / CRDs: `kai-scheduler`, `kueue`, `volcano`
- Inference: `dynamo-platform`, `kgateway`
- Training: `kubeflow-trainer`
- Observability: `kube-prometheus-stack`, `prometheus-adapter`
- Platform: `nfd`, `cert-manager`, `node-feature-discovery`
- Any other components in `recipes/registry.yaml` not listed here

Also check overlay-level pins in `recipes/overlays/*.yaml` and any version constraints embedded in mixins (`recipes/mixins/*.yaml`).

## Definition of done

- [ ] For each component, current pinned version is compared against latest stable upstream release.
- [ ] Where a newer version is selected, the version pin is updated and the change is justified (CVE fix, K8s compatibility, etc.) in the PR description.
- [ ] Where a version is intentionally held back, the reason is documented (in `recipes/registry.yaml` comments or a short note in `docs/contributor/data.md`).
- [ ] `make qualify` passes locally on the updated recipes.
- [ ] KWOK simulated cluster tests pass (`make kwok-test-all`).
- [ ] At least one full GPU CI run on H100 nvkind passes against the updated bundle.
- [ ] Documentation under `docs/user/component-catalog.md` is regenerated/updated if any displayName, repository, or default chart changes.

## Out of scope

- Adding new components (handled in separate issues).
- Upgrading recipes to a different Kubernetes minor (separate compatibility effort).
- Removing components.

## Notes

- Compatibility constraint: AICR targets Kubernetes 1.33+. Verify each upgraded component supports this floor.
[...]

PR #866, opened 2026-05-13
Add Slinky slurm-operator as platform-slurm

faganihajizada:
## Summary

Adds the Slinky slurm-operator (chart v1.1.0) as a first-class AICR platform under `--platform slurm`. Empty-base PR — ships the operator + CRDs plumbing only; leaf overlays follow in a separate PR.

## Motivation / Context

Slurm is the dominant HPC workload manager. Users running mixed AI + HPC workloads on Kubernetes need a validated path to install the [Slinky operator](https://github.com/SlinkyProject/slurm-operator) alongside the existing GPU stack (`gpu-operator`, `nvsentinel`, `nfd`, `cert-manager`, etc.) without bringing it in by hand. This PR adds the platform plumbing so a follow-up can add per-CSP / per-accelerator leaf overlays without re-litigating the operator/CRD shape, namespace, version pin, or value-override aliases.

Mirrors the operator-only pattern of `--platform dynamo` and `--platform kubeflow`: AICR ships only the platform; cluster-instance CRs (`Controller`, `LoginSet`, `NodeSet`, `Accounting`, `RestApi`, `Token`) remain user-authored. The Slinky `slurm` cluster Helm chart is intentionally deferred to a follow-up PR with its own short alias (e.g. `slurm-cluster`) so the `slurm` alias continues to route to the operator.

Fixes: N/A
Related: [SlinkyProject/slurm-operator#187](https://github.com/SlinkyProject/slurm-operator/pull/187) (upstream `nodeSelector` exposure — see Implementation Notes)

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] CLI (`cmd/aicr`, `pkg/cli`)
- [ ] API server (`cmd/aicrd`, `pkg/api`, `pkg/server`)
- [x] Recipe engine / data (`pkg/recipe`)
- [ ] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [ ] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [ ] Validator (`pkg/validator`)
- [ ] Core libraries (`pkg/errors`, `pkg/k8s`)
- [x] Docs/examples (`docs/`, `examples/`)
[...]

PR #872, opened 2026-05-13
chore(recipes): bump aws-efa to v0.5.26 keeping AICR's hardened securityContext

yuanchen8911:
## Summary

Bumps the EFA device-plugin chart from `v0.5.3` (2023) → `v0.5.26` / app `v0.5.18` (current upstream latest). AICR keeps its existing hardened `securityContext` override rather than accepting the chart's new upstream-default of `privileged: true`.

## Motivation / Context

The audit in #698 flagged aws-efa as blocked on a security review because chart `v0.5.4+` switched the device-plugin DaemonSet's default to `privileged: true` + `runAsUser: 0`, conflicting with AICR's narrow override (`allowPrivilegeEscalation: false`, `capabilities.drop: [ALL]`, `runAsNonRoot: false`).

Live-cluster testing on EKS `aicr2` (see Testing below) shows the narrow posture continues to work on AICR's primary target — the device-plugin binary doesn't actually require elevated privileges on Ubuntu 24.04 / kernel 6.14 / p5. This unlocks a third path beyond "accept `privileged: true`" or "stay on v0.5.3": **bump the chart, keep the narrow override**, get 23 patches of upstream fixes without expanding the security blast radius.

Fixes: follow-up #4 from #698 (aws-efa `v0.5.3 → v0.5.26`)
Related: #698, #715

## Type of Change

- [x] Bug fix / enhancement (component version refresh)
- [x] Build/CI/tooling (regenerated BOM doc)

## Component(s) Affected

- [x] Recipe engine / data (`recipes/registry.yaml`, `recipes/components/aws-efa/values.yaml`)
- [x] Docs (`docs/user/container-images.md` regenerated via `make bom-docs`)

## Implementation Notes

**Three changes only:**
1. `recipes/registry.yaml` — `defaultVersion: v0.5.3` → `v0.5.26`
2. `recipes/components/aws-efa/values.yaml` — `image.tag: v0.5.3` → `v0.5.18` (chart's bundled appVersion at v0.5.26), plus a long-form comment documenting the intentional divergence from upstream's `privileged: true` default
3. `docs/user/container-images.md` — regenerated via `make bom-docs`

**No change to the securityContext block itself** — AICR's existing override is preserved verbatim:
```yaml
securityContext:
  allowPrivilegeEscalation: false
  capabilities:
    drop:
      - ALL
  runAsNonRoot: false
```

**Why upstream went privileged.** Starting in chart `v0.5.4` AWS switched the daemonset to `privileged: true` + `runAsUser: 0` to cover the full matrix of EFA driver / kernel / OS combinations they support. Their default optimizes for "works everywhere" at the cost of blast radius. AICR optimizes for the narrower OS surface we actually ship, so we can keep narrow caps and still get the fixes.

**Revisit triggers** documented inline in the values file:
- Plugin logs return `operation not permitted` on a future EFA driver version
- AICR adds an AL2023 / Bottlerocket / GB200 GPU recipe (the test below only covers Ubuntu)

## Testing

```bash
make qualify    # all Go unit tests + 20/20 chainsaw + lint + scan: green
make bom-docs   # docs/user/container-images.md regenerated cleanly
```

**Live cluster regression test on EKS `aicr2` (account 615299774277, us-east-1, EKS 1.35):**

Baseline (chart `v0.5.3`):
- p5.48xlarge GPU node (`ip-10-0-185-161.ec2.internal`), Ubuntu 24.04.4, kernel `6.14.0-1018-aws`
- DaemonSet `aws-efa-k8s-device-plugin` 1/1 Running for 13d under AICR's narrow `securityContext`
- Node advertising `vpc.amazonaws.com/efa: 32` in both Capacity and Allocatable

Test procedure:
```bash
helm upgrade aws-efa eks/aws-efa-k8s-device-plugin \
  --version v0.5.26 --reuse-values --set image.tag=v0.5.18 -n kube-system
```

Result:
- New pod came up 1/1 Running in ~30s
- `kubectl get pod -o jsonpath='{.spec.containers[0].securityContext}'` confirmed AICR's narrow override preserved: `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"runAsNonRoot":false}` (no `privileged: true`)
- Pod logs showed clean EFA enumeration without `operation not permitted`:
  ```
  EfaDeviceTopology initialized successfully with 8 GPUs and 32 EFAs
  GPU topology initialized successfully.
  Starting to serve on /var/lib/kubelet/device-plugins/aws-efa-device-plugin.sock
  Registered device plugin with Kubelet
  ```
- Node continued advertising `vpc.amazonaws.com/efa: 32` in both Capacity and Allocatable

Rolled back via `helm rollback aws-efa 1` to restore v0.5.3 — cluster returned to baseline state.

**Test scope caveats (also documented in the values file comment):**
- Validated only on Ubuntu 24.04 / kernel 6.14 / p5. AICR ships no `*-eks-amazonlinux-*` or Bottlerocket+EFA GPU recipe today, so a broader OS matrix wasn't tested.
- Test exercised plugin enumeration + Kubelet registration. Actual workload EFA usage (a pod requesting `vpc.amazonaws.com/efa: N` and doing RDMA traffic) was not validated in this PR — that path uses the *workload's* securityContext, not the plugin's, so it's out of scope for the chart bump.
- If/when an AL2023 EKS GPU recipe is added, that PR's review should re-validate aws-efa's narrow-cap posture on AL2023 and either confirm the override holds or accept upstream's `privileged: true` default at that time.

## Risk Assessment

- [x] **Low** — Single component bump, narrow blast radius. Live regression test on the actual deployment target with a documented rollback path. Existing AICR EKS bundles regenerated against this version will run the same securityContext as today, just with a newer binary that picks up 15 app-version increments of upstream fixes.

**Rollout notes:** Existing AICR-deployed EKS clusters running `aws-efa` v0.5.3 can helm-upgrade in place by regenerating the bundle and running `deploy.sh`. The DaemonSet rotates one pod per EFA-capable node (~30s downtime per node for that node's EFA resource advertisement; running pods that already hold EFA allocations are unaffected).

## Checklist

- [x] I added/updated tests for new functionality (n/a — version bump only)
- [x] I updated docs if user-facing behavior changed (`docs/user/container-images.md` regenerated; values-file comment added)


recipes/registry.yaml
```diff
@@ -142,7 +142,7 @@ components:
     helm:
       defaultRepository: https://aws.github.io/eks-charts
       defaultChart: aws-efa-k8s-device-plugin
-      defaultVersion: v0.5.3
+      defaultVersion: v0.5.26
```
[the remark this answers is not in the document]
yuanchen8911: This finding is incorrect — `v0.5.26` is the current latest published version in the upstream `eks/aws-efa-k8s-device-plugin` Helm repo, not v0.5.25.

Direct verification just now (2026-05-13):

```
$ helm repo update eks
$ helm search repo eks/aws-efa-k8s-device-plugin --versions | head -3
NAME                         	CHART VERSION	APP VERSION	DESCRIPTION
eks/aws-efa-k8s-device-plugin	v0.5.26      	v0.5.18    	A Helm chart for EFA device plugin.
eks/aws-efa-k8s-device-plugin	v0.5.25      	v0.5.17    	A Helm chart for EFA device plugin.
eks/aws-efa-k8s-device-plugin	v0.5.24      	v0.5.16    	A Helm chart for EFA device plugin.

$ helm pull eks/aws-efa-k8s-device-plugin --version v0.5.26
# (succeeds, chart artifact downloaded)
```

I also already deployed v0.5.26 end-to-end against the `aicr2` EKS cluster as the live regression test described in the PR description — `helm upgrade aws-efa eks/aws-efa-k8s-device-plugin --version v0.5.26` returned `Release "aws-efa" has been upgraded` and the DaemonSet pod came up 1/1 Running. The chart pulls and renders cleanly.

CodeRabbit's web search appears to have hit a stale ArtifactHub index (which lists v0.5.25 as latest) and missed the more recent v0.5.26 published to the actual chart repo at `aws.github.io/eks-charts`. The authoritative source is the chart repo, not the ArtifactHub mirror.

Not actionable — leaving `v0.5.26` in place.

recipes/components/aws-efa/values.yaml
```diff
@@ -39,14 +39,37 @@
...
+# defaults — so any field omitted here would inherit the chart's
+# `privileged: true` / `runAsUser: 0`. We must therefore set
+# `privileged: false` and `runAsUser: null` explicitly, not just by
+# omission.
+#
+# Validated against chart v0.5.26 + app v0.5.18 on EKS 1.35 — see PR
+# description for evidence. Revisit this override if a future EFA driver
+# version returns `operation not permitted` from the plugin logs, or if
+# AICR adds an AL2023 / Bottlerocket / GB200 GPU recipe without re-
+# validating this posture there.
```
mchmarny: Strong divergence rationale — the comment explains *why* (upstream's "works everywhere" optimization vs AICR's narrow OS surface), the Helm merge mechanic (so the next maintainer doesn't strip the explicit `false`/`null` thinking it's redundant), and the revisit triggers. This is the right shape for documenting an intentional drift from upstream.

recipes/components/aws-efa/values.yaml
```diff
@@ -39,14 +39,37 @@
...
 securityContext:
   allowPrivilegeEscalation: false
+  privileged: false
   capabilities:
     drop:
       - ALL
   runAsNonRoot: false
+  runAsUser: null
```
mchmarny, line 144: **Test methodology gap — the live validation didn't actually exercise these new keys.** The PR description's test used:

```
helm upgrade aws-efa eks/aws-efa-k8s-device-plugin \
  --version v0.5.26 --reuse-values --set image.tag=v0.5.18 -n kube-system
```

`--reuse-values` reuses the *previously merged* user values from the v0.5.3 release and does *not* re-merge with the new chart's defaults. So what was validated is: "v0.5.3-shape values (without `privileged: false` and `runAsUser: null`) running against the v0.5.26 binary." The new explicit keys added in this PR weren't on the cluster during the test.

The path AICR users actually take is `aicr bundle` → fresh `helm install` of v0.5.26 with these new values, where Helm *does* merge values-file against chart defaults. That's where the explicit `privileged: false` and `runAsUser: null` matter — and that's what hasn't been validated end-to-end.

Suggest re-running with either:
- `helm install` clean against v0.5.26 with the values file from this PR, or
- `helm upgrade --reset-then-reuse-values --version v0.5.26 -f recipes/components/aws-efa/values.yaml ...`

then confirming the rendered pod's `securityContext` still shows no `privileged: true` and no `runAsUser: 0`. The chart-merge claim in lines 52-58 is precise; the live cluster just hasn't seen it in action.

mchmarny, line 144: `runAsUser: null` in a values file merges as the literal value `null`, not as "key absent." The chart's `toYaml .Values.securityContext` will emit `runAsUser: null` in the rendered pod spec, and Kubernetes' API server normalizes that to "unset" → image-default user.

Worth confirming on cluster that `kubectl get pod -o jsonpath='{.spec.containers[0].securityContext.runAsUser}'` returns empty (not `0`, not `null` literal). If a future K8s version stops normalizing the literal null, behavior changes silently. Cheap to assert as part of `aws-efa` health check.

Related: since `runAsNonRoot: false` is set, the image is permitted to run as whatever USER its Dockerfile declares. Worth a sentence in the comment block clarifying that the override allows root-via-image-default but blocks `runAsUser: 0` being forced by chart defaults — they're different things.

docs/user/container-images.md
```diff
@@ -190,6 +192,15 @@ _No images extracted._
...
 
+### slinky-slurm-operator
+
+- `ghcr.io/slinkyproject/slurm-operator-webhook:1.1.0`
+- `ghcr.io/slinkyproject/slurm-operator:1.1.0`
+
+### slinky-slurm-operator-crds
+
+_No images extracted._
+
```
mchmarny, line 203: nit: BOM regen picked up unrelated drift from main — `slinky-slurm-operator` / `slinky-slurm-operator-crds` (#866), `kubeflow-trainer` pytorch 2.9.1 → 2.11.0, `network-operator` busybox 1.36 → 1.37. Those aren't this PR's changes.

Either rebase onto main before merge so the BOM regen reflects only the aws-efa delta, or call this out in the PR description so reviewers don't have to puzzle out where those bumps came from. Not blocking — the rendered BOM does have to match registry state — but it makes the diff noisier than it needs to be.

mchmarny (commented):
Right call on the divergence — keeping the narrow `securityContext` is the correct posture for AICR's EKS surface, and the values-file comment is one of the better divergence rationales I've seen. CI is fully green (all Tier 2 e2e + KWOK + grype + ClamAV).

One real concern: the live cluster test used `helm upgrade --reuse-values`, which bypasses chart-default merging — so the new explicit `privileged: false` and `runAsUser: null` keys this PR adds weren't actually exercised on cluster. Worth a fresh-install validation before merge since that's the path AICR users actually take. Other notes are minor (BOM doc carries unrelated drift; `runAsUser: null` literal handling worth asserting in health check).

recipes/components/aws-efa/values.yaml
yuanchen8911: Good catch — `--reuse-values` was the wrong instrument. After Codex's P1 review ("chart-default `privileged: true` survives merge unless explicitly overridden"), I re-tested with a fresh `-f values.yaml` install that actually exercises the new keys, but I never updated the PR description with the re-test transcript. Posting it here for the record.

**Re-test on aicr2 (EKS 1.35, p5.48xlarge, Ubuntu 24.04, kernel 6.14):**

```bash
helm upgrade aws-efa eks/aws-efa-k8s-device-plugin --version v0.5.26 \
  -f recipes/components/aws-efa/values.yaml \
  --set 'nodeSelector.nodeGroup=gpu-worker' \
  --set 'tolerations[0].key=dedicated,tolerations[0].operator=Equal,tolerations[0].value=worker-workload,tolerations[0].effect=NoSchedule' \
  --set 'tolerations[1].key=dedicated,tolerations[1].operator=Equal,tolerations[1].value=worker-workload,tolerations[1].effect=NoExecute' \
  -n kube-system
```

(No `--reuse-values` — passes the committed values file directly, simulating what `aicr bundle` produces.)

**DaemonSet rendered `securityContext`** (`kubectl get ds aws-efa-k8s-device-plugin -o jsonpath='{.spec.template.spec.containers[0].securityContext}'`):

```json
{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"privileged":false,"runAsNonRoot":false}
```

`privileged: false` is **explicit in the rendered spec** — the chart default of `true` got overridden. `runAsUser` is absent entirely — answers your next thread.

**Pod status:** 1/1 Running in ~23s under the explicit narrow posture.

**Plugin logs (clean enumeration, no `operation not permitted`):**
```
EfaDeviceTopology initialized successfully with 8 GPUs and 32 EFAs
GPU topology initialized successfully.
Starting to serve on /var/lib/kubelet/device-plugins/aws-efa-device-plugin.sock
Registered device plugin with Kubelet
```

**Node still advertising EFA:** `vpc.amazonaws.com/efa: 32` in both Capacity and Allocatable, unchanged from baseline.

Rolled back to v0.5.3 after the re-test. So the explicit-override path is validated end-to-end on cluster, not just via `helm template`.

yuanchen8911: Confirmed: Kubernetes normalizes `runAsUser: null` to "unset." Evidence from the same re-test I posted in the sibling thread — the rendered DaemonSet `securityContext` came out as:

```json
{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"privileged":false,"runAsNonRoot":false}
```

No `runAsUser` field at all. The API server dropped the explicit `null` during the strategic-merge-patch admission step, which is what we want — the image's own `USER` directive applies. So both the chart-default of `runAsUser: 0` *and* our explicit `null` produce the same effective behavior at admission, but the explicit `null` correctly *removes* the chart default during the values-merge layer (which happens before admission), and admission then drops the null itself.

Pod started 1/1 Running, EFA enumerated cleanly, node advertised `vpc.amazonaws.com/efa: 32` — so whatever UID the image actually runs as has enough access to open the InfiniBand device nodes. (The aws-efa image still uses root in v0.5.18; the comment in the values file notes we'd inherit any future image hardening upstream chooses to apply.)

docs/user/container-images.md
yuanchen8911: The rebase did *not* fully eliminate the drift. `git diff origin/main..HEAD -- docs/user/container-images.md` after rebase still shows:

```
- aws-efa v0.5.3 → v0.5.26  (this PR's intent)
+ slinky-slurm-operator + slinky-slurm-operator-crds BOM rows  (catch-up; #866 merged without regenerating BOM)
- pytorch 2.9.1 → 2.11.0  (upstream kubeflow-trainer chart drift)
- busybox 1.36 → 1.37  (upstream chart drift)
```

Root cause: `make bom-docs` regenerates by rendering each chart fresh against the live OCI sources, so it picks up upstream image bumps in addition to anything in `registry.yaml`. Main's BOM appears to be genuinely stale — \#866 added the slinky-slurm registry entries but the committed BOM doesn't reflect them, and two upstream chart contents have shifted since main was last regenerated.

The catch  here: I can't simply revert the non-aws-efa BOM rows on this branch, because `make qualify` enforces BOM-up-to-dateness against the current registry + chart contents. Reverting the catch-up would make this PR's `make qualify` fail.

Filed a clean up PR: #879

yuanchen8911: Going with option (b). Filed the catch-up as #878 — a tiny pure-regen PR (one file, just `docs/user/container-images.md`) so the three drift items (slinky-slurm BOM rows, pytorch image bump, busybox bump) land independently of this PR.

Once #878 merges, I'll rebase #872 onto the new main and the BOM diff here will be just the aws-efa rows.

yuanchen8911:
Heads up on the BOM diff Mark flagged: I've split the catch-up out into #878 (pure `make bom-docs` regen — slinky-slurm BOM rows that #866 missed, plus upstream pytorch + busybox image drift). Will rebase this PR after #878 merges; the BOM diff here will then be just aws-efa rows. Both PRs CI green.

mchmarny (approved):

Issue #859, opened 2026-05-12
bundler: `manifestFiles` always emitted as `-post`, breaking prereq ConfigMaps (GB200/EKS deploy hangs)


njhensley:
## Summary

The GB200/EKS deploy hangs in `deploy.sh` at `009-gpu-operator` because the chart's `--wait` blocks on `ClusterPolicy` reaching Ready, but `ClusterPolicy` reconciliation requires the `nvidia-kernel-module-params` ConfigMap that ships in the recipe's `manifestFiles`. #706 (cf3cd339) moved `manifestFiles` from a pre-install pass to a `-post` folder applied after the chart, so the ConfigMap never lands before reconcile.

## Reproduction

Clean GB200/EKS cluster:

```bash
aicr recipe --service eks --accelerator gb200 --os ubuntu --intent training -o recipe.yaml
aicr bundle --recipe recipe.yaml -o ./bundles
cd ./bundles && ./deploy.sh
```

`009-gpu-operator` retries 5×10m and fails with:

```
"error":"ERROR: could not get ConfigMap nvidia-kernel-module-params from client:
ConfigMap \"nvidia-kernel-module-params\" not found"
```

## Root cause

`pkg/bundler/deployer/localformat/doc.go:51-64` classifies any `manifestFiles` entry as **post-install** (intended for CR-shaped manifests that need the chart's CRDs to exist first). But `nvidia-kernel-module-params` is **prereq-shaped** — the chart's controller reads it *during* reconcile, so it must apply before `helm install --wait`. There's currently no way to express that direction.

## Scope

`manifestFiles` is used by multiple recipes; some entries are post-shaped, some are prereq-shaped. Known prereqs that break the same way:

- `gb200-eks-training.yaml`, `gb200-eks-inference.yaml` → `gpu-operator/manifests/kernel-module-params.yaml`

Other `manifestFiles` callers (`h100-eks-*`, `h100-gke-cos-*`, `mixins/platform-inference.yaml`, `aks.yaml`, `gke-cos.yaml`, `base.yaml`) need audit; any that the chart's controller reads at reconcile time has the same latent deadlock.

## Proposed fix

Add a direction tag to `manifestFiles`:

```yaml
manifestFiles:
  - path: components/gpu-operator/manifests/kernel-module-params.yaml
    phase: pre   # default remains post
```

[...]

PR #873, opened 2026-05-13
feat(evidence): emit Recipe Evidence v1 bundle from aicr validate

### LINKED ISSUE #754: Define evidence bundle format and emit via aicr validate --emit-evidence

njhensley:
## Summary

Adds Recipe Evidence Bundle v1 emission (ADR-007) to `aicr validate`: signed, in-toto / Sigstore-bundled, cosign-verifiable attestations of a validate run, optionally pushed to an OCI registry as an artifact + Referrers-attached signature.

## Motivation / Context

Recipe-test evidence (ADR-007) is the missing piece between "this recipe was advertised as working on hardware X" and "someone signed off on actually proving it." This PR delivers the emit half: a contributor (or CI) running `aicr validate` against a real cluster can produce a tamper-evident artifact binding the resolved recipe, the captured snapshot, per-phase CTRF reports, a CycloneDX BOM, and a fingerprint of the cluster, signed by Sigstore keyless OIDC and discoverable on the registry via the OCI Referrers API.

Fixes: #754
Related: #859 (filed during testing — separate bundler deadlock surfaced on clean GB200/EKS clusters, not addressed here)

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Bundlers (`pkg/bundler`, `pkg/component/*`) — only `pkg/bundler/attestation`, reused as a primitive
- [x] Other: `pkg/evidence/attestation` (new), `pkg/config`, `pkg/oci`

### Surface

New flags on `aicr validate`:

| Flag | Required with | Behavior |
|---|---|---|
| `--emit-attestation <dir>` | — | Always emits the bundle directory locally (manifest, statement, recipe, snapshot, BOM, ctrf/*). Safe to run offline. |
| `--bom <path>` | `--emit-attestation` (optional) | When omitted, the bundle synthesizes a CycloneDX BOM from the recipe's component refs + cluster-observed images + the validator catalog images that ran. Pass an explicit path for an exhaustive BOM (e.g. produced by `make bom`). |
| `--push <oci-ref>` | `--emit-attestation` | Pushes the bundle as an OCI artifact, signs the manifest via Sigstore keyless, attaches the Sigstore Bundle as an OCI Referrer of the artifact. |
| `--identity-token`, `--oidc-device-flow`, `--plain-http`, `--insecure-tls` | `--push` | Mirror the existing `aicr bundle --attest` knobs. OIDC chain: `--identity-token` → `COSIGN_IDENTITY_TOKEN` env → ambient GH Actions OIDC → device flow → interactive browser. |

A new `spec.validate.evidence` block in AICRConfig (`pkg/config`) wires the same surface through `--config`.

### Bundle layout

```
<dir>/
├── pointer.yaml                       # Slim locator — safe to commit to recipes/evidence/<recipe>.yaml
└── summary-bundle/
    ├── manifest.json                  # Per-file sha256 inventory
    ├── statement.intoto.json          # Unsigned in-toto Statement (subject = recipe digest, portable identity)
    ├── recipe.yaml                    # Canonicalized recipe
    ├── snapshot.yaml                  # Validate-time snapshot
    ├── bom.cdx.json                   # CycloneDX 1.6
    └── ctrf/                          # Per-phase CTRF reports
        ├── deployment.json
        ├── performance.json
        └── conformance.json
```

When `--push` is set, the Sigstore Bundle v0.3 is attached as a **separate OCI artifact** whose manifest references the main artifact's digest via the OCI `subject` field, so `cosign verify-attestation <ref>` discovers it via `/v2/<name>/referrers/<digest>`. The signed Statement's `subject[0]` is the OCI artifact digest (anchoring cosign verification to the artifact ref); the recipe digest is preserved in the predicate body (`predicate.recipe.{name,digest}`) for portable provenance even if the artifact is later repackaged.

### Notable design choices

- **Sign sequence:** the OCI artifact is packed and pushed before the OIDC token is fetched. This was a fix during testing — Fulcio binds the token to a nonce at issue time, and a multi-minute validate-then-sign gap invalidated it ("error processing the identity token"). The token is now resolved adjacent to the SignBundle call.
- **Shared OCI push path no longer fast-paths the no-subdir case.** `pkg/oci/push.go` previously returned `sourceDir` directly when `subDir == ""`; that shortcut let the oras file store leak annotation-named manifest blobs back into the caller's source tree (regression test in `pkg/oci/push_test.go`). Every push — both the new evidence flow and the existing `pkg/cli/bundle.go` `PackageAndPush` — now copies through a tempdir via hardlink. Hardlink walk cost is negligible on typical bundles; reviewers touching the bundle path should note the behavior shift.
- **OCI Referrers attach:** the signature is a discoverable second artifact, not a layer inside the main one. Older cosign storage layouts (`<image>.sig` sibling tag) are not produced — Referrers covers all modern registries (ttl.sh, ghcr.io, ECR, Harbor, GHCR).
- **No `aicr verify-evidence` yet:** verification is via standard `cosign verify-attestation`. The dedicated wrapper command (#753) is intentional out-of-scope follow-up.
- **`pointer.yaml` is local-only metadata, deliberately unsigned.** Every claim in it (artifact digest, signer identity, Rekor log index) is independently verifiable from the bundle it references. Contributors commit it to `recipes/evidence/<recipe>.yaml` to publish a reviewable reference.
- **OIDC resolution reuses the existing `pkg/bundler/attestation` primitive.** No new sign/verify code paths; `aicr validate --push` and `aicr bundle --attest` share the same precedence chain and the same Fulcio/Rekor URLs.

### Out of scope (intentional)

- `aicr verify-evidence` CLI — #753 follow-up.
- `--include-logs` / `--push-logs` — flags were prototyped, then dropped pending validator log-capture rework. The manifest already pre-commits log file digests when logs are produced; the directory emission can be wired later without schema change.
- Per-validator `stdout` redaction — CTRF stdout currently carries cluster-specific data (hostnames, pod UUIDs, GPU UUIDs). Worth scrubbing in a follow-up; not blocking for the audit + sign + verify flow.

## Testing

End-to-end run on a real GB200/EKS cluster (6 nodes, 2 GB200, EKS 1.34.7, ubuntu 24.04):

```bash
unset GITLAB_TOKEN && make build
AICR_VALIDATOR_IMAGE_TAG=latest \
  KUBECONFIG=/path/to/kubeconfig \
  dist/aicr/aicr validate \
    --recipe recipe.yaml \
    --data recipes/ \
    --emit-attestation ./attestation \
    --push ttl.sh/aicr-evidence:1h
```

| Phase | Validators | Result | Duration |
|---|---|---|---|
| deployment | operator-health, expected-resources, gpu-operator-version, check-nvidia-smi | **4/4 passed** | 25s |
| performance | nccl-all-reduce-bw-net (329 GB/s), nccl-all-reduce-bw-nvls (840 GB/s) | **2/2 passed** | 1m55s |
| conformance | dra-support, gang-scheduling, accelerator-metrics, ai-service-metrics, pod-autoscaling, cluster-autoscaling, gpu-operator-health, platform-health | **8/8 passed** | 1m29s |

Sign + push outcome:
- OCI artifact `sha256:80307d35…` packed and pushed before OAuth (avoids token-staleness)
- OAuth via GitHub identity provider; cert SAN recorded the signer's verified email
- Rekor inclusion proof captured (publicly auditable)
- Sigstore Bundle v0.3 attached as Referrer (`sha256:56bb53ab…` → main artifact)
- `cosign verify-attestation <ref> --type 'https://aicr.nvidia.com/recipe-evidence/v1' --certificate-identity ... --certificate-oidc-issuer ...` → **Verified OK** (cosign claims + Rekor offline + cert chain)
- `oras discover <ref>` cleanly shows the Referrer relationship

```bash
make qualify  # passes locally; coverage gate clean
```

Unit + integration tests cover:
- Predicate marshaling (recipe-subject vs artifact-subject)
- Bundle build deterministic across runs (same recipe → same subject digest)
- OCI Referrers push round-trips via `registry:2`
- Sigstore signing with a `NoOpSigner` for offline tests
- BOM auto-synthesis from recipe + catalog
- OIDC resolver chain precedence

## Risk Assessment

- [x] **Medium** — New CLI surface and a new signing-and-push pipeline, but the implementation reuses the existing `pkg/bundler/attestation` primitive (proven by `aicr bundle --attest`). The default behavior (no `--emit-attestation`) is completely unchanged. The `--emit-attestation` (no `--push`) path is fully offline and safe to ship.

**Rollout notes:** purely additive. No flag renames, no config-schema breaking changes. Operators not using `--emit-attestation` see no behavior change. The `recipes/evidence/<recipe>.yaml` convention is documented but no recipes ship a pointer file in this PR.

## Checklist

- [x] I updated docs if user-facing behavior changed (`docs/user/cli-reference.md`, `docs/design/007-recipe-evidence.md`)


mchmarny:
Whoa, that escalated quickly... 23 commits, 4618 additions, 28 files ;)

pkg/evidence/attestation/signer.go
```diff
@@ -0,0 +1,114 @@
...
+// Token discovery is the caller's responsibility.
+type KeylessSigner struct {
+	OIDCToken string
+	FulcioURL string
+	RekorURL  string
+}
+
+// NewKeylessSigner returns a signer wired to Sigstore public-good URLs.
+func NewKeylessSigner(oidcToken string) *KeylessSigner {
+	return &KeylessSigner{
```
mchmarny: **Parallel attestation surfaces will drift.** `KeylessSigner` here duplicates `pkg/bundler/attestation.KeylessAttester` (same Fulcio/Rekor URLs, same OIDC token, same delegation to `SignStatement`). The schema/predicate types belong here, but the signer interface is general — having two `Sign()`/`Attest()` shapes invites them to evolve independently (error handling, identity surfacing, Rekor knobs).

Suggest: promote one signer to be the shared primitive and import it from both places. The PR description says "no new sign/verify code paths" — this file contradicts that.

Separately: `OIDCToken string` is exported here, but unexported in `KeylessAttester`. Tokens are short-lived secrets; don't invite callers to read them back off the struct after `Sign()` returns.

pkg/cli/validate_evidence.go
```diff
@@ -0,0 +1,466 @@
...
+		slog.Debug("resolving OIDC token", "mode", "identity-token")
+	case cfg.OIDCResolve.AmbientURL != "" && cfg.OIDCResolve.AmbientToken != "":
+		slog.Debug("resolving OIDC token", "mode", "ambient-github-actions")
+	case cfg.OIDCResolve.DeviceFlow:
+		slog.Info("resolving OIDC token via device-code flow (will print a code to enter at the URL shown)")
+	default:
+		slog.Info("resolving OIDC token via browser flow (will open a local browser)")
+	}
+	resolveOpts := cfg.OIDCResolve
+	resolveOpts.PromptWriter = os.Stderr
```
mchmarny: **Token-resolution timing now diverges between `aicr validate --push` and `aicr bundle --attest`.** This call site correctly defers `ResolveOIDCToken` until adjacent to `Sign` (Fulcio nonce-binding window — exactly the bug described in the PR description). But `pkg/cli/bundle.go:602` still calls `ResolveAttester`, which resolves the token *before* the Helm-render-and-push runs. For large bundles, that path has the same staleness window the validate path now avoids.

Either apply the same fix to bundle (preferred — the refactor exposing `ResolveOIDCToken` already enables it), or explicitly document in resolver.go why bundle's timing is safe (it may be: bundle work is shorter). Carrying the asymmetry silently means whoever debugs the next "error processing the identity token" on bundle has to rediscover this.

pkg/cli/validate_evidence.go
```diff
@@ -0,0 +1,466 @@
...
+			}
+			for name, reading := range st.Data {
+				if name == "" || reading == nil {
+					continue
+				}
+				ref := name + ":" + reading.String()
+				if _, dup := seen[ref]; dup {
+					continue
+				}
+				seen[ref] = struct{}{}
```
mchmarny: **Business logic in `pkg/cli` violates the CLI/API separation rule.** Per CLAUDE.md: "pkg/cli and pkg/api = user interaction only, no business logic. Business logic lives in functional packages so CLI and API can both use it."

`loadOrGenerateBOM`, `buildAutoBOM`, `dedupValidatorImages`, `validatorImagesForPredicate`, `observedImagesFromSnapshot`, `catalogVersion`, and the orchestration in `emitRecipeEvidence`/`signAndPushBundle` are all genuine business logic — they shape evidence content from recipe/snapshot/catalog inputs. None of this can be reused from `pkg/api` when the API surface for evidence emission lands.

Natural homes:
- BOM synthesis (lines 289–408) → `pkg/evidence/attestation/bom.go` or extend `pkg/bom`
- Snapshot→image walking (410–438) → same package (it's coupled to the predicate shape)
- Orchestration (`emitRecipeEvidence`, `signAndPushBundle`) → `pkg/evidence/attestation/emit.go` taking typed inputs and an `EmitOptions`

Leave only flag parsing (`buildRecipeEvidenceConfig`) and the call into the new emit primitive in this file.

mchmarny: **Snapshot→BOM coupling is brittle and untested at the collector contract boundary.** This function string-matches `m.Type == TypeK8s` and `st.Name == "image"`, and emits `"<name>:<tag>"` with the registry intentionally stripped (collector contract).

Two concerns:
1. A future collector rename of either constant silently produces an empty observed-images section in the BOM — no failure, no warning. There's no test exercising the actual collector → evidence path against a real-shaped snapshot.
2. The BOM lists image refs that can't be uniquely resolved across mirrors. Auditors comparing the BOM against a registry may find this surprising. Worth surfacing in the predicate/pointer that this is a "names without registries" view, or having an explicit `--bom` recommendation in user-facing docs for any verification flow.

Recommend: contract test against the collector's measurement schema (golden snapshot → expected BOM image set), and a brief note in the predicate's `bom` block describing the registry-stripping behavior.

pkg/evidence/attestation/builder.go
```diff
@@ -0,0 +1,451 @@
...
+		logsDir = filepath.Join(opts.OutputDir, LogsBundleDirName)
+		if logErr := writeLogsBundle(logsDir, opts.PhaseLogs); logErr != nil {
+			return nil, logErr
+		}
+	}
+
+	// Manifest pre-commits per-file hashes for any logs the contributor
+	// may publish later, so the integrity chain extends to log content
+	// the bundle doesn't physically carry.
+	manifest, err := buildManifestWithLogPreCommits(summaryDir, opts.PhaseLogs)
```
mchmarny: **Pre-baked log-handling infrastructure for an out-of-scope feature.** `PhaseLogs`/`IncludeLogs`/`writeLogsBundle`/`streamCopy`/`logBasename`/`buildManifestWithLogPreCommits` are all wired here, but the PR description states `--include-logs`/`--push-logs` "were prototyped, then dropped pending validator log-capture rework." So today every caller passes an empty/nil `PhaseLogs`, and these branches are dead.

Two specific concerns with shipping the dead wiring:

1. **Incomplete trust boundary for when the feature lands.** `streamCopy` (line 375) uses `os.Open(filepath.Clean(f.SourcePath))` with `//nolint:gosec`, and `logBasename` only sanitizes the basename — `SourcePath` is a full path with no sandbox check. Whoever later wires `PhaseLogs` from validator output needs to know `SourcePath` is a trust-boundary input. That contract is not enforced (no prefix check, no allow-list, no documentation on the field). When this comes back, the security review needs to happen against the *whole* path, not be an afterthought.

2. **Dead code drifts.** Six months from now, when log capture is rewritten, this scaffolding may no longer match the new design and will need to be redone anyway.

Recommend: strip the dead branches (and `LogFile`, `PhaseLogs`, `IncludeLogs`, `writeLogsBundle`, `streamCopy`, `logBasename`, `buildManifestWithLogPreCommits`) and bring them back with the feature. Or, if keeping, add an explicit `SourcePath` prefix/sandboxing requirement and a TODO referencing #754 or the follow-up issue.

pkg/oci/push.go
```diff
@@ -460,39 +627,44 @@ func jitterDuration(d time.Duration) time.Duration {
...
+	dstPath := tempDir
+	if subDir != "" {
+		srcPath = filepath.Join(sourceDir, subDir)
+		dstPath = filepath.Join(tempDir, subDir)
+	}
+	if err := hardLinkDir(srcPath, dstPath); err != nil {
+		cleanup()
+		return "", nil, apperrors.Wrap(apperrors.ErrCodeInternal, "failed to create hard links", err)
+	}
+
```
mchmarny, line 694: **Semantic change in shared push path — call this out in the PR description.** This change removes the `subDir == ""` shortcut that returned `sourceDir` directly; every push now goes through tempdir+hardlink. The fix is justified (oras file store leaking annotation-named manifest blobs into source — the regression test in `push_test.go` documents the failure well). But this affects `pkg/cli/bundle.go:620` (`PackageAndPush`) too, not just the new evidence flow.

Currently the change is buried as a sub-bullet under "Sign sequence" in the PR description. Worth promoting to its own line in "Notable design choices" so reviewers of bundle-path changes see it.

Minor performance: every existing bundle push now adds a hardlink walk over the source tree. For typical-sized bundles this is fine; for the largest case worth confirming.

mchmarny (changes requested):
Solid end-to-end work — bundle layout, in-toto/Sigstore plumbing, and the GB200/EKS test run all look right. Requesting changes on architectural seams that will be hard to walk back later: parallel attestation packages (`pkg/evidence/attestation` duplicating `pkg/bundler/attestation` signer surface), inconsistent OIDC token-resolution timing between `aicr validate --push` and `aicr bundle --attest`, and a 465-line orchestration file in `pkg/cli` carrying business logic that should live in functional packages per CLAUDE.md. Plus dead log-handling scaffolding with an incomplete trust boundary, and a brittle snapshot→BOM coupling without contract tests. Details inline.

pkg/oci/push.go
njhensley: Promoted to its own bullet under **Notable design choices** in the PR description (between `Sign sequence` and `OCI Referrers attach`). Covers the rationale — oras file-store leaking annotation-named manifest blobs into the source — and points at the regression test in `pkg/oci/push_test.go`.

On the perf note: the walk is `os.Link` per file (no copy on the hot path), so the cost for typical bundles is negligible. I haven't run an explicit stress case against the largest bundle path — happy to put a number on it if you want it confirmed before merge.

mchmarny (approved):

PR #879, opened 2026-05-14
feat(evidence): aicr evidence verify (directory input)

njhensley:
## Summary

First slice of `aicr evidence verify` (ADR-007 PR-B): offline verification of an unpacked bundle directory produced by `aicr validate --emit-attestation`. Ships the receiving end of the bundle integrity chain plus a Markdown/JSON report.

## Motivation / Context

ADR-007 ships a recipe-evidence v1 bundle so contributors on hardware AICR maintainers can't reach can hand reviewers a signed, verifiable artifact. The emitter landed in a prior PR; this is the start of the consumer. Shipped as a vertical slice so each PR adds one capability an end user can invoke independently. This PR gives contributors directory self-debug — `aicr evidence verify ./out/summary-bundle` after `aicr validate --emit-attestation ./out`.

Signature verification, inline constraint replay, OCI pull, and pointer files follow in three subsequent PRs.

Fixes: N/A (#753 closes with PR 4 of this series)
Related: #753, [ADR-007](docs/design/007-recipe-evidence.md)

## Type of Change

## Component(s) Affected

- [x] Other: `pkg/evidence/verifier` (new package)

## Implementation Notes

Four steps run: `materialize-bundle` → `predicate-parse` → `manifest-hash-check` → `render-summary`.

**Trust property shipped — the receiving end of the bundle integrity chain.** Every file in the bundle is sha256-verified against `manifest.json`, and `sha256(manifest.json)` is verified against `predicate.Manifest.Digest` read from the bundled in-toto Statement (`statement.intoto.json`).

The Statement file is unsigned at this slice — both sides of the manifest-digest comparison come from the same untrusted bundle directory, so a tamperer who rewrites `manifest.json` and the unsigned Statement in lockstep passes verification. The check catches accidental corruption and lazy tampering, which is the right contract for the contributor-self-debug case PR 1 targets.

**How PR 3 closes the chain.** When signature verification lands, predicate extraction switches to the DSSE payload inside `attestation.intoto.jsonl` (a Sigstore Bundle with a Fulcio-issued cert and Rekor inclusion proof). At that point, `predicate.Manifest.Digest` is cryptographically anchored to an OIDC identity at sign-time, and PR 1's manifest-hash check extends that anchor to every bundled file. PR 4 (OCI pull) adds redundancy via content-addressed registry storage.

ADR-007 §"Verifier steps" enumerates twelve steps. This implementation runs four, collapsing the redundant per-field digest checks the manifest-hash chain already covers and folding display of signed fields (fingerprint, phase counts, BOM) into the renderer. The package doc records the mapping.

Non-directory inputs (pointer-like `.yaml`, OCI refs) are rejected with a clear "not yet supported" message rather than silently failing — keeps the contract visible in the error surface.

**Output flags.** Standard project pattern: `--output`/`-o` selects the destination (stdout when empty, a file otherwise); `--format`/`-t` selects text (Markdown) or json. Format applies regardless of destination; no double-rendering.

**Exit codes.** OS exit codes are `0` (valid) and `2` (anything else, including recorded phase failures), per `pkg/errors/exitcode.go`'s code-to-OS mapping. The 0/1/2 distinction from ADR-007 is preserved in the structured output — `VerifyResult.Exit` field in JSON / Markdown carries `1` for the informational phase-failure case so JSON consumers can branch. Shell consumers can branch via `jq '.exit'`.

## Testing

```bash
go test ./pkg/evidence/verifier/... ./pkg/cli/...
golangci-lint -c .golangci.yaml run ./pkg/evidence/verifier/... ./pkg/cli/...
./tools/check-docs-mdx
./tools/check-docs-sidebar
```

All green locally (lint 0 issues; tests pass).

Tests cover: happy-path verify of a synthesized bundle, tampered-file → exit 2, stray-file → exit 2, manifest-digest mismatch → exit 2, manifest schemaVersion rejection, traversal-entry rejection, ctx cancellation in inventory, predicate-parse failure attributed to its own step, auto-detect rejects unsupported input forms with clear errors, `-o` writes to file with empty stdout, `-o -t json` writes JSON to file, `-t json` emits the `steps` array on stdout.

## Risk Assessment

- [x] **Low** — additive new subcommand under `aicr evidence`; no existing behavior changes; fully covered by unit tests; the trust property (self-consistency on an unsigned Statement) is partial-by-design and explicitly described in the package doc.

**Rollout notes:** N/A. Users who don't invoke `aicr evidence verify` see no change.

## Checklist

- [x] Tests pass locally
- [x] Linter passes
- [x] I added tests for new functionality
- [x] I updated docs (`docs/user/cli-reference.md`)
- [x] Changes follow existing patterns
- [x] Commits are cryptographically signed


pkg/evidence/verifier/inventory.go
```diff
@@ -0,0 +1,164 @@
...
+		return nil, errors.New(errors.ErrCodeInvalidRequest, "materialized bundle is required")
+	}
+	body, err := os.ReadFile(filepath.Join(mat.BundleDir, attestation.ManifestFilename)) //nolint:gosec
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeNotFound, "failed to read manifest.json", err)
+	}
+	var manifest attestation.Manifest
+	if uErr := json.Unmarshal(body, &manifest); uErr != nil {
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "manifest.json is not valid JSON", uErr)
+	}
```
mchmarny, line 75: **Critical — manifest-digest binding to the predicate is never checked.**

The PR description, package doc, and CLI description all claim this slice ships the "manifest hash chain": `manifest.json` bound to `predicate.manifest.digest` ⇒ every bundled file transitively bound to the predicate. But this code reads `manifest.json` and hashes files against its entries — it never recomputes `sha256(manifest.json)` and compares against `predicate.Manifest.Digest`. Predicate parsing in `verify.go:70` doesn't consult `pred.Manifest.Digest` either.

Concrete consequence: tamper a bundle file, recompute `manifest.json` with the new hash, and the verifier reports "all bundle files verified against manifest.json" with `Exit=0`. Until signature verification lands, this digest comparison is the only thing that prevents a one-line attack. The existing `TestVerify_TamperedFileFails` happens to pass only because the test doesn't also rewrite the manifest.

Suggested fix: after reading `manifest.json` bytes here, compute `attestation.HashBytesSHA256(body)` and compare against `pred.Manifest.Digest` (will need to plumb the predicate or its `ManifestRef` into `CheckInventory`, or do the check in `verify.go` between `loadPredicate` and `CheckInventory`). Add a test that tampers both the file and the manifest and expects `Exit=2`.

pkg/cli/evidence_verify.go
```diff
@@ -0,0 +1,129 @@
...
+	switch result.Exit {
+	case verifier.ExitValidPassed:
+		return nil
+	case verifier.ExitValidPhaseFailures:
+		return errors.New(errors.ErrCodeConflict,
+			"bundle valid; recorded validator results show failures")
+	default:
+		return errors.New(errors.ErrCodeInvalidRequest,
+			"bundle verification failed; see the verifier output for details")
+	}
```
mchmarny, line 112: The PR description says "CLI maps the informational exit 1 (phase failures recorded in bundle) via `ErrCodeConflict` so consumers can branch." Consumers cannot actually branch on the OS exit code: `pkg/errors/exitcode.go` maps both `ErrCodeConflict` and `ErrCodeInvalidRequest` to `ExitInvalidInput = 2`. So `aicr evidence verify` returns OS exit 0 for valid, OS exit 2 for both "phase failures" and "bundle invalid" — no way to distinguish from a shell.

The `docs/user/cli-reference.md` exit-code table is accurate (it lists 0 and 2 only), and the `VerifyResult.Exit` field in JSON/Markdown does carry the 1-vs-2 distinction, so this is more a wording issue in the PR/code-comment than a behavior bug. Worth either:
- adding a dedicated exit-code mapping in `exitcode.go` for `ErrCodeConflict` if shell-level branching is wanted, or
- updating the inline reasoning to clarify that JSON output is the only branching channel.

pkg/evidence/verifier/verify.go
```diff
@@ -0,0 +1,145 @@
...
+	// Parse the predicate so display steps can surface signed fields.
+	// The predicate is read but not cryptographically verified —
+	// callers see this via the empty Signer field in the rendered
+	// report until signature verification lands.
+	pred, perr := loadPredicate(mat)
+	if perr != nil {
+		record(r, stepInventory, StepFailed, "could not parse predicate: "+perr.Error(), nil)
+		r.Exit = ExitInvalid
+		return r, nil
+	}
```
mchmarny, line 77: If predicate parsing fails (missing `statement.intoto.json`, bad JSON, wrong `predicateType`), the failure is recorded as `stepInventory` — the "manifest-hash-check" step. Users seeing `Step 2 manifest-hash-check failed: could not parse predicate` will reasonably read this as a hash-check failure, but the manifest was never inspected.

Cheap fix: record this against a distinct step name (e.g. `stepPredicate` between materialize and inventory, or record materialize as failed with predicate detail) so the step name matches what actually went wrong. The `stepNames`/`stepMaterialize|stepInventory|stepRender` block is the place to add it.

pkg/evidence/verifier/types.go
```diff
@@ -0,0 +1,76 @@
...
+const (
+	ExitValidPassed        = 0
+	ExitValidPhaseFailures = 1
+	ExitInvalid            = 2
+)
+
+// VerifyOptions configures one Verify run.
+type VerifyOptions struct {
+	Input        string
+	MarkdownPath string
```
mchmarny: nit: `MarkdownPath` is dead inside the library — `Verify()` never reads it; the CLI calls `WriteMarkdown` directly after `Verify` returns. Either drop the field from `VerifyOptions` or have `Verify` honor it (and remove the separate `WriteMarkdown` call from `runEvidenceVerifyCmd`). Carrying it through the options struct invites future callers to think setting it is sufficient.

pkg/evidence/verifier/fetch.go
```diff
@@ -0,0 +1,76 @@
...
+func (m *MaterializedBundle) Cleanup() {
+	if m == nil || m.cleanup == nil {
+		return
+	}
+	m.cleanup()
+	m.cleanup = nil
+}
+
+// MaterializeBundle dispatches on InputForm. Only InputFormDir is handled.
+func MaterializeBundle(_ context.Context, opts VerifyOptions, form InputForm) (*MaterializedBundle, error) {
```
mchmarny: nit: `ctx` is accepted but discarded (`_`). For dir-only it's cheap (a couple of `os.Stat` calls), but once OCI fetch lands the convention will need to flip. Either check `ctx.Err()` once at the top to keep behavior consistent with the rest of the package, or document why directory materialization is exempt.

pkg/evidence/verifier/inventory.go
```diff
@@ -0,0 +1,164 @@
...
+// both nil on success.
+func CheckInventory(ctx context.Context, mat *MaterializedBundle) ([]KV, error) {
+	if mat == nil || mat.BundleDir == "" {
+		return nil, errors.New(errors.ErrCodeInvalidRequest, "materialized bundle is required")
+	}
+	body, err := os.ReadFile(filepath.Join(mat.BundleDir, attestation.ManifestFilename)) //nolint:gosec
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeNotFound, "failed to read manifest.json", err)
+	}
+	var manifest attestation.Manifest
```
mchmarny, line 72: nit: `manifest.SchemaVersion` is parsed but never validated. If/when the manifest schema evolves a v2 bundle would silently pass against a v1 verifier (as long as field shapes overlap). Worth either checking against an expected constant here, or adding a TODO if the schema bump is planned in a follow-up.

mchmarny (changes requested):
Nice clean structure — three-step pipeline, good separation between library and CLI, traversal guard via `filepath.IsLocal`, sensible test coverage including the stray-file and traversal cases. CI is green.

One blocker on the trust property this slice is meant to ship: the manifest-digest binding from `manifest.json` → `predicate.Manifest.Digest` is never actually checked. The inventory step recomputes per-file hashes against `manifest.json`, but `manifest.json` itself is trusted as-is, so if someone wanted to fake a bundle file and also rewrote `manifest.json` to match will passe verification with `Exit=0`. Until signature verification lands in a follow-up, this single digest comparison is what closes the loop — see the inline comment on `inventory.go` for the suggested fix and a regression test.

The other inline comments are smaller: a step-naming clarity issue (predicate parse failures surface as Step 2 / manifest-hash-check), a wording mismatch around exit-code branching, and a few nits (dead `MarkdownPath` field, ignored `ctx`, unvalidated `SchemaVersion`).

Blocking on the manifest-digest check; everything else can be follow-up.

pkg/evidence/verifier/inventory.go
njhensley: Fixed. `CheckInventory` now takes `expectedManifestDigest string` (sourced from `predicate.Manifest.Digest`) and verifies `sha256(manifest.json)` matches before iterating per-file hashes; empty digest is rejected.

In this slice the predicate comes from the unsigned `statement.intoto.json`, so the check is self-consistency only — a tamperer who rewrites both files in lockstep still passes. PR 3 will switch predicate extraction to the verified Sigstore Bundle payload inside `attestation.intoto.jsonl`, at which point PR 1's manifest-hash check extends a Fulcio-anchored value to every bundled file. The PR description has been rephrased to be honest about this — the slice ships the receiving end of the chain, not the chain itself.

Test: `TestCheckInventory_TamperedManifestDigestFails`.

pkg/cli/evidence_verify.go
njhensley: Fixed (doc-only). PR description rewritten: OS exit codes are `0` (valid) and `2` (anything else, including recorded phase failures), per `pkg/errors/exitcode.go`'s code → OS mapping. The 0/1/2 distinction is preserved in `VerifyResult.Exit` for JSON consumers via `jq '.exit'`. The CLI code is unchanged — it was correct; the description was misleading.

pkg/evidence/verifier/verify.go
njhensley: Fixed. Added `stepPredicate` between `stepMaterialize` and `stepInventory`; predicate parse failures now record as `predicate-parse`, not `manifest-hash-check`. Step count grew from 3 to 4; `doc.go` and the cli-reference table updated. Test: `TestVerify_PredicateParseFailureRecordedAsPredicateStep`.

pkg/evidence/verifier/types.go
njhensley: Fixed. `MarkdownPath` dropped from `VerifyOptions`; `verifier.WriteMarkdown` deleted (no remaining callers). Subsequently took a second pass on the CLI output surface: dropped `--output-markdown` entirely and replaced with the project's standard `--output`/`-o` (destination) + `--format`/`-t` (format) pattern. Single dispatch via `openVerifyOutput`; stdout is empty when `-o` is set, no double-rendering. Tests: `TestEvidenceVerifyCmd_OutputMarkdownToFile`, `TestEvidenceVerifyCmd_OutputJSONToFile`.

pkg/evidence/verifier/fetch.go
njhensley: Fixed. `MaterializeBundle` now checks `ctx.Err()` at the top and returns wrapped `ErrCodeUnavailable` on cancellation, matching the rest of the package.

pkg/evidence/verifier/inventory.go
njhensley: Fixed. Added `isSupportedManifestSchema` (accepts `1.0.x`); `CheckInventory` rejects with a clear error otherwise. Matches the pointer-schema validation pattern. Test: `TestCheckInventory_RejectsUnsupportedSchema`.

pkg/evidence/verifier/verify.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// Step 2 — manifest hash check (covers every file in the bundle).
+	mismatches, invErr := CheckInventory(mat)
+	if invErr != nil {
+		record(r, stepInventory, StepFailed, invErr.Error(), mismatches)
+		r.Exit = ExitInvalid
+	} else {
+		record(r, stepInventory, StepPassed,
+			"all bundle files verified against manifest.json", nil)
+	}
```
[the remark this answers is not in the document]
njhensley: Fixed. `CheckInventory` now takes `ctx context.Context` and checks `ctx.Err()` between each per-file hash; `findExtras` also takes `ctx` and checks it on every WalkDir callback. Cancellation returns the partial mismatches plus a wrapped `ErrCodeUnavailable`. Test: `TestCheckInventory_RespectsCancellation`.

pkg/evidence/verifier/inventory.go
```diff
@@ -0,0 +1,151 @@
...
+	if info.IsDir() {
+		return "", errors.New(errors.ErrCodeInvalidRequest, "manifest entry "+rel+" is a directory")
+	}
+	if expectedSize > 0 && info.Size() != expectedSize {
+		return "", errors.New(errors.ErrCodeInvalidRequest,
+			"size mismatch for "+rel+
+				" (got "+strconv.FormatInt(info.Size(), 10)+
+				", want "+strconv.FormatInt(expectedSize, 10)+")")
+	}
+	return attestation.HashFileSHA256(full)
```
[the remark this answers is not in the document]
njhensley: Fixed. `hashFile` now captures the error from `attestation.HashFileSHA256` and wraps with `errors.PropagateOrWrap(err, errors.ErrCodeInternal, "failed to hash bundle file: "+rel)`, preserving the inner code (already `ErrCodeInternal` from `checksum.SHA256Raw`) while adding the bundle-relative context that matches how the other error paths in `hashFile` are anchored.

pkg/evidence/verifier/inventory.go
```diff
@@ -0,0 +1,156 @@
...
+
+	if len(mismatches) > 0 {
+		return mismatches, errors.New(errors.ErrCodeInvalidRequest,
+			"manifest inventory check failed for "+strconv.Itoa(len(mismatches))+" file(s)")
+	}
+	return nil, nil
+}
+
+func hashFile(bundleDir, rel string, expectedSize int64) (string, error) {
+	full := filepath.Join(bundleDir, filepath.FromSlash(rel))
```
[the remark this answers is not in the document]
njhensley: Fixed. `hashFile` now calls `filepath.IsLocal(localRel)` at the top and rejects with `ErrCodeInvalidRequest` before any filesystem access. Matches the existing project idiom in `pkg/recipe/provider.go` and `pkg/oci/push.go`. Test: `TestCheckInventory_RejectsTraversal`.

pkg/evidence/verifier/inventory.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+		if _, ok := exempt[rel]; ok {
+			return nil
+		}
+		extras = append(extras, rel)
+		return nil
+	})
+	if walkErr != nil {
+		return nil, errors.Wrap(errors.ErrCodeInternal, "failed to walk bundle dir", walkErr)
+	}
```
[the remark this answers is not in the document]
njhensley: Fixed. `findExtras` now inspects `walkErr` with `stderrors.Is` against `context.Canceled` / `context.DeadlineExceeded` and translates to `ErrCodeUnavailable` (instead of `ErrCodeInternal`), so cancellation reads the same way it does in the per-file loop above.

docs/user/cli-reference.md
```diff
@@ -1890,6 +1890,41 @@ aicr verify ./my-bundle --format json
...
+| 2 | Bundle invalid (manifest hash mismatch or predicate malformed), OR recorded validator results show failures. |
+
+The JSON/Markdown output's `exit` field (and `VerifyResult.Exit` from the library API) still distinguishes the two non-zero cases as `1` (recorded phase failures) vs `2` (bundle invalid).
+
+**Examples:**
+```shell
+aicr evidence verify ./out/summary-bundle
+aicr evidence verify ./out/summary-bundle --output-markdown ./summary.md
+aicr evidence verify ./out/summary-bundle --format json
+```
```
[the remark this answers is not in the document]
njhensley: Skipped. The repo doesn't run markdownlint (no `.markdownlint*` config, nothing in CI / Makefile references MD031/MD058), and the new `aicr evidence verify` section matches the established compact style used by every sibling subcommand section in `cli-reference.md` — `aicr verify`, `aicr bundle`, `aicr validate` all use `**Label:**` immediately followed by a fence or table with no blank line. Applying the rule just to the new section would create inconsistency with its neighbors; applying it everywhere is out of scope for this PR.

pkg/cli/evidence_verify.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return err
+	}
+	if err := writeVerifyOutput(cmd.Root().Writer, format, result); err != nil {
+		return err
+	}
+	if opts.MarkdownPath != "" {
+		if err := verifier.WriteMarkdown(opts.MarkdownPath, result); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
njhensley: Considered, not changed. `verifier.Verify` and the helpers it calls return structured errors with codes already set; per CLAUDE.md's anti-pattern list ("Don't double-wrap errors that already have proper codes"), bare `return err` here preserves the inner code rather than masking it. If the verifier API ever surfaces uncoded errors at this boundary, `errors.PropagateOrWrap` would be the right pattern — but today every error path through `Verify` carries a code.

mchmarny (approved):

Issue #736, opened 2026-05-04
Export FilterEntriesBySpec() for external catalog filtering


xdu31:
## Overview

Export the `FilterEntriesBySpec()` function for external consumers to filter validator catalog entries.

## Proposed API

```go
package validator

// FilterEntriesBySpec filters catalog entries based on validation configuration.
func FilterEntriesBySpec(catalog *ValidatorCatalog, spec *ValidationConfig) []ValidatorEntry {
    // Implementation
}
```

## Use Cases

- External consumers filtering catalogs before building Jobs
- Previewing which validators will run
- Custom catalog filtering logic

## Implementation

Location: `pkg/validator/catalog/filter.go`

Export the existing internal function with public visibility.

## Acceptance Criteria

- [ ] Function exported from package
- [ ] Unit tests for filtering logic
- [ ] Documentation with examples

## Related

- Parent: #731 (External validation controller integration)
- Depends on: #732, #733 (portable types must exist first)

mchmarny:
Implemented as `FilterEntriesByValidation` in `pkg/api/validator/v1/catalog.go` (exported with unit tests in `catalog_test.go`). The function name in the proposed API differed but the capability — filtering catalog entries by a validation predicate — is in place and exported. Closing — please re-file if a distinct API surface is still needed.

PR #880, opened 2026-05-14
refactor(validator): move Phase to v1 and export FilterEntriesByValidation

xdu31:
## Summary

Moves the `Phase` type and constants from `pkg/validator` to `pkg/api/validator/v1` for proper API layering. Exports `FilterEntriesByValidation` function from v1 package to enable external consumers to filter catalog entries based on validation configuration (addresses #736).

## Motivation and Context

The `Phase` type was defined in `pkg/validator` but used throughout the v1 API types. This created an inverted dependency where the API package imported from the validator package. Moving Phase to v1 establishes proper layering and allows external consumers to work with validator catalog entries.

## Type of Change

- [x] Refactoring (code restructuring without changing behavior)

## Components Affected

- `pkg/api/validator/v1` - added Phase type, constants, and FilterEntriesByValidation function
- `pkg/validator` - added type alias and constant re-exports for backward compatibility

## Implementation Notes

- Phase type and constants (PhaseDeployment, PhasePerformance, PhaseConformance) are now defined in `pkg/api/validator/v1/catalog.go`
- `pkg/validator/phases.go` re-exports the type and constants via type alias for backward compatibility
- `FilterEntriesByValidation` function moved to v1 package and made public
- `ValidatorEntry.Phase` remains as `string` type to avoid extensive conversions throughout the codebase
- Added comprehensive test coverage for `FilterEntriesByValidation` (7 test cases)

## Testing

- All unit tests pass with race detector
- E2E tests: 20 passed, 0 failed
- Added `pkg/api/validator/v1/catalog_test.go` with test coverage for FilterEntriesByValidation
- Coverage: pkg/api/validator/v1: 87.0%

## Risk Assessment

**Low risk** - Refactoring with backward compatibility maintained through type aliases and re-exports. No behavior changes.

## Checklist

- [x] Tests pass locally (`make test`)
- [x] Linting passes (`golangci-lint run`)
- [x] E2E tests pass (`make qualify`)
- [x] Documentation updated (godoc comments added)
- [x] Backward compatibility maintained


pkg/api/validator/v1/catalog.go
```diff
@@ -24,6 +24,20 @@ const (
 	CatalogKind = "ValidatorCatalog"
 )
 
+// Phase represents a validation phase.
```
mchmarny, line 27: The PR description justifies this move by "the API package imported from the validator package" / "inverted dependency". On main, `pkg/api/validator/v1/catalog.go` only imports `"time"` and uses `phase string` — there is no `pkg/api → pkg/validator` import to fix. `grep -rn "NVIDIA/aicr/pkg/validator" pkg/api/` returns nothing on main and nothing on this branch. The motivation as written doesn't match the diff.

The real change is "introduce a typed `Phase` in the v1 API so external catalog consumers can use the constants without depending on `pkg/validator`" — which is a fine thing, just describe it accurately. The "inverted dependency" claim has a way of getting cited in future PRs as historical context, so worth correcting now.

pkg/validator/validator_test.go
```diff
@@ -396,9 +396,9 @@ func TestFilterEntriesByValidationEmptyChecksList(t *testing.T) {
 		},
 	}
 	validationInput := v1.ToValidationInput(rec)
-	got := filterEntriesByValidation(entries, PhaseDeployment, validationInput)
+	got := v1.FilterEntriesByValidation(entries, PhaseDeployment, validationInput)
 	if len(got) != 0 {
-		t.Errorf("filterEntriesByValidation(empty checks list) returned %d entries, want 0 (skip)", len(got))
+		t.Errorf("v1.FilterEntriesByValidation(empty checks list) returned %d entries, want 0 (skip)", len(got))
```
mchmarny: After the move, this file keeps five tests (`TestFilterEntriesByValidationNilValidation`, `…NilPhaseConfig`, `…NoChecks`, `…WithChecks`, `…EmptyChecksList`) that just call `v1.FilterEntriesByValidation(...)` — same call the new `pkg/api/validator/v1/catalog_test.go` already covers from inside the package. The logic is fully tested in v1; carrying duplicate tests for a function this package no longer owns leaves two places to maintain on every behavior tweak.

Either delete these from `pkg/validator/validator_test.go` or trim them to a single integration smoke test that exercises the orchestrator's call path through `runPhase`, not the filter function directly.

pkg/api/validator/v1/catalog_test.go
```diff
@@ -0,0 +1,228 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import (
+	"testing"
+
+	"github.com/NVIDIA/aicr/pkg/recipe"
```
mchmarny: This v1 unit test imports `pkg/recipe` only to build a `*recipe.RecipeResult` that the test body then manually translates into a `*ValidationInput` (see `TestFilterEntriesByValidation_AllPhases` lines 188–202). The other tests in the same file construct `ValidationInput` directly without going through recipe types.

Since the goal of moving `FilterEntriesByValidation` here is precisely to let v1 stand on its own, the test should reflect that and avoid the recipe dependency entirely — build `ValidationInput` directly in the table rows. Otherwise you've created a circular-feeling test layer where the v1 package depends on its own dependent (recipe) for unit-test setup.

pkg/api/validator/v1/catalog.go
```diff
@@ -102,13 +116,57 @@ type EnvVar struct {
 	Value string `json:"value" yaml:"value"`
 }
 
-// ForPhase returns validators filtered by phase name.
-func (c *ValidatorCatalog) ForPhase(phase string) []ValidatorEntry {
+// ForPhase returns validators filtered by phase.
+func (c *ValidatorCatalog) ForPhase(phase Phase) []ValidatorEntry {
```
mchmarny, line 120: nit: `ForPhase(phase Phase)` accepts the typed phase, but `ValidatorEntry.Phase` is still a plain `string` (line 79 on this file, unchanged), so the comparison has to round-trip via `v.Phase == string(phase)` here and via `Phase(phase)` at the test call site (`validator_test.go:487`).

PR description acknowledges this ("`ValidatorEntry.Phase` remains as `string` type to avoid extensive conversions"), but the result is a half-typed surface: callers think they're in the typed world, then the struct field forces them out. Worth either committing to `Phase` on `ValidatorEntry` in this PR or accepting it as a follow-up — but the casts on every comparison signal the half-step.

mchmarny (commented):
Reasonable refactor — typed `Phase` at the v1 boundary, `FilterEntriesByValidation` exported alongside `ForPhase`. CI green. CodeRabbit already flagged the OOB in the test loop and the stale "run all" comment on `validator.go:265` — won't repeat those.

Four additional notes:

1. PR motivation cites an "inverted dependency" / `pkg/api → pkg/validator` import that doesn't exist on main — `catalog.go` only imported `time`. The change is still defensible (lets external catalog consumers reach `Phase` without depending on the orchestrator), just describe it accurately.
2. Test duplication — five `TestFilterEntriesByValidation*` cases in `pkg/validator/validator_test.go` now mirror the new v1 tests on a function this package no longer owns.
3. The v1 catalog test imports `pkg/recipe` just to build inputs that are translated to `ValidationInput` manually inside the test body. Construct `ValidationInput` directly and drop the dep.
4. `ValidatorEntry.Phase` is still a plain `string` while the surface is typed — every comparison round-trips through `string(phase)`. Half-typed.

Nothing blocks merge.

xdu31:
Review comments addressed:
- Fixed potential panic in catalog_test.go by checking bounds before array access
- Updated stale comment about FilterEntriesByValidation behavior
- Removed duplicate tests from validator_test.go (already covered in v1/catalog_test.go)
- Removed unnecessary recipe import by constructing ValidationInput directly

mchmarny (approved):

PR #881, opened 2026-05-14
fix(validator): make ExtractResult sidecar-safe by reading 'validator' container explicitly

### LINKED ISSUE #795: Make ExtractResult sidecar-safe by reading from 'validator' container explicitly
## Problem

The current `ExtractResult()` implementation in `pkg/validator/job/result.go` is **not sidecar-safe**. It assumes the first container in the pod is the validator:

```go
// Line 57 - assumes first container is validator
cs := jobPod.Status.ContainerStatuses[0]  // ❌ Breaks with sidecars!

// Line 82 - empty container name defaults to first container
logs, logErr := pod.GetPodLogs(ctx, d.clientset, d.namespace, jobPod.Name, "")  // ❌ Not explicit!
```

**Impact**: When external controllers inject sidecars for log streaming or result processing, `ExtractResult()` may read from the sidecar container instead of the validator, returning incorrect exit codes and logs.

## Solution

Make `ExtractResult()` explicitly read from the `"validator"` container by name, and freeze `"validator"` as part of the public contract (ADR-002).

### Changes Required

1. **Find container by name `"validator"`** instead of using `ContainerStatuses[0]`
2. **Pass explicit container name** to `GetPodLogs(..., "validator")`
3. **Update function documentation** to document the `"validator"` contract
4. **Add helper function** `findContainerStatus()` for name-based lookup
5. **Apply same fix to `HandleTimeout()`** (line 122 has the same issue)

### Implementation

```go
// Add helper function
func findContainerStatus(statuses []corev1.ContainerStatus, name string) (corev1.ContainerStatus, bool) {
    for _, cs := range statuses {
        if cs.Name == name {
            return cs, true
        }
    }
    return corev1.ContainerStatus{}, false
}

// Update ExtractResult
func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
    // ...
    
    // BEFORE: cs := jobPod.Status.ContainerStatuses[0]
    // AFTER:
    cs, found := findContainerStatus(jobPod.Status.ContainerStatuses, "validator")
    if !found {
        result.ExitCode = -1
        result.TerminationMsg = "container 'validator' not found (ADR-002 violation)"
        return result
    }
    
    // ...
    
    // BEFORE: logs, logErr := pod.GetPodLogs(ctx, d.clientset, d.namespace, jobPod.Name, "")
    // AFTER:
    logs, logErr := pod.GetPodLogs(ctx, d.clientset, d.namespace, jobPod.Name, "validator")
}
```

### Documentation Update

```go
// ExtractResult reads the exit code, termination message, and stdout from the
// "validator" container in a completed validator pod.
//
// CONTRACT (ADR-002): The container name MUST be "validator". This is a frozen
// public contract to ensure sidecar-safety — ExtractResult will only read from
// the "validator" container, ignoring any sidecar containers that may be
// injected by external controllers (e.g., log streaming, result processing).
//
// This method must be called after WaitForCompletion returns, when the Job is
// in a terminal state (Complete or Failed).
func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
```

## Files to Modify

- `pkg/validator/job/result.go` (lines 29-35, 51-57, 82, 122)

## Acceptance Criteria

- [ ] `ExtractResult()` finds container by name `"validator"` instead of using index 0
- [ ] `GetPodLogs()` is called with explicit container name `"validator"`
- [ ] `HandleTimeout()` uses same sidecar-safe lookup
- [ ] Function documentation explicitly states `"validator"` contract per ADR-002
- [ ] Unit tests verify behavior with multiple containers (validator + sidecar)
- [ ] Error message when `"validator"` container not found mentions ADR-002

## Related

- Parent: #731 (External validation controller integration)
- ADR-002: Container-per-validator contract

xdu31:
## Summary

Makes `ExtractResult()` and `HandleTimeout()` sidecar-safe by explicitly reading from the "validator" container by name instead of assuming the first container (index 0) is the validator.

## Motivation and Context

Fixes #795

The current implementation assumes the first container in the pod is the validator, which breaks when external controllers inject sidecars for log streaming or result processing. This change enforces the ADR-002 contract that the validator container must be named "validator".

## Type of Change

## Components Affected

- `pkg/validator/job` - result extraction logic

## Implementation Notes

- Added `findContainerStatus()` helper function for name-based container lookup
- Updated `ExtractResult()` to find container by name "validator" instead of using index 0
- Updated `HandleTimeout()` with same sidecar-safe lookup
- Pass explicit container name "validator" to `GetPodLogs()` calls
- Updated function documentation to document the ADR-002 contract
- Error messages mention ADR-002 when validator container not found

## Testing

- All unit tests pass with race detector
- Added 4 new test cases for sidecar scenarios:
  - `TestExtractResultWithSidecar` - validator + sidecar, reads from validator only
  - `TestExtractResultSidecarOnlyNoValidator` - sidecar only, fails with ADR-002 violation
  - `TestHandleTimeoutWithSidecar` - timeout with sidecar present
  - `TestHandleTimeoutSidecarOnlyNoValidator` - timeout with no validator container
- Updated all existing tests to set container Name field to "validator"
- E2E tests: 20 passed, 0 failed

## Risk Assessment

**Low risk** - Defensive change that makes existing functionality more robust. The "validator" container name is already part of the public contract (ADR-002), so this change enforces what should already be true.

## Acceptance Criteria

- [x] `ExtractResult()` finds container by name "validator" instead of using index 0
- [x] `GetPodLogs()` is called with explicit container name "validator"
- [x] `HandleTimeout()` uses same sidecar-safe lookup
- [x] Function documentation explicitly states "validator" contract per ADR-002
- [x] Unit tests verify behavior with multiple containers (validator + sidecar)
- [x] Error message when "validator" container not found mentions ADR-002

## Checklist

- [x] Tests pass locally (`make test`)
- [x] Linting passes (`golangci-lint run`)
- [x] E2E tests pass (`make qualify`)
- [x] Documentation updated (godoc comments)


pkg/validator/job/result.go
```diff
@@ -26,9 +26,16 @@ import (
...
-// ExtractResult reads the exit code, termination message, and stdout from a
-// completed validator pod. Returns a ValidatorResult regardless of how the
-// container terminated — the caller maps the result to a CTRF status.
+// ExtractResult reads the exit code, termination message, and stdout from the
+// "validator" container in a completed validator pod.
+//
+// CONTRACT (ADR-002): The container name MUST be "validator". This is a frozen
+// public contract to ensure sidecar-safety — ExtractResult will only read from
+// the "validator" container, ignoring any sidecar containers that may be
+// injected by external controllers (e.g., log streaming, result processing).
```
mchmarny: The comment, error messages, and docstrings repeatedly cite this as an "ADR-002 contract" / "ADR-002 violation", but [docs/design/002-validatorv2-adr.md](https://github.com/NVIDIA/aicr/blob/f6ed26b1f13447528feb3f5321d42ea49e3eba8e/docs/design/002-validatorv2-adr.md) doesn't actually codify the container name `validator` anywhere — `grep -n "container name\|MUST be\|frozen" docs/design/002-validatorv2-adr.md` returns nothing.

The contract is real and worth enforcing, but right now this PR is *establishing* it rather than *referring to* it. Two ways to make the references accurate:

1. Add the contract to ADR-002 in this PR (a short "Validator container name" subsection under the existing "Every validator container" / "Each validator runs as a K8s Job" sections is the right home).
2. Reword the comments/messages to attribute the contract to the validator package — e.g. `"container 'validator' not found (orchestrator contract: container must be named 'validator')"`.

Without one of these, anyone reading the error message and clicking through to ADR-002 will be confused. Suggest option 1 since the PR claims the contract is "frozen public".

pkg/validator/job/result.go
```diff
@@ -109,27 +115,28 @@ func (d *Deployer) HandleTimeout(ctx context.Context) *ctrf.ValidatorResult {
...
-			result.ExitCode = -1
-			result.TerminationMsg = fmt.Sprintf("timeout: validator did not complete within %s", d.entry.Timeout)
-		}
+	// Try to get container status from "validator" container (ADR-002 contract)
+	cs, found := findContainerStatus(jobPod.Status.ContainerStatuses, "validator")
+	if !found {
+		result.ExitCode = -1
+		result.TerminationMsg = fmt.Sprintf("timeout: validator did not complete within %s (container 'validator' not found)", d.entry.Timeout)
+		return result
+	}
```
mchmarny: `GetPodLogs` is called before checking whether the `validator` container exists in `Status.ContainerStatuses`. On a pod that has only a sidecar (the `TestHandleTimeoutSidecarOnlyNoValidator` case), this issues a wasted apiserver call that the kubelet will reject with `container "validator" is not valid for pod ...`, then silently swallows the error via `logErr == nil && logs != ""`.

Cheap reorder: do the `findContainerStatus` lookup first, return early if not found, and only call `GetPodLogs` once the container is known to exist. Same wire behavior, one fewer apiserver round trip on the timeout path, and the error path becomes deterministic.

pkg/validator/job/result.go
```diff
@@ -172,3 +179,18 @@ func filterStdoutLines(lines []string, maxLineLen int) []string {
...
 	return pod.GetPodForJob(ctx, d.clientset, d.namespace, d.jobName)
 }
+
+// findContainerStatus finds a container status by name in the pod's container
+// status list. Returns the container status and true if found, or a zero value
+// and false if not found.
+//
+// This helper ensures sidecar-safety by allowing explicit container name lookup
+// instead of assuming index 0 is the validator container (ADR-002).
+func findContainerStatus(statuses []corev1.ContainerStatus, name string) (corev1.ContainerStatus, bool) {
```
mchmarny, line 196: nit: if you take CodeRabbit's suggestion to extract `"validator"` to a constant, `pkg/validator/job/deployer.go:175` (`.WithName("validator")`) is the producer side — it should reference the same constant so the producer and verifier can't drift. Putting the constant in `pkg/validator/job` (this package) is simpler than `pkg/defaults`, since this is a validator-job-specific contract and `pkg/defaults` is currently scoped to timeouts/limits.

mchmarny (commented):
Targeted fix, good test coverage of the sidecar-present and sidecar-only cases. CI green.

CodeRabbit already covered the magic-string extraction, the missing ADR-002 wording in `HandleTimeout`, and the test assertion strengthening — won't repeat those. Three additional notes:

1. The "ADR-002 contract" / "ADR-002 violation" wording in comments and error messages isn't actually backed by ADR-002 — the doc doesn't codify the `validator` container name. Either add the contract to ADR-002 in this PR or reword to attribute it to the validator package.
2. In `HandleTimeout`, the log fetch happens before the container-existence check — small reorder saves a wasted apiserver call on the missing-validator path.
3. Producer side (`deployer.go:175` `.WithName("validator")`) should consume the same constant CodeRabbit suggested extracting; pinning it inside `pkg/validator/job` is simpler than `pkg/defaults`.

Nothing blocks merge.

xdu31:
Review comments addressed:
- Extracted 'validator' string to ValidatorContainerName constant
- Updated references from 'ADR-002' to 'validator package contract' (per feedback that ADR-002 doesn't document this)
- Reordered HandleTimeout to check container existence before fetching logs
- Updated deployer.go to use the constant
- Strengthened test assertions to explicitly check for container name

mchmarny (approved):

PR #890, opened 2026-05-14
feat(evidence): aicr evidence verify (signed OCI bundles)

### LINKED ISSUE #753: aicr verify-evidence command for reviewers and CI

njhensley:
## Summary

Second slice of `aicr evidence verify` (closes #753): adds Sigstore signature verification, OCI pull with Referrers-API discovery of the attached Sigstore Bundle, pointer-file input form, and the end-to-end producer→consumer demo. The trust handoff is complete after this PR.

## Motivation / Context

PR 1 (#879) shipped offline directory verification — the receiving end of the bundle integrity chain. The chain was self-consistency only because `statement.intoto.json` is never signed; the actual signature lives in `attestation.intoto.jsonl`. This PR closes that gap by verifying the Sigstore-signed Statement and switching predicate extraction to the cryptographically anchored payload, plus the two transport forms most reviewers and CI workflows will actually use (OCI ref + pointer file).

Constraint replay (ADR-007's step 5) is a refinement that re-derives constraint outcomes against the bundled snapshot; it's tracked as a follow-up enhancement.

Fixes: #753
Related: ADR-007 PR-B; depends on emitter from #873

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `pkg/evidence/verifier` (extends existing package; no API breaks)

## Implementation Notes

**Five-step pipeline** (was three in PR 1):

1. `materialize-bundle` — directory / pointer / OCI ref dispatch. Pointer and OCI forms pull the artifact via ORAS, then discover the Sigstore Bundle attached as an OCI Referrer (artifact type `application/vnd.dev.sigstore.bundle.v0.3+json`) and stage it locally as `attestation.intoto.jsonl`. Referrer is best-effort: an unsigned bundle has no referrer and the signature step records Skipped.
2. `signature-verify` — sigstore-go verifies the DSSE-wrapped Statement against the Sigstore trusted root (Fulcio cert chain + Rekor entry). Cross-checks the signed subject digest against the pulled artifact's OCI manifest digest. Extracts the verified predicate body from inside the DSSE payload.
3. `predicate-parse` — prefers the verified predicate from step 2; falls back to the bundle's unsigned `statement.intoto.json` when no signature is attached (PR 1's path).
4. `manifest-hash-check` — unchanged from PR 1, but `predicate.Manifest.Digest` is now cryptographically anchored when the bundle is signed.
5. `render-summary`.

**Trust chain when signed:**

```
Sigstore trusted root → Fulcio cert (Rekor-logged)
  → DSSE-signed Statement
    → predicate.Manifest.Digest
      → manifest.json
        → every bundled file's sha256
```

Tampering anywhere below the signature breaks the chain. The OCI input form locks the signed subject digest to the pulled artifact's OCI manifest digest — a substituted bundle with a stale signature fails.

**Pointer signer cross-check.** Pointer files can declare `signer.identity` / `signer.issuer` / `signer.rekorLogIndex`. The orchestrator extracts the actual signer claims from the verified Sigstore Bundle and compares: a pointer that claims `identity: alice` against a bundle signed by `mallory` fails loudly with a field-by-field diff. Catches the malicious-pointer scenario where the pointer points at the right OCI bundle but lies about who signed it.

**Tag-pin enforcement.** Tag-only OCI references are refused by default — tags are registry-rewritable and not content-addressable. `--allow-unpinned-tag` opts in for one-off debugging. Pointer-driven flows accept tag-form `bundle.oci` only when the pointer also carries a non-empty `bundle.digest` (that digest is the pin and is cross-checked against the pulled artifact).

**Sigstore error sanitization.** `sigstore-go`'s `pkg/verify/tsa.go:88` wraps an empty `errors.Join(nil...)` with `%w`, producing a `%!w(<nil>)` Go format artifact in the user-visible error. `sanitizeSigstoreError` strips it before wrapping. Tracked upstream.

**Header rendering.** The Markdown report's Signer line distinguishes "signature verification failed" from "unsigned bundle" — a signed bundle whose signature didn't verify is meaningfully different from a bundle that carries no signature at all.

**Failed-step details.** The Markdown report appends a "Failed check details" section after the steps table that enumerates per-file sub-rows for any failed step. Markdown tables can't render nested lists; this section is the bullet form.

**`--no-rekor` removed.** The flag was effectively dead code for the current emitter path (cosign keyless → Rekor only): without Rekor and without a TSA timestamp, sigstore-go's threshold check fails with zero verified timestamps. Will come back when the emitter grows TSA support.

### What this PR does NOT bind

Worth flagging for reviewers — the signature covers the post-resolution `recipe.yaml` and the predicate body (via the manifest hash chain). It does NOT bind:

- `recipes/components/<name>/values.yaml` or `manifests/*.yaml` content. The `componentRefs` entry stores file *paths*; the bundler reads the bytes at bundle-time. An edit to those files between attest-time and verify-time silently changes deployment without changing the canonical recipe digest. ADR-007's deferred `chainManifest` is the answer.
- Upstream Helm chart bytes at a given version string. ADR-006's image-digest pinning covers the container side; chart content is a separate gap.
- Whether the bundled snapshot reflects a real cluster (cluster-physicality gap, ADR-007 §"Trust model").

The package doc records the depth honestly.

## Testing

Validated against a real GB200/EKS bundle (6 nodes, 14 validators across deployment / performance / conformance) signed via cosign keyless + Sigstore + Rekor, published as an OCI Referrer.

**Happy-path inputs (all verify clean, exit 0):**

- Directory: `aicr evidence verify ./bundle/` (unsigned local copy — signature step skips cleanly)
- OCI ref by digest: `…@sha256:f90b…` (pulls + discovers Sigstore Bundle Referrer + verifies)
- Pointer file: same as OCI, plus pointer-claim cross-check against the cert

**Negative-path sweep — six probes against the same bundle:**

| # | Scenario | Initial behavior | After fix |
|---|---|---|---|
| 1 | Tamper byte in `recipe.yaml` | ✅ Fails at manifest-hash-check, exit 2 | + offending file now named in "Failed check details" |
| 2 | Bogus digest in pointer | ✅ Fails at materialize-bundle (both digests printed) | unchanged |
| 3 | Attacker email in pointer's `signer.identity` | 🔴 **Passed silently** — pointer claim was cosmetic | ✅ Cross-checks vs. cert SAN, fails loud with field-by-field diff |
| 4 | Tag-only OCI ref | 🟡 **Passed** — registry retag could swap content under us | ✅ Refuses by default; `--allow-unpinned-tag` opt-in |
| 5 | `-t json` output | ✅ Clean structured shape (suitable for CI/policy) | unchanged |
| 6 | Pure-digest URI (no tag) | ✅ Sig binds to digest, verifies | unchanged |

Two real security gaps closed (pointer-signer enforcement in #3, digest-pin enforcement in #4) plus one UX gap (#1's "Failed check details" rendering). Three secondary issues surfaced while probing — a `--no-rekor` flag that could never succeed against keyless-signed bundles, a `_unsigned bundle_` header that fired on signature-failure too, and a Go format-string artifact leaking through sigstore-go errors — all addressed in the commit history.

### Local checks

```bash
make qualify
```

All green locally:

```bash
go test ./pkg/evidence/verifier/... ./pkg/cli/...
golangci-lint -c .golangci.yaml run ./pkg/evidence/verifier/... ./pkg/cli/...
./tools/check-docs-mdx
./tools/check-docs-sidebar
```

**Coverage highlights:**
- `pointer_test.go` — load + validate happy path; rejects unsupported schema, missing recipe, no/multiple attestations, wrong predicate type, bad digest format, oversized file.
- `signature_test.go` — sigstore-error sanitizer table; cert-chain-error matcher; identity matcher with and without pinning; absent-file returns `ErrUnsignedBundle`; nil-bundle errors; pure-function Statement parser with malformed inputs.
- `referrer_test.go` — single-layer happy path against an in-memory fake fetcher; rejects multi-layer manifest, oversized manifest, oversized layer (claimed size), malformed JSON.
- `fetch_test.go` — parses tag / digest / oci-scheme / invalid forms; directory-input rejects non-bundles.
- `crosscheck_test.go` — pointer signer cross-check 10-case table; tag-pin enforcement (tag rejected without flag, accepted with flag); `isDigestPinned` helper.
- `verify_test.go` — directory happy path asserts the 5-step transcript; manifest tamper / stray file / unsupported schema / traversal entry / ctx cancellation; predicate-parse failure records under its own step; Markdown header distinguishes passed/skipped/failed signature; Failed-check-details section appears only when needed.

Sigstore-actual verification (TLS handshake to Fulcio + Rekor) depends on the trusted root which depends on network access; the existing `pkg/bundler/verifier` integration tests cover the sigstore-go primitives end-to-end. The verifier-side tests cover everything that surrounds them.

## Risk Assessment

Adds cryptographic verification paths, OCI network I/O, and a new pointer schema dependency. Tag-pin and signer cross-check fail closed by default. Directory input from PR 1 is unchanged. The new transport forms are gated behind explicit user invocation; users who only run `aicr evidence verify ./out/summary-bundle` see no behavior change.

**Rollout notes:** No migration. The directory verify command from PR 1 continues to work. New flags (`--bundle`, `--expected-issuer`, `--expected-identity-regexp`, `--registry-plain-http`, `--registry-insecure-tls`, `--allow-unpinned-tag`) are additive and default-off.

## Checklist

### DIFF
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -1912,7 +1912,6 @@ The positional argument is auto-detected as one of:
 |------|-------|------|---------|-------------|
 | `--output` | `-o` | string | | Write output to this file. When empty, output goes to stdout. |
 | `--format` | `-t` | string | `text` | Output format: `text` (Markdown) or `json`. Applies regardless of destination. |
-| `--no-rekor` | | bool | `false` | Skip Rekor transparency-log cross-check; rely on the cert + signature in the Sigstore Bundle alone. Useful in air-gapped environments. |
 | `--expected-issuer` | | string | | Pin the OIDC issuer URL on the signing certificate. Empty allows any issuer. |
 | `--expected-identity-regexp` | | string | | Pin the signer's `SubjectAlternativeName` via regex. Empty allows any identity. |
 | `--bundle` | | string | | OCI reference override when the pointer carries no `bundle.oci`. |
@@ -1943,9 +1942,6 @@ aicr evidence verify ./out/summary-bundle
 aicr evidence verify recipes/evidence/<recipe>.yaml \
   --expected-issuer https://token.actions.githubusercontent.com \
   --expected-identity-regexp '^https://github\.com/myorg/.*$'
-
-# Air-gapped: skip Rekor cross-check, rely on the Sigstore Bundle alone.
-aicr evidence verify recipes/evidence/<recipe>.yaml --no-rekor
 
 # CI pipelines: JSON output.
 aicr evidence verify recipes/evidence/<recipe>.yaml -o result.json -t json
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -1892,37 +1892,68 @@ aicr verify ./my-bundle --format json
 
 ### aicr evidence verify
 
-Verify a recipe-evidence v1 bundle produced by `aicr validate --emit-attestation`. Recomputes every file's sha256 against the bundle's `manifest.json` (which is bound to the predicate's `manifest.digest`) and surfaces the predicate's fingerprint, phase counts, and BOM info.
+Verify a recipe-evidence v1 bundle produced by `aicr validate --emit-attestation`. When the bundle carries a signature, verifies it against the Sigstore trusted root and extracts the cryptographically anchored predicate. Recomputes every file's sha256 against `manifest.json` (which the predicate's `manifest.digest` field anchors), and surfaces the predicate's fingerprint, phase counts, and BOM info.
 
-Only directory input is supported today. Cryptographic signature verification, inline constraint replay, OCI pull, and pointer-file support are not yet implemented.
+Inline constraint replay is reserved for a follow-up PR.
 
 **Synopsis:**
 ```shell
-aicr evidence verify <directory> [flags]
+aicr evidence verify <input> [flags]
 ```
+
+The positional argument is auto-detected as one of:
+
+* `recipes/evidence/<recipe>.yaml` — pointer file (verifier fetches the OCI artifact named inside).
+* `ghcr.io/<owner>/aicr-evidence@sha256:...` or `oci://...` — OCI reference.
+* `./out/summary-bundle/` (or a parent containing it) — unpacked directory.
 
 **Flags:**
 | Flag | Alias | Type | Default | Description |
 |------|-------|------|---------|-------------|
 | `--output` | `-o` | string | | Write output to this file. When empty, output goes to stdout. |
 | `--format` | `-t` | string | `text` | Output format: `text` (Markdown) or `json`. Applies regardless of destination. |
+| `--no-rekor` | | bool | `false` | Skip Rekor transparency-log cross-check; rely on the cert + signature in the Sigstore Bundle alone. Useful in air-gapped environments. |
+| `--expected-issuer` | | string | | Pin the OIDC issuer URL on the signing certificate. Empty allows any issuer. |
+| `--expected-identity-regexp` | | string | | Pin the signer's `SubjectAlternativeName` via regex. Empty allows any identity. |
+| `--bundle` | | string | | OCI reference override when the pointer carries no `bundle.oci`. |
+| `--registry-plain-http` | | bool | `false` | Use HTTP for registry traffic (local-registry tests only). |
+| `--registry-insecure-tls` | | bool | `false` | Skip TLS verification for the registry (self-signed certificates). |
 
 **Exit codes:**
 
 | Code | Meaning |
 |------|---------|
 | 0 | Bundle valid; every check passed. |
-| 2 | Bundle invalid (manifest hash mismatch or predicate malformed), OR recorded validator results show failures. |
+| 2 | Bundle invalid (signature, integrity, or predicate failure), OR recorded validator results show failures. |
 
-The JSON/Markdown output's `exit` field (and `VerifyResult.Exit` from the library API) still distinguishes the two non-zero cases as `1` (recorded phase failures) vs `2` (bundle invalid).
+The JSON/Markdown output's `exit` field (and `VerifyResult.Exit` from the library API) still distinguishes the two non-zero cases as `1` (recorded phase failures) vs `2` (bundle invalid). Shell consumers can branch via `jq '.exit'` on `--format json` output.
 
 **Examples:**
 ```shell
-aicr evidence verify ./out/summary-bundle                  # Markdown to stdout
-aicr evidence verify ./out/summary-bundle -o summary.md    # Markdown to file
-aicr evidence verify ./out/summary-bundle -t json          # JSON to stdout
-aicr evidence verify ./out/summary-bundle -o r.json -t json   # JSON to file
+# Verify the pointer that a contributor committed alongside their recipe change.
+aicr evidence verify recipes/evidence/h100-eks-ubuntu-training.yaml
+
+# Verify a pushed OCI bundle directly (no repo checkout required).
+aicr evidence verify ghcr.io/myorg/aicr-evidence@sha256:abc...
+
+# Verify a local bundle directory (contributor self-debug before push).
+aicr evidence verify ./out/summary-bundle
+
+# Pin the expected OIDC signer.
+aicr evidence verify recipes/evidence/<recipe>.yaml \
+  --expected-issuer https://token.actions.githubusercontent.com \
+  --expected-identity-regexp '^https://github\.com/myorg/.*$'
+
+# Air-gapped: skip Rekor cross-check, rely on the Sigstore Bundle alone.
+aicr evidence verify recipes/evidence/<recipe>.yaml --no-rekor
+
+# CI pipelines: JSON output.
+aicr evidence verify recipes/evidence/<recipe>.yaml -o result.json -t json
 ```
+
+See [demos/evidence.md](../../demos/evidence.md) for a full producer-and-consumer walkthrough.
+
+> **Stale root:** If verification fails with certificate chain errors, run `aicr trust update` to refresh the Sigstore trusted root.
 
 ---
 
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -1917,6 +1917,7 @@ The positional argument is auto-detected as one of:
 | `--bundle` | | string | | OCI reference override when the pointer carries no `bundle.oci`. |
 | `--registry-plain-http` | | bool | `false` | Use HTTP for registry traffic (local-registry tests only). |
 | `--registry-insecure-tls` | | bool | `false` | Skip TLS verification for the registry (self-signed certificates). |
+| `--allow-unpinned-tag` | | bool | `false` | Accept tag-only OCI references. By default the verifier refuses unpinned refs because tags are registry-rewritable; opt in only for one-off debugging. Pointer-driven flows ignore this flag when the pointer carries a `sha256:` digest. |
 
 **Exit codes:**
 
@@ -1947,7 +1948,7 @@ aicr evidence verify recipes/evidence/<recipe>.yaml \
 aicr evidence verify recipes/evidence/<recipe>.yaml -o result.json -t json
 ```
 
-See [demos/evidence.md](../../demos/evidence.md) for a full producer-and-consumer walkthrough.
+See [`demos/evidence.md`](https://github.com/NVIDIA/aicr/blob/main/demos/evidence.md) for a full producer-and-consumer walkthrough.
 
 > **Stale root:** If verification fails with certificate chain errors, run `aicr trust update` to refresh the Sigstore trusted root.
 
--- a/pkg/evidence/verifier/crosscheck_test.go
+++ b/pkg/evidence/verifier/crosscheck_test.go
@@ -15,8 +15,10 @@
 package verifier
 
 import (
+	"context"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/NVIDIA/aicr/pkg/evidence/attestation"
 )
@@ -143,15 +145,24 @@ func TestMaterializeBundle_RejectsTagOnlyOCI(t *testing.T) {
 }
 
 func TestMaterializeBundle_TagOnlyWithAllowFlag(t *testing.T) {
-	// With AllowUnpinnedTag, we get past the pin check and fail at
-	// the OCI pull (no actual registry available in the test). The
-	// pin-error message must NOT appear.
-	form, _ := DetectInputForm("ghcr.io/owner/repo:v1")
-	_, err := MaterializeBundle(t.Context(),
-		VerifyOptions{Input: "ghcr.io/owner/repo:v1", AllowUnpinnedTag: true},
+	// With AllowUnpinnedTag, the pin check is bypassed and execution
+	// proceeds into oras.Copy. Use 127.0.0.1:1 (a port nothing should
+	// be listening on) so connection fails fast — never reaching the
+	// real internet — and a 250ms per-test context so flaky DNS or a
+	// slow stack can't pay the 2-minute pull timeout. The assertion
+	// is only that the pin-error message is NOT in the result.
+	const unreachable = "127.0.0.1:1/repo:v1"
+	form, err := DetectInputForm(unreachable)
+	if err != nil {
+		t.Fatalf("DetectInputForm: %v", err)
+	}
+	ctx, cancel := context.WithTimeout(t.Context(), 250*time.Millisecond)
+	defer cancel()
+	_, err = MaterializeBundle(ctx,
+		VerifyOptions{Input: unreachable, AllowUnpinnedTag: true},
 		form, nil)
 	if err == nil {
-		t.Fatalf("expected pull error (no registry), got nil")
+		t.Fatalf("expected pull error (unreachable registry), got nil")
 	}
 	if strings.Contains(err.Error(), "tag-only") {
 		t.Errorf("AllowUnpinnedTag should bypass tag-pin check; got %v", err)
--- a/pkg/evidence/verifier/crosscheck_test.go
+++ b/pkg/evidence/verifier/crosscheck_test.go
@@ -1,0 +1,159 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package verifier
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/NVIDIA/aicr/pkg/evidence/attestation"
+)
+
+func i64ptr(v int64) *int64 { return &v }
+
+func TestCrossCheckPointerSigner(t *testing.T) {
+	tests := []struct {
+		name      string
+		claimed   *attestation.PointerSigner
+		actual    *SignerClaims
+		wantErr   bool
+		errSubstr string
+	}{
+		{
+			name:    "no claim, no actual",
+			claimed: nil,
+			actual:  nil,
+		},
+		{
+			name:    "no claim, signed bundle (claim-agnostic pointer)",
+			claimed: nil,
+			actual:  &SignerClaims{Identity: "x", Issuer: "y"},
+		},
+		{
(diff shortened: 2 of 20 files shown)


pkg/evidence/verifier/fetch.go
```diff
@@ -81,3 +116,265 @@ func hasBundleMarkers(dir string) bool {
...
+	// The Sigstore Bundle is attached as an OCI Referrer of the main
+	// artifact, not part of the artifact's own layers. Discover and
+	// stage it as attestation.intoto.jsonl so signature verification
+	// can read it from disk the same way it does for directory input.
+	// Best-effort: an unsigned bundle has no referrer and the signature
+	// step records Skipped (matches the unsigned-on-disk behavior).
+	if err := discoverAndWriteReferrer(pullCtx, remoteRepo, desc, resolved); err != nil {
+		slog.Debug("no Sigstore Bundle referrer discovered",
+			"reference", registry+"/"+repo, "error", err.Error())
+	}
```
mchmarny, line 247: Silent downgrade risk: every error from `discoverAndWriteReferrer` collapses to `slog.Debug` here — including the "registry returned a multi-layer / oversized / malformed manifest" cases. A registry that mid-MITMs the Referrers response can downgrade a signed bundle to "unsigned" because the signature step then sees no `attestation.intoto.jsonl` on disk and records Skipped.

This violates the project's "Fail closed on evaluator errors" rule (CLAUDE.md anti-patterns table) and works around a real attack only thanks to `mat.Digest` cross-check on the signed subject — but only when a signed bundle is actually parsed, which is exactly what this swallow prevents.

Suggestion: distinguish `errors.Is(err, ErrCodeNotFound)` (legitimately unsigned, debug-log) from any other failure (warn at minimum; ideally surface as a failed signature step when the bundle was supposed to be signed). For pointer-driven flows the pointer's `signer != nil` already implies "must be signed" — propagating the error there would catch this cleanly.

pkg/evidence/verifier/fetch.go
```diff
@@ -81,3 +116,265 @@ func hasBundleMarkers(dir string) bool {
...
+					continue
+				}
+				if err := fetchAndWriteReferrerLayer(ctx, repo, r, bundleDir); err != nil {
+					return err
+				}
+				found = true
+				// Take the first matching referrer. Multi-signature
+				// bundles aren't a V1 case; if one ever lands, we'd
+				// need a selection policy.
+				return nil
```
mchmarny: nit: `return nil` here only short-circuits this page's `for` loop — `repo.Referrers` will continue fetching subsequent pages and re-invoke the callback. If the registry ever returns a second page with another sigstore-bundle referrer, `fetchAndWriteReferrerLayer` would overwrite `attestation.intoto.jsonl` from page N with the page-(N+1) layer.

Today this is benign (V1 emits one referrer), but the "Take the first matching referrer" comment immediately above promises stop-after-first semantics that the code doesn't actually deliver. Either return a sentinel error and unwrap it after `Referrers` returns, or weaken the comment.

docs/user/cli-reference.md
```diff
@@ -1892,38 +1892,65 @@ aicr verify ./my-bundle --format json
...
 **Flags:**
 | Flag | Alias | Type | Default | Description |
 |------|-------|------|---------|-------------|
 | `--output` | `-o` | string | | Write output to this file. When empty, output goes to stdout. |
 | `--format` | `-t` | string | `text` | Output format: `text` (Markdown) or `json`. Applies regardless of destination. |
+| `--expected-issuer` | | string | | Pin the OIDC issuer URL on the signing certificate. Empty allows any issuer. |
+| `--expected-identity-regexp` | | string | | Pin the signer's `SubjectAlternativeName` via regex. Empty allows any identity. |
+| `--bundle` | | string | | OCI reference override when the pointer carries no `bundle.oci`. |
+| `--registry-plain-http` | | bool | `false` | Use HTTP for registry traffic (local-registry tests only). |
+| `--registry-insecure-tls` | | bool | `false` | Skip TLS verification for the registry (self-signed certificates). |
```
mchmarny, line 1919: Missing flag in this table: `--allow-unpinned-tag` is wired into the CLI (`pkg/cli/evidence_verify.go`) and is the supported escape hatch for tag-form OCI refs, but it isn't documented here. Worth adding — it's the only opt-out for the new "tag-only refs refused by default" behavior, which is a meaningful change reviewers will want to discover from the reference page rather than from `--help`.

pkg/evidence/verifier/crosscheck_test.go
```diff
@@ -0,0 +1,159 @@
...
+	_, err := MaterializeBundle(t.Context(),
+		VerifyOptions{Input: "ghcr.io/owner/repo:v1", AllowUnpinnedTag: true},
+		form, nil)
+	if err == nil {
+		t.Fatalf("expected pull error (no registry), got nil")
+	}
+	if strings.Contains(err.Error(), "tag-only") {
+		t.Errorf("AllowUnpinnedTag should bypass tag-pin check; got %v", err)
+	}
+}
```
mchmarny: This test reaches out to the real internet (`ghcr.io/owner/repo:v1`). With `AllowUnpinnedTag: true` the pin check is bypassed and execution proceeds into `oras.Copy` against the live ghcr.io. Two real problems:

1. **Hostile to offline environments** — sandboxed CI runners and dev workstations behind firewalls will fail with DNS / connect errors which technically satisfies "err != nil" but isn't what the test is asserting.
2. **Worst-case 2-minute hang** — `defaults.EvidenceBundlePushTimeout` is 2m. If the registry responds slowly to an anonymous tag pull on a non-existent owner, the test pays that full timeout per run.

Suggest pointing at a known-unreachable host (e.g., `127.0.0.1:1/repo:v1` or an `httptest.Server` you stop before the call) and/or pass a context with `time.Millisecond * 100` so the failure is fast and deterministic. The assertion you care about — that the error is *not* "tag-only" — works just as well against either.

pkg/evidence/verifier/signature.go
```diff
@@ -0,0 +1,350 @@
...
+	idLit, idRe := "", ".+"
+	if opts.ExpectedIdentityRegexp != "" {
+		idLit, idRe = "", opts.ExpectedIdentityRegexp
+	}
+	identity, err := verify.NewShortCertificateIdentity(issuerLit, issuerRe, idLit, idRe)
+	if err != nil {
+		return verify.CertificateIdentity{}, errors.Wrap(errors.ErrCodeInternal,
+			"failed to build certificate identity matcher", err)
+	}
+	return identity, nil
```
mchmarny, line 183: When the user runs `aicr evidence verify` with neither `--expected-issuer` nor `--expected-identity-regexp` (the default), this builds a matcher that accepts `.+` for both — i.e., *any* Fulcio-issued certificate from *any* OIDC provider. The signature is still cryptographically valid, but the verifier no longer says anything about *who* signed it.

For a tool whose entire purpose is "did the right person attest this recipe," that default is a meaningful foot-gun. The cosign-style convention is to fail closed without pinning, or at minimum emit a `slog.Warn` at the call site so an operator skimming the report sees "no signer policy applied."

A `slog.Warn("signature verified but no signer pinned — any Fulcio identity will pass; consider --expected-issuer/--expected-identity-regexp", ...)` here (or in the success branch of `VerifySignature`) would close the loop without changing default behavior.

pkg/evidence/verifier/fetch.go
```diff
@@ -81,3 +116,265 @@ func hasBundleMarkers(dir string) bool {
...
+		ref = opts.BundleRef
+	}
+	if ref == "" {
+		return nil, errors.New(errors.ErrCodeInvalidRequest,
+			"pointer carries no bundle.oci — re-run with --bundle <oci-ref> or point at the unpacked directory")
+	}
+	// Pointer-driven path: pointer.bundle.digest is the pin. When it
+	// is empty (e.g., a local-only pointer plus --bundle override),
+	// fall through to the direct-OCI digest-pinning rule.
+	requirePin := !opts.AllowUnpinnedTag && att.Bundle.Digest == ""
```
mchmarny, line 159: nit: the function-level comment above says "A tag-only ref is allowed only when the pointer carries a non-empty bundle.digest," but the actual gate is the validator in `pointer.go` which requires `digest: "sha256:..."` (not just non-empty) when `bundle.oci` is set. The implementation is stricter than the prose — would be good to either tighten the comment to match (`bundle.digest must be a sha256 digest`) or have the validator accept any non-empty pin format. PR description has the same drift.

mchmarny (commented):
Nice slice — five-step pipeline reads cleanly, the trust-chain story in the package doc is accurate, the test coverage on the deterministic helpers (DSSE parse, identity matcher, cross-check table, referrer fetcher via in-memory fake) is strong, and gating `--allow-unpinned-tag` behind explicit opt-in is the right default. The pointer-signer cross-check that catches a malicious pointer + valid bundle pairing is a particularly nice touch.

A few things to look at before merge:

- **Medium:** `discoverAndWriteReferrer` errors collapse to `slog.Debug` at the call site, which means a registry that returns a malformed/oversized referrer manifest looks indistinguishable from "unsigned bundle" — a silent downgrade path. Worth distinguishing `ErrCodeNotFound` from real failures.
- **Low/security:** default identity matcher accepts any Fulcio cert (`.+` regex) when no `--expected-*` flags are passed; a `slog.Warn` would surface the missing pin without changing default behavior.
- **Low:** `--allow-unpinned-tag` is implemented but not in the `docs/user/cli-reference.md` flag table.
- **Low:** `TestMaterializeBundle_TagOnlyWithAllowFlag` reaches the real `ghcr.io` and can hang up to 2 minutes on flaky networks.
- **Nit:** comment in `materializeFromPointer` and the PR description say "non-empty bundle.digest" but the validator requires the stricter `sha256:` prefix.

**CI:** the `Build` job (specifically the "Build site (PR)" step) is failing on this head — worth a quick look before requesting re-review. The PR is also `behind` main and will need a rebase.

Nothing blocks merge after the silent-downgrade and pin-warning items are addressed.

pkg/evidence/verifier/fetch.go
njhensley: Fixed. `materializeOCIRef` now uses `stderrors.Is(err, errors.New(errors.ErrCodeNotFound, ""))` to distinguish legitimately-unsigned (debug-log, signature step records Skipped) from any other failure (fail-closed, wrapped `ErrCodeInvalidRequest` with the message "registry returned a malformed Sigstore Bundle referrer"). A MITM serving a multi-layer / oversized / malformed referrer now hits the failure path instead of being silently swallowed.

You're right that the project's "fail closed on evaluator errors" rule applied here — appreciate the catch.

pkg/evidence/verifier/fetch.go
njhensley: Fixed. Added an unexported sentinel `errReferrerFound` (package-level `stderrors.New`); the callback returns it after a successful fetch, and `discoverAndWriteReferrer` unwraps it via `stderrors.Is(cbErr, errReferrerFound)` after `Referrers` returns. The "first match wins, stop pagination" promise the comment makes is now actually delivered — a multi-page response with another sigstore-bundle referrer on page N+1 can no longer overwrite the file written from page N.

docs/user/cli-reference.md
njhensley: Fixed. Added a `--allow-unpinned-tag` row to the flag table with the full description — what it does, why default-off, and the pointer-driven exception (ignored when the pointer carries a sha256 digest). It's the only opt-out for the new tag-pin behavior; reference page is the right place for it.

pkg/evidence/verifier/crosscheck_test.go
njhensley: Fixed. Test now points at `127.0.0.1:1/repo:v1` (reserved port, instant connection refused) and runs under a 250 ms per-test context. Connection failure happens long before the 2-minute pull timeout, and offline CI runners get a deterministic "connection refused" instead of DNS-timeout drift. The assertion is unchanged: error is *not* `tag-only`.

pkg/evidence/verifier/signature.go
njhensley: Fixed. Added a `slog.Warn` in the success branch of `VerifySignature` when both `ExpectedIssuer` and `ExpectedIdentityRegexp` are empty: `"signature verified but no signer pinned — any Fulcio identity will pass"` with the identity / issuer values and a hint pointing at the two `--expected-*` flags.

Doesn't change default behavior (the report still passes), but operators skimming output now see the open-policy footgun. Closing the loop without breaking workflows that legitimately want "any OIDC-bound signer is fine" (e.g., third-party attestations on a public registry) — that's a maintainer-judgment call we leave to them.

pkg/evidence/verifier/fetch.go
njhensley: Fixed. Comment now reads "pointer carries a sha256-prefixed bundle.digest (the validator in pointer.go rejects any other shape when bundle.oci is set) — that digest is the pin." Prose matches the validator's actual rule, and the cross-reference makes it obvious where the strictness lives.

njhensley (commented):
Thanks for the review. All six inline findings addressed; replies on each thread.

**Build site (PR) CI failure** — diagnosed and fixed in the same pass: the cli-reference.md section had a relative link `../../demos/evidence.md` that VitePress flagged as a dead link (the site only mounts `docs/`, not `demos/`). Switched to an absolute GitHub URL matching the convention in `docs/README.md`, `docs/design/00*-*.md`, and `docs/integrator/recipe-development.md`. Local site build now passes (`cd site && npm run build` → `build complete`).

**Rebase onto main** — pending. Will rebase after this fixup commit lands so the rebase only deals with my freshly-committed state. The intervening merges (#886, #888, #889) don't touch `pkg/evidence/verifier` so it should be clean.

mchmarny (approved):

PR #602, opened 2026-04-17
fix(bundler): fix undeploy template pre/post-flight checks

yuanchen8911:
## Summary

Fixes bugs in [`pkg/bundler/deployer/helm/templates/undeploy.sh.tmpl`](https://github.com/NVIDIA/aicr/blob/73f09b9cf8b390298d8f3d19161876b08e93297f/pkg/bundler/deployer/helm/templates/undeploy.sh.tmpl), which generates the bundle `undeploy.sh`, by hardening pre-flight checks, cleanup behavior, and post-flight diagnostics.

## Motivation / Context

The generated `undeploy.sh` had a few real correctness gaps:

- transient `kubectl` / `jq` failures in cleanup could abort the script mid-run under `set -euo pipefail`
- pre-flight could miss some release-owned CRDs, or falsely block safe undeploys for bundle-managed resources that are deleted before controller uninstall
- post-flight did not clearly surface all leftover Helm-owned CRDs after uninstall

This PR fixes those behaviors in the `undeploy.sh` template so newly generated bundles get the corrected undeploy flow.

Fixes: N/A
Related:
- issue `#406`: https://github.com/NVIDIA/aicr/issues/406
- PR `#561`: https://github.com/NVIDIA/aicr/pull/561

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [x] Docs/examples (`docs/`, `examples/`)

### 1. Cleanup pipelines are now best-effort under `set -euo pipefail`

The template now emits warning-and-continue wrappers for cleanup pipelines that were previously able to abort `undeploy.sh` mid-run on a single transient API failure.

That applies to:

- `delete_release_cluster_resources`
- `force_clear_namespace_finalizers`
- the inline Helm-owned orphan-CRD cleanup loop

This keeps undeploy moving into later cleanup phases and post-flight reporting instead of exiting halfway through cleanup.

### 2. Pre-flight CRD discovery is release-scoped and explicit

The final pre-flight design intentionally stays narrow. The template now checks CRDs from only these sources:

- CRDs present in `helm get manifest`
- CRDs annotated to the Helm release via `meta.helm.sh/release-*`
[...]

PR #561, opened 2026-04-13
fix(bundler): add pre-flight finalizer check to undeploy.sh (#406)

lockwobr:
## Summary

Add a pre-flight check to `undeploy.sh` that detects custom resources with active finalizers on CRDs owned by bundle operators. Warns the
user and exits before teardown to prevent the unrecoverable hang described in #406.

## Motivation / Context

`undeploy.sh` hangs indefinitely when CRs have finalizers that require their controller to process, but the controller has already been
uninstalled. The deadlock: `delete_release_cluster_resources()` deletes CRDs without `--wait=false` (line 81), the CRD's
`customresourcecleanup` finalizer blocks until all CR instances are gone, but CRs in user namespaces have controller-managed finalizers that
  can never be processed.

Fixes: #406
Related: N/A

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] CLI (`cmd/aicr`, `pkg/cli`)
- [ ] API server (`cmd/aicrd`, `pkg/api`, `pkg/server`)
- [ ] Recipe engine / data (`pkg/recipe`)
- [x] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [ ] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [ ] Validator (`pkg/validator`)
- [ ] Core libraries (`pkg/errors`, `pkg/k8s`)
- [ ] Docs/examples (`docs/`, `examples/`)
- [ ] Other: ____________

## Implementation Notes

**Pre-flight check** (`check_release_for_stuck_crds` / `check_crd_for_stuck_resources`):
- Uses `helm get manifest` to discover CRDs owned by each Helm component (generic — no hardcoded CRD group list needed)
- For each CRD, queries all namespaces for CR instances with non-empty `.metadata.finalizers`
- If found: lists blocking objects with their finalizer names, then exits with instructions
- Silently skips releases that aren't installed (`helm get manifest` failure → `return 0`)
- `--skip-preflight` flag to bypass when the user knows what they're doing

**Component manifest**: Lists all components that will be removed (with type tags: `[helm]`, `[kustomize]`, `[manifests]`) so users can see
exactly what undeploy will do — important for shared components like cert-manager.

[...]

PR #416, opened 2026-03-16
fix(bundler): clean up orphaned KAI and Kubeflow Trainer CRDs on undeploy

yuanchen8911:
## Summary

Fix undeploy.sh to clean up orphaned CRDs from KAI scheduler and Kubeflow Trainer that survive Helm uninstall.

## Motivation / Context

After `undeploy.sh` completes, 6 orphaned CRDs remain on the cluster:
- `configs.kai.scheduler`, `schedulingshards.kai.scheduler`, `topologies.kai.scheduler` — created by the KAI operator at runtime, not by the Helm chart
- `trainjobs.trainer.kubeflow.org`, `trainingruntimes.trainer.kubeflow.org`, `clustertrainingruntimes.trainer.kubeflow.org` — installed by the validator during conformance/performance tests

Neither set carries Helm labels, so the existing `delete_release_cluster_resources` label-based sweep misses them.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Component(s) Affected

- [x] Bundlers (`pkg/bundler`, `pkg/component/*`)

## Implementation Notes

- Added explicit CRD cleanup in `undeploy.sh.tmpl` after namespace deletion, targeting `kai.scheduler` and `trainer.kubeflow.org` API groups
- Added orphaned CRD detection in post-flight verification
- Approach uses API group matching (not hardcoded CRD names) so new CRDs in these groups are automatically covered

## Testing

```bash
make test  # All bundler tests pass
```

Verified generated undeploy.sh includes the new cleanup steps.

## Risk Assessment

- [x] **Low** — Isolated change, well-tested, easy to revert

**Rollout notes:** Existing bundles won't have the fix — only newly generated bundles. Users with stale CRDs can manually delete them with `kubectl delete crd <name>`.

## Checklist

- [x] Tests pass locally (`make test` with `-race`)
- [x] Linter passes (`make lint`)
- [x] I did not skip/disable tests to make CI green
- [x] I added/updated tests for new functionality
- [x] I updated docs if user-facing behavior changed
- [x] Changes follow existing patterns in the codebase
- [x] Commits are cryptographically signed (`git commit -S`)

PR #253, opened 2026-02-28
fix(bundler): improve deploy/undeploy script reliability

yuanchen8911:
## Summary

Fix deploy and undeploy script templates to prevent deadlocks, handle interrupted deploys, clean up orphaned cluster-scoped resources, and respect --best-effort semantics.

## Motivation / Context

Multiple issues discovered during dynamo inference deployment on EKS:

1. **deploy.sh deadlock**: `helm install --wait` blocks on pods that need pre-existing resources (e.g., gpu-operator dcgm-exporter ConfigMap applied *after* helm completes)
2. **undeploy.sh hangs**: Slow helm cleanup hooks (gpu-operator, skyhook-operator) block indefinitely; stale webhooks and CRDs survive namespace deletion and block subsequent deployments
3. **Interrupted deploys**: Helm releases left in `pending-install` state can't be cleaned up by normal `helm uninstall`
4. **Orphaned webhooks**: Operator-created webhooks (kai-scheduler admission) without Helm labels block pod creation after their service namespace is deleted
5. **--best-effort not honored**: Namespace creation and manifest apply failures exit immediately instead of continuing

Fixes: N/A
Related: N/A

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)

## Component(s) Affected

- [x] Bundlers (`pkg/bundler`, `pkg/component/*`)

## Implementation Notes

**deploy.sh.tmpl:**
- Apply component manifests both before and after helm install. Pre-install handles ConfigMaps (e.g., gpu-operator dcgm-exporter), post-install handles CRD-dependent resources (e.g., kai-scheduler Config CR patch). Pre-install filters CRD-missing errors specifically.
- All per-component steps (namespace create, manifest apply) respect `--best-effort` via `|| helm_failed`

**undeploy.sh.tmpl:**
- `helm_force_uninstall()`: tries normal uninstall first, retries with `--no-hooks` if it fails (handles pending-install state without fragile status parsing)
- `--timeout` flag (default 120s) with input validation
- `delete_release_cluster_resources()`: deletes Helm-labeled webhooks and CRDs after each component uninstall
- `delete_orphaned_webhooks_for_ns()`: finds webhooks referencing services in the component namespace, deletes only if namespace or service explicitly returns NotFound (skips transient API errors). Scoped to component namespace to avoid touching unrelated platform webhooks.
- All jq usage guarded behind availability check with graceful fallback
[...]

PR #899, opened 2026-05-14
feat(bundler): add helmfile deployer (#632)

### LINKED ISSUE #632: [Feature]: Helmfile deployer for declarative, bash-free deployments
## Feature Summary

Add `--deployer helmfile` alongside the existing `helm`, `argocd`, and `argocdhelm` deployers. The bundler emits a `helmfile.yaml` describing the component release graph; operators run `helmfile apply` / `helmfile diff` / `helmfile destroy` to drive rollouts.

## Problem / Use Case

When the bundler needed a deploy/undeploy story, the project deliberately chose **generated bash scripts over adding a deployment-tool dependency.** The reasoning was that the scripts would stay simple enough that introducing another tool (helmfile, argo-rollouts, etc.) would be unnecessary overhead. That was a reasonable call at the time.

The evidence since suggests the bet didn't hold:

- `deploy.sh.tmpl` is ~450 lines and `undeploy.sh.tmpl` is ~700 lines — together encoding orchestration logic (finalizer races, stale hook cleanup, webhook orphan detection, operator-specific post-install, pre/post-flight) that isn't really "config generation."
- Seven PRs on `undeploy.sh.tmpl` alone have chased finalizer-handling edge cases (#253, #282, #364, #416, #477, #561, #602 — see #610 for context).
- The bash keeps growing because orchestration is genuinely hard, and the template is where that complexity lands.

AICR's charter is that it **generates validated configuration; it is not a deployment tool.** The existing `helm` deployer has quietly drifted past that line. A helmfile deployer is the circle-back: emit the release graph as **data** (`helmfile.yaml`) and delegate orchestration to a tool purpose-built for it.

**This is not a deprecation proposal.** The `helm` deployer stays in place; this issue adds a parallel option — the two can coexist indefinitely. But if the helmfile deployer is well received, it opens a future conversation about simplifying or retiring the bash deployer. That's a separate decision for a later issue, not a commitment this one makes.

### Bonus: what users get beyond generated bash

On top of the scope-alignment case, users pick up capabilities the bash deployer doesn't offer:

- **Readability** — a `helmfile.yaml` is easier to scan, review, and reason about than a ~450-line templated bash script.
- **`helmfile diff`** — preview changes before apply. No AICR deployer offers this today.
- **Idempotent apply and destroy** — safe to re-run after partial failure; generated bash isn't.
- **Reviewable release graph** — `needs:` dependencies and per-release flags in YAML, instead of ordering baked into bash template iteration.
- **Standard tool in the ecosystem** — teams already running helmfile in their CI get a one-tool workflow instead of a bundle-plus-custom-bash hybrid.
- **Parallel installs** where `needs:` allows; bash is sequential.

## Proposed Solution

A new `helmfile` deployer implementing AICR's existing `Deployer` interface. The bundler emits:

- `helmfile.yaml` — declarative release graph
- Per-component values files (reuses existing bundler output)

The deployer lives entirely within helmfile's native lifecycle. No bash wrapper, no pre-flight logic, no extra scripting layered on top. Operators drive helmfile directly (`helmfile apply` / `diff` / `destroy`) using the tool's standard workflow — the same workflow the helmfile community already uses.

`helmfile.yaml` maps cleanly from AICR's current deploy logic:

| Current deploy.sh behavior | Helmfile equivalent |
|---|---|
| Ordered install across components | `needs:` dependency graph |
| `ASYNC_COMPONENTS` skipping `--wait` | per-release `wait: false` |
| `COMPONENT_HELM_TIMEOUT` overrides | per-release `timeout:` |
| Component values.yaml + cluster-values.yaml | `values:` list |

## Success Criteria

- `aicr bundle --deployer helmfile -r recipe.yaml -o ./bundle` produces a working `helmfile.yaml`.
- `helmfile apply` and `helmfile destroy` succeed end-to-end against Kind/KWOK for at least one recipe.
- `helmfile diff` shows expected changes for recipe edits.
- Existing `helm`, `argocd`, `argocdhelm` deployers continue to work unchanged.
- Docs updated in `docs/user/cli-reference.md`.

## Sequencing Considerations

This issue is independently valuable, but sequencing with related bundler work pays off:

- **#610 (bridge Jobs for deploy/undeploy lifecycle)** — without it, `helmfile destroy` hangs on the same finalizer race as `helm uninstall`. Strongly prefer landing #610 first so the deployer's uninstall path works from day one.
- **#516 (local charts)** — uniform local-chart bundles make `helmfile.yaml` trivially point at `./001-component/` paths. Without it, `helmfile.yaml` references remote repos (works, but misses #516's air-gap goal).

Neither is a hard block. This deployer could ship against today's bundle format and today's finalizer handling, then gain polish as #610/#516 land.

## Hints for the Implementer

Design decisions intentionally left open — these are starting points, not prescriptions:

- **No pre-flight emitted.** Helmfile has native lifecycle hooks (`prepare:`, `presync:`, `postsync:`, `postuninstall:`) that could carry pre-flight logic. Per Non-Goals, this deployer intentionally ships without any. The hooks remain available if a future need justifies them — start with nothing shipped.
- **Kustomize / raw manifest components.** Helmfile is Helm-first. If #516 ships local charts first, this concern disappears. If not, the `chartify` project (helmfile org) handles the conversion.
- **helm-secrets plugin.** Opens a future door for SOPS-encrypted values files. Not required for MVP, but worth keeping the deployer design compatible.
- **Concurrency.** Helmfile installs in parallel where `needs:` allows. Declaring `needs:` correctly preserves current sequential behavior; concurrency becomes a later opt-in.
- **Tool install.** Helmfile becomes a new deploy-time dependency for users of this deployer. Document in cli-reference; consider pinning via `make tools-setup`.
- **Wiring points.** The deployer lives at `pkg/bundler/deployer/helmfile/`, implements `Deployer` from `pkg/bundler/deployer/deployer.go`, and is wired into `buildDeployer()` in `pkg/bundler/bundler.go` via a new `config.DeployerHelmfile` case in `pkg/bundler/config/config.go:ParseDeployerType`.

## Alternatives Considered

- **Do nothing; users keep wrapping bash.** Works but reinvents helmfile per team.
- **Push users toward ArgoCD.** Forces a controller into every target cluster — wrong fit for CI / ephemeral / air-gapped use cases.
- **Replace the helm deployer with helmfile.** Drops the "bash I can read" posture some users rely on. Additive is better than replacement.

## Non-Goals

**Guiding principle:** the deployer lives entirely within helmfile's native lifecycle (`apply`, `diff`, `destroy`, its own `hooks:`). That lifecycle is what the helmfile community expects and what an operator picking up this deployer would assume. AICR doesn't extend or wrap it. If AICR developers ever need deeper operations (cluster scrubbing, debug workflows, scorched-earth cleanup), those belong in `tools/` — separate from what the bundler emits.

Three concrete consequences:

- **No `deploy.sh` / `undeploy.sh` wrapper.** Shipping one would reintroduce the shell-script accretion pattern this issue is trying to leave behind.

- **No pre-flight checks emitted.** AICR's bash `deploy.sh` includes ~120 lines of pre-flight today (terminating namespaces, stale webhooks, orphaned API services, orphaned CRDs, stale node taints). Helmfile has `prepare:` hooks that could carry equivalent logic, but shipping any reintroduces the same accretion problem. Cluster pre-flight is the operator's concern.

- **No replicating today's `undeploy.sh` cleanup behavior.** CRD deletion, finalizer force-clearing, orphaned webhook sweep, post-flight verification — this arguably doesn't belong in an AICR-emitted undeploy path at all. `helmfile destroy` stays the simple graceful-uninstall primitive helmfile already provides. The broader "leave no trace" cleanup is a separate problem for a follow-up ticket. Two plausible directions there:
  - A `tools/nuke` script operators opt into when they explicitly want a scorched-earth clean.
  - The bundle-time opt-in `cleanup.yaml` bridge-Job pattern from #610, which resolves the finalizer race declaratively.

## Component

Bundlers

## Priority

Additive — existing deployers unchanged. New `--deployer helmfile` flag value; no migration impact for current users.

## Operational Considerations

- New runtime dependency on the `helmfile` binary for users of this deployer.
- Helmfile tracks release state separately from Helm by default (reconcilable from the cluster); worth documenting semantics.
- Air-gap posture improves if paired with #516; neutral otherwise.

Maybe, with guidance

lockwobr:
## Summary

Adds `--deployer helmfile` alongside `helm` / `argocd` / `argocd-helm` / `flux`. The bundler emits a declarative `helmfile.yaml` release graph + per-component chart dirs (via the shared `localformat` writer); operators drive rollouts with the upstream `helmfile` CLI (`apply` / `diff` / `destroy`). No bash wrapper, no pre-flight, no scripted post-install — AICR's charter is to generate validated configuration, not orchestrate deployments.

## Motivation / Context

`deploy.sh.tmpl` (~450 lines) and `undeploy.sh.tmpl` (~700 lines) have accreted finalizer-handling, stale-hook cleanup, webhook-orphan detection, and per-operator post-install logic. Seven PRs on `undeploy.sh.tmpl` alone have chased finalizer edge cases (#253, #282, #364, #416, #477, #561, #602). That's orchestration logic, and it belongs in a tool purpose-built or it — not in a bash template AICR maintains.

This change is **additive**: the `helm` deployer stays in place; the two coexist indefinitely. If the helmfile path is well received it opens a future conversation about simplifying the bash deployer, but that's a separate decision for a later issue.

Beyond scope alignment, users pick up `helmfile diff` (no AICR deployer offers preview today), idempotent apply/destroy, a reviewable release graph (`needs:` in YAML instead of bash iteration order).

closes #632
## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Implementation Notes

**Mapping from AICR concepts to helmfile primitives:**

| AICR concept | Helmfile mapping |
|---|---|
| `DeploymentOrder` | `release.needs` (immediate predecessor) |
| Chart repo (HTTP/OCI) | `repositories[]`, deduped by URL |
| `DynamicValues[name]` | `cluster-values.yaml` appended to release `values:` |
| Global `HELM_TIMEOUT="10m"` | `helmDefaults.timeout: 600` |
| `ASYNC_COMPONENTS` (kai-scheduler) | per-release `wait: false` |
| `kai-scheduler → 20m` | per-release `timeout: 1200` |
| Manifest-only / Kustomize / mixed-post | `chart: ./NNN-name` (localformat wraps as `KindLocalHelm`) |
| `VendorCharts` | localformat pulls tarball into `./NNN-name/charts/`; release still `chart: ./NNN-name` |

**Reuses existing infrastructure:**

- `localformat.Write()` emits the uniform `NNN-<name>/` per-component layout. The helmfile deployer doesn't re-classify components; it walks the returned `Folder` list and emits one release per folder.
- `deployer.SortComponentRefsByDeploymentOrder` for ordering.
| `ASYNC_COMPONENTS` (kai-scheduler) | per-release `wait: false` |
| `kai-scheduler → 20m` | per-release `timeout: 1200` |

**Non-goals (per #632):**

- No `deploy.sh` / `undeploy.sh` emitted alongside `helmfile.yaml`.
- No `prepare:` / `presync:` / `postsync:` / `postuninstall:` hooks emitted.
- No replication of today's `undeploy.sh` CRD/finalizer/webhook cleanup — `helmfile destroy` stays the simple graceful-uninstall primitive.

## Testing

```bash
make qualify
```

## Risk Assessment

**Rollout notes:** 

## Checklist


tools/setup-tools
```diff
@@ -390,6 +392,39 @@ if [[ "${SKIP_TOOLS}" == "false" ]]; then
...
+                sudo mv "${HELMFILE_TMP}/helmfile" /usr/local/bin/helmfile
+                sudo chmod +x /usr/local/bin/helmfile
+                rm -rf "${HELMFILE_TMP}"
+                log_success "helmfile installed"
+            else
+                log_error "Failed to install helmfile. Please install manually."
+            fi
+        fi
+    fi
+
```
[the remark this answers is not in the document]
lockwobr: NO, none of the others do this. Leaving it a lolne.

pkg/bundler/deployer/helmfile/helmfile_test.go
```diff
@@ -0,0 +1,464 @@
...
+// input.
+func TestGenerate_NilRecipeResult(t *testing.T) {
+	g := &Generator{Version: testBundlerVersion}
+	_, err := g.Generate(context.Background(), t.TempDir())
+	if err == nil {
+		t.Fatalf("Generate() with nil RecipeResult expected error, got nil")
+	}
+	if !strings.Contains(err.Error(), "RecipeResult is required") {
+		t.Fatalf("Generate() error = %q, want substring %q",
+			err.Error(), "RecipeResult is required")
```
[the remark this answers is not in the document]
lockwobr: Addressed. `helmfile_test.go` now uses `stderrors.Is(err, errors.New(<code>, ""))` paired with a secondary substring check on the message:

- L165 — `ErrCodeInvalidRequest` (Generate with nil RecipeResult)
- L246 — `ErrCodeTimeout` (canceled context)
- L313 — `ErrCodeInternal` (buildHelmfile with nil Upstream)
- L332 — `ErrCodeInvalidRequest` (unsupported folder kind)

The production sites were also fixed: `releases.go` now returns `errors.New(<code>, ...)` instead of `fmt.Errorf` for both the nil-Upstream and unsupported-kind paths, and `helmfile.go` uses `errors.PropagateOrWrap` when re-wrapping `buildHelmfile` results so the inner `ErrCodeInvalidRequest`/`ErrCodeInternal` codes survive.

pkg/bundler/deployer/helmfile/releases.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// componentOverrides mirrors the special cases hardcoded in
+// pkg/bundler/deployer/helm/templates/deploy.sh.tmpl
+// (ASYNC_COMPONENTS and the COMPONENT_HELM_TIMEOUT case block).
+// The two files must be updated together; promoting these into the recipe
+// schema is tracked as a follow-up to issue #632.
+var componentOverrides = map[string]overrides{
+	"kai-scheduler": {wait: false, timeout: 20 * 60},
+}
```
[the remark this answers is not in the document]
lockwobr: Addressed with an automated parity check rather than the heavy lift of unifying the two sites.

Added `TestComponentOverrides_ParityWithHelmDeployScript` in `pkg/bundler/deployer/helmfile/helmfile_test.go` that reads `pkg/bundler/deployer/helm/templates/deploy.sh.tmpl` and asserts both directions:

- Every entry in `componentOverrides` must appear in `ASYNC_COMPONENTS="…"` (when `wait: false`) and in the `COMPONENT_HELM_TIMEOUT` case block (when `timeout != 0`), with matching duration.
- Every name in either deploy.sh.tmpl construct must exist in `componentOverrides`.

Either side drifting (kai-scheduler timeout changes, a new async component added, an entry removed) now fails CI with a message telling the author to update both places together. The error strings name both sources of truth to make the next step obvious.

Unifying into a single shared registry is still the right long-term fix and is tracked as a follow-up to #632, but this guard closes the silent-drift risk without requiring schema work.

.github/actions/load-versions/action.yml
```diff
@@ -178,6 +190,7 @@ runs:
         echo "  ctlptl: ${{ steps.versions.outputs.ctlptl }}"
         echo "  tilt: ${{ steps.versions.outputs.tilt }}"
         echo "  helm: ${{ steps.versions.outputs.helm }}"
+        echo "  helmfile: ${{ steps.versions.outputs.helmfile }}"
```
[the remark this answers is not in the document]
lockwobr: Done. Added both `helmfile_sha256_linux_amd64` and `helmfile_sha256_linux_arm64` to the action's logging output, matching the chainsaw pattern on the same file.

pkg/bundler/bundler_test.go
```diff
@@ -759,6 +759,79 @@ func TestMake_ArgoCD(t *testing.T) {
...
+		t.Errorf("deployment type = %q, want %q",
+			output.Deployment.Type, "Helmfile release graph")
+	}
+	if output.TotalFiles == 0 {
+		t.Error("expected generated files")
+	}
+	// Sanity-check the deployer emitted helmfile.yaml at the bundle root.
+	if _, statErr := os.Stat(filepath.Join(tmpDir, "helmfile.yaml")); statErr != nil {
+		t.Errorf("helmfile.yaml missing at bundle root: %v", statErr)
+	}
```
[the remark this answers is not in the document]
lockwobr: Done — and went a bit further than the suggestion. `TestMake_Helmfile` now asserts the bundle root does NOT contain any peer-deployer artifact:

- `deploy.sh` / `undeploy.sh` (helm deployer)
- `app-of-apps.yaml` (argocd / argocd-helm deployer)
- `kustomization.yaml` (flux deployer)

Uses `stderrors.Is(statErr, os.ErrNotExist)` since the file already aliases stdlib `errors` as `stderrors`.

pkg/bundler/deployer/helmfile/helmfile_test.go
```diff
@@ -0,0 +1,559 @@
...
+		),
+		ComponentValues: map[string]map[string]any{"cert-manager": {}},
+		Version:         testBundlerVersion,
+		DataFiles:       []string{"does-not-exist.yaml"},
+	}
+	_, err := g.Generate(context.Background(), t.TempDir())
+	if err == nil {
+		t.Fatalf("Generate() with missing data file expected error, got nil")
+	}
+}
```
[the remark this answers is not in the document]
lockwobr: Done. The five scenario tests are now a single `TestGenerate_Scenarios` with a table of `{name, gen, goldens}` entries and a `t.Run(tc.name, …)` per row. Each row keeps its inline rationale (what code path the scenario exercises) so the subtest list still reads well from `go test -v`. The other tests (`TestGenerate_EmptyRecipe`, `TestGenerate_NilRecipeResult`, `TestGenerate_WithChecksums`, `TestGenerate_WithDataFiles`, `TestGenerate_ContextCanceled`, `TestGenerate_MissingDataFile`) stay separate as you suggested — they assert error codes / side-effects rather than golden files and don't fit the `runScenario` shape.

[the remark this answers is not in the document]
lockwobr: Done. `TestGenerate_MissingDataFile` now asserts `stderrors.Is(err, errors.New(errors.ErrCodeInternal, ""))` — `output.AddDataFiles` wraps the `os.Stat` failure as `ErrCodeInternal`, so a regression that surfaces an uncoded error there fails the test.

pkg/bundler/deployer/helmfile/releases.go
```diff
@@ -0,0 +1,229 @@
...
+	var prevName string
+	for _, f := range folders {
+		rel := Release{
+			Name:      f.Name,
+			Namespace: namespaceByComponent[f.Parent],
+			Values:    valuesFilesForFolder(f, dynamicValues[f.Parent]),
+		}
+		if prevName != "" {
+			rel.Needs = []string{prevName}
+		}
```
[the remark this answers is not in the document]
lockwobr: Pushing back on this one — the linear chain is the right behavior for the recipe contract we have today, not a defect.

The recipe carries `DeploymentOrder` as a flat `[]string`. There is no edge/tier/parallelism information in the schema (`pkg/recipe/recipe.go::RecipeResult`). The existing deployers all preserve that linear semantics:

- `helm` deployer: `deploy.sh.tmpl` iterates `Components` in order, with no parallelism.
- `flux` deployer: `pkg/bundler/deployer/flux/flux.go::buildDependsOn` returns `dependsOn` pointing at the immediate predecessor (the exact pattern this PR mirrors).
- `argocd` / `argocd-helm`: assigns sync-waves 0, 1, 2, … with no concept of "same wave".

If this PR opted helmfile out of `needs:` to parallelize, the resulting bundle would not match the install order the recipe author specified — a real regression for anyone relying on (for example) cert-manager landing before kube-prometheus-stack. Helmfile would happily install them concurrently and the prometheus webhooks would race.

Expressing genuine parallelism is a recipe-schema change (DAG / tiers / explicit `dependsOn` per component) and applies equally to flux/argocd/argocd-helm. That's out of scope for #632 and worth its own ticket once we know what shape the schema should take. I'd rather ship the parity-with-other-deployers behavior here and revisit parallelism uniformly across all four deployers.

pkg/bundler/deployer/helmfile/releases.go
```diff
@@ -0,0 +1,229 @@
...
+				rel.Wait = &waitFalse
+			}
+			if ov.timeout > 0 {
+				rel.Timeout = ov.timeout
+			}
+		}
+
+		releases = append(releases, rel)
+		prevName = rel.Name
+	}
```
mchmarny, line 161: The PR description sells "parallel installs where `needs:` allows" as a benefit over the bash deployer, but `buildHelmfile` always chains every release to its immediate predecessor (`rel.Needs = []string{prevName}`). The emitted graph is strictly linear — helmfile has no opportunity to parallelize. This mirrors the helm deployer's sequential `deploy.sh` iteration, which is fine for parity, but either tighten the PR description (drop the parallelism claim) or wire `needs:` from the recipe's real dependency edges so independent components can run concurrently.

pkg/bundler/deployer/helmfile/releases.go
```diff
@@ -0,0 +1,229 @@
...
+// Paths are written with an explicit "./" prefix to match the chart: form
+// used for KindLocalHelm folders. Helmfile accepts both forms; the prefix
+// is purely a stylistic choice for visual consistency in golden files.
+func valuesFilesForFolder(f localformat.Folder, dynamicPaths []string) []string {
+	values := []string{"./" + f.Dir + "/values.yaml"}
+	if len(dynamicPaths) > 0 {
+		values = append(values, "./"+f.Dir+"/cluster-values.yaml")
+	}
+	return values
+}
```
mchmarny, line 212: `cluster-values.yaml` is appended for every folder whose parent has dynamic paths — including injected `-pre` / `-post` folders. `localformat.writeLocalHelmFolder` does write a `cluster-values.yaml` into those auxiliary folders (split from the parent's values), so the file exists and helmfile resolves it, but the dynamic keys are typically only meaningful to the primary chart, not to the wrapped raw-manifest releases. Not strictly broken, just a behavior worth a one-line comment near `valuesFilesForFolder` so a future reader doesn't think the parent-keyed lookup is accidental.

tests/chainsaw/cli/bundle-helmfile/chainsaw-test.yaml
```diff
@@ -0,0 +1,382 @@
...
+              fi
+              cd "${WORK}/basic"
+              helmfile build >/dev/null \
+                || { echo "helmfile build failed" >&2; exit 1; }
+            check:
+              ($error == null): true
+      cleanup:
+        - script:
+            content: |
+              rm -rf /tmp/chainsaw-bundle-helmfile
```
mchmarny: nit: `cleanup:` is attached only to the last step (`helmfile-build-validates`). Chainsaw step-level cleanup runs after that step's `try`, so if any earlier step fails, `/tmp/chainsaw-bundle-helmfile` is left behind between runs. The setup step does `rm -rf "${WORK}"` so this self-heals on next run, but moving the cleanup to step 1 (or to a top-level `catch:`) makes the test idempotent under partial failure.

tests/chainsaw/cli/bundle-helmfile/chainsaw-test.yaml
```diff
@@ -0,0 +1,382 @@
...
+              HF="${WORK}/basic/helmfile.yaml"
+              # Extract every release name in declaration order.
+              names=$(awk '
+                /^releases:/ {in_releases=1; next}
+                in_releases && /^  - name: / {print $3}
+              ' "${HF}")
+              [ -n "$names" ] || { echo "no releases parsed" >&2; exit 1; }
+              count=$(printf "%s\n" "$names" | wc -l | tr -d ' ')
+              [ "$count" -ge 2 ] \
+                || { echo "need at least 2 releases, got $count" >&2; exit 1; }
```
mchmarny: nit: hard-coding `count >= 2` makes this test brittle against future recipe-content changes — if the eks/h100/training recipe ever ends up with a single component, this fails for the wrong reason. Consider degrading to `[ "$count" -ge 1 ]` and skipping the chain-walk when only one release exists.

pkg/bundler/deployer/helmfile/helmfile.go
```diff
@@ -0,0 +1,329 @@
...
+	if g.RecipeResult == nil {
+		return nil, errors.New(errors.ErrCodeInvalidRequest, "RecipeResult is required")
+	}
+	if err := ctx.Err(); err != nil {
+		return nil, errors.Wrap(errors.ErrCodeTimeout, "context canceled before generation", err)
+	}
+
+	output := &deployer.Output{Files: make([]string, 0)}
+
+	if err := os.MkdirAll(outputDir, 0755); err != nil {
```
mchmarny: nit: `0755` here and `0600` on line 287 use the decimal-leading-zero form. Most of the new code in this package and in `pkg/bundler/deployer/localformat` uses the explicit `0o755` / `0o600` Go octal literal. Trivial — flagged only because the rest of the package is consistent.

pkg/bundler/deployer/helmfile/helmfile_test.go
```diff
@@ -0,0 +1,596 @@
...
+				"(helmfile bundles would --wait on a release the helm deployer treats as async)", name)
+		}
+	}
+	for name := range scriptTimeouts {
+		if _, ok := componentOverrides[name]; !ok {
+			t.Errorf("deploy.sh.tmpl has COMPONENT_HELM_TIMEOUT case for %q but componentOverrides has no entry "+
+				"(helmfile bundles would use the global timeout for a release the helm deployer special-cases)", name)
+		}
+	}
+}
```
mchmarny, line 538: Nice. This parity test is the load-bearing piece that keeps the two deployers honest about kai-scheduler's async/timeout semantics — without it, deploy.sh and helmfile.yaml would silently drift the first time anyone touches either side. Both directions (Go-map → script, script → Go-map) are checked. Worth keeping in mind when the follow-up to #632 promotes these to the recipe schema: this test should fail loudly once the duplication is removed.

tools/update-helmfile-checksums
```diff
@@ -0,0 +1,137 @@
...
+  if awk -v sha="$sha" '
+        /^  chainsaw_checksums:[[:space:]]*$/ { in_block = 1; next }
+        in_block && /^  [a-zA-Z_]/ && !/^    / { in_block = 0 }
+        in_block && index($0, sha) > 0 { found = 1 }
+        END { exit !found }
+      ' "${NEW_FILE}"; then
+    echo "ERROR: helmfile SHA ${sha} leaked into chainsaw_checksums block" >&2
+    exit 1
+  fi
+done
```
mchmarny, line 134: Solid defensive shell — `set -euo pipefail`, awk-scoped rewrite over yq to avoid full-file reformatting, post-rewrite verify that the four target lines actually changed, and the cross-block contamination check against `chainsaw_checksums`. The comment on lines 121-123 documents the latter as paranoia, but it's exactly the kind of regression-by-copy-paste this catches. Good shape for a Renovate post-upgrade task.

mchmarny (commented):
Solid additive change — the helmfile deployer slots into the existing `localformat.Write()` infrastructure cleanly, reuses `SortComponentRefsByDeploymentOrder`, and the golden-file scenarios cover the four shapes that actually matter (upstream-only, dynamic-values, kai-scheduler async, mixed + manifest-only). The `TestComponentOverrides_ParityWithHelmDeployScript` regression guard is the right call — it locks the helmfile and helm deployers together until the special cases move into the recipe schema.

CI is green where it has reported (Lint, Test, E2E, CLI E2E, GPU smoke, GPU inference, analyze, malware/grype/ClamAV); the GPU Training Test was still in progress at review time. CodeQL is `neutral` — confirm that's the expected outcome and not a regressed config.

Feedback is all low/nit-grade — nothing blocks merge:
- The PR description's "parallel installs where `needs:` allows" claim doesn't match the implementation (every release chains to its predecessor, no fan-out).
- `valuesFilesForFolder` references `cluster-values.yaml` for injected `-pre` / `-post` folders too — works, but worth a code comment.
- Chainsaw cleanup is attached only to the last step.
- One brittle `count >= 2` assertion and a couple of `0755` vs `0o755` style nits.

Nice work overall.

pkg/bundler/deployer/helmfile/releases.go
lockwobr: Agreed — the linear chain is parity-with-helm-deployer, not parallel-where-possible. The recipe only carries a flat `DeploymentOrder`; encoding real DAG edges is a recipe-schema change that affects all four deployers (helm, argocd, argocd-helm, flux all preserve the same linear semantics today).

Will tighten the PR description to drop the "parallel installs where `needs:` allows" line. The bullet stays accurate as a *future capability* once the recipe gains DAG edges, but it's misleading as written about what this PR delivers.

pkg/bundler/deployer/helmfile/releases.go
lockwobr: Good call. Added a comment to `valuesFilesForFolder` documenting the rationale: `localformat.writeLocalHelmFolder` propagates the parent's split `cluster-values.yaml` into every -pre / -post auxiliary folder, so the wrapped releases reference the same file by design. The dynamic keys are usually only meaningful to the primary chart, but referencing the file uniformly keeps a partial-cluster-values edit from being silently ignored by the wrapped releases. Comment is on the helper that materializes this so it sits where the next reader will look.

tests/chainsaw/cli/bundle-helmfile/chainsaw-test.yaml
lockwobr: Done. Split into two pieces: a spec-level `catch:` that always fires on step failure, plus a final `teardown` step for the happy path. The setup step's `rm -rf` stays as the self-heal on the next run. /tmp/chainsaw-bundle-helmfile is now cleaned on success (teardown), on failure (catch), and on rerun (setup) — no path leaves it behind.

tests/chainsaw/cli/bundle-helmfile/chainsaw-test.yaml
lockwobr: Done — `count >= 2` degraded to `count >= 1`. Comment added inline: the chain-walk loop naturally short-circuits when count==1 (only the release[0] no-needs guard runs), so the test still validates everything it can, but stops failing for the wrong reason if the recipe ever pares down to a single component.

pkg/bundler/deployer/helmfile/helmfile.go
lockwobr: Fixed both — `0755` → `0o755` (MkdirAll), `0600` → `0o600` (WriteFile). Matches the rest of the package and `pkg/bundler/deployer/localformat`.

pkg/bundler/deployer/helmfile/helmfile_test.go
lockwobr: Thanks — exactly the intent. When the schema work lands and `componentOverrides` goes away, the failure mode will be "this test no longer compiles" rather than silent drift, which forces the cleanup work to happen in the same PR. Will leave a note on the follow-up issue pointing at this test as the canary.

tools/update-helmfile-checksums
lockwobr: Thanks. The block-scoped awk was the second pass — the first cut used a global sed that promptly stomped `chainsaw_checksums` (both blocks have the same arch keys at the same indent), so the cross-block guard is paranoia born of actually shipping the bug locally. Will let Renovate find out for real when v1.5.2 lands.

pkg/bundler/deployer/helmfile/releases.go
lockwobr: Yea, let me update the PR description. I kept it linear.

mchmarny (approved):

Issue #874, opened 2026-05-13
validator orchestrator writes recipe.Validation but :edge expects v1.ValidationInput (constraints silently lost)


njhensley:
## Summary

When `aicr validate` runs against the `:edge` validator images (e.g. `AICR_VALIDATOR_IMAGE_TAG=edge`), the orchestrator and validator disagree about the YAML shape of the validation payload. As a result:

- Phase constraints declared in `validation.performance.constraints` and `validation.deployment.constraints` are invisible to the validators that need them.
- `nccl-all-reduce-bw-net` / `nccl-all-reduce-bw-nvls` report **skipped** with `no <name> constraint in recipe`.
- `gpu-operator-version` reports **passed** but silently bypasses the recipe's version constraint (`return nil` path in `gpu_operator_version.go` when the lookup misses).

The same flow on `:latest` works because `:latest` is from before #830 and reads the legacy `ctx.Recipe.*` paths.

## Reproduction

On any recipe whose `validation.performance.constraints` and/or `validation.deployment.constraints` are populated (e.g. a GB200/EKS training recipe with explicit `Deployment.gpu-operator.version` and `nccl-all-reduce-bw-net` constraints):

```bash
AICR_VALIDATOR_IMAGE_TAG=edge aicr validate \
  --recipe recipe.yaml \
  --emit-attestation ./out \
  --no-cleanup \
  --timeout 15m
```

Then:

```bash
kubectl -n aicr-validation logs <nccl-all-reduce-bw-net-pod>
# → level=INFO msg=SKIP reason="[INTERNAL] no nccl-all-reduce-bw-net constraint in recipe: skip"

kubectl -n aicr-validation logs <gpu-operator-version-pod>
# → level=INFO msg="no version constraint in recipe, reporting detected version only"
#   …and the validator exits 0 (reported as passed).
```

## Root cause

The orchestrator writes the wrong wire shape into the `aicr-validation-<runID>` ConfigMap the validator pod consumes.

- **`pkg/recipe.Validation`** (orchestrator side) inlines `ValidationConfig`:
  ```go
  type Validation struct {
      // ...
      ValidationConfig `json:",inline" yaml:",inline"`
      ComponentRefs []ComponentRef `…`
      Criteria      Criteria       `…`
      Constraints   []Constraint   `…`
  }
  ```
[...]

PR #900, opened 2026-05-14
fix(validator): surface skip reason in CTRF, treat missing constraint as skip

### LINKED ISSUE #875: Validator skip signal is invisible: termination-log not written, gpu-operator-version reports skip as pass
## Summary

Two related defects in the validator-runtime skip path. Together they erase the difference between "constraint satisfied" and "constraint missing" in the signed evidence bundle. Surfaced during the #874 investigation.

## 1. Skip reasons never reach the CTRF report

When a validator's `CheckFunc` returns `validators.Skip(reason)`, the runner logs the reason to stderr and exits with code 2 (`validators/runner.go:93-97`), but never writes the reason to `/dev/termination-log`. Kubernetes only populates `ContainerStatus.State.Terminated.Message` from that file, so the orchestrator's `ValidatorResult.TerminationMsg` is empty, and `Builder.AddResult` (`pkg/validator/ctrf/builder.go:154-156`) leaves `TestResult.Message` unset. CTRF reports — including the signed predicate — record `"message": null` for every skipped validator.

Reproduction: run any recipe where a phase validator legitimately skips (e.g. a performance check whose constraint isn't declared) and inspect the resulting CTRF JSON or attestation. `tests[].status` is `"skipped"` but `tests[].message` is `null`. Diagnosing why anything skipped requires `--no-cleanup` and hand-tailing pod logs.

Fix: in `validators/runner.go:runCheck`, write the reason to the termination log inside the `errSkip` branch before returning `exitCodeSkip`. The orchestrator already lifts `TerminationMsg` into `tr.Message` via `AddResult`. Add a runner-level test that exercises a `Skip("because X")` return and asserts the resulting `TestResult.Message` contains `"because X"`.

## 2. `gpu-operator-version` reports skip as pass

`validators/deployment/gpu_operator_version.go:41-44` does `if !found { return nil }` when the recipe has no `Deployment.gpu-operator.version` constraint, which the runner reports as **passed** even though nothing was evaluated. Other validators on the same path use `validators.Skip(...)` — see `inference_perf.go` and `nccl_all_reduce_bw.go`. `gpu_operator_version.go` is the outlier.

Fix: replace `return nil` with `return validators.Skip("no version constraint in recipe")`. Best landed after #1 — until skip reasons surface in CTRF, the new skip signal has no visible destination.

## Severity

P2. Passes and fails are recorded correctly; the skipped category is the gap. The signed predicate cannot currently distinguish "constraint satisfied" from "constraint missing," which weakens the trust property the bundle is meant to carry.

## Related

- #874 — wire-shape mismatch

ayuskauskas:
## Summary

Two related defects in the validator-runtime skip path. Together they erased the difference between "constraint satisfied" and "constraint missing" in the signed evidence bundle. Skip reasons now write to `/dev/termination-log` (so they appear as `TestResult.Message` in CTRF reports), and `gpu-operator-version` reports a skip — not a pass — when no version constraint is set.

## Motivation / Context

Without these fixes, every skipped validator records `"message": null` in the signed CTRF predicate, and `gpu-operator-version` masquerades skips as passes — weakening the trust property the bundle is meant to carry. Diagnosing why anything skipped required `--no-cleanup` and hand-tailing pod logs.

Fixes: #875
Related: #874 (originally surfaced this)

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `validators/` runtime (in-container CheckFunc runner)

## Implementation Notes

**Runner (`validators/runner.go`)**
- Extracted `handleCheckResult(err) int` so the skip-vs-fail decision is unit-testable without `LoadContext` (which needs a real K8s client).
- Extracted `writeTerminationFile(msg)` as the shared file-write seam; `writeTerminationLog` (the FAIL helper) now delegates to it.
- Skip branch now writes the reason to the termination log *after* the existing `slog.Info("SKIP", ...)`.
- `writeTerminationFile` logs `slog.Warn` on `os.WriteFile` failure — silent loss here is the exact class of bug being fixed.
- `terminationLogPath` is now a `var` (with a guard-rail comment) only so unit tests can redirect to a temp file; the test helper documents the non-parallel constraint.

**gpu-operator-version (`validators/deployment/gpu_operator_version.go`)**
- `return nil` (= pass) replaced with `return validators.Skip(fmt.Sprintf("no Deployment.gpu-operator.version constraint in recipe; detected %s", version))`. The detected version is intentionally embedded in the skip message — it's diagnostic information that belongs in the report consumer's view. Pattern matches sibling validators (`inference_perf.go`, `nccl_all_reduce_bw.go`) but adds the detected value.

## Testing

```bash
make qualify                                                # green: tests + lint + e2e + scan + license headers
golangci-lint run -c .golangci.yaml ./validators/...        # 0 issues
go test -race -count=1 ./validators/...                     # all 6 sub-packages green
```

**New tests (validators/runner_test.go):**
- `TestHandleCheckResult` — 5 subtests: nil/skip/empty-skip/oversize-skip-truncates/fail. The oversize-skip case proves end-to-end truncation through the skip branch.
- `TestHandleCheckResultOverwrites` — two sequential skips; second overwrites first (guards against O_APPEND regressions).
- `TestWriteTerminationFileTruncatesKeepsPrefix` — distinguishable prefix asserts truncation keeps the head, not the tail.

**Coverage (changed packages, vs origin/main):**
- `validators/`: 29.9% → 44.0% (+14.1%)
- `validators/deployment/`: 43.8% → 43.9% (+0.1%)

**Note on `make test-coverage`:** the project-wide gate currently fails at 74.5% / 75% threshold. This is a **pre-existing failure on `origin/main`** (verified by running the same command against the unmodified baseline) — not introduced by this PR. This PR moves the needle in the right direction on the changed packages. Separate ticket may be warranted to investigate which package(s) dropped recently.

## Risk Assessment

**Rollout notes:** Behavior change is observable but is the intended outcome of the fix:
1. CTRF reports for previously-skipped validators will now have a non-null `message` field (was `null`). Downstream consumers parsing CTRF that asserted `message == null` for skipped tests will need to update.
2. `gpu-operator-version` results will now show `status: "skipped"` (was `"passed"`) when the recipe has no version constraint. Pass/fail are unchanged when a constraint is present.

## Checklist

- [x] I updated docs if user-facing behavior changed — N/A (no validator-name level docs reference `gpu-operator-version`; CTRF message field already existed)

### DIFF
--- a/validators/deployment/gpu_operator_version.go
+++ b/validators/deployment/gpu_operator_version.go
@@ -17,7 +17,6 @@ package main
 import (
 	"context"
 	"fmt"
-	"log/slog"
 	"strings"
 
 	"github.com/NVIDIA/aicr/pkg/constraints"
@@ -40,8 +39,7 @@ func checkGPUOperatorVersion(ctx *validators.Context) error {
 	// Find the version constraint in the recipe
 	constraintExpr, found := findDeploymentConstraint(ctx, "Deployment.gpu-operator.version")
 	if !found {
-		slog.Info("no version constraint in recipe, reporting detected version only")
-		return nil
+		return validators.Skip(fmt.Sprintf("no Deployment.gpu-operator.version constraint in recipe; detected %s", version))
 	}
 
 	// Evaluate constraint
--- a/validators/runner.go
+++ b/validators/runner.go
@@ -26,8 +26,9 @@ import (
 )
 
 // terminationLogPath is the standard K8s termination message path.
-// Kept as a constant (not configurable) since it matches the Job spec.
-const terminationLogPath = "/dev/termination-log"
+// Declared as a var (not const) only so unit tests can redirect writes
+// to a temp file; production code must not reassign it.
+var terminationLogPath = "/dev/termination-log"
 
 // exitCodeSkip is the exit code for a skipped check (not applicable).
 const exitCodeSkip = 2
@@ -89,24 +90,44 @@ func runCheck(checkFn CheckFunc) int {
 		return 1
 	}
 	defer ctx.Cancel()
+	return handleCheckResult(checkFn(ctx))
+}
 
-	if err := checkFn(ctx); err != nil {
-		if errors.Is(err, errSkip) {
-			slog.Info("SKIP", "reason", err.Error())
-			return exitCodeSkip
-		}
-		writeTerminationLog("%v", err)
-		return 1
+// handleCheckResult maps a CheckFunc return value to an exit code and
+// surfaces the reason (skip or failure) via slog and the termination log.
+// Writing the skip reason to the termination log is required for the
+// orchestrator's ValidatorResult.TerminationMsg to flow into the CTRF
+// report's TestResult.Message; without it skipped checks appear with a
+// null message in the signed evidence bundle.
+func handleCheckResult(err error) int {
+	if err == nil {
+		return 0
 	}
+	if errors.Is(err, errSkip) {
+		slog.Info("SKIP", "reason", err.Error())
+		writeTerminationFile(err.Error())
+		return exitCodeSkip
+	}
+	writeTerminationLog("%v", err)
+	return 1
+}
 
-	return 0
+// writeTerminationFile writes msg to the termination log path, truncated
+// to TerminationLogMaxSize. Does not emit any slog entry for the message
+// itself; callers log separately so the slog level can reflect skip vs.
+// failure. A failed write is logged at WARN — silent loss here is the
+// exact class of bug this code path exists to prevent.
+func writeTerminationFile(msg string) {
+	if len(msg) > defaults.TerminationLogMaxSize {
+		msg = msg[:defaults.TerminationLogMaxSize]
+	}
+	if err := os.WriteFile(filepath.Clean(terminationLogPath), []byte(msg), 0o600); err != nil { //nolint:gosec // Fixed path, not user-controlled
+		slog.Warn("failed to write termination log", "path", terminationLogPath, "error", err)
+	}
 }
 
 func writeTerminationLog(format string, args ...any) {
 	msg := fmt.Sprintf(format, args...)
 	slog.Error("FAIL", "message", msg)
-	if len(msg) > defaults.TerminationLogMaxSize {
-		msg = msg[:defaults.TerminationLogMaxSize]
-	}
-	_ = os.WriteFile(filepath.Clean(terminationLogPath), []byte(msg), 0o600) //nolint:gosec // Fixed path, not user-controlled
+	writeTerminationFile(msg)
 }
--- a/validators/runner_test.go
+++ b/validators/runner_test.go
@@ -16,8 +16,12 @@ package validators
 
 import (
 	"errors"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
+
+	"github.com/NVIDIA/aicr/pkg/defaults"
 )
 
 func TestSkip(t *testing.T) {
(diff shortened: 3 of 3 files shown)


validators/runner.go
```diff
@@ -89,24 +90,44 @@ func runCheck(checkFn CheckFunc) int {
...
+// orchestrator's ValidatorResult.TerminationMsg to flow into the CTRF
+// report's TestResult.Message; without it skipped checks appear with a
+// null message in the signed evidence bundle.
+func handleCheckResult(err error) int {
+	if err == nil {
+		return 0
+	}
+	if errors.Is(err, errSkip) {
+		slog.Info("SKIP", "reason", err.Error())
+		writeTerminationFile(err.Error())
```
mchmarny, line 108: nit: `slog.Info("SKIP", "reason", err.Error())` and `writeTerminationFile(err.Error())` both call `err.Error()` — minor duplication. Could hoist to `msg := err.Error()` and pass to both. Trivial; ignore if not worth the churn.

validators/deployment/gpu_operator_version.go
```diff
@@ -40,8 +39,7 @@ func checkGPUOperatorVersion(ctx *validators.Context) error {
 	// Find the version constraint in the recipe
 	constraintExpr, found := findDeploymentConstraint(ctx, "Deployment.gpu-operator.version")
 	if !found {
-		slog.Info("no version constraint in recipe, reporting detected version only")
-		return nil
+		return validators.Skip(fmt.Sprintf("no Deployment.gpu-operator.version constraint in recipe; detected %s", version))
```
mchmarny, line 42: Heads-up on what this actually writes into CTRF: `Skip(...)` wraps `errSkip` via `pkg/errors.Wrap(ErrCodeInternal, reason, errSkip)`, and `StructuredError.Error()` formats as `[%s] %s: %v`. So the `TestResult.Message` field will read:

```
[INTERNAL] no Deployment.gpu-operator.version constraint in recipe; detected v25.3.0: skip
```

The `[INTERNAL]` prefix on a skip is misleading for downstream readers — it visually reads like an internal error. Not introduced by this PR (it's how `Skip()` was already formatting things to the log line), but this PR is the first time those bytes land in the signed evidence bundle where humans will read them.

Out of scope to fix here, but worth a follow-up: either give `Skip` its own code (e.g. `ErrCodeSkipped`) or have `Skip` return a plain `fmt.Errorf("%s: %w", reason, errSkip)` and skip the structured-error wrapping (it has no useful code/context for the skip case).

validators/runner.go
```diff
@@ -26,8 +26,9 @@ import (
 )
 
 // terminationLogPath is the standard K8s termination message path.
-// Kept as a constant (not configurable) since it matches the Job spec.
-const terminationLogPath = "/dev/termination-log"
+// Declared as a var (not const) only so unit tests can redirect writes
+// to a temp file; production code must not reassign it.
```
mchmarny, line 30: nit: const→var with a "tests only" comment is a known-but-ugly pattern — nothing in Go enforces "production code must not reassign it." A common alternative is to keep the const and have the runner take an injectable `io.Writer` (or path) for the termination file, then tests construct the runner with a `*os.File` to a temp path. Not blocking — the comment makes intent clear and the blast radius is small (one package-local var). Filing for future reference if this seam grows.

validators/deployment/gpu_operator_version.go
mchmarny, line 42: Are there other validators with the same masquerade-as-pass bug? PR mentions `inference_perf.go` and `nccl_all_reduce_bw.go` match the new pattern — worth a quick grep across `validators/` for any `return nil` paths reached from a "no constraint / not applicable" check that should also be reporting skip.

mchmarny (approved):
LGTM. Clean, narrow bug fix backed by good tests — the `handleCheckResult` extraction makes the skip-vs-fail seam unit-testable, and the new `writeTerminationFile` slog.Warn on write failure is exactly the right call given this is the bug class being fixed. Coverage delta is honest about the existing project-wide gate.

Three nits inline, none blocking:

1. The bigger one: `Skip(reason)`'s wrapping via `pkg/errors.Wrap(ErrCodeInternal, ...)` means CTRF readers will see `[INTERNAL] <reason>: skip` for what is conceptually a skip. Pre-existing in `Skip()`, but this PR is the first time those bytes leave stderr and land in the signed evidence bundle. Worth a follow-up.
2. `terminationLogPath` const→var is a documented test seam — fine for now, an injected writer would be cleaner if this grows.
3. Quick grep suggestion for other `return nil` paths that should be `Skip(...)` (PR references `inference_perf.go` / `nccl_all_reduce_bw.go` as already-matching).

CI: lint, test, security scan, codeql green; E2E and analyze still running but unrelated to the change shape.

PR #906, opened 2026-05-15
refactor(validator): extract job planning to public v1 API

### LINKED ISSUE #734: Add public RenderJob() function returning Job specs
## Overview

Add a public Go package export that returns Kubernetes Job specs for validators without deploying them. This enables external controllers to use AICR as a validation planning engine.

## Go Package Design

**Primary consumer function:**
```go
// RenderJob returns a ready-to-deploy Kubernetes Job spec for a validator.
// This is the primary contract for external consumers.
func RenderJob(ctx context.Context, config ValidationConfig, catalog ValidatorCatalog,
    snapshotCMRef, recipeCMRef string, opts ...Option) (*batchv1.Job, error)
```

**Internal/advanced function:**
```go
// Plan returns a JobPlan data structure (internal representation).
// This is exposed for:
// 1. AICR internal testability (pure function testing)
// 2. Advanced consumers who need to inspect the plan before rendering
//
// Most consumers should use RenderJob() instead.
func Plan(ctx context.Context, config ValidationConfig, catalog ValidatorCatalog,
    snapshotCMRef, recipeCMRef string, opts ...Option) (*JobPlan, error)
```

## Rationale

**Why RenderJob() as primary contract:**
- Prevents consumer divergence from AICR's renderer (addressing review feedback)
- Simple one-call function for consumers
- Version pinning in go.mod handles breaking changes
- **Terminology note**: This is Go package consumption, not an API (per architectural guidance)

**Why expose Plan() at all:**
- Internal testability (pure function without K8s types)
- Advanced consumers can inspect before rendering
- Not required for basic usage

## JobPlan Structure

JobPlan is an internal data structure not intended for consumer manipulation:

```go
type JobPlan struct {
    ValidatorName   string
    Image          string
    Command        []string
    Args           []string
    Env            []corev1.EnvVar
    Volumes        []corev1.Volume
    VolumeMounts   []corev1.VolumeMount
    Resources      corev1.ResourceRequirements
    // ... other fields
}
```

**Contract:** JobPlan structure may evolve in minor versions. Consumers should not build custom renderers from JobPlan - use RenderJob() instead.

### Basic Consumer (Controller)
```go
// Simple: just call RenderJob
job, err := validator.RenderJob(ctx, config, catalog, snapCM, recipeCM)
if err != nil {
    return err
}

// Deploy
_, err = clientset.BatchV1().Jobs(ns).Create(ctx, job, metav1.CreateOptions{})
```

### Advanced Consumer (Custom Inspection)
```go
// Advanced: inspect plan before rendering
plan, err := validator.Plan(ctx, config, catalog, snapCM, recipeCM)
if err != nil {
    return err
}

// Inspect plan (e.g., for dry-run reporting)
fmt.Printf("Will run validator: %s with image: %s\n", plan.ValidatorName, plan.Image)

// Still use AICR's renderer (don't build custom)
job := validator.RenderJobFromPlan(plan)
```

## Acceptance Criteria

- [ ] `RenderJob()` function exported from `pkg/validator/job/` package
- [ ] Returns `*batchv1.Job` ready for deployment
- [ ] `Plan()` function exported but documented as internal/advanced use
- [ ] JobPlan struct exported with godoc warning about stability
- [ ] Unit tests for RenderJob() (integration)
- [ ] Unit tests for Plan() (pure function)
- [ ] Godoc clearly states RenderJob() is primary contract
- [ ] Godoc clarifies this is Go package consumption (not an API)
- [ ] Example code in package documentation

## Related

- Parent: #731 (External validation controller integration)
- Depends on: #732 (Validation type), #733 (ValidatorCatalog type)
- References: GitLab MR platform/architecture!64 (ADR-0028)

xdu31:
## Summary

Extracts validator job planning logic into a public API (`pkg/api/validator/v1/`) to enable external controllers to build and render validator Jobs without depending on internal deployer implementation.

## Motivation/Context

Resolves #734

External controllers (e.g., DGXC platform operator) need to create validator Jobs programmatically. Previously, job building logic was encapsulated in the internal `job.Deployer` type, making it inaccessible to external consumers.

This refactoring exposes the job planning API as stable, documented public functions that external controllers can import and use.

## Type of Change

- [x] Refactoring (code restructuring without behavior change)
- [x] Bug fix (addressing code review findings)

## Components Affected

- `pkg/api/validator/v1/` - New public job planning API
- `pkg/validator/job/deployer.go` - Simplified to use v1 API
- `pkg/validator/catalog/` - Removed duplicate ImagePullPolicy
- `pkg/validator/validator.go` - Fixed duplicate godoc

## Implementation Notes

**New Public API:**
- `JobPlan` struct - Decomposes Job into reusable components (image, args, env, volumes, resources, etc.)
- `BuildJobPlan()` - Builds a JobPlan from ValidatorEntry and runtime parameters
- `RenderPlan()` - Renders a complete Kubernetes Job from JobPlan
- `RenderPlanToApplyConfig()` - Renders ApplyConfiguration for server-side apply
- `ImagePullPolicy()` - Determines pull policy based on image reference
- `GenerateRunID()` - Creates unique run identifiers

**Key Design Decisions:**
- Added `Namespace` field to `JobPlan` for self-contained plan data
- Split public API (`job_plan.go`) from internal helpers (`job_plan_internal.go`)
- Deployer now uses v1 API instead of internal implementation
- Consolidated all job planning tests into `v1/job_plan_test.go`

**Code Verification:**
Detailed field-by-field comparison confirms the new v1 API produces equivalent Job specs to the original deployer implementation (with one minor enhancement for future flexibility - see Resources Handling below).

### Complete Field Preservation in ApplyConfigurations
Fixed `buildEnvVarApply()` and `buildVolumesApply()` to preserve all fields when converting to ApplyConfigurations:
- **EnvVar sources**: Now preserves Optional, APIVersion, Divisor for all 4 ValueFrom types (FieldRef, SecretKeyRef, ConfigMapKeyRef, ResourceFieldRef)
- **Volume sources**: Now preserves Items, DefaultMode, Optional, ReadOnly for all 7 volume types (ConfigMap, Secret, EmptyDir, HostPath, PVC, Projected, DownwardAPI)
- Split `buildVolumesApply()` into helper functions to satisfy funlen linter while maintaining complete field coverage

### Entropy Failure Validation
Changed `GenerateRunID()` and `generateJobName()` to panic on entropy failures instead of silent fallback:
- Validates both error return and bytes read from `rand.Read()`
- Prevents predictable ID generation that could cause collisions
- Added DNS-1123 validation to `generateJobName()` with length checks

### Resources Handling (Behavior Enhancement)
`buildResources()` introduces a minor enhancement over the original deployer:

**Original deployer:** Always hardcoded to `1 CPU / 1Gi` (ignored catalog `entry.Resources`)
**Refactored code:** Honors `entry.Resources` if set, else defaults to `1 CPU / 1Gi`

**Current Impact:**
- No observable difference today - no catalog entry currently sets `Resources`
- All validators get `1 CPU / 1Gi` (identical behavior to old deployer)
- Divergence is latent and only surfaces if/when catalog entries add custom requirements

**Validation:** If parsing fails (invalid CPU/memory values), logs warning and falls back to defaults, preventing invalid quantities from reaching the API server.

This enhancement enables future validators to declare custom resource requirements in the catalog without code changes.

### Code Quality Improvements
- Replaced O(n²) bubble sort with `sort.Strings()` in `serializeNodeSelector()`
- Fixed duplicate godoc on `EnsureDataConfigMaps()`
- Added comprehensive DNS-1123 label validation for generated job names

### Deferred Issues
- **tolerations parameter naming** (#1): Deferred for architectural discussion (rename vs honor parameter)
- **os.Getenv reads** (#5): Tracked in #954 - pre-existing behavior from deployer, now visible in public API

## Testing

- [x] All existing tests pass (21/21)
- [x] New comprehensive tests in `pkg/api/validator/v1/job_plan_test.go`
- [x] Test coverage: v1 package 67.4%, overall 75.3%
- [x] Verified output equivalence with old deployer implementation
- [x] `make qualify` passes (tests, lint, e2e, vulnerability scan)

## Risk Assessment

**Low Risk:**
- Pure refactoring - no breaking behavior changes
- All existing tests pass
- New API produces identical Job specs (verified)
- Deployer continues to work using v1 API
- Review findings addressed with backward-compatible fixes

## Checklist

- [x] Code follows project style and patterns
- [x] Tests added/updated and passing
- [x] Linting passes (0 issues)
- [x] Documentation updated (README.md added for v1 API)
- [x] No breaking changes to existing internal APIs
- [x] Commit message follows conventional commits format
- [x] Review findings addressed or tracked in issues


xdu31:
## Response to CodeRabbit Findings

### ✅ Tolerations (job_plan.go:215) - No change needed

The current implementation is **correct by design**. The orchestrator Job must **always** use tolerate-all (`{Operator: Exists}`) regardless of the incoming tolerations parameter.

**Rationale:**
- The orchestrator Job is the validator container that runs as a Kubernetes Job
- It needs to schedule on any CPU node, so it must tolerate all taints
- The `tolerations` parameter is **only** for inner workloads that validators spawn (e.g., GPU benchmarks)
- These inner tolerations are serialized into the `AICR_TOLERATIONS` env var and applied by validators to their child Jobs

**Evidence:**
Test `TestDeployJobOrchestratorToleratesTolerateAll` explicitly validates this behavior with the comment:
> "The orchestrator Job must always have tolerate-all so it can schedule on any CPU node, regardless of what tolerations are passed for inner workloads."

The test verifies that even when passing narrow GPU tolerations, the orchestrator Job still uses tolerate-all.

---

### ✅ ConfigMap ResourceVersion (validator.go:480-498) - Already implemented correctly

The ResourceVersion **is** being preserved. The code at line 482 fetches the existing ConfigMap:
```go
existing, getErr := clientset.CoreV1().ConfigMaps(namespace).Get(ctx, cm.name, metav1.GetOptions{})
```

Then at line 498, we update using that `existing` object:
```go
_, updateErr := clientset.CoreV1().ConfigMaps(namespace).Update(ctx, existing, metav1.UpdateOptions{})
```

Since `existing` comes from the Get call, it already contains `existing.ObjectMeta.ResourceVersion`, which is automatically preserved in the Update call. This is the standard Kubernetes pattern for optimistic concurrency control.

---

### ✅ README Issues - Fixed

All function signature and parameter name issues in the README have been corrected:
- Removed `namespace` parameter from `RenderPlan()` and `RenderPlanToApplyConfig()` signatures
- Changed `imagePullSecret` (singular string) to `imagePullSecrets` (plural slice) throughout
- Updated all examples to match actual function signatures

Changes are staged and ready to commit.

xdu31:
## ✅ Rebased on upstream main

Successfully rebased on latest upstream/main (932f59e8) and resolved merge conflict.

**Conflict Resolution:**
- **File:** `pkg/validator/validator.go` imports
- **Issue:** Upstream added imports for `GenerateRunID` which we moved to v1 package
- **Fix:** Removed unused imports (`crypto/rand`, `encoding/hex`), kept `stderrors`

**Verification:**
```
✅ make qualify PASSED
   - Tests: 21/21 passed
   - Coverage: 75.4% (above threshold)
   - Linting: passed
   - E2E: passed
   - Security: no vulnerabilities
```

**Current commits:**
```
723e33c5 docs(validator): fix README function signatures and imagePullSecrets references
d0e9e66f fix(validator): address CodeRabbit review findings for job planning API
72d49021 refactor(validator): extract job planning to public v1 API
```

Ready for review.

pkg/api/validator/v1/job_plan.go
```diff
@@ -0,0 +1,397 @@
...
+		Namespace:        namespace,
+		Image:            entry.Image,
+		Args:             entry.Args,
+		Env:              env,
+		Volumes:          volumes,
+		VolumeMounts:     volumeMounts,
+		Resources:        resources,
+		Timeout:          int64(timeout.Seconds()),
+		ServiceAccount:   serviceAccount,
+		Tolerations:      []corev1.Toleration{{Operator: corev1.TolerationOpExists}},
```
mchmarny, line 239: The `tolerations` parameter passed to `BuildJobPlan` is silently dropped here — `JobPlan.Tolerations` is hardcoded to `[]Toleration{{Operator: TolerationOpExists}}`. The caller's `tolerations` only reach the inner workloads via the `AICR_TOLERATIONS` env var.

This faithfully replicates the deployer's prior behavior, but the public API surface advertises something different: the README describes `JobPlan.Tolerations` as "Pod tolerations" and the `Plan()` parameter docs say "Pod tolerations (forwarded to inner workloads)". An external controller passing `tolerations=[…]` will reasonably assume those land on the validator Pod itself; only the env-var forwarding actually happens.

Two ways to resolve:
- Honor `tolerations` for the Pod (replace or merge with the hardcoded `OpExists`) — principle-of-least-surprise.
- Rename the parameter to `innerWorkloadTolerations` in the signature **and** README, and document that the orchestrator Pod always uses `OpExists`.

Same concern applies to `nodeSelector` — it's not applied to the validator Pod's `NodeSelector`, only forwarded via env. Worth clarifying in the README too.

pkg/api/validator/v1/job_plan_internal.go
```diff
@@ -0,0 +1,367 @@
...
+			if v.DownwardAPI.DefaultMode != nil {
+				downwardAPI = downwardAPI.WithDefaultMode(*v.DownwardAPI.DefaultMode)
+			}
+			volApply = volApply.WithDownwardAPI(downwardAPI)
+		}
+
+		volumesApply = append(volumesApply, volApply)
+	}
+	return volumesApply
+}
```
mchmarny, line 503: `buildVolumesApply` and `buildEnvVarApply` silently drop fields that `RenderPlan` preserves, so the two render paths are not equivalent for a customized `JobPlan`:

- `ConfigMapVolumeSource`: only `Name` copied; `Items`, `DefaultMode`, `Optional` dropped.
- `SecretVolumeSource`: only `SecretName`; `Items`, `DefaultMode`, `Optional` dropped.
- `PersistentVolumeClaimVolumeSource`: only `ClaimName`; `ReadOnly` dropped.
- `DownwardAPIVolumeSource`: only `DefaultMode`; `Items` dropped (the entire reason most users pick DownwardAPI).
- `Projected.Sources.{ConfigMap,Secret}`: only top-level `Name`; `Items`, `Optional` dropped.
- `EnvVar.ValueFrom.SecretKeyRef` / `ConfigMapKeyRef`: `Optional` dropped.

This is fine for AICR's internal use (the deployer only feeds two simple ConfigMap volumes and basic env), but the README explicitly invites external controllers to customize JobPlan and pick either render path:

> Both Create and server-side Apply deployment strategies are supported

A caller who customizes `plan.Volumes[0].ConfigMap.Items` and renders via `RenderPlanToApplyConfig` will silently lose those items vs the same plan run through `RenderPlan`. Either round-trip everything via the typed-to-apply conversion (k8s.io/client-go has helpers for this), or scope the public API down to the supported subset and document explicitly what is/isn't preserved.

pkg/api/validator/v1/job_plan.go
```diff
@@ -0,0 +1,397 @@
...
+// External controllers should use this to generate runIDs before creating
+// ConfigMaps and rendering Jobs.
+func GenerateRunID() string {
+	timestamp := time.Now().Format("20060102-150405")
+	randomBytes := make([]byte, 8)
+	if n, err := rand.Read(randomBytes); err != nil || n != len(randomBytes) {
+		return timestamp
+	}
+	return fmt.Sprintf("%s-%s", timestamp, hex.EncodeToString(randomBytes))
+}
```
mchmarny, line 102: Silent fallback on `rand.Read` failure is risky here. The runID is embedded in ConfigMap names (`aicr-snapshot-{runID}`, `aicr-validation-{runID}`) and Job names — if the system has degraded entropy and two concurrent runs both hit this path in the same second, they will collide on those resources and clobber each other's data.

Prefer one of:
- Return `(string, error)` and let the caller decide.
- Panic — entropy outages are exceptional and we'd rather fail fast than ship predictable IDs that look unique.

Same pattern in `generateJobName` (`job_plan_internal.go:222`) — its fallback `aicr-{name}-{unix}` also doesn't match the documented format `aicr-{validatorName}-{random-hex}`, so downstream regex/label parsers may break.

pkg/api/validator/v1/job_plan_internal.go
```diff
@@ -0,0 +1,367 @@
...
+func serializeNodeSelector(ns map[string]string) string {
+	keys := make([]string, 0, len(ns))
+	for k := range ns {
+		keys = append(keys, k)
+	}
+	// Sort keys for deterministic output
+	for i := 0; i < len(keys); i++ {
+		for j := i + 1; j < len(keys); j++ {
+			if keys[i] > keys[j] {
+				keys[i], keys[j] = keys[j], keys[i]
```
mchmarny: This is a regression from the deployer, which used `slices.Sort(keys)` (see the deleted code in `pkg/validator/job/deployer.go`). Hand-rolled O(n²) bubble sort is slower, harder to read, and there's no reason to avoid the stdlib here. Replace with:

```go
import "sort"
// ...
sort.Strings(keys)
```

or `slices.Sort(keys)` if you prefer the generic helper.

pkg/api/validator/v1/job_plan.go
```diff
@@ -0,0 +1,397 @@
...
+		// also required for disconnected/air-gapped deployments.
+		return corev1.PullIfNotPresent
+	}
+	if os.Getenv("AICR_VALIDATOR_IMAGE_TAG") != "" {
+		return corev1.PullAlways
+	}
+	if strings.HasSuffix(image, ":latest") {
+		return corev1.PullAlways
+	}
+	return corev1.PullIfNotPresent
```
mchmarny, line 126: `ImagePullPolicy` reads `AICR_VALIDATOR_IMAGE_TAG` from the caller's process env. Same in `buildEnv` (`job_plan_internal.go:85,88` — `AICR_VALIDATOR_IMAGE_REGISTRY`, `AICR_VALIDATOR_IMAGE_TAG`).

For AICR's own deployer this is fine — the CLI process and the planning live in the same binary. But this is now a public API surface for external controllers: an external operator calling `v1.Plan(...)` will silently behave differently based on the operator pod's environment, with no signal in the function signature.

At minimum, hoist this into the API contract: either accept the override values as struct/parameter inputs (e.g., on `JobPlan` or a `BuildOptions` struct), or document this prominently in the README as "the planning functions inspect these env vars in the caller's process".

pkg/api/validator/v1/job_plan_internal.go
```diff
@@ -0,0 +1,367 @@
...
+		{Name: "snapshot", MountPath: "/data/snapshot", ReadOnly: true},
+		{Name: "validation", MountPath: "/data/validation", ReadOnly: true},
+	}
+}
+
+// buildResources creates resource requirements for the validator container
+func buildResources(entry ValidatorEntry) corev1.ResourceRequirements {
+	// Use catalog entry resources if specified, otherwise use defaults
+	if entry.Resources != nil && entry.Resources.CPU != "" && entry.Resources.Memory != "" {
+		// Parse user-provided quantities with error handling
```
mchmarny, line 138: Behavior change relative to the original deployer, which always hardcoded `1 CPU / 1Gi memory` regardless of `entry.Resources`. The new code respects `entry.Resources` when set.

That's almost certainly a desirable change, but it contradicts the PR description's claim that "the new v1 API produces byte-identical Job specs as the original deployer implementation" / "Verified output equivalence with old deployer implementation". Currently moot because no catalog entry sets `resources:`, but the moment one does, the spec diverges.

Either update the PR description to call this out as an intentional improvement, or revert `buildResources` to the hardcoded values and surface `entry.Resources` support in a follow-up PR. Quietly enabling a code path that mutates production Job specs on the next catalog change is a footgun.

pkg/api/validator/v1/job_plan_internal.go
```diff
@@ -0,0 +1,367 @@
...
+	}
+	return strings.Join(parts, ",")
+}
+
+// int32Ptr returns a pointer to an int32
+func int32Ptr(i int32) *int32 {
+	return &i
+}
+
+// int64Ptr returns a pointer to an int64
```
mchmarny, line 218: Two concerns on this helper:

1. **No K8s name validation.** The format `aicr-{validatorName}-{hex}` assumes `validatorName` is a valid DNS-1123 label. If a catalog entry has an uppercase or underscore character — or a long name that pushes the total >63 chars — Job creation will fail at apply time with an opaque server error. The catalog should constrain names, but exposing this via the public API means external controllers can also feed in arbitrary `ValidatorEntry.Name` values. Worth a `validateName()` guard, or a comment documenting the caller's responsibility.

2. **Fallback format inconsistency** (line 226): `fmt.Sprintf("aicr-%s-%d", validatorName, time.Now().Unix())` doesn't match the documented `aicr-{validatorName}-{random-hex}` format. Downstream tools that regex on the hex suffix will break. Combined with the silent-fallback concern in `GenerateRunID`, prefer surfacing the rand failure instead.

pkg/validator/validator.go
```diff
@@ -435,12 +433,19 @@ func (v *Validator) phaseSkipped(cat *catalog.ValidatorCatalog, phase Phase, rea
 	}
 }
 
-// ensureDataConfigMaps creates snapshot and validation ConfigMaps for this run.
-// The validation payload is marshaled as v1.ValidationInput (the wire shape
-// that validator containers consume) with phase configs nested under `config:`.
-func (v *Validator) ensureDataConfigMaps(
+// EnsureDataConfigMaps creates or updates snapshot and validation ConfigMaps.
+// Creates ConfigMaps named aicr-snapshot-{runID} and aicr-validation-{runID} with
+// create-or-update semantics. External controllers should call this after generating
```
mchmarny, line 438: nit: duplicate godoc on `EnsureDataConfigMaps`. Two paragraphs separated by a blank line — only the second one attaches to the func. Collapse into one block.

pkg/api/validator/v1/README.md
```diff
@@ -0,0 +1,355 @@
...
+
+Rules:
+- `ko.local/` or `kind.local/` → `PullNever` (side-loaded images)
+- Digest-pinned (`@sha256:...`) → `PullIfNotPresent` (immutable)
+- `AICR_VALIDATOR_IMAGE_TAG` env set → `PullAlways` (override mode)
+- `:latest` tag → `PullAlways` (mutable tag)
+- Default → `PullIfNotPresent` (versioned releases)
+
+## RBAC Considerations
+
```
mchmarny, line 344: "Pod tolerations (forwarded to inner workloads)" is misleading — `tolerations` is *only* forwarded to inner workloads via `AICR_TOLERATIONS` env var. The orchestrator Pod itself ignores the parameter and uses a hardcoded `OpExists`. Same for `nodeSelector` (line 345) — it's forwarded via env, not applied to the validator Pod's `nodeSelector` field.

Either fix the implementation (see comment on `job_plan.go`) or fix this doc to say "forwarded to inner workloads via env var; does not affect the validator Pod's own scheduling".

mchmarny (changes requested):
Solid refactor in shape — the deployer now delegates to a v1 package and the integration looks correct for AICR's own call site. CI is green. The concerns are about the *public API contract*, not the deployer integration, but they matter because this is being shipped as a stable surface for external controllers.

**Blocking:**
- `BuildJobPlan` accepts `tolerations` and `nodeSelector` but neither lands on the validator Pod — they only flow through env vars to inner workloads. The README advertises them as Pod-level. Either honor them or rename + document. (inline on `job_plan.go`)
- `RenderPlan` vs `RenderPlanToApplyConfig` are not equivalent for customized JobPlans — the apply path silently drops `Items`, `DefaultMode`, `Optional`, `PVC.ReadOnly`, `DownwardAPI.Items`, etc. The README claims both paths are supported equally. (inline on `job_plan_internal.go`)
- `GenerateRunID` and `generateJobName` silently fall back on entropy failure, producing predictable / format-divergent IDs that are used as ConfigMap and Job names. (inline)

**Non-blocking but worth fixing in this PR:**
- "Byte-identical" claim in the PR description is wrong — `buildResources` now respects `entry.Resources` (catalog has none today, but the divergence is real).
- `serializeNodeSelector` regressed from `slices.Sort` to manual bubble sort.
- `os.Getenv` reads inside the planning API leak the caller's process env into the public contract — at minimum, document; ideally, parameterize.
- Duplicate godoc on `EnsureDataConfigMaps`.

Also note the PR is marked `needs-rebase` / `mergeable_state: blocked` — rebase onto `main` before merging.

pkg/api/validator/v1/job_plan.go
xdu31: @mchmarny we may not making the change at this point

This requires renaming the field consistently throughout the entire codebase (including Validator.Tolerations, WithTolerations(), and potentially the --toleration CLI flag), which is a much larger user facing breaking change.

The same thing is applied to `nodeSelector`

mchmarny: This is all net-new so no breaking change, is there? But if global rename is not an option then let's honor tolerations for the Pod then. Right now it drops it silently

xdu31: > os.Getenv reads inside the planning API leak the caller's process env into the public contract

Good catch. This behavior existed in the deployer before the refactoring (see https://github.com/NVIDIA/aicr/blob/main/pkg/validator/job/deployer.go#L256) - we moved it to the v1 API during extraction but didn't introduce the pattern. That said, you're absolutely right that it's more problematic now that it's in a public API contract.

I've created #954 to track this. Two approaches:

1. Document (quick fix): Add godoc explaining that BuildJobPlan() reads AICR_VALIDATOR_IMAGE_REGISTRY and AICR_VALIDATOR_IMAGE_TAG from the caller's process environment. Non-breaking, makes the contract explicit.
2. Parameterize (proper fix): Add imageRegistryOverride and imageTagOverride parameters to replace the os.Getenv() calls. Gives callers explicit control but requires updating all call sites.

Preference for Option 1 in this PR (document the existing behavior), then tackle Option 2 in a follow-up that includes the internal deployer refactoring? That way the breaking API change happens once with a comprehensive fix.

mchmarny: ye, docs are faster, don't think that adding these 2 params is that much more work, is it? The min this merges, it will become usable as a library so let's do it right.

mchmarny (approved):

PR #924, opened 2026-05-15
fix(bundler): wire PreManifestFiles through flux deployer with terminal-aware dependsOn

### LINKED ISSUE #915: [Bug]: helmfile/helm bundles fail on GKE — system-*-critical priorityClass pods rejected by ResourceQuota admission
### Summary

On a fresh GKE Standard cluster, AICR bundle installation fails because the `gpu-operator` controller Deployment declares `priorityClassName: system-node-critical` and lands in the `gpu-operator` namespace, where GKE's `ResourceQuota` admission plugin rejects it. The ReplicaSet event makes the cause explicit:

```
FailedCreate: Error creating: insufficient quota to match these scopes:
[{PriorityClass In [system-node-critical system-cluster-critical]}]
```

The Deployment never gets to image-pull or readiness — admission blocks the pod before it's created — so the failure surfaces only as a `helmfile apply` `--wait` timeout 10 minutes later, which is high time-to-debug for a first-time user.

### Repro

1. Provision the UAT GKE H100 cluster (e.g. `gh workflow run uat-gcp.yaml -f skip_delete=true -f skip_tests=true`).
2. Generate a helmfile bundle following `demos/cuj1-gke-config.md` (`spec.bundle.deployment.deployer: helmfile`).
3. `cd ./bundle && helmfile apply --skip-diff-on-install` (after installing the `helm-diff` plugin per #914).
4. After ~10 min, the gpu-operator release fails with `Deployment/gpu-operator/gpu-operator not ready. status: InProgress, message: Replicas: 0/1` + `context deadline exceeded`.
5. `kubectl -n gpu-operator describe deploy/gpu-operator` shows the ReplicaSet stuck at 0/1 with `ReplicaFailure: True (FailedCreate)`.
6. `kubectl -n gpu-operator get events` shows the quota rejection above.

### Root cause

GKE Standard enables the `ResourceQuota` admission plugin and ships a `ResourceQuota` in `kube-system` with a `scopeSelector` admitting `system-*-critical` priority classes for that namespace only. Per the Kubernetes spec ([`ResourceQuota` scope semantics](https://kubernetes.io/docs/concepts/policy/resource-quotas/#resource-quota-per-priorityclass)): once *any* quota in the cluster scopes by `PriorityClass` for those values, pods that request a matching priority class can only be created in namespaces that have a matching quota. AICR emits gpu-operator (and likely a few other charts) into per-component namespaces with the upstream chart's default `priorityClassName: system-node-critical`, so admission blocks every pod.

EKS, AKS, and bare-metal clusters generally don't ship this default quota, which is why `cuj1-eks.md` and the existing EKS UAT path don't hit it.

### Affected components (audit needed)

Any release emitted by AICR bundlers whose pod spec sets `priorityClassName: system-node-critical` (or `system-cluster-critical`) outside `kube-system`. Confirmed today: `gpu-operator` controller. Suspected, worth grepping templates for: `network-operator`, `nvidia-dra-driver-gpu`, `nodewright-operator`. NVIDIA's k8s-device-plugin, MIG manager, and the toolkit DaemonSets that gpu-operator manages all run in the `gpu-operator` namespace and inherit the same priority class — they would have hit the same wall on subsequent reconciliation.

### Fix options (preferred → fallback)

- **A — Bundler emits a matching `ResourceQuota` alongside any release that requests `system-*-critical`.** Localized to the affected namespace, opaque to end users, no chart-value changes. The quota stays a permissive admission allowlist (not a real quota cap):

  ```yaml
  apiVersion: v1
  kind: ResourceQuota
  metadata:
    name: <release>-critical-pods
    namespace: <release-namespace>
  spec:
    hard:
      pods: \"32\"   # generous; covers DaemonSets across N GPU nodes
    scopeSelector:
      matchExpressions:
        - operator: In
          scopeName: PriorityClass
          values:
            - system-node-critical
            - system-cluster-critical
  ```

  Wire it into `pkg/bundler/component/` so any rendered release whose pod-spec walks turn up `priorityClassName: system-*-critical` triggers a sibling `ResourceQuota`. Idempotent re-apply is safe.

- **B — Default the chart's `operator.priorityClassName: \"\"` (or a non-critical class) when the deployer target is GKE.** Smaller blast radius than A in some respects (no extra resource), but globally suppresses the priority signal even where it does work (EKS/AKS/bare-metal), and we lose the operator's eviction-safety guarantee on resource-pressured GPU nodes. Not preferred.

- **C — Document the manual `ResourceQuota` step in the GKE demo and the helmfile bundle README.** Worst option — pushes Kubernetes admission trivia onto users for a problem AICR can fix once, centrally. Current state of `demos/cuj1-gke-config.md` is effectively at this level (the unblock is in chat, not in the file).

A is the right answer. Implementation pointers: `pkg/bundler/component/` is the natural seam — render the chart, scan rendered templates for `priorityClassName: system-*-critical`, append a synthesized `ResourceQuota` for each affected namespace. Same fix benefits the `helm`, `helmfile`, `flux`, and `argocd` deployers because they all share the rendered-component plumbing.

### Severity

User-visible: GKE demo blocking on a fresh cluster with no actionable error in the helmfile output — the user sees `Replicas: 0/1` + `context deadline exceeded`, which sends them debugging the gpu-operator chart, not Kubernetes admission. Time-to-debug for a first-time user is high (10 min stall + opaque error). Risk is low: the fix is additive, no data loss path.

### Workaround until fixed

```shell
kubectl apply -f - <<'EOF2'
apiVersion: v1
kind: ResourceQuota
metadata:
  name: gpu-operator-critical-pods
  namespace: gpu-operator
spec:
  hard:
    pods: \"16\"
  scopeSelector:
    matchExpressions:
      - operator: In
        scopeName: PriorityClass
        values:
          - system-node-critical
          - system-cluster-critical
EOF2

cd ./bundle && helmfile apply --skip-diff-on-install
```

`helmfile apply` is idempotent — completed releases skip; only gpu-operator (and any downstream `needs:` releases) re-runs.

Repeat the same manifest, swapping `metadata.namespace`, for any other affected release the audit above turns up.

### Related

- #914 (helmfile bundle fresh-cluster CRD/CR ordering) — different problem, same demo path. Both should be resolved before `cuj1-gke-config.md` runs cleanly end-to-end with no manual steps.

Maybe, with guidance — happy to PR option A once a maintainer signs off on the seam (component bundler post-process vs. deployer-layer hook).

### LINKED ISSUE #923: [Bug]: --deployer flux drops PreManifestFiles, so GKE ResourceQuota fix (#921) does not reach flux bundles
## Summary

PR #921 fixed issue #915 (GKE `ResourceQuota` admission blocks `system-*-critical` pods outside `kube-system`) for the `helm`, `helmfile`, `argocd`, and `argocd-helm` deployers by synthesizing a pre-phase `ResourceQuota` via `collectComponentPreManifests`. The PR description claims flux benefits too, but the flux deployer never consumes `ComponentPreManifests` — so a GKE bundle generated with `--deployer flux` still hits the original symptom from #915.

This is also a pre-existing gap from #856 (`feat: os-talos mixin + bundler preManifestFiles support`), which added `PreManifestFiles` to the recipe schema and wired four of the five deployers. The flux deployer was missed at that point and was never caught because the only production user of `PreManifestFiles` at the time was os-talos, which `--deployer flux` users don't exercise.

## Affected component

`pkg/bundler/deployer/flux` — `flux.Generator` has no `ComponentPreManifests` field, and `bundler.buildDeployer`'s `DeployerFlux` case only calls `collectComponentManifests` (post), not `collectComponentPreManifests`. The synthesized GKE quota is computed but silently dropped before reaching the flux generator.

## Repro

1. Generate a GKE recipe (e.g. via `cuj1-gke-config.md` flow).
2. `aicr bundle --recipe recipe.yaml --deployer flux --output ./bundle`.
3. Apply the bundle to a fresh GKE Standard cluster.
4. `gpu-operator` HelmRelease stays unready; `kubectl -n gpu-operator describe deploy/gpu-operator` shows the same `FailedCreate: insufficient quota to match these scopes: [{PriorityClass In [system-node-critical system-cluster-critical]}]` from #915.

Verify the bundle omits the synthesized quota:

\`\`\`bash
find ./bundle -name 'gke-critical-pods-quota.yaml'   # empty
ls ./bundle/gpu-operator-pre 2>/dev/null              # missing
\`\`\`

## Code references

- `pkg/bundler/bundler.go:388-409` — `DeployerFlux` switch case calls only `collectComponentManifests`.
- `pkg/bundler/deployer/flux/flux.go:99-150` — `flux.Generator` struct: `ComponentManifests` only, no `ComponentPreManifests` field.
- Contrast with `helm.Generator` (`pkg/bundler/deployer/helm/helm.go:80`), `helmfile.Generator` (`pkg/bundler/deployer/helmfile/helmfile.go:75`), `argocd.Generator` (`pkg/bundler/deployer/argocd/argocd.go:138`), `argocdhelm.Generator` (`pkg/bundler/deployer/argocdhelm/argocdhelm.go:123`) — all four have the field and forward to `localformat.Options.ComponentPreManifests`.

## Documentation gap

`pkg/bundler/deployer/flux/doc.go` describes the generated bundle structure but only shows `<name>-post/` for mixed components (line 84). The `<name>-pre/` wrapper (which other deployers emit when `PreManifestFiles` is non-empty) is absent from the example tree because the feature isn't implemented. This package-level doc should grow a `<name>-pre/` entry once the fix lands, and the \"Deployment Ordering\" section needs a sentence on how pre-folders thread into the \`dependsOn\` chain (\`previous → <name>-pre → <name> → <name>-post → next\`).

## Fix outline

1. Add `ComponentPreManifests map[string]map[string][]byte` to `flux.Generator`.
2. Wire `b.collectComponentPreManifests(...)` in `bundler.buildDeployer`'s `DeployerFlux` case alongside the existing post-manifest collection.
3. In `generateComponentResources`, when `g.ComponentPreManifests[ref.Name]` is non-empty, emit a `<name>-pre` HelmRelease via `generateManifestHelmChart` (reusing the same helper as `-post`), and rewire `dependsOn` so the primary depends on `<name>-pre` instead of the previous component. Chain becomes: `previous → <name>-pre → <name> → <name>-post → next`.
4. Add a `<name>-pre` collision guard mirroring the rule in `pkg/bundler/deployer/localformat/writer.go:258-263`.
5. Update `pkg/bundler/deployer/flux/doc.go`: add `<name>-pre/` to the generated structure example and document the dependsOn chain ordering.
6. Tests in `flux_test.go`: pre-only component, pre+post on same component, pre on index-0 component (head of chain), `<name>-pre` collision rejection. End-to-end: rebuild GKE bundle with `--deployer flux` and confirm `gpu-operator-pre/helmrelease.yaml` is emitted.

## Severity

Medium for GKE+flux users (full regression of #915 symptom), low overall because `--deployer flux` on GKE H100 isn't the current demo path. Bundle is well-formed but silently incomplete — no error surfaces at generation time, only at reconcile time on the cluster.

## Related

- #915 — original GKE ResourceQuota admission bug
- #921 — fix for the other four deployers (incorrectly claims flux is covered)
- #856 — original introduction of `PreManifestFiles`; flux deployer wiring was missed here

yuanchen8911:
## Summary

Wire `ComponentPreManifests` through the flux deployer so the synthesized GKE critical-priority `ResourceQuota` from #921 (and any future `PreManifestFiles` consumer such as the os-talos mixin) actually reaches `--deployer flux` bundles. Also fix a pre-existing chain-ordering bug where the next component depended on the previous component's primary name instead of its terminal release.

## Motivation / Context

PR #921 fixed #915 (GKE `ResourceQuota` admission blocks `system-*-critical` pods outside `kube-system`) for the `helm`, `helmfile`, `argocd`, and `argocd-helm` deployers by synthesizing a pre-phase `ResourceQuota` via `collectComponentPreManifests`. The PR description claimed flux benefits too, but `flux.Generator` never had a `ComponentPreManifests` field and the `DeployerFlux` switch case never called `collectComponentPreManifests` — so the synthesized quota was computed and silently dropped. A GKE bundle generated with `--deployer flux` still hit the original symptom.

The same gap also blocked the `os-talos` mixin (#856) from working on flux, since both consumers ride on `ComponentRef.PreManifestFiles`.

While addressing the issue, Codex review surfaced a second, pre-existing bug: every component's `dependsOn` resolved to the previous component's primary name, even when that previous component had `-post` manifests. Flux could reconcile the next component in parallel with the previous component's post manifests, defeating the chain's purpose. Codex verified this against a local GKE Flux bundle: `gke-nccl-tcpxo.dependsOn = gpu-operator` instead of `gpu-operator-post`.

Fixes: #923
Related: #915, #921, #856

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- **`flux.Generator` field** — add `ComponentPreManifests map[string]map[string][]byte`, mirroring the field already present on `helm`, `helmfile`, `argocd`, and `argocd-helm` generators.
- **Bundler wiring** — `bundler.buildDeployer`'s `DeployerFlux` case now calls `collectComponentPreManifests` alongside `collectComponentManifests` and threads both into the flux generator.
- **`<name>-pre` emission** — when `ComponentPreManifests[ref.Name]` is non-empty, `generateComponentResources` emits a `<name>-pre` HelmRelease via the existing `generateManifestHelmChart` helper (same path used for `-post`) and rewires the primary's `dependsOn` to `<name>-pre`. Chain becomes `previous → <name>-pre → <name> → <name>-post → next`.
- **Terminal-aware `dependsOn`** — replace free `buildDependsOn(sortedRefs, index)` with method `g.buildPrimaryDependsOn`, backed by a new free function `terminalReleaseNameFor(ref, postManifests)` that returns `<name>-post` for mixed components and `<name>` otherwise. The README renderer (`buildComponentSummaries`) uses the same helper, so the rendered README matches the actual chain.
- **Collision guard** — at `Generate()` entry, a recipe declaring both `foo` (with pre-manifests) and a separate `foo-pre` is rejected at bundle time. Mirrors the rule in `pkg/bundler/deployer/localformat/writer.go`.
- **No localformat re-use** — flux has its own template-driven path and doesn't go through `localformat`, so the synthesis lives directly in `flux.go`. Same applies to `validatePreManifestNamespace`; flux doesn't yet have an os-talos-shaped consumer, so the namespace-drift guard isn't ported. Can be added later if a flux user needs it.

## Testing

```bash
make qualify   # all 21 chainsaw tests pass, including cli-bundle-flux
```

New / updated tests in `pkg/bundler/deployer/flux/flux_test.go`:

- `TestBuildPrimaryDependsOn` (4 sub-cases) — exercises terminal resolution for mixed, manifest-only, and chart-only previous components. Replaces `TestBuildDependsOn` since `buildDependsOn` is gone.
- `TestGenerate_WithPreManifests` — pre-only component emits `<name>-pre/` with Chart.yaml + templates + HelmRelease, pre HelmRelease depends on previous component, primary depends on `<name>-pre` (not previous), root kustomization references the pre folder.
- `TestGenerate_PreAndPostManifests` — full pre + post + downstream-next shape. Asserts `pre` is head of chain (no dependsOn), primary depends on `<name>-pre`, post depends on `<name>`, **and the next component depends on `<name>-post`** (regression guard for Codex's finding).
- `TestGenerate_PreManifestsCollision` — recipe declaring both `foo` (with pre-manifests) and `foo-pre` is rejected with an error naming the offending pair.

Coverage: `pkg/bundler/deployer/flux` 83.5% → 83.9% (+0.4%).

## Risk Assessment

- [x] **Low** — Isolated to the flux deployer; non-GKE / non-pre-manifest recipes generate the same bytes as before (terminal-aware `dependsOn` only changes output for *mixed* previous components, which previously had the next release racing the post anyway — a latent correctness improvement, not a behavior change for the common case). Well-tested, easy to revert.

**Rollout notes:** For flux bundles where the previous component is mixed (chart + post-manifests), the next component's `dependsOn` will now resolve to `<prev>-post` rather than `<prev>`. This is the documented chain — operators tracking reconcile order should see the expected sequencing rather than parallel reconcile of the next component with the previous component's post manifests. No CLI flags, no API surface changes.

## Checklist

### DIFF
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -1210,7 +1210,7 @@ All deployers respect the `deploymentOrder` field from the recipe, ensuring comp
 
 - **Helm**: Components listed in README in deployment order
 - **Argo CD**: Uses `argocd.argoproj.io/sync-wave` annotation (0 = first, 1 = second, etc.)
-- **Flux**: Uses `dependsOn` references in HelmRelease/Kustomization CRs (each component depends on its predecessor)
+- **Flux**: Uses `dependsOn` references in HelmRelease CRs (each component depends on the previous component's terminal release — its `<prev>-post` release when post-manifests are present, otherwise `<prev>`). Components with pre-manifests insert a `<name>-pre` release that the primary HelmRelease depends on, so the chain becomes `previous → <name>-pre → <name> → <name>-post → next`. The bundle's root `kustomization.yaml` is a plain Kustomize file (not a Flux Kustomization CR).
 - **Helmfile**: Uses `needs:` references in each release (each component depends on its predecessor)
 
 #### Value Overrides
--- a/pkg/bundler/bundler.go
+++ b/pkg/bundler/bundler.go
@@ -387,26 +387,28 @@ func (b *DefaultBundler) buildDeployer(ctx context.Context, recipeResult *recipe
 		}, nil
 
 	case config.DeployerFlux:
+		componentPreManifests, preErr := b.collectComponentPreManifests(ctx, recipeResult)
+		if preErr != nil {
+			return nil, errors.PropagateOrWrap(preErr, errors.ErrCodeInternal,
+				"failed to collect component pre-manifests")
+		}
 		componentManifests, manifestErr := b.collectComponentManifests(ctx, recipeResult)
 		if manifestErr != nil {
-			var se *errors.StructuredError
-			if stderrors.As(manifestErr, &se) {
-				return nil, manifestErr
-			}
-			return nil, errors.Wrap(errors.ErrCodeInternal,
-				"failed to collect component manifests", manifestErr)
+			return nil, errors.PropagateOrWrap(manifestErr, errors.ErrCodeInternal,
+				"failed to collect component manifests")
 		}
 		return &flux.Generator{
-			RecipeResult:       recipeResult,
-			ComponentValues:    componentValues,
-			Version:            b.Config.Version(),
-			RepoURL:            b.Config.RepoURL(),
-			TargetRevision:     b.Config.TargetRevision(),
-			IncludeChecksums:   b.Config.IncludeChecksums(),
-			DataFiles:          dataFiles,
-			ComponentManifests: componentManifests,
-			DynamicValues:      dynamicValues,
-			VendorCharts:       b.Config.VendorCharts(),
+			RecipeResult:          recipeResult,
+			ComponentValues:       componentValues,
+			Version:               b.Config.Version(),
+			RepoURL:               b.Config.RepoURL(),
+			TargetRevision:        b.Config.TargetRevision(),
+			IncludeChecksums:      b.Config.IncludeChecksums(),
+			DataFiles:             dataFiles,
+			ComponentPreManifests: componentPreManifests,
+			ComponentManifests:    componentManifests,
+			DynamicValues:         dynamicValues,
+			VendorCharts:          b.Config.VendorCharts(),
 		}, nil
 
 	case config.DeployerHelmfile:
--- a/pkg/bundler/deployer/flux/flux.go
+++ b/pkg/bundler/deployer/flux/flux.go
@@ -44,6 +44,11 @@ const (
 	fileReadme        = "README.md"
 )
 
+// summaryTypeHelmRelease is the value placed in ComponentSummary.Type for
+// every row of the README's Components table — every emitted release
+// (pre / primary / post) is a Flux HelmRelease CR.
+const summaryTypeHelmRelease = "HelmRelease"
+
 //go:embed templates/configmap-values.yaml.tmpl
 var configMapTemplate string
 
@@ -125,6 +130,14 @@ type Generator struct {
 	// components. Components without manifests do not appear in the map.
 	ComponentManifests map[string]map[string][]byte
 
+	// ComponentPreManifests maps component name → manifest path → rendered bytes.
+	// Emitted as a <name>-pre HelmRelease that the primary HelmRelease
+	// dependsOn, ensuring pre-phase manifests reconcile before the chart.
+	// Wired by the bundler from ComponentRef.PreManifestFiles and the
+	// synthesized GKE critical-priority ResourceQuota (see issue #915).
+	// Components without pre-manifests do not appear in the map.
+	ComponentPreManifests map[string]map[string][]byte
+
 	// DynamicValues maps component names to their dynamic value paths.
 	// When non-empty, dynamic paths are split from inline values into a
 	// ConfigMap and referenced via spec.valuesFrom in the HelmRelease.
@@ -202,6 +215,10 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 		}
 	}
 
+	if err := g.detectInjectedReleaseCollisions(sortedRefs); err != nil {
+		return nil, err
+	}
+
 	// Create sources directory.
 	sourcesDir, err := deployer.SafeJoin(outputDir, "sources")
 	if err != nil {
@@ -258,7 +275,7 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 	// Write README.md.
 	readmeData := ReadmeData{
 		BundlerVersion: deployer.NormalizeVersionWithDefault(g.Version),
-		Components:     buildComponentSummaries(sortedRefs, g.ComponentManifests),
+		Components:     buildComponentSummaries(sortedRefs, g.ComponentPreManifests, g.ComponentManifests),
 	}
 	if err := writeTemplate(output, readmeTemplate, readmeData,
 		outputDir, fileReadme, "failed to write README.md"); err != nil {
@@ -337,9 +354,36 @@ func (g *Generator) generateComponentResources(ctx context.Context, ref recipe.C
 			fmt.Sprintf("failed to create component directory %s", ref.Name), err)
 	}
 
-	dependsOn := buildDependsOn(sortedRefs, index)
+	primaryDependsOn := g.buildPrimaryDependsOn(sortedRefs, index)
+	hasPreManifests := len(g.ComponentPreManifests[ref.Name]) > 0
 	hasManifests := len(g.ComponentManifests[ref.Name]) > 0
 	var resources []string
+
+	// Emit the <name>-pre HelmRelease BEFORE the primary when pre-manifests
+	// exist, and rewire the primary's dependsOn to point at the pre release.
+	// Chain becomes: previous → <name>-pre → <name> → <name>-post → next.
+	if hasPreManifests {
+		preName := ref.Name + "-pre"
+		preDir, preDirErr := deployer.SafeJoin(outputDir, preName)
+		if preDirErr != nil {
+			return nil, preDirErr
+		}
+		if mkErr := os.MkdirAll(preDir, 0750); mkErr != nil {
+			return nil, errors.Wrap(errors.ErrCodeInternal,
+				fmt.Sprintf("failed to create pre directory %s", preName), mkErr)
+		}
+		preWroteCM, preErr := g.generateManifestHelmChart(ref.Name, preName, ref.Namespace, preDir,
+			g.ComponentPreManifests[ref.Name], gitSources, primaryDependsOn, output)
+		if preErr != nil {
+			return nil, preErr
+		}
+		resources = append(resources, filepath.Join(preName, fileHelmRelease))
+		if preWroteCM {
+			resources = append(resources, filepath.Join(preName, fileConfigMap))
+		}
+		// Primary chart now waits for the pre release.
+		primaryDependsOn = []DependsOnRef{{Name: preName}}
+	}
 
 	switch ref.Type { //nolint:exhaustive // only Helm is supported; default rejects others
 	case recipe.ComponentTypeHelm:
@@ -347,15 +391,15 @@ func (g *Generator) generateComponentResources(ctx context.Context, ref recipe.C
 		// Package as a local Helm chart so Flux renders the templates natively.
 		if ref.Chart == "" && ref.Source == "" && hasManifests {
 			wroteCM, genErr := g.generateManifestHelmChart(ref.Name, ref.Name, ref.Namespace, compDir,
-				g.ComponentManifests[ref.Name], gitSources, dependsOn, output)
+				g.ComponentManifests[ref.Name], gitSources, primaryDependsOn, output)
 			if genErr != nil {
 				return nil, genErr
 			}
-			res := []string{filepath.Join(ref.Name, fileHelmRelease)}
+			resources = append(resources, filepath.Join(ref.Name, fileHelmRelease))
 			if wroteCM {
-				res = append(res, filepath.Join(ref.Name, fileConfigMap))
+				resources = append(resources, filepath.Join(ref.Name, fileConfigMap))
 			}
-			return res, nil
+			return resources, nil
 		}
 
 		// Vendored Helm component: pull chart tarball, write wrapper,
@@ -364,7 +408,7 @@ func (g *Generator) generateComponentResources(ctx context.Context, ref recipe.C
 		// manifests (the existing flow handles them correctly).
 		if g.VendorCharts && isVendorable(ref) {
 			wroteCM, rec, vendErr := g.generateVendoredHelmComponent(
-				ctx, ref, compDir, dependsOn, gitSources, puller, output)
+				ctx, ref, compDir, primaryDependsOn, gitSources, puller, output)
 			if vendErr != nil {
 				return nil, vendErr
 			}
@@ -377,7 +421,7 @@ func (g *Generator) generateComponentResources(ctx context.Context, ref recipe.C
 				"component", ref.Name,
 				"chart", rec.Chart, "version", rec.Version, "sha256", rec.SHA256)
 		} else {
-			wroteCM, helmErr := g.generateHelmComponent(ref, compDir, dependsOn, helmSources, output)
+			wroteCM, helmErr := g.generateHelmComponent(ref, compDir, primaryDependsOn, helmSources, output)
 			if helmErr != nil {
 				return nil, helmErr
 			}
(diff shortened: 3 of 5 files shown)


pkg/bundler/deployer/flux/flux.go
```diff
@@ -202,6 +210,26 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
...
+	}
+	for _, ref := range sortedRefs {
+		if len(g.ComponentPreManifests[ref.Name]) == 0 {
+			continue
+		}
+		if _, clash := declared[ref.Name+"-pre"]; clash {
+			return nil, errors.New(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("component %q has preManifestFiles and would inject %q-pre, but a component named %q-pre is already declared in the recipe — rename one to avoid collision",
+					ref.Name, ref.Name, ref.Name))
+		}
```
mchmarny: Medium: parity with localformat is half-implemented. `pkg/bundler/deployer/localformat/writer.go` declares BOTH `-pre` and `-post` collision rules (search "would inject"). The flux deployer just gained the `-pre` check here but has been silently missing the `-post` check since post-manifests landed — a recipe declaring `foo` (with post-manifests) and `foo-post` would collide just like the `-pre` case.

Since the docstring above advertises mirror parity with localformat, the matching `-post` block is a small addition that closes the gap and makes the docstring accurate. Out-of-scope for the literal issue, but the parity claim makes the omission stand out.

pkg/bundler/deployer/flux/flux.go
```diff
@@ -364,7 +419,7 @@ func (g *Generator) generateComponentResources(ctx context.Context, ref recipe.C
 		// manifests (the existing flow handles them correctly).
 		if g.VendorCharts && isVendorable(ref) {
 			wroteCM, rec, vendErr := g.generateVendoredHelmComponent(
-				ctx, ref, compDir, dependsOn, gitSources, puller, output)
+				ctx, ref, compDir, primaryDependsOn, gitSources, puller, output)
```
mchmarny, line 411: Medium: no test coverage for `vendored chart + pre-manifests`. `g.generateVendoredHelmComponent(..., primaryDependsOn, ...)` now receives the rewired `-pre` reference, but `TestGenerate_WithPreManifests` and `TestGenerate_PreAndPostManifests` only exercise the non-vendored path. A regression that broke the rewire on the vendored branch (e.g., a future refactor that re-shadowed `primaryDependsOn`) would slip through.

A small test enabling `VendorCharts: true` with a pre-manifest on a vendorable chart would close the gap.

pkg/bundler/bundler.go
```diff
@@ -397,16 +402,17 @@ func (b *DefaultBundler) buildDeployer(ctx context.Context, recipeResult *recipe
...
+			TargetRevision:        b.Config.TargetRevision(),
+			IncludeChecksums:      b.Config.IncludeChecksums(),
+			DataFiles:             dataFiles,
+			ComponentPreManifests: componentPreManifests,
+			ComponentManifests:    componentManifests,
+			DynamicValues:         dynamicValues,
+			VendorCharts:          b.Config.VendorCharts(),
 		}, nil
 
 	case config.DeployerHelmfile:
```
mchmarny, line 414: nit: inconsistent error-wrapping pattern within this same case block — the new `componentPreManifests` uses `errors.PropagateOrWrap` (matching the helm / helmfile / argocd cases above) while the existing `componentManifests` immediately below still uses the older `errors.As + errors.Wrap` form.

Both are functionally equivalent, but the asymmetry within one switch arm reads as a leftover. Trivial follow-up to switch the post-manifests block to `errors.PropagateOrWrap` too.

pkg/bundler/deployer/flux/flux.go
```diff
@@ -470,15 +525,34 @@ func filterEnabled(refs []recipe.ComponentRef) []recipe.ComponentRef {
...
+// Pre-manifests never extend the tail — they live BEFORE the primary.
+//
+// Free function so the README renderer (buildComponentSummaries) can reuse it
+// without a Generator handle.
+func terminalReleaseNameFor(ref recipe.ComponentRef, postManifests map[string]map[string][]byte) string {
+	hasChart := ref.Chart != "" || ref.Source != ""
+	if hasChart && len(postManifests[ref.Name]) > 0 {
+		return ref.Name + "-post"
+	}
+	return ref.Name
```
mchmarny, line 577: Nice catch on the pre-existing terminal-aware bug. The `hasChart` guard correctly excludes the manifest-only path (which folds manifests into the primary and never emits `-post`), and the function being a free function lets the README renderer reuse it without dragging the `*Generator` into `buildComponentSummaries`. Clean separation.

mchmarny (commented):
Solid fix that closes the gap PR #921 left for the flux deployer, plus a real win on the pre-existing terminal-aware `dependsOn` bug that Codex caught. Test coverage for the main paths is thorough, docs are updated in sync, and the collision guard pattern is the right shape.

Two medium notes worth considering before merge: the `-post` half of the localformat-parity collision check is still missing in flux (latent pre-existing gap, but the new docstring makes the omission visible), and the vendored-chart + pre-manifests path is exercised by the code but not by tests. Plus one nit on error-wrapping inconsistency in `bundler.go`. None are blockers.

CI looks healthy — `tests / E2E` and `analyze` are still running but every completed check is green.

docs/user/cli-reference.md
```diff
@@ -1210,7 +1210,7 @@ All deployers respect the `deploymentOrder` field from the recipe, ensuring comp
 
 - **Helm**: Components listed in README in deployment order
 - **Argo CD**: Uses `argocd.argoproj.io/sync-wave` annotation (0 = first, 1 = second, etc.)
-- **Flux**: Uses `dependsOn` references in HelmRelease/Kustomization CRs (each component depends on its predecessor)
+- **Flux**: Uses `dependsOn` references in HelmRelease/Kustomization CRs (each component depends on the previous component's terminal release — its `<prev>-post` release when post-manifests are present, otherwise `<prev>`). Components with pre-manifests insert a `<name>-pre` release that the primary HelmRelease depends on, so the chain becomes `previous → <name>-pre → <name> → <name>-post → next`.
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 2693e411 — dropped `/Kustomization` and added a clarifying sentence that the root `kustomization.yaml` is a plain Kustomize file, not a Flux Kustomization CR.

pkg/bundler/deployer/flux/flux.go
```diff
@@ -535,8 +609,10 @@ func buildComponentSummaries(sortedRefs []recipe.ComponentRef, manifests map[str
 		}
 
 		dependsOnStr := "-"
-		if deps := buildDependsOn(sortedRefs, i); len(deps) > 0 {
-			dependsOnStr = deps[0].Name
+		// README mirrors the actual chain ordering — terminal-aware so the
+		// previous component's -post is named when it exists.
+		if i > 0 {
+			dependsOnStr = terminalReleaseNameFor(sortedRefs[i-1], manifests)
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 2693e411 — `buildComponentSummaries` now takes both pre and post manifest maps, emits a `<name>-pre` HelmRelease summary row when pre-manifests exist, and rewires the primary row to depend on it. `TestGenerate_PreAndPostManifests` got new README-content assertions that pin the rendered table against the actual chain (primary `dependsOn` is `gpu-operator-pre`, next component is `gpu-operator-post`).

pkg/bundler/deployer/flux/flux.go
yuanchen8911: Fixed in 2693e411 — added the matching `-post` collision rule alongside `-pre`. Mirrors `pkg/bundler/deployer/localformat/writer.go` exactly: same "would inject" wording, same skip-when-manifest-only carve-out (manifest-only components fold manifests into the primary release and never emit a `-post` folder, so they cannot collide). New `TestGenerate_PostManifestsCollision` covers the rule.

Also factored the whole block into `(*Generator).detectInjectedReleaseCollisions` so `Generate()` stays under the `funlen` 70-statement cap.

pkg/bundler/deployer/flux/flux.go
yuanchen8911: Fixed in 2693e411 — added `TestGenerate_VendoredChartWithPreManifests` which runs the vendored path (`VendorCharts: true` + `stubChartPuller`) with a pre-manifest on `gpu-operator`. Asserts the `gpu-operator-pre/helmrelease.yaml` is emitted, the vendored wrapper primary depends on `gpu-operator-pre`, and the wrapper still references its local `GitRepository` source (sanity check that the vendored branch was actually exercised).

pkg/bundler/bundler.go
yuanchen8911: Fixed in 2693e411 — switched the post-manifests block to `errors.PropagateOrWrap` so both calls in the `DeployerFlux` arm use the same form. `stderrors` is still imported (used by other deployer cases above this one).

pkg/bundler/deployer/flux/flux.go
yuanchen8911: Thanks\! Yeah, factoring it out into a free function fell out naturally once the README renderer also needed the same logic — keeping it free meant `buildComponentSummaries` could stay a pure helper rather than gaining a `*Generator` receiver just for one lookup. The manifest-only carve-out is the same one used in `generateComponentResources` (the manifest-only branch returns before the `-post` injection runs), so the README/code/dependsOn all agree on the same "terminal = `-post` iff chart + post-manifests" rule.

yuanchen8911:
@mchmarny — all 6 inline comments addressed at `2693e411` and replied inline. Ready for another look when you have a minute.

| # | Reviewer | File:line | Finding | Fix |
|---|----------|-----------|---------|-----|
| 1 | CodeRabbit | `cli-reference.md:1213` | "HelmRelease/Kustomization CRs" — flux doesn't emit Kustomization CRs | Dropped "/Kustomization", added clarifying sentence about plain `kustomization.yaml` |
| 2 | CodeRabbit | `flux.go:615` | README missing `<name>-pre` row + wrong primary `dependsOn` | `buildComponentSummaries` now takes both manifest maps, emits `-pre` row, rewires primary. README assertions added to existing test. |
| 3 | Mark | `flux.go:230` | `-pre` collision rule landed but `-post` parity missing (latent localformat gap) | Added `-post` collision rule with same wording; `TestGenerate_PostManifestsCollision` covers it |
| 4 | Mark | `flux.go:422` | Vendored chart + pre-manifests path untested | New `TestGenerate_VendoredChartWithPreManifests` exercises `VendorCharts: true` with pre-manifest |
| 5 | Mark | `bundler.go:418` | Error-wrapping inconsistency within `DeployerFlux` arm | Switched post-manifests block to `errors.PropagateOrWrap` for symmetry |
| 6 | Mark | `flux.go:555` | Positive feedback on terminal-aware fix | Acknowledged; no code change needed |

Bonus refactor: collision detection extracted into `(*Generator).detectInjectedReleaseCollisions` so `Generate()` stays under the `funlen` 70-statement cap. `make qualify` clean (21/21 chainsaw, no vulns, lint clean).

pkg/bundler/deployer/flux/flux_test.go
```diff
@@ -1647,3 +1664,244 @@ func extractSourceName(t *testing.T, yamlContent string) string {
...
+		Version: "v0.9.0",
+	}
+
+	_, err := g.Generate(ctx, outputDir)
+	if err == nil {
+		t.Fatal("expected collision error, got nil")
+	}
+	if !strings.Contains(err.Error(), `would inject "foo"-pre`) {
+		t.Errorf("expected collision error to name the offending pair, got: %v", err)
+	}
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 7194d8fc — added `stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, ""))` assertions to both `TestGenerate_PreManifestsCollision` and `TestGenerate_PostManifestsCollision` ahead of the existing message-substring check. The substring check is kept because it asserts the *content* (which name actually collided), which the code-only check cannot verify.

Used the project-idiomatic form (`stderrors.Is` against a `*StructuredError` constructed with the target code) rather than `errors.As` + manual `.Code` comparison, per the rule in `.claude/CLAUDE.md`:

> **Code-based matching with `errors.Is`:** `*StructuredError.Is` reports a match when the target is a `*StructuredError` with the same `Code`. Prefer this over `errors.As` + manual code comparison.

yuanchen8911:
Adding a 7th item I missed in the table above — CodeRabbit also posted [a nit](https://github.com/NVIDIA/aicr/pull/924#discussion_r3249883908) on `flux_test.go:1970` after my first force-push.

| # | Reviewer | File:line | Finding | Fix |
|---|----------|-----------|---------|-----|
| 7 | CodeRabbit | `flux_test.go:1970` | Collision tests asserted error type via `strings.Contains(err.Error(), ...)` rather than structured error code | Added `stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, ""))` to both `TestGenerate_PreManifestsCollision` and `TestGenerate_PostManifestsCollision` ahead of the existing substring check (substring kept — it asserts *which* name collided). Used `*StructuredError.Is` form per `.claude/CLAUDE.md` rule, not the `errors.As` suggestion. |

Branch is now at `7194d8fc`. `make qualify` clean.

mchmarny (approved):

PR #926, opened 2026-05-15
fix(bundler): stratify helmfile bundle by DAG level (#914)

### LINKED ISSUE #914: [Bug]: helmfile bundle fails on fresh cluster — CRD/CR render order between nodewright-operator and nodewright-customizations
### Summary

On a fresh cluster, `helmfile apply` (and `helmfile diff`) against an AICR helmfile bundle aborts before installing anything because `helm-diff` renders `nodewright-customizations` (which uses `kind: Skyhook` from `skyhook.nvidia.com/v1alpha1`) against live cluster discovery, and the `Skyhook` CRD is not yet registered — it is installed by `nodewright-operator`, which `helmfile` would install first by `needs:`, but the diff pre-pass runs before any release lands.

Same root cause for both `diff` and `apply`. `helmfile sync` would avoid it (no diff plugin), and `helmfile apply --skip-diff-on-install` works around it (current demo uses the latter).

### Repro

1. Provision a fresh GKE H100 cluster (e.g. via `gh workflow run uat-gcp.yaml -f skip_delete=true -f skip_tests=true`).
2. Generate a helmfile-deployer bundle:

   ```yaml
   spec:
     bundle:
       deployment:
         deployer: helmfile
   ```

   ```shell
   aicr bundle --config aicr-config.yaml
   ```

3. Install the `helm-diff` plugin: `helm plugin install https://github.com/databus23/helm-diff`.
4. From `./bundle/`, run `helmfile apply` (or `helmfile diff`).

### Observed

```
in ./helmfile.yaml: command \"/opt/homebrew/bin/helm\" exited with non-zero status:

ARGS:
  0: helm
  1: diff
  2: upgrade
  3: --allow-unreleased
  4: nodewright-customizations
  5: 006-nodewright-customizations
  ...
  8: --namespace
  9: skyhook

STDERR:
  Error: Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: resource mapping not found for name: \"tuning\" namespace: \"skyhook\" from \"\": no matches for kind \"Skyhook\" in version \"skyhook.nvidia.com/v1alpha1\"
  ensure CRDs are installed first
  Error: plugin \"diff\" exited with error
```

helmfile's `needs:` correctly orders `nodewright-operator` → `nodewright-customizations`, but `helm-diff` performs its render pass against the live cluster's REST mapper *before* any release is installed, so the dependent chart can't resolve the kind.

### Expected

`helmfile apply` against a freshly bundled output should succeed on a fresh cluster with no flag gymnastics. A user following [`demos/cuj1-gke-config.md`](../tree/main/demos/cuj1-gke-config.md) shouldn't need to know about `--skip-diff-on-install`, plugin internals, or CRD/CR ordering quirks.

### Workarounds (in increasing invasiveness)

- **A. CLI flag, no bundle change:** `helmfile apply --skip-diff-on-install`. helmfile bypasses `helm diff` for releases that don't yet exist, so the first install renders against the in-progress cluster (CRDs from `nodewright-operator` land before `nodewright-customizations` is rendered). Subsequent re-applies on the populated cluster diff normally. *Currently used in `demos/cuj1-gke-config.md`.*
- **B. Two-phase install:** `helmfile -l name=nodewright-operator sync` first, then `helmfile apply`. Explicit, no flag, but adds a step.
- **C. Bundle-side fix (preferred):** stop emitting a single helmfile that mixes CRD-owner and CRD-consumer charts. Options:
  - **C1.** Generate a multi-`helmfiles:` document with a `helmfile.yaml` that pulls a `crds.yaml` sub-helmfile first (helmfile renders sub-helmfiles sequentially). `nodewright-operator` (and any other CRD-owning charts — `cert-manager`, `kube-prometheus-stack`, `nvsentinel`, etc.) moves into the CRD layer.
  - **C2.** Continue emitting one `helmfile.yaml` but tag releases with `crd: true` on CRD owners and have the template wrap them in a `helmfiles:` block, so `--skip-diff-on-install` is the documented norm only as a re-apply fallback.
  - **C3.** Pre-install CRDs out-of-band via a generated `crds/` directory + a small `apply-crds.sh` the bundle README invokes before `helmfile apply`. Mirrors what some upstream operator charts ship as a separate install path.

C1 is the most idiomatic helmfile pattern and survives future additions of CRD-providing charts.

### Affected files

- `pkg/bundler/deployer/helmfile/` — `releases.go` (release graph), `templates/helmfile.yaml.tmpl` (top-level shape), `templates/README.md.tmpl` (currently documents `helmfile diff` / `helmfile apply` with no plugin or CRD-ordering caveat).
- `recipes/registry.yaml` / `recipes/overlays/*` — any component that ships a CRD that other components reference would feed into the same bug. `nodewright-operator` is the one tripping today; the broader set includes any chart whose `helm install` registers a CRD that another chart's templates use directly.
- `demos/cuj1-gke-config.md` — already updated to `--skip-diff-on-install`; revert the workaround when this issue lands.

### Operational considerations

- Bundle reproducibility shouldn't change — same chart versions, same values, just different `helmfile.yaml` layout.
- The `helm-diff` plugin is now effectively mandatory for any helmfile bundle (helmfile `apply` calls it under the hood). Worth surfacing in the generated README regardless of how this is resolved.

### Severity

User-visible: demo-blocking on a fresh cluster, and the workaround is non-obvious. Low data-loss risk (we abort before mutating cluster state), high time-to-debug for a first-time user.

Maybe, with guidance — happy to PR the README/template + `--skip-diff-on-install` hint immediately; the C1 bundle-side restructure benefits from a maintainer signoff on the preferred layout.

lockwobr:
## Summary

Replace #922's 2-way `crds.yaml` / `releases.yaml` split with an N-way DAG-stratified
layout: one sub-helmfile per topological level (`level-0.yaml` … `level-N.yaml`).
Cross-chart CRD ordering is now handled by `dependencyRefs` alone; the
`installsCRDs` flag is removed. Self-reference cases (CRD + CR in the same
release) get a per-release `disableValidation: true` driven by a new
`hasSelfRefCRDs` registry field.

## Motivation / Context

Issue #914 is the helm-diff "CRD not yet registered" race that fires on a fresh
cluster when a chart's templates reference CRDs installed by a sibling (or by
the same chart). #922 fixed the cross-chart case for a hand-marked subset of
charts via a 2-way sub-helmfile split. In live testing on kind we discovered
several gaps:

- nvsentinel was incorrectly marked `installsCRDs: true` (no in-bundle consumer
  of HealthCheck / NodeHealth CRs), causing it to land in `crds.yaml` and race
  against prometheus-operator-crds' diff in the same tier.
- network-operator's `NodeFeatureRule` CR depends on nfd's CRD, which wasn't
  marked.
- gpu-operator, kai-scheduler, and kubeflow-trainer all self-reference CRDs in
  their own `crds/` directory from their own templates — a failure mode the
  2-way split can't solve regardless of which tier they land in.
- kube-prometheus-stack's CRDs ship in a subchart `templates/`, so its own
  templates render against an unregistered REST mapper even on fresh installs.

The fix is to stop hand-marking charts and let the DAG do the ordering. A
chart that produces CRDs and a chart that consumes them are already related
via `dependencyRefs`; the bundler now reads that graph directly.

Fixes: #914
Related: #922

## Type of Change

- [x] Refactoring (no functional changes — installsCRDs removal, levels infra)

## Component(s) Affected

## Implementation Notes

**`recipe.TopologicalLevels`** (new in `pkg/recipe/metadata.go`): Kahn variant
that drains the entire current zero-in-degree generation as one batch before
processing its dependents. Returns `[][]string` where every component in level
`i` is mutually independent (no edges among them) and depends only on
components in levels `<i`. 14 unit-test cases cover cycle detection
(self-loop, two- and three-node), missing-dep surfacing as cycle, diamond,
wide root, the realistic AICR-shape, alphabetical tie-break within a level,
and a cross-check against `TopologicalSort` that flattened levels respect
every `dependencyRefs` edge.

**Helmfile bundler refactor** (`pkg/bundler/deployer/helmfile/`):

- `splitFoldersByCRD` → `splitFoldersByLevel`. N-way partition by parent
  component's DAG level. Folders with unknown parents default to level 0
  (defensive).
- `writeHelmfileLayout` emits one sub-helmfile per non-empty level
  (`level-0.yaml`, `level-1.yaml`, ...) and a top-level `helmfile.yaml`
  carrying `helmfiles: [level-0.yaml, ...]`. Single-level bundles collapse
  to a single `helmfile.yaml` (no sub-helmfile overhead).
- Cross-level `needs:` edges dissolve because helmfile processes the
  `helmfiles:` list sequentially. Within a level there's still a `needs:`
  chain for deterministic install order.

**Registry schema:**

- New `hasSelfRefCRDs: bool` field. Drives per-release `disableValidation: true`
  in helmfile. Marked components: `gpu-operator` (ClusterPolicy template +
  CRD), `kai-scheduler` (default-queue / default-shard), `kubeflow-trainer`
  (default ClusterTrainingRuntimes).
- `installsCRDs` field removed. The DAG-level partition makes it redundant
  for placement, and helmfile v1 doesn't accept `args:` per release (so we
  can't piggyback SSA on it). Removed from `ComponentConfig`, every registry
  entry, and `TestRegistry_CRDOwnersStillMarked`.

**kube-prometheus-stack split:**

- New `prometheus-operator-crds` component (chart `28.0.1`, matches operator
  `v0.90.1` pinned by kube-prometheus-stack `84.4.0`).
- `kube-prometheus-stack` ships with `crds.enabled: false` and declares
  `dependencyRefs: [prometheus-operator-crds]`, so the operator chart lands
  in a later level than its CRDs.

**Server-side apply NOT emitted:** helmfile v1 removed per-release `args:`,
and `helmDefaults.args` passes flags to helm-diff too (which doesn't accept
`--server-side`). The supported escape hatch is `HELM_DIFF_IGNORE_UNKNOWN_FLAGS=true`
as an env var — explicitly rejected because the bundle should be self-contained.
If the annotation-too-large failure surfaces in practice, follow-up will mark
specific releases with another registry flag.

**Known follow-up (NOT in this PR):**

- `TopologicalSort` (Kahn with per-dequeue tie-break) and `TopologicalLevels`
  (Kahn with generation-drain) produce different flat orderings. Folder
  numbering (001-N, driven by `TopologicalSort`) doesn't strictly match
  install order (driven by levels). Cosmetic — bundle deploys correctly
  either way. Unifying is a separate PR because it touches every chainsaw
  + UAT recipe golden.

## Testing

```bash
# 21/21 chainsaw e2e tests pass
# golangci-lint, yamllint, license headers, MDX, chart-pin checks all clean
# Grype vulnerability scan clean
# Coverage: 75.1% (no per-package regression >0.5%)
```
Live tested against a fresh kind cluster:

- helmfile apply on an eks/h100/training/kubeflow bundle → fixes nvsentinel
PodMonitor race, network-operator NodeFeatureRule race
- gpu-operator ClusterPolicy self-reference resolved via disableValidation: true on the gpu-operator release
- kai-scheduler default-queue / default-shard self-reference resolved likewise
- kubeflow-trainer default ClusterTrainingRuntime self-reference resolved
likewise

Per-package coverage on the affected paths:

- pkg/recipe: 90.2% (added 14 cases for TopologicalLevels)
- pkg/bundler/deployer/helmfile: stable; goldens regenerated for the new
level-N.yaml layout

Risk Assessment

- Medium — Touches recipe, bundler, registry schema, and every
generated bundle layout. The DAG-based partition is a strictly more general
algorithm than the previous flag-based one; existing recipes produce
correct (and structurally simpler) bundles. The schema change is removing
a field, not adding one — recipes that referenced installsCRDs: true
in YAML overlays would be silently ignored on the new schema, so a quick
grep across consumers before merge is worthwhile.

Rollout notes:

- No CLI flag, no API surface change.
- Operators regenerating bundles after merge will see the new
level-N.yaml layout instead of crds.yaml + releases.yaml. Both layouts
are valid helmfile.yaml inputs; existing helmfile apply / helmfile diff invocations against the bundle directory work without changes.
- Any downstream tooling that grepped for crds.yaml / releases.yaml in a
generated bundle needs to be updated to level-*.yaml. (None known
in-tree; chainsaw bundle-helmfile test updated.)

Checklist

- Tests pass locally (make test with -race)
- Linter passes (make lint)
- I did not skip/disable tests to make CI green
- I added/updated tests for new functionality
- I updated docs if user-facing behavior changed
- Changes follow existing patterns in the codebase
- Commits are cryptographically signed (git commit -S)


pkg/bundler/deployer/helmfile/helmfile.go
```diff
@@ -165,9 +162,9 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 		}
 	}
 
-	// Emit either the legacy single-file helmfile.yaml or the split
-	// helmfiles: layout depending on whether any referenced component
-	// is registry-marked InstallsCRDs (issue #914).
+	// Emit either the single-file helmfile.yaml or the stratified
+	// helmfiles: layout depending on whether the dependency DAG
+	// produces more than one level (issue #914).
 	splitLayout, err := g.writeHelmfileLayout(outputDir, output, writeResult.Folders, sortedRefs, namespaceByComponent)
```
mchmarny, line 168: **Medium — stale user-facing README.** The `splitLayout` bool flows into `readmeData.HasCRDLayer` and gates a block in `templates/README.md.tmpl` (lines 26-31) that still tells operators to look for `crds.yaml` and `releases.yaml`:

```
For bundles that include CRD-owning charts,
`helmfile.yaml` is a thin orchestration document with a `helmfiles:` list:
`crds.yaml` is processed first to register CRDs, then `releases.yaml`
installs every chart that consumes them.
```

Those files no longer exist in this layout — the sub-helmfiles are now `level-0.yaml` … `level-N.yaml`. As-is, anyone reading the bundle's own README will be looking for files the bundler doesn't emit. The first sentence ("For bundles that include CRD-owning charts…") is also no longer accurate — any recipe whose DAG produces more than one level will render this block.

Action items:
- Rewrite the template block to describe the stratified layout (one sub-helmfile per dependency level, processed in order).
- Rename `readmeData.HasCRDLayer` (and the related `splitLayout` local) to something like `HasSubHelmfiles` / `stratified` — the current name claims something the field no longer means.
- Regenerate the golden at `pkg/bundler/deployer/helmfile/testdata/upstream_helm_only/README.md` (lines 25-30 carry the same stale text) via `go test -update ./pkg/bundler/deployer/helmfile/...`.

pkg/bundler/deployer/helmfile/releases.go
```diff
@@ -194,6 +206,17 @@ func buildHelmfile(folders []localformat.Folder, namespaceByComponent map[string
 			rel.CreateNamespace = &falseVal
 		}
 
+		// Registry-driven per-release flags. HasSelfRefCRDs emits
+		// Registry-driven per-release flags. HasSelfRefCRDs emits
```
mchmarny, line 210: nit: this comment line is duplicated — `// Registry-driven per-release flags. HasSelfRefCRDs emits` appears twice. Drop one.

pkg/bundler/deployer/helmfile/releases.go
```diff
@@ -194,6 +206,17 @@ func buildHelmfile(folders []localformat.Folder, namespaceByComponent map[string
...
+		// disableValidation: true so helm-diff skips the live-cluster
+		// mapper check for charts whose own templates reference CRDs
+		// they ship in `crds/` (e.g., gpu-operator's clusterpolicy
+		// template). Keyed by f.Parent so injected -pre / -post
+		// folders inherit the same flag.
+		if flags[f.Parent].HasSelfRefCRDs {
```
mchmarny, line 216: Inheriting `disableValidation: true` to every `f.Parent == gpu-operator` folder (primary + injected `-pre`/`-post`) is over-broad and shows up in goldens like `testdata/mixed_gpu_operator/helmfile.yaml` where `gpu-operator-post` (just a RuntimeClass manifest) ends up with `disableValidation: true`. The post-folder is a local-helm wrapper around arbitrary manifests — it doesn't ship the chart's `crds/` and doesn't self-reference any CRD, so there's no reason to disable validation on it.

The trade-off is small (only the post-folder loses helm-diff's mapper sanity check), so this is fine to defer if it simplifies the code, but worth a comment acknowledging the choice. Today the inline comment says "so injected -pre / -post folders inherit the same flag" without explaining why we *want* that inheritance — it reads as accidental coupling rather than intentional.

recipes/registry.yaml
```diff
@@ -253,9 +255,13 @@ components:
...
-    # Registers HealthCheck / NodeHealth CRDs consumed by sibling
-    # observability and triage components. Issue #914.
-    installsCRDs: true
+    # nvsentinel's templates create monitoring.coreos.com/v1 PodMonitor
+    # CRs that prometheus-operator-crds must register first. Ordering
+    # is satisfied transitively in the DAG-stratified helmfile layout:
+    # nvsentinel depends on gpu-operator, which depends on
+    # kube-prometheus-stack, which depends on prometheus-operator-crds,
+    # so by the time nvsentinel's level diffs, the monitoring CRDs are
+    # already registered. Issue #914.
```
mchmarny, line 264: **Medium — brittle transitive chain.** This ordering relies on a three-hop transitive path: nvsentinel → gpu-operator → kube-prometheus-stack → prometheus-operator-crds. If anyone removes the `gpu-operator → kube-prometheus-stack` edge in `recipes/overlays/base.yaml` (it's plausible — gpu-operator doesn't strictly need kube-prometheus-stack to install, only to scrape metrics), nvsentinel's helm-diff race against PodMonitor CRD silently returns and there's no test that catches it.

The whole PR's thesis is "let dependencyRefs encode the ordering" — applying that consistently means nvsentinel should declare a direct `dependencyRefs: [prometheus-operator-crds]` (in addition to its existing edges). Same applies to any other component whose templates create monitoring.coreos.com CRs. The DAG-flattening makes the extra edge a no-op when the transitive path already exists, but it survives refactors of the intermediate nodes.

Worth doing in this PR while the rationale is fresh; otherwise drop a "TODO: harden direct edges for monitoring-CR creators" so the gap is documented.

mchmarny (commented):
The DAG-stratification thesis is right: removing the hand-marked `installsCRDs` flag and letting `dependencyRefs` drive ordering is structurally simpler and handles the cross-chart cases that #922's 2-way split couldn't. `TopologicalLevels` (with the level-by-level Kahn batching) is well-tested, and the chainsaw test was updated to walk the new layout. Per-release `disableValidation: true` is the right call for self-reference cases like gpu-operator and kai-scheduler.

Findings:

- **Medium — stale bundle README.** The template at `pkg/bundler/deployer/helmfile/templates/README.md.tmpl` lines 26-31 still references `crds.yaml` / `releases.yaml`. Every multi-level bundle ships an inaccurate operator-facing README. The `HasCRDLayer` field that gates it also needs renaming. Details inline at `helmfile.go:168`.
- **Medium — brittle transitive chain for monitoring CRDs.** nvsentinel relies on a three-hop dep path to prometheus-operator-crds; a future refactor of an intermediate node silently re-introduces the helm-diff race. Worth a direct edge or at least a TODO. Details inline at `recipes/registry.yaml`.
- **Low** — `releases.go:216` cascades `disableValidation: true` to injected `-pre`/`-post` folders of a self-ref-CRDs parent (visible in `mixed_gpu_operator` golden). Probably intentional but the inline rationale doesn't explain *why*. 
- **Nit** — duplicated comment line in `releases.go:209-210`. Stale `crds.yaml`/`releases.yaml` text also lingers in `helmfile.go:363-365` and `helmfile_test.go:753,790` (those lines are outside the diff hunks, mentioning here for the sweep).

Nothing blocks merge; the README fix is worth catching before this lands since it ships into every bundle. CI Tier 2 jobs were still running at review time.

mchmarny (approved):

Issue #933, opened 2026-05-15
[Bug]: ai-service-metrics validator non-deterministically fails on multi-SG clusters because orchestrator pod can land on a node that can't reach Prometheus


njhensley:
## Bug Description

The `ai-service-metrics` conformance validator non-deterministically fails on clusters that segregate worker and system pods into different Security Groups (or otherwise gate cross-subnet pod-to-pod traffic on port 9090). Whether the validator passes is decided by where the kube-scheduler lands its orchestrator pod — which on a clean cluster is a tie-break, and on a re-used cluster is dominated by image-locality. The root cause is that the validator's Job spec sets `tolerations: [{operator: Exists}]` with no `nodeAffinity` or pod-affinity, so it has no preference for co-locating with the Prometheus pod it queries.

This is reproducible today on DGXC EKS GB200 clusters where customer-workload and system-workload ENIs use separate SGs with asymmetric ingress rules.

## Impact

Medium (workaround exists — fix the cluster SG, or accept that a re-run on a fresh cluster may pass by luck)

## Component

Validator

## Regression?

No — this is the steady-state behavior across all aicr versions where ai-service-metrics has shipped. It only surfaces when the kube-scheduler's tie-break + image-locality history leaves the validator pod on a node whose ENI lives in a SG that can't reach Prometheus.

## Steps to Reproduce

1. Provision a Kubernetes cluster with two ENI configs / two security groups, where the SG attached to the Prometheus pod's node does **not** accept ingress from the SG attached to other worker nodes. (DGXC EKS with separate customer/system ENI subnets is one example.)
2. Deploy a CUJ2 inference bundle (e.g. `gb200-eks-ubuntu-inference-dynamo`).
3. Run `aicr validate --recipe recipe.yaml --phase conformance` repeatedly.

On a fresh cluster, the first run may pass (if the validator pod lands on a system-ENI node by tie-break). Once the validator image is cached on a customer-ENI node, every subsequent run will land there via image-locality scoring and ai-service-metrics will fail.

## Expected Behavior

The validator should deterministically succeed against the same cluster state, regardless of where the kube-scheduler decides to place the orchestrator pod. Either:

- The orchestrator pod is co-scheduled with the Prometheus pod (nodeAffinity / podAffinity), or
- The check is networking-tolerant by routing through the Prometheus Service (it already does — but it relies on the underlying pod-to-pod path being reachable, which in this cluster topology requires same-SG placement).

## Actual Behavior

```
[...]

PR #938, opened 2026-05-15
refactor(gpu-operator): inline dcgm-exporter ConfigMap into values

njhensley:
## Summary

Inline the dcgm-exporter metrics CSV into the gpu-operator chart's `dcgmExporter.config.data` (replacing the separate post-manifest), and document the `ai-service-metrics` Prometheus SG prerequisite for multi-SG EKS clusters.

## Motivation / Context

The upstream gpu-operator chart already renders the dcgm-exporter ConfigMap directly from values when `config.create` is true and `config.data` is non-empty. Shipping the same content as a post-manifest adds a second `kubectl apply` pass after `helm --wait`, which is also the root cause of the kind-overlay deadlock previously worked around by disabling the ConfigMap entirely. Moving the data into values lets Helm own install/upgrade ordering.

The second commit documents a separate cluster-side prerequisite: the conformance check `ai-service-metrics` dials Prometheus over the cluster Service from an orchestrator Job that tolerates every taint and has no node-affinity toward Prometheus, so on clusters with split customer/system ENI subnets (DGXC EKS) the dial can fail non-deterministically when the Job lands on an SG that cannot reach the Prometheus pod. Documents the symmetric `tcp/9090` ingress rule and adds a validation troubleshooting entry.

Fixes: N/A
Related: #933

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Bundlers (`pkg/bundler`, `pkg/component/*`) — test comment only

## Implementation Notes

**Inlining behavior.** The chart gates the ConfigMap on both `config.create == true` and non-empty `config.data`. Setting `data: ""` (as the kind overlay now does alongside `create: false`) suppresses the ConfigMap deterministically.

**Overlay updates.**
- `recipes/overlays/base.yaml`, `recipes/overlays/gke-cos.yaml`, `examples/recipes/aks-training.yaml` — dropped the `manifestFiles: [.../dcgm-exporter.yaml]` reference.
- `recipes/overlays/kind.yaml` — explicit `data: ""` so the chart cannot accidentally render the ConfigMap on simulated GPUs that lack the targeted datacenter metrics.

**Chainsaw assertion rewrite.** The previous assertion globbed for a post-manifest under `*-gpu-operator-post/templates/`. The new assertion verifies the values block carries `name: dcgm-exporter` and the custom metrics list, with a defense-in-depth check that no stray post-manifest is emitted.

**Pre-existing manifestFiles example.** Doc comments in `pkg/recipe/{adapter.go,adapter_test.go,metadata.go}` and the contributor/data-flow/CLI/API docs were using `dcgm-exporter.yaml` as the manifestFiles example; switched the example to `kernel-module-params.yaml` (which is the remaining real example after this change).

**Prometheus prereq docs.** `docs/integrator/eks-dynamo-networking.md` calls out the `tcp/9090` ingress rule as symmetrically required from every worker SG to the Prometheus-hosting SG; `docs/user/validation.md` adds a troubleshooting entry pointing at the SG topology. Tracked at #933 until the validator gains podAffinity toward Prometheus.

## Testing

```bash
# Pending CI verification — not run locally.
# Targeted local checks suggested for reviewers:
make test
golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./pkg/bundler/...
# kind/KWOK validation of dcgm ConfigMap rendering:
make kwok-e2e RECIPE=eks-training
```

**Local render verification:** built the binary from this branch and rendered the eks/h100/training/ubuntu/kubeflow recipe:
- `bundles/009-gpu-operator/values.yaml` carries `dcgmExporter.config.{create:true, name:dcgm-exporter, data:|...}` with the full metrics CSV.
- No `009-gpu-operator-post/` directory is emitted (default deployer, helmfile, and `--set gpuoperator:dcgmExporter.config.create=false` override all confirmed).
- `install.sh` is a single `helm upgrade --install` (no second `kubectl apply` pass).

**Live-cluster migration test:** generated an `eks/gb200/inference/dynamo` recipe + bundle from this branch and exercised the upgrade against an existing GB200 EKS cluster previously deployed with the old (`gpu-operator-post`) layout.

1. Built the binary and ran `aicr recipe --service eks --accelerator gb200 --intent inference --platform dynamo` → `aicr bundle`. Bundle structure confirmed:
   - `014-gpu-operator/values.yaml` carries `dcgmExporter.config.{create: true, data: |...}` with the full metrics CSV.
   - Neither `013-gpu-operator-pre/templates/` nor any `*-post/` directory contains a `dcgm-exporter` ConfigMap manifest.
2. Ran `013-gpu-operator-pre/install.sh` → upgraded cleanly.
3. Ran `014-gpu-operator/install.sh` against the cluster *before* removing the stale post release. Got the exact error documented in the upgrade notes:

   ```text
   Error: UPGRADE FAILED: unable to continue with update: ConfigMap "dcgm-exporter"
   in namespace "gpu-operator" exists and cannot be imported into the current release:
   invalid ownership metadata; annotation validation error: key
   "meta.helm.sh/release-name" must equal "gpu-operator": current value is
   "gpu-operator-post"
   ```

4. Followed the raw-Helm path in [Component Catalog → Upgrade Notes](https://github.com/NVIDIA/aicr/blob/main/docs/user/component-catalog.md#gpu-operator-dcgm-exporter-configmap-moved-into-the-main-release): `helm uninstall gpu-operator-post --namespace gpu-operator`, then re-ran `014-gpu-operator/install.sh` → upgraded cleanly.
5. Verified end state with the doc's verification one-liner:

   ```bash
   kubectl get configmap dcgm-exporter -n gpu-operator \
     -o jsonpath='{.metadata.annotations.meta\.helm\.sh/release-name}'
   # → gpu-operator
   ```

   Final Helm releases in the namespace: `gpu-operator` and `gpu-operator-pre` only.

The committed raw-Helm migration steps work end-to-end against a live GB200 EKS cluster.

The chainsaw assertion at `tests/chainsaw/bundle-templates/gpu-operator/chainsaw-test.yaml` verifies the values-based rendering directly; `make qualify` will exercise it.

## Risk Assessment

**Rollout notes:**

- **Fresh installs:** no operator action required. The main `gpu-operator` Helm release now owns the `dcgm-exporter` ConfigMap directly.
- **Upgrade from a prior AICR-generated bundle:** the previous bundle shipped the ConfigMap as a separate Helm release named `gpu-operator-post`, so the in-cluster ConfigMap carries `meta.helm.sh/release-name: gpu-operator-post` annotations. On `helm upgrade gpu-operator`, Helm 3 refuses to claim the ConfigMap because the ownership annotation does not match the upgrading release (this is **not** auto-adoption — an earlier draft of this PR description incorrectly claimed it would be; thanks @mchmarny for catching it in [#938 (comment)](https://github.com/NVIDIA/aicr/pull/938#discussion_r3251460539)). Operators must remove the stale `gpu-operator-post` release before applying the new bundle. The bundler no longer emits a `gpu-operator-post` release entry, so helmfile / Argo CD / Flux will not self-prune it. Step-by-step instructions for raw Helm, Helmfile, Argo CD, and Flux are documented in [Component Catalog → Upgrade Notes → gpu-operator: dcgm-exporter ConfigMap moved into the main release](https://github.com/NVIDIA/aicr/blob/main/docs/user/component-catalog.md#gpu-operator-dcgm-exporter-configmap-moved-into-the-main-release).
- **kind overlay:** the previous workaround (`create: false`, name: `""`) is preserved and now additionally sets `data: ""` so the upstream gate cannot accidentally render the ConfigMap on a future chart revision that loosens the create check.
- **Conformance validator behavior is unchanged** — only the documentation around multi-SG prereqs is new.

## Checklist

- [ ] Tests pass locally (`make test` with `-race`) — deferred to CI
- [ ] Linter passes (`make lint`) — deferred to CI
- [x] I added/updated tests for new functionality (chainsaw assertion rewritten)


recipes/components/gpu-operator/manifests/dcgm-exporter.yaml
mchmarny, line 1: **High — upgrade rollout will hit a Helm ownership conflict, not adopt.**

The PR description claims: *"on `helm upgrade`, the chart will start creating the ConfigMap from values. If a cluster had the AICR-shipped post-manifest ConfigMap of the same name (`dcgm-exporter`) already applied, Helm will adopt it on next upgrade because the metadata matches; no manual cleanup expected."*

That's not how Helm 3 adoption works. This post-manifest is currently rendered into a **separate Helm release** named `gpu-operator-post` (see `pkg/bundler/deployer/helmfile/testdata/mixed_gpu_operator/helmfile.yaml:16` and `flux_test.go` references to `gpu-operator-post/`). So the existing in-cluster ConfigMap carries:

```
meta.helm.sh/release-name: gpu-operator-post
meta.helm.sh/release-namespace: gpu-operator
```

After this change, the main `gpu-operator` release will try to create a ConfigMap of the same name. Helm will refuse with:

```
Error: rendered manifests contain a resource that already exists. Unable to continue with update:
ConfigMap "dcgm-exporter" in namespace "gpu-operator" exists and cannot be imported into the
current release: invalid ownership metadata; annotation validation error: key
"meta.helm.sh/release-name" must equal "gpu-operator": current value is "gpu-operator-post"
```

What actually unblocks the upgrade depends on the deployer:
- **helmfile** with `installed: true` toggled off on the removed release: prunes `gpu-operator-post` first → ConfigMap goes away → main release creates it. Clean *if* the bundle still emits `gpu-operator-post` as an absent release rather than just omitting it entirely (helmfile only prunes what it knows about).
- **Argo CD / Flux**: requires `prune: true` AND the old `gpu-operator-post` HelmRelease/Application to be removed by this PR's bundle output, not just orphaned. Otherwise the old release lingers and the new ConfigMap creation fails.
- **Raw `helm install`** (per the README in each component bundle): operators must run `helm uninstall gpu-operator-post -n gpu-operator` (or `kubectl delete configmap dcgm-exporter -n gpu-operator`) before `helm upgrade`. The current `Rollout notes` say no cleanup is expected, which will surprise them.

Two fixes worth considering:
1. Verify what the bundler emits for the `gpu-operator-post` release after this change — if it's gone entirely from the helmfile/argocd/flux output, only prune-capable deployers self-heal.
2. Update the PR description's rollout notes to spell out the manual step for raw-helm operators (`helm uninstall gpu-operator-post` before upgrade), and ideally add a short migration note to `docs/user/validation.md` or wherever upgrades are documented.

tests/chainsaw/bundle-templates/gpu-operator/chainsaw-test.yaml
```diff
@@ -56,20 +56,26 @@ spec:
...
-                --file ./assert-dcgm-defaults.yaml \
-                --no-color --timeout 5s
+              # The upstream gpu-operator chart renders the dcgm-exporter
+              # ConfigMap directly from values.yaml under
+              # dcgmExporter.config.data; AICR no longer ships a separate
+              # post-manifest for it. Verify the values block carries the
+              # config (name + non-empty data) the chart needs to render
+              # the ConfigMap at install time.
+              VALUES=$(ls "${WORK}"/bundle-defaults/[0-9][0-9][0-9]-gpu-operator/values.yaml 2>/dev/null | head -1)
+              [ -n "${VALUES}" ] || { echo "values.yaml not found in any *-gpu-operator/" >&2; exit 1; }
```
mchmarny, line 71: **Medium — substring greps replace a structured chainsaw assertion; consider keyed lookups.**

The old assertion ran `chainsaw assert` against the rendered ConfigMap, which is a structured check (kind+name match). The new test substring-greps in the values blob:

```
grep -q 'name: dcgm-exporter' "${VALUES}"
grep -q 'DCGM_FI_DEV_GPU_TEMP' "${VALUES}"
```

Two failure modes this won't catch:
1. `name: dcgm-exporter` matches anything containing that string — including a future `serviceMonitor.name`, a label value, an env var, etc. There's nothing pinning it to `dcgmExporter.config.name`.
2. The path assertion is gone — if a future refactor accidentally puts `data: |` under the wrong key (e.g. under `dcgmExporter.config.serviceMonitor` or top-level), the metrics list and name still appear in the file and the test passes, but the chart will not render the ConfigMap.

A `yq` keyed lookup tightens both:
```bash
yq -e '.dcgmExporter.config.name == "dcgm-exporter"' "${VALUES}" >/dev/null
yq -e '.dcgmExporter.config.create == true' "${VALUES}" >/dev/null
yq -e '.dcgmExporter.config.data | test("DCGM_FI_DEV_GPU_TEMP")' "${VALUES}" >/dev/null
```

Not a merge-blocker if `yq` isn't already available in the chainsaw runner, but worth doing — chainsaw already brings `jq`/`yq` in most rigs.

tests/chainsaw/bundle-templates/gpu-operator/chainsaw-test.yaml
```diff
@@ -87,16 +93,19 @@ spec:
...
               WORK="/tmp/chainsaw-bundle-gpu-operator-templates"
-              ## When create=false, the template gate prevents any YAML objects
-              ## from being rendered. The bundler removes empty manifests.
-              MATCHES=$(ls "${WORK}"/bundle-dcgm-disabled/[0-9][0-9][0-9]-gpu-operator-post/templates/dcgm-exporter.yaml 2>/dev/null | wc -l)
-              if [ "${MATCHES}" -ne 0 ]; then
-                echo "FAIL: dcgm-exporter.yaml should not exist when create=false"
+              # The upstream chart gates the ConfigMap on `config.create`
+              # AND non-empty `config.data` — when the override sets
+              # create=false, the chart skips the ConfigMap at install time.
+              VALUES=$(ls "${WORK}"/bundle-dcgm-disabled/[0-9][0-9][0-9]-gpu-operator/values.yaml 2>/dev/null | head -1)
```
mchmarny, line 110: nit: `grep -Eq '^\s*create:\s*false\s*$'` will match any `create: false` line at any nesting. Today there's only one `create:` in the gpu-operator values, but anchor it to the right key if you adopt the `yq` approach in the sibling comment: `yq -e '.dcgmExporter.config.create == false' "${VALUES}"`.

recipes/overlays/kind.yaml
```diff
@@ -54,14 +54,14 @@ spec:
...
-        # ClusterPolicy stays InProgress waiting for exporter pods that need
-        # the ConfigMap. Use the exporter's built-in default metrics instead.
+        # Use the exporter's built-in default metrics in kind — the custom
+        # metrics list targets real datacenter GPUs (NVLink/profiling/etc.),
+        # most of which are unavailable on the simulated devices kind uses.
         dcgmExporter:
           config:
             create: false
             name: ""
+            data: ""
```
mchmarny, line 64: Belt-and-suspenders here is correct — the upstream chart's gate is `config.create == true AND config.data non-empty`, so either `create: false` or `data: ""` alone would suppress the ConfigMap. Keeping both is fine and survives an upstream gate loosening. No change needed.

docs/integrator/eks-dynamo-networking.md
```diff
@@ -8,11 +8,42 @@ If system components and GPU workloads are on different node groups/security gro
...
+first scheduling decision is a tie-break, and subsequent runs are dominated by
+image-locality scoring (whichever node pulled the validator image first keeps
+winning). A re-run on a "freshly working" cluster is therefore not a reliable
+signal that the SG topology is correct.
+
+This is tracked in [issue #933](https://github.com/NVIDIA/aicr/issues/933);
+this page documents the cluster-side prerequisite until the validator gains
+`podAffinity` toward Prometheus.
+
 ## Required Security Group Rules
```
mchmarny, line 35: nit: the doc lands well, but the bullet list says *"required from every worker SG to the Prometheus-hosting SG"* — and a few lines up the prose says *"the system SG must accept ingress from the customer SG (and any other worker SG), not only from itself."* Worth one more clarifying line: if customers operate >2 worker node groups (e.g. a separate inference NG), they need N×rules, not just one customer→system. The "If the cluster has more than two worker security groups..." paragraph at the bottom already covers this — consider moving that paragraph up next to the bullet so readers see it before the code block.

mchmarny (commented):
Refactor direction is right — moving the dcgm-metrics CSV into `dcgmExporter.config.data` lets Helm own install/upgrade ordering and eliminates the kind deadlock root cause. The networking doc additions are useful context for the multi-SG case.

One substantive concern: the rollout notes assume Helm will adopt the existing in-cluster ConfigMap on upgrade, but it's owned by a different release (`gpu-operator-post`), so Helm 3 will refuse the import with an ownership-annotation error. Self-heals for prune-capable deployers (helmfile/argo/flux with prune), but raw-helm operators need a manual `helm uninstall gpu-operator-post` (or ConfigMap delete) step that should be documented. Details in the inline comment on the deleted `manifests/dcgm-exporter.yaml`.

Other findings are smaller: the chainsaw asserts now rely on loose substring greps instead of structured assertions — would tighten to `yq` keyed lookups so future values restructuring can't silently pass. CI checks at time of review are green on the completed items (lint, CLI E2E, security scan), with Tier 2 e2e and main Test job still in progress — none failing.

Not blocking merge; please address the rollout-note accuracy before merging or confirm the bundler removes the old `gpu-operator-post` release entry across all deployers.

yuanchen8911 (changes requested):
A few issues remained to be addressed. PTAL.

docs/contributor/component.md
```diff
@@ -124,11 +124,19 @@ componentRefs:
     type: Helm
     version: v25.3.3
     manifestFiles:
-      - components/gpu-operator/manifests/dcgm-exporter.yaml
+      - components/gpu-operator/manifests/kernel-module-params.yaml
```
yuanchen8911: This snippet puts `kernel-module-params.yaml` under `manifestFiles`, but in the real GB200 overlays (`recipes/overlays/gb200-eks-training.yaml:42-52`) it is declared under `preManifestFiles` because the gpu-operator chart's ClusterPolicy reconcile reads this ConfigMap during `helm install --wait`. Following this example would emit it as a `*-post` chart and reproduce the install deadlock the pre-manifest path avoids (#859).

Suggest either changing the snippet to `preManifestFiles`, or picking a real post-safe example file (e.g. one of the skyhook-customizations / nodewright-customizations post-manifests).

pkg/recipe/metadata.go
```diff
@@ -100,7 +100,7 @@ type ComponentRef struct {
 
 	// ManifestFiles lists manifest files to include in the component bundle.
 	// Paths are relative to the data directory.
-	// Example: ["components/gpu-operator/manifests/dcgm-exporter.yaml"]
+	// Example: ["components/gpu-operator/manifests/kernel-module-params.yaml"]
```
yuanchen8911: Same misleading example as the contributor docs: `kernel-module-params.yaml` is declared in `preManifestFiles` in the real GB200 overlays (`recipes/overlays/gb200-eks-training.yaml:42-52`), not `manifestFiles`. Integrators copying this godoc as a template for a new component would author a deadlocked recipe. Suggest swapping to a real post-safe example.

docs/user/cli-reference.md
```diff
@@ -1620,7 +1620,7 @@ bundles/
     ├── values.yaml
     ├── cluster-values.yaml
     └── templates/
-        └── dcgm-exporter.yaml
+        └── kernel-module-params.yaml
```
yuanchen8911: Bundle layout shows `kernel-module-params.yaml` under the post-phase chart templates. Production output for this file lands in `*-pre/` because the real recipe declares it in `preManifestFiles` (see `recipes/overlays/gb200-eks-training.yaml:42-52`). Same misleading-phase pattern as the contributor docs and `pkg/recipe/metadata.go` godoc — recommend fixing all sites in one pass.

docs/user/api-reference.md
```diff
@@ -438,8 +438,8 @@ bundles.zip
 ├── gpu-operator/
 │   ├── values.yaml              # Helm chart values
 │   ├── manifests/
-│   │   ├── clusterpolicy.yaml   # ClusterPolicy CR
-│   │   └── dcgm-exporter.yaml   # DCGM Exporter config
+│   │   ├── clusterpolicy.yaml      # ClusterPolicy CR
+│   │   └── kernel-module-params.yaml  # GB200 kernel module params
```
yuanchen8911: The example places `kernel-module-params.yaml` under `gpu-operator/manifests/` (post-phase bucket). Production bundles emit this file under `gpu-operator-pre/manifests/` because the recipe declares it in `preManifestFiles` (see `recipes/overlays/gb200-eks-training.yaml:42-52`). Same misleading-phase pattern as the other layout examples in this PR.

docs/integrator/data-flow.md
```diff
@@ -677,7 +677,7 @@ bundle-output/
 │   ├── values.yaml        # Component Helm values
 │   ├── README.md          # Component install/upgrade/uninstall
 │   └── manifests/         # Optional manifest files
-│       └── dcgm-exporter.yaml
+│       └── kernel-module-params.yaml
```
yuanchen8911: The "Helm Deployer (default)" bundle output diagram places `kernel-module-params.yaml` under `gpu-operator/manifests/`. Production output would emit this file under `gpu-operator-pre/manifests/` because the recipe declares it in `preManifestFiles` (see `recipes/overlays/gb200-eks-training.yaml:42-52`). Same misleading-phase pattern as the layout examples in `docs/user/api-reference.md` and `docs/user/cli-reference.md`.

recipes/components/gpu-operator/manifests/dcgm-exporter.yaml
```diff
@@ -1,103 +0,0 @@
-# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
yuanchen8911, line 1: Deleting this embedded manifest path breaks `aicr bundle --recipe <file>` for any RecipeResult saved by a prior release whose merged `manifestFiles` still references it.

The bundler's manifest-collection loop at `pkg/bundler/bundler.go:1284-1295` calls `recipe.GetManifestContent(manifestPath)` for every entry in `manifestFiles` and returns `ErrCodeInternal` if the file is missing — no fallback, no warn-and-skip:

```go
content, err := recipe.GetManifestContent(manifestPath)
if err != nil {
    return nil, errors.PropagateOrWrap(err, errors.ErrCodeInternal,
        fmt.Sprintf("failed to load manifest %s for component %s",
            manifestPath, ref.Name))
}
```

**Repro:** generate `recipe.yaml` with a pre-PR binary → upgrade `aicr` to a post-PR binary → run `aicr bundle --recipe recipe.yaml`. The saved RecipeResult still carries `components/gpu-operator/manifests/dcgm-exporter.yaml` in its merged `manifestFiles` (the pre-PR base/gke-cos overlays declared it, and the resolved RecipeResult bakes in the merge).

Evidence the project treats RecipeResults as persisted artifacts: `examples/recipes/aks-training.yaml` is a check-in committed RecipeResult, and this PR had to manually remove the `manifestFiles` entry from it.

**Suggested fix:** filter out the legacy `components/gpu-operator/manifests/dcgm-exporter.yaml` path during manifest collection (skip it in the read loop in `bundler.go:1285`). Do **not** substitute an empty `[]byte` for the deleted path — empty content can still cause an unnecessary `gpu-operator-post` chart/release to be emitted depending on downstream deployer handling. Filtering is the clean fix.

The PR's existing upgrade notes in `docs/user/component-catalog.md` cover the live-cluster Helm release rename but not this CLI-level break.

yuanchen8911 (commented):
Cross-review by Claude Code + Codex + CodeRabbit.

**Summary**

Two cross-cutting issues to address before merge:

1. **Five coordinated docs examples** (flagged inline on each file) use `kernel-module-params.yaml` as a `manifestFiles` / post-phase example. In the real GB200 overlays (`recipes/overlays/gb200-eks-training.yaml:42-52`) this file is declared in `preManifestFiles` because gpu-operator's ClusterPolicy reads the ConfigMap during `helm install --wait` — applying it post deadlocks the install (#859). An integrator copying any of these examples would author a deadlocked recipe. Recommend fixing all five sites in one pass.

2. **Saved-RecipeResult backward-compat break** (flagged inline on the deleted manifest): `aicr bundle --recipe <file>` fails for any RecipeResult saved by a prior release whose merged `manifestFiles` still references the deleted path. See inline comment for repro and proposed filter-in-loader fix.

**Minor nit (file not in this PR, so noted here)**

`pkg/bundler/deployer/flux/doc.go:95-99` package docstring still uses `gpu-operator-post/templates/dcgm-exporter.yaml` as the post-phase example. The bundler's post-phase capability remains live for other components, so this is a stale example rather than a broken feature — but worth refreshing in the same pass.

**Dismissed**

A preliminary integration-analysis claim that `TestBundleGolden_MixedGPUOperator` would fail post-PR was independently checked by Claude Code and Codex and dismissed: synthetic-input unit test, PR does not touch `pkg/bundler/`, test passes locally (`go test -run TestBundleGolden_MixedGPUOperator ./pkg/bundler/deployer/helm/` → ok).

docs/contributor/component.md
njhensley: Good catch — the example was already inaccurate before this PR (`dcgm-exporter.yaml` was a post-manifest that's now correctly inlined into values; `kernel-module-params.yaml` is real but lives in `preManifestFiles`). The swap from one to the other was a conflation, not a fix.

Latest push points the example at `components/nodewright-customizations/manifests/tuning.yaml`, which is declared under `manifestFiles` in `recipes/overlays/h100-eks-training.yaml` and `gb200-eks-training.yaml` — so the example now maps to a real in-tree recipe in the correct phase.

The walkthrough body in this file was a small ConfigMap, but tuning.yaml is a 100-line Skyhook CR. Reproducing it in full would drown the section, so the example body is now an abbreviated Skyhook skeleton with the same `{{ index .Values "<component>" }}` templating pattern, plus a pointer to the in-tree complete file. Same fix applied across `pkg/recipe/metadata.go`, `pkg/recipe/adapter.go`, `pkg/recipe/adapter_test.go`, `docs/integrator/data-flow.md`, `docs/user/cli-reference.md`, `docs/user/api-reference.md`, and the stale `gpu-operator-post/dcgm-exporter.yaml` example in `pkg/bundler/deployer/flux/doc.go`.

pkg/recipe/metadata.go
njhensley: Fixed in latest push — swapped to `components/nodewright-customizations/manifests/tuning.yaml`, a real `manifestFiles` (post-phase) entry declared in `recipes/overlays/h100-eks-training.yaml` and `gb200-eks-training.yaml`. Same swap across all five sites. See the consolidated response on the `docs/contributor/component.md` thread for context.

docs/user/cli-reference.md
njhensley: Fixed in latest push — bundle-layout tree now shows `nodewright-customizations/.../tuning.yaml` under the post-phase chart, which is a real `manifestFiles` entry in `recipes/overlays/h100-eks-training.yaml` and `gb200-eks-training.yaml`. Same swap across all five sites. See the consolidated response on the `docs/contributor/component.md` thread for context.

docs/user/api-reference.md
njhensley: Fixed in latest push — bundle-structure example now shows `nodewright-customizations/manifests/tuning.yaml`, a real `manifestFiles` entry declared in `recipes/overlays/h100-eks-training.yaml` and `gb200-eks-training.yaml`. Same swap across all five sites. See the consolidated response on the `docs/contributor/component.md` thread for context.

docs/integrator/data-flow.md
njhensley: Fixed in latest push — Helm deployer layout now shows `nodewright-customizations/manifests/tuning.yaml`, a real `manifestFiles` entry declared in `recipes/overlays/h100-eks-training.yaml` and `gb200-eks-training.yaml`. Same swap across all five sites. See the consolidated response on the `docs/contributor/component.md` thread for context.

recipes/components/gpu-operator/manifests/dcgm-exporter.yaml
njhensley: The fail-closed behavior in the manifest-collection loop is intentional — we're not going to silently filter unknown paths in the loader. A RecipeResult is a point-in-time artifact produced by a specific AICR binary: the embedded data set (registry, overlays, manifest paths, chart pins) is part of that binary's surface, and the supported workflow is `aicr recipe` → `aicr bundle` against the same binary. Adding a "tolerate missing manifest files" path in the bundler would make removals silently lossy across versions and erode the determinism guarantee the validate/bundle pipeline depends on.

Re: `examples/recipes/aks-training.yaml` as evidence that RecipeResults are persisted artifacts the project ships — it's regenerated in-tree alongside changes to the embedded data, so it's a fixture, not a portability promise.

What's fair to call out is that this contract wasn't documented anywhere user-facing. Latest push adds:

- A paragraph at the top of `docs/user/component-catalog.md` Upgrade Notes stating recipes are tied to the generating binary and should be regenerated on AICR upgrade.
- A troubleshooting entry in `docs/user/cli-reference.md` under "Bundle Generation Fails" for the `failed to load manifest <path> for component <name>` error pattern, pointing users at recipe regeneration.

njhensley:
Pushed the doc/godoc updates addressing the cross-review findings:

- All five `kernel-module-params.yaml` `manifestFiles` examples swapped to `nodewright-customizations/tuning.yaml` — a real post-phase manifest declared in the h100/gb200 EKS training overlays. Also fixed the stale `gpu-operator-post/dcgm-exporter.yaml` example in `pkg/bundler/deployer/flux/doc.go`.
- On the saved-RecipeResult back-compat finding: leaving the bundler's fail-closed behavior in place by design and documenting the contract instead — recipes are tied to the binary that produced them. New paragraph in `docs/user/component-catalog.md` Upgrade Notes and a troubleshooting entry in `docs/user/cli-reference.md` cover the regenerate-on-upgrade workflow. Details in the inline replies.

yuanchen8911 (commented):
Follow-up to the prior cross-review. The five docs/godoc deadlock-misleading examples and the Flux stale post-manifest example are resolved — thanks. Two items remain: a structural contradiction in the new CLI layout example, and the contract/behavior gap on saved RecipeResults (the original F4 — Codex re-flagged on PR head `0d76b1b8`).

docs/user/cli-reference.md
```diff
@@ -1609,18 +1609,18 @@ bundles/
 │   ├── values.yaml
 │   ├── cluster-values.yaml        # Dynamic-path overrides (operator-edited)
 │   └── upstream.env               # CHART, REPO, VERSION (sourced by install.sh)
-├── 002-gpu-operator/              # Mixed component primary (upstream-helm)
+├── 002-nodewright-customizations/      # Mixed component primary (upstream-helm)
```
yuanchen8911: The registry marks nodewright-customizations as manifest-only (`defaultRepository: ""`), and the layout rules immediately below this example say manifest-only components emit a single local-helm folder — not the upstream-primary + `-post` pair shown here. The example contradicts the rules printed in the same file.

The previous `gpu-operator + dcgm-exporter` example was structurally correct (gpu-operator is a real mixed component); the deadlock concern was about the post-manifest file, not the mixed layout. Either pick a different real mixed component, or rewrite this example as the manifest-only single-folder layout.

docs/user/cli-reference.md
```diff
@@ -2343,6 +2343,8 @@ aicr --debug bundle -r recipe.yaml
 
 **"helm CLI not found on PATH" with `--vendor-charts`** — the bundle-time vendoring path shells out to `helm pull`. Install Helm v3 or later (`brew install helm` / package manager) and re-run, or drop `--vendor-charts` for a registry-referencing bundle. See [Vendoring Charts for Air-Gap](#vendoring-charts-for-air-gap).
 
+**"failed to load manifest \<path\> for component \<name\>"** — the recipe references a manifest path that does not exist in the current AICR binary's embedded data. This usually means the recipe was generated by an older binary and a referenced manifest has since been removed or relocated. Regenerate the recipe with the current binary (`aicr recipe ...`) and re-bundle. AICR recipes are a point-in-time artifact of the binary that produced them; bundling a stale recipe against a newer binary is not supported.
```
yuanchen8911, line 2346: The docs document the contract, but `pkg/bundler/bundler.go:1284-1294` still emits a cryptic `ErrCodeInternal: failed to load manifest X for component Y` — operators hit it, grep the docs by error string, and only then learn it was intentional version skew. Documented contract without code-level enforcement.

Close the gap with either:

1. **Enforce explicitly** — detect the not-in-embed-data condition during recipe load, return `ErrCodeInvalidRequest` with *"recipe references path X not present in this binary; regenerate with `aicr recipe ...`"*.

2. **Shim silently** — filter the legacy `components/gpu-operator/manifests/dcgm-exporter.yaml` path in the read loop. No-op for the one known removal.

The repo ships committed RecipeResults (`examples/recipes/aks-training.yaml`) and this PR had to manually edit one to drop the deleted `manifestFiles` entry — evidence the format is treated as a persisted, edit-able artifact, which weakens the "always regenerate" framing.

docs/user/cli-reference.md
njhensley: You're right — `nodewright-customizations` is registered as manifest-only (`defaultRepository: ""`), so the bundler emits it as a single local-helm folder, not the upstream-primary + `-post` pair the example claimed. Reverted the three bundle-layout tree diagrams (`docs/user/cli-reference.md`, `docs/user/api-reference.md`, `docs/integrator/data-flow.md`) to `network-operator` + `nfd-network-rule.yaml`. `network-operator` is a real mixed component (`defaultRepository: https://helm.ngc.nvidia.com/nvidia`, `defaultChart: nvidia/network-operator`) and `nfd-network-rule.yaml` is declared under `manifestFiles` in `recipes/overlays/aks.yaml` — same structural shape as the original `gpu-operator` + `dcgm-exporter.yaml` example, no deadlock concern. `nodewright-customizations/tuning.yaml` stays in the godoc and walkthrough examples where only the path matters, not the bundle shape.

docs/user/cli-reference.md
njhensley: Agreed — the documented contract didn't help operators staring at a cryptic `ErrCodeInternal`. Implemented option 1 (enforce explicitly):

- At the `recipe.GetManifestContent` call site in `pkg/bundler/bundler.go`, detect `stderrors.Is(err, fs.ErrNotExist)` and map to `ErrCodeInvalidRequest` with a remediation message naming the path, component, and pointing at `aicr recipe ...` for regeneration.
- The message branches on whether `--data` is in play (the same `fs.ErrNotExist` could be a misconfigured external overlay or a hand-edited recipe path, not just version skew). The bundler type-asserts the data provider against `*recipe.LayeredDataProvider` at the error site to make the distinction — matches the existing pattern in `copyDataFiles` rather than widening the `DataProvider` interface for a single caller.
- Other read errors stay as `ErrCodeInternal`.
- `TestCollectComponentManifests_MissingPath` covers both paths: asserts `ErrCodeInvalidRequest` via the code-keyed sentinel pattern plus message content (path, component, "regenerate", presence/absence of `--data` per branch).

Re: `examples/recipes/aks-training.yaml` as evidence RecipeResults are persisted artifacts — it's a fixture maintained in lockstep with the embedded data in the same PR that changes either side. The "manually edit one to drop the deleted entry" step is the same pattern any in-tree fixture follows when the schema or data shifts, not a portability promise.

docs/contributor/component.md
```diff
@@ -96,21 +96,29 @@ For components that need additional Kubernetes manifests (beyond the Helm chart)
 
 **Step 1: Create manifest file**
 
-Create `recipes/components/gpu-operator/manifests/dcgm-exporter.yaml`:
+Create the manifest under `recipes/components/<name>/manifests/`. The bundler
+renders each manifest as a Helm template, so files can use the same
+`{{ index .Values "<component>" }}` lookup the component's `values.yaml`
+exposes. Abbreviated skeleton (the in-tree
+`recipes/components/nodewright-customizations/manifests/tuning.yaml` is the
```
yuanchen8911: This section describes manifests "beyond the Helm chart," which is the mixed-component case: upstream chart plus injected -post wrapper. The example uses `nodewright-customizations`, but that component is manifest-only (`defaultRepository: ""`), so it produces a single local-helm folder rather than an upstream primary plus -post.

Please switch this example to `network-operator` / `components/network-operator/manifests/nfd-network-rule.yaml`. That is a real upstream-chart component with a post manifest, is already referenced from `recipes/overlays/aks.yaml`, and matches the bundle-layout examples updated in 7fe890b4.

Places to update in this section:

- prose around L103
- YAML example around L106-L122
- recipe snippet around L130-L134

pkg/recipe/metadata.go
```diff
@@ -100,7 +100,7 @@ type ComponentRef struct {
 
 	// ManifestFiles lists manifest files to include in the component bundle.
 	// Paths are relative to the data directory.
-	// Example: ["components/gpu-operator/manifests/dcgm-exporter.yaml"]
+	// Example: ["components/nodewright-customizations/manifests/tuning.yaml"]
```
yuanchen8911: For consistency with the contributor-doc example, please change this to:

```go
// Example: ["components/network-operator/manifests/nfd-network-rule.yaml"]
```

Same for `pkg/recipe/adapter.go:34`. This is lower priority than the docs issue, but keeping the canonical `manifestFiles` example aligned avoids future drift.

yuanchen8911 (commented):
Two follow-up notes after the latest push. The first is the substantive one (docs/godoc example mismatch flagged by codex). Also leaving a top-level comment about reconciling `examples/recipes/README.md` with the new fail-closed bundler enforcement.

yuanchen8911:
Follow-up doc note (file not in this PR diff, so leaving as a top-level comment) — `examples/recipes/README.md` near L12 ("Using Example Recipes"):

With the new fail-closed missing-manifest error in `pkg/bundler/bundler.go`, we should document the version boundary near the example-recipe usage section. Suggested wording:

Example recipes are persisted, editable artifacts, but they are not a cross-version compatibility boundary. After upgrading AICR, regenerate recipes with `aicr recipe ...` before bundling because embedded manifest paths, chart pins, and defaults can change between versions.

This reconciles the repo shipping committed `RecipeResult`s with the new explicit enforcement in the bundler — operators who hit the new `ErrCodeInvalidRequest` will have a doc to grep for.

docs/contributor/component.md
njhensley: You're right — the "beyond the Helm chart" framing in this section is the mixed-component case, and `nodewright-customizations` is manifest-only. Inconsistent with the layout examples in `docs/user/cli-reference.md` et al. that already use `network-operator`. Switched the walkthrough to `network-operator` + `nfd-network-rule.yaml`: updated prose, example body (NFD NodeFeatureRule skeleton matching the real file's structure — the real file has no Helm templating, so the templating capability moves to prose rather than the example body), and the recipe snippet. Also flipped `pkg/recipe/metadata.go:103`, `pkg/recipe/adapter.go:34`, and the `pkg/recipe/adapter_test.go:496` fixture path for canonical alignment.

pkg/recipe/metadata.go
njhensley: Done — both godoc examples (`pkg/recipe/metadata.go:103`, `pkg/recipe/adapter.go:34`) swapped to `components/network-operator/manifests/nfd-network-rule.yaml`. The test fixture in `pkg/recipe/adapter_test.go:496` swapped too so the resolved path stays a real in-tree file.

njhensley:
Good call on `examples/recipes/README.md` — added the version-boundary paragraph near "Using Example Recipes" with your suggested wording verbatim. Pairs with the troubleshooting entry in `docs/user/cli-reference.md` and the Upgrade Notes paragraph in `docs/user/component-catalog.md` so the contract is documented in all three spots an operator might land.

yuanchen8911 (approved):
All previously flagged issues addressed across the latest commits:

- Helm ownership-conflict claim corrected; raw Helm / Helmfile / Argo CD / Flux migration steps documented.
- Chainsaw assertions anchored to `yq` keyed paths.
- Bundler fail-closed enforcement returns `ErrCodeInvalidRequest` with operator-actionable guidance for missing embedded manifest paths; covered for embedded-only and `--data` provider modes.
- All five `manifestFiles` example sites (contributor doc, godoc in `pkg/recipe/{metadata.go,adapter.go,adapter_test.go}`, and the three bundle-layout trees) consistent on `network-operator` + `nfd-network-rule.yaml` — a real mixed-component example.
- Cross-version compatibility note added to `examples/recipes/README.md`, reconciling committed `RecipeResult`s with the new bundler enforcement.

CI green. Non-blocking nits (PR-description Testing-section staleness, deployer test-fixture filenames still using `dcgm-exporter.yaml`) can land as follow-up cleanup.

PR #948, opened 2026-05-18
feat(slinky-slurm): add cluster chart and EKS/Kind leaves

faganihajizada:
## Summary

Wires the Slinky `slurm` cluster chart into AICR as an opt-in component (`platform-slurm-cluster` mixin) and ships the first two leaf recipes that consume both the Slinky operator and cluster:

- `h100-eks-ubuntu-training-slurm`
- `h100-kind-training-slurm`

## Changes

**Component (`recipes/components/slinky-slurm/values.yaml`):**
- Overrides the chart's native `slinky` keys for `nodesets` / `loginsets` so user values merge with chart defaults (avoids the `nodeset.logfile` typed-merge failure observed with custom sub-keys).
- Production-leaning defaults: priority/multifactor scheduler tuning via `extraConfMap` (strings explicitly quoted to dodge YAML 1.1 coercion), cgroup + gres config files, ClusterIP `restapi`.
- Accounting (slurmdbd) and DCGM job-mapping are **disabled by default** — accounting needs an external MariaDB AICR does not bundle; DCGM needs `dcgm-exporter` on workers. Both opt in via valuesFile / `--set`. `accounting.storageConfig` is kept as an inert documented example (chart wraps the entire `Accounting` CR in `if accounting.enabled`).

**Mixin (`recipes/mixins/platform-slurm-cluster.yaml`):** opt-in `slinky-slurm` cluster with deps on `slinky-slurm-operator` + `slinky-slurm-operator-crds`.

**Health check (`recipes/checks/slinky-slurm/health-check.yaml`):** asserts Slinky CRs (`Controller`, `LoginSet`, `NodeSet`, `RestApi` at `slinky.slurm.net/v1beta1` under the `slinky-slurm` release name) AND underlying workload readiness (`Deployment` / `StatefulSet` `availableReplicas > 0`) before the generic pod-health step. Standardized on `availableReplicas` across all sub-checks.

**Registry:** adds `slinky-slurm` with `nodeScheduling`:
- system tier: `nodeSelectorPaths` + `tolerationPaths` for `controller.podSpec.*`, `restapi.podSpec.*`, `loginsets.slinky.podSpec.*`
- accelerated tier: `nodeSelectorPaths` + `tolerationPaths` for `nodesets.slinky.podSpec.*`

Both `--system-node-*` and `--accelerated-node-*` CLI flags now steer the correct pods.

**Leaves:**
- Both leaves intentionally have no `validation:` block — inherit the full set from their `*-training` parents (mirrors `*-kubeflow` leaves).

**KWOK (`kwok/scripts/validate-scheduling.sh`):** extracts the `platform` criterion so `*-slurm` overlays resolve to the slinky bundle they ship. Flag is intentionally scoped to `slurm` only — a full matrix run revealed the harness's sequential CRD cleanup is broken for NFD / KAI / run.ai (orphans block pre-flight on the second recipe onward), so widening to kubeflow/dynamo is deferred to a follow-up.

**Tests:**
- `pkg/recipe/deployment_order_guard_test.go`: 2 new cases asserting `cert-manager → slinky-slurm-operator-crds → slinky-slurm-operator → slinky-slurm`.
- `pkg/recipe/components_test.go`: new targeted test verifies registry node-scheduling paths line up with the actual `slinky` map-keys in `components/slinky-slurm/values.yaml` (prevents silent drift on key renames).

**Docs:** `docs/integrator/recipe-development.md` documents the two-mixin opt-in pattern; `docs/user/container-images.md` regenerated via `make bom-docs`.

## Verified against upstream chart v1.1.0

All component values cross-checked against [SlinkyProject/slurm-operator helm/slurm](https://github.com/SlinkyProject/slurm-operator/tree/main/helm/slurm) v1.1.0 and `main`. Notably:
- `accounting.storageConfig.passwordKeyRef` matches chart defaults (`name: mariadb-password`, `key: password`).
- All `extraConfMap` values quoted as strings — chart serializes the map as `map[string]string` and bare YAML 1.1 `no` would coerce to bool.

## Test plan

- [x] `make test` — full unit suite (includes deployment-order guard and components_test).
- [x] `make lint` — go/yaml/license/agents-sync/docs-sidebar.
- [x] `make qualify` — end-to-end pre-PR pass.
- [x] `aicr recipe` + `aicr bundle` smoke for both leaves.
- [x] `make check-health` against live Kind cluster for both `slinky-slurm-operator` and `slinky-slurm`.
- [x] `aicr validate` against deployed Kind cluster.
- [x] KWOK scheduling validation for `h100-kind-training-slurm` and `h100-eks-ubuntu-training-slurm`.
- [ ] CI: kwok-recipes, chainsaw, lint, build.

## Notes for reviewers

- The slurm-only KWOK guard is deliberate — widening to other platforms is blocked on harness CRD-cleanup fixes, not on the resolver behavior itself.
- The `slinky` map-key alignment between registry and values is enforced by the new components_test to catch future renames at unit-test time rather than via opaque scheduling failures.


recipes/overlays/h100-kind-training-slurm.yaml
```diff
@@ -0,0 +1,45 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
mchmarny, line 1: CI: the `Tier 2: h100-kind-training-slurm` and `Tier 2: h100-eks-ubuntu-training-slurm` KWOK lanes both got cancelled at ~15min (other lanes finish in 5–7min), which tripped `KWOK Test Summary`. That's a 2–3× wall-time bump on the slurm leaves vs every other lane, and it lands on the very recipes this PR is wiring up. Locally the description says these passed, so it could be an OCI pull stall (`oci://ghcr.io/slinkyproject/charts/slurm`) under the CI runner network, or just more CRs to evaluate. Worth pinning down whether the lane timeout needs raising, the chart needs to be cached/mirrored, or something in the resolver/bundle path got slower — and re-running until green before merge.

Run: https://github.com/NVIDIA/aicr/actions/runs/26031275885

recipes/checks/slinky-slurm-operator/health-check.yaml
```diff
@@ -55,17 +51,9 @@ spec:
...
-        # chainsaw `error` assertions pass when no matching resource exists,
-        # which would let this step trivially pass on an empty namespace.
-        # The two preceding deployment-availability steps prevent that:
-        # they require both deployments to have at least one ready replica
-        # in `slinky`, which guarantees pods exist and are inspectable
-        # before this step runs.
+        # Catch Pending / Failed / Unknown phases. The preceding
+        # deployment asserts fail loudly if the workloads are missing,
+        # so chainsaw `error` here is not asked to detect that case.
         - error:
```
mchmarny, line 58: nit: the previous comment block explained the *why* — that chainsaw `error` assertions vacuously pass on an empty namespace, so the two preceding deployment-availability steps are doing load-bearing work guaranteeing pods exist before this step runs. The new wording ("`error` here is not asked to detect that case") leaves a reader to infer that invariant. The old paragraph was longer but the information was non-obvious; consider keeping at least the "passes vacuously when no resource matches" sentence. Same trim applied verbatim to the new `slinky-slurm/health-check.yaml` step.

recipes/components/slinky-slurm/values.yaml
```diff
@@ -0,0 +1,177 @@
...
+        storage: 4Gi
+  # All values quoted as strings: extraConfMap is map[string]string and
+  # bare YAML 1.1 scalars (e.g. `no`) would coerce to boolean false.
+  # Accounting-only directives belong in the accounting opt-in valuesFile.
+  extraConfMap:
+    PriorityType: "priority/multifactor"
+    PriorityWeightAge: "2000"
+    PriorityWeightFairShare: "10000"
+    PriorityWeightJobSize: "1000"
+    PriorityWeightPartition: "1000"
```
mchmarny: `PriorityFlags: NO_NORMAL_ALL` + the raw weight values (2000–10000) is a deliberate non-normalized priority setup — it works, but it bakes a specific site-policy into the AICR default. Most upstream Slurm clusters either run normalized (default) or omit `PriorityFlags` and let admins tune. Two suggestions: (a) call this out in the file's docstring as a chosen policy rather than "production-leaning defaults", and (b) consider whether the AICR-shipped default should be the *simpler* normalized form, with this multifactor recipe documented as an opt-in valuesFile example for sites that want it. Not blocking, but worth a thought before this becomes the de-facto default for every leaf that opts into the cluster mixin.

recipes/components/slinky-slurm/values.yaml
```diff
@@ -0,0 +1,177 @@
...
+      enabled: true
+      configMap:
+        State: UP
+        Default: "YES"
+        MaxTime: UNLIMITED
+
+vendor:
+  nvidia:
+    dcgm:
+      enabled: false
```
mchmarny, line 179: nit: `MaxTime: UNLIMITED` on the default partition + `Default: "YES"` means any job submitted without `-t` runs forever. Fine for a single-tenant research cluster; unsafe for any shared environment. Same comment as above — flag this as a chosen policy rather than a generic default, and consider whether the shipped default should be a finite cap (e.g. `MaxTime: "1-00:00:00"` for 1 day) with `UNLIMITED` documented as the opt-in path. Won't block.

recipes/checks/slinky-slurm/health-check.yaml
```diff
@@ -0,0 +1,141 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Slinky Slurm cluster health check.
+#
+# Names assume the AICR localformat release name `slinky-slurm`; the
+# trailing `-slinky` comes from the chart's default map-key for
+# loginsets/nodesets. Other deployers using a different release name
+# need a parameterized variant of this check.
+apiVersion: chainsaw.kyverno.io/v1alpha1
```
mchmarny, line 24: The release-name caveat is documented here — good. One follow-up: this check is registered in `registry.yaml` and embedded in the binary, so any non-localformat deployer that lands later (flux/argocd already exist, and they may rewrite release names) will silently break this assertion because the names are hardcoded. Either drop a TODO referencing the deployer matrix or parameterize the names via a JMESPath/`($.metadata.name ~ ...)` match once that path is supported. Not blocking for this PR; just don't let "needs a parameterized variant" rot in this comment forever.

kwok/scripts/validate-scheduling.sh
```diff
@@ -311,21 +311,30 @@ generate_bundle() {
...
     [[ -n "$accelerator" ]] && recipe_args+=(--accelerator "$accelerator")
     [[ -n "$intent" ]] && recipe_args+=(--intent "$intent")
     [[ -n "$os" ]] && recipe_args+=(--os "$os")
+    # Only forward --platform for platforms validated under KWOK. Other
+    # platforms (kubeflow, dynamo, nim) historically resolve to their
+    # non-platform parent here; preserve that behavior to avoid regressing
+    # existing matrix lanes. Extend as additional platforms are validated.
+    if [[ "$platform" == "slurm" ]]; then
+        recipe_args+=(--platform "$platform")
+    fi
```
mchmarny, line 339: The `[[ "$platform" == "slurm" ]]` allowlist is fine for now, but it'll silently bit-rot: a future `*-kubeflow` or `*-dynamo` leaf with criteria.platform set will land here and continue to be resolved against its non-platform parent without any signal. Consider failing-closed for unknown platforms (or at minimum `log_info "platform=$platform not yet validated under KWOK — resolving without --platform"`) so the next person adding a leaf gets a hint instead of mysterious bundle diffs.

pkg/recipe/components_test.go
```diff
@@ -210,6 +211,78 @@ func TestComponentRegistry_NodeSchedulingPaths(t *testing.T) {
...
+	if _, ok := values.Nodesets["slinky"]; !ok {
+		t.Errorf("%s must define nodesets.slinky to match the registry's "+
+			"nodeScheduling paths (got nodesets keys: %v)", valuesPath, slices.Sorted(maps.Keys(values.Nodesets)))
+	}
+	if _, ok := values.Loginsets["slinky"]; !ok {
+		t.Errorf("%s must define loginsets.slinky to match the registry's "+
+			"nodeScheduling paths (got loginsets keys: %v)", valuesPath, slices.Sorted(maps.Keys(values.Loginsets)))
+	}
+}
+
```
mchmarny, line 285: Nice cross-check — the registry-vs-values-file alignment is exactly the kind of drift that would otherwise surface as opaque scheduling failures only at deploy time. Slight suggestion for future-proofing: this is a per-component test; if more components end up with `<key>` map-key conventions (the chart's `slinky` slot here, but `kueue.podTemplate` and similar elsewhere), it might be worth a single table-driven `TestComponentRegistry_NodeSchedulingPathsAlignWithValuesFiles` that loops over the registry and validates every leaf-path's first map segment resolves in the loaded values. Not for this PR — just a thought.

mchmarny (changes requested):
Solid wiring overall — the mixin split (`platform-slurm` always-on operator vs opt-in `platform-slurm-cluster`) is the right shape, and the new `TestComponentRegistry_SlinkySlurm_NodeSchedulingPaths` cross-check between registry paths and `values.yaml` map-keys is exactly the kind of guard this codebase needs more of. Deployment-order test additions also look correct.

One blocker and a few non-blocking nits:

- **CI (blocker):** both `Tier 2: h100-kind-training-slurm` and `Tier 2: h100-eks-ubuntu-training-slurm` KWOK lanes were *cancelled* at ~15min (every other lane finished in 5–7min), which is what failed `KWOK Test Summary`. Since those are the very recipes this PR adds, please confirm whether it's a runner timeout vs. OCI pull stall vs. resolver/bundle slowdown, and get the lanes green before merge.
- **Defaults posture (non-blocking):** `PriorityFlags: NO_NORMAL_ALL` + raw priority weights + `MaxTime: UNLIMITED` are real site-policy choices baked into the shipped default. Reasonable for a research cluster; worth at least re-labeling these as "chosen policy" rather than generic "production-leaning defaults", or moving the multifactor / unlimited bits behind an opt-in valuesFile.
- **Health-check release-name pinning (non-blocking):** the `slinky-slurm/health-check.yaml` resource names assume the localformat release name — fine for `make check-health`, but flux/argocd deployers will silently break this. The file flags it; just don't let "needs a parameterized variant" stay a comment forever.
- **KWOK platform allowlist (non-blocking):** the `platform == "slurm"` allowlist will bit-rot the next time someone adds a leaf with a non-allowlisted criteria.platform — consider a log line so the next author gets a hint.
- **Comment trim regression (nit):** the `slinky-slurm-operator/health-check.yaml` rewrite drops the non-obvious "chainsaw `error` passes vacuously on empty namespace" rationale.

Also, the branch is `behind` main — please rebase before re-pushing.

recipes/checks/slinky-slurm-operator/health-check.yaml
faganihajizada: Addressed. Restored the "passes vacuously when no resource matches" sentence in both so the load-bearing role of the preceding deployment-availability asserts is explicit, not inferred.

recipes/components/slinky-slurm/values.yaml
faganihajizada: Good point! I say we remove for now and document in comment

recipes/overlays/h100-kind-training-slurm.yaml
faganihajizada: Fixed. Final root cause was not OCI pull or lane timeout — it was a three-layer cascade:

1. `slinky-slurm-operator` submits `cert-manager.io/Certificate` at install time → cert-manager webhook unreachable on KWOK fakes → Helm install hangs.
2. Even after gating that off, the operator installs its own `ValidatingWebhookConfiguration` for Slurm CRs → cluster chart's CR submissions also unreachable.
3. The `Controller` CR maps to a StatefulSet with `persistence.enabled=true` (chart default) → Kind's `local-path` provisioner binds `WaitForFirstConsumer` to whichever node the pod schedules on → KWOK fakes can't actually back local storage → pod hangs `Pending` forever with `NominatedNodeName` set and no `FailedScheduling` event.

All three fixed via `--set` overrides scoped to the KWOK bundle (commit `6fe4ad0c`):
- `slinkyslurmoperator:webhook.enabled=false`
- `slinkyslurmoperator:certManager.enabled=false`
- `slurmcluster:controller.persistence.enabled=false`

Production recipes generated without these CI-only flags keep all three enabled. Both slurm lanes now green; full Tier 1/2 matrix passing.

recipes/checks/slinky-slurm-operator/health-check.yaml
faganihajizada: Restored. Both `recipes/checks/slinky-slurm-operator/health-check.yaml:54-57` and `recipes/checks/slinky-slurm/health-check.yaml:118-121` now spell out:

> Chainsaw `error` passes vacuously when no resource matches, so the preceding deployment-availability asserts are load-bearing: they guarantee the pods exist before this step runs.

recipes/components/slinky-slurm/values.yaml
faganihajizada: Agreed — dropped the entire multifactor block. `values.yaml` now omits `PriorityType`, `PriorityWeight*`, `PriorityMaxAge`, `PriorityDecayHalfLife`, and `PriorityFlags`, falling through to Slurm defaults (`priority/basic`, FIFO). Operators who want fairshare opt in per-deployment via `controller.extraConfMap`. Keeps the generic base simpler and avoids the surprise you flagged.

recipes/components/slinky-slurm/values.yaml
faganihajizada: `MaxTime` on the default partition is now `"24:00:00"` with a comment noting it's a single-tenant default and operators raise it per partition via the accounting opt-in or `controller.extraConfMap`. Avoids the "first user submits sleep infinity, queue dies" failure mode you flagged.

recipes/checks/slinky-slurm/health-check.yaml
faganihajizada: Added an inline `TODO` at `recipes/checks/slinky-slurm/health-check.yaml` flagging the hardcoded release-name coupling and the follow-up to use a Chainsaw label selector once we standardize one. Tracked separately so it doesn't block this PR.

kwok/scripts/validate-scheduling.sh
faganihajizada: Tightened in commit `88418d18`. The allowlist now emits a `log_info` hint when a non-slurm `platform` value is passed, so future contributors adding a new platform see exactly what to extend instead of getting silent drops. Kept the strict-fail path out for now since `make` callers may legitimately pass other values that this script just doesn't translate to `--platform`; a future PR adding kubeflow / dynamo support here can promote it to a hard fail in the same place.

pkg/recipe/components_test.go
faganihajizada: Thanks. yeah, the divergence between `registry.yaml` nodeScheduling paths and the actual chart map-keys is exactly the kind of drift that's hard to catch via Helm-render alone, so a registry-vs-values structural assert felt cheap to add. Happy to extract this into a generic table-driven assertion for any component declaring `nodeScheduling` paths in a follow-up PR if you think it's worth standardizing.

mchmarny (approved):

PR #956, opened 2026-05-18
feat(ci,bundler): validate deployer OCI bundles matrix in KWOK

### LINKED ISSUE #843: CI does not validate deployer-specific bundle output (Argo CD bundles can ship broken)
### Prerequisites

### Feature Summary

Extend CI to validate deployer-specific bundle output for every supported `--deployer` value, not just the default (`helm`). Today, `argocd` and `argocd-helm` bundles are generated but never schema-validated or applied, so semantically broken Argo CD `Application` manifests can land on `main` undetected.

### Problem / Use Case

`aicr bundle` supports three deployers (`pkg/bundler/config/config.go:26-39`):

- `helm` (default) — per-component Helm bundles + `deploy.sh`
- `argocd` — app-of-apps with one `Application` per component (`pkg/bundler/deployer/argocd/templates/application.yaml.tmpl`)
- `argocd-helm` — Helm chart whose templates are Argo CD `Application` manifests

Current CI coverage:

| Layer | What it validates | Deployer coverage |
|---|---|---|
| KWOK recipes (`.github/workflows/kwok-recipes.yaml`, `kwok/scripts/validate-scheduling.sh:354`) | Bundle deploys and pods reach `Running` | **`helm` only** — `aicr bundle` invoked with no `--deployer` flag |
| Bundler unit tests (`pkg/bundler/bundler_test.go:685` `TestMake_ArgoCD`) | File presence, sync-wave annotation strings | Structural only |
| Chainsaw CLI (`tests/chainsaw/cli/bundle-variants/chainsaw-test.yaml:167-193`) | `yamllint`; `helm template` for `argocd-helm` | Syntax + Helm render only |

Gaps:

- No schema validation of `Application` CRDs (`apiVersion`, required `spec.source`, `destination.server`, `project`, etc.).
- No dry-run apply (`kubectl apply --dry-run=server` or kubeconform with the Argo CD CRD bundle).
- No end-to-end \"install Argo CD, apply app-of-apps, observe Sync\" path.
- A template regression in `application.yaml.tmpl` would only be caught after a user tries to deploy a release.

Concrete risks that would slip past today's CI:

- Invalid `apiVersion` / `kind` after an Argo CD CRD bump.
- Missing required fields (e.g., multi-source `spec.sources[].repoURL`).
- Broken Helm value references when `argocd-helm` is rendered through Argo CD (vs. raw `helm template`).
- Wave/finalizer annotations that pass `yamllint` but cause sync ordering bugs.

### Proposed Solution

Add a CI gate that exercises every deployer. Three options, ordered cheapest → most thorough; recommend landing **Option A first** and treating B/C as follow-ups.

**Option A — Static schema validation (fast, no cluster)**

- Add a step that, for each deployer, runs `aicr bundle --deployer <d>` over the representative recipes already used by KWOK.
- Validate every emitted YAML with `kubeconform` against:
  - upstream Kubernetes schemas, and
  - Argo CD CRDs (`argoproj.io/v1alpha1` `Application`, `AppProject`) loaded from a pinned `argo-cd` release.
- Extend `tests/chainsaw/cli/bundle-variants/chainsaw-test.yaml` to parse `Application` YAML and assert required `spec.source{,s}`, `destination`, `project`, sync-policy fields — not just string presence.

Pros: <30 s, no cluster, catches the bulk of regressions, easy to add to `make qualify`.
Cons: does not exercise actual Argo CD reconciliation (sources, repo creds, kustomize/helm render).

**Option B — Render-through-Argo CD (medium)**

- In KWOK or a Kind cluster, install Argo CD (pin a chart version in `.settings.yaml`), apply the generated app-of-apps with `--dry-run=server`, and assert each child `Application` reaches `Synced=Unknown`/`Healthy=Unknown` without `ComparisonError` / `ConditionError`.
- Skip actual workload sync (no GPU/operator install) — purpose is template correctness, not deploy success.

Pros: catches reference errors, Helm value plumbing, multi-source paths.
Cons: adds Argo CD install time (~30-60 s) and a new CI dependency.

**Option C — Full deploy parity with KWOK (most thorough)**

- Matrix the existing KWOK workflow over `deployer ∈ {helm, argocd, argocd-helm}`.
- For Argo CD variants, install Argo CD into the KWOK cluster, apply the bundle, and reuse the existing pod-Running assertions.

Pros: end-to-end parity, would have caught every historical bundle bug.
Cons: ~3× KWOK runtime; Argo CD reconciling against repos that don't exist in CI needs a local repo-server or `targetRevision: HEAD` against the test workspace.

### Success Criteria

- `make qualify` (or a dedicated `make validate-bundles` target wired into CI) fails when any deployer emits a manifest that fails schema validation.
- Per-deployer coverage matrix is documented in `docs/contributor/` and `DEVELOPMENT.md`.
- A deliberate regression in `pkg/bundler/deployer/argocd/templates/application.yaml.tmpl` (e.g., dropping `spec.destination.server`) causes a red CI run.
- At least Option A is implemented; Option B/C tracked as follow-up issues with explicit accept/defer decision.

### Alternatives Considered

- **Status quo + pre-release manual smoke test** — relies on humans, doesn't scale, already failed to catch [hypothetical] regressions.
- **Generate then `argocd app validate` against a live Argo CD** — requires a hosted Argo CD; equivalent to Option B but with stronger dependencies.
- **Schema-validate only `argocd` and skip `argocd-helm`** — leaves the Helm-chart-of-Applications path uncovered; rejected.

### Component

### Priority

None for end users. CI runtime increases (Option A: negligible; B: ~1 min; C: ~3× current KWOK time). Adds `kubeconform` and the Argo CD CRD bundle as pinned CI tools — should be wired through `.settings.yaml` to honor the \"local development equals CI\" rule.

- Pin Argo CD CRD source by release tag in `.settings.yaml` to keep \"local = CI\".
- New failure surface: schema validator false positives when Argo CD bumps CRDs — mitigated by version-pinning and a Renovate rule.
- No production blast radius; CI-only change.

### Willing to contribute

Maybe, with guidance

mchmarny:
## Summary

Adds end-to-end validation of `argocd`, `argocd-helm`, and `flux` bundles
in the KWOK CI matrix. Each lane bundles the recipe, pushes the artifact
to an in-cluster OCI registry, lets the corresponding GitOps controller
pull and reconcile (or for `flux-oci`, apply), and asserts the bundle's
GitOps shape was produced. Closes the long-standing gap where Argo CD
and Flux bundle template regressions could ship to `main` undetected.

Three latent bundler bugs that this round-trip surfaced are fixed in
the same PR. Two further gaps are deferred to tracked follow-ups
(#963, #964).

Fixes: #843, #960, #961
Related: #962 (migrate KWOK argocd-* sync gate to Chainsaw),
         #963 (Gitea-in-cluster for filesystem-bundle round-trip),
         #964 (flux local-chart HelmReleases unreachable under OCI)

## Deployer × output-format CI coverage

Before / after this PR, by deployer:

| Deployer | FS bundle | OCI image |
|---|---|---|
| `helm` | ✅ Chainsaw (offline, cuj1-training) + KWOK live install | ➖ Intentionally unscoped — bundle pushes work generically but no downstream consumer of a `helm` deployer's OCI artifact exists |
| `argocd` | ✅ Chainsaw `bundle-variants` (app-of-apps shape) | ✅ **New** — KWOK lane (bundle → push → Argo CD pull → reconcile → pods Running) |
| `argocd-helm` | ✅ Chainsaw `bundle-dynamic` (Chart.yaml + templates/) | ✅ **New** — KWOK lane (Helm-OCI mediatype push + `helm install` round-trip) |
| `flux` | ✅ Chainsaw `bundle-flux` | ✅ **New** — KWOK lane (OCI pull → Kustomization apply → HelmRelease CRs created). Reconciliation deferred per #964. |
| `helmfile` | ✅ Chainsaw `bundle-helmfile` (incl. real `helmfile build`) | **N/A** — `helmfile.yaml`'s `oci: true` flags chart repos, not the bundle artifact |

The KWOK matrix goes from 1 deployer (helm) to 4 (helm, argocd-oci,
argocd-helm-oci, flux-oci).

For filesystem (Git-source) bundle round-trip coverage of `argocd` and
`flux`, see follow-up [#963](https://github.com/NVIDIA/aicr/issues/963).

## What "pass" means per lane

The KWOK lanes validate **bundle-format correctness**, not chart
reconciliation completeness. Each lane stops at the strongest signal
its controller can produce without fighting KWOK fidelity gaps.

| Lane | Pull artifact | Apply manifests | Reconcile to Ready |
|---|---|---|---|
| `helm` | n/a (filesystem) | ✅ (`helm install`) | ✅ pods scheduled |
| `argocd-oci` | ✅ (repo-server OCI pull) | ✅ (Argo CD sync) | ✅ terminal pass set (below) |
| `argocd-helm-oci` | ✅ (`helm pull` OCI) | ✅ (wrapper chart install) | ✅ terminal pass set (below) |
| `flux-oci` | ✅ (source-controller OCI pull) | ✅ (kustomize-controller apply) | ⏳ **deferred — see [#964](https://github.com/NVIDIA/aicr/issues/964)** |

The argocd-* lanes admit four reality-shaped terminal-pass outcomes to
handle KWOK pod-readiness simulation gaps and operator-mutation drift:

- `Synced + Healthy` — canonical pass
- `Synced + Progressing` — KWOK pod-readiness simulation gap (Prometheus StatefulSet, DaemonSet probes); not a bundle defect
- `OutOfSync + Healthy + operationState.phase=Succeeded` — operator-mutation drift (gpu-operator's ClusterPolicy, nvidia-dra-driver-gpu's DeviceClass); production users handle this with `ignoreDifferences`, not a bundle defect
- `Synced + Degraded + operationState.phase=Succeeded` — chart applied but health controller marks at least one managed resource Degraded (typically a Pod whose Ready signal KWOK can't produce, e.g. kai-scheduler leader election); same `phase=Succeeded` guard as arm 3 rejects genuinely-broken syncs

The matrix does **not** assert that every chart eventually reaches
Healthy under KWOK — that conflates bundle shape with chart-author
contracts and KWOK simulation fidelity.

`flux-oci` intentionally stops one step earlier ("bundle applied" =
OCIRepository pulled + Kustomization applied + HelmRelease CRs created)
because the bundler emits local-chart HelmReleases pointing at a
placeholder Git URL (`https://github.com/YOUR_ORG/YOUR_REPO.git`).
Under OCI consumption these can't reconcile, and Flux's `dependsOn`
cascade stalls every downstream HelmRelease. Flux's HelmRelease v2 API
blocks the obvious fix (`spec.chart.spec.sourceRef.kind` rejects
`OCIRepository`; `spec.chartRef` has no path support into a multi-chart
artifact), so the proper fix needs an ADR-level design decision.
Tracked in [#964](https://github.com/NVIDIA/aicr/issues/964).

## Issues this PR exposed

The round-trip lanes immediately surfaced latent bundler bugs. Each
was invisible to existing CI because no lane previously exercised a
GitOps controller pulling an AICR bundle from a registry.

| # | Symptom | Root cause |
|---|---|---|
| #960 | `argocd-oci` failed with `unsupported scheme "oci"` from Argo CD repo-server | `application.yaml.tmpl` rendered multi-source `$values` ref against an OCI URL; `$values` is Git-only in Argo CD |
| #961 | `argocd-helm-oci` failed with `unable to locate any tags in provided repository` from `helm pull` | Bundle pushed with generic AICR mediatype; helm v3 filters tags by `helm.config.v1+json` config mediatype + semver tag |
| (no issue, fixed in this PR) | `argocd-helm-oci` failed `component "gpu-operator-pre" not found in registry` on recipes with `preManifestFiles` (e.g. `gke-cos-training`) | `argocdhelm.processFolders` parent-resolution stripped `-post` suffix but not `-pre` |
| [#964](https://github.com/NVIDIA/aicr/issues/964) (deferred) | `flux-oci` local-chart HelmReleases stall under OCI; `dependsOn` cascade stalls the rest of the graph | Bundler emits placeholder `GitRepository` URL for inner local charts; Flux's HelmRelease API rejects `OCIRepository` as a chart sourceRef and `chartRef` has no path support, so a wrapper OCIRepository can't address sub-paths. Proper fix needs ADR-level choice (render at bundle time vs. publish per-chart). |

## What's in this PR

- **#960** — `argocd.Generator` gains an `InlineUpstreamValues` flag that, when set (auto-enabled by the bundler for OCI RepoURLs), inlines per-component values as `helm.valuesObject` under a single-source Application. The Git-only `$values` multi-source ref is no longer rendered for OCI targets.

- **#961** — New `oci.PackageAndPushHelmChart` flow: builds a chart-root-prefixed `chart.tgz`, pushes with `application/vnd.cncf.helm.config.v1+json` config + `application/vnd.cncf.helm.chart.content.v1.tar+gzip` layer, and rewrites `Chart.yaml.version` to the OCI tag so `helm install --version <tag>` resolves. The CLI rejects non-semver tags at bundle time with an actionable remediation hint.

- **argocdhelm `-pre` resolution** — Parent-component lookup in `processFolders` now symmetrically strips both `-pre` and `-post` suffixes. Recipes with `preManifestFiles` (gke-cos OS overlay, etc.) bundle correctly.

- **CLI contract clarification** — `--deployer argocd-helm` no longer auto-fills `repoURL` from the OCI reference; that bundle is URL-portable by design and `repoURL` is supplied at `helm install` time.

- **KWOK matrix scaffolding** — `deployer` dimension added to `kwok-recipes.yaml` (Tier 1 + Tier 3). Per-cell: Kind cluster + in-cluster `registry:2` + the GitOps controller(s) the selected deployer needs (Argo CD chart `9.5.14` or Flux 2 install manifest `v2.8.7`), repo-creds template Secret (Argo CD lanes only), `aicr bundle --output oci://...`, then either `kubectl apply` (argocd-oci, flux-oci) or `helm install` (argocd-helm-oci), then sync-gate, then `verify_pods`.

- **`flux-oci` lane** — Installs Flux 2 (source-controller / kustomize-controller / helm-controller) via the pinned `install.yaml`. Applies an `OCIRepository` (with `spec.layerSelector.mediaType: application/vnd.oci.image.layer.v1.tar+gzip` for the AICR generic OCI mediatype, plus `spec.insecure: true` for the in-cluster HTTP registry) and a `Kustomization` pointing at it. Sync gate validates pull + apply + HelmRelease-CRs-created; does NOT wait for HelmRelease reconciliation — see [#964](https://github.com/NVIDIA/aicr/issues/964).

- **Sync gate (argocd)** — Four-state terminal-pass predicate, non-root Application fallback for argocd-helm-oci wrappers whose `status.resources` hasn't populated yet, and `set -e`-safe `kubectl` probe.

- **3-strike rule** — `run-all-recipes.sh` bails the whole job with exit 50 after three consecutive GitOps sync timeouts (argocd-* or flux-oci) so a wedged cluster doesn't burn the full matrix budget.

- **Cleanup hardening** — Three call sites switched from `kubectl delete ns --wait=true --timeout=120s` to async delete + force-finalize. KWOK can't run real controller finalizers, so `--wait=true` blocked the full 2 min per namespace; with ~10 namespaces/recipe this exhausted the 15 min job budget on cleanup *after* the actual validation passed. Force-finalize bypasses the protocol by PATCHing empty finalizers — safe in the ephemeral KWOK cluster. Cleanup drops from ~20 min/recipe to ~5–10 s.

- **Retry hardening** — `KWOK_COMMAND_RETRIES` default bumped from 3 to 5 (cumulative sleep 15s → 75s) so transient GitHub releases CDN 502s on `kwok-stage-fast` chart fetch don't cancel CI runs.

- **SIGPIPE safety** — `list_bundle_entries` now traps SIGPIPE in its subshell so a downstream `head -N` closing the pipe early can't propagate exit 141 to the whole script.

- **macOS port-5500 default** — `kwok/kind-config.yaml` exposes the in-cluster registry on host port 5500 (ControlCenter holds 5000); Linux CI runners are unaffected.

- **Test hermeticity** — `TestWrite_VendorCharts_KustomizeFallthrough` sets `GIT_TERMINAL_PROMPT=0` so kustomize's git fetcher fails fast instead of prompting `Username for 'https://github.com':` mid-`make qualify` on macOS.

## Testing

`make qualify` passes locally on the final commit (test + lint + e2e + scan).

KWOK acceptance against `eks-training` on macOS:

- `helm`: ✅ PASS — baseline byte-identical to `main`
- `argocd-oci`: ✅ PASS — 15/15 Applications terminal, 44/44 pods scheduled
- `argocd-helm-oci`: ✅ PASS — 15/15 Applications terminal, 44/44 pods scheduled
- `flux-oci`: ✅ PASS — OCIRepository pulled, Kustomization applied, 14 HelmReleases created (reconciliation deferred per #964)

CI matrix covers 16 recipes × 4 deployers; gke-cos-* and other recipes with COS-OS `preManifestFiles` specifically exercise the `-pre` fix.

## Risk

`helm` and non-OCI deployer paths are byte-identical to `main`. Changes
that affect OCI users:

- `argocd` + OCI: per-component Applications now render single-source with `helm.valuesObject` inlined. Per-component `values.yaml` files remain on disk for operator readability but are no longer referenced via `$values`. Git-backed `argocd` is unchanged.
- `argocd-helm` + OCI: bundle is now published as a Helm OCI chart (`helm.config.v1+json` config + `helm.chart.content.v1.tar+gzip` layer). `Chart.yaml.version` is rewritten to match the OCI tag so `helm install --version <tag>` resolves correctly.
- `argocd-helm` + OCI now rejects non-semver tags at bundle time with an actionable error message ("Wrap as a semver pre-release, e.g. `0.0.0-<tag>`"), instead of producing an artifact that `helm pull` silently refuses to find.
- `flux` + OCI: **No bundler changes.** The bundler-side investigation found Flux's HelmRelease v2 API blocks the natural fix (`spec.chart.spec.sourceRef.kind` rejects `OCIRepository`; `spec.chartRef` has no path support). The flux-oci lane validates the bundle-apply step only and tracks the proper fix in [#964](https://github.com/NVIDIA/aicr/issues/964).

## Checklist

- [x] Linter passes (`golangci-lint` 0 issues)
- [x] No skipped/disabled tests
- [x] New tests added (argocdhelm `-pre` resolution; argocd inline values; oci helm mediatypes + tag semver; oci helm tgz reproducibility)
- [x] Docs updated (`docs/contributor/kwok-testing.md`, `docs/design/008-kwok-deployer-matrix.md`)
- [x] Commits cryptographically signed (`git commit -S`)

### DIFF
--- a/kwok/scripts/lib/cleanup.sh
+++ b/kwok/scripts/lib/cleanup.sh
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+# shellcheck shell=bash
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Shared cleanup helpers for the KWOK deployer-matrix CI lane.
+#
+# Sourced by validate-scheduling.sh and run-all-recipes.sh so the
+# system-namespace allowlist (the only thing standing between the
+# force-finalize sweep and the install-infra-owned ns argocd /
+# flux-system / aicr-registry) and the KWOK-context safety guard
+# live in exactly one place. Drift in either across scripts has
+# silently broken the test infra before — keep it centralized.
+#
+# Source guard: this file uses constants and functions only, no
+# side effects at source time.
+
+# SYSTEM_NS_PATTERN matches the namespaces that MUST NOT be touched by
+# the test cleanup paths:
+#   - default, kube-*, local-path-storage, kwok-system: cluster
+#     control-plane and CNI / storage plumbing. Wiping any of them
+#     bricks the cluster.
+#   - argocd, flux-system, aicr-registry: owned by install-infra.sh,
+#     reused across recipes. Wiping any of them forces a full re-
+#     install (~3-5 min) on every subsequent recipe.
+#
+# Used in `grep -vE "^(${SYSTEM_NS_PATTERN})$"` and
+# `grep -vE "^(${SYSTEM_NS_PATTERN})\s"` call sites. Anchors are NOT
+# baked in here so callers can choose space-anchor vs eol-anchor
+# depending on the kubectl output they're filtering.
+# shellcheck disable=SC2034  # consumed by sourcing scripts
+readonly SYSTEM_NS_PATTERN="default|kube-node-lease|kube-public|kube-system|kwok-system|local-path-storage|argocd|aicr-registry|flux-system"
+
+# ensure_kwok_context_loose checks that the kubectl context is a known
+# KWOK Kind cluster name. Cheap check, no node lookup. Use this from
+# cold-start paths where kwok nodes may not exist yet (e.g. run-all-
+# recipes.sh's initial cleanup before any recipe has applied nodes).
+#
+# Hard-fail on mismatch — the cleanup paths the caller is about to run
+# force-finalize namespaces, and we'd rather refuse to act than risk
+# wiping a real cluster.
+ensure_kwok_context_loose() {
+    local ctx
+    ctx=$(kubectl config current-context 2>/dev/null || true)
+    case "$ctx" in
+        kind-aicr-kwok|kind-aicr-kwok-test)
+            return 0
+            ;;
+        *)
+            echo "[ERROR] ensure_kwok_context: refusing to run — current kubectl context '${ctx:-<none>}' is not a known KWOK Kind cluster (expected kind-aicr-kwok or kind-aicr-kwok-test)" >&2
+            echo "[ERROR] If you intended to run against a different cluster, double-check KUBECONFIG and the cluster's purpose — this script force-finalizes namespaces" >&2
+            exit 1
+            ;;
+    esac
+}
+
+# ensure_kwok_context strict-checks BOTH the context name AND that at
+# least one node carries `type=kwok` label. Use this from in-recipe
+# paths where apply-nodes.sh has already populated the cluster. The
+# node-label check is load-bearing: a developer could `kind create
+# cluster --name aicr-kwok-test` and then point at a different cluster
+# whose context name happens to be the same — only the kwok-typed
+# nodes prove apply-nodes.sh ran against THIS cluster.
+ensure_kwok_context() {
+    ensure_kwok_context_loose
+
+    if ! kubectl get nodes -l type=kwok -o name 2>/dev/null | grep -q . ; then
+        echo "[ERROR] ensure_kwok_context: refusing to run — current context has no kwok-typed nodes" >&2
+        echo "[ERROR] Run kwok/scripts/apply-nodes.sh <recipe> first, or verify you're against the right cluster" >&2
+        exit 1
+    fi
+}
--- a/kwok/scripts/run-all-recipes.sh
+++ b/kwok/scripts/run-all-recipes.sh
@@ -15,15 +15,57 @@
 
 # Run all KWOK recipe tests sequentially in a shared cluster.
 # Usage:
-#   ./run-all-recipes.sh           # Run all testable recipes
-#   ./run-all-recipes.sh recipe1   # Run specific recipe(s)
+#   ./run-all-recipes.sh                           # Run all testable recipes (helm)
+#   ./run-all-recipes.sh recipe1                   # Run specific recipe(s)
+#   ./run-all-recipes.sh --deployer <name>         # Select deployer for all recipes
+#   ./run-all-recipes.sh --deployer=<name> recipe1 # `=` form supported
+#
+# Flags:
+#   --deployer <name>   Deployer to exercise for every recipe. One of:
+#                         helm             (default — original Helm path)
+#                         argocd-oci       (in-cluster registry + Argo CD
+#                                           app-of-apps)
+#                         argocd-helm-oci  (in-cluster registry + Argo CD
+#                                           via OCI Helm chart wrapper)
+#                         flux-oci         (in-cluster registry + Flux 2
+#                                           OCIRepository → Kustomization →
+#                                           HelmRelease)
+#                       When != helm, install-infra.sh is run ONCE before the
+#                       recipe loop with DEPLOYER exported so it installs the
+#                       shared in-cluster OCI registry plus the GitOps
+#                       controllers the selected deployer needs (Argo CD or
+#                       Flux). Their owning namespaces (`aicr-registry`,
+#                       `argocd`, `flux-system`) survive across recipe
+#                       iterations.
+#
+# Exit codes:
+#    0  all recipes passed
+#    1  one or more recipes failed (non-sync-timeout); install-infra.sh failed;
+#       or invalid arguments
+#   50  Argo CD sync deadline hit on 3 consecutive recipes (3-strike rule,
+#       ADR-008 §"Error Handling and Failure Modes"). Distinct so CI can
+#       distinguish infra/controller-wide failure from per-recipe issues.
+#       Mirrors validate-scheduling.sh's EXIT_ARGOCD_SYNC_TIMEOUT.
 
 set -euo pipefail
+
+# Mirrors validate-scheduling.sh's EXIT_ARGOCD_SYNC_TIMEOUT. Kept in sync by
+# the matching documentation in both script headers.
+readonly EXIT_ARGOCD_SYNC_TIMEOUT=50
+# Max consecutive sync timeouts before bailing the whole job (ADR-008
+# §"Error Handling and Failure Modes"). Only applied for argocd-* deployers.
+readonly ARGOCD_SYNC_STRIKE_LIMIT=3
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 KWOK_DIR="${SCRIPT_DIR}/.."
 REPO_ROOT="${KWOK_DIR}/.."
 OVERLAYS_DIR="${REPO_ROOT}/recipes/overlays"
+
+# Shared cleanup helpers (SYSTEM_NS_PATTERN, ensure_kwok_context). Same
+# source for validate-scheduling.sh so the system-ns allowlist regex
+# lives in one place.
+# shellcheck source=lib/cleanup.sh
+source "${SCRIPT_DIR}/lib/cleanup.sh"
 
 CLUSTER_NAME="${KWOK_CLUSTER:-aicr-kwok-test}"
 CONTEXT="kind-${CLUSTER_NAME}"
@@ -41,7 +83,11 @@ retry_command() {
     local description="$1"
     shift
 
-    local max_attempts="${KWOK_COMMAND_RETRIES:-3}"
+    # 5 attempts with doubling backoff: 5+10+20+40 = 75s cumulative sleep.
+    # The prior default (3 attempts → 15s cumulative) was not enough headroom
+    # for transient GitHub releases CDN 502s on kwok-stage-fast chart fetch,
+    # which observably persist 30-60s. Overridable via env for local tuning.
+    local max_attempts="${KWOK_COMMAND_RETRIES:-5}"
     local delay="${KWOK_COMMAND_RETRY_DELAY:-5}"
     local attempt=1
 
@@ -124,13 +170,59 @@ cleanup_between_tests() {
     # Clean up orphaned CRDs from cert-manager (cluster-scoped, not cleaned by ns delete)
     kubectl delete crd -l app.kubernetes.io/instance=aicr-test --ignore-not-found 2>/dev/null || true
 
-    # Wait for any still-terminating namespaces before next recipe
-    local system_ns="default|kube-node-lease|kube-public|kube-system|kwok-system|local-path-storage"
-    local terminating
-    terminating=$(kubectl get ns -o jsonpath='{range .items[?(@.status.phase=="Terminating")]}{.metadata.name}{"\n"}{end}' 2>/dev/null | grep -vE "^(${system_ns})$" || true)
-    for ns in $terminating; do
-        log_info "Waiting for namespace $ns to terminate..."
-        kubectl wait --for=delete "ns/$ns" --timeout=120s 2>/dev/null || true
+    # Force-finalize any still-terminating namespaces before the next recipe.
+    # `argocd`, `flux-system`, and `aicr-registry` are owned by install-
+    # infra.sh and MUST survive between recipes for the GitOps deployer
+    # lanes — installing them once and reusing avoids 5+ minutes of Helm
+    # install + CRD Established + controller-ready overhead per recipe.
+    # The three namespaces are listed unconditionally (even under --deployer
+    # helm) because they simply won't exist on the helm path, so the
+    # allowlist is harmless there. Do not gate by DEPLOYER.
+    #
+    # See the long comment in validate-scheduling.sh cleanup() for the
+    # rationale: KWOK cannot run real controller finalizers, so
+    # `kubectl wait --for=delete` against a Terminating namespace blocks
+    # the full --timeout=120s per namespace. Recipes touch ~10 namespaces;
+    # the previous version spent ~20 minutes of cleanup between recipes
+    # while doing zero useful work. Force-finalize bypasses the protocol
+    # by PATCHing spec.finalizers to []; safe in the ephemeral KWOK
+    # cluster (no real workloads to leak; cluster destroyed at job end).
+    local system_ns="${SYSTEM_NS_PATTERN}"  # see SYSTEM_NS_PATTERN in lib/cleanup.sh
+
+    # Two-phase cleanup, matching validate-scheduling.sh::cleanup_old_tests:
+    # the previous version only iterated namespaces ALREADY in
+    # status.phase=Terminating, which silently missed `Active` namespaces
+    # carrying a `deletionTimestamp` (Helm hook resources mid-uninstall,
+    # for example) and let them collide on the next recipe.
+    #
+    # Phase 1 — issue async deletes for every non-system namespace so the
+    # controller starts tearing down. Phase 2 — sleep briefly, then
+    # force-finalize anything still present. The two phases together
+    # match the in-recipe cleanup so the between-test path can't drift
+    # away from it over time.
+    local test_namespaces
+    test_namespaces=$(kubectl get ns -o jsonpath='{.items[*].metadata.name}' 2>/dev/null \
+        | tr ' ' '\n' \
+        | { grep -vE "^(${system_ns})$" || true; })
+    if [[ -z "$test_namespaces" ]]; then
+        return 0
+    fi
+
+    for ns in $test_namespaces; do
+        kubectl delete ns "$ns" --ignore-not-found --wait=false 2>/dev/null || true
+    done
+
+    # Brief grace for graceful deletion (real finalizers run in well
+    # under a second when there are no real workloads).
+    sleep 2
+    for ns in $test_namespaces; do
+        if kubectl get ns "$ns" >/dev/null 2>&1; then
+            log_info "Force-finalizing stuck namespace: $ns"
+            kubectl get ns "$ns" -o json 2>/dev/null \
+                | jq '.spec.finalizers = [] | .metadata.finalizers = []' \
+                | kubectl replace --raw "/api/v1/namespaces/${ns}/finalize" -f - >/dev/null 2>&1 \
+                || true
+        fi
     done
 }
 
@@ -138,7 +230,7 @@ run_recipe_test() {
     local recipe="$1"
     echo ""
     log_info "========================================"
-    log_info "Testing recipe: ${recipe}"
+    log_info "Testing recipe: ${recipe} (deployer=${DEPLOYER})"
     log_info "========================================"
 
     cleanup_between_tests
@@ -146,30 +238,171 @@ run_recipe_test() {
     # Create nodes (pass recipe name, script infers from overlay)
     bash "${SCRIPT_DIR}/apply-nodes.sh" "${recipe}" || return 1
 
-    # Run validation
-    bash "${SCRIPT_DIR}/validate-scheduling.sh" "${recipe}" || return 1
+    # Run validation. Preserve validate-scheduling.sh's exit code so callers
+    # can distinguish EXIT_ARGOCD_SYNC_TIMEOUT (50) from generic failures (1)
+    # for the 3-strike rule.
+    local rc=0
+    bash "${SCRIPT_DIR}/validate-scheduling.sh" --deployer "${DEPLOYER}" "${recipe}" || rc=$?
+    return "$rc"
 }
+
+# Print usage to stderr.
+usage() {
+    cat >&2 <<'EOF'
+Usage: run-all-recipes.sh [--deployer <name>] [recipe1 recipe2 ...]
+
+Flags:
+  --deployer <name>   Deployer to exercise for every recipe. One of:
+                        helm (default), argocd-oci, argocd-helm-oci, flux-oci
+
+Examples:
+  run-all-recipes.sh
+  run-all-recipes.sh eks-training
+  run-all-recipes.sh --deployer argocd-oci eks-training
+  run-all-recipes.sh --deployer=argocd-helm-oci
+  run-all-recipes.sh --deployer flux-oci eks-training
+EOF
+}
+
+# Deployer selection (issue #843). Set by --deployer in main(); default helm.
+# Under --deployer helm, this script's behavior is byte-identical to pre-#843
+# except the cleanup_between_tests system_ns allowlist gains two harmless
+# entries (argocd|aicr-registry) that never exist on the helm path.
+DEPLOYER="helm"
 
 main() {
     local recipes failed=() passed=()
+    local positional=()
 
-    if [[ $# -gt 0 ]]; then
-        recipes="$*"
+    # Parse flags + positional recipe names. Supports `--deployer X` and
+    # `--deployer=X`. Anything else with a leading `-` is rejected.
+    while [[ $# -gt 0 ]]; do
+        case "$1" in
+            --deployer)
+                if [[ $# -lt 2 ]]; then
+                    log_error "--deployer requires a value"
+                    usage
+                    return 1
+                fi
+                DEPLOYER="$2"
+                shift 2
+                ;;
+            --deployer=*)
+                DEPLOYER="${1#--deployer=}"
+                shift
+                ;;
+            -h|--help)
+                usage
+                return 0
+                ;;
+            -*)
+                log_error "Unknown option: $1"
+                usage
+                return 1
+                ;;
+            *)
+                positional+=("$1")
+                shift
+                ;;
+        esac
+    done
+
+    case "$DEPLOYER" in
+        helm|argocd-oci|argocd-helm-oci|flux-oci)
+            ;;
+        *)
+            log_error "Invalid --deployer value: '${DEPLOYER}'"
+            log_error "Must be one of: helm, argocd-oci, argocd-helm-oci, flux-oci"
+            usage
+            return 1
+            ;;
+    esac
+
+    if [[ ${#positional[@]} -gt 0 ]]; then
+        recipes="${positional[*]}"
     else
         recipes=$(get_recipes)
     fi
 
-    log_info "Found $(echo "${recipes}" | wc -w | tr -d ' ') recipe(s) to test"
+    log_info "Found $(echo "${recipes}" | wc -w | tr -d ' ') recipe(s) to test (deployer=${DEPLOYER})"
     ensure_cluster
+
+    # Safety: refuse to start the cleanup sweep unless the kubectl
+    # context points at a known KWOK Kind cluster. ensure_cluster has
+    # just created/reused the cluster, but it does not validate the
+    # context name — a developer running locally with a stale
+    # KUBECONFIG could otherwise direct the upcoming force-finalize
+    # sweep at a real cluster. Loose check only: kwok nodes don't
+    # exist yet on the initial run (apply-nodes.sh runs per recipe
+    # inside run_recipe_test).
+    ensure_kwok_context_loose
 
     # Clean up any stale resources from previous runs
     cleanup_between_tests
 
+    # Install shared in-cluster registry + the controller(s) the selected
+    # deployer needs (Argo CD for argocd-*, Flux for flux-oci). install-
+    # infra.sh is idempotent but unnecessary work per recipe; a failure
+    # here is fatal — the lane cannot run without it. DEPLOYER is exported
+    # so install-infra.sh can branch on it. Its exit code map: 10=yq/
+    # settings, 20=registry Deployment not Ready, 21=registry not reachable
+    # on host port, 30=Argo CD Helm install failed, 31=Application CRD not
+    # Established, 40=Repository secret apply failed, 60=Flux install
+    # manifest apply failed, 61=Flux controller not Ready, 62=Flux CRDs not
+    # Established. Surface the raw rc in the log line so an operator can
+    # map it.
+    if [[ "${DEPLOYER}" != "helm" ]]; then
+        log_info "Installing shared infra (in-cluster registry + controllers for ${DEPLOYER})..."
+        local infra_rc=0
+        DEPLOYER="${DEPLOYER}" bash "${SCRIPT_DIR}/install-infra.sh" || infra_rc=$?
+        if (( infra_rc != 0 )); then
+            log_error "install-infra.sh failed (exit code ${infra_rc}); cannot run ${DEPLOYER} deployer lane"
+            log_error "See kwok/scripts/install-infra.sh header for exit-code taxonomy"
+            return 1
+        fi
+    fi
+
+    # 3-strike rule for Argo CD sync timeouts (ADR-008 §"Error Handling and
+    # Failure Modes"). Tracks CONSECUTIVE EXIT_ARGOCD_SYNC_TIMEOUT failures.
+    # The counter resets on any rc != 50 (pass OR generic failure) so only
+    # genuinely consecutive sync timeouts trip the bail. A pass-fail-pass-
+    # timeout-timeout-timeout sequence is 3 consecutive; a timeout-pass-
+    # timeout-timeout sequence is 2 consecutive.
+    # Only argocd-* deployers can hit 50; helm path never trips this.
+    local consecutive_sync_timeouts=0
+
     for recipe in ${recipes}; do
-        if run_recipe_test "${recipe}"; then
+        local rc=0
+        run_recipe_test "${recipe}" || rc=$?
+        if (( rc == 0 )); then
             passed+=("${recipe}")
+            consecutive_sync_timeouts=0
         else
             failed+=("${recipe}")
+            # 3-strike rule is GitOps-only: helm path never returns 50, so
+            # the gate is currently implicit. Make it explicit by checking
+            # DEPLOYER too — keeps the contract auditable from this site
+            # alone instead of a chain of "but only X can produce 50"
+            # comments scattered across files.
+            if (( rc == EXIT_ARGOCD_SYNC_TIMEOUT )) \
+                    && [[ "$DEPLOYER" == argocd-* || "$DEPLOYER" == flux-oci ]]; then
+                consecutive_sync_timeouts=$(( consecutive_sync_timeouts + 1 ))
+                log_warn "GitOps sync timeout strike ${consecutive_sync_timeouts}/${ARGOCD_SYNC_STRIKE_LIMIT} on recipe ${recipe}"
+                if (( consecutive_sync_timeouts >= ARGOCD_SYNC_STRIKE_LIMIT )); then
+                    log_error "========================================"
+                    log_error "3-strike rule tripped: ${consecutive_sync_timeouts} consecutive GitOps sync timeouts"
+                    log_error "Bailing remainder of recipe loop (partial coverage > no coverage is false here)"
+                    log_error "Failed recipes so far:"
+                    for r in "${failed[@]:-}"; do [[ -n "$r" ]] && log_error "  - ${r}"; done
+                    log_error "========================================"
+                    cleanup_between_tests
+                    return "$EXIT_ARGOCD_SYNC_TIMEOUT"
+                fi
+            else
+                # Any other failure (including a hypothetical rc=50 from a
+                # path that shouldn't produce one) breaks the streak.
+                consecutive_sync_timeouts=0
+            fi
         fi
     done
 
--- a/kwok/scripts/validate-scheduling.sh
+++ b/kwok/scripts/validate-scheduling.sh
@@ -16,21 +16,89 @@
 # validate-scheduling.sh - Validate bundle scheduling on KWOK cluster
 #
 # Usage:
-#   ./validate-scheduling.sh <recipe-name>
+#   ./validate-scheduling.sh [--keep-namespace] [--deployer <name>] <recipe-name>
 #   ./validate-scheduling.sh h100-eks-ubuntu-training-kubeflow
+#   ./validate-scheduling.sh --deployer argocd-oci eks-training
+#
+# Flags:
+#   --keep-namespace    Preserve releases and namespaces after the run for
+#                       inspection (skips cleanup).
+#   --deployer <name>   Deployer to exercise. One of:
+#                         helm             (default — original Helm path)
+#                         argocd-oci       (aicr bundle --deployer argocd ->
+#                                           OCI push -> kubectl apply
+#                                           app-of-apps.yaml)
+#                         argocd-helm-oci  (aicr bundle --deployer argocd-helm
+#                                           -> OCI push -> helm install OCI
+#                                           chart in argocd namespace)
+#                         flux-oci         (aicr bundle --deployer flux ->
+#                                           OCI push -> apply OCIRepository +
+#                                           Kustomization wrappers and let
+#                                           Flux reconcile)
+#                       For argocd-* values the in-cluster registry + Argo CD
+#                       must already be installed; for flux-oci the registry
+#                       + Flux 2 controllers must be installed. Both are
+#                       managed by `kwok/scripts/install-infra.sh` keyed on
+#                       the DEPLOYER env var.
+#
+# Environment variables (GitOps deployers only):
+#   KWOK_REGISTRY_HOST_PORT  Host port the in-cluster registry is reachable on
+#                            from the runner. MUST match the value used when
+#                            running install-infra.sh / kind-config.yaml.
+#                            Default: 5500 (avoids the macOS ControlCenter
+#                            port-5000 collision; Linux CI runners have 5500
+#                            free as well).
+#   KWOK_ARGOCD_SYNC_TIMEOUT Seconds to wait for all Argo CD Applications to
+#                            reach Synced + Healthy (or Progressing) before
+#                            failing. Default: 300.
+#   KWOK_ARGOCD_ROOT_GRACE   Seconds to wait for the root Argo CD Application
+#                            (nvidia-stack for argocd-oci, aicr-stack for
+#                            argocd-helm-oci) to appear in the argocd
+#                            namespace before failing. A missing root after
+#                            this grace window indicates `kubectl apply` /
+#                            `helm install` silently produced no Application
+#                            resource. Default: 30.
+#   KWOK_FLUX_SYNC_TIMEOUT   Seconds to wait for the outer Kustomization +
+#                            all HelmReleases to reach a terminal state
+#                            before failing. Default: 300.
+#   KWOK_FLUX_ROOT_GRACE     Seconds to wait for the outer Kustomization to
+#                            appear in the cluster before failing. A missing
+#                            Kustomization after the grace window indicates
+#                            kubectl apply silently produced no resource
+#                            (RBAC denial, CRD missing). Default: 30.
 #
 # This script:
 # 1. Generates a recipe from the cluster config
 # 2. Generates a bundle from the recipe
-# 3. Deploys the bundle to the KWOK cluster
+# 3. Deploys the bundle to the KWOK cluster (Helm, Argo CD, or Flux per --deployer)
 # 4. Verifies all pods reach Running state (KWOK auto-transitions them)
 # 5. Reports success/failure
+#
+# Exit codes:
+#    0  success
+#    1  generic failure (bundle gen, deploy, pod scheduling, RBAC, CRD missing,
+#       apiserver unreachable, etc.)
+#   50  GitOps sync deadline hit (KWOK_ARGOCD_SYNC_TIMEOUT for argocd-*,
+#       KWOK_FLUX_SYNC_TIMEOUT for flux-oci). Distinct so run-all-recipes.sh
+#       can apply the 3-strike rule (ADR-008 §"Error Handling and Failure
+#       Modes") without grepping stderr. Only emitted for GitOps deployers;
+#       the helm path never returns 50.
 
 set -euo pipefail
+
+# Distinct exit code for Argo CD sync timeout. Documented in the header above
+# AND in kwok/scripts/run-all-recipes.sh so callers can branch on it.
+readonly EXIT_ARGOCD_SYNC_TIMEOUT=50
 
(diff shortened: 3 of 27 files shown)


mchmarny:
## Acceptance-test status (live KWOK run on macOS, 2026-05-18)

After pushing `4f9a7c05` (wiring + CodeRabbit fixes), running `make kwok-test-deployer RECIPE=eks-training DEPLOYER=<d>` against a local Kind cluster produces:

| Deployer | Result | Root cause |
|---|---|---|
| `helm` | ✅ PASS | Regression baseline — helm-path is byte-identical to `main` |
| `argocd-oci` | ❌ FAIL | Argo CD `$values` multi-source ref is **Git-only**. The child Applications rendered by `pkg/bundler/deployer/argocd/templates/application.yaml.tmpl` use `$values/<dir>/values.yaml` against the OCI repoURL → Argo CD's repo-server attempts `git ListRefs` on `oci://...` → `unsupported scheme "oci"`. Blocks the lane until the bundler template inlines values for OCI output. Tracked: **#960** |
| `argocd-helm-oci` | ❌ FAIL | `helm upgrade --install oci://... --version <tag> --plain-http` returns `unable to locate any tags in provided repository`. AICR's OCI push uses `application/vnd.oci.image.layer.v1.tar+gzip` for the layer + `application/vnd.nvidia.aicr.artifact` for the manifest config — Argo CD's raw OCI source accepts this, but Helm OCI likely expects `application/vnd.cncf.helm.chart.content.v1.tar+gzip` + `application/vnd.cncf.helm.config.v1+json`. Tracked: **#961** |
| Deliberate regression (`destination.server` removed) | Not run | Pending decision on argocd-oci scope |

## What's already fixed in commit `4f9a7c05`

Seven distinct issues found by actually running the lanes:

1. **`validate-scheduling.sh`** — OCI bundle invocation needs `cd "${WORK_DIR}"` because `aicr bundle --output oci://...` writes the local bundle to `./bundle` (CWD-relative). Without the `cd`, the rendered `app-of-apps.yaml` was being looked up at `${WORK_DIR}/bundle/` but lived in the repo root.
2. **`validate-scheduling.sh`** — Pass `--repo oci://registry.aicr-registry.svc.cluster.local:5000/aicr/<recipe>` explicitly. The auto-derive in `pkg/cli/bundle.go:172` would set both push and pull URLs to `oci://localhost:5500/...` (the runner's host port), but Argo CD's repo-server runs in-cluster and must use the Service DNS.
3. **`install-infra.sh`** — Switch the Argo CD secret from `secret-type: repository` to `secret-type: repo-creds`. Verified against `util/db/repository_secrets.go`: Repository uses exact URL match (`git.SameURL`), one secret per recipe wouldn't scale. RepoCreds uses longest-prefix match — one template at the `…/aicr` prefix covers every recipe.
4. **`kwok/kind-config.yaml`** — Re-enable CoreDNS. The historical skip was correct for KWOK-only scheduling validation; with Argo CD pods in the cluster, they need in-cluster DNS to reach `argocd-redis` and to resolve the OCI registry's Service DNS. (Symptom before fix: `redis: connection pool: failed to dial after 1 attempts: dial tcp: lookup argocd-redis: i/o timeout`.)
5. **`validate-scheduling.sh`** — Helm OCI install syntax. The deploy step was `helm upgrade --install <release> oci://host/repo:<tag>`; Helm OCI does NOT accept the `:<tag>` suffix on the ref. Fix: `helm upgrade --install <release> oci://host/repo --version <tag>`.
6. **`run-all-recipes.sh`** (CodeRabbit) — Reset `consecutive_sync_timeouts` on non-timeout failures so the ADR-008 3-strike rule fires only on genuinely consecutive sync timeouts.
7. **`validate-scheduling.sh`** (CodeRabbit) — `apps_rc=$?` was inside the `if \! kubectl get ...; then` block; the shell's negation rewrote the captured status, so the diagnostic was logging `rc=0` on failure. Fix: move the `apps_rc=$?` capture before the `if`.

## How to resume from another machine

```bash
git fetch origin
git checkout feat/kwok-deployer-matrix-843
git log --oneline main..HEAD
# Should show 2 commits:
#   4f9a7c05 fix(kwok): live-test wiring fixes + CodeRabbit findings (#843)
#   5bc50950 feat(ci): KWOK deployer matrix — helm + argocd-oci + argocd-helm-oci (#843)

unset GITLAB_TOKEN && make build
make kwok-test-deployer RECIPE=eks-training DEPLOYER=helm           # passes
make kwok-test-deployer RECIPE=eks-training DEPLOYER=argocd-oci     # fails — blocked by #960
make kwok-test-deployer RECIPE=eks-training DEPLOYER=argocd-helm-oci # fails — blocked by #961
# Between deployers: kind delete cluster --name aicr-kwok-test

# macOS port-5000 collision (ControlCenter):
# Temporarily edit kwok/kind-config.yaml hostPort: 5000 → 5500 and run with
# KWOK_REGISTRY_HOST_PORT=5500 make kwok-test-deployer …
# DO NOT commit that edit.
```

For diagnostics during a stuck argocd-oci run:

```bash
kubectl --context kind-aicr-kwok-test get applications -A
kubectl --context kind-aicr-kwok-test -n argocd get app nvidia-stack -o json | jq '.status'
kubectl --context kind-aicr-kwok-test -n argocd logs deploy/argocd-repo-server --tail=200
```

## Open decisions

- **Scope of this PR**: ship `{helm}` only with the wiring scaffolding for future deployer lanes, then enable `argocd-oci` / `argocd-helm-oci` in follow-up PRs as #960 / #961 land? Or keep the full matrix declaration in CI and accept the red Tier 1 cells until those issues resolve?
- **ADR-008 sync**: the ADR still says push raw bundle as OCI + Argo CD pulls + `directory.recurse`. The argocd-oci portion of that is blocked by #960 (`$values` is Git-only); the argocd-helm-oci portion is blocked by #961 (Helm OCI mediaType). The "matrix achievable as designed" finding from the spike was over-optimistic.
- **PR status**: currently draft. Don't merge until #960 + #961 are resolved or argocd-* lanes are explicitly removed.

## What's NOT in this PR

- Phase 5.2 deliberate-regression acceptance test — pending unblocking of at least one argocd-* lane.
- ADR-008 update — left as-is per earlier direction.
- Bundler template inline-values change (would fix #960) — out of scope; substantial refactor.
- Bundler OCI Helm chart mediaType fix (would fix #961) — out of scope; `pkg/oci/push.go` mediaType conditional.

kwok/scripts/validate-scheduling.sh
```diff
@@ -567,17 +1758,29 @@ main() {
...
+            exit 1
+            ;;
+    esac
+
+    # Must run before cleanup trap so the trap sees the resolved root name.
+    resolve_argocd_root_app
+    resolve_flux_root_names "$recipe"
+
     # Set up cleanup trap
     trap cleanup EXIT
```
njhensley, line 1802: **Blocker.** This force-finalize path runs before `verify_kwok_nodes` (step 3), so the only thing standing between a misconfigured `KUBECONFIG` and a finalizer wipe on whatever cluster the user is authenticated to is the static system-namespace allowlist. Since `validate-scheduling.sh` is independently invokable (per the header `./validate-scheduling.sh <recipe>`) and the `EXIT` trap at line 1779 also fires unconditionally, I'd add a positive identity gate up front — before either path can run.

Something like:
```bash
ensure_kwok_context() {
  local ctx; ctx=$(kubectl config current-context 2>/dev/null || true)
  if [[ "$ctx" != "kind-${CLUSTER_NAME:-aicr-kwok}" ]]; then
    log_error "refusing to run: current kubectl context is '$ctx', not a KWOK Kind cluster"
    exit 1
  fi
  kubectl get node -l type=kwok -o name | grep -q . \
    || { log_error "refusing to run: no kwok-typed nodes in current context"; exit 1; }
}
```

Call it before `trap cleanup EXIT` at line 1779 and before `cleanup_old_tests` at line 1790. The same guard should fire from `run-all-recipes.sh:cleanup_between_tests` (line 157) for the same reason.

pkg/oci/helm.go
```diff
@@ -0,0 +1,453 @@
...
+	// marshal is therefore mandatory — the chartYAML struct itself is
+	// stable, but `yaml.v3` makes formatting choices (block-vs-flow
+	// scalar style, line wrapping) that can vary; using the project's
+	// deterministic serializer is the lighter belt-and-suspenders
+	// matching the rest of the digest-feeding paths in this repo.
+	rewritten, err := serializer.MarshalYAMLDeterministic(meta)
+	if err != nil {
+		return nil, apperrors.Wrap(apperrors.ErrCodeInternal, "failed to marshal Chart.yaml", err)
+	}
+	if err := os.WriteFile(chartPath, rewritten, 0o600); err != nil {
```
njhensley, line 307: `loadAndRewriteChartYAML` rewrites `Chart.yaml` in place inside the caller-supplied `SourceDir`. The push completes cleanly, but the source dir is now mutated — a subsequent push to a different tag, or a code path that re-derives version from `Chart.yaml` between push attempts, reads stale tag data. It's also non-atomic; a crash mid-write leaves a corrupt `Chart.yaml`.

Consider copying `SourceDir` to a temp dir at the top of `PackageAndPushHelmChart`, rewriting and packaging from the temp copy, and leaving the caller's source dir untouched. The same approach also fixes the atomicity concern for free.

kwok/scripts/validate-scheduling.sh
```diff
@@ -64,9 +197,80 @@ cleanup() {
...
+                || true)
+            if [[ -n "$stale_hrs" ]]; then
+                while IFS=' ' read -r ns name; do
+                    if [[ -n "$ns" && "$ns" != "flux-system" ]]; then
+                        log_info "Deleting HelmRelease ${name} in ${ns}..."
+                        timeout 30s kubectl delete helmrelease "$name" -n "$ns" \
+                            --ignore-not-found --wait=false 2>/dev/null || true
+                        # Strip finalizers if helm-controller's reconciler is stuck.
+                        kubectl patch helmrelease "$name" -n "$ns" \
+                            --type=json -p='[{"op":"remove","path":"/metadata/finalizers"}]' \
```
njhensley: RFC 6902 `remove` errors when the target path is absent; the `|| true` then silently swallows it, which means this guard is a no-op on exactly the case it exists for — a HelmRelease whose helm-controller crashed before attaching its finalizer.

Switch to a merge patch:
```bash
kubectl patch helmrelease "$hr" -n "$ns" \
  --type=merge -p '{"metadata":{"finalizers":null}}' >/dev/null 2>&1 || true
```
Idempotent whether finalizers are present or absent.

kwok/scripts/run-all-recipes.sh
```diff
@@ -124,52 +164,201 @@ cleanup_between_tests() {
...
+    # rationale: KWOK cannot run real controller finalizers, so
+    # `kubectl wait --for=delete` against a Terminating namespace blocks
+    # the full --timeout=120s per namespace. Recipes touch ~10 namespaces;
+    # the previous version spent ~20 minutes of cleanup between recipes
+    # while doing zero useful work. Force-finalize bypasses the protocol
+    # by PATCHing spec.finalizers to []; safe in the ephemeral KWOK
+    # cluster (no real workloads to leak; cluster destroyed at job end).
+    local system_ns="default|kube-node-lease|kube-public|kube-system|kwok-system|local-path-storage|argocd|aicr-registry|flux-system"
     local terminating
     terminating=$(kubectl get ns -o jsonpath='{range .items[?(@.status.phase=="Terminating")]}{.metadata.name}{"\n"}{end}' 2>/dev/null | grep -vE "^(${system_ns})$" || true)
```
njhensley: This only iterates namespaces already in `status.phase=="Terminating"`. A namespace that's still `Active` with a pending `deletionTimestamp` is skipped, and the next recipe collides on the existing namespace name. The full `cleanup_old_tests` in `validate-scheduling.sh` handles this with a two-phase pattern (async `delete --wait=false`, then force-finalize the stuck remainder).

Either mirror the same two-phase logic here, or factor both into a shared helper sourced by both scripts. Otherwise the between-test cleanup silently drifts away from the in-recipe cleanup over time.

kwok/scripts/run-all-recipes.sh
```diff
@@ -124,52 +164,201 @@ cleanup_between_tests() {
...
+    #
+    # See the long comment in validate-scheduling.sh cleanup() for the
+    # rationale: KWOK cannot run real controller finalizers, so
+    # `kubectl wait --for=delete` against a Terminating namespace blocks
+    # the full --timeout=120s per namespace. Recipes touch ~10 namespaces;
+    # the previous version spent ~20 minutes of cleanup between recipes
+    # while doing zero useful work. Force-finalize bypasses the protocol
+    # by PATCHing spec.finalizers to []; safe in the ephemeral KWOK
+    # cluster (no real workloads to leak; cluster destroyed at job end).
+    local system_ns="default|kube-node-lease|kube-public|kube-system|kwok-system|local-path-storage|argocd|aicr-registry|flux-system"
```
njhensley: This regex is the only thing preventing the force-finalize sweep from hitting the install-infra-owned namespaces (`argocd`, `flux-system`, `aicr-registry`). It's now duplicated in three locations — here, `validate-scheduling.sh:290`, and `validate-scheduling.sh:497`. Any drift (typo, missing pipe, one site updated and not the other two) silently strips finalizers from those namespaces and breaks the next recipe.

Hoist to a single sourced constant — either a tiny `kwok/scripts/lib/cleanup.sh` or a `readonly SYSTEM_NS_PATTERN=…` exported from one of the existing scripts.

kwok/scripts/run-all-recipes.sh
```diff
@@ -124,52 +164,201 @@ cleanup_between_tests() {
...
+    # genuinely consecutive sync timeouts trip the bail. A pass-fail-pass-
+    # timeout-timeout-timeout sequence is 3 consecutive; a timeout-pass-
+    # timeout-timeout sequence is 2 consecutive.
+    # Only argocd-* deployers can hit 50; helm path never trips this.
+    local consecutive_sync_timeouts=0
+
     for recipe in ${recipes}; do
-        if run_recipe_test "${recipe}"; then
+        local rc=0
+        run_recipe_test "${recipe}" || rc=$?
```
njhensley, line 376: The 3-strike counter at line 346 increments on any `rc==50`, but per the script header at line 45 it's only meaningful for argocd-* paths. Helm and flux paths can't return 50 today, but the gate is implicit. Tighten to make intent explicit:

```bash
if (( rc == EXIT_ARGOCD_SYNC_TIMEOUT )) && [[ "${DEPLOYER}" == argocd-* ]]; then
    consecutive_sync_timeouts=$(( consecutive_sync_timeouts + 1 ))
    # …
fi
```

pkg/oci/helm.go
```diff
@@ -0,0 +1,453 @@
...
+// loadAndRewriteChartYAML reads Chart.yaml from sourceDir, sets its
+// version to tag, and writes the result back. Returns the parsed chart
+// metadata so callers can build the OCI config blob without re-parsing.
+//
+// The rewrite is intentional and idempotent — see
+// PackageAndPushHelmChart for the chart-version-equals-OCI-tag
+// invariant.
+func loadAndRewriteChartYAML(sourceDir, tag string) (*chartYAML, error) {
+	chartPath := filepath.Join(sourceDir, "Chart.yaml")
+	data, err := os.ReadFile(chartPath) //nolint:gosec // path is under caller-supplied source dir
```
njhensley: Unbounded `os.ReadFile` — flagged as an anti-pattern in CLAUDE.md. `SourceDir` is caller-supplied, so under unusual ops conditions (a SourceDir pointing at a network mount or a symlink into `/proc`) this could OOM the process before the YAML parse fails. Wrap with `os.Open` + `io.LimitReader` against a Chart.yaml-shaped cap (~1 MiB) from `pkg/defaults`.

pkg/oci/helm.go
```diff
@@ -0,0 +1,453 @@
...
+// which the OCI registry then dedups.
+//
+// ctx is checked on entry and per directory entry so a parent timeout
+// (push deadline, server shutdown) terminates the walk without
+// orphaning a partially-written buffer.
+func buildHelmChartTGZ(ctx context.Context, sourceDir, chartName string) ([]byte, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, apperrors.Wrap(apperrors.ErrCodeUnavailable, "chart packaging canceled", err)
+	}
+	if chartName == "" {
```
njhensley: `chartName` is rejected when empty but accepted with `/`, `..`, or control chars; downstream it's used as the tar entry prefix at line 308 (`entry := chartName + "/" + filepath.ToSlash(rel)`). The current CLI flow always supplies a safe literal, but `PackageAndPushHelmChart` is a public function and a future caller with a Chart.yaml-derived name could produce a tarball with entries that escape the chart root on extraction.

Cheap defense: `if !filepath.IsLocal(chartName) || strings.ContainsAny(chartName, "/\x00") { … }`.

pkg/oci/helm.go
```diff
@@ -0,0 +1,453 @@
...
+		"chart_name", chartMeta.Name,
+		"chart_version", chartMeta.Version,
+	)
+
+	chartTGZ, err := buildHelmChartTGZ(ctx, absSourceDir, chartMeta.Name)
+	if err != nil {
+		return nil, err
+	}
+
+	configBlob, err := json.Marshal(chartMeta)
```
njhensley, line 184: Worth a one-line comment pinning the contract: this struct is intentionally the OCI **config blob** (registry-visible metadata), not the full Chart.yaml. Full Chart.yaml content still ships in the tar layer, so a Helm chart key not modeled here is not lost. Without the comment, a future reader is likely to "fix" this by widening the struct or marshaling the raw map.

pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
```diff
@@ -235,6 +235,60 @@ func TestGenerate(t *testing.T) {
...
+// `-pre` folder path that recipes with ComponentPreManifests emit
+// (e.g. the gke-cos OS overlay, which injects a Talos-style driver
+// prerequisite Namespace ahead of gpu-operator). The folder is named
+// `NNN-gpu-operator-pre`; the bundler must strip the `-pre` suffix
+// when resolving the override key, otherwise resolveOverrideKey is
+// called with "gpu-operator-pre" and fails `component %q not found
+// in registry` at bundle time. Caught in the KWOK gke-cos-training
+// CI lane after the pre-fix; before this commit the bug only
+// surfaced for recipes that combined Helm + pre-manifest overlays.
+func TestGenerate_PreManifestParentResolution(t *testing.T) {
```
njhensley, line 248: `-pre` resolution covered here, `-post` covered elsewhere, but no test exercises a component with **both** a `-pre` and a `-post` sibling — which is the actual gke-cos OS overlay shape described in the production-code comment. The loop `for _, suffix := range []string{"-pre", "-post"}` runs twice across two distinct folder iterations in that case, and the override-key lookup for each must resolve back to the same parent.

A new row (or a sibling subtest) staging `001-gpu-operator-pre/` and `003-gpu-operator-post/` against a single registered `gpu-operator` and asserting both wrapper templates point at the same parent override-key would cover the case for ~20 lines.

pkg/oci/helm_test.go
```diff
@@ -0,0 +1,646 @@
...
+		}
+	}
+}
+
+// TestBuildHelmChartTGZ_Reproducible: two builds against the same
+// source tree must produce byte-identical tar.gz so the registry
+// dedups the blob. A regression (e.g., embedding time.Now() in the
+// gzip header or in tar mtimes) would silently make every push a
+// new blob, defeating registry-level caching.
+func TestBuildHelmChartTGZ_Reproducible(t *testing.T) {
```
njhensley, line 242: Same-process determinism doesn't prove cross-run reproducibility — a regression that caches `time.Now()` once at package init would still pass this test. Consider walking the produced tar headers and asserting absolute invariants:

```go
for _, hdr := range hdrs {
    if !hdr.ModTime.IsZero() { /* fail */ }
    if hdr.Uid != 0 || hdr.Gid != 0 { /* fail */ }
    if hdr.Uname != "" || hdr.Gname != "" { /* fail */ }
}
// plus gzip header MTIME zero check
```

pkg/oci/helm_test.go
```diff
@@ -0,0 +1,646 @@
...
+	}
+}
+
+// newFakeOCIRegistry stands up an httptest server that handles just
+// enough of the OCI distribution-spec to accept a push: blob uploads
+// (POST/PATCH/PUT under /v2/<name>/blobs/uploads/), blob HEADs (to
+// satisfy oras's existence check), and manifest PUTs. Anything else
+// returns 404 — the goal isn't a conformant registry, just the
+// endpoints PackageAndPushHelmChart's push path actually touches.
+func newFakeOCIRegistry(t *testing.T) *httptest.Server {
```
njhensley, line 665: The fake registry accepts manifest PUTs without inspecting `manifest.config.mediaType`, which is the exact field whose handling motivated the #961 fix. The full round-trip (`TestPackageAndPushHelmChart_PushesAgainstFakeRegistry`) currently only asserts the outer manifest envelope mediaType.

Either have the fake assert `manifest.Config.MediaType == helmConfigMediaType` on the manifest PUT handler, or GET the manifest back at the end of the round-trip test and decode it to verify the Helm config mediaType survived.

kwok/scripts/validate-scheduling.sh
```diff
@@ -372,85 +648,460 @@ generate_bundle() {
...
+            log_debug "Components in bundle:"
+            list_bundle_components "$bundle_dir" | head -10
+
+            # Run the generated deploy script without --wait since KWOK clusters
+            # only validate scheduling, not pod readiness
+            chmod +x "${bundle_dir}/deploy.sh"
+            log_info "Running deploy.sh --no-wait..."
+
+            local deploy_output
+            if ! deploy_output=$("${bundle_dir}/deploy.sh" --no-wait 2>&1); then
```
njhensley, line 957: *(Optional, not blocking.)* The four arms here collapse to roughly `(sync == "Synced" || operationState.phase == "Succeeded") && health != "Missing"`. Arms 1–2 reduce to `Synced` (any non-Missing health); arms 3–4 to `phase == Succeeded`. The per-arm prose belongs in ADR-008 §"Terminal Pass Predicate"; the wait loop just needs the predicate. Happy to take this as a follow-up.

njhensley (changes requested):
This is a substantial and well-scoped contribution — the round-trip lanes immediately surfaced three latent bundler bugs, which is exactly the case for the lanes existing. The OCI tag/version handling in `pkg/oci/helm.go` and `pkg/bundler/deployer/argocd/argocd.go` is carefully thought through (inline comments preempt several reasonable misreadings). PR description honestly tracks the deferred items (#963, #964) rather than glossing over them.

I have one blocker and a handful of follow-up suggestions. None of the follow-ups need to block merge if you'd rather chase them as a stacked PR.

**Blocker:** the new force-finalize namespace cleanup path runs against whatever the current kubectl context happens to be, with no positive identity check that it's the KWOK Kind cluster. If `KUBECONFIG` or `kubectl config current-context` points anywhere else when someone runs `validate-scheduling.sh` directly (it's independently invokable per the script header) or via `make` from a stale shell, the EXIT trap and `cleanup_old_tests` will strip finalizers from every non-allowlisted namespace on the live cluster. The pre-existing `--wait=true --timeout=120s` path was slow but did not bypass the finalizer protocol. This expansion to PATCH `spec.finalizers=[]` materially widens blast radius and needs a guard before merge. Concrete fix in the inline comment.

**Themes worth a follow-up PR (not blocking):**

- **System-namespace allowlist drift.** Same regex is copy-pasted in three places (`run-all-recipes.sh:184`, `validate-scheduling.sh:290`, `:497`). Any one of them gaining `flux-system` or `aicr-registry` and the others not will silently strip finalizers from the infra namespaces. Hoist to one sourced constant.
- **Sync-gate complexity.** The four-arm Argo CD sync predicate at `validate-scheduling.sh:944-1056` collapses to roughly `(sync == Synced || phase == Succeeded) && health != Missing`. ~100 lines → ~15. The per-arm prose belongs in ADR-008, not the wait loop.
- **Mirror wait-loops.** `wait_for_argocd_*` and `wait_for_flux_*` are ~250 lines of structural duplication. At minimum, fold `wait_for_*_root_app` into one helper.

**Pre-merge checklist:**
- [ ] Add KWOK-context guard in front of `cleanup_old_tests` and the EXIT trap (blocker below)
- [ ] Rebase onto `origin/main` (PR has the `needs-rebase` label)
- [ ] Optional but recommended: address the in-place `Chart.yaml` rewrite and the HelmRelease finalizer-patch shape in the same push since the fixes are tiny

Nice work overall.

kwok/scripts/validate-scheduling.sh
mchmarny: Addressed in f76ee55a. Added `ensure_kwok_context_loose` (context-name only) and `ensure_kwok_context` (context + kwok-node-label) in `kwok/scripts/lib/cleanup.sh`. The strict variant is called from `validate-scheduling.sh::main()` before `trap cleanup EXIT` (apply-nodes.sh has populated the cluster by then). The loose variant is called from `run-all-recipes.sh::main()` after `ensure_cluster` returns — split because at that point no recipe has applied nodes yet, so the strict label check would false-positive on cold start.

pkg/oci/helm.go
mchmarny: Addressed in f76ee55a. `PackageAndPushHelmChart` now copies SourceDir to `<OutputDir>/helm-chart-work` and runs `loadAndRewriteChartYAML` + `buildHelmChartTGZ` against the copy. The caller's source tree stays byte-identical; the atomicity concern is gone too (crash mid-write leaves a corrupt file in the disposable work dir, not the caller's tree). Reused the existing `copyDir` helper from push.go.

kwok/scripts/validate-scheduling.sh
mchmarny: Addressed in f76ee55a. Switched the HelmRelease finalizer strip to `--type=merge -p '{"metadata":{"finalizers":null}}'`. Idempotent whether finalizers are present or absent — the previous `--type=json remove` would error when the path was absent and the `|| true` then silently swallowed the exact case the guard was meant for.

kwok/scripts/run-all-recipes.sh
mchmarny: Addressed in f76ee55a. `cleanup_between_tests` now mirrors `validate-scheduling.sh::cleanup_old_tests`'s two-phase shape: phase 1 issues async deletes (`--wait=false`) against every non-system namespace, phase 2 sleeps briefly and force-finalizes anything still present. Also factored the system-ns allowlist into `kwok/scripts/lib/cleanup.sh::SYSTEM_NS_PATTERN` (your other comment) so both cleanup paths now consume the same source.

kwok/scripts/run-all-recipes.sh
mchmarny: Addressed in f76ee55a. Created `kwok/scripts/lib/cleanup.sh` with `readonly SYSTEM_NS_PATTERN=...` and sourced from both scripts. All three sites (`run-all-recipes.sh:cleanup_between_tests`, `validate-scheduling.sh:cleanup`, `validate-scheduling.sh:cleanup_old_tests`) now use `${SYSTEM_NS_PATTERN}` with a `# see SYSTEM_NS_PATTERN in lib/cleanup.sh` annotation pointing back at the canonical definition.

kwok/scripts/run-all-recipes.sh
mchmarny: Addressed in f76ee55a. Strike gate is now `(( rc == EXIT_ARGOCD_SYNC_TIMEOUT )) && [[ "$DEPLOYER" == argocd-* || "$DEPLOYER" == flux-oci ]]`. Note: `flux-oci`'s `wait_for_flux_sync` also returns 50 on deadline (the constant is reused for symmetry of the 3-strike rule across GitOps deployers), so the explicit gate includes both `argocd-*` and `flux-oci`. Helm path stays implicitly excluded.

pkg/oci/helm.go
mchmarny: Addressed in f76ee55a. `loadAndRewriteChartYAML` now opens Chart.yaml and reads it through `io.LimitReader(f, MaxChartYAMLBytes+1)` with a typed `ErrCodeInvalidRequest` if the bound is exceeded. Added `defaults.MaxChartYAMLBytes = 1 MiB` to `pkg/defaults/timeouts.go` next to `MaxConfigBytes` so the cap lives with the other defensive size limits.

pkg/oci/helm.go
mchmarny: Addressed in f76ee55a. New `validateChartName` helper rejects names that fail `filepath.IsLocal` OR contain `/`, `\`, or `\x00`. Called from `loadAndRewriteChartYAML` (when reading the in-source Chart.yaml.name) AND from `buildHelmChartTGZ` (defense-in-depth — `buildHelmChartTGZ` is exported and a future caller could supply chartName from a different source).

pkg/oci/helm.go
mchmarny: Addressed in f76ee55a. Expanded the godoc on `chartYAML` to pin the contract: it's intentionally the OCI **config blob** (registry-visible metadata), NOT a model of full Chart.yaml; full Chart.yaml content still ships in the tar layer; widening this struct is the wrong fix.

pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
mchmarny: Addressed in f76ee55a. New `TestGenerate_PreAndPostManifestParentResolution` stages a single registered `gpu-operator` with both `ComponentPreManifests` AND `ComponentPostManifests` populated, then asserts both `001-gpu-operator-pre/` and `003-gpu-operator-post/` folders exist AND both `templates/gpu-operator-pre.yaml` and `templates/gpu-operator-post.yaml` wrapper templates were produced. Catches the asymmetric-suffix-strip regression the production code comment warns about.

pkg/oci/helm_test.go
mchmarny: Addressed in f76ee55a. `TestBuildHelmChartTGZ_Reproducible` now walks the produced tar and asserts per-entry `ModTime` / `AccessTime` / `ChangeTime` are zero (or Unix epoch — the tar wire format normalizes the Go zero time to epoch), `Uid`/`Gid` are 0, `Uname`/`Gname` are empty, AND the outer `gzip.Header.ModTime` is zero. A `time.Now()` cached at package-init that passed the same-process byte-equality would now fail these absolute-value assertions.

pkg/oci/helm_test.go
mchmarny: Addressed in f76ee55a. `TestPackageAndPushHelmChart_PushesAgainstFakeRegistry` now round-trips: after the push, it GETs the manifest back from the fake registry via HTTP, decodes the JSON, and asserts `manifest.Config.MediaType == helmConfigMediaType` AND `manifest.Layers[0].MediaType == helmLayerMediaType`. The exact field whose handling motivated #961 is now under test.

kwok/scripts/validate-scheduling.sh
mchmarny: Tracking as a follow-up — not addressing in this PR per your note. Filed a mental note to revisit when the ADR-008 §"Terminal Pass Predicate" section gets the predicate-collapse rewrite.

kwok/scripts/validate-scheduling.sh
```diff
@@ -348,88 +463,458 @@ generate_bundle() {
...
+        # the negation rewrites the captured status.
+        apps_json=$(kubectl get applications -n argocd -o json 2>/dev/null)
+        apps_rc=$?
+        if (( apps_rc != 0 )); then
+            log_error "kubectl get applications -n argocd failed (rc=${apps_rc})"
+            log_error "Likely cause: applications.argoproj.io CRD missing, RBAC denied, or apiserver unreachable."
+            log_error "--- kubectl get applications stderr ---"
+            kubectl get applications -n argocd 2>&1 | tail -20 || true
+            dump_argocd_failures
+            return 1
```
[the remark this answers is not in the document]
mchmarny: Verified against current code — the if/then/else wrap around `apps_json=$(kubectl ...)` is already in place (`wait_for_argocd_sync` and `wait_for_flux_sync` both use the pattern). The current squash carries the fix; no further change needed.

kwok/scripts/validate-scheduling.sh
```diff
@@ -348,88 +463,458 @@ generate_bundle() {
...
+                log_warn "Bundle may have been wrapped-only (no expansion). total=${total}"
+                # Surfacing as success would mask the bug from MUST-FIX 1.
+                # Fail closed: a child-less root is never a real PASS for
+                # the deployer matrix (every recipe ships >= 1 component).
+                dump_argocd_failures
+                return 1
+            fi
+            log_debug "Root Application present but children not yet declared (total=${total})..."
+            sleep 5
+            continue
```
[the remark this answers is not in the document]
mchmarny: Verified against current code — the non-root Application fallback is already implemented in `wait_for_argocd_sync` (counts non-root Applications as observed_children when status.resources is empty). The current squash carries the fix; no further change needed.

pkg/oci/helm.go
```diff
@@ -0,0 +1,397 @@
...
+	}
+	meta.Version = tag
+
+	rewritten, err := yaml.Marshal(&meta)
+	if err != nil {
+		return nil, apperrors.Wrap(apperrors.ErrCodeInternal, "failed to marshal Chart.yaml", err)
+	}
+	if err := os.WriteFile(chartPath, rewritten, 0o600); err != nil {
+		return nil, apperrors.Wrap(apperrors.ErrCodeInternal, "failed to rewrite Chart.yaml with OCI tag version", err)
+	}
```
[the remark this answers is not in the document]
mchmarny: Analysis chain only — no actionable change requested.

kwok/scripts/lib/cleanup.sh
```diff
@@ -0,0 +1,77 @@
...
+# recipes.sh's initial cleanup before any recipe has applied nodes).
+#
+# Hard-fail on mismatch — the cleanup paths the caller is about to run
+# force-finalize namespaces, and we'd rather refuse to act than risk
+# wiping a real cluster.
+ensure_kwok_context_loose() {
+    local ctx
+    ctx=$(kubectl config current-context 2>/dev/null || true)
+    case "$ctx" in
+        kind-aicr-kwok|kind-aicr-kwok-test)
```
njhensley, line 51: *(Minor, non-blocking follow-up.)* This hardcodes acceptance of `kind-aicr-kwok` or `kind-aicr-kwok-test`, but `KWOK_CLUSTER` is a publicly documented env var (see `.github/actions/kwok-test/action.yml:115` and `run-all-recipes.sh:64`'s `CLUSTER_NAME="${KWOK_CLUSTER:-aicr-kwok-test}"`). A developer running locally with `KWOK_CLUSTER=my-dev-cluster` legitimately creates `kind-my-dev-cluster` and is then hard-blocked by this safety check.

The default CI path is unaffected, but the env var exists specifically to permit non-default names, and the symptom (hard-fail with a generic error) is bad UX for the only population who'd encounter it.

Suggested fix — parameterize against the caller's expected context name:

```bash
ensure_kwok_context_loose() {
    local ctx expected="${1:-${CONTEXT:-}}"
    ctx=$(kubectl config current-context 2>/dev/null || true)
    case "$ctx" in
        kind-*)
            if [[ -n "$expected" && "$ctx" != "$expected" ]]; then
                echo "[ERROR] ensure_kwok_context: context '$ctx' does not match expected '$expected'" >&2
                exit 1
            fi
            return 0
            ;;
        *)
            echo "[ERROR] ensure_kwok_context: context '${ctx:-<none>}' is not a kind- cluster — refusing" >&2
            exit 1
            ;;
    esac
}
```

`run-all-recipes.sh` already computes `CONTEXT="kind-${CLUSTER_NAME}"` at line 65, so callers can pass it through or let the function pick it up from the environment. Preserves the safety invariant (context must be a `kind-` cluster) while honoring the existing customization knob.

njhensley (approved):
Thanks for the thorough response — all 12 actionable items addressed cleanly in `45f17cad`, and the explicit deferral on the sync-gate predicate collapse (with ADR-008 reference) is acceptable.

A few highlights worth calling out:

- The `os.MkdirTemp` staging approach with `TestPackageAndPushHelmChart_SourceEqualsOutputDir` as a regression guard is exactly the right shape. The self-discovered same-dir recursion bug being captured as a test is good engineering.
- The widened `defaults.MaxChartYAMLBytes` + `validateChartName` (called from both `loadAndRewriteChartYAML` AND the exported `buildHelmChartTGZ`) gives the OCI push path solid defense-in-depth.
- The strike-gate widening to include `flux-oci` (rather than just argocd-*) is a sharper read than the original review — `wait_for_flux_sync` also returns 50, and gating on both deployers is the right call.
- The two-phase `cleanup_between_tests` mirroring `cleanup_old_tests`, with `SYSTEM_NS_PATTERN` centralized in `lib/cleanup.sh`, eliminates the drift class entirely.

One minor follow-up surfaced from reading the new safety guard — see the inline comment. Not blocking; consider for a stacked PR.

Approving.

PR #957, opened 2026-05-18
Warn on missing PVC storageClassName in bundles

### LINKED ISSUE #898: Bundle silently omits storageClassName when --storage-class is not provided — deploy hangs on Pending PVC
## Motivation

#503 introduced `--storage-class` at bundle time as the deployer-facing seam
for injecting a cluster-specific StorageClass into components that need
persistent storage. That flag is **optional**, and `aicr bundle` exits 0
when it is omitted — but the resulting bundle can be a silent footgun.

## Problem

Cloud service overlays configure `kube-prometheus-stack` with a
`volumeClaimTemplate` but deliberately leave `storageClassName` unset, so the
PVC falls through to whatever StorageClass the target cluster has marked as
default. See [`recipes/overlays/eks.yaml`](https://github.com/NVIDIA/aicr/blob/main/recipes/overlays/eks.yaml#L50),
[`recipes/overlays/gke-cos.yaml`](https://github.com/NVIDIA/aicr/blob/main/recipes/overlays/gke-cos.yaml#L49),
[`recipes/overlays/aks.yaml`](https://github.com/NVIDIA/aicr/blob/main/recipes/overlays/aks.yaml#L80),
[`recipes/overlays/oke.yaml`](https://github.com/NVIDIA/aicr/blob/main/recipes/overlays/oke.yaml#L52).

That fallback works on clusters with a default class, but a real cluster
without one (or with a default whose provisioner can't satisfy `ReadWriteOnce`
+ 50Gi) leaves the PVC `Pending` indefinitely. `deploy.sh` hangs on the
component's `helm install --wait`, with no signal back to `aicr` that the
upstream gap was a missing `--storage-class`.

The current state:

- `aicr bundle` accepts (or omits) `--storage-class` with no diagnostic
- The registry already declares `storageClassPaths` for affected components
  (`recipes/registry.yaml#L311`)
- Demo CUJ guides now call this out in prose (#897), but that's
  documentation working around a tooling gap

## Proposal

Make the missing-StorageClass case visible at one of the earlier seams in the
pipeline. Possible levers, not prescriptive — pick whichever fits best:

1. **Bundle-time warning.** If any component in the recipe has a registry
   `storageClassPaths` entry and the rendered value at that path ends up
   empty (no overlay value, no `--set`, no `--storage-class`), emit
   `slog.Warn` listing the components and paths. Cheap, non-blocking,
   surfaces the gap on the *aicr* side rather than after `helm install`
   hangs.
2. **`--storage-class` becomes required when the recipe needs it.** Fail
   `aicr bundle` with `ErrCodeInvalidRequest` if the recipe references a
   component with `storageClassPaths` and no class is resolvable. Stricter;
   has the downside of breaking clusters that *do* have a usable default
   StorageClass and currently rely on it.
3. **`deploy.sh` runtime check.** Have the generated installer probe
   `kubectl get storageclass` for a default before running components that
   create PVCs, and print a clear error if one is needed but absent.
   Closest to where the failure manifests, but happens latest.

Option 1 is the smallest delta and matches the existing
`storageClassPaths` registry contract. Option 3 catches the case even when
the recipe relies on a cluster default that turns out to be missing, but it
runs after `aicr` has already handed off to the installer.

## Acceptance

- A user who runs `aicr bundle` against a recipe that produces a PVC and
  omits `--storage-class` sees a clear signal — warning, error, or
  installer-side check — instead of an apparently-successful bundle that
  hangs at `helm install` time.
- The signal names the component(s) and the registry path(s) that would
  have received the value, so the remediation (`--storage-class <name>`
  or `--set <component>:<path>=<name>`) is obvious.

## Notes

- Closes the UX loop opened by #503.
- The demo CUJ guides (#897) document the workaround in prose, but
  that surface decays — a tooling check is more durable.

yuanchen8911:
## Summary

Emit a non-blocking bundle warning when rendered component values create a PVC without a usable `storageClassName`.

Document the behavior and add regression coverage for PVC, `emptyDir`, global `--storage-class`, and explicit blank `--set` precedence cases.

## Motivation / Context

Cloud overlays can render `kube-prometheus-stack` with a `volumeClaimTemplate` but no `storageClassName`. On clusters without a default StorageClass, deployment can hang on a Pending PVC. This warns users at bundle time without making storage class selection a hard requirement.

Fixes: #898
Related: N/A

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Implementation Notes

The warning runs after component values are resolved, so it respects overlay, `--storage-class`, and explicit per-component `--set` precedence. It only warns when the PVC spec parent exists, avoiding false positives for `emptyDir` storage configurations.

## Testing

```bash
go test ./pkg/bundler -run 'TestApplyNodeSchedulingOverrides_StorageClass|TestWarnMissingStorageClassForPVCs' -count=1
env -u GITLAB_TOKEN make qualify
env GOFLAGS=-mod=vendor go build -o /tmp/aicr/issue-898-aicr2/aicr-reviewed ./cmd/aicr
/tmp/aicr/issue-898-aicr2/aicr-reviewed bundle --recipe /tmp/aicr/issue-898-aicr2/recipe.yaml --storage-class gp2 --set 'kube-prometheus-stack:prometheus.prometheusSpec.storageSpec.volumeClaimTemplate.spec.storageClassName= ' --output /tmp/aicr/issue-898-aicr2/bundle-blank-set
```

`make qualify` completed successfully. The `aicr2` EKS smoke test confirmed the warning still fires when a blank per-component `--set` wins over global `--storage-class`.

## Risk Assessment

**Rollout notes:** N/A. This only adds a warning note and does not change generated values.

## Checklist

### DIFF
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -1190,6 +1190,8 @@ aicr bundle --recipe recipe.yaml \
 
 When `--storage-class` is not set, any `storageClassName` values already defined in the recipe overlays are preserved as defaults. When it is set, `--set <component>:<path>=<value>` on the same path still wins — `--storage-class` only fills in paths that were not explicitly overridden.
 
+If a rendered component creates a PVC at a registry-declared `storageClassPaths` entry and no usable `storageClassName` is set after overlay, `--storage-class`, and `--set` precedence is resolved, `aicr bundle` emits a non-blocking warning. The bundle still relies on the target cluster's default StorageClass in that case.
+
 #### Deployment Methods
 
 The `--deployer` flag controls how deployment artifacts are generated:
--- a/pkg/bundler/bundler.go
+++ b/pkg/bundler/bundler.go
@@ -284,6 +284,10 @@ func (b *DefaultBundler) Make(ctx context.Context, recipeResult *recipe.RecipeRe
 			"failed to extract component values", err)
 	}
 
+	if warningErr := b.warnMissingStorageClassForPVCs(ctx, recipeResult, componentValues); warningErr != nil {
+		return nil, warningErr
+	}
+
 	// Run component-specific validations
 	if validationErr := b.runComponentValidations(ctx, recipeResult); validationErr != nil {
 		return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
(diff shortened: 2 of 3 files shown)


mchmarny (approved):
Clean, focused change. The warning fires after overlay/`--storage-class`/`--set` precedence is resolved, which is the right hook point, and the `storageClassPathHasPVCSpec` parent-must-be-a-map check is what keeps `emptyDir` and other storage modes from false-positiving. The `appendWarning` extraction is a nice incidental DRY-up of the existing validation path. Test matrix covers all the interesting combinations (omitted, blank, set, emptyDir, global, blank-beats-global). `if b.Config == nil` and `ErrCodeTimeout` on `ctx.Err()` both mirror existing patterns in this file. CI green on the completed checks; E2E still running at review time but unrelated to the change surface.

Issue #969, opened 2026-05-19
Close deployment-phase validation coverage gaps for accelerator-bound GPU recipes


yuanchen8911:
## Summary

Audit of \`recipes/overlays/\` showed that accelerator-bound GPU overlays (h100-\*, gb200-\*, rtx-pro-6000-\*) resolved to conformance-only validation, missing the deployment-phase \`operator-health\` / \`expected-resources\` / \`gpu-operator-version\` / \`check-nvidia-smi\` checks and — critically — the \`Deployment.gpu-operator.version\` constraint that makes \`gpu-operator-version\` a real assertion rather than a no-op.

**Scope (rescoped):**
- **In scope:** deployment-phase coverage for **accelerator-bound recipes** — every recipe a user can actually deploy (specifies \`--accelerator\` on the CLI, or has an \`accelerator:\` criterion that selects one).
- **Out of scope:** accelerator-unbound intermediates (e.g., \`eks\`, \`eks-training\`, \`aks-inference\`). These are not user-facing deployable recipes — \`aicr recipe --service eks --intent training\` without \`--accelerator\` produces a bundle with no accelerator-specific GPU operator config (no driver pin, no chart values per accelerator). The deployment-phase version pin has no meaningful value at the intent layer because the pin is accelerator-specific (H100 ≥ v24.6.0, GB200 ≥ v25.10.0, RTX Pro 6000 ≥ v25.10.0). The original 37-entry audit listed these intermediates, but covering them would require either inventing a service-scoped version pin (no anchor in reality) or shipping accelerator-unspecified deployment checks that don't help operators.
- **Performance-phase work:** tracked separately (see "Tracked separately" below).

## Background — how validation resolves

The validation block merges per-phase across the \`base:\` chain (\`pkg/recipe/metadata.go:459-477\`). Each leaf overlay's effective coverage is the OR of \`deployment\` / \`performance\` / \`conformance\` phases set anywhere in its chain.

Service-root overlays historically defined only \`conformance:\`. Deployment-phase coverage was therefore absent from every descendant that did not declare its own \`deployment:\` block — which was most of them.

## Resolution

[...]

PR #989, opened 2026-05-20
test(recipe): enforce per-intent validation phase floor on overlay leaves

### LINKED ISSUE #970: Add CI guard to enforce per-intent validation phase coverage on recipe overlays
## Summary

Add a CI test that resolves every overlay's `base:` chain and rejects any new or modified overlay whose resolved validation falls below the per-intent floor. Closes the loophole that let 27 of 41 GPU overlays drift to conformance-only without anyone noticing (see companion issue on closing the gaps).

## Background

The validation block merges per-phase across the `base:` chain (`pkg/recipe/metadata.go:459-477`). There is no test today that asserts a leaf overlay's resolved validation meets a per-intent contract — so a new overlay can be added that inherits only the service-root `conformance:` block and ship with no GPU operator-health check and no NCCL or `inference-perf` gate.

## Scope

For each overlay in `recipes/overlays/` (excluding intermediates like `eks-training.yaml` that are not user-selectable):

1. Walk the `base:` chain and resolve the merged validation block per `pkg/recipe/metadata.go` `Merge` semantics.
2. Classify the overlay by `criteria.intent` and `criteria.service` plus filename heuristics (Dynamo / NIM / plain inference, Kind vs non-Kind).
3. Assert that the resolved validation contains the required phases for that classification.

### Per-intent floor

| Intent | Required (blocking) | Recommended (warn-only) |
|---|---|---|
| Training (non-Kind) | deployment + conformance | performance (NCCL) |
| Inference Dynamo / NIM (non-Kind) | deployment + conformance | performance (`inference-perf`) |
| Inference (plain) | deployment + conformance | — |
| Kind (any intent) | deployment + conformance | — |

**Why performance is recommended-only initially:** some service overlays (notably Azure AKS and Oracle OKE) lack a performance testbed today (see companion issue). Tighten to required once testbeds exist.

## Implementation notes

- Wire the check into `make qualify` so local runs catch it before push.
- On failure, print the overlay name, its classification, resolved phases, and which phase is missing — so the message is self-explanatory.
- Consider adding a `validation: skip-deployment` (or similar) escape hatch only if a legitimate use case appears; default behavior should be strict.
- Future: when Azure / OCI performance testbeds exist, tighten performance from recommended to required for those services.

## Done when

- New overlay added without required phases → test fails with a clear message.
- Existing overlay modified to drop a required phase → test fails.
- `make qualify` runs the check by default; CI workflow runs it on every PR that touches `recipes/overlays/`.

## Related

- Companion issue: close the existing 27-overlay phase coverage gap.

mchmarny:
## Summary

Adds a Go test that walks each leaf overlay's `spec.base` chain, resolves the merged `ValidationConfig` per `pkg/recipe/metadata.go` Merge semantics, and asserts the resolved validation contains the per-intent required phases. Closes the loophole that let leaf overlays drift to conformance-only without a CI gate.

## Motivation / Context

Today there is no test that asserts a leaf overlay's resolved validation meets a per-intent contract — so a new overlay can inherit only the service-root `conformance:` block and ship with no GPU operator-health check, no NCCL gate, and no `inference-perf` gate. 27 of 41 GPU overlays already drifted this way.

This is the *static* gate from #970. The *runtime* gate (signed evidence bundles from #751) only proves the declared validators ran — it can't catch a missing declaration. #970 forces the declaration to be complete; #751 forces it to actually run.

Fixes: #970
Related: #969, #751

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

## Implementation Notes

**Per-intent floor enforced today:**

| Intent | Required (blocking) | Recommended (warn-only) |
|---|---|---|
| Training, non-Kind | deployment + conformance | performance (NCCL) |
| Inference Dynamo / NIM, non-Kind | deployment + conformance | performance (`inference-perf`) |
| Inference (plain) | deployment + conformance | — |
| Any intent, Kind | deployment + conformance | — |

**Strict mode:** `AICR_VALIDATION_FLOOR_STRICT=1` globally promotes the recommended performance phase from warn-only to required. Default OFF until #969 closes the data gap and Azure/OCI performance testbeds land. When ready, flip this in CI to tighten enforcement.

**Known-gap allowlist:** the test ships with a `knownGaps` map listing the 7 overlays that fail the floor today, so the contract can land independently of the data fix in #969. Entries downgrade `Errorf` to a `Logf` prefixed with `KNOWN GAP (#969):`. As #969 lands per-overlay fixes, drain this map. New overlays added without the floor will fail because they are not allowlisted.

**Wildcard fragment exclusion:** overlays with `criteria.intent == any` or `criteria.service == any` (`b200-any-training`, `gb200-any-training`, `monitoring-hpa`) are cross-cutting fragments — not user-selectable leaves — and are excluded from the gate. See `docs/contributor/data.md#criteria-wildcard-overlays`.

**No new CI wiring needed:** runs under `make test`, which `make qualify` already invokes; CI's existing `pkg/recipe/...` test job picks it up on every PR.

**Deferred (per scope):**
- Per-phase `validation: skip-deployment` escape hatch — wait for first legitimate case
- YAML-side `metadata.annotations` opt-out — would require modifying `RecipeMetadataHeader`; `knownGaps` serves the same purpose for now
- `aicr recipe lint` CLI surface — not requested

## Testing

```bash
make qualify     # passed
# default mode: TestOverlayValidationPhaseFloor — PASS (17 leaves, 7 known gaps logged)
AICR_VALIDATION_FLOOR_STRICT=1 go test ./pkg/recipe/ -run TestOverlayValidationPhaseFloor
# strict mode: PASS for allowlisted overlays + 1 non-allowlisted Azure/Dynamo gap caught
```

**Coverage:** `pkg/recipe`: 89.8% → 89.8% (test-only addition).

## Risk Assessment

- [x] **Low** — Isolated test addition, no production code touched, easy to revert. `knownGaps` allowlist absorbs the existing 7 failures so CI stays green; the gate triggers only on *new* overlays that ship below the floor.

**Rollout notes:** N/A. To enable strict performance enforcement after #969 closes and testbeds exist, set `AICR_VALIDATION_FLOOR_STRICT=1` in the CI test step.

## Checklist

- [x] I updated docs if user-facing behavior changed (no user-facing change)
- [x] Changes follow existing patterns in the codebase (mirrors `conformance_test.go` and the leaf-discovery pattern at `metadata_store_test.go:512`)

### DIFF
--- a/pkg/recipe/validation_phase_floor_test.go
+++ b/pkg/recipe/validation_phase_floor_test.go
@@ -1,0 +1,323 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// validation_phase_floor_test.go enforces a per-intent validation phase
+// floor on every leaf overlay. For each user-selectable overlay it walks
+// the spec.base chain, resolves the merged ValidationConfig per
+// pkg/recipe/metadata.go Merge semantics, classifies the overlay by
+// intent/service/platform, and asserts the resolved validation contains
+// the required phases.
+//
+// Closes the loophole that let 27 of 41 GPU overlays drift to
+// conformance-only without a CI gate (see issue #970, companion #969).
+//
+// Per-intent floor:
+//   Training (non-Kind)               : deployment + conformance   [performance recommended]
+//   Inference Dynamo / NIM (non-Kind) : deployment + conformance   [performance recommended]
+//   Inference (plain)                 : deployment + conformance
+//   Kind (any intent)                 : deployment + conformance
+//
+// Strict toggle: AICR_VALIDATION_FLOOR_STRICT=1 promotes the recommended
+// performance phase from warn-only to required. Default OFF until #969
+// closes the data gap and Azure/OCI performance testbeds land.
+//
+// knownGaps allowlist: overlays tracked by #969 are listed below so the
+// contract can land independently of the data fix. Entries downgrade
+// failures to logs. Empty this map as #969 closes; the test fails any
+// NEW overlay added without the floor because it is not allowlisted.
+
+package recipe
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+	"testing"
+)
+
+const strictEnvVar = "AICR_VALIDATION_FLOOR_STRICT"
+
+// knownGaps lists leaf overlays that fail the floor today and are tracked
+// under #969. Each entry downgrades an Errorf to a Logf prefixed with
+// "KNOWN GAP (#969):". Drain this map as #969 lands; delete the map and
+// gap handling once empty.
+var knownGaps = map[string]bool{
+	"gb200-oke-ubuntu-inference-dynamo":  true,
+	"gb200-oke-ubuntu-training-kubeflow": true,
+	"h100-kind-inference-dynamo":         true,
+	"h100-kind-training-kubeflow":        true,
+	"h100-kind-training-slurm":           true,
+	"rtx-pro-6000-lke-ubuntu-inference":  true,
+	"rtx-pro-6000-lke-ubuntu-training":   true,
+}
+
+// classification captures the inputs that drive the per-intent floor.
+type classification struct {
+	Intent   CriteriaIntentType
+	Service  CriteriaServiceType
+	Platform CriteriaPlatformType
+	IsKind   bool
+}
+
+// String renders a classification for failure messages.
+func (c classification) String() string {
+	return fmt.Sprintf("intent=%s service=%s platform=%s kind=%t",
+		c.Intent, c.Service, c.Platform, c.IsKind)
+}
+
+// requiresPerformance reports whether the per-intent floor recommends
+// the performance phase for this classification.
+func (c classification) requiresPerformance() bool {
+	if c.IsKind {
+		return false
+	}
+	if c.Intent == CriteriaIntentTraining {
+		return true
+	}
+	dynamoOrNIM := c.Platform == CriteriaPlatformDynamo || c.Platform == CriteriaPlatformNIM
+	return c.Intent == CriteriaIntentInference && dynamoOrNIM
+}
+
+// classifyOverlay derives the classification from resolved criteria.
+func classifyOverlay(criteria *Criteria) classification {
+	return classification{
+		Intent:   criteria.Intent,
+		Service:  criteria.Service,
+		Platform: criteria.Platform,
+		IsKind:   criteria.Service == CriteriaServiceKind,
+	}
+}
+
+// resolvedPhases returns the names of phases that are set on v.
+func resolvedPhases(v *ValidationConfig) []string {
+	if v == nil {
+		return nil
+	}
+	var out []string
+	if v.Readiness != nil {
+		out = append(out, "readiness")
+	}
+	if v.Deployment != nil {
+		out = append(out, "deployment")
+	}
+	if v.Performance != nil {
+		out = append(out, "performance")
+	}
+	if v.Conformance != nil {
+		out = append(out, "conformance")
+	}
+	return out
+}
+
+// resolveValidation walks the base: chain for the named recipe and
+// returns the merged ValidationConfig using the same Merge semantics
+// the production resolver uses.
+func resolveValidation(s *MetadataStore, name string) (*ValidationConfig, error) {
+	chain, err := s.resolveInheritanceChain(name)
+	if err != nil {
+		return nil, err
+	}
+	merged := &RecipeMetadataSpec{}
+	for _, recipe := range chain {
+		merged.Merge(&recipe.Spec)
+	}
+	return merged.Validation, nil
+}
+
+// discoverLeafOverlays returns the names of every user-selectable leaf
+// overlay. A leaf has criteria, is not referenced as another overlay's
+// spec.base, and is not a criteria-wildcard fragment (one whose
+// criteria.intent or criteria.service is "any"). Wildcard overlays are
+// cross-cutting fragments applied via the resolver's wildcard-match
+// path — see docs/contributor/data.md#criteria-wildcard-overlays — not
+// user-selectable entry points subject to the per-intent floor.
+// Sorted for deterministic test output.
+func discoverLeafOverlays(s *MetadataStore) []string {
+	referencedAsBase := make(map[string]bool)
+	for _, overlay := range s.Overlays {
+		if overlay.Spec.Base != "" {
+			referencedAsBase[overlay.Spec.Base] = true
+		}
+	}
+	var leaves []string
+	for name, overlay := range s.Overlays {
+		if referencedAsBase[name] {
+			continue
+		}
+		c := overlay.Spec.Criteria
+		if c == nil {
+			continue
+		}
+		if c.Intent == CriteriaIntentAny || c.Service == CriteriaServiceAny {
+			continue
+		}
+		leaves = append(leaves, name)
+	}
+	sort.Strings(leaves)
+	return leaves
+}
+
+// TestOverlayValidationPhaseFloor asserts every leaf overlay's resolved
+// validation block contains the per-intent required phases. See file
+// header for the floor matrix and the strict-mode toggle.
+func TestOverlayValidationPhaseFloor(t *testing.T) {
+	ctx := context.Background()
+	store, err := loadMetadataStore(ctx)
+	if err != nil {
+		t.Fatalf("loadMetadataStore: %v", err)
+	}
+
+	strict := os.Getenv(strictEnvVar) == "1"
+	t.Logf("strict mode (%s=1): %t", strictEnvVar, strict)
+	t.Logf("knownGaps allowlist size (#969): %d", len(knownGaps))
+
+	leaves := discoverLeafOverlays(store)
+	t.Logf("leaf overlays discovered: %d", len(leaves))
+
+	for _, name := range leaves {
+		t.Run(name, func(t *testing.T) {
+			overlay := store.Overlays[name]
+			class := classifyOverlay(overlay.Spec.Criteria)
+
+			validation, err := resolveValidation(store, name)
+			if err != nil {
+				t.Fatalf("resolveValidation: %v", err)
+			}
+			phases := resolvedPhases(validation)
+
+			report := func(severity, kind, phase string) string {
+				return fmt.Sprintf(
+					"%s overlay %q [%s]\n  resolved phases: %s\n  missing %s: %s",
+					severity, name, class,
+					strings.Join(phases, ", "),
+					kind, phase,
+				)
+			}
+
+			// fail records a missing required phase. Overlays in knownGaps
+			// are downgraded to logs so the contract can land before #969
+			// finishes closing the data gap.
+			fail := func(phase string) {
+				msg := report("FAIL", "required", phase)
+				if knownGaps[name] {
+					t.Logf("KNOWN GAP (#969): %s", msg)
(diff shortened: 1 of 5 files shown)


pkg/recipe/validation_phase_floor_test.go
```diff
@@ -0,0 +1,323 @@
...
+// gap handling once empty.
+var knownGaps = map[string]bool{
+	"gb200-oke-ubuntu-inference-dynamo":  true,
+	"gb200-oke-ubuntu-training-kubeflow": true,
+	"h100-kind-inference-dynamo":         true,
+	"h100-kind-training-kubeflow":        true,
+	"h100-kind-training-slurm":           true,
+	"rtx-pro-6000-lke-ubuntu-inference":  true,
+	"rtx-pro-6000-lke-ubuntu-training":   true,
+}
```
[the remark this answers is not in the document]
mchmarny: Implemented in c06e5a38. Added a `triggeredKnownGaps` map plus a post-loop check that fails if any entry didn't downgrade a failure. Verified by injecting a bogus entry — fails with the expected message naming the entry to remove.

Note: only one allowlist exists in this file (`knownGaps`); the suggestion's reference to allowlists at lines 210-220 and 240-242 doesn't match the code — those ranges are the per-overlay fail closure and the performance phase logic, not separate allowlists.

pkg/recipe/validation_phase_floor_test.go
```diff
@@ -0,0 +1,323 @@
...
+		t.Fatalf("loadMetadataStore: %v", err)
+	}
+
+	strict := os.Getenv(strictEnvVar) == "1"
+	t.Logf("strict mode (%s=1): %t", strictEnvVar, strict)
+	t.Logf("knownGaps allowlist size (#969): %d", len(knownGaps))
+
+	leaves := discoverLeafOverlays(store)
+	t.Logf("leaf overlays discovered: %d", len(leaves))
+
```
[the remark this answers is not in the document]
mchmarny: Implemented in c06e5a38 with `t.Fatal` (no formatting) and a message pointing the reader at `discoverLeafOverlays` and `recipes/overlays/` so a future regression is obvious.

pkg/recipe/validation_phase_floor_test.go
```diff
@@ -0,0 +1,349 @@
...
+// returns the merged ValidationConfig using the same Merge semantics
+// the production resolver uses.
+func resolveValidation(s *MetadataStore, name string) (*ValidationConfig, error) {
+	chain, err := s.resolveInheritanceChain(name)
+	if err != nil {
+		return nil, err
+	}
+	merged := &RecipeMetadataSpec{}
+	for _, recipe := range chain {
+		merged.Merge(&recipe.Spec)
```
yuanchen8911: This merge mutates the shared cached metadata store because `Merge` aliases `other.Validation` when the destination has no validation, then later phase writes go through that alias (see PR summary, finding 1).

The result is that this test is order-dependent. In a full run, `h100-gke-cos-inference-dynamo` resolves alphabetically before `h100-gke-cos-training-kubeflow` and plants `Deployment` onto the cached `gke-cos` validation. Running only `go test -run 'TestOverlayValidationPhaseFloor/h100-gke-cos-training-kubeflow'` shows the unpolluted result: `performance, conformance` — missing `deployment`. The overlay is not in `knownGaps`, so today the test silently passes for an overlay that actually fails the floor.

Fixing `Merge` in production is the best fix; otherwise this helper needs to deep-copy validation before merging so the gate does not pass due to polluted shared state.

pkg/recipe/validation_phase_floor_test.go
```diff
@@ -0,0 +1,349 @@
...
+	}
+	return merged.Validation, nil
```
yuanchen8911: This helper does not actually follow the production resolution path. It walks only the `spec.base` chain, while production uses `FindMatchingOverlays` (`metadata_store.go:316-342`), per-query maximal-leaf filtering, inheritance-chain merge, and then `mergeMixins` (`metadata_store.go:454`).

That matters for criteria-wildcard overlays. For example, production applies `gb200-any-training` (which carries `validation.performance` in `recipes/overlays/gb200-any-training.yaml`) to `gb200-oke-ubuntu-training-kubeflow` and contributes `performance` to the resolved validation; this helper's chain-only resolution never sees it. The header comment claiming this uses the same production merge semantics is therefore misleading, and some floor results can be different from what users actually get. Mixins are also skipped — fine today (no current mixin carries validation) but a divergence to track.

pkg/recipe/validation_phase_floor_test.go
```diff
@@ -0,0 +1,349 @@
...
+	referencedAsBase := make(map[string]bool)
+	for _, overlay := range s.Overlays {
+		if overlay.Spec.Base != "" {
+			referencedAsBase[overlay.Spec.Base] = true
+		}
+	}
+	var leaves []string
+	for name, overlay := range s.Overlays {
+		if referencedAsBase[name] {
+			continue
```
yuanchen8911: This discovers leaves globally, but production filters ancestors only among overlays that match the current query in `filterToMaximalLeaves` (`metadata_store.go:347-381`).

That means concrete intermediate overlays can be user-selectable but omitted by this test. For example, a query matching `service=gke accelerator=h100 intent=training` (no platform) will not match `h100-gke-cos-training-kubeflow` (which requires `platform=kubeflow`), so production returns `h100-gke-cos-training` as the maximal leaf — but this global `referencedAsBase` filter skips it entirely. Same story for `h100-eks-training`, `gb200-eks-training`, `h100-aks-training`, `eks-training`, `gke-cos-training`, etc. The PR body's "27 of 41 GPU overlays" framing vs the 17 leaves the test discovers points at this gap.

Please mirror production selection with `FindMatchingOverlays`-style discovery, or otherwise gate every overlay that production can return as a maximal candidate for some query. Note this will likely surface more existing gaps to add to `knownGaps`.

pkg/recipe/validation_phase_floor_test.go
```diff
@@ -0,0 +1,349 @@
...
+		}
+		c := overlay.Spec.Criteria
+		if c == nil {
+			continue
+		}
+		if c.Intent == CriteriaIntentAny || c.Service == CriteriaServiceAny {
+			continue
```
yuanchen8911, line 177: This wildcard-fragment filter is narrower than the criteria matcher. `MatchesCriteriaField` (in `criteria.go`) treats both `"any"` and empty string as wildcard across service, accelerator, intent, OS, and platform; this check only handles explicit `intent: any` and `service: any`.

There is no current overlay with `accelerator: any`, `os: any`, or `platform: any` that changes the outcome, so this is latent. Still, if this helper continues to define "non-user-selectable wildcard fragments," it should use the same wildcard semantics as criteria matching rather than a partial field check. (If the leaf-discovery comment above is taken — using `FindMatchingOverlays`-style discovery — this filter goes away entirely.)

pkg/recipe/validation_phase_floor_test.go
```diff
@@ -0,0 +1,349 @@
...
+			// are downgraded to logs so the contract can land before #969
+			// finishes closing the data gap.
+			fail := func(phase string) {
+				msg := report("FAIL", "required", phase)
+				if knownGaps[name] {
+					triggeredKnownGaps[name] = true
+					t.Logf("KNOWN GAP (#969): %s", msg)
+					return
+				}
+				t.Error(msg)
```
yuanchen8911, line 258: This allowlist is keyed only by overlay name, so it suppresses every missing required phase for that overlay, not just the known gap.

Concrete impact: `gb200-oke-ubuntu-inference-dynamo` currently has `conformance` and lacks `deployment`. If a future PR adds `deployment` but regresses `conformance`, the floor test would still downgrade the new failure to a log — the regression is invisible. Please make the allowlist phase-specific, e.g. `map[string]map[string]bool` keyed by `(overlay, phase)`, so only the known missing phase is downgraded. Pairs well with the wider-leaf-discovery fix above: a tighter allowlist is more useful when the leaf set expands.

yuanchen8911 (commented):
Cross-review by Claude Code + Codex + CodeRabbit. Three reviewers plus an integration-analysis pass converged on the same set of findings; cross-review verified each one against the PR head independently. Inline comments below cover the test-file findings. The headline issue is a production bug exposed by this test — since `pkg/recipe/metadata.go` is outside the diff, I'm raising it here.

## 1. Production bug: `RecipeMetadataSpec.Merge` aliases `other.Validation` and writes phase pointers through that alias

[`pkg/recipe/metadata.go#L460-L477`](https://github.com/NVIDIA/aicr/blob/c06e5a38306dcc5734f817a631c055ea4c55afde/pkg/recipe/metadata.go#L460-L477)

When `s.Validation == nil`, the function assigns `s.Validation = other.Validation` (pointer alias), then on later iterations the per-phase writes (`s.Validation.Deployment = other.Validation.Deployment`, etc.) go through that alias. Because the metadata store is cached and reused across recipe builds (`metadata_store.go:65`, `BuildRecipeResult` → `mergeOverlayChains` at `metadata_store.go:674`), resolving one recipe mutates an ancestor overlay for every later resolution.

Reproduction: building `h100-gke-cos-inference-dynamo` mutates cached `gke-cos` validation from `conformance` to `deployment, performance, conformance`. That pollution masks the missing `deployment` phase on `h100-gke-cos-training-kubeflow`, which is not in `knownGaps`. Running just `go test -run 'TestOverlayValidationPhaseFloor/h100-gke-cos-training-kubeflow'` shows the unpolluted result: `performance, conformance` — failing the floor.

Suggested fix: in `Merge`, deep-copy `other.Validation` (or copy each non-nil phase pointer into a newly allocated `ValidationConfig`) instead of aliasing. Fixing this in production removes the test order dependence as a side effect.

## Severity ordering

- **Blocker:** production `Merge` aliasing bug (above)
- **Should-fix in this PR:** chain-only resolver (L128), global leaf discovery (L148), per-overlay `knownGaps` granularity (L222)
- **Latent:** wildcard filter narrowness (L164)

yuanchen8911 (changes requested):
There are some issues. PTAL

pkg/recipe/validation_phase_floor_test.go
mchmarny: Verified by running a single subtest — the result differed from the full run, confirming the alias chain you described.

Fixed in 5d4f4cad. Added `cloneValidationConfig`/`cloneValidationPhase` and clone at both `Merge` assignment branches plus `initBaseMergedSpec`. Regression test in `metadata_test.go` (`merge does not alias source validation`) detects the leak: without the fix, the second merge plants `Deployment` into the first source's `ValidationConfig` and the assertion fires.

pkg/recipe/validation_phase_floor_test.go
mchmarny: Agreed. Switched to `store.BuildRecipeResult` in 19c2a8db — same code path the CLI and API use, so wildcard overlays and mixins now contribute to the resolved validation just as they do in production. Updated the file header to drop the misleading 'same Merge semantics' phrasing.

pkg/recipe/validation_phase_floor_test.go
mchmarny: Agreed. Replaced `discoverLeafOverlays` with `enumerateGateableOverlays` in 19c2a8db: every overlay with concrete criteria (minus wildcard fragments) is now gated by running `BuildRecipeResult` with the overlay's own criteria as the query. Per-query maximal-leaf filtering is then handled by production code.

Gateable set went from 17 → 56 overlays. As you predicted, this surfaced more gaps — the new allowlist has 39 (overlay, phase) pairs, including the concrete intermediates you named (`eks-training`, `h100-gke-cos-training`, `h100-eks-training`, etc.). Yuan's #991 already drained two during the rebase.

pkg/recipe/validation_phase_floor_test.go
mchmarny: With the switch to `BuildRecipeResult` (comment above), production resolution handles wildcard matching, so this filter only needs to identify queries that are not user-facing — `intent: any` and `service: any` are the only ones today, so the filter stays minimal. If `accelerator: any` / `os: any` / `platform: any` wildcards ever ship as standalone overlays, that's the place to widen it.

pkg/recipe/validation_phase_floor_test.go
mchmarny: Fixed in 19c2a8db. `knownGaps` is now `map[string]map[string]bool` keyed by `(overlay, phase)`. The staleness check tracks per-pair triggering, so a regression in a previously-passing phase fires immediately even when the overlay has a different phase allowlisted.

yuanchen8911 (approved):
All 6 review points are addressed with concrete code changes that go beyond the minimum.

PR #992, opened 2026-05-20
fix(recipe): deep-copy validation config during merge

### LINKED ISSUE #984: fix(recipe): deep-clone ValidationConfig phases in Merge to prevent cache corruption
Part of #983.

## Summary

`RecipeMetadataSpec.Merge` (in `pkg/recipe/metadata.go`, around line 490) aliases overlay validation phases into the merge result by pointer:

```go
s.Validation.Readiness  = other.Validation.Readiness
s.Validation.Deployment = other.Validation.Deployment
s.Validation.Performance = other.Validation.Performance
s.Validation.Conformance = other.Validation.Conformance
```

`other.Validation` lives inside the cached `MetadataStore` that the package's `sync.Once`-backed loader populates once per process. Every call to `BuildRecipeResult` for an overlay that contributes a validation phase aliases that same cached pointer into its result. Any caller mutating `result.Validation.Readiness.Constraints[i]`, `result.Validation.Performance.NodeSelection.ExcludeNodes`, or any other nested slice/map under a phase corrupts the cached overlay for every other concurrent caller of the same overlay.

## Reproduction (conceptual)

The hazard is dormant because no current in-tree caller mutates a result's validation in-place. It surfaces immediately if:

1. Two concurrent `BuildRecipeResult` calls produce results that both include the same overlay's `Validation.Readiness`.
2. One caller appends to `result.Validation.Readiness.Constraints`.
3. The second caller observes the appended constraint in its own result.

A targeted unit test that explicitly mutates a returned phase and then re-loads the same overlay via the store will catch this.

## Why it matters now

This is documented as an anti-pattern in `.claude/CLAUDE.md`:

> Deep-copy helper that recurses into maps but copies `[]any` by reference — recurse into both `map[string]any` and `[]any`; scalars fall through the default branch by value. Slice aliasing leaks mutations across overlay merges.

The cache is currently process-local and the bug is invisible. As soon as we expose more API surface that returns structured validation data to in-process consumers (see the parent), the assumption that nobody mutates phase data breaks silently and across goroutines. Fix it now while the contract is still clean.

## Proposed change

Introduce three private clone helpers in `pkg/recipe/metadata.go`:

```go
func cloneValidationConfig(in *ValidationConfig) *ValidationConfig
func cloneValidationPhase(in *ValidationPhase) *ValidationPhase
func cloneNodeSelection(in *NodeSelection) *NodeSelection
```

Each must:

- Return nil when given nil (so they can stand in directly for pointer assignments).
- Deep-copy every slice (`Constraints`, `Checks`, `ExcludeNodes`) and map (`NodeSelection.Selector`) so the clone shares no backing storage with the input.
- Be updated alongside any future field added to `ValidationConfig`, `ValidationPhase`, or `NodeSelection`. Add a comment on each helper noting this contract.

`cloneNodeSelection` in particular must clone **both** `Selector` (map) and `ExcludeNodes` (slice). The most common form of this bug is a clone that copies one and forgets the other.

Rewrite the four overlay-precedence branches in `Merge` to call the clone helpers instead of aliasing.

## Acceptance

- New table-driven test in `pkg/recipe/metadata_test.go`:
  - For each of the four phases (Readiness/Deployment/Performance/Conformance), mutate the merged result's `Constraints`, `Checks`, and `NodeSelection.ExcludeNodes` / `Selector`, then re-load the underlying overlay via the store and assert it is unchanged.
  - Repeat with two concurrent `BuildRecipeResult` calls (use `-race`) to demonstrate independence.
- `make test -race ./pkg/recipe/...` clean.
- Grep `pkg/recipe/metadata.go` for `s.Validation.X = other.Validation.X` patterns and confirm none remain.

## Out of scope

- Cloning `ComponentRefs` and `Constraints` at the top level of `RecipeMetadataSpec` — those merge by name and don't have the aliasing problem.
- Refactoring the cache itself. That's covered by the per-provider isolation child issue.

## Dependencies

- None. This is standalone and the smallest of the four issues — recommended to land first.

kiwigitops:
## Summary
- clone validation configs/phases when merging recipe metadata instead of assigning overlay pointers directly
- deep-copy validation slices and node-selection maps/slices
- add regression coverage for readiness, deployment, performance, and conformance phases

Closes #984.

## Testing
- `git diff --check`

I could not run `go test ./pkg/recipe` locally because this Windows environment does not have `go` on PATH.


pkg/recipe/metadata.go
```diff
@@ -617,6 +617,44 @@ func mergeComponentRef(base, overlay ComponentRef) ComponentRef {
...
+func cloneValidationPhase(in *ValidationPhase) *ValidationPhase {
+	if in == nil {
+		return nil
+	}
+	out := *in
+	out.Constraints = append([]Constraint(nil), in.Constraints...)
+	out.Checks = append([]string(nil), in.Checks...)
+	out.NodeSelection = cloneNodeSelection(in.NodeSelection)
+	return &out
+}
```
mchmarny, line 641: Both `cloneValidationConfig` and `cloneValidationPhase` already exist on main since [`05b6ae37`](https://github.com/NVIDIA/aicr/commit/05b6ae37) (in `pkg/recipe/validation.go`, lines 69 and 84), which fixed #984 directly. Adding them here is a redeclaration — `go build ./pkg/recipe/...` on this branch fails:

```
pkg/recipe/validation.go:69:6: cloneValidationConfig redeclared in this block
    pkg/recipe/metadata.go:620:6: other declaration of cloneValidationConfig
pkg/recipe/validation.go:84:6: cloneValidationPhase redeclared in this block
    pkg/recipe/metadata.go:632:6: other declaration of cloneValidationPhase
```

The reason CI didn't catch this is that the `gate` workflow likely requires maintainer approval for first-time external contributors — the only checks that ran here were `Label PRs` and the skipped auto-merge job.

Please rebase onto main; once you do, you'll see the helpers (and the `s.Validation = cloneValidationConfig(...)` call sites in `Merge`) are already in place. Both helpers in this hunk should then be removed.

pkg/recipe/metadata.go
```diff
@@ -617,6 +617,44 @@ func mergeComponentRef(base, overlay ComponentRef) ComponentRef {
...
+	out := *in
+	if in.Selector != nil {
+		out.Selector = make(map[string]string, len(in.Selector))
+		for key, value := range in.Selector {
+			out.Selector[key] = value
+		}
+	}
+	out.ExcludeNodes = append([]string(nil), in.ExcludeNodes...)
+	return &out
+}
```
mchmarny, line 656: `cloneNodeSelection` is a new helper, but main's `cloneValidationPhase` ([`validation.go:84`](https://github.com/NVIDIA/aicr/blob/main/pkg/recipe/validation.go#L84)) already clones `NodeSelection` inline — both the `Selector` map and `ExcludeNodes` slice. Extracting it as a named function isn't a bad idea on its own, but it should live in `validation.go` next to its siblings, not `metadata.go`. After rebasing, if you still want this extraction, open it as a small refactor PR against the already-fixed code.

pkg/recipe/metadata_test.go
```diff
@@ -1012,6 +1012,72 @@ func TestMergeValidationConfig(t *testing.T) {
...
+					t.Error("checks should not alias overlay")
+				}
+				if mergedPhase.NodeSelection.Selector["accelerator"] != "h100" {
+					t.Error("node selector should not alias overlay")
+				}
+				if mergedPhase.NodeSelection.ExcludeNodes[0] != "node-a" {
+					t.Error("excluded nodes should not alias overlay")
+				}
+			})
+		}
```
mchmarny, line 1078: This is the salvageable part of the PR — the per-phase aliasing matrix (`readiness`/`deployment`/`performance`/`conformance`) is genuinely better coverage than what landed with `05b6ae37`, which only exercises the deployment+conformance paths in `t.Run("merge does not alias source validation", ...)` at the bottom of `TestMergeValidationConfig`. After dropping the duplicate helpers in `metadata.go`, this table-driven subtest is worth keeping. Consider trimming the focused PR to just this test addition.

mchmarny (changes requested):
Thanks for digging into #984. The branch was opened before the fix landed on main as [`05b6ae37`](https://github.com/NVIDIA/aicr/commit/05b6ae37), which already solves the same root cause: it adds `cloneValidationConfig` / `cloneValidationPhase` in `pkg/recipe/validation.go` and wires them into `Merge`. After a rebase onto main, the additions to `metadata.go` in this PR become duplicate declarations and the package won't compile (verified locally with `go build ./pkg/recipe/...`).

The PR description mentioned `go` wasn't on PATH in the Windows environment — that's why this slipped through; CI didn't run either because the `gate` workflow appears to require maintainer approval for new external contributors.

Recommendation: rebase onto main, remove the helper additions in `metadata.go` (they're already in `validation.go`), and keep only the per-phase aliasing test in `metadata_test.go` — that subtest is the one genuinely useful improvement here, since current main only exercises two of the four phases. Inline comments mark each location.

If a rebase ends up trivial enough that nothing remains beyond the test, closing this PR in favor of a small follow-up scoped to just the test addition would also work.

PR #997, opened 2026-05-21
feat(slinky-slurm): add h100-gke-cos leaf and inline platform refs

faganihajizada:
## Summary

Adds the first-class `h100-gke-cos-training-slurm` leaf and refactors the existing EKS/Kind slurm leaves to inline the three Slinky components per leaf (mirroring the `dynamo-platform` pattern), unblocking leaf-level GPU GRES tuning that the old `platform-slurm` / `platform-slurm-cluster` mixin shape made impractical.

## Motivation / Context

The original `platform-slurm` / `platform-slurm-cluster` mixins contributed `slinky-slurm-operator-crds`, `slinky-slurm-operator`, and `slinky-slurm` to every slurm leaf. Because mixins are merged *after* the inheritance chain (`pkg/recipe/metadata_store.go:mergeMixins`) and the mixin's own componentRefs set identity fields (`type`, `valuesFile`, `dependencyRefs`), any leaf that pre-declared the same component name to attach `overrides` would cause `mixinComponentRefSafeForMerge` to reject the mixin's identity-field entry as a conflict against the chain-existing component.

In practice this meant the only way to push leaf-specific GPU GRES (`extraConfMap.Gres=gpu:h100:8` and `slurmd.resources.limits.nvidia.com/gpu: 8`) into `nodesets.slinky` was at install time via `--set slinkyslurm:...`, which made the leaf incomplete: `srun --gres=gpu:N` saw `Gres=(null)` from any straight `aicr bundle`.

Switching to the same inline pattern used by `dynamo-platform` across `*-inference-dynamo` leaves moves Slinky off the mixin merge path and onto the standard inheritance/overlay path, so each leaf carries its own hardware-specific tuning as part of the recipe itself.

This change was developed and rehearsed end-to-end on the `aicr-eidosx` GKE cluster: full 17-component bundle deploy on 2x H100 nodes, `sinfo` showing both NodeSet pods idle with `gpu:h100:8`, and `srun -N2 --gres=gpu:8 nvidia-smi -L` returning 16 unique H100 UUIDs across both hosts.

Fixes: N/A
Related: NKX-9586, #948 (initial slinky-slurm chart + EKS/Kind leaves)

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: recipes (`recipes/overlays`, `recipes/mixins`, `recipes/registry.yaml`)

## Implementation Notes

**Why inline instead of relaxing the mixin guard?** The `mixinComponentRefSafeForMerge` guard exists for a reason (ADR-005's "no silent chart identity override" mitigation): mixin-contributed components must be replaceable wholesale, not partially merged, to keep the mixin contract simple. The `dynamo-platform` family already established the precedent of declaring platform components inline per leaf — this PR brings slurm in line with that pattern rather than weakening the guard.

**Why omit `source` / `version` / `valuesFile` on the slinky `componentRefs`?** Unlike `dynamo-platform`, the Slinky charts are not declared in `base.yaml`. They live only in `registry.yaml` with `defaultRepository` / `defaultVersion` / `defaultChart`, and `ApplyRegistryDefaults` (`pkg/recipe/metadata.go:156-184`) populates them at recipe-resolve time. The registry comment on `slinky-slurm` explicitly documents `defaultVersion` as the single source of truth (one-line bumps instead of touching every leaf). This shape diverges from dynamo's inlining but is engine-supported and reduces leaf duplication.

**GRES wiring requires two fields, not one.** The chart does not derive `Gres=` in `slurm.conf` from pod resource limits, so leaves must declare GRES in both places: `nodesets.slinky.extraConfMap.Gres` (tells slurmctld which GPUs slurmd has) and `nodesets.slinky.slurmd.resources.limits.nvidia.com/gpu` (reserves the GPUs from the device plugin so `gres.conf`'s `AutoDetect=nvidia` finds them). Both are documented inline in the GKE leaf with a cross-reference to the mars-prod reference values.

**API reference docs gap.** While updating component docs I noticed `slinky-slurm` (the cluster instance) was missing from `docs/user/api-reference.md`'s bundler table — a pre-existing gap from when it was opt-in via the deleted `platform-slurm-cluster` mixin. Added in the same refactor commit so the docs match the inline reality.

**Out of scope** (logged separately): multi-node NCCL via Pyxis/Enroot requires swapping to `ghcr.io/slinkyproject/slurmd-pyxis` and configuring `plugstack.conf`; not attempted here.

## Testing

```bash
# Recipe + ordering tests
go test ./pkg/recipe/... -count=1 -short
# ok  github.com/NVIDIA/aicr/pkg/recipe        1.188s
# ok  github.com/NVIDIA/aicr/pkg/recipe/oskind 0.389s

# Full short test suite (all pkg/...)
go test ./pkg/... -count=1 -short
# All 30+ packages: ok

# Lint sweep (lint-go skipped: local golangci-lint built with go1.25,
# module targets go 1.26.3; CI runs the pinned version)
make lint-yaml license check-docs-filenames check-docs-mdx \
     bom-pinning-check check-agents-sync bom-check
# All pass
```

**End-to-end rehearsal on `aicr-eidosx` (GKE, 2x H100):**

```text
$ sinfo
PARTITION  AVAIL  TIMELIMIT  NODES  STATE NODELIST
all*          up   infinite      2   idle slinky-[0-1]

$ srun -N2 --gres=gpu:8 nvidia-smi -L | sort -u | wc -l
16   # 16 unique H100 UUIDs across both slurmd pods
```

## Risk Assessment

- [x] **Low** — Isolated to slurm leaves and the platform-slurm mixins; cohabitant components in mixed bundles are untouched. Rehearsed in a live cluster against the full 17-component AICR bundle; cohabitant Helm releases received a revision bump from `helm upgrade --install` but no manifest churn.

**Rollout notes:** No data migration. The deleted mixins were always paired (no leaf ever used `platform-slurm` without `platform-slurm-cluster`), so removing them does not regress any composition that was actually used. Any out-of-tree leaf still referencing the mixins will fail recipe validation with a clear error and needs a one-time port to the inline shape — see the EKS leaf for the template.

## Checklist

- [x] Linter passes (`make lint`) — except `lint-go` (local Go-version build mismatch on `golangci-lint`; CI runs pinned version)
- [x] I added/updated tests for new functionality (`deployment_order_guard_test.go`)
- [x] I updated docs if user-facing behavior changed (`recipe-development.md`, `component-catalog.md`, `api-reference.md`)
- [x] Changes follow existing patterns in the codebase (mirrors `dynamo-platform` inline pattern)
- [ ] Commits are cryptographically signed (`git commit -S`) — verify before pushing


recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+  # K8s-native NCCL bandwidth check is dropped on slurm leaves: on a
+  # Slinky cluster the equivalent validation is a slurm-launched job
+  # (`srun nccl-tests/all_reduce_perf`) that exercises slurmd + the
+  # GPUDirect TCPXO plugin already deployed by the parent leaf via
+  # gke-nccl-tcpxo. Conformance checks are inherited unchanged.
+  validation:
+    performance:
+      checks: []
+      constraints: []
```
ArangoGutierrez: Consistency question: the first sentence ("K8s-native NCCL bandwidth check is dropped on slurm leaves") reads as a general rule, but the refactored EKS slurm leaf (`recipes/overlays/h100-eks-ubuntu-training-slurm.yaml`) doesn't apply it — it still inherits `performance.checks: [nccl-all-reduce-bw]` from the EKS training chain.

Two options:
1. If the rule generalizes to any Slinky cluster (`srun` works equally on either platform), also override `validation.performance` in the EKS slurm leaf for consistency.
2. If the GKE-specific TCPXO path is the actual reason, tighten this comment to say so — e.g., "the K8s-native bandwidth check measures the wrong path on GKE because TCPXO traffic only flows through slurmd-launched jobs."

Minor nit on the same block: `Conformance checks are inherited unchanged` — since both `deployment.checks` and `conformance.checks` are inherited and only `performance.checks` is overridden, `Deployment and conformance checks are inherited unchanged` would be more precise.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -0,0 +1,101 @@
...
+        - slinky-slurm-operator-crds
+      overrides:
+        nodesets:
+          slinky:
+            extraConfMap:
+              Gres: "gpu:h100:8"
+            slurmd:
+              resources:
+                limits:
+                  nvidia.com/gpu: 8
```
ArangoGutierrez, line 90: Non-blocking observation: this `overrides` block is byte-identical to the one in `h100-eks-ubuntu-training-slurm.yaml`. The PR description correctly explains why the mixin path won't work today (identity-field guard), so duplication is the right call. If a third H100 slurm leaf (AKS/OKE) lands later, it may be worth revisiting whether a per-accelerator GRES contribution mechanism — separate from the mixin merge path — is warranted.

pkg/recipe/deployment_order_guard_test.go
```diff
@@ -198,6 +198,29 @@ func TestDeploymentOrderGuards(t *testing.T) {
...
+				"slinky-slurm":          {"slinky-slurm-operator", "slinky-slurm-operator-crds"},
+			},
+			requiredOrdering: [][2]string{
+				{"cert-manager", "slinky-slurm-operator"},
+				{"slinky-slurm-operator-crds", "slinky-slurm-operator"},
+				{"slinky-slurm-operator", "slinky-slurm"},
+				{"slinky-slurm-operator-crds", "slinky-slurm"},
+				{"gpu-operator", "nvsentinel"},
+			},
+		},
```
ArangoGutierrez, line 223: Non-blocking suggestion (pre-existing gap): `TestNFDTopologyUpdater_OverlayCoverage` in `pkg/recipe/metadata_test.go` enumerates every GPU overlay variant (training-kubeflow, inference-dynamo, inference-nim, etc.) but doesn't include any slurm leaves. Its docstring claims to guard "every GPU overlay rooted at a real-cluster platform base," and its stated purpose is exactly to catch overlays that accidentally lose NRT publishing.

Since this PR adds a new slurm leaf and refactors the others, it's a natural moment to extend coverage. Could be a follow-up if scope is a concern.

ArangoGutierrez (dismissed):
Approving — the refactor is well-motivated and well-documented.

**Summary:** The inline shape mirrors the established `dynamo-platform` precedent, the PR description traces the engine constraint (`mixinComponentRefSafeForMerge` identity-field guard) precisely, and the end-to-end evidence on `aicr-eidosx` (sinfo + 16 unique H100 UUIDs across both slurmd pods) is concrete. Nice catch on adding `slinky-slurm` to the `docs/user/api-reference.md` bundler table — that was a pre-existing gap from when it was opt-in via the deleted cluster mixin.

**Verified locally at HEAD `e51b4b16`:**
- `TestDeploymentOrderGuards` passes all 12 cases including the new `h100-gke-cos-training-slurm`.
- `go test ./pkg/recipe/... -short` green.
- All function references in the PR (`mergeMixins`, `mixinComponentRefSafeForMerge`, `ApplyRegistryDefaults`) exist at the cited locations.
- No orphaned consumers of the deleted mixins outside of stale planning docs in `docs/superpowers/`.

One consistency question on the EKS vs GKE NCCL validation handling (inline), plus a non-blocking suggestion about extending `TestNFDTopologyUpdater_OverlayCoverage` to cover slurm leaves (pre-existing gap, not introduced by this PR).

recipes/overlays/h100-gke-cos-training-slurm.yaml
faganihajizada: Good catch, went with option 1 (the rule generalizes). Both EKS and GKE slurm leaves now override `validation.performance.checks: []` with the same rationale (the K8s-native check launches a Pod past slurmd, so it bypasses the fabric integration entirely; the real signal is `srun nccl-tests/all_reduce_perf` going through slurmd + TCPXO on GKE / EFA on EKS).

Also tightened the comment per your nit: "Deployment and conformance checks are inherited unchanged."

Fix in `2d6614c4`.

recipes/overlays/h100-gke-cos-training-slurm.yaml
faganihajizada: Agree on the future direction. A per-accelerator GRES contribution mechanism — `accelerator-h100` mixin that contributes only `Overrides` to an existing `slinky-slurm` componentRef without tripping the identity-field guard — would be the cleanest fix once we have a third H100 slurm leaf to justify it. Keeping that out of scope for this PR.

pkg/recipe/deployment_order_guard_test.go
faganihajizada: Done: extended `TestNFDTopologyUpdater_OverlayCoverage` to cover all three slurm leaves in `24093c09`. EKS+GKE assert TU is on (inherited from `h100-eks-ubuntu-training` / `h100-gke-cos-training`), kind asserts TU stays off. Local run is green.

docs/integrator/recipe-development.md
```diff
@@ -115,7 +115,15 @@ spec:
 
 Mixins use `kind: RecipeMixin` and carry only `constraints` and `componentRefs`. They live in `recipes/mixins/` and are applied after inheritance chain merging. See [Data Architecture](../contributor/data.md#mixin-composition) for details.
 
-A platform may split into multiple mixins when parts of the stack are independently opt-in. For example, `--platform slurm` resolves through two mixins: `platform-slurm` always contributes the SchedMD Slinky operator and CRDs, and `platform-slurm-cluster` is opt-in for the Slinky-managed Slurm cluster instance (Controller / LoginSet / NodeSet / RestApi). A leaf that wants operator-only composes just `platform-slurm`; a leaf that wants the cluster too composes both — see `recipes/overlays/h100-eks-ubuntu-training-slurm.yaml` for the latter.
+A platform's full component stack is declared inline per leaf overlay rather than via a platform mixin. This lets each leaf carry its own hardware-specific tuning — most commonly GPU GRES strings and accelerator resource limits — without going through the mixin merge path.
```
mchmarny: This generalization is too broad — `platform-kubeflow` and `platform-inference` are still active mixins (used by `h100-eks-ubuntu-training-kubeflow.yaml`, `eks-inference.yaml`, etc.), and the example immediately above on this page (line ~107) shows `platform-kubeflow` being composed via `spec.mixins`. As written, the new paragraph contradicts that example and the actual state of `recipes/mixins/`.

Suggest scoping the claim to slurm (and dynamo, if you want to call it out as the precedent), e.g.:

> Some platforms declare their full component stack inline per leaf overlay rather than via a platform mixin — most notably `--platform slurm` and `--platform dynamo`, where each leaf carries hardware-specific tuning (GPU GRES strings, accelerator resource limits) that the mixin merge path can't represent cleanly. Other platforms (`--platform kubeflow`, `--platform inference`) still use mixins where leaf-specific tuning isn't required.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -0,0 +1,104 @@
...
+    intent: training
+    platform: slurm
+
+  # Unlike the EKS slurm leaf, no os-ubuntu / os-cos mixin is needed:
+  # gke-cos.yaml already disables GPU driver installation and pins
+  # DRA / nodewright paths for COS's read-only rootfs.
+
+  constraints:
+    - name: K8s.server.version
+      value: ">= 1.32"
```
mchmarny: nit: this duplicates the parent `h100-gke-cos-training.yaml`'s `K8s.server.version: ">= 1.32"`. Same-name constraints get overridden during merge so it's harmless, but if it's not strengthening the floor it's noise. The EKS leaf bumps to `>= 1.32.4` which justifies the restatement; here either drop it or tighten it.

recipes/overlays/h100-eks-ubuntu-training-slurm.yaml
```diff
@@ -33,18 +34,69 @@ spec:
...
+  #      AutoDetect=nvidia finds nothing. `requests` is omitted: Kubernetes
+  #      auto-mirrors requests=limits for extended resources, matching the
+  #      mars-prod reference (manifests/clusters/mars-prod/
+  #      mars-dgxc-k8s-tgr-bwi-prd1/components/slurm/values.yaml).
   # Accelerated nodeSelector/tolerations on slurmd are injected via the
   # registry's nodesets.slinky.podSpec.{nodeSelector,tolerations} paths.
-  componentRefs: []
+  componentRefs:
+    - name: slinky-slurm-operator-crds
+      type: Helm
```
mchmarny, line 64: The double-declaration commentary (extraConfMap.Gres + nvidia.com/gpu limit, with the explanation of why both are needed) is exactly the kind of context that future-you will thank present-you for. Nice.

Two micro-suggestions to consider — neither blocking:

1. The `mars-prod` filesystem path (`manifests/clusters/mars-prod/mars-dgxc-k8s-tgr-bwi-prd1/...`) is from an internal NVIDIA repo and will be opaque to anyone reading this in the public OSS tree. Either qualify with "(internal NVIDIA reference cluster)" or drop the parenthetical — the rationale (requests auto-mirror to limits for extended resources) stands on its own.
2. The same ~20-line GRES rationale block is now copy-pasted verbatim in the GKE and EKS slurm leaves. If a future slurm leaf changes the wording in one and forgets the other they'll drift. Acceptable for two leaves; worth keeping an eye on if a third lands.

mchmarny (commented):
Clean refactor — moving slurm off the mixin merge path onto the same inline shape as `dynamo-platform` is the right call, the rationale in the description and YAML comments is thorough, and the live rehearsal (`sinfo` + `srun -N2 nvidia-smi -L` returning 16 H100 UUIDs) is solid evidence the GRES wiring actually works end-to-end. CI is green across all 142 checks.

Three findings, none blocking:

- **Medium** — `docs/integrator/recipe-development.md` line 118 overgeneralizes: claims "a platform's full component stack is declared inline … rather than via a platform mixin," but `platform-kubeflow` / `platform-inference` are still active mixins and the example just above on the same page composes `platform-kubeflow` via `spec.mixins`. Scope the claim to slurm (and dynamo, if calling it out as precedent).
- **Nit** — the GKE leaf's `K8s.server.version: ">= 1.32"` constraint duplicates the parent. The EKS leaf's restatement is justified because it tightens to `>= 1.32.4`; here it's noise.
- **Nit** — the ~20-line GRES rationale block is now duplicated verbatim across the EKS and GKE slurm leaves, and one of the references (`manifests/clusters/mars-prod/...`) points at an internal NVIDIA repo that will read as opaque in the OSS tree.

Nothing here blocks merge — the medium finding is a doc accuracy issue, not a code defect.

docs/integrator/recipe-development.md
faganihajizada: Fixed in 7782ff74. Scoped the claim to slurm and dynamo, and explicitly preserved that kubeflow/inference still use `platform-kubeflow` / `platform-inference` mixins (referencing the example two paragraphs up). Wording is now consistent with the actual state of `recipes/mixins/`.

recipes/overlays/h100-gke-cos-training-slurm.yaml
faganihajizada: Dropped the restatement in 7782ff74. Replaced with a short comment noting the floor is inherited from `h100-gke-cos-training.yaml` (cf. the EKS leaf, which restates `>= 1.32.4` to tighten its parent). Cleaner than leaving merge-noise in the constraint list.

recipes/overlays/h100-eks-ubuntu-training-slurm.yaml
faganihajizada: Dropped the `mars-prod` parenthetical from both EKS and GKE slurm leaves in 7782ff74. Agreed it reads as opaque in the OSS tree; the auto-mirror rationale stands on its own. On the duplication: keeping both copies for now (matches your "acceptable for two leaves" call); if a third slurm leaf lands we can extract the GRES rationale into a shared comment fragment or a leaf-template.

faganihajizada:
@mchmarny could you please take a look? thanks!

mchmarny (approved):

PR #1017, opened 2026-05-24
fix(bundler): replace placeholder GitRepository with ArtifactGenerator for Flux OCI

### LINKED ISSUE #964: bundler(flux): local-chart HelmReleases unreachable under OCI consumption
## Summary

The `flux` deployer's OCI output (`aicr bundle --deployer flux --output oci://...`) is not consumable by Flux for **local-chart components** — recipes that include `manifestFiles`, `preManifestFiles`, or vendored Helm wrappers. Discovered during PR #956 while adding the `flux-oci` KWOK lane.

## Symptom

When the bundle is consumed via an outer `OCIRepository` + `Kustomization`, the per-component `HelmRelease` CRs for local-chart components reference:

```yaml
spec:
  chart:
    spec:
      chart: ./<component-name>
      sourceRef:
        kind: GitRepository
        name: github-com-your-org-your-repo
```

The `GitRepository` resource inside the bundle points at the bundler placeholder URL `https://github.com/YOUR_ORG/YOUR_REPO.git`, which Flux's source-controller cannot fetch. helm-controller observes:

```
release is in a failed state
terminal error: exceeded maximum retries: cannot remediate failed release
```

Because Flux components are chained via `dependsOn`, the entire HelmRelease graph downstream of any local-chart component stalls on "dependency not ready".

## Why this PR doesn't fix it

Two attempts to point the inner HelmReleases at the outer OCIRepository both fail against Flux's API:

1. **`HelmRelease.spec.chart.spec.sourceRef.kind: OCIRepository`** — rejected by the validating webhook with `"supported values: HelmRepository, GitRepository, Bucket"`. The legacy chart-fetch path goes through a `HelmChart` resource whose `sourceRef` only accepts those three kinds.
2. **`HelmRelease.spec.chartRef`** — accepts `OCIRepository`, but points at the **entire OCI artifact as a single chart**. There is no path support; a multi-chart bundle cannot be addressed this way.

So a single wrapper OCIRepository cannot serve sub-paths to inner local charts. The fix has to come from the bundler's side: either eliminate the local-chart HelmReleases or publish each one as its own OCI artifact.

## Mitigation in PR #956

The `flux-oci` KWOK lane validates that the bundle artifact can be **pulled** (OCIRepository Ready=True), **applied** (Kustomization Ready=True), and **produces the expected GitOps shape** (per-component HelmRelease CRs created). It does NOT wait for HelmReleases to reach terminal state — that wait would always fail under the current placeholder-URL behavior.

This covers the most common bundle-format regression vector (pull, parse, schema correctness) for the `flux` deployer's OCI output, at the cost of deferring reconciliation validation.

### Option A: Render local-chart manifests at bundle time

For local-chart components (those that today emit `Chart.yaml + templates/` inside the bundle), run `helm template` at `aicr bundle` time and write the **rendered manifests** into the bundle. The root `kustomization.yaml` lists them directly; no per-component `HelmRelease` is emitted for these.

External-chart components (cert-manager, gpu-operator from upstream Helm repos) keep their `HelmRelease` + `HelmRepository` source CRs.

- Pros: single OCI artifact, no per-chart push, mirrors what the `helm` deployer already does on its local-chart path.
- Cons: changes the bundle layout for the `flux + OCI` case (manifests at component root, not under `templates/`); partially duplicates the `helm` deployer's rendering.

### Option B: Push each local chart as its own OCI helm artifact

For each local-chart component, push the chart with `helm.config.v1+json` config + `helm.chart.content.v1.tar+gzip` layer (the same flow we added for `argocd-helm-oci` in this PR — see `pkg/oci.PackageAndPushHelmChart`). The bundle then contains one `OCIRepository` per local chart, plus a `HelmRelease` with `spec.chartRef` pointing at it.

- Pros: native Flux pattern; uses `chartRef` (the post-v2.1 API).
- Cons: N+1 OCI pushes per bundle (one outer wrapper + one per local chart); more registry traffic; more cleanup.

### Option C: Both

Option A as the default; Option B behind a flag for users who want strict GitOps separation per chart.

## Acceptance criteria

- [ ] `aicr bundle --deployer flux --output oci://...` produces a bundle whose local-chart components reconcile successfully under Flux without external Git access.
- [ ] PR #956's `flux-oci` KWOK lane is updated to wait for HelmRelease terminal state (Ready=True OR Released+history=deployed), restoring the same depth of validation `argocd-oci` has today.
- [ ] `docs/contributor/kwok-testing.md` "Local-chart skip" note is removed.
- [ ] Decision between Option A, B, or C captured in an ADR or design doc before implementation.

## Related

- PR #956 (introduced the `flux-oci` lane with this gap deferred)
- ADR-008
- Issue #843 (parent KWOK deployer-matrix work)
- #960, #961 (similar bundle-format gaps surfaced by adding OCI lanes)

haarchri:
## Summary

Local-chart `HelmReleases` emitted by the Flux deployer referenced a placeholder `GitRepository` (https://github.com/YOUR_ORG/YOUR_REPO.git) that stalls under OCI consumption because `spec.chart.spec.sourceRef `rejects `OCIRepository`. This PR switches local-chart components to emit `ArtifactGenerator` + `ExternalArtifact` pairs via `spec.chartRef,` enabling full HelmRelease reconciliation in the KWOK flux-oci lane.

## Motivation / Context

When consuming AICR bundles via Flux OCI, `helm-controller` cannot fetch local charts from a `GitRepository` that points at a placeholder URL. Every local-chart HelmRelease stalls, and the `dependsOn` chain cascades the failure through the entire bundle.

Fixes: #964
Related: N/A

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: KWOK test infrastructure (`kwok/scripts/`)

## Implementation Notes

Bundler changes (pkg/bundler/deployer/flux/):

New ArtifactGenerator CR template (`artifactgenerator.yaml.tmpl`) extracts each local chart's subdirectory from the outer `OCIRepository` into an `ExternalArtifact`.
New `helmrelease-chartref.yaml.tmpl` template emits `HelmReleases` with `spec.chartRef` pointing at the `ExternalArtifact` instead of `spec.chart.spec.sourceRef` pointing at a `GitRepository`.
When `OCISourceName` is set (via `--flux-oci-source-name`), local-chart paths (manifest-only, pre-manifests, vendored wrappers) all route through the ArtifactGenerator path. When empty, the existing GitRepository code path is preserved.
All Flux CR templates now accept an explicit Namespace field instead of hardcoding flux-system.

CLI changes (`pkg/cli/bundle.go`):
New `--flux-oci-source-name` flag (default: aicr-bundle) sets the `OCIRepository` CR name that `ArtifactGenerators` reference.
New `--flux-namespace` flag (`default: flux-system`) sets the namespace for all generated Flux CRs.
Both flags are wired through `config.WithOCISourceName` / `config.WithFluxNamespace`.

KWOK test changes (`kwok/scripts/`):
install-infra.sh: Switched from raw manifest apply to flux install -`-components-extra=source-watcher`, patches helm-controller with `--feature-gates=ExternalArtifact=true`, and waits for `ArtifactGenerator`/`ExternalArtifact` CRDs.
validate-scheduling.sh: wait_for_flux_sync now validates full HelmRelease reconciliation (`Ready=True`) and `ArtifactGenerator` readiness instead of only checking bundle application. Deadline bumped from 300s to 500s to accommodate the additional reconciliation steps. dump_flux_failures extended with `ArtifactGenerator`, `ExternalArtifact`, and source-watcher diagnostics. Kind control-plane node is excluded from bundle DaemonSets to prevent CrashLoops on non-cloud nodes.
tools/setup-tools: Added flux CLI installation.

_One issue discovered during testing:_ 

Flux appends + to the chart version (e.g. `0.1.0+09d01b0b4d7d`), and the `+` character is invalid in Kubernetes label values. This breaks any manifest template that uses `helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}`. This is a known Flux/Helm interaction issue documented in [fluxcd/flux2#2868 (comment)](https://github.com/fluxcd/flux2/issues/2868). Fixed by sanitizing the label with `{{ .Chart.Version | replace "+" "_" | trunc 63 }}` and adding `replace/trunc` to the bundler's manifest template engine.

_KWOK Kind control-plane node exclusion:_ 
In the KWOK test lane, the Kind control-plane node (the only real node) must be excluded from bundle DaemonSets, otherwise pods like `ebs-csi-node` and `nfd-worker` actually start on it, CrashLoop because cloud services are unavailable, and stall the entire HelmRelease dependsOn chain. Rather than skipping or manipulating HelmRelease readiness checks, we exclude the node properly:

Label `eks.amazonaws.com/compute-type=hybrid , the ebs-csi-driver DaemonSet has a NotIn [fargate, auto, hybrid] nodeAffinity, so this prevents scheduling.
Taint `nfd-excluded=true:NoSchedule`, NFD worker respects this and skips the node.

## Testing

```bash
make qualify
```

```bash
unset GITLAB_TOKEN && make kwok-test-deployer RECIPE=gb200-eks-training DEPLOYER=flux-oci 2>&1
[...]
[INFO] ========================================
[INFO] Results
[INFO] ========================================
  ✓ gb200-eks-training
[INFO] Cleaning up for next test...
node "gpu-0" force deleted
node "gpu-1" force deleted
node "gpu-2" force deleted
node "gpu-3" force deleted
node "system-0" force deleted
node "system-1" force deleted
No resources found
[INFO] All 1 recipe(s) passed!
[INFO] Cluster preserved. Delete with: kind delete cluster --name aicr-kwok-test
```

- Unit tests added for `ArtifactGenerator` and chartRef HelmRelease generation
- Unit tests added for replace/trunc template funcs
- Unit tests added for OCISourceName/FluxNamespace config options
- KWOK flux-oci lane now validates full HelmRelease reconciliation (Ready=True) and ArtifactGenerator readiness

## Risk Assessment

**Rollout notes:**
The new `ArtifactGenerator` path only activates when `--flux-oci-source-name` is set (which happens automatically when --deployer flux + OCI output). Git-based Flux deployments are unaffected. Target clusters require Flux v2.7+ with `source-watcher` deployed and `ExternalArtifact=true` feature gate on `helm-controller`.

## Checklist


recipes/components/network-operator/manifests/nic-cluster-policy-aks.yaml
mchmarny, line 1: The same `helm.sh/chart` truncation fix CodeRabbit flagged on `kernel-module-params.yaml` and `nfd-network-rule.yaml` applies to **5 more files** modified by this PR — each carries the identical `{{ .Chart.Name }}-{{ .Chart.Version | replace "+" "_" | trunc 63 }}` pattern that truncates only the version:

- `recipes/components/network-operator/manifests/ib-node-config-aks.yaml`
- `recipes/components/network-operator/manifests/nic-cluster-policy-aks.yaml` (this file)
- `recipes/components/nodewright-customizations/manifests/no-op.yaml`
- `recipes/components/nodewright-customizations/manifests/tuning-gke.yaml`
- `recipes/components/nodewright-customizations/manifests/tuning.yaml`

Apply the same `printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-"` shape across all 7 manifests so the label is guaranteed under 63 chars regardless of `.Chart.Name` length.

pkg/bundler/deployer/flux/helm.go
```diff
@@ -371,16 +438,49 @@ func (g *Generator) generateVendoredHelmComponent(ctx context.Context, ref recip
...
+
+		crData := ChartRefHelmReleaseData{
+			Name:            ref.Name,
+			Namespace:       g.resolveNamespace(),
+			TargetNamespace: ref.Namespace,
+			ChartRefName:    agName,
+			DependsOn:       dependsOn,
+			ValuesFrom:      valuesFrom,
+			ValuesYAML:      valuesYAML,
+		}
```
mchmarny: The OCI-mode emit block here (`ArtifactGenerator` + chartRef `HelmRelease`) is near-identical to the one in `generateManifestHelmChart` ~30 lines higher. Same `ArtifactGeneratorData` shape, same `ChartRefHelmReleaseData` shape, only `dirName` vs `ref.Name` differs.

Worth extracting a helper, e.g.:

```go
func (g *Generator) writeOCIArtifactPair(name, targetNS string, dependsOn []DependsOnRef,
    valuesFrom []ValuesFromRef, valuesYAML string, compDir string, output *deployer.Output,
) ([]string, error) { ... }
```

Two near-duplicate blocks make any future change to the OCI shape (e.g., adding `interval`, RBAC scoping, or fixing a subdir glob bug) prone to drift between the manifest-only and vendored paths.

pkg/manifest/render.go
```diff
@@ -130,5 +130,14 @@ func helmFuncMap() template.FuncMap {
 			}
 			return val
 		},
+		"replace": func(old, new, src string) string {
+			return strings.ReplaceAll(src, old, new)
+		},
+		"trunc": func(c int, s string) string {
+			if c < 0 || c >= len(s) {
+				return s
+			}
```
mchmarny, line 145: `trunc` here diverges from Sprig's semantics on **negative `c`**:

- Sprig: `trunc -5 "hello world"` → `"world"` (last `|c|` chars)
- This impl: `trunc -5 "hello world"` → `"hello world"` (returns input unchanged)

The test at `render_test.go:198` explicitly pins the non-Sprig behavior (`trunc(-1, "test") == "test"`). For the current usage in chart-label templates `c` is always positive (63), so this is moot today — but the function name and chart-template idiom strongly imply Sprig parity, and a contributor copying a Sprig-style template that uses negative truncation (`trunc -10 ...`) will silently get the wrong output.

Either match Sprig:

```go
"trunc": func(c int, s string) string {
    if c < 0 {
        if -c >= len(s) { return s }
        return s[len(s)+c:]
    }
    if c >= len(s) { return s }
    return s[:c]
}
```

…or rename and document explicitly (`truncFront(c, s)` / godoc that calls out the divergence). Same applies to `replace` — minor in scope (drop-in match for `strings.ReplaceAll`) but worth a one-line godoc that says "use Sprig parity" or "use these for chart-label fields only" so future contributors aren't surprised.

kwok/scripts/validate-scheduling.sh
```diff
@@ -1824,6 +1857,25 @@ main() {
...
+    # (AWS IMDS, GCP metadata, etc.) are unavailable.
+    #
+    # Two exclusions are needed:
+    #   1. eks.amazonaws.com/compute-type=hybrid  — the ebs-csi-driver
+    #      DaemonSet uses a NotIn affinity that skips fargate/auto/hybrid
+    #      nodes. Without this label the Kind node passes the NotIn check.
+    #   2. nfd-excluded=true:NoSchedule taint — NFD worker DaemonSet
+    #      respects this taint and will not schedule on tainted nodes.
+    local cp_node
+    cp_node=$(kubectl get nodes --selector='!kwok.x-k8s.io/node' -o jsonpath='{.items[0].metadata.name}' 2>/dev/null || true)
```
mchmarny: `--selector='!kwok.x-k8s.io/node' -o jsonpath='{.items[0].metadata.name}'` picks only the first non-kwok node, but the comment claims "the only real node". That assumption isn't enforced — if someone runs this against a multi-node Kind cluster (e.g., 1 control-plane + 1 worker), only one node will be labeled/tainted and the bundle DaemonSets may still schedule on the others, reproducing the CrashLoop the comment is trying to prevent.

Cheap hardening:

```bash
cp_nodes=$(kubectl get nodes --selector='!kwok.x-k8s.io/node' -o jsonpath='{.items[*].metadata.name}' 2>/dev/null || true)
if [[ -z "$cp_nodes" ]]; then
    log_warn "No real (non-kwok) nodes found to exclude — DaemonSet CrashLoops may occur"
fi
for cp_node in $cp_nodes; do
    log_info "Excluding real node ($cp_node) from bundle DaemonSets"
    kubectl label node "$cp_node" eks.amazonaws.com/compute-type=hybrid --overwrite
    kubectl taint nodes "$cp_node" nfd-excluded=true:NoSchedule --overwrite
done
```

Alternatively, assert exactly one real node exists and fail loud otherwise.

pkg/cli/bundle.go
```diff
@@ -441,6 +472,22 @@ Package with explicit tag (overrides CLI version):
...
+				Value:    "aicr-bundle",
+				Category: catDeployment,
+			},
+			&cli.StringFlag{
+				Name: "flux-namespace",
+				Usage: "Kubernetes namespace where Flux CRs (HelmRelease, sources, " +
+					"ArtifactGenerator) are deployed. Must match the namespace of " +
+					"the Flux installation in the target cluster.",
+				Value:    config.DefaultFluxNamespace,
+				Category: catDeployment,
```
mchmarny, line 502: `--flux-oci-source-name` and `--flux-namespace` are registered in the global `Flags` array, so they show up in `aicr bundle --help` for **every** deployer (helm/argocd/argocd-helm/helmfile/flux). The parser only reads them when `opts.deployer == config.DeployerFlux` (lines 197-204), so they're silently ignored elsewhere.

Two options:

1. **Document scope in the `Usage` field** — append `" (--deployer flux only)"` so the help line is self-documenting.
2. **Validate non-flux deployers don't pass them** — emit a warning or error when these flags are non-default but deployer != flux, so a user who misreads the help and sets them on `--deployer helm` knows their config was ignored.

(1) is the lower-friction fix; (2) is sturdier. Mostly a UX nit but Slack-shareable feature-flag discoverability matters for new users.

mchmarny (commented):
The `ArtifactGenerator` + `ExternalArtifact` shape via `spec.chartRef` is the right way to make local-chart `HelmReleases` reconcile under Flux OCI consumption, and the KWOK lane now actually validates `HelmRelease Ready=True` (was previously deferred). 

Few net-new items that CodeRabbit has not already covered. All medium/low — none blocking:

- **Chart label fix applies to 5 more manifests** beyond the 2 CodeRabbit flagged (same `replace+trunc` pattern across `nodewright-customizations/manifests/*.yaml` and `network-operator/manifests/{ib,nic,nfd}-*.yaml`).
- **Code duplication** between the OCI-mode emit blocks in `generateManifestHelmChart` and `generateVendoredHelmComponent` — worth extracting a helper before more deployer changes drift the two paths.
- **Kind control-plane node selection** picks `.items[0]` and labels/taints exactly one node. Comment says "the only real node" but isn't enforced — fragile if anyone runs against a multi-node Kind setup.
- **`--flux-*` flags visible in `--help` for all deployers** but only honored when `--deployer flux`.

haarchri:
@mchmarny after you updated the branch - github incident jumped in - can you rerun the CI ? - thanks

mchmarny (approved):

Issue #980, opened 2026-05-19
feat(gpu-operator-post): add Helm post-install hook to re-roll DRA kubelet plugin across all deployers


yuanchen8911:
## Summary

Add a Helm `post-install,post-upgrade` hook to the `gpu-operator-post` chart that waits for `ClusterPolicy.status.state == ready` and then re-rolls the `nvidia-dra-driver-gpu` kubelet-plugin DaemonSet. This extends the same protection PR #965 added to the Helm-deployer `deploy.sh` flow to **every deployer** that consumes the rendered Helm releases — Helm direct, helmfile, Flux, Argo CD, argocd-helm.

## Motivation

PR #965 closed two regressions introduced by the `gpu-operator v25.10.1 → v26.3.1` bump:

1. **Stale NVML after driver migration.** The `nvidia-dra-driver-gpu` kubelet-plugin's NVML handle goes stale when gpu-operator's `k8s-driver-manager` reloads kernel modules. Mitigated by a `aicr.nvidia.com/gpu-operator-chart-version` annotation on the DRA pod template that forces a re-roll on chart bumps (every deployer renders the template, so every deployer re-rolls).
2. **Timing race within a deployer run.** Even after the annotation fires correctly, `k8s-driver-manager` runs the per-node module reload **asynchronously** after `helm upgrade gpu-operator` returns. On multi-GPU-node clusters, the DRA pod can re-roll *before* the slowest node's migration finishes — leaving the freshly-rolled pod pinned to the pre-migration driver state, which produces:
   ```
   invalid CDI Spec: failed add device "<uuid>-gpu-N": empty device edits
   ```
   Reproduced on a GB200 EKS cluster (`yljtrxpmzu`, p6e-gb200.36xlarge × 2 nodes) during PR #965 validation. PR #965 closes this with a `kubectl wait` + `kubectl rollout restart` block in `pkg/bundler/deployer/helm/templates/deploy.sh.tmpl` — but **only the default Helm deployer runs `deploy.sh`**. Users who consume the same bundle via `helmfile apply`, `flux reconcile`, or Argo CD's sync get the rendered manifests but skip the post-deploy script. They still hit the race.

This issue tracks the **deployer-agnostic** durable fix.

## Proposed mechanism: Helm post-install hook on `gpu-operator-post`

Add a Job manifest to `recipes/components/gpu-operator-post/templates/` (or wherever the chart templates live) annotated as a Helm hook:

```yaml
apiVersion: batch/v1
[...]

Issue #894, opened 2026-05-14
chore(recipes): upgrade gpu-operator to latest stable


yuanchen8911:
## Summary

Audit and update the `gpu-operator` chart pin in `recipes/registry.yaml` to the current stable upstream release.

## Context

Carved out from #698 (broader component-version audit). All other components in that audit have been updated; only `gpu-operator` remains.

Current pin: `v25.10.1` (chart `gpu-operator-v25.10.1`).

## Definition of done

- [ ] Compare the current pin against the latest stable upstream release of NVIDIA `gpu-operator`.
- [ ] If a newer version is selected, update `recipes/registry.yaml`, regenerate `docs/user/container-images.md` (`make bom-docs`), and justify the bump in the PR description.
- [ ] If the current pin is intentionally held back, document the reason inline in `recipes/registry.yaml`.
- [ ] `make qualify` passes locally.
- [ ] At least one full GPU CI run (H100 nvkind) passes against the updated bundle.

## Out of scope

- Other components in the audit (handled in #698 and its predecessors).

PR #1033, opened 2026-05-26
feat(bundler): derive DRA chart-version annotation from resolved recipe (#973)

### LINKED ISSUE #973: chore(recipes): make DRA-rollout trigger durable (don't rely on manual annotation bump)
## Summary

PR #965 mitigates the stale-NVML class of bug by hard-coding the gpu-operator chart version into a DRA pod-template annotation. The annotation works as long as it's kept in lockstep with the gpu-operator chart pin — which depends on a maintainer remembering to bump both in the same PR.

That's brittle. A future gpu-operator bump that forgets the annotation will silently re-introduce the original failure (stale NVML, DRA `FailedPrepareDynamicResources`, GPU workloads stuck in `ContainerCreating`).

This issue tracks the **durable** fix.

## Pre-PR-#965 repro (context)

After upgrading gpu-operator from v25.10.1 (driver `580.105.08`) → v26.3.1 (driver `580.126.20`), the `nvidia-dra-driver-gpu` kubelet-plugin DaemonSet pod is not restarted by `helm upgrade`. Its in-pod NVML handle remains bound to the old driver, and DRA allocation fails with:

```
FailedPrepareDynamicResources: Failed to prepare dynamic resources:
  NodePrepareResources failed for ResourceClaim ...:
  unable to create CDI spec file for claim: ...:
  failed to initialize NVML: Driver/library version mismatch
```

Workloads that request GPUs via DRA (Dynamo `DynamoGraphDeployment`, `aicr validate --phase performance` `inference-perf`, etc.) stay in `ContainerCreating` indefinitely. Manual workaround pre-#965: `kubectl delete pod -n nvidia-dra-driver -l app.kubernetes.io/name=nvidia-dra-driver-gpu`.

## What's already in place (don't reinvent)

`deploy.sh.tmpl` ([pkg/bundler/deployer/helm/templates/deploy.sh.tmpl:354](https://github.com/NVIDIA/aicr/blob/main/pkg/bundler/deployer/helm/templates/deploy.sh.tmpl#L354)) already emits a post-install `kubectl rollout restart` for the DRA kubelet-plugin DaemonSet, gated on `name == "nvidia-dra-driver-gpu"`. So:

- ✅ Users running the bundled `./deploy.sh` (**default Helm deployer**) get the restart automatically.
- ❌ Users consuming `helmfile`, `flux`, `argocd`, or `argocd-helm` artifacts — there's no equivalent in those deployer templates (`grep -rn "kubelet-plugin" pkg/bundler/deployer/{helmfile,flux,argocd*}/` returns zero matches).
- ❌ Users running `helm upgrade` directly on the upstream chart without going through AICR — out of scope for AICR; only an upstream chart change can cover them.

(`localformat` is the shared folder writer that all deployers compose, not a user-facing deployer itself.)

PR #965's annotation closes the GitOps-deployer gap by making the pod-template change visible in every deployer's rendered artifact — they all re-roll the DaemonSet on next reconcile. The durability question is how to keep the annotation correct across future chart bumps.

## Why PR #965's mitigation isn't durable

```yaml
# recipes/components/nvidia-dra-driver-gpu/values.yaml
kubeletPlugin:
  podAnnotations:
    aicr.nvidia.com/gpu-operator-chart-version: v26.3.1   # ← must be bumped manually
```

- The annotation value is a free-form string. Helm has no awareness it's supposed to mirror `gpu-operator`'s chart version.
- A future PR that bumps `gpu-operator` (e.g. v26.4.0) but forgets the annotation produces identical rendered DaemonSet manifests for `nvidia-dra-driver-gpu`. `helm upgrade` skips the roll. Stale NVML returns.
- `make qualify` does not catch this — no static or runtime check verifies the two pins are coherent.

## Recommended durable fix: bundler-derived annotation

**Inject the annotation at bundle-generation time from the resolved recipe**, not at hand-edit time.

Concrete shape:

- **Source**: read the `gpu-operator` `ComponentRef.Version` from the resolved recipe (the same value the resolver writes into the bundle for the gpu-operator chart).
- **Destination**: inject into both
  - `controller.podAnnotations.aicr.nvidia.com/gpu-operator-chart-version`
  - `kubeletPlugin.podAnnotations.aicr.nvidia.com/gpu-operator-chart-version`
  of the `nvidia-dra-driver-gpu` rendered values.
- **Trigger condition**: only inject when **both** `gpu-operator` and `nvidia-dra-driver-gpu` are *enabled* in the **filtered resolved recipe**. `DefaultBundler.Make` ([`pkg/bundler/bundler.go:213-254`](https://github.com/NVIDIA/aicr/blob/main/pkg/bundler/bundler.go#L213)) builds an `enabledRefs` slice that excludes disabled components before any value extraction; the helper should consume that filtered `recipeResult.ComponentRefs`, not the unfiltered recipe.
- **Injection point**: in `DefaultBundler.Make`, **after** `extractComponentValues` ([line 277](https://github.com/NVIDIA/aicr/blob/main/pkg/bundler/bundler.go#L277)) has loaded recipe values, applied user `--set` overrides, and applied scheduling overrides — but **before** `buildDeployer` ([line 311](https://github.com/NVIDIA/aicr/blob/main/pkg/bundler/bundler.go#L311)). This window guarantees:
  - every deployer (Helm, helmfile, Flux, Argo CD, argocd-helm) receives the same final `componentValues` map;
  - a user `--set gpuoperator:rollout-trigger=foo` does **not** accidentally defeat the internal rollout annotation (unless AICR intentionally decides to allow that, which would be a separate explicit knob).
- **Maintainer-facing values file** in `recipes/components/nvidia-dra-driver-gpu/values.yaml` then carries no chart-version string at all — that data lives only in the resolved recipe.

Why this is the right shape:
- **Works for every deployer.** All deployers consume `componentValues` from the same code path, so all of them get the pod-template change.
- **Single source of truth.** The annotation is derived from the resolved recipe (which is exactly the source #966 is converging on for chart versions), so there's no separate string to maintain.
- **Bisects cleanly.** A maintainer who only edits `gpu-operator`'s pin gets the DRA annotation update for free in the bundle output.

## Near-term guardrail (safe to ship while the bundler-derived fix is in flight)

A static coherence check, ideally wired into `make qualify` or a new `make check-dra-rollout-trigger`:

- Compare the **resolved recipe's** `gpu-operator` `ComponentRef.Version` against the annotation value in `recipes/components/nvidia-dra-driver-gpu/values.yaml`. Fail if they differ.
- **Compare against the resolved recipe, not against `recipes/components/gpu-operator/values.yaml` or `helm.defaultVersion` in the registry.** This aligns with #966 (resolved recipe as single source of truth for chart versions) and avoids creating a new drift point.

This is the cheapest guardrail and is worth shipping immediately if the bundler-derived fix isn't ready for the same release.

## Definition of done

- [ ] **Primary**: `DefaultBundler.Make` injects `aicr.nvidia.com/gpu-operator-chart-version` (or equivalent annotation) into the rendered `nvidia-dra-driver-gpu` `componentValues` for both `controller` and `kubeletPlugin`, sourced from the resolved gpu-operator `ComponentRef.Version`, gated on both refs being enabled in the filtered recipe, injected after `extractComponentValues` and before `buildDeployer`. Maintainer-facing values file no longer carries the chart-version string.
- [ ] **Acceptance test (generated-artifact parity)**: bump the gpu-operator version in a fixture; assert the rendered DRA `podAnnotations` change in at least the **default Helm-deployer** output AND one GitOps-deployer output (helmfile or flux or argocd). Catches the cross-deployer parity risk.
- [ ] **Acceptance test (disabled-component negative case)**: with `nvidia-dra-driver-gpu` *disabled* in the recipe, assert the helper does not inject the annotation anywhere AND does not emit a warning. Proves the gating is correct.
- [ ] **Near-term guardrail (optional, ship if bundler change slips)**: `make` target that fails when the resolved gpu-operator chart version diverges from the DRA annotation. Compares against resolved recipe (not registry default).
- [ ] **Cluster verification**: on a cluster running an older gpu-operator, `aicr bundle` + deploy via the chosen deployer rolls the DRA pods automatically, and `aicr validate --phase performance` passes without manual intervention.

## Out of scope

- Bumping `nvidia-dra-driver-gpu` itself (still v25.12.0; not driver-version-bound in its registry pin).
- Coverage for users running the upstream `k8s-dra-driver-gpu` Helm chart directly without AICR artifacts — only an upstream chart change can cover that population. Worth a separate upstream conversation, not AICR-tracked.

## Related

- #894 — gpu-operator chart bump that surfaced this class of bug
- #965 — PR that ships the v26.3.1 bump AND the (manual-coupling) annotation mitigation
- #966 — separate refactor (resolved recipe as single source of truth for chart versions); the bundler-derived annotation slots into that work cleanly

---

**Credit:** v2 reframing (issue is a durability question, not the original bug) from Codex review of v1. v3 refinements (deploy.sh hook already present, bundler-derived annotation as the right primary, resolved-recipe basis for the coherence check, generated-artifact tests, upstream-chart-direct caveat) from Codex review of v2. v4 refinements (filtered-recipe gating, exact injection point between `extractComponentValues` and `buildDeployer`, "default Helm deployer" terminology, disabled-component negative test) from Codex review of v3.

yuanchen8911:
## Summary

Replace the hand-maintained `aicr.nvidia.com/gpu-operator-chart-version` annotation in `recipes/components/nvidia-dra-driver-gpu/values.yaml` with a bundler-side injection that derives the value from the resolved `gpu-operator` componentRef and writes it onto both `controller` and `kubeletPlugin` `podAnnotations` in the rendered DRA values. The annotation can no longer drift from the actual chart pin.

## Motivation / Context

PR #965 closed the stale-NVML class of bug (gpu-operator chart bump → `k8s-driver-manager` reloads kernel modules async → DRA kubelet plugin's NVML handle goes stale → CDI spec generation fails with `Driver/library version mismatch` → DRA-allocated pods stuck in `ContainerCreating`) by hard-coding the gpu-operator chart version into the DRA pod-template annotation. The annotation works, but only as long as a maintainer remembers to bump both pins in lockstep. A future PR that bumped gpu-operator and forgot the annotation would produce identical rendered DaemonSet manifests, helm upgrade would skip the re-roll, and stale NVML would return silently. `make qualify` did not catch the drift because no static check verified the two pins.

Fixes: #973
Related: #965 (the chart-version annotation this durabilizes), #980 (orthogonal cross-deployer corrective restart), #894 (chart bump that first surfaced the stale-NVML class of bug)

## Type of Change

- [x] Refactoring (no functional changes) — net behavior is identical when gpu-operator chart pin matches the value previously hand-edited; durability change makes future drift impossible

## Component(s) Affected

- [x] Bundlers (`pkg/bundler`, `pkg/component/*`) — new `injectDRAChartVersionAnnotation` helper + call site in `DefaultBundler.Make`
- [x] Recipe engine / data (`pkg/recipe`) — no Go change, only `recipes/components/nvidia-dra-driver-gpu/values.yaml` (stripped hand-maintained chart-version string)

## Implementation Notes

**Helper.** `injectDRAChartVersionAnnotation` iterates the filtered `recipeResult.ComponentRefs` (disabled components are removed upstream in `DefaultBundler.Make` at `pkg/bundler/bundler.go:219-257`), extracts the gpu-operator chart version, and gates on BOTH `gpu-operator` and `nvidia-dra-driver-gpu` being enabled. When both are present, it writes `aicr.nvidia.com/gpu-operator-chart-version` onto `controller` and `kubeletPlugin` `podAnnotations` of the nvidia-dra-driver-gpu values map, creating nested maps lazily so existing values (`priorityClassName`, other annotations) are preserved.

**Injection point.** In `DefaultBundler.Make` AFTER `extractComponentValues` — so the values map is populated and user `--set` overrides have already landed — and BEFORE `buildDeployer` — so every deployer receives the same final `componentValues` map. Placing the call AFTER `--set` is deliberate: a user override of this specific annotation key is intentionally NOT honored; the annotation must always reflect the actual resolved gpu-operator chart version, or the rollout-trigger semantics break. `TestInjectDRAChartVersionAnnotation_OverridesUserSet` pins this invariant.

**Cross-deployer parity (structural).** All five supported deployers consume `Generator.ComponentValues[ref.Name]` from the same post-injection map:

| Deployer | Consumption site |
|----------|------------------|
| Helm | `pkg/bundler/deployer/helm/helm.go:144` |
| helmfile | `pkg/bundler/deployer/helmfile/helmfile.go:136` |
| Flux | `pkg/bundler/deployer/flux/helm.go:130` |
| Argo CD | `pkg/bundler/deployer/argocd/argocd.go:552` |
| argocd-helm | `pkg/bundler/deployer/argocdhelm/argocdhelm.go:387` |

The parity test (see Testing) directly exercises both internal consumption shapes — inline values marshal (helmfile) and `localformat.Component.Values` write (Helm, reused by argocd / argocd-helm) — and asserts the annotation lands in the rendered `values.yaml`. The remaining deployers are guaranteed by the single-map design; renaming `Generator.ComponentValues` would break compilation everywhere simultaneously.

**Pattern.** Mirrors the GKE critical-priority quota helper (`pkg/bundler/bundler.go:injectGKECriticalPriorityQuotas`) so readers familiar with that pattern can pick this one up without context-switching.

**Recipe values change.** `recipes/components/nvidia-dra-driver-gpu/values.yaml` no longer carries the chart-version string at all. The old comment block is replaced with a short pointer to this helper. `controller.priorityClassName: ""` and `kubeletPlugin.priorityClassName: ""` remain unchanged.

## Behavior Delta

**Standalone-DRA recipes (DRA enabled, gpu-operator disabled) no longer render the annotation, by design.** The previously hand-maintained annotation lived inside `recipes/components/nvidia-dra-driver-gpu/values.yaml`, so it was written unconditionally whenever the DRA component was bundled — including the (currently unused) case where DRA is deployed without gpu-operator (for example, a host-installed driver via `--set gpuoperator:enabled=false`). The new helper gates on BOTH components being enabled, so a standalone-DRA bundle renders without the annotation.

This is the intended semantic: the annotation exists solely to force a DaemonSet re-roll when the **gpu-operator chart pin** bumps. With gpu-operator disabled, there is no chart pin to mirror and the rollout trigger has nothing to fire on. No in-tree overlay currently exercises this configuration, so the change is theoretical for today's recipes; it is the right shape going forward.

## Testing

```bash
unset GITLAB_TOKEN ; make qualify
# All 22 chainsaw tests pass.
# Coverage: 76.9% (threshold: 75%) — passes the project floor.
# Vulnerability scan: clean.

GOFLAGS="-mod=vendor" go test -coverprofile=/tmp/cover.out ./pkg/bundler
# pkg/bundler coverage 67.0% → 69.0% (+2.0% on this PR)
# injectDRAChartVersionAnnotation: 100% statement coverage
```

**Tests added:**

- **Unit tests** (`pkg/bundler/bundler_dra_annotation_test.go`):
  - positive case both pod paths, annotation matches resolved gpu-operator version
  - negative case nvidia-dra-driver-gpu disabled → no-op, no DRA entry materialized
  - negative case gpu-operator disabled → no-op (pins the Behavior Delta above)
  - preserves existing `priorityClassName` + unrelated annotations under both pod paths
  - overrides user `--set` of the annotation key (proves the "must reflect actual chart version" invariant)
  - version variants: `v`-prefixed, unprefixed, pre-release, build metadata
  - nil-tolerant inputs (nil componentValues, nil recipe, both nil)
  - lazy-creation of `componentValues[nvidia-dra-driver-gpu]` when the entry isn't initialized
- **Generated-artifact parity test** (`pkg/bundler/bundler_dra_annotation_parity_test.go`):
  - bundles the same recipe through the **Helm deployer** AND the **helmfile deployer**, asserts the annotation lands in the rendered DRA `values.yaml` for both outputs. These two deployers exercise both internal consumption shapes; Flux, Argo CD, and argocd-helm are guaranteed by the single-map design described in Implementation Notes.
  - disabled-recipe negative case: when nvidia-dra-driver-gpu is filtered out, the gpu-operator values must not leak any DRA-related annotation key

## Risk Assessment

- [x] **Low** — Isolated bundler-side helper with 100% unit coverage and a cross-deployer parity test. Net rendered output for the current `gpu-operator v26.3.1 / nvidia-dra-driver-gpu v25.12.0` pair is byte-identical to the prior hand-maintained annotation. Easy revert (single function + call site + values comment).

**Rollout notes:** No flag, no opt-in. The next gpu-operator chart bump after this PR lands gets the annotation derived automatically; no values-file edit needed. Existing bundles regenerated from the same gpu-operator pin produce the same output as today. See **Behavior Delta** for the only semantic change (annotation absent when gpu-operator is disabled).

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — via `make qualify`
- [x] Linter passes (`make lint`) — via `make qualify`
- [x] I updated docs if user-facing behavior changed — no user-facing surface; recipe values comment refreshed
- [x] Changes follow existing patterns in the codebase — mirrors `injectGKECriticalPriorityQuotas`

### DIFF
--- a/pkg/bundler/bundler.go
+++ b/pkg/bundler/bundler.go
@@ -284,6 +284,13 @@ func (b *DefaultBundler) Make(ctx context.Context, recipeResult *recipe.RecipeRe
 			"failed to extract component values", err)
 	}
 
+	// Bundler-derived annotations that must reflect the final resolved
+	// recipe state, applied AFTER extractComponentValues so that user
+	// --set overrides cannot defeat them. Every deployer (Helm,
+	// helmfile, Flux, Argo CD, argocd-helm) sees the same final map.
+	// See issue #973.
+	b.injectDRAChartVersionAnnotation(componentValues, recipeResult)
+
 	if warningErr := b.warnMissingStorageClassForPVCs(ctx, recipeResult, componentValues); warningErr != nil {
 		return nil, warningErr
 	}
@@ -1502,4 +1509,136 @@ func renderGKECriticalPriorityQuota(namespace string, pods int) ([]byte, error) 
 		},
 	}
 	return serializer.MarshalYAMLDeterministic(quota)
+}
+
+// draChartVersionAnnotation is the key written onto the
+// nvidia-dra-driver-gpu controller and kubelet-plugin pod templates.
+// Its value mirrors the resolved gpu-operator componentRef version
+// so that any gpu-operator chart bump produces a rendered pod-template
+// diff that forces helm upgrade (and every other deployer) to re-roll
+// the DaemonSet — clearing the kubelet plugin's stale NVML handle that
+// would otherwise pin to the pre-migration driver state.
+const draChartVersionAnnotation = "aicr.nvidia.com/gpu-operator-chart-version"
+
+// draComponentName / gpuOperatorComponentName are the registry-level
+// component names this injection couples together. Both must be
+// enabled in the filtered resolved recipe before the annotation is
+// written; recipes that disable either remain untouched.
+const (
+	draComponentName         = "nvidia-dra-driver-gpu"
+	gpuOperatorComponentName = "gpu-operator"
+)
+
+// injectDRAChartVersionAnnotation writes the resolved gpu-operator
+// chart version into the nvidia-dra-driver-gpu controller and
+// kubelet-plugin podAnnotations on the bundler's componentValues map.
+// Replaces the prior hand-maintained value in
+// recipes/components/nvidia-dra-driver-gpu/values.yaml — see #973.
+//
+// Why this exists:
+//
+// PR #965 mitigated the stale-NVML class of bug (gpu-operator chart
+// bump → k8s-driver-manager reloads kernel modules async → DRA kubelet
+// plugin's NVML handle goes stale → CDI spec generation fails) by
+// hard-coding the gpu-operator chart version into the DRA pod-template
+// annotation. The annotation works as long as it stays in lockstep
+// with the chart pin, but the coupling depends on a maintainer
+// remembering to bump both in the same PR. A future PR that bumps
+// gpu-operator and forgets the annotation produces identical rendered
+// DaemonSet manifests, so helm upgrade skips the re-roll and stale
+// NVML returns silently. Bundler-derived injection removes the manual
+// step entirely.
+//
+// Trigger gating: BOTH gpu-operator and nvidia-dra-driver-gpu must be
+// enabled in the filtered recipe. The caller has already removed
+// disabled components from recipeResult.ComponentRefs at this point,
+// so iterating the filtered slice gives the right gating for free.
+// Recipes that disable either component leave componentValues
+// untouched.
+//
+// Injection point: called from DefaultBundler.Make AFTER
+// extractComponentValues (so the values map is populated and user
+// --set overrides have already been applied) and BEFORE buildDeployer
+// (so every deployer — Helm, helmfile, Flux, Argo CD, argocd-helm —
+// receives the same final map). Placing the call after --set means a
+// user override of this specific annotation key is intentionally NOT
+// honored; the annotation must always reflect the actual resolved
+// gpu-operator chart version, or the rollout-trigger semantics break.
+//
+// Mutates componentValues in place; nested controller / kubeletPlugin
+// / podAnnotations maps are created lazily so existing values under
+// either path (priorityClassName, other annotations, etc.) are
+// preserved.
+func (b *DefaultBundler) injectDRAChartVersionAnnotation(
+	componentValues map[string]map[string]any,
+	recipeResult *recipe.RecipeResult,
+) {
+
+	if componentValues == nil || recipeResult == nil {
+		return
+	}
+
+	var gpuOperatorEnabled, draEnabled bool
+	var gpuOperatorVersion string
+	for _, ref := range recipeResult.ComponentRefs {
+		switch ref.Name {
+		case gpuOperatorComponentName:
+			gpuOperatorEnabled = true
+			gpuOperatorVersion = ref.Version
+		case draComponentName:
+			draEnabled = true
+		}
+	}
+	if !draEnabled || !gpuOperatorEnabled {
+		// Either component disabled: nothing to mirror. Silent skip
+		// matches the "no chart pin, no rollout trigger" semantic and
+		// is exercised by the disabled-component unit tests.
+		return
+	}
+	if gpuOperatorVersion == "" {
+		// gpu-operator is enabled but the resolver produced an empty
+		// Version string. This shouldn't happen in normal recipe
+		// resolution; if it does the rollout-trigger semantics break
+		// silently — the same drift class this helper exists to
+		// eliminate. Warn so operators have a debugging signal, then
+		// skip injection rather than write an empty annotation value
+		// that itself would lock the DaemonSet to a meaningless pin.
+		slog.Warn("gpu-operator enabled with empty Version, skipping DRA chart-version annotation injection",
+			"component", gpuOperatorComponentName,
+			"draComponent", draComponentName)
+		return
+	}
+
+	draValues := componentValues[draComponentName]
+	if draValues == nil {
+		draValues = make(map[string]any)
+		componentValues[draComponentName] = draValues
+	}
+
+	// Both pod templates carry the same annotation. The controller is
+	// a Deployment (one replica per chart) and the kubeletPlugin is the
+	// DaemonSet whose NVML handle is at risk; rolling both keeps the
+	// two halves of the chart consistent across upgrades.
+	//
+	// The `, _` on the type assertions below is deliberate: values come
+	// from `extractComponentValues`, which produces `map[string]any` via
+	// yaml.v3 decoding, so a wrong-type value at `controller` /
+	// `kubeletPlugin` / `podAnnotations` cannot happen in practice. If
+	// it ever did (e.g., a hand-crafted unit test or a malformed
+	// override that's also broken for the DRA chart itself), the helper
+	// silently replaces the wrong-type value with a fresh map and lands
+	// the annotation on top.
+	for _, podPath := range []string{"controller", "kubeletPlugin"} {
+		section, _ := draValues[podPath].(map[string]any)
+		if section == nil {
+			section = make(map[string]any)
+			draValues[podPath] = section
+		}
+		annotations, _ := section["podAnnotations"].(map[string]any)
+		if annotations == nil {
+			annotations = make(map[string]any)
+			section["podAnnotations"] = annotations
+		}
+		annotations[draChartVersionAnnotation] = gpuOperatorVersion
+	}
 }
--- a/pkg/bundler/bundler_dra_annotation_parity_test.go
+++ b/pkg/bundler/bundler_dra_annotation_parity_test.go
@@ -1,0 +1,331 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bundler
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/NVIDIA/aicr/pkg/bundler/config"
+	"github.com/NVIDIA/aicr/pkg/recipe"
+)
+
+// TestMake_DRAChartVersionAnnotation_GeneratedArtifactParity is the
+// generated-artifact acceptance test the issue calls out. It bundles
+// the same recipe through two deployer code paths — the default Helm
+// deployer AND the helmfile (GitOps-style) deployer — and asserts the
+// bundler-derived aicr.nvidia.com/gpu-operator-chart-version
+// annotation lands in the rendered nvidia-dra-driver-gpu values.yaml
+// at the structurally correct `controller.podAnnotations` and
+// `kubeletPlugin.podAnnotations` paths for BOTH outputs.
+//
+// This catches the cross-deployer parity risk for the deployers whose
+// values land in a known `<order>-<name>/values.yaml` layout. Coverage
+// for Flux / Argo CD / argocd-helm (which emit different file shapes,
+// e.g. configmap-values.yaml or Application sources) lives in
+// TestMake_DRAChartVersionAnnotation_AllDeployersCarryAnnotation
+// below — that test asserts the annotation reaches *some* file in the
+// rendered bundle for every supported deployer, without coupling to
+// per-deployer layout conventions.
+func TestMake_DRAChartVersionAnnotation_GeneratedArtifactParity(t *testing.T) {
+	const (
+		gpuOpVersion       = "v26.4.0"
+		expectedAnnotation = "aicr.nvidia.com/gpu-operator-chart-version"
+	)
+
+	deployers := []struct {
+		name     string
+		deployer config.DeployerType
+		// where to look for the rendered DRA values inside the bundle dir
+		valuesPath string
+	}{
+		{
+			name:       "helm-deployer",
+			deployer:   config.DeployerHelm,
+			valuesPath: "002-nvidia-dra-driver-gpu/values.yaml",
+		},
+		{
+			name:       "helmfile-deployer",
+			deployer:   config.DeployerHelmfile,
+			valuesPath: "002-nvidia-dra-driver-gpu/values.yaml",
+		},
+	}
+
+	for _, tc := range deployers {
+		t.Run(tc.name, func(t *testing.T) {
+			cfg := config.NewConfig(
+				config.WithDeployer(tc.deployer),
+				config.WithVersion("v1.0.0"),
+			)
+			b, err := New(WithConfig(cfg))
+			if err != nil {
+				t.Fatalf("New() error = %v", err)
+			}
+
+			rr := &recipe.RecipeResult{
+				APIVersion: "aicr.nvidia.com/v1alpha1",
+				Kind:       "Recipe",
+				Criteria: &recipe.Criteria{
+					Service:     "eks",
+					Accelerator: "h100",
+					Intent:      "training",
+				},
+				ComponentRefs: []recipe.ComponentRef{
+					{
+						Name:    gpuOperatorComponentName,
+						Version: gpuOpVersion,
+						Type:    "helm",
+						Source:  "https://helm.ngc.nvidia.com/nvidia",
+					},
+					{
+						Name:    draComponentName,
+						Version: "25.12.0",
+						Type:    "helm",
+						Source:  "https://helm.ngc.nvidia.com/nvidia",
+					},
+				},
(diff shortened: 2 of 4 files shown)


pkg/bundler/bundler_dra_annotation_parity_test.go
```diff
@@ -0,0 +1,199 @@
...
+						Version: "25.12.0",
+						Type:    "helm",
+						Source:  "https://helm.ngc.nvidia.com/nvidia",
+					},
+				},
+				DeploymentOrder: []string{gpuOperatorComponentName, draComponentName},
+			}
+
+			tmpDir := t.TempDir()
+			if _, err := b.Make(context.Background(), rr, tmpDir); err != nil {
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in 8b5ef744 — both `Make()` call sites in `pkg/bundler/bundler_dra_annotation_parity_test.go` now wrap the call in a 30s `context.WithTimeout` with `defer cancel()`, matching the project's `CLAUDE.md` rule on bounded contexts for I/O.

pkg/bundler/bundler.go
```diff
@@ -1503,3 +1510,109 @@ func renderGKECriticalPriorityQuota(namespace string, pods int) ([]byte, error)
...
+	}
+	if gpuOperatorVersion == "" || !draEnabled {
+		return
+	}
+
+	draValues := componentValues[draComponentName]
+	if draValues == nil {
+		draValues = make(map[string]any)
+		componentValues[draComponentName] = draValues
+	}
```
mchmarny, line 1616: The early-return conflates two cases: (1) gpu-operator is disabled / absent from refs — silent skip is correct, and (2) gpu-operator is present in refs but has `Version: ""` — silent skip is suspicious. The second case shouldn't happen in normal recipe resolution, but if it does, the rollout-trigger semantics break silently and `make qualify` won't catch it (same drift class the PR is trying to eliminate).

Consider distinguishing the two: a `slog.Warn("gpu-operator enabled with empty Version, skipping DRA chart-version annotation injection", ...)` for case 2 gives operators a debugging signal. Cheap defensive logging that pairs well with the "annotation can never drift from the actual chart pin" guarantee in the function header.

Worth a test case too — `TestInjectDRAChartVersionAnnotation_GPUOperatorEmptyVersion` would pin the current behavior either way.

pkg/bundler/bundler_dra_annotation_parity_test.go
```diff
@@ -0,0 +1,204 @@
...
+		name     string
+		deployer config.DeployerType
+		// where to look for the rendered DRA values inside the bundle dir
+		valuesPath string
+	}{
+		{
+			name:       "helm-deployer",
+			deployer:   config.DeployerHelm,
+			valuesPath: "002-nvidia-dra-driver-gpu/values.yaml",
+		},
```
mchmarny: The PR description's structural argument ("Flux, Argo CD, and argocd-helm are guaranteed by the single-map design") holds for *whether the annotation is in the values map*, but the parity test is also implicitly asserting *where in the bundle filesystem* the rendered values land (`002-nvidia-dra-driver-gpu/values.yaml`). The other three deployers may follow different naming/numbering conventions for their per-component output — argocd in particular emits `Application` manifests with a different shape.

Extending this table to all five deployers (`config.DeployerFlux`, `config.DeployerArgoCD`, `config.DeployerArgoCDHelm`) is a few extra rows + correct `valuesPath` per deployer. Cheap insurance against a future deployer refactor that quietly drops the annotation from one output shape. The current single-map structural argument doesn't cover bundle-layout drift.

pkg/bundler/bundler.go
```diff
@@ -1503,3 +1510,109 @@ func renderGKECriticalPriorityQuota(namespace string, pods int) ([]byte, error)
...
+		return
+	}
+
+	var gpuOperatorVersion string
+	var draEnabled bool
+	for _, ref := range recipeResult.ComponentRefs {
+		switch ref.Name {
+		case gpuOperatorComponentName:
+			gpuOperatorVersion = ref.Version
+		case draComponentName:
```
mchmarny, line 1588: nit: leading blank line at the start of the function body is unusual in this codebase and reads like accidental whitespace. Drop it.

pkg/bundler/bundler.go
```diff
@@ -1503,3 +1510,109 @@ func renderGKECriticalPriorityQuota(namespace string, pods int) ([]byte, error)
...
+	// two halves of the chart consistent across upgrades.
+	for _, podPath := range []string{"controller", "kubeletPlugin"} {
+		section, _ := draValues[podPath].(map[string]any)
+		if section == nil {
+			section = make(map[string]any)
+			draValues[podPath] = section
+		}
+		annotations, _ := section["podAnnotations"].(map[string]any)
+		if annotations == nil {
+			annotations = make(map[string]any)
```
mchmarny, line 1639: nit: the discarded `ok` from `.(map[string]any)` means if a values file ever has a non-map value at `controller` or `controller.podAnnotations` (e.g., somebody set `controller: false` as a chart-disable shortcut), the existing value gets silently overwritten by an empty map and the annotation lands on top.

Practically this won't happen — `extractComponentValues` produces `map[string]any` via yaml.v3 decoding, and a wrong-type override there would already be malformed for the DRA chart — but a comment noting "wrong-type path is silently replaced; values come from yaml.v3 so this is safe in practice" would document the assumption. Or check `ok` and skip with a warn. Take or leave.

mchmarny (commented):
Clean refactor — moves the chart-version annotation from a hand-maintained values entry to a bundler-derived injection, killing the drift risk that #973 called out. Helper mirrors `injectGKECriticalPriorityQuotas` for shape consistency, and the function header doc is unusually clear about *why* the annotation exists, *when* it fires, and *where* the injection point sits in `Make`. Test coverage is thorough (positive, both gating directions, preserves-existing, overrides --set, version-string variants, nil-tolerance, lazy-map creation).

Two low-severity suggestions and two nits inline; no blockers:

- Edge case worth pinning: gpu-operator present in refs but with empty `Version` is currently a silent skip — consider a `slog.Warn` + a test row to distinguish "disabled" from "enabled-with-missing-version".
- Parity test covers helm + helmfile; the structural argument for the other three deployers is sound, but extending the table to all five would also catch bundle-layout drift (argocd/argocd-helm in particular emit a different shape).

All 102 CI checks green. Mergeable state shows "blocked" — looks like branch protection / required reviewer, not a CI failure.

pkg/bundler/bundler.go
yuanchen8911: Done in 229946ba — split into two branches: silent skip when either component is disabled, `slog.Warn` + skip when gpu-operator is enabled with empty `Version`. Added `TestInjectDRAChartVersionAnnotation_GPUOperatorEmptyVersion` to pin the behavior.

pkg/bundler/bundler_dra_annotation_parity_test.go
yuanchen8911: Good call. Added `TestMake_DRAChartVersionAnnotation_AllDeployersCarryAnnotation` in 229946ba — bundles through all 5 deployers (Helm, helmfile, Flux, Argo CD, argocd-helm) and asserts the annotation key + version both appear in at least one rendered file per bundle. Avoids hard-coding per-deployer layout conventions while still catching "deployer X silently drops the annotation." Existing strict `controller.podAnnotations` / `kubeletPlugin.podAnnotations` structural assertion stays for Helm/helmfile.

pkg/bundler/bundler.go
yuanchen8911: The blank line is actually required by the project's `golangci-lint` config: `whitespace.multi-func: true` at `.golangci.yaml:107-109` enforces a newline after a multi-line function signature, and removing it fails lint with "multi-line statement should be followed by a newline". The same pattern exists at `injectGKECriticalPriorityQuotas` for the same reason. Leaving it as-is to keep `make qualify` happy — happy to revisit if the lint rule itself is the thing we want to change.

pkg/bundler/bundler.go
yuanchen8911: Took the "leave" path with a documenting comment in 229946ba explaining the assumption: `extractComponentValues` produces `map[string]any` via yaml.v3 decoding, so a wrong-type value at `controller` / `kubeletPlugin` / `podAnnotations` can't happen in practice. If it ever did, the helper silently replaces it with a fresh map — and a wrong type at those paths would already be malformed for the DRA chart itself, so we're not masking a real-world configuration.

mchmarny (approved):

Issue #1019, opened 2026-05-25
bug(bundler): argocd-helm hardcodes parent chart name as `aicr-bundle` instead of using the published OCI artifact name


atif1996:
## Summary

When generating the app-of-apps `aicr-stack` ArgoCD Application, the `argocd-helm` deployer hardcodes `spec.source.chart: aicr-bundle` regardless of what name the user actually pushed the bundle under.

If a user publishes to `oci://<registry>/<org>/<my-bundle-name>:<tag>`, ArgoCD pulls the correct chart at install time (via `helm install <release> oci://<registry>/<org>/<my-bundle-name>:<tag>`), but the **child** `aicr-stack` Application then references `chart: aicr-bundle` at `repoURL: oci://<registry>/<org>`, which does not exist in that registry.

## Reproduction

```bash
aicr bundle \
  --recipe recipe.yaml \
  --deployer argocd-helm \
  --output oci://<registry>/<org>/<my-bundle-name>:<tag>

helm install <release> oci://<registry>/<org>/<my-bundle-name>:<tag> \
  --namespace argocd \
  --set repoURL=oci://<registry>/<org>
```

## Generated `aicr-stack` Application

```yaml
spec:
  source:
    repoURL: oci://<registry>/<org>
    chart: aicr-bundle           # ← hardcoded; actual artifact is <my-bundle-name>
    targetRevision: <tag>
```

ArgoCD error (identical class of failure as the `gpu-operator-pre` bug — same root: wrong OCI coordinates):

```
Failed to load target state: failed to generate manifest for source 1 of 1:
... <registry>/<org>:<tag>: not found
```

(Because `aicr-bundle` doesn't exist under that org, ArgoCD falls back to trying the `repoURL` as a full artifact and that also fails.)

## Suggested Fix

Derive the parent chart name from the `--output` OCI reference the user provided. The bundler already parses this URL to perform the push; the same parsed name should flow into the app-of-apps Application template instead of the constant `aicr-bundle`.

If users genuinely want to override the parent chart name independently of the OCI artifact name, expose a flag (e.g. `--parent-chart-name`) — but the default should match what was actually pushed.

## Impact

- Every user who picks a non-default OCI artifact name hits this
- Combined with #(gpu-operator-pre path-vs-chart bug), the bundle effectively cannot bootstrap on a fresh cluster without manual patching
- The `aicr-stack` app-of-apps is the entry point: when it fails to sync, all child Applications that it would have managed are orphaned

[...]

PR #1035, opened 2026-05-26
fix(bundler): fix bugs -  assemble full OCI artifact in path-based child apps; quote chart name (#1034)

### LINKED ISSUE #1034: argocd-helm OCI bundle: path-based child Applications miss .Chart.Name; Chart.yaml emits unquoted name
## Summary

Two correctness bugs in the argocd-helm bundler's OCI publishing path:

1. **P1 (regression from #1032)** — Path-based child Applications resolve to a wrong artifact URL when the bundle is published to OCI.
2. **P2** — `Chart.yaml` emits an unquoted chart name; valid OCI artifact paths whose last segment is a YAML reserved scalar produce a chart Helm rejects.

Both reproduce against `origin/main` at HEAD `51eb4b5f`.

## P1 — Path-based child Applications miss `.Chart.Name` in repoURL

**Where:** `pkg/bundler/deployer/argocdhelm/argocdhelm.go:699-704` in `injectValuesIntoSingleSource`.

**What's wrong:** #1032 changed the contract so users pass only the parent namespace (e.g., `oci://reg/org`) and the parent Application appends `.Chart.Name` to assemble the full artifact reference (e.g., `oci://reg/org/my-bundle`). The parent-App template at line 407-408 implements this correctly:

```yaml
repoURL: {{ required "..." .Values.repoURL | quote }}
chart: {{ .Chart.Name | quote }}
```

But the path-based child-App template at line 703 only emits a `repoURL`, no `chart` field (path-based sources don't have one — Argo CD's generic OCI source uses `repoURL` directly as the full artifact). The template still emits just `.Values.repoURL`:

```go
Value: ` + "`" + `{{ required "...do NOT include the chart name..." .Values.repoURL }}` + "`" + `,
```

So a bundle pushed as `oci://reg/org/my-bundle:v1` with manifest-only or mixed-local-chart children (e.g., `gpu-operator-post`, `nodewright-customizations` raw manifests) installs with `--set repoURL=oci://reg/org`, the parent Application resolves correctly, but every path-based child Application resolves to `oci://reg/org:v1` — an artifact that doesn't exist — and fails to sync. The instruction text in the `required` message and the template body are out of sync.

**Severity:** High — silently broken bundles for any recipe with raw manifests under the new (post-#1032) OCI contract.

**Proposed fix:** Assemble the full artifact reference in the path-based child template too, matching the parent-App pattern:

```go
Value: ` + "`" + `{{ required "..." .Values.repoURL }}/{{ .Chart.Name }}` + "`" + `,
```

## P2 — Unquoted `name:` in Chart.yaml mis-parses YAML reserved scalars

**Where:** `pkg/bundler/deployer/argocdhelm/argocdhelm.go:944` in `writeChartYAML`.

**What's wrong:** `(*Reference).ChartName()` (`pkg/oci/reference.go:152`) returns `path.Base(Repository)`. For valid OCI repository names like `oci://reg/org/null:v1`, `oci://reg/org/false:v1`, `oci://reg/org/123:v1`, or any other YAML reserved scalar (`null`, `true`, `false`, `yes`, `no`, `on`, `off`, `~`, numeric-looking strings), `path.Base` returns the value verbatim. `writeChartYAML` then emits unquoted:

```go
fmt.Fprintf(&buf, "name: %s\n", name)
```

Producing `name: null` (or `name: 123`, etc.). Helm's YAML parser interprets these as non-string scalars and `chart.Metadata.Name` becomes empty (or fails to unmarshal), so the subsequent `helm package` / `helm push` step rejects the chart with "chart.metadata.name is required" or a YAML type error.

**Severity:** Low — requires an unusual but valid OCI path. No known reports in the wild; defense-in-depth fix.

**Proposed fix:** Quote the value safely. `%q` produces a YAML-safe double-quoted scalar for all printable ASCII:

```go
fmt.Fprintf(&buf, "name: %q\n", name)
```

Same defense should be applied to the `version:` line for symmetry (a version string like `"1.0"` would parse as a float and fail similarly).

## Acceptance criteria

- [ ] Path-based child Application template appends `.Chart.Name` so the rendered `repoURL` is the full OCI artifact reference (parent namespace + chart name).
- [ ] `writeChartYAML` quotes both `name` and `version` so YAML reserved scalars round-trip as strings.
- [ ] Unit test: path-based template renders `{{ .Values.repoURL }}/{{ .Chart.Name }}` (or a Helm-rendered fixture asserting the assembled URL).
- [ ] Unit test: `writeChartYAML` produces YAML that unmarshals `name` back as a string for inputs like `"null"`, `"true"`, `"123"`.

## Related

- #1019 — the original "argocd-helm chart name must match OCI artifact" thread.
- #1032 — the fix for #1019 that introduced the path-based-children regression.

Found via Codex review of the AICR codebase.

yuanchen8911:
## Summary

Two argocd-helm OCI publishing bugs against the post-#1032 contract:

1. **P1** — Path-based child Applications were resolving to a non-existent artifact because their `spec.source.repoURL` template never appended `.Chart.Name`. Under #1032's contract (`--set repoURL=<parent namespace>`), the parent App appends `.Chart.Name` via its separate `source.chart` field, but path-based children have no `chart` field — Argo CD's generic OCI source resolves `repoURL` directly. Bundles with manifest-only / mixed-local-chart children silently failed to sync.
2. **P2** — `writeChartYAML` emitted `name:` and `version:` unquoted. Valid OCI artifact paths whose last segment is a YAML reserved scalar (`null`, `true`, `false`, `yes`, `no`, numeric strings) produced `name: null` etc., which Helm's YAML parser interprets as the non-string scalar, failing the chart with "chart.metadata.name is required".

## Motivation / Context

Found via Codex review of `origin/main` at `51eb4b5f`. P1 is a regression introduced today by #1032 (which landed the parent-App + bundle-chart-name fix for #1019 but left the path-based child template on the old `--set repoURL=<full bundle URL>` contract). P2 is a latent corner case that the same review surfaced.

Fixes: #1034
Related: #1032 (the parent-App contract change this PR completes), #1019 (original chart-name bug), #965 (DRA stale-NVML mitigation, unrelated but in the same OCI chain)

## Type of Change

## Component(s) Affected

- [x] Bundlers (`pkg/bundler`, `pkg/component/*`) — `pkg/bundler/deployer/argocdhelm/argocdhelm.go`

## Implementation Notes

**P1 — `injectValuesIntoSingleSource` (line 703).** Append `/{{ .Chart.Name }}` to the rendered `repoURL` value so the assembled URL matches the artifact the parent's `repoURL/chart:tag` triple resolves to. Updated the `required` error message to reflect that the *path-based template itself* is doing the chart-name appending (the parent App also appends it, but via the separate `source.chart` field; the wording was misleading on the child).

**P2 — `writeChartYAML` (line 944).** Emit `name` and `version` via `fmt.Fprintf` with `%q`. OCI artifact path segments are constrained to printable ASCII by the docker reference grammar, which is exactly the safe charset for Go's `%q` (so the rendered YAML is always a clean double-quoted string).

## Testing

```bash
unset GITLAB_TOKEN ; make qualify
# All 22 chainsaw tests pass.
# Coverage: 76.9% > 75% floor.
# Vulnerability scan: clean.
```

**Tests added:**

- `TestInjectValuesIntoSingleSource_AppendsChartName` — asserts the rendered child `source.repoURL` value is `…/{{ .Chart.Name }}` (with the original `required` directive intact).
- `TestWriteChartYAML_QuotesYAMLReservedScalarsAsName` — table-driven case across `null`, `true`, `false`, `123`, `yes`, and a normal name. For each, writes `Chart.yaml` and asserts `yaml.Unmarshal` round-trips `name` back as the expected string.

**Tests updated:**

- `TestHelmTemplate_RendersWithSetRepoURL` — exercises the new contract (`--set repoURL=oci://reg/org`, parent namespace only) end-to-end via real `helm template`. Asserts the parent App's `spec.source.repoURL` equals the parent namespace and the path-based child's `spec.source.repoURL` equals parent-namespace + `/aicr-bundle`. This test was the canary: it was passing under the old contract (full bundle URL) because the parent App's `source.chart` was being ignored at install time; my fix surfaces the contract drift.
- `TestGenerate_CustomChartName` and the existing `version: 2.5.0` assertion updated to expect quoted scalars.
- Golden fixtures (`testdata/helm_and_manifest_only/` and `testdata/mixed_component/`) regenerated: child `repoURL` lines and `Chart.yaml` `name`/`version` lines now reflect the fix.

## Risk Assessment

- [x] **Low** — narrow scope, two-line code change plus assertion updates. No new external dependencies, no surface-area expansion. Bundles produced by this PR are correct for the new contract; bundles produced by the pre-#1032 codepath are unaffected (the child template change is additive — `/{{ .Chart.Name }}` only takes effect under the new `--set repoURL=<parent namespace>` contract).

**Rollout notes:** Anyone who picked up #1032's contract change since today (2026-05-26) needs this fix before publishing argocd-helm bundles with raw-manifest children to OCI. Existing bundles already pushed to OCI work as-is; the fix takes effect on the next `aicr bundle --output oci://…` run.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — via `make qualify`
- [x] Linter passes (`make lint`) — via `make qualify`
- [x] I updated docs if user-facing behavior changed — no user-facing surface change; the error-message text inside the `required` directive is the only "doc" touched

### DIFF
--- a/pkg/bundler/deployer/argocdhelm/argocdhelm.go
+++ b/pkg/bundler/deployer/argocdhelm/argocdhelm.go
@@ -696,11 +696,23 @@ func injectValuesIntoSingleSource(app map[string]any, overrideKey string) error 
 	// in single quotes — that's the only YAML scalar style that doesn't
 	// require escaping the embedded double quotes inside `required "..."`,
 	// which would corrupt Helm's parsing of the template.
+	//
+	// Path-based child Applications have no `chart` field — Argo CD's
+	// generic OCI source uses `repoURL` directly as the full artifact
+	// reference and then resolves `path` inside it. We therefore append
+	// .Chart.Name here ourselves so the rendered value is
+	// `<parent namespace>/<chart name>` (e.g. `oci://reg/org/my-bundle`),
+	// matching the artifact the parent Application's `repoURL/chart:tag`
+	// triple resolves to. Without the append, --set repoURL=<parent
+	// namespace> (the contract every other site in this file documents)
+	// produces a child source pointing at `oci://reg/org:tag` — an
+	// artifact that does not exist — and the child Application fails
+	// to sync. See issue #1034.
 	source["repoURL"] = &yaml.Node{
 		Kind:  yaml.ScalarNode,
 		Tag:   yamlStringTag,
 		Style: yaml.SingleQuotedStyle,
-		Value: `{{ required "repoURL is required: pass --set repoURL=<parent namespace> (e.g., oci://<registry>/<path>) — do NOT include the chart name; the parent App appends .Chart.Name to assemble the full OCI artifact reference" .Values.repoURL }}`,
+		Value: `{{ required "repoURL is required: pass --set repoURL=<parent namespace> (e.g., oci://<registry>/<path>) — do NOT include the chart name; this template appends .Chart.Name to assemble the full OCI artifact reference" .Values.repoURL }}/{{ .Chart.Name }}`,
 	}
 	source["targetRevision"] = &yaml.Node{
 		Kind:  yaml.ScalarNode,
@@ -939,12 +951,20 @@ func writeChartYAML(outputDir, name, version string) (string, int64, error) {
 		return "", 0, err
 	}
 
+	// Quote name and version so OCI artifact paths whose last segment is
+	// a YAML reserved scalar ("null", "true", "false", "yes", "no",
+	// "123", etc.) round-trip as strings instead of getting reinterpreted
+	// by Helm's YAML parser as the underlying scalar type and producing
+	// a chart with an empty Metadata.Name (or a type-mismatch error).
+	// fmt %q emits a Go-quoted string that is also valid YAML for all
+	// printable ASCII; OCI artifact path segments are constrained to
+	// that charset by the docker reference grammar. See issue #1034.
 	var buf strings.Builder
 	buf.WriteString("apiVersion: v2\n")
-	fmt.Fprintf(&buf, "name: %s\n", name)
+	fmt.Fprintf(&buf, "name: %q\n", name)
 	buf.WriteString("description: AICR deployment bundle with dynamic install-time values\n")
 	buf.WriteString("type: application\n")
-	fmt.Fprintf(&buf, "version: %s\n", version)
+	fmt.Fprintf(&buf, "version: %q\n", version)
 
 	content := buf.String()
 	if writeErr := os.WriteFile(chartPath, []byte(content), 0600); writeErr != nil {
--- a/pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
+++ b/pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
@@ -199,8 +199,11 @@ func TestGenerate(t *testing.T) {
 				if err != nil {
 					t.Fatalf("failed to read Chart.yaml: %v", err)
 				}
-				if !strings.Contains(string(content), "version: 2.5.0") {
-					t.Error("Chart.yaml should contain version: 2.5.0")
+				// writeChartYAML quotes the version scalar so YAML
+				// reserved scalars (e.g. "1.0", "null") round-trip as
+				// strings; see issue #1034.
+				if !strings.Contains(string(content), `version: "2.5.0"`) {
+					t.Errorf("Chart.yaml should contain version: \"2.5.0\", got:\n%s", string(content))
 				}
 			},
 		},
@@ -1044,12 +1047,19 @@ func TestHelmTemplate_RendersWithSetRepoURL(t *testing.T) {
 		t.Fatalf("Generate() error = %v", err)
 	}
 
-	const wantRepoURL = "oci://example.test/myorg/aicr-bundle"
+	// Under the post-#1032 contract, --set repoURL carries the parent
+	// namespace only — the parent App appends .Chart.Name via its
+	// source.chart field, and (per #1034) the path-based child template
+	// appends /<chart-name> directly into its source.repoURL so Argo
+	// CD's generic OCI source resolves to the same artifact.
+	const setRepoURL = "oci://example.test/myorg"
+	const wantParentRepoURL = setRepoURL
+	const wantChildRepoURL = setRepoURL + "/" + DefaultChartName
 	const wantTagName = "v9.9.9-render-test"
 	cmdCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
 	cmd := exec.CommandContext(cmdCtx, "helm", "template", "test-release", outputDir, //nolint:gosec // controlled args
-		"--set", "repoURL="+wantRepoURL,
+		"--set", "repoURL="+setRepoURL,
 		"--set", "targetRevision="+wantTagName,
 	)
 	out, err := cmd.CombinedOutput()
@@ -1090,8 +1100,8 @@ func TestHelmTemplate_RendersWithSetRepoURL(t *testing.T) {
 	if !ok {
 		t.Fatalf("rendered output missing parent Application 'aicr-stack'\noutput:\n%s", out)
 	}
-	if parent.Spec.Source.RepoURL != wantRepoURL {
-		t.Errorf("parent App repoURL: got %q, want %q", parent.Spec.Source.RepoURL, wantRepoURL)
+	if parent.Spec.Source.RepoURL != wantParentRepoURL {
+		t.Errorf("parent App repoURL: got %q, want %q", parent.Spec.Source.RepoURL, wantParentRepoURL)
 	}
 	if parent.Spec.Source.Chart != DefaultChartName {
 		t.Errorf("parent App chart: got %q, want %q", parent.Spec.Source.Chart, DefaultChartName)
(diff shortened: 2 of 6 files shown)


pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -696,11 +696,23 @@ func injectValuesIntoSingleSource(app map[string]any, overrideKey string) error
...
+	// namespace> (the contract every other site in this file documents)
+	// produces a child source pointing at `oci://reg/org:tag` — an
+	// artifact that does not exist — and the child Application fails
+	// to sync. See issue #1034.
 	source["repoURL"] = &yaml.Node{
 		Kind:  yaml.ScalarNode,
 		Tag:   yamlStringTag,
 		Style: yaml.SingleQuotedStyle,
-		Value: `{{ required "repoURL is required: pass --set repoURL=<parent namespace> (e.g., oci://<registry>/<path>) — do NOT include the chart name; the parent App appends .Chart.Name to assemble the full OCI artifact reference" .Values.repoURL }}`,
+		Value: `{{ required "repoURL is required: pass --set repoURL=<parent namespace> (e.g., oci://<registry>/<path>) — do NOT include the chart name; this template appends .Chart.Name to assemble the full OCI artifact reference" .Values.repoURL }}/{{ .Chart.Name }}`,
```
mchmarny, line 715: nit: if a user passes `--set repoURL=oci://reg/org/` (trailing slash, easy to copy-paste from a registry UI), this renders `oci://reg/org//aicr-bundle` — Argo CD will fail to sync with no obvious hint that the slash is the culprit. Cheap fix:

```
Value: `{{ required "..." .Values.repoURL | trimSuffix "/" }}/{{ .Chart.Name }}`
```

Same robustness for the parent App's `source.chart`/`source.repoURL` pair would be worth a follow-up if it doesn't already do this. Not a blocker.

pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
```diff
@@ -1272,3 +1284,100 @@ func assertGolden(t *testing.T, outDir, goldenDir, relPath string) {
...
+				t.Errorf("Chart.yaml name = %q (after YAML unmarshal), want %q\nraw:\n%s",
+					parsed.Name, tt.wantUnmarsh, content)
+			}
+			if parsed.Version != "1.0.0" {
+				t.Errorf("Chart.yaml version = %q, want \"1.0.0\"\nraw:\n%s",
+					parsed.Version, content)
+			}
+		})
+	}
+}
```
mchmarny, line 1383: nit: the table covers the "name is a YAML reserved scalar" case but the same risk applies to `version` — a user-supplied version like `"1.0"` (no patch) renders as the YAML float `1` once Helm reparses, which is what makes `version: "%q"` defense-in-depth and not just cosmetic. Worth adding a case here (e.g., `chartName: "aicr-bundle", version: "1.0"`) so the version branch of `writeChartYAML` is exercised by the same test that documents the rationale.

mchmarny (approved):
Both fixes look correct and the test coverage is thorough.

**P1 (path-based child `repoURL`):** Append of `/{{ .Chart.Name }}` matches the artifact the parent App's `repoURL/chart:tag` triple resolves to (`g.chartName()` flows into both the parent's `source.chart` and the chart's `Chart.yaml`, so they can't drift). The updated `required` message accurately describes child-template behavior. The `TestHelmTemplate_RendersWithSetRepoURL` change is the right canary — it now exercises the new contract end-to-end via real `helm template`.

**P2 (Chart.yaml quoting):** `%q` is safe here because OCI artifact path segments are constrained to printable ASCII by the docker reference grammar, which is exactly Go's `%q` safe charset. Defense-in-depth quoting for `version` is also sound — Helm's semver validator would reject most reserved scalars downstream, but failing at `Chart.yaml` parse time with a clear YAML round-trip is friendlier than a Helm push error.

Two nits inline (trailing-slash robustness on the rendered template, and adding a `version` case to the reserved-scalar table) — neither blocks merge. All 22 chainsaw tests and the rest of CI are green.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
yuanchen8911: Addressed in d8adac98 — the rendered value is now `{{ required "..." .Values.repoURL | trimSuffix "/" }}/{{ .Chart.Name }}`. So `--set repoURL=oci://reg/org/` (with trailing slash) renders cleanly as `oci://reg/org/aicr-bundle` instead of the double-slash artifact path. Goldens regenerated.

Good catch on the parent App's `source.chart` / `source.repoURL` pair too — I'll file that as a follow-up issue since the parent template would need its own `trimSuffix` treatment and the same caveat applies (silent Argo-CD sync failure with no operator hint).

pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
yuanchen8911: Addressed in d8adac98 — restructured the table to take an explicit (chartName, chartVersion) pair and added rows for float-looking (`"1.0"`) and numeric-looking (`"123"`) versions alongside the existing reserved-scalar name rows. The assertion now round-trips both fields through `yaml.Unmarshal` and checks each against its exact input string, so the `version` branch of `writeChartYAML`'s `%q` wrap is exercised by the same test that documents the rationale.

Issue #745, opened 2026-05-05
Supply-chain provenance audit per component


mchmarny:
Parent: #739

## Summary

Audit each component's container images for supply-chain provenance: cosign signatures, SBOMs, and SLSA provenance attestations. Produce a matrix that informs verification policy and customer security review.

## Motivation

AICR pulls images from many public registries with varying provenance maturity. Some publish signed images with SBOMs (e.g., NVIDIA, Kubernetes SIG); others may not. Customers performing security review need to know which can be cryptographically verified and which cannot.

## Proposed approach

For each component (and each image it deploys, per BOM output), check:

| Attribute | How to check |
|-----------|--------------|
| Cosign signature | `cosign verify <image>` against publisher's pubkey |
| SBOM availability | `cosign download sbom <image>` or registry attachment |
| SLSA provenance | `cosign verify-attestation --type slsaprovenance` |
| Vulnerability scan results | grype/trivy reference |

Output: a matrix in `docs/security/provenance-matrix.md` showing per-component status, with links to verification commands.

## Success criteria

- Every component has documented provenance status.
- Verification commands are reproducible.
- Gaps (unsigned images, missing SBOMs) are flagged for follow-up with upstream maintainers.
- Optionally: a `cosign verify` step in our deployment validation for components where signatures are available.

## Notes

- Depends on BOM tooling (separate issue) for the canonical image list.
- AICR's own validator/runtime images are out of scope (covered by goreleaser/SLSA already).
- This is an audit/documentation issue first; enforcement (e.g., admission policy) is a follow-up if warranted.

PR #1037, opened 2026-05-26
feat(recipes): pin nodewright-customizations packages by digest

### LINKED ISSUE #1031: [Feature]: Move nodewright-customizations to use sha for their container value
### Prerequisites

### Feature Summary

Related: #745 
Nodewright supports the package key: `containerSHA: sha256:....` in addition to version which is used to track upgrade/downgrade logic. Update the customizations to use the sha to better address supply chain assurance.

### Problem/Use Case

Supply chain attacks for nodewright-customizations

### Proposed Solution

Update all recipes/nodewright-customizations to use the proper sha for their container

 Example:
```yaml
  packages:
    my-pkg:
      image: ghcr.io/nvidia/nodewright/example
      version: 1.2.3
      containerSHA: sha256:1234...cdef
```

### Success Criteria

- All nodewright-customizations use the containerSHA

### Alternatives Considered

_No response_

### Component

Recipe engine / data

### Priority

Nice to have

_No response_

_No response_

Yes, I can open a PR

ayuskauskas:
## Summary

Pin every Skyhook Package in `nodewright-customizations` by `containerSHA: sha256:...` and teach the BOM extractor to recognize that field as the OCI digest, removing the per-image digest-pin test exemptions for the three nodewright-packages refs.

## Motivation / Context

The Skyhook Package CRD does not accept an `@sha256:...` digest inside its `image:` value, but it does carry an explicit `containerSHA` field for the same purpose. We were exempting these refs from the ADR-006 layer 2 digest-pin gate in `recipes/manifest_images_test.go`. Now that the CRD field is populated, the gate can enforce the pin directly.

Fixes: #1031
Related: #745

## Type of Change

## Component(s) Affected

## Implementation Notes

- `recipes/components/nodewright-customizations/manifests/tuning.yaml` and `tuning-gke.yaml`: added `containerSHA` to every package entry (`nvidia-setup-kernel`, `nvidia-tuned`, `nvidia-setup-full`, and the GKE `tuning` package). Digests resolved via `crane digest <ref>`:
  - `ghcr.io/nvidia/nodewright-packages/nvidia-setup:0.2.2` → `sha256:76913a5d…`
  - `ghcr.io/nvidia/nodewright-packages/nvidia-tuned:0.3.0` → `sha256:cc99c8c0…`
  - `ghcr.io/nvidia/nodewright-packages/nvidia-tuning-gke:0.1.2` → `sha256:6671d49f…`
- `pkg/bom/extract.go`: `walkForImages` now also collects a sibling `containerSHA` scalar, and a new `appendContainerSHA` helper folds it onto the combined CRD-style ref as `@<sha>`. An in-line `@digest` in `image` continues to win — `containerSHA` never overwrites or double-appends. This keeps the change local to the extractor: callers (BOM, digest-pin test, PURL) consume the resulting `@sha256:...` without any special-casing.
- `recipes/manifest_images_test.go`: removed the three `nodewright-packages/*` entries from `imageDigestExemptions`. `skyhook-packages/shellscript:1.1.1` stays exempt — upstream does not currently surface a `containerSHA` for it.
- `docs/user/container-images.md`: regenerated via `make bom-docs`; the three nodewright-packages refs now appear as `:tag@sha256:...`.

## Testing

```bash
go test -race ./pkg/bom/... ./pkg/bundler/... ./recipes/...
golangci-lint run -c .golangci.yaml ./pkg/bom/... ./recipes/...
make bom-docs
```

All green. Two new test cases were added in `pkg/bom/extract_test.go`:

- `Skyhook Package with containerSHA sibling folds in as digest` — verifies the fold-in path.
- `inline @digest in image takes precedence over containerSHA sibling` — guards the precedence rule so a stale `containerSHA` cannot silently overwrite an explicit in-line digest.

Coverage delta: `pkg/bom: 89.7% → 90.0% (+0.3%)`.

The `TestComponentManifestImagesAreDigestPinned` log now lists only the four expected exemptions (three Mellanox NicClusterPolicy refs + shellscript) — the three nodewright-packages refs are accepted as digest-pinned.

## Risk Assessment

**Rollout notes:** Pure supply-chain hardening. The image tags are unchanged, so resolved manifests deploy the same artifacts they did before — Skyhook now additionally validates the digest at install time. No migration needed.

## Checklist


pkg/bom/extract.go
```diff
@@ -177,6 +182,32 @@ func combineCRDTriplet(image, repository, version string) string {
...
+// `sha256:<hex>`), matching the Skyhook CRD convention and the
+// `@<digest>` form ParseImageRef already understands.
+func appendContainerSHA(image, sha string) string {
+	if sha == "" {
+		return image
+	}
+	if strings.Contains(image, "@") {
+		return image
+	}
+	return image + "@" + sha
```
mchmarny: Clean precedence rules — `@`-in-image wins, empty sha is a no-op, otherwise append. The two test cases (fold-in + precedence) cover both branches well, and the in-line `@digest` check uses `strings.Contains` so an alias-resolved value that already includes a digest is also handled correctly.

nit: the function will happily append a malformed `sha` if the YAML carries something like `containerSHA: not-a-digest` — the result is `image:tag@not-a-digest`, which `isLikelyImage` still accepts. The existing digest-pin test in `recipes/manifest_images_test.go` is the safety net for this (it regexes `@sha256:[a-f0-9]{64}`), so the BOM would render the bogus ref but CI would fail. Worth a sentence in the doc comment to flag that validation is downstream, or a cheap `strings.HasPrefix(sha, "sha256:")` guard here if you prefer fail-loud at extraction. Non-blocking either way.

mchmarny (dismissed):
Clean supply-chain hardening change. The extractor logic for `containerSHA` matches the existing CRD-triplet pattern (`image`/`repository`/`version` sibling scalars), preserves precedence (inline `@digest` wins, empty is a no-op, otherwise append), and the two new test cases lock both branches in. Removing the three nodewright-packages exemptions from `recipes/manifest_images_test.go` and keeping just the shellscript carve-out is the right call.

CI: ClamAV, grype, and all completed Tier-1 matrix jobs are green; a handful of flux-oci variants are still running but the extractor change is unrelated to those code paths. CodeQL conclusion is "neutral" (skipped, normal here).

One non-blocking nit on `appendContainerSHA` — see inline.

pkg/bom/extract.go
ayuskauskas: Took the fail-loud path in `e881b3ed`. `appendContainerSHA` now validates against `^sha256:[a-f0-9]{64}$` and returns an error; `walkForImages` and `ExtractImagesFromYAML` propagate it. Three regression cases cover non-sha256 prefix, short hex, and uppercase hex (OCI digest grammar is lowercase, so e.g. `sha256:CC99…` is rejected). Net effect: a typo or `--set`-injected bogus digest in a Skyhook Package fails BOM extraction with the offending value in the message, instead of silently shipping a malformed ref into the SBOM/PURL output. Thanks for catching it.

mchmarny (approved):

PR #1047, opened 2026-05-26
fix(kwok): drop recipe suffix from argocd-helm-oci in-cluster repoURL

yuanchen8911:
## Summary

Fix the KWOK matrix wrapper script so the `argocd-helm-oci` deployer tier passes the parent-namespace-only repoURL that the Argo CD parent Application template (PR #1032 / #1035) actually expects.

## Motivation / Context

PR #1032 ("configurable parent Application name") and PR #1035 ("assemble full OCI artifact in path-based child apps") changed the argocd-helm parent Application template to append `.Chart.Name` itself via the `source.chart` field. The wrapper script in `kwok/scripts/validate-scheduling.sh` was not updated to match — it kept passing the **full** bundle URL (including the per-recipe chart name) as `--set repoURL=…`. The parent App template then appended `.Chart.Name` on top, producing a doubled recipe segment:

```
oci://registry.aicr-registry.svc.cluster.local:5000/aicr/<recipe>/<recipe>:<tag>
                                                       ^^^^^^^^ ^^^^^^^^
                                                       passed in   appended
                                                       repoURL     by Argo
```

The OCI artifact actually lives at `oci://…/aicr/<recipe>:<tag>` (single segment), so the doubled form 404s. Argo CD's repo-server logs the resolved digest lookup failure:

```
failed to resolve revision "0.0.0-<sha>-argocd-helm-oci":
cannot get digest for revision 0.0.0-<sha>-argocd-helm-oci:
registry.aicr-registry.svc.cluster.local:5000/aicr/<recipe>/<recipe>:0.0.0-<sha>-argocd-helm-oci: not found
```

The `argocd-helm-oci` parent Application never syncs, `gpu-operator-post` (a child of the parent App) never reconciles, and the matrix job times out on `GitOps sync timeout strike 1/3`.

**Blast radius:** every PR with Tier-1 KWOK coverage on `argocd-helm-oci` fails this way — currently 10+ unrelated services (aks, aks-inference, aks-training, eks-inference, gke-cos, gke-cos-inference, gke-cos-training, kind-inference, lke-inference, oke-inference). The `argocd-oci`, `flux-oci`, and plain `helm` deployers are unaffected because their resolution paths don't go through `helm install --set repoURL=…` against the parent App template.

Fixes the `argocd-helm-oci` CI breakage currently blocking PR #1046 (rtx-pro-6000 EKS overlays) and any other open PR.

Fixes: N/A (no dedicated tracking issue)
Related: #1032, #1035, #1046

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: KWOK deployer matrix test wrapper (`kwok/scripts/validate-scheduling.sh`)

## Implementation Notes

Single change to the `argocd-helm-oci` branch of `generate_bundle()`:

```diff
- OCI_IN_CLUSTER_REF="oci://registry.aicr-registry.svc.cluster.local:5000/aicr/${recipe}"
+ OCI_IN_CLUSTER_REF="oci://registry.aicr-registry.svc.cluster.local:5000/aicr"
```

Plus an explanatory comment block citing PR #1032 / #1035, and a log-line tweak so the printed pull URL shows the full path (`${in_cluster_repo}/${recipe}:${tag}`) — i.e., the URL Argo CD will actually dereference — instead of the truncated parent-namespace URL.

**Why not also change the `flux-oci` branch (line 844):**
The `flux-oci` branch stays on `…/aicr/${recipe}` because Flux's source-controller doesn't go through the same parent-App template path. The Flux deployer doesn't bake `repoURL` into the bundle — the OCIRepository CR is applied at deploy time with the full path baked in directly. Its current value already matches the pushed artifact at `aicr/${recipe}:${tag}` and Flux works correctly today. Touching it would break a working path.

## Testing

```bash
# Shell syntax check
bash -n kwok/scripts/validate-scheduling.sh

# Static analysis
shellcheck kwok/scripts/validate-scheduling.sh
# Only pre-existing SC1091 info on line 101 (unrelated `source` directive)
```

Both pass. **Full `make qualify` was skipped** because this is an infra-only change to a CI shell script — Go tests, e2e, vulnerability scan, and yamllint cannot regress from a bash one-liner. Per `CLAUDE.local.md`'s doc-only/infra-only carve-out, scoped checks are the appropriate gate here.

**Real validation will come from CI itself:** the `argocd-helm-oci` matrix jobs in this PR's run should turn green where the rtx-pro-6000 PR (#1046) currently shows 10+ failures, confirming the contract is restored.

## Risk Assessment

- [x] **Low** — Isolated single-line change to CI test infrastructure, easy to revert (replace `aicr` back with `aicr/${recipe}` on line 758).

**Rollout notes:** No runtime impact — affects only the KWOK CI matrix test driver. After merge, every open PR's `argocd-helm-oci` Tier-1 jobs should pass when rebased onto the new main.

## Checklist

- [x] Tests pass locally (`bash -n` + `shellcheck` — infra-only change, full `make qualify` not applicable)
- [x] Linter passes (shellcheck, no new warnings)
- [ ] I added/updated tests for new functionality (N/A — this is a contract realignment between the wrapper script and the existing parent App template)
- [ ] I updated docs if user-facing behavior changed (N/A — CI internal)


mchmarny (approved):

yuanchen8911:
Codex caught a real regression — thank you. The `OCI_IN_CLUSTER_REF` assignment at line 769 lives inside the shared `argocd-oci|argocd-helm-oci)` case branch (label at [line 722](https://github.com/NVIDIA/aicr/blob/c70e84727b48bbb7220765e3788342ddfc8f1fce/kwok/scripts/validate-scheduling.sh#L722)), so the change strips the per-recipe segment for both lanes. With this PR as-is, `argocd-oci` would start failing — the root app resolves `repoURL = oci://…/aicr` to `oci://…/aicr:\${tag}`, but the artifact actually lives at `oci://…/aicr/\${recipe}:\${tag}`.

The fix is to make the assignment per-lane, matching the existing `[[ \"\$DEPLOYER\" == \"argocd-helm-oci\" ]]` conditionals at lines 742 and 763:

\`\`\`bash
if [[ \"\$DEPLOYER\" == \"argocd-helm-oci\" ]]; then
    OCI_IN_CLUSTER_REF=\"oci://registry.aicr-registry.svc.cluster.local:5000/aicr\"
else
    OCI_IN_CLUSTER_REF=\"oci://registry.aicr-registry.svc.cluster.local:5000/aicr/\${recipe}\"
fi
\`\`\`

And update the log line at 777 so each lane prints the URL it'll actually dereference (full path for `argocd-oci`, parent + `.Chart.Name` for `argocd-helm-oci`).

Amending the PR with this fix shortly.

yuanchen8911:
Amended in 0bb6b56f to make `OCI_IN_CLUSTER_REF` per-lane:

- `argocd-oci` keeps the full `aicr/${recipe}` form (its root `nvidia-stack` app references the artifact by full path).
- `argocd-helm-oci` uses the parent-only `aicr` form so the parent App template can append `.Chart.Name` itself.

The `log_info` line at 777 is also split per-lane so each prints the URL it will actually dereference. Verified `bash -n` and `shellcheck` clean (only the pre-existing SC1091 info on the unrelated line 101 `source` directive).

yuanchen8911:
Follow-up PR opened: #1048. Restores the per-recipe segment for the `argocd-oci` lane (which this PR's case-branch-wide change accidentally stripped along with `argocd-helm-oci`). Codex confirmed the per-lane conditional reads correctly.

Issue #962, opened 2026-05-19
Refactor KWOK argocd-* sync gate from bash/jq to Chainsaw


mchmarny:
## Background

PR #956 introduced a KWOK CI deployer matrix with three lanes: `helm`, `argocd-oci`, `argocd-helm-oci`. The argocd-* lanes need to wait for Argo CD reconciliation to complete and then assert that every Application reached a terminal pass state. This is currently implemented as a bash function (`wait_for_argocd_sync`) in `kwok/scripts/validate-scheduling.sh` (~150 lines of bash + jq).

The function works and handles three real-world pass states correctly (commit `cc2e87e1`):

1. `Synced + Healthy` — canonical pass
2. `Synced + Progressing` — KWOK pod-readiness simulation gap (per ADR-008)
3. `OutOfSync + Healthy + operationState.phase=Succeeded` — operator-mutation drift (gpu-operator's ClusterPolicy, nvidia-dra-driver-gpu's DeviceClass, etc.)

…and rejects two false-positive traps (sync that genuinely failed but resources happen to look healthy; OutOfSync + Degraded).

It works. It is also harder to read, harder to test in isolation, and not consistent with the rest of the AICR test surface (`tests/chainsaw/cli/bundle-variants/chainsaw-test.yaml`).

## Proposal

Replace the bash/jq sync-wait logic in `validate-scheduling.sh` (and the surrounding orchestration where it makes sense) with a Chainsaw scenario, matching the pattern already established under `tests/chainsaw/cli/`.

## Why a separate issue (not in #956)

This is an architectural refactor of how the KWOK lane drives its assertions, not a bug fix. PR #956 is already large (KWOK matrix scaffolding, two blocker fixes, hermeticity cleanup), and the Chainsaw migration trade-offs deserve their own review:

- Output format: bash logs stream into the `kwok-test` composite-action diagnostic dump; Chainsaw emits its own test-result format. The CI summary action needs adaptation.
- Exit-code semantics: `run-all-recipes.sh` reads `EXIT_ARGOCD_SYNC_TIMEOUT` (50) to implement the 3-strike rule. Chainsaw's exit code is fixed pass/fail.
- Diagnostics on failure: `dump_argocd_failures` collects per-Application status, repo-server logs, and application-controller logs in one place. Reproducing that in Chainsaw requires `outputs` blocks plus an explicit logs step.
[...]

PR #1062, opened 2026-05-27
fix(kwok): require Succeeded operationState in argocd sync gate

### LINKED ISSUE #1061: KWOK argocd-* sync gate races ahead of child Application materialization (regression from #1050)
## Summary

The KWOK `argocd-*` sync gate at `tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml` (introduced in #1050, merged at `d31e1f5d`) returns PASS within ~4–5 s — sometimes before Argo CD has finished its initial App-of-Apps sync and created the child `Application` resources. When the race is lost, the subsequent `validate-scheduling.sh` step runs against a half-deployed cluster and the percent-scheduled check trips on KIND-default kube-system pods.

**Severity: intermittent flake, not deterministic break.** Re-runs of the same SHA usually pass; the failure rate is low per-job but the 68-job workflow surfaces it at a meaningful rate.

## Evidence

The cleanest race-condition signature: **same commit, different outcome.**

| Run | When (UTC) | Branch | SHA | Outcome |
|---|---|---|---|---|
| `26523455927` (initial) | 2026-05-27 16:12 | `feat/973-…` | `8b5ef744` | ❌ |
| `26523455927` (re-run) | 2026-05-27 later | `feat/973-…` | `8b5ef744` (same) | ✅ |

Pod-count snapshots at the scheduling-validation step (recipe `eks-training`, deployer `argocd-helm-oci`):

| Run | When (UTC) | Chainsaw gate | Pods | Outcome |
|---|---|---|---|---|
| `26482828522` (pre-#1050 sync gate, my orig commit) | 2026-05-27 00:18 | bash gate | **35** scheduled, 0 unscheduled | ✅ |
| `26520380918` (`feat/bcm-service-type`, SHA `59ab2ff4`) | 2026-05-27 15:17 | **4.54 s** | **13** scheduled, 0 unscheduled | ✅ |
| `26521802328` (`feat/973-…`, SHA `6b7da4b9`) | 2026-05-27 15:42 | similar | **5** scheduled, 3 unscheduled | ❌ |
| `26523455927` (initial, SHA `8b5ef744`) | 2026-05-27 16:12 | **4.62 s** | **5** scheduled, 3 unscheduled | ❌ |
| `26523455927` (re-run, SHA `8b5ef744`) | 2026-05-27 later | (re-run) | (passes) | ✅ |

The 3 unscheduled pods on losing-the-race runs are the KIND-default `coredns` × 2 and `local-path-provisioner` × 1 — `kwok/scripts/validate-scheduling.sh:1489` taints the kind-control-plane node with `nfd-excluded=true:NoSchedule`, and those kube-system pods don't carry that toleration. When the AICR bundle's pods haven't deployed yet, the only remaining pods in the cluster are those 3 stragglers, so the percent-scheduled check trips.

### What this is NOT

- Not a deterministic break — same SHA can pass on retry.
- Not "every PR fails after #1050." Several PRs have passed first try since the merge (e.g., `renovate/busybox-1.37`, `feat/bcm-service-type` first run, and the retry on this PR).
- The earlier `feat/bcm-service-type` 15:17→15:36 transition I cited compared **different SHAs** (`59ab2ff4` → `e324ef97`); the pattern fits a race but isn't clean same-content evidence on that pair.

## Root Cause

In `tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml` (step `assert-all-applications-pass`, line ~122):

1. **The pass predicate doesn't require an Argo CD sync to have completed.** Arms 1 and 2 of the JMESPath only check `sync.status` and `health.status`; only arms 3 and 4 require `operationState.phase == 'Succeeded'`. For the root App-of-Apps, the controller can briefly set `sync.status=Synced` (manifest matches cluster) before running the sync operation that creates the child Applications.

2. **The selector has no minimum-count guard.** `assert` says "every `Application` in `argocd` namespace must pass" — when only the root exists, "every" = 1, so the assertion passes once the root alone matches. Children that haven't been created yet aren't in the match set.

The old bash gate (`wait_for_argocd_sync` in `validate-scheduling.sh`, replaced by #1050) didn't have this race because its polling loop took long enough that by the time it re-listed Applications, the children had materialized and joined the pass-check. Replacing it with a fast chainsaw assertion removed the implicit wait without adding an explicit one.

## Proposed Fix (minimum-invasive)

Add `operationState.phase == 'Succeeded'` to arms 1 and 2. All four arms then require Argo CD to have completed a sync operation, which for the App-of-Apps means children have been created and will be in the match set on subsequent polls.

```yaml
# tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml ~line 122
(operationState.phase == 'Succeeded'
 && ((sync.status == 'Synced' && health.status == 'Healthy')
     || (sync.status == 'Synced' && health.status == 'Progressing')
     || (sync.status == 'OutOfSync' && health.status == 'Healthy')
     || (sync.status == 'Synced' && health.status == 'Degraded'))): true
```

### Belt-and-suspenders option

Insert an intermediate step between `assert-root-app-present` and `assert-all-applications-pass` that waits for the root's `status.resources` list to be non-empty (or for at least N child Applications to exist). This makes the children-materialized invariant explicit instead of relying on `operationState.phase` propagation.

## Impact / Why Fix It Even Though Re-runs Clear It

- A flake at the workflow level translates to roughly half the runs hitting at least one failing tier-1 job (68 jobs × low per-job failure rate). That's enough to burn CI minutes, slow reviewer iteration, and erode trust in red CI on legitimate breakage.
- The fix removes the race rather than masking it — it doesn't make the chainsaw assertion wait longer; it makes its pass predicate require a real terminal state.

## Reproduction

Push any PR based on `main` ≥ `d31e1f5d` and watch any `Tier 1: <recipe> (argocd-helm-oci)` / `(argocd-oci)` job across multiple runs. Expect occasional failures with the signature "5 total pods, 3 unscheduled (coredns × 2, local-path-provisioner × 1)" that clear on retry of the same SHA.

## Related

- Introduced by: #1050 (`refactor(kwok): migrate argocd-* and flux-oci sync gate from bash to chainsaw`)
- Tracking: #962 (the migration this PR implemented)
- Observed on PR: #1033 (failed → retry passed on same SHA), `feat/bcm-service-type` (intermittent across SHAs)

yuanchen8911:
## Summary

Tighten the argocd-* KWOK chainsaw sync gate so all four arms of the pass-state predicate require `operationState.phase == 'Succeeded'`. Removes a race in which the assertion could PASS within ~4-5 s of `helm install` — before Argo CD had completed its initial App-of-Apps sync and created the child Applications.

Also add `tests/chainsaw/kwok/**` to the `KWOK Cluster Validation` workflow path filters so the scenarios actually re-run on edit (per Codex review of the first push).

## Motivation / Context

The chainsaw assertion at `tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml` says "every `Application` in the argocd namespace must satisfy the pass predicate." Right after `helm install` of an argocd-helm-oci bundle, the **only** Application present is the root App-of-Apps. The controller can briefly set the root's `sync.status=Synced` (manifest matches cluster) before running the sync operation that creates the children. In that window the assertion sees a single matching Application that satisfies arm 1 or arm 2 and returns PASS — children haven't been created yet, so they aren't in the match set.

Pre-#1050, the bash `wait_for_argocd_sync` loop avoided the race accidentally: its polling overhead took long enough that children materialized between iterations and joined the pass-check by the time it re-listed.

**Evidence the race is real** (issue #1061 has the full table):

| Run | SHA | Outcome |
|---|---|---|
| `26523455927` (initial) | `8b5ef744` | ❌ |
| `26523455927` (re-run) | `8b5ef744` (same) | ✅ |

Same commit, different result — textbook race-condition signature. Per-job failure rate is low (~1-3%), but with 68 parallel Tier 1 jobs the workflow surfaces it on a meaningful fraction of runs, costing CI minutes and re-runs.

Fixes: #1061
Related: #1050 (the chainsaw migration that introduced the race), #962 (the parent tracking issue)

## Type of Change

- [x] Build/CI/tooling (`.github/workflows/kwok-recipes.yaml` path filter expansion)

## Component(s) Affected

- [x] Other: KWOK CI scenarios (`tests/chainsaw/kwok/argocd-sync/`) and workflow path filters

## Implementation Notes

**Predicate change.** The JMESPath OR previously required `operationState.phase == 'Succeeded'` only on arms 3 and 4 (the relaxations for operator mutation and health-controller divergence). Arms 1 and 2 (`Synced + Healthy` / `Synced + Progressing`) could match purely on `sync.status` + `health.status`, which is what the App-of-Apps root briefly satisfies before its sync operation runs. Moving the `operationState.phase` guard out as a conjunct that gates all four arms ensures an Application only passes after Argo CD has completed at least one sync operation.

```yaml
# Before — only arms 3, 4 required Succeeded
((Synced && Healthy) || (Synced && Progressing) || (OutOfSync && Healthy && Succeeded) || (Synced && Degraded && Succeeded)): true

# After — all four arms require Succeeded
(Succeeded && ((Synced && Healthy) || (Synced && Progressing) || (OutOfSync && Healthy) || (Synced && Degraded))): true
```

**Why this isn't too strict.** Within the existing 5-minute assertion timeout, Argo CD's automated sync completes well within the budget (typical KWOK convergence: 30-90 s including child sync). The four sync/health combinations the old bash gate accepted remain valid; only the "pre-sync transient" window is now rejected.

**Workflow path filter.** `.github/workflows/kwok-recipes.yaml` previously gated KWOK on `recipes/**`, `kwok/**`, the workflow file, and `.github/actions/kwok-test/**`. PR #1050 introduced `tests/chainsaw/kwok/` as the new home for the migrated argocd / flux chainsaw scenarios, but didn't expand the filter — so edits to the scenarios silently skipped KWOK CI. The first push of this PR exhibited the symptom (0 KWOK jobs scheduled, caught by Codex review). Added `tests/chainsaw/kwok/**` to both `push.paths` and `pull_request.paths`.

**Flux sibling not affected.** `tests/chainsaw/kwok/flux-sync/chainsaw-test.yaml` asserts `Ready=True` conditions on `HelmRelease` and `Kustomization`, which already require their controllers to have completed reconciliation. No parallel change needed.

**Comment block updated** to explain why all four arms now share the `operationState.phase` requirement.

## Testing

- `yamllint tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml` — clean
- `chainsaw lint test -f tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml` — "The document is valid"
- `yamllint .github/workflows/kwok-recipes.yaml` — clean
- KWOK validation in CI is the canonical proof: with the path-filter fix in place, this PR now triggers the full 68-job Tier 1 matrix. If the original race survives, the same `5 pods total, 3 unscheduled` signature reappears; if the fix lands, the assertion converges normally or times out at 5 m with a clear "child apps not Succeeded" diagnostic.
- `make qualify` was not run end-to-end: change is test-infrastructure-only (no Go, no recipes, no docs-sidebar impact), and the relevant gates (yamllint, chainsaw lint) pass.

## Risk Assessment

- [x] **Low** — Single-file YAML predicate change + workflow path-filter expansion. The assertion still accepts all the same sync/health combinations as before, just gated behind a real terminal state. Worst case if the predicate is too strict: assertion times out at 5 m on a real-but-slow convergence rather than racing past it. Easy revert.

**Rollout notes:** No flag, no opt-in. The next KWOK run after merge exercises the new predicate. Existing passing PRs are unaffected. The path-filter expansion is purely additive.

## Follow-up

Codex review noted that the old bash gate had an explicit child-Application count guard, while this fix uses root-sync success as the materialization proxy. The proxy is sufficient for the reported flake (root cannot reach `Succeeded` without having created children), but a follow-up "wait for `root.status.resources` non-empty" step would restore full pre-#1050 parity for cases where the count itself is the invariant. Tracked for a follow-up if reviewers want belt-and-suspenders.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — N/A (no Go change)
- [x] Linter passes (`make lint`) — `yamllint` clean on both touched files, `chainsaw lint` valid
- [x] I added/updated tests for new functionality — N/A (the change IS in the test infrastructure)
- [x] I updated docs if user-facing behavior changed — N/A (no user-facing surface)

### DIFF
--- a/.github/workflows/kwok-recipes.yaml
+++ b/.github/workflows/kwok-recipes.yaml
@@ -21,6 +21,7 @@ on:
     paths:
       - 'recipes/**'
       - 'kwok/**'
+      - 'tests/chainsaw/kwok/**'
       - '.github/workflows/kwok-recipes.yaml'
       - '.github/actions/kwok-test/**'
       - '!**.md'
@@ -30,6 +31,7 @@ on:
     paths:
       - 'recipes/**'
       - 'kwok/**'
+      - 'tests/chainsaw/kwok/**'
       - '.github/workflows/kwok-recipes.yaml'
       - '.github/actions/kwok-test/**'
       - '!**.md'
--- a/tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml
+++ b/tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml
@@ -114,12 +114,25 @@ spec:
                 # The 4-arm OR is encoded as a single JMESPath expression
                 # because YAML doesn't accept multi-line mapping keys.
                 # Field paths are relative to `status:` here (e.g.,
-                # `sync.status` not `status.sync.status`). Arms:
-                #   1. Synced + Healthy                                — canonical pass
-                #   2. Synced + Progressing                            — KWOK pod-readiness sim gap (ADR-008)
-                #   3. OutOfSync + Healthy + op.phase == 'Succeeded'   — operator mutation (ClusterPolicy, DeviceClass)
-                #   4. Synced + Degraded + op.phase == 'Succeeded'     — health controller divergence after a successful op
-                ((sync.status == 'Synced' && health.status == 'Healthy') || (sync.status == 'Synced' && health.status == 'Progressing') || (sync.status == 'OutOfSync' && health.status == 'Healthy' && operationState.phase == 'Succeeded') || (sync.status == 'Synced' && health.status == 'Degraded' && operationState.phase == 'Succeeded')): true
+                # `sync.status` not `status.sync.status`). All four arms
+                # require `operationState.phase == 'Succeeded'` so that
+                # an Application can only count as passing AFTER Argo
+                # CD has completed at least one sync operation. For an
+                # App-of-Apps, that's the gate ensuring the child
+                # Applications have been materialized before the
+                # selector's "every Application in argocd ns" sweep
+                # accepts the root alone — otherwise the assertion
+                # races ahead and passes on the root's transient
+                # Synced state, returning before children even exist.
+                # See #1061 (and #1050 for the migration that
+                # introduced the race).
+                #
+                # Arms:
+                #   1. Synced + Healthy        — canonical pass
+                #   2. Synced + Progressing    — KWOK pod-readiness sim gap (ADR-008)
+                #   3. OutOfSync + Healthy     — operator mutation (ClusterPolicy, DeviceClass)
+                #   4. Synced + Degraded       — health controller divergence after a successful op
+                (operationState.phase == 'Succeeded' && ((sync.status == 'Synced' && health.status == 'Healthy') || (sync.status == 'Synced' && health.status == 'Progressing') || (sync.status == 'OutOfSync' && health.status == 'Healthy') || (sync.status == 'Synced' && health.status == 'Degraded'))): true
       catch:
         - script:
             content: |


mchmarny (approved):
Tight, focused fix. The predicate change is the right shape: factoring `operationState.phase == 'Succeeded'` out as a top-level conjunct makes all four arms require a completed sync operation, which rejects the pre-sync window where the App-of-Apps root can briefly satisfy `Synced` before children are materialized. Boolean check holds (`A ∨ B ∨ (C∧S) ∨ (D∧S)` → `S ∧ (A ∨ B ∨ C ∨ D)` is strictly stricter), and JMESPath fails closed on missing `operationState` so there's no surprise behavior on the cold path.

Path-filter addition is the other half of the fix — caught by Codex review on the first push, now correctly added to both `push` and `pull_request` paths so future chainsaw scenario edits actually re-trigger KWOK CI.

Comment block is also a notable upgrade — explains the race, references #1061 and #1050, and keeps the arms table readable after pulling the per-arm `op.phase` notes into the unified conjunct.

CI green (101 checks). "blocked" mergeable state is required-reviewer, not a CI signal. The `status.resources` non-empty follow-up the author flagged is a sensible belt-and-suspenders that doesn't block this PR.

Issue #1063, opened 2026-05-27
/ok-to-test checks out fork PR refs before running local actions in a privileged workflow


fallintoplace:
This looks like it has been this way since `/ok-to-test` was added in `2ca619ab` on 2026-03-05.

The problem is pretty simple:

- `.github/workflows/ok-to-test.yaml` grabs `pr.head.sha`
- passes it into `qualification.yaml`
- `qualification.yaml` checks out that SHA
- then it runs local actions from `./.github/actions/*`
- the caller gives that path `id-token: write` and `security-events: write`

So once a maintainer comments `/ok-to-test` on a fork PR, the PR author controls the local action definitions that run in a privileged workflow.

I don't think "a maintainer had to approve it" is enough protection here. It's too easy to approve a test run without separately auditing changes under `.github/actions/**` in the fork.

This repo already uses the safer pattern elsewhere: `on-push-comment.yaml` explicitly calls out guarded checkout to avoid running untrusted code in a privileged workflow.

Reasonable fixes would be either:

- make `/ok-to-test` unprivileged and skip the jobs that need elevated permissions there, or
- keep the privileged path on trusted base-repo workflow/action code only.

mchmarny:
Thanks @fallintoplace — confirmed the analysis. The local `./.github/actions/*` paths in `qualification.yaml` resolve against the workspace after the fork ref is checked out, so the privileged grants flow straight to fork-controlled YAML. Maintainer comment gating doesn't cover that.

Opened #1067 with the minimal fix from your first suggestion: drop `id-token: write` and `security-events: write` from the `/ok-to-test` → qualification call. The two jobs that requested them already degrade gracefully (SLSA predicate is gated to non-fork; SARIF upload uses `continue-on-error`), so no further qualification changes are needed.

A broader refactor (workflow_run-style guarded checkout for any future privileged step) is worth tracking separately but felt out of scope for the immediate close.

PR #1064, opened 2026-05-27
refactor(ci): make /ok-to-test's unprivileged status explicit via privileged_ci input

fallintoplace:
This keeps the `/ok-to-test` fork path unprivileged, but makes that intent explicit in the reusable workflow instead of relying on implicit permission loss and step-level soft degradation.

Context:

- #1063 pointed out that `/ok-to-test` checks out a fork PR head SHA and then runs local actions from that checkout.
- #1067 closed the immediate security gap by dropping the privileged grants from `ok-to-test.yaml`.
- This follow-up keeps the same behavior, but makes the unprivileged path obvious and reusable.

What changed:

- add a `privileged_ci` input to `qualification.yaml` with a default of `true`
- skip `cli-e2e` and `security-scan` when `privileged_ci` is `false`
- set `privileged_ci: false` in `ok-to-test.yaml`
- document that skipping `security-scan` under the unprivileged fork path is intentional so fork-derived SARIF is never uploaded under the base repo identity

Why this is better than the minimum-diff fix alone:

- the `/ok-to-test` call site now says what it means
- `cli-e2e` and `security-scan` are cleanly skipped instead of half-degrading under `issue_comment` context
- the same input can be reused by any future unprivileged caller of `qualification.yaml`

Validation:

- `actionlint -color -shellcheck= .github/workflows/qualification.yaml .github/workflows/ok-to-test.yaml`
- Ruby `YAML.load_file` parse of both edited workflow files


.github/workflows/ok-to-test.yaml
```diff
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 # Allows maintainers to run the full qualification suite on fork PRs
-# by commenting "/ok-to-test" on the PR. This runs in the base repo
-# context, so id-token: write and other elevated permissions are available.
+# by commenting "/ok-to-test" on the PR. Because this path checks out
+# an untrusted PR head SHA, it intentionally avoids elevated permissions
+# and skips qualification jobs that require them.
```
mchmarny: Heads up: the security gap from #1063 was closed by #1067 (merged 2026-05-27) — that PR dropped `id-token: write` and `security-events: write` from this same workflow call. This branch is now `mergeable_state: dirty` against main for that reason.

What's left here after rebasing onto post-#1067 main is the architectural improvement: explicit `privileged_ci: false` + job-level `if:` skips, vs #1067's implicit "drop perms and let `continue-on-error` / fork-aware step gating soft-degrade." The explicit approach is the cleaner outcome — see my comment on `qualification.yaml:25` for the substantive merits.

Recommend reframing as `refactor(ci): make /ok-to-test's unprivileged status explicit` and rebasing. Or close if the maintainer is satisfied with the minimum-diff fix from #1067.

.github/workflows/qualification.yaml
```diff
@@ -22,6 +22,11 @@ on:
         type: boolean
         required: false
         default: false
+      privileged_ci:
```
mchmarny, line 25: This `privileged_ci` input is the substantive improvement over the minimum-diff fix in #1067. Three reasons it's worth keeping if you rebase:

1. **Explicit > implicit.** A reader of `ok-to-test.yaml` sees `privileged_ci: false` and knows the intent. With #1067 alone the call site looks like a normal workflow invocation; the unprivileged status is inferred by what's *missing* from the `permissions` block.
2. **Cleaner CI output.** Skipping `cli-e2e` and `security-scan` entirely (via `if: inputs.privileged_ci` on lines 93 and 144) means no "SARIF upload skipped" notice, no partial chainsaw coverage in cli-e2e. The current `cli-e2e` on `/ok-to-test` has both `if: ...fork == false` and `if: ...fork == true` step gates that *both* evaluate false under `issue_comment` context (no `pull_request` payload), so both chainsaw passes silently skip — half-coverage that's hard to spot.
3. **Default `true` preserves behavior for `merge-gate`, `on-push`, `on-tag` callers** (all four call `qualification.yaml` and none of them set `privileged_ci`).

The pattern is also reusable for any future unprivileged-context invocation (e.g., scheduled scans on stale forks).

The one trade-off: `privileged_ci` is a fairly generic name. `trusted_source` or `trusted_workflow` would more accurately describe what the flag means semantically (it's about the trust level of the checked-out ref, not about CI privileges in the abstract). Not blocking — `privileged_ci` is clear enough in context.

.github/workflows/qualification.yaml
```diff
@@ -135,6 +141,7 @@ jobs:
 
   security-scan:
     name: Security Scan
+    if: inputs.privileged_ci
```
mchmarny, line 146: nit: when `security-scan` is skipped under `privileged_ci: false`, no SARIF is ever produced for the fork-PR run — so the GitHub code-scanning tab won't show new findings for that PR even when run via `/ok-to-test`. That's fine (the alternative, posting fork-derived findings under the base repo's identity, is worse), but worth a comment here noting that intentional gap so a future contributor doesn't try to "fix" it by gating only the SARIF upload step.

mchmarny (commented):
Thanks for the careful write-up on #1063 and for following through with a fix — really appreciate it. The analysis in the issue was spot-on and is what made the security fix easy to land quickly.

Quick context: we merged the minimum-diff fix in #1067 yesterday to close the immediate gap (dropped `id-token: write` and `security-events: write` from `ok-to-test.yaml`'s call to `qualification.yaml`). That's why this branch is now showing `dirty` against main.

That said, the `privileged_ci` toggle here is a meaningfully cleaner answer than the implicit-degradation we landed in #1067 — explicit > implicit, no half-coverage from the `cli-e2e` fork-gate quirk, and the pattern is reusable for any future unprivileged-context invocation. I'd like to keep this PR open and merge it as a refactor on top.

Could you rescope to just the still-net-new parts? After rebasing onto current `main`:

- **`qualification.yaml`**: add the `privileged_ci` input (default `true`, preserves behavior for `merge-gate` / `on-push` / `on-tag` callers) and the `if: inputs.privileged_ci` gates on `cli-e2e` and `security-scan`. Same as you have today.
- **`ok-to-test.yaml`**: only add `privileged_ci: false` to the `with:` block. The permissions block is already cleaned up by #1067 — nothing to do there.
- **PR title/description**: reframe as `refactor(ci): make /ok-to-test's unprivileged status explicit via privileged_ci input`. References #1063 / #1067 for context.

One small naming consideration in my inline review — `privileged_ci` is fine, but `trusted_source` or `trusted_workflow` more accurately captures what the flag governs (the trust level of the checked-out ref). Take it or leave it.

Thanks again for the contribution.

.github/workflows/ok-to-test.yaml
fallintoplace: @mchmarny Thank you, I will have a look later today.

mchmarny (approved):

PR #1066, opened 2026-05-27
feat(validator): co-locate ai-service-metrics with Prometheus

### LINKED ISSUE #933: [Bug]: ai-service-metrics validator non-deterministically fails on multi-SG clusters because orchestrator pod can land on a node that can't reach Prometheus

njhensley:
## Summary

Add a `dependencyAffinity` field to the validator catalog and use it to pin the `ai-service-metrics` orchestrator pod to the node running Prometheus. Eliminates intermittent failures on multi-Security-Group clusters where the dial would otherwise traverse cross-SG ingress.

## Motivation / Context

On multi-SG EKS clusters (e.g., DGXC GB200 with separate customer/system ENI subnets), `ai-service-metrics` was nondeterministically landing on nodes that could not dial the Prometheus service. Image-locality scoring decided placement once the validator image was cached, masking the underlying network-reachability constraint. The fix declares the dependency in the catalog and the deployer materializes it as a `podAffinity` term whose weight (100) dominates image-locality on first scheduling, then reinforces itself via locality on subsequent runs.

Fixes: #933
Related: N/A

## Type of Change

- [ ] Breaking change
- [ ] Documentation update
- [ ] Refactoring
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] CLI
- [ ] API server
- [ ] Recipe engine / data
- [ ] Bundlers
- [x] Validator (`pkg/validator`, `pkg/api/validator/v1`)
- [ ] Core libraries
- [x] Docs/examples (`docs/integrator/validator-extension.md`)
- [x] Other: validator catalog (`recipes/validators/catalog.yaml`)

## Implementation Notes

- New `DependencyAffinity` type on catalog entries: `componentRef`, `podLabelSelector`, `topologyKey` (defaults to `kubernetes.io/hostname`), and `requirement` (`preferred` default, or `required`).
- `BuildOrchestratorAffinity` resolves each dependency against the recipe's `componentRefs`:
  - `required` + missing/disabled/no-namespace → `ErrCodeInvalidRequest` (caller must abort before deploying).
  - `preferred` + missing → `slog.Warn` and skip the term (preserves behavior on flat networks).
  - Resolvable refs produce a `PodAffinityTerm` with the component's resolved namespace.
- Pre-flight gate added to `pkg/validator/validator.go`: walks selected validators and validates `dependencyAffinity` against the recipe before any Job is deployed.
- Deployer threads `componentRefs` into the JobPlan and renders the affinity into the orchestrator pod's `applyconfigurations` spec.
- `ai-service-metrics` catalog entry uses `requirement: preferred` so flat-cluster recipes that omit `kube-prometheus-stack` still run (validator schedules wherever, warning logged).
- Outer wrap in `pkg/validator/validator.go:290` returns `ErrCodeInternal` over an inner `ErrCodeInvalidRequest`; minor cosmetic deviation from the no-double-wrap rule, no behavior impact, tracked as follow-up.

## Testing

```bash
make qualify
go test ./pkg/api/validator/v1/... ./pkg/validator/...
```

Manually verified end-to-end on a GB200/EKS multi-SG cluster:

1. **Prometheus label sanity** — `monitoring/prometheus-kube-prometheus-prometheus-0` carries `app.kubernetes.io/name=prometheus`; selector matches.
2. **Deterministic placement** — 4/4 consecutive `validate --phase conformance` runs landed the `ai-service-metrics` orchestrator on the Prometheus node. Injected Job spec confirmed: `preferredDuringSchedulingIgnoredDuringExecution` with `weight: 100`, `matchLabels: app.kubernetes.io/name: prometheus`, `namespaces: [monitoring]`, `topologyKey: kubernetes.io/hostname`.
3. **Negative path** — with a `--data` overlay flipping `ai-service-metrics` to `requirement: required` and a recipe missing `kube-prometheus-stack` from componentRefs, validate aborts at pre-flight: `[INVALID_REQUEST] required dependencyAffinity component "kube-prometheus-stack" is not in the recipe's componentRefs`. Zero Jobs deployed.

**Coverage delta:**
- `pkg/api/validator/v1`: 67.4% → 73.6% (+6.2%)
- `pkg/validator/catalog`: 97.4% → 97.5% (+0.1%)
- `pkg/validator/job`: 12.9% → 15.7% (+2.8%)
- `pkg/validator`: 25.1% → 24.8% (-0.3%, within noise)

## Risk Assessment

- [x] **Low** — Isolated to validator orchestration. Default `requirement: preferred` preserves existing behavior on recipes without `kube-prometheus-stack`; the only validator using the new field today is `ai-service-metrics`. Easy to revert by removing the catalog entry's `dependencyAffinity` block.

**Rollout notes:** No migration required. Existing recipes work unchanged. Recipes wanting to enforce co-location for custom validators can opt-in via `requirement: required`.

## Checklist

### DIFF
--- a/docs/integrator/validator-extension.md
+++ b/docs/integrator/validator-extension.md
@@ -64,6 +64,58 @@ validators:
     env: []
 ```
 
+### Co-locating with a dependency (dependencyAffinity)
+
+A validator catalog entry can declare `dependencyAffinity` to control where its
+orchestrator Pod is scheduled relative to the components it queries. Use this
+when a check's correctness depends on pod-to-pod network reachability — the
+canonical case is `ai-service-metrics`, which dials Prometheus over a ClusterIP
+Service.
+
+```yaml
+- name: ai-service-metrics
+  phase: conformance
+  image: ghcr.io/nvidia/aicr-validators/conformance:latest
+  timeout: 5m
+  args: ["ai-service-metrics"]
+  dependencyAffinity:
+    - componentRef: kube-prometheus-stack
+      podLabelSelector:
+        app.kubernetes.io/name: prometheus
+      requirement: preferred       # or "required"; default "preferred"
+      topologyKey: kubernetes.io/hostname  # default
+```
+
+**Fields:**
+
+- `componentRef` *(required)* — the name of a component in the recipe.
+  The deployer resolves it to a namespace at spawn time using the resolved
+  recipe's `componentRefs`. If the named component is not in the recipe and
+  `requirement: required` is set, the run fails before any Job is deployed
+  with `ErrCodeInvalidRequest` — fix the recipe (add the component) or drop
+  the validator from the validation phase.
+- `podLabelSelector` *(required)* — labels that match the dependency's pods.
+  All key/value pairs must match.
+- `requirement` *(optional, default `preferred`)* — `required` emits
+  `requiredDuringSchedulingIgnoredDuringExecution`; the scheduler will refuse
+  to place the orchestrator anywhere else. `preferred` emits
+  `preferredDuringSchedulingIgnoredDuringExecution` with weight 100; the
+  scheduler treats it as the dominant scoring signal but can fall back to
+  another node if the dependency is unschedulable.
+- `topologyKey` *(optional, default `kubernetes.io/hostname`)* — the node
+  label whose value defines co-location. The default pins to the same node;
+  use `topology.kubernetes.io/zone` for zone-level locality.
+
+When in doubt, prefer `preferred`. The high weight (100) dominates the
+scheduler's image-locality scoring on the first run, after which image
+locality reinforces (rather than counteracts) the affinity. Use `required`
+only when the check has no chance of succeeding without exact co-location.
+
+See [#933](https://github.com/NVIDIA/aicr/issues/933) for the motivating case:
+on multi-Security-Group EKS clusters where customer-workload and system-workload
+ENIs sit in separate SGs with asymmetric ingress rules, the orchestrator's
+ability to dial Prometheus depends entirely on which node the scheduler picks.
+
 ### Step 4: Reference in Recipe
 
 Add the check to your recipe's validation section:
--- a/docs/integrator/validator-extension.md
+++ b/docs/integrator/validator-extension.md
@@ -106,9 +106,9 @@ Service.
   label whose value defines co-location. The default pins to the same node;
   use `topology.kubernetes.io/zone` for zone-level locality.
 
-When in doubt, prefer `preferred`. The high weight (100) dominates the
-scheduler's image-locality scoring on the first run, after which image
-locality reinforces (rather than counteracts) the affinity. Use `required`
+When in doubt, prefer `preferred`. The high weight (100) has a strong
+influence on the scheduler's scoring on the first run, after which image
+locality can support (rather than oppose) the affinity. Use `required`
 only when the check has no chance of succeeding without exact co-location.
 
 See [#933](https://github.com/NVIDIA/aicr/issues/933) for the motivating case:
--- a/pkg/api/validator/v1/affinity.go
+++ b/pkg/api/validator/v1/affinity.go
@@ -67,8 +67,8 @@ func BuildOrchestratorAffinity(
 
 	for _, dep := range deps {
 		if err := dep.Validate(); err != nil {
-			return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
-				"invalid dependencyAffinity", err)
+			return nil, errors.PropagateOrWrap(err,
+				errors.ErrCodeInvalidRequest, "invalid dependencyAffinity")
 		}
 
 		ref, found := refByName[dep.ComponentRef]
--- a/pkg/api/validator/v1/affinity.go
+++ b/pkg/api/validator/v1/affinity.go
@@ -66,15 +66,34 @@ func BuildOrchestratorAffinity(
 	var preferred []corev1.WeightedPodAffinityTerm
 
 	for _, dep := range deps {
+		if err := dep.Validate(); err != nil {
+			return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+				"invalid dependencyAffinity", err)
+		}
+
 		ref, found := refByName[dep.ComponentRef]
-		hasNamespace := found && ref.Namespace != ""
+		// Skip if the component is in the recipe but disabled or unresolved
+		// (no namespace). Without this check, "required" would block the
+		// orchestrator forever waiting for a pod that never deploys.
+		resolved := found && ref.IsEnabled() && ref.Namespace != ""
 		req := dep.RequirementOrDefault()
 
-		if !hasNamespace {
+		if !resolved {
 			if req == DependencyRequirementRequired {
+				var reason string
+				switch {
+				case !found:
+					reason = "is not in the recipe's componentRefs"
+				case !ref.IsEnabled():
+					reason = "is disabled (overrides.enabled=false)"
+				case ref.Namespace == "":
+					reason = "has no resolved namespace"
+				default:
+					reason = "is unresolved"
+				}
 				return nil, errors.New(errors.ErrCodeInvalidRequest,
-					fmt.Sprintf("required dependencyAffinity component %q is not present in the resolved recipe; either add the component to the recipe or remove this validator from the validation phase",
-						dep.ComponentRef))
+					fmt.Sprintf("required dependencyAffinity component %q %s; either fix the recipe or remove this validator from the validation phase",
+						dep.ComponentRef, reason))
 			}
 			slog.Warn("preferred dependencyAffinity component not present in recipe; skipping affinity term",
 				"componentRef", dep.ComponentRef)
--- a/pkg/api/validator/v1/affinity.go
+++ b/pkg/api/validator/v1/affinity.go
@@ -1,0 +1,109 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package v1
+
+import (
+	"fmt"
+	"log/slog"
+
+	"github.com/NVIDIA/aicr/pkg/errors"
+	"github.com/NVIDIA/aicr/pkg/recipe"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// preferredAffinityWeight is the scheduler weight for preferred dependency
+// affinity terms. 100 (max) so the term dominates image-locality scoring on
+// the first scheduling decision; once the validator image lands on the
+// dependency node, image-locality reinforces affinity on subsequent runs.
+const preferredAffinityWeight = 100
+
+// BuildOrchestratorAffinity composes the orchestrator pod's full affinity from
+// the validator's declared dependencies and the resolved recipe's component
+// refs. The result always includes the default prefer-CPU NodeAffinity; each
+// resolvable dependency adds one PodAffinity term.
+//
+// Resolution rules (per https://github.com/NVIDIA/aicr/issues/933):
+//   - A "required" dependency whose componentRef is missing from componentRefs
+//     returns ErrCodeInvalidRequest. The caller should treat this as a recipe
+//     misconfiguration and fail the run before deploying any Job.
+//   - A "preferred" dependency whose componentRef is missing is logged at
+//     slog.Warn and produces no PodAffinity term. The orchestrator schedules
+//     wherever the scheduler picks; this preserves backward-compatible behavior
+//     on flat networks where the dependency may not be present.
+//   - Components whose Namespace is empty after recipe resolution are treated
+//     as missing (a dependency without a known namespace cannot produce a
+//     well-formed PodAffinityTerm).
+func BuildOrchestratorAffinity(
+	deps []DependencyAffinity,
+	componentRefs []recipe.ComponentRef,
+) (*corev1.Affinity, error) {
+
+	affinity := preferCPUNodeAffinity()
+
+	if len(deps) == 0 {
+		return affinity, nil
+	}
+
+	refByName := make(map[string]recipe.ComponentRef, len(componentRefs))
+	for _, ref := range componentRefs {
+		refByName[ref.Name] = ref
+	}
+
+	var required []corev1.PodAffinityTerm
+	var preferred []corev1.WeightedPodAffinityTerm
+
+	for _, dep := range deps {
+		ref, found := refByName[dep.ComponentRef]
+		hasNamespace := found && ref.Namespace != ""
+		req := dep.RequirementOrDefault()
+
+		if !hasNamespace {
+			if req == DependencyRequirementRequired {
+				return nil, errors.New(errors.ErrCodeInvalidRequest,
+					fmt.Sprintf("required dependencyAffinity component %q is not present in the resolved recipe; either add the component to the recipe or remove this validator from the validation phase",
+						dep.ComponentRef))
+			}
+			slog.Warn("preferred dependencyAffinity component not present in recipe; skipping affinity term",
+				"componentRef", dep.ComponentRef)
+			continue
+		}
+
+		term := corev1.PodAffinityTerm{
+			LabelSelector: &metav1.LabelSelector{MatchLabels: dep.PodLabelSelector},
+			Namespaces:    []string{ref.Namespace},
+			TopologyKey:   dep.TopologyKeyOrDefault(),
+		}
+
+		if req == DependencyRequirementRequired {
+			required = append(required, term)
+		} else {
+			preferred = append(preferred, corev1.WeightedPodAffinityTerm{
+				Weight:          preferredAffinityWeight,
+				PodAffinityTerm: term,
+			})
+		}
+	}
+
+	if len(required) == 0 && len(preferred) == 0 {
+		return affinity, nil
+	}
+
+	affinity.PodAffinity = &corev1.PodAffinity{
+		RequiredDuringSchedulingIgnoredDuringExecution:  required,
+		PreferredDuringSchedulingIgnoredDuringExecution: preferred,
+	}
+	return affinity, nil
+}
--- a/pkg/api/validator/v1/job_plan.go
+++ b/pkg/api/validator/v1/job_plan.go
@@ -22,12 +22,14 @@ import (
 	"time"
 
 	"github.com/NVIDIA/aicr/pkg/defaults"
+	"github.com/NVIDIA/aicr/pkg/recipe"
 	"github.com/NVIDIA/aicr/pkg/validator/labels"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	applybatchv1 "k8s.io/client-go/applyconfigurations/batch/v1"
 	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
+	applymetav1 "k8s.io/client-go/applyconfigurations/meta/v1"
 )
 
 // JobPlan contains all components needed to build a validator Job.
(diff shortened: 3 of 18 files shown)


docs/integrator/validator-extension.md
```diff
@@ -64,6 +64,58 @@ validators:
...
+  scheduler treats it as the dominant scoring signal but can fall back to
+  another node if the dependency is unschedulable.
+- `topologyKey` *(optional, default `kubernetes.io/hostname`)* — the node
+  label whose value defines co-location. The default pins to the same node;
+  use `topology.kubernetes.io/zone` for zone-level locality.
+
+When in doubt, prefer `preferred`. The high weight (100) dominates the
+scheduler's image-locality scoring on the first run, after which image
+locality reinforces (rather than counteracts) the affinity. Use `required`
+only when the check has no chance of succeeding without exact co-location.
```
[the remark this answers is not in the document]
njhensley: Done in `2c22a086`. Used slightly softer wording than the suggested diff — "has a strong influence on the scheduler's scoring on the first run, after which image locality can support (rather than oppose) the affinity" — but matches the intent (observation rather than guarantee).

pkg/api/validator/v1/affinity.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	var required []corev1.PodAffinityTerm
+	var preferred []corev1.WeightedPodAffinityTerm
+
+	for _, dep := range deps {
+		if err := dep.Validate(); err != nil {
+			return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+				"invalid dependencyAffinity", err)
+		}
```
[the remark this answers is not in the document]
njhensley: Applied verbatim in `2c22a086`.

pkg/validator/catalog/catalog.go
```diff
@@ -275,6 +275,12 @@ func validate(c *ValidatorCatalog) error {
 			return errors.New(errors.ErrCodeInvalidRequest,
 				fmt.Sprintf("validator %q: image is required", v.Name))
 		}
+		for j, dep := range v.DependencyAffinity {
+			if err := dep.Validate(); err != nil {
+				return errors.Wrap(errors.ErrCodeInvalidRequest,
+					fmt.Sprintf("validator %q: dependencyAffinity[%d]", v.Name, j), err)
+			}
```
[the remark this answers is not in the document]
njhensley: Applied in `2c22a086`. Side-effect: the existing `TestParseDependencyAffinity` substring assertions previously locked in the `validator "X": dependencyAffinity[N]` wrap prefix — they were updated in the same commit to assert on the inner `Validate()` diagnostic content (`componentRef is required`, `podLabelSelector is required`, `invalid requirement "soft"`), which is the actionable bit and is producer-stable.

pkg/validator/validator.go
```diff
@@ -279,6 +279,18 @@ func (v *Validator) runPhase(
...
+	// resolve before any Job is deployed. This honors the per-validator
+	// contract (BuildOrchestratorAffinity returns ErrCodeInvalidRequest for
+	// missing required components) at phase scope, so a single misconfigured
+	// entry doesn't strand a partial deploy of earlier entries.
+	for _, entry := range entries {
+		if _, err := v1.BuildOrchestratorAffinity(entry.DependencyAffinity, validationInput.GetComponentRefs()); err != nil {
+			return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("dependencyAffinity pre-flight failed for validator %q", entry.Name), err)
+		}
+	}
```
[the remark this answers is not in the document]
njhensley: Done in `2c22a086`. `select { case <-ctx.Done(): return nil, errors.Wrap(errors.ErrCodeTimeout, "context canceled during dependencyAffinity pre-flight", ctx.Err()) }` block added at the top of the pre-flight loop body (validator.go:288-292).

pkg/validator/validator.go
```diff
@@ -279,6 +279,18 @@ func (v *Validator) runPhase(
...
+	// Pre-flight: validate all dependencyAffinity for required components
+	// resolve before any Job is deployed. This honors the per-validator
+	// contract (BuildOrchestratorAffinity returns ErrCodeInvalidRequest for
+	// missing required components) at phase scope, so a single misconfigured
+	// entry doesn't strand a partial deploy of earlier entries.
+	for _, entry := range entries {
+		if _, err := v1.BuildOrchestratorAffinity(entry.DependencyAffinity, validationInput.GetComponentRefs()); err != nil {
+			return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("dependencyAffinity pre-flight failed for validator %q", entry.Name), err)
+		}
```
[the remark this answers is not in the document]
njhensley: Applied verbatim in `2c22a086`.

recipes/validators/catalog.yaml
```diff
@@ -107,6 +107,15 @@ validators:
...
     env: []
+    # Co-locate the orchestrator pod with Prometheus so the dial is loopback
+    # / same-node, not subject to cross-Security-Group ingress rules. Required
+    # for determinism on multi-SG clusters (e.g., DGXC EKS with separate
+    # customer/system ENI subnets). See https://github.com/NVIDIA/aicr/issues/933.
+    dependencyAffinity:
+      - componentRef: kube-prometheus-stack
+        podLabelSelector:
+          app.kubernetes.io/name: prometheus
+        requirement: preferred
```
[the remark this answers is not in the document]
njhensley: Done in `2c22a086`. Reworded to "Best-effort co-location: prefer the same node as Prometheus … Uses requirement: preferred — schedule still falls back to any node if Prometheus is unschedulable." Matches the configured `requirement: preferred`.

njhensley:
Re: the outside-diff finding on `pkg/api/validator/v1/README.md:322-336` (Plan(...) parameter list missing `imageRegistryOverride` / `imageTagOverride` / `componentRefs`): added three bullets in `2c22a086`, matching the existing parameter-description style and placement.

pkg/validator/validator.go
```diff
@@ -279,6 +279,23 @@ func (v *Validator) runPhase(
...
+	// contract (BuildOrchestratorAffinity returns ErrCodeInvalidRequest for
+	// missing required components) at phase scope, so a single misconfigured
+	// entry doesn't strand a partial deploy of earlier entries.
+	for _, entry := range entries {
+		select {
+		case <-ctx.Done():
+			return nil, errors.Wrap(errors.ErrCodeTimeout, "context canceled during dependencyAffinity pre-flight", ctx.Err())
+		default:
+		}
+		if _, err := v1.BuildOrchestratorAffinity(entry.DependencyAffinity, validationInput.GetComponentRefs()); err != nil {
```
mchmarny, line 293: Pre-flight calls `BuildOrchestratorAffinity` purely for its validation side-effect — the returned `*corev1.Affinity` is discarded, but each invocation allocates the full affinity tree, builds the `refByName` map, and runs the converter logic.

Extract a `ValidateDependencyAffinity(deps, componentRefs) error` that performs only the resolution checks, then have `BuildOrchestratorAffinity` call it internally. Pre-flight uses the validate path; deploy path keeps building. Cheaper and the intent is clearer at the call site (you're validating, not building).

Not blocking — the entry count per phase is small.

pkg/api/validator/v1/job_plan.go
```diff
@@ -416,3 +429,164 @@ func RenderPlanToApplyConfig(plan JobPlan, jobName string) *applybatchv1.JobAppl
...
+	}
+	return preferCPUNodeAffinity()
+}
+
+// affinityToApplyConfig converts a *corev1.Affinity to the
+// applyconfigurations/core/v1 type used by server-side apply. We hand-write the
+// walk because client-go does not expose a generated converter for this pair.
+// A nil or empty Affinity falls back to preferCPUNodeAffinity() to preserve
+// pre-PR behavior for external callers that construct a JobPlan manually.
+func affinityToApplyConfig(a *corev1.Affinity) *applycorev1.AffinityApplyConfiguration {
```
mchmarny, line 449: Hand-written `corev1.Affinity` → `applycorev1.AffinityApplyConfiguration` walker. The docstring acknowledges client-go doesn't ship a generated converter; the risk is that when k8s API adds new affinity fields in future bumps (e.g. recent additions like `MatchLabelKeys` / `MismatchLabelKeys` were added in v1.29-v1.30), this walker will silently drop them — server-side apply won't error, the field just won't materialize.

Two suggestions, either acceptable:
1. Add an explicit allowlist test that fails when `reflect.TypeOf(corev1.PodAffinityTerm{}).NumField()` changes, so the next k8s bump forces a re-audit of this file.
2. Drop the walker entirely and round-trip via JSON marshal → unmarshal into the apply config type. Slower, but immune to API drift.

Round-tripping `TestAffinityToApplyConfig_RoundTripsAllFields` covers today's surface but won't catch tomorrow's additions.

pkg/api/validator/v1/affinity.go
```diff
@@ -0,0 +1,128 @@
...
+		// orchestrator forever waiting for a pod that never deploys.
+		resolved := found && ref.IsEnabled() && ref.Namespace != ""
+		req := dep.RequirementOrDefault()
+
+		if !resolved {
+			if req == DependencyRequirementRequired {
+				var reason string
+				switch {
+				case !found:
+					reason = "is not in the recipe's componentRefs"
```
mchmarny, line 86: `required` resolution checks recipe *membership* (`ref.IsEnabled()`, `ref.Namespace != ""`), not runtime *presence*. If `kube-prometheus-stack` is in `componentRefs` but its pods haven't started yet (or crashlooped), a `required` PodAffinity term on the orchestrator will leave the Job stuck in Pending until the deploy timeout fires — with no signal in the validator output beyond "ran out of time."

The docstring above (lines 36-39) implies the recipe-membership check is the guard, but the failure mode is actually different from what it suggests. Worth either:
- explicitly noting in the docstring that `required` assumes the dependency's pods will eventually be Running, OR
- having the deployer surface the orchestrator's Pending reason (`PodScheduled=False`) into the validator output when wait-for-Job times out.

For `ai-service-metrics` this is moot (you use `preferred`). It matters once someone declares the first `required` dependency.

pkg/api/validator/v1/job_plan.go
```diff
@@ -192,7 +210,13 @@ func BuildJobPlan(
 	nodeSelector map[string]string,
 	imageRegistryOverride string,
 	imageTagOverride string,
-) JobPlan {
+	componentRefs []recipe.ComponentRef,
+) (JobPlan, error) {
+
+	affinity, err := BuildOrchestratorAffinity(entry.DependencyAffinity, componentRefs)
```
mchmarny, line 216: nit: `Plan` and `BuildJobPlan` are now at 13 positional parameters and the test files are a sea of `nil, nil, nil, "", "", nil` placeholders that have to be re-aligned every time anything is added. The README also has to enumerate them in prose.

Worth a follow-up to convert to a config struct (`JobPlanConfig` / `PlanConfig`) or functional options — same shape this repo already uses for `recipe.NewBuilder` and `server.New`. Not blocking, but the friction will compound the next time someone needs to thread another field through.

recipes/validators/catalog.yaml
```diff
@@ -107,6 +107,17 @@ validators:
...
+    # is loopback / same-node and avoids cross-Security-Group ingress filtering.
+    # Uses requirement: preferred — schedule still falls back to any node if
+    # Prometheus is unschedulable. Motivated by multi-SG clusters (e.g., DGXC
+    # EKS with separate customer/system ENI subnets) where pod-to-pod
+    # reachability is asymmetric. See https://github.com/NVIDIA/aicr/issues/933.
+    dependencyAffinity:
+      - componentRef: kube-prometheus-stack
+        podLabelSelector:
+          app.kubernetes.io/name: prometheus
+        requirement: preferred
```
mchmarny, line 120: The PR description confirms `monitoring/prometheus-kube-prometheus-prometheus-0` carries `app.kubernetes.io/name=prometheus` on the test cluster — good. One caveat to lock in: when `kube-prometheus-stack` chart is upgraded (or when overrides set a custom release name), the Prometheus pods sometimes only carry `prometheus=kube-prometheus-stack-prometheus` or `app=kube-prometheus-stack-prometheus` and *not* the generic `app.kubernetes.io/name`. That isn't a bug in this PR, but it does mean a downstream chart bump can silently turn the affinity into a no-op (since `preferred`, no error).

Optional defense: render a `slog.Warn` from the deployer when the LabelSelector matches zero pods at deploy time (informer is right there).

mchmarny (commented):
Mechanism is sound: pinning to `kubernetes.io/hostname` with weight=100 is the right shape for the multi-SG reachability problem, and `preferred` as the catalog default keeps flat-cluster recipes working. Test coverage is strong — `TestAffinityToApplyConfig_RoundTripsAllFields` is especially welcome.

Five non-blocking comments inline: one on the apply-config converter's drift risk on future k8s bumps (most important), one on the pre-flight gate doing more work than it needs to, one on `required` not actually guaranteeing runtime presence (only recipe membership), a long-parameter-list nit, and an observation about Prometheus label fragility across chart versions.

CI is green. Nothing here blocks merge.

mchmarny (approved):

pkg/validator/validator.go
njhensley: Good catch. Refactored in the staged changes: extracted `resolveDeps` as an unexported helper, added a public `ValidateDependencyAffinity(deps, componentRefs) error` that calls `resolveDeps` with warning emission suppressed, and switched `runPhase`'s pre-flight to use it. `BuildOrchestratorAffinity` now also delegates to `resolveDeps`, so the validate and build paths share one source of truth and pre-flight no longer double-emits the missing-preferred warning.

pkg/api/validator/v1/job_plan.go
njhensley: Took option (1). Added `TestAffinityTypeFieldCountInvariant` in `affinity_test.go` that asserts the field counts of `corev1.Affinity` / `NodeAffinity` / `NodeSelectorTerm` / `PodAffinity` / `PodAntiAffinity` / `PodAffinityTerm` against current expected values. Future k8s vendor bumps that add or remove a field will fail the test with a message instructing the bumper to audit `affinityToApplyConfig` before updating the count. Cheaper than JSON round-trip, gives the next k8s bumper a forcing function.

pkg/api/validator/v1/affinity.go
njhensley: Real gap. Clarified `BuildOrchestratorAffinity`'s docstring (and `ValidateDependencyAffinity`'s) to explicitly call out that resolution checks recipe membership only, not runtime readiness, and to point operators at PodScheduled + the dependency component's replica status when triaging a hung run. Deferring the deployer's scheduling-diagnostics-on-timeout change to a separate follow-up issue since it's structural (a Job timeout would need to read pod conditions before returning), out of scope for this PR, and doesn't affect `ai-service-metrics` which uses `preferred`.

PR #1069, opened 2026-05-28
refactor(serializer): kubeconfig-aware writer for ConfigMap output

### LINKED ISSUE #826: refactor(serializer): add kubeconfig-aware NewFileWriterOrStdoutWithKubeconfig and update all CLI callers
## Background

Every CLI command that performs kubeconfig-aware reads for input (`query`, `snapshot`, `recipe`, `validate`, `diff`) calls plain `serializer.NewFileWriterOrStdout` for output, which does not propagate the kubeconfig. This creates a read/write asymmetry — writes to ConfigMap destinations (e.g. `cm://namespace/name`) cannot use the same kubeconfig that was used for reads, limiting correctness for multi-cluster workflows.

This was identified during review of #824 (feat: aicr diff) and deferred because adding a kubeconfig-aware writer only for `diff` would be inconsistent with the rest of the codebase. All callers need to be updated in a coordinated sweep.

## Scope

- Add `serializer.NewFileWriterOrStdoutWithKubeconfig(path, kubeconfig string) ...` (or equivalent) to the serializer package.
- Update all CLI command callers in lockstep: `pkg/cli/query.go`, `pkg/cli/diff.go`, `pkg/cli/snapshot.go`, `pkg/cli/recipe.go`, `pkg/cli/validate.go`, and any others identified during implementation.

## References

- PR #824 (feat: aicr diff for configuration drift detection) — original context: https://github.com/NVIDIA/aicr/pull/824
- Requested by: @sanjeevrg89

ayuskauskas:
## Summary

Adds `serializer.NewFileWriterOrStdoutWithKubeconfig` (and `NewConfigMapWriterWithKubeconfig`) mirroring the existing `FromFile` / `FromFileWithKubeconfig` reader pair, then threads `--kubeconfig` through every CLI command that writes a snapshot/recipe/diff/validation report to a `cm://` destination. Reads and writes in a single invocation now target the same cluster.

## Motivation / Context

Identified during review of #824 (`aicr diff`): every CLI that does a kubeconfig-aware read for input still called the plain `NewFileWriterOrStdout` for output, so writes to `cm://namespace/name` could not honor the same kubeconfig — broken for multi-cluster workflows.

Fixes: #826
Related: #824

## Type of Change

(The new function is additive; existing `NewFileWriterOrStdout` is preserved as a thin delegate, matching the reader-side pattern.)

## Component(s) Affected

- [x] Other: `pkg/serializer`

## Implementation Notes

- **New API:** `NewFileWriterOrStdoutWithKubeconfig(format, path, kubeconfig)` + `NewConfigMapWriterWithKubeconfig(namespace, name, kubeconfig, format)`. Existing `NewFileWriterOrStdout` / `NewConfigMapWriter` delegate with `""`.
- **CLI threading:** `diff.go` (`writeDiffResult` takes `kubeconfig`), `validate.go` (new field on `validationConfig`), `snapshot.go` (`createSnapshotSerializer` takes `kubeconfig`), `recipe.go` (passes `cmd.String("kubeconfig")`).
- **Bonus fixes (folded in to keep the change atomic):**
  - Per-path cache in `GetKubeClientWithConfig`: a single `validate --kubeconfig=X` run hitting recipe read + snapshot read + ConfigMap write now reuses one client + one TLS handshake instead of three. Empty/whitespace paths delegate to the existing singleton.
  - `--template` + `cm://` output rejected at parse time with a clear `INVALID_REQUEST` error — the template writer only emits local files and would otherwise fail later with a confusing "open cm://...: no such file" path error.
  - `TrimSpace` kubeconfig in `BuildKubeClient` / `GetKubeClientWithConfig` as defense-in-depth so a stray space doesn't bypass default discovery. (clientcmd appears to already tolerate this in practice, but normalizing at our layer keeps the cache key clean and the contract explicit.)
- **Docs:** `--kubeconfig` description updated for snapshot, recipe, validate to note ConfigMap-output behavior. `pkg/serializer/doc.go` example added.

## Testing

```bash
make qualify
golangci-lint run -c .golangci.yaml ./pkg/k8s/client/... ./pkg/serializer/... ./pkg/cli/...
go test -race -count=1 ./pkg/k8s/client/... ./pkg/serializer/... ./pkg/cli/...
```

- Unit tests pass, race detector clean.
- Lint clean on all touched packages.
- `make qualify` passes everything except `chainsaw/cli-bundle-helmfile`, which fails locally because the `helmfile` binary isn't on PATH (environment issue — unrelated to this change; touches the bundler's helmfile output, not the serializer or CLI writer).

### Manual end-to-end verification (kind, two clusters)

Built `main` and PR binaries, ran the same command against both with `KUBECONFIG=B` env and `--kubeconfig=A` flag pointing at separate kind clusters:

```bash
aicr recipe --service eks --accelerator h100 --intent training --os ubuntu \
  --kubeconfig /tmp/kc-a.yaml --output cm://default/test-826-recipe
```

| Binary | KUBECONFIG env | `--kubeconfig` flag | ConfigMap landed in |
|---|---|---|---|
| `main` | B | A | **B** ❌ (flag ignored for output — the bug) |
| this PR | B | A | **A** ✅ (flag honored end-to-end) |

The exact bug from issue #826 reproduced on `main` and fixed on this branch. `aicr diff` with `cm://` baseline + `cm://` target + `cm://` output verified the same fix end-to-end. `--template cm://...` rejection verified as well.

### New test cases
- `pkg/serializer`: kubeconfig propagation into `ConfigMapWriter`, default-discovery preservation, format coercion through the back-compat wrapper, file destinations ignore kubeconfig.
- `pkg/cli/diff_test.go`: `writeDiffResult` with `cm://` output + bogus kubeconfig — error must reference the threaded path (proves propagation, not silent fallback).
- `pkg/cli/snapshot_test.go`: `--template` + `cm://` output rejection.
- `pkg/k8s/client`: whitespace-normalization, per-path cache identity on repeat calls, empty/whitespace delegation to singleton.

**Coverage:**
- `pkg/serializer`: 73.8% → 73.5% (-0.3%)
- `pkg/cli`: unchanged
- `pkg/k8s/client`: improved (new tests on previously unexercised branches)
- Module total: 76.2% → 76.3%

## Risk Assessment

- [x] **Low** — Isolated change, additive API surface, backward-compatible delegating wrappers preserved. The kubeconfig argument defaults to `""` (existing behavior) so any external importer of `NewFileWriterOrStdout` / `NewConfigMapWriter` is unaffected.

**Rollout notes:** No migration needed. `--kubeconfig` continues to behave as before; it now also governs `--output cm://...` for consistency with input reads.

## Checklist

- [x] Changes follow existing patterns in the codebase (mirrors `FromFile` / `FromFileWithKubeconfig`)

### DIFF
--- a/pkg/cli/recipe.go
+++ b/pkg/cli/recipe.go
@@ -179,6 +179,8 @@ Override snapshot-detected criteria:
 				return err
 			}
 
+			kubeconfig := cmd.String("kubeconfig")
+
 			result, err := buildRecipeFromCmdWithConfig(ctx, cmd, cfg)
 			if err != nil {
 				return errors.PropagateOrWrap(err, errors.ErrCodeInternal, "error building recipe")
@@ -197,7 +199,7 @@ Override snapshot-detected criteria:
 			}
 
 			output := recipeOutputPath(cmd, cfg)
-			ser, err := serializer.NewFileWriterOrStdout(outFormat, output)
+			ser, err := serializer.NewFileWriterOrStdoutWithKubeconfig(outFormat, output, kubeconfig)
 			if err != nil {
 				return errors.Wrap(errors.ErrCodeInternal, "failed to create output writer", err)
 			}
--- a/pkg/cli/snapshot_test.go
+++ b/pkg/cli/snapshot_test.go
@@ -15,13 +15,16 @@
 package cli
 
 import (
+	"context"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
 
+	"github.com/urfave/cli/v3"
 	corev1 "k8s.io/api/core/v1"
 
+	"github.com/NVIDIA/aicr/pkg/config"
 	"github.com/NVIDIA/aicr/pkg/serializer"
 	"github.com/NVIDIA/aicr/pkg/snapshotter"
 )
@@ -154,12 +157,36 @@ func TestSnapshotTemplateFlagCombinations(t *testing.T) {
 			output:       "",
 			wantErr:      false,
 		},
+		// Template + ConfigMap URI output must be rejected: the template
+		// writer only emits to local files, so a cm:// path would silently
+		// create a file literally named "cm:..." instead of writing to K8s.
+		{
+			name:         "template with ConfigMap URI output is rejected",
+			templatePath: templatePath,
+			format:       "yaml",
+			formatSet:    false,
+			output:       "cm://aicr/snap",
+			wantErr:      true,
+			errContains:  "ConfigMap",
+		},
+		{
+			name:         "no template with ConfigMap URI output is allowed",
+			templatePath: "",
+			format:       "yaml",
+			formatSet:    false,
+			output:       "cm://aicr/snap",
+			wantErr:      false,
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			// Validate the combination
-			err := validateTemplateFlagCombination(tt.templatePath, tt.format, tt.formatSet)
+			// Exercise the real production function rather than a hand-copied
+			// mirror — the prior helper drifted from the actual validation
+			// rules during the ConfigMap-rejection addition.
+			cmd := buildSnapshotCmdForTemplateTest(t, tt.templatePath, tt.format, tt.formatSet, tt.output)
+			outFormat := serializer.Format(tt.format)
+			_, err := parseSnapshotTemplateOptions(cmd, outFormat, &config.SnapshotResolved{})
 
 			if tt.wantErr {
 				if err == nil {
@@ -176,29 +203,38 @@ func TestSnapshotTemplateFlagCombinations(t *testing.T) {
 	}
 }
 
-// validateTemplateFlagCombination validates the template + format combination.
-// This mirrors the validation logic in snapshotCmd.Action.
-func validateTemplateFlagCombination(templatePath, format string, formatSet bool) error {
-	if templatePath == "" {
-		return nil // No template, no validation needed
+// buildSnapshotCmdForTemplateTest constructs a parsed *cli.Command with
+// --template / --format / --output set so parseSnapshotTemplateOptions can be
+// exercised in isolation. formatSet=false omits --format so the test exercises
+// the cmd.IsSet("format") branch.
+func buildSnapshotCmdForTemplateTest(t *testing.T, templatePath, format string, formatSet bool, output string) *cli.Command {
+	t.Helper()
+	cmd := snapshotCmd()
+	app := &cli.Command{Name: "aicr", Commands: []*cli.Command{cmd}}
+
+	args := []string{"aicr", "snapshot"}
+	if templatePath != "" {
+		args = append(args, "--template", templatePath)
+	}
+	if formatSet {
+		args = append(args, "--format", format)
+	}
+	if output != "" {
+		args = append(args, "--output", output)
 	}
 
-	// Validate format is YAML when using template
-	if formatSet && format != string(serializer.FormatYAML) {
-		return &validationError{msg: "--template requires YAML format; --format must be \"yaml\" or omitted"}
+	var captured *cli.Command
+	cmd.Action = func(_ context.Context, c *cli.Command) error {
+		captured = c
+		return nil
 	}
-
-	// Validate template file exists
-	return serializer.ValidateTemplateFile(templatePath)
-}
-
-// validationError is a simple error type for validation failures.
-type validationError struct {
-	msg string
-}
-
-func (e *validationError) Error() string {
-	return e.msg
+	if err := app.Run(t.Context(), args); err != nil {
+		t.Fatalf("flag parse setup failed: %v", err)
+	}
+	if captured == nil {
+		t.Fatal("flag parse setup did not capture cmd")
+	}
+	return captured
 }
 
 // TestParseResourceList covers the --requests / --limits parser, including
--- a/pkg/k8s/client/client.go
+++ b/pkg/k8s/client/client.go
@@ -17,6 +17,7 @@ package client
 import (
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 
 	"github.com/NVIDIA/aicr/pkg/errors"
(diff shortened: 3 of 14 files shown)


pkg/k8s/client/client.go
```diff
@@ -117,17 +139,35 @@ func BuildKubeClient(kubeconfig string) (*kubernetes.Clientset, *rest.Config, er
...
+		return GetKubeClient()
+	}
+
+	pathClientMu.Lock()
+	defer pathClientMu.Unlock()
+	if entry, ok := pathClientCache[key]; ok {
+		return entry.client, entry.config, entry.err
+	}
+	client, config, err := BuildKubeClient(key)
+	pathClientCache[key] = &cachedPathClient{client: client, config: config, err: err}
```
mchmarny: Errors are cached for process lifetime alongside successful clients, with no distinction between transient and permanent failures. A momentary `EAGAIN` reading the kubeconfig, a brief filesystem hiccup, or a token-rotation race on first call gets pinned as the cached answer for every subsequent call to that path — for the *entire process lifetime*. The doc says "permanently-bad kubeconfig path fails fast," but the code can't tell the two apart.

For the CLI this is harmless (process exits in seconds). The function is also reachable from `pkg/serializer/reader.go` and `pkg/snapshotter/agent.go`, either of which could be invoked from a future long-running daemon path; today neither does, but the trap is set.

Two reasonable options:
1. Cache successes only; on error, return the error but don't memoize it.
2. Add a TTL or `Invalidate(path)` so a caller that detects "this is stale" can force a refresh.

Option 1 is simpler and matches the principle that the cache is an optimization, not a circuit breaker. Tests would lose one assertion (error identity across calls) — that assertion is what surfaced this.

pkg/k8s/client/client.go
```diff
@@ -117,17 +139,35 @@ func BuildKubeClient(kubeconfig string) (*kubernetes.Clientset, *rest.Config, er
...
-//   - kubeconfig: Path to kubeconfig file
+//   - kubeconfig: Path to kubeconfig file. Empty or whitespace-only falls back to
+//     default discovery via the singleton.
 //
 // Returns:
 //   - Interface: The Kubernetes client interface
 //   - *rest.Config: The rest configuration
-//   - error: Any error encountered
+//   - error: Any error encountered (cached on subsequent calls for the same path)
 func GetKubeClientWithConfig(kubeconfig string) (Interface, *rest.Config, error) {
```
mchmarny, line 167: `pathClientCache` is an unbounded `map[string]*cachedPathClient` at package scope. For a CLI run that's fine — one or two entries, process exits. But the function is exported and the package singleton survives across any long-lived embedding (today: none; tomorrow: easy to add).

Worth a one-line docstring on the cache var declaring this is a CLI-lifetime cache, *not* sized for daemon use — so a future contributor doesn't reach for it from `aicrd` server code without revisiting bounds.

pkg/serializer/configmap.go
```diff
@@ -61,7 +79,14 @@ func (w *ConfigMapWriter) Serialize(ctx context.Context, snapshot any) error {
 	writeCtx, cancel := context.WithTimeout(ctx, defaults.ConfigMapWriteTimeout)
 	defer cancel()
 
-	client, config, err := client.GetKubeClient()
+	var k8sClient client.Interface
+	var config *rest.Config
+	var err error
+	if w.kubeconfig != "" {
+		k8sClient, config, err = client.GetKubeClientWithConfig(w.kubeconfig)
+	} else {
```
mchmarny: nit: the `if w.kubeconfig != ""` branch is now redundant — the new `GetKubeClientWithConfig` already delegates to `GetKubeClient` on empty/whitespace input (client.go:154-156). Collapsing to a single call:

```go
k8sClient, config, err = client.GetKubeClientWithConfig(w.kubeconfig)
```

…removes the only place in the codebase that has to know about the two-function dispatch. `pkg/serializer/reader.go:434` is already using the single-call pattern.

pkg/cli/snapshot_test.go
```diff
@@ -176,9 +196,9 @@ func TestSnapshotTemplateFlagCombinations(t *testing.T) {
 	}
 }
 
-// validateTemplateFlagCombination validates the template + format combination.
-// This mirrors the validation logic in snapshotCmd.Action.
-func validateTemplateFlagCombination(templatePath, format string, formatSet bool) error {
+// validateTemplateFlagCombination validates the template + format + output
+// combination. This mirrors the validation logic in parseSnapshotTemplateOptions.
+func validateTemplateFlagCombination(templatePath, format string, formatSet bool, output string) error {
```
mchmarny: nit: `validateTemplateFlagCombination` is a hand-copied mirror of the validation logic in `parseSnapshotTemplateOptions` (the comment says as much). The two paths can drift — e.g., the next time someone adds a rule to the production function, this test won't catch them omitting it from the mirror.

Test `parseSnapshotTemplateOptions` directly. The function takes `*cli.Command` + format + outputPath; construct a minimal `*cli.Command` the same way `buildDiffCommandWithOutput` already does in `diff_test.go` and call the real thing.

pkg/cli/recipe.go
```diff
@@ -197,7 +197,7 @@ Override snapshot-detected criteria:
 			}
 
 			output := recipeOutputPath(cmd, cfg)
-			ser, err := serializer.NewFileWriterOrStdout(outFormat, output)
+			ser, err := serializer.NewFileWriterOrStdoutWithKubeconfig(outFormat, output, cmd.String("kubeconfig"))
```
mchmarny: nit: this is the only one of the four CLI commands that reads kubeconfig inline via `cmd.String("kubeconfig")` at the call site. `snapshot.go`, `validate.go`, and `diff.go` all bind it to a named local (or struct field) and pass that. Easy to lose track of where `--kubeconfig` is being consumed when one command does it differently.

Pull a local `kubeconfig := cmd.String("kubeconfig")` near the top of this action (same scope as `outFormat`) for consistency.

mchmarny (commented):
Symmetric `WithKubeconfig` variants mirror the reader-side pattern cleanly and the e2e reproduction table in the PR description is the right way to prove the multi-cluster fix. CI green, all the threading is straightforward.

Five non-blocking comments inline: one medium on the per-path cache memoizing errors permanently (transient kubeconfig failures become permanent for the process), one related note on the cache being CLI-lifetime by design without that being doc'd, and three nits — a now-redundant `if w.kubeconfig != ""` branch in `configmap.go`, the test-side mirror of `parseSnapshotTemplateOptions` validation logic that can drift, and one inconsistent inline `cmd.String("kubeconfig")` in `recipe.go`.

Nothing here blocks merge.

ayuskauskas:
Thanks for the review. Force-pushed `f5f82907` addressing all 6 comments:

1. **(@mchmarny — medium, cache errors)** — Cache successes only. Errors are recomputed every call so a transient first-call failure can't pin permanent. Removed the `err` field from `cachedPathClient`; replaced the error-identity assertion with `TestGetKubeClientWithConfig_ErrorsNotCached` that asserts the cache stays empty after error returns and a retry re-attempts.
2. **(@mchmarny — cache lifetime docs)** — Added a CLI-lifetime note on `pathClientCache` warning future contributors not to reach for it from `aicrd` daemon code without bounds.
3. **(@mchmarny — redundant branch in `configmap.go`)** — Collapsed to single `GetKubeClientWithConfig` call; removed the unused `rest` import.
4. **(@mchmarny — test mirror drift)** — `TestSnapshotTemplateFlagCombinations` now calls `parseSnapshotTemplateOptions` directly via a constructed `*cli.Command` (same pattern as `buildDiffCommandWithOutput`). Removed the hand-copied `validateTemplateFlagCombination` helper and the `validationError` shim.
5. **(@mchmarny — inline `cmd.String("kubeconfig")`)** — Pulled a `kubeconfig := cmd.String("kubeconfig")` local in `recipe.go` for consistency with `snapshot.go` / `validate.go` / `diff.go`.
6. **(@coderabbitai — validate `--output` doc inconsistency)** — Updated the row to "file path, ConfigMap URI (`cm://namespace/name`), or stdout" matching snapshot/recipe/diff phrasing.

Lint clean and `go test -race` passes on `pkg/k8s/client`, `pkg/serializer`, `pkg/cli`.

ayuskauskas:
@coderabbitai good catch — pushed `721a434c` pinning HOME (and USERPROFILE for Windows) to a t.TempDir in TestBuildKubeClient_WhitespaceTreatedAsEmpty so a malformed real `~/.kube/config` on a dev box can't false-positive the assertion.

mchmarny (approved):

Issue #1049, opened 2026-05-26
bug(bundler): argocd-helm Application repoURL contains oci:// scheme; ArgoCD OCI Helm sources reject the scheme and ignore the chart field


atif1996:
## Summary

Post #1032 + #1035, the `argocd-helm` contract is `--set repoURL=oci://<registry>/<path>` (parent namespace, **with** `oci://` scheme), and the parent App template appends `.Chart.Name` itself via the separate `source.chart` field. The bundler templates the user-supplied `--set repoURL=…` value verbatim into both `spec.source.repoURL` and `spec.source.helm.valuesObject.repoURL` on the parent `aicr-stack` Application — `oci://` scheme included.

In our test against ArgoCD `v3.4.2` + a real OCI registry, the repo-server rejected the prefixed shape: it appears to treat `oci://`-prefixed `repoURL` as a generic OCI artifact and resolves `HEAD /v2/<path>/manifests/<tag>` with **no chart segment**, instead of the Helm OCI shape `HEAD /v2/<path>/<chart>/manifests/<tag>` you'd expect when `source.chart` is set. The Argo CD [OCI Helm docs](https://argo-cd.readthedocs.io/en/stable/user-guide/oci/) explicitly call out that `oci://` is not included on `repoURL` for Helm OCI sources (`repoURL: registry-1.docker.io/bitnamicharts`).

If that's the canonical Argo contract, then the contract introduced in #1032's `required` message + README ("pass `--set repoURL=oci://<registry>/<path>`") emits an `oci://`-prefixed `repoURL` into the rendered Application that real Argo CD then rejects. The existing test surface doesn't catch this because every regression test (`TestHelmTemplate_*`) is a local `helm template` render that never hits a real Argo repo-server.

**Test caveat:** `aicr` commit `8b4939` (2026-05-20), predates #1032/#1035/#1036/#1038/#1039. Rebuild from `main` is in progress and we'll follow up. Filing now because none of the merged fixes touch `oci://` scheme handling, so this should still reproduce post-fix. Close out if it doesn't.

## Environment

- ArgoCD: `quay.io/argoproj/argocd:v3.4.2`
- Registry: private OCI registry serving Helm-format artifacts
- Cluster: EKS GB200, Ubuntu 24.04

## Reproduction

```bash
aicr bundle --recipe recipe.yaml --deployer argocd-helm \
  --output oci://<registry>/<org>/<path>/aicr-bundle:<tag>

# Per the bundler's own `required` message:
[...]

PR #1070, opened 2026-05-28
docs: explain argocd-helm OCI credential matching

kiwigitops:
## Summary
- document the Argo CD repo-credential URL matching rule for argocd-helm OCI bundles
- call out the private-registry 401 case where local Helm auth succeeds but Argo repo-server credentials are keyed to a different source URL

Related #1049.

## Testing
- `git diff --check`


mchmarny (dismissed):

mchmarny (dismissed):

yuanchen8911 (commented):
/lgtm A couple of minor issues.

docs/user/cli-reference.md
```diff
@@ -2486,6 +2486,11 @@ aicr --debug bundle -r recipe.yaml
...
 2. **Argo CD older than v2.13.** Path-based children rely on Argo CD's generic OCI artifact source type, added in v2.13. Older Argo treats the source as Git and fails to resolve. Check with `kubectl -n argocd get deploy argocd-repo-server -o jsonpath='{.spec.template.spec.containers[0].image}'`. Upgrade Argo, or use `--deployer helm` if Argo upgrade is not an option.
 3. **Tag missing from the registry.** Verify the published artifact exists at the exact tag the parent expects: `oras manifest fetch <registry>/<path>/<chart>:<tag>`. If `aicr bundle` is invoked without a tag (`oci://<registry>/<path>/<chart>` with no `:<tag>` suffix), the CLI version is used as the default — make sure `--set targetRevision=<chart-version>` at install time matches.
+4. **Private registry credentials keyed to a different source URL.** Problem: Argo CD matches repository credentials against the source URL it dereferences.
```
yuanchen8911, line 2589: Minor / non-blocking. Adding this fourth item makes the section intro stale — line 2484 still reads "Three common causes, in order of likelihood:". Consider bumping it to "Four common causes" or just "Common causes:".

(Editorial nuance: a private-registry 401 is a different error signature than the "...not found" target-state in the heading, so it sits a little loosely under "in order of likelihood" — dropping the count resolves both at once.)

docs/user/cli-reference.md
```diff
@@ -2486,6 +2486,11 @@ aicr --debug bundle -r recipe.yaml
 1. **Chart name doubled in `--set repoURL`.** Under the current contract, `--set repoURL` carries the **parent namespace only** (e.g., `oci://ghcr.io/myorg`). The parent Application appends `.Chart.Name` via its `source.chart` field, and path-based children append it directly into their rendered `source.repoURL`. Passing `--set repoURL=oci://ghcr.io/myorg/aicr-bundle` produces a double-suffixed reference (`.../aicr-bundle/aicr-bundle:<tag>`) that does not exist. Drop the trailing chart segment.
```
yuanchen8911, line 2585: Minor / non-blocking — pre-existing wording, not introduced by this PR, but adjacent to your change and worth fixing while we are here. For the OCI case (the whole subject of this section), the parent does **not** use a `source.chart` field — the template appends `.Chart.Name` into `source.repoURL` with `path: "."` (see `aicr-stack.yaml`, the `hasPrefix "oci://"` branch). `source.chart` is only used on the non-OCI Helm-repo branch. Suggested rewording:

```suggestion
1. **Chart name doubled in `--set repoURL`.** Under the current contract, `--set repoURL` carries the **parent namespace only** (e.g., `oci://ghcr.io/myorg`). The parent Application appends `.Chart.Name` into its OCI `source.repoURL`, and path-based children append it directly into their rendered `source.repoURL`. For non-OCI Helm repositories, the parent uses `source.chart` instead. Passing `--set repoURL=oci://ghcr.io/myorg/aicr-bundle` produces a double-suffixed reference (`.../aicr-bundle/aicr-bundle:<tag>`) that does not exist. Drop the trailing chart segment.
```

The same `via its `source.chart`` wording also appears in the install example comment block (line 1593) — that line is outside this diff so I cannot attach a suggestion to it, but it should get the same correction.

mchmarny (approved):

Issue #983, opened 2026-05-20
[Epic]: Library-readiness hardening for in-process Go consumers


mchmarny:
## Summary

AICR's `pkg/recipe` package is currently designed around a single-tenant process model: one CLI invocation or one API server instance, with a process-global `DataProvider` and `sync.Once`-cached metadata store and component registry. That works for our shipping consumers but blocks AICR from being a safe, reusable Go library for **in-process consumers** that need to maintain more than one recipe source at a time — operators, controllers, multi-tenant automation, in-process testing harnesses, and tooling like the BOM generator (see #966).

This is a tracking issue for the related hardening work. Each child issue is independently reviewable and mergeable; the order below reflects landing order, not file order.

## Motivation

Two concrete pressures expose the gap:

1. **#966's BOM-per-leaf rewrite** needs to resolve "effective values" (base → values file → overrides) for every enumerated leaf. The merge logic exists but is package-private, and the bundler — the only public consumer — writes to disk. The BOM rewrite either reimplements values merging (drift hazard) or duplicates bundler internals (coupling hazard) unless we promote the resolution primitives to public API.

2. **In-process consumers** of AICR cannot safely construct two Builders in the same process. The global `DataProvider` plus the `sync.Once`-cached metadata store and component registry mean that whichever Builder's source populates the cache first wins for every concurrent Builder, regardless of what `WithDataProvider`-style option they were given (today, there is no such option). The hazard is dormant only because no in-tree consumer triggers it.

The underlying patterns are already called out as anti-patterns in `.claude/CLAUDE.md`:

> `sync.Once` caching state that depends on a settable global (e.g., a registry tied to a DataProvider) … key the cache by a generation counter the setter increments; recompute on miss so late-bound configuration takes effect.

> Deep-copy helper that recurses into maps but copies `[]any` by reference — slice aliasing leaks mutations across overlay merges.

This work makes those guidelines true of the recipe package itself.

## Scope

Four child issues, in **landing order**:

- [x] **#984** — `fix(recipe): deep-clone ValidationConfig phases in Merge to prevent cache corruption` (correctness bug; standalone, smallest)
- [x] **#985** — `feat(recipe): expose effective-values and manifest helpers on RecipeResult` (unblocks #966; standalone)
- [ ] **#986** — `feat(recipe): per-Builder DataProvider isolation via WithDataProvider option` (largest; enables multi-tenant in-process use)
[...]

PR #1072, opened 2026-05-28
feat(aicr): top-level Go library facade for in-process consumers

### LINKED ISSUE #1071: [Feature]:  Stable embeddable Go library surface for in-process recipe resolution
### Prerequisites

### Feature Summary

A small, semver-tracked Go API that lets external programs use AICR's
recipe resolution, bundling, snapshotting, and validation logic
in-process — without importing AICR's internal packages directly.

### Problem/Use Case

AICR ships as a CLI binary and a REST server today. Both work well
for one-shot invocations from a user or a CI pipeline.

A new class of consumer wants to call AICR's logic from inside a
long-running Kubernetes controller — for example, a Crossplane
provider that reconciles a GPUCluster custom resource and needs to
re-resolve recipes whenever the source data or the cluster's
declared intent changes. These consumers need three things AICR does
not offer today:

A stable Go API contract. Today, external programs have to
import several internal AICR packages and stitch them together
with knowledge of how AICR initializes its data provider,
metadata store, and component registry. Every refactor of those
internals breaks external builds, and consumers have nothing to
pin against except a Git SHA.
Safe concurrent use within a single process. A controller
that reconciles two GPU clusters whose recipe sources differ
needs both resolutions to run side-by-side without
cross-contaminating each other's results. AICR's current caches
are process-global and keyed implicitly to the most recently
configured data source, so two sources active in one process
silently corrupt each other.
A way to release the caches. Long-running processes that ever
touch N distinct recipe sources accumulate N caches with no
eviction path, so memory grows monotonically.

### Proposed Solution

Introduce a top-level package that external consumers can import
directly. The package exposes a single Client type with methods
that mirror AICR's existing workflow stages:

Resolve a recipe from criteria
Bundle the components of a resolved recipe into deployable
Helm values and rendered manifests
Collect a snapshot of live cluster state
Validate a resolved recipe against an observed snapshot
Release the client's caches when done
Each Client owns its own data source and its own caches.
Constructing two clients against different sources is the supported
pattern for the multi-source case; the clients do not share state.

The package becomes the project's semver-tracked surface. Internal
packages remain importable, but the documentation states explicitly
that they sit below the semver line — the same pattern Go's standard
database/sql uses for driver internals.

### Success Criteria

- A new package, importable as `github.com/NVIDIA/aicr`, ships and is
  documented under the integrator guide.
- The CLI and REST server external behavior is unchanged; their
  existing tests pass without modification. Migrating their internal
  implementation to consume the facade is tracked separately and is
  not in scope for this issue (see "Out of Scope" below).
- Two clients constructed from different recipe sources in the same
  process resolve concurrently and return results scoped to their own
  source. A test exercises this directly.
- Releasing a client frees its caches; a test confirms memory does
  not grow when N clients are created and released in a loop.
- A migration note documents how external callers move from the
  internal packages to the new client.
- `make qualify` is green. 

### Alternatives Considered

- **Leaving the internal packages as the public surface and adding
  more documentation.** Rejected — documentation does not fix the
  concurrency bug, and pinning to internal packages forces every
  external consumer to absorb internal refactors.
- **Forking AICR's logic into the downstream consumer.** Rejected —
  duplicates recipe and bundler logic, diverges from AICR's recipe
  catalog over time, and forfeits the value of a curated runtime.
- **A REST-only integration where the controller spawns an `aicrd`
  sidecar.** Rejected — adds a network hop and an extra process to
  every cluster, and complicates the failure modes (sidecar
  liveness, connection pool, retries) for a problem that an
  in-process library call solves directly.

### Component

New component

### Priority

Critical (blocking adoption or major use case)

Strictly additive at the API level. The new package is net-new;
nothing previously importable is removed. The CLI and REST surfaces are unchanged.

We'd welcome the maintainers' guidance on whether the addition
warrants a minor-version bump (e.g., v0.14.0) or whether it can
ship inside an existing planned release.

- **Observability:** the new package exposes the same structured-log
  signals AICR already emits; no new log surface.
- **Upgrades:** downstream consumers track AICR via `go.mod`, so a
  semver-tracked surface lets them adopt patch releases without
  re-validating their own integration.
- **Config drift:** the per-client design eliminates a class of latent
  drift bug (one client's source bleeding into another) that exists
  today in any process running two sources.
- **Multi-tenancy:** the per-client isolation is what makes
  multi-tenant controllers (one client per tenant configuration)
  safe.
- **Blast radius:** zero blast radius on the CLI and REST. The change
  is contained to the new package plus the internal call sites the
  CLI and REST already own.

Yes, I can open a PR

hkii:
## Summary

Introduce `github.com/NVIDIA/aicr` as the project's stable, semver-tracked
Go API for external consumers that need to call AICR's recipe-resolution,
bundling, snapshot, and validation logic in-process — without importing
internal packages directly.

The facade is a sibling of the existing CLI binary and REST server (not a
layer on top); each `Client` owns its own DataProvider, eliminating the
process-global state that previously caused two consumers in one process
to silently corrupt each other's recipe caches.

## Motivation / Context

Fixes NVIDIA/aicr#1071

A new class of consumer — long-running Kubernetes controllers and
similar in-process integrations — needs to call AICR's logic without
spawning a subprocess or running a REST sidecar. Today they have to
stitch together internal `pkg/recipe` and `pkg/bundler` calls, have no
semver-tracked contract to pin against, and hit a latent cross-source
corruption bug when two recipe sources are active in one process. This
PR closes that.

## Type of Change

- [x] New feature (non-breaking change which adds functionality)

## Components Affected

- New top-level package `github.com/NVIDIA/aicr`
- `pkg/recipe` — per-DataProvider caching + `WithDataProvider` option + `Evict` helpers
- `pkg/bundler` — `WithDataProvider` option; all internal reads (values, manifests, registry) routed through `*WithProvider` variants
- `pkg/defaults` — new `SnapshotOperationTimeout` / `ValidationOperationTimeout`
- `validators/deployment` — switched off deprecated `GetManifestContent`
- Docs: integrator guide, public-API matrix

## Implementation Notes

See the commit message for the full breakdown. Highlights:

- **Per-Client isolation** is the load-bearing correctness change. Metadata-
  store and component-registry caches now key on `DataProvider` pointer
  identity; `Client.Close()` explicitly evicts via the existing public
  `Evict` helpers.
- **Shutdown safety**: `Close` drains in-flight `ResolveRecipe` /
  `BundleComponents` / `CollectSnapshot` / `ValidateState` calls before
  evicting. Without this drain, a resolve that released the read lock
  before `LoadMetadataStoreFor` could repopulate `storeCache[dp]` AFTER
  Close's eviction — silently leaking cache entries past Client lifetime.
  Each entry point registers under the read lock via a new
  `inflight sync.WaitGroup`; Close write-locks, releases, then `Wait()`s.
  `TestClient_CloseDrainsInflightResolve` pins this race deterministically
  with a `blockingDataProvider` parking `WalkDir` on a channel.
- **Transient-error resilience**: `LoadMetadataStoreFor` no longer caches
  `context.Canceled` / `context.DeadlineExceeded` errors permanently. The
  `sync.Once` semantics would otherwise poison the entry for the Client's
  lifetime; now the entry is `CompareAndDelete`d on transient error so a
  follow-up call with a healthy ctx loads from scratch.
  `builder.BuildFromCriteria`'s wrap switched to `PropagateOrWrap` so the
  inner `ErrCodeTimeout` survives — callers (and the eviction itself)
  rely on the structured timeout code.
- **Owner-token guard**: `ResolveRecipe` stamps the producing `Client` on
  its returned `RecipeResult`; `BundleComponents` / `ValidateState` reject
  cross-client misuse with `ErrCodeInvalidRequest`. Without this, a
  controller managing N ProviderConfigs could silently mix
  resolved-against-A refs with values-loaded-via-B reads.
- **Bounded timeouts at every facade entry point** so a controller
  passing `context.Background()` can't hang reconciliation on stuck
  Kubernetes I/O.
- **CLI and REST surfaces are unchanged.** Migrating their internal
  implementation to consume the facade is intentionally deferred to a
  follow-up issue to keep this PR scoped and reviewable.

## Testing

- `make qualify` passes locally end-to-end (unit + `-race` + chainsaw
  e2e + yamllint + license headers + sidebar check + vulnerability scan).
- New facade-package tests cover the two acceptance criteria from #1071
  directly:
  - `TestClient_ConcurrentResolveScopesToOwnSource` — two clients, two
    overlay sources, concurrent resolves under `-race`, each returns its
    own marker component and never the other's.
  - `TestClient_NoCacheGrowthAcrossManyCloseCycles` — 50 NewClient →
    ResolveRecipe → Close cycles, asserts metadata-store and registry
    cache counts return to baseline.
  - `TestClient_CloseDrainsInflightResolve` — `blockingDataProvider`
    parks `WalkDir` on a channel; `Close` must wait for the in-flight
    resolve before evicting; assertion uses per-DataProvider
    `CachedStoreContainsForTesting` / `CachedRegistryContainsForTesting`
    (robust under parallel test execution).
- New `pkg/recipe` test
  `TestLoadMetadataStoreFor_TransientErrorIsNotCached` — pre-canceled
  ctx on first load surfaces an error; second call with healthy ctx
  succeeds, proving the transient-error eviction.
- New facade tests for the field-level guards: negative-`Nodes`
  rejection, OS-enables-OS-pinned-overlays, cross-client
  `BundleComponents` / `ValidateState` rejection.

## Risk Assessment

- **Blast radius on existing surfaces: zero.** CLI flags, REST endpoints,
  OpenAPI schema, and existing tests are untouched. The chainsaw e2e
  suite is the regression gate and passes.
- **Public-surface lock-in:** this PR establishes the v0.x contract for
  `github.com/NVIDIA/aicr`. `ValidateOption` is already wrapped as a
  facade-owned functional-option type so `pkg/validator.Option` changes
  don't propagate. The remaining re-exports (`Snapshot`, `AgentConfig`,
  `PhaseResult`, `Phase`, and the phase constants) are still transparent
  aliases of `pkg/snapshotter` / `pkg/validator`; promoting them to
  facade-owned wrappers is worth scoping separately.

## Checklist

- [x] Tests added and passing
- [x] Documentation updated (`docs/integrator/go-library.md`, public-API
      matrix, architecture deep-dive)
- [x] `make qualify` is green
- [x] Commit signed (`-S`)
- [x] No `Co-Authored-By` lines
- [x] PR title under 70 characters


hkii:
Thanks for the review. All 8 fixes landed in `2efc0d9f` (force-pushed). Per-finding:

1. `aicr_test.go:54-56`: hardcoded `/nonexistent/path/...` replaced with `filepath.Join(t.TempDir(), ...)`. Doesn't depend on what files happen to exist on the runner anymore.
2. `aicr_test.go:258`: added `t.Cleanup` to close the client. Good catch on the isolation leak.
3. `aicr.go:148`: `NewClient`'s loop now skips nil `Option` entries instead of panicking. One branch per option, no crash if a caller assembles opts conditionally.
4. `aicr.go:298`: nil-ctx guard added to all four entry points (`ResolveRecipe`, `BundleComponents`, `CollectSnapshot`, `ValidateState`). Returns `ErrCodeInvalidRequest` before `context.WithTimeout` gets a chance to panic.
5. `docs/integrator/go-library.md:178`: doc was pointing at `RecipeBuildTimeout`, but the facade uses `RecipeOperationTimeout`. Rewrote that section to list the four per-operation caps and document the nil-ctx rejection.
6. `docs/integrator/public-api.md:23`: "four-phase" -> "three-phase (Deployment, Performance, Conformance)".
7. `pkg/recipe/metadata_store_test.go:1680`: the transient-error test now asserts the structured `ErrCodeTimeout` code on the first failure, not just `err != nil`. A regression to bare `context.Canceled` would have slipped past the old check.
8. `pkg/recipe/metadata_store.go:127`: godoc still said transient cancellations stay cached forever, which stopped being true after the `CompareAndDelete` change. Rewritten to match actual behavior, and noted that callers no longer need the manual `EvictCachedStore` workaround.

`make qualify` is green locally; CI will retrigger on the force-push.

hkii:
Second round of fixes in `938d5c79` (force-pushed). All 5 new findings addressed:

1. `aicr_test.go:131,135` (around `:138`): both clients in the isolation test now have `t.Cleanup(func() { _ = client.Close() })`. Same cache-leak pattern as the previous round, missed it on the first pass.
2. `aicr_test.go:394-399`: closed-client `BundleComponents` assertion now also pins `ErrCodeInvalidRequest`, not just `err != nil`.
3. `aicr.go` `loadManifestFiles`: takes `ctx context.Context` and checks `ctx.Err()` before each manifest read. Both call sites updated. Worst case after cancellation is one in-flight `provider.ReadFile` (the DataProvider interface itself isn't ctx-aware, so a single read in progress runs to completion). Returns wrapped `ErrCodeTimeout` on mid-load cancellation.
4. `options.go:55-60`: `applyValidateOptions` godoc was wrong about the lock scope. `ValidateState` calls it after releasing `Client.mu`, not inside the locked region. Rewritten to match.
5. `aicr_test.go:200-203`: the concurrency lifecycle test now captures `Close()`'s error via a buffered channel and asserts it after `wg.Wait`. The old `_ = client.Close()` swallowed regressions in drain/eviction.

`make qualify` green locally on the Go path.

hkii:
Third round of fixes in `e9c2a70c` (force-pushed):

1. `docs/integrator/go-library.md:145`: dropped the "Additional sources (HTTP, Git) may be added..." line. Replaced with a one-liner naming `FilesystemSource` as the only production source today. No roadmap promises in integrator docs.
2. `docs/integrator/go-library.md:168-176`: Context handling section reworded. Was implying the per-operation caps fire only when the caller's context has no deadline; the facade actually calls `context.WithTimeout` unconditionally and the effective deadline is `min(caller, cap)`. Now reads that way explicitly.
3. `options.go:119-145`: `WithValidationImagePullSecrets`, `WithValidationTolerations`, `WithValidationNodeSelector` now defensive-copy their inputs via three small helpers (`cloneStringSlice`, `cloneTolerations`, `cloneStringMap`). Each preserves the nil-in/nil-out distinction so `applyValidateOptions`'s "skip the validator.With* call when unset" path still works. Closes the data race where a caller mutating the original map/slice after handoff could trip `serializeNodeSelector`'s map iteration.

`make qualify` green on the Go path.

aicr.go
```diff
@@ -0,0 +1,874 @@
...
+	// silently clobber each other. The per-Builder binding makes
+	// each Client's resolve path use its own cached metadata
+	// store and component registry.
+	// Construction-time write to builder/dp doesn't need the lock —
+	// the Client isn't visible to other goroutines until NewClient
+	// returns — but using the same mu Lock pattern here keeps the
+	// access pattern uniform and makes the field-mutation rule
+	// trivial to verify by grep.
+	c.mu.Lock()
+	c.builder = recipe.NewBuilder(recipe.WithDataProvider(dp))
```
mchmarny, line 180: This is now the third site that wires `DataProvider → Builder → Bundler → Validator`:

- `pkg/cli/root.go:382,396` — uses `recipe.SetDataProvider` (the deprecated global-state path this facade exists to avoid; flagged `//nolint:staticcheck // tracked by #983 Stage 2`)
- `pkg/api/server.go:85-91` — uses `recipe.NewBuilder` + `bundler.New` directly
- here, with `recipe.WithDataProvider`

The PR description defers the CLI/REST migration "to keep this PR scoped." Until that follow-up lands, the codebase has strictly more aggregation surface than before, the CLI is still on the deprecated path, and changes to the wiring policy need to be applied in three places. Per CLAUDE.md ("business logic lives in functional packages so CLI and API can both use it"), the facade should be the single aggregation point — that's the work that pays back this PR.

Open a tracked follow-up issue for CLI + API server adoption and link it here before merge.

types.go
```diff
@@ -0,0 +1,243 @@
...
+)
+
+// Snapshot is the captured cluster-state artifact returned by
+// Client.CollectSnapshot. It is a transparent alias of
+// pkg/snapshotter.Snapshot — no facade wrapper is needed because the
+// type carries no per-Client lifetime state.
+//
+// A Snapshot embeds an api header (apiVersion, kind, metadata) and a
+// Measurements slice, one entry per collector that ran.
+type Snapshot = snapshotter.Snapshot
```
mchmarny, line 37: `type Snapshot = snapshotter.Snapshot` is a transparent alias. Consumers using `aicr.Snapshot` are pinned to `pkg/snapshotter.Snapshot`'s identity. `docs/integrator/public-api.md:31` marks `pkg/snapshotter` as **Internal** ("may change without notice") — so the "Public (stable)" facade is, by transitivity, semver-pinning an Internal type. The promise breaks the first time `snapshotter.Snapshot` gains/renames/removes a field.

Same shape on lines 40 (`AgentConfig` ← `snapshotter`), 46 (`PhaseResult` ← `validator`), 51 (`Phase` ← `validator`), and the phase constants (55-59).

Two ways to fix, pick one before merge:

1. Wrap each: define facade-owned structs that mirror only the fields the facade contract guarantees, translate at the boundary (same pattern `ValidateOption` already uses in `options.go:24`).
2. Promote `pkg/snapshotter` (and the relevant subset of `pkg/validator` if any types beyond `PhaseResult`/`Phase` leak) to "Public (evolving)" in `public-api.md`, and document that `aicr.Snapshot == snapshotter.Snapshot` is a deliberate identity alias under semver.

The PR description acknowledges (1) is deferred. Deferring it is fine if (2) is done now — otherwise the documented contract is dishonest from day one.

docs/integrator/public-api.md
```diff
@@ -0,0 +1,58 @@
...
+| `pkg/bundler` | Public (evolving) | Per-component Helm/Kustomize bundle generation. |
+| `pkg/validator` | Public (evolving) | Constraint evaluation, three-phase validation (Deployment, Performance, Conformance). |
+| `pkg/collector` | Public (evolving) | Observed state collection from clusters. |
+| `pkg/measurement` | Public (evolving) | Typed measurement model used by collectors and validators. |
+| `pkg/version` | Public (evolving) | Semver constraint evaluation. |
+| `pkg/errors` | Public (evolving) | Structured errors with error codes. Consumed at API boundaries. |
+| `pkg/defaults` | Public (evolving) | Shared timeout and limit constants. |
+| `pkg/component` | Internal | Bundler utilities and test helpers. |
+| `pkg/constraints` | Internal | Constraint type definitions. |
+| `pkg/snapshotter` | Internal | Snapshot orchestration. |
```
mchmarny: `pkg/snapshotter` is marked **Internal** here, but `types.go:32, 40` re-export `snapshotter.Snapshot` and `snapshotter.AgentConfig` through transparent aliases on the facade tier marked **Public (stable)**. An Internal package can't be reached "without notice" if its types are identity-aliased into a stable surface — the matrix is currently lying about the actual contract.

Decide which side moves: wrap the aliases at the facade (preserves the matrix as written), or upgrade `pkg/snapshotter` to **Public (evolving)** with a note that selected types are stably aliased from the top-level facade.

Same applies to `pkg/validator` (line 23) — it's `Public (evolving)` not `Internal`, but `aicr.PhaseResult` / `aicr.Phase` being identity-aliased pins those specific types under the facade's stable contract; an "evolving" backing package isn't compatible with that.

pkg/recipe/components.go
```diff
@@ -258,6 +258,41 @@ func EvictCachedRegistry(dp DataProvider) {
...
+// from outside the recipe package would be to reach into unexported
+// state via reflection.
+//
+// Test-only by convention (the _ForTesting suffix); never call from
+// production code.
+//
+// NOTE: global count across every DataProvider. Tests that need a
+// stable signal scoped to a specific DataProvider should prefer
+// CachedRegistryContainsForTesting.
+func CachedRegistryCountForTesting() int {
```
mchmarny, line 274: `CachedRegistryCountForTesting` / `CachedRegistryContainsForTesting` here, and `CachedStoreCountForTesting` / `CachedStoreContainsForTesting` in `metadata_store.go:342, 367`, are exported on `pkg/recipe`, which the matrix lists as "Public (evolving)." The `_ForTesting` suffix is convention-only — nothing prevents production code from calling them, and once they're in the exported surface they're subject to the same evolve-with-care contract as any other public symbol.

Options:
- Move to a `pkg/recipe/internal/cachetest` subpackage (only importable from inside the repo). Facade tests in `aicr_internal_test.go` already live in the right place to use that.
- Move to a `cache_export_test.go` in `pkg/recipe` with `//go:build test` or use `package recipe_test` + a separate `helpers_for_test.go`.
- Accept the export and document them as part of the `pkg/recipe` contract (least preferable — they're test-only by intent).

Inline cycle test in the facade can also drop count assertions and rely entirely on `*ContainsForTesting(dp)` against the facade's own dp — that path doesn't even need the count helpers.

aicr.go
```diff
@@ -0,0 +1,874 @@
...
+//
+// A nil owner means the caller bypassed ResolveRecipe (e.g., constructed
+// the RecipeResult directly). That's a programmer error too — the
+// internal field requires the facade to populate it, and the only public
+// path is ResolveRecipe — so the check rejects nil owner as well.
+//
+// The error embeds %p of both pointers; in controller logs that lets an
+// operator distinguish "wrong Client" from "no Client at all" without
+// adding telemetry surface.
+func (c *Client) assertOwns(r *RecipeResult) error {
```
mchmarny, line 252: The cross-Client guard is a facade-only safety net. It catches misuse for callers going through `aicr.Client`, but the underlying cross-source bleed risk that motivated #1071 sits in `pkg/recipe.Builder` itself — any consumer constructing a Builder directly (today: `pkg/api/server.go:85`) is still exposed.

This is fine as a facade contract, but worth noting that #1071's root cause isn't actually fixed at the layer where the bug lives. A future-leaning fix would either:
- move the owner-token concept down to `recipe.RecipeResult` so non-facade callers get the same protection, or
- accept the asymmetry but document explicitly in `public-api.md` that "Public (evolving)" packages do not carry the cross-source guard the facade provides.

Same point applies to the `inflight` drain in `Close` — facade callers get a clean lifecycle, direct `pkg/recipe` callers manage it themselves.

mchmarny (commented):
Foundational correctness work (inflight drain, transient-error eviction, owner-token guard, bounded timeouts) is excellent and the race-detector tests pin the actual bugs.

Five inline comments along three axes against #1071's contract: (1) the facade creates a third `DataProvider → Builder → Bundler → Validator` wiring path without retiring CLI's `recipe.SetDataProvider` or migrating the REST server — both must follow before this PR's architectural payoff lands; (2) `aicr.Snapshot`/`AgentConfig`/`PhaseResult`/`Phase` are transparent aliases of packages the matrix marks Internal or "Public (evolving)" — semver-stable facade can't honestly identity-alias unstable types; (3) `_ForTesting` helpers leaked onto `pkg/recipe`'s public surface, and the cross-Client guard protects only facade callers, not direct `pkg/recipe.Builder` users (which today includes `pkg/api/server.go`).

Not blocking the spirit of the change — (1) and (3) are fine as tracked follow-ups, (2) needs to be decided before merge so the documented contract is honest from day one.

mchmarny:
Filed follow-up tracking for the deferred work this PR introduces:

- **Epic:** #1076 — Post-facade adoption and surface hardening
  - #1077 — CLI + REST server adopt `aicr.Client` (retires the three-way wiring split and the deprecated `SetDataProvider` call)
  - #1078 — Wrap the facade alias types (`Snapshot`, `AgentConfig`, `PhaseResult`, `Phase`) as facade-owned structs — addresses the `types.go:32` review thread on the public-api.md contradiction
  - #1079 — Move the `_ForTesting` cache helpers off `pkg/recipe`'s public surface
  - #1080 — Push the owner-token guard down to `pkg/recipe.RecipeResult` so direct `pkg/recipe.Builder` consumers get the same cross-source protection

#1078 is the only one I'd consider blocking on for this PR — the others are clean follow-ups. The contradiction between `public-api.md`'s "Internal" tier and the facade's identity-aliased types is the one item that locks in a dishonest contract from day one.

hkii:
Thanks for filing #1076 / #1077 / #1078 / #1079 / #1080 — much appreciated.

For the #1078 contradiction blocker, taking the lighter path in `1deb73f1`: docs now match reality without doing the full wrap, leaving #1078 open as the tracked code-side fix.

Concretely:

1. `docs/integrator/public-api.md`: moved `pkg/snapshotter` from `Internal` → `Public (evolving)`. The previous matrix was claiming "may change without notice" for a package whose types are identity-aliased into a "Public (stable)" facade — that's the dishonesty the thread flagged.
2. Added a new "Facade type aliases" section enumerating the five inherited symbols (`Snapshot`, `AgentConfig`, `PhaseResult`, `Phase`, the three phase constants) and naming their source tier explicitly. Tells consumers these specific symbols inherit `pkg/snapshotter` / `pkg/validator`'s evolving tier rather than the facade's stable tier, with the link to #1078 for the full wrap.
3. `types.go`: each affected type's godoc now carries a "Stability:" note cross-referencing the new section and #1078.

Does this resolve #1078's "blocking on this PR" concern, with the full alias-wrap staying tracked in #1078 for the next milestone? Happy to do the full wrap in this PR instead if you'd prefer that — wanted to check before paying the ~300 LoC cost on speculation.

mchmarny (approved):

mchmarny:
> Does this resolve https://github.com/NVIDIA/aicr/issues/1078's "blocking on this PR" concern?

yes

> Happy to do the full wrap in this PR instead if you'd prefer that

no, lets land this and use the epic issues above to drive the full scope to completion

PR #1085, opened 2026-05-28
refactor(validator): split dependencyAffinity build/validate paths

njhensley:
## Summary

Follow-ups to PR #1066: split `BuildOrchestratorAffinity` so the pre-flight gate validates resolvability without building the affinity tree, add a tripwire test that catches silent k8s-API field drops on `affinityToApplyConfig`, and add a deploy-time selector smoke test that warns when an affinity term would match zero pods.

## Motivation / Context

Three small follow-ups surfaced reviewing #1066:

1. The pre-flight gate in `runPhase` was calling `BuildOrchestratorAffinity` purely for side-effect validation, allocating the full affinity tree and re-emitting `slog.Warn` for missing preferred deps on every run.
2. `affinityToApplyConfig` walks the k8s `Affinity` / `NodeAffinity` / `PodAffinity*` types by hand. A future k8s vendor bump that adds a field (precedent: `MatchLabelKeys` / `MismatchLabelKeys` on `PodAffinityTerm` in v1.29-v1.30) would silently drop it on the wire.
3. Dependency charts can relabel their pods (e.g., `kube-prometheus-stack` upstream chart bumps). A `preferred` term then becomes a no-op and a `required` term blocks scheduling, both without an actionable signal at deploy time.

Related: #933, #1066
Fixes: N/A

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] CLI
- [ ] API server
- [ ] Recipe engine / data
- [ ] Bundlers
- [x] Validator (`pkg/validator`, `pkg/api/validator/v1`)
- [x] Core libraries (`pkg/defaults`)
- [ ] Docs/examples
- [ ] Other

## Implementation Notes

- **Extract `resolveDeps`** from `BuildOrchestratorAffinity` so both the build path and a new validate-only path share resolution semantics. `ValidateDependencyAffinity(deps, refs) error` performs resolvability checks without constructing `corev1.Affinity` and suppresses the `slog.Warn` that the build path emits (pre-flight is a quiet check; build-time is where missing-preferred warnings belong).
- `runPhase` pre-flight switched from `BuildOrchestratorAffinity` to `ValidateDependencyAffinity`. `TestValidateDependencyAffinity_MatchesBuildSemantics` locks the two paths to identical error/no-error verdicts so they cannot drift.
- **`TestAffinityTypeFieldCountInvariant`** asserts the field count of each k8s type the hand-rolled `affinityToApplyConfig` walker iterates. A k8s vendor bump that adds a field fails this test and forces an audit before the field is silently lost on the wire. Test docstring spells out the audit procedure.
- **`scanMissingPodAffinityDeps`** in `Deployer.DeployJob`: for each `PodAffinityTerm` in the rendered affinity, list one pod with the term's `LabelSelector` in the term's namespace. If zero match, emit `slog.Warn`; if the List itself errors, emit a different `slog.Warn`. Per-namespace timeout via new `defaults.PodAffinitySelectorLookupTimeout` (5s) so a slow apiserver cannot delay deploy.
  - Logging only — does not fail-close. `preferred` terms are best-effort by design; `required` terms are already surfaced as Pending by the scheduler, which is more authoritative than a point-in-time List.

## Testing

```bash
make test
golangci-lint run -c .golangci.yaml ./pkg/api/validator/v1/... ./pkg/validator/... ./pkg/defaults/...
```

- `golangci-lint`: 0 issues on affected packages.
- New tests:
  - `TestAffinityTypeFieldCountInvariant` (6 subcases)
  - `TestValidateDependencyAffinity_MatchesBuildSemantics` (5 subcases)
  - `TestScanMissingPodAffinityDeps` (5 subcases — nil/empty/required-match/required-miss/preferred-miss/mixed)
  - `TestScanMissingPodAffinityDeps_ListErrorReturnsWarning`

**Coverage delta** (vs `upstream/main`):
- `pkg/api/validator/v1`: 73.6% → 74.3% (+0.7%)
- `pkg/validator/job`: 15.7% → 25.9% (+10.2%)
- `pkg/validator`, `pkg/validator/catalog`, `pkg/validator/ctrf`: unchanged

## Risk Assessment

- [x] **Low** — Pure refactor + additive defensive checks. `BuildOrchestratorAffinity` callers see no behavior change (it still returns the same `*corev1.Affinity` and the same error class); the new `ValidateDependencyAffinity` is verifiably semantics-equivalent via the matching test. `scanMissingPodAffinityDeps` is logging-only and bounded by a short per-namespace timeout. The field-count test fails at compile/test time only on k8s vendor bumps that demand the audit anyway.

**Rollout notes:** No migration. Existing recipes are unaffected.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — no user-visible behavior change)


pkg/validator/job/deployer.go
```diff
@@ -143,6 +154,75 @@ func (d *Deployer) DeployJob(ctx context.Context) error {
...
+		// "selector lookup failed; error=context canceled" warning per
+		// remaining (term, namespace) pair, turning one cancellation into
+		// N misleading apiserver-flake warnings.
+		if ctx.Err() != nil {
+			return warnings
+		}
+		if term.LabelSelector == nil || len(term.Namespaces) == 0 {
+			continue
+		}
+		selector := metav1.FormatLabelSelector(term.LabelSelector)
```
mchmarny: `metav1.FormatLabelSelector` is a display helper, not a query-parameter builder — it returns the literal string `"<error>"` on parse failure and `"<none>"` on an empty selector (see `vendor/k8s.io/apimachinery/pkg/apis/meta/v1/helpers.go:170`). The apiserver then rejects `LabelSelector=<error>` as malformed and this scan emits a misleading "selector lookup failed" warning.

For AICR-built affinity this can't bite today because `BuildOrchestratorAffinity` only constructs `MatchLabels`. But `scanMissingPodAffinityDeps` takes an arbitrary `*corev1.PodAffinity` from `plan.Affinity`, so any future path that injects `MatchExpressions` (e.g. operator override, recipe extension) silently breaks the diagnostic.

Use `metav1.LabelSelectorAsSelector(term.LabelSelector)` and surface the error as one of the existing warning categories:

```go
sel, err := metav1.LabelSelectorAsSelector(term.LabelSelector)
if err != nil {
    warnings = append(warnings, fmt.Sprintf(
        "dependencyAffinity selector is malformed; skipping lookup: selector=%+v error=%v",
        term.LabelSelector, err))
    continue
}
selector := sel.String()
```

pkg/validator/job/deployer.go
```diff
@@ -143,6 +154,75 @@ func (d *Deployer) DeployJob(ctx context.Context) error {
...
+	var warnings []string
+	for _, term := range terms {
+		// Stop early on cancellation. Continuing here would only emit a
+		// "selector lookup failed; error=context canceled" warning per
+		// remaining (term, namespace) pair, turning one cancellation into
+		// N misleading apiserver-flake warnings.
+		if ctx.Err() != nil {
+			return warnings
+		}
+		if term.LabelSelector == nil || len(term.Namespaces) == 0 {
```
mchmarny, line 250: Two silent-skip cases worth a one-line guard comment so a future reader knows they're intentional:

1. `term.NamespaceSelector` isn't handled — per the k8s `PodAffinityTerm` godoc, the effective namespace set is `Namespaces ∪ NamespaceSelector`, and when both are empty/nil it falls back to the pod's own namespace. AICR's `BuildOrchestratorAffinity` always populates `Namespaces` from the resolved `ref.Namespace` (affinity.go:105), so this scanner only needs to handle `Namespaces`. Worth saying so in a comment — otherwise the next person to add a `NamespaceSelector` path elsewhere won't realize this scan goes blind.

2. `LabelSelector == nil` silently continues. Same story — AICR always sets it, but a defensive `slog.Debug` would help triage if something upstream ever produces a malformed term.

Neither is a bug today; both are footguns for the next change.

pkg/validator/job/deployer.go
```diff
@@ -122,6 +122,17 @@ func (d *Deployer) DeployJob(ctx context.Context) error {
...
 	d.jobName = plan.JobName
 
+	// Best-effort: warn if any dependencyAffinity selector matches zero
+	// pods at deploy time. Catches silent label drift on dependency-chart
+	// bumps (e.g., kube-prometheus-stack relabels its Prometheus pods).
+	// Logging only — we don't block deploy, because the scheduler will
+	// surface a hard match miss as Pending if the affinity is `required`.
+	if plan.Affinity != nil && plan.Affinity.PodAffinity != nil {
+		for _, w := range scanMissingPodAffinityDeps(ctx, d.clientset, plan.Affinity.PodAffinity) {
+			slog.Warn(w, "validator", d.entry.Name)
```
mchmarny: nit: `slog.Warn(w, "validator", d.entry.Name)` passes the full pre-formatted message as the `msg` arg, which defeats the structured-attr advantage — `namespace=%q selector=%q` ends up jammed into `msg` rather than as queryable attrs. AGENTS.md's logging rules call this out: prefer fixed messages + key/value attrs.

If you want to keep the function testable as a `[]string`, the simplest fix is to return a small struct:

```go
type affinityScanWarning struct {
    Namespace string
    Selector  string
    Reason    string // "zero-match" or "lookup-failed: <err>"
}
```

Then the caller writes:

```go
slog.Warn("dependencyAffinity selector matched zero pods at deploy time",
    "validator", d.entry.Name,
    "namespace", w.Namespace,
    "selector", w.Selector,
    "reason", w.Reason)
```

Not blocking — the current form greps fine, just doesn't slice/dice in structured-log tools.

pkg/api/validator/v1/affinity_test.go
```diff
@@ -194,3 +196,116 @@ func TestBuildOrchestratorAffinity_ExplicitTopologyKeyPreserved(t *testing.T) {
...
+		{"PodAntiAffinity", reflect.TypeOf(corev1.PodAntiAffinity{}), 2},
+		{"PodAffinityTerm", reflect.TypeOf(corev1.PodAffinityTerm{}), 6},
+	}
+	for _, c := range cases {
+		t.Run(c.name, func(t *testing.T) {
+			if got := c.t.NumField(); got != c.want {
+				t.Fatalf("corev1.%s has %d fields, expected %d. K8s API has added/removed a field; audit the apply-config converter in pkg/api/validator/v1/job_plan.go and update this test only after confirming the new field is handled or intentionally excluded.",
+					c.name, got, c.want)
+			}
+		})
```
mchmarny, line 228: Nicely done — reflection-based field-count invariant + a docstring that spells out the audit procedure is exactly the pattern you want for a hand-rolled walker against a versioned upstream API. Future k8s bumps that add fields will fail this loudly with the right next step embedded in the failure message.

mchmarny (commented):
Clean refactor — splitting `BuildOrchestratorAffinity` so the pre-flight gate can validate without allocating the affinity tree (or double-firing the missing-preferred warning) is exactly the right shape, and the `MatchesBuildSemantics` test locks the two paths against drift. The `TestAffinityTypeFieldCountInvariant` tripwire is a particularly nice addition.

Two non-blocking findings on `scanMissingPodAffinityDeps`: `metav1.FormatLabelSelector` is a display helper that returns the literal string `"<error>"` on parse failure, so the diagnostic silently degrades if a non-`MatchLabels` selector ever reaches the scanner — defense-in-depth fix is `LabelSelectorAsSelector(...).String()`. And `NamespaceSelector` + nil-`LabelSelector` cases are silently skipped today, which is correct for AICR-produced affinity but worth a one-line guard comment so future contributors know. Plus a slog-attr-vs-message nit.

CI green, behavior change is logging-only, low risk. Note: PR is `behind main` — rebase before merge.

pkg/validator/job/deployer.go
njhensley: Done in the staged changes — switched to `metav1.LabelSelectorAsSelector(term.LabelSelector).String()` and emit a new `affinityScanReasonMalformedSelector` warning on parse failure. Added `TestScanMissingPodAffinityDeps_MalformedSelectorEmitsWarning` exercising an invalid operator (`MatchExpressions[].Operator = "Bogus"`) to lock in the new branch.

pkg/validator/job/deployer.go
njhensley: Done — added an "Intentional coverage gaps" paragraph to `scanMissingPodAffinityDeps`'s doc explicitly calling out both cases (NamespaceSelector not resolved; nil LabelSelector silently skipped), plus an inline comment at the silent-skip line pointing back to the doc. Left as silent-continue (no `slog.Debug`) since the coverage gap is documented; happy to add the debug log if you'd prefer.

pkg/validator/job/deployer.go
njhensley: Done with the struct refactor you sketched — `scanMissingPodAffinityDeps` now returns `[]affinityScanWarning` (`Message`/`Namespace`/`Selector`/`Reason`/`Err`). `DeployJob` emits stable msg + structured attrs:

```go
slog.Warn(w.Message,
    "validator", d.entry.Name,
    "namespace", w.Namespace,
    "selector", w.Selector,
    "reason", w.Reason,
    "error", w.Err)
```

`Reason` is one of `zero-match` / `lookup-failed` / `malformed-selector` — short ASCII so it works as a metric label too. Tests were updated to assert on the struct fields instead of substring-matching the old pre-formatted strings.

mchmarny (approved):

PR #1092, opened 2026-05-28
fix(kwok): single-snapshot verify_pods + settle retry (#1090)

### LINKED ISSUE #1090: bug(kwok): verify_pods samples pod state across 5 kubectl calls — TOCTOU race produces impossible math + misleading diagnostic
## Summary

`kwok/scripts/validate-scheduling.sh::verify_pods()` takes five sequential `kubectl get pods` snapshots and treats them as a coherent view of cluster state. They are not — controllers reconcile between calls, producing impossible Pod-status math and a misleading FailedScheduling diagnostic that has misdirected multiple recent investigations.

This is a CI harness race (TOCTOU), not a Go race and not a recipe regression. It explains the recent uptick in KWOK Tier-1 flakes across unrelated branches (#1053, #1086, #1046, and others).

## Symptom (signature)

The job log shows arithmetic that cannot come from any single Kubernetes state:

```
Pod status: 20 total, 20 running, 0 pending (1 unscheduled), 0 failed
ERROR: Scheduling validation FAILED
ERROR: 1 pods could not be scheduled to nodes
ERROR: Unscheduled pods:
       (kubectl returns "No resources found")
ERROR: Events for unscheduled pods:
       kube-system  coredns-…  FailedScheduling  untolerated taint(s)
       kube-system  coredns-…  FailedScheduling  untolerated taint(s)
       local-path-storage  local-path-provisioner-…  FailedScheduling  untolerated taint(s)
```

`unscheduled` is a subset of `pending`, so `pending=0` and `unscheduled=1` are mutually exclusive in a single state snapshot. The diagnostic `kubectl get pods --phase=Pending -o wide` then returns "No resources found" because the bundle pod that was momentarily Pending has already been bound by the scheduler. The event dump that follows is unfiltered, so kube-system FailedScheduling noise (from the `nfd-excluded=true:NoSchedule` taint applied to the kind control-plane node) surfaces as if it were the cause — but those pods are explicitly excluded from the actual count.

## Evidence

Two independent samples on different branches:

- **Job `78343519974`** (PR #1053, recipe `aks-inference`, deployer `argocd-helm-oci`, run `26589273745` attempt 1):
  `20 total, 20 running, 0 pending (1 unscheduled), 0 failed` — then "No resources found" on the diagnostic. Same SHA passed on rerun.
- **Job `78341110233`** (recipe + deployer differ): `2 total, 2 running, 1 pending (3 unscheduled)` — a fresh `nvidia-dra-driver` pending pod appeared between snapshots. Same race class.

Issue #1061 fixed a sibling race (chainsaw sync gate firing too early — PR #1062). That fix tightened the gate but does not close the in-script sampling race in `verify_pods()`, which is still latent.

## Root cause

`kwok/scripts/validate-scheduling.sh:1283-1296` issues 5 separate `kubectl get pods` calls, each ~hundreds of ms apart due to `-o json | jq` overhead:

```bash
total_pods=$(kubectl get pods … | wc -l)                              # T0
pending_pods=$(kubectl get pods --phase=Pending … | wc -l)            # T1
failed_pods=$(kubectl get pods --phase=Failed … | wc -l)              # T2
running_pods=$(kubectl get pods --phase=Running … | wc -l)            # T3
unscheduled_pods=$(kubectl get pods --phase=Pending -o json | jq …)   # T4
```

During those several seconds, operators (gpu-operator, cert-manager, prometheus, dynamo, slurm, dra-driver, …) continue reconciling and creating new pods. A new Pending-unscheduled pod that appears between T1 and T4 makes `pending=0, unscheduled=1` possible. By the time the diagnostic at line 1326-1327 re-queries, the scheduler has bound it — hence "No resources found".

The two earlier (incorrect) hypotheses that this same signature has provoked:
- **"It's the `nfd-excluded` taint"** — wrong: kube-system and local-path-storage are explicitly excluded from the unscheduled count at line 1281; only the event dump at line 1330 is unfiltered, which is why the noise appears in the log but cannot be the trigger.
- **"It's the Argo CD sync gate firing too early"** (#1061) — that was a real race that #1062 fixed, but it's distinct from this one. Tightening the sync gate just changed the timing without removing the in-script race.

Why this surfaces more often recently:
- #1050/#1062 shortened the time between deploy-complete and `verify_pods` running, leaving less natural settle time before the race window opens.
- The bundle now includes more components (slurm, dynamo, grove, nvsentinel, kgateway, kai-scheduler), each with its own operator and reconcile cadence, widening the create-pod window between snapshots.

## Fix

All three changes in `kwok/scripts/validate-scheduling.sh:1283-1336`:

1. **Single snapshot:** replace the 5 separate `kubectl get pods` calls with one `kubectl get pods --all-namespaces -o json` snapshot and derive total/running/pending/failed/unscheduled from that single coherent view with `jq`.
2. **Reuse the snapshot for diagnostics:** the "Unscheduled pods:" listing at line 1326-1327 must be rendered from the same snapshot, not a fresh `kubectl get`. This eliminates the "No resources found" diagnostic that has misled reviewers.
3. **Filter the FailedScheduling event dump** at line 1330 to the same namespace exclusions used by the count (kube-system, kube-node-lease, kube-public, local-path-storage, kwok-system, argocd, flux-system, aicr-registry) or to the exact set of unscheduled-pod names. This kills the `nfd-excluded` red herring that has provoked at least two wrong diagnoses.

Optional belt-and-suspenders (deferrable): if `unscheduled_pods > 0` after a single snapshot, sleep ~5s and re-snapshot once before declaring failure. Covers the legitimate-but-transient case where the scheduler genuinely hasn't bound yet.

Estimated size: ~20 lines net, single commit, doc-only / script change. No Go, no chart, no recipe.

## Reproduction

Race; not deterministic. Best signal:

```bash
# Triggers ~30-40% per attempt across recent KWOK runs:
gh run rerun <recent-failing-kwok-run-id> --failed
```

Or search the `KWOK Cluster Validation` workflow history for runs failing with the `Pod status: X total, X running, 0 pending (Y unscheduled)` signature where `Y > 0`.

## Acceptance criteria

- [ ] `verify_pods()` derives all pod counts from one `kubectl get pods -o json` snapshot
- [ ] "Unscheduled pods:" diagnostic uses the same snapshot — never returns "No resources found" when count > 0
- [ ] FailedScheduling event dump filters by the same namespace exclusions as the count
- [ ] Add a unit-style chainsaw test or shellcheck-compatible smoke check, if feasible — non-blocking
- [ ] Three consecutive clean KWOK runs on `main` after merge (sanity check that the dominant flake class is gone)

## Related

- #1061 (closed) — sibling sync-gate race, fixed by #1062. Different mechanism, same symptom class.
- PR #1053 KWOK Tier-1 failure (run 26589273745 attempt 1) — passed on retry, evidence captured above.

yuanchen8911:
## Summary

Rewrites `verify_pods()` in `kwok/scripts/validate-scheduling.sh` to take one coherent `kubectl get pods -o json` snapshot per attempt, polls on a bounded scheduler-settle window when pods are still unscheduled, and derives every count, the pod distribution, and the failure diagnostic from the same snapshot — eliminating both the TOCTOU race (impossible math) and the post-sync-gate scheduler-latency flake.

## Motivation / Context

`verify_pods()` had two flake classes producing false failures on KWOK Tier-1 lanes:

**1. TOCTOU race across five separate `kubectl get pods` calls.** total / pending / failed / running / unscheduled were sampled sequentially over several seconds (the unscheduled query alone takes hundreds of ms due to `-o json | jq`). Between calls, operator controllers (gpu-operator, dynamo, dra-driver, …) kept creating pods, so the snapshots disagreed:

```
Pod status: 20 total, 20 running, 0 pending (1 unscheduled), 0 failed
ERROR: Unscheduled pods:
       (kubectl returned "No resources found")
ERROR: Events for unscheduled pods:
  kube-system  coredns-…  FailedScheduling  untolerated taint(s)
```

`unscheduled` is a subset of `pending`, so `pending=0 + unscheduled=1` cannot come from one state. The unfiltered event dump then surfaced kube-system noise from the intentional `nfd-excluded=true:NoSchedule` taint, misleading multiple prior reviews.

**2. Post-sync-gate scheduler latency.** Even after switching to a single coherent snapshot, two lanes failed with internally-consistent math but a real Pending pod that simply hadn't been bound yet:

- `eks-inference (argocd-oci)`: `cert-manager/cert-manager-786cc54854-kngvm` — `Pod status: 2 total, 0 running, 2 pending (1 unscheduled), 0 failed`, zero FailedScheduling events
- `oke-training (argocd-oci)`: `skyhook/skyhook-operator-controller-manager-596b99b857-8shfm` — `Pod status: 12 total, 11 running, 1 pending (1 unscheduled), 0 failed`, zero FailedScheduling events

Zero FailedScheduling events means the scheduler hadn't tried to schedule yet — pure latency between Argo CD reporting Synced+Healthy and the scheduler catching up. A single instantaneous snapshot would fail-fast on this transient.

Issue #1090 has the full evidence trail (jobs `78343519974`, `78341110233` for the TOCTOU race; jobs `78353433292`, `78353436030` for the post-sync latency).

Fixes: #1090
Related: #1061 (closed, sibling sync-gate race fixed by #1062 — distinct from this one)

## Type of Change

## Component(s) Affected

- [x] Other: KWOK CI harness (`kwok/scripts/validate-scheduling.sh`)

## Implementation Notes

Six changes inside `verify_pods()`:

1. **Single snapshot per attempt.** One `kubectl get pods --all-namespaces -o json` per loop iteration; all five counts (total / running / pending / failed / unscheduled) plus pod distribution derived from it via jq. Within one iteration the math is guaranteed coherent — impossible-math signatures cannot reappear.

2. **Bounded scheduler-settle poll loop.** When `unscheduled_pods > 0`, the loop re-snapshots up to `KWOK_VERIFY_PODS_DEADLINE` seconds (default `60`) at `KWOK_VERIFY_PODS_INTERVAL` spacing (default `5`). Early-exits as soon as a snapshot shows `unscheduled == 0` (or any `Failed > 0`, which is terminal). Declares failure only if the deadline elapses with `unscheduled` still positive — uses the FINAL snapshot for diagnostics so the named pod is the one that was still unscheduled at the deadline.

3. **Diagnostics from the same snapshot.** "Unscheduled pods:" listing and pod distribution come from the final snapshot — never prints "No resources found" when the count is positive.

4. **Filtered FailedScheduling event dump.** Piped through `grep -vE "^(${exclude_ns})\s"` so kube-system / local-path-storage noise from the intentional `nfd-excluded` taint stops surfacing as if it were the cause.

5. **Fail-closed on kubectl error.** Pre-existing `2>/dev/null | wc -l` silently produced `total_pods=0` on apiserver outage / auth failure / bad context and flowed through the "bundle may be empty" PASS branch. Now captures kubectl's stderr into a `mktemp` file and surfaces it in `log_error` before `return 1`.

6. **(Drive-by, same shape)** Rendering the failed-pod diagnostic from the same snapshot for consistency.

### Tunables

| Env var | Default | Purpose |
|---|---|---|
| `KWOK_VERIFY_PODS_DEADLINE` | `60` | Total seconds to wait for the scheduler to bind remaining Pending pods before declaring failure. |
| `KWOK_VERIFY_PODS_INTERVAL` | `5` | Seconds between re-snapshots within the deadline window. |

Per-iteration cost: 1 kubectl call + 5 jq calls (~0.5s total). With default deadline, max ~13 iterations. Loop exits early on success, so the additional cost is paid only when scheduler latency is real — for hung pods that would have failed anyway, this adds the deadline as a fixed cost (worth it for stability).

### What this PR does NOT change

- The intentional `nfd-excluded=true:NoSchedule` taint on the kind control-plane (still applied; only the diagnostic stopped surfacing its noise).
- The Argo CD sync-gate behavior from #1062 (separate race, separately fixed).
- Recipe overlays or any Go code.
- Genuine scheduling failures (missing toleration, no matching `nodeSelector`, insufficient resources): those persist through the full deadline and are reported with the same diagnostic.

## Testing

```bash
# Static checks
shellcheck -x kwok/scripts/validate-scheduling.sh   # 0 errors

# Local smoke test with synthetic snapshot
# Covered: excluded namespaces (kube-system, local-path-storage, argocd),
#          Job-owned helm hooks, Failed pods, ContainerCreating pods.

# Project gate
unset GITLAB_TOKEN
make qualify    # 0 issues lint, 22/22 chainsaw, no vulnerabilities, license OK
```

**End-to-end verification:** the prior CI run on this PR (commit `d702b9e9`) hit the post-sync-gate scheduler latency on two argocd-oci lanes — exactly the case the new poll loop addresses. The current revision (`e4827765`) is the response to that finding. The next KWOK run on this PR is the natural integration test; if any lane still fails, the failure will name a specific pod that remained unscheduled for the full 60s deadline, which is a real bug not a flake.

**Coverage:** script-only change (no Go), so the per-package coverage gate does not apply per CLAUDE.md.

## Risk Assessment

- [x] **Low** — Single-function rewrite in a CI-only harness script. Success-path log unchanged on the happy path; failure-path output strictly more useful. The poll loop adds bounded latency (default 60s max) only when pods are not yet scheduled — already-scheduled bundles see zero added latency. Easy to revert (one file, one commit).

**Rollout notes:** None. Takes effect on first run after merge. No flag, no migration. The two new env vars default to the values used during development; operators only need to set them if they want a shorter/longer settle window.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — covered by `make qualify`
- [x] Linter passes (`make lint`) — 0 issues
- [x] I added/updated tests for new functionality — N/A, no Go path; smoke-tested locally and end-to-end on the prior PR commit
- [x] I updated docs if user-facing behavior changed — N/A, CI internals only

### DIFF
--- a/kwok/scripts/validate-scheduling.sh
+++ b/kwok/scripts/validate-scheduling.sh
@@ -1277,45 +1277,152 @@ verify_pods() {
     # Exclude system namespaces that aren't part of our bundle (control-plane
     # infra plus GitOps controller namespaces — argocd/flux-system/aicr-
     # registry are owned by install-infra.sh, not the bundle under test).
-    local ns_filter="--all-namespaces"
     local exclude_ns="kube-system|kube-node-lease|kube-public|local-path-storage|kwok-system|argocd|flux-system|aicr-registry"
 
+    # Take a coherent snapshot of pod state and derive every count, the
+    # pod distribution, and the failure diagnostic from it. The previous
+    # implementation issued 5+ separate `kubectl get pods` calls over
+    # several seconds; controllers (gpu-operator, dynamo, dra-driver, …)
+    # reconciled new pods between calls, producing impossible math (e.g.
+    # pending=0 + unscheduled=1) plus a "No resources found" diagnostic
+    # when the transient pod had already been bound by the scheduler.
+    # See #1090 for the failure signature.
+    #
+    # Poll-then-verify: after the Argo CD / Flux sync gate reports
+    # Synced+Healthy, controllers continue creating pods that haven't
+    # been bound by the scheduler yet. Without a settle window, every
+    # snapshot taken in that gap reports the pre-bind transient as a
+    # failure (observed on the eks-inference/argocd-oci and
+    # oke-training/argocd-oci lanes — cert-manager and skyhook-operator
+    # respectively, both with zero FailedScheduling events, meaning
+    # the scheduler simply had not gotten to them yet). The loop below
+    # re-snapshots up to KWOK_VERIFY_PODS_DEADLINE seconds at
+    # KWOK_VERIFY_PODS_INTERVAL spacing and only declares failure if
+    # unscheduled_pods is still positive at the deadline. Each snapshot
+    # is internally coherent (single `kubectl get pods -o json`), so the
+    # impossible-math signature from #1090 cannot reappear regardless
+    # of how many iterations the loop runs.
+    local deadline="${KWOK_VERIFY_PODS_DEADLINE:-60}"
+    local interval="${KWOK_VERIFY_PODS_INTERVAL:-5}"
+
+    # Validate the tunables before the loop relies on them. The settle
+    # window is only bounded if `waited` advances every iteration:
+    #   - interval=0 → `sleep 0` returns instantly AND `waited+=0` never
+    #     grows, so an unschedulable pod (Pending, failed=0) spins the
+    #     loop forever until the CI job timeout — defeating the deadline.
+    #   - a non-integer value (e.g. "abc", "2.5") aborts opaquely under
+    #     `set -euo pipefail` at `sleep`/`$(( ))` with no clear cause.
+    #   - a leading-zero value (e.g. "08", "09") matches the digit regex
+    #     but bash arithmetic reads it as octal ("08: value too great for
+    #     base"), so the `waited >= deadline` test and `$(( ))` increment
+    #     misbehave. Canonicalize to base 10 with `10#...` after the regex
+    #     so the loop only ever sees clean decimal integers.
+    # Reject non-digits up front, then normalize. deadline may be 0
+    # (fail fast, no settle); interval must be >= 1.
+    if ! [[ "$deadline" =~ ^[0-9]+$ ]]; then
+        log_error "KWOK_VERIFY_PODS_DEADLINE must be a non-negative integer (got: '${deadline}')"
+        return 1
+    fi
+    if ! [[ "$interval" =~ ^[0-9]+$ ]]; then
+        log_error "KWOK_VERIFY_PODS_INTERVAL must be a positive integer (got: '${interval}')"
+        return 1
+    fi
+    deadline=$((10#$deadline))
+    interval=$((10#$interval))
+    if (( interval < 1 )); then
+        log_error "KWOK_VERIFY_PODS_INTERVAL must be a positive integer >= 1 (got: '${KWOK_VERIFY_PODS_INTERVAL:-}')"
+        return 1
+    fi
+
+    local waited=0
+    local attempt=0
+    local snap
     local total_pods pending_pods failed_pods running_pods unscheduled_pods
-    total_pods=$(kubectl get pods ${ns_filter} --no-headers 2>/dev/null | { grep -vE "^(${exclude_ns})\s" || true; } | wc -l | tr -d ' ')
-    pending_pods=$(kubectl get pods ${ns_filter} --field-selector=status.phase=Pending --no-headers 2>/dev/null | { grep -vE "^(${exclude_ns})\s" || true; } | wc -l | tr -d ' ')
-    failed_pods=$(kubectl get pods ${ns_filter} --field-selector=status.phase=Failed --no-headers 2>/dev/null | { grep -vE "^(${exclude_ns})\s" || true; } | wc -l | tr -d ' ')
-    running_pods=$(kubectl get pods ${ns_filter} --field-selector=status.phase=Running --no-headers 2>/dev/null | { grep -vE "^(${exclude_ns})\s" || true; } | wc -l | tr -d ' ')
 
-    # Count truly unscheduled pods (Pending with no node assigned)
-    # Pods in ContainerCreating are Pending but scheduled - they have a node
-    # Exclude cleanup/webhook Jobs - these are Helm hooks that may not have proper tolerations
-    # Use awk to count lines, avoiding issues with empty output or newlines
-    unscheduled_pods=$(kubectl get pods ${ns_filter} --field-selector=status.phase=Pending \
-        -o json 2>/dev/null | \
-        jq -r '.items[] | select(.metadata.namespace as $ns | "'${exclude_ns}'" | split("|") | map(. == $ns) | any | not) | select(.spec.nodeName == null or .spec.nodeName == "") | select(.metadata.ownerReferences == null or (.metadata.ownerReferences | map(.kind) | contains(["Job"]) | not)) | .metadata.name' | \
-        awk 'NF {count++} END {print count+0}')
+    while :; do
+        attempt=$((attempt + 1))
 
-    log_info "Pod status: $total_pods total, $running_pods running, $pending_pods pending ($unscheduled_pods unscheduled), $failed_pods failed"
+        # Fail closed on kubectl error: apiserver outage, auth/RBAC
+        # failure, or a bad kube context must not be reported as a
+        # scheduling pass. The pre-existing `2>/dev/null | wc -l` chain
+        # did abort under `set -euo pipefail` (pipefail propagated
+        # kubectl's non-zero status), but opaquely — stderr was
+        # discarded, so the job log showed a bare non-zero exit with no
+        # cause. Capture stderr and emit an explicit message + clean
+        # `return 1` instead.
+        local snap_err_file
+        snap_err_file=$(mktemp)
+        if ! snap=$(kubectl get pods --all-namespaces -o json 2>"$snap_err_file"); then
+            log_error "Failed to list pods (apiserver unreachable, auth/RBAC failure, or bad kube context):"
+            log_error "$(cat "$snap_err_file")"
+            rm -f "$snap_err_file"
+            return 1
+        fi
+        rm -f "$snap_err_file"
 
-    # Show pod distribution across nodes.
-    # grep -vE returns exit 1 when no lines match (e.g. all pods are in
-    # excluded namespaces — happens on the flux-oci lane where every pod
-    # the controller has placed so far lives in flux-system / kube-system).
-    # Under `set -euo pipefail` that exit 1 propagates as the pipeline's
-    # exit code and kills the script silently AFTER Pod-status logging,
-    # masquerading as a pass-then-fail. Wrap with `{ ... || true; }` so
-    # an empty result is a no-op, matching the total_pods / unscheduled_pods
-    # extraction above.
+        # All counts derive from the same $snap in a SINGLE jq pass: the
+        # namespace exclusion is applied once (`as $f`), then each count
+        # is a sub-filter over $f, emitted as one tab-separated line and
+        # split by `read`. One jq invocation per iteration (vs five)
+        # reinforces the "all counts from one snapshot" invariant. The
+        # exclusion filter uses the same pipe-delimited list passed via
+        # --arg pat and split inside jq, so the bash regex pattern and
+        # the jq namespace filter agree by construction. The fifth field,
+        # "truly unscheduled", is Pending with no node assigned, excluding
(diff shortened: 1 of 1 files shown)


kwok/scripts/validate-scheduling.sh
```diff
@@ -1277,57 +1277,159 @@ verify_pods() {
...
+        #                          the final snapshot
+        if [[ "$unscheduled_pods" -eq 0 ]] \
+                || [[ "$failed_pods" -gt 0 ]] \
+                || [[ "$waited" -ge "$deadline" ]]; then
+            break
+        fi
+
+        log_info "Attempt ${attempt}: $total_pods total, $running_pods running, $pending_pods pending ($unscheduled_pods unscheduled), $failed_pods failed — waiting for scheduler (${waited}/${deadline}s)"
+        sleep "$interval"
+        waited=$((waited + interval))
```
mchmarny, line 1405: nit (not blocking): `KWOK_VERIFY_PODS_INTERVAL=0` is a footgun — `sleep 0` returns immediately and `waited=$((waited + 0))` never advances, so the `waited >= deadline` exit never fires and the loop spins indefinitely if the scheduler doesn't catch up. Probably worth either clamping to a minimum of 1 (`interval=$(( interval < 1 ? 1 : interval ))`) or asserting `[[ "$interval" -gt 0 ]]` near the defaults on L1306 — same shape as your fail-closed treatment of kubectl errors.

kwok/scripts/validate-scheduling.sh
```diff
@@ -1277,57 +1277,159 @@ verify_pods() {
...
+        # Job-owned cleanup pods (Helm hooks may legitimately lack
+        # tolerations).
+        unscheduled_pods=$(jq --arg pat "$exclude_ns" '
+            [.items[]
+             | select(.metadata.namespace as $ns | $pat | split("|") | map(. == $ns) | any | not)
+             | select(.status.phase == "Pending")
+             | select((.spec.nodeName // "") == "")
+             | select((.metadata.ownerReferences // []) | map(.kind) | contains(["Job"]) | not)
+            ] | length
+        ' <<<"$snap")
```
mchmarny: nit (not blocking): five `jq` invocations on the same `$snap` per iteration. Worst case (13 iterations × 5 jq calls = 65 jq runs) is still fine cost-wise as you noted, but a single pass would be both faster and visually reinforce the "all counts from one snapshot" invariant. Sketch:

```bash
read -r total_pods running_pods pending_pods failed_pods unscheduled_pods < <(
    jq -r --arg pat "$exclude_ns" '
        [.items[] | select(.metadata.namespace as $ns | $pat | split("|") | map(. == $ns) | any | not)] as $f
        | "\($f | length) \
           \([$f[] | select(.status.phase == "Running")] | length) \
           \([$f[] | select(.status.phase == "Pending")] | length) \
           \([$f[] | select(.status.phase == "Failed")] | length) \
           \([$f[] | select(.status.phase == "Pending") | select((.spec.nodeName // "") == "") | select((.metadata.ownerReferences // []) | map(.kind) | contains(["Job"]) | not)] | length)"
    ' <<<"$snap"
)
```

Same `$pat` filter, factored once via `as $f`. Take it or leave it.

mchmarny (dismissed):
Solid fix. The single-snapshot pattern eliminates the impossible-math signature from #1090 by construction, and the bounded settle loop addresses the post-sync scheduler latency observed on `eks-inference (argocd-oci)` / `oke-training (argocd-oci)` without adding latency on the happy path. CI is green across all 22 KWOK Tier-1 lanes on this revision.

Two nits on the loop machinery (neither blocking), plus props on the comment density — the inline rationale is going to save the next person a lot of root-cause time.

kwok/scripts/validate-scheduling.sh
yuanchen8911: Fixed in bae72e26. Went with the assertion you suggested: both `KWOK_VERIFY_PODS_INTERVAL` and `KWOK_VERIFY_PODS_DEADLINE` are now validated right after the defaults and rejected with a fail-closed `log_error` + `return 1` (same shape as the kubectl-error handling) — interval must be a positive integer `>= 1`, deadline a non-negative integer. While in there I also canonicalize both to base 10 (`deadline=$((10#$deadline))` / `interval=$((10#$interval))`) so leading-zero values like `08`/`09` cant pass the digit regex and then get misread as octal by the loop arithmetic. Thanks for catching this.

kwok/scripts/validate-scheduling.sh
yuanchen8911: Done in bae72e26 — collapsed the five per-iteration `jq` calls into a single pass using your `as $f` factoring, emitting the five counts as one `@tsv` line consumed by `read -r total_pods running_pods pending_pods failed_pods unscheduled_pods`. Verified the single-pass output matches the old per-call results on a synthetic snapshot (total/running/pending/failed/unscheduled all equal). Appreciate the sketch.

mchmarny (approved):

Issue #610, opened 2026-04-17
Bridge Jobs for deploy/undeploy lifecycle flow control


lockwobr:
## Summary

Introduce a convention-based pattern for deploy-time readiness and undeploy-time CR cleanup: each component can ship `readiness.yaml` and/or `cleanup.yaml` alongside its existing `values.yaml` and `manifests/*.yaml`, rendered through the same template pipeline. AICR generates manifests with the right annotations for each supported deployer; deploy tools execute them via their own native primitives.

## Problem

Two related gaps currently live in generated bash:

**Deploy-time readiness.** `deploy.sh` prints `Deployment complete.` when installs finish, but for GPU bundles that's not "ready for workloads." Recent CUJ2 run: 7 min install → 26 min actually ready.

**Undeploy finalizer race.** `undeploy.sh.tmpl` has grown ~700 lines across 7 PRs (#253, #282, #364, #416, #477, #561, #602) chasing a race: today's pre-flight scans for stuck finalizers *after* helm has removed the controller — too late, nothing left to release them. Hence escalation to `force_clear_namespace_finalizers`.

A `pre-uninstall` bridge Job (or argocd sync-wave ordering — see below) runs while the controller is alive, so finalizers resolve normally and the race disappears.

Related: #602, #607, #516.

## Proposed Convention

Extend the existing `recipes/components/<name>/` layout:

| File | Purpose | Status |
|---|---|---|
| `values.yaml`, `values-*.yaml` | Helm values + overlays | existing |
| `manifests/*.yaml` | Raw manifests | existing |
| `readiness.yaml` | Post-install bridge Job | new |
| `cleanup.yaml` | Pre-uninstall bridge Job | new |

Bundler picks up the new files by presence, renders through the existing `pkg/manifest` template pipeline (full access to `.Values`, `.Release.Namespace`, overlay values), and annotates for each target deployer. Components that don't need a bridge simply don't ship the file.

## Example — `recipes/components/gpu-operator/readiness.yaml`

Waits for `ClusterPolicy.status.state == ready`. Reuses the upstream chart's `gpu-operator` SA — no new RBAC.

```yaml
{{- $gpuOp := index .Values "gpu-operator" }}
{{- if ne (toString ($gpuOp.readiness.enabled | default true)) "false" }}
---
apiVersion: batch/v1
kind: Job
metadata:
  name: gpu-operator-readiness
  namespace: {{ .Release.Namespace }}
  annotations:
[...]

PR #1095, opened 2026-05-28
feat(bundler): drop generated undeploy.sh; delegate to helm uninstall

### LINKED ISSUE #671: Move undeploy.sh to tools/ for dev use; direct users to deployer-native uninstall
## Summary

Remove `undeploy.sh` from the generated bundle and move its teardown capability to `tools/` as a scoped developer utility. Replace bundle-script teardown guidance in user-facing docs with deployer-native uninstall for the deployers AICR actually supports (`helm`, `argocd`, `argocd-helm`).

## Motivation / Context

The generated `undeploy.sh` has accumulated complexity across #253, #282, #364, #416, #477, #561, #602, #661 chasing CRD lifecycle, finalizer resolution, and multi-tenancy safety — problems Helm itself deliberately avoids and that purpose-built deployers handle more safely.

AICR's stated role is to generate validated design-time bundles, not to be the deployer. Shipping `undeploy.sh` makes AICR implicitly responsible for safe teardown (including the blast-radius of cluster-wide CRD deletion) without the machinery to do it correctly. Each iteration of hardening reinforces this scope creep.

Consensus from #610, #661 (closed), and #669 (closed) is to commit to this direction.

Related: #610, #602, #661, #669.

## Proposed Scope

**In scope:**

1. Remove `undeploy.sh.tmpl` from bundle generation in `pkg/bundler/deployer/helm/`, along with every in-tree reference to it:
   - `pkg/bundler/deployer/helm/helm.go` (generation wiring)
   - `pkg/bundler/deployer/helm/helm_test.go` (bundle-output assertions)
   - Any bundle checksum / manifest entries emitting `undeploy.sh`
   - Generated bundle `README.md` text that tells users to run `./undeploy.sh`
   - `deploy.sh.tmpl` error hints that direct users at `./undeploy.sh` for cleanup
2. Add a standalone developer utility under `tools/` (e.g. `tools/nuke-bundle`) that operates on a generated bundle directory and performs the teardown semantics current CI/KWOK/Kind runs depend on.
3. **CI/e2e sweep** — update `tools/e2e`, `tests/chainsaw/**`, KWOK workflows (`make kwok-test-all`), and any GitHub Actions referencing bundle `undeploy.sh` to use the new tool.
4. **Docs** — rewrite the `Undeploy Script Behavior` section in both the primary (`docs/user/cli-reference.md`) and mirrored (`site/docs/user/cli-reference.md`) copies to document that bundles no longer include `undeploy.sh` and to point users at the deployer-native uninstall path matching their generated bundle:
   - `helm`: `helm uninstall` + deployer-driven CRD lifecycle
   - `argocd`: app deletion with `resources-finalizer.argocd.argoproj.io`
   - `argocd-helm`: ArgoCD app deletion driving Helm uninstall under the hood
   - Flux / Helmfile / other deployers AICR does not generate for: external-only pointer, not required scope
5. **Release notes + migration guidance** — release-notes callout for the removal plus an explicit `./undeploy.sh` → `tools/<name>` (dev) / deployer-native command (users) migration table.

**Open questions (must be decided before implementation starts):**

1. **Deprecation vs. immediate removal** — gating decision. This removes a generated user-facing bundle artifact, so a one-release deprecation window (undeploy.sh still emitted but prints a removal notice) may be preferable to a hard removal. Implementation cannot proceed until the team picks one.
2. **Scope of `tools/<name>`** — full parity with current `undeploy.sh` (pre-flight, post-flight, orphan-CRD cleanup, PVC handling, webhooks, namespaces), or only what CI/KWOK/Kind workflows actually exercise?
3. **Safety contract for `tools/<name>`:**
   - Source-tree-only — never copied into a generated bundle, so it can only run against a local checkout?
   - Preserves the current shared-cluster safety defaults (Helm release-name + release-namespace annotation guard, foreign-release CRD skip)?
   - Any destructive CRD cleanup guarded behind an explicit opt-in flag (same posture as PR #661 proposed)?
4. **Depth of "deployer-native uninstall" docs** — per-deployer walkthroughs or upstream-doc pointers?

## Non-goals

- Re-implementing CRD lifecycle management outside the script (that's deployer territory now).
- Any change to `deploy.sh` — out of scope for this issue.

## Acceptance Criteria

- [ ] Bundle generation no longer emits `undeploy.sh`
- [ ] No generated bundle file, `README.md`, checksum entry, or `deploy.sh` hint references `undeploy.sh`
- [ ] `tools/<name>` provides equivalent teardown for developer workflows, with its safety contract documented in-source
- [ ] All CI / e2e / KWOK paths previously invoking bundle undeploy use the new tool
- [ ] Both `docs/user/cli-reference.md` and `site/docs/user/cli-reference.md` no longer document bundle undeploy; each points at the deployer-native uninstall path matching the bundle's deployer
- [ ] Release notes call out the change and include a `./undeploy.sh` → `tools/<name>` / deployer-native migration table

mchmarny:
## Summary

Removes the generated `undeploy.sh` from helm bundles and delegates teardown to the deployer-native uninstall path (`helm uninstall`, ArgoCD app deletion). Rewrites `docs/user/cli-reference.md` `Undeploy Script Behavior` as a `Bundle Uninstall` section with per-deployer walkthroughs.

## Motivation / Context

Per the issue, `undeploy.sh` accumulated complexity across #253, #282, #364, #416, #477, #561, #602, #661 chasing CRD lifecycle, finalizer resolution, and multi-tenancy safety — problems Helm itself deliberately avoids and that purpose-built deployers handle more safely. AICR's role is to generate validated design-time bundles, not to be the deployer.

The implementation matches the gating decisions made during planning:
- **Immediate removal** (no one-release deprecation window)
- **No `tools/` utility** — verified that nothing in CI / Makefile / chainsaw / KWOK / `tools/e2e` actually invokes `undeploy.sh`, so the "minimal scope" choice maps to zero migration target
- **Per-deployer walkthroughs** in the docs (helm / argocd / argocd-helm)

Fixes: #671
Related: #610, #602, #661, #669

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

**Deviation from issue AC #3 (tools/ utility):** The issue's acceptance criteria say "tools/ provides equivalent teardown for developer workflows," but a repo-wide sweep found that nothing in CI (`.github/workflows/`), `Makefile`, `tools/` scripts, chainsaw tests, KWOK workflows, or developer docs (`DEVELOPMENT.md`, `CONTRIBUTING.md`) actually invokes `undeploy.sh`. The only existing chainsaw reference is a negative assertion that helmfile bundles don't leak `undeploy.sh`. With the "Only what CI/KWOK/Kind exercise" scope choice, the resulting tool would be empty. Skipping the `tools/` utility entirely — surfacing this discrepancy here.

**Chainsaw assertion:** Dropped the `undeploy.sh leaked into helmfile bundle` check from `tests/chainsaw/cli/bundle-helmfile/chainsaw-test.yaml`. After this change `undeploy.sh` is no longer generated by any deployer, so the negative assertion is universally true and adds noise.

**Golden testdata regenerated** to reflect the new bundle shape — no more `undeploy.sh` golden files, updated `README.md` Uninstall section, updated `deploy.sh` pre-flight error hint.

**Helper function cleanup:** `uniqueNamespaces` (only used by undeploy generation) removed. `reverseComponents` retained — still used by README.md's per-component manual-uninstall block.

## Testing

```bash
make qualify
```

- `go test -race ./pkg/bundler/...` — all packages pass
- `golangci-lint run -c .golangci.yaml ./...` — 0 issues
- `yamllint` — clean on the modified chainsaw test
- Bundle golden tests regenerated and reverified

**Coverage delta (`pkg/bundler/deployer/helm`):** 84.2% → 83.7% (-0.5%, at the threshold). The drop comes from removing the 14 `TestUndeployScript_*` shell-behavior tests that exercised the deleted template. Absolute level still well above the 70% project floor.

**Pre-existing `pkg/trust` failures** in the local run are network-related (sandbox can't reach `tuf-repo-cdn.sigstore.dev`) and reproduce on `origin/main`; unrelated to this PR.

## Risk Assessment

- [x] **High** — Breaking change, affects critical paths, or complex rollout

**Rollout notes:** This is a breaking change for any user who scripted against `./undeploy.sh` in a generated helm bundle. Migration:

| Before                          | After (helm)                                       |
| ------------------------------- | -------------------------------------------------- |
| `./undeploy.sh`                 | `helm uninstall <release> -n <namespace>`          |
| `./undeploy.sh --delete-pvcs`   | `kubectl -n <ns> delete pvc --all`                 |
| `./undeploy.sh --keep-namespaces` | (default; Helm doesn't delete namespaces)        |

| Before          | After (argocd / argocd-helm)                       |
| --------------- | -------------------------------------------------- |
| `./undeploy.sh` | `kubectl -n argocd delete application <parent>`    |

Full per-deployer walkthrough (CRD handling, PVC handling, stuck-release recovery) is in `docs/user/cli-reference.md#bundle-uninstall`. The bundle's `README.md` now embeds the rendered `helm uninstall` commands per component and links to the doc section.

## Checklist

- [x] I added/updated tests for new functionality (N/A — net removal; golden tests verify the absence)


pkg/bundler/deployer/helm/templates/README.md.tmpl
```diff
@@ -96,20 +96,21 @@ bash install.sh
...
 ## Uninstall
 
-To remove components (reverse order):
-
-```bash
-./undeploy.sh
-```
-
-Or remove a single release manually:
+Bundles do not ship an `undeploy.sh`. Uninstall components in reverse
```
yuanchen8911: **Issue 1 — Bundle README uninstall instructions are incomplete**

`pkg/bundler/deployer/helm/templates/README.md.tmpl:99–105`

The `## Uninstall` block ranges only over `.ComponentsReversed` (recipe components from `buildComponentDataList`), but the install loop at `pkg/bundler/deployer/helm/templates/deploy.sh.tmpl:291` (`for dir in "${SCRIPT_DIR}"/[0-9][0-9][0-9]-*/`) installs every numbered `NNN-*` folder as a separate Helm release. Bundles with injected `*-pre`/`*-post` folders therefore get incomplete teardown guidance. Verified: `mixed_with_pre` installs release `foo-pre` (`001-foo-pre/install.sh:29`) yet its regenerated README (`testdata/mixed_with_pre/README.md:87`) lists only `helm uninstall foo`. The deleted `undeploy.sh` previously globbed folders, so this is a real regression.

**Recommendation:** Generate the uninstall list from the actual emitted releases/folders (the same `[0-9][0-9][0-9]-*` set the deploy template globs), not from recipe components; regenerate the 6 goldens.

pkg/bundler/deployer/helm/helm.go
```diff
@@ -190,15 +189,6 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 	output.Files = append(output.Files, deployPath)
```
yuanchen8911, line 200: **Issue 2 — Stale `undeploy.sh` survives reused output dirs**

`pkg/bundler/deployer/helm/helm.go:183–194`

The generator no longer writes `undeploy.sh`, and nothing removes an existing one. `localformat.Write` prunes only `NNN-*` folders and explicitly leaves top-level orchestration files intact (`writer.go:189–190`). Regenerating over a pre-PR bundle leaves a stale teardown script that is executable, misleading, contradicts the new README, and is excluded from `checksums.txt`.

**Recommendation:** Remove the stale top-level `undeploy.sh` during Helm bundle generation. (Documenting the caveat is insufficient — the file remains executable, misleading, and unchecksummed.)

docs/user/cli-reference.md
```diff
@@ -1930,36 +1929,79 @@ aws ec2 reboot-instances --instance-ids <instance-id>
...
-**Stuck release handling:**
+See [Helm 3 uninstall docs](https://helm.sh/docs/helm/helm_uninstall/) for
+the full flag reference.
 
-If a Helm release is in a `pending-install` or `pending-upgrade` state (from an interrupted deploy), the script retries with `--no-hooks` to force removal.
+##### argocd
 
-**Orphaned webhook cleanup:**
+Delete the parent `Application` that owns the bundle's child Applications
+(app-of-apps). The `resources-finalizer.argocd.argoproj.io` finalizer that
```
yuanchen8911: **Issue 3 — ArgoCD uninstall docs claim a finalizer AICR does not set**

`docs/user/cli-reference.md:1979–1981`

The new `##### argocd` walkthrough states AICR sets `resources-finalizer.argocd.argoproj.io` on each Application, making `kubectl delete application` cascade. The argocd templates (`application.yaml.tmpl`, `app-of-apps.yaml.tmpl`) set no finalizer — verified zero matches. Without it the delete is non-cascading and orphans the deployed stack, defeating the designated replacement teardown path.

**Recommendation:** Either change the doc to `argocd app delete <app> --cascade` / manual finalizer addition, or add `resources-finalizer.argocd.argoproj.io` to the generated Applications. If choosing the finalizer path, cover both the parent (app-of-apps) and child Application resources as appropriate for cascade behavior, then regenerate the ArgoCD testdata.

docs/user/cli-reference.md
```diff
@@ -1930,36 +1929,79 @@ aws ec2 reboot-instances --instance-ids <instance-id>
...
+
+The CRD and PVC notes above still apply — Argo follows Helm's
+non-destructive-by-default semantics for both.
+
+See [ArgoCD app deletion docs](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/)
+for finalizer behavior and selective deletion.
+
+##### argocd-helm
+
+The ArgoCD application drives `helm uninstall` under the hood, so deletion
```
yuanchen8911: One doc wording issue to fix alongside the ArgoCD finalizer path: the `argocd-helm` section says the ArgoCD application drives `helm uninstall` under the hood. That is misleading. Argo CD uses Helm only to render manifests with `helm template`; Argo CD then manages/prunes those rendered Kubernetes resources directly.

Suggested wording:

Argo CD renders Helm charts into Kubernetes manifests and manages those resources directly. Deleting the Application with cascade enabled prunes the resources Argo CD tracks; it does not run `helm uninstall`.

Refs:
- https://argo-cd.readthedocs.io/en/latest/user-guide/helm/
- https://argo-cd.readthedocs.io/en/latest/faq/#after-deploying-my-helm-application-with-argo-cd-i-cannot-see-it-with-helm-ls-and-other-helm-commands

yuanchen8911 (changes requested):

docs/user/cli-reference.md
```diff
@@ -1930,36 +1929,79 @@ aws ec2 reboot-instances --instance-ids <instance-id>
...
+follows the same path as plain ArgoCD:
+
+```bash
+kubectl -n argocd delete application <bundle-parent-app>
+```
 
-After uninstalling each component, the script checks for orphaned validating/mutating webhooks whose backing service no longer exists. Fail-closed webhooks with missing services block all pod creation, so these are deleted proactively.
+The cascade is Helm-driven, so the same CRD/PVC preservation behavior
+applies. Manual cleanup of CRDs and PVCs follows the **helm** walkthrough
+above.
```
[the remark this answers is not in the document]
mchmarny: Fixed in 125f49b8.

- `argocd-helm` wording corrected: section now states Argo CD uses `helm template` to render manifests, then prunes the rendered resources directly via finalizers — it does **not** run `helm uninstall`, and `helm ls` won't see the releases. Links to the [Argo CD Helm user-guide](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/) and the [`helm ls` FAQ entry](https://argo-cd.readthedocs.io/en/stable/faq/#after-deploying-my-helm-application-with-argo-cd-i-cannot-see-it-with-helm-ls-and-other-helm-commands) for context.
- Added new `##### flux` subsection recommending `kubectl delete helmrelease <release>` so `helm-controller` runs the uninstall via `spec.uninstall`, then cleanup of source objects. Links to the [Flux helm-controller uninstall reference](https://fluxcd.io/flux/components/helm/helmreleases/#uninstall-configuration).
- Added a brief `##### helmfile` subsection pointing at `helmfile destroy` as the canonical primitive.

The flux walkthrough targets `HelmRelease` resources directly — happy to refine the wording if the AICR flux bundle wraps them in a different root than I assumed.

pkg/bundler/deployer/helm/templates/README.md.tmpl
mchmarny: Fixed in 125f49b8. Good catch — verified the regression on `testdata/mixed_with_pre` (was listing only `foo`, now lists `foo` then `foo-pre` in reverse install order).

Approach:
- Added a `Namespace` field to `localformat.Folder` (populated at the 3 construction sites in `local_helm.go`, `upstream_helm.go`, `vendor_folder.go`) so the caller doesn't have to re-derive namespace per release.
- Threaded `writeResult.Folders` from `localformat.Write` into `generateRootREADME` and exposed it on `readmeTemplateData` as `ReleasesReversed []releaseRef` (built via a new `reverseReleases` helper, tested in `TestReverseReleases`).
- `README.md.tmpl`'s `## Uninstall` block now ranges over `.ReleasesReversed` — one `helm uninstall` per emitted `NNN-*` folder (`*-pre` + primary + `*-post`).
- Regenerated all 6 golden bundles. Verified `mixed_gpu_operator` now lists `gpu-operator-post`, `gpu-operator`, `gpu-operator-pre` in reverse install order.

The now-dead `reverseComponents` helper + its test were replaced with `reverseReleases` / `TestReverseReleases`.

pkg/bundler/deployer/helm/helm.go
mchmarny: Fixed in 125f49b8.

`Generate()` now removes any pre-existing top-level `undeploy.sh` up front (right after `MkdirAll`), before `localformat.Write` runs. ENOENT is swallowed (no stale file to remove); any other error (e.g., `EACCES` on a read-only parent dir) is wrapped with `ErrCodeInternal` and propagated so we fail closed rather than silently overwrite the rest of the bundle.

Coverage:
- `TestGenerate_RemovesStaleUndeployScript` seeds a stale `undeploy.sh`, runs `Generate`, asserts removal.
- `TestGenerate_StaleUndeployRemovalErrorPropagates` chmod-locks the parent dir to 0o500 and asserts `Generate` returns an error mentioning the stale-removal failure. (Skips when EUID is 0 — root bypasses the permission check.)

Net per-package coverage on `pkg/bundler/deployer/helm`: 84.2% (main) → 84.0% (-0.2%).

docs/user/cli-reference.md
mchmarny: Fixed in 125f49b8. Confirmed by grep that the argocd templates set no finalizer — the doc was wrong, not the templates.

Rewrote the `##### argocd` walkthrough to:
- State explicitly that AICR does **not** set `resources-finalizer.argocd.argoproj.io` on generated Applications.
- Recommend `argocd app delete <app> --cascade=foreground` as the primary teardown path (the CLI handles the cascade for you).
- Document the kubectl + manual finalizer-patch fallback for users who can't use the CLI:
  ```bash
  kubectl -n argocd patch application <bundle-parent-app> --type=merge \
    -p '{"metadata":{"finalizers":["resources-finalizer.argocd.argoproj.io"]}}'
  kubectl -n argocd delete application <bundle-parent-app>
  ```
- Clarify that Argo CD does not run `helm uninstall` for Helm-templated children (related to your thread on argocd-helm wording), so CRD/PVC cleanup from the **helm** walkthrough still applies.

I went with the doc-fix path rather than adding the finalizer to generated Applications, since the latter would expand scope beyond this issue's "remove undeploy.sh, redirect users to deployer-native uninstall" mandate. Happy to file a follow-up issue if the team would prefer the templates set the finalizer by default.

docs/user/cli-reference.md
mchmarny: Fixed in 125f49b8. The `##### argocd-helm` section now reads:

> Same path as plain `argocd`: Argo CD uses Helm only to render charts into Kubernetes manifests (via `helm template`) and then manages those resources itself. Deleting the Application with cascade enabled prunes the resources Argo CD tracks; it does **not** run `helm uninstall`, and `helm ls` will not show the bundle's releases.

Links to the [Argo CD Helm user-guide](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/) and the [`helm ls` FAQ entry](https://argo-cd.readthedocs.io/en/stable/faq/#after-deploying-my-helm-application-with-argo-cd-i-cannot-see-it-with-helm-ls-and-other-helm-commands) you cited. CRD/PVC cleanup follows the **helm** walkthrough notes; the kubectl + finalizer-patch fallback from **argocd** applies here too.

docs/user/cli-reference.md
```diff
@@ -1930,36 +1929,133 @@ aws ec2 reboot-instances --instance-ids <instance-id>
...
+`resources-finalizer.argocd.argoproj.io` finalizer on generated
+Applications, so a plain `kubectl delete` removes only the Application CR
+and leaves the managed resources running. Use one of the cascade-aware
+flows instead:
 
-When multiple components share a namespace (e.g., `monitoring` contains `kube-prometheus-stack`, `prometheus-adapter`, and `k8s-ephemeral-storage-metrics`), all components are uninstalled first, then PVC and namespace cleanup runs once. This prevents hangs caused by `kubernetes.io/pvc-protection` finalizers — if a StatefulSet owner is still running when PVC deletion is attempted, the delete blocks indefinitely.
+```bash
+# Argo CD CLI — cascade is the default; foreground waits for resources
+argocd app delete <bundle-parent-app> --cascade=foreground
+```
```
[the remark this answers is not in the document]
mchmarny: Both points fixed in 4db8d82b:

1. **Cascade flag**: replaced `--cascade=foreground` with `--cascade --propagation-policy foreground` at both occurrences (lines 1987 and 2017). Matches the official [`argocd app delete` reference](https://argo-cd.readthedocs.io/en/stable/user-guide/commands/argocd_app_delete/).

2. **MD031 (outside-diff comment)**: tightened the `## Uninstall` range body in `README.md.tmpl` so multi-release bundles get exactly one blank line between fenced blocks (was emitting back-to-back fenced blocks for `mixed_with_pre` / `mixed_gpu_operator`). Also removed the trailing triple-blank-line that the previous form left before the "CRDs installed by these charts…" paragraph. Regenerated all 6 golden bundles to lock in the new spacing.

lalitadithya (approved):

Issue #1076, opened 2026-05-28
[Epic]: Post-facade adoption and surface hardening (follow-ups to #1072)


mchmarny:
## Summary

Tracking epic for follow-up work introduced by #1072 (the top-level `github.com/NVIDIA/aicr` facade). Each child is independently reviewable and mergeable.

**Status: ✅ closed — all four children merged.**

## Motivation

PR #1072 ships the `aicr.Client` facade defined by #1071 and lands the per-Builder DataProvider isolation that #986 was scoped to provide. The facade is functional, but to actually deliver the architectural payoff and honor its documented semver contract, several deferred items need to land.

Without these:
- The codebase carries three independent wiring paths for `DataProvider → Builder → Bundler → Validator` (CLI, REST, facade).
- The facade's "Public (stable)" contract identity-aliases types from packages marked Internal / "Public (evolving)" in `docs/integrator/public-api.md`, undermining the semver promise.
- The cross-Client safety guard #1071 motivates only protects facade callers; direct `pkg/recipe.Builder` users (today: `pkg/api/server.go`) remain exposed.
- Test-only helpers leaked onto `pkg/recipe`'s public surface.

## Final status

**Closed: 2026-05-29**

### Children

| # | Issue | Title | PR(s) | Status |
|---|---|---|---|---|
| 1 | #1079 | `refactor(recipe): move _ForTesting cache helpers off the public surface` | — | ✅ merged |
| 2 | #1077 | `feat(cli,api): migrate CLI and REST server to consume aicr.Client` | #1107 + #1108 | ✅ merged |
| 3 | #1078 | `feat(aicr): wrap facade alias types as facade-owned structs` | #1111 | ✅ merged |
| 4 | #1080 | `feat(recipe): push owner-token guard down to pkg/recipe.RecipeResult` | #1113 | ✅ merged |

### Adjacent work landed alongside the epic

These weren't original children but ship in the same area and were folded into the rollout:

- **#1107** (PR 1 of 2 for #1077) — per-DataProvider sources + criteria-registry de-globalization (foundation for #1077)
- **#1112** — move public facade from module root to `pkg/aicr/` (cleaner project layout)
- **#1111** — also moved the facade from `pkg/aicr/` to `pkg/client/v1/` (versioned SDK location, matches `pkg/api/validator/v1` pattern) in the same PR as the alias wrapping
[...]

Issue #1080, opened 2026-05-28
feat(recipe): push owner-token guard down to pkg/recipe.RecipeResult


mchmarny:
## Summary

The cross-Client guard in `aicr.go:252` (`assertOwns`) prevents a `RecipeResult` produced by `Client A` from being passed to `Client B.BundleComponents` / `ValidateState`. The PR (#1072) calls out exactly why this matters:

> Without this guard, a result from Client A passed to Client B silently mixed A's component refs with B's DataProvider reads — wrong Helm values or supplemental manifests with no error.

This is the bug class that #1071 is filed against. But the guard lives on the facade, not on `pkg/recipe`. Direct `pkg/recipe.Builder` consumers — today: `pkg/api/server.go:85-91`, and any external project that imports `pkg/recipe` per the "Public (evolving)" tier — get no protection.

## Motivation

Two reasons to move the guard down:

1. **Defense at the layer where the bug lives.** The cross-source bleed is a property of `recipe.RecipeResult` carrying refs that only make sense relative to the `DataProvider` that produced them. The owner-token belongs on the result type, enforced wherever results are consumed.

2. **Facade-symmetric protection.** After #1076 child 1 (CLI/REST adopting the facade), this matters less in-tree, but the `pkg/recipe` package is publicly importable. Any external consumer that constructs two Builders today (even though the package was unsafe for that until this PR landed) hits the same bug class once they start trusting the per-Builder isolation.

## Acceptance

- [ ] `recipe.RecipeResult` carries an unexported `owner` field stamped by the producing `Builder`.
- [ ] `recipe.Builder.BuildFromCriteria` (and other producers) stamp the owner.
- [ ] Consumers in `pkg/recipe` (e.g., `RecipeResult.GetValuesForComponent`) — and the bundler's `BundleComponents` call sites — assert ownership, returning `ErrCodeInvalidRequest` on mismatch.
- [ ] `aicr.assertOwns` in `aicr.go:252` can either be removed (the inner check fires) or kept as a clearer error message at the facade boundary.
- [ ] Test: a `pkg/recipe`-level test analogous to `TestClient_CrossClientBundleRejected` exercises the bundler-level rejection without going through the facade.

## Same applies to `Close()` lifecycle

[...]

Issue #1078, opened 2026-05-28
feat(aicr): wrap facade alias types as facade-owned structs


mchmarny:
## Summary

Replace the transparent type aliases in `types.go:32-59` (`Snapshot`, `AgentConfig`, `PhaseResult`, `Phase`, and the phase constants) with facade-owned structs and translate at the boundary — same pattern `ValidateOption` already uses in `options.go:24`.

## Motivation

`docs/integrator/public-api.md` advertises `github.com/NVIDIA/aicr` as **Public (stable)** and marks `pkg/snapshotter` as **Internal** ("may change without notice"). But `types.go:32` is `type Snapshot = snapshotter.Snapshot` — a transparent type alias. Consumers using `aicr.Snapshot` are identity-pinned to `snapshotter.Snapshot`; any field rename/removal in the "Internal" package breaks them.

Same shape for `AgentConfig` (`types.go:40`, ← `snapshotter`), `PhaseResult` (line 46), `Phase` (line 51), and the phase constants (lines 55-59). `pkg/validator` is "Public (evolving)" not "Internal" but it still can't supply identity for a "stable" surface — "evolving" by definition admits minor-bump breakage.

The PR description (#1072) acknowledges this: "promoting them to facade-owned wrappers is worth scoping separately."

## Acceptance

- [ ] `aicr.Snapshot`, `aicr.AgentConfig`, `aicr.PhaseResult`, `aicr.Phase`, and the phase constants are facade-owned types with no transparent alias to internal packages.
- [ ] Translation between facade types and `pkg/snapshotter` / `pkg/validator` types is done inside the facade (analogous to `applyValidateOptions` in `options.go`).
- [ ] `docs/integrator/public-api.md` is consistent: every type reachable from the **Public (stable)** facade is either facade-owned or comes from a package the matrix lists as **Public (stable)**.
- [ ] No regression in facade tests; CLI/REST behavior unchanged.

## Notes on translation cost

Most of these types carry transitively-referenced types of their own (e.g., `validator.PhaseResult` embeds a CTRF report; `snapshotter.Snapshot` embeds a `measurement.Measurement` slice). The fix has to either:

1. Wrap the immediate type and re-expose only the fields the facade guarantees (lossy but stable), or
2. Wrap recursively (high effort, mirrors more types).

Pick (1) unless a consumer concretely needs a field that (1) drops.

## Related

- #1076 (epic)
- #1072 (the PR that introduced the aliases)

Issue #987, opened 2026-05-20
feat(recipe): deprecate SetDataProvider / GetDataProvider globals


mchmarny:
Part of #983. Depends on #986.

## Summary

Once per-Builder `DataProvider` isolation lands (#986), the package-global `recipe.SetDataProvider` and `recipe.GetDataProvider` are the only remaining process-global state in the recipe path. Mark them as deprecated and point new callers to `WithDataProvider`.

This is Stage 1 (godoc deprecation only). Removing the symbols is a follow-up that must wait for CLI / API server migration.

## Motivation

The point of #986 is that recipe resolution no longer needs a global provider — every consumer can hold its own. Leaving `SetDataProvider` as a documented entry point invites new code to reproduce the multi-tenant hazard the rest of the cleanup eliminated. Deprecating the symbol communicates the intended direction without breaking any existing caller.

The CLI's `initDataProvider` at `pkg/cli/root.go` and any other call site that still uses `SetDataProvider` continue to work unchanged through Stage 1. They get migrated in Stage 2.

### 1. Godoc deprecation markers

Add the standard Go deprecation marker to both functions:

```go
// SetDataProvider sets the global data provider.
//
// Deprecated: prefer recipe.NewBuilder(recipe.WithDataProvider(dp))
// to bind a provider to a specific Builder instance. The package-
// global provider is retained for back-compat with the CLI and API
// server but will be removed in a future release. See #983 for the
// migration plan.
func SetDataProvider(provider DataProvider) { ... }

// GetDataProvider returns the global data provider.
//
// Deprecated: callers that need a specific provider should hold their
// own reference (typically obtained from NewLayeredDataProvider or
// NewEmbeddedDataProvider) and pass it via WithDataProvider rather
// than relying on package state. See #983.
func GetDataProvider() DataProvider { ... }
```

The Go vet / staticcheck deprecation lint (`SA1019`) will surface these as warnings to anyone importing the symbols, including the AICR codebase itself — that's intentional, because it gives us a TODO list for Stage 2 migration.

### 2. Optional: lint-allowlist the existing in-tree call sites

If `staticcheck`'s `SA1019` is enabled in `.golangci.yaml` (or becomes enabled), the existing in-tree uses in `pkg/cli/root.go` and any test helpers will start failing CI. Either:

- Add `//nolint:staticcheck // tracked by #983 Stage 2` comments at the current call sites (preferred — keeps the TODO list visible), or
[...]

PR #1104, opened 2026-05-29
feat(cli,api): migrate CLI and REST server to consume aicr.Client

### LINKED ISSUE #1077: feat(cli,api): migrate CLI and REST server to consume aicr.Client
## Summary

Migrate `pkg/cli` and `pkg/api/server.go` from constructing `recipe.Builder` / `bundler.Bundler` / `validator.Validator` directly to consuming the `github.com/NVIDIA/aicr` facade landed in #1072.

## Motivation

After #1072 lands the facade, the repository carries three independent wiring paths:

| Caller | Mechanism | Source |
|---|---|---|
| CLI | `recipe.SetDataProvider` (process-global, deprecated) | `pkg/cli/root.go:382, 396` (`//nolint:staticcheck // tracked by #983 Stage 2`) |
| REST server | `recipe.NewBuilder` + `bundler.New` directly | `pkg/api/server.go:85-91` |
| Facade | `recipe.WithDataProvider(dp)` + per-Client isolation | `aicr.go:180` |

Every change to wiring policy (new constructor option, new pre-flight, new lifecycle hook) has to be applied in all three places. Per `CLAUDE.md` — "business logic lives in functional packages so CLI and API can both use it" — the facade should be the single aggregation point; CLI and REST become thin presentation layers over `aicr.Client`.

CLI is also still on the deprecated global-state path; migrating onto `aicr.Client` retires it.

## Acceptance

- [ ] `pkg/cli/root.go` no longer calls `recipe.SetDataProvider`; bootstraps an `aicr.Client` per command invocation.
- [ ] `pkg/api/server.go` constructs `aicr.Client` instances (one per recipe-source configuration), not `recipe.Builder` / `bundler.Bundler` directly.
- [ ] All existing CLI flags and REST endpoints behave identically (chainsaw e2e + `make qualify` clean).
- [ ] If the facade lacks a knob the CLI/REST need, extend the facade rather than reaching past it; document any such extensions.

## Out of scope

- Removing `pkg/recipe`'s `SetDataProvider` / `GetDataProvider` — that's #987.
- Wrapping the facade's alias types — separate child of #1076.

## Related

- #1076 (epic)
- #1072 (the facade PR)
- #987 (sibling: deprecates the globals; this issue removes the last in-tree caller)

hkii:
## Summary

Route the CLI and API server entirely through the public `aicr.Client` facade, and replace the process-global `DataProvider`/criteria registry with per-`Client` state. This makes the facade the single integration surface for both internal consumers and external Go importers, and lets multiple `Client`s resolve recipes concurrently without clobbering shared globals.

## Motivation / Context

AICR maintained three independent component-initialization paths: the **CLI** used the deprecated `recipe.SetDataProvider` process-global; the **REST server** directly constructed `recipe.Builder`, `bundler.Bundler`, and `validator.Validator`; and the **facade** used the new per-instance `aicr.Client`. Every change to wiring policy had to be applied in all three places. Per the project principle that "business logic lives in functional packages so CLI and API can both use it," the facade is the proper single aggregation point.

This PR migrates the CLI and REST server to consume `aicr.Client` as thin presentation layers, retires the legacy global-state and direct-construction wiring, and extends the facade (rather than bypassing it) where new capabilities were needed — preserving behavioral parity across all CLI flags and REST endpoints.

Fixes: #1077
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: public facade (`aicr.go`, `bundle.go`, `query.go`, `options.go`, `types.go`)

## Implementation Notes

**Facade surface (new):**
- `Client.LoadRecipe` (provider-bound file load), `MakeBundle` (full deployer-mode bundle), `ResolveRecipeFromSnapshot` (constraint eval vs a snapshot), `ResolveRecipeFromCriteria`, `SelectFromRecipe` (hydrate + select), `LoadCatalog`, `Resolved()`, plus `CriteriaRegistry` accessor.
- New options: `WithVersion`, `WithAllowLists` + `ParseAllowListsFromEnv` (with enforcement), `EmbeddedSource` for embedded-only data, and validation options `WithValidationPhases/Timeout/Commit/ImageRegistryOverride/ImageTagOverride`.
- Lossless type aliases (`Recipe`/`AllowLists`/`Criteria`).

**Per-`Client` isolation (replaces process global):**
- `pkg/recipe` gains a per-`DataProvider` criteria registry with a global shim for back-compat; `LoadFromFileWithProvider` and `catalog.Load` now take a `DataProvider` instead of reading a hidden global. `pkg/config` adds `ResolveCriteriaWithRegistry` for per-provider validation.
- CLI `initDataProvider` is dropped; `recipe`/`query`/`mirror`/`bundle`/`validate` and API `/v1/recipe` `/v1/query` `/v1/bundle` all consume `aicr.Client`. `bundler.New` is removed from `server.go`.

**Correctness fixes folded in:**
- Deep-copy `ComponentRef.Overrides` nested values in clone, and deep-copy the recipe in `AdoptRecipe`, to preserve `Client` isolation (no slice/map aliasing across overlay merges).
- `validate` preserves structured error codes via `PropagateOrWrap`.
- Validation timeout is now **opt-in** (`*time.Duration`): the CLI passes `0` to uncap so an all-phase run isn't cut short by a fixed cap; controllers keep the default. Tolerations use an explicit `tolerationsSet` flag so a nil/empty override correctly clears the validator's default tolerate-all instead of being dropped.

## Testing

```bash
make qualify
```

New/expanded coverage: `aicr_test.go`, `aicr_internal_test.go`, `bundle_test.go`, `pkg/api/{recipe,bundle}_handler_test.go`, `pkg/recipe/{criteria_registry,loader_provider}_test.go`, `pkg/validator/catalog/catalog_test.go`.

## Risk Assessment

- [ ] **Low**
- [x] **Medium** — Touches CLI, API, recipe engine, and validator, and changes how the `DataProvider`/criteria registry is wired. Behavior is preserved via a global shim and defensive copies; no external API contract changes.
- [ ] **High**

**Rollout notes:** No migration required for external importers — the `aicr.Client` surface is additive and the prior process-global path retains a back-compat shim. Concurrent-`Client` isolation is a v0.12+ guarantee (documented in `aicr.go`); long-running consumers should retain and reuse `Client`s and call `Close()` to avoid monotonic cache growth.

## Checklist

- [x] I updated docs if user-facing behavior changed (`docs/integrator/go-library.md`, `docs/integrator/public-api.md`)
- [ ] Commits are cryptographically signed (`git commit -S`)


mchmarny (commented):
Thanks for taking this on, @hkii — you've been carrying the facade thread since #1072 and this PR is the logical next step. Direction is right and matches the project principle that the facade should be the single aggregation point. A few architectural shape questions before line-level review starts; flagging them now since the PR is still draft.

**1. Two parallel public-API tiers on the Client.**
`types.go` introduces transparent aliases (`type Recipe = recipe.RecipeResult`) alongside the existing facade-owned shape (`type RecipeResult struct {…}`), and the surface has parallel resolve methods returning each (`ResolveRecipe → *RecipeResult` vs `ResolveRecipeFromCriteria/FromSnapshot → *Recipe`). The Stability note on `Snapshot` already concedes the alias tier "inherits pkg/snapshotter's Public (evolving) tier rather than the facade's stable tier — field-shape changes propagate without notice," and a comment points to #1078 as the future wrap. The risk: this PR ships the alias surface that #1078 will need to wrap, and any caller built against `*Recipe` between now and #1078 has to migrate twice. Suggest landing **#1078 first** so the facade-owned wrapper protects external Go importers; then this PR consumes the stable types.

**2. The PR bundles a wiring refactor with several behavior changes.**
Title says "migrate CLI and API to consume aicr.Client" but the implementation notes also include: deep-copy fix in `AdoptRecipe` + `ComponentRef.Overrides` (fixes a prior aliasing bug — i.e., behavior change), validation timeout `time.Duration → *time.Duration` with the CLI passing `0` to **uncap** (real behavior change for CLI users), tolerations explicit-`set` flag (semantic change), and `validate` propagating structured codes via `PropagateOrWrap`. Each of those is independently bisectable and reviewable. Bundling them with a 4,900-line refactor across 41 files in 38 commits makes regression hunting much harder if something breaks downstream. The "preserves behavioral parity" framing also fights the substance — uncapping validation timeout in the CLI is a deliberate change worth calling out plainly.

Suggested split:
- Pre-PR: each correctness fix on its own commit with a regression test (`AdoptRecipe` deep-copy is a particularly good standalone candidate — small, testable, ships behavior improvement without coupling to wiring).
- Pre-PR: per-`DataProvider` criteria registry (replaces the process global) with the back-compat shim's deprecation marker and a tracking comment pointing at #987 for removal.
- This PR (slimmed): only the CLI/API migration onto the now-stable Client surface. Should drop 1,500–2,000 lines once the above ship separately.

**3. Back-compat shim sunset plan.**
"Per-`DataProvider` criteria registry with a global shim for back-compat" — shims tend to be sticky, and #987 is the sibling that removes the global entirely. Worth confirming the shim is flagged with `//nolint:staticcheck` + a `Deprecated:` doc tag pointing at #987, and that there's a test asserting no in-tree caller still touches the global. Otherwise we end up with a permanent dual code path.

**4. PR mechanics.** None of this is substance; flagging for the squash-and-rebase pass:
- Squash: 38 commits → 1 per logical change (after splitting per concern #2 above)
- Rebase: `mergeable_state: dirty` after #1100/#1101/#1102 landed
- Sign: `git commit -S` (checkbox is unchecked)
- `make qualify` checkbox: unchecked — confirm pass locally before flipping to ready-for-review

**What's clearly good and worth keeping:** the direction (single integration surface, per-instance state), the docs updates landing in the same PR (`go-library.md`, `public-api.md`), the per-file test coverage discipline, and the `Close()` + per-Client guarantee documented in rollout notes. The architectural boundaries are respected — no kubectl orchestration, no in-cluster controllers, no new deployers.

Happy to pair on any of the splits, especially the #1078 sequencing question. Thanks again for the work.

hkii:
> Thanks for taking this on, @hkii — you've been carrying the facade thread since #1072 and this PR is the logical next step. Direction is right and matches the project principle that the facade should be the single aggregation point. A few architectural shape questions before line-level review starts; flagging them now since the PR is still draft.
> 
> **1. Two parallel public-API tiers on the Client.** `types.go` introduces transparent aliases (`type Recipe = recipe.RecipeResult`) alongside the existing facade-owned shape (`type RecipeResult struct {…}`), and the surface has parallel resolve methods returning each (`ResolveRecipe → *RecipeResult` vs `ResolveRecipeFromCriteria/FromSnapshot → *Recipe`). The Stability note on `Snapshot` already concedes the alias tier "inherits pkg/snapshotter's Public (evolving) tier rather than the facade's stable tier — field-shape changes propagate without notice," and a comment points to #1078 as the future wrap. The risk: this PR ships the alias surface that #1078 will need to wrap, and any caller built against `*Recipe` between now and #1078 has to migrate twice. Suggest landing **#1078 first** so the facade-owned wrapper protects external Go importers; then this PR consumes the stable types.
> 
> **2. The PR bundles a wiring refactor with several behavior changes.** Title says "migrate CLI and API to consume aicr.Client" but the implementation notes also include: deep-copy fix in `AdoptRecipe` + `ComponentRef.Overrides` (fixes a prior aliasing bug — i.e., behavior change), validation timeout `time.Duration → *time.Duration` with the CLI passing `0` to **uncap** (real behavior change for CLI users), tolerations explicit-`set` flag (semantic change), and `validate` propagating structured codes via `PropagateOrWrap`. Each of those is independently bisectable and reviewable. Bundling them with a 4,900-line refactor across 41 files in 38 commits makes regression hunting much harder if something breaks downstream. The "preserves behavioral parity" framing also fights the substance — uncapping validation timeout in the CLI is a deliberate change worth calling out plainly.
> 
> Suggested split:
> 
> * Pre-PR: each correctness fix on its own commit with a regression test (`AdoptRecipe` deep-copy is a particularly good standalone candidate — small, testable, ships behavior improvement without coupling to wiring).
> * Pre-PR: per-`DataProvider` criteria registry (replaces the process global) with the back-compat shim's deprecation marker and a tracking comment pointing at #987 for removal.
> * This PR (slimmed): only the CLI/API migration onto the now-stable Client surface. Should drop 1,500–2,000 lines once the above ship separately.
> 
> **3. Back-compat shim sunset plan.** "Per-`DataProvider` criteria registry with a global shim for back-compat" — shims tend to be sticky, and #987 is the sibling that removes the global entirely. Worth confirming the shim is flagged with `//nolint:staticcheck` + a `Deprecated:` doc tag pointing at #987, and that there's a test asserting no in-tree caller still touches the global. Otherwise we end up with a permanent dual code path.
> 
> **4. PR mechanics.** None of this is substance; flagging for the squash-and-rebase pass:
> 
> * Squash: 38 commits → 1 per logical change (after splitting per concern #2 above)
> * Rebase: `mergeable_state: dirty` after #1100/#1101/#1102 landed
> * Sign: `git commit -S` (checkbox is unchecked)
> * `make qualify` checkbox: unchecked — confirm pass locally before flipping to ready-for-review
> 
> **What's clearly good and worth keeping:** the direction (single integration surface, per-instance state), the docs updates landing in the same PR (`go-library.md`, `public-api.md`), the per-file test coverage discipline, and the `Close()` + per-Client guarantee documented in rollout notes. The architectural boundaries are respected — no kubectl orchestration, no in-cluster controllers, no new deployers.
> 
> Happy to pair on any of the splits, especially the #1078 sequencing question. Thanks again for the work.

Thanks, this is a useful read and most of it I agree with. Going point by point:

  1. The two API tiers and #1078 ordering. Fair concern, the double-migrate risk is real. One thing I'd add though: the *Recipe-returning
  methods (ResolveRecipeFromCriteria, ResolveRecipeFromSnapshot, LoadRecipe) aren't an accident. They exist because the lossy facade
  RecipeResult (just Name/Version/Components) physically can't reproduce what the CLI and /v1/recipe actually emit: full metadata,
  constraints, applied overlays, deployment order. So whatever #1078 lands, it can't just be "wrap everything in the lossy struct"; it needs a
   lossless path for these exact callers or the migration regresses output. With that caveat I'm fine landing #1078 first and rebasing on top.
   For what it's worth, nothing external consumes the *Recipe methods yet (in-tree CLI/REST only), so the cost of getting the order wrong
  today is basically nil. But agreed, worth locking down before we invite outside use.

  2. The bundling. You're right. It's already a chain of small, independently bisectable commits, so splitting is mostly mechanical:
  - PR-1: the correctness fixes (AdoptRecipe/Overrides deep-copy, PropagateOrWrap), each with its own regression test. The deep-copy is the
  clean standalone you flagged.
  - PR-2: the per-DataProvider criteria registry plus shim.
  - PR-3 (this one, slimmed): just the CLI/API wiring onto the stable surface.

  And you caught the right thing on "parity": the validation-timeout uncap is a real behavior change, not parity. The old CLI was bounded only
   by per-validator timeouts, and the facade's fixed ValidationOperationTimeout would cancel an all-phase run that hits the 50m inference-perf
   check. I'll say that plainly in the description instead of filing it under parity.

  3. Shim sunset. Will do: Deprecated: plus //nolint:staticcheck on the global shims, and a test that fails if any in-tree caller still
  reaches for the global. One correction though, the global here is the criteria registry (DefaultRegistry/ParseCriteria*), not
  SetDataProvider/GetDataProvider. #987 was the DataProvider globals and it's already closed, so I'll open a separate issue to track the
  registry removal and point the shim there. Otherwise we imply #987 still owns this, and it doesn't.

  4. Mechanics. Commits are -S signed; GitHub shows them Verified. The local --show-signature "No signature" is just this machine missing an
  allowedSignersFile. make qualify passes locally and I'll flip the box after the rebase. I'll rebase onto main once #1100/#1101/#1102 are in
  and squash to one commit per logical change before taking it out of draft.

  The #1078 ordering is the one with real consequences, so I'd like to settle that before I cut PR-1. Happy to pair on it.

mchmarny:
Closing in favor of the 2-PR split @hkii landed:

- **#1107** — `refactor(recipe,validator): per-DataProvider sources + criteria-registry de-globalization` (base: `main`)
- **#1108** — `feat(cli,api): consume aicr.Client facade (#1077)` (stacked on #1107; retargets to `main` once #1107 merges)

Thanks @hkii for splitting this. The 3 correctness fixes called out in earlier review (deep-copy `ComponentRef.Overrides`, deep-copy in `AdoptRecipe`, `PropagateOrWrap` for structured error codes) ride along in #1108's squash rather than as a separate pre-PR — happy to accept that bundling since the diff is small and self-contained within the squash.

Epic tracking moves to #1107 → #1108, then #1078 + #1080 to close #1076.

PR #1101, opened 2026-05-29
feat(recipes): apply nodewright generic tuning to rtx-pro-6000 EKS

yuanchen8911:
## Summary

Wire `nodewright-customizations` with `accelerator=generic` into the rtx-pro-6000 EKS inference overlay, so RTX PRO 6000 nodes get a baseline GPU node-tuning profile — closing a gap vs `h100`/`gb200`/`b200`, which already wire tuning.

## Motivation / Context

The `rtx-pro-6000` EKS overlays deployed only `nodewright-operator` (the controller) with **no** `nodewright-customizations`, so aicr applied no tuning profile (left entirely to the platform). `nvidia-tuned` has no `rtx-pro-6000` profile; the `generic` profile is purpose-built for single-GPU cloud VMs and deliberately omits the NVLink/InfiniBand/hugepage/CPU-isolation settings the `h100`/`gb200` profiles carry. It already ships in `nvidia-tuned` 0.3.0 — the version the tuning manifest pins.

Fixes: N/A
Related: NVIDIA/nodewright-packages#37 (adds the `generic` profile)

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe`) — `recipes/overlays`

## Implementation Notes

- Added the `nodewright-customizations` componentRef to `rtx-pro-6000-eks-inference.yaml` — the `<accelerator>-<service>-<intent>` level, mirroring `h100-eks-inference`/`gb200-eks-inference`. The `-ubuntu-inference`, `-dynamo`, and `-nim` leaves inherit it via overlay merge (the established convention; leaves never re-declare tuning).
- `overrides.accelerator=generic` (the recipe criteria stays `rtx-pro-6000`); added `nodewright-customizations` to the `gpu-operator` `dependencyRefs` for ordering, same as h100.
- eks rtx-pro-6000 is inference-only, so this is a single-file change (no training counterpart).

## Alternatives Considered

[...]

PR #1105, opened 2026-05-29
feat(recipes): add nodewright to bcm with reapply-on-reboot

ayuskauskas:
## Summary

Add the `nodewright-operator` and `nodewright-customizations` components to the BCM overlay, with `reapplyOnReboot` enabled so nodewright customizations are re-applied after BCM rewrites the OS on boot.

## Motivation / Context

BCM re-writes the node OS on every reboot, which wipes any nodewright customizations applied to the host. Enabling `reapplyOnReboot: "true"` ensures the operator re-applies the BCM setup customizations after each reboot, keeping nodes in the expected state.

Fixes: N/A
Related: N/A

## Type of Change

## Component(s) Affected

## Implementation Notes

- Adds `nodewright-operator` (Helm) to `recipes/overlays/bcm.yaml` with `controllerManager.manager.env.reapplyOnReboot: "true"`.
- Adds `nodewright-customizations` (Helm) referencing `components/nodewright-customizations/manifests/bcm-setup.yaml`, with a `dependencyRefs` edge to `nodewright-operator`.
- Node scheduling (selectors/tolerations) for the operator is intentionally left to be supplied at recipe time rather than hardcoded in the overlay.

## Testing

```bash
make qualify
```

## Risk Assessment

- [x] **Low** — Isolated overlay/data change, easy to revert

**Rollout notes:** N/A — additive recipe overlay change.

## Checklist

### DIFF
--- a/recipes/components/nodewright-customizations/manifests/bcm-setup.yaml
+++ b/recipes/components/nodewright-customizations/manifests/bcm-setup.yaml
@@ -1,0 +1,72 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Nodewright EKS Ubuntu GPU Customization
+# Minimal templating - only tolerations/nodeSelectors are dynamic for CLI flag support
+#
+# This customization configures BCM nodes with the minimum necessary for AICR to function.
+{{- $cust := index .Values "nodewright-customizations" }}
+{{- if ne (toString (index $cust "enabled")) "false" }}
+---
+apiVersion: skyhook.nvidia.com/v1alpha1
+kind: Skyhook
+metadata:
+  annotations:
+    # Helm hooks ensure this CR is applied after the Skyhook CRD is installed
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "10"
+    "helm.sh/hook-delete-policy": before-hook-creation
+  labels:
+    app.kubernetes.io/part-of: skyhook-operator
+    app.kubernetes.io/created-by: aicr
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    helm.sh/chart: {{ printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+  name: setup
+  namespace: {{ .Release.Namespace }}
+spec:
+  autoTaintNewNodes: false
+  # Dynamic: Supports --accelerated-node-toleration CLI flag
+  additionalTolerations:
+    {{- if $cust.acceleratedTolerations }}
+    {{- toYaml $cust.acceleratedTolerations | nindent 4 }}
+    {{- else }}
+    # Tolerate-all default, matching ParseTolerations()/DefaultTolerations()
+    # (operator Exists, no key). Only reached if no acceleratedTolerations are
+    # injected from scheduling defaults/flags.
+    - operator: Exists
+    {{- end }}
+  # Dynamic: Supports --accelerated-node-selector CLI flag
+  {{- if $cust.acceleratedNodeSelector }}
+  nodeSelectors:
+    matchLabels:
+      {{- toYaml $cust.acceleratedNodeSelector | nindent 6 }}
+  {{- end }}
+  # Dynamic: Supports --workload-selector CLI flag
+  {{- if $cust.workloadSelector }}
+  podNonInterruptLabels:
+    matchLabels:
+      {{- toYaml $cust.workloadSelector | nindent 6 }}
+  {{- end }}
+  # Tuning
+  packages:
+    nvidia-setup:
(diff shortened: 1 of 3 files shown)


recipes/components/nodewright-customizations/manifests/bcm-setup.yaml
```diff
@@ -0,0 +1,70 @@
...
+  additionalTolerations:
+    {{- if $cust.acceleratedTolerations }}
+    {{- toYaml $cust.acceleratedTolerations | nindent 4 }}
+    {{- else }}
+    - key: dedicated
+      operator: Exists
+    {{- end }}
+  # Dynamic: Supports --accelerated-node-selector CLI flag
+  {{- if $cust.acceleratedNodeSelector }}
+  nodeSelectors:
```
mchmarny, line 51: Toleration default regresses the consistency fix from #1101 (merged into this PR's base). The other three manifests in this directory (`tuning.yaml`, `tuning-gke.yaml`, `no-op.yaml`) now use `- operator: Exists` (tolerate-all) to match `pkg/snapshotter/agent.go:354 DefaultTolerations()`. Change to:

```yaml
- operator: Exists
```

…and the comment block above can mirror the one in `tuning.yaml` ("Tolerate-all default, matching ParseTolerations()/DefaultTolerations()…").

recipes/components/nodewright-customizations/manifests/bcm-setup.yaml
```diff
@@ -0,0 +1,70 @@
...
+    matchLabels:
+      {{- toYaml $cust.workloadSelector | nindent 6 }}
+  {{- end }}
+  # Tuning
+  packages:
+    nvidia-setup:
+      image: ghcr.io/nvidia/nodewright-packages/nvidia-setup
+      version: "0.3.0"
+      containerSHA: sha256:f17c951d60b519d097c20a3d9f49668f043a996adb31b9bb4db24a112a8f60a2
+      configMap:
```
mchmarny, line 67: BOM regen needed. The existing `tuning.yaml` pins `nvidia-setup:0.2.2` (currently in `docs/user/container-images.md`); this new manifest pins `0.3.0` with a new SHA. Per `AGENTS.md`: "After any change to a chart version pin in registry, overlay, or mixin: run `make bom-docs` and commit the regenerated `docs/user/container-images.md` in the same PR."

Run `make bom-docs` and commit the updated BOM. The chart-pin verification gate (`bom-pinning-check`) won't catch this drift on its own since `make bom-check` is opt-in, not wired into the merge gate.

(Separately: is the bump from `0.2.2` to `0.3.0` intentional, or could BCM share the existing 0.2.2 pin to keep the BOM smaller? `0.3.0` is what `tuning-generic.yaml` from #1101 also uses, so there's now a fleet on 0.3.0; tuning.yaml's other consumers — h100/gb200/etc. — still pin 0.2.2.)

recipes/overlays/bcm.yaml
```diff
@@ -93,6 +93,25 @@ spec:
...
+              # Because BCM re-writes the OS on boot, we need to reapply the nodewright customizations on reboot
+              reapplyOnReboot: "true"
+
+    - name: nodewright-customizations
+      type: Helm
+      manifestFiles:
+        - components/nodewright-customizations/manifests/bcm-setup.yaml
+      overrides:
+        service: bcm
+        accelerator: h100
```
mchmarny, line 115: nit: `accelerator: h100` is hardcoded in the BCM overlay's nodewright override, but `bcm.yaml`'s `criteria` block only filters on `service: bcm` (no accelerator dimension). A BCM deployment on a non-H100 accelerator would silently get the H100 tuning profile via this override. Two options:

1. Add `accelerator` to the BCM criteria so the overlay only applies to H100-BCM, then hardcoding is correct.
2. Parameterize the override and create `bcm-h100`-style leaves like the h100/gb200 EKS overlays do.

If BCM is H100-only today by design, a one-line comment here documenting "BCM is H100-only until <X>" closes the loop for future maintainers.

mchmarny (commented):
Focused, real motivation — BCM rewrites the OS on reboot, so `reapplyOnReboot: "true"` is the right knob. CI green across 118 checks.

Three inline comments — two are addressable in a fixup:
- `bcm-setup.yaml`: toleration default regresses the #1101 consistency fix (one-line change to match the post-#1101 pattern on the sibling manifests).
- `bcm-setup.yaml`: BOM regen needed for `nvidia-setup:0.3.0` (existing BOM has `0.2.2`).
- `bcm.yaml`: `accelerator: h100` hardcoded in the override even though the BCM criteria block has no accelerator filter — worth a comment or scoping the override.

Two non-blocking nits for the next push:
- **Title:** `feat: add nodewright to bcm…` → `feat(recipes): add nodewright to bcm with reapply-on-reboot` to match the conventional-commits + `feat(area):` convention in [AGENTS.md "Pull Request Requirements"](https://github.com/NVIDIA/aicr/blob/main/AGENTS.md#pull-request-requirements).
- **Rebase:** branch is `mergeable_state: behind` after #1100/#1101/#1102 landed.

Not blocking the substance — the additive overlay + reapply-on-reboot wiring is good.

recipes/components/nodewright-customizations/manifests/bcm-setup.yaml
ayuskauskas: fixed in latest force push

recipes/components/nodewright-customizations/manifests/bcm-setup.yaml
ayuskauskas: done

ayuskauskas: It needs to update to 0.3.0 for BCM, the others could as well as he 0.2.2 -> 0.3.0 is no-op but I wanted to keep the PR very focused to avoid a big re-qualification.

recipes/overlays/bcm.yaml
ayuskauskas: done

mchmarny (approved):

Issue #1086, opened 2026-05-28
Address recipe gaps surfaced during BCM / H200 NVL validation


yuanchen8911:
## Summary

Cluster-side validation of the freshly-landed BCM service overlay (#feat/bcm-service-type merged to main) on a krusty test cluster with a real H200 NVL node surfaced a small set of recipe-layer gaps spanning shared component values, BCM overlays, and the criteria registry. None block usage today — `aicr validate` deployment+conformance phases pass end-to-end on real Hopper hardware — but they're worth tightening before BCM is called production-ready and before users with H200 hardware start asking why the recipe metadata says `accelerator: h100`.

## Context

Validated against:
- `aicr recipe --service bcm --accelerator h100 --os ubuntu --intent training` (passing `h100` because there's no `h200` criteria value yet — see gap #4 below)
- BCM-provisioned k8s 1.34.8, Ubuntu 24.04, kernel 6.8.0-71
- 2× NVIDIA H200 NVL (141GB HBM3e each) on `node007`
- gpu-operator v26.3.1, nvidia-dra-driver-gpu 25.12.0
- Driver: containerized (GPU operator's `nvidia-driver-daemonset` running)
- DRA ResourceSlice publishes both H200s correctly with the right capacity/arch attributes

## Architectural decision

**Keep the BCM service overlay containerized-driver-only.** Earlier validate failures from the same cluster (DRA pod stuck `Init:0/1`, empty `/run/nvidia/driver`) happened because the BCM base image shipped a pre-installed driver while the AICR overlay assumes the containerized path. The right fix is to strip the driver from the BCM image (in progress with the BCM team), not to add a host-installed-driver toggle in the overlay. One mode, fewer footguns.

## Gaps to fix

- [ ] **Shared DRA component** — `recipes/components/nvidia-dra-driver-gpu/values.yaml:74,78`: `controller.priorityClassName: ""` and `kubeletPlugin.priorityClassName: ""` neutralize the chart default (`system-node-critical`) without explanation. Either add an inline comment with the rationale, or restore the chart default. Neutralizing `system-node-critical` lets DRA pods be evicted under node pressure, which would break GPU allocation for running pods.

[...]

PR #1106, opened 2026-05-29
harden(recipes): lockstep DRA driver root with operator install dir

### LINKED ISSUE #1087: Add integration test: nvidiaDriverRoot and gpu-operator.hostPaths.driverInstallDir must stay in lockstep across overlays
## Summary

`nvidia-dra-driver-gpu.nvidiaDriverRoot` and `gpu-operator.hostPaths.driverInstallDir` MUST point to the same path. They're independently configurable across overlays and silently break if they drift. There's no schema link between the two, so an overlay editor can change one without the other and CI won't catch it.

## Why

The DRA kubelet plugin loads the NVIDIA driver userspace (`libnvidia-ml.so`, `nvidia-smi`, `nvidia-ctk`) from `nvidiaDriverRoot`. GPU operator's `hostPaths.driverInstallDir` is where the operator-managed driver container mounts its rootfs onto the host. If these diverge:

- DRA fails CDI spec generation (`Driver/library version mismatch` or missing `libnvidia-ml.so`)
- DRA-allocated pods stay in `ContainerCreating`
- `aicr validate` deployment phase fails

This is the symptom that surfaced on the krusty cluster during BCM service validation. The root cause was a different host-vs-container driver question, but the same class of bug (the two paths assuming different driver locations) is the underlying risk.

## Acceptance

- Test enumerates every overlay × intent × platform combination (use the recipe builder to fully hydrate) and asserts the two resolved values match
- Test runs in `make qualify` / CI
- Test emits a clear diff message identifying the overlay file when they don't match

## Implementation notes

Likely lives next to existing overlay-resolution tests in `pkg/recipe/...`. Iterate `AllCriteriaAcceleratorTypes`, `AllCriteriaServiceTypes`, `AllCriteriaIntentTypes`, etc., resolve each combination, and assert. Skip combinations the registry rejects.

Related: #1086 (BCM overlay gaps), where this lockstep would have prevented an earlier debug iteration.

mchmarny:
## Summary

Add `TestDriverRootLockstep` (and a supporting `TestStringAtPath`) in `pkg/recipe` to enforce that, for every leaf overlay shipping both `nvidia-dra-driver-gpu` and `gpu-operator`, the resolved `nvidia-dra-driver-gpu.nvidiaDriverRoot` equals the resolved `gpu-operator.hostPaths.driverInstallDir`. This is the lockstep guard requested in #1087.

## Motivation / Context

These two paths must point to the same on-host driver location, but they're independently configurable across overlays with no schema link between them. An overlay editor can change one without the other and CI won't notice. When they drift:

- DRA fails CDI spec generation (`Driver/library version mismatch` or missing `libnvidia-ml.so`)
- DRA-allocated pods stay in `ContainerCreating`
- `aicr validate` deployment phase fails

Fixes: #1087
Related: #1086 (BCM overlay gaps — where this lockstep would have prevented an earlier debug iteration)

## Type of Change

## Component(s) Affected

## Implementation Notes

**Discovery:** mirrors `TestBothBuildPathsProduceIdenticalContent` — walks `loadMetadataStore` for every overlay not referenced as `spec.base` (the leaves). Each leaf is resolved through the production `BuildRecipeResult` so the wildcard contributions and mixins are exercised.

**Assertion tiers** (per the validation-phase-floor test precedent):
- **Hard fail** when both values are explicitly set and differ — the core silent-drift case the issue is guarding against. This is what will catch a future regression.
- **Warn-only** (promoted to fail under `AICR_DRIVER_ROOT_LOCKSTEP_STRICT=1`) when exactly one side is explicitly set and the other falls through to the upstream chart default. The chart default isn't visible to this test, so the lockstep is unverifiable rather than definitively broken. Strict mode is for the eventual cleanup PR that makes every overlay's pair explicit.
- **Pass** when both are unset (relying on chart defaults — out of scope of overlay correctness) or both set and match.

**Why two-tier:** running the test against current `main` reveals 25 leaf overlays where `gpu-operator.hostPaths.driverInstallDir` falls through to the chart default while `nvidia-dra-driver-gpu.nvidiaDriverRoot` is explicitly set via the dra-driver's base `values.yaml`. In practice these coincidentally lockstep (both at `/run/nvidia/driver`), but the lockstep is not enforced by the recipe. Cleaning that up is a separate follow-up; this PR ships the guard now and leaves the cleanup tractable via the strict-mode toggle.

**Coverage:**
- Iterates all 32 discovered leaf overlays
- 7 of them have both values explicitly set today (gke-cos, eks-training, etc.) — those exercise the hard-fail path's positive case.
- The remaining 25 trigger the soft warn path; strict mode flips them to errors.
- `TestStringAtPath` covers the nested-map traversal helper.

## Testing

```bash
make qualify
```

Results:
- All Go tests pass with `-race`.
- `golangci-lint` clean (0 issues on `./pkg/recipe/...`).
- Chainsaw e2e: 22 passed / 0 failed.
- Vuln scan: clean.

Verified the test catches today's gaps in strict mode:

```bash
$ AICR_DRIVER_ROOT_LOCKSTEP_STRICT=1 go test -run TestDriverRootLockstep ./pkg/recipe/
FAIL: 25 overlays where one side is unset
```

…and passes in non-strict (default) mode, which is what `make qualify` runs.

Coverage delta (`pkg/recipe`): 91.9% (no change — the new file is mostly assertions exercising existing code paths).

## Risk Assessment

- [x] **Low** — Test-only change. No production code touched. Easy to revert; easy to evolve via the strict-mode toggle.

**Rollout notes:** Default-mode CI runs see the test pass today. A follow-up PR that makes the unset-side explicit on each overlay can flip CI to `AICR_DRIVER_ROOT_LOCKSTEP_STRICT=1` to promote the warnings to errors.

## Checklist

### DIFF
--- a/pkg/recipe/driver_root_lockstep_test.go
+++ b/pkg/recipe/driver_root_lockstep_test.go
@@ -16,21 +16,13 @@ package recipe
 
 import (
 	"context"
-	"fmt"
-	"os"
 	"testing"
 )
 
-// driverRootLockstepStrictEnvVar promotes "lockstep unverifiable"
-// findings (one side explicitly set, the other unset and falling
-// through to the upstream chart default) from warnings to test
-// failures. Used by CI jobs that want the stronger contract.
-const driverRootLockstepStrictEnvVar = "AICR_DRIVER_ROOT_LOCKSTEP_STRICT"
-
-// TestDriverRootLockstep enforces that, for every leaf overlay that ships
-// both nvidia-dra-driver-gpu and gpu-operator, the resolved values of
+// TestDriverRootLockstep enforces that, for every overlay carrying both
+// nvidia-dra-driver-gpu and gpu-operator, the resolved values of
 // nvidia-dra-driver-gpu.nvidiaDriverRoot and
-// gpu-operator.hostPaths.driverInstallDir do not silently diverge.
+// gpu-operator.hostPaths.driverInstallDir are explicitly set and identical.
 //
 // Why this lockstep matters (see issue #1087):
 // The DRA kubelet plugin loads the NVIDIA driver userspace
@@ -44,19 +36,24 @@ const driverRootLockstepStrictEnvVar = "AICR_DRIVER_ROOT_LOCKSTEP_STRICT"
 // between the two fields, so an overlay editor can change one without
 // the other and CI won't notice — this test is the only guard.
 //
-// The test walks every leaf overlay (same discovery pattern used by
-// TestBothBuildPathsProduceIdenticalContent), resolves it through the
-// production builder, and inspects both resolved values:
+// **Discovery.** The test iterates every overlay with non-nil
+// Spec.Criteria. The earlier draft restricted to "leaf" overlays
+// (overlays not referenced as spec.base by any other overlay), but
+// production resolution is per-query: FindMatchingOverlays →
+// filterToMaximalLeaves drops an overlay only when a matching descendant
+// exists for that query. E.g., h100-gke-cos-training is a base for
+// -kubeflow/-slurm leaves (which require platform=kubeflow/slurm), so a
+// {h100, gke-cos, training} query without platform resolves to it
+// directly in production. The earlier filter would miss that.
 //
-//   - Both explicitly set and differ: HARD FAIL. This is the core
-//     drift case the issue is guarding against.
-//   - Exactly one explicitly set: warn-only by default; promoted to
-//     fail when AICR_DRIVER_ROOT_LOCKSTEP_STRICT=1. The other side
-//     falls through to the upstream chart default which this test
-//     cannot read, so the lockstep is unverifiable rather than
-//     definitively broken. The strict mode is for the eventual
-//     cleanup PR that makes every overlay explicit.
-//   - Both unset / both set and match: pass.
+// **Assertion.** Both resolved values must be explicitly set (non-empty
+// in the resolved Helm values map) AND identical. An empty value falls
+// through to the upstream chart's bundled default, which the test cannot
+// read — and per-component defaults differ (GPU Operator chart 26.3.1
+// defaults driverInstallDir to /run/nvidia/driver, but DRA chart 25.12.0
+// defaults nvidiaDriverRoot to /). Relying on chart defaults is itself
+// drift waiting to happen on the next chart bump, so the test treats
+// "not explicitly set on both" as a failure.
 func TestDriverRootLockstep(t *testing.T) {
 	ctx := context.Background()
 	store, err := loadMetadataStore(ctx)
@@ -64,29 +61,13 @@ func TestDriverRootLockstep(t *testing.T) {
 		t.Fatalf("loadMetadataStore: %v", err)
 	}
 
-	strict := os.Getenv(driverRootLockstepStrictEnvVar) == "1"
-	t.Logf("strict mode (%s=1): %t", driverRootLockstepStrictEnvVar, strict)
-
-	// Discover leaf overlays: overlays not referenced as spec.base by any
-	// other overlay. Mirrors the discovery in
-	// TestBothBuildPathsProduceIdenticalContent.
-	referencedAsBases := make(map[string]bool, len(store.Overlays))
-	for _, overlay := range store.Overlays {
-		if overlay.Spec.Base != "" {
-			referencedAsBases[overlay.Spec.Base] = true
-		}
-	}
-
-	leafCount := 0
+	overlayCount := 0
 	checked := 0
 	for name, overlay := range store.Overlays {
-		if referencedAsBases[name] {
-			continue
-		}
 		if overlay.Spec.Criteria == nil {
 			continue
 		}
-		leafCount++
+		overlayCount++
 
 		t.Run(name, func(t *testing.T) {
 			result, err := store.BuildRecipeResult(ctx, overlay.Spec.Criteria)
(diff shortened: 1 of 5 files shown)


pkg/recipe/driver_root_lockstep_test.go
```diff
@@ -0,0 +1,230 @@
...
+//   - Exactly one explicitly set: warn-only by default; promoted to
+//     fail when AICR_DRIVER_ROOT_LOCKSTEP_STRICT=1. The other side
+//     falls through to the upstream chart default which this test
+//     cannot read, so the lockstep is unverifiable rather than
+//     definitively broken. The strict mode is for the eventual
+//     cleanup PR that makes every overlay explicit.
+//   - Both unset / both set and match: pass.
+func TestDriverRootLockstep(t *testing.T) {
+	ctx := context.Background()
+	store, err := loadMetadataStore(ctx)
```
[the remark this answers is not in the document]
mchmarny: Declining — `context.Background()` is the established pattern for tests at this level in `pkg/recipe/`:

- `TestOverlayValidationPhaseFloor` at `validation_phase_floor_test.go:235` (which this test mirrors structurally)
- `TestBothBuildPathsProduceIdenticalContent` at `metadata_store_test.go:518` (the discovery pattern I copied)
- Plus 8 more `context.Background()` call sites in `metadata_store_test.go`

`AGENTS.md` "Context Propagation Rules" explicitly carves out test setup: *"`context.Background()` is acceptable ONLY for: cleanup in deferred functions (when parent context is canceled), graceful shutdown, and `test setup`."* The `go test` runner already enforces a per-package wall-clock timeout via `-timeout` (default 10m), which is the right place for the bound — adding a redundant in-test timeout would shadow the global and surprise anyone debugging a hang under `go test -timeout 1h`.

yuanchen8911:
Thanks for tackling #1087 — the invariant is real and the direction is right. There are a few gaps.

pkg/recipe/driver_root_lockstep_test.go
```diff
@@ -0,0 +1,230 @@
...
+			// through to the upstream chart default, which this test
+			// cannot read — so the lockstep is unverifiable rather than
+			// definitively broken. Strict mode is for the eventual
+			// cleanup PR that makes every overlay's pair explicit.
+			report := func(msg string) {
+				if strict {
+					t.Error(msg)
+				} else {
+					t.Logf("UNVERIFIED LOCKSTEP: %s", msg)
+				}
```
yuanchen8911: **Issue 1 — One-sided drift passes default CI**

`make test` runs `go test` without `-v` (Makefile:200) and `AICR_DRIVER_ROOT_LOCKSTEP_STRICT` is not wired into CI, so the one-sided branch (`report()` → `t.Logf`) is both suppressed and non-failing by default. The hard-fail fires only when both values are explicit and differ. With 25 overlays one-sided today, an edit to the explicit side ships green — the guard for the common case is effectively dormant.

Recommended fix: require both resolved values to be explicit and equal, and make a violation a hard failure in normal CI (no opt-in). Avoid hardcoding a shared default — the defaults differ per component (GPU Operator v26.3.1 defaults `driverInstallDir` to `/run/nvidia/driver`, but the DRA chart 25.12.0 default is `/`; AICR's DRA values file is what sets `/run/nvidia/driver`). So either compare each side against its actual expected per-component default, or simply treat "not explicitly set on both" as unverifiable and fail it by default. This also lets you retire the two-tier toggle once overlays are explicit.

pkg/recipe/driver_root_lockstep_test.go
```diff
@@ -0,0 +1,230 @@
...
+		if overlay.Spec.Base != "" {
+			referencedAsBases[overlay.Spec.Base] = true
+		}
+	}
+
+	leafCount := 0
+	checked := 0
+	for name, overlay := range store.Overlays {
+		if referencedAsBases[name] {
+			continue
```
yuanchen8911: **Issue 2 — Leaf discovery skips production-resolvable intermediates**

The global `referencedAsBases` skip drops every overlay used as a base. But production resolution is per-query: `FindMatchingOverlays` → `filterToMaximalLeaves` (metadata_store.go:438/469) drops an overlay only when a matching descendant exists for that query. `h100-gke-cos-training` is a base for `-kubeflow`/`-slurm` (which require `platform=kubeflow`/`slurm`), so a `{h100, gke-cos, training}` query with no platform resolves it directly in production — yet this test never builds it. A mismatch there, overridden by every descendant, reaches users untested.

Recommended fix: drive discovery off the production path — enumerate the real criteria space and resolve via `FindMatchingOverlays`, or build every overlay with non-nil `Spec.Criteria` rather than only global leaves.

pkg/recipe/driver_root_lockstep_test.go
mchmarny: Accepted — fixed in ae912bdb. You're right that warn-mode was dormant by default (`make test` runs without `-v`; `AICR_DRIVER_ROOT_LOCKSTEP_STRICT` not wired into CI), so 25 overlays were unverified. And the chart-default-divergence point (gpu-operator chart 26.3.1 at `/run/nvidia/driver`, DRA chart 25.12.0 at `/`) is decisive — relying on unset values is itself drift waiting to happen on the next chart bump.

The test now hard-fails on all three cases (explicit divergence, one-sided, both unset) and the strict-mode toggle is retired. To make the tightened test pass, I made every overlay's gpu-operator side explicit:

- `recipes/components/gpu-operator/values.yaml`: base now sets `hostPaths.driverInstallDir: /run/nvidia/driver` (matches the nvidia-dra-driver-gpu base value).
- `recipes/components/gpu-operator/values-oke.yaml` + `values-oke-training.yaml`: `hostPaths.driverInstallDir: /` (matches OKE's `nvidia-dra-driver-gpu/values-oke.yaml`).
- `recipes/overlays/kind.yaml`: added gpu-operator override `hostPaths.driverInstallDir: "/"` (matches kind's existing `nvidiaDriverRoot: "/"`).

All 78 covered overlays now resolve to explicit, matching values on both sides.

pkg/recipe/driver_root_lockstep_test.go
mchmarny: Accepted — fixed in ae912bdb. The `referencedAsBases` filter is dropped; the test now iterates every overlay with non-nil `Spec.Criteria`. Coverage went from 32 (leaves only) to 78 (every criteria-bearing overlay), so production-resolvable intermediates like `h100-gke-cos-training` (resolvable directly when a query omits `platform=kubeflow/slurm`) are now exercised.

Implementation note in the test header now explains the choice with a reference to the `FindMatchingOverlays` → `filterToMaximalLeaves` behavior you cited, so the next person doesn't re-introduce the leaf-only filter for "simplicity."

yuanchen8911 (approved):
/lgtm

Both review findings are fully addressed (verified against the branch, including a negative test confirming the guard hard-fails on one-sided drift in default `make qualify` mode, and that intermediates like `eks`/`bcm` are now covered — 78/78 overlays pass).

One follow-up: update the PR description — it still describes the old strict-mode / leaf-only / test-only version (the `AICR_DRIVER_ROOT_LOCKSTEP_STRICT` toggle is retired, discovery now covers all 78 overlays not just 32 leaves, and the PR now also modifies gpu-operator values files, so it's no longer test-only).

PR #1107, opened 2026-05-29
refactor(recipe,validator): per-DataProvider sources + criteria-registry de-globalization

hkii:
## Summary
Thread an explicit `DataProvider` through the recipe loader, validator catalog, and criteria parsing so consumers can resolve against their own data source instead of the process-global provider/registry. Foundation (PR 1 of 2) for migrating the CLI/REST onto `aicr.Client` (#1077).

## What changes
- `recipe`: `GetCriteriaRegistryFor(dp)` per-provider criteria registry (+ `EvictCachedCriteriaRegistry`); registry-aware parsing + `BuildCriteriaWithRegistry`; `LoadFromFileWithProvider`; `RecipeResult.DeepCopy`/`BindDataProvider`.
- `validator`: `catalog.LoadWithDataProvider(dp, …)` + `WithDataProvider` option.
- `config`: `ResolveCriteriaWithRegistry`.

## Behavior
No consumer-facing behavior change. Global shims are retained for back-compat (`DefaultRegistry`, `ParseCriteria*`, `catalog.Load`, `LoadFromFile`, `config.ResolveCriteria`) so existing callers compile and behave identically. The criteria-registry global is distinct from `SetDataProvider`/`GetDataProvider` (#987, closed); shim removal will be tracked separately.

## Cache lifecycle
This PR adds `criteriaRegistryCache` (per-`DataProvider`, `sync.Map`-backed, lazy via `sync.Once`) alongside the existing `registryCache` (`pkg/recipe/components.go`) and `storeCache` (`pkg/recipe/metadata_store.go`). All three caches grow on first access per `DataProvider` and have `Evict*` primitives but no automatic eviction from this PR alone.

`Client.Close()` wiring that calls `EvictCachedCriteriaRegistry`, `EvictCachedStore`, and `EvictCachedRegistry` for the Client's `DataProvider` lands in #1108. Long-running consumers that churn through Clients should wait for #1108 (or evict manually) to avoid monotonic cache growth.

## Testing
`go build ./...`, `go test -race ./pkg/recipe/... ./pkg/validator/... ./pkg/config/...`, and `golangci-lint` all pass standalone.

Relates to #1077. PR 2 (CLI/API migration) stacks on this branch.

### DIFF
--- a/pkg/recipe/criteria.go
+++ b/pkg/recipe/criteria.go
@@ -639,7 +639,17 @@ func (c *Criteria) String() string {
 	return fmt.Sprintf("criteria(%s)", strings.Join(parts, ", "))
 }
 
-// CriteriaOption is a functional option for building Criteria.
+// CriteriaOption is a functional option for building Criteria. Each option
+// resolves its enum value via the package-level ParseCriteria*Type shims,
+// which consult DefaultRegistry() — the registry bound to the package-global
+// DataProvider.
+//
+// For per-provider isolation (e.g., when building Criteria against an
+// aicr.Client's own DataProvider so non-OSS values declared in a `--data`
+// overlay validate against THAT provider's registered values), use
+// RegistryCriteriaOption + BuildCriteriaWithRegistry instead. The two option
+// types deliberately have distinct signatures because RegistryCriteriaOption
+// takes the registry as an explicit parameter; they are not interchangeable.
 type CriteriaOption func(*Criteria) error
 
 // WithCriteriaService sets the service type.
@@ -713,10 +723,16 @@ func WithCriteriaNodes(n int) CriteriaOption {
 	}
 }
 
-// BuildCriteria creates a Criteria from functional options, resolving each
-// field against the package-global criteria registry. It is a shim over
-// BuildCriteriaWithRegistry(DefaultRegistry(), ...); callers holding an
-// explicit per-provider registry should call that directly.
+// BuildCriteria creates a Criteria from functional options. Each option
+// resolves its enum value through the package-level ParseCriteria*Type shims
+// against DefaultRegistry() (the registry bound to the package-global
+// DataProvider).
+//
+// For per-provider isolation, use BuildCriteriaWithRegistry with
+// RegistryCriteriaOption — that path threads an explicit *CriteriaRegistry
+// (typically from GetCriteriaRegistryFor) so registered values from a
+// caller-owned DataProvider's overlays are honored instead of (or in
+// addition to) the package-global registry.
 func BuildCriteria(opts ...CriteriaOption) (*Criteria, error) {
 	c := NewCriteria()
 	for _, opt := range opts {
--- a/pkg/recipe/criteria.go
+++ b/pkg/recipe/criteria.go
@@ -56,15 +56,25 @@ const (
 	CriteriaServiceBCM  CriteriaServiceType = "bcm"
 )
 
-// ParseCriteriaServiceType parses a string into a CriteriaServiceType.
+// ParseCriteriaServiceType parses a string into a CriteriaServiceType using
+// the package-global criteria registry. It is a shim over
+// (*CriteriaRegistry).ParseService bound to DefaultRegistry(); callers
+// holding an explicit registry (e.g., from GetCriteriaRegistryFor) should
+// call the method directly for per-provider isolation.
+func ParseCriteriaServiceType(s string) (CriteriaServiceType, error) {
+	return DefaultRegistry().ParseService(s)
+}
+
+// ParseService parses a string into a CriteriaServiceType against this
+// registry.
 //
 // The switch arms below are the canonical/aliased fast path for the
 // embedded OSS catalog. Any value not recognized here falls through to
-// the package criteria registry, which the data provider seeds from
-// loaded overlays (embedded + `--data`). This lets internal/proprietary
-// service values (e.g., undisclosed NCPs) be admitted at runtime via
-// `--data` without a binary rebuild.
-func ParseCriteriaServiceType(s string) (CriteriaServiceType, error) {
+// the registry, which the data provider seeds from loaded overlays
+// (embedded + `--data`). This lets internal/proprietary service values
+// (e.g., undisclosed NCPs) be admitted at runtime via `--data` without a
+// binary rebuild.
+func (reg *CriteriaRegistry) ParseService(s string) (CriteriaServiceType, error) {
 	switch strings.ToLower(strings.TrimSpace(s)) {
 	case "", CriteriaAnyValue, "self-managed", "self", "vanilla":
 		return CriteriaServiceAny, nil
@@ -83,7 +93,7 @@ func ParseCriteriaServiceType(s string) (CriteriaServiceType, error) {
 	case "bcm":
 		return CriteriaServiceBCM, nil
 	default:
-		if DefaultRegistry().Has(FieldService, s) {
+		if reg.Has(FieldService, s) {
 			return CriteriaServiceType(normalizeCriteriaValue(s)), nil
 		}
 		return CriteriaServiceAny, errors.New(errors.ErrCodeInvalidRequest, fmt.Sprintf("invalid service type: %s", s))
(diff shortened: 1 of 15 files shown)


pkg/recipe/criteria_registry.go
```diff
@@ -215,25 +215,89 @@ func (r *CriteriaRegistry) Reset() {
...
-func DefaultRegistry() *CriteriaRegistry {
-	defaultRegistryOnce.Do(func() {
-		defaultRegistry = newCriteriaRegistry()
+// criteriaRegistryCache holds criteriaRegistryCacheEntry pointers keyed by
+// DataProvider identity. Two callers bound to different DataProvider values
+// populate distinct entries; a single provider value yields a single shared
+// registry regardless of caller goroutine count. EvictCachedCriteriaRegistry
+// drops a single entry so a caller can force a rebuild after rotating a
+// provider's backing data without disturbing other providers.
+var criteriaRegistryCache sync.Map // map[DataProvider]*criteriaRegistryCacheEntry
```
mchmarny, line 235: `sync.Map.LoadOrStore` keys by interface value, which requires the dynamic type to be comparable. In-tree implementations (`*EmbeddedDataProvider`, `*LayeredDataProvider`) are pointer-receiver and safe, but `DataProvider` is exported — an external implementer using a struct receiver with a slice/map/func field will hit a runtime panic on first `GetCriteriaRegistryFor` call.

Same constraint already applies to the existing `registryCache` and `storeCache`, but the documented API surface doesn't say so. Suggest adding a sentence to the `DataProvider` interface godoc: "Implementations must be comparable per the Go language spec (typically: define methods on a pointer receiver)." Cheap, prevents a foot-gun for external authors per `docs/integrator/public-api.md`'s "Public (evolving)" contract on `pkg/recipe`.

pkg/recipe/criteria.go
```diff
@@ -653,6 +727,116 @@ func BuildCriteria(opts ...CriteriaOption) (*Criteria, error) {
...
 }
 
+// RegistryCriteriaOption is a functional option for building a Criteria
+// against an explicit *CriteriaRegistry. Unlike CriteriaOption (which
+// closes over the package-global registry through the ParseCriteria*Type
+// shims), a RegistryCriteriaOption resolves its enum value against the
+// registry threaded in by BuildCriteriaWithRegistry, so a caller holding a
+// per-provider registry (from GetCriteriaRegistryFor) builds and validates
+// criteria against THAT provider's registered values.
+type RegistryCriteriaOption func(reg *CriteriaRegistry, c *Criteria) error
```
mchmarny, line 753: API-duplication question — what's the long-term plan?

This PR ships a parallel API: every existing `With<X>` option now has a `With<X>Registry` sibling (Service, Accelerator, Intent, OS, Platform, Nodes), plus a parallel `BuildCriteriaWithRegistry`. Net new exports: 1 type + 7 funcs, all of which duplicate existing semantics except that they route through an explicit registry instead of `DefaultRegistry()`.

Two coherent options:

1. **Deprecate the originals** — add `// Deprecated: prefer With<X>Registry / BuildCriteriaWithRegistry for per-provider isolation` on `WithService`/`WithAccelerator`/.../`BuildCriteria`, with removal tracked alongside the criteria-registry shim. Migrates consumers cleanly.
2. **Keep both indefinitely** — add a doc note on `BuildCriteria` explaining why both APIs coexist (e.g., CLI uses package-global default for ergonomics, library callers thread per-provider for isolation).

The PR currently ships option (1)-shaped code without the `Deprecated:` markers, which makes the intent ambiguous for external Go importers. Pick one and document it.

pkg/recipe/criteria_registry.go
```diff
@@ -215,25 +215,89 @@ func (r *CriteriaRegistry) Reset() {
...
-// DefaultRegistry returns the process-wide criteria registry, creating
-// it on first access. Threading the registry through every call site
-// would add significant churn for a value that is set once at startup
-// (when the data provider populates it from loaded overlays) and read
-// many times thereafter, so a singleton matches the existing package
-// shape better than dependency injection.
-func DefaultRegistry() *CriteriaRegistry {
-	defaultRegistryOnce.Do(func() {
-		defaultRegistry = newCriteriaRegistry()
+// criteriaRegistryCache holds criteriaRegistryCacheEntry pointers keyed by
```
mchmarny, line 229: Cache-eviction wiring — worth a forward reference in the PR body.

`criteriaRegistryCache` joins the existing `registryCache` (components.go) and `storeCache` (metadata_store.go). All three need `Evict*` calls from `Client.Close()` to avoid monotonic growth in long-running processes that churn Clients (per `aicr.go`'s documented "retain and reuse Clients" guidance).

That `Close()` wiring lives in #1108. Two options:

- Note in this PR's description that "Close() wiring for all three caches lands in #1108" so the cache lifecycle isn't a hidden assumption when this PR merges alone.
- Or land a small test here that asserts `CachedCriteriaRegistryContainsForTesting(dp)` returns true after load and false after a manual `EvictCachedCriteriaRegistry(dp)` — gives #1108's wiring an observable target.

Not a blocker for this PR. The first option (PR-body note) is probably enough.

pkg/validator/catalog/catalog.go
```diff
@@ -59,8 +69,11 @@ type (
 //
 // Entries with explicit version tags (e.g., :v1.2.3) are never modified by
 // steps 1-2 but are replaced by step 3 if that env var is set.
-func Load(version, commit string) (*ValidatorCatalog, error) {
-	data, err := recipe.GetDataProvider().ReadFile("validators/catalog.yaml") //nolint:staticcheck // tracked by #983 Stage 2
+func LoadWithDataProvider(dp recipe.DataProvider, version, commit string) (*ValidatorCatalog, error) {
+	if dp == nil {
+		dp = recipe.GetDataProvider() //nolint:staticcheck // back-compat fallback; tracked by #983/#987
```
mchmarny: nit: inconsistent `nolint` reference style across the PR.

- Here: `// back-compat fallback; tracked by #983/#987`
- `pkg/recipe/criteria_registry.go:250` / `:300`: `// back-compat fallback for pre-WithDataProvider callers (#983 Stage 2)`

#987 is closed (Stage 1 deprecation marker). Either pick the actual shim-removal follow-up issue number (does one exist yet?) or standardize on `(#983 Stage 2)` everywhere. Three minutes of `sed`.

mchmarny (commented):
Clean foundation refactor — thanks for splitting this off, the per-provider isolation model is well-shaped and the doc comments are exceptional. Per-provider isolation test, eviction test, strict-per-registry test, and the marker-validator proof in `TestLoadHonorsExplicitProvider` together pin the behavior I'd want pinned. `RecipeResult.DeepCopy` handling nested `Overrides` via `deepCopyAnyMap` cleanly addresses the prior aliasing bug. All 32 checks green.

Four discussion points, none blocking:

- **M1** (`criteria_registry.go`): `sync.Map` keyed by `DataProvider` interface needs the dynamic type to be comparable — worth one sentence in the `DataProvider` interface godoc so external implementers don't trip a runtime panic.
- **M2** (`criteria.go`): the new `With<X>Registry` / `BuildCriteriaWithRegistry` ship as a parallel API alongside the existing `With<X>` / `BuildCriteria` — what's the long-term plan? Either `Deprecated:` markers on the originals or a doc note on why both coexist.
- **M3** (`criteria_registry.go`): cache-eviction wiring for the new `criteriaRegistryCache` lives in #1108 — worth a forward reference in this PR's body so the cache lifecycle isn't a hidden assumption.
- **L1** (nit): `//nolint:staticcheck` reference style varies (`#983 Stage 2` vs `#983/#987`); standardize.

Happy to take this as-is once M1/M2/M3 get a quick pass.

mchmarny:
Pushed eac3ecf9 addressing M1/M2/L1 inline; updated PR body with a Cache lifecycle section addressing M3. Summary:

- **M1** `pkg/recipe/provider.go`: `DataProvider` interface godoc now states the comparability requirement and explains the sync.Map foot-gun for external implementers.
- **M2** `pkg/recipe/criteria.go`: `CriteriaOption` and `BuildCriteria` docs now correctly describe the relationship to `RegistryCriteriaOption` / `BuildCriteriaWithRegistry` (dropping the misleading 'shim over' wording) and explicitly point per-provider callers at the new path. No `Deprecated:` markers yet — deferring those to a follow-up PR after #1108 migrates in-tree callers to avoid SA1019 churn.
- **L1** `pkg/validator/catalog/catalog.go`: `//nolint:staticcheck` reference style now matches `criteria_registry.go`.
- **M3** PR body: added a "Cache lifecycle" section noting all three caches (`criteriaRegistryCache` + the existing `registryCache` / `storeCache`) have their `Close()` wiring landing in #1108.

`make test` + `golangci-lint` clean on the affected packages. Stepping back from this review — handing off to another reviewer.

mchmarny (approved):

pkg/validator/catalog/catalog.go
```diff
@@ -59,8 +69,11 @@ type (
...
+	}
+	data, err := dp.ReadFile("validators/catalog.yaml")
 	if err != nil {
```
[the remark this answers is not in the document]
mchmarny: Out of scope for this PR — `LoadWithDataProvider` preserves the shape of the pre-existing `Load(version, commit)` on `main`, which also doesn't accept context. Adding context plumbing would require changing the `DataProvider` interface itself (`ReadFile(path)` → `ReadFile(ctx, path)`), which ripples through three in-tree implementations and any external implementers.

Concern is legitimate, though — tracked as #1109. Resolving here.

pkg/recipe/criteria_registry.go
mchmarny: Addressed in eac3ecf9 — `DataProvider` interface godoc now states the comparability requirement and explains the `sync.Map` keying foot-gun for external implementers. Resolving.

pkg/recipe/criteria.go
mchmarny: Addressed in eac3ecf9 — picked option (2) variant: `CriteriaOption` + `BuildCriteria` docs now correctly describe the relationship to `RegistryCriteriaOption` / `BuildCriteriaWithRegistry` (dropping the misleading "shim over" wording) and explicitly point per-provider callers at the new path. No `Deprecated:` markers yet — deferred to a follow-up PR after #1108 migrates in-tree callers, to avoid triggering SA1019 churn on still-active call sites. Resolving.

pkg/recipe/criteria_registry.go
mchmarny: Addressed in eac3ecf9 — PR body now has a "Cache lifecycle" section noting that `Close()` wiring for all three caches (`criteriaRegistryCache` + the existing `registryCache` / `storeCache`) lands in #1108. Resolving.

Issue #1011, opened 2026-05-21
feat(bundler): configurable parent Application name for multi-bundle clusters


xdu31:
## Summary

The `argocd-helm` and `argocd` deployers hardcode the parent Application name (`aicr-stack` and `nvidia-stack` respectively). When two bundles from different recipes (with non-overlapping components) are installed on the same cluster, the second bundle's parent Application silently overwrites the first, orphaning the first bundle's child Applications.

## Motivation / Context

Multiple CRDs targeting the same cluster would each install a bundle. Today this fails because both produce `name: aicr-stack` in the parent Application template.

## Current Behavior

**`argocd-helm` deployer** — `pkg/bundler/deployer/argocdhelm/argocdhelm.go:377`:
```yaml
metadata:
  name: aicr-stack      # hardcoded, no .Values override path
  namespace: argocd
```

**`argocd` deployer** — `pkg/bundler/deployer/argocd/templates/app-of-apps.yaml.tmpl:4`:
```yaml
metadata:
  name: nvidia-stack     # hardcoded
  namespace: argocd
```

Installing a second bundle overwrites the first parent Application. The first bundle's components become unmanaged.

## Expected Behavior

The parent Application name should be configurable so multiple bundles can coexist:

```bash
# Bundle A
aicr bundle -r gpu-recipe.yaml --deployer argocd-helm \
  --app-name gpu-runtime -o ./gpu-bundle

# Bundle B
aicr bundle -r ops-recipe.yaml --deployer argocd-helm \
  --app-name ops-runtime -o ./ops-bundle
```

## Proposed Solution

1. Add `--app-name` CLI flag to `aicr bundle`, defaulting to `aicr-stack` (argocd-helm) / `nvidia-stack` (argocd) for backward compatibility
2. In `argocd-helm`: expose via `.Values.appName` so it can also be overridden at `helm install` time
3. In `argocd`: template the name in `app-of-apps.yaml.tmpl`
4. Derive the template filename from the app name (e.g., `templates/<app-name>.yaml` instead of `templates/aicr-stack.yaml`)

## Additional Context

Audit of other hardcoded identifiers that affect multi-bundle:

| Identifier | Deployer | Workaround |
|-----------|----------|------------|
| Helm release name `aicr-bundle` | argocd-helm | Operator picks different name at `helm install` time |
| Chart name `aicr-bundle` in Chart.yaml | argocd-helm | Different OCI registry path at publish time |
[...]

PR #1110, opened 2026-05-29
Add opt-in component readiness gates (--readiness-hooks)

### LINKED ISSUE #904: feat: component readiness via chainsaw post-install hook (replaces #610)
## Summary

Establish an **optional, per-component** convention where a component may ship a `readiness.yaml` chainsaw test that serves a dual purpose:

1. **Deploy-time gate** — packaged as a standalone post-install hook chart so the deployer (Helm / Argo CD / Flux) blocks on component-specific readiness signals.
2. **Validation reuse** — the same chainsaw file is consumed by `aicr validate`'s readiness phase.

Replaces #610. **Undeploy is explicitly out of scope** — cleanup is a user/deployer concern; any dev-only cleanup needs live in `tools/`.

## Motivation

Deploy-time readiness for component-specific CRDs (`ClusterPolicy.status.state`, Skyhook completion, DRA kubelet-plugin registration, etc.) is the actual driver behind CUJ2's ~19-minute convergence gap. None of Helm, Argo CD, or Flux know about these CRDs out of the box:

- Helm `--wait` covers standard kinds only.
- Argo CD's built-in health assessment is limited to standard kinds + CRDs with upstream Lua scripts.
- Flux can read `.status.conditions[type=Ready]` directly *when the CRD exposes them* — many of ours don't.

A chainsaw-driven post-install Job fills the gap uniformly across deployers without modifying upstream charts.

## Proposed Design

**Convention:** components that need readiness gating MAY include `recipes/components/<name>/readiness.yaml`, a chainsaw `Test` describing the readiness assertion. Components without one fall back to deployer-native health checks. Not every component needs this.

**Deploy-time packaging:** for each component that has a `readiness.yaml`, the bundler emits a small **standalone post-install hook chart** alongside the component's own chart — no wrapping of the upstream chart required. The hook chart contains only:

- A Job that runs `chainsaw test` against the embedded `readiness.yaml`.
- A ServiceAccount bound to a shared cluster-wide read-only ClusterRole (see RBAC below).
- Hook annotations covering all deployers:
  - `helm.sh/hook: post-install,post-upgrade`
  - `argocd.argoproj.io/hook: PostSync`
  - Flux's helm-controller runs real Helm, so the Helm hook fires
    natively.

**Validator reuse:** `aicr validate`'s readiness phase loads the same per-component `readiness.yaml` files and runs them through the existing chainsaw pipeline. Migration from today's `recipes/checks/<name>/health-check.yaml` is TBD as part of this work (consolidate or alias).

**RBAC:** chainsaw assertions normally only read `.status` fields, so the default is a single **cluster-wide read-only ClusterRole** (get/list/watch on the CRDs and core kinds we assert against). Each component's hook chart ships a ServiceAccount bound to that shared role.

Per-component RBAC is **optional**: a component MAY ship its own ClusterRole/Role + binding alongside its `readiness.yaml` when it needs more than read-only access (e.g., create/exec for probe pods, namespace-scoped permissions, access to resources outside the shared role's allowlist). When present, the hook chart uses the component-supplied role instead of (or in addition to) the shared one.

**Timeouts:** owned by the chainsaw test itself — no bundler-level timeout knobs.

## Out of Scope

- Undeploy / finalizer handling — explicitly dropped (see #610 discussion). Cleanup tooling for dev workflows lives in `tools/`.
- Day-2 reconciliation — chainsaw is a one-shot assertion, not a continuous controller. Deployers handle ongoing health.
- Wrapping upstream charts for this feature — the hook ships as its own small chart.

## Open Questions

- Single canonical file (`readiness.yaml`) vs. keeping `health-check.yaml` as an alias during transition?
- Hook chart image: reuse the AICR validator image (already carries chainsaw) or build a dedicated minimal one?

## References

- Replaces #610

atif1996:
## Summary

Add an opt-in `--readiness-hooks` bundle flag that emits per-component readiness gates so a deploy blocks on component-specific readiness signals (e.g. `gpu-operator`'s `ClusterPolicy` reaching `status.state: ready`), not just on the chart's own resources reporting Ready. Supported on the `helm`, `argocd`, and `argocd-helm` deployers.

## Motivation / Context

Helm `--wait` and Argo CD's native health can't assess operator-specific readiness that lives in a custom resource — an operator Deployment going Available only means the process started, not that its operands rolled out. This wires a Chainsaw-test-driven gate (the `gate` CLI, run as a post-component Job) into the bundle so the deploy converges on the real signal.

Fixes: #904
Related: N/A

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: gate CLI + image (`cmd/gate`, `pkg/chainsawgate/runner`, `.github/workflows/on-tag.yaml`)

## Implementation Notes

- **Opt-in only.** Off by default. A component opts in by shipping `recipes/components/<name>/readiness.yaml` (a Chainsaw `Test`); the bundler discovers it by path — no registry field. `gpu-operator` is the first adopter (asserts `ClusterPolicy` state `ready`).
- **Gate library + CLI migrated into AICR.** `pkg/chainsawgate/runner` (the chainsaw-test evaluation + stability-window/deadline state machine, Kubernetes-API-free) and `cmd/gate` (the standalone polling gate) are ported from the `chainsaw-gate` repo so AICR is the single source of truth.
- **Deployer blocking.**
  - `helm`: `deploy.sh` runs the readiness folder with `helm upgrade --install --wait --wait-for-jobs`; Helm's `--timeout` is **derived** by the bundler from the gate's `--max-wait` + buffer so the gate owns the deadline.
  - `argocd` / `argocd-helm`: the readiness gate is emitted in the next sync-wave after its component and blocks on Argo CD's built-in `batch/Job` health (Progressing → Healthy / Degraded). No custom health Lua.
  - `flux` / `helmfile`: explicitly rejected when `--readiness-hooks` is set.
- **Image.** New `cmd/gate/Dockerfile` builds `ghcr.io/nvidia/aicr-gate`, baking in the Chainsaw version pinned in `.settings.yaml` (so the in-cluster gate runs the same Chainsaw AICR validates with). Wired into `on-tag.yaml` for multi-arch build, manifest, vuln-scan, and SLSA attestation alongside the validator images.
- **CLI-only.** `--readiness-hooks` is not exposed via the bundle API (parity with how the feature was scoped).
- **BOM unaffected.** `container-images.md` is component-scoped and auto-generated; the gate image and `readiness.yaml` don't feed it (same as the validator images), so `make bom-docs` is a no-op here.

## Testing

```bash
make qualify
```

- Green with the pinned toolchain (golangci-lint v2.12.2, helmfile v1.5.2): lint clean, all unit + e2e tests pass, no vulnerabilities.
- New unit coverage: `pkg/bundler/readiness_test.go`, `pkg/chainsawgate/runner/{runner,state}_test.go`, golden + sync-wave-ordering tests for the argocd / argocd-helm deployers, and localformat writer tests.
- `ghcr.io/nvidia/aicr-gate` built and smoke-tested locally: gate compiles, embedded chainsaw reports the pinned `0.2.15`.

## Risk Assessment

**Rollout notes:** Feature is off unless `--readiness-hooks` is passed; components without a `readiness.yaml` are unaffected. The `aicr-gate` image publishes on the next tag. No migration required.

## Checklist


pkg/bundler/readiness.go
```diff
@@ -0,0 +1,223 @@
...
+spec:
+  backoffLimit: 0
+  template:
+    spec:
+      restartPolicy: Never
+      serviceAccountName: %[1]s
+      containers:
+        - name: gate
+          image: %[5]s
+          imagePullPolicy: IfNotPresent
```
njhensley: **🔴 Blocker: the gate Job is never re-run on redeploy and breaks `helm upgrade`.**

This Job has a fixed name (`%[1]s` -> `<component>-readiness-gate`) and no recreation mechanism — no `helm.sh/hook` + `hook-delete-policy: before-hook-creation`, no `ttlSecondsAfterFinished`, no content-hash suffix. Each readiness folder is installed via `helm upgrade --install`, so on a second deploy:

- **same version** -> spec unchanged -> the completed Job is untouched -> `--wait-for-jobs` returns instantly on the already-`Complete` Job. The gate silently passes off the *previous* run.
- **version bump** -> `gateImage()` changes `spec.template.image` -> the apiserver rejects the patch (`Job.spec.template: field is immutable`); `MAX_RETRIES=1` makes it a hard deploy failure. ArgoCD `selfHeal` + ServerSideApply hits the same wall (immutability is enforced post-merge).

The repo already has the fix pattern — see `recipes/components/nodewright-customizations/manifests/no-op.yaml` etc. (`helm.sh/hook: post-install,post-upgrade` + `hook-delete-policy: before-hook-creation`). Please emit the Job as a hook (helm) and an `argocd.argoproj.io/hook` / `Replace=true` sync-option (argocd) so it re-runs every deploy.

pkg/bundler/readiness.go
```diff
@@ -0,0 +1,223 @@
...
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: %[1]s
+  namespace: {{ .Release.Namespace }}
+spec:
+  backoffLimit: 0
+  template:
+    spec:
+      restartPolicy: Never
```
njhensley: **🔴 Blocker-adjacent: a 90-minute, single-attempt, no-restart Job is operationally fragile.**

`--max-wait` defaults to 90 min (`defaults.ReadinessGateMaxWait`). With `backoffLimit: 0` + `restartPolicy: Never`, any node drain / spot reclaim / OOM / eviction in that window fails the Job permanently -> hard-fails the deploy (`MAX_RETRIES=1`). The poll loop is idempotent, so `backoffLimit > 0` (let it reschedule and resume polling) would make the gate survivable on exactly the large, long-rollout clusters this targets.

pkg/bundler/readiness.go
```diff
@@ -0,0 +1,223 @@
...
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["apps"]
+    resources: ["deployments", "daemonsets", "statefulsets", "replicasets"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["batch"]
+    resources: ["jobs", "cronjobs"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["nvidia.com"]
+    resources: ["clusterpolicies"]
+    verbs: ["get", "list", "watch"]
```
njhensley: **🟠 High: a "generic" feature ships a fixed, gpu-operator-shaped ClusterRole.**

The only CRD granted is `nvidia.com/clusterpolicies` (this block). The next adopter asserting a different CRD (network-operator `NicClusterPolicy`, dra-driver, kai-scheduler...) gets no read permission -> chainsaw asserts fail forever -> 90-min `--max-wait` -> deploy fails.

This contradicts `pkg/validator/job/rbac.go:54-72`, which deliberately binds the in-cluster chainsaw runner to `cluster-admin` precisely because the resources "do not exist at compile time." Please either template the CRD grants per-component or follow the validator's precedent — the static middle ground silently breaks adopter #2.

cmd/gate/main.go
```diff
@@ -0,0 +1,232 @@
...
+		if err != nil {
+			slog.Error("evaluate failed", "error", err)
+			return exitConfigErr
+		}
+
+		rs := runner.ComputeReadyState(
+			now, result.AllPass, result.Components, firstPass,
+			opts.StabilityWindow, opts.PollInterval,
+		)
+		firstPass = rs.FirstPassTime
```
njhensley: **🟡 Medium: the stability window can be silently shortened.**

`now := time.Now()` (L117) is captured *before* `runner.Evaluate` (which can run up to the 2-min exec timeout) and then stamped into `firstPass` (L128). Eval latency of intervening iterations is counted as "stable" time, so with `stability-window=30s` / `poll-interval=10s` / `exec-timeout=2m` the "continuously passing for 30s" guarantee can collapse to "passed twice ~10s apart." Capture `now` *after* `Evaluate` returns.

pkg/chainsawgate/runner/runner.go
```diff
@@ -0,0 +1,186 @@
...
+// of combined stdout+stderr. On context timeout, Result is ResultUnknown.
+func RunComponent(ctx context.Context, timeout time.Duration, namespace, compDir string) ComponentResult {
+	tctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	var buf bytes.Buffer
+	// G204: the command is a constant ("chainsaw"); namespace and compDir are
+	// operator-/runner-controlled inputs (a flag value and a temp dir), not
+	// attacker-reachable shell strings.
+	cmd := exec.CommandContext(tctx, "chainsaw", "test", "--no-color", "--namespace", namespace, compDir) //nolint:gosec // see comment above
```
njhensley: **🟡 Medium: chainsaw is exec'd with no pinned `--config`.**

Namespace create/cleanup behavior is left to chainsaw's defaults, and `CHAINSAW_VERSION` is a mutable tag (`cmd/gate/Dockerfile:31`). A future `.settings.yaml` chainsaw bump could change default namespace handling and interact badly with the narrow RBAC in this PR. Pass an explicit config (e.g. skip namespace create/delete) so the gate's contract doesn't drift with the base image.

pkg/bundler/readiness.go
```diff
@@ -0,0 +1,223 @@
...
+metadata:
+  name: %[1]s
+subjects:
+  - kind: ServiceAccount
+    name: %[1]s
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: %[1]s
```
njhensley: **🟡 Medium: cluster-scoped RBAC names collide across multi-bundle deployments.**

This `ClusterRoleBinding` and the `ClusterRole` above are named `<component>-readiness-gate` with no namespace/run-ID qualifier. Two bundles in different namespaces (the `--app-name` multi-tenant story, #1011) both declare `gpu-operator-readiness-gate` -> last-writer-wins, and the binding subject points at one namespace. The validator avoids this with a per-run-ID name (`ServiceAccountName(runID)`). Suffix the cluster-scoped names with the release namespace.

pkg/bundler/readiness.go
```diff
@@ -0,0 +1,223 @@
...
+// collectors so the localformat writer can treat readiness as another
+// auxiliary injection phase). Returns an empty map when --readiness-hooks is
+// off, so callers can forward the result unconditionally.
+//
+// For each component that ships recipes/components/<name>/readiness.yaml, the
+// gate manifests (ServiceAccount, read-only ClusterRole + binding, a ConfigMap
+// carrying the chainsaw Test, and the gate Job) are synthesized with the
+// resolved namespace templated via {{ .Release.Namespace }} — the same
+// mechanism the pre/post manifests use — and the gate image baked in.
+func (b *DefaultBundler) collectComponentReadiness(
```
njhensley, line 72: **🟡 Medium: no bundle-time validation that `readiness.yaml` is a Chainsaw `Test`.**

The file is embedded verbatim; a typo/malformed test surfaces only at deploy time as a 90-min hang then failure. A cheap `kind: Test` / `apiVersion` check here would fail fast at bundle time.

pkg/bundler/readiness.go
```diff
@@ -0,0 +1,223 @@
...
+// readinessManifestKey is the single multi-document manifest path emitted into
+// each readiness gate chart's templates/ directory. A single file keeps the
+// ServiceAccount, RBAC, ConfigMap, and Job together and ordered.
+const readinessManifestKey = "readiness.yaml"
+
+// gateImage returns the fully-qualified gate image reference. The tag tracks
+// the bundler version so a bundle pins the gate image to the AICR release that
+// produced it; an empty/"dev" version resolves to the locally-built dev tag
+// used by the Phase 1 Kind smoke test.
+func (b *DefaultBundler) gateImage() string {
```
njhensley, line 51: **🟢 Low: gate tag = `Config.Version()`.**

Bundles built from a non-release `aicr` reference an unpublished `aicr-gate:<tag>` -> `ImagePullBackOff`, and `imagePullPolicy: IfNotPresent` won't recover. Consistent with the validator images, but here an unpullable image silently blocks the deploy — worth a doc note on the failure mode.

pkg/chainsawgate/runner/runner.go
```diff
@@ -0,0 +1,186 @@
...
+	cmd.Stdout = &buf
+	cmd.Stderr = &buf
+
+	if err := cmd.Run(); err != nil {
+		out := buf.String()
+		if len(out) > maxMsgLen {
+			out = "..." + out[len(out)-maxMsgLen:]
```
njhensley: **🟢 Low: byte-slice truncation can split a UTF-8 rune.**

`"..." + out[len(out)-maxMsgLen:]` slices by bytes; same in `pkg/chainsawgate/runner/state.go:147` (`truncate`) and `cmd/gate/main.go:189` (`shortMsg`). Use a rune-aware trim.

cmd/gate/main.go
```diff
@@ -0,0 +1,232 @@
...
+)
+
+// version is overridden at build time via -ldflags -X.
+var version = "dev"
+
+const (
+	exitOK        = 0
+	exitDeadline  = 1
+	exitConfigErr = 2
+	exitSignal    = 130 // 128 + SIGINT; SIGTERM caller can read $? as 143
```
njhensley: **🟢 Low: comment vs. behavior.**

`signal.NotifyContext` collapses SIGINT/SIGTERM into one cancellation and `run` always returns `130`, but this comment says SIGTERM yields `143`. Fix the comment (or distinguish the signal).

pkg/bundler/deployer/helm/templates/deploy.sh.tmpl
```diff
@@ -304,21 +304,42 @@ for dir in "${SCRIPT_DIR}"/[0-9][0-9][0-9]-*/; do
   # Derive wait args: global --wait/--no-wait behavior + per-component timeout.
   COMPONENT_HELM_TIMEOUT="${HELM_TIMEOUT}"
   COMPONENT_MAX_RETRIES="${MAX_RETRIES}"
+  COMPONENT_WAIT_ARGS=""
   case "${name}" in
     kai-scheduler)
       COMPONENT_HELM_TIMEOUT="20m"
       COMPONENT_MAX_RETRIES="1"
       ;;
+    *-readiness)
```
njhensley, line 313: **🟢 Low: the `*-readiness` glob would also match a real component literally so-named.**

Safe today only because of the writer-side collision check (`pkg/bundler/deployer/localformat/writer.go:302`). Worth a comment noting the dependency.

cmd/gate/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+# CI so the in-cluster gate runs the same chainsaw AICR validates with. The
+# default below is documentation only — keep it in sync with .settings.yaml.
+#
+# GO_VERSION and CHAINSAW_VERSION are declared before the first FROM so both are
+# global build args usable in the FROM lines below (an ARG declared between
+# stages is not visible to the next FROM).
+ARG GO_VERSION=1.26.3
+ARG CHAINSAW_VERSION=v0.2.15
+
+FROM ghcr.io/kyverno/chainsaw:${CHAINSAW_VERSION} AS chainsaw
```
njhensley, line 31: **🟢 Low (partly pre-existing): chainsaw base is pinned by mutable tag, not digest.**

The comment says the binary is "cosign-attested," but copying from a tag (not `@sha256:`) doesn't enforce that at build, and the tag is mutable. Matches existing repo convention (validator/golang/distroless bases are unpinned too), so not introduced here — but it's the one soft spot in the otherwise-complete scan/attest/verify wiring for `GATE_IMAGE`.

njhensley (commented):
Solid, well-structured feature — k8s-free `runner`, a pure/tested state machine, deterministic rendering, complete supply-chain wiring, thorough docs. Two things should gate merge: **(1)** the gate Job isn't re-run on redeploy and breaks `helm upgrade` (silent stale-pass on same version, immutable-field hard fail on version bump — see inline), and **(2)** the 90-min single-attempt no-restart Job is too fragile for the target clusters. **(3)** the static CRD RBAC should be resolved or documented before this is called "generic."

One coverage note: `cmd/gate` has no `main_test.go` (the `run`/`loop`/`progressLine` contract is untested) and there's no e2e/KWOK exercising the readiness path — only the version smoke test in the PR description. The `runner`/`state` packages are well covered. (For the shipped gpu-operator adopter the read-only RBAC is actually exactly sufficient — assert-only, cluster-scoped, existing namespace — so this is a test-coverage gap, not an RBAC-correctness problem.)

pkg/bundler/readiness.go
```diff
@@ -0,0 +1,223 @@
...
+	tag := b.Config.Version()
+	if tag == "" {
+		tag = "dev"
+	}
+	return defaultGateImageRepo + ":" + tag
+}
```
yuanchen8911, line 59: **[Critical] Gate image tag is missing the `v` prefix → ImagePullBackOff on release builds.**

`gateImage()` tags from `Config.Version()` directly. At release, GoReleaser stamps `{{.Version}}` without the leading `v` (e.g. `0.13.0`), but `on-tag.yaml` publishes the gate image as `:${github.ref_name}` (e.g. `v0.13.0`). So a release bundle built with `--readiness-hooks` emits `ghcr.io/nvidia/aicr-gate:0.13.0`, which is never published — the gate Job hits ImagePullBackOff on first install. This is the feature's primary path on a real release.

The validator path already handles exactly this in `pkg/validator/catalog/catalog.go:149-162` (`replaceLatestTag`: *"GoReleaser strips the 'v' from the version but tags keep it"*).

**Recommendation:** normalize the tag so a release version always carries the `v` prefix. Extract the shared normalization if dependency boundaries allow; otherwise implement the same normalization locally in the bundler (note `replaceLatestTag` is unexported in `pkg/validator/catalog`, and importing validator from bundler may be the wrong dependency direction). Add a unit test pinning the release-tag contract (`0.13.0` → `:v0.13.0`).

pkg/bundler/readiness.go
```diff
@@ -0,0 +1,223 @@
...
+            - --stability-window=%[8]s
+            - --max-wait=%[9]s
+          volumeMounts:
+            - name: bundle
+              mountPath: /bundle
+              readOnly: true
+      volumes:
+        - name: bundle
+          configMap:
+            name: %[2]s
```
yuanchen8911: **[Major] Readiness Job is never re-run on redeploy and breaks on upgrade.**

The gate Job is a plain templated resource with a stable name, `backoffLimit: 0`, `restartPolicy: Never`, and no hook or recreate semantics. On a second `helm upgrade` / Argo resync:
- Unchanged spec → the completed/failed Job is left in place and the gate does **not** re-validate readiness.
- Changed spec (gate image bump or arg change) → `spec.template` is immutable, so the upgrade **fails** trying to patch it.

`deploy.sh`'s `cleanup_helm_hooks()` doesn't cover this — it only deletes Jobs carrying the `helm.sh/hook` annotation, which this Job lacks.

**Recommendation:** make the gate Job deterministically recreate on each deploy. A Helm hook with `helm.sh/hook-delete-policy: before-hook-creation` (and the equivalent `argocd.argoproj.io/hook` + delete policy for the Argo path), or a content-hash-suffixed Job name, gives a clean recreate. `ttlSecondsAfterFinished` alone is not sufficient — it is eventual cleanup, not deterministic recreate, so it doesn't fix an immediate redeploy/upgrade. Cover the upgrade/resync case in a test.

pkg/bundler/config/config.go
```diff
@@ -647,6 +663,17 @@ func WithAppName(name string) Option {
 	}
 }
 
+// WithReadinessHooks enables emission of per-component readiness gate charts
+// for components that ship a recipes/components/<name>/readiness.yaml. Off by
+// default; when enabled, each such component gets a standalone
+// NNN-<name>-readiness/ chart that runs the gate CLI as a post-component Job.
+// See #904.
+func WithReadinessHooks(enabled bool) Option {
```
yuanchen8911, line 747: **[Major] New exported functions have 0% coverage.**

`WithReadinessHooks()` and `ReadinessHooks()` (line 405) are new exported functions at 0.0% coverage; `collectComponentReadiness` is at ~16%; `cmd/gate` is at 0.0% (no test files). The deployer golden tests inject `ComponentReadiness` directly, bypassing the collector and the CLI poll loop. This trips the repo's coverage gate (block on any new exported function at 0%).

**Recommendation:** add unit tests for the config option round-trip, for `collectComponentReadiness` (the readiness.yaml-present and absent/`fs.ErrNotExist`-skip paths, plus the deployer-rejection branch), and for the `cmd/gate` `loop`/`progressLine`/`summaryReason` helpers (documented as package-private for testability).

pkg/bundler/deployer/helm/templates/deploy.sh.tmpl
yuanchen8911, line 313: **[Minor] `*-readiness)` glob over-matches release names.**

This branch matches any release whose name ends in `-readiness`, not only synthesized gate folders. A component literally named `foo-readiness` (shipping no gate) would silently inherit gate wait semantics (`--wait-for-jobs`, ~95-min timeout). No such component exists today, so this is latent robustness.

**Recommendation:** tighten the match (e.g. drive it off the known set of emitted readiness folders) or document the reserved `-readiness` suffix alongside the existing `-pre`/`-post` collision guard.

cmd/gate/main.go
yuanchen8911: **[Minor] Exit-code comment is inaccurate.**

The comment states SIGTERM yields `143`, but the shared return path (line 152) returns the constant `130` for both SIGINT and SIGTERM on `ctx.Done()`. Cosmetic.

**Recommendation:** fix the comment (or re-raise the signal if 143 on SIGTERM is actually desired).

yuanchen8911 (changes requested):
Reviewed the `--readiness-hooks` feature. Solid design overall — fail-closed flux/helmfile rejection and the derived helm `--timeout` are nice. Two blockers and a coverage-gate gap to resolve before merge; details inline.

- **Critical:** gate image tag missing `v` prefix → ImagePullBackOff on release builds.
- **Major:** readiness Job never re-runs on redeploy and breaks `helm upgrade` on image/arg change.
- **Major:** new exported functions at 0% coverage (`ReadinessHooks`, `WithReadinessHooks`, `cmd/gate`).
- **Minor:** `*-readiness)` glob over-matches release names; inaccurate exit-code comment.

A few open questions (gate RBAC scope, chainsaw assert-on-empty semantics, elapsed-time accounting) are noted inline as well.

yuanchen8911 (changes requested):
A few issues.

atif1996:
Follow-up commit b6a1effd addresses review items #1–#6 and #9:

- **#1**: `gateImage()` now normalizes release tags (`0.13.0` → `v0.13.0`); contract tests added.
- **#2**: Gate Job is deployer-aware — helm post-install/post-upgrade hook + `before-hook-creation`; Argo CD `Replace=true` on the Job. Helm `deploy.sh` readiness branch no longer uses `--wait-for-jobs` (hooks are waited on by `--wait`).
- **#3**: Scoped RBAC kept (no Secrets): `nvidia.com/*` + `apiextensions.k8s.io/customresourcedefinitions` read; core/apps/batch unchanged.
- **#4/#6**: Added config round-trip, `collectComponentReadiness`, `gateImage`, and synctest-based `cmd/gate` loop tests.
- **#5**: `ReadinessGateBackoffLimit=6` templated into the Job; `restartPolicy: Never` retained.
- **#6**: `now := time.Now()` moved to after `runner.Evaluate` in the gate poll loop.
- **#9**: Bundle-time validation that `readiness.yaml` is a chainsaw `kind: Test`.

Manifest rendering extracted to `pkg/bundler/gatemanifest` so deployer golden tests can import without an import cycle. Goldens regenerated; `make qualify` green.

**Deferred (unchanged):** #7 chainsaw `--config`, #8 multi-bundle RBAC name collision (#1011), #10–#15 low/minor.

**Scope note:** deploy-time gate remains one-shot by design — unchanged manifests do not re-sync on cluster scale-up; ongoing convergence is operator reconciliation + monitoring.

pkg/bundler/gatemanifest/manifest.go
```diff
@@ -0,0 +1,163 @@
...
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["batch"]
+    resources: ["jobs", "cronjobs"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["nvidia.com"]
+    resources: ["*"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["apiextensions.k8s.io"]
+    resources: ["customresourcedefinitions"]
+    verbs: ["get", "list", "watch"]
```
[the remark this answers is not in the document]
atif1996: Acknowledged, and intentional. We've scoped the grant to what our shipped readiness tests actually read (nvidia.com CRDs + core/apps) to keep the gate least-privileged. When we add tests that assert CRDs in other apiGroups, we'll expand the ClusterRole alongside them.

pkg/chainsawgate/runner/runner.go
atif1996: Fixed in 42bbf99. The gate image now bakes a pinned chainsaw `Configuration` (`cleanup.skipDelete: true`) at `/etc/chainsaw/config.yaml`, and `gate` passes it via `--config` (new `--chainsaw-config` flag, default points at the in-image path; runner appends `--config <path>` only when non-empty so local/CLI runs without a baked-in config still work). `skipDelete` matches the read-only ClusterRole (no delete verbs), so chainsaw never attempts to delete the release namespace or any resource regardless of base-image default drift across chainsaw version bumps.

pkg/bundler/gatemanifest/manifest.go
```diff
@@ -0,0 +1,163 @@
...
+  name: %[1]s
+subjects:
+  - kind: ServiceAccount
+    name: %[1]s
+    namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: %[1]s
+---
```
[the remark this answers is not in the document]
atif1996: Fixed in 42bbf99. The cluster-scoped `ClusterRole`, `ClusterRoleBinding`, and `roleRef` names are now namespace-qualified (suffixed with the resolved `{{ .Release.Namespace }}`), e.g. `gpu-operator-readiness-gate-gpu-operator`, so two bundles deploying the same component to different namespaces (the multi-tenant #1011 case) no longer overwrite each other's cluster-scoped objects. The namespaced ServiceAccount, binding subject, ConfigMap, and Job keep the bare component name. Deployer goldens regenerated.

pkg/bundler/readiness.go
atif1996: Fixed (52721eb3). Recreation is now deployer-aware: Helm-deployed gates carry `helm.sh/hook: post-install,post-upgrade` + `helm.sh/hook-delete-policy: before-hook-creation`, and ArgoCD gates carry `argocd.argoproj.io/sync-options: Replace=true`. The Job is recreated on every (re)deploy instead of colliding with the prior immutable Job. See `jobMetadataAnnotations` in `pkg/bundler/gatemanifest/manifest.go`.

pkg/bundler/readiness.go
atif1996: Addressed (52721eb3). The Job now uses `backoffLimit: 6` (`defaults.ReadinessGateBackoffLimit`) instead of `0`, so a node drain / spot reclaim retries the pod rather than failing the whole deploy on a single eviction.

pkg/bundler/readiness.go
atif1996: Addressed (52721eb3). The ClusterRole is no longer clusterpolicies-only: it now grants get/list/watch on `nvidia.com/*` (all CRD instances in that group), `apiextensions.k8s.io/customresourcedefinitions`, plus core/apps/batch read. Kept read-only (no secrets, no write verbs); apiGroups can be widened as new components add their own chainsaw tests.

cmd/gate/main.go
atif1996: Fixed (52721eb3). `now` is now captured *after* `runner.Evaluate` returns, so eval latency (up to the exec timeout) no longer eats into the stability window stamped at `firstPass`. See the loop in `cmd/gate/main.go`.

pkg/bundler/readiness.go
atif1996: Fixed (42bbf99). The cluster-scoped `ClusterRole`/`ClusterRoleBinding`/`roleRef` names are namespace-qualified (suffixed with the resolved namespace), e.g. `gpu-operator-readiness-gate-gpu-operator`, so same-component bundles in different namespaces no longer overwrite each other. The namespaced ServiceAccount, binding subject, ConfigMap, and Job keep the bare component name.

pkg/bundler/readiness.go
atif1996: Fixed (52721eb3). `validateReadinessTestYAML` now rejects at bundle time anything whose `apiVersion` is not `chainsaw.kyverno.io/*` or whose `kind` is not `Test` (returns `ErrCodeInvalidRequest`), so a malformed readiness.yaml fails the bundle instead of hanging the deploy for the full max-wait.

pkg/bundler/readiness.go
atif1996: Addressed (52721eb3). `gateImage()` now normalizes a release version missing the leading `v` (`0.13.0` -> `v0.13.0`) to match the GoReleaser-published tag; empty/`dev` still resolves to `:dev` for the Phase-1 kind smoke path.

pkg/bundler/readiness.go
atif1996: Fixed (52721eb3). `gateImage()` now prepends the `v` when a release version lacks it (`0.13.0` -> `v0.13.0`) so the tag matches the GoReleaser-published `aicr-gate` image; empty/`dev` still maps to `:dev`. No more ImagePullBackOff on release builds.

pkg/bundler/readiness.go
atif1996: Fixed (52721eb3). Recreation is deployer-aware now: Helm gates use `helm.sh/hook: post-install,post-upgrade` + `hook-delete-policy: before-hook-creation`; ArgoCD gates use `sync-options: Replace=true`. The Job re-runs on every redeploy/upgrade instead of colliding with the prior immutable Job.

pkg/bundler/config/config.go
atif1996: Addressed (52721eb3 + 42bbf99). Coverage now: `WithReadinessHooks` 100%, `ReadinessHooks` 100%, `gateImage` 100%, `validateReadinessTestYAML` 100%, `collectComponentReadiness` ~76% (was ~16%), and `cmd/gate` has a test suite at ~76% (was 0%, no test files).

docs/integrator/recipe-development.md
```diff
@@ -375,6 +375,36 @@ componentRefs:
...
+      try:
+        - assert:
+            resource:
+              apiVersion: nvidia.com/v1
+              kind: ClusterPolicy
+              status:
+                state: ready
+```
+
+When `--readiness-hooks` is set, the bundler wraps this test into a `NNN-<name>-readiness/` folder containing a `Job` that runs the `gate` CLI (`ghcr.io/nvidia/aicr-gate`, which embeds Chainsaw). The deploy blocks on that Job — via `helm --wait-for-jobs` for the helm deployer, or via Argo CD's built-in `batch/Job` health on the next sync-wave for the `argocd`/`argocd-helm` deployers. Keep `spec.timeouts.assert` shorter than the gate's per-test timeout so a single poll can't outlast one gate iteration. See [Readiness Gates](../user/cli-reference.md#readiness-gates) for the deploy-time behavior.
```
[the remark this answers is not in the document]
atif1996: Fixed (739ed84). Reworded to `helm --wait`: the gate Job is a `post-install,post-upgrade` hook, and `--wait` blocks on hook completion regardless of `--wait-for-jobs`, matching the generated `deploy.sh` (which uses `--wait --timeout`).

docs/user/cli-reference.md
```diff
@@ -1392,6 +1393,25 @@ HELM_REPOSITORY_PASSWORD=secret \
...
+
+With the flag set, the bundler emits an extra folder, `NNN-<name>-readiness/`, immediately after each opted-in component. The folder is a small chart containing a Kubernetes `Job` (plus the ServiceAccount/RBAC and a ConfigMap holding the Chainsaw test). The Job runs the `gate` CLI (`ghcr.io/nvidia/aicr-gate`, which embeds Chainsaw), which polls the test until it passes continuously for a stability window or a `--max-wait` ceiling elapses. The deploy blocks on that Job:
+
+- **`helm`** — `deploy.sh` runs the readiness folder with `helm upgrade --install --wait --wait-for-jobs`. Helm's own `--timeout` is derived by the bundler from the gate's `--max-wait` plus a buffer, so the gate owns the deadline (Helm never preempts it).
+- **`argocd` / `argocd-helm`** — the readiness folder inherits the next sync-wave after its component, and Argo CD blocks that wave on the gate Job via its built-in `batch/Job` health (Progressing → Healthy on success, Degraded on failure). No custom health Lua and no direct `ClusterPolicy` watch — the readiness logic stays encapsulated in the Chainsaw test the Job runs.
```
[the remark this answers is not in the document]
atif1996: Fixed (739ed84). Removed `--wait-for-jobs` from the helm line; the readiness folder runs `helm upgrade --install --wait`, and the gate hook completes under plain `--wait`.

cmd/gate/main.go
atif1996: Fixed (739ed84). Corrected the `exitSignal` comment: `signal.NotifyContext` collapses SIGINT/SIGTERM into one cancellation and `loop` always returns `130`, so the process exits `130` for either signal (not `143`).

atif1996: Fixed (739ed84). The `exitSignal` comment now states the process exits `130` for both SIGINT and SIGTERM (the shared `ctx.Done()` path), instead of the inaccurate `143`-for-SIGTERM note.

pkg/chainsawgate/runner/runner.go
atif1996: Fixed (feadebd). Added rune-aware `TruncHead`/`TruncTail` helpers that back off to a UTF-8 rune boundary before trimming, and routed all three sites through them: `RunComponent` failure tail (`runner.go`), `truncate` (`state.go`), and `shortMsg` (`cmd/gate/main.go`). Added table tests asserting valid UTF-8 output on multi-byte input.

pkg/chainsawgate/runner/runner.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		if wErr := os.WriteFile(filepath.Join(compDir, "chainsaw-test.yaml"), []byte(testYAML), 0o600); wErr != nil {
+			return EvalResult{}, errors.Wrap(errors.ErrCodeInternal, "write test file for "+comp, wErr)
+		}
+		res := runComponentFn(ctx, opts.Timeout, opts.Namespace, compDir)
+		components[comp] = res
+		if res.Result != ResultPass {
+			allPass = false
+		}
+	}
```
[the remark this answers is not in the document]
atif1996: Fixed (feadebd). `Evaluate` now checks `ctx.Err()` at the top of the per-component loop and returns early (wrapped `ErrCodeTimeout`) instead of spawning more chainsaw execs. The gate loop maps that cancellation to exit 130 (interrupt) rather than exit 2. Covered by `TestEvaluate_HonorsContextCancellation`.

pkg/chainsawgate/runner/runner.go
```diff
@@ -0,0 +1,186 @@
...
+		// G304: name comes from ReadDir over the operator-supplied bundle dir
+		// (a ConfigMap mount), constrained to *.yaml entries in that dir.
+		data, rErr := os.ReadFile(filepath.Join(dir, name)) //nolint:gosec // see comment above
+		if rErr != nil {
+			return nil, errors.Wrap(errors.ErrCodeInternal, "read "+name, rErr)
+		}
+		out[name] = string(data)
+	}
+	return out, nil
+}
```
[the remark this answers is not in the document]
atif1996: The per-component evaluation loop now honors `ctx.Err()` (feadebd), so context cancellation is propagated promptly rather than ignored. Resolving.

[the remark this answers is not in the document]
atif1996: Correction to my previous reply: this thread is about `LoadBundleDir` taking a `context`, not the eval loop. Keeping it contextless is intentional — `LoadBundleDir` does a single, bounded read of a local directory (the gate's mounted bundle ConfigMap) once at startup, with no network or unbounded I/O, so there is nothing for a context to cancel. The context-bearing path that can actually block (`Evaluate` -> `RunComponent` -> `chainsaw` exec) already threads `ctx` and now also checks `ctx.Err()` between components (feadebd). Leaving resolved on that basis.

lockwobr:
Need to review more yet, but I am liking it so far. I feel like might want to invert the flag tbh, `--no-readiness` to opt-out instead, feels like something we want!

pkg/bundler/deployer/helm/templates/deploy.sh.tmpl
atif1996: Fixed (6351f8c). Rather than loosen the glob, the `-readiness` suffix is now reserved at bundle time: `localformat.Write` rejects any declared component whose name ends in `-readiness` (`ErrCodeInvalidRequest`). Since `-pre`/`-post` folders end in those suffixes and the only `*-readiness` folders are synthesized gates, the deploy.sh `*-readiness)` match is now exact-by-construction. Note this also covers the `--readiness-hooks` off case, where the static glob would otherwise still match. Covered by `TestWrite_ReadinessSuffixReserved`.

atif1996: Fixed (6351f8c). The dependency you flagged is now explicit and enforced: `localformat.Write` reserves the `-readiness` suffix (rejects any declared component named `*-readiness`), with a comment pointing back at the deploy.sh `*-readiness)` glob. The glob can only ever match synthesized gate folders now. Covered by `TestWrite_ReadinessSuffixReserved`.

.github/workflows/on-tag.yaml
```diff
@@ -198,14 +202,78 @@ jobs:
...
+  # =============================================================================
+
+  build-gate:
+    name: Docker Gate (${{ matrix.arch }})
+    runs-on: ${{ matrix.runner }}
+    needs: [tests]
+    timeout-minutes: 15
+    permissions:
+      contents: read
+      packages: write
```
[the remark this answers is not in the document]
atif1996: Fixed (a23f6fb). Added a comment above the `build-gate` permissions block explaining the grants: `contents: read` for checkout, `packages: write` to push the per-arch aicr-gate image to GHCR.

yuanchen8911:
need rebase, LGTM, no blocking issues.

pkg/bundler 77.8% → 77.6% (−0.2%, sub-threshold). TestFetchAmbientOIDCToken fails on both base and PR (pre-existing, not a regression).

mchmarny (approved):

njhensley (approved):

atif1996:
> need rebase, LGTM, no blocking issues.
> 
> pkg/bundler 77.8% → 77.6% (−0.2%, sub-threshold). TestFetchAmbientOIDCToken fails on both base and PR (pre-existing, not a regression).

Rebased.  Should be good to go

yuanchen8911 (approved):
/lgtm

PR #1132, opened 2026-06-01
docs: ADR-009 — recipe health tracking and communication

njhensley:
## Summary

Adds ADR-009, a design-only proposal for tracking and communicating recipe
health across every supported criteria combination: an enumeration command
(`aicr recipe list`) plus a per-recipe structural-health matrix, generated
and published with the same advisory bot-PR machinery the BOM already uses.

## Motivation / Context

AICR ships ~50 leaf recipes across criteria combinations, but there is no
durable record of their health and no way to communicate it — not even a
command to enumerate the resolvable catalog. ADR-007 answers "did *this run*
pass, verifiably?" for one recipe; this ADR is its catalog-wide complement:
"across the whole matrix, what is each recipe's current structural state?"
It also lays the frame for incorporating ADR-007 evidence as it lands.

Fixes: N/A
Related: ADR-007 (Verifiable Recipe Test Evidence), ADR-006 (Image Pinning)

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Docs/examples (`docs/`)

## Implementation Notes

Design-only — no code ships in this PR; implementation will be tracked under
a follow-on epic if the ADR is accepted.

Key decisions:
- **Two orthogonal axes, never collapsed** — *structural soundness* (graded
  A–F) and *validation posture* (evidence-derived). Structural signals must
  never read as "validated"; this is the anti-false-confidence backbone.
- **V1 ships the structural axis + enumeration + one public matrix only.**
  Every V1 signal is hermetic and offline (no GPU, no network), so the
  generated doc is reproducible from a checkout.
- **`declared_coverage` descriptor** records *which named checks* each
  validation phase declares (not just whether the phase exists) — answering
  "which validations are defined for this recipe?" It is informational and
  does not move the grade.
- **Scope was deliberately narrowed after a multi-perspective design review**
  (architecture, supply-chain security, DX, CI/SRE, YAGNI). The
  evidence-driven validation axis, a committed JSON artifact, a daily cron,
  the PR-comment apparatus, and the KWOK / live-chart-drift signals are
  deferred — each with a concrete pull-trigger in *What V1 does not ship* —
  because they could not satisfy the ADR's own hermeticity/trust constraints
  while `recipes/evidence/` is empty and KWOK results are ephemeral.

## Testing

Documentation-only change; no Go sources touched. The relevant CI is
`fern-docs-ci.yaml` (lychee link check + MDX safety on `docs/**`); the ADR's
internal cross-links to `006-image-pinning-policy.md` and
`007-recipe-evidence.md` resolve.

## Risk Assessment

- [x] **Low** — Additive, design-only document; trivially revertible.

**Rollout notes:** N/A — the ADR proposes future work; nothing is implemented
or wired into any gate by this PR.

## Checklist

- [ ] Tests pass locally (`make test` with `-race`) — N/A (docs-only)
- [ ] Linter passes (`make lint`) — N/A (docs-only)
- [ ] I added/updated tests for new functionality — N/A
- [x] I updated docs if user-facing behavior changed (this PR *is* the doc; no behavior change)
- [x] Changes follow existing patterns in the codebase (matches ADR-006/007/008 house style)


docs/design/009-recipe-health-tracking.md
```diff
@@ -0,0 +1,459 @@
...
+recipe (e.g. the Slurm leaf that drops K8s checks) must not be penalized for
+declaring fewer checks.
+
+Per-dimension state (graded dimensions): `pass | warn | fail |
+not-applicable | unknown`. Grade: `A` all `pass`; `B` one `warn`; `C` ≥2
+`warn` or a non-blocking `not-applicable` gap; `D` one `fail` outside
+`resolves`; `F` `resolves: fail`. Transient resolver errors
+(`ErrCodeTimeout`/`ErrCodeInternal`) yield `unknown` and are **held, not
+graded** — re-run rather than penalize. `unknown` is a representable value in
+the serialized grade, never an empty field a consumer could misread as `A`.
```
mchmarny: The A–F vocabulary on a surface labeled "health" will read as a quality verdict beyond what v1 actually computes (resolves, constraints well-formed, chart pinned). The Evidence column + prose disclaimer mitigate, but a reader skimming the table sees "A" and infers "validated and good to deploy."

Concrete risk: a recipe with a `B` (one `warn` on `constraints_wellformed`) is structurally less sound than an `A` *and* potentially less validated — but a reader cannot tell which. The grade overloads two questions a user will (correctly) read as one.

Consider deferring letter grades until later, when the validation axis can better fuse into them. For v1, I'd recommend we surface `pass | warn | fail` per dimension and a single rolled-up `pass | warn | fail` column. Less catchy on a matrix, but it does not promise more than the signals justify. The tri-state vocabulary §3 reserves for v1.1 (Supported/Preview/Experimental) is exactly the same argument applied to a different label — apply it to the grade too.

docs/design/009-recipe-health-tracking.md
```diff
@@ -0,0 +1,459 @@
...
+
+Each combination is scored from signals that are **all hermetic and offline**
+(reproducible from a checkout, no GPU, no network). Three signals are
+**graded** (they move the `A`–`F` grade); one is a captured **descriptor**
+(surfaced, never graded):
+
+| Dimension | Role | Signal | Source |
+|---|---|---|---|
+| `resolves` | graded | `BuildFromCriteria` returns a `RecipeResult` without error | `pkg/recipe/builder.go` |
+| `constraints_wellformed` | graded | Every `Constraint.Value` parses; inline replay under `--no-cluster` is clean; no `ConstraintWarning`s | `pkg/recipe/metadata.go`, `pkg/validator` (no-cluster mode) |
```
mchmarny, line 176: Compute cost not bounded. `--no-cluster` validator replay across ~50 combos per weekly run is almost certainly fine, but the ADR sets no budget — a future maintainer reading this won't know whether 30 s, 5 min, or 30 min is the expected ceiling, or when to start worrying. One sentence on the target runtime (and what would trigger a redesign — caching, parallel replay, dropping the inline replay) keeps the door closed on this growing into a multi-minute job by accretion.

I'd also sync with @iamkhaledh on this.

docs/design/009-recipe-health-tracking.md
```diff
@@ -0,0 +1,459 @@
...
+`tools/health` generator calls `pkg/health.Compute` directly and needs no
+user-facing verb. Pull-trigger: a user asks to compute health for one
+ad-hoc combo outside the generated doc.
+
+**Business logic — new `pkg/health` package.** Not `pkg/recipe`:
+`pkg/evidence/attestation` already imports `pkg/recipe`, so placing health
+in `pkg/recipe` and importing evidence (needed by the deferred axis) would
+close a `recipe → evidence → recipe` cycle. A standalone `pkg/health` that
+depends on both is acyclic. V1 dependencies: `pkg/recipe`, `pkg/serializer`,
+`pkg/errors`, `pkg/defaults`. Sketch:
```
mchmarny: The cycle argument is forward-looking only. v1 has zero dependency on `pkg/evidence` — `pkg/recipe/health` (or `pkg/recipe.Health`) would compile cleanly today. The cycle only materializes in v1.1 once the validation axis imports evidence.

Not necessarily an issue — designing the package boundary now to avoid a v1.1 move is reasonable. But the ADR should say that's what it's doing ("placed in `pkg/health` to keep the v1.1 evidence import acyclic") rather than imply v1 itself requires the separation. The current wording reads as if the cycle exists today, which it doesn't.

mchmarny (commented):
Overall this makes sense. My inline notes are not blocking but may be worth sorting out before approval.

docs/design/009-recipe-health-tracking.md
njhensley: Agreed. Dropped the A–F letter grade for v1 in favor of a rolled-up `pass | warn | fail` status (per-dimension `pass | warn | fail | not-applicable | unknown` underneath, held as `unknown` on transient resolver errors). You're right that one letter overloads "how structurally sound?" and "how validated?" — the doc now defers letter grades to the *What V1 does not ship* table, to be reconsidered once the validation axis can fuse into a single graded verdict. Same reasoning §3 already applies to the Supported/Preview/Experimental vocabulary, now applied to the grade too.

docs/design/009-recipe-health-tracking.md
njhensley: Added a **Compute budget** note to §2. The dominant cost is the `--no-cluster` constraint replay across the ~50 leaf combos — pure in-process (no cluster, no GPU), so it's CI-runner wall-clock only. Target is a sub-minute generator run; if it grows past a few minutes the redesign levers, in order, are: cache per-combo results keyed by the resolved-recipe digest, parallelize the replay, or drop inline replay from the scheduled job and gate it to PRs touching `recipes/**`.

docs/design/009-recipe-health-tracking.md
njhensley: Good catch — reworded. The ADR now states the standalone `pkg/health` is a forward-looking boundary choice, not a v1 requirement: v1 has zero `pkg/evidence` dependency and would compile cleanly in `pkg/recipe` today. The separation exists only to keep the v1.1 evidence import acyclic (`pkg/evidence/attestation` already imports `pkg/recipe`, so health in `pkg/recipe` would close `recipe → evidence → recipe` once the validation axis lands) and to avoid a v1.1 package move.

mchmarny (approved):

PR #1133, opened 2026-06-01
feat(validators): enhance inference performance validation

### LINKED ISSUE #1116: Improve inference performance validation: tunable, per-accelerator params wired into recipes
## Summary

The `inference-perf` check defaulted to a tiny smoke-test model (Qwen/Qwen3-0.6B) with only global tuning knobs, so it couldn't characterize modern accelerators or adapt per GPU type. It is now **configurable per recipe**, backed by a **model-weights cache**, with **per-accelerator thresholds re-measured on real hardware** — so `aicr validate --phase performance` reflects GPU compute, not serving overhead, and works out of the box.

Implemented on branch `feat/inference-perf-configurable-model` (stacked on merged PR #1096); validated end to end on RTX PRO 6000, H100 (EKS + GKE), and GB200.

### 1. Per-recipe configuration

`inference-model` and `inference-concurrency-per-gpu` are now read from the recipe's `validation.performance.constraints` — symmetric with the `inference-throughput` / `inference-ttft-p99` thresholds that already live there. Each `<accel>-<service>-…-inference-dynamo` overlay pins its own values. Resolution precedence is **recipe constraint > catalog `env` (`AICR_INFERENCE_PERF_*`) > compiled default**. New defaults: **Qwen/Qwen3-8B at 256 concurrency/GPU** (the empirical throughput sweet spot across all tested accelerators).

### 2. Model-weights cache

A PVC-backed cache is **on by default**: a one-time populate Job downloads the model once, and all workers mount it read-only (`HF_HUB_OFFLINE=1`). This removes the per-worker Hugging Face download that tripped per-IP HTTP 429 throttling on large models, and needs no token for ungated models. Disable with `AICR_INFERENCE_PERF_MODEL_CACHE_SIZE=off`. The cache **fails fast** with actionable guidance when no StorageClass is resolvable (instead of leaving the PVC `Pending` until timeout); on EKS the `aws-ebs-csi-driver` component now provisions a default `gp3` StorageClass so it works zero-config (GKE already has `standard-rwo`).

### 3. Updated / improved thresholds

The per-accelerator gates (previously placeholders tuned to 0.6B at conc 16: `>= 5000` / `<= 200`) were re-measured at 8B / 256 and pinned alongside the model + concurrency in each overlay:

| Accelerator (overlay) | throughput gate | TTFT p99 gate | measured (8B/256) |
|---|---|---|---|
| RTX PRO 6000 (eks) | `>= 50000` | `<= 1000` | 59,260 tok/s · 445 ms · 100% util |
| H100 (eks + gke) | `>= 50000` | `<= 1000` | ~108k tok/s · ~700 ms · 100% util |
| GB200 (eks) | `>= 50000` | `<= 2000` | 65,952 tok/s · 1,240 ms · 80% util |
| B200 (gke) | `>= 50000` | `<= 2000` | untested — mirrors GB200 (flagged in-file) |

The evaluator applies a 10% tolerance on top of these.

## Remaining work

1. **Precision** (FP8 / NVFP4, typically via the checkpoint ID) — not yet a knob.
2. **Tensor-parallel size (TP > 1)** — the Dynamo deploy template still hardcodes TP=1 / one worker per GPU. Needed for models that exceed one GPU (e.g. 32B BF16 on an 80 GB H100) and to fully load a Blackwell GPU on GB200.

## Motivation (from 8B characterization)

At the original 0.6B size the benchmark was serving-overhead-bound and never loaded the GPU. Defaulting to **8B at 256 concurrency/GPU** fixes that on Hopper/Ada — H100 and RTX PRO 6000 reach **~100% GPU util** at their throughput peak. **GB200 reaches only ~80%** on 8B (it is over-provisioned for a model this small), which is what the remaining **precision + TP / larger-model** work is for. TTFT SLO context: interactive ~200 ms, voice ~500 ms, MLPerf server 2 s (70B) – 6 s (405B).

## Related

- PR #1096 (warmup + initial tunable knobs) — merged.
- Branch `feat/inference-perf-configurable-model` — the changes above; cache-on-by-default reproduces the hand-tuned sweep numbers within <1%.

yuanchen8911:
## Summary

Makes the `inference-perf` performance check configurable per recipe, defaults it
to a real-workload model, and backs it with a model-weights cache — so
`aicr validate --phase performance` reflects GPU compute (not serving overhead)
and works out of the box. Three headline changes: **per-recipe model/concurrency**,
**model cache on by default**, and **per-accelerator gate thresholds re-measured on
real hardware**.

> **Draft / WIP** — opening for early review. Full `make qualify` (e2e + scan) is
> pending before marking ready; `make test` (-race) and `make lint` pass locally.

## Motivation / Context

Fixes #1116.

The old check ran a tiny smoke-test model (Qwen3-0.6B, conc 16) that never loaded
the GPU and used global-only knobs. Characterization across RTX PRO 6000, H100
(EKS + GKE), and GB200 showed an 8B model at conc 256/GPU is the throughput sweet
spot and that the per-accelerator thresholds needed real numbers.

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe`, recipes/overlays, recipes/validators/catalog.yaml)
- [x] Validator (`pkg/validator`, `validators/performance`)
- [x] Docs/examples (`docs/`)

## Implementation Notes

- **Per-recipe inputs:** `inference-model` and `inference-concurrency-per-gpu` read
  from `validation.performance.constraints`, resolving **recipe > catalog env > default**
  (symmetric with the throughput/TTFT thresholds). Defaults: Qwen/Qwen3-8B at 256/GPU.
- **Model-weights cache (PVC) on by default:** one populate download shared by all
  workers (offline), avoiding per-IP HF 429s; `…MODEL_CACHE_SIZE=off` to disable.
  Fails fast with guidance when no StorageClass is resolvable instead of hanging.
- **EKS default StorageClass:** `aws-ebs-csi-driver` provisions a default gp3 SC
  (`defaultStorageClass.enabled`); GKE already has `standard-rwo`.
- **Per-accelerator gates** in the `*-inference-dynamo` overlays re-measured at
  8B/256, with model + concurrency pinned alongside.
- **Configurable timeouts:** workload-ready / endpoint-readiness / health timeouts
  (`AICR_INFERENCE_PERF_WORKLOAD_READY_TIMEOUT`, `…HEALTH_TIMEOUT`) for large-model
  bring-up. The inference-perf catalog timeout budget now accounts for the
  cache-populate phase (50m → 65m), and the facade `ValidationOperationTimeout` was
  raised to 75m (> the largest catalog check timeout) with a catalog-vs-facade
  guard test.
- **Security/scoping (Codex review):** model passed to AIPerf via the `AICR_MODEL`
  env referenced as `"$AICR_MODEL"` (no shell command-substitution); `HF_TOKEN`
  forwarded only to the inference-perf validator.
- **Input hardening (review rounds):**
  - Resolved model ID validated against the HF repo-ID character set — rejects
    YAML/shell metacharacters before it reaches the Dynamo template.
  - `SERVED_MODEL_NAME: "${MODEL}"` quoted in the deploy template so a
    scalar-looking model ID (pure-numeric / boolean-like) parses as a string, not
    int/bool/null (+ parse test).
  - Populate container carries explicit CPU/memory **requests**
    (`100m`/`512Mi`) so it schedules under ResourceQuota / LimitRange — but
    **no memory limit**: it streams tens of GB to the PVC and a hard limit
    OOMKills large-model downloads via page cache on cgroup v2 (caught on a
    live 8B GKE run).
  - Scale the `inference-throughput` gate to the free-GPU count actually
    benchmarked. The workload is sized to a node's *free* GPUs, so on a shared
    node the absolute full-node gate would false-fail a healthy per-GPU result;
    the gate is scaled by `freeGPUs / nodeGPUs` (TTFT p99, a latency, is not
    scaled). Full-node runs are a no-op.
  - Preserve the model-cache populate Job's timeout/unavailable error
    classification (`PropagateOrWrap` instead of flattening to `ErrCodeInternal`).
  - Propagate the validator pod's `imagePullSecrets` to the model-cache populate
    Job (parity with the AIPerf Job) so private-mirror / air-gapped clusters can
    pull the cache worker image.
  - Reject non-positive cache sizes (`0Gi` / `-1Gi`) with `ErrCodeInvalidRequest`.
  - Clear a stale HF-token Secret when the token is unset; carry `resourceVersion`
    on Secret update; bound the cache PVC/Job and HF-token Secret API calls with
    named timeouts.
  - Recognize the legacy beta default-StorageClass annotation.

## Testing

Commands run:
- `make test` (-race) — pass; total coverage 76.4% (≥ 70% floor).
- `make lint` (golangci-lint + yamllint + license/agents-sync) — 0 issues.

Validated end to end on live clusters with the **compiled defaults** (Qwen3-8B,
256 concurrency/GPU, cache on; no tuning flags beyond the per-region
registry/tag, plus `MODEL_CACHE_STORAGE_CLASS=gp2` on GB200 which has no default
StorageClass):

| Cluster | GPUs | throughput | TTFT p99 | peak GPU util |
|---------|-----:|-----------:|---------:|--------------:|
| EKS RTX PRO 6000 | 8 | 58,434 tok/s | 458 ms | 100% |
| GKE H100 | 8 | 107,742 tok/s | 796 ms | 100% |
| EKS H100 (aicr2, earlier) | 8 | ~108,790 tok/s | ~688 ms | 100% |
| EKS GB200 | 4 | 69,072 tok/s | 989 ms | ~80% |

All passed their per-accelerator gates. GB200 reaches only ~80% util on 8B (it is
over-provisioned for a model this small) — the documented case for the remaining
precision/TP work. The on-by-default cache bound to each cluster's StorageClass
(`standard-rwo` on GKE, `ebs-csi-default-sc` on the AICR-provisioned EKS, `gp2`
on GB200), and the no-default-SC fast-fail was confirmed on GB200 (~6 s, with
actionable guidance).

Re-verified the head-of-branch build end to end on GKE H100 (validator image
rebuilt from the latest commit): per-recipe model/concurrency resolved from the
recipe, cache on by default bound zero-config to the cluster's `standard-rwo`
default SC (100Gi RWO `aicr-model-cache`), and the populate container scheduled
with its new `100m`/`512Mi` requests + `4Gi` limit.

- **Pending before ready-for-review:** full `make qualify` (e2e + grype scan).

## Risk Assessment

- [x] **Medium** — Touches the inference-perf validator, recipe overlays, and the
  EBS-CSI default-SC; default model/concurrency changed (heavier default run).

**Rollout notes:** the default inference-perf run now uses 8B at conc 256 with the
cache on (slower, ~6–7 min, benefits from the cache); disable via
`AICR_INFERENCE_PERF_MODEL_CACHE_SIZE=off`.

> ⚠️ **Cluster-wide default StorageClass — wants conscious maintainer sign-off.**
> This PR sets `defaultStorageClass.enabled: true` on `aws-ebs-csi-driver`, which
> provisions a **cluster-default** gp3 SC (`ebs-csi-default-sc`) on **every**
> AICR-deployed EKS cluster that pulls in the component — **not just inference
> recipes**; training overlays inherit it too. Two consequences a maintainer
> should approve knowingly:
> 1. If a cluster already has a default SC, this creates a two-default ambiguity
>    (Kubernetes treats multiple defaults as undefined) — the operator must unset
>    the other default.
> 2. Any unrelated PVC that previously failed-fast on "no default SC" will now
>    silently bind gp3, which can mask a misconfiguration.
>
> This is intentional (EKS ships no default SC, so the model-weights cache PVC
> would otherwise hang `Pending`), and is the one change flagged in review
> (@njhensley) as needing explicit maintainer approval. The inline `NOTE:` in
> `recipes/components/aws-ebs-csi-driver/values.yaml` documents the same caveat.

**HF_TOKEN forwarding (reviewed, accepted):** the token reaches the validator pod
as a plaintext pod-spec env (workers/cache Job already use `secretKeyRef`). The
`secretKeyRef`-on-validator alternative was considered and declined — narrow
upside, doesn't remove the runtime plaintext, low-sensitivity rate-limit token in
an ephemeral per-run namespace. The trade-off is documented inline at
`pkg/validator/v1/job_plan_internal.go`.

## Checklist


docs/contributor/validator.md
```diff
@@ -149,16 +149,57 @@ The validator engine mounts snapshot and recipe data as ConfigMaps:
...
+    constraints:
+      - name: inference-model
+        value: Qwen/Qwen3-8B          # HF model ID (bare value, no comparator)
+      - name: inference-concurrency-per-gpu
+        value: "256"                  # positive integer
+      - name: inference-throughput
+        value: ">= 50000"
+      - name: inference-ttft-p99
+        value: "<= 500"
+```
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in `badaa911` — aligned this example's `inference-ttft-p99` to `<= 1000` to match the overlay guidance and the later example in the same doc. Thanks\!

docs/user/validation.md
```diff
@@ -131,12 +138,42 @@ validation:
...
+**Model-weights cache and `MODEL_CACHE_STORAGE_CLASS`.** The benchmark downloads
+the model **once** into a PVC and serves all workers from it (on by default;
+avoids per-IP Hugging Face throttling). The cache PVC needs a StorageClass: it
+uses the cluster's **default** StorageClass unless you set
+`AICR_INFERENCE_PERF_MODEL_CACHE_STORAGE_CLASS`. On a cluster with **no default
+StorageClass** (common on EKS — e.g. only a non-default `gp2`) and no value set,
+the check **fails fast** in seconds with guidance rather than hanging; set
+`AICR_INFERENCE_PERF_MODEL_CACHE_STORAGE_CLASS=<name>` (e.g. `gp2`/`gp3` on EKS,
+`standard-rwo` on GKE) on the `inference-perf` catalog entry's `env` (or via
+`aicr validate --data <overlay>`), or disable the cache with
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in `badaa911` — reworded to "a catalog overlay in the `aicr validate --data <dir>` directory", since `--data` reads a directory (not an overlay identifier). Thanks\!

pkg/validator/v1/job_plan_internal.go
```diff
@@ -89,6 +95,16 @@ func buildEnv(
...
+	// when present; absent leaves workers on anonymous downloads. Scoped to the
+	// inference-perf entry so the credential is not injected into unrelated
+	// validator Pods (deployment/conformance checks) that never use it.
+	if tok := os.Getenv("HF_TOKEN"); tok != "" && entry.Name == inferencePerfCheckName {
+		env = append(env, corev1.EnvVar{Name: "HF_TOKEN", Value: tok})
+	}
+
 	// Add catalog entry's custom env vars
 	for _, e := range entry.Env {
 		env = append(env, corev1.EnvVar{Name: e.Name, Value: e.Value})
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in `badaa911` — the `entry.Env` loop now skips any `HF_TOKEN` entry, so a catalog value can't override (or duplicate) the orchestrator-forwarded token. Good catch on the trust boundary.

recipes/overlays/b200-gke-cos-inference-dynamo.yaml
```diff
@@ -73,16 +73,22 @@ spec:
...
+          value: Qwen/Qwen3-8B
+        - name: inference-concurrency-per-gpu
+          value: "256"
         - name: inference-throughput
-          value: ">= 5000"
+          value: ">= 50000"
         - name: inference-ttft-p99
-          value: "<= 200"
+          value: "<= 2000"
     conformance:
```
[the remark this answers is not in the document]
yuanchen8911: Intentional for now (this is a WIP draft), keeping it with a clear justification:

The placeholder mirrors the **measured GB200** (Blackwell) gate — throughput `>= 50000` and a deliberately generous `<= 2000` ms TTFT. A B200 is the same architecture family and should meet or exceed those numbers, so this acts as a **conservative Blackwell-family floor**, not a too-loose limit, and it's commented UNTESTED in-file.

The alternative you suggest — disabling the inference-perf gate for B200 — would let B200 inference-perf pass with **no check at all**, which is strictly worse than a conservative floor. The values will be replaced with measured results once a B200 testbed run is available (tracked in #1116). Not blocking this PR.

validators/performance/inference_perf_constraint.go
```diff
@@ -758,13 +842,52 @@ func applyInferenceWorkerScheduling(obj *unstructured.Unstructured,
...
+	secret := &v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{Name: hfTokenSecretName, Namespace: namespace},
+		Type:       v1.SecretTypeOpaque,
+		StringData: map[string]string{hfTokenSecretKey: token},
+	}
+	_, err := secrets.Create(ctx.Ctx, secret, metav1.CreateOptions{})
+	if apierrors.IsAlreadyExists(err) {
+		if _, err = secrets.Update(ctx.Ctx, secret, metav1.UpdateOptions{}); err != nil {
+			return errors.Wrap(errors.ErrCodeInternal, "failed to update HF token secret", err)
+		}
```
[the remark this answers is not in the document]
yuanchen8911: Already addressed in `6bf61be7` (the commit before this round): `ensureHFTokenSecret` now does a Get-then-Update and sets `secret.ResourceVersion` from the existing object before `Update` (and additionally deletes a stale secret when `HF_TOKEN` is unset). Thanks\!

validators/performance/model_cache.go
```diff
@@ -0,0 +1,295 @@
...
+func clusterHasDefaultStorageClass(ctx *validators.Context) (bool, error) {
+	listCtx, cancel := context.WithTimeout(ctx.Ctx, defaults.DiagnosticTimeout)
+	defer cancel()
+	scs, err := ctx.Clientset.StorageV1().StorageClasses().List(listCtx, metav1.ListOptions{})
+	if err != nil {
+		return false, errors.Wrap(errors.ErrCodeInternal, "failed to list StorageClasses for cache pre-flight", err)
+	}
+	for i := range scs.Items {
+		if scs.Items[i].Annotations[defaultStorageClassAnnotation] == "true" {
+			return true, nil
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in `badaa911` — `clusterHasDefaultStorageClass` now treats both `storageclass.kubernetes.io/is-default-class` and the legacy `storageclass.beta.kubernetes.io/is-default-class` as default-class markers. Thanks\!

recipes/overlays/h100-eks-ubuntu-inference-dynamo.yaml
```diff
@@ -71,16 +71,21 @@ spec:
...
+      # gate catches real regressions without flaking. Model + concurrency are
+      # pinned so the gate stays valid independent of the compiled defaults.
       constraints:
+        - name: inference-model
+          value: Qwen/Qwen3-8B
+        - name: inference-concurrency-per-gpu
+          value: "256"
         - name: inference-throughput
-          value: ">= 5000"
+          value: ">= 50000"
```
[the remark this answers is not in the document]
yuanchen8911: Keeping `>= 50000` here intentionally. It's a deliberate **uniform floor** across accelerators (same value on RTX PRO 6000 and GB200), acting as a conservative "grossly-broken" gate rather than a tight per-accelerator regression detector. Tightening H100 to ~90k (≈83% of the ~108k baseline, matching the RTX/GB200 margin) was considered and explicitly declined for now — uniform and never-flaky while this is a WIP draft. Worth revisiting with proper per-accelerator floors once the gates leave WIP. Not changing in this PR.

recipes/overlays/h100-gke-cos-inference-dynamo.yaml
```diff
@@ -71,16 +71,21 @@ spec:
...
+        - name: inference-model
+          value: Qwen/Qwen3-8B
+        - name: inference-concurrency-per-gpu
+          value: "256"
         - name: inference-throughput
-          value: ">= 5000"
+          value: ">= 50000"
         - name: inference-ttft-p99
-          value: "<= 200"
+          value: "<= 1000"
```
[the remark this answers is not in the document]
yuanchen8911: Same as the EKS H100 overlay: keeping `>= 50000` as a deliberate uniform conservative floor for now (declined tightening to ~90k while WIP). See the reply on `h100-eks-ubuntu-inference-dynamo.yaml`.

validators/performance/inference_perf_test.go
```diff
@@ -724,6 +757,266 @@ func TestValidatePerfTuningEnvs(t *testing.T) {
...
+		t.Setenv(envHFToken, "hf_new")
+		client := fake.NewClientset(&v1.Secret{
+			ObjectMeta: metav1.ObjectMeta{Name: hfTokenSecretName, Namespace: ns, ResourceVersion: "7"},
+			StringData: map[string]string{hfTokenSecretKey: "hf_old"},
+		})
+		ctx := &validators.Context{Ctx: context.Background(), Clientset: client}
+		if err := ensureHFTokenSecret(ctx, ns); err != nil {
+			t.Fatalf("unexpected error updating existing secret: %v", err)
+		}
+	})
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in `c4deb44a` — the "existing secret → updated" test now asserts the stored token is actually replaced `hf_old` → `hf_new` (reads the Secret back and checks the value), not just that no error is returned.

validators/performance/inference_perf_constraint.go
```diff
@@ -758,13 +842,72 @@ func applyInferenceWorkerScheduling(obj *unstructured.Unstructured,
...
+	}
+	if exists {
+		secret.ResourceVersion = existing.ResourceVersion
+		if _, err := secrets.Update(ctx.Ctx, secret, metav1.UpdateOptions{}); err != nil {
+			return errors.Wrap(errors.ErrCodeInternal, "failed to update HF token secret", err)
+		}
+		return nil
+	}
+	if _, err := secrets.Create(ctx.Ctx, secret, metav1.CreateOptions{}); err != nil {
+		return errors.Wrap(errors.ErrCodeInternal, "failed to create HF token secret", err)
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in `c4deb44a` — the `Get`/`Delete`/`Update`/`Create` calls now run under `context.WithTimeout(ctx.Ctx, defaults.DiagnosticTimeout)`, so a slow/wedged apiserver can't burn the check's overall deadline during setup.

validators/performance/model_cache.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return errors.Wrap(errors.ErrCodeInternal, "failed to create model-cache PVC", err)
+	}
+
+	jobName := fmt.Sprintf("%s-%s", modelCachePopulateJobPrefix, config.runID)
+	job := buildModelCachePopulateJob(jobName, config)
+	_, err = ctx.Clientset.BatchV1().Jobs(config.namespace).Create(ctx.Ctx, job, metav1.CreateOptions{})
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		return errors.Wrap(errors.ErrCodeInternal, "failed to create model-cache populate Job", err)
+	}
```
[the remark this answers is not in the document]
yuanchen8911: Deferring — heavy lift for a near-unreachable case. The cache PVC and populate Job live in the **per-run namespace** (`aicr-inference-perf-<runID>`, unique per `aicr validate` invocation), so the `AlreadyExists` path can only be hit by a same-runID retry, which resolves the **same** `config.model`. A different model resolves a different namespace, so there's no stale-weights window in practice. Versioning the PVC by a model hash would only matter if we ever reuse namespaces across runs; tracked as future work in #1116.

recipes/validators/catalog.yaml
```diff
@@ -68,28 +68,57 @@ validators:
...
-    timeout: 50m
+    # model-cache populate (up to WORKLOAD_READY_TIMEOUT, 10m; cache is on by
+    # default) + namespace-termination (5m) + workload-ready (10m) + health (5m)
+    # + AIPerf pod scheduling (5m) + AIPerf benchmark (15m) = 50m worst-case
+    # phase sum, with headroom inside the ctx for inter-phase overhead. The
+    # Job's activeDeadlineSeconds is set from this value and kills the pod only
+    # after the ctx has surfaced a clean timeout; 65m leaves ~10m for the
+    # cache-populate phase plus deferred cleanup / CTRF-tail work. Raising the
+    # model (or disabling the cache) shifts this budget — see the knobs below.
+    timeout: 65m
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in `c4deb44a` — reworded to remove the ambiguity: the **50m is the worst-case serial sum that already includes** the 10m cache-populate phase; the 65m timeout = that 50m + 15m headroom (inter-phase overhead + deferred cleanup). Cache-populate is not additive on top of 50m.

validators/performance/inference_perf_constraint.go
```diff
@@ -1234,6 +1386,70 @@ func getOwnPullSecrets(ctx *validators.Context) []v1.LocalObjectReference {
 	return out
 }
 
+// resolveInferenceModel returns the Hugging Face model ID used for the
+// benchmark. Override via AICR_INFERENCE_PERF_MODEL to characterize a larger
+// model (e.g., Qwen/Qwen3-32B) without rebuilding the validator image; unset or
+// empty falls back to the default smoke-test model. A non-existent model ID
+// surfaces as a deploy / endpoint-ready timeout (fail-closed), never a silent pass.
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in `c4deb44a` — the comment now reads "falls back to the default model (inferenceModel, Qwen/Qwen3-8B)", dropping the stale "smoke-test model" wording (the default is now the real-workload 8B).

validators/performance/model_cache.go
```diff
@@ -0,0 +1,307 @@
...
+							{Name: "AICR_MODEL", Value: config.model},
+							{Name: "HF_HOME", Value: modelCacheMountPath},
+							{Name: "HF_TOKEN", ValueFrom: &v1.EnvVarSource{SecretKeyRef: &v1.SecretKeySelector{
+								LocalObjectReference: v1.LocalObjectReference{Name: hfTokenSecretName},
+								Key:                  hfTokenSecretKey,
+								Optional:             ptr.To(true),
+							}}},
+						},
+						VolumeMounts: []v1.VolumeMount{{Name: modelCacheVolumeName, MountPath: modelCacheMountPath}},
+					}},
```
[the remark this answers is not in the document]
yuanchen8911: Intentionally leaving the populate container without resource requests/limits. It downloads a model via `huggingface_hub` (which buffers in memory), so a memory **limit** risks OOM-killing a large download. The Job is pinned to the GPU node (`a3-megagpu`/`p5`/`p6e`/`g7e`), which has ample memory, so scheduling without a request isn't a concern here. Unbounded is the safer default for this one-shot downloader.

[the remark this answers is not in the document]
yuanchen8911: Added explicit resource requests/limits to the populate container in da8238b4: requests `cpu: 100m` / `memory: 512Mi`, limit `memory: 4Gi` (constants `cacheJobCPURequest`/`cacheJobMemoryRequest`/`cacheJobMemoryLimit` in model_cache.go). snapshot_download is network/IO-bound so CPU stays small; the memory limit guards against huggingface_hub download buffering, and the explicit requests let the pod schedule under ResourceQuota/LimitRange. Covered by a new assertion in TestBuildModelCachePopulateJob.

validators/performance/model_cache.go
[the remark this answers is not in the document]
yuanchen8911: @coderabbitai I looked into this, but I don't think the stale-reuse scenario can occur in normal operation, so I'm leaving the code as-is — please re-evaluate.

The PVC name (`aicr-model-cache`) is fixed, but it lives in a **per-run namespace**: `config.namespace` is `inference-workload-<runID>`, where `runID` is unique per `aicr validate` invocation (`deriveRunID` — random 4-byte hex for standalone runs, or SHA-256 of the CLI's per-invocation `AICR_RUN_ID`). The entire namespace is torn down at the end of every run via the deferred `cleanupInferenceWorkload` (`inference_perf_constraint.go:326` → `Namespaces().Delete` at `:1200`).

So the `AlreadyExists` branches here are intra-run retry/idempotency safety, not cross-run reuse — a second run with a different model gets a fresh namespace and a fresh PVC. Hitting the "stale weights / wrong model" case would require **all** of: pinning an identical `AICR_RUN_ID` across two separate invocations, changing the model between them, **and** the prior run's namespace cleanup having failed (crash/timeout). That is outside the supported usage.

On the cited guideline (generation-counter cache keying): that targets process-global `sync.Once`-style caches whose lifetime spans config changes. These artifacts are namespace-isolated and recreated per run, so the same hazard doesn't apply.

recipes/components/aws-ebs-csi-driver/values.yaml
```diff
@@ -56,8 +56,16 @@ node:
...
+# WaitForFirstConsumer, expandable, annotated is-default-class=true). EKS ships
+# no default StorageClass on its own, so without this any PVC created without an
+# explicit storageClassName — including the inference-perf model-weights cache —
+# stays Pending. Enabling it here makes dynamic provisioning "just work" on
+# AICR-deployed EKS clusters. NOTE: this sets a *cluster-wide* default; if the
+# cluster already has another default StorageClass, unset that one to avoid two
+# defaults (which Kubernetes treats as ambiguous).
 defaultStorageClass:
-  enabled: false
+  enabled: true
```
njhensley, line 68: **Cluster-wide blast radius — wants explicit sign-off.** Flipping `defaultStorageClass.enabled` to `true` makes *every* AICR-deployed EKS cluster get a cluster-default gp3 SC, not just inference recipes — training overlays that pull in `aws-ebs-csi-driver` inherit it too. Two concerns:

1. If a cluster already has a default SC, this creates the two-default ambiguity the comment warns about, but nothing prevents it.
2. Any unrelated PVC that previously failed-fast on "no default SC" will now silently bind gp3, masking misconfig.

Is the cluster-wide default intended, or should the default-SC requirement be scoped to the inference-perf path (or made opt-in)? Either way, please call this out prominently in the rollout notes. This is the one change I'd want a maintainer to consciously approve.

pkg/validator/v1/job_plan_internal.go
```diff
@@ -89,8 +95,25 @@ func buildEnv(
...
 
-	// Add catalog entry's custom env vars
+	// Forward a Hugging Face token from the orchestrator's environment (never
+	// the in-repo catalog) so inference-perf can provision it for model
+	// downloads that would otherwise hit HF anonymous rate limits. Only set
+	// when present; absent leaves workers on anonymous downloads. Scoped to the
+	// inference-perf entry so the credential is not injected into unrelated
+	// validator Pods (deployment/conformance checks) that never use it.
+	if tok := os.Getenv("HF_TOKEN"); tok != "" && entry.Name == inferencePerfCheckName {
+		env = append(env, corev1.EnvVar{Name: "HF_TOKEN", Value: tok})
```
njhensley, line 124: The forwarded `HF_TOKEN` lands in the validator Job's pod spec as a plaintext `EnvVar.Value`, so it's readable by anyone with `get pod`/`get job` in the run namespace and stored unencrypted in etcd (absent encryption-at-rest). The *workers* correctly consume it via the `aicr-hf-token` Secret + `secretKeyRef` — only this orchestrator→validator hop is plaintext.

Mitigated by the per-run ephemeral namespace and the token being a rate-limit lifter rather than a high-privilege credential, so not a blocker. The cleaner pattern would be to stage it into a Secret and reference it via `secretKeyRef` on the validator pod too. The trust-boundary handling (dropping any catalog-supplied `HF_TOKEN`) is otherwise excellent.

validators/performance/model_cache.go
```diff
@@ -0,0 +1,341 @@
...
+
+	// modelCachePopulateJobPrefix prefixes the one-time download Job (suffixed
+	// with the run ID so concurrent runs don't collide).
+	modelCachePopulateJobPrefix = "aicr-model-cache-populate"
+
+	// cacheWorkerImage is the image used by the populate Job to download
+	// weights. It must carry python3 + huggingface_hub; reuse the vLLM runtime
+	// image so it is already node-cached. Keep in sync with the worker image in
+	// testdata/inference/dynamo-deployment.yaml.
+	cacheWorkerImage = "nvcr.io/nvidia/ai-dynamo/vllm-runtime:0.9.0"
```
njhensley, line 94: `cacheWorkerImage` is hardcoded and not run through `catalog.ResolveImage`, so unlike the AIPerf image it won't be rewritten by `AICR_VALIDATOR_IMAGE_REGISTRY`. Propagating `imagePullSecrets` to the populate Job only helps when the private creds are for `nvcr.io` itself, not a different mirror host — a true air-gapped mirror would `ImagePullBackOff` here.

This is consistent with the existing Dynamo worker image (also pinned to `nvcr.io` in the template), and `TestCacheWorkerImageMatchesTemplate` guards drift, so it's a known-shape limitation rather than a regression. Worth a one-line note that mirror/air-gapped support for the whole inference-perf path is a separate follow-up.

validators/performance/model_cache.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+}
+
+// injectModelCacheMounts adds the read-only cache PVC volume, its mount, and the
+// HF_HOME / HF_HUB_OFFLINE env to a service's extraPodSpec so the worker loads
+// weights from the pre-populated PVC and never reaches out to Hugging Face.
+// Offline mode fails closed if the cache is incomplete rather than silently
+// falling back to a (rate-limited) download. Operates on the unstructured map
+// in place, merging with any env the template already declares.
+func injectModelCacheMounts(extraPodSpec map[string]interface{}) {
```
njhensley, line 343: nit: `applyInferenceWorkerScheduling` calls this for every service, so the `Frontend` pod also mounts the RWO cache read-only and gets `HF_HUB_OFFLINE=1`. RWO co-located on one node is fine, and `snapshot_download` fetches the full repo (tokenizer included) so an offline frontend still resolves — but the frontend doesn't need weights and the injection is incidental. Just confirming this is intentional rather than over-broad.

validators/performance/model_cache.go
```diff
@@ -0,0 +1,341 @@
...
+	}
+	return false, nil
+}
+
+// ensureModelCache provisions the model-weights cache when enabled: an RWO PVC
+// plus a one-time populate Job (pinned to the same node as the workers, so the
+// WaitForFirstConsumer RWO volume binds there) that downloads the model into it.
+// It blocks until the download completes so the cache is ready before the
+// DynamoGraphDeployment workers start. A no-op when the cache is disabled.
+func ensureModelCache(ctx *validators.Context, config *inferenceWorkloadConfig) error {
```
njhensley, line 179: The no-default-StorageClass fail-fast (`clusterHasDefaultStorageClass` → `ErrCodeInvalidRequest`) is the operator-facing happy-failure path you most want pinned, but I don't see a unit test for it. A fake clientset with no default-annotated SC + cache enabled + empty `MODEL_CACHE_STORAGE_CLASS`, asserting `ErrCodeInvalidRequest`, would lock in the behavior you verified on GB200.

recipes/overlays/b200-gke-cos-inference-dynamo.yaml
```diff
@@ -71,18 +71,16 @@ spec:
...
+      # inference-perf is DISABLED on B200 until a reference benchmark exists.
+      # B200 has not been measured, and shipping a hard gate with UNTESTED
+      # PLACEHOLDER thresholds risks passing against non-representative limits
+      # and hiding a real regression. Re-enable once a B200 GKE testbed run
+      # (e.g. nvcf-dgxc-k8s-gcp-azne1-prd7) publishes measured values: set
+      # checks to [inference-perf] and add a constraints block with the
+      # measured throughput / TTFT p99. As a conservative Blackwell-family
+      # starting point, the measured GB200 gate at Qwen/Qwen3-8B,
+      # concurrency=256/GPU was throughput ≈ 66k tok/s, TTFT p99 ≈ 1,240 ms.
+      checks: []
```
njhensley, line 83: Good call disabling the gate rather than shipping placeholder thresholds. Quick confirm: does the performance phase handle an **empty** `checks` list cleanly (skip, not error)? This is a new state for this overlay, so worth a 30-second manual `aicr validate --phase performance` against it (or pointing at the test that covers empty-checks) before merge.

njhensley (commented):
Architecturally sound and unusually well-tested — fail-closed env parsing, the `"$AICR_MODEL"` shell-injection avoidance (with a direct `$(...)` test), the catalog-vs-facade timeout guard test, and the cgroup-v2 page-cache reasoning behind the populate container's missing memory limit are all excellent.

My only real gating item is a conscious decision on the **cluster-wide default StorageClass** flip (inline below); everything else is nits, one low-severity security note, and a couple of process confirmations.

**Process confirmations before flipping out of Draft:**
1. `recipes/components/aws-ebs-csi-driver/values.yaml` changed — per CLAUDE.md, run `make bom-docs` and commit `docs/user/container-images.md` if it changes. Toggling `defaultStorageClass.enabled` renders only a StorageClass (no new images) so it's almost certainly a no-op, but `bom-check` isn't in the merge gate, so worth confirming.
2. PR notes `make qualify` (e2e + grype) is still pending — that's the gate that matters most for an XL validator change that touches cluster-wide defaults.

LGTM once the StorageClass decision is settled and `make qualify` is green.

recipes/components/aws-ebs-csi-driver/values.yaml
yuanchen8911: Confirmed intended, and surfaced for conscious maintainer sign-off. The cluster-wide default is deliberate — EKS ships no default StorageClass, so the model-cache PVC would otherwise hang `Pending`. Kept `defaultStorageClass.enabled: true`, documented the two-default ambiguity inline in `values.yaml`, and added a prominent ⚠️ rollout-note callout in the PR description (blast radius beyond inference recipes, two-default caveat, silent-bind risk) for explicit approval. (02679d79)

pkg/validator/v1/job_plan_internal.go
yuanchen8911: Agreed — not a blocker. I weighed the orchestrator-creates-Secret + `secretKeyRef` pattern and decided to document it as an accepted trade-off rather than implement it: the upside is narrow (keeps the value out of the pod spec, where `get pods` is granted far more liberally than `get secrets`) and it does **not** remove the runtime plaintext (still readable via `kubectl exec`/env; the Secret is likewise base64-in-etcd absent encryption-at-rest) — for a low-sensitivity, easily-rotated rate-limit token in an ephemeral per-run namespace. The rationale is now documented inline at `pkg/validator/v1/job_plan_internal.go` with a "revisit if a higher-privilege credential ever flows this path" note. (02679d79)

validators/performance/model_cache.go
yuanchen8911: Added the note. `cacheWorkerImage` is pinned to `nvcr.io` and is **not** run through `catalog.ResolveImage`, so `AICR_VALIDATOR_IMAGE_REGISTRY` does not rewrite it; propagating `imagePullSecrets` only helps when the creds are for `nvcr.io` itself, and a true air-gapped mirror host would `ImagePullBackOff`. Documented inline that this matches the Dynamo worker image (drift-guarded by `TestCacheWorkerImageMatchesTemplate`) and that full mirror/air-gap support for the inference-perf path is a separate follow-up. (02679d79)

validators/performance/model_cache.go
yuanchen8911: Confirmed intentional. `applyInferenceWorkerScheduling` injects per-service, so the Frontend also receives the RO mount — harmless: the RWO PVC is co-located on one node (which the validator already enforces) and `snapshot_download` fetches the full repo (tokenizer included), so an offline frontend still resolves. Added a comment to `injectModelCacheMounts` stating this is per-service-by-design and harmless. (02679d79)

validators/performance/model_cache.go
yuanchen8911: Test added — `TestEnsureModelCache_NoDefaultStorageClassFailsFast` uses a fake clientset with no default-annotated SC + cache enabled + empty `MODEL_CACHE_STORAGE_CLASS` and asserts `ErrCodeInvalidRequest`; `TestClusterHasDefaultStorageClass` separately covers the detection helper (including the "no default" case). This locks in the GB200-verified fail-fast behavior.

recipes/overlays/b200-gke-cos-inference-dynamo.yaml
yuanchen8911: Confirmed clean — an empty `checks` list is a no-op, not an error. `FilterEntriesByValidation` returns an empty set, the phase loop iterates zero entries, and `WarnPhasesAgainstRecipe` emits a `slog.Warn` ("phase requested but no checks defined; phase will be empty") rather than failing. So `b200-gke-cos-inference-dynamo` with `checks: []` skips the performance gate cleanly while keeping its conformance checks.

yuanchen8911:
@njhensley thanks for the thorough review — here's how each point was addressed (all in `02679d79`, rebased on current `main`):

| # | Comment | Resolution |
|---|---|---|
| 1 | Cluster-wide default StorageClass — wants conscious sign-off | Kept intentionally (EKS ships no default SC, so the model-cache PVC would otherwise hang `Pending`). Two-default caveat documented inline in `values.yaml`; added a prominent ⚠️ rollout-note callout in the PR description (blast radius beyond inference, two-default ambiguity, silent-bind risk) for explicit maintainer approval. |
| 2 | `HF_TOKEN` plaintext on validator pod (non-blocker) | Considered the orchestrator-creates-Secret + `secretKeyRef` pattern and **declined as an accepted trade-off**: narrow upside (keeps it out of the pod spec, where `get pods` is more widely granted than `get secrets`) that doesn't remove the runtime plaintext, for a low-sensitivity rate-limit token in an ephemeral namespace. Rationale documented inline at `pkg/validator/v1/job_plan_internal.go`, with a "revisit for higher-privilege credentials" note. |
| 3 | `cacheWorkerImage` not run through `ResolveImage` (air-gap) | Added inline note: pinned to `nvcr.io`, not registry-rewritten, `imagePullSecrets` only help for `nvcr.io` creds — matches the Dynamo worker image (drift-guarded by `TestCacheWorkerImageMatchesTemplate`); full mirror/air-gap support is a separate follow-up. |
| 4 | Frontend pod also mounts the RO cache (nit) | Confirmed intentional (per-service injection) and harmless (co-located RWO PVC, full-repo download resolves offline); added a clarifying comment to `injectModelCacheMounts`. |
| 5 | Missing unit test for no-default-SC fail-fast | Covered — `TestEnsureModelCache_NoDefaultStorageClassFailsFast` (fake clientset, no default SC, empty `MODEL_CACHE_STORAGE_CLASS` → asserts `ErrCodeInvalidRequest`) plus `TestClusterHasDefaultStorageClass`. |
| 6 | Empty `checks` list — skip not error? | Confirmed clean: `FilterEntriesByValidation` returns an empty set, the phase loop is a no-op, and `WarnPhasesAgainstRecipe` only `slog.Warn`s — so `b200-gke-cos-inference-dynamo` with `checks: []` skips the performance gate cleanly. |

Inline replies posted under each thread as well.

njhensley (dismissed):
LGTM

recipes/components/aws-ebs-csi-driver/values.yaml
```diff
@@ -56,8 +56,16 @@ node:
 
 storageClasses: []
 
+# Provision a cluster-default StorageClass (ebs-csi-default-sc: gp3,
+# WaitForFirstConsumer, expandable, annotated is-default-class=true). EKS ships
```
mchmarny, line 60: Reinforcing your own callout in the PR body: this is a **cluster-wide** change that affects every AICR-deployed EKS cluster pulling in `aws-ebs-csi-driver` — training overlays inherit it too, not just inference recipes. The inline `NOTE:` is the right place to document it for someone reading the values file, but an operator reading `docs/user/component-catalog.md` won't see it.

Suggest a one-paragraph mention under the `aws-ebs-csi-driver` entry in `docs/user/component-catalog.md` covering the two consequences (multi-default ambiguity, silent-binding of previously-failed PVCs) so the caveat is discoverable from the user-facing catalog without having to find the inline comment.

Not blocking — you already flagged this for explicit maintainer sign-off.

validators/performance/model_cache.go
```diff
@@ -0,0 +1,356 @@
...
+	modelCacheVolumeName = "model-cache"
+
+	// modelCachePopulateJobPrefix prefixes the one-time download Job (suffixed
+	// with the run ID so concurrent runs don't collide).
+	modelCachePopulateJobPrefix = "aicr-model-cache-populate"
+
+	// cacheWorkerImage is the image used by the populate Job to download
+	// weights. It must carry python3 + huggingface_hub; reuse the vLLM runtime
+	// image so it is already node-cached. Keep in sync with the worker image in
+	// testdata/inference/dynamo-deployment.yaml.
```
mchmarny, line 83: The inline comment is honest about the gap — `cacheWorkerImage` is pinned to `nvcr.io` and bypasses `catalog.ResolveImage`, so `AICR_VALIDATOR_IMAGE_REGISTRY` doesn't rewrite it and a true air-gap mirror host would `ImagePullBackOff` here. The `TestCacheWorkerImageMatchesTemplate` drift guard against `testdata/inference/dynamo-deployment.yaml` is the right floor.

Suggest filing a follow-up issue so the air-gap parity gap is tracked rather than just commented. The current state is fine to ship; just want it on the board.

pkg/validator/v1/job_plan_internal.go
```diff
@@ -29,6 +30,11 @@ import (
 	applycorev1 "k8s.io/client-go/applyconfigurations/core/v1"
 )
 
+// inferencePerfCheckName is the catalog name of the inference performance check.
+// Used to scope HF_TOKEN forwarding to that validator only (it is the sole
+// check that downloads models from Hugging Face).
+const inferencePerfCheckName = "inference-perf"
```
mchmarny: String coupling to the catalog entry name in `recipes/validators/catalog.yaml`. If someone ever renames the catalog entry, HF_TOKEN forwarding silently no-ops on the renamed entry without test failure.

Two options:
1. Reference a shared constant the catalog parser already exposes (preferred if one exists in `pkg/validator/catalog`).
2. Add a parity assertion in `pkg/validator/catalog/catalog_test.go` that an entry named `"inference-perf"` exists in the embedded catalog.

The `non-inference-perf entry → not forwarded` test (great test, btw) guards the negative case but not the rename — a fresh catalog entry with a different name would silently lose forwarding.

validators/performance/model_cache.go
```diff
@@ -0,0 +1,356 @@
...
+		Spec: batchv1.JobSpec{
+			BackoffLimit: &backoff,
+			Template: v1.PodTemplateSpec{
+				Spec: v1.PodSpec{
+					RestartPolicy:    v1.RestartPolicyNever,
+					NodeSelector:     config.gpuNodeSelector,
+					Tolerations:      config.gpuTolerations,
+					ImagePullSecrets: pullSecrets,
+					SecurityContext:  &v1.PodSecurityContext{FSGroup: ptr.To(int64(1000))},
+					Containers: []v1.Container{{
```
mchmarny, line 292: nit: `FSGroup: 1000` is hardcoded. If the vLLM runtime image rebuilds with a different non-root UID, the cache files won't be readable by the worker (and the offline mount would silently fail to load weights).

Worth either:
- A short comment noting that 1000 matches the `cacheWorkerImage` non-root UID, or
- Hoisting to a named constant alongside `cacheWorkerImage` so the drift surface is co-located.

Optional polish; not blocking.

validators/performance/model_cache.go
```diff
@@ -0,0 +1,356 @@
...
+	backoff := int32(2)
+	return &batchv1.Job{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: config.namespace,
+			Labels:    map[string]string{"app.kubernetes.io/name": "aicr-model-cache-populate"},
+		},
+		Spec: batchv1.JobSpec{
+			BackoffLimit: &backoff,
+			Template: v1.PodTemplateSpec{
```
mchmarny, line 285: nit: `BackoffLimit: 2` on a 100Gi-class download means a partial-failure case can consume ~3× the download budget against the parent ctx deadline before failing — and `snapshot_download` doesn't resume across pods, so each retry restarts from scratch.

Worth confirming intent. Two reasonable shapes:
- **`BackoffLimit: 0`** — one attempt, fail fast. The parent deadline still bounds it; `snapshot_download` retries internally for transient HTTP errors anyway.
- **Keep `2`** if observed behavior is that the bulk of failures are early (DNS, image pull, auth) where a cheap restart actually helps.

Either is defensible; the comment + test guard on memory-limit makes me think this was considered.

mchmarny (commented):
Excellent work — the code, comments, and test coverage are all top-shelf. Every failure mode I went looking for (cgroup-v2 page-cache OOMKill, scalar YAML coercion of model IDs, shell injection via the model name, missing default StorageClass, two-default-SC ambiguity, namespace reuse leaving a stale HF_TOKEN, image registry-rewrite parity, RWO-PVC binding under WaitForFirstConsumer) was already addressed with a fail-closed pattern, an inline comment explaining the *why*, and a regression test.

Three medium notes and two nits inline. Nothing blocking the substance.

**CI**: 130 pass, 9 expected skips (Tier 3 nightly, the standard `*-skip` push-only jobs, CodeQL on a non-scheduled run) — fully green for what runs on a PR.

**Leaving as COMMENT** rather than APPROVE since the body says full `make qualify` (e2e + grype scan) is pending before ready-for-review. Happy to flip to APPROVE once those run.

One overall thought: the EBS-CSI default-SC change is the one piece of this PR with a blast radius beyond inference-perf — flagging it inline so the catalog doc gets a parallel note for operators reading from that direction.

recipes/components/aws-ebs-csi-driver/values.yaml
yuanchen8911: Good call — added a "Cluster-wide default StorageClass" note to the `aws-ebs-csi-driver` entry in `docs/user/component-catalog.md` covering both consequences (multi-default ambiguity, silent-binding of previously-failed PVCs), so the caveat is discoverable from the user-facing catalog rather than only the inline values comment. (0b97d699)

validators/performance/model_cache.go
yuanchen8911: Filed as #1159. One adjustment worth flagging: I scoped it as a registry-override consistency fix ("route `cacheWorkerImage` through `ResolveImage` like the AIPerf image") rather than "air-gap parity" — because registry rewrite alone isn't sufficient for air-gap: the populate Job's `snapshot_download` still reaches huggingface.co for the weights, which is the dominant air-gap blocker. The issue calls that out as a separate concern (pre-staged weights / internal model registry / ModelExpress), and notes the gap is pre-existing (the Dynamo worker image was already pinned). Code comment updated to match. Happy to broaden #1159 into a full air-gapped-inference-perf epic if you'd prefer that framing. (0b97d699)

pkg/validator/v1/job_plan_internal.go
yuanchen8911: Addressed both ways you suggested: exported the constant as `v1.InferencePerfCheckName`, and added `TestEmbeddedCatalog_InferencePerfEntryExists` in `pkg/validator/catalog/catalog_test.go` asserting an embedded entry with that exact name exists (and is in the `performance` phase). Now a rename of the catalog entry without updating the constant — or vice versa — fails the test instead of silently no-op'ing HF_TOKEN forwarding. (0b97d699)

validators/performance/model_cache.go
yuanchen8911: Hoisted to a named constant `cacheWorkerFSGroup` co-located with `cacheWorkerImage`, with a comment noting it must match the image's non-root UID/GID and that the workers' RO mount depends on the same group ownership — so the drift surface is in one place. (0b97d699)

validators/performance/model_cache.go
yuanchen8911: Went with `BackoffLimit: 0` (fail-fast). Rationale in the code: `snapshot_download` doesn't resume across pods, so each Job retry restarts the multi-GB download from scratch and burns the parent deadline (a 100Gi-class download retried 2× could eat ~3× the budget); `huggingface_hub` already retries transient HTTP errors in-process, and the parent context deadline bounds the single attempt. Early/cheap failures (image pull, auth) still surface immediately. (0b97d699)

mchmarny (approved):

PR #1136, opened 2026-06-01
fix(ci): stamp publish with resolved tag instead of releases/latest API

### LINKED ISSUE #1134: fix(ci): publish stamp shows stale version due to releases/latest API race
## Summary

The publish workflow's stamp step queries the GitHub `releases/latest` API to get the version for the "Latest · vX.Y.Z" display name. On a tag push, the Release object may not exist yet, causing the stamp to show the previous release version.

Observed in run 26770200319: v0.14.0 publish stamped "Latest · v0.13.0".

## Fix

Use the already-resolved tag from the earlier `resolve` step instead of the API call. The tag is always correct since it comes from the trigger event itself.

nvsentinel and topograph already use this pattern.

pdmack:
## Summary

Use the already-resolved tag from the `resolve` step to stamp the "Latest · vX.Y.Z" display name, instead of querying the GitHub `releases/latest` API which races with Release creation on tag push.

## Motivation / Context

On a tag push, the Release object may not exist yet, causing the stamp to show the previous version. Observed in run 26770200319: v0.14.0 publish stamped "Latest · v0.13.0".

Fixes: #1134

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `.github/workflows/publish-fern-docs.yml`

## Implementation Notes

Replaces the `curl releases/latest` API call with `${{ steps.resolve.outputs.tag }}` which is already available from the earlier resolve step. This is the same pattern nvsentinel and topograph use.

## Testing

No test commands — CI workflow change. Verified by inspecting the existing step logic and comparing with production patterns in nvsentinel/topograph.

## Risk Assessment

**Rollout notes:** Next `Publish Fern Docs` run will use the resolved tag. No migration needed.

## Checklist

### DIFF
--- a/.github/workflows/publish-fern-docs.yml
+++ b/.github/workflows/publish-fern-docs.yml
@@ -200,17 +200,12 @@ jobs:
           done
 
       - name: Stamp version in docs.yml
+        env:
+          TAG: ${{ steps.resolve.outputs.tag }}
         run: |
-          # Back up the pre-stamp docs.yml so we can persist registry-only
-          # changes (no transient Latest stamp) in the post-publish PR step.
           cp fern/docs.yml /tmp/docs.yml.preStamp
-          VERSION=$(curl -sf https://api.github.com/repos/${{ github.repository }}/releases/latest | jq -r .tag_name || true)
-          if [ -n "$VERSION" ] && [ "$VERSION" != "null" ]; then
-            export VERSION
-            yq -i '(.products[0].versions[] | select(.slug == "latest")).display-name = "Latest · " + env(VERSION)' fern/docs.yml
-          else
-            echo "::warning::Could not resolve latest GitHub release tag; continuing without stamping"
-          fi
+          export TAG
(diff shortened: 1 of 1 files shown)


mchmarny (approved):

pdmack:
Re: CodeRabbit's suggestion to introduce a `STAMP_TAG` fallback — this is a false positive.

**Pre-releases:** The stamp step runs unconditionally, but that's fine — pre-release publishes *should* show the pre-release tag. The stamp is transient (reverted before the registry PR) and only affects the published site for that run.

**Manual dispatch with an older tag:** If someone explicitly passes `v0.12.0`, they *want* to republish that version — stamping "Latest · v0.12.0" is correct for that run. The stamp reflects what was just published.

**The race condition this PR fixes:** On a tag push, `releases/latest` returns the *previous* release because the new Release hasn't been created yet. Observed in run 26770200319: v0.14.0 publish stamped "Latest · v0.13.0". The resolved tag is always correct because it comes from the trigger event.

CodeRabbit's suggested fix reintroduces the exact race condition this PR eliminates by falling back to `LATEST_RELEASE_TAG` from the API.

PR #1140, opened 2026-06-01
feat(evidence): add `aicr evidence publish` for off-network signing

### LINKED ISSUE #1130: [Feature]: Decouple evidence signing/publishing from validate
### Prerequisites

### Feature Summary

Add a standalone `aicr evidence publish` (sign + push + write pointer) that operates on an already-emitted on-disk evidence bundle, so the cluster-bound validation step and the Fulcio/Rekor-bound signing step can run on different networks.

### Problem/Use Case

Validation must run where the cluster is reachable (often a corporate VPN), but keyless signing must reach `fulcio.sigstore.dev` + `rekor.sigstore.dev`, which corporate networks frequently block. Today `aicr validate --emit-attestation --push` does validate → emit → **sign** → push → write `pointer.yaml` as one atomic step, so if Fulcio is unreachable the whole publish fails (`POST fulcio.sigstore.dev/api/v2/signingCert: EOF`, exit 6) even though validation, OCI push, and OIDC all succeeded.

### Proposed Solution

`aicr evidence publish <bundle-dir> --push <ref> [signing/OIDC flags]` that reuses the existing sign+push leg (today embedded in the `validate` emit path) against a pre-built `summary-bundle/`: package as OCI, keyless-sign the in-toto statement, attach the Sigstore referrer, write `pointer.yaml`. It signs the bundle bytes as-is (the timestamp is already baked at emit time → no determinism concern). `validate --emit-attestation` (no `--push`) remains the "produce unsigned bundle" step.

### Success Criteria

- `validate --emit-attestation <dir>` on VPN → unsigned bundle; `evidence publish <dir> --push <ref>` off-VPN → signed artifact + `pointer.yaml`, identical to today's one-shot output.
- `evidence verify` on the resulting pointer/OCI hits `signature-verify | passed`.
- One-shot `validate --emit-attestation --push` behavior is unchanged (no breaking change).
- All signing/OIDC flags (`--identity-token`, `--oidc-device-flow`, `--insecure-tls`) are available on the new command.

### Alternatives Considered

- Run the whole `validate` from a hotspot/off-VPN network — requires the cluster to be reachable from there, which it usually isn't.
- IT allowlist for `*.sigstore.dev`.
- KMS-backed signing (#407) — different mechanism, and still couples signing into the `validate` run.

### Component

CLI (aicr)

### Priority

None — additive. The new subcommand reuses the existing sign+push leg; the one-shot `validate --emit-attestation --push` path is untouched.

Decoupling lets the passing validation result be preserved as an artifact and signed/published from a host with Sigstore egress (CI runner, jump box, hotspot). The on-disk bundle is content-addressable, so the signed artifact is identical regardless of which step happened where.

Maybe, with guidance

njhensley:
## Summary

Add `aicr evidence publish <bundle-dir> --push <ref>`, which signs, pushes, and writes the pointer for an evidence bundle that `aicr validate --emit-attestation` (without `--push`) left on disk — splitting the one-shot path into a cluster-bound leg and a Sigstore-bound leg.

## Motivation / Context

The one-shot `validate --emit-attestation --push` requires one host to reach **both** the target cluster *and* Fulcio/Rekor. In practice those live on different networks: validation needs the cluster (often behind a corporate VPN), while keyless signing needs public Sigstore egress (frequently blocked on that same VPN). This change lets the two legs run on different hosts/networks without changing the signed artifact.

Because the predicate — including its baked-in `attestedAt` timestamp — is read verbatim from the bundle's `statement.intoto.json`, the published result is byte-for-byte identical to the one-shot output regardless of which host ran which leg.

Fixes: #1130
Related: ADR-007 (`docs/design/007-recipe-evidence.md`)

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: evidence (`pkg/evidence/attestation`, `pkg/evidence/verifier`)

## Implementation Notes

- **New command** `aicr evidence publish` and **new `attestation.Publish`** entry point. Publish loads a pre-built on-disk bundle, validates the `--push` ref up front (so a malformed ref doesn't waste a Fulcio cert + Rekor inclusion proof), signs the unsigned `statement.intoto.json` verbatim, pushes as an OCI artifact, attaches the Sigstore bundle as an OCI referrer, and writes `pointer.yaml`.
- **Shared sign+push pipeline.** Refactored the sign+push leg out of `EmitOptions` into a `signPushOptions` struct so both `Emit` (in-memory bundle) and `Publish` (on-disk bundle) drive the identical pipeline without sharing build-only fields (Recipe/Snapshot/PhaseResults).
- **Shared OIDC resolution.** `oidcResolveOptionsFromFlags` centralizes the `--identity-token` / `COSIGN_IDENTITY_TOKEN` / ambient-GHA / `--oidc-device-flow` precedence chain across `validate --push` and `evidence publish`. Token resolution stays deferred until adjacent to signing so Fulcio's nonce-binding window is respected.
- **Bundle discrimination** via `HasBundleMarkers` (recipe.yaml + manifest.json co-occurrence), accepting either the parent dir or `summary-bundle/` itself, mirroring the verifier's directory handling.

## Testing

```bash
# Targeted gate on affected packages (race + lint + gofmt)
go test -race ./pkg/cli/... ./pkg/evidence/...      # all ok
golangci-lint run -c .golangci.yaml ./pkg/cli/... ./pkg/evidence/...   # 0 issues
```

Plus a full **live end-to-end run** against a real EKS GB200 (Ubuntu, Dynamo) cluster:

1. `validate --emit-attestation ./out` (no `--push`) — all phases passed (deployment / performance / conformance), unsigned bundle written.
2. `evidence publish ./out --push ttl.sh/…:72h` — pushed, Fulcio keyless-signed (Rekor log index recorded), Sigstore bundle attached as referrer, `pointer.yaml` populated.
3. `evidence verify` confirmed **exit 0** across all three input modes (OCI digest ref, pointer file); the tag-only ref was correctly rejected.

Verified the cross-host content-addressability guarantee held: `attestedAt` from leg 1 was preserved verbatim in the signed output from leg 2.

> Note: `make qualify`'s cluster-bound e2e was exercised manually against live hardware rather than via the KWOK path; unit/lint gates run clean.

## Risk Assessment

**Rollout notes:** Purely additive — new subcommand + new exported `Publish`. The existing one-shot `validate --emit-attestation --push` path is unchanged (the refactor is internal and output-preserving). No migration required.

## Checklist

### DIFF
--- a/pkg/evidence/attestation/publish.go
+++ b/pkg/evidence/attestation/publish.go
@@ -1,0 +1,223 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package attestation
+
+import (
+	"context"
+	"encoding/json"
+	"log/slog"
+	"os"
+	"path/filepath"
+
+	bundleattest "github.com/NVIDIA/aicr/pkg/bundler/attestation"
+	"github.com/NVIDIA/aicr/pkg/errors"
+	"github.com/NVIDIA/aicr/pkg/oci"
+)
+
+// PublishOptions controls a single Publish run. Publish operates on an
+// already-emitted on-disk bundle so the cluster-bound validate step and
+// the Fulcio/Rekor-bound signing step can run on different networks (see
+// ADR-007 and issue #1130).
+type PublishOptions struct {
+	// BundleDir is the on-disk evidence directory. It may be either the
+	// OutDir that `validate --emit-attestation` wrote (holds
+	// summary-bundle/ and is where pointer.yaml is written) or the
+	// summary-bundle/ directory itself.
+	BundleDir string
+
+	// Push is the OCI reference the summary bundle is pushed to. Unlike
+	// Emit, Push is required: a publish with nothing to push is a no-op.
+	Push        string
+	PlainHTTP   bool
+	InsecureTLS bool
+
+	// AICRVersion is stamped into the pushed OCI manifest's annotations.
+	// It does not alter the signed predicate, which is read verbatim from
+	// the bundle's statement.intoto.json — the bundle bytes (and their
+	// baked-in attestedAt timestamp) are signed as-is.
+	AICRVersion string
+
+	// OIDCResolve configures keyless-signing token resolution. Resolution
+	// is deferred until adjacent to SignStatement so Fulcio's
+	// nonce-binding window is respected.
+	OIDCResolve bundleattest.ResolveOptions
+}
+
+// Publish signs and pushes an already-emitted recipe-evidence v1 bundle,
+// then writes pointer.yaml beside it. It is the off-network second leg of
+// the workflow whose first leg is `aicr validate --emit-attestation`
+// (no --push): that step produces the unsigned on-disk bundle this one
+// consumes.
+//
+// The output is identical to the one-shot `validate --emit-attestation
+// --push` path: the predicate signed here is read verbatim from the
+// bundle's statement.intoto.json, so the timestamp baked at emit time is
+// preserved and the resulting signed artifact is content-identical
+// regardless of which host ran which leg.
+//
+// The *EmitResult return mirrors Emit for the future API surface: it carries
+// PushSummary's ref+digest and the Sign/Rekor info an API handler would
+// serialize. The success path that populates it requires a live push + Fulcio
+// sign, which unit tests can't exercise, so no in-repo caller consumes it yet.
+//
+//nolint:unparam // result is returned for the future API surface; see above.
+func Publish(ctx context.Context, opts PublishOptions) (*EmitResult, error) {
+	if opts.Push == "" {
+		return nil, errors.New(errors.ErrCodeInvalidRequest, "push reference is required for publish")
+	}
+	// Validate the ref up front so a malformed reference doesn't waste a
+	// Fulcio cert + Rekor inclusion proof on a sign the push would reject.
+	if _, err := oci.ParseOutputTarget(opts.Push); err != nil {
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid push reference", err)
+	}
+
+	bundle, outDir, err := loadOnDiskBundle(opts.BundleDir)
+	if err != nil {
+		return nil, err
+	}
+
+	slog.Info("publishing evidence bundle",
+		"summaryDir", bundle.SummaryDir,
+		"recipe", bundle.RecipeName,
+		"push", opts.Push)
+
+	out, err := signAndPush(ctx, bundle, signPushOptions{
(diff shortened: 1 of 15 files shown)


pkg/evidence/attestation/publish.go
```diff
@@ -0,0 +1,240 @@
...
+	clean := filepath.Clean(dir)
+	if HasBundleMarkers(clean) {
+		return clean, filepath.Dir(clean), nil
+	}
+	candidate := filepath.Join(clean, SummaryBundleDirName)
+	if HasBundleMarkers(candidate) {
+		return candidate, clean, nil
+	}
+	return "", "", errors.New(errors.ErrCodeInvalidRequest,
+		"directory "+dir+" does not look like an evidence bundle "+
```
mchmarny: nit: vocab drift with the verifier — `verifier/fetch.go` `materializeDir` says "does not look like a summary bundle" for the same condition. Worth aligning on one term (probably "summary bundle" since that's what the directory is called everywhere else) so users debugging an "X does not look like a Y bundle" error don't see two different Y's depending on which command surfaced it.

pkg/evidence/attestation/publish.go
```diff
@@ -0,0 +1,240 @@
...
+// bundle's statement.intoto.json, so the timestamp baked at emit time is
+// preserved and the resulting signed artifact is content-identical
+// regardless of which host ran which leg.
+//
+// The *EmitResult return mirrors Emit for the future API surface: it carries
+// PushSummary's ref+digest and the Sign/Rekor info an API handler would
+// serialize. The success path that populates it requires a live push + Fulcio
+// sign, which unit tests can't exercise, so no in-repo caller consumes it yet.
+//
+//nolint:unparam // result is returned for the future API surface; see above.
```
mchmarny: nit: the `*EmitResult` return + `nolint:unparam` is well-justified by the comment, but this kind of "for the future API" return tends to decay if the API doesn't land. Worth filing a tracking issue (or dropping the return to `error` now and reintroducing it when an actual caller exists) so the lint suppression doesn't outlive its justification.

mchmarny (dismissed):
Clean architectural split — splits the one-shot `validate --emit-attestation --push` into a cluster-bound leg and a Sigstore-bound leg without changing the signed artifact, which is exactly the right shape for the corporate-VPN-vs-Sigstore-egress problem. The `signPushOptions` extraction (so `Emit` and `Publish` drive the same pipeline without leaking build-only fields), the `oidcResolveOptionsFromFlags` consolidation, the shared `HasBundleMarkers`, and the flag-constant extraction are all good DRY moves that the new command motivated rather than adding incidentally. Up-front push-ref validation (saving a wasted Fulcio cert + Rekor inclusion proof on a malformed ref) and the bounded statement read against `MaxAttestationFileBytes` match the project's established patterns. CI is green and the description documents a live EKS GB200 end-to-end run.

Two nits inline, both optional — neither blocks merge.

pkg/evidence/attestation/publish.go
njhensley: Good catch — aligned the wording to "summary bundle" (matching `materializeDir`) so an "X does not look like a Y bundle" error reads consistently no matter which command surfaced it. Updated the test substring to match. In the next push.

pkg/evidence/attestation/publish.go
njhensley: Agreed — a "for the future API" return plus a `nolint:unparam` is exactly the kind of suppression that outlives its justification. Took your second option: dropped `Publish` to returning just `error` and removed the suppression. `Emit` keeps `*EmitResult` since its result is actually exercised by tests, so the asymmetry now reflects real usage rather than speculation — trivial to reintroduce on `Publish` when a real caller lands. In the next push.

mchmarny (approved):

Issue #1153, opened 2026-06-02
feat(verify): private Sigstore trust root for bundle verification


lockwobr:
## Problem

Private Sigstore signing (#408) lets an org sign against a self-hosted Fulcio/Rekor, but `aicr verify` can only verify against the public-good trusted root. `verifySigstoreBundle` calls `trust.GetTrustedMaterial()` (`pkg/bundler/verifier/trust.go`), which has no override for a private CA root or a custom TUF trust root. So privately-signed bundles fail verification.

## Proposal

Allow `aicr verify` to load a custom Sigstore trusted root so bundles signed against private Fulcio/Rekor verify against the matching root of trust.

## Behavior

- A flag (e.g. `--trust-root <path>` / `--certificate-chain`) or a TUF root URL supplies the private root of trust to the sigstore-go verifier.
- Public-good remains the default when unset.
- Composes with the keyless identity-pinning path (the identity matcher is unchanged; only the root material differs).

## Dependencies

- Pairs with: #408 (private Fulcio/Rekor URLs). This is its verification counterpart.

## Success criteria

- A bundle signed via `aicr bundle --fulcio-url <private> --rekor-url <private>` verifies with the corresponding private trust root.
- Public Sigstore verification is unchanged when no custom root is supplied.
- `make test` / `make lint` pass.

Issue #1149, opened 2026-06-02
[Epic]: Closed supply chain (V1) — enterprise signing and verification


lockwobr:
## Goal

Deliver the V1 "Closed Supply Chain" acceptance bar for AICR bundles and recipe data:

> every artifact (binary, image, recipe data, bundle) has retrievable provenance and a documented verification path that works in air-gapped, private, and public-trust deployments

This epic tracks the two halves that are not yet shipped — **signing** in enterprise environments and the matching **verification** path — plus **recipe-data provenance** and the **end-user verification guide**. (Build-time provenance for binaries and images already ships.)

Signing lives in `pkg/bundler/attestation/` (`SignStatement` in `signing.go`, the attester resolver in `resolver.go`) and uses sigstore-go's `sign` package. Verification lives in `pkg/bundler/verifier/` (`aicr verify` → `verifySigstoreBundle`), which today only checks keyless bundles against the public-good trusted root (`trust.GetTrustedMaterial()`).

## Signing sub-issues

| # | Title | Size | Notes |
|---|-------|------|-------|
| #408 | Private Sigstore via `--fulcio-url` / `--rekor-url` | **Small** | Library already plumbs `SignOptions.FulcioURL`/`RekorURL`; CLI wiring only. |
| #685 | OIDC token caching across runs | **Medium** | Iteration ergonomics; **not V1-blocking** — once #407 lands its audience is local-dev. Standalone. |
| #407 | KMS-backed signing via `--signing-key` | **Large** | Foundation. Key-based signer alongside the keyless path; pulls KMS-provider deps under sigstore-go. |
| #409 | Air-gapped signing via `--tlog-upload=false` | **Small (gated)** | One flag + conditional Rekor step + validation matrix; meaningful once #407 lands. |
| #1150 | Server-side bundle attestation for `/v1/bundle` | **Large** | `/v1/bundle` does not sign today; needs a non-interactive identity — **KMS, request-supplied identity token, or ambient OIDC** (browser OIDC can't work in a daemon). Gated on #407. |

## Verification & provenance sub-issues

Signing without a matching verification path breaks the chain at the consumer. Each signing mode needs its verify counterpart:

| Signing mode | Verification counterpart |
|---|---|
| KMS-backed signing (#407) | #1152 — verifier `aicr verify --key <kms-uri>` |
| Private Fulcio/Rekor (#408) | #1153 — verifier custom/private trust root |
[...]

PR #1158, opened 2026-06-02
feat(bundle): private Sigstore via --fulcio-url and --rekor-url

### LINKED ISSUE #408: feat(bundle): private Sigstore infrastructure support via --fulcio-url and --rekor-url
## Problem

Organizations running their own Fulcio CA and Rekor transparency log cannot direct AICR to use their private Sigstore instances. All signing currently targets the public Sigstore infrastructure, which may not meet internal compliance or network isolation requirements.

## Proposal

Add `--fulcio-url` and `--rekor-url` flags to `aicr bundle` (and/or `aicr bundle-sign`) to override the default public Sigstore endpoints.

### New Flags

| Flag | Type | Default |
|------|------|---------|
| `--fulcio-url` | string | Public Fulcio (`https://fulcio.sigstore.dev`) |
| `--rekor-url` | string | Public Rekor (`https://rekor.sigstore.dev`) |

### Behavior

- When either flag is set, AICR uses the specified endpoint(s) for signing and transparency log upload
- The verification process is identical—same cosign library, same bundle format—only the root of trust differs
- Verification against private infrastructure requires the corresponding root certificate (`--certificate-chain` or TUF root)
- Both flags are optional and independent: an org could run private Fulcio with public Rekor, or vice versa

### Validation

- URLs must be valid HTTPS endpoints
- AICR should fail fast with a clear error if the private endpoint is unreachable

## Implementation Notes

- Cosign already supports custom Fulcio/Rekor URLs via `cosign sign --fulcio-url` and `--rekor-url`
- Pass these through to the cosign signing options in the existing signing path
- No changes to bundle format or verification logic needed

## Success Criteria

- `aicr bundle --fulcio-url https://fulcio.internal.example.com --rekor-url https://rekor.internal.example.com` signs against private infrastructure
- Bundle is verifiable with matching private root of trust
- Public Sigstore remains the default when flags are omitted
- `make test` and `make lint` pass with no regressions

lockwobr:
## Summary

Adds `--fulcio-url` / `--rekor-url` flags (and matching config fields) to `aicr bundle` so `--attest` keyless signing can target a private Sigstore instance instead of the public-good endpoints.

## Motivation / Context

Organizations running their own Fulcio CA and/or Rekor transparency log could not point AICR at them — all signing hit public Sigstore. The signing library already accepted custom endpoints (`SignOptions.FulcioURL`/`RekorURL` with default fallback); this PR connects the CLI and config layers so the capability is actually reachable.

Fixes: #408
Related: #1149

## Type of Change

## Component(s) Affected

- [x] Core libraries (`pkg/errors`, `pkg/k8s`) — `pkg/defaults`

## Implementation Notes

- `ResolveOptions` carries `FulcioURL`/`RekorURL`; `NewKeylessAttester(token, fulcio, rekor)` applies the public-good default on empty, so existing callers are unaffected. Both the eager and lazy resolver paths thread the URLs to `SignStatement`.
- Config: `spec.bundle.attestation.fulcioURL`/`rekorURL` → `BundleResolved`.
- CLI validates both flags at parse time as absolute `https://` URLs (fail fast, before bundling), via a shared `config.ValidateHTTPSURL(label, raw)` helper that lives next to the sibling `ValidateAppName`.
- Moved the Sigstore default endpoints out of `pkg/bundler/attestation` into `pkg/defaults` as `SigstoreFulcioURL` / `SigstoreRekorURL` (matching the existing `SigstoreSignTimeout` convention) and updated all call sites, including `pkg/evidence/attestation/emit.go`.
- The two URLs are independent (private Fulcio + public Rekor, or vice versa); public Sigstore is the default when omitted.
- Verifier support for privately-signed bundles is intentionally out of scope here and tracked under epic #1149 (#1153).

## Testing

```bash
make qualify
```

`make qualify` passes end-to-end: unit tests (`-race`), golangci-lint (0 issues), chainsaw e2e (22 passed / 0 failed), vulnerability scan (no vulnerabilities), license-header and repo checks. New table-driven tests cover `NewKeylessAttester` URL defaulting, resolver URL propagation (eager + lazy), config resolution, CLI flag parsing + HTTPS-validation rejection, and the shared `ValidateHTTPSURL`.

## Risk Assessment

- [x] **Low** — Additive flags; default behavior (public Sigstore) is unchanged when the flags are omitted. The constant move is mechanical with all call sites updated and covered by the full gate.

**Rollout notes:** Backwards compatible. No migration. `--fulcio-url`/`--rekor-url` are opt-in and only take effect with `--attest`.

## Checklist

### DIFF
--- a/pkg/bundler/config/config.go
+++ b/pkg/bundler/config/config.go
@@ -16,6 +16,7 @@ package config
 
 import (
 	"fmt"
+	"net/url"
 	"sort"
 	"strings"
 
@@ -103,6 +104,32 @@ func ValidateAppName(name string) error {
 		return errors.New(errors.ErrCodeInvalidRequest,
 			fmt.Sprintf("invalid app name %q: must be a DNS-1123 subdomain (%s)",
 				name, strings.Join(errs, "; ")))
+	}
+	return nil
+}
+
+// ValidateHTTPSURL reports whether raw is a usable absolute https:// endpoint.
+// The empty string is allowed (callers treat it as "use the default"); any
+// non-empty value must parse as an absolute URL with an https scheme and a
+// host, and must not embed credentials. label names the field for the error
+// message (e.g. "fulcio URL"). Used to fail malformed signing endpoints at
+// bundle/parse time rather than at sign time. See #408.
+func ValidateHTTPSURL(label, raw string) error {
+	if raw == "" {
+		return nil
+	}
+	u, err := url.Parse(raw)
+	if err != nil {
+		return errors.Wrap(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("invalid %s %q", label, raw), err)
+	}
+	// Reject embedded credentials: url.Parse stashes "user:pass@" in u.User
+	// while leaving Scheme/Host intact, so a scheme+host-only check would
+	// otherwise accept "https://user:pass@host". Credentials have no place in
+	// a signing endpoint and would leak via config/flags/process listings.
+	if u.Scheme != "https" || u.Host == "" || u.User != nil {
+		return errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("invalid %s %q: must be an absolute https:// URL without embedded credentials", label, raw))
 	}
 	return nil
 }
--- a/pkg/cli/bundle.go
+++ b/pkg/cli/bundle.go
@@ -86,6 +86,12 @@ type bundleCmdOptions struct {
 	// (RFC 8628) for headless hosts where a browser callback is unavailable.
 	oidcDeviceFlow bool
 
+	// fulcioURL and rekorURL override the public-good Sigstore endpoints so
+	// keyless signing targets a private Fulcio CA and/or Rekor transparency
+	// log. Empty leaves the public defaults in place. See #408.
+	fulcioURL string
+	rekorURL  string
+
 	// OCI output reference (nil if outputting to local directory)
 	ociRef        *oci.Reference
 	plainHTTP     bool
@@ -110,7 +116,7 @@ type bundleCmdOptions struct {
 // config-supplied fields happens at the conversion boundary in Resolve;
 // errors from that boundary carry "spec.bundle.<path>" attribution.
 //
-//nolint:gocyclo // option resolution is inherently long but linear
+//nolint:gocyclo,funlen // option resolution is inherently long but linear
 func parseBundleCmdOptions(cmd *cli.Command, cfg *appcfg.AICRConfig) (*bundleCmdOptions, error) {
 	resolved, err := cfg.Bundle().Resolve()
 	if err != nil {
@@ -126,6 +132,8 @@ func parseBundleCmdOptions(cmd *cli.Command, cfg *appcfg.AICRConfig) (*bundleCmd
 		certificateIdentityRegexp: stringFlagOrConfig(cmd, "certificate-identity-regexp", resolved.CertIDRegexp),
 		identityToken:             cmd.String(flagIdentityToken),
 		oidcDeviceFlow:            boolFlagOrConfig(cmd, flagOIDCDeviceFlow, resolved.OIDCDeviceFlow),
+		fulcioURL:                 stringFlagOrConfig(cmd, flagFulcioURL, resolved.FulcioURL),
+		rekorURL:                  stringFlagOrConfig(cmd, flagRekorURL, resolved.RekorURL),
 		insecureTLS:               boolFlagOrConfig(cmd, flagInsecureTLS, resolved.InsecureTLS),
 		plainHTTP:                 boolFlagOrConfig(cmd, flagPlainHTTP, resolved.PlainHTTP),
 		imageRefsPath:             stringFlagOrConfig(cmd, "image-refs", resolved.ImageRefs),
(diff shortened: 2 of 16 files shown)


pkg/cli/bundle.go
```diff
@@ -567,6 +593,16 @@ Package with explicit tag (overrides CLI version):
 				Sources:  cli.EnvVars("AICR_OIDC_DEVICE_FLOW"),
 				Category: catDeployment,
 			},
+			&cli.StringFlag{
+				Name:     flagFulcioURL,
+				Usage:    "Override the Fulcio CA URL for --attest keyless signing (e.g. a private Sigstore instance). Must be an HTTPS URL. Defaults to the public-good Fulcio.",
+				Category: catDeployment,
+			},
+			&cli.StringFlag{
```
mchmarny, line 602: The two new flags don't take `Sources: cli.EnvVars(...)`, but the sibling `--oidc-device-flow` (a few lines above) does — `AICR_OIDC_DEVICE_FLOW`. For CI / automation targeting a private Sigstore that's a real ergonomic gap: every invocation has to thread the URLs explicitly rather than setting them once in the environment. Suggest `AICR_FULCIO_URL` / `AICR_REKOR_URL` for parity with the existing convention (and matching the long-standing Cosign analogs `COSIGN_FULCIO_URL` / `COSIGN_REKOR_URL`). Docs table in `cli-reference.md` would need the env column populated to match.

pkg/cli/bundle.go
```diff
@@ -567,6 +593,16 @@ Package with explicit tag (overrides CLI version):
...
+				Name:     flagRekorURL,
+				Usage:    "Override the Rekor transparency-log URL for --attest keyless signing (e.g. a private Sigstore instance). Must be an HTTPS URL. Defaults to the public-good Rekor.",
```
mchmarny: nit: usage says "Must be an HTTPS URL" but `ValidateHTTPSURL` also rejects embedded credentials, so the runtime error users see ("must be an absolute https:// URL without embedded credentials") doesn't match what `--help` advertises. Worth syncing the wording on both flags (and the docs blurb) so a user pasting `https://user:pass@…` isn't surprised by a rejection that the help text didn't warn them about.

mchmarny (commented):
CodeRabbit already owns the three open items (verifier-support wording, `validateSingleValueFlags` gap, GoDoc placement) plus the resolved ones, so I'll skip those.

Two additional points below. The env var gap is the substantive one; the usage-string mismatch is a nit.

pkg/bundler/config/config.go
```diff
@@ -107,6 +108,32 @@ func ValidateAppName(name string) error {
...
+	u, err := url.Parse(raw)
+	if err != nil {
+		return errors.Wrap(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("invalid %s %q", label, raw), err)
+	}
+	// Reject embedded credentials: url.Parse stashes "user:pass@" in u.User
+	// while leaving Scheme/Host intact, so a scheme+host-only check would
+	// otherwise accept "https://user:pass@host". Credentials have no place in
+	// a signing endpoint and would leak via config/flags/process listings.
+	if u.Scheme != "https" || u.Host == "" || u.User != nil {
```
mchmarny, line 130: Nice — the `u.User != nil` check (with the comment explaining that `url.Parse` stashes `user:pass@` in `u.User` while leaving Scheme/Host valid) is exactly the right guard. A scheme+host check alone would have let `https://user:pass@host` through and leaked creds via config/flags/process listings. No action needed.

pkg/evidence/attestation/emit.go
```diff
@@ -261,10 +261,15 @@ func signAndPush(ctx context.Context, bundle *Bundle, opts signPushOptions) (emi
...
+	// public Sigstore. Empty values fall back to the public-good defaults
+	// inside SignStatement. (The validate-emit path does not yet expose
+	// --fulcio-url/--rekor-url, so these are empty today; this keeps the
+	// signer from overriding a configured endpoint once it does.)
 	signRes, err := bundleattest.SignStatement(signCtx, artifactStmt, bundleattest.SignOptions{
 		OIDCToken: token,
-		FulcioURL: bundleattest.DefaultFulcioURL,
-		RekorURL:  bundleattest.DefaultRekorURL,
+		FulcioURL: opts.OIDCResolve.FulcioURL,
+		RekorURL:  opts.OIDCResolve.RekorURL,
```
mchmarny, line 272: The "validate-emit path does not yet expose `--fulcio-url`/`--rekor-url`, so these are empty today; this keeps the signer from overriding a configured endpoint once it does" framing is accurate but leaves the wire-through as an implicit TODO with no linked issue. Either reference the tracking issue (looks like #1149 / #1153 from the doc caveat) or open a follow-up and link it here so the work doesn't fall off — easy to miss in a year when someone wonders why recipe-evidence signing always hits public Sigstore.

pkg/cli/bundle.go
```diff
@@ -567,6 +593,18 @@ Package with explicit tag (overrides CLI version):
...
 				Category: catDeployment,
 			},
+			&cli.StringFlag{
+				Name:     flagFulcioURL,
+				Usage:    "Override the Fulcio CA URL for --attest keyless signing (e.g. a private Sigstore instance). Must be an absolute https:// URL with no embedded credentials. Defaults to the public-good Fulcio. Also reads AICR_FULCIO_URL.",
+				Sources:  cli.EnvVars("AICR_FULCIO_URL"),
+				Category: catDeployment,
+			},
+			&cli.StringFlag{
+				Name:     flagRekorURL,
```
mchmarny, line 603: nit: `--fulcio-url` / `--rekor-url` are accepted without `--attest=true` and silently no-op when signing is disabled. Matches the pattern of `--certificate-identity-regexp` / `--identity-token`, so it's internally consistent — but an operator who sets these expecting signing gets no feedback. Optional: `slog.Warn` in `selectAttester` when either URL is set but `opts.attest == false`. Not blocking.

mchmarny (approved):
Approving. Clean additive change: defaults preserved, flag + env + config + spec-path attribution all wired correctly, and both eager and lazy attester paths thread the URLs. The `ValidateHTTPSURL` credential-rejection check is well thought out (with a clear comment explaining why scheme+host alone isn't enough). Tests cover URL defaulting, flag-over-env precedence, HTTPS-only validation, and embedded-credential rejection. CI green.

Two low-severity notes inline plus a positive callout — nothing blocking. The doc caveat about `aicr verify` not yet supporting private trust roots is the right honest framing for the partial-coverage shape of this PR.

Issue #1161, opened 2026-06-02
bundle: support --set-json / --set-file for list and object value overrides


yuanchen8911:
### Summary

`aicr bundle --set <component>:<path>=<value>` is **scalar-only**, so it cannot express list/object value overrides. The concrete trap: `agentgateway.allowedSourceRanges` is a `[]string` rendered into the Service's `spec.loadBalancerSourceRanges`. Running:

```
aicr bundle --set agentgateway:allowedSourceRanges=216.228.127.128/30 ...
```

exits 0 but writes a **bare string** at that path; `toYaml` then emits a scalar under `loadBalancerSourceRanges`, producing a **type-invalid Service** — the field is dropped or rejected and the gateway can silently stay open at `0.0.0.0/0`. This is bad enough that the docs explicitly say *"Do NOT use `--set` for this key"* (`recipes/components/agentgateway/values.yaml`, `docs/user/component-catalog.md`).

Today the only ways to set a list override are editing the generated recipe's `componentRefs[].overrides` or authoring a `--data` overlay — neither is a clean CLI workflow.

### Proposal

Add **`--set-json`** (and/or **`--set-file`**) to `aicr bundle`, mirroring Helm's flags, so list/object overrides work from the CLI:

```
aicr bundle -r recipe.yaml \
  --set-json agentgateway:allowedSourceRanges='["216.228.127.128/30"]' ...
```

- Value is JSON-parsed and **deep-merged** into the same `ValueOverrides` → `ComponentRef.Overrides` path the recipe/bundler already use (`pkg/recipe/adapter.go` three-way merge: base → valuesFile → Overrides).
- `--set-file` reads the JSON/YAML value from a file for larger structures.
- `--set` stays as the scalar convenience; `--set-json` covers lists/maps.

### Rationale

1. **Closes the silent-`--set`-list trap** — operators get a real YAML list instead of a bare string, so a lockdown attempt can't silently leave the gateway open.
2. **Unblocks a clean lockdown workflow** — scope the inference-gateway (or any list/object field) to trusted CIDRs from the CLI without hand-editing the generated recipe.
3. General-purpose: applies to every list/object override, not just `allowedSourceRanges`.

### Pointers

- `--set` wiring: `pkg/cli/bundle.go` (StringSliceFlag `set`, `resolveComponentPaths(cmd, "set", ...)`, `config.ParseValueOverrides`)
- Merge target: `pkg/recipe/adapter.go` (`Overrides` highest precedence)

[...]

PR #1163, opened 2026-06-02
feat(bundler): warn on open agentgateway inference-gateway exposure

### LINKED ISSUE #1160: agentgateway: surface open inference-gateway exposure (bundle warning + conformance check)
### Summary

The `agentgateway` `inference-gateway` is provisioned as a **public `LoadBalancer`** (port 80, HTTP, unauthenticated). Its `allowedSourceRanges` defaults to `[]`, which — by deliberate design (#1138) — leaves the Service open to `0.0.0.0/0`. The default itself is defensible (a baked-in CIDR would firewall every downstream operator to NVIDIA's network), but the **exposure is silent**: nothing in `aicr bundle` output or validation surfaces that the inference gateway is internet-facing. An operator only discovers it by manually inspecting the live Service.

Discovered during GKE H100 inference e2e validation: the deployed `inference-gateway` had a public IP with `loadBalancerSourceRanges` empty (`0.0.0.0/0`), found only by hand.

### Proposal (two guardrails — keep open-by-default, remove the *silence*)

1. **Bundle-time warning** — when the `agentgateway` component is included with an empty `allowedSourceRanges`, emit a warning in `aicr bundle` output, e.g.:
   > ⚠ inference-gateway will be provisioned as an internet-facing LoadBalancer open to `0.0.0.0/0`. Scope it to trusted networks via `agentgateway.allowedSourceRanges` (recipe componentRef override). See docs/user/component-catalog.md.

   This mirrors the existing kube-prometheus-stack PVC `storageClassName` warning already printed by the bundler.

2. **Conformance/security check** — a validator check that, on a live cluster, flags an `inference-gateway` (or any agentgateway-managed) `LoadBalancer` Service whose `spec.loadBalancerSourceRanges` is empty while `type=LoadBalancer`. Reports the open exposure as a finding (warn/fail per policy).

### Rationale / scope

- Preserves the intentional open-by-default UX from #1138 (zero-config external reachability) while making the exposure a **conscious decision** rather than a silent default.
- **Non-goals:** (a) baking a CIDR into the shared overlay/default — would lock external operators out of their own gateway; (b) flipping the default to internal/ClusterIP — a stronger posture worth considering, but a separate maintainer decision with broad blast radius.

### Pointers

- Default + rationale: `recipes/components/agentgateway/values.yaml` (`allowedSourceRanges: []`)
- Render path: `recipes/components/agentgateway/manifests/inference-gateway.yaml` (`loadBalancerSourceRanges` under `AgentgatewayParameters.spec.service.spec`)
- Bundle warning precedent: kube-prometheus-stack PVC warning in `pkg/bundler` output

---

**Addressed by:** #1163 (bundle-time warning + conformance finding). Related: #1161 / #1162 (`--set-json` CLI flag for scoping `allowedSourceRanges`).

yuanchen8911:
## Summary

Surfaces the agentgateway `inference-gateway`'s open-by-default network exposure with two guardrails — a bundle-time warning and a conformance finding — without changing the intentional open-by-default behavior.

## Motivation / Context

The agentgateway `inference-gateway` is provisioned as a public `LoadBalancer` (HTTP, unauthenticated) whose `allowedSourceRanges` defaults to empty, leaving it open to `0.0.0.0/0` (#1138). The default is defensible — a baked-in CIDR would firewall every downstream operator to NVIDIA's network — but the exposure was **silent**: nothing in `aicr bundle` output or validation flagged that the inference endpoint is internet-facing. An operator only found it by inspecting the live Service. This keeps open-by-default while removing the silence.

Fixes: #1160
Related: #1161 / #1162 — now **merged** (`--set-json` / `--set-file` are the clean CLI mechanism for scoping `allowedSourceRanges` after these guardrails fire). This PR is rebased on top of #1162.

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Validator (`pkg/validator`, `validators/conformance`)

## Implementation Notes

**1. Bundle-time warning (`pkg/bundler`).** When a bundle includes `agentgateway` with an empty `allowedSourceRanges`, `aicr bundle` emits a non-blocking warning that the inference-gateway will be open to `0.0.0.0/0`, with remediation. It inspects the **merged** component values (mirroring the existing `storageClassName` PVC warning), so it catches the empty state regardless of how the value was (or wasn't) overridden. A non-list value (e.g. a bare string from a mistaken scalar `--set`) also triggers the warning, since that itself renders an invalid Service.

**2. Conformance finding (`validators/conformance`).** Rather than a new check requiring edits to 13 inference overlays + the catalog, this extends the existing `inference-gateway` conformance check — already wired into every inference overlay. It lists the `LoadBalancer` Service(s) in `agentgateway-system` and records their exposure as evidence: the source ranges if scoped, or an explicit "open to `0.0.0.0/0`" finding if not. Open-by-default stays a **non-fatal warning** (preserving #1138); setting `AICR_REQUIRE_SCOPED_INFERENCE_GATEWAY=true` on the validator environment escalates an open gateway to a check **failure** (fail-closed policy). This matches the issue's "warn/fail per policy."

**Non-goals (per the issue):** not baking a CIDR into the shared overlay (would lock external operators out), and not flipping the default to internal/ClusterIP (a separate maintainer decision).

## Testing

```bash
make test-coverage   # 76.4% total (threshold 75%)
make lint            # golangci-lint 0 issues, license headers, MDX-safe docs, agents-sync
make e2e             # 22/22 chainsaw tests pass (incl. cli-bundle-agentgateway-templates)
```

New tests: `pkg/bundler` (`warnAgentgatewayOpenExposure`: empty / empty-list / bare-string / scoped / absent), `validators/conformance` (`assessGatewayExposure`: open→warn, open+policy→fail, scoped→pass, non-LB ignored, nil clientset; `envTruthy`).

Verified end-to-end with a locally built binary against a generated EKS/H100/dynamo inference recipe:

```
$ aicr bundle -r recipe.yaml -o ./bundle
  ⚠ Warning: agentgateway: inference-gateway will be provisioned as an internet-facing
    LoadBalancer open to 0.0.0.0/0 (allowedSourceRanges is empty). Scope it to trusted
    networks via a recipe componentRef override or --set-json agentgateway:allowedSourceRanges='["<cidr>"]'.
```

`make scan` (grype) was not run: no new dependencies were introduced, so the vulnerability surface is unchanged.

## Risk Assessment

- [x] **Low** — Additive. The bundle warning is non-blocking; the conformance finding defaults to non-fatal (no change to existing pass/fail behavior unless the new opt-in env var is set). Open-by-default is preserved.

**Rollout notes:** Backwards compatible. `AICR_REQUIRE_SCOPED_INFERENCE_GATEWAY` is opt-in and defaults to off.

## Checklist


docs/user/cli-reference.md
```diff
@@ -1198,6 +1198,8 @@ When `--storage-class` is not set, any `storageClassName` values already defined
 
 If a rendered component creates a PVC at a registry-declared `storageClassPaths` entry and no usable `storageClassName` is set after overlay, `--storage-class`, and `--set` precedence is resolved, `aicr bundle` emits a non-blocking warning. The bundle still relies on the target cluster's default StorageClass in that case.
 
+Similarly, when a bundle includes the `agentgateway` component with an empty `allowedSourceRanges`, `aicr bundle` emits a non-blocking warning that the inference-gateway will be provisioned as an internet-facing `LoadBalancer` open to `0.0.0.0/0`. Scope it to trusted CIDRs via a recipe `componentRef` override or `--set-json agentgateway:allowedSourceRanges='["<cidr>"]'`. See [Inference Gateway Network Exposure](component-catalog.md#inference-gateway-network-exposure).
```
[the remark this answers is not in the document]
yuanchen8911: Correct that `--set-json` isn't on `main` yet — it's the flag added by #1162 (in review). This doc reference becomes valid once #1162 lands; the plan is to merge #1162 before #1163. Keeping the `--set-json` guidance (rather than dropping it) so the docs are complete the moment the flag ships.

pkg/bundler/bundler.go
```diff
@@ -1037,6 +1039,63 @@ func (b *DefaultBundler) warnMissingStorageClassForPVCs(ctx context.Context, rec
...
+		return
+	}
+
+	msg := fmt.Sprintf(
+		"%s: inference-gateway will be provisioned as an internet-facing LoadBalancer open to 0.0.0.0/0 "+
+			"(%s is empty). Scope it to trusted networks via a recipe componentRef override or "+
+			"--set-json %s:%s='[\"<cidr>\"]'. See docs/user/component-catalog.md.",
+		agentgatewayComponentName, agentgatewaySourceRangesPath,
+		agentgatewayComponentName, agentgatewaySourceRangesPath,
+	)
```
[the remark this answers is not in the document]
yuanchen8911: Same as the cli-reference note: the `--set-json` remediation in this warning string is the flag added by #1162. #1162 will merge before #1163, so the warning points at a real flag by the time this lands. (If merge order slips, we'll soften the string to the recipe `componentRef` override, which works today.)

validators/conformance/inference_gateway_check.go
```diff
@@ -137,9 +146,104 @@ func CheckInferenceGateway(ctx *validators.Context) error {
...
+		}
+		lbCount++
+		if len(svc.Spec.LoadBalancerSourceRanges) == 0 {
+			openSvcs = append(openSvcs, svc.Name)
+			fmt.Fprintf(&summary, "%-40s type=LoadBalancer sourceRanges=<empty> (OPEN to 0.0.0.0/0)\n", svc.Name)
+		} else {
+			fmt.Fprintf(&summary, "%-40s type=LoadBalancer sourceRanges=%s\n",
+				svc.Name, strings.Join(svc.Spec.LoadBalancerSourceRanges, ","))
+		}
+	}
```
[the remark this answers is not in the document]
yuanchen8911: Skipping — intentional, with reason. Every API call in this check uses `ctx.Ctx` directly (8/8; zero `context.WithTimeout` in the file), and `ctx.Ctx` is already deadline-bounded by the validator framework (`validators/context.go` → `AICR_CHECK_TIMEOUT` / `defaults.CheckExecutionTimeout`). So the `Services().List` is bounded and an unbounded hang isn't possible. Adding a per-call timeout + `ctx.Done()` loop guard to only this new function would diverge from the file's established pattern; if we want per-call timeouts here, that's a separate file-wide cleanup rather than something to introduce in this PR.

validators/conformance/inference_gateway_exposure_test.go
```diff
@@ -0,0 +1,138 @@
...
+		{"0", false},
+		{"garbage", false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.val, func(t *testing.T) {
+			if tt.val != "" {
+				t.Setenv("AICR_TEST_ENV_TRUTHY", tt.val)
+			}
+			got := envTruthy("AICR_TEST_ENV_TRUTHY")
+			if got != tt.want {
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 8772e223 — `t.Setenv("AICR_TEST_ENV_TRUTHY", tt.val)` is now called unconditionally for every case (including the empty string), so each subtest fully controls the var and can't read an ambient value.

yuanchen8911:
**Stacked on #1162** — this PR sits on top of `feat/bundle-set-json` (#1162) and should merge **after** it.

The open-exposure remediation points users at `--set-json agentgateway:allowedSourceRanges='["<cidr>"]'`, which is the flag added by #1162. The branch is rebased onto #1162's head, so that commit currently shows in this PR's diff.

Because this is a cross-fork PR, the base can't be retargeted to the fork branch `feat/bundle-set-json` (GitHub requires the base to be a branch on `NVIDIA/aicr`). So the base stays `main`. Once #1162 merges, this PR's diff collapses to just its own commit automatically — no rebase needed.

docs/user/cli-reference.md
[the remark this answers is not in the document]
yuanchen8911: Superseded by #1162, which merged as `f0490fbf` — `aicr bundle` now **does** support `--set-json` / `--set-file` for list/object values, so the `--set-json agentgateway:allowedSourceRanges='["<cidr>"]'` remediation is valid and is exactly the intended scoping mechanism (the earlier learning that there is no `--set-json` support no longer holds). This PR is rebased on top of #1162, and the flag is now cross-linked to the [List and Object Value Overrides](cli-reference.md#list-and-object-value-overrides) section. Keeping the remediation and resolving this thread.

pkg/bundler/bundler.go
```diff
@@ -1108,6 +1111,102 @@ func (b *DefaultBundler) warnMissingStorageClassForPVCs(ctx context.Context, rec
...
+// 0.0.0.0/1 + 128.0.0.0/1) — a deliberate-evasion case nobody hits by accident,
+// left as a documented limitation rather than expanded into range-union math.
+func isAnySourceCIDR(cidr string) bool {
+	_, ipNet, err := net.ParseCIDR(strings.TrimSpace(cidr))
+	if err != nil {
+		return false
+	}
+	ones, _ := ipNet.Mask.Size()
+	return ones == 0
+}
```
mchmarny: This `isAnySourceCIDR` is byte-for-byte identical to the one in [`validators/conformance/inference_gateway_check.go:279-286`](https://github.com/NVIDIA/aicr/blob/a9749fa91b20e75ba0133327a69f8af473b9a82f/validators/conformance/inference_gateway_check.go#L279-L286), including the multi-line "Limitation: ... range-union math" comment.

Per `AGENTS.md` anti-pattern (*"Duplicate K8s utilities across packages — Use shared utilities"*): pull this into a shared package (e.g., `pkg/util/net` or alongside `pkg/component/values`). Both call sites then import it, and the documented `/0` limitation only needs revisiting in one place if the policy ever tightens.

The validator binary is `package main` but already imports `github.com/NVIDIA/aicr/pkg/errors` and `pkg/validators`, so importing a `pkg/util/net` helper is no different.

pkg/bundler/bundler.go
```diff
@@ -1108,6 +1111,102 @@ func (b *DefaultBundler) warnMissingStorageClassForPVCs(ctx context.Context, rec
 	return nil
 }
 
+// agentgateway component name and the value path that renders into the
+// inference-gateway Service's spec.loadBalancerSourceRanges.
+const (
+	agentgatewayComponentName    = "agentgateway"
```
mchmarny, line 1117: nit: the validator side has [`TestEmbeddedCatalog_InferenceGatewayEntryExists`](https://github.com/NVIDIA/aicr/blob/a9749fa91b20e75ba0133327a69f8af473b9a82f/pkg/validator/catalog/catalog_test.go#L1095-L1109) explicitly to catch a silent no-op if the catalog entry is renamed — same risk applies here for the bundler. If `recipes/registry.yaml` ever renames `agentgateway` (or someone bumps to a chart whose values use a different key than `allowedSourceRanges`), `warnAgentgatewayOpenExposure` quietly stops firing.

Worth an analogous test that loads the registry and asserts an entry named `"agentgateway"` exists (mirrors the catalog test). The `allowedSourceRanges` chart-key drift is harder to guard against without running the chart, so a doc comment pointing back to the upstream chart values.yaml would help future-you.

validators/conformance/inference_gateway_check.go
```diff
@@ -137,9 +147,155 @@ func CheckInferenceGateway(ctx *validators.Context) error {
...
+func assessGatewayExposure(ctx *validators.Context) error {
+	if ctx.Clientset == nil {
+		return errors.New(errors.ErrCodeInvalidRequest,
+			"kubernetes client is not available for gateway exposure assessment")
+	}
+
+	svcs, err := ctx.Clientset.CoreV1().Services(agentgatewayNamespace).List(ctx.Ctx, metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(errors.ErrCodeInternal, "failed to list Services in agentgateway-system", err)
+	}
```
mchmarny, line 198: A transient apiserver error here fails the entire inference-gateway check, even when the data-plane portion already passed. Compare with [`validateGatewayDataPlane:319-329`](https://github.com/NVIDIA/aicr/blob/a9749fa91b20e75ba0133327a69f8af473b9a82f/validators/conformance/inference_gateway_check.go#L319-L329) which is more permissive about a missing Gateway resource.

Fail-closed is defensible (can't assess exposure → don't pretend it's safe), but the asymmetry with `validateGatewayDataPlane` is worth a comment so a future reader knows it was intentional. If you'd rather degrade like `validateGatewayDataPlane` does, you'd switch to a `recordRawTextArtifact` describing "could not assess exposure: <err>" and `return nil`.

validators/conformance/inference_gateway_check.go
```diff
@@ -137,9 +147,155 @@ func CheckInferenceGateway(ctx *validators.Context) error {
...
+			continue
+		}
+		// Only the inference-gateway proxy Service is in scope. Mirror the
+		// EndpointSlice readiness filter (kubernetes.io/service-name contains
+		// "inference-gateway") so a co-located LoadBalancer — controller
+		// manager, webhook — is neither mislabeled as the gateway nor able to
+		// fail the check under enforce mode. See #1160.
+		if !strings.Contains(svc.Name, "inference-gateway") {
+			continue
+		}
```
mchmarny: nit: `strings.Contains(svc.Name, "inference-gateway")` will also match e.g. `inference-gateway-controller-manager` or `inference-gateway-metrics` if either ever ships from the chart. The test for "unrelated open LoadBalancer is ignored" passes because the test fixture name (`agentgateway-controller-manager`) happens not to contain the substring — but the chart's own naming convention is what's actually being relied on here.

A label-selector List (`app.kubernetes.io/name=inference-gateway` or whatever the chart sets on the proxy Service) would be tighter. Consistent with the existing `validateGatewayDataPlane` filter, so not a regression — just flagging so the assumption is logged somewhere.

pkg/validator/v1/job_plan_test.go
```diff
@@ -360,6 +360,75 @@ func TestBuildJobPlan_ForwardsHFToken(t *testing.T) {
...
+		plan, err := BuildJobPlan(entry, "run-1", "ns", "1.0.0", "abc123", "sa", nil, nil, nil, "", "", nil)
+		if err != nil {
+			t.Fatalf("BuildJobPlan error: %v", err)
+		}
+		for _, e := range plan.Env {
+			if e.Name == requireScopedInferenceGatewayEnv {
+				values = append(values, e.Value)
+			}
+		}
+		if len(values) != 1 || values[0] != "true" {
```
mchmarny, line 425: The "catalog value cannot override forwarded value" sub-test is exactly the right test for this — a catalog-supplied value silently replacing an orchestrator-forwarded security toggle is precisely the kind of trust-boundary bug that would never get noticed without it. Good instinct mirroring the HF_TOKEN pattern.

mchmarny (commented):
Solid PR. The two-layer design — bundle-time warning + conformance finding with opt-in fail-closed escalation — is the right shape for a "open by default but make sure operators see it" guardrail, and the env-var forwarding pattern correctly mirrors the existing `HF_TOKEN` trust boundary (with a test asserting catalog entries can't override the forwarded value, which is genuinely the right paranoia).

One **medium** concern: `isAnySourceCIDR` is duplicated byte-for-byte (including the multi-line "Limitation: range-union math" comment) between `pkg/bundler/bundler.go:1201-1208` and `validators/conformance/inference_gateway_check.go:279-286`. Per `AGENTS.md` anti-pattern, this should live in a shared package. Not blocking merge — the bytes are identical today — but worth lifting out before someone tightens the `/0` policy in one place and forgets the other.

A few low / nit items inline: registry-name fragility on the bundler side (validator side has `TestEmbeddedCatalog_InferenceGatewayEntryExists`; bundler doesn't have an analog), substring-match Service filter (consistent with existing pattern), and one asymmetry vs. `validateGatewayDataPlane` on how a List failure degrades — fail-closed is defensible there, just worth a comment.

All 31 CI checks green including GPU training/inference/smoke. Coverage 76.4% above floor. Docs + catalog + tests all updated together. Nothing here blocks merge.

pkg/bundler/bundler.go
yuanchen8911: Good call. Extracted the shared helper to a new package `pkg/netutil` as `IsAnySourceCIDR` (with its own unit test); both the bundler and the validator now call it instead of carrying byte-identical copies. Fixed in 5bf8448f.

pkg/bundler/bundler.go
yuanchen8911: Added `TestAgentgatewayComponentExistsInRegistry` on the bundler side, mirroring the validator's `TestEmbeddedCatalog_InferenceGatewayEntryExists`. It asserts `registry.Get(agentgatewayComponentName)` is non-nil, so renaming the `agentgateway` registry entry now fails a test instead of silently turning the open-exposure warning into a no-op. 5bf8448f.

validators/conformance/inference_gateway_check.go
yuanchen8911: Agreed. The exposure assessment is an advisory finding layered on top of the data-plane checks that have already passed by this point, so a transient `List` error should not fail the whole check in the default mode. `assessGatewayExposure` now records the error as a best-effort artifact, warns, and passes when not enforcing; under `AICR_REQUIRE_SCOPED_INFERENCE_GATEWAY=true` it fails closed since scoping can no longer be confirmed. Added `TestAssessGatewayExposure_ListError` covering both modes (reactor-injected List failure). 5bf8448f.

validators/conformance/inference_gateway_check.go
yuanchen8911: Fixed. Added an `isInferenceGatewayProxyName` predicate that requires the `inference-gateway` substring but excludes co-located control-plane components (`controller-manager`, `webhook`, `metrics`). Applied it to both the exposure Service filter and the EndpointSlice readiness filter so the two stay in lockstep, so e.g. `inference-gateway-controller-manager` is no longer mistaken for the proxy. Added a predicate unit test plus an enforce-mode table case for the control-plane name. 5bf8448f.

pkg/validator/v1/job_plan_test.go
yuanchen8911: Thanks — agreed, that sub-test is the guard that keeps a catalog-supplied value from silently overriding the orchestrator-forwarded one. Left as-is.

yuanchen8911:
@mchmarny good review — thank you\! Addressed all five threads in 5bf8448f:

1. Extracted the duplicated `isAnySourceCIDR` into a shared `pkg/netutil.IsAnySourceCIDR` (with a unit test); bundler and validator both call it now.
2. Added `TestAgentgatewayComponentExistsInRegistry` so renaming the `agentgateway` registry entry fails a test instead of silently no-opping the warning.
3. `assessGatewayExposure` no longer fails the whole check on a transient apiserver `List` error in advisory mode — it records best-effort and passes, and fails closed only under enforce mode. Added a reactor-based test for both.
4. Replaced the `strings.Contains` filter with `isInferenceGatewayProxyName` (excludes controller-manager/webhook/metrics), applied to both the exposure Service filter and the EndpointSlice readiness filter so they stay in lockstep. Added predicate + table-case tests.
5. Left the forwarded-value test as-is per your note.

Rebased onto latest `main`, `make test` / lint green on the affected packages. Please take another look.

mchmarny (approved):

PR #1164, opened 2026-06-03
test(evidence): signed recipe-evidence pointers for h100 GKE + EKS inference

yuanchen8911:
## Summary

Adds two signed **recipe-evidence v1** pointers generated from live conformance validation on H100 inference clusters, to **exercise the warning-only recipe-evidence drift gate** (#1065, "Recipe Evidence: Verify").

| Recipe | Cluster | OCI digest | Rekor |
|--------|---------|-----------|-------|
| `h100-gke-cos-inference-dynamo` | GKE / COS (aicr-demo5) | `sha256:6a59465f…` | 1706764289 |
| `h100-eks-ubuntu-inference-dynamo` | EKS / Ubuntu (aicr3) | `sha256:da9d8838…` | 1706788485 |

Both bundles were built with `aicr validate --phase conformance --emit-attestation` against the live clusters (11/11 conformance checks passed on each), then signed + pushed to `ghcr.io/nvidia/aicr-evidence` via `aicr evidence publish` (Sigstore keyless OIDC).

## Motivation / Context

Validates the recipe-evidence verify pipeline end-to-end on real evidence: the gate should fetch each signed OCI artifact, verify its Sigstore signature, and compare the signed recipe digest against the current recipe.

Refs #1160

## Type of Change

- [x] CI / evidence (test of the recipe-evidence gate)

## Testing

- Conformance phase passed 11/11 on both clusters (GKE H100/COS, EKS H100/Ubuntu).
- Bundles signed (Sigstore keyless) and pushed to GHCR; pointers pin per-recipe digests.
- This PR's purpose is to trigger "Recipe Evidence: Verify" — observe its result.

PR #1165, opened 2026-06-03
fix(evidence): emit recipe-evidence pointer.yaml at 2-space indent

yuanchen8911:
## Summary

`aicr evidence publish` / `aicr validate --emit-attestation` wrote the recipe-evidence `pointer.yaml` with **4-space** sequence indentation (yaml.v3 default). Since the pointer is committed to `recipes/evidence/<recipe>.yaml` — where the repo's `.yamllint` (`spaces: 2`) lints it — the documented **publish → commit → PR** flow failed `make lint` on every generated pointer:

```
recipes/evidence/*.yaml  4:5 [indentation] wrong indentation: expected 2 but found 4
```

## Fix

`pkg/evidence/attestation/pointer.go` `MarshalPointer` now uses a `yaml.v3` encoder with `SetIndent(2)` instead of plain `yaml.Marshal`. Content/field order unchanged — only indentation. Fixes both code paths that emit a pointer (`validate --emit-attestation` and `evidence publish`, both via `WritePointer → MarshalPointer`).

(`emit.go` already serializes the recipe/snapshot via `serializer.MarshalYAMLDeterministic`; the pointer writer was the lone holdout on plain `yaml.Marshal`.)

## Type of Change

- [x] Bug fix (non-breaking)

## Testing

- New `TestMarshalPointer_TwoSpaceIndent`: asserts 2-space sequence indent, no 4-space, even leading-space on every line, and content round-trips.
- `go test ./pkg/evidence/attestation/...` pass; `golangci-lint` 0 issues; `gofmt` clean.
- Discovered while committing real signed evidence pointers (#1164), which failed `tests / Lint` until hand-reformatted; this removes the need for that manual step.

Refs #1164

### DIFF
--- a/pkg/evidence/attestation/pointer.go
+++ b/pkg/evidence/attestation/pointer.go
@@ -19,9 +19,8 @@ import (
 	"path/filepath"
 	"time"
 
-	"gopkg.in/yaml.v3"
-
 	"github.com/NVIDIA/aicr/pkg/errors"
+	"github.com/NVIDIA/aicr/pkg/serializer"
 )
 
 // PointerInputs carries the pointer-file fields that are not derived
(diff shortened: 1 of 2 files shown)


pkg/evidence/attestation/pointer.go
```diff
@@ -59,16 +60,28 @@ func BuildPointer(in PointerInputs) (*Pointer, error) {
...
+// The 2-space indent is load-bearing: the pointer is written to
+// recipes/evidence/<recipe>.yaml and committed, where the repo's
+// .yamllint (spaces: 2) lints it. yaml.v3's default 4-space sequence
+// indent makes the committed pointer fail `make lint`, breaking the
+// documented publish -> commit -> PR workflow, so set it explicitly.
 func MarshalPointer(p *Pointer) ([]byte, error) {
 	if p == nil {
 		return nil, errors.New(errors.ErrCodeInvalidRequest, "pointer is required")
 	}
-	body, err := yaml.Marshal(p)
```
njhensley, line 67: Your PR description calls this out. I think you could use it here instead. It's from "github.com/NVIDIA/aicr/pkg/serializer"
```
body, err := serializer.MarshalYAMLDeterministic(p)
```

njhensley (changes requested):

pkg/evidence/attestation/pointer.go
yuanchen8911: Good call — switched to `serializer.MarshalYAMLDeterministic(p)` in `df014037`, so the pointer now uses the same serializer as `emit.go` (recursively-sorted keys + 2-space indent) instead of a one-off encoder. Test updated for the sorted key order; build/tests/lint green.

njhensley (approved):
LGTM

PR #1166, opened 2026-06-03
ci: flag pointer-only changes in recipe-evidence gate

njhensley:
## Summary

Teach the recipe-evidence gate to flag a leaf recipe when the PR adds or modifies that recipe's own evidence pointer (`recipes/evidence/<slug>.yaml`), not just when an overlay/ancestor/values file changes.

## Motivation / Context

The discovery loop in `recipe-evidence-check.sh` flagged a recipe as "affected" only via rules 1–3 (overlay changed, ancestor overlay changed, or referenced component values file changed). A PR that *only* refreshes an evidence pointer — exactly the workflow the script's own "How to refresh evidence" section prescribes (`cp ./out/pointer.yaml recipes/evidence/<slug>.yaml`) — touched none of those, so the gate reported "No leaf overlays affected" and the freshly-submitted pointer was never `evidence verify`'d or digest-compared. This closes that gap.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `.github/scripts/recipe-evidence-check.sh` (warning-only recipe-evidence gate)

## Implementation Notes

- Adds **Rule 4** in the discovery loop: include the recipe when `recipes/evidence/<slug>.yaml` is in the diff. Mirrors Rule 1's `grep -qxF` whole-line exact match, so a sibling like `…/<slug>.yaml.bak` doesn't false-trigger.
- Scoped to the existing per-overlay loop on purpose: current-digest computation requires `-r <overlay>`, so a pointer without a matching leaf overlay couldn't be verified anyway.
- `promote_all` path is unaffected (every leaf is already included there), matching rules 1–3.
- Reachability holds end-to-end: the workflow's `paths: recipes/**` filter already covers `recipes/evidence/**`, so a pointer-only PR triggers the job; and the discovery loop is unconditional, so it evaluates Rule 4 even when no overlay/values file changed.
- Inherits the pre-existing `basename`-vs-criteria-slug naming limitation already tracked in the header (out of scope here).

## Testing

```bash
bash -n .github/scripts/recipe-evidence-check.sh   # syntax OK
```

Rule 4's matching logic was exercised in isolation against five diff scenarios (it is pure `grep`, independent of `yq`):

- pointer added as the only file → flagged
- pointer among unrelated files → flagged
- `…/<slug>.yaml.bak` sibling → not flagged
- a different slug's pointer → not flagged
- embedded substring (no whole-line match) → not flagged

The full script could not be run end-to-end locally because the sandbox's `yq` is the Python flavor while the script targets Go `yq eval` (as in CI); the new rule itself is yq-independent.

## Risk Assessment

- [x] **Low** — Isolated change to a warning-only gate that never blocks merge; additive discovery rule; easy to revert.

**Rollout notes:** N/A — gate is warning-only.

## Checklist

- [ ] Linter passes (`make lint`) — N/A, shell-only change; `bash -n` clean (shellcheck unavailable in sandbox)
- [ ] I added/updated tests for new functionality — N/A, no test harness for this script
- [x] I updated docs if user-facing behavior changed — header comment updated to document the four affected-recipe triggers

### DIFF
--- a/.github/scripts/recipe-evidence-check.sh
+++ b/.github/scripts/recipe-evidence-check.sh
@@ -173,6 +173,30 @@ done
 count=$(echo "$affected" | jq 'length')
 echo "Affected leaf overlays: ${count}"
 
+# Orphan-pointer check. The loop above only iterates existing overlays,
+# so a pointer added/modified for a slug with no matching leaf overlay
+# (a typo'd slug, or evidence for a retired recipe) never gets checked —
+# it would pass the gate unverified, defeating its purpose. Diff the
+# changed evidence pointers against the overlay set to surface those.
+# Only pointers present on disk at HEAD count: a *deleted* pointer is
+# absent here and is not an orphan (if its recipe still exists it is
+# already flagged via Rule 4 and reported as missing-pointer). This
+# shares the gate's basename==slug assumption (see header note 1).
+orphans="[]"
+while IFS= read -r pf; do
+  if [[ -z "$pf" ]]; then continue; fi
+  case "$pf" in
+    recipes/evidence/*.yaml) ;;
+    *) continue ;;
+  esac
+  pslug=$(basename "$pf" .yaml)
+  if [[ -f "$pf" && ! -f "recipes/overlays/${pslug}.yaml" ]]; then
+    orphans=$(echo "$orphans" | jq -c --arg s "$pslug" '. + [$s]')
+  fi
+done < <(printf '%s\n' "$changed_files")
+orphan_count=$(echo "$orphans" | jq 'length')
+echo "Orphan evidence pointers: ${orphan_count}"
+
 # --- Report build ------------------------------------------------------
 
 mkdir -p "$(dirname "$REPORT_OUT")"
@@ -191,7 +215,7 @@ mkdir -p "$(dirname "$REPORT_OUT")"
 
 warnings=0
 
-if [[ "$count" -eq 0 ]]; then
+if [[ "$count" -eq 0 && "$orphan_count" -eq 0 ]]; then
   {
     echo "No leaf overlays affected by this PR."
     echo
@@ -201,12 +225,14 @@ if [[ "$count" -eq 0 ]]; then
   exit 0
 fi
 
-{
-  echo "Affected leaf overlays: **${count}**"
-  echo
-  echo "| Recipe | Pointer | Verify | Digest match |"
-  echo "|---|---|---|---|"
-} >> "$REPORT_OUT"
+if [[ "$count" -gt 0 ]]; then
+  {
+    echo "Affected leaf overlays: **${count}**"
+    echo
+    echo "| Recipe | Pointer | Verify | Digest match |"
+    echo "|---|---|---|---|"
+  } >> "$REPORT_OUT"
+fi
 
 rows_written=0
 rows_truncated=0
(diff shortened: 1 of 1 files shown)


.github/scripts/recipe-evidence-check.sh
```diff
@@ -148,6 +153,18 @@ for overlay in recipes/overlays/*.yaml; do
...
   fi
 
+  # Rule 4: the recipe's own evidence pointer was added or modified.
+  # Refreshing evidence (`cp ./out/pointer.yaml recipes/evidence/<slug>.yaml`)
+  # touches no overlay or values file, so without this rule a pointer-only
+  # PR — the canonical "refresh evidence" workflow — would slip through
+  # unverified. `grep -qxF` matches the whole line, so a sibling like
+  # `recipes/evidence/${name}.yaml.bak` doesn't trigger it.
+  if [[ "$include" == "false" ]]; then
+    if printf '%s\n' "$changed_files" | grep -qxF "recipes/evidence/${name}.yaml"; then
```
yuanchen8911, line 163: Minor / non-blocking: this loop only iterates existing `recipes/overlays/*.yaml`, so a pointer added for a recipe with **no matching overlay** (typo'd slug, or a retired recipe) is silently skipped — it never hits an overlay iteration, so a misnamed evidence file wouldn't be flagged. Optional: diff added `recipes/evidence/*.yaml` against the overlay set to catch orphans.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -148,6 +153,18 @@ for overlay in recipes/overlays/*.yaml; do
     done <<<"$values_files"
   fi
 
+  # Rule 4: the recipe's own evidence pointer was added or modified.
```
yuanchen8911, line 156: Minor / non-blocking: `git diff BASE...HEAD` lists deletions too, so **removing** a pointer also marks the recipe affected → downstream reports it as missing-pointer (graceful, per existing handling). Probably intended; just flagging that a pointer delete and a pointer add both route through Rule 4.

yuanchen8911 (dismissed):
LGTM — correctly closes the evidence-only-PR gap (finding A): a pointer-only PR (like the recent test #1164) now gets verified instead of silently skipped. Reuses the existing `grep -qxF` whole-line match and integrates cleanly into the `affected` accumulation. Two minor, non-blocking edge-case notes inline.

.github/scripts/recipe-evidence-check.sh
njhensley: Good catch — addressed. Added an orphan-pointer check after the discovery loop that diffs the changed `recipes/evidence/*.yaml` against the overlay set, exactly as you suggested. It flags any added/modified pointer whose slug has no matching `recipes/overlays/<slug>.yaml` (typo'd slug or evidence for a retired recipe) in a dedicated "Orphan evidence pointers" section, counted as a warning.

Two details worth noting:
- Only pointers present on disk at HEAD are considered, so a *deleted* pointer is correctly not an orphan.
- The `count==0 & orphans>0` path is handled: the early "no overlays affected" exit now requires both to be zero, and the affected-recipes table is guarded behind `count>0` so it isn't rendered empty.

Verified the detection in isolation across has-overlay / typo'd-slug / retired-recipe / deleted / non-evidence-path cases.

.github/scripts/recipe-evidence-check.sh
njhensley: Yes — intended, thanks for flagging it explicitly. Deleting a pointer while its overlay still exists *should* flag the recipe: it now lacks evidence, and downstream renders the graceful "missing-pointer" warning.

The one edge — retiring a recipe by deleting both the pointer and its overlay in the same PR — produces no false warning: the overlay loop no longer iterates it, and the new orphan check skips it because the deleted pointer isn't on disk at HEAD.

yuanchen8911 (approved):

PR #1151, opened 2026-06-02
fix(docs): cover contributor docs in MDX check; catch autolinks

mchmarny:
## Summary

Closes #1148.

Expand \`tools/check-docs-mdx\` to scope \`docs/contributor/\` and add a new check for markdown autolinks (\`<https://...>\` / \`<http://...>\`) that break the MDX parser on the \`://\`. Fix the one offending autolink in \`docs/contributor/validator.md:461\` that slipped through when \#1143 added contributor docs to the Fern sidebar.

## Motivation/Context

PR \#1143 added \`docs/contributor/\` to the Fern sidebar, so contributor docs are now MDX-rendered. But the safety check in \`tools/check-docs-mdx\` only scoped \`docs/user/\` and \`docs/integrator/\`. \`validator.md:461\` had a \`<https://kyverno.github.io/...>\` autolink that broke MDX parsing with:

\`\`\`
Failed to parse ../docs/contributor/validator.md: Unexpected character \`/\` (U+002F) before local name
\`\`\`

Fixes: \#1148

## Type of Change

Bug fix (tooling + docs).

## Implementation Notes

**Three changes:**

1. \`SEARCH_DIRS\` in \`tools/check-docs-mdx\` now includes \`docs/contributor/\`.
2. New **Check 4** flags autolinks outside fenced code and inline code. The actual MDX-breaker class — once that fires, the \`://\` confuses the JSX tag parser.
3. \`docs/contributor/validator.md:461\` autolink rewritten as a markdown link.

**On the issue's listed 14 \`<word>\` patterns:** the other 13 (\`<phase>\`, \`<name>\`, \`<runID>\`, \`<component>\`) are all inside inline code spans (\`\` \` \`\`), which MDX treats as literal \`<code>\` content. They are not parser-breakers — the same placeholder convention is used in \`docs/user/cli-reference.md\` and \`docs/integrator/automation.md\`, both of which are MDX-rendered and pass. The new check follows the issue's own criterion — "outside code fences/inline code" — so it strips inline code spans before scanning, matching the existing braces/comments checks.

## Testing

- \`bash tools/check-docs-mdx\` — green
- \`make lint\` — green
- Synthetic test: dropped \`Visit <https://example.com> for details\` into a temp file under \`docs/contributor/\` → check flagged it as expected, then file removed.

## Risk Assessment

Low. Tightens a CI check and fixes one doc line; no runtime impact.

## Checklist

- [x] \`make lint\` passes
- [x] New check verified to catch the target pattern
[...]

PR #1170, opened 2026-06-03
fix(docs): catch bare <word> tags in MDX safety check

### LINKED ISSUE #1147: [Bug]: MDX safety check does not cover contributor docs
## Bug Description

`tools/check-docs-mdx` scopes its checks to `docs/user/` and `docs/integrator/` only. Since PR #1143 added contributor docs to the Fern sidebar, `docs/contributor/` is now rendered through Fern's MDX parser but is not covered by the safety check.

`docs/contributor/validator.md` currently has 14 bare `<word>` patterns (e.g., `<phase>`, `<name>`, `<runID>`, `<component>`) that break the MDX parser with:

```
Failed to parse ../docs/contributor/validator.md: Unexpected character `/` (U+002F)
before local name, expected a character that can start a name, such as a letter, `$`,
or `_` (note: to create a link in MDX, use `[text](url)`)
```

## Two fixes needed

1. **Escape the bare `<` patterns in `validator.md`** — wrap placeholder names in backticks or use `&lt;` (14 occurrences across lines 11, 13, 30, 91, 119, 137, 185, 413, 433, 434, 461, 513).

2. **Expand `tools/check-docs-mdx` scope** — add `docs/contributor/` to `SEARCH_DIRS` (line 25) and add a new check for bare `<word>` patterns outside code fences/inline code. The existing checks (void HTML elements, bare braces, HTML comments) don't catch this pattern class.

## Steps to Reproduce

```bash
cd fern && fern docs dev
# 500 error rendering docs/contributor/validator.md
```

Or grep for the patterns:
```bash
grep -n '<[a-zA-Z]' docs/contributor/validator.md | grep -v '```\|<!--'
```

pedjak:
## Summary

Add a bare `<word>` angle-bracket check (check 5) to `tools/check-docs-mdx` and fix two cross-line inline code spans that escaped detection.

## Motivation / Context

Fern's MDX parser interprets bare `<word>` placeholders as JSX component tags and fails with parse errors. The existing MDX safety script had no check for this pattern class.

Fixes: #1147
Related: #1151

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tools/check-docs-mdx`

## Implementation Notes

- Check 5 matches `<[a-zA-Z]` after stripping inline code and backslash-escaped patterns (`\<word\>`), excluding void HTML elements (check 1) and autolinks (check 4).
- Two cross-line backtick spans in `validator.md` and `collector.md` were collapsed to single-line spans so the per-line `gsub` stripper correctly excludes them.

## Testing

```bash
bash tools/check-docs-mdx  # OK: all doc files are MDX-safe
```

- Script passes clean on all docs
- Deliberate bare `<placeholder>` correctly flagged
- Backtick-wrapped, code-fenced, and `\<escaped\>` patterns not flagged

## Risk Assessment

**Rollout notes:** N/A

## Checklist

### DIFF
--- a/tools/check-docs-mdx
+++ b/tools/check-docs-mdx
@@ -19,6 +19,9 @@
 #   3. HTML comments (<!-- -->) outside fenced code blocks
 #   4. Autolinks (<https://...> / <http://...>) outside fenced and inline code
 #      — MDX tries to parse them as JSX tags and chokes on the `://`.
+#   5. Bare angle-bracket placeholders (<word>) outside fenced and inline code
+#      — MDX interprets them as JSX component tags and fails on unexpected
+#      content that follows.
 
 set -euo pipefail
 
@@ -49,7 +52,7 @@ for dir in "${SEARCH_DIRS[@]}"; do
   done < <(find "${dir}" -type f \( -name '*.md' -o -name '*.mdx' \) ! -name "${EXCLUDE}" 2>/dev/null)
 done
 
-# --- Checks 2, 3, 4: bare braces, HTML comments, and autolinks ---
+# --- Checks 2, 3, 4, 5: bare braces, HTML comments, autolinks, bare tags ---
 # All evaluated outside fenced code blocks; inline code spans are stripped
 # before the line is scanned so legitimate uses inside backticks don't trip.
 for dir in "${SEARCH_DIRS[@]}"; do
@@ -62,6 +65,9 @@ for dir in "${SEARCH_DIRS[@]}"; do
         # Strip inline code spans so backticked tokens don'\''t trigger.
         line = $0
         gsub(/`[^`]*`/, "", line)
+        # Strip backslash-escaped angle brackets (\<word\>) — these are
+        # valid markdown escapes and MDX-safe.
+        gsub(/\\<[^>]*\\>/, "", line)
       }
       # Bare { not preceded by backslash escape. The leading
       # (^|[^\\]) form catches { in column 1 as well.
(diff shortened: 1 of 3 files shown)


tools/check-docs-mdx
```diff
@@ -79,6 +85,13 @@ for dir in "${SEARCH_DIRS[@]}"; do
         printf "MDX: autolink outside code fence: %s:%d: %s\n", FILENAME, NR, $0
         errors++
       }
+      # Bare angle-bracket placeholders: <word> that are not void HTML
+      # elements (check 1) or autolinks (check 4). MDX interprets these
+      # as JSX component tags.
+      line ~ /<[a-zA-Z]/ && line !~ /<(img|br|hr|input|source|meta|link|area|base|col|embed|param|track|wbr)([ \t>\/])/ && line !~ /<https?:\/\// {
```
mchmarny, line 91: nit: line-level negation means a line containing **both** a void element (or autolink) **and** a bare `<word>` won't flag the placeholder. Reproduced:

```
$ echo '<br /> and <foo>' > t.md && bash tools/check-docs-mdx /path/with/t.md
OK
```

The void-element regex matches anywhere on the line, suppressing the whole line. Same shape as check 4 (autolinks), so this is consistent with existing code — flagging as an observation, not a request to change. If you wanted positional accuracy you'd strip the void/autolink matches from `line` (like inline code spans are stripped on lines 67/70) before the bare-tag predicate.

mchmarny, line 91: nit: the safe-list is HTML *void* elements only, but valid non-void containers like `<details>`, `<summary>`, `<kbd>`, `<sub>`, `<sup>` will get flagged if anyone uses them outside backticks. MDX accepts those fine. Quick repro:

```
$ echo '<details>' > t.md && bash tools/check-docs-mdx /path/with/t.md
MDX: bare <word> tag outside code fence: ...:1: <details>
```

No doc currently triggers this (the one `<details>` reference in `docs/contributor/maintaining.md:99` is inside backticks), so this is latent. Worth keeping in mind if someone later adds a collapsible `<details>` block — they'll get flagged and may not realize the fix is to escape with `&lt;...&gt;` or wrap in backticks.

mchmarny (approved):
Welcome to AICR, @pedjak — thanks for picking this up. The fix is well-scoped: check 5 follows the same shape as the existing check 4, the backslash-escape strip is the right complement to the inline-code strip, and the two cross-line backtick spans in `validator.md` / `collector.md` are exactly the kind of collateral fix that has to land in the same PR (the per-line `gsub` stripper can't see across lines).

Verified locally that the script still passes clean on all docs and correctly flags a bare `<word>` while leaving `\<word\>` and `` `<word>` `` alone.

Two low-priority observations inline — both about edge cases the predicate doesn't cover (lines with mixed content, and non-void HTML containers like `<details>`). Neither blocks merge; flagged for future iteration. PR is `behind` main, so a rebase + force-push will be needed before merge, and CI requires NVIDIA-runner vetting for a first-time contributor.

LGTM.

tools/check-docs-mdx
mchmarny: Actually, this is a blocker, the project docs generator will fail until this is resolved. Otherwise this LGTM

Issue #1172, opened 2026-06-03
Consolidate KWOK tier workflows: single reusable test workflow + one deployer-list source


njhensley:
## Summary

Consolidate the KWOK recipe-test workflow so all three tiers share **one** reusable workflow for the test steps and **one** source of truth for the deployer list. Today the deployer set and the `Run KWOK test` step block are duplicated across tiers, guarded only by a code comment.

This is a maintainability follow-up to the Tier 3 matrix-sharding fix (#1171) — that change deliberately scoped itself to the 256-config overflow and left this de-duplication for a focused PR.

## Background

The Tier 3 sharding fix extracted the Tier 3 test steps into a reusable workflow (`.github/workflows/kwok-tier3-shard.yaml`) and moved the deployer list into the `discover` job's bash. That solved the overflow but left two kinds of duplication in place.

### Duplication that remains

**Deployer list** — `["helm","argocd-oci","argocd-helm-oci","flux-oci"]` appears in:
- `.github/workflows/kwok-recipes.yaml` — `discover` bash var (drives Tier 3)
- `.github/workflows/kwok-recipes.yaml` — the `test-tier1` matrix axis
- `.github/actions/kwok-test/action.yml` — the input's doc string (prose only)

Tier 1 and Tier 3 must agree on the full four-deployer set; today that agreement is enforced only by a "keep this list in sync" comment. Tier 2 has **no** deployer axis (recipe-only matrix) — it exercises just the action's default deployer.

**`Run KWOK test` step block** — the checkout → load-versions → `kwok-test` trio with its ~13 version inputs is copied in three places: `test-tier1`, `test-tier2`, and the new `kwok-tier3-shard.yaml`.

## Proposal

- Generalize `kwok-tier3-shard.yaml` into a single `kwok-test-run.yaml` reusable workflow that takes a `pairs` input (JSON array of `{recipe, deployer}`) and owns the **only** copy of the checkout/load-versions/kwok-test steps.
- Have `discover` emit the deployer list (or pre-built `{recipe, deployer}` pairs) as an **output**, defined once in bash. Tier 1 and Tier 3 both consume it. Tier 1 builds pairs the same way Tier 3 does, over the smaller generic-overlay set (well under 256 → no batching needed).
- All three tiers become thin matrix callers of `kwok-test-run.yaml`. The step block exists once; the deployer set exists once; `action.yml`'s string stays as the only prose copy.

[...]

PR #1173, opened 2026-06-03
fix(ci): shard Tier 3 KWOK matrix to stay under 256-config cap

### LINKED ISSUE #1171: [Bug]: Tier 3 KWOK matrix exceeds GitHub's 256-configuration limit (288 configs)
## Summary

The `test-tier3` job in `.github/workflows/kwok-recipes.yaml` fails to start because its build matrix produces **288 configurations**, exceeding GitHub Actions' hard maximum of **256 configurations per job**.

```
.github/workflows/kwok-recipes.yaml (Line: 325, Col: 7):
Strategy for job 'test-tier3' produced 288 configurations which exceeds the maximum of 256 configurations
```

## Impact

**High** — Tier 3 is the full-matrix backstop that runs on push-to-`main` and the nightly schedule (per ADR-003). When the matrix is rejected, the job never launches, so the full recipe × deployer coverage silently stops running. The `summary` job treats a non-`failure`/`cancelled` Tier 3 result as acceptable, so this can pass unnoticed while coverage is actually lost.

## Root Cause

The Tier 3 matrix is a two-dimension cross-product (`.github/workflows/kwok-recipes.yaml:326-328`):

```yaml
matrix:
  recipe: ${{ fromJSON(needs.discover.outputs.tier3) }}   # all testable overlays
  deployer: [helm, argocd-oci, argocd-helm-oci, flux-oci]  # 4 deployers
```

`tier3` is the `all` set from the `discover` job — every overlay with a concrete `service` criterion. That set has grown to **72 recipes**:

- 72 recipes × 4 deployers = **288 configurations** > 256 cap.

The matrix has crossed the threshold organically as accelerator/service overlays were added; nothing in the workflow guards against it.

## Steps to Reproduce

1. Push a change touching `recipes/**` (or the workflow itself) to `main`, or wait for the nightly `0 3 * * *` schedule.
2. Observe the `test-tier3` job fail to start with the "288 configurations exceeds the maximum of 256" error.

## Expected Behavior

Tier 3 runs the intended recipe × deployer coverage without exceeding GitHub's matrix limit, and coverage scales as more overlays are added.

## Actual Behavior

The matrix is rejected before any job runs; Tier 3 coverage does not execute.

## Possible Fixes (for discussion)

1. **Shard the matrix** — split `test-tier3` into N jobs each handling a slice of recipes (e.g., a `recipe-shard` dimension), keeping each job's cross-product under 256. Most robust against future growth.
2. **Reduce deployer coverage per recipe** — test all recipes against one canonical deployer (`helm`), and run the full 4-deployer matrix only against a representative subset.
3. **Construct an explicit `include:` matrix in `discover`** — emit `{recipe, deployer}` pairs as JSON and cap the count, so the limit is enforced at generation time rather than discovered at runtime.
4. **Add a guard in `discover`** — fail fast with a clear message (and/or auto-shard) when `recipes × deployers` would exceed 256, so the failure is actionable instead of opaque.

Option 1 or 3 is preferred since the recipe count will keep growing.

## Additional Context

- Affected job: `test-tier3` — `.github/workflows/kwok-recipes.yaml:312-354`
- Matrix definition: lines `324-328`
- `tier3` source (`all` set): `discover` job, lines `96-111` and `215-216`
- The `summary` job (lines `394-398`) only flags Tier 3 on `failure`/`cancelled` — a matrix-generation rejection may not surface as a hard failure, so this should be treated as a coverage gap, not just a red X.

njhensley:
## Summary

Shards the Tier 3 KWOK `recipe × deployer` matrix into batches fanned out to a reusable workflow, so the job stays under GitHub Actions' hard cap of 256 matrix configurations per job.

## Motivation / Context

The `test-tier3` job crossed every testable recipe with all four deployers (72 × 4 = 288), exceeding GitHub's 256-config limit. GitHub rejected the matrix before any leg ran, so the post-merge/nightly full-matrix backstop silently stopped executing.

Fixes: #1171
Related: #1172 (follow-up: deployer-list / step-block de-duplication across tiers)

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: CI workflows (`.github/workflows/kwok-recipes.yaml`, `.github/workflows/kwok-tier3-shard.yaml`)

## Implementation Notes

- **`discover`** now builds the `{recipe, deployer}` cross-product and chunks it into batches of `TIER3_BATCH_SIZE` (200, with headroom under 256), emitting a new `tier3_batches` output of `{id, pairs}` objects. A fail-closed guard errors loudly if the batch size is ever raised past 256, rather than resurfacing GitHub's opaque rejection mid-fan-out.
- **`test-tier3`** is now a thin matrix over those batches that fans each one out to the new **`kwok-tier3-shard.yaml`** reusable workflow (matching the repo's existing `workflow_call` pattern), which expands its batch as its own ≤ 256 matrix. Batch count grows automatically as overlays are added — no manual job duplication.
- **Concurrency:** the group gains a `-${{ matrix.batch.id }}` suffix so every shard of a single run lands in its own group and runs in parallel; the SHA key still keeps successive merges independent (ADR-003's no-cancel-on-successive-merges guarantee preserved). The inner pair-matrix in the shard workflow has no concurrency block, so it is plainly parallel.
- **Edge cases:** `workflow_dispatch` emits `tier3_batches=[]` so the output contract is total; `test-tier3`'s `if:` guard is keyed on `tier3_batches` (the output it actually consumes).
- **ADR-003** amended: updated the concurrency snippet and added a "Tier 3 matrix sharding (GitHub 256-config cap)" subsection and refreshed structure diagram.

## Testing

This is a CI-workflow + docs change (no Go); the merge gate for workflows is actionlint + yamllint, run directly:

```bash
# actionlint pinned to the CI version (v1.7.11), exact CI flags — clean repo-wide
actionlint -color -shellcheck=

# yamllint with repo config — clean
yamllint -c .yamllint.yaml .github/workflows/kwok-recipes.yaml .github/workflows/kwok-tier3-shard.yaml
```

Batching logic validated locally with synthetic inputs (0, 1, 50, 72, 200, 201, 400 recipes × 4 deployers): no dropped or duplicated pairs, contiguous integer batch ids, every batch ≤ 200. Reviewed via a three-persona review (CI/security, distributed-systems correctness, maintainability) — no blockers; all feedback applied.

Note: Tier 3 only runs on push-to-`main` and the nightly schedule (never on PRs, by design), so the sharded matrix is first exercised end-to-end on the post-merge run.

## Risk Assessment

- [x] **Low** — Isolated CI/docs change, no product code; easy to revert (single commit).

**Rollout notes:** No migration. `KWOK Test Summary` remains the required check (per ADR-003), so the renamed/nested Tier 3 matrix job names do not affect branch protection.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — N/A (no Go changes); ran actionlint + yamllint instead
- [ ] I added/updated tests for new functionality — N/A (CI workflow change)
- [x] I updated docs if user-facing behavior changed (ADR-003)


.github/workflows/kwok-recipes.yaml
```diff
@@ -215,14 +219,47 @@ jobs:
...
+          deployers='["helm","argocd-oci","argocd-helm-oci","flux-oci"]'
+          readonly TIER3_BATCH_SIZE=200  # headroom under GitHub's 256 cap
+
+          # Fail closed if the batch size is ever raised past the hard limit —
+          # a too-large batch would otherwise resurface the opaque
+          # "produced N configurations which exceeds the maximum of 256" error.
+          if (( TIER3_BATCH_SIZE > 256 )); then
+            echo "::error::TIER3_BATCH_SIZE=${TIER3_BATCH_SIZE} exceeds GitHub's 256-configuration matrix limit"
+            exit 1
+          fi
```
mchmarny, line 240: Right shape for the guard — failing loud at config time beats letting GitHub reject the matrix mid-fan-out with its opaque "produced N configurations" message. Static check on the constant is sufficient because batches are slices of size ≤ `TIER3_BATCH_SIZE`, so the invariant holds by construction.

.github/workflows/kwok-recipes.yaml
```diff
@@ -307,51 +344,28 @@ jobs:
...
-          kubectl_version: ${{ steps.versions.outputs.kubectl }}
-          yq_version: ${{ steps.versions.outputs.yq }}
-          flux_version: ${{ steps.versions.outputs.flux }}
-          chainsaw_version: ${{ steps.versions.outputs.chainsaw }}
-          chainsaw_sha256: ${{ steps.versions.outputs.chainsaw_sha256_linux_amd64 }}
-          kind_node_image: ${{ steps.versions.outputs.kind_node_image }}
+        batch: ${{ fromJSON(needs.discover.outputs.tier3_batches) }}
+    uses: ./.github/workflows/kwok-tier3-shard.yaml
+    with:
+      pairs: ${{ toJSON(matrix.batch.pairs) }}
```
mchmarny, line 368: The `concurrency` + matrix + reusable-workflow combo is exactly the right idiom: matrix expansion creates one reusable-workflow invocation per batch, each with its own concurrency group (`<sha>-<batch.id>`), so shards within a single run go parallel while ADR-003's no-cancel-on-successive-merges contract holds via the SHA key.

One thing to eyeball on the first post-merge run: the GH UI now renders Tier 3 jobs as `test-tier3 / KWOK Tier 3 Shard / Tier 3: <recipe> (<deployer>)` instead of the flat `Tier 3: ...`. Branch protection still keys on `KWOK Test Summary` per ADR-003, so unaffected — just worth a sanity check that the summary correctly aggregates `needs.test-tier3.result` across all matrix legs.

mchmarny (dismissed):
Clean fix to a real problem. The reusable-workflow + batched-matrix pattern is the right idiom for the 256-config cap, and the implementation gets the details right: fail-closed batch-size guard, `tier3_batches=[]` on `workflow_dispatch` keeps the output contract total, `if:` condition keyed on the output actually consumed, concurrency group's `matrix.batch.id` suffix preserves ADR-003's no-cancel-on-successive-merges contract while letting shards within a single run parallelize, and path triggers correctly add the new shard workflow.

ADR-003 is updated in the same PR (diagram + new sharding subsection) and the deployer-list duplication is already tracked in #1172. CI green on the pre-merge commit.

Two inline notes — both positive observations, no asks. The first true end-to-end exercise of the sharded matrix is on the post-merge run since Tier 3 only fires on push-to-main + nightly; worth eyeballing the GH UI / summary aggregation once it lands.

LGTM.

mchmarny (approved):

PR #1180, opened 2026-06-03
fix(validator): honor tag override for main-container pull policy

### LINKED ISSUE #1177: validator: tag override doesn't force PullAlways on the main container (stale cached image runs)
## Summary

When `AICR_VALIDATOR_IMAGE_TAG` is set (e.g. `edge`), the **main validator
container** is created with `imagePullPolicy: IfNotPresent`, so a node that has
an older copy of that tag cached reuses the stale image. The tag override
silently does **not** take effect for the validator itself. Only the inner
AIPerf benchmark sidecar correctly gets `PullAlways`.

## Impact

Stale validator logic runs against a fresh recipe. Concrete symptom we hit:

- Recipe declared `inference-model: Qwen/Qwen3-8B` and
  `inference-concurrency-per-gpu: "256"`.
- The inference-perf benchmark actually ran `model=Qwen/Qwen3-0.6B`,
  `concurrency=64` — the old hardcoded smoke-test defaults from a cached `:edge`
  image that predated the recipe-driven model/concurrency wiring (#1133, #1096).
- The recipe's `inference-throughput: ">= 50000"` threshold **was** applied
  (that path is older), so an 8B-sized gate was checked against a 0.6B workload
  → guaranteed false failure.

Beyond confusing results, if `aicr validate --emit-attestation` is used, the
evidence attests a run that benchmarked the wrong model.

## Root cause

`ImagePullPolicy(image, imageTagOverride)` (`pkg/validator/v1/job_plan.go:116`)
returns `PullAlways` when `imageTagOverride != ""` (and for `:latest`), else
`PullIfNotPresent` for ordinary tags like `:edge`. See the table in
`pkg/validator/v1/README.md:348-356`.

The two call sites are inconsistent:

- **Main validator container(s):** `pkg/validator/v1/job_plan.go:284` and `:349`
  call `ImagePullPolicy(plan.Image, "")` — the override is passed as `""`, so the
  "override ⇒ PullAlways" rule never applies. With `:edge` this resolves to
  `PullIfNotPresent`.
- **Inner AIPerf container:** `validators/performance/inference_perf_constraint.go:1667`
  calls `ImagePullPolicy(aiperfImage, os.Getenv("AICR_VALIDATOR_IMAGE_TAG"))` —
  correctly forces `PullAlways` when the override is set.

So `AICR_VALIDATOR_IMAGE_TAG=edge` forces a fresh pull of the benchmark sidecar
but not of the validator that actually picks the model and deploys the workload.

## Proposed fix

Honor the tag override consistently across **all** validator containers: pass
the resolved `AICR_VALIDATOR_IMAGE_TAG` (or a shared "override in effect" signal)
into the pull-policy computation at `job_plan.go:284` and `:349`, matching what
the AIPerf sidecar already does — so a mutable tag override (and `:latest`)
forces `PullAlways` everywhere while digest-pinned refs stay `IfNotPresent`.

Net effect: when an operator overrides the validator tag, the node always pulls
the current image for the validator, not just the benchmark sidecar.

## Reproduce

1. On a GPU node that already cached an older `…/aicr-validators:edge`.
2. Run `AICR_VALIDATOR_IMAGE_TAG=edge aicr validate --phase performance …` on an
   inference recipe whose `inference-model` differs from the cached image's
   default.
3. Observe the benchmark uses the cached image's old model/concurrency, not the
   recipe's.

## Workaround

Use a unique immutable tag the node has not cached (forces a pull), e.g.
`AICR_VALIDATOR_IMAGE_TAG=sha-<commit>`; or `crictl rmi` the cached image / cycle
the node.

yuanchen8911:
## Summary

Force `PullAlways` on the **main validator container** when an image tag override (`AICR_VALIDATOR_IMAGE_TAG`) is in effect, so a node does not silently reuse a cached older image for a mutable tag like `:edge`. Previously only the inner AIPerf sidecar honored the override.

Also folds in two small doc corrections to `docs/contributor/validator.md` (the guide touched by #1176), since they describe this exact pull-policy helper and the tag-resolution flow.

## Motivation / Context

`ImagePullPolicy(image, imageTagOverride)` returns `PullAlways` when `imageTagOverride != ""` (or `:latest`), else `PullIfNotPresent`. The two render paths for the main validator container called it with `""`:

- `RenderPlan` (`pkg/validator/v1/job_plan.go`)
- `RenderPlanToApplyConfig` (same file)

So with `AICR_VALIDATOR_IMAGE_TAG=edge`, the tag was overridden but the pull policy stayed `IfNotPresent` for the main container — a node that had cached an older `:edge` ran **stale validator logic**. The inner AIPerf sidecar (`validators/performance/inference_perf_constraint.go`) already passed the override and pulled fresh, which masked the bug.

Concrete symptom (from #1177): a recipe pinning `inference-model: Qwen/Qwen3-8B` + `inference-concurrency-per-gpu: "256"` benchmarked `Qwen3-0.6B` at low concurrency from a stale cached `:edge` validator, while the (older) throughput-threshold path still applied the 8B-sized gate → guaranteed false failure. With `--emit-attestation`, the bundle would attest the wrong run.

Fixes: #1177
Related: #1176 (docs), #1133 (recipe-driven model/concurrency)

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- Added `JobPlan.ImageTagOverride`, set from `imageTagOverride` in `BuildJobPlan` (the override is already applied to `Image`; this preserves the "override in effect" signal for pull-policy computation).
- Both render sites now call `ImagePullPolicy(plan.Image, plan.ImageTagOverride)` instead of `ImagePullPolicy(plan.Image, "")`, mirroring the AIPerf sidecar.
- Net behavior: a mutable override tag (and `:latest`) forces `PullAlways` on **all** validator containers; digest-pinned refs stay `IfNotPresent` (air-gap-safe), and an immutable `:sha-<commit>` with no override stays `IfNotPresent` (safe — the tag is immutable).
- Docs (`docs/contributor/validator.md`): corrected the **Image-pull policy** prose to the real two-arg signature `ImagePullPolicy(image, imageTagOverride)`, and reworded step 1 of the tag-resolution list to note that `ResolveImage` checks the version first — a **release** build resolves to `:vX.Y.Z`, while `:sha-<commit>` is the dev/`main` case. (Follow-up nits from the #1176 cross-review, applied here since they describe this code.)

## Testing

```bash
gofmt -l pkg/validator/v1/...                       # clean
go test -race ./pkg/validator/v1/...                # ok
golangci-lint run -c .golangci.yaml ./pkg/validator/v1/...   # 0 issues
go test -cover ./pkg/validator/v1/...               # 75.0%
```

Added `TestRenderPlanTagOverridePullPolicy`: asserts `PullAlways` on the main container for `:edge` + override, and `IfNotPresent` for `:edge` without override and for an immutable `:sha` pin — across **both** render paths (`RenderPlan` and `RenderPlanToApplyConfig`).

The `docs/contributor/validator.md` change is Markdown-only (no Go behavior change); it is covered by the `docs/**` MDX/lychee CI gates — all added angle-brackets stay inside backticks and no headings/anchors changed.

## Risk Assessment

- [x] **Low** — Two call-site changes + one plan field, plus a docs-only correction; covered by a regression test. Only affects pull policy when an override is set; digest-pin and no-override behavior unchanged.

**Rollout notes:** Backwards compatible. Operators using `AICR_VALIDATOR_IMAGE_TAG=edge` will now correctly re-pull the main validator (one extra pull when the cached tag is stale).

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — `./pkg/validator/v1/...`
- [x] Linter passes (`make lint`) — `golangci-lint` 0 issues on the package

### DIFF
--- a/pkg/validator/v1/job_plan_test.go
+++ b/pkg/validator/v1/job_plan_test.go
@@ -472,6 +472,56 @@ func TestRenderPlan(t *testing.T) {
 	}
 }
 
+// TestRenderPlanTagOverridePullPolicy is the regression guard for #1177: a
+// mutable image tag override (e.g. :edge) must force PullAlways on the MAIN
+// validator container, not just the inner AIPerf sidecar — otherwise a node
+// that cached an older :edge silently reuses stale validator logic. Both
+// render paths (typed RenderPlan and the server-side-apply
+// RenderPlanToApplyConfig) must honor plan.ImageTagOverride.
+func TestRenderPlanTagOverridePullPolicy(t *testing.T) {
+	tests := []struct {
+		name     string
+		image    string
+		override string
+		want     corev1.PullPolicy
+	}{
+		{name: "mutable :edge with override → Always", image: "ghcr.io/nvidia/aicr-validators/performance:edge", override: "edge", want: corev1.PullAlways},
+		{name: "mutable :edge without override → IfNotPresent", image: "ghcr.io/nvidia/aicr-validators/performance:edge", override: "", want: corev1.PullIfNotPresent},
+		{name: "immutable :sha pin, no override → IfNotPresent", image: "ghcr.io/nvidia/aicr-validators/performance:sha-abc1234", override: "", want: corev1.PullIfNotPresent},
+		// Digest pin + override → IfNotPresent: the @digest short-circuit wins
+		// over the override, so air-gap/disconnected behavior is preserved on
+		// the main container too (the explicit invariant this fix advertises).
+		{name: "digest pin with override → IfNotPresent (air-gap-safe; digest wins)", image: "ghcr.io/nvidia/aicr-validators/performance@sha256:deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef", override: "edge", want: corev1.PullIfNotPresent},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			plan := JobPlan{
+				JobName:          "j",
+				Namespace:        "ns",
+				Image:            tt.image,
+				ImageTagOverride: tt.override,
+				Timeout:          300,
+			}
+
+			// Typed render path (job_plan.go RenderPlan).
+			job := RenderPlan(plan)
+			if got := job.Spec.Template.Spec.Containers[0].ImagePullPolicy; got != tt.want {
+				t.Errorf("RenderPlan main container ImagePullPolicy = %q, want %q", got, tt.want)
+			}
+
+			// Server-side-apply render path (RenderPlanToApplyConfig).
+			jobApply := RenderPlanToApplyConfig(plan, "j")
+			gotApply := jobApply.Spec.Template.Spec.Containers[0].ImagePullPolicy
+			if gotApply == nil {
+				t.Fatal("RenderPlanToApplyConfig main container ImagePullPolicy is nil")
+			}
+			if *gotApply != tt.want {
+				t.Errorf("RenderPlanToApplyConfig main container ImagePullPolicy = %q, want %q", *gotApply, tt.want)
+			}
+		})
+	}
+}
+
 func TestRenderPlanToApplyConfig(t *testing.T) {
 	plan := JobPlan{
 		ValidatorName:    "test-validator",
--- a/docs/contributor/validator.md
+++ b/docs/contributor/validator.md
@@ -187,8 +187,8 @@ prevents concurrent runs from clobbering each other's RBAC. External
 tooling selects by label `app.kubernetes.io/name=aicr-validator`, not
 literal name.
 
-**Image-pull policy** is computed by `v1.ImagePullPolicy(image)` in
-`pkg/validator/v1/job_plan.go`:
+**Image-pull policy** is computed by `v1.ImagePullPolicy(image,
+imageTagOverride)` in `pkg/validator/v1/job_plan.go`:
 side-loaded (`ko.local/*`, `kind.local/*`) → `Never`;
 digest-pinned (`name@sha256:…`) → `IfNotPresent`;
 `AICR_VALIDATOR_IMAGE_TAG` set or `:latest` suffix → `Always`;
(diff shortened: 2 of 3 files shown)


njhensley (dismissed):
LGTM

pkg/validator/v1/job_plan_test.go
```diff
@@ -472,6 +472,52 @@ func TestRenderPlan(t *testing.T) {
...
+// validator container, not just the inner AIPerf sidecar — otherwise a node
+// that cached an older :edge silently reuses stale validator logic. Both
+// render paths (typed RenderPlan and the server-side-apply
+// RenderPlanToApplyConfig) must honor plan.ImageTagOverride.
+func TestRenderPlanTagOverridePullPolicy(t *testing.T) {
+	tests := []struct {
+		name     string
+		image    string
+		override string
+		want     corev1.PullPolicy
```
mchmarny, line 486: nit: the PR description calls out "digest-pinned refs stay `IfNotPresent` (air-gap-safe)" as the explicit invariant of this fix. Worth adding a 4th case here — `@sha256:...` *with* override → `IfNotPresent` — so the render-path test pins that promise too. `TestImagePullPolicy` already exercises the helper directly, so this is optional, but it would lock the digest-short-circuit behavior into the wiring test as well.

mchmarny (dismissed):
Clean targeted fix. Both render paths (`RenderPlan` and `RenderPlanToApplyConfig`) now mirror the AIPerf sidecar pattern in `validators/performance/inference_perf_constraint.go:1667`, and `ImagePullPolicy` short-circuits on `@digest` before checking the override — so digest pins stay `IfNotPresent` and air-gap behavior is preserved. Storing the override on `JobPlan` is a reasonable way to carry the signal across the build → render boundary without re-plumbing params.

Doc corrections in `docs/contributor/validator.md` are accurate against the current source (two-arg `ImagePullPolicy` signature, release vs. dev tag resolution).

All required checks pass. One optional nit inline on the test table — non-blocking.

pkg/validator/v1/job_plan_test.go
yuanchen8911: Good call — added a 4th case to `TestRenderPlanTagOverridePullPolicy`: a digest pin (`@sha256:…`) with the override set still resolves to `IfNotPresent`, so the render path now pins the air-gap invariant directly (the `@digest` short-circuit wins over the override). In b5ca28af.

njhensley (approved):

PR #1198, opened 2026-06-04
fix(validator): run all phases by default; add --fail-fast opt-in

### LINKED ISSUE #1178: validator: a failing phase skips all later phases by order, not dependency (perf failure skips conformance)
## Summary

`ValidatePhases` runs phases in a fixed `PhaseOrder` and latches a single
`overallFailed` flag on the first phase that fails; every **subsequent** phase
is then reported `skipped` — regardless of whether it actually depends on the
failed phase.

```go
// pkg/validator/validator.go:200-225
overallFailed := false
for _, phase := range phases {          // PhaseOrder: deployment → performance → conformance
    if overallFailed {
        pr := v.phaseSkipped(cat, phase, "skipped due to previous phase failure")
        results = append(results, pr); continue
    }
    pr, _ := v.runPhase(...)
    if pr.Status == ctrf.StatusFailed { overallFailed = true }   // sticky
}
```

`PhaseOrder` and the documented behavior: `pkg/validator/phases.go:40,44`
(*"If a phase fails, subsequent phases are skipped."*).

## Impact

The skip is by **order, not dependency**. So a `performance` failure (e.g.
`inference-perf` missing a throughput gate) skips the **entire `conformance`
suite**, even though conformance does not depend on performance. You lose all
conformance signal on any perf failure, and with `aicr validate
--emit-attestation` the evidence bundle records conformance as `skipped`.

`--fail-on-error=false` does **not** help — it only changes the final exit code,
not the phase orchestration.

## Proposed fix

Either:

- add an opt-in to run all phases regardless (`--continue-on-phase-failure` /
  config field), collecting every phase's results; and/or
- narrow the fail-fast so only a **deployment** failure gates later phases
  (deployment is the real prerequisite — nothing is deployed to test). A
  `performance` failure should not skip `conformance`, since they're independent.

## Reproduce

Run `aicr validate` (all phases) on an inference recipe where the performance
phase fails its threshold; observe `conformance` reported `skipped due to
previous phase failure` with zero conformance tests executed.

njhensley:
## Summary

`aicr validate` now runs all requested phases by default instead of skipping every phase after the first failure. A new `--fail-fast` flag (and `spec.validate.execution.failFast` config field) restores the old stop-on-first-failure behavior as an opt-in.

## Motivation / Context

`ValidatePhases` latched a single `overallFailed` flag on the first failing phase and reported every *subsequent* phase as `skipped` — by execution order, not dependency. Because the order is `deployment → performance → conformance`, a `performance` threshold miss (e.g. an `inference-perf` throughput gate) skipped the **entire conformance suite**, even though conformance does not depend on performance. With `aicr validate --emit-attestation` this silently recorded conformance as `skipped` in the evidence bundle, losing all conformance signal. `--fail-on-error=false` did not help — it only changes the exit code, not phase orchestration.

This flips the default to "run all phases, collect every result" and makes fail-fast opt-in, which is the behavior the issue requested.

Fixes: #1178
Related: N/A

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe`) — config schema (`pkg/config`)
- [x] Other: `pkg/client/v1` facade

## Implementation Notes

- The phase loop in `ValidatePhases` was extracted into a `runPhases(ctx, runner, cat, phases)` helper that takes an injected per-phase runner, so the orchestration logic (fail-fast gating, skip recording) is unit-testable without a live cluster.
- The old unconditional latch `if overallFailed { skip }` became `if anyFailed && v.FailFast { skip }`. The skip-reason string stays mechanism-neutral (no CLI flag name) since it surfaces in the CTRF report consumed by non-CLI callers.
- `FailFast` is plumbed CLI flag → `pkg/config` (`*bool`, nil = inherit CLI default) → `pkg/client/v1` (`WithValidationFailFast`) → `validator.WithFailFast`. This mirrors the existing `FailOnError` plumbing. `FailOnError` (exit-code policy) stays CLI-only; `FailFast` (orchestration policy) lives on the `Validator` because `ValidatePhases` is the single drive point for both CLI and server callers.
- Rebased onto `main` to pick up the inference-gateway exposure feature (#1163). One conflict in `pkg/validator/v1/job_plan_internal.go`: that PR added `requireScopedInferenceGatewayEnv` to the orchestrator env-skip guard while this branch's lint commit replaced the `"HF_TOKEN"` literal with the `hfTokenEnvVar` constant. Resolved by keeping both — `if e.Name == hfTokenEnvVar || e.Name == requireScopedInferenceGatewayEnv`.

## Testing

```bash
# Affected-package unit suites:
GOFLAGS="-mod=vendor" go test ./pkg/validator/... ./pkg/client/v1/... ./pkg/config/... ./pkg/cli/...
```

All affected-package tests pass. New coverage: `runPhases` default-runs-all, fail-fast-stops, and passed-phase-does-not-gate cases; `WithValidationFailFast` round-trip through the facade; `--fail-fast` flag default; config resolution of `failFast`.

**Live verification — EKS GB200 inference-dynamo.** Validated the orchestration change end-to-end with a release-built binary (`make build`, `aicr_darwin_arm64_v8.0`) against a live EKS GB200 inference-dynamo cluster (`gb200-eks-ubuntu-inference-dynamo` recipe, 18 components, fresh cluster snapshot).

| Scenario | deployment | performance | conformance |
|---|---|---|---|
| default, all pass | passed (26s) | passed (10m29s) | **passed (1m39s) — all phases ran** |
| default, deployment forced to fail | failed (30s) | — (scoped out) | **passed — still ran** |
| `--fail-fast`, deployment forced to fail | failed (25s) | — | **skipped (0s)** — "skipped due to previous phase failure" |

Key confirmations:
- **Default runs all phases regardless of failure** — with deployment forced to fail (gpu-operator constraint `>= v99.0.0` on a throwaway recipe copy), conformance still executed and passed. Confirms the removed `overallFailed` latch.
- **`--fail-fast` short-circuits on first failure** — conformance reported `skipped` (0s); orchestrator logged `skipping phase due to fail-fast: phase=conformance`.
- **Flag wiring intact** CLI → config → `client/v1` → `runPhases`: binary exposes `--fail-fast`; CLI override logged (`flag=fail-fast config=false override=true`).
- **Full 3-phase happy path passed**: throughput 76,309 tok/s (constraint `>= 50000`), TTFT p99 1032 ms (constraint `<= 2000`), 4× GB200, driver 580.126.20.

Failure injection used only a throwaway recipe copy — production code and the real recipe were untouched. The validator cleaned up its `aicr-validation` namespace after each run.

## Risk Assessment

- [x] **Low** — Isolated to phase orchestration; default flip is well-tested (unit + live EKS GB200) and the old behavior is one flag away.

**Rollout notes:** Behavioral default change — clusters where a `performance` failure was previously skipping `conformance` will now run conformance and emit its results. Operators who relied on stop-on-first-failure must add `--fail-fast` or set `spec.validate.execution.failFast: true`. No API/wire breakage; the `failFast` config field is additive and optional.

## Checklist

- [x] Tests pass locally — full `make qualify` green (unit tests with `-race`, coverage 75.9% ≥ 75% threshold, 22/22 chainsaw e2e); also verified live on EKS GB200.
- [x] Linter passes (`make lint`) — golangci-lint reports `0 issues`; `go vet` clean; license headers + AGENTS sync OK.
- [x] Commits are cryptographically signed (`git commit -S`) — all 8 commits show a good signature (`git log --show-signature`). [GPG signing info](https://docs.github.com/en/authentication/managing-commit-signature-verification)

### DIFF
--- a/docs/contributor/validator.md
+++ b/docs/contributor/validator.md
@@ -104,6 +104,11 @@ per run. Per-phase containers are built from
 `ParsePhaseSelection` collapses it to nil-meaning-everything. It is
 **exclusive** — combining `all` with any other phase is rejected.
 
+By default all phases run and produce results regardless of earlier failures —
+a performance threshold miss no longer silences conformance results. Pass
+`--fail-fast` (or set `spec.validate.execution.failFast: true` in config) to
+restore stop-on-first-failure behavior for cost-sensitive runs.
+
 `readiness` is also a field on `ValidationConfig` (see
 `pkg/recipe/validation.go`) and appears in overlay examples, but it
 is **not** a container-per-validator phase. Readiness runs as
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -630,6 +630,7 @@ aicr validate [flags]
 | `--config` | | string | | Path or HTTP/HTTPS URL to an AICRConfig file (YAML/JSON). CLI flags override values from this file. See [Validate Config File Mode](#validate-config-file-mode). |
 | `--phase` | | string[] | all | Validation phase to run: deployment, performance, conformance, all (repeatable) |
 | `--fail-on-error` | | bool | true | Exit with non-zero status if any constraint fails |
+| `--fail-fast` | | bool | false | Stop after the first phase that fails. By default all phases run and produce results. |
 | `--output` | `-o` | string | stdout | Output destination: file path, ConfigMap URI (`cm://namespace/name`), or stdout |
 | `--kubeconfig` | `-k` | string | ~/.kube/config | Path to kubeconfig file (used when `--recipe`, `--snapshot`, or `--output` is a ConfigMap URI) |
 | `--namespace` | `-n` | string | aicr-validation | Kubernetes namespace for validation Job deployment |
@@ -669,7 +670,7 @@ Validation can be run in different phases to validate different aspects of the d
 | `deployment` | Validates component deployment completeness plus post-install GPU readiness signals (see below) | After deploying components |
 | `performance` | Validates system performance and network fabric health | After components are running |
 | `conformance` | Validates workload-specific requirements and conformance | Before running production workloads |
-| `all` | Runs all phases sequentially with dependency logic | Complete end-to-end validation |
+| `all` | Runs all phases sequentially; results collected regardless of failures | Complete end-to-end validation |
 
 > **Note:** Readiness constraints (K8s version, OS, kernel) are always evaluated implicitly before any phase runs. If readiness fails, validation stops before deploying any Jobs.
 
@@ -689,6 +690,7 @@ The `deployment` phase verifies that the cluster is actually ready for GPU workl
 
 **Phase Dependencies:**
 - Phases run sequentially when using `--phase all`
+- All phases run by default and produce results regardless of earlier failures. Use `--fail-fast` to stop after the first phase that fails (e.g., to skip a 65-minute inference-perf run when deployment already failed).
 - If a phase fails, subsequent phases are skipped
 - Use individual phases for targeted validation during specific deployment stages
 
--- a/docs/user/cli-reference.md
+++ b/docs/user/cli-reference.md
@@ -691,7 +691,6 @@ The `deployment` phase verifies that the cluster is actually ready for GPU workl
 **Phase Dependencies:**
 - Phases run sequentially when using `--phase all`
 - All phases run by default and produce results regardless of earlier failures. Use `--fail-fast` to stop after the first phase that fails (e.g., to skip a 65-minute inference-perf run when deployment already failed).
-- If a phase fails, subsequent phases are skipped
 - Use individual phases for targeted validation during specific deployment stages
 
 #### Constraint Format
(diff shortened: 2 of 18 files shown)


mchmarny (approved):
LGTM. The `overallFailed` → `anyFailed && v.FailFast` flip is exactly the minimal change #1178 asks for, and extracting `runPhases` with an injected per-phase runner makes the orchestration logic unit-testable without a cluster — the three new tests (default-runs-all, fail-fast-stops, passed-phase-doesn't-gate) cover the matrix cleanly.

Plumbing is consistent: `*bool failFast` mirrors `FailOnError`'s tri-state (nil = inherit CLI default) through `pkg/config` → `pkg/client/v1` → `pkg/validator`. The round-trip test in `aicr_internal_test.go` is well-constructed — pre-seeding `WithFailFast(true)` to catch a regression where an unset `failFast` would still emit `WithFailFast(false)` is the kind of test that only pays off the day someone breaks it. Docs are updated in all three relevant pages, including the skip-reason table.

Behavioral default change is well-flagged in the PR body's Rollout notes; the live EKS GB200 matrix (default-all-pass / default-deployment-fail-conformance-still-runs / fail-fast-skips) is solid verification end-to-end. CI green. No correctness, security, or architecture concerns. The two unrelated constant extractions (`hfTokenEnvVar`, `defaultStorageClassAnnotationValue`) are tiny enough not to block — the former is explained as a rebase artifact, the latter is a small consistency win.

PR #1199, opened 2026-06-04
fix(snapshotter): auto-target GPU nodes and warn on placement mismatch

### LINKED ISSUE #1179: snapshot: silently collects no GPU data when the Job lands on a non-GPU node
## Summary

`aicr snapshot` deploys a Job whose node placement comes entirely from
`--node-selector` / `--toleration` (defaults: empty selector, and **tolerate all
taints**), and it requests no GPU resource by default. On a heterogeneous
(CPU + GPU) cluster the scheduler can therefore place the agent on a **non-GPU
node**, where `nvidia-smi` is absent → no GPU data is collected and the
accelerator fingerprint is empty. Placement is non-deterministic, so this
happens **intermittently**, and the snapshot still reports success — there is no
warning or failure.

## Impact

- Downstream recipe resolution / fingerprint / criteria-match get an empty (or
  wrong) `accelerator` dimension.
- With `aicr validate --emit-attestation`, the evidence then attests a snapshot
  that collected no GPU data — silently misleading, and directly undercutting
  the recipe-evidence trust story.
- Hard to catch precisely because it's intermittent and silent.

Observed agent output when it lands on a CPU node:

```
[agent] node topology collection complete: nodes=6 taints=2 labels=189
nvidia-smi not found - no GPU data will be collected
```

Note the cluster-wide node topology (from the K8s API) still shows the GPU
nodes/labels — so the discrepancy (GPU nodes exist, but local GPU collection
found nothing) is detectable.

## Root cause

- The agent Job's `NodeSelector` / `Tolerations` come from the CLI flags
  (`pkg/k8s/agent/job.go:113-114`); tolerations default to tolerate-all, and the
  default selector is empty.
- `aicr snapshot --help` advises `--node-selector` on heterogeneous clusters,
  but nothing enforces it or warns when it's missing and the agent ends up off a
  GPU node.

## Proposed fix

When the collected snapshot has **zero GPU data but the cluster topology shows
GPU nodes** (e.g. `nvidia.com/gpu.present=true`, GPU labels, or
`nvidia.com/gpu` allocatable), emit a loud warning — or fail (configurable) —
rather than silently producing a GPU-less snapshot. Optionally: auto-target a
GPU node, or make `--require-gpu` the default when the requested/criteria
accelerator implies a GPU.

## Workaround

Pin the snapshot Job to a GPU node:

```shell
aicr snapshot -o snapshot.yaml \
  --node-selector nvidia.com/gpu.present=true \
  --runtime-class nvidia        # nvidia-smi access without consuming a GPU
# or: --require-gpu  (fails fast if no GPU; requests an nvidia.com/gpu resource)
```

njhensley:
## Summary

`aicr snapshot` now auto-targets GPU nodes when run on a heterogeneous cluster, and warns when a captured snapshot has no GPU data despite the cluster having GPU-capable nodes — closing the silent "snapshot landed on a CPU node" failure mode.

## Motivation / Context

`aicr snapshot` deploys a Job whose placement defaults to an empty node selector with tolerate-all tolerations, and it requests no GPU resource. On a mixed CPU+GPU cluster the scheduler can place the agent on a non-GPU node where `nvidia-smi` is absent, so no GPU data is collected — yet the snapshot still reports success. Placement is non-deterministic, so this happens intermittently and silently, corrupting downstream recipe resolution, fingerprinting, and `--emit-attestation` evidence.

Two complementary defenses:
- **Proactive:** before deploy, if no placement constraint is set (`--node-selector` / `--require-gpu` / `--runtime-class`) and the cluster has `nvidia.com/gpu.present=true` nodes, auto-inject that selector so the Job lands on a GPU node.
- **Reactive safety net:** after collection, if the snapshot has no GPU data but topology shows GPU-capable nodes, emit a loud warning with remediation options.

Fixes: #1179
Related: N/A

## Type of Change

## Component(s) Affected

- [x] Core libraries (`pkg/errors`, `pkg/k8s`) — adds `defaults.GPUNodeDetectionTimeout`

## Implementation Notes

- `maybeInjectGPUNodeSelector` (new `pkg/snapshotter/detect.go`) is a no-op when any explicit placement constraint is already set, so user intent is always preserved. The pre-flight node `List` is `limit=1` (presence check only), bounded by `GPUNodeDetectionTimeout` (5s), and **fail-open** — an API error logs a warning and proceeds without injection rather than blocking the snapshot.
- **TOCTOU:** a GPU node can be cordoned between detection and pod scheduling. When that happens the Job pends to timeout (a better failure mode than silently landing on a CPU node); the wait-timeout error names the auto-injected selector so operators understand the cause.
- `detectGPUPlacementMismatch` keys on the `nvidia.com/gpu.` label prefix (covers `.present` and `.product`). Detection relies on NFD/GFD labels — clusters without them are not auto-detected, which the warning message states explicitly.
- `hasGPUData` reuses `measurement.Subtype.GetInt64`, which handles `int`/`int64`/`float64` — required because snapshots round-tripped through YAML/JSON deliver integers as `float64` (the prior `int`-only assertion would have mis-fired on parsed snapshots).

## Testing

**Manual verification** — live EKS GB200 inference (Dynamo) cluster: K8s 1.34.7-eks, Ubuntu 24.04, us-east-1, 6 nodes / 2 GPU nodes (gb200).

- **Single run:** snapshot completed successfully. Auto-targeting fired as designed (`auto-targeting GPU nodes: selector=nvidia.com/gpu.present=true`); fingerprint correctly resolved `accelerator=gb200`, `gpuNodeCount=2`, `service=eks`. No placement-mismatch warning (agent landed on a GPU node as expected).
- **20 sequential runs:** 20/20 succeeded. Auto-targeting fired every run; zero placement-mismatch warnings; no failures — confirms stable, deterministic behavior across repeated runs.
- **20 parallel runs (stress):** failed as expected on the shared `aicr` Job name (`jobs.batch "aicr" already exists`) and GPU-node saturation timeouts. These are pre-existing concurrency collisions on the single Job name, not regressions from this change — the auto-selector logic behaved identically across all attempts.

**Unit tests:**

```bash
make qualify
```

- `pkg/snapshotter` covers `warnOnGPUPlacementMismatch`, `hasGPUData` (incl. the `float64` JSON round-trip path), auto-injection no-op paths, and GPU node detection across NFD label variants.
- `golangci-lint` clean on changed packages.
- Coverage: `pkg/snapshotter` 45.9% → 48.7% (+2.8%).

## Risk Assessment

- [x] **Low** — Additive change behind a guard that no-ops whenever a placement constraint is already set; auto-detection and the warning both fail open and never block a snapshot.

**Rollout notes:** Backwards compatible. Existing invocations that pass `--node-selector`, `--require-gpu`, or `--runtime-class` are unaffected. The only behavioral change for unconstrained invocations is the auto-injected selector on clusters that advertise `nvidia.com/gpu.present=true`.

## Checklist


mchmarny (approved):
LGTM. Two-defense design (proactive auto-injection + reactive warning) is well-shaped and bounded throughout — guards on `NodeSelector`/`RequireGPU`/`RuntimeClassName` preserve user intent, the pre-flight List is `Limit:1` with a 5s timeout in `defaults.GPUNodeDetectionTimeout`, and the path is fail-open on API errors. TOCTOU is explicitly documented and handled: the wait-timeout error names the auto-injected selector so operators know why the Job pended.

The `hasGPUData` refactor (`r.Any().(int)` → `m.Subtypes[i].GetInt64(...)`) fixes exactly the AGENTS.md anti-pattern about type switches on `Any()` missing the `float64` JSON round-trip case — the dedicated float64 test entry is direct evidence the prior code mis-fired on parsed snapshots. `detectGPUPlacementMismatch` keys on the `nvidia.com/gpu.` prefix so it covers both `gpu.present` and `gpu.product.*` labels.

Tests are thorough: 5 sub-tests for inject covering each guard branch, table-driven coverage for `hasGPUData` (int/int64/float64/zero/missing) and `detectGPUPlacementMismatch` (`.present`, `.present.true`, `.product.*`, no-mismatch). Live verified 20/20 sequential on EKS GB200; the parallel-stress failure is an unrelated pre-existing Job-name collision, correctly characterized as not a regression. Confirmed by reading the call graph: `measureWithAgent` and `DeployAndGetSnapshot` are separate entry points that each wrap `deployAndWaitForResult` and parse independently — `warnOnGPUPlacementMismatch` fires at most once per snapshot path. CI green including GPU smoke/training/inference tests, which are the most relevant gate for a snapshot-scheduling change.

Minor (non-blocking): the docs checkbox is unchecked even though auto-injection is a small user-visible behavior change for unconstrained invocations. Worth a short note in the snapshot section of `docs/user/cli-reference.md` in a follow-up.

njhensley:
> Minor (non-blocking): the docs checkbox is unchecked even though auto-injection is a small user-visible behavior change for unconstrained invocations. Worth a short note in the snapshot section of docs/user/cli-reference.md in a follow-up.

Will follow-up with small docs update.

PR #1201, opened 2026-06-05
fix(validators): reorder validation phases and fix stale paths and documentation issues

yuanchen8911:
## Summary

Two validator fixes surfaced while running `aicr validate --phase all` on a single-node EKS H100 cluster after a gpu-operator upgrade:

1. **Run the performance phase last** — reorder the default phase order from `deployment → performance → conformance` to **`deployment → conformance → performance`**.
2. **Fix stale `Equivalent: kubectl apply -f …` hint paths** in conformance check artifacts and evidence script (`docs/conformance/cncf/manifests/` and `pkg/evidence/scripts/manifests/` don't exist).

## Motivation / Context

On a single GPU node (8×H100), `--phase all` had **conformance `dra-support` fail** with `Unschedulable: cannot allocate all claims`. Root cause: the **performance** phase ran immediately before conformance; its inference-perf benchmark uses **all 8 GPUs** (8 vLLM workers via DRA) and the DynamoGraphDeployment teardown releases those ResourceClaims **asynchronously** — so `dra-support`'s 1-GPU test pod couldn't schedule. "Worked before" only because earlier runs were standalone phases. DRA itself was verified healthy (a standalone ResourceClaim pod allocated an H100 + ran `nvidia-smi`). Running performance **last** frees GPUs for conformance and keeps a flaky perf phase from blocking conformance under `--fail-fast`.

The hint paths are cosmetic ("equivalent kubectl" text); the checks build resources **in-code via the K8s API**, not from those files.

## Changes

**Phase order → `deployment, conformance, performance`** (kept consistent across every surface):
- `pkg/validator/phases.go` `PhaseOrder` (+ rationale comment) and `pkg/validator/validator_test.go` `TestPhaseOrder`.
- `pkg/client/v1`: SDK facade — `ValidateState` method godoc, the package-level overview godoc, and `TestValidateState_PhaseSelection` "unset runs all phases" (was a **CI-breaking** stale assertion).
- `pkg/validator/v1/job_plan.go` `Plan()` literal (test-only helper; removes latent drift).
- Docs: `docs/contributor/validator.md` (order + why-last rationale), `docs/user/validation.md` (intro, "When to validate" table, "equivalent to" line), `docs/design/002-validatorv2-adr.md` (pipeline list + execution diagram).
- **Not changed (deliberate):** `pkg/evidence/attestation/AllPhases` stays frozen — it's the canonical *output* order for reproducible attestation bytes; reordering it would change provenance output. Added a comment documenting that it's intentionally independent of execution `PhaseOrder`.

**Stale manifest hint paths → `pkg/evidence/cncf/scripts/manifests/`:**
- `dra_support_check.go` → `dra-gpu-test.yaml`; `pod_autoscaling_check.go` → `hpa-gpu-test.yaml`.
- `cluster_autoscaling_check.go`: referenced a non-existent `hpa-gpu-scale-test.yaml`; since no cluster-autoscaling manifest exists and `hpa-gpu-test.yaml` is the *pod*-autoscaling manifest, the misleading apply hint is **removed** — the artifact now states the resources are built in-code (HPA + GPU deployment + Karpenter NodePool).
- `collect-evidence.sh`: fixed all three `**Test manifest:**` text paths (dra / gang-scheduling / hpa) to the real `pkg/evidence/cncf/scripts/manifests/` location.

## Type of Change

- [x] Bug fix (conformance false-negative on single-node clusters + misleading/stale hint paths)

## Component(s) Affected

- [x] Validator (`pkg/validator`, `pkg/validator/v1`, `validators/conformance`)
- [x] SDK facade (`pkg/client/v1`)
- [x] Docs + evidence tooling

## Testing

```bash
go test ./pkg/client/v1 ./pkg/validator ./pkg/validator/v1 ./pkg/evidence/attestation ./validators/conformance   # ok
golangci-lint run -c .golangci.yaml ./pkg/client/v1/... ./pkg/validator/... ./pkg/evidence/attestation/... ./validators/conformance/...   # 0 issues
env -u GITLAB_TOKEN make test   # pass
```

**Validated end-to-end** on EKS H100 (aicr3, post gpu-operator v26.3.2 upgrade): reordered `aicr validate --phase all` **passed all three phases** — deployment 4/4, **conformance 11/11 (incl. `dra-support`)**, performance 1/1 (107,350 tok/s / 1,939 ms). The reorder resolved the contention exactly as intended.

## Risk Assessment

- [x] **Low** — phase-order slice change + artifact/string/doc edits; no control-flow or deployment-topology change. Attestation output order intentionally unchanged.

## Checklist

- [x] Tests pass locally (incl. the previously-failing `pkg/client/v1` assertion)
- [x] Linter passes (0 issues)
- [x] Docs updated (contributor + user + ADR, all order references aligned)
- [x] Commit cryptographically signed (`-S`)


mchmarny (approved):

njhensley (approved):
## Review — #1201 `fix(validators): reorder validation phases and fix stale paths`

**Verdict: Approve.** The diff is small, correct, well-tested, and the one subtle correctness risk (attestation reproducibility) is handled deliberately and verified. A few optional doc-consistency nits below — none block merge.

Reviewed the actual PR diff only: commit `1bdc29f2`, 16 files, +46/−24. (Note for other reviewers: a three-dot `git diff` against a stale `main` pulls in already-landed sibling commits — the FailFast default change (#1198) and the inference-perf tuning (#1196) are **not** part of this PR.)

### What's right

- **Single source of truth updated correctly.** `pkg/validator.PhaseOrder` is the live execution driver (`validator.go` falls back to it when `--phase` is unset). The flip to `deployment → conformance → performance` is applied there, and the parallel literal in the legacy-but-still-imported `pkg/validator/v1/job_plan.go::Plan()` is kept in sync (removes latent drift; `Plan()` has no non-test callers, so no behavior change).
- **Attestation order is correctly left frozen.** `attestation.AllPhases` intentionally stays `deployment, performance, conformance`. Traced both consumers — `predicate.go` copies into a phase-keyed `map` (JSON-marshalled with sorted keys) and `verifier/report.go` looks up by key — so execution order genuinely cannot mis-map, drop, or reorder a phase's results, and predicate bytes stay reproducible. The added comment documenting this decoupling is accurate and worth having.
- **Tests match production and are order-sensitive.** `TestPhaseOrder` and `TestValidateState_PhaseSelection` ("unset runs all phases") both assert the new order against index-sensitive comparisons; no test anywhere still asserts the old execution sequence (`phases_test.go` checks parsing/dedup only; the recipe phase-floor test iterates config-declaration order, not execution order — unaffected).
- **CodeRabbit's one comment is resolved.** All three `**Test manifest:**` lines in `collect-evidence.sh` now point to `pkg/evidence/cncf/scripts/manifests/`; no stale `pkg/evidence/scripts/manifests/` references remain in the file.
- **The `cluster_autoscaling_check.go` free-text swap is correct.** There is genuinely no static cluster-autoscaling manifest (the only `hpa-gpu-scale-test.yaml` is an unrelated KWOK fixture), so replacing the bogus `kubectl apply -f` hint with the "constructed via the Kubernetes API" note is accurate; `recordRawTextArtifact` treats that arg as informational text, so no behavior change.
- **The three frozen CNCF evidence files are correctly *not* touched.** `docs/conformance/cncf/v1.35/nim-eks/evidence/{dra-support,pod-autoscaling,gang-scheduling}.md` still contain the old path, but they are point-in-time conformance-submission snapshots (committed once, only ever touched by a copyright sweep) and `collect-evidence.sh` is deprecated and doesn't regenerate them. Rewriting them would falsify a historical record. This is intentional, not a gap.

### Optional nits (pre-existing drift this PR could tidy, since it's the one establishing the new order)

These files aren't in the diff, but they now contradict the order this PR makes canonical:

- **`docs/integrator/go-library.md:113`** — the `ValidateState` godoc example says it runs "(Deployment, Performance, Conformance) **in canonical order**." That's now wrong; it should read `Deployment, Conformance, Performance`. (Most worth fixing — it explicitly asserts execution order in public-API docs.)
- **`README.md` (lines 63, 96, 118)** and **`demos/valid.md:153`** — prose / expected-output still list `deployment, performance, conformance` as the run sequence.
- **`docs/user/cli-reference.md`** — the `aicr validate` phase table and `--phase` enum still list the old order (descriptive, low impact, but this is the natural PR to align it).
- **PR title is 87 chars**, over the repo's 70-char convention. The commit headline (`run performance phase last; fix stale conformance manifest hint paths`) is already a good shorter form.

### Out of scope, but worth a follow-up issue (not a defect in this diff)

The reorder fixes the **default** path, which is the right call to unblock single-node runs. It's worth recording that it doesn't remove the underlying hazard: `ParsePhaseSelection` preserves user-supplied order, so an explicit `aicr validate --phase performance --phase conformance` would still run perf first and can re-trigger the same "cannot allocate all claims" starvation. The deeper cause is that the perf phase's workload teardown releases DRA ResourceClaims asynchronously (no readiness gate on claim release) and `dra-support` fails fast on transient `Unschedulable`. A follow-up that makes perf teardown block until its ResourceClaims are released (and/or lets `dra-support` tolerate transient unschedulability) would make phase ordering a latency choice rather than a correctness crutch. None of that is required for this PR — flagging for a tracking issue.

Issue #1152, opened 2026-06-02
feat(verify): KMS public-key verification (aicr verify --key)


lockwobr:
## Problem

KMS-backed signing (#407) lets CI sign a bundle with a cloud KMS key, but `aicr verify` cannot verify the result. The verifier (`pkg/bundler/verifier/verifier.go` → `verifySigstoreBundle`) only verifies keyless Sigstore bundles against the public-good trusted root (`trust.GetTrustedMaterial()`); there is no public-key verification path.

## Proposal

Add a `--key <kms-uri-or-pubkey>` option to `aicr verify` that verifies a key-signed bundle against the corresponding public key (KMS URI or a local PEM), mirroring `cosign verify --key`.

## Behavior

- `--key` selects public-key verification instead of keyless certificate-identity verification.
- Accepts the same KMS URI schemes as signing (`awskms://`, `gcpkms://`, `azurekms://`, `hashivault://`) and local public-key files.
- Mutually exclusive with the keyless identity-pinning path (`--certificate-identity-regexp`).

## Dependencies

- Pairs with: #407 (KMS-backed signing). This is its verification counterpart.

## Success criteria

- A bundle signed with `aicr bundle --signing-key gcpkms://...` verifies with `aicr verify --key gcpkms://...`.
- `make test` / `make lint` pass.

Issue #1150, opened 2026-06-02
feat(server): bundle attestation for /v1/bundle (non-interactive signing)


lockwobr:
## Problem

The `/v1/bundle` endpoint (aicrd) does not attest or sign bundles at all. The handler builds `aicr.BundleOptions{Config, OutputDir, Timeout}` with **no `Attester`** (`pkg/server/bundle_handler.go`), so server-generated bundles are always unsigned — there is no server-side equivalent of `aicr bundle --attest`. The OpenAPI spec reflects this, framing attestation as a CLI-only flow that "requires OIDC interactive or ambient authentication" (`api/aicr/v1/server.yaml`).

This blocks the server from participating in the enterprise signing use cases tracked by the rest of the epic: a CI pipeline or platform that bundles via the API cannot get a signed artifact back.

## Why this is separate from the CLI signing work

The CLI's default signing path is **interactive browser OIDC**, which is a non-starter for a long-running daemon. So server-side attestation is not just "pass the existing attester through" — it needs a signing identity that works without a human at a browser:

- **KMS-backed signing** (depends on #407) — the server signs with a configured KMS key (`awskms://` / `gcpkms://` / `azurekms://` / `hashivault://`). This is the natural fit for a daemon and the most likely default.
- **Request-supplied / ambient OIDC** — a pre-fetched identity token (e.g. workload identity) passed on the request, or ambient OIDC when aicrd itself runs in a CI context. Carries trust/expiry questions (the server signs on behalf of a caller) that need a design pass.

The client facade already has the seam: `aicr.BundleOptions.Attester` and `MakeBundle` honor an injected attester (`pkg/client/v1/bundle.go`). The missing piece is *how the server constructs that attester* and *how the request/config expresses it*.

## Scope (proposed)

- Decide and document the server signing identity model (KMS key from server config vs. request-supplied identity token vs. ambient), including the trust boundary.
- Construct an `Attester` in `bundleHandler.HandleBundles` (or the client facade) from that model and pass it through `BundleOptions`.
- Once the server can sign, expose the signing-target knobs that already exist at the library layer: private Fulcio/Rekor URLs (#408) and `--tlog-upload=false` (#409) as request/config fields where they make sense.
- OpenAPI updates: new request fields and/or server config, plus revised attestation description in `api/aicr/v1/server.yaml`.
- Docs: `docs/user/api-reference.md`, `docs/contributor/api-server.md`.

## Dependencies

[...]

Issue #409, opened 2026-03-13
feat(bundle): air-gapped signing with --tlog-upload=false for restricted networks


mchmarny:
## Problem

Fully disconnected (air-gapped) environments cannot upload to any transparency log—public or private. Currently there is no way to sign bundles without network connectivity to a Rekor instance.

## Proposal

Add a `--tlog-upload` flag (default `true`) to `aicr bundle` (and/or `aicr bundle-sign`) that disables transparency log uploads when set to `false`. Combined with KMS-backed signing (#407), this enables fully offline bundle attestation.

### New Flag

| Flag | Type | Default |
|------|------|---------|
| `--tlog-upload` | bool | `true` |

### Behavior

- When `--tlog-upload=false`, AICR skips the Rekor transparency log upload during signing
- Must be combined with `--signing-key <kms-uri>` (keyless signing requires Fulcio, which requires network)
- No operational metadata leaves the network perimeter
- The resulting bundle is still a valid cosign bundle, verifiable with `--insecure-ignore-tlog` or offline verification

### Validation Combinations

| Signing Method | tlog-upload | Valid? | Use Case |
|----------------|-------------|--------|----------|
| Keyless (OIDC) | true | Yes | Default public Sigstore |
| Keyless (OIDC) | false | **No** | Keyless requires Fulcio network access |
| KMS key | true | Yes | KMS + public/private transparency log |
| KMS key | false | Yes | **Air-gapped**: fully offline signing |

AICR should reject the invalid combination (keyless + no tlog) at flag validation time with a clear error message.

### Verification in Air-Gapped Environments

Verification uses the KMS public key and skips tlog verification:
```bash
cosign verify --key <kms-uri> --insecure-ignore-tlog <bundle>
```

## Implementation Notes

- Cosign supports `--tlog-upload=false` natively
- Pass through to cosign signing options alongside the existing signing path
- Depends on KMS-backed signing support (see related issue)

## Dependencies

- Requires: #407 (KMS-backed signing — `--signing-key` flag)
- Complementary: #408 (Private Sigstore infrastructure — `--fulcio-url` / `--rekor-url`)

## Success Criteria

- `aicr bundle --signing-key gcpkms://... --tlog-upload=false` produces a valid signed bundle with no network calls to Rekor
- Invalid combination `--tlog-upload=false` without `--signing-key` is rejected with a clear error
- Bundle is verifiable offline with the KMS public key
- `make test` and `make lint` pass with no regressions

[...]

PR #1205, opened 2026-06-05
feat(bundle): KMS-backed signing via --signing-key (#407)

### LINKED ISSUE #407: feat(bundle): KMS-backed signing for CI/CD environments without OIDC
## Problem

CI environments without OIDC token support—such as Jenkins, TeamCity, or internal pipelines—cannot attest bundles using keyless (Fulcio) signing. There is currently no way to sign bundles with a KMS-backed key.

## Proposal

Add a `--signing-key` flag to `aicr bundle` (and/or `aicr bundle-sign`) that accepts a KMS URI for cosign-compatible key-based signing. This ensures key material never touches disk.

### Supported Backends

All backends natively supported by cosign via URI scheme:

| Backend | URI Scheme |
|---------|------------|
| AWS KMS | `awskms://arn:aws:kms:...` |
| GCP KMS | `gcpkms://projects/...` |
| Azure Key Vault | `azurekms://...` |
| HashiCorp Vault | `hashivault://...` |

### Behavior

- `--signing-key <kms-uri>` triggers KMS-backed signing instead of keyless OIDC flow
- Cosign resolves the URI to the appropriate KMS provider at runtime
- The resulting signature bundle uses the same format as keyless signing (cosign bundle)
- Verification works with the corresponding public key (`cosign verify --key <kms-uri>`)

### Mutual Exclusivity

`--signing-key` is mutually exclusive with keyless OIDC signing flags. Providing both is a validation error.

## Implementation Notes

- Cosign's `SignerVerifier` interface already supports KMS URIs via `cosign.LoadKeyFromKMS()`
- The signing path in `pkg/evidence/` or `pkg/bundler/` would branch on whether `--signing-key` is set
- No new dependencies required—cosign's KMS support is built-in

## Success Criteria

- `aicr bundle --signing-key gcpkms://projects/my-proj/locations/global/keyRings/kr/cryptoKeys/k` produces a valid signed bundle
- Signature is verifiable with `cosign verify --key <same-kms-uri>`
- Works in CI without OIDC token availability
- `make test` and `make lint` pass with no regressions

lockwobr:
## Summary

Adds `aicr bundle --attest --signing-key <kms-uri>` to sign bundles with a cloud-KMS-backed key (AWS / GCP / Azure KMS) instead of keyless OIDC, for CI/CD environments that have no OIDC token.

## Motivation / Context

CI environments without OIDC (Jenkins, TeamCity, internal pipelines) cannot attest bundles via keyless Fulcio signing. This adds a key-based signing path and is the foundation issue of the closed-supply-chain epic: it unblocks air-gapped signing (#409), server-side `/v1/bundle` signing (#1150), and KMS verification (#1152).

Fixes: #407
Related: #1149 (epic), #409, #1150, #1152

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

## Implementation Notes

Signing is decomposed into two orthogonal, composable axes consumed by a new `SignStatementWith` primitive in `pkg/bundler/attestation`:

- **`SigningIdentity`** (`identity.go`): supplies the signing keypair and, for keyless, the Fulcio certificate provider. `keylessIdentity` uses an ephemeral key + Fulcio; `kmsIdentity` uses a KMS key and returns no certificate.
- **`TransparencyPolicy`** (`transparency.go`): selects the Rekor transparency log, or none (the no-tlog policy is reserved for offline signing, #409).

`SignStatement` (keyless) now delegates to `SignStatementWith` with **no behavioral change**; `KMSAttester` pairs a KMS identity with Rekor. Key points:

- KMS-signed bundles carry **public-key** verification material (no Fulcio certificate) and **upload to Rekor by default**, mirroring keyless.
- The KMS keypair adapts a `sigstore/sigstore` KMS `SignerVerifier` to sigstore-go's `sign.Keypair`; signing passes a precomputed SHA-256 digest via `options.WithDigest` so there is no double-hashing (the signature verifies over `sha256(data)`).
- Supported schemes: `awskms://`, `gcpkms://`, `azurekms://`. **HashiCorp Vault (`hashivault://`) is intentionally excluded** because its client libraries are MPL-2.0, which the project license policy disallows. (Vault/OpenBao can be supported later via sigstore's out-of-process KMS plugin, which keeps the binary MPL-free.)
- `--signing-key` is mutually exclusive with `--identity-token`, `--oidc-device-flow`, and `--fulcio-url`; `--rekor-url` **may** be combined with it (KMS + private Rekor).
- `kms.Get` failures are classified: unknown scheme (`kms.ProviderNotFoundError`) → `ErrCodeInvalidRequest`; provider init/auth/network failure → `ErrCodeUnavailable` (matters for the future server path, #1150).

Verification today is via `cosign verify ... --key <kms-uri>`; native `aicr verify --key` is tracked in #1152. The new sigstore KMS provider dependencies are vendored in a separate commit.

## Testing

```bash
make qualify
```

`make qualify` passes (test `-race`, lint, e2e, scan, repo checks; 0 failed tests). New tests are fully offline (local ECDSA signer + no-tlog policy; no network/cloud/cluster):

- `pkg/bundler/attestation`: **78.1%** coverage; every new exported function/method covered (the key-only/no-tlog path is asserted end-to-end to produce public-key verification material with no tlog entries; `classifyPublicKey` fails closed on non-P256 / non-2048 keys).
- Project-wide `make test-coverage`: **75.9%** vs the 75% floor.
- `make scan`: no new findings from the KMS provider deps. (Grype reports two pre-existing `k8s.io/kubernetes` advisories that arrived via the rebase onto `main`, unrelated to this change.)

## Risk Assessment

**Rollout notes:** Additive and opt-in behind a new `--signing-key` flag. The keyless signing path is unchanged (it now delegates to the shared composer with identical behavior). No migration required.

## Checklist


.claude/skills/creating-slide-decks/skeleton.html
mchmarny, line 1: Out of scope for this PR (KMS signing). This copyright header addition and the matching one on `cmd/gate/chainsaw-config.yaml` belong in a separate housekeeping commit so the KMS change stays focused. Not blocking.

pkg/bundler/attestation/keylessidentity.go
```diff
@@ -0,0 +1,58 @@
...
+
+import (
+	"context"
+
+	"github.com/sigstore/sigstore-go/pkg/sign"
+
+	"github.com/NVIDIA/aicr/pkg/defaults"
+	"github.com/NVIDIA/aicr/pkg/errors"
+)
+
```
mchmarny, line 25: nit: the third parameter is silently discarded (`_ string`). The doc says "accepted for call-site symmetry" but the only caller (`SignStatement`) builds `NewRekorPolicy(opts.RekorURL)` separately — there's no symmetry being preserved. Drop the dead parameter, or if reserved for a future signature, leave a `// reserved for ...` note.

pkg/bundler/attestation/kmsidentity.go
```diff
@@ -0,0 +1,152 @@
...
+// lazily on first Keypair() call so a bad URI or missing provider credentials
+// surface at sign time. HashiCorp Vault is unsupported (MPL-2.0 license).
+func NewKMSIdentity(keyURI string) SigningIdentity { return &kmsIdentity{keyURI: keyURI} }
+
+// Keypair resolves the KMS provider for the key URI, reads the public key, and
+// returns a sign.Keypair backed by the KMS signing seam. SHA-256 is fixed as
+// the bundle digest algorithm (the cloud KMS defaults — ECDSA P-256 / RSA-2048
+// — all sign over a SHA-256 digest).
+func (k *kmsIdentity) Keypair(ctx context.Context) (sign.Keypair, error) {
+	// Bound the provider lookup + PublicKey RPC so a caller that passes an
```
mchmarny, line 63: `kmsIdentity` is stored once on `KMSAttester` but `Keypair` re-runs `kms.Get` + the `PublicKey` RPC on every `Attest` call. The single-bundle CLI path is unaffected, but the server path (#1150) and any future batch caller will pay 2× RPC per sign. Consider caching the resolved `sign.Keypair` behind a `sync.Once` (or invalidate on auth-error to handle credential rotation). Follow-up, not blocking this PR.

pkg/bundler/attestation/kmsidentity.go
```diff
@@ -0,0 +1,152 @@
...
+// (awskms:// | gcpkms:// | azurekms://). The provider and key are resolved
+// lazily on first Keypair() call so a bad URI or missing provider credentials
+// surface at sign time. HashiCorp Vault is unsupported (MPL-2.0 license).
+func NewKMSIdentity(keyURI string) SigningIdentity { return &kmsIdentity{keyURI: keyURI} }
+
+// Keypair resolves the KMS provider for the key URI, reads the public key, and
+// returns a sign.Keypair backed by the KMS signing seam. SHA-256 is fixed as
+// the bundle digest algorithm (the cloud KMS defaults — ECDSA P-256 / RSA-2048
+// — all sign over a SHA-256 digest).
+func (k *kmsIdentity) Keypair(ctx context.Context) (sign.Keypair, error) {
```
mchmarny, line 62: When invoked from `SignStatementWith`, `ctx` is already wrapped with `defaults.SigstoreSignTimeout`, so this second `context.WithTimeout` at the same duration is effectively a no-op (tighter deadline wins, but they're equal so the parent deadline is preserved). The comment correctly notes it nests harmlessly — only flagging because if `SigstoreSignTimeout` is ever shortened for the signing call but the KMS resolution genuinely needs longer, this nesting won't help. Consider a separate `defaults.KMSResolveTimeout` if those budgets ever need to diverge.

pkg/cli/bundle.go
```diff
@@ -643,6 +678,11 @@ Package with explicit tag (overrides CLI version):
 				Sources:  cli.EnvVars("COSIGN_IDENTITY_TOKEN"),
 				Category: catDeployment,
 			},
+			&cli.StringFlag{
+				Name:     flagSigningKey,
+				Usage:    "Sign --attest bundles with a KMS-backed key (awskms:// | gcpkms:// | azurekms://) instead of keyless OIDC, for CI/CD without OIDC. Mutually exclusive with --identity-token, --oidc-device-flow, --fulcio-url. Combine with --rekor-url to log to a private Rekor.",
```
mchmarny: nit: the `--signing-key` Usage string is comprehensive on what's supported but doesn't mention that verification today still requires `cosign verify --key` (native `aicr verify --key` is #1152). One short sentence here would save users a round trip — e.g., "Verify with `cosign verify --key <uri>` until `aicr verify --key` lands (#1152)."

pkg/bundler/attestation/kms.go
```diff
@@ -0,0 +1,73 @@
...
+	// INFO (unlike the keyless Fulcio SAN, which stays at Debug).
+	slog.Info("bundle attestation signed successfully (KMS)", "identity", res.Identity)
+	return res.BundleJSON, nil
+}
+
+// Identity returns the KMS key URI used for signing.
+func (k *KMSAttester) Identity() string { return k.keyURI }
+
+// HasRekorEntry returns true — KMS attestations upload to Rekor by default.
+func (k *KMSAttester) HasRekorEntry() bool { return true }
```
mchmarny: `HasRekorEntry()` is hardcoded to `true` here, but the no-tlog policy (`NewNoTLogPolicy`) exists in `transparency.go` and `KMSAttester` is the consumer that will need it for #409 (offline signing). Worth making the `tlog` field drive this (`_, ok := k.tlog.(noTLogPolicy); return !ok`) now so #409 doesn't have to revisit the contract. Non-blocking.

mchmarny (commented):
Clean refactor — the `SigningIdentity` × `TransparencyPolicy` split keeps keyless behaviorally unchanged while making KMS a single composition apart, and the error classification at `kms.Get` (InvalidRequest vs Unavailable) is right for the future server path. Fail-closed key-type validation (P-256 / RSA-2048 only) with explicit ed25519/RSA-4096 test coverage is solid.

A few non-blocking observations inline: two unrelated copyright-header additions worth splitting out, a dead `_ string` parameter on `NewKeylessIdentity`, per-Attest KMS RPC re-resolution worth caching before #1150, hardcoded `HasRekorEntry()` that #409 will want driven from `tlog`, and a Usage-string nit about verification still needing `cosign verify --key`. None block merge.

.claude/skills/creating-slide-decks/skeleton.html
lockwobr: Reverted both `skeleton.html` and `cmd/gate/chainsaw-config.yaml` to origin/main; the license-header additions were swept in by a hook and are now removed from this PR. Thanks for catching the scope creep.

pkg/bundler/attestation/keylessidentity.go
lockwobr: Dropped the dead third parameter. `NewKeylessIdentity` is now `(oidcToken, fulcioURL string)`, with the caller and test updated; Rekor stays a TransparencyPolicy concern.

pkg/bundler/attestation/kmsidentity.go
lockwobr: Good call. Filed as a follow-up in #1214 (KMS e2e + hardening). It touches credential-rotation semantics and is non-blocking for the single-bundle CLI path, so keeping the eager per-Attest resolution for now.

pkg/bundler/attestation/kmsidentity.go
lockwobr: Tracked in #1214 too. Both resolve and sign currently share SigstoreSignTimeout; I'll split out a dedicated KMSResolveTimeout if/when those budgets need to diverge.

mchmarny (approved):
Full review (CodeRabbit was down). Walked through every changed `pkg/bundler/attestation/*.go`, `pkg/cli/bundle.go`, `docs/user/cli-reference.md`, and re-verified each open review thread against the head SHA.

**All 7 outstanding comments are addressed in the code, even though the threads weren't marked resolved on GitHub:**

| Thread | Verdict |
|---|---|
| Vault exclusion wording (CodeRabbit) | ✅ `cli-reference.md:2147` now reads "deliberate, ongoing exclusion… MPL-2.0" |
| `PropagateOrWrap` for `BuildStatement` (CodeRabbit) | ✅ `kms.go:56` |
| `PropagateOrWrap` on `SignData` (CodeRabbit) | ✅ `kmskeypair.go:101` |
| Precedence test pins `SigningKey` over keyless (CodeRabbit) | ✅ `resolver_test.go` — both tests now include `IdentityToken:"should-be-ignored"` + `DeviceFlow:true` |
| Empty-URL fallback assertion (CodeRabbit) | ✅ `transparency_test.go` — white-box assert on `rekorPolicy.url == defaults.SigstoreRekorURL` |
| `HasRekorEntry` driven by tlog policy (my own #409 nit) | ✅ `kms.go:76` — `_, noTLog := k.tlog.(noTLogPolicy); return !noTLog` |
| `cosign verify --key` hint in `--signing-key` Usage (my own nit) | ✅ `bundle.go:702` |

Resolving the 6 stale threads since the concerns are addressed.

**One thing worth calling out beyond the comments**: the new `extractIssuerExtension` in `signing.go` properly ASN.1-decodes the current Fulcio OID `1.3.6.1.4.1.57264.1.8` as a UTF8String (with trailing-byte rejection) and falls back to the legacy OID as raw bytes. The previous code treated both as raw bytes, which would have silently embedded the ASN.1 tag/length prefix in the issuer string for any Fulcio cert using the current OID. Nice quiet fix.

`validateSigningKeyExclusivity` correctly fixes the config-bypass concern from the earlier CodeRabbit comment — checks both `cmd.IsSet` (for empty-value fail-fast) and the resolved opts (catching config-sourced keyless flags). The `--rekor-url` deliberate omission is well-justified in the comment (enterprise "KMS key + private Rekor").

CI: all required checks green except `tests / E2E` still in_progress; CodeQL conclusion is `neutral` (informational, not blocking).

Approving. Architecture (composable `SigningIdentity` × `TransparencyPolicy`) sets up #409 and #1150 cleanly; failure-mode classification (`ProviderNotFoundError` → `ErrCodeInvalidRequest` vs. init failures → `ErrCodeUnavailable`) is exactly what the future server path needs.

Issue #1207, opened 2026-06-06
Add aicr recipe list to enumerate curated overlays as JSON (with criteria filters)


atif1996:
## Problem

Tools that consume AICR as a library/CLI with external `--data` (e.g. an internal
recipe repo that packages scenarios as OCI bundles) need to know **which criteria
combinations resolve to a curated leaf overlay** vs. a generic fallback, so they can
build a bundle catalog automatically instead of hand-maintaining one.

Today there is no way to enumerate the catalog. The only workarounds are both fragile:

1. **Discovering valid enum values** — the accelerator/service lists only appear as
   `e.g.` hint text in `aicr recipe --help` (`Accelerator/GPU type (e.g. a100, b200,
   gb200, h100, h200, l40, rtx-pro-6000)`). That's not a guaranteed-complete machine API,
   so a new accelerator (e.g. a future `l40s`) may not be discoverable programmatically.
2. **Distinguishing curated from fallback** — we must probe every
   `service × accelerator × intent × os` combination, run `aicr recipe`, and inspect
   `metadata.appliedOverlays` to check whether a dedicated `<accel>-<service>` leaf was
   applied (curated) or only `<accel>-any` + a generic `<service>-training` overlay
   (fallback). `aicr recipe` resolves *anything* via fallback, so "it resolves" is not a
   usable filter.

## Request

Add a read-only `aicr recipe list` (or `aicr catalog`) subcommand that enumerates the
**curated overlays** and emits **JSON** (machine-readable, `--format json|yaml`, default
text table). Each entry should carry at least:

- the overlay `metadata.name` (the leaf)
- its `criteria` (`service`, `accelerator`, `os`, `intent`, `platform`)
- whether it is a curated leaf vs. a generic/base overlay
- (nice to have) the source: embedded (OSS) vs. external `--data`

It should honor `--data` (so external overlays are included with correct precedence) and
accept **optional filter flags** — `--service`, `--accelerator`, `--intent`, `--os` — that
narrow the same output. Per-cloud filtering is just a view over the full catalog, so
supporting both the full dump (default) and filters costs little extra and covers every
caller.

### Example

```console
$ aicr recipe list --format json --service eks --intent training
[
  {"name":"gb200-eks-ubuntu-training","criteria":{"service":"eks","accelerator":"gb200","os":"ubuntu","intent":"training"},"curated":true,"source":"embedded"},
[...]

PR #1208, opened 2026-06-07
feat(recipe): add aicr recipe list subcommand for catalog enumeration

rsd-darshan:
## Summary

Adds a read-only `aicr recipe list` subcommand that enumerates overlay recipes in the catalog with machine-readable output (JSON, YAML, table), solving the catalog discoverability gap in #1207.

## Motivation / Context

Tools consuming AICR as a library or CLI with external `--data` had no programmatic way to discover which criteria combinations resolve to a curated leaf overlay versus a generic fallback. The only workarounds were parsing `--help` hint text (not a complete API) or probing every `service × accelerator × intent × os` combination and inspecting `appliedOverlays` — both fragile and expensive.

Fixes: N/A
Related: #1207

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Implementation Notes

**Source tracking (`pkg/recipe/metadata_store.go`):** Added `OverlaySources map[string]string` to `MetadataStore`, populated with `provider.Source(path)` during the catalog walk alongside the existing criteria-registry staging. This avoids re-walking the provider at list time.

**Curated definition (`pkg/recipe/catalog.go`):** An overlay has `IsLeaf: true` when no other overlay in the catalog lists it as `spec.base` — i.e., it is a maximal leaf. Intermediate overlays (e.g., `eks-training`, which is the base of `h100-eks-training`) get `IsLeaf: false`. This matches the issue's intent of distinguishing dedicated leaf recipes from generic shared layers.

**Filter semantics (`matchesCatalogFilter`):** Uses simple equality rather than the asymmetric `Criteria.Matches` logic used for recipe resolution. `--accelerator h100` returns only overlays explicitly specifying `h100`; it does not match overlays whose accelerator is `any`. This is the correct semantic for a discovery query: "show me what targets h100", not "show me what a query for h100 would pick up via wildcard".

**CLI output (`pkg/cli/recipe_list.go`):** Default format is `table` (human-friendly with tabwriter); `--format json` and `--format yaml` are available for machine consumption. Writes to `cmd.Root().Writer` per project conventions. Filter flags are validated through the client's criteria registry so `--data` overlay values are accepted.

**Facade (`pkg/client/v1`):** `Client.ListCatalog` follows the same inflight/lock protocol as `LoadCatalog` and reuses the existing `toInternalCriteria`/`WrapCriteria` translation helpers to avoid parallel projection logic.

## Testing

```bash
make qualify
```

All tests pass, linter clean. New unit tests in `pkg/recipe/catalog_test.go` cover:
- Full enumeration (no filter)
- Source field propagation
- Per-dimension filters: service, accelerator, OS, service+intent combined
- Empty result when no overlays match
- All-`any` filter behaves identically to no filter
- Ascending name sort
- `matchesCatalogFilter` nil filter, exact match, mismatched dimension, wildcard-overlay vs specific-filter

## Risk Assessment

**Rollout notes:** Purely additive. No existing commands or API endpoints are modified. The only change to `MetadataStore` is a new field populated during the existing catalog walk; no behavior changes to recipe resolution or bundling.

## Checklist

### DIFF
--- a/pkg/cli/recipe_list.go
+++ b/pkg/cli/recipe_list.go
@@ -1,0 +1,257 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/urfave/cli/v3"
+
+	aicr "github.com/NVIDIA/aicr/pkg/client/v1"
+	"github.com/NVIDIA/aicr/pkg/errors"
+	"github.com/NVIDIA/aicr/pkg/recipe"
+	"github.com/NVIDIA/aicr/pkg/serializer"
+)
+
+func recipeListCmd() *cli.Command {
+	return &cli.Command{
+		Name:  cmdNameRecipeList,
+		Usage: "List recipe overlays in the catalog.",
+		Description: `Enumerate all overlay recipes in the catalog and their criteria.
+
+Each entry shows the overlay name, its criteria dimensions, whether it is a
+leaf overlay (no other overlay inherits from it), and its data source
+(embedded or external).
+
+Filter flags narrow the output to overlays that carry the specified criteria
+value. Unspecified flags match all overlays for that dimension.
+
+Examples:
+
+List all overlays (table format):
+  aicr recipe list
+
+List all overlays as JSON:
+  aicr recipe list --format json
+
+Filter to EKS training overlays:
+  aicr recipe list --service eks --intent training
+
+Filter to H100 overlays as JSON:
+  aicr recipe list --accelerator h100 --format json
+
+Include overlays from an external data directory:
+  aicr recipe list --data /path/to/custom-recipes`,
+		Flags: []cli.Flag{
+			withCompletions(&cli.StringFlag{
+				Name:     flagService,
+				Usage:    fmt.Sprintf("Filter by service type (e.g. %s)", strings.Join(recipe.GetCriteriaServiceTypes(), ", ")),
+				Category: catQueryParameters,
+			}, recipe.GetCriteriaServiceTypes),
+			withCompletions(&cli.StringFlag{
+				Name:     flagAccelerator,
+				Aliases:  []string{"gpu"},
+				Usage:    fmt.Sprintf("Filter by accelerator type (e.g. %s)", strings.Join(recipe.GetCriteriaAcceleratorTypes(), ", ")),
+				Category: catQueryParameters,
+			}, recipe.GetCriteriaAcceleratorTypes),
+			withCompletions(&cli.StringFlag{
+				Name:     flagIntent,
+				Usage:    fmt.Sprintf("Filter by workload intent (e.g. %s)", strings.Join(recipe.GetCriteriaIntentTypes(), ", ")),
+				Category: catQueryParameters,
+			}, recipe.GetCriteriaIntentTypes),
+			withCompletions(&cli.StringFlag{
+				Name:     flagOS,
+				Usage:    fmt.Sprintf("Filter by OS type (e.g. %s)", strings.Join(recipe.GetCriteriaOSTypes(), ", ")),
+				Category: catQueryParameters,
+			}, recipe.GetCriteriaOSTypes),
+			withCompletions(&cli.StringFlag{
+				Name:     flagPlatform,
+				Usage:    fmt.Sprintf("Filter by platform type (e.g. %s)", strings.Join(recipe.GetCriteriaPlatformTypes(), ", ")),
+				Category: catQueryParameters,
+			}, recipe.GetCriteriaPlatformTypes),
+			dataFlag(),
+			withCompletions(&cli.StringFlag{
+				Name:     flagFormat,
+				Aliases:  []string{"t"},
+				Value:    string(serializer.FormatTable),
+				Usage:    "Output format (json, yaml, table)",
+				Category: catOutput,
+			}, func() []string { return []string{"json", "yaml", "table"} }),
+		},
+		Action: func(ctx context.Context, cmd *cli.Command) error {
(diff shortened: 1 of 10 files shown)


pkg/client/v1/types.go
```diff
@@ -351,6 +351,30 @@ type ComponentBundle struct {
...
+type CatalogEntry struct {
+	// Name is the overlay name, e.g. "h100-eks-ubuntu-training".
+	Name string `json:"name" yaml:"name"`
+
+	// Criteria is the set of dimensions this overlay targets.
+	Criteria Criteria `json:"criteria" yaml:"criteria"`
+
+	// Curated is true when this overlay is a catalog leaf (no other
+	// overlay inherits from it).
+	Curated bool `json:"curated" yaml:"curated"`
```
mchmarny: `Curated` is a misnomer here. Everything merged to `main` is curated by definition — that's what the review process is for. What this field actually encodes is a structural property of the overlay inheritance tree: whether this overlay is a leaf (no other overlay lists it as `spec.base`). Calling an intermediate overlay like `eks-training` "not curated" implies it was carelessly added, which misrepresents the overlay system.

Two options:

1. **Rename to `IsLeaf bool`** (or `IsTerminal`) — names the structural property accurately without a quality connotation. Callers who only want leaf overlays can filter on it.

2. **Drop the field entirely** — the leaf/non-leaf distinction is an internal implementation detail of how the overlay chain composes. Expose it only if there's a demonstrated caller need, and even then name it structurally.

Separately: if the primary use case is "show me what I can generate for h100+eks+ubuntu+training", that's a resolution question — the answer already comes from `aicr recipe`. The catalog list is useful for enumeration, but intermediate overlays may just add noise for most callers. Consider whether `list` should default to showing only leaf overlays (with a `--all` flag to include intermediates), rather than relying on a `Curated` boolean that the caller has to filter on.

mchmarny (changes requested):
The `Curated` field in `CatalogEntry` needs to be addressed before merge — see inline comment for details. Everything in `main` is "curated" by definition; the field conflates editorial quality with internal overlay tree structure. Either rename it to something structurally accurate (`IsLeaf`) or drop it entirely.

@rsd-darshan do you mind also creating an issue for this so we can align on the scope before the PR lands?

rsd-darshan:
@mchmarny good catch — renamed `Curated` to `IsLeaf` throughout (commit 7f19b3ac). The JSON/YAML tag is now `is_leaf`. Agreed that the structural name is clearer and avoids the quality connotation.

On your suggestion about defaulting to leaf-only with `--all`: happy to do that in a follow-up if the maintainers prefer it, but wanted to keep this PR scoped to the minimum needed to address #1207 and not grow it further while it's under review.

pkg/client/v1/aicr.go
```diff
@@ -316,6 +316,64 @@ func (c *Client) LoadCatalog(ctx context.Context) error {
...
+
+	store, err := recipe.LoadMetadataStoreFor(ctx, dp)
+	if err != nil {
+		return nil, errors.PropagateOrWrap(err, errors.ErrCodeInternal, "failed to load recipe catalog")
+	}
+
+	raw := store.ListCatalog(toInternalCriteria(filter))
+	entries := make([]CatalogEntry, len(raw))
+	for i, e := range raw {
+		var c Criteria
```
mchmarny: `var c Criteria` shadows the method receiver `c *Client`. The compiler accepts it because the outer `c` isn't used after the lock section, but `govet -shadow` (wired into golangci-lint) may flag it. Use `crit` or `criteria` instead:

```go
var crit Criteria
if wrapped := WrapCriteria(e.Criteria); wrapped != nil {
    crit = *wrapped
}
entries[i] = CatalogEntry{
    Name:     e.Name,
    Criteria: crit,
    IsLeaf:   e.IsLeaf,
    Source:   e.Source,
}
```

pkg/cli/recipe_list.go
```diff
@@ -0,0 +1,258 @@
...
+func writeCatalogEntries(ctx context.Context, cmd *cli.Command, entries []aicr.CatalogEntry, format serializer.Format) error {
+	w := cmd.Root().Writer
+
+	switch format {
+	case serializer.FormatJSON:
+		data, err := json.MarshalIndent(entries, "", "  ")
+		if err != nil {
+			return errors.Wrap(errors.ErrCodeInternal, "failed to marshal catalog entries as JSON", err)
+		}
+		if _, err := fmt.Fprintln(w, string(data)); err != nil {
```
mchmarny, line 202: nit: project convention for machine-readable YAML output is `serializer.MarshalYAMLDeterministic` (see `query.go`). `CatalogEntry` has no map fields so `yaml.Marshal` is deterministic in practice, but keeping it consistent avoids a future footgun if `Criteria` ever gains one.

mchmarny (commented):
Good response on `IsLeaf` — the rename is clean and consistent across all layers. Two small items inline.

One thing to clean up before merge: the PR description's Implementation Notes still say `Curated: true` / `Curated: false` — update that to match the renamed field.

Also: `CatalogEntry.Source` uses string values `"embedded"` / `"external"` that are only documented in comments, while the actual constants live unexported in `pkg/recipe`. Consider exporting them (e.g. `CatalogSourceEmbedded`, `CatalogSourceExternal`) so callers can compare without hardcoding magic strings.

PR mechanics before merge: squash the 5 commits to 1, add the `enhancement` label, and confirm CI runs the full build/lint/test gate (only label/auto-merge checks are visible on the head commit right now).

ayuskauskas:
This seems reasonable. The `--all` work can be deferred as long as the documentation is comprehensive on examples for getting everything vs just the leafs.

rsd-darshan:
Addressed all inline items (commit 33950bd7):

- `var c Criteria` renamed to `var crit Criteria` to avoid shadowing the `*Client` receiver
- YAML output now uses `serializer.MarshalYAMLDeterministic` (consistent with `query.go`)
- Exported `CatalogSourceEmbedded` / `CatalogSourceExternal` from `pkg/recipe` and re-exported from `pkg/client/v1` so callers can compare `Source` without hardcoding strings

Also squashed the 5 commits to 1 as requested.

Could a maintainer add the `enhancement` label? No permission to add labels from a fork.

mchmarny (dismissed):

mchmarny:
@rsd-darshan few issues from CR, otherwise it generally LGTM;

rsd-darshan:
@mchmarny all three CR items were addressed in commit 3f46c002 (pushed just before your comment):

- **Defensive Criteria copy** — `critCopy := *overlay.Spec.Criteria` before appending to entries in `ListCatalog`
- **Table-driven tests** — 10 single-case filter/matchesCatalogFilter tests consolidated into two `t.Run` tables
- **MD031 blank lines** — blank lines added before/after all fenced code blocks in `cli-reference.md`

mchmarny (dismissed):
All review findings addressed. Good work on the quick turnaround — `IsLeaf` rename, defensive `Criteria` copy, table-driven tests, `MarshalYAMLDeterministic`, exported source constants, and the flag name constants cleanup are all clean. Also good to see the PR squashed to a single commit. LGTM.

mchmarny (approved):

PR #1216, opened 2026-06-08
feat(recipe): signed catalog provenance for V1 closed supply chain

### LINKED ISSUE #1155: feat: recipe-data provenance (signed catalog/registry)
## Problem

The V1 "Closed Supply Chain" bar requires **recipe data** (the catalog/registry that drives recipe resolution) to have retrievable provenance, alongside binaries, images, and bundles. Today the recipe data served by the provider (`pkg/recipe/provider.go`, `criteria_registry.go`, `metadata_store.go`) is not signed and has no verification path: a consumer cannot confirm the recipe inputs they resolved came from a trusted source unmodified.

## Proposal

Sign the recipe catalog/registry and give consumers a way to verify it, so recipe inputs carry provenance comparable to bundles.

## Scope (to be refined)

- Decide the unit of signing (whole registry snapshot vs. per-entry) and the format (detached Sigstore bundle / attestation over the catalog digest).
- Produce the signature as part of the release/publish flow.
- Add a verification path (CLI and/or library) that checks the recipe data against its signature before use.
- Document the trust model.

## Dependencies

- Independent of the bundle signing/verifying sub-issues, but shares the Sigstore primitives in `pkg/bundler/attestation` and the verifier in `pkg/bundler/verifier`.

## Success criteria

- The published recipe catalog/registry has a retrievable signature.
- A consumer can verify recipe data provenance before resolving a recipe.
- `make test` / `make lint` pass.

mchmarny:
## Summary

Adds retrievable provenance and a consumer verification path for the recipe catalog (`registry.yaml` + `validators/catalog.yaml`), filling the recipe-data signing workstream of the V1 closed-supply-chain bar.

## Motivation / Context

The V1 \"Closed Supply Chain\" acceptance criterion requires every artifact — binary, image, recipe data, bundle — to have retrievable provenance and a documented verification path. Today the recipe catalog is shipped unsigned with no verification path; this PR closes that gap.

Fixes: #1155
Related: #1149 (parent epic)

## Type of Change

## Component(s) Affected

- [x] CLI (`pkg/cli`)
- [x] Recipe engine / data (`pkg/recipe`, new `pkg/recipe/catalog`)
- [x] Bundlers (`pkg/bundler/attestation` — additive `BuildType` field)
- [x] Build/CI/tooling (`.goreleaser.yaml`)

## Implementation Notes

Implementation reuses `pkg/bundler/attestation` (Sigstore keyless OIDC + DSSE + in-toto + SLSA Provenance v1) and `pkg/bundler/verifier` (Fulcio identity pinning), so catalog attestations are structurally identical to bundle attestations — same bundle file format, same predicate type, different `buildType` URI (`https://aicr.nvidia.com/recipe-catalog/v1`).

- **`pkg/recipe/catalog`** (new): `ComputeDigest` (deterministic SHA-256 over `serializer.MarshalYAMLDeterministic` of both files), `Sign` (delegates to `attestation.Attester`), `Verify` (delegates to `verifier.VerifyBinaryAttestation` with NVIDIA CI identity pinning).
- **`pkg/bundler/attestation`**: `StatementMetadata.BuildType` is optional — empty falls back to `BundleBuildType`, so existing bundle attestation callers are unchanged. New `CatalogBuildType` constant.
- **`pkg/recipe`**: `RegistryFileName`/`CatalogFileName` exported; unexported aliases preserved for internal callers (zero churn elsewhere).
- **CLI**: hidden `aicr recipe sign-catalog --output <path>` (CI-only, reuses the same OIDC precedence as `aicr bundle --attest`); user-facing `aicr recipe verify-catalog <bundle-path>` with optional `--identity-pattern` for testing.
- **Release**: goreleaser after-hook signs the catalog once per release (gated on `linux/amd64` + `SLSA_PREDICATE` env var, same guard as the binary attestation hook so local goreleaser runs skip the OIDC flow). The bundle is published as a `recipe-catalog.sigstore.json` release asset and included in the aicr tarball.
- **Optional runtime verification hook** in `LayeredDataProvider` was intentionally **not** implemented; the CLI subcommand satisfies the V1 verification-path requirement and avoids a `recipe → catalog → recipe` import cycle.

## Testing

```
make qualify
```

Results:
- All unit tests pass with race detector (`pkg/recipe/catalog` 69.5% statement coverage)
- Total coverage 76.5% (threshold 75%)
- `go vet`: pass
- `golangci-lint`: 0 issues

New tests cover determinism, missing-file error paths, NoOp signing, digest equivalence between `ComputeDigest` and `Sign`, bundle-not-found and invalid-bundle verification paths, and CLI subcommand registration (visible/hidden).

## Risk Assessment

- [x] **Low** — Isolated, additive change; existing code paths unchanged. `BuildType` field defaults to empty → `BundleBuildType` (no regression). The goreleaser after-hook is gated on `SLSA_PREDICATE` env presence so local/PR builds are unaffected.

**Rollout notes:** The first release after this PR merges will publish the signature; older releases remain unsigned. Verification is opt-in via the new CLI subcommand.

## Checklist

- [x] I added tests for new functionality
- [x] N/A — no user-facing behavior shipped to existing commands; new commands documented inline (`--help`)
- [x] Changes follow existing patterns in the codebase (pkg/errors, slog, deterministic YAML, Sigstore primitive reuse)


pkg/recipe/catalog/verify_test.go
```diff
@@ -0,0 +1,52 @@
...
+	bundlePath := filepath.Join(t.TempDir(), "garbage.sigstore.json")
+	if err := os.WriteFile(bundlePath, []byte(`{"not":"a real bundle"}`), 0o600); err != nil {
+		t.Fatalf("write garbage bundle: %v", err)
+	}
+
+	_, err := catalog.Verify(ctx, bundlePath, newFullProvider(), catalog.VerifyOptions{})
+	if err == nil {
+		t.Fatal("expected error for invalid bundle, got nil")
+	}
+}
```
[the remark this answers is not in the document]
mchmarny: Skipping. A real success-path test requires either a pre-signed Sigstore bundle fixture (which would need to be regenerated whenever the trust root rotates) or refactoring `verifier.VerifyBinaryAttestation` into an interface so it can be mocked. Neither carries its weight here: the digest computation and bundle-stat error paths are unit-tested, and the actual sigstore-go verification is exercised end-to-end by the existing `pkg/bundler/verifier` integration tests. Same reasoning was accepted on #1205 for the parallel KMS verification path.

pkg/recipe/catalog/verify.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeInternal, "invalid catalog digest", err)
+	}
+
+	identityPattern := opts.CertificateIdentityRegexp
+	if identityPattern == "" {
+		identityPattern = verifier.TrustedRepositoryPattern
+	}
+
+	identity, err := verifier.VerifyBinaryAttestation(ctx, bundlePath, identityPattern, digestBytes)
```
njhensley, line 78: 🔴 **Blocking — NVIDIA/aicr identity pinning is bypassable.**

`opts.CertificateIdentityRegexp` (CLI `--identity-pattern`, env `AICR_CATALOG_IDENTITY_PATTERN`) is passed straight to `verifier.VerifyBinaryAttestation` **without calling `verifier.ValidateIdentityPattern`**. The bundle path calls it (`verifier.go:120`), and that validation is the *only* thing enforcing the pattern contains `://github.com/NVIDIA/aicr/`. `VerifyBinaryAttestation` pins only the OIDC *issuer*.

The issuer pin to `token.actions.githubusercontent.com` still holds, so the blast radius is "any GitHub Actions OIDC identity" — but `aicr recipe verify-catalog bundle.json --identity-pattern '.*'` (or the same via env) then reports **"catalog verified"** for a catalog signed by *any* workflow in *any* repo, including an attacker-controlled fork. The help text "Must contain 'NVIDIA/aicr'." is unenforced.

**Fix:** call `verifier.ValidateIdentityPattern(identityPattern)` immediately after the empty-string fallback (before `VerifyBinaryAttestation`), returning `ErrCodeInvalidRequest` on failure. Add a test asserting `--identity-pattern '.*'` is rejected.

pkg/cli/recipe.go
```diff
@@ -133,6 +130,11 @@ Override config file values with flags:
 
 Override snapshot-detected criteria:
   aicr recipe --snapshot cm://gpu-operator/aicr-snapshot --service gke`,
+		Commands: []*cli.Command{
+			recipeListCmd(),
+			recipeSignCatalogCmd(),
+			recipeVerifyCatalogCmd(),
```
njhensley, line 136: 💡 **Suggestion (non-blocking) — consider a top-level `aicr catalog` namespace.**

These register `sign-catalog` / `verify-catalog` as hyphenated verbs under `recipe`. It might read more consistently to follow the shape used for the evidence workstream — `evidenceCmd()` is its own top-level verb-group exposing `aicr evidence digest|publish|verify` — i.e. **`aicr catalog sign` / `aicr catalog verify`** under a new `catalogCmd()` parent.

Worth weighing now mainly because the CLI surface is user-facing: once `aicr recipe verify-catalog` ships in a release, renaming it later is a breaking change. If you'd rather keep it under `recipe` (which already has `recipe list`), that's a reasonable call too — just flagging the precedent.

pkg/recipe/catalog/digest.go
```diff
@@ -0,0 +1,88 @@
...
+const digestAlgoSHA256 = "sha256"
+
+// ComputeDigest reads registry.yaml and validators/catalog.yaml through
+// provider, deterministically re-serializes each, and returns an
+// attestation.AttestSubject whose Digest is the SHA-256 of the concatenated
+// canonical bytes. ResolvedDependencies holds the individual file digests.
+func ComputeDigest(ctx context.Context, provider recipe.DataProvider) (attestation.AttestSubject, error) {
+	regBytes, err := deterministicYAML(ctx, provider, recipe.RegistryFileName)
+	if err != nil {
+		return attestation.AttestSubject{}, errors.PropagateOrWrap(err, errors.ErrCodeNotFound,
```
njhensley, line 48: 🟠 **Error code clobbered.** This (and the catalog read at :46) unconditionally re-wraps `deterministicYAML`'s result with `errors.Wrap(ErrCodeNotFound, …)`. But that helper can surface a context-cancel (`ErrCodeTimeout` from the provider) or a YAML-parse failure (`ErrCodeInternal`). `errors.Wrap` sets `Code` unconditionally and `*StructuredError.Is` matches the outermost code, so a timeout/parse error is reported to callers as `ErrCodeNotFound` — the "don't double-wrap errors that already have proper codes" anti-pattern in CLAUDE.md.

**Fix:** use `errors.PropagateOrWrap(err, ErrCodeNotFound, "failed to read registry file")` (as `LayeredDataProvider.ReadFile` already does), so genuine not-found maps to `ErrCodeNotFound` while timeout/internal keep their codes.

pkg/recipe/catalog/digest.go
```diff
@@ -0,0 +1,88 @@
...
+
+// deterministicYAML reads a YAML file through provider and re-marshals it
+// with sorted keys so the output is byte-stable across runs.
+func deterministicYAML(ctx context.Context, provider recipe.DataProvider, path string) ([]byte, error) {
+	raw, err := provider.ReadFile(ctx, path)
+	if err != nil {
+		return nil, err
+	}
+	var v any
+	if err := yaml.Unmarshal(raw, &v); err != nil {
```
njhensley: 🟡 **Digest covers re-serialized YAML, not the shipped bytes.** This `yaml.Unmarshal(raw, &v any)` → `MarshalYAMLDeterministic` round-trip means the signed digest attests a *normalized re-rendering*, not the bytes embedded in the binary. The lossy step is the decode-into-`any` (not `MarshalYAMLDeterministic`, which preserves comments via its `yaml.Node` round-trip):

- **Comments are dropped** and scalars canonicalized — an attestation a consumer reads as "covers the shipped catalog" doesn't cover comments.
- **Multi-document YAML is silently truncated** — `Unmarshal` into a single `any` decodes only the first doc. Neither file is multi-doc today, but there's no guard.

**Fix:** hash the raw `provider.ReadFile` bytes directly (embedded-FS bytes are already deterministic — the map-ordering rationale for `MarshalYAMLDeterministic` doesn't apply to bytes you don't decode). If re-serialization must stay, reject multi-doc input and document that the digest covers normalized content.

pkg/recipe/catalog/digest.go
```diff
@@ -0,0 +1,88 @@
...
+		return attestation.AttestSubject{}, errors.PropagateOrWrap(err, errors.ErrCodeNotFound,
+			"failed to read catalog file")
+	}
+
+	regHash := sha256.Sum256(regBytes)
+	catHash := sha256.Sum256(catBytes)
+
+	combined := sha256.New()
+	combined.Write(regBytes)
+	combined.Write(catBytes)
```
njhensley: 🟢 **Low (defense-in-depth) — no length delimiter between the two files.** `Write(regBytes); Write(catBytes)` is boundary-ambiguous: `(reg=A, cat=BC)` and `(reg=AB, cat=C)` hash identically. Low risk today (fixed file count, both NVIDIA-controlled), but hashing the two per-file hex digests instead of the concatenated bodies removes the footgun and self-documents the binding. Note the per-file `ResolvedDependencies` digests are informational — only this combined digest is verified.

pkg/recipe/catalog/sign.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+
+	return &SignResult{
+		Digest:     subject.Digest[digestAlgoSHA256],
+		BundleJSON: bundleJSON,
+	}, nil
+}
+
+func writeBundle(path string, data []byte) error {
```
njhensley, line 87: 🟠 **Test gap — the real-bundle write path is 0% covered.** Both `Sign` tests use `NewNoOpAttester()`, whose `Attest` returns `(nil, nil)`, so the `bundleJSON != nil` branch (line 71) and all of `writeBundle` (success + `os.Create` error + close error) never execute.

Add a stub attester returning non-nil bytes: assert the file is written with exactly those bytes, and a separate case with an unwritable `Output` (e.g. a path under a non-existent dir) asserting `ErrCodeInternal`. Also: the error-path tests (`TestSign_MissingRegistry`, `TestComputeDigest_Missing*`) only assert `err != nil` — switch to `stderrors.Is(err, errors.New(ErrCodeNotFound, ""))` so a code regression is caught (this also locks in the fix for the error-code finding).

pkg/cli/recipe_sign_catalog.go
```diff
@@ -0,0 +1,111 @@
...
+	result, err := recipecat.Sign(ctx, provider, recipecat.SignOptions{
+		Attester:    attester,
+		Output:      output,
+		ToolVersion: version,
+	})
+	if err != nil {
+		return err
+	}
+
+	if result.BundleJSON == nil {
```
njhensley, line 103: 🟠 **Test gap — neither CLI Action is exercised.** `TestRecipeSignCatalogCmd_Registered` / `..._VerifyCatalogCmd_Registered` assert *registration only*. This nil-bundle branch (`ErrCodeInternal`) fires in every non-OIDC environment (i.e. every local/test run with NoOp), so it's both important and trivially testable — invoke `runRecipeSignCatalogCmd` without OIDC env and assert `ErrCodeInternal`. Likewise add a `verify-catalog` test for the `NArg()!=1` guard (`ErrCodeInvalidRequest`).

.goreleaser.yaml
```diff
@@ -45,6 +45,17 @@ builds:
...
+        ## Runs once per release (linux/amd64) and only when SLSA_PREDICATE is set
+        ## (same gate as the binary attestation hook) so local goreleaser runs skip
+        ## the OIDC flow. Output bundle is published as a release asset (see
+        ## release.extra_files below).
+        - cmd: >-
+            bash -c 'if [ "{{ .Os }}/{{ .Arch }}" != "linux/amd64" ]; then exit 0; fi;
+            if [ -z "${SLSA_PREDICATE:-}" ]; then exit 0; fi;
+            "{{ .Path }}" recipe sign-catalog
+            --output "$(dirname "{{ .Path }}")/recipe-catalog.sigstore.json"'
+          output: true
```
njhensley, line 62: 🟡 **Two issues on this hook.**

1. **Comment inaccuracy:** "same gate as the binary attestation hook" — it's actually a *superset*. The binary hook has no os/arch filter and runs once per build target; this one adds `linux/amd64` so it runs once. Worth correcting the comment.
2. **Release-abort surface:** `output: true` only streams the hook's output to the log — it is **not** non-fatal. A non-zero exit from `recipe sign-catalog` (OIDC unresolved / Fulcio unreachable) aborts the whole release *after* the binaries are built, with no retry. This mirrors the existing binary hook, so it isn't new, but it doubles the per-release Fulcio dependency surface. Worth confirming the release runner has a clean path to public-good Fulcio/Rekor.

.goreleaser.yaml
```diff
@@ -144,6 +159,8 @@ archives:
     files:
       - src: "dist/aicr_{{ .Os }}_{{ .Arch }}*/aicr-attestation.sigstore.json"
         strip_parent: true
+      - src: "dist/aicr_linux_amd64*/recipe-catalog.sigstore.json"
```
njhensley, line 173: 🟡 **Catalog bundle ships in all four archives, not just linux/amd64.** `archives.files` is evaluated per-archive, and this **fixed** (non-templated) `dist/aicr_linux_amd64*/...` path matches for all four archives — so the darwin/arm64 tarballs each bundle the linux/amd64 catalog bundle. Harmless (catalog is arch-independent) but contradicts the "for linux/amd64" comment above. Confirm intent and reconcile the comment.

Separately: the loose `release.extra_files` copy of this bundle is not covered by `checksums.txt` (and not SBOM'd). Authority comes from the Sigstore signature, not the checksums manifest — worth a one-line note, or rely on the in-archive copy (already covered by the tarball checksum) and drop the loose asset.

njhensley (changes requested):
## Multi-persona review — signed catalog provenance

**Verdict: Request changes.** One blocking security defect; otherwise a clean, well-structured PR that closes a real V1 supply-chain gap by reusing the existing `attestation`/`verifier` primitives. The reuse keeping catalog attestations structurally identical to bundle attestations is the right call.

Detailed, line-anchored findings are in the **inline comments** below. Index:

| # | Sev | Where | Finding |
|---|-----|-------|---------|
| 1 | 🔴 Blocking | `catalog/verify.go:65-70` | Identity pinning is bypassable — `--identity-pattern` skips `ValidateIdentityPattern` |
| 2 | 💡 Suggestion | `cli/recipe.go:135` | Consider a top-level `aicr catalog` namespace (mirrors `evidence`) instead of `recipe`-nested verbs |
| 3 | 🟠 High | `catalog/digest.go:40` | `ComputeDigest` clobbers inner error code (timeout/internal → NotFound) |
| 4 | 🟠 High | *docs* | Missing `docs/user/cli-reference.md` entry for the new user-facing `verify-catalog` (CLAUDE.md requires it in-PR) |
| 5 | 🟠 High | `catalog/sign.go:83`, `cli/recipe_sign_catalog.go:103` | Real-bundle write path + both CLI Actions are 0% covered; error tests assert `err != nil` not codes |
| 6 | 🟡 Medium | `catalog/digest.go:84` | Digest hashes re-serialized YAML, not shipped bytes (drops comments, silently truncates multi-doc) |
| 7 | 🟡 Medium | `.goreleaser.yaml:48,162` | Catalog bundle ships in all 4 archives; "same gate" comment inaccurate; loose asset not in checksums.txt |
| 8 | 🟡 Medium | `.goreleaser.yaml:48` | Catalog-signing failure aborts the whole release (Fulcio dependency, no retry) |

**Low / nits** (not separately inlined): combined digest has no length delimiter (`digest.go:54-55`, also inlined); duplicated `"recipe-catalog"` literal (`digest.go`/`sign.go:59`) — extract a const; `Sign(ctx, provider, opts)` vs `Verify(ctx, bundlePath, provider, opts)` arg-order asymmetry; `ComputeDigest` exported with no external caller; redundant `os.Stat` in `Verify` (benign); `SLSA_PREDICATE` reused as a presence sentinel; `verify-catalog` never asserts the predicate `buildType` (defense-in-depth — replay already blocked by digest binding).

**Confirmed compliant (no action):** `writeBundle` Close()-error handling, the CI-only `//nolint:gosec`, slog-vs-`cmd.Root().Writer` split, the `RegistryFileName`/`CatalogFileName` export + unexported-alias trick, and `{{ .Path }}` / darwin-arm clean-skip in the goreleaser hook. Context-timeout omission on the package entry points is acceptable (the provider honors `ctx.Err()`; the sigstore paths bound their own I/O).

**Bottom line:** Fix #1 before merge — it nullifies the PR's central guarantee. #3–#5 should land in the same PR. #2 is a suggestion worth weighing before release (CLI surface is compatibility-sensitive), but it's the author's call.

_Reviewed across four personas (supply-chain security, Go conventions, release engineering, API/test design) and meta-reviewed for accuracy._

pkg/recipe/catalog/verify.go
mchmarny: Fixed in `6326b4ce`. `Verify` now calls `verifier.ValidateIdentityPattern` after the empty-string fallback and returns `ErrCodeInvalidRequest` if the pattern is missing the `NVIDIA/aicr` repo prefix. Added `TestVerify_RejectsOverlyBroadIdentityPattern` (asserts `--identity-pattern '.*'` is rejected with `ErrCodeInvalidRequest`) to lock the contract.

pkg/recipe/catalog/verify.go
```diff
@@ -0,0 +1,79 @@
...
+	digestHex := subject.Digest[digestAlgoSHA256]
+	digestBytes, err := hex.DecodeString(digestHex)
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeInternal, "invalid catalog digest", err)
+	}
+
+	identityPattern := opts.CertificateIdentityRegexp
+	if identityPattern == "" {
+		identityPattern = verifier.TrustedRepositoryPattern
+	}
```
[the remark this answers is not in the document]
mchmarny: Fixed in `6326b4ce` — `Verify` now calls `verifier.ValidateIdentityPattern` and returns `ErrCodeInvalidRequest` on violation. Same root cause as njhensley's blocking finding above; one fix covers both.

pkg/recipe/catalog/digest.go
mchmarny: Fixed in `7ddad887` (and reaffirmed in `6326b4ce`). Both call sites now use `errors.PropagateOrWrap(err, ErrCodeNotFound, ...)`, so timeout/internal codes from the provider survive the boundary. Tests now assert the structured code (`stderrors.Is(err, errors.New(ErrCodeNotFound, ""))`) rather than just `err != nil`, so a future regression is caught.

pkg/recipe/catalog/digest.go
mchmarny: Fixed in `6326b4ce`. `ComputeDigest` now hashes raw `provider.ReadFile` bytes directly — dropped the `yaml.Unmarshal → MarshalYAMLDeterministic` round-trip entirely. The digest now covers the exact shipped bytes (comments preserved, multi-document YAML tolerated). Per-file SHA-256s in `ResolvedDependencies` are now over the same raw bytes, so what gets recorded matches what gets verified.

pkg/recipe/catalog/digest.go
mchmarny: Fixed in `6326b4ce`. The combined hasher now writes `u64-BE(len(reg)) || reg || u64-BE(len(cat)) || cat`, so `(regA, catBC)` and `(regAB, catC)` produce different digests. Removes the footgun for free even though the file count is fixed today.

pkg/recipe/catalog/digest.go
```diff
@@ -0,0 +1,88 @@
...
+	combinedHex := hex.EncodeToString(combined.Sum(nil))
```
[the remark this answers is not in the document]
mchmarny: Fixed in `6326b4ce`. Same fix as njhensley's parallel concern — length-prefixed encoding now used. See reply on `digest.go:55` thread for details.

pkg/recipe/catalog/sign.go
mchmarny: Real-bundle write path is now covered by `TestSign_WritesBundleToFile` (added in `61c57d48`) — it injects a `fakeAttester` returning canned bytes and asserts file existence + byte-equality. The error-path tests (`TestSign_MissingRegistry`, `TestComputeDigest_MissingRegistry`, `TestComputeDigest_MissingCatalog`) now assert `stderrors.Is(err, errors.New(ErrCodeNotFound, ""))` in `6326b4ce` so a code regression is caught. New `TestSign_RejectsNilAttester` locks in the `ErrCodeInvalidRequest` from the nil-Attester guard. Catalog package coverage is now **85.2%**.

pkg/recipe/catalog/sign.go
```diff
@@ -0,0 +1,97 @@
...
+		RecipeSource:  provider.Source(recipe.RegistryFileName),
+		BuildType:     attestation.CatalogBuildType,
+		ToolVersion:   opts.ToolVersion,
+		Deterministic: true,
+	}
+
+	bundleJSON, err := opts.Attester.Attest(ctx, subject)
+	if err != nil {
+		return nil, errors.PropagateOrWrap(err, errors.ErrCodeInternal, "catalog attestation signing failed")
+	}
```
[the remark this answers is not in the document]
mchmarny: Fixed in `6326b4ce`. `Sign` now returns `errors.New(ErrCodeInvalidRequest, "attester is required")` when `opts.Attester == nil`, before any call to `Attest`. `TestSign_RejectsNilAttester` locks the contract.

pkg/cli/recipe_sign_catalog.go
mchmarny: Partially addressed in `6326b4ce`: added `TestRecipeVerifyCatalog_RejectsMissingPositional` which drives `runRecipeVerifyCatalogCmd` through `urfave/cli` and asserts the usage error when `<bundle-path>` is omitted.

Skipping the `sign-catalog` nil-bundle path test deliberately. That branch (`result.BundleJSON == nil` → `ErrCodeInternal`) only fires when `attestation.ResolveAttesterLazy` returns a `NoOpAttester`, which it never does when `opts.Attest = true` — the sign-catalog command hardcodes `Attest: true`, so reaching that branch via the CLI would require either (a) a regression in `ResolveAttesterLazy` (out of scope) or (b) a mock attester. The branch is defensive-only "this should never happen" code; instrumenting the CLI to exercise it would mean exposing an attester-injection seam purely for the test. The CI-only hidden command remains exercised end-to-end by the goreleaser hook on every release.

pkg/cli/recipe_verify_catalog.go
```diff
@@ -0,0 +1,77 @@
...
+Exit code is 0 on success, non-zero on any verification failure.`,
+		Flags: []cli.Flag{
+			&cli.StringFlag{
+				Name:    "identity-pattern",
+				Usage:   "Override the NVIDIA CI certificate identity regexp. Must contain 'NVIDIA/aicr'.",
+				Sources: cli.EnvVars("AICR_CATALOG_IDENTITY_PATTERN"),
+			},
+		},
+		Action: runRecipeVerifyCatalogCmd,
+	}
```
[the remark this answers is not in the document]
mchmarny: Skipping. `aicr verify` and `aicr evidence verify` (the parallel verification commands in this repo) both produce fixed human-readable text without `--format`. The verify-catalog output is two lines — digest and identity — and a programmatic consumer can scrape `sha256:<64-hex>` directly. Adding `--format json|yaml|table` here would diverge from the existing verify-command convention without solving a concrete consumer need. If a downstream automation needs structured output, we can add `--format` to all three verify commands together so the surface stays consistent.

pkg/cli/recipe.go
mchmarny: Keeping under `recipe` for now. The reasoning cuts both ways: `evidence` is its own top-level verb-group, but `recipe list` (which landed in #1208 right before this PR) is also under `recipe`, and the catalog *is* recipe data — putting catalog operations adjacent to `recipe list` reads naturally. The cost of being wrong here is one release of compat-shim aliasing (`aicr catalog verify` → `aicr recipe verify-catalog`) once we see how operators actually use it; the cost of an early namespace split is harder to walk back. Filed [#1217](https://github.com/NVIDIA/aicr/issues/1217) to revisit before V1 GA if a third catalog verb appears.

.goreleaser.yaml
mchmarny: Fixed in `6326b4ce`. The inaccurate "same gate" comment is replaced with an explicit "two-axis filter" explanation: `SLSA_PREDICATE` (release-only) AND `linux/amd64` (single-combination pin so the catalog is signed exactly once).

On the release-abort surface: confirmed acceptable. The release workflow's Fulcio dependency surface is **identical** to the existing binary-attestation hook — same Sigstore public-good endpoints, same OIDC token, same network path — so this PR doesn't add a new failure mode, it just doubles the per-release call count. If Fulcio outage becomes a recurring release pain point, the fix is workflow-level (retry-with-backoff wrapping the goreleaser invocation), not hook-level. Tracked separately under the supply-chain hardening epic (#1149).

.goreleaser.yaml
mchmarny: Reconciled in `6326b4ce`. Made the intent explicit in two places:

1. The build-hook comment now describes both publication paths (loose release asset + inside every archive) and explains why both: the loose asset is for one-shot `curl + verify-catalog` UX; the per-archive copy is because the signed artifact is arch-independent and a darwin user shouldn't need a second download.
2. The `archives.files` entry has a comment noting "signed once, arch-independent, ship everywhere".

On the `checksums.txt` coverage gap for the loose asset: kept the loose copy deliberately and added a comment to `release.extra_files` explaining that the authority comes from the Sigstore signature (Fulcio cert + Rekor inclusion proof), not from the checksums manifest. The in-archive copy is still covered by the tarball checksum, which is sufficient defense-in-depth.

mchmarny:
@njhensley thanks for the multi-persona pass — that's exactly the level of scrutiny this corner of the supply chain needs. Disposition for all 8 findings from your review:

| # | Sev | Finding | Disposition |
|---|---|---|---|
| 1 | 🔴 Blocking | Identity pinning bypassable | **Fixed** in `6326b4ce`. `Verify` now calls `verifier.ValidateIdentityPattern` after the empty-string fallback; returns `ErrCodeInvalidRequest` on violation. New `TestVerify_RejectsOverlyBroadIdentityPattern` locks the `--identity-pattern '.*'` rejection contract. |
| 2 | 💡 Suggestion | Top-level `aicr catalog` namespace | **Kept under `recipe`** — full reasoning on the inline thread. Filed #1217 to revisit before V1 GA. |
| 3 | 🟠 High | `ComputeDigest` error code clobber | **Fixed** in `7ddad887` + `6326b4ce`. `PropagateOrWrap` at both call sites; tests now assert structured codes via `stderrors.Is`. |
| 4 | 🟠 High | Missing `cli-reference.md` entry | **Fixed** in `6326b4ce`. Added `### aicr recipe verify-catalog` section between `recipe list` and `query` with synopsis, flags, the `curl + verify` example, and the expected success output. |
| 5 | 🟠 High | Test gaps (real-bundle write + CLI Actions) | **Mostly fixed**: bundle-write covered by `TestSign_WritesBundleToFile` (`61c57d48`); error-path tests now assert error codes (`6326b4ce`); `TestSign_RejectsNilAttester` + `TestRecipeVerifyCatalog_RejectsMissingPositional` added. Skipping the sign-catalog defensive-only `BundleJSON == nil` branch deliberately — full rationale on the inline thread. Coverage 69.5% → **85.2%**. |
| 6 | 🟡 Medium | Digest covers re-serialized YAML | **Fixed** in `6326b4ce`. Dropped the `yaml.Unmarshal → MarshalYAMLDeterministic` round-trip entirely — digest now covers raw `provider.ReadFile` bytes. Comments preserved; multi-doc YAML no longer silently truncated. |
| 7 | 🟡 Medium | Goreleaser comment + archive globbing | **Fixed** in `6326b4ce`. Comment now describes the actual two-axis filter (`SLSA_PREDICATE` ∧ `linux/amd64`); the per-archive `recipe-catalog.sigstore.json` placement got a dedicated note explaining "signed once, arch-independent, ship everywhere"; loose-asset `extra_files` placement is now documented as deliberate (authority comes from the Sigstore signature, not the checksums manifest). |
| 8 | 🟡 Medium | Catalog-signing failure aborts release | **Confirmed acceptable**; no behavior change. Same Fulcio dependency surface as the existing binary-attestation hook — failure modes are identical. If Fulcio outage becomes a recurring release pain point, the right fix is workflow-level (retry-with-backoff around the goreleaser invocation), not hook-level. Tracked under epic #1149. |

Plus the **low / nit** items from your tail paragraph:
- **Length delimiter on combined digest** — fixed alongside #6 (u64-BE length prefix per file).
- **Duplicate `"recipe-catalog"` literal** — extracted to `catalogSubjectName` const in `digest.go`, referenced by both `ComputeDigest` and `Sign`.
- **`Sign(ctx, provider, opts)` vs `Verify(ctx, bundlePath, provider, opts)` arg-order asymmetry**, **`ComputeDigest` exported with no external caller**, **redundant `os.Stat` in `Verify`**, **`SLSA_PREDICATE` reused as presence sentinel**, **`verify-catalog` doesn't assert predicate `buildType`** — acknowledged, not changed. The asymmetry is unavoidable because `verify-catalog`'s bundle-path is positional (UX) while `Sign`'s output is a flag; `ComputeDigest` stays exported as a small public API for downstream callers that want to compute the digest without signing (which is also what makes the `Verify` happy-path testable from outside the package); the `os.Stat` gives `Verify` a clean `ErrCodeNotFound` before the verifier's more cryptic deeper error; `SLSA_PREDICATE`-as-sentinel is consistent with the existing binary-attest hook; the `buildType` assertion is genuinely defense-in-depth — once a malicious predicate is digest-bound to a tampered catalog, the digest mismatch fires first.

All 13 inline threads resolved. `make qualify` passes locally; CI re-run on the amended commit.

cc @coderabbitai — the `verify-catalog` output-format thread was pushed back on (skipping `--format json|yaml|table` to stay consistent with the other two verification commands in this repo); rationale on that inline thread.

lockwobr:
## Review reconciliation — remaining open items

Reconciled across the three reviews on this PR (CodeRabbit, @njhensley's multi-persona review, and a separate pass of my own) against the current head `6326b4ce`. The author's follow-up commits (`61c57d4`, `6326b4ce`) already resolved every blocking and high-severity finding, so this just tracks what is left.

### ✅ Resolved since the original reviews
- **Identity-pin bypass (was 🔴 blocking):** `catalog.Verify` now calls `verifier.ValidateIdentityPattern` before honoring `--identity-pattern` / `AICR_CATALOG_IDENTITY_PATTERN`, so the `NVIDIA/aicr` repo pin can no longer be silently disabled.
- **Error-code masking (was 🟠):** `ComputeDigest` switched to `errors.PropagateOrWrap`, preserving inner `ErrCodeTimeout`/`ErrCodeInternal` codes.
- **Digest over re-serialized YAML (was 🟡):** now hashes the exact shipped bytes (no YAML normalization), so comments and multi-doc files are covered.
- **Digest ambiguity:** length-prefixed (`u64-BE(len) ‖ bytes`) injective encoding added.
- **Nil-attester panic guard, duplicated `"recipe-catalog"` literal → const, write-path test (`TestSign_WritesBundleToFile`), and the missing `docs/user/cli-reference.md` entry** are all in.

### Remaining (none blocking)

| # | Sev | Where | Item |
|---|-----|-------|------|
| 1 | 💡 Design | `pkg/cli/recipe.go` | CLI surface: `recipe sign-catalog` / `recipe verify-catalog` vs a top-level `aicr catalog {sign,verify}` namespace (mirrors `evidence`). Worth settling **before** the first release that ships the command, since the surface is compatibility-sensitive. Author's call. |
| 2 | 🟡 Release eng | `.goreleaser.yaml` | A Fulcio/OIDC hiccup during `sign-catalog` aborts the **entire** release (no `continue-on-error`/retry). This is consistent with the existing binary-attestation hook and fail-closed is defensible for a supply-chain release — please just confirm it is intended rather than incidental. |
| 3 | 🔵 Enhancement | `pkg/cli/recipe_verify_catalog.go` | `verify-catalog` emits human-readable text only. A `--format json\|yaml` would make the verdict consumable from CI/scripts. Optional. |
| 4 | 🔵 Test | `pkg/recipe/catalog/verify_test.go`, `pkg/cli/recipe_sign_catalog.go` | No success-path test for `catalog.Verify` (needs a committed signed-bundle fixture; the network-dependent sigstore path is reasonably left uncovered), and `runRecipeSignCatalogCmd` is still 0% (OIDC resolution is hard to unit-test). Acceptable, noting for completeness. |
| 5 | 🔵 Nit | `pkg/recipe/catalog/{sign,verify}.go` | Arg-order asymmetry: `Sign(ctx, provider, opts)` vs `Verify(ctx, bundlePath, provider, opts)`. Cosmetic. |

**Bottom line:** no blockers remain. Item 1 (namespace) and item 2 (confirm release-abort is intended) are the two worth a decision before merge; 3–5 are optional polish. Nice, clean reuse of the existing attestation/verifier primitives.

njhensley (approved):
Approving — thanks for the fast, thorough turnaround.

The blocking identity-pin bypass is fixed (`Verify` now calls `verifier.ValidateIdentityPattern` and returns `ErrCodeInvalidRequest`, with `TestVerify_RejectsOverlyBroadIdentityPattern` locking in that `--identity-pattern '.*'` is rejected). The error-code clobbering (`PropagateOrWrap`), the docs entry, and the test gaps (real-bundle write path via `fakeAttester`, nil-attester guard, NArg guard, code-based assertions) are all addressed.

Nice resolution on the digest: hashing the raw `ReadFile` bytes with length-prefixed framing (`u64-BE(len) || bytes`) fixes both the re-serialization concern (what ships is what's signed, multi-doc safe) and the boundary-ambiguity nit in one change.

Remaining items are non-blocking and your calls: keeping the commands under `recipe` (reasonable either way), skipping the positive `Verify` path test (consistent with how `aicr verify` / `evidence verify` are tested), and the release-abort note (#8), which mirrors the existing binary hook. LGTM.

mchmarny:
Thanks @lockwobr — taking each of the remaining items in turn:

**1. CLI namespace (`recipe sign-catalog` vs `aicr catalog {sign,verify}`) — decision: keep under `recipe`.** `recipe list` (which landed in #1208 the same week) already established the precedent that recipe-data operations live under `recipe`, and the catalog *is* recipe data. The cost of being wrong is one release of compat-shim aliasing (`aicr catalog verify` → `aicr recipe verify-catalog`) once we see how operators actually use it; the cost of an early namespace split is harder to walk back. Filed [#1217](https://github.com/NVIDIA/aicr/issues/1217) to revisit before V1 GA if a third catalog verb appears.

**2. Release-abort on Fulcio failure — confirmed intentional.** Fail-closed is the right posture for a supply-chain release: a release missing its catalog signature is worse than a delayed release. The Fulcio dependency surface is *identical* to the existing binary-attestation hook above it (same endpoints, same OIDC token, same network path), so this PR doesn't introduce a new failure mode, it just doubles the per-release call count to public-good Sigstore. If Fulcio outage ever becomes a recurring release pain point, the right fix is workflow-level retry-with-backoff wrapping the goreleaser invocation, not hook-level — tracked under the supply-chain hardening epic #1149.

**3. `--format json|yaml` for `verify-catalog`** — skipping for the same reason given on the inline thread: `aicr verify` and `aicr evidence verify` (the two parallel verification commands in this repo) both produce fixed human-readable text without a format flag. Adding one only to `verify-catalog` would diverge from the convention; adding it to all three at once is the right move when a downstream consumer surfaces a concrete need. Filed [#1218](https://github.com/NVIDIA/aicr/issues/1218) to track the cross-cutting `--format` addition.

**4. Success-path tests** — agreed, accepting the gap. The sigstore-go verification is exercised end-to-end by the existing `pkg/bundler/verifier` integration tests, and committing a pre-signed bundle fixture means regenerating it on every Sigstore trust-root rotation — net cost > coverage gain. The `runRecipeSignCatalogCmd` OIDC-resolution path stays uncovered for the same reason — instrumenting it would require an attester-injection seam purely for the test. The CI-only hidden command is exercised end-to-end by the goreleaser hook on every release, which is the real signal.

**5. Arg-order asymmetry (`Sign(ctx, provider, opts)` vs `Verify(ctx, bundlePath, provider, opts)`)** — acknowledged as cosmetic, leaving as-is. `verify-catalog`'s bundle path is the positional argument users actually pass on the command line; threading it through `VerifyOptions` would push CLI shape into library shape. The asymmetry mirrors the equivalent in `pkg/bundler/verifier` (`Verify(ctx, bundleDir, opts)`).

All five threads were addressed via individual inline replies and resolved earlier. CI re-run on `6326b4ce` is green. Ready to merge from my end.

Issue #1220, opened 2026-06-08
Ship chainsaw binary + wire deployment-phase runner (#660 PR 2)


mchmarny:
## Summary

Ship the `chainsaw` binary in the deployment validator image and wire the
Chainsaw runner into `aicr validate --phase deployment` so the
hydrated `ComponentRef.HealthCheckAsserts` content from #660 PR 1
actually executes. Codify timeouts, parallelism, and Job envelope margin
as named constants. Restrict registry-declared assert content to a
read-only allowlist.

Depends on #660 PR 1 (hydration mechanism).

### Validator image + library

- Ship the `chainsaw` binary in the deployment validator image:
  - Reuse the version and per-platform checksums already pinned in
    `.settings.yaml` (`chainsaw: 'v0.2.14'` and `chainsaw_checksums`); do
    not introduce a second pin in `Dockerfile` or goreleaser config.
  - Use multi-stage `COPY --from=...` from an upstream chainsaw release
    artifact into the existing `gcr.io/distroless/static-debian12:nonroot`
    base (replace the base only if signing/scan coverage requires it).
  - The `exec.LookPath` / `/usr/local/bin/chainsaw` fallback in
    `validators/chainsaw/binary.go:41-47` must find the binary at runtime.
- Bump the `kyverno/chainsaw` Go library in `go.mod` in lockstep with
  the image binary; raw-K8s-YAML (library path) and Chainsaw Test format
  (binary path) dispatch in `validators/chainsaw/runner.go:132-143` must
  stay aligned.

### Runtime semantics

- Per-component Chainsaw execution timeout: `defaults.ChainsawAssertTimeout`.
- Outer envelope: the enclosing `expected-resources` validator catalog
  timeout (and therefore Job `activeDeadlineSeconds`) must exceed
  `defaults.ChainsawAssertTimeout` by a named-constant margin
  (`defaults.JobEnvelopeMargin`, ≥ 60s) so Chainsaw has headroom for
  process termination, temp-dir cleanup, and log flushing before the
  Job SIGKILLs the container.
- Bounded parallelism across components: `defaults.ChainsawMaxParallel`.
- Dispatch: raw K8s resource YAML via the chainsaw Go library
  (`checks.Check`); `apiVersion: chainsaw.kyverno.io/v1alpha1` `kind:
  Test` via the binary (per the package comment at
  `validators/chainsaw/runner.go:15-20`).

### Gate relaxation + CLI output

[...]

Issue #660, opened 2026-04-23
Reuse Chainsaw health checks to improve readiness check  in `aicr validate --phase deployment`


yuanchen8911:
## Summary

Follow-up to closed issue #622. After discussion, the project is pivoting away
from the pure-Go `readiness` contract proposed there and instead adopting the
**Chainsaw-reuse approach** (Option 2 in #622) to deepen `aicr validate --phase
deployment`.

The problem is unchanged: today's deployment validation is shallow. Baseline
checks (`Active` namespaces, `expectedResources`) plus the three targeted
GPU-specific gaps from #611 are not symmetric across the registry, so most
components still get only thin post-install validation.

This issue has two deliverables:

1. **Wire the existing Chainsaw health checks into `aicr validate --phase
   deployment`** so deployment validation uses the same assertions that already
   live under `recipes/checks/*` and are referenced by `healthCheck.assertFile`
   in `recipes/registry.yaml`.
2. **Enhance the existing Chainsaw checks** where they are too shallow, and
   add checks for the components that don't have one yet, so deployment
   validation is symmetric across the whole registry.

## Problem / Use Case

After #611, deployment validation is stronger but still asymmetric:

- baseline checks run for enabled components (`Active` namespaces, healthy
  declared `expectedResources`)
- targeted deep checks run for three GPU-specific gaps
  (`skyhook-customizations`, `nvidia-dra-driver-gpu`, `gpu-operator`)
- most other components get only shallow validation unless they declare
  `expectedResources`

Meanwhile, `recipes/checks/*` already contains Chainsaw health checks for
19 of the 22 registry components, and `healthCheck.assertFile` is already
declared for them in `recipes/registry.yaml`. These checks are the
highest-fidelity readiness logic the project has, but they are not wired
into `aicr validate --phase deployment` today:

- `healthCheck.assertFile` is not hydrated into
  `ComponentRef.HealthCheckAsserts` during recipe resolution
- `pkg/recipe/metadata.go` intentionally skips loading the assert content
- the deployment validator image does not ship the `chainsaw` binary
- the deployment-phase Chainsaw runner that exists in the code path is
  therefore dormant

The result: the existing Chainsaw assertions are maintained but unused by
`aicr validate --phase deployment`, and the deployment phase re-implements a
shallower version of the same intent.

## Goal

Make `aicr validate --phase deployment` execute the existing Chainsaw health
checks for every registry component that has one, and enhance / backfill
[...]

PR #631, opened 2026-04-21
docs: add ADR for shared component checks (deprecated)

yuanchen8911:
## Summary

Add ADR-006 to document the proposed registry-driven component check-definition
model and the intended convergence on a shared validator entrypoint.

## Motivation / Context

Issue `#622` became large enough to warrant a durable architecture record, and
the discussion with `#610` made the broader direction clearer. This ADR now
captures:

- a shared check-definition model in `recipes/registry.yaml`
- the initial structured contract (`readiness`, `customChecks`, `crds`)
- the intended convergence on `aicr validate` as the shared validator
  entrypoint, including a wait-capable mode for deploy / cleanup and
  post-install validation flows
- the rationale for keeping the structured contract small rather than turning
  it into a general-purpose DSL
- the tradeoff versus reusing Chainsaw directly at validator runtime, while
  still leaving room for a Chainsaw- or asset-based override path when the
  structured schema is not sufficient

This PR is design-only. It does not include the `#622` implementation; that
remains follow-up work.

Fixes: N/A
Related: #622, #610

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] CLI (`cmd/aicr`, `pkg/cli`)
- [ ] API server (`cmd/aicrd`, `pkg/api`, `pkg/server`)
- [x] Recipe engine / data (`pkg/recipe`)
- [ ] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [ ] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [x] Validator (`pkg/validator`)
- [ ] Core libraries (`pkg/errors`, `pkg/k8s`)
- [x] Docs/examples (`docs/`, `examples/`)
- [ ] Other: ____________

## Implementation Notes

- Adds `docs/design/006-deployment-validation.md` in the existing ADR format.
- Reframes the ADR from a deployment-validator-specific contract to a shared
  component check-definition model that different deployment-related phases can
  consume.
- Defines the small structured contract in `recipes/registry.yaml`:
  `readiness`, `customChecks`, and `crds`.
- Records the initial validator subset, including exact workload identities,
  CRD establishment, merge semantics, wait-capable validator convergence, and
[...]

Issue #622, opened 2026-04-20
Add registry-driven deployment readiness for all components


yuanchen8911:
## Summary

> Follow-up note: this issue is the broader all-components follow-up to [#607](https://github.com/NVIDIA/aicr/issues/607) and its narrow implementation PR [#611](https://github.com/NVIDIA/aicr/pull/611). It does not widen or reopen that GPU-scoped work.

Follow up to issue `#607` and PR `#611` with a registry-driven readiness contract so `aicr validate --phase deployment` becomes **deeper, more symmetric, and deployer-neutral for all current registry components**, not just baseline namespace / `expectedResources` checks plus the targeted GPU-specific checks.

This follow-up is the broader next step: make deployment-phase readiness declarative in the registry and apply it consistently across all components.

This issue has two explicit deliverables:
1. **Registry-driven deep deployment checks**: for every component, deployment validation should run the deep checks declared in `recipes/registry.yaml`:
   - generic workload readiness
   - CRD readiness where applicable
   - fixed custom checks where generic readiness is not sufficient
2. **Registry enforcement for new components**: after the current migration lands, adding a new registry component must require declaring at least one readiness coverage path:
   - `readiness`
   - `customChecks`
   - top-level `crds`

Important framing: this issue is **not** trying to reproduce the full expressiveness of the existing repo-side Chainsaw health checks. The v1 contract is intentionally narrower and more operationally stable.

## Problem / Use Case

After `#611`, deployment validation is stronger but still asymmetric:
- baseline checks run for enabled components (`Active` namespaces, healthy declared `expectedResources`)
- targeted deep checks run for the three GPU-specific gaps (`skyhook-customizations`, `nvidia-dra-driver-gpu`, `gpu-operator`)
- many other components still have only shallow validation unless they declare `expectedResources`

That means:
- new registry components do **not** automatically get deep deployment validation
- the validator surface is still partly code-driven instead of registry-driven
- some shared-namespace components still cannot be mapped cleanly without an exact-workload escape hatch
- deployer-neutral deep readiness is not yet symmetric across the registry

The goal of this follow-up is to make the deployment validator:
1. deeper and more symmetric across all registry components
2. deployer-neutral (no Helm API, no release metadata, no release-scoped labels)
3. simple, manageable, and extensible when new components are added

### Option 1. Add explicit readiness fields and implement deployment validation in pure Go

Add a registry-driven readiness contract and have the deployment validator interpret it using Kubernetes API state.

- Shape:
[...]

PR #1231, opened 2026-06-08
feat(recipe): hydrate healthCheck.assertFile + suppression sentinel

### LINKED ISSUE #1219: Hydrate healthCheck.assertFile + suppression sentinel (#660 PR 1)
## Summary

Hydrate `healthCheck.assertFile` content into
`ComponentRef.HealthCheckAsserts` during recipe resolution so the
deployment validator can later consume registry-declared Chainsaw asserts.
Add an overlay-driven suppression sentinel as the rollback path.

This PR introduces the **mechanism only**. The deployment-phase runtime
remains unchanged in this PR — see parent #660 PR 2 for the runner wiring
and image change. The hydrated content is dormant until PR 2 ships the
`chainsaw` binary.

## Scope

- Hydrate `healthCheck.assertFile` content into
  `ComponentRef.HealthCheckAsserts` during recipe resolution. Routing
  goes through the `DataProvider` bound to the `Builder` / result, so
  per-tenant isolation and external `--data` overlays continue to work.
- Validator Jobs must continue to consume only the mounted `recipe.yaml`;
  they must not read `registry.yaml` at runtime.
- Add a per-recipe **suppression sentinel** so a leaf overlay can clear
  an inherited hydration (rollback path for a regressing upstream check,
  and the mechanism by which an external `--data` directory overrides
  registry-declared content). Extend the overlay merge in
  `pkg/recipe/metadata.go` to honor the sentinel.
- Rewrite the stale `// NOTE: healthCheck.assertFile content is
  intentionally NOT loaded here.` block in
  `pkg/recipe/metadata.go:199-205` so the comment describes the new
  hydration contract rather than the abandoned reason.
- Make hydration behavior explicit for components disabled via
  `overrides.enabled: false` — should the hydrated content land in the
  on-disk `recipe.yaml` artifact or be filtered? Document and test.
- Tests:
  - hydration loads content from registry-declared `assertFile`
  - suppression sentinel in overlay clears inherited content
  - external `--data` directory overrides registry-declared content
  - `applyRegistryDefaults` is idempotent under repeated calls
  - existing `pkg/recipe` table-driven tests still pass

## Out of Scope (lands in later PRs under #660)

- Shipping the `chainsaw` binary in the deployment validator image (PR 2)
- Relaxing `validators/deployment/expected_resources.go:86` (PR 2)
- Read-only assertion allowlist enforcement (PR 2 enforces at runtime,
  PR 5 enforces at lint time)
- Backfilling missing checks (PR 3)
- Enhancing existing checks (PR 4)

## Acceptance Criteria

- Recipe resolution hydrates `healthCheck.assertFile` content onto each
  resolved `ComponentRef.HealthCheckAsserts` in `RecipeResult`.
- Hydration routes through the result's bound `DataProvider`.
- Suppression sentinel in an overlay clears an inherited
  `HealthCheckAsserts` value.
- The stale `NOTE` comment in `pkg/recipe/metadata.go` is rewritten.
- No change to deployment-phase runtime behavior in this PR; the new
  field on disk is dormant content.
- Table-driven tests cover hydration, suppression, and external-data
  override paths.

## Related

- Parent epic: #660
- Closed predecessor: #622
- ADR: #631

mchmarny:
## Summary

Hydrate registry-declared `healthCheck.assertFile` content onto
`ComponentRef.HealthCheckAsserts` during recipe resolution. Add an
overlay-driven suppression sentinel as the rollback path. PR 1 of 5
under epic #660.

## Motivation / Context

Today `recipes/checks/*` carries Chainsaw health checks for 22 registry
components, referenced by `healthCheck.assertFile` in
`recipes/registry.yaml`, but the content is never loaded:
`pkg/recipe/metadata.go` intentionally skipped hydration because the
deployment validator image does not ship the `chainsaw` binary, so
populating `ref.HealthCheckAsserts` would activate
`validators/deployment/expected_resources.go:86` and fail every
component invocation. The dormant content is the foundation for
deepening `aicr validate --phase deployment` — see epic #660 (
predecessor #622, ADR #631).

This PR introduces the hydration mechanism only. Runtime behavior is
preserved by gating activation on `ChainsawBinary.Available()`; the
binary ships in #1220.

Fixes: #1219
Related: #660, #1220, #622, #631

## Type of Change

## Component(s) Affected

- [x] Validator (`validators/deployment`, `validators/chainsaw`)

## Implementation Notes

- **Hydration** (`pkg/recipe/metadata_store.go`): `applyRegistryDefaults`
  now calls `hydrateHealthCheckAsserts`, which loads each
  registry-declared `assertFile` through the bound `DataProvider` and
  stamps the content onto the matching `ComponentRef`. Routing goes
  through the per-result provider so per-tenant isolation and external
  `--data` overlays continue to work.
- **Suppression sentinel**: new `HealthCheckSkip bool` field on
  `ComponentRef`. Leaf overlay or external `--data` overlay sets it to
  clear inherited hydration (rollback for a regressing upstream check).
  Merge semantics mirror `Cleanup` (set-if-true).
  `mixinComponentRefSafeForMerge` rejects mixins that set it so a mixin
  cannot silently suppress an inherited check.
- **Inline-wins**: when an overlay declares `HealthCheckAsserts` inline,
  hydration leaves it alone — never silently overwrite caller intent.
- **Disabled components**: hydrated unconditionally so the on-disk
  `recipe.yaml` artifact carries the same content regardless of
  enablement; runtime filtering by `enabledComponentRefs` strips them
  from execution.
- **Runtime gate** (`validators/chainsaw/binary.go`,
  `validators/deployment/expected_resources.go`): `ChainsawBinary` gains
  an `Available()` method (returns `false` when `exec.LookPath("chainsaw")`
  fails). The deployment validator checks it before dispatching
  hydrated assert content; absent binary logs once and skips, preserving
  today's behavior (these components had no validation path at all
  pre-hydration). PR #1220 ships the binary and the gate naturally
  lights up.
- **Stale NOTE rewrite**: the `// NOTE: healthCheck.assertFile content
  is intentionally NOT loaded here.` block in
  `pkg/recipe/metadata.go:199-205` now describes the new contract
  instead of the abandoned reason.

## Testing

```bash
go test -race -count=1 ./pkg/recipe/... ./validators/...
golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./validators/deployment/... ./validators/chainsaw/...
```

- New `pkg/recipe/hydrate_test.go` covers:
  - hydration from registry assertFile
  - skip sentinel suppresses
  - inline `HealthCheckAsserts` is preserved (overlay wins)
  - unknown-component no-op
  - registered-but-no-assertFile no-op
  - provider read failure → wrapped `ErrCodeInternal` with `component` + `assertFile` context
  - nil-provider falls back to embedded
  - merge propagates `HealthCheckSkip` (set-if-true)
  - mixin merge rejects `HealthCheckSkip`
- New `validators/chainsaw/binary_test.go` covers both branches of
  `ChainsawBinary.Available()` (PATH miss → false, PATH hit → true).
- Coverage: `pkg/recipe` 86.0% → 86.2% (+0.2%); `validators/chainsaw`
  11.2% → 14.4% (+3.2%).
- All existing `pkg/recipe`, `validators/deployment`,
  `validators/chainsaw` tests pass under `-race`.
- `pkg/trust` `TestUpdate_Success` fails locally due to Sigstore CDN
  network restrictions in my sandbox — pre-existing, not from this PR.

## Risk Assessment

**Rollout notes:** Validation outcomes are unchanged — the chainsaw
dispatch path is skipped because `ChainsawBinary.Available()` returns
false until #1220 ships the binary, and components with a registry
`assertFile` had no validation path at all pre-hydration. One
user-visible delta: a single `slog.Warn("chainsaw binary not
available; skipping registry-declared health check assertions", ...)`
now fires per deployment-phase run when hydrated content is present
and the binary is missing. The resolved recipe + bundled `recipe.yaml`
now carry `healthCheckAsserts` content (previously empty); downstream
consumers tolerate the extra field since it was already in the
`ComponentRef` schema. To revert: drop the hydration call from
`applyRegistryDefaults` and remove `HealthCheckSkip` / the `Available()`
gate.

## Checklist

- [x] No user-facing CLI/API/recipe-field behavior changed (docs
      update follows in #1220 when the runtime path activates)

### DIFF
--- a/pkg/recipe/metadata_store.go
+++ b/pkg/recipe/metadata_store.go
@@ -26,6 +26,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/NVIDIA/aicr/pkg/defaults"
 	aicrerrors "github.com/NVIDIA/aicr/pkg/errors"
 	"gopkg.in/yaml.v3"
 )
@@ -1129,20 +1130,27 @@ func mixinComponentRefSafeForMerge(c ComponentRef) (string, bool) {
 		return "expectedResources", false
 	case c.HealthCheckAsserts != "":
 		return "healthCheckAsserts", false
+	case c.HealthCheckSkip:
+		return "healthCheckSkip", false
 	}
 	return "", true
 }
 
-// applyRegistryDefaults fills in ComponentRef fields from ComponentConfig defaults.
-// This allows registry.yaml to specify default values that are applied to components
-// that don't explicitly set them in recipes. Returns an error if the registry
-// cannot be loaded — silently no-op'ing would emit partial ComponentRefs that
-// downstream bundlers would reject far from the root cause.
+// applyRegistryDefaults fills in ComponentRef fields from ComponentConfig
+// defaults AND hydrates healthCheck.assertFile content via the bound
+// DataProvider. Returns an error if the registry cannot be loaded or if
+// hydration cannot read a declared assertFile — silently no-op'ing would
+// emit partial ComponentRefs that downstream bundlers / validators would
+// reject far from the root cause.
 //
-// The provider parameter routes the registry lookup through a specific
-// DataProvider so per-provider isolation holds even when the package-global
-// provider has since been swapped. A nil provider falls back to the
-// package-global DataProvider via GetComponentRegistryFor.
+// The provider parameter routes both the registry lookup and assertFile
+// reads through a specific DataProvider so per-provider isolation holds
+// even when the package-global provider has since been swapped. A nil
+// provider falls back to the package-global DataProvider via
+// GetComponentRegistryFor.
+//
+// See hydrateHealthCheckAsserts for the skip / inline-wins / no-assertFile
+// rules and #1219 for the motivation.
 func applyRegistryDefaults(provider DataProvider, refs []ComponentRef) error {
 	registry, err := GetComponentRegistryFor(provider)
 	if err != nil {
@@ -1154,6 +1162,61 @@ func applyRegistryDefaults(provider DataProvider, refs []ComponentRef) error {
 		if config != nil {
 			refs[i].ApplyRegistryDefaults(config)
 		}
+	}
+
+	return hydrateHealthCheckAsserts(provider, registry, refs)
+}
+
+// hydrateHealthCheckAsserts loads each registry-declared healthCheck.assertFile
+// through the bound DataProvider and stamps the content onto the matching
+// ComponentRef.HealthCheckAsserts. Hydration is skipped when:
+//   - the overlay has set HealthCheckSkip (rollback / external-data override
+//     path — see ComponentRef field doc), or
+//   - the overlay already declared HealthCheckAsserts inline (the inline value
+//     wins; never silently overwrite caller intent), or
+//   - the registry has no assertFile entry for this component.
+//
+// Disabled components (overrides.enabled: false) ARE hydrated unconditionally
+// so the on-disk recipe.yaml artifact carries the same content regardless of
+// enablement; runtime execution is filtered separately by enabledComponentRefs.
+//
+// Read-bounded with defaults.FileReadTimeout per call to mirror
+// loadComponentRegistryFor — a hung backing store can't park recipe
+// resolution indefinitely. nil provider falls back to the embedded default,
+// matching applyRegistryDefaults / GetComponentRegistryFor.
+//
+//nolint:contextcheck // ctx is bounded internally; threading caller ctx would require a public-API change to applyRegistryDefaults and every transitive caller — same tradeoff as loadComponentRegistryFor.
+func hydrateHealthCheckAsserts(provider DataProvider, registry *ComponentRegistry, refs []ComponentRef) error {
+	if provider == nil {
+		provider = defaultEmbeddedProvider
+	}
+	for i := range refs {
+		ref := &refs[i]
+		if ref.HealthCheckSkip {
+			continue
+		}
+		if ref.HealthCheckAsserts != "" {
+			continue
+		}
+		config := registry.Get(ref.Name)
+		if config == nil || config.HealthCheck.AssertFile == "" {
+			continue
+		}
+		ctx, cancel := context.WithTimeout(context.Background(), defaults.FileReadTimeout)
+		data, err := provider.ReadFile(ctx, config.HealthCheck.AssertFile)
+		cancel()
+		if err != nil {
+			// PropagateOrWrap so a structured ReadFile error (e.g.,
+			// ErrCodeTimeout from the bounded ctx, or ErrCodeNotFound from
+			// the layered provider) preserves its inner code instead of
+			// being flattened to ErrCodeInternal. Falls back to Wrap for
+			// non-structured stdlib errors (e.g., fs.ErrNotExist), which
+			// is what the embedded provider surfaces today.
+			return aicrerrors.PropagateOrWrap(err, aicrerrors.ErrCodeInternal,
+				fmt.Sprintf("failed to read healthCheck.assertFile for component %q from %q",
+					ref.Name, config.HealthCheck.AssertFile))
+		}
+		ref.HealthCheckAsserts = string(data)
 	}
 	return nil
 }
--- a/validators/chainsaw/binary.go
+++ b/validators/chainsaw/binary.go
@@ -30,21 +30,36 @@ type ChainsawBinary interface {
 	// RunTest executes chainsaw test against the given test directory.
 	// Returns whether all tests passed, the combined output, and any execution error.
 	RunTest(ctx context.Context, testDir string) (passed bool, output string, err error)
+	// Available reports whether the chainsaw binary is callable from this
+	// process. Used by the deployment validator to skip Chainsaw Test-format
+	// dispatch when the binary is absent (e.g., the validator image has not
+	// shipped chainsaw yet), preserving today's no-op behavior while
+	// registry-declared HealthCheckAsserts content hydrates upstream in
+	// pkg/recipe.
+	Available() bool
 }
 
 type chainsawBinary struct {
-	binPath string
+	binPath   string
+	available bool
 }
 
 // NewChainsawBinary creates a ChainsawBinary that invokes the chainsaw CLI.
 // It resolves the binary path from PATH, falling back to /usr/local/bin/chainsaw.
+// Availability is recorded at construction time so callers can branch on it
+// without repeating the exec.LookPath probe.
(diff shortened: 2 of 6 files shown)


lockwobr (dismissed):

pkg/recipe/metadata_store.go
```diff
@@ -1155,5 +1163,58 @@ func applyRegistryDefaults(provider DataProvider, refs []ComponentRef) error {
...
+		data, err := provider.ReadFile(ctx, config.HealthCheck.AssertFile)
+		cancel()
+		if err != nil {
+			return aicrerrors.WrapWithContext(aicrerrors.ErrCodeInternal,
+				"failed to read healthCheck.assertFile",
+				err,
+				map[string]any{
+					"component":  ref.Name,
+					"assertFile": config.HealthCheck.AssertFile,
+				})
```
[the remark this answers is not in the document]
mchmarny: Good catch — applied in 330e5990. Switched to `PropagateOrWrap` so a structured ReadFile error (e.g., `ErrCodeTimeout` from the bounded ctx) survives propagation instead of being flattened to `ErrCodeInternal`. Component name and assertFile path are now embedded in the message string (the structured Context map is dropped, since `PropagateOrWrap`'s fallback `Wrap` doesn't carry context). Test `TestHydrateHealthCheckAsserts_ReadError` updated to assert on message contents accordingly.

lockwobr (approved):

njhensley (approved):

yuanchen8911 (commented):
Todo

validators/deployment/expected_resources.go
```diff
@@ -104,25 +104,40 @@ func checkExpectedResources(ctx *validators.Context) error {
...
+		// Gate activation on chainsaw binary availability. Hydration of
+		// healthCheck.assertFile populates ref.HealthCheckAsserts upstream
+		// (pkg/recipe), but the deployment validator image does not ship the
+		// chainsaw binary yet — see issue #1220. Without this gate, every
+		// component with a registry-declared assertFile would trigger a
+		// failed binary invocation. When the binary is absent, log once and
+		// skip; this preserves the pre-hydration runtime behavior (these
+		// components had no validation path at all before).
+		bin := chainsaw.NewChainsawBinary()
+		if !bin.Available() {
```
yuanchen8911, line 116: **Raw asserts silently skipped when the chainsaw binary is absent → false PASS.**

This `if !bin.Available()` gate skips the *entire* `chainsawAsserts` batch. But `chainsaw.Run` dispatches **raw K8s resource YAML to the Go assertion path** (`validators/chainsaw/runner.go:123-130`, `assertRawResources` at `:184-234`) — only `kind: Test` content actually needs the CLI binary. So an inline overlay `HealthCheckAsserts` or an external `--data` raw assert is silently ignored, and deployment validation can **pass without ever evaluating it**.

Registry-declared checks are all `kind: Test` today, so this bites inline/external-data raw asserts specifically.

**Suggested fix:** partition the asserts — gate only the Test-format subset on `bin.Available()`, and always run the raw-YAML subset through the Go library.

validators/chainsaw/binary.go
```diff
@@ -30,22 +30,37 @@ type ChainsawBinary interface {
...
+// Availability is recorded at construction time so callers can branch on it
+// without repeating the exec.LookPath probe.
 func NewChainsawBinary() ChainsawBinary {
 	binPath, err := exec.LookPath("chainsaw")
 	if err != nil {
-		binPath = "/usr/local/bin/chainsaw"
+		// Fall through with the canonical install path; RunTest will surface
+		// the missing-binary error if invoked, but Available() reports false
+		// so the deployment validator can short-circuit upstream.
+		return &chainsawBinary{binPath: "/usr/local/bin/chainsaw", available: false}
```
yuanchen8911, line 57: **`Available()` reports false even when the fallback binary is callable.**

`NewChainsawBinary` returns `available: false` whenever `exec.LookPath("chainsaw")` fails, without checking the documented fallback `/usr/local/bin/chainsaw`. If chainsaw is installed there but that dir isn't on `PATH`, `RunTest` would invoke it successfully — yet `Available()` reports false and the deployment validator skips all Test-format checks.

**Suggested fix:** `os.Stat` + executable-bit check the fallback path before returning unavailable, so the new gate doesn't disable a binary that `RunTest` can actually run.

validators/deployment/expected_resources.go
```diff
@@ -104,25 +104,40 @@ func checkExpectedResources(ctx *validators.Context) error {
 	failures = append(failures, verifyGPUReadinessSignals(ctx, enabledRefs)...)
 
 	if len(chainsawAsserts) > 0 {
```
yuanchen8911, line 106: **Hydrated `HealthCheckAsserts` and `ExpectedResources` are mutually exclusive at runtime** (refers to the enqueue gate at **line 86**: `if ref.HealthCheckAsserts != "" && len(ref.ExpectedResources) == 0`).

Asserts are only enqueued when `len(ref.ExpectedResources) == 0`, but hydration now stamps registry assert content onto refs that *also* declare expected resources — so the chainsaw check is silently dropped for them.

In-tree example: `k8s-nim-operator` has a registry `assertFile` (`recipes/registry.yaml:515-521`) **and** overlay `expectedResources` (`recipes/overlays/h100-eks-ubuntu-inference-nim.yaml:43-55`).

**Suggested fix:** either run both checks, or make `assertFile` fallback-only explicit and skip hydration when `ExpectedResources` is present (and document it on the field).

yuanchen8911 (commented):
Three inline findings below. All are latent today because the chainsaw binary isn't shipped yet (#1220 gates the branch off), but each becomes an active correctness gap the moment the binary lands.

mchmarny:
@yuanchen8911 — thanks for the careful cross-review. All three findings addressed in follow-up #1234:

1. **Major (binary gate over-blocks raw YAML):** validator now partitions Test-format vs raw-YAML asserts and gates only the Test-format subset on `bin.Available()`. `chainsaw.IsChainsawTest` exported for the classifier.
2. **Medium (fallback path executability):** `NewChainsawBinary` now stats `/usr/local/bin/chainsaw` for an executable regular file after `exec.LookPath` misses, so `Available()` is consistent with what `RunTest` would actually invoke.
3. **Medium (hydration + ExpectedResources mutex):** `hydrateHealthCheckAsserts` now skips when `ExpectedResources` is non-empty, so the resolved recipe matches what the validator actually runs. PR #1220 will simultaneously drop both the validator-side mutex and this hydration skip so both paths run side-by-side.

Tests added for each. Dismissals confirmed (external read bound + mock surface) — no further action there.

Issue #1223, opened 2026-06-08
Registry lint guard for healthCheck.assertFile + allowlist (#660 PR 5)


mchmarny:
## Summary

Add a `pkg/recipe` registry-validation guard that blocks new registry
components from landing without a `healthCheck.assertFile`, and enforces
the read-only assertion allowlist on registry-declared Chainsaw content.

Lands after #660 PR 3 (backfill) so all current components are covered
before the guard begins enforcing.

### Required `healthCheck.assertFile` on new components

- Lint-level check in `pkg/recipe` that runs as part of the existing
  repo CI / `make qualify` surface (not new runtime behavior).
- Rejects any `recipes/registry.yaml` entry that lacks
  `healthCheck.assertFile` after this PR lands.
- Existing components are grandfathered only insofar as they all
  declare `assertFile` by the time this PR ships (#660 PR 3 must land
  first).

### Read-only allowlist enforcement

- Same lint-time check rejects any registry-declared Chainsaw content
  that uses an operation other than `assert` / `error`.
- Mirrors the runtime enforcement from #660 PR 2 so violations are
  caught at PR time, not in CI runs that exercise the validator.

### Surface decision (open)

- Lint at `make qualify` time (custom Go check in `pkg/recipe` test
  suite) vs. load-time at `aicr recipe` (recipe resolution rejects).
- The original #660 wording leans lint-time. Confirm during
  implementation; both are valid, but pick one — not both.

## Out of Scope

- Migrating any existing check or backfilling missing checks (those
  are #660 PR 3 / PR 4).
- Removing `expectedResources` (the long-term disposition of the
  parallel reporting path is a separate decision; see #660 PR 2 for
  the transition period).

## Acceptance Criteria

- `make qualify` (or the chosen surface) fails when a new registry
  entry lacks `healthCheck.assertFile`.
- `make qualify` fails when registry-declared assert content uses an
  operation outside the `assert` / `error` allowlist.
- The check produces a structured `ErrCodeInvalidRequest` with the
  offending component name + operation.
- Documentation in `docs/contributor/recipe.md` (or equivalent)
  explains the contract for new components.

## Related

- Parent epic: #660
- Depends on: #660 PR 3 (backfill must be complete first)

PR #611, opened 2026-04-18
enhance(validator): add targeted post-deployment GPU readiness checks

yuanchen8911:
## Summary

Strengthen `aicr validate --phase deployment` by adding the targeted GPU-workload readiness signals the current validator misses — Skyhook customizations, NVIDIA DRA driver, and GPU Operator (`ClusterPolicy`) — on top of the existing namespace and `expectedResources` checks.

Concretely, this PR:
- keeps using the existing deployment phase
- preserves the existing baseline checks:
  - enabled namespaces are `Active`
  - declared `expectedResources` are healthy
- adds the targeted GPU-specific readiness checks the phase was missing:
  - `skyhook-customizations`: Skyhook CR completion
  - `nvidia-dra-driver-gpu`: kubelet-plugin readiness
  - `gpu-operator`: `ClusterPolicy.status.state == ready`

So the main new value is the addition of the targeted Skyhook, DRA, and GPU Operator deep readiness checks on top of the preserved baseline deployment checks.

> Follow-up note: the broader registry-driven work to make deployment-phase validation symmetric and deep for **all** components is now tracked in [#622](https://github.com/NVIDIA/aicr/issues/622). This PR remains intentionally scoped to the targeted GPU readiness gap narrowed in `#607`.

The check stays strictly deployer-neutral: no Helm API calls, no dependence on Argo CD or any other deployer, no reads of release metadata, no release-identity assumptions. Every lookup is a plain Kubernetes API call against live-cluster object shape, so the command can be used as a standalone tool to verify any Kubernetes cluster's runtime install regardless of how the bundle was deployed (`deploy.sh`, Argo CD, kapp, helmfile, plain `kubectl apply`). Validated end-to-end on a real EKS H100 cluster.

## Motivation / Context

`#607` narrowed this follow-up to reusing and enhancing the existing deployment validator path instead of adding a new phase, flag, or Chainsaw runtime dependency. The previous deployment phase verified install/resource existence, but it did not cover the late-converging readiness signals that explain the post-install gap on fresh GPU clusters.

Fixes: #607
Related:
- #602
- #609
- #622
- #516

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

[...]

PR #1235, opened 2026-06-08
feat(validator): ship chainsaw binary; activate deployment-phase runner

### LINKED ISSUE #1220: Ship chainsaw binary + wire deployment-phase runner (#660 PR 2)

mchmarny:
## Summary

Activate the dormant Chainsaw runner so the `HealthCheckAsserts`
content hydrated by #1219 / #1234 actually executes during
`aicr validate --phase deployment`. Ships the chainsaw binary in the
deployment validator image, codifies the Job activeDeadlineSeconds
envelope, enforces a runtime read-only operation allowlist, and lifts
the previous mutex between `ExpectedResources` and registry-declared
`HealthCheckAsserts`.

## Motivation / Context

Predecessor PR #1219 hydrated registry-declared
`healthCheck.assertFile` content onto `ComponentRef.HealthCheckAsserts`,
but execution was gated behind `ChainsawBinary.Available()` because the
deployment validator image had not yet shipped the binary. This PR
ships the binary, removes the gate, and codifies the surrounding
contracts (timeouts, read-only allowlist, dedup output) so the
deployment phase is now driven by registry-declared Chainsaw content.

Fixes: #1220
Related: #660, #1219, #1234, #1223

## Type of Change

## Component(s) Affected

- [x] Validator (`validators/chainsaw`, `validators/deployment`,
      `pkg/validator/catalog`)
- [x] Build/CI/tooling (Makefile, Dockerfile, `.settings.yaml` consumption)
- [x] Docs (`docs/contributor/validator.md`)

### Image

- `validators/deployment/Dockerfile`: new `chainsaw-fetch` multi-stage
  downloads `chainsaw_linux_amd64.tar.gz` from the kyverno/chainsaw
  release, verifies the sha256 against `.settings.yaml`, and `COPY`s
  the binary into the distroless final stage at
  `/usr/local/bin/chainsaw`.
- `Makefile` `image-validators` target reads `CHAINSAW_VERSION` and
  `CHAINSAW_SHA256_LINUX_AMD64` from `.settings.yaml` via `yq` and
  passes them as `--build-arg` **only** to the deployment-phase build
  (conformance + performance images don't use chainsaw).
- Single-arch (linux/amd64). The `linux_arm64` checksum is already in
  `.settings.yaml` so the multi-arch upgrade path is clear.
- `go.mod`'s `github.com/kyverno/chainsaw v0.2.15` is already in
  lockstep with the `.settings.yaml` pin (no bump needed).

### Envelope

- New `defaults.JobEnvelopeMargin = 60s`. Catalog timeout for
  `expected-resources` becomes the Job's `activeDeadlineSeconds`;
  both must exceed `ChainsawAssertTimeout` so chainsaw has headroom
  to terminate, clean up its temp dir, and flush logs before SIGKILL.
- `recipes/validators/catalog.yaml`: `expected-resources` timeout
  bumped `5m → 7m` (= 6m + 60s). `TestExpectedResourcesCatalogEnvelope`
  in `pkg/validator/catalog` asserts the invariant — raising
  `ChainsawAssertTimeout` later without bumping the catalog will fail
  the test.

### Read-only allowlist

- New `validators/chainsaw.ValidateTestReadOnly` parses each
  registry-declared Chainsaw Test YAML and rejects any operation
  outside `{assert, error}`. State-changing (apply, create, delete,
  patch, update) and side-effecting (script, command, wait, sleep,
  podLogs, events, describe, get) operations are rejected with
  `ErrCodeInvalidRequest`.
- Catch / finally / cleanup blocks equivalently restricted.
- `TestValidateTestReadOnly_RegistryContent` sweeps every
  `recipes/checks/*/health-check.yaml` — guarantees no registry-
  declared check violates the contract.
- Bounds the cluster-admin RBAC posture of the deployment validator
  Job (see `pkg/validator/job/rbac.go`). PR #1223 will add the same
  enforcement at lint time.

### Validator runtime

- Dropped the `&& len(ref.ExpectedResources) == 0` gate in
  `validators/deployment/expected_resources.go`. Both paths now run
  per component. Output is source-tagged `[expectedResources]` /
  `[chainsaw]`.
- Hard-removed `ChainsawBinary.Available()` and the associated PR
  #1231 skip path (per pre-PR decision). The binary is now a hard
  requirement of the deployment image; a regression that drops it
  surfaces as a clear `RunTest` error rather than a silent skip.
- `pkg/recipe.hydrateHealthCheckAsserts` no longer skips when
  `ExpectedResources` is set. The transitional skip added in #1234 is
  reverted in lockstep with the validator-side gate drop so the
  artifact matches what runs.

### GPU deep-check migration (deferred)

- `clusterPolicyReady` (issue #611 deep check) migration to
  registry-declared Chainsaw YAML is **deferred** to a focused
  follow-up. The migration must prove assertion equivalence against
  `expected_resources_test.go`'s heavily-mocked ClusterPolicy state
  coverage; bundling that semantic argument here would bloat review.
  Disposition documented in the new doc on `verifyGPUReadinessSignals`.
- `verifyNodewrightReady` (formerly `skyhookReady`) stays in Go:
  names are derived dynamically from recipe `ManifestFiles` at
  validate-time. Static Chainsaw YAML cannot express the
  dynamic-name selector.
- `verifyDRAKubeletPluginReady` stays in Go: chart's full DaemonSet
  name is release-derived. Encoding it would violate the
  deployer-neutrality constraint (#660 issue body).

## Testing

```bash
go test -race -count=1 ./pkg/recipe/ ./pkg/defaults/ ./pkg/validator/catalog/ ./validators/...
golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./pkg/defaults/... ./pkg/validator/catalog/... ./validators/...
```

New tests:

- `pkg/defaults.TestJobEnvelopeMarginInvariant` — guards the
  envelope contract.
- `pkg/validator/catalog.TestExpectedResourcesCatalogEnvelope` —
  guards the catalog timeout is `>= ChainsawAssertTimeout +
  JobEnvelopeMargin`.
- `validators/chainsaw.TestValidateTestReadOnly` (9 sub-cases) +
  `TestValidateTestReadOnly_RegistryContent` (sweeps in-tree checks).
- `validators/chainsaw.TestNewChainsawBinary` (PATH hit + miss).
- `pkg/recipe.TestHydrateHealthCheckAsserts_RunsAlongsideExpectedResources`
  — replaces #1234's transitional `_ExpectedResourcesWins` test.

Coverage delta on changed packages:

- `validators/chainsaw`: 14.4% → 22.1% (+7.7%).
- `pkg/defaults`: 100% (unchanged).
- `pkg/validator/catalog`: 97.5%.
- `validators/deployment`: 44.1%.

**Live-cluster validation required by #660 acceptance criteria —
deferred to @mchmarny per out-of-band agreement (no live cluster in
PR author's environment).**

## Risk Assessment

- [x] **Medium** — Touches the validator image build, drops a runtime
  gate, and adds a runtime allowlist. Easy to revert per-component
  (Dockerfile, catalog.yaml, runtime branches all isolated).

**Rollout notes:** No CLI-flag changes. Output gains source-tag
prefixes (`[expectedResources]` / `[chainsaw]`). Resolved
`recipe.yaml` artifacts for components with both `assertFile` AND
`ExpectedResources` now carry the hydrated chainsaw content
(previously suppressed transitionally). Recipe authors who declared
`expectedResources` to suppress hydrated content must add
`healthCheckSkip: true` instead (existing field from #1219). The
chainsaw binary is now required in the deployment validator image —
operators running custom images must rebuild.

## Checklist

- [x] I updated docs for user-facing behavior changes
  (`docs/contributor/validator.md`)

### DIFF
--- a/docs/contributor/validator.md
+++ b/docs/contributor/validator.md
@@ -10,7 +10,7 @@ contributor view for all four.
 | [**Constraint**](#constraints-declarative) (declarative) | `aicr validate` against a snapshot | Recipe overlay `validation:` block | `pkg/constraints` evaluator (in-process) |
 | [**Container-per-validator check**](#container-per-validator-checks) | `aicr validate` against a live cluster | `validators/<phase>/` + `recipes/validators/catalog.yaml` | One K8s Job per check |
 | [**Component validation**](#component-validations-bundle-time) (bundle-time) | `aicr bundle` | `pkg/bundler/validations/checks.go` + `registry.yaml` `validations:` | In-process Go `ValidationFunc` |
-| [**Chainsaw health check**](#chainsaw-health-checks) | `make check-health` post-deploy | `recipes/checks/<name>/health-check.yaml` | Chainsaw YAML assertions |
+| [**Chainsaw health check**](#chainsaw-health-checks) | Two surfaces: `make check-health` post-deploy locally, AND `aicr validate --phase deployment` in-cluster | `recipes/checks/<name>/health-check.yaml` | Chainsaw YAML (Test format runs via the pinned chainsaw binary shipped in the deployment validator image; raw K8s YAML via the chainsaw Go library) |
 
 Rule of thumb: declarative constraint against a snapshot value → surface 1.
 Active probe of a live cluster → surface 2 or 4. Pre-deployment sanity
@@ -650,11 +650,21 @@ A **chainsaw health check** is a YAML test in
 `recipes/checks/<component>/health-check.yaml` that asserts a
 deployed component's state. Runs against a real cluster (typically a
 Kind cluster after `aicr bundle` + `helm install`) via the
-[Chainsaw](https://kyverno.github.io/chainsaw/) test runner. The
-separation from container-per-validator checks: chainsaw is
-**post-deploy** Helm-chart-author sanity, declarative YAML, no Go
-code. Container checks are **part of `aicr validate`** and run as
-part of the AICR validation contract.
+[Chainsaw](https://kyverno.github.io/chainsaw/) test runner.
+
+The same assertion file now powers TWO surfaces:
+
+1. **`make check-health` / `make check-health-all`** — local Kind-cluster
+   sanity invoked manually by chart authors.
+2. **`aicr validate --phase deployment`** — registry-declared content is
+   loaded into `ComponentRef.HealthCheckAsserts` during recipe
+   resolution (PR #1219) and executed by the deployment validator's
+   chainsaw runner (PR #1220). The deployment validator image
+   ships the pinned chainsaw binary at `/usr/local/bin/chainsaw`; the
+   version + sha256 come from `.settings.yaml` and are wired into the
+   image via Makefile build args. CLI output is source-tagged
+   `[chainsaw]` vs `[expectedResources]` so operators can disambiguate
+   when both paths report on the same component.
 
 **Registration.** A component opts in by declaring
 `healthCheck.assertFile` in `recipes/registry.yaml`:
@@ -691,12 +701,22 @@ spec:
               status: { (availableReplicas > `0`): true }
 ```
 
-Use Chainsaw's `assert` (expected match), `error` (unexpected match
-must not exist), and `script` (shell). Always include an existence
-guard before phase assertions so an empty namespace can't yield a
-vacuous pass. See the
+Use Chainsaw's `assert` (expected match) and `error` (unexpected match
+must not exist). Always include an existence guard before phase
+assertions so an empty namespace can't yield a vacuous pass. See the
 [Chainsaw assert reference](https://kyverno.github.io/chainsaw/latest/operations/check/assert/)
 for the full operator list.
+
+**Read-only allowlist.** Registry-declared assert files MUST use only
+`assert` and `error` operations. The deployment validator Job runs
+under a ServiceAccount bound to cluster-admin, so registry content is
+restricted at runtime to read-only Chainsaw operations
+(`validators/chainsaw/allowlist.go`). Any other operation (`script`,
+`apply`, `create`, `delete`, `patch`, `update`, `wait`, `command`,
+`sleep`, `podLogs`, `events`, `describe`, `get`) is rejected with
+`ErrCodeInvalidRequest`. PR #1223 will add the same enforcement at
+lint time so violations are caught before they ever reach the
+validator.
 
 **Running:**
 
--- a/pkg/defaults/timeouts.go
+++ b/pkg/defaults/timeouts.go
@@ -274,6 +274,21 @@ const (
 	// assertion retries. Assertions are retried at this interval until
 	// they pass or the ChainsawAssertTimeout expires.
 	AssertRetryInterval = 5 * time.Second
+
+	// JobEnvelopeMargin is the headroom added on top of ChainsawAssertTimeout
+	// when computing the validator Job's outer activeDeadlineSeconds and the
+	// expected-resources catalog timeout. Chainsaw needs time after the inner
+	// assert deadline elapses to terminate the binary process, clean up the
+	// temp test directory, and flush log output before the Job's SIGKILL
+	// arrives. Without this headroom the binary is killed mid-cleanup and
+	// operators see truncated output, masking the actual failure cause.
+	//
+	// 60s = 30s for the default Pod terminationGracePeriodSeconds (the
+	// SIGTERM→SIGKILL window — see K8sPodTerminationWaitTimeout above)
+	// plus 30s headroom for pre-chainsaw helper.VerifyResource iteration
+	// and chainsaw startup variance. Tune upward if chainsaw output
+	// truncation is observed in CI runs.
+	JobEnvelopeMargin = 60 * time.Second
 )
 
 // Readiness gate (deploy-time) configuration drives the `gate` CLI Job the
--- a/recipes/validators/catalog.yaml
+++ b/recipes/validators/catalog.yaml
@@ -38,7 +38,17 @@ validators:
     phase: deployment
     description: "Verify expected Kubernetes resources exist and are healthy"
     image: ghcr.io/nvidia/aicr-validators/deployment:latest
-    timeout: 5m
+    # Must be strictly > defaults.ChainsawAssertTimeout (6m) +
+    # defaults.JobEnvelopeMargin (60s) so chainsaw has headroom to
+    # terminate, clean up its temp dir, and flush log output before the
+    # Job's SIGKILL AND so pre-chainsaw work (helper.VerifyResource
+    # iteration) can't starve the chainsaw budget. 8m provides 60s for
+    # cleanup plus another 60s for pre-chainsaw work. The invariant is
+    # verified by TestExpectedResourcesCatalogEnvelope in
+    # pkg/validator/catalog — raising ChainsawAssertTimeout without
+    # raising this in tandem will fail the test. See issue #1220 and
+    # PR #1235 review.
+    timeout: 8m
     args: ["expected-resources"]
     env: []
   - name: gpu-operator-version
--- a/validators/chainsaw/allowlist.go
+++ b/validators/chainsaw/allowlist.go
@@ -1,0 +1,221 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package chainsaw
+
+import (
+	"bytes"
+	stderrors "errors"
+	"fmt"
+	"io"
+
+	"github.com/kyverno/chainsaw/pkg/apis/v1alpha1"
+	yamlv3 "gopkg.in/yaml.v3"
+	"sigs.k8s.io/yaml"
+
+	"github.com/NVIDIA/aicr/pkg/errors"
+)
+
+// ValidateTestReadOnly parses chainsaw Test YAML content (possibly
+// multi-document) and rejects any operation other than `assert` or
+// `error`. Used at runtime to bound the blast radius of registry-declared
+// health checks: the deployment validator Job runs under a ServiceAccount
+// bound to cluster-admin (pkg/validator/job/rbac.go:41-67), so registry
+// content must not be able to invoke state-changing chainsaw operations
+// (apply, create, delete, patch, update) or side-effecting collectors
+// (script, command, wait, sleep, podLogs, events, describe, get, proxy).
+//
+// Multi-document support: a single `---`-separated stream may carry more
+// than one Test; each is unmarshaled and validated independently. Empty
+// documents and non-Test documents (different apiVersion/kind) are
+// skipped. Per PR #1235 review.
+//
+// Both per-step (`spec.steps[].try/catch/finally/cleanup`) and top-level
+// (`spec.catch`) operation lists are validated.
+//
+// Caller contract: invoke only on content that IsChainsawTest reports as
+// Test format. Raw K8s YAML asserts have no operations and are
+// unaffected.
+//
+// Returns ErrCodeInvalidRequest naming the offending document index +
+// step + operation so the operator can pinpoint the registry entry that
+// violated the allowlist. PR #1223 will surface the same rule at lint
+// time so violations are caught before they ever reach the validator.
+func ValidateTestReadOnly(component, yamlContent string) error {
+	dec := yamlv3.NewDecoder(bytes.NewReader([]byte(yamlContent)))
+	for docIdx := 0; ; docIdx++ {
+		var node yamlv3.Node
+		err := dec.Decode(&node)
+		if stderrors.Is(err, io.EOF) {
+			return nil
+		}
+		if err != nil {
+			return errors.Wrap(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("failed to parse chainsaw Test YAML for component %q (document %d)", component, docIdx), err)
+		}
+		// Re-marshal this single document and run through sigs.k8s.io/yaml
+		// so the Test struct gets the JSON-tag-aware unmarshal path (the
+		// chainsaw types are tagged with json, not yaml).
+		buf, err := yamlv3.Marshal(&node)
+		if err != nil {
+			return errors.Wrap(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("failed to re-marshal chainsaw Test document for component %q (document %d)", component, docIdx), err)
+		}
+		var test v1alpha1.Test
+		if err := yaml.Unmarshal(buf, &test); err != nil {
+			return errors.Wrap(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("failed to decode chainsaw Test YAML for component %q (document %d)", component, docIdx), err)
+		}
+		if err := validateTest(component, docIdx, &test); err != nil {
+			return err
+		}
+	}
+}
+
+// validateTest walks both the top-level Spec.Catch list and the
+// per-step Try/Catch/Finally/Cleanup lists, rejecting any operation
+// outside the read-only allowlist.
+func validateTest(component string, docIdx int, test *v1alpha1.Test) error {
+	docLabel := fmt.Sprintf("doc[%d]", docIdx)
+
+	// Top-level spec.catch (sibling of spec.steps, not nested under a
+	// step). Independent of any step's lifecycle and would otherwise
+	// slip past a steps-only walker. EVERY entry is rejected because
+	// CatchFinally has no read-only variant. Per PR #1235 review.
+	for opIdx, cf := range test.Spec.Catch {
+		return errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("component %q %s spec.catch[%d]: operation %q is not in the read-only allowlist (assert, error)",
+				component, docLabel, opIdx, catchFinallyOpName(cf)))
+	}
+
+	for stepIdx, step := range test.Spec.Steps {
+		stepLabel := step.Name
+		if stepLabel == "" {
+			stepLabel = fmt.Sprintf("step[%d]", stepIdx)
+		}
+		for opIdx, op := range step.Try {
+			if name, ok := disallowedOperation(op); ok {
+				return errors.New(errors.ErrCodeInvalidRequest,
+					fmt.Sprintf("component %q %s step %q try[%d]: operation %q is not in the read-only allowlist (assert, error)",
+						component, docLabel, stepLabel, opIdx, name))
+			}
+		}
+		// catch / finally / cleanup blocks are rejected unconditionally
+		// — CatchFinally carries no read-only variant.
+		for opIdx, cf := range step.Catch {
+			return errors.New(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("component %q %s step %q catch[%d]: operation %q is not in the read-only allowlist (assert, error)",
+					component, docLabel, stepLabel, opIdx, catchFinallyOpName(cf)))
+		}
+		for opIdx, cf := range step.Finally {
+			return errors.New(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("component %q %s step %q finally[%d]: operation %q is not in the read-only allowlist (assert, error)",
+					component, docLabel, stepLabel, opIdx, catchFinallyOpName(cf)))
+		}
+		for opIdx, cf := range step.Cleanup {
+			return errors.New(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("component %q %s step %q cleanup[%d]: operation %q is not in the read-only allowlist (assert, error)",
+					component, docLabel, stepLabel, opIdx, catchFinallyOpName(cf)))
+		}
+	}
+	return nil
+}
+
+// disallowedOperation enforces a TRUE allowlist on op: an Operation is
+// valid IFF Assert is set OR Error is set, and nothing else. Per PR #1235
+// review (yuanchen8911), the previous denylist shape was unsafe — chainsaw
+// v0.2.15 added a `proxy` op that issues HTTP requests through the
+// API-server proxy subresource (side-effecting under the validator Job's
+// cluster-admin binding), and the denylist let it through by default.
+// Inverting to allowlist semantics fails closed on every future chainsaw
+// op upstream adds, until a maintainer explicitly audits and permits it.
+//
+// The known-bad switch below produces actionable diagnostics for the
+// current 14 disallowed ops; anything not in either set is reported as
+// "<unrecognized>" and rejected too.
+func disallowedOperation(op v1alpha1.Operation) (string, bool) {
+	switch {
+	case op.Apply != nil:
+		return "apply", true
+	case op.Command != nil:
+		return "command", true
+	case op.Create != nil:
+		return "create", true
+	case op.Delete != nil:
+		return "delete", true
+	case op.Describe != nil:
+		return "describe", true
+	case op.Events != nil:
+		return "events", true
+	case op.Get != nil:
+		return "get", true
+	case op.Patch != nil:
+		return "patch", true
+	case op.PodLogs != nil:
+		return "podLogs", true
+	case op.Proxy != nil:
+		return "proxy", true
+	case op.Script != nil:
+		return "script", true
+	case op.Sleep != nil:
+		return "sleep", true
+	case op.Update != nil:
+		return "update", true
+	case op.Wait != nil:
+		return "wait", true
+	}
+	// True allowlist: anything that isn't Assert or Error is rejected,
+	// including future chainsaw ops not yet enumerated above.
+	if op.Assert == nil && op.Error == nil {
+		return "<unrecognized>", true
+	}
+	return "", false
+}
+
+// catchFinallyOpName returns a diagnostic name for the operation set on
+// a CatchFinally entry. CatchFinally has no Assert or Error variant;
+// every operation it can carry (command, delete, describe, events, get,
+// podLogs, script, sleep, wait) is either state-changing or
+// side-effecting, and is outside the read-only assertion surface. So
+// EVERY non-empty CatchFinally is rejected unconditionally by the
+// caller — this helper exists only to name the offender for diagnostic
+// clarity. Per PR #1235 review (yuanchen8911): inverted from the prior
+// boolean-returning denylist shape so future chainsaw additions to
+// CatchFinally fail closed without needing to update this function.
+func catchFinallyOpName(cf v1alpha1.CatchFinally) string {
+	switch {
+	case cf.Command != nil:
+		return "command"
+	case cf.Delete != nil:
+		return "delete"
+	case cf.Describe != nil:
+		return "describe"
+	case cf.Events != nil:
+		return "events"
+	case cf.Get != nil:
+		return "get"
+	case cf.PodLogs != nil:
+		return "podLogs"
+	case cf.Script != nil:
+		return "script"
+	case cf.Sleep != nil:
+		return "sleep"
+	case cf.Wait != nil:
+		return "wait"
+	}
+	// No known op set, but a non-empty catch/finally/cleanup entry is
+	// still outside the read-only contract — name it generically so the
+	// caller's unconditional rejection has a sensible diagnostic.
+	return "<catch-finally-block>"
+}
--- a/validators/chainsaw/binary_test.go
+++ b/validators/chainsaw/binary_test.go
@@ -20,113 +20,58 @@ import (
 	"testing"
 )
 
-// TestNewChainsawBinary_Available exercises every branch of the
-// availability probe in NewChainsawBinary by manipulating PATH and the
-// canonical fallback path so the binary is found / not found. The
-// fallback-path branch was added in PR #1231 review (yuanchen8911):
-// previously NewChainsawBinary reported Available()==false even when
-// /usr/local/bin/chainsaw existed and was callable.
+// TestNewChainsawBinary covers the two discovery branches:
+//   - exec.LookPath hit  → use the PATH-resolved path
+//   - exec.LookPath miss → fall back to the canonical install path
 //
-// canonicalChainsawPath is swapped to a TempDir entry per subtest so
-// the real /usr/local/bin/chainsaw on developer machines doesn't make
-// the unavailable assertions flake.
-func TestNewChainsawBinary_Available(t *testing.T) {
-	withCanonicalPath := func(t *testing.T, p string) {
-		orig := canonicalChainsawPath
-		canonicalChainsawPath = p
-		t.Cleanup(func() { canonicalChainsawPath = orig })
-	}
-
-	t.Run("unavailable when missing from PATH and fallback path", func(t *testing.T) {
-		t.Setenv("PATH", t.TempDir())
-		withCanonicalPath(t, filepath.Join(t.TempDir(), "chainsaw"))
-		bin := NewChainsawBinary()
-		if bin.Available() {
-			t.Fatal("Available() = true, want false when chainsaw is on neither PATH nor fallback")
-		}
-	})
-
-	t.Run("available when discoverable on PATH", func(t *testing.T) {
-		dir := t.TempDir()
-		stub := filepath.Join(dir, "chainsaw")
-		// 0o755: an exec.LookPath-discoverable executable is the whole
-		// point of this fixture.
-		if err := os.WriteFile(stub, []byte("#!/bin/sh\nexit 0\n"), 0o755); err != nil { //nolint:gosec // test fixture
-			t.Fatalf("write stub: %v", err)
-		}
-		t.Setenv("PATH", dir)
-		withCanonicalPath(t, filepath.Join(t.TempDir(), "chainsaw"))
-		bin := NewChainsawBinary()
-		if !bin.Available() {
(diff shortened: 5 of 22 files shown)


docs/contributor/validator.md
```diff
@@ -650,11 +650,21 @@ A **chainsaw health check** is a YAML test in
...
+   sanity invoked manually by chart authors.
+2. **`aicr validate --phase deployment`** — registry-declared content is
+   loaded into `ComponentRef.HealthCheckAsserts` during recipe
+   resolution (PR #1219) and executed by the deployment validator's
+   chainsaw runner (PR #1220). The deployment validator image
+   ships the pinned chainsaw binary at `/usr/local/bin/chainsaw`; the
+   version + sha256 come from `.settings.yaml` and are wired into the
+   image via Makefile build args. CLI output is source-tagged
+   `[chainsaw]` vs `[expectedResources]` so operators can disambiguate
+   when both paths report on the same component.
```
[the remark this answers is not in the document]
mchmarny: Fixed in c6e3949d. The Chainsaw row in the surface table now reads: "Two surfaces: `make check-health` post-deploy locally, AND `aicr validate --phase deployment` in-cluster" — and mentions that Test format dispatches via the pinned chainsaw binary while raw K8s YAML uses the Go library.

validators/deployment/expected_resources.go
```diff
@@ -82,74 +82,53 @@ func checkExpectedResources(ctx *validators.Context) error {
...
 	failures = append(failures, verifyNamespacesActive(ctx, enabledRefs)...)
 
+	// Both paths run side-by-side now (#1220). Previously the chainsaw
+	// path was gated on len(ExpectedResources) == 0; that mutex was
+	// dropped because (a) the registry-declared assertFile is the deeper
+	// readiness signal and should always run, and (b) the hydration step
+	// in pkg/recipe now suppresses HealthCheckAsserts when an overlay
+	// declares ExpectedResources — see PR #1234. Until #1234 lands,
+	// components with both populated will execute both paths; output is
+	// source-tagged so operators can disambiguate.
```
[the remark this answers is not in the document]
mchmarny: Fixed in c6e3949d. The comment block now describes the current behavior without referencing a future PR: both paths execute when both fields are populated; output is source-tagged `[expectedResources]` / `[chainsaw]`; the `len(ExpectedResources) == 0` gate was dropped in this PR; the transitional hydration skip from #1234 was reverted in lockstep.

validators/deployment/expected_resources.go
```diff
@@ -82,74 +82,53 @@ func checkExpectedResources(ctx *validators.Context) error {
...
+				if r.Output != "" {
+					msg += fmt.Sprintf(":\n%s", r.Output)
 				}
+				if r.Error != nil {
+					msg += fmt.Sprintf("\nerror: %v", r.Error)
+				}
+				failures = append(failures, msg)
 			}
 		}
 	}
```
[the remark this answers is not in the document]
mchmarny: Holding the serial order, declining the fan-out. Rationale:

1. **The actual chainsaw budget is the inner timer, not the outer envelope.** `chainsaw.Run` receives `defaults.ChainsawAssertTimeout` (6m) directly as its per-assert timeout. The outer Job `activeDeadlineSeconds` (7m = 6m + 60s margin) is the SIGKILL backstop, not chainsaw's budget. Chainsaw gets its full 6m regardless of how long expected-resources verification took.

2. **Expected-resources verification is fast.** `helper.VerifyResource` does a single K8s API `Get` per resource — typically sub-100ms. Even for 22 components × 1 resource at 200ms each, that's ~4.4s, well below the 60s envelope margin.

3. **The envelope margin is the right knob, not the serialization model.** If a cluster ever pushes expected-resources verification past 60s, the fix is to bump `JobEnvelopeMargin` (a single named constant), not to add mutex + errgroup concurrent-failure-aggregation complexity.

4. **Complexity cost.** Concurrent fan-out here means: shared `failures` slice under mutex, `buildResourceFetcher` error path threaded through the group, ordering of output lines becomes non-deterministic (operators reading CLI output prefer stable order). Three new failure modes for a problem that doesn't manifest at current scale.

If we ever see real evidence of envelope preemption, the right follow-up is to either bump the margin or partition expected-resources verification itself with errgroup (independent of chainsaw). Tracking as a possible future optimization but not blocking this PR.

validators/chainsaw/allowlist.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// disallowedOperation returns the lowercase name of a disallowed operation
+// set on op, if any. Only Assert and Error are permitted. The chainsaw
+// Operation struct exposes each operation as a discriminated optional
+// pointer; we walk each known pointer and reject any non-allowlisted one
+// that is non-nil. Adding a new chainsaw operation upstream would leave
+// it implicitly allowed only if its pointer is also added here — which is
+// a deliberate fail-loud trigger when the dependency bumps.
+func disallowedOperation(op v1alpha1.Operation) (string, bool) {
```
yuanchen8911, line 147: The read-only allowlist is a denylist (`disallowedOperation` enumerates the bad ops), and it omits chainsaw's `proxy` operation — which exists in the pinned v0.2.15 (`vendor/github.com/kyverno/chainsaw/pkg/apis/v1alpha1/operation.go:88`, `Proxy *Proxy`). A registry-declared or inline `assertFile` with a `proxy:` step under `try:` returns `("", false)` from `disallowedOperation`, passes `ValidateTestReadOnly`, and is dispatched to chainsaw — which runs under the cluster-admin-bound ServiceAccount (`pkg/validator/job/rbac.go`). `proxy` issues an HTTP request through the API-server proxy subresource to a pod/service/node, i.e. a side-effecting op, not read-only. So the "only assert and error are permitted" guarantee is already broken today (not on a future dependency bump), and `allowlist_test.go` has no `proxy` coverage.

Suggested fix: add `case op.Proxy != nil: return "proxy", true` plus a regression test. Better, invert to a true allowlist — reject anything that is not exactly `Assert` or `Error` — so future upstream operations fail closed by default instead of fail-open.

validators/deployment/Dockerfile
```diff
@@ -13,6 +13,30 @@
...
+# raw-K8s-YAML library path and Chainsaw Test binary path dispatch in
+# validators/chainsaw/runner.go must stay aligned. See issue #1220.
+#
+# Single-arch: linux/amd64 only. Multi-arch support (linux/arm64) is a
+# follow-up that requires buildx adoption repo-wide; the .settings.yaml
+# already carries a linux_arm64 checksum so the upgrade path is clear.
+FROM debian:bookworm-slim AS chainsaw-fetch
+ARG CHAINSAW_VERSION
+ARG CHAINSAW_SHA256_LINUX_AMD64
+RUN test -n "${CHAINSAW_VERSION}" || (echo "CHAINSAW_VERSION must be set (Makefile injects from .settings.yaml)" && exit 1) && \
```
yuanchen8911: Shell operator precedence makes this guard print the wrong diagnostic. `A || B && C || D` parses as `((A||B) && C) || D`; when `CHAINSAW_VERSION` is empty, the first `|| (echo VERSION && exit 1)` subshell exits 1, `(A||B)` is false, `&& C` is skipped, and control falls to the trailing `|| (echo SHA256 && exit 1)` — printing the SHA256-missing message for a missing-VERSION failure. The build still fails closed (correct outcome); only the message is misleading. Split into two separate `RUN` lines, or use `if [ -z "$VAR" ]; then echo ...; exit 1; fi`.

yuanchen8911 (commented):
Three findings from a review of the chainsaw deployment-runner activation. The first is worth fixing before merge; the other two are low/cosmetic.

yuanchen8911 (changes requested):
Left comments. PTAL.

recipes/validators/catalog.yaml
```diff
@@ -38,7 +38,14 @@ validators:
...
     image: ghcr.io/nvidia/aicr-validators/deployment:latest
-    timeout: 5m
+    # Must be >= defaults.ChainsawAssertTimeout (6m) +
+    # defaults.JobEnvelopeMargin (60s) so chainsaw has headroom to
+    # terminate, clean up its temp dir, and flush log output before the
+    # Job's SIGKILL. The invariant is verified by
+    # TestExpectedResourcesCatalogEnvelope in pkg/validator/catalog —
+    # raising ChainsawAssertTimeout without raising this in tandem will
+    # fail the test. See issue #1220.
+    timeout: 7m
```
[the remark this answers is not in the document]
mchmarny: Applied in 601c9a8d. Bumped expected-resources timeout to 8m (= 6m chainsaw + 60s pre-chainsaw work headroom + 60s SIGKILL cleanup margin). Tightened TestExpectedResourcesCatalogEnvelope to assert strict-greater (`>` not `>=`) so an exact-match regression fails the test even if envelope arithmetic shifts later.

validators/deployment/Dockerfile
```diff
@@ -13,6 +13,30 @@
...
+    test -n "${CHAINSAW_SHA256_LINUX_AMD64}" || (echo "CHAINSAW_SHA256_LINUX_AMD64 must be set (Makefile injects from .settings.yaml)" && exit 1)
```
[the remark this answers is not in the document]
mchmarny: Applied in 601c9a8d. Dockerfile build-arg guard now rejects both empty and the literal string "null" (yq -r's default for missing keys). Failure message names the offending arg and points back to .settings.yaml.

validators/deployment/expected_resources.go
```diff
@@ -82,74 +82,56 @@ func checkExpectedResources(ctx *validators.Context) error {
...
+				fmt.Printf("  [chainsaw] %s: health check passed\n", r.Component)
+			} else {
+				msg := fmt.Sprintf("[chainsaw] %s: health check failed", r.Component)
+				if r.Output != "" {
+					msg += fmt.Sprintf(":\n%s", r.Output)
 				}
+				if r.Error != nil {
+					msg += fmt.Sprintf("\nerror: %v", r.Error)
+				}
+				failures = append(failures, msg)
```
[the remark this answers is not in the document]
mchmarny: Applied in 601c9a8d. `checkExpectedResources` now captures the first `*errors.StructuredError` from chainsaw result `r.Error` (e.g., `ErrCodeInvalidRequest` from `ValidateTestReadOnly` when registry assert content violates the read-only allowlist). On failure, the function prefers the structured error over the generic `ErrCodeNotFound` summary so downstream catalog/CLI surfaces classify the failure correctly. Human-readable failures list is still printed for operator visibility.

validators/chainsaw/allowlist.go
mchmarny: @yuanchen8911 — sharp catch, this was a real gap not a theoretical one. Fixed in 25af1bc3:

1. **Added `case op.Proxy \!= nil: return "proxy", true`** to the switch so the v0.2.15 proxy op is named explicitly with an actionable error.
2. **Inverted to true allowlist** as you suggested. After the switch, an Operation with neither `Assert` nor `Error` set falls through to `return "<unrecognized>", true` — so anything chainsaw upstream adds in a future release fails closed by default until a maintainer audits and enumerates it. No more denylist-shaped fail-open exposure.

New tests:
- `TestValidateTestReadOnly/proxy_rejected` — covers the v0.2.15 gap.
- `TestDisallowedOperation_AllowlistFailsClosed` — direct test that an all-zero `v1alpha1.Operation` is rejected with name `"<unrecognized>"`, plus positive cases for Assert-only and Error-only.

validators/deployment/Dockerfile
mchmarny: @yuanchen8911 — you're right that the original `A || B && C || D` chain would have printed the wrong diagnostic. That form was already replaced in 601c9a8d (one commit before your review landed) with two separate `if [ -z "$VAR" ] || [ "$VAR" = "null" ]; then echo ...; exit 1; fi` blocks chained with `&&`. Each block does its own `exit 1` with the correct arg-named message, so the precedence trap is gone:

```Dockerfile
RUN if [ -z "${CHAINSAW_VERSION}" ] || [ "${CHAINSAW_VERSION}" = "null" ]; then \\
      echo "CHAINSAW_VERSION must be set to a non-null value (Makefile injects from .settings.yaml)"; exit 1; \\
    fi && \\
    if [ -z "${CHAINSAW_SHA256_LINUX_AMD64}" ] || [ "${CHAINSAW_SHA256_LINUX_AMD64}" = "null" ]; then \\
      echo "CHAINSAW_SHA256_LINUX_AMD64 must be set to a non-null value (Makefile injects from .settings.yaml)"; exit 1; \\
    fi
```

That fix was in response to a separate CodeRabbit comment about `yq -r` emitting the literal string `"null"` — happens to also resolve your precedence concern. Confirm the current file looks right?

validators/chainsaw/binary_test.go
```diff
@@ -20,114 +20,43 @@ import (
...
-			t.Fatal("Available() = false, want true when fallback path is executable")
+		fakeCanonical := filepath.Join(t.TempDir(), "chainsaw")
+		withCanonicalPath(t, fakeCanonical)
+
+		bin := NewChainsawBinary().(*chainsawBinary)
+		if bin.binPath != fakeCanonical {
+			t.Errorf("binPath = %q, want %q (the canonical fallback)", bin.binPath, fakeCanonical)
 		}
 	})
 }
```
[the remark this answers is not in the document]
mchmarny: Applied in 4d802f67. Converted to a table-driven harness with a 2-case slice (`name`, `createStub`, `wantBinPath`). Adding more discovery cases (e.g., "PATH hit but stub isn't executable") is now a one-line entry. The closure-shaped `wantBinPath` lets cases name TempDir paths that aren't known until `t.Run` runs.

validators/deployment/Dockerfile
```diff
@@ -13,6 +13,38 @@
...
+# kyverno/chainsaw Go library in go.mod is bumped in lockstep — both the
+# raw-K8s-YAML library path and Chainsaw Test binary path dispatch in
+# validators/chainsaw/runner.go must stay aligned. See issue #1220.
+#
+# Single-arch: linux/amd64 only. Multi-arch support (linux/arm64) is a
+# follow-up that requires buildx adoption repo-wide; the .settings.yaml
+# already carries a linux_arm64 checksum so the upgrade path is clear.
+FROM debian:bookworm-slim AS chainsaw-fetch
+ARG CHAINSAW_VERSION
+ARG CHAINSAW_SHA256_LINUX_AMD64
```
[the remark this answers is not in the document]
mchmarny: Real bug, fixed in 4d802f67. Multi-arch via Docker TARGETARCH ARG (BuildKit auto-populates; default since ~2023).

- Dockerfile: chainsaw-fetch now declares ARG TARGETARCH plus ARG CHAINSAW_SHA256_LINUX_AMD64 and ARG CHAINSAW_SHA256_LINUX_ARM64. A `case TARGETARCH` block picks the matching SHA at build time; the curl URL also interpolates TARGETARCH. Unsupported arches fail loudly. Empty TARGETARCH (legacy non-BuildKit) is rejected with a clear DOCKER_BUILDKIT=1 hint.
- Makefile: image-validators now reads both linux_amd64 and linux_arm64 checksums from .settings.yaml and passes both as --build-arg. Dockerfile picks the right one via TARGETARCH.

Net: `make image-validators` on an M1 Mac produces a working linux/arm64 image; `docker buildx build --platform linux/amd64` on the same host produces a working linux/amd64 image. Both use .settings.yaml as the single source of truth.

pkg/defaults/timeouts.go
```diff
@@ -274,6 +274,17 @@ const (
...
+
+	// JobEnvelopeMargin is the headroom added on top of ChainsawAssertTimeout
+	// when computing the validator Job's outer activeDeadlineSeconds and the
+	// expected-resources catalog timeout. Chainsaw needs time after the inner
+	// assert deadline elapses to terminate the binary process, clean up the
+	// temp test directory, and flush log output before the Job's SIGKILL
+	// arrives. Without this headroom the binary is killed mid-cleanup and
+	// operators see truncated output, masking the actual failure cause.
+	// 60s mirrors the kubelet's default terminationGracePeriodSeconds; tune
+	// upward if chainsaw output truncation is observed in CI runs.
```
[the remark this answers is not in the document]
mchmarny: Good catch, the rationale was wrong. Fixed in 0ff057f6: the comment now says `60s = 30s for the default Pod terminationGracePeriodSeconds (the SIGTERM→SIGKILL window — see K8sPodTerminationWaitTimeout above) plus 30s headroom for pre-chainsaw helper.VerifyResource iteration and chainsaw startup variance.` Test message also updated to match. The constant value stays at 60s; only the rationale changed.

validators/deployment/expected_resources.go
```diff
@@ -78,78 +79,79 @@ func checkExpectedResources(ctx *validators.Context) error {
...
 	var chainsawAsserts []chainsaw.ComponentAssert
 	var failures []string
+	// firstStructuredErr captures the first structured error surfaced by
+	// chainsaw.Run results (e.g., ErrCodeInvalidRequest from
+	// ValidateTestReadOnly when a registry assert violates the read-only
+	// allowlist). Without this, the function would flatten such errors
+	// into the generic ErrCodeNotFound "expected resource check failed"
+	// summary at the bottom, losing the actionable classification. Per
+	// PR #1235 review.
+	var firstStructuredErr error
```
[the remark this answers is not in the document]
mchmarny: Fixed in 0ff057f6. `verifyGPUReadinessSignals` now returns `([]string, error)` — the first `*errors.StructuredError` it encounters across `verifyNodewrightReady`, `verifyClusterPolicyReady`, and `verifyDRAKubeletPluginReady`. The caller merges that into the existing `firstStructuredErr` fast-path so the final return preserves the original code (e.g., `ErrCodeInternal` from a discovery/RBAC failure) instead of collapsing into the generic `ErrCodeNotFound` summary. Refactored helper `capture` inside `verifyGPUReadinessSignals` walks each helper's err once for both the printable failures list and the structured-error fast-path.

validators/deployment/Dockerfile
```diff
@@ -13,6 +13,55 @@
...
+ARG TARGETARCH
+ARG CHAINSAW_VERSION
+ARG CHAINSAW_SHA256_LINUX_AMD64
+ARG CHAINSAW_SHA256_LINUX_ARM64
+# Reject empty / "null" build args. mikefarah/yq -r emits "null" for a
+# missing key, so explicit equality with "null" is required.
+RUN if [ -z "${TARGETARCH}" ]; then \
+      echo "TARGETARCH must be set (BuildKit auto-populates; for legacy docker build, pass --build-arg TARGETARCH=<amd64|arm64>)"; exit 1; \
+    fi && \
+    if [ -z "${CHAINSAW_VERSION}" ] || [ "${CHAINSAW_VERSION}" = "null" ]; then \
```
yuanchen8911, line 42: [P1] This new required `CHAINSAW_VERSION` guard is now wired in the main `on-push` and release validator-image jobs, but the same deployment Dockerfile is still built elsewhere without these args. `vuln-scan-images.yaml` passes only `GO_VERSION`, and the UAT AWS/GCP shell loops also pass only `--build-arg GO_VERSION` while iterating over `deployment performance conformance`. Those paths publish/build the deployment validator directly, not through `make image-validators`, so the deployment image will fail here before the scan/UAT can run. Please either pass `CHAINSAW_VERSION`, `CHAINSAW_SHA256_LINUX_AMD64`, and `CHAINSAW_SHA256_LINUX_ARM64` in those workflows too, or route those builds through the same Make target.

validators/chainsaw/allowlist.go
```diff
@@ -0,0 +1,161 @@
...
+// the operator can pinpoint the registry entry that violated the
+// allowlist. PR #1223 will surface the same rule at lint time so
+// violations are caught before they ever reach the validator.
+func ValidateTestReadOnly(component, yamlContent string) error {
+	var test v1alpha1.Test
+	if err := yaml.Unmarshal([]byte(yamlContent), &test); err != nil {
+		return errors.Wrap(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("failed to parse chainsaw Test YAML for component %q", component), err)
+	}
+	for stepIdx, step := range test.Spec.Steps {
```
yuanchen8911, line 102: [P1] This only walks `test.Spec.Steps`, but the vendored `v1alpha1.TestSpec` also has a top-level `Spec.Catch []CatchFinally` sibling. A Chainsaw Test can put `spec.catch: [{delete: ...}]` outside any step; that content is populated by YAML unmarshal, skipped by this loop, and then run by the deployment validator under its cluster-admin service account. Please validate `test.Spec.Catch` with the same read-only guard before returning nil.

validators/chainsaw/allowlist.go
```diff
@@ -0,0 +1,161 @@
...
+	case cf.PodLogs != nil:
+		return "podLogs", true
+	case cf.Script != nil:
+		return "script", true
+	case cf.Sleep != nil:
+		return "sleep", true
+	case cf.Wait != nil:
+		return "wait", true
+	}
+	return "", false
```
yuanchen8911: [P2] `CatchFinally` is still denylist-shaped here: any field not in the switch falls through as allowed. That is weaker than the fail-closed `disallowedOperation` helper above and leaves future Chainsaw `CatchFinally` operations open by default; it also currently allows `compiler` even though this type has no `assert`/`error` operation to allow. Since catch/finally/cleanup blocks are not part of the read-only assertion surface, invert this helper so any non-empty `CatchFinally` is rejected unless a field is explicitly audited and allowed.

recipes/validators/catalog.yaml
```diff
@@ -38,7 +38,17 @@ validators:
...
+    # defaults.JobEnvelopeMargin (60s) so chainsaw has headroom to
+    # terminate, clean up its temp dir, and flush log output before the
+    # Job's SIGKILL AND so pre-chainsaw work (helper.VerifyResource
+    # iteration) can't starve the chainsaw budget. 8m provides 60s for
+    # cleanup plus another 60s for pre-chainsaw work. The invariant is
+    # verified by TestExpectedResourcesCatalogEnvelope in
+    # pkg/validator/catalog — raising ChainsawAssertTimeout without
+    # raising this in tandem will fail the test. See issue #1220 and
+    # PR #1235 review.
+    timeout: 8m
```
yuanchen8911, line 51: [P3] This timeout is now `8m`, but `recipes/validators/README.md` still documents `expected-resources` as `5m` in the validator catalog table. Please update the README in the same PR so the catalog documentation matches the shipped timeout.

validators/chainsaw/allowlist.go
```diff
@@ -0,0 +1,161 @@
...
+// Caller contract: invoke only on content that IsChainsawTest reports as
+// Test format. Raw K8s YAML asserts have no operations and are unaffected.
+//
+// Returns ErrCodeInvalidRequest naming the offending step + operation so
+// the operator can pinpoint the registry entry that violated the
+// allowlist. PR #1223 will surface the same rule at lint time so
+// violations are caught before they ever reach the validator.
+func ValidateTestReadOnly(component, yamlContent string) error {
+	var test v1alpha1.Test
+	if err := yaml.Unmarshal([]byte(yamlContent), &test); err != nil {
```
yuanchen8911: [P3/non-blocking] This unmarshals a single `Test`, and the vendored YAML decoder documents that `Unmarshal` decodes only the first document in a multi-document YAML stream. Registry checks are conventionally one Test per file, so this is defense-in-depth rather than a merge blocker, but a second `---` document containing side-effecting operations would not be checked by this guard. Consider rejecting multi-doc Chainsaw Test files or decoding and validating every document.

yuanchen8911 (changes requested):
Posted inline findings verified against head 0ff057f62c18c247ea9b4a22abd9db8dec8ba689.

yuanchen8911 (changes requested):
Merge-blocking CI build-arg gaps remain, and the Chainsaw read-only guard still misses top-level/future catch surfaces. README timeout drift is minor.

yuanchen8911:
The three GPU check failures look like an infrastructure/runner flake.  Re-running the GPU jobs should clear it.

validators/deployment/Dockerfile
mchmarny: @yuanchen8911 — confirmed and fixed in 01091a09. Three more workflows were missing the chainsaw args:

- `.github/workflows/vuln-scan-images.yaml`: now uses the same `load-versions` composite action + 3 new build-args pattern as `on-push.yaml` / `on-tag.yaml`.
- `.github/workflows/uat-aws.yaml` and `.github/workflows/uat-gcp.yaml`: the shell-loop docker builds at lines 96-106 / 94-104 now read CHAINSAW_VERSION + both checksums via `yq` from `.settings.yaml` and pass them as `--build-arg` alongside `GO_VERSION`. Same approach as the Makefile's `image-validators` target.

The four-line block (`yq` × 3 + 3 extra `--build-arg`s) is duplicated across uat-aws/uat-gcp; consolidating into a composite action is a separate cleanup not in scope here. Other phase Dockerfiles ignore unknown build-args, so the unconditional pass-through is safe.

validators/chainsaw/allowlist.go
mchmarny: @yuanchen8911 — real bug, fixed in 01091a09. `validateTest` (extracted from the previously-monolithic `ValidateTestReadOnly`) now walks `test.Spec.Catch` BEFORE the per-step lists. Any entry in the top-level catch fails the read-only check with `spec.catch[N]` in the message so the operator can pinpoint which sibling block violates the contract. New test case `top-level spec.catch rejected` covers it.

validators/chainsaw/allowlist.go
mchmarny: @yuanchen8911 — applied in 01091a09. Inverted to fail-closed:

- The old `disallowedCatchFinally(cf) (string, bool)` helper became `catchFinallyOpName(cf) string` — names the offender for diagnostics only.
- Callers in `validateTest` no longer branch on a `bool`. ANY entry in `spec.catch`, `step.catch`, `step.finally`, or `step.cleanup` is rejected unconditionally — the rejection is structural (the for-loop body always returns an error on the first iteration). Compiler-only / description-only entries hit the generic `<catch-finally-block>` diagnostic from `catchFinallyOpName`.
- New test case `catch with only metadata still rejected (fail-closed)` exercises a description-only entry to prove the catch-all.

Future chainsaw additions to `CatchFinally` (new operation field, new metadata key) no longer need updates to this code to remain rejected.

recipes/validators/catalog.yaml
mchmarny: @yuanchen8911 — fixed in 01091a09. `recipes/validators/README.md` line 41 now shows `8m` and a brief note that the entry runs ExpectedResources + Chainsaw assert paths side-by-side.

validators/chainsaw/allowlist.go
mchmarny: @yuanchen8911 — applied in 01091a09 even though non-blocking. `ValidateTestReadOnly` now uses `yamlv3.NewDecoder` and iterates documents until `io.EOF`, validating each via `validateTest`. Errors include `doc[N]` so an offender in a multi-doc stream is pinpointable. Per-doc unmarshal goes through `sigs.k8s.io/yaml` (the chainsaw types are `json` tagged, not `yaml`) by re-marshaling the parsed yamlv3.Node and feeding it back through `yaml.Unmarshal`. New test case `second document in multi-doc YAML is rejected` covers the path.

validators/deployment/expected_resources.go
```diff
@@ -78,78 +79,88 @@ func checkExpectedResources(ctx *validators.Context) error {
...
 			if err := helper.VerifyResource(ctx.Ctx, ctx.Clientset, er); err != nil {
-				failures = append(failures, fmt.Sprintf("%s %s/%s (%s): %s",
+				failures = append(failures, fmt.Sprintf("[expectedResources] %s %s/%s (%s): %s",
 					er.Kind, er.Namespace, er.Name, ref.Name, err.Error()))
 			} else {
-				fmt.Printf("  %s %s/%s: healthy\n", er.Kind, er.Namespace, er.Name)
+				fmt.Printf("  [expectedResources] %s %s/%s: healthy\n", er.Kind, er.Namespace, er.Name)
 			}
 		}
 	}
```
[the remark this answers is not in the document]
mchmarny: Applied in 01091a09. Added two `select { case <-ctx.Ctx.Done(): return errors.Wrap(ErrCodeTimeout, ...) default: }` checks:

- At the top of the `for _, ref := range enabledRefs` loop, so a canceled run stops issuing more `helper.VerifyResource` calls between components.
- Inside the `if len(chainsawAsserts) > 0` block before `chainsaw.Run`, so a doomed run skips fetcher construction and the `slog.Info` noise.

`chainsaw.Run` itself already honors ctx mid-flight; the short-circuit here just avoids paying the dispatch setup cost on an already-canceled run.

mchmarny:
Tracking the GPU CI flake separately in #1237 — environmental, all 3 failing GPU jobs hit the same `nvidia-container-toolkit 1.19.1-1` + containerd-restart fingerprint inside the kind worker. Not in #1235's blast radius (failure occurs in `nvkind cluster create`'s internal apt-install on the worker, before any AICR validator code runs). Same trace appears on main since ~2026-06-07 19:00 UTC.

Live-cluster validation by @njhensley is the load-bearing GPU signal for this PR; the GPU-CI failures don't block that.

yuanchen8911 (dismissed):
All the issues are fully resolved. Thanks. /lgtm

The only remaining red checks are the three GPU tests (Inference/Smoke/Training). I re-triggered the failed jobs — hopefully they pass this time.

lalitadithya (approved):

Issue #1221, opened 2026-06-08
Backfill health checks: gke-nccl-tcpxo, dynamo-crds, kgateway-crds (#660 PR 3)


mchmarny:
## Summary

Backfill `healthCheck.assertFile` content for the three registry
components that don't yet have one, so the deployment phase's Chainsaw
coverage matches the registry once PR 2 ships the runner.

Depends on #660 PR 1 (hydration) + PR 2 (runtime).

### `gke-nccl-tcpxo`

- Predecessor #622 named `nccl-tcpxo-installer` and `device-injector`
  DaemonSets in `kube-system`; confirm these are the target before
  writing the check.
- Add `recipes/checks/gke-nccl-tcpxo/health-check.yaml` with read-only
  `assert` / `error` content only.
- Live-run against `h100-gke-cos-training` (or document why that
  environment was unavailable).

### `dynamo-crds`

- CRD-only component. The pinned chart is not vendored in-repo.
- Document the CRD-name extraction procedure (e.g., `helm template`
  against the pinned chart version) and pin the enumerated list of
  CRD names in the assert file.
- Assert exact CRD names show `Established=True`.

### `kgateway-crds`

- CRD-only component. Pinned chart:
  `oci://cr.kgateway.dev/kgateway-dev/charts/kgateway-crds:v2.0.0`
  (declared in `recipes/registry.yaml:400-408`).
- Per the comment in `recipes/components/kgateway-crds/values.yaml:15-22`,
  the chart installs six CRDs (`Backends`, `DirectResponses`,
  `GatewayExtensions`, `GatewayParameters`, `HTTPListenerPolicies`,
  `TrafficPolicies` — all under `gateway.kgateway.dev`).
- The Gateway API and Inference Extension CRDs under
  `recipes/components/kgateway-crds/manifests/` are prerequisites, not
  CRDs this chart creates. Scope of the check:
  - If the bundler installs both the chart and the vendored manifests
    under this component's install path, assert both sets.
  - If the vendored manifests are owned by a different component or by
    a separate bundle step, keep this check to the chart's six CRDs
    only and file a follow-up to stand up a prerequisite component
    (e.g., `gateway-api-crds` / `inference-extension-crds`) with its
    own health check.
- Re-verify the exact CRD set against the pinned upstream chart; if
  any names drift from the `values.yaml` comment, reconcile both in
  the same PR.

## Acceptance Criteria

[...]

PR #1244, opened 2026-06-09
feat(recipe): lint guard requiring healthCheck.assertFile + allowlist

### LINKED ISSUE #1223: Registry lint guard for healthCheck.assertFile + allowlist (#660 PR 5)

mchmarny:
## Summary

Final piece of epic #660. Adds the PR-time lint contract:

1. **Every registry entry must declare `healthCheck.assertFile`**, and
   the path must be readable through the embedded data provider.
2. **Every declared assertFile must pass the read-only allowlist**
   (`assert` / `error` operations only; no `script`, `apply`, `wait`,
   `command`, etc.).

Together with #1235's runtime allowlist enforcement, this means
violations are caught at PR review and never reach the validator
runtime.

Fixes: #1223
Closes epic #660 once merged.
Related: #1219, #1220, #1221.

## Type of Change

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe` — new test only)
- [x] Validator (`validators/chainsaw` — tightened existing test)
- [x] Docs (`docs/contributor/recipe.md`)

### New: `pkg/recipe.TestComponentRegistry_RequiresHealthCheck`

Walks the embedded registry; per-component subtest verifies:

- `healthCheck.assertFile` is non-empty
- The declared path resolves through the same data provider used by
  recipe-resolution hydration
  (`pkg/recipe/metadata_store.go:hydrateHealthCheckAsserts`)
- The file is non-empty

Failure messages name the offending component and the suggested
remediation (ship `recipes/checks/<name>/health-check.yaml` + cite
the allowlist contract).

### Tightened: `validators/chainsaw.TestValidateTestReadOnly_RegistryContent`

Was a `recipes/checks/*` directory walk; now walks via
`recipe.GetComponentRegistryFor(provider)`. This closes a path-
mismatch gap: a registry entry pointing at a non-conventional path
(anything other than `recipes/checks/<name>/health-check.yaml`) is
now validated, because runtime hydration honors whatever the registry
says.

Empty-assertFile and unreadable-path failure modes fall through here
gracefully — the `pkg/recipe` test above is the single source of
truth for those failures, so this test doesn't double-report on the
same root cause.

### Surface decision: lint-time, not load-time

The issue's open question was "lint at `make qualify` time vs.
load-time at `aicr recipe`". This PR uses lint-time. Two reasons
documented inline:

1. Catches violations at PR review, before any operator runs `aicr`.
2. External `--data` overlays may add components that aren't subject
   to this contract — only the in-tree `recipes/registry.yaml` is.
   The embedded registry is the source of truth this test asserts on.

### Docs

`docs/contributor/recipe.md`:

- Table row for `healthCheck.assertFile` flipped from optional (`no`)
  to required (**`yes`**) with a description that names both tests
  enforcing the contract.
- "Adding a Recipe" gains a new step (#3) describing the requirement
  for new components, including pointers to the chainsaw assertion
  patterns currently in use (DaemonSet `numberReady ==
  desiredNumberScheduled`, Deployment `Available=True`, CRD
  `Established=True`).

## Testing

```bash
go test -race -count=1 ./pkg/recipe/ ./validators/chainsaw/
golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./validators/chainsaw/...
```

- `TestComponentRegistry_RequiresHealthCheck`: 27/27 components
  pass.
- `TestValidateTestReadOnly_RegistryContent`: registry walker
  validates all 27 declared assertFiles pass the allowlist.
- All existing `pkg/recipe` and `validators/chainsaw` tests still
  pass.
- Lint clean.

**Regression-positive check:** to manually verify the new lint
guard works, remove any `healthCheck` block from a registry entry
locally and re-run `go test ./pkg/recipe/`. The subtest for that
component fails with a clear pointer to the missing assertFile +
the read-only contract.

## Risk Assessment

- [x] **Low** — Tests + docs only. No production code changes. No
  runtime behavior change. Easy to revert.

**Rollout notes:** No user-facing CLI/API change. After merge, any
PR that adds a registry entry without `healthCheck.assertFile` (or
points at a missing file, or ships a file with a disallowed
chainsaw op) fails `make qualify` with a per-component diagnostic.

## Checklist

- [x] I added/updated tests for new functionality (this PR is the
      test)
- [x] I updated docs for the new contract (`docs/contributor/recipe.md`)

### DIFF
--- a/docs/contributor/recipe.md
+++ b/docs/contributor/recipe.md
@@ -97,7 +97,7 @@ components:
 | `podScheduling` | `PodSchedulingConfig` | no | Helm value paths for workload pod scheduling injection |
 | `storageClassPaths` | []string | no | Helm value paths where `--storage-class` is written |
 | `validations` | []`ComponentValidationConfig` | no | Bundle-time validation checks (function, severity, conditions, message) |
-| `healthCheck.assertFile` | string | no | Chainsaw assert YAML (relative to data dir) for conformance phase |
+| `healthCheck.assertFile` | string | **yes** | Chainsaw assert YAML (relative to data dir) consumed by `aicr validate --phase deployment` (runtime — #1220) and by `make check-health` locally. Content is restricted to the read-only `assert` / `error` operation allowlist. Enforced at PR time by `pkg/recipe.TestComponentRegistry_RequiresHealthCheck` (every component must declare a path) and `validators/chainsaw.TestValidateTestReadOnly_RegistryContent` (every declared path must pass the allowlist) — see #1223. |
 | `gkeCriticalPriority` | bool | no | Synthesize ResourceQuota on GKE so `system-*-critical` pods admit |
 | `hasSelfRefCRDs` | bool | no | Tells helmfile to emit `disableValidation: true` (chart ships CRD + CR in same release) |
 
@@ -403,17 +403,31 @@ externally-visible product. Fields beyond `ComponentRefs` and
 2. **Write the YAML** in the correct directory. For an overlay, set
    `spec.base` to the most specific shared ancestor and let the chain
    carry shared constraints; only declare what differs.
-3. **Run `make bom-docs` and commit `docs/user/container-images.md`**
+3. **Ship the chainsaw health check** (registry entries only). Every
+   new component in `recipes/registry.yaml` MUST declare
+   `healthCheck.assertFile` pointing at
+   `recipes/checks/<name>/health-check.yaml`, and that file MUST use
+   only the read-only `assert` / `error` operation allowlist (no
+   `script`, `apply`, `wait`, `command`, etc. — see
+   `validators/chainsaw/allowlist.go`). The contract is enforced at
+   PR time by `pkg/recipe.TestComponentRegistry_RequiresHealthCheck`
+   and `validators/chainsaw.TestValidateTestReadOnly_RegistryContent`
+   — both gate `make qualify`. See #1223 and the
+   [chainsaw health check section in validator.md](validator.md#chainsaw-health-checks)
+   for the assertion patterns currently in use (DaemonSet
+   `numberReady == desiredNumberScheduled`, Deployment
+   `Available=True`, CRD `Established=True`).
+4. **Run `make bom-docs` and commit `docs/user/container-images.md`**
    if your change touches `registry.yaml`, a component's `values.yaml`,
    or a chart version pin (see [BOM regeneration](#bom-regeneration)).
-4. **Unit tests.** `make test` runs the recipe-resolution suite —
+5. **Unit tests.** `make test` runs the recipe-resolution suite —
    `pkg/recipe/yaml_test.go` (static catalog: parse, refs, enum
    values, inheritance depth, no cycles) and
    `pkg/recipe/metadata_test.go` (runtime merge, topological sort).
    Both gate `make qualify`. If your change adds a registry entry, a
    new overlay file, or a mixin, the static suite typically picks it
    up without new test code.
-5. **Integration validation.** For a new chart pin, run `make qualify`
+6. **Integration validation.** For a new chart pin, run `make qualify`
    and let the e2e pipeline render the bundle. KWOK simulated
    clusters (`make kwok-e2e RECIPE=<name>`) catch most resolution
    regressions without GPU hardware.
--- a/pkg/recipe/components_test.go
+++ b/pkg/recipe/components_test.go
@@ -15,6 +15,7 @@
 package recipe
 
 import (
+	"context"
 	"maps"
 	"slices"
 	"strings"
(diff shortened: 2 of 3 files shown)


ArangoGutierrez (approved):
## ✅ Approve

Reviewed statically against the PR head (`a981918`); did not execute the suite locally. The tests are genuine (not theater), and the contract is well-constructed. Nice close on the registry side of #660.

### Verified
- **Anti-vacuous guard present** — [`TestValidateTestReadOnly_RegistryContent`](https://github.com/NVIDIA/aicr/blob/a98191814d4b38fb110f6f9e9a44029f422277ae/validators/chainsaw/allowlist_test.go) ends with `if checked == 0 { t.Fatal("...registry walker is broken") }`, so the registry walker can't pass over an empty set.
- **Compiles against base** — `validators/chainsaw/allowlist.go`, `GetComponentRegistryFor`, and `defaultEmbeddedProvider` are all present on the base commit (`cad0142`).
- **No nil-panic** — `HealthCheck` is a value type (`HealthCheckConfig`), so the unguarded `comp.HealthCheck.AssertFile` access is safe.
- **Docs anchor valid** — the new `validator.md#chainsaw-health-checks` link matches the existing `## Chainsaw health checks` heading. Worth calling out since lychee runs in CI but not in `make qualify`.

### Strengths
- Both tests fail when the contract is broken (missing/empty/unreadable `assertFile` → recipe test; disallowed chainsaw op → allowlist test). Per-component subtests plus remediation pointers in [`TestComponentRegistry_RequiresHealthCheck`](https://github.com/NVIDIA/aicr/blob/a98191814d4b38fb110f6f9e9a44029f422277ae/pkg/recipe/components_test.go) are exactly right.
- Retargeting the chainsaw test from a `recipes/checks/*` filesystem walk to a **registry walk** is a real correctness improvement — it now validates whatever path runtime hydration would actually honor, not just the conventional `recipes/checks/<name>/` layout.

### Minor (non-blocking)
1. **Implicit cross-test coupling.** The chainsaw test `continue`s (silently skips) on empty/unreadable `assertFile`, delegating those failure modes to `TestComponentRegistry_RequiresHealthCheck`. Correct under `make qualify` (runs `./...`), but if the recipe test is ever removed, the chainsaw test would silently skip broken entries and still go green (as long as ≥1 valid entry keeps `checked > 0`). A one-line guard comment on the recipe test noting the dependency would harden this against future erosion.
2. **Provider-prefix nit.** The chainsaw test builds its provider with prefix `"."` while the canonical `defaultEmbeddedProvider` uses `""`. Harmless (`filepath.Join(".", path)` cleans the `.`), but passing `""` would match the runtime path exactly and remove the head-scratch.
3. **Readability ≠ hydration.** The recipe test asserts the path *reads* through the provider, framed as exercising the hydration path (`hydrateHealthCheckAsserts`) — it won't catch a file that reads fine but fails downstream parsing/transform. Fine for a lint's scope.

Ready to merge once checks go green (`BLOCKED` is pending checks/review, not conflicts).

ArangoGutierrez (approved):

PR #1243, opened 2026-06-09
feat(recipes): backfill chainsaw health checks for 5 missing components

mchmarny:
## Summary

Backfill `healthCheck.assertFile` for the 5 registry components that
don't have one. After this PR lands, every current component has a
chainsaw health check — the precondition for #1223's lint guard.

Fixes: #1221
Related: #660 (epic), #1223

## Type of Change

- [x] New feature (non-breaking change that adds functionality)

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe` — registry.yaml + new
      recipes/checks/* health-check files)

### Scope drift from #1221's original text

#1221 named `gke-nccl-tcpxo`, `dynamo-crds`, and `kgateway-crds`. The
registry has drifted since the issue was written: `dynamo-crds` and
`kgateway-crds` aren't in the current `recipes/registry.yaml` (they
were renamed or folded into other components). The actual 5 current
gaps are covered here. The acceptance criterion is "every current
registry component has a `healthCheck.assertFile`" — sticking to the
dated list would leave 4 gaps and prevent #1223 from enforcing.

Per out-of-band agreement, the issue body is not updated; the scope
drift is captured here in the PR description instead.

### The 5 components

| Component | Type | Assertions |
|---|---|---|
| `gke-nccl-tcpxo` | Manifest-only | 2 DaemonSets in `kube-system` (`nccl-tcpxo-installer`, `device-injector`) — `numberReady == desiredNumberScheduled` |
| `grove` | Helm chart | `grove-operator` Deployment in `dynamo-system` — `Available=True` + Pod-phase/container-state error guards |
| `agentgateway-crds` | CRD-only | 10 vendored Gateway API + Inference Extension CRDs — `Established=True` |
| `prometheus-operator-crds` | CRD-only | 8 `monitoring.coreos.com/v1` CRDs — `Established=True` |
| `slinky-slurm-operator-crds` | CRD-only | 6 `slinky.slurm.net` CRDs — `Established=True` |

### Resource-name provenance

- **`gke-nccl-tcpxo`** — DaemonSet names verified from in-tree
  `recipes/components/gke-nccl-tcpxo/manifests/{nccl-tcpxo-installer,nri-device-injector}.yaml`.
- **`grove`** — Deployment name + namespace verified from the existing
  `tests/chainsaw/ai-conformance/kind-inference-dynamo/assert-dynamo.yaml`
  assertion (so the registry check produces the same signal the
  ai-conformance suite exercises).
[...]

Issue #1236, opened 2026-06-08
Replace chainsaw binary with in-process executor for assert/error allowlist subset


mchmarny:
## Summary

Investigate replacing the chainsaw CLI binary in the deployment
validator image with an in-process Go executor for the
`assert`/`error` allowlist subset. The goal is a pure-Go validator
image built by `ko` (no Dockerfile, no GitHub-release fetch, no
sha256 dance, no `.settings.yaml` build-arg plumbing), with the
read-only allowlist becoming **structural** rather than a YAML
validator.

## Background

PR #1235 (closed #1220) ships the chainsaw binary in the deployment
validator image via a multi-stage Dockerfile that downloads + checksums
the upstream release tarball. Two dispatch paths exist in
`validators/chainsaw/runner.go`:

1. **Raw K8s YAML asserts** → `assertRawResources` →
   `github.com/kyverno/chainsaw/pkg/engine/checks` (in-process Go
   library). No binary needed; `ko` would handle this path today.
2. **Chainsaw `Test` format** (`apiVersion: chainsaw.kyverno.io/v1alpha1`,
   `kind: Test`) → `runChainsawBinary` → external chainsaw CLI. This
   is what the Dockerfile multi-stage exists for.

After #1235 lands, registry-declared content is restricted to a
read-only allowlist (`assert` and `error` operations only). That means
the in-process executor would only need to drive those two ops, not
the full chainsaw operation set (`script`, `apply`, `wait`, `catch`,
`finally`, ...). Significantly smaller surface than reimplementing
chainsaw.

## Proposal

Write a `runChainsawTestInProcess` shim that:

1. Unmarshals the Test YAML to `v1alpha1.Test` (already done by
   `ValidateTestReadOnly` in `validators/chainsaw/allowlist.go`).
2. Walks `test.Spec.Steps[].Try[]`, dispatching each `assert` /
   `error` to
   `github.com/kyverno/chainsaw/pkg/engine/checks.Check` (the same
   library `assertRawResources` already uses).
3. Applies a retry loop with `defaults.AssertRetryInterval` until the
   step passes or the per-step timeout from `defaults.ChainsawAssertTimeout`
   elapses.
4. Returns the same `chainsaw.Result` shape as `runChainsawBinary`.

## Acceptance Criteria

- New `runChainsawTestInProcess` function in `validators/chainsaw/`
  with parity tests covering every registry check (use the existing
  `TestValidateTestReadOnly_RegistryContent` corpus as the source).
[...]

PR #1245, opened 2026-06-09
feat(recipes): deepen 21 chainsaw health checks; close epic #660

### LINKED ISSUE #1222: Enhance existing Chainsaw health checks (#660 PR 4)
## Summary

Audit and enhance the existing 19 Chainsaw health checks under
`recipes/checks/*` so they are deep enough to serve as post-install
readiness, not just existence checks.

Depends on #660 PR 1 + PR 2. Best landed after #660 PR 3 (backfill) so
all 22 components are covered before the audit.

## Scope

For each existing check, audit and enhance where the current assertion
is too shallow. Targets include (non-exhaustive):

- **Workload-level readiness:** verify
  `Deployment.availableReplicas >= desired`,
  `StatefulSet.readyReplicas >= desired`,
  `DaemonSet` (`desiredNumberScheduled > 0` and
  `numberReady == desiredNumberScheduled`) where checks currently only
  assert object existence.
- **CRD `Established=True`:** verify the condition where the component
  owns CRDs.
- **Selector tightening in shared namespaces:** in `kube-system`,
  `monitoring`, and other shared namespaces, tighten label selectors so
  checks don't false-match sibling components.
- **Component-specific readiness:** add or strengthen custom assertions
  the component genuinely requires
  (e.g., `ClusterPolicy.status.state == "ready"`, Skyhook customization
  completion, DRA kubelet-plugin readiness).

## Constraints

- Read-only `assert` / `error` allowlist only — no `script`, `wait`,
  `catch`, `apply`, `create`, `delete`, `patch`, `update`, ... (per
  #660 PR 2 enforcement).
- Deployer-neutral: no release-scoped labels
  (`app.kubernetes.io/instance`), no Helm API calls, no
  release-metadata reads.

## Acceptance Criteria

- Per-component audit summary in the PR description identifying which
  checks were shallow and what was strengthened.
- Workload-level readiness assertions added where missing.
- CRD `Established=True` assertions added where the component owns
  CRDs.
- Shared-namespace selectors are tight enough to avoid sibling-component
  false matches.
- Live-cluster validation run against
  `h100-eks-ubuntu-inference-dynamo` confirms every enhanced check
  passes; unexercised mapped components (e.g., `aws-efa`,
  `network-operator`, `gke-nccl-tcpxo`, `kubeflow-trainer`) documented.

## Related

- Parent epic: #660
- Depends on: #660 PR 1, #660 PR 2; best after #660 PR 3

### LINKED ISSUE #1248: fix(recipes): NFD worker-DaemonSet check asserts omitempty numberUnavailable (fails on healthy cluster)
## Summary

`recipes/checks/nfd/health-check.yaml` gates the NFD worker DaemonSet rollout on a `numberUnavailable == 0` assertion. `DaemonSetStatus.NumberUnavailable` is `omitempty`, so on a fully-healthy DaemonSet (0 unavailable) the field is **absent** from the serialized object. Chainsaw/kyverno-json then evaluates that comparison as `null == 0` → **false**, the assert fails, and the NFD health check fails on an otherwise-healthy cluster.

Discovered during review of #1245, which deliberately avoided this pattern for its `aws-efa` / `nvidia-dra-driver-gpu` DaemonSet checks.

## Evidence

- `recipes/checks/nfd/health-check.yaml` — the `validate-worker-daemonset` step asserts that `numberUnavailable` equals 0 (see the Fix section for the exact chainsaw expression).
- `vendor/k8s.io/api/apps/v1/types.go:722` declares `NumberUnavailable` with a `json:"numberUnavailable,omitempty"` struct tag — the `omitempty` is the trap. By contrast `DesiredNumberScheduled` (L698) and `NumberReady` (L702) have no `omitempty`, so they are always serialized.
- kyverno-json scalar matching: a missing field projects to `null`, and `null == 0` is `false`.

## Impact

The NFD worker-DaemonSet rollout gate is structurally unsatisfiable when the DaemonSet is healthy — the field it tests is omitted exactly when the condition it wants (`== 0`) is true. The master/GC Deployment asserts, the phase checks, and the container-state checks are unaffected; this is isolated to the worker-DaemonSet rollout assertion.

## Fix

Match the corrected pattern #1245 uses for its two DaemonSets (`numberReady` and `desiredNumberScheduled` are always serialized, so this is robust):

```yaml
- assert:
    resource:
      apiVersion: apps/v1
      kind: DaemonSet
      metadata:
        name: node-feature-discovery-worker
        namespace: node-feature-discovery
      status:
        # Guard against vacuous pass when 0 nodes match the DaemonSet.
        (desiredNumberScheduled > `0`): true
        # Full rollout: every scheduled pod is ready.
        (numberReady == desiredNumberScheduled): true
```

## Related

- Introduced in: #1234
- Correct pattern reference: #1245 (`aws-efa`, `nvidia-dra-driver-gpu`)
- Epic: #660 · original deepening issue: #1222

## Note

#1245 carries comments in `aws-efa` / `nvidia-dra-driver-gpu` that cite this NFD file as "same rationale" — those cross-references are flagged for correction in that PR (NFD uses a different, broken mechanism).

mchmarny:
## Summary

Closes #1222. With this merged, **epic #660 is fully done.**

Audit + deepen 21 chainsaw health checks that landed before the epic
hit "every check is deep" depth. The PR-time runtime contract from
#1219 / #1220 / #1223 is in place; this PR closes the
depth-of-coverage gap on the actual assertions inside each
`recipes/checks/*/health-check.yaml`.

Fixes: #1222
Closes: #1248 (NFD `numberUnavailable` omitempty bug — the core fix was
landed pre-PR via #1235's `d7e68a24` patch; this PR's broader NFD
touch reinforces with the projection-form container-state fix
below).
Closes epic: #660

## Post-merge review fixes (@ArangoGutierrez)

After the initial 21-file enhancement landed, a follow-up review
surfaced a load-bearing correctness gap and several smaller items.
All three are now addressed in this PR:

### Multi-container projection fix (was: single-element silently passes sidecars)

The pre-fix container-state error blocks used single-element list
match:

```yaml
status:
  containerStatuses:
    - state: { waiting: { reason: CrashLoopBackOff } }
```

kyverno-json's `sliceNode.Assert`
(`vendor/github.com/kyverno/kyverno-json/pkg/core/assertion/assertion.go:43`)
gates on exact slice length BEFORE comparing elements, so this only
matches pods with exactly one container. Multi-container pods
(sidecar / proxy-injected / init+main) silently passed.

**Fixed via JMESPath projection** across all 22 files (21 in this PR
+ NFD's pre-existing container-state blocks + grove's):

```yaml
status:
  (containerStatuses[?state.waiting.reason == 'CrashLoopBackOff' || ... ] | length(@) > `0`): true
```

Now fires on any container at any index. Each file also gains a
**parallel `initContainerStatuses` block** — init-container
failures (e.g., `nvidia-operator-validator`'s 4 init containers)
are invisible to `containerStatuses` and were silently missed by
the prior pattern. Inline rationale comment added to every block.

### NFD cross-reference clarification

The aws-efa and nvidia-dra-driver-gpu DaemonSet-fix headers cite
NFD as the exemplar for the `numberReady == desiredNumberScheduled`
pattern. ArangoGutierrez's review note flagged this as misleading
because NFD's pre-#1235 state used `numberUnavailable == 0` (the
omitempty trap that #1248 calls out). NFD on this branch (post-#1235's
`d7e68a24` patch from @njhensley) uses the correct
`numberReady == desiredNumberScheduled` pattern — verified in
`recipes/checks/nfd/health-check.yaml`. The cross-reference is
accurate as of the current head. #1248's core ask is therefore
already resolved in main; this PR additionally fixes NFD's
container-state blocks via the projection form above.

### Empty-prefix DataProvider consistency (follow-up)

`validators/chainsaw/allowlist_test.go` now uses `""` (matching the
canonical `defaultEmbeddedProvider`); `pkg/recipe/provider_test.go`
still uses `"."` at multiple sites — same behavior (filepath.Join
cleans both) but mixed style. Out of scope for this PR; tracked
separately if it becomes a maintenance issue.

## Audit

| Component | Pre-PR depth | Action |
|---|---|---|
| `agentgateway-crds` | Full 10 CRDs Established=True | skip (#1243) |
| `agentgateway` | Deploy + pod-phase | +container-state |
| `aws-ebs-csi-driver` | Deploy + label pod-phase | +container-state (label-scoped) |
| `aws-efa` | DS `numberReady > 0` PARTIAL + label pod-phase | **fix to full rollout** + container-state |
| `cert-manager` | 1 Deploy + pod-phase | **+ 2 missing Deploys + 3 CRDs** + container-state |
| `dynamo-platform` | Deploy + pod-phase | + DCD CRD + container-state |
| `gatekeeper` | 2 Deploys + pod-phase | +container-state |
| `gke-nccl-tcpxo` | Full DS rollout | skip (#1243) |
| `gpu-operator` | Deploy + pod-phase | **+ ClusterPolicy CRD + cluster-policy CR `state=ready`** + container-state |
| `grove` | Deploy Available + 1 container-state | +3 missing container-state |
| `k8s-ephemeral-storage-metrics` | Deploy + label pod-phase | +container-state (label-scoped) |
| `k8s-nim-operator` | Deploy + pod-phase | +container-state |
| `kai-scheduler` | Deploy + pod-phase | +container-state |
| `kube-prometheus-stack` | Deploy + label pod-phase | +container-state (label-scoped) |
| `kubeflow-trainer` | Deploy + pod-phase | +container-state |
| `kueue` | Deploy + pod-phase | +container-state |
| `network-operator` | Deploy + pod-phase | +container-state |
| `nfd` | Full | skip (#1234) |
| `nodewright-customizations` | Skyhook CR `status=complete` | skip (deep CR state is the right signal) |
| `nodewright-operator` | Deploy + pod-phase | **+ Skyhook CRD** + container-state |
| `nvidia-dra-driver-gpu` | DS `numberReady > 0` PARTIAL + pod-phase | **fix to full rollout** + container-state |
| `nvsentinel` | Deploy + pod-phase | +container-state |
| `prometheus-adapter` | Deploy + label pod-phase | +container-state (label-scoped) |
| `prometheus-operator-crds` | Full 8 CRDs Established=True | skip (#1243) |
| `slinky-slurm` | 4 CRs + 3 Deploys + StatefulSet + NodeSet + pod-phase | +container-state |
| `slinky-slurm-operator` | 2 Deploys + pod-phase | +container-state |
| `slinky-slurm-operator-crds` | Full 6 CRDs Established=True | skip (#1243) |

Files touched: **21** (6 untouched: 3 CRD-only that are already deep,
+ NFD/gke-nccl-tcpxo/nodewright-customizations that already have the
right signal).

### 1. DaemonSet partial → full rollout (2 files)

`aws-efa` and `nvidia-dra-driver-gpu` were gating on `numberReady > 0`,
which passed on partial failures. Switched to `numberReady ==
desiredNumberScheduled` with `desiredNumberScheduled > 0` as the
vacuous-pass guard — same pattern as the NFD fix in #1234 and the new
`gke-nccl-tcpxo` in #1243. (We don't use `numberUnavailable == 0`
because that field is `omitempty` and disappears when zero — `null
== 0` evaluates false on a fully-healthy DaemonSet.)

### 2. Owned-CRD `Established=True` assertions (4 components)

| Component | CRDs asserted |
|---|---|
| `gpu-operator` | `clusterpolicies.nvidia.com` |
| `dynamo-platform` | `dynamocomponentdeployments.nvidia.com` |
| `nodewright-operator` | `skyhooks.skyhook.nvidia.com` |
| `cert-manager` | `certificates.cert-manager.io`, `issuers.cert-manager.io`, `clusterissuers.cert-manager.io` |

CRD names verified against
`tests/chainsaw/ai-conformance/cluster/assert-crds.yaml` and the Go
`verifyClusterPolicyReady` check.

### 3. gpu-operator: ClusterPolicy `state=ready` migration

`gpu-operator/health-check.yaml` now asserts the singleton
`cluster-policy` CR has `status.state == "ready"`. **This migrates the
`verifyClusterPolicyReady` Go check that PR #1235 task 7 deferred for
assertion-equivalence proof.** The static Chainsaw assertion is now
the canonical readiness signal for the full GPU stack (driver +
toolkit + cuda + plugin DaemonSets + validators all healthy →
operator reconciles ClusterPolicy.status.state to "ready").

### 4. cert-manager scope fix (real bug)

The chart deploys three Deployments: `cert-manager` (controller),
`cert-manager-cainjector`, `cert-manager-webhook`. The previous
check asserted only the controller. The webhook in particular is
load-bearing for admission: a healthy controller with a dead webhook
silently passes the check while the cluster rejects every
Certificate / Issuer CR. Now asserts all three.

### 5. Universal container-state error coverage (21 files)

Every check that previously only gated on pod phase (Pending /
Failed / Unknown) now also rejects pods stuck in Running phase with
a known-bad container-waiting reason:

- `CrashLoopBackOff`
- `ImagePullBackOff`
- `ErrImagePull`
- `CreateContainerConfigError`

The phase-only filter missed `CrashLoopBackOff` entirely because such
pods typically remain in Running phase with unhealthy containers.
Same pattern NFD adopted via #1234. Label-scoped in shared
namespaces (kube-system, monitoring) to avoid sibling false-matches;
namespace-only otherwise.

## Also Addresses Post-Merge Review on PR #1244

`validators/chainsaw/allowlist_test.go`:

- Provider prefix changed from `"."` to `""` to match the canonical
  `defaultEmbeddedProvider` used at runtime. No behavior change
  (`filepath.Join` cleans both forms), but matches the runtime path
  exactly without the head-scratch.
- Added a CROSS-TEST DEPENDENCY comment block noting that
  empty-assertFile and unreadable-path failure modes are owned by
  `pkg/recipe.TestComponentRegistry_RequiresHealthCheck`. If that
  test is ever removed or weakened, this test must be strengthened
  in lockstep so a broken registry can't silently green-light
  here.

## Out of Scope

Deferred to a possible follow-up:

- **CRD `Established=True` for 6 more operators that own CRDs**:
  `gatekeeper`, `kueue`, `kubeflow-trainer`, `network-operator`,
  `k8s-nim-operator`, `slinky-slurm-operator`. Each needs
  verification against its pinned chart version; bundling them
  blindly would risk introducing typo'd CRD names that fail on live
  clusters.
- **StatefulSet readiness for kube-prometheus-stack** (Prometheus /
  Alertmanager StatefulSets shipped by the stack): chart's exact
  StatefulSet names and label conventions need verification.

## Testing

```bash
go test -race -count=1 ./pkg/recipe/ ./validators/chainsaw/ ./validators/deployment/ ./pkg/defaults/ ./pkg/validator/catalog/
golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./validators/...
```

- `TestComponentRegistry_RequiresHealthCheck` (pkg/recipe): 27/27
  components pass — every entry declares assertFile + path resolves.
- `TestValidateTestReadOnly_RegistryContent` (validators/chainsaw):
  walks every registry-declared assertFile and validates against the
  read-only allowlist. 27/27 pass.
- All in-tree tests still pass.
- Lint clean.

**Live-cluster validation deferred to @mchmarny per the same
agreement as #1235 / #1243.** Priority verification targets for
this PR specifically:

1. **gpu-operator's `ClusterPolicy.status.state == "ready"`** — the
   load-bearing assertion that migrates the Go check. If it
   diverges from the Go check's behavior on real clusters, the
   migration needs adjusting.
2. **cert-manager's 3 Deployments** — confirm `cert-manager-cainjector`
   and `cert-manager-webhook` are the actual names produced by the
   chart's fullname template (vs. release-prefixed variants).
3. **The 4 owned-CRD assertions** — confirm each CRD name exists
   exactly as asserted on a live cluster.

## Risk Assessment

- [x] **Medium** — 21 file changes; semantically the depth bump is a
  real correctness improvement; isolated per-file revert is trivial.

**Rollout notes:** After merge, the deployment-phase chainsaw runner
exercises the deeper assertions on every `aicr validate --phase
deployment` run. A component that previously passed the existence
check but had a partially-healthy DaemonSet (or a CRD that isn't
Established yet, or a webhook Deployment down) now correctly
reports the failure. The fail-loud direction is the goal — see
"Live-cluster validation" above for the verification targets.

## Checklist

- [x] No new tests added — the existing two lint guards
      (`TestComponentRegistry_RequiresHealthCheck` +
      `TestValidateTestReadOnly_RegistryContent`) already cover the
      validation; this PR is the registry content they validate
- [x] No user-facing CLI/API behavior change


recipes/checks/kueue/health-check.yaml
```diff
@@ -68,3 +68,47 @@ spec:
...
+            resource:
+              apiVersion: v1
+              kind: Pod
+              metadata:
+                namespace: kueue-system
+              status:
+                containerStatuses:
+                  - state:
+                      waiting:
+                        reason: CreateContainerConfigError
```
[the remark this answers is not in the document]
mchmarny: Good catch, root cause was that my bash-append helper for the universal container-state errors didn't update the pre-existing pod-phase comment. Fixed in 7e13be4e. Applied universally across all 14 files that had the same stale phase-only narrative (not just the 2 you flagged — same root cause): kueue, network-operator, agentgateway, aws-ebs-csi-driver, gatekeeper, k8s-ephemeral-storage-metrics, k8s-nim-operator, kai-scheduler, kube-prometheus-stack, kubeflow-trainer, nvsentinel, prometheus-adapter, slinky-slurm, slinky-slurm-operator. Each comment now names both the phase set (Pending/Failed/Unknown) and the container-waiting reasons (CrashLoopBackOff/ImagePullBackOff/ErrImagePull/CreateContainerConfigError) plus why phase-only filtering misses CrashLoopBackOff. Label-scoping note preserved where the namespace is shared.

recipes/checks/network-operator/health-check.yaml
```diff
@@ -68,3 +68,47 @@ spec:
...
+            resource:
+              apiVersion: v1
+              kind: Pod
+              metadata:
+                namespace: nvidia-network-operator
+              status:
+                containerStatuses:
+                  - state:
+                      waiting:
+                        reason: CreateContainerConfigError
```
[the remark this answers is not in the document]
mchmarny: Good catch, root cause was that my bash-append helper for the universal container-state errors didn't update the pre-existing pod-phase comment. Fixed in 7e13be4e. Applied universally across all 14 files that had the same stale phase-only narrative (not just the 2 you flagged — same root cause): kueue, network-operator, agentgateway, aws-ebs-csi-driver, gatekeeper, k8s-ephemeral-storage-metrics, k8s-nim-operator, kai-scheduler, kube-prometheus-stack, kubeflow-trainer, nvsentinel, prometheus-adapter, slinky-slurm, slinky-slurm-operator. Each comment now names both the phase set (Pending/Failed/Unknown) and the container-waiting reasons (CrashLoopBackOff/ImagePullBackOff/ErrImagePull/CreateContainerConfigError) plus why phase-only filtering misses CrashLoopBackOff. Label-scoping note preserved where the namespace is shared.

ArangoGutierrez (commented):
## Review summary — deepen 21 chainsaw health checks

Verified against the vendored kyverno-json engine, the `verifyClusterPolicyReady` Go check, and `tests/chainsaw/ai-conformance/cluster/assert-crds.yaml` — **not** a live cluster. Net: real correctness wins, with one high-severity gap in the headline feature. Detail is in the inline comments.

### 🔴 Container-state checks only fire on single-container pods (inline on `gpu-operator`)
The literal one-element `containerStatuses:` assertion is matched by an exact-length gate (`sliceNode.Assert` → "lengths of slices don't match", `vendor/.../kyverno-json/pkg/core/assertion/assertion.go:43`). Any pod with ≥2 containers (device-plugin + `config-manager` sidecar, prometheus + config-reloader, injected proxies) **silently passes**. Affects all 21 files. Fix: the JMESPath projection form `(containerStatuses[?...] | length(@) > `0`): true` — the same style already used correctly for the CRD `conditions[?type=='Established']` asserts in this PR. No false-positive risk, but it's a silent false-negative that a passing live run won't surface.

### 🟡 Misleading NFD cross-reference + NFD #1234 likely broken (inline on `aws-efa`)
The DaemonSet fix here is correct (`numberReady`/`desiredNumberScheduled` are not `omitempty`). But the comments cite `nfd/health-check.yaml` as "same rationale," and NFD uses `numberUnavailable == 0` — which IS `omitempty`, so `null == 0` is false on a healthy DS. Filing a separate NFD issue. Please also fix the cross-ref comment here and in `nvidia-dra-driver-gpu`.

### ✅ Solid (verified)
- All 4 owned-CRD names match `assert-crds.yaml` exactly.
- ClusterPolicy `state=ready` is equivalent to `verifyClusterPolicyReady` (`status.state == "ready"`, name `cluster-policy`). Right call to flag as live-validation priority #1.
- cert-manager 3-Deployment scope fix — genuine bug catch (the webhook is admission-gating).
- Go test provider-prefix change correctly matches the runtime `globalDataProvider` (empty prefix).

### Verdict
Recommend addressing the 🔴 before merge (it's the PR's central feature) and the 🟡 comment fix. The CRD / ClusterPolicy / cert-manager / test changes are good as-is. Posted as a **comment**, not a formal block — escalate at your discretion.

recipes/checks/gpu-operator/health-check.yaml
```diff
@@ -71,3 +101,51 @@ spec:
...
+        # misses CrashLoopBackOff/ImagePullBackOff because those pods
+        # typically remain in Running phase with unhealthy containers.
+        - error:
+            resource:
+              apiVersion: v1
+              kind: Pod
+              metadata:
+                namespace: gpu-operator
+              status:
+                containerStatuses:
```
ArangoGutierrez: **🔴 This container-state check only fires on single-container pods — and the same pattern is repeated in all 21 files in this PR.**

Chainsaw compiles this `error` block via kyverno-json, whose `sliceNode.Assert` gates on exact slice length *before* comparing any element (`vendor/github.com/kyverno/kyverno-json/pkg/core/assertion/assertion.go:43`):

```go
if valueOf.Len() \!= len(node) {
    errs = append(errs, field.Invalid(path, value, "lengths of slices don't match"))
}
```

So a one-element `containerStatuses:` matches only a pod with **exactly one** container:

| Pod | Result |
|---|---|
| 1 container, CrashLoopBackOff | length 1==1 → flagged ✓ |
| 2+ containers, one crashlooping | length ≠ 1 → no match → **silently passes** ✗ |

This namespace is exactly where it bites: the device-plugin DaemonSet pods carry a `config-manager` sidecar and dcgm-exporter often has init+main. Same gap in every namespace-scoped check (cert-manager, dynamo-system, kueue-system, …) and any sidecar/proxy-injected pod. No false-positive risk, but it's a silent false-negative — the opposite of the "fail loud" goal.

Fix — use the projection form (same style as the CRD `conditions[?type=='Established']` asserts in this PR). It fires on any container at any index and collapses the four blocks into one:

```yaml
- error:
    resource:
      apiVersion: v1
      kind: Pod
      metadata:
        namespace: gpu-operator
      status:
        (containerStatuses[?state.waiting.reason == 'CrashLoopBackOff'
          || state.waiting.reason == 'ImagePullBackOff'
          || state.waiting.reason == 'ErrImagePull'
          || state.waiting.reason == 'CreateContainerConfigError'] | length(@) > `0`): true
```

Consider adding `initContainerStatuses` too — the `nvidia-operator-validator` runs its checks as init containers, and an init-container `ImagePullBackOff` is invisible to `containerStatuses`. Quick confirm: `chainsaw assert` this block against a 2-container pod fixture with `container[1]` crashlooping — it passes today (i.e. fails to detect).

recipes/checks/aws-efa/health-check.yaml
```diff
@@ -14,11 +14,15 @@
...
-# kube-system namespace. Checks that the aws-efa-k8s-device-plugin DaemonSet
-# has at least one ready pod and that no EFA plugin pods (selected by label)
-# are stuck in Pending, Failed, or Unknown phases. Label selectors are used
-# because kube-system is a shared namespace.
+# Validates that the AWS EFA K8s device plugin is running and healthy
+# in the kube-system namespace. The DaemonSet must roll out fully —
+# numberReady == desiredNumberScheduled (with desiredNumberScheduled > 0
+# as a guard against vacuous pass on a 0-node selector). The previous
+# check gated only on numberReady > 0, which passed on partial
+# failures; see #1222 / recipes/checks/nfd/health-check.yaml for the
```
ArangoGutierrez, line 22: **🟡 The DaemonSet fix is correct, but this NFD cross-reference is misleading — and points at a check that is itself broken.**

`numberReady == desiredNumberScheduled` (+ the `desiredNumberScheduled > 0` guard) is the right pattern: both fields are non-`omitempty` in `DaemonSetStatus`, so they are always serialized. 👍

But NFD's `health-check.yaml` does **not** use "the same rationale" — it asserts `(numberUnavailable == `0`): true`, and `NumberUnavailable` **is** `omitempty` (`k8s.io/api/apps/v1/types.go:722`). On a fully-healthy DaemonSet the field is absent, so JMESPath evaluates `null == 0` → false → the assert fails. That makes NFD #1234 fail on a healthy cluster (this PR's own description explains the mechanism).

Two asks:
1. Fix the comment here **and** in `nvidia-dra-driver-gpu/health-check.yaml` (same wording) — don't hold NFD up as the exemplar; it propagates the `omitempty` trap.
2. NFD itself needs the same fix — I'm filing a separate issue so it's tracked.

validators/chainsaw/allowlist_test.go
```diff
@@ -341,8 +341,22 @@ func TestDisallowedOperation_AllowlistFailsClosed(t *testing.T) {
...
+// silently skipping them; otherwise a broken registry could green-
+// light here as long as ≥1 valid entry keeps `checked > 0`. Post-merge
+// review on PR #1244.
 func TestValidateTestReadOnly_RegistryContent(t *testing.T) {
-	provider := recipe.NewEmbeddedDataProvider(recipe.GetEmbeddedFS(), ".")
+	// Empty prefix matches the canonical defaultEmbeddedProvider used at
+	// runtime (pkg/recipe/components.go); "." also works because
+	// filepath.Join cleans it, but matches the runtime path exactly
+	// without the head-scratch.
+	provider := recipe.NewEmbeddedDataProvider(recipe.GetEmbeddedFS(), "")
```
ArangoGutierrez, line 359: **🔵 Minor:** good alignment with the runtime `globalDataProvider` (empty prefix), and `""`/`"."` are equivalent here (`filepath.Join` cleans both; `WalkDir` handles `prefix == ""` explicitly). One consistency note: `provider_test.go` still uses `"."` at ~18 sites, so the suite is now mixed. Consider a follow-up sweep to `""` everywhere — otherwise the next reader gets the same head-scratch in reverse.

validators/chainsaw/allowlist_test.go
```diff
@@ -341,8 +341,22 @@ func TestDisallowedOperation_AllowlistFailsClosed(t *testing.T) {
...
+//
+// CROSS-TEST DEPENDENCY: empty-assertFile and unreadable-path failure
+// modes are deliberately skipped here and owned by
+// TestComponentRegistry_RequiresHealthCheck in pkg/recipe — that test
+// fails first with a clearer message, so we avoid double-reporting on
+// the same root cause. If that test is ever removed or weakened,
+// strengthen this one to fail on empty/unreadable entries rather than
+// silently skipping them; otherwise a broken registry could green-
+// light here as long as ≥1 valid entry keeps `checked > 0`. Post-merge
+// review on PR #1244.
```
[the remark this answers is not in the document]
mchmarny: Intentional, not a typo — the `#1244` reference is correct as-is. The CROSS-TEST DEPENDENCY comment block was added in response to a post-merge review on PR #1244 (the lint guard PR that landed first). It documents *why this code exists*, which is the #1244 review feedback, not the #1245 PR that's currently carrying the comment.

Changing to `#1245` would lose the provenance trail (a future reader following the link would land on this PR's body, which describes 21-file health-check enhancements, and wonder how that relates to a cross-test-dependency note). The current form correctly points to the actual review thread.

No change needed.

recipes/checks/gpu-operator/health-check.yaml
mchmarny: @ArangoGutierrez — verified the kyverno-json source you cited (`vendor/github.com/kyverno/kyverno-json/pkg/core/assertion/assertion.go:43`), confirmed the bug, fixed universally in `fcb5bdcf`.

**Scope:** the projection fix applies to all 22 files with container-state error blocks — 21 enhanced in this PR plus NFD and grove which had the same pre-existing pattern. The CRD-only files (agentgateway-crds, prometheus-operator-crds, slinky-slurm-operator-crds, gke-nccl-tcpxo) don't have Pod-level errors and were untouched.

**Shape:** collapsed the four per-reason blocks into one JMESPath projection per pod-array:

```yaml
status:
  (containerStatuses[?state.waiting.reason == 'CrashLoopBackOff' || state.waiting.reason == 'ImagePullBackOff' || state.waiting.reason == 'ErrImagePull' || state.waiting.reason == 'CreateContainerConfigError'] | length(@) > `0`): true
```

Plus a **parallel `initContainerStatuses` block** in every file — your `nvidia-operator-validator` init-container point was exactly right; those failures are invisible to `containerStatuses` and the old shape would have missed them.

**Note on YAML:** the expression has to be on a single line because YAML's nested-mapping parser can't handle a multi-line key with `:` characters inside without explicit quoting (which chainsaw assert syntax doesn't accept). First attempt with line-broken `||` chains hit `yaml: line N: mapping values are not allowed in this context`; collapsed to single line. Inline rationale comment block added above each pair of projection errors so future readers know why this isn't the obvious shape.

**Verification:** `TestValidateTestReadOnly_RegistryContent` (allowlist sweep) + `TestComponentRegistry_RequiresHealthCheck` (lint guard) both pass on the 27 components. Live-cluster verification still on @mchmarny per the standing agreement.

The deeper structural fix (chainsaw moving to subset-of-slice match by default, or AICR migrating to the in-process Go executor #1236 where this kind of trap goes away entirely) is the right long-term path. The projection workaround unblocks the current PR.

recipes/checks/aws-efa/health-check.yaml
mchmarny: @ArangoGutierrez — the cross-reference is intentional and points at the correct exemplar. Verified:

- **NFD on this branch** uses `(numberReady == desiredNumberScheduled): true` — the GOOD pattern. Check `recipes/checks/nfd/health-check.yaml:1146-1153`.
- The `numberUnavailable == 0` pattern you describe was the **pre-#1235** state of NFD. It was fixed by @njhensley's patch applied in commit `d7e68a24` (merged into main as part of PR #1235's broader chainsaw runtime activation work).
- So when this PR's aws-efa / nvidia-dra-driver-gpu headers say "see #1222 / recipes/checks/nfd/health-check.yaml for the same rationale", they correctly point at NFD's current state where `numberReady == desiredNumberScheduled` is the documented pattern (with inline comment explaining the omitempty trap as the reason for the choice).

Likely cause of the confusion: a local checkout that doesn't include `d7e68a24`, or reviewing the file via a `main` snapshot from before #1235 merged.

**Issue #1248** that you filed asks for the exact fix that's already in main. I've added `Closes #1248` to this PR's body — the PR's broader NFD touch (the projection-form container-state fix from your other comment) is the additional improvement that makes this PR a fit closer.

No code change needed on this comment; the cross-reference stands.

validators/chainsaw/allowlist_test.go
mchmarny: Thanks for the heads-up on the mixed-style state. Agreed it's a `provider_test.go` cleanup worth doing but out of scope for #1245 — the 18 sites all live in a different test file from this PR's diff. Will spin out a separate small sweep if/when it becomes a maintenance issue, or someone passing through that file can roll it into a refactor commit. Not blocking the merge.

recipes/checks/gpu-operator/health-check.yaml
ArangoGutierrez: Verified the new shape empirically against the exact pinned engine (`kyverno-json v0.0.4-...b294ee72`, `chainsaw v0.2.15`, jmespath-community) using chainsaw's own `apis.DefaultCompilers` and the `assertion.Parse` → `Assert` path — reproduced two independent ways (a standalone harness, and a second in-repo `-mod=vendor` harness that replicates the engine's assertion nodes line-for-line); trivially-true/false controls pass in both. Two findings — one good, one that needs a fix before merge.

## ✅ The index fix is correct

The projection flags a bad container at **any** index, the init block catches init-container failures, and there are no false positives:

| fixture | result |
|---|---|
| 2 containers, `[1]` ImagePullBackOff | **MATCH** (flagged) ✓ |
| init container ErrImagePull (→ init block) | **MATCH** ✓ |
| single running / 2× running / empty `[]` / benign `PodInitializing` | NO-MATCH ✓ |

## 🔴 But `length(@)` is null-unsafe, and as written it breaks the check on a healthy cluster

When the status array is **absent**, the projection resolves to `null`, and `length(null)` is a hard type error in jmespath-community:

```
EVAL: status.(initContainerStatuses[?...] | length(@) > `0`):
  Internal error: invalid type for: <nil>, expected: ["string","array","object"]
```

That is fatal because chainsaw's `error` operation **aborts the whole step on an evaluation error** — it does not treat it as no-match. From `pkg/engine/operations/error/operation.go` (v0.2.15):

```go
_errs, err := checks.Check(ctx, o.compilers, candidate.UnstructuredContent(), bindings, ...)
if err != nil {
    return false, err   // poll terminates immediately → the step ERRORS
}
```

The repo's own raw-YAML path does the same — `validators/chainsaw/runner.go:268`:

```go
errs, err := checks.Check(ctx, apis.DefaultCompilers, actual, nil, &check)
if err != nil {
    return errors.Wrap(errors.ErrCodeInternal, ...)   // hard failure
}
```

And `initContainerStatuses` is `omitempty`, so **a pod with no init containers has the field absent entirely** — the common steady-state case (the gpu-operator controller pod, cert-manager controllers, most operator Deployments). So on a fully-healthy cluster the phase blocks pass, the container block passes (a Running pod always populates `containerStatuses`), then the **init block** hits the first no-init pod → `length(null)` → the check errors. That fires across essentially all 22 components.

(The container block is mostly shielded in practice — the `phase: Pending` error block runs first, and Running/Succeeded pods always have `containerStatuses` — but the init block has no such shield, so the guard is needed on both.)

It won't be caught by the structural tests (they never evaluate the assertion against a Pod), and it sits in the deferred live-cluster gap — so it surfaces on the first real `validate --phase deployment` run.

Note the old single-element shape was null-**safe** (kyverno-json's `sliceNode` treats a null array as a plain no-match), so this is a regression introduced by the rewrite — and it's on me: the snippet I suggested omitted the null guard. Apologies for sending you down that path.

## Fix — default the projection to an empty array before `length`, verified

```yaml
status:
  (containerStatuses[?state.waiting.reason == 'CrashLoopBackOff' || state.waiting.reason == 'ImagePullBackOff' || state.waiting.reason == 'ErrImagePull' || state.waiting.reason == 'CreateContainerConfigError'] || `[]` | length(@) > `0`): true
```

`null || ` + the empty-array literal → `[]`, and `length([])` = 0 (precedence is `||` then the `|` pipe, so the default applies before the pipe). I re-ran the whole matrix with the guard added on both blocks: absent array → NO-MATCH (no error), `[1]`-bad → still MATCH, healthy → NO-MATCH. The same guard is needed on the container **and** init blocks in all 22 files.

Happy to share the standalone Go harness (chainsaw's exact compiler, ~15 fixtures incl. the controls) if it's useful for a regression test.

recipes/checks/aws-efa/health-check.yaml
ArangoGutierrez: You're right, and I was wrong here — apologies. I checked NFD on both `main` and this PR's head (`fcb5bdc`): both use `(numberReady == desiredNumberScheduled): true` with the `desiredNumberScheduled > 0` guard, and the file even carries a comment documenting the `numberUnavailable` omitempty trap. Commit `d7e68a24` (the #1235 chainsaw-runtime work) is what fixed it. My finding came from a stale local checkout on a different branch that predated `d7e68a24` — I read the wrong copy of the file. The cross-reference is correct and points at the right exemplar; the comment stands as written. I've closed #1248 as already-fixed. Thanks for the careful push-back.

lalitadithya (approved):

PR #1252, opened 2026-06-09
feat(validators): replace chainsaw binary with in-process executor

### LINKED ISSUE #1236: Replace chainsaw binary with in-process executor for assert/error allowlist subset

mchmarny:
## Summary

The deployment validator no longer exec's `/usr/local/bin/chainsaw` to run registry-declared health checks. A pure-Go executor walks the same `chainsaw.kyverno.io/v1alpha1` Test AST and dispatches `assert` / `error` to kyverno-json's `checks.Check` engine — the exact subset the read-only allowlist already restricts content to.

## Motivation / Context

The deployment validator image shipped a pinned chainsaw binary and a Dockerfile `chainsaw-fetch` multi-stage just to invoke two operations (`assert` / `error`) that the in-process kyverno-json engine already implements. Dropping the binary eliminates a CHAINSAW_* build-arg matrix across five workflows, removes a release-tarball download from the image build, and shrinks the validator's supply-chain footprint to a single static Go binary.

Fixes: #1236
Related: #1219, #1220, #1222, #1223

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Build/CI

## Implementation Notes

- `validators/chainsaw/inprocess.go`: unmarshals a v1alpha1 Test, walks `spec.steps[].try[]`, honors per-test `spec.timeouts.assert`, retries until deadline, dispatches `Assert` / `Error` to `checks.Check` with the project's `apis.DefaultCompilers`.
- `validators/chainsaw/fetch.go`: `clusterFetcher` gains `List()` via the dynamic client + label-selector composition so `Error` operations with label selectors work.
- `validators/chainsaw/runner.go`: package doc rewritten for the in-process model; `runConfig` / `RunOption` / `WithChainsawBinary` removed.
- `validators/chainsaw/binary.go` + `binary_test.go`: deleted.
- `validators/chainsaw/inprocess_test.go`: parity sweep across all 27 in-tree `recipes/checks/*/health-check.yaml` files plus unit cases for happy / assert-fails / error-fires.
- `validators/deployment/Dockerfile`: drops the `chainsaw-fetch` multi-stage.
- `Makefile` + five workflows: drop `CHAINSAW_VERSION` / `CHAINSAW_SHA256_*` build-args. `cmd/gate/Dockerfile` and `.github/actions/e2e` still consume them for their own chainsaw runtime — unchanged.
- `docs/contributor/validator.md`: documents the in-process executor.

## Testing

```bash
make qualify  # passed
```

Parity tests in `validators/chainsaw/inprocess_test.go` cover all 27 in-tree health-check YAMLs (each wrapped in a 200ms context to short-circuit retry loops without a live cluster).

## Risk Assessment

- [x] **Medium** — Touches the deployment validator's execution model; image build matrix changes ripple across 5 workflows.

**Rollout notes:** No CLI / API surface change. Allowlist already restricted registry content to assert/error before this PR, so existing `recipes/checks/*/health-check.yaml` content is byte-for-byte compatible with the in-process executor. The build-gate image still bundles chainsaw for its own runtime (separate code path).

## Checklist


ArangoGutierrez (commented):
Solid refactor. The dispatch split, the allowlist enforcement point, and the bounded-concurrency Run are all preserved, and clusterFetcher.List mirrors the existing Fetch scope/namespace logic exactly. Verified: the single chainsaw.Run caller is updated cleanly, removed symbols are referenced nowhere else, checks.Check is invoked identically to the raw-YAML path, and all in-tree Test checks carry explicit namespaces.

Two items go a bit past a pure no-op refactor, both about parity with the binary rather than crashes, and both mostly latent for the current corpus:
- Named error blocks fail open on transient API errors (inprocess.go L285).
- The 6m process cap became a per-step deadline with no overall cap, leaving ChainsawAssertTimeout stale (inprocess.go L78).

Plus two minors inline: error-code wrapping and missing unit coverage on the new List. Since the assertion-parity claim rests on the live-cluster E2E rather than the unit tests, worth confirming green E2E across the recipes before merge. Details inline.

validators/chainsaw/inprocess.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	check := v1alpha1.NewCheck(resourceSpec)
+	if name != "" {
+		// Single-resource: error passes if the resource doesn't exist
+		// OR if it doesn't match the shape.
+		actual, err := fetcher.Fetch(ctx, apiVersion, kind, namespace, name)
+		if err != nil {
+			// NotFound is the happy path for error blocks.
+			return nil
```
ArangoGutierrez: Fetch wraps every Get failure as ErrCodeNotFound (fetch.go:59), not just genuine 404s. So a named error block treats a transient API error (500 / timeout / forbidden) as resource-absent and passes here without retry — it fails open. The chainsaw binary distinguishes NotFound from other errors and fails closed, so this is a behavior change, and a negative health check passing spuriously is the dangerous direction. List-based error blocks are fine (List returns un-flattened errors that get retried); only the named path is exposed. Suggest preserving apierrors.IsNotFound through Fetch and short-circuiting only on a true NotFound.

validators/chainsaw/inprocess.go
```diff
@@ -0,0 +1,357 @@
...
+	var test v1alpha1.Test
+	if err := yaml.Unmarshal([]byte(yamlContent), &test); err != nil {
+		result.Error = errors.Wrap(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("failed to parse chainsaw Test YAML for component %q", component), err)
+		return result
+	}
+
+	effectiveTimeout := stepTimeout
+	if test.Spec.Timeouts != nil && test.Spec.Timeouts.Assert != nil && test.Spec.Timeouts.Assert.Duration > 0 {
+		effectiveTimeout = test.Spec.Timeouts.Assert.Duration
```
ArangoGutierrez, line 79: Timeout model changed here. Old path capped the whole test at ChainsawAssertTimeout (6m) via context.WithTimeout in runChainsawBinary. New path makes timeouts.assert (5m) a per-step deadline, recomputed fresh each step (executeStepInProcess L113), with no overall test cap — an N-step all-passing test can run up to 5m x N. Since every in-tree check sets timeouts.assert, the 6m constant is now only an unused fallback, and its doc comment (outer timeout for the chainsaw binary process ... must be greater than the internal assert timeout) is stale on every clause. Either wrap the whole test in context.WithTimeout(ctx, stepTimeout) to keep the outer cap, or update the constant comment to its new fallback role.

validators/chainsaw/inprocess.go
```diff
@@ -0,0 +1,357 @@
...
+			return errors.New(errors.ErrCodeInternal,
+				fmt.Sprintf("%s %s/%s: %s", kind, namespace, name, formatFieldErrors(errs)))
+		}
+		return nil
+	}
+
+	// List-and-match: assert passes if at least one item matches.
+	items, err := fetcher.List(ctx, apiVersion, kind, namespace, labels)
+	if err != nil {
+		return errors.Wrap(errors.ErrCodeInternal,
```
ArangoGutierrez: fetcher.List already returns structured errors (ErrCodeInvalidRequest for a bad apiVersion, ErrCodeNotFound for no REST mapping). Wrapping with ErrCodeInternal overwrites the inner code — the do-not-double-wrap rule in CLAUDE.md. Propagate as-is, or wrap preserving the code. Same pattern in evaluateError (~L304).

validators/chainsaw/fetch.go
```diff
@@ -62,3 +63,52 @@ func (f *clusterFetcher) Fetch(ctx context.Context, apiVersion, kind, namespace,
...
 }
+
+// List enumerates resources of the given kind in the given namespace,
+// optionally narrowed by label match. labels is a string→string map
+// converted to the canonical "k=v,k=v" Kubernetes label selector
+// format. An empty labels map yields no selector (all resources of the
+// kind in the namespace).
+//
+// Returns an empty slice (not error) when no resources match; callers
+// distinguish "no matches" from "list failed".
```
ArangoGutierrez, line 86: clusterFetcher.List (REST mapping, scope branch, k=v selector join, UnstructuredContent extraction) is only exercised via the in-memory fakeFetcher; this dynamic-client path has no direct unit test. Per the coverage gate (new exported method at 0% blocks), a client-go/dynamic/fake-backed test would cover it. A cluster-scoped case plus a label-selector case would also lock in the Fetch/List scope parity.

validators/chainsaw/inprocess.go
```diff
@@ -0,0 +1,357 @@
...
+	result.Passed = true
+	slog.Info("health check passed", "component", component)
+	return result
+}
+
+// executeStepInProcess walks a step's Try operations sequentially.
+// Each operation has its own retry-until-deadline loop (matching the
+// chainsaw binary's per-operation semantics).
+func executeStepInProcess(ctx context.Context, try []v1alpha1.Operation, fetcher ResourceFetcher, stepTimeout time.Duration) error {
+	deadline := time.Now().Add(stepTimeout)
```
ArangoGutierrez, line 131: Nit: the comment says this matches the chainsaw binary per-operation semantics, but all ops in a step share one deadline computed at step entry — chainsaw gives each op its own clock. Benign for the current corpus (error ops pass instantly when healthy, and a failing op short-circuits the step). Also only timeouts.assert is read; timeouts.error would be ignored, though no in-tree check sets it today.

ArangoGutierrez (commented):
Re-reviewed at `8e1e7143ca`. All five points are addressed and the fixes hold up:

- **Named `error` fail-open** — `Fetch` now splits `apierrors.IsNotFound` → `ErrCodeNotFound` from everything else → `ErrCodeUnavailable`, and `evaluateError` only treats a true NotFound as the happy path (`stderrors.Is`), otherwise it propagates and retries. Fails closed on transient apiserver errors now, matching the old binary.
- **Timeout cap** — `context.WithTimeout(ctx, effectiveTimeout)` bounds the whole Test, so the N × effectiveTimeout blowup in the retrying case is gone.
- **Error-code wrapping** — the `ErrCodeInternal` wraps around the fetcher calls are all gone; codes propagate untouched through `executeStepInProcess` and `runChainsawTestInProcess`. Went a bit further than I'd flagged.
- **`List` coverage** — `fetch_test.go` is solid: table-driven over namespaced + cluster-scoped, label filtering, and the error-code mappings, plus a dedicated reactor test that `ServiceUnavailable` maps to `ErrCodeUnavailable` — that one pins the fail-closed fix directly. Real assertions against the real dynamic/fake client, no theater.
- **Per-op comment** — now accurately describes the shared per-step deadline and the ignored `timeouts.error`.

One non-blocking leftover: `pkg/defaults/timeouts.go` still describes `ChainsawAssertTimeout` as the outer timeout for the chainsaw binary process that must exceed the 5m internal assert timeout. There's no binary now, and the outer cap is `effectiveTimeout` — which for every in-tree check is `timeouts.assert` (5m), so the 6m constant only kicks in as a fallback when a Test omits `timeouts.assert`. Net budget for a multi-step Test is ~5m vs the old ~6m, fine for these checks. Worth a one-line comment refresh next time that file is touched.

LGTM with that noted. Good turnaround.

mchmarny:
Thanks for the thorough re-review, @ArangoGutierrez — picked up the `ChainsawAssertTimeout` comment refresh in `bb3fd047`. Rewrote it to describe its actual role as the fallback per-Test budget when YAML omits `spec.timeouts.assert`, and called out that #1236 removed the binary the prior comment referenced.

Holding off on merging for one more cycle — added a small unconditional dump of `/etc/containerd/config.toml` and `/etc/containerd/conf.d/99-nvidia.toml` to `collect-debug-artifacts.sh` in `9ecbf064` to capture the toml that nvidia-ctk emits, since the next GPU run will help unblock #1237.

lockwobr (approved):

PR #1261, opened 2026-06-09
fix(validator): fail fast on eval errors; nil-safe health checks

yuanchen8911:
## Summary

Three regression fixes for deployment-phase health-check validation, introduced by the recent in-process executor + deepened-check stack (#1235/#1245/#1252):

1. **In-process executor retries permanent eval errors** (`inprocess.go`)
2. **Nil-unsafe container-status expressions** (22 health checks)
3. **dynamo-platform check asserts a non-existent Deployment name**

## Motivation / Context

On an EKS H100 cluster, `aicr validate --phase deployment` hung ~8 min and returned `status=other` ("pod not found for Job") instead of real results. Root cause is three interacting issues:

1. **#1252** — `runAssertWithRetry` / `runErrorWithRetry` retry on *any* non-nil error, so a permanent JMESPath evaluation error is retried for the full 6m assert window. The old chainsaw binary treated eval errors as terminal and failed fast.
2. **#1245** — `(init)containerStatuses[?...] | length(@)` throws `invalid type for: <nil>` on any pod without (init) containers (the common case), which *feeds* defect #1.
3. **dynamo-platform** — `validate-deployment-exists` asserts a Deployment named `dynamo-operator`, but the chart prefixes it with the release name (`dynamo-platform-dynamo-operator-controller-manager`), so the assert never matched and retried to the deadline — a false negative against a healthy component.

(Defects #1/#2 independently confirmed by Nathan.)

Related: #1235, #1245, #1252

## Type of Change

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe`) — `recipes/checks/*`
- [x] Validator (`pkg/validator`) — `validators/chainsaw/inprocess.go`

## Implementation Notes

1. Classify assertion-engine errors as `ErrCodeInvalidRequest` (terminal) and short-circuit both retry loops via a new `isTerminalAssertErr` helper, so a non-evaluable expression fails fast instead of burning the assert deadline. Transient failures (mismatch, NotFound) still retry. Adds a regression test asserting a terminal eval error returns in `< AssertRetryInterval` rather than retrying to the deadline.
2. Coalesce to an empty array — `(initContainerStatuses || ` + "`[]`" + `)[?...]` — across all 22 affected component health checks (both `initContainerStatuses` and `containerStatuses`).
3. Match the dynamo-operator manager Deployment by its stable `app.kubernetes.io/name=dynamo-operator` label (List-and-match) instead of a release-name-dependent hardcoded name.

**Scope note (`conditions[?...]` filters):** the remaining `(conditions[?type == '...'])` filters are intentionally *not* coalesced — they are projection asserts without `| length(@)` (so nil `.status.conditions` yields an empty match, not a `<nil>` throw) and target CRDs/Deployments that reliably populate `.status.conditions`. They are not the same throwing class as the `length()` expressions fixed here.

## Testing

```bash
golangci-lint run -c .golangci.yaml ./validators/chainsaw/...   # 0 issues
go test ./validators/chainsaw/...                                # ok, coverage 60.7%
```

New regression test `TestRunChainsawTestInProcess_TerminalEvalErrorFailsFast` reproduces the exact `invalid type for: <nil>` throw and asserts fast-fail (returns in 0.00s vs. retrying to deadline).

On-cluster (EKS H100): deployment phase went from an 8m timeout (`status=other`) to **PASSED in ~21s**; full run on this branch = **deployment 4/4 + conformance 11/11 green**. (Performance phase ran a real Qwen3-8B benchmark and reports a true TTFT-p99 SLA miss — a separate cluster/SLA matter, not this change.) Full `make qualify` not run locally — recommend CI / a maintainer run before merge.

## Risk Assessment

- [x] **Low** — Data/expression fixes plus a localized executor error-classification change; no API or schema changes, easy to revert.

**Rollout notes:** Reclassifying assertion-engine errors `ErrCodeInternal → ErrCodeInvalidRequest` is a caller-visible semantic shift (CLI exit 8→2, HTTP 500→400) for a malformed bundled-registry expression — intentional and anticipated by the design comment; backwards compatible otherwise.

## Checklist

- [x] Linter passes (`golangci-lint` on changed package — 0 issues)
- [x] Tests pass locally (`go test ./validators/chainsaw/...`), incl. new regression test
- [ ] Full `make qualify` (recommend running in CI / before merge)
- [x] I added a regression test for the terminal-eval-error fast-fail

### DIFF
--- a/validators/chainsaw/inprocess.go
+++ b/validators/chainsaw/inprocess.go
@@ -167,6 +167,9 @@ func runAssertWithRetry(ctx context.Context, a *v1alpha1.Assert, fetcher Resourc
 		if lastErr == nil {
 			return nil
 		}
+		if isTerminalAssertErr(lastErr) {
+			return lastErr
+		}
 		remaining := time.Until(deadline)
 		if remaining <= 0 {
 			return lastErr
@@ -194,6 +197,9 @@ func runErrorWithRetry(ctx context.Context, e *v1alpha1.Error, fetcher ResourceF
 		if lastErr == nil {
 			return nil
 		}
+		if isTerminalAssertErr(lastErr) {
+			return lastErr
+		}
 		remaining := time.Until(deadline)
 		if remaining <= 0 {
 			return lastErr
@@ -208,6 +214,17 @@ func runErrorWithRetry(ctx context.Context, e *v1alpha1.Error, fetcher ResourceF
 		case <-time.After(wait):
 		}
 	}
+}
+
+// isTerminalAssertErr reports whether err is a non-retryable failure: a
+// malformed or non-evaluable assert/error expression (ErrCodeInvalidRequest,
+// e.g. a JMESPath operation that throws "invalid type for: <nil>"). Such an
+// error will never become valid by retrying, so the retry loops fail fast
+// instead of burning the full assert deadline. Transient failures — a resource
+// not yet in the desired state (assertion mismatch) or not-found — carry other
+// codes and continue to retry until the deadline.
+func isTerminalAssertErr(err error) bool {
+	return stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, ""))
 }
 
 // evaluateAssert runs a single positive assertion against the cluster.
@@ -239,7 +256,7 @@ func evaluateAssert(ctx context.Context, a *v1alpha1.Assert, fetcher ResourceFet
 		}
 		errs, checkErr := checks.Check(ctx, apis.DefaultCompilers, actual, nil, &check)
 		if checkErr != nil {
-			return errors.Wrap(errors.ErrCodeInternal,
+			return errors.Wrap(errors.ErrCodeInvalidRequest,
 				fmt.Sprintf("%s %s/%s: assertion engine error", kind, namespace, name), checkErr)
 		}
 		if len(errs) > 0 {
@@ -267,7 +284,7 @@ func evaluateAssert(ctx context.Context, a *v1alpha1.Assert, fetcher ResourceFet
 		}
 		errs, checkErr := checks.Check(ctx, apis.DefaultCompilers, actual, nil, &check)
 		if checkErr != nil {
-			return errors.Wrap(errors.ErrCodeInternal,
+			return errors.Wrap(errors.ErrCodeInvalidRequest,
 				fmt.Sprintf("%s in %q: assertion engine error", kind, namespace), checkErr)
 		}
 		if len(errs) == 0 {
@@ -314,7 +331,7 @@ func evaluateError(ctx context.Context, e *v1alpha1.Error, fetcher ResourceFetch
 		}
 		errs, checkErr := checks.Check(ctx, apis.DefaultCompilers, actual, nil, &check)
 		if checkErr != nil {
-			return errors.Wrap(errors.ErrCodeInternal,
+			return errors.Wrap(errors.ErrCodeInvalidRequest,
 				fmt.Sprintf("%s %s/%s: assertion engine error", kind, namespace, name), checkErr)
 		}
 		if len(errs) == 0 {
@@ -339,7 +356,7 @@ func evaluateError(ctx context.Context, e *v1alpha1.Error, fetcher ResourceFetch
 		}
 		errs, checkErr := checks.Check(ctx, apis.DefaultCompilers, actual, nil, &check)
 		if checkErr != nil {
-			return errors.Wrap(errors.ErrCodeInternal,
+			return errors.Wrap(errors.ErrCodeInvalidRequest,
 				fmt.Sprintf("%s in %q: assertion engine error", kind, namespace), checkErr)
 		}
 		if len(errs) == 0 {
--- a/validators/chainsaw/inprocess_test.go
+++ b/validators/chainsaw/inprocess_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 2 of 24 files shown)


validators/chainsaw/inprocess.go
```diff
@@ -210,6 +216,19 @@ func runErrorWithRetry(ctx context.Context, e *v1alpha1.Error, fetcher ResourceF
...
+// isTerminalAssertErr reports whether err is a non-retryable failure: a
+// malformed or non-evaluable assert/error expression (ErrCodeInvalidRequest,
+// e.g. a JMESPath operation that throws "invalid type for: <nil>"). Such an
+// error will never become valid by retrying, so the retry loops fail fast
+// instead of burning the full assert deadline. Transient failures — a resource
+// not yet in the desired state (assertion mismatch) or not-found — carry other
+// codes and continue to retry until the deadline. Fixes the #1252 regression
+// where the baked chainsaw binary classified eval errors as terminal but the
+// in-process executor retried them for the entire window.
+func isTerminalAssertErr(err error) bool {
```
mchmarny, line 226: The fast-fail path (`isTerminalAssertErr` → early return in both retry loops) has no unit test. The original bug was an 8-minute silent hang; if the error code classification regresses (e.g. someone changes `ErrCodeInvalidRequest` → `ErrCodeInternal` on checkErr), the hang silently comes back.

A table-driven test for `runAssertWithRetry` with a deadline of `time.Now().Add(10*time.Second)` and a mock evaluator that always returns `ErrCodeInvalidRequest` could assert the function returns in < 1 second. Same for `runErrorWithRetry`. Worth adding to lock in this invariant.

validators/chainsaw/inprocess.go
```diff
@@ -210,6 +216,19 @@ func runErrorWithRetry(ctx context.Context, e *v1alpha1.Error, fetcher ResourceF
...
+	return stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, ""))
+}
+
 // evaluateAssert runs a single positive assertion against the cluster.
 // Returns nil if the assertion passes (resource exists AND matches the
 // shape), non-nil error otherwise.
```
mchmarny, line 232: nit: last sentence (`Fixes the #1252 regression where…`) is PR-description context rather than load-bearing code context — it'll drift as the executor evolves. The first two sentences (what it detects and why it's terminal) are the part that helps a future reader. Consider trimming.

mchmarny (dismissed):
Three clean, targeted fixes for the deployment-phase hang regression. The error-code reclassification (`ErrCodeInternal` → `ErrCodeInvalidRequest` for checkErr) is semantically correct — eval errors are a malformed-expression problem, not an internal one — and the `isTerminalAssertErr` placement in both retry loops is right (after nil check, before sleep). The `|| \`[]\`` JMESPath coalescing is applied consistently across all 22 files. The dynamo label-match is a solid fix that generalizes correctly to any release-name prefix.

One medium concern: the fast-fail path that prevents the 8-minute hang has no unit test. If someone later changes the error code on `checkErr`, the hang silently returns. Suggest adding a short retry-loop test before this merges. One nit on comment length. CI fully green including on-cluster Tier 1.

validators/chainsaw/inprocess.go
yuanchen8911: Done — added `TestRunChainsawTestInProcess_TerminalEvalErrorFailsFast` (table-driven over both an `assert` op and an `error` op, since they share the `isTerminalAssertErr` guard). It feeds a terminal eval error (`length(@)` on an absent field → `invalid type for: <nil>`) with a 30s step budget and asserts the result is `ErrCodeInvalidRequest` **and** that it returns in `< AssertRetryInterval` — so a future change of the checkErr code back to `ErrCodeInternal` (re-introducing the 8-minute hang) fails the test. Coverage on the package is back to +0.3% over main.

validators/chainsaw/inprocess.go
yuanchen8911: Good call — trimmed. Dropped the trailing `Fixes the #1252 regression where…` sentence; the comment now keeps just what the helper detects and why it is terminal.

validators/chainsaw/inprocess_test.go
```diff
@@ -295,3 +296,65 @@ func TestRunChainsawTestInProcess_RegistryCorpusParses(t *testing.T) {
...
+			if !stderrors.Is(r.Error, errors.New(errors.ErrCodeInvalidRequest, "")) {
+				t.Errorf("expected ErrCodeInvalidRequest (terminal), got %v", r.Error)
+			}
+			if elapsed >= defaults.AssertRetryInterval {
+				t.Fatalf("terminal eval error was retried (took %s >= AssertRetryInterval %s) — #1252 regression",
+					elapsed, defaults.AssertRetryInterval)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
yuanchen8911: Good catch — addressed. The regression test is now table-driven over the full matrix: both ops (`assert` → `runAssertWithRetry`, `error` → `runErrorWithRetry`) **× both fetch paths** (named single-Get and List-and-match). The List subcase omits `metadata.name` and seeds the fixture via `f.addList(...)`, so the List-path `checkErr` sites (inprocess.go:271/343) are now pinned alongside the named-Get sites — a future regression on the List path (the one the `(init)containerStatuses` bug actually rode in on) now fails the test. Package coverage went 60.7% → 65.4%.

mchmarny (approved):

PR #1233, opened 2026-06-08
feat(recipe): add OKE GB200 perf check

yuanchen8911:
## Summary

Adds GB200 OKE performance phases following the existing training/inference split: NCCL goals for training overlays and `inference-perf` throughput/TTFT goals for the Dynamo inference overlay. OKE training advertises NVLS only for now; OKE NET/RDMA stays out until the OCI testbed proves a non-Socket NCCL transport end to end.

## Motivation / Context

The OKE GB200 leaves were missing the recommended performance phase. This adds conservative smoke goals by mirroring the established EKS/GKE overlay pattern while avoiding an unverified OKE NET contract.

Fixes: #1007
Related: N/A

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] CLI (`cmd/aicr`, `pkg/cli`)
- [ ] API server (`cmd/aicrd`, `pkg/server`)
- [x] Recipe engine / data (`pkg/recipe`)
- [ ] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [ ] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [x] Validator (`pkg/validator`)
- [ ] Core libraries (`pkg/errors`, `pkg/k8s`)
- [ ] Docs/examples (`docs/`, `examples/`)
- [ ] Other: ____________

## Implementation Notes

- Adds `nccl-all-reduce-bw-nvls >= 500` at `gb200-oke-training` so `gb200-oke-ubuntu-training` and `gb200-oke-ubuntu-training-kubeflow` inherit the training performance goal.
- Registers only `CriteriaServiceOKE + CriteriaAcceleratorGB200` for the NCCL NVLS variant and ships the OKE NVLS runtime template.
- Keeps the OKE NVLS runtime template on the IMEX/ComputeDomain claim path via `resourceClaimTemplateName: nccl-all-reduce-imex`.
- Leaves OKE NET/RDMA out of the support matrix and recipe goals until #1007 or OCI testbed data proves pod RDMA exposure and non-Socket NCCL logs.
[...]

PR #1262, opened 2026-06-09
fix(validators): make GKE NCCL node selector dynamic

### LINKED ISSUE #1256: nccl-all-reduce-bw training gate is a fixed absolute fabric-specific busbw value applied to SKU-agnostic recipes → false-fails EKS/H100 small SKUs
## Summary

The `nccl-all-reduce-bw` training gate is a fixed absolute bus-bandwidth (busbw) value, but the training recipes it lives in are **SKU-agnostic** (matched only on `service` + `accelerator`, no node shape or fabric class). On a node whose fabric differs from the one the floor was calibrated against, the gate false-fails a healthy run.

This is a sibling of #1254 (the inference throughput gate), **but the fix is different**: inference throughput is ~count-linear, so it gets per-GPU normalization. NCCL busbw is **topology/transport-class dependent and is *not* count-linear** — per-GPU scaling must **not** be applied. The correct invariant is **fabric/transport class**, not GPU count.

## Evidence

**Flat absolute compare, no scaling.** `validators/performance/nccl_all_reduce_bw_constraint.go:257`:
```go
passed := bandwidth >= (threshold * 0.9)
```
The parsed busbw is compared directly against the recipe value with 10% tolerance — no GPU-count or topology scaling.

**The clean active threshold false-fail today is H100 on EKS small SKUs.** The check only runs for `(variant, service, accelerator)` tuples present in `supportedNCCLCombinations` (`nccl_all_reduce_bw_constraint.go:151`); the lookup is exact-service with no `Any` fallback (`:183`). So the **active** default-variant gates are just:

| Overlay | constraint | value | status |
|---|---|---|---|
| `h100-eks-training` | `nccl-all-reduce-bw` | `>= 300` | **active** (EKS/H100 supported) |
| `h100-gke-cos-training` | `nccl-all-reduce-bw` | `>= 250` | **active** (GKE/H100 supported) |
| `h100-aks-training` | `nccl-all-reduce-bw` | `>= 100` | declared but **skips** (AKS not in matrix) |
| `h200-eks-training` | `nccl-all-reduce-bw` | `>= 300` | declared but **skips** (EKS/H200 not in matrix) |
| `b200-gke-cos-training` | `nccl-all-reduce-bw` | `>= 100` | declared but **skips** (GKE/B200 not in matrix) |
| `gb200-eks-training` | `nccl-all-reduce-bw-net` / `-nvls` | `>= 40` / `>= 500` | **active** (EKS/GB200 NET+NVLS supported) — reference pattern |

So the clean active threshold false-fail is concretely **EKS/H100 `p5.4xlarge × 2`**: a supported tuple, the gate runs, and a healthy 1-GPU-per-node run measures network-class busbw well under `>= 300`. **GKE/H100 small SKUs expose the same SKU-agnostic recipe problem but along a broader axis**: the GKE runtime is hardcoded to `a3-megagpu-8g` (TCPXO/FastRak, `cloud.google.com/gke-accelerator=nvidia-h100-mega-80gb` worker selector at `nccl_all_reduce_bw_constraint.go:836`, and GPU-NIC-network discovery at `:563`), so an `a3-highgpu-1g × 2` cluster likely fails earlier on runtime/scheduling/template assumptions rather than reaching the busbw threshold. The AKS/H200/B200-default rows are **declared-but-inactive** gates — alignment/config concerns to fix before enabling support, not current runtime false-fails.

**The gate runs on 2-node × 1-GPU clusters.** It skips only when fewer than 2 GPU nodes are present (`:217`); a 2-node, 1-GPU-per-node cluster on a supported service/accelerator runs the full gate.

**The thresholds are fixed absolute topology/fabric-specific values, not GPU-count-scaled.** The `300`/`250` H100 numbers were calibrated on 8-GPU high-bandwidth node templates with NVLink/NVSwitch plus multi-NIC transport; a 1-GPU-per-node configuration measures a materially different path (pure inter-node NIC, network class, tens of GB/s) that cannot meet such a floor.

**Sub-8-GPU / different-fabric H100 nodes exist on every provider:** EKS `p5.4xlarge` (1), GKE `a3-highgpu-1g/2g/4g`, Azure `NC40ads` (1) / `NC80adis` (2, PCIe-NVL). SKU-agnostic recipes will legitimately land on them.

## Network performance: why busbw is fabric-class-bound, not GPU-count-bound

busbw is defined as `algbw × 2(n-1)/n` — deliberately constructed to report **per-link** bandwidth, not aggregate. For a multi-node all-reduce the bottleneck is the inter-node NIC path, and on a balanced fabric each GPU has its own fixed-bandwidth NIC (e.g., 1× 400 Gbps EFA per H100 on `p5`). So 8-GPU → 8 NICs, 4-GPU → 4, 2-GPU → 2, but **per-link bandwidth is identical**, and busbw normalizes to that per-link figure. A single floor **may be valid within a measured equivalent fabric class** — it is not a guarantee (measured bandwidth still depends on rank count, NCCL channel selection, NIC mapping, and fabric), but within one balanced fabric busbw lands in the same band regardless of GPUs/node.

### What actually breaks it: fabric class, not count

In the real H100 catalog, the smaller-GPU shapes usually aren't "the same node with fewer GPUs" — they're a **weaker fabric tier**:
- Azure **NCads** (NC40/NC80) is PCIe H100 NVL with **no InfiniBand**; the 8-GPU **ND** family is SXM + IB.
- GCP **a3-highgpu** (1/2/4/8g) has less per-GPU network than **a3-megagpu-8g**.
- AWS `p5.4xlarge` (1 GPU) vs `p5.48xlarge` differ in per-GPU NIC provisioning.

That's why the calibrated numbers diverge (EKS SXM `300` vs AKS NCads `100`) — the gap is **per-GPU fabric bandwidth**, not GPU count. A single target is portable across 2/4/8-GPU nodes **only if they are a measured-equivalent balanced RDMA fabric**; it is **not** portable if the 2-GPU shape is a cheaper, RDMA-less SKU.

### One caveat even within a fabric

At small total rank counts (2 nodes × 2 GPU = 4 ranks), the `2(n-1)/n` factor and fewer NCCL channels make measured busbw run a bit lower than at 16 ranks. Set the floor with ~10–15% headroom below the large-node value — the existing 10% tolerance roughly absorbs it, but don't calibrate the floor right at the 8-GPU measurement.

## Proposed fix: make NCCL gates fabric/transport-class aware, not GPU-count scaled

- **Keep `nccl-all-reduce-bw` as the legacy absolute gate** for fixed-shape recipes (unchanged semantics — same backward-compat stance as #1254).
- **Do not infer NVLS from `gpuCountPerNode`.** `nccl-all-reduce-bw-nvls` is MNNVL across an NVL72 IMEX domain (`recipes/validators/catalog.yaml:219`), a GB200 NVL72-class transport — GPU count alone does not imply it.
- **Add or extend named variants only where the validator actually has runtime templates and transport assertions** for that `(service, accelerator, fabric)` — i.e. extend `supportedNCCLCombinations` (`nccl_all_reduce_bw_constraint.go:151`) and ship the matching testdata template, rather than declaring a variant a recipe can't run.
- **For SKU-agnostic recipes, pick one of:**
  - **(a)** set a conservative **weakest-in-scope fabric floor** and document it as a coarse liveness gate (simple, portable, weaker on premium fabric), or
  - **(b)** introduce a **recipe criteria dimension for fabric / SKU class** and carry measured per-class thresholds (correct, more work, requires per-class hardware to calibrate).
- The NET vs NVLS variants then separate "inter-node fabric healthy" from "NVLS/MNNVL fabric healthy where that variant is implemented" — and keep the 1-GPU case coherent (NET-only).

**Do not auto-skip `gpuCountPerNode == 1`.** A 2-node × 1-GPU all-reduce is still a useful signal: it proves distributed training can launch and the inter-node path carries traffic. Treat it as a **NET / network-class baseline** and name/document it honestly as such — not full-node bandwidth validation.

### Implementation sketch
1. **Floor/variant by actual supported fabric class, not by GPU count.** Keep the legacy `nccl-all-reduce-bw` absolute gate. Where per-fabric coverage is wanted, gate on the NET/NVLS variants only for `(service, accelerator)` combinations the validator supports.
2. **Align the declared-but-skipping overlays** (`h100-aks`, `h200-eks`, `b200-gke`): either add real support (matrix entry + testdata template + transport assertion) with a measured per-fabric floor, or drop the declared constraint until support exists, so a skipped gate isn't mistaken for a passing one.
3. **SKU-agnostic floors**: apply option (a) or (b) above; if (a), set the H100 EKS/GKE floors to a weakest-in-scope value that a small SKU can still meet, and document the coarse-liveness intent.
4. Update `nccl` unit tests for the chosen approach (fabric-class selection / weakest-in-scope floors / 1-GPU NET-only path).

## Why the inference fix does NOT transfer

For #1254, inference throughput is ~count-linear at fixed concurrency-per-GPU, so a per-GPU floor × benchmarked GPU count is correct. busbw is **not** count-linear — multiplying a busbw threshold by GPU count would bake in a hardware coincidence (NIC count ≈ GPU count on today's SKUs) and is wrong in general. The correct invariant for NCCL is **per-fabric / per-transport**, which busbw already normalizes for count.

## Scope / related
- Sibling of #1254 (inference throughput). Same root cause — SKU-agnostic recipes vs node-shape/fabric-fixed perf gates — but distinct fix (fabric/transport-class floors here; per-GPU normalization there).
- Clean active threshold false-fail today: **EKS/H100 `p5.4xlarge × 2`**. GKE/H100 small SKUs hit the same SKU-agnostic recipe problem but current runtime assumptions (`a3-megagpu-8g` hardcoding) may fail before threshold evaluation. `gb200-eks` already uses the NET/NVLS variant pattern and is the reference. `h100-aks` / `h200-eks` / `b200-gke` declare thresholds that currently skip — align before enabling.

## References (fabric-class point)
- AWS EC2 accelerated computing: https://aws.amazon.com/ec2/instance-types/accelerated-computing/
- AWS EFA / multi-network-card instance types: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-acc-inst-types.html
- GCP GPU machine types: https://cloud.google.com/compute/docs/gpus/

xdu31:
## Summary

Fixes the GKE NCCL all-reduce bandwidth validator to dynamically discover the `cloud.google.com/gke-accelerator` label from target nodes instead of hardcoding `nvidia-h100-mega-80gb`. Now fails explicitly when GKE nodes have missing or mixed accelerator labels to prevent WorkerCount divergence.

## Motivation / Context

The NCCL validator hardcoded `cloud.google.com/gke-accelerator=nvidia-h100-mega-80gb` in `platformWorkerScheduling`, so worker pods could only schedule on `a3-megagpu-8g` nodes. Any GKE H100 SKU with a different accelerator label (e.g. `a3-highgpu-1g` with `nvidia-h100-80gb`) failed at scheduling.

Additionally, when GKE nodes had missing or mixed accelerator labels, the code silently fell through with no nodeSelector while `WorkerCount` was sized from `config.Nodes`. This could cause worker placement to diverge from the counted cohort, leading to invalid bandwidth measurements.

Fixes: #1256

## Type of Change

## Component(s) Affected

## Implementation Notes

- Added `commonGKEAccelerator(nodes []v1.Node) string` helper that returns the `cloud.google.com/gke-accelerator` label shared by all nodes, or `""` when nodes disagree or any node lacks the label (mirrors `commonGPUProduct` for OKE).
- Modified `platformWorkerScheduling()` signature to return `(map[string]string, []v1.Toleration, error)` instead of just the first two values.
- GKE branch now fails explicitly (returns error) when `commonGKEAccelerator()` returns `""`, preventing silent WorkerCount/scheduling divergence.
- Error message uses the `gkeAcceleratorLabel` constant as requested in review.
- Updated all call sites and tests to handle the new error return.

## Testing

```bash
golangci-lint run -c .golangci.yaml ./validators/performance/...  # 0 issues
go test ./validators/performance/...                              # ok (all tests pass)
```

Updated `TestPlatformWorkerScheduling`:
- Existing homogeneous-label cases (`a3-megagpu-8g`, `a3-highgpu-1g`) updated to handle error return
- Previous "missing label" test now expects error instead of nil selector
- Previous "mixed-SKU pool" test now expects error instead of nil selector

## Risk Assessment

- [x] **Low** — Isolated change to one validator package. Changes function signature from `(ns, tols)` to `(ns, tols, err)` but this is an internal function. GKE nodes with homogeneous accelerator labels work as before; mixed/missing labels now fail fast instead of silently diverging.

**Rollout notes:** N/A — non-breaking for valid clusters. Clusters with mixed/missing GKE accelerator labels will now fail explicitly with a clear error message instead of silently producing invalid results.

## Checklist

- [x] Tests pass locally (`go test ./validators/performance/...`)
- [x] Linter passes (golangci-lint on changed packages)
- [ ] I updated docs if user-facing behavior changed — N/A, no user-facing surface changed
- [x] Changes follow existing patterns in the codebase (mirrors `commonGPUProduct`)

### DIFF
--- a/validators/performance/nccl_all_reduce_bw_constraint.go
+++ b/validators/performance/nccl_all_reduce_bw_constraint.go
@@ -599,7 +599,10 @@ func applyNCCLResources(ctx *validators.Context, dynamicClient dynamic.Interface
 	}
 
 	// Build effective worker scheduling: user override takes precedence over platform default.
-	defaultNodeSelector, defaultTolerations := platformWorkerScheduling(service, instanceType, config.Nodes)
+	defaultNodeSelector, defaultTolerations, err := platformWorkerScheduling(service, instanceType, config.Nodes)
+	if err != nil {
+		return err
+	}
 	effectiveNodeSelector := defaultNodeSelector
 	// Gate on len() rather than != nil so an explicit but empty selector does
 	// not silently clear the platform default for scheduling while
@@ -834,21 +837,32 @@ func createUnstructured(ctx context.Context, dynamicClient dynamic.Interface, gv
 // platformWorkerScheduling returns the default nodeSelector and tolerations
 // for NCCL worker pods on the given service. instanceType is only used for EKS;
 // nodes (the accelerator-narrowed target set from resolveTargetGPUNodes) is
-// only used for OKE, where the selector is derived from the shared
-// nvidia.com/gpu.product label.
-func platformWorkerScheduling(service recipe.CriteriaServiceType, instanceType string, nodes []v1.Node) (map[string]string, []v1.Toleration) {
+// used for GKE (the gke-accelerator label) and OKE (the shared
+// nvidia.com/gpu.product label).
+func platformWorkerScheduling(service recipe.CriteriaServiceType, instanceType string, nodes []v1.Node) (map[string]string, []v1.Toleration, error) {
 	switch service {
 	case recipe.CriteriaServiceEKS:
 		return map[string]string{
 			"node.kubernetes.io/instance-type": instanceType,
-		}, []v1.Toleration{{Operator: v1.TolerationOpExists}}
+		}, []v1.Toleration{{Operator: v1.TolerationOpExists}}, nil
 	case recipe.CriteriaServiceGKE:
-		return map[string]string{
-				"cloud.google.com/gke-accelerator": "nvidia-h100-mega-80gb",
-			}, []v1.Toleration{
-				{Operator: v1.TolerationOpExists},
-				{Key: "nvidia.com/gpu", Operator: v1.TolerationOpEqual, Value: "present", Effect: v1.TaintEffectNoSchedule},
-			}
+		// GKE accelerator node pools carry cloud.google.com/gke-accelerator
+		// (set by the node pool spec). Pin workers to the SKU shared by every
+		// target node so the selector matches the cohort WorkerCount was sized
+		// against — narrowByAccelerator only filters by the H100 product
+		// prefix, so a mixed pool (e.g. a3-megagpu-8g + a3-highgpu-1g) can
+		// reach this branch. Fail if accelerator labels are missing or mixed
+		// to prevent WorkerCount divergence (sizing for N nodes but scheduling
+		// to a subset).
+		acc := commonGKEAccelerator(nodes)
+		if acc == "" {
+			return nil, nil, aicrErrors.New(aicrErrors.ErrCodeInvalidRequest,
+				fmt.Sprintf("GKE nodes have missing or mixed %s labels — cannot derive a nodeSelector that matches the full WorkerCount cohort", gkeAcceleratorLabel))
+		}
+		return map[string]string{gkeAcceleratorLabel: acc}, []v1.Toleration{
+			{Operator: v1.TolerationOpExists},
+			{Key: "nvidia.com/gpu", Operator: v1.TolerationOpEqual, Value: "present", Effect: v1.TaintEffectNoSchedule},
+		}, nil
 	case recipe.CriteriaServiceOKE:
 		// OKE bare-metal GB200 pools are commonly tainted and may coexist
 		// with other GPU shapes under one control plane. Tolerate the pool
@@ -862,11 +876,11 @@ func platformWorkerScheduling(service recipe.CriteriaServiceType, instanceType s
 		if product := commonGPUProduct(nodes); product != "" {
 			nodeSelector = map[string]string{gpuProductLabel: product}
 		}
-		return nodeSelector, []v1.Toleration{{Operator: v1.TolerationOpExists}}
+		return nodeSelector, []v1.Toleration{{Operator: v1.TolerationOpExists}}, nil
 	case recipe.CriteriaServiceAny, recipe.CriteriaServiceAKS, recipe.CriteriaServiceKind, recipe.CriteriaServiceLKE, recipe.CriteriaServiceBCM:
-		return nil, nil
+		return nil, nil, nil
 	default:
-		return nil, nil
+		return nil, nil, nil
 	}
 }
 
@@ -891,6 +905,27 @@ func commonGPUProduct(nodes []v1.Node) string {
 		}
 	}
 	return product
+}
+
+// commonGKEAccelerator returns the cloud.google.com/gke-accelerator label
+// shared by every node, or "" when the nodes disagree or any node lacks the
+// label. Mirrors commonGPUProduct, applied to the GKE-specific label NCCL
+// workers must pin to so the nodeSelector matches the same cohort
+// WorkerCount was sized against.
+func commonGKEAccelerator(nodes []v1.Node) string {
+	accelerator := ""
+	for _, n := range nodes {
+		a := n.Labels[gkeAcceleratorLabel]
+		if a == "" {
+			return ""
+		}
+		if accelerator == "" {
+			accelerator = a
+		} else if a != accelerator {
+			return ""
+		}
+	}
+	return accelerator
 }
 
 // applyNCCLWorkerScheduling sets the nodeSelector and tolerations on the "node"
--- a/validators/performance/nccl_gke_utils.go
+++ b/validators/performance/nccl_gke_utils.go
@@ -27,6 +27,12 @@ import (
 	"k8s.io/client-go/dynamic"
 )
 
+// gkeAcceleratorLabel is set by GKE node-pool spec on every accelerator node
+// (e.g. "nvidia-h100-mega-80gb" for a3-megagpu-8g, "nvidia-h100-80gb" for
+// a3-highgpu-1g). Read off the target node to pin NCCL workers to the SKU
+// the test was sized against, without hardcoding any single shape.
+const gkeAcceleratorLabel = "cloud.google.com/gke-accelerator"
+
 // discoverGKEGPUNICNetworks lists networks.networking.gke.io and returns
 // GPU NIC network names (those containing "gpu-nic"), sorted alphabetically.
 // GKE clusters provision these with cluster-specific prefixes (e.g.,
--- a/recipes/overlays/h100-eks-training.yaml
+++ b/recipes/overlays/h100-eks-training.yaml
@@ -87,6 +87,15 @@ spec:
       # false-fail. Making the gate fabric/transport-class aware is tracked in
       # https://github.com/NVIDIA/aicr/issues/1256.
       constraints:
+        # Fixed absolute floor calibrated on p5.48xlarge (8x H100 SXM,
+        # 32x EFA, NVLink/NVSwitch). This recipe is SKU-agnostic — it
+        # matches all EKS/H100 instance types, including small shapes
+        # (e.g. p5.4xlarge: 1 GPU, 1 EFA) whose pure inter-node NIC path
+        # measures tens of GB/s and cannot clear this floor. See #1256.
+        # The proper fix is either a weakest-in-scope liveness floor
+        # (option a) or a per-fabric criteria dimension (option b);
+        # both need measured numbers from the small SKUs that we don't
+        # have yet. Keeping the calibrated value here until then.
         - name: nccl-all-reduce-bw
           value: ">= 300"
     conformance:
--- a/recipes/overlays/h100-gke-cos-training.yaml
+++ b/recipes/overlays/h100-gke-cos-training.yaml
@@ -95,6 +95,16 @@ spec:
       # false-fail. Making the gate fabric/transport-class aware is tracked in
       # https://github.com/NVIDIA/aicr/issues/1256.
       constraints:
+        # Fixed absolute floor calibrated on a3-megagpu-8g (8x H100,
+        # 8x GPUDirect TCPXO NICs). This recipe is SKU-agnostic — it
+        # matches all GKE/H100 instance types, including small shapes
+        # (e.g. a3-highgpu-1g: 1 GPU, 1 gVNIC ~100 Gbps) whose pure
+        # inter-node NIC path measures single-digit GB/s and cannot
+        # clear this floor. See #1256. The proper fix is either a
+        # weakest-in-scope liveness floor (option a) or a per-fabric
+        # criteria dimension (option b); both need measured numbers
+        # from the small SKUs that we don't have yet. Keeping the
+        # calibrated value here until then.
         - name: nccl-all-reduce-bw
           value: ">= 250"
     conformance:
--- a/validators/performance/nccl_test.go
+++ b/validators/performance/nccl_test.go
@@ -396,7 +396,10 @@ func TestAcceleratorProductMatchers(t *testing.T) {
 
 func TestPlatformWorkerScheduling(t *testing.T) {
 	t.Run("EKS returns instance-type selector", func(t *testing.T) {
-		ns, tols := platformWorkerScheduling(recipe.CriteriaServiceEKS, "p5.48xlarge", nil)
+		ns, tols, err := platformWorkerScheduling(recipe.CriteriaServiceEKS, "p5.48xlarge", nil)
+		if err != nil {
+			t.Fatalf("platformWorkerScheduling() error = %v", err)
+		}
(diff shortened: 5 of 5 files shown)


validators/performance/nccl_gke_utils.go
```diff
@@ -22,11 +22,25 @@ import (
 
 	"github.com/NVIDIA/aicr/pkg/defaults"
 	aicrErrors "github.com/NVIDIA/aicr/pkg/errors"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/dynamic"
 )
```
mchmarny, line 29: nit: the error message duplicates the label string that's already in the `gkeAcceleratorLabel` constant one line above. `fmt.Sprintf("GPU node missing %s label", gkeAcceleratorLabel)` avoids the drift risk if the label name ever changes.

mchmarny (dismissed):
Clean fix. The `discoverGKENodeConfig` helper faithfully mirrors `discoverEKSNodeConfig` — same single-node-read pattern, same `ErrCodeInternal` for a missing label, same `config.Nodes[0]` call site (safe: `determineGPUConfig` guarantees a non-empty slice before returning). `platformWorkerScheduling` thread-through is correct; the `gkeAccelerator` zero-value-when-not-GKE pattern matches `instanceType` for EKS. Tests are table-driven and cover both valid SKUs plus the missing-label path. CI is fully green.

One nit inline (use the `gkeAcceleratorLabel` constant in the error message). Two mechanics items before merge:

- Missing type label: add `bug` (per PR policy — type labels aren't auto-assigned).
- Branch is behind `main`: rebase + squash before merging.

validators/performance/nccl_gke_utils.go
xdu31: Good catch — and self-resolved by the rebase. The new commit removes `discoverGKENodeConfig` entirely; the GKE lookup now happens inline in `platformWorkerScheduling` and uses the `gkeAcceleratorLabel` constant directly (matches the OKE pattern added in #1233). No more duplication to drift on.

validators/performance/nccl_all_reduce_bw_constraint.go
```diff
@@ -834,21 +834,30 @@ func createUnstructured(ctx context.Context, dynamicClient dynamic.Interface, gv
...
+		// GKE accelerator node pools carry cloud.google.com/gke-accelerator
+		// (set by the node pool spec). Read it from the first target node so
+		// the selector matches the actual SKU (a3-megagpu-8g, a3-highgpu-1g,
+		// etc.) instead of being hardcoded to a single shape. Falls through to
+		// no selector on missing label, mirroring the OKE non-GFD path.
+		var nodeSelector map[string]string
+		if len(nodes) > 0 {
+			if acc := nodes[0].Labels[gkeAcceleratorLabel]; acc != "" {
+				nodeSelector = map[string]string{gkeAcceleratorLabel: acc}
 			}
```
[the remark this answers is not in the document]
xdu31: Addressed in the latest push. The GKE branch now uses a new `commonGKEAccelerator(nodes)` helper that mirrors `commonGPUProduct` exactly — returns the label only when every target node carries the same non-empty value, otherwise "" (and the selector is omitted, same as the OKE non-GFD path). Mixed-pool case can no longer pin to a subset.

Added `TestCommonGKEAccelerator` (all-share / mixed / missing-label) and a `platformWorkerScheduling` GKE subtest covering the mixed-SKU pool. Tests pass.

validators/performance/nccl_all_reduce_bw_constraint.go
```diff
@@ -834,21 +834,31 @@ func createUnstructured(ctx context.Context, dynamicClient dynamic.Interface, gv
...
+		// GKE accelerator node pools carry cloud.google.com/gke-accelerator
+		// (set by the node pool spec). Pin workers to the SKU shared by every
+		// target node so the selector matches the cohort WorkerCount was sized
+		// against — narrowByAccelerator only filters by the H100 product
+		// prefix, so a mixed pool (e.g. a3-megagpu-8g + a3-highgpu-1g) can
+		// reach this branch and would otherwise size for N nodes while the
+		// nodes[0]-derived selector matched only a subset. Falls through to
+		// no selector on missing/mixed labels, mirroring the OKE non-GFD path.
+		var nodeSelector map[string]string
+		if acc := commonGKEAccelerator(nodes); acc != "" {
```
yuanchen8911: Mixed/missing GKE accelerator labels fail open here, so worker placement can diverge from the sized cohort.

`narrowByAccelerator` keeps every `NVIDIA-H100-*` node (it filters only on the `nvidia.com/gpu.product` prefix), and `determineGPUConfig` derives `WorkerCount` and `GPUCountPerNode` from that target set (`GPUCountPerNode` is taken from `targetNodes[0]` only). On a mixed H100 pool — e.g. `a3-megagpu-8g` + `a3-highgpu-1g` — both SKUs survive into the counted set, so `commonGKEAccelerator(nodes)` returns `""` and this branch emits no nodeSelector. Since `applyNCCLWorkerScheduling` only writes a selector when one is non-empty, the `WorkerCount` workers can then schedule across the heterogeneous pool rather than the cohort the test was sized against. The all-reduce bandwidth result is measured against a mismatched set and silently passes/fails the floor instead of erroring.

The `len(gpuNICs) < 8` prerequisite does not gate this: `discoverGKEGPUNICNetworks` counts cluster-level `networking.gke.io/Network` CRs, not per-node NICs, so a mixed cluster that has the 8 `gpu-nic` networks provisioned (because the mega-8g nodes need them) passes that check and reaches this branch. There is partial implicit mitigation via GPU resource requests (if `targetNodes[0]` is a mega-8g node, workers request 8 GPUs and can't fit on a 1-GPU node), but it is ordering-dependent — if `targetNodes[0]` is the small SKU, workers request 1 GPU and can land anywhere.

Suggestion: fail closed when `commonGKEAccelerator` returns `""` on a non-empty GKE target set (mirroring how `narrowByAccelerator` hard-errors on zero matches), or narrow the counted node set to the same cohort the selector will match, so sizing and placement can't diverge.

yuanchen8911 (changes requested):
Requesting changes on two items (details in the inline comment and review above):

1. Mixed/missing GKE accelerator labels fail open in `platformWorkerScheduling` — when `commonGKEAccelerator(nodes)` returns `""` on a non-empty GKE target set, no nodeSelector is emitted while `WorkerCount`/`GPUCountPerNode` were sized from `config.Nodes`, so worker placement can diverge from the counted cohort and the bandwidth result is measured against a mismatched set instead of erroring. Please fail closed (error on a mixed/missing set, mirroring `narrowByAccelerator`'s zero-match error) or narrow the counted node set to the selected cohort.

2. The PR description is stale — it references `discoverGKENodeConfig`, a new `gkeAccelerator` parameter, and `TestDiscoverGKENodeConfig` that don't exist; the code adds `commonGKEAccelerator` with an unchanged `platformWorkerScheduling` signature. Please update the description before merge.

Open question: do heterogeneous H100 GKE pools (`a3-megagpu-8g` + `a3-highgpu-1g`) occur in supported deployments? If pools are always single-SKU, item 1 is latent and can be deferred to #1256; if mixed pools are realistic, the fail-closed guard belongs in this PR.

xdu31:
Updated to address review comments:

**Fixed:**
✅ Made `platformWorkerScheduling` fail when GKE nodes have missing/mixed `cloud.google.com/gke-accelerator` labels (prevents WorkerCount divergence)
✅ Error message uses `gkeAcceleratorLabel` constant as requested
✅ Updated PR description (removed references to non-existent `discoverGKENodeConfig`, `gkeAccelerator` parameter, and `TestDiscoverGKENodeConfig`)
✅ All tests pass

**Label issue:**
❌ Cannot add `bug` label — it doesn't exist in the repo label list. The closest available labels are `invalid` or `needs-triage`. Should I use one of those, or can you add the `bug` label?

**Branch status:**
Branch is behind `main` — will rebase + squash once approved.

yuanchen8911 (dismissed):
Thanks, both findings are fixed. Two minor issues from CodeRabbit remain unresolved.

yuanchen8911 (dismissed):
/lgtm

yuanchen8911 (approved):
need another rebase

Issue #1228, opened 2026-06-08
aicr recipe list: add structural-health + coverage columns


njhensley:
**Parent:** #1224 (Epic: Recipe health tracking, ADR-009 V1)

## Summary

Extend the `aicr recipe list` command **delivered by #1207 / #1208 (merged)** with the ADR-009 §4 health view — a structural status and a compact declared-coverage summary per recipe — delegating computation to `pkg/health`. #1208 already delivers the command, enumeration, criteria/curated/source columns, `--format json|yaml|table`, the `--service/--accelerator/--intent/--os` filters, and `--data`; A4 adds **only** the health-derived columns. `pkg/cli` stays logic-free.

## Scope

- Wire the existing `recipe list` command to call `pkg/health.Compute()` for the listed combos and add: **table** → a structural status column + compact coverage (`R:2 D:4 P:1 C:10`); **json/yaml** → the full `declared_coverage` (named checks per phase) plus the per-dimension status map. Preserve #1208's curated/source fields and filter behavior.
- **Degrade gracefully:** if #1227 hasn't merged yet, `constraints_wellformed` surfaces as `unknown`; the status column still renders.
- The command renders its own curated tabwriter grid (the generic `pkg/serializer` table writer flattens to two-column FIELD/VALUE and cannot render a multi-row grid).
- **Docs same PR:** extend the `aicr recipe list` section in `docs/user/cli-reference.md` with the health columns.

## Key files

`pkg/cli/recipe_list.go` (the merged `list` command), `pkg/cli/recipe.go`, `pkg/health` (`Compute`), `pkg/cli/root.go` (`cmd.Root().Writer`), `docs/user/cli-reference.md`.

## Tests

Capture `cmd.Root().Writer`: the table carries the status + coverage columns atop #1208's criteria/curated/source columns; `--format json` carries the full `declared_coverage` + per-dimension status map; the status column degrades to `unknown` when #1227 is absent. The enumeration round-trip test (feed each `CatalogEntry.Criteria` from `ListCatalog` back through `BuildFromCriteria`) may live here or with the enumeration command.

## Acceptance

`aicr recipe list` shows the structural status + coverage columns in all formats, delegating to `pkg/health`, without losing #1208's enumeration/curated/source output; documented in `cli-reference.md`.

## Dependencies

Depends on the merged #1207/#1208 command and on #1226 (#1227 for full grading).

Issue #1227, opened 2026-06-08
pkg/health: constraints_wellformed signal (parse-only, hermetic)


njhensley:
**Parent:** #1224 (Epic: Recipe health tracking, ADR-009 V1)

## Summary

Add the third graded signal, `constraints_wellformed`, **hermetically**.

ADR-009's original phrasing ("inline replay under `--no-cluster` is clean") is **not achievable offline**: the only evaluator, `pkg/constraints.Evaluate(constraint, *snapshotter.Snapshot)`, calls `path.ExtractValue(snap)` *before* parsing, and the sole wiring rejects a nil snapshot (`pkg/client/v1/aicr.go`). With no snapshot — ADR-009's hermetic premise — there is no "clean replay" to run. So the V1 signal is **parse-only well-formedness**.

## Scope

- **Primary (graded, fully hermetic):** for every merged `Constraint`, call the exported, snapshot-free parsers directly — `ParseConstraintPath(c.Name)` (`pkg/constraints/extractor.go`) and `ParseConstraintExpression(c.Value)` (`pkg/constraints/constraint.go`). Any parse failure → `fail`, with the offending constraint + parser error in `Detail`.
- **Secondary (graded `warn`, hermetic):** resolve the leaf via `BuildFromCriteriaWithEvaluator` with an **injected stub** `ConstraintEvaluatorFunc` (a plain `func(Constraint) ConstraintEvalResult`, trivially mockable) returning a deterministic *satisfied* result so the constraint-aware path executes without cluster data. Surface any resulting `RecipeResult.Metadata.ConstraintWarnings`/`ExcludedOverlays` as `warn`. Hermetically this catches resolution/evaluation-setup problems and mixin-composition inconsistencies — **not** cluster-dependent constraint outcomes.
- **Fail honestly:** a malformed constraint is `fail`, never a silent `pass`.
- Fold the signal into the #1225 rollup.

## Seam note

A3 needs **no `pkg/validator` import**. The acyclic integration is `pkg/health → pkg/constraints → pkg/recipe`. *(The stale `pkg/recipe/builder.go` doc comment "wrapping `validator.EvaluateConstraint`" references a symbol that does not exist; flag for a separate one-line fix.)*

## Out of scope (deferred)

[...]

Issue #1226, opened 2026-06-08
pkg/health: chart_pinned + declared_coverage signals


njhensley:
**Parent:** #1224 (Epic: Recipe health tracking, ADR-009 V1)

## Summary

Add the two structural signals that are **pure reads** of the resolved `RecipeResult` — no external integration — on top of the #1225 package core.

## Scope

- **`chart_pinned`** (graded): iterate `RecipeResult.ComponentRefs`; for every `Type == ComponentTypeHelm`, require a non-empty `Version`. Empty → `fail`. Pure in-repo field check — **no Helm render**.
  - **ADR-006 layer 1 only:** this checks the **chart version** pin, *not* image digests. Layer-2 image overrides / layer-3 chart-default sub-images need a render and are out of scope. Kustomize `defaultTag` is also out of scope, so a pure-Kustomize recipe scores `chart_pinned` as a **vacuous `pass`** — note this in the signal's `Detail` and the matrix header so the column isn't misread as "supply-chain pinned".
- **`declared_coverage`** (descriptor, **never graded**): from `RecipeResult.Validation.{Readiness,Deployment,Performance,Conformance}`, record per phase `Declared bool`, sorted `Checks []string`, `len(Constraints)`. A minimal recipe (e.g. Slurm dropping K8s checks) must not be penalized.

## Key files

`pkg/recipe/metadata.go` (`ComponentRef`: `Type`, `Version`), `pkg/recipe/validation.go` (`ValidationConfig`, `ValidationPhase`: `Checks`, `Constraints`), `docs/design/006-image-pinning-policy.md` (layer model).

## Tests

Unpinned Helm component → `fail` on `chart_pinned`; pure-Kustomize recipe → vacuous `pass` with an explanatory `Detail`; minimal recipe → `pass` with sparse `declared_coverage`, never penalized; `Checks` emitted sorted.

## Acceptance

Both signals computed; `chart_pinned` is layer-1-only and says so; `declared_coverage` is descriptive and never moves status.

## Dependencies

Depends on #1225. Feeds the rollup; consumed by #1228 and #1229.

Issue #1224, opened 2026-06-08
Epic: Recipe health tracking (ADR-009 V1) — enumeration, structural soundness, public matrix


njhensley:
## Summary

Implement [ADR-009: Recipe Health Tracking](https://github.com/NVIDIA/aicr/blob/main/docs/design/009-recipe-health-tracking.md) V1.

AICR ships ~50 leaf recipes resolved from ~79 overlays, but there is no command that enumerates them and no durable, communicated record of per-recipe health. This epic closes both gaps using **only hermetic, offline, deterministic signals** — reproducible from a checkout, no GPU, no network, no wall-clock on the computed path.

V1 ships exactly three things (ADR-009 §1):

1. **`aicr recipe list`** — enumerate resolvable criteria combinations. ✅ **Delivered by #1207 / #1208 (merged):** the command plus the `MetadataStore.ListCatalog` enumeration seam.
2. **`pkg/health.Compute()`** — per-recipe **structural soundness** from three graded signals (`resolves`, `chart_pinned`, `constraints_wellformed`) plus one descriptor (`declared_coverage`), rolled up to `pass | warn | fail` (`unknown` is held, never rolled up).
3. **One generated public matrix** (`docs/user/recipe-health.md`), refreshed weekly by a drift-only bot PR — a faithful clone of the BOM generate-and-PR loop.

## Children

- [x] #1207 — `aicr recipe list` + `MetadataStore.ListCatalog` enumeration seam (delivered by #1208, merged; subsumes the originally-planned A1)
- [x] #1225 — `pkg/health`: package scaffold, `Compute()` loop, `resolves` signal, rollup + `unknown`-held, determinism harness
- [x] #1226 — `pkg/health`: `chart_pinned` (graded) + `declared_coverage` (descriptor)
- [x] #1227 — `pkg/health`: `constraints_wellformed` via parse-only well-formedness (hermetic)
- [x] #1228 — augment `aicr recipe list` with structural-health status + coverage columns
- [x] #1229 — `tools/health` generator + `make recipe-health-docs`/`recipe-health-check` + public matrix + nav/ownership/budget
- [x] #1230 — `.github/workflows/health-refresh.yaml` weekly bot PR

## Acceptance criteria

[...]

PR #1291, opened 2026-06-10
feat(health): pkg/health core Compute loop, resolves signal, rollup

### LINKED ISSUE #1225: pkg/health: core Compute() loop, resolves signal, rollup, determinism
**Parent:** #1224 (Epic: Recipe health tracking, ADR-009 V1)

## Summary

Stand up the new `pkg/health` package: the enumeration→resolution loop, the `resolves` signal, the status rollup with `unknown`-held semantics, and the determinism harness. The read-only signals are #1226; the constraint signal is #1227.

`pkg/health` is a **standalone** package (not in `pkg/recipe`) to keep the future v1.1 evidence import acyclic — `pkg/evidence/attestation` already imports `pkg/recipe`, so hosting health in `pkg/recipe` would later close a `recipe → evidence → recipe` cycle. V1 deps: `pkg/recipe`, `pkg/serializer`, `pkg/errors`, `pkg/defaults`.

## Scope

- New package `pkg/health` with the ADR-009 §4 types: `Options`, `PhaseCoverage`, `DeclaredCoverage`, `StructureHealth`, `ComboHealth`, `Report`, and `SchemaVersion = "1.0.0"`.
- `func Compute(ctx, opts Options) (*Report, error)`: enumerate via `MetadataStore.ListCatalog(nil)` filtered to `Curated == true` (the **merged #1208** seam — each `CatalogEntry` gives `Name` → `ComboHealth.LeafOverlay` and `Criteria`), resolve each through a **single shared** `recipe.NewBuilder(...)`'s `BuildFromCriteria` (so the owner-stamp invariant holds across combos). `ListCatalog` is name-sorted (deterministic); consume as-is or re-sort by criteria tuple — either is deterministic.
- **`resolves`** signal (graded, intrinsic to the loop): resolves → `pass`; non-transient error → `fail` (error in `Detail`); `ErrCodeTimeout`/`ErrCodeInternal` → `unknown` (held, excluded from rollup, never an empty field).
- **Rollup:** `fail` if any graded dimension fails; else `warn`; else `pass`. `unknown` is held. The rollup iterates `Dimensions` generically so #1226/#1227 add dimensions without changing rollup code.
- Add `pkg/health/**` to `.github/CODEOWNERS` and a `pkg/health/**` area glob to `.github/labeler.yml`.

## Determinism

No `time.Now()` on the computed path. Caveat: the `unknown`-from-`ErrCodeTimeout` mapping is the one non-deterministic input — byte-determinism is asserted **on the non-error path only**; tests inject errors via a fake `DataProvider`, never real timeouts.

## Key files

`pkg/recipe/catalog.go` (`ListCatalog`, `CatalogEntry` — from #1208), `pkg/recipe/builder.go` (`NewBuilder`, `BuildFromCriteria`), `pkg/recipe/metadata.go` (`RecipeResult`), `pkg/serializer` (deterministic marshal).

## Tests

Clean recipe → `pass`; injected transient error (fake provider) → `unknown` held out of rollup; **determinism: serialize the `Report` via the deterministic marshaller and assert byte-equality across two `Compute` runs** (not struct `DeepEqual` — `Report` carries `map[string]string`, whose nondeterminism only manifests at marshal time); empty-catalog → empty `Report`, no panic.

## Acceptance

`pkg/health` exists with `Compute()`, the `resolves` signal, a generic rollup with `unknown`-held, and a marshal-and-compare byte-determinism test. CODEOWNERS + labeler glob land here. 75% coverage floor cleared.

## Dependencies

Builds on the merged `MetadataStore.ListCatalog` (#1208). Blocks #1226, #1227, #1228, #1229.

mchmarny:
## Summary

Adds the standalone `pkg/health` package per ADR-009 V1: the core enumeration→resolution loop (`Compute`), the graded `resolves` signal, a generic per-recipe status rollup with `unknown`-held semantics, and byte-deterministic output.

## Motivation / Context

First building block of the Recipe health tracking epic (ADR-009 V1). `pkg/health` is standalone (not in `pkg/recipe`) so the future v1.1 evidence axis stays acyclic — `pkg/evidence/attestation` already imports `pkg/recipe`, so hosting health there would later close a `recipe → evidence → recipe` cycle. This PR ships only the package core and the `resolves` dimension; the read-only signals and the constraint signal land on top of it.

Fixes: #1225
Related: #1224 (epic), #1226, #1227

## Type of Change

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe`) — new sibling package `pkg/health`
- [x] Build/CI/tooling — CODEOWNERS + labeler

## Implementation Notes

- `Compute` enumerates leaf overlays via `MetadataStore.ListCatalog(filter)` (filtered to `IsLeaf`) and resolves each through a **single shared** `recipe.NewBuilder(...)` so the owner-stamp invariant holds across combos. Combos are sorted by criteria string (tie-broken by leaf name) for determinism.
- `resolves` grading: resolves → `pass`; non-transient error → `fail` (message in `Detail`); `ErrCodeTimeout`/`ErrCodeInternal` → `unknown`, **held** out of the fail/warn rollup (re-run rather than penalize).
- Rollup iterates the `Dimensions` map generically (precedence `fail > warn > unknown > pass`), so #1226/#1227 add dimensions without touching rollup code. A held `unknown` surfaces as the recipe status when nothing fails/warns — never silently as `pass`.
- The `resolves`-only core leaves `DeclaredCoverage` at its zero value; the descriptor types are defined now so later signals slot in without a schema change.
- No `time.Now()` on the computed path. The one non-deterministic input (unknown-from-real-timeout) is not exercised offline; the determinism test asserts byte-equality on the non-error embedded-catalog path via `serializer.MarshalYAMLDeterministic`.

## Testing

```bash
golangci-lint run -c .golangci.yaml ./pkg/health/... ./pkg/defaults/...   # 0 issues
make test                                                                 # pass, total 76.5%
```

- Unit: table-driven `classifyResolve`, `isTransient`, `rollup` (incl. unknown-held precedence).
- Integration: embedded catalog → every leaf carries a `resolves` dimension with a status consistent with the rollup; byte-determinism across two `Compute` runs; non-matching filter → empty report; failing provider → surfaced load error.
- `pkg/health` coverage: **95.5%** (new package; floor is 75%).

## Risk Assessment

- [x] **Low** — Additive new package, no existing call sites, no user-facing surface (CLI/docs deferred to #1228/#1229).

**Rollout notes:** N/A

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — no user-facing surface in this PR)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -21,6 +21,9 @@
 # Data
 recipes/** @nvidia/aicr-maintainer
 
+# Recipe health (ADR-009): the area/recipes owners triage the weekly refresh
+pkg/health/** @nvidia/aicr-maintainer
+
 # CI/CD and automation
 /Makefile @nvidia/aicr-maintainer
 /.goreleaser.yaml @nvidia/aicr-maintainer
--- a/pkg/health/doc.go
+++ b/pkg/health/doc.go
@@ -1,0 +1,58 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package health computes per-recipe structural health across the whole
+// criteria matrix, as specified by ADR-009 (Recipe Health Tracking).
+//
+// # Overview
+//
+// Compute enumerates every leaf recipe in the catalog and scores each
+// against a set of hermetic, offline, deterministic signals — reproducible
+// from a checkout with no GPU, no network, and no wall-clock. The V1 core
+// (this package) ships the enumeration→resolution loop, the graded resolves
+// signal, the per-recipe status rollup, and a byte-deterministic report.
+//
+// # Standalone package
+//
+// health is a standalone package rather than living in pkg/recipe to keep a
+// future evidence-driven validation axis acyclic: pkg/evidence/attestation
+// already imports pkg/recipe, so hosting health in pkg/recipe would later
+// close a recipe → evidence → recipe import cycle once health consumes
+// evidence. A standalone package depending on both stays acyclic.
+//
+// # Signals and rollup
+//
+// Each leaf is scored on graded dimensions whose per-dimension state is one
+// of pass, warn, fail, or unknown. The resolves dimension (this package) is
+// graded from whether the recipe builder resolves the criteria without
+// error. Additional dimensions (chart_pinned, declared_coverage,
+// constraints_wellformed) are layered on by later work and feed the same
+// generic rollup without changing rollup logic.
+//
+// The rolled-up per-recipe status is fail if any graded dimension fails,
+// else warn if any warns, else unknown if any is held (transient resolver
+// errors are held rather than penalized — re-run rather than mark fail),
+// else pass. unknown is a representable value, never an empty field a
+// consumer could misread as pass.
+//
+// # Determinism
+//
+// There is no time.Now() on the computed path, so output is a pure function
+// of the checkout. Report carries map-typed fields whose nondeterminism
+// only manifests at marshal time; callers serializing the report must use
+// serializer.MarshalYAMLDeterministic. The sole non-deterministic input is
+// the unknown-from-transient mapping, which depends on real timeouts; tests
+// inject errors via a fake DataProvider and assert byte-determinism on the
+// non-error path only.
+package health
--- a/pkg/health/health.go
+++ b/pkg/health/health.go
@@ -1,0 +1,283 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package health
+
+import (
+	"context"
+	stderrors "errors"
+	"sort"
+
+	"github.com/NVIDIA/aicr/pkg/defaults"
+	"github.com/NVIDIA/aicr/pkg/errors"
+	"github.com/NVIDIA/aicr/pkg/recipe"
+)
+
+// SchemaVersion is the version of the Report schema. It is forward-compat
+// metadata, emitted if/when the report is serialized so consumers can detect
+// shape changes across releases.
+const SchemaVersion = "1.0.0"
+
+// Per-dimension and rolled-up status values. These are the only legal values
+// for a graded dimension's state and for ComboHealth status.
+const (
+	// StatusPass means the dimension (or recipe) is structurally sound.
+	StatusPass = "pass"
+
+	// StatusWarn means the dimension surfaced a non-fatal concern.
+	StatusWarn = "warn"
+
+	// StatusFail means a graded dimension failed; it forces the rollup to fail.
+	StatusFail = "fail"
+
+	// StatusUnknown is held: a transient resolver error (ErrCodeTimeout or
+	// ErrCodeInternal) prevented a confident verdict. It is excluded from the
+	// fail/warn rollup so a re-runnable hiccup never penalizes a recipe, but it
+	// surfaces as the recipe status when nothing fails or warns — never as pass.
+	StatusUnknown = "unknown"
+)
+
+// Graded dimension keys. Later work adds more keys (chart_pinned,
+// declared_coverage, constraints_wellformed) without changing the rollup.
+const (
+	// DimResolves is the graded dimension scoring whether the recipe builder
+	// resolves the criteria into a RecipeResult without error.
+	DimResolves = "resolves"
+)
+
+// Options configures a Compute run.
+type Options struct {
+	// Provider is the recipe DataProvider to enumerate and resolve against.
+	// Nil selects the package-global embedded catalog.
+	Provider recipe.DataProvider
+
+	// Version stamps the recipe builder version used during resolution. It
+	// does not affect health scoring; it mirrors the CLI version for parity
+	// with normal recipe generation.
+	Version string
+
+	// Filter narrows enumeration to leaf overlays matching every explicitly
+	// set criteria dimension. Nil enumerates all leaf combos.
+	Filter *recipe.Criteria
+}
+
+// PhaseCoverage records which named checks and how many phase-level
+// constraints a single validation phase declares. It is a descriptor, never
+// graded. Populated by the declared_coverage signal (added in later work);
+// the resolves-only core leaves it at its zero value.
+type PhaseCoverage struct {
+	// Declared is true when the phase block is present after overlay merge.
+	Declared bool `json:"declared" yaml:"declared"`
+
+	// Checks are the named checks the phase declares, sorted.
+	Checks []string `json:"checks,omitempty" yaml:"checks,omitempty"`
+
+	// Constraints is the phase-level constraint count.
+	Constraints int `json:"constraints" yaml:"constraints"`
+}
+
+// DeclaredCoverage captures, per validation phase, which validations a recipe
+// defines. It is a descriptor — it never moves the rolled-up status, so a
+// deliberately minimal recipe is not penalized for declaring fewer checks.
+type DeclaredCoverage struct {
+	Readiness   PhaseCoverage `json:"readiness"   yaml:"readiness"`
+	Deployment  PhaseCoverage `json:"deployment"  yaml:"deployment"`
+	Performance PhaseCoverage `json:"performance" yaml:"performance"`
+	Conformance PhaseCoverage `json:"conformance" yaml:"conformance"`
+}
+
+// StructureHealth is the structural-soundness axis for one recipe.
+type StructureHealth struct {
+	// Status is the rolled-up verdict: pass | warn | fail | unknown (held).
+	Status string `json:"status" yaml:"status"`
+
+	// Dimensions maps each graded dimension key to its state. The rollup
+	// iterates this map generically, so adding a dimension does not change
+	// rollup logic.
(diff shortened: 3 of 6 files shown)


pkg/health/health.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// isTransient reports whether err carries a re-runnable resolver code
+// (ErrCodeTimeout or ErrCodeInternal) anywhere in its chain.
+func isTransient(err error) bool {
+	var se *errors.StructuredError
+	if !stderrors.As(err, &se) {
+		return false
+	}
+	return se.Code == errors.ErrCodeTimeout || se.Code == errors.ErrCodeInternal
```
njhensley: **The `unknown` bucket is too wide: `ErrCodeInternal` is overwhelmingly a *deterministic* failure on the resolve path, not a transient one.**

`isTransient` holds both `ErrCodeTimeout` and `ErrCodeInternal` as `unknown` ("re-run rather than penalize"). That's correct for timeouts, but in the path `BuildFromCriteria` actually executes, `ErrCodeInternal` is reserved for deterministic structural defects — re-running never clears them, so the recipe sits at `unknown` forever instead of surfacing as `fail`, defeating the point of the `resolves` signal.

The genuinely reachable trigger today is a registry `healthCheck.assertFile` pointing at a missing path: the embedded provider returns a bare `fs.ErrNotExist`, which `hydrateHealthCheckAsserts` flattens via `PropagateOrWrap(..., ErrCodeInternal, ...)` (`pkg/recipe/metadata_store.go:1226`) on the per-combo path (`finalizeRecipeResult:862`). That deterministic defect grades `unknown` rather than `fail`.

Note genuine cancellation does **not** need `ErrCodeInternal` here — store-walk and per-combo cancellation are both wrapped as `ErrCodeTimeout` (`metadata_store.go:199`, `builder.go:140`), so dropping `ErrCodeInternal` loses no transient coverage for this package.

Suggest narrowing to genuinely re-runnable signals and letting everything else fall through to `fail`:

```go
func isTransient(err error) bool {
    if stderrors.Is(err, context.DeadlineExceeded) || stderrors.Is(err, context.Canceled) {
        return true
    }
    var se *errors.StructuredError
    if !stderrors.As(err, &se) {
        return false
    }
    return se.Code == errors.ErrCodeTimeout
}
```

This flips the `ErrCodeInternal → unknown` cases in `TestClassifyResolve`/`TestIsTransient`, so it's a deliberate call — worth confirming against the ADR-009 intent for the held-`unknown` semantics. Latent rather than urgent (all current leaves resolve `pass`), but it's a landmine for the next `assertFile` typo or future `ErrCodeInternal`-returning resolve path.

pkg/health/health.go
```diff
@@ -0,0 +1,260 @@
...
+	// A single shared builder bound to the same provider used for enumeration,
+	// so every combo resolves against one consistent metadata store and carries
+	// the same owner stamp.
+	builder := recipe.NewBuilder(
+		recipe.WithVersion(opts.Version),
+		recipe.WithDataProvider(opts.Provider),
+	)
+
+	report := &Report{SchemaVersion: SchemaVersion}
+	for _, entry := range store.ListCatalog(opts.Filter) {
```
njhensley, line 169: **The loop never checks `ctx.Err()`, so a timed-out run is indistinguishable from a healthy one.**

If `HealthComputeTimeout` (or the parent ctx) fires mid-loop, every remaining `BuildFromCriteria` short-circuits at `builder.go:140` → `ErrCodeTimeout` → graded `unknown`, and `Compute` still returns `(report, nil)`. A truncated run then looks byte-for-byte like a healthy catalog that happens to have transient unknowns — no error, no signal. This also trips the repo's "check `ctx.Done()` in loops" rule.

Suggest failing loud at the top of the loop:

```go
for _, entry := range store.ListCatalog(opts.Filter) {
    if err := ctx.Err(); err != nil {
        return nil, errors.Wrap(errors.ErrCodeTimeout,
            "health computation canceled before completing the catalog", err)
    }
    if !entry.IsLeaf {
        continue
    }
    ...
}
```

pkg/health/health_test.go
```diff
@@ -0,0 +1,212 @@
...
+				t.Errorf("rollup = %q, want %q", got, tt.want)
+			}
+		})
+	}
+}
+
+// TestComputeEmbeddedCatalog resolves the real embedded catalog: every leaf
+// must carry a resolves dimension and a status consistent with the rollup of
+// its dimensions, and at least one leaf must resolve cleanly.
+func TestComputeEmbeddedCatalog(t *testing.T) {
```
njhensley, line 115: **No integration test drives a `fail` (or `unknown`) through the real builder.** This test asserts "≥1 leaf passes" and "status == rollup(dimensions)"; the non-pass grading has only synthetic `classifyResolve` unit coverage. That gap is precisely what would let the over-wide `unknown` bucket (see `isTransient`) ship unnoticed.

Recommend a case that resolves a deliberately-broken leaf — e.g. via a fake provider whose catalog has a component with a missing `assertFile` — and asserts the combo grades `fail`. Pairs naturally with the `isTransient` fix.

njhensley (commented):
Solid, well-documented foundation for ADR-009 — the generic rollup, the descriptor-vs-graded split, byte-deterministic output, and the standalone-package rationale (avoiding the future `recipe → evidence → recipe` import cycle) are all well-judged, and coverage is strong.

I have one correctness concern about the `unknown` classification being too wide (latent today — every current leaf resolves `pass`), one context-handling gap, and a test gap worth closing alongside the first. Details inline.

pkg/health/health.go
```diff
@@ -0,0 +1,260 @@
...
+	})
+
+	return report, nil
+}
+
+// computeCombo scores a single leaf overlay's structural health.
+func computeCombo(ctx context.Context, builder *recipe.Builder, entry recipe.CatalogEntry) ComboHealth {
+	structure := StructureHealth{
+		Dimensions: make(map[string]string),
+		Detail:     make(map[string]string),
```
njhensley, line 199: Nit: `Detail` is always allocated even when no detail is recorded. Harmless — `yaml:",omitempty"` drops the empty map at marshal time — just noting the struct always carries a non-nil empty map. Fine to leave as-is for symmetry with `Dimensions`.

pkg/health/doc.go
```diff
@@ -0,0 +1,58 @@
...
+//
+// The rolled-up per-recipe status is fail if any graded dimension fails,
+// else warn if any warns, else unknown if any is held (transient resolver
+// errors are held rather than penalized — re-run rather than mark fail),
+// else pass. unknown is a representable value, never an empty field a
+// consumer could misread as pass.
+//
+// # Determinism
+//
+// There is no time.Now() on the computed path, so output is a pure function
```
njhensley, line 51: The byte-determinism guarantee is only tested on the clean path; `Detail` carries the raw `err.Error()`, which is deterministic for *currently reachable* errors (slice-ordered cycle paths, static `fs.ErrNotExist`) but could embed map-ordered content from a future error that formats a `map[string]any`. Already acknowledged in the PR body — no action for V1, but worth a regression fixture once more recipes can grade `fail`.

.github/CODEOWNERS
```diff
@@ -21,6 +21,9 @@
 # Data
 recipes/** @nvidia/aicr-maintainer
 
+# Recipe health (ADR-009): recipes/area owners triage the weekly refresh
```
njhensley: Nit: "recipes/area owners" reads a little oddly — suggest "the `area/recipes` owners triage the weekly refresh" for clarity.

pkg/health/health.go
mchmarny: Agreed — narrowed in fe7da04a. Verified your claim against the source: per-combo cancellation/timeout are wrapped `ErrCodeTimeout` (builder.go:140, metadata_store.go:888) and the store-walk cancel path is `ErrCodeTimeout` (metadata_store.go:199), while `ErrCodeInternal` on the resolve path comes from deterministic defects — the missing `assertFile` flattened at metadata_store.go:1226, and `TopologicalSort` at finalizeRecipeResult:859 (though cycles are actually caught earlier as `ErrCodeInvalidRequest` in `detectCycles`). So dropping `ErrCodeInternal` loses no transient coverage.

`isTransient` now holds only context cancellation/deadline + `ErrCodeTimeout`; everything else falls through to `fail`. This intentionally diverges from ADR-009 §2's literal "ErrCodeTimeout/ErrCodeInternal → unknown" to match its stated intent ("transient … re-run rather than penalize" + "fail honestly, never silently pass"). I left a code comment recording the divergence and flagged it here so the ADR §2 wording can be corrected in a follow-up. Unit tests for the flipped cases updated, and there's now an end-to-end `fail` test (below).

pkg/health/health.go
mchmarny: Fixed in fe7da04a — the catalog loop now checks `ctx.Err()` at the top and returns `ErrCodeTimeout` ("health computation canceled before completing the catalog"), so a truncated run fails loud instead of looking like a healthy catalog with transient unknowns. Used your suggested shape.

pkg/health/health_test.go
mchmarny: Added in fe7da04a: `TestComputeGradesDeterministicDefectAsFail`. It builds a minimal in-memory catalog (base + a `broken-leaf` overlay) whose registry component declares `healthCheck.assertFile` pointing at a missing path; the embedded read returns bare `fs.ErrNotExist`, flattened to `ErrCodeInternal` in `finalizeRecipeResult`, and the test asserts the combo grades `fail` (resolves + rolled-up status) with the `assertFile` error in `Detail`. This drives the exact `ErrCodeInternal`-flip through the real builder, so reverting the `isTransient` narrowing would flip it back to `unknown` and fail this test.

pkg/health/health.go
mchmarny: Acknowledged — leaving as-is per your note. `yaml:",omitempty"` drops the empty `Detail` map at marshal time, and the always-non-nil map keeps it symmetric with `Dimensions` and lets callers write without a nil-guard.

pkg/health/doc.go
mchmarny: Agreed, no action for V1. The doc comment already records this caveat; a regression fixture is worth adding once more recipes can grade `fail` and an error can embed map-ordered content. I'll leave it for the signal that introduces that (the `Detail` strings reachable today — slice-ordered cycle paths, static `fs.ErrNotExist` — are deterministic).

.github/CODEOWNERS
mchmarny: Done in fe7da04a — reworded to "the `area/recipes` owners triage the weekly refresh".

njhensley (approved):
Approving — all review comments addressed in the force-push.

- **`isTransient` narrowed** (the High-latent finding): `ErrCodeInternal` now grades `fail`; only `ErrCodeTimeout` + `context.DeadlineExceeded/Canceled` are held as `unknown`. Unit cases flipped accordingly.
- **Loop now fails loud on `ctx.Err()`** instead of emitting a truncated report that looks healthy.
- **New `TestComputeGradesDeterministicDefectAsFail`** drives a real broken leaf (missing `healthCheck.assertFile` → `ErrCodeInternal`) through `Compute` and asserts `fail` end-to-end — exactly the trigger path the bucket fix protects, guarding against regression.
- CODEOWNERS wording fixed.

The two remaining nits (always-allocated `Detail` map, fail-path determinism fixture) are intentionally deferred and fine for V1. Nice, well-tested foundation for ADR-009.

PR #1302, opened 2026-06-11
feat(cli): add structural-health columns to recipe list

### LINKED ISSUE #1228: aicr recipe list: add structural-health + coverage columns

njhensley:
## Summary

Extend `aicr recipe list` with the ADR-009 §4 structural-health view: a rolled-up status column and a compact per-phase coverage summary, delegating all computation to `pkg/health` so `pkg/cli` stays logic-free.

## Motivation / Context

Part of the recipe health tracking epic (ADR-009 V1). #1208 delivered the `recipe list` enumeration command (criteria/leaf/source columns + `--format`/filter flags); this adds the health-derived columns on top, without changing #1208's output shape.

Fixes: #1228
Related: #1224 (epic), #1226/#1227 (health grading dependencies)

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Other: `pkg/client/v1` facade (`Client.ComputeHealth`)

## Implementation Notes

- New facade method `Client.ComputeHealth(ctx, filter)` wraps `health.Compute`, binding the client's own `DataProvider` + version so `--data` overlays are scored against the same catalog the listing enumerates (not the process-global embedded catalog). It follows the sibling facade guard pattern and deliberately omits the per-op timeout since `health.Compute` applies its own `HealthComputeTimeout`.
- **table**: adds `STATUS` (rolled-up verdict) and `COVERAGE` (compact per-phase named-check summary, `R:n D:n P:n C:n`); non-leaf overlays render `-`.
- **json/yaml**: each entry gains an additive `health` block (per-dimension status map + full `declared_coverage`), omitted for non-leaf overlays. The embedded `CatalogEntry` carries `yaml:",inline"` so #1208's fields stay top-level — `yaml.v3` does not auto-inline anonymous struct fields the way `encoding/json` does.
- **Degrades gracefully**: rendering reads the dimension set generically, so an absent `constraints_wellformed` dimension (#1227 not yet merged) doesn't break the status column.

## Testing

```bash
gofmt -l ...                          # clean
golangci-lint run (pinned v2.12.2)    # 0 issues
go test -race ./pkg/cli/... ./pkg/client/v1/...   # pass
```

Coverage: `pkg/cli` 64.0% → 67.7%, `pkg/client/v1` 75.1% → 75.3% (new `ComputeHealth` 81.2%). New tests cover the table/json/yaml formats, the non-leaf `-` placeholder, filtered-CLI pass-through, the empty-result branch, and a YAML-inline regression guard.

This change was reviewed by a multi-persona review pass (Go architecture, CLI/API-contract, test-quality), which caught and fixed a YAML serialization regression before push.

## Risk Assessment

- [x] **Low** — Additive output only; #1208's existing fields/columns are unchanged and guarded by tests. Easy to revert.

**Rollout notes:** Backwards compatible. The `health` block is additive (omitted for non-leaf); existing json/yaml/table consumers see no field removals or renames.

## Checklist


pkg/cli/recipe_list.go
```diff
@@ -128,7 +151,20 @@ Include overlays from an external data directory:
...
 			}
 
-			return writeCatalogEntries(ctx, cmd, entries, format)
+			// Delegate the structural-health computation to pkg/health (via the
+			// facade, which binds this Client's own provider so --data overlays
+			// are scored). Health is keyed by leaf overlay name so each catalog
+			// entry can be paired with its verdict during rendering.
+			report, err := client.ComputeHealth(ctx, filter)
+			if err != nil {
+				return err
```
mchmarny, line 160: nit: the `ListCatalog` sibling two lines up wraps with `errors.PropagateOrWrap(err, errors.ErrCodeInternal, "failed to list catalog")`, while this returns `err` bare. Functionally fine since `ComputeHealth` already wraps internally with `PropagateOrWrap`, so re-wrapping here would double the message — flagging only for symmetry with the adjacent call. No change needed.

pkg/cli/recipe_list.go
```diff
@@ -128,7 +151,20 @@ Include overlays from an external data directory:
 					fmt.Sprintf("unknown output format %q, valid formats are: json, yaml, table", cmd.String(flagFormat)))
 			}
 
-			return writeCatalogEntries(ctx, cmd, entries, format)
+			// Delegate the structural-health computation to pkg/health (via the
+			// facade, which binds this Client's own provider so --data overlays
+			// are scored). Health is keyed by leaf overlay name so each catalog
+			// entry can be paired with its verdict during rendering.
+			report, err := client.ComputeHealth(ctx, filter)
```
mchmarny, line 158: nit: `recipe list` now resolves every leaf on every invocation, so the previously-cheap enumeration always pays the full health-compute cost with no way to opt out. Within ADR-009's sub-minute budget so not blocking — but if the added latency is noticeable interactively, a `--no-health` / `--skip-health` flag would be a reasonable follow-up.

mchmarny (approved):
Clean, additive `recipe list` health columns. The structural-health computation is correctly delegated to `pkg/health` via the new `Client.ComputeHealth` facade method, keeping `pkg/cli` logic-free, and the facade follows the sibling guard/inflight pattern. The leaf-name join with the catalog listing is correct, non-leaf overlays render the `-` placeholder, and the `yaml:",inline"` regression is explicitly guarded by a dedicated test. CI fully green; coverage up on both packages.

Two non-blocking nits inline (error-wrap symmetry, and the always-on health compute on every `list`). Approving.

pkg/cli/recipe_list.go
njhensley: Thanks — agreed, keeping the bare return. `ComputeHealth` already wraps with `PropagateOrWrap(err, ErrCodeInternal, "failed to compute recipe health")`, so re-wrapping here would double the message and is exactly the "don't double-wrap errors that already have proper codes" case in CLAUDE.md. (The adjacent `ListCatalog` call is arguably the asymmetric one, since the facade `ListCatalog` already wraps internally too — but that's pre-existing #1208 code and out of scope here.)

pkg/cli/recipe_list.go
njhensley: Good call-out. Health compute runs within ADR-009's sub-minute budget so it's fine for v1, but an opt-out (`--no-health` / `--skip-health`) is a sensible follow-up if interactive latency becomes noticeable. Tracking it as a follow-up rather than expanding this PR's scope.

njhensley: Filed as #1311 for the follow-up.

PR #1303, opened 2026-06-11
fix(health): exempt manifest-only Helm components from chart_pinned

njhensley:
## Summary

`chart_pinned` flagged manifest-only "Helm" components (typed `Helm` but shipping local `manifestFiles` with no external chart) as unpinned. This exempts them, since there is no chart version to pin.

## Motivation / Context

Running the health generator against current main reported 17 of 32 recipes as `fail` — every one for the same reason:

```
chart_pinned: fail — unpinned Helm chart version for: nodewright-customizations
```

`nodewright-customizations` is a manifest-only component: overlays declare it `type: Helm` but it ships local `manifestFiles` with empty repository/chart/version. The merged `classifyChartPinned` (#1226) flags any enabled Helm ref with an empty `Version` as `fail` and did not exempt components with no external chart to pin — so this was a false-positive class, not a real pinning gap.

This matters because ADR-009 is emphatic that the public matrix must never mislead ("a stale public 'health' surface is worse than none"). Publishing a doc that says half the catalog fails — when it is really one known false-positive class — would do exactly that.

Fixes: N/A
Related: #1226, #1293

## Type of Change

## Component(s) Affected

- [x] Other: recipe health signal (`pkg/health`)

## Implementation Notes

- Added `isManifestOnlyHelm(ref)` — a Helm-typed ref with empty `Chart` and `Source` but non-empty `ManifestFiles`/`PreManifestFiles`. This mirrors the bundler's canonical manifest-only detection (`pkg/bundler/deployer/flux/flux.go`: `ref.Chart == "" && ref.Source == "" && hasManifests`), keeping the health signal aligned with what is actually bundled and deployed.
- `classifyChartPinned` now skips manifest-only refs the same way it skips disabled components: not counted toward `helmCount`, never added to the `unpinned` list.
- Edge case: a recipe whose only enabled Helm component is manifest-only now reports the existing vacuous-pass "not applicable" detail (consistent with the disabled-only precedent). No current recipe hits this — the affected 17 all also carry real pinned charts.

## Testing

```bash
go test ./pkg/health/       # ok
go vet ./pkg/health/...     # clean
gofmt -l pkg/health/*.go    # clean
```

- Added three table cases to `TestClassifyChartPinned`: manifest-only alongside a pinned chart → pass; manifest-only not appearing in the unpinned list; manifest-only-only → vacuous "not applicable" pass. They fail before the change and pass after.
- Verified real-catalog impact with a temporary probe against the embedded catalog (removed afterward): `chart_pinned` went from **17/32 fail → 0/32 fail**.
- `-race` could not run in the dev sandbox (no cgo/gcc) and `golangci-lint` is not on PATH locally; CI covers both.

## Risk Assessment

- [x] **Low** — Isolated change to one scoring function, table-test covered, easy to revert.

**Rollout notes:** N/A — read-only health signal; no API, schema, or deployment behavior changes.

## Checklist

- [ ] Tests pass locally (`make test` with `-race`) — `-race` unavailable in sandbox (no cgo); `go test ./pkg/health/` passes, CI runs `-race`
- [ ] Linter passes (`make lint`) — `golangci-lint` not on PATH locally; `go vet` + `gofmt` clean, CI runs the pinned linter
- [x] I updated docs if user-facing behavior changed (N/A — internal health signal)


mchmarny (dismissed):
Correctly exempts manifest-only "Helm" components from `chart_pinned` — a component typed Helm but shipping local `manifestFiles` with no external chart has no version to pin, so the prior `fail` was a false-positive class (17/32 → 0/32). The `isManifestOnlyHelm` predicate faithfully mirrors the bundler's manifest-only detection (`flux.go`: `Chart == "" && Source == "" && hasManifests`), keeping the health signal aligned with what's actually bundled and deployed. Edge cases hold: a Helm ref with a real chart but empty version is still flagged, and a truly-empty Helm ref still falls through to the unpinned check. Well-tested (three new table cases), isolated, CI green. No findings.

mchmarny (approved):

Issue #1230, opened 2026-06-08
CI: weekly health-refresh bot PR for the recipe-health matrix


njhensley:
**Parent:** #1224 (Epic: Recipe health tracking, ADR-009 V1)

## Summary

Structurally clone `.github/workflows/bom-refresh.yaml` so the public matrix stays current via a weekly, drift-only bot PR (ADR-009 §5).

## Scope

New workflow mirroring `bom-refresh.yaml`:

- **Weekly cron, offset from `bom-refresh`** — `bom-refresh` is `0 6 * * 1`; use e.g. `30 6 * * 1` so two bot PRs don't land simultaneously every Monday and contend for `/ok` re-fire and runner capacity. Plus `workflow_dispatch`.
- Weekly, not daily — V1 health changes only on code/registry/chart merges, not the calendar.
- `permissions: contents: read` globally; the job overrides `contents: write` + `pull-requests: write`.
- Checkout (`persist-credentials: false`), `load-versions`, setup Go. **No Helm setup** — health is render-free (the one deliberate divergence from `bom-refresh`).
- `run: make recipe-health-docs`.
- `peter-evans/create-pull-request` (pin the same SHA as `bom-refresh`): `commit-message`/`title` `chore(health): refresh recipe health matrix`, `branch: chore/health-refresh`, `delete-branch: true`, `labels: documentation / area/docs / area/recipes`, built-in `GITHUB_TOKEN` (the `/ok` reviewer-comment policy re-fires CI on bot PRs).
- Write per-dimension detail to `$GITHUB_STEP_SUMMARY` for at-a-glance maintainer triage.

## Out of scope

PR-time sticky-comment apparatus (deferred); a blocking gate (health is advisory).

## Key files

`.github/workflows/bom-refresh.yaml` (template — note its `GITHUB_TOKEN` + `/ok` comment), `.github/actions/load-versions`.

## Tests

Workflow lint; manual `workflow_dispatch` dry run. CODEOWNERS already routes `/.github/**` to `@nvidia/aicr-maintainer`.

## Acceptance

A drift-only weekly PR opens against `chore/health-refresh`, labeled, with step-summary detail, cron-offset from `bom-refresh`, no Helm/network dependency.

## Dependencies

Depends on #1229 (the `recipe-health-docs` target + committed matrix doc).

Issue #1229, opened 2026-06-08
tools/health generator + make targets + public recipe-health matrix


njhensley:
**Parent:** #1224 (Epic: Recipe health tracking, ADR-009 V1)

## Summary

Mirror the BOM generate-and-PR precedent (ADR-009 §5): a deterministic generator, make targets that splice into a marked region, and the public matrix doc — but **hermetic/offline** (no live Helm render, unlike `tools/bom`).

## Scope

- New `tools/health/main.go` paralleling `tools/bom/main.go`: flags `-deterministic` and `-no-title`, `-out-dir`, `-aicr-version`. Calls `pkg/health.Compute()` directly and renders the Markdown matrix (rolled-up status, `Coverage` column, and an `Evidence` column that is a literal `pending` for every recipe in V1).
- `Makefile` targets cloning `bom-docs`/`bom-check`, **named to avoid the existing `check-health`/`check-health-all`/`component-health` cluster-chainsaw family** (the `health-check` vs `check-health` homophone is a footgun):
  - `make recipe-health-docs` → `go run ./tools/health -deterministic -no-title`, then `awk`-splice into `docs/user/recipe-health.md` between `<!-- BEGIN AICR-HEALTH -->` / `<!-- END AICR-HEALTH -->`.
  - `make recipe-health-check` → run the docs target, fail on `git diff`; **advisory, not wired into `make qualify`** (comment it like `bom-check`).
- **Compute-budget gate:** a measurable assertion that the generator stays under the ADR-009 sub-minute target (e.g. a `go test` timing ceiling on `Compute()` over the full catalog), referenced by the epic acceptance criteria.
- **Committed-golden determinism:** the staleness check is a regenerate-and-`git diff` against the committed `recipe-health.md` — the byte-stability backstop complementing #1225's in-process marshal test.
- `docs/user/recipe-health.md`: hand-written header (incl. a note on the uniform `pending` Evidence column and the layer-1-only `chart_pinned` meaning) + the marked region.
- Register in `docs/index.yml` under **User Guide** (after Component Catalog); cross-link from `docs/user/component-catalog.md`.
- `CODEOWNERS`: add `tools/health/**` and `docs/user/recipe-health.md` (→ `@nvidia/aicr-maintainer`). (`pkg/health/**` is pinned to #1225.)

## Out of scope

Committed `recipe-health.json`; committed internal detail doc (the weekly job writes per-dimension detail to `$GITHUB_STEP_SUMMARY` — #1230).

[...]

PR #1304, opened 2026-06-11
feat(health): add tools/health generator and recipe-health matrix

njhensley:
## Summary

Add a hermetic, offline recipe-health generator (`tools/health`) plus the publication machinery (make targets, public doc, nav/ownership) that surfaces per-recipe structural health across the whole criteria matrix — mirroring the BOM generate-and-PR precedent, but with no live Helm render. The matrix body is intentionally deferred (see Implementation Notes); the tooling itself is complete and functional.

## Motivation / Context

ADR-009 (Recipe Health Tracking, V1) calls for a hermetic generator, BOM-style make targets that splice into a marked region, and a public matrix doc. This PR delivers the `tools/health` generator + targets + doc; the validation logic it consumes (`pkg/health.Compute`) already landed via #1226/#1225.

Fixes: N/A — implements the #1229 tooling; public-matrix *publication* is deferred (see Implementation Notes), so I left #1229 open for the maintainer to close after the matrix publishes.
Related: #1229, #1224 (epic), #1226/#1227 (deps), #1230 (blocked: weekly refresh)

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tools/health` generator + `Makefile` targets + `tools/check-docs-mdx`

## Implementation Notes

- **Generator (`tools/health`)** parallels `tools/bom` *minus the network render*: it calls `pkg/health.Compute()` against the embedded catalog (`Provider: nil`) and renders a Markdown matrix — per-recipe rolled-up status, a declared-coverage column (`R:n D:n P:n C:n`), and a literal `pending` Evidence column. Flags `-deterministic`/`-no-title` make the output byte-stable and embeddable.
- **Targets** clone `bom-docs`/`bom-check`: `make recipe-health-docs` splices the body between `<!-- BEGIN/END AICR-HEALTH -->`; `make recipe-health-check` is advisory (not wired into `make qualify`/lint/merge gate). Named `recipe-health-*`, not the ADR's original `health-*`, to avoid the `check-health`/`check-health-all`/`component-health` chainsaw family (homophone footgun); ADR-009 §5 updated to match.
- **Matrix publication is intentionally deferred** behind an `AICR-HEALTH-DEFERRED` sentinel. `pkg/health`'s `chart_pinned` signal currently over-reports `fail` for manifest-only Helm components (declared `type: Helm` but shipping local manifests with no external chart to pin), which would make 17/32 recipes show `fail` and render a misleading public matrix. The generator/targets are fully functional; `recipe-health-check` skips while the sentinel is present. Publishing is a follow-up gated on fixing `classifyChartPinned` (a `pkg/health` change, CODEOWNERS-pinned to the #1225/#1226 owners).
- **#1227 forward-compat:** the renderer reads only the rolled-up `Status` (+ coverage), never per-dimension keys, and `pkg/health.rollup()` is generic over the dimension map — so `constraints_wellformed` folds in with zero code change, just a doc regen.
- `docs/user/recipe-health.md` excluded from the MDX check (carries splice-marker HTML comments, exactly like `container-images.md`).

## Testing

```bash
make lint                                   # golangci-lint 0 issues, MDX, license, agents-sync, bom-pinning
go test -race ./tools/health/...            # pass (CGO race)
go test -coverprofile ./tools/health/...    # 72.1%
make recipe-health-docs                     # splices the 32-row matrix cleanly (verified, then reverted to placeholder)
make recipe-health-check                    # skips honestly while deferral sentinel present (exit 0)
```

Note: I ran `make lint` + race/coverage on `tools/health` + the target/splice behavior, not the full `make qualify` (e2e/scan) — CI covers those. New package coverage 72.1% (above the 70% floor); no pre-existing package's coverage changed (no edits outside `tools/health` + docs/build files).

## Risk Assessment

- [x] **Low** — Isolated, additive tooling; no CLI/server/runtime path touched. The advisory check is unwired from any gate, and the matrix body is a placeholder, so nothing here can turn CI red or affect deployments. Trivially revertible.

**Rollout notes:** No migration. The public matrix stays a placeholder until a follow-up fixes the `chart_pinned` false-positive and runs `make recipe-health-docs`.

## Checklist


tools/health/markdown.go
```diff
@@ -0,0 +1,165 @@
...
+func coverageCell(cov *health.DeclaredCoverage) string {
+	if cov == nil {
+		return "—"
+	}
+	return fmt.Sprintf("R:%d D:%d P:%d C:%d",
+		len(cov.Readiness.Checks),
+		len(cov.Deployment.Checks),
+		len(cov.Performance.Checks),
+		len(cov.Conformance.Checks),
+	)
```
mchmarny, line 164: nit: this `R:n D:n P:n C:n` formatter is identical to `compactCoverage` in #1302's `pkg/cli/recipe_list.go` — both render the same string from a `*health.DeclaredCoverage`. Once both land, that's duplicated format logic in two packages that has to be kept in sync. Consider lifting it to a shared `pkg/health` helper (e.g. a `DeclaredCoverage.Compact()` method) so the CLI table and this matrix can't drift. Not blocking.

mchmarny (dismissed):
Hermetic, offline `tools/health` generator + BOM-style splice targets + public matrix doc, faithfully mirroring the `tools/bom` precedent minus the network render. Determinism and error handling are solid and well-covered: `run` checks both render and `Close` errors with the right priority, the `stickyWriter` collects the first write failure, and byte-stability is tested directly. Deferring the matrix body behind a documented `AICR-HEALTH-DEFERRED` sentinel — pending the `chart_pinned` false-positive fix — is the right call to avoid publishing a misleading surface, and the `recipe-health-check` skip handles it honestly. The `check-docs-mdx` array-exclude conversion is correct. One non-blocking nit inline on duplicated coverage formatting. Approving.

tools/health/markdown.go
njhensley: Good catch — agreed they should not drift. Holding the consolidation as a follow-up rather than doing it here, for two reasons: (1) #1302 (`pkg/cli/recipe_list.go`) is still open, so neither formatter is settled yet; and (2) lifting to `DeclaredCoverage.Compact()` lands in `pkg/health`, which is CODEOWNERS-pinned to the #1225/#1226 owners and explicitly out of scope for #1229. Filed as #1312: whichever of #1302 / #1304 merges second adds the `Compact()` helper and repoints both call sites.

mchmarny (approved):

PR #1314, opened 2026-06-11
feat(cli): add --no-health opt-out to recipe list

### LINKED ISSUE #1311: [Feature]: add --no-health opt-out to aicr recipe list
### Prerequisites

### Feature Summary

Add an opt-out flag (`--no-health` / `--skip-health`) to `aicr recipe list` that skips the per-leaf structural-health computation and renders only the #1208 enumeration columns/fields.

### Problem / Use Case

As of #1228, `aicr recipe list` resolves every leaf overlay through `pkg/health.Compute` on **every** invocation to populate the `STATUS`/`COVERAGE` columns (table) and the `health` block (json/yaml). The previously-cheap catalog enumeration now always pays the full health-compute cost, with no way to opt out.

This is within ADR-009's sub-minute budget so it's acceptable for v1, but for purely interactive "what overlays exist?" lookups — or scripted/CI callers that only consume `name`/`criteria`/`is_leaf`/`source` — the health resolution is wasted latency.

Raised as a non-blocking follow-up in PR #1302 review (https://github.com/NVIDIA/aicr/pull/1302).

### Proposed Solution

Add a boolean flag to the `recipe list` command:

- `--no-health` (or `--skip-health`): when set, skip the `client.ComputeHealth(...)` call entirely and render the original #1208 output:
  - **table**: omit the `STATUS` and `COVERAGE` columns (or render `-`).
  - **json/yaml**: omit the `health` block (already `omitempty`, so passing a nil/empty health map yields the pre-#1228 shape).

Wiring is minimal: gate the `ComputeHealth` call on the flag and pass an empty `healthByName` map to `writeCatalogEntries` when skipped. `pkg/cli` stays logic-free; no `pkg/health` or facade changes needed.

### Success Criteria

- `aicr recipe list --no-health` renders the #1208 columns/fields only, with no health computation performed (verifiable: no per-leaf resolve cost).
- Default behavior (no flag) is unchanged from #1228 — health columns/block present.
- Documented in `docs/user/cli-reference.md` (flag table + a note).
- Table/json/yaml covered by tests, including that the `health` block is absent in json when `--no-health` is set.

### Alternatives Considered

- **Always compute (status quo):** simplest, but no escape hatch for latency-sensitive callers.
- **Lazy/cached health:** more complex; caching across invocations is out of scope for a CLI that re-loads the catalog each run.
- **`--columns` selector:** more general but heavier; a single opt-out flag is the minimal fix for the raised concern.

### Component

CLI (aicr)

njhensley:
## Summary

Add a `--no-health` (alias `--skip-health`) opt-out flag to `aicr recipe list` that skips the per-leaf structural-health computation and renders only the #1208 enumeration columns/fields.

## Motivation / Context

As of #1228, `aicr recipe list` resolves every leaf overlay through `pkg/health.Compute` (via `client.ComputeHealth`) on **every** invocation to populate the `STATUS`/`COVERAGE` columns (table) and the `health` block (json/yaml). The previously-cheap catalog enumeration now always pays the full health-compute cost, with no way to opt out. For purely interactive "what overlays exist?" lookups — or scripted/CI callers that only consume `name`/`criteria`/`is_leaf`/`source` — the health resolution is wasted latency.

Raised as a non-blocking follow-up in PR #1302 review.

Fixes: #1311
Related: #1228, #1302

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- The `client.ComputeHealth(...)` call is gated behind `showHealth := !cmd.Bool(flagNoHealth)` and skipped **entirely** when the flag is set — a real latency win, not just a column drop.
- When skipped, `healthByName` stays `nil`: the table emits the pre-#1228 8-column header (no `STATUS`/`COVERAGE`), and json/yaml omit the `health` block via the existing `omitempty` (reading a nil map yields the zero-value pointer, so each entry's `Health` is nil).
- Gating lives in `pkg/cli` only; no `pkg/health` or facade changes. Default behavior (no flag) is unchanged from #1228.

## Testing

```bash
go test ./pkg/cli/...          # pass
golangci-lint run -c .golangci.yaml ./pkg/cli/...   # 0 issues (pinned v2.12.2)
```

New tests cover the flag across table/json/yaml, the `--skip-health` alias (byte-identical output), and the empty-result + `--no-health` intersection — including that the `health` block is absent in json/yaml when set. The YAML test decodes and asserts no `health` key (rather than a substring scan) plus a leaf-present guard.

`pkg/cli` coverage: 68.7% → 68.7% (flat, no regression). No new exported functions.

## Risk Assessment

- [x] **Low** — Isolated, additive CLI flag; default path unchanged; easy to revert.

**Rollout notes:** Backwards compatible. The flag defaults to `false`, so existing invocations and scripts see the #1228 output unchanged.

## Checklist

### DIFF
--- a/pkg/cli/recipe_list.go
+++ b/pkg/cli/recipe_list.go
@@ -116,6 +116,12 @@ Include overlays from an external data directory:
 				Usage:    "Output format (json, yaml, table)",
 				Category: catOutput,
 			}, func() []string { return []string{"json", "yaml", "table"} }),
+			&cli.BoolFlag{
+				Name:     flagNoHealth,
+				Aliases:  []string{"skip-health"},
+				Usage:    "Skip per-leaf structural-health computation. Omits the STATUS/COVERAGE table columns and the health block in json/yaml output.",
+				Category: catOutput,
+			},
 		},
 		Action: func(ctx context.Context, cmd *cli.Command) error {
 			if err := validateSingleValueFlags(cmd, flagService, flagAccelerator, flagIntent, flagOS, flagPlatform, flagFormat); err != nil {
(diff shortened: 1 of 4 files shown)


lockwobr (approved):

pkg/cli/recipe_list.go
```diff
@@ -151,20 +157,30 @@ Include overlays from an external data directory:
 					fmt.Sprintf("unknown output format %q, valid formats are: json, yaml, table", cmd.String(flagFormat)))
 			}
```
mchmarny, line 158: nit: `pre-#1228` will rot — PR references in code comments belong in the commit/PR description. Drop the issue number; the behaviour description ("no STATUS/COVERAGE columns, no health block") stands on its own.

mchmarny (approved):
Clean implementation. The `showHealth` gate is in exactly the right place (CLI layer, facade untouched), nil `healthByName` correctly signals both the table renderer and the json/yaml `omitempty` path, and test coverage is thorough (table, JSON, YAML, alias, empty-result intersection). One nit on a PR reference in a code comment; same pattern appears in the `writeCatalogEntries` godoc at ~line 254 — both can drop the `#1228` number. No correctness issues; good to merge once CI clears.

PR #1324, opened 2026-06-11
fix(validator): bound absent-resource retries in health checks of deployment validation

### LINKED ISSUE #1323: Deployment validator stalls ~8m and reports 'other' on absent resources
## Summary

The `expected-resources` deployment validator stalls for the full Job
deadline (~8m) and reports an opaque `status=other` (instead of a clean
`failed`) whenever a recipe-declared component is **absent or in a different
namespace** than the recipe expects. This was observed validating a live
GB300 EKS cluster managed by an external platform (components deployed by the
platform's GitOps into different namespaces than the recipe assumes, plus one
component intentionally removed).

## Symptoms

- `expected-resources` runs to ~8m24s and ends as `status=other`
  (phase: `passed=3 failed=0`), so the operator gets **no actionable verdict**.
- Per-component failures surface a misleading error:
  `client rate limiter Wait returned an error: context deadline exceeded`,
  which looks like a client bug but is really the per-check deadline firing on
  an in-flight GET.
- Healthy components are also delayed to the deadline (the check returns nothing
  useful until it is killed).

## Root cause

1. **Absent resources retry until the full readiness budget.** A resource that
   does not exist (`ErrCodeNotFound`) is retried every `AssertRetryInterval`
   until `ChainsawAssertTimeout` (6m). With `ChainsawMaxParallel = 4` worker
   slots, a few permanently-absent components occupy the pool for the full 6m
   and **starve the healthy components** queued behind them. The check then
   brushes the Job's `activeDeadlineSeconds`, the pod is killed, and a
   non-0/1/2 exit code maps to `other`.
2. **The real reason is lost at the deadline.** On context cancellation the
   retry loops return an `ErrCodeInternal` "context canceled" wrap (or the
   in-flight client-go rate-limiter error), masking the substantive assertion
   reason (e.g. "DaemonSet not found").
3. **(Minor) client QPS is the client-go default (5 QPS / 10 burst)** — too low
   for a one-shot enumeration across 18+ components, which throttles even the
   all-healthy sweep.

## Proposed fixes

1. **Bound only the absent (NotFound) case** with a short
   `AbsentResourceGracePeriod` (~30s): a resource that does not exist will not
   appear by waiting out the readiness budget. A resource that **exists but is
   not ready** returns a shape-mismatch error (`ErrCodeInternal`), and a
   transient API failure returns `ErrCodeUnavailable` — **both keep the full
   6m**, so slow-but-healthy rollouts are not failed prematurely. (Verified:
   shape mismatch and transient errors are distinct codes from NotFound, so the
   gate is precise.)
2. **Preserve the substantive assertion error at the deadline** so the verdict
   is a clean `failed` with the real reason ("... not found"), not an opaque
   `other` driven by the context/rate-limiter artifact.
3. **Raise the validator REST client limits** (e.g. QPS 50 / Burst 100) so the
   healthy enumeration sweep is not throttled. apiserver APF still protects the
   server.

## Evidence (prototyped + verified on a live GB300 EKS cluster)

| Metric | Before | After (fixes 1+2+3) |
|---|---|---|
| `expected-resources` status | `other` | `failed` |
| Phase result | `passed=3 failed=0` (other) | `passed=3 failed=1` |
| Duration | 8m24s | 1m28s (~5.7x faster) |
| Failure reason | `client rate limiter Wait ... context deadline exceeded` | `Nodewright tuning: not found (recipe declared it but the cluster has no such CR)` |

Healthy components still pass; not-ready/slow-rollout components still get the
full 6m (no false-failure regression).

## Affected code

- `validators/chainsaw/inprocess.go` — `runAssertWithRetry` / `runErrorWithRetry`
- `validators/chainsaw/runner.go` — `assertRawResources`
- `pkg/defaults/timeouts.go` — `ChainsawAssertTimeout`, `ChainsawMaxParallel`,
  `AssertRetryInterval` (+ proposed `AbsentResourceGracePeriod`)
- `pkg/k8s/client/client.go` / `validators/context.go` — REST client QPS/Burst

A working prototype of all three fixes (with unit tests) has been validated
end-to-end on a live GB300 EKS cluster; a PR can follow.

yuanchen8911:
## Summary

Fix the deployment validator's `expected-resources` check so it fails **fast and cleanly** when a recipe-declared resource is absent or in a different namespace than the recipe expects — instead of retrying to the Job deadline and reporting an opaque `other`.

Fixes #1323

## Problem

On a cluster where a recipe component is absent or misplaced, `expected-resources` retried the missing resource for the full `ChainsawAssertTimeout` (6m). With `ChainsawMaxParallel = 4` worker slots, a few absent components starved the healthy ones and pushed the check past the Job's `activeDeadlineSeconds`; the pod was killed and the result surfaced as `status=other` (~8m) with a misleading `client rate limiter Wait … context deadline exceeded` message rather than an actionable failure.

## Changes

- **Absent-resource grace** (`pkg/defaults.AbsentResourceGracePeriod = 30s`): bound only the entirely-absent (`ErrCodeNotFound`) case. A not-ready resource (shape mismatch → `ErrCodeInternal`) and a transient API failure (`ErrCodeUnavailable`) keep the full readiness budget, so slow-but-healthy rollouts are not failed prematurely.
- **Stateful grace**: the grace is disabled once the resource is actually observed (a shape-mismatch `ErrCodeInternal`), so a later transient `NotFound` (e.g. a pod recreate mid-rollout) keeps the full deadline. A transient `ErrCodeUnavailable` does **not** latch (it proves nothing about existence), so a clean `NotFound` after an API blip still gets the fast-fail.
- **Preserve the substantive reason at the deadline**: the retry loops surface the last real assertion error (e.g. "resource not found"), not the `ErrCodeInternal` context-cancellation/rate-limiter artifact — so the verdict is a clean `failed` with the real cause, not an opaque `other`.
- **Align the raw-resource path**: `assertSingleDocument` now propagates the fetcher's structured code (`NotFound`/`Unavailable`) instead of blanket-wrapping as `ErrCodeInternal`, so the grace/latch logic works identically in both the in-process and raw paths.
- **Raise validator REST client limits** (`QPS 50 / Burst 100`) above the client-go defaults so the multi-component enumeration sweep is not throttled. apiserver APF still protects the server.

## Tests

`preferSubstantiveErr`, `notFoundGraceDeadline` (incl. the recreate case), `resourceObservedErr` (incl. `Unavailable` does-not-latch), `assertSingleDocument` propagates the structured code, and a canceled-context retry that asserts the substantive error is surfaced. Build, lint, and `-race` tests pass.

## Verified on live clusters

| Cluster | Topology | Result |
|---|---|---|
| Managed RTX Pro 6000 (components match recipe) | healthy | **all-pass, 28s** |
| GB300 (components in different namespaces + a component removed) | divergent | **`failed` in 1m26s**, clean reason (`Nodewright tuning: not found …`) |

Same divergent GB300 cluster, before vs after: **8m24s → 1m26s**, `other` → `failed` with the real cause.

## Risk

- [x] **Low** — behavior change scoped to the absent (`NotFound`) case; not-ready and transient paths keep their full timeout. Easy to revert.

### DIFF
--- a/pkg/defaults/k8s.go
+++ b/pkg/defaults/k8s.go
@@ -1,0 +1,42 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package defaults
+
+// Kubernetes REST client rate limits for validator containers.
+//
+// client-go's default client-side rate limiter (QPS 5, Burst 10) is tuned for
+// controllers issuing a steady trickle of requests, not for a one-shot check
+// that enumerates every component in a recipe. The expected-resources
+// deployment validator, for example, runs health-check assertions across 18+
+// components, each issuing several GETs (Deployments, DaemonSets, custom
+// resources) within a single per-check context deadline. At 5 QPS the limiter
+// queues the later requests until the deadline elapses, surfacing as
+// "client rate limiter Wait returned an error: context deadline exceeded" and
+// flipping an otherwise-healthy check to status=other even though the cluster
+// is fine.
+//
+// Raising the limits lets these enumeration-heavy checks complete within their
+// deadline. The apiserver's own API Priority and Fairness (APF) still protects
+// the server from overload, so a higher client-side ceiling is safe.
+const (
+	// ValidatorClientQPS is the steady-state requests-per-second ceiling for a
+	// validator container's Kubernetes REST client.
+	ValidatorClientQPS = 50
+
+	// ValidatorClientBurst is the burst capacity for a validator container's
+	// Kubernetes REST client. Sized above QPS so the initial resource sweep
+	// (discovery + the first batch of GETs) is not immediately throttled.
+	ValidatorClientBurst = 100
+)
--- a/pkg/defaults/timeouts.go
+++ b/pkg/defaults/timeouts.go
@@ -323,6 +323,24 @@ const (
 	// they pass or the ChainsawAssertTimeout expires.
 	AssertRetryInterval = 5 * time.Second
 
+	// AbsentResourceGracePeriod bounds how long a health-check assertion
+	// retries a resource that does not exist at all (the fetch returns
+	// ErrCodeNotFound). A resource that is missing entirely — wrong
+	// namespace, never installed — will not appear by waiting out the full
+	// ChainsawAssertTimeout, and retrying it for minutes holds one of the
+	// ChainsawMaxParallel worker slots and starves healthy components behind
+	// it (and can push the check past the Job's activeDeadlineSeconds, which
+	// surfaces as an opaque "other" status instead of a clean failure).
+	//
+	// This grace bounds ONLY the entirely-absent (NotFound) case. A resource
+	// that EXISTS but is not yet ready returns a shape-mismatch error
+	// (ErrCodeInternal), and a transient API failure returns
+	// ErrCodeUnavailable — both keep the full ChainsawAssertTimeout so slow
+	// but healthy rollouts are not failed prematurely. The grace allows brief
+	// creation lag (a resource that appears within the window switches to the
+	// full readiness budget) while failing permanently-absent resources fast.
+	AbsentResourceGracePeriod = 30 * time.Second
+
 	// JobEnvelopeMargin is the headroom added on top of ChainsawAssertTimeout
 	// when computing the validator Job's outer activeDeadlineSeconds and the
 	// expected-resources catalog timeout. Chainsaw needs time after the inner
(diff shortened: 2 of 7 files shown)


yuanchen8911:
@coderabbitai Addressed in 2388e4d6.

Good catch on the grace window being anchored too early. Made the grace **stateful**: a `sawResource` flag is set once any non-NotFound result is observed (resource exists / not-ready / transient error), and `notFoundGraceDeadline(err, sawResource, absentDeadline, deadline)` returns the full `deadline` whenever `sawResource` is true. So the short `AbsentResourceGracePeriod` now only bounds a resource that has been NotFound from the start; a later transient NotFound (e.g. a pod recreate mid-rollout, after the resource had already appeared) keeps the full readiness budget instead of fast-failing on the stale cutoff.

Applied the same pattern in both `validators/chainsaw/inprocess.go` (`runAssertWithRetry`) and `validators/chainsaw/runner.go` (`assertRawResources`). Added a `notFoundGraceDeadline` unit case for the recreate flow (`sawResource=true` + NotFound → full deadline), plus tests for `preferSubstantiveErr` and a canceled-context retry that asserts the substantive error is surfaced.

yuanchen8911:
@coderabbitai The raw-resource path finding (assertSingleDocument collapsing Fetch errors to ErrCodeInternal) is fixed in d9a91e29.

`assertSingleDocument` now propagates the fetcher's structured code as-is — `ErrCodeNotFound` (absent) and `ErrCodeUnavailable` (transient) are no longer blanket-wrapped — and `ErrCodeInternal` is reserved for an actual shape mismatch (the `checks.Check` error list) and a malformed assertion is returned as `ErrCodeInvalidRequest`. So the raw path now matches `evaluateAssert`: the absent-resource grace fast-fail fires and `resourceObservedErr` distinguishes "exists but not ready" from "absent / transient" correctly in both paths.

Added `TestAssertSingleDocument_PropagatesStructuredCode` (absent ⇒ NotFound propagated, not Internal) to guard it. Also fixed the stale `notFoundGraceDeadline` godoc.

mchmarny (approved):

PR #1350, opened 2026-06-13
fix(fingerprint): match GPU SKUs on token boundaries not substrings

### LINKED ISSUE #1349: fix: GPU SKU classifier substring match silently mismaps SKUs
## Summary

The GPU SKU classifier (`ParseGPUSKU`) identifies a node's accelerator by **substring-matching** `nvidia-smi`'s `ProductName` against a pattern table. Because it uses `strings.Contains`, any SKU whose product name contains a shorter registered pattern is silently collapsed onto the wrong accelerator. This defeats the classifier's own `unknown-sku` safety signal and produces fingerprints that misrepresent the hardware.

- Classifier: [`pkg/fingerprint/gpu_sku.go#L48`](https://github.com/NVIDIA/aicr/blob/bbd08495558ab6d3fe1fccec559743db6569fb8c/pkg/fingerprint/gpu_sku.go#L48)
- Pattern table: [`pkg/fingerprint/gpu_sku.go#L24-L41`](https://github.com/NVIDIA/aicr/blob/bbd08495558ab6d3fe1fccec559743db6569fb8c/pkg/fingerprint/gpu_sku.go#L24-L41)
- `unknown-sku` fallback this defeats: [`pkg/fingerprint/from_measurements.go#L186-L194`](https://github.com/NVIDIA/aicr/blob/bbd08495558ab6d3fe1fccec559743db6569fb8c/pkg/fingerprint/from_measurements.go#L186-L194)

## Current behavior / evidence

On an Oracle OKE cluster with `nvidia.com/gpu.product=NVIDIA-L40S`, the snapshot fingerprint records:

```yaml
fingerprint:
  accelerator:
    source: nodeTopology.label.nvidia.com/gpu.product
    value: l40        # the trailing "S" is dropped
```

`NVIDIA-L40S` contains the substring `L40`, so it matches `{"L40", "l40"}` and is reported as a (valid-but-wrong) `l40` instead of being flagged `unknown-sku`.

This is a **class of bug**, not a one-off. The table already carries two hand-patched workarounds for the same root cause:

- `{"GB200", ...}` is ordered before `{"B200", ...}` so a Grace-Blackwell node isn't read as `b200`.
- `{"GH200", ""}` is an explicit exclusion so Grace-Hopper isn't read as `h200`.

Each new collision currently requires another manual ordering/exclusion entry. Known/likely live collisions that are **not** yet patched:

- `NVIDIA RTX A1000` → `"A1000"` contains `"A100"` → silently fingerprints as datacenter `a100`.
- `NVIDIA L40S` → `l40` (the reported case).

## Proposed approach

Replace substring matching with **token-boundary matching** in `ParseGPUSKU`: tokenize the product name (split on whitespace and `-`) and match whole tokens, with contiguous-token-sequence matching for multi-word patterns (`RTX PRO 6000`). Unrecognized tokens fall through to `unknown-sku` (the honest signal).

This structurally eliminates the entire collision class and removes the need for per-SKU ordering/exclusion hacks:

- `NVIDIA L40S` → token `L40S` (not `L40`) → no false match → `unknown-sku`.
- `NVIDIA RTX A1000` → token `A1000` (not `A100`) → no false `a100`.
- `NVIDIA A100-SXM4-40GB` → token `A100` → `a100`. ✓
- `NVIDIA GB200 NVL72` → token `GB200` → `gb200`; `B200` is no longer a substring concern, so the **GB200-before-B200 ordering hack and the GH200 exclusion become unnecessary**.

## Acceptance criteria

- [ ] `ParseGPUSKU` matches on token boundaries (handles space and `-` delimiters; supports the multi-word `RTX PRO 6000` pattern).
- [ ] Unrecognized product strings return `""` so the fingerprint records `unknown-sku` (raw model already preserved in measurements).
- [ ] Ordering/exclusion workarounds removed where token matching makes them redundant (GB200/B200, GH200), with tests proving the behavior still holds.
- [ ] Test matrix asserts **decoys do not false-map**: `RTX A1000`, `L40S`, `L4`, `A800`, `H800`, `GH200`, plus all currently supported SKUs continue to resolve correctly.

## Notes / related

- Making `l40s` (or any newly distinguished SKU) a **first-class, selectable accelerator** is a separate, larger change (criteria enum, OpenAPI, CLI/API, docs) and should land **together with actual recipe coverage** for that SKU — otherwise selecting it just yields "no matching recipe." This issue is only about fixing the misclassification, not adding new enum values.

mchmarny:
## Summary

Fix the GPU SKU classifier so it matches accelerator product names on whole-token boundaries instead of raw substrings, eliminating silent misclassification (e.g. L40S read as `l40`, RTX A1000 read as `a100`).

## Motivation / Context

`ParseGPUSKU` matched `nvidia-smi`/NFD product names with `strings.Contains`, so any SKU whose name contained a shorter registered pattern was collapsed onto the wrong accelerator — and it bypassed the `unknown-sku` fallback that surfaces registry gaps. This surfaced on an Oracle OKE cluster where `nvidia.com/gpu.product=NVIDIA-L40S` fingerprinted as `l40`. The same trap silently mislabels `RTX A1000` as datacenter `a100`. The pattern table already carried two hand-patched workarounds for the identical root cause (GB200-before-B200 ordering; GH200 exclusion).

Fixes: #1349
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Collectors / snapshotter (`pkg/fingerprint` — snapshot SKU classification)

## Implementation Notes

- `ParseGPUSKU` now upper-cases and tokenizes the product name on whitespace **and** hyphens, then matches registry entries as contiguous token sequences. Hyphen splitting normalizes nvidia-smi (`NVIDIA A100 80GB`) and NFD label (`NVIDIA-A100-80GB`) forms identically.
- Because tokens match exactly, entry order is no longer significant, so the **GB200-before-B200 ordering** and the **GH200 exclusion** workarounds are removed. Unrecognized SKUs resolve to `""`, preserving the `unknown-sku` signal.
- No enum/API/recipe surface changed. Adding a SKU as a *first-class selectable accelerator* (criteria enum, OpenAPI, docs) is intentionally out of scope and should land with actual recipe coverage to avoid a "no matching recipe" footgun.

## Testing

```bash
make test   # PASS — total coverage 77.1% (gpu_sku.go funcs 100%)
make lint   # PASS — 0 issues; AGENTS.md in sync
```

`TestParseGPUSKU` expanded to 70 cases: every supported SKU across its real nvidia-smi and NFD-label product-name variants (H100/H200/A100/B200/GB200/L40/RTX PRO 6000), plus an exhaustive decoy set proving look-alikes do **not** misclassify — GH200, L40S, L4/L20/L2, A800/A40/A30/A16/A10/A2, H800/H20, B100, RTX A1000/A2000/A4000/A5000/A6000, RTX 6000 Ada, Quadro RTX 6000, GeForce RTX 4090/5090, V100/T4/P100, Grace.

Note: `make qualify`'s e2e (KWOK/kind) and grype scan stages were not run locally (they don't exercise this code path and require cluster/tooling); CI covers them.

## Risk Assessment

- [x] **Low** — Isolated change to one pure-Go function, exhaustively unit-tested, easy to revert.

**Rollout notes:** Snapshots of unsupported SKUs (e.g. L40S) now correctly fingerprint as `unknown-sku` rather than a wrong-but-valid accelerator. No previously-supported SKU changes classification (covered by tests). N/A migration.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — no user-facing surface changed)


ArangoGutierrez (commented):
Nice fix — token matching is the right call, and it lets you delete both existing workarounds (GB200-before-B200 ordering, GH200 exclusion) instead of adding a third. The behavior reads correctly: single-token SKUs match exactly, RTX PRO 6000 matches as a contiguous run, and the look-alikes fall through to "".

**Code review**
- Correctness looks right. `tokenizeProductName` splitting on whitespace and hyphens normalizes the nvidia-smi and NFD-label forms to identical tokens, and `containsTokenSequence`'s bounds are correct.
- The decoy half of the test table is what makes this trustworthy: L40S, RTX A1000, GH200, B100 all assert "", so the guard discriminates rather than just confirming the happy path. Nice that both the product-name and NFD-label forms are covered.

**Go review**
- Pure function, no error or concurrency surface, nothing for the linter. The per-call allocation in `tokenizeProductName` runs once per node, so no concern.
- One nit inline: the registry is order-independent now.

No blockers. The unknown-sku fallback is intact and it's a clean revert if anything regresses.

pkg/fingerprint/gpu_sku.go
```diff
@@ -14,46 +14,83 @@
...
-	// distinct Grace-Hopper SKU, not the discrete H200 GPU. It has no
-	// first-class accelerator enum, so match it explicitly before the "H200"
-	// rule and leave it unresolved ("") rather than mislabeling it as h200 —
-	// same reason "GB200" precedes "B200" above.
-	{"GH200", ""},
-	{"H200", "h200"},
-	{"A100", "a100"},
-	{"RTX PRO 6000", "rtx-pro-6000"},
-	{"L40", "l40"},
+	{[]string{"GB200"}, "gb200"},
```
ArangoGutierrez, line 38: Order no longer matters now that matching is exact-token (the doc comment says as much) — GB200 vs B200 here is cosmetic, since GB200 is its own token and never collides with B200. That's also why dropping the old GH200 exclusion is safe: GH200 tokenizes distinct from H200, locked in by the decoy test. Flagging only so nobody later "restores" a perceived ordering dependency.

pkg/fingerprint/gpu_sku_test.go
```diff
@@ -22,26 +22,102 @@ func TestParseGPUSKU(t *testing.T) {
...
+
+		// --- Collision guards: distinct SKUs sharing a substring with a ---
+		// --- supported one must NOT be misclassified (the bug this fixes) ---
+		// GH200 (Grace Hopper) vs H200 / H100.
+		{"GH200 480GB", "NVIDIA GH200 480GB", ""},
+		{"GH200 bare", "NVIDIA GH200", ""},
+		{"GH200 144GB HBM3e", "NVIDIA GH200 144GB HBM3e", ""},
+		{"GH200 label", "NVIDIA-GH200-480GB", ""},
+		// L40S / L4 / L20 / L2 vs L40.
+		{"L40S", "NVIDIA L40S", ""},
```
ArangoGutierrez, line 74: This decoy block is the part that makes the fix real — each of these mis-resolved under the old `strings.Contains` (L40S->l40, RTX A1000->a100). Asserting "" here means the test goes red if anyone reintroduces substring matching, which is exactly the regression you want guarded.

pkg/fingerprint/gpu_sku.go
mchmarny: Good catch — the test comment on line 43 said "must win over the B200 token in the same family", which implied ordering still mattered. Fixed in 2e1a02a: changed to "distinct token; no ordering dependency — 'GB200' ≠ 'B200' under exact-token matching" to match the registry doc comment's intent.

pkg/fingerprint/gpu_sku_test.go
mchmarny: Exactly right — those `""` assertions are the regression gate. If anyone reintroduces `strings.Contains`, L40S→l40 and RTX A1000→a100 go red immediately. No code change needed here.

ArangoGutierrez (approved):
Re-reviewed at 4d4430a. The only change since my review is the GB200 comment fix (2e1a02a), which addresses the ordering-dependency note I left — ParseGPUSKU and all test assertions are unchanged. CI is green except tests/E2E and analyze still running; the merge gate holds the merge until those pass. No blockers. Approving to unblock the SMI refactor in #1352.

PR #1352, opened 2026-06-13
feat(collector): driver-free GPU SKU discovery; remove nvidia-smi + CUDA base

### LINKED ISSUE #1351: Remove nvidia-smi dependency: driver-free SKU via GFD label + PCI device-ID
## Summary

Derive the GPU accelerator SKU without `nvidia-smi`, then remove the SMI collector entirely. This lets the snapshot/agent image drop its CUDA base (`nvcr.io/nvidia/cuda`) and ship as a pure-Go static image built with `ko`, eliminating the recurring CUDA-image CVE surface.

## Motivation / Context

The only data the rest of AICR consumes from the SMI collector is the GPU product name → accelerator SKU (`gpu.smi.gpu.model`, [`pkg/fingerprint/from_measurements.go#L41`](https://github.com/NVIDIA/aicr/blob/bbd08495558ab6d3fe1fccec559743db6569fb8c/pkg/fingerprint/from_measurements.go#L41)). Driver/CUDA/vbios/MIG/memory fields are collected but consumed by nothing (no fingerprint dimension, no validator code, no recipe constraint).

The SKU is already obtainable without SMI:

- **GFD node label** `nvidia.com/gpu.product` (set by GPU Feature Discovery / GPU Operator) is already captured by the topology collector and already feeds `reconcileAccelerator` ([`from_measurements.go#L102`](https://github.com/NVIDIA/aicr/blob/bbd08495558ab6d3fe1fccec559743db6569fb8c/pkg/fingerprint/from_measurements.go#L102)). In the motivating OKE L40S snapshot, SMI returned 0 GPUs yet the fingerprint still resolved an accelerator from this label.
- **PCI device ID** is already discovered by the NFD PCI source — `extractHardwareInfo` iterates the `device` instances and reads `vendor`/`class` ([`pkg/collector/gpu/nfd.go#L113-L119`](https://github.com/NVIDIA/aicr/blob/bbd08495558ab6d3fe1fccec559743db6569fb8c/pkg/collector/gpu/nfd.go#L113)) but **discards `dev.Attributes["device"]`**, the device ID needed to identify the SKU driver-free.

`nvidia-smi` never covered the day-0 case anyway (no driver image ⇒ no `nvidia-smi`); PCI enumeration is the genuinely driver-free path. So removing SMI in favor of label + PCI is strictly better for AICR's pre-deployment use case, and removes a heavy, CVE-prone base image.

The CUDA dependency is wired into the build, not just the agent pod:

```
.ko.yaml:15         defaultBaseImage: nvcr.io/nvidia/cuda:13.2.1-runtime-ubuntu24.04
.goreleaser.yaml:152               base_image: nvcr.io/nvidia/cuda:13.2.1-runtime-ubuntu24.04
.settings.yaml:150  snapshot_agent_cuda_image: nvcr.io/nvidia/cuda:13.1.0-base-ubuntu24.04
```

Fixes: N/A
Related: #1349, #1350 (token-boundary SKU matching — prerequisite; the PCI/label paths must reuse that same normalization + `unknown-sku` logic)

## Proposed approach (phased)

1. **(done in #1350)** Token-boundary SKU matching so product names / labels classify correctly and unknown SKUs fail safe to `unknown-sku`.
2. **PCI device-ID → accelerator mapping.** In `extractHardwareInfo`, read `dev.Attributes["device"]`, map it to the AICR accelerator enum via a curated device-ID table, and surface it on `HardwareInfo` ([`pkg/collector/gpu/hardware.go#L30`](https://github.com/NVIDIA/aicr/blob/bbd08495558ab6d3fe1fccec559743db6569fb8c/pkg/collector/gpu/hardware.go#L30)) + the `hardware` subtype. Add a fingerprint source for it. Reuse #1350's normalization; unknown device IDs → `unknown-sku`.
3. **Make the GFD label the primary SKU source** and **remove the SMI collector** (`collectSMI`, the `NVSMIDevice` XML model, and the `nvidia-smi` exec in `pkg/collector/gpu`). Fingerprint precedence becomes: GFD label → PCI device-ID → `unknown-sku`.
4. **Drop the CUDA base image.** Flip `.ko.yaml` / `.goreleaser.yaml` base to distroless/static, remove `snapshot_agent_cuda_image`, and build the agent/server image with `ko` like `aicr`.

## Open decisions (call out before implementation)

- **Device-ID table ownership / freshness.** Where it lives and how it's kept current (CI check vs periodic review). New device IDs fail safe to `unknown-sku` until added; NVIDIA-employee maintainers have launch visibility. AICR's enum is coarse (~7 buckets), so the table maps device-ID → bucket, not exact marketing names.
- **GFD availability assumption.** `nvidia.com/gpu.product` requires GPU Operator/GFD to have labeled the node. The PCI path is the fallback for nodes before that.
- **Residual host access.** PCI/module detection still needs host `/sys` (+ `/proc`) mounts ([`pkg/snapshotter/doc.go#L195`](https://github.com/NVIDIA/aicr/blob/bbd08495558ab6d3fe1fccec559743db6569fb8c/pkg/snapshotter/doc.go#L195)). Removing SMI drops the driver / CUDA image / `--runtime-class nvidia` requirement, not the sysfs mount.

## Acceptance criteria

- [ ] `extractHardwareInfo` maps PCI device IDs to the accelerator enum; covered by a table-driven test for every supported SKU's known device IDs, with unknown IDs → `unknown-sku`.
- [ ] Fingerprint resolves the SKU with `nvidia-smi` absent, on both a GFD-labeled node (label path) and an unlabeled node (PCI path).
- [ ] SMI collector code removed; no remaining consumers of SMI-only fields.
- [ ] Image base switched to static/distroless; `nvidia-smi` and the CUDA base no longer required to fingerprint; `grype` scan no longer surfaces CUDA-image CVEs.
- [ ] Docs updated (`pkg/collector/gpu/doc.go`, snapshotter agent docs, any image/runtime-class guidance).

## Type of Change

- [x] New feature / enhancement (driver-free SKU detection; dependency removal)
- [ ] Bug fix

## Risk Assessment

- [ ] **Low**
- [x] **Medium** — touches GPU collection, fingerprint precedence, and image build; phased and reversible (SMI removal is the last, gated step).

mchmarny:
> [\!IMPORTANT]
> **Stacked on #1350.** This PR is based on the `fix/1349-gpu-sku-token-matching` branch and **depends on #1350 landing in `main` first** (it reuses the token-boundary `ParseGPUSKU` from that PR). The base will be retargeted to `main` once #1350 merges. Review the 4 commits here; the diff excludes #1350's commit.

## Summary

Discover the GPU accelerator SKU driver-free (NFD/PCI device ID + GFD label) and remove the `nvidia-smi` collector entirely, so the `aicr` image can drop its CUDA base and ship as a pure-Go static image — eliminating the recurring CUDA-image CVE surface.

## Motivation / Context

The only data anything consumed from the SMI collector was the GPU product name → accelerator SKU. The SKU is obtainable without `nvidia-smi`: the NFD PCI source already enumerates devices (we now read the device ID), and the GPU-operator `nvidia.com/gpu.product` label already feeds the fingerprint. `nvidia-smi` never covered day-0 (no driver ⇒ no `nvidia-smi`) anyway, so PCI is strictly better for AICR's pre-deployment use case.

Fixes: #1351
Related: #1349, #1350 (prerequisite)

## Type of Change

- [x] New feature (driver-free SKU discovery)
- [x] Refactoring (remove SMI collector)
- [x] Build/CI/tooling (drop CUDA base image)

## Component(s) Affected

- [x] Collectors / snapshotter (`pkg/collector/gpu`, `pkg/fingerprint`)
- [x] Build/CI (`.ko.yaml`, `.goreleaser.yaml`, `.github/actions/*`, `.settings.yaml`)
- [x] Docs/examples (`docs/`, `examples/`, `demos/`)

## Implementation Notes

Organized as 4 phase commits (each builds + tests green):

1. **PCI device-ID SKU discovery** — `extractHardwareInfo` reads the PCI `device` attribute and maps it via `device_ids.go` (curated datacenter vocabulary from pci.ids: Pascal→Blackwell + RTX PRO 6000, with converged/consumer/legacy excluded). The real SKU lands in the `hardware.model` measurement and a new **descriptive** `Fingerprint.GPUModel` field.
2. **Remove SMI collector** — deletes the `nvidia-smi` exec, the ~400-line `NVSMIDevice` XML model, and the 4.2k-line `gpu.xml` fixture. Fingerprint accelerator resolution: **GFD label (primary) → PCI device ID (gated to recipe-supported SKUs) → unknown-sku**.
3. **Drop CUDA base image** — remove the `base_image`/`defaultBaseImage` pins so `aicr`/`aicrd` use ko's default static base (`cgr.dev/chainguard/static`); rebase the CI smoke-test agent image; rename `snapshot_agent_cuda_image` → `snapshot_agent_base_image`.
4. **Docs/examples SMI sweep** — update the `smi` subtype / `gpu.smi.*` references to the `hardware` subtype / `gpu.hardware.model`.

**Design guarantees (aligned with maintainer first principles):**
- **No footgun:** the *selectable* accelerator in CLI/API/SDK stays the recipe enum; the broader discovery vocabulary is output-only.
- **Matching/attestation unchanged in logic:** the matching `Accelerator` dimension stays enum-limited (`ParseGPUSKU` from #1350 is unchanged); PCI backfills it only for recipe-supported SKUs. `ToCriteria` and `Fingerprint.Match` are unaffected for labeled nodes; the one additive change is that a driver-less day-0 node now gets a definitive supported-SKU accelerator instead of `unknown`.

**Driver/CUDA version note:** the snapshot no longer records GPU driver/CUDA version *via SMI*. On GPU-operator clusters these remain available in the snapshot through the GFD node labels (`nvidia.com/cuda.driver-version.full`, etc.) already captured by the topology collector — only a driver-installed-but-GFD-absent node loses them.

## Testing

```bash
make test   # PASS (full repo, -race)
make lint   # PASS — 0 issues; AGENTS.md in sync; doc checks pass
```

`pkg/trust/TestUpdate_Success` fails locally only due to the sandbox blocking `tuf-repo-cdn.sigstore.dev` (network) — unrelated; passes in CI. `make qualify`'s e2e (KWOK/kind) + grype scan + the smoke-test image build were not run locally (need cluster/docker/CI) — CI covers them; the smoke-test image base change is exercised there.

## Risk Assessment

- [x] **Medium** — touches GPU collection, fingerprint precedence, and the image build. Phased and reversible.

**Rollout notes:** Snapshots of unsupported GPUs now fingerprint with an empty `accelerator` value carrying the `unknown-sku` note + `gpuModel: <real-sku>` (e.g. `l40s`) instead of a wrong-but-valid value. The `unknown-sku` note (preserved from the old nvidia-smi path) keeps "GPU present but unsupported" visible. External consumers reading `gpu.smi.*` must switch to `gpu.hardware.model` (or the GFD driver-version labels). N/A migration for recipes.

## Checklist

- [x] I updated docs for user-facing behavior changes


.github/actions/aicr-build/build-snapshot-agent.sh
```diff
@@ -25,20 +25,21 @@ if [[ ! -f dist/aicr ]]; then
...
 
-# Load onto all nodes. The snapshot agent requests nvidia.com/gpu but does not
-# set a node selector, so it can land on any GPU-capable node including the
-# control-plane in the L40G smoke test.
+# Load onto all nodes. The snapshot agent has no node selector, so it can land
+# on any node including the control-plane in the smoke test; PCI enumeration
+# reads sysfs and needs no GPU resource/runtime-class.
 timeout 900 kind load docker-image ko.local:smoke-test --name "${KIND_CLUSTER_NAME}" || {
   echo "::warning::kind load attempt 1 failed for ko.local:smoke-test, retrying..."
   timeout 900 kind load docker-image ko.local:smoke-test --name "${KIND_CLUSTER_NAME}"
```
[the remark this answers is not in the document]
mchmarny: Done in 57c3dc70 — added an explicit `KIND_CLUSTER_NAME` precheck (matching the `SNAPSHOT_AGENT_BASE_IMAGE` guard) so it fails with a clear CI diagnostic.

docs/user/cli-reference.md
```diff
@@ -248,8 +248,8 @@ Example template extracting key cluster info:
 cluster:
   kubernetes: {{ with index .Measurements 0 }}{{ range .Subtypes }}{{ if eq .Name "server" }}
     version: {{ (index .Data "version").String }}{{ end }}{{ end }}{{ end }}
-  gpu: {{ range .Measurements }}{{ if eq .Type.String "GPU" }}{{ range .Subtypes }}{{ if eq .Name "smi" }}
-    model: {{ (index .Data "gpu.model").String }}
+  gpu: {{ range .Measurements }}{{ if eq .Type.String "GPU" }}{{ range .Subtypes }}{{ if eq .Name "hardware" }}
+    model: {{ (index .Data "model").String }}
     count: {{ (index .Data "gpu-count").String }}{{ end }}{{ end }}{{ end }}{{ end }}
```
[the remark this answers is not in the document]
mchmarny: Done in 57c3dc70 — updated the subtype-name example from `smi` to `hardware`.

docs/user/installation.md
```diff
@@ -177,14 +177,13 @@ sudo chmod +x /usr/local/bin/aicr
...
-
-# If missing, install NVIDIA drivers for your platform
-```
+Snapshot GPU detection is driver-free: it enumerates PCI devices via sysfs
+(NFD) and resolves the accelerator SKU from the device ID — no `nvidia-smi` or
+NVIDIA driver required. When run in-cluster, the agent also reads the GPU
+Operator's `nvidia.com/gpu.product` node label. If a snapshot reports no GPU on
+a node you expect to have one, confirm the agent landed on the GPU node (it
+needs host `/sys` access) and, for the most accurate SKU, that the GPU Operator
+has labeled the node.
```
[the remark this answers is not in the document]
mchmarny: Done in 57c3dc70 — the note now states the `nvidia.com/gpu.product` label is **not** required for GPU detection (driver-free via PCI/sysfs); it only improves SKU accuracy and powers the placement-mismatch warning.

.settings.yaml
```diff
@@ -146,8 +146,10 @@ testing:
   # GPU CI runtime pins consumed through .github/actions/load-versions.
   # renovate: datasource=helm depName=gpu-operator depType=testing
   gpu_operator_chart_version: 'v25.10.1'
-  # renovate: datasource=docker depName=nvcr.io/nvidia/cuda depType=testing
-  snapshot_agent_cuda_image: 'nvcr.io/nvidia/cuda:13.1.0-base-ubuntu24.04'
+  # Base image for the CI smoke-test snapshot-agent image. The agent binary is
+  # static Go and detects GPUs driver-free via NFD/PCI, so it no longer needs a
+  # CUDA base / nvidia-smi. Matches ko's default static base for the aicr image.
+  snapshot_agent_base_image: 'cgr.dev/chainguard/static:latest'
```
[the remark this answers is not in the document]
mchmarny: Intentionally leaving the floating tag. This is a CI-internal smoke-test base image, not a published/release artifact, and the published `aicr`/`aicrd` images deliberately use ko's default *unpinned* static base (`cgr.dev/chainguard/static`). Digest-pinning only this CI value would be inconsistent with that and add Renovate churn for an ephemeral test image. Resolving.

.github/actions/gpu-snapshot-validate/validate-snapshot-gpu.sh
```diff
@@ -15,16 +15,19 @@
...
+GPU_MODEL=$(yq eval '.measurements[] | select(.type == "GPU") | .subtypes[] | select(.subtype == "hardware") | .data.model' snapshot.yaml)
+GPU_COUNT=$(yq eval '.measurements[] | select(.type == "GPU") | .subtypes[] | select(.subtype == "hardware") | .data["gpu-count"]' snapshot.yaml)
 echo "GPU model: ${GPU_MODEL}"
 echo "GPU count: ${GPU_COUNT}"
 if ! [[ "${GPU_COUNT}" =~ ^[0-9]+$ ]]; then
   echo "::error::Expected numeric gpu-count in snapshot, got: ${GPU_COUNT}"
   exit 1
 fi
-if [[ "${GPU_MODEL}" != *"${EXPECTED_GPU_MODEL}"* ]]; then
+if [[ "${GPU_MODEL,,}" != *"${EXPECTED_GPU_MODEL,,}"* ]]; then
```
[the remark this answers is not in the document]
mchmarny: Done in fc49fd69 — switched to exact normalized equality so e.g. `gh200` can't satisfy an expected `h200`.

tests/e2e/run.sh
```diff
@@ -447,22 +447,20 @@ test_snapshot() {
...
-  gpu_name=$(echo "$snapshot_data" | grep "gpu-product:" | head -1 | sed 's/.*gpu-product: //' || echo "unknown")
+  # Extract and display GPU info from the driver-free "hardware" subtype
+  # (the SMI subtype was removed; model is the PCI-derived accelerator SKU).
+  local gpu_name gpu_count driver_loaded
+  gpu_name=$(echo "$snapshot_data" | grep -E "^\s*model:" | head -1 | sed 's/.*model: //' || echo "unknown")
   gpu_count=$(echo "$snapshot_data" | grep "gpu-count:" | head -1 | sed 's/.*gpu-count: //' || echo "0")
-  gpu_mem=$(echo "$snapshot_data" | grep "gpu-memory:" | head -1 | sed 's/.*gpu-memory: //' || echo "unknown")
-  driver_ver=$(echo "$snapshot_data" | grep "driver-version:" | head -1 | sed 's/.*driver-version: //' || echo "unknown")
-  cuda_ver=$(echo "$snapshot_data" | grep "cuda-version:" | head -1 | sed 's/.*cuda-version: //' || echo "unknown")
+  driver_loaded=$(echo "$snapshot_data" | grep "driver-loaded:" | head -1 | sed 's/.*driver-loaded: //' || echo "unknown")
```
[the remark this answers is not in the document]
mchmarny: Done in fc49fd69 — extraction is now `yq`-scoped to the GPU `hardware` subtype, so a `model` key elsewhere in the snapshot can't be misread as the GPU SKU.

.goreleaser.yaml
```diff
@@ -163,7 +166,6 @@ kos:
     build: aicrd
```
ArangoGutierrez, line 166: Heads up: removing aicrd's `base_image: gcr.io/distroless/static:nonroot` (just below) means aicrd falls through to ko's default base, and with `defaultBaseImage` also gone from `.ko.yaml` that's `cgr.dev/chainguard/static`. So aicrd's runtime base changes too, not just aicr's CUDA drop. Two things to confirm: (1) that's intended for aicrd (the PR scope reads as aicr-only), and (2) the nonroot posture carries over — distroless `:nonroot` runs as UID 65532; please verify the chainguard default does the same (or set `runAsNonRoot`/USER explicitly) so we don't quietly start running as root.

pkg/collector/gpu/nfd.go
```diff
@@ -100,23 +100,36 @@ func (d *NFDHardwareDetector) Detect(ctx context.Context) (*HardwareInfo, error)
...
 			vendor := dev.Attributes["vendor"]
 			class := dev.Attributes["class"]
 			if vendor == nvidiaVendorID && (class == pciClassVGA || class == pciClass3D) {
 				info.GPUCount++
+				if sku := skuForDeviceID(dev.Attributes["device"]); sku != "" {
+					skus[sku] = struct{}{}
+				}
+			}
+		}
+		if len(skus) == 1 {
```
ArangoGutierrez, line 130: This heterogeneous guard (one distinct SKU -> set it; 0 or >1 -> leave "") is new and not covered by a unit test — there's no `nfd_test.go` change here. Worth a small table test for `extractHardwareInfo`: single SKU resolves, two distinct SKUs -> SKU="", and the known-among-unknown case (one recognized device + one unrecognized device-id) where the node still reports the single known SKU. That last one is a real behavior worth pinning so it doesn't silently drift.

pkg/fingerprint/from_measurements.go
```diff
@@ -80,25 +82,57 @@ func FromMeasurements(measurements []*measurement.Measurement) *Fingerprint {
...
+	if st == nil {
+		return
+	}
+	sku, err := st.GetString(keyGPUHardwareSKU)
+	if err != nil || sku == "" {
+		return
+	}
+	fp.GPUModel = Dimension{Value: sku, Source: sourceAcceleratorPCI}
+
+	if fp.Accelerator.Value == "" && fp.Accelerator.Note != noteMultiGPU && isSupportedAccelerator(sku) {
```
ArangoGutierrez: Behavior change worth confirming: on a day-0 node with no GFD label and an unsupported SKU (e.g. l40s), `Accelerator` ends up empty with no note and the signal lives only in `GPUModel`. The old smi path set `Accelerator.Note=unknown-sku` here — removing `TestFromMeasurements_GPUUnknownModel` makes this look deliberate. Two follow-ups: (1) the description's rollout note says "accelerator: unknown", which now means "empty + GPUModel set" rather than the unknown-sku note, so tighten the wording; (2) confirm nothing downstream keys on `Accelerator.Note==unknown-sku` to mean "GPU present but unsupported." If that distinction still matters, setting the note here when GPUModel resolved-but-unsupported would preserve it.

pkg/collector/gpu/device_ids.go
```diff
@@ -0,0 +1,169 @@
...
+	"2bb1": skuRTXPro6000, // RTX PRO 6000 Blackwell Workstation Edition
+	"2bb4": skuRTXPro6000, // RTX PRO 6000 Blackwell Max-Q Workstation Edition
+	"2bb5": skuRTXPro6000, // RTX PRO 6000 Blackwell Server Edition
+}
+
+// skuForDeviceID returns the normalized accelerator SKU for an NVIDIA PCI
+// device ID, or "" when the ID is not a recognized datacenter GPU. The lookup
+// is case-insensitive and tolerates an optional "0x" prefix.
+func skuForDeviceID(deviceID string) string {
+	id := strings.ToLower(strings.TrimPrefix(strings.TrimSpace(deviceID), "0x"))
```
ArangoGutierrez: Small robustness nit: `TrimPrefix(..., "0x")` is case-sensitive and runs before `ToLower`, so a `0X2330` (uppercase X) input wouldn't get the prefix stripped and would miss the map. sysfs gives lowercase without a prefix, so this is theoretical, but reordering to `TrimPrefix(ToLower(TrimSpace(id)), "0x")` makes it uniform. The test covers lowercase `0x` and uppercase hex, just not the uppercase-`0X` combination.

pkg/collector/gpu/device_ids.go
```diff
@@ -0,0 +1,169 @@
...
+	"20b5": skuA100, // A100 PCIe 80GB
+	"20f1": skuA100, // A100 PCIe 40GB
+	"20b7": skuA30,  // A30 PCIe
+	"20bd": skuA800, // A800 SXM4 40GB
+	"20f3": skuA800, // A800-SXM4-80GB
+	"20f5": skuA800, // A800 80GB PCIe
+	"20f6": skuA800, // A800 40GB PCIe
+	"2235": skuA40,  // A40
+	"2236": skuA10,  // A10
+	"25b6": skuA16,  // A2 / A16
```
ArangoGutierrez, line 116: Concrete example of the table-correctness question: 25b6 is the shared A2/A16 device id (GA107GL), mapped here to a16 — a judgment call, and since neither is recipe-supported it only affects the descriptive GPUModel. More broadly, the positive cases in `device_ids_test.go` derive `want` from the same map they check, so they pin the table against accidental edits but can't catch a mis-sourced id; the exclusion cases (converged / consumer / NVSwitch -> "") are the real guards. Worth a reviewer spot-check of the datacenter entries against pci.ids, or generating the table from pci.ids if it's expected to grow.

ArangoGutierrez (commented):
Big, clean change. The SMI removal is thorough — collector, the ~400-line XML model, and the 4.2k-line fixture all gone — and the four-commit structure makes it reviewable. Driver-free PCI discovery with GFD-label precedence is a real day-0 improvement, and dropping the CUDA base is a solid CVE-surface win. The precedence tests (`TestFromMeasurements_PCIBackfill`: supported->backfill, unsupported->GPUModel-only, label-wins-over-PCI) are the right matrix.

**Code review**
- `.goreleaser.yaml`: removing aicrd's `base_image` looks like a side effect past the stated scope — aicrd moves from `gcr.io/distroless/static:nonroot` to ko's chainguard default. Confirm it's intended and that the nonroot UID posture survives. Inline.
- `nfd.go`: the new heterogeneous / multi-SKU resolution in `extractHardwareInfo` has no direct unit test (no `nfd_test.go` change in the PR). Inline.
- `from_measurements.go`: a day-0 unsupported SKU now leaves `Accelerator` empty with no note, where the old smi path set `unknown-sku`. Looks deliberate (you dropped the old test), but the rollout note says "accelerator: unknown" — worth reconciling. Inline.
- `device_ids.go`: the table is only as correct as the pci.ids curation; the positive tests can't catch a mis-sourced id. Inline.

**Go review**
- `skuForDeviceID`: the `0x` strip is case-sensitive and runs before `ToLower`, so a `0X`-prefixed id slips through. Minor, but easy to make uniform. Inline.
- Error and degradation paths read well — `Collect` returns no subtypes rather than erroring, the `GetString` early-returns are consistent, and no concurrency surface is introduced.

Non-blocking: this is stacked on #1350 (base `fix/1349-...`, not `main`), so re-target once #1350 lands. Review order is #1350 first, then this.

.goreleaser.yaml
mchmarny: Good catch on the aicrd posture. Confirming both points:

1. **Intended for aicrd** — yes. The intent is to consolidate both `aicr` and `aicrd` on ko's default static base rather than leave aicrd on its own distroless pin. Made that explicit in 42f1f15f with a comment on the aicrd ko block.
2. **Nonroot posture preserved** — ko's default base `cgr.dev/chainguard/static` runs as nonroot **UID 65532** by default (that's precisely why ko adopted it as its default, replacing the old `gcr.io/distroless/static:nonroot` which is also 65532). So the posture carries over identically; the new comment documents the 65532 ↔ 65532 equivalence so this stays intentional and visible.

pkg/collector/gpu/nfd.go
mchmarny: Added in 42f1f15f. `TestExtractHardwareInfo` now sets `device` attributes and asserts the resolved SKU, with the three cases you called out:
- single SKU resolves (2×`2330` → `h100`),
- two distinct SKUs → `""` (`2330` H100 + `20b0` A100, heterogeneous),
- known-among-unknown → single known SKU (`2330` H100 + `ffff` unrecognized → `h100`).

That last one pins the behavior so it can't silently drift.

pkg/fingerprint/from_measurements.go
mchmarny: Both follow-ups addressed in 42f1f15f.

**(2) Downstream:** confirmed nothing keys on `Accelerator.Note == "unknown-sku"` — the constant is referenced only within `from_measurements.go`. So the distinction is observability-facing, not logic-bearing.

**(1) + the restore:** rather than let the day-0 unsupported case go silent, I restored the signal. `populatePCIDiscovery` now sets `Accelerator.Note = unknown-sku` (via the PCI source) when a GPU resolves to a SKU outside the recipe enum and no higher-priority source already set a value/note — symmetric with the topology path in `reconcileAccelerator`. The matching `Value` still stays empty (so `Match`/`ToCriteria` are untouched) and the descriptive SKU still lands in `GPUModel`. Tightened the rollout-note wording in the PR description to match (empty `accelerator` value + `unknown-sku` note, not a bare empty).

pkg/collector/gpu/device_ids.go
mchmarny: Fixed in 42f1f15f — reordered to `TrimPrefix(ToLower(TrimSpace(deviceID)), "0x")` so the prefix strip is uniform, and added a `0X2330` (uppercase-X) test case alongside the existing lowercase `0x` one.

pkg/collector/gpu/device_ids.go
mchmarny: Spot-checked the datacenter entries against pci.ids. `25b6` is `GA107GL`, the shared **A2 / A16** device id — the inline `// A2 / A16` note flags that, and since neither is recipe-supported the choice only affects the descriptive `GPUModel`, never matching. Agreed the positive cases derive `want` from the same map (so they guard against accidental edits, not mis-sourcing) and the exclusion cases (converged / consumer / NVSwitch → `""`) are the real guards.

Leaving the table hand-curated for now rather than generating from pci.ids: it's intentionally a small datacenter-only vocabulary with curation decisions (converged/consumer/legacy excluded) that a raw pci.ids import wouldn't capture. If it grows materially, generating + a curated exclusion list is the right follow-up — worth its own issue rather than expanding this PR's scope. No code change here.

ArangoGutierrez (commented):
Re-reviewed at 42f1f15 — this commit cleanly addresses all five points from the last pass, and I traced each:

- **aicrd base image (`.goreleaser.yaml`):** resolved. The comment documents that aicrd intentionally rides ko's chainguard/static default and stays nonroot UID 65532, matching the old `distroless/static:nonroot`. That's correct for ko's default user. Only nice-to-have: assert the nonroot user on the built image in CI so the posture can't silently regress later — not a blocker.
- **Heterogeneous PCI path (`nfd_test.go`):** resolved. The new `TestExtractHardwareInfo` cases cover the three I flagged — single SKU resolves, two distinct SKUs -> "", known-among-unknown -> the known SKU — with real `wantSKU` assertions that go red if the guard breaks.
- **unknown-sku note (`from_measurements.go`):** resolved, and the precedence holds up. A day-0 unsupported SKU now records `Accelerator{Note: unknown-sku}` with the descriptive value in `GPUModel`; supported SKUs still backfill the value; a label-resolved value or multi-gpu note is left untouched; and a topology-set unknown-sku note isn't clobbered by the PCI source. The early-return refactor also reads cleaner than the old compound condition, and matches the smi behavior the rollout notes describe.
- **0x prefix (`device_ids.go`):** resolved. `ToLower` runs before `TrimPrefix` now, and the new `0X2330` test case discriminates against the old ordering.

One standing, non-blocking item: the device-ID table's positive entries are still only as trustworthy as the pci.ids curation (the tests pin them but can't catch a mis-sourced id). Nothing for this commit to change — leaving it as a spot-check / generator suggestion if the table grows.

No blockers from the code. CI is green except the two H100 GPU e2e jobs still running; the full Merge Gate suite will run once this retargets to main after #1350 lands. Stacked-on-#1350 merge order still stands.

ArangoGutierrez (dismissed):
Approving as an aicr-maintainer code owner. Re-reviewed at 42f1f15 — all five points from the prior pass are addressed and verified.

This satisfies the code-owner review for the maintainer-owned files in this PR (`.goreleaser.yaml`, `.settings.yaml`, `.github/**`). The Go code, docs, and examples are under the default `aicr-write` owner and still need an approval from that team.

Heads-up: this approval will likely need re-doing once #1350 lands and this branch rebases/retargets onto main (new commit SHAs will mark it stale).

ArangoGutierrez (approved):
Re-approving after the retarget and rebase onto main (the prior approval was dismissed when the SHA changed). I verified the rebased tree at 11f2004 matches what I approved: same 32-file change, gpu_sku.go correctly dropped (now in main via #1350), and all five review fixes intact — the 0x prefix ordering, the unknown-sku note on the PCI path, the nonroot-65532 note on the aicrd base, and the new heterogeneous and known-among-unknown nfd tests.

This satisfies the aicr-maintainer code-owner review for the maintainer-owned files (.goreleaser.yaml, .settings.yaml, .github). The Go code, docs, and examples are under aicr-write and still need an approval from that team.

Note: the full Merge Gate is running against main for the first time on this branch; the merge will correctly hold until those checks are green.

PR #1367, opened 2026-06-15
fix(docs): escape bare < for Fern MDX + harden MDX checker

mchmarny:
## Summary

Unblock the Fern docs publish (MDX) and close the two gaps in the local MDX checker that let the breakage reach the release tag.

## Motivation / Context

Fern parses docs as MDX, which treats a bare `<` followed by a name-start character as a JSX tag. `landed <30 s` in `docs/contributor/inference-perf-fluctuation.md` failed the `v0.15.0-rc2` Publish Fern Docs job:

```
Failed to parse ../docs/contributor/inference-perf-fluctuation.md: Unexpected character `3` (U+0033) before name ...
```

There is already a local gate for this — `tools/check-docs-mdx`, exposed as `make check-docs-mdx` and wired into `make lint` (hence `make qualify`) and the PR-CI `fern-docs-ci.yaml`. It should have caught both this and the earlier recipe-health.md breakage, but had two gaps:

1. **Digit gap:** the bare-tag check only matched `<` before a *letter* (`<[a-zA-Z]`), so `<30 s` / `<13%` slipped through.
2. **Stale exclusion:** `recipe-health.md` was excluded (on a now-obsolete VitePress HTML-marker assumption) — which is why its HTML comment reached the rc1 publish.

(`fern check` does not catch MDX *parse* errors, so publish was the first place these surfaced.)

Fixes: N/A
Related: https://github.com/NVIDIA/aicr/actions/runs/27557526326

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: docs MDX checker (`tools/check-docs-mdx`, `make check-docs-mdx`)

## Implementation Notes

Three commits:
1. **Doc fix** — escape `<30 s` as `\<30 s` (literal `<` in MDX and GitHub CommonMark). `<13%` on line 162 is inside a code fence, so MDX leaves it literal — consistent with the error pointing at `<30` (char `3`).
2. **Checker: digit gap** — broaden the bare-tag match to `<[0-9a-zA-Z]`, and strip `\<` / `\>` individually so a lone `\<` escape doesn't false-positive.
3. **Checker: exclusion + MDX comments** — drop the `recipe-health.md` exclusion (it is Fern-published and now uses MDX-safe `{/* ... */}` markers), and teach the checker to treat `{/* ... */}` comments (single- and multi-line) as safe so the valid markers don't trip the bare-brace check while real hazards on the same line are still caught. `container-images.md` stays excluded (HTML markers, not Fern-published).

## Testing

```bash
make check-docs-mdx     # OK on the full tree (now including recipe-health.md)
# probes: bare `<30 s` / `<13%` and `{foo}` are flagged; `\<...\>`, backticked
# `<...>`, and `{/* ... */}` comments are treated as safe.
```

A sweep of all 31 Fern-nav docs found `<30 s` to be the only remaining bare-`<` prose hazard and no stray `{` hazards beyond valid MDX comments.

## Risk Assessment

- [x] **Low** — one-char doc escape plus regex/handling fixes in a lint tool; easy to revert.

**Rollout notes:** With both checker gaps closed, `make lint` / `make qualify` now catch this class pre-merge — no separate pre-push hook needed.

## Checklist

- [ ] Tests pass locally (`make test` with `-race`) — N/A
- [x] Linter passes (`make check-docs-mdx`)

### DIFF
--- a/tools/check-docs-mdx
+++ b/tools/check-docs-mdx
@@ -9,9 +9,10 @@
 # common violations before they reach CI or break versioned doc snapshots.
 #
 # Scope: docs/user/, docs/integrator/, and docs/contributor/ (all published
-# to Fern). container-images.md and recipe-health.md carry auto-generated
-# HTML comment splice markers that VitePress still needs; both are excluded
-# until VitePress is removed (see EXCLUDE_NAMES below).
+# to Fern). container-images.md carries auto-generated HTML comment splice
+# markers that VitePress still needs and is not published to Fern, so it is
+# excluded (see EXCLUDE_NAMES below). recipe-health.md IS published to Fern
+# and uses MDX-safe {/* ... */} comment markers, so it is checked.
 # Override with positional args for frozen version content.
 #
 # Checks:
@@ -28,10 +29,11 @@ set -euo pipefail
 
 ERRORS=0
 SEARCH_DIRS=("docs/user/" "docs/integrator/" "docs/contributor/")
-# Docs with auto-generated HTML-comment splice markers that VitePress still
-# needs (excluded until VitePress is removed): the BOM inventory and the
-# recipe-health matrix, both spliced by `make *-docs` between marker comments.
-EXCLUDE_NAMES=("container-images.md" "recipe-health.md")
+# Docs excluded from the MDX scan: container-images.md carries auto-generated
+# HTML-comment splice markers that VitePress still needs and is not published
+# to Fern. (recipe-health.md is intentionally NOT excluded — it is Fern-
+# published and uses MDX-safe {/* ... */} markers.)
+EXCLUDE_NAMES=("container-images.md")
 EXCLUDE_ARGS=()
 for name in "${EXCLUDE_NAMES[@]}"; do
   EXCLUDE_ARGS+=(! -name "${name}")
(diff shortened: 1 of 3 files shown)


tools/check-docs-mdx
```diff
@@ -70,12 +72,24 @@ for dir in "${SEARCH_DIRS[@]}"; do
...
+        # MDX comments {/* ... */} are valid (and used for license headers
+        # and the auto-generated splice markers in recipe-health.md). Treat
+        # them like code: continue an open multi-line comment, strip
+        # single-line ones, and open a region for an unterminated {/*.
+        if (in_mdx_comment) {
+          ci = index(line, "*/}")
+          if (ci > 0) { line = substr(line, ci + 3); in_mdx_comment = 0 }
+          else { next }
+        }
+        gsub(/\{\/\*.*\*\/\}/, "", line)
```
ArangoGutierrez: Greedy `.*` strips from the first `{/*` to the last `*/}` on the line, so two comments on one line take out whatever sits between them:

```
{/* a */} landed <30 s here {/* b */}
```

The whole line gets blanked and the `<30 s` is missed. `[^}]*` covers the in-tree markers; a C-comment-style match or an iterative strip is fully correct if a comment body can ever contain `}`.

tools/check-docs-mdx
```diff
@@ -70,12 +72,24 @@ for dir in "${SEARCH_DIRS[@]}"; do
...
+        if (match(line, /\{\/\*/)) { line = substr(line, 1, RSTART - 1); in_mdx_comment = 1 }
```
ArangoGutierrez: This opens the multi-line comment region before the inline code spans are stripped (L88), so a backticked `{/*` with no close on the same line trips it:

```
Write an MDX comment like `{/*` to start one.
runs landed <30 s and passed
```

The region then swallows every line until EOF or a stray `*/}`, so the `<30 s` slips through. Stripping code spans before this detection avoids it. Not in the tree today, just flagging since it matches the gap this PR is fixing.

ArangoGutierrez (commented):
Pulled the head tree and ran the new checker end to end. It does what the description says:

- Passes clean on the full docs tree, so the digit broadening doesn't add false positives.
- Catches a bare `<30 s` where the old `<[a-zA-Z]` regex missed it.
- `\<30 s`, backticked `` `<30 s` ``, and `{/* ... */}` (single and multi-line) all pass; a hazard after a same-line or multi-line comment close is still flagged.
- recipe-health.md's license header and BEGIN/END splice markers are handled now that it's scanned.

Two latent edge cases in the comment handling, both inline below. Neither is hit by the current tree, and both are the same gap-class this PR closes, so I'd treat them as a fast follow rather than a blocker.

Separately, there's no regression test for the checker itself. A small testdata/ of pass/fail snippets would lock the behavior in and would have caught both edges below. Happy to send that plus the fixes as a follow-up.

Nothing here blocks unblocking the publish.

tools/check-docs-mdx
mchmarny: Fixed in c112c432 — the greedy `.*` was replaced with an iterative single-segment strip (one `{/* … */}` at a time), so a hazard between two comments on one line (`{/* a */} <30 s {/* b */}`) survives and is still checked. Resolving.

tools/check-docs-mdx
mchmarny: Fixed in 57ff7685 — inline code spans are now stripped *before* the MDX-comment detection, so a backticked, unterminated `` `{/*` `` no longer opens a bogus multi-line region. While verifying I also found and fixed a `set -e` issue in the same loop that aborted the scan at the first offending file; both fixes are in that commit. Resolving.

PR #1369, opened 2026-06-15
docs(integrator): document Dynamo 1.2 NATS SG symptoms on EKS

yuanchen8911:
## Summary

Extend `docs/integrator/eks-dynamo-networking.md` with the failure signature actually observed when the GPU→system security group blocks NATS `4222` after the Dynamo 1.2 bump. The required SG rules (4222 + 9090) were already documented; this adds the symptoms and a quick reachability probe so the failure is recognizable.

## Motivation / Context

Dynamo 1.2 (#1308) switched discovery to the NATS event plane (TCP 4222). On DGXC EKS clusters whose system-node SG only allowlisted the pre-1.2 control-plane ports (e.g. 27017/6379/8080), 4222 is closed, so a bundle that worked on Dynamo 1.0.x starts failing purely from the version bump. The crash presents as a workload bug (CrashLoopBackOff) rather than a network/SG issue, which cost real debugging time. Reproduced on GB200 (yljtrxpmzu) and H100 (aicr3) clusters; the same rc2 bundle runs fine on GKE (flat network).

Related: #1308

## Type of Change

- [x] Documentation update

## Components Affected

- [x] Docs/examples (`docs/`)

## Implementation Notes

- Adds a note that the 4222 requirement is new as of the Dynamo 1.2 NATS switch.
- Expands the symptom list: frontend `Failed to connect to NATS: timed out`, worker `CrashLoopBackOff`, startup-probe `connection refused`, and an `inference-perf` timeout while deployment/conformance pass.
- Adds a one-line busybox `nc` reachability probe from a GPU node.

## Testing

Doc-only change. Verified Markdown renders; no headings renamed/removed (no anchor-link impact). Full `make qualify` not run — tests/e2e/Go-lint cannot regress from a docs-only edit; CI lychee covers link checks.

## Risk Assessment

Minimal — additive documentation only.

## Checklist

- [x] Docs updated in the same PR as the behavior they describe
- [x] No code changes

PR #1376, opened 2026-06-16
test(uat): allow NATS ingress on UAT AWS/GCP clusters

mchmarny:
## Summary

Allow NATS (4222/tcp) ingress on the UAT AWS and GCP clusters.

## Motivation / Context

Dynamo's NATS clients (in worker pods) could not reach the NATS server on the system nodes in the UAT environments.

Fixes: N/A
Related: #1369

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT cluster infra config (`tests/uat/`)

## Implementation Notes

- **AWS:** added an EKS security-group ingress rule for 4222/tcp. It sources **both** the worker node subnet (`10.0.128.0/17`) **and** the pod subnet (`100.65.0.0/16`) — AWS VPC-CNI does not SNAT in-VPC pod-to-pod traffic, so the system node hosting NATS sees the *pod* source IP, not the node IP. A node-only rule would never match.
- **AWS:** added the HQ CIDR `128.77.49.32/30` to the control-plane endpoint `allowedCidrs`, for parity with the GKE `authorizedNetworks` `nw3` entry. (This is an HQ-access addition, not the worker CIDR — corrected from the earlier description.)
- **GCP:** declared explicit `system`/`worker` node subnets. **No dedicated NATS firewall rule is added** — the existing `nccl-internal` rule already allows all intra-VPC TCP from `10.0.0.0/8` (covering node primaries and the GKE pod secondary ranges), which includes 4222. The initial draft tightened `nccl-internal` to node-only CIDRs and added a redundant `allow-nats` rule; both were reverted after review since they would have dropped pod-range coverage.
- Port 9090 (Prometheus query, `ai-service-metrics`) is already enabled by default, so no rule is needed for Prometheus queries from worker nodes.

## Testing

```bash
# Config-only change to UAT infra YAML; no code paths exercised by make qualify.
```

## Risk Assessment

- [x] **Low** — Isolated change to UAT-only infra config, easy to revert

**Rollout notes:** Applied when the UAT clusters are next provisioned/reconciled. N/A for runtime code.

## Checklist

### DIFF
--- a/tests/uat/aws/cluster-config.yaml
+++ b/tests/uat/aws/cluster-config.yaml
@@ -35,6 +35,7 @@ cluster:
         - 216.228.127.128/30
         - 52.41.222.58/32
         - 54.68.11.139/32
+        - 128.77.49.32/30
     addOns:
       coreDns: ""
       kubeProxy: ""
@@ -64,6 +65,17 @@ network:
         - cidr: 100.65.0.0/16
           zone: us-east-1e
           disableEndpoints: true
+    securityGroups:
+      additionalRules:
+        - target: system
+          direction: ingress
+          fromPort: 4222
+          toPort: 4222
+          protocol: tcp
+          # VPC-CNI does not SNAT in-VPC pod-to-pod traffic, so NATS on the
+          # system node sees the pod source IP (pod subnet), not the node IP.
+          cidrBlocks: ["10.0.128.0/17", "100.65.0.0/16"]  # worker node + pod subnet CIDRs
+          description: "Allow NATS from worker nodes and pods"
 
 compute:
   eks:
--- a/tests/uat/gcp/cluster-config.yaml
+++ b/tests/uat/gcp/cluster-config.yaml
@@ -41,12 +41,21 @@ cluster:
 
 network:
   gke:
+    subnets:
+      nodes:
+        - name: system-subnet
+          cidr: "10.0.0.0/22"
+        - name: worker-subnet
+          cidr: "10.0.128.0/17"
     gpuNets:
       count: 8
       mtu: 8244
       cidrBase: "192.168.0.0/16"
       gvnicCidr: "192.168.128.0/20"
     firewallRules:
+      # Allows all intra-VPC TCP/UDP/ICMP from 10.0.0.0/8, which covers node
+      # subnets and the GKE pod secondary ranges. NATS (4222) is already
+      # permitted by this rule, so no dedicated allow-nats rule is needed.
       - name: nccl-internal
         direction: INGRESS
         priority: 900


tests/uat/aws/cluster-config.yaml
```diff
@@ -64,6 +65,15 @@ network:
...
           zone: us-east-1e
           disableEndpoints: true
+    securityGroups:
+      additionalRules:
+        - target: system
+          direction: ingress
+          fromPort: 4222
+          toPort: 4222
+          protocol: tcp
+          cidrBlocks: ["10.0.128.0/17"]  # worker subnet CIDR
```
yuanchen8911: AWS NATS rule sources the wrong CIDR

This 4222 ingress rule sources the worker node subnet 10.0.128.0/17, but Dynamo's NATS client runs in worker pods, which get IPs from the pod subnet (network.eks.subnets.pod = 100.65.0.0/16). AWS VPC-CNI does not SNAT in-VPC pod-to-pod traffic, so the system node hosting NATS sees the pod source IP (100.65.x), which is not in 10.0.128.0/17. The rule won't match and NATS stays blocked. Upstream secgroup.tf follows the opposite pattern: its default pod-admitting rules (NFD 8080, Prometheus 9090) source both the VPC and pod CIDRs (10.0.0.0/16 and 100.65.0.0/16).

Recommended fix: Source the pod CIDR 100.65.0.0/16 in this rule (alongside or instead of the node subnet), or use a worker/pod security-group reference, which matches regardless of pod IP.

tests/uat/gcp/cluster-config.yaml
```diff
@@ -51,12 +57,24 @@ network:
...
         allowed:
           - protocol: tcp
           - protocol: udp
           - protocol: icmp
 
+      - name: allow-nats
+        direction: INGRESS
+        sourceRanges:
+          - "10.0.0.0/22"    # system-subnet
+          - "10.0.128.0/17"  # worker-subnet
```
yuanchen8911: GCP allow-nats rule misses the pod ranges

Description: On VPC-native GKE the firewall sees the pod alias IP from the secondary pod ranges, not the node primary CIDRs. The default secondary ranges are system-pods 10.100.0.0/17 and worker-pods 10.100.128.0/17. Restricting allow-nats to only 10.0.0.0/22 and 10.0.128.0/17 means pod-originated 4222 traffic won't match. (Note: GKE's flat pod network historically allowed NATS without any firewall change, so this rule may be redundant.)

Recommended fix: Add the pod secondary ranges 10.100.0.0/17 and 10.100.128.0/17 to sourceRanges, or drop the rule if it's not actually needed on GKE.

tests/uat/gcp/cluster-config.yaml
```diff
@@ -51,12 +57,24 @@ network:
         direction: INGRESS
         priority: 900
         sourceRanges:
-          - "10.0.0.0/8"
+          - "10.0.0.0/22"    # system-subnet
+          - "10.0.128.0/17"  # worker-subnet
```
yuanchen8911: GCP nccl-internal tightening drops pod ranges

Description: Narrowing nccl-internal from 10.0.0.0/8 to [10.0.0.0/22, 10.0.128.0/17] removes the GKE pod secondary ranges (10.100.0.0/17, 10.100.128.0/17) and any other intra-VPC 10.x range the broad rule previously allowed. Pod-sourced east/west traffic governed by this rule would regress. Blast radius is bounded: VPC-native GKE also provisions auto-managed firewall rules (gke-<cluster>-...) that allow node/pod ranges independently, so this may not break pod east/west unless something relies specifically on nccl-internal -- worth verifying, but the tightening is still a real regression risk.

Recommended fix: Keep the nccl-internal source set broad enough to retain the pod ranges (re-add 10.100.0.0/17 and 10.100.128.0/17), and confirm no other 10.x range relied on the prior 10.0.0.0/8 before tightening.

tests/uat/aws/cluster-config.yaml
```diff
@@ -35,6 +35,7 @@ cluster:
         - 216.228.127.128/30
         - 52.41.222.58/32
         - 54.68.11.139/32
+        - 128.77.49.32/30
```
yuanchen8911, line 38: Control-plane allowlist addition is mislabeled

Description: The PR body states the change "added the worker CIDR to the EKS endpoint access list," but the CIDR added to controlPlane.allowedCidrs is 128.77.49.32/30 -- a public/HQ-style CIDR (it also appears as nw3 in the GCP authorizedNetworks), not the worker subnet 10.0.128.0/17. This is unrelated to NATS.

Recommended fix: Correct the PR description, and confirm whether adding 128.77.49.32/30 to the endpoint allowlist is intentionally in scope for this PR.

yuanchen8911 (changes requested):
The first issue is a blocker. Please take a look

tests/uat/aws/cluster-config.yaml
mchmarny: Good catch, fixed. The rule now sources both `10.0.128.0/17` (worker node subnet) and `100.65.0.0/16` (pod subnet), matching the upstream NFD/Prometheus pattern. Confirmed the pod subnet is defined at `network.eks.cidrs.pod`.

tests/uat/gcp/cluster-config.yaml
mchmarny: Agreed it's redundant — dropped the `allow-nats` rule entirely. The existing `nccl-internal` rule already allows all TCP from `10.0.0.0/8` (incl. the pod secondary ranges), so 4222 is already permitted cluster-wide on GKE.

tests/uat/gcp/cluster-config.yaml
mchmarny: Reverted `nccl-internal` back to `10.0.0.0/8` — the tightening would have dropped the pod secondary ranges, as you noted. This restores the prior all-TCP-from-10.x behavior with certainty (no need to enumerate exact pod CIDRs).

tests/uat/aws/cluster-config.yaml
mchmarny: Correct — `128.77.49.32/30` is the HQ CIDR (matches GKE `authorizedNetworks` nw3), not the worker subnet. Fixed the PR description. The addition is intentional: it grants HQ access to the EKS control-plane endpoint for parity with GKE.

yuanchen8911 (approved):
All four findings are addressed:

1. AWS 4222 rule now sources the pod CIDR `100.65.0.0/16` (alongside the node subnet) — pod-originated NATS will match.
2. `nccl-internal` kept at `10.0.0.0/8` (tightening reverted) — pod secondary ranges retained.
3. `allow-nats` dropped as redundant — `nccl-internal` already permits intra-VPC 4222.
4. Control-plane allowlist (`128.77.49.32/30`) description corrected (HQ-access, not worker CIDR).

LGTM.

Issue #1388, opened 2026-06-16
[Bug]: argocd-sync gate uses exists-semantics, not all-Applications (argocd half of #1288)


haarchri:
### Prerequisites

- [x] I searched existing issues and found no duplicates
- [x] I can reproduce this issue consistently
- [x] This is not a security vulnerability (use [Security Advisories](https://github.com/NVIDIA/aicr/security/advisories/new) instead)

### Bug Description

The KWOK Argo CD sync gate `tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml`
(step `assert-all-applications-pass`) uses a name-less chainsaw `assert`, which
has **exists-semantics** — it passes the moment *any one* Application matches the
4-arm terminal-pass predicate, not when *all* do. This is the Argo CD half of
#1288; PR #1290 fixed only the flux side (`flux-sync`, `flux-git-sync`) and left
`argocd-sync` as a documented "Known follow-up" (see ADR-010 "Sync Gate:
All-Resources Semantics"). `tests/chainsaw/kwok/argocd-git-sync/` inherits the
same step verbatim as a faithful sibling, so this covers `argocd-oci`,
`argocd-helm-oci`, and `argocd-git`.

Surfaced while implementing the `argocd-git` lane (#963).

### Impact

Blocking (cannot proceed)

### Component

CLI (aicr)

### Regression?

Unknown / first time using this feature

### Steps to Reproduce

-

### Expected Behavior

- [ ] `argocd-sync` (and its sibling `argocd-git-sync`) fail when *any* Application
      is outside the terminal-pass set at the deadline; zero Applications still fails.
- [ ] The gate does not pass vacuously at apply-time (root must converge first).
- [ ] The gate does not hang on KWOK-pinned `Progressing` apps whose sync `op`
      never reaches `Succeeded` (e.g. `kube-prometheus-stack`).
- [ ] Not flaky under selfHeal churn across repeated runs.
- [ ] `argocd-oci` / `argocd-helm-oci` / `argocd-git` all stay green after the change
      (budget a fidelity-iteration round per ADR-008; bump `KWOK_ARGOCD_SYNC_TIMEOUT`
      to match `KWOK_FLUX_SYNC_TIMEOUT`=500s if the honest gate needs it).

### Actual Behavior

-

### Environment

- AICR version (CLI `aicr version`, API image tag, or commit SHA):
- Install method (release binary / build from source / container image):
[...]

PR #1290, opened 2026-06-10
ci(kwok): add flux-git lane with in-cluster gitea

haarchri:
## Summary

Adds the `flux-git` KWOK deployer lane: an in-cluster Gitea receives the filesystem bundle via `git push`, and Flux reconciles it from a `GitRepository` source — closing the Git-source round-trip gap in the deployer matrix. Along the way this fixes the chainsaw sync-gate exists-semantics bug (#1288).

## Motivation / Context

The KWOK matrix only validated OCI round-trips (`argocd-oci`, `argocd-helm-oci`, `flux-oci` — ADR-008 / PR #956). The flux deployer's **Git output shape** (`sources/gitrepo-*.yaml` GitRepository CRs, `branch: main` default, GitRepository `sourceRef` on local-chart HelmReleases) had zero end-to-end coverage, so regressions in it would ship undetected. The first live run of the new lane immediately surfaced two such masked bugs — validating the premise. Design rationale in ADR-010 (`docs/design/010-kwok-git-source-lanes.md`).

Fixes: #1288
Related: #963 (flux half; `argocd-git` follows on the same Gitea infra), #843, #1285, https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/issues/1184

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [x] Build/CI/tooling

## Component(s) Affected

- [ ] CLI (`cmd/aicr`, `pkg/cli`)
- [ ] API server (`cmd/aicrd`, `pkg/server`)
- [ ] Recipe engine / data (`pkg/recipe`)
- [ ] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [ ] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [ ] Validator (`pkg/validator`)
- [ ] Core libraries (`pkg/errors`, `pkg/k8s`)
- [x] Docs/examples (`docs/`, `examples/`)
- [x] Other: KWOK test infra (`kwok/`, `tests/chainsaw/kwok`), CI workflows (`.github/`), conformance validators (`validators/`), component values (`recipes/components`, `recipes/checks`)

## Testing

```bash
unset GITLAB_TOKEN
make qualify                                          # PASS ("Codebase qualification completed")
[...]

Issue #1288, opened 2026-06-10
[Bug]:  ci(kwok): chainsaw sync gates pass when ANY resource converges, not ALL


haarchri:
### Prerequisites

- [x] I searched existing issues and found no duplicates
- [x] I can reproduce this issue consistently
- [x] This is not a security vulnerability (use [Security Advisories](https://github.com/NVIDIA/aicr/security/advisories/new) instead)

### Bug Description

The KWOK GitOps sync gates (`tests/chainsaw/kwok/flux-sync`,
`tests/chainsaw/kwok/argocd-sync`) use name-less chainsaw `assert` blocks to
express "ALL resources of kind X must reach a terminal-pass state". A
name-less chainsaw `assert` actually has **exists-semantics**: it succeeds as
soon as **any** resource of that kind matches the predicate. The gates
therefore pass while most of the fleet is still converging — or never
converges at all.

### Impact

Blocking (cannot proceed)

### Component

CLI (aicr)

### Regression?

Yes, this worked before (please specify version below)

### Steps to Reproduce

Run any flux-oci KWOK cell and watch the HelmReleases while the sync gate
runs:

```
$ kubectl get helmrelease -n flux-system
NAME                            AGE   READY   STATUS
aws-ebs-csi-driver              9s    True    Helm install succeeded ...
aws-efa                         9s    False   dependency 'flux-system/aws-ebs-csi-driver' is not ready
cert-manager                    9s    False   dependency 'flux-system/aws-efa' is not ready
... (10 more False)
```

The `assert-all-helmreleases-ready` step passes the moment the FIRST release
hits `Ready=True` — the remaining 12 are still `False` in the `dependsOn`
chain. Verified on `eks-training`: the gate even passes when one of those
releases is **permanently failing** (`InstallFailed` — see the companion
nvidia-dra-driver-gpu issue, which this bug masked).

`verify_pods` provides no backstop for this class: a failed Helm install
creates no pods, and zero pods means zero Pending pods.

### Expected Behavior

Use the existence + error-polarity pattern the ArtifactGenerator step already
demonstrates:

1. Keep an existence `assert` (resource kind + namespace, no status
   predicate) so zero resources fails instead of passing vacuously.
2. Replace the terminal-pass `assert` with an `error:` op carrying the
   **De Morgan negation** of the predicate — chainsaw's `error` polls until
   NO resource matches the bad state, which is the actual "all resources
   converged" semantics.
[...]

PR #1397, opened 2026-06-21
fix(test): argocd-sync gate: use all-semantics, handle health-gated apps

rsd-darshan:
## Summary

Fixes issue #1388. Replaces exist-semantics assert with inverted-polarity error: to enforce that ALL Applications reach terminal-pass state, not just any one.

## Changes

1. **Added staged assert-root-app-succeeded step** — gates fleet sweep against apply-time races where kubectl apply creates root with empty status and the sweep evaluates before children materialize
2. **Switched assert-all-applications-pass to inverted-polarity error:** blocks when ANY Application fails predicate (all-semantics vs exist-semantics)
3. **Removed operationState.phase==Succeeded from per-child check** — fixes hang on health-gated apps (e.g., kube-prometheus-stack) that stay Progressing on KWOK but are genuinely synced
4. **Increased timeout from 5m to 8m** — gives the all-semantics gate (which waits for every Application, not just the first) more budget than the old exists-semantics check needed
5. **Bumped `KWOK_ARGOCD_SYNC_TIMEOUT` default from 300 to 480** in `validate-scheduling.sh` — this env var feeds `--assert-timeout`, which overrides `spec.timeouts.assert` at invocation, so the YAML's 8m bump was a no-op until the script's default was raised too

## Root Cause

The original code used a name-less `assert:` block that passed when ANY Application matched the predicate (exist-semantics). Two failure modes emerged:
- **Apply-time race**: Root app created with empty status; sweep evaluates before children exist → vacuous pass  
- **Health-gate hang**: Apps with op=Running waiting for Pod readiness never reach op=Succeeded → timeout

## Test Coverage

- Modified `tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml`, `tests/chainsaw/kwok/argocd-git-sync/chainsaw-test.yaml`, `kwok/scripts/validate-scheduling.sh`, `docs/contributor/tests.md`
- No Go code changes
- Syntax validated

## References

- ADR-010 "Sync Gate: All-Resources Semantics"  
- Issue #1061 (operationState.phase rationale)
- Issue #1050 (migration introducing apply-time race)
- Parent issue #1288 (flux half fixed in #1290)

### DIFF
--- a/tests/chainsaw/kwok/argocd-git-sync/chainsaw-test.yaml
+++ b/tests/chainsaw/kwok/argocd-git-sync/chainsaw-test.yaml
@@ -29,12 +29,16 @@
 # which is exactly what Argo CD's repo-server clones from.
 #
 # Pass predicate — every Application in the `argocd` namespace must match
-# one of these 4 terminal-pass states (mirroring argocd-sync):
+# one of these 4 terminal-pass states (mirroring argocd-sync). Note:
+# operationState.phase==Succeeded is validated only on the root Application
+# (assert-root-app-succeeded step) to close apply-time races; per-child
+# checks omit it to avoid hanging on health-gated apps like
+# kube-prometheus-stack that stay op=Running forever.
 #
-#   1. Synced + Healthy                                   — canonical pass
-#   2. Synced + Progressing                               — KWOK pod-readiness gap (ADR-008)
-#   3. OutOfSync + Healthy + operationState=Succeeded     — operator mutation (ClusterPolicy, DeviceClass, etc.)
-#   4. Synced + Degraded + operationState=Succeeded       — health controller divergence after successful op
+#   1. Synced + Healthy        — canonical pass
+#   2. Synced + Progressing    — KWOK pod-readiness gap (ADR-008)
+#   3. OutOfSync + Healthy     — operator mutation (ClusterPolicy, DeviceClass, etc.)
+#   4. Synced + Degraded       — health controller divergence after successful op
 apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
@@ -58,8 +62,16 @@ spec:
     - name: repoURL
       value: ($values.repoURL || '')
   timeouts:
-    # KWOK_ARGOCD_SYNC_TIMEOUT default. Matches argocd-sync.
-    assert: 5m
+    # KWOK_ARGOCD_SYNC_TIMEOUT default. Matches argocd-sync (8 minutes for
+    # all-semantics gate; see issue #1288).
+    # `error` needs the same budget as `assert`: assert-all-applications-pass
+    # uses error-polarity (poll until NO Application matches the bad-state
+    # predicate), and chainsaw's default error timeout is only 10s — far
+    # less than a fleet of Applications needs to converge. Both are
+    # overridden per-run by the driver (validate-scheduling.sh passes
+    # --assert-timeout AND --error-timeout from KWOK_ARGOCD_SYNC_TIMEOUT).
+    assert: 8m
+    error: 8m
   # See argocd-sync for the skipDelete rationale (KWOK has no
   # kube-controller-manager to finalize the auto-created test namespace).
   skipDelete: true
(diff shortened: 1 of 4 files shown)


tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml
```diff
@@ -68,11 +70,13 @@ spec:
...
-    # Application has to render the child Apps before they begin
-    # reconciling. 5 minutes is conservative for KWOK pod-readiness sim.
-    assert: 5m
+    # KWOK_ARGOCD_SYNC_TIMEOUT default (480s, matching the bash driver's
+    # default below). The argocd-helm-oci wrapper takes slightly longer to
+    # converge than argocd-oci because the parent Application has to render
+    # the child Apps before they begin reconciling. 8 minutes accommodates
+    # the additional latency of the more honest gate (all-semantics vs
+    # exists-semantics) without racing; see issue #1288.
+    assert: 8m
```
mchmarny, line 85: 🔴 Blocking: this needs `error: 8m` alongside `assert: 8m`.

`assert-all-applications-pass` was converted from `assert:` to `error:`, and chainsaw bounds `error:` operations by `timeouts.error` — not `timeouts.assert`. The repo default is `error: 10s` (`tests/chainsaw/chainsaw-config.yaml:26`), so the new all-semantics gate gets ~10s to converge instead of 8m. On any real cluster (apps need ~30s+ to reach Synced+Healthy) the step fails spuriously, and the 5m→8m bump is a no-op for the converted step.

The flux tests already hit this and set both — see `tests/chainsaw/kwok/flux-sync/chainsaw-test.yaml:70` (`error: 8m`) with a comment spelling out exactly this reason. Mirror it:

```yaml
  timeouts:
    assert: 8m
    error: 8m
```

tests/chainsaw/kwok/argocd-git-sync/chainsaw-test.yaml
```diff
@@ -58,8 +62,9 @@ spec:
     - name: repoURL
       value: ($values.repoURL || '')
   timeouts:
-    # KWOK_ARGOCD_SYNC_TIMEOUT default. Matches argocd-sync.
-    assert: 5m
+    # KWOK_ARGOCD_SYNC_TIMEOUT default. Matches argocd-sync (8 minutes for
+    # all-semantics gate; see issue #1288).
+    assert: 8m
```
mchmarny, line 73: 🔴 Blocking: same `timeouts.error` gap as argocd-sync — add `error: 8m` here too. The `error:`-polarity `assert-all-applications-pass` step falls back to the default `error: 10s` without it. Mirror `flux-git-sync` (`error: 8m`).

tests/chainsaw/kwok/argocd-sync/chainsaw-test.yaml
```diff
@@ -105,41 +109,82 @@ spec:
...
-                # Arms:
-                #   1. Synced + Healthy        — canonical pass
-                #   2. Synced + Progressing    — KWOK pod-readiness sim gap (ADR-008)
-                #   3. OutOfSync + Healthy     — operator mutation (ClusterPolicy, DeviceClass)
-                #   4. Synced + Degraded       — health controller divergence after a successful op
-                (operationState.phase == 'Succeeded' && ((sync.status == 'Synced' && health.status == 'Healthy') || (sync.status == 'Synced' && health.status == 'Progressing') || (sync.status == 'OutOfSync' && health.status == 'Healthy') || (sync.status == 'Synced' && health.status == 'Degraded'))): true
+                # NOT (any of the 4 pass states) == NOT pass
+                # Negated via DeMorgan: NOT (A OR B OR C OR D) == (NOT A AND NOT B AND NOT C AND NOT D)
+                # Encode as: all 4 conditions false => application fails predicate
+                ((sync.status == 'Synced' && health.status == 'Healthy') || (sync.status == 'Synced' && health.status == 'Progressing') || (sync.status == 'OutOfSync' && health.status == 'Healthy') || (sync.status == 'Synced' && health.status == 'Degraded')): false
```
mchmarny, line 193: nit: the DeMorgan comment describes `NOT(A∨B∨C∨D) == (¬A∧¬B∧¬C∧¬D)`, but the encoding below keeps the disjunction and asserts it `== false` (simpler and correct). The comment is vestigial/misleading — worth trimming so the next reader isn't hunting for a DeMorgan expansion that isn't there.

mchmarny (changes requested):
The design is sound and correctly mirrors the proven flux-sync/flux-git-sync `error:` inverted-polarity pattern — the staged `assert-root-app-succeeded` gate guarding the zero-apps vacuous-pass is the right structure, and catching that the bash `KWOK_ARGOCD_SYNC_TIMEOUT` default made the YAML bump a no-op is a sharp fix.

One blocker, though: the same class of no-op was reintroduced on the YAML side. The converted `assert-all-applications-pass` step uses `error:` polarity, which chainsaw bounds by `timeouts.error` (default `10s` in `tests/chainsaw/chainsaw-config.yaml`), not `timeouts.assert`. Both files need `error: 8m` added — the flux tests set it explicitly for exactly this reason. Without it the new all-semantics gate gets ~10s and fails spuriously on real clusters. Inline comments on both `timeouts` blocks, plus one nit on a vestigial DeMorgan comment.

Also worth noting: the argocd-sync / argocd-git-sync KWOK lanes don't appear to have run on this SHA (only Label/Auto-Merge checks present) and the branch is `behind` main — rebasing and running those lanes would catch the timeout gap directly. Not blocking beyond the code fix above.

rsd-darshan:
@mchmarny good catch on the error timeout — added `error: 8m` to both files and `--error-timeout` in the driver, squashed and rebased onto current main. Should be ready for another look.

mchmarny (approved):

Issue #1197, opened 2026-06-04
inference-perf: stochastic worker-stall / throughput degradation on EKS H100 at 2048 concurrency


yuanchen8911:
Part of epic #1043.

## Summary

The `inference-perf` validation phase shows **large run-to-run performance fluctuation on EKS H100** clusters at the 2048-concurrency saturation point (256/GPU × 8 vLLM decode workers). A subset of runs exhibit a **severe worker-stall / throughput-degradation** in which TTFT p99 jumps from ~700 ms (healthy) to **4–77 s** and aggregate throughput drops ~30–85% (123k → 18–84k tok/s), while the same deployment on **GKE H100 is consistently clean**. The behavior is **stochastic and time-varying** — entire windows pass cleanly, then a later window fails repeatedly with byte-identical configuration.

This is the underlying performance issue surfaced by (now-closed) #1192. The gate-hardening work in PR #1196 (relax TTFT p99 → `<= 2000 ms` + pin AIPerf inputs) removes the *false-negative knee-jitter* but **does not fix the severe stall** — a 4–77 s degradation fails regardless of the ceiling. Filing this to track the root cause and the long-term fix separately from the gate change.

## Environment

- **Cluster:** EKS, H100 GPU nodes (8× H100 per node), DRA-based 1-GPU-per-worker.
- **Workload:** `DynamoGraphDeployment` — Dynamo frontend + 8 vLLM decode workers; AIPerf benchmark Job (`aicr-validation` namespace) drives 2048 concurrent requests.
- **Dynamo runtime:** 1.0.2 (post #1193; the 0.9.0 `futures-util` panic is a separate, fixed issue).
- **Routing:** default round-robin (capacity-blind).
- **Comparison baseline:** GKE H100, identical recipe/overlay path.

### 1. Combined gate-hardening image (PR #1193 + #1196), 3× EKS H100 + 3× GKE H100 @ 2048

| Cluster | Run | Throughput (tok/s) | TTFT p99 | Peak Running | Verdict |
|---|---|---|---|---|---|
| EKS H100 | 1 | 18,868 | **77,106 ms** | 986 | ❌ severe stall |
| EKS H100 | 2 | 119,483 | 780 ms | 256 | ✅ |
| EKS H100 | 3 | 84,230 | **7,374 ms** | 178* | ❌ degraded |
| GKE H100 | 1 | 139,790 | 304 ms | 257 | ✅ |
[...]

Issue #1043, opened 2026-05-26
[Epic]: Performance-phase constraints: empirically-grounded thresholds for testbed-blocked overlays


mchmarny:
## Goal

Every overlay shipped by AICR must carry a performance-phase constraint anchored in empirically-grounded thresholds from a real testbed, so `aicr validate` ships a meaningful runtime gate instead of a placeholder. Overlays without a constraint either get one or are explicitly exempted in `recipes/overlays_validation_floor_test.go` with the tracking issue.

Parent initiative: #1041.

## Success criteria

1. Every entry in `recipes/overlays_validation_floor_test.go` flagged by `AICR_VALIDATION_FLOOR_STRICT=1` either has the required constraint or carries an exemption pointing at the issue tracking the testbed.
2. Each added constraint uses thresholds derived from a real run on the corresponding service (no placeholder values).
3. The validator the constraint invokes (e.g., `inference-perf`, `nccl-all-reduce-bw`) actually executes against the workload variant the overlay deploys — if it doesn't, the gap is filed under the validator capability Epic instead.

## Scope

In scope:

- Adding `nccl-*` / `inference-perf` / future performance constraints to existing strict-floor-flagged overlays once their testbed lands.
- Re-running the threshold derivation when a service or accelerator generation changes.

Out of scope (other epics under #1041):

- New overlays themselves (covered under the overlay-coverage Epic).
- Validator changes that would make an existing skip path actually run on a NIM / new workload variant (validator capability Epic).

New testbed-blocked performance gaps should be filed as standalone issues and attached here.

yuanchen8911:
#1413 tracks the RoCE NCCL NET test/validation (impl #1428): enables the nccl-all-reduce-bw-net gate on ConnectX RoCE (GB300 p6e-gb300r) and captures a RoCE-GB300 baseline, replacing the current EFA-inherited placeholder threshold.

yuanchen8911:
Closing: all 14 sub-issues are closed. Performance-phase floors are in place for the active overlays (EKS/GKE/AKS H100 NCCL via #1005/#442/#1695, GB200 NET/NVLS variants, inference-perf gates via #448/#937/#1116 and the reliability fixes #1181/#1192/#1197, RoCE NET support via #1413). The SKU-agnostic threshold-model follow-ups (#1254 inference, #1256 training) are closed as not planned — deferred until smaller-GPU SKUs need support.

Issue #1374, opened 2026-06-15
feat(validator): make inference-perf routing strategy configurable (AICR_INFERENCE_PERF_ROUTER_MODE)


yuanchen8911:
## Summary

The inference-perf performance validator hardcodes the Dynamo frontend router mode to `DYN_ROUTER_MODE: kv` (in `validators/performance/testdata/inference/dynamo-deployment.yaml`). Make it configurable like the other `AICR_INFERENCE_PERF_*` methodology knobs, so operators can A/B the routing strategy (e.g. `kv` vs `least-loaded`) without editing the image.

## Motivation

Dynamo v1.2 adds frontend router strategies (`kv`, `round-robin`, `random`, `least-loaded`, `device-aware-weighted`). Today there's no way to characterize their throughput/TTFT impact through the validator — the only routing lever is recipe `inference-routing-mode` (`dynamo-router` | `gateway-epp`), which selects the routing *layer*, not the dynamo-router sub-strategy. The strategies only apply under `inference-routing-mode=dynamo-router`.

## Proposed change

- Add env knob **`AICR_INFERENCE_PERF_ROUTER_MODE`** (default `kv`), validated against the Dynamo frontend enum (`kv`, `round-robin`, `random`, `least-loaded`, `device-aware-weighted`); invalid value fails closed (`ErrCodeInvalidRequest`), consistent with the other perf knobs.
- Template the value into the DGD as `DYN_ROUTER_MODE: ${ROUTER_MODE}` (replacing the hardcoded `kv`).
- Only affects the `dynamo-router` path (gateway-epp routes via the EPP).

## Notes

A working implementation already exists locally and was used to A/B `least-loaded` vs `kv` on the aicr3 H100 cluster (built as `aicr-validators/performance:lb-exp`). This issue tracks landing it as a first-class, documented knob (catalog `env` doc + cli/validation docs) and wiring it through release image builds. Can submit the PR.

## Type of Change

- New feature (validation methodology knob)

yuanchen8911:
## Validated end-to-end on EKS H100 (aicr3) — `kv` vs `least-loaded`

Implemented the knob (`AICR_INFERENCE_PERF_ROUTER_MODE`, default `kv`, validated against the Dynamo frontend enum; `${ROUTER_MODE}` placeholder in the inference-perf DGD template) and ran an A/B on the same cluster/workload.

[...]

PR #1399, opened 2026-06-22
feat(validation): switch inference-perf frontend routing to least-loaded

yuanchen8911:
## Summary

Switch the `inference-perf` validation workload's Dynamo frontend from KV-cache-aware routing (`DYN_ROUTER_MODE=kv`) to load-aware least-loaded routing (`DYN_ROUTER_MODE=least-loaded`).

## Motivation / Context

The inference-perf frontend routed by KV overlap, which can keep sending a transiently-slow worker its full share of requests. On EKS H100 at the 2048-concurrency saturation knee this produced stochastic worker-stall / throughput degradation — TTFT p99 spiking from ~700 ms to 4–77 s and aggregate throughput dropping 30–85% — while GKE H100 stayed consistently clean. `least-loaded` balances by each worker's active in-flight load, so a backed-up worker stops receiving new requests until it drains, removing the positive-feedback loop behind the stall.

**Why least-loaded is the best fit for the AICR performance benchmark.** The performance phase drives the workload with AIPerf under deliberately uniform, reproducible conditions — a single served model, a pinned prompt pool, fixed input/output token counts (stddev 0), and greedy decoding (`temperature: 0`) — against a homogeneous pool of identical single-GPU vLLM workers (one H100 per worker via DRA). Under those conditions KV-overlap (`kv`) routing has no advantage to exploit: there is no worker heterogeneity to weigh, and the synthetic prompts carry no meaningful cross-request prefix-cache reuse for the router to capitalize on. What KV routing *does* do is bias requests toward an "overlap-matching" worker, which lets a transiently-slow worker keep accumulating load — exactly the saturation-knee failure mode in #1197. Least-loaded optimizes for the one signal that actually governs tail latency in this benchmark — instantaneous per-worker queue depth — so it is the most appropriate mode for AICR's AIPerf-driven throughput/TTFT measurement. (KV-aware routing remains the right default for production traffic with real prompt-prefix reuse; this change scopes only the benchmark workload.)

Fixes: N/A
Related: #1197, #1043 (epic)

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- One-line behavior change: `DYN_ROUTER_MODE` on the Frontend container in `validators/performance/testdata/inference/dynamo-deployment.yaml` flips `kv` → `least-loaded`.
- `least-loaded` is a first-class Dynamo 1.2 router mode (set via the same frontend env var). No image/runtime bump required — the manifest already targets Dynamo 1.2.0.
- Workers keep the vLLM ZMQ KV-cache event publisher; least-loaded simply does not consume those events. Left in place to avoid widening scope and to keep `kv` mode trivially restorable.
- The `gateway-epp` routing path (`--router-mode direct` sidecars, external endpoint-picker) is unchanged and out of scope.
- Docs updated to match: `docs/user/validation.md`, `docs/contributor/validator.md`, and the investigation report `docs/contributor/inference-perf-fluctuation.md`.

## Testing

```bash
yamllint validators/performance/testdata/inference/dynamo-deployment.yaml   # clean
```

YAML + testdata/docs only — no Go source changed, so test/e2e/Go-lint cannot regress. End-to-end behavior (degradation no longer reproduces under least-loaded on EKS H100) needs a cluster run before this leaves draft.

## Risk Assessment

- [x] **Low** — Single env-value change in a validator testdata manifest, easy to revert; `kv` mode remains available.

**Rollout notes:** Affects only the inference-perf validation workload, not user-deployed recipes/bundles.

## Checklist

- [x] Linter passes (yamllint on changed manifest)
- [ ] I added/updated tests for new functionality — N/A (testdata value change)

### DIFF
--- a/docs/contributor/inference-perf-fluctuation.md
+++ b/docs/contributor/inference-perf-fluctuation.md
@@ -25,8 +25,9 @@ H100/RTX clusters showed:
 **Mitigations shipped/in-PR:** dynamo runtime image bump 0.9.0→1.0.2 (#1193, the
 panic fix); reproducible gate — TTFT `<= 2000` ms + pinned AIPerf inputs (#1196);
 centralized dynamo image reference + drift guard (#1194).
-**Long-term:** Dynamo 1.2 KV-cache-aware routing with live worker KV events,
-plus a sub-knee, throughput-primary gate.
+**Long-term:** Dynamo 1.2 load-aware least-loaded routing
+(`DYN_ROUTER_MODE=least-loaded`, shipped for #1197), plus a sub-knee,
+throughput-primary gate.
 
 ---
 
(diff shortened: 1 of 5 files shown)


mchmarny (commented):
Tightly-scoped, well-reasoned change. The env flip is on the Frontend container, the worker's vLLM ZMQ KV-events publisher is correctly retained (matching the "least-loaded doesn't consume them, kept for trivial revert" rationale), and all three doc files plus the `resolveRoutingMode` comment are updated coherently. CI is green.

One concern blocks my approval — not the diff itself, but verification:

- The Testing section gates this on an EKS H100 cluster run ("degradation no longer reproduces under least-loaded ... needs a cluster run before this leaves draft"), but the PR is **no longer a draft** and no run is recorded. The whole premise — that least-loaded stops the worker-stall, and that `least-loaded` is even an accepted `DYN_ROUTER_MODE` value for the Dynamo 1.2 frontend — rests on that run. A bad enum value would only surface at frontend startup on a real cluster, so I'd want the deploy + non-reproduction confirmed before merge. Everything is reversible (`kv` restorable), so this is a "verify, then merge," not a redesign.

Non-blocking nit: CodeRabbit flagged the tension with #1374 (make `DYN_ROUTER_MODE` configurable via `AICR_INFERENCE_PERF_ROUTER_MODE` rather than hardcoding). Hardcoding is consistent with the current template, so fine to keep here and let #1374 generalize it later — worth a one-line note in the PR that this is intentional.

Holding as COMMENT rather than approve pending the cluster confirmation.

yuanchen8911:
Validated on EKS H100 (aicr3) before it left draft. inference-perf passed at 8-GPU/2048 concurrency: 136,054 tok/s, TTFT p99 703 ms. The stall from #1197 didn't reproduce. least-loaded does a much better job than round-robin and kv-aware here — performance is much more consistent and predictable, and better overall.

mchmarny (approved):

Issue #1405, opened 2026-06-22
GP5 — GitHub Pages publish CI (merge-to-main cadence)


njhensley:
**Parent:** #1400 · **Labels:** `enhancement`, `area/ci` (auto, `.github/**`) · **Size:** M · **Depends on:** GP4 (the generator), GP3 (read-only bucket access)

**Goal.** Establish the repo's **first GitHub Pages** pipeline: a merge-to-main-cadence CI job that runs the GP4 generator against the GCS bucket and publishes the static site via `upload-pages-artifact` + `deploy-pages`.

**Scope.**
- New `.github/workflows/evidence-dashboard-publish.yaml`: on merge-to-main (and `workflow_dispatch`), authenticate to GCS read-only (GP3 read SA via the existing federation), run the GP4 generator to produce the site + JSON (`index.json` + `series/*.json`), then `actions/configure-pages` → `actions/upload-pages-artifact` → `actions/deploy-pages`, with `permissions: pages: write, id-token: write` (the **first** Pages workflow in the repo — a repo-wide grep for `upload-pages-artifact`/`deploy-pages`/`pages: write` returns nothing today; Fern publishes to docs.nvidia.com via `publish-fern-docs.yml`, a separate surface).
- Because the generator is byte-deterministic, the job can run as a straightforward publish (no drift-PR needed for the *site*); a determinism check can be wired so a non-reproducible build fails loudly.
- Pin all action SHAs (repo norm, e.g. `uat-gcp.yaml:109,:115`); bounded `timeout-minutes`; `runs-on: ubuntu-latest`.
- Establish the Pages **environment**; a custom domain is deferred (see Deferred).

**Out of scope.** The generator/model (GP4); the ingest (GP2); the bucket/SA (GP3); the docs page (GP6). A custom domain.

**Key files.** New `.github/workflows/evidence-dashboard-publish.yaml`; the GP4 generator (invoked); GCS read auth `uat-gcp.yaml:53-54,:109,:115`; HTML-artifact precedent `demos/*.html`. (No existing Pages workflow to model on — this child establishes the pattern.)

[...]

Issue #1404, opened 2026-06-22
GP4 — Consensus model + deterministic generator (the viz)


njhensley:
**Parent:** #1400 · **Labels:** `enhancement` (**adds** a `tools/<generator>/**` glob to `.github/labeler.yml`, e.g. `area/ci` or a new `area/tools` — `tools/**` matches no glob today) · **Size:** L · **Depends on:** GP3 (bucket to read from); soft on GP2 (real data to render — can develop against fixtures)

**Goal.** A deterministic Go generator (`pkg/<corroborate>` model + `tools/<generator>` CLI, modeled on `tools/bom`) that pulls all CTRFs + metadata from GCS, computes the consensus model **counting distinct verified signers, never builds**, and emits a deterministic **`index.json` + per-recipe `series/*.json`** + a self-contained static HTML/CSS/JS renderer that reads them.

**Scope.**
- **Consensus model package** (`pkg/<corroborate>`): given the set of `(signer, run, phase, row, result, aicr_version, k8s_version, attested_at)` tuples for a recipe coordinate, compute per row: `S_pass`/`S_fail` over **distinct allowlisted signers** (latest-per-signer in scope), then the cell state (CONFIRMED / SINGLE / CONTESTED / FAILING / UNTESTED) per the *Decision* rules. Bucket each signer's latest CTRF status **totally and explicitly** (`pkg/validator/ctrf/types.go:26-39`): `passed`→`S_pass`; `failed`/`other`→`S_fail`; `skipped`/`pending`→**NOT-RUN** for that signer+row (a coverage gap, never a passing corroboration — so a `skipped`/`pending` row can neither promote to CONFIRMED nor suppress a CONTESTED). CONFIRMED requires ≥ 2 distinct allowlisted signers and empty `S_fail`; an unknown-verified community signer is a non-corroborating "reported" dot. Phase rollup `CONTESTED > FAILING > UNTESTED > SINGLE > CONFIRMED`. The row set is the **union of CTRF Names** across all sources (`ctrf/types.go:114`); a missing Name is NOT-RUN for that signer.
- **Source class** from the verified signer + the GP1 allowlist (first-party issuer pin / community / partner sections); never a free flag.
- **Recency/version drift:** default scope latest-per-signer; aicr-version and k8s `major.minor` (from `meta.json`, sourced from the bundle predicate, never the clock) carried per dot and exposed as facets in the emitted site.
[...]

Issue #1283, opened 2026-06-09
RQ1 — recipe-health ⇄ TestGrid Evidence deep-link


njhensley:
**Parent:** #1265 · **Labels:** `enhancement`; `area/docs` auto-assigns from the `docs/**` change. **No net-new `pkg/recipe` helper is introduced by RQ1** — the shared mapping function is shipped and owned by the TestGrid epic (§3), so RQ1 carries no `labeler.yml`/`CODEOWNERS` add of its own; it imports the TestGrid epic's package. · **Size:** M · **Depends on:** #1224 children #1225–#1230 (specifically **#1229** `tools/health` generator and the `pkg/health.Compute` + `docs/user/recipe-health.md` they introduce — none exist at HEAD), **the TestGrid epic's shipped mapping function + coordinate/URL scheme (TG2/TG4a/TG6)**, and **the TestGrid epic's coordinate-presence query surface (TG4a)** that RQ2 keys off. RQ1 must land **after** #1224 V1 ships its `pending` column.

**Goal.** Turn the literal `pending` Evidence column (`docs/design/009-recipe-health-tracking.md:263`) into a stable, deterministic deep-link to each recipe's TestGrid `(group, dashboard, tab)` coordinate — without duplicating any TestGrid data into the matrix — as the #1224 v1.1 Evidence rendering.

**Scope.**
- **Consume the TestGrid epic's shipped recipe → coordinate function** (§2): `service` → group, `accelerator`-`os` → dashboard, `intent`[-`platform`] → tab, applying the TestGrid epic's path-sanitize rules (lowercase; collapse `/\:;[]{}, `+whitespace to `-`; dedupe; trim; empty→`unknown` for required segments; **bare-intent tab when platform absent**). This is **one function**, owned by the TestGrid epic (TG2/TG6), imported here and by #1224's join key — not re-implemented. Confirming the import path against the TestGrid epic is a **hard dependency**, not a note: there must be a single source of truth.
[...]

Issue #1273, opened 2026-06-09
TG7 — Operability (worker/API/host-cluster liveness + runbook + failure surfacing)


njhensley:
**Parent:** #1263 · **Labels:** `enhancement`, `area/infra` (auto, manifests under
`infra/testgrid/**`) · **Size:** M · **Depends on:** TG1 (host cluster), TG3 (workers),
TG4a (API), TG5 (publish path)

**Goal.** Make the running TestGrid instance operable by a small team — the
"operate" half of the epic title. TG7 is the **single accountable owner** of
operability for the running instance; TG1/TG3/TG4a/TG5 deliver the probes/metrics/signals
it monitors and gates on. Cover liveness/health of the host cluster, the
long-running workers, and the API; a runbook for the common failure modes; and
non-silent surfacing of publish-path and produced-but-unpublished failures.
CLAUDE.md's design principles make observability mandatory (structured logging,
metrics).

**Scope.**
- **Host-cluster + worker health:** liveness/readiness for the updater/tabulator/
  summarizer GKE pods (long-running, resumable `queue/*.json`); a check that a
  worker loop is making progress (queue not wedged), surfaced via structured
  logs/metrics; host-cluster/node-pool health (the TG1 cluster).
- **API health:** verify TG4a's liveness/readiness probes are wired and monitored.
- **Publish-path failure surfacing:** the TG5 ingestion job must not silently swallow
  a publish failure — emit a clear failure signal (job failure + structured log) so
  an operator sees a dropped bundle.
- **Produced-but-unpublished reconciliation (the dynamic-clusters→TG5→TG3 seam).** A lightweight
  check: for each completed UAT run with an emitted bundle (the dynamic-clusters side succeeded),
  assert a TestGrid column **materializes within a bounded number of worker cycles**;
  if not, emit a warning. This closes the cross-epic gap where a bundle exists but
  never reaches the feed (ingestion failed, queue wedged) — otherwise invisible for
  up to the 1-week `alertStaleResultsHours=168` window. TG7 owns detecting this gap
  (the TestGrid epic owns the feed).
- **Runbook** (`docs/contributor/` or `infra/testgrid/RUNBOOK.md`): stuck queue,
  wedged worker, failed publish, produced-but-unpublished bundle, stale columns,
  config-gen drift, **host cluster down / node pool unschedulable** — symptoms,
  diagnosis, recovery (resume from `queue/*.json`, re-run config-gen, re-trigger TG5
[...]

PR #1409, opened 2026-06-22
feat(recipe): shared coordinate mapping + taxonomy spec (TG6)

mchmarny:
## Summary

Adds `pkg/recipe.CoordinateFor` — the canonical, pure recipe→board-coordinate mapping (`group=service`, `dashboard=accelerator-os`, `tab=intent[-platform]`) — plus its golden test and the `docs/design/012` taxonomy spec that pins the mapping, stable URL form, column-metadata key schema, and the ADR-009 / #1224 reconciliation.

## Motivation / Context

This is **TG6 — the Wave 0 linchpin** of the evidence-dashboard tracks. Per the epic plan, one shared mapping function gates four workstreams: GP4 (#1404), GP5 (#1405), RQ1 (#1283), and the entire TestGrid track (#1266–#1273) all import `CoordinateFor` rather than re-deriving the coordinate. Landing it first lets those consumers build against a fixed contract instead of a stub.

Scope is deliberately the **function + spec**. The user-facing `docs/user/` TestGrid page is **deferred** (soft dependency on TG4a's served URL form, which doesn't exist yet), so #1272 stays open after this merges.

Fixes: N/A
Related: #1272 (partial — docs/user page deferred), epic #1263, unblocks #1404 / #1405 / #1283

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- `CoordinateFor` is **pure**: no clock, no maps, no registry, no I/O — same `Criteria` always yields the same `Coordinate`. It consumes already-resolved `Criteria` and **never parses `metadata.name`** (the recipe name is accelerator-first; the coordinate is service-first — they stay independently correct).
- **Fails closed**: `service`/`accelerator`/`os`/`intent` are required; a nil `Criteria`, empty value, or the `"any"` wildcard returns `ErrCodeInvalidRequest` naming the offending dimension. Platform is the only optional sub-segment — dropped when unset, never `unknown`-substituted.
- The golden test enumerates the runtime `CriteriaRegistry` union (not a closed enum), so a new const or `--data`-contributed value is covered without editing the table.
- The spec doc reflects the **current** repo state: `pkg/health` + `docs/user/recipe-health.md` have shipped (structural matrix), with the Evidence column a literal `pending` placeholder today — the coordinate deep-link is what RQ1 will wire.

## Testing

```bash
make qualify   # passed (exit 0)
go test -race ./pkg/recipe/...          # PASS
golangci-lint run -c .golangci.yaml ./pkg/recipe/...   # 0 issues
```

Coverage: `pkg/recipe/coordinate.go` — `CoordinateFor` / `Path` / `String` / `requireConcrete` all 100%; `pkg/recipe` package total ~86.8% (well above the 75% floor).

## Risk Assessment

- [x] **Low** — Additive only (one new file + tests + one design doc); no existing code paths touched; pure function, easy to revert.

**Rollout notes:** N/A — no consumers wired yet; this PR only publishes the contract.

## Checklist

---

> **Note for the epic author:** #1272's body example `h100-eks-ubuntu-training → eks/h100-ubuntu/training-kubeflow` contradicts its own empty-platform→bare-intent rule (that overlay has no platform). This PR implements the rule — that recipe maps to `eks/h100-ubuntu/training` — treating the `-kubeflow` suffix in the example as a typo. Flag if that was intentional.


docs/design/012-recipe-coordinate-mapping.md
```diff
@@ -0,0 +1,232 @@
...
+   the facet and the default scope are required to keep the benefit without
+   paying the fusion cost.
+
+## Stable URL form
+
+The canonical coordinate string is:
+
+```
+<group>/<dashboard>/<tab>
+```
```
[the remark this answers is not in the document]
mchmarny: Fixed in 54c4447d — added `text` language identifier to the fenced block. Thanks\!

pkg/recipe/coordinate.go
```diff
@@ -0,0 +1,105 @@
...
+	// Tab is the intent, optionally suffixed with the platform when one
+	// is set (e.g. "training" or "training-kubeflow").
+	Tab string
+}
+
+// Path returns the coordinate as "<group>/<dashboard>/<tab>". The
+// host/navigation scheme built around this path is owned by the
+// consumer (GP5/TG4a); this package only emits the canonical segments.
+func (co Coordinate) Path() string {
+	return co.Group + "/" + co.Dashboard + "/" + co.Tab
```
njhensley, line 45: **[High] `Path()` is not round-trippable, but ADR-012 markets it as the parseable "stable URL form."**

`Dashboard = accelerator + "-" + os` joins two free-form fields with `-`, and `CriteriaAcceleratorRTXPro6000 = "rtx-pro-6000"` (criteria.go:121) already contains hyphens. So `eks/rtx-pro-6000-ubuntu/training` has no positional split point — a consumer cannot recover `(rtx-pro-6000, ubuntu)` vs `(rtx, pro-6000-ubuntu)`. ADR §"Stable URL form" says RQ1 deep-links to and TG4a's presence endpoint checks this string.

Not a blocker for *this* PR — nothing parses `Path()` yet; deep-links/presence checks only need it stable+equal, not decomposable. **Recommend:** track on the first parsing consumer (RQ1 #1283 / TG4a #1284) — ship an explicit `ParseCoordinate`, or keep consumers on the struct fields rather than string-splitting. Add an ADR note that `Path()` is a stable *opaque identity*, not a decomposable key, until then.

pkg/recipe/coordinate.go
```diff
@@ -0,0 +1,105 @@
...
+		Dashboard: accelerator + "-" + os,
+		Tab:       tab,
+	}, nil
+}
+
+// requireConcrete returns value when it is a concrete (non-empty,
+// non-wildcard) dimension, or an ErrCodeInvalidRequest naming dim when it
+// is empty or the "any" wildcard.
+func requireConcrete(dim, value string) (string, error) {
+	if value == "" || value == CriteriaAnyValue {
```
njhensley, line 106: **[High] No charset guard → a `/` in a value silently breaks the segment count.**

`requireConcrete` rejects only empty/`"any"`. Registry values admitted via `--data` are merely lowercased/trimmed (no charset validation), so a service value like `acme/ncp` yields `acme/ncp/h100-ubuntu/training` — a 4-segment path that any positional `/` split mis-parses. That is exactly the "ambiguous placement / silent fusion" the ADR claims to prevent.

This one is cheap and self-contained — **reasonable to fold into this PR**: reject values containing `/` (and arguably `-` in the joined fields), failing closed with `ErrCodeInvalidRequest` naming the dimension, plus a negative test.

**[Low, same spot]** `value == CriteriaAnyValue` compares to the literal `"any"` without trim/lowercase, so `"ANY"`/`" any "` from a hand-built (unnormalized) `Criteria` would pass as concrete. Only reachable if a caller bypasses the resolve path — either normalize here or document that `CoordinateFor` assumes resolved input.

pkg/recipe/coordinate_test.go
```diff
@@ -0,0 +1,170 @@
...
+				t.Errorf("String() = %q, want == Path() %q", got.String(), got.Path())
+			}
+		})
+	}
+}
+
+func TestCoordinateForRegistryUnionConcrete(t *testing.T) {
+	r := NewCriteriaRegistry()
+
+	services := concreteValues(r.AllServiceTypes())
```
njhensley, line 138: **[High] The registry-union test never actually exercises the registry-driven fail-closed path.**

`concreteValues()` strips `"any"`, but `NewCriteriaRegistry()` is empty and the static `Get*Types()` lists never contain `"any"` — so the filter removes a token that is never present, and no wildcard ever reaches `CoordinateFor` through this test. If `concreteValues` were removed or a `--data` registry ever surfaced `"any"`, this test would still pass.

Fail-closed *is* covered by the literal cases in `TestCoordinateFor` (nil + each dimension `any`), so this is dead-code/clarity rather than a true coverage hole. **Recommend:** either feed an explicit `"any"`/empty into the loop and assert it fails closed, or drop the dead filter with a comment pointing at the literal cases that own that coverage.

pkg/recipe/coordinate.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		return Coordinate{}, err
+	}
+	intent, err := requireConcrete("intent", string(c.Intent))
+	if err != nil {
+		return Coordinate{}, err
+	}
+
+	tab := intent
+	if platform := string(c.Platform); platform != "" && platform != CriteriaAnyValue {
```
njhensley, line 88: **[Medium] `tab = intent-platform` round-trips only by the unenforced accident that intents are hyphen-free.**

Splitting `training-kubeflow` back to `(training, kubeflow)` works only because today's intents (`training`, `inference`) contain no `-`. A future/`--data` intent like `fine-tuning` would make `fine-tuning-kubeflow` un-splittable. Same delimiter family as the `Path()` issue above — the structural-parsing fix covers it. **Recommend:** add an explicit ADR invariant: "dimension values must not contain the active delimiter."

docs/design/012-recipe-coordinate-mapping.md
```diff
@@ -0,0 +1,232 @@
...
+
+Notes:
+
+- `k8s_version` is **observed** (what the cluster actually reported);
+  `k8s_constraint` is **declared** (what the recipe asked for). Keeping both
+  distinct is what lets the board show satisfied-vs-violated without re-resolving
+  the recipe.
+- `signer_identity` + `signer_issuer` together key the **latest-per-signer**
+  default scope referenced in the k8s-in-column countermeasure.
+- The key strings and their count (7) are the contract: TG2 emits exactly these;
```
njhensley, line 189: **[Medium] The 7-key `started.json` schema is declared as a binding contract but is prose-only — nothing enforces it.**

"TG2 emits exactly these; TG3 reads exactly these" reads as a frozen ABI, yet no Go constant/struct/test pins the key names, order, or count. A typo or reorder on either side drifts silently — the same anti-drift failure the ADR's Problem section warns against, now in the half of the contract that isn't code. Deferring the emitter to TG2 (#1267) is fine, but **reword to**: "specified here, enforced by TG2," rather than implying it is pinned in code.

docs/design/012-recipe-coordinate-mapping.md
```diff
@@ -0,0 +1,232 @@
+# ADR-012: Recipe → Board-Coordinate Mapping
```
njhensley, line 1: **[Low] Missing `## Status` section.** 7 of 8 prior ADRs (002–005, 007–009) open with a `## Status`. Add `## Status: Accepted` (or `Proposed`) so consumers importing this contract know whether the mapping is ratified.

pkg/recipe/coordinate_test.go
```diff
@@ -0,0 +1,170 @@
...
+		name    string
+		c       *Criteria
+		want    string // Coordinate.Path()
+		wantErr bool
+	}{
+		{
+			name: "eks h100 ubuntu training kubeflow",
+			c: &Criteria{Service: CriteriaServiceEKS, Accelerator: CriteriaAcceleratorH100,
+				OS: CriteriaOSUbuntu, Intent: CriteriaIntentTraining, Platform: CriteriaPlatformKubeflow},
+			want: "eks/h100-ubuntu/training-kubeflow",
```
njhensley, line 50: **[Low] No format-lock golden test independent of the fixtures.** The canonical string is only asserted via the table's `want` fields; a refactor that changes the delimiter in both `Path()` and these `want` strings stays green while breaking the five downstream consumers that hardcode `<group>/<dashboard>/<tab>`. Consider one standalone `Path() == "eks/h100-ubuntu/training-kubeflow"` assertion explicitly framed as "this exact string must never change."

pkg/recipe/coordinate_test.go
```diff
@@ -0,0 +1,170 @@
...
+		{name: "intent any", c: &Criteria{Service: CriteriaServiceEKS, Accelerator: CriteriaAcceleratorH100, OS: CriteriaOSUbuntu, Intent: CriteriaIntentAny}, wantErr: true},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := CoordinateFor(tt.c)
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("err = %v, wantErr %v", err, tt.wantErr)
+			}
+			if tt.wantErr {
+				if !stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, "")) {
```
njhensley, line 86: **[Low] Negative cases assert the error *code* but not that the message names the offending dimension** — which both the code comment and ADR explicitly promise ("naming the offending dimension"). A refactor dropping the dim name from the message (e.g. a generic "dimension must be concrete") still passes. Add a substring assertion on the dimension name in one negative case.

docs/design/012-recipe-coordinate-mapping.md
```diff
@@ -0,0 +1,232 @@
...
+- The #1224 cross-link contract that recipe-health consumes.
+
+**Out of scope:**
+
+- The consumers themselves — TG2 (#1267), TG3, TG4a, RQ1 (#1283), GP4 (#1404).
+  They *import* this mapping; their internals are owned by their own issues.
+- The API and UI that render the board (GP5).
+- The live board / hosting / navigation host.
+- The user-facing `docs/user/` TestGrid page. It is **deferred**: it has a soft
+  dependency on TG4a's served URL form, so it is not authored here. **#1272 stays
```
njhensley: **[Nitpick] Durable ADR embeds transient issue-lifecycle state** ("**#1272 stays open until that user-facing page lands**"). Once #1272 closes, the ADR still asserts it's open. Keep the design record timeless and let the issue tracker own lifecycle — consistent with the repo's "don't reference transient state in durable artifacts" guidance.

njhensley (commented):
## Multi-persona review + senior-engineer meta-review

Reviewed by four independent personas (correctness, API-contract, docs, test), then adjudicated by a senior-engineer meta-review that calibrated severity to this PR's **additive, contract-publishing scope** (no consumers wired yet).

**TL;DR — no blockers.** Clean, pure, well-tested function. The personas converged on one real defect: the coordinate string is **provably not round-trippable** (`rtx-pro-6000` and `--data` values collide with the `-`/`/` delimiters). But since nothing parses `Path()` yet, that is a **High-priority design issue to resolve with the first parsing consumer (RQ1/TG4a)** — not a merge blocker. Deep-links and the presence endpoint only need the string *stable and equal*, not decomposable.

**Recommendation: merge-with-changes (minimal) + tracked follow-ups.**
- Fold into this PR (cheap, contract-strengthening): the `/` charset guard, the dead-filter test cleanup, the `## Status` line, and rewording the 7-key schema as "specified here, enforced by TG2."
- Block on the first consumer issue (cross-linked from ADR-012): the round-trippability fix / `ParseCoordinate` / delimiter invariant — so no string-splitting consumer can land before parseability is settled.

Inline comments below, tiered High → Nitpick. One persona finding was **dismissed**: "the `Coordinate` struct omits the `metadata.name` join key" — the ADR deliberately keeps `name` (identity) off `Coordinate` (placement) since the caller holds both; that's a correct design choice.

pkg/recipe/coordinate.go
mchmarny: Folded in (820eef35). \`requireConcrete\` now routes through a new \`rejectPathSeparator\`, so service/accelerator/os/intent **and** platform fail closed with \`ErrCodeInvalidRequest\` naming the dimension when a value contains \`/\`. Added negative cases (\`service contains slash\`, \`platform contains slash\`). On the [Low]: left as-is — the function documents its input as already-resolved \`Criteria\`, and normalizing \`"any"\`/case here would duplicate the resolve path; the resolved-input assumption is stated in the doc comment.

pkg/recipe/coordinate.go
mchmarny: Documented (820eef35). Added an ADR note under "Stable URL form": \`Path()\` is a stable **opaque identity**, not a decomposable key — \`rtx-pro-6000\` already carries the join char, so consumers compare/deep-link the string but address dimensions via the \`Coordinate\` struct fields, never by splitting. An explicit \`ParseCoordinate\` (with a "values must not contain the active delimiter" invariant) is deferred to the first parsing consumer (RQ1 #1283 / TG4a #1284).

pkg/recipe/coordinate.go
mchmarny: Covered by the same opaque-identity note (820eef35): the ADR now states values may contain \`-\` (\`rtx-pro-6000\`, a future \`fine-tuning\` intent), so the tab is not \`-\`-decomposable, and the delimiter invariant is deferred to \`ParseCoordinate\`. Only \`/\` is enforced now, since it is the one char that changes the segment count.

pkg/recipe/coordinate_test.go
mchmarny: Fixed (820eef35). Added an explicit assertion that a wildcard service fails closed through the registry-driven path, plus a comment noting \`concreteValues\` is defensive against a \`--data\`-seeded registry surfacing \`"any"\` and that the literal per-dimension cases in \`TestCoordinateFor\` own the rest of that coverage.

docs/design/012-recipe-coordinate-mapping.md
mchmarny: Reworded (820eef35): "specified here and **enforced by TG2 (#1267)**" — and added that no Go constant pins the names today, so TG2 must lock the key names/order/count in a shared constant + round-trip test when it lands, rather than relying on the prose.

docs/design/012-recipe-coordinate-mapping.md
mchmarny: Added \`## Status: Accepted\` (820eef35).

pkg/recipe/coordinate_test.go
mchmarny: Added \`TestCoordinatePathFormatLock\` (820eef35) — a standalone assertion that \`Path()\` equals \`eks/h100-ubuntu/training-kubeflow\`, framed as a frozen contract string that must fail even if the table want-strings change in lockstep.

pkg/recipe/coordinate_test.go
mchmarny: Added \`TestCoordinateForErrorNamesDimension\` (820eef35) — asserts the failure message contains \`accelerator\` for an \`any\`-accelerator input.

docs/design/012-recipe-coordinate-mapping.md
mchmarny: Reworded (820eef35) to drop the issue-lifecycle state — the deferred page is now described as "authored alongside that work rather than here," leaving #1272 lifecycle to the tracker.

njhensley:
## Re-review — feedback addressed ✅

Re-reviewed at `820eef35`. All 9 inline findings from the prior review are resolved; I ran `go test ./pkg/recipe/... -run Coordinate -v` against the updated code — **all pass**, including the new negative/contract tests.

| # | Finding | Resolution |
|---|---------|-----------|
| High 1 | `Path()` not round-trippable, marketed as parseable | ADR now has a dedicated **"`Path()` is a stable *opaque identity*, not a decomposable key"** section — spells out the `rtx-pro-6000` ambiguity, directs consumers to the struct fields (never string-split), and defers `ParseCoordinate` + the delimiter invariant to RQ1 #1283 / TG4a #1284 |
| High 2 | No `/` charset guard on `--data` values | Added `rejectPathSeparator`, applied to all four required dims **and** platform; fails closed naming the dimension |
| High 3 | Registry-union test never fed a wildcard | Added an explicit wildcard-service `CoordinateFor` assertion that must fail closed, with a comment pointing at the literal cases that own the rest |
| Med 4 | 7-key schema declared binding but prose-only | Reworded to **"specified here and enforced by TG2"**; notes no Go constant pins it today and that TG2 must lock names/order/count + round-trip test |
| Med 5 | `intent-platform` delimiter invariant unenforced | Folded into the opaque-identity section + the deferred "values must not contain the active delimiter" invariant |
| Low 6 | Missing `## Status` | Added `## Status: Accepted` |
| Low 8 | No format-lock golden test | Added `TestCoordinatePathFormatLock` (frozen-string assertion independent of fixtures) |
| Low 9 | Negative tests didn't assert dimension name | Added `TestCoordinateForErrorNamesDimension` |
| Nit 10 | ADR embedded issue-lifecycle state | "#1272 stays open until…" reworded to "authored alongside that work rather than here" |

**One optional item intentionally left** (Low, from the High-2 comment): `requireConcrete` still compares to the literal `"any"` without trim/lowercase. It's only reachable via a hand-built unnormalized `Criteria`, so defense-in-depth — fine to leave or document as "assumes resolved input."

**LGTM** — the cheap contract-strengthening fixes landed in this PR and the round-trippability work is correctly deferred and cross-linked to the first parsing consumer. No remaining blockers.

njhensley (approved):

PR #1414, opened 2026-06-23
feat(evidence): minimize sensitive content in evidence bundles

### LINKED ISSUE #1345: Minimize sensitive content in evidence bundles while preserving the verifiable signal
# Minimize sensitive content in evidence bundles while preserving the verifiable signal

## Summary

An evidence bundle today ships the full backing payloads it attests to — the
validate-time **snapshot** (`snapshot.yaml`), the resolved **recipe**
(`recipe.yaml`), and per-phase **CTRF reports** (`ctrf/<phase>.json`) including
captured pod stdout. As snapshot collectors grow (more node/OS/topology detail)
and CTRF accumulates more log lines, the bundle increasingly carries
operational detail that is sensitive to publish — node names, cloud provider
instance IDs, the full node label/taint set, kernel/sysctl tuning, and raw
container stdout that can leak internal endpoints, hostnames, and identifiers.

The goal of this issue is to reduce what the bundle *physically ships* to the
minimum needed for a strong, verifiable conformance signal — without weakening
the cryptographic binding that makes the bundle trustworthy.

## Key observation: the verifiable signal is already digest-based

The trust signal does **not** come from the raw payloads — it comes from the
signed in-toto predicate, which commits to every artifact by hash
(`pkg/evidence/attestation/types.go`):

- `Predicate.Recipe.Digest` = `sha256(canonicalize(recipe.yaml))`
- `Predicate.Phases[phase].CTRFDigest` = `sha256(ctrf/<phase>.json)`
- `Predicate.BOM.Digest`, `Predicate.BOM.ImageCount`
- `Predicate.Manifest.Digest` → `manifest.json`, which carries a per-file
  `sha256` for every file in the bundle, transitively binding all of them to
  the signature
- Per-phase `Passed`/`Failed`/`Skipped` counts, `Fingerprint`, `CriteriaMatch`,
  validator image digests, catalog version

So a verifier can confirm "recipe X, validated against a cluster matching
criteria Y, passed/failed N tests per phase, signed by identity Z, logged in
Rekor" **entirely from the predicate** — the bundled `snapshot.yaml` and CTRF
stdout are the *backing content* a digest points at, not the signal itself.
That gives us room to shrink or sanitize the backing content while keeping the
binding intact.

This is consistent with the direction already taken in ADR-007
(`docs/design/007-recipe-evidence.md`), which split large/sensitive **logs**
into a separate opt-in artifact precisely because they are "large,
contributor-controlled ... and frequently sensitive (cluster identifiers,
internal endpoints)." This issue extends that same reasoning to the snapshot
and CTRF payloads that still ship in the summary bundle by default.

## What's currently bundled and where the sensitivity lives

**Snapshot (`snapshot.yaml`)** — emitted in full. Sensitive fields include:
- `pkg/collector/k8s/node.go`: `source-node` (node name) and `provider-id`
  (e.g. `aws:///us-west-2a/i-0123456789abcdef0`)
- `pkg/collector/topology`: the **complete** node label and taint set across the
  cluster — custom labels routinely encode account IDs, cost centers,
  environment/tenant names, regions
- `pkg/collector/os`: GRUB/kernel boot args, full sysctl tree, loaded modules —
  reveals hardening/tuning posture
- `pkg/collector/gpu`, `systemd`: driver/CUDA versions, service config

**CTRF reports (`ctrf/<phase>.json`)** — `pkg/validator/ctrf/types.go`:
- `TestResult.Stdout` (captured pod stdout, capped by line count/length) and
  `TestResult.Message` (termination-log text) — free-form log text that can
  contain IPs, DNS names, cert/secret names, internal URLs

**Recipe (`recipe.yaml`)** — generally the least sensitive (criteria, component
refs, version pins, constraints), but worth confirming nothing operator-specific
leaks via overlay resolution / `matchedRules`.

## Goal

Define and implement a "minimal evidence" bundle that:
1. Preserves the full cryptographic verification story (predicate digests,
   manifest binding, DSSE signature, Rekor).
2. Ships only the data an external auditor needs to trust the conformance claim.
3. Keeps potentially sensitive operational detail out of the published bundle by
   default, with an explicit opt-in for operators who *want* to publish more.

## Possible approaches (to be discussed — not yet decided)

These are mutually compatible building blocks, not exclusive options:

- **A. Field-level snapshot redaction/allowlist.** Emit a sanitized snapshot
  that drops/obfuscates known-sensitive fields (node names, provider IDs, the
  raw label/taint set) while retaining what constraints actually evaluated
  against. Risk: must keep the snapshot digest meaningful — decide whether the
  predicate attests to the sanitized snapshot or to the raw one (and keep the
  raw one local). An allowlist (fail-closed) is safer than a denylist as
  collectors grow.
- **B. CTRF stdout/message scrubbing or omission by default.** Strip or redact
  `Stdout`/`Message` from published CTRF, or replace full logs with a digest +
  summary, keeping raw logs in the existing opt-in local/logs path. The
  pass/fail counts and per-test status — the actual signal — stay.
- **C. Digest-only backing content.** Replace bundled payloads with their
  digests in the predicate and ship the raw payloads only when the operator
  opts in (mirrors the ADR-007 logs split). Verifiers still get the binding;
  publication of the bytes becomes a choice.
- **D. Built-in redaction transforms with a documented allowlist.** A small,
  audited set of regex/structured redactors (cf. the existing ELB-hostname
  redaction in `pkg/evidence/cncf/scripts/collect-evidence.sh:2263`) applied
  consistently, with the redaction policy itself recorded in the predicate so
  verifiers know what was scrubbed.

## Open questions

- Should the predicate attest to the **sanitized** artifact (so the published
  bundle self-verifies) or to the **raw** artifact (so the local full bundle
  remains the canonical signed thing)? This is the central design decision.
- Do downstream consumers (CNCF conformance submission, recipe-health) depend on
  any field we'd want to redact? Inventory those consumers first.
- Default posture: minimal-by-default with `--include-snapshot`/`--include-logs`
  opt-in, vs. full-by-default with `--minimal`/`--redact`?
- How do we make redaction auditable — record the redaction policy version in
  the predicate so a verifier can tell *what* was removed?

## Acceptance criteria (draft)

- [ ] A documented definition of "minimal evidence bundle" contents.
- [ ] Sensitive snapshot fields and CTRF log text are excluded or redacted from
      the published bundle by default (or behind an explicit minimal mode).
- [ ] `aicr evidence verify` still passes on the minimal bundle: predicate
      digests, manifest binding, and signature all validate.
- [ ] Redaction policy is deterministic (same input → same bytes) and recorded
      so verification stays reproducible.
- [ ] Docs updated (`docs/design/007-recipe-evidence.md`, CLI/evidence docs) and
      ADR amended for the new default posture.

## References

- `pkg/evidence/attestation/types.go` — `Predicate`, `*Ref` digest fields,
  bundle file constants
- `pkg/evidence/attestation/manifest.go` — per-file sha256 integrity inventory
- `pkg/collector/{k8s,os,gpu,systemd,topology}` — snapshot data sources
- `pkg/validator/ctrf/types.go`, `pkg/validator/job/result.go` — CTRF + stdout
  capture
- `docs/design/007-recipe-evidence.md` — summary/logs split rationale (logs
  already treated as sensitive/opt-in)

njhensley:
## Summary

Evidence bundles are now **minimized by default**: `snapshot.yaml` is reduced to a fail-closed allowlist and per-test CTRF `stdout`/`message` are omitted, keeping sensitive operational detail out of the published artifact. A new `--full` flag restores the raw payloads.

## Motivation / Context

An evidence bundle shipped the full validate-time snapshot and per-phase CTRF stdout, publishing detail that is sensitive to expose: node names, cloud provider instance IDs, the full node label/taint set, kernel/sysctl tuning, loaded modules, systemd service config, and raw container logs.

The verifiable trust signal is digest-based — the signed in-toto predicate commits to artifacts by hash and carries the derived fingerprint, criteria-match, and per-phase counts. The snapshot and CTRF logs are the *backing content* those digests point at, not the signal itself, so they can be minimized without weakening the cryptographic binding. This extends the same reasoning ADR-007 already applied to logs.

Fixes: #1345
Related: N/A

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `pkg/evidence/{redact,attestation,verifier}`

## Implementation Notes

- **New package `pkg/evidence/redact`** — pure, deterministic, non-mutating `Snapshot()` and `CTRF()`. The snapshot allowlist is fail-closed: only enumerated measurement subtypes/keys survive, so a field a future collector adds is dropped until explicitly allowlisted. v1 keeps `K8s.server`, an allowlisted `K8s.node` subset (drops `source-node`, `provider-id`, `container-runtime-id`), `GPU.hardware`, `OS.release`, `NodeTopology.summary`; drops `OS.{grub,sysctl,kmod}`, all of `SystemD`, `NodeTopology.{label,taint}`, subtype `Context`, and header `metadata.source-node`.
- **Fingerprint is computed from the raw snapshot, not the shipped one.** `fingerprint.FromMeasurements` reads fields the allowlist drops (e.g. `NodeTopology.label`), so deriving the predicate's fingerprint/criteriaMatch from the full snapshot keeps the conformance signal identical whether or not the shipped bytes were minimized. The verifier trusts the fingerprint via the signature; it never recomputes it from `snapshot.yaml`.
- **Self-verifying.** The predicate digests (recipe, CTRF, BOM, manifest) cover the bytes that actually ship, so a minimal bundle passes `aicr evidence verify`. The applied policy is recorded in a new `predicate.redaction` block (`omitempty`), so a `--full` predicate is byte-identical to prior output.
- **Default behavior change:** `aicr validate --emit-attestation` now writes a redacted bundle. No documented downstream consumer (recipe-health, CNCF submission) reads the bundle's `snapshot.yaml` back. `--full` is the opt-out.
- **Not a schema bump:** `redaction` is an additive `omitempty` predicate field, consistent with ADR-007's additive-field policy; `PredicateSchemaVersion` stays `1.0.0`.

## Testing

```bash
go test ./pkg/evidence/... ./pkg/cli/...          # pass
golangci-lint run -c .golangci.yaml ./...          # 0 issues (pinned v2.12.2)
```

- `pkg/evidence/redact` coverage 98.4%; new round-trip test confirms a minimal bundle self-verifies (`TestVerify_MinimalBundleSelfVerifies`); `--full` byte-identity and fingerprint-from-raw decoupling are guarded by tests.
- Not runnable in my local sandbox (no C compiler for `-race`; `sigstore.dev` TUF/Fulcio network blocked) — the `-race` run and the `pkg/bundler/attestation`/`pkg/trust` sigstore tests rely on CI. Those packages do not import the changed code.

## Risk Assessment

- [x] **Medium** — Changes the default contents of the published evidence bundle

**Rollout notes:** Default `--emit-attestation` output is now redacted; operators who want the prior full payloads pass `--full`. Verification is unaffected (minimal bundles self-verify; `--full` predicates are byte-identical to prior output). No migration required.

## Checklist

- [ ] Tests pass locally (`make test` with `-race`) — non-race suite passes locally; `-race` requires CI (no local C compiler)
- [x] Linter passes (`make lint`) — pinned `golangci-lint v2.12.2`, 0 issues
- [x] I updated docs if user-facing behavior changed (ADR-007, cli-reference, validation, artifact-verification)


pkg/evidence/redact/redact.go
```diff
@@ -0,0 +1,253 @@
...
+//
+// The subtype names and data keys mirror the literals the collectors author
+// (pkg/collector/{k8s,os,gpu,topology}) — the same names pkg/fingerprint keys
+// off. They are not shared constants, so a collector that renames a subtype or
+// key must update this table too; otherwise the renamed field is silently
+// dropped from the minimal snapshot (fail-closed, but lossy). Bump PolicyVersion
+// when the allowlist changes.
+var snapshotAllowlist = map[measurement.Type]map[string]subtypePolicy{
+	measurement.TypeK8s: {
+		"server": keep(), // version etc. — not sensitive
```
mchmarny, line 96: Medium, non-blocking — fail-closed claim is only partial. The package doc (lines 27-31) says "anything a future collector adds is dropped until explicitly allowlisted," but that holds only at **type** and **subtype** granularity, plus **key** granularity for key-constrained subtypes like `node`. The `keep()`-all subtypes here (`server`, `hardware`, `release`, `summary`) are **fail-open at the key level**: a sensitive key a future collector attaches to one of these ships by default.

Nothing leaks today — verified these four currently carry only version/model/count/distro-identity keys. But the asymmetry with `node` (which you correctly key-constrained *because* it has sensitive keys) is exactly the trap. Two options: key-constrain these four as well, or tighten the doc and add an invariant note that `server`/`hardware`/`release`/`summary` must never accrue node-identifying keys (and bump `PolicyVersion` if they do). Either way the stated guarantee and the implementation should match.

mchmarny (approved):
LGTM. The core insight — that the trust signal is digest-based, so the snapshot/CTRF backing content can be minimized without weakening the binding, and the fingerprint must therefore be derived from the *raw* snapshot rather than the shipped bytes — is exactly right, and the implementation matches it cleanly. `--full` byte-identity, the `omitempty` `redaction` block, and self-verification of minimal bundles are all handled well.

One medium, non-blocking note inline: the "fail-closed" guarantee is only partial — the unconstrained `keep()`-all subtypes are fail-open at the key level, which the package doc overclaims. Worth reconciling doc and implementation, but nothing leaks today and it doesn't gate this merge.

Nice work — the ADR-007 extension and the docs are thorough.

Issue #828, opened 2026-05-09
Integrate k8s-launch-kit as a snapshot Collector


mchmarny:
Part of #827.

## Goal

Persist NIC/fabric/rail topology in AICR's snapshot by consuming [k8s-launch-kit](https://github.com/NVIDIA/k8s-launch-kit) (l8k) as a **Go library** that implements AICR's `Collector` interface (`pkg/collector/types.go:27`).

## Background

l8k already encodes the discovery logic AICR needs (kernel module graph from `/sys/module`, GPU topology via `nvidia-smi`, NIC inventory via `NicDevice` CRs, machine-type detection, node-label patching). Today that logic is invoked by the `l8k` CLI via a *mutating* flow:

1. Patch/create `NicClusterPolicy` so the network-operator stands up `nic-configuration-daemon`.
2. Exec into those daemon pods to collect kernel/hardware data.
3. Read `NicDevice` CRs.
4. Patch node labels.
5. Defer-cleanup the patch.

See `pkg/networkoperatorplugin/discovery.go:99-260`. **This violates AICR's collector contract** (read-only, bounded, no side effects). Phase 1 must reconcile that.

### Option A — Static collector via `presets` + `releases` catalog (no cluster contact)

Wrap `pkg/presets.LoadPreset(machineType, gpuType)` and `pkg/networkoperatorplugin.LookupRelease`. Caller (or AICR's existing GPU collector) supplies machine type + GPU product; the new collector returns l8k's `LaunchKubernetesConfig` populated from baked-in presets.

- **Pros:** ships immediately. Zero footprint. No cluster mutation. No new RBAC.
- **Cons:** no live data — degenerate \"collector\" that just looks up a table. Doesn't validate actual hardware. Useful as a fallback / offline mode, not as the primary path.

### Option B — Wrap `DiscoverClusterConfig` as-is

AICR collector calls l8k's existing mutating `DiscoverClusterConfig` directly.

- **Pros:** full topology truth — fabric, rails, NIC inventory, NUMA, GPU↔NIC proximity.
- **Cons:**
  - **Violates `Collector` contract:** mutates `NicClusterPolicy`, requires pod-exec RBAC.
  - **Requires network-operator already installed** — chicken-and-egg if AICR is meant to *produce* that config.
  - **Race conditions** on parallel snapshots against the same cluster (both will fight over the patch).
  - Hard-coded namespace fallback (`discovery.go:201`) may conflict with multi-tenant deployments.

[...]

PR #1417, opened 2026-06-23
feat(measurement): NetworkTopology type + Subtype.Items for cross-repo l8k integration

almaslennikov:
…
# Why

aicr models cluster state as a Snapshot of `Measurement`s (K8s, GPU, OS, SystemD, NodeTopology). NodeTopology is cluster-wide and only carries node taints + labels. The model has no slot for per-hardware-group network topology -- PFs and their PCI addresses, rails, NUMA, RDMA capability, link type, kernel module lists, machine/GPU identity -- which makes it impossible to recipe or validate networking beyond a single all-or-nothing network-operator toggle.

k8s-launch-kit (l8k) already discovers exactly this data via its `cluster-config.yaml`. The intended integration is library-mode: l8k imports `pkg/measurement` and emits `*Measurement` values directly that aicr Snapshots can consume. The dependency direction is l8k -> aicr; the Measurement schema is aicr's contract.

This commit is Phase 1 -- the schema affordances l8k needs to implement that contract. No l8k import is introduced here; the snapshotter and CLI are untouched. The producer side (Phase 2, in the l8k repo) and the consumer side (Phase 3, the aicr collector + CLI flag) land separately.

Part of https://github.com/NVIDIA/aicr/issues/828

# What changes

Schema additions to `pkg/measurement`:

  - New `TypeNetworkTopology` enum value. Today at most one Measurement of this Type per snapshot, keeping all find-first-by-Type consumers (constraints extractor, recipe validation, diff indexing, fingerprint) correct. Multi-instance support is a deliberate future step.

  - New `Subtype.Items []ItemEntry` field. `ItemEntry` mirrors Subtype's scalar contract: `Data map[string]Reading` for scalar measurements plus `Context map[string]string` for descriptive metadata. This lets homogeneous record lists (per-PF entries under a `pfs` subtype) be represented as real YAML arrays rather than packed-string-encoded scalars or numbered subtypes (`pf-0`, `pf-1`, ...). `Reading` itself is NOT extended -- it stays scalar-only.

  - Custom `UnmarshalJSON`/`UnmarshalYAML` on `ItemEntry` so `Reading` values inside `ItemEntry.Data` round-trip the same way they do in `Subtype.Data` today.

  - `Subtype.Validate` relaxed: a subtype is valid with either non-empty `Data` or non-empty `Items` (or both). The existing `len(Data) > 0` invariant predated Items.

  - `Items` carries `omitempty` so subtypes that don't use it serialize byte-identically to today's output. `Data`'s tag also gains `omitempty` so Items-only subtypes don't emit a noisy `data: {}`; this is behavior-safe because every existing subtype satisfies `len(Data) > 0` (Validate guaranteed it).

  - New SubtypeBuilder helpers: `WithContext(k, v)`, `WithContextMap(map)`, `WithItem(entry)`, `WithItems(slice)`.

Constraint path grammar extension in `pkg/constraints`:

  - New optional item-selector segment: `{Type}.{Subtype}[<selector>].{Key}`. Two selector forms:
      Index:     `NetworkTopology.pfs[0].rail`
      Predicate: `NetworkTopology.pfs[rail=3].pciAddress`
    Predicates evaluate against `ItemEntry.Data` first, then
    `ItemEntry.Context`. First match wins; multiple matches return
    `ErrCodeConflict`. Key resolution inside the matched entry tries
    `Data` then `Context`. Paths without a selector continue to use the
    legacy `Subtype.Data` lookup unchanged.

OpenAPI (`api/aicr/v1/server.yaml`):

  - `Measurement.type` enum corrected from the stale `[systemd, os, k8s, gpu]` to match `measurement.Types`: `[K8s, GPU, OS, SystemD, NodeTopology, NetworkTopology]`. The pre-existing miss (`NodeTopology` and `SystemD` were never added) is bundled into this fix.
  - New `ItemEntry` schema mirroring `data` + `context`.
  - `Subtype` gains an optional `items` array property; `data` is no longer required at the schema level (Validate still enforces "at least one of data/items").

Docs:

  - New `docs/integrator/measurement-api.md` publishing the cross-repo Measurement schema contract: Type cardinality table, Subtype layout rules, the NetworkTopology shape (subtype names, what lives in Context vs Data, item structure for `pfs`), and the constraint-path grammar including selectors. l8k's Phase 2 implementation targets this spec.
  - `pkg/measurement/doc.go` updated to flag the package as part of aicr's cross-repo public API surface.
  - `docs/integrator/index.md` adds the new page to the integrator TOC.

# Testing

  - `pkg/measurement/types_test.go`: Items-only subtype validation, JSON + YAML round-trip of Items with Reading values inside ItemEntry, byte-identity check that pre-existing Items-less subtypes don't suddenly emit an `items:` key.
  - `pkg/measurement/builder_test.go`: 100% coverage on each new builder helper (WithContext / WithContextMap / WithItem / WithItems).
  - `pkg/constraints/extractor_test.go`: parser tests for both selector forms and every error path; ExtractValue tests for index/predicate hits, out-of-bounds, no-match, no-items, ambiguous-predicate (ErrCodeConflict).

Coverage on touched packages vs main:

  pkg/measurement: 96.0% -> 96.6% (+0.6%)
  pkg/constraints: 80.9% -> 87.6% (+6.7%)

`go test ./...` passes (the pre-existing `pkg/trust` sandbox failure is unrelated -- it writes to `~/.sigstore` which is outside this environment's write allowlist; passes outside the sandbox). `go vet ./...` clean. `golangci-lint` not available in this environment per project notes -- run `make qualify` before pushing.

# Not in this commit

Phase 2 (in the l8k repo): embed l8k's discovery-only runtime artifacts (presets, default config; profiles deferred), rename `LaunchKubernetesConfig` -> `LaunchKitConfig`, add a library-friendly `Discover(ctx, kubeClient)` entry point, vendor aicr's `pkg/measurement`, and add `DiscoverMeasurements` / `ParseClusterConfigMeasurements` wrappers that emit `*aicrmeasurement.Measurement` values per the contract this commit publishes.

Phase 3 (in aicr): new `pkg/collector/network/` calling l8k's library, `--cluster-config` / `--discover-network` flags on `aicr snapshot`, RBAC additions for the agent Job, single-group enforcement.

Multi-group `NetworkTopology` is a future iteration; the consumer rewrites it requires (constraint @<id> grammar, recipe-validator iteration, diff (Type, identifier) keying, fingerprint group-aware logic) are deliberately deferred until the single-group integration has shaken out.

## Motivation / Context

Fixes: 
Related: 

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Testing

```bash
make qualify
```

## Risk Assessment

**Rollout notes:** 

## Checklist

### DIFF
--- a/api/aicr/v1/server.yaml
+++ b/api/aicr/v1/server.yaml
@@ -1301,12 +1301,35 @@ components:
           description: Optional unit of measurement
           example: "MB"
 
+    ItemEntry:
+      type: object
+      description: |
+        One ordered element of a Subtype.items list. Mirrors Subtype's
+        scalar-value contract: data holds Reading scalars; context holds
+        string-typed descriptive metadata. Used when a subtype carries a
+        homogeneous array of structured records (e.g. one entry per PF).
+      properties:
+        data:
+          type: object
+          additionalProperties:
+            $ref: "#/components/schemas/Reading"
+          description: Key-value pairs of scalar measurements for this item.
+        context:
+          type: object
+          additionalProperties:
+            type: string
+          description: Optional descriptive metadata for this item (string-typed).
+
     Subtype:
       type: object
-      description: A subtype within a measurement containing specific configuration data
-      required: [name, data]
+      description: |
+        A subtype within a measurement. A subtype must carry at least one
+        entry in `data` or `items`. `items` is used when the subtype payload
+        is naturally a list of structured records (e.g. per-PF entries under
+        a NetworkTopology `pfs` subtype).
+      required: [subtype]
       properties:
-        name:
+        subtype:
           type: string
           description: Subtype identifier (e.g., service name, configuration category)
           example: containerd.service
@@ -1320,17 +1343,28 @@ components:
           additionalProperties: true
           description: Optional metadata about the source and collection method
           nullable: true
+        items:
+          type: array
+          items:
+            $ref: "#/components/schemas/ItemEntry"
+          description: |
+            Optional ordered list of structured records. Independent of `data` —
+            a subtype may carry either or both.
 
     Measurement:
       type: object
-      description: A measurement grouping related configuration subtypes
+      description: |
+        A measurement grouping related configuration subtypes. Most types
+        appear at most once per snapshot; `NetworkTopology` is currently
+        emitted at most once but is the planned multi-instance exception
+        (one Measurement per discovered hardware group).
       required: [type, subtypes]
       properties:
         type:
           type: string
           description: Measurement type category
-          enum: [systemd, os, k8s, gpu]
-          example: systemd
+          enum: [K8s, GPU, OS, SystemD, NodeTopology, NetworkTopology]
+          example: SystemD
         subtypes:
           type: array
           items:
--- a/pkg/measurement/builder_test.go
+++ b/pkg/measurement/builder_test.go
@@ -15,6 +15,7 @@
 package measurement
 
 import (
+	"strconv"
 	"testing"
 )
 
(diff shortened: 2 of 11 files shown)


api/aicr/v1/server.yaml
```diff
@@ -1301,12 +1301,35 @@ components:
...
     Subtype:
       type: object
-      description: A subtype within a measurement containing specific configuration data
-      required: [name, data]
+      description: |
+        A subtype within a measurement. A subtype must carry at least one
+        entry in `data` or `items`. `items` is used when the subtype payload
+        is naturally a list of structured records (e.g. per-PF entries under
+        a NetworkTopology `pfs` subtype).
+      required: [subtype]
```
mchmarny, line 1330: nit: `required: [subtype]` here, but the Go field is `Name string \`json:"subtype,omitempty"\``. With `omitempty`, an empty-named subtype serializes without the field and would violate the schema that marks it required. Pre-existing tension (and `Validate` doesn't enforce a non-empty name either), just flagging since you're correcting this block anyway — either drop `subtype` from `required` or have `Validate` reject an empty name.

pkg/measurement/builder_test.go
```diff
@@ -135,6 +135,141 @@ func TestSubtypeBuilder(t *testing.T) {
...
+		t.Fatalf("Items len = %d, want 3", len(st.Items))
+	}
+	for i := range items {
+		if got := st.Items[i].Data["rail"].String(); got != fmtIntForTest(i) {
+			t.Errorf("Items[%d].Data[rail] = %q, want %d", i, got, i)
+		}
+	}
+}
+
+func fmtIntForTest(i int) string {
```
mchmarny: nit: `strconv.Itoa(i)` is clearer than hand-rolling the switch — importing `strconv` in a test is fine.

mchmarny (commented):
Clean, well-scoped Phase 1 — schema affordances only, no collector/CLI wiring, exactly the shape described. The `Items` model, the constraint-path selector grammar, and the contract doc are all coherent and the test coverage is thorough (parser forms + every error path, index/predicate extraction, JSON+YAML round-trip, and a byte-identity check that `omitempty` keeps existing snapshots unchanged).

A couple of correctness points I checked and am satisfied with: predicate string-matching (`Reading.String() == value`) is robust across YAML and JSON sources since `%v` on `float64(3.0)` (how JSON delivers integers) prints `3`, so `pfs[rail=3]` resolves regardless of source format; and ambiguous predicate → `ErrCodeConflict` / negative index / unclosed bracket all fail closed, which is the right call.

Only two minor nits inline, neither blocking. Content LGTM.

Note: this isn't approvable as-is on CI — **Lint** and the **gate** check (unsigned commit) are both red. Worth a `make qualify` pass and a `git commit -S --amend` (+ squash to one signed commit) before merge; both are mechanics, not content.

mchmarny (approved):

Issue #1263, opened 2026-06-09
Epic: TG — Deploy & operate the AICR TestGrid instance


njhensley:
### Summary

Stand up a new, public, AICR-native TestGrid in the `eidosx` UAT GCP project. A
`tools/` CI publisher converts AICR
evidence bundles (CTRF → deterministic jUnit) into a TestGrid-wire-compatible
GCS feed laid out CSP/service-first (group=`service`, dashboard=`accelerator`-`os`,
tab=`intent`[-`platform`], k8s-in-column). Upstream Kubernetes TestGrid
updater/tabulator/summarizer workers roll it up; an AICR-native config-gen, API,
and public UI replace the upstream config-gen, API, aggregator, and web UI. Both first-party UAT
runs and accepted external-contributor ADR-007 bundles are ingested by an
AICR-owned CI job — contributors never write the bucket. The write identity reuses
the existing GitHub-Actions→GCP WIF **federation** on a **dedicated, write-only**
publish service account (not the shared high-privilege UAT SA). Launch covers AWS
H100 (`eks`/`h100-ubuntu`) and GCP H100 (`gke`/`h100-cos`) — two dashboards,
because the clouds run different OS families.

### Children
- [ ] #1266 — TG1: Infra foundation (buckets, hardening, dedicated publish SA, GKE host cluster, serving posture)
- [ ] #1267 — TG2: Publish path (`tools/testgrid-publish`: OCI bundle → CTRF → jUnit → GCS feed)
- [ ] #1268 — TG3: Processing (upstream workers + AICR-native config-gen + metadata-key contract)
- [ ] #1269 — TG4a: AICR-native read-only API (greenfield)
- [ ] #1270 — TG4b: Public UI (greenfield)
- [ ] #1271 — TG5: Evidence ingestion CI (UAT + accepted external bundles)
- [ ] #1272 — TG6: IA/taxonomy spec + shared mapping function + public docs + #1224 reconciliation/cross-link contract
- [ ] #1273 — TG7: Operability (worker/API/host-cluster liveness + runbook + failure surfacing)

### Epic acceptance criteria

1. A nightly UAT run produces an evidence bundle that is ingested and published to
   the AICR TestGrid GCS feed at the correct CSP-first coordinate, and the build
   appears as a column in the public UI within one worker cycle — AWS H100 lands
   under `eks` / `h100-ubuntu`; GCP H100 lands under `gke` / `h100-cos`.
[...]

PR #1447, opened 2026-06-24
feat(testgrid): add testgrid-publish CLI tool (TG2)

### LINKED ISSUE #1267: TG2 — Publish path (`tools/testgrid-publish`: OCI bundle → CTRF → jUnit → GCS feed)
**Parent:** #1263 · **Labels:** `enhancement` · **Size:** L · **Depends on:** TG1
*(no auto area label today — `tools/**` matches no glob in `.github/labeler.yml`;
this child MUST add a `tools/**` (or `tools/testgrid-publish/**`) glob mapping to
a new/existing area, e.g. `area/ci` or a new `area/tools`.)*

**Goal.** A deterministic, least-privilege CI Go tool that consumes an AICR
evidence OCI bundle, converts its per-phase CTRF to canonical jUnit XML, assembles
`started.json`/`finished.json`, and writes the 4-level TestGrid-wire-compatible
GCS feed at the CSP-first coordinate with the pinned column-metadata key schema.

**Scope.**
- New `tools/testgrid-publish/` Go tool (`GOFLAGS=-mod=vendor go run
  ./tools/testgrid-publish`), modeled on the `tools/bom` Go-tool precedent (flags,
  `-deterministic` posture, README).
- **Input = OCI evidence bundle.** Pull via ORAS (the same transport ADR-007's
  `--push`/verifier paths use), materialize the `summary-bundle/`, read
  `recipe.yaml` (for criteria → coordinate) and `ctrf/{deployment,performance,
  conformance}.json`. The publisher's job: OCI bundle → CTRF → jUnit →
  started/finished → GCS.
- **Coordinate via the shared mapping function.** Derive `(group, dashboard, tab)`
  by calling the **shared `pkg/recipe` coordinate helper** (the single source of
  truth; TG6 owns the spec), taking the recipe's resolved `Criteria` as input —
  **never** by parsing `metadata.name`. The same helper is imported by TG3/TG4a/RQ1.
- **Deterministic CTRF→jUnit converter.** Map CTRF `TestResult`
  (`pkg/validator/ctrf/types.go`: `Name`, `Status` ∈ {`passed`,`failed`,`skipped`,
  `pending`,`other`}, `Duration` int ms, `Suite`, `Message`, `Stdout`) to jUnit:
  `failed`→`<failure>`, `other`→`<error>`, `skipped`/`pending`→`<skipped/>`,
  `passed`→bare `<testcase>`; `time` = `strconv.FormatFloat(ms/1000, 'f', 3, 64)`;
  suite totals = sum-of-rounded; preserve CTRF order; omit timestamp/hostname;
  fixed XML declaration + indentation + `encoding/xml` escaping; rune-safe
  line-boundary `<system-out>` cap with a fixed sentinel. **Byte-deterministic**
  per the `pkg/serializer.MarshalYAMLDeterministic` / `tools/bom -deterministic`
  precedent. Rows are named `<phase>/<check>`. **Multi-CTRF merge:** all
  `ctrf/*.json` (deployment/performance/conformance, plus any phases the dynamic-clusters epic DC4
  adds) merge into one deterministic `junit.xml` with phase-namespaced row names so
  the per-build column never collides duplicate test-case names into one flaky row.
- **Tab sanitize — empty platform yields bare intent.** When `platform` is absent,
  the tab is the bare `intent` (`training`, `inference`) with **no trailing
  hyphen** and **no `unknown` substitution** — the generic `empty→unknown` rule
  applies only to a *required* segment that resolves empty, never to an
  intentionally-omitted optional sub-segment. Path-construction tests include a
  no-platform recipe so TG2 and RQ1 are provably consistent.
- **build-id (load-bearing for idempotency AND time-series).** Define `{build-id}`
  so that (a) distinct first-party UAT runs of the same recipe are **distinct,
  chronologically-ordered columns** and (b) re-ingesting a bundle for the *same
  run* is idempotent. The idempotency key is `{recipe-coordinate + run identity}`,
  **not** raw bundle bytes (a re-signed/re-pushed bundle of the same run gets a new
  digest because the predicate's `AttestedAt` is a `Truncate(time.Second)`
  wall-clock, distinct from `BundleHash`). Confirm against the upstream updater
  whether columns sort by `started.json` timestamp or by build-id string; if the
  latter, prefix the build-id with a zero-padded/ISO timestamp (`timestamp +
  short-digest`) so it is both **monotonic** and idempotent. For community
  re-ingestion of an identical bundle, the digest makes it idempotent. A column
  ordering golden/integration check lives in TG3's seeded-feed smoke.
- **started.json / finished.json — timestamps from the bundle, not the clock.**
  `started.json = {timestamp, metadata}`; `finished.json = {timestamp,
  passed(bool), result: SUCCESS|FAILURE, metadata}`. Both timestamps are sourced
  from the **bundle predicate** (`AttestedAt` / phase timestamps), **never**
  `time.Now()` at publish — otherwise determinism and same-run idempotency break.
  `passed` is **binary**, derived from the CTRF summary (any phase with
  `Summary.Failed > 0` or `Summary.Other > 0` → FAILURE); nuance stays in the
  jUnit. Upload order **started → junit → finished** so a partial failure never
  shows a result without evidence.
- **GCS path = the contract.** Write
  `gs://<bucket>/groups/{group}/{dashboard}/{tab}/{build-id}/` with `started.json`,
  `artifacts/junit.xml`, `finished.json`. group=`service`, dashboard=`accelerator`-`os`,
  tab=`intent`[-`platform`], all from the bundle's recipe criteria via the shared
  helper. **4-level depth is load-bearing** — never emit a 5th level. Sanitize every
  required segment (lowercase; collapse `/\:;[]{}, `+whitespace to `-`; dedupe;
  trim; empty→`unknown`; k8s truncated to major.minor for column metadata).
- **Column metadata under a pinned key schema.** Emit the started.json metadata
  with the **exact key strings, count, and order** defined once in TG6's taxonomy
  spec and referenced verbatim by TG3's config-gen: the **aicr version**
  (`aicr_version`, the released/`main`-tip AICR version that produced the run); the
  **observed runtime k8s version** (`k8s_version`, truncated to `major.minor`, from
  the bundle predicate / snapshot); the **recipe-declared k8s constraint**
  (`k8s_constraint`,
  the `K8s.server.version` range read from the recipe — e.g. `">= 1.32.4"`,
  `recipes/overlays/gb200-eks-training.yaml:32`; cf. `">= 1.30"` at
  `aks-training.yaml:31` / `gke-cos-inference.yaml:35` — read **for display only**,
  never as a path segment); **verified signer identity + issuer** (the
  verified SAN from `aicr evidence verify`, not the raw bundle-asserted string);
  **source** (`uat` iff the verified SAN matches the UAT-workflow identity regexp,
  else `community` — derived, not a TG5 free flag); and the **evidence bundle
  digest**. These map onto the column-header `extra` array the upstream workers and
  API understand (`Header.extra`; e.g. the upstream convention uses
  `[release_stream, release_tag, runtime_release, operations_release]`,
  `test-result-ingestion-reference.md` §3, §6).
- **Least-privilege write.** Authenticate via GCP WIF using the **dedicated**
  write-only publish identity from TG1; the tool only creates objects (no
  read/list/delete). Idempotent (replace-on-write at the same build-id blob path),
  so re-runs are safe.
- **Add a `tools/**` labeler glob + CODEOWNERS.** `tools/` has no `labeler.yml`
  glob today, so a `tools/testgrid-publish/**` PR gets **no** area label until this
  child adds one (e.g. `area/ci` or a new `area/tools`). A narrower
  `tools/testgrid-publish/**` CODEOWNERS entry is optional (precedent: ADR-009's
  planned `pkg/health/**` self-labeling); the `* @nvidia/aicr-write` catch-all
  already owns the path.

**Out of scope.** Deploying the workers (TG3); the config-gen (TG3); the API/UI
(TG4a/TG4b); deciding *which* bundles to publish or the signer/registry policy (TG5);
a `junit/` directory inside the OCI evidence bundle (the parallel
ADR-007-extension epic).

**Key files.** New `tools/testgrid-publish/` (main.go, converter, gcs writer,
README); the shared coordinate helper under `pkg/recipe`; `pkg/validator/ctrf/types.go`
(CTRF model), `pkg/validator/ctrf/builder.go:56` (`ExitCodeToCTRFStatus`);
`pkg/validator/phases.go:34-45` (`PhaseDeployment`/`PhasePerformance`/
`PhaseConformance`, `PhaseOrder`); determinism precedent `pkg/serializer/yaml.go:31`,
`tools/bom/main.go:63,145`; pointer/signer shape `pkg/evidence/attestation/pointer.go`,
`signer.go`; verified-SAN provenance from `aicr evidence verify`
(`pkg/cli/evidence_verify.go:78-84`). `.github/labeler.yml`.

**Tests.** Golden-file converter tests (each CTRF status incl. `pending`→`<skipped/>`
and `other`→`<error>`; multi-CTRF phase-namespaced merge with no row collision;
mixed-suite aggregate counts; rounding-boundary `time`; over-cap rune-safe
`<system-out>`; byte-identical across two marshals; XML escaping). A determinism
test that runs the publisher twice with **clock skew between runs** and asserts
byte-identical output (proving timestamps come from the bundle, not `time.Now()`).
GCS path-construction table tests (sanitization, 4-level depth, empty→`unknown` for
required segments, **bare-intent tab for no-platform**, k8s truncation, build-id
monotonic+idempotent). started/finished assembly tests (binary `passed` from CTRF
summary; pinned metadata keys present; timestamps from predicate). A WIF/GCS write
test against a fake or emulated bucket asserting create-only behavior, that two
distinct runs produce two columns, and that re-ingesting the same run replaces
rather than duplicates. **75%** floor on the new package (`pkg=tools/testgrid-publish`
if measured as a package, else the converter sub-package). The shared coordinate
helper's golden test (both launch clouds) lives in `pkg/recipe`.

**Acceptance.** `tools/testgrid-publish` consumes an OCI evidence bundle, emits a
byte-deterministic 4-level CSP-first GCS feed via the shared coordinate helper
(timestamps from the bundle predicate; build-id monotonic+idempotent so distinct
runs are distinct ordered columns), with the pinned column-metadata key schema and
verified provenance, writes with the dedicated create-only identity, and is covered
by golden + path + determinism + write tests; `tools/**` self-labels.

---
_Child of #1263 · staged from `docs/design/010-aicr-testgrid.md` (TG2)._

srao-nv:
## Summary

Adds \`tools/testgrid-publish\`, a new CLI tool that converts AICR evidence OCI bundles into TestGrid-compatible GCS feed entries. Also wires the binary into \`.goreleaser.yaml\` and adds \`make testgrid-publish\` / \`make testgrid-publish-dryrun\` targets.

## Motivation / Context

Implements the TG2 publish path from the AICR TestGrid epic.

Fixes: #1267
Related: #1263, #1409

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tools/testgrid-publish` (new internal ops tool)

## Implementation Notes

**Pipeline** (`--bundle <oci-ref>` or `--bundle-dir <path>` → GCS):

```
bundle dir
  ├── recipe.yaml          → CoordinateFor() → GCS group/dashboard/tab
  ├── statement.intoto.json → AttestedAt, k8s_version, aicr_version
  └── ctrf/*.json           → jUnit XML
                                   ↓
  gs://<bucket>/groups/<service>/<accel-os>/<intent[-platform]>/<build-id>/
    started.json   (written first)
    artifacts/junit.xml
    finished.json  (written last — triggers TestGrid updater)
```

**Key decisions:**
- Upload order is strict (started → junit → finished); TestGrid updater ingests a build only after `finished.json` lands
- `--bundle-dir` flag allows local dry-runs without an OCI registry (`--dry-run` prints planned GCS paths)
- Criteria fields normalized (lowercase + trim) so `EKS` and ` eks ` both map to the correct GCS group
- Zero-test bundles rejected — crashed test runners write empty CTRF and would otherwise publish as SUCCESS
- `CoordinateFor` now delegates to `pkg/recipe.CoordinateFor` (#1409 merged); gains `/`-in-dimension rejection guard
- Signer extraction (`signer_identity`, `signer_issuer`) is a tracked stub — DSSE envelope parsing deferred to #1267 follow-up
- Bounded file reads (`readBoundedFile`) prevent OOM from hostile/corrupt bundle files

**Not included in this PR (follow-ups):**
- TG5: GitHub Actions workflow that calls `testgrid-publish` after `aicr validate`
- Cross-project IAM binding for GCS bucket `aicr-testgrid-staging`

## Testing

```bash
# Unit tests — all pass, -race clean
GOFLAGS="-mod=vendor" go test ./tools/testgrid-publish/... -race -count=1
# Coverage: 75.7% (project gate: 75%)

# Local dry-run (no OCI or GCS needed)
make testgrid-publish
./dist/testgrid-publish \
  --bundle-dir /tmp/bundle \
  --bucket aicr-testgrid-staging \
  --source-class uat \
  --dry-run
```

5 cross-review passes, 35 issues found and fixed before opening.

## Risk Assessment

- [x] **Low** — New isolated tool with no callers yet; only additive changes to `.goreleaser.yaml` and `Makefile`

**Rollout notes:** Tool is not invoked by any existing CI workflow. TG5 will wire it in a follow-up PR once GCS IAM is confirmed.

## Checklist


tools/testgrid-publish/main.go
```diff
@@ -0,0 +1,299 @@
...
+				Input:       cfg.bundleRef,
+				PlainHTTP:   cfg.plainHTTP,
+				InsecureTLS: cfg.insecureTLS,
+				// Allow unpinned tags for CI convenience; digest-pinning is
+				// enforced at the GitHub Actions level (TG5 uses digest refs).
+				AllowUnpinnedTag: true,
+			},
+			verifier.InputFormOCI,
+			nil, // pointer not used for direct OCI input
+		)
```
njhensley: 🛑 **Blocker — publishes a verdict to a trust surface without verifying the bundle signature.**

`MaterializeBundle` is called with `InputFormOCI` + `AllowUnpinnedTag:true`, and `verifier.VerifySignature` is never called anywhere in the tool. The bundle is pulled *by tag* and its SUCCESS/FAILURE verdict is published to TestGrid with no provenance check — content can be swapped between tag resolution and pull. Deferring pinning "to the GitHub Actions level (TG5)" is a legitimate decision, but it needs to be explicitly defended now, because TG5 will inherit it as a comment rather than a gate. Either verify the signature for `--source-class=uat`, or document loudly that the feed carries no provenance and is unsuitable for trust decisions.

.goreleaser.yaml
```diff
@@ -84,9 +84,26 @@ builds:
...
+      - GOFLAGS=-mod=vendor
+    flags:
+      - -trimpath
+    ldflags:
+      - -w -s -extldflags "-static"
+    goos:
+      - linux
+    goarch:
+      - amd64
+      - arm64
```
njhensley: 🛑 **Blocker — this build compiles every release but ships nowhere.**

The `testgrid-publish` build has no consumer: `archives` lists only `id: aicr`, `kos` covers only `aicr`/`aicrd`, and there is no `dockers`/`extra_files` entry for it. Unlike `aicrd` (archive-less *because* it ships via `kos:`), this binary has neither an archive nor an image — so goreleaser builds it on every release and never distributes it. Decide now: drop it from goreleaser and build via the new `make testgrid-publish` target inside the consuming workflow, **or** give it a real distribution path (`archives` id / `extra_files`).

🟡 *Related (minor):* `sboms:` (line ~113) uses `artifacts: binary` with no `ids:` filter, so if this build stays it gets an SBOM but no signing/provenance — half-in the supply-chain surface, which can read as "attested". Resolve together: either bring it under the attestation story or exclude it from SBOM with an `ids` filter + comment.

tools/testgrid-publish/main.go
```diff
@@ -0,0 +1,299 @@
...
+// readPointerFromAttestation reads signer information from the DSSE-wrapped
+// attestation bundle. Returns nil without error when no attestation is present
+// (unsigned bundles are valid).
+//
+// TODO: implement DSSE envelope parsing to extract PointerSigner fields
+// (identity, issuer) from the Sigstore signing certificate in
+// attestation.AttestationFilename. Tracking issue: NVIDIA/aicr#1267.
+func readPointerFromAttestation(_ string) (*attestation.Pointer, error) {
+	return nil, errors.New(errors.ErrCodeInternal, "signer extraction not yet implemented")
+}
```
njhensley, line 317: 🔴 **Major — signer columns are permanently blank.**

`readPointerFromAttestation` unconditionally returns an error, so `extractSigner` always returns `("", "")` and the `att.Signer` dereference just above is unreachable dead code. Result: `signer_identity`/`signer_issuer` — 2 of the 7 "single source of truth" schema columns — render "missing" in the TestGrid UI for **every** build, including correctly-signed UAT bundles. That is the exact drift the `schema.go` comment warns against. Either implement the DSSE/Sigstore-cert parse, or remove the two columns + dead deref and emit a one-time `slog.Warn` so operators know why they're blank rather than assuming a bug. (Tracked under #1267.)

tools/testgrid-publish/main.go
```diff
@@ -0,0 +1,299 @@
...
+		return errors.Wrap(errors.ErrCodeInvalidRequest, "coordinate resolution failed", err)
+	}
+
+	// ── 3. Extract AttestedAt, aicr_version, and k8s_version from the predicate
+	pred, predErr := loadPredicate(dir)
+	if predErr != nil {
+		// Unsigned / local bundles may not have a full predicate; use now.
+		slog.Warn("could not load predicate, using current time", "error", predErr)
+		pred = &attestation.Predicate{AttestedAt: time.Now().UTC()}
+	}
```
njhensley, line 196: 🔴 **Major — a corrupt predicate is silently masked as a successful publish.**

Any `loadPredicate` error — including `ErrCodeInvalidRequest` for a truncated/corrupt `statement.intoto.json`, not just the not-found case — is downgraded to `slog.Warn` + a fabricated `time.Now()` timestamp. The comment only justifies the *unsigned/local → no predicate* case. This is the CLAUDE.md "fail-closed vs warn-and-continue" anti-pattern applied to a parse failure. Distinguish `ErrCodeNotFound` (benign → fall back) from a malformed predicate (→ return the error).

tools/testgrid-publish/gcs_test.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	err := writeGCS(context.Background(),
+		"aicr-testgrid-staging",
+		"groups/eks/h100-ubuntu/training/1749600000-abc12345",
+		started, finished, []byte("<testsuites/>"))
+	if err != nil {
+		t.Fatalf("writeGCS() unexpected error: %v", err)
+	}
+}
```
njhensley, line 130: 🔴 **Major (test) — the upload-order invariant, the tool's core safety property, is never asserted.**

The fake `gcloud` is `#!/bin/sh` (+ optional `exit 1`) — it copies and records nothing. `TestWriteGCS` only proves "three invocations exited 0"; it cannot catch a reordering of the `uploads` slice that would break TestGrid's `finished.json`-last completeness detection (the invariant `gcs.go` documents as mandatory). Make the fake record its `dst` argument to a temp log and assert the sequence ends `started.json` → `artifacts/junit.xml` → `finished.json`.

tools/testgrid-publish/main.go
```diff
@@ -0,0 +1,299 @@
...
+func run(ctx context.Context, cfg runConfig) error {
+	var dir, digest string
+	var cleanup func()
+
+	if cfg.bundleDir != "" {
+		// Pre-materialized bundle directory — skip OCI pull.
+		dir = cfg.bundleDir
+		digest = "local"
+		cleanup = func() {}
+		slog.Info("using pre-materialized bundle", "dir", dir)
```
njhensley, line 130: 🟡 **Minor — `--bundle-dir` bypasses bundle-marker validation and nested `summary-bundle/` resolution.**

The OCI path resolves a nested `summary-bundle/` subdir and validates markers via `materializeDir`; `--bundle-dir` hand-assigns `dir` verbatim. Pointing at a bundle *parent* (natural, since bundles unpack with a `summary-bundle/` root) yields a confusing "recipe.yaml not found". Route `--bundle-dir` through `verifier.MaterializeBundle(ctx, VerifyOptions{Input: cfg.bundleDir}, verifier.InputFormDir, nil)` to reuse both behaviors.

tools/testgrid-publish/coordinate.go
```diff
@@ -0,0 +1,51 @@
...
+// (lowercase, trimSpace) has already been applied by parseCriteria.
+func toRecipeCriteria(c RecipeCriteria) *recipe.Criteria {
+	return &recipe.Criteria{
+		Service:     recipe.CriteriaServiceType(c.Service),
+		Accelerator: recipe.CriteriaAcceleratorType(c.Accelerator),
+		OS:          recipe.CriteriaOSType(c.OS),
+		Intent:      recipe.CriteriaIntentType(c.Intent),
+		Platform:    recipe.CriteriaPlatformType(c.Platform),
+	}
+}
```
njhensley, line 52: 🟡 **Minor — criteria values are raw-cast, never validated against the enum allowlist.**

`toRecipeCriteria` does bare type conversions; `CoordinateFor` only rejects empty/`any`/`/`. A hostile `recipe.yaml` (`service: "foo bar"`, `accelerator: ".."`) yields a valid `Coordinate` and publishes under an arbitrary prefix in the operator's bucket. Confirmed **not** exploitable as RCE/traversal (the `gs://` prefix + `/`-rejection + direct `exec` argv neutralize flag/shell injection), but it pollutes the namespace. Validate against the registry/allowlist (`ParseService`/`ValidateWithRegistry`) and fail closed on unknown dimensions.

tools/testgrid-publish/gcs.go
```diff
@@ -0,0 +1,123 @@
...
+		{filepath.Join(tmp, "finished.json"), fmt.Sprintf("gs://%s/%s/finished.json", bucket, prefix)},
+	}
+	for _, u := range uploads {
+		if err := ctx.Err(); err != nil {
+			return errors.Wrap(errors.ErrCodeUnavailable, "GCS upload canceled", err)
+		}
+		if err := gcloudCopy(ctx, u.localPath, u.gcsObject); err != nil {
+			return err
+		}
+	}
```
njhensley, line 112: 🟡 **Minor — partial-upload window (+ `--bundle-dir` build-id collision).**

If `started.json` lands but `junit`/`finished` fails, a `started`-without-`finished` build is left in GCS with no rollback. This self-heals via the deterministic build-id (retry overwrites) — **but only for OCI bundles**. For `--bundle-dir`, `digest="local"` → `shortHex` returns `"local"` → build-id `{now-unix}-local`, so two distinct local bundles in the same second collide and the idempotency mitigation does not hold. Either best-effort `gcloud storage rm` the partial objects on failure, or fold a content hash into the `local` suffix (and document the dev-path limitation).

tools/testgrid-publish/main.go
```diff
@@ -0,0 +1,299 @@
...
+		plainHTTP   bool
+		insecureTLS bool
+		dryRun      bool
+	)
+	flag.StringVar(&bundleRef, "bundle", "", "OCI reference of the evidence bundle")
+	flag.StringVar(&bundleDir, "bundle-dir", "", "pre-materialized bundle directory (alternative to --bundle)")
+	flag.StringVar(&bucket, "bucket", "", "GCS bucket to publish to (required)")
+	flag.StringVar(&sourceClass, "source-class", sourceClassUAT, "bundle origin: "+sourceClassUAT+" or "+sourceClassCommunity)
+	flag.BoolVar(&plainHTTP, "plain-http", false, "use plain HTTP for OCI registry (dev only)")
+	flag.BoolVar(&insecureTLS, "insecure-tls", false, "skip TLS verification for OCI registry")
```
njhensley: 🟡 **Minor — `--insecure-tls` lacks the "(dev only)" label and isn't env-gated.**

`--plain-http` says "(dev only)"; `--insecure-tls` doesn't, and neither is gated behind a dev/CI env. Combined with `AllowUnpinnedTag` + no signature verification, `--insecure-tls` enables a trivial MITM bundle swap. Mirror the wording and ideally refuse both unless an explicit dev env is set.

tools/testgrid-publish/schema_test.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+	seen := make(map[string]bool)
+	for _, k := range keys {
+		if k == "" {
+			t.Error("MetaKeys() contains empty string")
+		}
+		if seen[k] {
+			t.Errorf("duplicate key %q in MetaKeys()", k)
+		}
+		seen[k] = true
```
njhensley, line 57: 🟡 **Minor (test) — this "contract" test is a self-consistency tautology.**

`TestMetaKeys` checks `MetaKeys()` against the constants it's built from — it never verifies that the `started.Metadata` map actually emits exactly those keys, which is the stated purpose. Add a test asserting set-equality between the emitted started.json keys and `MetaKeys()`. (A true cross-repo contract with config-gen may be out of scope — at least re-document the limitation.)

tools/testgrid-publish/bundle.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	defer func() { _ = f.Close() }()
+	data, err := io.ReadAll(io.LimitReader(f, maxBytes+1))
+	if err != nil {
+		return nil, err
+	}
+	if int64(len(data)) > maxBytes {
+		return nil, errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("%s exceeds %d-byte size limit", path, maxBytes))
+	}
```
njhensley, line 47: 🟡 **Minor (test) — safety-critical branches are uncovered (grouped).**

The branches a publish tool most wants tested are uncovered: this `readBoundedFile` oversize-rejection branch; `convertCTRF`'s `StatusPending` mapping and the `default:` unknown-status fail-closed branch (`ctrf.go`); and `TestGcloudCopyCanceled` asserts nothing (`_ = err`). These are the realistic source of the 75.7%↔77.1% coverage drift. Add the oversize case, the pending + bogus-status cases, and either make the cancel test deterministic or drop it.

tools/testgrid-publish/main.go
```diff
@@ -0,0 +1,299 @@
...
+		fmt.Fprintln(os.Stderr, "testgrid-publish: --bucket is required")
+		os.Exit(1)
+	}
+	if sourceClass != sourceClassUAT && sourceClass != sourceClassCommunity {
+		fmt.Fprintf(os.Stderr, "testgrid-publish: --source-class must be %q or %q\n",
+			sourceClassUAT, sourceClassCommunity)
+		os.Exit(1)
+	}
+
+	ctx := context.Background()
```
njhensley, line 96: 🟡 **Minor — pre-`oras.Copy` registry handshake runs under an unbounded context.**

`context.Background()` flows into `MaterializeBundle`. `oras.Copy` itself is internally bounded, but the registry resolve/auth handshake before it is not — the realistic hang. Per CLAUDE.md, derive a bounded root here: `ctx, cancel := context.WithTimeout(context.Background(), defaults.EvidenceBundlePushTimeout)`.

tools/testgrid-publish/main.go
```diff
@@ -0,0 +1,299 @@
...
+		}
+	}
+}
+
+// junitAllPassed returns true when the jUnit XML has no failures or errors.
+func junitAllPassed(xml []byte) bool {
+	// Quick scan: if any <failure or <error element exists, the run failed.
+	s := string(xml)
+	return !strings.Contains(s, "<failure") && !strings.Contains(s, "<error")
+}
```
njhensley: 🔵 **Nitpick — `junitAllPassed` re-derives pass/fail by string-scanning the marshaled XML.**

It scans for `<failure`/`<error` instead of using the exact `suite.Failures`/`suite.Errors` counts `convertCTRF` already computed. No false positive today (XML escapes names/messages), but brittle to any future formatting change. Have `convertCTRF` return the aggregate counts and compute `allPassed` from them.

tools/testgrid-publish/gcs.go
```diff
@@ -0,0 +1,123 @@
...
+	return nil
+}
+
+// gcloudCopy uploads a single local file to GCS via `gcloud storage cp`.
+// Authentication is handled by ADC / the gcloud CLI configuration.
+func gcloudCopy(ctx context.Context, src, dst string) error {
```
njhensley, line 118: 🔵 **Nitpick — no `gcloud` preflight (PATH / auth).**

Failure surfaces late as a generic exec error after pull/parse/stage. An `exec.LookPath("gcloud")` + lightweight auth probe up front (and in `--dry-run`) would make a green dry-run actually predict a working publish.

njhensley (changes requested):
## Multi-persona review (Go correctness · Security · CI/CD · Test-quality · Operability), confirmed by senior meta-review

Solid, well-iterated tool — hermetic tests, bounded reads, fail-closed CTRF handling, deterministic output. Requesting changes on **2 merge-blockers** + **3 majors** before this is wired into TG5. The tool has no callers yet, so most latent bugs are major/minor — except two design decisions that future wiring will silently paper over.

**Tier legend:** 🛑 Blocker · 🔴 Major · 🟡 Minor · 🔵 Nitpick (inline below).

### Top items
- 🛑 **No signature verification before publishing a verdict** to a trust surface (`AllowUnpinnedTag` + no `VerifySignature`).
- 🛑 **Orphaned goreleaser build** — compiles every release, ships nowhere.
- 🔴 **Signer columns permanently blank** (`readPointerFromAttestation` is a hard error stub → 2 of 7 schema columns always "missing").
- 🔴 **Corrupt predicate masked as success** (any parse error → `time.Now()` fallback).
- 🔴 **Upload-order invariant never tested** (fake `gcloud` records nothing; a reordering bug passes green).

### Considered and intentionally *not* flagged
- **gcloud arg/shell injection** — clean (no shell; every arg prefixed with literal `gs://`/tmpdir).
- **Path traversal** — clean (all `filepath.Join` segments are trusted constants).
- **`mod_timestamp` anchor refactor** — valid YAML; `&build-timestamp` is defined earlier on the `aicr` build, so the alias is sound and reproducibility is unchanged.
- **Bounded reads, temp-dir/file-handle cleanup, error wrapping/codes** — verified correct.
- **Dropped as noise:** 2-min timeout sizing for 3 gcloud cold-starts (theoretical); `make` target omitting release build flags (expected dev-convenience divergence).

*Review assembled from 5 reviewer personas and a senior meta-reviewer that independently re-verified each finding against source at `8535f2e` (zero false positives).*

tools/testgrid-publish/coordinate.go
```diff
@@ -0,0 +1,52 @@
...
+// plain-string RecipeCriteria to a typed *recipe.Criteria.
+func CoordinateFor(c RecipeCriteria) (recipe.Coordinate, error) {
+	return recipe.CoordinateFor(toRecipeCriteria(c))
+}
+
+// toRecipeCriteria converts our local plain-string struct to the typed
+// *recipe.Criteria required by pkg/recipe.CoordinateFor. Normalization
+// (lowercase, trimSpace) has already been applied by parseCriteria.
+// Validation (empty, "any", "/") is enforced by recipe.CoordinateFor itself.
+func toRecipeCriteria(c RecipeCriteria) *recipe.Criteria {
```
njhensley, line 44: 🟡 **Follow-up (non-blocking) — F6.** Criteria are still raw-cast here; `recipe.CoordinateFor` only rejects empty/`any`/`/`, so a hostile `recipe.yaml` (`service: "foo bar"`, `accelerator: ".."`) still resolves to a valid coordinate and publishes under an arbitrary prefix in the operator's bucket. Confirmed not an escape/RCE (the `gs://` prefix + `/`-rejection neutralize traversal/flag injection), so fine to defer — but validating against the registry allowlist (`ParseService`/`ValidateWithRegistry`) before building the coordinate would close the namespace-pollution gap.

tools/testgrid-publish/buildid.go
```diff
@@ -0,0 +1,58 @@
...
+// digest is empty or too short.
+func shortHex(digest string, n int) string {
+	// Strip algorithm prefix (e.g. "sha256:").
+	if i := strings.IndexByte(digest, ':'); i >= 0 {
+		digest = digest[i+1:]
+	}
+	if len(digest) >= n {
+		return digest[len(digest)-n:]
+	}
+	if digest != "" {
```
njhensley, line 54: 🟡 **Follow-up (non-blocking) — F12.** For `--bundle-dir`, `digest="local"` so `shortHex` returns `"local"` and the build-id collapses to `{now-unix}-local`; two distinct local bundles validated in the same second collide and overwrite in GCS. Dev-path only, so deferring is fine — folding a content hash into the suffix when `digest=="local"` would make local runs collision-safe.

njhensley (approved):
## Re-review (head `192457b9`) — addressed; approving ✅

Re-checked every prior finding against the new sources, verified the dependency APIs, and ran the package tests locally (pass, 77.1% coverage, `go vet` clean). Both blockers and all three majors are resolved or explicitly defended; what remains is two minor, low-risk items on a tool with no callers yet. Clearing my earlier `CHANGES_REQUESTED`.

**Blockers**
- 🛑→✅ **Orphaned goreleaser build** — build block removed from `.goreleaser.yaml`; built via `make testgrid-publish` in the workflow instead. Also dissolves the SBOM-without-provenance concern.
- 🛑→⚠️ **No signature verification** — `AllowUnpinnedTag` kept, now explicitly defended in-code with digest-pinning deferred to TG5 at the workflow level. Acceptable while unwired — please make sure TG5 actually enforces the pin.

**Majors**
- 🔴→✅ **Corrupt predicate** — now fails closed; only a *missing* predicate falls back to `time.Now()`. Bonus `predicateType` validation added.
- 🔴→✅ **Upload-order invariant** — fake `gcloud` records destinations and `TestWriteGCS` now asserts started→junit→finished. 👍
- 🔴→⚠️ **Signer columns blank** — still a stub, but the doc is now honest and warns future implementers; deferred to #1267.

**Minors/nitpicks fixed:** unbounded pull context (now bounded), `--bundle-dir` summary-bundle auto-resolution, counts-based `allPassed` (no more XML string-scan), `gcloud` preflight (`LookPath`), `--insecure-tls` "(dev only)" label, real `MetaKeys()`↔started.json contract test, and coverage for oversize/pending/unknown-status/cancel paths.

**Remaining (non-blocking follow-ups, inline below):** F6 criteria enum validation and F12 `--bundle-dir` build-id collision — both minor and fine to defer.

Nice iteration — the test hardening in particular is exactly right.

Issue #1437, opened 2026-06-23
Evidence gate: surface underlying verify failure cause (e.g. registry 403)


njhensley:
Part of #1431.

## Summary

When `aicr evidence verify` fails inside the gate, the PR comment table and the job logs only convey a coarse verdict (effectively "invalid") with no actionable cause. In #1427 the contributor's fork GHCR package was private, so the gate's verify pull got an HTTP **403 (registry not accessible)** — but that root cause was not discoverable from either the table output or the workflow logs. Surface the underlying error.

## Motivation / Context

- `aicr evidence verify` distinguishes "bundle invalid" (exit 2) from "recorded validator failures" (exit 1) from "valid" (exit 0), but the gate collapses the failure into a single column value without the *reason* (registry auth/403, not-found/404, signature mismatch, schema/integrity error).
- A contributor can't self-serve the fix (make the package public) if the gate never tells them the pull was a 403.

## Proposed approach

- Ensure `aicr evidence verify --format json` emits a machine-readable error/cause field for the failure path (registry status code, error code, message); add it if missing.
- In `.github/scripts/recipe-evidence-check.sh`, capture stderr + the JSON cause from the verify call and:
  - render a short classified cause in the table's verify column or as a footnote (e.g. "invalid — registry 403 (package not public?)"), and
  - echo the full error into the job log (currently swallowed).
- Classify the common cases so the message is actionable: `403/401` → "registry not accessible (make the fork's aicr-evidence package public)"; `404` → "bundle not found at ref"; signature/schema → as reported.

## Acceptance criteria

- [ ] A 403 on the evidence pull shows a clear, actionable cause in the PR comment and in the job log — not just "invalid."
- [ ] `evidence verify --format json` carries a structured error/cause for non-zero exits.
- [ ] Other failure classes (404, signature mismatch, schema/integrity) render distinct, readable causes.

## References

- `.github/scripts/recipe-evidence-check.sh` (verify invocation + report rendering)
- `pkg/cli/evidence_verify.go` (exit codes, `--format json`); `pkg/evidence/verifier/`
- Related: the fork-based signing workflow / two-phase publish issues (fork GHCR visibility is the fix the message should point at).

mchmarny:
Implemented in #1448 (classified verify-failure causes rendered in the gate comment + structured failureCause from #1445). Merged to main.

Issue #1435, opened 2026-06-23
Evidence gate: scope registry/component cascade to actually-affected recipes


njhensley:
Part of #1431.

## Summary

The gate currently treats any change to `recipes/registry.yaml` (or `recipes/overlays/base.yaml`) as a broad-impact event and promotes **every** leaf recipe to "affected" (`promote_all=true` in `recipe-evidence-check.sh`). PR #1418 only changed the `aws-efa` registry entry, yet the gate flagged non-AWS recipes (e.g. `h100-gke-cos-training`) as drifted. Tighten the cascade to the recipes that transitively reference the *specific* registry entries / component values that changed.

## Motivation / Context

- The registry is the canonical source for per-component defaults, so the whole-file "registry changed → all recipes affected" heuristic is correct-but-coarse: an `aws-efa`-only edit can't change the rendered digest of a recipe that never references `aws-efa`.
- The material-slice model (`docs/design/007-recipe-evidence/04-material-slice-algorithm.md`) already defines a recipe's material inputs as the registry entries for components *that recipe transitively references* — the gate's file-level promotion is coarser than the documented model.

## Proposed approach

- Diff `registry.yaml` at the entry (component) level rather than file level; build the set of changed component names.
- Mark a recipe affected only if its resolved component set intersects the changed components. The digest re-computation (`aicr evidence digest -r`) is the ground truth — a recipe whose digest is unchanged shouldn't be listed as drifted regardless of promotion.
- Consider: skip listing a "promoted" recipe entirely when its recomputed digest matches its committed pointer (no drift = no row) — this also shrinks the scary list from the curation issue.
- Apply the same treatment to `base.yaml` if feasible (base-chain membership is already computed for the ancestor rule).

## Acceptance criteria

- [ ] An `aws-efa`-only registry edit flags only recipes that reference `aws-efa`.
- [ ] Recipes whose recomputed digest matches their pointer are not reported as drifted.
- [ ] Behavior documented in `04-material-slice-algorithm.md` / `09-ci-and-maintainer.md`.

## References

- `.github/scripts/recipe-evidence-check.sh` (`promote_all`, affected-detection rules, digest compare)
- `aicr evidence digest`: `pkg/cli/evidence_digest.go`, `pkg/evidence/attestation/recipe_digest.go`
- Material model: `docs/design/007-recipe-evidence/04-material-slice-algorithm.md`

mchmarny:
[...]

Issue #1431, opened 2026-06-23
Evidence gate pilot: curation, fork-based publishing, and cascade scoping


njhensley:
## Summary

The recipe-evidence drift gate (ADR-007, `docs/design/007-recipe-evidence.md`) went live as a **warning-only** pilot. Real PR traffic — #1418 (aws-efa registry bump) and #1427 (re-sign of the two recipes that carry pointers) — surfaced three tunable rough edges and a contributor DevX blocker. This epic tracks tuning the pilot into something low-noise and easy to contribute to.

## Context

- **#1418** changed a single registry entry (`aws-efa`) but the gate promoted ~all leaf recipes to "affected," producing an alarming warning list (despite being warning-only).
- **#1427** re-signed the only two recipes that carry committed pointers (`gb200-eks-ubuntu-training`, `h100-gke-cos-training`); the gate then reported them "invalid" because it got an HTTP **403** pulling the evidence bundle from a contributor fork's (private) container registry — a cause that was not discoverable from the logs or the table.
- Sigstore keyless signing from a laptop is blocked on corporate VPN and some home networks (upstream Fulcio IP-level rejection), forcing contributors onto a phone hotspot.

The gate is warning-only and never blocks merge — this epic is pilot tuning, not a correctness fix.

## Child issues

- [ ] #1432 — Curated "protected recipes" set (implicit: has a committed pointer)
- [ ] #1433 — Fork-based evidence signing workflow, driven by committed unsigned pointers
- [ ] #1434 — `aicr evidence`: two-phase push-without-sign + sign-existing support
- [ ] #1435 — Scope registry/component cascade to actually-affected recipes
- [ ] #1436 — Docs: Sigstore/Fulcio network blocking + recommended publishing path
- [ ] #1437 — Surface the underlying verify failure cause (e.g. registry 403)

mchmarny:
All six child issues delivered across #1445, #1446, #1448, #1451 (merged to main): two-phase publish (--no-sign + evidence sign), curated protected set, component-scoped cascade, classified verify causes, pending-signature verify state, fork-based signing workflow, and connectivity docs.

PR #1448, opened 2026-06-24
feat(evidence-gate): protected/other partition + component-scoped cascade

### LINKED ISSUE #1432: Evidence gate: curated 'protected recipes' set (implicit by committed pointer)
Part of #1431.

## Summary

Distinguish recipes the evidence gate actively *protects* from the ~61 leaf recipes that have no evidence yet. A recipe is **protected implicitly** — iff it has a committed pointer in `recipes/evidence/`. Today only `gb200-eks-ubuntu-training` and `h100-gke-cos-training` carry pointers; every other recipe renders as "missing," which makes the gate comment on broad-impact PRs (e.g. #1418) look alarming despite being warning-only.

## Motivation / Context

- Evidence is hardware-gated: a bundle is only valid on a cluster matching the recipe's full criteria (accelerator × service × OS × intent × platform). Most clusters aren't readily available, and `*-kind-*` recipes can't produce real evidence. "Add pointers to all 63 recipes" is impractical.
- The signal we want is: *for the recipes we've chosen to protect, is their committed evidence stale or invalid?* Everything else is best-effort.
- Using "has a committed pointer" as the definition means zero new files to maintain — the set grows naturally as evidence is added, and converges with the maturing testgrid work (the recipes we validate on real hardware).

## Proposed approach

- In `.github/scripts/recipe-evidence-check.sh`, treat "protected" as "has a pointer file in `recipes/evidence/`."
- Split the report into **Protected** (full status: pointer / verify / digest-match, drift highlighted) and **Other affected** (collapsed `<details>` summary count, or omitted). Keep warning-only semantics.

## Acceptance criteria

- [ ] Gate comment clearly separates protected recipes (have pointers) from "no evidence yet" recipes.
- [ ] On a broad-impact PR, the all-recipes list is collapsed/summarized, not the headline.
- [ ] The implicit "has a pointer = protected" definition is documented in ADR-007 / `09-ci-and-maintainer.md`.

## References

- `.github/scripts/recipe-evidence-check.sh` (affected-detection + report rendering)
- `.github/workflows/recipe-evidence-comment.yaml`
- `recipes/evidence/{gb200-eks-ubuntu-training,h100-gke-cos-training}.yaml`
- ADR-007 trust model: `docs/design/007-recipe-evidence.md`

mchmarny:
## Summary

Tunes the warning-only recipe-evidence gate (`.github/scripts/recipe-evidence-check.sh`) so broad-impact PRs stop rendering an alarming wall of "missing"/"drifted" rows. Three changes: a protected-vs-other partition, a component-scoped registry cascade, and classified verify-failure causes in the comment.

## Motivation / Context

Part of the evidence-gate pilot epic. PR #1418 (an `aws-efa`-only registry bump) promoted ~all leaf recipes to "affected" and #1427 surfaced an unhelpful bare "invalid". This is the gate-report half of that tuning; it consumes the `failureCause` JSON contract added in #1445.

Fixes: #1432, #1435, #1437
Related: #1431

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Other: CI gate script (`.github/scripts/recipe-evidence-check.sh`) + ADR-007

## Implementation Notes

- **Protected vs. other (#1432):** "protected" = has a committed pointer in `recipes/evidence/`. The gate verifies the protected set (status table); affected-but-unevidenced recipes collapse into a single `<details>` count — no longer per-recipe "missing" warnings. No new files to maintain.
- **Component-scoped cascade (#1435):** `registry.yaml` changes are diffed at the component-entry level; a recipe is affected only if its resolved component set (walked across base chain + `spec.mixins`) intersects the changed entries. `base.yaml` stays broad. The recomputed digest remains ground truth (match ⇒ not drift).
- **Classified causes (#1437, script half):** parse `failureCause {class, httpStatus, hint}` + `pending` from `aicr evidence verify --format json`; render an actionable cause in the verify column and echo the full error to the job log.
- ADR-007 gains a "Gate report structure" subsection.

## Testing

Local smoke tests (built `aicr`, temp commits, `recipe-evidence-check.sh`):

| Scenario | Result |
|---|---|
| `aws-efa`-only registry edit | 23 affected = **1 protected** (`gb200-eks-ubuntu-training`) + **22 EKS collapsed**; GKE `h100-gke-cos-training` **not affected** |
| `base.yaml` edit (broad) | 63 affected → **2 protected + 61 collapsed**; warnings **63 → 4** |
| single overlay edit | Rules 1–4 still scope correctly (3 affected) |

`bash -n` + `shellcheck -S warning` clean. (No Go changed; the consumed `failureCause`/`pending` JSON shipped in #1445.)

## Risk Assessment

- [x] **Low** — Warning-only gate; never blocks merge. Changes only the CI script + a doc. Worst case is over/under-listing in a PR comment, with the digest comparison as ground truth.

**Rollout notes:** No config or workflow changes; the workflow invokes the same script with the same env contract.

## Checklist

- [x] Linter passes (`shellcheck`, `bash -n`)
- [x] I updated docs if user-facing behavior changed (ADR-007)

### DIFF
--- a/.github/scripts/recipe-evidence-check.sh
+++ b/.github/scripts/recipe-evidence-check.sh
@@ -8,9 +8,16 @@
 # canonical digest. Writes a Markdown report; never blocks the PR.
 #
 # A recipe is "affected" when the PR touches its overlay, an ancestor
-# overlay in its base chain, a referenced component values file, or its
-# own evidence pointer (recipes/evidence/<slug>.yaml) — so a pointer-only
+# overlay in its base chain, a referenced component values file, a
+# *changed* registry component it transitively references, or its own
+# evidence pointer (recipes/evidence/<slug>.yaml) — so a pointer-only
 # "refresh evidence" PR is verified rather than silently skipped.
+#
+# The report separates "protected" recipes — those with a committed
+# pointer in recipes/evidence/, the set the gate actively verifies — from
+# "other affected" recipes that have no evidence yet (best-effort, not a
+# warning). This keeps a broad-impact PR from rendering dozens of alarming
+# "missing" rows for recipes nobody has produced evidence for.
 #
 # Required env:
 #   AICR        path to the aicr binary (built from a trusted source)
@@ -22,8 +29,8 @@
 #   REPO_URL    base URL for absolute links in the report (e.g.
 #               https://github.com/NVIDIA/aicr); when unset, the
 #               trust-model link in the trailer is omitted
-#   MAX_ROWS    cap on per-recipe rows (default 80) to keep the report
-#               under GitHub's ~65KB comment-body cap on broad-impact PRs
+#   MAX_ROWS    cap on protected-recipe rows (default 80) to keep the
+#               report under GitHub's ~65KB comment-body cap
 #
 # Local invocation:
 #   make build  # or `go build -o ./bin/aicr ./cmd/aicr`
@@ -38,10 +45,6 @@
 #     `aicr validate --emit-attestation` writes pointers named after
 #     the criteria-derived slug (RecipeNameFor). Overlays not following
 #     the canonical naming will be reported as missing-pointer.
-#   * Discovery walks `spec.base` ancestors but not `spec.mixins`, and
-#     promote-all only matches recipes/registry.yaml or recipes/
-#     overlays/base.yaml. Mixin/check/component edits outside literal
-#     valuesFile refs are not promoted. Parity with kwok-recipes.
 #   * `aicr evidence verify` fetches OCI artifacts from PR-controlled
 #     URLs; an allow-list of trusted registries is a follow-up.
 
@@ -54,6 +57,25 @@ set -euo pipefail
 
 REPO_URL="${REPO_URL:-}"
 MAX_ROWS="${MAX_ROWS:-80}"
+# Cap on the collapsed "other affected" list so a broad PR can't blow past
+# GitHub's ~65KB comment-body limit.
+OTHER_MAX="${OTHER_MAX:-200}"
+
+# md_escape sanitizes a value rendered into a Markdown table cell. Slugs and
+# verifier-supplied hints can in principle carry characters that break the
+# table or the surrounding backticks; collapse newlines, escape pipes, and
+# drop backticks so a value can't break out of a code span.
+md_escape() {
+  printf '%s' "$1" | tr '\n' ' ' | sed 's/|/\\|/g; s/`//g'
+}
+
+# log_sanitize strips characters from PR-controlled text before it is echoed
+# into a `::warning::`/`::error::` workflow command, so the text cannot inject
+# its own workflow commands. Collapses newlines and carriage returns, and
+# neutralizes `::`.
+log_sanitize() {
+  printf '%s' "$1" | tr '\n\r' '  ' | sed 's/::/:_:/g'
+}
 
 # Three-dot range so we only see what the PR actually changed (relative
 # to its merge base) — not main commits that landed after PR open.
@@ -63,10 +85,89 @@ if ! changed_files=$(git diff --name-only "${BASE_SHA}...${HEAD_SHA}" 2>&1); the
   exit 1
 fi
 
+# A change to recipes/overlays/base.yaml is genuinely broad — it sits at
+# the root of (almost) every base chain — so it still promotes all leaf
+# recipes. A registry.yaml change is NOT treated as broad: instead we
+# compute which component *entries* changed and promote only the recipes
+# that transitively reference one of them (see changed_components + Rule 5).
 promote_all=false
-if echo "$changed_files" | grep -qE '^recipes/(registry\.yaml|overlays/base\.yaml)$'; then
+if printf '%s\n' "$changed_files" | grep -qxF 'recipes/overlays/base.yaml'; then
   promote_all=true
 fi
+
+# Component-level registry scoping (#1435). When recipes/registry.yaml
+# changes, diff it at the component-entry level between BASE and HEAD and
+# collect the names whose entry was added, removed, or modified. An
+# `aws-efa`-only edit then flags only recipes that reference `aws-efa`,
+# not every leaf. Empty when registry.yaml is unchanged or only changed
+# cosmetically (no entry differs).
+registry_changed=false
+changed_components=""
+if printf '%s\n' "$changed_files" | grep -qxF 'recipes/registry.yaml'; then
+  registry_changed=true
+  reg_base=$(mktemp)
+  reg_head=$(mktemp)
+  # A revision missing the file (e.g. registry.yaml newly added) yields an
+  # empty doc, so every HEAD component reads as added — fail-open to broad
+  # for that recipe set, which is the safe direction for a warning gate.
+  git show "${BASE_SHA}:recipes/registry.yaml" >"$reg_base" 2>/dev/null || : >"$reg_base"
+  git show "${HEAD_SHA}:recipes/registry.yaml" >"$reg_head" 2>/dev/null || : >"$reg_head"
+  comp_names=$( { yq eval '.components[].name // ""' "$reg_base" 2>/dev/null || true; \
+                  yq eval '.components[].name // ""' "$reg_head" 2>/dev/null || true; } \
+                | grep -v '^$' | sort -u )
+  while IFS= read -r cn; do
+    [[ -z "$cn" ]] && continue
+    # Pass the name via strenv() rather than string-interpolating it into the
+    # yq expression: a PR-controlled component name must be data, not code.
+    # strenv (not env) keeps it a string so a scalar-looking name (`true`,
+    # `null`, `123`) isn't YAML-typed and made to miss its registry entry.
+    b=$(CN="$cn" yq eval '.components[] | select(.name == strenv(CN))' "$reg_base" 2>/dev/null || true)
+    h=$(CN="$cn" yq eval '.components[] | select(.name == strenv(CN))' "$reg_head" 2>/dev/null || true)
+    if [[ "$b" != "$h" ]]; then
+      changed_components="${changed_components}${cn}"$'\n'
+    fi
+  done <<<"$comp_names"
+  changed_components=$(printf '%s' "$changed_components" | grep -v '^$' | sort -u || true)
+  rm -f "$reg_base" "$reg_head"
+  cc_count=$(printf '%s\n' "$changed_components" | grep -c . || true)
+  echo "Changed registry components: ${cc_count}"
+fi
+
+# recipe_components <overlay-path> — print the sorted, unique set of
+# component names a recipe resolves to, using the *same* engine the digest
+# uses: `aicr recipe` resolves by criteria across all overlays, with an
+# unconditional base.yaml merge and criteria-matched wildcard overlays
+# (e.g. monitoring-hpa injecting prometheus-adapter). Hand-walking
+# spec.base/spec.mixins misses base.yaml and wildcard injections, which
+# would make Rule 5 a false-negative for base/wildcard components — exactly
+# the registry entries this scoping targets. Resolving via aicr keeps the
+# affected-set consistent with the digest comparison downstream.
+recipe_components() {
+  local overlay="$1" service accel os intent platform
+  service=$(yq eval '.spec.criteria.service // ""' "$overlay" 2>/dev/null || true)
+  accel=$(yq eval '.spec.criteria.accelerator // ""' "$overlay" 2>/dev/null || true)
+  os=$(yq eval '.spec.criteria.os // ""' "$overlay" 2>/dev/null || true)
+  intent=$(yq eval '.spec.criteria.intent // ""' "$overlay" 2>/dev/null || true)
+  platform=$(yq eval '.spec.criteria.platform // ""' "$overlay" 2>/dev/null || true)
+
+  local args=()
+  [[ -n "$service" && "$service" != "null" ]] && args+=(--service "$service")
+  [[ -n "$accel" && "$accel" != "null" ]] && args+=(--accelerator "$accel")
+  [[ -n "$os" && "$os" != "null" ]] && args+=(--os "$os")
+  [[ -n "$intent" && "$intent" != "null" ]] && args+=(--intent "$intent")
+  [[ -n "$platform" && "$platform" != "null" ]] && args+=(--platform "$platform")
+
+  : "${AICR_TIMEOUT:=30s}"
+  # Return non-zero when the resolver itself fails (timeout / error), so the
+  # caller can fail OPEN (include the recipe) rather than mistake a resolver
+  # hiccup for "references no changed component" — a false negative is the
+  # dangerous direction for a protection gate.
+  local out
+  if ! out=$(timeout "$AICR_TIMEOUT" "$AICR" recipe "${args[@]}" --format json 2>/dev/null); then
+    return 1
+  fi
+  printf '%s\n' "$out" | jq -r '.componentRefs[].name // empty' 2>/dev/null | sort -u
+}
 
 affected="[]"
 for overlay in recipes/overlays/*.yaml; do
@@ -108,7 +209,7 @@ for overlay in recipes/overlays/*.yaml; do
       parent=$(yq eval '.spec.base // ""' "$current" 2>/dev/null || true)
       if [[ -z "$parent" || "$parent" == "null" ]]; then break; fi
       if [[ -n "${visited_r2[$parent]:-}" ]]; then
-        echo "::warning::cyclic base chain detected at overlay '${name}' (re-visited '${parent}')"
+        echo "::warning::cyclic base chain detected at overlay '$(log_sanitize "$name")' (re-visited '$(log_sanitize "$parent")')"
         break
       fi
       visited_r2[$parent]=1
@@ -136,7 +237,7 @@ for overlay in recipes/overlays/*.yaml; do
       parent=$(yq eval '.spec.base // ""' "$current" 2>/dev/null || true)
       if [[ -z "$parent" || "$parent" == "null" ]]; then break; fi
       if [[ -n "${visited_r3[$parent]:-}" ]]; then
-        echo "::warning::cyclic base chain detected at overlay '${name}' (re-visited '${parent}')"
+        echo "::warning::cyclic base chain detected at overlay '$(log_sanitize "$name")' (re-visited '$(log_sanitize "$parent")')"
         break
       fi
       visited_r3[$parent]=1
@@ -165,6 +266,25 @@ for overlay in recipes/overlays/*.yaml; do
     fi
   fi
 
+  # Rule 5: a changed registry component entry that this recipe
+  # transitively references (component-level cascade scoping, #1435). Only
+  # the recipes that actually reference a changed component are promoted —
+  # not every leaf — so an `aws-efa`-only registry edit no longer flags
+  # non-AWS recipes.
+  if [[ "$include" == "false" && "$registry_changed" == "true" && -n "$changed_components" ]]; then
+    if rc=$(recipe_components "$overlay"); then
+      # Resolved cleanly: include only on a real intersection.
+      if [[ -n "$rc" ]] && comm -12 <(printf '%s\n' "$rc") <(printf '%s\n' "$changed_components") | grep -q .; then
+        include=true
+      fi
+    else
+      # Resolver failed (aicr recipe timeout/error). Fail open: include the
+      # recipe so a transient failure can't hide real drift on a protected one.
+      echo "::warning::component resolution failed for $(log_sanitize "$overlay"); including conservatively"
+      include=true
+    fi
+  fi
+
   if [[ "$include" == "true" ]]; then
     affected=$(echo "$affected" | jq -c --arg r "$name" '. + [$r]')
   fi
@@ -172,6 +292,31 @@ done
 
 count=$(echo "$affected" | jq 'length')
 echo "Affected leaf overlays: ${count}"
+
+# Partition affected recipes into "protected" (a committed pointer exists
+# in recipes/evidence/) and "other" (affected but no evidence yet). The
+# gate actively verifies the protected set; the others are surfaced as a
+# best-effort summary, not per-recipe warnings (#1432).
+#
+# A pointer present in BASE or HEAD counts as protected: a PR that *deletes*
+# a pointer leaves no file at HEAD, but removing evidence from a previously
+# protected recipe is a de-protection we must surface — not silently demote
+# it to "no evidence yet". The protected loop renders such a recipe as
+# "evidence pointer removed" rather than verifying it.
+protected="[]"
+others="[]"
+while IFS= read -r slug; do
+  [[ -z "$slug" ]] && continue
+  if [[ -f "recipes/evidence/${slug}.yaml" ]] \
+     || git cat-file -e "${BASE_SHA}:recipes/evidence/${slug}.yaml" 2>/dev/null; then
+    protected=$(echo "$protected" | jq -c --arg r "$slug" '. + [$r]')
+  else
+    others=$(echo "$others" | jq -c --arg r "$slug" '. + [$r]')
+  fi
+done < <(echo "$affected" | jq -r '.[]')
+protected_count=$(echo "$protected" | jq 'length')
+other_count=$(echo "$others" | jq 'length')
+echo "Protected (have pointers): ${protected_count}; other affected (no evidence yet): ${other_count}"
 
 # Orphan-pointer check. The loop above only iterates existing overlays,
 # so a pointer added/modified for a slug with no matching leaf overlay
@@ -206,9 +351,13 @@ mkdir -p "$(dirname "$REPORT_OUT")"
   echo "## Recipe evidence check"
   echo
   if [[ "$promote_all" == "true" ]]; then
-    echo "> **Broad impact:** \`recipes/registry.yaml\` or \`recipes/overlays/base.yaml\` changed;"
-    echo "> every leaf recipe is potentially affected. The list below covers all of them — each"
-    echo "> one would ideally have refreshed evidence before merge."
+    echo "> **Broad impact:** \`recipes/overlays/base.yaml\` changed; every leaf recipe is"
+    echo "> potentially affected. Recipes that carry committed evidence are verified below;"
+    echo "> the rest have no evidence yet (best-effort)."
+    echo
+  elif [[ "$registry_changed" == "true" && -n "$changed_components" ]]; then
+    echo "> **Registry change:** scoped to recipes that reference a changed component"
+    echo "> entry in \`recipes/registry.yaml\` (not every leaf)."
     echo
   fi
 } >> "$REPORT_OUT"
@@ -225,17 +374,19 @@ if [[ "$count" -eq 0 && "$orphan_count" -eq 0 ]]; then
   exit 0
 fi
 
-if [[ "$count" -gt 0 ]]; then
-  {
-    echo "Affected leaf overlays: **${count}**"
-    echo
-    echo "| Recipe | Pointer | Verify | Digest match |"
-    echo "|---|---|---|---|"
-  } >> "$REPORT_OUT"
-fi
-
 rows_written=0
 rows_truncated=0
+
+if [[ "$protected_count" -gt 0 ]]; then
+  {
+    echo "### Protected recipes"
+    echo
+    echo "Recipes with committed evidence (\`recipes/evidence/<slug>.yaml\`) that this PR affects: **${protected_count}**"
+    echo
+    echo "| Recipe | Verify | Digest match |"
+    echo "|---|---|---|"
+  } >> "$REPORT_OUT"
+fi
 
 # `while IFS= read -r` (not `for x in $(...)`) so slugs with shell
 # metachars don't word-split or glob-expand.
@@ -249,6 +400,16 @@ while IFS= read -r slug; do
   overlay="recipes/overlays/${slug}.yaml"
   pointer="recipes/evidence/${slug}.yaml"
 
+  # Protected via a BASE pointer that this PR deletes (absent at HEAD).
+  # Removing evidence from a previously protected recipe is a de-protection
+  # to surface, not a file to verify.
+  if [[ ! -f "$pointer" ]]; then
+    echo "| \`$(md_escape "$slug")\` | :warning: evidence pointer removed | — |" >> "$REPORT_OUT"
+    warnings=$((warnings + 1))
+    rows_written=$((rows_written + 1))
+    continue
+  fi
+
   # Wall-clock cap on aicr invocations so a hung / tarpit OCI registry
   # behind a PR-controlled pointer URL can't burn the whole job budget.
   # `timeout` exits 124 on timeout; the existing verify-exit default
@@ -258,21 +419,14 @@ while IFS= read -r slug; do
   digest_err=$(mktemp)
   current_digest=""
   if ! current_digest=$(timeout "$AICR_TIMEOUT" "$AICR" evidence digest -r "$overlay" 2>"$digest_err"); then
-    echo "::warning::digest failed for ${overlay}: $(head -c 500 "$digest_err")"
-    echo "| \`${slug}\` | — | — | :warning: could not compute current digest |" >> "$REPORT_OUT"
+    echo "::warning::digest failed for $(log_sanitize "$overlay"): $(log_sanitize "$(head -c 500 "$digest_err")")"
+    echo "| \`$(md_escape "$slug")\` | — | :warning: could not compute current digest |" >> "$REPORT_OUT"
     warnings=$((warnings + 1))
     rows_written=$((rows_written + 1))
     rm -f "$digest_err"
     continue
   fi
   rm -f "$digest_err"
-
-  if [[ ! -f "$pointer" ]]; then
-    echo "| \`${slug}\` | :warning: missing | — | — |" >> "$REPORT_OUT"
-    warnings=$((warnings + 1))
-    rows_written=$((rows_written + 1))
-    continue
-  fi
 
   verify_json=$(mktemp)
   verify_err=$(mktemp)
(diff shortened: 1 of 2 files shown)


njhensley (commented):
## 📋 Review — multi-persona, meta-confirmed

**Method:** four persona passes (shell-correctness, security, CI/DX, domain-correctness), each finding independently confirmed/refuted by a senior meta-reviewer against the resolved code. **Tier legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick.

### Overall — no blockers; two Majors to resolve or defer in writing

Clean, well-commented tuning of a **warning-only, never-blocking** gate, and the noise-reduction direction is right. The privilege split (unprivileged `pull_request` build + `workflow_run` comment with no PR checkout) is implemented correctly — the most important control here, and it's sound. Nothing in this PR can break CI or merge a bad artifact.

Two **Major** items keep it from a clean approve (both inline):
1. The component-scoped cascade walks a *different inheritance graph* than the one the digest is computed over → systematic false-negatives on **registry-only** edits to shared/base/wildcard components (the exact case the feature targets).
2. The classified-cause/pending rendering reads verifier JSON fields that **don't exist yet** → graceful dead code that the ADR nonetheless marks "(implemented)".

### 🟡 Minor (not anchorable inline — file not in this diff)

**Stale "Known limitations" comment in `.github/workflows/recipe-evidence.yaml` L32–35** still says discovery *"walks `spec.base` ancestors but not `spec.mixins`"* and *"promote-all only matches registry.yaml or base.yaml"* — both made false by this PR. The script header was updated; the duplicated workflow copy was not.

`recipe_components()` **has** its own base-chain cycle guard (`seen[]`); removing the `[[ ! -f "$pointer" ]]` branch is safe (non-pointer recipes route to the `others` partition; a vanished pointer fails-closed to `:x: verify error`); `mktemp` files are cleaned on every `continue`; `comm -12 | grep -q` is correct under `pipefail`; the new `protected=`/`other_affected=` stdout keys break no consumer; the truncation row was correctly updated to 3 columns; the lower `MAX_ROWS` header comment **was** updated; trigger-model split is correct.

### Summary

| Tier | Count |
|---|---|
| 🔴 Blocker | 0 |
| 🟠 Major | 2 |
| 🟡 Minor | 6 |
| 🔵 Nitpick | 2 |

**Recommendation:** not a blocker for a warning-only gate, but please either fix the **cascade graph mismatch** (it defeats the feature's core purpose for registry edits) or scope it down + document the gap, and reconcile the **dead-code vs. "implemented" ADR** before merge.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -63,11 +71,81 @@ if ! changed_files=$(git diff --name-only "${BASE_SHA}...${HEAD_SHA}" 2>&1); the
...
+  cc_count=$(printf '%s\n' "$changed_components" | grep -c . || true)
+  echo "Changed registry components: ${cc_count}"
+fi
+
+# recipe_components <overlay-path> — print the sorted, unique set of
+# component names a recipe references, accumulated across its base chain
+# and its mixins. Used by Rule 5 to intersect against changed_components.
+# Over-approximation by design (see header limitations); the digest
+# comparison remains the ground truth for recipes that carry a pointer.
+recipe_components() {
```
njhensley, line 145: 🟠 **Major — `recipe_components()` walks the `spec.base`/`spec.mixins` chain, but the digest resolves by criteria-match across *all* overlays.**

`aicr evidence digest -r <overlay>` resolves via `ComputeRecipeDigest → LoadFromFileWithProvider → BuildFromCriteria`, which re-resolves purely by **criteria matching across every overlay** (`FindMatchingOverlays`) plus an unconditional `base.yaml` merge (`initBaseMergedSpec`) — not the base/mixins graph this function walks.

- No iterated leaf reaches `base.yaml` via `spec.base` (service overlays terminate the chain), yet `base.yaml`'s componentRefs merge into every recipe's digest.
- Wildcard overlays like `monitoring-hpa` (`criteria.intent: any`) inject components (`prometheus-adapter`) by criteria, referenced by no leaf's base/mixins.

**Blast radius:** a registry-only edit to a base/wildcard component (`cert-manager`, `prometheus-adapter`, `kai-scheduler`, `prometheus-operator-crds`, `nvsentinel`) flags **zero** protected recipes via Rule 5, while their digests depend on it — a systematic false-negative for exactly the case this feature targets. Major (not blocker) only because the gate is warning-only and Rules 1–4 still catch drift when an overlay/values file is co-touched.

**Fix:** resolve the set with the same engine the digest uses, e.g.
```bash
recipe_components() {
  "$AICR" recipe -r "$1" --format json 2>/dev/null \
    | jq -r '.spec.componentRefs[].name // empty' | sort -u
}
```
or, at minimum, treat `base.yaml` as an implicit ancestor of every leaf and union in every criteria-matching overlay's componentRefs.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -259,54 +373,74 @@ while IFS= read -r slug; do
...
+  # ErrCodeInvalidRequest (bundle invalid) to OS exit 2, so we disambiguate
+  # via the JSON, not the exit code alone.
   signed_digest=""
+  pending="false"
+  cause_class=""
+  cause_status=""
+  cause_hint=""
   if [[ -s "$verify_json" ]]; then
     signed_digest=$(jq -r '.predicate.recipe.digest // empty' "$verify_json" 2>/dev/null || true)
+    pending=$(jq -r '.pending // false' "$verify_json" 2>/dev/null || echo false)
```
njhensley, line 460: 🟠 **Major — classified-cause + pending rendering reads verifier JSON fields that don't exist yet (graceful dead code).**

`VerifyResult` in `pkg/evidence/verifier/types.go` has no `failureCause` or `pending` field, and this PR changes no `.go` (it's the "#1437 *script half*"). Today: the `:hourglass: pending signature` branch is unreachable, every exit-2 failure renders bare `:x: invalid` (the advertised hint never appears). The jq `// empty`/`// false` defaults degrade gracefully, but ADR-007 documents this under a heading marked **"(implemented)"** — describing behavior the shipped binary can't produce.

**Fix:** land the verifier `failureCause`/`pending` fields in this PR, **or** soften the ADR heading + script header to "surfaced once the verifier emits `failureCause` (#1437 Go half)" and add a comment that these fields are forward-looking.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -206,9 +314,13 @@ mkdir -p "$(dirname "$REPORT_OUT")"
...
   echo
   if [[ "$promote_all" == "true" ]]; then
-    echo "> **Broad impact:** \`recipes/registry.yaml\` or \`recipes/overlays/base.yaml\` changed;"
-    echo "> every leaf recipe is potentially affected. The list below covers all of them — each"
-    echo "> one would ideally have refreshed evidence before merge."
+    echo "> **Broad impact:** \`recipes/overlays/base.yaml\` changed; every leaf recipe is"
+    echo "> potentially affected. Recipes that carry committed evidence are verified below;"
+    echo "> the rest have no evidence yet (best-effort)."
+    echo
+  elif [[ "$registry_changed" == "true" ]]; then
```
njhensley: 🟡 **Minor — registry banner renders even when a cosmetic registry edit affects 0 recipes.**

This `elif registry_changed` banner is emitted before the `count == 0` early-return. On a whitespace/comment-only registry edit (`changed_components` empty), the reader sees "scoped to recipes that reference a changed component entry…" immediately followed by "No leaf overlays affected by this PR."

**Fix:** gate the banner on `registry_changed && -n "$changed_components"` (or on `count > 0`).

.github/scripts/recipe-evidence-check.sh
```diff
@@ -259,54 +373,74 @@ while IFS= read -r slug; do
...
+        # Render the classified, actionable cause instead of a bare
+        # "invalid" (#1437): e.g. "invalid — registry-forbidden (HTTP 403):
+        # make the fork's aicr-evidence package public".
+        verify_cell=":x: invalid"
+        if [[ -n "$cause_class" ]]; then
+          verify_cell="${verify_cell} — ${cause_class}"
+          [[ -n "$cause_status" ]] && verify_cell="${verify_cell} (HTTP ${cause_status})"
+          [[ -n "$cause_hint" ]] && verify_cell="${verify_cell}: ${cause_hint}"
+        fi
+        warnings=$((warnings + 1))
```
njhensley, line 503: 🟡 **Minor — an invalid bundle increments `warnings` twice.**

An exit-2 invalid verify (empty `signed_digest`) does `warnings++` here, then reaches the digest section and hits the `else` (no signed digest) for a second `warnings++` — so one broken recipe contributes 2 warnings vs 1 for a digest failure. Cosmetic (no consumer parses `warnings=`), but inconsistent.

**Fix:** drop this increment and let the digest-skip arm own the single warning.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -322,13 +456,35 @@ while IFS= read -r slug; do
     warnings=$((warnings + 1))
   fi
 
-  echo "| \`${slug}\` | :white_check_mark: present | ${verify_cell} | ${digest_cell} |" >> "$REPORT_OUT"
+  echo "| \`${slug}\` | ${verify_cell} | ${digest_cell} |" >> "$REPORT_OUT"
```
njhensley: 🟡 **Minor — PR-controlled `slug`/`cause_hint` rendered unescaped into a comment posted verbatim.**

`recipe-evidence-comment.yaml` posts `report.md` verbatim under a `pull-requests: write` identity. `slug` (PR-added overlay filename) and `cause_hint` (PR-controlled OCI artifact, once #1437 lands) flow into table cells / `<details>` with no escaping — a crafted value with `|`, a newline, `</details>`, or `](http://…)` can break layout or inject a link. Bounded to defacing a bot comment (no privilege gain).

**Fix:** pipe-escape and strip newlines from cells (`tr -d '\n\r' | sed 's/|/\\|/g'`) and length-cap `cause_hint`.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -63,11 +71,81 @@ if ! changed_files=$(git diff --name-only "${BASE_SHA}...${HEAD_SHA}" 2>&1); the
...
+  # empty doc, so every HEAD component reads as added — fail-open to broad
+  # for that recipe set, which is the safe direction for a warning gate.
+  git show "${BASE_SHA}:recipes/registry.yaml" >"$reg_base" 2>/dev/null || : >"$reg_base"
+  git show "${HEAD_SHA}:recipes/registry.yaml" >"$reg_head" 2>/dev/null || : >"$reg_head"
+  comp_names=$( { yq eval '.components[].name // ""' "$reg_base" 2>/dev/null || true; \
+                  yq eval '.components[].name // ""' "$reg_head" 2>/dev/null || true; } \
+                | grep -v '^$' | sort -u )
+  while IFS= read -r cn; do
+    [[ -z "$cn" ]] && continue
+    b=$(yq eval ".components[] | select(.name == \"${cn}\")" "$reg_base" 2>/dev/null || true)
```
njhensley: 🟡 **Minor — registry component name string-interpolated into a yq expression (fails-open / injection-shaped).**

`select(.name == \"${cn}\")` splices a PR-controlled component name into the yq filter. A name containing `\"` breaks the expression; `|| true` swallows it and the component reads as *unchanged* — a real change silently dropped from the cascade. Expression-injection-shaped (yq evaluates expressions, not shell; runs in the secretless `pull_request` job — no RCE). Clean today, so latent.

**Fix:** pass via env, mirroring the comment workflow's jq-safety pattern: `cn="$cn" yq eval '.components[] | select(.name == strenv(cn))' "$reg_base"`.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -322,13 +456,35 @@ while IFS= read -r slug; do
...
+    echo
+    echo "<details><summary>Other affected recipes without evidence yet: <b>${other_count}</b></summary>"
+    echo
+    echo "These recipes are affected by this PR but carry no committed evidence pointer, so there is"
+    echo "nothing to verify. This is expected — evidence is hardware-gated and added over time."
+    echo
+  } >> "$REPORT_OUT"
+  while IFS= read -r oslug; do
+    [[ -z "$oslug" ]] && continue
+    echo "- \`${oslug}\`" >> "$REPORT_OUT"
```
njhensley: 🟡 **Minor — "other affected" list is uncapped against the ~65KB comment limit.**

`MAX_ROWS` now caps only the protected table; this loop emits one bullet per recipe uncapped. At ~91 leaves that's ~2.5KB (scaling-only risk), but the header comment claims `MAX_ROWS` keeps "the report under GitHub's ~65KB cap," which it no longer fully guarantees.

**Fix:** apply the same cap here, or reword the comment to "bounds the protected table; the others list is unbounded but compact".

.github/scripts/recipe-evidence-check.sh
```diff
@@ -259,54 +373,74 @@ while IFS= read -r slug; do
...
   verify_err=$(mktemp)
   set +e
   timeout "$AICR_TIMEOUT" "$AICR" evidence verify "$pointer" --format json >"$verify_json" 2>"$verify_err"
   verify_exit=$?
   set -e
   if [[ "$verify_exit" -ne 0 && -s "$verify_err" ]]; then
-    echo "::warning::verify exit ${verify_exit} for ${pointer}: $(head -c 500 "$verify_err")"
+    # Echo the full stderr into the job log (not just the table) so a
+    # maintainer can see the underlying cause beyond the classified row.
+    echo "::warning::verify exit ${verify_exit} for ${pointer}: $(head -c 1000 "$verify_err")"
```
njhensley: 🔵 **Nitpick — PR-controlled stderr echoed into `::warning::` workflow commands.**

`aicr` stderr (influenceable via PR overlays / the OCI artifact) is echoed after `::warning::`; this PR widens the slice 500→1000 bytes. Worst case is cosmetic log-annotation spoofing in an unprivileged job.

**Fix (optional):** strip `%`/CR/LF/`::` before echoing, or route the raw cause through a non-command `echo`.

docs/design/007-recipe-evidence.md
```diff
@@ -376,6 +376,41 @@ surface item 1 below for the full schema).
...
+- **Component-scoped registry cascade.** A change to
+  `recipes/registry.yaml` does not promote every leaf recipe. The gate
+  diffs the registry at the component-*entry* level and marks a recipe
+  affected only if its resolved component set — walked across the base
+  chain and `spec.mixins` — intersects the changed entries, so an
+  `aws-efa`-only edit flags only recipes that reference `aws-efa`. A change
+  to `recipes/overlays/base.yaml` is still treated as broad (it sits at the
+  root of nearly every base chain). The recomputed digest remains the
+  ground truth: a protected recipe whose digest still matches its pointer
+  is reported as `matches`, never as drift, regardless of promotion.
```
njhensley: 🔵 **Nitpick — ADR over-states the digest cell.**

"is reported as `matches`, never as drift, regardless of promotion" is true for the match case, but the script also has `:warning: skipped` and `:warning: could not compute current digest` non-drift outcomes. Soften to "…is reported as a match, never as drift".

.github/scripts/recipe-evidence-check.sh
```diff
@@ -173,6 +263,24 @@ done
...
+protected="[]"
+others="[]"
+while IFS= read -r slug; do
+  [[ -z "$slug" ]] && continue
+  if [[ -f "recipes/evidence/${slug}.yaml" ]]; then
+    protected=$(echo "$protected" | jq -c --arg r "$slug" '. + [$r]')
+  else
+    others=$(echo "$others" | jq -c --arg r "$slug" '. + [$r]')
+  fi
+done < <(echo "$affected" | jq -r '.[]')
```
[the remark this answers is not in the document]
mchmarny: Fixed — the partition now treats a pointer present in **BASE or HEAD** as protected. A PR that deletes `recipes/evidence/<slug>.yaml` keeps the recipe in the Protected table, rendered as `:warning: evidence pointer removed` (a warning) rather than silently demoted to "other / no evidence yet". Smoke-tested: deleting the gb200 pointer yields warnings=1 with that row.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -259,54 +373,74 @@ while IFS= read -r slug; do
...
+          verify_cell="${verify_cell} — ${cause_class}"
+          [[ -n "$cause_status" ]] && verify_cell="${verify_cell} (HTTP ${cause_status})"
+          [[ -n "$cause_hint" ]] && verify_cell="${verify_cell}: ${cause_hint}"
+        fi
+        warnings=$((warnings + 1))
       fi
       ;;
     *)
       verify_cell=":x: verify error (exit ${verify_exit})"
+      warnings=$((warnings + 1))
```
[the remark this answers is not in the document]
mchmarny: Fixed — the gate now branches on the JSON `.exit` field from `evidence verify --format json` (0 valid / 1 valid-with-phase-failures / 2 invalid), not the OS exit code. Confirmed via pkg/errors/exitcode.go that aicr collapses both ErrCodeConflict (Exit 1) and ErrCodeInvalidRequest (Exit 2) to OS exit 2, so exit 1 is now rendered as "phase failures recorded (informational)" instead of a generic verify error. A missing/unparseable result still falls back to "verify error (exit N)".

.github/scripts/recipe-evidence-check.sh
mchmarny: Fixed — `recipe_components` now resolves via `aicr recipe --<criteria> --format json | jq ".componentRefs[].name"`, the same engine `aicr evidence digest` uses (criteria match + unconditional base.yaml merge + wildcard overlays). Verified: a `cert-manager` registry edit now flags all 63 recipes (the old hand-walk flagged 0); `aws-efa` still scopes to the 23 EKS recipes. Dropped the base/mixins hand-walk and its limitation note.

.github/scripts/recipe-evidence-check.sh
mchmarny: Not dead code — `VerifyResult.failureCause` and `.pending` shipped in #1445, which is merged to main (this branch is based on it; see pkg/evidence/verifier/types.go on the branch). The pending and classified-cause branches are live against the shipped binary, so the ADR "(implemented)" heading is accurate. No change needed.

.github/scripts/recipe-evidence-check.sh
mchmarny: Fixed — the registry banner is now gated on `registry_changed && -n "$changed_components"`, so a cosmetic registry edit (0 changed components) no longer prints a "scoped to…" banner immediately above "No leaf overlays affected".

.github/scripts/recipe-evidence-check.sh
mchmarny: Fixed — added a `verify_ok` flag; the digest "skipped (no signed digest)" arm now only increments `warnings` when the bundle otherwise verified, so an invalid bundle contributes exactly one warning instead of two.

.github/scripts/recipe-evidence-check.sh
mchmarny: Fixed — added `md_escape` (collapses newlines, escapes pipes, drops backticks) applied to `slug` and `cause_hint` (and orphan slugs) wherever they render into a Markdown table cell.

.github/scripts/recipe-evidence-check.sh
mchmarny: Fixed — the registry component name is now passed to yq via `env(CN)` instead of being string-interpolated into the expression, so a PR-controlled component name is treated as data, not code.

.github/scripts/recipe-evidence-check.sh
mchmarny: Fixed — the collapsed "other affected" list is now capped at `OTHER_MAX` (default 200) with a "+N more (list capped)" line; the summary count remains the full total.

.github/scripts/recipe-evidence-check.sh
mchmarny: Fixed — PR-controlled stderr echoed into `::warning::` is now run through `log_sanitize` (collapses newlines, neutralizes `::`) so it cannot inject its own workflow commands.

docs/design/007-recipe-evidence.md
mchmarny: Fixed — softened to "reported as a match, never as drift" since the digest cell also has skipped / could-not-compute outcomes.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -63,11 +84,81 @@ if ! changed_files=$(git diff --name-only "${BASE_SHA}...${HEAD_SHA}" 2>&1); the
...
+  local args=()
+  [[ -n "$service" && "$service" != "null" ]] && args+=(--service "$service")
+  [[ -n "$accel" && "$accel" != "null" ]] && args+=(--accelerator "$accel")
+  [[ -n "$os" && "$os" != "null" ]] && args+=(--os "$os")
+  [[ -n "$intent" && "$intent" != "null" ]] && args+=(--intent "$intent")
+  [[ -n "$platform" && "$platform" != "null" ]] && args+=(--platform "$platform")
+
+  : "${AICR_TIMEOUT:=30s}"
+  timeout "$AICR_TIMEOUT" "$AICR" recipe "${args[@]}" --format json 2>/dev/null \
+    | jq -r '.componentRefs[].name // empty' 2>/dev/null | sort -u || true
```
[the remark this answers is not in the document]
mchmarny: Fixed — `recipe_components` now returns non-zero when `aicr recipe` itself fails (timeout/error), and Rule 5 fails **open** on that: it includes the recipe (with a `::warning::`) rather than treating a resolver hiccup as "references no changed component". A clean resolution still includes only on a real intersection.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -63,11 +84,88 @@ if ! changed_files=$(git diff --name-only "${BASE_SHA}...${HEAD_SHA}" 2>&1); the
...
+
+  local args=()
+  [[ -n "$service" && "$service" != "null" ]] && args+=(--service "$service")
+  [[ -n "$accel" && "$accel" != "null" ]] && args+=(--accelerator "$accel")
+  [[ -n "$os" && "$os" != "null" ]] && args+=(--os "$os")
+  [[ -n "$intent" && "$intent" != "null" ]] && args+=(--intent "$intent")
+  [[ -n "$platform" && "$platform" != "null" ]] && args+=(--platform "$platform")
+
+  : "${AICR_TIMEOUT:=30s}"
+  # Return non-zero when the resolver itself fails (timeout / error), so the
```
njhensley, line 161: ✅ Nice — resolving the affected set through the same engine the digest uses (`aicr recipe … --format json`) closes the base/wildcard false-negative cleanly. Verified the jq path: `RecipeResult.ComponentRefs` carries json tag `componentRefs` at top level (`pkg/recipe/metadata.go:415`), so `.componentRefs[].name` is correct.

🔵 Non-blocking note: this now execs `aicr recipe` once per candidate overlay on a registry change (~60 resolutions on a broad registry edit). Fine for a warning gate, but worth confirming it's acceptable or memoizing if registry edits become frequent.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -258,57 +416,96 @@ while IFS= read -r slug; do
...
   rm -f "$verify_err"
 
-  # aicr maps both ErrCodeConflict (phase failures recorded; bundle
-  # valid) and ErrCodeInvalidRequest (bundle invalid) to OS exit 2 —
-  # see pkg/errors/exitcode.go. Disambiguate via whether the JSON
-  # payload contains a parseable predicate.
+  # Pull structured fields surfaced by `aicr evidence verify --format json`
+  # (see pkg/evidence/verifier): VerifyResult.Exit (0 valid, 1 valid with
+  # recorded phase failures, 2 invalid), the predicate digest, the
+  # pending-signature flag, and — on Exit 2 — the classified failureCause
```
njhensley, line 447: 🟠 Follow-up (non-blocking): `pending` and `failureCause.{class,httpStatus,hint}` still don't exist on `VerifyResult` (`pkg/evidence/verifier/types.go:108-118` — only `.exit` landed). So the `:hourglass: pending signature` branch and the classified-cause rendering remain inert until the #1437 Go half lands. It degrades gracefully (→ "passed" / bare ":x: invalid"), so no regression — switching to the real `.exit` field was a genuine improvement — but the headline feature isn't observable yet.

docs/design/007-recipe-evidence.md
```diff
@@ -376,6 +376,41 @@ surface item 1 below for the full schema).
...
+  diffs the registry at the component-*entry* level and marks a recipe
+  affected only if its resolved component set — walked across the base
+  chain and `spec.mixins` — intersects the changed entries, so an
+  `aws-efa`-only edit flags only recipes that reference `aws-efa`. A change
+  to `recipes/overlays/base.yaml` is still treated as broad (it sits at the
+  root of nearly every base chain). The recomputed digest remains the
+  ground truth: a protected recipe whose digest still matches its pointer
+  is reported as a match, never as drift, regardless of promotion.
+
+- **Classified verify failures.** When `aicr evidence verify` fails, the
```
njhensley, line 407: 🟡 Since `failureCause` isn't emitted by the verifier yet (see the script thread), this subsection describes behavior the shipped binary can't produce, under a heading marked "(implemented)". Either land the verifier fields in this PR, or soften to "surfaced once the verifier emits `failureCause` (#1437 Go half)".

njhensley (dismissed):
## ✅ Approve with notes

Re-reviewed the force-push (`e3dad1f` → `1894bfc`) and verified each fix against the code. **8 of 10 findings fixed, both Majors addressed substantively** — approving. The two Majors in particular:

- **Cascade graph mismatch** — fully fixed: `recipe_components()` now resolves via `aicr recipe --format json` (the same criteria-resolution engine the digest uses); jq path confirmed correct against `RecipeResult` (`pkg/recipe/metadata.go:415`). The base/wildcard false-negative is closed.
- **Classified causes / pending** — improved (now branches on the real `VerifyResult.Exit` rather than OS-exit heuristics, no regression), but the `pending`/`failureCause` fields still aren't emitted by the verifier, so that rendering stays inert until the #1437 Go half. See inline.

Also confirmed fixed: cosmetic-edit banner guard, double-counted warnings (`verify_ok`), Markdown escaping (`md_escape`), yq `env(CN)`, `OTHER_MAX` cap, `log_sanitize` on `::warning::` echoes, ADR "matches" wording. **Bonus:** the de-protection handling (BASE-pointer present, deleted at HEAD → `:warning: evidence pointer removed`) is a good catch beyond the review.

### Two non-blocking items remain (fine as follow-ups)
1. **🟠 ADR overstates classified causes as "(implemented)"** while the verifier doesn't emit `failureCause`/`pending` yet — land the Go fields or soften the wording (inline).
2. **🟡 Stale comment** in `.github/workflows/recipe-evidence.yaml:32-33` still says discovery "walks `spec.base` ancestors but not `spec.mixins`" and "promote-all only matches registry.yaml" — both now false (resolution is via `aicr`; registry is component-scoped). The script's own header note was correctly removed; this duplicated copy was missed.

None of these block a warning-only gate. Thanks for the thorough turnaround.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -63,11 +84,88 @@ if ! changed_files=$(git diff --name-only "${BASE_SHA}...${HEAD_SHA}" 2>&1); the
...
+  git show "${HEAD_SHA}:recipes/registry.yaml" >"$reg_head" 2>/dev/null || : >"$reg_head"
+  comp_names=$( { yq eval '.components[].name // ""' "$reg_base" 2>/dev/null || true; \
+                  yq eval '.components[].name // ""' "$reg_head" 2>/dev/null || true; } \
+                | grep -v '^$' | sort -u )
+  while IFS= read -r cn; do
+    [[ -z "$cn" ]] && continue
+    # Pass the name via env() rather than string-interpolating it into the
+    # yq expression: a PR-controlled component name must be data, not code.
+    b=$(CN="$cn" yq eval '.components[] | select(.name == env(CN))' "$reg_base" 2>/dev/null || true)
+    h=$(CN="$cn" yq eval '.components[] | select(.name == env(CN))' "$reg_head" 2>/dev/null || true)
```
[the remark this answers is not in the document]
mchmarny: Fixed — switched the component-name match to `strenv(CN)` so a scalar-looking name (`true`/`null`/`123`) stays a string and matches its registry entry.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -165,6 +263,25 @@ for overlay in recipes/overlays/*.yaml; do
...
+  if [[ "$include" == "false" && "$registry_changed" == "true" && -n "$changed_components" ]]; then
+    if rc=$(recipe_components "$overlay"); then
+      # Resolved cleanly: include only on a real intersection.
+      if [[ -n "$rc" ]] && comm -12 <(printf '%s\n' "$rc") <(printf '%s\n' "$changed_components") | grep -q .; then
+        include=true
+      fi
+    else
+      # Resolver failed (aicr recipe timeout/error). Fail open: include the
+      # recipe so a transient failure can't hide real drift on a protected one.
+      echo "::warning::component resolution failed for ${overlay}; including conservatively"
```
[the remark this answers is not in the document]
mchmarny: Fixed — the resolver-failure warning now routes `$overlay` through `log_sanitize` before emitting the `::warning::`, consistent with the other workflow-command echoes.

.github/scripts/recipe-evidence-check.sh
mchmarny: Acknowledged — confirmed `.componentRefs[].name` is correct. The ~60 `aicr recipe` resolutions only run on a registry-entry change, for leaves not already caught by Rules 1–4, each guarded by `AICR_TIMEOUT`; well within budget for a warning gate. Memoizing by criteria-tuple is a clean future optimization if registry edits get hot — leaving it out for now to keep the script simple.

.github/scripts/recipe-evidence-check.sh
mchmarny: These fields do exist on the base branch: `Pending` and `FailureCause` shipped in #1445 (merged to main — `e8781612`), which this PR branches from. On this branch `pkg/evidence/verifier/types.go` declares `Pending` (`json:"pending"`) and `FailureCause` (`json:"failureCause"`), and the verifier populates them (`verify.go` sets `r.Pending = true`; `failure.go`/`setFailureCause` set `FailureCause`). So the pending and classified-cause rendering is live against the shipped binary — not inert. The PR diff doesn’t touch types.go precisely because the fields already landed in #1445.

docs/design/007-recipe-evidence.md
mchmarny: The verifier does emit `failureCause` (and `pending`) — they shipped in #1445 (merged), which this branch is based on; see `pkg/evidence/verifier/types.go` (`Pending`/`FailureCause` fields) and `failure.go` (classifier). So the "(implemented)" heading is accurate. No change needed.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -54,6 +57,24 @@ set -euo pipefail
...
+md_escape() {
+  printf '%s' "$1" | tr '\n' ' ' | sed 's/|/\\|/g; s/`//g'
+}
+
+# log_sanitize strips characters from PR-controlled text before it is echoed
+# into a `::warning::`/`::error::` workflow command, so the text cannot inject
+# its own workflow commands. Collapses newlines and neutralizes `::`.
+log_sanitize() {
+  printf '%s' "$1" | tr '\n' ' ' | sed 's/::/:_:/g'
+}
```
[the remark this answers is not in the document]
mchmarny: Fixed — `${overlay}`, `${pointer}`, and the cyclic-base-chain `${name}`/`${parent}` are now all routed through `log_sanitize` before being written into `::warning::` messages (the only remaining raw echo is the git-diff `::error::`, which carries no PR-controlled value). Also updated `log_sanitize` to collapse `\r` alongside `\n`.

.github/scripts/recipe-evidence-check.sh
```diff
@@ -165,6 +266,25 @@ for overlay in recipes/overlays/*.yaml; do
...
+  # not every leaf — so an `aws-efa`-only registry edit no longer flags
+  # non-AWS recipes.
+  if [[ "$include" == "false" && "$registry_changed" == "true" && -n "$changed_components" ]]; then
+    if rc=$(recipe_components "$overlay"); then
+      # Resolved cleanly: include only on a real intersection.
+      if [[ -n "$rc" ]] && comm -12 <(printf '%s\n' "$rc") <(printf '%s\n' "$changed_components") | grep -q .; then
+        include=true
+      fi
+    else
+      # Resolver failed (aicr recipe timeout/error). Fail open: include the
```
njhensley, line 281: ✅ This is the right call — failing **open** (include the recipe + emit a `::warning::`) when `aicr recipe` times out or errors means a transient resolver hiccup can't silently hide real drift on a protected recipe. Fail-open is the correct direction for a protection gate. Verified the control flow: `if rc=$(recipe_components …)` captures the non-zero return cleanly under `set -e`, and an empty-but-successful resolve still falls through to "no intersection → not included".

.github/scripts/recipe-evidence-check.sh
```diff
@@ -63,11 +85,90 @@ if ! changed_files=$(git diff --name-only "${BASE_SHA}...${HEAD_SHA}" 2>&1); the
...
+  comp_names=$( { yq eval '.components[].name // ""' "$reg_base" 2>/dev/null || true; \
+                  yq eval '.components[].name // ""' "$reg_head" 2>/dev/null || true; } \
+                | grep -v '^$' | sort -u )
+  while IFS= read -r cn; do
+    [[ -z "$cn" ]] && continue
+    # Pass the name via strenv() rather than string-interpolating it into the
+    # yq expression: a PR-controlled component name must be data, not code.
+    # strenv (not env) keeps it a string so a scalar-looking name (`true`,
+    # `null`, `123`) isn't YAML-typed and made to miss its registry entry.
+    b=$(CN="$cn" yq eval '.components[] | select(.name == strenv(CN))' "$reg_base" 2>/dev/null || true)
```
njhensley, line 124: ✅ `strenv(CN)` over `env(CN)` is a nice catch beyond the original note — keeps a scalar-looking component name (`true`, `null`, `123`) typed as a string so it can't be YAML-coerced and miss its registry entry.

njhensley (approved):
## ✅ Approve

Re-reviewed `1894bfc` → `fb91a2d`. Script-only delta, and every change is hardening with no regressions (`bash -n` clean):

- `log_sanitize` now strips `\r` as well as `\n` — fully closes the `::warning::` injection nitpick.
- `strenv(CN)` instead of `env(CN)` — fixes a scalar-typing edge case beyond the original ask (inline).
- `recipe_components()` now **fails open** on resolver timeout/error (include + warn) instead of treating a hiccup as "references no changed component" — correct direction for a protection gate, and it neatly resolves the perf/robustness note from my last pass (inline).
- `log_sanitize` applied consistently to `overlay`/`name`/`parent`/`pointer` in the remaining `::warning::` echoes.

Approving. The two non-blocking follow-ups from the prior review remain open and are fine to land separately: the ADR's "Classified verify failures" still sits under "(implemented)" while the verifier doesn't emit `failureCause`/`pending` yet, and the stale comment in `.github/workflows/recipe-evidence.yaml:32-33`. Neither blocks a warning-only gate. Nice iteration.

Issue #1139, opened 2026-06-01
agentgateway inference-gateway provisions a public LoadBalancer open to 0.0.0.0/0


yuanchen8911:
## Summary

The `agentgateway` component's `inference-gateway` Gateway is materialized by the controller into a `type: LoadBalancer` Service with no source-range restriction. On every cloud this provisions an **internet-facing load balancer open to `0.0.0.0/0`**, exposing the (plaintext HTTP, unauthenticated) inference endpoint to the public internet.

## Impact

A security scan flagged this on an AICR-provisioned EKS cluster. Any inference recipe deployment got a publicly reachable gateway by default, with no first-class way to restrict it from a recipe.

## Resolution

Fixed by #1138 — adds `agentgateway.allowedSourceRanges`, rendered into the generated Service's `spec.loadBalancerSourceRanges` (honored by the AWS, GCP, Azure, and OCI cloud load balancers), so operators scope the gateway via a recipe overlay / `componentRef` override.

Live remediation already applied to the affected clusters:
- EKS `aicr2` (H100) and `b40` — scoped to the NV HQ range (`216.228.127.128/30`); ELB security groups confirmed admitting only that CIDR.
- GB200 EKS — pre-scoped to the same range (its gateway LoadBalancer had not provisioned, so no live public surface).

## Notes / follow-ups

- `allowedSourceRanges` filters by source IP only; it does **not** add TLS or authentication to the gateway listener (separate hardening follow-up).
- Default remains empty (non-breaking); operators opt into scoping. A secure-by-default `ClusterIP` alternative is documented in the PR for future consideration.

yuanchen8911:
Resolved by #1138 (merged as cfb0cb06). Live clusters remediated as noted above. Closing as completed.

PR #1450, opened 2026-06-24
feat(bundler): private-by-default agentgateway inference-gateway exposure

### LINKED ISSUE #1373: feat(bundler): private-by-default agentgateway exposure (deny public, open via --set-json)
## Summary

The `agentgateway` inference-gateway currently defaults **open to the internet**: when `allowedSourceRanges` is unset/empty, no `loadBalancerSourceRanges` is rendered and the LoadBalancer accepts `0.0.0.0/0`. `aicr bundle` only prints a non-blocking **warning**. Apply least-privilege: make the gateway **private by default** — never emit a publicly-open gateway unless the operator explicitly opts in.

## Motivation

An inference endpoint open to the whole internet by default is a real exposure (flagged on multiple clusters). The endpoint is plaintext HTTP and unauthenticated. Defaulting to private aligns with the project's existing "fail closed, don't `Warn; continue`" principle, while keeping bundles deployable.

## Proposed behavior

- When `agentgateway` is enabled and `allowedSourceRanges` is empty/unset, `aicr bundle` **injects the private RFC1918 ranges** (`10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`) into the generated Service's `loadBalancerSourceRanges` and records a bundle note. The deployed gateway is reachable from inside the cluster/VPC but **denied to the public internet** — it is never emitted open to `0.0.0.0/0`.
- Opening to specific clients (e.g. a corporate VPN, which egresses from a **public** IP not covered by the RFC1918 default) is via the existing list-aware flag: `--set-json agentgateway:allowedSourceRanges='["<cidr>"]'` (or a recipe `componentRef` override).
- **Explicit public opt-in** is still allowed but must be deliberate and is logged loudly: `--set-json agentgateway:allowedSourceRanges='["0.0.0.0/0"]'`.
- Validate CIDRs so a bundle that generates does not later fail Kubernetes 1.36+ strict `loadBalancerSourceRanges` validation at Service apply time: require **canonical** form (no host bits, e.g. reject `1.2.3.4/24`; no leading-zero prefix, e.g. reject `1.2.3.0/024`) and reject **IPv4-mapped IPv6** prefixes (e.g. `::ffff:192.12.2.0/120`). Continue to reject the bare-string `--set` footgun.

## Why private-by-default (RFC1918) rather than fail-closed-error or deny-all

Kubernetes has **no native deny-all** for a LoadBalancer — an empty `loadBalancerSourceRanges` means allow-all. So "deny by default" must be expressed as a real list. Three options were considered:

| Option | No-flag behavior | Trade-off |
|---|---|---|
| Fail-closed (error) | bundle refuses to generate | breaks existing inference pipelines; nothing deployable |
| Deny-all sentinel (`255.255.255.255/32`) | deployed but reachable by nobody | gateway inert until scoped; "too limited" |
| **RFC1918 private (chosen)** | deployed, private-only | denies public internet but not in-VPC; corp VPN still needs the flag |

RFC1918 is "secure but not too limited": no public exposure by default, gateway immediately usable inside the cluster/VPC, and `validate --phase deployment`/conformance pass with no flags. It's generic (not a customer-specific network), so it doesn't firewall every deployment to one site. An internal-LB-by-default scheme (per-cloud annotations) is a stronger but larger cross-cloud change — proposed as a follow-up `--internal` option, not this issue.

> Note: the chosen default is a named constant, so flipping to a deny-all sentinel later is a one-line change.

## Compatibility

Default change (open → private). No-flag inference bundles still generate and deploy (now private instead of public), so existing pipelines are not broken. Gate behind a release note.

## Implementation

Implemented in PR #1450. The default is applied at bundle time by `resolveAgentgatewayExposure` (a named constant `agentgatewayDefaultSourceRanges`), with canonical + 4-in-6 CIDR validation in `netutil.IsValidCIDR`. Verified end-to-end on a live EKS cluster: both the no-flag (private RFC1918) and `--set-json` (scoped) deployments pass `validate --phase deployment` and `--phase conformance` (including the `inference-gateway` conformance check).

## Type of Change

- New feature / security hardening

yuanchen8911:
## Summary

Make `aicr bundle` **private by default** for the agentgateway inference-gateway: when `agentgateway` is included and `allowedSourceRanges` is empty/unset, the bundler injects the private RFC1918 ranges into `loadBalancerSourceRanges` instead of emitting a LoadBalancer open to `0.0.0.0/0`. The deployed gateway is reachable inside the cluster/VPC but denied to the public internet.

## Motivation / Context

The inference-gateway defaulted **open to the internet** — an unset `allowedSourceRanges` rendered no `loadBalancerSourceRanges`, and Kubernetes treats that as allow-all. `aicr bundle` only printed a non-blocking warning. An unauthenticated, plaintext-HTTP inference endpoint open to the world is a real exposure (flagged on multiple clusters).

Fixes: #1373
Related: #1139, #1160, #1161

## Type of Change

- [x] New feature (security hardening)
- [x] Documentation update
- [x] Behavior change (open → private default; bundles still generate)

## Component(s) Affected

- [x] Core libraries — `pkg/netutil`
- [x] Docs/examples (`docs/`)

## Implementation Notes

`resolveAgentgatewayExposure` classifies `agentgateway.allowedSourceRanges` and resolves it **default-then-validate**:

- **Unset/empty** → injects `agentgatewayDefaultSourceRanges` (RFC1918: `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16`) into the merged values + bundle note. Denies the public internet; keeps in-VPC reachability. Generation succeeds, so no-flag inference pipelines keep working.
- **Scoped** (valid canonical CIDRs, no any-source) → passes silently.
- **Open** (explicit `0.0.0.0/0` / `::/0`) → allowed, logged loudly + bundle warning (deliberate opt-in).
- **Invalid** (bare-string `--set`, non-list, unparseable, or non-canonical CIDR) → `ErrCodeInvalidRequest`.

`netutil.IsValidCIDR` now uses `netip.ParsePrefix` and requires canonical form (`p == p.Masked()`) and rejects IPv4-mapped IPv6 prefixes (`p.Addr().Is4In6()`), so a CIDR like `1.2.3.4/24`, `1.2.3.0/024`, or `::ffff:192.12.2.0/120` is rejected at bundle time rather than failing Kubernetes 1.36+ strict CIDR validation at Service apply.

The default is a named constant, so switching to a deny-all sentinel (`255.255.255.255/32`) later is a one-line change. An internal-LB default (per-cloud annotations) is a possible follow-up (`--internal`), per #1373.

**Design choice (private RFC1918 vs fail-closed vs deny-all):** Kubernetes has no native deny-all for a LoadBalancer (empty = allow-all), so the default must be a real list. RFC1918 is "secure but not too limited" — no public exposure, gateway usable inside the VPC, and `validate`/conformance pass with no flags — without firewalling every deployment to one customer network. Note: the corporate VPN egresses from a **public** IP (`216.228.127.128/30`), so it is *not* covered by the RFC1918 default and still requires the flag — that's the intended private-vs-public separation.

**Resolves the CI-breakage findings from cross-review:** because no-flag bundles now generate (private) rather than error, `runtime-install/generate-bundle.sh` and the kwok inference lanes are unaffected — no per-workflow `--set-json` needed.

## Testing

```bash
go test ./pkg/bundler/ ./pkg/netutil/...   # pass
golangci-lint run -c .golangci.yaml ./pkg/bundler/... ./pkg/netutil/...   # 0 issues
```

Manual end-to-end with a freshly built binary on an `eks/h100/ubuntu/inference/dynamo` recipe:
- no flag → generates; renders RFC1918 `10.0.0.0/8`,`172.16.0.0/12`,`192.168.0.0/16` into the inference-gateway Service; bundle note recorded
- `--set-json …='["216.228.127.128/30"]'` → renders only the corp CIDR
- non-canonical / invalid CIDR → errors
- explicit `0.0.0.0/0` → generates with a loud warning

Verified on EKS cluster `aicr3`: redeploying `agentgateway-post` from a scoped bundle correctly reconciles `loadBalancerSourceRanges` onto the live `inference-gateway` Service (sentinel-swap test confirmed the field is live-controlled).

Updated chainsaw `bundle-templates/agentgateway` to assert the private-by-default render — all source-range checks match the rendered `loadBalancerSourceRanges` list entries (not the surrounding comments), confirming RFC1918 present and `0.0.0.0/0` absent. `ai-conformance/offline` and `uat/{aws,gcp,azure}/cuj2-inference` now exercise the default path (the redundant scoped `--set-json` was removed). Docs updated: `component-catalog.md`, `cli-reference.md`, `bundling.md`, plus the inline comments in `recipes/components/agentgateway/values.yaml` and `manifests/inference-gateway.yaml`.

Coverage: `netutil.IsValidCIDR` covered in `cidr_test.go` (incl. non-canonical rejection); `resolveAgentgatewayExposure`/`classifyAgentgatewaySourceRanges` covered by the rewritten `TestResolveAgentgatewayExposure` table (default-injection, scoped, any-source, invalid, non-canonical).

## Risk Assessment

- [x] **Medium** — Changes the default exposure for inference bundles (open → private), but bundles still generate/deploy and the change is isolated to one bundle-time step.

**Rollout notes:** No-flag inference bundles now deploy a private gateway instead of a public one. Operators who need public/VPN access pass `--set-json agentgateway:allowedSourceRanges='["<cidr>"]'` (corp VPN: `216.228.127.128/30`) or `'["0.0.0.0/0"]'` for deliberate public exposure.

**Release note (behavior change — inference-gateway open → private):** The bundled `agentgateway` inference-gateway is now **private by default**. Operators who currently reach the endpoint from a **public** IP (corporate VPN egress, a public client) will be **blocked on upgrade** until they opt back in. Remediation: regenerate the bundle with `--set-json agentgateway:allowedSourceRanges='["<your-cidr>"]'` (or `'["0.0.0.0/0"]'` to restore the previous open behavior). This must be called out in the release notes.

## Checklist

- [x] Tests pass locally (affected packages, `-race` via `make test` scope)
- [x] Linter passes (affected packages)


tests/chainsaw/bundle-templates/agentgateway/chainsaw-test.yaml
```diff
@@ -57,32 +62,20 @@ spec:
...
-              ## regardless of network). Operators opt in via an overlay override;
-              ## see assert-gateway-source-ranges below for the configured path.
-              ! grep -q 'loadBalancerSourceRanges:' "${MANIFEST}"
+              ## The private-by-default RFC1918 ranges must be rendered onto the
+              ## Service, and an any-source range must NOT appear.
+              grep -q 'loadBalancerSourceRanges:' "${MANIFEST}"
+              grep -q '10.0.0.0/8' "${MANIFEST}"
+              grep -q '172.16.0.0/12' "${MANIFEST}"
+              grep -q '192.168.0.0/16' "${MANIFEST}"
+              ! grep -q '0.0.0.0/0' "${MANIFEST}"
```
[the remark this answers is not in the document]
yuanchen8911: Good catch — the whole-file grep matched the `0.0.0.0/0` token in the template comments, so the assertion could fail even when the rendered list is private. Fixed in 7ac34f00: all source-range checks now match `loadBalancerSourceRanges` list entries only (`^[[:space:]]*-[[:space:]]*…`), so the comment text no longer interferes.

pkg/bundler/bundler.go
```diff
@@ -1158,76 +1160,129 @@ const (
...
+			"%s: inference-gateway is explicitly opened to the entire internet "+
+				"(%s includes an any-source CIDR such as 0.0.0.0/0). This is a deliberate opt-in; "+
+				"scope it to trusted networks unless public exposure is intended.",
+			agentgatewayComponentName, agentgatewaySourceRangesPath,
+		)
+		b.appendWarning(msg)
+		slog.Warn("agentgateway inference-gateway is explicitly opened to 0.0.0.0/0 (deliberate any-source opt-in)",
+			"component", agentgatewayComponentName,
+			"path", agentgatewaySourceRangesPath,
+		)
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 7ac34f00: the log line is now protocol-agnostic — "explicitly opened via an any-source CIDR (deliberate public opt-in)" — so an explicit `::/0` opt-in no longer logs as IPv4.

recipes/components/agentgateway/manifests/inference-gateway.yaml
```diff
@@ -39,8 +39,11 @@ spec:
   # inference endpoint to the entire internet. spec.service.spec is merged onto
   # the generated Service, and loadBalancerSourceRanges is a portable Service
   # field honored by the AWS, GCP, Azure, and OCI cloud load balancers — so this
-  # single setting locks the gateway down on every platform. An empty
-  # allowedSourceRanges list leaves the Service unrestricted (0.0.0.0/0).
+  # single setting locks the gateway down on every platform. Through `aicr
+  # bundle`, allowedSourceRanges is never empty here: the bundler injects the
+  # private RFC1918 ranges by default (see #1373). A literally-empty value (e.g.
```
njhensley, line 44: 🟡 **Minor — the safe default lives only in the bundler, not the chart.** As this comment acknowledges, `values.yaml` still ships `allowedSourceRanges: []`, and the RFC1918 default is injected in Go (`resolveAgentgatewayExposure`). Anyone rendering this component chart outside `aicr bundle` (raw `helm template`, or a GitOps flow that consumes the recipe's `componentRef` + values directly) gets the Service open to `0.0.0.0/0`.

Baking the RFC1918 set into `values.yaml` as the actual chart default would be defense-in-depth (safe standalone too) and simpler than the Go-side mutation. Worth confirming the threat model intends bundler-only enforcement — if so, fine to leave as documented.

recipes/components/agentgateway/values.yaml
```diff
@@ -40,11 +40,15 @@ inferenceExtension:
...
-# firewall every downstream deployment to NVIDIA's network and lock external
-# operators out of their own gateway.
+# This value is EMPTY here, but it is NOT what ships: when it is left empty,
+# `aicr bundle` injects the private RFC1918 ranges (10.0.0.0/8, 172.16.0.0/12,
+# 192.168.0.0/16) before rendering, so the deployed load balancer is reachable
+# from inside the cluster/VPC but denied to the public internet — it is never
+# emitted open to 0.0.0.0/0 (see #1373). It is intentionally NOT defaulted to an
+# NVIDIA-specific range: a baked-in customer CIDR would firewall every downstream
+# deployment to NVIDIA's network and lock external operators out of their own
+# gateway. A corporate VPN egresses from a public IP and is therefore NOT covered
```
njhensley, line 50: 🟡 **Minor — behavior change vs current `main` (open → private) deserves a release-note callout.** Today the gateway is reachable from any source; after this change it is scoped to RFC1918. Operators who currently reach the endpoint from a **public** IP (corporate VPN egress, a public client) will be silently blocked on upgrade. The caveat is well-documented here and the opt-out is explicit, but a BREAKING/behavior-change note in the release notes (with the `--set-json` remediation) would save an operator a debugging session.

pkg/bundler/bundler.go
```diff
@@ -1158,76 +1160,164 @@ const (
...
+// (and does not override it via a recipe componentRef). These are the RFC1918
+// private ranges: they keep the gateway reachable from inside the cluster/VPC and
+// from privately-routed peers, while denying the public internet (which includes
+// VPN egress that presents a public source IP). Operators open specific public
+// clients with --set-json agentgateway:allowedSourceRanges='["<cidr>"]', or
+// expose it publicly with ["0.0.0.0/0"]. This is deliberately generic (not a
+// customer-specific network) to avoid firewalling every deployment to one site.
+// To flip to a deny-all default instead, replace this with a single unreachable
+// CIDR such as "255.255.255.255/32". See #1373.
+var agentgatewayDefaultSourceRanges = []string{
```
njhensley, line 1233: 🔵 **Nitpick — RFC1918 ≠ cluster-only.** `10.0.0.0/8` (and the others) admit the entire private space; in a shared/multi-tenant VPC this trusts more than just this cluster's peers. It's a defensible "private by default" stance (it does deny the public internet), and the comment helpfully documents the deny-all alternative (`255.255.255.255/32`) — just flagging that "private" here means "all privately-routed traffic," not "this cluster."

Separately (body, not inline): `IsAnySourceCIDR` still matches only a literal `/0`, so a deliberate subnet-union covering all space (`0.0.0.0/1` + `128.0.0.0/1`) classifies as scoped — already a documented deliberate-evasion non-goal, so no action needed.

njhensley (dismissed):
## ✅ Review — private-by-default agentgateway inference-gateway exposure

**Method.** Reviewed across four lenses (correctness, security, domain/architecture, operability) with an adversarial verification pass; findings re-derived against the resolved code at `7ac34f0`.

### Overall

The private-by-default design is sound and a clear improvement over an open default: an unset/empty `allowedSourceRanges` now injects the RFC1918 private set into `loadBalancerSourceRanges` so the gateway denies the public internet while bundles still generate (no breaking failure). I verified:

- `resolveAgentgatewayExposure` mutates the merged `componentValues` in the unset case **before** render, at the same call site as the prior warning — ordering is correct, so overlay/`componentRef`/`--set-json` overrides are all seen first.
- `IsValidCIDR` is now strict (`netip.ParsePrefix` + `p == p.Masked()` + reject `Is4In6`). This **fail-fasts non-canonical CIDRs** before a K8s apply failure and **closes the IPv4-mapped-IPv6 (`::ffff:…`) any-source blind spot**. No existing in-repo CIDR (all canonical) is broken by the stricter check.
- The `bundle-defaults-private` chainsaw step asserts the three RFC1918 entries render and `0.0.0.0/0` does not, with anchored regex.

No blockers, no Majors. The remaining items are minor/discussion-level (see inline). Recommendation: **approve** — please consider a release-note callout for the open→private reachability change before merge.

### Tier summary

|---|---|---|---|
| 0 | 0 | 2 | 2 |

docs/user/component-catalog.md
```diff
@@ -68,9 +68,11 @@ The output lists every component with its pinned version and configuration value
...
+
+To restrict it to specific trusted networks instead — for example to allow a corporate VPN, which egresses from a **public** IP and is therefore *not* covered by the RFC1918 default — set `agentgateway.allowedSourceRanges` to a list of CIDR (Classless Inter-Domain Routing) blocks. The values replace the default and are rendered into the generated Service's `spec.loadBalancerSourceRanges`, which the AWS, GCP, Azure, and OCI cloud load balancers all honor — so one setting locks the gateway down on every platform.
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in a15308dc. Reworded the opening paragraph so the section is internally consistent: the raw `LoadBalancer` the agentgateway controller materializes is internet-facing, and `aicr bundle` scopes it to private networks by default — with the opt-in and validation behavior described in the paragraphs below. The `--set` example and Exposure guardrails text already described the private-by-default + opt-in + validation flow, so no further change there.

pkg/bundler/bundler_test.go
```diff
@@ -2069,59 +2069,79 @@ func TestAgentgatewayComponentExistsInRegistry(t *testing.T) {
...
 			present:     true,
 			wantWarning: true,
 		},
 		{
-			name:        "does not warn when agentgateway is absent",
-			present:     false,
-			wantWarning: false,
+			name:    "no-op when agentgateway is absent",
+			present: false,
 		},
```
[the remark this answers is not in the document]
yuanchen8911: Added in a15308dc. The table now has a `[]string` scoped-list case (pins the `case []string` branch) and a `[]any{"10.0.0.0/8", 123}` case (pins the non-string-list-entry rejection). `go test ./pkg/bundler/...` passes.

tests/chainsaw/bundle-templates/agentgateway/chainsaw-test.yaml
```diff
@@ -57,32 +62,23 @@ spec:
...
+              ## The private-by-default RFC1918 ranges must be rendered onto the
+              ## Service, and an any-source range must NOT appear.
+              grep -q 'loadBalancerSourceRanges:' "${MANIFEST}"
+              ## Match list entries, not the surrounding template comments (which
+              ## mention these tokens), so the assertions key off the rendered
+              ## loadBalancerSourceRanges list only.
+              grep -Eq '^[[:space:]]*-[[:space:]]*10\.0\.0\.0/8([[:space:]]|$)' "${MANIFEST}"
+              grep -Eq '^[[:space:]]*-[[:space:]]*172\.16\.0\.0/12([[:space:]]|$)' "${MANIFEST}"
+              grep -Eq '^[[:space:]]*-[[:space:]]*192\.168\.0\.0/16([[:space:]]|$)' "${MANIFEST}"
+              ! grep -Eq '^[[:space:]]*-[[:space:]]*0\.0\.0\.0/0([[:space:]]|$)' "${MANIFEST}"
```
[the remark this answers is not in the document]
yuanchen8911: Added in a15308dc — an anchored `\! grep -Eq '^[[:space:]]*-[[:space:]]*::/0([[:space:]]|$)'` negative assertion now sits alongside the `0.0.0.0/0` check, so the default-private step fully matches the any-source contract for both families.

recipes/components/agentgateway/manifests/inference-gateway.yaml
yuanchen8911: Confirming the threat model: bundler-only enforcement is intentional. `aicr bundle` (recipe -> bundle) is AICR's supported delivery path, and enforcing there lets the bundler distinguish *unset* (-> inject RFC1918 default) from *explicitly set* (overlay / `componentRef` / `--set-json`) so overrides are honored while the unset case still fails safe. Baking the RFC1918 set into `values.yaml` would make the chart default a real list, which loses that unset-vs-explicit distinction the Go path keys off of. The standalone `helm template` caveat is documented in the manifest comment, so leaving the safe default in the bundler as-is. Open to revisiting if we want chart-level defense-in-depth as a follow-up.

recipes/components/agentgateway/values.yaml
yuanchen8911: Agreed — added an explicit "Release note (behavior change — inference-gateway open → private)" callout to the PR description, with the `--set-json` remediation (and `'["0.0.0.0/0"]'` to restore prior open behavior). Flagged that it must be carried into the release notes so an operator reaching the endpoint from a public IP knows why they're blocked on upgrade.

pkg/bundler/bundler.go
yuanchen8911: Acknowledged — "private by default" here means privately-routed (RFC1918) traffic, not strictly this cluster's peers; in a shared/multi-tenant VPC it trusts the whole private space. It's the deliberate stance (denies the public internet while keeping bundles generating), and the deny-all alternative (`255.255.255.255/32`) is documented in the comment. On the `IsAnySourceCIDR` literal-`/0` point: agreed, a deliberate subnet-union evasion is a documented non-goal, so no change.

yuanchen8911:
@coderabbitai full review

njhensley (approved):
## ✅ Re-approval — `a15308d`

My prior approval (`7ac34f0`) was auto-dismissed by the new push. Re-reviewed the incremental diff `7ac34f0..a15308d`: the changes are **non-substantive improvements, no production-logic change**.

- `pkg/bundler/bundler.go` and `pkg/netutil/cidr.go` are **byte-identical** to the approved revision — the private-by-default injection, call-site ordering, and strict `IsValidCIDR` (canonical + reject `Is4In6`) are unchanged.
- `bundler_test.go`: +2 cases — a `[]string` scoped list passes, and a non-string list entry errors. These close the two previously-untested branches in `classifyAgentgatewaySourceRanges`. 👍
- chainsaw `bundle-defaults-private`: +1 assertion that `::/0` is absent from the default-private manifest. Strengthens the check.
- `component-catalog.md`: intro reworded for clarity.

The earlier inline notes (🟡 bundler-only safe default / 🟡 open→private release-note callout / 🔵 RFC1918 breadth) still stand as non-blocking discussion items. **Approving.**

Issue #1436, opened 2026-06-23
Docs: Sigstore/Fulcio network blocking + recommended evidence publishing path


njhensley:
Part of #1431.

## Summary

Document the known Fulcio/Sigstore connectivity issue and the recommended way to publish evidence. Corporate VPN and some home networks block TLS to `fulcio.sigstore.dev` (upstream IP-level rejection, not an AICR bug), which forces contributors onto a phone hotspot for keyless signing. The recommended remedy is the fork-based GitHub Actions signing workflow, with the local split-leg `validate` (on VPN) → `evidence publish` (off VPN) as the fallback.

## Acceptance criteria

- [ ] A contributor-facing doc page (under `docs/contributor/` or the evidence section) explains the symptom, the cause (upstream Sigstore IP blocking), and the CI-based remedy.
- [ ] Notes the requirement that the fork's `aicr-evidence` GHCR package be **public** so the gate's verify pull (and the signing workflow's pre-sign pull) doesn't 403.
- [ ] Cross-links the fork-based signing workflow and the two-phase publish issues.

## References

- `docs/design/007-recipe-evidence.md` and subpages
- `pkg/cli/evidence_publish.go`

mchmarny:
Implemented in #1451 (docs/contributor/evidence-publishing.md — Sigstore/Fulcio connectivity + recommended publishing path). Merged to main.

Issue #1434, opened 2026-06-23
aicr evidence: two-phase push-without-sign + sign-existing support


njhensley:
Part of #1431.

## Summary

Let AICR produce a "mostly-formed" pointer and bundle **without** a signature, and add a separate path that signs an already-pushed bundle and patches its pointer in place. This decouples the network-light work (compute digest, assemble bundle, push to registry, write pointer skeleton) from the Fulcio-bound work (sign + log to Rekor + fill in signer identity/index). Today signing is mandatory whenever `--push` is set — there is no no-sign path — which couples the whole flow to a network that can reach Fulcio.

## Motivation / Context

- The pointer types already model an unsigned state: `signer.rekorLogIndex` is a pointer that is nil when unsigned (`pkg/evidence/attestation/types.go`). The schema supports it; the CLI doesn't produce it.
- This is what the fork-based signing workflow consumes: contributor publishes unsigned + commits pointer locally; CI fills in the signature.

## Proposed approach

**(a) Push-without-sign:** add `--no-sign` (or `--unsigned`) to `validate --emit-attestation` and/or `evidence publish` that:
- assembles the bundle and pushes it to the registry (content-addressed, unsigned), and
- writes a pointer with `bundle.oci` / `bundle.digest` populated and `signer` empty / `rekorLogIndex` nil.

**(b) Sign-existing:** add an `evidence sign <pointer>` path (or extend `evidence publish`) that takes the already-pushed bundle referenced by the pointer, signs it with ambient OIDC, logs to Rekor, and patches the pointer's `signer` block in place — without re-emitting the bundle. This is exactly what the fork workflow calls.

**(c) Verifier:** `evidence verify` should treat an unsigned pointer as a distinct "pending signature" status rather than "invalid," so an in-flight PR isn't flagged as broken (gate stays warning-only).

## Acceptance criteria

- [ ] `aicr ... --no-sign` produces a committable pointer with bundle digest set and signer empty, and pushes the unsigned bundle.
- [ ] A sign-existing step fills in `signer.identity`, `signer.issuer`, `signer.rekorLogIndex` from the committed pointer without re-emitting the bundle.
- [ ] `evidence verify` reports unsigned pointers as a non-failing "pending" state.
- [ ] Unit coverage for the unsigned pointer round-trip in `pkg/evidence/attestation`.

## References

- `pkg/evidence/attestation/types.go` (`Pointer`, `PointerSigner.RekorLogIndex *...`)
[...]

PR #1451, opened 2026-06-24
feat(evidence): fork-based signing workflow + publishing docs

### LINKED ISSUE #1433: Fork-based evidence signing workflow driven by committed unsigned pointers
Part of #1431.

## Summary

Add a GitHub Actions workflow that a contributor runs in their fork to complete the *signing* leg of evidence publishing. The contributor first does the network-light work locally with AICR: publish the **unsigned** bundle to their own registry and commit the pointer (see two-phase publish issue). The workflow then operates purely on the branch — it discovers any pointer in `recipes/evidence/*.yaml` whose `signer` is empty / `rekorLogIndex` is nil, signs the bundle that pointer already references (ambient OIDC → Fulcio/Rekor), patches the pointer's `signer` block in place, and commits it back. No recipe-name or bundle-ref inputs to thread through `workflow_dispatch`.

## Why this shape

- Eliminates the friction of passing a bundle artifact or recipe name into the workflow — the committed pointer *is* the work item, and it already carries `bundle.oci` + `bundle.digest`.
- The only step that genuinely needs Fulcio (signing) runs where Fulcio is reachable (Actions); everything else stays local and VPN-friendly. Sigstore keyless signing from a laptop is currently blocked on corporate VPN and some home networks (upstream Fulcio IP-level rejection).
- The signer becomes the GitHub Actions OIDC identity, not a personal one.

## Contributor flow

1. Local (on VPN, no Fulcio): `aicr validate ... --emit-attestation ./out --push ghcr.io/<owner>/aicr-evidence --no-sign` → unsigned bundle pushed, pointer written with empty signer (two-phase publish issue).
2. Make the fork's `aicr-evidence` package **public** so verify pulls don't 403.
3. Commit the pointer, push the branch.
4. Workflow runs → signs, logs to Rekor, patches `signer.identity` / `issuer` / `rekorLogIndex`, commits back.

## Trigger — open design question (resolve in this issue)

- **Option A (explicit):** `workflow_dispatch` only — predictable, no surprise commits, but one manual click.
- **Option B (automatic on push):** trigger on push to fork branches when an unsigned pointer is present. Lowest friction, but a workflow that commits back to the branch that triggered it needs loop-guards (skip when no unsigned pointers; don't re-trigger on its own signing commit) and clear behavior on forks where Actions / `id-token` may be disabled by default.
- Recommend landing **A** first, with **B** as a fast-follow once the discover-and-patch path is proven.

## Acceptance criteria

- [ ] `.github/workflows/evidence-publish.yaml`, runnable in a fork, `permissions: id-token: write` (+ `contents: write` for commit-back); no required inputs.
- [ ] Discovers all unsigned pointers on the branch; signs each referenced bundle with ambient OIDC; patches the pointer and commits it back.
- [ ] No-op (clean exit, no commit) when there are no unsigned pointers.
- [ ] **Fork `aicr-evidence` package must be reachable for verification** — documented, and the workflow fails with a clear message if it can't pull the bundle it's about to sign (ties to the gate-error-feedback issue).
- [ ] Auto-trigger (Option B) tracked as a follow-up with the loop-guard requirements written down.

## References

- Discovery target: `recipes/evidence/*.yaml`; `pkg/evidence/attestation/types.go` (`Pointer`, nil `RekorLogIndex` = unsigned)
- Signing: `pkg/cli/evidence_publish.go` sign-existing path; ambient OIDC `pkg/bundler/attestation/oidc.go` (`FetchAmbientOIDCToken`)
- Depends on the two-phase push-without-sign issue.

mchmarny:
## Summary

Adds the Fulcio-bound signing leg of two-phase evidence publish: a `workflow_dispatch` workflow a contributor runs in their fork to sign committed-but-unsigned evidence pointers, plus a contributor doc explaining the Sigstore connectivity problem this works around.

## Motivation / Context

Final piece of the evidence-gate pilot epic. Keyless signing reaches `fulcio.sigstore.dev`/`rekor.sigstore.dev`, which corporate VPNs and some home networks block at the IP level. The fix is to keep the network-light push leg local (`--no-sign`, merged in #1445) and move only the signing leg to CI, where Sigstore is reachable. This consumes the `aicr evidence sign` CLI merged in #1446.

Fixes: #1433, #1436
Related: #1431, #1434

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: GitHub Actions workflow + script

## Implementation Notes

- **`.github/workflows/evidence-publish.yaml`** — `workflow_dispatch`, no inputs (the committed pointer is the work item). `id-token: write` (ambient keyless OIDC) + `contents: write` (commit the patched pointer back). Builds aicr, runs the discovery script, commits signed pointers back. Clean no-op when nothing is unsigned. **Option B (auto-trigger on push) is documented in the header as a follow-up** with its loop-guard requirements; Option A (manual) lands first per the issue.
- **`.github/scripts/evidence-sign-unsigned.sh`** — "unsigned" = first attestation has no `signer` block (the `--no-sign` state); already-signed pointers are skipped (idempotent re-runs). Fails closed with an actionable message when a bundle can't be pulled (a private fork package → HTTP 403 on the pre-sign pull).
- **`docs/contributor/evidence-publishing.md`** (+ nav entry) — the Fulcio block, the split-leg path, the **public-package requirement**, and the local-only fallback.
- **Security:** runs in the contributor's fork on a branch they control, signing with the fork's own OIDC identity. The write grants act on the fork, not upstream — no `pull_request_target`/untrusted-ref exposure (manual dispatch on the fork's own code).

## Testing

Local smoke tests (built aicr, ran the discovery script):

| Scenario | Result |
|---|---|
| Both committed pointers (signed) | skipped → `signed=0`, clean no-op |
| Temp unsigned pointer | detected → routed to `aicr evidence sign` → fails closed with the public-package hint when the bundle can't be pulled |

`shellcheck -S warning` + `actionlint` + `yamllint` all clean.

## Risk Assessment

- [x] **Low** — New, manually-triggered workflow + script + docs; nothing wired into existing PR/merge gates. No effect unless a contributor dispatches it in their fork.

**Rollout notes:** Contributors enable Actions in their fork and make their `aicr-evidence` package public (documented). Upstream behavior is unchanged.

## Checklist

- [x] Linter passes (`shellcheck`, `actionlint`, `yamllint`)
- [x] I added/updated tests for new functionality (local smoke tests; no unit-test surface for a dispatch workflow)
- [x] I updated docs (new contributor page + nav entry)

### DIFF
--- a/.github/scripts/evidence-sign-unsigned.sh
+++ b/.github/scripts/evidence-sign-unsigned.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Discover committed-but-unsigned evidence pointers and sign each in place.
+# Consumed by .github/workflows/evidence-publish.yaml (the fork-based
+# signing leg): a contributor pushes an unsigned bundle and commits the
+# pointer locally (`aicr ... --no-sign`), then this signs the bundle each
+# pointer references using the runner's ambient OIDC token and patches the
+# pointer's signer block in place.
+#
+# A pointer is "unsigned" iff its first attestation has no `signer` block —
+# the state `--no-sign` produces. Already-signed pointers are skipped, so
+# re-running is idempotent and safe.
+#
+# Required env:
+#   AICR   path to the aicr binary (built from a trusted source)
+#
+# Requires the mikefarah `yq` (v4) — the same `yq` the rest of the repo's
+# scripts assume. The kislyuk/python `yq` has incompatible syntax and is not
+# supported. On GitHub-hosted runners mikefarah yq is preinstalled.
+#
+# Any extra arguments are forwarded to `aicr evidence sign` (e.g.
+# --plain-http / --insecure-tls for local-registry tests).
+#
+# Outputs (and, when set, appends to $GITHUB_OUTPUT):
+#   skipped=<n>  pointers already signed (no-op)
+#   signed=<n>   pointers signed this run
+#   failed=<n>   pointers whose signing (or parsing) failed
+#
+# Exit status is non-zero when any signing failed, so the workflow surfaces
+# the cause (commonly a private fork registry returning HTTP 403 on the
+# pre-sign pull — make the aicr-evidence package public).
+
+set -euo pipefail
+
+: "${AICR:?AICR is required (path to aicr binary)}"
+
+extra_args=("$@")
+
+signed=0
+failed=0
+skipped=0
+
+shopt -s nullglob
+for pointer in recipes/evidence/*.yaml; do
+  # signer absent => unsigned (the --no-sign state). A signed-without-Rekor
+  # pointer still has a signer block, so it is correctly skipped.
+  #
+  # Fail closed on a yq error: a missing yq or malformed YAML must NOT look
+  # like an unsigned pointer (which would resign an already-signed bundle and
+  # break rerun idempotence). Treat a parse failure as a hard failure.
+  if ! signer=$(yq eval '.attestations[0].signer // "null"' "$pointer" 2>/dev/null); then
+    echo "::error::failed to parse ${pointer} with yq — invalid YAML, or mikefarah yq (v4) not installed"
+    failed=$((failed + 1))
+    continue
+  fi
+  if [[ "$signer" != "null" ]]; then
+    echo "skip (already signed): ${pointer}"
+    skipped=$((skipped + 1))
+    continue
+  fi
+
+  echo "signing unsigned pointer: ${pointer}"
+  # --yes: never pause for the interactive identity-disclosure prompt in CI
+  # (ambient OIDC does not prompt, but this is belt-and-suspenders).
+  if "$AICR" evidence sign "$pointer" --yes "${extra_args[@]}"; then
+    signed=$((signed + 1))
+  else
+    echo "::error::failed to sign ${pointer} — is the bundle's registry package public? (a private fork package returns HTTP 403 on the pre-sign pull)"
+    failed=$((failed + 1))
+  fi
+done
+
+echo "skipped=${skipped}"
+echo "signed=${signed}"
+echo "failed=${failed}"
+
+if [[ -n "${GITHUB_OUTPUT:-}" ]]; then
+  {
+    echo "skipped=${skipped}"
+    echo "signed=${signed}"
+    echo "failed=${failed}"
+  } >> "$GITHUB_OUTPUT"
+fi
+
+# Fail closed if any pointer could not be signed.
+[[ "$failed" -eq 0 ]]
--- a/.github/workflows/evidence-publish.yaml
+++ b/.github/workflows/evidence-publish.yaml
@@ -1,0 +1,135 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Fork-based evidence signing — the Fulcio-bound leg of two-phase publish.
+#
+# A contributor publishes an UNSIGNED evidence bundle locally (where the
+# cluster lives, often behind a VPN that blocks Sigstore) and commits the
+# pointer:
+#
+#   aicr validate -r recipes/overlays/<slug>.yaml -s snapshot.yaml \
+#     --emit-attestation ./out --push ghcr.io/<owner>/aicr-evidence --no-sign
+#   cp ./out/pointer.yaml recipes/evidence/<slug>.yaml
+#
+# They then run THIS workflow from their fork's Actions tab. It discovers
+# every committed pointer with an empty signer block, signs the bundle each
+# one references with the runner's ambient OIDC identity (where Sigstore IS
+# reachable), patches the pointer's signer block in place, and commits it
+# back to the branch. Clean no-op when there is nothing unsigned.
+#
+# Why workflow_dispatch (Option A): no recipe-name or bundle-ref inputs to
+# thread through — the committed pointer is the work item, and it already
+# carries bundle.oci + bundle.digest. Manual dispatch keeps the commit-back
+# predictable (no surprise commits) and avoids the loop-guards an automatic
+# push trigger would require.
+#
+# Follow-up — Option B (auto-trigger on push). Lower friction but needs:
+#   * skip the run entirely when no unsigned pointer is present;
+#   * a loop-guard so the workflow's own signing commit does not re-trigger
+#     it (e.g. gate on the actor / commit message, or [skip ci]);
+#   * graceful handling on forks where Actions or id-token are disabled.
+# Land Option A first; add B once the discover-and-patch path is proven.
+#
+# Security note: this runs in the *contributor's fork* on a branch they
+# control, signing with the fork's own GitHub Actions OIDC identity. The
+# id-token:write / contents:write grants act on the fork, not upstream —
+# there is no pull_request_target / untrusted-ref exposure (it is manual
+# dispatch on the fork's own code).
+#
+# The fork's aicr-evidence registry package must be PUBLIC: the pre-sign
+# pull (and the upstream evidence gate) fetch the bundle, and a private
+# package returns HTTP 403. The signing step fails with a clear message
+# pointing at this when it can't pull.
+
+name: "Recipe Evidence: Sign"
+
+on:
+  workflow_dispatch: {}
+
+permissions:
+  contents: write   # commit the patched pointer(s) back to the branch
+  id-token: write   # ambient OIDC for keyless Fulcio/Rekor signing
+
+concurrency:
+  # Serialize per branch; do NOT cancel in progress — a cancel mid-sign
+  # could leave a bundle signed in the registry but its pointer un-patched.
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  sign:
+    name: Sign unsigned evidence pointers
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          # Do not persist the push token in .git/config: it would sit there
+          # across the fork-controlled `go build` step. The push step below
+          # supplies the token explicitly only when it needs it.
+          persist-credentials: false
+
+      - name: Load versions
+        id: versions
+        uses: ./.github/actions/load-versions
+
+      - name: Set up Go
+        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16  # v6.5.0
+        with:
+          go-version: ${{ steps.versions.outputs.go }}
+          cache: false
+
+      - name: Build aicr
+        env:
+          GOFLAGS: -mod=vendor
+        run: |
+          # Verify the vendored dependency tree against go.sum before building
+          # the binary that will sign with the runner's OIDC identity.
+          go mod verify
+          go build -o ./bin/aicr ./cmd/aicr
+
+      - name: Sign unsigned evidence pointers
+        id: sign
+        env:
+          AICR: ${{ github.workspace }}/bin/aicr
+        run: .github/scripts/evidence-sign-unsigned.sh
+
+      - name: Commit signed pointers
+        # always() so that if one pointer fails to sign, the ones that DID
+        # sign are still committed — otherwise the registry would hold a signed
+        # bundle whose committed pointer stays unsigned. The job still fails
+        # (the sign step exited non-zero), surfacing the failure.
+        if: ${{ always() && steps.sign.outputs.signed != '0' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          if git diff --quiet -- recipes/evidence/; then
+            echo "No pointer changes to commit (nothing to sign)."
+            exit 0
+          fi
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add recipes/evidence/
+          # -s adds a `Signed-off-by:` (DCO) trailer matching the bot author,
+          # so the DCO bot passes on the commit-back. The commit is NOT
+          # cryptographically signed (no -S) — it fills in the evidence
+          # pointer's signer block (the bundle signature); repo commit-signing
+          # policy is satisfied by the eventual squash-merge.
+          git commit -s -m "chore(evidence): sign pending evidence pointers"
+          # Push with the token supplied inline (credentials are not persisted
+          # in .git/config). A non-fast-forward here is a clear, recoverable
+          # failure: re-dispatch after pulling.
+          git push "https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git" \
+            "HEAD:${GITHUB_REF_NAME}"
--- a/docs/contributor/evidence-publishing.md
+++ b/docs/contributor/evidence-publishing.md
@@ -1,0 +1,127 @@
+# Publishing Recipe Evidence
+
+Recipe evidence is a signed bundle that proves a recipe's validators passed
+on hardware matching its `criteria`. Producing it has two legs:
+
+1. **Validate + push (network-light).** Runs where the cluster lives —
+   often a corporate VPN. Captures a snapshot, runs the validators, builds
+   the bundle, and pushes it to an OCI registry.
+2. **Sign (Fulcio-bound).** Signs the pushed bundle with keyless OIDC
+   (Sigstore Fulcio + Rekor) and records the signature in the committed
+   pointer.
+
+The two legs need different network access, and the signing leg is the one
+contributors most often can't complete locally.
+
+## The Fulcio connectivity problem
+
+Keyless signing reaches `fulcio.sigstore.dev` and `rekor.sigstore.dev`.
+**Corporate VPNs and some home networks block TLS to these hosts** — the
+connection is rejected at the IP level upstream, not by AICR. The symptom
+is a hang or a TLS/connection-reset error during the sign step, even though
+`aicr validate` and the registry push succeed.
+
+This is not an AICR bug and AICR cannot work around it; the block is on the
+path to Sigstore's public-good infrastructure. Contributors have reported
+needing a phone hotspot to sign from a laptop.
+
+## Recommended path: split the legs, sign in CI
+
+The fix is to keep the network-light leg local and move only the
+Fulcio-bound leg to GitHub Actions, where Sigstore is reachable. The signer
+identity becomes your fork's GitHub Actions OIDC identity rather than a
+personal one.
+
+### 1. Validate and push an unsigned bundle (local, on VPN)
+
+```shell
+aicr snapshot -o snapshot.yaml
+aicr validate \
+  -r recipes/overlays/<slug>.yaml \
+  -s snapshot.yaml \
+  --emit-attestation ./out \
+  --push ghcr.io/<your-fork-owner>/aicr-evidence \
+  --no-sign
+cp ./out/pointer.yaml recipes/evidence/<slug>.yaml
+```
+
+`--no-sign` skips all OIDC/Fulcio/Rekor work, so this runs even where
+Sigstore egress is blocked. It pushes the content-addressed bundle and
+writes a pointer with an empty `signer` block. See
+[`aicr validate`](../user/cli-reference.md#aicr-validate) and
+[`aicr evidence publish`](../user/cli-reference.md#aicr-evidence-publish)
+(which also accepts `--no-sign` if you push as a separate step).
+
+> **Make the registry package public.** The signing step (and the
+> repository's evidence gate) pull the bundle back to sign and verify it.
+> If your fork's `aicr-evidence` package is private, the pull fails with an
+> HTTP **403** and the gate reports `registry-forbidden`. On GHCR, set the
+> `aicr-evidence` package visibility to **public** under your account's
+> *Packages* settings. Any OCI-1.1 registry works; it just has to be
+> readable.
+
+### 2. Commit the unsigned pointer and push your branch
+
+```shell
+git add recipes/evidence/<slug>.yaml
+# Use -s (DCO sign-off) — required for all contributors. NVIDIA org members
+# additionally use -S (cryptographic signing); external contributors use -s
+# alone. See CONTRIBUTING.md.
+git commit -s -m "evidence: <slug> (unsigned; sign in CI)"
+git push
+```
+
+`aicr evidence verify` reports an unsigned pointer as a non-failing
+**pending signature** state, so an in-flight PR is not flagged as broken.
+
+### 3. Sign in your fork via GitHub Actions
+
+Run the **Recipe Evidence: Sign** workflow
+(`.github/workflows/evidence-publish.yaml`) from your fork's *Actions* tab
+(it is `workflow_dispatch` — no inputs). The filename says *publish* because it
+anticipates a future auto-publish leg (see the workflow header); today it only
+signs. The workflow:
+
+- discovers every pointer in `recipes/evidence/*.yaml` with an empty
+  `signer` (i.e. unsigned),
+- signs the bundle each one already references using the runner's ambient
+  OIDC token ([`aicr evidence sign`](../user/cli-reference.md#aicr-evidence-sign)),
+- patches the pointer's `signer` block in place and commits it back to the
+  branch.
+
+It is a clean no-op when there are no unsigned pointers, and it fails with a
+clear message if it cannot pull a bundle (the public-package requirement
+above). Pull the commit it pushes (`git pull`) — your PR now carries a pointer
+whose `signer` block is filled in (the *bundle* is signed; the commit-back
+itself is a normal, unsigned GitHub Actions commit, which the eventual
+squash-merge re-signs under the repo's policy).
+
+> The workflow declares `id-token: write` in its own `permissions:` block —
+> it is not a default. Your fork must have GitHub Actions enabled and not
+> restrict workflow OIDC token issuance (the default fork setting allows it).
+>
+> Avoid pushing other commits to the branch while the workflow runs: it
+> commits the patched pointers back, and a concurrent push would cause a
+> non-fast-forward — re-dispatch after `git pull` if that happens.
+
+## Fallback: split the legs locally
+
+If you have a host with Sigstore egress (a jump box, CI runner, or
+hotspot), you can run the signing leg there instead of in Actions:
+
+```shell
+# On VPN, where the cluster is: produce an unsigned bundle.
+aicr validate -r recipes/overlays/<slug>.yaml -s snapshot.yaml --emit-attestation ./out
+
+# Off VPN, where Sigstore is reachable: sign, push, and write the pointer.
+aicr evidence publish ./out --push ghcr.io/<your-fork-owner>/aicr-evidence
+cp ./out/pointer.yaml recipes/evidence/<slug>.yaml
+```
+
+The signed artifact is content-addressed, so the result is identical
+regardless of which host ran which leg.
+
+## See also
+
+- [ADR-007: Verifiable Recipe Test Evidence](../design/007-recipe-evidence.md) — trust model and bundle format.
+- [`aicr evidence` CLI reference](../user/cli-reference.md#aicr-evidence-sign) — `sign`, `publish`, `verify`.
--- a/docs/index.yml
+++ b/docs/index.yml
@@ -90,3 +90,5 @@ navigation:
         path: contributor/skills.md
       - page: Maintaining AICR
         path: contributor/maintaining.md
+      - page: Publishing Recipe Evidence
+        path: contributor/evidence-publishing.md


faganihajizada (dismissed):
👍🏻

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,115 @@
...
+        run: go build -o ./bin/aicr ./cmd/aicr
+
+      - name: Sign unsigned evidence pointers
+        id: sign
+        env:
+          AICR: ${{ github.workspace }}/bin/aicr
+        run: .github/scripts/evidence-sign-unsigned.sh
+
+      - name: Commit signed pointers
+        if: ${{ steps.sign.outputs.signed != '0' }}
```
njhensley: **🟠 Major — Commit step's custom `if:` drops the implicit `success()` — partial sign-failure still commits & pushes**

A step `if:` expression with no status function (`success()`/`failure()`/`always()`) drops the implicit `success()` and evaluates as if wrapped in `always()`. The script writes `signed=<n>` to `$GITHUB_OUTPUT` *before* its final `[[ "$failed" -eq 0 ]]` fails the step, so on a partial failure (e.g. 3 signed, 1 returns HTTP 403) `signed != '0'` is still true.

**Blast radius:** The commit step runs anyway, commits the successfully-patched pointers, and `git push`es them — publishing a partial result that the script's fail-closed exit (`evidence-sign-unsigned.sh` L74) was meant to block. The red job status under-sells that pointers actually landed.

**Fix:** Restore the success guard explicitly: `if: ${{ success() && steps.sign.outputs.signed != '0' }}`.

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,115 @@
...
+        run: |
+          if git diff --quiet -- recipes/evidence/; then
+            echo "No pointer changes to commit (nothing to sign)."
+            exit 0
+          fi
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add recipes/evidence/
+          git commit -m "chore(evidence): sign pending evidence pointers [skip ci]"
```
njhensley: **🟠 Major — Bot commit-back is unsigned/unverified — contradicts the `required_signatures` ruleset and the doc's "fully signed" wording**

This is a raw `git commit -m "…"` (no `-S`) pushed over the persisted `GITHUB_TOKEN`. Only API-made commits (as `peter-evans/create-pull-request` produces) are auto-Verified; a raw git commit is unsigned. The `main` ruleset includes `required_signatures` (confirmed via `gh api repos/NVIDIA/aicr/rulesets/12304487`), CLAUDE.md mandates `git commit -S`, and the existing commit-back workflow (`bom-refresh.yaml`) uses `peter-evans/create-pull-request` to get verified commits.

**Blast radius:** The push succeeds on the fork (the ruleset is on upstream), but an unsigned commit lands on the PR head — contradicting step 3 of the new doc ("your PR now carries a fully signed pointer", which conflates Sigstore signing with commit signing) and risking the merge gate. Also ironic: a workflow whose purpose is signing evidence injects an unsigned commit into the provenance chain. Could rise to 🔴 if `required_signatures` blocks the PR merge (not just direct pushes).

**Fix:** Route the commit-back through `peter-evans/create-pull-request` with `persist-credentials: false` (matches repo convention, yields a verified commit, and also resolves F5 and F7).

.github/scripts/evidence-sign-unsigned.sh
```diff
@@ -0,0 +1,74 @@
...
+
+signed=0
+failed=0
+skipped=0
+
+shopt -s nullglob
+for pointer in recipes/evidence/*.yaml; do
+  # signer absent => unsigned (the --no-sign state). A signed-without-Rekor
+  # pointer still has a signer block, so it is correctly skipped.
+  signer=$(yq eval '.attestations[0].signer // "null"' "$pointer" 2>/dev/null || echo "null")
```
njhensley: **🟡 Minor — `yq` classification fails open on malformed YAML / missing `attestations` (caught downstream, but should fail closed here)**

`signer=$(yq eval '.attestations[0].signer // "null"' "$pointer" 2>/dev/null || echo "null")` coerces both "signer genuinely absent" and "yq errored / no `attestations` array" to the literal `"null"` → routed to the signing path. This is the CLAUDE.md fail-open anti-pattern (a presence check passing on an ambiguous condition).

**Blast radius:** Softened in practice — `aicr evidence sign` runs `ValidateSignablePointer`, rejects a malformed/invalid pointer, and the `else` branch counts it `failed`, so the *outcome* is fail-closed. Worth tightening so the script itself classifies correctly.

**Fix:** Distinguish absent-signer from parse error and count the latter as `failed`:
```bash
if ! signer=$(yq eval '.attestations[0].signer // "__ABSENT__"' "$pointer"); then
  echo "::error::cannot parse ${pointer}"; failed=$((failed + 1)); continue
fi
[[ "$signer" != "__ABSENT__" ]] && { echo "skip (already signed): ${pointer}"; skipped=$((skipped+1)); continue; }
```

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,115 @@
...
+    name: Sign unsigned evidence pointers
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          # Keep the token so the signing commit can be pushed back to the
+          # dispatched branch.
+          persist-credentials: true
```
njhensley: **🟡 Minor — `persist-credentials: true` keeps the push token in `.git/config` across the fork-controlled `go build`**

The token persists in `.git/config` for the whole job, including the steps that compile and then *execute* `./cmd/aicr` from the dispatched branch, while `contents:write` + `id-token:write` are live.

**Blast radius:** Fork-only (manual dispatch on the contributor's own code, token scoped to the fork) — not an upstream-compromise vector, hence Minor. But it is a needlessly long-lived push credential exposed across an arbitrary-code build; the repo default (`bom-refresh.yaml`) is `persist-credentials: false`.

**Fix:** Set `persist-credentials: false` and re-introduce the token only at push time — cleanest via the `peter-evans` path in F2.

.github/scripts/evidence-sign-unsigned.sh
njhensley: **🟡 Minor — `yq` is unpinned (unlike every other tool here) and breaks for contributors running the script locally with python/kislyuk yq**

`yq eval '… // …'` is mikefarah/yq (Go) syntax; the workflow pins Go via `load-versions`+`setup-go` but relies on whatever ambient `yq` the runner ships. The other widely-installed `yq` (kislyuk, a jq wrapper) has no `eval` subcommand and different `//` semantics.

**Blast radius:** Works in CI today (ubuntu-latest ships mikefarah yq) but is undocumented and version-drift-fragile; a contributor running the script locally with python-yq gets wrong/confusing results.

**Fix:** Add a guard (`yq --version | grep -qi mikefarah || { echo "requires mikefarah/yq"; exit 1; }`) and note the dependency in the script header.

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,115 @@
...
+        run: |
+          if git diff --quiet -- recipes/evidence/; then
+            echo "No pointer changes to commit (nothing to sign)."
+            exit 0
+          fi
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add recipes/evidence/
+          git commit -m "chore(evidence): sign pending evidence pointers [skip ci]"
+          git push
```
njhensley: **🟡 Minor — Bare `git push` with no rebase/retry → non-fast-forward failure leaves registry/pointer inconsistent**

If the contributor pushes to the branch during the ~minutes-long run, the final push is rejected non-fast-forward and the job fails *after* the bundle was already signed in the registry.

**Blast radius:** Re-running is idempotent for signing, but the user is left with a confusing red run and a transient registry/pointer mismatch. Lower severity given single-actor manual dispatch + per-branch concurrency serialization.

**Fix:** `git pull --rebase --autostash` before push, or a small fetch-rebase-retry loop with a clear "pull and re-dispatch" message. (Also resolved by the `peter-evans` path.)

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,115 @@
...
+# id-token:write / contents:write grants act on the fork, not upstream —
+# there is no pull_request_target / untrusted-ref exposure (it is manual
+# dispatch on the fork's own code).
+#
+# The fork's aicr-evidence registry package must be PUBLIC: the pre-sign
+# pull (and the upstream evidence gate) fetch the bundle, and a private
+# package returns HTTP 403. The signing step fails with a clear message
+# pointing at this when it can't pull.
+
+name: "Recipe Evidence: Sign"
```
njhensley, line 55: **🟡 Minor — Three names for one thing: file `evidence-publish.yaml`, workflow "Recipe Evidence: Sign", behavior = sign-only**

The file is named `evidence-publish.yaml`, the Actions UI shows "Recipe Evidence: Sign", the workflow only signs (never publishes/pushes a bundle), and the doc page is titled "Publishing recipe evidence." The doc directs readers to "the **Recipe Evidence: Sign** workflow (`.github/workflows/evidence-publish.yaml`)" — so anyone grepping by filename for the signing workflow won't find it.

**Blast radius:** Maintainer/contributor confusion when locating or referencing the workflow.

**Fix:** Rename the file to `evidence-sign.yaml` to match its `name:` and behavior. If the `publish` filename must stay to reserve room for the future Option B auto-publish leg (per the header), call that out explicitly in the doc once.

docs/contributor/evidence-publishing.md
```diff
@@ -0,0 +1,115 @@
...
+- patches the pointer's `signer` block in place and commits it back to the
+  branch.
+
+It is a clean no-op when there are no unsigned pointers, and it fails with a
+clear message if it cannot pull a bundle (the public-package requirement
+above). Pull the commit it pushes (`git pull`) and your PR now carries a
+fully signed pointer.
+
+> Your fork must have GitHub Actions enabled and allow the `id-token: write`
+> permission (the default for `workflow_dispatch` runs you trigger).
```
njhensley: **🟡 Minor — Doc misstates `id-token: write` as a `workflow_dispatch` default**

"allow the `id-token: write` permission (the default for `workflow_dispatch` runs you trigger)" is false — `id-token: write` is never a default; it is explicitly requested by the workflow's `permissions:` block. The real fork gotcha (which the workflow header itself acknowledges) is that Actions is disabled by default on forks.

**Blast radius:** Misleads a contributor debugging a failed run.

**Fix:** "Your fork must have GitHub Actions enabled (forks have it disabled by default — enable it under your fork's *Actions* tab). The workflow requests `id-token: write` itself; no per-run setting is needed once Actions is on."

docs/index.yml
```diff
@@ -90,3 +90,5 @@ navigation:
         path: contributor/skills.md
       - page: Maintaining AICR
         path: contributor/maintaining.md
+      - page: Publishing Recipe Evidence
```
njhensley, line 93: **🔵 Nitpick — Nav title casing doesn't match the page H1**

Nav uses "Publishing Recipe Evidence" (title case); the page H1 is "Publishing recipe evidence" (sentence case).

**Blast radius:** Cosmetic — not all sibling pages match nav==H1 either ("Skills" → "Claude Skills").

**Fix:** Recommend matching the H1: `- page: Publishing recipe evidence`.

.github/workflows/evidence-publish.yaml
njhensley: **🔵 Nitpick — `[skip ci]` on the commit-back is inert and misleading on a PR branch**

`[skip ci]` does nothing for this `workflow_dispatch`-only workflow. On the commit-back to a PR head it could skip the `pull_request`-triggered merge gate — but the ruleset's `strict_required_status_checks_policy: true` means the required `gate` check must be green on the latest commit, so a skipped run blocks merge (fails closed), not opens a hole.

**Blast radius:** Harmless but misleading on a PR branch.

**Fix:** Drop `[skip ci]` (unnecessary under manual dispatch) or scope it.

.github/scripts/evidence-sign-unsigned.sh
```diff
@@ -0,0 +1,74 @@
...
+  # (ambient OIDC does not prompt, but this is belt-and-suspenders).
+  if "$AICR" evidence sign "$pointer" --yes "${extra_args[@]}"; then
+    signed=$((signed + 1))
+  else
+    echo "::error::failed to sign ${pointer} — is the bundle's registry package public? (a private fork package returns HTTP 403 on the pre-sign pull)"
+    failed=$((failed + 1))
+  fi
+done
+
+echo "skipped=${skipped}"
```
njhensley, line 75: **🔵 Nitpick — `skipped` count is logged to stdout but never written to `$GITHUB_OUTPUT`**

`signed`/`failed` are emitted to both stdout and `$GITHUB_OUTPUT`; `skipped` only to stdout.

**Blast radius:** None today; asymmetry only matters if a future step wants to branch on it.

**Fix:** Add `skipped` to the `$GITHUB_OUTPUT` block for symmetry.

docs/contributor/evidence-publishing.md
```diff
@@ -0,0 +1,115 @@
...
+- discovers every pointer in `recipes/evidence/*.yaml` with an empty
+  `signer` (i.e. unsigned),
+- signs the bundle each one already references using the runner's ambient
+  OIDC token ([`aicr evidence sign`](../user/cli-reference.md#aicr-evidence-sign)),
+- patches the pointer's `signer` block in place and commits it back to the
+  branch.
+
+It is a clean no-op when there are no unsigned pointers, and it fails with a
+clear message if it cannot pull a bundle (the public-package requirement
+above). Pull the commit it pushes (`git pull`) and your PR now carries a
```
njhensley: **🔵 Nitpick — Doc doesn't warn against pushing other work while the workflow runs**

"Pull the commit it pushes (`git pull`)" gives no warning that pushing concurrent commits to the branch during the run will race the bot push (and trip the bare-push finding).

**Blast radius:** First-timer trap; possible non-fast-forward race.

**Fix:** Append "(don't push other commits to the branch while the workflow runs)."

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,115 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16  # v6.5.0
+        with:
+          go-version: ${{ steps.versions.outputs.go }}
+          cache: false
+
+      - name: Build aicr
+        env:
+          GOFLAGS: -mod=vendor
+        run: go build -o ./bin/aicr ./cmd/aicr
```
njhensley: **🔵 Nitpick — No `go mod verify` before building the signer binary**

The signing binary is built from the fork-controlled vendored tree with no integrity check.

**Blast radius:** Fork-scoped, defense-in-depth only — but this binary's provenance is the crux of an evidence-signing workflow.

**Fix:** Consider adding `go mod verify`, and ensure the upstream evidence gate pins the accepted Fulcio identity (out of scope for this diff).

njhensley (commented):
📋 **Multi-persona review — PR #1451**

**Method:** 3 independent persona passes (Security/CI supply-chain · Bash/Reliability · Docs/DX), each finding independently confirmed, refuted, or re-tiered by a senior meta-reviewer against the resolved code (CLI source on `main`, the repo's `required_signatures` ruleset, and GitHub Actions `if:` semantics). Line links pinned to head `00228d4`.

## Overall assessment — Approve with comments (resolve the two 🟠 first)

The two-phase split — push the network-light leg locally with `--no-sign`, do the Fulcio-bound leg in CI where Sigstore is reachable — is a sound design, and the `workflow_dispatch`-on-the-fork framing genuinely defuses the worst class of CI risk (no `pull_request_target`, no untrusted-ref exposure). **No hard blocker.** Two Major issues are merge-relevant: a workflow step whose `if:` silently defeats the script's own fail-closed exit, and an *unsigned* bot commit-back that contradicts both the doc's "fully signed pointer" wording and the repo's `required_signatures` ruleset. The remaining items are robustness, a real naming-confusion trap, and polish. Notably, the scariest candidate — a signed-in-registry-but-un-patched-pointer race — was **refuted** on inspection of `SignExisting`.

- **No signed-in-registry-but-un-patched-pointer race.** The highest-proposed-tier candidate was **refuted**: `SignExisting` calls `ValidateSignablePointer` (rejects a non-nil signer with `ErrCodeConflict`), and on success sets `Attestations[0].Signer` and writes the file. A successful sign therefore *always* mutates the pointer, so `git diff --quiet` cannot find "nothing to commit" after a real `--no-sign` pointer is signed; an idempotent re-sign returns non-zero and counts as `failed`, not `signed`.
- **`#aicr-evidence-sign`, `#aicr-evidence-publish`, `#aicr-validate`, and `../design/007-recipe-evidence.md` all resolve** on current `main` (lychee-safe). CLI flag claims are accurate: `--no-sign` on both `validate` and `evidence publish`; positional bundle dir; `evidence verify` reports unsigned as a non-failing "pending signature".
- **`concurrency` with `cancel-in-progress: false`** is correct — a mid-sign cancel could orphan a signed bundle from its un-patched pointer.
- **Fork-only `workflow_dispatch` trust model** is sound: no `pull_request_target`, no untrusted-ref injection; `actions/checkout` and `actions/setup-go` are SHA-pinned; `timeout-minutes: 20` set.

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major | 2 | commit-step `if:` defeats fail-closed exit · unsigned bot commit vs `required_signatures` + doc wording |
| 🟡 Minor | 6 | yq fail-open · persisted token across build · yq unpinned/local-incompat · bare `git push` · publish-vs-sign naming · `id-token` "default" misstatement |
| 🔵 Nitpick | 5 | nav casing · inert `[skip ci]` · `skipped` not in output · concurrent-push warning · `go mod verify` |

**Recommendation:** Approve with comments. No hard blocker. Fix the two 🟠 (one-line `success() &&` guard; commit-back via `peter-evans/create-pull-request`, which also clears the persisted-token and bare-push items) before merge — the second could escalate to 🔴 if `required_signatures` blocks the PR merge. The 🟡/🔵 items are safe to address in the same pass or defer in writing.

<sub>Generated via a multi-persona + adversarial-meta-reviewer pass; findings independently re-derived from resolved code before surviving.</sub>

.github/scripts/evidence-sign-unsigned.sh
```diff
@@ -0,0 +1,74 @@
...
+signed=0
+failed=0
+skipped=0
+
+shopt -s nullglob
+for pointer in recipes/evidence/*.yaml; do
+  # signer absent => unsigned (the --no-sign state). A signed-without-Rekor
+  # pointer still has a signer block, so it is correctly skipped.
+  signer=$(yq eval '.attestations[0].signer // "null"' "$pointer" 2>/dev/null || echo "null")
+  if [[ "$signer" != "null" ]]; then
```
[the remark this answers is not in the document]
mchmarny: Fixed — the signer check now fails closed: a yq error (invalid YAML / missing mikefarah yq) is treated as a hard failure (failed++, ::error::) rather than `|| echo null` which conflated it with an unsigned pointer. Smoke-tested with malformed YAML → failed=1.

docs/contributor/evidence-publishing.md
```diff
@@ -0,0 +1,115 @@
...
+> `aicr-evidence` package visibility to **public** under your account's
+> *Packages* settings. Any OCI-1.1 registry works; it just has to be
+> readable.
+
+### 2. Commit the unsigned pointer and push your branch
+
+```shell
+git add recipes/evidence/<slug>.yaml
+git commit -S -m "evidence: <slug> (unsigned; sign in CI)"
+git push
```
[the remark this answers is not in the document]
mchmarny: Fixed — the doc example now uses `git commit -s` (DCO sign-off, required for all contributors) and notes that NVIDIA org members additionally use `-S`, per CONTRIBUTING.md.

.github/workflows/evidence-publish.yaml
mchmarny: Addressed — the commit `if:` now explicitly uses `always() && ...signed \!= "0"`. This is intentional: on partial failure we DO commit the successfully-signed pointers (otherwise the registry/repo split in the sibling comment), while the job still fails overall because the sign step exited non-zero. So a partial failure is never silent.

.github/workflows/evidence-publish.yaml
mchmarny: Addressed in docs — the commit-back is a normal github-actions bot commit and is NOT cryptographically signed; the doc now says so explicitly and notes the squash-merge re-signs under the repo policy. Reworded "fully signed pointer" → "a pointer whose signer block is filled in (the bundle is signed)". Producing an API-verified commit is out of scope for this fork helper.

.github/scripts/evidence-sign-unsigned.sh
mchmarny: Fixed — see the fail-closed yq handling above (a parse error is now a hard failure, not "unsigned").

.github/workflows/evidence-publish.yaml
mchmarny: Fixed — checkout now uses `persist-credentials: false`, so the push token is not stored in .git/config across the fork-controlled `go build`. The commit-back pushes with the token supplied inline (`https://x-access-token:${GH_TOKEN}@…`) only at push time.

.github/scripts/evidence-sign-unsigned.sh
mchmarny: Documented — the script header now states it requires the mikefarah `yq` v4 (the same one the rest of the repo assumes) and that kislyuk/python yq is unsupported; GitHub-hosted runners ship mikefarah yq.

.github/workflows/evidence-publish.yaml
mchmarny: Acknowledged — kept the plain push: the workflow is manual `workflow_dispatch` with `cancel-in-progress: false` serializing per branch, so a concurrent push is unlikely, and a non-fast-forward is a clear, recoverable failure (re-dispatch after `git pull`). Documented that in both the commit-step comment and the contributor doc.

.github/workflows/evidence-publish.yaml
mchmarny: Kept the filename `evidence-publish.yaml` because it is the exact path specified in issue #1433s acceptance criteria; the workflow `name:` ("Recipe Evidence: Sign") and its behavior (sign-only) are consistent with each other. The "publish" in the filename refers to the overall two-phase publish flow this completes.

docs/contributor/evidence-publishing.md
mchmarny: Fixed — the doc no longer claims `id-token: write` is a default; it now states the workflow declares the permission in its own `permissions:` block and the fork just needs Actions enabled with OIDC issuance allowed (the default fork setting).

docs/index.yml
mchmarny: Fixed — page H1 changed to Title Case ("Publishing Recipe Evidence") to match the nav entry.

.github/workflows/evidence-publish.yaml
mchmarny: Fixed — removed `[skip ci]` from the commit-back message; the commit should let the evidence gate re-run and verify the now-signed pointer.

.github/scripts/evidence-sign-unsigned.sh
mchmarny: Fixed — `skipped` is now written to `$GITHUB_OUTPUT` alongside signed/failed.

docs/contributor/evidence-publishing.md
mchmarny: Fixed — the doc now warns against pushing other commits to the branch while the workflow runs (the commit-back would otherwise hit a non-fast-forward).

.github/workflows/evidence-publish.yaml
mchmarny: Fixed — the build step now runs `go mod verify` before `go build`, checking the vendored dependency tree against go.sum before building the binary that signs with the runner OIDC identity.

.github/workflows/evidence-publish.yaml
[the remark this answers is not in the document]
mchmarny: Fixed — the commit step now uses `if: always() && steps.sign.outputs.signed \!= "0"`, committing the successfully-signed pointers even when another pointer fails; the job still fails (sign step exited non-zero) so the partial failure is surfaced.

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,134 @@
...
+            exit 0
+          fi
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git add recipes/evidence/
+          # Note: this commit is made with the github-actions bot identity and
+          # is NOT cryptographically signed. It fills in the evidence pointer's
+          # signer block (the bundle signature); the git commit itself relies on
+          # the eventual squash-merge for repo commit-signing policy.
+          git commit -m "chore(evidence): sign pending evidence pointers"
```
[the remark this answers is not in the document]
mchmarny: Fixed — the commit-back now uses `git commit -s`, adding a `Signed-off-by:` trailer matching the github-actions[bot] author so the DCO bot passes on the workflow-generated commit. (It is still not cryptographically signed (-S); repo commit-signing policy is satisfied at squash-merge.)

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,135 @@
...
+# id-token:write / contents:write grants act on the fork, not upstream —
+# there is no pull_request_target / untrusted-ref exposure (it is manual
+# dispatch on the fork's own code).
+#
+# The fork's aicr-evidence registry package must be PUBLIC: the pre-sign
+# pull (and the upstream evidence gate) fetch the bundle, and a private
+# package returns HTTP 403. The signing step fails with a clear message
+# pointing at this when it can't pull.
+
+name: "Recipe Evidence: Sign"
```
njhensley, line 55: 🟡 **Minor (non-blocking) — F8: workflow file/name vs behavior still diverge.** File is `evidence-publish.yaml`, the workflow `name:` is "Recipe Evidence: Sign", and it only signs. Anyone grepping by filename for the signing workflow won't find it. Defensible to keep the `publish` name if it's reserved for the future Option B auto-publish leg (per the header) — if so, a one-line note in the doc pointing that out would close the loop. Leaving as a note, not a blocker.

docs/index.yml
njhensley, line 93: 🔵 **Nitpick (non-blocking) — F10: nav title-case vs page H1.** Nav reads "Publishing Recipe Evidence" (title case); the page H1 is "Publishing recipe evidence" (sentence case). Match the H1: `- page: Publishing recipe evidence`. Pure cosmetic.

njhensley (dismissed):
✅ **Approve — with two non-blocking notes**

Re-reviewed at head `e17b4b3`. The revision resolves **11 of 13** findings from the multi-persona pass, including both 🟠 Majors:

- **F1** (commit-step `if:`) — now `always() && …` with documented intent: partial-failure still commits the pointers that *did* sign so the registry never holds a signed bundle with an unsigned committed pointer, while the job still fails to surface the error. Deliberate, well-reasoned.
- **F2** (unsigned bot commit) — documented that the commit relies on the eventual squash-merge for the repo's commit-signing policy; `[skip ci]` dropped; doc softened from "fully signed pointer" to "a pointer."
- **F4** yq fail-open → fixed (fail-closed on parse error). **F5** `persist-credentials: false` + inline push token. **F6** mikefarah-yq dependency documented. **F7** non-fast-forward documented as recoverable. **F9** `id-token` "default" wording corrected. **F11** `[skip ci]` removed. **F12** `skipped` added to `$GITHUB_OUTPUT`. **F13** concurrent-push warning added. **F14** `go mod verify` added.

Two stragglers remain, both non-blocking (inline) — neither holds up the approve. Nice work on the turnaround.

**Tier legend:** 🟡 Minor · 🔵 Nitpick

.github/workflows/evidence-publish.yaml
mchmarny: Added a one-line note in the contributor doc: the file is named `evidence-publish.yaml` because it anticipates the future auto-publish (Option B) leg documented in the workflow header; today it only signs. Keeping the filename so it matches issue #1433s acceptance path.

docs/index.yml
mchmarny: Already aligned — the page H1 is "# Publishing Recipe Evidence" (Title Case) and the nav entry is "Publishing Recipe Evidence" (Title Case) on the current commit; they match. (The earlier sentence-case H1 was changed in a prior push.)

.github/workflows/evidence-publish.yaml
```diff
@@ -0,0 +1,135 @@
...
+          cache: false
+
+      - name: Build aicr
+        env:
+          GOFLAGS: -mod=vendor
+        run: |
+          # Verify the vendored dependency tree against go.sum before building
+          # the binary that will sign with the runner's OIDC identity.
+          go mod verify
+          go build -o ./bin/aicr ./cmd/aicr
```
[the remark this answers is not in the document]
mchmarny: Verified not needed — `go build -o ./bin/aicr` creates the `./bin` parent directory itself (confirmed with `rm -rf ./bin` then build → `./bin` created). The repos existing recipe-evidence.yaml gate builds the same way without a mkdir. So no `mkdir -p ./bin` step is required.

njhensley (approved):
✅ **Approve** — re-reviewed at head `96130c1`.

All 13 findings from the multi-persona pass are now resolved, including the two stragglers from the prior round:

- **F8** (publish-vs-sign naming) — doc now explains the filename: *"The filename says publish because it anticipates a future auto-publish leg (see the workflow header); today it only signs."* Exactly the note that closes the loop.
- **F10** (nav vs H1 casing) — H1 aligned to Title Case to match the nav.

Bonus hardening since the last revision: the bot commit-back now does `git commit -s` (DCO sign-off matching the `github-actions[bot]` author), with the rationale documented inline; step 2 of the doc mirrors it. Verified against `CONTRIBUTING.md` — external contributors `-s`, NVIDIA org members `-S` — the doc's claim is accurate.

CI is green across Lint, Test, CLI E2E, Security Scan, Fern Check (anchors/lychee), and actionlint; `tests / E2E` is still in progress (merge stays gated on it). Clean approve — nice, thorough iteration. 🟢

PR #1471, opened 2026-06-25
feat(bundler): support dynamic toleration injection for pre-built OCI bundles

### LINKED ISSUE #1371: Support dynamic toleration injection for pre-built OCI bundles
## Problem

When AICR generates a static OCI bundle, GPU node tolerations are baked in at build time via `--accelerated-node-toleration`. Once the image is built there is no mechanism to inject additional tolerations at install time without rebuilding the bundle.

This becomes a blocker when customers add arbitrary taints to their GPU nodes (e.g., `reserved-by:NoSchedule` during application upgrades). The GPU Operator daemonsets — driver, toolkit, device-plugin, DCGM, GFD, etc. — cannot schedule on those nodes, causing driver upgrades to stall.

## Request

Extend the dynamic value mechanism (`--dynamic component:path`) to support node scheduling toleration paths, so that `daemonsets.tolerations` (and equivalent paths registered under `nodeScheduling.accelerated.tolerationPaths` in `recipes/registry.yaml`) can be left as HelmRelease dynamic placeholders at bundle time and filled at install/deploy time.

## Proposed change

- `pkg/bundler/bundler.go:applyNodeSchedulingOverrides()` — when a toleration path is marked as dynamic, emit a placeholder in the HelmRelease rather than baking in the toleration values
- New flag or extension to existing `--dynamic` to mark accelerated/system node toleration paths for install-time injection
- At install time: the operator passes the additional taint(s) as a HelmRelease value override

## Context

This was surfaced from a customer issue on an mk8s-managed cluster (`vclpksojdd-dgxc-k8s-aws-use1-prod`) where `reserved-by:NoSchedule` was added to GPU nodes. The platform needs a consistent way to handle arbitrary customer taints across both AOR-managed (mk8s) and AICR-managed clusters.

Wildcard tolerations (`operator: Exists` with no key) are not an option because they bypass the Skyhook node hardening gate (`skyhook.nvidia.com=runtime-required:NoSchedule`, sync-wave 019).

mohityadav8:
## Summary

Closes #1371

When AICR generates an OCI bundle, GPU node tolerations are baked in at
build time via `--accelerated-node-toleration`. Once built, there was no
way to inject additional tolerations at install time without rebuilding
the entire bundle — a blocker when customers add arbitrary taints to GPU
nodes (e.g. `reserved-by:NoSchedule` during application upgrades).

## What changed

**`pkg/bundler/bundler.go`**

- Added `dynamicPathSetFor(componentName, provider)` — resolves
  `--dynamic` override keys to component names via the registry and
  returns the declared path set for that component.

- In `extractComponentValues()`, dynamic paths are merged into
  `policy.optOut` before `applyNodeSchedulingOverrides` runs. The
  existing `filterPaths()` machinery already skips injection for opt-out
  paths — so no changes to `applyNodeSchedulingOverrides` itself were
  needed. The result: `cluster-values.yaml` gets an empty placeholder
  instead of a baked-in toleration value.

**`pkg/bundler/bundler_test.go`**

- `TestDynamicPathSetFor` — verifies the helper correctly resolves
  override keys to component names and scopes paths per component.

- `TestDynamicTolerationPathExcludedFromBakeIn` — verifies that when
  `--dynamic gpuoperator:daemonsets.tolerations` is set alongside
  `--accelerated-node-toleration`, the toleration is NOT written into
  component values.

## How to use

```bash
aicr bundle \
  --accelerated-node-toleration reserved-by:NoSchedule \
  --dynamic gpuoperator:daemonsets.tolerations \
  --output ./bundle
```

`daemonsets.tolerations` will be an empty placeholder in
`cluster-values.yaml`. At install time, inject the taint:

```bash
helm upgrade --install gpu-operator ./bundle/gpu-operator \
  --set daemonsets.tolerations[0].key=reserved-by \
  --set daemonsets.tolerations[0].effect=NoSchedule
```

## Testing

```bash
go test ./pkg/bundler/... -run "TestDynamic"
```


pkg/bundler/bundler_test.go
```diff
@@ -2472,6 +2472,77 @@ func TestMake_DynamicValuesValidComponent(t *testing.T) {
 	}
 }
 
+// TestDynamicPathSetFor verifies the helper correctly resolves override
+// keys to component names and returns only that component's declared paths.
+func TestDynamicPathSetFor(t *testing.T) {
```
mchmarny: Both new test functions are indented with spaces; gofmt requires tabs. `gofmt -l` flags this file, so `make lint` / `make qualify` will fail the merge gate. Run `gofmt -w pkg/bundler/bundler_test.go` (or `make tidy`) before pushing.

pkg/bundler/bundler.go
```diff
@@ -1096,6 +1106,59 @@ func (b *DefaultBundler) applyNodeSchedulingOverrides(componentName string, valu
...
+// cluster-values.yaml carries an empty placeholder rather than a baked-in
+// value, letting operators supply tolerations at install time without
+// rebuilding the bundle. See #1371.
+func (b *DefaultBundler) dynamicPathSetFor(componentName string, provider recipe.DataProvider) map[string]struct{} {
+	if b.Config == nil || !b.Config.HasDynamicValues() {
+		return nil
+	}
+	registry, err := recipe.GetComponentRegistryFor(provider)
+	if err != nil {
+		return nil
```
mchmarny, line 1192: This fails open against the feature's intent. Both a registry error here and an unknown override key below (`GetByOverrideKey(key)` returning `nil`) silently yield no dynamic paths — so the toleration gets baked in anyway, the opposite of the opt-out the user asked for. A typo'd `--dynamic gpuoperatr:daemonsets.tolerations` would quietly ship a bundle with the toleration hard-coded.

If `--dynamic` keys are already validated upstream (a real component + valid path), this is defensive-only and fine — confirm that's the case. If not, at least `slog.Warn` on the unresolved key so the user isn't misled into thinking the path is dynamic.

mchmarny (changes requested):
Thanks for this — the core approach is clean: merging dynamic paths into `policy.optOut` and letting the existing `filterPaths` machinery skip injection is exactly the right seam, no churn to `applyNodeSchedulingOverrides`, and `optOut` is always initialized so the merge is panic-safe.

Blockers before this can merge:

**gofmt failure (inline)** `pkg/bundler/bundler_test.go` is space-indented; `gofmt -l` flags it, so `make qualify` fails.

One medium concern inline on `dynamicPathSetFor` (silent fail-open on bad `--dynamic` keys), plus two nits:
- "empty placeholder" (PR description + the comment at the `optOut` merge): the path ends up **absent** from values, not set to an empty placeholder — the test correctly asserts absence, but the wording overstates it.
- Worth a one-line note in `docs/user/bundling.md` that `--dynamic` paths are excluded from scheduling bake-in, since that's the user-visible behavior this enables.

None of the substance is in question — once the rebase shrinks the diff and gofmt passes, this should be quick to re-review.

mohityadav8:
@mchmarny `buildDynamicValuesMap` already validates all `--dynamic` keys upstream and returns `ErrCodeInvalidRequest` if `GetByOverrideKey` returns nil ---- so a typo like gpuoperatr fails loudly before extractComponentValues is ever reached. The silent return in dynamicPathSetFor is defence-in-depth only.
That said, added slog.Warn on both the registry-load-failure path and the unresolved-key path so any future gap in upstream validation surfaces immediately rather than silently baking in a toleration. Fixed in latest commit.

mchmarny:
`pkg/bundler/bundler.go:1171:1: File is not properly formatted (gofmt)`

mchmarny:
Make sure that `make qualify` passes locally.

mchmarny (approved):

PR #104, opened 2026-02-12
ci: add GPU smoke test workflow using nvkind

dims:
## Summary

Add a GPU smoke test CI workflow that validates real GPU access on self-hosted T4 runners and tests `eidos snapshot --deploy-agent` end-to-end.

Also adds a `--require-gpu` flag to `snapshot` and `validate` commands for CDI environments where GPU devices are only injected when explicitly requested.

## Motivation / Context

We need CI coverage to ensure GPU detection and the deploy-agent workflow function correctly on real hardware. This workflow runs on NVIDIA's self-hosted GPU runners via the copy-pr-bot trigger pattern.

Related: N/A

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Build/CI/tooling

## Component(s) Affected

- [x] CLI (`cmd/eidos`, `pkg/cli`)
- [x] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [x] Core libraries (`pkg/errors`, `pkg/k8s`)
- [x] Other: `.github/workflows/gpu-smoke-test.yaml`, `deployments/eidos-agent/2-job.yaml`

## Implementation Notes

**CI Workflow (`gpu-smoke-test.yaml`):**
- Runs on `linux-amd64-gpu-t4-latest-1` self-hosted runner via copy-pr-bot push trigger
- Scheduled to run 4x daily (every 6 hours)
- Creates GPU-enabled kind cluster using nvkind with CDI mode
- Installs GPU Operator (driver/toolkit disabled, NFD enabled) for device plugin + node labeling
- Validates GPU access with standalone nvidia-smi pod
- Builds eidos image with ko, loads into kind, runs `eidos snapshot --deploy-agent --require-gpu`
- Validates snapshot output contains T4 GPU data using yq
- Collects debug artifacts on failure

**`--require-gpu` flag:**
- Added to both `snapshot` and `validate` commands
- When set, adds `nvidia.com/gpu: 1` to the agent pod's resource limits
- Required in CDI environments (e.g., kind with nvkind) where GPU devices and nvidia-smi are only injected by the container runtime when explicitly requested
- On bare metal, not needed — privileged+hostPID gives direct access to `/dev/nvidia*`
- Also available via `EIDOS_REQUIRE_GPU` env var

**Key patterns from NVIDIA/OSMO:**
- CDI mode: `nvidia-ctk runtime configure --cdi.enabled` with `accept-nvidia-visible-devices-as-volume-mounts=true`
[...]

PR #1474, opened 2026-06-25
feat(network): NetworkTopology collector + l8k library integration

almaslennikov:
## Summary

Adds a `NetworkTopology` Measurement collector that ingests k8s-launch-kit (l8k) cluster-config — either from a file (`--cluster-config <path>`) or via live discovery (`--discover-network`) — and emits the canonical `NetworkTopology` shape documented in `docs/integrator/measurement-api.md`.

Re-submission of #1472 (closed). The blocker on the prior PR was `verify-licenses` rejecting MPL-2.0 `hashicorp/go-multierror`, transitively pulled in via Helm SDK from `pkg/networkoperatorplugin`. l8k PR #104 (merged as `a99c9684`) split discovery into a Helm-free sub-package `pkg/networkoperatorplugin/discovery`; this PR pins that merge and migrates the import. The MPL-2.0 dep is gone and the aicr vendor tree shrinks by ~950 files.

## Motivation / Context

aicr's snapshot model gained a `NetworkTopology` Measurement type in Phase 1 (shape contract in [`docs/integrator/measurement-api.md`](https://github.com/NVIDIA/aicr/blob/main/docs/integrator/measurement-api.md)) but nothing populated it. This PR wires the producer side: per-hardware-group PFs (PCI/PSID/partNumber/RDMA-device/rail/NUMA), per-group capabilities (sriov/rdma/ib), `linkType`, kernel-module lists, machine/GPU type, recommended firmware — all captured by l8k discovery and translated into one Measurement.

Fixes: N/A
Related: closes #1472 (this PR re-submits with the verify-licenses gate unblocked); l8k PR #104 (the upstream split that made this possible)

## Type of Change

## Component(s) Affected

- [x] CLI (`cmd/aicr`, `pkg/cli`) — new `--cluster-config` / `--discover-network` flags
- [x] Core libraries (`pkg/errors`, `pkg/k8s`) — added `client.ResolveKubeconfigPath` helper

## Implementation Notes

**New package `pkg/collector/network`** with two source modes:
- `ClusterConfigPath` — parse `cluster-config.yaml` from disk; no cluster contact.
- `DiscoverNetwork` — call l8k's live `Discover` against the cluster. NOT read-only: writes `nvidia.kubernetes-launch-kit.{machine,gpu}` labels and patches `NicClusterPolicy` via server-side-apply.

When both are set, file path wins. When neither is set, the collector is "inactive" — `Collect` returns `(nil, nil)`. Multi-group input is rejected with `ErrCodeInvalidRequest` + `groupCount` context; multi-group support is a future iteration.

**Translator** converts `l8kconfig.LaunchKitConfig` → `NetworkTopology` Measurement with `identity / capabilities / pfs / kernel-modules` subtypes. Per-PF Context carries `model`, `connectedGPU`, `gpuProximity` alongside `pciAddress`, `deviceID`, `psid`, `partNumber`, `rdmaDevice`, `networkInterface`. Per-PF Data carries `rail`, `numaNode`, `traffic`.

**Cross-repo dependency.** `go.mod` pins `github.com/nvidia/k8s-launch-kit v0.0.0-20260625130457-a99c9684627e` — the merge of l8k PR #104. No local `replace`. The discovery sub-package is Helm-free by construction; `go list -deps ./pkg/collector/network | grep -E 'helm\.sh|hashicorp/go-multierror'` is empty.

**CLI: `aicr snapshot --cluster-config <path> | --discover-network`** with env-var bindings (`AICR_CLUSTER_CONFIG_PATH`, `AICR_DISCOVER_NETWORK`) for in-pod Job mode. In local mode (`AICR_AGENT_MODE=true`), the CLI seeds `KUBECONFIG` from `--kubeconfig` before calling Measure so the l8k client and aicr's K8s client resolve the same path.

**Job-mode RBAC for `--discover-network`** lives in `pkg/k8s/agent/rbac.go`: `ensureClusterRole` conditionally appends `discoverNetworkClusterRules()` (CRDs, namespaces, daemonsets, serviceaccounts/configmaps, roles/rolebindings, pods/exec, nodes/patch, `configuration.net.nvidia.com/nicdevices`, `mellanox.com/nicclusterpolicies`) when `Config.DiscoverNetwork` is true. Non-network snapshots keep the existing minimal-priv 3-rule ClusterRole.

**Shared helper `pkg/k8s/client.ResolveKubeconfigPath`** — lifted the explicit-path → `KUBECONFIG` env → `~/.kube/config` → in-cluster chain out of `BuildKubeClient` so aicr's K8s client and the l8k discovery client resolve kubeconfigs identically.

**Snapshotter hardening:** `collectSafe` now skips nil Measurements (required for the network collector's inactive-mode contract; harmless to existing collectors).

## Testing

```bash
make qualify   # full gate
```

- `go test -race ./pkg/collector/network/...` — all tests pass; 9 cases in `translate_test.go`, 6 cases in `network_test.go` covering inactive contract / file-mode success / missing path / invalid YAML / multi-group rejection / file-path-wins-over-discover.
- `go test -race ./pkg/k8s/agent/...` — the new discover-network RBAC subtest pins that the extra rules (mellanox.com/nicclusterpolicies patch, pods/exec create, nodes/patch) land when `Config.DiscoverNetwork` is set.
- `golangci-lint run ./pkg/...` — 0 issues.
- `go list -deps ./pkg/collector/network | grep -E 'helm\.sh|hashicorp/go-multierror|containerd|lib/pq|jmoiron/sqlx'` — empty.

**End-to-end discovery** smoke-tested against a 4-node ThinkSystem SR675-V3 RTX-PRO-6000 cluster — `AICR_AGENT_MODE=true aicr snapshot --discover-network` writes a NetworkTopology Measurement with the contracted shape including `gpuProximity` / `connectedGPU` per-PF Context overlaid from l8k preset topology. The output was captured in the closed PR #1472's commit message; behaviour is unchanged from there modulo the API migration.

## Risk Assessment

- [x] **Medium** — New collector + new RBAC + cross-repo dependency on a freshly-merged l8k API. Behind opt-in CLI flags (`--cluster-config` / `--discover-network`); off-by-default for existing snapshot users. Discovery is cluster-mutating (writes node labels + patches `NicClusterPolicy`) but only when the user opts in.

**Rollout notes:** Backwards compatible for existing snapshot users — no new flags required. Job-mode `--discover-network` requires the new RBAC rules in `ensureClusterRole`; the ClusterRole is recreated on each snapshot Job deploy so existing deployments pick it up automatically. The l8k pseudo-version is the canonical merged-main commit, not a branch tip.

## Not in this PR

- ConfigMap mounting for Job-mode `--cluster-config` — file source works in local mode (`AICR_AGENT_MODE=true`) today.
- Multi-group NetworkTopology emission. The single-group enforcement is in `toMeasurement`; lifting it requires multi-Measurement-per-Type consumer rewrites in constraints extractor, recipe validation, diff indexing, and fingerprint — tracked as a follow-up phase.

## Checklist

- [x] Linter passes (`golangci-lint run ./pkg/...`)
- [x] I updated docs if user-facing behavior changed (`docs/integrator/measurement-api.md` updated in the schema-foundations work; this PR's new flags are CLI-only)


almaslennikov:
Addressing CodeRabbit feedback in `df3d81d6` (amend of `a24c8453`, force-pushed with `--force-with-lease`).

**Fixed (this revision):**

| # | Path | Status |
|---|---|---|
| 1 | `pkg/cli/snapshot.go:557` | ✅ Explicit `--kubeconfig` now wins over inherited `KUBECONFIG` — dropped the `== ""` guard. Cluster-mutating discovery should never silently target a stale env-pointed cluster when the user explicitly specified `--kubeconfig`. |
| 3 | `pkg/collector/network/network.go:109` | ✅ Discriminate `fs.ErrNotExist` (→ `ErrCodeNotFound`) from `fs.ErrPermission` (→ `ErrCodeUnauthorized`) from everything else (→ `ErrCodeInternal`). Applied to both `os.Stat` and `os.Open`. |
| 4 | `pkg/collector/network/network.go:122` | ✅ `os.Stat` now runs first and rejects oversize input with `ErrCodeInvalidRequest` + size/limit context. The `io.LimitReader` stays as a defense-in-depth bound on parser memory. |
| 5 | `pkg/collector/network/network.go:163` | ✅ `context.DeadlineExceeded` / `context.Canceled` from `Discover` are now surfaced as `ErrCodeTimeout`. Only non-context errors get `ErrCodeUnavailable`. The 10-min `CollectorNetworkTimeout` now reads as a timeout, not a cluster-availability problem. |
| 7 | `pkg/k8s/client/client.go:85` | ✅ `ResolveKubeconfigPath` returns `""` when `KUBECONFIG` env contains a path separator (`:` / `;`) rather than handing a multi-file value to `BuildConfigFromFlags("", v)` (which treats v as one explicit path and breaks the merge). Empty path lets clientcmd's default loading rules honor `KUBECONFIG` natively. |
| 8 | `pkg/snapshotter/agent.go:163` | ✅ `deployAndWaitForResult` now rejects `ClusterConfigPath != ""` up front with a clear `ErrCodeInvalidRequest` message pointing users at `--discover-network` (Job mode) or `AICR_AGENT_MODE=true` (local mode). ConfigMap-backed mounting is tracked as a follow-up — the previous behavior silently forwarded an unreadable host path. |
| 10 | `pkg/collector/network/translate.go:144` | ✅ `identity.context.linkType` is now emitted unconditionally — empty string is the documented "unknown fabric" sentinel per `docs/integrator/measurement-api.md`. |

**Deferred to follow-up (not blocking this PR):**

| # | Path | Reason |
|---|---|---|
| 2 | `pkg/collector/network/network_test.go:163` | Table-driven refactor is stylistic. Current per-case tests use distinct error-code assertions and `errors.As` matchers per case; collapsing them into one table risks losing the per-case context strings that make failures easy to triage. Open to doing it in a follow-up if the team prefers. |
| 6 | `pkg/collector/network/translate_test.go:73` | Optional PF context coverage (model / connectedGPU / gpuProximity). Worth adding but a test gap, not a correctness regression. Tracked. |
| 9 | `pkg/snapshotter/snapshot_test.go:131` | Mock shape needs an "inactive" override to exercise `collectSafe`'s nil-skip branch. Same — test gap, not correctness. Tracked. |
| 11 | `pkg/k8s/agent/deployer_test.go:224` | ClusterRole upgrade-path coverage. `ensureClusterRole` already handles the `AlreadyExists` → `Update` path; the production code uses it on every redeploy. Will add the explicit test case in a follow-up. |
| 12 | `pkg/k8s/agent/rbac.go:337` | Scoping the bootstrap-namespace resources (daemonsets, serviceaccounts, configmaps, roles, rolebindings) into a namespace-scoped `Role`/`RoleBinding` for `nvidia-k8s-launch-kit` is the right long-term shape. It's a non-trivial restructure: aicr would need to pre-create the namespace + namespaced Role before the snapshot Job runs (today l8k creates the namespace during discovery). Worth a dedicated PR. The current ClusterRole still beats cluster-admin and is gated behind `--discover-network`. |

CI: build / lint / test green locally. Re-pushed, awaiting fresh CI cycle.

pkg/cli/snapshot.go
```diff
@@ -383,6 +395,18 @@ func snapshotCmdFlags() []cli.Flag {
 			Sources:  cli.EnvVars("AICR_LIMITS"),
 			Category: catAgentDeployment,
 		},
+		&cli.StringFlag{
+			Name:     "cluster-config",
```
mchmarny, line 399: These two new user-facing flags (`--cluster-config`, `--discover-network`) aren't reflected in `docs/user/cli-reference.md`. The `aicr snapshot` flag table there documents every other snapshot flag (`--os`, `--runtime-class`, `--max-nodes-per-entry`, etc.), so this drifts. Per the project rule, a new CLI flag should update `docs/user/cli-reference.md` in the same PR — add the two rows to that table.

pkg/k8s/client/client.go
```diff
@@ -59,6 +59,51 @@ type cachedPathClient struct {
...
+	}
+	if env := strings.TrimSpace(os.Getenv("KUBECONFIG")); env != "" {
+		// A KUBECONFIG holding multiple files (':'-separated on
+		// Unix, ';'-separated on Windows) is a clientcmd loading-rules
+		// concept — feeding the raw multi-path string into
+		// BuildConfigFromFlags would treat the whole value as one
+		// explicit path. Defer to clientcmd's loading rules in that
+		// case by returning "".
+		if !strings.ContainsAny(env, ":;") {
+			return env
```
mchmarny: The comment says returning `""` here "lets the caller's clientcmd loading rules (which honor KUBECONFIG natively when no explicit path is supplied) merge the files according to standard precedence." That's true for l8k's `kubeclient.New`, but not for the in-repo caller `BuildKubeClient`: a `""` return routes straight to `rest.InClusterConfig()` (L156), never to clientcmd loading rules. So a multi-file `KUBECONFIG` run locally now fails with `failed to get in-cluster config` (and in-pod it silently ignores `KUBECONFIG` in favor of the SA token) rather than merging.

Not a regression of a previously-working path (the old code passed the joined value to `BuildConfigFromFlags` and also errored), but the comment overstates the outcome for this consumer. Suggest tightening the wording to scope the "clientcmd merges" claim to clientcmd-loading-rules callers, distinct from `BuildKubeClient`.

pkg/k8s/agent/rbac.go
```diff
@@ -277,3 +287,73 @@ func (d *Deployer) deleteClusterRoleBinding(ctx context.Context) error {
...
+			Verbs:     []string{verbGet, verbCreate, verbDelete},
+		},
+		{
+			APIGroups: []string{rbacAPIGroup},
+			Resources: []string{"roles", "rolebindings"},
+			Verbs:     []string{verbGet, verbCreate, verbDelete},
+		},
+		{
+			APIGroups: []string{""},
+			Resources: []string{"pods/exec"},
```
mchmarny, line 340: nit / operator-awareness: `pods/exec` create combined with `roles`/`rolebindings` create/delete and `namespaces` create/delete is cluster-scoped and effectively privilege-escalation-capable (an exec into any pod, or a self-granting rolebinding). It's gated behind opt-in `--discover-network` and thoroughly documented here and in the PR body, so this isn't blocking — just flagging that these are the broadest rules in the ClusterRole and worth keeping visible to operators reviewing the snapshot Job's footprint. If l8k always bootstraps into a single fixed namespace, scoping the roles/rolebindings/pods-exec grant to a namespaced Role there would shrink the surface, but that's an l8k-side change, not this PR.

mchmarny (commented):
Solid, well-engineered PR — the new `network` collector follows existing collector patterns closely, uses `pkg/errors` codes + context timeouts throughout, bounds the file read (`os.Stat` + `LimitReader`), fails closed on parse/garbage-env, and has a clean inactive-collector (`nil, nil`) contract that the snapshotter now honors. Test coverage is good (inactive/file-success/missing/invalid/multi-group/file-wins + the RBAC discovery subtest), and the MPL-2.0 blocker from #1472 is genuinely resolved via the Helm-free l8k discovery sub-package. CI is green (Lint, Test, Security Scan, verify-licenses, CLI E2E, GPU Smoke).

A few non-blocking notes inline:
- **Docs (medium):** the two new CLI flags aren't added to the `aicr snapshot` table in `docs/user/cli-reference.md`.
- **Comment accuracy (low):** the multi-file `KUBECONFIG` rationale in `ResolveKubeconfigPath` overstates the outcome for the `BuildKubeClient` caller (it goes to in-cluster config, not clientcmd loading rules).
- **Operator awareness (nit):** the discovery ClusterRole's `pods/exec` + `roles`/`rolebindings` grants are the broadest in the role — opt-in and well-documented, just flagging.

Nothing blocks merge. The doc table update is the one I'd ask for before merge; the other two are optional.

almaslennikov:
Addressing the new review (mchmarny + CodeRabbit re-review of `475e337d`) in `53c103ee`.

**@mchmarny:**

1. **`docs/user/cli-reference.md` rows for `--cluster-config` / `--discover-network`** — ✅ added in this revision. Both flags now appear in the `aicr snapshot` flag table with their full Usage text, mutual-exclusion behavior, and the local-mode-vs-Job-mode constraints called out.
2. **`pkg/k8s/client/client.go` doc-comment scoping** — ✅ tightened. The comment now explicitly scopes the "clientcmd loading rules merge multi-file `KUBECONFIG`" claim to clientcmd-loading-rules callers (l8k's `kubeclient.New("")`, anything using `NewNonInteractiveDeferredLoadingClientConfig`), and explicitly notes that this package's own `BuildKubeClient` falls through to `InClusterConfig()` on empty path — so a multi-file `KUBECONFIG` run locally through `BuildKubeClient` is not supported (and wasn't before this change either).
3. **`pkg/k8s/agent/rbac.go` operator-awareness on pods/exec + namespaces+roles+rolebindings** — agreed, non-blocking. Will tighten the scope when l8k confirms its bootstrap namespace is stable (it has been `nvidia-k8s-launch-kit` for several releases now); right now the rules are intentionally cluster-scoped because aicr can't pre-create the bootstrap namespace before l8k discovery runs. Tracking as a follow-up.

**CodeRabbit new findings on `475e337d`:**

| Path | Status |
|---|---|
| `pkg/cli/snapshot.go:403` (Minor) — Usage text didn't mention `--cluster-config` is local-only | ✅ Usage now reads "Local agent mode only (AICR_AGENT_MODE=true) — Job mode rejects this flag with INVALID_REQUEST..." |
| `pkg/cli/snapshot.go:559` (Major) — env not unset on negative case → could inherit stale `AICR_DISCOVER_NETWORK` | ✅ Both `AICR_CLUSTER_CONFIG_PATH` and `AICR_DISCOVER_NETWORK` are now `os.Setenv` on positive, `os.Unsetenv` on negative, so the Measure rebuild sees exactly the resolved CLI state |
| `pkg/k8s/agent/types.go:56` (Minor) — `AgentConfig.ClusterConfigPath` doc-comment confused local-mode with Job-mode | ✅ Comment rewritten — now explains Job-mode rejects this with `ErrCodeInvalidRequest`, file ingestion is local-mode-only |
| `pkg/collector/network/network_test.go:163` (Trivial) — no test for the MaxClusterConfigBytes rejection | ✅ Added `TestCollect_FileMode_OversizeRejected` — writes a `defaults.MaxClusterConfigBytes+1` file and asserts `ErrCodeInvalidRequest` |
| `pkg/collector/network/translate_test.go:40` (Trivial) — no test for the empty-`linkType` sentinel | ✅ Added `TestToMeasurement_IdentitySubtype_UnknownLinkType` — sets `LinkType = ""` on the fixture and asserts the key is present with empty value |
| `pkg/k8s/client/client.go:99` (Major, same as mchmarny #2) — multi-file KUBECONFIG scope claim | ✅ Same doc-comment fix above |

CodeRabbit also re-flagged several items I already addressed in `df3d81d6` (it appears to be re-running against the latest commit and reposting). For the record:

| Path | Status (already fixed) |
|---|---|
| `pkg/cli/snapshot.go:559` #1 kubeconfig override | ✅ fixed in df3d81d6 |
| `pkg/collector/network/network.go:216` #5 ctx-cancel | ✅ fixed in df3d81d6 |
| `pkg/snapshotter/agent.go:178` #8 Job-mode reject | ✅ fixed in df3d81d6 |
| `pkg/collector/network/network.go:111` (goconst) | ✅ fixed in 475e337d (ctxKeyPath const) |

Still deferred to follow-up:
- `pkg/collector/network/network_test.go` table-driven refactor (CR #2) — separate cases, distinct error-code assertions per case
- `pkg/k8s/agent/deployer_test.go:223` table-driven RBAC checks (new CR Trivial)
- `pkg/snapshotter/snapshot_test.go:131` mock-shape change for inactive network path (CR #9)
- `pkg/k8s/agent/deployer_test.go:224` ClusterRole upgrade-path test (CR #11)
- `pkg/k8s/agent/rbac.go:337` namespace-scoping (CR #12 / mchmarny #3)
- `pkg/collector/network/translate_test.go:73` optional PF context coverage (CR #6) — the new `TestToMeasurement_IdentitySubtype_UnknownLinkType` covers the linkType part; full model/connectedGPU/gpuProximity matrix is a separate add

CI: build / lint / test green locally. Re-pushed, awaiting fresh CI cycle.

almaslennikov:
Addressing the 3 new CodeRabbit findings on `53c103ee` in `d091f52a` (signed amend, force-pushed).

- **`pkg/cli/snapshot.go:572` [Major]** — agent-mode bypass dropped the pre-built factory and lost CLI-only options (`--os`, `--max-nodes-per-entry`) that have no env source. Fixed by removing `ns.Factory = nil` and the redundant env-mirroring: `factory` at line 504 already carries every CLI-resolved option (cluster-config-path, discover-network, kubeconfig, os, max-nodes-per-entry) via the `With*` options, and the flags' `cli.EnvVars` Sources already populated `opts` from Job-set env before this point, so reusing `ns.Factory` directly is correct for both the in-pod Job path and the dev-bypass path.
- **`pkg/collector/network/network.go:153` [Major]** — FIFO / device paths passed the size check and blocked on Open. Added an `st.Mode().IsRegular()` gate right after the Stat check, rejecting non-regular files with `ErrCodeInvalidRequest` + the mode in the error context.
- **`pkg/k8s/client/client.go:111` [Major]** — multi-file `KUBECONFIG` was falling through to `~/.kube/config`. Fixed: multi-path env now returns `""` immediately so loading-rules-aware callers (l8k's `kubeclient.New("")`, anything using `NewNonInteractiveDeferredLoadingClientConfig`) honor the env natively. A user who explicitly set `KUBECONFIG=/a:/b` no longer gets their home kubeconfig silently substituted for the merge they asked for.

Build / lint / test green locally. Re-pushed, awaiting fresh CI cycle.

mchmarny (approved):
All three of my prior comments are addressed in `d1d12aa6` — verified against the current head:

1. **`docs/user/cli-reference.md`** — both `--cluster-config` and `--discover-network` rows added to the `aicr snapshot` table with full usage, mutual-exclusion, and local-mode-vs-Job-mode constraints. ✅
2. **`pkg/k8s/client/client.go`** — doc comment now correctly scopes the "clientcmd merges" claim to loading-rules-aware callers and explicitly distinguishes `BuildKubeClient` (falls through to `InClusterConfig()` on empty path). Behavior matches: multi-file `KUBECONFIG` returns `""` instead of silently substituting `~/.kube/config`. ✅
3. **`pkg/k8s/agent/rbac.go`** — non-blocking operator-awareness nit; the broad cluster-scoped grant (pods/exec, roles/rolebindings, namespaces) is now documented per-rule in `discoverNetworkClusterRules()` and at the call site, gated behind `--discover-network`. Namespace-scoping deferred to an l8k-side follow-up, as expected. ✅

Original `verify-licenses` blocker is resolved (MPL-2.0 `go-multierror` gone via the l8k discovery split). LGTM.

Note for merge: branch is `behind` main — rebase before merge, and let the in-flight GPU/E2E/analyze checks finish.

PR #1477, opened 2026-06-25
feat(ci): add testgrid-publish workflow (TG5)

### LINKED ISSUE #1268: TG3 — Processing (upstream workers + AICR-native config-gen + metadata-key contract)
**Parent:** #1263 · **Labels:** `enhancement`, `area/infra` (auto, GKE manifests
under `infra/testgrid/**`) · **Size:** L · **Depends on:** TG1, TG2 (needs a feed to
roll up and the pinned key schema to mirror)

**Goal.** Deploy the upstream Kubernetes TestGrid updater/tabulator/summarizer on
the `eidosx` host cluster and an **AICR-native config generator** that infers the
CSP-first taxonomy from the bucket layout, so the raw `groups/` feed is rolled up
into the processed `grid/`/`tabs/`/`summary/` trees the API consumes — with a
contract test that pins the TG2↔TG3 metadata-key seam.

**Scope.**
- **Deploy the three upstream workers unchanged** as GKE Workload Identity pods (on
  the TG1 host cluster) with `roles/storage.objectAdmin` (the workers SA from TG1):
  updater (30m, `groups/`→`grid/`), tabulator (15m, `grid/`→`tabs/`), summarizer
  (5m, `tabs/`→`summary/`), each with `--confirm`, `--wait`, `--persist-queue`,
  `--json-logs`, resumable `queue/*.json`. Manifests live under
  `infra/testgrid/k8s/` (or Helm/Kustomize as the repo prefers).
- **AICR-native config-gen lives in its own tool: `tools/testgrid-config-gen/`**
  (Go), owned by TG3 — it consumes the **bucket layout** the workers read, not the
  OCI bundle TG2 reads, so it is a distinct artifact from TG2 with its own
  labeler/CODEOWNERS/coverage scope. It lists `gs://<bucket>/groups/`, infers the
  **CSP-first** taxonomy (DashboardGroup=`service`, Dashboard=`accelerator`-`os`,
  TestGroup/Tab=`intent`[-`platform`]), and writes the protobuf `Configuration` to
  `gs://<bucket>/config`. Honor the upstream invariants: **`GcsPrefix` must include
  the bucket name**; each `TestGroup` must set `result_source.gcs_config`; configure
  the AICR column-header set **keyed to the exact metadata key strings TG2 emits**
  (the pinned schema) as the `Header.extra` array; alerting knobs
  (`numFailuresToAlert=1`, `alertStaleResultsHours=168`, `minAcceptableRuns=3`,
  `brokenColumnThreshold=50%`) as sane defaults; `DisableProwjobAnalysis: true` (no
  podinfo.json).
- Run config-gen on a cadence (CI cron or an in-cluster CronJob) so new
  recipes/coordinates appear automatically as the feed grows.
- **TG2↔TG3 metadata-key contract test.** Against a real TG2-produced `started.json`
  fixture, assert that every `Header.extra` `configuration_value` key TG3 configures
  has a matching key emitted by TG2 (and vice versa). A drift fails CI rather than
  silently rendering blank provenance columns.

**Out of scope.** The publisher (TG2); the API/UI (TG4a/TG4b); provisioning the host
cluster (TG1). We do **not** fork the worker source — only their deployment + an
AICR-native config.

**Key files.** New `infra/testgrid/k8s/**` (worker manifests, Workload Identity SA
annotations); new `tools/testgrid-config-gen/` (needs the same `tools/**` labeler
glob TG2 adds; a narrower CODEOWNERS entry is optional); the pinned key schema in
TG6's spec. Upstream behavior reference: `test-result-ingestion-reference.md` §3,
§4, §9.

**Tests.** config-gen unit tests: given a fixture `groups/` listing, assert the
generated `Configuration` has the right DashboardGroup/Dashboard/TestGroup nesting,
`GcsPrefix` includes the bucket, every TestGroup has `result_source.gcs_config`,
the AICR header set matches the pinned keys, and the alerting thresholds match. The
TG2↔TG3 metadata-key contract test (above). A `--dry-run` text-proto golden. An
integration smoke (against an emulated bucket or a staging GCS) that the workers
produce non-empty `grid/`/`tabs/`/`summary/` from a seeded feed, **and that columns
appear in chronological order** for distinct seeded runs. **75%** floor on
`tools/testgrid-config-gen`.

**Acceptance.** The three upstream workers run on the `eidosx` host cluster with
`objectAdmin` and resumable queues; `tools/testgrid-config-gen` infers CSP-first
taxonomy from `groups/` and writes a valid protobuf `config`; the TG2↔TG3 metadata-key
contract test passes; a seeded feed rolls up into `grid/`/`tabs/`/`summary/` with
correct column ordering.

---
_Child of #1263 · staged from `docs/design/010-aicr-testgrid.md` (TG3)._

srao-nv:
## Summary

Adds `.github/workflows/testgrid-publish.yml` — the GitHub Actions workflow that automatically publishes AICR UAT evidence bundles to the TestGrid dashboard after every completed cluster run.

## Motivation / Context

TG5 from the AICR TestGrid epic (#1263). Once TG2 (`tools/testgrid-publish`) merges, this wires it into CI so every UAT run appears as a column in the TestGrid dashboard without manual intervention.

Fixes: #1268
Related: #1263, #1267

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `.github/workflows/testgrid-publish.yml` (new CI workflow)

## Implementation Notes

**Trigger:** `workflow_run` on "UAT GCP" and "UAT AWS" (success **or** failure — failing runs are valid TestGrid columns showing which tests broke). Also supports `workflow_dispatch` for manual backfills.

**Pipeline:**
```
UAT completes → download evidence pointer artifact
             → extract OCI bundle ref + sha256 digest from pointer.yaml
             → GCP WIF auth (publish SA, groups/ prefix write-only)
             → go build ./tools/testgrid-publish
             → testgrid-publish --bundle <oci-ref>@<digest> --bucket aicr-testgrid
```

**Auth:** WIF pool `aicr-testgrid-prod-github` (provisioned in aicr-testgrid Terraform). Workflow filename starts with `testgrid-publish` to satisfy the WIF `attribute_condition` in `wif-publish.tf`.

**Environment:** Defaults to prod (`gs://aicr-testgrid`). Override to staging via `workflow_dispatch(environment=staging)` for testing. WIF pool, publish SA, and bucket are all derived from the environment input.

**Dependency:** Requires TG2 (`tools/testgrid-publish`) — PR #1447.

**Before merge:**
- Apply prod Terraform in `aicr-testgrid` repo (creates `aicr-testgrid-prod-github` WIF pool and `aicr-testgrid-prod-publish` SA with `objectCreator` on `gs://aicr-testgrid/groups/`)
- TG2 (#1447) must be merged first

## Testing

Tested the full pipeline manually today:
```bash
# Built testgrid-publish from source
make testgrid-publish

# Ran against real GKE H100 cluster data
./dist/testgrid-publish \
  --bundle-dir /tmp/tg-bundle \
  --bucket aicr-testgrid-staging \
  --source-class uat

# Verified GCS objects landed in correct order
# Manually ran config-gen + updater + tabulator + summarizer
# Confirmed gke-h100-cos/training tab appeared in TestGrid UI
```

## Risk Assessment

- [x] **Low** — New workflow file only; no existing code modified. Cannot break existing CI. Will be a no-op until prod Terraform is applied.

**Rollout notes:** Workflow is inert until prod WIF pool and SA exist. Safe to merge ahead of Terraform apply.

## Checklist


.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+            if [ -z "${DIGEST}" ] || [ "${DIGEST}" = "null" ]; then
+              echo "::error::Could not extract digest from ${POINTER}"
+              cat "${POINTER}"
+              exit 1
+            fi
+            BUNDLE_REF="${OCI}@${DIGEST}"
+          fi
+          # Validate format: ghcr.io/nvidia/...@sha256:<64 hex chars>
+          # Also guards against newline injection into GITHUB_OUTPUT.
+          if ! echo "${BUNDLE_REF}" | grep -qE '^ghcr\.io/nvidia/[^@]+@sha256:[0-9a-f]{64}$'; then
```
mchmarny: `source_class: community` (line 71) is offered as a dispatch option, but this regex hard-codes `^ghcr\.io/nvidia/`. Committed community evidence bundles live under contributor namespaces — e.g. `recipes/evidence/gb200-eks-ubuntu-training.yaml` points at `ghcr.io/yuanchen8911/aicr-evidence:...`. So a dispatch with `source_class: community` and a real community `bundle_ref` always fails this check. Either the `community` option is dead, or the regex needs to permit non-nvidia namespaces (while still requiring `@sha256:<64hex>`).

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+      # a clear error message rather than a cryptic WIF rejection.
+      - name: Validate producer
+        if: github.event_name == 'workflow_run'
+        env:
+          PRODUCER_REPO: ${{ github.event.workflow_run.repository.full_name }}
+          PRODUCER_BRANCH: ${{ github.event.workflow_run.head_branch }}
+        run: |
+          set -euo pipefail
+          if [ "${PRODUCER_REPO}" != "${GITHUB_REPOSITORY}" ]; then
+            echo "::error::Producer workflow is from '${PRODUCER_REPO}', expected '${GITHUB_REPOSITORY}' — rejecting"
```
mchmarny: The comment says this "rejects workflow_run events from forks," but `github.event.workflow_run.repository.full_name` is always the repo that *hosts* the workflow (the base repo) — never the fork. For a fork-originated run the fork identity is in `workflow_run.head_repository.full_name`, so this `!= GITHUB_REPOSITORY` check can never fire on a fork.

In practice this is moot: UAT only triggers on `workflow_dispatch` + `schedule`, never fork PRs, so no fork can ever produce a `UAT - GCP/AWS` run. But the comment overstates the guard. Either soften the comment to match what the check actually does (base-repo sanity check) or, if fork defense-in-depth is the intent, check `head_repository.full_name`.

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+          set -euo pipefail
+          if [ "${TG_ENV}" = "prod" ]; then
+            CLUSTER="aicr-testgrid"
+          else
+            CLUSTER="aicr-testgrid-${TG_ENV}"
+          fi
+          echo "cluster=${CLUSTER}" >> "${GITHUB_OUTPUT}"
+          echo "bucket=${CLUSTER}" >> "${GITHUB_OUTPUT}"
+          echo "wif_provider=projects/${GCP_PROJECT_NUMBER}/locations/global/workloadIdentityPools/${CLUSTER}-github/providers/github-actions" >> "${GITHUB_OUTPUT}"
+          echo "wif_sa=${CLUSTER}-publish@${GCP_PROJECT_ID}.iam.gserviceaccount.com" >> "${GITHUB_OUTPUT}"
```
mchmarny: `vars.GCP_PROJECT_NUMBER` / `GCP_PROJECT_ID` are consumed unguarded. If either repo var is unset, `wif_provider` silently becomes `projects//locations/global/...` and `wif_sa` becomes `...-publish@.iam.gserviceaccount.com`, so the failure surfaces later as a cryptic WIF rejection. A quick `[ -n "${GCP_PROJECT_NUMBER}" ]`/`[ -n "${GCP_PROJECT_ID}" ]` guard with an `::error::` would fail fast here — same "fail fast with a clear message" rationale you gave for the producer guard.

mchmarny (commented):
Clean, well-structured CI workflow — injection hygiene is solid (all user-controlled values flow through `env`, anchored format regex, digest-pinned bundle refs), actions are SHA-pinned, permissions are least-privilege at both workflow and job level, and `cancel-in-progress: false` correctly protects in-flight GCS writes. Verified against the repo: UAT workflow names, the `uat-*-evidence-pointer-<run_id>` artifact naming, the `attestations[0].bundle.oci/.digest` pointer schema, and the referenced composite actions all line up.

No blocking issues. Three non-critical notes inline: a `community` source_class option that the nvidia-only bundle regex makes unusable (medium), a producer-guard comment that overstates what the repo-name check actually does (low — moot since UAT can't be fork-triggered), and unguarded `vars.GCP_PROJECT_*` lookups (low).

WIP as titled — `go build ./tools/testgrid-publish` will fail until TG2 (#1447) merges, and the workflow is inert until the prod WIF pool/SA exist; both are already documented in the description. Nothing here blocks merge once those land.

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+      # other → aicr-testgrid-<env>
+      - name: Resolve GCP config
+        id: gcp
+        env:
+          TG_ENV: ${{ inputs.environment || 'prod' }}
+          GCP_PROJECT_NUMBER: ${{ vars.GCP_PROJECT_NUMBER }}
+          GCP_PROJECT_ID: ${{ vars.GCP_PROJECT_ID }}
+        run: |
+          set -euo pipefail
+          if [ "${TG_ENV}" = "prod" ]; then
```
njhensley, line 222: **🟠 Major — prod WIF pool/SA naming contradicts the PR's stated Terraform**

The prod branch emits unsuffixed names: WIF pool `aicr-testgrid-github`, SA `aicr-testgrid-publish`, bucket `aicr-testgrid`. The PR body states prod Terraform creates `aicr-testgrid-prod-github` and `aicr-testgrid-prod-publish` (only the bucket matches). CodeRabbit flagged the inverse on an earlier commit; the bucket got reconciled but the pool/SA names still disagree.

**Blast radius:** Once prod Terraform is applied, a name mismatch = silent WIF auth failure on every publish → the dashboard never updates. With TG2 now merged, this is the single most likely rollout-breaker.

**Fix:** Reconcile against the actual `wif-publish.tf` in the aicr-testgrid repo before/with the apply. Drop the `-prod` suffix in Terraform, or set `CLUSTER="aicr-testgrid-prod"` here. (Source: persona panel + CodeRabbit history.)

.github/workflows/testgrid-publish.yml
njhensley: **🟠 Major — `grep -qE` is line-based — does NOT prevent GITHUB_OUTPUT newline injection; the comment is wrong**

`echo "${BUNDLE_REF}" | grep -qE '^...$'` matches per line, so a value like `valid-ref\nfoo=bar` passes (line 1 matches) and the subsequent `echo "bundle_ref=${BUNDLE_REF}" >> GITHUB_OUTPUT` writes both lines, injecting `foo=bar` as a second output. The inline comment claiming this 'guards against newline injection into GITHUB_OUTPUT' is therefore inaccurate.

**Blast radius:** Bounded — exploitable via the workflow_dispatch bundle_ref input by a main-branch actor (or a newline reaching the trusted pointer). Low likelihood, but the guard does not do what its comment says.

**Fix:** Use a whole-string Bash match instead of line-based grep: `if [[ ! "${BUNDLE_REF}" =~ ^ghcr\.io/nvidia/[A-Za-z0-9._/-]+@sha256:[0-9a-f]{64}$ ]]; then ...`. Then fix the comment. (Source: CodeRabbit — corrects an earlier claim in my own panel.)

njhensley (commented):
# 🔍 Reconciled multi-persona review

**Method:** 4 independent persona reviewers (Security/CI-supply-chain · Operability/CI-DX · Correctness · Domain & Architecture) → adversarial senior meta-reviewer re-derived every finding from the resolved code, then **reconciled against all prior reviews on this PR** (CodeRabbit ×5 runs, @mchmarny's human review, pre-merge bot + zizmor). Findings are deduplicated and re-tiered to the cross-source consensus; line links pin to head `37366ea`.

## Overall

This is a carefully built, security-conscious workflow — all three review sources agree the logic is safe (injection hygiene via env-routing, SHA-pinned actions, least-privilege per-job permissions, `persist-credentials: false`, `cancel-in-progress: false`, build-from-vendor) and that the live risks are **operability + the prod-Terraform sequencing**, not the workflow logic. @mchmarny found no blocking issues; this reconciliation concurs — **no hard blocker remains now that TG2 (#1447) has merged.**

**Recommendation: Approve-with-comments**, gated on two operational items before prod traffic:
1. **Reconcile the prod WIF pool/SA naming** against the actual `wif-publish.tf` (the inline 🟠 at the Resolve-GCP-config step) — with TG2 merged this is the most likely rollout-breaker.
2. **Merge after the prod Terraform apply, or add a feature-flag guard** — otherwise every passing UAT goes red at WIF auth until the pool/SA exist.

The other two 🟠 (line-based `grep` + its inaccurate comment; unguarded `vars.GCP_PROJECT_*`) are quick fixes worth landing in this PR.

## TG2 status (confirmed on `main`)

`tools/testgrid-publish` is now merged, and its flags match the workflow exactly — **`--bundle` / `--bucket` / `--source-class`** (`main.go:71-74`), `--source-class` defaulting to `uat` and accepting `community`. So the earlier flag-drift concern is **resolved**; the build step will now succeed.

## Reconciliation with existing reviews

**Already addressed by the author on earlier commits** (CodeRabbit threads marked ✅): producer workflow names `"UAT - GCP"/"UAT - AWS"`, `source_class` wired through env, `bundle_ref` moved to env, `actions: read` added, prod **bucket** un-suffixed.

**Moot for the current PR:** CodeRabbit's `.goreleaser.yaml` release-artifact-integrity / unsigned-tarball findings — the PR no longer touches `.goreleaser.yaml` and the workflow builds from source only. zizmor `concurrency-limits` is resolved (concurrency block present); zizmor's `workflow_run` `dangerous-triggers` is informational and mitigated by the producer guard.

**Process note — linked-issue / scope mismatch:** the PR says `Fixes #1268`, but the pre-merge bot flags #1268 as **TG3** (processing workers + config-gen), not TG5 (publishing) → "Out of Scope" + "Linked Issues" warnings. Worth correcting the linked issue or the `Fixes` target.

**Correction to my own panel:** my Security/meta pass initially called the `grep` regex a newline-injection guard — CodeRabbit is right that it is not (see the inline 🟠). Reconciliation defers to CodeRabbit there.

## Confirmed non-issues (examined, agreed by panel + @mchmarny)

Script-injection hygiene (all event/user values env-routed, never inline `${{ }}` in `run:`), SHA-pinned actions, least-privilege per-job permissions, `persist-credentials: false`, `cancel-in-progress: false` on the upload job, build-from-vendor in the privileged context, and the producer-repo guard running before any credential exchange (defense-in-depth — moot since UAT can't be fork-triggered). Digest construction, yq null/empty handling, the `skip` output coverage, and the top-level `if:` matrix were all traced correct.

## Tier tally (this review)

|:---:|:---:|:---:|:---:|
| 0 | 5 | 7 | 2 |

🤖 Posted via the reviewing-prs-multipersona skill (multi-persona + adversarial meta-review, reconciled across CodeRabbit and human reviews).

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+          fi
+
+      # Resolve environment-specific GCP resource names.
+      # prod  → aicr-testgrid          (no env suffix — matches Terraform cluster_name)
+      # other → aicr-testgrid-<env>
+      - name: Resolve GCP config
+        id: gcp
+        env:
+          TG_ENV: ${{ inputs.environment || 'prod' }}
+          GCP_PROJECT_NUMBER: ${{ vars.GCP_PROJECT_NUMBER }}
```
njhensley, line 213: **🟠 Major — Unguarded vars.GCP_PROJECT_NUMBER / GCP_PROJECT_ID → malformed WIF, cryptic failure**

No non-empty guard (`set -u` does not catch empty-but-set env vars). If either repo var is unset, the step still succeeds and emits `projects//locations/.../aicr-testgrid-github/...` and `...-publish@.iam.gserviceaccount.com`, which then fail opaquely inside google-github-actions/auth three steps later.

**Blast radius:** Operator sees a cryptic WIF rejection instead of 'GCP_PROJECT_NUMBER is unset' — the same legibility the Validate-producer step provides, missing for config vars.

**Fix:** Fail closed early: `: "${GCP_PROJECT_NUMBER:?repo var GCP_PROJECT_NUMBER is unset}"` (and GCP_PROJECT_ID) with an ::error::. (Source: persona panel + mchmarny + CodeRabbit.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+  publish:
+    name: Publish to TestGrid
+    runs-on: ubuntu-latest
+    timeout-minutes: 15  # Go build ~2 min + gcloud uploads ~3 min; 15 min is generous
+
+    # For workflow_run: run whether UAT succeeded or failed —
+    # a failure run is a valid TestGrid column showing what broke.
+    # For workflow_dispatch: restrict to main to prevent feature-branch
+    # code from obtaining WIF credentials (see comment on trigger above).
+    if: >
```
njhensley, line 154: **🟠 Major — Job `if:` silently drops cancelled / timed_out / skipped UAT conclusions**

The gate admits only workflow_dispatch-on-main or workflow_run conclusion success/failure. A cancelled, timed_out, action_required, or skipped UAT (e.g. an operator-aborted or 180-min-capped run) makes the whole job skip with no step output.

**Blast radius:** Observability gap that contradicts the header comment ('after every completed cluster run'): no TestGrid column AND no log explaining the no-op. On-call investigating 'why no column for run X' finds a job that simply didn't run.

**Fix:** Either broaden the gate and let Resolve-bundle-ref decide (it already handles no-pointer → skip), or add a ::notice:: documenting that non-terminal conclusions are intentionally not published. (Source: persona panel — net-new, not raised by prior reviewers.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+            echo "::error::BUNDLE_REF '${BUNDLE_REF}' does not match expected format ghcr.io/nvidia/...@sha256:<64hex>"
+            exit 1
+          fi
+          echo "bundle_ref=${BUNDLE_REF}" >> "${GITHUB_OUTPUT}"
+          echo "skip=false" >> "${GITHUB_OUTPUT}"
+          echo "Resolved bundle: ${BUNDLE_REF}"
+
+      # Authenticate to GCP via WIF using the testgrid publish SA.
+      # This SA has objectCreator on gs://aicr-testgrid[-<env>]/groups/ only.
+      - name: Authenticate to GCP
```
njhensley, line 360: **🟠 Major — Not 'inert until Terraform': successful UATs go red at WIF auth until prod Terraform applies**

TG2 is now merged so `go build` succeeds, but on a passing UAT the job still reaches Authenticate to GCP and fails there until the prod WIF pool/SA exist. The skip/no-pointer path is genuinely green; the success path is not.

**Blast radius:** Every passing UAT produces a red TestGrid Publish run (after exercising prod auth) between merge and the Terraform apply — alert noise that erodes CI signal.

**Fix:** Merge after the prod Terraform apply, OR gate the job behind a `vars.TESTGRID_PUBLISH_ENABLED == 'true'` feature flag (also makes 'inert until Terraform' literally true and gives on-call one switch). (Source: persona panel + mchmarny; re-pointed from the TG2 gate now that TG2 merged.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+  # Restricted to main — WIF attribute_condition gates on workflow filename
+  # but not branch; manual dispatch from a feature branch with modified
+  # workflow code could otherwise obtain GCS write credentials.
+  workflow_dispatch:
+    inputs:
+      bundle_ref:
+        description: "OCI bundle ref (ghcr.io/nvidia/...@sha256:...)"
+        type: string
+        required: true
+      source_class:
```
njhensley, line 107: **🟡 Minor — `source_class: community` is a dead option under the nvidia-only bundle regex**

The dispatch offers community, and TG2 genuinely supports `--source-class community`, but the BUNDLE_REF regex hardcodes `^ghcr.io/nvidia/...`, so a community bundle outside that namespace can never validate. The option is workflow-side unusable.

**Blast radius:** A manual community backfill always fails validation — a confusing dead end.

**Fix:** Either broaden the regex for the community class or drop the community choice until community bundles are namespaced under ghcr.io/nvidia. (Source: mchmarny, medium.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+          PRODUCER_BRANCH: ${{ github.event.workflow_run.head_branch }}
+        run: |
+          set -euo pipefail
+          if [ "${PRODUCER_REPO}" != "${GITHUB_REPOSITORY}" ]; then
+            echo "::error::Producer workflow is from '${PRODUCER_REPO}', expected '${GITHUB_REPOSITORY}' — rejecting"
+            exit 1
+          fi
+          # Release branches (e.g. release/v1.x) intentionally publish —
+          # their UAT runs qualify the release and belong in TestGrid.
+          if [ "${PRODUCER_BRANCH}" != "main" ]; then
```
njhensley: **🟡 Minor — Branch guard publishes from ANY non-main same-repo branch (notice only)**

Validate producer rejects forks but only emits ::notice:: for non-main producer branches and publishes anyway. UAT producers allow workflow_dispatch on any branch, so a dev-branch UAT lands a column in prod TestGrid.

**Blast radius:** Dashboard noise / false signal from experimental branches; not a security hole (WIF still gated, repo validated).

**Fix:** Allowlist explicitly: `case "${PRODUCER_BRANCH}" in main|release/*) ... ;; *) echo '::error::...'; exit 1 ;; esac` — matches the stated 'release branches are intentional' intent. (Source: persona panel + CodeRabbit + mchmarny.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+        type: choice
+        default: "prod"
+        options:
+          - prod
+          - staging
+
+concurrency:
+  # One publish per triggering run — prevents duplicate columns if the
+  # workflow is re-run manually while a prior publish is still in flight.
+  group: testgrid-publish-${{ github.event.workflow_run.id || github.run_id }}
```
njhensley: **🟡 Minor — Concurrency group doesn't dedup workflow_dispatch backfills; comment overstates it**

Group is `testgrid-publish-${{ github.event.workflow_run.id || github.run_id }}`. For dispatch, workflow_run.id is empty → falls back to the unique github.run_id, so two dispatches of the same bundle_ref get distinct groups and both run. The 'prevents duplicate columns if re-run manually' comment only holds for workflow_run re-runs.

**Blast radius:** Duplicate/racing columns possible on manual backfill unless testgrid-publish writes deterministic object paths.

**Fix:** Key the dispatch fallback on the bundle: `${{ github.event.workflow_run.id || inputs.bundle_ref || github.run_id }}`, or scope the comment. (Source: persona panel.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+      # Install yq (used to parse pointer.yaml) and Go using the repo's
+      # shared composite action — keeps tool versions in sync with .settings.yaml.
+      - name: Set up build tools
+        uses: ./.github/actions/setup-build-tools
+        with:
+          install_yq: 'true'
+          yq_version: ${{ steps.versions.outputs.yq }}
+
+      - name: Set up Go
+        uses: actions/setup-go@d60b41a563a30eb6578e24a53e1f85c8e82d1623  # v6.4.0
```
njhensley: **🟡 Minor — setup-go pin drift: v6.4.0 vs repo-standard v6.5.0**

This workflow pins actions/setup-go@d60b41... (v6.4.0); all ~18 other workflows pin @924ae3... (v6.5.0). Matches nothing else in the repo.

**Blast radius:** Dependency-bump noise and one-off SHA drift; no functional risk.

**Fix:** Bump to actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16  # v6.5.0. (Source: persona panel.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+            echo "::notice::Producer ran on branch '${PRODUCER_BRANCH}' — publishing (release branches are intentional)"
+          fi
+
+      # Resolve environment-specific GCP resource names.
+      # prod  → aicr-testgrid          (no env suffix — matches Terraform cluster_name)
+      # other → aicr-testgrid-<env>
+      - name: Resolve GCP config
+        id: gcp
+        env:
+          TG_ENV: ${{ inputs.environment || 'prod' }}
```
njhensley, line 212: **🟡 Minor — `environment` input selects bucket/WIF but is never passed to the tool**

environment (prod/staging) flows only into bucket/WIF resource names; the binary call passes --bucket and --source-class but no --environment. Architecturally fine iff the bucket fully encodes env (it does: bucket == cluster == aicr-testgrid[-<env>]).

**Blast radius:** None today; a coupling worth documenting so a future change doesn't assume the tool sees the env dimension.

**Fix:** Add a one-line comment that env is encoded in the bucket. (Source: persona panel.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+      - "UAT - AWS"
+    types:
+      - completed
+  # Allow manual dispatch for testing and backfills.
+  # Restricted to main — WIF attribute_condition gates on workflow filename
+  # but not branch; manual dispatch from a feature branch with modified
+  # workflow code could otherwise obtain GCS write credentials.
+  workflow_dispatch:
+    inputs:
+      bundle_ref:
```
njhensley, line 103: **🟡 Minor — workflow_dispatch bundle_ref provenance unverified beyond prefix+digest**

The regex constrains namespace+digest shape but not provenance. A main-branch operator (or anyone with push to ghcr.io/nvidia/*) could feed an arbitrary conforming bundle to the publish binary, which pulls and parses it under WIF creds.

**Blast radius:** Bounded by the SA's groups/-prefix-only objectCreator scope and the main-only dispatch gate. Real but well-contained.

**Fix:** Optionally verify the bundle's attestation/signature before pull, or document workflow_dispatch as trusted-operator-only. (Source: persona panel.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+          if [ "${EVENT_NAME}" = "workflow_dispatch" ]; then
+            BUNDLE_REF="${DISPATCH_BUNDLE_REF}"
+          else
+            # If the download step failed, distinguish "artifact not found"
+            # (conformance did not complete — benign skip) from a real error
+            # (network failure — should not silently drop the publish).
+            POINTER="$(find evidence-pointer/ -name 'pointer.yaml' 2>/dev/null | head -1)"
+            DOWNLOAD_OUTCOME="${DOWNLOAD_POINTER_OUTCOME}"
+            if [ -z "${POINTER}" ]; then
+              if [ "${DOWNLOAD_OUTCOME}" = "failure" ] && [ "${UAT_CONCLUSION}" = "success" ]; then
```
njhensley, line 302: **🟡 Minor — Narrow false hard-error on a manual skip_tests dispatch**

Producer uploads the pointer only if conformance succeeded. The `DOWNLOAD_OUTCOME=='failure' && UAT_CONCLUSION=='success'` branch is meant to catch a network error, but a deliberate skip_tests UAT concludes success with no pointer → this branch fires exit 1. (Downgraded from Major by meta-review: a normal conformance failure forces conclusion=failure, so this path is reachable only via the manual skip_tests dispatch, which has no evidence to publish anyway.)

**Blast radius:** A skip_tests UAT shows a confusing red publish run; the not-found-vs-network distinction remains imperfect (both surface as outcome=failure).

**Fix:** Accept the narrow edge, or have the producer always upload a sentinel pointer on success so absence unambiguously means failure. (Source: persona panel.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+          UAT_CONCLUSION: ${{ github.event.workflow_run.conclusion }}
+        run: |
+          set -euo pipefail
+          if [ "${EVENT_NAME}" = "workflow_dispatch" ]; then
+            BUNDLE_REF="${DISPATCH_BUNDLE_REF}"
+          else
+            # If the download step failed, distinguish "artifact not found"
+            # (conformance did not complete — benign skip) from a real error
+            # (network failure — should not silently drop the publish).
+            POINTER="$(find evidence-pointer/ -name 'pointer.yaml' 2>/dev/null | head -1)"
```
njhensley: **🔵 Nitpick — `find evidence-pointer/ ... | head -1` brittle under set -euo pipefail**

If the download dir is ever absent, `find <missing-dir>` exits 1 and pipefail+set -e kills the step before the -z disambiguation; head -1 closing the pipe can also SIGPIPE find on many matches. In practice download-artifact@v8 pre-creates the path: dir and merge-multiple yields one file, so it's benign today.

**Blast radius:** Latent; would surface only on an unexpected layout/failure mode, producing an opaque exit-1 instead of the intended notice/error.

**Fix:** Cheap insurance: `POINTER="$(find evidence-pointer/ -name pointer.yaml 2>/dev/null | head -1 || true)"` (or guard `[ -d evidence-pointer/ ]` first). (Source: persona panel + CodeRabbit.)

.github/workflows/testgrid-publish.yml
```diff
@@ -0,0 +1,295 @@
...
+            if [ -z "${POINTER}" ]; then
+              if [ "${DOWNLOAD_OUTCOME}" = "failure" ] && [ "${UAT_CONCLUSION}" = "success" ]; then
+                echo "::error::UAT succeeded but evidence pointer download failed — possible network error"
+                exit 1
+              fi
+              echo "::notice::No evidence pointer found (conformance did not complete) — skipping TestGrid publish"
+              echo "skip=true" >> "${GITHUB_OUTPUT}"
+              exit 0
+            fi
+            OCI=$(yq '.attestations[0].bundle.oci' "${POINTER}")
```
njhensley, line 324: **🔵 Nitpick — yq `.attestations[0]` hardcodes index 0**

Correct for schema V1 (BuildPointer always emits exactly one attestation), but couples to that invariant with no guard.

**Blast radius:** None today; a future V2 multi-attestation pointer could silently pick the wrong entry.

**Fix:** Optional fail-loud guard: assert `yq '.attestations | length' == 1`, or select by `.bundle.predicateType`. (Source: persona panel.)

mchmarny (approved):
## ✅ Approving — all review findings verified addressed at head `523be8b`

Re-derived every open finding from the three review passes (CodeRabbit ×5, mine, @njhensley's multi-persona) against the current file. Everything material landed in `3f348ede`:

**Fixed:**
- **Feature-flag gate** — `vars.TESTGRID_PUBLISH_ENABLED == 'true'` makes the workflow a true no-op (skipped, not failed) until prod Terraform is applied. Resolves the "not inert until Terraform" Major and removes the merge-sequencing constraint.
- **Conclusion coverage** — job now runs for any completed `workflow_run` conclusion, with a documented benign-skip path when no evidence pointer exists (cancelled/timed_out no longer silently dropped).
- **WIF naming** — pool/SA derivation parameterized (`${CLUSTER}-github` / `${CLUSTER}-publish`); trigger retargeted to `uat-run.yaml` with corrected trust-model docs.
- **Fail-fast var guards** — `: "${GCP_PROJECT_NUMBER:?...}"` / `GCP_PROJECT_ID` with actionable messages.
- **Output-injection hardening** — `grep -qE` replaced with anchored whole-string `[[ =~ ]]`, with a comment explaining the line-based-grep pitfall.
- **Branch guard** — `case` allowlist (main/release) failing closed before WIF auth.
- **`community` source-class** — bundle-ref regex broadened beyond `ghcr.io/nvidia/`, so the dispatch option is usable.
- **Pin drift** — `setup-go` bumped to repo-standard v6.5.0; concurrency group now dedups dispatch backfills by `environment+bundle_ref`.

**Accepted with in-code rationale** (fine as-is): `environment` selects bucket/WIF only ("the tool only needs `--bucket`"), the narrow `skip_tests` dispatch edge case, and `yq .attestations[0]` (correct for pointer schema V1).

The open review threads all predate `3f348ede` and are addressed by it — resolving them alongside this approval. CI is green at this head except the still-running `tests / E2E`; nothing in it exercises this workflow (new file, `workflow_run`-triggered), so no concerns there.

Nice work on the flag gate — it turns the rollout ordering problem into a one-variable flip.

PR #1483, opened 2026-06-26
feat(corroborate): consensus model and dashboard generator

njhensley:
## Summary

A deterministic generator that computes the recipe **corroboration consensus model** and emits the interim evidence dashboard: `index.json` + per-recipe `series/<recipe>.json` plus a self-contained static HTML/CSS/JS renderer that reads them.

## Motivation / Context

GP4 of the interim evidence dashboard — turn per-signer CTRF evidence into a sybil-resistant, byte-deterministic corroboration board. Modeled on `tools/bom`.

Fixes: #1404
Related: #1400

## Type of Change

- [x] New feature (non-breaking change that adds functionality)
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: new `pkg/corroborate` (consensus model + generator) and `tools/corroborate` (CLI); `.github/labeler.yml` glob

## Implementation Notes

- **Consensus counts distinct verified signers, never builds** — the five states (CONFIRMED/SINGLE/CONTESTED/FAILING/UNTESTED) over allowlisted signers, with `skipped`/`pending` → NOT-RUN and unallowlisted → zero-weight "reported" dots.
- **Source class** derived from the verified signer via an in-tree allowlist (regex-AST over-broad lint, disjoint/overlap validation); falls back to the class baked into `meta.json` when no allowlist is supplied.
- **Coordinates** via the shared `pkg/recipe.CoordinateFor` (imported, never parsing `metadata.name`); criteria inverted for display.
- **Byte-deterministic emit** — no clock/random/UUID; timestamps from the bundle predicate; fixed ordering throughout.
- **Renderer** — catalog → consensus grid → per-signer drilldown, k8s/aicr facets, HTML-escaped output + CSP, resolvable hash routes (`#/group/dashboard/tab[/signer]`); shows only groups with evidence.
- Deterministic demo-data synthesizer under `testdata/bigdemo`.

## Testing

```bash
make qualify
```

- `go build ./...` clean; `golangci-lint` (pinned) **0 issues**; `go test ./pkg/corroborate/... ./tools/corroborate/...` pass at **93% / 95%** coverage.
- Determinism verified: emit byte-identical across runs (incl. injected TZ skew).
- Note: `-race` was not runnable in the dev sandbox (no C compiler) — covered by CI.

## Risk Assessment

- [x] **Low** — new, isolated, additive package + tool; no changes to existing code paths; easy to revert.

[...]

Issue #1400, opened 2026-06-22
Epic: GP — Interim corroboration dashboard on GitHub Pages


njhensley:
### Summary

Establish AICR's interim public evidence surface: a **static, byte-deterministic GitHub Pages dashboard** that visualizes **corroboration across heterogeneous, signed sources** (first-party / community / partner) for each recipe. In-repo pointers are **per-source and add-only** (`recipes/evidence/<slug>/<source>/<bundle-digest>.yaml`, [#1347](https://github.com/NVIDIA/aicr/issues/1347) Option A; source = verified signer, CODEOWNERS-scoped) — the **community/partner** contribution channel; **first-party NVIDIA UAT ingests directly per run (no repo commit, avoiding churn)**. An AICR-owned CI job ORAS-mirrors the bundle, unpacks CTRF via `MaterializeBundle`, and verifies the signer (pinned issuer + in-tree allowlist) and OCI registry **before counting** — contributors never write the bucket. A deterministic `tools/` Go generator computes a **consensus model** (CONFIRMED / SINGLE / CONTESTED / FAILING / UNTESTED, counting **distinct verified signers, never builds**) and emits a deterministic **`index.json` + per-recipe `series/*.json`** + a self-contained HTML/CSS/JS **renderer that reads them** (catalog overview → per-recipe consensus grid → per-cell source dot-strip by class → per-source drilldown; CSP-first nav; k8s + aicr-version facets), published via `upload-pages-artifact` + `deploy-pages`. Reuses the TestGrid epic's shared recipe→coordinate mapping and publish/ingest concepts; **defers** that epic's live workers/API/UI/host cluster.

### Children
- [x] #1401 — GP1: Per-source evidence-pointer contract (#1347) + first-party direct-ingest + signer allowlist
- [x] #1402 — GP2: Ingest CI (pointers → verify-before-count → source-keyed GCS)
- [x] #1403 — GP3: GCS bucket + dedicated write-only publish SA (no host cluster)
- [x] #1404 — GP4: Consensus model + deterministic generator (the viz)
- [x] #1405 — GP5: GitHub Pages publish CI (merge-to-main cadence)
- [x] #1406 — GP6: #1263 re-scope + #1224/RQ1 cross-link + public docs

### Epic acceptance criteria

[...]

PR #1484, opened 2026-06-26
feat(evidence): GP2 ingest — verify bundles to source-keyed GCS

### LINKED ISSUE #1402: GP2 — Ingest CI (pointers → verify-before-count → source-keyed GCS)
**Parent:** #1400 · **Labels:** `enhancement`, `area/ci` (auto, `.github/**`) · **Size:** L · **Depends on:** GP1 (pointers + allowlist), GP3 (bucket + write SA)

**Goal.** An AICR-owned CI job — triggered both on merge-to-`main` over `recipes/evidence/**` (community/partner pointers) **and directly from each first-party UAT run** (by bundle ref, no repo commit) — that ORAS-mirrors the bundle to a durable registry, unpacks CTRF + metadata via `MaterializeBundle`, verifies the signature/issuer/identity and the OCI registry **before counting**, and writes a source-keyed GCS layout — so contributors never write the bucket and unverified evidence never reaches the model.

**Scope.**
- **Two triggers.** (a) On merge to `recipes/evidence/**`: discover **per-source pointer files** by glob `recipes/evidence/<slug>/<source>/*.yaml` (#1347 Option A) and read each via `LoadAndValidatePointer` (`pointer.go:37`) — reusing/extending the PR-time discovery harness in `.github/scripts/recipe-evidence-check.sh`. (b) Invoked **directly from each first-party UAT run** with the run's bundle OCI ref (`uat-gcp.yaml:291`), so first-party evidence ingests without a per-run repo commit. Both paths converge on the same verify→write steps below.
- **ORAS-mirror** each referenced bundle (`PointerBundle.OCI` + `.Digest`, `types.go:211`) to a durable registry as a backup against the source registry going dark — an **ORAS copy** (`pkg/oci/push.go` `oras.Copy`/`PushFromStore`, `:277,:294,:342`), *not* `aicr mirror` (which is image discovery, `pkg/cli/mirror.go:61-64`).
- **Pull + unpack** via `MaterializeBundle` (`pkg/evidence/verifier/fetch.go:80` → `materializeOCIRefRequireDigest`, `:195`), reusing the existing ORAS transport; read `ctrf/<phase>.json` (deployment/performance/conformance, written by `builder.go:136-155`) and predicate metadata (`AttestedAt`, criteria, aicr version, k8s, digest).
- **Verify before counting** in a step holding **no** bucket-write credentials: re-verify via the same verification engine the `evidence verify` CLI uses (`verifier.Verify`, `pkg/evidence/verifier/verify.go:52`), pinning **both** `--expected-issuer` (allowed issuer set) and `--expected-identity-regexp` (the GP1 in-tree allowlist — empty pins are a no-op; these are the pin flags defined at `evidence_verify.go:78-84`, `uat-aws.yaml:258-260`), reject tag-only refs (`--allow-unpinned-tag=false`, `evidence_verify.go:103`), and enforce a **trusted-registry allowlist** on the OCI ref (closing the open follow-up documented verbatim at `.github/workflows/recipe-evidence.yaml:37` / `.github/scripts/recipe-evidence-check.sh:46`), bounding layer/bundle size against SSRF/exfil/DoS. The verified signer (`PointerSigner` identity+issuer, `types.go:223`) and source **class** are recorded from the *verified* SAN, never the raw pointer string.
- **Write source-keyed GCS** in a *separate* step holding the dedicated write-only publish identity (GP3): `results/<recipe-coordinate>/<signer-id-hash>/<run-id>/{ctrf/<phase>.json, meta.json}`, where `<recipe-coordinate>` is from the shared mapping helper and `meta.json` carries verified signer id+issuer+class, aicr version, k8s `major.minor`, evidence digest, and `AttestedAt`. Create-only writes; same-bundle re-ingest is idempotent on the digest (`PointerBundle.Digest`).
- **Add a `tools/**` labeler glob** if the ingest reuses a shared Go helper under `tools/`; if it is pure workflow + `aicr` CLI invocations, `.github/**` → `area/ci` already covers it.

**Out of scope.** The bucket/SA Terraform (GP3); the generator/consensus model (GP4); Pages publish (GP5). Deciding corroboration weight (GP4 reads the GP1 allowlist).

**Key files.** New `.github/workflows/evidence-ingest.yaml` (+ any shared Go helper under `tools/` or `pkg/`); `pkg/evidence/verifier/fetch.go:80,:195` (`MaterializeBundle`); `pkg/evidence/verifier/verify.go:52` (`verifier.Verify`, the verification engine); `pkg/oci/push.go:277,:294,:342` (ORAS copy); `pkg/cli/evidence_verify.go:78-84,:103` (issuer/identity/unpinned-tag pin flags), `:123` (CLI's `Verify` call); pointer loader + single-attestation constraint `pkg/evidence/verifier/pointer.go:37,:74-83`; pointer schema `pkg/evidence/attestation/types.go:188-226`; verify-leg prior art (PR-time workflow + pointer discovery) `.github/workflows/recipe-evidence.yaml:30-48` / `.github/scripts/recipe-evidence-check.sh:40-52`, registry-allowlist follow-up `recipe-evidence.yaml:37` / `recipe-evidence-check.sh:46`; WIF auth `uat-gcp.yaml:53-54,:109,:115`; the shared `pkg/recipe` coordinate helper (imported).

**Tests.** A fixture/dry-run test that the ingest constructs the right verify + mirror + write invocations; an assertion the verify step uses **non-empty** issuer + identity pins and rejects tag-only refs; an assertion the registry-allowlist rejects out-of-list refs; a fork-safety assertion (no bucket credentials exposed to a PR/fork context, per `hippo:reviewer-cicd` norms) and that the write step uses the **dedicated `objectCreator`-only** SA (not the shared `github-actions@eidosx`); an idempotency test (same digest re-ingest replaces, does not duplicate); a source-keying test (two distinct signers land under two `<signer-id-hash>` prefixes). **75%** floor on any new Go helper.

**Acceptance.** On merge to `recipes/evidence/**`, the job ORAS-mirrors the bundle, unpacks CTRF via `MaterializeBundle`, verifies signer + issuer + identity + registry **in a credential-free step before counting**, and writes a source-keyed GCS entry with verified metadata using the dedicated write-only identity; unverified or out-of-registry bundles are rejected; contributors hold no bucket credentials.

---
_Child of #1400 · staged from `docs/design/013-interim-evidence-dashboard.md` (GP2)._

njhensley:
## Summary

Add the GP2 evidence-ingest producer: verify published, signed recipe-evidence bundles and synthesize the source-keyed tree the GP4 corroborate generator consumes, then publish it to the corroboration GCS bucket.

## Motivation / Context

Nothing on `main` turns the per-run signed evidence bundles (from `aicr validate --emit-attestation --push`) into the aggregated, source-keyed layout the corroboration dashboard reads. GP2 is that bridge — with **verify-before-count**: signature, issuer, identity, and source registry are all checked in a credential-free step *before* any results are recorded, so unverified evidence never reaches the model and contributors never hold bucket credentials.

Fixes: #1402
Related: #1400 (parent), #1404 / #1483 (GP4 consumer)

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Docs/examples (`docs/`)
- [x] Other: evidence ingest (`pkg/evidence/project`, `tools/evidence-project`), CI (`.github/workflows/evidence-ingest.yaml`, `uat-{aws,gcp}.yaml`, composite action)

## Implementation Notes

- **`pkg/evidence/project`** — pure, offline synthesis library. Derives the recipe coordinate + signer id-hash and writes `results/<group>/<dashboard>/<tab>/<idHash>/<runId>/{meta.json, ctrf/<phase>.json}` (schema `aicr-corroboration-meta/v1`). Deterministic, idempotent-replace; fails closed when no verified signer is present.
- **`tools/evidence-project`** — ingest CLI: enforce a trusted-registry allowlist, materialize once, verify on the unpacked dir with non-empty issuer/identity pins (`pkg/evidence/verifier`), classify, then synthesize. Holds no bucket credentials.
- **`evidence-ingest.yaml` (+ `evidence-ingest.sh`)** — two triggers (push to `recipes/evidence/**`; `workflow_call`/`dispatch` with `bundle_ref`). Credential-free verify job; separate fork-gated publish job that `gcloud storage rsync`s to `gs://aicr-testgrid-staging/results` via the eidosx WIF SA from `uat-gcp.yaml`.
- **UAT wiring** — `uat-aws.yaml`/`uat-gcp.yaml` call the ingest workflow as a dependent `ingest-evidence` job on a successful run, via the shared `export-evidence-ref` composite action; `runId` is set from the GHA run ID (the same one embedded in the OCI artifact), keeping the entry unique and traceable.
- **Allowlist** — loader/classifier shares the GP1/GP4 `recipes/evidence/allowlist.yaml` schema (`firstParty`/`community`/`partner`, exact-or-`^…$`-regex identity, over-broad + overlap lint) so producer and consumer classify a signer identically. A documented interim first-party heuristic applies only while that file is absent.
- **idHash** = first 12 hex of `sha256(issuer + "\n" + identity)`; the GP4 consumer treats it as opaque.

### Known / forward-looking limitations
- Statement-only community bundles whose signer is verified against Rekor by digest are **not yet ingestable** — today's verifier yields a verified signer only from an in-artifact `attestation.intoto.jsonl`, so such bundles fail closed.
- The publish job uses the shared `github-actions@eidosx` SA pending GP3's dedicated `objectCreator`-only identity.
- Discovers the flat `recipes/evidence/<recipe>.yaml` layout today; forward-compatible with GP1's nested per-source glob.

## Testing

```bash
# Run locally on the rebased base:
go build ./...                                                  # ok
go test ./pkg/evidence/project/... ./tools/evidence-project/... ./pkg/evidence/verifier/...
#   project 88.7%  cli 67.4%  verifier 67.0%
go vet ./pkg/evidence/project/... ./tools/evidence-project/...  # clean
golangci-lint run -c .golangci.yaml ./pkg/evidence/project/... ./tools/evidence-project/...  # 0 issues
actionlint + yamllint (3 workflows + composite action + labeler) # clean
check-docs-filenames / check-docs-mdx / check-agents-sync       # ok
```

`-race`, the project-wide `make test-coverage` floor, and the e2e/grype legs of `make qualify` were not run locally (no C compiler / Docker in the dev sandbox); the new Go code is goroutine-free and CI runs the full gate.

## Risk Assessment

- [x] **Medium** — Touches the AWS/GCP UAT workflows (adds an output + a dependent job) and introduces a CI job that writes to a shared GCS bucket.

**Rollout notes:** The new publish job is gated to the canonical repo (never forks) and writes only under a fresh `results/` prefix that does not collide with the bucket's existing paths (`config`, `grid/`, `groups/`, `queue/`, `summary/`, `tabs/`). The UAT `ingest-evidence` job only runs when a bundle was produced; it is additive and skips cleanly otherwise.

## Checklist

- [x] Tests pass locally (`make test` — `-race` deferred to CI; no C compiler in dev sandbox)
- [x] Linter passes (`make lint` — pinned golangci-lint 2.12.2, yamllint, actionlint)


.github/workflows/evidence-ingest.yaml
```diff
@@ -0,0 +1,210 @@
...
+    env:
+      GCP_WIF_PROVIDER: "projects/116689922666/locations/global/workloadIdentityPools/github-actions-pool/providers/github-actions-provider"
+      GCP_WIF_SERVICE_ACCOUNT: "github-actions@eidosx.iam.gserviceaccount.com"
+      BUCKET: "aicr-testgrid-staging"
+    steps:
+      - name: Download evidence tree
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c  # v8.0.1
+        with:
+          name: evidence-tree
+          path: out
```
mchmarny, line 198: nit (non-blocking): empty-ingest edge. In push mode `evidence-ingest.sh` exits 0 with no `out/results` when the push touched only the allowlist, deleted a pointer, or had no changed pointers. The verify job's upload then hits `if-no-files-found: ignore` → no `evidence-tree` artifact. But `publish` (`needs: verify`) still runs and `download-artifact` fails hard when the named artifact is absent — a spurious red run on an otherwise-fine push.

The `if [ ! -d out/results ]` guard in the rsync step never gets reached because download fails first. Options: gate `publish` on a verify-job output flag (e.g. emit `produced=true` only when `out/results` exists), or set the download step tolerant and keep the existing dir guard. The UAT `bundle_ref` path always produces output, so this only bites the `recipes/evidence/**` push trigger.

mchmarny (dismissed):
GP2 ingest — solid, security-conscious work. Walked the full verify-before-count path and it's sound and fail-closed at every gate:

- **Trust gate** (`tools/evidence-project/main.go`): trusted-registry allowlist fails closed on an empty list, matches on a path-segment boundary (`ghcr.io/nvidia` won't match `ghcr.io/nvidia-evil`), and is enforced *before* any pull. Digest-pinned refs are always gated regardless of `--allow-unpinned-tag`.
- **Verify before synthesize**: issuer + identity pins are required (non-empty), and synthesis refuses `ExitInvalid` / nil signer / nil predicate. Ingesting `ExitValidPhaseFailures` is correct — it records attested phase results.
- **Anti-sybil allowlist** (`classify.go`): `Classify` is nil-safe, `Validate` rejects over-broad regex constructs and overlapping entries, full-string identity match. The interim first-party heuristic is tightly anchored and only active when the allowlist file is absent.
- **Path safety** (`synthesize.go`): `filepath.IsLocal` on the recipe-derived name and runId, plus a `Rel`+`IsLocal` recheck on ctrf extraction — a verified-but-unallowlisted community bundle can't traverse out of its prefix.
- **Fork safety**: verify job holds no GCS creds; the credentialed publish job is separate and repo-gated; the GCS rsync `--delete` is scoped to a single run prefix.

One non-blocking robustness nit inline (empty-ingest → publish artifact download). Doesn't block merge.

Note: several CI legs (Test, E2E, analyze, Mirror E2E) were still pending at review time — worth a glance before merge, though the change is goroutine-free and the Go lint/CLI-E2E legs already passed.

.github/workflows/evidence-ingest.yaml
```diff
@@ -0,0 +1,230 @@
...
+          # so a re-ingest of the same run drops stale ctrf/*.json from a
+          # prior, larger run instead of leaving them behind. The delete is
+          # confined to the single run prefix, so other sources, other runs,
+          # and the bucket's other top-level prefixes are never touched.
+          find out/results -type f -name meta.json -print0 \
+            | while IFS= read -r -d '' meta; do
+              run_dir="$(dirname "${meta}")"
+              rel="${run_dir#out/results/}"
+              echo "syncing run ${rel}"
+              gcloud storage rsync -r \
```
mchmarny, line 227: nit: the `gcloud storage rsync` call inside the per-`meta.json` loop isn't `timeout`-wrapped, so a wedged object-store call parks the publish step until the job `timeout-minutes` backstop fires. Low risk — UAT produces one run dir per call — but `timeout 300 gcloud storage rsync ...` makes a hung call fail fast and loud rather than burning the whole job budget.

tools/evidence-project/main.go
```diff
@@ -0,0 +1,305 @@
...
+		PlainHTTP:        o.plainHTTP,
+	}
+	mat, err := verifier.MaterializeBundle(ctx, matOpts, form, pointer)
+	if err != nil {
+		return err
+	}
+	defer mat.Cleanup()
+
+	// Step 2 — verify on the materialized directory with non-empty pins.
+	vr, err := verifier.Verify(ctx, verifier.VerifyOptions{
```
mchmarny, line 172: This `Verify` runs on `mat.BundleDir` (a directory), so `verifier` re-detects the input form, `r.Pointer` is nil, and `CrossCheckPointerSigner` does not run on this pass. That's not a hole today — in push mode `evidence-ingest.sh` derives the identity pin (`^<claimed-identity>$`) from the pointer's claimed signer, so a lying pointer still fails the identity pin. But a future refactor could wrongly assume the pointer→signer cross-check fires here. Worth a one-line comment noting the cross-check is intentionally covered by the derived pin rather than this pass.

.github/scripts/evidence-ingest.sh
```diff
@@ -0,0 +1,108 @@
...
+  exit 0
+fi
+
+: "${BEFORE_SHA:?BEFORE_SHA is required in push mode}"
+: "${HEAD_SHA:?HEAD_SHA is required in push mode}"
+
+# Discover added/modified pointer files in this push. A deleted pointer
+# is absent at HEAD and is skipped (the tree entry it produced is left
+# in the bucket; pruning retired sources is a separate concern).
+changed=$(git diff --name-only --diff-filter=AM "${BEFORE_SHA}" "${HEAD_SHA}" -- 'recipes/evidence/*.yaml' || true)
```
mchmarny, line 66: nit: `|| true` swallows a genuine `git diff` failure (e.g. a bad SHA range) into an empty `changed`, indistinguishable from "no pointers changed" → exit 0. Soft false-pass. Low risk here since `BEFORE_SHA`/`HEAD_SHA` come from trusted `github.event.before`/`github.sha` with `fetch-depth: 0`, but capturing `rc` and failing on non-zero is the fail-loud form.

mchmarny (dismissed):
Verify-before-count is correctly fail-closed and pin-enforced: the ingest CLI requires non-empty issuer + identity pins before any pull, enforces the trusted-registry allowlist on a path-segment boundary before the network is touched, and the sigstore policy actually enforces the pins so an unsigned/non-matching bundle yields no verified signer and `synthesizeVerified` refuses it. Fork-safety is sound — no `pull_request_target`, no PR-HEAD checkout, both ingest and upstream UAT jobs gated on `github.repository`, credential-free verify split from the WIF publish job, and the GCS write scoped to a `results/` prefix with delete-confined rsync. Synthesis is pure/offline, deterministic, atomic (stage+rename), and confines bundle-local reads against symlink escape. Test coverage hits the security-critical paths.

Three minor inline notes (two low, one nit) — none blocking. LGTM.

mchmarny (approved):

Issue #607, opened 2026-04-17
Enhance deployment-phase validation to reflect post-install workload readiness


yuanchen8911:
> Note: this issue has been revised from its original version based on the discussions in the comments. The original scope (deploy.sh messaging/docs fix) is now handled separately in [#609](https://github.com/NVIDIA/aicr/pull/609).
>
> Follow-up note: the broader registry-driven work to make deployment-phase validation symmetric and deep for **all** components is now tracked in [#622](https://github.com/NVIDIA/aicr/issues/622). This issue remains intentionally scoped to strengthening the existing deployment phase for deployment completeness plus the targeted GPU readiness signals below.

## Summary

We already have most of the machinery for a deployer-neutral readiness check in `aicr validate`. The gap is not a missing standalone tool or a missing new phase; the gap is that the current `deployment` phase is still narrower than **post-install workload readiness**.

The proposal is to strengthen the existing `deployment` phase so that:

```bash
aicr validate --recipe recipe.yaml --phase deployment
```

answers the practical post-install question users care about: **is this cluster actually ready for GPU workloads now?**

Concretely, this issue is about:
- keep using the existing deployment phase
- preserve the existing baseline checks:
  - enabled namespaces are `Active`
  - declared `expectedResources` are healthy
- add the targeted GPU-specific readiness checks the phase was missing:
  - `skyhook-customizations`: Skyhook CR completion
  - `nvidia-dra-driver-gpu`: kubelet-plugin readiness
  - `gpu-operator`: `ClusterPolicy.status.state == ready`

So the main new value is the addition of the targeted Skyhook, DRA, and GPU Operator deep readiness checks on top of the preserved baseline deployment checks.

## Problem

Today, `deployment` validation is closer to:
- components were installed
- some expected resources exist
- basic operator/resource checks passed

That is still weaker than real post-install workload readiness for GPU bundles.

For example, the cluster may still be converging after install while:
- Skyhook tuning is still running
- the DRA kubelet plugin is not fully ready
- GPU components are present but not yet usable end-to-end

So the missing piece is not a new validator binary or a new CLI entrypoint. The missing piece is that the existing `deployment` phase does not yet include the specific readiness checks needed for post-install workload readiness.

## Scope Assumption

The scope of `--phase deployment` here is:
- deployment completeness for all enabled components, and
- the specific post-install GPU readiness signals needed to close the current gap

It is **not** intended to become:
- a generic per-component health-check aggregator
- a full end-to-end workload smoke test
[...]

PR #1495, opened 2026-06-26
fix(validator): drop single-shot ClusterPolicy check from expected-resources

njhensley:
## Summary

Removes the redundant single-shot Go ClusterPolicy readiness check from the deployment-phase `expected-resources` validator. The registry-declared chainsaw `gpu-operator` health check already asserts the same `ClusterPolicy state: ready` with a polling timeout, so the Go check added nothing but flake.

## Motivation / Context

The UAT - AWS run failed the `expected-resources` gate it gates on, with a single failure: `[INTERNAL] ClusterPolicy state=notReady (want ready)`. Investigation showed the GPU Operator readiness signal was being checked **two ways** inside the same validator:

- `verifyClusterPolicyReady` (Go) read `ClusterPolicy.status.state` **once, with no retry** — it sampled ~13s into the run.
- The chainsaw `gpu-operator` health check (`recipes/checks/gpu-operator/health-check.yaml`, step `validate-cluster-policy-ready`) asserts the same `state: ready` but **polls up to 5m**.

On a fresh node the `nvidia-operator-validator` runs four sequential init containers (driver → toolkit → cuda → plugin) before ClusterPolicy reconciles to `ready` — ~2.5m in the failing run. The Go check fired far too early and flaked the whole deployment gate, while the chainsaw check waited and passed. (Nodewright tuning was already `complete` and the DRA kubelet-plugin healthy, so neither was the cause.)

This completes the `clusterPolicyReady` migration that was deferred in #1220 / #1235 (the chainsaw replacement landed but the Go original was never removed).

Fixes: N/A
Related: #1220, #1235

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

## Implementation Notes

- Removed `verifyClusterPolicyReady`, its call site in `verifyGPUReadinessSignals`, and the now-unused `gpuOperatorComponent` / `clusterPolicyName` / `clusterPolicyReadyState` consts and `clusterPolicyGVR` var.
- The chainsaw assert is **co-extensive** with the removed Go check (both fire only when `gpu-operator` is enabled) and strictly superior (polls instead of single-shot), so there is no coverage loss.
- `verifyNodewrightReady` and `verifyDRAKubeletPluginReady` deliberately **stay in Go** — their resource names are dynamically/release-derived and cannot be expressed in static chainsaw YAML. Neither was implicated in this flake.
- Fail-closed-on-discovery-error coverage is preserved by re-targeting `TestCheckExpectedResources_FailsWhenDiscoveryReturnsNonNotFoundError` at the Nodewright discovery gate, which shares the same pattern.

## Testing

```bash
go build ./validators/...
go test ./validators/...                                  # all packages pass
golangci-lint run -c .golangci.yaml ./validators/deployment/...   # 0 issues (pinned v2.12.2)
```

- `go test ./validators/...` — all packages pass (deployment, chainsaw, conformance, helper, performance).
- Lint clean with the repo-pinned golangci-lint v2.12.2.
- **Note:** `-race` could not be exercised in my environment (no C compiler); CI's `make qualify` will run the race detector.

## Risk Assessment

**Rollout notes:** No behavior change for healthy clusters — the deeper `ClusterPolicy state: ready` signal is still asserted via the chainsaw `gpu-operator` health check, now with correct polling semantics. The only residual consideration: the chainsaw assert's 5m poll must cover ClusterPolicy reconcile on a fresh node (observed ~2.5m); a genuinely slower reconcile would now fail legitimately rather than flakily.

## Checklist


mchmarny (approved):
Clean removal of the redundant single-shot Go `ClusterPolicy` readiness check. Verified the chainsaw `gpu-operator` health check (`validate-cluster-policy-ready`) asserts the same `state: ready` with a 5m poll, registered via `assertFile` so it fires exactly when `gpu-operator` is enabled — co-extensive and strictly superior to the deleted Go check. Fail-closed and CR-missing coverage are preserved via the Nodewright gate. CI fully green.

Minor non-blocking note: the old Go check *skipped* (per #607) when the ClusterPolicy CRD wasn't registered; the chainsaw step instead asserts the CRD is Established and fails after 5m. That's correct for the deployment phase (operator already installed) and arguably stricter-better — not a regression.

LGTM.

PR #1500, opened 2026-06-26
fix(corroborate): escape boot() error + fail-closed meta schema gate

njhensley:
## Summary

Two post-merge review follow-ups from #1483: escape the one unescaped error string in the dashboard renderer, and make the `meta.json` schema-version gate fail closed on an incompatible major (symmetric with the allowlist loader).

## Motivation / Context

#1483 merged with two non-blocking review threads still open. This addresses both.

Fixes: N/A
Related: #1483, #1404

## Type of Change

## Component(s) Affected

- [x] Other: corroboration dashboard generator (`pkg/corroborate`)

## Implementation Notes

- **Renderer (`renderer/index.html`):** `boot()`'s `catch` interpolated `err.message` into `innerHTML` unescaped — the one data-adjacent string bypassing the otherwise-universal `esc()` discipline (some browsers embed a snippet of the offending `index.json` into `SyntaxError` messages). CSP (`script-src 'self' 'unsafe-inline'`) blocks execution so it was not exploitable, but it is now wrapped with `esc(err.message)` to match the series error path.
- **`loadRun` schema gate:** a `meta.json` schema-version mismatch previously only warned and continued, asymmetric with `LoadAllowlist`, which fails closed on an unsupported `schemaVersion` because a future schema may change trust semantics. An incompatible major (e.g. `meta/v2` repurposing `signer`/`allowlisted`) is now skipped via `errSkipRun` rather than parsed under v1 assumptions. An empty version stays tolerated by design (GP2 may omit it); a same-major future minor still loads with a warning. New `compatibleMetaSchema`/`schemaMajor` helpers carry the logic.

## Testing

```bash
go test ./pkg/corroborate/... ./tools/corroborate/...
golangci-lint run -c .golangci.yaml ./pkg/corroborate/... ./tools/corroborate/...
```

All pass; lint reports 0 issues. Added a table test for `compatibleMetaSchema` and an integration test asserting a `meta/v2` run is skipped.

## Risk Assessment

**Rollout notes:** N/A — behavior change only affects malformed/future-major evidence (skipped instead of mis-parsed) and an error-banner code path.

## Checklist

### DIFF
--- a/pkg/corroborate/generate.go
+++ b/pkg/corroborate/generate.go
@@ -232,14 +232,49 @@ func collectRuns(ctx context.Context, inputDir string, allowlist *Allowlist) ([]
 	return runs, nil
 }
 
+// compatibleMetaSchema reports whether a meta.json schemaVersion is safe to parse
+// under the current rules. An empty version is tolerated (GP2 may omit it); a
+// non-empty version is compatible only when it shares RunMetaSchemaVersion's
+// major (so a backward-compatible future minor still loads, but a new major does
+// not). See the gate in loadRun for the fail-closed rationale.
+func compatibleMetaSchema(got string) bool {
+	if got == "" {
+		return true
+	}
+	return schemaMajor(got) == schemaMajor(RunMetaSchemaVersion)
+}
+
+// schemaMajor returns the "<family>/v<major>" portion of a meta schemaVersion,
+// stripping any ".<minor>[...]" suffix. A value with no minor suffix is returned
+// unchanged.
+func schemaMajor(v string) string {
+	if i := strings.IndexByte(v, '.'); i >= 0 {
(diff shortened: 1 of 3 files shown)


mchmarny (approved):
Both follow-ups are correct and well-scoped:

- `esc(err.message)` closes the one data-adjacent string bypassing the otherwise-universal escaping discipline — good hardening regardless of the CSP backstop.
- The schema gate fails closed on an incompatible major and stays symmetric with `LoadAllowlist`. `schemaMajor`/`compatibleMetaSchema` correctly tolerate empty (GP2) and same-major minors while skipping a future major via `errSkipRun`. Table test + the `meta/v2` integration test cover the behavior.

CI green, lint clean. LGTM.

PR #1499, opened 2026-06-26
feat: migrate API domain to aicr.run, bump v1alpha2/v1beta2

mchmarny:
## Summary

Hard-break migration of the AICR artifact API domain `aicr.nvidia.com` → `aicr.run`, with the artifact version bumped `v1alpha1` → `v1alpha2` and the build-spec version `v1beta1` → `v1beta2`. Centralized through a single `pkg/header.Domain` constant.

## Motivation / Context

Tying the artifact contract to `nvidia.com` couples a vendor-neutral format to one company's brand/DNS and is verbose in every `apiVersion`, label key, and attestation URI. A group rename is a breaking change, so it is cheapest to land **before v1** while still on alpha/beta. The version bump is **signal-only** (no schema change) — it makes the breaking rename explicit so consumers can distinguish migrated artifacts at a glance. See [ADR-013](https://github.com/NVIDIA/aicr/blob/main/docs/design/013-aicr-run-domain-migration.md).

Fixes: #1486, #1487, #1488, #1490, #1491, #1493
Related: #1489 (closed as won't-do — superseded by the hard-break decision below)

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Component(s) Affected

- [x] CLI (`cmd/aicr`, `pkg/cli`)
- [x] API server (`cmd/aicrd`, `pkg/server`)
- [x] Recipe engine / data (`pkg/recipe`)
- [x] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [x] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [x] Validator (`pkg/validator`)
- [x] Docs/examples (`docs/`, `examples/`)
- [x] Other: `pkg/header`, `pkg/evidence`, OpenAPI, CI SLSA buildType

## Implementation Notes

Three coordinated changes land atomically so CI is never left half-migrated:

- **Domain** flips on all 5 semantic roles: artifact apiVersion group, K8s label/annotation keys, in-toto/SLSA attestation+provenance URIs, and the UUIDv5 namespace seed — all derived from `header.Domain`.
- **Versions** bump: artifacts → `aicr.run/v1alpha2`; build spec (`AICRRuntime`) → `aicr.run/v1beta2`.
[...]

PR #1507, opened 2026-06-27
fix(evidence,validator): correct catalog apiVersion examples, drop unverifiable demo evidence (#1499 follow-up)

yuanchen8911:
## Summary

Follow-up to #1499 (aicr.run domain migration). Fixes two defects that PR's domain sweep introduced: ValidatorCatalog doc examples pointed at an invalid apiVersion, and two committed demo evidence pointers became unverifiable.

## Motivation / Context

#1499 mechanically swept `aicr.nvidia.com` → `aicr.run`. A cross-review surfaced that the sweep touched two surfaces it shouldn't have:

**F3 — ValidatorCatalog examples (invalid apiVersion).** The validator catalog is a *separate* schema and is explicitly NOT part of the artifact-domain migration. Its loader fail-closes on anything other than `validator.nvidia.com/v1alpha1` (`pkg/validator/v1/catalog.go` `CatalogAPIVersion`; rejected in `pkg/validator/catalog/catalog.go`). #1499 swept four example blocks `aicr.nvidia.com/v1` → `aicr.run/v1` — still invalid — so an integrator copying any of them hits `unsupported apiVersion "aicr.run/v1"` at load. Fixed all four to `validator.nvidia.com/v1alpha1`:
- `docs/integrator/validator-extension.md:52`, `:150`
- `recipes/validators/README.md:8`
- `docs/design/002-validatorv2-adr.md:70`

Also aligned the three `ValidatorCatalog` test fixtures in `pkg/recipe/provider_test.go` to `validator.nvidia.com/v1alpha1` (they passed before only because the layered provider merges external catalogs under the embedded header; this keeps test data honest).

**F2 — unverifiable demo evidence pointers.** #1499 edited `predicateType` in place to `https://aicr.run/recipe-evidence/v1` in the two committed demo pointers, but `bundle.digest`/`oci` still pin the pre-migration, immutable OCI artifacts whose signed in-toto statements carry the old predicate type. `aicr evidence verify` materializes the bundle, then fails with `unexpected predicateType https://aicr.nvidia.com/recipe-evidence/v1` (`pkg/evidence/verifier/verify.go` / `signature.go` compare the *fetched* statement, not the YAML). A YAML-only fix is impossible in both directions: legacy fails `LoadAndValidatePointer` (`pointer.go`), and `aicr.run` fails the pulled-bundle check. The only true fix — regenerate + re-sign + republish the bundles — needs signing keys and OCI push, so this PR instead **removes** the two stale pointers (they also pinned a personal fork org, `ghcr.io/yuanchen8911/aicr-evidence`, not a canonical NVIDIA source) and the now-orphaned `community` allowlist anchor that existed solely to back them.

Two user docs (`docs/user/cli-reference.md`, `docs/user/artifact-verification.md`) showed the now-deleted concrete pointer path as the preferred `aicr evidence verify` example; switched both to the generic `recipes/evidence/<recipe>/<src>/<digest>.yaml` form.

Regenerating proper signed demo evidence under the NVIDIA org is a separate follow-up (it is the only path that restores verifiable evidence; editing strings cannot).

Fixes: N/A (follow-up to #1499)
Related: #1499

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Other: `pkg/evidence/allowlist`, `recipes/evidence/`

## Implementation Notes

- The two evidence pointers were the only ones anchored by the `community` source slug `7c4c0edc…`, so the `community` allowlist entry is removed too (→ `community: []`); the blocking `evidence-pointer-contract` gate enforces pointer→allowlist (not the reverse), so an empty community list with no community pointers is consistent.
- `pkg/evidence/allowlist/allowlist_test.go` `TestClassify` loads the real committed `allowlist.yaml`; its "community by slug" case is updated to assert that a community-issuer signer is now admitted as "reported" only (no community entries). This is a fixture update to match the intentional allowlist change, not a disabled test.
- Verified outside the repo while reviewing: `https://get.aicr.run` resolves and serves the install script (HTTP 200). The `https://aicr.run/...` predicate/buildType URIs are in-toto/SLSA *type identifiers* (compared as strings, never dereferenced), so the apex not resolving is expected and not a defect.

## Testing

```bash
go test ./pkg/evidence/allowlist/... ./pkg/evidence/verifier/... ./pkg/validator/catalog/... ./tools/evidence-project/...   # ok
golangci-lint run -c .golangci.yaml ./pkg/evidence/allowlist/...                                                          # 0 issues
```

Scope is docs + sample-evidence removal + one test fixture; no production Go logic changed. Targeted package tests and lint pass locally; full `make qualify` (e2e/scan/cluster) runs in CI.

## Risk Assessment

- [x] **Low** — Doc corrections plus removal of unverifiable sample evidence; no production code paths affected. Easy to revert.

**Rollout notes:** Demo evidence for `h100-gke-cos-training` / `gb200-eks-ubuntu-training` is removed until regenerated under the NVIDIA org (tracked follow-up). No runtime/API impact.

## Checklist

- [x] Tests pass locally (targeted packages with `-race`-equivalent unit runs)
- [x] Linter passes (`golangci-lint` on the changed package: 0 issues)


pkg/evidence/allowlist/allowlist_test.go
```diff
@@ -78,12 +78,13 @@ func TestClassify(t *testing.T) {
...
-			identity:  "yuanchen97@gmail.com",
-			wantClass: ClassCommunity,
-			wantOK:    true,
+			// Community is empty (the demo signer's evidence was removed in the
+			// #1499 follow-up), so a community-issuer signer is admitted as
+			// "reported" only, not classified.
+			name:     "community-issuer signer admitted as reported (no community entries)",
+			issuer:   "https://github.com/login/oauth",
+			identity: "yuanchen97@gmail.com",
+			wantOK:   false,
```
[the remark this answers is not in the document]
yuanchen8911: Good catch — fixed in 20989848. Replaced the retained personal email with a synthetic identity (`community-signer@example.com`); the case only asserts the unclassified (`wantOK: false`) outcome now that `community` is empty, so no real signer data is needed.

recipes/evidence/allowlist.yaml
```diff
@@ -51,10 +51,10 @@ firstParty:
...
 # Community: individual external contributors, keyed by source slug only.
 # Each commits pointers under recipes/evidence/<recipe>/<src>/ where <src> is
-# the slug below.
-community:
-  - issuer: https://github.com/login/oauth
-    source: 7c4c0edc8c765a95a0f3afdb3bbb8e91
+# the slug below. None onboarded yet — the prior entry anchored the
+# demo evidence removed in this PR (see #1499 follow-up); re-add alongside
+# regenerated, signed pointers.
+community: []
```
mchmarny, line 57: Deletion is the correct *immediate* fix — these can't verify post-migration in either direction (signed OCI statement is immutable and carries the pre-migration predicate type; reverting the YAML to the old domain fails `pointer.go`'s `PredicateTypeV1` check instead).

But the predicate type itself already migrated to `aicr.run` (`PredicateTypeV1 = "https://" + header.Domain + "/recipe-evidence/v1"`), so regenerating + re-signing produces statements that verify cleanly. Before merging with `community: []`: can we regenerate `h100-gke-cos-training` / `gb200-eks-ubuntu-training` under a canonical NVIDIA org (`ghcr.io/NVIDIA/...`) in this same PR, rather than landing with zero demo evidence and a follow-up? If that's blocked on signing-key / OCI-push access, delete-now + tracked follow-up is acceptable — just want to confirm we're not leaving the evidence surface empty longer than necessary. Either way, dropping the personal-fork pointer (`ghcr.io/yuanchen8911/...`) is right.

mchmarny (commented):
Verified both fixes against `main`. This appears to be a domain revert: this PR does **not** reset `aicr.run` back to `aicr.nvidia.com`. Every added line is either `validator.nvidia.com/v1alpha1` or a generic pointer placeholder.

**F3 (validator apiVersion) — correct, approve.** `validator.nvidia.com/v1alpha1` is a Kubernetes-style API group, not the AICR service domain, and it's the value the catalog loader fail-closes on (`pkg/validator/v1/catalog.go:21`, enforced at `pkg/validator/catalog/catalog.go:249`). #1499 only mis-swept the doc/test *examples* to `aicr.run/v1` — the `CatalogAPIVersion` constant itself was never changed — so copy-pasting the current docs hits `unsupported apiVersion "aicr.run/v1"` at load. This just realigns the examples to the unchanged code. Good catch.

**F2 (evidence removal) — diagnosis correct, one question on the remedy (non-blocking).** The signed OCI statements are immutable and predate the migration, so they can't be reconciled with a YAML edit in either direction. Deletion is the right immediate fix; see the inline note on whether we can regenerate under a canonical NVIDIA org in this PR rather than landing with `community: []` + a follow-up.

Not blocking merge. CI is green and the recipe-evidence gate is warning-only.

mchmarny (approved):

recipes/evidence/allowlist.yaml
yuanchen8911: Done in the follow-up #1509 (it landed as a separate PR only because #1507 merged before regeneration finished).

Both recipes were re-validated on live clusters, signed keyless, and pushed to ghcr.io/nvidia/aicr-evidence:
- h100-gke-cos-training — ghcr.io/nvidia/aicr-evidence:h100-gke-cos-training-3619d2b8a5ba (Rekor 1981934662), NCCL busbw 338 GB/s
- gb200-eks-ubuntu-training — ghcr.io/nvidia/aicr-evidence:gb200-eks-ubuntu-training-788a1bf8955a (Rekor 1982171931), NCCL nvls 841 GB/s

Both carry predicateType https://aicr.run/recipe-evidence/v1 and verify clean (`aicr evidence verify` -> exit 0: signature, predicate, manifest-hash). The signer is unchanged, so #1509 also restores the community allowlist entry (source 7c4c0edc...) removed here.

Note: the earlier NCCL performance failures were pure GPU-capacity contention (a standing slinky-slurm NodeSet held all GPUs/IMEX), not an IMEX/DRA problem — freeing it let the all-reduce workers schedule.

PR #1515, opened 2026-06-28
feat(ci): consolidate KWOK tier workflows into single reusable runner

### LINKED ISSUE #1172: Consolidate KWOK tier workflows: single reusable test workflow + one deployer-list source

mohityadav8:
## Summary

* Replace `kwok-tier3-shard.yaml` with `kwok-test-run.yaml`; the `checkout/load-versions/kwok-test` step block now lives in exactly one place.
* Define readonly `DEPLOYERS` once in `discover`; Tier 1, Tier 2, and Tier 3 all derive from it, removing hardcoded matrix axes and sync comments.
* Emit `tier1_pairs` and `tier2_pairs` from `discover`; all three tiers call `kwok-test-run.yaml` with pre-built `{recipe, deployer}` pairs.
* Keep Tier 2 deployer coverage Helm-only (deliberate; full coverage remains in Tier 3).
* Fix `workflow_dispatch` early-exit to also emit `tier1_pairs` and `tier2_pairs`.
* Add a Tier 1 pair-count guard (warning at >200).
* Update workflow path triggers and the ADR-003 workflow diagram.

Closes #1172


.github/workflows/kwok-recipes.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+﻿# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
mchmarny, line 1: This line now starts with a UTF-8 BOM (`EF BB BF`) — `main` doesn't have it, so it crept in via an editor. Strip it; yamllint can flag a BOM and it's just noise on a workflow file.

.github/workflows/kwok-recipes.yaml
```diff
@@ -237,16 +251,42 @@ jobs:
...
+            exit 1
+          elif (( tier1_pair_count > 200 )); then
+            echo "::warning::Tier 1 has ${tier1_pair_count} pairs (>200) — consider adding batching before it reaches 256"
+          fi
+
+          # --- Pre-build Tier 2 pairs: diff-affected recipes, helm-only ---
+          # Coverage-policy decision: Tier 2 uses helm only to keep PR wall-clock
+          # time proportional to the change scope. Full deployer coverage runs in
+          # Tier 3 on every push to main and on the nightly schedule. See ADR-003
+          # §"Tier 2 deployer coverage" for rationale and how to revisit this.
```
mchmarny, line 273: `ADR-003 §"Tier 2 deployer coverage"` doesn't exist — `docs/design/003-scaling-recipe-tests.md` has no such section. This same reference appears in the `test-tier2` job comment too.

More broadly, ADR-003 wasn't updated in this PR (only the two workflow files changed), so it's now stale:
- lines 8–9, 153, 178 still point at `kwok-tier3-shard.yaml` (renamed here to `kwok-test-run.yaml`)
- the diagram shows `test-tier2  matrix: tier2 × deployer`, but this PR makes Tier 2 helm-only

The PR description says it updates the ADR-003 workflow diagram, but that change isn't in the diff. Please update ADR-003 in this PR: fix the workflow name, the Tier 2 diagram line, and add the `Tier 2 deployer coverage` section these comments cite (or drop the `§` reference).

.github/workflows/kwok-recipes.yaml
```diff
@@ -237,16 +251,42 @@ jobs:
           # --- Tier 3: full matrix (all testable overlays) ---
           tier3="$all"
 
+          # Local alias so jq --argjson calls below can use $deployers
+          # (DEPLOYERS is readonly; this assignment is intentional).
+          # shellcheck disable=SC2034
+          deployers="${DEPLOYERS}"
```
mchmarny: nit: this alias isn't needed. `readonly` only blocks writes, so `--argjson deployers "$DEPLOYERS"` works directly in the jq calls below. And `DEPLOYERS`/`deployers` is read later (`deployer_count`, the tier1 pairs jq), so SC2034 wouldn't actually fire — the disable comment is misleading. Drop the alias and use `$DEPLOYERS` throughout.

mchmarny (commented):
Nice consolidation — single-source `DEPLOYERS`, the fail-closed Tier 1 pair-count guard, and the `workflow_dispatch` early-exit fix are all solid. The Tier 2 helm-only switch is behavior-preserving: the `kwok-test` action already defaults to `helm`, so the old deployer-less Tier 2 matrix was running helm anyway. Path-trigger parity is maintained too.

Three non-blocking issues (all inline): an accidental UTF-8 BOM on line 1, and a doc gap — ADR-003 still describes the old `kwok-tier3-shard.yaml` and `tier2 × deployer`, and the `§"Tier 2 deployer coverage"` section the comments cite doesn't exist, yet the PR body claims the ADR diagram was updated. Plus one nit on the `deployers` alias. The workflow logic itself looks correct.

Heads-up: the PR is currently `needs-rebase` (mergeable state blocked) — rebase onto `origin/main` before merge.

.github/workflows/kwok-recipes.yaml
njhensley, line 273: **🟠 Major — New workflow comments cite a nonexistent ADR-003 §"Tier 2 deployer coverage"**

This comment (and the duplicate at L337-338 on the test-tier2 job) directs readers to `ADR-003 §"Tier 2 deployer coverage"` for rationale. No such section exists — `grep -in "deployer coverage"` against `docs/design/003-scaling-recipe-tests.md` at this head returns nothing. It's a defect introduced by this PR: a shipped pointer into a phantom anchor.

**Blast radius:** A contributor revisiting the helm-only trade-off follows the pointer and finds nothing, defeating the comment's stated "how to revisit this" intent. Same dead pointer recurs at L337-338.

**Fix:** Add a `### Tier 2 deployer coverage` subsection to ADR-003 capturing the rationale (helm-only keeps PR wall-clock proportional to change scope; full coverage runs in Tier 3 on push/nightly) so both comments resolve — or repoint them at the existing rationale in `docs/contributor/tests.md` / ADR-008.

.github/workflows/kwok-recipes.yaml
njhensley, line 1: **🟡 Minor — UTF-8 BOM accidentally prepended to kwok-recipes.yaml**

Line 1 now begins with bytes `EF BB BF`. Verified authoritatively: the `origin/main` blob starts `23 20 43` (`# C`); this head blob starts `ef bb bf` — the BOM is introduced by this PR (an editor artifact; it's the diff's first hunk). No other workflow in the repo carries a BOM.

**Blast radius:** None functional — GitHub Actions and yamllint 1.38.0 both tolerate a leading BOM (lint passes), so CI won't fail. Cost is hygiene: git-blame noise, an encoding that differs from every sibling file, and a latent trap for tooling that keys on the first byte.

**Fix:** Strip it so the first hunk disappears: `sed -i '1s/^\xEF\xBB\xBF//' .github/workflows/kwok-recipes.yaml` (or re-save as UTF-8 without BOM).

.github/workflows/kwok-recipes.yaml
```diff
@@ -237,16 +251,42 @@ jobs:
           # --- Tier 3: full matrix (all testable overlays) ---
           tier3="$all"
 
+          # Local alias so jq --argjson calls below can use $deployers
```
njhensley: **🔵 Nitpick — Dead `# shellcheck disable=SC2034` on a used variable, plus an inaccurate comment**

The `deployers="${DEPLOYERS}"` alias is read at L262, L302, and L318, so SC2034 (unused variable) would never fire — the disable is a no-op. The comment also claims the alias is needed "so jq --argjson calls can use $deployers," but `${DEPLOYERS}` can be passed to `--argjson` directly (exactly as done at L115 in the dispatch branch), so the alias buys nothing.

**Blast radius:** None functional; maintainer-confusing — a future reader trusts the "intentional / unused" comment and may keep propagating an alias that could just be deleted.

**Fix:** Either drop the alias and use `${DEPLOYERS}` at the three call sites, or keep it and delete the misleading comment + SC2034 directive.

njhensley (commented):
📋 **Multi-persona review — PR #1515**

▎ Method: 4 persona passes (Correctness, Security, Operability/CI-DX, Domain-Architecture/Docs), each finding independently confirmed or refuted by a senior meta-reviewer against the resolved code. Line links/anchors pinned to head `dcd8f12c`.

The workflow logic is clean and well-factored: collapsing the six-deployer list to a single `readonly DEPLOYERS`, deriving all tiers from it, and generalizing the ex–Tier-3 shard into one shared `kwok-test-run.yaml` runner is a real maintainability win. Every jq/bash path checks out (pair + batch generators dry-run correctly, all call sites guard the empty-matrix case, the pair-count guards are fail-closed with correct boundaries), and the security surface (permissions, secrets, `pull_request` trigger, SHA-pinned checkout) is unchanged and not over-privileged. **No correctness or security blockers.**

What keeps it from a clean approve is **documentation consistency**, which this repo treats as a same-PR requirement: the PR body states it "updates the ADR-003 workflow diagram," but the diff touches only the two workflow files — ADR-003 is untouched, so it now carries three dangling references to the renamed file and a stale diagram, and the workflow's own new comments point to an ADR section that doesn't exist. None of this breaks CI, but it ships known-stale architecture docs.

---
## Findings anchored to files outside this diff (can't be inline)

🟠 **Major — ADR-003 left stale: dangling refs to the renamed workflow + stale diagram (PR body claims it was updated)**
`docs/design/003-scaling-recipe-tests.md` L9, L153, L178 (and diagram at L175)

This PR renames `kwok-tier3-shard.yaml` → `kwok-test-run.yaml`, but ADR-003 still names the old file at L9, L153, and L178 — `git grep kwok-tier3-shard` at head returns exactly these three hits, all now pointing at a file that no longer exists. The diagram at L175 still shows Tier 2 as `matrix: tier2 × deployer`, and the model no longer has any `deployer:` matrix axis at all. The PR Summary explicitly claims it updates "the ADR-003 workflow diagram," but `gh pr view --json files` shows only the two workflow files change.
**Fix:** In this PR, update ADR-003 — replace the three `kwok-tier3-shard.yaml` references with `kwok-test-run.yaml` (and reword "shard lane" → shared runner for all tiers), and redraw the diagram (Tier 1 = recipes × all deployers, Tier 2 = recipes × helm only, Tier 3 = full cross-product batched → `kwok-test-run.yaml`). If the ADR edits are truly out of scope, at minimum correct the PR body so it stops claiming an update it didn't make.

🟡 **Minor — Onboarding docs still tell contributors to edit a `deployer:` matrix this PR removes**
`docs/contributor/tests.md` L435 (parallel wording at `docs/design/008-kwok-deployer-matrix.md` L246)

This PR removes the literal `deployer: [helm, …]` matrix axes from Tier 1/Tier 3; deployers now come solely from `readonly DEPLOYERS` ("change this one line"). But `tests.md:435` still says *"Add the value to the `deployer:` matrix in Tier 1 and Tier 3"* and ADR-008:246 says *"Add `deployer: [...]` to Tier 1 and Tier 3 matrices"* — those matrices no longer exist. (The "leave Tier 2 alone" guidance remains correct.)
**Fix:** Update both to: "Add the value to the `DEPLOYERS` array in `discover` (single source of truth; Tier 1 and Tier 3 derive from it). Tier 2 stays helm-only by design." Ideally folded into this PR since it's the same behavioral change.

🔵 **Nitpick — `discover` still exports `tier1`/`tier2`/`tier3` outputs that nothing consumes**
`.github/workflows/kwok-recipes.yaml` L83–85 (unchanged context, so not inline)

Downstream jobs now reference only `tier1_pairs`, `tier2_pairs`, and `tier3_batches`. The `tier1`/`tier2`/`tier3` job outputs are consumed only by discover's own summary `echo`s. Harmless leftover surface — optionally drop the three unused `outputs:` entries (keep the internal shell vars, which the summary/batching still need).

> The remaining findings (🟠 phantom-ADR comment, 🟡 BOM, 🔵 dead SC2034) are attached as **inline comments** on the diff.

---

- **Branch-protection safe.** The required check is the aggregate `KWOK Test Summary` job (per ADR-003), which is unchanged. Converting Tier 1/Tier 2 to reusable-workflow calls renames only the per-recipe leaf check-runs, which aren't individually required.
- **Empty-matrix guarded everywhere.** All three `uses:` call sites gate on `!= '[]' && != ''`; tier3 batches are non-empty by construction; the `workflow_dispatch` early-exit sets every consumed output.
- **Pair-count guards correct.** Warn `>200` / error `>256` measures the exact quantity passed to the matrix; `256` (GitHub's cap) is allowed, `257+` fails closed. No off-by-one.
- **Tier 2 helm-only is behavior-preserving, not a coverage cut.** The `kwok-test` action has `default: 'helm'` and the old Tier 2 passed no `deployer` — so it already ran helm-only. The PR just makes it explicit; full deployer coverage still runs in Tier 3 on push/nightly.
- **Security unchanged.** Caller and reusable workflow are both `contents: read`; no secrets passed or needed; `on: pull_request` (not `pull_request_target`) keeps forks read-only/secret-less; checkout stays SHA-pinned with `persist-credentials: false`; the reusable ref is a local path. The one real script-injection sink (`inputs.recipe`/`inputs.deployer` → `run:` in `.github/actions/kwok-test/action.yml`) is pre-existing and untouched, low severity in a read-only secret-less runner — optional future hardening, not for this PR.
- **Rename hygiene within the workflows is complete.** Path triggers updated; no residual `kwok-tier3-shard` references inside the two workflow files (the only stragglers are in docs — see the Major above).

---
## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major   | 2 | ADR-003 stale (dangling refs + diagram, PR body overclaims); workflow comments cite phantom ADR section (inline) |
| 🟡 Minor   | 2 | UTF-8 BOM on line 1 (inline); onboarding docs reference removed `deployer:` matrix |
| 🔵 Nitpick | 2 | Dead SC2034 + misleading alias comment (inline); unused `discover` outputs |

**Recommendation:** Approve with comments. No code blockers — the workflow refactor is sound. Before merge, please fold the ADR-003 updates (and ideally `tests.md`/ADR-008) into this PR to satisfy the repo's same-PR docs rule, since the PR body already claims the ADR was updated; strip the BOM while you're in the file. The two nitpicks are optional.

<sub>Reviewed with a multi-persona + adversarial meta-review workflow.</sub>

mchmarny:
## 📋 Review status — verified against head `ed7311e`

### Addressed ✅
- **ADR-003 doc gap** — the ADR is now in the diff: `kwok-tier3-shard.yaml` reference updated to `kwok-test-run.yaml`, diagram redrawn for `tier1_pairs`/`tier2_pairs`, and the promised `### Tier 2 deployer coverage` section added, so the workflow comments' `§` pointer now resolves.
- Earlier CodeRabbit items (conflict markers in the workflow, Tier 1 hard-fail at >256, `workflow_dispatch` early-exit emitting pairs) — fixed in prior commits.

### Still broken 🔴
1. **Merge-conflict markers committed into ADR-003.** The Status section of `docs/design/003-scaling-recipe-tests.md` now contains literal `<<<<<<< Updated upstream` / `=======` / `>>>>>>> Stashed changes` blocks. Resolve the conflict — keep the updated status wording and drop the markers.
2. **UTF-8 BOM still on line 1 of `kwok-recipes.yaml`** ([thread](https://github.com/NVIDIA/aicr/pull/1515#discussion_r3502156676)). The first hunk still adds `EF BB BF`. `sed -i '1s/^\xEF\xBB\xBF//' .github/workflows/kwok-recipes.yaml`
3. **New: trailing whitespace + missing EOF newline** at the end of `kwok-recipes.yaml` (last hunk appends a whitespace-only line with no newline at EOF) — yamllint's trailing-spaces rule will flag this.

### Still open 🟡
4. Dead `# shellcheck disable=SC2034` + unneeded `deployers` alias ([thread](https://github.com/NVIDIA/aicr/pull/1515#discussion_r3502157711)) — still present verbatim.
5. `docs/contributor/tests.md` (~L435) and ADR-008 (~L246) still instruct contributors to edit the removed `deployer:` matrix; neither file is in the diff. Unused `tier1`/`tier2`/`tier3` discover outputs also remain.

The workflow refactor itself remains sound — items 1–3 are one cleanup commit away.

.github/workflows/kwok-test-run.yaml
```diff
@@ -33,7 +37,7 @@ permissions:
 
 jobs:
   test:
-    name: 'Tier 3: ${{ matrix.pair.recipe }} (${{ matrix.pair.deployer }})'
+    name: '${{ matrix.pair.recipe }} (${{ matrix.pair.deployer }})'
     runs-on: ubuntu-latest
     timeout-minutes: 15
```
mchmarny: **[high] Give Flux lanes enough timeout headroom.** On the current head, [`kind-inference (flux-oci)`](https://github.com/NVIDIA/aicr/actions/runs/29258579569/job/86847045696) reached successful test and cleanup steps but was cancelled at 15:03, so the aggregate summary failed. The same lane took 12:26 on current `main` and 12:53 on the prior PR head; its bounded Flux readiness check alone consumes about seven minutes. Raise the shared runner budget (20 minutes appears reasonable) or shorten that bounded wait, then rerun.

.github/workflows/kwok-recipes.yaml
```diff
@@ -100,13 +102,25 @@ jobs:
         run: |
           set -euo pipefail
 
+          # Deployer list — single source of truth, consumed by all tiers and by
+          # the workflow_dispatch early-exit below. To add or remove a deployer,
+          # change this one line; Tier 1, Tier 2, and Tier 3 all derive from it.
```
mchmarny: **[medium] This is not a single source for all tiers.** Tier 2 is intentionally constructed with the literal `deployer: "helm"` at line 276, so changing `DEPLOYERS` does not update Tier 2. Narrow this claim to Tier 1, Tier 3, and manual dispatch, or derive/assert `helm` from `DEPLOYERS`; otherwise the maintenance promise is false.

.github/workflows/kwok-recipes.yaml
```diff
@@ -84,6 +84,8 @@ jobs:
       tier2: ${{ steps.classify.outputs.tier2 }}
       tier3: ${{ steps.classify.outputs.tier3 }}
       tier3_batches: ${{ steps.classify.outputs.tier3_batches }}
+      tier1_pairs: ${{ steps.classify.outputs.tier1_pairs }}
```
mchmarny, line 84: **[low] Remove the obsolete recipe-list outputs.** `tier1`, `tier2`, and `tier3` are no longer consumed downstream; only `tier1_pairs`, `tier2_pairs`, and `tier3_batches` are. Remove outputs 83–85 and writes 306–308 so the consolidated workflow exposes one output contract.

docs/design/003-scaling-recipe-tests.md
```diff
@@ -150,7 +150,7 @@ To stay under the limit without sacrificing coverage, the `discover` job builds
 the `{recipe, deployer}` pairs and chunks them into batches of `TIER3_BATCH_SIZE`
 (200, with headroom under 256), emitting a `tier3_batches` output of
 `{id, pairs}` objects. `test-tier3` is a thin matrix over those batches that fans
-each one out to the **`kwok-tier3-shard.yaml`** reusable workflow, which expands
+each one out to the **`kwok-test-run.yaml`** reusable workflow, which expands
```
mchmarny, line 153: **[medium] Finish the workflow rename in this ADR.** The Status section at line 9 still points to deleted `.github/workflows/kwok-tier3-shard.yaml`. Update it to `.github/workflows/kwok-test-run.yaml`; otherwise the repository retains a dangling implementation reference.

docs/design/003-scaling-recipe-tests.md
```diff
@@ -163,19 +163,18 @@ and a summary. Tier 3 fans out to a reusable shard workflow (see above):
...
 
 test-tier2  (PR only, skip if empty)
-  matrix: tier2 × deployer
+  uses kwok-test-run.yaml  pairs=tier2_pairs  [helm-only]
 
 test-tier3  (push to main + schedule, skip on PR)
-  matrix: tier3_batches → uses kwok-tier3-shard.yaml (matrix: pairs)
+  matrix: tier3_batches → uses kwok-test-run.yaml (matrix: pairs)
 
 summary
```
mchmarny, line 179: **[high] Make the aggregate check actually required.** Active main ruleset `12304487` requires only `gate`, and `.github/workflows/merge-gate.yaml` does not depend on this workflow. The current head proves the gap: `gate` passed before [`KWOK Test Summary`](https://github.com/NVIDIA/aicr/actions/runs/29258579569/job/86852147120) failed. Issue #1172 explicitly requires this stable summary to remain required, so add it to the ruleset or provide equivalent required-gate integration before closing the issue.

mchmarny (changes requested):
Most earlier feedback is addressed: conflict markers, BOMs, the dead deployer alias, manual-dispatch outputs, the Tier 1 matrix guard, and the ADR structure are fixed. Local workflow/docs checks pass, but the current-head KWOK run still failed because a Flux lane exceeded the shared 15-minute budget, and the claimed required aggregate check is not enforced. Findings are inline; requesting changes until the blocking CI items are resolved and rerun.

Issue #739, opened 2026-05-05
[Epic]: Software supply chain security: visibility, reproducibility, verification


mchmarny:
## Goal

Raise AICR's software supply chain security to a level customers and internal security review can sign off on. Three maturity stages, mapped to the SLSA / NIST SSDF outcomes:

1. **Visibility** — every container image AICR can deploy is enumerable, auditable, and consumable by standard tooling (CycloneDX, Trivy, Grype, Cosign).
2. **Reproducibility** — the same recipe rendered today and a year from now produces the same artifacts. Recipe → image set is deterministic.
3. **Provenance & verification** — every artifact AICR deploys can be cryptographically tied to a known publisher and a known build, with admission-time enforcement where chart internals make in-tree pinning impractical.

## Why this matters

- **Customer security review.** Enterprise and regulated customers ask for an SBOM, a signing/provenance matrix, and a reproducibility statement before deploying. The SBOM and pinning story now exist; the signing/provenance matrix is in flight.
- **Air-gap deployment.** AICR pulls from 11+ public registries; customers behind a corporate firewall need a complete mirror list.
- **Compromise blast radius.** Tag-mutable references mean an upstream account compromise silently changes our deployments. Digest pinning + signature verification caps the blast radius and enables detection.
- **Reproducible builds.** "Same recipe → same bytes" is the foundation for diff-based audit, regression analysis, and incident response.
- **SLSA / SSDF alignment.** Visibility (level 1), Provenance (level 2), Hermetic builds (level 3) — each stage below corresponds to one of these.

## Current state

From the published BOM (`docs/user/container-images.md`, generated by `make bom`):

| Metric | Value |
|--------|-------|
| Components | 22 |
| Unique images | 69 |
| Distinct registries | 11 |
| Helm components without a chart-version pin | 0 ✅ |
| In-tree manifest images digest-pinned | 4 / 11 (the 7 remaining are CRD-triplet refs whose schemas reject digests; tracked via 4 upstream signing requests below) |
| Components with documented signing/SBOM status | 0 |

[...]

PR #1516, opened 2026-06-28
feat(ci): extend daily scan to full BOM image set at Medium+ with auto-generated VEX drafts (#1473)

### LINKED ISSUE #1473: Scan full BOM image set (Medium+) with auto-generated VEX
## Summary

Extend the daily vulnerability scan to cover **every image in the BOM** (the full ~82-image upstream set), not just the 6 images NVIDIA builds (`aicr`, `aicrd`, and the four validators). Set the severity cutoff at **Medium and higher**, and **auto-generate OpenVEX statements** for findings so that triaged/non-applicable results are captured as evidence rather than re-reviewed every run.

## Motivation

Today the [`Daily Image Vulnerability Scan`](https://github.com/NVIDIA/aicr/blob/main/.github/workflows/vuln-scan-images.yaml) scans only the images NVIDIA itself produces. The ~57 chart-default sub-images and the explicit in-tree refs that AICR deploys — the bulk of what actually lands on a customer cluster — are never scanned by us.

The auto-generated BOM (`docs/user/container-images.md`, `make bom`) already enumerates all images at pinned `@sha256:` digests and maps each to its component and pinned version. That inventory is the exact input a full-set scan needs. The capability is in reach now; what's missing is the operational wiring, not a technical barrier.

This closes the gap called out during the supply-chain epic (#739): we can *attribute* a CVE to an image/component/version via the BOM, but we do not continuously *detect* CVEs across the full set. For upstream images we can't fix directly, the scan output feeds escalation and VEX, not an in-tree code change — but that visibility is the precondition for any response.

## Scope

- **Images:** the full BOM image set (currently ~82 across 11 registries), sourced from `make bom` output rather than a hand-maintained matrix, so new components are picked up automatically.
- **Severity cutoff:** Medium and higher (`severity-cutoff: medium` in `anchore/scan-action`). Critical/High already drive notification; including Medium gives earlier signal on upstream drift without the Low/Negligible noise floor.
- **No build step:** images we don't produce are scanned by reference (Grype/`anchore/scan-action` pulls the remote digest) — no rebuild required.
- **Does not require mirroring.** This scans upstream digests in place. It is the cheaper, independent first step toward the broader mirror-attest-verify program (central SBOMs + custody attestations + admission verification), and a stepping stone to it — but it stands alone.

## Automatic VEX generation

- For each Medium+ finding, emit/maintain an OpenVEX statement keyed by the affected PURL + vuln ID, consistent with the existing `.openvex.json` consumed by the scan (`vex:` input) and the `aicr-managing-openvex` workflow.
- Default new findings to `under_investigation`; require a human to move a statement to `not_affected`/`affected`/`fixed` with a justification + evidence (do **not** auto-suppress — fail open on triage, not closed).
- Goal: a finding is reviewed once. Re-runs apply the existing VEX statement so the daily signal is *new/changed* findings, not the full standing backlog.
- Statements should carry reachability/justification evidence where available, matching the per-CVE pattern already used for the aiperf-bench suppressions.

## Open questions

- **Registry auth:** most of the 11 registries are public; any gated sources (private `nvcr.io` paths, ECR) need credentials wired into the scan job. Enumerate which require auth.
- **CI cost / cadence:** pulling ~82 images daily (some multi-GB, e.g. gpu-operator/CUDA) has real bandwidth/time cost. Confirm daily is right vs. a lighter cadence for the upstream set; consider caching and sharding the matrix.
- **VEX storage model:** single `.openvex.json` vs. per-image documents; how auto-generated drafts are reviewed/merged (bot PR vs. in-workflow artifact).
- **Notification routing:** Medium+ for ~82 images is a different volume than today's Critical/High over 6 — define the Slack threshold (e.g. notify on Critical/High always, Medium as a digest) so the channel stays actionable.

## Success criteria

- [ ] Daily scan covers the full BOM image set, sourced from `make bom` (no hand-maintained image list).
- [ ] Severity cutoff is Medium+; results report counts and IDs per image as today.
- [ ] New Medium+ findings produce OpenVEX draft statements (`under_investigation`) without auto-suppressing.
- [ ] Re-runs apply existing VEX so notifications surface only new/changed findings.
- [ ] Registry auth documented for any gated sources.
- [ ] Notification volume is tuned so the signal stays actionable.

## Related

- Parent epic: #739 (Software supply chain security)
- #745 — provenance audit + admission-time verification
- #743 — air-gap / private-registry mirroring guide
- Mirror-attest-verify program (central SBOM + custody attestations) — full-set scanning is a prerequisite/stepping stone; not yet filed.

mohityadav8:
## Summary

Extends `vuln-scan-images.yaml` to cover the full ~82-image upstream set (sourced from `make bom`) at Medium+ severity, and adds a `gen-vex-drafts.py` script that auto-generates `under_investigation` OpenVEX draft statements for new findings and opens a PR for human triage.

Closes #1473.
Parent epic: #739.

### `.github/workflows/vuln-scan-images.yaml`

Three new jobs added after the existing `scan` job; `notify` and `cleanup` updated.

| Job | Depends on | What it does |
|---|---|---|
| `bom-list` | *(independent)* | Runs `make bom`, extracts all container image refs from `bom.cdx.json` via `jq`, outputs a JSON array. New components are picked up automatically — no hand-maintained list. |
| `scan-bom` | `bom-list` | Dynamic matrix over all BOM images. `severity-cutoff: medium`, `only-fixed: false` (upstream images — escalation/VEX feeds rather than in-tree fixes). `max-parallel: 10` to cap bandwidth and registry rate-limit risk. Per-job upload: `.txt` summary + `.json` (full Grype output for VEX) + `.ref` (raw image ref). |
| `vex-draft` | `scan-bom` | Downloads all `bom-scan-*` artifacts, runs `gen-vex-drafts.py`, opens a PR against `.openvex.json` when new `(vuln-id, product-purl)` pairs are found. Skipped cleanly when `bom-list` fails or produces an empty list. |
| `notify` *(updated)* | `scan` + `scan-bom` | Two-tier Slack: NVIDIA-built images → Critical/High with full CVE IDs (unchanged); BOM upstream images → Critical/High with IDs always, Medium as a one-line count digest so the channel stays actionable. `scan-bom` being skipped does not block the NVIDIA notification. |
| `cleanup` *(updated)* | `notify` + `vex-draft` | Waits for the VEX draft PR to open before deleting scan images. |

### `.github/scripts/gen-vex-drafts.py` *(new)*

- Reads existing `(vuln-name, product-purl)` pairs from `.openvex.json` — idempotent, never duplicates.
- Reads each `*.ref` / `*.json` pair from `bom-scan-results/`.
- For each Medium+ match not already covered, emits an `under_investigation` statement with an OCI PURL keyed to the exact image ref.
- **Does not auto-suppress.** `under_investigation` leaves Grype reporting findings; a human must update status to `not_affected`/`affected`/`fixed` with evidence before suppression takes effect.
- Appends to `.openvex.json`, increments `version`, stamps `timestamp` and `tooling`.
- Writes `new_count=N` to `$GITHUB_OUTPUT`; the workflow gates the `create-pull-request` step on `new_count > 0`.

## Success criteria (from #1473)

- [x] Daily scan covers the full BOM image set sourced from `make bom` — no hand-maintained image list
- [x] Severity cutoff is Medium+; counts and IDs reported per image
- [x] New Medium+ findings produce `under_investigation` OpenVEX draft statements without auto-suppressing
- [x] Re-runs apply existing VEX so notifications surface only new/changed findings (idempotency tested locally)
- [x] Notification volume tuned: Critical/High always, Medium as a count digest
- [ ] Registry auth for gated sources — see open question below

## Testing

- `yamllint -c .yamllint.yaml` — 0 warnings
- `python3 -m py_compile gen-vex-drafts.py` — clean
- Idempotency: second run on same scan data outputs `new_count=0`, `.openvex.json` unchanged
- Job dependency graph verified via `python3 -c "import yaml; ..."` — all `needs:` chains correct


.github/workflows/vuln-scan-images.yaml
```diff
@@ -369,26 +368,276 @@ jobs:
...
   # =============================================================================
   # Notify: Post scan results to Slack (only on critical/high findings)
   # =============================================================================
 
   notify:
     name: Slack Notification
     runs-on: ubuntu-latest
-    needs: [scan]
-    if: always() && needs.scan.result != 'cancelled'
+    needs: [scan, scan-bom]
```
mchmarny, line 611: Recommend dropping the Slack wiring for the BOM scan entirely for now — keep the `bom-scan-*` results as artifacts only (the VEX draft PR already carries the per-finding detail for triage).

Rationale: the upstream set is ~82 images with no in-tree fix path, and gated images can't be scanned yet (auth is still open), so a Medium digest into the channel is high-volume and partially blind on day one. Until registry auth lands and the signal is proven actionable, the uploaded artifacts + the VEX draft PR are a sufficient record without adding channel noise.

Concretely: revert the `notify` changes here (keep `needs: [scan]` and the existing NVIDIA-only Slack path unchanged), and rely on the `bom-scan-*` artifacts. The BOM Slack tier can be added back in a follow-up once auth/coverage is in place and we know what's worth paging on.

.github/workflows/vuln-scan-images.yaml
```diff
@@ -369,26 +368,276 @@ jobs:
...
+
+          # Persist the raw image reference so the vex-draft job can reconstruct
+          # the product PURL without re-parsing Grype internals.
+          echo "${IMAGE}" > "bom-scan-results/${SHORT_NAME}.ref"
+
+          if [[ -z "${JSON_FILE}" || ! -f "${JSON_FILE}" ]]; then
+            echo "${SHORT_NAME}: scan output missing" > "${OUT}"
+            exit 0
+          fi
+
```
mchmarny, line 503: Gated `nvcr.io/nvidia/*` images can't be pulled without registry auth (the unchecked success-criteria box). When `scan-action` produces no JSON, this branch writes `scan output missing` + `exit 0`, so the image silently contributes 0 findings and is dropped from VEX (no `.json`).

Since the BOM results are now the artifact-of-record (per the notify comment above), make the artifact distinguish **"scanned: 0 findings"** from **"scan failed/missing"** — otherwise a future reader of the artifact can't tell a clean image from an unscanned one. A non-zero exit here (or an explicit `status: error` marker in the `.txt`/`.json`) would make the gap legible rather than silent.

.github/scripts/gen-vex-drafts.py
```diff
@@ -0,0 +1,262 @@
...
+    existing_keys = load_existing_keys(openvex_path)
+    scan_pairs = collect_scan_pairs(results_dir)
+
+    if not scan_pairs:
+        print(
+            "Error: no scan result pairs found in bom-scan-results/ after scan-bom ran.",
+            file=sys.stderr,
+        )
+        _write_output("new_count", "0")
+        return 1
```
mchmarny, line 229: `return 1` here fails the `vex-draft` job whenever `scan_pairs` is empty. That's reachable on a normal run where every BOM scan produced a `.ref` but no `.json` (transient registry/auth/pull failures) — the generator then shows a red job rather than the intended "skip cleanly." Consider `return 0` with `new_count=0` and a warning, or distinguish a true generator fault from "scan-bom ran but nothing was scannable," so a coverage gap doesn't masquerade as a generator error.

.github/scripts/gen-vex-drafts.py
```diff
@@ -0,0 +1,262 @@
...
+import json
+import os
+import sys
+from datetime import datetime, timezone
+
+
+_MEDIUM_PLUS = {"Critical", "High", "Medium"}
+
+
+def oci_purl(image_ref: str) -> str:
```
mchmarny, line 44: The BOM already emits a canonical `purl` per container component (`pkg:oci/<name>@<version>?repository_url=...`, see `tools/bom/README.md`), but the workflow passes only the raw `.ref` here, so this re-derives the PURL with its own scheme. If the two ever diverge (tag vs digest, qualifier order/encoding), the VEX `product` key won't match what Grype/BOM use and a draft could fail to apply. Threading the BOM's purl through the `.ref` artifact (or alongside it) would keep the product identity consistent end-to-end.

.github/scripts/gen-vex-drafts.py
```diff
@@ -0,0 +1,262 @@
...
+def _write_output(key: str, value: str) -> None:
+    """Append key=value to $GITHUB_OUTPUT if set."""
+    github_output = os.environ.get("GITHUB_OUTPUT", "")
+    if github_output:
+        with open(github_output, "a") as fh:
+            fh.write(f"{key}={value}\n")
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
mchmarny, line 261: nit: trailing whitespace after `sys.exit(main())` and no newline at end of file. `gofmt`-equivalents aside, a final newline keeps diffs/`cat` clean.

mchmarny (commented):
Well-structured. The BOM-derived image set (no hand-maintained list), idempotent keying by `(vuln, product-purl)`, and the non-suppressing `under_investigation` posture (human triage required before anything is hidden) are all the right calls.

Let's **drop the Slack wiring for the BOM scan in this PR and keep the `bom-scan-*` results as artifacts only for now.** With ~82 upstream images, no in-tree fix path, and gated images still unscannable (auth open), the BOM Slack tier is high-volume and partially blind on day one. The uploaded artifacts plus the VEX draft PR are a sufficient record.

Other items (none blocking): the gated-registry gap means unscannable images silently report 0 — since artifacts become the record, make them distinguish "0 findings" from "scan failed"; `gen-vex-drafts.py` returns `1` (red job) when no scannable results exist, which is reachable on a transient all-fail run; and the script re-derives the OCI PURL the BOM already emits (divergence risk for VEX product matching). Plus two nits (EOF newline, PR title ~96 chars vs the 70-char guideline).

It's reasonable to land the framework first for what it can reach. As you mentioned, full coverage will take some auth work, which will require some more thinking.

mchmarny:
@mohityadav8 First — thank you for the real work here. The BOM-sourced matrix, the idempotent `under_investigation` draft generation, the two-tier notification tuning, and the "never auto-suppress" discipline are all solid, and the implementation closely tracks the success criteria in #1473. None of that is wasted.

That said, the findings from this first pass have surfaced a more fundamental design question that I think we need to resolve before we invest further in the implementation. The short version:

**AICR is a configuration *definition* tool, not a deployment system.** When a full-set scan flags a CVE in one of the ~80 upstream images, our actual options are limited — we don't own those images and can't patch them. The meaningful response is usually to open an issue upstream, not to act in-tree. And emitting `under_investigation` VEX statements across 80+ images we didn't build risks generating standing noise rather than actionable signal — the opposite of what the notification tuning here is trying to achieve.

So this isn't a critique of the code; it's that the *strategy* the code implements needs more thought. Specifically: what response is AICR actually positioned to take on an upstream-image finding, and does VEX-across-the-full-set serve that, or just add maintenance burden?

**Proposal:** let's **close this PR but keep the branch** (`feat/vuln-scan-bom-1473`) so the work is preserved and easy to revive, and move the design discussion back to #1473 to align on the right path forward. Once we've settled the strategy, much of what you've built here should drop right back in.

Thanks again — genuinely good work, and I'd like to keep building on it once we've nailed the direction.

Issue #1401, opened 2026-06-22
GP1 — Per-source evidence-pointer contract (#1347) + first-party direct-ingest + signer allowlist


njhensley:
**Parent:** #1400 · **Labels:** `enhancement`, `area/recipes` (auto, `recipes/**`), `area/ci` (auto, `.github/**`) · **Size:** M · **Depends on:** none · **Implements:** [#1347](https://github.com/NVIDIA/aicr/issues/1347)

**Goal.** Establish the in-repo `recipes/evidence/` pointer contract per **[#1347](https://github.com/NVIDIA/aicr/issues/1347)** as the community/partner entry point, wire **first-party UAT to ingest directly (no per-run commit)**, and stand up the maintained signer allowlist + path-ownership enforcement.

**Scope.**
- Create `recipes/evidence/` (does not exist at HEAD) and adopt the **#1347 Option A per-source, add-only layout** `recipes/evidence/<slug>/<source>/<bundle-digest>.yaml` (see *Pointer file topology*): one **immutable** single-attestation V1 pointer per run, where `<source>` is a stable slug derived from the **verified** signer OIDC identity. Each file is the existing schema (`pkg/evidence/attestation/types.go:188`, `PointerSchemaVersion = "1.0.0"`, `:34`), parseable by `LoadAndValidatePointer` (`pointer.go:37`); discovery is by glob `<slug>/<source>/*.yaml`; nothing is overwritten. (Supersedes the earlier per-signer-file sketch.)
- **Path ownership (anti-squat).** Add `CODEOWNERS` rules (and/or a CI check) per `recipes/evidence/<slug>/<source>/` so a party can only write under its own `<source>/` — the path is otherwise squattable (#1347 cross-cutting).
- **First-party UAT does NOT commit a pointer per run.** That would churn `main` nightly. The UAT workflow already pushes a signed bundle to OCI (`uat-gcp.yaml:291`) with its OIDC identity pinned (`uat-aws.yaml:258-260`); it **invokes GP2 ingest directly with the bundle ref** instead. (A single low-cadence first-party pointer per launch recipe MAY be committed for discoverability/parity, drift-refreshed à la `bom-refresh.yaml` — but not per run.)
[...]

Issue #1535, opened 2026-06-29
Evidence pointer-contract gate trusts pointer-supplied signer (no crypto verification)


yuanchen8911:
## Summary

The blocking *Evidence Pointer Contract* gate establishes signer **identity from fields the pointer itself supplies**, not from a cryptographic verification of the bundle's signature. A committed pointer can therefore *claim* an allowlisted signer (issuer/identity) and pass the blocking gate, even if the referenced bundle was not actually signed by that identity. The cryptographic OCI signature verification is a separate, currently **warning-only** check.

## Evidence

`pkg/evidence/verifier/discover.go` `checkPointerFile` (~L156-186):
- reads `signer := ptr.Attestations[0].Signer` — the signer block **carried in the committed pointer**,
- derives the source slug from those supplied `issuer`/`identity` values (`SourceSlug`),
- checks path ownership + allowlist `Classify` against the same supplied values.

Nothing in this blocking path pulls the bundle and verifies a Fulcio/Rekor signature actually binds that identity to the bundle digest. So the gate trusts a *claimed* identity.

## Impact

A contributor (or a compromised/typo'd pointer) can commit a pointer that names an allowlisted community/partner signer it does not control and pass the blocking gate. Corroboration/consensus downstream may then count it. Severity depends on whether the warning-only OCI verification is later promoted to blocking.

## Fix options

1. Make the pointer-contract gate cryptographically verify the bundle signature (pull the referrer, verify Fulcio cert identity + Rekor inclusion) before trusting the signer block — i.e. promote the OCI verification from warning-only to blocking in the gate path.
2. Or document clearly that the on-disk gate establishes a *claimed* (not cryptographically proven) signer, and that trust derives from the separate signature-verification step.

## Related

- Surfaced during the docs correctness review of #1528; the demo (`demos/evidence.md`) now describes the pointer's signer as *claimed* at gate time and links here.
- Adjacent to #1530 (evidence signing flow) but distinct: this is about what the gate *verifies*, not where pointers live.

PR #1538, opened 2026-06-29
fix(evidence): make unsigned-commit CI-sign flow coherent (#1530)

### LINKED ISSUE #1530: Evidence signing: unsigned-commit flow contradicts pointer contract; sign script scans wrong path
## Summary

The unsigned-commit-then-CI-sign evidence-publishing flow is internally contradictory with the on-disk pointer contract, and the signing script it relies on is effectively dead (and mis-fires on `allowlist.yaml`). This needs a maintainer design decision, not a one-line path fix.

## Details

`.github/scripts/evidence-sign-unsigned.sh` (invoked by `.github/workflows/evidence-publish.yaml:130`) discovers "committed-but-unsigned" pointers and signs them in place. It globs the **flat** path:

```bash
for pointer in recipes/evidence/*.yaml; do
```

Three problems:

1. **Wrong location.** Every committed pointer in the tree is **nested**, e.g. `recipes/evidence/h100-gke-cos-training/<source>/sha256-….yaml`. The flat glob matches none of them — it matches only `recipes/evidence/allowlist.yaml`.

2. **Mis-fires on the allowlist.** `allowlist.yaml` is not a pointer and has no `.attestations[0].signer`, so the script's `yq` check returns `"null"`, classifies it as "unsigned," and would attempt `aicr evidence sign allowlist.yaml` (which fails).

3. **No relocation, and the flow is circular.** The script signs *in place* and never moves the pointer to its derived nested path. But the nested `<source>` segment is `SourceSlug(issuer, identity)` — derived from the **signer**, which an unsigned (`--no-sign`) pointer does not have (`pkg/evidence/attestation/pointer.go` `PointerCopyToHint` returns guidance, not a path, when there is no signer). So an unsigned pointer cannot be committed at its nested path to begin with. The only coherent unsigned-commit design is **commit-flat → CI-sign → CI-relocate-to-nested**, and the relocate step does not exist.

4. **Blocking gate rejects the intermediate.** `.github/workflows/evidence-pointer-contract.yaml` runs on any `recipes/evidence/**` PR change and requires every committed pointer to be a **signed, nested** single-attestation V1 pointer under `<recipe>/<source>/`. A flat unsigned pointer is rejected on the PR before the signing leg can help.

Net: the *commit-unsigned-then-CI-sign* path cannot pass the contract gate as designed; the script that supports it scans the wrong location and trips on the allowlist.

## Options (maintainer decision)

1. **Deprecate the unsigned-commit flow.** Require contributors to sign locally and commit the nested signed pointer directly (or use the ingest-by-ref GP2 path), then delete `evidence-sign-unsigned.sh` and its `evidence-publish.yaml` step.
2. **Make the flow coherent.** Have the script scan nested pointer paths, exclude `allowlist.yaml`, sign, and then relocate the now-signed pointer to its derived nested path — and reconcile that ordering with the blocking contract gate.

## Related

- Surfaced during the docs correctness sweep #1528 (follow-up note: https://github.com/NVIDIA/aicr/pull/1528#issuecomment-4835150636).
- The evidence-publishing doc in #1528 was aligned to the (broken) flat-commit step; once a direction is chosen here, that doc section should be updated to match (e.g. sign-locally-then-commit-nested).

njhensley:
## Summary

Make the commit-unsigned-then-CI-sign evidence-publishing flow internally coherent: a contributor commits an unsigned pointer **flat**, the fork-based CI leg signs it and **relocates** it to its canonical per-source path, and the contract gate accepts the flat pending pointer as a valid intermediate.

## Motivation / Context

The unsigned-commit flow was broken three ways: `evidence-sign-unsigned.sh` globbed a flat path that matched no committed pointer (and mis-fired on `allowlist.yaml`); an unsigned pointer could not be placed at its nested `<source>` path because that segment derives from the signer it lacks; and the blocking contract gate rejected the flat intermediate before signing could run. The only coherent design is **commit-flat → CI-sign → CI-relocate-to-nested**, and the relocate step did not exist.

Fixes: #1530
Related: #1401 (per-source pointer contract), #1402 (GP2 ingest)

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Core libraries (`pkg/errors`, `pkg/k8s`) — `pkg/evidence`
- [x] Other: `.github/` (evidence signing workflow + script)

## Implementation Notes

- **attestation**: `RelocatePointerToCanonical` moves a signed flat pointer to `recipes/evidence/<recipe>/<source>/<digest>.yaml`; `canonicalPointerSuffix` is the single source of truth for the per-source layout, shared with `PointerCopyToHint`. A `filepath.IsLocal` guard on the recipe segment blocks path traversal even when the function is called directly (no CI gate in front).
- **cli**: `aicr evidence sign --relocate` — sign-then-relocate, with an idempotent relocate-only recovery path for an already-signed flat pointer (and a no-op when already canonical).
- **verifier**: `CheckEvidenceTree` accepts a flat `<recipe>.yaml` as an unsigned, single-attestation, bundle-referencing pending pointer; still rejects signed-flat, misnamed, and unpushed files. Path-ownership and allowlist remain enforced on the signed, relocated pointer — anti-squat is not weakened.
- **ci**: `evidence-sign-unsigned.sh` scans flat pointers, skips `allowlist.yaml`, calls `sign --relocate`; `evidence-publish.yaml` stages the move with `git add -A` and commits the relocation.

## Testing

```bash
make test    # -race, affected packages green
make lint    # golangci-lint: 0 issues
```

New tests cover relocation (move, no-op, clobber-refusal, unsigned/no-digest rejection, recipe path-traversal rejection), the flat-pending gate cases, and the CLI `--relocate` flag (relocate-only idempotence + already-signed-without-relocate fails closed).

## Risk Assessment

- [x] **Low** — Isolated to the evidence-publishing path; thoroughly tested; the contract gate's security invariants (path-ownership, allowlist) are unchanged for signed pointers.

**Rollout notes:** No migration. The flat-pending acceptance is additive; existing nested signed pointers are unaffected. `--relocate` is opt-in.

## Checklist

### DIFF
--- a/.github/scripts/evidence-sign-unsigned.sh
+++ b/.github/scripts/evidence-sign-unsigned.sh
@@ -2,33 +2,42 @@
 # Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 #
-# Discover committed-but-unsigned evidence pointers and sign each in place.
-# Consumed by .github/workflows/evidence-publish.yaml (the fork-based
-# signing leg): a contributor pushes an unsigned bundle and commits the
-# pointer locally (`aicr ... --no-sign`), then this signs the bundle each
-# pointer references using the runner's ambient OIDC token and patches the
-# pointer's signer block in place.
+# Discover committed-but-unsigned evidence pointers, sign each, and relocate
+# it to its canonical per-source path. Consumed by
+# .github/workflows/evidence-publish.yaml (the fork-based signing leg).
 #
-# A pointer is "unsigned" iff its first attestation has no `signer` block —
-# the state `--no-sign` produces. Already-signed pointers are skipped, so
-# re-running is idempotent and safe.
+# The two-phase publish flow (#1530): a contributor pushes an unsigned bundle
+# and commits the pointer FLAT at recipes/evidence/<recipe>.yaml
+# (`aicr ... --no-sign`). A flat path is the only committable state for an
+# unsigned pointer — the nested <source> path segment derives from the signer,
+# which a `--no-sign` pointer does not have. This script then:
+#
+#   1. signs the bundle each flat pointer references with the runner's ambient
+#      OIDC token (patching the pointer's signer block in place), and
+#   2. relocates the now-signed pointer to its canonical per-source path
+#      recipes/evidence/<recipe>/<source>/<digest>.yaml — the layout the
+#      per-source contract gate requires.
+#
+# Both steps run inside `aicr evidence sign --relocate`, so the SourceSlug
+# derivation stays in one place (Go) rather than being reimplemented here.
+# `--relocate` is idempotent: an already-signed flat pointer (a prior run
+# signed it but the relocation didn't land) is moved without re-signing, so
+# re-running is safe.
+#
+# Pointers already relocated under <recipe>/<source>/ are not scanned (the
+# glob is flat), so they are never re-processed. allowlist.yaml is skipped.
 #
 # Required env:
 #   AICR   path to the aicr binary (built from a trusted source)
-#
-# Requires the mikefarah `yq` (v4) — the same `yq` the rest of the repo's
-# scripts assume. The kislyuk/python `yq` has incompatible syntax and is not
-# supported. On GitHub-hosted runners mikefarah yq is preinstalled.
 #
 # Any extra arguments are forwarded to `aicr evidence sign` (e.g.
 # --plain-http / --insecure-tls for local-registry tests).
 #
 # Outputs (and, when set, appends to $GITHUB_OUTPUT):
-#   skipped=<n>  pointers already signed (no-op)
-#   signed=<n>   pointers signed this run
-#   failed=<n>   pointers whose signing (or parsing) failed
+#   signed=<n>   flat pointers signed and/or relocated this run
+#   failed=<n>   flat pointers whose signing/relocation failed
 #
-# Exit status is non-zero when any signing failed, so the workflow surfaces
+# Exit status is non-zero when any pointer failed, so the workflow surfaces
 # the cause (commonly a private fork registry returning HTTP 403 on the
 # pre-sign pull — make the aicr-evidence package public).
 
@@ -40,49 +49,38 @@ extra_args=("$@")
 
 signed=0
 failed=0
-skipped=0
 
 shopt -s nullglob
+# The evidence root is also defined in Go as verifier.EvidenceDirName
+# ("recipes/evidence"); a shell script can't import the constant, so keep this
+# glob in sync if that path ever changes.
 for pointer in recipes/evidence/*.yaml; do
-  # signer absent => unsigned (the --no-sign state). A signed-without-Rekor
-  # pointer still has a signer block, so it is correctly skipped.
-  #
-  # Fail closed on a yq error: a missing yq or malformed YAML must NOT look
-  # like an unsigned pointer (which would resign an already-signed bundle and
-  # break rerun idempotence). Treat a parse failure as a hard failure.
-  if ! signer=$(yq eval '.attestations[0].signer // "null"' "$pointer" 2>/dev/null); then
-    echo "::error::failed to parse ${pointer} with yq — invalid YAML, or mikefarah yq (v4) not installed"
-    failed=$((failed + 1))
-    continue
-  fi
-  if [[ "$signer" != "null" ]]; then
-    echo "skip (already signed): ${pointer}"
-    skipped=$((skipped + 1))
+  # allowlist.yaml is not a pointer; never feed it to `evidence sign`.
+  if [[ "${pointer##*/}" == "allowlist.yaml" ]]; then
     continue
   fi
 
-  echo "signing unsigned pointer: ${pointer}"
+  echo "signing + relocating flat pending pointer: ${pointer}"
+  # --relocate: sign in place, then move to the canonical per-source path.
   # --yes: never pause for the interactive identity-disclosure prompt in CI
   # (ambient OIDC does not prompt, but this is belt-and-suspenders).
-  if "$AICR" evidence sign "$pointer" --yes "${extra_args[@]}"; then
+  if "$AICR" evidence sign "$pointer" --relocate --yes "${extra_args[@]}"; then
     signed=$((signed + 1))
   else
-    echo "::error::failed to sign ${pointer} — is the bundle's registry package public? (a private fork package returns HTTP 403 on the pre-sign pull)"
+    echo "::error::failed to sign/relocate ${pointer} — is the bundle's registry package public? (a private fork package returns HTTP 403 on the pre-sign pull)"
     failed=$((failed + 1))
   fi
 done
 
-echo "skipped=${skipped}"
 echo "signed=${signed}"
 echo "failed=${failed}"
 
 if [[ -n "${GITHUB_OUTPUT:-}" ]]; then
   {
-    echo "skipped=${skipped}"
     echo "signed=${signed}"
     echo "failed=${failed}"
   } >> "$GITHUB_OUTPUT"
 fi
 
-# Fail closed if any pointer could not be signed.
+# Fail closed if any pointer could not be signed/relocated.
 [[ "$failed" -eq 0 ]]
--- a/docs/contributor/evidence-publishing.md
+++ b/docs/contributor/evidence-publishing.md
@@ -42,12 +42,20 @@ aicr validate \
   --emit-attestation ./out \
   --push ghcr.io/<your-fork-owner>/aicr-evidence \
   --no-sign
-cp ./out/pointer.yaml recipes/evidence/<slug>.yaml
+cp ./out/pointer.yaml recipes/evidence/<recipe>.yaml
 ```
 
 `--no-sign` skips all OIDC/Fulcio/Rekor work, so this runs even where
 Sigstore egress is blocked. It pushes the content-addressed bundle and
-writes a pointer with an empty `signer` block. See
+writes a pointer with an empty `signer` block.
+
+Commit it **flat** at `recipes/evidence/<recipe>.yaml` (where `<recipe>` is
+the pointer's `recipe:` field). This is the *only* committable location for
+an unsigned pointer: the nested per-source path
+`recipes/evidence/<recipe>/<src>/<digest>.yaml` includes a `<src>`
+segment derived from the **signer**, which a `--no-sign` pointer does not yet
+have. The signing leg below relocates the pointer to that nested path once it
+is signed. See
 [`aicr validate`](../user/cli-reference.md#aicr-validate) and
 [`aicr evidence publish`](../user/cli-reference.md#aicr-evidence-publish)
 (which also accepts `--no-sign` if you push as a separate step).
@@ -63,11 +71,11 @@ writes a pointer with an empty `signer` block. See
 ### 2. Commit the unsigned pointer and push your branch
 
 ```shell
-git add recipes/evidence/<slug>.yaml
+git add recipes/evidence/<recipe>.yaml
 # Use -s (DCO sign-off) — required for all contributors. NVIDIA org members
 # additionally use -S (cryptographic signing); external contributors use -s
 # alone. See CONTRIBUTING.md.
-git commit -s -m "evidence: <slug> (unsigned; sign in CI)"
+git commit -s -m "evidence: <recipe> (unsigned; sign in CI)"
 git push
 ```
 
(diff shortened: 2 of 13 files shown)


pkg/evidence/verifier/discover.go
```diff
@@ -108,6 +122,44 @@ func CheckEvidenceTree(root, allowlistPath string) ([]TreeProblem, error) {
...
+		return "flat pending pointer must be named <recipe>.yaml (expected " + want + ", got " + base + ")"
+	}
+	att := ptr.Attestations[0]
+	if att.Signer != nil {
+		return "signed pointer must live under <recipe>/<source>/, not flat at the evidence root"
+	}
+	if att.Bundle.OCI == "" || att.Bundle.Digest == "" {
+		return "pending pointer must reference a pushed bundle (bundle.oci + bundle.digest) to be signable"
+	}
+	return ""
```
mchmarny, line 175: The gate now accepts a flat, **unsigned** `<recipe>.yaml` with no path-ownership or allowlist check, and `evidence-pointer-contract.yaml` runs on every PR with no branch scoping. Nothing enforces that the sign→relocate leg actually ran before merge, so a PR carrying only a flat pending pointer passes the contract gate and can merge to `main` — where it then sits permanently unsigned, since the auto-sign trigger is fork-only and never runs on the canonical repo.

The anti-squat invariant still holds (an unsigned pointer carries no verified signer, and ownership/allowlist are enforced once it's signed + relocated), so this isn't a trust bypass — but it does let an unverifiable, potentially recipe-squatting flat pointer land in the tree on `main`.

Is there a merge-time guard I'm missing? If not, consider either rejecting flat pending pointers when the contract runs against the `main` base, or documenting that reviewers must confirm the relocation landed before merge. Non-blocking — want to understand the intended guarantee.

pkg/cli/evidence_sign.go
```diff
@@ -107,10 +120,29 @@ func runEvidenceSignCmd(ctx context.Context, cmd *cli.Command) error {
...
+	// moved to its canonical path without re-signing. Without --relocate this
+	// falls through to ValidateSignablePointer, which fails closed on an
+	// already-signed pointer rather than re-signing it. LoadAndValidatePointer
+	// already guarantees exactly one attestation, so [0] is safe.
+	if relocate && pointer.Attestations[0].Signer != nil {
+		dest, rerr := attestation.RelocatePointerToCanonical(path, pointer)
+		if rerr != nil {
+			return rerr
+		}
+		slog.Info("evidence pointer already signed; relocated to canonical path",
```
mchmarny: nit: when the already-signed pointer is already at its canonical path, `RelocatePointerToCanonical` returns `currentPath` unchanged (no-op), but this still logs "already signed; relocated to canonical path" with `from == to`. Slightly misleading for the idempotent re-run case — consider logging "already canonical" when `dest == path`.

mchmarny (commented):
Solid, well-tested supply-chain change. The commit-flat → CI-sign → CI-relocate flow is coherent, the relocation logic lives in `pkg/evidence/attestation` (CLI just orchestrates), and the defensive details are right — `filepath.IsLocal` traversal guard on the recipe segment that doesn't depend on the CI gate, `ErrCodeConflict` on the immutable-pointer clobber refusal, and thorough tests (move/no-op/clobber/unsigned/no-digest/traversal + the flat-pending gate cases + CLI idempotence). Docs updated coherently.

One non-blocking design question on the contract gate: it now accepts an unsigned flat pending pointer with no ownership/allowlist check on every PR, and nothing enforces the sign+relocate leg ran before merge — so an unsigned flat pointer can land on `main` and stay unsigned (auto-sign is fork-only). Not a trust bypass since unsigned pointers are unverifiable and ownership is enforced once signed/relocated, but worth confirming the intended merge-time guarantee. Plus one nit on a log line. Nothing blocks merge.

pkg/evidence/attestation/pointer.go
```diff
@@ -88,20 +88,111 @@ func MarshalPointer(p *Pointer) ([]byte, error) {
...
+	if !filepath.IsLocal(p.Recipe) {
+		return "", errors.New(errors.ErrCodeInvalidRequest,
+			"pointer.recipe is not a local path segment: "+p.Recipe)
 	}
 	source, err := SourceSlug(att.Signer.Issuer, att.Signer.Identity)
 	if err != nil {
-		return unpushed
+		return "", err
 	}
 	file := strings.ReplaceAll(att.Bundle.Digest, ":", "-") + ".yaml"
```
mchmarny, line 140: This is the one open item worth fixing before merge (CodeRabbit flagged it too). `validatePointer` only checks the `sha256:` *prefix* (`discover.go`/`pointer.go:89`), not that the remainder is hex — so a digest like `sha256:../../../../etc/passwd` passes validation, and here it's fed straight into `filepath.Join` after the `:`→`-` swap with no traversal guard. The `filepath.IsLocal` check above covers `recipe` but not the digest, so the comment's claim that "the digest filename is validated sha256:<hex>" isn't actually true.

CI's sign path likely fails earlier at the bundle pull, but the relocate-only path (and a direct `aicr evidence sign --relocate`, which bypasses the gate as the comment notes) reaches `os.Rename` with the attacker-influenced digest. Cheap fix: reject `/`, `\`, `..` (or validate hex) on the digest stem before `filepath.Join`, mirroring the `IsLocal` guard you already apply to `recipe`.

mchmarny (commented):
Re-reviewed at `08a117a`. The two follow-ups from my first pass are handled: the relocate-only log nit is fixed, and the unsigned-pending merge-gate concern is resolved — `CheckEvidenceTree` now takes `allowPending`, and the merge gate runs without `--allow-pending` so a flat unsigned pointer can't land on `main`. Resolved that thread. CodeRabbit's substantive items (TOCTOU on the rename, doc/signing-policy wording) are also addressed.

One open item remains (inline): the bundle digest isn't validated as a safe filename before it's joined into the canonical path — a `..`-bearing digest passes the prefix-only check and reaches `os.Rename`. Limited real-world reach but a cheap, correct hardening, and it makes the existing code comment accurate. The `allowPending`-as-functional-option nit is fine to take or leave. Neither blocks merge once the digest guard is in.

mchmarny (dismissed):

pkg/evidence/attestation/pointer.go
```diff
@@ -88,20 +88,118 @@ func MarshalPointer(p *Pointer) ([]byte, error) {
...
+		return "", errors.Wrap(errors.ErrCodeInternal, "failed to create canonical pointer directory", err)
+	}
+	// Move with an atomic no-clobber guarantee: os.Link fails with EEXIST if
+	// dest already exists, so the exclusivity check and the placement are a
+	// single operation — no TOCTOU gap (a Stat+Rename would let another writer
+	// create dest after the Stat, and Rename silently clobbers it on Unix). The
+	// per-source pointer is immutable, so a pre-existing dest is a real conflict
+	// for the operator to resolve, not something to overwrite.
+	if err := os.Link(currentPath, dest); err != nil {
+		if os.IsExist(err) {
```
yuanchen8911, line 214: Idempotent recovery breaks on a partial prior move: if `os.Link` succeeds but `os.Remove` fails (or the process stops between), both the flat source and the canonical dest exist (same inode). A retry sees the still-present flat source (which doesn't satisfy the no-op suffix check), re-runs `os.Link`, hits `EEXIST`, and returns `ErrCodeConflict` — contradicting the documented idempotent recovery and leaving the flat source stuck. Treat an identical/same-inode dest as a completed placement (remove the stale source); keep the conflict only when dest differs.

.github/scripts/evidence-sign-unsigned.sh
```diff
@@ -40,49 +49,38 @@ extra_args=("$@")
...
     continue
   fi
 
-  echo "signing unsigned pointer: ${pointer}"
+  echo "signing + relocating flat pending pointer: ${pointer}"
+  # --relocate: sign in place, then move to the canonical per-source path.
   # --yes: never pause for the interactive identity-disclosure prompt in CI
   # (ambient OIDC does not prompt, but this is belt-and-suspenders).
-  if "$AICR" evidence sign "$pointer" --yes "${extra_args[@]}"; then
+  if "$AICR" evidence sign "$pointer" --relocate --yes "${extra_args[@]}"; then
```
yuanchen8911, line 67: Sign + relocate run in one call: the referrer is attached and the flat pointer rewritten before relocation. If relocation then fails, the command returns nonzero, the script increments only `failed` (so `signed=0` on the usual single-pointer run), and the workflow skips its commit step — discarding the signer patch. The next dispatch re-signs (duplicate referrer) instead of taking the relocate-only recovery path. Ensure the signer-patched pointer still reaches the commit step, or report partial signing separately.

pkg/cli/evidence_sign.go
```diff
@@ -107,10 +120,36 @@ func runEvidenceSignCmd(ctx context.Context, cmd *cli.Command) error {
...
 		return err
 	}
+
+	// Idempotent recovery: with --relocate, an already-signed flat pointer
+	// (signing succeeded on a prior run but the relocation didn't land) is
+	// moved to its canonical path without re-signing. Without --relocate this
+	// falls through to ValidateSignablePointer, which fails closed on an
+	// already-signed pointer rather than re-signing it. LoadAndValidatePointer
+	// already guarantees exactly one attestation, so [0] is safe.
+	if relocate && pointer.Attestations[0].Signer != nil {
```
yuanchen8911, line 142: The relocate-only recovery path takes any non-nil pointer-supplied signer without materializing or cryptographically verifying the bundle (validation only checks signer strings are non-empty). Combined with the contract gate validating only the claimed signer/source/allowlist metadata, this normalizes the trust gap tracked in #1535. Not a new privilege escalation (a contributor can already construct the nested path directly), but worth verifying the signature/bundle before relocating, or scoping `--relocate` to CI-signed pointers.

docs/contributor/evidence-publishing.md
```diff
@@ -88,19 +95,29 @@ runs two ways in your fork:
...
-above). Pull the commit it pushes (`git pull`) — your PR now carries a pointer
-whose `signer` block is filled in (the *bundle* is signed; the commit-back
-itself is a normal, unsigned GitHub Actions commit, which the eventual
-squash-merge re-signs under the repo's policy).
+above). Pull the commit it pushes (`git pull`) — your PR now carries a
+**signed, nested** pointer (the flat pending file is gone). The *bundle* is
+signed; the commit-back itself is a normal, unsigned GitHub Actions commit,
+which the eventual squash-merge re-signs under the repo's policy.
+
+> **Run this leg before merge.** The blocking per-source contract gate
```
yuanchen8911, line 113: This states a signed, nested pointer is sufficient, but a new fork signs with the GitHub Actions OIDC issuer + workflow identity, producing a `<src>` that isn't in the allowlist (whose community entry uses the personal GitHub OAuth issuer). `checkPointerFile` rejects an unlisted signer, so the PR stays blocked until it also adds its issuer/source to the community/partner allowlist. Add the post-sign allowlist-onboarding step (using the generated `<src>`), or link to the onboarding instructions.

yuanchen8911 (commented):
A few non-blocking correctness items on the relocate flow — the digest path-traversal from the earlier pass is fixed now, thanks. None are hard blockers, but good to fix before merge if feasible.

One higher-impact item that isn't on a changed line, so noting it here:

**Fork sign leg can't publish the signature referrer** (`.github/workflows/evidence-publish.yaml` permissions ~L80-82). The workflow grants `contents: write` + `id-token: write` but not `packages: write`, and there's no GHCR/registry login step. `SignExisting` pushes the Sigstore bundle as an OCI referrer — a registry write — while public-package visibility only enables the pre-sign pull. So a clean hosted fork runner can pull the bundle but can't attach its signature, leaving the pointer unsigned and the documented flow unable to complete end-to-end. Add `packages: write` and a registry login before the sign step.

pkg/evidence/attestation/pointer.go
yuanchen8911: Addressed in 4ba93e39 — `os.SameFile` now treats a same-inode dest as a completed placement (idempotent recovery), keeping the conflict only when dest differs.

docs/contributor/evidence-publishing.md
yuanchen8911: Addressed in 4ba93e39 — new "Add your signer to the allowlist" section keyed by the generated `<src>`.

.github/scripts/evidence-sign-unsigned.sh
yuanchen8911: Largely mitigated by the idempotent-relocate fix — a rerun now recovers a partial move. A narrow edge remains (a true relocation failure still counts `failed` and skips the commit), but it's much less likely now; fine as a follow-up.

pkg/cli/evidence_sign.go
yuanchen8911: Acknowledged — this is the pre-existing #1535 gate-trust gap (not a new privilege escalation, since a contributor can construct the nested path directly). Leaving it tracked by #1535.

yuanchen8911 (approved):
Re-reviewed at 88375d85 — the prior findings are addressed: packages:write + the ghcr-login step let the sign leg publish the referrer; the digest filename is path-guarded; relocation is idempotent (os.SameFile); and the allowlist-onboarding step is documented. Remaining items are non-blocking — a narrow signed-state edge on relocation failure, and the relocate-trust point tracked by #1535. LGTM.

yuanchen8911:
Non-blocking follow-ups (not merge blockers — fine to track as issues or a quick follow-up):

- **Relocation retry is inode-based.** The idempotent path uses `os.SameFile`, so it only recognizes a completed move within the same filesystem session; after a git round trip the flat and nested copies are distinct inodes and would still hit `EEXIST`/conflict. Low likelihood (a partial run leaves the flat pointer uncommitted, so you don't normally end up with both committed), but a content-equality fallback would close the committed-both case.

- **Fork package write access isn't documented.** The sign leg now has `packages: write` + the `ghcr-login` step, but a contributor must also grant their fork's Actions write access to the `aicr-evidence` package, and a locally-pushed package may be linked to the upstream repo rather than the fork — this prerequisite isn't covered in the guide.

- **`--relocate` still relocates on an unverified signer block.** The broader cryptographic-verification fix (the gate trusts the claimed signer) is tracked in #1535.

- **Minor:** the digest guard rejects path separators / `..` but doesn't require exactly 64 SHA-256 hex characters (e.g. `sha256:nothex` still validates).

PR #1543, opened 2026-06-29
fix(evidence): relocate idempotency, strict digest, package docs

njhensley:
## Summary

Non-blocking follow-ups to the merged evidence relocate flow (#1538 / #1530): close the relocation idempotency gap on a committed-both retry, tighten the bundle-digest guard to a canonical `sha256:<64-hex>`, and document the fork package-**write** prerequisites for the CI signing leg.

## Motivation / Context

These were raised as non-blocking review items on #1538 and tracked for a quick follow-up.

Fixes: N/A
Related: #1538, #1530, #1535

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Core libraries (`pkg/errors`, `pkg/k8s`) — `pkg/evidence/attestation`

## Implementation Notes

- **Relocation content-equality recovery** (`pkg/evidence/attestation/pointer.go`): the `os.Link` EEXIST recovery recognized only same-inode dests (`os.SameFile`), so after a git round trip — where the flat source and the committed nested copy are distinct inodes but byte-equal — a retry hit a spurious `ErrCodeConflict`. Added a bounded-read + `bytes.Equal` fallback (`pointerAlreadyPlaced`); only a dest holding **different** bytes is a real immutable-pointer conflict.
- **Strict digest** (`isHexDigest`): now requires exactly `sha256:` + 64 lowercase hex, rejecting `sha256:nothex`, truncated hashes, and wrong-algorithm prefixes instead of any hex-ish string. Test fixtures updated to the real 64-char h100 digest.
- **Fork package-write docs** (`docs/contributor/evidence-publishing.md`): documented the two GHCR write prerequisites the sign leg needs — the package's *Actions access* must include the fork with the **Write** role, and a locally-pushed package may be linked to `NVIDIA/aicr` rather than the fork and must be re-linked.
- **Not in scope:** `--relocate` trusts the on-disk signer block without cryptographic bundle verification — the same metadata-only trust the contract gate applies (not a new escalation), tracked repo-wide in **#1535**.

## Testing

```bash
make test    # -race, evidence + cli packages green
make lint    # golangci-lint 0 issues; gofmt clean; MDX OK
```

New/updated tests: relocation recovery on byte-identical dest (distinct inode), conflict on different-content dest, same-inode partial-move recovery, and digest-guard cases for non-hex / too-short / too-long / wrong-prefix.

## Risk Assessment

- [x] **Low** — Isolated to the evidence relocate path + docs; idempotency change only widens the already-placed recognition (a stricter conflict definition), and the digest guard only tightens validation.

**Rollout notes:** No migration. N/A.

## Checklist


mchmarny (approved):
Approving — clean, well-tested follow-up to #1538. (Re-reviewed at `7d99f4d`: the only change since the prior head was tightening the conflict test to assert `ErrCodeConflict` via `stderrors.Is` instead of bare non-nil — strictly better.)

- **Strict digest** (`isHexDigest` → exact `sha256:`+64 lowercase hex) is correctly wired into `canonicalPointerSuffix` before `filepath.Join`, alongside the `filepath.IsLocal(p.Recipe)` guard. Both attacker-influenced path segments now fail closed — this fully closes the digest-traversal concern from #1538, and the previously-spurious `sha256:..` case now genuinely exercises the guard.
- **Idempotent recovery** via `pointerAlreadyPlaced` (same-inode OR byte-identical) correctly handles the committed-both git-round-trip case; deterministic serialization means byte-equality ⟺ same signing output, so only differing bytes trip the conflict. Bounded `os.Open` + `io.LimitReader` read, fail-closed on read error — matches repo patterns.
- **Package-write docs** are the right operator-side complement to the workflow `packages: write` grant.

Tests cover the new paths well (byte-identical-distinct-inode recovery, different-content conflict with code assertion, same-inode recovery, full strict-digest matrix). The `--relocate` unverified-signer item is appropriately scoped out to #1535. Nothing blocks merge once e2e lands.

PR #1548, opened 2026-06-29
docs(demos): add dynamic-values and private-signing test walkthroughs

lockwobr:
## Summary

Two manual, use-case-driven testing walkthroughs under `demos/`, for exercising the bundle and supply-chain surface of a release candidate by hand: dynamic install-time values, and private/enterprise signing & verification.

## Motivation / Context

We have strong automated coverage (chainsaw + e2e), but no by-hand walkthroughs a teammate can follow to manually validate an RC's surface. These two docs fill that gap, framed identically to the existing `demos/bundle-attestation.md`.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- `demos/dynamic.md` — pre-built OCI bundle dynamic install-time values, driven by a single `AICRConfig` (`--config`) so the same file serves end users and testing. The `--output` flag overrides `spec.bundle.output.target` for the OCI step (ttl.sh, no local registry needed). Includes the `dynamic` vs `set` guidance and a corrected deployer table: plain `argocd` **rejects** `--dynamic` (use `argocd-helm`); `flux` uses `configmap-values.yaml` via `spec.valuesFrom`, not `cluster-values.yaml`.
- `demos/private-signing.md` — private/enterprise counterpart to `bundle-attestation.md`: self-hosted Sigstore (`--fulcio-url`/`--rekor-url` + `verify --trust-root`), KMS (`--signing-key` + `verify --key`), headless OIDC. Flag-based (config can't yet carry the KMS key, identity token, or verify options — see Follow-up). Trusted-root assembly and verify commands are taken from the `bundle-attestation-private-sigstore` chainsaw suite; `--ctfe` uses a dedicated `$CTFE_URL`.
- Manual walkthroughs by design (not scripts); each links to its automated chainsaw equivalent.

## Testing

Docs-only change. `demos/dynamic.md` was validated **end to end against the RC binary (v0.16.0-rc1)**: `recipe --config` + `bundle --config` produce the dynamic split (value removed from `values.yaml`, present in `cluster-values.yaml`, `install.sh` wires it), and the OCI round-trip via ttl.sh preserves it. `demos/private-signing.md` was **not** validated end to end — the local Kind runner got past the binary-attestation gate (after `aicr trust update` cached the public TUF root) but the Sigstore stack hit a transient `trillian-mysql` rollout timeout; its commands are taken verbatim from the verified chainsaw suite and CLI usage strings.

## Risk Assessment

- [x] **Low** — Docs only, easy to revert

**Rollout notes:** N/A

## Follow-up work

Surfaced while writing these docs; filed under epic #1149:

- #1566 — add KMS `signingKey` to `AICRConfig` (`spec.bundle.attestation`). Would let the KMS path of `private-signing.md` be `--config`-driven.
- #1567 — support `aicr verify` via `AICRConfig` (`spec.verify`). Scope expansion (config is producer-pipeline-only today); pending team decision.

## Checklist

- [x] I updated docs (this PR is docs)
- [x] Changes follow existing patterns in the codebase (mirrors `demos/bundle-attestation.md`)
- [ ] `make qualify` — not yet run (docs-only draft; will run before marking ready)


demos/dynamic.md
```diff
@@ -0,0 +1,207 @@
...
+```text
+bundle/
+├── checksums.txt
+├── recipe.yaml
+├── deploy.sh
+├── 010-gpu-operator/
+│   ├── values.yaml            # driver.version REMOVED
+│   ├── cluster-values.yaml    # driver.version lives here now
+│   ├── install.sh             # consumes cluster-values.yaml last
+│   └── README.md
```
njhensley: **🟡 Minor — Step-4 bundle-layout tree places README.md inside the component dir; it lives only at the bundle root**

A real build emits `values.yaml`, `cluster-values.yaml`, `install.sh`, and `upstream.env` per component — the helm deployer delegates per-component content to `localformat`, which writes NO per-component `README.md` (`pkg/bundler/deployer/helm/helm.go:46` — "no longer owns per-component folder content"; README is a bundle-root file via `generateRootREADME`). The tree draws `└── README.md` under `010-gpu-operator/` and omits `upstream.env`.

**Blast radius:** A reader running `cat ./bundle/010-gpu-operator/README.md` (as the tree implies) gets "No such file or directory." Cosmetic, but factually wrong.

**Fix:** Move `README.md` to the bundle-root section of the tree and add `upstream.env` under the component.

demos/private-signing.md
```diff
@@ -0,0 +1,275 @@
...
+`--fulcio-url` / `--rekor-url` redirect keyless signing to your stack. Both must be absolute `https://` URLs with no embedded credentials. `--identity-token` is an OIDC token your stack's Fulcio trusts (e.g. minted by your IdP, or a Kubernetes ServiceAccount token):
+
+```shell
+# a bundle is built from a recipe — generate one first (offline, no signing)
+aicr recipe --service eks --accelerator gb200 --os ubuntu --intent training -o recipe.yaml
+
+aicr bundle --recipe recipe.yaml --output ./bundle --attest \
+  --fulcio-url "$FULCIO_URL" \
+  --rekor-url "$REKOR_URL" \
+  --identity-token "$OIDC_TOKEN"
```
njhensley: **🟡 Minor — Step 2 passes the OIDC token via --identity-token (argv-exposed), contradicting step 7's own warning**

Step 2 signs with `--identity-token "$OIDC_TOKEN"`, placing the bearer token in the child `aicr` process argv (visible in `ps` / `/proc/<pid>/cmdline`). Step 7 (line 251) explicitly warns about exactly this and prefers `COSIGN_IDENTITY_TOKEN` — but the primary sign example readers copy first uses the exposed form with no caveat.

**Blast radius:** On the shared bastion/CI hosts this mode targets, a co-tenant can capture the token for the process lifetime. The doc contradicts its own later guidance.

**Fix:** Use `COSIGN_IDENTITY_TOKEN="$OIDC_TOKEN" aicr bundle …` in step 2 (or forward-reference the step-7 note), keeping `--identity-token` as the documented-but-discouraged alternative.

demos/private-signing.md
```diff
@@ -0,0 +1,275 @@
...
+curl -fsS "$REKOR_URL/api/v1/log/publicKey" -o rekor.pub
+# CTLog (CTFE) public key — MANDATORY (aicr does not disable Signed Certificate
+# Timestamp checks, so the private Fulcio cert fails verification without it).
+# Local store: extract it from the cluster. Your own stack: get it from your CTLog.
+kubectl --context kind-aicr-sigstore-e2e -n ctlog-system get secret ctlog-public-key \
+  -o jsonpath='{.data.public}' | base64 -d > ctfe.pub
+
+cosign trusted-root create \
+  --no-default-fulcio --no-default-rekor --no-default-ctfe --no-default-tsa \
+  --fulcio="url=$FULCIO_URL,certificate-chain=fulcio-chain.pem,start-time=1970-01-01T00:00:00Z" \
```
njhensley, line 154: **🔵 Nitpick — Epoch start-time is caveated but doesn't name the security consequence**

`start-time=1970-01-01T00:00:00Z` in `cosign trusted-root create` disables the time-based lower bound of the trust anchor. The doc caveats it ("set it to your CA/log's real start time") but frames the reason only as "widens it so fresh test certs fall inside." Since `--trust-root` is additive to the public-good root, a prod trusted root copy-pasted with the epoch value would accept trust material regardless of issuance date.

**Blast radius:** Low — this is an explicitly throwaway demo root, and production use is already caveated. Purely a clarity improvement.

**Fix:** Add a clause naming the consequence, e.g. "the epoch value accepts trust material regardless of issuance date; in production set the CA/log's real inception timestamp."

njhensley (commented):
📋 **Multi-Persona Review — PR #1548**

Method: 4 persona passes (Technical-Accuracy, Security, Docs, CI-DX), each finding independently confirmed or refuted by a senior meta-reviewer against the resolved code. Line links pinned to head `908b41ea`.
Tier legend: 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick · ✅ Confirmed non-issue

This is a strong, genuinely useful pair of hand-run walkthroughs, and the command-level accuracy is excellent — the `AICRConfig` schema, every documented flag (`--config/--output/--dynamic/--fulcio-url/--rekor-url/--signing-key/--identity-token/--oidc-device-flow`, `verify --trust-root/--key`), the `argocd`-rejects-`dynamic` behavior, the `--signing-key` mutual-exclusion set, the `install.sh -f` ordering, the flux/argocd-helm deployer semantics, and all anchor targets were verified against the source (and a live build). No blockers, nothing that breaks CI or ships a bad artifact.

Two things worth fixing before merge: the two new demos are **orphaned from `demos/README.md`** (major, below), and the `dynamic.md` step-4 layout tree points a reader at a `README.md` path that doesn't exist inside the component dir (inline). Everything else is minor/optional.

## 🟠 Major — New demos are not indexed in `demos/README.md` and have no inbound links

https://github.com/NVIDIA/aicr/blob/908b41eaa192b1d437cbfa2f71ebfd4b2466bf54/demos/README.md#L26

The "Available Demos" table indexes every runbook including `bundle-attestation.md`, but neither `dynamic.md` nor `private-signing.md` is added. `git grep 'dynamic.md|private-signing.md'` across the repo at head returns **zero** inbound references, so both docs are reachable only by knowing the filename — the new walkthroughs link *out* to their siblings, but nothing links *back* in.

**Blast radius:** anyone browsing the demos index (the maintained entry point) never discovers these two walkthroughs.

**Fix:** add two rows to the table, next to the `bundle-attestation` rows:
```markdown
| [dynamic.md](dynamic.md) | Dynamic install-time values (cluster-values.yaml split, install-time override, OCI artifact) |
| [private-signing.md](private-signing.md) | Private/enterprise signing & verification (self-hosted Sigstore, KMS-backed, headless OIDC) |
```
*(Surfaced here rather than inline — `demos/README.md` is not part of the diff, so it isn't anchorable.)*

## 🔵 Nitpick — `demos/**` links are not covered by the lychee CI gate

https://github.com/NVIDIA/aicr/blob/908b41eaa192b1d437cbfa2f71ebfd4b2466bf54/.github/workflows/fern-docs-ci.yaml#L56

The link checker triggers only on `docs/**`/`fern/**` and runs lychee against the glob `'docs/**/*.md'`; `demos/**` is outside both. All links in these files resolve today (manually verified), but future renames in `docs/` or `tests/` will break these inbound links with no CI signal. Optional follow-up: extend the trigger paths + glob to include `demos/**/*.md`.

- **KMS run.sh — forcing dummy AWS creds** (`bundle-attestation-kms-ministack/run.sh`): `unset AWS_SESSION_TOKEN AWS_SECURITY_TOKEN AWS_PROFILE` + hardcoded `AWS_ACCESS_KEY_ID=test` is a deliberate robustness/security improvement (prevents a stale ambient session leaking into the emulator as `ExpiredTokenException`). The only CI caller passes `test`/`test` and assumes no AWS role — nothing real is clobbered; region stays overridable.
- **`trust update || log_warning`** (`bundle-attestation-private-sigstore/run.sh`): swallowing is intentional and safe — it warms the public Sigstore TUF cache, is non-fatal so a pre-cached air-gapped host still passes, and the downstream sign step fails closed if the root is truly absent. Correct placement (runs for both binary paths); `msg`/`log_warning` are real helpers.
- **ttl.sh / `plainHTTP` / `insecureTLS` guidance** (`dynamic.md`): adequately warned and correctly scoped to test-only.
- **`.gitignore` additions**: no secret left un-ignored, nothing masked that should be committed (public keys/signatures/archives only; key material stays under `$HOME`, tokens in env).
- **Attested-binary gate framing**: accurate — a fully-private signing flow does reach public Sigstore once to verify the binary's own attestation; trust boundary correctly described.

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major   | 1 | Demos not indexed in `demos/README.md` (orphaned) |
| 🟡 Minor   | 2 | Layout tree misplaces `README.md`; step-2 token via argv |
| 🔵 Nitpick | 2 | Epoch `start-time` consequence; `demos/**` outside lychee |

Recommendation: **Approve with comments.** Add the two `demos/README.md` index rows and fix the step-4 tree before merge; the token-handling consistency is a quick win; the two nitpicks are optional.

njhensley (approved):
✅ **Approve**

Re-reviewed against head `e37c2f21`. All actionable findings from the earlier multi-persona review are resolved:

- 🟠 The two demos are now indexed in `demos/README.md`.
- 🟡 `dynamic.md` bundle-layout tree corrected — `README.md` moved to the bundle root, `upstream.env` added under the component.
- 🟡 `private-signing.md` step 2 now passes the OIDC token via `COSIGN_IDENTITY_TOKEN` (env var) with the ps/proc caveat, instead of the argv-exposed `--identity-token` flag.
- 🔵 Epoch `start-time` caveat now names the consequence and the additive-trust-root interaction.

Command-level accuracy was verified against the source and a live build in the first pass. The remaining note — extending lychee CI coverage to `demos/**` — was flagged as an optional out-of-PR follow-up. Nice, well-grounded docs.

Issue #1496, opened 2026-06-26
Add Topograph to AICR as a topology-aware-scheduler component


yuanchen8911:
## Summary

Add [Topograph](https://github.com/NVIDIA/topograph) to AICR as a new component so topology-aware schedulers (Slinky/Slurm, KAI) can consume cluster network topology.

## Background

Topograph was validated end-to-end on the `aicr-demo5` GKE H100 cluster (Slurm/Slinky + kai-scheduler). After resolving GCP Workload Identity / `roles/compute.viewer` IAM setup, Topograph queries the GCP Compute/Topology API and produces a valid Slurm `topology.conf` (SwitchName/Nodes mapping for the slinky nodesets). Discussion in the `aicr` Slack thread "Add Topograph to AICR" reached consensus on the modeling approach.

## Approach

Register topograph as a component, and have the leaf overlays pull it in via `dependencyRefs` (not installed by default) so it only ships where the scheduler is topology-aware — per Kayne's snippet below. Then add topograph to the recipes and wire dependencies as needed, with the right config per consumer (Slinky vs KAI).

Key points from the discussion:
- Topograph is **not** valuable standalone — it provides topology data to a scheduler, so it should only be deployed where the scheduler is topology-aware.
- Topograph's config depends on the consumer engine (Slinky/Slurm vs Kubernetes/KAI).
- Topograph must install **after** Slinky (operator/CRDs) and before the Slurm cluster instance.
- Not installed by default; leaf overlays opt in via `dependencyRefs`.

Proposed overlay wiring (from Kayne):

```yaml
    - name: slinky-topograph
      type: Helm
      valuesFile: components/topograph/values.yaml
      dependencyRefs:
        - slinky-slurm-imex-compute-domain
        - slinky-slurm-operator
        - slinky-slurm-operator-crds

    - name: slinky-slurm
      type: Helm
      valuesFile: components/slinky-slurm/values.yaml
      dependencyRefs:
        - slinky-topograph
```

Start from GKE.

## Tasks

- [ ] Add a `topograph` registry entry in `recipes/registry.yaml` (name, displayName, Helm repo/chart, default namespace, nodeScheduling paths, optional health-check)
- [ ] Add `recipes/components/topograph/values.yaml`
- [ ] Wire topograph into the GKE Slurm/Slinky leaf overlay via `dependencyRefs` (not default-installed)
- [ ] Model config per consumer (Slinky/Slurm vs KAI)
[...]

PR #1554, opened 2026-06-30
feat(slinky-topograph): topograph component with slinky engine

ravisoundar:
## Summary

Registered `slinky-topograph` component, and added health checks.
Updated the recipes `h100-kind-training-slurm` and `h100-gke-cos-training-slurm` to include topograph.

## Motivation / Context

Fixes: 
Related: #1496 

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `recipes/registry.yaml`, `recipes/components/`, `recipes/checks/`, `recipes/overlays/`

## Implementation Notes

**New slingy-topograph component** Added a Slinky/Slurm-scoped instance of Topograph — the topology utility that generates Slurm topology.conf by querying cloud provider topology APIs (GCP, AWS, OCI …). Feeds topology data to the Slinky-managed Slurm scheduler so it can make topology-aware placement decisions.

## Testing

```bash
make qualify
```

| Action    | Result |
| -------- | ------- |
| test coverage  | PASS — 77.3% coverage (threshold 75%)    |
| lint | PASS — 0 issues     |
| e2e -chainsaw tests    | PASS — 23/23    |

## Risk Assessment

**Rollout notes:** 

## Checklist


mchmarny (changes requested):
Commits are not properly signed, see https://github.com/NVIDIA/aicr/blob/main/CONTRIBUTING.md#developer-certificate-of-origin

yuanchen8911 (commented):
Review at head d3fda5f5. Main concern, detailed in the inline comments: the component installs and its workloads report Ready, but the topology pipeline is not functional end-to-end — the generated topology is never consumed by Slurm, the Kind leaf's DRA provider cannot generate topology at all, and the health check cannot detect either condition.

Open questions:
- What is the intended, ownership-safe delivery path for `topology.conf` into slurmctld — Topograph updating an already-referenced ConfigMap (Helm ownership and install ordering handled), or extending the chart to append the Topograph map as an external `configFileRef`? Either way the matching `TopologyPlugin` also needs to be set (`topology/block` for DRA-based leaves); `useDynamicNodes` alone is insufficient.
- Was the accelerated selector deliberately intended to narrow the observer's watched events? Independently, how should `node-observer.nodeSelector` be wired for observer-pod placement? Both intentions can coexist.
- Should the shared values' hardcoded `namespace: slurm` / `topologyConfigmapName: slinky-slurm-config-topo` be per-overlay overrides like provider/engine, so a namespace-moving overlay or mixin does not silently break them?
- Was end-to-end generation validated on a live cluster? The PR's Testing table exercises none of the Topograph generation -> ConfigMap -> Slurm-consumption pipeline.

recipes/components/slinky-topograph/values.yaml
```diff
@@ -0,0 +1,80 @@
...
+    type: ClusterIP
+    port: 49021
+  engine:
+    name: slinky
+    params:
+      plugin: topology/tree
+      podSelector:
+        matchLabels:
+          app.kubernetes.io/name: slurmd
+      topologyConfigmapName: slinky-slurm-config-topo
```
yuanchen8911, line 50: Topograph is configured here to update `slurm/slinky-slurm-config-topo`, but the rendered Slurm v1.1.0 Controller references only the chart-owned `slinky-slurm-config-extra` ([controller-cr.yaml#L48-L55](https://github.com/SlinkyProject/slurm-operator/blob/275035106a192e80e78a3de00773a7aa78224ad8/helm/slurm/templates/controller/controller-cr.yaml#L48-L55)). The chart has no value for appending an external `Controller.spec.configFileRefs` entry, and the operator projects only referenced ConfigMaps into `/etc/slurm` ([controller_app.go#L98-L118](https://github.com/SlinkyProject/slurm-operator/blob/275035106a192e80e78a3de00773a7aa78224ad8/internal/builder/controllerbuilder/controller_app.go#L98-L118)). Additionally, no overlay or values file sets `TopologyPlugin`, so Slurm retains `topology/flat` and would ignore `topology.conf` even if it were mounted. Net effect: the generated topology never reaches slurmctld, so topology-aware scheduling does not materialize on either leaf. Topograph's own Slinky example points `topologyConfigmapName` at a ConfigMap that Slinky mounts; this names a new ConfigMap nothing references. `useDynamicNodes` alone would not fix this — a mounted topology definition is still required.

Suggested direction: either have Topograph update an already-referenced ConfigMap (handling Helm ownership and install ordering), or extend the chart to append the Topograph map as an external `configFileRef` — and set the matching `TopologyPlugin` (`topology/block` for DRA-based leaves).

recipes/overlays/h100-kind-training-slurm.yaml
```diff
@@ -58,12 +58,26 @@ spec:
...
+    - name: slinky-topograph
+      type: Helm
+      valuesFile: components/slinky-topograph/values.yaml
+      dependencyRefs:
+        - slinky-slurm-operator
+        - slinky-slurm-operator-crds
+      overrides:
+        global:
+          provider:
+            name: dra
```
yuanchen8911: `provider: dra` cannot produce topology on this leaf. The overlay is explicitly CPU-only ("Kind has no GPUs"), and topograph v0.5.0's DRA provider keys domains off the `nvidia.com/gpu.clique` node label, returning HTTP 502 when no node matches ([provider.go#L87-L125](https://github.com/NVIDIA/topograph/blob/253cb10aafae29d3463d95cee414c1eaabdc5757/pkg/providers/dra/provider.go#L87-L125)). Even if sufficient metadata were added for generation to succeed, the DRA provider returns block domains only, which the shared values' `plugin: topology/tree` rejects with "missing tree topology" ([topology.go#L82-L94](https://github.com/NVIDIA/topograph/blob/253cb10aafae29d3463d95cee414c1eaabdc5757/pkg/translate/topology.go#L82-L94)). Every accepted generation request on this recipe is therefore guaranteed to fail. Consider the `test` provider or a block-topology setup for this leaf.

recipes/checks/slinky-topograph/health-check.yaml
```diff
@@ -0,0 +1,129 @@
...
+# limitations under the License.
+
+# slinky-topograph Health Check
+#
+# Validates that slinky-topograph and its sub-charts (node-observer,
+# node-data-broker) are running and healthy, and that the topology generation
+# pipeline has produced a result. Specifically:
+#   1. All three Deployments are available (topograph API, node-observer, node-data-broker).
+#   2. No pods in the topograph namespace are in a failed/stuck state.
+#   3. The topology ConfigMap written by topograph exists in the slurm namespace
```
yuanchen8911, line 22: Item 3 in this header promises an assertion that the topology ConfigMap exists in the slurm namespace with a non-empty `topology.conf` key ("confirming end-to-end pipeline success — not just process readiness"), but no ConfigMap assertion exists among the steps below. That is precisely the assertion that would catch real failure modes: `/v1/generate` returns 202 with a UID after enqueueing ([http_server.go#L136-L159](https://github.com/NVIDIA/topograph/blob/253cb10aafae29d3463d95cee414c1eaabdc5757/pkg/server/http_server.go#L136-L159)); node-observer discards that response and never queries `/v1/topology` ([status_informer.go#L442-L467](https://github.com/NVIDIA/topograph/blob/253cb10aafae29d3463d95cee414c1eaabdc5757/pkg/node_observer/status_informer.go#L442-L467) — it retries failed submissions, not eventual generation results); and `/healthz` returns 200 unconditionally. So topology generation can fail persistently (e.g. the Kind DRA configuration, or an unreplaced GKE Workload Identity placeholder) while every pod stays Ready and this check passes.

Also, line 20 says "All three Deployments" — node-data-broker is asserted as a DaemonSet below.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -73,12 +73,36 @@ spec:
...
+        serviceAccount:
+          annotations:
+            # GKE Workload Identity: replace with the GCP service account
+            # bound to this cluster's Workload Identity pool.
+            # Or update it at bundle time by passing the command line argument to `aicr bundle`:
+            # --set-json 'slinkytopograph:serviceAccount.annotations={"iam.gke.io/gcp-service-account":"<sa-name>@<project-id>.iam.gserviceaccount.com"}'
+            # Format: <sa-name>@<project-id>.iam.gserviceaccount.com
+            # The SA needs roles/compute.viewer on the GCP project so
+            # topograph can call the Compute/Topology API.
+            iam.gke.io/gcp-service-account: <gcp-sa>@<project-id>.iam.gserviceaccount.com
```
yuanchen8911: This bakes a literal placeholder as an active annotation value — the only live placeholder in any shipped overlay (existing overlays keep placeholders in comments). Kubernetes accepts the annotation string, so bundle and deploy succeed, and the documented Workload Identity path is broken by default: topograph's GCP topology calls fail at request time unless both the placeholder and the IAM binding are replaced. Combined with the process-readiness-only health check, the failure is invisible to validation. Consider omitting the annotation by default and requiring the documented `--set-json` override instead.

recipes/registry.yaml
```diff
@@ -695,6 +695,31 @@ components:
...
+      defaultNamespace: topograph
+    nodeScheduling:
+      system:
+        nodeSelectorPaths:
+          - nodeSelector
+        tolerationPaths:
+          - tolerations
+      accelerated:
+        nodeSelectorPaths:
+          - node-observer.topograph.trigger.nodeSelector
```
yuanchen8911, line 721: `node-observer.topograph.trigger.nodeSelector` is the observer's event filter (which node changes trigger topology regeneration — sibling of `trigger.podSelector` in the sub-chart values), not a pod-scheduling field, yet it receives the accelerated node selector via `nodeScheduling`. Meanwhile the pod's actual scheduling selector `node-observer.nodeSelector` is not wired under either `system` or `accelerated`, while `node-observer.tolerations` is — leaving an unpinned observer pod carrying accelerated-node tolerations (nondeterministic placement). If narrowing the watch to accelerated nodes is intended, both wirings can coexist — but pod placement still needs `node-observer.nodeSelector`.

recipes/registry.yaml
```diff
@@ -695,6 +695,31 @@ components:
...
+    displayName: slinky-topograph
+    valueOverrideKeys:
+      - slinkytopograph
+    healthCheck:
+      assertFile: checks/slinky-topograph/health-check.yaml
+    helm:
+      defaultRepository: https://nvidia.github.io/topograph
+      defaultChart: topograph/topograph
+      defaultVersion: "0.5.0"
+      defaultNamespace: topograph
```
yuanchen8911, line 708: `tools/cleanup`'s `AICR_NAMESPACES` (per its own comment, sourced from `defaultNamespace` entries in this registry) was not updated with `topograph`, so full-stack teardown leaves the namespace and its workloads behind. Adjacent pre-existing gap: `slurm` (slinky-slurm's defaultNamespace) is missing from that list too.

docs/user/component-catalog.md
```diff
@@ -54,6 +55,7 @@ Not every component appears in every recipe. The recipe engine selects component
...
++ - **Topology-aware optional components** (`slinky-topograph`) are not installed by default. Opting in requires two things in the leaf overlay: an explicit `componentRef` entry for `slinky-topograph` (which installs it) and a `dependencyRef` on `slinky-slurm` pointing to it (which enforces ordering). `dependencyRefs` alone does not install the component. See the wiring example in the [Recipe Development Guide](../integrator/recipe-development.md#slinky-slurm-inline-components).
```
yuanchen8911: This line begins with a stray leading `+ ` (`+ - **Topology-aware...`) while sibling bullets use `- `; under CommonMark/GitHub rendering it becomes an unintended nested bullet and breaks the list structure.

yuanchen8911: The link targets `../integrator/recipe-development.md#slinky-slurm-inline-components`, but no heading in that file produces this slug — the wiring example added by this PR is prose without a heading of its own. CI will not catch it: the lychee invocation in [fern-docs-ci.yaml#L53-L56](https://github.com/NVIDIA/aicr/blob/d3fda5f5b0ab21e2d678d59fd7f0f9ec02c62428/.github/workflows/fern-docs-ci.yaml#L53-L56) omits `--include-fragments`, so fragment targets are not validated (adding that flag does catch this one). Either add a matching heading above the wiring example or drop the fragment from the link.

docs/user/component-catalog.md
```diff
@@ -54,6 +55,7 @@ Not every component appears in every recipe. The recipe engine selects component
 - **Cloud-specific components** (aws-efa, aws-ebs-csi-driver) are added when the service matches. OCP recipes replace base components (gpu-operator, nfd, network-operator) with OLM+CR pairs (e.g., `gpu-operator-ocp-olm` + `gpu-operator-ocp`).
 - **Intent-specific components** (agentgateway, agentgateway-crds) are added based on workload intent (e.g., inference recipes include the inference gateway).
 - **Platform-specific components** (slinky-slurm-operator, slinky-slurm, kubeflow-trainer, dynamo-platform) are added when the recipe selects a matching `--platform`. For `--platform slurm`, all three Slinky pieces (`slinky-slurm-operator-crds`, `slinky-slurm-operator`, `slinky-slurm`) are declared inline per slurm leaf overlay — the same shape `dynamo-platform` uses across `*-inference-dynamo` leaves. Leaves that want the operator only inline the CRDs + operator and omit the `slinky-slurm` componentRef. For an end-to-end walkthrough (recipe → bundle → install → validate → `srun` smoke job on EKS, GKE, or Kind), see [`demos/cuj1-slinky-slurm.md`](https://github.com/NVIDIA/aicr/blob/main/demos/cuj1-slinky-slurm.md).
```
yuanchen8911: `demos/cuj1-slinky-slurm.md` (linked from this bullet) is now stale for both leaves this PR changes: it still says each slurm leaf resolves to three inline Slinky components, its deploy order omits `slinky-topograph`, and its full-stack cleanup omits the `slinky-topograph` release and the `topograph` namespace.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -73,12 +73,36 @@ spec:
...
+
     - name: slinky-slurm
       type: Helm
       valuesFile: components/slinky-slurm/values.yaml
       dependencyRefs:
         - slinky-slurm-operator
         - slinky-slurm-operator-crds
+        - slinky-topograph
```
yuanchen8911, line 95: The ordering contract introduced by this dependencyRef (topograph after the operator, before slinky-slurm) is not guarded: the `h100-gke-cos-training-slurm` and `h100-kind-training-slurm` cases in `pkg/recipe/deployment_order_guard_test.go` still assert only the original deps/orderings, and the assertions are subset-based — so the test passes today without covering the new edges, and a future overlay edit dropping this dependencyRef would regress silently. Consider extending `requiredDeps`/`requiredOrdering` with the `slinky-topograph` edges in both cases.

recipes/components/slinky-topograph/values.yaml
```diff
@@ -0,0 +1,80 @@
...
+  topograph:
+    trigger:
+      podSelector:
+        matchLabels:
+          app.kubernetes.io/name: slurmd
+
+# node-data-broker sub-chart
+# Bundled as a local chart dependency.
+node-data-broker:
+  fullnameOverride: slinky-topograph-node-data-broker
```
yuanchen8911: Missing trailing newline at EOF — the only values file under `recipes/components/` in this state. yamllint will not flag it (`new-line-at-end-of-file` is disabled), so this is cosmetic, but it adds noise to future diffs that append to the file.

yuanchen8911 (changes requested):
Requesting changes per the inline review comments (pullrequestreview-4620253145): the component installs and health-checks green, but the topology pipeline is not functional end-to-end — the generated topology is never consumed by Slurm (unmounted ConfigMap, no TopologyPlugin), the Kind leaf's DRA provider cannot generate topology, and the health check omits the ConfigMap assertion its header promises, so none of this surfaces in validation. Also: the GKE overlay ships a live Workload Identity placeholder, and tools/cleanup, the cuj1 demo, and the deployment-order guard test were not updated for the new component.

recipes/overlays/h100-kind-training-slurm.yaml
ravisoundar: Replaced the `dra` provider with the `test` provider.

recipes/overlays/h100-gke-cos-training-slurm.yaml
ravisoundar: Removed it, and the instruction to add it during bundle generation is provided in the `receipe_development.md`

recipes/checks/slinky-topograph/health-check.yaml
ravisoundar: Add a test to validate the configMap generation with the desired values.

recipes/registry.yaml
ravisoundar: Updated the nodeSelectors.

docs/user/component-catalog.md
ravisoundar: Updated it.

recipes/registry.yaml
ravisoundar: Added the namespaces to the cleanup script.

mchmarny:
## 📋 Review status — verified against head `be113d7`

Re-checked every finding from the review passes against the latest push. Most items are addressed; the central functional finding is still open.

### Addressed ✅

| Finding | Status |
|---|---|
| CodeRabbit items (queue-depth claim, provider list, `node-data-broker` comment + `nodee-data-broker` typo, DaemonSet gating, opt-in wording) | ✅ Fixed, threads resolved |
| Kind leaf `provider: dra` cannot generate topology | ✅ Replaced with `test` provider |
| Live Workload Identity placeholder in GKE overlay | ✅ Removed; documented as a `--set` override in recipe-development docs |
| `node-observer` nodeSelector wiring (trigger filter vs pod placement) | ✅ Updated |
| `tools/cleanup` missing `topograph` (and `slurm`) namespaces | ✅ Added |
| Stale `demos/cuj1-slinky-slurm.md` | ✅ Updated in diff |
| Deployment-order guard test missing `slinky-topograph` edges | ✅ `pkg/recipe/deployment_order_guard_test.go` extended |
| Health check missing the promised ConfigMap assertion | ✅ `validate-topology-configmap` step added (asserts non-empty `topology.conf`) |
| Markdown stray `+` bullet, EOF newline, broken anchor | ✅ Fixed |

### Still open 🔴

1. **Topology is generated but never consumed by Slurm** ([unresolved thread](https://github.com/NVIDIA/aicr/pull/1554#discussion_r3515338889)). Topograph writes `slinky-slurm-config-topo`, but the rendered Slurm v1.1.0 Controller references only chart-owned ConfigMaps, and no overlay sets `TopologyPlugin` — so slurmctld retains `topology/flat` and topology-aware scheduling doesn't materialize. The head diff still shows `topologyConfigmapName: slinky-slurm-config-topo` with no `configFileRef` / `TopologyPlugin` change, and the new health check asserts a ConfigMap that nothing mounts. This was the basis of the changes-requested review and needs either a fix (referenced ConfigMap or external `configFileRef`, plus matching `TopologyPlugin`) or a design answer on the thread.
2. **Commit signing** — DCO `Signed-off-by` trailers are now present on all commits 👍, but the GPG signatures verify as `unknown_key` (signing key not registered on the GitHub account), so commits still show **Unverified**. Please upload the public key to your GitHub account ([docs](https://docs.github.com/en/authentication/managing-commit-signature-verification/adding-a-gpg-key-to-your-github-account)) or re-sign with a registered key.

Great progress on the mechanical sweep — item 1 is the remaining substantive gate.

recipes/checks/slinky-topograph/health-check.yaml
```diff
@@ -0,0 +1,149 @@
...
+              status:
+                (availableReplicas > `0`): true
+    - name: validate-node-data-broker-daemonset
+      # node-data-broker is enabled by default. When an overlay sets
+      # node-data-broker.enabled: false the DaemonSet is not deployed and this
+      # step must be skipped rather than retried until timeout. A script step
+      # handles the conditional: if the DaemonSet is absent, exit 0 (skip);
+      # if present, wait for a full rollout before declaring success.
+      try:
+        - script:
```
ArangoGutierrez, line 64: This script step trips the read-only allowlist for registry-declared health checks: TestValidateTestReadOnly_RegistryContent fails at this head, since only assert and error operations are permitted (the deployment-validator Job runs under cluster-admin, so the allowlist fails closed). The conditional DaemonSet skip needs an assert/error-only shape; make test fails until then.

recipes/checks/slinky-topograph/health-check.yaml
```diff
@@ -0,0 +1,149 @@
...
+      try:
+        - assert:
+            resource:
+              apiVersion: v1
+              kind: ConfigMap
+              metadata:
+                name: slinky-slurm-config-topo
+                namespace: slurm
+              data:
+                topology.conf: |
```
ArangoGutierrez, line 146: This asserts the exact topology.conf rendered by the test provider's small-tree.yaml fixture, but the assertFile is hydrated onto every overlay carrying the component with no per-provider scoping, so the GKE leaf (provider gcp) fails this step deterministically against real topology. A provider-agnostic assertion (key present and non-empty) would hold everywhere; the exact-content check fits a Kind-only test.

ArangoGutierrez (commented):
Found 2 issues, both in the new health check. Also two branch-hygiene notes per CLAUDE.md: the branch merges main into the feature branch instead of rebasing (the needs-rebase label agrees), and the four commits should be squashed into a single signed commit before push.

1. This script step trips the read-only allowlist for registry-declared health checks: TestValidateTestReadOnly_RegistryContent fails at this head, since only assert and error operations are permitted (the deployment-validator Job runs under cluster-admin, so the allowlist fails closed). The conditional DaemonSet skip needs an assert/error-only shape; make test fails until then. (recipes/checks/slinky-topograph/health-check.yaml:64)
2. This asserts the exact topology.conf rendered by the test provider's small-tree.yaml fixture, but the assertFile is hydrated onto every overlay carrying the component with no per-provider scoping, so the GKE leaf (provider gcp) fails this step deterministically against real topology. A provider-agnostic assertion (key present and non-empty) would hold everywhere; the exact-content check fits a Kind-only test. (recipes/checks/slinky-topograph/health-check.yaml:146)

recipes/components/slinky-topograph/values.yaml
ArangoGutierrez, line 50: Topograph writes topology.conf into slinky-slurm-config-topo, but nothing mounts that ConfigMap: the slurm chart v1.1.0 renders configFileRefs only from its own configFiles values (the -config-extra ConfigMap), with no mechanism to reference an external one. Pointing topologyConfigmapName at the chart's -config-extra ConfigMap and seeding a topology.conf stub via slinky-slurm's configFiles would close the loop; as-is the generated topology never reaches slurmctld. Same gap yuanchen8911's changes-requested review raised.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -73,12 +73,26 @@ spec:
...
+            name: slinky
+
     - name: slinky-slurm
       type: Helm
       valuesFile: components/slinky-slurm/values.yaml
       dependencyRefs:
         - slinky-slurm-operator
         - slinky-slurm-operator-crds
+        - slinky-topograph
       overrides:
```
ArangoGutierrez, line 96: Nothing in the PR or the base slinky-slurm values sets TopologyPlugin (extraConfMap carries only SelectType and ScronParameters), so slurmctld ignores topology.conf even once it is mounted. The slurm leaves that carry topograph need e.g. controller.extraConfMap TopologyPlugin: topology/tree.

recipes/checks/slinky-topograph/health-check.yaml
ArangoGutierrez, line 64: This script step trips the read-only allowlist for registry-declared health checks: TestValidateTestReadOnly_RegistryContent fails at this head, since only assert and error operations are permitted. Chainsaw can assert the DaemonSet (or its absence) directly without shelling out, or the allowlist needs a deliberate, signed-off extension.

recipes/checks/slinky-topograph/health-check.yaml
ArangoGutierrez, line 146: This asserts the exact small-tree.yaml fixture content, but healthCheck.assertFile hydrates this file onto every overlay carrying the component — on the GKE leaf (provider gcp) it fails deterministically against real topology. Scope the exact-content assertion to the test provider and assert only a non-empty key elsewhere.

ArangoGutierrez (commented):
Found 4 issues. The three functional ones share a root cause: the topology pipeline currently ends at the ConfigMap — nothing mounts it into the controller, and slurmctld is never told to load it.

1. Topograph writes topology.conf into slinky-slurm-config-topo, but nothing mounts that ConfigMap: the slurm chart v1.1.0 renders configFileRefs only from its own configFiles values (the -config-extra ConfigMap), with no mechanism to reference an external one. Pointing topologyConfigmapName at the chart's -config-extra ConfigMap and seeding a topology.conf stub via slinky-slurm's configFiles would close the loop; as-is the generated topology never reaches slurmctld. Same gap yuanchen8911's changes-requested review raised. (recipes/components/slinky-topograph/values.yaml:50)
2. Nothing in the PR or the base slinky-slurm values sets TopologyPlugin (extraConfMap carries only SelectType and ScronParameters), so slurmctld ignores topology.conf even once it is mounted. The slurm leaves that carry topograph need e.g. controller.extraConfMap TopologyPlugin: topology/tree. (recipes/overlays/h100-gke-cos-training-slurm.yaml:96)
3. This script step trips the read-only allowlist for registry-declared health checks: TestValidateTestReadOnly_RegistryContent fails at this head, since only assert and error operations are permitted. Chainsaw can assert the DaemonSet (or its absence) directly without shelling out, or the allowlist needs a deliberate, signed-off extension. (recipes/checks/slinky-topograph/health-check.yaml:64)
4. This asserts the exact small-tree.yaml fixture content, but healthCheck.assertFile hydrates this file onto every overlay carrying the component — on the GKE leaf (provider gcp) it fails deterministically against real topology. Scope the exact-content assertion to the test provider and assert only a non-empty key elsewhere. (recipes/checks/slinky-topograph/health-check.yaml:146)

ArangoGutierrez:
Superseding this PR with #1737 so the changes can land while @ravisoundar is on PTO (done with approval). His work is preserved as the first commit of the successor, authored by him; the second commit addresses the open review threads from @mchmarny, @yuanchen8911, and @ArangoGutierrez (signing, read-only health checks, and wiring topology.conf end-to-end into slurmctld — verified live on a Kind cluster). Thank you Ravi for the component — please pick up review credit on #1737 when you're back.

PR #1560, opened 2026-06-30
fix(lke): add provider ID mapping for Akamai Cloud / Linode LKE

eljohnson92:
## Summary

Map Akamai Cloud / Linode LKE nodes (`linode://<instance-id>` provider IDs) to the `lke` service in the K8s collector and fingerprint paths, so live LKE clusters are correctly detected, and the correct overlays are chosen.

## Motivation / Context

The `lke` service criteria, overlays, and docs already existed, but real LKE clusters were never detected as `lke`: the collector's `parseProvider` fell through to the raw `linode` prefix, and the fingerprint's `normalizeProviderID` had no Linode case. This closes that gap end-to-end (snapshot collection and fingerprint-based service detection).

Fixes: N/A
Related: N/A

## Type of Change

## Component(s) Affected

- [x] Other: `pkg/fingerprint`

## Implementation Notes

Provider normalization happens in two layers and both are updated:

- **Collector** (`pkg/collector/k8s/node.go`): `parseProvider` adds `case "linode": return "lke"`. The scheme is lowercased first, so `LINODE://` normalizes too. This is the live-cluster path.
- **Fingerprint** (`pkg/fingerprint/from_measurements.go`): adds a `serviceLKE` constant and teaches `normalizeProviderID` to handle both the raw `linode://<id>` URL and the bare `linode` string. This is defensive for hand-crafted snapshots that carry the unnormalized value (mirrors the existing OKE `oci`/`ocid1.` handling).

Also updates the `aicr-analyzing-snapshots` skill doc and package godoc (`doc.go`) with the `linode://` → `lke` mapping and the `lke.linode.com/pool-id` node label. No new service enum value is introduced — `CriteriaServiceLKE` and the LKE recipes already existed.

## Testing

```bash
make qualify
```

## Risk Assessment

  Additive single-case mapping in two well-isolated functions, fully covered by unit tests. Nodes that previously surfaced the raw linode prefix now resolve to lke; no other provider behavior changes.

**Rollout notes:** 
 N/A — backwards compatible, no migration required.
## Checklist

### DIFF
--- a/pkg/fingerprint/from_measurements.go
+++ b/pkg/fingerprint/from_measurements.go
@@ -54,8 +54,7 @@ const (
 	noteUnknownSKU          = "unknown-sku"
 )
 
-//	FromMeasurements builds a Fingerprint from a snapshot's measurement
-//
+// FromMeasurements builds a Fingerprint from a snapshot's measurement
 // slice. Dimensions whose source signal is missing keep their zero
 // value (empty string for Dimension/OSDimension, 0 for IntDimension);
 // callers compare those against criteria using Match, which treats
--- a/pkg/fingerprint/from_measurements.go
+++ b/pkg/fingerprint/from_measurements.go
@@ -28,6 +28,7 @@ import (
 // don't want to import them just for the names.
 const (
 	serviceOKE              = "oke"
+	serviceLKE              = "lke"
 	subtypeK8sServer        = "server"
 	subtypeK8sNode          = "node"
 	subtypeGPUHardware      = "hardware"
@@ -53,7 +54,8 @@ const (
 	noteUnknownSKU          = "unknown-sku"
 )
 
-// FromMeasurements builds a Fingerprint from a snapshot's measurement
+//	FromMeasurements builds a Fingerprint from a snapshot's measurement
+//
 // slice. Dimensions whose source signal is missing keep their zero
 // value (empty string for Dimension/OSDimension, 0 for IntDimension);
 // callers compare those against criteria using Match, which treats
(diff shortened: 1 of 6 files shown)


pkg/fingerprint/from_measurements.go
```diff
@@ -53,7 +54,8 @@ const (
 	noteUnknownSKU          = "unknown-sku"
 )
 
-// FromMeasurements builds a Fingerprint from a snapshot's measurement
+//	FromMeasurements builds a Fingerprint from a snapshot's measurement
```
mchmarny: Looks like an accidental edit unrelated to the LKE change — the leading tab turns this synopsis into a godoc code block, and the new blank `//` line splits the opening sentence ("...measurement" / blank / "slice."). Revert to the single line:

```go
// FromMeasurements builds a Fingerprint from a snapshot's measurement
```

mchmarny (commented):
Welcome to AICR, @eljohnson92 — and thanks for a really clean first PR. The two-layer fix is exactly right: mapping `linode://` in both `parseProvider` (live-cluster path) and `normalizeProviderID` (hand-crafted snapshots) mirrors the existing OKE `oci`/`ocid1.` handling precisely, and the test coverage is thorough — uppercase scheme, bare `linode`, and the `linode://` URL are all exercised. Docs and godoc updated to match. Nicely scoped.

One non-blocking nit (flagged inline): an accidental doc-comment edit on `FromMeasurements` that's unrelated to the LKE change — please revert that one line.

Otherwise this looks good to merge once CI runs (currently gated on first-contributor vetting, so the checks haven't executed yet).

pkg/fingerprint/from_measurements.go
eljohnson92: Good catch, just reverted this edit. should be all set now

njhensley (approved):
LGTM. Clean, minimal, and symmetric with the existing OKE handling.

Verified:
- Collector `parseProvider` lowercases the scheme before the new `case "linode": return "lke"`, so `LINODE://` normalizes correctly (covered by the added uppercase test).
- Fingerprint `normalizeProviderID` maps both `linode://<id>` and bare `linode` to `lke`, mirroring the defensive `oci://`/`oci` cases.
- `CriteriaServiceLKE = "lke"` already exists, so no new enum value is introduced — the doc-audit obligation for new enums doesn't apply.
- `serviceLKE` constant is used (no unused-const risk); docs and godoc updated consistently; both code paths have test coverage.

No correctness issues found.

PR #1572, opened 2026-07-01
fix(recipes): guard overlay version pins against registry defaults

### LINKED ISSUE #1424: BOM renders from registry defaults, not resolved overlay pins — per-overlay version drift is invisible
## Problem

`make bom-docs` renders `docs/user/container-images.md` from each component's **registry `defaultVersion`**, not from the **version a recipe actually resolves** after overlay pins are applied. Since overlays routinely pin component versions (`componentRefs[].version`) that override the registry default, the BOM can advertise a version that **no real recipe installs** — and per-overlay version drift is therefore invisible in the BOM.

## Evidence (surfaced by #1418)

`aws-efa` is pinned in two places:
- `recipes/registry.yaml` → `defaultVersion`
- `recipes/overlays/eks.yaml` → `componentRefs[].version` (overrides the registry default for every EKS recipe — and EKS is the only `aws-efa` consumer)

In #1418 the registry default was bumped (`v0.5.26` → `v0.5.29`) and the BOM regenerated to `v0.5.29`, but `eks.yaml` still pinned `v0.5.26`. Result: the BOM (and PR summary) advertised `v0.5.29` while every EKS recipe actually rendered `v0.5.26`. Nothing flagged the mismatch — the BOM is generated from the registry default, so it looked correct. The fix in #1418 only re-aligned the two pins by hand; the underlying invisibility remains.

## Impact

- The BOM is the user-facing source of truth for shipped image/chart versions (and feeds supply-chain expectations). It can misreport the actual installed version for any component whose overlay pin diverges from the registry default.
- Version bumps can appear applied (green BOM regen) while real recipes are unchanged — a silent correctness gap, exactly the class of bug #1418 hit.

## Underlying question — version-management model

Version pinning is currently split and inconsistent:
- Some components carry a registry `defaultVersion`.
- Most are pinned per-overlay (`base.yaml`, `aks.yaml`, `oke.yaml`, …) — 6+ components, 35+ pins.
- `aws-efa` has **both** (registry default *and* an overlay pin), which is what allows divergence.

So the real decision is **where the canonical version lives** (registry default vs overlay pin vs both), and how the BOM should reflect it.

## Options

1. **Render the BOM from resolved recipes** (per service/accelerator/intent) instead of registry defaults, so it reflects what actually installs. Most correct; changes the BOM generator and likely the doc's shape (a component may render multiple versions across overlays).
2. **Consistency guard** — a test asserting overlay `version` pins match the registry `defaultVersion` unless an explicit exemption is declared. Cheap; prevents *unintended* divergence, but needs an exemption mechanism for legitimate per-overlay pins.
3. **Single source of truth** — pick one location per component (registry default OR overlay pin, not both) and lint that the other is absent. Removes the divergence class structurally.

## Acceptance

- The BOM cannot silently advertise a version that no recipe installs (either it renders resolved versions, or a check fails when a pin diverges unexpectedly).
- A bump to a component version is reflected in *both* the BOM and the recipes that install it, or CI fails.

## Notes

- Not a regression introduced by #1418 — pre-existing on `main`; #1418 removed the one concrete instance (re-aligned the `aws-efa` pins) and surfaced the general gap.

mchmarny:
## Summary

The container-images BOM renders each chart at its registry `defaultVersion`, but a `componentRef` version/tag pin in `base.yaml`/an overlay/a mixin overrides that default at resolution. Add a guard test so an overlay pin can no longer silently diverge from the registry default (which the BOM reads), and fix the one live divergence it caught.

## Motivation / Context

The BOM (`docs/user/container-images.md`) is the user-facing source of truth for shipped chart/image versions, yet it renders from the registry `defaultVersion` — a *fallback* that overlay/mixin `version` pins override. So the BOM equals what recipes actually install only when no pin diverges. When a component's sole consumer diverges, the registry default is installed by zero recipes and the BOM is pure fiction (the #1418 `aws-efa` bug).

The guard caught a **live** instance on `main`: `base.yaml` pinned `nvsentinel v1.3.0` while the registry default — and the committed BOM (6 images) — advertised `v1.9.0`. Every recipe shipped v1.3.0.

Fixes: #1424
Related: #1418

## Type of Change

- [x] Build/CI/tooling (guard test wired into `make test` → `make qualify`)

## Component(s) Affected

## Implementation Notes

- **`pkg/recipe/version_pin_guard_test.go`** — `TestOverlayVersionPinsMatchRegistry` asserts every `base`/overlay/mixin `componentRefs[].version` (Helm) / `.tag` (Kustomize) equals the component's registry default, unless declared in `versionPinExemptions` with a justification. Also fails on **stale exemptions** (pin re-aligned) and **vacuous discovery** (no pins found). Reuses `loadMetadataStore` + `GetComponentRegistry`; runs under `make test`, so no new gate wiring.
- **`recipes/overlays/base.yaml`** — nvsentinel `v1.3.0 → v1.9.0`. This changes what every recipe deploys; the BOM already advertised v1.9.0, so this makes reality match the BOM (regenerated `make bom-docs` → **zero diff**).
- **Exemption** — `aks/kube-prometheus-stack 83.7.0` (intentional, validated cluster state in #700; the registry default 84.4.0 is still installed by every non-AKS recipe, so the BOM's default is not fictional).
- **Example recipes (`examples/recipes/*.yaml`)** — refreshed the hand-maintained example snapshots so their pins match the current registry defaults: nvsentinel `v1.3.0 → v1.9.0`, gpu-operator `v25.10.1 → v26.3.2`, and (AKS) kube-prometheus-stack `84.4.0 → 83.7.0`, plus dropping a stale `driver.version: 580.82.07` override in the gb200 example. Examples bypass overlay resolution and are not covered by the new guards (a follow-up will track guarding them); these edits keep them consistent for now.
- **Approach choice** — a consistency guard (over rendering the BOM from resolved recipes) keeps the BOM generator unchanged and directly implements the acceptance clause "a check fails when a pin diverges unexpectedly." Chosen with the maintainer.
- Docs: single-source version model documented in `docs/contributor/recipe.md`.

## Testing

```bash
golangci-lint run -c .golangci.yaml ./pkg/recipe/...   # 0 issues
go test -race ./pkg/recipe/...                          # ok
go test ./...                                           # ok (except pkg/trust + pkg/bundler/attestation: known sandbox TUF/network)
make lint                                               # go vet, golangci (0), yamllint, license, agents-sync all OK
make bom-docs                                           # zero diff after nvsentinel align
```

Guard verified RED (caught the live nvsentinel divergence) before the fix, GREEN after.

## Risk Assessment

- [x] **Medium** — Touches the base recipe: nvsentinel `v1.3.0 → v1.9.0` changes the deployed chart for every recipe.

**Rollout notes:** The registry default and the rendered BOM were already v1.9.0; this aligns the recipes to match. nvsentinel values render cleanly at v1.9.0 (BOM regen produced no diff). Easy to revert (single-line pin + test).

## Checklist


mchmarny:
@CodeRabbit full review

pkg/recipe/version_pin_guard_test.go
```diff
@@ -0,0 +1,239 @@
...
+			}
+			if def == "" {
+				// No registry default to diverge from. `make bom-pinning-check`
+				// separately enforces that every Helm component is pinned.
+				continue
+			}
+
+			checked++
+			if pin == def {
+				continue
```
yuanchen8911, line 131: The guard asserts each pin matches the registry default, but nothing asserts the committed BOM matches that default. A coordinated bump — registry defaultVersion and every overlay pin moved to a new version together — passes this test even if docs/user/container-images.md was never regenerated. make bom-check is opt-in and not wired into qualify/lint/the merge gate (Makefile:294); lint runs only bom-pinning-check, which asserts pins exist, not that the BOM is fresh.

The "chosen with the maintainer" note concerns preferring a consistency guard over resolved-BOM generation. That is a separate decision from #1424's acceptance requirement that CI catch a stale committed BOM, which this guard does not do. The documented "desirable follow-up" (recipe.md:479-483) has no open tracking issue: the BOM refresh workflow comment (.github/workflows/bom-refresh.yaml:24-27) explicitly points the planned PR-time bom-check gate at #739, which is closed.

If this PR carries Fixes #1424, this is a merge blocker. Recommended: either add a lightweight BOM-version freshness gate (path-filtered bom-check triggered on BOM-affecting paths — recipes/**, docs/user/container-images.md, tools/bom/**, pkg/bom/** — or a deterministic test that renders the version table from the registry and diffs it against the committed doc), or downgrade Fixes #1424 to Related #1424 and leave #1424 open with explicit follow-up tracking.

recipes/overlays/base.yaml
```diff
@@ -50,7 +50,7 @@ spec:
     - name: nvsentinel
       type: Helm
       source: oci://ghcr.io/nvidia
-      version: v1.3.0
+      version: v1.9.0
```
yuanchen8911, line 53: This aligns the resolved recipes, but the hydrated example recipes are not updated. They bypass overlay resolution (pkg/recipe/loader.go:81-85), and the guard added here (which scans only base/overlays/mixins) does not cover them. They are stale beyond nvsentinel — for example aks-training.yaml pins gpu-operator v25.10.1 (current default v26.3.2), kube-prometheus-stack 84.4.0 (resolved AKS actually installs 83.7.0), and nvsentinel v1.3.0 (now v1.9.0):

- examples/recipes/eks-training.yaml
- examples/recipes/aks-training.yaml
- examples/recipes/eks-gb200-ubuntu-training-with-validation.yaml

Recommended fix: regenerate these examples from resolved output rather than hand-editing individual pins. If a guard is added for examples, it must compare each example against freshly resolved output for that example's criteria — not against registry defaults — because intentional overrides such as AKS kube-prometheus-stack 83.7.0 are legitimate and would be flagged incorrectly by a registry-default comparison.

Address this, or explicitly scope it out of the PR with follow-up tracking.

pkg/recipe/version_pin_guard_test.go
```diff
@@ -0,0 +1,239 @@
...
+					t.Errorf("out-of-date versionPinExemptions entry for %s/%s: exemption "+
+						"blesses pin=%q vs default=%q, but the recipe now has %s=%q vs default=%q.\n"+
+						"  Update the exemption's expectedPin/expectedDefault and re-justify the "+
+						"divergence, or re-align the pin. See issue #1424.",
+						source, ref.Name, e.expectedPin, e.expectedDefault, field, pin, def)
+					continue
+				}
+				t.Logf("exempted divergence: %s/%s pins %s=%q vs registry default %q — %s",
+					source, ref.Name, field, pin, def, e.reason)
+				continue
```
yuanchen8911, line 153: An exemption passes purely from its recorded expectedPin/expectedDefault pair. If every recipe that installs the registry default is later removed or disabled, this test stays green while the BOM advertises a version no recipe installs. Because an exemption is explicit, the divergence is arguably no longer "silent" — but the test still does not enforce its own stated sole-consumer policy ("Do NOT exempt a component whose only consumer diverges").

Recommended fix (non-blocking, follow-up): for each exemption, require at least one resolved, enabled recipe to install the registry default; fail if none does.

yuanchen8911 (dismissed):
The version-pin guard works for the current embedded Helm catalog, but does not fully enforce #1424's end-to-end acceptance criteria. Details inline: finding 1 is a merge blocker if this PR carries Fixes #1424; finding 2 should be addressed or explicitly scoped out; finding 3 is non-blocking follow-up.

pkg/recipe/version_pin_guard_test.go
mchmarny: Good catch — the recipe guard alone doesn't fully close #1424: a coordinated bump (registry default + every pin moved together) would pass it even with a stale committed BOM.

Added `TestCommittedBOMVersionsMatchRegistry` (`tools/bom/freshness_test.go`, 5c1b6759): it diffs the committed `docs/user/container-images.md` version column against the registry pins — no Helm rendering, runs under `make test` → `make qualify`. So a pin bump that forgets `make bom-docs` now fails CI deterministically. Both halves of #1424's acceptance now hold: recipes match the registry (recipe guard) **and** the committed BOM matches the registry (freshness test). Verified it fails on a stale version and passes when fresh.

Scope note: this gates the version column, not upstream *image drift* within a chart (a chart bumping its own image without a pin change) — that still needs opt-in `make bom-check`, now documented as such in `recipe.md`. Keeping `Fixes #1424`.

recipes/overlays/base.yaml
mchmarny: Strictly this is pre-existing drift outside this PR's scope — the `examples/recipes/*` files are hand-curated illustrative snapshots (README: "not a cross-version compatibility boundary"), and e.g. `gpu-operator v25.10.1` predates and is independent of this change — but fixed it here anyway.

Note on approach: full `aicr recipe` regeneration is **not** viable. The tool now inlines every component's `healthCheckAsserts`, so a regenerated `eks-training` is 1746 lines vs the curated 73 — that would destroy the examples' readability and change their nature. Instead I aligned each drifted pin to verified resolved output for that example's criteria (5c1b6759): `gpu-operator → v26.3.2`, `nvsentinel → v1.9.0`, and — per your point — AKS `kube-prometheus-stack → 83.7.0` (the resolved AKS value, not the registry default). All three files' pins now match resolved output.

pkg/recipe/version_pin_guard_test.go
mchmarny: Also strictly outside #1424's scope, but fixed here anyway. `TestOverlayVersionPinsMatchRegistry` now resolves every overlay and, per exemption, fails if the component's registry default is installed by no enabled recipe (5c1b6759) — machine-enforcing the sole-consumer policy rather than leaving it a comment. Verified it fails when the default is made un-findable, and passes for the current AKS `kube-prometheus-stack` exemption (default `84.4.0` is installed by EKS/base and every non-AKS recipe).

docs/contributor/recipe.md
```diff
@@ -473,11 +476,40 @@ when a chart bumps an image inside its own templates without a
...
+`TestCommittedBOMVersionsMatchRegistry` (`tools/bom/freshness_test.go`,
+run by `make test` → `make qualify`) diffs the committed doc's version
+column against the registry pins with no Helm rendering, so a chart-pin
+bump that forgets `make bom-docs` fails CI. It does **not** catch
+*upstream image drift* — a chart bumping an image inside its own
+templates without a pin change on our side. Full `make bom-check`
+verifies that too by re-rendering, but it is **opt-in only** — not
+wired into `make qualify`, `make lint`, or the PR gate. So you still
+must run `make bom-docs` after a values change; wiring `bom-check` into
+the gate remains a desirable follow-up.
```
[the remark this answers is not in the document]
mchmarny: Done — promoted to `### Version freshness is gated; image drift is not` (f881fd4c), matching the sibling `### Version pinning is single-source` heading. Dropped the italics/period per heading style and kept the paragraphs beneath unchanged.

tools/bom/freshness_test.go
```diff
@@ -0,0 +1,176 @@
...
+
+		name := strings.TrimSpace(cells[nameCol])
+		ver := strings.TrimSpace(cells[verCol])
+		if name == "" || ver == "" {
+			continue
+		}
+		out[name] = ver
+	}
+	return out
+}
```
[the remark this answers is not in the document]
mchmarny: Good catch on the latent edge case — hardened it (f881fd4c). Header detection now drives off `inTable` and keeps searching every pipe row until a single row resolves **both** columns, committing them atomically; a row carrying only one header no longer stops the search, so a preceding unrelated table can't lock us out. Verified the freshness test still passes and still fails on a stale version after the refactor.

examples/recipes/eks-gb200-ubuntu-training-with-validation.yaml
```diff
@@ -140,7 +140,7 @@ componentRefs:
     chart: nvsentinel
     type: Helm
     source: oci://ghcr.io/nvidia
-    version: v1.3.0
+    version: v1.9.0
```
yuanchen8911, line 142: One more stale value in this refreshed example, separate from the chart-version pins: further up (around lines 119-122) it still sets overrides.driver.version: 580.82.07, while the current base gpu-operator pin is 580.126.20 (recipes/components/gpu-operator/values.yaml:152) — the driver version the BOM advertises. Bundling this example therefore installs a driver version nothing else ships. This is a nested image override rather than a registry/chart pin, so the chart-version refresh in this PR didn't touch it.

Recommended: resolve the GB200 criteria against the current recipes and, if it produces 580.126.20, drop this override (or update it to match). Non-blocking, but cheap to fix while this file is already being refreshed.

yuanchen8911 (changes requested):
Version-management model and guard tests look correct — both new tests compile, lint clean, and pass, the version bumps are internally consistent with the registry defaults, and the AKS kube-prometheus-stack divergence is covered by a well-formed exemption. A few items to address before merge:

**examples/recipes/*.yaml are guarded by neither new test.** `TestOverlayVersionPinsMatchRegistry` iterates only base/overlays/mixins (`store.Base`, `store.Overlays`, `store.Mixins`), and `TestCommittedBOMVersionsMatchRegistry` checks only the registry against the committed doc. The persisted examples are updated by hand in this PR, so they are consistent now, but a future registry or overlay bump can leave the examples advertising a stale chart version while both guards stay green — the same drift class #1424 targets, on the examples surface. Please either extend the guard (or a regeneration/consistency check) to cover `examples/recipes/`, or file a tracked follow-up.

**assertExemptionDefaultsInstalled excludes base.yaml as a standalone consumer.** The sole-consumer check resolves only overlays with non-nil `Spec.Criteria`; `base.yaml` pins kube-prometheus-stack to the 84.4.0 default but has no criteria, so it is not in `results`. The AKS exemption stays satisfied only because a criteria-bearing non-AKS overlay inherits base's 84.4.0 default. This holds today (and the test fails loudly if no overlay carries criteria), but the guarantee depends on overlay structure. Worth a comment documenting the coupling.

**PR description omits two example bumps.** The Implementation Notes cover only the nvsentinel bump and the AKS exemption, but the diff also bumps the example gpu-operator (v25.10.1 -> v26.3.2) and AKS kube-prometheus-stack (84.4.0 -> 83.7.0). The changes are correct; please note them in the description so reviewers reading the summary know examples were touched.

examples/recipes/eks-gb200-ubuntu-training-with-validation.yaml
mchmarny: Good catch — dropped the stale `driver.version: 580.82.07` override (04cd0e4b). Confirmed no gb200 overlay in the resolution chain (`gb200-eks-ubuntu-training` → `gb200-eks-training` → `eks-training`) sets `driver.version`; it inherits the base gpu-operator values pin `580.126.20` (`recipes/components/gpu-operator/values.yaml:152`), which is what the BOM advertises. Rather than update the literal, I removed the override entirely to match resolved output — sibling examples (`eks-training.yaml`, `aks-training.yaml`) carry no `driver.version` override either; the version comes from the referenced valuesFile. Recipe + BOM tests still green, yamllint clean.

yuanchen8911 (approved):
Approving. The version-management model and both guard tests are correct — they compile, lint clean, pass, and version pins align with registry defaults; CI coverage is above the floor. The stale gb200 driver.version override is resolved (the example now inherits the chart default), and the review findings are addressed.

Two housekeeping items before merge:

1. Please rebase onto origin/main and squash before merging — the branch currently carries several merge-from-main commits plus a follow-up doc-consistency commit. A single squashed, rebased commit keeps history clean per repo policy (not the "Update branch"/merge path).

2. The example recipes (examples/recipes/*) bypass overlay resolution and are not covered by either new guard; they're kept consistent by hand in this PR. Let's track guarding them (or a regeneration check) in a follow-up issue so the drift class can't silently return there.

Non-blocking: the example chart-version bumps are now noted in the PR description, and the Common Pitfalls doc bullet was reconciled with the new version-freshness gating section.

PR #1575, opened 2026-07-01
fix(ci): make helm v4.2.2 pin effective, verify helm-diff provenance

### LINKED ISSUE #1574: UAT helm-diff install fails: v4.2.2 pin bypassed by preinstalled Helm 3 on ubuntu-latest
## Summary

UAT (`uat-aws`, `uat-gcp`) fails at the **Install helm-diff plugin (v3.15.10)** step:

```
Error: Unable to get repository: Cloning into '/home/runner/.cache/helm/plugins/https-github.com-databus23-helm-diff-releases-download-v3.15.10-helm-diff-linux-amd64.tgz'...
remote: Not Found
fatal: repository 'https://github.com/databus23/helm-diff/releases/download/v3.15.10/helm-diff-linux-amd64.tgz/' not found
: exit status 128
```

## Root cause (two compounding bugs)

**1. The pinned Helm 4.2.2 is never installed on `ubuntu-latest`.**
`.github/actions/setup-build-tools/action.yml` (Install Helm step) guards installation with:

```bash
if [[ -x /usr/local/bin/helm ]]; then echo "helm already installed"; helm version; exit 0; fi
```

GitHub's `ubuntu-24.04` image (the `ubuntu-latest` label) **preinstalls Helm 3.21.2** at exactly `/usr/local/bin/helm`. The guard short-circuits and the `.settings.yaml` pin `helm: 'v4.2.2'` becomes a **no-op** — every UAT run actually executes **Helm 3.21.2**. The guard checks only for the binary's *presence*, never its version.

**2. PR #1510 switched helm-diff to a tarball-URL install that Helm 3 routes to git.**
Before #1510 the command was the git-repo form (`helm plugin install https://github.com/databus23/helm-diff`), which Helm 3 installs by cloning the repo and running `install-binary.sh` — this is why UAT was green. #1510 changed `tests/uat/{aws,gcp}/run` to:

```bash
helm plugin install ".../releases/download/${HELM_DIFF_VERSION}/helm-diff-linux-amd64.tgz"
```

with a comment describing it as "the recommended Helm **4** method." But the runtime is Helm **3.21.2**, whose source detection routes the GitHub release-download URL (302 redirect, `application/octet-stream`) to the **VCS installer**, which `git clone`s the `.tgz` URL → `remote: Not Found`. That is the observed error.

Additionally (verified locally on Helm 4.2.0): even once bug #1 is fixed and Helm 4 is genuinely in use, the tarball-URL install fails provenance verification (`--verify` defaults to `true` in Helm 4) unless the maintainer's GPG key is imported into a legacy keyring and passed via `--keyring`. Both must be addressed for the Helm 4 path to work end-to-end.

## Fix (chosen approach: make the v4.2.2 pin real)

1. `.github/actions/setup-build-tools/action.yml`: version-check the preinstalled `helm` and reinstall the pinned version when it does not match (do not skip on mere presence). This makes the whole repo actually run the pinned Helm 4 (shared action — validate ripple across dependent workflows).
2. `tests/uat/{aws,gcp}/run`: import the helm-diff maintainer GPG key into a legacy keyring and pass `--keyring` on the tarball install so Helm 4 provenance verification passes.

## Impact

- UAT-AWS and UAT-GCP are fully blocked at deploy.
- Latent: the `helm: v4.2.2` pin is currently ineffective repo-wide anywhere `setup-build-tools` runs on `ubuntu-latest`; all such jobs silently use the runner's preinstalled Helm 3.21.2.

njhensley:
## Summary

Make the pinned Helm `v4.2.2` actually install on GitHub-hosted runners, and verify helm-diff plugin provenance so the UAT install step (`uat-aws`, `uat-gcp`) stops failing with `remote: Not Found`.

## Motivation / Context

UAT failed at **Install helm-diff plugin (v3.15.10)** with a `git clone` of the `.tgz` URL → `remote: Not Found`. Two compounding bugs:

1. **The `helm: v4.2.2` pin was a no-op.** `setup-build-tools` skipped Helm install whenever `/usr/local/bin/helm` merely existed. `ubuntu-latest` (ubuntu-24.04) **preinstalls Helm 3.21.2** at that path, so every job silently ran Helm 3.
2. **#1510 switched helm-diff to a tarball-URL install that only works under Helm 4.** Under the actually-running Helm 3, that GitHub release-download URL is routed to the git VCS installer and `git clone` fails. The prior git-repo install form worked on Helm 3, which is why UAT was green before #1510.

Even after fixing (1), Helm 4 verifies plugin provenance by default (`--verify=true`) and the install fails unless the maintainer's signing key is available in a keyring — so (2)'s install path also needed provenance wiring.

Fixes: #1574
Related: #1510

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: CI (`.github/actions/setup-build-tools`, `tests/uat/{aws,gcp}/run`)

## Implementation Notes

- **`setup-build-tools`**: reinstall Helm unless the existing binary's version already matches the pin (`helm version --template '{{.Version}}'`), instead of skipping on mere presence. This makes the `.settings.yaml` pin effective everywhere the action runs.
- **`tests/uat/{aws,gcp}/run`**: import the helm-diff maintainer's release key into an isolated `GNUPGHOME`, **pin it to the published fingerprint `C5645EF47482257A1F806D2BEA17A2A206AFF8CD` and fail closed on mismatch**, export it to a legacy keyring, and run `helm plugin install … --keyring …` so Helm 4 provenance verification passes. Temp keyring/GNUPGHOME are cleaned up in all paths.

**Blast radius (intentional):** making the pin real flips these jobs from the runner's Helm 3.21.2 to the pinned Helm 4.2.2: `gpu-cluster-setup` (gpu-h100 tests), `integration`, `kwok-test`, `mirror-e2e`, `sigstore-scaffolding-e2e`, `uat-aws`, `uat-gcp`. helmfile v1.6.0 supports Helm 4.

## Testing

Verified the full plugin-install flow end-to-end against a local Helm 4.2.0 (close to the pinned 4.2.2): fingerprint pin check passes, provenance verifies (`Plugin Hash Verified …`), `helm diff version` → `3.15.10`. `bash -n` and `yamllint` clean on the edited files. actionlint (merge-gate) lints the action YAML; no new `${{ }}` expressions were introduced.

```bash
bash -n tests/uat/aws/run tests/uat/gcp/run   # syntax OK
yamllint .github/actions/setup-build-tools/action.yml   # clean
```

## Risk Assessment

- [x] **Medium** — Touches a shared composite action; several CI jobs move from Helm 3 to the pinned Helm 4.

**Rollout notes:** No user-facing/runtime behavior change (CI-only). Easily reverted. Watch the first run of the affected workflows for any latent Helm 3→4 incompatibilities.

## Checklist

- [ ] Tests pass locally (`make test` with `-race`) — N/A, CI-only shell/YAML change
- [x] Linter passes (`bash -n`, `yamllint`)
- [ ] I added/updated tests for new functionality — N/A
- [ ] I updated docs if user-facing behavior changed — N/A (internal CI)

### DIFF
--- a/tests/uat/aws/run
+++ b/tests/uat/aws/run
@@ -167,9 +167,26 @@ phase_install() {
       helm plugin remove diff
     fi
     # Install from the signed release tarball (helm-diff's recommended Helm 4
-    # method). The git-URL form requires --verify=false under Helm 4; the
-    # tarball form works without it and carries the upstream release signature.
-    helm plugin install "https://github.com/databus23/helm-diff/releases/download/${HELM_DIFF_VERSION}/helm-diff-linux-amd64.tgz"
+    # method). Helm 4 verifies plugin provenance by default (--verify=true):
+    # the .prov signature is checked against a keyring. Import the maintainer's
+    # release key, pin it to the published fingerprint (fail closed on
+    # mismatch), export it to a legacy keyring, and verify against it.
+    local HELM_DIFF_KEY_FPR="C5645EF47482257A1F806D2BEA17A2A206AFF8CD"
+    local HELM_DIFF_GNUPGHOME HELM_DIFF_KEYRING
+    HELM_DIFF_GNUPGHOME="$(mktemp -d)"
+    HELM_DIFF_KEYRING="$(mktemp)"
+    GNUPGHOME="${HELM_DIFF_GNUPGHOME}" curl -fsSL "https://github.com/databus23.gpg" \
+      | GNUPGHOME="${HELM_DIFF_GNUPGHOME}" gpg --import
+    if ! GNUPGHOME="${HELM_DIFF_GNUPGHOME}" gpg --with-colons --fingerprint \
+        | awk -F: '/^fpr:/{print $10}' | grep -qx "${HELM_DIFF_KEY_FPR}"; then
+      echo "::error::helm-diff release key fingerprint mismatch (expected ${HELM_DIFF_KEY_FPR})" >&2
+      rm -rf "${HELM_DIFF_GNUPGHOME}" "${HELM_DIFF_KEYRING}"
+      exit 1
+    fi
+    GNUPGHOME="${HELM_DIFF_GNUPGHOME}" gpg --export "${HELM_DIFF_KEY_FPR}" > "${HELM_DIFF_KEYRING}"
+    helm plugin install "https://github.com/databus23/helm-diff/releases/download/${HELM_DIFF_VERSION}/helm-diff-linux-amd64.tgz" \
+      --keyring "${HELM_DIFF_KEYRING}"
+    rm -rf "${HELM_DIFF_GNUPGHOME}" "${HELM_DIFF_KEYRING}"
   fi
   echo "::endgroup::"
 
--- a/.github/actions/setup-build-tools/action.yml
+++ b/.github/actions/setup-build-tools/action.yml
@@ -244,9 +244,23 @@ runs:
       shell: bash
       run: |
         set -euo pipefail
-        if [[ -x /usr/local/bin/helm ]]; then echo "helm already installed"; helm version; exit 0; fi
         HELM_VERSION="${{ inputs.helm_version }}"
         HELM_VERSION="${HELM_VERSION#v}"
+        # The runner image may ship a preinstalled helm (ubuntu-latest bundles
+        # Helm 3 at /usr/local/bin/helm). Only skip installation when the
+        # existing binary already matches the pinned version; a bare presence
+        # check silently ignores the .settings.yaml pin and leaves jobs running
+        # the runner's stock Helm 3 instead of the pinned Helm 4.
+        if [[ -x /usr/local/bin/helm ]]; then
+          INSTALLED="$(/usr/local/bin/helm version --template '{{.Version}}' 2>/dev/null || true)"
+          INSTALLED="${INSTALLED#v}"
+          if [[ "${INSTALLED}" == "${HELM_VERSION}" ]]; then
+            echo "helm ${HELM_VERSION} already installed"
+            helm version
+            exit 0
+          fi
+          echo "helm ${INSTALLED:-unknown} present; replacing with pinned ${HELM_VERSION}"
+        fi
         HELM_BASE="https://get.helm.sh"
         HELM_TAR="helm-v${HELM_VERSION}-linux-amd64.tar.gz"
         HELM_TMP="$(mktemp -d)"
--- a/tests/uat/gcp/run
+++ b/tests/uat/gcp/run
@@ -167,9 +167,26 @@ phase_install() {
       helm plugin remove diff
     fi
     # Install from the signed release tarball (helm-diff's recommended Helm 4
-    # method). The git-URL form requires --verify=false under Helm 4; the
-    # tarball form works without it and carries the upstream release signature.
-    helm plugin install "https://github.com/databus23/helm-diff/releases/download/${HELM_DIFF_VERSION}/helm-diff-linux-amd64.tgz"
+    # method). Helm 4 verifies plugin provenance by default (--verify=true):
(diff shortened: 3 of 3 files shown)


tests/uat/aws/run
```diff
@@ -167,9 +167,26 @@ phase_install() {
...
+    if ! GNUPGHOME="${HELM_DIFF_GNUPGHOME}" gpg --with-colons --fingerprint \
+        | awk -F: '/^fpr:/{print $10}' | grep -qx "${HELM_DIFF_KEY_FPR}"; then
+      echo "::error::helm-diff release key fingerprint mismatch (expected ${HELM_DIFF_KEY_FPR})" >&2
+      rm -rf "${HELM_DIFF_GNUPGHOME}" "${HELM_DIFF_KEYRING}"
+      exit 1
+    fi
+    GNUPGHOME="${HELM_DIFF_GNUPGHOME}" gpg --export "${HELM_DIFF_KEY_FPR}" > "${HELM_DIFF_KEYRING}"
+    helm plugin install "https://github.com/databus23/helm-diff/releases/download/${HELM_DIFF_VERSION}/helm-diff-linux-amd64.tgz" \
+      --keyring "${HELM_DIFF_KEYRING}"
+    rm -rf "${HELM_DIFF_GNUPGHOME}" "${HELM_DIFF_KEYRING}"
```
mchmarny, line 189: Cleanup isn't guaranteed on all paths: under `set -e` (L41), if `helm plugin install` (L187) fails, the script aborts before this `rm -rf`, leaking the temp `GNUPGHOME` + keyring — so the PR's "cleaned up in all paths" doesn't hold for the install-failure path. Harmless on ephemeral runners, but a `trap 'rm -rf "$HELM_DIFF_GNUPGHOME" "$HELM_DIFF_KEYRING"' RETURN` right after the `mktemp`s would make it truly all-paths and drop both explicit `rm` lines. Same block in `tests/uat/gcp/run`. Non-blocking.

tests/uat/aws/run
```diff
@@ -167,9 +167,26 @@ phase_install() {
...
-    helm plugin install "https://github.com/databus23/helm-diff/releases/download/${HELM_DIFF_VERSION}/helm-diff-linux-amd64.tgz"
+    # method). Helm 4 verifies plugin provenance by default (--verify=true):
+    # the .prov signature is checked against a keyring. Import the maintainer's
+    # release key, pin it to the published fingerprint (fail closed on
+    # mismatch), export it to a legacy keyring, and verify against it.
+    local HELM_DIFF_KEY_FPR="C5645EF47482257A1F806D2BEA17A2A206AFF8CD"
+    local HELM_DIFF_GNUPGHOME HELM_DIFF_KEYRING
+    HELM_DIFF_GNUPGHOME="$(mktemp -d)"
+    HELM_DIFF_KEYRING="$(mktemp)"
+    GNUPGHOME="${HELM_DIFF_GNUPGHOME}" curl -fsSL "https://github.com/databus23.gpg" \
```
mchmarny, line 178: nit: the `GNUPGHOME=…` prefix on `curl` is a no-op — curl doesn't read it; only the `gpg --import` on the right of the pipe needs it. Harmless, just slightly misleading. Same in `gcp/run`.

mchmarny (approved):
Correct two-part fix: the version-match guard makes the `.settings.yaml` Helm pin actually take effect (the bare-presence check silently left jobs on the runner's stock Helm 3), and the helm-diff provenance wiring is soundly fail-closed — pinning to a hardcoded fingerprint and exporting only that key to the verification keyring means a swapped/compromised key source can't smuggle a plugin through. CI green (24/24).

Non-blocking nits inline (cleanup not guaranteed on install failure under `set -e`; a no-op env on `curl`), plus the provenance block is duplicated verbatim across `aws/run` and `gcp/run` — fine for now, just a drift point when the pinned fingerprint rotates. Nothing blocks merge.

PR #1582, opened 2026-07-01
chore(license): header hand-authored files, skip generated evidence pointers

yuanchen8911:
## Summary

Stop `make license` from repeatedly re-injecting Apache headers into header-less files: add headers to the two hand-authored files, and exclude the generated evidence pointers from `addlicense` so they stop churning.

## Motivation / Context

`make lint` runs `make license` (addlicense in add-mode), which kept adding headers to files that lack them. Two are hand-authored source that should carry a header permanently; the evidence pointer YAMLs are written header-less by `MarshalPointer` (`serializer.MarshalYAMLDeterministic`), so any committed header is stripped on the next `aicr evidence publish` / signing step — churn that never converges.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: build tooling (`Makefile`), license headers

## Implementation Notes

- Add the standard Apache header to:
  - `pkg/corroborate/renderer/index.html` — a `//go:embed`-ed static dashboard asset (hand-authored).
  - `recipes/evidence/allowlist.yaml` — the hand-maintained evidence trust root (no generator writes it).
- Ignore exactly the two generated pointer shapes in `LICENSE_IGNORES` so `make license` no longer targets them:
  - `recipes/evidence/*.yaml` — the transient flat `<recipe>.yaml` pending pointer (the `--no-sign` / commit-flat state CI later signs and relocates).
  - `recipes/evidence/*/*/*.yaml` — the final `<recipe>/<source>/sha256-*.yaml` pointers.
  The patterns are deliberately narrow (not a recursive `**`) so any unrelated YAML added under `recipes/evidence/` still gets header enforcement.
- The flat glob also covers `allowlist.yaml`, so `make license` re-headers it with an explicit second `addlicense` pass.
- The `DO NOT ADD IGNORES` note in the Makefile governs `go-licenses` `--allowed_licenses`/`--ignore` in `license-check` (dependency licensing), not `addlicense`'s file-glob ignores.

## Testing

```bash
# make license is now idempotent and correctly scoped:
#  - flat pending pointer + nested sha256 pointer: stay header-less
#  - an unrelated YAML placed under recipes/evidence/: DOES get a header (enforcement retained)
#  - allowlist.yaml: re-headered by the explicit pass
make license
go run ./tools/evidence-pointercheck -root recipes/evidence   # OK — all pointers honor the contract
yamllint recipes/evidence/allowlist.yaml                       # clean
```

Infra/tooling-only change (Makefile ignore globs + license-comment headers); no Go logic touched, so tests/e2e/Go-lint cannot regress. The evidence verifier `yaml.Unmarshal`s pointer bodies (comments ignored) and the pointer filename digest is the bundle digest (a field value), not the file's own bytes — so headers on the YAML/HTML files are parse- and verification-safe.

## Risk Assessment

- [x] **Low** — Comment-only header additions plus two narrow build-ignore globs; trivially reversible.

**Rollout notes:** N/A

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -174,13 +174,23 @@ LICENSE_IGNORES = \
 	-ignore 'dist/**' \
 	-ignore 'vendor/**' \
 	-ignore '**/testdata/**' \
+	-ignore 'recipes/evidence/*.yaml' \
+	-ignore 'recipes/evidence/*/*/*.yaml' \
 	-ignore 'THIRD_PARTY_NOTICES.md' \
 	-ignore '.licenses-cache/**'
 
+# The two recipes/evidence patterns in LICENSE_IGNORES match exactly the
+# generated, header-less pointer shapes MarshalPointer emits — the transient
+# flat <recipe>.yaml pending state and the final <recipe>/<source>/sha256-*.yaml
+# (any committed header is stripped on the next publish/sign). They are kept
+# narrow (not a recursive **) so unrelated YAML added under the tree still gets
+# header enforcement. The flat glob also covers the hand-maintained
+# allowlist.yaml, so the license target headers it explicitly below.
 .PHONY: license
 license: ## Add/verify license headers in source files
 	@echo "Ensuring license headers..."
 	@addlicense -f .github/headers/LICENSE $(LICENSE_IGNORES) .
+	@addlicense -f .github/headers/LICENSE recipes/evidence/allowlist.yaml
 
 #### DO NOT CHANGE THIS SET OF ALLOWED LICENSES, DO NOT ADD IGNORES
 license-check: ## Check license is approved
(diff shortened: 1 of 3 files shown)


mchmarny (approved):
LGTM. Clean, well-scoped tooling fix that makes `make license` idempotent.

Verified against the repo:
- **Glob depth is correct** — pointers live at `recipes/evidence/<recipe>/<hash>/sha256-*.yaml`, matched exactly by `recipes/evidence/*/*/*.yaml`; the flat `recipes/evidence/*.yaml` covers the pending pointer + `allowlist.yaml`. Single `*` (not recursive `**`) keeps enforcement on unrelated nested YAML, as described.
- **Idempotency holds** — `allowlist.yaml` is ignored in the main pass then re-headered by the explicit second pass; addlicense is idempotent so repeat runs don't duplicate. Pending pointers stay header-less.
- **HTML header is embed/render-safe** — comment after `<!DOCTYPE html>`, before `<html>`.

The ignore-then-re-add two-pass on `allowlist.yaml` is a minor wart, but it's unavoidable (the pending-pointer glob can't distinguish it by shape) and already documented in the Makefile comment. No blockers — rebase for the `behind` state and merge.

PR #1594, opened 2026-07-02
feat(tuning): auto-generate nodewright tuning-status table

### LINKED ISSUE #1563: Document nodewright tuning status per service in docs/integrator/components/nodewright.md
## Summary

Add a "Tuning status" section to `docs/integrator/components/nodewright.md` that surfaces, per service, the state of nodewright/tuned tuning so integrators and users can see at a glance which (service, accelerator, intent) combinations are tuned and with what package version.

The status table **must be generated automatically from the existing recipes** (overlays/mixins + nodewright package pins) — not hand-maintained. This keeps it in sync as recipes change, similar in spirit to the BOM doc regeneration (`make bom-docs`).

## Motivation / Context

From a discussion in #aicr: nodewright tuning status is currently implicit (spread across recipes and the nodewright-packages repo) and not easy to discover from the AICR docs. The repo already has `docs/integrator/components/nodewright.md`, which is the natural home for this.

Proposed table, derived from recipe criteria + nodewright package pins:

| Service | Accelerator | Intent | Package version | Status |
|---------|-------------|--------|-----------------|--------|

## Type of Change

- [x] Documentation

## Components Affected

- nodewright / nodewright-customizations docs
- recipes (the source of truth the table is generated from)

## Implementation Notes

- Add a generator that walks the recipes (criteria: service, accelerator, intent) and the nodewright `nvidia-tuned` package pins, and emits the table into the doc.
- Wire it like the existing BOM generation so the doc is regenerated rather than edited by hand.

## Open Questions

- Where exactly within the doc the section lives (new "Tuning status" section vs. promoting existing per-service detail into the generated table).

## Notes

Related context: nodewright `nvidia-tuned` package config takes `intent`, `accelerator`, `service`; supported accelerators currently `h100`, `gb200`. The ALLOW_NEWER kernel/EFA work in NVIDIA/nodewright-packages#53 is related but separate.

ayuskauskas:
## Summary

Auto-generate a per-(service, accelerator) nodewright **tuning-status** table into `docs/integrator/components/nodewright.md`, rendered hermetically from the recipe catalog and the nodewright package pins — regenerated by `make tuning-docs`, never hand-maintained (like the BOM / recipe-health docs).

## Motivation / Context

nodewright tuning status was implicit (spread across recipes and the nodewright-packages repo) and not discoverable from the AICR docs. This surfaces it in its natural home, generated from the source of truth so it stays in sync as recipes change.

Fixes: #1563
Related: N/A

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: new `pkg/tuning`, `tools/tuning`, `tools/internal/docgen`; `pkg/health` refactor; `Makefile`; `.github/CODEOWNERS`

## Implementation Notes

Mirrors the hermetic `recipe-health` doc pipeline (`pkg/health` + `tools/health` + `make recipe-health-docs`).

- **`pkg/recipe.ResolveLeaves`** — extracted the "enumerate + resolve every leaf recipe" loop out of `health.Compute` into a shared helper; `pkg/health` now consumes it. The refactor is **behavior-preserving** (generated `recipe-health.md` is byte-identical).
- **`pkg/tuning`** — resolves every leaf, finds the `nodewright-customizations` component, reads its selected manifest, and extracts the `nvidia-setup` / `nvidia-tuned` / `nvidia-tuning-gke` version pins **by image basename** (not the Skyhook packages-map key), skipping nested `dependsOn:` blocks. Rows are keyed by `(service, accelerator)` with a **fail-loud consistency check** (divergence across a group errors rather than silently picking one). Rendering: empty criteria dimension → `*`; `Profile` shows the tuning-profile accelerator when it differs from the selected one, `-` when identical / when no package is applied. Columns are left-justified for readable raw `.md`.
- **`tools/tuning` + `tools/internal/docgen`** — thin generator + a shared `WriteRendered` helper (also adopted by `tools/health`).
- **`make tuning-docs` / `make tuning-check`** — regenerate + staleness check. `tuning-check` is wired into `make qualify` **and** enforced in CI by a path-filtered `tuning-freshness` merge-gate job (mirroring `bom-freshness`). This is a **deliberate divergence** from the advisory `recipe-health-check` precedent: `tuning-check` is hermetic and narrow (a small table keyed by service × accelerator), so fail-closed enforcement keeps the committed table from drifting when a pin or recipe changes. MDX-comment splice markers (`{/* … */}`), since the doc is Fern-published.

**Divergence from the #1563 proposal (flagging for the issue author):** the proposed columns were `Service | Accelerator | Intent | Package version | Status`. This implementation drops **Intent** (pins do not vary by intent — enforced by the consistency check), drops **Status** (replaced by the self-describing `Setup`/`Tuning` package-version columns), and adds **Profile**.

## Testing

```bash
make qualify
```

- `go test -race` on affected packages (`pkg/tuning`, `pkg/recipe`, `pkg/health`, `tools/...`): **pass**.
- `make test-coverage`: **77.7% ≥ 75%** floor (`pkg/tuning` 88.6%; new exported symbols covered).
- `golangci-lint` on all changed paths: **0 issues**. `make check-docs-mdx`: **OK**. `make tuning-check`: **up to date**.
- `make qualify`: test / vet / golangci-lint / license / yamllint **pass**. The `e2e` phase could not complete in the local environment (the pinned `helmfile` binary — a `make tools-setup` dependency — is absent from PATH; the failing step `cli-bundle-helmfile` is unrelated to this change, which touches no bundling/helmfile/recipe-manifest code). CI runs the full e2e + scan.

## Risk Assessment

**Rollout notes:** N/A. Additive (new packages + a generated doc region + make targets + a path-filtered `tuning-freshness` merge-gate check); the single refactor (`ResolveLeaves`) is behavior-preserving; no runtime, API, or deployment path is touched.

## Checklist


ayuskauskas:
Thanks for the review. Addressed in `64d37881`:

- **Vague "readme" reference** (`nodewright.md`) — linked the nvidia-setup README explicitly.
- **MD058 (blank line before table)** — fixed at the source: `renderTable` now emits a leading blank line before the header, so the spliced body is always separated from the marker; the doc was regenerated.
- **Canceled-context edge case** in `ResolveLeaves` — added a pre-loop `ctx.Err()` check so a canceled context with a zero-match filter errors instead of returning `(nil, nil)`, plus a regression test (`TestResolveLeaves_CanceledContextZeroMatchFilter`).
- **Redundant `order` slice** in `Compute` — removed; `rows` is sorted regardless, so `groups` is iterated directly.
- **Duplicate not-applicable marker** — exported `tuning.NotApplicable` and switched the renderer to it, removing the duplicated literal.

Deliberately **not** changed:

- **Shared Make splice helper** (`tuning-docs`/`tuning-check` duplicating `bom-docs`/`recipe-health-docs`) — the duplication is the established pattern across all three doc generators, `checkmake` isn't part of the merge gate, and extracting a shared `$(call splice-doc,…)` helper would touch the out-of-scope `bom`/`recipe-health` targets. Better as its own focused cleanup PR than bundled here.
- The `MD037` markdownlint warnings on the legend paragraph don't reproduce against the current text (no spaces inside emphasis markers), and markdownlint isn't gated in this repo.

ayuskauskas:
@coderabbitai review

pkg/tuning/pins.go
```diff
@@ -0,0 +1,166 @@
...
+		break
+	}
+	if entryIndent < 0 {
+		return nil, errors.New(errors.ErrCodeInternal, "packages: block has no entries")
+	}
+	// Assumes a 2-space YAML indent step (true for all current nodewright
+	// manifests). A violation is caught fail-loud: a recognized image whose
+	// version: line is missed at this offset errors in flush rather than
+	// silently blanking the pin.
+	fieldIndent := entryIndent + 2
```
mchmarny: The fail-loud backstop this comment claims doesn't cover the dangerous case. With a 4-space indent step, both `image:` and `version:` lines sit at `entryIndent+4`, so neither matches `fieldIndent` — `curImage` stays empty, `flush` is a no-op, and the manifest yields zero pins: the row silently renders all `-`. That's exactly the silent-blanking this invariant exists to prevent (the flush error only fires when `image:` *is* parsed and `version:` is missed).

Derive `fieldIndent` from the first observed field line instead of hardcoding `+2` — e.g. the first non-template, non-comment line with indent > `entryIndent` after an entry key — and keep the flush errors as-is. That removes the magic `+2` while staying fail-loud on genuinely malformed blocks.

pkg/tuning/pins.go
```diff
@@ -0,0 +1,166 @@
...
+// indentOf returns the count of leading spaces on a line.
+func indentOf(line string) int {
+	return len(line) - len(strings.TrimLeft(line, " "))
+}
+
+// classifyPins maps extracted image→version pins into the Setup and Tuning
+// columns. Setup is the nvidia-setup pin; Tuning is nvidia-tuned or
+// nvidia-tuning-gke (a manifest ships at most one). Unrecognized images yield
+// zero-value pins (rendered "-").
+func classifyPins(pins map[string]string) (setup, tuning PackagePin) {
```
mchmarny: nit: "a manifest ships at most one" is asserted but not enforced — if a manifest ever carries both `nvidia-tuned` and `nvidia-tuning-gke`, the gke pin silently wins. An error on both-present would match the fail-loud ethos of the rest of this package (conflicting versions, missing version, group divergence all error).

docs/integrator/components/nodewright.md
```diff
@@ -57,23 +57,38 @@ Uses a set of bash scripts to do the necessary actions to bring an ubuntu worker
...
+{/* BEGIN AICR-TUNING */}
+
+| Service | Accelerator  | Profile | Setup              | Tuning                  |
+|---------|--------------|---------|--------------------|-------------------------|
+| bcm     | *            | h100    | nvidia-setup 0.3.0 | -                       |
+| bcm     | h100         | -       | nvidia-setup 0.3.0 | -                       |
+| eks     | a100         | h100    | nvidia-setup 0.2.2 | nvidia-tuned 0.3.0      |
+| eks     | gb200        | -       | nvidia-setup 0.2.2 | nvidia-tuned 0.3.0      |
+| eks     | h100         | -       | nvidia-setup 0.2.2 | nvidia-tuned 0.3.0      |
+| eks     | h200         | h100    | nvidia-setup 0.2.2 | nvidia-tuned 0.3.0      |
```
mchmarny: The committed table is stale against this branch's own manifests — reproduced locally: `make tuning-docs` rewrites these four eks rows from `nvidia-setup 0.2.2` to `0.4.0` (the pin was bumped after the table was generated). This is what's failing the `tuning-freshness` job on this PR. Run `make tuning-docs` and commit the refresh — nice proof the gate works, at least.

mchmarny (changes requested):
Solid work overall — the `ResolveLeaves` extraction is clean (with the sort-order preservation explicitly reasoned), and the structure test, BOM cross-check, and orphan-manifest test give this table real gating teeth. Three things before merge:

1. The committed tuning table is stale against this branch's own manifests (`nvidia-setup 0.2.2` vs `0.4.0`) — that's the failing `tuning-freshness` job. `make tuning-docs` + commit fixes it (inline comment on the rows).
2. One fail-open gap in the pin extractor's indent assumption (inline comment on `pins.go`), plus a small fail-loud nit in `classifyPins`.
3. The PR description says tuning-check is "NOT wired into `qualify`/lint/merge gate (matches `recipe-health-check`)" — but the diff wires it into `make qualify` *and* adds the `tuning-freshness` merge-gate job. The implementation is the better call (fail-closed, mirrors `bom-freshness`), so please update the description to match and note the deliberate divergence from the `recipe-health-check` precedent.

The other CI failures (`Tier 1: oke-ol-inference (argocd-git)` and the cascading `KWOK Test Summary`) look like an infra flake unrelated to this docs/tooling change — worth a rerun after the regen lands. Nothing else blocks; #1 and the description fix are the must-dos.

ayuskauskas:
@mchmarny thanks for the careful review — all three addressed in `2e8d1290`:

1. **Stale table** — regenerated for the `nvidia-setup 0.4.0` EKS bump (#1599). The four eks rows now show `0.4.0`; bcm stays `0.3.0`. `tuning-freshness` passes locally now. (And yes — nice confirmation the gate does its job.)

2. **pins.go fail-open** — correct, the hard-coded `+2` was a genuine silent-blank on a wider indent step (neither `image:` nor `version:` matched, so the entry produced no pins and rendered all `-`, with no flush error). `extractPackagePins` now derives the field indent from the first field line under each entry key instead of assuming `+2`, and still ignores the deeper `dependsOn`/`configMap` lines. Added a 4-space-step regression test that also checks the `dependsOn` version isn't leaked. Separately, `classifyPins` now **errors** when a manifest carries both `nvidia-tuned` and `nvidia-tuning-gke` rather than letting the gke pin silently win — matching the fail-loud ethos of the rest of the package.

3. **PR description** — updated. It now states that `tuning-check` is wired into `make qualify` **and** enforced in CI via the path-filtered `tuning-freshness` merge-gate job, and calls out the deliberate divergence from the advisory `recipe-health-check` precedent (hermetic + narrow → fail-closed is the better call, mirroring `bom-freshness`).

Agreed the `oke-ol-inference (argocd-git)` / `KWOK Test Summary` failures look like unrelated infra flake — will re-run once this lands.

mchmarny (approved):

PR #1598, opened 2026-07-03
fix(validator): poll GPU readiness signals to ride through reboots

njhensley:
## Summary

The deployment-phase `expected-resources` check verified Skyhook (nodewright) tuning and the DRA kubelet-plugin DaemonSet with **one-shot** Gets, so a single sample landing in a GPU-node reboot window failed the whole phase even though the cluster converged moments later. Both Go readiness signals now **poll with a stability dwell**, and the UAT readiness gate requires consecutive passes before admitting the conformance phase.

## Motivation / Context

Observed in a live UAT `aicr validate` run: the `expected-resources` check failed with `Nodewright tuning: status=in_progress (want complete)` even though the component's chainsaw health-check had passed ~2 minutes earlier, and `check-nvidia-smi` reported *no GPU nodes* at that instant — i.e. a node mid-reboot.

Root cause: Skyhook tuning packages carry `interrupt: reboot`, so the GPU node reboots more than once and re-opens `status=in_progress` after each reboot and for each newly-joined node; the DRA kubelet-plugin DaemonSet likewise churns to `0/0` while no GPU node is schedulable. Both signals are **non-monotonic** during rollout, so a one-shot Get is racy. The pre-validate readiness gate only proved "converged at instant T", letting a not-yet-settled cluster into `--phase all`.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `validators/deployment` (Go-resident deployment checks), `pkg/defaults`, `tests/uat/{aws,gcp}/run`

## Implementation Notes

- New shared `pollUntilStable(ctx, label, probe, onStable)` helper: polls any single-pass readiness probe until it reports healthy **continuously for `GPUReadinessStabilityWindow` (60s)**, bounded by `GPUReadinessTimeout` (8m) and capped by the parent check budget so the subsequent chainsaw asserts still fit the phase. On timeout it surfaces the last observed state (e.g. `status=in_progress`, `0/0 pods ready`) so gate logs still show *why*.
- `verifyNodewrightReady` and `verifyDRAKubeletPluginReady` refactored to single-pass probes wrapped by the helper. This is the Go analogue of the components' chainsaw health-check siblings (retry-until-timeout), hardened with a dwell so a check cannot pass on a momentary lull between reboots. No change to deployer-neutrality (no Helm/release-metadata/label dependence). The Skyhook CRD-not-registered skip (#607) is unchanged.
- New `pkg/defaults` constants `GPUReadinessPollInterval` / `GPUReadinessStabilityWindow` / `GPUReadinessTimeout`.
- UAT gate (`tests/uat/{aws,gcp}/run`): the readiness gate now requires the deployment phase to pass `READINESS_CONSECUTIVE_PASSES` (default 2) times in a row — any failure resets the streak — so a lone pass in a reboot lull no longer admits the conformance phase. Budget raised 20m → 30m for headroom.

## Testing

```bash
make qualify
```

Passed: unit tests (`-race`), golangci-lint + yamllint, e2e chainsaw (23/23), Grype scan (no vulnerabilities), license headers. New tests exercise both ride-through (transient `in_progress`/`0/0` → healthy ⇒ pass) and fail-closed timeout (persistent unready ⇒ error surfacing last-observed state) for nodewright and DRA. Poll tunables are shrunk once in `TestMain` so the tests run in milliseconds and stay race-free under `t.Parallel`.

## Risk Assessment

- [x] **Low** — Isolated to two Go deployment checks (now stricter *and* more tolerant of transients) and the UAT gate script; well-tested and easy to revert. The poll respects the parent context budget, so it cannot exceed the phase timeout.

**Rollout notes:** N/A — no config/flag/API surface change. Poll behavior is tunable via `pkg/defaults` constants; UAT gate via `READINESS_CONSECUTIVE_PASSES` / `READINESS_TIMEOUT_SECONDS` env overrides.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — no user-facing surface)


mchmarny (dismissed):

validators/deployment/expected_resources.go
```diff
@@ -485,38 +644,57 @@ func verifyDRAKubeletPluginReady(ctx *validators.Context, namespace string) erro
...
+func draKubeletPluginProbe(ctx *validators.Context, namespace string) (string, error) {
+	matches, seenNames, err := listDRAKubeletPluginDaemonSets(ctx, namespace)
+	if err != nil {
+		return "", err
+	}
 
 	switch len(matches) {
 	case 0:
-		return errors.New(errors.ErrCodeNotFound,
+		return "", errors.New(errors.ErrCodeNotFound,
```
mchmarny, line 684: nit/question: zero-match stays pollable, but a rebooting node churns the DaemonSet's *pods* (0/0), not the DaemonSet object itself — the object persists through the reboot. Chart-not-installed is therefore deterministic like the ambiguous case, and now takes the full 8m `GPUReadinessTimeout` to fail instead of instantly. The only transient this protects is the brief post-helmfile creation window, which the gate's outer retry already covers. Consider fail-fast on zero-match in the upfront gate too; fine to keep as-is if you'd rather stay conservative on the creation race.

pkg/defaults/timeouts.go
```diff
@@ -335,6 +335,42 @@ const (
...
+// (retry-until-timeout, 5m) already do, hardened with a continuous-pass window
+// so a check cannot pass on a momentary lull between reboots.
+const (
+	// GPUReadinessPollInterval is the sleep between readiness-signal samples.
+	GPUReadinessPollInterval = 10 * time.Second
+
+	// GPUReadinessStabilityWindow is how long a signal must report healthy
+	// continuously before the check passes, absorbing the flaps a reboot
+	// introduces.
+	GPUReadinessStabilityWindow = 60 * time.Second
```
mchmarny, line 363: nit: worth noting the healthy-path cost — every successful `--phase deployment` run now pays the 60s dwell twice (nodewright + DRA run sequentially), so ~2min per validate pass. Across 2 consecutive gate passes plus the conformance `--phase all` run that's ~6min of steady-state added per UAT run. The budget raises cover it, so just confirming this is intentional; if it ever pinches, a shorter dwell on an already-healthy first sample would be the lever.

validators/deployment/expected_resources.go
```diff
@@ -59,8 +63,85 @@ var (
...
+			stableSince = time.Time{}
+		}
+
+		if !time.Now().Before(deadline) {
+			break
+		}
+		select {
+		case <-ctx.Ctx.Done():
+			// Parent budget exhausted mid-poll.
+			return timedOut(lastErr)
```
mchmarny, line 145: nit: when the *parent* ctx expires mid-dwell with `lastErr == nil`, `timedOut(nil)` reports "did not hold it for the 60s stability window within 8m" even though the parent budget — not `gpuReadinessTimeout` — was the binding constraint. Slightly misleading when reading gate logs. Cheap fix: append `ctx.Ctx.Err()` context on this branch (e.g. "parent check budget exhausted") to distinguish it from the 8m path.

tests/uat/aws/run
```diff
@@ -221,13 +234,20 @@ phase_install() {
...
-  # its own per-attempt HELMFILE_TIMEOUT_SECONDS budget; the workflow's
-  # step `timeout-minutes` bounds total wall time.
-  local success=false
+  # partial-install state converges on the desired set. The 3 attempts SHARE a
+  # single HELMFILE_TIMEOUT_SECONDS wall-clock budget (each attempt is capped at
+  # the remaining budget, like the readiness gate below), not a full budget each:
+  # transient errors fail fast so sharing leaves ample time for retries, while a
+  # genuine hang cannot stretch install to 3x the budget. This bounds worst-case
+  # install (helmfile + gate) within the workflow step's timeout-minutes.
+  local helmfile_deadline=$(( SECONDS + HELMFILE_TIMEOUT_SECONDS ))
```
mchmarny, line 243: Flagging the behavior change for visibility (same in `tests/uat/gcp/run`): attempts previously each got a fresh `HELMFILE_TIMEOUT_SECONDS`; now an attempt that consumes the full 20m — including a slow-but-progressing apply, not just a hang — leaves zero retry budget, and the `(( SECONDS < helmfile_deadline )) || break` path exits without logging why attempts 2–3 never ran. The comment documents the trade-off, so no change requested; consider an explicit "shared budget exhausted after attempt N" echo on the break path so a budget-starved run reads differently from a 3-strike failure.

mchmarny (commented):
Well-built fix for the reboot-window race. `pollUntilStable` gets the hard parts right: the dwell restarts on regression, the poll is capped by the parent check budget (8m well under the 55m `CheckExecutionTimeout`), timeouts surface the last-observed state, and the ambiguous-match fail-fast is the correct transient-vs-deterministic split. The zero-expected-names fail-closed guard is untouched, the `TestMain`/sequential-phase tunable reasoning is correct, and the tests pin both ride-through and fail-closed paths. CI is green including the GPU smoke test. Four nits inline (zero-match polled as transient, healthy-path dwell cost, a misleading timeout message on the parent-budget path, and log visibility on the shared helmfile budget) — none block merge.

mchmarny (dismissed):
Approving — all inline findings are nits, none blocking. The poll/dwell design is correct (regression restarts the dwell, parent budget caps the poll, timeouts surface last-observed state), fail-closed paths are preserved, and the tests pin both ride-through and timeout behavior.

mchmarny (approved):

PR #1618, opened 2026-07-06
fix(ci): reap orphaned LB ELB/SG before UAT VPC teardown (#1617)

### LINKED ISSUE #1617: UAT teardown leaks VPCs: orphaned k8s-elb SG blocks DeleteVpc
## Summary

The AICR UAT pipeline is leaking a VPC on (nearly) every run whose test cluster provisions a Kubernetes `Service type=LoadBalancer`. `terraform destroy` fails at `DeleteVpc` with `DependencyViolation` because an orphaned `k8s-elb-*` security group (and sometimes the classic ELB itself) remains in the VPC — a resource created out-of-band by the in-tree AWS cloud provider that Terraform never manages.

This is distinct from the credential-expiry teardown bug (#1456). It reproduces even when destroy runs with valid credentials.

## Observed error

```
Error: deleting EC2 VPC (vpc-082fa88694f096a5d): operation error EC2: DeleteVpc,
https response error StatusCode: 400, api error DependencyViolation:
The vpc 'vpc-082fa88694f096a5d' has dependencies and cannot be deleted.
```

## Root cause

When a UAT test cluster creates a `Service type=LoadBalancer`, the in-tree AWS cloud provider provisions two resources **outside Terraform state**:

1. a classic ELB (`a<hash>`), and
2. a security group `k8s-elb-<hash>` attached to it.

The cloud controller only deletes these when the **Service** is deleted gracefully. The teardown deletes the cluster/nodes abruptly, so the cloud controller never runs its finalizer. Terraform then destroys everything it owns and hits `DeleteVpc`, which fails because the orphaned `k8s-elb-*` SG is still a dependency of the VPC.

Same underlying pattern as the leaked `eks-*` dev clusters (orphaned LoadBalancer ELBs + FSx PVs) — Kubernetes-managed cloud resources are not cleaned up before infrastructure teardown.

## Evidence (us-east-1, account 615299774277)

9 leaked VPCs found on 2026-07-06, **every one** carrying an orphaned `k8s-elb-*` SG; 2 also still had the ELB:

| VPC | Orphaned `k8s-elb` SG | ELB present |
|---|---|---|
| vpc-06f868e9f523f53b5 | ✓ | — |
| vpc-01f9ff45e17be19a7 | ✓ | — |
| vpc-082fa88694f096a5d | ✓ | — |
| vpc-08dd94fe41bc3e28a | ✓ | — |
| vpc-05db21ea7c95b5fc5 | ✓ | — |
| vpc-03d622b49ccdfb355 | ✓ | — |
| vpc-08c9b0739bf2613b4 | ✓ | — |
| vpc-05d4a6b5054ef5b6b | ✓ | ✓ (+2 subnets, IGW) |
| vpc-0e5dbbb7219325639 | ✓ | ✓ (+2 subnets, IGW) |

(All 9 have since been manually reaped.)

## Proposed fix

Add a **pre-`terraform destroy`** cleanup step to the UAT teardown:

1. Delete all LoadBalancer Services so the cloud controller reaps the ELB + SG, then wait for completion:
   ```
   kubectl delete svc --all-namespaces --field-selector spec.type=LoadBalancer --wait --timeout=5m
   ```
2. **Fallback** (in case the API server is already gone): before `DeleteVpc`, force-delete any classic ELBs and `k8s-elb-*` security groups in the target VPC.

Without this, any UAT run that exercises a LoadBalancer Service will keep leaking a VPC (and burning VPC-limit slots + ~$18/mo per run that still has the ELB).

## Acceptance criteria

- [ ] UAT teardown deletes Kubernetes LoadBalancer Services (and waits) before `terraform destroy`.
- [ ] Fallback sweep removes orphaned `a<hash>` ELBs and `k8s-elb-*` SGs in the VPC prior to `DeleteVpc`.
- [ ] A full UAT run + teardown leaves zero `aicr-uat-*` VPCs behind.

njhensley:
## Summary

Reap the out-of-band Kubernetes `LoadBalancer` ELB + `k8s-elb-*` security group before the UAT AWS `terraform destroy`, so an orphaned SG no longer blocks `DeleteVpc` and leaks the VPC.

## Motivation / Context

A UAT test cluster that creates a `Service type=LoadBalancer` makes the in-tree AWS cloud provider provision a classic ELB (`a<hash>`) and a security group (`k8s-elb-<hash>`) **outside Terraform state**, both tagged `kubernetes.io/cluster/<name>=owned`. These are only reaped when the Service is deleted gracefully. The teardown deletes the cluster abruptly, so the cloud-controller finalizer never runs; the orphaned SG stays a VPC dependency and `terraform destroy` fails at `DeleteVpc` with `DependencyViolation`, leaking the VPC (9 leaked on 2026-07-06, every one carrying a `k8s-elb-*` SG). Distinct from the credential-expiry teardown bug (#1456) — reproduces even with valid credentials.

Fixes: #1617
Related: #1456

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT AWS CI teardown (`.github/workflows/uat-aws.yaml`, `.github/scripts/`)

## Implementation Notes

New `.github/scripts/uat-aws-cleanup-lb.sh` with two modes, both wired into the `uat-aws` teardown and scoped strictly to this cluster's ownership tag:

- **graceful** — runs before `Destroy Cluster`: enumerates LoadBalancer Services (via jsonpath; `spec.type` is not a registered Service field selector, so `--field-selector` from the issue's sketch won't work) and deletes each with `--wait`, blocking on the `load-balancer-cleanup` finalizer so the cloud controller actually removes the ELB + SG. Refreshes its own kubeconfig, so it also covers `daytime-down` (which never ran *Connect to Cluster*). Best-effort — always exits 0.
- **sweep** — fallback for when the API server is already gone: runs **between failed destroy attempts** inside the existing retry loop. By the time an attempt fails at `DeleteVpc`, Terraform has already deleted the EKS cluster/nodes, leaving only the out-of-band LB SG; the sweep force-deletes the cluster-tagged classic ELBs (waiting for ENI release) then the `k8s-elb-*` SGs (retried through the transient `DependencyViolation`), so the next attempt's `DeleteVpc` succeeds.

Both filter on `kubernetes.io/cluster/<name>` (and `k8s-elb-*` name / VPC-scoped tag checks) so unrelated ELBs/SGs are never touched.

## Testing

```bash
# YAML + shell only (no Go); relevant gates:
yamllint .github/workflows/uat-aws.yaml          # clean
actionlint .github/workflows/uat-aws.yaml         # clean (matches merge-gate)
shellcheck .github/scripts/uat-aws-cleanup-lb.sh  # clean
bash -n .github/scripts/uat-aws-cleanup-lb.sh     # syntax OK; usage/mode guards exercised
```

Full end-to-end verification is the acceptance criterion — a UAT run + teardown leaving zero `aicr-uat-*` VPCs behind — validated by the next nightly UAT batch.

## Risk Assessment

**Rollout notes:** Teardown-only, best-effort, tag-scoped to the run's own cluster; the `terraform destroy` retry loop remains the source of truth. Reverting the two workflow hunks + deleting the script fully restores prior behavior.

## Checklist

- [x] Linter passes (yamllint + actionlint + shellcheck)

### DIFF
--- a/.github/scripts/uat-aws-cleanup-lb.sh
+++ b/.github/scripts/uat-aws-cleanup-lb.sh
@@ -1,0 +1,228 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# UAT teardown cleanup for Kubernetes-managed AWS load-balancer resources that
+# Terraform does not own (issue #1617).
+#
+# When a UAT test cluster creates a Service type=LoadBalancer, the in-tree AWS
+# cloud provider provisions, OUTSIDE Terraform state:
+#   * a classic ELB           (name  a<hash>)
+#   * a security group        (name  k8s-elb-<hash>) attached to it
+# Both are tagged kubernetes.io/cluster/<cluster-name>=owned. They are only
+# reaped when the Service is deleted gracefully; an abrupt cluster teardown
+# skips that finalizer, so the orphaned k8s-elb SG stays a dependency of the
+# VPC and `terraform destroy` fails at DeleteVpc with DependencyViolation —
+# leaking the VPC (and, when the ELB survives too, ~$18/mo per run).
+#
+# Two modes:
+#   graceful  Delete every LoadBalancer Service and wait for the cloud
+#             controller to reap the ELB + SG. Requires a reachable API server;
+#             best-effort (the sweep is the backstop) — always exits 0.
+#   sweep     Fallback for when the API server is already gone: force-delete the
+#             orphaned classic ELBs and k8s-elb-* SGs tagged for this cluster so
+#             a subsequent DeleteVpc succeeds. Operates purely against the AWS
+#             API. Scoped to this cluster's ownership tag, never touching
+#             unrelated resources.
+#
+# Usage:
+#   AWS_REGION=us-east-1 ./uat-aws-cleanup-lb.sh graceful <cluster-name>
+#   AWS_REGION=us-east-1 ./uat-aws-cleanup-lb.sh sweep    <cluster-name>
+#
+# No -e: each step tolerates partial failure so cleanup keeps going and the
+# terraform-destroy retry loop remains the source of truth for teardown.
+set -uo pipefail
+
+MODE="${1:-}"
+CLUSTER="${2:-}"
+
+if [[ -z "${MODE}" || -z "${CLUSTER}" ]]; then
+  echo "Usage: $0 <graceful|sweep> <cluster-name>" >&2
+  exit 2
+fi
+: "${AWS_REGION:?Set AWS_REGION}"
+
+# Fail closed on an unexpected cluster name: it is interpolated into AWS filter
+# values (the ownership tag key) and a JMESPath predicate, so a stray '*' or
+# "'" would widen the sweep beyond this cluster (e.g. wildcard-matching every
+# daytime cluster's tag in the shared account). EKS cluster names are
+# [A-Za-z0-9-] anyway; reject anything else rather than trust the caller.
+if ! [[ "${CLUSTER}" =~ ^[A-Za-z0-9][A-Za-z0-9-]*$ ]]; then
+  echo "refusing to run: invalid cluster name '${CLUSTER}' (want ^[A-Za-z0-9][A-Za-z0-9-]*\$)" >&2
+  exit 2
+fi
+
+# Wait budget for a LoadBalancer Service deletion (cloud-controller reap). Kept
+# well under the teardown budget — the sweep is the backstop, so a slow reap
+# must not starve `terraform destroy` of the job-timeout it needs.
+LB_SVC_TIMEOUT="${LB_SVC_TIMEOUT:-3m}"
+# Per-API-call ceiling so a reachable-but-blackholed API server can't hang the
+# step past the LB_SVC_TIMEOUT window.
+KUBECTL_REQUEST_TIMEOUT="${KUBECTL_REQUEST_TIMEOUT:-30s}"
+# Ownership tag the in-tree AWS cloud provider stamps on the ELB and its SG.
+# The value distinguishes cluster-owned resources (owned) from shared ones
+# (shared, e.g. a co-tenant VPC's subnets). Match BOTH key and value so the
+# sweep never touches a `shared`-tagged resource from another cluster.
+OWNER_TAG_KEY="kubernetes.io/cluster/${CLUSTER}"
+OWNER_TAG_VALUE="owned"
+
+# graceful: delete LoadBalancer Services and wait for the cloud controller to
+# tear down the backing ELB + SG. Best-effort; the sweep covers the case where
+# the API server is unreachable.
+graceful() {
+  echo "graceful: refreshing kubeconfig for ${CLUSTER}"
+  if ! aws eks update-kubeconfig --region "${AWS_REGION}" --name "${CLUSTER}"; then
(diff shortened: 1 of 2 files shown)


.github/scripts/uat-aws-cleanup-lb.sh
```diff
@@ -0,0 +1,207 @@
...
+      || echo "sweep: delete of ELB ${name} failed"
+  done
+
+  # Wait for the ELB(s) to fully deregister so the SG stops being an attachment.
+  local remaining
+  for _ in $(seq 1 12); do
+    remaining="$(aws elb describe-load-balancers --region "${AWS_REGION}" \
+      --query "LoadBalancerDescriptions[?VPCId=='${vpc}'].LoadBalancerName" \
+      --output text 2>/dev/null)"
+    [[ -z "${remaining//[[:space:]]/}" ]] && return 0
```
mchmarny: This wait polls for *zero* classic ELBs in the VPC, but the loop above deliberately leaves untagged ELBs alone ("not tagged for ${CLUSTER}; leaving it"). If a co-tenant ELB shares the VPC, `remaining` never empties, so every sweep burns the full 120s and returns without actually confirming the owned ELBs are gone. Track the names you deleted and wait only on those (e.g. re-describe with `--load-balancer-names` and treat `LoadBalancerNotFound` as done). Low likelihood on a per-run UAT VPC, but the wait as written can't distinguish "still deregistering" from "never going away".

.github/scripts/uat-aws-cleanup-lb.sh
```diff
@@ -0,0 +1,207 @@
...
+  echo "graceful: deleting LoadBalancer Services:"
+  echo "${lbs}"
+  # Delete with --wait so kubectl blocks on the load-balancer-cleanup finalizer,
+  # i.e. until the cloud controller has actually removed the ELB + SG.
+  while read -r ns name; do
+    [[ -z "${ns}" || -z "${name}" ]] && continue
+    echo "graceful: deleting svc ${ns}/${name}"
+    kubectl delete svc "${name}" -n "${ns}" --wait=true \
+      --request-timeout="${KUBECTL_REQUEST_TIMEOUT}" --timeout="${LB_SVC_TIMEOUT}" \
+      || echo "graceful: delete of ${ns}/${name} did not complete cleanly; sweep will reconcile"
```
mchmarny: Verify `--request-timeout=30s` doesn't truncate the 3m `--wait`: the per-request timeout also governs the watch/poll requests backing the delete waiter. kubectl's wait path re-establishes on `ErrWatchClosed` in recent versions, but this behavior has been inconsistent across releases — if it errors out at ~30s, the graceful mode's effective wait is 30s, not `LB_SVC_TIMEOUT`, and everything falls to the sweep. Consider dropping `--request-timeout` on the delete call (keeping it on the `get` above) or setting it >= `LB_SVC_TIMEOUT` here. Non-blocking either way — the sweep backstops.

.github/scripts/uat-aws-cleanup-lb.sh
```diff
@@ -0,0 +1,207 @@
...
+  # the ELB's ENIs deregister.
+  local sg attempt
+  while read -r sg _vpc; do
+    [[ -z "${sg}" ]] && continue
+    for attempt in 1 2 3 4 5 6; do
+      if aws ec2 delete-security-group --region "${AWS_REGION}" --group-id "${sg}"; then
+        echo "sweep: deleted SG ${sg}"
+        break
+      fi
+      echo "sweep: SG ${sg} still has dependencies (attempt ${attempt}); retrying..."
```
mchmarny: nit: on attempt 6 this prints "retrying..." and then the loop silently gives up — the final-failure case is indistinguishable from a retry in the logs. Guard the message like the destroy loop does (`if attempt < 6 ... else "giving up on ${sg}"`), so a persistently stuck SG is visible when triaging a leaked VPC.

mchmarny (commented):
Solid fix for the VPC leak: the tag scoping (key AND value=owned), fail-closed cluster-name validation, best-effort design with the terraform-destroy retry as source of truth, and skipping the sweep after the final attempt are all right. CI is green. Three non-blocking findings inline: the ELB-teardown wait can never converge when an untagged ELB shares the VPC (medium-low), a question on whether `--request-timeout` truncates the graceful `--wait`, and a logging nit on the SG retry loop. Nothing blocks merge.

mchmarny (approved):
Approving — all three review findings addressed in e05a2e4: the post-delete wait now polls the specific ELB names it deleted (treating LoadBalancerNotFound as done) so a co-tenant ELB can't stall convergence, `--request-timeout` was dropped from the `--wait` delete so the 3m budget actually applies, and the SG retry loop now emits a `::warning::` giving-up line on the final attempt. Tag-scoped, fail-closed on bad input, best-effort by design with the destroy retry as source of truth.

Issue #1611, opened 2026-07-06
chore(recipes): represent explicit per-source version variants in the BOM


yuanchen8911:
## Summary

Follow-up to #966 (superseded by #1424 / #1572 / #1580). Scope is the remaining BOM-variant gap only; the empty-version check and redundant-pin cleanup are tracked independently in #1615 and #1616.

## Problem

`tools/bom` renders each component once, at its registry `defaultVersion`. A base, overlay, or mixin can explicitly pin a Helm `componentRef.version` that differs from that default. The differing pin is part of the recipe data that deployment actually consumes, but the corresponding chart version and rendered image set are invisible in both `docs/user/container-images.md` and `bom.cdx.json`.

The recipe pin and registry default are the source facts. `versionPinExemptions` is validation policy for deciding whether a divergence is allowed; it must not be the BOM's source of deployed variants.

## Design decisions

1. **Fidelity: catalog parity, not deployment fidelity.** Variant entries are rendered at catalog parity—the chart at the variant version with the shared `recipes/components/<name>/values.yaml`—using exactly the inputs default entries use today. Effective overlay `overrides:` remain invisible to both default and variant entries; state this uniform limitation in both the Markdown header and CycloneDX metadata. Full effective-values rendering would relax #1572's non-goal and is out of scope. Variant image sets must be rendered, not copied from the default entry.

2. **Variant discovery: derive from recipe data.** Load the component registry and recipe metadata from the same `-repo-root`. Inspect explicit Helm `componentRef.version` pins declared by the base, overlays, and mixins. Compare each pin with that component's registry `defaultVersion`; when they differ, aggregate a variant by `(component, version)` and record the sorted declaring source names. Default-equal pins produce no variant. `tools/bom` must not import or read `versionPinExemptions`.

3. **CycloneDX identity.** Preserve existing default-component bom-refs unchanged. Only variant entries receive version-qualified refs (for example, `<name>@<version>`). Attach a sorted `aicr:variant:sources` property. Ordering must be deterministic, and duplicate bom-refs must fail closed.

4. **Markdown projection and freshness.** Render variants as a separate table so the existing one-row-per-registry-component structure remains intact. Extend the freshness gate bidirectionally against the variants derived from recipe pins: a derived variant missing from the document fails, and a documented variant with no backing differing pin fails. The existing default-row gate remains unchanged.

[...]

PR #1619, opened 2026-07-06
chore(recipes): move version-pin exemptions to a declarative file

yuanchen8911:
## Summary

Move the version-pin exemption table out of `pkg/recipe/version_pin_guard_test.go` into a declarative, schema-validated file — `recipes/version-pin-exemptions.yaml` — loaded by the new `internal/versionpins` package. No intended behavior change: the guard test enforces the same invariant, from the same single entry (aks/kube-prometheus-stack).

## Motivation / Context

Preparation for #1611 (represent exempted per-overlay version variants in the BOM): `tools/bom` cannot import a Go table defined in a `_test.go` file, and the policy must not be duplicated. A declarative file under `recipes/`, read from repo root, also guarantees `tools/bom -repo-root` reads exemption policy and `recipes/registry.yaml` from the same checkout — a compiled-in table could apply one checkout's policy to another checkout's registry.

Fixes: N/A
Related: #1611

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `internal/versionpins` (new), `pkg/defaults` (new file-size cap constant)

## Implementation Notes

- The loader lives under `internal/` (not `pkg/`) because it is repository-only dev/CI tooling, not public Go API; both the guard test and (with #1611) `tools/bom` can import it within the module.
- `recipes/version-pin-exemptions.yaml` is deliberately NOT added to the `recipes/data.go` embed patterns, so it does not ship in the binary and cannot interact with the runtime data providers or external `--data` merging. `recipes/README.md` documents the exception.
- `versionpins.Load(ctx, path)` fails closed on every defect — missing file (`ErrCodeNotFound` only on `fs.ErrNotExist`), non-regular file (fstat on the opened handle inside the bounded worker, so no stat-to-open race), oversize (`defaults.MaxVersionPinExemptionsBytes`), unknown fields (strict decode), wrong `apiVersion`/`kind` (validated against the canonical `header.GroupVersion`), a second YAML document, an omitted-or-null `exemptions` key (a truncated file must not read as "no exemptions"; an explicit `[]` is required), missing or whitespace-only entry fields, duplicate `(source, component)` keys, and `expectedPin == expectedDefault`. Validation problems are aggregated into one `ErrCodeInvalidRequest` error.
- Open, the fstat type check, and the read all run inside context-bounded workers under `defaults.FileReadTimeout` (tightening any caller deadline) — a FIFO blocking in `open(2)` or a hung network/FUSE mount cannot stall the caller past the deadline. Handle delivery is an unbuffered rendezvous, so the worker always knows whether the caller took the handle and closes it itself on every error/late-completion path — no drainer goroutine. If the underlying syscall never returns, exactly one goroutine (the worker) stays pinned until it does; Go cannot cancel a blocked syscall (the same trade `pkg/serializer.readAllBounded` makes). Helpers classify failures as structured `pkg/errors` codes once, and `Load` propagates them without re-wrapping. Deterministic seam tests drive blocked-open and blocked-fstat with explicit synchronization (async Load, await started, cancel, bounded done check, release) and verify the late handle is closed by requiring fs.ErrClosed. Every fatal exit path releases the blocked worker and then joins the async Load via a t.Cleanup registered before the seam restore (release is idempotent on success and retryable on failure), so no failure can strand a worker or restore seams while a Load is still in flight. A darwin/linux-gated test proves the same against a genuine writer-less FIFO, using explicit post-started cancellation (no deadline race) and a deterministic release that fails the test if the worker would be left pinned.
- The static checks previously inlined in the guard test (duplicates, empty fields, pin == default) moved into the loader so every consumer gets them; the guard-only checks (out-of-date expectations, stale entries, sole-consumer enforcement) stay in the test.
- `tools/bom` is intentionally NOT wired to consume the file in this PR — that is #1611's implementation PR. This PR only makes the shared source available.
- `make bom-docs` regen is not required: no registry entry, chart-version pin, or component values changed.

## Testing

```bash
make qualify
```

- `make qualify` passes end-to-end (test-coverage 78.6% vs 75% threshold, lint 0 issues, tuning-check, e2e "All tests passed", scan "No vulnerabilities found", license-check).
- New package `internal/versionpins`: 90.0% statement coverage (new package, no baseline). Table-driven tests cover valid/empty documents, omitted/null exemptions key, strict-decode rejection, identity (kind/apiVersion) mismatch, multi-document rejection, per-field and whitespace-only validation, duplicates, pin==default, missing file, non-regular file, canceled context, oversize file, and the committed repository file itself.
- `TestOverlayVersionPinsMatchRegistry` passes against the file-based table: 38 pinned componentRefs verified, 1 declared exemption fires with identical semantics to the previous Go table.
- Coverage delta on changed packages: `pkg/recipe` 87.3% → 87.3% (test-only edit); `pkg/defaults` 100% → 100% (constant only).

## Risk Assessment

**Rollout notes:** N/A — dev/CI-time policy only; the file is not embedded in the binary and no runtime path reads it.

## Checklist

### DIFF
--- a/docs/contributor/recipe.md
+++ b/docs/contributor/recipe.md
@@ -518,8 +518,10 @@ registry default. A version bump must update the registry default (the
 BOM reads it) **and** every pin that sets it, or CI fails. If an
 overlay must legitimately run a different chart version (e.g. a
 platform validated against an older chart), add an entry to
-`versionPinExemptions` with a justification — a declared divergence is
-not a silent one. Do **not** exempt a component whose only consumer
+`recipes/version-pin-exemptions.yaml` with a justification — a declared
+divergence is not a silent one. The file is loaded and validated by
+`internal/versionpins`, the shared source consumed by the guard test
+today and by `tools/bom` once #1611 lands. Do **not** exempt a component whose only consumer
 diverges; that reinstates the exact fiction the guard prevents.
 
 ## Common Pitfalls
--- a/internal/versionpins/versionpins.go
+++ b/internal/versionpins/versionpins.go
@@ -1,0 +1,326 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package versionpins loads the repository's declarative version-pin
+// exemption table (recipes/version-pin-exemptions.yaml): componentRefs whose
+// overlay/mixin version pin is INTENTIONALLY different from the component's
+// registry defaultVersion/defaultTag.
+//
+// The file is repository-only policy data — deliberately NOT embedded into
+// the aicr binary (see recipes/data.go) and not part of the runtime recipe
+// data model, which is why this package lives under internal/ rather than
+// pkg/: it is dev/CI tooling, not public Go API. It is consumed from a
+// repository checkout by the version-pin guard test
+// (pkg/recipe/version_pin_guard_test.go) and, once #1611 wires it, by the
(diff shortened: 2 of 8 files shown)


yuanchen8911:
Closing because #1611 has been revised to derive BOM variants directly from actual base/overlay/mixin version pins compared with registry defaults. A shared exemption file is therefore no longer a BOM prerequisite, and this loader/policy machinery is unnecessary. The existing guard policy can remain test-local while its longer-term representation is considered separately.

PR #1631, opened 2026-07-07
feat(validator): GKE NCCL preflight + launcher failure mitigations

njhensley:
## Summary

Add a GKE GPUDirect-TCPXO readiness preflight to the NCCL all-reduce check, widen the launcher's SSH retry budget, and surface worker-pod diagnostics on launcher failure — so a GKE/H100 run that currently dies with an opaque "pod failed" fails fast and legibly (or tolerates slow worker bring-up) instead.

## Motivation / Context

On GKE H100, the `nccl-all-reduce-bw` check applies a Kubeflow TrainJob whose launcher `mpirun` SSHes to worker pods. On freshly provisioned UAT nodes two things race the launcher: (1) the worker containers install `openssh-server` and source the TCPXO env profile at container start before sshd comes up, and (2) the `nccl-tcpxo-installer` DaemonSet may not have laid down `nccl-env-profile.sh` / FastRak libs yet. Either causes `mpirun` to exhaust its SSH connection attempts and the launcher pod to enter `Failed`. The check then reports only `pod failed` because the launcher log-fetch error was discarded and no worker context was gathered — making root-cause impossible after the fact.

Fixes: N/A
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Validator (`pkg/validator`) — specifically the standalone `validators/performance` NCCL check binary

## Implementation Notes

- **GKE TCPXO preflight** (`nccl_preflight_tcpxo.go`): `gkeTCPXOPreflightApplies` gates on `default variant + GKE + H100`; `preflightGKETCPXOReady` fans out a per-node busybox probe that checks for `/home/kubernetes/bin/nvidia/lib64/nccl-env-profile.sh` (the exact artifact the worker startup chain sources). Missing → `ErrCodeInvalidRequest` naming the unready nodes with a fix hint. Modeled on the existing GB200 NVreg preflight.
- **Launcher SSH retry budget** (`testdata/h100/gke/runtime.yaml`): `ConnectionAttempts=10` → `-o ConnectTimeout=10 -o ConnectionAttempts=300` (~5-min bounded reachability; refused-port retries are ~1s apart, so the ceiling ≈ attempts, and the outer NCCLTrainJobTimeout bounds the unreachable-IP case). Mirrors the rationale already present in the RoCE/EKS template.
- **Failure diagnostics** (`nccl_all_reduce_bw_constraint.go`): `waitForLauncherPodAndGetLogs` no longer discards the launcher log-fetch error and, on failure, appends `collectNCCLWorkerDiagnostics` (per-worker phase, container terminated/waiting reason + exit code + message, and best-effort `node`/`tcpxo-daemon` logs).
- **Refactor**: extracted the shared per-node preflight fan-out into `runPerNodeProbe` (`nccl_preflight.go`) so the new TCPXO preflight and the existing NVreg preflight share one implementation; added `shellBin`/`nodeJobName` constants.

## Testing

```bash
go build ./validators/...
golangci-lint run -c .golangci.yaml ./validators/performance/...   # 0 issues
go test -race ./validators/performance/...                          # ok
```

- Added unit tests: `TestGKETCPXOPreflightApplies` (gate matrix) and `TestCollectNCCLWorkerDiagnostics` (no-pods + failed/waiting container states).
- Package coverage: `validators/performance` 39.4% → 39.6% (+0.2%).
- GKE runtime template re-validated as a well-formed `TrainingRuntime` after the edit.
- Scoped runs above rather than a full `make qualify` locally; relying on CI for the module-wide gate.

## Risk Assessment

- [x] **Low** — Isolated to the GKE/H100 NCCL check path; the preflight only adds a fail-fast gate, the retry bump is strictly more permissive, and the diagnostics only run on the existing failure path. Easy to revert.

**Rollout notes:** N/A — no config, API, or recipe surface changes; behavior change is confined to the `nccl-all-reduce-bw` validator on GKE/H100.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — affected package run with `-race`
- [x] Linter passes (`make lint`) — `golangci-lint` on affected package, 0 issues
- [x] I updated docs if user-facing behavior changed — N/A (internal validator behavior only)


validators/performance/testdata/h100/gke/runtime.yaml
```diff
@@ -87,7 +87,13 @@ spec:
...
                   - --mca
                   - plm_rsh_args
-                  - -o StrictHostKeyChecking=no -o ConnectionAttempts=10
+                  # Worker "node" pods install openssh-server and source the
+                  # TCPXO env profile at container start before sshd comes up,
+                  # so on freshly provisioned nodes sshd can lag the launcher by
+                  # minutes. Give mpirun's rsh a bounded reachability budget
+                  # (ConnectTimeout 10s x ConnectionAttempts 60 ~= 10 min) so a
+                  # slow worker bring-up retries instead of aborting the launcher.
+                  - -o StrictHostKeyChecking=no -o ConnectTimeout=10 -o ConnectionAttempts=60
```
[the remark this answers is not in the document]
njhensley: Good catch on the arithmetic — corrected in 5e7aeb29. You are right that a refused port (sshd not yet listening on a Running worker) fails immediately and `ConnectionAttempts` retries ~1s apart, so `ConnectTimeout` does not extend the window. Rather than add a separate in-shell readiness poll (mpirun's `plm_rsh` already *is* the retry loop, and computing the worker hostnames in-shell is fragile), I raised `ConnectionAttempts` 60 → 300 so the real ceiling is ~5 min, comfortably covering fresh-node worker bring-up, and rewrote the comment to state the ~1s/attempt semantics accurately. The new GKE TCPXO preflight also fails fast when the installer artifacts are missing, so the retry window only has to absorb sshd startup, not the whole node provisioning path.

validators/performance/nccl_all_reduce_bw_constraint.go
```diff
@@ -1180,6 +1205,70 @@ func waitForLauncherPodAndGetLogs(ctx *validators.Context, podHelper *helper.Pod
...
+		for _, container := range []string{nodeJobName, "tcpxo-daemon"} {
+			logs, logErr := k8spod.GetPodLogs(diagCtx, clientset, namespace, p.Name, container)
+			if logErr != nil {
+				fmt.Fprintf(&b, "  [%s logs unavailable: %v]\n", container, logErr)
+				continue
+			}
+			if trimmed := strings.TrimSpace(logs); trimmed != "" {
+				fmt.Fprintf(&b, "  --- %s/%s logs ---\n%s\n", p.Name, container, trimmed)
+			}
+		}
```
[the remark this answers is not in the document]
njhensley: Addressed in 5e7aeb29.

validators/performance/nccl_all_reduce_bw_constraint.go
```diff
@@ -1180,6 +1205,70 @@ func waitForLauncherPodAndGetLogs(ctx *validators.Context, podHelper *helper.Pod
...
+// phase, each container's terminal state (reason/exitCode/message) or waiting
+// reason, and the tail of the "node" and "tcpxo-daemon" container logs. The
+// most common root cause — worker sshd never started (slow apt-get, missing
+// TCPXO env profile) or the tcpxo-daemon sidecar crashing — shows up here even
+// when the launcher's own log is empty.
+func collectNCCLWorkerDiagnostics(ctx context.Context, clientset kubernetes.Interface, namespace string) string {
+	diagCtx, cancel := context.WithTimeout(ctx, defaults.DiagnosticTimeout)
+	defer cancel()
+
+	selector := fmt.Sprintf("jobset.sigs.k8s.io/jobset-name=%s,jobset.sigs.k8s.io/replicatedjob-name=node", ncclTrainJobName)
```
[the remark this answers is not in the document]
njhensley: Addressed in 5e7aeb29.

validators/performance/nccl_preflight_tcpxo_test.go
```diff
@@ -0,0 +1,128 @@
...
+			"container tcpxo-daemon: terminated reason=Error exitCode=137",
+			"fastrak init failed",
+			"container node: waiting reason=CrashLoopBackOff",
+		} {
+			if !strings.Contains(got, want) {
+				t.Errorf("diagnostics missing %q\nfull output:\n%s", want, got)
+			}
+		}
+	})
+}
```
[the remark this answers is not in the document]
njhensley: Addressed in 5e7aeb29.

njhensley:
Addressed both outside-diff-range comments in 696551d0:

- **Launcher logs unbounded (1181-1194):** the launcher log is now tailed to `maxDiagLogLines` too, matching the worker-diagnostics cap.
- **Sequential per-pod fetches share one deadline (1246-1288):** per-worker diagnostics now run concurrently via `errgroup` bounded by `preflightNodeConcurrency`, with order preserved through an indexed result slice (extracted into `workerPodDiagnostics`), so later workers on a large job are no longer starved of the shared `DiagnosticTimeout`.

validators/performance/nccl_preflight_tcpxo.go
```diff
@@ -0,0 +1,204 @@
...
+
+	created, err := podsClient.Create(ctx, pod, metav1.CreateOptions{})
+	if err != nil {
+		return false, aicrErrors.Wrap(aicrErrors.ErrCodeInternal,
+			"failed to create TCPXO preflight pod", err)
+	}
+	// Cleanup runs on an independent context so it fires even when the probe
+	// context has been canceled (timeout, parallel-sibling error).
+	defer func() { //nolint:contextcheck // Fresh context: parent may be canceled during cleanup
+		cleanupCtx, cancel := context.WithTimeout(context.Background(), defaults.PreflightCleanupTimeout)
```
mchmarny, line 188: The `HostPath` volume has no explicit `Type`. If `/home/kubernetes/bin/nvidia` doesn't exist at all on a node (installer DaemonSet never ran — arguably the strongest form of the condition this preflight exists to catch), the empty type leaves behavior runtime-dependent: some runtimes create a root-owned dir (probe then correctly reports not-ready), others fail the mount and leave the pod stuck in ContainerCreating until the 2-minute `DiagnosticTimeout` expires — which then aborts the whole fan-out as `ErrCodeInternal` ("probe failed") instead of naming the node as not-ready.

Consider an explicit `Type: &hostPathDirectoryOrCreate` (matches what containerd does implicitly, and the dir is harmless — the installer owns its contents), or mount a guaranteed-existing parent like `/home/kubernetes/bin` and test the full relative path.

validators/performance/nccl_all_reduce_bw_constraint.go
```diff
@@ -1180,6 +1211,117 @@ func waitForLauncherPodAndGetLogs(ctx *validators.Context, podHelper *helper.Pod
...
+		case cs.State.Running != nil:
+			fmt.Fprintf(&b, "  container %s: running (ready=%t)\n", cs.Name, cs.Ready)
+		}
+	}
+	// Best-effort container logs. "node" is the worker itself (sshd + NCCL);
+	// "tcpxo-daemon" is the FastRak sidecar whose failure also downs the run.
+	// GetPodLogs streams the full log — a verbose NCCL/apt-get worker can emit
+	// thousands of lines — so tail each container to the last maxDiagLogLines
+	// before appending, keeping the failure payload bounded. The tail (not the
+	// head) is kept because the fatal error is almost always the last output.
```
mchmarny: `collectNCCLWorkerDiagnostics` runs on every platform's launcher-failure path (EKS, OKE, ...), but the container list here hardcodes `tcpxo-daemon`, which only exists in the GKE template. On non-GKE failures every worker gets a noise line: `[tcpxo-daemon logs unavailable: container "tcpxo-daemon" not found ...]`.

Consider deriving the container names from `p.Spec.InitContainers`/`p.Spec.Containers` instead — it removes the noise and also stays correct if a template renames its sidecar.

validators/performance/testdata/h100/gke/runtime.yaml
```diff
@@ -87,7 +87,16 @@ spec:
...
+                  # Worker "node" pods install openssh-server and source the
+                  # TCPXO env profile at container start before sshd comes up,
+                  # so on freshly provisioned nodes sshd can lag the launcher by
+                  # minutes. mpirun's rsh IS the retry loop here: while sshd is
+                  # not listening the port is refused, so ssh retries ~1s apart
+                  # (ConnectTimeout only bounds the unreachable-IP case, not a
+                  # refusal), making the real ceiling ~= ConnectionAttempts
+                  # seconds. Size it to ~5 min so a slow worker bring-up retries
+                  # instead of aborting the launcher.
+                  - -o StrictHostKeyChecking=no -o ConnectTimeout=10 -o ConnectionAttempts=300
```
mchmarny, line 105: Two things here:

1. The comment's "~ConnectionAttempts seconds" ceiling only holds for the connection-refused case. For an unreachable IP (worker pod never scheduled, network policy drop) each attempt burns the full `ConnectTimeout=10`, so the worst case is ~300 × 11s ≈ 55 min — past `NCCLTrainJobTimeout` (30 min), meaning that path fails as a wait-timeout rather than an mpirun abort. Not wrong (the outer timeout still bounds it), but worth a sentence in the comment so nobody "fixes" the timeout interplay later.

2. nit: the PR description says `ConnectionAttempts=60 (~10-min)` — stale vs. the actual `300` here; update the body before merge.

validators/performance/nccl_all_reduce_bw_constraint.go
```diff
@@ -1180,6 +1211,117 @@ func waitForLauncherPodAndGetLogs(ctx *validators.Context, podHelper *helper.Pod
...
+	}
+	if len(pods.Items) == 0 {
+		return "\n--- no NCCL worker pods found (workers may never have been scheduled) ---\n"
+	}
+
+	// Fetch each worker's diagnostics concurrently with bounded concurrency:
+	// sequential per-pod log fetches would share the single DiagnosticTimeout,
+	// so on a large job later workers could exhaust it and lose diagnostics.
+	// Order is preserved via an indexed result slice; workerPodDiagnostics is
+	// best-effort and never errors, so g.Wait never cancels the group early.
```
mchmarny, line 1256: nit: `preflightNodeConcurrency` is documented (and named) as a cap on in-flight probe *Pods* creating scheduler pressure; here it's reused to bound log fetches, which are much cheaper apiserver reads. Fine to share the value, but a locally named constant (or a rename to something like `perNodeFanoutConcurrency`) would keep the doc comment honest.

mchmarny (commented):
Adds a GKE/H100 TCPXO readiness preflight, widens the launcher SSH retry budget, and surfaces worker-pod diagnostics on launcher failure — all confined to the NCCL check path. Overall this is well done: the `runPerNodeProbe` extraction is a clean dedup of the NVreg fan-out, the probe correctly distinguishes not-ready from infrastructure errors (marker-on-stdout is a nice touch), and the new diagnostics are unit-tested. CI is fully green.

Nothing blocking. One medium item (unset `HostPath` type can turn the "installer never ran" case — the exact scenario the preflight targets — into a 2-minute ErrCodeInternal timeout instead of a not-ready report), plus a cross-platform noise issue in the worker diagnostics and two nits. Also: branch is behind `main` — rebase before merge, and the PR body's `ConnectionAttempts=60` is stale vs. the actual `300`.

validators/performance/nccl_preflight_tcpxo.go
njhensley: Fixed in 1041ab1b — the probe HostPath now sets `Type: DirectoryOrCreate`. That makes the installer-never-ran case deterministic: the dir is created empty, `test -f` reports not-ready, and the node is named in the fail-fast error instead of the probe hanging in ContainerCreating until DiagnosticTimeout. The empty dir is harmless (the installer owns and populates its contents).

validators/performance/nccl_all_reduce_bw_constraint.go
njhensley: Fixed in 1041ab1b — `workerPodDiagnostics` now derives container names from `p.Spec.InitContainers` + `p.Spec.Containers` rather than hardcoding `[node, tcpxo-daemon]`. No more spurious "container not found" line on EKS/OKE, and a renamed sidecar still gets its logs captured.

validators/performance/testdata/h100/gke/runtime.yaml
njhensley: Both addressed in 1041ab1b: expanded the comment to spell out the unreachable-IP case (each attempt burns the full `ConnectTimeout`, worst case ~55 min, but bounded by the outer `NCCLTrainJobTimeout` and surfacing as a wait-timeout, with a note to keep the two timeouts in mind together), and corrected the stale `60 (~10-min)` in the PR description to `300 (~5-min)`.

validators/performance/nccl_all_reduce_bw_constraint.go
njhensley: Renamed to `perNodeFanoutConcurrency` in 1041ab1b and generalized its doc to say it bounds both preflight probe Pods and diagnostics log fetches — so the name and comment now match both use sites.

mchmarny (approved):

PR #1643, opened 2026-07-07
fix(validator): skip orphan pods in negative health assertions

njhensley:
## Summary

The in-process chainsaw executor's negative (`error`) health assertions now skip orphaned pods (Terminating or NodeLost) so a ghost pod left behind by node churn no longer fails an otherwise-healthy check.

## Motivation / Context

In an AWS UAT inference run, the `aws-ebs-csi-driver` `expected-resources` check failed with `Pod kube-system/ebs-csi-node-xhpmz matches forbidden shape` — after the same check had passed twice earlier in the run. The failure burned the full 5-minute `assert` timeout, meaning the pod matched a forbidden shape continuously and never recovered.

Root cause: `ebs-csi-node-*` is a DaemonSet (one pod per node). When a GPU/inference node is replaced or lost mid-run (validate stage), its `ebs-csi-node` pod lingers in phase `Unknown`/`Failed` (or with a `deletionTimestamp`) for minutes until pod GC removes it, while a healthy replacement already runs on the live node. The negative list-and-match loop in `evaluateError` had **no filter for orphaned pods**, so it flagged the ghost — the exact orphan-pod trap documented in the project anti-patterns (`pods.Items[0]` after a label-selector List → filter `DeletionTimestamp != nil`).

Fixes: N/A
Related: N/A

## Type of Change

## Component(s) Affected

## Implementation Notes

- `evaluateError` skips resources with a `metadata.deletionTimestamp` or `status.reason == "NodeLost"` before evaluating the forbidden shape, so negative assertions reflect the state of live nodes only. Positive (`assert`) matching is intentionally unchanged.
- The `forbidden shape matched` message now appends `(phase=…, waiting=…, node=…)` so future failures are self-diagnosing from the check report instead of an opaque shape reference.

## Testing

```bash
make qualify
```

- Full `make qualify` passed (unit tests with `-race`, lint, chainsaw e2e, vulnerability scan).
- Added table cases (terminating ghost skipped, NodeLost ghost skipped, live pod beside a ghost still fails) plus focused `TestIsTerminatingOrLost` / `TestDescribePodStatus`.
- New helpers covered 100% / 100% / 88.2%; `validators/chainsaw` package coverage unchanged-or-up.

## Risk Assessment

- [x] **Low** — Isolated change to the negative-assertion path, well-tested, easy to revert.

**Rollout notes:** N/A — health-check evaluation only; no API or recipe surface changes.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — internal evaluator behavior)

### DIFF
--- a/validators/chainsaw/inprocess.go
+++ b/validators/chainsaw/inprocess.go
@@ -19,6 +19,7 @@ import (
 	stderrors "errors"
 	"fmt"
 	"log/slog"
+	"strings"
 	"time"
 
 	"github.com/kyverno/chainsaw/pkg/apis"
@@ -446,6 +447,15 @@ func evaluateError(ctx context.Context, e *v1alpha1.Error, fetcher ResourceFetch
 			return errors.Wrap(errors.ErrCodeInternal,
 				fmt.Sprintf("list-match canceled for %s in %q", kind, namespace), err)
 		}
+		// Skip ghosts (Terminating or NodeLost): a node replaced or lost
+		// mid-run leaves its DaemonSet/workload pod behind for minutes until
+		// pod GC, while a healthy replacement already runs on the live node.
+		// A negative (error) assertion must not fire on such an orphan — that
+		// is the orphan-pod trap documented in the project anti-patterns, and
+		// it turned a healthy ebs-csi-node DaemonSet into a false failure.
+		if isTerminatingOrLost(actual) {
+			continue
+		}
 		errs, checkErr := checks.Check(ctx, apis.DefaultCompilers, actual, nil, &check)
 		if checkErr != nil {
 			return errors.Wrap(errors.ErrCodeInvalidRequest,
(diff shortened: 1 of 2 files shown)


validators/chainsaw/inprocess.go
```diff
@@ -446,6 +447,15 @@ func evaluateError(ctx context.Context, e *v1alpha1.Error, fetcher ResourceFetch
 			return errors.Wrap(errors.ErrCodeInternal,
 				fmt.Sprintf("list-match canceled for %s in %q", kind, namespace), err)
 		}
+		// Skip ghosts (Terminating or NodeLost): a node replaced or lost
+		// mid-run leaves its DaemonSet/workload pod behind for minutes until
+		// pod GC, while a healthy replacement already runs on the live node.
+		// A negative (error) assertion must not fire on such an orphan — that
+		// is the orphan-pod trap documented in the project anti-patterns, and
+		// it turned a healthy ebs-csi-node DaemonSet into a false failure.
+		if isTerminatingOrLost(actual) {
```
mchmarny, line 456: nit: the skip is kind-agnostic — any listed resource with a `deletionTimestamp` is exempt from every negative assertion, so a pod stuck terminating indefinitely (wedged finalizer) that genuinely matches a forbidden shape will now never fail the check. Acceptable trade-off given pod GC normally clears these in minutes and the false-positive direction was actively burning UAT runs; just flagging that this is deliberately fail-open for terminating resources in case a future check wants to assert on terminating state itself.

mchmarny (approved):
Correct fix for the documented orphan-pod trap: `evaluateError`'s list-and-match loop had no filter for Terminating/NodeLost ghosts, so a DaemonSet pod left behind by node churn failed an otherwise-healthy check for the full assert timeout. The filter is scoped to negative assertions only (positive `assert` unchanged — the safe direction), and the `describePodStatus` suffix makes future forbidden-shape failures self-diagnosing. Test coverage is thorough, including the load-bearing "live pod beside a ghost still fails" case. One non-blocking note inline on the kind-agnostic breadth of the skip. CI green — no blockers.

PR #1656, opened 2026-07-08
feat(bundle): multi-tenant Argo deploy options

### LINKED ISSUE #1625: feat(bundle): support multi-tenant Argo deploy options
## Summary

Extend AICR Argo bundle generation so generated Argo Application fields can be parameterized for multi-tenant and remote-cluster deployments.

Today `--dynamic` supports component Helm values, but not the generated Argo Application wrapper fields. This blocks usage where the same AICR bundle is installed many times across namespaces, tenants, and clusters.

## Problem

The generated `argocd-helm` Argo resources currently have fields that are static or only partially configurable:

- The `spec.destination` blocks are not parameterized, so the bundle cannot be installed cleanly into different target clusters or namespaces at render time.
- Generated child Application names are hardcoded to component app names such as `gpu-operator` and related component names these names will collide in the shared Argo CD namespace.

The existing `--app-name` support helps with the parent Application name, but does not cover child app names, destination, or other generated Argo Application fields.

## Proposed Solution

Allow deployer-level Argo Application options to be supplied dynamically for `--deployer argocd-helm`, and possibly `--deployer argocd` where applicable.

It should allow installers to set:

- Parent Application name
- Child Application name prefix or naming template
- Generated Application destination
- Any related Argo Application deploy options needed for multi-tenant installs

### LINKED ISSUE #1628: feat(bundle): parameterize Argo resource finalizer
## Summary

Allow generated Argo Applications to optionally include `resources-finalizer.argocd.argoproj.io`.

## Problem

The generated `argocd-helm` parent Application does not include Argo CD's resource finalizer, so deleting it may leave managed child Applications or resources behind.

## Proposal

Add an install-time option for Argo deployers to enable the finalizer while preserving the current default behavior.

Docs:
- https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/#best-practices-for-app-of-apps-pattern
- https://argo-cd.readthedocs.io/en/stable/operator-manual/cluster-bootstrapping/#cascading-deletion

mchmarny:
## Summary

Adds a reserved `--set deployer:<key>=<value>` vocabulary (`namePrefix`, `destinationServer`, `project`, `cascadeDelete`) that parameterizes generated Argo CD Applications for multi-tenant and remote-cluster installs — bundle-time for both `argocd` and `argocd-helm` deployers, plus install-time `helm --set deployer.<key>=...` overrides for `argocd-helm`.

## Motivation / Context

Generated Argo Application wrapper fields (child names, `spec.destination.server`, `spec.project`) were hardcoded, blocking installs of the same bundle across tenants, namespaces, and clusters; deleting the parent Application also left managed resources behind (no finalizer option). Design follows the consensus posted on both issues.

Fixes: #1625
Fixes: #1628
Related: N/A

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- **Allowlisted, fail-closed key set** (`pkg/bundler/config/argo_deployer.go`): unknown `deployer:` keys, invalid values, non-Argo deployers, and the typed override path (`--set-json`/`--set-file`) are all rejected with `ErrCodeInvalidRequest`. The `deployer` name is reserved by a registry guard test.
- **Children-only scoping** for `namePrefix`/`destinationServer`/`project`: the parent app-of-apps stays on the Argo CD control-plane cluster in project `default` — Application CRs are reconciled only from the cluster running Argo CD ([apps-in-any-namespace](https://argo-cd.readthedocs.io/en/stable/operator-manual/app-any-namespace/) broadens namespaces, not source clusters), and [App-of-Apps is an admin-only pattern](https://argo-cd.readthedocs.io/en/stable/operator-manual/cluster-bootstrapping/). Parent name remains covered by `--app-name`.
- **`cascadeDelete`** adds `resources-finalizer.argocd.argoproj.io` to parent AND children ([app deletion docs](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/)); bundle-time only, because `metadata.finalizers` is a list field that cannot round-trip as a Helm template expression.
- **`argocd-helm` install-time layering**: baked values ship as chart defaults in the root `values.yaml` `deployer:` map; child templates read `.Values.deployer.*` with nil-safe sprig defaults. A live `helm template` test proves `--set deployer.namePrefix=t-` renders `t-gpu-operator`.
- **No behavior change without the options**: default output is semantically equivalent for both deployers — the review-hardening commits quote `name`/`project`/`server`/`namespace` in the `argocd` templates (goldens re-pinned), so the YAML bytes differ by quoting only; parsed values are unchanged. `argocd-helm` default rendering is equivalent (fields templated with today's values as defaults).
- Composed child names are validated as DNS-1123 subdomains at generation; `destinationServer` must be a credential-free `https://` URL.

## Testing

```bash
unset GITLAB_TOKEN
make qualify          # full gate: test (-race) + lint + e2e + scan — PASS
make test-coverage    # project floor — PASS
```

Coverage deltas (branch vs origin/main):
- `pkg/bundler`: 72.5% → 72.9% (+0.4%)
- `pkg/bundler/config`: 96.7% → 96.9% (+0.2%)
- `pkg/bundler/deployer/argocd`: 81.2% → 82.9% (+1.7%)
- `pkg/bundler/deployer/argocdhelm`: 81.8% → 81.8% (±0.0%)

New tests: allowlist parser table tests, generator option/rejection tests for both deployers, typed-override rejection, registry reserved-key tripwire, live `helm template` override render. `argocd` goldens unchanged; `argocd-helm` goldens regenerated (changes confined to the three templated child fields + `values.yaml` `deployer:` block).

## Risk Assessment

**Rollout notes:** Backwards compatible — bundles generated without `deployer:` options are semantically unchanged (scalar quoting differs in `argocd` output; parsed manifests identical). No migration needed.

## Checklist


pkg/bundler/config/argo_deployer.go
```diff
@@ -0,0 +1,116 @@
...
+			if err != nil {
+				return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+					fmt.Sprintf("invalid deployer cascadeDelete value %q: must be a boolean", value), err)
+			}
+			opts.CascadeDelete = parsed
+		default:
+			return nil, errors.New(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("unknown deployer option %q: valid keys are cascadeDelete, destinationServer, namePrefix, project", key))
+		}
+	}
```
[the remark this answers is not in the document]
mchmarny: Applied in 6e75d979 — keys are now iterated in sorted order, with a test pinning the deterministic first-error key.

pkg/bundler/config/argo_deployer.go
```diff
@@ -0,0 +1,116 @@
...
+			if err := ValidateHTTPSURL("deployer destinationServer", value); err != nil {
+				return nil, err
+			}
+			opts.DestinationServer = value
+		case deployerKeyProject:
+			if errs := validation.IsDNS1123Subdomain(value); len(errs) > 0 {
+				return nil, errors.New(errors.ErrCodeInvalidRequest,
+					fmt.Sprintf("invalid deployer project %q: must be a DNS-1123 subdomain", value))
+			}
+			opts.Project = value
```
mchmarny, line 183: Applied in 6e75d979 — the project error now embeds the concrete DNS-1123 violation (`errs[0]`).

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -701,6 +793,16 @@ func (g *Generator) processFolders(ctx context.Context, tmpDir, outputDir, templ
...
+		// Validate the composed child name at bundle time: the rendered
+		// template prepends `.Values.deployer.namePrefix` (defaulting to
+		// the baked NamePrefix in root values.yaml), so a prefix that
+		// composes into an invalid DNS-1123 name must fail here rather
+		// than at apiserver admission.
+		if nameErr := bundlercfg.ValidateAppName(g.NamePrefix + folderComponent); nameErr != nil {
+			return errors.Wrap(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("deployer namePrefix produces invalid child Application name %q", g.NamePrefix+folderComponent), nameErr)
+		}
+
```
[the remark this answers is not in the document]
mchmarny: Applied in 6e75d979 — both `Generate` boundaries now call the new `config.ValidateNamePrefix` upfront, covering zero-folder recipes; test `TestGenerate_NamePrefixValidatedUpfront` added.

mchmarny:
**Fixed:**
- **Install-time fail-open** — argocd-helm bundles now ship a `values.schema.json` scoped to the `deployer` key (`additionalProperties: false` + type/pattern validation), so `helm install --set deployer.destinationSever=...` fails loudly instead of silently targeting the in-cluster default. Root-level properties stay open for component override maps.
- **Helm 53-char release-name cap** — composed child names are rejected at generation in both deployers, and argocd-helm child templates carry a `fail`-based guard for install-time `deployer.namePrefix` overrides ([Argo CD defaults the Helm release name to the Application name](https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-release-name)).
- **Parent/child name collision** (`--app-name tenant-cert-manager` + `namePrefix=tenant-`) — rejected at generation in both deployers, plus an install-time template guard mirroring the parent's `appName` fallback.
- **YAML type/quoting safety** — `name`/`project`/`destination.server` are now `%q`-quoted in the argocd templates (`project=true` renders as the string `"true"`, not a boolean), and `destinationServer` rejects quotes/whitespace/control characters at parse time and in the schema pattern (a validator-accepted URL with an apostrophe previously broke the single-quoted Helm expression).
- **Reserved `deployer` key** — now enforced at registry load (single choke point covering embedded and external `--data` registries), not just by the tripwire test; a cross-package test pins the constant equality.
- **Docs** — remote-destination prerequisites (cluster must be [registered with Argo CD](https://argo-cd.readthedocs.io/en/stable/operator-manual/declarative-setup/#clusters), [AppProject](https://argo-cd.readthedocs.io/en/stable/user-guide/projects/) must permit the emitted destinations/sources) and the new name-length/collision limits are documented in the CLI reference.

**Not changed — destination-namespace support:** the consensus on #1625 explicitly deferred namespace remapping pending a concrete variant (per-tenant prefix vs. single shared namespace) — `destinationServer` + `namePrefix` cover the remote-cluster and shared-Argo-namespace cases raised there, and remapping can break namespace-coupled chart values. Opened #1658 to track it as a follow-up; this PR still closes #1625 per that consensus.

Full gate (`make qualify`) and coverage floor pass on the updated branch.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -463,6 +561,90 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
+				Type:                 "object",
+				AdditionalProperties: false,
+				Properties: valuesSchemaDeployerProps{
+					NamePrefix: valuesSchemaProperty{
+						Type:    schemaTypeString,
+						Pattern: `^$|^[a-z0-9][a-z0-9-]*$`,
+					},
+					DestinationServer: valuesSchemaProperty{
+						Type:    schemaTypeString,
+						Pattern: `^https://[^'"\s]+$`,
```
njhensley: **🟡 Minor — install-time `destinationServer` schema accepts embedded credentials the bundle-time validator rejects**

Bundle-time `ValidateDestinationServer` → `ValidateHTTPSURL` rejects userinfo (`u.User != nil`), and `argo_deployer_test.go` pins that `https://u:p@host:6443` is rejected. But `helm install --set deployer.destinationServer=...` is gated ONLY by this schema regex `^https://[^'"\s]+$`, which matches `https://user:pass@host:6443` — the Go validator never runs at install time.

**Blast radius:** Low — the schema still bars quotes/whitespace so the value renders safely into the single-quoted scalar (no YAML breakout); Argo CD matches `destination.server` by exact string against registered clusters, so a credentialed URL just fails to sync. The issue is fail-open drift in a schema whose stated purpose is to "fail loudly instead of silently."

**Fix:** Forbid an authority `@`, e.g. `^https://[^'"\s@]+$` — or deliberately document the divergence in the schema comment. Better still, generate the install-time schema from the same source as the Go validators so the two can't drift (see the `project` finding below — same root cause).

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -463,6 +561,90 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
+					},
+					Project: valuesSchemaProperty{
+						Type:      schemaTypeString,
+						Pattern:   `^[a-z0-9]([-a-z0-9.]*[a-z0-9])?$`,
```
njhensley: **🟡 Minor — install-time `project` schema is looser than `IsDNS1123Subdomain`**

Bundle-time `project` is validated with `validation.IsDNS1123Subdomain` (forbids consecutive dots, caps each label at 63). The install-time schema `^[a-z0-9]([-a-z0-9.]*[a-z0-9])?$` (maxLength 253, no per-label cap) accepts `a..b` and a 70-char single label — values `IsDNS1123Subdomain` rejects. Again schema-only at install time.

**Blast radius:** Low — `--set deployer.project=a..b` renders a child Application referencing an AppProject name that can never exist, so the child silently fails to sync rather than erroring at install.

**Fix:** Tighten to per-label form, e.g. `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`. F1+F2 share one root cause: the install-time `values.schema.json` is a hand-rolled weaker mirror of the Go allowlist — generating one from the other kills both (and the cascadeDelete nit below) at the root.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -463,6 +561,90 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
+	if err != nil {
+		return "", 0, err
+	}
+	schema := valuesSchema{
+		Schema: "http://json-schema.org/draft-07/schema#",
+		Type:   "object",
+		Properties: valuesSchemaRootProps{
+			Deployer: valuesSchemaDeployer{
+				Type:                 "object",
+				AdditionalProperties: false,
```
njhensley, line 631: **🔵 Nitpick — `helm --set deployer.cascadeDelete=true` fails with an opaque `additionalProperties` error**

`cascadeDelete` is intentionally bundle-time-only (finalizers is a list field), and the schema sets `additionalProperties: false` listing only the three string keys. An operator who tries the install-time form gets a generic "additional property … not allowed" rather than a targeted message.

**Blast radius:** Confusing UX for the exact operator the feature targets (3 of 4 keys are install-time overridable, 1 isn't).

**Fix:** Add a note in the values.yaml header ("cascadeDelete is bundle-time only; use `aicr bundle --set deployer:cascadeDelete=true`"), or surface a clearer signal.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -800,6 +1032,65 @@ func transformApplication(srcDir, templatesDir, folderName, componentName, overr
...
+// child Application template. The bundle-time checks in processFolders
+// cover baked NamePrefix values, but install-time `--set
+// deployer.namePrefix=...` bypasses them — these guards fail `helm
+// template`/`helm install` when the composed child name exceeds Helm's
+// release-name cap (Argo CD uses the Application name as the Helm release
+// name) or collides with the parent Application's name. The guard lines
+// are Helm template control flow that renders to nothing; Helm template
+// files need not be valid YAML pre-render. The appName fallback MUST
+// mirror parentAppTemplate's `.Values.appName | default "aicr-stack"`.
+func childNameGuard(componentName string) string {
```
njhensley, line 1088: **🔵 Nitpick — 53-char Helm-release-name cap is enforced for non-Helm children too**

The `childNameGuard` (and the bundle-time check) fires for every child regardless of `FolderKind`, but only Helm-source children derive a release name from the Application name; kustomize/manifest-only children (e.g. `nodewright-customizations`) are bounded only by DNS-1123's 253.

**Blast radius:** Over-strict and the message ("Argo CD uses the Application name as the Helm release name") is inaccurate for those children. Fail-closed, so cosmetic.

**Fix:** Gate the 53-cap on Helm folders, or soften the message.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -800,6 +1032,65 @@ func transformApplication(srcDir, templatesDir, folderName, componentName, overr
...
+
+	spec, ok := app["spec"].(map[string]any)
+	if !ok {
+		return errors.New(errors.ErrCodeInternal, "application manifest missing 'spec'")
+	}
+	spec["project"] = &yaml.Node{
+		Kind:  yaml.ScalarNode,
+		Tag:   yamlStringTag,
+		Style: yaml.SingleQuotedStyle,
+		Value: `{{ (.Values.deployer | default dict).project | default "default" }}`,
```
njhensley: **🔵 Nitpick — default literals duplicated instead of referencing the constants**

`applyDeployerTemplates` hardcodes `default "default"` (L1079) and `default "https://kubernetes.default.svc"` (L1089), duplicating `argocd.DefaultProject` / `argocd.DefaultDestinationServer` that `deployerValues()` uses via the constants.

**Blast radius:** If either constant changes, these template fallbacks drift silently. Low risk today since `deployerValues()` always bakes a non-empty value.

**Fix:** Interpolate `argocd.DefaultProject` / `argocd.DefaultDestinationServer` into the template strings via `fmt.Sprintf`.

njhensley (dismissed):
📋 **Multi-persona review — PR #1656 · feat(bundle): multi-tenant Argo deploy options**

**Method:** 4 persona passes (Correctness · Security · Domain/Architecture · Test-coverage/Operability), each finding independently confirmed or refuted by a senior meta-reviewer against the resolved code. Line links pinned to head `6e75d979`.

This is a high-quality, thoroughly-tested PR. The security-sensitive surface — user values rendered into Argo `Application` YAML — is handled well: the bundle-time path validates through an allowlist (`ParseArgoDeployerOptions`) **and** double-escapes with `printf "%q"`; the argocd-helm path emits single-quoted scalars with a schema that bans quotes/whitespace. **No injection, no blockers, no majors.** The finalizer-on-both-parent-and-children cascade model, the "children-only / parent-stays-in-cluster-default" app-of-apps design, the reserved-key layering seam with its cross-package equality test, and determinism (struct-ordered `values.schema.json` with a byte-golden) are all correct.

The one theme worth a follow-up: the argocd-helm **install-time `values.schema.json` is a hand-rolled, weaker mirror of the bundle-time Go validators**, so a value rejected at `aicr bundle` can slip through at `helm install`. Both instances (inline) are low blast-radius — bad value → child Application fails to sync, no crash, no breakout — hence Minor, but they're real drift sharing one root cause.

## Summary-only notes (land on unchanged context / non-code)

- 🔵 **CI helm fragility** (`.github/actions/go-test/action.yml`): the install-time guard/schema tests (`TestHelmTemplate_*`) `t.Skip` when helm is absent, and the go-test action installs only Go + envtest. They run today only because `ubuntu-latest` ships helm ambiently (unpinned) — unlike the repo's `setup-build-tools` which pins a version. Not a live gap, but a pinned `azure/setup-helm` step (or a CI env that makes absence fail rather than skip) would keep the safety net from silently disappearing.
- 🔵 **PR-description wording**: "argocd deployer default output is byte-identical" — the goldens now quote scalars (`name: "gpu-operator"`, `project: "default"`, `server: "https://…"`), so it's *semantically equivalent*, not byte-identical. Worth softening the claim (the `%q` quoting is also what makes the non-helm path injection-safe, so it's a good change).
- 🔵 **Docs**: consider a one-liner that (a) `deployer:destinationServer` supports only `spec.destination.server`, not `destination.name` (name-registered clusters won't resolve), and (b) `cascadeDelete=true` bakes the `resources-finalizer`, which blocks `kubectl delete application` if Argo CD is already gone (the classic stuck-`Terminating` trap).

- **YAML / Helm-template injection** — clean on both paths. Bundle-time renders via `printf "%q"` on pre-validated inputs; install-time uses single-quoted `yaml.Node` scalars with schema-banned quotes/whitespace; `{{…}}` in a `.Values` lookup is emitted as inert text (Helm is single-pass). The `childNameGuard` `printf` interpolates only registry-controlled `componentName` (`%q`) + a constant.
- **Finalizer `strings.Replace` injection** — anchor `"  namespace: argocd\nspec:"` occurs exactly once in `parentAppTemplate`; the `!Contains(ResourcesFinalizer)` fail-loud guard is sound.
- **Length/collision math** — Go `> 53` and guard `gt (len …) 53` agree (no off-by-one); the `appName | default "aicr-stack"` fallback matches across guard, parent template, and the Go collision check.
- **Cross-deployer parity** — the four options render equivalently across `argocd` and `argocd-helm`; cascadeDelete is bundle-time-only in both.
- **Determinism** — `values.schema.json` is struct-ordered and pinned by a byte-golden (`TestBundleGolden_HelmAndManifestOnly`).
- **Fail-closed integrity** — `argoDeployerOptions()` runs unconditionally before the deployer switch; rejects typed (`--set-json/--set-file`) overrides, unknown keys, and the `deployer:` prefix on non-Argo deployers; registry guard fails closed for embedded + `--data` registries; cross-package constant equality pinned.
- **"multi-tenant" scope** — correctly delivers per-tenant *naming/placement*, not AppProject RBAC *isolation* (honestly documented in the `Project` field comment).

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major | 0 | — |
| 🟡 Minor | 2 | install-time schema accepts credentialed `destinationServer`; install-time `project` schema looser than DNS-1123 |
| 🔵 Nitpick | 5 | opaque `cascadeDelete` install error · 53-cap on non-Helm children · duplicated default literals · CI helm fragility · "byte-identical" wording (+ optional doc notes) |

**Recommendation: Approve with comments.** No blockers or majors. The two Minors share one root cause — the install-time `values.schema.json` diverging from the Go validators — and are worth a follow-up (ideally generating one from the other); everything else is optional polish.

<sub>🤖 Generated with multi-persona review + adversarial meta-review.</sub>

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: Fixed in 24d4784c — pattern now forbids `@` (`^https://[^'"\s@]+$`), matching the bundle-time rejection of userinfo; unit test pins `https://u:p@host:6443` against the compiled pattern. Agreed on the shared-source idea — kept as hand-written for now since the pattern tests now pin schema↔validator agreement.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: Fixed in 24d4784c — project pattern is now the exact per-label DNS-1123 form with `{0,61}` label caps, rejecting `a..b` and >63-char labels; pinned by the same schema-pattern unit test.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: Fixed in 24d4784c — the values.yaml header now states cascadeDelete is bundle-time only (`aicr bundle --set deployer:cascadeDelete=true`) and that the install-time form fails schema validation; the schema doc comment notes the deliberate omission.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: Addressed in 24d4784c via your option B — the cap stays uniform (single invariant, no folder-kind branching in the guard), and the messages in both bundle-time checks and the template guard now say the release-name derivation applies to Helm-rendered children, with a code comment explaining the deliberate uniform application.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: Fixed in 24d4784c — both template fallbacks are now built via `fmt.Sprintf` from `argocd.DefaultProject` / `argocd.DefaultDestinationServer`, so a constant change can no longer drift silently.

pkg/bundler/deployer/argocd/templates/application.yaml.tmpl
```diff
@@ -33,7 +37,7 @@ spec:
       ref: values
 {{- end }}
   destination:
-    server: https://kubernetes.default.svc
+    server: {{ printf "%q" .DestinationServer }}
     namespace: {{ .Namespace }}
```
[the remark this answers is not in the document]
mchmarny: Applied in 5030c283 — `destination.namespace` now uses the same `printf "%q"` quoting as name/project/server; goldens regenerated (quotes only).

pkg/bundler/config/argo_deployer.go
```diff
@@ -0,0 +1,161 @@
...
+			if err := ValidateDestinationServer(value); err != nil {
+				return nil, err
+			}
+			opts.DestinationServer = value
+		case deployerKeyProject:
+			if errs := validation.IsDNS1123Subdomain(value); len(errs) > 0 {
+				return nil, errors.New(errors.ErrCodeInvalidRequest,
+					fmt.Sprintf("invalid deployer project %q: %s", value, errs[0]))
+			}
+			opts.Project = value
```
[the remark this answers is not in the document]
mchmarny: Skipping this one — the case is unreachable through every supported input path: `ComponentPath.Parse` rejects `--set deployer:project=` with "value cannot be empty" (pkg/bundler/config/component_path.go) before `ParseArgoDeployerOptions` ever runs, and the same parse guards the API's `?set=` parameter. Treating `""` as unset inside the parser would only affect direct library callers passing a hand-built map, where silently coercing an explicit empty to "default" is less safe than rejecting it (fail-closed). `Generator.project()`'s empty fallback exists for the no-override-at-all case, not for explicit empties.

pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
```diff
@@ -1901,3 +1905,349 @@ func TestWriteChartYAML_QuotesYAMLReservedScalarsAsName(t *testing.T) {
...
+			re, err := regexp.Compile(tt.pattern)
+			if err != nil {
+				t.Fatalf("pattern %q does not compile: %v", tt.pattern, err)
+			}
+			if got := re.MatchString(tt.input); got != tt.match {
+				t.Errorf("pattern %q match(%q) = %v, want %v", tt.pattern, tt.input, got, tt.match)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
mchmarny: Applied in ba4f8fd3, with one strengthening: the struct-based assertion alone cannot detect the omitempty-drop scenario (unmarshal defaults the field to false either way), so the test also asserts the raw JSON bytes contain `"additionalProperties": false` explicitly — that is the check that actually catches the fail-open regression.

njhensley (dismissed):
✅ **Re-approve on `ba4f8fd3`** — all review feedback addressed.

Verified the follow-up commits against my earlier findings:

- 🟡 **`destinationServer` credential parity** — schema pattern tightened to `^https://[^'"\s@]+$`, matching the bundle-time `ValidateHTTPSURL` credential rejection. Pinned by `TestValuesSchemaPatterns` (`https://u:p@host:6443` now rejected).
- 🟡 **`project` DNS-1123 parity** — schema now enforces exact per-label form (`≤63`/label, no empty labels); `a..b` is rejected, matching `IsDNS1123Subdomain`.
- 🔵 **`cascadeDelete` UX** — documented as bundle-time-only in the schema comment and the `values.yaml` install header.
- 🔵 **53-char cap message** — reworded to be accurate for non-Helm children (uniform cap kept by design as a single invariant).
- 🔵 **Duplicated defaults** — now reference `argocd.DefaultProject` / `argocd.DefaultDestinationServer`.

`values.schema.json` golden regenerated to match, and a new fail-open guard test asserts `additionalProperties: false` is emitted. No blockers, no majors. LGTM.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
yuanchen8911: Follow-up after 24d4784c — embedded credentials are now rejected, but three gaps remain and the fix introduced one false positive (all reproduced with `helm template` against a chart generated at 5030c283):

- `https:///nohost` and `https://:6443` still pass the schema and render into the child Applications' `spec.destination.server`. `https://:6443` is also accepted at bundle time: `url.Parse` returns `Host=":6443"`, which passes the `u.Host == ""` check in ValidateHTTPSURL even though `Hostname()` is empty — switching that check to `u.Hostname()` closes the bundle-time half.
- An explicitly-empty value (`--set deployer.destinationServer=""`) is still rejected by the schema, while bundle-time treats empty as "use the in-cluster default", so a remote-destination bundle cannot be reset to in-cluster at install time. Adding the `^$|` alternative (as the namePrefix pattern has) fixes this; the template's `default` fallback already handles empty.
- The new blanket `@` exclusion rejects legal URLs with `@` in the path or query (e.g. `https://host/a@b`), which bundle-time accepts — `url.Parse` only treats `@` before the host as userinfo.

A regex cannot reproduce net/url semantics, so suggest a render-time guard in the child template prologue instead: sprig `urlParse`, failing when scheme is not `https`, hostname is empty, or userinfo is non-empty. It must check `hostname`, not `host` — for `https://:6443`, `host` is `":6443"` (non-empty) while `hostname` is empty, the same trap as `u.Host` in Go.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
yuanchen8911: Follow-up after 24d4784c — `a..b` is now rejected, but the `{0,61}` per-label cap introduces a divergence in the opposite direction: `validation.IsDNS1123Subdomain` enforces only the 253-char total cap and the subdomain regex — it has no per-label 63-char limit — so a 64+-char label (e.g. a 70-char single-label project name) is accepted at bundle time but now rejected at install time. TestValuesSchemaPatterns codifies the over-restriction ("project rejects 70-char label") and compiles the patterns with Go regexp only, without consulting the Go validators, so it cannot catch this class of drift.

Suggest dropping the per-label bound and mirroring the Kubernetes regex exactly — `^$|^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` with `maxLength: 253` — and having the test assert each input against both `validation.IsDNS1123Subdomain` and the compiled pattern with an explicit `wantValid` per case, so both layers are checked against the contract rather than against each other.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -800,6 +1046,65 @@ func transformApplication(srcDir, templatesDir, folderName, componentName, overr
...
+// release-name cap (Argo CD uses the Application name as the Helm release
+// name) or collides with the parent Application's name. The guard lines
+// are Helm template control flow that renders to nothing; Helm template
+// files need not be valid YAML pre-render. The appName fallback MUST
+// mirror parentAppTemplate's `.Values.appName | default "aicr-stack"`.
+func childNameGuard(componentName string) string {
+	return fmt.Sprintf(
+		`{{- $childName := printf "%%s%%s" ((.Values.deployer | default dict).namePrefix | default "") %q -}}
+{{- if gt (len $childName) %d }}{{ fail (printf "deployer.namePrefix produces child Application name %%q (%%d chars): child Application names are capped at %d characters because Argo CD derives the Helm release name from the Application name for Helm-rendered children" $childName (len $childName)) }}{{ end -}}
+{{- if eq $childName (.Values.appName | default %q) }}{{ fail (printf "child Application name %%q collides with the parent Application name: choose a different deployer.namePrefix or appName" $childName) }}{{ end -}}
```
yuanchen8911: `helm template` crashes when appName is a scalar-looking value: `--set appName=true` (bool) or `--set appName=123` (int64) makes this `eq` fail with "incompatible types for comparison". On the base commit both values render fine because the parent template pipes appName through `quote`, so this is a regression of an existing install-time input — and appName is deliberately outside values.schema.json, so nothing catches it earlier.

Suggested fix: coerce before comparing — `eq $childName (.Values.appName | default "aicr-stack" | toString)`.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -263,6 +313,13 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 		// parent chart level via writeStaticValuesAndBuildStubs.
 		AllowDynamicValueSplit: true,
 		VendorCharts:           g.VendorCharts,
+		// Forward ONLY CascadeDelete: child finalizers survive the YAML
+		// round-trip in transformApplication untouched. NamePrefix /
+		// DestinationServer / Project are NOT forwarded — those fields
+		// are rewritten into `.Values.deployer.*` template expressions
+		// during transform, so forwarding would bake values the rewrite
+		// then discards.
+		CascadeDelete: g.CascadeDelete,
```
yuanchen8911, line 333: The inner argocd.Generator is constructed without AppName, so the parent-collision check inside the delegated Generate compares child names against argocd.DefaultAppName ("nvidia-stack") rather than this deployer's effective parent ("aicr-stack" or `--app-name`). A component named `nvidia-stack` — possible via an external `--data` registry — is falsely rejected, wrapped as ErrCodeInternal "failed to generate base Argo CD output" with error text citing `--app-name`. The outer processFolders check already re-validates against the correct parent name, so the inner check only adds this false-positive path.

Suggested fix: forward the effective parent name (g.AppName, defaulting to this package's DefaultAppName) to the inner Generator; argocd-helm discards the inner app-of-apps.yaml, so nothing else changes.

pkg/recipe/components.go
```diff
@@ -350,6 +358,23 @@ func loadComponentRegistryFor(provider DataProvider) (*ComponentRegistry, error)
...
 	}
 
+	// Fail closed on the reserved deployer key for EVERY loaded registry
+	// (embedded and external --data alike): a component named "deployer"
+	// or aliasing it via valueOverrideKeys would make `--set deployer:...`
+	// ambiguous between component Helm values and deployer-level Argo
+	// Application options. See #1625.
+	for i := range registry.Components {
+		comp := &registry.Components[i]
+		if comp.Name == ReservedDeployerKey {
```
yuanchen8911, line 368: This guard covers registry loads (embedded and `--data`), but not recipe input: a recipe passed to `aicr bundle -r` can declare a component named `deployer` with an explicit type — coherence validation never inspects ref.Name and registry backfill is skipped for explicit types (reproduced through the CLI). For such a recipe, bundle-time `--set deployer:<key>` becomes ambiguous or unusable across deployers: with `--deployer argocd` an allowlisted key like namePrefix applies both as an Argo option and as a chart value merged into that component, and any other key fails as "unknown deployer option"; with `--deployer helm`/`flux`/`helmfile` every `--set deployer:*` is rejected outright, so that component's chart values cannot be set via `--set` at all.

Suggested fix: reject `ref.Name == "deployer"` (mirroring this error message) at the common RecipeResult validation boundary, including disabled refs, so recipe input fails closed the same way registry input does.

pkg/bundler/deployer/argocd/argocd.go
```diff
@@ -353,6 +434,15 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 	if err := bundlercfg.ValidateAppName(g.AppName); err != nil {
 		return nil, err
 	}
+	if err := bundlercfg.ValidateNamePrefix(g.NamePrefix); err != nil {
+		return nil, err
+	}
+	if err := bundlercfg.ValidateDestinationServer(g.DestinationServer); err != nil {
+		return nil, err
+	}
+	if err := bundlercfg.ValidateAppName(g.Project); err != nil {
```
yuanchen8911: Project is validated here with ValidateAppName, so a direct library caller with an invalid Project gets "invalid app name …" pointing at the wrong field (same at argocdhelm.go:277). CLI/API callers are unaffected — ParseArgoDeployerOptions validates project first with a field-specific message. The sibling checks (ValidateNamePrefix, ValidateDestinationServer) have purpose-built messages; suggest a ValidateProject twin (same accept set: empty allowed, DNS-1123 subdomain) with a project-specific message.

docs/user/cli-reference.md
```diff
@@ -1462,6 +1462,44 @@ aicr bundle -r recipe.yaml \
...
+#### Argo CD Deployer Options
+
+The `deployer` prefix is reserved: with `--deployer argocd` or `--deployer argocd-helm`, `--set deployer:<key>=<value>` configures the generated Argo CD Applications instead of component chart values. Unknown `deployer:` keys are rejected, and the prefix is rejected entirely with any other `--deployer` type (`helm`, `flux`, `helmfile`).
+
+| Key | Applies to | Default | Example |
+|-----|------------|---------|---------|
+| `namePrefix` | Child Application names | (none) | `--set deployer:namePrefix=tenant-a-` |
+| `destinationServer` | Child Applications' `spec.destination.server` | `https://kubernetes.default.svc` | `--set deployer:destinationServer=https://prod.example.com:6443` |
+| `project` | Child Applications' `spec.project` | `default` | `--set deployer:project=gpu-infra` |
+| `cascadeDelete` | Parent and child Applications | `false` | `--set deployer:cascadeDelete=true` |
```
yuanchen8911, line 1474: With cascadeDelete shipping, the cleanup section later in this page ("AICR does **not** set the `resources-finalizer.argocd.argoproj.io` finalizer on generated Applications", around L2402 — outside this diff) is no longer unconditionally true: bundles generated with `--set deployer:cascadeDelete=true` carry exactly that finalizer on the parent and every child, and a plain `kubectl delete` on the parent then does cascade. Suggest qualifying that section in this PR: default bundles have no finalizer (use the cascade-aware flows documented there); cascadeDelete bundles do, with a link back to this options table.

docs/user/cli-reference.md
```diff
@@ -1462,6 +1462,44 @@ aicr bundle -r recipe.yaml \
...
+
+**Install-time overrides with argocd-helm:** for `--deployer argocd-helm`, the three string keys ship as defaults in the bundle chart's root `values.yaml` and can also be overridden at install time via `helm install --set deployer.<key>=...` (note the dot, not colon). The bundle ships a `values.schema.json` that Helm applies on install/upgrade/template/lint: unknown `deployer.*` keys (e.g. a `destinationSever` typo) and malformed values are rejected at install time instead of silently falling back to defaults. `cascadeDelete` is bundle-time only — it adds the [`resources-finalizer.argocd.argoproj.io` finalizer](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/) (a list field, not overridable via `--set`) so deleting an Application also deletes its deployed resources.
```
yuanchen8911, line 1482: The plain `helm --set` path documented here breaks for legal values Helm type-infers: `--set deployer.project=true` or `=123` (valid DNS-1123 project names, accepted at bundle time) fail schema validation with "got boolean/number, want string"; the same applies to `deployer.namePrefix`. `--set-string` works. Keeping the string-typed schema is right — suggest documenting `--set-string` as the required form for values Helm would parse as a non-string (booleans, bare numbers), with one example.

yuanchen8911:
Three notes that do not anchor to changed lines:

**CI does not install Helm for unit tests.** `.github/actions/go-test/action.yml` sets up Go and envtest only, and the two new Helm-dependent test functions (covering four scenarios) skip silently when `helm` is not on PATH. They currently run only because the runner image bundles Helm — 3.x, while `.settings.yaml` pins `v4.2.2` — so the install-time schema/guard behavior is validated against a different Helm major than the project pins, and would silently drop out of CI if the runner image changes. Suggest installing the pinned Helm in the go-test action and making these tests fail rather than skip when `CI` is set.

**Validation-contract drift.** The destinationServer/project/namePrefix findings on this PR are all instances of the install-time gate (schema pattern or template guard) diverging from the bundle-time Go validators — in both directions. A parity test would close the class: drive the same inputs through the Go validator and through `helm template` against a generated chart (`--set-string`), each case asserting an explicit `wantValid` — asserting the two paths merely agree with each other would miss shared bugs such as both accepting `https://:6443`. Include live plain `--set` cases for `true`/`false`/`123`/`0`/`null`, since Helm type inference is outside anything a regex can validate.

**PR body wording.** The "byte-identical" claim for the argocd deployer is not accurate — the twelve argocd goldens each changed four scalars from unquoted to quoted (name, project, server, namespace). The output is semantically equivalent, not byte-identical; suggest rewording to avoid overstating the compatibility guarantee.

mchmarny:
All three addressed, @yuanchen8911 — and the second one paid for itself immediately.

**CI Helm install** — `62f14aeb` (originally 687cf26a): `.github/actions/go-test` now installs the `.settings.yaml`-pinned Helm (v4.2.2, sha256-verified, replacing the runner's stock 3.x) via the same `load-versions` flow the other tool installs use, and the 8 Helm-dependent test sites now route through a shared `requireHelm(t)` that **fails** (not skips) when `CI` is set and helm is missing.

**Validation-contract parity test** — added as `TestValidationContractParity`: every case carries an explicit `wantValid` and is asserted against BOTH the Go validators and a live `helm template` run — exactly to catch shared bugs, and it caught two real ones on the first run:
- `https://:6443` was accepted by `ValidateHTTPSURL` (`u.Host` is `":6443"`, non-empty) — fixed by checking `u.Hostname()`, which also hardens the shared fulcio/rekor endpoint validation (IPv6 literals verified unaffected).
- `IsDNS1123Subdomain` doesn't enforce the per-label 63-char cap (only the 253 total), so a 64-char project label passed bundle-time but failed the schema — new `config.ValidateProject` closes it.

The type-inference cases are in too (`--set deployer.project=true/false/123/0` all fail the schema's `type: string`; `=null` deletes the key and correctly renders the baked default — pinned with a comment). A follow-up commit also closed the two residual drifts the parity review then surfaced (`HTTPS://` uppercase scheme, `@` in path) — the gate is now mirror-exact in both directions.

**PR body** — reworded: the argocd output is semantically equivalent, not byte-identical (the hardening commits quote name/project/server/namespace; goldens re-pinned).

pkg/bundler/deployer/argocdhelm/argocdhelm.go
yuanchen8911: Verified at 62f14aeb: `https://:6443` and `https:///nohost` are now rejected by both gates (schema first-character class + the `u.Hostname()` fix), and the `@` handling is aligned by narrowing the bundle-time validator to match the schema — a deliberate shared narrower grammar, which works: `https://host/a@b` now fails identically on both paths, so there is no drift left there.

One residual asymmetry remains, empirically confirmed: an explicit empty value is accepted at bundle time (`ValidateDestinationServer("")` returns nil, meaning "use the in-cluster default") but rejected by the install-time schema. 24d4784c/62f14aeb resolved exactly this for `project` by rejecting explicit empty at parse time — suggest the same one-line treatment for `destinationServer` in ParseArgoDeployerOptions so the two keys behave consistently and the parity test can carry an empty-string case with a single wantValid.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
yuanchen8911: Verified at 62f14aeb: parity is achieved — ValidateProject now enforces the per-label 63-character cap at bundle time, both generators call it, and a 70-char single-label value fails identically on both gates (reproduced against the Go validator and `helm template`). Going narrower than `IsDNS1123Subdomain` on both sides is a consistent, documented policy choice, and TestValidationContractParity's explicit wantValid per case is exactly the shape that keeps this from drifting again. Resolving.

pkg/bundler/deployer/argocd/argocd.go
yuanchen8911: Fixed at 62f14aeb: ValidateProject exists with a field-specific message and both generator boundaries call it instead of ValidateAppName. Resolving.

yuanchen8911:
Verified at 62f14aeb: the CI gap and the parity-test recommendation from the earlier note are both addressed — the go-test action now installs the Helm version pinned in .settings.yaml and the live-render tests fail (rather than skip) when helm is missing in CI, and TestValidationContractParity drives the bundle-time Go validators and `helm template` against explicit per-case wantValid, including plain `--set` type-inference cases. Remaining open items from this review: the appName `toString` guard crash, the delegated generator's parent name, the reserved `deployer` recipe-input boundary, and the two documentation notes (finalizer cleanup section, `--set-string`).

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: Fixed in 40a4e0ff — the guard coerces with `toString` before the `eq`, and a live test pins `--set appName=true` rendering successfully (parent name `"true"` via the existing `quote`).

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: Fixed in 40a4e0ff — the effective parent (`g.AppName` or `aicr-stack`) is now forwarded to the inner generator. Tests prove it load-bearing both ways: a component named `nvidia-stack` now bundles fine with the default parent, and a deliberate `--app-name gpu-operator` collision still fails.

pkg/recipe/components.go
mchmarny: Fixed in 40a4e0ff — `RecipeResult.PrepareAndValidate` now rejects componentRefs named `deployer` (before type backfill, disabled refs included) with the registry guard's message. That boundary is shared by CLI file load, the API adopt path, and defensively by `bundler.Make`, so both entry points fail closed.

docs/user/cli-reference.md
mchmarny: Fixed in 40a4e0ff — the cleanup section is now qualified: no finalizer on default bundles (existing cascade-aware flows apply), but `--set deployer:cascadeDelete=true` bundles carry it on parent and children so plain `kubectl delete` cascades; cross-linked to the deployer-options table.

docs/user/cli-reference.md
mchmarny: Fixed in 40a4e0ff — the options section now documents `--set-string` as the required form for values Helm would type-infer (booleans, bare numbers), with an example, and notes the schema is intentionally string-typed.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: Closing the loop on your follow-up (now spread across 5e78fc8d/62f14aeb/40a4e0ff): `https://:6443` and `https:///nohost` are rejected by BOTH gates (`u.Hostname()` bundle-time; schema forbids `:`/`/` as the first authority char); explicit-empty at install now resets to the in-cluster default (`^$|` alternative + template `default` fallback, live-tested); on `@` we went the other direction from the urlParse suggestion — both gates reject `@` anywhere, keeping the two gates mirror-exact (a regex cannot reproduce net/url authority parsing, and cluster API endpoints never carry `@` in paths). If a real endpoint with `@` ever shows up we can revisit with the render-time urlParse guard.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
mchmarny: You were right on this one — adopted in 40a4e0ff: a 64+-char label is a legal Kubernetes object name, so the per-label cap was a false positive with no workaround. Both gates now mirror `IsDNS1123Subdomain` exactly (`^$|` + no label cap, maxLength 253), `config.ValidateProject` dropped the cap, and the parity test asserts each input with explicit `wantValid` against both the Go validator and a live `helm template` — not against each other.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
yuanchen8911: Fixed at 40a4e0ff — verified: `--set appName=true` and `--set appName=123` now render (the `| toString` coercion), matching the base-commit behavior. Resolving.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
yuanchen8911: Fixed at 40a4e0ff — the inner generator now receives `cmp.Or(g.AppName, DefaultAppName)`, so the delegated collision check tests against this deployer's effective parent. Resolving.

pkg/recipe/components.go
yuanchen8911: Fixed at 40a4e0ff — PrepareAndValidate now rejects componentRefs named with the reserved deployer key before backfill/coherence, for all refs including disabled ones, with coverage in componentref_coherence_test.go. Resolving.

docs/user/cli-reference.md
yuanchen8911: Fixed at 40a4e0ff — the cleanup section now qualifies the default-vs-cascadeDelete behavior and links back to the options table. Resolving.

docs/user/cli-reference.md
yuanchen8911: Fixed at 40a4e0ff — the install-time section now documents `--set-string` for type-inferred values with an example. Resolving.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
yuanchen8911: Verified at 40a4e0ff: the `^$|` alternative landed, and an explicit-empty install-time value now renders the baked default (reproduced) — the last asymmetry on this key is closed. Bundle-parse explicit-empty remains distinct by design (nothing to reset at bundle time), which the code comments now document. Nothing further on this thread.

mchmarny:
Thanks for the verification, @yuanchen8911. The five remaining items from your list all landed in 40a4e0ff (each thread has an inline reply with details): the `toString` coercion in the child-name guard, the effective parent name forwarded to the delegated generator, the reserved `deployer` rejection at `RecipeResult.PrepareAndValidate` (covers hand-authored recipes on both the CLI and API paths, disabled refs included), and both doc corrections (finalizer cleanup section qualified + `--set-string` documented). All review threads are now resolved; `make qualify` passes on the head.

yuanchen8911 (approved):
All review findings verified fixed at 40a4e0ff — bundle-time and install-time validation confirmed aligned end-to-end (helm template matrix + Go validator probes), parity test and pinned-Helm CI in place, docs updated.

PR #1657, opened 2026-07-08
ci(uat-aws): refresh AWS creds per stage across the UAT phase chain

njhensley:
## Summary

Re-mint a fresh 1h AWS STS session at each long UAT-AWS phase boundary (before `install`, `validate`, and the CUJ) so no single phase runs on an expiring token. Workflow-only — no IAM/terraform change.

## Motivation / Context

The initial `Configure AWS credentials` step mints a 1-hour STS session (the action default). kubectl and `aicr validate` authenticate to EKS through the `aws eks get-token` exec plugin, which signs each token with the **ambient** session credentials. Build + EKS bringup + `install` (60m) + `validate` (40m) run well past 1h, so by the time `validate` runs the session has expired and every `get-token` fails with `ExpiredToken` mid-phase. This is distinct from the teardown-only expiry fixed earlier in #1456.

Fixes: N/A
Related: #1456

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: CI (`.github/workflows/uat-aws.yaml`)

## Implementation Notes

Three new `configure-aws-credentials` steps re-mint a fresh session before `install`, `validate`, and the intent-selected CUJ (train/serve), each with a distinct `role-session-name`. Each phase therefore starts inside a full 1h window (worst-case phase budget < 60m), so `get-token` never signs with an expired session.

Deliberately kept at the STS default duration: **no** `role-duration-seconds` and **no** change to the role's `max_session_duration` — so this is a pure workflow change requiring no `terraform apply` and no longer-lived credentials. The existing pre-teardown re-auth (#1456) stays as the final backstop; its comment was updated to match the 1h-per-stage model.

## Testing

```bash
yamllint .github/workflows/uat-aws.yaml   # clean
actionlint .github/workflows/uat-aws.yaml # only pre-existing SC2016 at an untouched step
```

No Go source changed, so `make test`/coverage gates are N/A. Validated end-to-end by the next nightly UAT-AWS run (`validate` no longer expiring mid-phase).

## Risk Assessment

- [x] **Low** — Isolated change, easy to revert

**Rollout notes:** None — workflow-only, effective on merge.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — internal CI)

### DIFF
--- a/.github/workflows/uat-aws.yaml
+++ b/.github/workflows/uat-aws.yaml
@@ -260,6 +260,15 @@ jobs:
           done
 
       # Auth
+      #
+      # Mints the action-default 1h STS session. kubectl and `aicr validate`
+      # reach EKS through the `aws eks get-token` exec plugin, which signs each
+      # token with these ambient env credentials — so once the session expires
+      # (build + bringup + install + validate can exceed 1h) those get-token
+      # calls fail with ExpiredToken. The per-stage refresh steps below re-mint a
+      # fresh 1h session at each long phase boundary so no single phase runs on
+      # an expiring token; the role's max_session_duration is left at the STS
+      # default, so we deliberately do NOT set role-duration-seconds here.
       - name: Configure AWS credentials
         id: auth
         uses: aws-actions/configure-aws-credentials@517a711dbcd0e402f90c77e7e2f81e849156e31d  # v6.2.2
@@ -455,6 +464,22 @@ jobs:
           AICR_IMAGE: ${{ inputs.aicr_version == '' && format('ghcr.io/nvidia/aicr-validators/agent:{0}', env.VALIDATOR_TAG) || format('ghcr.io/nvidia/aicr:{0}', inputs.aicr_version) }}
         run: ./tests/uat/aws/run prep "${TEST_CONFIG}"
 
+      # Per-stage credential refresh. Re-mint a fresh 1h session at each long
+      # phase boundary so no single phase runs on an expiring token — the UAT
+      # chain (build + bringup + install + validate + CUJ) outlives one STS
+      # session, and kubectl/`aicr validate` sign every EKS get-token with the
+      # ambient session credentials. `id-token: write` is granted for the whole
+      # job, so configure-aws-credentials can re-assume here. Keyed to prep
+      # success so it is skipped in the skip_tests / daytime-up-with-no-prep
+      # paths, matching the install step's own gate.
+      - name: Refresh AWS credentials before install
+        if: steps.prep.outcome == 'success'
+        uses: aws-actions/configure-aws-credentials@517a711dbcd0e402f90c77e7e2f81e849156e31d  # v6.2.2
+        with:
+          role-to-assume: "arn:aws:iam::${{ env.AWS_ACCOUNT_ID }}:role/${{ env.GITHUB_ACTIONS_ROLE_NAME }}"
+          aws-region: ${{ env.AWS_REGION }}
+          role-session-name: GitHubActions-UAT-AICR-Install
+
       - name: UAT - install (helmfile apply)
         id: install
         if: steps.prep.outcome == 'success'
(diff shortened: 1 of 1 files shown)


.github/workflows/uat-aws.yaml
```diff
@@ -455,6 +464,22 @@ jobs:
...
+      # phase boundary so no single phase runs on an expiring token — the UAT
+      # chain (build + bringup + install + validate + CUJ) outlives one STS
+      # session, and kubectl/`aicr validate` sign every EKS get-token with the
+      # ambient session credentials. `id-token: write` is granted for the whole
+      # job, so configure-aws-credentials can re-assume here. Keyed to prep
+      # success so it is skipped in the skip_tests / daytime-up-with-no-prep
+      # paths, matching the install step's own gate.
+      - name: Refresh AWS credentials before install
+        if: steps.prep.outcome == 'success'
+        uses: aws-actions/configure-aws-credentials@254c19bd240aabef8777f48595e9d2d7b972184b  # v6.2.1
```
mchmarny: The three new steps pin `254c19bd...` (v6.2.1), but #1653 bumped main to `517a711d...` (v6.2.2) — that's also the merge conflict blocking CI here. After rebasing, use the v6.2.2 SHA in all three new steps so the workflow doesn't carry two pins of the same action (renovate will otherwise churn on it). Same applies to the `before validate` and `before CUJ` steps.

.github/workflows/uat-aws.yaml
```diff
@@ -498,6 +531,16 @@ jobs:
       # DynamoGraphDeployment (phase_serve, DC3) — deploy, wait for readiness,
       # hit the OpenAI-compatible endpoint, assert a completion. Exactly one
       # fires per run; verify/evidence cover the deployed stack either way.
+      # One refresh before the intent-selected CUJ (train OR serve); both gate on
+      # conformance success, and exactly one runs per intent.
+      - name: Refresh AWS credentials before CUJ
```
mchmarny, line 539: Question, non-blocking: `UAT - verify (evidence verify)` runs after the CUJ with no refresh of its own. If train runs near its budget (~60m), verify starts on an expiring session minted here. If verify only checks signatures/OCI it doesn't matter; if it hits EKS via get-token, it's the same failure mode one step later. Worth confirming — if it does touch the cluster, either move this refresh to after the CUJ or add a fourth one.

mchmarny (commented):
The per-stage refresh approach is right, the gating conditions correctly mirror each phase's own gate, and keeping the STS default duration (no terraform change) is the right trade-off. Two things before this can merge: the branch has a merge conflict with main (#1653 bumped `configure-aws-credentials` to v6.2.2), and the three new steps should adopt the v6.2.2 pin on rebase — inline comment on the first occurrence. Also one non-blocking question inline about whether `verify` needs a refresh after a long CUJ. CI hasn't run on this head due to the conflict; rebase will trigger it.

.github/workflows/uat-aws.yaml
njhensley: Done. Rebased onto current main (resolved the #1653 conflict) and squashed to a single commit; all three new steps now use the v6.2.2 pin `517a711d…`, so the workflow carries exactly one pin of the action (no renovate churn).

.github/workflows/uat-aws.yaml
njhensley: Confirmed `verify` does not touch EKS, so no fourth refresh is needed. `phase_verify` (tests/uat/aws/run) runs only `aicr trust update` (Sigstore TUF root, network to the TUF mirror) and `aicr evidence verify ./evidence/pointer.yaml` (pulls the OCI evidence bundle from GHCR — authenticated by the earlier GHCR login, not AWS — and checks Fulcio/Rekor signatures). No kubectl, no `aws eks get-token`. The CUJ refresh preceding it is incidental; verify would pass on an expired AWS session.

mchmarny (approved):

Issue #1652, opened 2026-07-08
validator(inference-perf): generalized DRA topologies/drivers are fail-fast


yuanchen8911:
## Summary

`inference-perf` validates exactly the two GPU allocation configurations AICR ships (per #1327 / PR #1620):

1. Node-local (`spec.nodeName`) `gpu.nvidia.com` ResourceSlices on the worker node → DRA-claims worker wiring, sized from validated device counts.
2. ComputeDomain-only / no full-GPU slices → device-plugin worker wiring, sized from scalar `nvidia.com/gpu` allocatable.

Configurations outside that matrix **fail fast with an actionable message** rather than being modeled: non-NVIDIA "gpu"-named DRA drivers (e.g. mixed NVIDIA+AMD slices), non-node-local full-GPU slice topologies (`nodeSelector` / `allNodes`), and `gpu.nvidia.com` allocations from pools that no node-local ResourceSlice publishes or that multiple nodes publish (ambiguous). Arbitrary pool *names* are supported — attribution goes through the slice-derived pool → node map (`spec.nodeName`), not name equality — so only unpublished or ambiguous pools reject. This issue tracks *if and when* to support any of those states, so the fail-fast boundary has a home.

## Context

During PR #1620's review, successive rounds began building a partial kai-scheduler compatibility model inside `inference-perf` (raw-slice mirrors of kai v0.14.1's classification, capacity caps, dual occupancy ledgers, pod-bound claim attribution) to behave correctly on configurations no AICR recipe produces. Review convergence (2026-07-08) deliberately reversed that: the validator validates supported configurations and fails fast on unsupported ones; it does not emulate the scheduler. Generic ResourceSlice topology validation remains in the conformance probe (`validators/internal/allocmode` / `dra-support`), where compliance-checking belongs.

## Known accepted divergences

[...]

Issue #1644, opened 2026-07-07
Re-enable GCP UAT serve phase: fix ~12GB vllm-runtime Frontend image pull


njhensley:
## Context

The GCP UAT `serve` phase (the inference CUJ — `tests/uat/gcp/run serve`, driven by the `UAT - serve` step in `.github/workflows/uat-gcp.yaml`) was **temporarily disabled** because it consistently times out.

### Root cause

The served `DynamoGraphDeployment` (`vllm-agg`) has two components sharing the `nvcr.io/nvidia/ai-dynamo/vllm-runtime:1.2.1` image (~12GB):

- **`VllmDecodeWorker`** is pinned to the GPU node (`nodeSelector: nodeGroup=gpu-worker`), where the image is already cached — it comes up in ~90s (`Container image "...vllm-runtime:1.2.1" already present on machine`).
- **`Frontend`** has no GPU nodeSelector, so it schedules onto a fresh non-GPU node with no cached copy and must pull the full ~12GB image from scratch. It wedges in `ContainerCreating` past the 30m `SERVE_READY_TIMEOUT_SECONDS`, so the phase fails:

  ```
  Error: DynamoGraphDeployment vllm-agg did not become ready within 1800s
  status: Resources not ready: podclique/vllm-agg-0-frontend: desired=1, ready=0
  ```

Observed on run 28889370008.

## What was done

The `UAT - serve` step in `.github/workflows/uat-gcp.yaml` was commented out. The inference **stack** (dynamo operator + KAI scheduler + DRA driver) is still stood up and validated by `prep → install → conformance → verify`; only the served-workload run is skipped.

## Proposed fix (options)

- **Pin the Frontend to the GPU node** (it uses the same runtime image, so it reuses the cached layer). Simplest; keeps everything on one node.
- **Pre-pull / cache** the `vllm-runtime` image onto the pool (e.g. a warmup DaemonSet or node image seeding) before the serve phase.
- **Use a smaller/frontend-only image** for the Frontend component if one exists.

## Acceptance

- Re-enable the `UAT - serve` step (uncomment) and restore the `Serve` summary row.
- A GCP inference UAT run reaches a served `/v1/chat/completions` completion within the readiness budget.

mchmarny:
Triaged: confirmed at P2 / Ready. Root cause and fix options (pre-pull or nodeSelector for the Frontend component) are well understood and self-contained — actionable as-is, appropriately prioritized.

mchmarny:
[...]

Issue #1629, opened 2026-07-07
evidence: collect-evidence.sh hardcodes full-GPU DRA in dra-support and secure-access paths


yuanchen8911:
## Summary

`pkg/evidence/cncf/scripts/collect-evidence.sh` still hardcodes **full-GPU DRA** (`gpu.nvidia.com` DeviceClass + `resource.k8s.io/v1` ResourceClaim) in two of its CNCF evidence paths:

- the **dra-support** path — [collect-evidence.sh#L241-L275](https://github.com/NVIDIA/aicr/blob/b6da88af58266bd014936d0cce778a7c1a63700e/pkg/evidence/cncf/scripts/collect-evidence.sh#L241-L275)
- the **secure-access** path — [collect-evidence.sh#L389-L499](https://github.com/NVIDIA/aicr/blob/b6da88af58266bd014936d0cce778a7c1a63700e/pkg/evidence/cncf/scripts/collect-evidence.sh#L389-L499)

The Go validators were made **capability-driven** in #1327 (PR #1620): they probe which GPU allocation mechanism is actually usable (device plugin `nvidia.com/gpu` vs full-GPU DRA), discover the served `resource.k8s.io` API version (`v1` | `v1beta2` | `v1beta1`), and exercise the matching secure-access test. The NVIDIA DRA driver's supported configuration is ComputeDomain-only (no `gpu.nvidia.com` DeviceClass), so on supported clusters the hardcoded DRA evidence path fails or produces misleading evidence even though the cluster is conformant via the device plugin.

This was called out during review of PR #1620. It is a **non-blocking follow-up** for that PR (reclassified 2026-07-07): the `--cncf-submission` evidence script is self-contained and still runs correctly on DRA-enabled clusters, and its DRA-only limitation is documented in `docs/conformance/cncf/index.md` until this lands.

## Scope

Bring `collect-evidence.sh` to parity with the capability-driven validator behavior:

- **Allocation-mode detection**: probe DeviceClass/ResourceSlices and node allocatable to decide device-plugin vs full-GPU DRA (mirror `allocmode.Detect` in `validators/internal/allocmode/allocmode.go`).
- **Served-version discovery**: use the newest served `resource.k8s.io` version (`v1`, `v1beta2`, `v1beta1`) instead of hardcoding `v1`; build the version-correct ResourceClaim shape (`v1beta1` has no `exactly` wrapper).
[...]

PR #1671, opened 2026-07-09
feat(recipes): device-plugin GPU allocation production default

### LINKED ISSUE #1327: GPU allocation policy: device-plugin production default with recipe-configured DRA opt-in
## Summary

AICR's production default for whole-GPU allocation must be the device plugin
(`nvidia.com/gpu`), with full-GPU DRA available as an explicit, recipe-level
opt-in — and AICR validation workloads must request GPUs in the form matching
the configured mode. The governing contract:

**Configuration selects the allocation policy. Cluster inspection verifies it.
Mismatch fails closed. Only recipe-less standalone runs select automatically.**

This supersedes the original framing of this issue ("device plugin, not DRA").
DRA remains load-bearing for ComputeDomain/IMEX (GB200/GB300 MNNVL) and
networking claims in every configuration; this contract governs whole-GPU
allocation only.

## Allocation policy

One enum, resolved from fully hydrated recipe values (base + platform values +
overlays + mixins) via `GetValuesForComponentWithContext`, carried to
validators in `ValidationInput`:

| Policy | Meaning |
|---|---|
| `device-plugin-extended-resource` | Whole GPUs via device plugin, `nvidia.com/gpu` requests |
| `dra-resource-claim` | Whole GPUs via DRA, `gpu.nvidia.com` ResourceClaims |
| `dra-extended-resource` | Reserved: KEP-5004 mapped `nvidia.com/gpu` on a DRA backend (not yet validated by AICR) |
| `unspecified` | No recipe context (standalone validator / evidence runs): capability-driven automatic selection |

Backend and request syntax are deliberately distinct: after KEP-5004, an
extended-resource request can be served by either backend, so
`extended-resource` alone would be ambiguous.

## Configuration states

Inputs: `nvidia-dra-driver-gpu` values `resources.gpus.enabled` (the switch)
and `gpuResourcesEnabledOverride` (the chart-guard waiver only), plus
`gpu-operator` `devicePlugin.enabled` (to be pinned explicitly).

| # | `gpus.enabled` | `Override` | `devicePlugin` | Verdict | Policy |
|---|---|---|---|---|---|
| 1 | `false` | `false` | `true` | **Supported — production default** | `device-plugin-extended-resource` |
| 2 | `true` | `true` | `false` | **Experimental — DRA-only opt-in** (see below) | `dra-resource-claim` |
| 3 | `true` | `true` | `true` | Transitional: warn before the default flip, error after | `dra-resource-claim` |
| 4 | `true` | `false` | any | Invalid immediately: upstream chart install guard rejects | — |
| 5 | `false` | `true` | `true` | Transitional: warn before the default flip, error after (inert waiver disarms the tripwire) | `device-plugin-extended-resource` |
| 6 | `false` | any | `false` | Invalid immediately: no whole-GPU advertiser | — |
| 7 | — no recipe context — | | | Standalone runs | `unspecified` |

Invariant: **exactly one whole-GPU advertiser per node** (`gpus.enabled` XOR
`devicePlugin.enabled`). Rows 3 and 5 exist only because today's recipes are
dual-advertised; they warn until the default flip lands, then become errors.

Production defaults (all intent pinned explicitly, no upstream inheritance):

```yaml
# nvidia-dra-driver-gpu
gpuResourcesEnabledOverride: false
resources:
  gpus:
    enabled: false
  computeDomains:
    enabled: true
```

```yaml
# gpu-operator
devicePlugin:
  enabled: true
```

The DRA opt-in overlay must change all three: GPU DRA on, waiver on, device
plugin off.

## Enforcement

- Validators compare the configured policy against inspected cluster state
  (`InspectGPUAllocationFacts` / `VerifyGPUAllocation` / `SelectGPUAllocation`,
  the last for `unspecified` only) and **fail closed on mismatch** — no silent
  fallback to the other mechanism. A cluster drifted from its recipe is a
  validation failure by design.
- Bundle-time changes to allocation-policy keys are phased out (current docs
  recommend them, so no immediate break): reject `--dynamic` allocation-policy
  paths immediately; warn/deprecate static bundle-time changes while docs
  migrate to recipe overlays; later reject actual policy changes with
  `ErrCodeInvalidRequest` by comparing the recipe tuple against the final
  effective tuple. Enforced in `pkg/bundler` after alias resolution (covers
  CLI `--set`/`--set-json`/`--set-file`, config, REST, SDK, parent-object
  overrides). `argocd-helm` and manual post-generation edits are explicitly
  outside this guarantee.

## Experimental status of DRA-only (row 2)

"Experimental" means: AICR renders and deploys the three-part opt-in, and the
mode-aware checks exercise ResourceClaims. Full `aicr validate` is NOT
guaranteed. Graduation to supported requires all of:

- [ ] inference node discovery and sizing from ResourceSlices
- [ ] NCCL discovery and runtime-template compatibility
- [ ] deployment nvidia-smi verification on DRA-only nodes
- [ ] Karpenter / cluster-autoscaling compatibility
- [ ] `validate --require-gpu`
- [ ] kai-scheduler mapped-extended-resource support, or mode-aware kai workloads
- [ ] CNCF evidence collection (`--cncf-submission` bypasses Go validator dispatch; separate work)
- [ ] live managed-cluster validation
- [ ] separate verification of the KEP-5004 mapped `nvidia.com/gpu` path (`dra-extended-resource`)

Docs must not call row 2 supported until this checklist passes.

## Delivery plan

1. **PR #1620** (foundation, scope frozen): capability inspection, permanent
   standalone automatic selection (`unspecified`), validator execution
   hardening. Explicitly NOT: recipe-configured selection, full DRA-only
   support, KEP-5004 mapped-path validation.
2. **PR "1.5"** (stacked on #1620): policy resolution from hydrated values;
   explicit `devicePlugin.enabled: true` pin (makes the tuple observable);
   `ValidationInput` serialization; Inspect/Verify/Select separation;
   fail-closed gates at secure-accelerator-access, inference-perf, and
   dra-support's full-GPU behavioral subtest; inference DRA-only node-discovery
   fix (removes one graduation blocker; does not graduate DRA-only);
   transitional warnings (rows 3/5); immediate rejection of rows 4/6;
   `--dynamic` rejection + static bundle-override deprecation warnings.
   Preferred merge: back-to-back with #1620 after both are approved.
3. **PR 2** (atomic): flip production defaults to row 1; convert stock
   kai-scheduled workloads to `nvidia.com/gpu`; regenerate the BOM; rendered-
   output regression tests; promote rows 3/5 to errors.
4. **Follow-ups:** DRA-only graduation checklist items; runtime detector's
   broader dual-advertisement warning; #1652 generalized topologies (stay
   fail-fast).

## Non-goals

- Modeling unsupported GPU DRA topologies/drivers (#1652 — fail fast).
- Making `--cncf-submission` evidence collection capability- or policy-driven.
- Externally managed whole-GPU advertisers (row 6 fails closed; a future
  platform-supplied advertiser is a registry/component decision).
- Changing ComputeDomain/IMEX or networking DRA usage in any way.

## Note on KEP-5004

KEP-5004 does not make DRA "the default allocator": it lets the same
`nvidia.com/gpu` extended-resource request be served by either backend, and
forbids both backends advertising the same resource on one node. Beta/on in
K8s 1.36; GA targeted for 1.37 (unreleased). kai-scheduler v0.14.1 does not
implement mapped-extended-resource scheduling. The production default remains
row 1 until the ecosystem (kai, DCGM attribution, quota, autoscaling) is
validated together — readiness-based, not version-based.

---

*Original issue body (2025 framing, superseded by the contract above) is
preserved in the edit history. The flag-semantics truth table and live
verification evidence are in the decision-record comment below.*

yuanchen8911:
## Summary

Device-plugin whole-GPU allocation becomes the **production default**: `nvidia-dra-driver-gpu` ships `resources.gpus.enabled: false` + `gpuResourcesEnabledOverride: false` (armed tripwire) with `resources.computeDomains.enabled: true` pinned explicitly (load-bearing for GB200/MNNVL IMEX), all per-overlay DRA-GPU overrides removed, **every kai-scheduled workload converted to scalar `nvidia.com/gpu` requests in the same commit** (pre-flip conversion would break them: kai rejects scalar requests on ResourceSlice-bearing nodes), and the transitional configuration states promoted from warnings to errors.

> Final PR (3 of 3) of the #1327 plan. Rebased onto main after #1664 merged — now a single-commit diff (+393/−339, 25 files).

Fixes #1327

## Motivation / Context

#1327's contract and sequencing: #1620 (merged) shipped capability inspection; #1664 (merged) added configuration-selected, capability-verified allocation policy; this PR completes the plan by flipping the stock catalog from the transitional dual-advertised configuration to the device-plugin production default. Atomicity is the load-bearing constraint: the recipe flip, the overlay-override removals, and ALL kai-scheduled workload conversions must land together — kai-scheduler rejects scalar GPU requests on nodes bearing raw GPU ResourceSlices (the scalar-GPU-on-DRA-node rejection in kai's `node_info.go`, upstream-marked "temporary fix"), so converting workloads before the flip breaks them on stock clusters, while flipping without converting strands the DRA-form workloads after it.

- **Component defaults** (`recipes/components/nvidia-dra-driver-gpu/values.yaml` + `values-oke.yaml` — the OKE chain deep-merges the latter, so both flip): production tuple with guidance comments covering the three-part experimental DRA opt-in and the upgrade boundary.
- **9 overlay overrides removed** (the per-overlay `gpuResourcesEnabledOverride: true` waivers) — every stock overlay now inherits the flipped defaults.
- **Workload conversions** (whole-GPU ResourceClaims → `nvidia.com/gpu` limits; ComputeDomain/IMEX claims untouched): `demos/workloads/inference/vllm-agg.yaml`, the `dynamo-vllm-agg` + `gang-scheduling-test` CNCF evidence manifests, `tests/manifests/dynamo-vllm-smoke-test.yaml`, and `phase_serve` in `tests/uat/{aws,gcp}/run` (runner source converted and ready; the workflow step remains disabled pending #1644).
- **Rows 3/5 promoted to errors** (`pkg/validator/v1`): dual advertisement and the inert waiver now fail resolution with `ErrCodeInvalidRequest` and actionable guidance naming the actual components (`gpu-operator-ocp` on OpenShift); the three-part opt-in still resolves `dra-resource-claim` cleanly. The runtime detector's dual-advertisement warning stays a warning (follow-up scope).
- **Catalog acceptance guard** (`TestStockCatalogResolvesDevicePluginPolicy`): every stock overlay must resolve `device-plugin-extended-resource` with the explicit pins present (override explicitly false, operator `devicePlugin.enabled` explicitly true, `computeDomains.enabled` true) — presence-aware so resolver fallbacks can't mask deleted pins, with non-vacuity counters. This is a hydrated-tuple sweep; chart rendering is covered by the bundler/e2e suites.
- **Upgrade boundary documented** (`docs/user/validation.md` + component values header): applying the flip does not drain running full-GPU claim workloads — a surviving claim keeps its prepared GPU while the scalar ledger admits new workloads onto the same device. The documented procedure uses a **driver-specific, fail-closed drain check** (filters `.status.allocation.devices.results[].driver == "gpu.nvidia.com"` so legitimate IMEX claims don't block; captures each stage so a failed query can't masquerade as safe).
- **Docs sweep**: device-plugin as default in the AKS guide and component catalog with corrected enforcement wording ("recipe-backed validation rejects... skipping validation bypasses the check"), configuration-selected terminology, accurate KEP-5004 timeline/semantics, consistent UAT `phase_serve` status. BOM regen: no change.

Related: #1620, #1664, #1652, #1629, #1644

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change) — see Rollout notes: transitional configurations now fail validation-time resolution; in-place upgrades of dual-advertised clusters must follow the documented drain procedure

## Component(s) Affected

- [x] Recipe engine / data (`recipes/` components, overlays, values)
- [x] Docs/examples (`docs/`, `demos/`)
- [x] Other: CNCF evidence manifests, UAT runner sources, smoke-test manifests

## Implementation Notes

The values-oke.yaml flip was required beyond the spec'd nine overlays — the OKE service chain deep-merges it over the base values, so leaving it would have kept every OKE stock recipe in the now-rejected dual row (caught by the catalog sweep). Post-flip, `dra-support` hard-requires a validating NVIDIA ResourceSlice; on MNNVL platforms the compute-domain slices satisfy it (proven live below); the committed H100 nim-eks evidence shows the compute-domain kubelet plugin publishes slices there too.

## Testing

```bash
make qualify        # full gate at the final head — PASS (isolated golangci-lint cache)
go test -race ./... # via qualify; catalog sweep covers every stock overlay
```

**Live upgrade-transition matrix** (EKS GB200 `yljtrxpmzu`, multi-arch images `pr3-82ea13cd`, 2026-07-08; the PR-owned delta after evidence head `82ea13cd` is documentation-only; the branch was subsequently rebased over newer main changes not covered by this hardware run — e.g. #1651 and #1670 on the NCCL execution path — so the numerical results are scoped to `82ea13cd`):

| Cell | Result |
|---|---|
| Negative resolution (`--no-cluster`): dual-advertised + inert-waiver recipes | both rejected with the promoted row-3/5 errors; stock control clean |
| Pre-flip: real full-GPU `gpu.nvidia.com` claim workload | allocated + Running (recreates the pre-flip world) |
| **Drain check and transition sequence reproduced** (this 2026-07-08 run used chart/runtime `0.4.1-rc.1` — a temporary bug workaround then deployed on the test cluster, never a version AICR pinned; stock is `0.4.1` throughout — see the shipping-runtime confirmation row below) | fail-closed check caught the live claim (exit 1, printed ns/name/reservedFor), passed after drain, flip applied, `gpu.nvidia.com` slices + DeviceClass gone in ~15s, compute-domain slices remained, scalar allocatable restored |
| **Shipping-runtime confirmation on chart `0.4.1`** (2026-07-09) | cluster upgraded `0.4.1-rc.1` → `0.4.1` in the pre-flip dual-advertised state, then the values flip applied on the shipping chart: `gpu.nvidia.com` DeviceClass removed immediately, GPU ResourceSlices gone in ~20s, compute-domain slices persisted; drain check exit 0 with live ComputeDomain claims present (no false positive); scalar workloads rescheduled post-flip (2× 4-GPU workers); ComputeDomain recreated — fresh IMEX channel claims allocated end-to-end on `0.4.1`; conformance phase **8/8 passed** incl. `dra-support` and `gang-scheduling` |
| Post-flip validation using **unmodified stock recipe inputs** on the `0.4.1-rc.1`-deployed cluster | inference **15/15** (policy artifact `device-plugin-extended-resource` from stock hydration; `dra-support` passed via compute-domain slice validation — the #1620 forward constraint; 108,533 tok/s, TTFT p99 477.6 ms) · training **10/10** (NCCL net 320.87 GB/s, nvls 840.98 GB/s, with fresh IMEX channel claims allocated mid-TrainJob — ComputeDomain load-bearing post-flip) |
| Converted scalar `vllm-agg` under kai-scheduler | both pods bound and scheduled, decode worker 1/1 Running with `nvidia.com/gpu: 1`, DynamoGraphDeployment Ready |
| Restore + IMEX-ignore proof | helm values byte-identical to pre-run state; drain check exits 0 with 4 allocated compute-domain claims present |

Cross-reviewed pre-push over five review rounds; all findings addressed (upgrade boundary + fail-closed drain check, enforcement-wording accuracy, KEP-5004 semantics, explicit-pin sweep guards, consistent UAT status).

## Risk Assessment

**Rollout notes:** For freshly provisioned clusters the flip is transparent — stock recipes deploy device-plugin allocation and validate cleanly. Two deliberate behavior changes: (1) transitional configurations (dual advertisement, inert waiver) now fail validation-time policy resolution with actionable errors instead of warnings; (2) **in-place upgrades of clusters running full-GPU DRA claim workloads must follow the documented drain procedure** (`docs/user/validation.md`) — the Helm change alone does not drain running claims, and a surviving claim can contend with newly admitted scalar workloads for the same physical GPU. The experimental three-part DRA opt-in remains available per #1327. The source change is a single-commit revert, but in-cluster rollback crosses the same independent-ledger boundary in reverse: drain scalar GPU workloads before re-enabling full-GPU DRA, verify they are gone, apply the rollback, wait for `gpu.nvidia.com` ResourceSlices to appear, and only then start claim workloads.

## Checklist


validators/conformance/secure_access_check.go
```diff
@@ -314,11 +315,27 @@ func CheckSecureAcceleratorAccess(ctx *validators.Context) error {
...
+	// Only unspecified (standalone runs) keeps the capability-driven
+	// dispatch below.
+	policy := ctx.ValidationInput.GetGPUAllocationPolicy()
+	// Evidence BEFORE the verdict: a Verify failure below must leave the
+	// configured policy in the artifacts.
+	recordRawTextArtifact(ctx, "GPU allocation policy", "",
+		fmt.Sprintf("configured policy: %s", policy))
+	if verifyErr := verifyGPUAllocationPolicy(policy, mode); verifyErr != nil {
+		return verifyErr
+	}
```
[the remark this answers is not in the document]
yuanchen8911: This thread is anchored to the obsolete stacked diff: `validators/conformance/secure_access_check.go` is not part of this PR (it belongs to #1664, since merged) and left this PR's diff when the branch was rebased onto main after that merge. If the suggestion still applies to the code on main, it is best re-raised there rather than expanding this PR's scope.

validators/internal/allocmode/allocmode.go
```diff
@@ -652,26 +679,42 @@ func UsableDriverDeviceCounts(ctx context.Context, items []unstructured.Unstruct
...
+			}
 		}
 	}
-	return false, fmt.Sprintf(
-		"extended-resource attribution: no DeviceClass maps %s to DRA", resourceNVIDIAGPU), nil
+	if !backed {
+		detail = fmt.Sprintf(
+			"extended-resource attribution: no DeviceClass maps %s to DRA", resourceNVIDIAGPU)
+	}
+	return backed, detail, gpuClass, nil
```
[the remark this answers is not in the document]
yuanchen8911: This thread is anchored to the obsolete stacked diff: `validators/internal/allocmode/allocmode.go` is not part of this PR (it belongs to #1664, since merged) and left this PR's diff when the branch was rebased onto main after that merge. If the suggestion still applies to the code on main, it is best re-raised there rather than expanding this PR's scope.

docs/user/validation.md
```diff
@@ -376,9 +381,50 @@ recipes always do), `gpus.enabled=true` without
...
+result), and proceed only when it exits 0:
+
+```shell
+claims="$(kubectl get resourceclaims -A -o json)" || { echo "claim query FAILED — do not proceed"; exit 1; }
+unsafe="$(printf '%s' "$claims" | jq -r '
+  .items[]
+  | select([.status.allocation.devices.results[]?.driver] | index("gpu.nvidia.com"))
+  | "\(.metadata.namespace)/\(.metadata.name) reservedFor=\([.status.reservedFor[]?.name] | join(","))"')" \
+  || { echo "claim filter FAILED — do not proceed"; exit 1; }
+[ -z "$unsafe" ] || { printf 'unsafe full-GPU claims remain:\n%s\n' "$unsafe"; exit 1; }
```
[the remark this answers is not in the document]
yuanchen8911: A reserved-but-unallocated claim cannot exist server-side, so the allocation-driver filter already covers every reserved claim. Upstream `validateResourceClaimStatusUpdate` forbids the state this suggests covering: when `status.reservedFor` is non-empty and `status.allocation` is nil, the API server rejects the update with "may not be specified when `allocated` is not set" (https://github.com/kubernetes/kubernetes/blob/v1.36.2/pkg/apis/resource/validation/validation.go#L441-L446). So every claim with `reservedFor` necessarily carries the `allocation.devices.results` the jq filter keys on, and the procedure's "allocated or reserved" wording is satisfied. Broadening the filter beyond the `gpu.nvidia.com` allocation driver would also false-positive on legitimate ComputeDomain/IMEX claims, which must survive the flip — verified live on the GB200 cluster where the check passed with compute-domain claims present (exit 0) and failed closed on a full-GPU claim.

pkg/validator/v1/allocation_policy.go
```diff
@@ -209,12 +215,13 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult)
...
 		}
 		if devicePluginEnabled {
-			// TODO(#1327 follow-up): promote to an error after the
-			// production-default flip lands.
-			slog.Warn("transitional GPU allocation configuration: both full-GPU DRA and the device plugin are enabled — dual advertisement risks GPU over-admission; resolving policy to dra-resource-claim (that is what schedules under kai-scheduler)",
-				"policy", GPUAllocationPolicyDRAResourceClaim,
-				draDriverGPUComponentName, valuePathGPUsEnabled+"=true",
-				gpuOperatorComponentName, valuePathDevicePluginEnabled+"=true")
+			return "", errors.New(errors.ErrCodeInvalidRequest, fmt.Sprintf(
+				"invalid GPU allocation configuration: dual advertisement — %s %s=true and %s %s=true both advertise whole GPUs, risking GPU over-admission; exactly one advertiser is required. For the production default keep %s=false (with %s=false); for the experimental DRA opt-in flip all three values together: %s=true, %s=true, and %s=false (issue #1327)",
```
mchmarny: nit: two small wording points in this message. "For the production default **keep** `resources.gpus.enabled=false`" — the config being rejected has it `true`, so "set" is more accurate than "keep". And the opt-in guidance here says `devicePlugin.enabled=false` without the operator component name, while the inert-waiver message below includes it (`%s %s=false` with `operatorName`) — worth matching so OpenShift users see `gpu-operator-ocp` in both.

pkg/validator/v1/allocation_policy_sweep_test.go
```diff
@@ -0,0 +1,162 @@
...
+				if !waiverSet || waiver {
+					t.Errorf("%s: want explicitly false (present=%t value=%t)\n"+
+						"  The armed-tripwire pin must be EXPLICIT in stock hydrated values, not a\n"+
+						"  resolver fallback (issue #1327).",
+						valuePathGPUsEnabledOverride, waiverSet, waiver)
+				}
+				// ComputeDomain guard: load-bearing for GB200/MNNVL IMEX —
+				// a values change flipping it off would otherwise pass the
+				// policy sweep silently.
+				cdEnabled, err := lookupBoolValue(values, draRef.Name, "resources.computeDomains.enabled", false)
```
mchmarny: nit: `"resources.computeDomains.enabled"` is a string literal while the sibling pins use the `valuePath*` constants. A `valuePathComputeDomainsEnabled` constant next to the others makes a future path rename fail loudly here instead of silently checking a dead path (the `false` fallback would then error, so it fails closed either way — but the constant makes the intent explicit).

docs/user/validation.md
```diff
@@ -376,9 +381,50 @@ recipes always do), `gpus.enabled=true` without
...
+`kubectl ... | jq ...` masks a failed List as an empty, safe-looking
+result), and proceed only when it exits 0:
+
+```shell
+claims="$(kubectl get resourceclaims -A -o json)" || { echo "claim query FAILED — do not proceed"; exit 1; }
+unsafe="$(printf '%s' "$claims" | jq -r '
+  .items[]
+  | select([.status.allocation.devices.results[]?.driver] | index("gpu.nvidia.com"))
+  | "\(.metadata.namespace)/\(.metadata.name) reservedFor=\([.status.reservedFor[]?.name] | join(","))"')" \
+  || { echo "claim filter FAILED — do not proceed"; exit 1; }
```
mchmarny, line 424: low: the filter catches *allocated* `gpu.nvidia.com` claims — correct priority, since those are the double-allocation hazard. A pending/unallocated claim whose spec references the `gpu.nvidia.com` DeviceClass slips through, though: it holds no GPU (no over-admission risk), but post-flip the DeviceClass disappears and it strands unschedulable forever. Worth one sentence telling operators to delete those too while migrating.

mchmarny (dismissed):
Completes the #1327 plan with the atomicity constraint respected — the recipe flip, all 9 overlay-override removals (verified exhaustive against main), and every kai-scheduled workload conversion land in one commit, with the `values-oke.yaml` deep-merge hole closed. The promoted row-3/5 errors are correctly ordered after the row-6 no-advertiser gate so the guidance is always actionable, and the catalog sweep test is genuinely strong: presence-aware pin assertions that resolver fallbacks can't mask, plus non-vacuity counters. Live GB200 upgrade-transition matrix and the fail-closed drain procedure give good confidence in the breaking-change boundary.

Three nits inline (error-message wording, a test string literal, one drain-check edge for pending claims) — none blocking. Unit tests/lint/CLI e2e/security scan are green; E2E and GPU lanes were still in progress at review time, and the `needs-rebase` label means a rebase is needed before merge regardless.

pkg/validator/v1/allocation_policy.go
yuanchen8911: Both fixed in 8bf3214d: "keep" is now "set", and the opt-in guidance ends with `%s %s=false` using `operatorName`, matching the inert-waiver message — OpenShift recipes now name `gpu-operator-ocp` in both errors.

pkg/validator/v1/allocation_policy_sweep_test.go
yuanchen8911: Fixed in 8bf3214d: added `valuePathComputeDomainsEnabled` beside the other `valuePath*` constants (with a comment noting it is a pinned invariant, not a policy input) and the sweep test now uses it.

docs/user/validation.md
yuanchen8911: Fixed in 8bf3214d: added a sentence after the drain-check snippet telling operators to also delete pending (unallocated) claims whose spec references the `gpu.nvidia.com` DeviceClass — no over-admission risk, but they strand unschedulable once the flip removes the DeviceClass.

mchmarny (approved):

PR #1679, opened 2026-07-09
feat(supply-chain): default keyless and KMS signing to Rekor v2

### LINKED ISSUE #1650: feat(supply-chain): move release signing to Rekor v2
## Problem

AICR release signing writes to **Rekor v1** (`rekor.sigstore.dev`). Verified on
the latest release: v0.16.0's `recipe-catalog.sigstore.json` has
`checkpoint origin: rekor.sigstore.dev`, logIndex ~2.0B. Even though we run
cosign v3.1.1, the log target is set by the TUF-distributed SigningConfig, which
still routes everyone to v1 — and per Sigstore's 2026-06-28 post
([rekor-evolution](https://blog.sigstore.dev/rekor-evolution/)), the public-good
instance "will continue using Rekor v1 as the default log for the foreseeable
future."

This blocks transparency-log **identity monitoring** of our release signer
(#1623): v1 has ~1.9B entries, no way to query the index by certificate SAN, and
the only monitoring option is a linear firehose scan measured at ~50x slower than
the log grows. Rekor **v2** solves this — tile-based reads make an identity scan
feasible (measured ~570 entries/sec, single worker keeps up) — but only for
entries that are actually *in* v2. So to get monitoring, our signing must move to
v2 first.

We are pre-1.0, so the breaking verification change (consumers need cosign v3+)
is acceptable now.

## Validated (end-to-end spike)

Signed a test blob with the v2 signing config and verified it through our exact
stack:

- **Signs to v2:** entry landed in `log2025-1.rekor.sigstore.dev`, logIndex ~10.2M,
  `hashedrekord 0.0.2`.
- **TSA is free:** v2 returns no Rekor SET, but the Sigstore signing config already
  lists `timestamp.sigstore.dev`; the bundle carried 1 RFC3161 TSA timestamp with
  no extra work.
- **Our verifier already handles v2:** `sigstore-go v1.2.2` (our pinned version)
  returned `VERIFY OK`, validating both the tiles inclusion proof and the TSA
  timestamp. No dependency bump. The consumer-facing `cosign verify-blob-attestation`
  command is byte-for-byte unchanged (bundle-based verification is self-describing).

## Proposal / scope

1. **Signing config from TUF, not hardcoded.** Sigstore publishes a maintained
   `signing_config_rekor_v2.v0.2.json` target (v2 shard + TSA + `validFor`
   windows). Fetch it via a TUF client at release time so shard rotation
   (`log2025-1` → future `log2026-*`) is handled by Sigstore, not us. Do **not**
   hardcode the shard URL (Sigstore explicitly advises against it).
2. **`cosign attest-blob`** (`.goreleaser.yaml`): add `--signing-config <fetched>`.
3. **`aicr recipe sign-catalog`** (`pkg/bundler/attestation`): add a
   `--signing-config` flag and build the signer from a SigningConfig instead of the
   hardcoded v1 rekor URL (`keyless.go` `defaults.SigstoreRekorURL`).
4. **Resolve the `ANY`-selector question.** The Sigstore v2 config lists both v2
   and v1 with `rekorTlogConfig.selector: ANY`, both currently valid. Confirm cosign
   writes to v2 (not v1, not both); if ambiguous, ship a v2-only override derived
   from the fetched config.
5. **Docs:** state the cosign v3.0.1+ floor for verifiers; note the cutover release
   (entries ≤ vX on v1, ≥ vX+1 on v2) in release notes + supply-chain-verification.
6. **Tests:** e2e sign→verify round-trip on v2; keep private-Sigstore e2e green.

## Out of scope / follow-on

- Flipping `rekor-monitor` to v2 with identity scanning re-enabled — that's #1623,
  which this unblocks.
- Historical v1 entries stay on v1 forever and remain verifiable via the trust root
  (which carries both v1 and v2 log keys); no migration of past artifacts.

Blocks #1623. Part of #1149.

lockwobr:
## Summary

AICR attestations now record their transparency-log entry in **Rekor v2** by default instead of Rekor v1, for both keyless OIDC and KMS-backed signing, with opt-outs to v1 and a `cosign v3` verification floor.

## Motivation / Context

Rekor v2 is Sigstore's tile-based transparency log. It matters here because it makes **release-identity monitoring feasible**: monitoring AICR's release signer on the Rekor v1 firehose is ~50x too slow to keep up (there is no index query for a keyless SAN identity), whereas v2's bulk tile reads make a single-worker scan keep up. This PR is the signing half of that move; the monitoring half rides on it.

The public-good Sigstore default still points other clients at Rekor v1, so AICR opts into v2 explicitly by consuming the Sigstore-maintained v2 signing config from TUF. Pre-1.0, adopting it early is a deliberate, low-cost choice.

Fixes: N/A
Related: #1650, #1149

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `pkg/trust`, `pkg/evidence`, `.goreleaser.yaml`, `.github/actions`

## Implementation Notes

- **v2 signing config from TUF, not hardcoded.** `pkg/trust` fetches Sigstore's `signing_config_rekor_v2` target; `aicr trust update` refreshes it alongside the trusted root, signing resolves it from cache with a bounded network fallback on a cold cache, and shard rotation is handled by Sigstore. Selection uses sigstore-go's `root.SelectServices` (prefers highest API version, so v2 wins), matching cosign.
- **Default computed at the CLI layer.** Library, server, and test callers of `ResolveOptions` keep their v1 zero-value behavior; only the signing commands flip. All signing paths (bundle, catalog, evidence, KMS) go through one `transparencyForOptions` policy selection and one `SignOptionsFromResolve` mapper, so a signing field cannot be dropped by one caller.
- **Fail-closed rules.** A v2 selection with no timestamp authority is rejected (v2 has no inline SET, so the bundle would lack trusted time); the release hooks error if signing is enabled but the config path is unset (no silent fall-back to v1).
- **Opt-outs preserve v1.** `--rekor-url` (private/public v1) and `--signing-config` (custom) are mutually exclusive; existing users on `--rekor-url` / `spec.bundle.rekorURL` are unaffected. KMS follows the same rules.
- **Release wiring.** The `go-build-release` action runs `aicr trust update --emit-signing-config` (cosign cannot fetch the TUF v2 config itself) and feeds `cosign attest-blob --signing-config` + `aicr recipe sign-catalog --signing-config`.

## Testing

```bash
make qualify
```

Full `make qualify` passes: `make test` (`-race`, 0 failures), coverage **77.8%** (floor 75%), lint, tuning-check, e2e (CLI chainsaw, 23/23), scan (no vulnerabilities). Proven end-to-end: a default keyless sign lands in `log2025-1.rekor.sigstore.dev` (v2) with an RFC3161 TSA timestamp and verifies through sigstore-go v1.2.2 (AICR's pinned version). A code review (superpowers:code-reviewer) caught and fixed a bug where evidence signing still went to v1; a drift-guard test now covers the shared mapper. `bundle-headless-oidc-ci` asserts the default lands in a v2 shard.

## Risk Assessment

**Rollout notes:** Verifying v2 bundles requires **Cosign v3.0.1+**; the verify *commands* are unchanged (a bundle self-describes its log), and releases published before the cutover remain in Rekor v1 and verify with any recent Cosign. `aicr verify` handles both. KMS+v2 is exercised by the `kms-ministack` e2e workflow but was not run locally (no KMS key). A private/local v2 e2e stack (`rekor-tiles` + `timestamp-authority`) is deferred and tracked on #1650.

## Checklist

### DIFF
--- a/.github/actions/generate-slsa-predicate/action.yml
+++ b/.github/actions/generate-slsa-predicate/action.yml
@@ -52,3 +52,30 @@ runs:
         }
         EOF
         echo "SLSA_PREDICATE=${PREDICATE}" >> "$GITHUB_ENV"
+
+        # AICR signs release attestations to Rekor v2 (NVIDIA/aicr#1650). Fetch
+        # the TUF-distributed v2 SigningConfig (Rekor v2 endpoint + timestamp
+        # authority) so the goreleaser cosign + sign-catalog hooks can point
+        # --signing-config at it. Co-located with SLSA_PREDICATE so the invariant
+        # "signing enabled (SLSA_PREDICATE set) => signing config present" holds
+        # for every signing workflow, not just the release. go run compiles aicr
+        # from source (the built binaries do not exist yet). Requires Go, which
+        # every caller sets up before this action.
+        SC="${RUNNER_TEMP}/aicr-signing-config.json"
+        # Retry the TUF fetch (3 attempts, 5s/10s backoff) to absorb transient
+        # Sigstore TUF CDN blips, mirroring the goreleaser cosign hook — this is a
+        # hard prerequisite for the whole signing flow, which fails closed if the
+        # config is missing. The `-s` check treats an exit-0-but-empty-file result
+        # as a failure so a truncated write is retried, not silently exported.
+        for n in 1 2 3; do
+          GOFLAGS=-mod=vendor go run ./cmd/aicr trust update --emit-signing-config "${SC}" && [ -s "${SC}" ] && break
+          if [ "$n" -lt 3 ]; then
+            echo "trust update attempt $n failed; retrying" >&2
+            sleep $((n * 5))
+          else
+            echo "trust update failed after 3 attempts" >&2
+            exit 1
+          fi
+        done
+        echo "AICR_SIGNING_CONFIG=${SC}" >> "$GITHUB_ENV"
+        echo "Rekor v2 signing config written to ${SC}"
--- a/.github/actions/install-aicr-release/action.yml
+++ b/.github/actions/install-aicr-release/action.yml
@@ -27,15 +27,32 @@ inputs:
   aicr-version:
     description: 'Released aicr version tag to install (e.g. v1.2.3).'
     required: true
+  cosign_version:
+    description: 'Cosign version (from .settings.yaml via load-versions; must be >= v3.1.0 to verify release provenance logged to Rekor v2 as hashedrekord/PAE).'
+    required: true
 
 runs:
   using: composite
   steps:
+    # GitHub does not enforce `required: true` on composite-action inputs at
+    # runtime, so a caller that omits cosign_version would pass '' to
+    # cosign-release and silently get the installer default (which lags the
+    # v3.1.0 floor). Fail closed instead. See #1650.
+    - name: Require cosign_version
+      shell: bash
+      env:
+        COSIGN_VERSION: ${{ inputs.cosign_version }}
+      run: '[ -n "$COSIGN_VERSION" ] || { echo "cosign_version input is required (>= v3.1.0 for Rekor v2)" >&2; exit 1; }'
     # verify-blob-attestation needs no OIDC token (verification, not signing);
     # it fetches only the Sigstore trusted root, so the caller's contents:read
     # ceiling suffices. Pinned to the same SHA the release/build workflows use.
     - name: Install Cosign
       uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6  # v4.1.2
+      with:
+        # Pin from .settings.yaml (>= v3.1.0). The release binary provenance is
+        # logged to Rekor v2 as hashedrekord/PAE; keep the verifier in lockstep
+        # with the signer rather than the installer default (v3.0.6). See #1650.
+        cosign-release: '${{ inputs.cosign_version }}'
     - name: Download, verify, and install released aicr
       shell: bash
       env:
--- a/docs/contributor/rekor-v2-signing.md
+++ b/docs/contributor/rekor-v2-signing.md
@@ -1,0 +1,189 @@
+# Rekor v2 Signing
+
+AICR signs its keyless attestations to **Rekor v2** by default. This page
+explains what Rekor v2 is, why AICR moved to it, and how the signing path is
+wired, so contributors can reason about the transparency-log behavior and its
+verification and monitoring consequences.
+
+## What Rekor v2 is
+
+Rekor is Sigstore's signature transparency log: the append-only, publicly
+auditable record that every keyless signature is entered into, so that a
+signature cannot be produced and then hidden. Rekor v2 is a redesign of that log
+with a few properties that matter here:
+
+- **Tile-based.** The log is served as immutable, content-addressed tiles (the
+  C2SP tlog-tiles layout, backed by Trillian-Tessera) rather than a live
+  database queried entry-by-entry. Tiles are static files, cheap to host on a
+  CDN and cheap to read in bulk.
+- **Yearly shards.** Each year gets a fresh log shard at
+  `log<year>-<rev>.rekor.sigstore.dev` (for example `log2025-1`). A shard is
+  bounded in size, and old shards are frozen and archived as static tiles. This
+  is the core of v2's "cheaper to run" thesis, and it keeps consistency proofs
+  and monitoring bounded.
+- **Integrated witnessing.** v2 folds witnessing into the log, strengthening the
+  append-only guarantee.
+- **No inline timestamp.** A v2 entry does not carry a signed entry timestamp
+  (the "SET" that v1 provided). A bundle needs trusted time from a separate
+  RFC3161 **timestamp authority (TSA)** instead. This is why AICR attaches a TSA
+  when signing to v2 and fails closed if a v2 target has no TSA.
+- **No search index.** v2 removed v1's `/api/v1/index/retrieve` search. Finding
+  entries is delegated to clients reading tiles.
+
+For the migration status and Sigstore's own guidance, see the
+[Rekor v2 GA post](https://blog.sigstore.dev/rekor-v2-ga/) and the
+[Rekor evolution post](https://blog.sigstore.dev/rekor-evolution/).
+
+## Why AICR moved to Rekor v2
+
+The motivating driver is **release-identity monitoring**. AICR wants to watch the
+public-good log for any entry made under its release-signing identity that a
+release did not produce, since such an entry signals OIDC or key compromise. On
+Rekor v1 that is not feasible: v1 has no way to query the index by certificate
+SAN, and AICR's release identity is keyless (an ephemeral Fulcio certificate, no
+email, no fixed public key), so the only way to find AICR's entries is a linear
+walk of the entire public-good firehose. Measured, that scan runs roughly 50x
+slower than the log grows, so it can never keep up in a bounded job. The analysis
+and the pivot are tracked in [#1623](https://github.com/NVIDIA/aicr/issues/1623).
+
+Rekor v2 makes monitoring feasible: tiles are read in bulk (256-entry bundles),
+so a scan runs about two orders of magnitude faster than the v1 per-entry fetch,
+and a single worker keeps up with the ecosystem's write rate. A yearly shard is
+also small and bounded rather than billions of entries. Signing to v2 is the
+prerequisite for that monitoring, tracked in
+[#1650](https://github.com/NVIDIA/aicr/issues/1650), under the closed
+supply-chain epic [#1149](https://github.com/NVIDIA/aicr/issues/1149).
+
+Beyond monitoring, v2 is Sigstore's forward direction (cheaper to operate,
+stronger append-only guarantees), and AICR is pre-1.0, so adopting it early is a
+deliberate, low-cost choice.
+
+Note that the public-good Sigstore default still points other clients at Rekor
+v1 for now, so AICR opts into v2 explicitly by consuming the v2 signing config
+(below) rather than waiting for the ecosystem default to flip.
+
+## How AICR signs to v2
+
+### Signing config from TUF, not hardcoded
+
+Signing endpoints (which Fulcio, which Rekor, which TSA) come from a Sigstore
+**SigningConfig**. Sigstore distributes a v2 SigningConfig as a TUF target
+(`signing_config_rekor_v2.v0.2.json`) that lists the current v2 shard and a TSA,
+each with a validity window. AICR fetches that target through the TUF client
+(`pkg/trust`), so shard rotation is handled by Sigstore and nothing is
+hardcoded. `aicr trust update` refreshes both the trusted root (verification
+material) and this signing config (sign-side endpoints); signing also resolves
+the config from the local cache and falls back to a bounded network fetch on a
+cold cache, so a signer works without a prior explicit update.
+
+Selection of the active Rekor and TSA from the config uses sigstore-go's
+`root.SelectServices`, which prefers the highest supported API version and never
+mixes versions. Because the v2 config lists both a v2 and a v1 Rekor, this
+selects v2; the v1 entry is only a fallback for clients that support v1 only.
+
+### The default is set at the CLI layer
+
+Rekor v2 is the default for keyless signing, but the default is computed in the
+CLI commands, not in the library's zero-value options. This keeps library,
+server, and test callers of `attestation.ResolveOptions` on their existing v1
+behavior, so the flip is contained to the signing commands and does not
+surprise embedders. All three keyless signing paths default to v2 consistently:
+
+- `aicr bundle --attest`
+- `aicr recipe sign-catalog`
+- evidence signing (`aicr validate --emit-attestation --push`,
+  `aicr evidence publish`, `aicr evidence sign`)
+
+The single `attestation.SignOptionsFromResolve` mapper carries the signing-target
+fields (Fulcio, Rekor, signing config) from `ResolveOptions` into `SignOptions`
+for every path, so a new field cannot be dropped by one caller and silently sign
+to the wrong log.
+
+### Opting out to Rekor v1 or a custom config
+
+Signing target selection has this precedence: a `--signing-config` file, then the
+TUF v2 config (default), then a Rekor v1 URL.
+
+- `--rekor-url <url>` signs to **Rekor v1** at that URL, for a private instance
+  or the public-good v1 URL. Existing users who set `--rekor-url` or
+  `spec.bundle.rekorURL` are therefore unaffected by the default flip.
+- `--signing-config <file>` signs with a custom SigningConfig JSON, for an edited
+  config or a private v2 instance.
+- KMS signing (`--signing-key`) follows the same rules: it signs to Rekor v2 by
+  default and honors the same `--rekor-url` / `--signing-config` opt-outs. The
+  `KMSAttester` builds its transparency policy through the same
+  `transparencyForOptions` path as keyless signing, so both stay in lockstep. A
+  KMS v2 entry carries public-key verification material (no Fulcio certificate).
+
+`--rekor-url` and `--signing-config` are mutually exclusive, so an operator's
+private Rekor is never silently overridden by the v2 default.
+
+### Fail-closed rules
+
+- A v2 selection with no timestamp authority is rejected: a v2 bundle carries no
+  inline timestamp, so it would have no trusted time and could not be verified.
+- In the release pipeline, if signing is enabled (`SLSA_PREDICATE` set) but the
+  signing config path is unset, the goreleaser hooks error rather than let cosign
+  silently fall back to Rekor v1.
+
+### Release wiring
+
+At release time the `generate-slsa-predicate` action runs
+`aicr trust update --emit-signing-config` to materialize the v2 SigningConfig to
+a file and exports its path as `AICR_SIGNING_CONFIG` (co-located with
+`SLSA_PREDICATE` so every signing workflow inherits it). The goreleaser hooks pass
+that file to both `cosign attest-blob --signing-config` (the binary attestation)
+and `aicr recipe sign-catalog --signing-config` (the recipe catalog), so both
+land in the same v2 log. `cosign` cannot fetch the TUF-distributed v2 config
+itself (`cosign signing-config create` only builds one from hardcoded URLs), so
+`aicr` supplies it.
+
+**The signer must be Cosign >= v3.1.0** (pinned in `.settings.yaml` and passed to
+`cosign-installer` via `cosign-release` in every signing workflow). Only v3.1.0+
+logs a DSSE attestation to Rekor v2 as a `hashedrekord` entry over the envelope's
+pre-auth encoding (PAE); older Cosign writes the legacy `dsse` entry type, which
+no released sigstore-go can verify. The `bundle-headless-oidc-ci` chainsaw guard
+only catches a regression on the `bundle --attest` path, so keep the pin in step
+with the installer default (which lags well behind v3.1.0). See #1650.
+
+## Verification
+
+The consumer verification commands do not change. A bundle self-describes which
+log it is in (the entry, its inclusion proof, and the checkpoint origin are all
+embedded), and the public-good trusted root already carries the v2 log's key and
+a TSA, so `aicr verify` and `cosign verify-blob-attestation` verify a v2 bundle
+with the same flags as a v1 one.
+
+`aicr verify` needs only the `aicr` binary and handles v1 and v2 transparently
+(it embeds a v2-capable sigstore-go). The **Cosign v3.0.1+** floor applies only to
+the `cosign verify-blob-attestation` path: older Cosign cannot parse a v2 tiles
+inclusion proof or its RFC3161 timestamp. Releases published before the v2
+cutover remain in Rekor v1 and verify with any recent Cosign.
+
+## Operational notes
+
+- **Shard rotation** is Sigstore's responsibility: when a new yearly shard is
+  deployed, Sigstore updates the TUF signing-config target, and AICR picks it up
+  on the next `trust update` or on-demand fetch. Do not hardcode a shard URL.
+- **Stale-cache downgrade window.** `root.SelectServices` applies each service's
+  validity window at sign time. If a cached signing config's v2 shard ages past
+  its `validFor.end` before the cache refreshes, selection falls back to the
+  still-valid v1 entry — a silent Rekor v1 downgrade. The window is bounded: the
+  TUF client refreshes the cache when its timestamp metadata expires, so the
+  worst case is that expiry interval, and a v1 selection under the v2 default is
+  logged as a warning (see `NewSigningConfigPolicyFromTUF`). Run `aicr trust
+  update` to refresh eagerly; the `bundle-headless-oidc-ci` shard assertion
+  guards the CI path but not individual end users.
+- **A private or local Rekor v2 test stack** would need both a `rekor-tiles`
+  server and a `timestamp-authority`, driven via `--signing-config`. The existing
+  private-Sigstore chainsaw test exercises the Rekor v1 path via `--rekor-url` and
+  is unaffected.
+- **`AICR_SIGNING_CONFIG` binds as `--signing-config` process-wide.** Because
+  `generate-slsa-predicate` exports it into the job environment, any later
+  `aicr … --rekor-url …` in the same job auto-picks it up as `--signing-config`
+  and fails closed with the mutual-exclusivity error (the two select different
+  logs). A caller combining a private `--rekor-url` with an inherited config env
+  must `unset AICR_SIGNING_CONFIG` first — as the private-Sigstore chainsaw test
+  does. The error message names both env vars so the collision is diagnosable.
+- The `bundle-headless-oidc-ci` chainsaw test asserts that a default signature
+  lands in a v2 shard, guarding against a silent revert to v1.
--- a/docs/integrator/supply-chain-verification.md
+++ b/docs/integrator/supply-chain-verification.md
@@ -16,6 +16,18 @@ the [GitHub CLI](https://cli.github.com/) (`gh`), `crane` (recommended; `docker 
 `jq`, and — for in-cluster enforcement — `kubectl`. Binary and bundle
 verification (`aicr verify`) need only the `aicr` binary.
 
+**Cosign version.** AICR release signatures (the binary attestation and the
+signed recipe catalog) are recorded in **Rekor v2** as of the v2 cutover (see the
+release notes for the exact version). Verifying those bundles **with Cosign**
+requires **Cosign v3.0.1+**: older Cosign cannot parse a Rekor v2
+inclusion proof or its RFC3161 timestamp. `aicr verify` needs only the `aicr`
+binary and verifies both v1 and v2 transparently, so the Cosign floor does not
+apply to it. Releases published before the cutover are in Rekor v1 and verify
+with any recent Cosign. The verification *commands* are identical either way: a
+bundle self-describes which log it is in, so nothing in your workflow changes
+beyond the Cosign version. For why AICR signs to Rekor v2 and how the signing
+path works, see [Rekor v2 Signing](../contributor/rekor-v2-signing.md).
+
 Export the following variables once; the rest of this guide reuses them.
 Tags are mutable and can be repointed to a different image, so resolve the
 tag to an immutable `@sha256:` digest and verify against the digest.
--- a/pkg/bundler/attestation/kms.go
+++ b/pkg/bundler/attestation/kms.go
@@ -30,16 +30,32 @@ var _ Attester = (*KMSAttester)(nil)
 type KMSAttester struct {
 	keyURI   string
 	identity SigningIdentity
-	tlog     TransparencyPolicy
+
+	// target carries the transparency-target fields (RekorURL, SigningConfigPath,
+	// UseTUFSigningConfig; OIDC/Fulcio fields are unused for KMS). It is built by
+	// the shared SignOptionsFromResolve mapper — the same single mapping the
+	// keyless and evidence paths use — so a new signing-target field cannot be
+	// silently dropped on the KMS path (guarded by TestSignOptionsFromResolve).
+	target SignOptions
+
+	// tlog, when non-nil, overrides the transparency policy that Attest would
+	// otherwise compute from target. Used to inject a no-tlog policy for offline
+	// testing, and reserved for a future --tlog-upload=false opt-out (#409).
+	tlog TransparencyPolicy
 }
 
-// NewKMSAttester returns a KMSAttester for keyURI. Empty rekorURL falls back to
-// the Sigstore public-good Rekor default.
-func NewKMSAttester(keyURI, rekorURL string) *KMSAttester {
+// NewKMSAttester returns a KMSAttester for keyURI. Like keyless signing, KMS
+// signs to Rekor v2 by default (via the TUF-distributed signing config, which
+// also supplies the timestamp authority v2 requires) unless target opts out to a
+// Rekor v1 URL or a custom signing config. target is produced by
+// SignOptionsFromResolve so the KMS path shares the keyless path's signing-target
+// mapping. The transparency-log entry carries public-key verification material
+// (no Fulcio certificate). See #1650.
+func NewKMSAttester(keyURI string, target SignOptions) *KMSAttester {
 	return &KMSAttester{
 		keyURI:   keyURI,
 		identity: NewKMSIdentity(keyURI),
-		tlog:     NewRekorPolicy(rekorURL),
+		target:   target,
 	}
 }
 
@@ -55,7 +71,15 @@ func (k *KMSAttester) Attest(ctx context.Context, subject AttestSubject) ([]byte
 		return nil, errors.PropagateOrWrap(err, errors.ErrCodeInternal, "failed to build attestation statement")
 	}
 
-	res, err := SignStatementWith(ctx, statementJSON, k.identity, k.tlog)
+	tlog := k.tlog
+	if tlog == nil {
+		tlog, err = transparencyForOptions(ctx, k.target)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	res, err := SignStatementWith(ctx, statementJSON, k.identity, tlog)
 	if err != nil {
 		return nil, err
 	}
@@ -69,10 +93,11 @@ func (k *KMSAttester) Attest(ctx context.Context, subject AttestSubject) ([]byte
 // Identity returns the KMS key URI used for signing.
 func (k *KMSAttester) Identity() string { return k.keyURI }
 
-// HasRekorEntry reports whether produced attestations include a Rekor entry,
-// derived from the configured transparency policy. KMS signing uses Rekor by
-// default; this returns false only when paired with the no-tlog policy (the
-// offline/air-gapped path, #409).
+// HasRekorEntry reports whether produced attestations include a Rekor entry.
+// KMS signing records one by default (Rekor v2, or v1 via --rekor-url); it is
+// false only when an explicit no-tlog policy override is set (the offline path,
+// #409). A nil override (the production default) computes a real Rekor policy at
+// Attest time, so this correctly reports true.
 func (k *KMSAttester) HasRekorEntry() bool {
 	_, noTLog := k.tlog.(noTLogPolicy)
 	return !noTLog
--- a/pkg/bundler/attestation/signing.go
+++ b/pkg/bundler/attestation/signing.go
@@ -43,8 +43,39 @@ type SignOptions struct {
 	FulcioURL string
 
 	// RekorURL is the Rekor transparency log URL. Empty falls back to
-	// defaults.SigstoreRekorURL.
+	// defaults.SigstoreRekorURL. Ignored when SigningConfigPath or
+	// UseTUFSigningConfig is set (both take precedence).
 	RekorURL string
+
+	// SigningConfigPath, when non-empty, is the path to a Sigstore SigningConfig
+	// JSON file (e.g. the TUF-distributed signing_config_rekor_v2 target written
+	// to disk). It takes precedence over UseTUFSigningConfig and RekorURL and
+	// drives transparency-log and timestamp-authority selection via
+	// root.SelectServices — this is how AICR targets Rekor v2. See #1650.
+	SigningConfigPath string
+
+	// UseTUFSigningConfig selects the Rekor v2 signing config from the local TUF
+	// cache (populated by "aicr trust update"). It is the preferred, rotation-safe
+	// way to target Rekor v2 — the endpoint set is Sigstore-maintained and nothing
+	// is hardcoded. Ignored when SigningConfigPath is set. See #1650.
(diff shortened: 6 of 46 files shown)


pkg/bundler/attestation/kms.go
```diff
@@ -28,18 +28,32 @@ var _ Attester = (*KMSAttester)(nil)
...
-		keyURI:   keyURI,
-		identity: NewKMSIdentity(keyURI),
-		tlog:     NewRekorPolicy(rekorURL),
+		keyURI:              keyURI,
+		identity:            NewKMSIdentity(keyURI),
+		rekorURL:            rekorURL,
+		signingConfigPath:   signingConfigPath,
+		useTUFSigningConfig: useTUFSigningConfig,
 	}
 }
```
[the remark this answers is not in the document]
lockwobr: Considered this and I'm keeping the positional constructors here, for a couple of reasons:

- `NewKeylessAttester` / `NewKMSAttester` are internal and constructed only in the resolver (a handful of sites). Their signing-target fields (`rekorURL`, `signingConfigPath`, `useTUFSigningConfig`) are already funneled through the single `SignOptionsFromResolve` mapper that was added in the previous review round specifically to prevent the field-drift this PR guards against (a dropped field silently signing to the wrong Rekor).
- Functional options would make the resolver call sites more verbose without shrinking the field set, and grouping the three into a struct would re-introduce a second `ResolveOptions → signing` mapping surface competing with that single mapper.
- The parameter set is bounded and stable (Rekor v1 URL / custom signing config / TUF v2). Happy to switch to options if it grows beyond this.

pkg/trust/trust.go
```diff
@@ -104,6 +104,135 @@ func GetTrustedMaterial() (root.TrustedMaterial, error) {
...
 }
 
+// signingConfigTarget is the TUF target name for the Sigstore signing config
+// that routes signing to Rekor v2. AICR signs to Rekor v2 by default, so this is
+// the config the signer resolves unless the caller opts out to a Rekor v1 URL.
+// Sigstore distributes it alongside the default (v1) signing config; the
+// public-good default still points *other* clients at Rekor v1, so we consume
+// this v2 target explicitly. When Sigstore makes v2 the ecosystem default this
+// can move to the default "signing_config.v0.2.json" target. See NVIDIA/aicr#1650.
+const signingConfigTarget = "signing_config_rekor_v2.v0.2.json"
```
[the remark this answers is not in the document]
lockwobr: This one is a false positive — `signing_config_rekor_v2.v0.2.json` **is** published in the public-good TUF repo and resolves correctly. Two independent confirmations:

- A live `sigstore-go` TUF client (`tuf.New(DefaultOptions())` → `GetTarget("signing_config_rekor_v2.v0.2.json")` → `root.NewSigningConfigFromJSON`) returns the v2 config right now: `log2025-1.rekor.sigstore.dev` (v2), `rekor.sigstore.dev` (v1), and the `timestamp.sigstore.dev` TSA.
- This PR's e2e proves it end to end: signing reached `log2025-1.rekor.sigstore.dev` (v2, logIndex ~13M) in the KMS and keyless jobs — which is only possible if the v2 signing config was fetched from TUF.

On the cold-cache point: `GetSigningConfig` (ForceCache) intentionally misses when the cache hasn't been warmed, but `ResolveSigningConfig` falls back to a bounded network TUF fetch that succeeds (and `aicr trust update` / the release `generate-slsa-predicate` step warm it explicitly). So a cold cache does not cause a miss in practice.

(Note: the real signing e2e failures on this PR are unrelated — Rekor v2 DSSE `0.0.2` entries have no verification type in `rekor-tiles` yet, which is a separate upstream gap.)

pkg/trust/trust.go
```diff
@@ -104,6 +104,135 @@ func GetTrustedMaterial() (root.TrustedMaterial, error) {
...
+}
+
+// ResolveSigningConfig returns the Rekor v2 signing config, preferring the local
+// TUF cache (offline) and falling back to a bounded network TUF fetch when the
+// cache is cold. Because AICR signs to Rekor v2 by default, signers must succeed
+// without a prior explicit "aicr trust update"; signing is already an online
+// operation, so a one-time fetch of the endpoint config is acceptable and it
+// also warms the cache for next time. Bounded by defaults.TUFUpdateTimeout.
+func ResolveSigningConfig(ctx context.Context) (*root.SigningConfig, error) {
+	if sc, err := GetSigningConfig(); err == nil {
```
mchmarny, line 148: Cache-first here can produce a *silent* fall-back to Rekor v1 — the failure mode this PR otherwise fails closed on. `root.SelectServices` applies validity windows at sign time, so if a cached config's v2 shard entry passes its `validFor.end` (shard rotation) before the cache refreshes, selection quietly picks the still-valid v1 entry for any CLI user who hasn't run `trust update`. The `bundle-headless-oidc-ci` shard assertion only guards the CI path, not end users.

This is probably bounded in practice — `WithForceCache` still refreshes when the cached TUF timestamp metadata expires, so staleness is capped at that expiry — but worth (a) confirming that's actually the behavior of the pinned sigstore-go, and (b) a sentence in `rekor-v2-signing.md`'s operational notes on the worst-case downgrade window. Alternatively, log a warning when the selected service is v1 while `UseTUFSigningConfig` is set, so the downgrade is at least visible.

pkg/cli/bundle.go
```diff
@@ -398,6 +416,13 @@ func validateSigningKeyExclusivity(cmd *cli.Command, opts *bundleCmdOptions) err
 // values (non-empty endpoints that are not absolute https:// URLs). The
 // HTTPS check is shared with the config layer via config.ValidateHTTPSURL.
 func validateSigstoreEndpoints(opts *bundleCmdOptions) error {
+	// --rekor-url (a Rekor v1 endpoint) and --signing-config (a custom signing
+	// config file) are two different ways to opt out of the Rekor v2 default;
+	// reject the combination rather than silently picking one.
+	if opts.rekorURL != "" && opts.signingConfigPath != "" {
+		return errors.New(errors.ErrCodeInvalidRequest,
+			"--"+flagRekorURL+" and --"+flagSigningConfig+" are mutually exclusive")
```
mchmarny: The error message names two flags, but `--signing-config` also arrives via `AICR_SIGNING_CONFIG` — so a user with that env var set who passes only `--rekor-url` gets told two flags they never typed are mutually exclusive. The private-sigstore chainsaw test hit exactly this and had to `unset AICR_SIGNING_CONFIG`. Cheap fix: mention the env sources in the message, e.g. `"--rekor-url and --signing-config (or AICR_REKOR_URL / AICR_SIGNING_CONFIG) are mutually exclusive"`. Same applies to the copy in `runRecipeSignCatalogCmd`.

pkg/bundler/attestation/testdata/signing_config_v2_no_tsa.json
```diff
@@ -0,0 +1,40 @@
...
+    }
+  ],
+  "tsaUrls": [],
+  "rekorTlogConfig": {
+    "selector": "ANY"
+  },
+  "tsaConfig": {
+    "selector": "ANY"
+  }
+}
```
mchmarny: nit: missing trailing newline (the other two testdata configs have one).

mchmarny (commented):
Strong PR — the single `transparencyForOptions` selection shared by keyless and KMS, the `SignOptionsFromResolve` drift-guard mapper (with the test that caught the evidence-signs-v1 bug), the fail-closed v2-without-TSA rule, the fail-closed release hooks, and the CI assertion that the default actually lands in a v2 shard are all exactly the right shape, and library callers keep their v1 zero-value behavior. CI is fully green. Nothing blocks merge: one medium question on whether a stale TUF cache can silently downgrade end users to v1 after shard rotation (confirm + document, or warn on a v1 selection under the v2 default), one low UX item on the mutual-exclusivity error not naming its env-var sources, and one testdata nit.

pkg/bundler/attestation/signingconfig.go
```diff
@@ -0,0 +1,164 @@
...
+	usesV2 := false
+	for _, svc := range rekorSvcs {
+		logs = append(logs, sign.NewRekor(&sign.RekorOptions{
+			BaseURL: svc.URL,
+			Version: svc.MajorAPIVersion,
+		}))
+		if svc.MajorAPIVersion >= 2 {
+			usesV2 = true
+		}
+	}
```
njhensley, line 127: **🟡 Minor — Default/TUF path can silently select Rekor v1 with no "require-v2" assertion**

The public v2 signing config lists both a v2 and a v1 Rekor (selector ANY). Vendored `root.SelectServices` walks highest-version-first but returns the v1 service without error if the v2 shard isn't valid at `now` — the PR's own `TestSelectSigningServices` "falls back to v1 before v2 window" case proves it. On that fallback `usesV2=false`, this `usesV2 && len(tsas)==0` gate is skipped, and a valid v1 policy is returned silently. Triggers only on clock skew or a shard-rotation gap (low probability on NTP-synced CI), and fails safe (a valid signature in the wrong log, not a forgeable one). But nothing asserts a v2 log was actually chosen on the explicit-v2 paths, and the only shard-origin grep guards the `bundle --attest` chainsaw, not the release `cosign attest-blob` / `sign-catalog` outputs.

**Blast radius:** A release (or default-path signer) silently lands in Rekor v1 while everyone believes it's v2, defeating the release-monitoring purpose (#1650) with no error and no CI signal on the release artifacts.

**Fix:** On `NewSigningConfigPolicyFromTUF` + the `UseTUFSigningConfig` default only (not `--signing-config`, which may legitimately be v1), return an error when `!usesV2`; and add a v2-shard-origin assertion on the release attestation outputs.

pkg/bundler/attestation/kms.go
```diff
@@ -55,7 +69,19 @@ func (k *KMSAttester) Attest(ctx context.Context, subject AttestSubject) ([]byte
 		return nil, errors.PropagateOrWrap(err, errors.ErrCodeInternal, "failed to build attestation statement")
 	}
 
-	res, err := SignStatementWith(ctx, statementJSON, k.identity, k.tlog)
+	tlog := k.tlog
+	if tlog == nil {
+		tlog, err = transparencyForOptions(ctx, SignOptions{
```
njhensley: **🟡 Minor — KMS signing bypasses the SignOptionsFromResolve "no-dropped-field" guard**

The invariant "a signing field cannot be dropped by one caller" is enforced by the `SignOptionsFromResolve` mapper (guarded by `TestSignOptionsFromResolve`) for keyless + both evidence paths. KMS instead hand-copies the target fields positionally (ResolveOptions → `NewKMSAttester` args in resolver.go:170,198 → struct → this fresh `SignOptions{}` literal in `Attest`). `TestResolveAttester_KMS` only checks the returned type, not field propagation.

**Blast radius:** A future signing-target field added to the mapper is silently dropped on the KMS path with no failing test — the exact drift the mapper was introduced to prevent, left unguarded for KMS.

**Fix:** Route the KMS path through a shared `transparencyTargetFromResolve(ResolveOptions)` helper that `SignOptionsFromResolve` also calls, or extend the drift test to assert the KMS attester carries every target field.

pkg/cli/bundle.go
```diff
@@ -156,11 +168,17 @@ func parseBundleCmdOptions(cmd *cli.Command, cfg *appcfg.AICRConfig) (*bundleCmd
...
+		signingConfigPath:         cmd.String(flagSigningConfig),
 		insecureTLS:               boolFlagOrConfig(cmd, flagInsecureTLS, resolved.InsecureTLS),
 		plainHTTP:                 boolFlagOrConfig(cmd, flagPlainHTTP, resolved.PlainHTTP),
 		imageRefsPath:             stringFlagOrConfig(cmd, "image-refs", resolved.ImageRefs),
 	}
 
+	// Rekor v2 is the default for both keyless and KMS --attest signing: the
+	// TUF-distributed v2 signing config is used unless the caller opts out to a
+	// Rekor v1 URL or supplies a custom signing config file.
+	opts.useTUFSigningConfig = opts.rekorURL == "" && opts.signingConfigPath == ""
```
njhensley: **🟡 Minor — v2-default derivation + --rekor-url XOR --signing-config guard triplicated across CLI commands**

The same `useV2 = rekorURL=="" && signingConfigPath==""` derivation and the same mutual-exclusivity check are re-derived in bundle.go, recipe_sign_catalog.go, and (hardcoded true) validate_evidence.go. This is the exact drift class the library `SignOptionsFromResolve` mapper eliminated — re-created at the CLI layer.

**Blast radius:** A future policy change (a third target, or evidence gaining `--rekor-url`) must touch all three sites in lockstep; a missed site signs to the wrong log.

**Fix:** A single shared helper, e.g. `attestation.SigningTargetFromFlags(rekorURL, signingConfigPath) (useTUF bool, err error)`, that both computes the default and enforces exclusivity; call it from all three sites.

pkg/bundler/attestation/signing.go
```diff
@@ -101,10 +132,35 @@ func SignStatement(ctx context.Context, statementJSON []byte, opts SignOptions)
...
+func transparencyForOptions(ctx context.Context, opts SignOptions) (TransparencyPolicy, error) {
+	switch {
+	case opts.SigningConfigPath != "":
+		return NewSigningConfigPolicyFromPath(opts.SigningConfigPath)
+	case opts.UseTUFSigningConfig:
+		// Bound the Rekor v2 signing-config resolution (cache read plus a
+		// possible network TUF fetch on a cold cache) so an unbounded caller
+		// context cannot hang here, before SignStatementWith applies its own
+		// sign deadline. Shared by keyless SignStatement and KMSAttester.Attest.
+		tufCtx, cancel := context.WithTimeout(ctx, defaults.SigstoreSignTimeout)
```
njhensley, line 164: **🟡 Minor — Cold-cache TUF fetch shares the 2-minute signing deadline with the retry loop**

`transparencyForOptions` bounds the pre-sign TUF fetch with `WithTimeout(ctx, SigstoreSignTimeout)` where `ctx` is already the caller's sign context (evidence passes `signCtx = WithTimeout(ctx, EvidenceBundleSignTimeout)` = 2m). So the cold-cache fetch and the downstream `SignStatementWith` retry loop draw from one budget, undercutting `TestSigstoreRetryBudgetInvariant`'s premise that the retry loop owns nearly all of it.

**Blast radius:** Release CI pre-warms the cache via `trust update`, so only cold ad-hoc `evidence sign` / `validate --emit-attestation` is affected — a spurious first-sign timeout (fails closed, no wrong-log output).

**Fix:** Resolve the signing config on its own budget outside the sign context, and annotate `TestSigstoreRetryBudgetInvariant` that the v2 path may prepend a TUF fetch.

pkg/bundler/attestation/signing_targeting_test.go
```diff
@@ -0,0 +1,163 @@
...
+		{
+			name: "signing config file takes precedence over rekor URL",
+			opts: SignOptions{SigningConfigPath: "testdata/signing_config_v2.json", RekorURL: "https://rekor.example.com"},
+			want: "signingConfig",
+		},
+		{
+			name:      "TUF v2 config takes precedence over rekor URL",
+			opts:      SignOptions{UseTUFSigningConfig: true, RekorURL: "https://rekor.example.com"},
+			want:      "notRekor",
+			ignoreErr: true,
```
njhensley, line 60: **🟡 Minor — TestTransparencyForOptions makes an ungated live TUF call; its offline assertion is vacuous**

The "TUF v2 config" case drives real `trust.ResolveSigningConfig` (network) but is NOT behind `testing.Short()` like its sibling in trust_test.go, so `make test` can block up to 2m on a cold/offline runner. And with `ignoreErr: true`, on fetch failure `p==nil`, so the `if _, ok := p.(rekorPolicy); ok` "notRekor" check passes trivially — offline it proves nothing about the precedence it claims to cover.

**Blast radius:** Occasionally slow/flaky unit run; weaker-than-advertised precedence coverage offline.

**Fix:** Gate the case behind `testing.Short()` (or move to the integration test); rely on the deterministic `SigningConfigPath`-file case for the precedence assertion.

docs/integrator/supply-chain-verification.md
```diff
@@ -16,6 +16,18 @@ the [GitHub CLI](https://cli.github.com/) (`gh`), `crane` (recommended; `docker
 `jq`, and — for in-cluster enforcement — `kubectl`. Binary and bundle
 verification (`aicr verify`) need only the `aicr` binary.
 
+**Cosign version.** AICR release signatures (the binary attestation and the
+signed recipe catalog) are recorded in **Rekor v2** as of the v2 cutover (see the
+release notes for the exact version). Verifying those bundles **with Cosign**
+requires **Cosign v3.0.1+** (or v2.6.0+): older Cosign cannot parse a Rekor v2
```
njhensley: **🟡 Minor — Cosign version floor inconsistent across the three touched docs; "(or v2.6.0+)" unsourced**

Only this integrator guide says "v3.0.1+ (or v2.6.0+)"; cli-reference.md and rekor-v2-signing.md say bare "v3.0.1+". The v2.6.0 backport is uncorroborated anywhere in the PR (signer floor is >= v3.1.0 per workflow comments; .settings.yaml pins v3.1.1), and a 2.6.x cosign parsing v2 tiles + hashedrekord/PAE is dubious.

**Blast radius:** A consumer following the integrator guide could install cosign 2.6.x and hit a hard verify failure, then distrust a legitimately-signed release.

**Fix:** Pick one floor, use it verbatim in all three passages; drop the parenthetical unless a specific cosign release is cited. Also confirm v3.0.1 is genuinely the correct verify floor for the hashedrekord/PAE entries a v3.1.x signer writes (the PR only validated the v3.1.0 write floor).

docs/contributor/rekor-v2-signing.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Fail-closed rules
+
+- A v2 selection with no timestamp authority is rejected: a v2 bundle carries no
+  inline timestamp, so it would have no trusted time and could not be verified.
+- In the release pipeline, if signing is enabled (`SLSA_PREDICATE` set) but the
+  signing config path is unset, the goreleaser hooks error rather than let cosign
+  silently fall back to Rekor v1.
+
+### Release wiring
```
njhensley, line 129: **🟡 Minor — Contributor doc omits the load-bearing cosign >= v3.1.0 signer requirement**

This "Release wiring" section documents the `--signing-config` plumbing but never states that cosign < 3.1.0 silently writes the legacy `dsse` entry sigstore-go can't verify. That safety-critical constraint lives only in `.github/**` YAML comments.

**Blast radius:** A contributor editing the signing path from this doc could reintroduce the exact unverifiable-attestation bug this PR pins cosign to fix, caught only (if at all) by the bundle-headless-oidc-ci chainsaw guard.

**Fix:** Add one line to "Release wiring" stating the signer must be cosign >= v3.1.0 (pinned in .settings.yaml) so DSSE is logged as hashedrekord/PAE, cross-referencing #1650.

pkg/cli/bundle.go
```diff
@@ -732,10 +761,16 @@ Package with explicit tag (overrides CLI version):
 			},
 			&cli.StringFlag{
 				Name:     flagRekorURL,
-				Usage:    "Override the Rekor transparency-log URL for --attest keyless signing (e.g. a private Sigstore instance). Must be an absolute https:// URL with no embedded credentials. Defaults to the public-good Rekor. Also reads AICR_REKOR_URL.",
+				Usage:    "Sign --attest bundles to Rekor v1 at this URL instead of the Rekor v2 default (e.g. a private Sigstore instance, or the public-good v1 URL). Must be an absolute https:// URL with no embedded credentials. Also reads AICR_REKOR_URL.",
 				Sources:  cli.EnvVars("AICR_REKOR_URL"),
 				Category: catDeployment,
 			},
```
njhensley, line 781: **🟡 Minor — AICR_SIGNING_CONFIG auto-binds as --signing-config and collides with --rekor-url**

Because release CI exports `AICR_SIGNING_CONFIG` process-wide (generate-slsa-predicate), any later `aicr … --rekor-url …` in that job auto-picks the env as `--signing-config` and hard-errors "mutually exclusive" (guard at bundle.go:422) — the PR already had to `unset AICR_SIGNING_CONFIG` in the private-sigstore chainsaw test to work around it.

**Blast radius:** Fails closed (clear error, not a bad signature), so Minor — but a real footgun for future callers combining a private `--rekor-url` with an inherited env.

**Fix:** Document the interaction, or let an explicit flag `--rekor-url` take precedence over an env-sourced (vs flag-sourced) `--signing-config`.

pkg/bundler/attestation/signingconfig.go
```diff
@@ -0,0 +1,164 @@
...
+// #1650.
+type signingConfigPolicy struct {
+	logs []sign.Transparency
+	tsas []*sign.TimestampAuthority
+}
+
+// NewSigningConfigPolicyFromPath loads a Sigstore SigningConfig JSON file (e.g.
+// the TUF-distributed signing_config_rekor_v2 target written to disk) and
+// returns a TransparencyPolicy that signs to the services it selects for the
+// current time.
```
njhensley, line 54: **🔵 Nitpick — --signing-config file read via unbounded os.ReadFile**

`NewSigningConfigPolicyFromPath` → sigstore-go `root.NewSigningConfigFromPath` → bare `os.ReadFile` on the user path (follows symlinks, no size cap) — the CLAUDE.md anti-pattern the repo already avoids for the analogous `LoadTrustedMaterialFromFile`.

**Blast radius:** It's the operator's own local signer input (not adversarial verify material) and lives in vendored code — low practical impact.

**Fix:** Optionally bound it with `io.LimitReader` against a new `defaults.MaxSigningConfigBytes` + `root.NewSigningConfigFromJSON` (already used by `signingConfigFromClient`).

tests/chainsaw/signing/bundle-headless-oidc-ci/chainsaw-test.yaml
```diff
@@ -83,6 +83,17 @@ spec:
...
+              # The freshly signed bundle attestation must land in a Rekor v2
+              # shard (log<YEAR>-<rev>.rekor.sigstore.dev), not v1
+              # (rekor.sigstore.dev). Guards against the signing default silently
+              # reverting to Rekor v1. The regex is rotation-safe across yearly
+              # shards; validated against log2025-1.rekor.sigstore.dev. If a
+              # future shard adopts a non-numeric revision suffix, widen it here.
+              # See NVIDIA/aicr#1650.
+              ORIGIN="$(jq -r '.verificationMaterial.tlogEntries[0].inclusionProof.checkpoint.envelope' \
+                "${WORK}/b-flag/attestation/bundle-attestation.sigstore.json" | head -1)"
+              echo "${ORIGIN}" | grep -qE '^log[0-9]{4}-[0-9]+\.rekor\.sigstore\.dev'
```
njhensley, line 96: **🔵 Nitpick — v2-shard chainsaw assertion covers only the b-flag variant**

The `^log[0-9]{4}-…` origin grep guards the b-flag (identity-token-flag) invocation only; the b-env invocation gets just `test -s`. Same default signing code path, so single-path coverage is largely sufficient.

**Blast radius:** A regression affecting only the env path wouldn't be caught by the v2-shard guard.

**Fix:** Optionally duplicate the origin check to also cover the b-env invocation.

.github/actions/generate-slsa-predicate/action.yml
```diff
@@ -52,3 +52,30 @@ runs:
...
+        # from source (the built binaries do not exist yet). Requires Go, which
+        # every caller sets up before this action.
+        SC="${RUNNER_TEMP}/aicr-signing-config.json"
+        # Retry the TUF fetch (3 attempts, 5s/10s backoff) to absorb transient
+        # Sigstore TUF CDN blips, mirroring the goreleaser cosign hook — this is a
+        # hard prerequisite for the whole signing flow, which fails closed if the
+        # config is missing. The `-s` check treats an exit-0-but-empty-file result
+        # as a failure so a truncated write is retried, not silently exported.
+        for n in 1 2 3; do
+          GOFLAGS=-mod=vendor go run ./cmd/aicr trust update --emit-signing-config "${SC}" && [ -s "${SC}" ] && break
```
njhensley, line 71: **🔵 Nitpick — generate-slsa-predicate adds a mandatory network TUF fetch to PR-gating e2e jobs**

`go run … trust update --emit-signing-config` (3-retry, fail-closed) now runs in kms-ministack-e2e and sigstore-scaffolding-e2e. Both jobs already sign via cosign (already Sigstore-coupled) and skip on fork PRs, and the retry absorbs blips — acceptable.

**Blast radius:** A sustained Sigstore TUF outage now fails these checks slightly earlier than before (they'd have failed at the existing signing step anyway).

**Fix:** No change required; noting the added external dependency for awareness.

njhensley (commented):
# 🔎 Multi-Persona Review — Rekor v2 signing default

**Method:** 5 independent persona reviewers (Correctness · Supply-chain Security · Operability/CI-DX · Domain & Architecture · Docs/API-contract) → every finding adversarially re-derived from the resolved code + vendored sigstore-go by a senior meta-reviewer. Anchored to head `1d03d54`.

## Overall assessment

This is a carefully built, well-tested supply-chain change. The core v2-targeting logic is correct, the `v2-requires-TSA` invariant fails closed and is thoroughly unit-tested, the CI wiring is complete (every `outputs.cosign` reference resolves in-job, ordering + Go availability correct, all `install-aicr-release`/`cli-e2e` callers pass the new `cosign_version`, cosign pinned `v3.1.1` ≥ the `v3.1.0` floor), and the shared `SignOptionsFromResolve` mapper genuinely closes the "evidence silently signs v1" drift it was introduced for.

**No blocker or major finding survived adversarial verification.** The headline candidate — a silent v1 downgrade (inline finding on `signingconfig.go`) — is real but externally-gated, low-probability, and fails *safe* (a valid signature in the wrong log, not a broken/forgeable one): a hardening gap, not a bug. Everything else is Minor/Nitpick.

**Recommendation: Approve with comments.** The two highest-value fast-follows are the v1-downgrade assertion and routing KMS through the shared mapper — worth doing, but neither need block merge.

## Confirmed non-issues (examined, refuted or cleared)

- **KMS `HasRekorEntry()` nil-interface assertion** — comma-ok on a nil interface returns `false`, no panic; correctly reports `true`.
- **`ResolveSigningConfig` cold-cache goroutine** — buffered chan + select, identical to the existing `Update` pattern; no leak, TUF-integrity-verified fetch, errors propagate (no silent v1 fallback).
- **`selectSigningServices` swallowing the TSA error to `nil`** — safe: yields an empty TSA set, and a v2 selection with empty TSA is caught by the downstream fail-closed gate.
- **`//nolint:unparam` in resolver.go** — still accurate (lazy path returns nil error on all branches).
- **`sbom-and-attest` cosign "unpinned/orphaned"** — **refuted:** it *is* called via `attest-image-from-tag` (used in `attest-images.yaml`) and attests images via GitHub attestation, not the v2 signing-config path — out of scope.
- **Verify-time TSA trust** — handled by the public-good trusted root; informational only, no defect.
- **OpenAPI update for `--signing-config`** — not needed; signing is CLI/library-only, the `POST /v1/bundle` schema never exposed `rekorURL`.
- **`recipe sign-catalog` missing `--signing-config` in cli-reference** — correct; that command is `Hidden: true` / CI-only.

## Summary

| 🔴 Blocker | 🟠 Major | 🟡 Minor | 🔵 Nitpick | **Recommendation** |
| 0 | 0 | 8 | 3 | **Approve with comments** |

<sub>Generated by a multi-persona review harness (parallel persona reviewers + adversarial meta-reviewer). Inline comments carry each finding's mechanism, blast radius, and suggested fix.</sub>

pkg/trust/trust.go
lockwobr: Addressed in 741304f7. Added a `warnOnV1Downgrade` flag on the TUF path (`NewSigningConfigPolicyFromTUF`): when a stale cache selects a Rekor v1 service under the v2 default, it now emits a `slog.Warn` with the selected URL/API version, pointing the operator to `aicr trust update` — so the fall-back is visible rather than silent. A user-supplied `--signing-config` file that is legitimately v1 does not warn. I also documented the bounded worst-case window (capped at the TUF timestamp-metadata expiry, per the pinned sigstore-go `WithForceCache` refresh) in `rekor-v2-signing.md`'s operational notes.

pkg/cli/bundle.go
lockwobr: Good catch — this is exactly what the private-sigstore chainsaw test had to `unset AICR_SIGNING_CONFIG` around. Both mutual-exclusivity messages (bundle and `sign-catalog`) now name the env sources: `--rekor-url and --signing-config (or their env vars AICR_REKOR_URL / AICR_SIGNING_CONFIG) are mutually exclusive`. 741304f7.

pkg/bundler/attestation/testdata/signing_config_v2_no_tsa.json
lockwobr: Fixed — added the trailing newline to match the other two testdata configs. 741304f7.

pkg/bundler/attestation/signingconfig.go
lockwobr: Same silent-v1-downgrade concern @mchmarny raised. Addressed in 6f428663 by the `warnOnV1Downgrade` path in `NewSigningConfigPolicyFromTUF`: a v1 selection under the v2 default now logs a `slog.Warn` pointing to `aicr trust update`, so the fall-back is visible rather than silent. Failing fully closed would break signing during a legitimate shard-rotation gap (the fallback is a valid signature in the wrong log, not a forgeable one), so a visible warning is the tradeoff. Bounded worst-case window is now documented in `rekor-v2-signing.md`.

pkg/bundler/attestation/kms.go
lockwobr: Fixed in 6f428663. `KMSAttester` now carries its signing target as a `SignOptions` built by the shared `SignOptionsFromResolve` mapper (resolver passes `SignOptionsFromResolve("", opts)`), so the KMS path no longer hand-copies fields. `TestResolveAttester_KMS` now asserts `ka.target == SignOptionsFromResolve("", opts)`, so a new mapper field can't be silently dropped on the KMS path.

pkg/cli/bundle.go
lockwobr: Fixed in 6f428663. Extracted `signingTargetFromFlags(rekorURL, signingConfigPath) (useTUF bool, err error)` that both computes the v2 default and enforces the `--rekor-url`/`--signing-config` exclusivity; `bundle` and `recipe sign-catalog` both call it. (`validate --emit-attestation` exposes no such flags and hardcodes the TUF default.)

pkg/bundler/attestation/signing.go
lockwobr: Documented rather than refactored in 6f428663: added a budget note to `transparencyForOptions` and annotated `TestSigstoreRetryBudgetInvariant` that a cold-cache fetch draws from the caller budget before the retry loop. The full own-budget decoupling needs context-cancellation tradeoffs (detaching the fetch from the sign deadline also drops parent cancellation); for a fails-closed Minor that release CI pre-warms via `trust update`, the annotation felt proportionate. Happy to do the decoupling if you'd prefer it.

pkg/bundler/attestation/signing_targeting_test.go
lockwobr: Fixed in 6f428663. The live-TUF case is now tagged `network: true` and skipped under `testing.Short()`; the deterministic `SigningConfigPath` file cases carry the load-bearing precedence assertions.

docs/integrator/supply-chain-verification.md
lockwobr: Fixed in 6f428663. Dropped the unsourced `(or v2.6.0+)`; all three docs now state a consistent `v3.0.1+`. Kept v3.0.1 as the verify floor the other two docs already asserted (the PR validated the v3.1.0 *write* floor separately); flagging the deeper 'is v3.0.1 the exact verify floor' question as pre-existing, happy to pin it down in a follow-up if you want.

docs/contributor/rekor-v2-signing.md
lockwobr: Fixed in 6f428663. Added a 'the signer must be Cosign >= v3.1.0' paragraph to Release wiring (older Cosign writes the legacy `dsse` entry no sigstore-go can verify), and fixed a stale `go-build-release`->`generate-slsa-predicate` reference in the same section.

pkg/cli/bundle.go
lockwobr: 6f428663 names both env vars in the error (`AICR_REKOR_URL / AICR_SIGNING_CONFIG`) so the collision is diagnosable, and documents the interaction in `rekor-v2-signing.md` operational notes. Source-based precedence (explicit flag beats env-sourced value) is hard to do cleanly in urfave/cli v3 — `IsSet` doesn't distinguish flag vs env — so I went with the message + docs; the private-sigstore `unset AICR_SIGNING_CONFIG` is the documented workaround.

pkg/bundler/attestation/signingconfig.go
lockwobr: Fixed in 6f428663. `NewSigningConfigPolicyFromPath` now reads via `io.LimitReader` against a new `defaults.MaxSigningConfigBytes` (1 MiB) then `root.NewSigningConfigFromJSON`, mirroring `loadPEMPublicKey` and the trusted-root loader instead of sigstore-go's bare `os.ReadFile`.

tests/chainsaw/signing/bundle-headless-oidc-ci/chainsaw-test.yaml
lockwobr: Fixed in 6f428663. Duplicated the `^log[0-9]{4}-...` v2-shard origin assertion onto the b-env invocation.

.github/actions/generate-slsa-predicate/action.yml
lockwobr: Thanks — noting for awareness, no change. Both jobs already sign via cosign (already Sigstore-coupled) and skip on fork PRs, and the 3-attempt retry + empty-file guard absorb transient TUF blips.

.github/actions/install-aicr-release/action.yml
```diff
@@ -27,6 +27,9 @@ inputs:
   aicr-version:
     description: 'Released aicr version tag to install (e.g. v1.2.3).'
     required: true
+  cosign_version:
+    description: 'Cosign version (from .settings.yaml via load-versions; must be >= v3.1.0 to verify release provenance logged to Rekor v2 as hashedrekord/PAE).'
+    required: true
```
mchmarny, line 32: GitHub does not enforce `required: true` on composite-action inputs at runtime — a caller that omits `cosign_version` gets `cosign-release: ''`, which cosign-installer treats as "use my default". That's the exact silent drift this pin exists to prevent (the installer default lags the v3.1.0 floor per the comments this PR adds). Cheap fail-closed guard as the first step of the action:

```yaml
- name: Require cosign_version
  shell: bash
  run: '[ -n "${{ inputs.cosign_version }}" ] || { echo "cosign_version input is required" >&2; exit 1; }'
```

Same applies to the new input in `cli-e2e/action.yml`. Concrete near-miss: the in-flight uat-azure workflow calls this action without `cosign_version`, so whichever PR merges second would silently install the default cosign in that job.

pkg/cli/bundle.go
```diff
@@ -398,12 +420,33 @@ func validateSigningKeyExclusivity(cmd *cli.Command, opts *bundleCmdOptions) err
...
+// and the two are mutually exclusive. Either value can also arrive via
+// AICR_REKOR_URL / AICR_SIGNING_CONFIG, so the message names the env sources
+// rather than citing a flag the caller may never have typed. Centralized so a
+// future policy change lands in one place instead of drifting across per-command
+// copies.
+func signingTargetFromFlags(rekorURL, signingConfigPath string) (useTUF bool, err error) {
+	if rekorURL != "" && signingConfigPath != "" {
+		return false, errors.New(errors.ErrCodeInvalidRequest,
+			"--"+flagRekorURL+" and --"+flagSigningConfig+
+				" (or their env vars AICR_REKOR_URL / AICR_SIGNING_CONFIG) are mutually exclusive")
```
mchmarny: nit: `opts.rekorURL` can also arrive via `spec.bundle.rekorURL` in a `--config` file (`stringFlagOrConfig` in `parseBundleCmdOptions`), so a user with that config plus an inherited `AICR_SIGNING_CONFIG` env hits this error naming two flags/envs they never typed — same footgun shape as the env-var case this message already covers, one source deeper. Cheap fix: add `spec.bundle.rekorURL` to the message, e.g. `(or their env vars AICR_REKOR_URL / AICR_SIGNING_CONFIG, or spec.bundle.rekorURL in --config)`.

mchmarny (commented):
Verified all three of my earlier comments are properly fixed in 6f42866: the `warnOnV1Downgrade` path (with the stale-cache window documented in rekor-v2-signing.md and the warn branch under test), the centralized `signingTargetFromFlags` with env-source-naming error message, and the testdata newline. @njhensley's threads and CodeRabbit's findings all check out in the diff too — the KMS-through-the-shared-mapper unification with the `target` equality test is a particularly solid close.

Two new findings, neither blocking: a fail-closed guard for the unenforced `cosign_version` composite-action input (real cross-PR risk — the in-flight uat-azure workflow already omits it), and a nit on the exclusivity message omitting the `spec.bundle.rekorURL` config source. CI is fully green; the branch is behind main and needs a rebase before merge.

.github/actions/install-aicr-release/action.yml
lockwobr: Good catch on the runtime-enforcement gap — and it was live in-tree: `uat-azure.yaml` calls this action without `cosign_version`, so it would have silently installed the cosign-installer default. Fixed in aaa57927 three ways: (1) added `cosign_version` to the uat-azure caller, (2) added a fail-closed `Require cosign_version` guard as the first step of this action (env-var form to keep it injection-safe), and (3) the same guard in `cli-e2e/action.yml`. A caller that omits the input now errors loudly instead of drifting to the default.

pkg/cli/bundle.go
lockwobr: Fixed in aaa57927. Extended the shared message to name the config source too: `--rekor-url and --signing-config (or their env vars AICR_REKOR_URL / AICR_SIGNING_CONFIG, or spec.bundle.rekorURL in --config) are mutually exclusive`.

mchmarny (approved):
All 35 review threads are resolved, and I verified the fixes are actually in the head (33088eda) rather than trusting the thread state: the silent-v1-downgrade concern (raised twice) is addressed by `warnOnV1Downgrade` wired through `newSigningConfigPolicy` with the bounded-window docs; the triplicated CLI guard is consolidated into `signingTargetFromFlags` (bundle + sign-catalog); KMS now flows through the shared `SignOptionsFromResolve` mapper with the test asserting field propagation; the `--signing-config` read is size-bounded via `defaults.MaxSigningConfigBytes`; the mutual-exclusivity error names all three sources (flags, env vars, `spec.bundle.rekorURL`); the `cosign_version` fail-closed guards landed in both actions plus the live uat-azure caller that was silently omitting it; the live-TUF test case is short-skipped; the chainsaw v2-shard origin assertion covers both invocations; and the docs are consistent on v3.0.1+ verify / v3.1.0+ signer floors with the unsourced v2.6.0 claim dropped. CI is 14/15 green with one check still running — merge gates on it as usual. Nothing outstanding.

Issue #1666, opened 2026-07-08
helmfile: fresh-cluster apply fails diff on -post wrappers referencing parent-shipped CRDs


mchmarny:
### Bug Description

On a fresh cluster, `helmfile apply` of an AKS H100 bundle fails at diff time on the `network-operator-post` release:

```
Error: Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: resource mapping not found for name: "nic-cluster-policy" namespace: "" from "": no matches for kind "NicClusterPolicy" in version "mellanox.com/v1alpha1"
ensure CRDs are installed first
Error: plugin "diff" exited with error
```

The `-post` wrapper release (`004-network-operator-post`, containing [`nic-cluster-policy-aks.yaml`](https://github.com/NVIDIA/aicr/blob/038ae9cefc850b2b172d68bf40713566d8ddf5b6/recipes/components/network-operator/manifests/nic-cluster-policy-aks.yaml) wired in by [`recipes/overlays/aks.yaml#L67`](https://github.com/NVIDIA/aicr/blob/038ae9cefc850b2b172d68bf40713566d8ddf5b6/recipes/overlays/aks.yaml#L67)) creates a `NicClusterPolicy` CR whose CRD is shipped by its **parent** `network-operator` release. On first install the parent is not deployed yet, so helm-diff's live-mapper check cannot resolve the kind and the whole `helmfile apply` aborts before anything installs.

### Root Cause

#914 introduced the `hasSelfRefCRDs` registry flag, which emits `disableValidation: true` so helm-diff skips the live-mapper check for charts that reference CRDs they ship themselves. #929 then deliberately scoped the flag to the primary release only ([`pkg/bundler/deployer/helmfile/releases.go#L222`](https://github.com/NVIDIA/aicr/blob/038ae9cefc850b2b172d68bf40713566d8ddf5b6/pkg/bundler/deployer/helmfile/releases.go#L222), `f.Name == f.Parent`), on the assumption that injected `-pre`/`-post` wrapper folders contain only raw manifests with no self-reference of their own.

That assumption misses the **cross-release** case: a wrapper's manifests can create CRs of CRDs installed by the *parent* release. `needs:` ordering guarantees apply order, but helm-diff validates every release against the live cluster before anything is applied — so on a fresh cluster the wrapper's diff always fails.

[...]

PR #1686, opened 2026-07-09
feat(recipes): functional kueue with default quota CRs

ArangoGutierrez:
## Summary

Makes the kueue component functional in place: chart 0.17.1 → 0.18.2, default ResourceFlavor/ClusterQueue/LocalQueue quota CRs shipped via new registry-level `manifestFiles` defaults, an explicit trimmed integrations pin, and a health check that proves Kueue actually reconciles.

## Motivation / Context

Kueue is registered in the component registry, health-checked, and BOM'd — but referenced by zero overlays/mixins and ships no queue/quota CRs, so a deployed Kueue admits nothing out of the box (a "ghost component"). This PR fixes the component itself; composing it into recipes (a `platform-batch` mixin + standalone JobSet component) is a deliberate follow-up.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- **Registry-level `manifestFiles` defaults** (`ComponentConfig` + `ApplyRegistryDefaults` clone-fill): a component can ship default manifests wherever a recipe references it; ref-declared (overlay/mixin) lists take precedence, no merge. Applied post-merge in hydration, so overlay composition is unaffected.
- **Quota CRs** carry `helm.sh/hook` post-install/post-upgrade annotations (weights 1/5/10 encode ResourceFlavor→ClusterQueue→LocalQueue ordering) — required by `TestManifestHelmHooksRequired` for custom-apiVersion manifests. Upgrade semantics (delete-and-recreate can wedge on Kueue's `resource-in-use` finalizer; tuned quotas reset) are documented in the component catalog.
- **Integrations pin-and-trim:** the chart's default `controllerManagerConfigYaml` enables pod/deployment/statefulset/ray/etc. webhook interception; the pinned copy is byte-identical to the 0.18.2 default except `integrations.frameworks` trimmed to `batch/job`, `jobset.x-k8s.io/jobset`, `trainer.kubeflow.org/trainjob`.
- **Health check** adds assert-only steps (read-only allowlist compliant): ClusterQueue/LocalQueue `Active=True` via JMESPath condition projections, ResourceFlavor existence.

## Testing

```bash
make qualify   # rc=0: unit+race, 78.2% coverage, golangci-lint 0 issues, yamllint, 23/23 chainsaw e2e, grype
make bom-check # rc=0: 33 components, 89 image refs; committed BOM up to date
```

- Coverage delta `pkg/recipe`: 87.4% → 87.5% (+0.1%); new `ApplyRegistryDefaults` branch at 100%.
- Live Kind smoke: `helm install` with the pinned values (rc=0), the three CRs admitted by the webhook, ClusterQueue+LocalQueue `Active=True`, `make check-health COMPONENT=kueue` rc=0 (including the new quota-CR step), and a `kueue.x-k8s.io/queue-name`-labeled Job went QuotaReserved→Admitted→Finished end-to-end.

## Risk Assessment

**Rollout notes:** Component is referenced by zero overlays today, so no deployed recipe changes behavior. The new registry field is additive with precedence preserved. Upgrade/uninstall caveats for the hook-shipped CRs are documented in `docs/user/component-catalog.md`.

## Checklist


recipes/registry.yaml
```diff
@@ -549,10 +549,17 @@ components:
       - kueue
     healthCheck:
       assertFile: checks/kueue/health-check.yaml
+    # Default quota CRs so a referenced kueue admits workloads out of the
+    # box. Applied post-install (wave N+1) via the injected -post local
+    # chart, after the main release ships the CRDs.
+    manifestFiles:
+      - components/kueue/manifests/resource-flavor.yaml
```
mchmarny, line 580: These quota CRs instantiate kinds whose CRDs the kueue chart itself ships — which makes the injected `kueue-post` release exactly the #1666 shape: on a fresh cluster, helmfile bundles fail the wrapper's helm-diff REST-mapper check (`no matches for kind "ClusterQueue"`) because helmfile diffs every release in a level before applying any, and the hook annotations don't help (helm-diff renders hooks too). #1687 (in flight) adds `manifestsUseChartCRDs: true` for precisely this shape — network-operator and kubeflow-trainer carry it there. Once both PRs land, kueue's entry needs the flag too; whichever merges second should pick it up. Not a defect in this PR — nothing references kueue today — but the component won't survive its first helmfile-bundled reference without it.

recipes/checks/kueue/health-check.yaml
```diff
@@ -96,3 +98,34 @@ spec:
                 namespace: kueue-system
               status:
                 ((initContainerStatuses || `[]`)[?state.waiting.reason == 'CrashLoopBackOff' || state.waiting.reason == 'ImagePullBackOff' || state.waiting.reason == 'ErrImagePull' || state.waiting.reason == 'CreateContainerConfigError'] | length(@) > `0`): true
+    - name: validate-quota-crs-active
```
mchmarny, line 101: This step hard-asserts the *default* CR names, but the mechanism it depends on is override-able: a ref-declared `manifestFiles` list replaces the registry defaults, so an overlay shipping its own queue CRs (different names) makes the component health check fail on missing `default-flavor`/`cluster-queue` even though Kueue is healthy. Worth a comment here (and/or in the catalog) stating that overriding `manifestFiles` requires also overriding `healthCheck` asserts — or a follow-up to decouple the functional proof from the default names.

pkg/recipe/metadata.go
```diff
@@ -212,6 +213,14 @@ func (ref *ComponentRef) ApplyRegistryDefaults(config *ComponentConfig) {
 		}
 	}
 
+	// ManifestFiles: default from the registry only when the resolved ref
+	// carries none. Clone (not alias) so per-ref mutations cannot leak
+	// into the shared registry config (see slice-aliasing overlay-merge
+	// anti-pattern in CLAUDE.md).
+	if len(ref.ManifestFiles) == 0 && len(config.ManifestFiles) > 0 {
```
mchmarny, line 220: The clone-fill isn't conditioned on component type. `validateComponentRef` rejects Kustomize refs that carry `manifestFiles` ("either Kustomize (tag/path) or raw manifest files, not both"), so registry-level `manifestFiles` on a future Kustomize component would be injected here and then fail validation for *every* recipe referencing it — a config mistake that only surfaces at resolution time, in every consumer at once. Cheap insurance: reject the combination at registry load (mirroring the `TestLoadRegistry_RejectsReservedDeployerKey` pattern) or skip the fill for Kustomize-typed configs.

pkg/recipe/components.go
```diff
@@ -70,6 +70,15 @@ type ComponentConfig struct {
...
 	HealthCheck HealthCheckConfig `yaml:"healthCheck,omitempty"`
 
+	// ManifestFiles lists manifest files (relative to the recipes data
+	// root, e.g. "components/kueue/manifests/cluster-queue.yaml") bundled
+	// with the component whenever a recipe references it and the
+	// componentRef does not declare its own manifestFiles. A ref-declared
+	// (overlay/mixin) manifestFiles list takes precedence — these defaults
+	// are not merged into it. Consumers cannot opt out of the defaults
+	// except by declaring their own list.
+	ManifestFiles []string `yaml:"manifestFiles,omitempty"`
```
mchmarny, line 83: nit: the godoc is honest that there's no opt-out, but note *why*: `manifestFiles: []` on a ref is indistinguishable from absent (len == 0 → defaults filled), so an empty list silently gets the defaults back. Worth one line in `docs/contributor/component.md` too, since an integrator hitting this will look there first — the workaround is declaring a replacement list.

pkg/recipe/components_test.go
```diff
@@ -781,51 +782,11 @@ func TestComponentConfig_GetType(t *testing.T) {
...
-				Name: "test",
-				Kustomize: KustomizeConfig{
-					DefaultSource: "https://github.com/example/repo",
-					DefaultPath:   "deploy/production",
-					DefaultTag:    "v1.0.0",
-				},
-			},
-			expected: ComponentTypeKustomize,
-		},
+		{name: "nil config returns Helm", config: nil, expected: ComponentTypeHelm},
```
mchmarny: nit: this reformatting of `TestComponentConfig_GetType` is unrelated to the PR — it adds diff churn and diverges from the repo's multi-line table-test style (the CLAUDE.md example uses one field per line). Suggest reverting to keep the diff scoped to the kueue change.

recipes/components/kueue/manifests/resource-flavor.yaml
```diff
@@ -0,0 +1,37 @@
...
+    # kueue.x-k8s.io is a custom (CRD-backed) apiVersion — required by
+    # TestManifestHelmHooksRequired so this CR is applied post-install,
+    # after kueue's own chart has installed the ResourceFlavor CRD.
+    # Weight 1: no dependency on the other two kueue manifests here, but
+    # ClusterQueue references this flavor by name, so it applies first.
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "1"
+    "helm.sh/hook-delete-policy": before-hook-creation
+spec:
+  tolerations:
```
mchmarny, line 40: nit: since this is the single default flavor, the toleration gets injected into *every* admitted workload — including CPU-only jobs, which can then schedule onto `nvidia.com/gpu`-tainted nodes (the taint no longer repels them). Fine as an admission-neutral default, but worth a sentence in this comment so operators tuning quotas know CPU-only queues warrant a separate untolerated flavor.

mchmarny (commented):
Solid work making kueue functional in place — the quota-CR trio with hook-weight ordering, the trimmed integrations pin with its maintenance note, and the honest upgrade/uninstall caveats in the catalog are all the right shape, and the live Kind smoke (end-to-end Job admission through the default queue) is convincing. I verified the `ApplyRegistryDefaults` timing: hydration-time fill with clone semantics, serialized into saved recipes, so the saved-recipe → bundle path works.

Three medium comments (none blocking today since nothing references kueue yet): coordination with #1687's `manifestsUseChartCRDs` so the first helmfile-bundled reference doesn't hit the #1666 fresh-cluster diff failure, the health check's coupling to override-able default CR names, and a load-time guard for registry `manifestFiles` on Kustomize components. Three nits (opt-out doc, unrelated test reformatting, flavor-toleration scheduling note).

CI: the KWOK Summary failure is an infra flake — `Tier 1: aks-inference (flux-oci)` hit the 15-min job timeout and was cancelled (same class as seen on #1687); everything else is green. The branch is also `mergeable_state: dirty` — needs a rebase onto main before merge.

recipes/registry.yaml
ArangoGutierrez: Addressed — the branch is rebased onto main with #1687, and kueue's entry now carries `manifestsUseChartCRDs: true` with a comment pointing at the #914/#929 pattern (570da26eb).

recipes/checks/kueue/health-check.yaml
ArangoGutierrez: Documented in both places: a COUPLING note on the `validate-quota-crs-active` step and a matching sentence in the catalog entry (override `manifestFiles` → must also override `healthCheck.assertFile`). Kept the default-name asserts as the functional proof of the shipped CRs — a name-less any-match assert would decouple but weakens the proof to "some queue is Active"; decoupling properly is a follow-up if overriding becomes common.

pkg/recipe/metadata.go
ArangoGutierrez: Went with the load-time rejection: `GetComponentRegistryFor` now fails closed when a Kustomize-typed component declares `manifestFiles`, mirroring the reserved-deployer-key guard; covered by `TestLoadRegistry_RejectsKustomizeManifestFiles`. The config mistake surfaces at load instead of at resolution in every consumer.

pkg/recipe/components.go
ArangoGutierrez: Godoc now states the why — an empty ref-declared list is indistinguishable from absent after YAML decoding (len == 0 → defaults filled), so `manifestFiles: []` silently gets the defaults back — and the `manifestFiles` row in docs/contributor/component.md carries the same note plus the replacement-list workaround.

pkg/recipe/components_test.go
ArangoGutierrez: Reverted — `TestComponentConfig_GetType` is back to the multi-line table form; the hunk is gone from the diff against main.

recipes/components/kueue/manifests/resource-flavor.yaml
ArangoGutierrez: Added a TRADE-OFF paragraph to the manifest comment: as the single default flavor, the toleration is injected into every admitted workload including CPU-only jobs (the taint no longer repels them); operators tuning quotas for CPU-only queues should add a separate flavor without the toleration.

pkg/recipe/components_test.go
```diff
@@ -1200,3 +1161,29 @@ func TestLoadRegistry_RejectsReservedDeployerKey(t *testing.T) {
...
+				continue
+			}
+			if len(bytes.TrimSpace(data)) == 0 {
+				t.Errorf("component %q manifestFiles entry %q is empty", comp.Name, mf)
+			}
+			checked++
+		}
+	}
+	if checked == 0 {
+		t.Fatal("no registry manifestFiles entries were validated — expected at least kueue's quota CRs")
```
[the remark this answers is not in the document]
ArangoGutierrez: Sentinel is now Kueue-specific: the test asserts `registry.Get("kueue")` exists and its `manifestFiles` equal the three dependency-ordered quota CR paths; the generic readability sweep over all components stays.

mchmarny (dismissed):
That PR is in good shape now, @ArangoGutierrez, I approved, though at least one of the non-blocking CodeRabbit items may be worth closing in this or follow-up PR.

docs/user/component-catalog.md
```diff
@@ -34,7 +34,7 @@ The source of truth is [`recipes/registry.yaml`](https://github.com/NVIDIA/aicr/
...
+| **kueue** | Kubernetes-native job queuing system. Manages quotas and admits jobs for batch and AI workloads. Ships default quota CRs (ResourceFlavor `default-flavor`, ClusterQueue `cluster-queue`, LocalQueue `default` in the `default` namespace) so admission works out of the box — tune the ClusterQueue's nominal quotas to cluster capacity to enact real limits. Managed frameworks are pinned to batch/job, JobSet, and TrainJob. Upgrade note: the quota CRs are helm post-install/post-upgrade hooks with a delete-and-recreate policy — quiesce queues before upgrading the bundle (Kueue's resource-in-use finalizer on an active ClusterQueue/ResourceFlavor blocks the delete and can wedge the upgrade), and re-apply tuned quotas afterwards since upgrades reset them to the shipped defaults. Uninstalling leaves the hook-created CRs behind; delete them manually when removing Kueue. Overlays that override the component's `manifestFiles` (replacing the default quota CRs) must also override its health check — the shipped check asserts the default CR names above. | [Kueue](https://github.com/kubernetes-sigs/kueue) |
```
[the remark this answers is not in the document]
ArangoGutierrez: Checked this against the tree: there is no undeploy.sh in the repo — raw helm uninstall is the supported teardown, and the hook-created quota CRs do survive it (helm hooks are not chart-managed resources). The caveat as written matches actual behavior, so leaving it unscoped.

pkg/recipe/components_test.go
```diff
@@ -1200,3 +1201,77 @@ func TestLoadRegistry_RejectsReservedDeployerKey(t *testing.T) {
...
+	})
+	_, err := GetComponentRegistryFor(dp)
+	if err == nil {
+		t.Fatal("expected registry load to fail on Kustomize component with manifestFiles, got nil error")
+	}
+	if !strings.Contains(err.Error(), `"my-kustomize-app"`) {
+		t.Errorf("error %q does not name the offending component", err.Error())
+	}
+	if !strings.Contains(err.Error(), "manifestFiles") {
+		t.Errorf("error %q does not mention manifestFiles", err.Error())
```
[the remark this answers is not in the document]
ArangoGutierrez: Added the structured-code assertion in 1b84c6b6 — the test now fails if the code regresses from ErrCodeInvalidRequest.

pkg/recipe/metadata_test.go
```diff
@@ -2086,6 +2087,73 @@ func TestComponentRefApplyRegistryDefaults_HealthCheckAsserts(t *testing.T) {
...
+		}
+		ref := &ComponentRef{Name: "test-helm"}
+
+		ref.ApplyRegistryDefaults(config)
+
+		if len(ref.ManifestFiles) != 0 {
+			t.Errorf("ManifestFiles = %v, want empty", ref.ManifestFiles)
+		}
+	})
+}
```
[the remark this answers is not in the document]
ArangoGutierrez: Fixed in 5fa27841: the three uniform manifest-default cases are now table-driven, with defensive slice cloning kept as a separate subtest. The focused race test and full qualification pass.

recipes/checks/kueue/health-check.yaml
```diff
@@ -14,10 +14,12 @@
...
-# Validates that Kueue is running and healthy in the kueue-system namespace.
-# Checks that the kueue-controller-manager deployment has at least one
-# available replica and that no pods in the namespace are stuck in Pending,
-# Failed, or Unknown phases.
+# Validates that Kueue is running and healthy in the kueue-system
+# namespace, and that the default quota CRs shipped with the component
+# (ResourceFlavor default-flavor, ClusterQueue cluster-queue, LocalQueue
+# default/default) were admitted and reconciled to Active — a functional
+# proof (CRDs installed, controller reconciling, webhook admitting,
+# flavor resolving) within the assert/error-only allowlist.
```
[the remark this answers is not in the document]
ArangoGutierrez: Fixed in 5fa27841: the health-check description now claims only CRD installation, controller reconciliation, and flavor/queue resolution; it no longer claims webhook admission.

mchmarny (approved):
Makes kueue functional in place: chart bump to 0.18.2, default quota CRs via the new registry-level `manifestFiles` mechanism, trimmed integrations pin, and a health check that proves reconciliation (quota CRs Active) rather than just pod liveness. Reviewed the full diff: the `ApplyRegistryDefaults` clone-fill avoids slice aliasing and is covered by a dedicated mutation test, the Kustomize+manifestFiles combination fails closed at registry load with the error code asserted via `stderrors.Is`, the kueue-specific regression sentinel pins the exact quota CR paths, and the health-check/manifestFiles coupling plus upgrade-wedge and toleration trade-offs are all documented where an integrator will look. All prior review threads resolved, CI fully green, BOM and docs updated in-PR. No blockers.

Issue #929, opened 2026-05-15
refine(bundler): scope disableValidation: true to primary release, not -pre/-post wrapper folders


mchmarny:
## Summary

`pkg/bundler/deployer/helmfile/releases.go:216` keys the `disableValidation: true` flag off `f.Parent`, so injected `-pre` / `-post` wrapper folders inherit the flag from their primary component. Visible in goldens like `pkg/bundler/deployer/helmfile/testdata/mixed_gpu_operator/helmfile.yaml`:

```yaml
- name: gpu-operator-post
  namespace: gpu-operator
  chart: ./002-gpu-operator-post
  values:
    - ./002-gpu-operator-post/values.yaml
  needs:
    - gpu-operator
  disableValidation: true
```

`gpu-operator-post` here is a local-helm wrapper around a single RuntimeClass manifest. It doesn't ship the chart's `crds/`, doesn't reference any CRD that's not already registered by the predecessor `gpu-operator` release, and doesn't self-reference any kind. There's no operational reason to disable helm-diff's mapper sanity check on it.

## Impact

Low — the post-folder loses helm-diff's REST-mapper validation, so a typo in a resource kind in a pre/post manifest would render successfully at diff time and fail at apply time. Not a functional regression vs. main, but a small loss of defense-in-depth.

## Options

1. **Scope the flag to the primary release only.** Track folder kind in the per-release decision so only `KindUpstreamHelm` (or whichever shape carries the actual self-ref CRD) gets `disableValidation: true`.
2. **Keep the cascade but document why.** Today the inline comment says "Keyed by f.Parent so injected -pre / -post folders inherit the same flag." That reads as accidental coupling. If the cascade is intentional, the comment should explain *why* we want it (e.g., a self-ref CRD in the primary chart's `crds/` could be referenced by a post-manifest in some scenarios).

## Acceptance Criteria

- [ ] Pick option 1 or 2.
- [ ] If option 1: update `releases.go` and regenerate goldens (`go test -update ./pkg/bundler/deployer/helmfile/...`).
- [ ] If option 2: rewrite the inline comment at `releases.go:209-215` to explain the cascade rationale.

## Related

- PR #926
- Review comment: https://github.com/NVIDIA/aicr/pull/926#discussion_r3250796753

PR #1687, opened 2026-07-09
fix(bundler): emit disableValidation on CRD-dependent -post wrappers

### LINKED ISSUE #1666: helmfile: fresh-cluster apply fails diff on -post wrappers referencing parent-shipped CRDs

mchmarny:
## Summary

Helmfile bundles now emit `disableValidation: true` on an injected `-post` wrapper release when its manifests instantiate CRs of CRDs the parent chart installs, via a new registry flag `manifestsUseChartCRDs` (set on `network-operator`). Fixes fresh-cluster `helmfile apply` failing with `no matches for kind "NicClusterPolicy" in version "mellanox.com/v1alpha1"`.

## Motivation / Context

The AKS overlay attaches `nic-cluster-policy-aks.yaml` (a `NicClusterPolicy` CR) as `manifestFiles` on `network-operator`; the bundler wraps it into the injected `network-operator-post` release. That wrapper shares the parent's sub-helmfile level, and helmfile diffs every release in a level before applying any — the `needs:` edge orders apply, not diff — so on a fresh cluster helm-diff's REST-mapper check runs before the parent has installed the CRDs and can never pass. The existing `disableValidation` mitigation (#914) is deliberately scoped to primary releases (#929), which didn't account for wrapper manifests depending on parent-shipped CRDs.

Fixes: #1666
Related: #914, #929

## Type of Change

## Component(s) Affected

## Implementation Notes

- New registry flag `manifestsUseChartCRDs` mirrors `hasSelfRefCRDs` (#914) and is plumbed through the same `componentFlags` path; emission is scoped to `f.Name == f.Parent+"-post"` so primary releases and `-pre` wrappers (whose manifests apply before the chart and cannot depend on its CRDs) keep helm-diff's mapper typo-check per #929.
- `make bom-docs` run: no change (comment/flag-only registry edit).

## Testing

```bash
make qualify   # PASS
```

- New `TestBuildHelmfile_DisableValidationPostManifests` pins pre/primary/post emission (written first, red on the missing field).
- Coverage: `pkg/bundler/deployer/helmfile` 85.6% → 86.9% (+1.3%).
- Live verification on a fresh AKS cluster (`aicr-test1`, K8s 1.35.5) that reproduced #1666: with the flag emitted, the previously-fatal `network-operator-post` diff passes, `helmfile apply` completes (exit 0), and the `NicClusterPolicy` CR is created; regenerating the bundle from the same recipe with the patched binary emits `disableValidation: true` on `network-operator-post` automatically.

## Risk Assessment

**Rollout notes:** Only affects newly generated helmfile bundles for recipes whose components carry the new flag (today: `network-operator` with attached manifests, i.e. AKS). Existing bundles are unchanged; regenerate to pick up the fix.

## Checklist

### DIFF
--- a/pkg/bundler/deployer/helmfile/releases.go
+++ b/pkg/bundler/deployer/helmfile/releases.go
@@ -95,15 +95,15 @@ type Release struct {
 	// (true) applies otherwise. Pointer so a deliberate false survives
 	// YAML marshaling and a nil value omits the field entirely.
 	CreateNamespace *bool `yaml:"createNamespace,omitempty"`
-	// DisableValidation is emitted when the chart self-references its
-	// own CRDs at diff time (registry flag HasSelfRefCRDs). helm-diff
-	// will skip the live-cluster REST-mapper check for this release,
-	// letting render succeed when a template creates a CR whose CRD
-	// is shipped under `crds/` in the same chart and not yet present
-	// in the cluster. Scoped to the primary release only — injected
-	// -pre / -post wrapper folders carry raw manifests, not the
-	// self-ref chart, so they keep the safety check. Issues #914,
-	// #929.
+	// DisableValidation tells helm-diff to skip the live-cluster
+	// REST-mapper check for this release, letting render succeed when
+	// a manifest creates a CR whose CRD is not yet present in the
+	// cluster. Emitted in two registry-driven cases: on the primary
+	// release when the chart self-references its own CRDs
+	// (HasSelfRefCRDs), and on the injected -post wrapper when its
+	// manifests instantiate CRs of CRDs the parent chart installs
+	// (ManifestsUseChartCRDs). All other wrappers keep the safety
+	// check. Issues #914, #929.
 	DisableValidation bool `yaml:"disableValidation,omitempty"`
 }
 
@@ -220,6 +220,24 @@ func buildHelmfile(folders []localformat.Folder, namespaceByComponent map[string
 		// the wrapper's manifests — a typo in a resource Kind would
 		// render at diff time and fail at apply. See issue #929.
 		if f.Name == f.Parent && flags[f.Parent].HasSelfRefCRDs {
+			rel.DisableValidation = true
+		}
+		// ManifestsUseChartCRDs is the post-manifest counterpart: the
+		// component's attached manifests instantiate CRs whose CRDs the
+		// component's own chart installs. The manifest-bearing folder
+		// shares the parent's DAG level, and helmfile diffs every
+		// release in a level before applying any — the `needs:` edge
+		// orders apply, not diff — so on a fresh cluster its mapper
+		// check can never pass (network-operator's NicClusterPolicy on
+		// AKS; kubeflow-trainer's ClusterTrainingRuntime). Keyed off
+		// Folder.CarriesPostManifests, which both the injected -post
+		// wrapper and the collapsed vendored mixed folder set — a
+		// release-name suffix check would silently miss the vendored
+		// layout, where Name == Parent. Primaries without post
+		// manifests and -pre wrappers (whose manifests apply before
+		// the chart and cannot depend on its CRDs) keep the mapper
+		// sanity check per issue #929.
+		if f.CarriesPostManifests && flags[f.Parent].ManifestsUseChartCRDs {
 			rel.DisableValidation = true
 		}
 
--- a/recipes/registry.yaml
+++ b/recipes/registry.yaml
@@ -125,6 +125,17 @@ components:
     displayName: network-operator
     valueOverrideKeys:
       - networkoperator
+    # Overlays attach NicClusterPolicy CRs (mellanox.com API group) as
+    # manifestFiles — e.g. aks.yaml's nic-cluster-policy-aks.yaml —
+    # which the bundler wraps into the injected network-operator-post
+    # release. Those CRDs are installed by this chart at apply time,
+    # but the wrapper shares the parent's sub-helmfile level and
+    # helm-diff runs before anything in the level applies, so on a
+    # fresh cluster the wrapper's REST-mapper check can never pass
+    # ("no matches for kind NicClusterPolicy"). This flag emits
+    # disableValidation: true on the -post wrapper release only.
+    # See issues #914 and #929 for the pattern.
+    manifestsUseChartCRDs: true
     healthCheck:
       assertFile: checks/network-operator/health-check.yaml
     helm:
@@ -574,6 +585,15 @@ components:
     # on this release tells helm-diff to skip the live-mapper check.
     # See https://github.com/NVIDIA/aicr/issues/914.
     hasSelfRefCRDs: true
+    # recipes/mixins/platform-kubeflow.yaml additionally attaches
+    # torch-distributed-cluster-training-runtime.yaml (a
+    # ClusterTrainingRuntime CR of the CRD this chart ships) as
+    # manifestFiles, which the bundler wraps into the injected
+    # kubeflow-trainer-post release (or the chart's own vendored
+    # wrapper under --vendor-charts). hasSelfRefCRDs above covers only
+    # the primary release (#929); this flag extends the helm-diff
+    # bypass to the manifest-bearing folder. See #1666.
+    manifestsUseChartCRDs: true
     healthCheck:
       assertFile: checks/kubeflow-trainer/health-check.yaml
     helm:
--- a/docs/contributor/component.md
+++ b/docs/contributor/component.md
@@ -85,7 +85,7 @@ operator:
 - `healthCheck.assertFile:` — chainsaw conformance assertions ([validator.md](validator.md))
 - `storageClassPaths:` — where `--storage-class` is injected
 - `podScheduling.workload.workloadSelectorPaths` — for workload-pod placement
-- `gkeCriticalPriority`, `hasSelfRefCRDs` — narrow service-specific quirks (see godoc on `ComponentConfig` for when these apply)
+- `gkeCriticalPriority`, `hasSelfRefCRDs`, `manifestsUseChartCRDs` — narrow service-specific quirks (see godoc on `ComponentConfig` for when these apply)
 
 **4. Run `make bom-docs`** and commit the regenerated
 `docs/user/container-images.md` in the same PR. CLAUDE.md treats this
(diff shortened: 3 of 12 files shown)


pkg/bundler/deployer/helmfile/releases.go
```diff
@@ -222,6 +222,19 @@ func buildHelmfile(folders []localformat.Folder, namespaceByComponent map[string
...
+		// wrapper's manifests instantiate CRs whose CRDs the PARENT
+		// chart installs. The wrapper shares the parent's DAG level,
+		// and helmfile diffs every release in a level before applying
+		// any — the `needs:` edge orders apply, not diff — so on a
+		// fresh cluster the wrapper's mapper check can never pass
+		// (network-operator's NicClusterPolicy on AKS). Scoped to the
+		// -post wrapper only: the primary chart and -pre wrappers
+		// (whose manifests apply before the chart and cannot depend on
+		// its CRDs) keep the mapper sanity check per issue #929.
+		if f.Name == f.Parent+"-post" && flags[f.Parent].ManifestsUseChartCRDs {
```
yuanchen8911: This condition can't fire for vendored bundles, so the --vendor-charts path keeps the exact failure this PR fixes.

Under --vendor-charts, mixed components never split into primary + -post: writer.go gates post-injection on !opts.VendorCharts ("mixed collapses into the primary"), and the vendored wrapper folder is written by vendor_folder.go with Name == Parent == component name, with the attached manifests rendered into the wrapper chart's templates/ as post-install hooks (vendor_wrapper.go). So neither this branch nor the hasSelfRefCRDs branch matches, and the vendored network-operator release is emitted without disableValidation. On a fresh cluster, helm-diff then renders the NicClusterPolicy CR from the wrapper's templates while its CRD is still inside the vendored subchart tarball — the same "no matches for kind NicClusterPolicy" failure, in the layout air-gapped installs (vendoring's main use case) actually use. The helmfile deployer supports vendoring (bundler.go passes VendorCharts into the helmfile Generator), so this is a reachable, supported combination.

Suggested fix: key the decision off the generated folder shape instead of the release-name suffix — e.g. a localformat.Folder field marking "this folder carries the component's post-manifests", set by both the -post injection writer and the vendored mixed-wrapper writer, with the emission here becoming flags[f.Parent].ManifestsUseChartCRDs && f.<marker>. That keeps -pre wrappers and non-mixed primaries on the #929 safety check in both layouts, and a vendored-mode test mirroring TestBuildHelmfile_DisableValidationPostManifests would pin it.

If you'd rather keep this PR release-name-scoped to land the AKS fix quickly, that's workable too — but then the vendored gap should be documented on the flag's godoc as a known limitation and tracked in a follow-up issue, since the flag is otherwise a silent no-op under a supported CLI flag.

recipes/registry.yaml
```diff
@@ -125,6 +125,17 @@ components:
...
+    # manifestFiles — e.g. aks.yaml's nic-cluster-policy-aks.yaml —
+    # which the bundler wraps into the injected network-operator-post
+    # release. Those CRDs are installed by this chart at apply time,
+    # but the wrapper shares the parent's sub-helmfile level and
+    # helm-diff runs before anything in the level applies, so on a
+    # fresh cluster the wrapper's REST-mapper check can never pass
+    # ("no matches for kind NicClusterPolicy"). This flag emits
+    # disableValidation: true on the -post wrapper release only.
+    # See issues #914 and #929 for the pattern.
+    manifestsUseChartCRDs: true
```
yuanchen8911, line 138: kubeflow-trainer has the same shape and needs this flag too.

recipes/mixins/platform-kubeflow.yaml attaches components/kubeflow-trainer/manifests/torch-distributed-cluster-training-runtime.yaml as manifestFiles on kubeflow-trainer, which the bundler wraps into kubeflow-trainer-post. That manifest instantiates a ClusterTrainingRuntime CR whose CRD ships in the kubeflow-trainer chart's crds/. The existing hasSelfRefCRDs: true on kubeflow-trainer covers only the primary release, deliberately per #929, not the wrapper. So on a fresh cluster, any training bundle that includes platform-kubeflow fails the wrapper's helm-diff REST-mapper check exactly like network-operator-post does here; the AKS deploy just never got far enough to hit it.

One-line fix: add manifestsUseChartCRDs: true to the kubeflow-trainer entry (with a comment pointing at platform-kubeflow's manifestFiles), and ideally extend the new test with a kubeflow-trainer-style case so the flag's intended coverage is pinned.

yuanchen8911 (changes requested):
Two gaps in the flag's coverage, detailed inline: the --vendor-charts layout collapses post manifests into the primary wrapper so the -post release-name condition never fires, and kubeflow-trainer has the same post-manifest/parent-CRD shape via platform-kubeflow but doesn't get the flag. The mechanism and scoping of the change itself look right (the -post-only emission, #929 preservation on primaries and -pre wrappers, and the test).

pkg/bundler/deployer/helmfile/releases.go
yuanchen8911: Rechecked current head fbda78f after the merge-from-main update. This still applies: releases.go still emits the new flag only when f.Name == f.Parent+"-post", while the vendored mixed wrapper is Name == Parent and there is still no folder marker for carried post manifests. Leaving this unresolved.

recipes/registry.yaml
yuanchen8911: Rechecked current head fbda78f. This still applies: kubeflow-trainer still has only hasSelfRefCRDs and does not set manifestsUseChartCRDs, even though platform-kubeflow attaches the ClusterTrainingRuntime manifest as a post manifest. Leaving this unresolved.

pkg/bundler/deployer/helmfile/releases.go
mchmarny: Verified and implemented the folder-marker approach in 49f5da55 — you're right that the suffix check was a silent no-op under `--vendor-charts` (confirmed against `writer.go`'s `!opts.VendorCharts` post-injection gate and `vendor_folder.go`'s `Name == Parent` wrapper).

- New `localformat.Folder.CarriesPostManifests`, set by `injectAuxiliaryFolder` for `phasePost` and by `writeVendoredHelmFolder` when the collapsed wrapper embeds manifests as hook templates.
- Emission is now `f.CarriesPostManifests && flags[f.Parent].ManifestsUseChartCRDs`; `-pre` wrappers and manifest-less primaries keep the #929 mapper check in both layouts.
- Pinned by `TestBuildHelmfile_DisableValidationVendoredMixed` (collapsed mixed folder gets the bypass, vendored pure-Helm sibling doesn't) plus marker assertions in `TestWrite_VendorCharts_Mixed` / `TestWrite_VendorCharts_PureHelm` and the injected-wrapper writer test, so both writers' marker behavior is covered, not just the helmfile consumer.

recipes/registry.yaml
mchmarny: Confirmed — `platform-kubeflow.yaml` attaches the `ClusterTrainingRuntime` manifest whose CRD ships in the trainer chart's `crds/`, same shape exactly. Fixed in 49f5da55:

- `manifestsUseChartCRDs: true` on the kubeflow-trainer registry entry with a comment pointing at the mixin's `manifestFiles`.
- `TestBuildHelmfile_DisableValidationBothFlags` pins the dual-flag shape: `hasSelfRefCRDs` covers the primary, the new flag covers the `-post` wrapper, each scoped to its own release.
- Verified end-to-end: a regenerated `eks/h100/training/kubeflow` bundle emits `disableValidation: true` on both `kubeflow-trainer` and `kubeflow-trainer-post`.

yuanchen8911 (approved):

PR #1689, opened 2026-07-09
fix(recipes): pin toolkit containerd config source to file on AKS

### LINKED ISSUE #1688: gpu-operator toolkit writes v4 containerd drop-in on AKS, kills containerd
## Symptom

Deploying the stock AKS recipe bundle on AKS Ubuntu 24.04 GPU nodes (containerd 2.x; 2.3.2 on current node images) takes both GPU nodes `NotReady` shortly after the gpu-operator container-toolkit daemonset starts. All pods on the node fail with `no runtime for "nvidia" is configured`, then `dial unix /run/containerd/containerd.sock: connect: connection refused`. On the node:

```
containerd: drop-in config version 4 higher than root config version 2
containerd.service: Main process exited, code=exited, status=1/FAILURE
```

## Root cause

nvidia-container-toolkit v1.19.1 (as shipped by gpu-operator v26.3.2) defaults its containerd config source to `["command", "file"]`. The `command` loader runs `containerd config dump`, which returns the **migrated effective config** — `version = 4` on containerd 2.x — while AKS's on-disk `/etc/containerd/config.toml` declares `version = 2`. The toolkit then writes `/etc/containerd/conf.d/99-nvidia.toml` with `version = 4`; containerd refuses to merge a drop-in with a higher version than the root config and exits, killing the node's runtime (and taking kubelet down with it).

## Fix

Set the toolkit env `RUNTIME_CONFIG_SOURCE=file` in the AKS gpu-operator values so the toolkit reads the on-disk config (version 2) and emits a matching version-2 drop-in. Verified live on AKS `aicr-test1` (K8s 1.35.5, Ubuntu 24.04, containerd 2.3.2): drop-in regenerates as `version = 2`, containerd stays up, full gpu-operator stack converges, `nvidia.com/gpu: 8` allocatable.

Toolkit source refs (v1.19.1): default `ConfigSources` in `cmd/nvidia-ctk-installer/container/runtime/runtime.go` (env `RUNTIME_CONFIG_SOURCE`), version taken from the source loader in `pkg/config/engine/containerd/containerd.go` (`parseVersion`), `CommandLineSource` = `containerd config dump`.

mchmarny:
## Summary

Sets `RUNTIME_CONFIG_SOURCE=file` on the gpu-operator container-toolkit in both AKS values variants, so the toolkit's containerd drop-in matches the on-disk root config version instead of the migrated (`containerd config dump`) version that kills containerd on AKS Ubuntu 24.04.

## Motivation / Context

nvidia-container-toolkit v1.19.1 (shipped by gpu-operator v26.3.2) defaults its containerd config source to `["command", "file"]`; the `command` loader returns the migrated effective config — `version = 4` on containerd 2.2 — while AKS's on-disk `/etc/containerd/config.toml` declares `version = 2`. The generated `/etc/containerd/conf.d/99-nvidia.toml` then carries a higher version than the root config, containerd fails to start (`drop-in config version 4 higher than root config version 2`), and every GPU node goes NotReady with kubelet down. Toolkit refs (v1.19.1): default `ConfigSources` in `cmd/nvidia-ctk-installer/container/runtime/runtime.go` (env `RUNTIME_CONFIG_SOURCE`), version derived from the source loader in `pkg/config/engine/containerd/containerd.go`.

Fixes: #1688

## Type of Change

## Component(s) Affected

- [x] Recipe engine / data (`recipes/` components, overlays, values)

## Implementation Notes

- The env is added to **both** `values-aks.yaml` and `values-aks-training.yaml` because `aks-training.yaml` *replaces* (does not chain) the componentRef's `valuesFile`.
- Scoped to AKS, where the mismatch is reproduced and the fix verified; other services' node images have not exhibited the mismatch (nightly UAT green on EKS/GKE with the same gpu-operator pin).
- `make bom-docs` run: no change (env-only values edit).

## Testing

```bash
make qualify   # PASS
```

- Hydration verified for both AKS recipe shapes: `aicr query --service aks --accelerator h100 --intent {training,inference} --os ubuntu --selector components.gpu-operator.values.toolkit` returns the env in both.
- Live verification on AKS `aicr-test1` (K8s 1.35.5, Ubuntu 24.04, containerd 2.2), which reproduced #1688 with both GPU nodes NotReady: with `RUNTIME_CONFIG_SOURCE=file` applied via ClusterPolicy, the toolkit regenerates the drop-in as `version = 2`, containerd stays up, the full gpu-operator stack converges (validators Completed), and `nvidia.com/gpu: 8` is allocatable on both nodes.

## Risk Assessment

**Rollout notes:** Affects newly generated AKS bundles only. Clusters already broken by the v4 drop-in additionally need the stale `/etc/containerd/conf.d/99-nvidia.toml` removed and containerd/kubelet restarted before the corrected toolkit pod can run (documented in #1688).

## Checklist

### DIFF
--- a/recipes/components/gpu-operator/values-aks.yaml
+++ b/recipes/components/gpu-operator/values-aks.yaml
@@ -68,3 +68,17 @@ driver:
 #   serviceMonitor:
 #     additionalLabels:
 #       release: kube-prometheus  # Must match kube-prometheus-stack Helm release name
+
+# AKS Ubuntu 24.04 GPU nodes run containerd 2.x (2.3 on current node
+# images) with a version=2 root config at /etc/containerd/config.toml. The container-toolkit's default
+# config source ("command" = `containerd config dump`) returns the
+# MIGRATED effective config (version=4), so the generated
+# /etc/containerd/conf.d/99-nvidia.toml declares a higher version than
+# the root file and containerd refuses to start ("drop-in config
+# version 4 higher than root config version 2"), taking the node
+# NotReady. Reading the on-disk file instead emits a matching
+# version=2 drop-in. See issue #1688.
+toolkit:
+  env:
+    - name: RUNTIME_CONFIG_SOURCE
+      value: file
--- a/recipes/components/gpu-operator/values-aks-training.yaml
+++ b/recipes/components/gpu-operator/values-aks-training.yaml
@@ -40,3 +40,17 @@ validator:
     env:
       - name: WITH_WORKLOAD
         value: "false"
+
+# AKS Ubuntu 24.04 GPU nodes run containerd 2.x (2.3 on current node
+# images) with a version=2 root config at /etc/containerd/config.toml. The container-toolkit's default
+# config source ("command" = `containerd config dump`) returns the
+# MIGRATED effective config (version=4), so the generated
+# /etc/containerd/conf.d/99-nvidia.toml declares a higher version than
+# the root file and containerd refuses to start ("drop-in config
+# version 4 higher than root config version 2"), taking the node
+# NotReady. Reading the on-disk file instead emits a matching
+# version=2 drop-in. See issue #1688.
+toolkit:
+  env:
+    - name: RUNTIME_CONFIG_SOURCE
+      value: file


recipes/components/gpu-operator/values-aks.yaml
```diff
@@ -68,3 +68,17 @@ driver:
 #   serviceMonitor:
 #     additionalLabels:
 #       release: kube-prometheus  # Must match kube-prometheus-stack Helm release name
+
+# AKS Ubuntu 24.04 GPU nodes run containerd 2.2 with a version=2 root
```
yuanchen8911: Version nit: aicr-test1 nodes (where this was reproduced and verified) run containerd 2.3.2-1, not 2.2 (kubectl get nodes -o wide). Same text in the values-aks-training.yaml comment and the #1688 writeup. Mechanism and fix are unchanged — on any containerd 2.x the command source returns the migrated config (version = 4) while the AKS root config stays version = 2 — but worth correcting so readers comparing against their node's containerd version aren't thrown off.

yuanchen8911 (dismissed):
Verified end-to-end on aicr-test1: reproduced the v4 drop-in containerd crash from the default command config source, and confirmed that with RUNTIME_CONFIG_SOURCE=file the toolkit regenerates the drop-in as version 2, containerd stays up, the gpu-operator stack converges, and both GPU nodes advertise 8 GPUs. Both-file placement is right since values-aks-training.yaml replaces the base valuesFile. One inline nit on the containerd version in the comments, non-blocking.

recipes/components/gpu-operator/values-aks.yaml
mchmarny: Good catch — confirmed `containerd://2.3.2-1` on both nodes. Corrected to "containerd 2.x (2.3 on current node images)" in both values comments and the commit message (a4462bb), and updated the #1688 writeup to match.

yuanchen8911 (approved):

PR #1697, opened 2026-07-09
feat(recipe): auto-detect pre-installed GPU driver from snapshot

### LINKED ISSUE #475: feat: GPU operator should skip driver installation if drivers already exist and install otherwise
### Prerequisites

### Feature Summary

GPU operator includes a flag to install the GPU drivers with the `driver.enabled` flag. But providers like EKS, GKE, and AKS can also have their own managed GPU driver. It'll conflict if GPU operator tries to install its own driver. And if the provider doesn't install a driver and GPU operator also skips, then the cluster fails too. I'd like AICR to figure out if the GPU nodes already have a pre-installed driver to determine how to set the driver.enabled flag in GPU operator values. Idea would be such that we just make the cluster conformant and ready to run dynamo, etc. 

As of now, the Helm values for GKE always sets `driver.enabled` to false while the templates for AKS sets `driver.enabled` to true. This feature would remove the inconsistency between providers and ensure there the drivers and operator work regardless of what the user brings in their cluster.

Note: If some nodes have GPU drivers installed while others do not, that could be a tricky situation. For example, if you create one AKS node pool with the `--gpu-driver None` flag and another AKS nodepool `--gpu-driver Install` flag, we couldn't just query one node to check for the driver and assume it applies to all nodes.

### Success Criteria

When you run AICR and deploy to a cluster with pre installed driver, GPU operator will work and there won't be driver conflict. When you run AICR and deploy to a cluster without a pre installed driver, it will install GPU operator along with the driver.

### Alternatives Considered

_No response_

### Component

Recipe engine / data

### Priority

_No response_

_No response_

Yes, I can open a PR

xdu31:
## Summary

When resolving a recipe from a snapshot, read the sampled GPU node's `driver-loaded` measurement and inject `components.gpu-operator.overrides.driver.enabled=false` when the NVIDIA kernel module is already loaded. Prevents the GPU Operator from installing a second driver on top of one the platform has already provisioned.

## Motivation / Context

Today `driver.enabled` is set statically per provider overlay:
- `values.yaml` = `true` (base, inherited by EKS and AKS)
- `values-gke-cos*.yaml` = `false` (GKE-COS preinstalls drivers)
- `values-oke*.yaml` = `false` (OKE preinstalls drivers on GPU shapes)
- `values-aks*.yaml` — unset, falls back to helm chart default `true`

Broken case: an AKS nodepool created with `--gpu-driver Install` (Azure installs the driver) receives AICR's default `driver.enabled=true`, and the GPU Operator tries to install a *second* driver on top → conflict, cluster left in a bad state. The same failure mode exists on EKS GPU-optimized AMIs that ship an NVIDIA driver.

Fixes: #475
Related: #464 (heterogeneous per-node visibility for mixed pools)

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe`) — behavior only; the mutation lives in `pkg/client/v1`
- [x] Other: `pkg/client/v1` (facade — the injection layer)

## Implementation Notes

- **Approach.** Compute the driver state at the facade layer (`pkg/client/v1`) and post-process the internal `recipe.RecipeResult` before it is projected to the facade shape. Zero coupling of `pkg/recipe` to `pkg/snapshotter`. The injection lands in `ComponentRef.Overrides` at the top of the merge chain (`base values.yaml → ValuesFile → Overrides`), so it wins over every provider values file at bundle time.
- **Reducer** (`pkg/client/v1/gpu_driver_state.go`). `computeGPUDriverState(snap)` maps the snapshot to one of four states — `Unknown` / `NotObserved` / `Preinstalled` / `Absent`. A missing `driver-loaded` key is deliberately `Unknown` (not `Absent`), so a stale snapshot from an older CLI cannot flip a hardened overlay.
- **Injection policy: only-false.** `applyGPUDriverAutoOverride` never forces `driver.enabled=true`. Safe under no-cluster mode (`Unknown` → no-op). Idempotent for GKE-COS/OKE (values files already carry `false`). Fixes AKS-with-`--gpu-driver Install` and EKS-with-preinstalled-driver AMIs. AKS-with-`--gpu-driver none` and stock EKS AMIs are unaffected (state=Absent → no-op).
- **Behavior matrix.**

| Overlay | Snapshot state | Static valuesFile | Injection | Final `driver.enabled` |
|---|---|---|---|---|
| AKS + `--gpu-driver Install` | Preinstalled | unset (chart default `true`) | injects false | **false** (bug fixed) |
| AKS + `--gpu-driver none` | Absent | unset (chart default `true`) | no-op | true (correct) |
| EKS + preinstalled-driver AMI | Preinstalled | true | injects false | **false** (bug fixed) |
| EKS + stock AMI | Absent | true | no-op | true (correct) |
| GKE-COS | Preinstalled | false | injects false | false (idempotent) |
| OKE | Preinstalled | false | injects false | false (idempotent) |
| Any + no snapshot / no-cluster | Unknown | as-is | no-op | as-is (unchanged) |

- **Single-sample limitation.** The GPU collector (`pkg/collector/gpu/gpu.go`) runs on one `nvidia.com/gpu.present=true` node the snapshotter Job schedules onto, so the reducer trusts a single-node sample. On homogeneous clusters — the common case — that's representative. Mixed-pool clusters (some nodes with a preinstalled driver, some without) need per-node collector fanout and stay parked behind #464; documented in both the component catalog and the AKS integrator guide.
- **Observability.** One `slog.Info` at the injection point when the override actually lands. Other states are silent to avoid noise on every criteria-only resolve.

## Testing

```bash
make qualify
golangci-lint run -c .golangci.yaml ./pkg/client/v1/...
```

- `make qualify`: passes. Total coverage **78.2%** (threshold 75%).
- `golangci-lint run -c .golangci.yaml ./pkg/client/v1/...`: **0 issues** (mandatory Go-change gate).
- Coverage delta on `pkg/client/v1` via the `git worktree` baseline method from `AGENTS.md`: comparable to baseline; every new unexported func has coverage (`computeGPUDriverState` — 9-case table; `applyGPUDriverAutoOverride` — 8-case table + sibling-preservation test).
- New tests:
  - `TestComputeGPUDriverState` (9 subtests) — reducer table covers nil snapshot, empty, K8s-only, GPU-no-subtypes, missing driver-loaded, gpu-present=false, gpu-count=0, driver-loaded=true, driver-loaded=false.
  - `TestApplyGPUDriverAutoOverride` (8 subtests) + `TestApplyGPUDriverAutoOverride_PreservesSiblingsUnderDriver` — injection policy: nil result, non-Preinstalled states never inject, sibling under `driver` (e.g. `driver.version`) preserved, existing `driver.enabled=true` gets flipped, no gpu-operator ref is a no-op.
  - `TestResolveRecipeFromSnapshot_GPUDriverAutoDetect_*` (4 integration tests) — end-to-end through `ResolveRecipeFromSnapshot` on the embedded recipes: AKS+Preinstalled → override present; AKS+Absent → no override; EKS+Preinstalled → override present (proves provider-agnostic wire-up); AKS+K8s-only snapshot → no override.
- No changes to `recipes/registry.yaml` or version pins, so `make bom-docs` is not required.

**Manual verification (not yet run).** The change is exercised end-to-end by the integration tests using the embedded recipe catalog. A live e2e on the `aicr-test1` AKS testbed with a real `--gpu-driver Install` nodepool is worth confirming before backport if any is planned; the assertion is that `aicr recipe --snapshot snap.yaml --service aks --accelerator h100 --intent training --os ubuntu` emits `components.gpu-operator.overrides.driver.enabled=false`.

## Risk Assessment

**Rollout notes:**
- Additive and behind the `ResolveRecipeFromSnapshot` code path. Recipes resolved without a snapshot (criteria-only, no-cluster mode) see zero behavior change.
- The only-false policy makes this monotonically safe: for any overlay, the injection either lands the same value the static file already carried (idempotent) or fixes the double-install bug. It cannot regress a working overlay to a broken state.
- Revert = drop the four new/edited files under `pkg/client/v1` and the doc paragraphs. No data migration, no persisted state, no on-disk artifacts.
- Follow-ups (separate issues, not blocking this PR): heterogeneous per-node collector fanout (#464); driver-version-aware pre-install (when the loaded driver's version disagrees with `driver.version` in the recipe).

## Checklist


pkg/client/v1/gpu_driver_state.go
```diff
@@ -0,0 +1,176 @@
...
+		ref := &r.ComponentRefs[i]
+		if ref.Overrides == nil {
+			ref.Overrides = map[string]any{}
+		}
+		driverAny, _ := ref.Overrides["driver"].(map[string]any)
+		if driverAny == nil {
+			driverAny = map[string]any{}
+			ref.Overrides["driver"] = driverAny
+		}
+		driverAny["enabled"] = false
```
njhensley, line 391: **🟠 Major — A lone driver.enabled=false violates the documented driver/toolkit lockstep**

The auto-detect injects only `driver.enabled=false`. But `recipes/components/gpu-operator/values-aks.yaml:20-24` states the AKS driver-only profile requires `driver.enabled=false`, `toolkit.enabled=false`, AND `operator.runtimeClass=nvidia-container-runtime` set *together — never just one of them*. On the exact target cases (AKS `--gpu-driver Install`, EKS preinstalled-driver AMIs inheriting base `values.yaml`), base leaves `toolkit.enabled=true` (values.yaml:50) and `gdrcopy.enabled=true` (values.yaml:144) with no `hostPaths.driverInstallDir`. The idempotent GKE-COS/OKE cases only work because their values files already carry the *full* holistic set (driver=false + gdrcopy=false + driverInstallDir) — the fix cases don't.

**Blast radius:** The very clusters this feature targets (AKS/EKS with a platform driver) get a half-configured Operator: driver off, but toolkit/gdrcopy still on and expecting an operator-managed driver root that no longer exists.

**Fix:** Inject the coordinated set the profile requires (toolkit.enabled=false + operator.runtimeClass where applicable), or gate auto-detect to overlays that already carry a full preinstalled profile and treat bare AKS/EKS as out-of-scope until one exists.

pkg/client/v1/gpu_driver_state.go
```diff
@@ -0,0 +1,176 @@
...
+				return gpuDriverNotObserved
+			}
+		case int64:
+			if v == 0 {
+				return gpuDriverNotObserved
+			}
+		}
+	}
+
+	loaded := hw.Get(measurement.KeyGPUDriverLoaded)
```
njhensley, line 151: **🟠 Major — driver-loaded=true has no provenance — re-snapshot can tear down the driver AICR installed**

`DriverLoaded` (pkg/collector/gpu/nfd.go) is set purely on `nvidia` kernel-module presence — installer-agnostic. The GPU Operator's driver container loads that module into the node-global kernel, so a cluster where a prior AICR deploy set driver.enabled=true reports driver-loaded=true on re-snapshot → re-resolve flips true→false → next `helm upgrade` removes the operator-managed driver DaemonSet → new/rebooted GPU nodes come up driverless. The 'monotonically safe, cannot regress a working overlay' claim holds against static values files but not against this self-referential signal, and nothing enforces 'snapshots must be pre-deploy' (re-snapshot on drift-check / GitOps reconcile is normal). Only a slog.Info marks the flip.

**Blast radius:** Any workflow that resolves a recipe from a snapshot taken after the GPU Operator has run — silent driver teardown on redeploy.

**Fix:** The guard is already in the snapshot: pkg/collector/k8s/policy.go collects the GPU-Operator ClusterPolicy. Only auto-disable when driver-loaded=true AND no gpu-operator ClusterPolicy is present. At minimum, document the pre-deploy-snapshot requirement and warn when both a loaded driver and a ClusterPolicy are observed.

pkg/client/v1/gpu_driver_state.go
```diff
@@ -0,0 +1,176 @@
...
+const gpuOperatorComponentName = "gpu-operator"
+
+// gpuHardwareSubtypeName is the subtype name emitted by
+// pkg/collector/gpu when writing the driver-loaded reading (see the
+// subtypeHardware constant there). Re-declared locally so this file
+// does not pull the collector package.
+const gpuHardwareSubtypeName = "hardware"
+
+// gpuDriverState reports what the snapshot tells us about the NVIDIA
+// kernel driver on the sampled GPU node.
```
njhensley, line 48: **🟡 Minor — Single-node sample makes mixed pools fail TOWARD a missing driver (new, silent, worse mode)**

One node is sampled. Before this PR a mixed cluster resolved driver.enabled=true → a driver on every pool. Now one preinstalled-node sample injects false globally → non-preinstalled pools get no driver = dead GPUs.

**Blast radius:** Mixed-pool clusters — a strictly worse and undetectable failure than the double-install being prevented. Parking mixed pools behind #464 is fine; parking them into a worse silent mode is the concern.

**Fix:** The topology collector already sees all node labels — at least warn when GPU pools appear non-uniform before injecting, and note the fail-direction in the #464 deferral + catalog doc.

pkg/client/v1/aicr.go
```diff
@@ -675,6 +675,12 @@ func (c *Client) ResolveRecipeFromSnapshot(ctx context.Context, criteria *Criter
 	if err != nil {
 		return nil, err
 	}
+	// Snapshot-driven post-processing: when the sampled GPU node already
+	// has the NVIDIA kernel driver loaded, inject
+	// gpu-operator.driver.enabled=false so the Operator does not install
+	// a second driver on top. Only-false policy — safe across every
+	// provider (see gpu_driver_state.go for the full rationale).
+	applyGPUDriverAutoOverride(internal, computeGPUDriverState(toInternalSnapshot(snap)))
```
njhensley: **🟡 Minor — Snapshot-driven recipe shaping lives in the pkg/client/v1 facade, not pkg/recipe**

Mutating ComponentRef.Overrides from measured state is recipe business logic; CLAUDE.md designates the facade as a no-business-logic adapter. Latent today (server has no snapshot-resolve path; CLI routes through the facade), but the moment a snapshot endpoint calls pkg/recipe.BuildFromCriteriaWithEvaluator directly, the override silently vanishes with no compile-time or test signal. The evaluator already threads the snapshot into pkg/recipe, which is the natural home.

**Blast radius:** Latent divergence for any future non-facade snapshot-resolve caller. Trade-off, not a defect.

**Fix:** Move the post-process into pkg/recipe as an explicit measurement-driven step, or document at the pkg/recipe seam that snapshot-driven overrides are applied by the facade.

pkg/client/v1/gpu_driver_state.go
```diff
@@ -0,0 +1,176 @@
...
+				return gpuDriverNotObserved
+			}
+		}
+	}
+
+	loaded := hw.Get(measurement.KeyGPUDriverLoaded)
+	if loaded == nil {
+		return gpuDriverUnknown
+	}
+	b, ok := loaded.Any().(bool)
```
njhensley, line 155: **🟡 Minor — Fail-closed non-bool branch untested; reducer never exercised through real deserialization**

computeGPUDriverState sits at 90% (package 75.5%, passes). Untested: the fail-closed guard (driver-loaded present but non-bool → Unknown) the design comment leans on, and the dead int64 count branch. Every fixture builds Readings via SubtypeBuilder (Go-native int/bool), so the ToReading→float64 path is never exercised — which is what masks the float64 gap below.

**Blast radius:** The untested branch is the security-relevant fail-closed path; untested = unverified.

**Fix:** Add a table row for non-bool driver-loaded → Unknown, and one reducer test that round-trips a fixture through json.Unmarshal / sigs.k8s.io/yaml before reducing.

pkg/client/v1/gpu_driver_state.go
```diff
@@ -0,0 +1,176 @@
...
+	// A hardware subtype that explicitly reports no GPU on the node
+	// is a distinct signal from "we couldn't tell" — the sampled node
+	// simply is not a GPU node. Bucket separately so it cannot be
+	// confused with a driver-loaded=false GPU node.
+	if present := hw.Get(measurement.KeyGPUPresent); present != nil {
+		if b, ok := present.Any().(bool); ok && !b {
+			return gpuDriverNotObserved
+		}
+	}
+	if count := hw.Get(measurement.KeyGPUCount); count != nil {
```
njhensley, line 145: **🔵 Nitpick — gpu-count type switch omits float64 (documented anti-pattern)**

The count switch handles int/int64 only; a JSON / sigs.k8s.io/yaml-decoded snapshot delivers gpu-count as float64. Zero impact today (only caller loads via yaml.v3 → int; and gpu-present=false, a bool, short-circuits the no-GPU case first, so the count check never changes an outcome). Still the CLAUDE.md 'handles int/int64 but not float64' anti-pattern.

**Blast radius:** Latent — trips only if a future server/external caller feeds a JSON-decoded snapshot.

**Fix:** Add a `case float64` branch (reject non-integral via float64(int64(v)) != v), or reuse hw.GetInt64 which already normalizes int/int64/float64.

pkg/client/v1/gpu_driver_state.go
```diff
@@ -0,0 +1,176 @@
...
+// wins over every provider values file at bundle time.
+func applyGPUDriverAutoOverride(r *recipe.RecipeResult, state gpuDriverState) {
+	if r == nil || state != gpuDriverPreinstalled {
+		return
+	}
+	for i := range r.ComponentRefs {
+		if r.ComponentRefs[i].Name != gpuOperatorComponentName {
+			continue
+		}
+		ref := &r.ComponentRefs[i]
```
njhensley, line 379: **🔵 Nitpick — In-place Overrides mutation aliases the sync.Once-cached MetadataStore**

initBaseMergedSpec shallow-copies ComponentRefs, so each ref's Overrides map pointer aliases the cached base. Safe only because the base gpu-operator ref carries no Overrides (valuesFile only) and overlays deep-merge into fresh maps. If anyone ever adds an Overrides block to the base ref, this mutation poisons every future resolve (including snapshot-less ones).

**Blast radius:** None today; latent cross-resolve state poisoning if the base ref ever gains Overrides.

**Fix:** Defensive — assign a fresh driver submap / deep-copy (serializer.DeepCopyAnyMap) before writing.

pkg/client/v1/gpu_driver_state.go
```diff
@@ -0,0 +1,176 @@
...
+		if ref.Overrides == nil {
+			ref.Overrides = map[string]any{}
+		}
+		driverAny, _ := ref.Overrides["driver"].(map[string]any)
+		if driverAny == nil {
+			driverAny = map[string]any{}
+			ref.Overrides["driver"] = driverAny
+		}
+		driverAny["enabled"] = false
+		slog.Info("auto-disabled gpu-operator driver install: pre-installed driver detected in snapshot",
```
njhensley, line 392: **🔵 Nitpick — Observability is inject-only**

Only the Preinstalled→inject path logs. Unknown (incl. stale snapshot missing the reading), NotObserved, Absent, and Preinstalled-but-no-gpu-operator-ref are all silent.

**Blast radius:** An operator debugging 'why didn't it disable' (or an unexpected double-install) gets no trace of the computed state.

**Fix:** Add a slog.Debug of the resolved gpuDriverState (and reason) on the no-op paths.

pkg/client/v1/gpu_driver_state.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// applyGPUDriverAutoOverride injects driver.enabled=false into the
+// gpu-operator ComponentRef's Overrides map when the snapshot reports
+// a pre-installed driver on the sampled GPU node.
+//
+// Policy is only-false: the function never forces driver.enabled=true.
+// That keeps the change safe across every provider — GKE-COS and OKE
+// values files already carry driver.enabled=false, so the injection is
+// byte-for-byte idempotent for them; AKS with --gpu-driver Install and
```
njhensley: **🔵 Nitpick — "byte-for-byte idempotent" is imprecise in the code comment**

For GKE-COS/OKE the *rendered* Helm value is unchanged, but the resolved-recipe *artifact* gains an overrides.driver.enabled block a criteria-only resolve omits — so `aicr recipe --snapshot -o recipe.yaml` diffs are not byte-identical. The catalog doc scopes this correctly ('the rendered value is unchanged'); only this comment drops the qualifier.

**Blast radius:** Could mislead an integrator diffing emitted recipes.

**Fix:** Scope the phrasing to rendered Helm output, e.g. 'the rendered driver.enabled value is unchanged (the resolved recipe now records the override explicitly).'

njhensley (commented):
## 🔬 Multi-Persona Review

**Method:** 4 independent persona reviewers (Correctness · Domain/Architecture · Test-coverage · Docs) → an adversarial senior meta-reviewer that re-derived every finding from the resolved code at `33740818`. Refuted claims are listed under *Confirmed non-issues* so you can see they were examined.

### Overall assessment

The mechanism is clean, well-tested, and correctly wired — the internal `RecipeResult` mutation reaches both `Resolved()` and rendered Helm output, nil-safety is thorough, and the only-false policy is genuinely safe against the *static* provider values files it enumerates. The four-state reducer and the Unknown-not-Absent fail-closed default are nicely done.

Two Major findings converge on one root cause: **injecting a lone `driver.enabled=false` is under-specified.** The signal can't distinguish a platform-preinstalled driver from one the GPU Operator itself installed (so a re-snapshot can tear down a working driver — see inline at `gpu_driver_state.go:124`), and a single key doesn't reproduce the *holistic* preinstalled-driver contract the maintainers deliberately encoded in the provider profiles (inline at `gpu_driver_state.go:170`). Both have a real, already-collected in-snapshot signal the reducer ignores (the GPU-Operator `ClusterPolicy`, and the full profile values).

**Recommendation: address the two 🟠 before merge.** No blocker — the feature is opt-in behind the snapshot path — but the two Major items undercut the core premise on the very clusters this targets.

### Additional finding (off-diff, folded here)

**🟡 `docs/user/cli-reference.md:323` — the `recipe --snapshot` bullet omits the new behavior.** It enumerates snapshot-detected effects (accelerator/service/OS/node-count) but not the new `gpu-operator` driver override, and doesn't cross-link the catalog section this PR adds. Per CLAUDE.md (CLI behavior change → `cli-reference.md`), add one sentence + a link to `component-catalog.md#gpu-operator-driver-auto-detect`.

### ✅ Confirmed non-issues (examined, refuted)

- **Nil-safety** — every dereference in both functions is guarded (nil snap/measurement/subtype, missing keys, nil Overrides, absent gpu-operator ref). No panic path.
- **Mutation reaches consumers** — `facadeResultFromInternal` stores the same `*recipe.RecipeResult` pointer; the override reaches `Resolved()` and `BundleComponents` (confirmed by the end-to-end test).
- **`--set` still force-enables** — merge is base→ValuesFile→Overrides, then CLI `--set` on top, so `--set gpuoperator:driver.enabled=true` beats the auto-injected false. Works as intended.
- **`driver-loaded` bool assertion** — bool round-trips as bool under both `yaml.v3` and `encoding/json`→`ToReading`; the float64 nitpick is integer-only.
- **Doc anchor links valid** — both `#alternative-use-the-aks-driver-only-profile` and `#gpu-operator-driver-auto-detect` resolve to real headings; lychee will pass.
- **`ResolveRecipeFromCriteria` skips the injection** — consistent (only-false needs a snapshot signal).
- **PR body "4 integration tests"** — accurate; it just omits mentioning the 5th (`TestBundleComponents_..._RendersInHelmValues`).

### Summary

| Tier | Count |
|---|---|
| 🔴 Blocker | 0 |
| 🟠 Major | 2 |
| 🟡 Minor | 4 |
| 🔵 Nitpick | 4 |

*Generated via the multi-persona + adversarial-meta-review workflow. Findings were independently reproduced against the code before posting.*

njhensley (commented):
## 🔬 Follow-up review — feedback resolution (head `387c32f0`)

Re-reviewed after `9e51e6b` (gate GPU driver auto-detect to preinstalled-profile overlays) + the main merge. Thanks for the thorough rework — **9 of the 10 findings from the prior review are resolved with real code + tests, and CI is fully green** (148 pass / 14 skip / 0 fail; `pkg/client/v1` tests pass locally).

### ✅ Resolved

| Finding | How it was addressed |
|---|---|
| 🟠 Driver/toolkit lockstep | `hasPreinstalledDriverProfile` gate — injection fires only on overlays whose merged base+valuesFile already declares `driver.enabled=false` (the full coordinated profile); bare AKS/EKS get a `slog.Warn` instead of a half-configured Operator. |
| 🟠 `driver-loaded` provenance / re-snapshot teardown | Largely neutralized by the gate above (injection now only lands where the value is already `false`, so the dangerous true→false flip can't occur) + `hasGPUOperatorClusterPolicy` warning + "capture snapshot BEFORE deploy" doc callout. |
| 🟡 Mixed pools fail toward missing driver | `hasHeterogeneousGPUPool` warning on non-uniform GPU topology labels. |
| 🟡 Recipe logic in the facade | Documented at the `pkg/recipe` seam (`builder.go` note telling non-facade callers to apply measurement-driven overrides themselves). |
| 🟡 Fail-closed non-bool branch + no round-trip test | `TestComputeGPUDriverState_YAMLRoundTrip` (JSON **and** sigs-yaml), non-bool→Unknown case, `TestIsZeroCount`. |
| 🔵 `gpu-count` omits float64 | `isZeroCount` handles int/int64/float64 with non-integral fail-closed. |
| 🔵 Overrides mutation aliasing | `serializer.DeepCopyAnyMap` before mutating. |
| 🔵 Observability inject-only | `gpuDriverState.String()` + `slog.Debug` on every no-op/gated path. |
| 🔵 "byte-for-byte idempotent" imprecise | Reworded to "semantically idempotent — rendered value unchanged; recipe records the override explicitly," in comment + docs. |

### 🟡 One remaining minor item

**`docs/user/cli-reference.md:323`** — the `recipe --snapshot` bullet still enumerates only the *criteria* auto-detected from a snapshot (accelerator/service/OS/node-count); it doesn't mention the new `gpu-operator.driver.enabled` override or cross-link the [catalog section](../user/component-catalog.md#gpu-operator-driver-auto-detect). One sentence + a link closes it.

### ⚠️ Worth an explicit sign-off (not a defect — the gate is the right call)

The profile gate is a **scope pivot** worth making conscious:

- The injection now only fires on GKE-COS/OKE, where `driver.enabled=false` is *already* the static default — so **the change alters zero rendered Helm output for any overlay today.** Its net effect is (a) recording the override explicitly in the resolved recipe for auditability, and (b) the warnings.
- The original motivation — the AKS `--gpu-driver Install` / EKS-preinstalled-AMI double-install — is now **deferred**: those cases only warn, with the actual fix pushed to a future full preinstalled-profile overlay.

That's the correct, safe resolution (a no-op + warning beats a half-configured Operator). But if the description still reads **"Fixes #475,"** that's now overstated — #475's double-install isn't fixed here, it's gated + warned. Suggest **"Refs #475"** and a one-line note in the PR body that the actual AKS/EKS fix lands with the follow-up profile overlay, so maintainers merge with that scope in mind.

### Net

Ready to merge once the #475 wording is reconciled; the `cli-reference.md` line is a trivial nice-to-have. Nice work turning the two Majors into a coherent gate-and-warn design backed by tests.

*Generated via the multi-persona + adversarial-meta-review workflow; resolutions independently verified against the code and CI at `387c32f0`.*

njhensley (approved):
## ✅ Approve — all review feedback resolved (head `1fa5cd17`)

Re-reviewed after `9e51e6b` (profile gate) and `b2510789` (cli-reference doc). **All 10 findings from the multi-persona review are addressed**, and CI is green (148 pass / 13 skip / 0 fail; `pkg/client/v1` tests pass locally). Approving.

### Resolution summary

| Finding | Resolution |
|---|---|
| 🟠 Driver/toolkit lockstep | `hasPreinstalledDriverProfile` gate — injection fires only on overlays already declaring the coordinated `driver.enabled=false` profile; bare AKS/EKS warn instead of shipping a half-configured Operator. |
| 🟠 `driver-loaded` provenance / re-snapshot teardown | Neutralized by the gate (injection only lands where the value is already `false`) + `hasGPUOperatorClusterPolicy` warning + pre-deploy-snapshot doc callout. |
| 🟡 Mixed pools fail toward missing driver | `hasHeterogeneousGPUPool` warning on non-uniform GPU topology labels. |
| 🟡 Recipe logic in the facade | Documented at the `pkg/recipe` seam (`builder.go`). |
| 🟡 Fail-closed non-bool branch + no round-trip test | `TestComputeGPUDriverState_YAMLRoundTrip` (JSON + sigs-yaml), non-bool→Unknown case, `TestIsZeroCount`. |
| 🟡 `cli-reference.md` `--snapshot` omitted the behavior | `b2510789` adds the sentence + cross-link to the catalog section. |
| 🔵 `gpu-count` omits float64 | `isZeroCount` (int/int64/float64, non-integral fail-closed). |
| 🔵 Overrides mutation aliasing | `serializer.DeepCopyAnyMap` before mutating. |
| 🔵 Observability inject-only | `gpuDriverState.String()` + `slog.Debug` on every no-op/gated path. |
| 🔵 "byte-for-byte idempotent" imprecise | Reworded (comment + docs). |

### Non-blocking note

The description still reads **"Fixes: #475."** Because the profile gate scopes the injection to overlays where `driver.enabled=false` is already the static default, the AKS `--gpu-driver Install` / EKS-preinstalled-AMI double-install this feature set out to fix is now **gated + warned**, with the actual fix deferred to a follow-up preinstalled-profile overlay. Consider **"Refs #475"** (or leave #475 open on merge) so the tracking stays accurate — entirely your call, not a merge blocker.

Nice work turning the two Majors into a coherent gate-and-warn design backed by tests.

*Generated via the multi-persona + adversarial-meta-review workflow; resolutions independently verified against the code and CI at `1fa5cd17`.*

njhensley (approved):
## ✅ Approve — all review feedback resolved (head `1fa5cd17`)

Re-reviewed after `9e51e6b` (profile gate) and `b2510789` (cli-reference doc). **All 10 findings from the multi-persona review are addressed**, and CI is green (148 pass / 13 skip / 0 fail; `pkg/client/v1` tests pass locally). Approving.

### Resolution summary

| Finding | Resolution |
|---|---|
| 🟠 Driver/toolkit lockstep | `hasPreinstalledDriverProfile` gate — injection fires only on overlays already declaring the coordinated `driver.enabled=false` profile; bare AKS/EKS warn instead of shipping a half-configured Operator. |
| 🟠 `driver-loaded` provenance / re-snapshot teardown | Neutralized by the gate (injection only lands where the value is already `false`) + `hasGPUOperatorClusterPolicy` warning + pre-deploy-snapshot doc callout. |
| 🟡 Mixed pools fail toward missing driver | `hasHeterogeneousGPUPool` warning on non-uniform GPU topology labels. |
| 🟡 Recipe logic in the facade | Documented at the `pkg/recipe` seam (`builder.go`). |
| 🟡 Fail-closed non-bool branch + no round-trip test | `TestComputeGPUDriverState_YAMLRoundTrip` (JSON + sigs-yaml), non-bool→Unknown case, `TestIsZeroCount`. |
| 🟡 `cli-reference.md` `--snapshot` omitted the behavior | `b2510789` adds the sentence + cross-link to the catalog section. |
| 🔵 `gpu-count` omits float64 | `isZeroCount` (int/int64/float64, non-integral fail-closed). |
| 🔵 Overrides mutation aliasing | `serializer.DeepCopyAnyMap` before mutating. |
| 🔵 Observability inject-only | `gpuDriverState.String()` + `slog.Debug` on every no-op/gated path. |
| 🔵 "byte-for-byte idempotent" imprecise | Reworded (comment + docs). |

### Non-blocking note

The description still reads **"Fixes: #475."** Because the profile gate scopes the injection to overlays where `driver.enabled=false` is already the static default, the AKS `--gpu-driver Install` / EKS-preinstalled-AMI double-install this feature set out to fix is now **gated + warned**, with the actual fix deferred to a follow-up preinstalled-profile overlay. Consider **"Refs #475"** (or leave #475 open on merge) so the tracking stays accurate — entirely your call, not a merge blocker.

Nice work turning the two Majors into a coherent gate-and-warn design backed by tests.

*Generated via the multi-persona + adversarial-meta-review workflow; resolutions independently verified against the code and CI at `1fa5cd17`.*

PR #1698, opened 2026-07-09
feat(recipes): replace AKS ib-node-config DaemonSet with nodewright tuning

ayuskauskas:
## Summary

Replaces the privileged `ib-node-config-aks` DaemonSet with nodewright: the AKS accelerator overlays now wire the same `nodewright-customizations` tuning manifest EKS uses (setup and tuning packages), with `service: aks`, and `tuning.yaml` bumps `nvidia-setup` to 0.5.0 which adds the `aks-h100` combination.

## Motivation / Context

The DaemonSet carried a TODO to remove it once nodewright gained AKS RDMA/MEMLOCK support. That support shipped in [NVIDIA/nodewright-packages#42](https://github.com/NVIDIA/nodewright-packages/pull/42) (released as `nvidia-setup/0.5.0`): the package performs the same host setup (persistent `ib_umad`/`rdma_ucm` module loading, memlock limits, `LimitMEMLOCK=infinity` drop-ins for containerd and kubelet), with nodewright handling service restarts and post-interrupt verification. This removes a privileged always-running DaemonSet in favor of lifecycle-managed, checked host configuration, and brings AKS onto the same tuning manifest as EKS (nvidia-tuned applies the `aks-h100-{intent}` profiles).

Fixes: N/A (DaemonSet TODO)
Related: [NVIDIA/nodewright-packages#42](https://github.com/NVIDIA/nodewright-packages/pull/42), [NVIDIA/nodewright#296](https://github.com/NVIDIA/nodewright/issues/296), #1697 (host GPU driver auto-detect; orthogonal, no file overlap)

## Type of Change

## Component(s) Affected

- [x] Other: recipes (overlays, component manifests), `pkg/bundler/validations`, `pkg/tuning` tests

## Implementation Notes

- One shared manifest, no split: `tuning.yaml` moves to `nvidia-setup 0.5.0` for all consumers. 0.5.0 is identical to 0.4.0 for EKS (its only delta is the aks-h100 combination and an `ensure_kernel` no-op when `KERNEL` is unset), so EKS overlays are unaffected beyond the pin.
- AKS wiring mirrors EKS: `h100-aks-inference` (intent `inference`), `h100-aks-training` (intent `multiNodeTraining`), and `a100-aks-training` (h100 profile reuse with the same rationale and comment as `a100-eks-training`). Ubuntu variants inherit. The stale "Nodewright packages do not support service: aks" comments are removed.
- On AKS, `nvidia-setup-kernel` is a designed no-op (aks defaults intentionally omit `KERNEL`; AKS manages its own Ubuntu kernel); this is documented in the manifest.
- Documented AKS caveat on the manifest's tolerate-all fallback: AKS admission collapses a pod's toleration list to just the wildcard when one is present, which defeats the nodewright operator's drain exemption for its own package pods and deadlocks interrupt-bearing packages ([NVIDIA/nodewright#296](https://github.com/NVIDIA/nodewright/issues/296), found while verifying on a live cluster). Keyed `acceleratedTolerations` (for example `nvidia.com/gpu`) must be injected on AKS.
- Guarded at bundle time (per review): new `CheckWildcardAcceleratedToleration` validation in `pkg/bundler/validations`, registered on `nodewright-customizations` with `severity: warning` and `conditions: service: [aks]`. The default `aicr bundle` path injects the `DefaultTolerations()` wildcard into `acceleratedTolerations`, so a plain AKS bundle would otherwise hit the nodewright#296 deadlock silently; the warning surfaces in deployment notes with the `--accelerated-node-toleration` remedy. The keyed-toleration requirement and RDMA opt-out flags are also documented in `docs/integrator/aks-gpu-setup.md` ("InfiniBand RDMA Host Setup (nodewright)"), and the function is listed in `docs/contributor/validator.md`.
- The RDMA opt-out note in `aks.yaml` now includes `--set nodewrightcustomizations:enabled=false`.
- Regenerated `make tuning-docs` (aks rows: `nvidia-setup 0.5.0` + `nvidia-tuned 0.3.1`, matching eks) and `make bom-docs`.

## Testing

- The exact package chain (nvidia-setup-kernel no-op, nvidia-tuned with `aks-h100-inference` profile, nvidia-setup-full IB RDMA + memlock, reboot interrupts) was verified end to end on a live AKS v1.35.5 cluster with 2x `Standard_ND96isr_H100_v5`: all four config files written, modules loaded and reloaded from `modules-load.d` after reboot, `LimitMEMLOCK=infinity` configured and effective on containerd/kubelet, fresh pods see `ulimit -l` unlimited, `umad*`/`rdma_cm` devices present. Full evidence: [nodewright-packages#42 verification comment](https://github.com/NVIDIA/nodewright-packages/pull/42#issuecomment-4929585619).
- `go test ./pkg/tuning/... ./tools/tuning/... ./pkg/bom/...` pass (structure test gains the two aks rows).
- New validation exercised through the real bundle path: default `aicr bundle` of an AKS H100 inference recipe emits the wildcard-toleration warning in deployment notes; with `--accelerated-node-toleration nvidia.com/gpu:NoSchedule` the warning is gone and the rendered Skyhook CR carries the keyed toleration.
- Coverage: `pkg/bundler/validations`: 84.6% → 86.9% (+2.3%); `CheckWildcardAcceleratedToleration` at 100%.
- `make qualify` passes on the rebased branch (test-coverage, lint, tuning-check, e2e — 23 chainsaw tests, scan, license-check).
- `yamllint` clean on the changed recipe files.


ayuskauskas:
Reworked per review: no separate AKS manifest. The AKS accelerator overlays now reuse the shared `tuning.yaml` (setup and tuning) with `service: aks`, and `tuning.yaml` bumps `nvidia-setup` to 0.5.0, which is EKS-compatible (the 0.5.0 delta is only the aks-h100 combination and the KERNEL-unset no-op). This matches the CR shape that was verified on the live AKS cluster.

recipes/components/nodewright-customizations/manifests/tuning.yaml
```diff
@@ -49,6 +49,11 @@ spec:
     # Tolerate-all default, matching ParseTolerations()/DefaultTolerations()
     # (operator Exists, no key). Only reached if no acceleratedTolerations are
     # injected from scheduling defaults/flags.
+    # AKS caveat: AKS admission collapses a pod toleration list to just the
+    # wildcard when one is present, which defeats the nodewright operator
+    # drain exemption for its own package pods and deadlocks packages that
+    # declare interrupts (NVIDIA/nodewright#296). On AKS always inject keyed
+    # acceleratedTolerations (for example nvidia.com/gpu).
```
mchmarny: The default bundle path guarantees the wildcard this caveat warns against. With no `--accelerated-node-toleration` flag, `resolveTolerations` ([pkg/cli/bundle_config.go#L97](https://github.com/NVIDIA/aicr/blob/57056916cb1ffc9074e0fae92322075ad8920899/pkg/cli/bundle_config.go#L97)) falls back to `DefaultTolerations()` — a single bare `operator: Exists` — and the bundler injects it into `acceleratedTolerations` via the registry's `tolerationPaths` ([pkg/bundler/bundler.go#L1203](https://github.com/NVIDIA/aicr/blob/57056916cb1ffc9074e0fae92322075ad8920899/pkg/bundler/bundler.go#L1203)). So the `$cust.acceleratedTolerations` branch above renders the wildcard anyway, and this fallback comment is effectively unreachable from the CLI.

Since `nvidia-setup-full` declares a reboot interrupt, a plain `aicr bundle` of any AKS recipe produces the nodewright#296 deadlock on first install — the RDMA/memlock setup this PR delivers never applies, where the old DaemonSet worked with defaults. The only warning is this template comment, which users never see.

The validations framework already supports `service` conditions ([pkg/bundler/validations/checks.go#L140](https://github.com/NVIDIA/aicr/blob/57056916cb1ffc9074e0fae92322075ad8920899/pkg/bundler/validations/checks.go#L140)), so a registry validation on nodewright-customizations conditioned on `service: aks` that warns when no keyed toleration is provided (same pattern as `CheckAcceleratedSelectorMissing`) would surface this at bundle time. A note in the AKS user docs alongside the RDMA opt-out instructions would help too.

recipes/overlays/aks.yaml
```diff
@@ -67,11 +71,11 @@ spec:
       manifestFiles:
         - components/network-operator/manifests/nfd-network-rule.yaml
```
mchmarny, line 72: `components/network-operator/manifests/nfd-network-rule.yaml:23` still says the rule exists partly to target "the ib-node-config DaemonSet", which this PR deletes. Update that comment in the same PR so it doesn't go stale.

recipes/overlays/aks.yaml
```diff
@@ -67,11 +71,11 @@ spec:
...
         - components/network-operator/manifests/nic-cluster-policy-aks.yaml
-        - components/network-operator/manifests/ib-node-config-aks.yaml
       dependencyRefs:
         - nfd
         - cert-manager
 
+
```
mchmarny: nit: double blank line introduced here — drop one.

mchmarny (changes requested):
Solid replacement of the privileged `ib-node-config` DaemonSet — the AKS wiring exactly mirrors the EKS pattern, the live-cluster verification is thorough, and docs/BOM/tuning-docs are all regenerated. One high-severity concern blocks: the default `aicr bundle` path injects the wildcard `operator: Exists` toleration into the Skyhook CR, which is exactly the nodewright#296 deadlock the manifest comment warns about — and since `nvidia-setup-full` carries a reboot interrupt, a default AKS bundle deadlocks on first install where the old DaemonSet worked out of the box. A bundle-time validation warning (the framework already supports `service` conditions) and/or a user-docs note would resolve it. Two minor items: a stale comment in `nfd-network-rule.yaml` and a double blank line. CI green so far (Tier 1 matrix still in progress).

recipes/components/nodewright-customizations/manifests/tuning.yaml
ayuskauskas: Fixed in 73a37a5a. Added `CheckWildcardAcceleratedToleration` ([pkg/bundler/validations/checks.go#L184](https://github.com/NVIDIA/aicr/blob/73a37a5a3ffaac96e6ec48c4925619b146b16aeb/pkg/bundler/validations/checks.go#L184)), registered on `nodewright-customizations` in `registry.yaml` with `severity: warning` and `conditions: service: [aks]`. It flags an effective accelerated-toleration list that is empty (template wildcard fallback) or contains any keyless toleration — the latter also catches the keyed+wildcard mix, since AKS collapses to the wildcard whenever one is present.

Verified through the real bundle path: a default `aicr bundle` of an AKS H100 inference recipe now emits the warning in deployment notes ("⚠ Warning: nodewright-customizations renders a wildcard (keyless) accelerated-node toleration. AKS admission collapses a pod's tolerations to just the wildcard... Pass keyed tolerations instead, e.g. --accelerated-node-toleration nvidia.com/gpu:NoSchedule."), and bundling with `--accelerated-node-toleration nvidia.com/gpu:NoSchedule` silences it and renders the keyed toleration into the Skyhook CR.

Also added the user-docs note: [docs/integrator/aks-gpu-setup.md](https://github.com/NVIDIA/aicr/blob/73a37a5a3ffaac96e6ec48c4925619b146b16aeb/docs/integrator/aks-gpu-setup.md) gains an "InfiniBand RDMA Host Setup (nodewright)" section with the keyed-toleration requirement and the RDMA opt-out flags (previously only in overlay comments), and the function is listed in `docs/contributor/validator.md`. Kept it a warning rather than a blocking error since the right taint key is cluster-specific and nodewright#296 may land a server-side fix.

recipes/overlays/aks.yaml
ayuskauskas: Fixed in 73a37a5a — dropped the DaemonSet reference; the comment now reads "This label is used by the NicClusterPolicy nodeAffinity to target only IB-capable nodes."

recipes/overlays/aks.yaml
ayuskauskas: Fixed in 73a37a5a.

pkg/tuning/compute_test.go
```diff
@@ -37,6 +37,8 @@ func TestCompute_Structure(t *testing.T) {
 		t.Fatalf("Compute: %v", err)
 	}
 	want := []wantRow{
+		{"aks", "a100", "h100", "nvidia-setup", "nvidia-tuned"},
+		{"aks", "h100", "-", "nvidia-setup", "nvidia-tuned"},
```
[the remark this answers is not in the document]
ayuskauskas: Skipping the hard-coded version columns — the version-agnostic row shape is deliberate (see the `wantRow` comment): this test guards the tuple that does not change on a pin bump, while exact pins are already CI-gated twice. `make tuning-check` (run by `make qualify` and the merge-gate) fails when the committed tuning table — whose aks rows carry `nvidia-setup 0.5.0` — drifts from a fresh render of the manifest, and `TestCompute_VersionsMatchBOM` cross-checks every extracted pin against the committed BOM. That second gate demonstrably catches the failure mode described here: during the rebase of this branch it failed with `pin "nvidia-setup:0.5.0" not found in container-images.md (BOM drift)` until the BOM was regenerated. Asserting versions in the structure test would add a third place to edit on every pin bump without new coverage.

recipes/registry.yaml
```diff
@@ -263,6 +263,15 @@ components:
...
         message: "Without this selector, the customization will run on all nodes. Consider setting --accelerated-node-selector to target specific nodes."
+      - function: CheckWildcardAcceleratedToleration
+        severity: warning
+        conditions:
+          service:
+            - aks
+        message: "AKS admission collapses a pod's tolerations to just the wildcard,
+          which deadlocks nodewright packages that declare interrupts on first
+          install (NVIDIA/nodewright#296). Pass keyed tolerations instead, e.g.
+          --accelerated-node-toleration nvidia.com/gpu:NoSchedule."
```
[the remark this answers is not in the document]
ayuskauskas: Agreed — flipped to `severity: error` in b4850ef5. Given the deadlock needs manual node cordon/reboot to recover, blocking the bundle beats a missable warning.

Two things I paired with it so the block is correct rather than just strict:

1. **Disabled-component guard.** The documented RDMA opt-out is `--set nodewrightcustomizations:enabled=false` with no toleration — that renders no package pods and cannot deadlock, so a naive error would have hard-failed a legitimate, safe config. `CheckWildcardAcceleratedToleration` now skips when the component is disabled via `--set`, checking the exact component name plus its registry aliases (`nodewrightcustomizations`/`skyhookcustomizations`) to match how the bundler resolves overrides (`DefaultBundler.componentOverrideKeys`). Firing on a disabled component was a false positive even as a warning, so this is a correctness fix independent of severity.

2. **Test fixture.** `tests/chainsaw/cli/recipe-overlays` bundles every leaf overlay with a bare `bundle` (no toleration) and now correctly hit the block on the AKS overlays, so it passes `--accelerated-node-toleration nvidia.com/gpu=present:NoSchedule` (matching the sibling bundle tests).

Verified through the real bundle path: default AKS bundle now exits non-zero with the actionable message and produces no bundle; `--accelerated-node-toleration nvidia.com/gpu:NoSchedule` and the `enabled=false` opt-out both succeed. Docs (`aks-gpu-setup.md`, `validator.md`) and the manifest comment updated to say it blocks. `make qualify` green.

mchmarny (approved):
LGTM. All review threads resolved and verified against the head commit:

- `CheckWildcardAcceleratedToleration` wired on `nodewright-customizations` at `severity: error`, with the disabled-component guard so the `nodewrightcustomizations:enabled=false` opt-out isn't false-flagged — correct, not just strict.
- Stale `nfd-network-rule.yaml` DaemonSet comment rewritten to the NicClusterPolicy nodeAffinity intent.
- `wantRow` kept version-agnostic; pins stay CI-gated by `make tuning-check` + `TestCompute_VersionsMatchBOM`.

The default-bundle deadlock concern is now fenced off by a blocking check with two documented escape hatches (keyed toleration / component disable), both live-verified. Coverage up (+2.31% on `pkg/bundler/validations`). Approving once the in-flight CI run is green.

PR #1707, opened 2026-07-10
fix(kwok): mirror registry and gitea images off Docker Hub

ArangoGutierrez:
## Summary

Repoints the KWOK CI lane's two in-cluster image pins off anonymous Docker Hub to first-party mirrors: `registry_image` → `public.ecr.aws/docker/library/registry:3.1.1`, `gitea_image` → `docker.gitea.com/gitea:1.26.4-rootless` (same versions, only the registry host changes).

## Motivation / Context

Main was red on 4 consecutive KWOK Cluster Validation runs on 2026-07-09 (29036321619, 29043219254, 29047218828; plus PR #1686 run 29036085188). Root cause: the in-cluster registry Deployment pulls `registry:3.1.1` anonymously from Docker Hub via containerd inside the kind node; shared GitHub-runner IPs hit anonymous pull rate limits → `install-infra.sh` exit 20 ("registry Deployment not Ready in 120s", verbatim in job 86218808173's log) or setup slowed from ~5m to ~8m. `gitea_image` (flux-git/argocd-git lanes) had the same exposure. ECR Public is AWS's first-party mirror of Docker official images; docker.gitea.com is Gitea's first-party registry — neither rate-limits anonymous pulls, and the fix works identically for fork PRs (unlike authenticated pulls).

Fixes: N/A
Related: #1686 (diagnosis context)

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: KWOK CI lane (`.settings.yaml` `testing_tools` pins; `kwok/scripts/install-infra.sh` header comment only — no script logic change)

## Implementation Notes

- Renovate hints carry the full mirror paths in `depName` (the repo's `.settings.yaml` regex manager captures `depName` immediately after `datasource=`; a `packageName=` key there does not match), so version bumps keep flowing from the mirror registries.
- Both pins carry a rationale comment so a future bump doesn't regress to a bare Docker Hub ref.
- Accepted risk: no fallback if a mirror is down — both are first-party mirrors, and that failure mode is a clean, diagnosable image-pull error rather than a rate-limit flake.
- Remaining anonymous Docker Hub pull in the lane is runner-side `kindest/node` (different machinery — cache/auth; follow-up issue to come). `ubuntu:22.04` in the karpenter scale-test manifest schedules onto KWOK fake nodes and is never pulled.

## Testing

```bash
make build                                # rc=0
[...]

PR #1708, opened 2026-07-10
fix(kwok): deadline-derived sync-gate budgets + 18m tier timeout

ArangoGutierrez:
## Summary

Makes chainsaw sync-gate diagnostics structurally guaranteed in the KWOK CI lane: gate budgets derive from the remaining job time (new `KWOK_SYNC_DEADLINE_EPOCH` contract) with a fail-fast floor, and tier job timeouts move 15m → 18m so the full budget actually fits.

## Motivation / Context

KWOK tier jobs (`timeout-minutes: 15`) raced the chainsaw sync gates' fixed budgets (flux 500s, argocd 480s): GitHub killed the job ~1 minute before chainsaw would time out and print its `catch:` diagnostics, so failures surfaced as CANCELLED with zero error report (job 86181256207 on PR #1686 shows the shape). The budget was over-committed even on the good path: ~300s setup + 500s gate + ~180s post-gate diagnostics/verify/artifact-upload ≈ 980s > 900s. Sibling PR #1707 fixes the Docker Hub throttling that made setup slow enough to hit this consistently; this PR guarantees that whenever anything slow happens again, the lane reports instead of silently dying.

Fixes: N/A
Related: #1707, #1686 (diagnosis context)

## Type of Change

- [x] Build/CI/tooling
- [x] Documentation update

## Component(s) Affected

- [x] Docs/examples (`docs/`)
- [x] Other: KWOK CI lane (`kwok/scripts/`, `.github/actions/kwok-test`, `.github/workflows/kwok-recipes.yaml`, `.github/workflows/kwok-tier3-shard.yaml`)

## Implementation Notes

- `kwok/scripts/lib/sync-budget.sh`: `compute_sync_budget` returns `min(gate default, deadline − now)`; below a 120s floor the gate fails fast with the existing GitOps sync-deadline exit code (50) and an explicit "setup consumed the budget" error — itself the diagnosis a silent CANCELLED used to destroy. Env unset (local runs) ⇒ behavior unchanged.
- Both gates (argocd + flux) consume it; existing `KWOK_ARGOCD_SYNC_TIMEOUT`/`KWOK_FLUX_SYNC_TIMEOUT` overrides still work; the derived budget is logged on every run so the margin is tunable from evidence.
- The kwok-test action derives the deadline in its **first step** — before toolchain setup and the aicr build — so unaccounted job-start drift stays within the 240s margin's 60s allowance (180s is reserved for post-gate diagnostics/artifacts). The `job_timeout_minutes` input is required (no default), integer-guarded (positive, no leading zeros), and passed explicitly by all three call sites — a sync-budget_test.sh case asserts each caller's value matches its job's `timeout-minutes`.
- 18m tier budget: 1080s fits ~300s setup + 500s gate + 180s margin with ~100s slack. Passing runs are unaffected; only failing runs use the extra time.
- The sync-budget unit harness runs as a step in the `discover` job, so a regression fails the lane before any cluster spins up.
- Chainsaw test files untouched (flux-sync/flux-git-sync byte-identical sibling contract preserved).

## Testing

```bash
bash kwok/scripts/lib/sync-budget_test.sh                        # 6/6 pass
shellcheck -x -P SCRIPTDIR kwok/scripts/validate-scheduling.sh \
  kwok/scripts/lib/sync-budget.sh kwok/scripts/lib/sync-budget_test.sh  # rc=0
actionlint .github/workflows/kwok-recipes.yaml \
  .github/workflows/kwok-tier3-shard.yaml       # zero new findings vs base
make kwok-e2e RECIPE=aks-inference               # rc=0
make qualify                                     # rc=0
```

Live behavioral verification on a local kind cluster:
- Ample deadline, full flux-oci lane: **rc=0**, gate logs `Flux sync timeout: 500s (default 500s; deadline-derived when KWOK_SYNC_DEADLINE_EPOCH is set)`.
- Nearly-consumed deadline: gate prints `[ERROR] Flux sync gate: <120s left before the job deadline (…) — setup consumed the budget` and exits 50 — the failure now reports instead of CANCELLED-with-nothing.

## Risk Assessment

**Rollout notes:** CI-only behavior; local runs are unaffected (env unset ⇒ fixed defaults). Revert restores the fixed budgets and 15m timeouts. If the 240s margin proves mistuned, the derived budget is logged on every run for evidence-based adjustment.

## Checklist

### DIFF
--- a/.github/actions/kwok-test/action.yml
+++ b/.github/actions/kwok-test/action.yml
@@ -77,9 +77,11 @@ inputs:
       KWOK_SYNC_DEADLINE_EPOCH (deadline = step start + this budget - 240s
       diagnostics margin) so the chainsaw sync gates always finish - and
       print their catch-block diagnostics - before GitHub kills the job.
-      Keep in sync with the calling job's timeout-minutes.
-    required: false
-    default: '18'
+      Required with no default: every caller must wire its own
+      timeout-minutes here, so a job that changes its timeout cannot
+      silently drift from a stale default (a missing value fails the
+      integer validation below loudly).
+    required: true
 
 runs:
   using: 'composite'
--- a/.github/actions/kwok-test/action.yml
+++ b/.github/actions/kwok-test/action.yml
@@ -71,10 +71,42 @@ inputs:
     description: 'Number of days to retain debug artifacts'
     required: false
     default: '7'
+  job_timeout_minutes:
+    description: >-
+      The calling job's timeout-minutes. Used to derive
+      KWOK_SYNC_DEADLINE_EPOCH (deadline = step start + this budget - 240s
+      diagnostics margin) so the chainsaw sync gates always finish - and
+      print their catch-block diagnostics - before GitHub kills the job.
+      Keep in sync with the calling job's timeout-minutes.
+    required: false
+    default: '18'
 
 runs:
   using: 'composite'
   steps:
+    - name: Derive sync-gate deadline
+      shell: bash
+      env:
+        JOB_TIMEOUT_MINUTES: ${{ inputs.job_timeout_minutes }}
+      run: |
+        # Anchor the sync-gate deadline as early as possible in the job:
+        # only checkout + load-versions run before this step, so the
+        # unaccounted job-start drift stays inside the 240s margin's 60s
+        # allowance (the other 180s is reserved for post-gate work:
+        # chainsaw catch diagnostics, verify_pods, debug-artifact upload).
+        # Anchoring later (at the test step, after setup-go + tool installs
+        # + make build) lets the deadline land past GitHub's job kill and
+        # silently re-creates the CANCELLED-without-diagnostics failure
+        # this mechanism exists to prevent.
+        # See kwok/scripts/lib/sync-budget.sh for the consuming side.
+        if ! [[ "${JOB_TIMEOUT_MINUTES}" =~ ^[0-9]+$ ]]; then
+          echo "::error::job_timeout_minutes must be an integer, got '${JOB_TIMEOUT_MINUTES}'"
+          exit 1
+        fi
+        KWOK_SYNC_DEADLINE_EPOCH=$(( $(date +%s) + JOB_TIMEOUT_MINUTES * 60 - 240 ))
+        echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH}" >> "$GITHUB_ENV"
+        echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH} (job_timeout_minutes=${JOB_TIMEOUT_MINUTES}, margin=240s)"
+
     - name: Set up Go
       uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5  # v6.2.0
       with:
--- a/.github/actions/kwok-test/action.yml
+++ b/.github/actions/kwok-test/action.yml
@@ -75,12 +75,14 @@ inputs:
     description: >-
       The calling job's timeout-minutes. Used to derive
       KWOK_SYNC_DEADLINE_EPOCH (deadline = step start + this budget - 240s
-      diagnostics margin) so the chainsaw sync gates always finish - and
-      print their catch-block diagnostics - before GitHub kills the job.
-      Required with no default: every caller must wire its own
-      timeout-minutes here, so a job that changes its timeout cannot
-      silently drift from a stale default (a missing value fails the
-      integer validation below loudly).
+      diagnostics margin) so each chainsaw sync gate's budget stays within
+      that margin, letting the gate finish - and print its catch-block
+      diagnostics - before GitHub kills the job in the expected
+      single-gate-dominates case (this bounds each gate operation, not the
+      job's overall wall time). Required with no default: every caller
+      must wire its own timeout-minutes here, so a job that changes its
+      timeout cannot silently drift from a stale default (a missing value
+      fails the integer validation below loudly).
     required: true
 
 runs:
@@ -101,13 +103,13 @@ runs:
         # silently re-creates the CANCELLED-without-diagnostics failure
         # this mechanism exists to prevent.
         # See kwok/scripts/lib/sync-budget.sh for the consuming side.
-        if ! [[ "${JOB_TIMEOUT_MINUTES}" =~ ^[0-9]+$ ]]; then
-          echo "::error::job_timeout_minutes must be an integer, got '${JOB_TIMEOUT_MINUTES}'"
+        if ! [[ "${JOB_TIMEOUT_MINUTES}" =~ ^[1-9][0-9]*$ ]]; then
+          echo "::error::job_timeout_minutes must be a positive integer, no leading zeros, got '${JOB_TIMEOUT_MINUTES}'"
           exit 1
         fi
-        KWOK_SYNC_DEADLINE_EPOCH=$(( $(date +%s) + JOB_TIMEOUT_MINUTES * 60 - 240 ))
+        KWOK_SYNC_DEADLINE_EPOCH=$(( $(date +%s) + 10#${JOB_TIMEOUT_MINUTES} * 60 - 240 ))
         echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH}" >> "$GITHUB_ENV"
-        echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH} (job_timeout_minutes=${JOB_TIMEOUT_MINUTES}, margin=240s)"
+        echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH} (job_timeout_minutes=${JOB_TIMEOUT_MINUTES}, margin=240s, job kill in ~$(( 10#${JOB_TIMEOUT_MINUTES} * 60 ))s)"
 
     - name: Set up Go
       uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5  # v6.2.0
--- a/.github/actions/kwok-test/action.yml
+++ b/.github/actions/kwok-test/action.yml
@@ -107,6 +107,14 @@ runs:
           echo "::error::job_timeout_minutes must be a positive integer, no leading zeros, got '${JOB_TIMEOUT_MINUTES}'"
           exit 1
         fi
+        # 120 must equal SYNC_BUDGET_FLOOR_SECONDS in kwok/scripts/lib/sync-budget.sh
+        # (hand-synced literal, guarded by sync-budget_test.sh — same contract as the
+        # 240s margin). A budget under the floor would compute an already-unusable
+        # deadline that only fails later, inside the job, after setup burned CI minutes.
+        if (( 10#${JOB_TIMEOUT_MINUTES} * 60 - 240 < 120 )); then
+          echo "::error::job_timeout_minutes=${JOB_TIMEOUT_MINUTES} leaves no usable sync budget after the 240s diagnostics margin (need >= 6)"
+          exit 1
+        fi
         KWOK_SYNC_DEADLINE_EPOCH=$(( $(date +%s) + 10#${JOB_TIMEOUT_MINUTES} * 60 - 240 ))
         echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH}" >> "$GITHUB_ENV"
         echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH} (job_timeout_minutes=${JOB_TIMEOUT_MINUTES}, margin=240s, job kill in ~$(( 10#${JOB_TIMEOUT_MINUTES} * 60 ))s)"
--- a/.github/workflows/kwok-recipes.yaml
+++ b/.github/workflows/kwok-recipes.yaml
@@ -323,6 +323,7 @@ jobs:
           chainsaw_version: ${{ steps.versions.outputs.chainsaw }}
           chainsaw_sha256: ${{ steps.versions.outputs.chainsaw_sha256_linux_amd64 }}
           kind_node_image: ${{ steps.versions.outputs.kind_node_image }}
+          # keep == this job's timeout-minutes (asserted by sync-budget_test.sh)
           job_timeout_minutes: '18'
 
   # ── Tier 2: diff-aware accelerator tests (PR only, conditional) ──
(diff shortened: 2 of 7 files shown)


.github/actions/kwok-test/action.yml
```diff
@@ -71,10 +71,42 @@ inputs:
...
     default: '7'
+  job_timeout_minutes:
+    description: >-
+      The calling job's timeout-minutes. Used to derive
+      KWOK_SYNC_DEADLINE_EPOCH (deadline = step start + this budget - 240s
+      diagnostics margin) so the chainsaw sync gates always finish - and
+      print their catch-block diagnostics - before GitHub kills the job.
+      Keep in sync with the calling job's timeout-minutes.
+    required: false
+    default: '18'
```
mchmarny, line 87: nit: all three call sites pass `job_timeout_minutes` explicitly, and the "keep in sync with the calling job's timeout-minutes" contract is manual — a future caller that bumps its `timeout-minutes` and relies on this stale default silently re-creates the CANCELLED-without-diagnostics failure this mechanism exists to prevent. Dropping the default (making the input required) turns that drift into a loud failure at wiring time.

mchmarny (dismissed):
Sound design: budget = min(default, deadline − now) with a 120s fail-fast floor that converts the silent CANCELLED into an explicit "setup consumed the budget" diagnosis, env-unset preserves local behavior, and the unit harness gates the lane from the discover job before any cluster spins up. The flux path reusing `EXIT_ARGOCD_SYNC_TIMEOUT` matches the pre-existing documented pattern, and anchoring the deadline in the action's first step is the right call. One nit on the `job_timeout_minutes` default inline; nothing blocking. CI green.

.github/actions/kwok-test/action.yml
ArangoGutierrez: Made it required and dropped the default in 624a947d — a missing value now fails the derive step's integer validation loudly.

njhensley (commented):
## 🔍 Multi-Persona Review

**Method:** 4 independent persona reviewers (Correctness · CI-DX/Operability · Domain & Architecture · Docs), each finding routed through an adversarial senior meta-reviewer that re-derived it from the resolved code at `624a947d`.

### Overall assessment

This is a well-reasoned, well-tested CI fix. The core mechanism is sound: the deadline is anchored in the action's *first* step so pre-setup drift stays inside the margin, both gates consume `min(default, deadline−now)` with a 120s fail-fast floor, the unit harness runs in `discover` before any cluster spins up, and local runs are unaffected when the env var is unset. The highest-risk theories were independently cleared: the flux gate's reuse of `EXIT_ARGOCD_SYNC_TIMEOUT` (50) is **intended and consumer-correct** (`run-all-recipes.sh` matches `argocd-*||flux-*` on rc 50), there's **no double-charge** (one recipe×deployer per job; the absolute deadline shrinks budgets monotonically), the `readonly` const is **sourced exactly once**, and the 6 unit tests cover the floor boundary precisely.

Nothing here blocks. What surfaced is **latent foot-guns and doc drift**, all inline below.

- **Flux gate returns `EXIT_ARGOCD_SYNC_TIMEOUT` (50)** — intentional shared code; the header comment scopes 50 to both gates and `run-all-recipes.sh` matches `argocd-*||flux-*` on it. Not a naming bug.
- **Double-charging the budget** — one recipe×deployer per job; the absolute deadline shrinks budgets monotonically. Correct.
- **`readonly SYNC_BUDGET_FLOOR_SECONDS` re-source error** — sourced exactly once by `validate-scheduling.sh`; `run-all-recipes.sh` invokes it as a subprocess. Matches the `cleanup.sh` contract.
- **`set -e` arithmetic landmines in `compute_sync_budget`** — every `(( ))` is an if-condition; the `if ! budget=$(…)` capture correctly propagates `return 1`.
- **Cold-cache exit-50 misattribution** — already handled: the fail-fast branch logs `"…<120s left before the job deadline… — setup consumed the budget"`, which *is* the diagnosis.
- **`.github/actions/README.md` not updated** — this PR doesn't touch it, and its kwok-test example already omits other `required` inputs; illustrative, not copy-paste-complete. Not a regression this PR introduced (though a broader README refresh wouldn't hurt).
- **`required: true` is advisory for composite inputs** — true, but the regex guard rejects the empty/missing case loudly, so the "fails loudly" claim holds.

### Summary

```
🔴 Blocker 0 | 🟠 Major 0 | 🟡 Minor 3 | 🔵 Nitpick 2      Recommendation: Approve with comments
```

The three 🟡s are all cheap: a sync-assertion for the twin `18`s, one doc line, and a two-char regex tightening. None gate merge.

<sub>Review method: multi-persona fan-out + adversarial meta-review.</sub>

.github/workflows/kwok-recipes.yaml
```diff
@@ -320,6 +323,7 @@ jobs:
           chainsaw_version: ${{ steps.versions.outputs.chainsaw }}
           chainsaw_sha256: ${{ steps.versions.outputs.chainsaw_sha256_linux_amd64 }}
           kind_node_image: ${{ steps.versions.outputs.kind_node_image }}
+          job_timeout_minutes: '18'
```
njhensley, line 327: **🟡 Minor — The two `18`s are unguarded hand-synced literals — the dangerous drift direction re-creates the exact bug this PR fixes**

The deadline derives from the passed `job_timeout_minutes`, but GitHub kills the job at the separate `timeout-minutes` (L294). These are two literals per job, hand-synced across three call sites (tier1 L294↔L326, tier2 L337↔L367, kwok-tier3-shard L38↔L69). If a future edit ever makes `job_timeout_minutes > timeout-minutes` (e.g. bump the input to 25, forget the job cap), `deadline = derive_time + 1500 − 240` lands past the `job_start + 1080` kill → `min()` never shrinks → the gate runs until GitHub kills it → silent CANCELLED-without-diagnostics, the precise failure this PR exists to prevent.

**Blast radius:** All KWOK tier1/2/3 lanes. Nothing is broken today (all six literals = 18), hence Minor — but a future timeout bump that touches one literal and not the other silently reintroduces the diagnostics-loss regression. GitHub exposes no context for a job's own `timeout-minutes`, so it genuinely can't be auto-derived.

**Fix:** Add a grep/unit assertion (alongside `sync-budget_test.sh`) that each caller's `job_timeout_minutes` equals its `timeout-minutes`, plus a paired `# keep == timeout-minutes above` comment at each of the 3 sites.

docs/contributor/tests.md
```diff
@@ -378,6 +378,16 @@ independent.
 | `KWOK_ARGOCD_ROOT_GRACE` | `30` s | Grace period for the root Application before deadline counting starts |
 | `KWOK_FLUX_SYNC_TIMEOUT` | `500` s | Deadline for source fetch (OCIRepository or GitRepository) + Kustomization apply + HelmReleases `Ready=True` + ArtifactGenerators Ready |
 | `KWOK_FLUX_ROOT_GRACE` | `30` s | Grace period for the outer Kustomization before deadline counting starts |
+| `KWOK_SYNC_DEADLINE_EPOCH` | unset | Absolute epoch deadline for sync-gate work (CI only). When set, each gate budget becomes min(default, deadline − now); below a 120 s floor the gate fails fast with exit 50 so chainsaw's catch-block diagnostics always print before GitHub's job timeout |
+
+In CI, the `kwok-test` action derives `KWOK_SYNC_DEADLINE_EPOCH` in its
+first step — before toolchain setup and the `aicr` build, so the anchor
+sits within ~60 s of job start — from its `job_timeout_minutes` input
+(default `18`, kept in sync with the calling job's `timeout-minutes`)
```
njhensley: **🟡 Minor — Docs claim `job_timeout_minutes` has "default `18`" — but the action declares it `required: true` with no default**

The prose says the deadline is derived "from its `job_timeout_minutes` input (default `18`…)". But the second commit flipped `action.yml` to `required: true` with no default and a description that explicitly forbids a stale default — the docs still describe the exact behavior the action was redesigned to forbid. A contributor reading this will believe omitting the input is safe; it isn't (the empty string fails the `^[0-9]+$` regex). It fails loudly, which is why this is Minor and not higher.

**Blast radius:** Doc-only inaccuracy; also misrepresents the deliberate no-default design rationale. The same wording appears in the PR body — worth correcting there too for the permanent record.

**Fix:** Drop "default `18`"; state the input is required, integer, and must match the caller's `timeout-minutes` (currently `18` for the KWOK jobs).

.github/actions/kwok-test/action.yml
```diff
@@ -71,10 +71,44 @@ inputs:
...
+        # Anchoring later (at the test step, after setup-go + tool installs
+        # + make build) lets the deadline land past GitHub's job kill and
+        # silently re-creates the CANCELLED-without-diagnostics failure
+        # this mechanism exists to prevent.
+        # See kwok/scripts/lib/sync-budget.sh for the consuming side.
+        if ! [[ "${JOB_TIMEOUT_MINUTES}" =~ ^[0-9]+$ ]]; then
+          echo "::error::job_timeout_minutes must be an integer, got '${JOB_TIMEOUT_MINUTES}'"
+          exit 1
+        fi
+        KWOK_SYNC_DEADLINE_EPOCH=$(( $(date +%s) + JOB_TIMEOUT_MINUTES * 60 - 240 ))
```
njhensley: **🟡 Minor — Integer guard `^[0-9]+$` admits zero-padded values → octal misparse**

The regex accepts leading zeros, but the arithmetic `$(( … + JOB_TIMEOUT_MINUTES*60 − 240 ))` reads them in base-8. `010` → 8 (deadline silently too short, computed with no error — the exact silent-miscompute class this PR fights); `08`/`09` → `value too great for base`, which under the composite step's `set -eo pipefail` aborts the step before the friendly `::error::` line fires.

**Blast radius:** Latent — all three callers pass `18` — but the validation block's whole job is to reject bad input loudly, and for zero-padded inputs it doesn't.

**Fix:** Tighten to `^[1-9][0-9]*$`, or force base-10: `$(( … + 10#${JOB_TIMEOUT_MINUTES}*60 − 240 ))`.

kwok/scripts/validate-scheduling.sh
```diff
@@ -1638,8 +1657,13 @@ wait_for_flux_sync() {
             ;;
     esac
 
-    local sync_timeout="${KWOK_FLUX_SYNC_TIMEOUT:-500}s"
-    log_info "Flux sync timeout: ${sync_timeout}"
+    local budget_seconds
+    if ! budget_seconds=$(compute_sync_budget "${KWOK_FLUX_SYNC_TIMEOUT:-500}"); then
```
njhensley, line 1661: **🔵 Nitpick — "Always finishes before the kill" is mildly overstated — the budget is per-chainsaw-operation, not a whole-test wall bound**

The single derived budget is passed as both `--assert-timeout` and `--error-timeout`, which chainsaw applies per operation. The flux gate has ~4 sequential asserts; a chain of slow-but-eventually-passing steps can accumulate `N×budget` and overrun the deadline.

**Blast radius:** Pre-existing (identical per-op mechanics with the old fixed 480/500s) and low-probability (first timed-out step aborts the test; KWOK resources go ready-fast-or-stuck). A wording caveat, not a defect.

**Fix:** Note in the `sync-budget.sh` header that the budget is per-operation — the guarantee holds for a single stuck step, not N slow-passing ones.

.github/actions/kwok-test/action.yml
```diff
@@ -71,10 +71,44 @@ inputs:
...
+        echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH}" >> "$GITHUB_ENV"
+        echo "KWOK_SYNC_DEADLINE_EPOCH=${KWOK_SYNC_DEADLINE_EPOCH} (job_timeout_minutes=${JOB_TIMEOUT_MINUTES}, margin=240s)"
```
njhensley: **🔵 Nitpick — Derive step logs the deadline epoch but not the expected-kill offset**

If pre-derive drift ever exceeds 240s (unlikely — shallow checkout), the deadline lands past the kill and the mechanism silently self-disables. The step already logs the epoch/minutes/margin, so it's mostly observable; what's missing is a comparison against the expected kill time to make the degenerate case greppable.

**Blast radius:** Observability nit only; low probability given shallow checkout and the 60s drift allowance.

**Fix:** Also log the expected kill epoch and warn if `deadline ≥ kill`.

.github/workflows/kwok-recipes.yaml
ArangoGutierrez: Added three cases to sync-budget_test.sh (tier1/tier2/tier3-shard) asserting each caller's job_timeout_minutes equals that job's timeout-minutes, plus paired comments at all three sites — mutation-checked: drifting one literal to 19 fails the matching case. cfb49617.

docs/contributor/tests.md
ArangoGutierrez: Reworded in cfb49617: required, integer, must match the caller's timeout-minutes (currently 18). Corrected the same claim in the PR description.

.github/actions/kwok-test/action.yml
ArangoGutierrez: Tightened to ^[1-9][0-9]*$ with a clearer error, and forced base-10 (10#) at both arithmetic sites. cfb49617.

kwok/scripts/validate-scheduling.sh
ArangoGutierrez: Softened in cfb49617 — the overstated claim actually lived in sync-budget.sh's header comment and the action description (validate-scheduling.sh sources the lib); both now state the per-operation bound rather than a whole-test wall guarantee.

.github/actions/kwok-test/action.yml
ArangoGutierrez: The derive step's log line now includes the implied kill offset (job kill in ~Ns) next to the deadline, so a drifted deadline is visible at a glance. cfb49617.

.github/actions/kwok-test/action.yml
```diff
@@ -71,10 +71,42 @@ inputs:
...
+
```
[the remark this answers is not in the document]
ArangoGutierrez: Added the derive-time floor check in a30bb2e6 — after the integer validation, the step now fails fast when `job_timeout_minutes*60 - 240 < 120` (the 120 is hand-synced with `SYNC_BUDGET_FLOOR_SECONDS` and test-guarded, same contract as the 240s margin). Regression cases extract and execute the actual derive block from action.yml: `5` → rc 1, `6` → rc 0 (boundary exactly at the floor), mutation-checked — deleting the guard fails the suite.

.github/actions/kwok-test/action.yml
```diff
@@ -71,10 +71,54 @@ inputs:
...
+  job_timeout_minutes:
+    description: >-
+      The calling job's timeout-minutes. Used to derive
+      KWOK_SYNC_DEADLINE_EPOCH (deadline = step start + this budget - 240s
+      diagnostics margin) so each chainsaw sync gate's budget stays within
+      that margin, letting the gate finish - and print its catch-block
+      diagnostics - before GitHub kills the job in the expected
+      single-gate-dominates case (this bounds each gate operation, not the
+      job's overall wall time). Required with no default: every caller
+      must wire its own timeout-minutes here, so a job that changes its
```
mchmarny, line 83: nit: `job_timeout_minutes` is now a required input, but `.github/actions/README.md` (the `kwok-test` input catalog + usage example) wasn't updated — its Inputs list omits this input and the example would fail the integer guard if copy-pasted. Worth adding in this PR.

mchmarny (approved):
Solid CI reliability fix — deadline-derived sync budgets with a fail-fast floor turn silent CANCELLED-with-no-diagnostics into an explicit exit-50 report. Logic is correct, boundaries between `action.yml` and `sync-budget.sh` are consistent (jtm=6 ↔ 120s floor), the deadline is anchored early so setup drift is accounted for, and the test harness extracts CI logic from source rather than duplicating it. One low nit inline: the new required `job_timeout_minutes` input isn't reflected in `.github/actions/README.md`. Nothing blocking.

Issue #1280, opened 2026-06-09
DC7 — Azure UAT revive-or-retire


njhensley:
**Parent:** #1264 · **Labels:** `enhancement`, `area/tests` (auto), `area/ci`
(auto), `area/docs` (auto) · **Size:** S (retire) / M (revive) · **Depends on:**
none (independent of DC1–DC6; carries its own design debate, peeled from DC6)

**Goal.** Make and implement a recorded revive-or-retire decision for the dead Azure
UAT scaffolding so nothing dead lingers.

**Scope.**
- `tests/uat/azure/tests/cuj{1,2}-*/` exist with no workflow referencing them (no
  `uat-azure.yaml`). Either add a `uat-azure.yaml` + `tests/uat/azure/run`
  + `cluster-config.yaml` (revive, **only** if Azure GPU capacity is actually
  available and a reservation row can be added to `infra/uat/reservations.yaml`), or
  **remove the dead `tests/uat/azure/` tree**.
- Record the decision in `docs/contributor/uat.md`.
- **Default recommendation: retire** unless a concrete Azure reservation exists
  (mirrors ADR-007's "don't pre-build without demand").

**Out of scope.** Any Azure work beyond the decision and its implementation. Adding
Azure to the launch matrix (would be a future reservation-row data edit, DC1/DC5).

**Key files.** `tests/uat/azure/**`, `infra/uat/reservations.yaml` (DC1, only if
reviving), `docs/contributor/uat.md`.

**Tests.** The decision is reflected: a `uat-azure.yaml` added + green (revive), or
the tree removed and referenced nowhere with the docs note (retire).

**Acceptance.** Azure decided and implemented (workflow added + green, or tree
removed + docs updated); decision recorded in `docs/contributor/uat.md`.

> **Decomposition note.** DC6 originally bundled five heterogeneous gap-fills; the two
> items carrying independent design debate are peeled out: **Azure revive-or-retire
> is now its own child (DC7)**, and the **merge-gate cluster-smoke is a *Deferred*
> proposal** (not committed scope) rather than a DC6 line item. DC6 retains only the
> three mechanical fixes (no-hard-fail/superseded-run guard, orphaned-Kind, opt-in
> per-PR GPU) and is splittable into small PRs.

---
_Child of #1264 · staged from `docs/design/011-uat-dynamic-clusters.md` (DC7)._

mchmarny:
[...]

Issue #1264, opened 2026-06-09
Epic: DC — Comprehensive UAT + dynamic, CI-managed GPU clusters


njhensley:
### Summary

Make AICR's real-hardware UAT comprehensive and contention-safe by time-sharing the
single per-cloud reservation across a **day/night operating model**. Introduce a
simple day/night **scheduler + capacity request/queue broker** that arbitrates the
scarce reservation-backed GPU time (replacing today's AWS hard-fail-on-busy): **night**
runs a `main`-first, time-boxed **version matrix** — the training and inference CUJ on
AWS and GCP for `main` + the previous N stable AICR releases — sequentially per cloud;
**morning** hands each freed reservation to a long-lived **daytime human-access
deployment** (one training cluster on one cloud, one inference cluster on the other,
for human use outside CI); **day** is untouched-by-CI human use; **evening** tears the
daytime deployment down — an **enforced cleanup boundary** — before the next night
batch. Make the per-run cluster **shape per intent** (training vs inference) from the
same reservation, and **version-parameterize the install** (released `aicr`/validator
images at version X for each release row, `main` at tip) so the matrix is
cross-version regression. Wire the existing CUJ2 inference assets into the UAT runner
at CUJ1 parity, broaden CLI-boundary coverage (`query`, `diff`, `evidence
verify/publish`, real-cluster `snapshot`, more bundle deployers), and drive resolvable
recipes across the dynamic clusters (and KWOK where hardware-independent) so the matrix
fills out beyond the two launch recipes. Every UAT run emits an ADR-007 evidence bundle
handed to the TestGrid epic's TestGrid ingestion (authenticating as the TestGrid epic's
dedicated write-only publish SA, never the shared UAT actuator SA); the
dynamic-clusters epic does not write the bucket, and the daytime human-access clusters
emit no bundle and produce no column. Launch is phased: `training-kubeflow` on both
clouds first (`eks`/`h100-ubuntu`, `gke`/`h100-cos`), `inference-dynamo` once DC3
lands. Designed for more reservations, clouds, and accelerators to come (and eventually
both daytime flavors per cloud). Cost governance is deferred (the only variable cost is
the CPU/system node pool — autoscaled on GCP, fixed on AWS).

### Children
- [x] #1274 — DC1: Day/night scheduler + capacity request/queue broker
- [x] #1275 — DC2: Dynamic per-intent cluster acquisition
[...]

PR #1709, opened 2026-07-10
feat(uat): Azure AKS UAT — OIDC federation, dispatch, pipeline (1/2)

mchmarny:
## Summary

Adds Azure (AKS) as the third UAT cloud: Entra OIDC account federation (`infra/uat-azure-account/`), a quota-backed `azure-h100` reservation row (manual-dispatch only during bring-up), `run-azure` dispatch + a fail-closed unmapped-cloud guard in `uat-run.yaml`, and a full `uat-azure.yaml` pipeline. PR 1 of 2 — the staged bring-up; PR 2 adds `tests/uat/azure/run` + per-intent test configs and flips nightly enrollment after green manual runs.

## Motivation / Context

Revives the dormant Azure UAT scaffolding (the AKS-adapted CUJ assert files under `tests/uat/azure/tests/` had no workflow referencing them). The committed `tests/uat/azure/cluster-config.yaml` was validated manually against the AKS actuator (clean bring-up on westus NDSH100v5 quota).

Fixes: N/A
Related: #1280 (DC7 revive arm; acceptance completes with PR 2's green run), #1264

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT infrastructure (`infra/uat*`, `.github/workflows/uat-*`, `pkg/uatbroker`, `tools/uat-broker`)

## Implementation Notes

- **Broker semantics** (`pkg/uatbroker`): `azure` is a valid cloud; `reservation-id` is now optional (quota-backed capacity has no reservation object); `nightly-intents` distinguishes absent (nil → `[training]` default, unchanged) from explicit `[]` (nightly opt-out — the bring-up mode). The nightly batch skips opted-out legs gracefully but still fails closed when the broker output key is missing entirely.
- **Fail-closed dispatch**: `uat-run.yaml` gains an `unmapped-cloud` guard job — previously a registry cloud with no `run-<cloud>` job silently no-opped while consuming the reservation lease.
- **Auth model**: GitHub OIDC → Entra federated credential (`azure/login`, SHA-pinned v3.0.0) → az CLI context in `~/.azure`, mounted into the AKS actuator container for apply and destroy. Federated sessions cannot self-refresh, so the pipeline re-runs `azure/login` at each long phase boundary (mirrors the AWS per-stage STS refreshes). Subjects restricted to `main` + `test/uat` refs; no client secret exists.
- **Federation identifiers are live**: `infra/uat-azure-account/` has been `terraform apply`d (managed-identity federation; the tenant blocks app registrations), and `uat-azure.yaml` commits the real `AZURE_CLIENT_ID`/`AZURE_TENANT_ID`/`AZURE_SUBSCRIPTION_ID` — non-secret identifiers, same posture as the AWS account ID in `uat-aws.yaml` (access is gated by the federated-credential subject pins, not knowledge of the IDs). `./tests/uat/azure/run` and the `h100-<intent>-config.yaml` files ship in PR 2; the `Validate inputs` guard skips the test-config check for `skip_tests` / `daytime-down`, so the burn-in path (manual `uat-run` dispatch with `skip_tests: true`: provision → connect → teardown) works now, and `azure-h100` stays `nightly-intents: []` until PR 2 lands.
- Azure capacity posture: quota-backed, no pre-flight assertion (GCP posture — the actuator fails loudly at provision time); recorded in `docs/contributor/uat.md`.

## Testing

```bash
make qualify        # pass (unit -race, lint, e2e 23/23, scan, license)
make test-coverage  # pass (project floor)
```

Coverage deltas: `pkg/uatbroker` 96.9% → 97.1% (+0.2%); `tools/uat-broker` 91.5% → 91.5% (±0.0%).

New tests: azure-cloud validity, optional reservation-id, nil-vs-empty nightly-intents contract (unit + end-to-end through YAML parsing and the CLI resolver), committed-registry expectations for `azure-h100`.

Post-merge burn-in (manual): `gh workflow run uat-run.yaml -f reservation=azure-h100 -f skip_tests=true` → verify clean teardown (`az group list` shows no `aicr-*` groups). Account setup (terraform apply + role assignments + state bootstrap) is already done.

## Risk Assessment

**Rollout notes:** Azure is inert until the Terraform is applied and IDs are baked in: the reservation row is nightly-opted-out, the daytime rotation excludes it, and the only reachable path is explicit manual dispatch. Existing AWS/GCP pipelines are untouched except comment updates in `uat-run.yaml` / `uat-nightly-batch.yaml`.

## Checklist

### DIFF
--- a/.github/workflows/uat-azure.yaml
+++ b/.github/workflows/uat-azure.yaml
@@ -1,0 +1,826 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: UAT - Azure
+
+# Reusable per-cloud UAT pipeline — invoked by uat-run.yaml, the shared
+# dispatch surface that owns the per-reservation concurrency lease (#1274,
+# DC1). It is intentionally NOT directly dispatchable or scheduled: the
+# nightly cron lives in uat-nightly-batch.yaml and ad-hoc runs go through
+# uat-run.yaml, so every run on the Azure quota pool serializes on the same
+# lease. The keyless-cosign signing identity (and the EXPECTED_IDENTITY_REGEXP
+# pin below) is uat-azure.yaml@…; kept in lockstep with uat-gcp.yaml /
+# uat-aws.yaml.
+on:
+  workflow_call:
+    inputs:
+      reservation:
+        description: 'Reservation name (lease key) this run targets; used for labeling and the daytime cluster name.'
+        type: string
+        required: true
+      aicr_version:
+        description: 'Released aicr version to install (e.g. v1.2.3); empty = build from source (main tip).'
+        type: string
+        default: ''
+      intent:
+        description: 'Recipe intent — selects the per-intent test config (h100-<intent>-config.yaml).'
+        type: string
+        default: training
+      lifecycle:
+        description: 'nightly (provision→CUJ→teardown) | daytime-up (provision+deploy+HOLD) | daytime-down (tear down the held daytime cluster).'
+        type: string
+        default: nightly
+      cluster_config_path:
+        description: 'Cluster-config the AKS actuator provisions from (from infra/uat/reservations.yaml).'
+        type: string
+        required: true
+      test_config_dir:
+        description: 'Directory holding the AICRConfig test files (from infra/uat/reservations.yaml).'
+        type: string
+        required: true
+      skip_delete:
+        description: 'Skip cluster teardown (manual runs only; the nightly batch always deletes).'
+        type: boolean
+        default: false
+      skip_tests:
+        description: 'Skip UAT test execution (manual runs only; the nightly batch always tests).'
+        type: boolean
+        default: false
+
+permissions:
+  contents: read
+
+# NOTE: the per-reservation concurrency lease is owned by the caller
+# (uat-run.yaml). A reusable workflow must not declare its own concurrency
+# here, or every reservation would serialize against one group instead of
+# queueing per reservation.
+
+jobs:
+  uat-azure:
+    if: github.repository == 'nvidia/aicr'
+    runs-on: ubuntu-latest
+    # Budget = uncapped non-UAT steps (build, validator image push, AKS
+    # provisioning, evidence upload ~30m) + UAT phase steps (prep 15 + install 60
+    # + validate 40 + the intent-selected CUJ (train 25 OR serve 40) + verify 5
+    # = up to 160) + the always()-run Destroy Cluster teardown (3 retries, ~10m
+    # each = up to ~40m). Only one of train/serve runs per intent, so the budget
+    # counts the larger (serve). The teardown MUST fit in this budget: a job-level
+    # timeout cancels pending always() steps, so an undersized cap would skip
+    # teardown and leak the GPU node. 160 UAT + ~30 setup + ~40 teardown = ~230,
+    # so 270 leaves the teardown ample headroom even on a worst-case phase run.
+    timeout-minutes: 270
+    permissions:
+      contents: read
+      actions: read
+      id-token: write
+      packages: write
+    outputs:
+      # Digest-pinned OCI ref of the signed evidence bundle, consumed by
+      # the ingest-evidence job below. Empty when no bundle was produced.
+      bundle_ref: ${{ steps.evidence_ref.outputs.ref }}
+    env:
+      # Entra federation identifiers (infra/uat-azure-account terraform
+      # output; the client id is the github-actions-aicr user-assigned
+      # managed identity). Identifiers, NOT secrets — the federated-credential
+      # subject pins (main + test/uat refs of NVIDIA/aicr), not knowledge of
+      # these ids, gate access.
+      AZURE_CLIENT_ID: "047f0427-5dcf-4e7c-97c3-9c71d92c5abb"
+      AZURE_TENANT_ID: "43083d15-7273-40c1-b7db-39efd9ccc17a"
+      AZURE_SUBSCRIPTION_ID: "e88faa01-b4fd-49d3-b934-0ad9f9fca307"
+      # Managed-identity principal (object) id — the assignee for the
+      # run-scoped cluster-admin grant below (terraform output
+      # AZURE_PRINCIPAL_ID). Inlined because the federated az session may
+      # lack the Graph access needed to resolve the client id to a principal.
+      AZURE_PRINCIPAL_ID: "3c061691-eded-40cb-869e-d5fdb2944a62"
+      # Cluster name (AKS actuator derives both the cluster name and its
+      # resource group `<id>-rg` from .deployment.id). The nightly
+      # per-run lifecycle uses a run-scoped name for per-run isolation; the
+      # daytime lifecycles use a STABLE, reservation-tagged name so the evening
+      # teardown (daytime-down) and the nightly pre-batch guard can find the
+      # held cluster. Reservation names are lowercase alnum+hyphen, so the
+      # derived name is a valid AKS cluster name.
+      DEPLOYMENT_ID: >-
+        ${{ (inputs.lifecycle == 'daytime-up' || inputs.lifecycle == 'daytime-down')
+        && format('aicr-day-{0}', inputs.reservation)
+        || format('aicr-{0}', github.run_id) }}
+      CLUSTER_CONFIG: ${{ inputs.cluster_config_path }}
+      # DC2 selects the AICRConfig by intent; both intents drive the same
+      # cluster-config (GPU pool from the reservation, system/CPU pools dynamic).
+      TEST_CONFIG: ${{ inputs.test_config_dir }}/h100-${{ inputs.intent }}-config.yaml
+      AKS_ACTUATOR_IMAGE: "ghcr.io/mchmarny/cluster/aks@sha256:d3831f21c397b0a03a281e7d55e39e6c9535ba7e099c11dba5c9c0c28cb13c4a"
+      VALIDATOR_IMAGE_PREFIX: "ghcr.io/nvidia/aicr-validators"
+      VALIDATOR_TAG: "uat-${{ github.run_id }}"
+      # Pin validator-image resolution for MAIN cells to this run's freshly
+      # built+pushed validators (the `Build and push validator images` step
+      # tags them `:uat-<run_id>`), mirroring how AICR_IMAGE pins the agent — so
+      # the validators pair with the binary under test rather than a stale
+      # :latest. The main binary is stamped version=main (below), which the
+      # catalog would otherwise leave on the :latest path. Empty for release
+      # cells (aicr_version set): they use the released :<version> validators the
+      # released binary self-resolves, and an empty AICR_VALIDATOR_IMAGE_TAG is
+      # a no-op (the catalog ignores an empty override).
+      AICR_VALIDATOR_IMAGE_TAG: ${{ inputs.aicr_version == '' && format('uat-{0}', github.run_id) || '' }}
+
+    steps:
+      # Checkout
+      - name: Checkout
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          persist-credentials: false
+
+      # Fail closed on an unrecognized intent/lifecycle before any provisioning:
+      # workflow_dispatch constrains these to a choice, but workflow_call passes
+      # free strings, so a caller typo must surface as a clear error rather than
+      # a missing-file failure deep in a phase. Also assert the per-intent config
+      # exists so a missing sibling config fails here, not mid-run.
+      - name: Validate inputs
+        env:
+          INTENT: ${{ inputs.intent }}
+          LIFECYCLE: ${{ inputs.lifecycle }}
+          SKIP_TESTS: ${{ inputs.skip_tests }}
+        run: |
+          set -euo pipefail
+          case "${INTENT}" in
+            training|inference) ;;
+            *) echo "::error::unknown intent '${INTENT}' (want training|inference)"; exit 1 ;;
+          esac
+          case "${LIFECYCLE}" in
+            nightly|daytime-up|daytime-down) ;;
+            *) echo "::error::unknown lifecycle '${LIFECYCLE}' (want nightly|daytime-up|daytime-down)"; exit 1 ;;
+          esac
+          # The per-intent config is only consumed by the UAT phase steps, so
+          # only require it when tests will actually run: skip_tests burn-in
+          # runs and daytime-down (teardown-only) must not fail on a missing
+          # test config before provisioning/teardown even starts.
+          if [[ "${SKIP_TESTS}" != "true" && "${LIFECYCLE}" != "daytime-down" && ! -f "${TEST_CONFIG}" ]]; then
+            echo "::error::test config not found: ${TEST_CONFIG}"; exit 1
+          fi
+          echo "intent=${INTENT} lifecycle=${LIFECYCLE} test-config=${TEST_CONFIG} cluster=${DEPLOYMENT_ID}"
+
+      # Versions
+      - name: Load versions
+        id: versions
+        uses: ./.github/actions/load-versions
+
+      # Build from source
+      - name: Setup Go
+        # Main cells only: release cells install a prebuilt binary and must not
+        # depend on the source Go toolchain (a broken Go pin shouldn't fail them).
+        if: inputs.aicr_version == '' && inputs.skip_tests != true
+        uses: actions/setup-go@924ae3a1cded613372ab5595356fb5720e22ba16  # v6.5.0
+        with:
+          go-version: '${{ steps.versions.outputs.go }}'
+          cache: true
+          cache-dependency-path: |
+            go.sum
+            vendor/modules.txt
+
+      # Main cell (aicr_version empty): build the aicr binary from the
+      # checked-out source (tip). Release cells install the released binary
+      # instead (next step). daytime-down tears down the held cluster only — it
+      # builds nothing and pushes no images.
+      - name: Build aicr binary
+        if: inputs.aicr_version == '' && inputs.skip_tests != true && inputs.lifecycle != 'daytime-down'
+        env:
+          GOFLAGS: -mod=vendor
+          # github.sha via env (never inline ${{ }} in a shell run — expression
+          # injection). It is a 40-hex commit, so no real risk, but env is the
+          # repo norm and keeps zizmor/CodeRabbit quiet.
+          COMMIT_SHA: ${{ github.sha }}
+        run: |
+          set -euo pipefail
+          # Stamp the main-cell binary's version metadata via ldflags (mirroring
+          # goreleaser's pkg/cli.{version,commit,date} targets) so the emitted
+          # recipe/evidence report "main" + the commit instead of the un-stamped
+          # "dev"/"unknown" defaults. version="main" is deliberately NOT a
+          # release-shaped string, so it does not trip the catalog's
+          # :latest→:<version> validator-image rewrite; validator resolution is
+          # pinned explicitly below (AICR_VALIDATOR_IMAGE_TAG). Release cells
+          # install a goreleaser-built binary already stamped with its tag, so
+          # this step is main-only.
+          BUILD_DATE="$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+          go build \
+            -ldflags "-X github.com/NVIDIA/aicr/pkg/cli.version=main -X github.com/NVIDIA/aicr/pkg/cli.commit=${COMMIT_SHA} -X github.com/NVIDIA/aicr/pkg/cli.date=${BUILD_DATE}" \
+            -o ./aicr ./cmd/aicr
+          # Assert the stamp actually landed — a broken ldflags target path would
+          # otherwise silently leave the binary on the default dev/unknown
+          # metadata (which then propagates to the evidence/TestGrid version).
+          VER_OUT="$(./aicr --version)"
+          echo "${VER_OUT}"
+          if ! printf '%s' "${VER_OUT}" | grep -qF "main (commit: ${COMMIT_SHA}"; then
+            echo "::error::aicr binary not stamped (want 'main (commit: ${COMMIT_SHA}…'); got: ${VER_OUT}" >&2
+            exit 1
+          fi
+
+      # Release cell (aicr_version set): install the RELEASED aicr binary at that
+      # version instead of building. The released binary self-resolves its own
+      # version's images — validators via pkg/validator/catalog (:latest ->
+      # :<version>) and the snapshot-agent via ghcr.io/nvidia/aicr:<version> — so
+      # the validator/agent build+push steps below are skipped for release cells.
+      - name: Install released aicr
+        if: inputs.aicr_version != '' && inputs.skip_tests != true && inputs.lifecycle != 'daytime-down'
+        uses: ./.github/actions/install-aicr-release
+        with:
+          aicr-version: ${{ inputs.aicr_version }}
+
+      - name: Authenticate to GHCR
+        uses: ./.github/actions/ghcr-login
+
+      # Main cell only: release cells use the released validator images the
+      # released binary self-resolves to (skipped when aicr_version is set).
+      - name: Build and push validator images
+        if: inputs.aicr_version == '' && inputs.skip_tests != true && inputs.lifecycle != 'daytime-down'
+        run: |
+          GO_VERSION="$(cat .go-version)"
+          # CHAINSAW_* args dropped in #1236 — the deployment validator
+          # no longer ships /usr/local/bin/chainsaw; it executes Chainsaw
+          # Test format in-process via validators/chainsaw/inprocess.go.
+          #
+          # aiperf-bench is an ancillary (non-Go) validator image the
+          # performance phase's inference-perf check pulls at runtime, resolved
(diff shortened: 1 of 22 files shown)


.github/workflows/uat-azure.yaml
```diff
@@ -0,0 +1,738 @@
...
+      # the ingest-evidence job below. Empty when no bundle was produced.
+      bundle_ref: ${{ steps.evidence_ref.outputs.ref }}
+    env:
+      # Entra federation identifiers (infra/uat-azure-account terraform
+      # output). Identifiers, NOT secrets — the federated-credential subject
+      # pins (main + test/uat refs of NVIDIA/aicr), not knowledge of these
+      # ids, gate access.
+      AZURE_CLIENT_ID: "00000000-0000-0000-0000-000000000000"  # PLACEHOLDER — replace with `terraform output AZURE_CLIENT_ID` from infra/uat-azure-account (fast-follow after terraform apply)
+      AZURE_TENANT_ID: "00000000-0000-0000-0000-000000000000"  # PLACEHOLDER — replace with `terraform output AZURE_TENANT_ID` from infra/uat-azure-account (fast-follow after terraform apply)
+      AZURE_SUBSCRIPTION_ID: "e88faa01-b4fd-49d3-b934-0ad9f9fca307"
```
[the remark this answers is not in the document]
mchmarny: Partially fixed in 0b47ed43: `AZURE_TENANT_ID` is now the real tenant id from the `infra/uat-azure-account` terraform output. `AZURE_CLIENT_ID` stays a placeholder because the app registration does not exist yet (the terraform state holds only the subscription resource; the client id is minted by the pending apply — fast-follow, as the inline comment says). Deliberately not switching to `vars.*`: the sibling pipelines inline these identifiers in workflow env (`AWS_ACCOUNT_ID` in uat-aws.yaml, `GCP_WIF_*` in uat-gcp.yaml) so the workflow is self-contained and reviewable, and `infra/uat-azure-account/README.md` already documents wiring the outputs.

.github/workflows/uat-azure.yaml
```diff
@@ -0,0 +1,738 @@
...
+      # Destroy below). A federated az session cannot self-refresh (the OIDC
+      # token is single-exchange; access tokens last ~60-75 min), so the long
+      # phases below each re-run azure/login — mirroring the AWS pipeline's
+      # per-stage STS refreshes.
+      - name: Authenticate to Azure
+        uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43  # v3.0.0
+        with:
+          client-id: ${{ env.AZURE_CLIENT_ID }}
+          tenant-id: ${{ env.AZURE_TENANT_ID }}
+          subscription-id: ${{ env.AZURE_SUBSCRIPTION_ID }}
```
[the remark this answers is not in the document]
mchmarny: Skipping — the premise does not hold: the aicr binary build and the validator image builds run *before* this `azure/login` step, not after it. Between login and `Bringup Infra` there are only the pre-batch guard (which itself needs az auth, so login must precede it), `setup-build-tools`, one `ko build`, and a `yq` edit — a few minutes against the ~60–75 min access-token lifetime. Every long phase after bringup already re-runs `azure/login` (before install, validate, CUJ, and teardown).

.github/workflows/uat-azure.yaml
```diff
@@ -0,0 +1,738 @@
...
+            -e AUTO_APPROVE=true \
+            -v "$HOME/.azure:/root/.azure" \
+            "${AKS_ACTUATOR_IMAGE}" apply
+
+      # Connect. kubelogin ships with `az aks install-cli`; when the cluster
+      # kubeconfig uses Entra RBAC (exec plugin), convert it to azurecli mode
+      # so kubectl reuses the same az session the per-phase re-logins refresh.
+      - name: Install kubelogin
+        if: steps.infra.outcome == 'success'
+        run: sudo az aks install-cli
```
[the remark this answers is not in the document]
mchmarny: Fixed in 0b47ed43 — `kubelogin` is now pinned in `.settings.yaml` (renovate-tracked against Azure/kubelogin), exposed through `load-versions`, and passed via `--kubelogin-version`. Also passing `--client-version` with the pinned kubectl version, since `az aks install-cli` would otherwise install latest kubectl over the one `setup-build-tools` pinned.

tests/uat/azure/cluster-config.yaml
```diff
@@ -0,0 +1,75 @@
...
+          maxNodes: 3
+      workers:
+        - name: cpuworker1
+          vmSize: Standard_D8s_v6 # 2 vCPU/8 GiB
+          osDiskType: Managed
+          osDiskSizeGb: 200
+          autoscaling:
+            enabled: true
+            minNodes: 1
+            maxNodes: 1
```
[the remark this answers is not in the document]
mchmarny: The pool sizing is correct; the comments were stale. Verified against the live subscription (`az vm list-usage --location westus`): `StandardDsv6Family` limit is **200 vCPUs** (not 10), so the 3×D8 system pool + 1×D8 CPU worker (32 vCPUs total) fits comfortably. Fixed the stale "10 vCPU per family" claim and the wrong "2 vCPU/8 GiB" annotation (Standard_D8s_v6 is 8 vCPU / 32 GiB) in 0b47ed43.

infra/uat-azure-account/variables.tf
```diff
@@ -0,0 +1,38 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+variable "tenant_id" {
+  description = "Entra ID tenant that hosts the app registration. Discover with `az account show --query tenantId -o tsv`."
+  type        = string
+  default     = "43083d15-7273-40c1-b7db-39efd9ccc17a"
```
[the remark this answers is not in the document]
mchmarny: Skipping — the default is deliberate, not hidden state. This directory (`infra/uat-azure-account`) provisions exactly one environment: the NVIDIA UAT tenant/subscription that `tests/uat/azure/cluster-config.yaml` and `uat-azure.yaml` are hard-wired to. Pinning the tenant id in code (alongside the `subscription_id` default two stanzas down, and the same identifiers inlined in the workflow env) is what makes the setup reproducible and reviewable; it is an identifier, not a secret. An operator running this from a different tenant would fail fast at `azuread` provider auth, not silently provision elsewhere.

infra/uat-azure-account/README.md
```diff
@@ -0,0 +1,95 @@
...
+
+**No client secret is created** — authentication is OIDC-only. GitHub mints a short-lived token per workflow run, `azure/login` exchanges it against the federated credentials above, and the resulting `az` CLI context is what the AKS actuator container consumes.
+
+## Usage
+
+```bash
+cd infra/uat-azure-account
+terraform init
+terraform plan
+terraform apply
```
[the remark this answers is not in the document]
mchmarny: Skipping — premise follows from the `variables.tf` suggestion above, which we are not taking: `tenant_id` keeps its default (this directory targets exactly one environment), so plain `terraform plan` / `terraform apply` is correct and non-interactive as documented.

.github/workflows/uat-azure.yaml
```diff
@@ -0,0 +1,765 @@
...
+        id: install
+        if: steps.prep.outcome == 'success'
+        # helmfile apply (up to HELMFILE_TIMEOUT_SECONDS, 20m) + the post-install
+        # readiness gate, which runs `aicr validate --phase deployment` until it
+        # passes READINESS_CONSECUTIVE_PASSES times (up to READINESS_TIMEOUT_SECONDS,
+        # 60m, spanning nodewright tuning + reboots, which can run past 30m on a
+        # cold GPU node). Sized to exceed one helmfile attempt + the full gate
+        # window (20 + 60 = 80, + margin) so the gate's own fail-closed path runs
+        # before GitHub Actions kills the step.
+        timeout-minutes: 90
```
njhensley: **🟠 Major — install (90m) can outlive the non-refreshable federated Azure token; no mid-phase re-auth**

`azure/login` exchanges the single-use GitHub OIDC token for an az session via a client-credentials (federated) grant — which issues no refresh token, so the session cannot self-refresh (the PR's own stated premise, and why each long phase re-runs `azure/login`). `kubelogin` azurecli mode pulls a token from that context on every `kubectl` call. But `install` has ONE preceding refresh and a 90m budget (helmfile ~20m + the readiness gate polling `aicr validate --phase deployment` up to READINESS_TIMEOUT_SECONDS=60m). Once the initial access token expires (~60–75m in), the readiness gate's `kubectl` calls start failing and cannot mint a new token mid-step. AWS/GCP cap `install` at 60m (≈ token lifetime); Azure widened it to 90m precisely for the slow cold-GPU bring-up that would trigger this.

**Blast radius:** Fails safe — an install failure still fires the `always()` teardown, so no GPU-node leak; it just burns an H100 run and defeats the purpose of the widened budget. Manifests once azure-h100 hits a long readiness window.

**Fix:** Re-run `azure/login` inside/around the readiness-gate loop (`tests/uat/azure/run`), split `install` into helmfile-apply + a separately-refreshed readiness step, or cap at ≤60m. At minimum, document that 90m exceeds the federated token lifetime so it isn't copied to future clouds.

infra/uat-azure-account/federation.tf
```diff
@@ -0,0 +1,128 @@
...
+  condition            = "((!(ActionMatches{'Microsoft.Authorization/roleAssignments/write'})) OR (@Request[Microsoft.Authorization/roleAssignments:RoleDefinitionId] ForAnyOfAnyValues:GuidEquals {4d97b98b-1d4f-4787-a291-c67834d212e7, 7f951dda-4ed3-4680-a7ca-43fe172d538d})) AND ((!(ActionMatches{'Microsoft.Authorization/roleAssignments/delete'})) OR (@Resource[Microsoft.Authorization/roleAssignments:RoleDefinitionId] ForAnyOfAnyValues:GuidEquals {4d97b98b-1d4f-4787-a291-c67834d212e7, 7f951dda-4ed3-4680-a7ca-43fe172d538d}))"
+}
+
+# Data-plane Kubernetes access. The actuator provisions AKS with Entra RBAC
+# and local accounts disabled, so `az aks get-credentials` alone grants
+# nothing: without a Kubernetes data-plane role the pipeline's first kubectl
+# call is Forbidden (the managed identity is not in the cluster-config's
+# adminGroups Entra group). Grant the built-in AKS RBAC Cluster Admin role
+# so kubelogin's azurecli mode can authorize kubectl for the UAT phases.
+resource "azurerm_role_assignment" "aks_rbac_cluster_admin" {
```
njhensley: **🟡 Minor — Subscription-scoped AKS RBAC Cluster Admin + Contributor in a subscription documented as shared**

Both `aks_rbac_cluster_admin` and `contributor` are scoped to `data.azurerm_subscription.current.id`. "Azure Kubernetes Service RBAC Cluster Admin" at subscription scope grants the cluster-admin ClusterRole on EVERY Azure-RBAC-enabled AKS cluster in the subscription — not just the run-scoped `aicr-<run_id>` clusters. `tests/uat/azure/cluster-config.yaml` documents the vCPU quota as "capped at 200 vCPUs subscription-wide (shared with other clusters)", i.e. the subscription is not UAT-dedicated. A compromise of the CI managed identity (or the mounted actuator container) would yield kubectl cluster-admin on co-tenant Azure-RBAC clusters plus Contributor CRUD on their RGs.

**Blast radius:** Data-plane cluster-admin + Contributor blast radius extends to co-tenant clusters if the CI identity is compromised. Contributor-at-subscription is justified in the README (no pre-created RG); this mirrors GCP's project-scoped grants, so may be accepted org posture — the data-plane grant is the sharpest edge.

**Fix:** If the subscription genuinely hosts other Azure-RBAC clusters, have the actuator self-assign "AKS RBAC Cluster Admin" scoped to each run's RG (the same iam.tf mechanism it already uses for Network Contributor/AcrPull). If the subscription is UAT-dedicated, state that in the README and this is fine as-is.

pkg/uatbroker/model.go
```diff
@@ -68,13 +74,17 @@ type Reservation struct {
...
+// pre-DC3 behavior, so an un-annotated reservation keeps running the training
+// CUJ nightly. An EXPLICIT empty list ([]) is a nightly opt-out and returns
+// empty: the reservation stays manually dispatchable through uat-run.yaml but
+// the nightly batch skips it (used for bring-up of a new cloud before its
+// pipeline has earned nightly enrollment). Validate guarantees any listed
+// value is a recognized, non-duplicate intent. The returned slice is a fresh
+// copy the caller may mutate freely.
 func (r *Reservation) NightlyIntentsOrDefault() []string {
-	if len(r.NightlyIntents) == 0 {
+	if r.NightlyIntents == nil {
```
njhensley, line 94: **🔵 Nitpick — Bare `nightly-intents:` (YAML null) silently opts INTO the nightly batch; only `[]` opts out**

The opt-out contract keys on `r.NightlyIntents == nil`. yaml.v3 decodes a bare `nightly-intents:` (explicit null) to nil — identical to absent — so it resolves to [training] (opt-IN), while only a non-nil `[]` opts out. `KnownFields(true)` won't catch it (the key is valid). A maintainer who deletes the `[]` value but leaves the key on a bring-up row would silently re-enroll it in the nightly training batch, provisioning real GPU capacity, with no parse error.

**Blast radius:** Low likelihood (committed file + docs consistently use `[]`), but a sharp authoring edge that provisions expensive GPU capacity unintentionally.

**Fix:** Add a doc caveat ("use `[]`, not a bare/null value") or a ParseRegistry test locking the null→[training] behavior.

.github/workflows/uat-azure.yaml
```diff
@@ -0,0 +1,765 @@
...
+          SUMMARY_LIFECYCLE: ${{ inputs.lifecycle }}
+        run: |
+          {
+            echo "## UAT Results (Azure)"
+            echo ""
+            printf '**Reservation:** `%s` · **Intent:** `%s` · **Lifecycle:** `%s`\n' \
+              "$SUMMARY_RESERVATION" "$SUMMARY_INTENT" "$SUMMARY_LIFECYCLE"
+            printf '**Cluster:** `%s`\n' "$DEPLOYMENT_ID"
+            printf '**AICR version:** `%s`\n' "${SUMMARY_AICR_VERSION:-main (build from source)}"
+            echo "**Build:** \`${{ github.sha }}\` (branch: \`${{ github.ref_name }}\`)"
```
njhensley: **🔵 Nitpick — github.sha / github.ref_name inlined into run/summary blocks, against the file's own convention**

The Build step deliberately routes `github.sha` through a `COMMIT_SHA` env var citing "never inline ${{ }} in a shell run — expression injection" — but the Test Summary (`${{ github.ref_name }}`) and the ko step (`--image-label=...${{ github.sha }}`) inline directly. `github.sha` is 40-hex (safe); `github.ref_name` can contain shell-active characters and is only bounded by the federation restricting dispatch to main/test/uat.

**Blast radius:** Low severity, pre-existing pattern in uat-aws/gcp, but zizmor/CodeRabbit will flag it.

**Fix:** Route both through `env` for consistency (optionally sweep the sibling workflows).

.github/workflows/uat-azure.yaml
```diff
@@ -0,0 +1,765 @@
...
+        timeout-minutes: 15
+        shell: bash
+        env:
+          AICR_BIN: ${{ github.workspace }}/aicr
+          RUN_ID: ${{ github.run_id }}
+          # Snapshot agent: main cell = the PR-built image (see "Build and push
+          # snapshot-agent image"); release cell = the released aicr image at that
+          # version. Honored by `aicr snapshot`.
+          AICR_IMAGE: ${{ inputs.aicr_version == '' && format('ghcr.io/nvidia/aicr-validators/agent:{0}', env.VALIDATOR_TAG) || format('ghcr.io/nvidia/aicr:{0}', inputs.aicr_version) }}
+        run: ./tests/uat/azure/run prep "${TEST_CONFIG}"
```
njhensley, line 502: **🔵 Nitpick — tests/uat/azure/run + h100-<intent>-config.yaml are absent (this is PR 1/2)**

Every phase runs `./tests/uat/azure/run <phase>`, and TEST_CONFIG points at `tests/uat/azure/tests/h100-<intent>-config.yaml` — none of which exist yet (they ship in PR 2). This is declared and fails closed: a normal run hits the `Validate inputs` guard (`! -f "${TEST_CONFIG}"`, L162) BEFORE provisioning; only the `skip_tests` burn-in path runs (and tears down cleanly).

**Blast radius:** No leak; expected 1/2 incompleteness. But if `nightly-intents` is flipped off `[]` before PR 2 lands, the enrolled nightly leg fails every night at `Validate inputs`.

**Fix:** No change needed here — gate flipping `nightly-intents` off `[]` on PR 2 landing.

.github/workflows/uat-azure.yaml
```diff
@@ -0,0 +1,765 @@
...
+      # Digest-pinned OCI ref of the signed evidence bundle, consumed by
+      # the ingest-evidence job below. Empty when no bundle was produced.
+      bundle_ref: ${{ steps.evidence_ref.outputs.ref }}
+    env:
+      # Entra federation identifiers (infra/uat-azure-account terraform
+      # output; the client id is the github-actions-aicr user-assigned
+      # managed identity). Identifiers, NOT secrets — the federated-credential
+      # subject pins (main + test/uat refs of NVIDIA/aicr), not knowledge of
+      # these ids, gate access.
+      AZURE_CLIENT_ID: "047f0427-5dcf-4e7c-97c3-9c71d92c5abb"
```
njhensley, line 98: **🔵 Nitpick — PR description says AZURE_* are all-zeros placeholders, but committed workflow ships real GUIDs**

The committed env holds real IDs (AZURE_CLIENT_ID=047f0427-…, TENANT=43083d15-…, SUBSCRIPTION=e88faa01-…), which contradicts the PR body's "all-zeros until terraform apply fast-follow." The committed values are correct and required for the burn-in `azure/login` to work, and are non-secret identifiers (access is gated by the federated-credential subject pins, not knowledge of the IDs).

**Blast radius:** None functional — committed content is correct and necessary. Only a reviewer-trust/accuracy gap between the PR body and the diff.

**Fix:** Update the PR description to match (the aws/gcp precedent already commits real non-secret IDs); drop the all-zeros/placeholder wording.

njhensley (commented):
# 🔬 Multi-Persona Review — Azure AKS UAT (1/2)

**Method:** 4 independent persona reviewers (Correctness/Go · Security/OIDC-IAM · Operability/CI-DX · Domain & Architecture) → an adversarial senior meta-review re-deriving each finding from the resolved code at `53c43ec`. Verdicts: **CONFIRMED** (independently reproduced) / **PLAUSIBLE** (real but hinges on an unverified runtime/tenancy fact).

## Overall assessment

This is a careful, internally consistent third-cloud addition that faithfully mirrors the already-reviewed AWS/GCP UAT pattern — and on the IAM axis is actually *tighter* than the GCP baseline (ABAC-constrained RBAC-admin limited to two role GUIDs vs GCP's project-scoped `projectIamAdmin`). The Go changes (nil-vs-`[]` nightly-intents opt-out, optional `reservation-id`, strict `KnownFields` decode) are correct and well-tested; the nil-vs-empty distinction was empirically confirmed against yaml.v3. Every teardown path fires when a cluster was provisioned, and the only reachable green path in this 1/2 PR (manual `skip_tests` burn-in) is safe. **No blockers.** One fail-safe reliability concern worth a look before nightly enrollment, plus a least-privilege question and some hygiene nits (inline).

## Confirmed non-issues (examined, cleared)

- **nil-vs-`[]` core mechanism** — verified against yaml.v3: `[]`→non-nil empty, absent→nil; the `len()==0`→`==nil` switch is correct both directions, and the broker/nightly-batch missing-key-vs-empty-value handling is sound.
- **`KnownFields(true)` strict decode** — all committed reservation keys map to struct fields; won't break `reservations.yaml`; io.EOF empty-doc path returns the canonical validation error.
- **Evidence trust regex** — all four copies (`evidence-ingest.yaml`, `classify.go`, `allowlist.yaml`, `uat-azure.yaml` `EXPECTED_IDENTITY_REGEXP`) add `azure` to the same fully-anchored `^…uat-(aws|gcp|azure)\.yaml@refs/heads/.+$`; `refs/heads/.+` breadth is pre-existing, not widened.
- **OIDC federation & ABAC** — two exact-match FICs (`main`, `test/uat`), no wildcards → tags/PRs/forks fail closed; the ABAC condition constrains both `roleAssignments/write` *and* `/delete` to the same two GUIDs.
- **Action pinning** — every third-party action + the AKS actuator image are SHA/digest-pinned.
- **Teardown / leak safety** — `Destroy Cluster` fires on every provisioned path, correctly skips only daytime-up (intentional hold), self-contained, retries 3× and fails closed; 270m job budget leaves teardown headroom.
- **Pre-batch guard** — `az aks show` parsing fails closed on non-definitive (throttle/auth) errors.
- **ingest recipe naming** — `aks-h100-<intent>` is consistent with `eks-h100`/`gke-h100`.
- **Docs "serve disabled in both cloud workflows pending #1644"** — the aws/gcp serve steps are actually *live* and `phase_serve` is un-guarded, so this sentence looks inaccurate — **but it is pre-existing text carried unchanged across the diff**, not introduced by this PR. Out of scope; worth a separate tidy if you want to correct it while in the file.

## Tier table

|:---------:|:-------:|:-------:|:---------:|
| 0 | 1 | 1 | 4 |

**Recommendation: Approve with comments** — no blockers; the 🟠 is a fail-safe reliability issue to resolve before nightly enrollment, and the 🟡 is a least-privilege question likely closed by confirming subscription tenancy.

<sub>🤖 Multi-persona review (Correctness · Security · Operability · Domain) + adversarial meta-review.</sub>

.github/workflows/uat-azure.yaml
mchmarny: Fixed in 9c9a4a55 — took the cap option: `install` is now `timeout-minutes: 60`, matching the AWS/GCP install caps and one federated-token lifetime, with the step comment documenting exactly this constraint and warning "do NOT copy a >60m phase budget to future clouds without mid-phase re-auth". Re-widening for slow cold-GPU bring-up comes with PR 2, where the readiness loop in `tests/uat/azure/run` re-logins mid-phase via `az login --federated-token` with a freshly minted ACTIONS_ID_TOKEN (GitHub lets a job mint OIDC tokens repeatedly). Job-budget arithmetic updated (190→160 UAT minutes).

infra/uat-azure-account/federation.tf
mchmarny: Fixed in 9c9a4a55, per your first option (adapted to run in the workflow rather than the pinned actuator): the subscription is indeed shared, so the standing subscription-scoped "AKS RBAC Cluster Admin" grant is gone. The ABAC condition on the RBAC Administrator grant now covers three role definitions (Network Contributor, AcrPull, + `b1ff04bb…` AKS RBAC Cluster Admin), and a new "Grant run-scoped cluster admin" step self-assigns the role scoped to the run's resource group right after Bringup — idempotent for daytime re-runs, deleted with the RG at teardown, with the Connect step retrying kubectl while the assignment propagates. README updated; the terraform change is applied to the live account (old subscription-scoped assignment destroyed).

pkg/uatbroker/model.go
mchmarny: Fixed in 9c9a4a55 — both suggestions: an AUTHORING CAVEAT block on `Reservation.NightlyIntents` (use `[]`, not a bare/null value; bare null = opt-IN via the training default) and `TestParseRegistryBareNullNightlyIntents` locking the null→`[training]` decode behavior so a semantic change surfaces as a test failure.

.github/workflows/uat-azure.yaml
mchmarny: Fixed in 9c9a4a55 — the ko step routes `github.sha` through a `COMMIT_SHA` env var, and the Test Summary routes `github.sha`/`github.ref_name` through `SUMMARY_SHA`/`SUMMARY_REF` (the ref_name one being the actually shell-active case). Sweeping the same pattern in uat-aws/uat-gcp is left for a follow-up since it's pre-existing there.

.github/workflows/uat-azure.yaml
mchmarny: Confirmed, no change — exactly the intended 1/2 staging: the `Validate inputs` guard fails closed before provisioning for normal runs, only the `skip_tests` burn-in path executes, and `azure-h100` stays `nightly-intents: []` until PR 2 lands (now stated explicitly in the PR description). Flipping the nightly enrollment is gated on PR 2.

.github/workflows/uat-azure.yaml
mchmarny: Fixed — PR description updated: the placeholder/all-zeros wording is gone; it now states the federation is applied and the committed workflow ships the real non-secret identifiers (same posture as the AWS account ID), with access gated by the federated-credential subject pins.

njhensley (approved):
## ✅ Approve — all review findings resolved (`109b34d`)

Re-reviewed against the two commits pushed since the multi-persona pass at `53c43ec`. Every finding is resolved, and the new code is correct:

- **🟠 install token-expiry** — install capped to **60m** (matches AWS/GCP), with a comment documenting the federated-token lifetime and that any future widening requires mid-phase re-auth (PR 2).
- **🟡 subscription-scoped AKS RBAC Cluster Admin** — replaced with a **per-run, RG-scoped self-assignment** ("Grant run-scoped cluster admin"): idempotent, fails closed, gated to `steps.infra.outcome == 'success'` (skipped for `daytime-down`), and torn down with the RG. ABAC condition widened to permit exactly that third role definition (`b1ff04bb…`, the built-in role). This genuinely raises the security posture over the original.
- **🔵 bare `nightly-intents:` null footgun** — authoring-caveat doc comment + `TestParseRegistryBareNullNightlyIntents`.
- **🔵 inlined `github.sha`/`github.ref_name`** — both routed through `env` (`COMMIT_SHA`, `SUMMARY_SHA`/`SUMMARY_REF` + `printf`).
- **🔵 PR body placeholder wording** — corrected to "Federation identifiers are live".

Also reviewed the new Connect retry loop (absorbs RBAC-assignment propagation delay) — sound.

Remaining work is the tracked **PR 2/2** (`tests/uat/azure/run` + per-intent configs); `azure-h100` correctly stays `nightly-intents: []` until then, and every real-run path fails closed at `Validate inputs` in the meantime.

<sub>🤖 Multi-persona review (Correctness · Security · Operability · Domain) + adversarial meta-review — re-review pass.</sub>

PR #1714, opened 2026-07-10
fix(validator): validate NCCL fabric resource homogeneity across nodes

### LINKED ISSUE #1710: NCCL fabric resource discovery: validate homogeneity across target node set (EKS + AKS)
## Summary

The NCCL all-reduce performance validator derives the fabric device-plugin resource count from only the **first** resolved GPU node, then applies that count to every worker in the TrainJob. This is true for both services:

- **EKS** — `discoverEKSNodeConfig(config.Nodes[0])` reads `vpc.amazonaws.com/efa` from `Nodes[0]`
- **AKS** — `discoverAKSRdmaCount(config.Nodes[0])` reads `rdma/hca_shared_devices_a` from `Nodes[0]`

`warnIfHeterogeneousNodes` only *warns* on `instance-type` label mismatch — it does not check the fabric resource itself.

## Why it matters

On a homogeneous, fully-converged cluster this is correct (the device plugin advertises the same count on every GPU node). But on a **mixed or partially-rolled-out** node set — e.g. the RDMA/EFA device plugin present on some target nodes but not others — sampling `Nodes[0]` applies one node's fabric assumption to all workers. Today the behavior fails *loudly* (a worker requesting a resource its node lacks goes unschedulable; TCP fallback misses the bandwidth floor), so it is not a silent-pass hazard — but it produces a confusing failure rather than a clear diagnostic, and the single-node assumption is fragile.

## Proposed change

Make NCCL fabric resource discovery **validate homogeneous adapter/device-plugin availability across the full resolved target node set** for both EKS and AKS: scan all `config.Nodes`, confirm the fabric resource count is consistent, and **fail closed with a clear error** on mixed/partial rollout (e.g. "RDMA/EFA present on X of Y target GPU nodes — cluster still converging or device plugin not fully rolled out; re-run once uniform") instead of silently using `Nodes[0]`.

## Scope

- `validators/performance/nccl_eks_utils.go` (`discoverEKSNodeConfig`)
- `validators/performance/nccl_aks_utils.go` (`discoverAKSRdmaCount` / `applyAKSTemplateData`)
- Keep the existing zero-count TCP fallback (uniformly-zero across all nodes is still valid → TCP).

## Context

Follow-up from PR #1695 (AKS H100 NCCL support). Raised in CodeRabbit review; deliberately deferred to keep #1695 scoped to enabling the AKS path, and to cover EKS symmetrically rather than fixing only AKS.

yuanchen8911:
## Summary

Make the NCCL all-reduce performance validator validate fabric device-plugin resource homogeneity across the full resolved target GPU node set (EKS EFA and AKS shared RDMA), failing closed on a mixed/partial rollout instead of silently sizing every worker from `config.Nodes[0]`.

## Motivation / Context

The validator derived the fabric resource count from only the first resolved GPU node and applied it to every worker in the TrainJob:

- EKS — `discoverEKSNodeConfig(config.Nodes[0])` read `vpc.amazonaws.com/efa`
- AKS — `discoverAKSRdmaCount(config.Nodes[0])` read `rdma/hca_shared_devices_a`

`warnIfHeterogeneousNodes` only *warns* on `instance-type` label mismatch — it never checked the fabric resource itself. On a mixed or partially-rolled-out node set (device plugin present on some target nodes but not others) this applied one node's fabric assumption to all workers, producing a confusing unschedulable/bandwidth failure rather than a clear diagnostic.

Fixes: #1710
Related: #1695

This is the follow-up carved out of PR #1695 (AKS H100 NCCL support): the homogeneity gate was deliberately deferred there to keep that PR scoped to enabling the AKS path, and to cover EKS symmetrically rather than fixing only AKS.

## Type of Change

## Component(s) Affected

## Implementation Notes

- New shared helper `uniformFabricResourceCount(nodes, resource)` in `nccl_eks_utils.go` scans all `config.Nodes`, requires the extended-resource allocatable count to be uniform across the cohort, and returns it. A uniform zero is valid and preserves the existing TCP fallback (reduced max message size). A disagreement returns a `pkg/errors` `ErrCodeInternal` with a clear message: `<resource> present on X of Y target GPU nodes — cluster still converging or device plugin not fully rolled out; re-run once uniform`.
- Parameterized by resource name so EKS (`vpc.amazonaws.com/efa`) and AKS (`rdma/hca_shared_devices_a`) reuse the same logic without duplicating the scan/verify.
- `discoverEKSNodeConfig` now takes the full `[]v1.Node` (instance type still from the first node; EFA count via the shared helper). `applyAKSTemplateData` now returns an `error`; the single-node `discoverAKSRdmaCount` is removed (superseded by the helper). Both call sites in `nccl_all_reduce_bw_constraint.go` propagate the error so the validator fails closed.

## Testing

```bash
unset GITLAB_TOKEN
go test -race ./validators/performance/...
golangci-lint run -c .golangci.yaml ./validators/performance/...
make qualify
make test-coverage
```

- `go test -race ./validators/performance/...`: pass (full package passes with sandbox network access for the unrelated httptest port-bind test).
- `golangci-lint ./validators/performance/...`: 0 issues.
- `make qualify`: test + lint + e2e (23/23 e2e tests pass) all green. `make scan` reports pre-existing upstream dependency CVEs only (`golang.org/x/net`, `stdlib` go1.26.x, `runc`); this change adds no dependencies, so the scan finding is pre-existing on `main` and unrelated.
- `make test-coverage`: passed — total 78.8% (threshold 75%).
- Coverage delta (`validators/performance`): 50.7% → 51.0% (package). New/changed funcs `uniformFabricResourceCount`, `discoverEKSNodeConfig`, `applyAKSTemplateData` are each at 100%.
- Added/extended table-driven tests: uniform-nonzero (proceed), uniform-zero (TCP fallback), mixed present/absent and mixed-nonzero (fail closed), and the empty-node-set guard, for both EKS and AKS.

## Risk Assessment

- [x] **Low** — Isolated change to the NCCL performance validator; well-tested; easy to revert. Uniform (converged) clusters behave identically to before; only mixed/partial rollouts now fail closed with a clear diagnostic instead of a confusing downstream failure.

**Rollout notes:** N/A — no API, recipe, or dependency changes.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — no user-facing surface changed)

### DIFF
--- a/validators/performance/nccl_eks_utils.go
+++ b/validators/performance/nccl_eks_utils.go
@@ -43,21 +43,77 @@ func warnIfHeterogeneousNodes(nodes []v1.Node) {
 	}
 }
 
-// discoverEKSNodeConfig reads the instance type label and EFA adapter count
-// from a GPU node. Returns an error if the label is missing. EFA count of 0
-// is valid (device plugin not installed — NCCL falls back to TCP).
-func discoverEKSNodeConfig(node v1.Node) (string, int, error) {
-	instanceType := node.Labels["node.kubernetes.io/instance-type"]
+// efaResourceName is the EFA extended resource published by the AWS EFA
+// Kubernetes device plugin on EKS GPU nodes.
+const efaResourceName = v1.ResourceName("vpc.amazonaws.com/efa")
+
+// discoverEKSNodeConfig reads the instance type label (from the first target
+// node) and the EFA adapter count validated as uniform across all target GPU
+// nodes. Returns an error if the instance-type label is missing or if the EFA
+// count is not uniform across the cohort (partial device-plugin rollout). An
+// EFA count of 0 is valid when uniform (device plugin not installed — NCCL
+// falls back to TCP).
+func discoverEKSNodeConfig(nodes []v1.Node) (string, int, error) {
+	if len(nodes) == 0 {
+		return "", 0, aicrErrors.New(aicrErrors.ErrCodeInternal,
+			"no target GPU nodes for EKS discovery")
+	}
+
+	instanceType := nodes[0].Labels["node.kubernetes.io/instance-type"]
 	if instanceType == "" {
 		return "", 0, aicrErrors.New(aicrErrors.ErrCodeInternal,
 			"GPU node missing node.kubernetes.io/instance-type label")
 	}
 
-	efaResource := v1.ResourceName("vpc.amazonaws.com/efa")
-	efaQuantity := node.Status.Allocatable[efaResource]
-	efaCount := int(efaQuantity.Value())
+	efaCount, err := uniformFabricResourceCount(nodes, efaResourceName)
+	if err != nil {
+		return "", 0, err
+	}
 
 	return instanceType, efaCount, nil
+}
+
+// uniformFabricResourceCount returns the allocatable count of an extended
+// fabric resource (EFA on EKS, the shared RDMA pool on AKS) across the full set
+// of resolved target GPU nodes, requiring every node to advertise the SAME
+// count. NCCL sizes every worker identically, so deriving the count from only
+// the first node silently mis-sizes the fabric request during a partial
+// device-plugin rollout or a mixed-instance cohort.
+//
+// A uniform count is returned as-is; a uniform zero is valid and selects the
+// existing TCP fallback. A disagreement across nodes (mixed/partial rollout)
+// returns an error so the validator fails closed and the operator re-runs once
+// the cluster has converged.
+func uniformFabricResourceCount(nodes []v1.Node, resource v1.ResourceName) (int, error) {
+	if len(nodes) == 0 {
+		return 0, aicrErrors.New(aicrErrors.ErrCodeInternal,
+			fmt.Sprintf("no target GPU nodes to inspect for %s", resource))
+	}
+
+	firstQuantity := nodes[0].Status.Allocatable[resource]
+	first := int(firstQuantity.Value())
+
+	present := 0
+	uniform := true
+	for i := range nodes {
+		quantity := nodes[i].Status.Allocatable[resource]
+		count := int(quantity.Value())
+		if count > 0 {
+			present++
+		}
+		if count != first {
+			uniform = false
+		}
+	}
+
+	if !uniform {
+		return 0, aicrErrors.New(aicrErrors.ErrCodeInternal,
+			fmt.Sprintf("%s present on %d of %d target GPU nodes — cluster still converging "+
+				"or device plugin not fully rolled out; re-run once uniform",
+				resource, present, len(nodes)))
+	}
+
+	return first, nil
 }
 
 // buildEFAResourceLine returns the YAML line for EFA resource requests/limits
--- a/validators/performance/nccl_eks_utils_test.go
+++ b/validators/performance/nccl_eks_utils_test.go
@@ -23,87 +23,83 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// eksNode builds an EKS GPU node with the given instance type and EFA count.
+// A negative efa omits the EFA resource entirely (device plugin absent).
+func eksNode(name, instanceType string, efa int) v1.Node {
+	labels := map[string]string{}
+	if instanceType != "" {
+		labels["node.kubernetes.io/instance-type"] = instanceType
+	}
+	alloc := v1.ResourceList{
+		v1.ResourceName("nvidia.com/gpu"): resource.MustParse("8"),
+	}
+	if efa >= 0 {
+		alloc[efaResourceName] = *resource.NewQuantity(int64(efa), resource.DecimalSI)
+	}
+	return v1.Node{
+		ObjectMeta: metav1.ObjectMeta{Name: name, Labels: labels},
+		Status:     v1.NodeStatus{Allocatable: alloc},
+	}
+}
+
 func TestDiscoverEKSNodeConfig(t *testing.T) {
 	tests := []struct {
 		name         string
-		node         v1.Node
+		nodes        []v1.Node
 		wantInstance string
 		wantEFA      int
 		wantErr      bool
 		wantErrMsg   string
 	}{
 		{
-			name: "p5.48xlarge with 32 EFA",
-			node: v1.Node{
-				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{
-						"node.kubernetes.io/instance-type": "p5.48xlarge",
-					},
-				},
-				Status: v1.NodeStatus{
-					Allocatable: v1.ResourceList{
-						v1.ResourceName("nvidia.com/gpu"):        resource.MustParse("8"),
-						v1.ResourceName("vpc.amazonaws.com/efa"): resource.MustParse("32"),
-					},
-				},
+			name: "uniform p5.48xlarge with 32 EFA",
+			nodes: []v1.Node{
+				eksNode("n1", "p5.48xlarge", 32),
+				eksNode("n2", "p5.48xlarge", 32),
 			},
 			wantInstance: "p5.48xlarge",
 			wantEFA:      32,
 		},
 		{
-			name: "p4d.24xlarge with 4 EFA",
-			node: v1.Node{
-				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{
-						"node.kubernetes.io/instance-type": "p4d.24xlarge",
-					},
-				},
-				Status: v1.NodeStatus{
-					Allocatable: v1.ResourceList{
-						v1.ResourceName("nvidia.com/gpu"):        resource.MustParse("8"),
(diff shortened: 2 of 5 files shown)


validators/performance/nccl_eks_utils.go
```diff
@@ -43,23 +43,79 @@ func warnIfHeterogeneousNodes(nodes []v1.Node) {
...
+		if count != first {
+			uniform = false
+		}
+	}
+
+	if !uniform {
+		return 0, aicrErrors.New(aicrErrors.ErrCodeInternal,
+			fmt.Sprintf("%s present on %d of %d target GPU nodes — cluster still converging "+
+				"or device plugin not fully rolled out; re-run once uniform",
+				resource, present, len(nodes)))
```
mchmarny, line 113: nit: for the mixed-nonzero case (e.g. 32 vs 16, your `mixed nonzero counts` test) this prints "present on 3 of 3 target GPU nodes — cluster still converging or device plugin not fully rolled out", which describes the wrong failure: everything is present, the counts just disagree. Since this PR's whole point is a clear diagnostic, consider including the differing counts (e.g. min/max or a per-count tally) so the operator sees *what* disagreed, not just that something did.

mchmarny (approved):
Right shape for the fix: one shared `uniformFabricResourceCount` helper covering EKS EFA and AKS RDMA symmetrically, fail-closed on disagreement, uniform-zero preserving the TCP fallback, and both call sites propagating the error. Test coverage is thorough — uniform/zero/absent/mixed/empty for both paths, plus the helper's own table including the mixed-nonzero case. One nit inline on the diagnostic wording for mixed-nonzero counts; nothing blocking. CI green.

validators/performance/nccl_eks_utils.go
```diff
@@ -43,23 +43,79 @@ func warnIfHeterogeneousNodes(nodes []v1.Node) {
...
+		quantity := nodes[i].Status.Allocatable[resource]
+		count := int(quantity.Value())
+		if count > 0 {
+			present++
+		}
+		if count != first {
+			uniform = false
+		}
+	}
+
```
njhensley, line 108: **🟡 Minor — Non-uniform error misdiagnoses heterogeneous counts as a partial rollout**

`present` counts nodes with `count > 0`, but uniformity fails on *any* differing count. For a mixed-nonzero cohort (e.g. EFA `[32,16,32]`) all nodes are present, so the message reads *"present on 3 of 3 target GPU nodes — cluster still converging or device plugin not fully rolled out; re-run once uniform"* — the stated cause is false and *"re-run once uniform"* never resolves a stable heterogeneous cohort. The test pins this wording (`nccl_eks_utils_test.go`, `"mixed nonzero counts"` → `"present on 3 of 3"`).

**Blast radius:** Diagnostic-only, and low reachability: the default EKS path narrows by instance-type → uniform EFA, and AKS is pinned to 1000. Mixed-nonzero is only reachable via a `--node-selector` override spanning instance types (which already warns). For the common absent/present case the message is accurate.

**Fix:** Branch the message — `present < total` → "partial rollout, re-run once converged"; `present == total` with differing counts → "nodes advertise differing counts; NCCL requires uniform fabric — check SKU/instance-type consistency (will not self-resolve)". Reporting the distinct observed values also subsumes the two node-detail nitpicks below.

validators/performance/nccl_eks_utils.go
```diff
@@ -43,23 +43,79 @@ func warnIfHeterogeneousNodes(nodes []v1.Node) {
...
+	if !uniform {
+		return 0, aicrErrors.New(aicrErrors.ErrCodeInternal,
+			fmt.Sprintf("%s present on %d of %d target GPU nodes — cluster still converging "+
```
njhensley, line 111: **🔵 Nitpick — Error names neither the disagreeing nodes nor their values**

The loop already visits each `node[i]` and its `count`, but the error carries only `present`/`len(nodes)`.

**Blast radius:** On a non-uniform cohort the operator must manually inspect allocatable on every target node to find the outlier.

**Fix:** Collect `(node.Name, count)` outliers during the loop and include them in the message.

validators/performance/nccl_eks_utils.go
njhensley, line 108: **🔵 Nitpick — No slog node-level detail before the fail-closed return**

The success path logs (`slog.Info "Discovered ... pool"`), but the failure branch returns the terse error with no structured per-node breakdown — the only in-cluster forensic artifact is the single-line error.

**Blast radius:** Operators can't see the per-node fabric-count distribution that triggered the failure without reproducing manually.

**Fix:** `slog.Warn("non-uniform fabric resource across target nodes", "resource", resource, "counts", perNodeCounts)` before returning.

validators/performance/nccl_eks_utils.go
```diff
@@ -43,23 +43,79 @@ func warnIfHeterogeneousNodes(nodes []v1.Node) {
...
+		count := int(quantity.Value())
+		if count > 0 {
+			present++
+		}
+		if count != first {
+			uniform = false
+		}
+	}
+
+	if !uniform {
```
njhensley, line 109: **🔵 Nitpick — ErrCodeInternal for an operator-actionable / transient cluster-state condition**

Semantically this is external cluster state ("still converging"), for which `ErrCodeUnavailable` reads truer than `ErrCodeInternal`.

**Blast radius:** Low — `validators/performance` is a standalone `package main` in-cluster Job, not imported by `pkg/server`/`pkg/client`, so the 5xx-cause-withholding concern does not apply and the diagnostic surfaces intact. Consistent with sibling errors in the same file.

**Fix:** Optionally use `ErrCodeUnavailable`; purely a hygiene nit.

validators/performance/nccl_eks_utils_test.go
```diff
@@ -124,6 +120,76 @@ func TestDiscoverEKSNodeConfig(t *testing.T) {
...
+			nodes: []v1.Node{node(0), node(0)},
+			want:  0,
+		},
+		{
+			name:  "uniform absent (resource missing)",
+			nodes: []v1.Node{node(-1), node(-1)},
+			want:  0,
+		},
+		{
+			name:       "mixed present and absent",
```
njhensley, line 155: **🔵 Nitpick — Test gap: no case where node[0] is the divergent/absent node**

All non-uniform rows put the odd node last. The code derives `first` from `nodes[0]` and is order-independent, so there's no current defect — but no test locks that symmetry in.

**Blast radius:** A future refactor making divergence detection order-sensitive could slip through green tests.

**Fix:** Add one row with the divergent node first, e.g. `{node(-1), node(32)}` → `"present on 1 of 2"`.

njhensley (approved):
## 🧭 Multi-Persona Review — Approve with notes

**Method:** 4 independent persona reviewers (Correctness · Domain/Architecture · Test-coverage · Operability) → an adversarial senior meta-reviewer that re-derived every finding from the resolved code at `4c01a9b`. The crux finding (AKS fabric semantics) was independently reproduced and **refuted**; two load-bearing facts were verified against the actual config and resolver.

### Assessment

Clean, well-scoped defensive fix. Replacing `Nodes[0]`-based fabric sizing with a cohort-wide uniformity gate (`uniformFabricResourceCount`) that fails closed on a mixed/partial device-plugin rollout — parameterized so EKS EFA and AKS shared-RDMA share one scan — is the right call. The logic is correct and order-independent, error propagation follows the "don't double-wrap already-coded errors" rule, empty-node guards are present on both entry points, and the table-driven tests cover every branch. Uniform/converged clusters behave identically to before — **no regression path found**. 👍

Everything below is diagnostic/observability polish; none blocks merge.

- **🟠→ refuted: "AKS gate wrongly requires exact pool-size uniformity when only presence matters."** Verified: `buildAKSRdmaResourceLine` does request `"1"` regardless of magnitude, but `nic-cluster-policy-aks.yaml` pins `rdmaHcaMax: 1000` in a single cluster-wide `NicClusterPolicy`, so two *healthy* IB nodes cannot advertise different nonzero ceilings — the only reachable violation is presence-mismatch (0 vs 1000), which *should* fail closed. The exact-value check is also genuinely required on EKS, where the count is load-bearing. No false fail-closed.
- **Error-propagation / double-wrapping:** bare `return err` correctly propagates already-`ErrCodeInternal` errors.
- **Empty-node guards / panic safety:** both entry points guard `len==0`; `config.Nodes` is non-empty at runtime (`WorkerCount >= 2` upstream gate).
- **GKE untouched:** not a symmetry gap — GKE claims fabric via cluster-scoped multi-NIC network objects (already gated `>= 8`), not a per-node allocatable.
- **Deleted `discoverAKSRdmaCount` coverage:** fully re-covered (`"1k"`→1000 parse, absent→0) by the new tables.

### Summary

| Tier | Count |
|------|-------|
| 🔴 Blocker | 0 |
| 🟠 Major | 0 |
| 🟡 Minor | 1 |
| 🔵 Nitpick | 4 |

**Approving** — the sole 🟡 is a message-wording improvement on a niche override-only path and can be a follow-up. The homogeneity gate is correct, well-tested, and fails closed exactly as intended.

Issue #1703, opened 2026-07-10
Validation applicability is compiled-in, not data-driven — external --data recipes can't run NCCL/conformance validation


atif1996:
### Summary

`aicr validate` can only run its conformance / performance (NCCL benchmark) phases for recipes that AICR knows from its **embedded** catalog. Recipes introduced purely through external `--data` (a new `criteria.service` value, or a service/accelerator/OS combination not shipped embedded) have the performance phase **SKIPPED** — even when the recipe's own validation block declares the checks.

### Details

The NCCL benchmark applicability (the `variantNVLS` / `variantNET` selection) is **compiled into the binary** and keyed to specific embedded `service`+`accelerator` pairs (e.g. `eks`-GB200, `oke`-GB200). It is not derived from the recipe data. As a result:

- A custom service registered only through an external `--data` overlay has no applicability entry, so `aicr validate` reports the NCCL / performance check as **SKIPPED**.
- Extending an existing service to a new accelerator/OS via `--data` hits the same wall.
- Supplying NCCL limits in the recipe does not help — the gate is the compiled applicability, not the recipe's thresholds.

### Why it matters

External `--data` is the supported way to define recipes on top of the OSS base (new services, internal OS images, new accelerator combinations). If validation applicability is hardcoded to the embedded catalog, those recipes are second-class: they cannot be certified (NCCL / GPUDirect performance) through `aicr validate`, which is the standard gate.

### Proposal

Make validation applicability **data-driven from the recipe** rather than a compiled matrix:

- Let a recipe's `validation` block declare its own conformance/performance applicability, benchmark variant (NVLS/NET), and thresholds, and have `aicr validate` honor it for any recipe loaded via `--data`, not just embedded ones.
- At minimum, allow the NCCL benchmark variant + limits to be specified in recipe data and keyed on the recipe's own `criteria`, so a new `service` can opt into the same benchmark the embedded recipes use.

### Environment

- `aicr validate --recipe <recipe.yaml> --data <dir>` with a recipe whose `criteria.service` is registered only via `--data`.

PR #1721, opened 2026-07-10
fix(validator): warn on unknown check names; fix omitted-checks doc

### LINKED ISSUE #1717: aicr validate silently drops unknown check names in validation checks
### Summary

`aicr validate` silently ignores a check name in `validation.{phase}.checks` that matches no validator catalog entry. There is no error, no warning, and no log line — the phase simply runs without that check and can report **passed**.

### Details

Check selection is a pure name-set intersection: `FilterEntriesByValidation` (`pkg/validator/v1/catalog.go`) builds an `allowed` set from the recipe's declared checks for the phase and keeps only catalog entries whose `name` is in the set. Declared names that match nothing are dropped on the floor. No other layer validates them:

- Recipe load/merge does not check the names against the catalog.
- `WarnPhasesAgainstRecipe` (`pkg/validator/phases.go`) only warns when a *requested phase* declares no checks at all — it never compares individual names.
- A name that exists in the catalog but under a **different phase** is dropped just as silently.

### Why it matters

Validation is a gate. A typo'd check name (`nccl-all-reduce-bw-nvsl`, `expected-resource`) produces a spuriously-passing phase — the fail-open direction for a negative check. This is especially easy to hit with external `--data` catalogs, where a custom validator's `name` in `validators/catalog.yaml` must match the recipe's `checks` entry by hand.

### Proposal

After filtering, diff the declared check list against the matched entries and:

- at minimum `slog.Warn` per unmatched name (naming the phase, and noting when the name exists under a different phase), or
- fail closed with `ErrCodeInvalidRequest`, possibly gated by `--criteria-strict`-style strictness, so a misspelled check cannot certify a cluster.

### Environment

Found while auditing check selection for #1703 (validate path in `pkg/validator/v1/catalog.go`, `pkg/validator/validator.go` `runPhase`).

### LINKED ISSUE #1718: docs: validator-extension.md wrongly claims omitted checks list runs all catalog entries
### Summary

`docs/integrator/validator-extension.md` ("Step 4: Reference in Recipe") states:

> If you omit the `checks` list, all catalog entries for the phase run (embedded + custom).

The code does the opposite: `FilterEntriesByValidation` (`pkg/validator/v1/catalog.go`) returns **nil when the phase declares no checks** ("No checks declared for this phase → skip it"), so an omitted or empty `checks` list runs **zero** validators for that phase. `WarnPhasesAgainstRecipe` (`pkg/validator/phases.go`) warns when an explicitly requested phase has no checks — consistent with skip-, not run-all-, semantics.

### Fix

Correct the sentence to match the code: omitting `checks` skips the phase, and every check to run (embedded or custom) must be listed explicitly. If run-all was the intended contract for integrators, that is a code change to discuss separately — but current behavior, tests, and the empty-list-clears merge semantics (`checks: []` deliberately drops inherited checks) all encode skip.

### Related

- #1717 — unknown check *names* are silently dropped by the same intersection; both issues concern silent divergence between a recipe's `checks` list and what actually runs.
- Found while auditing check selection for #1703.

njhensley:
## Summary

Make `aicr validate` surface check names that match no validator catalog entry (a typo or a check declared under the wrong phase), and correct the validator-extension doc that described the opposite of the actual omitted-`checks` behavior.

## Motivation / Context

Check selection is a pure name-set intersection in `FilterEntriesByValidation`: it keeps only catalog entries whose `name` appears in a phase's declared `checks`, and drops every unmatched name silently — no error, no warning, no log line. A typo'd or misplaced check name therefore produces an empty, spuriously-*passing* phase — the fail-open direction for a validation gate.

Separately, `docs/integrator/validator-extension.md` Step 4 claimed omitting `checks` runs *all* catalog entries for the phase, when the code returns `nil` (skips the phase, runs zero validators) — the opposite.

Fixes: #1717, #1718
Related: #1703

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- New `ValidatorCatalog.UnmatchedChecks(phase, validationInput)` diffs a phase's declared checks against the full catalog and reports each unmatched name, flagging names that exist under a *different* phase as likely misplacements (distinct from outright typos).
- `runPhase` and the no-cluster path call `warnUnmatchedChecks`, emitting a structured `slog.Warn` per unmatched name so typos are visible during both live and offline (`--no-cluster`) validation.
- **Advisory only** — it never fails the run, mirroring the existing `WarnPhasesAgainstRecipe` posture. Failing closed under a strictness flag was considered (see the issue) but kept out of scope to avoid breaking recipes with pre-existing stale names; the warning is the minimal fix for the silent-drop bug.
- Doc rewritten to state every check must be listed explicitly, that an empty list clears inherited checks and skips the phase, and that unmatched names are dropped with a warning.

## Testing

```bash
go test -race ./pkg/validator/...
golangci-lint run -c .golangci.yaml ./pkg/validator/... ./pkg/validator/v1/...
```

- Added `TestUnmatchedChecks` (all-matched, typo, wrong-phase, dedup) and `TestUnmatchedChecks_NilReceiverAndNoChecks`.
- Full `pkg/validator/...` suite passes with `-race`; lint clean (0 issues).

## Risk Assessment

- [x] **Low** — Additive warning + doc fix, no behavior change to what runs, easy to revert.

**Rollout notes:** N/A — no config or API surface change; new log lines only.

## Checklist

### DIFF
--- a/docs/integrator/validator-extension.md
+++ b/docs/integrator/validator-extension.md
@@ -129,7 +129,12 @@ validation:
       - my-custom-check        # Your custom validator
 ```
 
-If you omit the `checks` list, all catalog entries for the phase run (embedded + custom).
+Every check to run must be listed explicitly. If you omit the `checks` list (or
+set it to `[]`, which clears any inherited checks), the phase runs **zero**
+validators and is skipped — an empty phase reports as passing, so a missing entry
+is easy to overlook. A declared name that matches no catalog entry for the phase
+is likewise dropped (it logs a warning but does not fail the run), so keep names
+in sync with the `name` fields in `validators/catalog.yaml`.
 
 ### Step 5: Run Validation
 
--- a/pkg/validator/v1/catalog.go
+++ b/pkg/validator/v1/catalog.go
@@ -145,28 +145,38 @@ func (c *ValidatorCatalog) ForPhase(phase Phase) []ValidatorEntry {
 	return result
 }
 
-// FilterEntriesByValidation filters catalog entries based on the validation's declared checks for the given phase.
-// Returns nil if the validation has no phase configuration or no checks declared.
-func FilterEntriesByValidation(entries []ValidatorEntry, phase Phase, validationInput *ValidationInput) []ValidatorEntry {
+// checksForPhase returns the check names declared for phase in the validation
+// input, or nil when the validation has no configuration for that phase.
+func checksForPhase(phase Phase, validationInput *ValidationInput) []string {
 	if validationInput == nil {
 		return nil
 	}
-
-	var phaseChecks []string
 	switch phase {
 	case PhaseDeployment:
 		if validationInput.Config.Deployment != nil {
-			phaseChecks = validationInput.Config.Deployment.Checks
+			return validationInput.Config.Deployment.Checks
 		}
 	case PhasePerformance:
 		if validationInput.Config.Performance != nil {
-			phaseChecks = validationInput.Config.Performance.Checks
+			return validationInput.Config.Performance.Checks
 		}
 	case PhaseConformance:
 		if validationInput.Config.Conformance != nil {
-			phaseChecks = validationInput.Config.Conformance.Checks
+			return validationInput.Config.Conformance.Checks
 		}
 	}
+	return nil
+}
+
+// FilterEntriesByValidation filters catalog entries based on the validation's declared checks for the given phase.
+// Returns nil if the validation has no phase configuration or no checks declared.
+//
+// Declared check names that match no entry are dropped here; use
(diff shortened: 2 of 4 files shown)


mchmarny (approved):
Closes the fail-open direction of check selection cleanly: unmatched names now warn on both the live path (`runPhase`) and `--no-cluster`, with no double-warn, and the cross-phase misplacement hint is a nice touch over a bare "not found". The advisory-only posture with the fail-closed alternative explicitly deferred (and reasoned about in the issue) is the right call for existing recipes. Doc fix matches the actual `nil`-return behavior. Tests cover typo, wrong-phase, dedup, and nil-receiver paths. No findings — nothing blocking.

Issue #1276, opened 2026-06-09
DC3 — CUJ2 inference E2E in the UAT runner/workflows


njhensley:
**Parent:** #1264 · **Labels:** `enhancement`, `area/tests` (auto), `area/ci`
(auto), `area/docs` (auto) · **Size:** L · **Depends on:** DC2

**Goal.** Bring real-cluster inference to CUJ1 parity: deploy a Dynamo
`DynamoGraphDeployment`, hit the OpenAI-compatible endpoint, capture conformance
evidence, and emit a signed bundle — the inference counterpart of the existing
`phase_train`. DC3 is the gate on the `inference-dynamo` half of the launch scope:
no inference column exists on any cloud until DC3 lands.

**Scope.**
- Add `phase_serve` to `tests/uat/{aws,gcp}/run`, parallel to `phase_train`
  (`tests/uat/aws/run:193-256`): apply the Dynamo deployment, wait for readiness,
  issue a sample OpenAI-compatible request, assert a valid completion, capture
  logs. Select it via the test config / `intent`, so `run all` runs
  `prep → install → conformance → serve → verify` for inference and
  `…→ train →…` for training. The committed `cuj2-inference/chainsaw-test.yaml`
  is today a `--no-cluster` simulation (`cuj2-inference/chainsaw-test.yaml:86-92`)
  and is **not** invoked by any workflow — this child makes inference a *real*
  cluster phase, not a dry run.
- Wire the inference path into the workflows so a `workflow_dispatch`
  `intent=inference` (DC2) reaches `phase_serve`. **Resolve the cron-scheduling
  decision explicitly** (not as a parenthetical): either add a daily inference cron
  lane or alternate the single existing cron between intents. **Do not schedule two
  daily crons against the single AWS reservation** (`cr-0cbe491320188dfa6`) unless
  the superseded-run surfacing from DC1 is in place — two crons + a human dispatch on
  one reservation is a routine three-contender case, and the loser is silently
  cancelled. Because GCP draws from a separate reservation, **alternating intents
  across clouds** (e.g. training cron on one, inference cron on the other) is the
  safer launch posture. Make "two crons (or cron + dispatch) on one reservation queue
  cleanly, third is reported superseded" the documented validation scenario for the
  broker, and record the chosen cadence in `docs/contributor/uat.md`.
- Conformance + evidence emission already exist (`phase_conformance`,
  `tests/uat/aws/run:175-191`); confirm the inference recipe's criteria
[...]

Issue #1275, opened 2026-06-09
DC2 — Dynamic per-intent cluster acquisition


njhensley:
**Parent:** #1264 · **Labels:** `enhancement`, `area/ci` (auto), `area/tests` (auto,
`tests/**`) · **Size:** M · **Depends on:** DC1

**Goal.** Acquire a cluster from a reservation **through the broker** and shape it
for the requested intent (training vs inference), replacing AWS's hard-fail-on-busy
pre-flight with queue-aware acquisition — and decide what the GCP side does, since
GCP has no pre-flight capacity gate today. DC2 also owns the **two lifecycles** a
cluster can have on a reservation: the **per-run nightly-batch cluster** (provision →
CUJ → teardown, the existing behavior) and the **long-lived daytime deployment** (the
morning-handoff cluster DC8 hands to humans), plus the **enforced cleanup boundary**
between them.

**Scope.**
- **AWS:** convert the existing capacity step (`uat-aws.yaml:122-147`) from a
  race-and-fail check into a **post-lease assertion**: once the broker grants the
  lease, assert the reservation has the desired count (a genuinely exhausted
  reservation still fails; contention no longer does, because the lease was the
  gate).
- **GCP:** there is **nothing to convert** — `uat-gcp.yaml` has no capacity-check
  step. Explicitly decide between (a) **adding** a symmetric GCP-side
  capacity/quota assertion (post-lease, parallel to the AWS one) or (b)
  **documenting** that GCP relies on the GKE actuator failing at provision time.
  Record the decision in `docs/contributor/uat.md`. Do not imply a symmetric edit
  that does not exist.
- Select the test config and recipe criteria by an `intent` input rather than the
  hard-coded `TEST_CONFIG: …/h100-training-config.yaml` (`uat-aws.yaml:55`,
  `uat-gcp.yaml:56`). Introduce `tests/uat/{aws,gcp}/tests/h100-inference-config.yaml`
  as a sibling of the training config (same AICRConfig shape as
  `tests/uat/aws/tests/h100-training-config.yaml`), with
  `spec.validate.recipe.criteria.intent: inference` / `platform: dynamo`. For the
  push target, set `spec.validate.evidence.attestation.out` + `.push` to a stable
  per-intent prefix (e.g. `ghcr.io/nvidia/aicr-uat-eks-h100-inference`, mirroring the
  training config's `:98`); the runner's `inject_push_target` (`tests/uat/aws/run:69-76`)
[...]

PR #1722, opened 2026-07-10
feat(uat): Azure AKS UAT phase runner + intent configs (2/2)

mchmarny:
## Summary

Completes the Azure AKS UAT pipeline started in #1709 (2/2): the phase runner (`tests/uat/azure/run`), the per-intent AICRConfig pair (`h100-{training,inference}-config.yaml`), and the `install` step re-widened to 90m now that the readiness gate can keep the federated az session alive mid-phase.

## Motivation / Context

PR #1709 shipped the account federation, dispatch surface, and pipeline; the burn-in path (provision → connect → destroy) is green on main as of run 29122546423. This PR adds what a full (non-`skip_tests`) run needs.

The runner is a port of the AWS/GCP siblings (byte-identical there apart from comments) with one Azure-only addition: a federated az session cannot self-refresh (5-minute OIDC assertion, ~60–75m access tokens — the AADSTS700024 failure diagnosed during burn-in), and the post-install readiness gate can outlive one token. The gate therefore re-logins in-loop every `AZ_RELOGIN_INTERVAL_SECONDS` (25m): mint a fresh `ACTIONS_ID_TOKEN`, `az login --federated-token`, then eagerly warm the AKS-audience token (the assertion is only 5-minute valid, so the mint cannot be deferred to kubelogin's next cache miss). Transient re-login failures warn and retry rather than killing the gate.

The configs codify the manually validated AKS commands (aicr-test5): system components on the `cpu-worker` pool (the AKS system pool carries `CriticalAddonsOnly`), GPU toleration `nvidia.com/gpu=present:NoSchedule` only. Deliberately no value overrides — the toolkit `RUNTIME_CONFIG_SOURCE` env is baked into `values-aks.yaml` (#1689), and the empty `agentgateway:allowedSourceRanges` receives the bundler's RFC1918 injection (#1373), sufficient because `phase_serve` reaches the endpoint via `kubectl port-forward`, never the LoadBalancer. The NVIDIA-egress override from the manual validation is a human/VPN concern, documented in the inference config.

Fixes: N/A
Related: #1709, #1275, #1276

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT pipeline (`tests/uat/azure/`, `.github/workflows/uat-azure.yaml`)

## Implementation Notes

- `azure-h100` stays `nightly-intents: []` (bring-up, manual dispatch only) until a green manual full run; flipping it is a follow-up one-liner in `infra/uat/reservations.yaml`.
- Runner divergence from siblings is limited to the re-login machinery and three cloud-specific comments — kept diffable on purpose.

## Testing

```bash
bash -n tests/uat/azure/run && shellcheck -x tests/uat/azure/run   # clean
yamllint -c .yamllint.yaml tests/uat/azure/tests/*.yaml .github/workflows/uat-azure.yaml  # clean
actionlint .github/workflows/uat-azure.yaml  # no new findings
```

- Local end-to-end smoke of both configs against a freshly built `aicr`: `recipe --config` resolves (training: 13 components, inference: 16) and `bundle --config` renders; the inference bundle's rendered gateway carries the RFC1918 `loadBalancerSourceRanges` and the gpu-operator values carry `RUNTIME_CONFIG_SOURCE` — matching the validated manual bundles without overrides.
- No Go changes; coverage gate N/A.
- Post-merge: manual `uat-run` dispatch (full run, no `skip_tests`) is the acceptance test before nightly enrollment.

## Risk Assessment

- [x] **Low** — Additive files plus a timeout widening; the burn-in path is unaffected, and a revert restores the 60m cap.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — no Go changes
- [x] I added/updated tests for new functionality (N/A — CI tooling; validated by dispatch)
- [x] I updated docs if user-facing behavior changed (N/A — contributor docs already describe the per-intent config contract)

### DIFF
--- a/tests/uat/azure/run
+++ b/tests/uat/azure/run
@@ -1,0 +1,799 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# UAT runner. Codifies the workflow demonstrated in
+# demos/cuj1-training.md against the AKS cluster provisioned by
+# ../cluster-config.yaml, and is invoked phase-by-phase from
+# .github/workflows/uat-azure.yaml so per-phase failures surface in the
+# GitHub Actions UI.
+#
+# Usage:
+#   AICR_BIN=./aicr RUN_ID=12345 ./run <phase> <test-config.yaml>
+#
+# Phases:
+#   prep         snapshot + recipe + dry-run validate + bundle
+#   install      helmfile apply (deploys gpu-operator, kubeflow/dynamo, ...)
+#   conformance  validate ALL phases (deployment + conformance + performance)
+#                + emit signed evidence bundle
+#   train        submit TrainJob, wait for completion, capture logs
+#                (intent=training CUJ)
+#   serve        deploy a DynamoGraphDeployment, wait for readiness, hit the
+#                OpenAI-compatible endpoint, assert a completion, capture logs
+#                (intent=inference CUJ — the DC3 counterpart of `train`)
+#   verify       aicr evidence verify against the signed bundle
+#   all          run every phase in order (for local reproduction); the CUJ
+#                phase is chosen by the config's recipe intent — `train` for
+#                training, `serve` for inference
+#
+# Required env:
+#   AICR_BIN    Path to the aicr binary
+#   RUN_ID      Per-run correlation tag (`run-<id>`) applied to the evidence
+#               push target (e.g. ${{ github.run_id }} in CI, or `local-$(date +%s)`)
+#
+# Working files are written to $PWD: snapshot.yaml, recipe.yaml, bundle/,
+# dry-run.json, report.json, evidence/, evidence-result.json, train-logs/.
+
+set -euo pipefail
+
+phase="${1:-}"
+config="${2:-}"
+
+if [[ -z "${phase}" || -z "${config}" ]]; then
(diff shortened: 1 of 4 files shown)


njhensley (approved):

tests/uat/azure/run
```diff
@@ -0,0 +1,799 @@
...
+    -H "Authorization: bearer ${ACTIONS_ID_TOKEN_REQUEST_TOKEN}" \
+    "${ACTIONS_ID_TOKEN_REQUEST_URL}&audience=api://AzureADTokenExchange" | jq -r .value)"
+  az login --service-principal \
+    --username "${AZURE_CLIENT_ID}" \
+    --tenant "${AZURE_TENANT_ID}" \
+    --federated-token "${oidc_token}" \
+    --output none
+  az account set --subscription "${AZURE_SUBSCRIPTION_ID}" --output none
+  az account get-access-token --resource "${AKS_SERVER_APP_ID}" --output none
+  echo "federated az session refreshed (ARM + AKS audiences)"
```
yuanchen8911, line 160: `az_federated_relogin` can report success after a failed token refresh (L150-L160, invoked at L360).

Because the helper is called as `if az_federated_relogin; then ...`, Bash does not apply `errexit` inside the function body for failures in that conditional context. The refresh steps at L150-L159 have no explicit failure propagation, so a failed OIDC `curl | jq`, `az login`, `az account set`, or `az account get-access-token` can fall through to the unconditional `echo` at L160, whose 0 exit status becomes the function result.

That makes the caller advance `last_relogin` as if refresh succeeded and skip the warning/retry path, contradicting the comment that the timer only advances on success. During the 90m install gate, a transient OIDC/az failure would be masked: no warning is emitted and the next refresh is deferred for the full relogin interval. If the cached token later expires, `aicr validate` can fail with AADSTS700024 without the diagnostic this branch was intended to provide.

Please explicitly propagate each refresh step failure, for example with `|| return 1` or `if ! ...; then return 1; fi`, and only print `federated az session refreshed` after all mint/login/warm-up commands have succeeded.

mchmarny: Confirmed and fixed in follow-up #1725 — every mint/login/warm-up step now carries an explicit `|| return 1`, the minted token is rejected when empty/`"null"`, and the success echo prints only after all steps pass, so the caller's "timer only advances on success" contract is actually enforced. Good catch: `pipefail` survives the conditional context (so the `curl | jq` pipe was covered) but `errexit` does not, and the trailing echo converted every failure into rc 0.

PR #1724, opened 2026-07-10
feat(bundler): argocd-helm children-only render via includeRootApp

### LINKED ISSUE #1723: argocd-helm: option to render children-only (suppress the aicr-stack root app-of-apps)
## Problem

The `argocd-helm` bundle ships the root app-of-apps as an ordinary chart template (`templates/aicr-stack.yaml`), written unconditionally by [`writeParentApplicationTemplate`](https://github.com/NVIDIA/aicr/blob/8f198eb966f144d8abc3ab8a1a1e769dad08d6f4/pkg/bundler/deployer/argocdhelm/argocdhelm.go#L846).

An external controller (e.g. a CAPI addon provider) that creates **its own** root Argo CD Application pointing at the published OCI bundle therefore renders the child Applications *plus* [`aicr-stack`](https://github.com/NVIDIA/aicr/blob/8f198eb966f144d8abc3ab8a1a1e769dad08d6f4/pkg/bundler/deployer/argocdhelm/argocdhelm.go#L783) — which points back at the same chart. Two root apps now own the identical set of children, and because `aicr-stack` carries `automated: {prune: true, selfHeal: true}`, the two parents fight continuously (SharedResourceWarning, sync thrash) rather than surfacing a one-time warning.

There is no workaround today short of `helm template` + applying the children directly, which defeats the GitOps loop. Falling back to the plain `argocd` deployer is not an option for this integration because install-time `--dynamic` values are `argocd-helm`-only (the `argocd` deployer bakes values at bundle time).

## Proposed solution

A **render-time values toggle**, not a bundle-time flag: gate `templates/aicr-stack.yaml` behind `deployer.includeRootApp` (default `true`).

- Bring-your-own-root consumers set `deployer: {includeRootApp: false}` in their root Application's `helm.valuesObject` and get children-only.
- Standalone `helm install` keeps today's behavior unchanged.
- The same published OCI artifact serves both flows — no re-bundle, preserving build-once-deploy-many (#516).
- Fits the existing `deployer.*` values namespace (`namePrefix`, `destinationServer`, `project` — #1625).

## Alternatives considered

- **Bundle-time flag to omit the parent template** — forces two published artifacts for the two consumption modes.
- **Separate chart for the child Applications that the app-of-apps consumes** — doubles the publish surface and breaks the single-artifact model.

## Acceptance criteria

- [ ] `deployer.includeRootApp=false` renders zero `aicr-stack` Application; children unchanged
- [ ] Default (unset / `true`) output is byte-identical to today
- [ ] Root `values.yaml` documents the key (visible via `helm show values`)
- [ ] Docs updated (`docs/user/bundling.md` argocd-helm section)

Reported by the CAPI addon provider for AICR integration.

mchmarny:
## Summary

Adds an install-time boolean, `deployer.includeRootApp` (default `true`), that gates the argocd-helm bundle's parent app-of-apps template so consumers that bring their own root Argo CD Application can render children-only from the same published chart.

## Motivation / Context

An externally managed root Application pointing at the published OCI bundle — e.g. one created by a CAPI addon provider controller — previously rendered a second parent (`aicr-stack`) that owned the same child Applications; the two roots fought indefinitely through the parent's `automated: {prune: true, selfHeal: true}` sync policy. The plain `argocd` deployer is not a fallback for such consumers because install-time `--dynamic` values are argocd-helm-only.

Fixes: #1723
Related: N/A

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- **Render-time toggle, not a bundle-time flag or second chart:** the same published OCI artifact serves both the standalone `helm install` flow (root app included, unchanged) and the bring-your-own-root flow (`deployer: {includeRootApp: false}` in the external root's `helm.valuesObject`, or `--set deployer.includeRootApp=false`).
- The template gate uses sprig `dig` rather than `default` — `default` coerces an explicit `false` back to `true`. `{{- if ... -}}` trim markers keep the enabled-path render byte-identical to today.
- Baked `includeRootApp: true` into the bundle's root `values.yaml` so `helm show values` documents it, and added the key to `values.schema.json` as a **boolean** — `additionalProperties: false` would otherwise reject the `--set`, and a string value (`--set-string`) fails the schema loudly instead of rendering truthy.
- Install-time only, deliberately: not added to the bundle-time `--set deployer:` allowlist (`pkg/bundler/config`), which continues to reject it fail-closed.
- `cascadeDelete`'s injected finalizer anchor in `parentAppTemplate` is unaffected (golden-covered).

## Testing

```bash
make qualify        # PASS (test -race, lint, e2e 23/23, scan, licenses)
```

- New `TestHelmTemplate_IncludeRootAppToggle` (live `helm template`): default renders parent+children; `--set deployer.includeRootApp=false` renders children-only; `--set-string` value fails the values schema.
- Extended `TestGenerate_DeployerDefaults_NoOptions` to pin the baked default, the template gate, and the schema property.
- Golden bundles regenerated (`-update`): `values.yaml`, `values.schema.json`, `templates/aicr-stack.yaml`.
- Coverage: `pkg/bundler/deployer/argocdhelm` 82.4% → 82.3% (-0.1%) vs `origin/main`; no new exported funcs.

## Risk Assessment

**Rollout notes:** Backwards compatible — default (`true`) output renders byte-identical to today; the key only takes effect when explicitly set to `false`. Older bundles (without the gate) ignore the value harmlessly: the schema in *those* bundles rejects the unknown key at install time, surfacing the version mismatch loudly.

## Checklist


yuanchen8911 (approved):
/lgtm

PR #1725, opened 2026-07-10
fix(uat): propagate az_federated_relogin step failures

mchmarny:
## Summary

Follow-up to #1722, addressing @yuanchen8911's review comment: `az_federated_relogin` could report success after a failed refresh. Every mint/login/warm-up step now propagates failure explicitly, and the success message prints only after all steps pass.

## Motivation / Context

The helper is invoked as `if az_federated_relogin; then …`, and Bash suspends `errexit` inside a function body called in a conditional context. A failed OIDC `curl | jq`, `az login`, `az account set`, or `az account get-access-token` therefore fell through to the unconditional success `echo`, whose 0 exit became the function result — the caller advanced `last_relogin` as if the refresh succeeded, skipped the warning/retry path, and a later `AADSTS700024` would surface without the diagnostic this mechanism exists to provide.

Now each step carries `|| return 1`, and the minted token is additionally rejected when empty or the JSON-null string (`jq -r .value` on an unexpected response shape). `pipefail` (set script-wide, unaffected by the conditional context) already propagates a `curl` failure through the `jq` pipe.

Fixes: N/A
Related: #1722 (review follow-up)

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT pipeline (`tests/uat/azure/run`)

## Implementation Notes

Single-function change; the caller's warn-and-retry contract ("the timer only advances on success") is now actually enforced rather than merely documented.

## Testing

```bash
bash -n tests/uat/azure/run     # clean
shellcheck -x tests/uat/azure/run  # clean
```

- No Go changes; coverage gate N/A.
- Behavior exercised end-to-end by the next full `uat-run` dispatch (the gate logs each refresh).

## Risk Assessment

- [x] **Low** — Strictly tightens an error path in a CI helper; the success path is byte-identical.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — no Go changes
- [x] I added/updated tests for new functionality (N/A — CI shell helper)
- [x] I updated docs if user-facing behavior changed (N/A)


yuanchen8911 (approved):
Verified the fix: each mint/login/warm-up step now propagates failure via `|| return 1`, and the success message prints only after all steps pass, so the caller's warn-and-retry branch actually runs and `last_relogin` no longer advances on a failed refresh. Reproduced the `set -euo pipefail` + `if az_federated_relogin` conditional-context semantics locally to confirm. The empty/`null` token guard is a good addition. Thanks for the quick turnaround.

PR #1729, opened 2026-07-11
feat(bundle): KMS-backed signing via HashiCorp Vault (hashivault://)

### LINKED ISSUE #1577: feat(bundle): KMS-backed signing via HashiCorp Vault (hashivault://)
## Problem

AICR's KMS-backed signing and verification (#407, #1152) supports the
cosign-style cloud providers — `awskms://`, `gcpkms://`, and `azurekms://` —
but not HashiCorp Vault (`hashivault://`). Vault is a common enterprise key
store, and users who standardize on it currently have no first-class signing
path.

The provider is not yet wired in because its client libraries carry a license
(MPL-2.0) that needs sign-off under our dependency policy before we can vendor
it. That review is in progress.

## Proposal

Add `hashivault://` to the supported KMS URI schemes once the dependency is
cleared, so Vault-held keys work everywhere the other KMS providers already do:

- `aicr bundle --signing-key hashivault://...`
- `aicr verify --key hashivault://...`

Implementation is expected to be small and symmetric with the existing
providers: blank-import `github.com/sigstore/sigstore/pkg/signature/kms/hashivault`
in `pkg/bundler/attestation/kmsidentity.go` and add the scheme to the shared
`kmsURISchemes` set (single source of truth for signing + verifying).

## Blocked on

Dependency license approval for the Vault KMS client (MPL-2.0). Tracking here so
the work is queued and ready to land as soon as that clears; approval is
expected.

If approval is ultimately not granted, close this issue and open a follow-up to
track an alternative that avoids the dependency (e.g. the out-of-tree cosign KMS
plugin protocol, or an optional build tag).

## Success criteria

- A bundle signed with `aicr bundle --signing-key hashivault://<key>` verifies
  with `aicr verify --key hashivault://<key>`.
- The inline rationale in `pkg/bundler/attestation/kmsidentity.go`, `doc.go`, and
  `docs/user/cli-reference.md` is updated to reflect supported status.
- `make qualify` passes.

lockwobr:
## Summary

Add HashiCorp Vault (`hashivault://`) as a supported KMS backend for `aicr bundle --signing-key` (signing) and `aicr verify --key` (verification), alongside the existing `awskms://` / `gcpkms://` / `azurekms://` providers.

## Motivation / Context

Vault KMS was previously excluded on license grounds: its client libraries are MPL-2.0, which the project's license policy disallowed. That exclusion has now been formally approved and lifted, so this change wires up the provider and flips the code and docs from "intentionally unsupported" to "supported."

Fixes: #1577
Related: #1149 (Closed supply chain epic), #407 (KMS-backed signing)

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: license-check policy (Makefile), e2e suite + CI workflow

## Implementation Notes

- **Core enablement is minimal and symmetric.** The set of KMS schemes lives in one place, `kmsURISchemes` in `pkg/bundler/attestation/kmsidentity.go`, shared by both the signing side (`NewKMSIdentity` / `resolveKMSPublicKey`) and the verifying side (`NewKeyVerificationIdentity`). Adding `hashivault://` there plus blank-importing the sigstore hashivault provider is the whole enablement; sign/verify parity is structural. Every other spot that enumerated the scheme trio (godoc, CLI usage strings, docs) was updated to match.
- **License policy — fail-closed.** The approved MPL-2.0 packages are exempted via explicit per-package `--ignore` entries in the `Makefile` `license-check` target, not by widening `--allowed_licenses`. This keeps the gate fail-closed: an unrelated future MPL-2.0 dependency still fails for review. The 8 ignored import paths are exactly the ones `go-licenses check` flags for the Vault client subtree. The `DO NOT CHANGE` directive comment was rewritten to sanction only these named, approved exceptions.
- **e2e coverage.** New chainsaw suite `tests/chainsaw/signing/bundle-attestation-vault/` (sign → verify-by-URI → verify-by-PEM → trust levels → tamper → wrong-key) plus a dedicated `vault-kms-e2e.yaml` workflow, mirroring the existing MiniStack KMS pattern but against OpenBAO (the Apache-2.0 fork of Vault) in dev mode over HTTP. The OpenBAO image is pinned in `.settings.yaml` and surfaced via the `load-versions` action. As with MiniStack, the full `--attest` round-trip runs only in CI (it needs a CI-attested binary); the local `run.sh` falls back to smoke mode.
- **Dependencies.** `go mod tidy` + `go mod vendor` pulled the sigstore hashivault provider (`v1.10.8`) and the Vault client subtree; `THIRD_PARTY_NOTICES.md` was regenerated via `make notices`.

## Testing

```bash
make qualify            # green: tests (-race), lint (0 issues), e2e, scan, license-check
make license-check      # PASS with the new per-package ignores
```

- Added `TestResolveKMSPublicKey_HashivaultProviderRegistered` — pins the blank import: a `hashivault://` URI must route to a registered provider (classified `ErrCodeUnavailable` with no `VAULT_ADDR`), not the unknown-scheme `ErrCodeInvalidRequest`. Also extended `TestIsKMSURI` / `TestNormalizeURIScheme` with hashivault cases.
- Exercised the e2e plumbing against a real **OpenBAO 2.4.1** container (Transit enable, `ecdsa-p256` key, PEM export via the HTTP API) and ran the local `run.sh` end to end — `Vault KMS E2E: PASSED`.

## Risk Assessment

**Rollout notes:** Additive. The new scheme is opt-in per KMS URI; the existing `awskms://` / `gcpkms://` / `azurekms://` paths are unchanged. The license-check change only adds narrowly-scoped, approved exceptions and preserves the fail-closed posture. No migration required.

## Checklist


tests/chainsaw/signing/bundle-attestation-vault/run.sh
```diff
@@ -0,0 +1,321 @@
...
+
+# =============================================================================
+# Configuration
+# =============================================================================
+
+OPENBAO_CONTAINER_NAME="aicr-vault-kms-e2e-openbao"
+OPENBAO_PORT="${OPENBAO_PORT:-8200}"
+# Image from .settings.yaml (testing_tools.openbao_image; the CI source of
+# truth), with a pinned fallback if yq or the key is unavailable. Never :latest.
+OPENBAO_IMAGE="${OPENBAO_IMAGE:-$(yq '.testing_tools.openbao_image // "openbao/openbao:2.4.1"' "${ROOT}/.settings.yaml" 2>/dev/null || echo 'openbao/openbao:2.4.1')}"
```
mchmarny: nit: the hardcoded `openbao/openbao:2.4.1` fallback will silently drift — Renovate bumps `.settings.yaml` but not this string, so a machine without `yq` (or a renamed key) quietly tests against a stale image. Since `.settings.yaml` is the single source of truth, prefer failing fast when the lookup comes back empty instead of falling back to a duplicate pin.

tests/chainsaw/signing/bundle-attestation-vault/run.sh
```diff
@@ -0,0 +1,321 @@
...
+    export AICR_BIN
+    # Detect the binary attestation (FindBinaryAttestation convention:
+    # <binary>-attestation.sigstore.json) so the --attest / verified-level steps
+    # run only when the binary actually carries one.
+    if [ -f "${AICR_BIN}-attestation.sigstore.json" ]; then
+      export AICR_ATTESTED=true
+      msg "Binary attestation found; --attest steps will run"
+    else
+      export AICR_ATTESTED=false
+      warn "No attestation next to ${AICR_BIN}; --attest steps will fail (provide a CI-attested binary)"
```
mchmarny, line 241: nit: when the attestation sidecar is found but `AICR_IDENTITY_REGEXP` is unset, the full chainsaw suite runs with an empty regexp, which falls back to the production on-tag identity default and rejects the `build-attested.yaml`-attested binary with a confusing sign-time failure. The README documents the pairing, but a `warn` (or fail-fast) here when `AICR_ATTESTED=true` and `AICR_IDENTITY_REGEXP` is empty would save the debugging round-trip.

mchmarny (commented):
Adds `hashivault://` as a fourth KMS scheme for `bundle --signing-key` / `verify --key`, with the enablement kept to the single shared scheme list plus a blank provider import — sign/verify parity is structural, as advertised. Verified the load-bearing claims: the `BAO_ADDR`/`BAO_TOKEN` and `TRANSIT_SECRET_ENGINE_PATH` doc statements match the vendored sigstore hashivault client exactly; the 8 license-check ignores are precisely the MPL-2.0 Vault subtree (the other new deps are MIT and correctly remain under allowlist enforcement); no scheme-trio enumeration was missed anywhere in the repo; and the new OpenBAO e2e mirrors the MiniStack suite's labels, gating, and fail-closed negative checks (explicit `rc` capture on wrong-key, value-asserting greps on tamper).

All 165 checks are green on head `5fca139`, and the three earlier CodeRabbit findings are addressed. Two nits inline on `run.sh` (a duplicated image pin that can drift from `.settings.yaml`, and a missing warning for the attested-binary-without-identity-regexp combination) — neither blocks merge. Only real gate left is the `needs-rebase` label: base is behind `main`, so rebase + re-run before merging.

lockwobr:
Thanks for the review. Rebased onto current `main` (the branch is now 1 commit ahead, 0 behind — the `needs-rebase` was resolved by the prior push and the label should clear on this push) and addressed all four nits:

**@mchmarny (run.sh):**
- **Image-pin drift** — `run.sh` no longer duplicates a literal tag; `OPENBAO_IMAGE` is read solely from `.settings.yaml` (`testing_tools.openbao_image`) and errors out if that read fails, so `.settings.yaml` stays the single source of truth shared with CI via `load-versions`.
- **Missing attested-without-regexp warning** — `build_binary` now warns when a binary attestation is present but `AICR_IDENTITY_REGEXP` is unset (the `--attest` steps would otherwise fall back to the on-tag default identity and fail confusingly).

**CodeRabbit:**
- **Eager `has_tools`** — now checks only `docker curl python3 yq` upfront; `goreleaser` is checked lazily in `build_binary` and `chainsaw` in `run_chainsaw`, so the pre-built-binary and smoke paths aren't blocked. Dropped the unused `make`.
- **Hardcoded wrong-key name** — the wrong-key probe name is now derived as `${VAULT_KMS_KEY:-aicr}-wrong`, so it can never alias the signing key.

On the earlier `go.mod` comment (regenerate vendor without touching `go.sum`): that was a false positive. The vendor tree *is* regenerated and committed, and `go.sum` is correctly updated to match — the `verify-licenses` and build checks both passed, which can't happen with an inconsistent vendor tree. CodeRabbit can't see `vendor/**` or `go.sum` (both path-filtered), so it inferred a drift that isn't there.

lockwobr:
Addressed the two follow-ups from the latest review — both were staleness introduced by the earlier `has_tools` change:

- **`run.sh` PREREQUISITES comment** — now lists `yq` as an always-required tool, drops `make` (never invoked), and splits `goreleaser`/`chainsaw` out as lazily-checked (in `build_binary`/`run_chainsaw`).
- **README Prerequisites table** — added `yq`, and a "When" column marking `goreleaser`/`chainsaw` as conditional.

Note on the prior CI run: the wave of failures (128 jobs, including `tests / Test`, the whole Tier-1 matrix, and `startup_failure` on **`Private Sigstore E2E`** and **`KMS MiniStack E2E`** — both unchanged by this PR) was a GitHub Actions platform incident, not a regression here. The amend that triggered it changed only `run.sh` and a `requires=openbao` chainsaw file; the Go code and `main` base were byte-identical to the earlier all-green run. This push starts a fresh run.

lockwobr:
Addressed the latest review (its inline comments failed to post, so working from the summary):

- **`cli-reference.md` / `artifact-verification.md` "cloud KMS" wording** — reworded to "KMS" across the `--signing-key` descriptions, since `hashivault://` (Vault/OpenBAO) can be self-hosted, not just cloud.
- **Spurious `github.com` entry in `THIRD_PARTY_NOTICES.md`** — good catch, and it was introduced by this PR (not on `main`). Root-caused it: `go-licenses csv` deterministically emits a bare-domain `github.com` row (no path, attributed to the repo's own LICENSE) once the Vault subtree is vendored. Fixed at the source in `tools/generate-notices` — dropped any row whose package field has no `/` (a bare domain is never a valid Go module path) — and regenerated the file. The spurious index row and its `### github.com` section are gone; all 165 real `github.com/*` deps remain.
- **`Makefile` license-check line-continuation** — this one is a false positive (the inline comment didn't post, so the structure was likely misjudged): the last `--ignore=.../hcl` line ends with `\` and `--ignore=$STDLIB_IGNORE` is the final argument, so the command does run through `STDLIB_IGNORE`. `make license-check` passes.
- **markdownlint noise on `THIRD_PARTY_NOTICES.md`** — agreed it's expected for a generated file embedding verbatim third-party license text; not fixing per the review's own suggestion.

tools/generate-notices
```diff
@@ -80,8 +80,13 @@ go-licenses save "${PACKAGES[@]}" \
...
 INDEX_FILE="$(mktemp)"
 trap 'rm -f "${INDEX_FILE}"' EXIT
+# go-licenses can emit a spurious bare-domain row (e.g. "github.com" with no
+# path, attributed to the local repo's LICENSE) when it cannot map a package to
+# a specific module. A bare domain is never a valid Go module path, so drop any
+# row whose package field contains no "/".
 go-licenses csv "${PACKAGES[@]}" \
     --ignore="${IGNORE}" \
+    | awk -F, '$1 ~ /\//' \
     | sort -t, -k1,1 -u \
```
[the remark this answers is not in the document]
lockwobr: https://github.com/NVIDIA/aicr/pull/1753 is going to cover this.

lockwobr:
Triaged the latest review (4 items):

- **`doc.go:22` "cloud-KMS-backed" → "KMS-backed"** — done (consistent with the CLI/verification docs; covers self-hosted Vault/OpenBAO).
- **`run.sh` `build_binary` — `AICR_BIN` set-but-not-executable** — done. It now fails fast with a clear error instead of silently falling through to a snapshot build (which would ignore the caller's explicit intent).
- **`vault-kms-e2e.yaml` SLSA-predicate ordering** — this is a false positive. The workflow already runs `Generate SLSA predicate` (step at ~L137) before `Build aicr binary` (~L192), and the last green `Vault KMS E2E` run confirms the binary was attested (`Binary attestation found: …/aicr-attestation.sigstore.json`) and `verify-min-trust-verified` executed (not skipped).
- **`tools/generate-notices` bare-domain filter** — deferring to #1753 (`feat/notices-freshness`), which owns `THIRD_PARTY_NOTICES.md` + the generator comprehensively (cross-platform union + a `notices-freshness` merge-gate) and supersedes the single-platform approach. To avoid conflicting with and duplicating that work, I've **reverted the notices regeneration and the generator change out of this PR** entirely — this PR no longer touches `THIRD_PARTY_NOTICES.md` or `tools/generate-notices`. #1753 will pick up the Vault subtree when it regenerates.

pkg/bundler/attestation/kmsidentity.go
```diff
@@ -46,14 +48,12 @@ const ctxKeySigningKey = "signingKey"
 // supports. It is the single source of truth shared by the signing side
 // (NewKMSIdentity / resolveKMSPublicKey) and the verifying side
 // (NewKeyVerificationIdentity), so the two cannot drift on what counts as a
-// "KMS URI". The set mirrors the blank-imported provider packages above;
-// HashiCorp Vault (hashivault://) is intentionally omitted (MPL-2.0 license,
-// see issue #407).
-var kmsURISchemes = []string{"awskms://", "gcpkms://", "azurekms://"}
+// "KMS URI". The set mirrors the blank-imported provider packages above.
+var kmsURISchemes = []string{"awskms://", "gcpkms://", "azurekms://", "hashivault://"}
```
mchmarny, line 52: The vendored sigstore hashivault provider registers two schemes: `ReferenceScheme = "hashivault://"` and `AlternativeScheme = "openbao://"` (vendor/github.com/sigstore/sigstore/pkg/signature/kms/hashivault/signer.go). Because the signing side has no scheme gate (`--signing-key` flows straight to `NewKMSAttester` -> `kms.Get`), `aicr bundle --attest --signing-key openbao://<key>` will sign successfully — but `aicr verify --key openbao://<key>` hits `isKMSURI() == false` here and falls through to the local-PEM path, failing with a confusing file-not-found error. That breaks the sign/verify parity this list exists to guarantee, and since this PR's own e2e runs against OpenBAO, users are likely to try the alias. Suggest either adding `"openbao://"` to `kmsURISchemes` (it is the same provider) or rejecting it explicitly at sign time; either way, document the outcome. Failure direction is closed, so this is a UX/consistency issue, not a security one.

pkg/bundler/attestation/kmspublickey_test.go
```diff
@@ -32,6 +32,33 @@ func TestResolveKMSPublicKey_UnknownScheme(t *testing.T) {
...
+// call, which resolveKMSPublicKey classifies as ErrCodeUnavailable — distinct
+// from the ErrCodeInvalidRequest returned for an unregistered scheme (see
+// TestResolveKMSPublicKey_UnknownScheme). If the blank import were dropped,
+// kms.Get would surface ProviderNotFoundError and the code would flip to
+// ErrCodeInvalidRequest, failing this test.
+func TestResolveKMSPublicKey_HashivaultProviderRegistered(t *testing.T) {
+	t.Setenv("VAULT_ADDR", "")
+	t.Setenv("BAO_ADDR", "")
+
+	_, _, err := resolveKMSPublicKey(context.Background(), "hashivault://transit/keys/k")
```
mchmarny, line 48: This test passes via a different mechanism than its comment describes. The provider's reference regex is `^(?:hashivault|openbao)://(?P<path>\w(([\w-.]+)?\w)?)$` — no slashes allowed — so `hashivault://transit/keys/k` is rejected by `ValidReference()` at the top of `newHashivaultClient`, *before* the VAULT_ADDR/BAO_ADDR check ever runs. The two `t.Setenv` lines are therefore inert, and the "provider fails to initialize with VAULT_ADDR unset" path documented in the comment never executes (the assertion still holds because an invalid-reference error is also non-ProviderNotFound -> ErrCodeUnavailable, so the blank import is still pinned). Use a valid single-segment reference, e.g. `hashivault://k`, so the test actually exercises the missing-address init failure it claims to pin — otherwise a future change that classifies invalid references as ErrCodeInvalidRequest (arguably the correct code for a malformed URI) will break this test for a confusing reason.

nit: the `TestIsKMSURI` hashivault cases below (`hashivault://transit/keys/k`) use the same provider-invalid form; fine for a prefix check, but a valid example would be less misleading.

mchmarny (approved):
Clean, well-scoped enablement: the core change is two lines (`hashivault://` added to the single-source-of-truth `kmsURISchemes` plus the blank provider import), reusing the existing timeout-bounded `resolveKMSPublicKey`, error-classification, and scheme-normalization infrastructure, so sign/verify parity and error-code semantics carry over structurally. The license-check change keeps the gate fail-closed via per-package ignores rather than widening the allowlist, vendored code matches the go.mod additions (sigstore hashivault v1.10.8 + hashicorp/vault/api subtree, all pinned), and the OpenBAO e2e mirrors the established MiniStack pattern with proper fail-closed negative tests. CI is fully green, including the new Vault KMS E2E lane.

Two non-blocking medium findings inline: the provider also registers an `openbao://` alias that signs but cannot verify, and the new unit test passes via an unintended failure path. Worth a follow-up, but nothing blocks merge. Suggest a rebase before merging given the vendor churn (the `needs-rebase` label is stale — GitHub reports the branch as mergeable).

PR #1737, opened 2026-07-13
feat(slinky-topograph): topograph component with slinky engine

ArangoGutierrez:
## Summary

Adds the `slinky-topograph` component (NVIDIA Topograph, `slinky` engine) to the H100 GKE and Kind training-slurm recipes, wiring the generated `topology.conf` end-to-end into `slurmctld` via the chart-owned `slinky-slurm-config-extra` ConfigMap.

## Motivation / Context

Supersedes #1554, authored by @ravisoundar. This lands his work with his approval while he is on PTO, so review and merge can proceed — his original commit is preserved as the first commit here, unmodified, with his authorship intact; he should pick up review credit on this PR when he's back. The second commit here addresses the 5 open review threads that were blocking #1554 (from @mchmarny, @yuanchen8911, and @ArangoGutierrez) — see the thread-to-fix table under Implementation Notes.

Fixes: N/A
Related: #1496

## Type of Change

## Component(s) Affected

## Implementation Notes

**Deploy-order flip.** `slinky-topograph`'s `slinky` engine create-or-updates the `slinky-slurm-config-extra` ConfigMap, patching only the `topology.conf` key. The `slinky-slurm` chart must create that ConfigMap first, or the Helm install for `slinky-slurm` fails on ownership metadata once Topograph has already claimed it. So `slinky-topograph` now declares `slinky-slurm` in `dependencyRefs` (previously the reverse), and `pkg/recipe/deployment_order_guard_test.go` guards the new ordering on both slurm leaves (`h100-gke-cos-training-slurm`, `h100-kind-training-slurm`). Live-verified on a Kind cluster: `managedFields` on the ConfigMap show `helm(Apply)` followed by `topograph(Update)`, and `cgroup.conf`/`gres.conf` (chart-owned) survive Topograph's key-level patch of `topology.conf`.

**Open review-thread → fix mapping** (5 threads, all previously blocking #1554):

| Thread | Reviewer(s) | Fix |
|---|---|---|
| Commit history not rebased/squashed, signing unclear | @mchmarny | Rebuilt as exactly 2 signed commits on top of current upstream/main: Ravi's original feature commit (author preserved, `Signed-off-by: Ravi Shankar` + GPG `-S`), followed by one squashed fix commit addressing every other thread below (also `-S -s`, second `Signed-off-by` trailer added by the committer). |
| Health check's `script` step violates the read-only allowlist (`TestValidateTestReadOnly_RegistryContent` fails: only `assert`/`error` permitted) | @yuanchen8911, @ArangoGutierrez | Rewrote `recipes/checks/slinky-topograph/health-check.yaml` to assert the DaemonSet's state directly instead of shelling out — no `script` op remains. |
| Exact-content `topology.conf` assertion fails deterministically on the GKE leaf (assertFile hydrates onto every overlay, but GKE renders real cluster topology, not the test fixture) | @yuanchen8911, @ArangoGutierrez | Registry-level check now asserts only "key present and non-empty" (provider-agnostic, passes on every leaf). The Kind leaf carries its own inline `healthCheckAsserts` with the exact 3-switch fixture content, scoped to that leaf only. |
| Generated ConfigMap is never mounted into `slurmctld` (`topologyConfigmapName` pointed at a CM the slurm chart doesn't reference) | @yuanchen8911, @ArangoGutierrez | Repointed `topologyConfigmapName` at the chart-owned `slinky-slurm-config-extra`; `slinky-slurm`'s overlay now seeds a `topology.conf` placeholder via `configFiles` so the chart creates and mounts the key (via `configFileRefs`) before Topograph's first sync. |
| `TopologyPlugin` is never set, so `slurmctld` ignores `topology.conf` even once mounted | @yuanchen8911, @ArangoGutierrez | Added `controller.extraConfMap.TopologyPlugin: topology/tree` to both slurm overlays. |

## Testing

```bash
```

- `make test` (race detector): **PASS**, rc=0. `pkg/recipe` package: 87.5% of statements.
- `make qualify` (test + lint + e2e + scan + repo checks): **PASS**, rc=0.
  - e2e (`chainsaw/cli-recipe-overlays`): 23 passed / 0 failed / 0 skipped.
  - vulnerability scan: one pre-existing advisory (`golang.org/x/crypto v0.54.0`, `GO-2026-5932`, severity Unknown) — non-blocking, untouched by this branch.
- `pkg/recipe` coverage delta vs the `079e1734` fork-point baseline: **87.4% → 87.4% (0.0%)** — the branch adds only a test file (`pkg/recipe/deployment_order_guard_test.go`); no new exported func/method to gate.
- `golangci-lint run -c .golangci.yaml ./pkg/recipe/...`: 0 issues.
- KWOK scheduling simulation (`make kwok-e2e RECIPE=h100-gke-cos-training-slurm`): **PASS**, 48/48 pods scheduled; `slinky-topograph` present in the 17-component GKE leaf.
- Live Kind cluster (k8s v1.36.1): deployed the resolved slurm-stack subset — `cert-manager`, `slinky-slurm-operator-crds`, `slinky-slurm-operator`, `slinky-slurm`, `slinky-topograph` (the remaining 11 non-slurm components in the recipe need GPU hardware unavailable on this laptop). Verified on the live cluster:
  - `slinky-slurm-controller-0` **2/2 Running, restarts=0** — no crashloop against the seeded `topology.conf`.
  - `scontrol show config` on `slurmctld` reports **`TopologyPlugin = topology/tree`**.
  - `slinky-slurm-config-extra` holds all three chart-owned keys (`cgroup.conf`, `gres.conf`, `topology.conf`); Topograph key-merges only `topology.conf` (confirmed via `managedFields` = `helm(Apply)` + `topograph(Update)`).
  - The exact 3-switch fixture is present both in the ConfigMap and mounted at `/etc/slurm/topology.conf` on `slurmctld`.
  - The Kind leaf's inline chainsaw health check (`slinky-topograph-health-check-kind`): **PASS**, 1/0/0.
  - **Note:** `make validate-local` does **not** deploy the recipe's components — it loads validator images and runs `aicr validate --phase deployment`, which deploys validator Jobs (chainsaw health-check runners) against the active kube-context; it does not `helm install` the recipe's charts. To exercise the real end-to-end path, the slurm-stack subset above was deployed directly via `helm install`, then `aicr validate --phase deployment` was run against the live cluster using the **published** validator image (`ghcr.io/nvidia/aicr-validators/deployment:latest`) — local `make image-validators` build is blocked by a pre-existing `k8s-launch-kit` vendoring issue unrelated to this branch (base `079e1734` has the identical import; `GOFLAGS=-mod=vendor go build ./validators/deployment` succeeds locally, and `:latest` maps to `PullAlways` so the Job ran the current published machinery). Through that real path, all 5 deployed slurm-stack components — including `slinky-topograph` — reported `health check passed`; the 11 undeployed non-slurm components reported the expected failures (no GPU hardware to bring them up).
- **Post-rebase re-verification:** upstream/main advanced from `079e17345` to `bc3ea1709` (4 commits, incl. a global dep upgrade) during this task. Rebasing produced a genuine content conflict in the generated `docs/user/container-images.md` (BOM), resolved by regenerating with `make bom-docs` rather than hand-merging. Because the rebase touched a conflicting file, the full `make qualify` gate above was re-run post-rebase (not just `make test`) and passed, rc=0.

## Risk Assessment

- [x] **Medium** — new optional component, not deployed by default (opted into only by the two slurm leaf overlays via explicit `componentRef`s), but ships a cluster-wide RBAC grant and one review-flagged validation gap (see below).

**Rollout notes:**
- `slinky-topograph` is off by default; it is scheduled only by an explicit `componentRef` on the two slurm leaf overlays (`dependencyRefs` only orders it). Existing recipes/overlays that don't reference it are unaffected.
- **Upstream chart RBAC:** Topograph chart v0.5.0 grants a cluster-wide (unscoped `ClusterRole`) ConfigMap create/get/list/update when `engine: slinky`, needed so the `topograph` namespace's controller can patch the ConfigMap living in the `slurm` namespace. This is the upstream chart's shape, disclosed here unchanged — not introduced or narrowed by this PR.
- **Seeded-check gap — resolved per review:** the shared/registry health check originally asserted only that `topology.conf` is present and non-empty, which the seeded placeholder satisfies from first boot. Per @mchmarny's review, the assert is now strengthened to `(contains("topology.conf", 'SwitchName')): true` — an unconfigured Topograph (e.g. GKE with no Workload Identity binding) now fails closed at the 5m assert timeout instead of passing vacuously, while a healthy deployment passes once the first sync lands (Kind synced in ~2.5m). Verified both directions against a scratch cluster: synced content → PASS; seed-only content → FAIL (timeout). The Kind leaf's inline variant still asserts exact fixture content.

## Checklist


docs/integrator/recipe-development.md
```diff
@@ -125,6 +125,67 @@ For example, `--platform slurm` leaves inline three `componentRefs`:
 
 This is the same shape `dynamo-platform` uses across the `*-inference-dynamo` leaves. See `recipes/overlays/h100-eks-ubuntu-training-slurm.yaml` for the full example.
 
+### Slinky Slurm Inline Components
```
mchmarny, line 162: This new `### Slinky Slurm Inline Components` heading lands mid-section inside `### Multi-Level Inheritance`, before the existing Talos-mixin paragraph and the **Cross-cutting overlays with wildcard criteria** content — both now render nested under the Slinky heading even though they're unrelated to it. Move the new section down so it sits after the cross-cutting-overlay content (just before `### Inference performance constraints`), keeping Multi-Level Inheritance intact. The two inbound links (`demos/cuj1-slinky-slurm.md`, `component-catalog.md`) use the anchor, not position, so they're unaffected by the move.

recipes/checks/slinky-topograph/health-check.yaml
```diff
@@ -0,0 +1,166 @@
...
+      try:
+        - assert:
+            resource:
+              apiVersion: v1
+              kind: ConfigMap
+              metadata:
+                name: slinky-slurm-config-extra
+                namespace: slurm
+              data:
+                (length("topology.conf") > `0`): true
```
mchmarny: Answering the seeded-check question from the PR body: yes, strengthen this to fail closed. Because `slinky-slurm` seeds `topology.conf` non-empty from first boot, an unconfigured GKE deployment (no Workload Identity binding) passes this check while Topograph never completes a sync — exactly the "presence check that passes on an ambiguous condition" anti-pattern in AGENTS.md.

Since the component pins `plugin: topology/tree`, `(contains("topology.conf", 'SwitchName')): true` works, and chainsaw's 5m assert retry gives Topograph time to finish its first sync on a healthy cluster. The alternative — asserting the content no longer `starts_with` the seed marker — avoids coupling to the tree plugin but couples the registry check to overlay seed text instead; `contains SwitchName` is the better trade. Fine as a fast-follow if you'd rather not re-run live verification in this PR.

demos/cuj1-slinky-slurm.md
```diff
@@ -266,6 +266,16 @@ helm uninstall slinky-slurm-operator-crds -n slinky
 kubectl delete ns slurm slinky --ignore-not-found
 ```
 
+Full Slurm stack with topograph:
+
+```shell
+helm uninstall slinky-slurm -n slurm
+helm uninstall slinky-topograph -n topograph
```
mchmarny, line 272: nit: uninstall `slinky-topograph` before `slinky-slurm`. If node-observer triggers a sync in the window after the slurm chart is gone, Topograph's create-or-update recreates `slinky-slurm-config-extra` as an orphan. The trailing `kubectl delete ns` covers it, but swapping the first two lines removes the race.

mchmarny (commented):
Solid landing of #1554 — the deploy-order flip with the seeded `topology.conf` is the right fix for the ConfigMap ownership problem, and the `managedFields` live verification plus the deployment-order guard test give good confidence. Two mediums: the new docs section in recipe-development.md swallows the unrelated Talos/cross-cutting content (heading placement), and on the named question — strengthen the shared ConfigMap assert to fail closed on an unconfigured provider (inline comment has a concrete suggestion; fast-follow acceptable). One nit on demo uninstall order. Nothing blocking otherwise; happy to approve once the docs placement is fixed and we settle the seeded-check question.

docs/integrator/recipe-development.md
ArangoGutierrez: Moved in ee3698b6 — the Slinky Slurm Inline Components section now sits after the cross-cutting-overlay content, immediately before Inference performance constraints, with Multi-Level Inheritance contiguous again. Heading text unchanged; both inbound anchors (component-catalog.md, cuj1 demo) verified resolving.

recipes/checks/slinky-topograph/health-check.yaml
ArangoGutierrez: Implemented in this PR rather than as a fast-follow (ee3698b6): the shared assert is now `(contains("topology.conf", 'SwitchName')): true`. Verified both directions against a scratch cluster — synced tree content passes immediately; seed-only content fails closed at the assert timeout. contains(null, ...) on a missing key also fails closed, and the read-only allowlist test stays green. The step and header comments now describe the fail-closed semantics; the PR body's Risk Assessment bullet is updated to match.

demos/cuj1-slinky-slurm.md
ArangoGutierrez: Swapped in ee3698b6 — topograph uninstalls before slinky-slurm, closing the orphan-CM recreation window.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -88,6 +102,16 @@ spec:
...
+        configFiles:
+          # Seed so topology.conf exists in slinky-slurm-config-extra (and is
+          # mounted at /etc/slurm) from first boot. Topograph overwrites this
+          # key after each successful topology sync. NOTE: helm upgrade resets
+          # the key to this seed until the next Topograph sync.
+          topology.conf: |
+            # Managed by NVIDIA Topograph (engine: slinky). Pre-sync seed.
+        controller:
+          extraConfMap:
+            TopologyPlugin: "topology/tree"
```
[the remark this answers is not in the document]
ArangoGutierrez: Verified against live behavior before acting — the boot-failure claim doesn't hold on this stack, and the suggested fixes would regress the review fix that just landed:

1. **Empirically disproven for this deployment (slurm chart v1.2.0):** in the live Kind verification for this PR, slurmctld started with the comment-only seed in place and ran 2/2 Ready, restarts=0, through Topograph's first sync — and because `slinky-slurm` deploys *before* `slinky-topograph`, slurmd node registration also happened during the seed-only window. No segfault, no crashloop. Evidence is in the PR's Testing section.
2. **A `SwitchName=`-bearing seed would vacuously pass the strengthened health check** requested by @mchmarny in this very review round (`contains("topology.conf", 'SwitchName')`), silently re-opening the seeded-check gap it exists to close. That trade is strictly worse than the theoretical boot risk.
3. **"Defer plugin activation until after the first sync" has no mechanism** in static chart values — it would require post-sync reconfigure automation that doesn't exist in this design.

If a slurm version in a future chart bump does fail on zero-switch tree topology, it fails loudly (CrashLoopBackOff → health checks red), not silently. Added a comment in both overlays (419aca32) documenting why the seed is deliberately comment-only, so this doesn't resurface.

demos/cuj1-slinky-slurm.md
```diff
@@ -26,7 +26,7 @@ Slurm leaves are built from criteria flags (`--service`, `--platform slurm`, …
 
 ## Generate Recipe (Query Mode)
 
-Pick the row that matches your cluster. Each resolves to a slurm leaf with three inline Slinky components: `slinky-slurm-operator-crds`, `slinky-slurm-operator`, and `slinky-slurm`.
+Pick the row that matches your cluster. Each resolves to a slurm leaf with at least three inline Slinky components: `slinky-slurm-operator-crds`, `slinky-slurm-operator`, and `slinky-slurm`. The GKE and Kind leaves also include `slinky-topograph` (topology-aware scheduling) — GKE with the `gcp` provider, Kind with the `test` provider and a fixed topology fixture. The EKS leaf does not include it today; see [Slinky Slurm Inline Components](../docs/integrator/recipe-development.md#slinky-slurm-inline-components) to add it to another leaf.
```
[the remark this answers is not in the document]
ArangoGutierrez: Updated in 3bee1bbf — the deploy-order line now includes `slinky-topograph` with its leaf scoping: it follows `slinky-slurm` on the GKE and Kind leaves (matching the recipe dependencyRefs; the slurm chart must own the ConfigMap Topograph patches), and the EKS leaf doesn't carry it.

docs/user/container-images.md
```diff
@@ -19,8 +19,8 @@ A machine-readable **CycloneDX 1.6 JSON** companion to this page is produced by
 <!-- BEGIN AICR-BOM -->
 ## Summary
 
-- Components: **33**
-- Unique images: **85**
+- Components: **34**
+- Unique images: **86**
```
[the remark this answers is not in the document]
ArangoGutierrez: No inconsistency — the arithmetic in the finding is off. The Components table's Images column sums to **90** across its 35 rows (34 components + 1 variant row), not 82, and the image list below it contains 90 references of which **86 are unique** — shared images (e.g. cert-manager sidecars) appear under multiple components, so unique ≤ total-references is the expected relationship, not a drift signal. This section is machine-generated: `make bom-docs` was run fresh for this branch and again after the rebase onto current main, and `TestCommittedBOMVersionsMatchRegistry` (part of `make qualify`, green post-rebase) gates the version column and component set. No change needed.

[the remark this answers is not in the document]
ArangoGutierrez: Correction to my previous reply — the table-sum figure I quoted (90) was wrong; the Components table's Images column does sum to **82** as stated. The 82-vs-86 relationship is still not drift, though — it's the generator's semantics, and the numbers reconcile exactly:

- The Components table counts images at each component's **registry-default version** → 82.
- `## Version variants` lists kube-prometheus-stack 83.7.0 (declared by the aks overlay) with 8 images, **4 of which are version-differing tags** not present in the default set.
- Unique images = 82 + 4 = **86**.

The same relationship exists on main before this PR (33 components, table sum 81, unique 85 — identical +4 variant delta); this PR shifts each figure by exactly +1 for slinky-topograph. `make bom-docs` was already run fresh twice on this branch and reproduces these numbers — there is nothing to regenerate. If the summary should be self-explanatory (e.g. "Unique images: 86 (incl. version-variant tags)"), that's a change to the BOM generator, better as a follow-up than in this PR.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -88,6 +102,21 @@ spec:
...
+          # The seed is deliberately comment-only (no SwitchName=): slurmctld
+          # (slurm chart v1.2.0) boots and registers nodes with a zero-switch
+          # topology/tree config (live-verified pre-first-sync), and the
+          # registry health check fails closed until real SwitchName content
+          # lands — a SwitchName-bearing seed would pass that check vacuously.
+          topology.conf: |
+            # Managed by NVIDIA Topograph (engine: slinky). Pre-sync seed.
+        controller:
+          extraConfMap:
+            TopologyPlugin: "topology/tree"
```
yuanchen8911, line 125: The generated topology is never loaded into the running slurmctld, so topology-aware placement stays inactive on both leaves. In slurm-operator v1.2.0, controller.inplaceReconfigure defaults to false (helm/slurm/values.yaml L190-193), the reconfigure sidecar is created only when it is true, and the controller rollout hash covers only the controller's own config ConfigMap plus auth secrets — it excludes configFileRefs ConfigMaps (internal/builder/controllerbuilder/controller_app.go L102-168, L349-369). Topograph's patch to slinky-slurm-config-extra therefore reaches the projected file in the pod but triggers neither scontrol reconfigure nor a pod recreation; slurmctld keeps the boot-time (seed) topology indefinitely. Meanwhile the new health check asserts SwitchName content on the ConfigMap, so it passes while the feature is inactive.

Please add a reload path (e.g. controller.inplaceReconfigure: true, with the seed caveat noted on the topology.conf seed line) and verify with scontrol show topology after the first sync. Applies equally to the Kind overlay.

yuanchen8911 (commented):
Issues 1 and 2 (topology reload path, and the zero-switch seed constraint on that fix) should be fixed before merge. Issue 3 is a non-blocking fixture correction.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -88,6 +102,21 @@ spec:
...
+          # mounted at /etc/slurm) from first boot. Topograph overwrites this
+          # key after each successful topology sync. NOTE: helm upgrade resets
+          # the key to this seed until the next Topograph sync.
+          # The seed is deliberately comment-only (no SwitchName=): slurmctld
+          # (slurm chart v1.2.0) boots and registers nodes with a zero-switch
+          # topology/tree config (live-verified pre-first-sync), and the
+          # registry health check fails closed until real SwitchName content
+          # lands — a SwitchName-bearing seed would pass that check vacuously.
+          topology.conf: |
+            # Managed by NVIDIA Topograph (engine: slinky). Pre-sync seed.
```
yuanchen8911: The comment-only seed has zero switches, and Slurm's tree plugin fatals on zero-switch validation once node records exist (slurm-26.05 src/plugins/topology/tree/switch_record.c L427-431). At first boot this is masked (validation is skipped before nodes register), but it makes a bare inplaceReconfigure: true fix unsafe: the reconfigure sidecar hashes every file under /etc/slurm, and helm upgrade explicitly restores the seed (as this comment notes), so a reconfigure against the zero-switch seed after nodes have registered is reachable through an ordinary helm upgrade. The exact in-container failure mode still needs live verification; it may persist until another Topograph generation trigger restores valid topology, and without working GCP identity recovery cannot succeed.

Please use a syntactically valid nonzero placeholder topology (and make the health check distinguish the placeholder from synced output), or otherwise ensure reload occurs only after valid topology has landed. Live evidence to cover: controller restart after sync, and helm upgrade/seed reset with registered nodes. Applies equally to the Kind overlay.

recipes/overlays/h100-kind-training-slurm.yaml
```diff
@@ -58,12 +58,157 @@ spec:
...
+      dependencyRefs:
+        - slinky-slurm-operator
+        - slinky-slurm-operator-crds
+        - slinky-slurm
+      overrides:
+        global:
+          provider:
+            name: test
+            params:
+              modelFileName: small-tree.yaml
```
yuanchen8911: The small-tree.yaml fixture writes SwitchName entries for nodes I21-I25/I34-I36, but the Kind leaf's one-replica NodeSet registers slinky-0 — the generated topology never references a real node. Not a blocker: unknown node names in topology.conf produce a warning, not a fatal, in Slurm 26.05 (src/common/node_conf.c L1250-1283), and the one-node smoke workload uses the single-node selection path that bypasses tree-wide selection (select/cons_tres job_test.c, max_nodes==1 fast path). But it means the leaf never validates effective topology end-to-end, and would surface as a real scheduling problem if the NodeSet were scaled.

Since the fixture is embedded in Topograph v0.5.0 (tests/models/), please supply or mount a Kind test model whose generated topology references slinky-0 (the test provider accepts a path to a custom model file).

recipes/overlays/h100-gke-cos-training-slurm.yaml
ArangoGutierrez: Good catch — the previous revision generated `topology.conf` but never guaranteed the running `slurmctld` picked it up. Fixed: both Slurm leaves now set `controller.inplaceReconfigure: true`, with a comment explaining why. Without it the controller's rollout hash excludes the `configFileRefs` ConfigMap, so Topograph's `topology.conf` updates land in the ConfigMap but never reach the live daemon; with it, the reconfigure sidecar runs `scontrol reconfigure` whenever a file under `/etc/slurm` changes.

- [h100-gke-cos-training-slurm.yaml#L126-L132](https://github.com/NVIDIA/aicr/blob/274fe43fab95dc0c0949d55dbe825839c60e6b16/recipes/overlays/h100-gke-cos-training-slurm.yaml#L126-L132)
- [h100-kind-training-slurm.yaml#L227-L233](https://github.com/NVIDIA/aicr/blob/274fe43fab95dc0c0949d55dbe825839c60e6b16/recipes/overlays/h100-kind-training-slurm.yaml#L227-L233)

I verified this live on Kind. After the first sync, the ConfigMap and the running daemon both reflect the generated topology (the sidecar fired):

```
$ kubectl -n slurm get cm slinky-slurm-config-extra -o jsonpath='{.data.topology\.conf}'
SwitchName=SW1 Nodes=slinky-0

$ kubectl -n slurm exec slinky-slurm-controller-0 -c slurmctld -- scontrol show topology
SwitchName=SW1 Level=0 LinkSpeed=1 Nodes=slinky-0
```

And it survives a controller restart — I deleted `slinky-slurm-controller-0`, the StatefulSet recreated it, and the reloaded topology held with no fatal:

```
$ kubectl -n slurm exec pod/slinky-slurm-controller-0 -c slurmctld -- scontrol ping
Slurmctld(primary) at slinky-slurm-controller-0 is UP
$ kubectl -n slurm exec pod/slinky-slurm-controller-0 -c slurmctld -- scontrol show topology
SwitchName=SW1 Level=0 LinkSpeed=1 Nodes=slinky-0
```

The reload mechanics are provider-independent — the reconfigure sidecar and the `configFileRefs` projection are identical on every leaf. Kind is the evidence environment (it carries the `test` provider); on GKE only the provider's topology output and cloud identity differ, not the path that gets it into `slurmctld`.

recipes/overlays/h100-gke-cos-training-slurm.yaml
ArangoGutierrez: Right — a zero-switch seed would fatal the `topology/tree` plugin. The seed is now a valid one-switch `aicr-preseed` placeholder rather than a comment-only stub, so the tree plugin always has a well-formed switch to parse (at boot, after a restart, and after a helm-upgrade seed restore):

- [h100-gke-cos-training-slurm.yaml#L120-L122](https://github.com/NVIDIA/aicr/blob/274fe43fab95dc0c0949d55dbe825839c60e6b16/recipes/overlays/h100-gke-cos-training-slurm.yaml#L120-L122)

Fail-closed is preserved because the health check requires `SwitchName` content *without* the `aicr-preseed` marker, so an unconfigured provider stuck on the placeholder reports red.

I exercised the exact case you flagged — a helm upgrade reset the ConfigMap back to the placeholder while `slinky-0` was registered — and the running `slurmctld` reconfigured against it and stayed up, with zero restarts and zero fatal lines. Raw capture:

```
$ kubectl -n slurm exec slinky-slurm-controller-0 -c slurmctld -- scontrol ping   (against placeholder)
Slurmctld(primary) at slinky-slurm-controller-0 is UP
$ kubectl -n slurm exec slinky-slurm-controller-0 -c slurmctld -- scontrol show topology
SwitchName=aicr-preseed Level=0 LinkSpeed=1 Nodes=
slurmctld restartCount: 0
'fatal:' line count in slurmctld log: 0
```

`Nodes=` is empty only because `aicr-preseed-node` isn't a registered node; `slinky-0` is simply left unplaced. Forcing a Topograph resync recovered the running topology to `SwitchName=SW1 Level=0 LinkSpeed=1 Nodes=slinky-0`.

The health check discriminates on content — the `validate-topology-configmap` assert errors on the placeholder and passes once synced. Raw chainsaw lines:

```
RUN #1 (placeholder):
validate-topology-configmap | ASSERT | ERROR
  * data.topology.conf: Invalid value: "# Managed by NVIDIA Topograph (engine: slinky). Pre-sync placeholder.\nSwitchName=aicr-preseed Nodes=aicr-preseed-node\n": Expected value: "SwitchName=SW1 Nodes=slinky-0\n"

RUN #2 (synced):
validate-topology-configmap | ASSERT | DONE
```

One caveat on the green run so it isn't misread: chainsaw's overall exit is also `--- FAIL` (`Failed tests 1`), but that is solely the temp-namespace CLEANUP hitting its 30s deletion deadline (`context deadline exceeded`) *after* the assert had already passed (`validate-topology-configmap | ASSERT | DONE`) — not a health-check failure.

recipes/overlays/h100-kind-training-slurm.yaml
ArangoGutierrez: Agreed — the embedded `small-tree.yaml` fixture references nodes that never exist on this cluster, so it couldn't validate effective topology. The Kind leaf now ships a test-provider model (`kind-slinky.yaml`, `SwitchName=SW1 Nodes=slinky-0`) that references the real one-replica slurmd node, mounted into the Topograph Deployment via a `preManifestFiles` ConfigMap plus the chart's volume/volumeMount:

- [h100-kind-training-slurm.yaml#L68-L91](https://github.com/NVIDIA/aicr/blob/274fe43fab95dc0c0949d55dbe825839c60e6b16/recipes/overlays/h100-kind-training-slurm.yaml#L68-L91)
- [kind-test-model.yaml](https://github.com/NVIDIA/aicr/blob/274fe43fab95dc0c0949d55dbe825839c60e6b16/recipes/components/slinky-topograph/manifests/kind-test-model.yaml)

Live, Topograph reads that model and the generated `topology.conf` references the real node, effective in the running daemon:

```
$ kubectl -n slurm get cm slinky-slurm-config-extra -o jsonpath='{.data.topology\.conf}'
SwitchName=SW1 Nodes=slinky-0
$ kubectl -n slurm exec slinky-slurm-controller-0 -c slurmctld -- scontrol show topology
SwitchName=SW1 Level=0 LinkSpeed=1 Nodes=slinky-0

topograph API log:
  "Using simulated topology from" modelFileName="/etc/topograph/models/kind-slinky.yaml"
```

The pre-manifest is deliberately ConfigMap-only (no `Namespace` object): shipping a `Namespace` makes the bundler drop `--create-namespace` on the pre folder, and Helm 4 then fails validating the ConfigMap against the not-yet-existing namespace. Keeping it ConfigMap-only preserves `--create-namespace`, so the pre-manifest installs cleanly under helm v4.2.3 on a fresh cluster (topograph namespace absent). Raw helm output:

```
Release "slinky-topograph-pre" does not exist. Installing it now.
NAME: slinky-topograph-pre
NAMESPACE: topograph
STATUS: deployed
REVISION: 1
DESCRIPTION: Install complete
```

The run completed with no `namespaces "topograph" not found` and no ownership-import error; it created the topograph namespace and the `slinky-topograph-test-model` ConfigMap (Helm-release-owned).

recipes/components/slinky-topograph/values.yaml
```diff
@@ -0,0 +1,86 @@
...
+# Stabilize the Deployment name to "slinky-topograph"; health-check assertions
+# and dependencyRef consumers rely on it.
+fullnameOverride: slinky-topograph
+
+global:
+  service:
+    type: ClusterIP
+    port: 49021
+  engine:
+    name: slinky
```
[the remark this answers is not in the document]
ArangoGutierrez: Thanks — you're right that the comment was misleading, and I've corrected it (commit `274fe43f`). Two parts, one accepted and one adjusted:

- [values.yaml#L25-L34](https://github.com/NVIDIA/aicr/blob/274fe43fab95dc0c0949d55dbe825839c60e6b16/recipes/components/slinky-topograph/values.yaml#L25-L34)

On the provider list, the enum I now document is the authoritative one from the chart's own `values.schema.json` at tag `v0.5.0` — 20 values, which includes several the old comment omitted (`nscale`, `nscale-sim`, `dsx-sim`, `nebius-sim`, `aws-sim`, `cw`, `gcp-sim`, `infiniband-bm`, `lambdai`/`-sim`, `oci-imds`, `oci-sim`):

https://raw.githubusercontent.com/NVIDIA/topograph/v0.5.0/charts/topograph/values.schema.json

On `global.engine.name`: it stays pinned to `slinky` in this shared component values file rather than being a required per-overlay override. The chart schema requires `global.engine`, but this file already satisfies it, so overlays only need to set `provider.name`. The comment now says exactly that.

mchmarny (approved):
Adds the slinky-topograph component wired end-to-end into slurmctld via the chart-owned config-extra ConfigMap, cleanly superseding #1554 with all 5 blocking threads resolved. Reviewed the full diff: the deploy-order flip is correctly guarded by `deployment_order_guard_test.go` on both slurm leaves, the health check fails closed (requires `SwitchName` content without the `aicr-preseed` marker, so an unconfigured provider can't pass vacuously on the seed), the Kind leaf's inline asserts validate effective topology against the real `slinky-0` node, and the `inplaceReconfigure` reload path closes the "generated but never loaded" gap @yuanchen8911 flagged — with live Kind evidence (`scontrol show topology`, restart and helm-upgrade seed-reset both exercised). CI is fully green (KWOK, tiered e2e, lint, BOM freshness), docs/BOM/catalog updated in-PR. @yuanchen8911's three open threads all have fixes landed in the current diff with verification attached; nothing blocks merge.

PR #1394, opened 2026-06-17
feat(slinky-slurm): add conformance validator and CUJ demo

kaynetu:
## Summary

Adds the `slinky-slurm-health` conformance validator (pod-exec checks for `scontrol ping`, idle/mixed node inventory, and bounded `srun`), wires it into slurm training leaves and the validator catalog, documents the EKS/GKE/Kind CUJ in `demos/cuj1-slinky-slurm.md`, and tunes the Slurm deployment Chainsaw assert timeout under the `expected-resources` envelope.

## Motivation / Context

Slinky Slurm leaves already deploy via `--platform slurm`, but conformance had no workload-specific signal that slurmctld is reachable, workers are idle/mixed, and a minimal job can run through slurmd. This closes that gap and documents the full recipe → bundle → deploy → validate workflow (without snapshot) for EKS, GKE, and Kind.

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] CLI (`cmd/aicr`, `pkg/cli`)
- [ ] API server (`cmd/aicrd`, `pkg/server`)
- [x] Recipe engine / data (`pkg/recipe`)
- [ ] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [ ] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [x] Validator (`pkg/validator`, `validators/conformance`)
- [ ] Core libraries (`pkg/errors`, `pkg/k8s`)
- [x] Docs/examples (`docs/`, `demos/`)
- [x] Other: `recipes/validators/catalog.yaml`, `recipes/checks/slinky-slurm/`, slurm overlays, `vendor/`

## Implementation Notes

- **`slinky-slurm-health` conformance check** — discovers login/worker pods via Slinky LoginSet/NodeSet CR selectors, then execs from the login pod: `scontrol ping`, `sinfo -h -Ne -t idle,mix | grep -q .`, and `srun --immediate=5 --time=0:03 hostname`. Shared pod-exec helper in `validators/conformance/pod_exec.go` (pulls in `k8s.io/client-go/tools/remotecommand` + vendored deps).
[...]

PR #1744, opened 2026-07-14
feat(slinky-slurm): add AKS H100 recipe, Enroot/Pyxis support, and Gres/GPU config

kaynetu:
## Summary

Adds an H100 AKS Ubuntu training Slurm recipe and enables Enroot/Pyxis container execution by default for Slinky login and worker pods.

Completes GPU GRES and `task/cgroup` isolation across GPU-backed Slurm leaves, and extends `slinky-slurm-health` with a GPU-gated container execution check.

## Motivation / Context

AICR had H100 Slurm leaves for EKS and GKE, but not AKS. The shared Slinky defaults also used the vanilla login and slurmd images, so `srun --container-image` was not available without manual image and plugstack overrides.

GPU leaves declared their node GRES and Kubernetes GPU limits, but did not consistently enable `GresTypes=gpu` or `task/cgroup`. That left `srun --gpus` behavior and allocated-device isolation incomplete. Conformance validated basic Slurm scheduling but did not prove that a GPU allocation could launch a Pyxis container.

Fixes: N/A
Related: #1394, #1539

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: Slinky component values, Slurm leaf overlays, and `validators/conformance`

## Implementation Notes

- Adds `h100-aks-ubuntu-training-slurm`, including the Slinky CRDs, operator, cluster, eight-GPU H100 NodeSet configuration, inherited deployment/conformance checks, and AKS bundle guidance.
- Switches the shared Slinky login and slurmd defaults to the pinned `login-pyxis` and `slurmd-pyxis` images and loads Pyxis through `configFiles.plugstack.conf`.
- Configures `GresTypes=gpu` and `TaskPlugin=task/cgroup,task/affinity` for the GB200 EKS and H100 AKS/EKS/GKE Slurm leaves. The existing `ConstrainDevices=yes` setting then restricts jobs to their allocated GPU devices.
- Extends `slinky-slurm-health` to reuse the discovered runnable NodeSet pods, detect positive `nvidia.com/gpu` requests or limits, and append a bounded one-GPU `srun --container-image=docker://alpine:3.20` check only on GPU-backed clusters. CPU-only Kind and KWOK paths do not run it.
- Adds recipe and validator regression coverage for the AKS leaf, GRES/task-cgroup configuration, Pyxis defaults, GPU detection, command selection, inventory recording, and the shared IMEX NodeSet discovery path.
- Updates the Slinky CUJ, component and validator documentation, generated recipe-health report, and container-image inventory.

## Testing

```bash
make qualify
```

`make qualify` passed locally.

After pinning the conformance image to Alpine 3.20:

```bash
GOFLAGS=-mod=vendor go test -race ./validators/conformance -run TestCheckSlinkySlurmHealth -count=1
golangci-lint run -c .golangci.yaml ./validators/conformance/...
```

Both focused checks passed.

**Live GB200 cluster — GPU device isolation:** requested two GPUs in an interactive Slurm allocation and confirmed that `nvidia-smi` exposed exactly two GB200 devices:

```text
$ srun --gpus=2 --pty bash
root@slinky-1:/tmp# nvidia-smi
NVIDIA-SMI 580.126.20    Driver Version: 580.126.20    CUDA Version: 13.0

GPU 0  NVIDIA GB200
GPU 1  NVIDIA GB200
```

**Live GB200 cluster — GPU container conformance:** before the image reference was pinned, the check ran with `alpine:latest` from the login pod, received a GPU allocation, launched Alpine through Pyxis/Enroot, and exited successfully. The submitted implementation now pins `alpine:3.20` for reproducibility:

```text
Command: srun --immediate=30 --time=1:00 --nodes=1 --ntasks=1 \
  --cpus-per-task=1 --mem=128M --gpus=1 \
  --container-image=docker://alpine:latest cat /etc/os-release
ExitCode: 0

stdout:
NAME="Alpine Linux"
ID=alpine
VERSION_ID=3.24.1
PRETTY_NAME="Alpine Linux v3.24"
```

## Risk Assessment

**Rollout notes:** Generated Slinky bundles now use the Pyxis-enabled login and slurmd images by default. GPU-backed Slurm leaves enable GPU GRES plus task/cgroup isolation, while CPU-only leaves omit those GPU-specific controller settings. Rebuild and publish the conformance validator image before expecting the new live GPU container check; stage the updated Slinky workloads before production rollout.

## Checklist


validators/conformance/slinky_slurm_health_check.go
```diff
@@ -77,6 +77,24 @@ var slinkySlurmHealthCommands = []slinkySlurmHealthCommand{
...
+		"--cpus-per-task=1",
+		"--mem=128M",
+		"--gpus=1",
+		"--container-image=docker://alpine:latest",
+		"cat",
+		"/etc/os-release",
+	},
+	requireStdout: true,
+}
+
```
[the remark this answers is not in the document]
kaynetu: Intentionally keep the check to container execution health.  nvidia-smi check with image import will be too heavy at this moment.

demos/cuj1-slinky-slurm.md
```diff
@@ -112,6 +113,27 @@ GKE system nodes should **not** carry custom taints (konnectivity and other mana
...
+  --accelerated-node-toleration nvidia.com/gpu=present:NoSchedule \
+  --set slinkyslurm:nodesets.slinky.replicas=2 \
+  --set-json 'slinkyslurm:controller.podSpec={"nodeSelector":{"agentpool":"cpuworker1"}}' \
+  --set-json 'slinkyslurm:restapi.podSpec={"nodeSelector":{"agentpool":"cpuworker1"}}' \
+  --set-json 'slinkyslurm:loginsets.slinky.podSpec={"nodeSelector":{"agentpool":"cpuworker1"}}' \
+  --output bundle
+```
+
+AKS always ships `managed-csi` as the default StorageClass; omit `--storage-class` unless you need a non-default class.
+
```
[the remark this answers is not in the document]
kaynetu: It's a demo example. It's working fine.

recipes/overlays/h100-gke-cos-training-slurm.yaml
```diff
@@ -94,6 +98,18 @@ spec:
...
+            GresTypes: "gpu"
+            TaskPlugin: "task/cgroup,task/affinity"
+            TopologyPlugin: "topology/tree"
+          # Reload path (review: PR #1737): the reconfigure sidecar (slurm
+          # chart v1.2.0, controller.inplaceReconfigure) runs scontrol
+          # reconfigure when any file under /etc/slurm changes. Without it the
+          # controller rollout hash excludes configFileRefs ConfigMaps, so
+          # Topograph's topology.conf updates never reach the running
+          # slurmctld and topology-aware placement stays inactive.
+          inplaceReconfigure: true
```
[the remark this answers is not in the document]
kaynetu: this change is not from my change. It's from other PRs

validators/conformance/slinky_slurm_health_check.go
```diff
@@ -136,9 +157,39 @@ func resolveSlinkySlurmNamespace(ctx *validators.Context) string {
 	return slinkySlurmNamespace
 }
 
-func runSlinkySlurmHealthCommands(ctx *validators.Context, namespace, loginPodName string) []string {
+func slinkySlurmHealthCommandsForNodeSetPods(pods []corev1.Pod) []slinkySlurmHealthCommand {
+	commands := append([]slinkySlurmHealthCommand(nil), slinkySlurmHealthCommands...)
+	if nodeSetPodsRequestNVIDIAGPUs(pods) {
```
mchmarny, line 162: The GPU container check silently downgrades to the CPU-only command set when no NodeSet pod carries a positive `nvidia.com/gpu` request/limit. On a GPU leaf where the slurmd limit got dropped (bad `--set` override, chart values drift), conformance would still PASS without ever exercising Pyxis — a false-PASS path that's invisible in the evidence. Consider recording an artifact (or at least a log line) stating whether the GPU container check was included and why, so a downgrade shows up in the conformance output instead of passing quietly.

validators/conformance/slinky_slurm_health_check.go
```diff
@@ -77,6 +77,24 @@ var slinkySlurmHealthCommands = []slinkySlurmHealthCommand{
...
+	command: []string{
+		"srun",
+		"--immediate=30",
+		"--time=1:00",
+		"--nodes=1",
+		"--ntasks=1",
+		"--cpus-per-task=1",
+		"--mem=128M",
+		"--gpus=1",
+		"--container-image=docker://alpine:3.20",
```
mchmarny: Two things on the image reference: (1) nit: the component pins its Alpine sidecars to `3.23.3` (`components/slinky-slurm/values.yaml`) — aligning this to `alpine:3.23` keeps one Alpine pin story per component. (2) This adds a runtime Docker Hub pull to conformance: egress-restricted/air-gapped clusters and Hub rate limits will now fail `slinky-slurm-health` on GPU clusters for reasons unrelated to Slurm health. The validator catalog already supports `env:` — worth making the image overridable (env var with this default) or documenting the docker.io egress requirement.

docs/user/recipe-health.md
```diff
@@ -51,7 +51,8 @@ The matrix is computed **hermetically and offline**: every signal is a pure read
 | monitoring-hpa | — | — | — | — | — | pass | R:0 D:0 P:0 C:0 | pending |
 | a100-aks-ubuntu-training-kubeflow | aks | a100 | ubuntu | training | kubeflow | pass | R:0 D:4 P:0 C:10 | pending |
 | h100-aks-ubuntu-inference-dynamo | aks | h100 | ubuntu | inference | dynamo | pass | R:0 D:4 P:1 C:11 | pending |
-| h100-aks-ubuntu-training-kubeflow | aks | h100 | ubuntu | training | kubeflow | pass | R:0 D:4 P:0 C:10 | pending |
+| h100-aks-ubuntu-training-kubeflow | aks | h100 | ubuntu | training | kubeflow | pass | R:0 D:4 P:1 C:10 | pending |
```
mchmarny, line 54: `h100-aks-ubuntu-training-kubeflow` P:0 → P:1 changed here, but this PR doesn't touch that leaf or its parent. Is this pre-existing drift on main that the regeneration picked up? Fine either way — just confirming it's not an accidental side effect of the overlay changes.

demos/cuj1-slinky-slurm.md
```diff
@@ -112,6 +113,27 @@ GKE system nodes should **not** carry custom taints (konnectivity and other mana
...
+Example layout: 3× `system` (`CriticalAddonsOnly=true:NoSchedule`), 1× `cpuworker1` (untainted), 2× `gpu-worker` (`nvidia.com/gpu=present:NoSchedule`). Operator on **system**; controller / login / restapi on **cpuworker1**; slurmd on **gpu-worker**.
+
+```shell
+aicr bundle \
+  --recipe recipe.yaml \
+  --deployer helm \
+  --system-node-selector agentpool=system \
+  --system-node-toleration CriticalAddonsOnly=true:NoSchedule \
+  --accelerated-node-selector nodeGroup=gpu-worker \
```
mchmarny: nit: `nodeGroup=gpu-worker` — AKS node pools natively carry `agentpool=<pool>` (used for the system and CPU pools two lines up). If `nodeGroup` is a custom label from your test cluster, either switch the example to `agentpool=gpu-worker` or add a note that it's a user-applied label, so readers don't assume AKS sets it.

mchmarny (commented):
Solid, complete change — the AKS leaf mirrors the EKS/GKE shape exactly, the four-place GRES/isolation wiring is well documented in the overlay comments, and the live GB200 verification of both device isolation and the Pyxis launch path is exactly the right evidence for this kind of change. The KWOK-skip refactor (`runnableSlinkyNodeSetPods` returning the pods it already listed) is a nice cleanup too.

Four comments inline: one medium (GPU check downgrade is silent — a GPU leaf that loses its `nvidia.com/gpu` limit would pass conformance without exercising Pyxis), one on the conformance image (Docker Hub egress dependency + Alpine pin consistency), and two doc-level questions/nits. Nothing blocks merge. Branch is behind `main` — please rebase before merging per repo policy.

validators/conformance/slinky_slurm_health_check.go
```diff
@@ -136,9 +157,39 @@ func resolveSlinkySlurmNamespace(ctx *validators.Context) string {
...
 }
 
-func runSlinkySlurmHealthCommands(ctx *validators.Context, namespace, loginPodName string) []string {
+func slinkySlurmHealthCommandsForNodeSetPods(pods []corev1.Pod) []slinkySlurmHealthCommand {
+	commands := append([]slinkySlurmHealthCommand(nil), slinkySlurmHealthCommands...)
+	if nodeSetPodsRequestNVIDIAGPUs(pods) {
+		commands = append(commands, slinkySlurmGPUContainerHealthCommand)
+	}
+	return commands
+}
```
yuanchen8911: When no NodeSet pod carries a positive `nvidia.com/gpu` request or limit, this silently falls back to the CPU-only command set, so `slinky-slurm-health` can pass on a cloud GPU Slurm recipe whose deployment has lost its GPU resources — a false pass in the direction this check was added to close.

The state is reachable through documented behavior: `--set-json` deletes a key on JSON `null` (docs/user/cli-reference.md), and `slinkyslurm` is an exposed override alias, so a bundle can drop `nodesets.slinky.slurmd.resources` while the validation recipe still promises GPU workers.

The plumbing to fail closed already exists: checks receive `ctx.ValidationInput`, which carries `Criteria` copied from the resolved recipe (pkg/validator/v1/validation_input.go), and `GPUAllocationPolicy` (#1327) is precedent for a recipe-resolved value driving strict verification with a capability fallback for standalone runs. Suggested behavior:

- criteria with a non-empty/non-`any` service other than `kind` and a non-empty/non-`any` accelerator: fail when no NodeSet pod has a positive `nvidia.com/gpu` request or limit
- `kind`: keep CPU-only execution and record the skip as an artifact
- missing/incomplete criteria: keep the current capability-based gating

validators/conformance/slinky_slurm_health_check.go
yuanchen8911: This hardcodes a Docker Hub pull at validation time: enroot resolves `docker://alpine:3.20` from the Slurm worker, so egress-restricted or air-gapped GPU clusters (or anonymous-pull throttling) fail `slinky-slurm-health` even when Pyxis works. The reference also never surfaces in the image inventory — the BOM/mirror tooling extracts images from rendered charts only, so it cannot be discovered via `aicr mirror list` — and `AICR_VALIDATOR_IMAGE_REGISTRY`, which is forwarded into the validator Job env for exactly this purpose (pkg/validator/v1/job_plan_internal.go), is ignored.

Suggested fix: build the reference from the fully-qualified `docker.io/library/alpine:3.20` and apply a registry-only rewrite when `AICR_VALIDATOR_IMAGE_REGISTRY` is set. Two traps to avoid: `catalog.ResolveImage` is unsuitable as-is (its `AICR_VALIDATOR_IMAGE_TAG` step rewrites explicit tags too, and `replaceRegistry` returns single-segment names like `alpine:3.20` unchanged). For discoverability, document the image in the validator/air-gap docs rather than hand-editing the generated section of docs/user/container-images.md — the next `make bom-docs` would remove a manual entry.

recipes/components/slinky-slurm/values.yaml
```diff
@@ -129,6 +132,12 @@ loginsets:
   slinky:
     enabled: true
     replicas: 1
+    # Use Slinky's login-pyxis image (enroot + pyxis SPANK + hooks baked
+    # in) instead of the chart's vanilla login default.
+    login:
+      image:
+        repository: ghcr.io/slinkyproject/login-pyxis
+        tag: 26.05-ubuntu26.04
```
yuanchen8911, line 140: These new in-tree image overrides are tag-only. ADR-006 (docs/design/006-image-pinning-policy.md, Decision item 2 and the new-components contract) requires explicit in-tree overrides to carry an `@sha256:` digest, and the chart supports it: slurm 1.2.0 exposes `digest:` on every image block and renders `repository@digest` when set.

Current multi-arch index digests (verified against ghcr.io):

```yaml
loginsets:
  slinky:
    login:
      image:
        digest: sha256:9e782d1a645aff1dedc498d7a3256733cde55a152659f44716e8a5f0dca02028
```

```yaml
nodesets:
  slinky:
    slurmd:
      image:
        digest: sha256:0c03f87d5b5725df2d11392702fb647922b3060c076e9ce4b4f13c9a67c904b3
```

The slurmd image block later in this file needs the same treatment. After pinning, rerun `make bom-docs` — the rendered references become `repository@sha256:...`.

recipes/overlays/h100-aks-ubuntu-training-slurm.yaml
```diff
@@ -0,0 +1,113 @@
...
+    os: ubuntu
+    intent: training
+    platform: slurm
+
+  mixins:
+    - os-ubuntu
+
+  constraints:
+    - name: K8s.server.version
+      value: ">= 1.32.4"
```
yuanchen8911: This resolves the new recipe to `>= 1.32.4`, while recipes/overlays/aks.yaml states that all AKS recipes include DRA and pins `>= 1.34` because DRA graduated to GA in Kubernetes 1.34. The value itself is inherited — h100-aks-training and h100-aks-ubuntu-training already set `>= 1.32.4`, so removing this constraint would change nothing — but a new recipe should not silently extend the inconsistency.

Either bump this leaf to `>= 1.34` and file a family-wide follow-up for the existing AKS training leaves, or keep `>= 1.32.4` and reconcile the aks.yaml rationale.

yuanchen8911 (changes requested):
Requesting changes for four issues, each detailed inline:

1. The GPU container smoke test silently downgrades to CPU-only commands when NodeSet pods lack GPU resources, allowing a false pass on cloud GPU Slurm recipes.
2. The smoke-test image `docker://alpine:3.20` is hardcoded with no registry override and is invisible to the mirror/BOM inventory.
3. The new Pyxis image overrides are tag-only; ADR-006 requires digest pins and the chart exposes a digest field.
4. The new AKS leaf resolves to Kubernetes `>= 1.32.4` against the `>= 1.34` DRA-GA floor stated in aks.yaml.

docs/user/recipe-health.md
kaynetu: Yes — this is pre-existing generated-doc drift on main, not a side effect of this PR’s overlays. PR #1695 (23e97a7d) added nccl-all-reduce-bw to the shared h100-aks-training parent, which the Kubeflow leaf inherits, but the committed recipe-health row remained at P:0. Regenerating the matrix correctly updates that inherited performance count to P:1.

validators/conformance/slinky_slurm_health_check.go
kaynetu: Resolved in 361b9aa6 — the GPU container check now records an include/skip artifact with the decision reason.

validators/conformance/slinky_slurm_health_check.go
kaynetu: Resolved in 361b9aa6 — Alpine is pinned to 3.23.3 with AICR_VALIDATOR_IMAGE_REGISTRY rewrite and air-gap docs.

demos/cuj1-slinky-slurm.md
kaynetu: Resolved in 361b9aa6 — the AKS demo now uses agentpool=gpuworker1.

validators/conformance/slinky_slurm_health_check.go
kaynetu: Resolved in 361b9aa6 — concrete GPU criteria now fail closed when NodeSet pods lack nvidia.com/gpu resources.

validators/conformance/slinky_slurm_health_check.go
kaynetu: Resolved in 361b9aa6 — uses fully-qualified docker.io/library/alpine:3.23.3 with registry-only rewrite and air-gap docs.

recipes/components/slinky-slurm/values.yaml
kaynetu: Resolved in 361b9aa6 — login/slurmd Pyxis images are digest-pinned and the BOM was regenerated.

recipes/overlays/h100-aks-ubuntu-training-slurm.yaml
kaynetu: Resolved in 361b9aa6 — the AKS Slurm leaf now pins K8s >= 1.34.

docs/user/air-gap-mirror.md
```diff
@@ -273,3 +273,11 @@ generate the deployment-specific manifest for a specific deployment.
...
 > run reported no warnings — otherwise the mirrored set may be silently
 > incomplete.
+
+> **Runtime validation image:** `slinky-slurm-health` launches
+> `docker.io/library/alpine:3.23.3` dynamically through `srun` on GPU-backed
+> NodeSets. Because the reference is not part of a rendered chart or manifest,
+> it does not appear in `aicr mirror list`. Mirror it separately and set
+> `AICR_VALIDATOR_IMAGE_REGISTRY` to the destination registry; the validator
+> preserves the `library/alpine:3.23.3` repository and tag. Without a registry
+> override, Slurm workers need Docker Hub egress for this check.
```
yuanchen8911, line 285: The claim that this image does not appear in `aicr mirror list` is inaccurate. The check now uses `docker.io/library/alpine:3.23.3`, which is the same image the slinky-slurm chart's initconf/logfile sidecars pin in recipes/components/slinky-slurm/values.yaml — and mirror discovery extracts from rendered charts, so the reference is already in the mirror set for every Slurm recipe (it's listed in the slinky-slurm section of docs/user/container-images.md). The same wording appears in recipes/validators/README.md.

The rest of the guidance is correct and worth keeping: mirroring alone doesn't redirect the `srun` pull, so `AICR_VALIDATOR_IMAGE_REGISTRY` must still be set.

Suggested reword for both passages: the image is already covered by `aicr mirror list` via the component's sidecar pins; the registry override is what redirects the runtime pull.

One caveat: this coverage only holds while the Go constant (`slinkySlurmGPUContainerImage`) and the values.yaml sidecar tags stay in lockstep — a Renovate bump of one without the other silently breaks it. Consider a small drift-guard test asserting they match, same pattern as `TestCacheWorkerImageMatchesTemplate` in validators/performance/model_cache.go.

recipes/overlays/h100-aks-ubuntu-training-slurm.yaml
```diff
@@ -0,0 +1,113 @@
...
+    os: ubuntu
+    intent: training
+    platform: slurm
+
+  mixins:
+    - os-ubuntu
+
+  constraints:
+    - name: K8s.server.version
+      value: ">= 1.34"
```
yuanchen8911, line 40: The bump to `>= 1.34` resolves this leaf, but the sibling AKS training leaves (h100-aks-training, h100-aks-ubuntu-training, the kubeflow and a100 variants) still resolve to `>= 1.32.4` against the DRA-GA rationale in recipes/overlays/aks.yaml. The equivalent EKS alignment already landed on main in #1747, so this is the AKS half. Could you file the follow-up issue so it doesn't get lost once this merges? Happy to file it otherwise.

yuanchen8911 (commented):
Non-blocking follow-ups on the latest push, detailed inline: an inaccuracy in the new air-gap wording, the pending family-wide AKS K8s-floor follow-up, and a suggested regression test for the new floor.

recipes/overlays/h100-aks-ubuntu-training-slurm.yaml
yuanchen8911, line 40: Consider adding a direct regression test for this floor. The rebase brought in TestGB200EKSTrainingFloor (from #1747), which asserts the resolved `>= 1.34` floor for the GB200 EKS family including its slurm leaf, but nothing asserts the resolved floor for h100-aks-ubuntu-training-slurm. Since this is a leaf constraint under last-writer-wins merge with lower-valued ancestors (aks-training at `>= 1.30`, h100-aks-ubuntu-training at `>= 1.32.4`), a future constraint edit could silently regress it — the same shape TestGB200OKEFloorNotClobbered guards against. A small table-driven test mirroring the GB200 pattern would lock it in, and gives the family-wide follow-up a ready-made place to add the sibling leaves as they're bumped.

docs/user/air-gap-mirror.md
kaynetu: Good catch — corrected. The doc claimed the image wasn't in `aicr mirror list`, which was wrong: it's the same Alpine tag the initconf/logfile sidecars pin, so chart rendering already surfaces it. Reworded both `docs/user/air-gap-mirror.md` and `recipes/validators/README.md` to say it's covered via the sidecar pins, and that `AICR_VALIDATOR_IMAGE_REGISTRY` is what redirects the runtime `srun` pull. Also added `TestSlinkySlurmGPUContainerImageTagMatchesChartSidecars` (same shape as `TestCacheWorkerImageMatchesTemplate`) so a Renovate bump of the Go constant or the sidecar tag without the other fails loudly.

recipes/overlays/h100-aks-ubuntu-training-slurm.yaml
kaynetu: Filed the family-wide follow-up: https://github.com/NVIDIA/aicr/issues/1772. It captures the current drift — `aks.yaml` declares the DRA-GA `>= 1.34` floor, but `aks-training` and the `a100-aks-*` leaves resolve to `>= 1.30` and the `h100-aks-*` non-slurm leaves to `>= 1.32.4` — and proposes either bumping the remaining leaves to `>= 1.34` or reconciling the `aks.yaml` rationale. Referenced #1747 (EKS half) and this PR.

kaynetu: Added `TestH100AKSUbuntuTrainingSlurmFloorNotClobbered` in `pkg/recipe/metadata_store_test.go`, mirroring the `TestGB200OKEFloorNotClobbered` shape — it asserts the resolved `>= 1.34` floor for `h100-aks-ubuntu-training-slurm` so a future edit to a lower-valued ancestor (aks-training `>= 1.30`, h100-aks-ubuntu-training `>= 1.32.4`) can't silently clobber it. The family-wide follow-up (#1772) has a ready-made place to add the sibling leaves as they're bumped.

yuanchen8911 (approved):
All requested changes and follow-ups are addressed and verified: fail-closed GPU gating with recorded decisions and per-tier tests, registry-configurable smoke image aligned with the component's Alpine pin plus a sidecar lockstep guard, digest-pinned Pyxis images with the BOM regenerated, the AKS leaf at Kubernetes >= 1.34 with a floor regression test, corrected air-gap/mirror docs, and the family-wide floor follow-up filed as #1772.

Issue #1466, opened 2026-06-24
Unify the two topological-sort paths in pkg/recipe


yuanchen8911:
## Summary

Unify the two topological-sort implementations in `pkg/recipe/metadata.go` into a single code path. Today there are two parallel Kahn's-algorithm implementations:

- `TopologicalSort()` — flat order, used at recipe resolution to compute `DeploymentOrder`.
- `ComponentRefsTopologicalLevels()` / `TopologicalLevels()` — level-grouped order, used at bundle generation by the helmfile deployer.

## Motivation

The two paths are duplicated logic that must be kept in sync. This was called out as known tech debt left in under release pressure when the level-grouped variant was added. The duplication directly caused extra work in #1465: honoring `overrides: {enabled: false}` had to be implemented twice (skip-disabled + drop-edges-to-declared-but-disabled) — once in each function — and the bundler needed a separate dangling-dependency prune so the helmfile path didn't re-trip a false circular-dependency error. A single implementation would have required the fix in one place.

## Proposed approach

Have one implementation be the single source of truth and express the other in terms of it (e.g. derive the flat order by flattening the levels, or factor the shared graph build + enabled-filtering + cycle detection into one helper that both thin wrappers call). Preserve current behavior:

- Deterministic ordering (alphabetical tie-break within a level / queue).
- Enabled-filtering semantics from #1465: skip disabled components; treat an edge to a declared-but-disabled component as satisfied externally; keep edges to undeclared components so missing-dependency errors still surface.
- The level grouping the helmfile deployer relies on (mutually independent components share a level).

## Acceptance criteria

- One shared graph/sort implementation; `TopologicalSort` and `TopologicalLevels`/`ComponentRefsTopologicalLevels` are thin wrappers over it (no duplicated Kahn loop or enabled-filtering logic).
- Existing tests pass unchanged, including `TestRecipeMetadataSpecTopologicalLevelsMatchesSort` (flat order stays consistent with flattened levels).
- No behavior change for recipes without disabled components.

## Context

Follow-up to #1465 (which fixed `enabled: false` across both paths) and the discussion there. Refactor only — no functional change.

yuanchen8911:
Context: follow-up to bug #1464, fixed in #1465. This issue tracks the refactor to collapse the two sort implementations into one.

PR #1746, opened 2026-07-14
feat(deployer): deploy independent components in parallel

lockwobr:
## Summary

Deploy mutually independent components **concurrently** instead of strictly one at a time, across all four concurrent deployers, and add a `--serial` escape hatch. Ordering follows each component's declared `dependencyRefs`.

## Motivation / Context

The argocd deployer assigned each component a sync-wave equal to its linear position, so components with no dependency between them still rolled out serially. Flux chained every component to the previous one, and helmfile chained every release to its predecessor via positional `needs:` — so independent components were serialized on all three. This makes them parallelize independent components while preserving genuine dependency gating.

Fixes: N/A (enhancement)
Related: #1466 / #1728 (topological-sort unification this builds on)

## Type of Change

## Component(s) Affected

- [x] CLI (`cmd/aicr`, `pkg/cli`) — new `--serial` flag
- [x] Bundlers (`pkg/bundler`, `pkg/bundler/deployer/*`)
- [x] Docs/examples (`docs/`)

## Implementation Notes

- **argocd / argocd-helm:** `wave = tier*4 + phase` (phase orders `-pre` → primary → `-post` → `-readiness`). The stride-4 band keeps consecutive tiers disjoint so a tier's readiness gate still blocks the next. argocd-helm inherits this from the delegated argocd generator.
- **flux:** `dependsOn` projects each component's declared `dependencyRefs` directly onto Flux's native DAG (a true DAG, not tiers), the most faithful and most parallel rendering.
- **helmfile:** `needs:` now chains only within a component (`-pre` → primary → `-post`); independent components in a sub-helmfile carry no edge and apply concurrently. Cross-tier ordering stays the sequential sub-helmfile layout (issue #914). Previously every release was positionally chained to its predecessor, which serialized independent components.
- **helm:** unchanged (already serial `deploy.sh`).
- **`--serial`** (`config.Serial`): reverts argocd / argocd-helm to a linear per-folder sync-wave, flux to the previous-component `dependsOn` chain, and helmfile to one linear `needs:` chain — reproducing the pre-parallelism ordering for rollback or bisecting a rollout. A no-op for helm.
- Real recipes already declare their dependencies (`base.yaml`: `gpu-operator` depends on `nfd`, `cert-manager`, `kube-prometheus-stack`; etc.), so genuine ordering is preserved. Only independent components now roll out concurrently.

## Testing

```bash
go test -race ./pkg/cli/... ./pkg/bundler/...     # ok (attestation pkg needs network/TUF, run outside sandbox)
golangci-lint run -c .golangci.yaml ./pkg/...     # 0 issues
chainsaw test --no-cluster --test-dir tests/chainsaw/cli/bundle-flux/       # PASS (6 roots parallel / 1 root serial)
chainsaw test --no-cluster --test-dir tests/chainsaw/cli/bundle-helmfile/   # PASS (intra-component needs / linear chain under --serial)
```

New Go tests cover the argocd wave arithmetic (`TestWaveForFolder`, `TestGenerate_LevelBasedSyncWaves`, `TestGenerate_SerialSyncWaves`), the flux DAG (`TestDeclaredDependsOn`, `TestSerialDependsOn`), the helmfile intra-component chaining (`TestGenerate_IndependentComponentsCollapse`, `TestGenerate_SerialChainsIndependentComponents`), and `TestConfigSerial`. argocd / argocd-helm goldens regenerated (sync-wave values only); flux and helmfile goldens unchanged. The flux and helmfile CLI e2e chainsaw tests now assert the parallel DAG by default and the linear chain under `--serial`.

## Risk Assessment

- [x] **Medium** — Changes the **default** deploy ordering for argocd / argocd-helm / flux / helmfile bundles (now parallel within a tier). Mitigated: declared dependencies preserve all gating; byte-compare goldens pin the output; `--serial` restores the exact prior behavior.

**Rollout notes:** The default flips to parallel for the four concurrent deployers. Operators who hit trouble can pass `--serial` to get the old one-at-a-time ordering back. Helm output is unchanged.

## Checklist


lockwobr:
@coderabbitai review

lockwobr:
@coderabbitai full review

pkg/bundler/deployer/argocd/argocd.go
```diff
@@ -251,6 +261,13 @@ type Generator struct {
 	// vendoring shape.
 	VendorCharts bool
 
+	// Serial assigns each folder a strictly increasing sync-wave equal to its
+	// linear position instead of the dependency-depth band (waveForFolder),
+	// so components deploy one at a time in deployment order. Off by default
+	// (parallel). Operator escape hatch wired from --serial; see
+	// config.Serial. argocdhelm forwards its own Serial here.
+	Serial bool
+
```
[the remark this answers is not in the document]
lockwobr: Declining this one for consistency. The deployer Generators (argocd, argocd-helm, flux, helmfile) are uniformly configured via exported struct fields set in struct literals by the bundler — `RecipeResult`, `VendorCharts`, `CascadeDelete`, `IncludeChecksums`, `AllowDynamicValueSplit`, `AppName`, etc. There are zero `With*` functional options in `pkg/bundler/deployer/`. Converting only `Serial` to a functional option would make it the lone outlier on these types. The functional-options guideline applies to the builder-style constructors (`recipe.NewBuilder`, `server.New`); matching the established Generator pattern (exported field) is the more consistent choice here.

pkg/bundler/deployer/flux/flux.go
```diff
@@ -294,13 +301,36 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 	// Add source file paths to resources list.
 	resources = append(resources, sourceResourcePaths(helmSources, gitSources)...)
 
+	// Fail closed on a bad dependency graph before projecting it onto Flux's
+	// dependsOn: Generate is exported, so a direct caller could hand-build a
+	// cyclic or dangling RecipeResult that would otherwise become a cyclic
+	// Flux DAG and stall reconciliation. Reuse the same acyclicity /
+	// missing-dependency check argocd and helmfile run; the levels themselves
+	// are discarded because flux renders the exact DAG, not tiers.
+	if _, levelErr := recipe.ComponentRefsTopologicalLevels(sortedRefs); levelErr != nil {
```
mchmarny: This validates the graph on `sortedRefs`, which is already enabled-filtered (`filterEnabled` above). That defeats the declared-but-disabled handling that `buildDependencyGraph`/`edgeSatisfiedExternally` were built for: a disabled component is absent from the filtered slice, so an enabled component's edge to it looks *undeclared* and this errors with a false "circular dependencies exist" — exactly the failure mode the pkg/recipe comment says the declared/enabled distinction exists to prevent. argocd (argocd.go:558) and helmfile both validate on the unfiltered `ComponentRefs`, so flux is the odd one out for the direct-caller case this check was added for (DefaultBundler pre-prunes these edges, so the production path is unaffected). It also makes the absent-dep drop branch in `declaredDependsOn` (line 673, unit-tested via the "absent-external-dep" case) unreachable through Generate. Suggest validating `g.RecipeResult.ComponentRefs` instead of `sortedRefs` — the graph builder does its own enabled-filtering and still rejects genuinely undeclared deps and cycles.

pkg/bundler/deployer/helmfile/releases.go
```diff
@@ -152,15 +159,23 @@ func buildHelmfile(folders []localformat.Folder, namespaceByComponent map[string
...
-		if prevName != "" {
+		// A needs: edge orders apply. Within a component (its <name>-pre ->
+		// primary -> <name>-post folders, always contiguous) the order is
+		// mandatory, so chain those regardless of mode. Across independent
+		// components that share a sub-helmfile, chain them only in --serial
+		// mode; by default they carry no edge and helmfile applies them
+		// concurrently. Cross-tier ordering does not rely on needs at all —
+		// the stratified layout runs level-N's sub-helmfile only after
+		// level-(N-1) has fully applied (see writeHelmfileLayout).
+		if prevName != "" && (serial || prevParent == f.Parent) {
```
mchmarny: nit: the intra-component chain relies on a component's folders being adjacent in `folders` — if localformat (or `splitFoldersByLevel`) ever interleaved another component's folder between a primary and its `-post`, the `prevParent == f.Parent` check would silently drop the mandatory intra-component edge and the `-post` would apply concurrently with its primary. The contiguity invariant holds today and the comment documents it, but tracking the last release per parent (`lastByParent[f.Parent]`) instead of the single `prevName/prevParent` pair would make the edge independent of folder adjacency at no extra cost.

pkg/bundler/config/config.go
```diff
@@ -794,6 +812,16 @@ func WithReadinessHooks(enabled bool) Option {
 	}
 }
 
+// WithSerial forces deployers to sequence components strictly one at a time in
+// deployment order, disabling the per-tier / declared-dependency concurrency
+// the argocd, argocd-helm, flux, and helmfile deployers emit by default. Off
+// by default; opt-in via --serial. A rollback / debugging escape hatch.
+func WithSerial(enabled bool) Option {
```
mchmarny, line 839: nit: `serial` is CLI-only — POST /v1/bundle exposes `vendor-charts` and `app-name` as query parameters (pkg/bundler/handler.go) but not `serial`, so aicrd callers can't reach this escape hatch. This matches the `readiness-hooks` precedent of CLI-only bundle options, so likely intentional; flagging in case API parity is wanted (would need handler.go, api/aicr/v1/server.yaml, and docs/user/api-reference.md in a follow-up).

mchmarny (dismissed):
Well-executed change: parallelism is derived from declared `dependencyRefs` and projected onto each deployer's native mechanism (flux exact DAG via `dependsOn`, argocd tier-banded sync-waves with stride-4 phase offsets, helmfile intra-component `needs` within sequential sub-helmfiles), with a `--serial` escape hatch reproducing the prior linear ordering. Importantly, no Go-level concurrency is introduced — generation stays sequential and deterministic (sorted tiers/targets), so AICR remains an output adapter rather than a deployment engine. Test coverage is strong: cycle rejection in both modes, wave arithmetic pinned, DAG/serial chainsaw asserts, and goldens byte-pin the output. CI is fully green including the entire Tier-1 deployer matrix.

One medium inconsistency inline on the flux graph-validation placement (fail-closed direction, direct-caller-only) plus two nits; nothing blocking.

pkg/bundler/deployer/flux/flux.go
lockwobr: Good catch — fixed in 8882cf62. Validation now runs on the unfiltered `g.RecipeResult.ComponentRefs` (not the enabled-filtered `sortedRefs`), so `ComponentRefsTopologicalLevels` can treat an edge to a declared-but-disabled component as satisfied externally instead of mistaking it for an undeclared dep. Added `TestGenerate_DisabledDependencyNotACycle` (fails on the old code, passes now).

pkg/bundler/deployer/helmfile/releases.go
lockwobr: Done in 8882cf62 — the intra-component `needs:` edge is now keyed on `lastByParent[f.Parent]` (last release emitted per component) rather than the single `prevParent` pair, so the `-pre → primary → -post` ordering holds even if folder order ever interleaved another component. Output is unchanged for the contiguous case (goldens unaffected).

pkg/bundler/config/config.go
lockwobr: Added API parity in 69faaaa4: `serial` is now a POST /v1/bundle query param, wired `parseQueryParams → config.WithSerial` alongside `vendor-charts`/`app-name`. Updated the OpenAPI contract (`api/aicr/v1/server.yaml`) and `docs/user/api-reference.md`, and added `TestParseBundleConfig_Serial`.

pkg/bundler/deployer/flux/flux.go
```diff
@@ -594,19 +627,84 @@ func (g *Generator) detectInjectedReleaseCollisions(sortedRefs []recipe.Componen
...
+	out := make(map[string][]string, len(sortedRefs))
+	for _, c := range sortedRefs {
+		if len(c.DependencyRefs) == 0 {
+			continue
+		}
+		targets := make([]string, 0, len(c.DependencyRefs))
+		for _, dep := range c.DependencyRefs {
+			depRef, ok := refByName[dep]
+			if !ok {
+				// Disabled, externally provided, or undeclared: no release to
```
mchmarny, line 678: Dropping the edge to a disabled component also drops *transitive* ordering: with A → B → C and B disabled, A ends up with no `dependsOn` at all and reconciles concurrently with C. Before this PR the linear chain still installed one component at a time, so a latent mis-order like this was masked; now it can surface as a race. The "satisfied externally" semantic is pre-existing (`edgeSatisfiedExternally` in `pkg/recipe`) and arguably correct — an external B carries its own dependency on C — but the same collapse happens in the argocd/helmfile level computation (A falls to tier 0 beside C). Worth a sentence in the component.md ordering section ("disabling a mid-graph component removes its dependents' transitive ordering; use `--serial` if unsure"), or contracting edges through disabled nodes if you want the stronger guarantee. Not blocking.

mchmarny (approved):
Careful, complete work. The "one model, three projections" framing (exact DAG for flux, tier bands for argocd, sequential sub-helmfiles for helmfile) is the right decomposition, and the details hold up: unconditional cycle validation on both paths, component-keyed intra-component chaining in helmfile, a chainsaw DAG check with a real acyclicity pass, a test helper that fails loudly instead of passing vacuously on stratified layouts, and full doc/API-spec parity for the new flag. `--serial` reproducing the exact pre-parallelism ordering is the right escape hatch for a default-behavior flip.

One non-blocking question inline about transitive ordering through disabled components. Please squash the 3 commits before merge per repo policy.

PR #1753, opened 2026-07-14
ci(notices): gate THIRD_PARTY_NOTICES.md freshness cross-platform

lockwobr:
## Summary

Regenerate `THIRD_PARTY_NOTICES.md` as the union of dependency licenses across every released OS/arch target, and add a `notices-freshness` merge-gate job so the committed file can no longer silently drift when dependencies change.

## Motivation / Context

The committed `THIRD_PARTY_NOTICES.md` was badly out of date (noticed in #1729). Root cause: the generator ran `go-licenses` for a single host platform. `go-licenses` resolves the dependency graph for the host `GOOS/GOARCH` only, and `aicr` ships on darwin + linux (× amd64/arm64) with build-tagged sources (`*_linux.go` vs `*_darwin.go`) that import different transitive deps per platform. A macOS-generated file therefore omits linux-only deps (and vice versa) and is not reproducible between a dev machine and CI, so no freshness gate could hold. This is the "linux vs mac" issue the previous attempt hit.

Fixes: N/A
Related: #1729

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: build tooling (`tools/generate-notices`, `Makefile`), CI (`.github/workflows/merge-gate.yaml`)

## Implementation Notes

- **`tools/generate-notices`** — now generates the **union** across `{linux,darwin} × {amd64,arm64}`: re-runs `go-licenses save`/`csv` once per target (CGO disabled, so no C toolchain needed), merges per-platform license trees, recomputes the GOOS-dependent stdlib ignore set per target, and sorts the index with `LC_ALL=C` (BSD vs GNU `sort` collation was a second drift source). Output is complete and byte-for-byte reproducible on any host. The header documents why the upstream tool is not used directly (neither `go-licenses` nor goreleaser has a multi-GOOS union mode). Dead `VERSION` code removed.
- **`Makefile`** — adds `make notices-check` (regenerate + `git diff --quiet`), mirroring the proven `tuning-check`; `make notices` now covers all platforms.
- **`merge-gate.yaml`** — adds a `notices` path filter (`go.mod`, `go.sum`, `vendor/**`, `tools/generate-notices`, `THIRD_PARTY_NOTICES.md`) and a `notices-freshness` / `-skip` job pair wired into the aggregate `gate`. Runs only when dependencies change, so `make qualify` stays fast.
- **`THIRD_PARTY_NOTICES.md`** — regenerated: 141 → 188 dependencies (the full union).

## Testing

```bash
make notices          # 141 -> 188 deps
make notices-check     # exit 0, "THIRD_PARTY_NOTICES.md is up to date"
```

- Verified **byte-for-byte reproducible**: two independent runs produce identical output.
- `yamllint`, `shellcheck`, and `actionlint` clean on the changed files.
- No Go source changed, so the Go test/coverage gates are not applicable.

## Risk Assessment

- [x] **Low** — Isolated tooling + a regenerated artifact + a new, path-filtered gate; easy to revert.

**Rollout notes:** Overlaps #1729 on `THIRD_PARTY_NOTICES.md` (both regenerate it) — a mechanical conflict resolved by re-running `make notices`, not a hand-merge. If this lands first, #1729's dependency changes will trip the new gate and force a correct regeneration.

## Checklist

- [x] Linter passes (`make lint` inputs: yamllint/shellcheck/actionlint clean)
- [x] I updated docs if user-facing behavior changed (`docs/contributor/tests.md`, `RELEASING.md`)
- [x] Changes follow existing patterns in the codebase (mirrors `bom-freshness` / `tuning-check`)


THIRD_PARTY_NOTICES.md
```diff
@@ -5250,28 +4906,68 @@ THE SOFTWARE.
...
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
+in the documentation and/or other materials provided with the
+distribution.
+   * Neither the name of Google Inc. nor the names of its
+contributors may be used to endorse or promote products derived from
+this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYIGHT HOLDERS AND CONTRIBUTORS
```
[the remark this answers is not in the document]
lockwobr: This is from upstream, we should not make changes to these. Closing this.

.github/workflows/merge-gate.yaml
```diff
@@ -106,6 +107,19 @@ jobs:
...
+              # across the released OS/arch matrix, so any dependency-graph
+              # change can alter it. Watch the dependency manifests, the vendor
+              # tree, and the generator itself (plus the committed file, so a
+              # hand edit is re-verified). This is a standalone job — the full
+              # `tests` suite does NOT regenerate the notices file.
+              - 'go.mod'
+              - 'go.sum'
+              - 'vendor/**'
+              - 'tools/generate-notices'
+              - 'THIRD_PARTY_NOTICES.md'
```
lockwobr, line 134: Good call — addressed in 056b7494, and taken a step further. `.goreleaser.yaml` is now in the `notices` path filter, **and** `tools/generate-notices` asserts its `PLATFORMS` matrix equals the cross-product of every build's `goos × goarch` in `.goreleaser.yaml` (via `yq`), failing the run on drift. Without that assertion the filter alone would be a no-op — the matrix is hardcoded, so a goreleaser-only change would regenerate an identical file and pass. Now a new release target fails `make notices` until the matrix is updated.

THIRD_PARTY_NOTICES.md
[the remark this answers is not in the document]
lockwobr: This is intentional and must stay as-is. `COPYIGHT` is the **verbatim** text of the upstream license file, present in `vendor/github.com/aws/aws-sdk-go-v2/internal/sync/singleflight/LICENSE` and `vendor/github.com/aws/smithy-go/internal/sync/singleflight/LICENSE` (both vendor Go's `singleflight`, whose LICENSE carries this long-standing typo). `THIRD_PARTY_NOTICES.md` is a byte-for-byte reproduction of each dependency's license, so "correcting" it would misrepresent the license **and** immediately fail the new `notices-freshness` gate, since `make notices` regenerates it straight from vendor. Leaving unchanged.

tools/generate-notices
```diff
@@ -49,49 +65,89 @@ case "${LICENSES_DIR}" in
...
-INDEX_FILE="$(mktemp)"
-trap 'rm -f "${INDEX_FILE}"' EXIT
-go-licenses csv "${PACKAGES[@]}" \
-    --ignore="${IGNORE}" \
-    | sort -t, -k1,1 -u \
-    > "${INDEX_FILE}"
+# Deduplicate by package (first CSV field) and sort deterministically. LC_ALL=C
+# pins byte-ordering so macOS (BSD sort) and Linux (GNU sort) agree — locale
+# collation is the classic source of cross-platform diff noise.
+LC_ALL=C sort -t, -k1,1 -u "${COMBINED_CSV}" > "${INDEX_FILE}"
```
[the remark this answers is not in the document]
lockwobr: Verified this is not an issue. `sort -z` is supported by stock macOS `/usr/bin/sort` (`2.3-Apple (199)`, BSD) — tested `printf 'b\0a\0' | LC_ALL=C sort -z` returns `a\0b\0` correctly, and `notices-freshness` passes on Linux CI, so the output is identical on both. Note the `-z` sort applies to the `find -print0` license-file pipeline (filenames with spaces), not the CSV index sort, which is newline-delimited `LC_ALL=C sort -t, -k1,1 -u`. No portability change needed.

.github/workflows/merge-gate.yaml
```diff
@@ -106,6 +107,23 @@ jobs:
...
+              # the generator asserts its PLATFORMS matrix matches the release
+              # targets defined there — a new goos/goarch must re-run this check.
+              # This is a standalone job — the full `tests` suite does NOT
+              # regenerate the notices file.
+              - '.goreleaser.yaml'
+              - 'go.mod'
+              - 'go.sum'
+              - 'vendor/**'
+              - 'tools/generate-notices'
+              - 'THIRD_PARTY_NOTICES.md'
```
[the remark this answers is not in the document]
lockwobr: Agreed — applied in 48bbbf77. `Makefile` is now in the `notices` filter so a change to the `notices`/`notices-check` recipes re-runs `notices-freshness` rather than hitting the skip job. Matches the existing `renovate` filter, which also watches `Makefile` broadly.

Makefile
```diff
@@ -577,6 +577,17 @@ validate-local: image-validators ## Builds validator images and runs validation
...
+.PHONY: notices-check
+notices-check: ## Verifies THIRD_PARTY_NOTICES.md is up to date (run by the merge-gate on dependency changes)
+	@set -e; \
+	$(MAKE) notices; \
+	if ! git diff --quiet -- THIRD_PARTY_NOTICES.md; then \
+	   echo "ERROR: THIRD_PARTY_NOTICES.md is stale. Run 'make notices' and commit the change." >&2; \
+	   git --no-pager diff --stat -- THIRD_PARTY_NOTICES.md >&2; \
+	   exit 1; \
+	fi; \
+	echo "THIRD_PARTY_NOTICES.md is up to date"
```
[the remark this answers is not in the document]
lockwobr: Leaving as-is. `checkmake` is not part of this repo's lint gate (no reference in the Makefile, `.github/`, or `.settings.yaml`; `tests / Lint` is green), so this is a CodeRabbit-tool nit rather than a CI failure. More importantly, `notices-check` is deliberately byte-for-byte the same shape as the existing `tuning-check` target (same `@set -e; \ ... if ! git diff --quiet ...` block) — extracting only this one into a helper script would make the two freshness targets gratuitously inconsistent for no functional gain. Prefer to keep them parallel.

ayuskauskas (dismissed):

.github/workflows/merge-gate.yaml
```diff
@@ -106,6 +107,26 @@ jobs:
...
+              # recipes are the generation/check contract — a change to them
+              # must not be able to skip its own gate. This is a standalone
+              # job — the full `tests` suite does NOT regenerate the file.
+              - 'Makefile'
+              - '.goreleaser.yaml'
+              - 'go.mod'
+              - 'go.sum'
+              - 'vendor/**'
+              - 'tools/generate-notices'
+              - 'THIRD_PARTY_NOTICES.md'
```
mchmarny, line 134: `.settings.yaml` is missing from this filter. The generator's output depends on two pins that live there: `linting.go_licenses` (the CI job installs go-licenses at that version, and a version bump can change license classification/URL derivation) and `languages.go` (changes the `go list std` ignore set). A toolchain-bump PR touching only `.settings.yaml` would skip this gate and could leave the committed file stale — the next unrelated dependency PR then eats a confusing freshness failure it didn't cause. Add `.settings.yaml` to the list (same rationale as including `Makefile`: inputs to the generation contract).

Makefile
```diff
@@ -577,6 +577,17 @@ validate-local: image-validators ## Builds validator images and runs validation
 notices: ## Generates THIRD_PARTY_NOTICES.md aggregating every dependency's license
 	@bash tools/generate-notices
 
+.PHONY: notices-check
+notices-check: ## Verifies THIRD_PARTY_NOTICES.md is up to date (run by the merge-gate on dependency changes)
```
mchmarny, line 582: The PR mirrors `tuning-check`, but `tuning-check` is wired into `make qualify` and this isn't — so a dep-bump PR passes `make qualify` locally, then fails the merge gate, breaking the "qualify passes locally → CI passes" invariant from AGENTS.md. The speed tradeoff is fair; a cheap middle ground: run `make notices` at the end of `make tidy`, so the file regenerates exactly when deps change locally and there's nothing to forget. Non-blocking either way — `docs/contributor/tests.md` documents the pitfall.

tools/generate-notices
```diff
@@ -36,11 +53,17 @@ if ! command -v go >/dev/null 2>&1; then
     exit 1
 fi
 
+if ! command -v yq >/dev/null 2>&1; then
+    echo "ERROR: yq is not installed (needed to verify the release-target matrix)." >&2
+    echo "Install it with 'make tools-setup'." >&2
```
mchmarny, line 58: nit: while here — the go-licenses hint ~25 lines up still says `go install github.com/google/go-licenses@$(yq -r .linting.go_licenses .settings.yaml)`. With the pin at v2.0.1 that path fails (`go-licenses/v2` is required, as the new CI job correctly uses). Pre-existing, but this PR makes the pinned tool version load-bearing for reproducibility, so worth fixing the hint.

mchmarny (commented):
Solid fix — the per-target union is the right answer to the linux-vs-mac drift, the generator fails closed everywhere it matters (goreleaser-matrix guard, empty-index guard, `set -e` through the make recipe), and the gate follows the established real/skip pattern; nice that `notices-freshness` already validated itself on this PR. One medium-low gap (`.settings.yaml` missing from the path filter), one non-blocking suggestion on local/CI parity, and one nit. Nothing blocks merge.

.github/workflows/merge-gate.yaml
lockwobr: Good catch — applied in c0ac7380. `.settings.yaml` is now in the `notices` filter, with a comment noting both pins that feed the generator: `linting.go_licenses` (classification/URL derivation) and `languages.go` (the `go list std` ignore set). A toolchain-only bump now re-runs the gate instead of deferring a confusing stale-file failure to the next dependency PR.

tools/generate-notices
lockwobr: Fixed in c0ac7380 — the hint now reads `go install github.com/google/go-licenses/v2@$(yq -r .linting.go_licenses .settings.yaml)`, matching the v2.0.1 pin and the CI job. Thanks for flagging; agreed it matters more now that the pinned version is load-bearing for reproducibility.

Makefile
lockwobr: Took the middle ground you suggested — `make tidy` now runs `make notices` as its final step (right after `go mod vendor`) in ecd8116a. So the file regenerates exactly when deps change locally and there's nothing to forget, without adding the ~1-2min cross-platform pass to every `make qualify`. Updated the `docs/contributor/tests.md` gotcha to match (hand-editing go.mod/vendor still needs a manual `make notices`). Keeping `notices-check` out of `qualify` itself per the earlier speed call, but this closes the "qualify passes locally → merge gate fails" gap for the normal dependency-update flow.

mchmarny (approved):

PR #1754, opened 2026-07-14
feat(ci): own Rekor v2 identity monitoring via tools/rekor-monitor

### LINKED ISSUE #1623: feat(ci): Rekor v2 identity monitoring for the release signer
## Problem

Our transparency-log monitor (`.github/workflows/rekor-monitor.yaml`) has
**never completed an identity scan since it was merged** (#1480, 2026-06-26).
The `monitor` job hits the upstream reusable workflow's hard-coded
`timeout-minutes: 60`, is SIGKILLed mid-scan, and never advances its checkpoint.

Timeline (all verifiable from run history + checkpoint artifacts):

- **2026-06-26**: merged; then `startup_failure` on every scheduled run for ~4
  days — it never executed.
- **2026-06-30 19:59**: first run that ran, finished in **1.4 min** via the
  no-prev-checkpoint "just save head, skip the scan" path, saving tree size
  **1,904,155,885** as the initial checkpoint.
- **Every run since** (including the GHA-green "successes" on 07-01/02/05) left
  the checkpoint at **1,904,155,885**. The green ones took the consistency-check
  early-return path (`once=true` returns exit 0 without scanning or
  checkpointing on any consistency hiccup); the rest were cancelled at the 60-min
  cap. **Net identity-scan coverage since day one: zero.**

Current state: checkpoint still frozen at **1,904,155,885** while the
public-good head is **~1,969,169,705** — we are **~65M entries behind** and the
gap grows ~1–2M/hour. It cannot recover on its own.

### Root cause

We call `sigstore/rekor-monitor` with a `config` containing
`monitoredValues.certIdentities` (our release signer,
`on-tag.yaml@refs/tags/*`). That switches on an **identity scan**: a linear walk
of *every* entry from the last checkpoint to head — i.e. the entire public-good
Rekor firehose (~500–750 entries/sec added globally), not just our entries.
Upstream's own `check.yml` and most consumers pass **no config**, so they only
run the O(log n) consistency proof and never hit this. Our scan rate ≈ the log's
global write rate, so we sat at break-even and finally tipped over permanently.

### Why the reusable workflow can't fix it

- **No timeout knob** — `base_monitoring.yml` bakes in `timeout-minutes: 60`.
- **All-or-nothing checkpoint** — `internal/cmd/common.go` `MonitorLoop` calls
  `WriteCheckpoint` *only after* `IdentitySearch` fully completes ("to ensure
  identities are always searched even if something fails in the middle"). A
  timeout mid-scan discards all progress → next run rescans a bigger window →
  death spiral.
- **Can't parallelize instances** — the `detect-workflow` job pins
  `concurrency: group: rekor-consistency-check, cancel-in-progress: true`, so a
  second invocation cancels the first.
- **Checkpoint always persists head** — `WriteCheckpoint(prev, cur)` writes
  `curCheckpoint` (head at run start), not the scanned `endIndex`, so bounded
  windows corrupt the built-in cursor.

The identity-scan feature simply does not fit in a 60-min GitHub-hosted job at
today's Rekor throughput, and the reusable workflow exposes no lever to make it.

## Proposal

**Own the monitoring workflow** (vendor it under `.github/workflows/`, pinned to
an upstream SHA we track) and replace the single unbounded scan with **parallel
paged identity scanning** so we can keep up with — and catch up to — the live
log head.

Mechanics we can build on (already supported by the CLI, not the reusable
workflow): `startIndex` / `endIndex` are settable via the `--config` YAML
(`pkg/notifications/notifications.go`), so the scan window is fully addressable.

Design sketch:

1. **Shard by index range.** Split `[cursor, head]` into fixed-size pages and
   fan them out with a `strategy.matrix` (one job per page), each invoked with
   `once: true` and explicit `startIndex`/`endIndex`. Throttle `max-parallel` to
   stay a good citizen against the shared public-good instance (it 429s under
   load).
2. **Size the shard to a time budget.** Pick a page size that reliably scans in
   *X* minutes (well under the job timeout) at observed throughput, so every
   shard always completes and reports. Make page size + `max-parallel` the two
   tunables; document how to re-tune as Rekor throughput grows.
3. **Own the cursor.** Persist our own scanned-through index (the built-in
   checkpoint writes head and can't express partial progress). Advance the
   cursor only over ranges that actually completed; aggregate matches across
   shards and file one issue on any hit or failed entry.
4. **Split consistency from identity.** Keep the cheap append-only consistency
   proof on its own frequent cadence (this is the high-value tamper check and
   always finishes); run the paged identity scan on its own track so a slow
   identity sweep never blocks consistency monitoring.
5. **One-time backlog catch-up.** Before steady state, run a bounded parallel
   sweep across the ~65M-entry gap to clear it, then hand a fresh cursor to the
   steady-state job.

### Scaling note

Firehose identity scanning is O(all of Sigstore), so parallel paging is an arms
race against total ecosystem write volume: it buys headroom (2× parallel already
outpaces current growth) but the margin shrinks over time. This issue delivers
the scalable paging solution and the tunables to stay ahead; a follow-up should
evaluate a fundamentally cheaper posture (query-by-identity, O(our entries), or
release-time identity verification) so we are not brute-scanning the planet's
traffic forever to find our own release entries.

## Acceptance

- Vendored, SHA-pinned monitoring workflow under our control.
- Parallel paged identity scan with page-size + concurrency tunables; each shard
  completes within its time budget and always records progress.
- Consistency proof runs independently on a frequent cadence.
- Backlog cleared; steady-state runs track head without spiraling.
- Docs: how the paging works and how to re-tune as throughput grows.

Part of #1149.

lockwobr:
## Summary

Own Rekor v2 identity + consistency monitoring with a small tool (`tools/rekor-monitor`), replacing the merged-but-broken reusable-workflow approach from #1727 that fails on `main` every hour.

## Motivation / Context

The upstream `sigstore/rekor-monitor` reusable workflow resolves its Rekor API version and shard set from Sigstore's **default** signing config (`signing_config.v0.2.json`), which lists only Rekor v1 and, per [rekor-evolution](https://blog.sigstore.dev/rekor-evolution/), keeps v1 as the default "for the foreseeable future". AICR opted into Rekor v2 **early** (#1650) via the separate `signing_config_rekor_v2.v0.2.json` TUF target, which upstream never reads and exposes no flag to select. So pointing it at a v2 shard URL falls through to v1, and its v1 client times out against the v2 tile endpoint — every scheduled run on `main` fails (`error getting log verifier: … context deadline exceeded`).

This tool changes only the AICR-specific bit: it reads the v2 signing config AICR actually signs against (`pkg/trust`) and reuses the upstream rekor-monitor **library** packages (`pkg/rekor/v2`, `pkg/tiles`, `pkg/identity`) for the security-critical verification, so we do not reimplement transparency-log crypto. When Sigstore makes v2 the ecosystem default, the reusable workflow can monitor v2 directly and this tool can be retired.

Fixes: #1623
Related: #1650, #1727, #1149

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tools/rekor-monitor` (new), `.github/workflows/rekor-monitor.yaml`, `go.mod`/`vendor`

## Implementation Notes

- **`tools/rekor-monitor`** (structured across files): `doc.go` (rationale + structure, surfaced by `go doc`), `main.go` (`run()` orchestration + exit codes), `monitor.go` (`monitor` = resolved v2 shards + watched identity, its two checks, and the `outcome` of a pass), `checkpoint.go` (`checkpointStore`: restore the cursor from the fetched artifact zip via native `archive/zip`, read, write). Unit tests per file + a `README.md`.
- **Consistency** anchors the incremental identity scan (proves nothing was removed/rewritten between runs, so scanning only the new window is sound) and is the standard append-only check. **Identity** is the AICR-specific check for a rogue entry under the release signer.
- **Workflow**: runs the tool; fetches the prior checkpoint artifact via the REST API (cross-run) and hands it to the tool to restore natively; opens/closes a deduplicated alert issue on failure/success. Fresh artifact name (`rekor-v2-checkpoint`) so the stale v1 checkpoint is ignored — no migration.
- **Deps**: adds `sigstore/rekor-monitor` (SHA-pinned, no upstream releases) + `transparency-dev/formats`; built `MonitoredValues` directly to avoid pulling the go-github/SMTP notification deps.

## Testing

```bash
GOFLAGS=-mod=vendor go test -race ./tools/rekor-monitor/...   # pass
golangci-lint run -c .golangci.yaml ./tools/rekor-monitor/... # 0 issues
make license-check                                            # pass (new deps allowed)
actionlint / yamllint .github/workflows/rekor-monitor.yaml    # clean
```

End-to-end against live Rekor v2 (`log2025-1`): first run baselines at the current head and skips the scan; a subsequent run proves consistency forward and scans the new identity window (no match, exit 0); a catch-all subject matches real entries and exits non-zero (the alert path). Restore-from-zip round trip verified (a failed restore fails loudly instead of silently re-baselining). Full-suite coverage total 78.0% (≥ 75% floor).

## Risk Assessment

- [x] **Low** — New isolated tool + one CI workflow + docs; easy to revert. The monitor is read-only against the public log and never writes to release artifacts.

**Rollout notes:** On merge the hourly workflow starts using the tool; the first run baselines (no scan) and every run after scans the new window. No manual migration (new artifact name). Verifiers are unaffected — this only changes monitoring, not signing.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — tool package tested with `-race`


.github/workflows/rekor-monitor.yaml
```diff
@@ -73,56 +71,107 @@ concurrency:
...
-          echo "Resolved Rekor v2 shard: $url"
-          echo "url=$url" >> "$GITHUB_OUTPUT"
+          id="$(gh api "repos/${GITHUB_REPOSITORY}/actions/artifacts?name=${ARTIFACT_NAME}&per_page=100" \
+            --paginate --jq '[.artifacts[] | select(.expired == false)] | sort_by(.created_at) | last | .id // empty')"
+          if [ -z "$id" ]; then
+            echo "No prior checkpoint artifact found; treating this as the first run."
+            exit 0
+          fi
+          gh api "repos/${GITHUB_REPOSITORY}/actions/artifacts/${id}/zip" > checkpoint.zip
+          echo "Fetched checkpoint artifact ${id}."
```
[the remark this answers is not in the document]
lockwobr: Fixed in 4d14d1b9: added `actions: read` to the job permissions (the artifact fetch via the Actions API needs it), plus a `timeout-minutes: 20` backstop.

.github/workflows/rekor-monitor.yaml
```diff
@@ -73,56 +71,107 @@ concurrency:
...
         run: |
           set -euo pipefail
-          go run ./cmd/aicr trust update --emit-signing-config signing-config.json
-          # Newest currently-listed v2 shard (majorApiVersion 2). Any live v2
-          # shard selects v2; the monitor then discovers the full set from TUF.
-          url="$(jq -er '[.rekorTlogUrls[] | select(.majorApiVersion == 2)] | sort_by(.validFor.start) | last | .url' signing-config.json)"
-          echo "Resolved Rekor v2 shard: $url"
-          echo "url=$url" >> "$GITHUB_OUTPUT"
+          id="$(gh api "repos/${GITHUB_REPOSITORY}/actions/artifacts?name=${ARTIFACT_NAME}&per_page=100" \
+            --paginate --jq '[.artifacts[] | select(.expired == false)] | sort_by(.created_at) | last | .id // empty')"
```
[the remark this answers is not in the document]
lockwobr: Fixed in 4d14d1b9: switched to `gh api --paginate ... | jq -rs '[.[].artifacts[] | select(.expired==false)] | sort_by(.created_at) | last | .id'` so the reduction happens across all pages in one jq instead of per-page.

tools/rekor-monitor/checkpoint_test.go
```diff
@@ -0,0 +1,185 @@
...
+	t.Run("corrupt zip errors", func(t *testing.T) {
+		zp := filepath.Join(dir, "corrupt.zip")
+		if err := os.WriteFile(zp, []byte("not a zip"), 0o600); err != nil {
+			t.Fatalf("write: %v", err)
+		}
+		if err := extractCheckpointFromZip(zp, dest); err == nil {
+			t.Error("expected error for corrupt zip")
+		}
+	})
+}
```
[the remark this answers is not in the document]
lockwobr: Fixed in 4d14d1b9: `TestExtractCheckpointFromZip` is now table-driven and includes an oversized-entry case (`strings.Repeat("x", maxCheckpointBytes+1)`) that exercises the size guard.

tools/rekor-monitor/main.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		slog.Error("invalid arguments", "error", err)
+		os.Exit(2)
+	}
+	if err := run(context.Background(), opts, os.Stdout); err != nil {
+		slog.Error("rekor v2 monitor detected an issue or failed", "error", err)
+		os.Exit(1)
+	}
+	slog.Info("rekor v2 monitor completed cleanly")
+}
```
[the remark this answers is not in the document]
lockwobr: Fixed in 4d14d1b9: `main` now wraps the pass in `context.WithTimeout` (new `--timeout` flag, default 15m) and uses a `realMain() int` so `defer cancel()` runs (also satisfies gocritic's exit-after-defer).

tools/rekor-monitor/checkpoint_test.go
```diff
@@ -0,0 +1,185 @@
...
+	t.Run("malformed file errors", func(t *testing.T) {
+		p := filepath.Join(dir, "bad.txt")
+		if err := os.WriteFile(p, []byte("not a checkpoint"), 0o600); err != nil {
+			t.Fatalf("write: %v", err)
+		}
+		if _, err := (checkpointStore{path: p}).read(); err == nil {
+			t.Error("expected error for malformed checkpoint, got nil")
+		}
+	})
+}
```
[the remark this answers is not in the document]
lockwobr: Partially addressed: converted `TestExtractCheckpointFromZip` to table form. Leaving `TestCheckpointStoreRead`/`Restore` as `t.Run` subtests on purpose: each case needs distinct filesystem setup (missing file, empty file, malformed file, absent zip, present zip), which reads more clearly as discrete subtests than a table of setup closures. Leaving this thread open for maintainer preference.

tools/rekor-monitor/monitor_test.go
```diff
@@ -0,0 +1,159 @@
...
+				if tt.certIssuer == "" && len(cid.Issuers) != 0 {
+					t.Errorf("Issuers = %v, want empty", cid.Issuers)
+				}
+				if tt.certIssuer != "" && (len(cid.Issuers) != 1 || cid.Issuers[0] != tt.certIssuer) {
+					t.Errorf("Issuers = %v, want [%q]", cid.Issuers, tt.certIssuer)
+				}
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
lockwobr: Declining this one: the `certID.Verify()` error branch is unreachable through `buildMonitoredValues`. An empty subject returns early, and `Issuers` only ever holds a non-empty value, so `Verify()` cannot fail here; it also doesn't validate the regex (that's compiled later during the scan). A test would have to call `Verify()` on hand-built input, which wouldn't exercise this function. Keeping the check as defense-in-depth. Leaving open for a maintainer view.

tools/rekor-monitor/monitor_test.go
```diff
@@ -0,0 +1,169 @@
...
+		wantStart int64
+		wantEnd   int64
+		wantOK    bool
+	}{
+		{name: "first run (nil prev)", prev: nil, cur: cp(100), wantOK: false},
+		{name: "nil cur", prev: cp(100), cur: nil, wantOK: false},
+		{name: "zero prev size", prev: cp(0), cur: cp(100), wantOK: false},
+		{name: "normal window", prev: cp(100), cur: cp(150), wantStart: 99, wantEnd: 149, wantOK: true},
+		{name: "no growth (start==end)", prev: cp(100), cur: cp(100), wantStart: 99, wantEnd: 99, wantOK: true},
+		{name: "shrunk (start>end)", prev: cp(150), cur: cp(100), wantOK: false},
```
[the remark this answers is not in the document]
lockwobr: Fixed in e13b294e: `scanWindow` now returns `[prev.Size, cur.Size-1]` (first unseen entry through the current head) and returns `ok=false` when the log hasn't grown, so it no longer re-scans the boundary/pre-baseline entry. Tests updated.

.github/workflows/rekor-monitor.yaml
```diff
@@ -73,56 +74,156 @@ concurrency:
...
+        id: alert_issue
+        if: ${{ failure() }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          # Reuse an already-open alert issue (dedup) and surface its URL so the
+          # Slack step can link to it.
+          existing="$(gh issue list --repo "${GITHUB_REPOSITORY}" --state open \
+            --search "in:title \"${ALERT_TITLE}\"" --json url --jq '.[0].url // empty')"
```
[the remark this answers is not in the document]
lockwobr: Fixed in e13b294e: both the open-dedup and close steps now filter to an exact title match (`jq --arg t ... select(.title == $t)`) instead of relying on phrase search.

tools/rekor-monitor/main.go
```diff
@@ -0,0 +1,167 @@
...
+	fs.StringVar(&opts.userAgent, "user-agent", "aicr-rekor-v2-monitor", "User-Agent for requests to the log")
+	fs.DurationVar(&opts.timeout, "timeout", defaultTimeout, "maximum duration for the whole monitor pass")
+	fs.StringVar(&opts.restoreZip, "restore-zip", "", "path to a GitHub-artifact zip to extract the prior checkpoint from before monitoring (missing file = first run)")
+	if err := fs.Parse(args); err != nil {
+		return options{}, errors.Wrap(errors.ErrCodeInvalidRequest, "failed to parse flags", err)
+	}
+	if opts.certSubject == "" && opts.certIssuer != "" {
+		return options{}, errors.New(errors.ErrCodeInvalidRequest, "--cert-issuer requires --cert-subject")
+	}
+	return opts, nil
```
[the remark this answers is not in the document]
lockwobr: Fixed in e13b294e: `parseFlags` now rejects `--timeout <= 0` with `ErrCodeInvalidRequest` (+ a flag test).

tools/rekor-monitor/checkpoint.go
```diff
@@ -0,0 +1,159 @@
...
+		return nil
+	}
+	if _, err := os.Stat(s.restoreZip); err != nil {
+		if stderrors.Is(err, os.ErrNotExist) {
+			return nil // no prior artifact fetched: first run
+		}
+		return errors.Wrap(errors.ErrCodeInternal, "failed to stat checkpoint artifact zip", err)
+	}
+	return extractCheckpointFromZip(s.restoreZip, s.path)
+}
```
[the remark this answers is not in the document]
lockwobr: Fixed in e13b294e: threaded `ctx` through `restore` -> `extractCheckpointFromZip` -> `selectCheckpointEntry`, which checks `ctx.Err()` while walking the member list. Added a canceled-context test.

tools/rekor-monitor/monitor_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+func TestMonitorWatchesIdentity(t *testing.T) {
+	if (&monitor{}).watchesIdentity() {
+		t.Error("empty monitor should not watch identity")
+	}
+	m := &monitor{identities: identity.MonitoredValues{identity.CertIdentityValue{CertSubject: "^x$"}}}
+	if !m.watchesIdentity() {
+		t.Error("monitor with an identity should watch identity")
+	}
+}
```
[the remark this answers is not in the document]
lockwobr: Fixed in e13b294e: `TestMonitorWatchesIdentity` is now table-driven.

tools/rekor-monitor/monitor.go
```diff
@@ -0,0 +1,184 @@
...
+	return found, failed, nil
+}
+
+// buildMonitoredValues converts the identity flags into rekor-monitor's
+// MonitoredValues. It returns the zero value (no identities) when certSubject is
+// empty, which the caller treats as consistency-only.
+func buildMonitoredValues(certSubject, certIssuer string) (identity.MonitoredValues, error) {
+	if certSubject == "" {
+		return identity.MonitoredValues{}, nil
+	}
```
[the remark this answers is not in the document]
lockwobr: Fixed in e13b294e: `buildMonitoredValues` now returns `ErrCodeInvalidRequest` for an issuer without a subject (defense-in-depth; `parseFlags` already rejects it at the CLI). Added a test case.

tools/rekor-monitor/main.go
```diff
@@ -0,0 +1,166 @@
...
+	"time"
+
+	"github.com/NVIDIA/aicr/pkg/errors"
+)
+
+// defaultTimeout bounds a single monitor pass (TUF fetch, shard discovery,
+// consistency proof, and identity scan). It is generous relative to the observed
+// runtime (seconds to a couple of minutes) so a real scan never trips it, while
+// still capping a stalled network request rather than hanging the CI job.
+const defaultTimeout = 15 * time.Minute
```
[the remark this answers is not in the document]
lockwobr: Declining per maintainer preference: we keep `defaultTimeout` local to this standalone tool rather than importing the shared `pkg/defaults` for a single tool-specific constant. It's a named const (not a magic literal), and the tool otherwise imports only `pkg/trust`/`pkg/errors`, which are essential.

tools/rekor-monitor/monitor.go
```diff
@@ -0,0 +1,184 @@
...
+
+	// GetRekorShards filters the services to v2, so passing the full list (which
+	// also carries the legacy v1 entry) is fine and future-proof across rotation.
+	shards, shardOrigin, err := rekorv2.GetRekorShards(ctx, trustedRoot, signingConfig.RekorLogURLs(), opts.userAgent, "")
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeUnavailable, "failed to resolve Rekor v2 shards", err)
+	}
+
+	return &monitor{shards: shards, shardOrigin: shardOrigin, identities: identities}, nil
+}
```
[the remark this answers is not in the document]
lockwobr: Declining: the two clients share the same TUF cache (the second read is warm). The signing-config path must go through `pkg/trust` (it owns the private v2 target name) and the trusted-root path needs the concrete `*root.TrustedRoot` that `pkg/trust` doesn't expose, so sharing one client would require a `pkg/trust` API change for negligible gain (also tagged Heavy lift).

.github/workflows/rekor-monitor.yaml
```diff
@@ -73,56 +74,156 @@ concurrency:
...
-          echo "Resolved Rekor v2 shard: $url"
-          echo "url=$url" >> "$GITHUB_OUTPUT"
+          # --paginate with --jq evaluates the filter per page, so reduce across
+          # all pages in one jq (-s slurps the concatenated page objects); a
+          # per-page "last" would return several ids once artifacts span pages.
+          id="$(gh api --paginate "repos/${GITHUB_REPOSITORY}/actions/artifacts?name=${ARTIFACT_NAME}&per_page=100" \
+            | jq -rs '[.[].artifacts[] | select(.expired == false)] | sort_by(.created_at) | last | .id // empty')"
+          if [ -z "$id" ]; then
+            echo "No prior checkpoint artifact found; treating this as the first run."
+            exit 0
```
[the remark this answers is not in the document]
lockwobr: Declining: the forward-looking baseline is the intended v2 design — the 'clear the backlog' objective was from the abandoned parallel-paging approach. Pre-baseline entries are covered by release-time verification (documented in maintaining.md), and a one-time historical sweep is a tracked follow-up.

tools/rekor-monitor/monitor.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	// The previous checkpoint covered indices 0..prev.Size-1, so the first unseen
+	// entry is index prev.Size; there is nothing to scan when the log has not
+	// grown (cur.Size <= prev.Size). Starting at prev.Size (not prev.Size-1)
+	// avoids re-scanning the boundary entry, which would otherwise double-alert on
+	// it or, right after the first baseline, scan a pre-baseline entry.
+	if prev.Size == 0 || cur.Size <= prev.Size {
+		return 0, 0, false
+	}
+	start = int64(prev.Size)  //nolint:gosec // G115: log size never approaches int64 max
```
mchmarny: Off-by-one: the entry at index `prev.Size` is never scanned. The vendored `tiles.GetEntriesByIndexRange` (which `rekorv2.IdentitySearch` delegates to) is documented and implemented as exclusive-start `(start, end]`: it slices `endEntries[startOffset:]` with `startOffset = start+1`, and `start == end` returns zero entries. Upstream's own v2 loop accounts for this — `RekorV2MonitorLogic.GetStartIndex` in cmd/rekor_monitor/main.go passes `int64(prevCheckpoint.Size) - 1` (the last already-scanned index). Passing `start = prev.Size` here means each pass actually scans `[prev.Size+1, cur.Size-1]`, so the first new entry of every window escapes the identity check, and when exactly one entry was added (`cur.Size == prev.Size+1`, start==end) nothing is scanned at all. Because the cursor then advances, these indices are never revisited — a permanent, targetable blind spot (an attacker submitting a rogue entry immediately after a monitor pass has a good chance of landing at exactly `prev.Size`). Fix: `start = int64(prev.Size) - 1` (prev.Size >= 1 is already guaranteed by the guard above), and adjust the comment plus the `report` range output, which currently claims the boundary entry was covered.

.github/workflows/rekor-monitor.yaml
```diff
@@ -73,56 +74,161 @@ concurrency:
...
-          # Newest currently-listed v2 shard (majorApiVersion 2). Any live v2
-          # shard selects v2; the monitor then discovers the full set from TUF.
-          url="$(jq -er '[.rekorTlogUrls[] | select(.majorApiVersion == 2)] | sort_by(.validFor.start) | last | .url' signing-config.json)"
-          echo "Resolved Rekor v2 shard: $url"
-          echo "url=$url" >> "$GITHUB_OUTPUT"
+          # --paginate with --jq evaluates the filter per page, so reduce across
+          # all pages in one jq (-s slurps the concatenated page objects); a
+          # per-page "last" would return several ids once artifacts span pages.
+          id="$(gh api --paginate "repos/${GITHUB_REPOSITORY}/actions/artifacts?name=${ARTIFACT_NAME}&per_page=100" \
+            | jq -rs '[.[].artifacts[] | select(.expired == false)] | sort_by(.created_at) | last | .id // empty')"
```
mchmarny: Checkpoint artifact poisoning: this selects the newest non-expired artifact named `rekor-v2-checkpoint` from *any* workflow run in the repo — `actions/artifacts?name=` is repo-wide, and a fork `pull_request` run executes the PR's own workflow files, whose GITHUB_TOKEN can upload artifacts under any name into this repo. The restored checkpoint is parsed unsigned (`log.Checkpoint.Unmarshal`: origin/size/hash only), and a *genuine* checkpoint copied from the live log head passes the consistency proof trivially while collapsing the identity-scan window to empty — silently skipping every entry between the real cursor and head. That bypasses exactly the OIDC-compromise scenario this monitor exists to catch, and needs only the ability to open a PR. Mitigation: the artifacts API returns `workflow_run` metadata per artifact — filter in the jq to artifacts produced by this workflow on `main` (e.g. `select(.workflow_run.head_branch == "main")` plus a workflow match), or list this workflow's runs on main and fetch the artifact per-run.

tools/rekor-monitor/monitor.go
```diff
@@ -0,0 +1,191 @@
...
+func scanWindow(prev, cur *tlog.Checkpoint) (start, end int64, ok bool) {
+	if prev == nil || cur == nil {
+		return 0, 0, false
+	}
+	// The previous checkpoint covered indices 0..prev.Size-1, so the first unseen
+	// entry is index prev.Size; there is nothing to scan when the log has not
+	// grown (cur.Size <= prev.Size). Starting at prev.Size (not prev.Size-1)
+	// avoids re-scanning the boundary entry, which would otherwise double-alert on
+	// it or, right after the first baseline, scan a pre-baseline entry.
+	if prev.Size == 0 || cur.Size <= prev.Size {
```
mchmarny, line 170: Shard rotation opens a silent scan gap. On the first pass after yearly rotation (log2025-1 -> log2026-1), `checkConsistency` proves the *old* shard forward but returns the *new* shard's checkpoint (see vendored `VerifyConsistencyWithCheckpoint`), so `prev` and `cur` are sizes of different logs. The new shard is small, so this `cur.Size <= prev.Size` branch treats it as "no growth": the scan is skipped with no log line, and the cursor jumps to the new shard — old-shard entries appended after the last pass and new-shard entries `[0, cur.Size)` are never identity-scanned (the vendored `IdentitySearch` also only reads the latest shard: "TODO: handle sharding"). Recommend handling `prev.Origin != cur.Origin` explicitly: at minimum report the re-baseline in the output (so the gap is visible/triageable), ideally also scan the new shard from index 0 in that pass.

tools/rekor-monitor/main.go
```diff
@@ -0,0 +1,172 @@
...
+	// Advance the cursor after a completed consistency proof and scan. On a
+	// finding we still advance (the alert is surfaced below) so we do not
+	// re-alert every hour; the workflow deduplicates the alert issue by title.
+	if err := store.write(prev, cur); err != nil {
+		return err
+	}
+
+	out.report(w)
+	if out.hasFindings() {
+		return errors.New(errors.ErrCodeConflict, "unrecognized entry under the monitored release identity")
```
mchmarny: nit: `hasFindings()` is also true when only `out.failed` (unparseable) entries exist, but the error always reads "unrecognized entry under the monitored release identity" — mislabeling a parse failure as an identity match in the run log and the alert issue. Since the cursor has already advanced past these indices (they will never be re-scanned), accurate triage wording matters; include the found/failed counts, e.g. distinguish "N identity match(es)" from "M entries failed verification in the scanned window".

mchmarny (commented):
Replaces the broken upstream sigstore/rekor-monitor reusable workflow with a small in-repo tool that reads AICR's Rekor v2 signing config and reuses upstream library packages for consistency proofs and identity search. The structure, tests, docs, and workflow hygiene are strong (SHA-pinned actions, minimal permissions, env-var script inputs), and the fail-closed restore/advance ordering is well thought out. CI is fully green — though note CI cannot exercise the live-log scan boundary, so a green run doesn't rule out the first issue below.

Two blockers before merge, detailed inline: an off-by-one in the identity-scan window (the vendored `GetEntriesByIndexRange` is exclusive-start, so the first new entry of every pass is permanently skipped, and a single-entry window scans nothing), and the cross-run checkpoint artifact is fetched by repo-global name with no provenance filter, letting any workflow run (including fork PR runs) poison the cursor and silently collapse the scan window. Also one medium (shard-rotation scan gap) and one nit (finding-vs-parse-failure wording in alerts). Holding approval until the two blockers are addressed.

tools/rekor-monitor/monitor.go
lockwobr: Great catch, and confirmed in the vendored code (`tiles.GetEntriesByIndexRange`: `if start == end { return nothing }`, reads from `start+1`). Reverted to `start = prev.Size - 1` in a6516bce so the first new entry (index `prev.Size`) is covered, and fixed the comment + `report` to show the effective covered range `[prev.Size, cur.Size-1]`. This was my regression from an earlier suggestion I took without checking the exclusive-start semantics.

.github/workflows/rekor-monitor.yaml
lockwobr: Fixed in a6516bce: the fetch now scopes to THIS workflow's runs on `main` (`actions/workflows/rekor-monitor.yaml/runs?branch=main`, newest-first, first run with a live artifact) instead of the repo-wide `artifacts?name=` list, so a fork `pull_request` run's same-named upload can't poison the checkpoint (forks can't run on main).

tools/rekor-monitor/monitor.go
lockwobr: Fixed in a6516bce: `observe` now detects `prev.Origin != cur.Origin` and, instead of silently treating the smaller new shard as 'no growth', re-baselines on the new shard and reports the rotation loudly (pointing at the maintaining.md shard-rotation note). Full cross-shard coverage is bounded by the vendored `IdentitySearch` 'TODO: handle sharding', so this makes the boundary gap visible/triageable; a deeper backfill is a tracked follow-up. Added a shard-rotation test.

tools/rekor-monitor/main.go
lockwobr: Fixed in a6516bce: findings now surface via `out.findingError()` = "scanned window has N identity match(es) and M entr(y/ies) that failed verification", and `report` distinguishes `MATCH:` vs `FAILED:`, so a parse failure is no longer labeled an identity match.

.github/workflows/rekor-monitor.yaml
```diff
@@ -73,56 +74,173 @@ concurrency:
...
-          echo "url=$url" >> "$GITHUB_OUTPUT"
+          id=""
+          for run in $(gh api \
+            "repos/${GITHUB_REPOSITORY}/actions/workflows/rekor-monitor.yaml/runs?branch=main&per_page=30" \
+            --jq '.workflow_runs | sort_by(.created_at) | reverse | .[].id'); do
+            id="$(gh api "repos/${GITHUB_REPOSITORY}/actions/runs/${run}/artifacts" \
+              | jq -r --arg n "${ARTIFACT_NAME}" \
+                  '[.artifacts[] | select(.name == $n and .expired == false)] | .[0].id // empty')"
+            [ -n "$id" ] && break
+          done
```
mchmarny: `set -e` does not catch a failing command substitution in a `for` word list: if this `gh api` run-listing call fails (transient 500, rate limit), the loop body never runs, `id` stays empty, and the step prints "first run" and exits 0 — verified: `set -euo pipefail; for run in $(exit 1); do ...; done` continues and exits 0. The monitor then re-baselines at the current head, silently collapsing the identity-scan window — the exact silent-skip the SECURITY comment above defends against — and the gap is permanent because this run's fresh baseline becomes the newest artifact. Capture the listing first (`runs="$(gh api ...)"` — a failing substitution in an assignment *does* trip errexit), then iterate over `$runs`, so a listing failure fails the step and alerts instead of baselining.

tools/rekor-monitor/monitor.go
```diff
@@ -0,0 +1,211 @@
...
+
+// scanWindow computes the [start, end] entry-index window to scan between the
+// previous checkpoint and the current head. ok is false when there is no prior
+// checkpoint (first run) or the window is empty/degenerate, so the caller only
+// baselines.
+func scanWindow(prev, cur *tlog.Checkpoint) (start, end int64, ok bool) {
+	if prev == nil || cur == nil {
+		return 0, 0, false
+	}
+	if prev.Size == 0 || cur.Size <= prev.Size {
```
mchmarny, line 170: The `prev.Size == 0` guard creates a silent scan gap: baseline taken on an empty shard (size 0 — plausible right after a shard is created), next run has `cur.Size == N` → `ok` is false, the cursor advances to N, and entries 0..N-1 are never identity-scanned, with no report. If `GetEntriesByIndexRange` accepts `start = -1`, the exclusive-start semantics make `(−1, end]` exactly right and the guard can allow `prev.Size == 0`; if upstream rejects negative start, at least report the skipped `[0, N-1]` window loudly (like the rotation case) instead of advancing silently.

docs/contributor/maintaining.md
```diff
@@ -75,48 +75,49 @@ grows, so it can never keep up inside a bounded CI job: the earlier v1
...
+
+### Checkpoint and first run
+
+The monitor persists its cursor as the `rekor-v2-checkpoint` artifact between
+runs (a deliberately fresh name, so the stale v1 `checkpoint` artifact from the
+earlier design is simply ignored, no migration). The **first** run has no prior
+checkpoint, so it establishes a baseline at the current v2 tree head and skips
+the identity scan; every run after that scans only the newly-added window.
+Entries predating the baseline are covered by release-time verification (the
+`aicr verify` path), not by this monitor. Shard rotation (`log2025-1` ->
```
mchmarny: The tool's rotation report says "see the shard-rotation note in docs/contributor/maintaining.md", but this is the only rotation mention and it says rotation "needs no change here" — there's no note covering what the operator actually sees on rotation: the re-baseline, the un-scanned entries around the rotation boundary, and that those are covered by release-time `aicr verify`. Add that note here (or drop the pointer from the report).

mchmarny (changes requested):
This is the right shape for the fix — own only the AICR-specific shard/config resolution, reuse upstream's verification libraries, and keep the workflow the auditable source of truth for the monitored identity. The fail-closed checkpoint restore, bounded zip handling, rotation round-trip test, and off-by-one window tests are all careful work.

One high-severity finding blocks: the artifact-fetch step's `for run in $(gh api ...)` swallows run-listing failures under `set -e` (verified), so a transient API error silently re-baselines and permanently collapses the identity-scan window — the exact failure mode the step's own SECURITY comment guards against. One medium (silent scan gap when a shard is baselined at size 0) and one doc nit inline. All are small fixes; the design doesn't need to change.

.github/workflows/rekor-monitor.yaml
lockwobr: Fixed in ab8ebc28: the run list is captured in a standalone assignment (so `set -e` aborts on a failed API call) and iterated with `while IFS= read -r`, replacing the `for`-word-list command substitution that `set -e` doesn't catch. A transient run-listing failure now fails the step (and alerts) instead of falling through to a silent re-baseline.

tools/rekor-monitor/monitor.go
lockwobr: Investigated, and `prev.Size == 0` turns out to be unreachable: `WriteCheckpointRekorV2` skips size-0 writes ("skipping write of checkpoint: size is 0"), and the checkpoint format won't parse a crafted size-0 entry ("invalid checkpoint - too few newlines"). So an empty-shard baseline never persists; it collapses to a normal first run, which is reported ("baseline established at tree size N (first run; identity scan skipped)") and whose [0, N-1] gap is the documented forward-looking behavior covered by release-time `aicr verify`. I kept the `prev.Size == 0` guard as cheap defense (it also keeps start >= 0) with an explanatory comment, and documented the empty-shard case in maintaining.md rather than adding an unreachable report branch.

docs/contributor/maintaining.md
lockwobr: Fixed in ab8ebc28: added a 'Shard rotation (and what the operator sees)' subsection to docs/contributor/maintaining.md covering the re-baseline log line, the un-scanned boundary entries, the empty-shard-collapses-to-first-run case, and that these are covered by release-time `aicr verify`. The tool's report pointer now resolves to a real note.

lockwobr:
Thanks for the thorough security review, @mchmarny — several of these were real, and M1 in particular caught a regression I'd introduced by taking an earlier suggestion without checking the vendored semantics. All addressed, and the branch is rebased onto latest `main`.

**Fixed:**
- **Scan-window off-by-one** (M1): confirmed `tiles.GetEntriesByIndexRange` is exclusive-start (`start == end` returns nothing, reads from `start+1`), so reverted to `start = prev.Size - 1`. The first new entry (index `prev.Size`) is covered again, and the report now shows the true covered range `[prev.Size, cur.Size-1]`.
- **Checkpoint artifact poisoning** (M2 + N1): the fetch now scopes to *this workflow's runs on `main`* (a fork `pull_request` run can't run on main), and the run list is captured in a standalone assignment so `set -e` aborts on a failed API call instead of falling through to a silent re-baseline.
- **Shard rotation** (M3 + N3): `prev.Origin != cur.Origin` is now detected, re-baselines on the new shard with a loud log line, and `docs/contributor/maintaining.md` documents exactly what the operator sees (re-baseline, un-scanned boundary entries, covered by release-time `aicr verify`).
- **Findings message** (M4): distinguishes N identity match(es) from M entr(y/ies) that failed verification.

**Investigated, no code change:**
- **`prev.Size == 0` gap** (N2): unreachable. `WriteCheckpointRekorV2` skips size-0 writes ("skipping write of checkpoint: size is 0") and the checkpoint format won't parse a crafted size-0 entry ("invalid checkpoint - too few newlines"), so an empty-shard baseline never persists and collapses to a normal (reported) first run. Kept the guard as cheap defense and documented the case.

Each thread has the specifics and is resolved. Two follow-ups are intentionally out of scope (tracked separately): an external dead-man's-switch for liveness (a job that fails to *start* can't self-alert) and auto-derived known-good indices so identity hits stay high-signal. Ready for another pass when you have a minute.

.github/workflows/rekor-monitor.yaml
```diff
@@ -73,56 +74,180 @@ concurrency:
...
+        # leaves checkpoint.zip absent so the monitor baselines. The monitor
+        # reads the zip natively (--restore-zip); no unzip here.
+        #
+        # SECURITY: the repo-wide `artifacts?name=` list would also return
+        # same-named artifacts uploaded by *any* run, including a fork
+        # pull_request run whose GITHUB_TOKEN can upload under any name. A
+        # poisoned checkpoint (a genuine one copied from the live log head) passes
+        # the consistency proof while collapsing the identity-scan window to
+        # empty, silently skipping entries. So we trust only artifacts produced by
+        # THIS workflow's runs on `main` (forks cannot run on main).
```
mchmarny: **[high]** This still accepts checkpoint artifacts from untrusted fork PR runs. GitHub filters `branch=main` by the run head branch; a fork PR whose source branch is named `main` appears here, and the live workflow-runs API already returns such runs for this repo. A fork can modify this workflow to add a `pull_request` trigger, upload `rekor-v2-checkpoint`, and collapse the next identity window with a genuine head checkpoint. Filter to trusted events (`schedule` / `workflow_dispatch`) and require `head_repository.full_name == GITHUB_REPOSITORY` or repository ID equality before using any run.

tools/rekor-monitor/monitor.go
```diff
@@ -0,0 +1,215 @@
...
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeInternal, "consistency verification failed", err)
+	}
+	return cur, nil
+}
+
+// scanIdentity searches the [start, end] entry-index window for the monitored
+// identity, returning matching entries and entries that failed to parse.
+func (m *monitor) scanIdentity(ctx context.Context, start, end int64) ([]identity.MonitoredIdentity, []identity.FailedLogEntry, error) {
+	found, failed, err := rekorv2.IdentitySearch(ctx, m.shards, m.shardOrigin, m.identities, start, end)
```
mchmarny: **[high]** `IdentitySearch` is called without `WithCARootsFile` or `WithCAIntermediatesFile`. The vendored library returns success without certificate-chain validation when both paths are empty, so a self-signed certificate with the matching SAN and OIDC issuer can create a false compromise alert. Build the Fulcio root and intermediate files from `trustedRoot` via `ConfigureTrustedCAs`, retain them through the scan, and pass both search options.

tools/rekor-monitor/main.go
```diff
@@ -0,0 +1,182 @@
...
+			out.scanned = true
+			out.from, out.to = start+1, end // inclusive range actually covered
+			out.found, out.failed = found, failed
+		}
+	}
+
+	// Advance the cursor after a completed consistency proof and scan. On a
+	// finding we still advance (the alert is surfaced below) so we do not
+	// re-alert every hour; the workflow deduplicates the alert issue by title.
+	if err := store.write(prev, cur); err != nil {
```
mchmarny: **[high]** The cursor advances before `hasFindings`, including entries that failed verification. The next run never revisits those indices and satisfies the workflow `success()` condition, which auto-closes the alert without triage. Do not advance past failed entries, and keep identity findings in durable state until a maintainer explicitly acknowledges or allowlists them; a subsequent empty window is not resolution.

.github/workflows/rekor-monitor.yaml
```diff
@@ -73,56 +74,180 @@ concurrency:
...
-          url="$(jq -er '[.rekorTlogUrls[] | select(.majorApiVersion == 2)] | sort_by(.validFor.start) | last | .url' signing-config.json)"
-          echo "Resolved Rekor v2 shard: $url"
-          echo "url=$url" >> "$GITHUB_OUTPUT"
+          # Capture the run list in a standalone assignment so a failed API call
+          # aborts the step (set -e). A failing command substitution in a `for`
+          # word-list is NOT caught by set -e, so the loop would just not run and
+          # fall through to a silent re-baseline: the exact identity-scan gap this
+          # step exists to prevent.
+          runs="$(gh api \
+            "repos/${GITHUB_REPOSITORY}/actions/workflows/rekor-monitor.yaml/runs?branch=main&per_page=30" \
```
mchmarny: **[high]** Only the latest 30 workflow runs are searched although the artifact is retained for 30 days. After 30 artifact-less hourly runs, such as startup failures or API outages, an older valid cursor is ignored and this falls through to first-run re-baselining, permanently skipping the outage window. Paginate through retained run history and fail closed when prior state cannot be recovered.

mchmarny (changes requested):
Four high-severity security and correctness blockers remain in checkpoint provenance, Fulcio trust validation, alert durability, and cursor recovery. These can silently skip identity windows, accept untrusted certificates as monitored identities, or clear alerts before triage. Requesting changes before merge.

.github/workflows/rekor-monitor.yaml
lockwobr: Fixed in 7b9558df: the fetch now filters on `.workflow_run.head_repository_id == .workflow_run.repository_id`, which rejects every fork run (a fork's head-repo id differs from this repo's id) regardless of branch name, in addition to `head_branch == "main"`. So a fork PR whose source branch is named 'main' is excluded.

tools/rekor-monitor/monitor.go
lockwobr: Fixed in 7b9558df: `newMonitor` materializes the Fulcio CA root/intermediate PEM files from the trusted root via `rmutil.ConfigureTrustedCAs`, and `scanIdentity` passes `WithCARootsFile`/`WithCAIntermediatesFile`, so `IdentitySearch` validates the certificate chain (a self-signed cert with a matching SAN/issuer no longer registers as a match). The temp CA files are removed via a deferred cleanup in `run`.

tools/rekor-monitor/main.go
lockwobr: Fixed in 7b9558df: `observe` no longer advances the cursor on a finding — it returns the finding error without writing the checkpoint, so the same window is re-scanned every run (keeping the alert issue open) until a maintainer triages it. A later empty window can no longer auto-close it; that only happens once the cursor is deliberately advanced after triage. Durable allowlisting of acknowledged/known-good indices is the tracked follow-up.

.github/workflows/rekor-monitor.yaml
lockwobr: Fixed in 7b9558df: switched to the artifacts API with `--paginate` (full retained history, not just 30 runs), so an older valid cursor survives a long stretch of artifact-less runs (startup failures/outages). A failed API call aborts the step (`set -e`/pipefail on the assignment) so it fails closed rather than re-baselining; first-run is reached only when the API succeeds and returns no matching artifact.

mchmarny (approved):

PR #1756, opened 2026-07-14
feat(recipes): default AKS to Azure-managed GPU driver profile

### LINKED ISSUE #1842: nodewright tuning Skyhook CR orphaned on tuningEnabled true→false under Flux (Helm hook not release-managed)
## Summary

The nodewright tuning Skyhook CR (`tuning.yaml`, `tuning-gke.yaml`, `tuning-generic.yaml`) is applied as a Helm hook (`helm.sh/hook: post-install,post-upgrade`, `hook-delete-policy: before-hook-creation`). Helm hook resources are **not release-managed**, and `before-hook-creation` only deletes the previous resource immediately before creating a replacement hook. Under `--deployer flux` (raw template), flipping `nodewrightcustomizations:tuningEnabled` from `true` to `false` removes the CR from the render, so no replacement hook is created, the delete never fires, and the previously applied Skyhook CR is **orphaned** — it keeps tuning new nodes.

## Mechanism

- `recipes/components/nodewright-customizations/manifests/tuning-gke.yaml` (and siblings) annotate the Skyhook CR with `helm.sh/hook: post-install,post-upgrade` + `helm.sh/hook-delete-policy: before-hook-creation`.
- Per Helm's hook-lifecycle docs, hook resources are not tracked as part of the release and `before-hook-creation` deletes the prior resource only when a new hook of the same kind/name is about to be created.
- On an enabled→disabled transition under Flux, the CR disappears from the render → no new hook → `before-hook-creation` never runs → the old `tuning` Skyhook persists.

## Reachability

Requires the Flux deployer plus an enabled→disabled `tuningEnabled` transition on a `tuning-gke.yaml`/`tuning-generic.yaml` recipe (or `enabled: false`). The in-repo Helm/Argo path strips hook annotations at the bundle layer, so this is specific to the raw-template Flux path. Documented as a limitation in `docs/integrator/components/nodewright.md` (PR #1756).

## Proposed fix

Make the tuning Skyhook CR release-managed instead of a Helm hook, so Flux prunes it when it leaves the render. The hook annotations exist to order the CR after the Skyhook CRD install; replacing them requires an equivalent ordering mechanism (e.g. Flux `dependsOn` / healthChecks, or the bundle-layer NNN ordering the in-repo deployers already rely on) that holds across all deployers. This is a cross-deployer change and deserves its own PR.

Overlaps: #1835 (vendored-bundle install-only hook lifecycle), #1820 (in-place tuning residue on upgrade). Related: #1756 (introduces the `tuningEnabled` gate and the doc note).

yuanchen8911:
## Summary

Switches the AKS recipes' default GPU ownership mode from GPU-Operator-managed to the **AKS driver-only (Azure-managed) profile** — the Azure default, where the AKS node image preinstalls the NVIDIA driver and container toolkit: `driver.enabled=false`, `toolkit.enabled=false`, `operator.runtimeClass=nvidia-container-runtime`, plus DRA `nvidiaDriverRoot=/`. The AKS recipes also ship with the nodewright `nvidia-tuned` package **disabled by default** under this profile via a new per-package `tuningEnabled` gate (`nvidia-setup`/RDMA host prep still runs).

## Motivation / Context

AICR defaults should follow the CSP default. Azure's default node pool provisioning (`--enable-managed-gpu=false`, no `--gpu-driver none`) preinstalls the driver and container toolkit, but the AKS recipes assumed the GPU-Operator-managed profile (`--gpu-driver none` pools). This PR makes the CSP default the recipe default; GPU-Operator-managed mode remains supported as a documented, explicit bundle-time override set.

A blocking bundle-time driver-ownership coherence validation (driverless-cluster mismatch, partial profile flips, legacy-recipe handling) is deliberately **not** part of this PR — it is tracked in #1757 (release-blocking) and implemented by the stacked PR #1819. #1819 must actually merge in the **same tagged release** as this flip (not merely be labeled) so users transition atomically from a pre-flip release to a release carrying both the flip and the guard (agreed review disposition).

Fixes: N/A
Related: #1680, #1687, #1757; evidence in #1840

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Other: `pkg/client/v1` (driver auto-detect), `recipes/`, `tests/uat/azure`

## Implementation Notes

- `values-aks.yaml` and `values-aks-training.yaml` both restate the full profile because a training overlay's `valuesFile` replaces the base's. `driver.rdma.useHostMofed=false` and the toolkit `RUNTIME_CONFIG_SOURCE=file` env (#1688) stay declared — inert in driver-only mode, but correct if an ownership-mode flip re-enables them.
- **DRA driver root:** with `driver.enabled=false` nothing populates `/run/nvidia/driver`, so the AKS overlay retargets `nvidia-dra-driver-gpu.nvidiaDriverRoot` to `/` (host-installed driver), matching OKE/kind. `TestDriverRootLockstep` now rejects any preinstalled-driver overlay that inherits the container driver root instead of skipping. The GPU-Operator-managed override set therefore includes `--set dradriver:nvidiaDriverRoot=/run/nvidia/driver`.
- The AKS overlays now carry the preinstalled-driver marker, so the snapshot driver auto-detect **injects** the audited `driver.enabled=false` override on AKS (previously warn-and-skip). The inverse mismatch — an AKS recipe resolved against a snapshot whose sampled GPU node has no driver loaded (`--gpu-driver none` signature) — logs a warning naming the bundle-time GPU-Operator-managed override set; blocking enforcement is #1757.
- **nvidia-tuned disabled by default on AKS:** `tuning.yaml` gains a per-package `tuningEnabled` gate (default `true`); the AKS overlays set it to `false`. Disabling skips the package's node-reboot interrupt, and an A/B on an AKS H100 cluster measured the tuned inference profile (isolcpus/hugepages) regressing TTFT from ~550 ms to ~6 s, so untuned is the better default today. With the gate off, `nvidia-setup-full` chains `dependsOn` directly off `nvidia-setup-kernel` (no dangling reference to the omitted package). `nvidia-setup` stays enabled — on AKS it carries the RDMA host prep (MEMLOCK limits + IB module loading) that replaced the `ib-node-config` DaemonSet. Re-enable with `--set nodewrightcustomizations:tuningEnabled=true`. The `tuningEnabled` gate predicate itself leaves non-AKS (EKS/GKE) renderings byte-identical; note that the separate Helm-hook removal on the tuning Skyhook CR (below) does change that CR's annotations on all deployers — a one-time GitOps checksum drift for existing EKS/GKE users, intentional so `tuningEnabled=false` retracts the CR under Flux. Documented in `aks-gpu-setup.md`.
- **Sprig-safe template gates:** the `tuningEnabled` gates (package + `dependsOn` switch) and all four `autoTaintNewNodes` sites (`tuning.yaml`, `no-op.yaml`, `tuning-generic.yaml`, `tuning-gke.yaml`) use `ne (toString (index $cust "<key>")) "false"` — the same predicate as the file's `enabled` gate — instead of `| default true`. Sprig's `default` treats boolean `false` as empty, so under `--deployer flux` (raw template shipped to real Helm) an explicit `false` was silently swallowed and re-enabled the gated behavior; the in-repo engine was unaffected. Semantics: only explicit `false` disables; absent/`true` render.
- **gdrcopy disabled under the driver-only profile:** `gdrcopy.enabled: false` in `values-aks.yaml`/`values-aks-training.yaml`, and the inline `gdrcopy.enabled: true` overrides removed from the `h100-aks-training`/`a100-aks-training` overlays — the operator deploys gdrcopy as a sidecar container in the operator-managed driver pod, so it cannot deploy with `driver.enabled=false`. Whether GDRCopy capability exists on the Azure-managed image (node-image `gdrdrv`) is unverified; documented in `aks-gpu-setup.md` with the GPU-Operator-managed re-enable path.
- **GPUDirect RDMA (`nvidia-peermem`) preserved under the driver-only default:** with `driver.enabled=false` there is no operator driver DaemonSet to load `nvidia-peermem`, and the AKS-managed driver does not load it automatically — Azure's RDMA guidance requires a dedicated reloader in this mode. The AKS recipes now ship Azure's `nvidia-peermem-reloader` DaemonSet as a `network-operator` manifest (a probeless retry loop — Ready immediately — re-asserts `modprobe` via chroot into the host root, so the node's own kmod loads the module: no runtime apt, digest-pinned busybox image, and no readiness gate that could deadlock the GPU-Operator-managed fallback profile, where the driver only appears after gpu-operator installs and the operator's driver pod loads `nvidia-peermem` itself), scoped to IB-capable nodes via the same `pci-15b3` NFD label as the NicClusterPolicy and removed by the documented RDMA opt-out (`--set networkoperator:enabled=false`). Documented in `aks-gpu-setup.md`; example regenerated (sole delta: the added manifest file); BOM regenerated (adds the digest-pinned `docker.io/library/busybox` image to the network-operator section).
- **In-place upgrade residue:** upgrading a cluster that already applied `nvidia-tuned` to a `tuningEnabled: false` bundle does not revert the applied host state (no uninstall block; nodewright uninstall defaults off) — `aks-gpu-setup.md` documents node-pool recreation as the clean path; automated remediation is tracked in #1820.
- The generated tuning-status table in `docs/integrator/components/nodewright.md` cannot see value gates; an annotation below it notes AKS disables `nvidia-tuned` by default.
- `tests/uat/azure/cluster-config.yaml` flips `gpuDriverInstall: true` to match. **The existing Azure UAT cluster's GPU pool was provisioned driverless (`gpuDriverInstall: false`) and must be recreated before the next Azure UAT run.**
- **Validated end-to-end on AKS H100 (aicr-test6, 2026-07-21):** 2x `Standard_ND96isr_H100_v5`, node image `AKSUbuntu-2404gen2containerd-202606.19.0`, kernel `6.8.0-1059-azure`, preinstalled driver 580.126.09, `gpuProfile.driver=Install`. Deployment phase 4/4 checks passed for **both intents** (operator-health, expected-resources, gpu-operator-version, check-nvidia-smi); the training performance phase passed the calibrated NCCL all-reduce gate at **157.29 GB/s** (16 GiB message size, 2 nodes, gate >= 150). `aks-gpu-setup.md` records this in place of the earlier "not yet validated end-to-end" caveat, scoped to H100 / this node image (other SKUs unexercised). The GPU-Operator-managed alternative was previously validated on aicr-test1.
- **peermem nuance from the same cluster (documented in `aks-gpu-setup.md`):** this node image's preinstalled `nvidia-peermem.ko` is a **stub DKMS build** — `modprobe nvidia-peermem` fails with `Invalid argument` and no dmesg trace (the signature of NVIDIA's conftest finding no OFED-style headers at image bake and compiling the peer-memory support out) — even though the running kernel exports `ib_register_peer_memory_client` and the IB stack (`ib_core`, `mlx5_ib`) is loaded after nodewright host prep. The reloader's retry loop is the designed harmless steady state there; GPUDirect RDMA is carried by the **DMA-BUF path** (driver 580 + kernel 6.8 + mlx5 + a DMA-BUF-capable NCCL), which is what delivered the 157.29 GB/s. The doc's host-staging fallback warning is now scoped to images where neither path is available.
- **Device-isolation hardening under the driver-only profile:** with `driver.enabled=false`/`toolkit.enabled=false` there is no operator toolkit DaemonSet to harden `/etc/nvidia-container-runtime/config.toml`, so the AKS node image's permissive toolkit defaults let **any unprivileged pod with `NVIDIA_VISIBLE_DEVICES=all` grab all host GPUs unallocated** (a `secure-accelerator-access` failure). The AKS recipes ship an `nvidia-toolkit-hardening` DaemonSet that re-asserts the hardened keys via `nvidia-ctk config --in-place`, with a **fail-closed** readiness probe (Ready only while `nvidia-ctk` can harden the config — a missing/incompatible toolkit surfaces instead of latching green; a transient valid permissive rewrite is self-repaired by the next assert within 60s, no Ready transition). The DaemonSet renders **only** in the driver-only profile (`toolkit.enabled=false`); under the GPU-Operator-managed fallback (`--gpu-driver none` + `toolkit.enabled=true`) it is gated off (the operator owns the toolkit; the same hardened keys are set via `gpu-operator.toolkit.env` `ACCEPT_NVIDIA_VISIBLE_DEVICES_*` in `values-aks*.yaml`), which also avoids a helm-wait deadlock where `nvidia-ctk` isn't preinstalled. Scope: this closes the **env-var** device-request path (the one `secure-accelerator-access` exercises); the **volume-mounts** path remains open (same posture as operator-managed mode), so full multi-tenant isolation additionally needs an admission policy restricting a `/dev/null`-backed mount whose destination is under `/var/run/nvidia-container-devices`. Migrating into nodewright `nvidia-setup` (pre-schedulable, closes the boot-window race) is tracked in #1839. Documented in `aks-gpu-setup.md`.
- **Migration:** recipes generated before this change reference the AKS values files by path and will resolve the flipped values on their next bundle while retaining their old baked DRA overrides — regenerate pre-flip AKS recipes (`aicr recipe`) or supply the complete four-flag GPU-Operator-managed override set. Documented in `aks-gpu-setup.md`; automatic detection/enforcement is #1757 (release-blocking). The committed `examples/recipes/aks-training.yaml` is regenerated in this PR for exactly this reason — the pre-flip example rendered an incoherent driver/DRA combination on this branch (verified: the regenerated example bundles to `driver.enabled=false`, `toolkit.enabled=false`, `operator.runtimeClass=nvidia-container-runtime`, DRA root `/`). `make bom-docs` regenerated (the BOM renders from base `values.yaml`); the only delta across the branch is upstream digest drift, committed — see Testing.

## Testing

```bash
make qualify   # test-coverage, lint, tuning-check, e2e, scan, license-check — all pass
```

- Coverage: `pkg/client/v1` 76.8% → 77.0%, `pkg/recipe` 87.5% → 87.5%.
- Updated tests: `TestResolveRecipeFromSnapshot_GPUDriverAutoDetect` (AKS+preinstalled now injects; AKS/GKE-COS+absent warns, no injection), strengthened `TestDriverRootLockstep` (preinstalled-driver overlays must not inherit the container DRA root).
- Regenerated example verified: `aicr bundle -r examples/recipes/aks-training.yaml` renders `driver.enabled=false`, `toolkit.enabled=false`, `operator.runtimeClass=nvidia-container-runtime`, DRA root `/`. (Example regen + docs note landed after the full qualify run; scoped checks for that delta: yamllint on the regenerated YAML and the bundle-render verification above — generated-artifact/doc-only, no Go or lintable-config change.)
- Example regenerated post-tuning-gate: `examples/recipes/aks-training.yaml` re-resolved with the dev CLI at the final head, now carrying `tuningEnabled: false` on nodewright-customizations; bundle render of the example omits `nvidia-tuned`; yamllint + `go test -race ./pkg/recipe/...` green.
- nvidia-tuned gate: new `TestNodewrightTuningGateAcrossCatalog` (`pkg/recipe`) renders the real `tuning.yaml` for every catalog leaf wiring it — all 3 AKS leaves omit `nvidia-tuned` and rewire `dependsOn` to `nvidia-setup-kernel`; all 10 EKS leaves keep the full chain; flipping `tuningEnabled=true` over AKS values restores it. Render evidence: EKS output is byte-identical pre/post change; `--set nodewrightcustomizations:tuningEnabled=true` restores the pre-change AKS rendering byte-for-byte. Scoped gates for this delta: `go build ./...`, `go test -race` on `pkg/recipe`, `pkg/tuning`, `pkg/bom`, `pkg/manifest`, `golangci-lint` on `./pkg/recipe/...` (0 issues, isolated caches), yamllint on the changed overlays, and `make bom-docs` (no diff attributable to this change; a pre-existing upstream `ubuntu:24.04` → `26.04` image drift in an unrelated GKE component was observed; committed in a later revision — see Testing).
- End-to-end CLI verification: `aicr query` confirms rendered values `driver.enabled=false`, `toolkit.enabled=false`, `operator.runtimeClass=nvidia-container-runtime`, DRA root `/` for both intents; a driver-loaded snapshot resolves the full 7-overlay AKS chain with the injected override; a no-driver snapshot resolves with the mismatch warning; bundling with the four GPU-Operator-managed overrides renders `driver.enabled=true`, `runtimeClass=nvidia`, DRA root `/run/nvidia/driver`.
- Sprig-gate verification: real `helm template` (Helm v4.2.3) reproduces the pre-fix bug (`tuningEnabled: false` still rendered `nvidia-tuned`; `autoTaintNewNodes: false` rendered `true`) and confirms the fix (`false` disables, absent/`true` render, `dependsOn` rewires); an in-repo `pkg/manifest.Render` byte-compare across all 4 nodewright manifests x 3 value sets (absent / explicit false / explicit true) is byte-identical pre/post rewrite, so no deployed manifest changes for the in-repo engine.
- Live-cluster validation (aicr-test6: AKS, 2x `Standard_ND96isr_H100_v5`, Azure-managed driver 580.126.09, node image `AKSUbuntu-2404gen2containerd-202606.19.0`; dev build of this branch, driver-only profile, `tuningEnabled=false`, 2026-07-21) — complete results:

  | Recipe | Deployment | Conformance | Performance |
  |---|---|---|---|
  | `h100-aks-training` | 4/4 passed | — | `nccl-all-reduce-bw` **passed** — **157.29 GB/s** @ 16 GiB across 2 nodes (gate >= 150) |
  | `h100-aks-inference` | 4/4 passed | — | none defined (platform-agnostic recipe; validators=0, skipped — by design) |
  | `h100-aks-ubuntu-inference-dynamo` | 4/4 passed | **15/16 passed, 0 failed, 1 skipped** (see below) | `inference-perf` **passed** — **147,601 tok/s** (gate >= 50,000), TTFT p99 **564 ms** (gate <= 2,000); Qwen/Qwen3-8B @ 256 concurrency/GPU, dynamo-router |
  | `h100-aks-ubuntu-training-kubeflow` | 4/4 passed | **14/15 passed, 0 failed, 1 skipped** | `nccl-all-reduce-bw` **passed** — **157.48 GB/s** (gate >= 150) |

  Two full-phase runs (deployment + conformance + performance, each with an attestation bundle emitted) — `h100-aks-ubuntu-inference-dynamo` and `h100-aks-ubuntu-training-kubeflow` — passed every conformance check including **`secure-accelerator-access`**, confirming the device-isolation hardening (Implementation Notes) closes the gap on both intents: `cluster-autoscaling` is the only skip (no Karpenter on this cluster). An earlier deployment/performance-only cycle reported an unhardened-baseline `secure-accelerator-access` FAIL before the hardening DaemonSet was added; the fix flips it to pass while an allocated 1-GPU pod still sees exactly its device. The training-kubeflow run added only the `kubeflow-trainer` components on top of the existing stack (no teardown of the shared or inference components).

  Context: recipes were generated from a live cluster snapshot (driver auto-detect injected `driver.enabled=false`); full undeploy + clean-sweep between the training and inference cycles (0 releases/CRDs/webhooks residue); Skyhook ran `nvidia-setup` only (`tuningEnabled=false`), one rolling reboot on first deploy, host state survived redeploy; NCCL bandwidth was delivered via the DMA-BUF GPUDirect path with the peermem reloader in its expected retry loop (stub `nvidia-peermem.ko` on this node image; see Implementation Notes). The Dynamo runs also reconfirm the untuned-profile inference numbers from the earlier aicr-test5 A/B (~146k tok/s / 547 ms) on a second cluster.
- First `inference-perf` attempt failed with `DynamoGraphDeployment not ready: TIMEOUT` — storage-bound, not a performance failure: 8 vLLM decode workers cold-loading the 15.26 GiB checkpoint concurrently from one RWO model-cache PVC on the cluster-default Azure disk StorageClass read shards at ~95 s each, exceeding the default 10m `AICR_INFERENCE_PERF_WORKLOAD_READY_TIMEOUT`. Passed after the documented `--data` catalog-entry override (ready timeout 25m, health timeout 10m, entry timeout 65m → 80m); both this and the faster-StorageClass alternative (`AICR_INFERENCE_PERF_MODEL_CACHE_STORAGE_CLASS`) are now documented in `aks-gpu-setup.md`, and the AKS Dynamo overlay's gate comment records the Azure baseline (gate values deliberately unchanged — conservative smoke thresholds). Doc + overlay-comment deltas on top of the fully-qualified head: scoped verification = yamllint (repo config) on the changed overlay, no heading/anchor changes in `aks-gpu-setup.md` (inbound links grepped), `go test ./pkg/uatbroker/...` green — comment/doc-only, no Go gates needed.
- Post-fix scoped gates: yamllint on all changed YAML; `go test -race ./pkg/recipe/...` green (`TestNodewrightTuningGateAcrossCatalog` unchanged — the new predicate is behavior-identical in-repo); example regenerated with the rebuilt CLI, sole delta the removed gdrcopy override (`tuningEnabled: false` retained); `make bom-docs` regen committed: the sole delta is the upstream `ubuntu:24.04` -> `26.04` digest drift in the GKE section (the gdrcopy flip produced no BOM delta); committed per the repo rule that a values-file PR carries the fresh BOM, since `bom-check` is not wired into CI.

**Cross-review round (parallel session, reconciled into the single commit):** the resolution-time warning now recommends the GPU-Operator-managed tuple only on AKS (generic reprovisioning guidance on GKE-COS/OKE, where the tuple is incoherent; the stacked #1819 supersedes this with the fully provider/OS-aware remedy); the `tuningEnabled` gate applies uniformly to the single-package tuning manifests (`tuning-gke.yaml`, `tuning-generic.yaml`) with a catalog-wide render test; `examples/recipes/aks-training.yaml` is curated (1732 → 229 lines, registry-hydrated `healthCheckAsserts` stripped, rendered bundle byte-identical); bundling.md documents the vendored post-install-hook lifecycle and nodewright.md the health-check gate limitation. The UAT registry gates Azure **release training cells** to v0.18.0 (same mechanism as the existing inference gate): release cells provision from main's `gpuDriverInstall: true` config while running a released binary that embeds the pre-flip operator-managed profile — a second driver stack on the Azure-managed pool would invalidate the nightly compatibility cell.

## Known caveats (dispositioned, non-blocking)

These were raised in review and considered; none can produce a broken supported configuration, and the two that could have been real were closed empirically by the aicr-test6 validation.

- **CDI spec generation with `toolkit.enabled=false`** — validated in practice: full three-phase validation passed with the toolkit disabled (device allocation works, `check-nvidia-smi` and `secure-accelerator-access` granted-container both pass). The only residual is the narrow CDI-*opt-in* path (`cdi.default: false`), which no default workload hits.
- **`nvidia-ctk config --set` compatibility across node images** — confirmed working with `nvidia-ctk` 1.18.2 on the tested image. On an image without it, the hardening DaemonSet's new readiness probe means the pod never reports Ready, so it fails visibly at deploy time rather than silently.
- **Boot-window race in the hardening DaemonSet** — a post-schedulable DaemonSet cannot close the window where an unprivileged `NVIDIA_VISIBLE_DEVICES=all` pod (which requests no `nvidia.com/gpu`) schedules before the first assert. Shipping this as interim defense-in-depth with the readiness probe; the durable, pre-schedulable fix requires a nodewright `nvidia-setup` package change plus an aicr pin bump, tracked with a two-repo plan in #1839.
- **UAT version gate is nightly-only** — the scheduled compatibility cells are gated (`training: v0.18.0`); a *manual* `uat-run` dispatch could still pair `gpuDriverInstall: true` with a pre-flip release, but that is deliberate operator action that fails loudly at deploy, not an automated-path defect.
- **`gdrdrv` on the Azure node image** — moot by default: GDRCopy is disabled (`gdrcopy.enabled: false`); GPUDirect RDMA rides the DMA-BUF path (validated at 157 GB/s). Whether the image ships `gdrdrv` only matters if a user opts GDRCopy back in.

## Risk Assessment

**Rollout notes:** Existing AKS deployments keep working; regenerated bundles now default to the driver-only profile. Clusters with `--gpu-driver none` pools must either be reprovisioned with the AKS default driver install or bundle with `--set gpuoperator:driver.enabled=true --set gpuoperator:toolkit.enabled=true --set gpuoperator:operator.runtimeClass=nvidia --set dradriver:nvidiaDriverRoot=/run/nvidia/driver`. Pre-change AKS recipes should be regenerated. The Azure UAT cluster needs its GPU pool recreated (see Implementation Notes).

## Checklist


docs/user/component-catalog.md
```diff
@@ -70,12 +70,13 @@ See the upstream [Topology Updater docs](https://kubernetes-sigs.github.io/node-
...
+
+The inverse mismatch **fails closed**: when the snapshot reports *no* NVIDIA driver loaded on the sampled GPU node but the resolved overlay declares the preinstalled-driver profile (for example, the AKS driver-only default resolved against a cluster whose GPU pools were created with `--gpu-driver none`), resolution returns an `INVALID_REQUEST` error instead of a recipe. Nothing on such a cluster provides a driver and the recipe does not install one, so any bundle generated from that combination would leave GPU nodes driverless. The error message names the two ways out: reprovision the GPU pools with the platform's default driver install, or bundle with the GPU-Operator-managed overrides (`--set gpuoperator:driver.enabled=true --set gpuoperator:toolkit.enabled=true --set gpuoperator:operator.runtimeClass=nvidia`).
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 0a33e7f3: the sentence now reads "The case is still tracked as separate work." The remaining "pre-installed" on that line is inside a backticked quote of the literal log message the auto-detect emits (`gpu-operator driver auto-detect: pre-installed driver observed ...` in pkg/client/v1/gpu_driver_state.go), so the doc keeps it verbatim; the section's own prose uses "preinstalled" consistently.

pkg/bundler/validations/checks.go
```diff
@@ -299,3 +304,81 @@ func CheckHostMofedWithoutNetworkOperator(ctx context.Context, componentName str
...
+	// An explicit --set gpuoperator:driver.enabled=true flips the bundle
+	// to the GPU-Operator-managed profile (the operator installs the
+	// driver), so the recipe's preinstalled assumption no longer applies.
+	// Value overrides supersede recipe values at render time.
+	if bundlerConfig != nil {
+		if gpuOpOverrides := bundlerConfig.ValueOverrides()["gpuoperator"]; gpuOpOverrides != nil {
+			if v, ok := gpuOpOverrides["driver.enabled"]; ok && v == overrideValueTrue {
+				return nil, nil
+			}
+		}
```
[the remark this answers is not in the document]
yuanchen8911: The flagged code has been removed from this PR: the bundle-time mismatch gate was descoped to keep this change focused on the default flip, and is tracked in #1757 (marked release-blocking for the next tag). #1757's scope explicitly requires exactly what this comment asks for: the gate clears only on the complete effective override tuple (driver.enabled, toolkit.enabled, operator.runtimeClass, DRA nvidiaDriverRoot), evaluated on final rendered values with --set keys resolved under both canonical component names and registry aliases.

pkg/client/v1/gpu_driver_state.go
```diff
@@ -289,12 +293,12 @@ func isDisambiguatedLabelKey(k string) bool {
...
 //
-// Bare AKS/EKS overlays lack this marker; auto-detect skips them so
+// Bare EKS overlays lack this marker; auto-detect skips them so
 // callers get a warning instead of a half-configured Operator (driver
 // off, toolkit + gdrcopy still on with no operator-managed driver
-// root). Fixing those cases requires a full preinstalled-profile
+// root). Fixing that case requires a full preinstalled-profile
 // overlay (tracked separately) — this gate keeps the current PR from
-// regressing them into a strictly worse state.
+// regressing it into a strictly worse state.
```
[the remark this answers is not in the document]
yuanchen8911: This gate guards only the auto-detect injection, and the injection is only-false and semantically idempotent: it fires exactly when the merged values already declare driver.enabled=false, so the rendered Helm output is identical with or without it. An overlay whose remaining settings (toolkit, gdrcopy, driver root) are incoherent with driver.enabled=false is broken with or without the injection - the injection cannot make it worse, so widening this marker check adds no protection at this point in the pipeline. It would also reject legitimate profiles: GKE-COS carries driver.enabled=false with toolkit.enabled=true (host-provided driver at /home/kubernetes/bin/nvidia, operator-managed toolkit), so requiring toolkit.enabled=false here would break its injection. Overlay-side coherence is enforced statically by TestDriverRootLockstep - strengthened in this PR so a preinstalled-driver overlay that inherits the container DRA root fails the suite - and effective-values enforcement at bundle time (toolkit, runtime class, driver root, including user --set overrides) is the scope of #1757.

pkg/recipe/driver_root_lockstep_test.go
```diff
@@ -162,8 +162,25 @@ func TestDriverRootLockstep(t *testing.T) {
...
+					t.Errorf(
+						"overlay %q: gpu-operator has driver.enabled=false (preinstalled-driver profile)\n"+
+							"  but nvidia-dra-driver-gpu.nvidiaDriverRoot = %q.\n"+
+							"  /run/nvidia/driver is populated by the operator's driver container, which this\n"+
+							"  profile disables — the DRA kubelet plugin would read an empty driver root.\n"+
+							"  Set nvidiaDriverRoot to the host driver location (\"/\" for drivers installed\n"+
+							"  at the host root; GKE-COS uses /home/kubernetes/bin/nvidia).",
+						name, draRoot)
+					return
+				}
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 0a33e7f3: the test now normalizes both nvidiaDriverRoot and hostPaths.driverInstallDir with path.Clean before comparing (empty string preserved, since "unset" is treated differently from any real path), so a trailing-slash spelling can no longer dodge the container-root rejection or the lockstep equality.

tests/uat/azure/cluster-config.yaml
```diff
@@ -63,7 +63,12 @@ compute:
             nodeGroup: cpu-worker
         - name: gpuworker1
           gpuType: h100  # NDSH100v5 family: 2x96=192 of 1152 vCPUs
-          gpuDriverInstall: false  # no Azure driver — GPU software managed in-cluster (GPU Operator)
+          # AKS driver-only profile (the Azure default): the node image
+          # preinstalls the NVIDIA driver + container toolkit, matching the
+          # AKS recipe defaults (gpu-operator driver.enabled=false,
+          # toolkit.enabled=false). Existing pools provisioned with
+          # gpuDriverInstall: false must be recreated to pick this up.
+          gpuDriverInstall: true
```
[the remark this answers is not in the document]
yuanchen8911: Verified: the UAT pipeline provisions a fresh cluster per run - uat-azure.yaml names it aicr-<run_id>, the Bringup Infra step creates it from this cluster-config.yaml, and the always() teardown deletes it (the nightly batch always deletes; skip_delete is manual-only). So there is no standing gpuworker1 pool to retrofit - the next provision consumes gpuDriverInstall: true. The immutable-profile caveat applies to long-lived manually-kept clusters, which is why the PR notes the existing demo cluster's GPU pool must be recreated before the next Azure UAT-style run against it.

pkg/client/v1/aicr_test.go
```diff
@@ -904,13 +914,13 @@ func TestResolveRecipeFromSnapshot_GPUDriverAutoDetect(t *testing.T) {
...
 			snap:         gpuHardwareSnapshot(true),
-			wantInjected: false,
+			wantInjected: true,
 		},
 		{
-			name:         "aks + absent snapshot leaves defaults alone",
+			name:         "aks + absent snapshot warns, leaves defaults alone",
 			service:      "aks",
 			snap:         gpuHardwareSnapshot(false),
 			wantInjected: false,
```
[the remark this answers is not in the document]
yuanchen8911: Asserting the log text here has a structural problem: the warning is emitted through the process-global slog default, and TestResolveRecipeFromSnapshot_GPUDriverAutoDetect runs t.Parallel() with parallel subtests - capturing would mean swapping slog.SetDefault mid-run, which races with every other concurrently-running test that logs (nothing in pkg/client/v1 captures slog today for the same reason). Making the logger injectable just to assert this text is an API change disproportionate to the nitpick.

The durable fix is already scoped in #1757: the same mismatch condition becomes a blocking bundle-time validation that returns an error, which is directly assertable without log capture - that is what makes the safety behavior regression-proof, rather than pinning a warning string. The behavioral contract this PR introduces (absent + preinstalled profile -> no injection, defaults intact) is asserted by the existing rows.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
```diff
@@ -0,0 +1,162 @@
...
+# Deliberately NO readiness/startup probe: the pod reports Ready immediately
+# and re-asserts in-loop, matching nvidia-peermem-reloader.
+#
+# Boot-window caveat: because the keys are asserted only after this pod's
+# first loop iteration, there is a narrow window on a freshly booted or
+# reimaged node where the permissive node-image defaults are still live. In
+# the normal flow this window is not reachable by a workload — a GPU pod
+# cannot schedule until the device plugin (a gpu-operator operand that starts
+# alongside this DaemonSet) advertises nvidia.com/gpu, which lands after this
+# DaemonSet's first assert — but it is not closed by construction. The durable
```
mchmarny: [high] This hardening path fails open. The unauthorized pod described above does not request `nvidia.com/gpu`, so device-plugin advertisement is not a scheduling barrier: it can run as soon as the node is Ready, before this DaemonSet's first write. If `nvidia-ctk` is missing or incompatible, the loop also swallows the error forever while the hardening pod remains Ready. Please land the pre-schedulable nodewright assertion tracked in #1839 before making this the default, or add a fail-closed node taint/admission barrier that is removed only after both keys are verified; a readiness probe should also expose persistent assertion failures.

recipes/overlays/h100-aks-inference.yaml
```diff
@@ -43,8 +43,11 @@ spec:
 
     # Same tuning manifest as EKS: nvidia-setup 0.5.0 supports service: aks
     # (aks-h100), covering host IB RDMA + memlock setup that the
-    # ib-node-config-aks DaemonSet used to perform, and nvidia-tuned applies
-    # the aks-h100-inference profile.
+    # ib-node-config-aks DaemonSet used to perform. nvidia-tuned (which would
+    # apply the aks-h100-inference profile) is disabled by default under the
+    # Azure-managed driver profile — no node reboot, and the tuned profile's
```
mchmarny: [medium] Disabling `nvidia-tuned` does not mean "no node reboot": `tuning.yaml` still renders `nvidia-setup-full`, whose `interrupt.type` is `reboot`. Please say "no `nvidia-tuned`-triggered reboot". The same wording issue appears in `recipes/overlays/a100-aks-training.yaml:61-62`.

pkg/client/v1/aicr_test.go
```diff
@@ -884,18 +884,28 @@ func gpuOperatorDriverEnabled(t *testing.T, rec *aicr.RecipeResult) (bool, bool)
...
-//     driver.enabled at chart default true), so the gate refuses to
-//     land a lone driver.enabled=false and logs a warning instead.
-//     The bug fix waits on a full AKS driver-only-install overlay.
-//   - EKS + Preinstalled → SKIPPED for the same reason (base
-//     values.yaml has driver.enabled=true).
+//   - AKS + Preinstalled → INJECTED. The AKS values files declare the
+//     coordinated AKS driver-only profile (driver.enabled=false,
+//     toolkit.enabled=false, operator.runtimeClass) — the Azure
+//     default, where the node image preinstalls driver + toolkit — so
+//     the override is semantically idempotent and safe.
```
mchmarny: [medium] This comment says the test covers the coordinated AKS profile, but the assertions below observe only the `driver.enabled` injection. A regression that re-enables the toolkit, changes `operator.runtimeClass`, leaves GDRCopy enabled, or moves the DRA root would still pass. Add a catalog/recipe test that resolves both AKS values files and asserts the complete tuple: driver off, toolkit off, `nvidia-container-runtime`, GDRCopy off, and DRA root `/`.

mchmarny (changes requested):
This changes AKS to the Azure-managed driver profile and adds the supporting hardening/runtime manifests. I found one merge-blocking fail-open security issue in the toolkit hardening path, plus two medium correctness and coverage issues. Please address the security barrier before merging; #1757 also remains release-blocking before the next tag.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
yuanchen8911: Addressed the visibility half and documented the rest honestly at 0ba777a0.

- Readiness probe added: the loop touches a sentinel file only after a successful `nvidia-ctk config --in-place`, and the probe (`test -f /tmp/hardened`) gates Ready on it. A node where nvidia-ctk is missing or incompatible never reports Ready, so a persistently-failing assert surfaces at deploy time instead of a silently-green DaemonSet.
- The manifest comment now states plainly that the exploit pod requests no `nvidia.com/gpu` and can schedule before the first assert; a post-schedulable DaemonSet cannot close that window by construction.

I did not add a fail-closed taint/admission barrier: a DaemonSet that removes a NoSchedule taint only after asserting fights the scheduler and is fragile, and admission is out of scope for this component.

Decision: shipping this DaemonSet as interim defense-in-depth (it closes the permanent exposure), with the boot-window race documented as a known residual. The durable, race-free fix is asserting the toolkit-config keys during Skyhook host prep while the node is cordoned and not yet workload-schedulable — that lives in the `nvidia-setup` package. Because `nvidia-setup` is an external digest-pinned image (`ghcr.io/nvidia/nodewright-packages/nvidia-setup`), it needs a change in BOTH repos: the nodewright-packages `nvidia-setup` package must run the `nvidia-ctk config --set …` step, then this repo bumps the pin and removes this DaemonSet. Steps written up in #1839.

recipes/overlays/h100-aks-inference.yaml
yuanchen8911: Fixed at 0ba777a0. Reworded to "no nvidia-tuned-triggered reboot (nvidia-setup-full still reboots for host prep)" in `h100-aks-inference.yaml`, `a100-aks-training.yaml`, `h100-aks-training.yaml`, and the `tuning.yaml` gate comment — nvidia-setup-full's `interrupt.type: reboot` still applies.

pkg/client/v1/aicr_test.go
yuanchen8911: Added `TestAKSDriverOnlyProfileTuple` (pkg/recipe) at 0ba777a0: it resolves every AKS leaf and asserts the full gpu-operator tuple — `driver.enabled=false`, `toolkit.enabled=false`, `gdrcopy.enabled=false`, and `operator.runtimeClass=nvidia-container-runtime`. The DRA driver-root half (`nvidiaDriverRoot=/`) is enforced on effective values by `TestDriverRootLockstep`. Also tightened the overclaiming comment here to point at both tests.

docs/integrator/aks-gpu-setup.md
```diff
@@ -144,31 +148,156 @@ its own device plugin, DCGM exporter, and GPU health tooling, which duplicate
...
+--in-place`. The toolkit re-reads its config at every container start, so no
+containerd restart is needed and running pods are unaffected; the pair closes
+the env-var path while keeping legitimately allocated pods working (they
+receive devices via CDI / the device plugin's volume-mounts strategy, not the
+env var). Migrating this host-config hardening into the nodewright
+`nvidia-setup` package — the canonical host-configuration channel on AKS — and
+retiring the DaemonSet is tracked in
+[#1839](https://github.com/NVIDIA/aicr/issues/1839).
+
+**Migration note for pre-existing AKS recipes:** a recipe file references the
```
njhensley, line 315: **🟡 Minor — Only the plain-bundle re-resolve path is unwarned until #1757**

A pre-flip AKS `recipe.yaml` re-bundled via plain `aicr bundle recipe.yaml` (no `--snapshot`) resolves the new `driver.enabled=false` values while keeping its baked `dradriver:nvidiaDriverRoot=/run/nvidia/driver` override, producing an incoherent bundle. The `--snapshot` path warns; this path relies only on the docs migration note and the deploy-time gpu-operator-health backstop. Retiered major→minor because it fails closed at deploy and #1757 (the blocking coherence guard) is release-blocking to the same tag.

**Blast radius:** AKS users who committed a recipe.yaml before the flip and re-bundle without regenerating — broken GPU stack surfaced at deploy/validate, not silently shipped.

**Fix:** Safe to proceed *iff* #1757 is confirmed release-blocking to the same tag as the flip — worth stating that explicitly on the PR/issue. Optional: emit the same driver/DRA-root coherence warning on the no-snapshot bundle path to close the one unwarned route now.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
```diff
@@ -0,0 +1,176 @@
...
+              while true; do
+                if chroot /host nvidia-ctk config --in-place \
+                     --set accept-nvidia-visible-devices-envvar-when-unprivileged=false \
+                     --set accept-nvidia-visible-devices-as-volume-mounts=true >/dev/null 2>&1; then
+                  cur=hardened
+                  # Sentinel for the readiness probe — only touched after a
+                  # successful assert, so the pod reports Ready only once the
+                  # toolkit config has actually been hardened at least once.
+                  touch /tmp/hardened
+                else
```
njhensley, line 156: **🟡 Minor — Readiness probe is a one-shot latch**

The loop only ever `touch /tmp/hardened` on success (L138) and never removes it, so the `test -f /tmp/hardened` probe (L154) reports Ready permanently after the *first* success — even if a later in-loop re-assert fails (e.g. config.toml rewritten permissive without a pod restart and nvidia-ctk then failing).

**Blast radius:** A node whose toolkit config silently reverts after initial hardening keeps advertising Ready / passing secure-accelerator-access while the env-var isolation is off. Bounded but silent on multi-tenant nodes.

**Fix:** Add `rm -f /tmp/hardened` in the `else` branch (L139) so readiness tracks the *current* assert result, turning the probe into a live health signal at zero extra cost (the loop already re-asserts every 60s).

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
```diff
@@ -0,0 +1,176 @@
...
+# This DaemonSet re-asserts the hardened pair via `nvidia-ctk config
+# --in-place` (TOML-safe; the preinstalled toolkit ships nvidia-ctk). The
+# toolkit re-reads config.toml at every container start, so no containerd
+# restart is needed and already-running pods are unaffected. The loop
+# re-asserts after a node reimage/upgrade rewrites the config, mirroring the
+# nvidia-peermem-reloader design in this component set.
+#
+#   accept-nvidia-visible-devices-envvar-when-unprivileged = false
+#     Ignore the NVIDIA_VISIBLE_DEVICES env var for unprivileged containers —
+#     closes the unauthorized-GPU path.
```
njhensley: **🔵 Nitpick — Header over-claims "closes the unauthorized-GPU path"**

With `accept-nvidia-visible-devices-as-volume-mounts=true` (required for the device-plugin volume-mounts allocation strategy), an unprivileged pod mounting an emptyDir at `/var/run/nvidia-container-devices/all` can still request all GPUs unallocated. This equals GPU-Operator-managed mode's posture — not a regression and not closable here — but the header claims the env-var key closes "the unauthorized-GPU path" when it closes one of two.

**Blast radius:** Operator reading the header/conformance-green may over-trust isolation; the volume-mount request path remains open absent an admission policy.

**Fix:** Reword to "closes the env-var path; the volume-mounts path remains, same as GPU-Operator-managed mode" and note that full multi-tenant isolation needs an admission policy forbidding mounts under `/var/run/nvidia-container-devices`.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
```diff
@@ -0,0 +1,176 @@
...
+                  else
+                    echo "nvidia-ctk not available yet (toolkit not installed); retrying every 60s"
+                  fi
+                  prev=$cur
+                fi
+                sleep 60
+              done
+          readinessProbe:
+            exec:
+              command: ["test", "-f", "/tmp/hardened"]
```
njhensley, line 176: **🔵 Nitpick — toolkit-hardening readinessProbe reintroduces the helm-wait hazard the peermem reloader avoids (PLAUSIBLE)**

The sibling `nvidia-peermem-reloader` deliberately omits any probe to avoid deadlocking helm-wait under the GPU-Operator-managed fallback (tool/module appears only after gpu-operator installs). This DS's probe has the same exposure: under `--gpu-driver none`, `nvidia-ctk` is not preinstalled, so `/tmp/hardened` is never touched and the probe can hold the gpu-operator wait until timeout. Default driver-only profile is unaffected (nvidia-ctk preinstalled, probe passes fast).

**Blast radius:** Only the non-default GPU-Operator-managed fallback on AKS, and only if the deployer helm-waits on the gpu-operator post-manifests. PLAUSIBLE — verify against your deployer's wait behavior.

**Fix:** Drop/relax the readinessProbe (matching the peermem reloader rationale) or gate it so it cannot block the managed-fallback helm-wait; at minimum document the asymmetry in the header.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
```diff
@@ -0,0 +1,176 @@
...
+            initialDelaySeconds: 2
+            periodSeconds: 10
+          securityContext:
+            privileged: true
```
njhensley, line 180: **🔵 Nitpick — privileged:true where scoped capabilities may suffice**

Both new DaemonSets use `privileged: true`. Least-privilege: `SYS_CHROOT` (this DS) and `SYS_CHROOT`+`SYS_MODULE` (nvidia-peermem-reloader.yaml L112) plus the existing hostPath *may* suffice. Matches in-repo (gke-nccl-tcpxo) and Azure reference precedent, so not a defect.

**Blast radius:** Widened host attack surface on every AKS GPU/IB node if the busybox image/supply chain is ever compromised.

**Fix:** If pursued, test `capabilities.add: [SYS_CHROOT]` / `[SYS_CHROOT, SYS_MODULE]` with `privileged:false` on a real AKS ND-H100 node before adopting — chroot'd host binaries may need more caps or interact with AppArmor. Do not change blind.

pkg/bundler/deployer/localformat/vendor_folder.go
```diff
@@ -238,15 +238,15 @@ func writeMixedManifests(ctx context.Context, folderDir, dir string, c Component
...
 	templatesDir, err := deployer.SafeJoin(folderDir, "templates")
 	if err != nil {
 		return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
 			"templates dir path unsafe", err)
 	}
-	if err = os.MkdirAll(templatesDir, 0o755); err != nil {
-		return nil, errors.Wrap(errors.ErrCodeInternal,
-			fmt.Sprintf("create templates dir for %s", dir), err)
-	}
+	templatesDirCreated := false
```
njhensley, line 249: **🔵 Nitpick — Mixed-manifest lazy-templates/ path is untested**

The identical lazy `templates/` creation landed in both `writeLocalHelmFolder` (covered by the new `empty_render_test.go`) and `writeMixedManifests` here — but the latter is only exercised with a non-empty manifest, so the all-empty branch (`templatesDirCreated` stays false, no empty dir) is untested. No live defect (structurally identical, all-empty mixed case not currently reachable).

**Blast radius:** A future regression re-adding an unconditional MkdirAll on the mixed path would re-trigger the inventory-verifier abort and not be caught.

**Fix:** Add a `writeMixedManifests` all-empty-render test asserting no `templates/` dir remains, mirroring `TestWriteLocalHelmFolder_AllManifestsEmpty`.

recipes/components/gpu-operator/values-aks.yaml
```diff
@@ -15,12 +15,19 @@
...
+# node image preinstalls the NVIDIA driver and container toolkit, so the
+# GPU Operator runs with driver.enabled=false, toolkit.enabled=false, and
+# operator.runtimeClass=nvidia-container-runtime — always the three
+# together, never just one of them.
+# To let the GPU Operator manage the driver instead (nodepools created
+# with `--gpu-driver none`), flip all of these at bundle time:
+#   --set gpuoperator:driver.enabled=true \
+#   --set gpuoperator:toolkit.enabled=true \
+#   --set gpuoperator:operator.runtimeClass=nvidia \
+#   --set dradriver:nvidiaDriverRoot=/run/nvidia/driver
```
njhensley, line 29: **🔵 Nitpick — Flip tuple omits gpuoperator:hostPaths.driverInstallDir**

The documented GPU-Operator-managed override tuple sets `dradriver:nvidiaDriverRoot=/run/nvidia/driver` but not `driverInstallDir`, and `TestDriverRootLockstep` skips the equality check for AKS (resolves to driver.enabled=false). Correction to the raised concern: this is NOT fragile against a chart-default bump — AICR's base `values.yaml` pins `hostPaths.driverInstallDir=/run/nvidia/driver`, so the two stay in lockstep. The only residual is that the manual `--set` flip path is unexercised by the lockstep test.

**Blast radius:** None today; affects only the non-default managed fallback if base values.yaml ever stops pinning driverInstallDir.

**Fix:** Optional: add the override for explicitness, or note in docs that driverInstallDir is already pinned by base values.yaml so no override is needed.

njhensley (dismissed):
## Multi-persona review — approve with comments

**Method.** Four independent persona reviewers (Correctness · Security · Domain/Architecture · Operability/CI-DX) ran in parallel against head `0ba777a`, then every finding was routed through an adversarial senior meta-reviewer that re-derived it from source. Templates were rendered through the real `manifest.Render` to confirm truth tables; docs anchors, BOM entries, and UAT gates were cross-checked against the actual files.

### Overall

A large, unusually well-documented breaking change (AKS driver-ownership default flip), live-validated end-to-end on AKS H100 across all four intents. The core mechanics hold up: the driver-only tuple is coherent across base + training overlays, the Sprig-safe template gates have correct truth tables (`absent→true, false→false, true→true`), EKS renders are byte-identical, the dependency graph stays acyclic, and the two new privileged DaemonSets are correctly node-scoped with no injection surface and digest-pinned images.

**No blocker or major finding survived independent reproduction.** The persona-raised "major" (shipping the flip before the coherence guard) was retiered to minor because the incoherent path *fails closed at deploy* (gpu-operator-health / DRA CDI generation), is guarded by a docs migration note + snapshot-path warning, and the blocking guard (#1757) is release-blocking to the same v0.18.0 tag.

The one item worth a reply before merge is the sequencing condition (first inline comment) — confirm #1757 is release-blocking to the same tag as the flip. Everything else is optional polish; the readiness-latch `rm -f` (second inline comment) is the one I'd most encourage folding in as a one-line robustness fix.

### Confirmed non-issues (examined, refuted, or fully pre-documented)

- **`applyGPUDriverAutoOverride` control flow** — all 4 states × profile combinations traced; no fall-through, AKS branch nil-guarded, remediation `--set` keys match `registry.yaml`.
- **Template gates** — `tuningEnabled`/`autoTaintNewNodes` truth tables verified via real render; every branch is valid YAML; no dangling `dependsOn` on omitted `nvidia-tuned`; EKS output byte-identical; whole-CR suppression correct for the single-package manifests.
- **Driver-only tuple coherence** — base + training values both restate the full profile; `nfd.enabled=false` correctly inherited from generic `values.yaml`; `driver.rdma.enabled=true` under `driver.enabled=false` coherently inert with the peermem-reloader substitute; `gdrcopy` overrides fully removed from leaves.
- **Boot-window race / per-boot conformance masking** — fully and accurately documented in the manifest header (#1839 durable fix); no undisclosed signal.
- **DaemonSet security hygiene** — static chroot commands, no injection; digest-pinned busybox (same digest both manifests); node scopings correct (`pci-15b3` / `kubernetes.azure.com/accelerator`); drain-exemption nullifies the wildcard-toleration hazard; `toolkit.enabled=false` drops no isolation control beyond the two restored keys.
- **Docs / anchors / BOM / UAT** — override tuples byte-match across doc/values/code; all intra-repo anchors resolve (no dangling old slug); BOM regenerated with correct busybox entries and counts; `training: v0.18.0` UAT gate correctly mirrors the inference gate.
- **Lazy `templatesDir`** — no path leaves an empty `templates/`; template-less charts install as harmless no-ops.

### Summary

|:---:|:---:|:---:|:---:|
| 0 | 0 | 2 | 5 |

<sub>🤖 Reviewed with a multi-persona + adversarial meta-review harness.</sub>

docs/integrator/aks-gpu-setup.md
yuanchen8911: Disposition unchanged and confirmed. #1757 is release-blocking and implemented by the stacked #1819; the PR body now states #1819 must merge in the same tagged release as the flip (not merely be labeled), which is the condition to proceed. Kept the no-snapshot bundle coherence warning as #1757/#1819's scope rather than duplicating it on this path.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
yuanchen8911: Fixed at 85449d5e. The loop now removes /tmp/hardened on any assert failure, so the probe tracks the current assert result — a reverted or failing config drops the pod out of Ready instead of latching Ready permanently after the first success.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
yuanchen8911: Reworded at 85449d5e (manifest header + aks-gpu-setup.md). Setting the env-var key to false closes the env-var path (the one secure-accelerator-access exercises); accept-nvidia-visible-devices-as-volume-mounts=true is required for the device-plugin volume-mounts allocation strategy but leaves the volume-mounts device-request path open — same posture as GPU-Operator-managed mode. The docs now state that full multi-tenant isolation additionally requires an admission policy restricting user-declared mount destinations under /var/run/nvidia-container-devices, and that a green secure-accelerator-access is not proof of complete isolation.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
yuanchen8911: Fixed at 85449d5e. The DaemonSet now renders only when gpu-operator.toolkit.enabled==false, so it is omitted entirely in the --gpu-driver none managed fallback where nvidia-ctk is not preinstalled — no probe there, no helm-wait hazard. The fallback is instead hardened via gpu-operator.toolkit.env (ACCEPT_NVIDIA_VISIBLE_DEVICES_ENVVAR_WHEN_UNPRIVILEGED=false, ACCEPT_NVIDIA_VISIBLE_DEVICES_AS_VOLUME_MOUNTS=true in values-aks*.yaml), which the operator toolkit installer writes into config.toml. Verified: driver-only renders the DaemonSet, managed fallback omits it and carries the env vars.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
yuanchen8911: Filed as a follow-up: #1848 — test SYS_CHROOT (hardening) and SYS_CHROOT+SYS_MODULE (peermem-reloader) with privileged:false on a real AKS ND-H100 node before adopting, per your "do not change blind" note.

pkg/bundler/deployer/localformat/vendor_folder.go
yuanchen8911: Added at 85449d5e: TestWriteMixedManifests_AllManifestsEmpty asserts the vendored/mixed path (writeMixedManifests) leaves no templates/ dir and returns an empty path list on an all-empty render, mirroring TestWriteLocalHelmFolder_AllManifestsEmpty.

recipes/components/gpu-operator/values-aks.yaml
yuanchen8911: Added a doc note at 85449d5e: hostPaths.driverInstallDir is already pinned to /run/nvidia/driver by base values.yaml, so the GPU-Operator-managed override tuple needs no explicit driverInstallDir — it stays in lockstep with the DRA driver root.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
yuanchen8911: Follow-up: the hardening path was strengthened further at 85449d5e, addressing more of the original concern:

- The readiness probe is now fail-closed — the loop removes /tmp/hardened on any assert failure, so a reverted or failing config drops the pod out of Ready rather than latching Ready after the first success.
- The DaemonSet now renders only when gpu-operator.toolkit.enabled==false (driver-only). Under the --gpu-driver none managed fallback it is omitted entirely, so it cannot hold the gpu-operator-post helm --wait on nodes where nvidia-ctk is not preinstalled; the fallback is hardened instead via gpu-operator.toolkit.env (ACCEPT_NVIDIA_VISIBLE_DEVICES_* in values-aks*.yaml).

The boot-window race itself still can't be closed by a post-schedulable DaemonSet by construction — the durable pre-schedulable fix (nodewright nvidia-setup, node cordoned) is #1839, and we're shipping this as interim defense-in-depth per the earlier disposition.

recipes/components/gpu-operator/values-aks.yaml
njhensley, line 29: **🔵 Nitpick — Flip tuple omits gpuoperator:hostPaths.driverInstallDir (optional, carried over)**

The documented GPU-Operator-managed override tuple sets dradriver:nvidiaDriverRoot=/run/nvidia/driver but not driverInstallDir. This is benign — AICR's base values.yaml already pins hostPaths.driverInstallDir=/run/nvidia/driver, so the two stay in lockstep and it is NOT fragile against a chart-default bump. Only residual is that the manual --set flip path is unexercised by TestDriverRootLockstep.

**Blast radius:** None today; non-default managed fallback only.

**Fix:** Optional: add the override for explicitness, or note in docs that driverInstallDir is already pinned by base values.yaml so no override is needed. Not blocking.

recipes/components/gpu-operator/manifests/nvidia-toolkit-hardening-aks.yaml
```diff
@@ -0,0 +1,199 @@
...
```
njhensley, line 180: **🔵 Nitpick — privileged:true where scoped capabilities may suffice (optional, carried over)**

Both new DaemonSets use privileged: true. Least-privilege: SYS_CHROOT (this DS) and SYS_CHROOT+SYS_MODULE (nvidia-peermem-reloader.yaml L112) plus the existing hostPath may suffice. Matches in-repo (gke-nccl-tcpxo) and Azure reference precedent, so not a defect.

**Blast radius:** Widened host attack surface if the busybox image/supply chain is ever compromised.

**Fix:** Only if pursued: bench-test capabilities.add:[SYS_CHROOT] / [SYS_CHROOT,SYS_MODULE] with privileged:false on a real AKS ND-H100 node before adopting — chroot'd host binaries may need more caps or interact with AppArmor. Do not change blind. Not blocking.

njhensley (approved):
## Re-review @ `ce93ba1` — approve holds ✅

Re-reviewed the revision pushed since the first pass (`0ba777a` → `ce93ba1`). It addresses the review directly, with tests. **No blocker / major / minor.** Approval stands.

**Prior findings → resolved:**

- 🟡 **Readiness probe one-shot latch** — fixed: `rm -f /tmp/hardened` in the else branch, fail-closed semantics (a transient valid-but-permissive rewrite is repaired by the next re-assert, not flapped). Pinned by `TestToolkitHardeningGate`.
- 🔵 **Header over-claim ("closes the unauthorized-GPU path")** — reworded to the ENV-VAR path; the volume-mounts path is now explicitly documented as open (with the toolkit-v1.19.1 `/dev/null`-source detail + admission-policy guidance), and green conformance is noted as not proof of complete isolation.
- 🔵 **`readinessProbe` could deadlock the managed-fallback helm-wait** — fixed cleanly: the DaemonSet now renders only when `toolkit.enabled=false`, so it's omitted under the GPU-Operator-managed fallback; that profile is hardened instead via `toolkit.env` (`ACCEPT_NVIDIA_VISIBLE_DEVICES_*`). `TestAKSToolkitEnvHardened` asserts the exact name→value pairs.
- 🔵 **Untested mixed-manifest empty-render path** — fixed: `TestWriteMixedManifests_AllManifestsEmpty` (now load-bearing, since the managed fallback makes the hardening manifest render empty).
- Bonus: `hasPreinstalledDriverProfile` now warns on a values read error instead of silently suppressing the absent-driver mismatch warning.

**New in this revision (verified):** the tuning Skyhook CRs are now release-managed (Helm hooks removed → `aicr/skip-hook-validation`), resolving the Flux-orphaning limitation (was #1842). Confirmed safe: `stripHelmHooks` already strips hook annotations at bundle time, so helm/argo output is unchanged and Flux now prunes correctly on a `tuningEnabled` flip; CRD ordering is preserved via the `nodewright-customizations → nodewright-operator` dependency. Docs updated thoroughly, including the one-time Flux migration note.

The two inline comments below are optional, non-blocking nits carried over from the first pass (both already benign). The sequencing condition from the first review still applies: #1757 (bundle-time coherence guard) should be release-blocking to the same tag as this flip.

<sub>🤖 Re-reviewed with a multi-persona + adversarial meta-review harness.</sub>

mchmarny (approved):

Issue #1761, opened 2026-07-15
Implement ADR-015: recipe-declared configuration profiles


yuanchen8911:
## Summary

Implement the mechanism specified in ADR-015 (`docs/design/015-recipe-configuration-profiles.md` — **Accepted 2026-07-21**, merged via PR #1762): a single enumerated configuration profile per resolved recipe, selected at recipe generation, persisted in the hydrated artifact as a compact ownership record, gated behind a new recipe artifact apiVersion, and enforced at every recipe boundary.

## Scope (per the final ADR — read it first; it is authoritative)

- **Profile declaration & resolution** (`pkg/recipe`):
  - Singular `spec.profile` declaration (`name`, `description`, **mandatory `default` naming a declared value — catalog-validated**, `values`).
  - **Closed per-value fragment schema** — a value permits `advertiser`, `constraints`, and `componentRefs` only; each ref permits only `name` and `overrides` (every other `ComponentRef` field is rejected at catalog load). The admission criterion: a fragment field is permitted only when its effect is either representable in `ownedPaths` (value paths plus the synthetic `enabled`) or persisted verbatim in the artifact with no bundle-time override surface (`constraints` join the artifact's constraint set; `advertiser` is recorded in `selectedProfile`) — an effect the artifact cannot carry is an effect the lock cannot protect. Every other `ComponentRef` field fails both tests:
    - **`valuesFile`** — additionally rejected for a merge-semantics reason: a later `valuesFile` *replaces* the earlier one rather than merging, so a profile could change effective values by *removing* a file — through paths that appear nowhere in the fragment and that no syntactic union could lock. Deferred design sketch, should a consumer ever require it: derive owned paths from each value's semantic **before/after delta** rather than from fragment syntax.
    - **chart/version/tag/source fields**; **`namespace`**; **`patches`**; **health-check / cleanup fields**.
    - **`manifestFiles` / `preManifestFiles`** (raw-manifest injection) — shares `valuesFile`'s unrepresentable-effect hazard.
  - Fragments may reference **only components enabled in the pre-profile resolved composition** — membership validated at resolution step 4, against the post-filtering surviving composition, before fragment application, and **declaration-wide**: every component referenced by *any* declared value (not only the selected one) must be enabled, because `ownedPaths` is declaration-wide and a selection must never record lock paths for a component absent from its composition (a snapshot exclusion disabling any referenced component fails generation).
[...]

Issue #1755, opened 2026-07-14
validate: fail-closed check for GKE device-plugin ownership conflict


yuanchen8911:
## Problem

On GKE, default GPU node-pool provisioning leaves the `gke-no-default-nvidia-gpu-device-plugin` label absent, so GKE's managed `nvidia-gpu-device-plugin` DaemonSet runs on every GPU node and advertises `nvidia.com/gpu`. AICR's stock GKE recipes ship gpu-operator with `devicePlugin.enabled: true`, so deploying a bundle onto an unlabeled (default-provisioned) cluster produces two device plugins registering the same `nvidia.com/gpu` resource on each GPU node.

Kubelet maintains one endpoint and one device inventory per resource name ([device manager](https://github.com/kubernetes/kubernetes/blob/v1.33.0/pkg/kubelet/cm/devicemanager/manager.go) keys `endpoints[resourceName]`), so competing registrations and ListAndWatch updates replace each other — nondeterministic ownership, and potentially one plugin's device IDs (GKE: `nvidia0`-style names; NVIDIA: GPU UUIDs) passed to the other plugin's `Allocate` implementation. This can cause allocation and runtime failures; it is not a benign overlap.

AICR's stock profile intends the operator's plugin to be the sole advertiser, which requires GPU node pools to carry GKE's documented opt-out label `gke-no-default-nvidia-gpu-device-plugin=true` (see [NVIDIA GPU Operator GKE guidance](https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/google-gke.html)).

## Current gap — nothing detects the conflict

- `aicr bundle` is offline by design (recipes define *what*, independent of cluster state) and cannot see node labels.
- `aicr validate` has no guard: the GKE overlays' only constraint is `K8s.server.version`, and the generic constraint syntax cannot express "every GPU node carries this label" — `NodeTopology.label` readings encode `value|node1,node2,...` (with possible truncation) while the evaluator compares scalar readings. There is no way to cross-reference the GPU-node set against the label's node list.
- Deployment-phase health checks pass (the operator's device-plugin DaemonSet is healthy); the conflict surfaces only as incidental, nondeterministic workload failures with no pointer to the cause.

So a user pointing AICR at a default-provisioned GKE cluster gets a clean bundle and a validation run with no deterministic guard, and ships a latent plugin-ownership conflict.

## Proposed solution

[...]

PR #1762, opened 2026-07-15
docs(design): ADR-015 recipe-declared configuration profiles

yuanchen8911:
## Summary

Adds ADR-015: recipe-declared configuration profiles. An overlay may declare a single named profile enum (e.g. `gpuStack: operator | csp-managed`); exactly one value is selected at recipe generation and hydrated into the artifact with a compact ownership record — so multiple qualified cluster configurations per criteria combination stay validated against the recipe. The design is deliberately minimal:

- **Existing fragment syntax only** (`componentRefs[].overrides` + `constraints`) — no new templating, no parameter language.
- **No component-set mutation**: fragments only re-configure components already enabled in the composition; conditional installation is expressed through values-gated components (chart renders nothing unless the value selects it), and true addition/removal stays a future amendment.
- **Byte-identical behavior for unprofiled recipes**, digests included — nothing changes until a family explicitly declares a profile.
- **Explicit conversion boundaries**: converting a family is a documented qualification + cut-over event (new artifact apiVersion, regenerated evidence, `/v1`→`/v2` client migration), never an implicit upgrade.
- **One authoritative output invariant**: before any output is written, the effective configuration must equal the selected recipe at every locked path and presence state, and no output surface may expose a locked path as install-time-mutable; per-surface checks are diagnostics only.

## Motivation / Context

One criteria combination increasingly maps to more than one valid cluster configuration (GKE device-plugin ownership, AKS driver-only vs operator-managed pools). Today the only outlets are bundle-time `--set` (invisible to validation — the deployed config silently diverges from what `aicr validate` checks) or new criteria axes (recipe-matrix growth).

The ADR chooses a middle path. It specifies the required amendments to the #1327 allocation-policy model (recipe-declared external advertiser, recomputed complete-tuple locking, hydration-boundary coherence checks, one dependency-neutral descriptor + evaluator) and an explicit ADR-011 amendment (profile-bearing recipes ride a new artifact apiVersion so released binaries fail closed instead of silently dropping the profile).

Motivating examples (both specified in the ADR):

- **GKE GPU-stack ownership, three values** — `operator` (GKE driver + Operator's plugin; stock, opt-out label required), `csp-managed` (GKE driver + GKE's managed plugin; `advertiser: external`, opposite label constraint), and `operator-selfdriver` (driver installed by a new values-gated `gcp-driver-installer` component + Operator's plugin — the DGXC/NKX case, declared once its durable ownership-mode distinguishing signal is identified).
- **AKS unmanaged-pool driver/toolkit ownership** — GPU-Operator-managed vs the AKS "Driver only" install profile (`driver.enabled=false`, `toolkit.enabled=false`, `operator.runtimeClass=nvidia-container-runtime`, all three together).

The profile direction reached consensus in internal design discussion (2026-07-14). Remaining decision points are listed in the ADR's **Deferred Adoption Decisions** section — none changes the mechanism or gates acceptance; each is routed to #1761 or the consumer work that resolves it. The Deferred Decision 4 catalog-compatibility follow-up is filed as #1812.

Fixes: N/A
Related: #1761 (implementation tracking), #1755 (GKE label check — first GKE consumer dependency), #1327 (allocation-policy contract being amended)

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

Design-only PR — no code changes. Key decisions recorded in the ADR:

- **Single profile per resolved composition** — generic multi-params rejected (Alternative C).
- **Compact ownership record** persisted in the artifact: owned-path union across all declared values, component presence as the synthetic `enabled` path.
- **#1327 policy closure recomputed** at every `RecipeResult` boundary from one canonical descriptor + shared evaluator. Rendered outputs are the exception — descriptor-expansion remediation is outcome-based (regenerate every previously rendered profiled output that diverges or stays install-time-mutable at a newly locked path, or conservatively all affected outputs), and each expansion is a re-qualification/evidence re-signing event with profile-bearing evidence current only when both the recipe digest and the scoped descriptor identity (the descriptor entries contributing to that recipe's effective closure) match (a profile-scoped ADR-007 amendment).
- **`advertiser: external`** as the only reserved marker.
- **Closed, strictly-decoded fragment schema** — value level: `advertiser`/`constraints`/`componentRefs`; each ref: `name`/`overrides` only; no `valuesFile`; existing components only.
- **Resolved policy value unchanged** (`device-plugin-extended-resource`).
- **New recipe apiVersion** for profile-bearing artifacts (explicit ADR-011 amendment) so released binaries reject rather than silently ignore profiles.

**Compatibility:** landing the mechanism alone is non-breaking by construction.

- A composition without a profile declaration produces byte-identical artifacts (legacy apiVersion, unchanged digests) — existing overlays, generated recipes, committed evidence, `/v1` endpoints, and released binaries are unaffected.
- Exactly six day-1 behavior changes land, all fail-closed tightenings on inputs invalid under the ADR: `aicr mirror list` gains the shared context-aware validation gate (no rejection for well-formed recipes; it validates a defensive copy, so `Discover` output and the caller's artifact are unchanged); raw-artifact boundaries reject unknown-apiVersion artifacts that today pass unchecked; legacy/empty-version artifacts carrying `selectedProfile` are newly rejected (today's non-strict decode silently drops the field); legacy/empty-version profile-declaring overlays are newly rejected at catalog load (today the kind-only loader silently yields an unprofiled catalog); new-version overlays without a declaration are newly rejected at catalog load (today accepted kind-only); and `/v1` requests carrying explicit profile input are newly rejected (today the parameter is silently dropped).
- All breaking effects on well-formed inputs (new apiVersion, digest changes, family-wide `/v1` rejection of ordinary criteria requests, evidence re-signing) are deferred to each family's explicit profile conversion — a documented per-family qualification and cut-over event (see the ADR's Consequences section).

Design-review PR: implementation (#1761) starts after this design is accepted.

## Testing

Doc-only change: a single new file, `docs/design/015-recipe-configuration-profiles.md`. No code, no user-facing docs pages, no sidebar entries (design docs are not part of the Fern site).

Full `make qualify` was run against an earlier revision of this file and passed (78.8% coverage, zero lint issues, 23/23 E2E, vulnerability gate). Subsequent revisions are prose-only edits to the same single file with no new external links, verified with scoped checks per the doc-only rule: heading/anchor hygiene reviewed, `git diff --check` clean, and Fern Docs CI (including the link checker) passing on the pushed head.

## Risk Assessment

**Rollout notes:** N/A — no behavior change until #1761 lands.

## Checklist

### DIFF
--- a/docs/design/007-recipe-evidence.md
+++ b/docs/design/007-recipe-evidence.md
@@ -1,5 +1,12 @@
 # ADR-007: Verifiable Recipe Test Evidence
 
+> **Amended by [ADR-015](015-recipe-configuration-profiles.md).** For
+> profile-bearing recipes, evidence currentness requires both the recipe
+> digest and the canonical-descriptor identity in the signed predicate to
+> match, and profile-bearing evidence rides a new predicate type. Unprofiled
+> evidence keeps this ADR's digest-only currentness rule and the v1
+> predicate; the pointer layout and verification flow below are unchanged.
+
 ## Status
 
 **Accepted, partially implemented** — proposed 2026-05-08.
--- a/docs/design/011-artifact-apiversion-policy.md
+++ b/docs/design/011-artifact-apiversion-policy.md
@@ -1,5 +1,12 @@
 # ADR-011: Artifact apiVersion Policy and Compatibility Gate
 
+> **Amended by [ADR-015](015-recipe-configuration-profiles.md).** ADR-015
+> introduces kind-scoped version evolution: `RecipeMetadata` and
+> `RecipeResult` accept `aicr.run/v1alpha3` for profile-bearing artifacts
+> (dual-accepted alongside the legacy version), while snapshots, configs,
+> and `RecipeCriteria` stay on `aicr.run/v1alpha2`. The single-version
+> posture below is otherwise unchanged.
+
 > **Amended by [ADR-013](013-aicr-run-domain-migration.md).** This ADR
 > established the single-sourced, enforced `apiVersion` contract while the
 > domain was `aicr.nvidia.com` and the version was `v1alpha1`. ADR-013 later
--- a/docs/design/015-recipe-configuration-profiles.md
+++ b/docs/design/015-recipe-configuration-profiles.md
@@ -1,0 +1,1378 @@
+# ADR-015: Recipe-Declared Configuration Profiles
+
+## Status
+
+**Accepted** — 2026-07-21 (proposed 2026-07-14).
+
+Originated from an internal GKE device-plugin ownership discussion
+(2026-07-14) and
+[#1755](https://github.com/NVIDIA/aicr/issues/1755) (fail-closed check for
+GKE device-plugin ownership conflict), but the mechanism is
+**service-agnostic**: it applies wherever one criteria combination maps to
+more than one qualified cluster configuration.
+
+## Problem
+
+A single criteria combination increasingly maps to more than one *valid*
+cluster configuration. The recurring axis is **who owns a layer of the GPU
+stack — the cloud service or the GPU Operator**:
+
+- **AKS**: two documented *unmanaged*-pool ownership modes
+  (`docs/integrator/aks-gpu-setup.md`): GPU Operator-managed pools created
+  with `--gpu-driver none` (operator installs driver + toolkit) vs. the AKS
+  **"Driver only"** install profile with driver and toolkit preinstalled
+  (gpu-operator needs `driver.enabled=false`, `toolkit.enabled=false`,
+  `operator.runtimeClass=nvidia-container-runtime`, all three together —
+  mixing modes leaves containerd without a working `nvidia` runtime
+  handler). Fully AKS-managed GPU node pools (`--enable-managed-gpu=true`,
+  preview) additionally own the device plugin, DCGM exporter, and health
+  tooling; that mode conflicts with the operands AICR deploys and stays
+  **out of scope** here, as the AKS guide already states.
+- **GKE**: device-plugin ownership. The stock profile runs the GPU
+  Operator's device plugin and requires GPU node pools to carry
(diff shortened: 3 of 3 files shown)


yuanchen8911:
Ready for design review. Orientation is in the PR description; the full design is `docs/design/015-recipe-configuration-profiles.md`.

Please focus on (1) whether the proposed design is sound, and (2) opportunities to simplify — the ADR holds itself to "every element must trace to a concrete failure it prevents," and challenges under that bar are welcome. The Deferred Adoption Decisions section at the end of the doc is non-gating (each item is routed to implementation or consumer work); implementation tracking is #1761.

atif1996 (dismissed):

mchmarny:
Strong design. The "no signal, no declaration" rule keeps it honest, the fail-closed posture is consistent across every boundary (apiVersion gate, catalog load, /v1 rejection, selection typos), and the outcome-based pre-output invariant is the right shape (implicit validation on construction, future mutations are caught without anyone maintaining an enumeration (that's the main thing that disqualifies `--set` and `--dynamic` with criteria axes.

1 non-blocking nit: converting family breaks ordinary criteria requests on /v1, not just profile-aware ones (i.e. clients find out at cut-over). Suggest adoption step 2 check /v1 usage for the family (or announce a deprecation window) before the first conversion. Doesn't change the mechanism.

/approved

mchmarny (dismissed):

yuanchen8911:
Good catch on the `/v1` cut-over discovery. Fixed in 8da321de — adoption step 2 now checks the family's `/v1` usage, or announces a deprecation window, before the first conversion, so clients don't find out at cut-over.

The same commit executes the on-acceptance obligations: Status is now Accepted, and ADR-007 and ADR-011 carry the reciprocal "Amended by ADR-015" banners.

mchmarny (approved):

njhensley (approved):

PR #1764, opened 2026-07-15
feat(ci): collect UAT cluster debug bundle on failure

njhensley:
## Summary

On UAT failure, snapshot live cluster state into a `cluster-debug/` bundle (nodes, taints/reboot fingerprint, events, operator CRs incl. Skyhook, operator/check-Job logs) before teardown — so a failed run's artifact explains **why** it failed, not just **what** check failed.

## Motivation / Context

The existing `Upload failure debug` artifact captured only runner-local files (`report.json`, `recipe.yaml`, `snapshot.yaml`, `dry-run.json`) — **zero live cluster state**. Two recent UAT-AWS runs (`29393800033`, `29402956254`) failed at the `expected-resources` deployment check on a non-monotonic **skyhook re-tuning race** (`Skyhook /tuning status.status` goes `complete → in_progress` after the readiness gate certifies it, and the tuning reboot can evict the check Job's pod). Root-causing that required reconstructing the timeline from raw CI step logs, because the bundle contained none of the Skyhook status history, node events, or reboot markers needed. This closes that gap.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT harness (`tests/uat/**`) + UAT workflows (`.github/workflows/uat-*.yaml`)

## Implementation Notes

- **`tests/uat/lib/collect-debug.sh`** (new shared, cloud-agnostic lib):
  - `collect_cluster_debug` — full bundle: a `MANIFEST.yaml` (failing checks parsed straight from `report.json`), node conditions + **reboot fingerprint** (bootID / kernel / Ready `lastTransitionTime`) + taints, cluster events (reboots, pod evictions), operator CRs (**Skyhook full YAML first**), and per-namespace describe/events + operator/check-Job logs incl. `--previous`.
  - `capture_skyhook_snapshot` — fast, focused snapshot invoked **inline** the moment a validate failure is detected, seconds after the failing check (while `status.status` is most likely still `in_progress`), vs. the teardown-time full bundle taken minutes later.
- New **`debug` phase** wired into `tests/uat/{aws,gcp,azure}/run` (dispatch + usage/header docs).
- **`phase_install`** now persists the readiness-gate `status.status` **time-series** (each attempt, timestamped) into the bundle — so `complete → in_progress` tuning flips are captured as they happen, not lost to a single snapshot.
- **`phase_conformance`** captures the validate exit code and, on failure, calls `capture_skyhook_snapshot` before propagating (was: `set -e` aborted immediately).
- A `failure()`-gated **`Collect cluster debug bundle`** step (8-minute bound, before teardown) added to `uat-{aws,gcp,azure}.yaml`, with `cluster-debug/**` added to the upload artifact.
- Everything is **best-effort** — every `kubectl` call is guarded, the collector never fails the run, and the step is bounded so it can't eat the teardown budget or leak a GPU node. Secrets are not dumped.

Follow-up (out of scope, noted for reviewers): `aws/run` and `gcp/run` are functionally identical (differ only in 3 comment lines) and `azure/run` is the same core plus an `az_federated_relogin` hook — strong candidate for a further `tests/uat/lib/phases.sh` consolidation in a dedicated PR.

## Testing

```bash
# Shell + YAML only; no Go touched.
shellcheck -x tests/uat/lib/collect-debug.sh          # clean
shellcheck tests/uat/{aws,gcp,azure}/run              # clean (SC1091 source-resolution info only)
yamllint -c .yamllint.yaml .github/workflows/uat-*.yaml  # clean
actionlint .github/workflows/uat-*.yaml               # no new findings
```

Also smoke-tested the collector against the real `report.json` from run `29393800033` with no cluster reachable: it no-ops gracefully and the `MANIFEST.yaml` correctly surfaces both failing checks and points at the new time-series artifacts (`readinessGateSeries`, `skyhookAtFailure`).

## Risk Assessment

- [x] **Low** — Isolated, best-effort diagnostics that only run on `failure()` before teardown; no change to the pass path or to `aicr` itself. The `phase_conformance` change preserves the prior failure semantics (still exits non-zero on a validate failure).

**Rollout notes:** No migration. The new artifact fields appear on the next failed UAT run.

## Checklist

- [x] Linter passes (`make lint`) — shellcheck + yamllint + actionlint clean
- [ ] I added/updated tests for new functionality — N/A (shell harness; validated via shellcheck + no-cluster smoke test)
- [x] I updated docs if user-facing behavior changed — N/A (internal UAT harness, not user-facing `aicr` surface)
- [x] Changes follow existing patterns in the codebase (mirrors `serve_debug()` / `phase_prep` snapshot-debug idioms)


njhensley:
Thanks @coderabbitai — all three findings were valid and are addressed in the amended commit:

1. **Job timeout budget (`uat-{aws,gcp,azure}.yaml`)** — bumped each job `timeout-minutes` 270→**280** and updated the budget math to account for the new failure-path debug step (`190 UAT + ~30 setup + ~8 debug + ~40 teardown = ~268`), restoring ~12m teardown headroom so an oversized failure-path run can't cancel the always()-run `Destroy Cluster` and leak the GPU node.

2. **Security/privacy claim (`collect-debug.sh`)** — narrowed the claim: it now states only that **no Secret objects are fetched**, and explicitly warns that `describe pods` (plain env values) and `logs --previous` can carry credentials/PII, that the artifact is a private 30-day-retention bundle to be treated as sensitive, and that redaction is deliberately not attempted (fragile for an internal UAT diagnostic on a private repo).

3. **Per-command timeout (`_cd_section`)** — added `CLUSTER_DEBUG_CMD_TIMEOUT` (30s) and a portable `_cd_bounded` helper. Every external kubectl/`bash -c` call now runs under `timeout` when available (Linux CI) and directly otherwise (macOS dev boxes lack `timeout`), so one hung call can't starve the high-value per-namespace pod-log sections. Applied to `_cd_section`, the reboot-fingerprint, `capture_skyhook_snapshot`, and the per-namespace log loop. Note: `timeout` can't invoke a shell function, so the function-based fingerprint section is routed directly and self-bounds its own kubectl.

Validated: shellcheck + `bash -n` clean, yamllint + actionlint clean (no new findings), and the no-cluster smoke test still produces the bundle + MANIFEST correctly (the macOS run also exercised the no-`timeout` fallback path).

tests/uat/azure/run
```diff
@@ -787,6 +821,7 @@ case "${phase}" in
   train)       phase_train ;;
   serve)       phase_serve ;;
   verify)      phase_verify ;;
+  debug)       collect_cluster_debug ;;
```
mchmarny: The `debug` phase dispatches straight to `collect_cluster_debug`, but on AKS every kubectl goes through kubelogin's cached AAD token, which is why the long-running phases periodically call `az_federated_relogin`. A failure that surfaces after that token expires (exactly the long tuning-race runs this bundle is built for) leaves the collector with a dead credential — every kubectl no-ops behind its guard and the bundle comes out empty, silently. `debug) az_federated_relogin || true; collect_cluster_debug ;;` mirrors what the phases already do; `ACTIONS_ID_TOKEN_REQUEST_URL` is present in the step env since the job has `id-token: write`.

tests/uat/lib/collect-debug.sh
```diff
@@ -0,0 +1,276 @@
...
+  # Nodes carry the reboot/taint evidence central to tuning-race failures.
+  _cd_section "nodes (wide)" nodes.txt kubectl get nodes -o wide
+  _cd_section "nodes (yaml)" nodes.yaml kubectl get nodes -o yaml
+  _cd_section "nodes (describe)" nodes-describe.txt kubectl describe nodes
+  # Reboot fingerprint + taints: bootID (changes across a reboot), kernel, the
+  # Ready condition's lastTransitionTime, and taints. The skyhook.nvidia.com
+  # NoSchedule taint applied during tuning and removed on completion — plus a
+  # fresh Ready transition / changed bootID — is the durable "a reboot re-opened
+  # tuning" signal that distinguishes an in-flight reboot from a genuine resource
+  # gap. Routed directly (not via _cd_section) because it is a shell function,
```
mchmarny, line 214: `kubectl describe pods` renders literal `env:` values in its Environment section, and this loop describes every namespace — while artifacts on a public repo are downloadable by anyone. Operators shouldn't carry secrets as literal envs, but a UAT recipe override or third-party chart doing it would leak straight into the bundle. Consider limiting `describe pods` to the `CLUSTER_DEBUG_LOG_NAMESPACES` allowlist like the log collection (get/events for the rest), or piping describe through a `sed` that masks Environment values.

mchmarny (commented):
This closes a real gap — the two recent UAT-AWS failures took raw-CI-log archaeology to root-cause, and the three-layer design here (gate time-series during install, inline Skyhook snapshot at failure, full bundle before teardown) captures exactly the non-monotonic signal that made those hard. The shell discipline is good throughout: guarded kubectl, bounded step, unchanged pass-path semantics in the gate loop, and the errexit-suspension trap in the relogin helper is explicitly handled.

Two comments inline: one medium (the azure `debug` phase skips `az_federated_relogin`, so an expired kubelogin token yields a silently empty bundle in exactly the long-failure case this targets) and one low (describe-pods env exposure to public artifacts). Neither changes the design; the azure one is worth fixing before merge since it hollows out the feature on one of the three clouds. Branch is behind `main` — rebase before merge.

njhensley:
Thanks @mchmarny — both are real and now fixed in the amended commit.

**1. Dead credential at collect time (`azure/run` + AWS/GCP workflows).** Spot on — a failure that surfaces after the token expires is *exactly* the long tuning-race case this bundle targets, so an empty-but-silent bundle would defeat the purpose. Fixed the credential refresh at all three collection entry points:
- **Azure:** `debug)` now runs `az_federated_relogin || true; collect_cluster_debug` (your suggestion verbatim) — mirrors the phases; no-op outside CI.
- **AWS & GCP:** the same gap exists on a long run (the 1h AWS session has expired mid-job before — it caused a teardown leak previously). Added a `failure()`-gated **Refresh AWS credentials / Re-authenticate to GCP for debug collection** step immediately before each Collect step, reusing the existing `configure-aws-credentials` / `google-github-actions/auth` pattern. Also dropped the now-incorrect "session still valid inside the 1h window" comment on the AWS collect step.

**2. `describe pods` env leak across all namespaces on a public repo.** Correct — and the repo being public makes it worse than my earlier "private repo" note claimed (that note was wrong; fixed). `describe pods` is now limited to the `CLUSTER_DEBUG_LOG_NAMESPACES` operator/check allowlist (same gate as log collection); every other namespace gets only `get pods` + `get jobs` + `get events`. Went with the allowlist over a `sed` mask since masking `describe` output is fragile. The header privacy note now states honestly: no Secret objects fetched, describe scoped to the allowlist, but describe/logs on those namespaces can still carry app-emitted creds/PII, and the artifact is public — treat it as sensitive.

Re-validated: shellcheck / bash-syntax / yamllint / actionlint clean, and the no-cluster smoke test still produces the bundle (incl. the no-`timeout` fallback path). Appreciate the careful read.

mchmarny (approved):

Issue #1636, opened 2026-07-07
inference-perf fails on main UAT cells: aiperf-bench image tag never built/pushed


njhensley:
## Summary

The `inference-perf` performance check fails on from-source (main) UAT cells because the `aiperf-bench` benchmark image is never built or pushed with the per-run `uat-<run_id>` tag that the validator resolves. The AIPerf pod sits in `ImagePullBackOff` (phase `Pending`) for the full 15 min `InferencePerfJobTimeout` and the run fails.

Observed failure:

```
[INTERNAL] AIPerf job failed: [SERVICE_UNAVAILABLE] pod watch closed before pod reached terminal state
```

## Root cause

1. UAT main cells set `AICR_VALIDATOR_IMAGE_TAG: uat-<run_id>` (`.github/workflows/uat-gcp.yaml:121`, `uat-aws.yaml:125`).
2. `resolveAiperfImage()` → `catalog.ResolveImage("…/aiperf-bench:latest", "main", <sha>)`. `isReleaseVersion("main")` is false, so the `AICR_VALIDATOR_IMAGE_TAG` override rewrites the tag → `…/aiperf-bench:uat-<run_id>`.
3. But the UAT "Build and push validator images" step only builds `deployment performance conformance` — `aiperf-bench` is omitted (`uat-gcp.yaml:232`, `uat-aws.yaml:236`). So `aiperf-bench:uat-<run_id>` is never pushed → the pod 404s on pull → `Pending` until the 15 min job timeout.

`on-push.yaml` / `on-tag.yaml` `VALIDATOR_PHASES` **do** include `aiperf-bench`, so release cells (which resolve `aiperf-bench:<version>`) work; only from-source main cells break. The path is only exercised by the inference intent (DC2, #1275), which is why it surfaced now.

The `SERVICE_UNAVAILABLE / pod watch closed` message is a misleading surface symptom: the watch channel closed ~9 s before the context deadline, so `pkg/k8s/pod/wait.go` reported watch-closed instead of timeout. Same failure, different label.

## Fix

**Primary:** add `aiperf-bench` (Dockerfile `validators/performance/aiperf-bench.Dockerfile`, image title `aiperf-bench`) to the UAT build/push step in both `uat-gcp.yaml` and `uat-aws.yaml` so the `uat-<run_id>` tag exists.

**Secondary (diagnosability):**
[...]

PR #1765, opened 2026-07-15
fix(ci): guard UAT image cleanup on numeric version id

njhensley:
## Summary

Guard the end-of-run UAT validator-image cleanup on a numeric package-version id so a `gh api` 404 body can no longer be fed into the DELETE URL.

## Motivation / Context

Every UAT run ends with a step that deletes the `uat-<run_id>`-tagged validator images. It loops over phases `deployment performance conformance aiperf-bench agent`, but `agent` is never built by the build loop (which builds only `deployment performance conformance aiperf-bench`), and other phases 404 on runs that skipped them. On an error status `gh api` skips `--jq` and prints the raw JSON error body to **stdout**, which the `2>/dev/null || true` capture lands in `VERSION_ID`. The old `[[ -n "${VERSION_ID}" ]]` guard then passed and fed that multi-line JSON blob into the DELETE URL, so every run logged:

```
parse "repos/NVIDIA/aicr/packages/container/aicr-validators%2Fdeployment/versions/{\r\n  \"message\": \"Not Found\",...}": net/url: invalid control character in URL
```

The cleanup still worked for images that existed; the errors were noise, but noise that obscured real failures in the step.

Fixes: N/A
Related: #1636

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT workflows (`.github/workflows/uat-aws.yaml`, `.github/workflows/uat-gcp.yaml`)

## Implementation Notes

Changed the guard from a non-empty check to `[[ "${VERSION_ID}" =~ ^[0-9]+$ ]]`. The GitHub package-version API returns `.id` as a JSON integer, so on success `VERSION_ID` is bare digits and legitimate deletes are unaffected. The numeric guard fails closed on the three cases that should be skipped: the 404 JSON body (multi-line, control chars), the empty "no matching tag" case, and any multi-line jq output (e.g. if two versions ever shared a tag). Added a comment explaining the `gh api`-on-error stdout behavior and why `agent` reliably 404s.

## Testing

```bash
yamllint .github/workflows/uat-aws.yaml .github/workflows/uat-gcp.yaml  # clean
```

Workflow-only change (no Go sources touched), so the Go test/lint gates in `make qualify` do not apply. Verified `gh api` prints the 404 body to stdout and that the new regex rejects it while matching a bare integer id.

## Risk Assessment

- [x] **Low** — Isolated change to a best-effort cleanup step, trivially revertible

**Rollout notes:** N/A — takes effect on the next UAT run.

## Checklist

- [x] Linter passes (yamllint)
- [ ] I added/updated tests for new functionality — N/A (workflow shell guard)
- [x] I updated docs if user-facing behavior changed — N/A

### DIFF
--- a/.github/workflows/uat-aws.yaml
+++ b/.github/workflows/uat-aws.yaml
@@ -691,10 +691,15 @@ jobs:
           # clutter. aiperf-bench is included so the tag pushed by the build
           # step above (#1636) does not leak.
           for phase in deployment performance conformance aiperf-bench agent; do
+            # gh api prints the raw error body to stdout on a 404 (a phase whose
+            # package was never pushed — e.g. agent, which the build loop above
+            # does not build) and skips --jq, so VERSION_ID can capture a JSON
+            # blob with control chars. Require a numeric id before deleting so
+            # the 404 body, the empty case, and multi-line output all fail closed.
             VERSION_ID=$(gh api \
               "repos/${{ github.repository }}/packages/container/aicr-validators%2F${phase}/versions" \
               --jq ".[] | select(.metadata.container.tags[] == \"${VALIDATOR_TAG}\") | .id" 2>/dev/null) || true
-            if [[ -n "${VERSION_ID}" ]]; then
+            if [[ "${VERSION_ID}" =~ ^[0-9]+$ ]]; then
               gh api --method DELETE \
                 "repos/${{ github.repository }}/packages/container/aicr-validators%2F${phase}/versions/${VERSION_ID}" || true
             fi
--- a/.github/workflows/uat-azure.yaml
+++ b/.github/workflows/uat-azure.yaml
@@ -769,10 +769,15 @@ jobs:
           # clutter. aiperf-bench is included so the tag pushed by the build
           # step above (#1636) does not leak.
           for phase in deployment performance conformance aiperf-bench agent; do
+            # gh api prints the raw error body to stdout on a 404 (a phase whose
+            # package was never pushed — e.g. agent, which the build loop above
+            # does not build) and skips --jq, so VERSION_ID can capture a JSON
+            # blob with control chars. Require a numeric id before deleting so
+            # the 404 body, the empty case, and multi-line output all fail closed.
             VERSION_ID=$(gh api \
               "repos/${{ github.repository }}/packages/container/aicr-validators%2F${phase}/versions" \
               --jq ".[] | select(.metadata.container.tags[] == \"${VALIDATOR_TAG}\") | .id" 2>/dev/null) || true
-            if [[ -n "${VERSION_ID}" ]]; then
+            if [[ "${VERSION_ID}" =~ ^[0-9]+$ ]]; then
               gh api --method DELETE \
                 "repos/${{ github.repository }}/packages/container/aicr-validators%2F${phase}/versions/${VERSION_ID}" || true
             fi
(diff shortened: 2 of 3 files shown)


.github/workflows/uat-aws.yaml
```diff
@@ -691,10 +691,15 @@ jobs:
...
+            # gh api prints the raw error body to stdout on a 404 (a phase whose
+            # package was never pushed — e.g. agent, which the build loop above
+            # does not build) and skips --jq, so VERSION_ID can capture a JSON
+            # blob with control chars. Require a numeric id before deleting so
+            # the 404 body, the empty case, and multi-line output all fail closed.
             VERSION_ID=$(gh api \
               "repos/${{ github.repository }}/packages/container/aicr-validators%2F${phase}/versions" \
               --jq ".[] | select(.metadata.container.tags[] == \"${VALIDATOR_TAG}\") | .id" 2>/dev/null) || true
-            if [[ -n "${VERSION_ID}" ]]; then
+            if [[ "${VERSION_ID}" =~ ^[0-9]+$ ]]; then
```
mchmarny, line 702: Same loop, same `[[ -n "${VERSION_ID}" ]]` guard exists in `uat-azure.yaml` (line 775 on main) — this PR fixes aws and gcp but the azure cleanup keeps producing the exact `invalid control character in URL` noise described in the description. Apply the same comment + numeric guard there in this PR so all three UAT workflows stay in lockstep.

mchmarny (commented):
Correct diagnosis (gh api printing the error body to stdout and skipping `--jq` is exactly the trap) and the numeric guard is the right shape — it fails closed on the 404 body, the empty case, and multi-line output alike, and the inline comment will save the next person the archaeology. One gap inline: `uat-azure.yaml` carries the identical loop and guard and needs the same fix in this PR, otherwise the noise just moves to one cloud. Nothing else blocks.

.github/workflows/uat-aws.yaml
njhensley: Good catch — done. Applied the identical comment + numeric guard to `uat-azure.yaml` (cleanup loop, formerly line 775) in this PR, so all three per-cloud UAT workflows now fail closed on the 404 body / empty / multi-line cases. Rebased onto latest `main` (which had moved the block); the fix is intact in aws, gcp, and azure.

.github/workflows/uat-aws.yaml
```diff
@@ -722,10 +722,15 @@ jobs:
...
               gh api --method DELETE \
                 "repos/${{ github.repository }}/packages/container/aicr-validators%2F${phase}/versions/${VERSION_ID}" || true
             fi
```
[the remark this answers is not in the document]
njhensley: Skipping this one. The non-empty-non-numeric branch is the *expected* case here, not a rare anomaly: when a phase's package doesn't exist (`agent` is never built by the build loop, so it 404s on **every** run; other phases 404 whenever they weren't pushed), `gh api` prints the 404 JSON body to stdout and skips `--jq`, so `VERSION_ID` holds that non-numeric blob. An `elif [[ -n ... ]]` warning would therefore emit a `::warning::` for `agent` (plus any skipped phase) on every run — reintroducing exactly the per-run log noise this PR removes.

Making the warning meaningful would require distinguishing a 404 from a genuinely malformed 200 (e.g. gating the stdout capture on the call's exit status so a 404 yields empty), which is a larger restructure than this fail-closed one-liner warrants. Leaving it silent: the numeric guard already fails closed correctly, and a real cleanup regression would surface as images accumulating in GHCR.

mchmarny (approved):

PR #1769, opened 2026-07-15
refactor(ci): consolidate UAT runners into shared phases.sh

njhensley:
## Summary

Consolidate the three near-identical UAT runners (`tests/uat/{aws,gcp,azure}/run`) into a shared `tests/uat/lib/phases.sh`; each runner becomes a thin shim. Pure refactor — behavior preserved.

## Motivation / Context

`aws/run` and `gcp/run` differed in only **three comment lines**; `azure/run` was the same body plus a federated-session refresh. Every phase change had to be applied three times (as #1764 demonstrated — the same edit landed in all three runners). This extracts the shared body once.

Fixes: N/A
Related: #1764

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT harness (`tests/uat/**`)

## Implementation Notes

- **`tests/uat/lib/phases.sh`** (new) holds the cloud-agnostic body: `phase_prep/install/conformance/train/serve/verify`, `inject_push_target`/`serve_debug`, env defaults, and a `uat_main` dispatcher.
- Each runner is now a ~30-line shim: `source ../lib/phases.sh` + `uat_main "$@"`.
- **One hook** carries all cloud-specific behavior: `cloud_refresh_credentials` (default no-op) + `CLOUD_REFRESH_INTERVAL_SECONDS`. Azure's shim overrides the hook with `az_federated_relogin` (moved **verbatim**) and lowers the interval; the readiness gate calls the hook periodically and the `debug` phase forces it before collecting. AWS/GCP keep the no-op default.
- Net: ~2400 lines across three files → ~790 shared + ~30 per shim (**−1436 lines**).

**Behavior preservation — verified.** Every extracted phase function is byte-identical to the previous `aws/run` (checked function-by-function with `diff`), **except**:
- `phase_install`: the new `cloud_refresh_credentials` hook call in the gate loop (no-op for AWS/GCP).
- `phase_conformance` + `phase_serve`: two comments genericized for the multi-cloud context.
- `az_federated_relogin`: moved verbatim (diff-confirmed identical).

The Azure gate-refresh behavior is unchanged — the previous inline interval-gated `az_federated_relogin` is now the generic hook driven by the same interval; only the `::warning::` message text is more generic.

## Testing

```bash
shellcheck -x tests/uat/lib/phases.sh tests/uat/{aws,gcp,azure}/run   # clean
bash -n <each>                                                        # clean
# function-equivalence: each phase_* fn diffed against pre-refactor aws/run → identical modulo the hook + 2 comments
# dispatch: bad phase → exit 2; missing config → exit 2; usage on no args → matches original
```

## Risk Assessment

- [x] **Medium** — Touches the runners that drive the (expensive) UAT GPU jobs, but the change is a mechanical extraction with byte-level equivalence verified per function, and the one behavioral seam (the Azure refresh hook) reproduces the prior logic. No workflow or `collect-debug.sh` changes.

**Rollout notes:** No interface change — the workflows still call `./tests/uat/<cloud>/run <phase> <config>` identically. Best validated by a UAT dry-run per cloud before relying on it.

## Checklist

- [x] Tests pass locally — N/A (shell); shellcheck + bash-syntax + function-equivalence + dispatch checks all pass
- [x] Linter passes (`make lint`) — shellcheck clean
- [ ] I added/updated tests — N/A (shell harness)
- [x] I updated docs if user-facing behavior changed — N/A (internal harness, no interface change)
- [x] Changes follow existing patterns (mirrors the `collect-debug.sh` shared-lib pattern introduced in #1764)


njhensley:
Thanks @coderabbitai — valid, fixed in the amended commit. The `debug` dispatch now warns instead of swallowing: `cloud_refresh_credentials || echo "::warning::cloud credential refresh failed before debug collection; the bundle may come out empty" >&2`, then still proceeds (best-effort). So the exact failure mode the comment above warns about is now surfaced in the step log rather than silent. shellcheck/bash-syntax clean.

njhensley:
Thanks @coderabbitai — all three are **valid observations but pre-existing**, not introduced by this PR:

- **Platform check after evidence push** (`phase_conformance`) and **serve readiness `total -gt 0`** (`phase_serve`) exist byte-identically in the merged `main` runners (`tests/uat/aws/run` lines ~394/428 and ~648 respectively), and the `phase_verify` item likewise. This PR **moves those functions verbatim** into `phases.sh` — I verified each extracted phase function is byte-identical to the previous `aws/run` (the whole point of the refactor is zero behavior change), so these behaviors are unchanged from what's already on `main` and running in UAT today.

Deliberately **not fixing them here**: changing phase logic would break this PR's behavior-preserving guarantee (and its byte-equivalence verification) and muddy a pure-refactor diff. I'll track them as follow-ups:
1. Run the platform-CRD sanity check *before* `validate --phase all` publishes the bundle (Major — avoids routing evidence to the wrong TestGrid tab on a misdeclared platform).
2. Gate serve readiness on both declared components (`-ge 2`, or component-aware) so the Frontend becoming ready before the decode Worker can't pass prematurely (Major — flaky inference).
3. The `phase_verify` Minor.

These will land alongside the separate `collect-debug.sh` `set -e` early-abort fix as a small "UAT harness correctness" follow-up, so each is reviewed on its own merits rather than smuggled into the refactor.

njhensley:
📋 Tracking the deferred pre-existing findings in **#1770** (`UAT harness correctness`, type: Bug), to be fixed as a focused follow-up **after this refactor merges** — so #1769 stays a pure, behavior-preserving extraction. #1770 covers:
- the platform-CRD check running after evidence is published (`phase_conformance`),
- serve readiness passing on a single pod (`phase_serve`),
- the `phase_verify` `set -e` detail-masking, and
- (found separately) `collect_cluster_debug` aborting early under `set -e` → truncated bundle.

@coderabbitai these are all pre-existing on `main` (byte-identical to the code this PR moves) and are now tracked in #1770; please treat them as deferred for this PR.

tests/uat/lib/phases.sh
```diff
@@ -0,0 +1,820 @@
...
+#               push target (e.g. ${{ github.run_id }} in CI, or `local-$(date +%s)`)
+#
+# Working files are written to $PWD: snapshot.yaml, recipe.yaml, bundle/,
+# dry-run.json, report.json, evidence/, evidence-result.json, train-logs/.
+
+# Shared, cloud-agnostic cluster debug-bundle collector (collect_cluster_debug,
+# capture_skyhook_snapshot), invoked by the `debug` phase on failure and inline
+# from phase_conformance. Lives alongside this file in tests/uat/lib/.
+# shellcheck source=./collect-debug.sh
+source "$(dirname "${BASH_SOURCE[0]}")/collect-debug.sh"
```
mchmarny, line 61: blocker: Current main deleted collect-debug.sh, and all three runner shims now conflict. If the rebase keeps these shims while accepting main’s deletion, every runner exits while sourcing this file. Rebase and reconcile the complete collector/gate-log call chain with main rather than resolving only the visible wrapper conflicts.

mchmarny (dismissed):
The shared-runner shape is reasonable, but this head cannot be integrated with current main safely: all runner shims conflict and the shared library depends on debug infrastructure deleted on main. Rebase and reconcile that dependency chain before merge; required CI currently tests only the stale head.

njhensley:
@mchmarny You're right, thanks for catching it — and the root cause is broader than this PR. **#1744 (`7e54fca7`) reverted all of #1764**: its branch was based on a pre-#1764 `main`, and the squash-merge removed `tests/uat/lib/collect-debug.sh` and reverted all three runners **and** all three `uat-*.yaml` workflows — the exact seven files #1764 added. So the debug-bundle feature is currently gone from `main` (`aws/run` is back to the pre-#1764 phase list, no `debug` phase; the workflows have no `Collect cluster debug bundle` step). A stale-branch clobber, not a conflict in #1769 per se.

That means reconciling this PR also has to restore #1764's infra. I'm confirming the cleanest path (restore #1764 first, then rebase this refactor on top, so it stays a pure extraction — vs. folding the restore into this PR) and will flag the regression to maintainers so #1744's other changes can be checked for similar stale-clobber damage. Will update here once the approach is settled.

njhensley:
@mchmarny Rebased onto current `main` — the blocker is resolved. Maintainers reverted #1744 (via #1788), so `tests/uat/lib/collect-debug.sh` and the debug infra are reinstated; the rebase applied cleanly (my shims replace runners that #1788 restored to their post-#1764 state, so no conflict), and `phases.sh`'s `source .../collect-debug.sh` now resolves.

Re-validated against the rebased `main`:
- **File set unchanged** — only `tests/uat/lib/phases.sh` (new) + the three runner shims; no workflow changes.
- **Behavior still preserved** — every extracted phase function is byte-identical to current `main`'s `aws/run` except `phase_install`'s `cloud_refresh_credentials` hook and two genericized comments (`phase_conformance`, `phase_serve`); `az_federated_relogin` is byte-identical to `main`'s `azure/run`.
- **Sourcing works** — a shim now sources `collect-debug.sh` and reaches dispatch (no more exit-on-source); shellcheck/bash-syntax clean on all four.

Ready for another look when you have a moment.

mchmarny (approved):

PR #1771, opened 2026-07-15
feat(corroborate): add cross-version combined dashboard view

njhensley:
## Summary

The corroboration dashboard now defaults to a cross-version **combined** view — each source's single latest run, folded into one consensus grid — so every source that has attested a recipe is visible even when their latest runs are at different AICR releases. Selecting a specific AICR version switches to that version's strict same-version grid.

## Motivation / Context

The dashboard previously showed only the **newest** AICR version's baked consensus grid per recipe. Sources whose latest run was at an older release were silently hidden, so a recipe validated by several sources across different tool versions looked under-corroborated. This adds an "all versions" default (the equivalent of `AICR_VERSION=all`, but at the corroborate level) while keeping strict per-version corroboration one selection away.

Fixes: N/A
Related: N/A (follows the GP4 corroborate dashboard work)

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: corroboration dashboard (`pkg/corroborate`, `tools/corroborate`)

## Implementation Notes

- **Consensus math stays a single source of truth in Go.** `Tab` gains `Combined *TabVersion` (additive, `omitempty`); `buildRecipe` bakes the combined grid from each signer's single latest run (`latestAny`, version-blind) via the existing `computeGrid` path. A throwaway grid-signer set is used so the per-recipe series column set (built from the per-version grids) is unchanged — verified: a signer's latest-overall run is exactly its newest-at-that-version run, so combined never surfaces a cell without a series column.
- **The renderer change lands at one chokepoint.** `activeVer()` already drove the whole dashboard (landing/tree/catalog/matrix via `versionView`). Its "all" branch (the default, empty AICR-version selection) now returns the baked combined grid; a specific version returns its strict grid; an absent-version selection falls back to combined, flagged. The grid header labels each case ("all versions" vs `AICR <ver>`, with a "· no `<ver>` runs" suffix on fallback).
- **Cross-version trade-off is explicit.** Combining across versions is intentionally weaker than same-version reproduction; the header tooltip says so and the strict per-version grids remain available.
- **Backward/forward compatible.** `combined` is additive under the unchanged `aicr-corroboration/v1` schema; a renderer loading older data without it degrades to the newest version.

## Testing

```bash
# Run on the affected packages (CI runs the full `make qualify` gate)
GOFLAGS="-mod=vendor" go test -race ./pkg/corroborate/... ./tools/corroborate/...   # ok
golangci-lint run -c .golangci.yaml ./pkg/corroborate/... ./tools/corroborate/...    # 0 issues
GOFLAGS="-mod=vendor" go build ./...                                                  # ok
```

- New `TestGenerateCombinedCrossVersion` proves the core behavior: two sources at different AICR versions → the newest per-version grid is `SINGLE`, the combined grid is `CONFIRMED`.
- Package coverage 91.4% → **91.5%** (combined path 100% covered).
- Renderer JS parses cleanly; the real `activeVer()` and the header-label logic were exercised across all cases (default, strict, absent-version fallback, and legacy data with no `combined` grid).
- Stood up a local deploy from the `bigdemo` evidence tree (with a few sources nudged to lag) — the "all versions" default visibly surfaces sources the newest-version grid hides on 15/21 recipes; selecting a specific version narrows to that release's strict grid.

## Risk Assessment

- [x] **Low** — Isolated to the corroborate dashboard package and its embedded renderer; additive, `omitempty` JSON field; consensus algorithm unchanged; well-tested; trivial to revert.

**Rollout notes:** No migration. `combined` is additive/optional; the renderer is byte-embedded and emitted atomically with the data it reads, and degrades gracefully on older data lacking the field.

## Checklist


pkg/corroborate/model.go
```diff
@@ -131,11 +135,24 @@ type Tab struct {
...
+
+	// Combined is the cross-version "all versions" consensus grid: each distinct
+	// signer's single latest run (version-blind) folded into one grid. It is the
+	// dashboard's default (non-strict) view — it surfaces every source that has
+	// attested the recipe, including sources whose latest run predates the newest
+	// release (which the per-version Versions grids omit). Its own AICRVer is empty
+	// because it spans versions; each row's per-signer Latest.AICRVer still carries
+	// that source's real run version. Consensus here counts agreement ACROSS
+	// versions, which is weaker than same-version reproduction — the renderer makes
+	// that trade-off explicit and the strict Versions grids remain available.
```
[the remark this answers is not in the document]
njhensley: Addressed in 98bf3daa. Clarified both comments: a lagging source stays visible in its own version's strict grid — only the newest strict grid (`Versions[0]`) omits it, which is what the combined default surfaces. Thanks!

pkg/corroborate/renderer/index.html
```diff
@@ -1666,18 +1666,21 @@ <h3>No matching recipes for <span style="color:var(--green)">${esc(currentGroup)
...
+// selected version falls back to combined (flagged _fellback so the header says so).
 function activeVer(recipe){
   const vs = recipe.versions || [];
-  if (!vs.length) return { aicrVer: "", tests: recipe.tests || [], _phaseRollup: recipe._phaseRollup || {} };
-  if (facets.aicr){
+  const combined = recipe.combined || vs[0] || { aicrVer: "", tests: recipe.tests || [], _phaseRollup: recipe._phaseRollup || {} };
+  if (facets.aicr){                                   // strict: a specific AICR version
     const hit = vs.find(v => v.aicrVer === facets.aicr);
     if (hit) return hit;
+    return { ...combined, _fellback: true };          // recipe lacks that version → combined, flagged
```
mchmarny: When a recipe has no runs at the explicitly selected version, this substitutes the cross-version Combined grid. versionView then drops the _fellback marker before landing/tree/catalog consume it, so those views can show cross-version CONFIRMED even though the recipe has zero runs at the selected version. Preserve strict filter semantics by returning an empty/UNTESTED projection or excluding the recipe from version-scoped summaries instead of substituting Combined.

mchmarny (commented):
The combined view is well-contained overall. One version-filter correctness issue remains: recipes without the selected version can silently contribute cross-version consensus to aggregate views. Required CI passes.

pkg/corroborate/renderer/index.html
njhensley: Good catch — fixed in 98d570e0.

`activeVer` no longer substitutes the Combined grid when a specific version is selected and the recipe has no runs at it. It now returns an empty (`_empty`) projection, so `versionView` → the landing/tree/catalog aggregates roll up to **UNTESTED** — the truthful version-scoped state — instead of leaking cross-version consensus. The matrix shows an explicit "no runs at AICR <version>" empty state, and the grid header reads `AICR <version> · no runs`.

Verified the strict-miss path rolls up to UNTESTED (and default→combined / strict-present / legacy-data paths are unchanged). Thanks for the review.

mchmarny (approved):

Issue #1274, opened 2026-06-09
DC1 — Day/night scheduler + capacity request/queue broker


njhensley:
**Parent:** #1264 · **Labels:** `enhancement`, `area/ci` (auto, `.github/**`),
`area/infra` (auto, `infra/**`); **adds `tools/uat-broker/**` `labeler.yml` glob +
dedicated CODEOWNERS entry if a Go helper is introduced** · **Size:** M ·
**Depends on:** none

**Goal.** Own the day/night operating cycle and arbitrate scarce reserved GPU time so
contending runs queue instead of hard-failing, keyed by *reservation* and driven by a
checked-in reservation registry that new infra extends by data edit. The broker is the
single owner of: **(a)** the nightly UAT-batch state machine
(provision→CUJ→evidence→publish→teardown→next, sequential per cloud); **(b)** the
**version-matrix schedule** (time-boxed, `main`-first, then the previous N stable
releases, descending); **(c)** the **daytime human-access lease** (DC8) granted at the
morning handoff and released at the evening teardown; and **(d)** ad-hoc human/CI GPU
requests. The default cadence is **nightly-batch + daytime-lease**; ad-hoc dispatch is
still supported but is the exception, not the norm. Keep it boring — a lease record +
GH-Actions concurrency, no daemon.

**Scope.**
- Implement the **day/night cycle** as the scheduling contract: a nightly batch entry
  point that walks the version-matrix schedule sequentially per cloud (time-boxed,
  `main`-first, oldest releases dropped first when the window is tight), and a
  morning-handoff step that grants the daytime DC8 lease once the batch drains. Both
  the batch and the lease acquire/release the **same reservation lease** so CI and
  human use never overlap on one reservation.
- Add a reservation registry, `infra/uat/reservations.yaml` (sibling of the
  existing `infra/uat-aws-account/`, `infra/uat-gcp-account/`), enumerating rows
  `{name, cloud, reservation-id, accelerator, gpu-count, cluster-config-path,
  test-config-dir}`. Seed with the two known rows: AWS H100
  (`cr-0cbe491320188dfa6`, `tests/uat/aws/cluster-config.yaml`) and GCP H100
  (`h100-a3-mega-phase-1-…`, `tests/uat/gcp/cluster-config.yaml`).
- Implement the lease as a GitHub Actions **concurrency group keyed by the
  reservation name** (e.g. `concurrency.group: uat-${{ inputs.reservation }}`,
[...]

PR #1774, opened 2026-07-15
feat(ci): onboard GB200 AWS UAT reservation

njhensley:
## Summary

Onboard the `aws-gb200` UAT reservation (cross-account GB200 Capacity Block) and generalize the AWS UAT pipeline to select test configs by `<accelerator>-<intent>`.

## Motivation / Context

The AWS UAT pipeline hardcoded `h100-<intent>-config.yaml` and a fixed account, so it could only run the standard-tenancy H100 reservation. GB200 capacity lives in a **separate tenancy** (a time-bounded Capacity Block), so enabling it requires both per-accelerator test-config selection and cross-account role assumption.

Fixes: N/A
Related: #1274

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT CI pipeline (`.github/workflows/uat-*.yaml`), reservation registry (`infra/uat/reservations.yaml`, `pkg/uatbroker`), UAT test configs (`tests/uat/aws`)

## Implementation Notes

- **New reservation `aws-gb200`** in `infra/uat/reservations.yaml`: cross-account GB200 Capacity Block, opted **out** of the nightly batch (`nightly-intents: []`) during bring-up — manual dispatch only via `uat-run.yaml`.
- **New `tests/uat/aws/cluster-config-gb200.yaml`**: Local-Zone, arm64/Grace, Capacity-Block EKS cluster; `deployment.tenancy: 897722687756`.
- **New `tests/uat/aws/tests/gb200-training-config.yaml`**: mirrors the H100 training config; `platform: kubeflow` is load-bearing so the NCCL validators launch via the tolerated Kubeflow Trainer on the all-tainted cluster.
- **Per-accelerator test-config selection**: the AWS pipeline now resolves `TEST_CONFIG` as `<accelerator>-<intent>-config.yaml`; `accelerator` is plumbed from the reservation through `uat-run.yaml`.
- **Account derived, not supplied**: `AWS_ACCOUNT_ID` is now read from the cluster-config's `.deployment.tenancy` (the single source of truth) and exported to `$GITHUB_ENV`, instead of being a caller input. This makes it impossible to assume the OIDC role into a different account than the one being provisioned — the previous design carried both an `account` input and a drift-check assertion; both are removed.

## Testing

```bash
go test ./pkg/uatbroker/... ./tools/uat-broker/...   # ok
golangci-lint run -c .golangci.yaml ./pkg/uatbroker/... ./tools/uat-broker/...   # 0 issues
actionlint .github/workflows/uat-aws.yaml .github/workflows/uat-run.yaml   # clean (pre-existing SC2016 info note unrelated)
```

`TestCommittedRegistryValid` covers the new `aws-gb200` row (cloud + `nightly-intents: []` guard). The cross-account run itself validates end-to-end via a manual `uat-run.yaml` dispatch once the AWS account bootstrap lands (GitHub OIDC provider + role in `897722687756`).

## Risk Assessment

- [x] **Low** — Additive: new reservation row + configs, and a pipeline generalization that is behavior-preserving for the existing H100 row (`accelerator` defaults to `h100`, tenancy derivation yields the same `615299774277`). `aws-gb200` is opted out of the nightly batch, so no nightly capacity is provisioned until explicitly enrolled.

**Rollout notes:** Requires `infra/uat-aws-account` applied in account `897722687756` (GitHub OIDC provider + `github-actions-role-aicr`) before the first GB200 dispatch. Runs must fall inside the Capacity Block's active window.

## Checklist


tests/uat/aws/cluster-config-gb200.yaml
```diff
@@ -0,0 +1,147 @@
...
+        instanceType: m7i.xlarge
+        capacity:
+          desired: 3
+        labels:
+          nodeGroup: system-worker
+          dedicated: system-workload
+      workers:
+        - name: gpu-worker
+          instanceType: p6e-gb200.36xlarge
+          architecture: arm64               # GB200 = Grace/arm64 (arm64 Ubuntu EKS AMI)
```
mchmarny, line 125: blocker: This makes the GPU workers arm64, but the AWS main cell builds the snapshot-agent image only for linux/amd64, and the new test config pins that agent to this gpu-worker pool. The pod cannot start, so the default main UAT path fails before recipe generation. Build and push arm64/multi-arch images here, and verify the validator images scheduled on GPU workers support arm64 too.

.github/workflows/uat-aws.yaml
```diff
@@ -154,7 +167,20 @@ jobs:
           if [[ ! -f "${TEST_CONFIG}" ]]; then
```
mchmarny: blocker: This guard runs for daytime-down and before AWS_ACCOUNT_ID is exported below. If the accelerator/intent config is absent, the job fails before teardown re-authentication can construct the role ARN, potentially leaving the cluster and Capacity Block in use. Derive/export tenancy first, and skip the TEST_CONFIG existence check for daytime-down.

mchmarny (changes requested):
The cross-account wiring is coherent, but two runtime paths remain blocked: main-cell images do not match the GB200 worker architecture, and teardown can fail before account discovery. Required CI is green, but it does not exercise either path.

tests/uat/aws/cluster-config-gb200.yaml
njhensley: Addressed. Main-cell image builds are now arch-aware via a new `BUILD_PLATFORMS` job env — `linux/amd64,linux/arm64` for GB200, `linux/amd64` for h100 (so the common nightly path skips the QEMU cross-build).

- Snapshot-agent: `ko build --platform="${BUILD_PLATFORMS}"` (Go cross-compile, emits a manifest list).
- Validators: converted to `docker buildx build --platform="${BUILD_PLATFORMS}" --push`, with `Set up QEMU` gated on `contains(env.BUILD_PLATFORMS, 'arm64')` and `Set up Docker Buildx`.

On the validator arm64 point: the release pipeline (`on-push.yaml`/`on-tag.yaml`) already builds these exact phases (deployment/performance/conformance/aiperf-bench) for `linux/amd64 linux/arm64` with an `EXPECTED_PLATFORMS` assertion, so the Dockerfiles are proven arm64-capable — this change applies the same multi-arch build to main cells. Release cells are unaffected (they pull the already-multi-arch released images).

.github/workflows/uat-aws.yaml
njhensley: Addressed in the "Validate inputs" step. `AWS_ACCOUNT_ID` is now derived from `.deployment.tenancy` and exported to `$GITHUB_ENV` **first**, before any lifecycle-specific gate can exit — so `daytime-down` teardown can always construct the role ARN and re-authenticate. The `TEST_CONFIG` existence check is now guarded with `[[ "${LIFECYCLE}" != "daytime-down" && ! -f "${TEST_CONFIG}" ]]`, so a missing/absent config no longer fails the job before teardown and can no longer strand the cluster + Capacity Block.

mchmarny (approved):

Issue #1783, opened 2026-07-16
recipe: validate constraint measurement paths at load time


mchmarny:
Follow-up to #1542 (design section 10). `pkg/constraints.Evaluate` distinguishes malformed constraint paths (`ErrCodeInvalidRequest`, now fails resolution closed) from values absent in the snapshot (`ErrCodeNotFound`, graceful exclusion by design). A grammatically valid but typo'd constraint path evaluates as NotFound and still degrades silently for unstated dimensions. Validate constraint paths against the measurement catalog when loading recipe data so typos fail at load time instead of masquerading as missing measurements.

mchmarny:
Triaged: added to the **v1** milestone (API Stability).

Moving typo'd constraint paths from silent `NotFound` degradation to a load-time failure is a fail-closed semantics change on a frozen surface (ROADMAP §2). Pre-v1 work.

mchmarny:
Triaged: **P2 → P1** (Ready).

This is the cleanest false-PASS on the board. A grammatically valid but typo'd constraint path evaluates as `ErrCodeNotFound`, which the resolver treats as "value absent from the snapshot — exclude gracefully". The recipe then resolves successfully with that constraint silently never evaluated — a recipe passing because it skipped a gate.

It is also an input check for #2116 (`aicr recipe validate`): shipping the contributor-facing validator without it means shipping a validator that misses the highest-value class of authoring typo.

Issue #1782, opened 2026-07-16
recipe: joint-coverage formulation to subsume requireOSIfNeeded


mchmarny:
Follow-up to #1542 (design 4.1). The per-dimension coverage post-condition does not subsume the `requireOSIfNeeded` guard: coverage permits `service` and `accelerator` to be honored by separate applied overlays, while the guard demands one overlay carry both before the OS-agnostic tier counts as served. On current embedded data the two agree only by accident (OKE/GKE gate their whole trees; other services have OS-agnostic mid-tier accel overlays) — external providers can present catalogs where they diverge. Design a joint-coverage rule that genuinely subsumes the guard, then delete it. `TestBuildRecipeResult_GuardAndCoverageComposition` pins the synthetic catalog where the guard fires and coverage alone would pass.

mchmarny:
Triaged: added to the **v1** milestone (API Stability).

Recipe resolution semantics are part of the frozen schema surface (ROADMAP §2). The issue itself notes the coverage post-condition and `requireOSIfNeeded` agree on current embedded data "only by accident" and can diverge for external providers — that has to be resolved before the schema is declared stable.

Issue #1781, opened 2026-07-16
recipe: demote nodes criteria dimension to metadata-only


mchmarny:
Follow-up to #1542 (design 4.3). `nodes` participates in `Criteria.Matches` and `Specificity()` but no overlay gates on it, and it is exempt from the coverage post-condition — a stated `--nodes` value carries no coverage guarantee. Remove it from matching/specificity (making it advisory metadata everywhere) after auditing external providers (internal recipes repo) for nodes-gated overlays: a provider with such overlays would have them match every query if the dimension is dropped unaudited.

Also fold in: `completionTuple` in `pkg/recipe/coverage.go` ignores `nodes` when computing completion suggestions, so an external provider's nodes-gated overlay could be suggested as a completion that still would not match due to a nodes conflict (zero impact on embedded data).

mchmarny:
Triaged: added to the **v1** milestone (API Stability).

Removing `nodes` from `Criteria.Matches` and `Specificity()` changes recipe-schema semantics, and the recipe schema is one of the four surfaces frozen at v1 (ROADMAP §2). Doing this after `v1.0.0` costs a major version bump; doing it before costs nothing.

PR #1784, opened 2026-07-16
fix(recipe): enforce stated-criteria coverage on recipe resolution

### LINKED ISSUE #1542: fix(recipe): OS-required guard ignores platform/intent, allowing partial recipe resolution
## Summary

`requireOSIfNeeded` (`pkg/recipe/metadata_store.go`) short-circuits on the first matching overlay keyed by service+accelerator only. A query like `oke + gb200 + kubeflow` (no OS) passes the guard because a generic GB200+OKE overlay exists, but the returned recipe omits the kubeflow platform overlay (which lives on an OS-specific leaf). The caller gets a valid-looking recipe that is silently incomplete.

## Root cause

The success-shortcut loop in `requireOSIfNeeded` checks whether any already-resolved overlay matches service+accelerator — it does not account for platform or intent. So a query with a platform or intent dimension that only resolves via an OS-specific leaf can slip through the guard and return a partial result.

## Correct fix

Compare overlay coverage with vs without OS — if adding an OS produces a strictly richer match set (more leaves, more componentRefs), the guard should require it. This is non-trivial to get right and needs careful testing against the full overlay matrix.

## Context

Pre-existing limitation independent of the L40S work in PR #1510, where it was deferred to avoid a half-measure. Surfaced during yuanchen8911's round-4 review.

mchmarny:
## Summary

Recipe resolution now fails with an actionable `ErrCodeInvalidRequest` (HTTP 400) when a query-stated criteria dimension is not honored by any applied overlay, instead of silently returning a partial recipe that echoes criteria it did not apply. Constraint evaluation additionally fails closed on non-NotFound evaluator errors.

## Motivation / Context

The OS-required guard (`requireOSIfNeeded`) models only service+accelerator/OS, while the criteria matcher treats all dimensions uniformly — so a stated `platform` or `intent` could silently contribute nothing. Live repro on `main`: `aicr recipe --service eks --accelerator h100 --intent training --platform kubeflow` exited 0 with no `kubeflow-trainer` content while echoing `platform: kubeflow` in its criteria.

Fixes: #1542
Related: #1781, #1782, #1783 (follow-ups: nodes demotion, guard subsumption, constraint-path load-time validation)

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- **Coverage post-condition** (`pkg/recipe/coverage.go`): every stated dimension (service, accelerator, intent, os, platform — `nodes` is exempt/advisory) must appear with its exact value on ≥1 applied overlay (leaf or inheritance-chain ancestor). Runs on the FINAL applied set in both `BuildRecipeResult` and `BuildRecipeResultWithEvaluator` (after constraint exclusion and the mixin-failure fallback rebuild).
- **Actionable errors**: minimal completion tuples per uncovered dimension — `platform 'kubeflow' for criteria(service=eks, accelerator=h100, intent=training) requires os (valid: ubuntu)` vs `no recipe provides platform 'nim' for ...` — plus a machine-readable `details.uncovered` array (`dimension`, `requestedValue`, `validCompletions`); evaluator path adds `excludedOverlays`/`constraintWarnings` context.
- **`requireOSIfNeeded` stays and runs first**: per-dimension coverage does not subsume its joint service+accelerator OS gating (see #1782). Pinned by `TestBuildRecipeResult_GuardAndCoverageComposition`.
- **Fail-closed evaluation**: non-NotFound evaluator errors (malformed constraint path/expression → `ErrCodeInvalidRequest`, internal failures → `ErrCodeInternal`) now abort resolution with their codes preserved, in both per-overlay and mixin evaluation. `ErrCodeNotFound` ("value not found in snapshot") remains the designed graceful-exclusion signal, so `recipe --snapshot` degradation is unchanged and no new 404 surfaces on `/v1/recipe`.
- **Golden projection matrix** (`pkg/recipe/testdata/coverage_golden.yaml`, 415 entries): pins the resolution outcome of every projection of every overlay's criteria. Any registry/overlay data change that flips an outcome fails `TestCoverageGoldenMatrix`; regenerate deliberately with `AICR_UPDATE_GOLDEN=1 go test ./pkg/recipe/ -run TestCoverageGoldenMatrix`.
- **Blast-radius fixes**: 9 downstream test fixtures (pkg/client/v1, pkg/server, pkg/cli, chainsaw) queried silent partials; each individually triaged. One was genuinely wrong pre-branch (`--service gke` override onto an `os=ubuntu` base — no gke+ubuntu overlay exists).

## Testing

```bash
make qualify        # full gate: test + lint + e2e (23/23 chainsaw) + scan — clean
make test-coverage  # 79.7% vs 75% floor — pass
make kwok-e2e RECIPE=eks-training   # 35/35 pods scheduled — pass
```

Coverage deltas vs `origin/main`: `pkg/recipe: 87.3% → 88.0% (+0.7%)`; `pkg/client/v1`, `pkg/server`, `pkg/cli` unchanged. Every function in `pkg/recipe/coverage.go` exercised.

Live verification: the #1542 repro now exits non-zero with the completion-suggestion error; `--os ubuntu` and the generic-tier query (`eks+h100+training`, no platform) still exit 0.

## Risk Assessment

**Rollout notes:** Intentional contract change: queries that previously returned 200 with silently-partial recipes now return 400 with `details.uncovered`; malformed constraints in recipe data now hard-fail resolution instead of silently excluding an overlay. No transition flag (pre-1.0, per approved design). External recipe-data consumers (internal recipes repo) notified on #1542. Revert is a clean single-commit revert.

## Checklist


api/aicr/v1/server.yaml
```diff
@@ -228,7 +228,14 @@ paths:
...
         "400":
-          description: Invalid request parameters
+          description: >
+            Invalid request: malformed criteria, allowlist violation, or — new —
+            a stated criteria dimension not honored by any applicable recipe
+            overlay (uncovered dimension). Uncovered-dimension errors carry a
+            machine-readable `details.uncovered` array (dimension,
+            requestedValue, validCompletions) and, on snapshot-driven
+            resolution, `details.excludedOverlays` / `details.constraintWarnings`
+            when constraint exclusion occurred during resolution.
```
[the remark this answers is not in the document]
mchmarny: Done in 16392773 — GET and POST /v1/query 400 responses now mirror the `details.uncovered` contract, and `docs/user/api-reference.md` cross-references the /v1/recipe error shape from the query section.

pkg/cli/config_integration_test.go
```diff
@@ -129,8 +129,12 @@ func TestRecipeCmd_ConfigFlag_FlagOverride(t *testing.T) {
 	cfgPath := writeYAML(t, "config.yaml", testRecipeConfig)
 	root := newRootCmd()
 
+	// aks (like eks) is covered by accelerator=h100/intent=training/os=ubuntu;
+	// gke instead requires os=cos, so overriding to gke here would trip the
+	// criteria-coverage post-condition on an incoherent os/service pairing
+	// rather than exercise the flag-override behavior this test targets.
 	err := root.Run(context.Background(), []string{
-		name, "recipe", "--config", cfgPath, "--service", "gke", "-o", "-",
+		name, "recipe", "--config", cfgPath, "--service", "aks", "-o", "-",
```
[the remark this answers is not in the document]
mchmarny: Done in 3f7894bb — the test now writes the recipe to a file and asserts `service: aks` is present (and `service: eks` absent), so an ignored flag override fails the test.

pkg/recipe/metadata_store.go
```diff
@@ -757,6 +763,11 @@ func (s *MetadataStore) evaluateMixinConstraints(
 			continue // already evaluated per-overlay
 		}
 		result := evaluator(constraint)
+		if result.Error != nil && !isNotFoundEvalError(result.Error) {
+			// Fail closed on non-NotFound evaluation errors (design 5.2);
+			// the error already carries its structured code.
+			return mixinEvalResult{}, result.Error
+		}
```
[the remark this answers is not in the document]
mchmarny: Done in be0f69aa — both fail-closed branches now route through `aicrerrors.PropagateOrWrap`: an existing `*StructuredError` propagates with its code preserved (no double-wrap), and a plain error is wrapped as `ErrCodeInternal` with the cause retained. Unit tests cover both evaluation paths.

mchmarny:
**Update (round 1 fixes pushed, `be0f69aa..16392773`):**

The first CI run surfaced a real design-consequence failure worth calling out: on the nvkind GPU runners, the snapshot fingerprint detects `os=ubuntu` on Kind nodes, but the Kind overlay tree is deliberately OS-agnostic — strict coverage rejected the resolution (`no recipe provides os 'ubuntu' ...`) even though the OS-wildcard Kind recipe is the correct answer for that node. The committed golden matrix cannot catch this class: snapshot-derived values come from outside the catalog.

**Amendment (maintainer-approved):** the recipe engine stays strict, but the shared CLI snapshot path (recipe/query/mirror) now treats fingerprint-derived dimensions as advisory — when a coverage failure's uncovered dimensions are ALL snapshot-derived (not set by config or flags), they are relaxed with a warning and resolution retries once; any user-stated uncovered dimension still fails hard. `pkg/client/v1.ResolveRecipeFromSnapshot` takes caller-supplied criteria and derives nothing, so it is unchanged. Documented in `docs/user/cli-reference.md` and `docs/contributor/recipe.md`.

Also in this round: unstructured evaluator errors are now coded via `PropagateOrWrap` (CodeRabbit), `/v1/query` 400 docs mirror `details.uncovered` (CodeRabbit), the CLI override test asserts the override actually lands (CodeRabbit), and the e2e from-snapshot test uses a coverable intent on the GPU-less Kind cluster. `make qualify` clean.

mchmarny:
**GPU test root cause (round 2, `cd72d497`):** the failing H100 jobs pass `--os ubuntu` as an explicit CLI flag in `.github/actions/runtime-install/generate-recipe.sh` — not via snapshot detection, so the snapshot-relaxation correctly did not apply (user-stated dimensions are strict by design). The Kind overlay tree is deliberately OS-agnostic, so that flag never contributed content pre-branch (silent no-op) and is now correctly rejected as an uncovered stated dimension. Fixed by dropping `--os ubuntu` from the GPU test action and the same latent pattern in `tools/mirror-e2e`; verified locally by running the actual script for both failing CI shapes (inference+dynamo → 16 components, training → 12 components).

ayuskauskas (dismissed):
Approving. Faithful implementation of the #1542 dimension-uniform coverage post-condition, and retaining `requireOSIfNeeded` rather than subsuming it (as I'd claimed on the issue) is the correct call — coverage is satisfied when `service`/`accelerator` are honored by *separate* overlays, while the joint OS guard still needs to fire. Nicely pinned by the golden matrix.

One non-blocking follow-up on the CLI snapshot-relax path ([`pkg/cli/query.go`](https://github.com/NVIDIA/aicr/blob/e4d113233a7184f54584628aed29f951840d333a/pkg/cli/query.go)):

`uncoveredCoverageDimensions` depends on (a) the coverage error being the *first* `StructuredError` in the chain and (b) `Context["uncovered"]` asserting to exactly `[]map[string]any`. That holds today — the facade (`aicr.go`) and `buildWithStore` (`builder.go`) both return the builder's error unwrapped — but it's only unit-tested with a synthetic error (`mkCoverageErr`), and the e2e scripts were changed to drop `--os ubuntu` entirely, which sidesteps the exact scenario this path exists for. So relaxation is unexercised end-to-end, and a future re-wrap of the error (or a JSON boundary) would silently disable it with no failing test.

Suggestion: add an integration test through `ResolveRecipeFromSnapshot` that drives a real coverage failure → relax → retry, and/or make extraction robust by walking the chain for any `StructuredError` carrying `uncovered` instead of asserting on the first match.

mchmarny:
Both follow-ups addressed in 2715f3fd:

- `uncoveredCoverageDimensions` no longer stops at the first `StructuredError`: it walks every node in the wrap chain and identifies the coverage error by its own code + `uncovered` context, so an intermediate re-wrap between the builder and the CLI can't silently disable relaxation. It also accepts the decoded-JSON payload shape (`[]any` of `map[string]any`) alongside the in-process `[]map[string]any`.
- `TestRelaxSnapshotDerivedCoverage_RealResolutionError` drives the exact scenario the e2e scripts sidestepped: a snapshot fingerprinting to `service=kind` + `os=ubuntu` produces a real coverage failure from `ResolveRecipeFromSnapshot` against the embedded catalog, extraction is asserted against the builder's actual error shape, and the relaxed retry resolves. A future re-wrap or payload-shape drift now fails this test instead of silently disabling relaxation. Wrapped (same/different code) and JSON-shaped payloads are pinned in unit tests as well.

njhensley (commented):
# 📋 Multi-Persona Review

**Method:** 5 independent persona reviewers (Correctness, Domain/Architecture, API-contract, Test-coverage, CI-DX) → adversarial senior meta-reviewer that re-derived every finding from the resolved code at head `85eb09e`. 10 candidates raised; 8 survived, 2 refuted.

## Overall assessment

This is a well-engineered, well-tested bug fix with a sound core invariant. The coverage algorithm is correct: `appliedOverlays` genuinely includes inheritance-chain ancestors (so ancestor-supplied coverage counts) but excludes constraint-excluded overlays (so "honored only by an excluded overlay" fails loudly, as intended). The `requireOSIfNeeded` / per-dimension-coverage composition has no gap or double-gating. The shallow-copy retry in the CLI is safe (`Criteria` is all scalars). Error-code choice (400) is consistent with the sibling OS guard. Fail-closed evaluation aligns with the project's own anti-pattern rules.

**No blockers.** The one 🟠 is a *test-coverage* gap on the PR's central safety property, not a code defect. The 🟡s are accuracy/UX refinements.

**Recommendation: Approve with comments** — worth addressing the 🟠 (test the `touched` invariant) and the OpenAPI overpromise before or shortly after merge.

- **Coverage-set correctness** — `appliedOverlays` correctly includes inheritance-chain ancestors but not excluded overlays; both `verifyCriteriaCoverage` call sites run on the final applied set; single-retry relaxation is provably sufficient.
- **`next := *criteria` aliasing** — `Criteria` is all scalar enum/int fields, no slices/pointers; the shallow copy fully isolates the retry.
- **`touched` keys ↔ coverage dimension names** — identical constants in `consts.go` and `coverage.go`; `nodes` intentionally absent from both.
- **No internal leak** — `verifyCriteriaCoverage` uses `NewWithContext` (no `Cause`), so `WriteErrorFromErr`'s 4xx `details.error` branch never fires; message is only completion clauses. Struct↔JSON tags align.
- **Guard composition** — `requireOSIfNeeded` (joint service+accel) and per-dimension coverage neither subsume each other nor double-gate.
- **`PropagateOrWrap` / wrap-chain walk** — no code-clobbering; `uncoveredCoverageDimensions` inspects every chain node and won't false-match a bare `ErrCodeInvalidRequest`.
- **Operational call sites** — the 4 changed shell/chainsaw files adapt correctly (drop stated-but-undistinguished `--os`, retarget `gke`→`aks`); no `|| true` masking; no unmodified in-repo resolver call site breaks post-merge; no shell-injection.
- **Blast-radius fixtures** — spot-checked; none weaken assertions (`config_integration_test` was *strengthened*).
- **~~gpu-alias `gb200`-on-`gke` doc example returns a franken-recipe~~** — **refuted**: no `gb200`+`gke` overlay exists, so `service`/`os` are uncovered → this actually returns **400**, not a 200 partial. (Minor residue: the `api-reference.md:146` example's comment implies `os=cos` makes it valid when it would 400 on `service` — worth a re-check of the intended illustration.)
- **~~e2e `--intent training→inference` narrows coverage for no reason~~** — **refuted**: the flip is almost certainly coverage-forced (mirroring the sibling `--os` drops); the only residue is that `run.sh` lacks the explanatory comment its siblings got.

## Summary

```
🔴 Blocker 0 | 🟠 Major 1 | 🟡 Minor 5 | 🔵 Nitpick 2     Recommendation: Approve with comments
```

| Tier | Finding | Location |
|------|---------|----------|
| 🟠 | `touched`-map invariant untested (silent-relax regression could ship green) | `pkg/cli/recipe.go:374` |
| 🟡 | Fail-closed aborts on unparseable snapshot value (snapshot path only) | `pkg/recipe/metadata_store.go:1195` |
| 🟡 | Misleading "no recipe provides" when sole overlay was excluded | `pkg/recipe/coverage.go:248` |
| 🟡 | OpenAPI/docs promise HTTP fields that can't be emitted | `api/aicr/v1/server.yaml:237` |
| 🟡 | Snapshot-relaxation policy CLI-only, undocumented SDK asymmetry | `pkg/cli/query.go:237` |
| 🔵 | Golden matrix doesn't guard completion-suggestion content | `pkg/recipe/coverage_matrix_test.go:32` |
| 🔵 | Completions constraint-unaware + multi-element wording unasserted | `pkg/recipe/coverage.go:88` |

<sub>Generated via multi-persona review + adversarial meta-review. Findings independently re-derived from resolved code; refuted candidates listed above for transparency.</sub>

pkg/cli/recipe.go
```diff
@@ -346,7 +367,11 @@ func mergeCriteriaFromCmdAndConfig(cmd *cli.Command, cfg *appcfg.AICRConfig, reg
 // applyCriteriaOverrides applies CLI flag overrides to criteria, resolving each
 // enum value against the supplied per-provider registry. Logs a warning when a
 // flag overrides a value detected from the snapshot.
-func applyCriteriaOverrides(cmd *cli.Command, criteria *recipe.Criteria, reg *recipe.CriteriaRegistry) error {
+//
+// touched records which of the 5 coverage dimensions this call set from a CLI
+// flag — see applyCriteriaFromConfig. Callers not tracking that distinction
+// may pass nil.
+func applyCriteriaOverrides(cmd *cli.Command, criteria *recipe.Criteria, reg *recipe.CriteriaRegistry, touched map[string]bool) error {
```
njhensley, line 374: **🟠 Major — The `touched`-map invariant — "never relax a user-stated dimension" — has zero test coverage**

The safety property that a user-stated dimension is never silently relaxed hinges entirely on `applyCriteriaFromConfig`/`applyCriteriaOverrides` calling `markCriteriaTouched` for each dimension they set. But every production call site in tests passes `touched=nil` (`config_integration_test.go:63,94`, `recipe_test.go:582`), and the relax tests (`snapshot_coverage_test.go`) feed hand-built literal `touched` maps straight into `relaxSnapshotDerivedCoverage`. Nothing asserts these apply functions actually *produce* the marks.

**Blast radius:** A future edit that drops or mis-keys one `markCriteriaTouched` call (e.g. `--service` forgets its mark) makes the CLI treat a user-stated `--service gke` as snapshot-derived and silently clear it — shipping a recipe for a different service than requested, the exact silent-partial this PR exists to prevent — with CI green.

**Fix:** Add a table test driving `applyCriteriaOverrides`/`applyCriteriaFromConfig` with a live `touched := map[string]bool{}`, asserting each stated flag/config field lands a mark and unset ones do not. Ideally one end-to-end case through `buildRecipeFromCmdWithConfig` proving a stated uncovered dimension propagates the error instead of being relaxed.

pkg/recipe/metadata_store.go
```diff
@@ -1148,8 +1186,16 @@ func (s *MetadataStore) evaluateOverlayConstraints(overlay *RecipeMetadata, eval
 		result := evaluator(constraint)
 
 		switch {
+		case result.Error != nil && !isNotFoundEvalError(result.Error):
+			// Fail closed: a malformed constraint or internal evaluation
+			// failure must not silently degrade the recipe (issue #1542,
+			// design 5.2). ConstraintEvaluatorFunc returns a plain error;
+			// propagate as-is when it already carries a structured code,
+			// otherwise wrap so it doesn't reach the server layer uncoded.
+			return false, nil, aicrerrors.PropagateOrWrap(result.Error, aicrerrors.ErrCodeInternal, "constraint evaluation failed")
```
njhensley, line 1195: **🟡 Minor — Fail-closed evaluation turns a present-but-unparseable snapshot value into a hard build abort (snapshot path only)**

Only `ErrCodeNotFound` degrades gracefully; every other evaluator code now aborts the whole build. When a version operator (>=,>,<=,<) runs against a value present in the snapshot but not parseable as a version, `pkg/constraints/constraint.go` emits `ErrCodeInvalidRequest` ("cannot parse actual version") and `pkg/constraints/evaluate.go:64` re-wraps it as `ErrCodeInternal` — so this overlay used to be excluded-with-warning pre-PR but now aborts resolution.

**Blast radius:** Narrow and non-HTTP: reachable only via `aicr recipe --snapshot` / SDK `ResolveRecipeFromSnapshot` (the server never uses the evaluator path). Needs malformed snapshot data; the cause chain still names the bad value (not opaque). Failing closed here matches project policy — flagging as a behavior change, not a defect. The genuine sub-defect is the pre-existing `evaluate.go:64` code-clobber (InvalidRequest→Internal), which this PR makes consequential.

**Fix:** Optional: classify "actual value unusable" the same as NotFound (degrade), reserving abort for recipe-author parse errors and true internal faults.

pkg/recipe/coverage.go
```diff
@@ -0,0 +1,292 @@
...
+	}
+	return ""
+}
+
+// completionClause renders one uncovered dimension's error clause.
+// Single-field wording is used ONLY when all minimal tuples are singletons
+// over the same dimension (design 5.1).
+func completionClause(criteria *Criteria, dimName, want string, tuples []map[string]string) string {
+	stated := criteria.String()
+	if len(tuples) == 0 {
```
njhensley, line 268: **🟡 Minor — Misleading "no recipe provides X" error when the sole covering overlay was constraint-excluded**

On the evaluator path, a fully-specified query whose only matching overlay is constraint-excluded produces an empty completion tuple (every other dimension is stated). `minimalTuples` drops empty tuples, so `completionClause` hits the `len(tuples)==0` branch and prints "no recipe provides platform 'X' for ..." — but a recipe *does* provide it; it was excluded by a failing constraint.

**Blast radius:** Confusing operator-facing error on a legitimate snapshot resolution where the environment simply doesn't meet a constraint. The truthful story is in `details.excludedOverlays`, so the primary message contradicts its own context. Cosmetic/UX only — no incorrect resolution result.

**Fix:** When `len(tuples)==0` and this dimension's only providers are in `excluded`, word it as "provided only by overlays excluded by failing constraints (see excludedOverlays)".

api/aicr/v1/server.yaml
```diff
@@ -228,7 +228,14 @@ paths:
...
                         cudaVersion: "13.1"
         "400":
-          description: Invalid request parameters
+          description: >
+            Invalid request: malformed criteria, allowlist violation, or — new —
+            a stated criteria dimension not honored by any applicable recipe
+            overlay (uncovered dimension). Uncovered-dimension errors carry a
+            machine-readable `details.uncovered` array (dimension,
+            requestedValue, validCompletions) and, on snapshot-driven
+            resolution, `details.excludedOverlays` / `details.constraintWarnings`
```
njhensley: **🟡 Minor — OpenAPI/docs promise `details.excludedOverlays` / `constraintWarnings` the HTTP API can never emit**

These two context keys are populated only on the evaluator path (`BuildRecipeResultWithEvaluator`). Both HTTP handlers (`recipe_handler.go:144,265`) call `ResolveRecipeFromCriteria` → `BuildRecipeResult` (non-evaluator), and there is no snapshot-upload endpoint — so over HTTP those fields are never returned. (This description text repeats at server.yaml:423,638,771 and docs/user/api-reference.md:224.)

**Blast radius:** A client generating types/handlers from the OpenAPI contract codes for 400-response fields this API never returns. `details.uncovered` IS emittable, so scope the correction to the two constraint fields.

**Fix:** Drop the `excludedOverlays`/`constraintWarnings` clause from the four HTTP 400 descriptions (keep it in docs/contributor/recipe.md, which correctly scopes it to pkg/recipe), or mark it CLI/SDK-only.

pkg/cli/query.go
```diff
@@ -204,6 +220,134 @@ func buildRecipeFromCmdWithConfig(ctx context.Context, cmd *cli.Command, cfg *ap
...
+// a single retry.
+//
+// Design rationale: the recipe engine stays strict (pkg/recipe never
+// relaxes), but a Kind-style overlay tree can be deliberately OS-agnostic
+// while the snapshot fingerprint still detects a concrete os value on the
+// node — no recipe content distinguishes that detected value, so failing the
+// build on it would reject a legitimate query. Dimensions the caller
+// explicitly stated are never relaxed: if any uncovered dimension was
+// user-stated, ok is false and the caller must propagate err unchanged.
+func relaxSnapshotDerivedCoverage(err error, criteria *recipe.Criteria, touched map[string]bool) (relaxed *recipe.Criteria, ok bool) {
```
njhensley, line 237: **🟡 Minor — Snapshot-relaxation policy lives in the CLI, not the shared facade — SDK parity gap**

`relaxSnapshotDerivedCoverage` is CLI-private; `pkg/client/v1` `ResolveRecipeFromSnapshot` returns strict coverage failures with no relaxation. The facade is documented as the shared SDK for "CLI, server, and external Go callers," yet an external caller replicating `aicr recipe --snapshot` against a deliberately OS-agnostic overlay tree (Kind-style) gets a hard failure the CLI silently relaxes.

**Blast radius:** Undocumented behavioral asymmetry for third-party Go callers. Not a defect — the `touched` provenance only exists at the CLI layer — but a design smell.

**Fix:** Document on `ResolveRecipeFromSnapshot` that the coverage post-condition is strict and snapshot-derived relaxation is CLI-layer policy, or expose an optional `relaxDerivedDimensions` on the facade.

pkg/recipe/coverage_matrix_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"testing"
+
+	aicrerrors "github.com/NVIDIA/aicr/pkg/errors"
+	"gopkg.in/yaml.v3"
+)
+
+const goldenPath = "testdata/coverage_golden.yaml"
+
+// coverageClassification is the golden outcome for one projected query.
+type coverageClassification struct {
```
njhensley, line 32: **🔵 Nitpick — Golden matrix pins resolution outcome but not completion-suggestion content**

`coverageClassification` captures Outcome/Uncovered(names)/RequiresOS only — not `validCompletions` tuples or the rendered clause. So the suggestion machinery (`completionTuplesFor`/`minimalTuples`/`completionTuple`/`sameDimensionSingletons`/`completionClause`) is guarded against real catalog data only insofar as it affects the outcome, which it doesn't. (Credit: the golden does fail closed on missing/flipped/removed projections.)

**Blast radius:** A regression producing wrong or empty completion suggestions on the real catalog passes the golden matrix silently; only 3-4 synthetic unit cases guard it.

**Fix:** Add `ValidCompletions` (canonical `tupleKey` strings) to the classification struct so completion regressions on the real catalog flip a golden entry.

pkg/recipe/coverage.go
```diff
@@ -0,0 +1,292 @@
...
+	return uncovered
+}
+
+// completionTuplesFor returns the minimal sets of additional (unstated)
+// dimension values under which some overlay would cover dimName=want.
+// An overlay contributes a candidate tuple when it carries dimName=want and
+// does not conflict with any stated query dimension. Empty tuples (overlay
+// covers the value with no additions — possible only when it was
+// constraint-excluded) are dropped; the exclusion context tells that story.
+func (s *MetadataStore) completionTuplesFor(criteria *Criteria, dimName, want string) []map[string]string {
```
njhensley, line 88: **🔵 Nitpick — Completion suggestions are constraint-unaware; multi-element error wording is unasserted**

Two trivial gaps: (a) `completionTuplesFor` iterates all overlays with no exclusion awareness, so a suggested completion can itself be constraint-excluded (a suggestion that leads in a circle); (b) tests (coverage_test.go) only assert single-tuple / single-value error wording — the `(...), (...)` multi-tuple join and the `valid: a, b` multi-value list in completionClause/sameDimensionSingletons are never asserted.

**Blast radius:** A regression in the join/ordering of multi-tuple or multi-value error messages (user-facing guidance) ships unnoticed; a suggested completion may be unreachable against the caller's snapshot. Both low impact.

**Fix:** Add one verifyCriteriaCoverage case yielding >=2 minimal tuples (assert the full clause) and one same-key 2-value singleton (assert `valid: a, b`); optionally deprioritize completions whose only providers are excluded.

pkg/cli/recipe.go
mchmarny: Done in 760cf2a3 — `pkg/cli/touched_invariant_test.go` adds `TestApplyCriteriaOverridesMarksTouched` and `TestApplyCriteriaFromConfigMarksTouched`: table tests over all five dimensions with a live `touched` map, asserting each stated flag/config field lands exactly its mark and unset fields (plus `--nodes`) land none. Also added the end-to-end pair through `root.Run` with a real snapshot file: `TestRecipeCmd_Snapshot_StatedDimensionNotRelaxed` (user-stated `--os rhel` on a kind-fingerprinting snapshot propagates the coverage error with `uncovered=[os]`) and `TestRecipeCmd_Snapshot_DerivedDimensionRelaxed` (same snapshot without the flag relaxes the derived os and resolves). A dropped or mis-keyed `markCriteriaTouched` call now fails four tests instead of shipping green.

pkg/recipe/metadata_store.go
mchmarny: Keeping fail-closed for present-but-unparseable actual values, but the genuine sub-defect is fixed in 760cf2a3: `pkg/constraints/evaluate.go` now routes `parsed.Evaluate` errors through `errors.PropagateOrWrap`, so the version parser's `ErrCodeInvalidRequest` ("cannot parse actual version", naming the value) reaches the recipe engine un-clobbered instead of masquerading as `ErrCodeInternal`. New `TestEvaluate` in pkg/constraints pins the full eval error-code contract (NotFound for absent measurements, InvalidRequest preserved for unparseable actuals). On the degrade option: exclusion-with-warning on unusable snapshot data would let a corrupt value silently flip recipe content — the same masked-partial shape this PR removes — so abort-with-a-named-value seems the right trade for this narrow CLI/SDK-only surface. Happy to revisit if real snapshots hit it.

pkg/recipe/coverage.go
mchmarny: Done in 760cf2a3 — when a dimension has zero completion tuples but an excluded overlay carries the requested value (new `excludedOverlayProvides` check), the clause now reads `<dim> '<v>' for <criteria> is provided only by overlays excluded by failing constraints (see excludedOverlays)` instead of "no recipe provides ...", so the primary message no longer contradicts its own context. Pinned by the `excluded-only provider wording` subtest, which also asserts the old wording is absent.

api/aicr/v1/server.yaml
mchmarny: Done in 760cf2a3 — all four HTTP 400 descriptions in server.yaml plus the POST /v1/recipe error section in docs/user/api-reference.md now state that `excludedOverlays`/`constraintWarnings` are attached only on snapshot-driven resolution (CLI `--snapshot` / Go SDK) and are never emitted by this HTTP API. docs/contributor/recipe.md untouched (correctly scoped to pkg/recipe). One correction on the refuted-finding residue: the `gpu=gb200&service=gke&os=cos` example at api-reference.md:146 is accurate — verified live against the embedded catalog: with `os=cos` it resolves (per-dimension coverage: `gb200-any` covers the accelerator, `gke-cos` covers service+os), and without os it fails with `specify an OS (valid: cos)`, exactly what the example's comment claims.

pkg/cli/query.go
mchmarny: Done in 760cf2a3 — documented on `ResolveRecipeFromSnapshot`: the coverage post-condition is STRICT at the facade; snapshot-derived relaxation is CLI-layer policy (only the CLI knows which dimensions the user explicitly stated vs derived from the fingerprint), and callers replicating `aicr recipe --snapshot` must clear derived-not-received uncovered dimensions and retry themselves. Chose documentation over a `relaxDerivedDimensions` option: the facade has no touched-provenance to act on, so such an option would relax user-supplied criteria it cannot distinguish — reintroducing the silent-partial risk this PR removes.

pkg/recipe/coverage_matrix_test.go
mchmarny: Done in 760cf2a3 — `coverageClassification` now carries `ValidCompletions` (per-dimension canonical `tupleKey` strings, preserving `minimalTuples` order) and the comparison loop diffs it, so a wrong/empty/reordered completion suggestion on the real catalog flips a golden entry. Golden regenerated (~2k lines of pinned suggestion content). The regeneration also folded in the `h100-aks-ubuntu-training-slurm` overlay that landed on main — the merge commit's golden was missing those projections, which this update fixes.

pkg/recipe/coverage.go
mchmarny: Done (wording assertions) in 760cf2a3 — `multi-tuple wording joins all supported combinations` asserts the full `(...), (...)` clause verbatim and `same-dimension singletons list all valid values sorted` asserts `valid: rhel, ubuntu`. Skipped the excluded-completion deprioritization deliberately: completion suggestions describe the catalog, constraint exclusion is snapshot-specific, and the empty-tuple case now gets explicit excluded-overlays wording (per your coverage.go:248 comment), which covers the circular-suggestion path; deprioritizing non-empty tuples would make suggestions vary by snapshot for little gain.

mchmarny:
**Multi-persona review round addressed in 760cf2a3** — all 7 inline threads replied to and resolved (1 🟠, 4 🟡, 2 🔵; details in each thread). Notes on the review-body residues that had no inline thread:

- **api-reference.md:146 example (`gpu=gb200&service=gke&os=cos`)** — verified live: it resolves with `os=cos` (per-dimension coverage: `gb200-any` covers the accelerator, `gke-cos` covers service+os) and fails without os with `specify an OS (valid: cos)`. The example and its comment are accurate as written; no change.
- **e2e `run.sh` intent flip** — confirmed coverage-forced (`kind-inference` is the only accelerator-less kind intent overlay; training exists only h100-qualified, and the GPU-less kind cluster can't state an accelerator). Added the explanatory comment both siblings got.
- **Golden freshness** — regenerating for `ValidCompletions` also surfaced that the `main` merge (254a91f6) brought `h100-aks-ubuntu-training-slurm` without its golden projections (TestCoverageGoldenMatrix failed at that commit); the regenerated golden folds those in.

Gates: `make qualify` clean (one `tests/releasepolicy` deadline flake under parallel load; passes in isolation and on the clean rerun), `make test-coverage` 80.0% vs 75% floor. Coverage deltas vs pre-round HEAD: `pkg/constraints 87.6% → 97.9%`, `pkg/cli 72.3% → 73.1%`, `pkg/recipe 88.1% → 88.2%`, `pkg/client/v1` unchanged (doc-only).

njhensley (approved):
Re-reviewed at `760cf2a3` — all 7 findings from the multi-persona round are addressed, and every new test executes and passes locally (`pkg/constraints`, `pkg/recipe`, `pkg/cli` green).

- 🟠 **touched-map invariant** — `touched_invariant_test.go` pins production (exact-set table tests over all 5 dimensions) *and* end-to-end (stated `--os` propagates the coverage error; derived os relaxes and drops from output). A dropped/mis-keyed mark now fails four tests.
- 🟡 **fail-closed on unparseable value** — kept fail-closed (correct); fixed the real sub-defect via `PropagateOrWrap` in `evaluate.go` so `ErrCodeInvalidRequest` reaches the engine un-clobbered, pinned by `TestEvaluate`.
- 🟡 **misleading "no recipe provides"** — `excludedOverlayProvides` + `onlyExcluded` wording, with a subtest asserting the old wording is gone.
- 🟡 **OpenAPI overpromise** — all four HTTP 400 descriptions + api-reference now scope `excludedOverlays`/`constraintWarnings` to snapshot-only.
- 🟡 **SDK parity** — documented the STRICT post-condition on `ResolveRecipeFromSnapshot` (right call; the facade has no touched-provenance).
- 🔵 **golden completion content** — `ValidCompletions` added to the classification and diffed; golden regenerated (also folded in the missing `h100-aks-ubuntu-training-slurm` projections).
- 🔵 **multi-element wording** — verbatim multi-tuple and sorted multi-value assertions added; deprioritizing excluded completions declined with sound reasoning.

The two declined items are defensible design decisions, not gaps. Approving.

PR #1790, opened 2026-07-16
feat(ci): consolidate KWOK tier workflows into single reusable runner

### LINKED ISSUE #1172: Consolidate KWOK tier workflows: single reusable test workflow + one deployer-list source

mohityadav8:
## Summary

* Replace `kwok-tier3-shard.yaml` with `kwok-test-run.yaml`; the `checkout/load-versions/kwok-test` step block now lives in exactly one place.
* Define readonly `DEPLOYERS` once in `discover`; Tier 1, Tier 2, and Tier 3 all derive from it, removing hardcoded matrix axes and sync comments.
* Emit `tier1_pairs` and `tier2_pairs` from `discover`; all three tiers call `kwok-test-run.yaml` with pre-built `{recipe, deployer}` pairs.
* Keep Tier 2 deployer coverage Helm-only (deliberate; full coverage remains in Tier 3).
* Fix `workflow_dispatch` early-exit to also emit `tier1_pairs` and `tier2_pairs`.
* Add a Tier 1 pair-count guard (warning at >200).
* Update workflow path triggers and the ADR-003 workflow diagram.

Closes #1172


.github/workflows/kwok-recipes.yaml
```diff
@@ -80,13 +80,16 @@ jobs:
...
     outputs:
-      tier1: ${{ steps.classify.outputs.tier1 }}
-      tier2: ${{ steps.classify.outputs.tier2 }}
-      tier3: ${{ steps.classify.outputs.tier3 }}
+      tier1_pairs: ${{ steps.classify.outputs.tier1_pairs }}
+      tier2_pairs: ${{ steps.classify.outputs.tier2_pairs }}
       tier3_batches: ${{ steps.classify.outputs.tier3_batches }}
     steps:
       - name: Checkout
+<<<<<<< HEAD
```
njhensley: **🔴 Blocker — Unresolved git merge-conflict markers committed into the workflow**

The `discover` job's Checkout step contains literal conflict markers at lines 88/90/92 (`<<<<<<< HEAD` / `=======` / `>>>>>>> ff27aee1 ...`), left over from a rebase/merge onto newer `main` (which bumped actions/checkout to v7.0.1). This is invalid YAML — GitHub cannot parse the workflow, so `discover` and every tier fail to launch, and yamllint/actionlint error immediately.

**Blast radius:** Total CI break — nothing in this PR or the KWOK gate can run until it is resolved.

**Fix:** Resolve the conflict: keep the HEAD side (`actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1`) and delete the `<<<<<<<`/`=======`/`>>>>>>>` lines and the incoming v7.0.0 block. That matches the reusable `kwok-test-run.yaml`, which already pins v7.0.1 (line 51). Then run `grep -rnE '^(<<<<<<<|=======|>>>>>>>)' .github/` to confirm no other marker slipped in.

.github/workflows/kwok-recipes.yaml
```diff
@@ -441,3 +461,4 @@ jobs:
           fi
 
           echo "All recipe validations passed" >> $GITHUB_STEP_SUMMARY
+
```
njhensley: **🔴 Blocker — Trailing-whitespace line at EOF fails yamllint (trailing-spaces: enable)**

The earlier 'missing trailing newline' nit was 'fixed' by appending a line of 10 spaces after the final `echo "All recipe validations passed"` (confirmed via `cat -A`: line 464 is `          $`). `.yamllint.yaml` sets `trailing-spaces: enable`, and `.github/**` is not in the ignore list.

**Blast radius:** Breaks the lint gate in `make lint` / `make qualify` / CI, independently of the conflict markers.

**Fix:** Replace the whitespace-only line with a single clean newline: the file should end with the `echo ...` line followed by exactly one `\n`, no blank/space-only trailing line.

.github/workflows/kwok-recipes.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+﻿# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
njhensley: **🟡 Minor — UTF-8 BOM (U+FEFF) still on line 1 of both workflow files (not fixed)**

Both `kwok-recipes.yaml` and `kwok-test-run.yaml` still begin with bytes `ef bb bf` before `# Copyright` (re-confirmed at this head). It does not fail CI (no yamllint BOM rule; addlicense still matches the header), but it is an unintended editor artifact in the files this PR is meant to tidy.

**Blast radius:** Hygiene/consistency; can trip stricter YAML/actionlint/grep tooling later. Also affects .github/workflows/kwok-test-run.yaml line 1.

**Fix:** sed -i '1s/^\xEF\xBB\xBF//' .github/workflows/kwok-recipes.yaml .github/workflows/kwok-test-run.yaml (or re-save both as UTF-8 without BOM).

.github/workflows/kwok-recipes.yaml
```diff
@@ -105,11 +108,10 @@ jobs:
 
           # --- workflow_dispatch: test exactly the requested recipe ---
           if [[ -n "${DISPATCH_RECIPE}" ]]; then
-            single=$(jq -nc '[$r]' --arg r "${DISPATCH_RECIPE}")
-            echo "tier1=${single}"   >> "$GITHUB_OUTPUT"
-            echo "tier2=[]"          >> "$GITHUB_OUTPUT"
-            echo "tier3=[]"          >> "$GITHUB_OUTPUT"
-            echo "tier3_batches=[]"  >> "$GITHUB_OUTPUT"
+            single=$(jq -cn --arg r "${DISPATCH_RECIPE}" '[{recipe: $r, deployer: "helm"}]')
+            echo "tier1_pairs=${single}" >> "$GITHUB_OUTPUT"
```
njhensley, line 117: **🟡 Minor — workflow_dispatch runs helm-only (6 deployers → 1)**

The manual-dispatch path now emits `[{recipe:$r, deployer:"helm"}]`, where the old path emitted `tier1=[$r]` which the (now-removed) test-tier1 matrix crossed against all 6 deployers. So a manual dispatch now runs exactly one helm leg.

**Blast radius:** Bounded to the manual debug/repro escape hatch — the automated PR gate, push-to-main, and nightly Tier 3 still exercise all 6 deployers. But the reduction is silent, with no signal the other 5 legs were skipped.

**Fix:** Cross the dispatched recipe with the shared `deployers` array (same jq recipe×deployer expansion as tier1_pairs), or make the reduced coverage an explicit, documented choice.

.github/workflows/kwok-recipes.yaml
```diff
@@ -271,20 +287,24 @@ jobs:
...
-          echo "tier1=${tier1}" >> "$GITHUB_OUTPUT"
-          echo "tier2=${tier2}" >> "$GITHUB_OUTPUT"
-          echo "tier3=${tier3}" >> "$GITHUB_OUTPUT"
+          echo "tier1_pairs=${tier1_pairs}"     >> "$GITHUB_OUTPUT"
+          echo "tier2_pairs=${tier2_pairs}"     >> "$GITHUB_OUTPUT"
           echo "tier3_batches=${tier3_batches}" >> "$GITHUB_OUTPUT"
 
+                   tier1_count=$(echo "$tier1_pairs" | jq 'length')
+          if (( tier1_count > 200 )); then
+            echo "::warning::Tier 1 contains ${tier1_count} pairs and is approaching GitHub's 256-configuration matrix limit"
```
njhensley, line 293: **🟡 Minor — Tier 1 has no fail-closed 256-cap guard, only a >200 warning**

tier1_pairs (generic overlays × 6 deployers) is passed to kwok-test-run.yaml as a single un-batched matrix; discover only emits a `::warning::` at >200. Tier 3, by contrast, batches and hard-fails (exit 1) if the size is ever raised past 256 — precisely to avoid GitHub's opaque 'produced N configurations which exceeds the maximum of 256' rejection.

**Blast radius:** ~120 pairs today (20 generic overlays × 6), with headroom to ~43 overlays, so distant — but the architecture is now inconsistent: one tier fails closed with a legible message, the other only warns and would fail opaquely.

**Fix:** Upgrade the >200 warning to a fail-closed `::error::` + `exit 1` at >256 (naming the cause), or batch tier1_pairs through the same fan-out as Tier 3.

.github/workflows/kwok-recipes.yaml
```diff
@@ -271,20 +287,24 @@ jobs:
...
 
           # --- Output ---
-          echo "tier1=${tier1}" >> "$GITHUB_OUTPUT"
-          echo "tier2=${tier2}" >> "$GITHUB_OUTPUT"
-          echo "tier3=${tier3}" >> "$GITHUB_OUTPUT"
+          echo "tier1_pairs=${tier1_pairs}"     >> "$GITHUB_OUTPUT"
+          echo "tier2_pairs=${tier2_pairs}"     >> "$GITHUB_OUTPUT"
           echo "tier3_batches=${tier3_batches}" >> "$GITHUB_OUTPUT"
 
+                   tier1_count=$(echo "$tier1_pairs" | jq 'length')
```
njhensley: **🔵 Nitpick — Stray ~19-space over-indentation on the tier1_count assignment**

`tier1_count=$(echo "$tier1_pairs" | jq 'length')` is indented at ~19 leading spaces while the surrounding `run:` block is at 10. Harmless (bash ignores leading whitespace inside the block scalar) but reads as broken alignment.

**Blast radius:** Cosmetic; no runtime effect.

**Fix:** Re-indent the line to the surrounding 10-space level.

njhensley (changes requested):
## 📋 Multi-persona review — Request changes ⛔

**Method:** 4 persona passes (Operability/CI-DX · Correctness · Security · Domain & Architecture), each finding independently confirmed or refuted by a senior meta-reviewer against the resolved code at head `8fe16c5d`. Line links/anchors pinned to that SHA.

### Overall assessment

The design of this consolidation is sound, but **the latest push regressed the PR into a non-mergeable state.** A rebase/merge onto newer `main` (which bumped `actions/checkout` to v7.0.1) was left with **unresolved conflict markers committed into `kwok-recipes.yaml`** (invalid YAML — the workflow won't parse), and the attempt to fix an earlier missing-newline nit introduced a **trailing-whitespace line that fails `yamllint`**. Either 🔴 breaks CI on its own. The substantive design findings (dispatch helm-only, Tier 1 cap) are unchanged from the prior pass and remain open. Please push a clean fix, then this is ready for a merge-focused re-review.

### ✅ Confirmed non-issues (re-checked at this head)

- **Reusable-workflow checkout pin is correct** — `kwok-test-run.yaml` uses `v7.0.1` (`3d3c42e5…`), consistent with the `HEAD` side of the conflict, so resolving to `HEAD` needs no further pin change.
- **Supply chain clean** — `contents: read` on caller and reusable workflow; no `secrets: inherit`; `pull_request` (not `_target`); `persist-credentials: false`. Fork PRs gain nothing.
- **Tier 3 concurrency intact** (SHA + batch-id group lives on the unchanged `test-tier3` caller job); **Tier 2 helm-only is behavior-preserving** (action defaults to `helm`); **`KWOK Test Summary` still gates correctly** through the reusable-workflow caller jobs; **`sync-budget_test.sh`** update is complete and correct (11→9 checks, `jtm=20,tm=20`); the rename repoints every reference.

### Additional low-priority notes (not inlined)

- 🔵 **Deployer "single source of truth" slightly overstated** — the functional list is now single, but `.github/actions/kwok-test/action.yml`'s `deployer` input *description* still enumerates all 6 (doc-string only), and the old "keep in sync with action.yml" reminder comment was dropped. Consider re-adding a one-line pointer.
- 🔵 **`Tier N:` prefix dropped from check labels** — caller jobs lost their `name:`, so checks render terser (e.g. `test-tier1 / <recipe> (<deployer>)`). No gating impact (only `KWOK Test Summary` is required); optional `name: 'Tier 1'`/`'Tier 2'`/`'Tier 3'` on the callers restores the label.
- 🔵 **PR-gate per-leg timeout raised 18→20** with `sync-budget_test.sh` updated to match, but no stated rationale. Benign; a one-line note would help.

### Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 2 | committed merge-conflict markers (invalid YAML) · EOF trailing-whitespace line (yamllint fail) |
| 🟠 Major | 0 | — |
| 🟡 Minor | 3 | BOM ×2 files · dispatch helm-only · Tier 1 no fail-closed cap |
| 🔵 Nitpick | 3 | tier1_count indentation · deployer-SoT pointer · check-label / timeout notes |

**Recommendation:** Must fix before further review — (1) resolve the conflict markers keeping `v7.0.1`, (2) remove the trailing-whitespace EOF line. While there, strip the BOM and re-indent `tier1_count`, and run `make lint` locally (CI can't parse the file until the markers are gone). The dispatch-scope and Tier 1 cap findings can be fixed or deferred in writing.

<sub>Reviewed with a multi-persona + adversarial meta-review workflow.</sub>

.github/workflows/kwok-recipes.yaml
```diff
@@ -240,16 +247,16 @@ jobs:
           # --- Tier 3: full matrix (all testable overlays) ---
           tier3="$all"
 
+          tier1_pairs=$(jq -cn --argjson recipes "$tier1" --argjson deployers "$deployers" '
```
njhensley, line 250: **🔴 Blocker — tier1_pairs jq assignment is unterminated and tier2_pairs is never defined — classify script fails bash -n**

Line 250 `tier1_pairs=$(jq -cn --argjson recipes "$tier1" --argjson deployers "$deployers" '` opens a `$(` command substitution AND a single-quoted jq program, but the jq body, the closing `')`, and the entire `tier2_pairs=$(…)` block that used to follow were deleted in this push. The next `'` in the file is 24 lines down on the `tier3_batches` jq call, so the whole Tier 3 batching section gets swallowed into tier1_pairs' string and the `$(` never finds its matching `)`. Extracting the classify step and running `bash -n` yields exit 2: `unexpected EOF while looking for matching )'`. Separately, `tier2_pairs` is referenced at lines 282 and 289 but never assigned in the classify path — an unbound-variable error under `set -u`.

**Blast radius:** Total CI break — `discover` fails on every trigger (PR, push, schedule, dispatch), so no tier runs and `KWOK Test Summary` cannot pass.

**Fix:** Restore the two clobbered assignments before the `# --- Tier 3 batching ---` comment:
```bash
          tier1_pairs=$(jq -cn --argjson recipes "$tier1" --argjson deployers "$deployers" '
            [ $recipes[] as $r | $deployers[] as $d | {recipe: $r, deployer: $d} ]
          ')

          tier2_pairs=$(jq -cn --argjson recipes "$tier2" '
            [ $recipes[] as $r | {recipe: $r, deployer: "helm"} ]
          ')
```
Then extract the classify `run:` and `bash -n` it (or run actionlint) before pushing.

.github/workflows/kwok-test-run.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+﻿# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
njhensley: **🟡 Minor — UTF-8 BOM (U+FEFF) still on line 1 of both workflow files (carried over, not fixed)**

Both kwok-recipes.yaml and kwok-test-run.yaml still begin with bytes `ef bb bf` before `# Copyright` (re-confirmed at head 32c6878f). Unintended editor artifact, flagged in the two prior review rounds.

**Blast radius:** Does not fail CI (no yamllint BOM rule; addlicense still matches), but it is invisible noise in the files this PR is meant to tidy. Also affects .github/workflows/kwok-recipes.yaml line 1.

**Fix:** sed -i '1s/^\xEF\xBB\xBF//' .github/workflows/kwok-recipes.yaml .github/workflows/kwok-test-run.yaml (or re-save both as UTF-8 without BOM).

njhensley (changes requested):
## 📋 Re-review (round 3) — Request changes ⛔

Verified against the new push (head `32c6878f`). **The two prior 🔴 blockers are fixed, but this edit introduced a new 🔴 blocker that breaks the discover job outright.**

### Overall assessment

Good progress — the author cleared everything from the last round except the BOM. But the conflict-resolution edit **mangled the `tier1_pairs` / `tier2_pairs` construction**: `tier1_pairs=$(jq …'` is left unterminated and `tier2_pairs` is never assigned. I extracted the classify `run:` block and ran `bash -n` → **exit 2, `unexpected EOF while looking for matching )'`**. So the `discover` job fails to parse on every trigger and no tier runs. Net: still not mergeable, now for a higher-impact reason than before.

### ✅ Fixed since last review (verified at this head)

- **Merge-conflict markers** — gone; discover Checkout cleanly pins `actions/checkout@3d3c42e5… # v7.0.1`, consistent with `kwok-test-run.yaml`.
- **EOF trailing-whitespace line** — removed (no space-only trailing line; the missing final newline is not linted).
- **`workflow_dispatch` helm-only regression** — fixed; dispatch now crosses the recipe with the full `deployers` set.
- **Tier 1 fail-closed cap** — added (`::error::` + `exit 1` when `tier1_count > 256`).
- **`tier1_count` indentation** — corrected to the block's 10-space level.
- **`Tier N` check labels** — restored (`name: 'Tier 1'` / `'Tier 2'` / `'Tier 3'` on the caller jobs).
- **Timeout 18→20 rationale** — documented inline (flux-oci readiness ~7min).
- **Deployer single-source-of-truth** — `deployers` defined once before dispatch, with a comment pointing at `action.yml`'s doc-string as the un-enforced prose copy.

### Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 1 | broken `tier1_pairs` / missing `tier2_pairs` — classify script fails `bash -n` |
| 🟠 Major | 0 | — |
| 🟡 Minor | 1 | BOM ×2 files |
| 🔵 Nitpick | 0 | — |

**Recommendation:** One fix stands between this and mergeable — restore the `tier1_pairs`/`tier2_pairs` block and `bash -n` the classify step (or run actionlint) before pushing. Strip the BOM while you're in there. Everything else from the prior two rounds is resolved.

<sub>Reviewed with a multi-persona + adversarial meta-review workflow; the blocker was confirmed by extracting the classify step and running `bash -n`.</sub>

njhensley (approved):

PR #1758, opened 2026-07-15
fix: enforce closed-world bundle verification

mchmarny:
## Summary

Enforces a closed-world inventory for generated deployment bundles and carries a private, revalidated snapshot through verification, ZIP, and OCI publication. Hardens generic and Helm OCI publication, CLI image-reference output, cleanup/cancellation behavior, and the public trust contract.

## Motivation / Context

Listed-digest verification alone cannot establish that a bundle contains exactly the files AICR generated. This change makes exact inventory validation the common boundary for generation, trust evaluation, archive streaming, and registry publication; incomplete legacy bundles now fail closed and must be regenerated.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [ ] Refactoring (no functional changes)
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] CLI (`cmd/aicr`, `pkg/cli`)
- [x] API server (`cmd/aicrd`, `pkg/server`)
- [ ] Recipe engine / data (`pkg/recipe`)
- [x] Bundlers (`pkg/bundler`, `pkg/component/*`)
- [ ] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`)
- [ ] Validator (`pkg/validator`)
- [ ] Core libraries (`pkg/errors`, `pkg/k8s`)
- [x] Docs/examples (`docs/`, `examples/`)
- [x] Other: OCI and supply-chain publication (`pkg/oci`, `pkg/evidence/attestation`)

## Implementation Notes

- Parses and generates deterministic typed checksum manifests, validates the exact regular-file and directory inventory, and permits only the three explicit verification metadata paths.
- Rejects Unicode simple-fold aliases across payload, checksum-name, and reserved-metadata comparisons while retaining original path bytes for access and diagnostics.
- Finalizes every deployer output through the same inventory boundary and reports incomplete or unmanaged legacy bundles as `unknown` trust.
- Retains the first private verified snapshot as the immutable OCI source after caller-visible output publication.
- Replaces existing output through an anchored two-phase commit with pre-commit rollback, identity-aware cleanup, and commit-aware cancellation semantics.
- Anchors checksum publication and rollback to retained roots; opens inventory and payload files descriptor-first with bounded, nonblocking, no-follow reads.
[...]

PR #1794, opened 2026-07-16
fix(oci): restore layer title annotation for file-store unpack

### LINKED ISSUE #1793: UAT evidence verify fails on all main cells (all clouds) since #1758
### Summary

Since **#1758 ("enforce closed-world bundle verification", merged 2026-07-15)**, the UAT `verify` step fails on **every `@ main` cell** — AWS, GCP, and Azure, training *and* inference. `provision → install → validate → conformance → performance → sign → OCI push` all succeed; only the evidence-bundle **round-trip verification** fails. Released `@ v0.17.0` cells (which run the released binary, pre-#1758) pass. So `main`'s nightly evidence coverage is fully red.

### Evidence

| Version | aws | gcp | azure |
|---|---|---|---|
| `@ main` | ❌ | ❌ | ❌ |
| `@ v0.17.0` | ✅ | ✅ | ✅ |

Failing `@ main` runs (all same failure): `29515344959` (azure/inference), `29476365480` (aws/training), `29475780757` (gcp), `29472776223` (aws), `29472765075` (azure), `29448415427` (azure). Passing `@ v0.17.0`: `29486726163` (aws), `29485955338` (gcp), `29476494205` (azure).

### Failure detail

`verify` dies at the **`materialize-bundle`** step (`evidence-result.json`):

```json
{ "step": 1, "name": "materialize-bundle", "status": "failed",
  "detail": "[INVALID_REQUEST] pulled artifact does not contain a recognizable summary bundle" }
```

The bundle builds, signs (`rekorLogIndex` present), and pushes to GHCR fine — but when `verify` pulls the OCI artifact back and extracts it, the verifier can't find the `recipe.yaml` + `manifest.json` markers (`resolveBundleDir` → `HasBundleMarkers`). Because the **same `main` binary both packs and verifies within a run**, this is an internal pack↔verify inconsistency in #1758, not a cross-version skew.

### Root-cause narrowing

- The **only** evidence/OCI-path change between `v0.17.0` and `main` is #1758 (`git log v0.17.0..main -- pkg/evidence/ pkg/oci/` → one commit).
- The **pull/verify** side is *untouched* by #1758 — `pkg/evidence/verifier/fetch.go` (`materializeBundle`, `resolveBundleDir`) and `pkg/evidence/attestation/publish.go` (`HasBundleMarkers`) are not in the diff. So the regression is on the **pack** side.
- `builder.go` still writes `manifest.json`/`recipe.yaml` to the summary-bundle root, and the generic (evidence) pack path sets **no archive prefix** (`archiveOptions.Prefix` is only set for Helm, `helm.go:459`) — so in theory markers should land at the layer root.
- The behavioral change is the packaging **staging**: #1758 replaced `preparePushDir(SourceDir, SubDir)` with the closed-world `prepareSource(SourceDir, OutputDir, SubDir, SourceFiles)` workspace in `pkg/oci/{push.go,source_files.go}`. The archive entry names come directly from `preparedSource.relativeFiles()` (`pkg/oci/archive.go:187`), so the staged file enumeration is what now determines the layer layout / contents. **The exact line is most likely in the new `prepareSource`/`relativeFiles` staging** — a local pack→pull round-trip repro would pin it definitively.

### Impact

- Every `main` nightly cell (all clouds, both intents) has been red since 2026-07-15 — no `main` evidence signal.
- Blocks any green end-to-end `main` acceptance run — e.g. the manual Azure inference acceptance gate on PR #1776 (that PR itself is unrelated; it touches no evidence code).

### Reproduction

Any `@ main` UAT run (`gh workflow run uat-run.yaml -f reservation=aws-h100`) fails at `UAT - verify`. `@ v0.17.0` passes.

njhensley:
## Summary

Restore the `org.opencontainers.image.title` annotation on the **generic** OCI archive layer, so an ORAS file store can materialize the layer to disk on pull. Fixes the evidence-verify regression that broke **every `@ main` UAT cell** (all clouds, both intents).

## Motivation / Context

Fixes: #1793

Since #1758 (generic-path packaging rewrite), UAT `verify` fails on every `@ main` cell — AWS/GCP/Azure, training + inference — at the `materialize-bundle` step:

```
[INVALID_REQUEST] pulled artifact does not contain a recognizable summary bundle
```

`@ v0.17.0` cells pass. I pulled both artifacts from GHCR and diffed them — **the bundle content is identical and correct** (`recipe.yaml` + `manifest.json` at the tar root in both). The only difference is one layer annotation:

| layer annotation | v0.17.0 ✅ | `main` ❌ |
|---|---|---|
| `io.deis.oras.content.unpack` | `"true"` | `"true"` |
| **`org.opencontainers.image.title`** | **`"."`** | **absent** |

The verifier materializes bundles with an ORAS **file store** (`oras.Copy(…, file.New(tmp), …)`, `pkg/evidence/verifier/fetch.go`). ORAS writes/unpacks each layer to disk using `org.opencontainers.image.title` as the destination path. With `title:"."` + `unpack:"true"` it unpacks the tar into the store root (markers land at root); with the title **absent** the layer is an unnamed blob and is **never written to disk** → empty dir → `resolveBundleDir` fails.

#1758's generic-path rewrite dropped the annotation (`pkg/oci/archive.go`); the Helm path still sets its own title (`pkg/oci/helm.go`), which is why only the generic (evidence/bundle) path regressed. The removed doc comment in #1758 notes the title annotation "materializing as a literal filename" — but the verifier's file-store pull *depends* on it, so it must stay.

## Type of Change

## Component(s) Affected

- [x] Core libraries (`pkg/oci`)
- [x] Other: evidence verification (consumer)

## Implementation Notes

One line: re-add `ociv1.AnnotationTitle: "."` to the generic layer descriptor in `pkg/oci/archive.go`, restoring v0.17.0 behavior. Plus a regression test (`TestBuildDeterministicTarGzipSetsFileStoreUnpackTitle`) asserting both the `unpack` and `title` annotations, verified to fail without the fix (`title=""`) and pass with it.

@mchmarny — flagging you as the #1758 author in case dropping the title was deliberate (the removed comment suggests a stray-`.`-file concern); if so, the alternative would be teaching the verifier's materialize step to extract the layer without relying on the ORAS file-store title convention. Restoring the annotation is the minimal fix that matches the last-known-good behavior.

## Testing

```bash
go test -race ./pkg/oci/... ./pkg/evidence/... ./pkg/bundler/attestation/...   # pass
golangci-lint run -c .golangci.yaml ./pkg/oci/...                              # 0 issues
```
(Note: `TestHelmV4_2_3ExplicitVersionPull` fails identically on clean `origin/main` — a local helm-binary version mismatch, unrelated to this change.)

## Risk Assessment

- [x] **Low** — One additive annotation restoring v0.17.0 behavior; generic-path artifacts now materialize on file-store pull as they did before #1758. Helm path untouched. Reversible.

## Checklist

### DIFF
--- a/pkg/oci/archive.go
+++ b/pkg/oci/archive.go
@@ -161,6 +161,17 @@ func buildDeterministicTarGzipWithDependencies(
 		Annotations: map[string]string{
 			file.AnnotationDigest: tarDigester.Digest().String(),
 			file.AnnotationUnpack: "true",
+			// AnnotationTitle names the on-disk destination an ORAS file
+			// store (oras.Copy into file.New(dir)) unpacks this layer to.
+			// The generic archive packs the source tree at the tar root,
+			// so the store must unpack it into its own root (".") — without
+			// this the store treats the layer as an unnamed blob and writes
+			// NOTHING to disk, and a downstream file-store pull (e.g. the
+			// evidence verifier's materialize step) sees an empty directory
+			// and fails "does not contain a recognizable summary bundle".
+			// The Helm path sets its own title (the chart dir); only this
+			// generic path regressed when the annotation was dropped.
+			ociv1.AnnotationTitle: ".",
 		},
 	}
 	if err := verifyArchiveDescriptor(ctx, layout.child, name, descriptor, deps); err != nil {
--- a/pkg/oci/archive_test.go
+++ b/pkg/oci/archive_test.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/opencontainers/go-digest"
 	ociv1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"oras.land/oras-go/v2/content/file"
 
 	apperrors "github.com/NVIDIA/aicr/pkg/errors"
 )
@@ -85,6 +86,43 @@ func TestPackage_SourceFilesArchiveContainsOnlyFrozenSortedSelection(t *testing.
 	}
 	if _, ok := files["other.txt"]; ok {
 		t.Fatal("archive included an unselected file")
+	}
+}
+
+// TestBuildDeterministicTarGzipSetsFileStoreUnpackTitle locks the two layer
+// annotations an ORAS file store needs to MATERIALIZE the layer to disk on
+// pull. oras.Copy into a file.New(dir) store unpacks a layer into the path
+// named by org.opencontainers.image.title (here ".", the store root) when
+// io.deis.oras.content.unpack is "true"; WITHOUT the title annotation the
+// store treats the layer as an unnamed blob and writes NOTHING to disk, so a
+// downstream file-store pull (e.g. the evidence verifier's materialize step)
+// sees an empty directory. Regression guard for the generic archive path —
+// dropping the title annotation broke every `main` UAT verify (issue #1793).
+func TestBuildDeterministicTarGzipSetsFileStoreUnpackTitle(t *testing.T) {
+	sourceDir := t.TempDir()
+	outputDir := t.TempDir()
+	writeSourceFixture(t, sourceDir)
+	safeOCITempRoot(t)
+	prepared, err := preparePackageSource(context.Background(), sourceDir, outputDir, "", nil)
+	if err != nil {
+		t.Fatalf("preparePackageSource() error = %v", err)
+	}
+	t.Cleanup(func() {
+		if closeErr := prepared.Close(); closeErr != nil {
+			t.Error(closeErr)
+		}
+	})
+	layout := newTestOwnedLayout(t, outputDir)
+	_, desc, err := buildDeterministicTarGzip(context.Background(), prepared, layout, archiveOptions{})
+	if err != nil {
+		t.Fatalf("buildDeterministicTarGzip() error = %v", err)
+	}
+	if got := desc.Annotations[ociv1.AnnotationTitle]; got != "." {
+		t.Errorf("layer %q annotation = %q, want %q — an ORAS file store unpacks into this path; "+
+			"empty means the layer is never written to disk on pull", ociv1.AnnotationTitle, got, ".")
+	}
+	if got := desc.Annotations[file.AnnotationUnpack]; got != "true" {
+		t.Errorf("layer %q annotation = %q, want \"true\"", file.AnnotationUnpack, got)
 	}
 }
 


pkg/oci/archive.go
```diff
@@ -161,6 +161,17 @@ func buildDeterministicTarGzipWithDependencies(
...
+			// store (oras.Copy into file.New(dir)) unpacks this layer to.
+			// The generic archive packs the source tree at the tar root,
+			// so the store must unpack it into its own root (".") — without
+			// this the store treats the layer as an unnamed blob and writes
+			// NOTHING to disk, and a downstream file-store pull (e.g. the
+			// evidence verifier's materialize step) sees an empty directory
+			// and fails "does not contain a recognizable summary bundle".
+			// The Helm path sets its own title (the chart dir); only this
+			// generic path regressed when the annotation was dropped.
+			ociv1.AnnotationTitle: ".",
```
mchmarny, line 174: To close the question in the PR description: dropping this in #1758 wasn't a deliberate pull-side decision. The removed comment's stray-file concern was about the push-side file store — the old path pushed through `file.New` + `fs.Add(ctx, ".", ...)`, which could materialize the title as a literal file in the store root, hence the temp-dir staging it documented. The #1758 rewrite builds the tar directly into the OCI layout with no push-side file store, so that concern is moot and losing the layer title (and with it the verifier's materialization) was collateral. Restoring it on the layer is the right call; push-side it's inert now.

pkg/oci/archive_test.go
```diff
@@ -88,6 +89,43 @@ func TestPackage_SourceFilesArchiveContainsOnlyFrozenSortedSelection(t *testing.
...
+			t.Error(closeErr)
+		}
+	})
+	layout := newTestOwnedLayout(t, outputDir)
+	_, desc, err := buildDeterministicTarGzip(context.Background(), prepared, layout, archiveOptions{})
+	if err != nil {
+		t.Fatalf("buildDeterministicTarGzip() error = %v", err)
+	}
+	if got := desc.Annotations[ociv1.AnnotationTitle]; got != "." {
+		t.Errorf("layer %q annotation = %q, want %q — an ORAS file store unpacks into this path; "+
```
mchmarny, line 121: nit: this pins the annotations, but the actual contract is ORAS's materialization behavior — a future oras-go semantic change would keep this test green while UAT breaks again. A push→pull round-trip test (Package → `oras.Copy` into `file.New(t.TempDir())` → assert `recipe.yaml`/`manifest.json` exist on disk) would have caught #1793 at PR time and survives annotation-convention changes. Fine as a follow-up PR.

mchmarny (approved):
Root cause verified end to end: the ORAS file store routes unnamed layers to fallback storage (never to disk), v0.17.0 set the title via `fs.Add(ctx, ".", ...)`, and the verifier's file-store pull depends on it. I confirmed the new test fails without the one-line fix, and that the Helm path is unaffected (`buildHelmChartTGZ` nils layer annotations before packing). The #1758 drop was incidental, not a deliberate pull-side decision — details inline. One non-blocking suggestion on test depth; nothing blocks merge.

PR #1805, opened 2026-07-17
feat(validator): recipe-supplied NCCL benchmark runtime via --data

### LINKED ISSUE #1792: nccl-benchmark-profile escape hatch doesn't cover external-only service+accelerator pairs
## Background

#1703 reported that NCCL validation applicability is compiled-in, blocking \`aicr validate\` for recipes introduced via \`--data\` with service/accelerator combinations not in the embedded catalog.

PR #1719 (shipped in v0.17.0) added a \`nccl-benchmark-profile\` constraint as an opt-in escape hatch. The intent was to let external recipes reference an existing compiled benchmark variant explicitly.

## The remaining gap

The escape hatch still requires the target \`service+accelerator\` pair to be present in the compiled benchmark matrix. From the PR notes:

> profiles are limited to pairs already in the compiled matrix

We have a recipe for a private service introduced entirely via \`--data\` — the service type is intentionally not in the embedded catalog. Even after bumping to v0.17.0 and adding \`nccl-benchmark-profile\`, \`aicr validate\` fails because the pair has no compiled entry to point at. The profile key is a no-op in this case.

## What we have

The recipe's \`validation\` block declares NCCL performance checks and thresholds. We've validated externally that it works correctly — the issue is solely that \`aicr validate\` can't gate it in CI because the compiled matrix doesn't cover private services.

## Requested fix

Complete the data-driven validation vision from #1703: allow a recipe's \`validation.performance.checks\` and thresholds to drive \`aicr validate\` directly, without requiring a compiled applicability entry. The \`nccl-benchmark-profile\` approach is a step in the right direction but still requires the binary to know about the service — which defeats the purpose for private \`--data\`-only recipes.

Relates to #1703, #1719.

njhensley:
## Summary

Adds an `nccl-benchmark-runtime-ref` performance constraint so a recipe can ship its own Kubeflow `TrainingRuntime` as a `--data` file and have `aicr validate` run the NCCL benchmark against it — keyed on the recipe's own criteria, with no compiled applicability entry required.

## Motivation / Context

The NCCL check applicability is a compiled `service+accelerator` matrix (`supportedNCCLCombinations`), and each supported pair is backed by a runtime template baked into the validator image at `validators/performance/testdata/{accelerator}/{service}/`. The `nccl-benchmark-profile` escape hatch (#1719, v0.17.0) lets an external `--data` recipe *borrow* an embedded template, but only when one exists whose fabric matches the cluster. A **private service introduced entirely via `--data`** — a `criteria.service`/`accelerator` the validator ships no template for, with a fabric that reuses none of the embedded ones — has no compiled pair to point a profile at, so its NCCL checks are silently skipped and it cannot be performance-certified in CI.

This completes the data-driven validation vision: the recipe **supplies the benchmark itself**. It ships a `TrainingRuntime` in its `--data` tree at the same `validators/performance/testdata/{accelerator}/{service}/runtime.yaml` layout the embedded templates use, and references it with a bare `{accelerator}/{service}` value. Because the file already sits at the embedded path, upstreaming it later is a copy (move the file in, add the matrix tuple, drop the ref).

Fixes: #1792
Related: #1703, #1719

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `validators/performance` (in-pod check binary), `recipes/validators/catalog.yaml` (comment)

## Implementation Notes

Two-stage design, splitting authoring from execution across the existing orchestrator/pod boundary:

- **Orchestrator** (`pkg/validator/benchmark_runtime_ref.go`, `resolveBenchmarkRuntimeRef`, called at the top of `ValidatePhases`/`ValidatePhase`): parses the `{accelerator}/{service}` ref, reads `validators/performance/testdata/{accelerator}/{service}/runtime.yaml` through the recipe `DataProvider`, and **lowers its content into an `nccl-benchmark-runtime` carrier constraint** that rides to the pod via the existing `ValidationInput` ConfigMap. No new delivery plumbing. Resolution **consumes** the ref (idempotent; the raw ref never reaches the pod).
- **Pod** (`validators/performance`): renders the carrier in place of a baked-in template (shared `renderYAMLTemplate` core), **bypasses** the compiled `supportedNCCLCombinations` gate, and **skips every service-specific step** — EFA/TCPXO/RDMA NIC discovery, the GB200-NVreg / GKE-TCPXO preflights, RoCE claim application, NVLS/IMEX provisioning, and transport verification — because the supplied runtime owns its fabric wiring.

The two constraint names are defined once in `pkg/validator/v1` (`PerfConstraint*`) and imported by both ends, so the write side and read side cannot drift.

**Trust / fail-closed:** the runtime is applied only through `trainingRuntimeGVR` with a force-set name/namespace, so a recipe can supply a `TrainingRuntime` and nothing else — not an arbitrary resource kind or namespace. It fails closed (`ErrCodeInvalidRequest`, before any Job is deployed) on a malformed/traversal ref (segments can't contain separators or `..`; the joined path must stay under `.../testdata/`), a missing/empty file, an absent `--data` source, or a ref set alongside an inline carrier. The pod additionally fails closed on a non-`trainer.kubeflow.org/v1alpha1` `TrainingRuntime`, a runtime missing the `node` replicatedJob, or a `-ref` that reaches the validator unresolved (defense-in-depth for the external-controller `Plan()` path). Mutually exclusive with `nccl-benchmark-profile`. A ctx-timeout read propagates as `ErrCodeTimeout`, not `InvalidRequest`.

**Scope note:** resolution is wired into the CLI/facade `aicr validate` path (`ValidatePhases`/`ValidatePhase`). The external-controller `Plan()`/`EnsureDataConfigMaps()` SDK path does not resolve the ref (it has no in-repo callers today); the pod fails closed there rather than silently skipping. Lifting resolution into that shared layer is a small follow-up if a real `Plan()` consumer appears.

## Testing

```bash
# Affected packages (validators/ is excluded from `make test` and the coverage gate,
# so run these directly):
go test -race ./pkg/validator/... ./pkg/validator/v1/ ./validators/...   # PASS
golangci-lint run -c .golangci.yaml ./pkg/validator/... ./validators/... # 0 issues
go vet ./pkg/validator/... ./validators/...                             # clean
gofmt -l <changed .go files>                                            # clean
yamllint recipes/validators/catalog.yaml                                # clean
go build ./...                                                          # clean
```

New tests: orchestrator ref resolution (no-op / valid inject / malformed / traversal / missing / empty / nil-provider / ref+inline conflict / idempotent double-pass / path builder); pod resolve+validate (absent/blank/valid/wrong-kind/wrong-apiVersion/unparseable/missing-node-job/unresolved-ref), the applicability-gate bypass, the custom-runtime cluster path, and a regression guard that a recognized service (`eks`) + custom runtime does not clobber the runtime's own `nodeSelector`.

The change was adversarially reviewed (two multi-agent review passes); every confirmed finding is fixed and regression-tested. CI runs the full `make qualify` gate on this PR.

## Risk Assessment

- [x] **Low** — Additive, opt-in, well-tested, easy to revert

**Rollout notes:** Backwards compatible — recipes without the constraint are byte-for-byte unchanged. Version-skew: an older validator image that predates the shared carrier constant would ignore the constraint and fall back to the criteria-derived skip, so the feature needs a validator new enough to read it.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — affected packages, see Testing
- [x] Linter passes (`make lint`) — golangci-lint 0 issues on changed packages

### DIFF
--- a/pkg/validator/benchmark_runtime_ref.go
+++ b/pkg/validator/benchmark_runtime_ref.go
@@ -56,16 +56,16 @@ func (v *Validator) resolveBenchmarkRuntimeRef(ctx context.Context, vi *v1.Valid
 	// a blank earlier entry must not hide a later non-blank duplicate (and the
 	// overlay merge already collapses same-named constraints, so a duplicate here
 	// is a malformed input). Exactly one of each is allowed.
-	if n := countPerfConstraint(cs, v1.PerfConstraintNCCLBenchmarkRuntimeRef); n > 1 {
+	if n := v1.CountConstraint(cs, v1.PerfConstraintNCCLBenchmarkRuntimeRef); n > 1 {
 		return errors.New(errors.ErrCodeInvalidRequest,
 			fmt.Sprintf("declare at most one %s constraint (found %d)", v1.PerfConstraintNCCLBenchmarkRuntimeRef, n))
 	}
-	if n := countPerfConstraint(cs, v1.PerfConstraintNCCLBenchmarkRuntime); n > 1 {
+	if n := v1.CountConstraint(cs, v1.PerfConstraintNCCLBenchmarkRuntime); n > 1 {
 		return errors.New(errors.ErrCodeInvalidRequest,
 			fmt.Sprintf("declare at most one %s constraint (found %d)", v1.PerfConstraintNCCLBenchmarkRuntime, n))
 	}
 
-	ref, ok := findPerfConstraint(cs, v1.PerfConstraintNCCLBenchmarkRuntimeRef)
+	ref, ok := v1.FindConstraint(cs, v1.PerfConstraintNCCLBenchmarkRuntimeRef)
 	if !ok {
 		return nil
 	}
@@ -74,13 +74,20 @@ func (v *Validator) resolveBenchmarkRuntimeRef(ctx context.Context, vi *v1.Valid
 		return nil
 	}
 
-	// A ref and an inline runtime together are ambiguous — fail closed rather
-	// than silently preferring one.
-	inline, hasInline := findPerfConstraint(vi.Config.Performance.Constraints, v1.PerfConstraintNCCLBenchmarkRuntime)
-	if hasInline && strings.TrimSpace(inline.Value) != "" {
+	// A ref is mutually exclusive with an inline runtime and with a benchmark
+	// profile — supply your own runtime OR borrow an embedded one, never both.
+	// Both conflicts are pure recipe-authoring errors, so reject them here (before
+	// any Job is deployed) rather than leaving the ref+profile case to fail late
+	// in the pod.
+	if inline, ok := v1.FindConstraint(cs, v1.PerfConstraintNCCLBenchmarkRuntime); ok && strings.TrimSpace(inline.Value) != "" {
 		return errors.New(errors.ErrCodeInvalidRequest,
 			fmt.Sprintf("%s and an inline %s are both set — supply one",
 				v1.PerfConstraintNCCLBenchmarkRuntimeRef, v1.PerfConstraintNCCLBenchmarkRuntime))
+	}
+	if profile, ok := v1.FindConstraint(cs, v1.PerfConstraintNCCLBenchmarkProfile); ok && strings.TrimSpace(profile.Value) != "" {
+		return errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("%s and %s are mutually exclusive — supply a runtime or borrow an embedded profile, not both",
+				v1.PerfConstraintNCCLBenchmarkRuntimeRef, v1.PerfConstraintNCCLBenchmarkProfile))
 	}
 
 	relPath, err := benchmarkRuntimeRefPath(raw)
@@ -119,6 +126,16 @@ func (v *Validator) resolveBenchmarkRuntimeRef(ctx context.Context, vi *v1.Valid
 		return errors.New(errors.ErrCodeInvalidRequest,
 			fmt.Sprintf("%s=%q resolved to an empty file (%q)",
 				v1.PerfConstraintNCCLBenchmarkRuntimeRef, raw, relPath))
+	}
+
+	// Shape-check the resolved file against the same contract the pod enforces,
+	// so a runtime that isn't a valid TrainingRuntime fails fast here (offline,
+	// before any Job) instead of only in the pod after the Trainer install. The
+	// check parses the raw pre-substitution content; ${VAR} placeholders live in
+	// value positions and don't affect the identity / replicatedJob shape.
+	if err := v1.ValidateBenchmarkRuntime(string(content)); err != nil {
+		return errors.Wrap(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("%s=%q (%q) resolved to an invalid runtime", v1.PerfConstraintNCCLBenchmarkRuntimeRef, raw, relPath), err)
 	}
 
 	// Consume the ref: rebuild the constraint slice without it (and without any
@@ -177,25 +194,4 @@ func invalidRuntimeRef(ref string) error {
 // non-empty name — no path separators, no "."/".." traversal.
 func isSafeRefSegment(s string) bool {
 	return s != "" && s != "." && s != ".." && !strings.ContainsAny(s, `/\`)
-}
-
-// findPerfConstraint returns the first constraint with the given name.
-func findPerfConstraint(cs []recipe.Constraint, name string) (recipe.Constraint, bool) {
-	for _, c := range cs {
-		if c.Name == name {
-			return c, true
-		}
-	}
-	return recipe.Constraint{}, false
-}
-
-// countPerfConstraint counts constraints with the given name.
-func countPerfConstraint(cs []recipe.Constraint, name string) int {
-	n := 0
-	for _, c := range cs {
-		if c.Name == name {
-			n++
-		}
-	}
-	return n
 }
--- a/pkg/validator/benchmark_runtime_ref.go
+++ b/pkg/validator/benchmark_runtime_ref.go
@@ -1,0 +1,201 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


njhensley:
@coderabbitai review

njhensley:
@coderabbitai On the worker-sizing comment (nccl_all_reduce_bw_constraint.go): I've intentionally scoped sizing to the runtime's `nodeSelector` rather than simulating full scheduler eligibility (nodeAffinity, taint/toleration matching). Reimplementing scheduler placement in a benchmark validator adds significant bug surface, and the uncovered cases already **fail safe** — a runtime that restricts placement below the sized cohort leaves workers Pending and the launcher wait times out with a diagnostic, never a false pass. This is now documented in the code comment at the sizing site and in docs/user/validation.md (with a recommendation to pass `--node-selector` for an exact cohort match). Applied the docs-precision comment about orchestrator-vs-in-check failure staging.

pkg/validator/benchmark_runtime_ref.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	raw := strings.TrimSpace(ref.Value)
+	if raw == "" {
+		return nil
+	}
+
+	// A ref and an inline runtime together are ambiguous — fail closed rather
+	// than silently preferring one.
+	inline, hasInline := findPerfConstraint(vi.Config.Performance.Constraints, v1.PerfConstraintNCCLBenchmarkRuntime)
+	if hasInline && strings.TrimSpace(inline.Value) != "" {
```
mchmarny: The ref+inline conflict fails fast here in the orchestrator, but the ref+`nccl-benchmark-profile` conflict is only caught in the pod (`validateNcclAllReduceBw`), after RBAC creation and Job deploy. Both are pure recipe-authoring errors visible at resolve time — consider extending this guard to also reject a non-blank `nccl-benchmark-profile` so the documented "rejected before any validator Job is deployed" semantics cover both conflicts. Non-blocking: the pod check does fail closed, just late.

pkg/validator/benchmark_runtime_ref.go
```diff
@@ -0,0 +1,201 @@
...
+			return errors.PropagateOrWrap(err, errors.ErrCodeTimeout,
+				fmt.Sprintf("timed out resolving %s=%q (%q)",
+					v1.PerfConstraintNCCLBenchmarkRuntimeRef, raw, relPath))
+		}
+		return errors.Wrap(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("failed to resolve %s=%q: expected %q in the --data tree",
+				v1.PerfConstraintNCCLBenchmarkRuntimeRef, raw, relPath), err)
+	}
+	if strings.TrimSpace(string(content)) == "" {
+		return errors.New(errors.ErrCodeInvalidRequest,
```
mchmarny, line 126: The orchestrator checks only non-emptiness; a file that isn't a `trainer.kubeflow.org/v1alpha1` `TrainingRuntime` with a `node` replicatedJob is caught only in the pod, after the validator Job (and Trainer install) has run. The pod-side shape check (`validateBenchmarkRuntime`) is a cheap in-memory parse — hoisting it (or a shared copy in `pkg/validator/v1`, next to the constraint constants) into this resolve path would fail fast offline, matching how the malformed-ref/missing-file cases already behave. Non-blocking: current behavior is documented and fails closed either way.

pkg/validator/benchmark_runtime_ref.go
```diff
@@ -0,0 +1,201 @@
...
+
+// isSafeRefSegment reports whether a single ref path segment is a plain,
+// non-empty name — no path separators, no "."/".." traversal.
+func isSafeRefSegment(s string) bool {
+	return s != "" && s != "." && s != ".." && !strings.ContainsAny(s, `/\`)
+}
+
+// findPerfConstraint returns the first constraint with the given name.
+func findPerfConstraint(cs []recipe.Constraint, name string) (recipe.Constraint, bool) {
+	for _, c := range cs {
```
mchmarny: nit: these duplicate the pod-side `findPerformanceConstraint`/`countPerformanceConstraint` (`validators/performance/nccl_all_reduce_bw.go`). Since `pkg/validator/v1` already exists as the shared vocabulary between the two ends, a single `[]recipe.Constraint` helper there would keep the lookup semantics (first-match, count) from drifting the same way the constants are protected.

mchmarny (dismissed):
Well-engineered, additive, opt-in feature with an unusually thorough fail-closed posture: every silent-skip path this could introduce is explicitly converted to an error at both the orchestrator and pod layers, with tests pinning each one (ref parse, `filepath.IsLocal` path guard, non-empty read, pod-side duplicate/kind/apiVersion/node-job checks). The fabric short-circuit is safe — every fabric-dependent branch is gated on the custom-runtime path, and transport verification still runs. Trust model is consistent with existing `--data` semantics (invoker supplies the tree and holds the kubeconfig; carrier applied only through the pinned TrainingRuntime GVR with force-set name/namespace). Logic correctly lives below the CLI boundary; docs updated across all four pages. CI fully green including real-GPU lanes. Three non-blocking polish comments inline (two fail-faster suggestions, one helper-duplication nit). Nothing blocks merge.

njhensley:
Thanks @mchmarny — addressed all three in `4cc123f4`:

1. **ref + `nccl-benchmark-profile` conflict** — now rejected in `resolveBenchmarkRuntimeRef` (orchestrator), so both the ref+inline and ref+profile conflicts fail fast before any Job, matching the documented "rejected before any validator Job is deployed" semantics.
2. **runtime shape check hoisted** — the TrainingRuntime shape validation now lives in `pkg/validator/v1` (`ValidateBenchmarkRuntime`) and runs in the resolve path, so a file that isn't a valid `trainer.kubeflow.org/v1alpha1` TrainingRuntime with a `node` replicatedJob fails offline before the Trainer install (verified e2e: `… resolved to an invalid runtime: … got apiVersion="v1" kind="ConfigMap"`). The pod delegates to the same shared function.
3. **duplicate helpers** — `FindConstraint`/`CountConstraint` now live once in `pkg/validator/v1`; both the orchestrator and the pod's ctx-based lookups delegate to them, and the constraint-name constants + a lock test keep the pod's apply-side identifiers (`trainingRuntimeGVR`, `nodeJobName`) from drifting from the shared shape-check constants.

CI green, `-race` + lint clean, and re-verified end-to-end against the simulated `--data` folder.

mchmarny (approved):
Re-approving at `4cc123f`. All three review comments addressed and verified: the ref+profile conflict now fails fast in the orchestrator before any Job deploys, the runtime shape check is hoisted into the resolve path via the shared `v1.ValidateBenchmarkRuntime`, and the constraint-lookup vocabulary is consolidated in `pkg/validator/v1` so the two ends can't drift. GPU lanes are re-running post-rebase; everything else green.

PR #1807, opened 2026-07-17
ci(uat): refresh evidence dashboard on ingest, name UAT runs by intent

njhensley:
## Summary

Refresh the GP5 evidence dashboard after every successful `evidence-ingest` run, and render the intent in UAT run titles so nightly/daytime runs are self-describing in the Actions list.

## Motivation / Context

First-party UAT evidence reaches `evidence-ingest` as a nested `workflow_call` (`uat-run` → `uat-{aws,gcp,azure}` → `evidence-ingest`) that pushes nothing to `main`, so the published dashboard stayed stale until the next unrelated push to `main`. And every dispatched UAT run rendered the same `UAT <reservation> @ <version>` title regardless of intent, making training vs inference indistinguishable at a glance.

Fixes: N/A
Related: #1402 (GP2 ingest), #1405 (GP5 dashboard), #1275/#1276 (DC2/DC3 intents)

## Type of Change

- [x] Build/CI/tooling
- [x] Documentation update

## Component(s) Affected

- [x] Other: GitHub Actions (UAT + evidence pipeline workflows)

## Implementation Notes

**Dashboard on every ingest.** A new `trigger-dashboard` job in `evidence-ingest.yaml` dispatches `evidence-dashboard-publish.yaml` (`workflow_dispatch`, `--ref main`). It runs `needs: publish`, so it inherits the existing `produced == 'true'` gate — a no-op ingest never fires it. A dispatch (rather than a nested `workflow_call`) is required because the UAT path is already at GitHub's four-level reuse limit and a nested call emits no top-level run event.
  - `actions: write` is threaded from the `uat-run` `run-*` jobs (was `actions: read`) through each cloud pipeline's `ingest-evidence` job so the nested dispatch is permitted. Held in a separate job so the GCS-credentialed publish job never also carries `actions: write`.
  - The **concurrency: 1 + queue + replace-queued** behavior the request asks for is already provided by the dashboard's `concurrency: { group: pages, cancel-in-progress: false }`: one running build + one queued build, newest dispatch superseding the queued one. A multi-cell nightly batch therefore coalesces into a single rebuild instead of a backlog.

**Descriptive UAT run names.** `uat-run.yaml`'s `run-name` now renders `UAT <reservation> <intent> @ <version>`. `intent` defaults to `training` on both the dispatch and call paths, so it is always present. The nightly (`uat-nightly-batch.yaml`) and daytime (`uat-daytime.yaml`) controllers rebuild this exact title to resolve the run they dispatched, so their `title=` construction is updated in lockstep.

## Testing

```bash
actionlint .github/workflows/{evidence-ingest,uat-run,uat-nightly-batch,uat-daytime,uat-aws,uat-gcp,uat-azure}.yaml   # clean (pre-existing SC2016 infos only)
go test ./tools/corroborate/...   # ok (dashboard-publish structural gate unchanged)
```

No Go source changed. The `evidence-dashboard-publish.yaml` structural test in `tools/corroborate` is unaffected (that workflow is not modified).

## Risk Assessment

- [x] **Low** — CI-only wiring; isolated, easy to revert. The run-name/title change is the one coupling to watch (the controllers match on the exact title), kept in lockstep and called out in-comment.

**Rollout notes:** N/A — takes effect on merge; no migration.

## Checklist

- [x] Linter passes (actionlint)

### DIFF
--- a/.github/workflows/evidence-ingest.yaml
+++ b/.github/workflows/evidence-ingest.yaml
@@ -249,6 +249,14 @@ jobs:
   # concurrency group ("pages", cancel-in-progress:false) coalesces the
   # flurry from a multi-cell nightly batch into a single queued rebuild
   # that the newest dispatch supersedes.
+  #
+  # The refresh is BEST-EFFORT: the evidence is already in the bucket and the
+  # dashboard self-heals on the next ingest or push to main, so a transient
+  # dispatch failure must never fail this workflow_call chain — otherwise the
+  # nightly controller would record a fully-successful, multi-hour UAT cell as
+  # a failed leg. The dispatch step retries, then warns (never fails), so
+  # staleness is surfaced but not fatal — the same not-cry-wolf posture as
+  # uat-superseded-notice.yaml.
   # ---------------------------------------------------------------------
   trigger-dashboard:
     name: Trigger dashboard publish
@@ -269,6 +277,17 @@ jobs:
           # Always target main: the dashboard deploys only from the default
           # branch, and the evidence tree it renders is the canonical GCS
           # bucket regardless of which ref this ingest ran on.
-          gh workflow run evidence-dashboard-publish.yaml \
-            --repo "$REPO" --ref main
-          echo "dispatched Evidence: Dashboard Publish on main"
+          #
+          # Bounded retry, then warn-and-succeed: the `if` guard keeps a failed
+          # attempt from tripping `set -e`, and the trailing warning exits 0 so a
+          # best-effort refresh cannot fail the calling UAT run.
+          for attempt in 1 2 3; do
+            if gh workflow run evidence-dashboard-publish.yaml \
+              --repo "$REPO" --ref main; then
+              echo "dispatched Evidence: Dashboard Publish on main"
+              exit 0
+            fi
+            echo "dispatch attempt ${attempt} failed; retrying in 5s..."
+            sleep 5
+          done
+          echo "::warning::could not dispatch Evidence: Dashboard Publish after 3 attempts; the dashboard will refresh on the next ingest or push to main"
--- a/.github/workflows/evidence-ingest.yaml
+++ b/.github/workflows/evidence-ingest.yaml
@@ -229,3 +229,46 @@ jobs:
                 --delete-unmatched-destination-objects \
                 "${run_dir}" "gs://${BUCKET}/results/${rel}"
             done
+
+  # ---------------------------------------------------------------------
+  # Job 3: refresh the published dashboard. Every successful ingest lands
+  # new evidence in GCS, so the GP5 dashboard (evidence-dashboard-publish)
+  # must re-render to surface it — otherwise the site is stale until the
+  # next unrelated push to main. The push-to-main ingest path already
+  # re-triggers the dashboard via its own `push:` trigger, but the
+  # first-party UAT path is a NESTED workflow_call (up to four levels deep:
+  # uat-run -> uat-{aws,gcp,azure} -> evidence-ingest), which emits no
+  # top-level run event and cannot nest a fifth reusable workflow. A
+  # workflow_dispatch decouples cleanly: it starts a fresh top-level run
+  # regardless of how deep this ingest was invoked.
+  #
+  # Held in a SEPARATE job so the GCS-credentialed publish job never also
+  # carries actions:write. `needs: publish` inherits the produced gate — a
+  # skipped/failed publish skips this too, so the dashboard only re-renders
+  # after evidence is actually in the bucket. The dashboard's own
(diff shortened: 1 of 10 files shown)


.github/workflows/evidence-ingest.yaml
```diff
@@ -229,3 +229,46 @@ jobs:
...
+  #
+  # Held in a SEPARATE job so the GCS-credentialed publish job never also
+  # carries actions:write. `needs: publish` inherits the produced gate — a
+  # skipped/failed publish skips this too, so the dashboard only re-renders
+  # after evidence is actually in the bucket. The dashboard's own
+  # concurrency group ("pages", cancel-in-progress:false) coalesces the
+  # flurry from a multi-cell nightly batch into a single queued rebuild
+  # that the newest dispatch supersedes.
+  # ---------------------------------------------------------------------
+  trigger-dashboard:
```
mchmarny, line 261: The dashboard refresh is a best-effort side-effect, but a transient `gh workflow run` failure here fails the whole `workflow_call` chain — the nightly controller then records a multi-hour, fully-successful UAT cell as a failed leg (`leg_failed=1`) even though the evidence is already in the bucket, and the dashboard would self-heal on the next ingest or push to main anyway. Consider `continue-on-error: true` on this job (with a `::warning::` on dispatch failure so staleness isn't silent), or a small bounded retry around the `gh workflow run` call. Precedent: `uat-superseded-notice.yaml` deliberately skips on a `gh api` fetch failure rather than crying wolf.

.github/workflows/uat-run.yaml
```diff
@@ -21,7 +21,14 @@ name: UAT Run
...
 # Quoted so YAML does not treat the ` #` in the dispatch_key suffix as a comment.
-run-name: "UAT ${{ inputs.reservation }} @ ${{ inputs.aicr_version || 'main' }}${{ inputs.dispatch_key != '' && format(' #{0}', inputs.dispatch_key) || '' }}"
+#
+# The intent (training|inference) is rendered so nightly/daytime runs are
+# self-describing in the Actions list. It defaults to `training` on both the
+# dispatch and call paths, so it is always present. IMPORTANT: the nightly
+# (uat-nightly-batch.yaml) and daytime (uat-daytime.yaml) controllers rebuild
+# this exact title to resolve the run they dispatched — keep their `title=`
+# construction in lockstep with this format string.
+run-name: "UAT ${{ inputs.reservation }} ${{ inputs.intent }} @ ${{ inputs.aicr_version || 'main' }}${{ inputs.dispatch_key != '' && format(' #{0}', inputs.dispatch_key) || '' }}"
```
mchmarny, line 31: nit: `uat-superseded-notice.yaml` (line 89) also describes this run-name format (`"UAT <reservation> @ <version> ..."`) and is now stale — it only displays the title (no parsing), so nothing breaks, but since this PR updates the other two format describers in lockstep, worth updating that comment too.

mchmarny (dismissed):
Solid CI wiring: the dispatch-based dashboard refresh is the right mechanism given the four-level workflow-reuse limit, the `actions: write` permission is threaded completely through every caller in the chain (all three cloud pipelines + all three `uat-run` jobs — none missed) while staying isolated from the GCS-credentialed publish job, and the run-title format change is kept in lockstep with both controllers that exact-match on it. The `pages` concurrency group coalesces the push-trigger/dispatch race. One substantive non-blocking comment inline: a flaky dashboard dispatch currently fails an otherwise-successful multi-hour UAT cell — suggest making the refresh job best-effort. Plus one stale-comment nit. Nothing blocks merge.

njhensley:
Thanks @mchmarny — both addressed in 92ecc13a:

1. **Best-effort dashboard refresh** — the `trigger-dashboard` dispatch now runs a bounded retry (3×) and, on exhaustion, emits a `::warning::` and exits 0 so a transient `gh workflow run` failure can never fail the `workflow_call` chain (no more falsely-red nightly leg for a successful multi-hour cell). Same not-cry-wolf posture as `uat-superseded-notice.yaml`; the job comment now spells out the best-effort contract. Went with the retry-then-warn shape over `continue-on-error: true` so the job stays green and the staleness reason is visible in the log rather than the job showing as failed-but-ignored.
2. **Stale comment** — updated `uat-superseded-notice.yaml`'s run-name format comment to include `<intent>`, in lockstep with the other two describers.

mchmarny (approved):
Re-approving at `92ecc13`. Both comments addressed and verified: `trigger-dashboard` now does a bounded retry then warns-and-succeeds, so a flaky dispatch can no longer fail an otherwise-successful UAT leg, and the stale run-name format description in `uat-superseded-notice.yaml` is updated to the new `<reservation> <intent> @ <version>` shape. CI fully green.

PR #1810, opened 2026-07-17
fix(validate): never emit evidence in --no-cluster dry-run mode

njhensley:
## Summary

`validate --no-cluster` (an offline dry-run) still emitted and signed a full evidence bundle when the config set `spec.validate.evidence.attestation.*`, writing `attestation.intoto.jsonl` into the shared evidence output dir. In the UAT flow this lets the prep phase's leftover signature get packed into the authoritative run-tag artifact, which breaks `aicr evidence verify` and the `Evidence: Ingest` job with a signed-subject/pulled-digest mismatch.

## Motivation / Context

The UAT emit output dir is `attestation.out: ./evidence`, shared across phases in the runner workspace.

1. `prep` runs `validate --no-cluster` (dry-run, all checks `skipped`). It still honored the config's evidence block, signed a bundle to the aicr content-hash tag (`…-59266d32abc3`, digest `d3466f4c…`), and left `attestation.intoto.jsonl` (subject = `d3466f4c…`) in `./evidence/`.
2. `conformance` then runs the authoritative `validate --phase all`, which re-packs `./evidence/` into the `run-<id>` artifact (`2a486d77…`). The closed-world pack sweeps in prep's stale `attestation.intoto.jsonl`, so **the run-tag artifact embeds a signature for a foreign subject** (`d3466f4c…`, not `2a486d77…`).
3. Normally the conformance emit's own Sigstore **referrer overwrites** the staged attestation at verify time, masking this. On the failing run the referrer push hit GHCR timeouts and never attached, so `MaterializeBundle` found no referrer to overwrite the embedded stale file, and `VerifySignature` read the embedded `d3466f4c` attestation → mismatch.

Directly confirmed against the registry:
- `oras pull …@sha256:2a486d77…` → the artifact **embeds `attestation.intoto.jsonl`**, and its DSSE payload signs subject `d3466f4c…` (prep's content-tag ref).
- `oras discover …@sha256:2a486d77…` → **zero referrers** (the conformance signature never attached).

This is why it is intermittent / "new": it needs prep's stale attestation (present on every emit-enabled config) **and** the conformance referrer failing to attach (a GHCR write hiccup). Healthy runs overwrite the stale file via the referrer and pass — which is why the H100/release cells are green. Observed on run 29611408651 (verify) and 29619695743 (ingest). An offline dry-run over an all-`skipped` run attests to nothing, so it should never sign or write an attestation in the first place.

Fixes: N/A
Related: #1793 (@main verify regression since #1758 closed-world pack)

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT harness (`tests/uat/lib/phases.sh`)

## Implementation Notes

Two layers so a dry-run can never leave a signed attestation behind:

1. **CLI (authoritative):** `evidenceConfigForRunMode` drops the recipe-evidence config whenever `--no-cluster` is set, so an offline dry-run can never sign, push, or write `attestation.intoto.jsonl`, regardless of config. Logs once when it suppresses emission.
2. **UAT harness (belt-and-suspenders):** the `prep` phase validates against a config copy with `spec.validate.evidence` stripped (mirroring the readiness-gate strip already used in the conformance phase). Keeps `prep` safe even against an older released `aicr` that predates the CLI guard.

With prep no longer emitting, `./evidence/` holds no stale attestation for the conformance pack to embed, so the run-tag artifact carries only its own signature. Producer-side fix — no change needed in verify/ingest.

Possible follow-up (out of scope): have the emit exclude any pre-existing `attestation.intoto.jsonl` from the pack, and/or have `MaterializeBundle` prefer the referrer over an embedded attestation — defense-in-depth against any stray signed file in the output dir.

## Testing

```bash
go build ./pkg/cli/...
go test ./pkg/cli/ -count=1            # ok (incl. new TestEvidenceConfigForRunMode)
golangci-lint run -c .golangci.yaml ./pkg/cli/...   # 0 issues
shellcheck -x tests/uat/lib/phases.sh  # clean (SC1091 info only)
```

Added `TestEvidenceConfigForRunMode` covering live-cluster pass-through, `--no-cluster` suppression, and nil-in/nil-out. Verified the mechanism against the live registry with `oras pull` / `oras discover` (see Motivation).

## Risk Assessment

**Rollout notes:** Behavior change is limited to `--no-cluster` runs, which are dry-runs that should not have been producing evidence; live-cluster validate (the only path that emits authoritative evidence) is unchanged. Backwards compatible.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — ran `go test ./pkg/cli/`
- [x] Linter passes (`make lint`) — `golangci-lint` 0 issues on `pkg/cli`

### DIFF
--- a/pkg/cli/validate.go
+++ b/pkg/cli/validate.go
@@ -617,6 +617,10 @@ Run validation without failing on check errors (informational mode):
 			evidenceDir := stringFlagOrConfig(cmd, "evidence-dir", cncfCfg.Dir)
 			cncfSubmission := boolFlagOrConfig(cmd, "cncf-submission", cncfCfg.CNCFSubmission)
 			features := stringSliceFlagOrConfig(cmd, "feature", cncfCfg.Features)
+			// Resolved once here (flag > config) and reused by the guards below and
+			// the mode banner further down.
+			noCluster := boolFlagOrConfig(cmd, "no-cluster", resolved.NoCluster)
+			explicitAttest := cmd.IsSet("emit-attestation") || cmd.IsSet(flagPush)
 
 			// Validate flag combinations.
 			if cncfSubmission && evidenceDir == "" {
@@ -624,6 +628,27 @@ Run validation without failing on check errors (informational mode):
 			}
 			if len(features) > 0 && !cncfSubmission {
 				return errors.New(errors.ErrCodeInvalidRequest, "--feature requires --cncf-submission")
+			}
+			// --cncf-submission deploys GPU workloads and collects behavioral
+			// evidence against the active kube-context, so it is incompatible with
+			// the offline --no-cluster dry-run. The short-circuit below reaches the
+			// live-cluster collector directly, bypassing the noCluster handling
+			// further down, so reject the combination here rather than silently
+			// contacting the cluster.
+			if cncfSubmission && noCluster {
+				return errors.New(errors.ErrCodeInvalidRequest,
+					"--cncf-submission cannot be combined with --no-cluster: the behavioral evidence collector requires a live cluster")
+			}
+			// An explicit --emit-attestation/--push is a request to sign and
+			// (optionally) push an attestation; --no-cluster is an offline dry-run
+			// whose checks are all skipped, so the two conflict. Reject the
+			// explicit combination rather than warn-and-ignore an explicit CLI flag
+			// (repo rule) — consistent with the --cncf-submission guard above. A
+			// config-driven spec.validate.evidence.attestation is still silently
+			// suppressed in --no-cluster mode by evidenceConfigForRunMode below.
+			if noCluster && explicitAttest {
+				return errors.New(errors.ErrCodeInvalidRequest,
+					"--emit-attestation/--push cannot be combined with --no-cluster: an offline dry-run must not sign or push an attestation")
 			}
 
 			// Short-circuit: --cncf-submission bypasses normal validation and runs
@@ -657,7 +682,6 @@ Run validation without failing on check errors (informational mode):
 
 			failOnError := boolFlagOrConfig(cmd, "fail-on-error", derefBoolOr(resolved.FailOnError, true))
 			failFast := boolFlagOrConfig(cmd, "fail-fast", derefBoolOr(resolved.FailFast, false))
-			noCluster := boolFlagOrConfig(cmd, "no-cluster", resolved.NoCluster)
 
 			// Mode banner: make it explicit whether this run touches a live
 			// cluster (issue #1383). --no-cluster is an offline dry-run that
@@ -758,7 +782,7 @@ Run validation without failing on check errors (informational mode):
 					"cleanupHint", "kubectl delete clusterrolebinding -l app.kubernetes.io/name=aicr-validator")
 			}
 
-			evidenceCfg := buildRecipeEvidenceConfig(cmd, resolved)
+			evidenceCfg := evidenceConfigForRunMode(noCluster, buildRecipeEvidenceConfig(cmd, resolved))
 
 			return runValidation(ctx, client, rec, snap, validationConfig{
 				phases:                phases,
--- a/pkg/cli/validate_evidence.go
+++ b/pkg/cli/validate_evidence.go
@@ -16,6 +16,7 @@ package cli
 
 import (
 	"context"
+	"log/slog"
 	"os"
 
 	"github.com/urfave/cli/v3"
@@ -80,6 +81,28 @@ func buildRecipeEvidenceConfig(cmd *cli.Command, resolved *config.ValidateResolv
 		OIDCResolve: oidcResolveOptionsFromFlags(cmd),
 		AssumeYes:   cmd.Bool(flagAssumeYes),
 	}
+}
+
+// evidenceConfigForRunMode suppresses CONFIG-DRIVEN recipe-evidence emission in
+// --no-cluster mode. (An explicit --emit-attestation/--push alongside
+// --no-cluster is rejected upstream with ErrCodeInvalidRequest — an explicit CLI
+// flag must not be warn-and-ignored; this path handles only
+// spec.validate.evidence.attestation resolved from config.) An offline dry-run
+// reports every check as "skipped", so an emitted attestation would attest to
+// nothing. Worse, when the config sets attestation.{out,push} (as the UAT test
+// configs do), a dry-run would sign and push a bundle to the same OCI repo the
+// authoritative live-cluster validate later pushes to — leaving two
+// independently-signed bundles whose digests differ, which breaks `aicr evidence
+// verify`: the signed subject no longer matches the pulled run-tagged artifact.
+// A dry-run must never emit signed evidence, so drop the config and log once.
+// Returns cfg unchanged for a live-cluster run (including cfg == nil).
+func evidenceConfigForRunMode(noCluster bool, cfg *recipeEvidenceConfig) *recipeEvidenceConfig {
+	if noCluster && cfg != nil {
+		slog.Warn("skipping evidence emission: --no-cluster is an offline dry-run and must not sign or push an attestation",
+			"emitAttestation", cfg.OutDir, "push", cfg.Push)
+		return nil
+	}
+	return cfg
 }
 
 // oidcResolveOptionsFromFlags builds the keyless-signing OIDC resolution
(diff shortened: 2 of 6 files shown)


njhensley:
Addressed the CodeRabbit review (all 3 findings) in the amended commit:

- **`--cncf-submission --no-cluster` (Major):** now rejected in the flag-combination validation — the CNCF collector short-circuits straight to the live cluster, bypassing the `noCluster` handling, so the combination fails closed with `ErrCodeInvalidRequest` instead of silently contacting the cluster. Added a table case to `TestValidateCmd_CNCFSubmissionFlagValidation`.
- **Docs claim too broad (Minor):** narrowed the `--no-cluster` row to state that only the recipe-evidence *attestation* is suppressed (not signed/pushed); `--evidence-dir` CNCF conformance markdown is still rendered locally.
- **Temp config leak (Minor):** `phase_prep` now captures the validate rc and removes the temp config on the failure path too (mirrors the `vrc` pattern in `phase_conformance`).

`go build`/`go test ./pkg/cli/` (incl. new cases), `golangci-lint` (0 issues), and `shellcheck -x` all pass locally.

tests/uat/lib/phases.sh
```diff
@@ -170,11 +170,25 @@ phase_prep() {
...
+  local prep_config
+  prep_config="$(mktemp)"
+  yq 'del(.spec.validate.evidence)' "${config}" > "${prep_config}"
   "${AICR_BIN}" validate \
-    --config "${config}" \
+    --config "${prep_config}" \
     --phase deployment \
     --no-cluster \
     --output dry-run.json
+  rm -f "${prep_config}"
```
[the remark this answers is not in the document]
njhensley: Addressed in 168d1d3b: the dry-run validate now runs in a subshell with a `trap 'rm -f "${prep_config}"' EXIT`, so the temp config is removed on every exit path (normal, `set -e` abort, interrupt) — mirroring the signing subshell at phases.sh:240–248 rather than the normal-path-only `rm`.

pkg/cli/validate_test.go
```diff
@@ -205,6 +205,12 @@ func TestValidateCmd_CNCFSubmissionFlagValidation(t *testing.T) {
 			wantErr:    true,
 			errContain: "unknown feature",
 		},
+		{
+			name:       "cncf-submission with no-cluster",
+			args:       []string{"aicr", "validate", "--cncf-submission", "--evidence-dir", "/tmp/test", "--no-cluster"},
+			wantErr:    true,
+			errContain: "--cncf-submission cannot be combined with --no-cluster",
+		},
```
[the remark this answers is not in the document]
njhensley: Addressed the code assertion in 198e45cf: every rejection case in `TestValidateCmd_CNCFSubmissionFlagValidation` now asserts `stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, ""))` in addition to the message text.

On the config-sourced `no-cluster` case: skipping intentionally. The guard resolves the value through the shared `boolFlagOrConfig(cmd, "no-cluster", resolved.NoCluster)` helper, and the config→`resolved.NoCluster` mapping (`spec.validate.execution.noCluster`) is already covered by `pkg/config` resolve tests (`resolve_test.go`). Re-exercising that same helper through a temp-config fixture here would duplicate existing coverage and diverge from the sibling guards in this table (all flag-tested only), for no additional signal.

pkg/cli/validate_evidence.go
```diff
@@ -82,6 +83,25 @@ func buildRecipeEvidenceConfig(cmd *cli.Command, resolved *config.ValidateResolv
...
+// mode. An offline dry-run reports every check as "skipped", so an emitted
+// attestation would attest to nothing. Worse, when the config sets
+// spec.validate.evidence.attestation.{out,push} (as the UAT test configs do), a
+// dry-run would sign and push a bundle to the same OCI repo the authoritative
+// live-cluster validate later pushes to — leaving two independently-signed
+// bundles whose digests differ, which breaks `aicr evidence verify`: the signed
+// subject no longer matches the pulled run-tagged artifact. A dry-run must never
+// emit signed evidence, so drop the config and log once. Returns cfg unchanged
+// for a live-cluster run (including cfg == nil).
+func evidenceConfigForRunMode(noCluster bool, cfg *recipeEvidenceConfig) *recipeEvidenceConfig {
```
mchmarny, line 99: The config-sourced suppression here is exactly right (that's the UAT case this PR fixes), but when the user explicitly types `--emit-attestation` or `--push` alongside `--no-cluster`, silently dropping it is inconsistent with the hard `ErrCodeInvalidRequest` you added for `--cncf-submission --no-cluster`, and with the repo rule that an explicit CLI flag must not be warn-and-ignored. Consider `cmd.IsSet("emit-attestation") || cmd.IsSet(flagPush)` → return `ErrCodeInvalidRequest`, keeping the warn+nil path for config-driven values only.

pkg/cli/validate.go
```diff
@@ -758,7 +768,7 @@ Run validation without failing on check errors (informational mode):
 					"cleanupHint", "kubectl delete clusterrolebinding -l app.kubernetes.io/name=aicr-validator")
 			}
 
-			evidenceCfg := buildRecipeEvidenceConfig(cmd, resolved)
+			evidenceCfg := evidenceConfigForRunMode(noCluster, buildRecipeEvidenceConfig(cmd, resolved))
```
mchmarny, line 785: nit: this guard holds only at the CLI boundary — a `pkg/client/v1` caller can still emit/sign an all-skipped no-cluster result set via `Client.EmitRecipeEvidence`. Fine for this producer-side fix, but worth folding into your noted follow-up: a facade-level refusal (or pack exclusion of pre-existing `attestation.intoto.jsonl`) would make the invariant hold below the CLI, per the "safety guards belong below the CLI boundary" convention.

mchmarny (dismissed):
The fix is correct and complete: `--no-cluster` can no longer produce a signed attestation via any CLI path (config- or flag-sourced), the previously-silent `--cncf-submission --no-cluster` cluster contact is now rejected with a structured `ErrCodeInvalidRequest` before the short-circuit, live-cluster emission is provably untouched (pass-through identity is tested), and the UAT `yq del` strip mirrors the existing readiness-gate precedent exactly. CI fully green including the KWOK matrix and GPU lanes; docs updated in the same PR. Two non-blocking suggestions inline (explicit-flag warn-vs-error asymmetry, and the guard living only at the CLI boundary). Nothing blocks merge.

pkg/cli/validate_evidence.go
njhensley: Done in 89f117d0. Added a hard reject in the flag-combination block: `if noCluster && (cmd.IsSet("emit-attestation") || cmd.IsSet(flagPush))` → `ErrCodeInvalidRequest` ("--emit-attestation/--push cannot be combined with --no-cluster"), right next to the `--cncf-submission` guard for consistency. The warn+nil `evidenceConfigForRunMode` path now handles config-driven `spec.validate.evidence.attestation` only (its doc-comment says so explicitly). Covered by a new `TestValidateCmd_NoClusterEvidenceFlags` (asserts both the `ErrCodeInvalidRequest` code and the message, for `--emit-attestation` and `--push`).

pkg/cli/validate.go
njhensley: Agreed — keeping this PR producer-side and folding the below-the-CLI invariant into the follow-up. The stronger fix I noted in the PR description is a pack-side exclusion of any pre-existing `attestation.intoto.jsonl` in the emit output dir (in `pkg/evidence` packaging), which makes the invariant hold for every caller including `Client.EmitRecipeEvidence` — that directly kills the embedded-stale-signature root cause regardless of who calls it. I'll take that route rather than a facade-level `noCluster` refusal, since the facade cannot reliably tell an offline dry-run's result set apart. Will open the follow-up against this once it merges.

mchmarny (approved):
Re-approving at `71400da`. Both comments addressed and verified: an explicitly typed `--emit-attestation`/`--push` with `--no-cluster` now hard-rejects with `ErrCodeInvalidRequest` (config-sourced values keep the warn+suppress path, which is the right split), and the below-the-CLI facade guard is explicitly deferred to the noted follow-up with the pack-side exclusion approach. GPU training lane re-running post-rebase; everything else green.

PR #1811, opened 2026-07-17
fix(validator): gate GPU readiness on runtime-required taint clearance

### LINKED ISSUE #1775: Readiness gate keys off non-monotonic Skyhook status, not the tuning taint → flaky post-gate failures
## Summary

The GPU-stack readiness gate can certify a cluster "ready" and then have skyhook/nodewright node tuning re-open, producing intermittent downstream failures in the conformance phase. Root cause: the gate keys off a **non-monotonic** signal (`Skyhook status.status`) rather than the durable terminal signal (the tuning `NoSchedule` taint being removed). This is the RCA of the two UAT-AWS `expected-resources` failures that motivated the debug bundle (#1764).

## Evidence

Run `29393800033`, one `expected-resources` run, the **same** `Skyhook/tuning` CR:
- `07:03:53` — Go pre-flight printed `Nodewright tuning: complete (stable ≥1m0s)` (i.e. `status.status == complete` continuously for 60s) → passed.
- `07:08:53` — the chainsaw step `validate-skyhook-cr-complete` **failed** reading `status.status: in_progress` (persisted through its 5m assert budget).

So the same CR went `complete` → `in_progress` within one check. Run `29402956254` is the same race with the tuning reboot evicting the check Job's pod (`pod for job not found`).

## Why the signal flaps

The `tuning` Skyhook ([`recipes/components/nodewright-customizations/manifests/tuning.yaml`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/recipes/components/nodewright-customizations/manifests/tuning.yaml#L101)) is a multi-package chain where `nvidia-tuned` ([L105](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/recipes/components/nodewright-customizations/manifests/tuning.yaml#L105)) and `nvidia-setup-full` ([L126](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/recipes/components/nodewright-customizations/manifests/tuning.yaml#L126)) both carry `interrupt: reboot`. `status.status` is an aggregate over (packages × matching GPU nodes), and the cluster runs 2 GPU nodes — so it re-opens to `in_progress` on each package reboot and per newly-joined node.

## Why the current gate doesn't close it

Both gate layers watch `status.status`, **not** the taint:
- **Runner gate** — [`tests/uat/aws/run#L307`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/tests/uat/aws/run#L307): `validate --phase deployment` until it passes `READINESS_CONSECUTIVE_PASSES` ([=2](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/tests/uat/aws/run#L95)) times consecutively.
- **Go pre-flight** — [`verifyNodewrightReady`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/validators/deployment/expected_resources.go#L403) / [`pollUntilStable`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/validators/deployment/expected_resources.go#L90) polls `status.status` ([read at L505](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/validators/deployment/expected_resources.go#L505)) until "complete" stable for [`GPUReadinessStabilityWindow` = 60s](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/pkg/defaults/timeouts.go#L363).

Both are **probabilistic settling heuristics** — "quiet at N sampled instants." They cannot guarantee no further reboot, and the runner gate closes in `phase_install` minutes before `phase_conformance` opens a fresh run; a reboot in that gap is invisible. The code even acknowledges the non-monotonicity ([expected_resources.go#L442](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/validators/deployment/expected_resources.go#L442)).

## Proposed direction

Gate on the **durable terminal signal** instead of the flapping status:
1. **Gate on taint-absence being stable** — skyhook applies a `skyhook.nvidia.com` `NoSchedule` taint during tuning and removes it on completion. If that removal is skyhook's monotone terminal step (i.e. it does not re-taint for a later package/reboot in the same plan), "taint absent + stable" is a stronger "done, won't reboot again" signal than `status.status`.
2. **And/or** give the chainsaw `validate-skyhook-cr-complete` assert ([`recipes/checks/nodewright-customizations/health-check.yaml#L31`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/recipes/checks/nodewright-customizations/health-check.yaml#L31)) the same poll-until-stable ride-through budget the Go barrier has, so it doesn't fail on a transient re-open.
3. **Possibly** lengthen `GPUReadinessStabilityWindow` so "stable" better predicts no re-tuning.

## Open questions (needs data)

- **Is skyhook's taint removal monotone-terminal within a plan?** If skyhook re-taints between packages (nvidia-setup-kernel → nvidia-tuned → nvidia-setup-full, each a reboot), taint-absence would flap too and (1) buys nothing. This must be confirmed against skyhook's behavior before changing the gate.
- The debug bundle (#1764) now captures the per-node/per-package Skyhook status timeline + node reboot fingerprint + taint state at failure, which is exactly the data needed to answer this on the next occurrence.

## References
- Diagnostic tooling: #1764 (debug bundle). Runs: `29393800033`, `29402956254`.
- Related prior work: commit `d47ce33` (poll GPU readiness to ride through reboots) added the current `pollUntilStable` mitigation.

njhensley:
## Summary

Gate the deployment-phase GPU readiness check on the durable **runtime-required taint** being cleared, not just the non-monotonic `Skyhook status.status`.

## Motivation / Context

The readiness gate keyed off `Skyhook status.status`, which is non-monotonic during nodewright tuning: it aggregates over (packages × GPU nodes) and re-opens to `in_progress` on each package reboot, and can momentarily read `complete` in the lull between two package reboots while tuning is still in flight. The gate could therefore certify a cluster "ready" and then have tuning re-open, producing intermittent downstream failures in the conformance phase (RCA of the two UAT-AWS `expected-resources` failures behind runs `29393800033` / `29402956254`).

The durable terminal signal is the `skyhook.nvidia.com=runtime-required:NoSchedule` taint the nodewright (skyhook) operator applies when a GPU node joins and removes **once per node** when every `runtimeRequired: true` skyhook completes on that node. Per the operator's `runtime_required` docs, that removal is a per-node monotone terminal step that bridges the package reboots (it flaps only when `reapplyOnReboot=true`, which the failing `tuning.yaml` recipe does not set — only the `gke-cos`/`bcm` overlays do, and there the existing stability window rides through). So "taint absent + stable" is a stronger "done, won't reboot again" signal than the status. This resolves the open question in the issue.

Fixes: #1775
Related: #1764 (debug bundle that captured the taint timeline)

## Type of Change

## Component(s) Affected

- [x] Other: `validators/deployment` (deployment-phase readiness pre-flight) + its tests

## Implementation Notes

- `verifyNodewrightReady` now polls **two** signals to stability instead of one: every expected Skyhook CR `status.status == complete` **and** no node still carries the runtime-required NoSchedule taint. During the between-reboot lull the taint is still present, so the probe stays unhealthy until tuning is truly done on every node.
- A node-list error is returned so the poll **fails closed** (resets its dwell and retries) — "could not list nodes" must never be read as "taint absent".
- When the Skyhook CRD is not registered, the check still skips early, so a cluster without the operator is unaffected.
- Scoped to the Go pre-flight (the gate the UAT runner loop actually blocks on). Fixing it also fixes the downstream chainsaw `validate-skyhook-cr-complete` failure, since the Go gate and chainsaw run back-to-back in one `checkExpectedResources` — the CR is stably `complete` by the time chainsaw runs. The chainsaw assert and stability window (issue directions 2/3, framed as optional) are intentionally left unchanged.

## Testing

```bash
go test -race ./validators/deployment/...              # ok
golangci-lint run -c .golangci.yaml ./validators/deployment/...   # 0 issues
```

New tests: taint matcher table test; gate fails while a node is tainted despite `status: complete` (the race); gate passes once cleared while ignoring unrelated taints; poll rides through a lingering-then-cleared taint and times out while it persists; node-list error fails closed. New functions are 100% covered.

## Risk Assessment

- [x] **Low** — Isolated to the deployment-phase readiness pre-flight, additive (strictly stricter) gating, well-tested, easy to revert. No user-facing surface changes.

**Rollout notes:** N/A. The gate can only stay closed *longer* than before (until the taint clears); it never opens earlier. Clusters without the skyhook operator/CRD are unaffected (early skip).

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — internal readiness-gate behavior, no user-facing surface)

### DIFF
--- a/validators/deployment/expected_resources.go
+++ b/validators/deployment/expected_resources.go
@@ -57,6 +57,32 @@ const (
 	draKubeletPluginSuffix = "-kubelet-plugin"
 
 	nodewrightCompleteState = "complete"
+
+	// runtimeRequiredTaintKey / runtimeRequiredTaintValue identify the
+	// workload-gate taint the nodewright (skyhook) operator manages for Skyhook
+	// CRs with runtimeRequired: true (see tuning.yaml `runtimeRequired: true`).
+	//
+	// Why gate on this taint and not status.status: a GPU node joins carrying
+	// this NoSchedule taint, and the operator removes it once *all*
+	// runtime-required Skyhooks targeting that node are complete *on that node*
+	// (per-node, not per-package). Unlike status.status — an aggregate over
+	// (packages × matching nodes) that re-opens to in_progress on every package
+	// reboot and each newly-joined node — the taint is applied once and removed
+	// once as the monotone terminal step, so "taint absent" is a durable
+	// "done, won't reboot again" signal rather than a probabilistic settling
+	// heuristic (see issue #1775). Note the operator re-applies the taint across
+	// reboots only when configured with REAPPLY_ON_REBOOT/reapplyOnReboot=true
+	// (the gke-cos and bcm overlays); on those the taint flaps like the status
+	// and the stability window rides through it, so gating on the taint is never
+	// weaker than gating on the status.
+	//
+	// Values match the skyhook chart's default
+	// controllerManager.manager.env.runtimeRequiredTaint
+	// (skyhook.nvidia.com=runtime-required:NoSchedule), which AICR ships
+	// unchanged and the UAT GPU node pools pre-taint with verbatim
+	// (tests/uat/aws/cluster-config.yaml).
+	runtimeRequiredTaintKey   = "skyhook.nvidia.com"
+	runtimeRequiredTaintValue = "runtime-required"
 )
 
 var (
@@ -437,27 +463,66 @@ func verifyNodewrightReady(ctx *validators.Context, ref recipe.ComponentRef) err
 		return err
 	}
 
-	// Poll status.status until every expected Skyhook CR reports "complete"
-	// continuously for the stability window, or the budget elapses. status.status
-	// is non-monotonic during tuning: a reboot (or a newly-joined GPU node)
-	// re-opens it to in_progress, so a single sample is not terminal. Polling
-	// rides through the reboot flaps rather than failing the whole deployment
-	// phase on a transient in_progress. See pkg/defaults GPUReadiness* for sizing.
+	// Poll two signals until both hold continuously for the stability window, or
+	// the budget elapses:
+	//
+	//  1. Every expected Skyhook CR reports status.status == "complete".
+	//  2. No node still carries the runtime-required NoSchedule taint the
+	//     operator removes as its monotone terminal step.
+	//
+	// status.status alone is non-monotonic during tuning: a reboot (or a
+	// newly-joined GPU node) re-opens it to in_progress, and — worse — it can
+	// momentarily read "complete" in the lull between two package reboots while
+	// tuning is still in flight, which is exactly how the gate certified a
+	// cluster ready and then had tuning re-open post-gate (issue #1775). Adding
+	// the taint gate closes that hole: during such a lull the runtime-required
+	// taint is still present, so the probe stays unhealthy until tuning is truly
+	// done on every node. Polling rides through the reboot flaps rather than
+	// failing the deployment phase on a transient in_progress / re-taint. See
+	// pkg/defaults GPUReadiness* for sizing.
 	return pollUntilStable(ctx,
-		fmt.Sprintf("%d expected Nodewright(s)", len(expectedNames)),
+		fmt.Sprintf("%d expected Nodewright(s) + runtime-required taint clearance", len(expectedNames)),
 		func() error {
-			failures := nodewrightStatusFailures(ctx, dynClient, expectedNames)
+			// The Skyhook status Gets and the node-list taint scan are
+			// independent read-only calls, so fan them out (per repo CLAUDE.md
+			// "Sequential calls to N independent read-only K8s APIs → fan-out
+			// with errgroup") rather than paying both round-trips serially every
+			// poll iteration.
+			var statusFailures, taintFailures []string
+			var taintErr error
+			g := new(errgroup.Group)
+			g.Go(func() error {
+				statusFailures = nodewrightStatusFailures(ctx, dynClient, expectedNames)
+				return nil
+			})
+			g.Go(func() error {
+				taintFailures, taintErr = runtimeRequiredTaintFailures(ctx)
+				return nil
+			})
+			_ = g.Wait()
+
+			if taintErr != nil {
+				// A transient node-list failure (e.g. an apiserver hiccup while
+				// a GPU node reboots) must not be read as "taint absent". Return
+				// it so the poll resets the dwell and retries — fail closed.
+				return taintErr
+			}
+			failures := make([]string, 0, len(statusFailures)+len(taintFailures))
+			failures = append(failures, statusFailures...)
+			failures = append(failures, taintFailures...)
 			if len(failures) == 0 {
 				return nil
 			}
 			return errors.New(errors.ErrCodeInternal,
-				fmt.Sprintf("%d of %d expected Nodewright(s) not ready:\n  %s",
-					len(failures), len(expectedNames), strings.Join(failures, "\n  ")))
+				fmt.Sprintf("%d Nodewright readiness signal(s) not settled:\n  %s",
+					len(failures), strings.Join(failures, "\n  ")))
 		},
 		func() {
 			for _, name := range expectedNames {
 				fmt.Printf("  Nodewright %s: %s (stable ≥%s)\n", name, nodewrightCompleteState, gpuReadinessStabilityWindow)
 			}
+			fmt.Printf("  Nodewright runtime-required taint (%s=%s:%s): cleared from all nodes (stable ≥%s)\n",
+				runtimeRequiredTaintKey, runtimeRequiredTaintValue, corev1.TaintEffectNoSchedule, gpuReadinessStabilityWindow)
 		})
 }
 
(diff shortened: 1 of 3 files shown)


validators/deployment/expected_resources.go
```diff
@@ -57,6 +57,32 @@ const (
...
+	// (the gke-cos and bcm overlays); on those the taint flaps like the status
+	// and the stability window rides through it, so gating on the taint is never
+	// weaker than gating on the status.
+	//
+	// Values match the skyhook chart's default
+	// controllerManager.manager.env.runtimeRequiredTaint
+	// (skyhook.nvidia.com=runtime-required:NoSchedule), which AICR ships
+	// unchanged and the UAT GPU node pools pre-taint with verbatim
+	// (tests/uat/aws/cluster-config.yaml).
+	runtimeRequiredTaintKey   = "skyhook.nvidia.com"
```
mchmarny, line 84: The comment says the values match the skyhook chart default "which AICR ships unchanged", but `--workload-gate` is a documented bundle flag that overrides exactly this value (`controllerManager.manager.env.runtimeRequiredTaint` via `accelerated.taintStrPaths` in `recipes/registry.yaml`, applied in `pkg/bundler/bundler.go`). The `docs/user/cli-reference.md` example even uses a different taint shape (`skyhook.nvidia.com/runtime-required=true:NoSchedule` — key `skyhook.nvidia.com/runtime-required`, value `true`), which `isRuntimeRequiredTaint` will never match. On a cluster bundled with a custom workload gate, this new gate silently becomes a no-op and readiness reverts to status-only gating — the #1775 race stays open there. Never *weaker* than pre-PR, so non-blocking, but either (a) derive the taint from the nodewright-operator ref's values/overrides with this constant as fallback, or (b) at minimum correct the comment to state the known limitation so the next reader doesn't assume custom-gate clusters are covered.

validators/deployment/expected_resources.go
```diff
@@ -515,6 +562,52 @@ func nodewrightStatusFailure(verifyCtx context.Context, dynClient dynamic.Interf
...
+//
+// A List error (transient apiserver failure, RBAC gap) is returned so the
+// caller fails closed: "could not list nodes" must never be read as "taint
+// absent". The error rides through the poll's dwell reset like any other
+// unhealthy sample.
+func runtimeRequiredTaintFailures(ctx *validators.Context) ([]string, error) {
+	listCtx, cancel := ctx.Timeout(defaults.ResourceVerificationTimeout)
+	defer cancel()
+
+	nodes, err := ctx.Clientset.CoreV1().Nodes().List(listCtx, metav1.ListOptions{})
```
mchmarny, line 598: The gate requires the taint absent from every node in the cluster, but the operator only removes it from nodes matched by a `runtimeRequired: true` Skyhook. A pre-tainted node outside the Skyhook's node selector (pool sizing drift, UAT config mismatch with `tuning.yaml` selectors) will hold the gate closed for the entire poll budget with a message that reads as "tuning not complete" when tuning will in fact never touch that node. Fail-closed is the right direction, so this is non-blocking — but consider noting the "node not targeted by any Skyhook" possibility in the per-node failure string so operators debugging a timed-out gate check the selector, not the operator.

mchmarny (dismissed):
Solid fail-closed strengthening of the GPU readiness pre-flight: the taint-clearance gate closes the #1775 window where aggregate Skyhook status momentarily reads complete between package reboots, every error path fails closed (List errors reset the stability dwell rather than passing), and the tests cover the exact regression race plus the matcher, ride-through, timeout, and fail-closed paths. CI fully green. Two non-blocking comments inline: the hardcoded chart-default taint makes the gate a silent no-op on clusters bundled with a custom `--workload-gate` (never weaker than pre-PR, but worth a follow-up), and the all-nodes scope can stall the full poll budget on nodes no Skyhook targets. Nothing blocks merge.

njhensley:
Addressed both CodeRabbit findings in bb3ec0bd:

1. **Concurrent probes** — the Skyhook status Gets and the node-list taint scan now fan out via `errgroup` in the poll closure instead of running serially each iteration (matching the repo pattern for independent read-only K8s calls). The taint-list error is still surfaced fail-closed after `Wait()`.
2. **Cancellation in node scan** — `runtimeRequiredTaintFailures` now checks `listCtx.Done()` per node and returns an `ErrCodeTimeout`-wrapped error, so a canceled verification stops walking a large node list.

`go test -race ./validators/deployment/...` green, `golangci-lint` clean.

mchmarny (approved):
Re-approving at `bb3ec0b`. The delta since my last review (errgroup fan-out of the two read-only probes, ctx cancellation in the node scan) is sound and CI is fully green. For the record: my two earlier comments (hardcoded chart-default taint vs a custom `--workload-gate`, and the per-node failure-string hint for un-targeted nodes) are not in this delta — both were explicitly non-blocking and remain follow-up material, so they don't hold up this approval. I'm leaving those two threads open.

PR #1845, opened 2026-07-21
fix(corroborate): partial phase rollup + show untested categories

njhensley:
## Summary

Two evidence-dashboard category (phase) rollup fixes: mixed passing/untested categories now roll up to `PARTIAL` instead of `UNTESTED`, and an entirely-untested category is shown as an untested coverage gap instead of being omitted from the board.

## Motivation / Context

On the corroboration dashboard, a category like `conformance` with some rows tested-and-passing and others not yet run was dragged all the way to `UNTESTED` by the worst-first rollup precedence, hiding the coverage it did have. Separately, a category with no rows at all was dropped from the recipe matrix and per-source drilldown entirely, so a category nobody has run appeared silently missing rather than as a visible gap.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: evidence corroboration dashboard (`pkg/corroborate`, renderer)

## Implementation Notes

- **`PARTIAL` phase rollup** — `RollupPhase` (Go, authoritative in `pkg/corroborate/consensus.go`) and the renderer's `rollupPhase` now promote a category to `PARTIAL` when the worst state is `UNTESTED` but at least one row is passing (`CONFIRMED`/`SINGLE`). `CONTESTED` and `FAILING` still outrank `UNTESTED`, so a phase only reaches `PARTIAL` when nothing failed; an all-untested phase stays `UNTESTED`. `PARTIAL` remains a rollup-only state — individual rows never carry it. The overview grid picks this up automatically via the baked Go `PhaseRollup`.
- **Show empty categories** — the recipe matrix and per-source drilldown previously did `if (tests.length === 0) return;`, omitting empty phases. They now render the phase header plus an `UNTESTED` placeholder row (empty/not-run tiles across all columns).
- Updated `docs/user/evidence-dashboard.md` Phase rollup section for both behaviors.

## Testing

```bash
go test ./pkg/corroborate/...
golangci-lint run -c .golangci.yaml ./pkg/corroborate/...
gofmt -l pkg/corroborate/
```

`TestRollupPhase` updated: the former "untested outranks confirmed" case now asserts `PARTIAL`, with added cases covering partial/failing/contested combinations with untested rows. Package tests pass, lint reports 0 issues, gofmt clean.

## Risk Assessment

- [x] **Low** — Isolated to dashboard rollup/rendering; no change to evidence data, consensus counting, or verification. Easy to revert.

**Rollout notes:** N/A — the dashboard re-renders from existing evidence; no migration.

## Checklist

### DIFF
--- a/pkg/corroborate/renderer/index.html
+++ b/pkg/corroborate/renderer/index.html
@@ -986,10 +986,10 @@ const STATE_META = {
   CONTESTED:{ cls:"st-contested", ic:"⚠",  label:"CONTESTED",   desc:"sources disagree — inspect the runs" },
   FAILING:  { cls:"st-failing",   ic:"✗",  label:"FAILING",     desc:"corroborated as failing" },
   UNTESTED: { cls:"st-untested",  ic:"–",  label:"UNTESTED",    desc:"no evidence yet" },
-  // PARTIAL is an overview-only recipe rollup: corroborated on the phases that
-  // ran, but not all phases are covered (e.g. a runner without the hardware for
-  // performance). Never a per-row or per-phase state — see rollupRecipe.
-  PARTIAL:  { cls:"st-partial",   ic:"◐",  label:"PARTIAL",     desc:"passing where tested, but not all phases have run" },
+  // PARTIAL is a rollup-only state (phase or recipe), never a per-row state:
+  // corroborated where it ran, but some rows/phases have not been run yet (e.g.
+  // a runner without the hardware for performance). See rollupPhase/rollupRecipe.
+  PARTIAL:  { cls:"st-partial",   ic:"◐",  label:"PARTIAL",     desc:"passing where tested, but not everything has run" },
 };
 const STATE_COLOR = {
   CONFIRMED:"var(--confirmed)", SINGLE:"var(--single)", CONTESTED:"var(--contested)",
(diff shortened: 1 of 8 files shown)


mchmarny (dismissed):
The time-series fix still hides and mislabels historical-only phase rows. Please address the inline blocker before merge.

pkg/corroborate/renderer/index.html
```diff
@@ -2187,7 +2224,8 @@ <h3>No matching recipes for <span style="color:var(--green)">${esc(currentGroup)
     const tbody = document.createElement("tbody");
     PHASES.forEach(ph => {
       const phaseTests = tests.filter(t => t.phase === ph);
-      if (phaseTests.length === 0) return;
+      // Show an empty phase as an UNTESTED coverage gap instead of hiding it.
+      const empty = phaseTests.length === 0;
```
mchmarny, line 2237: `recipe.tests` is the latest combined grid, but `series` was built from the full historical row union. If an older build ran `conformance/foo` and every signer omits that phase in its latest build, the series still contains the historical result while this path renders “not run” and hides the row. Serialize the phase-aware all-build union with `Series`, render the drilldown from it, and cover an older-build-only phase fixture.

mchmarny (changes requested):
Historical-only phase rows are still hidden and mislabeled in the time-series drilldown. Please address the inline blocker before merge.

pkg/corroborate/renderer/index.html
njhensley: Good catch — fixed in 8e4abc8b.

- Added a phase-aware all-build row union to the series output (`Series.Rows` — `{name, phase}`, ordered PhaseOrder then name), populated in `buildSeries` from the same `unionRowsAllBuilds` row keys that feed the build columns.
- The drilldown now renders its rows from `series.rows` instead of `recipe.tests`, so an older-build-only phase (e.g. `conformance/foo` that no signer ran in its latest build) is shown with its historical result rather than dropped. Falls back to `recipe.tests` for series data written before the field existed.
- "Not reported by this source" is now derived from the source's own build columns rather than the latest-grid `runs`, so it's correct for historical-only rows.
- Added an older-build-only `conformance` fixture on nvidia's older run plus two tests: one asserting `Series.Rows` carries it in the `conformance` phase with the historical `pass`/newest `not-run` cells, and one confirming it's absent from the combined all-versions grid but present in the older version's strict grid.

mchmarny (approved):

PR #1847, opened 2026-07-21
fix(uat): harness correctness bugs + budget-aware nightly time-box

### LINKED ISSUE #1770: UAT harness correctness: 4 latent bugs in tests/uat runners + debug collector
## Summary

Four latent correctness bugs in the UAT harness (`tests/uat/**`), all **pre-existing on `main`**. They were surfaced while extracting the shared `tests/uat/lib/phases.sh` in #1769 (CodeRabbit reviewed the moved code as new); (a) was found separately while functionally testing that refactor. Filing here to fix as one focused "UAT harness correctness" PR **after #1769 merges**, so the refactor stays a pure, behavior-preserving change and each fix is reviewed on its own merits.

None of these are introduced by #1769 — the flagged lines are byte-identical to the current runners.

### (a) Debug collector aborts early under `set -e` → truncated bundle — **High**
[`tests/uat/lib/collect-debug.sh#L192`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/tests/uat/lib/collect-debug.sh#L192) (and `capture_skyhook_snapshot`, [L116](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/tests/uat/lib/collect-debug.sh#L116))

The runner sources these under `set -euo pipefail`. The MANIFEST block ends with:
```bash
for snap in "${CLUSTER_DEBUG_DIR}"/skyhook-at-failure-*.txt; do
  [[ -f "${snap}" ]] && echo "skyhookAtFailure: ..."
done
} | tee "${CLUSTER_DEBUG_DIR}/MANIFEST.yaml"
```
When the glob matches nothing, the loop's last command returns 1 → the `{ … } | tee` pipeline returns 1 (under `pipefail`) → `set -e` aborts `collect_cluster_debug` **before any kubectl runs**. For failures where no inline `skyhook-at-failure-*.txt` was captured (e.g. an **install-phase** failure), the uploaded bundle is truncated to a partial MANIFEST with none of the actual cluster state. The workflow's `|| true` masks the non-zero exit, so it looks like it "worked". Confirmed reproducible against the merged code.

**Fix:** make the best-effort collectors immune to the caller's errexit — `local -; set +e` at the top of both `collect_cluster_debug` and `capture_skyhook_snapshot` (bash-4.4 `local -` scopes the option change to the function). Undercuts the #1764 debug bundle, so highest priority.

### (b) Platform-CRD sanity check runs *after* evidence is published — **Major**
[`tests/uat/aws/run#L428`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/tests/uat/aws/run#L428) (the check) runs after [`#L394`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/tests/uat/aws/run#L394) (`validate --phase all` emits + pushes the signed bundle).

If `platform` is misdeclared, the CRD cross-check fails **only after** incorrectly-routed evidence has already been pushed to the wrong TestGrid tab. Move the CRD sanity check before `validate --phase all`.

### (c) Serve readiness passes with only one pod — **Major**
[`tests/uat/aws/run#L648`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/tests/uat/aws/run#L648): `if [[ "${total}" -gt 0 && "${total}" == "${ready_count}" ]]`

The manifest declares two components (Frontend + decode Worker). The condition passes as soon as the Frontend pod exists and is Ready, before the Worker is created — an immediate, flaky inference failure. Gate on `-ge 2` (or component-aware).

### (d) `phase_verify` masks failure details under `set -e` — **Minor**
[`tests/uat/aws/run#L724`](https://github.com/NVIDIA/aicr/blob/51232cd899cd10e2eb10b89f8088823ccb533550/tests/uat/aws/run#L724)

`"${AICR_BIN}" "${args[@]}"` (evidence verify) can exit non-zero before the following `jq`/`cat` runs, so under `set -e` the script aborts before printing `evidence-result.json` and never closes the `::group::`. Wrap in `if ! …; then` to print the result on failure, then propagate the status.

## Notes
- (b)–(d) apply identically to `tests/uat/gcp/run` and `tests/uat/azure/run` (same code); after #1769 merges they live once in `tests/uat/lib/phases.sh`, so the fix is single-site.
- Surfaced by: #1769 (refactor) · Related feature: #1764 (debug bundle).

### LINKED ISSUE #1778: UAT nightly: fix time-box vs drive job-timeout mismatch
`deadline_offset_hours=5` (300m) but `drive timeout-minutes=350`. A cell dispatched at ~minute 299 runs ~100m → ~399m > 350m, so the job is killed mid-cell and the leg hard-fails instead of gracefully dropping oldest cells.

**Evidence:** aws-h100 leg cancelled at 350m on run 29307281940.

**Fix:** lower the deadline so `deadline + max_cell_duration < 350`, or let per-cell time reductions shrink the overrun so the graceful drop-oldest fires first.

njhensley:
## Summary

Fixes four pre-existing correctness bugs in the shared UAT harness (`tests/uat/lib/**`, now single-site after #1769) and the nightly-batch time-box vs drive-job-timeout mismatch.

## Motivation / Context

Two focused correctness fixes filed separately after the #1769 refactor merged, so each is reviewed on its own merits (not bundled into a behavior-preserving refactor).

Fixes: #1770, #1778
Related: #1769 (refactor), #1764 (debug bundle)

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Docs/examples (`docs/`)
- [x] Other: UAT harness (`tests/uat/lib/**`), nightly-batch CI workflow

## Implementation Notes

**#1770 — four harness bugs (all fixed once in `tests/uat/lib/`, so gcp/azure inherit via the thin shims):**

- **(a) High — debug collector aborts early under `set -e`.** `collect_cluster_debug` and `capture_skyhook_snapshot` now run their bodies in a **subshell with `set +e`**, so the caller's errexit can't abort the best-effort collector mid-bundle (the MANIFEST `for snap in glob` loop returns 1 under `pipefail` → truncated bundle). The issue suggested `local -; set +e`, but that's fatal on bash 3.2 (which macOS ships and `collect-debug.sh` explicitly claims to support); the subshell isolates errexit portably on every bash version. Verified: a full bundle is now produced under `set -e` with an empty skyhook glob, where the old code aborted at the MANIFEST `| tee`.
- **(b) Major — platform-CRD check ran after evidence was pushed.** Moved the platform coordinate sanity check in `phase_conformance` to **before** `validate --phase all`, so a mis-declared platform fails the leg before any incorrectly-routed evidence reaches the wrong TestGrid tab.
- **(c) Major — serve readiness passed with only one pod.** `phase_serve` now derives the expected pod count from the DynamoGraphDeployment component replicas (fallback min 2) and gates on `total -ge expected_pods`, closing the Frontend-Ready-before-Worker-exists race.
- **(d) Minor — `phase_verify` masked failure under `set -e`.** Captures the `evidence verify` exit code so the script prints `evidence-result.json` and closes the `::group::` on failure before propagating.

**#1778 — budget-aware nightly time-box:** a fixed `deadline_offset_hours` can't be safe against variable cell durations. The drop-oldest cutoff is now **budget-aware** — the drive loop stops dispatching once fewer than `max_cell_minutes` (new input, default 150) remain before the drive job's own `timeout-minutes` (with a small setup slack for the checkout/setup skew). The effective cutoff is `min(user time-box, budget deadline)`, so the last cell always finishes cleanly instead of being killed mid-run. Adds input validation and a fail-closed guard against a reserve so large no cell could ever dispatch.

## Testing

```bash
# Shell: bash -n + shellcheck (all clean; only pre-existing SC1091 source-resolution info)
# Workflow: yamllint + actionlint (clean)
# Go: tests/uat/workflow_security_test.go passes
```

- Bug (a) reproduced against the merged code and verified fixed on bash 3.2 and via a fake-kubectl full-collection run under `set -e`.
- #1778 deadline arithmetic unit-tested in isolation (budget cutoff, offset-dominates, and rejection of `max_cell_minutes` = 0 / non-integer / too-large).
- No Go source changed, so the Go lint/coverage gates do not apply.

## Risk Assessment

- [x] **Low** — Isolated, best-effort/CI-only harness + workflow changes; easy to revert.

**Rollout notes:** N/A — no artifact or API surface changes. `max_cell_minutes` is a new optional workflow input with a safe default.

## Checklist

- [x] Linter passes (shellcheck / yamllint / actionlint)
- [x] I updated docs if user-facing behavior changed (`docs/contributor/uat.md`)


tests/uat/lib/phases.sh
```diff
@@ -652,6 +656,21 @@ EOF
   echo "::endgroup::"
 
   echo "::group::Wait for DynamoGraphDeployment readiness (timeout ${SERVE_READY_TIMEOUT_SECONDS}s)"
+  # Expected pod count = sum of replicas across the graph's components (Frontend +
+  # decode Worker = 2 here). Gating on "all pods Ready" alone races: the Frontend
+  # pod is created and goes Ready seconds before the Dynamo operator materializes
+  # the Worker, so `total == ready_count` is briefly true with only the Frontend
+  # present — the readiness gate would pass, then serve traffic with no worker.
```
mchmarny, line 663: The fallback two lines down is unreachable for the failure case it documents. The run scripts execute this under `set -euo pipefail`, so if `kubectl get` fails here (transient API error, CRD registration lag), pipefail makes the pipeline non-zero, the assignment propagates it, and errexit kills the script before the `expected_pods=2` fallback ever runs — with both stderr streams sent to `/dev/null`, the leg dies with no message and the `::group::` unclosed. The comment says "a transient kubectl/jq hiccup can never relax the gate," but what actually happens is the hiccup hard-fails the leg.

The loop below already handles this exact case with `|| echo '{}'` on its `kubectl get pods` (line 673). Same treatment here keeps the fallback reachable:

```bash
expected_pods="$(kubectl get dynamographdeployments.nvidia.com "${SERVE_NAME}" -n "${SERVE_NAMESPACE}" \
  -o json 2>/dev/null | jq '[.spec.components[]?.replicas // 1] | add // 0' 2>/dev/null)" || expected_pods=""
```

Failure direction is still closed either way (abort = leg fails), so this isn't dangerous — but it turns a transient hiccup into a hard failure of a multi-hour leg, the opposite of the stated intent.

.github/workflows/uat-nightly-batch.yaml
```diff
@@ -145,9 +151,22 @@ jobs:
...
         run: |
           set -euo pipefail
+          # Approximate this drive job's start (its timeout-minutes clock anchor).
+          # Captured at the top of the first substantive step; the checkout/setup
+          # steps before it already consumed a few minutes of the timeout budget,
+          # so the real GitHub kill is slightly EARLIER than DRIVE_START + timeout.
+          # SETUP_SLACK_MINUTES below absorbs that skew (plus gh dispatch/watch
+          # latency) so the budget-aware cutoff holds even for a full-length cell.
+          DRIVE_START=$(date +%s)
+          SETUP_SLACK_MINUTES=5
```
mchmarny: nit: `DRIVE_START` + 5m slack is an approximation of the job's `timeout-minutes` anchor, and this job's setup is on the heavier side (`fetch-depth: 0` checkout + setup-go). If setup ever exceeds the slack, the budget silently over-reserves in the wrong direction and the mid-cell kill this PR fixes comes back. If that ever bites, the exact anchor is available: `gh api "repos/${REPO}/actions/runs/${GITHUB_RUN_ID}/jobs" --jq '...started_at'` gives this job's true start time, leaving the slack to cover only dispatch/watch latency. Fine as-is with the documented assumption — flagging so the option is on the record.

mchmarny (commented):
Solid set of fixes — all four harness bugs are real and the fixes verified against the merged code. The subshell + `set +e` approach in `collect-debug.sh` is the right call over `local -` given the bash 3.2 portability claim, and I confirmed the one internal `return 0` → `exit 0` conversion covers every early-exit path in the subshell bodies. The platform-check move correctly lands before `validate --phase all` (which is what pushes evidence), and the budget-aware cutoff arithmetic checks out, including the fail-closed guards.

One medium finding: the new `expected_pods` read in `phase_serve` isn't errexit-safe, so the documented kubectl-hiccup fallback is unreachable — the leg hard-fails instead of falling back to 2. One optional nit on the `DRIVE_START` anchor approximation. Nothing blocking; CI is fully green.

tests/uat/lib/phases.sh
njhensley: Good catch — this is a real bug I introduced. Fixed in de57a32: added `|| expected_pods=""` on the command substitution so a transient `kubectl get`/`jq` failure under `set -euo pipefail` no longer propagates through the assignment and hard-kills the leg before the `expected_pods=2` fallback runs. Also rewrote the comment, which previously mis-described the behavior (it claimed a hiccup 'can never relax the gate' when it actually aborted the leg). The gate still only ever tightens.

tests/uat/lib/phases.sh
```diff
@@ -648,6 +652,21 @@ EOF
...
+  # Require at least the declared component count of pods before certifying ready.
+  local expected_pods
+  expected_pods="$(kubectl get dynamographdeployments.nvidia.com "${SERVE_NAME}" -n "${SERVE_NAMESPACE}" \
+    -o json 2>/dev/null | jq '[.spec.components[]?.replicas // 1] | add // 0' 2>/dev/null)"
+  # Fall back to the known-minimum (Frontend + Worker) if the spec read fails, so
+  # a transient kubectl/jq hiccup can never relax the gate below two components.
+  if [[ -z "${expected_pods}" || "${expected_pods}" -lt 2 ]]; then
+    expected_pods=2
+  fi
+  echo "expecting ${expected_pods} workload pod(s) (sum of component replicas)"
```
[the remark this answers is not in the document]
njhensley: Fixed in de57a32 — guarded the command substitution with `|| expected_pods=""` so the `set -e`/`pipefail` abort can't fire before the `expected_pods=2` fallback, same failure-tolerant shape as the `kubectl get pods` read in the loop below.

.github/workflows/uat-nightly-batch.yaml
njhensley: Thanks — agreed the fixed slack is a stand-in for the true anchor. Kept the `DRIVE_START` approach but bumped `SETUP_SLACK_MINUTES` 5→10m in de57a32 to comfortably clear the `fetch-depth: 0` + setup-go setup, and recorded your exact-anchor recipe (`.../jobs` `started_at`) inline as the escape hatch if setup ever outgrows the slack.

mchmarny (approved):

Issue #1820, opened 2026-07-20
nodewright: revert applied nvidia-tuned host state when tuning is disabled


yuanchen8911:
## Problem

PR #1756 defaults AKS recipes to the Azure-managed GPU driver profile and gates the `nvidia-tuned` Skyhook package behind `nodewright-customizations` `tuningEnabled: false`. That gate only controls whether the package **renders** into the Skyhook CR — it does not revert tuning already applied to nodes.

**Residue mechanism:**

- Skyhook package uninstall defaults off, and the `tuning.yaml` customization declares no `uninstall` block for `nvidia-tuned`.
- When an upgraded bundle omits `nvidia-tuned` from the Skyhook CR, the operator simply stops managing the package; the host state it applied (isolcpus, hugepages, GRUB cmdline changes, sysctl/tuned profile) remains in effect until the node is reimaged or recreated.

**Affected upgrade path:** AKS clusters deployed before #1756 (where `nvidia-tuned` 0.3.2 was applied, including the `aks-h100-inference` profile shown to regress inference TTFT from ~550 ms to ~6 s) that upgrade in place to a `tuningEnabled: false` bundle. They keep the tuned host state and do not see the untuned-default behavior the new recipes intend.

## Current disposition

`docs/integrator/aks-gpu-setup.md` documents the limitation: in-place upgrade does not revert previously applied tuning; recreating/reimaging the GPU node pool is the clean path.

## Candidate approaches

- Add an `uninstall` block to the `nvidia-tuned` package in the tuning customization so nodewright reverts the applied profile when the package is removed (needs upstream nodewright-packages support/verification, and implies a revert-triggered reboot).
- Keep documentation-only: reimage guidance in `aks-gpu-setup.md` (status quo, this issue tracks improving on it).
- Automate node-pool recreation guidance/tooling for the AKS profile flip.

## Related

- PR #1756 (Azure-managed driver default, `tuningEnabled` gate)

## Related follow-up: Sprig-divergence render-test gap

[...]

PR #1849, opened 2026-07-22
fix(validator): make nodewright deployment health check value-aware

### LINKED ISSUE #1844: nodewright-customizations deployment health check fails when tuningEnabled=false suppresses the Skyhook CR
## Summary

The `nodewright-customizations` deployment health check (`recipes/checks/nodewright-customizations/health-check.yaml`) unconditionally asserts the `tuning` Skyhook CR reaches `status.status: complete`. When `tuningEnabled=false` suppresses the CR on the single-package tuning manifests (`tuning-gke.yaml`, `tuning-generic.yaml`), `aicr validate --phase deployment` fails on the deliberately-untuned cluster because the CR is absent. The health check cannot see the `tuningEnabled` value gate.

## Reachability

Only when a user explicitly passes `--set nodewrightcustomizations:tuningEnabled=false` on a GKE-COS or generic recipe. No shipped recipe hits it: AKS (the only place the flag is a supported default) uses the multi-package `tuning.yaml`, which keeps the CR via `nvidia-setup`, so its health check passes; non-AKS recipes default the flag to true. So the check is correct for every default configuration and fails only on this off-label override. Documented as a limitation in `docs/integrator/components/nodewright.md` (PR #1756).

## Why not just tolerate an absent CR

Making the assertion pass when the CR is missing is a fail-open anti-pattern (the repo bans spuriously-passing negative checks): a check that tolerates absence would also pass when tuning was *enabled* but the Skyhook CR failed to appear — hiding a real deployment failure. So the check must stay fail-closed.

## Proposed fix

Make the health check value-aware: when the resolved/effective `nodewrightcustomizations:tuningEnabled` is `false` (and the recipe wires a single-package tuning manifest), skip or invert the `validate-skyhook-cr-complete` assertion (assert the CR is *absent*), rather than requiring `complete`. This needs the health check (or its hydration) to observe the effective value — the same "value gates are invisible to generated/validation artifacts" gap as #1820. Options to explore: drive `healthCheckSkip` on the component from the same value, or thread the effective value into the check's hydration.

Related: #1820 (in-place tuning residue), #1842 (Flux hook orphan — now fixed in #1756 by making the CR release-managed). Introduced by the `tuningEnabled` gate in #1756.

njhensley:
## Summary

Make the `nodewright-customizations` deployment-phase health checks value-aware: render each nodewright manifest with the component's effective Helm values and derive the expected `tuning` Skyhook CR set from the render, so a value-gated CR (e.g. `tuningEnabled=false` suppressing the whole CR on a single-package tuning manifest) is not asserted on a deliberately-untuned cluster.

## Motivation / Context

The deployment phase asserted the `tuning` Skyhook CR reaches `status.status: complete` in two places that both ignore value gates: the static chainsaw assert (`recipes/checks/nodewright-customizations/health-check.yaml`) and the Go readiness check `verifyNodewrightReady`, whose `expectedNodewrightNames` regex-scanned the **raw** manifest template. When the effective values suppress the whole CR, both failed on a cluster that was deliberately untuned.

Fixes: #1844
Related: #1756 (introduces the `tuningEnabled` gate this tolerates), #1820 (value gates invisible to generated/validation artifacts)

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

- New `recipe.GetComponentValues(ref)` resolves a bare `ComponentRef`'s effective values (base `values.yaml` → `valuesFile` → `overrides`) via the embedded provider, mirroring `GetManifestContent`; `GetValuesForComponentWithContext` now delegates to the shared `resolveComponentValues`.
- `expectedNodewrightNames` renders each manifest with those values (`manifest.Render`, the same renderer the bundler uses) before extracting Skyhook names — a value-gated CR drops out of the render.
- `verifyNodewrightReady` now distinguishes **no manifests declared** (still fails closed — misconfiguration) from **manifests present but rendered empty** (CR deliberately suppressed → pass).
- `nodewrightHealthCheckSuppressed` skips the redundant static chainsaw assert on the same render signal, scoped to `nodewright-customizations` only.
- **Fail-closed preserved:** whenever the effective values keep the CR, the render still lists it and the check runs, so a CR that *should* deploy but is missing on the cluster still fails. Only an intentionally-suppressed CR is tolerated. Render/read errors propagate rather than passing.

**Base-on-`main` note:** `#1756` (which adds `tuningEnabled`) is unmerged. The render mechanism is generic — today's manifests gate only on `enabled`, so the render produces an identical CR set to the old regex for every shipped recipe (verified by the existing suite). It becomes correct for `tuningEnabled=false` automatically once `#1756` lands, with no further validator change.

**Boundary (documented):** suppression must be expressed in the recipe (overlay-declared or inline `overrides:`, e.g. the AKS `tuningEnabled: false` default). The in-cluster validator Job renders from the recipe it is handed and has no `DataProvider`, so a bundle-time-only `--set` (never persisted to the recipe; `aicr validate` has no `--set`) is not honored, and `--data`-external files referenced by path are not readable in the Job (fails closed).

## Testing

```bash
go test -race ./validators/deployment/... ./pkg/recipe/... ./pkg/manifest/...
golangci-lint run -c .golangci.yaml ./validators/deployment/... ./pkg/recipe/...
```

- All affected packages pass with `-race`; downstream consumers (`pkg/bundler/...`, `pkg/validator/...`, `pkg/cli/...`) pass. `go build ./...` clean. `golangci-lint` (v2.12.2, matches `.settings.yaml`): 0 issues. `yamllint` clean.
- New tests: `TestExpectedNodewrightNames_RenderAware`, `TestNodewrightHealthCheckSuppressed`, `TestVerifyNodewrightReady_TolerantWhenAllCRsSuppressed`, `TestGetComponentValues_BareRef`. Coverage on every new/changed function ≥ 80%.

## Risk Assessment

**Rollout notes:** No behavior change for any shipped recipe (render == prior regex output while manifests gate only on `enabled`). UAT deployment lanes unaffected — nodewright tuning stays enabled, so `verifyNodewrightReady` (status + runtime-required taint gate + stability window) runs exactly as before.

## Checklist


mchmarny (approved):
No blocking issues found. The value-aware render path preserves fail-closed behavior, and the focused race/lint validation passes. The failing KWOK git-deployer lanes are unrelated Gitea image-pull failures.

Issue #1372, opened 2026-06-15
feat(recipes): add kubeflow training overlay for rtx-pro-6000 on EKS


yuanchen8911:
## Summary

There is no Kubeflow training recipe for **`rtx-pro-6000` on EKS**. Generating `--service eks --accelerator rtx-pro-6000 --intent training --os ubuntu --platform kubeflow` resolves to a **base training stack only** — `kubeflow-trainer`/`kubeflow-trainer-post` are absent and `platform: kubeflow` is effectively a no-op.

## Details

Applied overlays for that criteria are `base, monitoring-hpa, rtx-pro-6000-any, eks, eks-training` — no kubeflow platform wiring and no NCCL performance gate. Concretely:

- rtx-pro-6000 training overlays exist **only for LKE** (`rtx-pro-6000-lke-training.yaml`, `rtx-pro-6000-lke-ubuntu-training.yaml`); there is **no `rtx-pro-6000-eks-*-training`** overlay.
- No kubeflow overlay/mixin references rtx-pro-6000 at all (`grep kubeflow recipes/overlays/*rtx*` → none).
- Resulting recipe has 13 base components, empty conformance training checks, and no performance (NCCL) validator.

By contrast, GB200-EKS and H100-GKE ship real kubeflow training overlays (`kubeflow-trainer` + `kubeflow-trainer-post`, NCCL all-reduce perf gate), so training validates end-to-end there.

## Proposed change

Add an `rtx-pro-6000-eks-ubuntu-training-kubeflow` overlay (and any needed mixin wiring) mirroring the existing GB200-EKS / LKE training overlays:
- include `kubeflow-trainer` + `kubeflow-trainer-post` (torch-distributed runtime),
- wire an appropriate NCCL all-reduce performance validator + threshold for rtx-pro-6000,
- carry the standard eks training node-scheduling.

## Impact

Inference on rtx-pro-6000+EKS is unaffected (validated end-to-end: deployment 4/4, conformance 11/11, inference-perf ~64K tok/s on the b40 cluster). This is purely a training-recipe coverage gap.

## Type of Change

- New feature (recipe coverage)

mchmarny:
Triaged: added to the **v1** milestone (Sufficient Coverage).

This is an (accelerator × platform) cell of the v1 matrix — RTX PRO 6000 on EKS with Kubeflow.

yuanchen8911:
[...]

PR #1857, opened 2026-07-22
feat(recipes): add kubeflow training overlay for rtx-pro-6000 on EKS

mohityadav8:
Adds rtx-pro-6000-eks-training / -ubuntu-training / -ubuntu-training-kubeflow so training on EKS resolves kubeflow-trainer + kubeflow-trainer-post, matching the GB200-EKS / rtx-pro-6000-LKE pattern.

NCCL all-reduce gating is intentionally descoped from this PR pending calibration. Follow-up work is tracked in #<new-issue-number>.

Related to #1372


mohityadav8:
@coderabbitai full review

recipes/overlays/rtx-pro-6000-eks-ubuntu-training.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - os-ubuntu
+
+  # Constraints for RTX PRO 6000 on EKS with Ubuntu for training workloads
+  constraints:
+    - name: K8s.server.version
+      value: ">= 1.32.4"
+
+  componentRefs: []
+
+  # Validation is inherited from rtx-pro-6000-eks-training (not OS-specific)
```
yuanchen8911: The Merge Gate lint job is failing on this file: `42:3 [comments-indentation] comment not indented like content` (`make lint-yaml` exit 1). The trigger is that the file ends with a comment line that isn't newline-terminated — yamllint's comments-indentation rule misfires on an unterminated trailing comment. Adding a trailing newline makes lint pass (verified locally with the repo's `.yamllint.yaml`).

Consider adding trailing newlines to the other two new files as well for hygiene — they pass lint as-is since they don't end in a comment, and the repo config disables `new-line-at-end-of-file`, so that part is optional.

recipes/overlays/rtx-pro-6000-eks-training.yaml
```diff
@@ -0,0 +1,101 @@
...
+  # so there is neither a calibrated bandwidth floor nor a runtime testdata
+  # template for it today. Forcing the generic nccl-all-reduce-bw check here
+  # would either no-op or false-fail in the same way tracked for single-GPU-
+  # per-node H100 shapes in NVIDIA/aicr#1256.
+  validation:
+    conformance:
+      checks:
+        - platform-health
+        - gpu-operator-health
+        - dra-support
```
yuanchen8911: Two problems with this block as written.

First, the "single-GPU workstation card with no NVLink" rationale doesn't hold for the EKS target: `rtx-pro-6000-eks-inference.yaml` documents the AWS g7e family as 8x RTX PRO 6000 per node, and the dynamo inference overlay benchmarks g7e.48xlarge (8 GPUs). The card has no NVLink, but multi-GPU nodes with an inter-node fabric are exactly what the NCCL all-reduce validator tests, so NVLink absence alone doesn't make the check inapplicable. This rationale was carried over from the LKE overlay, where it may fit; here it doesn't.

Second, the actual limitation is real but different: there is no rtx-pro-6000 entry in `supportedNCCLCombinations` (any variant), no testdata runtime template, and no calibrated bandwidth floor — so the check genuinely can't be enabled today without false-failing. That makes the omission correct, but it means the PR doesn't deliver the NCCL validator + threshold that #1372's proposed change explicitly requests, so the `Fixes #1372` claim overstates.

Suggested resolution:
1. Rewrite this comment to state the real limitation (no matrix entry / runtime template / calibrated threshold for rtx-pro-6000 on EKS), dropping the single-GPU/NVLink framing.
2. Change `Fixes #1372` to `Related to #1372` in the PR description (or descope the NCCL bullet in the issue and keep `Fixes`).
3. File a follow-up issue for hardware calibration and NCCL matrix/testdata support for rtx-pro-6000 on EKS.

To be clear, this is not a request to implement uncalibrated NCCL support in this PR — the omission itself is the right call today.

yuanchen8911 (changes requested):
Thanks for the PR! Three findings: one CI blocker (trivial trailing newline), one scope/documentation correction on the NCCL omission rationale, and one non-blocking test-coverage gap. Otherwise this is pattern-consistent with the GB200-EKS / LKE training overlays and safe to merge once addressed. Details inline; the test-coverage item is below since it touches files outside the diff.

Also, the branch is currently 80 commits behind `main` — please rebase onto `origin/main` (rather than a merge commit or the "Update branch" button) when you push the fixes, so CI runs against current main.

**Test-coverage gap (non-blocking):** the topology-updater regression table's RTX PRO 6000 EKS section ([pkg/recipe/metadata_test.go#L2322-L2326](https://github.com/NVIDIA/aicr/blob/868349db11fe14e94b990fe218ea93b65bc7258f/pkg/recipe/metadata_test.go#L2322-L2326)) contains only the inference variants. The three new training leaves all resolve with `topologyUpdater: true` but aren't covered, so a future regression dropping it would go uncaught. The conformance table ([pkg/recipe/conformance_test.go#L304](https://github.com/NVIDIA/aicr/blob/868349db11fe14e94b990fe218ea93b65bc7258f/pkg/recipe/conformance_test.go#L304)) likewise doesn't cover the new training chain. The sibling inference overlays were added to both tables when introduced — please add the three training criteria (`rtx-pro-6000-eks-training`, `-ubuntu-training`, `-ubuntu-training-kubeflow`) for parity. Coverage debt only; no present runtime failure.

PR #1864, opened 2026-07-22
feat(slinky/slurm/snapshot): add snapshotting for Slinky Slurm and MariaDB operator

kaynetu:
## Summary

Adds snapshot signals for declared Slinky Slurm topology and the official MariaDB Operator API so future deployment constraints can detect existing resources before installation. The capture is intentionally secret-safe and observational: it does not reconstruct Helm values, infer `platform: slurm`, or claim Slurm, operator, or database health.

## Motivation / Context

AICR needs conclusive, pre-deployment evidence of an existing Slinky Controller and potential MariaDB Operator conflict without capturing workload configuration or credentials. These signals provide a stable input for future fail-closed constraint checks, while preserving compatibility with older snapshots that do not include either subtype.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: evidence redaction policy and snapshot-agent RBAC

## Implementation Notes

- Discovers the exact `slinky.slurm.net/controllers` API dynamically and reports `absent`, `detected`, `unsupported-multicluster`, or `unknown`. When exactly one Controller is observed, it projects only associated NodeSet, LoginSet, RestApi, and referenced Accounting identity plus an explicit scalar allowlist.
- Emits child items and counts only when every required child API and reference is collected conclusively. Incomplete projection retains the Controller item but omits child counts and children, preventing ambiguous zero values from becoming a passing constraint input.
- Detects the official `k8s.mariadb.com/mariadbs` API as conflict evidence with `absent`, `api-detected`, `crs-detected`, or `unknown`. `api-available` consistently means the exact target resource is served; it does not report operator or database availability.
- Adds least-privilege `list` RBAC for the targeted Slinky and MariaDB CRs, pins the Slinky v1.2.0 schema paths in tests, and keeps both new subtypes in raw/`--full` snapshots while dropping them from minimal evidence.
- Documents snapshot field-presence and fail-closed semantics, including hybrid snapshot recipe generation where Slurm remains an explicit platform choice.

**Observed snapshot evidence** (`/private/tmp/aws-snapshot-new-3.yaml`):

```yaml
- data:
    accounting-count: 0
    api-available: true
    api-version: v1beta1
    collection-state: detected
    controller-count: 1
    detected: true
    loginset-count: 1
    nodeset-count: 1
    restapi-count: 1
  items:
    - context:
        api-version: v1beta1
        id: controller/slurm/slinky-slurm
        kind: Controller
        name: slinky-slurm
        namespace: slurm
      data:
        accounting-ref-present: false
        external: false
    - context:
        api-version: v1beta1
        controller-id: controller/slurm/slinky-slurm
        id: loginset/slurm/slinky-slurm-login-slinky
        kind: LoginSet
        name: slinky-slurm-login-slinky
        namespace: slurm
    - context:
        api-version: v1beta1
        controller-id: controller/slurm/slinky-slurm
        id: nodeset/slurm/slinky-slurm-worker-slinky
        kind: NodeSet
        name: slinky-slurm-worker-slinky
        namespace: slurm
      data:
        partition-enabled: true
    - context:
        api-version: v1beta1
        controller-id: controller/slurm/slinky-slurm
        id: restapi/slurm/slinky-slurm
        kind: RestApi
        name: slinky-slurm
        namespace: slurm
  subtype: slinky-slurm
- data:
    api-available: false
    collection-state: absent
  subtype: mariadb-operator
```

The observed Controller has no Accounting reference (`accounting-count: 0`), while the exact Slinky Controller API is served. The MariaDB result means the official `mariadbs.k8s.mariadb.com` API is absent; it does not make a claim about an external database.

## Testing

```bash
make qualify
```

`make qualify` passed locally, including the race-enabled Go suite, coverage gate (79.7%; 75% threshold), lint, 23/23 E2E checks, vulnerability scan, license check, and repository checks.

## Risk Assessment

**Rollout notes:** Updated snapshot-agent RBAC is required to collect the new CRs. Existing snapshots remain valid; future constraints must treat `unknown` states and omitted child fields as non-passing rather than as absence. No deployment behavior or existing recipe resolution is changed.

## Checklist


pkg/k8s/agent/rbac.go
```diff
@@ -181,6 +181,22 @@ func (d *Deployer) ensureClusterRole(ctx context.Context) error {
...
+				slinkyRestAPIResource,
+				slinkyAccountingResource,
+			},
+			Verbs: []string{verbList},
+		},
+		{
+			APIGroups: []string{mariaDBAPIGroup},
+			Resources: []string{mariaDBResource},
+			Verbs:     []string{verbList},
+		},
```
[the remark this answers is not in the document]
kaynetu: Thanks for flagging this. We’re keeping the existing behavior because this PR only adds RBAC rules; it does not introduce the `AlreadyExists → Update` pattern.

Kubernetes RBAC resources explicitly allow updates without a `ResourceVersion`; the API server performs an unconditional update using the latest version. Adding `Get → Update` would introduce possible `409 Conflict` failures without retry handling.

If we want optimistic concurrency, we should handle it separately with `RetryOnConflict` or server-side apply.

pkg/collector/k8s/slinky.go
```diff
@@ -0,0 +1,757 @@
...
+		slog.Warn("Slinky Slurm discovery cancelled", slog.String("error", err.Error()))
+		return unknownSlinkySubtype()
+	}
+
+	discoveryClient := k.slinkyDiscovery
+	if discoveryClient == nil {
+		discoveryClient = k.ClientSet.Discovery()
+	}
+
+	groups, err := discoveryClient.ServerGroups()
```
mchmarny, line 167: **Blocking:** this call is not bounded by `ctx`. client-go implements both `ServerGroups()` and `ServerResourcesForGroupVersion()` with `Do(context.TODO())`, while the `rest.Config` returned by `BuildKubeClient` leaves `Timeout` at zero. A stalled discovery endpoint can therefore keep this goroutine—and `errgroup.Wait()`/the snapshot—blocked indefinitely past `CollectorK8sTimeout`. Please use discovery with a real bounded request/transport timeout and add a blocking-server regression test covering both discovery calls. The same issue applies to MariaDB’s `ServerGroups()` call and the shared resource-discovery helper.

mchmarny (changes requested):
One merge blocker: Kubernetes discovery is not cancellation- or timeout-bounded, so a stalled API server can hang snapshot collection indefinitely. Focused race tests and lint otherwise pass; the failed LKE KWOK cell is an action-download infrastructure timeout before its test ran.

pkg/collector/k8s/slinky.go
kaynetu: Thanks, confirmed. The underlying client-go discovery behavior predates this PR: the existing ServerVersion() and ServerPreferredResources() paths also use context.TODO() with an unbounded shared REST client. This PR adds ServerGroups() and ServerResourcesForGroupVersion() calls, which expands and amplifies that exposure.

I’m addressing it at the shared K8s collector discovery boundary rather than patching only the new Slinky/MariaDB paths. The fix binds all existing and new discovery requests to the collector context with a bounded timeout. I’m also adding blocking-server regressions for all four discovery APIs and for the full Collect() path to verify errgroup.Wait() cannot hang past the collector deadline.

mchmarny (approved):

PR #1916, opened 2026-07-25
fix: resolve goconst and errorlint warnings across OCI, coverage, validators

andrewwhitecdw:
## Summary

Fixes four small lint/test issues surfaced by golangci-lint and static analysis: one errorlint violation in OCI push tests and three goconst duplications in coverage tooling and validator packages.

## Motivation / Context

N/A — opportunistic cleanup from a lint/test sweep.

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: tools/coverage

## Implementation Notes

- Replaced `err == io.EOF` with `stderrors.Is(err, io.EOF)` in `pkg/oci/push_test.go`.
- Centralized the em dash placeholder as `emDash` in `tools/coverage`.
- Added `allocationModeExactCount` in `validators/conformance/secure_access_check.go`.
- Added `skipMsgNCCLFewNodes` in `validators/performance/nccl_all_reduce_bw_constraint.go` and updated tests.

## Testing

```bash
# Targeted package tests and lint on changed packages:
go test ./pkg/oci/... ./tools/coverage/... ./validators/conformance/... ./validators/performance/...
golangci-lint run -c .golangci.yaml ./pkg/oci/... ./tools/coverage/... ./validators/conformance/... ./validators/performance/...
```

All targeted tests pass and golangci-lint reports 0 issues. Full `make lint` reached the Go lint step successfully; the overall target stopped at the YAML lint step because `yamllint` is not installed in this environment.

## Risk Assessment

**Rollout notes:** N/A

## Checklist

### DIFF
--- a/pkg/oci/push_test.go
+++ b/pkg/oci/push_test.go
@@ -106,7 +106,7 @@ func extractFilesFromOCIArtifact(t *testing.T, ociLayoutDir, digest string) map[
 	tr := tar.NewReader(gzr)
 	for {
 		header, err := tr.Next()
-		if err == io.EOF {
+		if stderrors.Is(err, io.EOF) {
 			break
 		}
 		if err != nil {
--- a/tools/coverage/model.go
+++ b/tools/coverage/model.go
@@ -30,8 +30,6 @@ const (
 	// only the Azure UAT trees). Distinct from not-yet-covered: the assets are
 	// present but inert pending a revive-or-retire decision (DC6, #1280).
 	StatusStubbed Status = "stubbed"
-
-	emDash = "—"
 )
 
 // Harness identifies an execution mechanism that can exercise an item. The set
--- a/tools/coverage/model.go
+++ b/tools/coverage/model.go
@@ -30,6 +30,8 @@ const (
 	// only the Azure UAT trees). Distinct from not-yet-covered: the assets are
 	// present but inert pending a revive-or-retire decision (DC6, #1280).
 	StatusStubbed Status = "stubbed"
+
+	emDash = "—"
 )
 
 // Harness identifies an execution mechanism that can exercise an item. The set
--- a/validators/conformance/secure_access_check.go
+++ b/validators/conformance/secure_access_check.go
@@ -65,7 +65,11 @@ const (
 	gpuClaimPrefix   = "gpu-claim-"
 	// noAllocProbePrefix names the standalone probe pod that is granted no
 	// GPU at all (no ResourceClaims, no nvidia.com/gpu limits).
-	noAllocProbePrefix       = "no-alloc-probe-"
+	noAllocProbePrefix = "no-alloc-probe-"
+
+	// allocationModeExactCount is the Kubernetes DRA DeviceRequest
+	// allocationMode enum value. It is an upstream API constant, not an AICR
+	// internal token. See k8s.io/api/resource/{v1beta1,v1,v1beta2}.DeviceRequest.
 	allocationModeExactCount = "ExactCount"
 
 	// cudaTestImage runs the in-container GPU visibility verification.
--- a/validators/conformance/secure_access_check.go
+++ b/validators/conformance/secure_access_check.go
@@ -65,7 +65,8 @@ const (
 	gpuClaimPrefix   = "gpu-claim-"
 	// noAllocProbePrefix names the standalone probe pod that is granted no
 	// GPU at all (no ResourceClaims, no nvidia.com/gpu limits).
-	noAllocProbePrefix = "no-alloc-probe-"
+	noAllocProbePrefix       = "no-alloc-probe-"
+	allocationModeExactCount = "ExactCount"
 
 	// cudaTestImage runs the in-container GPU visibility verification.
 	cudaTestImage = "nvidia/cuda:12.9.0-base-ubuntu24.04"
@@ -1587,12 +1588,12 @@ func buildResourceClaim(run *gpuTestRun, version string) *unstructured.Unstructu
 	}
 	if version == versionV1beta1 {
 		request["deviceClassName"] = draDriverGPU
-		request["allocationMode"] = "ExactCount"
+		request["allocationMode"] = allocationModeExactCount
 		request["count"] = int64(1)
 	} else {
 		request["exactly"] = map[string]interface{}{
 			"deviceClassName": draDriverGPU,
-			"allocationMode":  "ExactCount",
+			"allocationMode":  allocationModeExactCount,
 			"count":           int64(1),
 		}
 	}
--- a/validators/performance/nccl_all_reduce_bw_constraint.go
+++ b/validators/performance/nccl_all_reduce_bw_constraint.go
@@ -62,9 +62,23 @@ const (
 	// ncclTrainingRuntimeName is the name of the TrainingRuntime resource.
 	// Must stay in sync with runtime.yaml.
 	ncclTrainingRuntimeName = "nccl-all-reduce-runtime"
-
-	skipMsgNCCLFewNodes = "skipped - requires at least 2 GPU nodes for EW fabric test"
 )
+
+// skipMsg* are the constraint-result strings returned when the NCCL check cannot
+// run. The "skipped " prefix is contractual: nccl_all_reduce_bw.go:78 dispatches
+// on it via strings.HasPrefix(actual, "skipped") to emit CTRF Skip status.
+const (
+	skipMsgNCCLNoInput        = "skipped - requires Service + Accelerator"
+	skipMsgNCCLNotImplemented = "skipped - requires Service + Accelerator to be implemented"
+	skipMsgNCCLFewNodes       = "skipped - requires at least 2 GPU nodes for EW fabric test"
+)
+
+// skipMsgNCCLNoProfile returns the skip result when the resolved benchmark
+// profile does not implement the requested NCCL variant.
+func skipMsgNCCLNoProfile(target ncclBenchmarkTarget, variant ncclVariant) string {
+	return fmt.Sprintf("skipped - benchmark profile %s does not implement the %s NCCL variant",
+		target.String(), ncclVariantDisplayName(variant))
+}
 
 // Package-level GVR definitions for Kubeflow Trainer CRDs used by both
 // applyNCCLResources and cleanupNCCLResources.
@@ -252,7 +266,7 @@ func validateNcclAllReduceBw(ctx *validators.Context, constraint recipe.Constrai
 	// Skip unless the validation targets a supported service + accelerator combination.
 	if ctx.ValidationInput == nil {
 		slog.Info("Skipping NCCL All Reduce bandwidth validation: no validation")
-		return "skipped - requires Service + Accelerator", true, nil
+		return skipMsgNCCLNoInput, true, nil
 	}
 
 	service := ctx.ValidationInput.Criteria.Service
@@ -323,10 +337,9 @@ func validateNcclAllReduceBw(ctx *validators.Context, constraint recipe.Constrai
 			// The profile itself is valid (resolveNCCLBenchmarkProfile fails
 			// closed on unknown pairs); it just doesn't implement this variant
 			// — e.g. gb200/eks covers net and nvls but not the default check.
-			return fmt.Sprintf("skipped - benchmark profile %s does not implement the %s NCCL variant",
-				target.String(), constraintNameForVariant(variant)), true, nil
+			return skipMsgNCCLNoProfile(target, variant), true, nil
 		}
-		return "skipped - requires Service + Accelerator to be implemented", true, nil
+		return skipMsgNCCLNotImplemented, true, nil
 	}
 
 	// Extract threshold from constraint
@@ -373,7 +386,7 @@ func validateNcclAllReduceBw(ctx *validators.Context, constraint recipe.Constrai
 	// NCCL all-reduce tests EW (East-West) fabric between nodes and requires at least
 	// two GPU nodes. Skip gracefully rather than fail when only one node is available.
 	if gpuConfig.WorkerCount < 2 {
-		slog.Info("Skipping NCCL All Reduce bandwidth validation: "+skipMsgNCCLFewNodes,
+		slog.Info("Skipping NCCL All Reduce bandwidth validation: requires at least 2 GPU nodes for EW fabric test",
 			"nodes", gpuConfig.WorkerCount)
 		return skipMsgNCCLFewNodes, true, nil
 	}
--- a/validators/performance/nccl_all_reduce_bw_constraint.go
+++ b/validators/performance/nccl_all_reduce_bw_constraint.go
@@ -62,6 +62,8 @@ const (
 	// ncclTrainingRuntimeName is the name of the TrainingRuntime resource.
 	// Must stay in sync with runtime.yaml.
 	ncclTrainingRuntimeName = "nccl-all-reduce-runtime"
+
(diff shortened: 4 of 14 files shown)


mohityadav8 (approved):
Thanks for the PR! I fetched this locally to test the changes. I ran golangci-lint against pkg/oci/, tools/coverage/, and the validator packages. I can confirm the previous goconst and errorlint warnings are completely resolved.
LGTM

validators/performance/nccl_all_reduce_bw_constraint.go
```diff
@@ -371,9 +373,9 @@ func validateNcclAllReduceBw(ctx *validators.Context, constraint recipe.Constrai
 	// NCCL all-reduce tests EW (East-West) fabric between nodes and requires at least
 	// two GPU nodes. Skip gracefully rather than fail when only one node is available.
 	if gpuConfig.WorkerCount < 2 {
-		slog.Info("Skipping NCCL All Reduce bandwidth validation: requires at least 2 GPU nodes for EW fabric test",
+		slog.Info("Skipping NCCL All Reduce bandwidth validation: "+skipMsgNCCLFewNodes,
```
njhensley: **🟡 Minor — Reusing the status constant inside the log message makes it stutter**

This now emits `Skipping NCCL All Reduce bandwidth validation: skipped - requires at least 2 GPU nodes for EW fabric test` — the skip is stated twice, and a machine-readable status prefix ends up mid-sentence in operator prose. That `skipped - ` prefix is load-bearing: `nccl_all_reduce_bw.go:78` does `strings.HasPrefix(actual, "skipped")` → `validators.Skip` → exit code 2 → CTRF skipped status. That is exactly why the two sibling skip paths in this same function (`:254`, `:320`) keep their log text independent of the returned status value; this line is now the only one that collapses them.

**Blast radius:** Operator-facing validator stderr on every single-GPU-node validate run. I grepped `tests/uat/lib/*.sh`, the chainsaw trees, workflows and `docs/` for the old log text and found no consumer, so this is cosmetic only — no functional impact.

**Fix:** Keep the log message as its own literal and return the constant separately:

```go
slog.Info("Skipping NCCL All Reduce bandwidth validation: requires at least 2 GPU nodes for EW fabric test",
    "nodes", gpuConfig.WorkerCount)
return skipMsgNCCLFewNodes, true, nil
```

validators/performance/nccl_benchmark_profile_test.go
```diff
@@ -373,7 +373,7 @@ func TestValidateNcclAllReduceBwProfileClusterPath(t *testing.T) {
 	if err != nil {
 		t.Fatalf("unexpected error: %v", err)
 	}
-	want := "skipped - requires at least 2 GPU nodes for EW fabric test"
+	want := skipMsgNCCLFewNodes
```
njhensley, line 376: **🔵 Nitpick — Tests now assert against the constant, dropping the only guard on the load-bearing `skipped` prefix**

Asserting `want := skipMsgNCCLFewNodes` is normal Go style, but it does remove the literal that incidentally pinned the `skipped ` prefix. Worth flagging that the failure mode is worse than it first looks: if the prefix were ever dropped, `strings.HasPrefix(actual, "skipped")` at `nccl_all_reduce_bw.go:78` falls through to `if !passed` — and `passed` is `true` on this path — so the check would silently report **PASS**, not fail. I checked and no test currently exercises that dispatch, so this is a pre-existing gap the PR merely stops accidentally masking.

**Blast radius:** A future edit to the constant could silently turn a skipped NCCL check into a green one in CTRF/evidence output. Latent, not live.

**Fix:** One cheap guard somewhere in the package tests:

```go
if !strings.HasPrefix(skipMsgNCCLFewNodes, "skipped") {
    t.Fatal("skip status prefix is contractual — nccl_all_reduce_bw.go dispatches on it")
}
```

validators/conformance/secure_access_check.go
```diff
@@ -65,7 +65,8 @@ const (
 	gpuClaimPrefix   = "gpu-claim-"
 	// noAllocProbePrefix names the standalone probe pod that is granted no
 	// GPU at all (no ResourceClaims, no nvidia.com/gpu limits).
-	noAllocProbePrefix = "no-alloc-probe-"
+	noAllocProbePrefix       = "no-alloc-probe-"
+	allocationModeExactCount = "ExactCount"
```
njhensley, line 73: **🔵 Nitpick — `allocationModeExactCount` is misfiled in the name-prefix group, undocumented, and only half-abstracts the DRA literals**

A DRA API enum is wedged into an otherwise homogeneous group of Kubernetes object *name prefixes*, directly beneath `noAllocProbePrefix`'s two-line doc comment — which now visually reads as if it documents both. Because the new name is longer, gofmt re-padded `noAllocProbePrefix`, so the diff touches an unrelated declaration. The abstraction is also half-applied: the sibling literals in the very same expression (`"exactly"`, `"deviceClassName"`, `"count"`, and the `"allocationMode"` key itself) were left inline at the same 2-occurrence count.

**Blast radius:** Readability and diff/blame noise in one file. I verified the emitted ResourceClaim is byte-identical for both the v1beta1 and v1 `exactly` shapes — no wire change.

**Fix:** Give it its own blank-line-separated stanza with a doc comment near the other DRA declarations — e.g. `// allocationModeExactCount is the Kubernetes DRA DeviceRequest allocationMode enum value; it is an upstream API constant, not an internal token.` — which also removes the forced realignment of `noAllocProbePrefix`. Or drop the constant.

validators/performance/nccl_all_reduce_bw_constraint.go
```diff
@@ -62,6 +62,8 @@ const (
 	// ncclTrainingRuntimeName is the name of the TrainingRuntime resource.
 	// Must stay in sync with runtime.yaml.
 	ncclTrainingRuntimeName = "nccl-all-reduce-runtime"
+
+	skipMsgNCCLFewNodes = "skipped - requires at least 2 GPU nodes for EW fabric test"
```
njhensley: **🔵 Nitpick — Only one of four sibling `"skipped - ..."` strings was extracted, and the new const lacks a doc comment**

Three structurally identical skip-result strings in the *same function* carry the identical load-bearing prefix and were left inline: `:255` (`"skipped - requires Service + Accelerator"`), `:326` (the `fmt.Sprintf("skipped - benchmark profile %s ...")`), and `:329`. So the file now has one skip message as a constant and three as literals, with no stated rule for which is which. The new constant is also the only member of this const block without a doc comment — `maxMessageSizeTCP`, `ncclTrainJobName` and `ncclTrainingRuntimeName` all have one, several with "Must stay in sync with runtime.yaml"-style rationale.

**Blast radius:** One file; a future reader has no principle for where the next skip message should go, so the inconsistency compounds.

**Fix:** Either finish the extraction (hoist all four into one documented `// skipMsg* are the constraint-result strings returned when the check cannot run.` stanza) or revert it. Either way add a doc comment recording that the `skipped ` prefix is contractual — dispatched on at `nccl_all_reduce_bw.go:78` — since nothing else in the codebase states that invariant.

tools/coverage/model.go
```diff
@@ -30,6 +30,8 @@ const (
 	// only the Azure UAT trees). Distinct from not-yet-covered: the assets are
 	// present but inert pending a revive-or-retire decision (DC6, #1280).
 	StatusStubbed Status = "stubbed"
+
+	emDash = "—"
```
njhensley: **🔵 Nitpick — `emDash` is parked inside the closed `Status` vocabulary const block**

This block is introduced by the `type Status string` doc comment describing "a small, closed vocabulary so the rendered table is unambiguous". `emDash` is a rendering placeholder and is the only member that is neither typed `Status` nor documented, while all three siblings are both. Neither of its call sites is in `model.go` — they are in `build.go` (:174, :180, :182) and `render.go` (:97, :115). Because it is an untyped string constant, `var s Status = emDash` compiles (I verified), letting a placeholder masquerade as a coverage verdict — theoretical, since nothing in the package does that.

**Blast radius:** `tools/coverage` only (build-time tooling). Comprehension and type cohesion, not runtime — I confirmed the rendered matrix output is byte-identical.

**Fix:** Move it to its own documented one-liner in `render.go` (the presentation layer both call sites already live in): `// emDash is the placeholder rendered for an empty matrix cell.`

pkg/oci/push_test.go
```diff
@@ -106,7 +106,7 @@ func extractFilesFromOCIArtifact(t *testing.T, ociLayoutDir, digest string) map[
 	tr := tar.NewReader(gzr)
 	for {
 		header, err := tr.Next()
-		if err == io.EOF {
+		if stderrors.Is(err, io.EOF) {
```
njhensley, line 109: **🔵 Nitpick — errorlint allowlists `io.EOF` — this change is style alignment, and three identical sites remain**

The change itself is correct and matches the CLAUDE.md rule. Flagging it only because errorlint did not require it: I reproduced in a scratch module that `err == io.ErrUnexpectedEOF`, `err == fs.ErrNotExist` and `err == os.ErrClosed` all fire under errorlint, while `err == io.EOF` on the adjacent line does not — go-errorlint special-cases `io.EOF` because `io.Reader` implementations are documented to return it unwrapped. Note `errorlint` is *not* in the `_test.go` exclusion list in `.golangci.yaml`, so it does apply to these files; it simply never fired. Three structurally identical tar-loop sites remain unconverted: `pkg/oci/archive_test.go:371`, `pkg/oci/archive_checkpoint_c_test.go:204`, `pkg/evidence/attestation/oci_test.go:291` — running `golangci-lint run -c .golangci.yaml ./pkg/oci/... ./pkg/evidence/attestation/...` reports 0 issues with them present.

**Blast radius:** Test files only, no runtime behavior. The risk is a future reader assuming the rule is uniformly applied in `pkg/oci` when it is not.

**Fix:** Either convert the other three in this PR for consistency, or note in the body that the sweep is deliberately scoped to `push_test.go`.

njhensley (commented):
📋 **Multi-persona review**

> **Method:** 3 independent persona passes (Go Correctness & Idiom · Observability & Contract · Consistency/Lint-Premise), with every finding re-derived from the resolved code, confirmed, refuted or re-tiered by a senior meta-reviewer. Line links pinned to head `5204c9cb`.
> **Tier legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick · ✅ Confirmed non-issue

## Overall assessment

The code is **behaviorally inert and correct**. Every substitution is value-identical: the DRA ResourceClaim payload serializes byte-for-byte the same for both the v1beta1 and v1 `exactly` shapes, the coverage-matrix output is unchanged (no golden files to regenerate), and the CTRF skip message is unchanged. `go vet`, `golangci-lint run -c .golangci.yaml` and `go test` are all clean on the four changed packages. Nothing here can break CI or ship a bad artifact.

The issue is the premise rather than the code.

### 🟡 Minor — none of the four changes were surfaced by golangci-lint

I reproduced this in scratch modules using this repo's exact linter settings, so it's worth laying out the evidence:

| Change | Why the linter never fired |
|---|---|
| `"—"` → `emDash` (tools/coverage) | goconst's `min-len` defaults to **3 and counts runes**, not bytes. An em dash is 1 rune (3 bytes), so it is never a goconst candidate at *any* occurrence count. Controls: `"abc"` ×6 fired; `"—"` ×6 did not; a 2-rune/6-byte `"——"` ×6 also did not — which rules out a byte-length reading. |
| `"ExactCount"` (validators/conformance) | 2 occurrences in non-test files, below `min-occurrences: 3`. The 2 further occurrences are in `_test.go`, and `ignore-tests: true` removes them from the **count**, not just from the report (verified with a control run at `ignore-tests: false`, which does fire). |
| NCCL skip message (validators/performance) | Exactly **1** non-test occurrence pre-change. The second grep hit in that file was the `slog` line, which was a *different* literal (`"Skipping NCCL All Reduce bandwidth validation: requires at least 2 GPU nodes for EW fabric test"`). The other 3 were in `_test.go`. |
| `err == io.EOF` → `stderrors.Is` (pkg/oci) | go-errorlint **allowlists `io.EOF`**, since `io.Reader` implementations are documented to return it unwrapped. See the inline comment for the reproduction. |

So the title — *"resolve goconst and errorlint warnings"* — is false for both linters it names, and the body's *"surfaced by golangci-lint / static analysis"* framing doesn't hold. Since this repo squash-merges, that becomes a permanent entry in `main`'s history and would mislead anyone later auditing why these constants exist.

**Suggested:** retitle and reword the body to describe this as voluntary readability cleanup (which is a perfectly good reason to land it) rather than a lint fix. If a real goconst sweep is wanted, run it and fix what it actually reports.

The one place the PR makes something measurably *worse* is the NCCL `slog` line — see inline.

## ✅ Confirmed non-issues (examined and cleared)

- **Em dash substitution is exact** — all five standalone `"—"` literals are U+2014 and were converted; the remaining em dashes in `tools/coverage` are inside prose sentences where substitution would be wrong. Correctly scoped.
- **No doc or golden regeneration owed** — no testdata under `tools/coverage`, and the constant equals the literal.
- **`stderrors` alias already existed** in `push_test.go` imports.
- **`"ExactCount"` literals kept in `secure_access_check_test.go:788,:1712`** — correct; those still act as a canary against the constant drifting.
- **PR checklist / unsquashed branch — raised by a persona, refuted on review.** The `- [x] Linter passes` box was flagged against the disclosed yamllint gap. Dismissed: this PR touches zero YAML, the gap was disclosed transparently in the body, `golangci-lint` was independently re-run clean on all four packages, and `origin/main` is squash-merged so the local commit count is immaterial.
- **Environment caveat, not a PR issue:** `TestHelmV4_2_3ExplicitVersionPull` fails on my machine only because it has helm v4.2.0 against the v4.2.3 pin in `.settings.yaml`. Unrelated file, unrelated code path — `make tools-setup` resolves it.

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major | 0 | — |
| 🟡 Minor | 2 | PR title/body premise is factually wrong; NCCL `slog` line stutters |
| 🔵 Nitpick | 5 | Tautological test assertions · const placement ×2 · arbitrary extraction · partial errorlint sweep |

**Recommendation:** approve-able with comments. Two things worth doing before merge — (1) retitle/reword so the squash-merged commit isn't a false record, and (2) revert the `slog` line to a plain sentence. Everything else is fine to defer in writing.

andrewwhitecdw:
Addressed the review nits and repushed:

- Fixed the stuttering NCCL skip message by introducing ncclVariantDisplayName; the message now reads e.g. "skipped - benchmark profile gb200/eks does not implement the default NCCL variant".
- Cleaned up the "Skipping ...: skipped - ..." log line in the few-nodes path.
- Moved emDash out of the Status const block into render.go.
- Documented allocationModeExactCount.
- Extracted the remaining inline inference skip strings into inferenceSkipMsgNoDynamoPlatform and inferenceSkipMsgCRDNotInstalled.
- Added prefix guard tests for both NCCL and inference skip messages.
- Applied errors.Is(err, io.EOF) consistently across the OCI archive tests.

Verification:
- go test -mod=mod ./validators/performance ./validators/conformance ./tools/coverage ./pkg/oci ./pkg/evidence/attestation: all pass
- golangci-lint run on the same packages: 0 issues
- Fork-local security sweep dispatched: https://github.com/andrewwhitecdw/aicr/actions/runs/30321084099

njhensley (approved):
Approving - and welcome, nice first PR.

Good response round. Thanks for extending the pattern to inference_perf.go and catching that skipMsgNCCLNoProfile was printing a constraint name where a variant name belonged. Build, vet, test and lint are all clean at 2357b9ee.

Issue #1757, opened 2026-07-15
[release blocker] Bundle-time GPU driver-ownership coherence validation for preinstalled-driver profiles


yuanchen8911:
> [!IMPORTANT]
> **Release-blocking:** this must land before the next tagged release. PR #1756 ships the AKS default flip without a bundle-time guard; recipes persisted before the flip render an incoherent driver/DRA combination when bundled by a post-flip binary. Users consume tagged releases, so landing this before the next tag keeps the transition atomic at the release boundary (agreed disposition in #1756 review).

## Problem

PR #1756 switches the AKS recipes to the AKS driver-only (Azure-managed) GPU profile by default (`gpu-operator driver.enabled=false`, `toolkit.enabled=false`, `operator.runtimeClass=nvidia-container-runtime`, DRA `nvidiaDriverRoot=/`). Driver ownership is now a coordinated multi-value profile that can be broken in ways nothing currently blocks before deploy:

1. **Driverless-cluster mismatch.** A recipe assuming a preinstalled driver, bundled for a cluster whose GPU pools were created with `--gpu-driver none`, deploys driverless GPU nodes. Snapshot-driven resolution warns today (`aicr recipe --snapshot`), but nothing fails: `aicr recipe` has no `--set`, so a resolution-time hard error would cut off the documented bundle-time GPU-Operator-managed override path (found in review of #1756).
2. **Partial profile flips.** `--set gpuoperator:driver.enabled=true` alone (without `toolkit.enabled=true` / `operator.runtimeClass=nvidia` / `dradriver:nvidiaDriverRoot=/run/nvidia/driver`) yields a bundle with no containerd `nvidia` handler or a DRA plugin pointed at the wrong driver root.
3. **Legacy recipes.** A `RecipeResult` stores `valuesFile` paths, not frozen values. A pre-#1756 AKS recipe bundled by a post-#1756 binary resolves the new `driver.enabled=false` values while keeping its old baked DRA overrides (no `nvidiaDriverRoot` retarget) — an incoherent bundle with no warning. Directly reproducible with the committed `examples/recipes/aks-training.yaml`, which also needs regeneration.

## Proposal

[...]

PR #1917, opened 2026-07-25
fix(recipe,client): reject duplicate ComponentRef names, omit disable…

### LINKED ISSUE #1874: Recipe/SDK correctness cleanups from #1819 review
## Context

Two small, general correctness cleanups surfaced by the #1819 cross-review. Both are unrelated to driver-ownership and were split out to keep #1819 focused. Reference implementations exist in #1819's pre-trim history (commit `25bd52fda`).

## C — Reject duplicate `ComponentRef.Name`

`RecipeResult.PrepareAndValidate` does not reject a recipe with two component refs sharing the same `Name`. Component names key value materialization, override routing, validation, filtering, and deployment ordering, so a duplicate creates a "validate first ref, emit last ref" ambiguity. AICR-generated recipes never produce duplicates, but a hand-authored or externally-adopted recipe can.

**Acceptance criteria:**
- `PrepareAndValidate` rejects duplicate component-ref names with `ErrCodeInvalidRequest`, before registry back-fill.
- Disabled refs participate (a disabled duplicate must not mask an enabled ref of the same name).
- Table-driven test covering adjacent, non-adjacent, disabled, and empty-name duplicates.

## D — Omit disabled components from the SDK facade

`facadeResultFromInternal` (`pkg/client/v1`) includes disabled components in the SDK facade's deployable set. The facade contract is the set of deployable components, so disabled refs should be filtered out.

**Acceptance criteria:**
- `facadeResultFromInternal` skips `!IsEnabled()` component refs.
- Test asserting only enabled components appear in the facade result.

## Notes

Neither is release-blocking — C needs a malformed/hand-authored recipe; D is a facade-contract nicety. Reference code exists in #1819's pre-trim history.

Related: #1819, #1757

mohityadav8:
## Summary
Two small, general correctness cleanups surfaced by the #1819 cross-review. Both are unrelated to driver-ownership and were split out to keep #1819 focused.

- **C** - `RecipeResult.PrepareAndValidate` now rejects a recipe with two component refs sharing the same `Name`.
- **D** - `facadeResultFromInternal` (`pkg/client/v1`) now omits disabled components from the SDK facade's deployable set.

## Motivation / Context
Component names key value materialization, override routing, validation, filtering, and deployment ordering throughout the bundler/validations/deployer paths, all of which assume a unique name to ref mapping. AICR-generated recipes never produce duplicates, but a hand-authored or externally-adopted recipe can, creating a "validate first ref, emit last ref" ambiguity.

Separately, the SDK facade's contract is the set of deployable components, so a disabled ref should not appear in it.

Fixes: #1874
Related: #1819, #1757

## Type of Change

## Component(s) Affected
- [x] Recipe engine / data (pkg/recipe)
- [x] Other: pkg/client/v1 (SDK facade)

## Implementation Notes

**C - duplicate ComponentRef.Name rejection.** `PrepareAndValidate` now walks all `ComponentRefs`, both enabled and disabled, and rejects the recipe with `ErrCodeInvalidRequest` on the first repeated non-empty `Name`, before registry back-fill. Disabled refs participate in the check so a disabled duplicate cannot mask an enabled ref of the same name, or vice versa. Empty names are exempt since they are never used as a lookup key. Placed immediately after the existing reserved-deployer-key check and before `backfillComponentTypes()`, matching the method's documented required order.

**D - omit disabled components from the SDK facade.** `facadeResultFromInternal` now calls `c.IsEnabled()` at the top of its component-projection loop and skips disabled refs, so the facade's `Components` slice only ever contains deployable components, consistent with the facade's documented contract.

**Follow-up from review (round 2).** Extracted the duplicate-name check into a shared `validateRefNames()` helper and wired it into `finalizeRecipeResult` (`pkg/recipe/metadata_store.go`), so the criteria-resolve boundary (`aicr recipe -o`, `aicr query`, `Client.ResolveRecipe`) now fails closed on duplicate names the same way the file-load/adopt/bundle boundary already does, closing the gap flagged in review. `BundleComponents` (`pkg/client/v1/aicr.go`) now rejects a recipe whose facade `Components` slice is empty, whether from every ref being disabled or from a zero-componentRef recipe, with `ErrCodeInvalidRequest`, matching `DefaultBundler.Make`'s parity. Also documented the componentRef name-uniqueness rule, including the empty-name exemption, in `api/aicr/v1/server.yaml` and `docs/user/api-reference.md`, and the `Components` field's enabled-only contract in `docs/integrator/go-library.md`, `docs/integrator/public-api.md`, and `pkg/client/v1/types.go`.

## Testing

gofmt -l pkg/recipe/metadata.go pkg/recipe/componentref_coherence_test.go pkg/client/v1/aicr.go pkg/client/v1/aicr_internal_test.go # clean
go build ./... # ok
go test -race ./pkg/recipe/... ./pkg/client/... # ok
golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./pkg/client/... # 0 issues
go test -race ./pkg/bundler/... ./pkg/validator/... ./pkg/server/... # ok, no regressions

Coverage: pkg/recipe 88.3% (unchanged), pkg/client/v1 77.4% updated after round-2 additions. `PrepareAndValidate` 94.1%, `facadeResultFromInternal` 100.0%, `validateRefNames` 100.0% (new, fully covered by `TestPrepareAndValidate_RejectsDuplicateNames`), and the `BundleComponents` all-disabled/zero-ref guard covered by `TestBundleComponents_RejectsAllDisabled` and `TestBundleComponents_RejectsZeroComponentRefs`.

New coverage:
- `TestPrepareAndValidate_RejectsDuplicateNames` - table-driven, covering adjacent, non-adjacent, disabled, and empty-name duplicates.
- `TestFacadeResultFromInternal_OmitsDisabledComponents` - asserts only enabled components appear in the facade result.
- `TestBundleComponents_RejectsAllDisabled` - asserts an all-disabled recipe is rejected with `ErrCodeInvalidRequest` instead of silently returning an empty bundle list.
- `TestBundleComponents_RejectsZeroComponentRefs` - asserts a recipe with no componentRefs at all is rejected the same way.
- `TestLoadFromFileWithProvider/hand-authored_duplicate_componentRef_names_are_rejected` - boundary-level test confirming the file-load path fails closed on duplicate names.

## Risk Assessment
- [x] Low - Isolated change, well-tested, easy to revert

Rollout notes: `PrepareAndValidate`'s fail-closed behavior is new; any existing hand-authored or adopted recipe with duplicate component names will now be rejected at load/adopt time. AICR-generated recipes are unaffected since they never produce duplicates. The facade change (D) narrows `Components` for any SDK caller currently relying on seeing disabled refs there, expected to be a no-op for existing callers per the facade's documented contract.

## Checklist
- [x] Tests pass locally (`go test -race`)
- [x] Linter passes (`golangci-lint`)


pkg/recipe/metadata.go
```diff
@@ -563,6 +563,20 @@ func (r *RecipeResult) PrepareAndValidate() error {
 				fmt.Sprintf("recipe component %q uses the reserved deployer override key as its name; %q is reserved for --set deployer:<key> Argo deployer options", r.ComponentRefs[i].Name, ReservedDeployerKey))
 		}
 	}
+	// Reject duplicate component-ref names (enabled or disabled) BEFORE
```
njhensley, line 566: **🟡 Minor — Criteria-resolve path still bypasses the duplicate check**

`finalizeRecipeResult` calls `ValidateCoherence` directly and never `PrepareAndValidate` ([metadata_store.go#L977](https://github.com/NVIDIA/aicr/blob/70b865d75a59b3ab672a03f715c8d12d9fd50b77/pkg/recipe/metadata_store.go#L977)). With an external `--data` base carrying duplicate refs and no matching overlay, `initBaseMergedSpec` plain-copies them and the duplicates reach `aicr recipe -o`, `aicr query`, and `Client.ResolveRecipe`.

**Blast radius:** Limited. Every *deploy* path still fails closed — `bundler.Make` calls `PrepareAndValidate` at `bundler.go:237` — so the worst case is an emitted recipe that a later `aicr bundle -r` rejects. #1874's acceptance criteria asked for exactly what shipped, and the #1656 reserved-name check has the identical gap, so this is a scope boundary rather than a miss.

**Fix:** Follow-up is fine: extract a shared `validateRefNames()` and call it from `finalizeRecipeResult` alongside `ValidateCoherence`, closing both gaps at once. Worth a code comment either way — `RecipeMetadataSpec.Merge` silently last-wins-collapses same-name refs via `componentMap`, so when an overlay *does* merge, the duplicate vanishes without warning.

pkg/client/v1/aicr.go
```diff
@@ -857,6 +857,9 @@ func facadeResultFromInternal(r *recipe.RecipeResult, name string) *RecipeResult
 		internal:     r,
 	}
 	for _, c := range r.ComponentRefs {
+		if !c.IsEnabled() {
```
njhensley, line 860: **🟡 Minor — Facade `Components` narrowing isn't reflected in the integrator docs**

To be clear up front, this change is better than advertised: `applyRegistryDefaults` populates `Type` on disabled refs too, so pre-PR `Client.BundleComponents` returned fully-populated Helm bundles for components that must never deploy — diverging from `DefaultBundler.Make`'s `filterEnabledComponents`. This is a genuine bug fix. The `types.go:271` godoc ("lists the deployable components") already supports it, and the `BundleComponents` "1:1 — same order, same length" promise is intact because bundles derive from the already-filtered `r.Components`.

**Blast radius:** Docs only, but observable: `recipes/overlays/ocp.yaml` disables ~10 componentRefs, so an SDK caller resolving an OCP recipe sees a materially shorter `Components` slice after upgrade. [`docs/integrator/go-library.md#L276`](https://github.com/NVIDIA/aicr/blob/70b865d75a59b3ab672a03f715c8d12d9fd50b77/docs/integrator/go-library.md#L276) and `docs/integrator/public-api.md:67` both describe the field with no enabled-only qualifier.

**Fix:** Add the qualifier to both doc pages — e.g. "`Components` — enabled (deployable) components only; disabled refs remain visible via `Resolved().ComponentRefs`" — and point the `types.go:271` godoc at `Resolved()` for the full set.

pkg/recipe/metadata.go
```diff
@@ -563,6 +563,20 @@ func (r *RecipeResult) PrepareAndValidate() error {
...
+	// registry back-fill. See #1874.
+	seen := make(map[string]bool, len(r.ComponentRefs))
+	for i := range r.ComponentRefs {
+		name := r.ComponentRefs[i].Name
+		if name == "" {
+			continue
+		}
+		if seen[name] {
+			return errors.New(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("recipe has duplicate component ref name %q", name))
```
njhensley: **🔵 Nitpick — Error names the duplicate but not the colliding positions**

Fail-fast is right here — it matches the reserved-key loop immediately above, and the split between cheap pre-flight name gates and `ValidateCoherence`'s aggregating shape pass is deliberate, not accidental. But the message identifies only the name.

**Blast radius:** On a 20-component machine-generated recipe the author can't tell which two entries collided or which one is the stray.

**Fix:** Make `seen` a `map[string]int` and emit `fmt.Sprintf("recipe has duplicate component ref name %q (refs %d and %d)", name, seen[name], i)`. Costs nothing.

pkg/recipe/metadata.go
```diff
@@ -563,6 +563,20 @@ func (r *RecipeResult) PrepareAndValidate() error {
 				fmt.Sprintf("recipe component %q uses the reserved deployer override key as its name; %q is reserved for --set deployer:<key> Argo deployer options", r.ComponentRefs[i].Name, ReservedDeployerKey))
 		}
 	}
+	// Reject duplicate component-ref names (enabled or disabled) BEFORE
+	// registry back-fill. See #1874.
+	seen := make(map[string]bool, len(r.ComponentRefs))
+	for i := range r.ComponentRefs {
+		name := r.ComponentRefs[i].Name
+		if name == "" {
```
njhensley: **🔵 Nitpick — Empty-name exemption leaves the same collision class open (pre-existing)**

Nothing rejects an empty `Name` — `coherenceProblem` never inspects it. An enabled nameless ref needs an explicit `type: Helm` + `source` + `version` to clear coherence (an unset type is caught by the `default:` branch), but if it does, two of them collide *deployer-dependently*: `SafeJoin(outputDir, "")` resolves to the base dir itself (`helpers.go:56`) so flux clobbers one `helmrelease.yaml`, while argocd's `IsSafePathComponent("")` rejects.

**Blast radius:** Entirely pre-existing — two empty-named refs collided identically before this PR — and outside #1874's scope, so not a defect this PR owns. Noted only because the PR is already adding a name-validity loop.

**Fix:** Optional: make the exemption an outright rejection (`return errors.New(errors.ErrCodeInvalidRequest, "component ref has an empty name")`), which also aligns Go with `api/aicr/v1/server.yaml:1628`'s `required: [name]`.

pkg/recipe/componentref_coherence_test.go
```diff
@@ -304,3 +304,65 @@ func TestPrepareAndValidate_RejectsReservedDeployerName(t *testing.T) {
...
+				{Name: "b", Type: ComponentTypeHelm, Source: "https://charts.example.com", Version: "1.0.0"},
+			},
+			false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			r := &RecipeResult{ComponentRefs: tt.refs}
+			err := r.PrepareAndValidate()
+			if (err != nil) != tt.wantErr {
```
njhensley, line 386: **🔵 Nitpick — Test asserts only `err != nil`, not the error code or message**

The check *is* currently pinned — mutation-tested via `go test -overlay` with the production block stripped, and all three `wantErr:true` subtests fail. So this is drift risk, not a present false pass: `PrepareAndValidate` has four rejection paths (reserved key, back-fill failure, canonicalization, coherence), so a future fixture tweak making one ref merely incoherent would keep this green while the duplicate rule silently rots.

**Blast radius:** Test only. The sibling `TestPrepareAndValidate_RejectsReservedDeployerName` fifteen lines up already models the fix, asserting both the code and the message.

**Fix:** Behind the existing `wantErr` branch, add `stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, ""))` and `strings.Contains(err.Error(), "duplicate")`. Two lines.

pkg/recipe/componentref_coherence_test.go
```diff
@@ -304,3 +304,65 @@ func TestPrepareAndValidate_RejectsReservedDeployerName(t *testing.T) {
...
+				{Name: "a", Type: ComponentTypeHelm, Source: "https://charts.example.com", Version: "1.0.0"},
+				{
+					Name:      "a",
+					Overrides: map[string]any{"enabled": false},
+				},
+			},
+			true,
+		},
+		{
+			"empty names not flagged",
```
njhensley: **🔵 Nitpick — `"empty names not flagged"` uses disabled refs, so it doesn't pin what it's named for**

Both refs are disabled, so `backfillComponentTypes` and `ValidateCoherence` skip them entirely (each guards on `IsEnabled()`). The case therefore passes even if the `name == ""` exemption were deleted — it passes for a reason unrelated to the exemption it's named after.

**Blast radius:** Test only. The configuration that actually reaches the deployers is two *enabled* nameless refs, and that has no coverage.

**Fix:** Swap the fixture to two enabled refs with `Type: ComponentTypeHelm` + `Source` + `Version` (the shape that clears coherence), or add that variant alongside. Makes the intended behavior explicit either way you decide the empty-name question above.

pkg/recipe/componentref_coherence_test.go
```diff
@@ -304,3 +304,65 @@ func TestPrepareAndValidate_RejectsReservedDeployerName(t *testing.T) {
...
 	}
 }
+
+func TestPrepareAndValidate_RejectsDuplicateNames(t *testing.T) {
+	tests := []struct {
+		name    string
+		refs    []ComponentRef
+		wantErr bool
+	}{
+		{
```
njhensley, line 322: **🔵 Nitpick — Positional struct literals where every sibling table in the file uses keyed fields**

`TestPrepareAndValidate_RejectsReservedDeployerName` twenty lines above uses `name:` / `refs:` / `wantErr:`, as does `TestComponentRefCoherenceProblem` earlier in the file.

**Blast radius:** Readability only. Keyed literals survive a field insertion into the anonymous struct without silently reassigning values, which is why the file already prefers them.

**Fix:** Add the `name:` / `refs:` / `wantErr:` keys.

pkg/recipe/componentref_coherence_test.go
```diff
@@ -304,3 +304,65 @@ func TestPrepareAndValidate_RejectsReservedDeployerName(t *testing.T) {
 		})
 	}
 }
+
+func TestPrepareAndValidate_RejectsDuplicateNames(t *testing.T) {
```
njhensley, line 316: **🔵 Nitpick — Worth one more edge case: reserved-key-vs-duplicate precedence**

Two refs both named `deployer` hit the reserved-key loop first (`metadata.go:560`) and surface the "reserved" message, never the "duplicate" one. That precedence is deliberate and nothing currently pins it — and since neither the code nor the message is asserted today, a precedence inversion would be invisible.

**Blast radius:** Test completeness only.

**Fix:** Add the precedence case. (Three-way duplicates exercise no new path, and whitespace-padded names are treated as distinct — consistent with the total absence of name normalization anywhere in the codebase — so both of those are skippable.)

njhensley (commented):
## 📋 Multi-persona review

> **Method:** 4 independent persona passes (Correctness & Invariants · API-Contract & Compatibility · Test-Coverage · Domain & Architecture), each finding then confirmed, refuted, or re-tiered by a senior meta-reviewer that re-derived it from the resolved code rather than the persona's claim. Line links pinned to head `70b865d7`.
>
> **Tier legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick · ✅ Confirmed non-issue

Both changes are correct, minimal, and land in the right place. The duplicate-name check is idiomatic with the reserved-deployer-key check directly above it and is genuinely pinned by its tests (mutation-verified: stripping the production block fails all three positive subtests).

The facade change is **better than advertised** — it isn't just a contract tidy-up, it's a bug fix. `applyRegistryDefaults` populates `Type` on disabled refs too, so pre-PR `Client.BundleComponents` returned fully-populated Helm bundles for components that must never deploy, diverging from `DefaultBundler.Make`'s `filterEnabledComponents`. That's observable today: `recipes/overlays/ocp.yaml` disables ~10 refs.

Nothing blocks merge. Four persona findings arrived at 🟠 Major; all four were demoted once re-derived against the code — three because the immediate precedent (#1656's reserved-name check) has the identical gap, one because the claimed "1:1 contract break" doesn't exist.

---

## Additional findings (not inline-anchorable)

These land on unchanged context lines or on files outside the diff, so they couldn't be posted inline.

### 🟡 `PrepareAndValidate` godoc enumerates its steps "in the required order" and now omits the new one
[`pkg/recipe/metadata.go#L535`](https://github.com/NVIDIA/aicr/blob/70b865d75a59b3ab672a03f715c8d12d9fd50b77/pkg/recipe/metadata.go#L535)

The comment reads "…in the required order: reject refs named with the reserved deployer override key, back-fill missing types, canonicalize the type casing, then validate coherence" and then "so the **three** steps cannot drift or be partially applied." There are now four. This is the one defect the PR itself introduces — the comment was accurate before — and it's the ordering contract the next editor reads before inserting anything else into this method, so a stale enumeration is precisely the drift it was written to prevent.

**Fix:** add "reject duplicate non-empty ref names (all refs, enabled or disabled)" after the reserved-key clause; change "three steps" → "four steps".

### 🟡 All-disabled recipe: `BundleComponents` returns `(empty, nil)` where `MakeBundle` returns 400
[`pkg/client/v1/aicr.go#L1061`](https://github.com/NVIDIA/aicr/blob/70b865d75a59b3ab672a03f715c8d12d9fd50b77/pkg/client/v1/aicr.go#L1061)

With every ref disabled the projection loop never executes, so `bundles := make([]ComponentBundle, 0, 0)` is returned with a `nil` error. The equivalent `MakeBundle` path fails with `ErrCodeInvalidRequest "recipe has no enabled components after filtering"` ([`bundler.go#L968`](https://github.com/NVIDIA/aicr/blob/70b865d75a59b3ab672a03f715c8d12d9fd50b77/pkg/bundler/bundler.go#L968)). The two public entry points should agree on whether "nothing to deploy" is an error; today an SDK caller that loops over the result and deploys silently no-ops. Newly reachable via the skip — though the pre-PR behavior was *worse*, not better.

**Fix:** at the top of `BundleComponents`, `if len(r.Components) == 0 && len(r.internal.ComponentRefs) > 0 { return nil, errors.New(errors.ErrCodeInvalidRequest, "recipe has no enabled components") }`.

### 🟡 The uniqueness rule isn't documented alongside its siblings
[`docs/user/api-reference.md#L486`](https://github.com/NVIDIA/aicr/blob/70b865d75a59b3ab672a03f715c8d12d9fd50b77/docs/user/api-reference.md#L486)

`PrepareAndValidate` gates `aicr bundle -r`, `aicr validate -r`, and `POST /v1/bundle`, so a hand-authored recipe that previously bundled now returns 400 — yet nothing in `docs/user/api-reference.md` or `api/aicr/v1/server.yaml` says componentRef names must be unique, even though both files carefully enumerate every sibling coherence rejection. Tiered Minor rather than Major because I ran the control: **#1656's reserved-`deployer`-name rejection is also undocumented**, so this raises the bar rather than backsliding from it. (A `yq` sweep of every `componentRefs` list under `recipes/` and `tests/` found zero duplicates, so no checked-in recipe is newly rejected.)

**Fix:** one sentence next to the existing "Incoherent refs are rejected with HTTP 400 naming the component," plus the `componentRefs.type` description at `server.yaml:1640`. And tick the PR's docs checklist item.

### 🔵 Testing section omits the per-package coverage delta
CLAUDE.md's coverage gate step 6 asks for a `before → after` line. Measured at HEAD: `pkg/recipe` 88.3%, `pkg/client/v1` 77.4%; `PrepareAndValidate` 94.1%, `facadeResultFromInternal` 100.0%. No coverage risk — bookkeeping only.

### 🔵 No boundary-level test through `LoadFromFileWithProvider` / adopt / `POST /v1/bundle`
Low value: all three boundaries call the same method and their call sites are covered, and the #1656 precedent has no boundary test either. A single loader-level duplicate fixture asserting `ErrCodeInvalidRequest` would cover the most-used path if you want one cheap.

---

- **`ErrCodeInvalidRequest` vs `ErrCodeConflict`** — 400 is correct. A duplicate name is a malformed payload, not a resource-state conflict; `ErrCodeConflict` is reserved for already-exists/version-mismatch where "the request itself is well-formed."
- **`BundleComponents` 1:1 contract "break"** — refuted. Bundles derive from the already-filtered `r.Components` (`aicr.go:1061-1062`), so "same order, same length" holds exactly. Raised 🟠 by a persona; does not survive.
- **Case-sensitive dedup** — two personas conflicted; tie broken. Every in-process lookup (`ComponentRegistry.Get`'s `byName` map, `componentValues[ref.Name]`, `GetComponentRef`) is exact-string, so there's no alias skew. Names *do* become path components, so `GPU-Operator`/`gpu-operator` collide on APFS/NTFS — but that predates the bundler's entire history, and case-folding would over-reject on Linux. If ever addressed, the right place is the deployer output boundary, not `PrepareAndValidate`.
- **Asserting `out.internal` retains disabled refs** — refuted as tautological. `facadeResultFromInternal` assigns `internal: r` in the struct literal *before* the loop (`aicr.go:857`), so the assertion can never fail. The meaningful pin would be `out.Resolved().ComponentRefs` length via the public accessor.
- **Two fixes in one PR** — non-issue. #1874 defines both as items C and D of one scoped cleanup with separate acceptance criteria; splitting would leave the issue half-closed for 8 production lines.
- **`IsEnabled()` fragility** — sound. Nil-map safe, missing key → enabled, non-bool → `slog.Warn` + fail-open. The facade skip uses the identical predicate as `filterEnabledComponents` and `ValidateCoherence`.
- **Ordering / placement before back-fill** — correct. Nothing before it touches `Name`; `canonicalizeComponentTypes` only rewrites `Type`.
- **Test suite** — green across `pkg/recipe`, `pkg/client/v1`, `pkg/bundler/...`, `pkg/server`, `pkg/cli`, `pkg/validator/...`, `pkg/evidence/...`.

---

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major | 0 | *(4 persona-raised, all demoted on re-derivation)* |
| 🟡 Minor | 5 | stale ordered-step godoc · `BundleComponents` `(empty, nil)` · facade docs · uniqueness rule undocumented · resolve-path gap |
| 🔵 Nitpick | 8 | error indices · empty-name exemption · test error-code assertion · disabled-ref empty-name case · positional literals · precedence case · coverage delta · boundary test |

**Recommendation: Approve with comments.** Worth fixing in-PR: the stale `PrepareAndValidate` godoc (self-inflicted by this diff) and the `BundleComponents` empty-vs-error divergence. The docs items are a fair ask, but the #1656 precedent is equally undocumented, so defer-in-writing is defensible. Everything else is optional polish.

pkg/client/v1/aicr.go
```diff
@@ -1055,6 +1058,16 @@ func (c *Client) BundleComponents(ctx context.Context, r *RecipeResult) ([]Compo
 		return nil, preflightErr
 	}
 
+	// Agree with DefaultBundler.Make (bundler.go), which fails closed with
+	// ErrCodeInvalidRequest when every component is disabled. Without this,
+	// an all-disabled recipe silently returns an empty, successful bundle
+	// list here — an SDK caller that loops over the result and deploys
+	// would silently no-op instead of getting an actionable error.
+	if len(r.Components) == 0 && r.internal != nil && len(r.internal.ComponentRefs) > 0 {
```
njhensley: **🟡 Minor — All-disabled guard: comment claims parity with `DefaultBundler.Make` it doesn't quite have, plus a dead conjunct and no test**

The `len(r.internal.ComponentRefs) > 0` conjunct means a **zero-componentRef** recipe still returns `([], nil)` here, while `DefaultBundler.Make` → `filterEnabledComponents` ([bundler.go#L969](https://github.com/NVIDIA/aicr/blob/486b11a7e5e9bcfa5e5d30dbf686e0279569214c/pkg/bundler/bundler.go#L969)) returns `ErrCodeInvalidRequest` for that same input — so "Agree with DefaultBundler.Make" overstates it. Separately `r.internal != nil` is dead: line 1004 already returns on nil internal. And `rg 'no enabled components'` hits only the two production sites — the guard has no test at all.

**Blast radius:** SDK callers of `Client.BundleComponents` on a zero-ref recipe (reachable via `LoadRecipe` of a hand-authored file) still silently no-op. Low frequency; the failure is a missing error rather than a wrong artifact.

**Fix:** Guard on `len(r.Components) == 0` alone — dropping both conjuncts makes the two paths agree exactly — and add an all-disabled plus a zero-ref case in `pkg/client/v1`.

---

**On @coderabbitai's version of this finding** — it argued the guard should key off `r.internal.ComponentRefs` because `Components` is caller-mutable. I'd respectfully decline that part: this method's own godoc tells callers to "pass that call's `*RecipeResult` unchanged," and `docs/user/api-reference.md:472` directs subsetting through the `bundlers` filter, so hand-trimming `Components` is unsupported by convention — and an actionable error beats the old silent empty success even in that case. The zero-ref divergence is the part worth fixing.

pkg/recipe/metadata_store.go
```diff
@@ -968,6 +968,18 @@ func finalizeRecipeResult(provider DataProvider, criteria *Criteria, mergedSpec
...
 
+	// Reject duplicate component-ref names before they reach ValidateCoherence
+	// or any caller of BuildRecipeResult/ResolveRecipe. RecipeMetadataSpec.Merge
+	// silently last-wins-collapses same-name refs via componentMap, so this
+	// only fires when an external --data base carries duplicates AND no
+	// matching overlay merge collapses them. Mirrors the same check
+	// PrepareAndValidate runs at the file-load/adopt/bundle boundary, so the
+	// criteria-resolve boundary (aicr recipe -o, aicr query,
+	// Client.ResolveRecipe) fails closed on the same recipes too. See #1874.
+	if err := validateRefNames(mergedSpec.ComponentRefs); err != nil {
```
njhensley, line 979: **🔵 Nitpick — This gate is effectively unreachable — honest defense-in-depth, but worth knowing (and untested)**

Reproduced twice, independently. `RecipeMetadataSpec.Merge` rebuilds `ComponentRefs` from a name-keyed map, so a single overlay merge last-wins-collapses duplicates *before* `validateRefNames` sees them. And because `LayeredDataProvider` layers an external `--data` dir *over* the embedded catalog, the embedded `monitoring-hpa` overlay (`criteria: {intent: any}`) always matches — an external tree cannot delete it — so ≥1 overlay always merges. Copying `recipes/` to `/tmp`, duplicating the `nfd` ref in `overlays/base.yaml`, and running `aicr recipe --data … --service eks --accelerator h100 --intent training --os ubuntu` exits 0 with a single `nfd`.

**Blast radius:** None today — no in-tree recipe has duplicates, and the deploy boundary fails closed regardless. Deliberately **not** tiered higher: the comment directly above already says exactly this ("only fires when an external --data base carries duplicates AND no matching overlay merge collapses them"), so the code doesn't misrepresent itself, and the docs added in this commit are correctly scoped to the POST /v1/bundle request body, which *does* reject via `adoptRecipe` → `PrepareAndValidate`.

**Fix:** No change required. Two notes if you want to revisit later: (1) to make it actually bite, validate pre-merge in `initBaseMergedSpec` plus per-overlay inside `mergeOverlayChains`; (2) there's no test covering this call site — a store-level test with a duplicate-carrying base and no matching overlay would have surfaced the inertness at authoring time, and without one the line is free to stay dead silently.

api/aicr/v1/server.yaml
```diff
@@ -1629,7 +1629,10 @@ components:
             properties:
               name:
                 type: string
-                description: Component name
+                description: >-
+                  Component name. Must be unique within componentRefs (across
```
njhensley, line 1657: **🔵 Nitpick — Uniqueness description omits the empty-name exemption**

Confirming @coderabbitai's finding, and I checked the escape hatch it didn't: `required: [name]` on line 1628 is a *presence* assertion, not non-emptiness (`"name": ""` satisfies it), and `pkg/server` performs **no runtime OpenAPI validation** — the spec is documentation-only (its only in-repo consumers are a doc comment in `pkg/server/doc.go` and the enum-drift test; there's no `kin-openapi`/`libopenapi` in `go.mod`). So `validateRefNames`' `name == ""` exemption really is reachable and repeated empty names really are accepted.

**Blast radius:** Documentation accuracy only — a nameless enabled ref is already degenerate. Same wording applies to `docs/user/api-reference.md:487-489`.

**Fix:** Append "(refs with an empty name are exempt)" to both descriptions — or, better, make an empty `Name` on an *enabled* ref a `coherenceProblem`, which makes the docs true as written and closes the case below.

pkg/recipe/componentref_coherence_test.go
```diff
@@ -304,3 +304,103 @@ func TestPrepareAndValidate_RejectsReservedDeployerName(t *testing.T) {
...
+			wantErr: true,
+		},
+		{
+			// Both refs are enabled and nameless — the actual shape that
+			// reaches the deployers if the name=="" exemption were ever
+			// removed. Disabled nameless refs would pass trivially for an
+			// unrelated reason (backfillComponentTypes/ValidateCoherence
+			// skip disabled refs entirely), so this must use enabled ones
+			// to pin the exemption it's named for.
+			name: "empty names not flagged",
```
njhensley, line 354: **🔵 Nitpick — The strengthened fixture is right, but note it now blesses a degenerate shape**

The rewrite from disabled to enabled refs is a genuine improvement and the comment explaining why is exactly right. Worth naming what it now durably pins, though: two **enabled, fully-shaped, nameless** Helm refs sail through the entire `PrepareAndValidate` boundary (reserved-key → validateRefNames → backfill → canonicalize → ValidateCoherence).

**Blast radius:** Pre-existing, not introduced here. But those refs are exactly as unroutable downstream as the duplicates this PR rejects — `Name` is the key for values materialization, override routing, and deployment ordering.

**Fix:** Nothing for this PR. A follow-up making an empty `Name` on an enabled ref a `coherenceProblem` would close it and make the new API docs true as written.

njhensley (dismissed):
## 📋 Re-review — head `486b11a7`

> **Method:** 2 fresh persona passes (Follow-Through Verifier · Correctness & Regression) over the delta `70b865d7…486b11a7`, plus adjudication of @coderabbitai's two findings, all routed through an adversarial meta-reviewer that re-derived each claim from the resolved code. Round-1 findings were re-checked individually rather than taken from the commit message.
>
> **Verification at head:** `go build ./...` ok · `go test -race ./pkg/recipe/... ./pkg/client/... ./pkg/bundler/...` all green · `golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./pkg/client/...` 0 issues.
>
> **Tier legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick · ✅ Confirmed non-issue

## Overall — Approving

Clean, well-commented response to round 1 that fixed more than it had to: the extracted shared `validateRefNames` helper and the load-boundary test both went beyond what I asked for. Nothing blocks merge. The single 🟡 is a three-line tightening plus a test.

## Round-1 follow-through: 12 of 13 addressed

Verified independently against the code, not from the commit message:

| # | Finding | Status |
|---|---|---|
| 1 | Stale `PrepareAndValidate` godoc | ✅ enumeration and "four steps" now match the body |
| 2 | `BundleComponents` `(empty, nil)` | ⚠️ partial — see the 🟡 inline |
| 3 | Facade docs (3 surfaces) | ✅ `types.go`, `go-library.md`, `public-api.md`; the dangling "When you" is source wrapping only, renders fine |
| 4 | Uniqueness rule undocumented | ✅ `server.yaml` + `api-reference.md`; "naming the conflicting positions" is now literally true |
| 5 | Resolve-path bypass | ⚠️ partial — helper extracted and wired in, but see the 🔵 on `metadata_store.go` |
| 6 | Error lacked positions | ✅ `map[string]int`, `"(refs %d and %d)"` |
| 7 | Empty-name exemption | ✅ kept, now documented in the helper's godoc |
| 8 | Test asserted only `err != nil` | ✅ code + message asserted. The `tt.name ==` branch is **not** silently-skipping-fragile — a renamed case falls through to the `"duplicate"` check and fails loudly |
| 9 | `"empty names not flagged"` fixture | ✅ now two enabled nameless refs — passes for the right reason |
| 10 | Positional literals | ✅ all keyed |
| 11 | Precedence case | ✅ added, pins the "reserved" message |
| 12 | No boundary test | ✅ **causality confirmed** — stubbing `validateRefNames` via `go test -overlay` makes the new subtest fail, so it genuinely reaches `PrepareAndValidate` rather than erroring incidentally |
| 13 | PR body | ❌ not addressed — see below |

## Additional findings (not inline-anchorable)

### 🔵 The new `BundleComponents` failure mode isn't in its godoc
[`pkg/client/v1/aicr.go#L953`](https://github.com/NVIDIA/aicr/blob/486b11a7e5e9bcfa5e5d30dbf686e0279569214c/pkg/client/v1/aicr.go#L953) — a new user-visible failure mode on a public SDK method, but the godoc has no `Errors:` block, unlike [`LoadRecipe`](https://github.com/NVIDIA/aicr/blob/486b11a7e5e9bcfa5e5d30dbf686e0279569214c/pkg/client/v1/aicr.go#L899) which models the convention. `docs/integrator/go-library.md` doesn't mention it either. A short `# Errors` section listing the nil-client / nil-ctx / nil-result / nil-internal / not-owned / all-disabled cases would cover it.

### 🔵 PR body is stale after the follow-up commit
The docs checklist item is still `[ ]` even though the branch now touches four doc/spec files; coverage is still reported as bare absolutes with no `origin/main` baseline (CLAUDE.md's coverage gate asks for `before% → after%`); and Implementation Notes still describe only the original C/D changes — the `BundleComponents` guard, the `finalizeRecipeResult` check, the extracted `validateRefNames` helper, and the new loader-boundary test are all undescribed.

Also: the description ends with a stray agent status footer — `Session: 0% / Reset in: Not set / Messages left: N/A`. Not part of `.github/PULL_REQUEST_TEMPLATE.md`, and adjacent to the org policy against agent attribution in PRs. Worth deleting when you refresh the body.

- **`r.internal` reachable as nil in the new guard** — no. Line 1004 already returns `ErrCodeInvalidRequest` on nil internal, and `assertOwns` rejects caller-built facades.
- **`validateRefNames` in `finalizeRecipeResult` regressing a shipped recipe** — no. It validates `mergedSpec.ComponentRefs`, which is the same slice assigned to `result.ComponentRefs` (no divergence), it sits correctly after `applyRegistryDefaults` and before `ValidateCoherence`, and a `yq` sweep of every `recipes/**` + `examples/**` + `tests/**` YAML found zero duplicate componentRef names. Full suite green.
- **`tt.name`-keyed assertion branch in the duplicate test** — not fragile. A renamed case falls through to `strings.Contains(err.Error(), "duplicate")`, which the reserved-key message fails, so it breaks loudly rather than degrading silently.
- **Caller mutating the exported `Components` field** (@coderabbitai's framing of the guard) — declined, with reasoning in the inline comment.
- **`docs/integrator/go-library.md` prose** — the dangling "When you" is source line-wrapping; the rendered paragraph reads correctly.
- **New loader test hygiene** — writes only under `t.TempDir()`, uses `t.Context()`, imports clean, lints clean.

## Summary

| Tier | Round 1 | Now |
|------|---------|-----|
| 🔴 Blocker | 0 | 0 |
| 🟠 Major | 0 | 0 |
| 🟡 Minor | 5 | **1** |
| 🔵 Nitpick | 8 | 6 |

**Approving.** The 🟡 (`BundleComponents` zero-ref divergence + dead conjunct + missing test) is worth picking up before merge but doesn't gate it; everything else is optional polish or PR-body hygiene.

pkg/client/v1/aicr_internal_test.go
```diff
@@ -367,6 +367,60 @@ func TestBundleComponents_HelmComponentLoadsManifestFiles(t *testing.T) {
 	}
 }
 
+// TestBundleComponents_RejectsAllDisabled locks in that an all-disabled
+// recipe is rejected the same way DefaultBundler.Make rejects it
+// (bundler.go), rather than silently returning an empty, successful
+// bundle list. See #1917.
+func TestBundleComponents_RejectsAllDisabled(t *testing.T) {
```
njhensley, line 374: **🔵 Nitpick — Fold the two new all-disabled / zero-ref tests into one table-driven test**

`TestBundleComponents_RejectsAllDisabled` and `TestBundleComponents_RejectsZeroComponentRefs` share the same shape — build a client, build a recipe, assert `ErrCodeInvalidRequest` — differing only in the recipe input. This is @coderabbitai's one remaining actionable comment, and it aligns with the repo's "table-driven tests (required for multiple cases)" convention in CLAUDE.md.

**Blast radius:** Test readability only — both tests pass and are correct as written.

**Fix:** Optional: one table with `{name, refs, facadeComponents}` cases ("all disabled", "zero componentRefs") and a shared assertion body. Two focused tests are also perfectly acceptable; take it or leave it.

njhensley (approved):
## ✅ Re-review — head `2109ea21` (approving)

> **Scope:** delta since my round-2 review — a `main` merge (`59de9b7c`) plus `2109ea21` "address remaining review feedback." My round-2 approval was auto-dismissed by the push.
>
> **Verification at head:** `go test -race ./pkg/client/v1/... -run 'BundleComponents_Rejects|Duplicate|OmitsDisabled'` green; the guard-simplification traced by hand (below).
>
> **Tier legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick · ✅ Confirmed non-issue

## Overall — Approving

Every in-scope round-2 item is resolved, with the exact fix I suggested plus tests. Only optional nitpicks remain (1 inline, 1 in this body; none blocking).

## Round-2 follow-through — all addressed

| Round-2 finding | Status |
|---|---|
| 🟡 `BundleComponents` guard (zero-ref divergence + dead conjunct + no test) | ✅ **fully** — now `if len(r.Components) == 0` alone, comment rewritten to cover both cases, and two new tests (`RejectsAllDisabled`, `RejectsZeroComponentRefs`) added, both green under `-race` |
| 🔵 Empty-name exemption undocumented (my confirm of @coderabbitai #1) | ✅ **fully** — `server.yaml:1633` and `api-reference.md:487` now say "when non-empty … Refs with an empty name are exempt" |
| 🔵 PR body stale (R2) | ✅ **fully** — docs checklist now `[x]`, coverage section describes the new tests + `validateRefNames` 100% |
| 🔵 Stray agent status footer (M1) | ✅ **removed** |

## ✅ Confirmed sound

- **Guard simplification is safe.** The nil-`internal` case already returns at `aicr.go:1004` *before* the guard, so by there `internal` is non-nil; an empty `Components` now always errors — exactly the parity with `DefaultBundler.Make` the comment claims (nothing to deploy → `ErrCodeInvalidRequest`, whether zero refs or all-disabled). No valid recipe reaches this with empty `Components` expecting success.
- **New test doubles are fair** — `facadeComponents: nil` directly exercises `len(r.Components) == 0`, and `newClientForBundleTest` binds a real per-Client DataProvider so the `-race` path is clean.

## Remaining — all optional, none blocking

- **🔵 One inline nitpick** — fold the two new tests into a table per @coderabbitai.
- **🔵 `BundleComponents` godoc still lacks an `Errors:` block** (carried from round 2, `aicr.go:994`): it documents the nil-internal `ErrCodeInvalidRequest` inline but not the new all-disabled one, so a public SDK method has an undocumented failure mode. `LoadRecipe` (aicr.go:~899) already models the `Errors:` convention. Not diff-anchorable, so noted here.
- The two round-2 items I'd explicitly marked "no change required / follow-up" (resolve-path gate inertness at `metadata_store.go:979`; the empty-*enabled*-name fixture blessing a degenerate shape) are unchanged — no regression.

## Summary

| Tier | R1 | R2 | Now |
|------|----|----|-----|
| 🔴 / 🟠 | 0 | 0 | 0 |
| 🟡 Minor | 5 | 1 | **0** |
| 🔵 Nitpick | 8 | 6 | 2 |

**Approving.** The round-2 🟡 is resolved with the exact fix suggested plus tests; everything remaining is optional polish. Nice, tight response across three rounds.

PR #290, opened 2026-03-05
feat(validation): container-per-validator execution engine

lalitadithya:
Closes #291
Closes #303
Closes #295
Closes #140

## Summary

Replace the Go `testing.T`-based validation pipeline with a **container-per-validator** execution engine. Each validator is a standalone OCI container run as a Kubernetes Job.

### Validator Contract

- Exit codes: `0`=pass, `1`=fail, `2`=skip (not applicable)
- Evidence via stdout (captured in CTRF report)
- Debug logs via stderr
- Error context via `/dev/termination-log`

### Key Design Decisions

| Decision | Rationale |
|----------|-----------|
| Exit code protocol | No custom IPC, no log parsing |
| One Job per validator | Fault isolation; partial results always available |
| Scoped ClusterRole | Purpose-built RBAC with minimum required permissions |
| Watch API for completion | Reliable event delivery, no informer complexity |
| CTRF reporting | Industry-standard format with existing tooling |
| Catalog in `recipes/` | Same embed pattern as registry and overlays |
| Self-selecting validators | Exit 2 (skip) when infrastructure unavailable |

### What Changed

| Area | Changes |
|------|---------|
| `pkg/validator/` | Orchestrator, catalog loader, CTRF builder, Job deployer, RBAC, result extraction |
| `pkg/validator/catalog/` | Catalog types, phase filtering, image registry override |
| `pkg/validator/ctrf/` | CTRF types, builder, ConfigMap writer/reader |
| `pkg/validator/job/` | SSA-based Job deployer, scoped ClusterRole, Watch-based completion |
| `validators/deployment/` | operator-health, expected-resources, helm-values, gpu-operator-version, check-nvidia-smi |
| `validators/conformance/` | DRA, gang-scheduling, autoscaling, metrics, gateway, security checks |
| `validators/performance/` | NCCL bandwidth, trainer lifecycle |
| `validators/helper/` | Shared pod, GPU, resource utilities |
| `validators/chainsaw/` | Chainsaw assertion runner |
| `pkg/cli/validate.go` | Unified `--namespace` flag, snapshot agent + validator deployment |
| `pkg/snapshotter/agent.go` | Concurrent log streaming (eliminates 2-min delay) |
| `pkg/constraints/` | Extracted constraint evaluation from validator package |
| `recipes/validators/catalog.yaml` | Declarative validator catalog (embedded via `recipes.FS`) |
| `docs/design/002-validatorv2-adr.md` | Architecture decision record |
[...]

PR #1918, opened 2026-07-25
fix(ci): run validators/ unit tests in make test, exclude from coverage gate

### LINKED ISSUE #1752: [Bug]: Validator unit tests are excluded from standard CI
## Bug Description

Validator unit tests are excluded from the repository's standard test and coverage gates. `make test` builds its package list with:

```make
go list ./... | grep -v -e /tests/chainsaw/ -e /validators
```

The reusable GitHub Actions Go test action runs `make test`, so tests below `validators/**` can compile locally yet never execute in the ordinary PR test/coverage path. This gap was introduced with the container-per-validator layout in #290 and was independently noticed while reviewing #1748 and #1745.

Concrete uncovered regression tests include:

- `validators/deployment/nvidia_smi_test.go` in #1748
- `validators/performance/inference_perf_test.go` in #1745

## Impact

Medium. Validator behavior can regress while the standard merge test and coverage gates remain green. GPU end-to-end workflows do not replace deterministic unit coverage for parsing, matching, and constraint logic.

## Component

Validator / CI

## Regression?

No; this is a coverage gap in the validator-v2 package layout introduced by #290.

## Steps to Reproduce

1. Add or modify a unit test under `validators/**`.
2. Run `make test` or the reusable `.github/actions/go-test` action.
3. Observe that the validator test package is absent from the executed package list and from `coverage.out`.

## Expected Behavior

All deterministic validator unit tests run with the repository's standard race and coverage gates, either through `make test` or a dedicated validator-unit target required by the merge gate.

## Actual Behavior

The package selection explicitly removes all paths containing `/validators`, so those tests do not run.

## Acceptance Criteria

- [ ] Add a validator unit-test target, or safely include eligible `validators/**` packages in `make test`.
- [ ] Invoke that target from the standard GitHub PR test/merge gate.
- [ ] Run validator unit tests with the race detector and a bounded timeout.
- [ ] Include eligible validator packages in coverage reporting, or document and enforce a separate coverage policy.
- [ ] Add a guard proving the tests from #1748 and #1745 are discovered and executed.

## Additional Context

Related: #1748, #1745, #290.

mohityadav8:
Makefile excluded /validators from 'go list ./...' in make test, so regression suites added in #1745/#1748 (and the older nvidia-smi, inference-perf, nccl, conformance tests) never ran in CI.

- Drop -e /validators from the make test package filter.
- Exclude validators/ from the test-coverage threshold calc (its package coverage is 41-92%, pulling aggregate from 80.0% to 75.7% and breaking the 80% gate) while still running its tests.

Fixes #1752


Makefile
```diff
@@ -226,13 +226,13 @@ test: test-shell ## Runs unit tests with race detector and coverage (use -short
...
 	AICR_CRITERIA_STRICT=1 \
-	GOFLAGS="-mod=vendor" go test -short -count=1 -race -timeout=$(TEST_TIMEOUT) -covermode=atomic -coverprofile=coverage.out $$(go list ./... | grep -v -e /tests/chainsaw/ -e /validators) || exit 1; \
+	GOFLAGS="-mod=vendor" go test -short -count=1 -race -timeout=$(TEST_TIMEOUT) -covermode=atomic -coverprofile=coverage.out $$(go list ./... | grep -v -e /tests/chainsaw/) || exit 1; \
 	echo "Test coverage:"; \
 	go tool cover -func=coverage.out | tail -1
 
 .PHONY: test-coverage
 test-coverage: test ## Runs tests and enforces coverage threshold (from .settings.yaml quality.coverage_threshold)
-	@coverage=$$(go tool cover -func=coverage.out | grep total | awk '{print $$3}' | sed 's/%//'); \
+	@coverage=$$(grep -v '/validators/' coverage.out | go tool cover -func=/dev/stdin | grep total | awk '{print $$3}' | sed 's/%//'); \
```
njhensley: **🔴 Blocker — Coverage exclusion is in a target CI never invokes; merge gate will hard-fail at 75.8% vs the 80% threshold**

The filter only affects `make test-coverage`. CI never runs that target — `grep -rn 'test-coverage' .github/` returns zero hits. The actual CI path is `qualification.yaml:61` → `.github/actions/go-test/action.yml:72` (`run: make test`) → `.github/actions/go-coverage` with `coverage_file: ./coverage.out`. That action recomputes `go tool cover -func="$COVERAGE_FILE"` on the **unfiltered** profile (`go-coverage/action.yml:65`), and its final `Enforce Threshold` step does `exit 1` (`:180-185`) — ungated, no `continue-on-error`.

The threshold is genuinely 80, traced end to end: `.settings.yaml:156 coverage_threshold: '80'` → `load-versions/action.yml:209` → `qualification.yaml:63` → `go-test/action.yml:79`. The action's `THRESHOLD=70` fallback is **not** taken. Both `merge-gate.yaml:174` and `on-push.yaml:54` pass `coverage_report: true`, and merge-gate's dorny/paths-filter (`predicate-quantifier: every`, `['**','!**.md','!docs/**','!LICENSE']`) matches `Makefile`, so the `tests` job does run for this PR.

Measured independently three times at this head:

```
unfiltered (what CI gates on):        total: (statements) 75.8%   → FAIL vs 80
filtered   (what make test-coverage): total: (statements) 80.2%   → pass
```

On main today, validators are never compiled into the profile at all, so main's `coverage.out` *is* the filtered set — CI passes at 80.2% with ~0.2pt of headroom. This PR pushes it 4.4pt under.

**Blast radius:** The `tests` job fails on this PR and on every subsequent PR/main push until reverted. It also breaks the repo's stated invariant ("if `make qualify` passes locally, CI will pass") in the most confusing direction: local `make qualify` goes green at 80.2% while the required check goes red at 75.8%, with no local way to reproduce.

**Fix:** Put the exclusion where CI actually reads it. Cleanest is to filter at profile-production time so every consumer (gate, badge, artifacts, delta comment) sees one number — in the `test` target:

```make
GOFLAGS="-mod=vendor" go test ... -coverprofile=coverage.full.out $$(go list ./... | grep -v -e /tests/chainsaw/) || exit 1; \
grep -v '^github.com/NVIDIA/aicr/validators/' coverage.full.out > coverage.out; \
```

…and revert this line to `go tool cover -func=coverage.out`. Alternatively add a `coverage_exclude` input to `.github/actions/go-coverage` and source the pattern from `.settings.yaml` so the Makefile and the action cannot drift. Either way, confirm the merge-gate `Test` job is green before merging — it has not run on this PR yet (only the labelers and CodeRabbit have).

njhensley: **🟡 Minor — Exemption is unbounded: no floor, no expiry, no tracking issue**

Per-package coverage today: `validators` 41.2%, `chainsaw` 73.3%, `conformance` 52.6%, `deployment` 58.3%, `helper` 10.8%, `internal/allocmode` 92.2%, `performance` 56.5%. Nothing now prevents that from decaying — the gated number literally cannot move for validator code, which is the part of AICR that runs unattended on GPU clusters.

**Blast radius:** Long-term erosion of validator test quality. Undocumented carve-outs tend to become permanent.

**Fix:** To be clear about proportionality: **"exempt now, ratchet later" is the right call for a two-line CI fix.** Don't lower the global threshold (that degrades the gate for all of `pkg/`) and don't write tests for 7 packages here. Just bound it — a follow-up issue for a separate lower validators floor (~4 lines of make: `grep '/validators/' coverage.out | go tool cover -func=/dev/stdin`, compared against e.g. 50) converts an open-ended carve-out into a ratchet.

njhensley: **🟡 Minor — Unanchored grep pattern will silently swallow future `*/validators/*` packages**

`grep -v '/validators/'` matches the segment anywhere in the import path. Zero false positives and zero false negatives today (verified in both directions), and the `mode: atomic` header is correctly preserved. But a future `internal/validators/` or `pkg/x/validators/` package would drop out of the gate with no error and no signal.

**Blast radius:** Future maintainers — a real coverage regression in a newly added `*/validators/*` package would not be caught.

**Fix:** Use `grep -v '^github.com/NVIDIA/aicr/validators/'` — verified to select the identical line set today. Credit where due: the current pattern correctly does **not** catch the singular `pkg/validator/**` tree, which was the likelier accidental-exclusion trap.

Makefile
```diff
@@ -226,13 +226,13 @@ test: test-shell ## Runs unit tests with race detector and coverage (use -short
...
 	echo "Coverage: $$coverage% (threshold: $(COVERAGE_THRESHOLD)%)"; \
```
njhensley, line 254: **🟡 Minor — Pre-existing fail-open in the coverage gate (informational — your change improves one case)**

`test-coverage` has no `set -o pipefail`. On a corrupt profile `$coverage` ends up empty, `bc` emits a parse error, `[ -eq 1 ]` errors with 'unary operator expected', the `if` body is skipped, and the recipe prints "Coverage check passed" and exits 0. Reproduced in a scratch Makefile on **both** the pre- and post-change recipes, so this is pre-existing and not introduced here. Your change incidentally *improves* one case: a missing `coverage.out` now fails **closed** (empty stdin → `total: 0.0%` → gate errors) where it previously failed open.

**Blast radius:** Local `make qualify` / `make test-coverage` only — CI is unaffected because `.github/actions/go-coverage` uses `set -euo pipefail` and explicitly checks for an empty `$COVERAGE`.

**Fix:** Not yours to fix in this PR. Worth a drive-by follow-up: `if [ -z "$$coverage" ]; then echo "ERROR: failed to compute coverage"; exit 1; fi` after the assignment, mirroring what the CI action already does.

Makefile
```diff
@@ -226,13 +226,13 @@ test: test-shell ## Runs unit tests with race detector and coverage (use -short
 	echo "Running tests with race detector..."; \
 	KUBEBUILDER_ASSETS=$$(setup-envtest use -p path 2>/dev/null || echo "") \
 	AICR_CRITERIA_STRICT=1 \
-	GOFLAGS="-mod=vendor" go test -short -count=1 -race -timeout=$(TEST_TIMEOUT) -covermode=atomic -coverprofile=coverage.out $$(go list ./... | grep -v -e /tests/chainsaw/ -e /validators) || exit 1; \
+	GOFLAGS="-mod=vendor" go test -short -count=1 -race -timeout=$(TEST_TIMEOUT) -covermode=atomic -coverprofile=coverage.out $$(go list ./... | grep -v -e /tests/chainsaw/) || exit 1; \
```
njhensley: **🟡 Minor — No guard that the revived tests stay revived (unmet acceptance criterion of #1752)**

#1752 lists verbatim: *"Add a guard proving the tests from #1748 and #1745 are discovered and executed."* The fix here is a negative-space change — deleting a grep term — and the exact regression class it repairs went unnoticed for four months. Nothing asserts that `validators/deployment/nvidia_smi_test.go` (#1748) and `validators/performance/inference_perf_test.go` (#1745) are in the executed package set.

**Blast radius:** The same silent-exclusion regression can recur; the new coverage filter is a second instance of the same footgun.

**Fix:** `test-shell` (Makefile:224) already runs `tools/*_test.sh` and is a natural home for a ~15-line hermetic assertion that the executed package set contains `github.com/NVIDIA/aicr/validators/deployment` and `.../validators/performance`. Low-probability regression, so reasonable to defer — but since the PR says `Fixes #1752`, either add it or re-scope the issue rather than letting it auto-close with 2 of 5 criteria unmet.

Makefile
```diff
@@ -226,13 +226,13 @@ test: test-shell ## Runs unit tests with race detector and coverage (use -short
...
 	echo "Test coverage:"; \
 	go tool cover -func=coverage.out | tail -1
```
njhensley, line 235: **🟡 Minor — Two divergent coverage numbers for the same commit; decide deliberately which one is published**

This line still prints the unfiltered total, so a single `make test-coverage` run emits `Test coverage: 75.8%` followed two lines later by `Coverage: 80.2% (threshold: 80%)` — a 4.4-point contradiction with no explanation, which reads like a computation bug. The same raw `coverage.out` also feeds the shields.io badge gist (75.8% flips the color bucket from `brightgreen` (>=80) to `green` (>=70)), the `coverage-pr`/`coverage-baseline` artifacts, and the `go-coverage-report` delta comment.

**Blast radius:** Contributor confusion locally, plus a ~4.4pt drop in the publicly reported coverage badge on main that reflects no real regression.

**Fix:** Filter once at profile-production time (see the blocker) so this line, the gate, the badge, and the delta report all agree. If the filter must stay in `test-coverage`, at minimum label the two numbers explicitly, e.g. `Total (incl. validators, ungated): …` vs `Gated (excl. validators): …`. The point is to make it a deliberate choice rather than an accident.

Makefile
```diff
@@ -226,13 +226,13 @@ test: test-shell ## Runs unit tests with race detector and coverage (use -short
...
 
 .PHONY: test-coverage
 test-coverage: test ## Runs tests and enforces coverage threshold (from .settings.yaml quality.coverage_threshold)
```
njhensley: **🔵 Nitpick — Carve-out is undocumented, and `docs/contributor/tests.md` newly says something false**

No Makefile comment explains the exclusion — which is exactly how the original `-e /validators` survived unexamined for four months. Separately, the word **"project-wide"** in `docs/contributor/tests.md:94` (and the mirrored `AGENTS.md`/`.claude/CLAUDE.md:585`) becomes factually wrong once validators are carved out.

**Blast radius:** Maintainability — a future contributor either "fixes" the filter (re-breaking the gate) or widens it without realizing it is a deliberate carve-out.

**Fix:** Add a comment naming the reason, the tracking issue, and the removal condition. Fix "project-wide" in `docs/contributor/tests.md:94` and `AGENTS.md`/`CLAUDE.md:585` (those two must change together — AGENTS.md is a CI-enforced mirror). Note the `75%`/`70%`-vs-actual-`80` drift in those same lines plus `DEVELOPMENT.md:458` is entirely **pre-existing** — flagging it, not asking you to fix it here.

njhensley (changes requested):
📋 **Multi-persona review — PR #1918**

▎ Method: 3 persona passes (Correctness/Shell-Make, CI-DX/Operability, Test-Strategy/Domain), each finding independently confirmed or refuted by a senior meta-reviewer against the resolved code. Measurements reproduced independently. Line links pinned to head `95040365`.

## Overall assessment — Request changes

The diagnosis is right and valuable: `-e /validators` was hiding 54 test files across 7 packages, and that exclusion was never deliberate — it was collateral in the 60-file refactor `b741e99e` (#290), with no rationale in the commit message and 8 validator test files already present at that commit. The revived tests are in good shape: all 7 packages pass under `-race`, hermetically, adding only ~60–100s to a 15-minute job budget.

The problem is that the mitigation is in the wrong layer. The coverage exclusion was added to `make test-coverage`, but **CI never runs that target** — `grep -rn 'test-coverage' .github/` returns zero hits. CI runs `make test` and then re-computes the threshold from the raw `coverage.out` via `.github/actions/go-coverage`, which hard-fails below 80. So as written this PR ships the coverage regression *without* the mitigation and turns the merge-gate `Test` job red. Details inline on `Makefile` L235.

## 🔵 Nitpick — PR title typo `covergae` will land in main's history

The commit subject at `95040365` is spelled correctly, but the repo squash-merges (verified: recent `main` subjects all end in `(#NNNN)`, GitHub's PR-title-derived format), so the PR **title** becomes the permanent commit subject. Suggest retitling to `fix(ci): run validators/ unit tests in make test, exclude from coverage gate` (68 chars, within the 70-char limit).

---

- **The revived tests are healthy.** All 7 packages `ok` under `-race -short`; re-run with `KUBECONFIG=/dev/null HOME=/tmp/nohome` still all `ok`. External interactions are `httptest.NewServer` and fake clientsets only — no cluster, envtest, network, or GPU dependency. `make test` is the correct home for them.
- **No timeout risk.** `TEST_TIMEOUT=10m` is a per-test-binary limit, not aggregate; the slowest new package is `validators/conformance` at ~30s, against a 15-minute job budget.
- **`validators/chainsaw` is not swallowed** by the surviving `-e /tests/chainsaw/` filter — its import path lacks the required `/tests/` segment. The filter is correctly specific.
- **`go tool cover -func=/dev/stdin` works correctly** — byte-identical output to reading a file, reads fine from a non-seekable pipe, and `mode: atomic` survives the grep. Verified on darwin/arm64 go1.26.5. (Linux-runner behavior unverified — no Docker in the review sandbox — but `/dev/stdin` → `/proc/self/fd/0` is standard on `ubuntu-latest`.)
- **Make escaping is correct** — `$$3`, `$$(...)`, the single-quoted grep pattern, `.PHONY`, and the `@` prefix are all fine.
- **Refuted during meta-review:** a reviewer claimed #1752's *"include validator packages in coverage reporting"* criterion was unmet. Not true — validators **are** in `coverage.out` now and in every downstream reporting consumer. Only the *threshold gate* excludes them.
- **Refuted during meta-review:** a reviewer claimed the first `go-coverage-report` delta comment would be skewed by baseline-vs-PR asymmetry. Not for this PR — the delta step is gated on changed `.go` files, and this PR changes none, so no delta comment fires at all.

---

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 1 | Coverage exclusion sits in a target CI never runs → merge gate fails at 75.8% vs 80 |
| 🟠 Major | 0 | — |
| 🟡 Minor | 5 | Unbounded exemption; unanchored grep; no anti-regression guard; two divergent coverage numbers; pre-existing fail-open |
| 🔵 Nitpick | 2 | Undocumented carve-out + newly-false "project-wide"; PR title typo |

**Recommendation:** The blocker must be fixed before merge — move the exclusion to where CI reads it, then confirm the merge-gate `Test` job is actually green (it has not run yet on this PR). The unanchored-grep fix is a one-token change worth folding in. The remaining minors and both nitpicks are fine to defer in writing, but the unbounded exemption should get a tracking issue before the carve-out becomes permanent by default.

Thanks for digging out a four-month-old silent test gap — the underlying find here is a good one.

tools/validators_tests_included_test.sh
```diff
@@ -0,0 +1,19 @@
...
+# asserts the package set make test runs still includes the packages
+# carrying the #1745/#1748 regression suites.
+set -euo pipefail
+cd "$(dirname "$0")/.."
+
+packages=$(GOFLAGS="-mod=vendor" go list ./... | grep -v -e /tests/chainsaw/)
+
+for pkg in \
+  github.com/NVIDIA/aicr/validators/deployment \
+  github.com/NVIDIA/aicr/validators/performance \
```
njhensley: **🔴 Blocker — The inclusion guard is broken, and (my scoping) can't catch what it's named for — recommend removing it**

**Context:** this script exists because of my prior finding F7. I'd flagged the missing #1752 guard as *Minor* and "reasonable to defer," and proposed a `tools/*_test.sh` asserting the package set contains the two validator packages. That suggestion was under-specified, and this is a faithful implementation of it — so treat this as a correction of my own advice as much as of the code.

Two problems, both pointing at removal:

1. **It fails CI as written.** Line 12 ends with a trailing `\`, which line-continues into `do`, so bash reads `for pkg in deployment performance do` and hits `if` where `do` was expected:
```
$ bash -n tools/validators_tests_included_test.sh
tools/…:14: syntax error near unexpected token `if'   (rc=2)
```
`test-shell` auto-runs every `tools/*_test.sh` under `set -e` and is a prerequisite of `make test`, so this aborts the Test job before any Go test runs — the same class of red gate this PR set out to fix.

2. **Even fixed, it guards nothing.** It hardcodes its *own* copy of the filter (`go list ./... | grep -v -e /tests/chainsaw/`) rather than exercising the Makefile's `test` recipe. So the regression it is named for — someone re-adding `-e /validators` to the **Makefile** — would leave this test green. It only fails if a validator package is deleted/renamed outright, which other checks already surface loudly. It asserts a constant it defined two lines earlier.

**Blast radius:** CI: the syntax error aborts `test-shell` → `make test` → the merge-gate Test job, on this PR and every push, until fixed. Separately, as a guard it provides false assurance against the exact silent-exclusion regression it advertises.

**Fix:** **Delete the file.** The #1752 "add a guard" criterion isn't load-bearing now that validators run and appear in the coverage profile, and a guard that gives false assurance is worse than none. If you'd rather keep the checkbox closed, the only version that earns its place asserts the artifact the recipe actually produces:
```bash
# coverage.full.out is the UNfiltered profile `make test` writes
for pkg in validators/deployment validators/performance; do
  grep -q "^github.com/NVIDIA/aicr/$pkg/" coverage.full.out \
    || { echo "FAIL: $pkg not in coverage profile — did make test stop running it?"; exit 1; }
done
```
That catches a Makefile filter re-adding `-e /validators` (the packages vanish from `coverage.full.out`); the current script can't. Either way, run `make test-shell` before pushing.

Makefile
```diff
@@ -220,13 +220,19 @@ license-check: ## Check license is approved
 test-shell: ## Runs shell unit tests (tools/*_test.sh; hermetic, no cluster)
 	@set -e; for t in tools/*_test.sh; do [ -e "$$t" ] || continue; echo "Running $$t..."; bash "$$t"; done
 
+# validators/ tests run as part of `make test` but are excluded from the
+# coverage.out this target emits: per-package coverage there runs 41-92%
+# (see #1752), which would pull the project-wide gate from ~80% to ~75.8%.
+# Tracked for a separate, lower validators coverage floor in #<ISSUE>.
```
njhensley: **🟡 Minor — `#<ISSUE>` is a literal unfilled placeholder**

`# Tracked for a separate, lower validators coverage floor in #<ISSUE>.` This is the deferred ratchet from prior finding F3, but the reference is a dangling placeholder — a literal `#<ISSUE>` would land in `main`.

**Blast radius:** A permanent placeholder in the tree; readers can't find the tracking issue because there isn't one.

**Fix:** Either file the tracking issue and drop in the real number, or remove the sentence. (This is also where a *real* coverage assertion on validators would eventually live, which is the honest reason the guard above can only check presence today — validators sit below the 80% bar and are carved out, so there's no enforced number to assert yet.)

Makefile
```diff
@@ -220,13 +220,19 @@ license-check: ## Check license is approved
...
+# Remove the grep below once that floor lands.
 .PHONY: test
 test: test-shell ## Runs unit tests with race detector and coverage (use -short to skip integration tests)
 	@set -e; \
 	echo "Running tests with race detector..."; \
 	KUBEBUILDER_ASSETS=$$(setup-envtest use -p path 2>/dev/null || echo "") \
 	AICR_CRITERIA_STRICT=1 \
-	GOFLAGS="-mod=vendor" go test -short -count=1 -race -timeout=$(TEST_TIMEOUT) -covermode=atomic -coverprofile=coverage.out $$(go list ./... | grep -v -e /tests/chainsaw/ -e /validators) || exit 1; \
+	GOFLAGS="-mod=vendor" go test -short -count=1 -race -timeout=$(TEST_TIMEOUT) -covermode=atomic -coverprofile=coverage.full.out $$(go list ./... | grep -v -e /tests/chainsaw/) || exit 1; \
```
njhensley, line 232: **🔵 Nitpick — New `coverage.full.out` isn't cleaned or ignored**

`make clean` (line 650) removes `./coverage.out` but not the newly-introduced `coverage.full.out`, and `.gitignore` covers neither.

**Blast radius:** A run leaves an untracked artifact in the working tree.

**Fix:** Add `coverage.full.out` to the `clean` target's `rm` list (and optionally to `.gitignore`).

njhensley (changes requested):
📋 **Re-review — PR #1918**

▎ Method: follow-up pass on the deltas since the prior multi-persona review, verified directly against head `a2edf95b`.

## Overall assessment — Request changes

Thanks for the fast turnaround — the substance of the prior review is addressed well. The original blocker is genuinely fixed: the filter moved into the `test` target so CI now gates on the validators-excluded `coverage.out`, and the docs/comment cleanups all landed. One item still blocks merge, and it's one I need to partly walk back: the anti-regression guard was **my suggestion** (prior finding F7, which I'd tiered *Minor / reasonable to defer*), and as I scoped it, it can't actually do its job. Recommend deleting it — details inline on `tools/validators_tests_included_test.sh`.

### ✅ Prior 🔴 Blocker — RESOLVED (verified)

`Makefile` now writes the unfiltered profile to `coverage.full.out`, then `grep -v '^github.com/NVIDIA/aicr/validators/' coverage.full.out > coverage.out`. Since CI's `go-coverage` action reads `./coverage.out`, it gates on the filtered ~80% number — the correct layer. I confirmed the anchored pattern preserves the `mode: atomic` header. This also closes prior **F2** (`make test`, the badge, and the artifacts now all read the same filtered `coverage.out`), **F4** (anchored `^github.com/NVIDIA/aicr/validators/`), and **F6** (docs no longer claim "project-wide").

---

## ✅ Confirmed non-issues / resolved since last pass

- Coverage-layer fix verified correct (see above) — the headline concern is fully addressed.
- Docs: `.claude/CLAUDE.md`, `AGENTS.md`, and `docs/contributor/tests.md` all updated to note the validators exclusion; the false "project-wide" wording is gone.
- (Superseded) the prior "missing trailing newline" nit on the guard file is moot if the file is deleted.

---

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 1 | Inclusion guard breaks CI and doesn't test its named regression → delete (our own F7, reframed) |
| 🟠 Major | 0 | — |
| 🟡 Minor | 1 | `#<ISSUE>` placeholder unfilled |
| 🔵 Nitpick | 1 | `coverage.full.out` not cleaned/ignored |

**Recommendation:** The coverage fix is right and done. The only thing blocking merge is the guard I asked for — simplest path is to **delete `tools/validators_tests_included_test.sh`** and either fill in or drop the `#<ISSUE>` line, at which point this is an easy approve. (The PR-title typo `covergae` → `coverage` is also still worth fixing before squash-merge.)

mohityadav8:
gh pr comment 1918 --body "@coderabbitai review"

njhensley (approved):

Issue #1246, opened 2026-06-09
Backfill CRD Established=True for 6 operator-owning health checks


mchmarny:
## Summary

Follow-up to PR #1245 (closed #1222, closed epic #660). Add
`Established=True` assertions for the CRDs owned by 6 operator
components that PR #1245 deferred because each needs chart-version
verification.

## Background

PR #1245 added CRD `Established=True` assertions for 4 operator-owning
components where the CRD names were verifiable from in-tree sources
(`tests/chainsaw/ai-conformance/cluster/assert-crds.yaml` and the Go
`verifyClusterPolicyReady` check):

- `gpu-operator` → `clusterpolicies.nvidia.com`
- `dynamo-platform` → `dynamocomponentdeployments.nvidia.com`
- `nodewright-operator` → `skyhooks.skyhook.nvidia.com`
- `cert-manager` → `certificates.cert-manager.io`,
  `issuers.cert-manager.io`, `clusterissuers.cert-manager.io`

Six operators that own CRDs were left out because their CRD lists
aren't enumerated in-tree at the pinned chart version. Bundling them
blindly would risk introducing typo'd CRD names that fail on live
clusters — better to verify each chart's actual output first.

## Components to enhance

| Component | Pinned chart | Expected CRD group |
|---|---|---|
| `gatekeeper` | (see `recipes/registry.yaml`) | `gatekeeper.sh`, `templates.gatekeeper.sh`, `config.gatekeeper.sh`, `expansion.gatekeeper.sh`, `mutations.gatekeeper.sh`, `status.gatekeeper.sh` |
| `kueue` | (see `recipes/registry.yaml`) | `kueue.x-k8s.io` (ClusterQueue / LocalQueue / ResourceFlavor / Workload / ProvisioningRequestConfig / AdmissionCheck / Topology / Cohort / MultiKueueCluster / MultiKueueConfig / WorkloadPriorityClass) |
| `kubeflow-trainer` | (see `recipes/registry.yaml`) | `trainer.kubeflow.org` (TrainJob / TrainingRuntime / ClusterTrainingRuntime / JobSet) |
| `network-operator` | (see `recipes/registry.yaml`) | `mellanox.com` / `nvidia.com` (NicClusterPolicy / MacvlanNetwork / IPoIBNetwork / HostDeviceNetwork) |
| `k8s-nim-operator` | (see `recipes/registry.yaml`) | `apps.nvidia.com` (NIMService / NIMCache / NIMPipeline) |
[...]

PR #1919, opened 2026-07-26
feat(recipes): backfill CRD Established=True for 6 operator-owning health checks

mohityadav8:
## Summary
Backfills CRD `Established=True` assertions for the 6 operator-owning health checks that PR #1245 deferred pending chart verification.

## Motivation / Context
PR #1245 added CRD `Established=True` checks for 4 components whose CRD names were verifiable in-tree (gpu-operator, dynamo-platform, nodewright-operator, cert-manager). It explicitly deferred gatekeeper, kueue, kubeflow-trainer, network-operator, k8s-nim-operator, and slinky-slurm-operator because their CRD names weren't enumerated in-tree at the pinned chart version - bundling them blind risked typo'd names failing on live clusters.

Related: #1245

## Type of Change
- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected
- [ ] Other: ____________

## Implementation Notes
Per component, in `recipes/checks/<comp>/health-check.yaml`:

| Component | CRD(s) asserted | Verification source |
|---|---|---|
| kueue | `resourceflavors/clusterqueues/localqueues.kueue.x-k8s.io` | `recipes/components/kueue/manifests/*.yaml` already ship CRs at this apiVersion - group is load-bearing today, not a guess |
| kubeflow-trainer | `trainjobs.trainer.kubeflow.org`, `trainingruntimes.trainer.kubeflow.org`, `clustertrainingruntimes.trainer.kubeflow.org` | all asserted Established=True. Verified against chart 2.2.0 via `helm template oci://ghcr.io/kubeflow/charts/kubeflow-trainer --version 2.2.0 --include-crds` (all three ship; jobsets.jobset.x-k8s.io also ships but is correctly not asserted - not load-bearing for this recipe). |
| network-operator | `nicclusterpolicies.mellanox.com` | `recipes/components/network-operator/manifests/nic-cluster-policy-aks.yaml` ships an actual `NicClusterPolicy` CR at `mellanox.com/v1alpha1` |
| k8s-nim-operator | `nimservices/nimcaches/nimpipelines.apps.nvidia.com` | `docs/conformance/cncf/v1.35/nim-eks/evidence/robust-operator.md` - live `kubectl get crds` output from an EKS run of chart 3.1.0 |
| gatekeeper | `constrainttemplates.templates.gatekeeper.sh`, `configs.config.gatekeeper.sh` | both asserted Established=True. Verified against chart 3.22.2 via `helm template gatekeeper/gatekeeper --version 3.22.2 --include-crds` - both present among 17 total CRDs; the feature-conditional mutation/expansion/external-data CRDs also render under default values, confirming these two unconditional core CRDs are the correct minimal assertion set. |
| slinky-slurm-operator | none - negative finding | Chart is deployed with `crds.enabled: false` (`recipes/registry.yaml`); all 6 `slinky.slurm.net` CRDs are owned exclusively by the sibling `slinky-slurm-operator-crds` chart, already covered by its own health check. Comment added explaining the omission per #1246's acceptance criteria. |

For network-operator, `MacvlanNetwork`/`IPoIBNetwork`/`HostDeviceNetwork` from the issue's guess table are left out pending confirmation.

## Testing
```bash
go test -race -count=1 ./validators/chainsaw/...
python -m yamllint -c .yamllint.yaml recipes/checks/gatekeeper/health-check.yaml recipes/checks/k8s-nim-operator/health-check.yaml recipes/checks/kubeflow-trainer/health-check.yaml recipes/checks/kueue/health-check.yaml recipes/checks/network-operator/health-check.yaml recipes/checks/slinky-slurm-operator/health-check.yaml
TestValidateTestReadOnly_RegistryContent: pass

TestRunChainsawTestInProcess_RegistryCorpusParses: pass

yamllint: clean, no output

Live-cluster validation not yet performed for any of the 6 - see Risk Assessment.

Risk Assessment
[x] Medium - Touches multiple components; CRD names for kueue/kubeflow-trainer/network-operator/k8s-nim-operator and gatekeeper are verified against real in-tree evidence or chart renders, and none of the 6 have been validated on a live cluster yet.

Rollout notes: Per-component revert is trivial (each is an isolated try step).

Checklist
[x] Tests pass locally (make test with -race)

[x] Linter passes (make lint)

[x] I did not skip/disable tests to make CI green

[ ] I added/updated tests for new functionality - no new tests added; existing TestComponentRegistry_RequiresHealthCheck + TestValidateTestReadOnly_RegistryContent lint guards already cover the registry content this PR touches, same as #1245

[ ] I updated docs if user-facing behavior changed - N/A, no user-facing CLI/API change

[x] Changes follow existing patterns in the codebase

[x] Commits are cryptographically signed (git commit -S)

### DIFF
--- a/recipes/checks/gatekeeper/health-check.yaml
+++ b/recipes/checks/gatekeeper/health-check.yaml
@@ -15,10 +15,14 @@
 # Gatekeeper Health Check
 #
 # Validates that gatekeeper-controller-manager and gatekeeper-audit are
-# running and healthy in the gatekeeper-system namespace. Guards against
-# vacuous pass on empty namespace by asserting both Deployments exist with
-# at least one available replica, then asserts no pods are stuck in
-# Pending, Failed, or Unknown phases.
+# running and healthy in the gatekeeper-system namespace. Also asserts
+# the two CRDs the chart installs unconditionally in crds/ —
+# constrainttemplates.templates.gatekeeper.sh and
+# configs.config.gatekeeper.sh — are Established=True (verified against
+# chart 3.22.2; re-verify on defaultVersion bump). Guards against
+# vacuous pass on empty namespace by asserting both Deployments exist
+# with at least one available replica, then asserts no pods are stuck
+# in Pending, Failed, or Unknown phases.
 apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
@@ -27,6 +31,34 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      # Verified against chart 3.22.2 (recipes/registry.yaml): rendered
+      # `helm template gatekeeper/gatekeeper --version 3.22.2 --include-crds`
+      # and confirmed both constrainttemplates.templates.gatekeeper.sh and
+      # configs.config.gatekeeper.sh are present (17 CRDs total; the
+      # feature-conditional mutation/expansion/external-data CRDs also
+      # render under default values, confirming these two unconditional
+      # core CRDs are the correct minimal assertion set). Re-verify on
+      # defaultVersion bump — tracked systemically in #1948.
+      try:
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: constrainttemplates.templates.gatekeeper.sh
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: configs.config.gatekeeper.sh
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
     - name: validate-controller-manager-deployment
       try:
         - assert:
--- a/recipes/checks/k8s-nim-operator/health-check.yaml
+++ b/recipes/checks/k8s-nim-operator/health-check.yaml
@@ -15,9 +15,12 @@
 # NIM Operator Health Check
 #
 # Validates that the NVIDIA NIM Operator is running and healthy in the
-# nvidia-nim namespace. Checks that the k8s-nim-operator deployment has
-# at least one available replica and that no pods in the namespace are
-# stuck in Pending, Failed, or Unknown phases.
+# nvidia-nim namespace. Also asserts nimservices / nimcaches /
+# nimpipelines.apps.nvidia.com are Established=True (verified against
+# chart 3.1.0; re-verify on defaultVersion bump). Checks that the
+# k8s-nim-operator deployment has at least one available replica and
+# that no pods in the namespace are stuck in Pending, Failed, or
+# Unknown phases.
 apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
@@ -26,6 +29,35 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      try:
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: nimservices.apps.nvidia.com
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: nimcaches.apps.nvidia.com
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: nimpipelines.apps.nvidia.com
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
     - name: validate-deployment-exists
       try:
         # Guard against vacuous pass on empty namespace: verify the
--- a/recipes/checks/kubeflow-trainer/health-check.yaml
+++ b/recipes/checks/kubeflow-trainer/health-check.yaml
@@ -14,10 +14,17 @@
 
 # Kubeflow Trainer Health Check
 #
-# Validates that the Kubeflow Trainer is running and healthy in the kubeflow
-# namespace. Checks that the kubeflow-trainer-controller-manager deployment has at least one
-# available replica and that no pods in the namespace are stuck in Pending,
-# Failed, or Unknown phases.
+# Validates that the Kubeflow Trainer is running and healthy in the
+# kubeflow namespace. Also asserts trainjobs.trainer.kubeflow.org,
+# trainingruntimes.trainer.kubeflow.org, and
+# clustertrainingruntimes.trainer.kubeflow.org are Established=True
+# (verified against chart 2.2.0; re-verify on defaultVersion bump) —
+# clustertrainingruntimes is load-bearing because platform-kubeflow.yaml
+# applies a ClusterTrainingRuntime CR
+# (torch-distributed-cluster-training-runtime.yaml) via manifestFiles.
+# Checks that the kubeflow-trainer-controller-manager deployment has at
+# least one available replica and that no pods in the namespace are
+# stuck in Pending, Failed, or Unknown phases.
 apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
@@ -26,6 +33,40 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      # trainjobs + trainingruntimes + clustertrainingruntimes verified
+      # against chart 2.2.0 (recipes/registry.yaml). clustertrainingruntimes
+      # is the CRD platform-kubeflow.yaml actually instantiates a CR
+      # against — see the kubeflow-trainer entry's manifestFiles block
+      # in registry.yaml.
+      try:
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: trainjobs.trainer.kubeflow.org
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: clustertrainingruntimes.trainer.kubeflow.org
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: trainingruntimes.trainer.kubeflow.org
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
     - name: validate-deployment-exists
       try:
         # Guard against vacuous pass on empty namespace: verify the
--- a/recipes/checks/kueue/health-check.yaml
+++ b/recipes/checks/kueue/health-check.yaml
@@ -15,11 +15,14 @@
 # Kueue Health Check
 #
 # Validates that Kueue is running and healthy in the kueue-system
-# namespace, and that the default quota CRs shipped with the component
-# (ResourceFlavor default-flavor, ClusterQueue cluster-queue, LocalQueue
-# default/default) were admitted and reconciled to Active — a functional
-# proof (CRDs installed, controller reconciling, and flavor/queue resolution)
-# within the assert/error-only allowlist.
+# namespace. Asserts resourceflavors / clusterqueues /
+# localqueues.kueue.x-k8s.io are Established=True (verified against
+# chart 0.18.2; re-verify on defaultVersion bump), and that the default
+# quota CRs shipped with the component (ResourceFlavor default-flavor,
+# ClusterQueue cluster-queue, LocalQueue default/default) were admitted
+# and reconciled to Active — a functional proof (CRDs installed,
+# controller reconciling, and flavor/queue resolution) within the
+# assert/error-only allowlist.
 apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
@@ -28,6 +31,39 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      try:
+        # kueue.x-k8s.io group verified in-tree: the manifests this
+        # component ships (recipes/components/kueue/manifests/*.yaml)
+        # use apiVersion kueue.x-k8s.io/v1beta2 for ResourceFlavor,
+        # ClusterQueue, and LocalQueue. Chart 0.18.2 (recipes/registry.yaml).
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: resourceflavors.kueue.x-k8s.io
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: clusterqueues.kueue.x-k8s.io
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: localqueues.kueue.x-k8s.io
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
     - name: validate-deployment-exists
       try:
         # Guard against vacuous pass on empty namespace: verify the
--- a/recipes/checks/network-operator/health-check.yaml
+++ b/recipes/checks/network-operator/health-check.yaml
@@ -14,10 +14,17 @@
 
 # Network Operator Health Check
 #
-# Validates that the NVIDIA Network Operator is running and healthy in the
-# nvidia-network-operator namespace. Checks that the network-operator
-# deployment has at least one available replica and that no pods in the
-# namespace are stuck in Pending, Failed, or Unknown phases.
+# Validates that the NVIDIA Network Operator is running and healthy in
+# the nvidia-network-operator namespace. Also asserts
+# nicclusterpolicies.mellanox.com is Established=True (verified against
+# chart 26.4.1; re-verify on defaultVersion bump). Chart 26.4.1 also
+# ships nicnodepolicies.mellanox.com, deliberately NOT asserted here:
+# no in-tree recipe instantiates a CR against it, so it fails the same
+# load-bearing-only bar the rest of this PR's assertions are held to.
+# Checks that the
+# network-operator deployment has at least one available replica and
+# that no pods in the namespace are stuck in Pending, Failed, or
+# Unknown phases.
 apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:
@@ -26,6 +33,17 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      try:
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: nicclusterpolicies.mellanox.com
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
     - name: validate-deployment-exists
       try:
         # Guard against vacuous pass on empty namespace: verify the
--- a/recipes/checks/slinky-slurm-operator/health-check.yaml
+++ b/recipes/checks/slinky-slurm-operator/health-check.yaml
@@ -18,6 +18,13 @@
 # running and healthy in the `slinky` namespace. Checks that both
 # deployments have at least one available replica and that no pods in
 # the namespace are stuck in Pending, Failed, or Unknown phases.
+#
+# No CRD Established=True block here (negative finding for #1246): the
+# slinky-slurm-operator chart is deployed with crds.enabled: false (see
+# the `crds.enabled: false` key in
+# components/slinky-slurm-operator/values.yaml; registry.yaml's
+# slinky-slurm-operator entry only carries the same flag as a prose
+# reminder to re-verify it on version bumps).
 apiVersion: chainsaw.kyverno.io/v1alpha1
 kind: Test
 metadata:

### CODE REVIEW

**Comment by mohityadav8** [comment]
@coderabbitai full review

File: recipes/checks/gatekeeper/health-check.yaml [hunk 1]
```diff
@@ -27,6 +27,26 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
```
njhensley, line 34: **🟡 Minor — The promised inline TODO doesn't exist in the file**

The PR body says gatekeeper is "Flagged inline with a TODO to confirm via `helm template gatekeeper/gatekeeper --version 3.22.2`", and commit 594c9041 repeats "flagged inline". Lines 30-49 contain zero comments — the step is bare `assert:` blocks, and the file header (lines 15-21) was not amended either. kueue's new block (lines 33-36) *does* carry its provenance comment, so the pattern was known and simply omitted here.

**Blast radius:** Traceability only — I verified both names are correct (see the summary), so there is no runtime defect. The effect is that the PR's own stated merge precondition is unenforceable and evaporates once the PR description scrolls out of view.

**Fix:** Since the names are now verified against chart 3.22.2, the better fix is a provenance comment in kueue's style rather than a TODO — e.g. `# Verified against chart 3.22.2 (recipes/registry.yaml): both CRDs ship in the chart's unconditional crds/ dir, so no values toggle can suppress them. Re-verify on defaultVersion bump.` Alternatively, drop the claim from the PR body.

recipes/checks/kubeflow-trainer/health-check.yaml
```diff
@@ -26,6 +26,17 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-trainjob-crd-established
```
njhensley: **🟡 Minor — Omits clustertrainingruntimes — the one CRD AICR actually instantiates a CR against**

The check asserts only `trainjobs.trainer.kubeflow.org`. But `recipes/registry.yaml:638-646` documents that `recipes/mixins/platform-kubeflow.yaml` attaches `torch-distributed-cluster-training-runtime.yaml` — a **ClusterTrainingRuntime CR** — as `manifestFiles`, wrapped into the injected `kubeflow-trainer-post` release. So `clustertrainingruntimes.trainer.kubeflow.org` is load-bearing in AICR's actual deployment path, and it is precisely the CRD this check does not assert. `helm show crds oci://ghcr.io/kubeflow/charts/kubeflow-trainer --version 2.2.0` confirms the chart ships `clustertrainingruntimes`, `trainingruntimes`, `trainjobs`, `jobsets` and `leaderworkersets`.

**Blast radius:** Applying a ClusterTrainingRuntime CR against a not-yet-Established CRD is the concrete failure mode this backfill exists to catch, and the check as written would miss it — on the GKE/EKS `*-training-kubeflow` overlays and the `platform-kubeflow` mixin. Same shape at lower stakes for network-operator (1 of 17 CRDs asserted).

**Fix:** Add `clustertrainingruntimes.trainer.kubeflow.org` (and optionally `trainingruntimes`), renaming the step to `validate-crds-established` to match the repo's multi-CRD convention. registry.yaml:630-637 is the same class of in-tree evidence already used to justify the kueue names. At minimum, add an in-file note stating the subset is deliberate and why.

recipes/checks/k8s-nim-operator/health-check.yaml
```diff
@@ -26,6 +26,35 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
```
njhensley, line 32: **🟡 Minor — File headers left stale after the first step changed (5 files)**

The convention in the checks PR #1245 landed is that the file header enumerates every step the file performs: gpu-operator lines 18-21 list the CRD assert as numbered item 2, cert-manager lines 26-28 say "Also asserts the three core CRDs are Established=True...", nodewright-operator and dynamo-platform likewise as item 1. Here all five touched headers still describe only deployment-replica + pod-phase checks while the file's *first* step is now a CRD assert — k8s-nim-operator (15-20), gatekeeper (15-21), kubeflow-trainer (15-20), network-operator (15-20), and kueue (15-22, which got an inline block comment but no header update).

**Blast radius:** Documentation drift, zero runtime effect. The practical cost is on chart bumps: CRD plural names are chart-version-coupled, and nothing in-tree tells a Renovate bumper of k8s-nim-operator 3.1.0 / kubeflow-trainer 2.2.0 / network-operator 26.1.1 / gatekeeper 3.22.2 that these asserts must be re-verified.

**Fix:** Extend each file's header to list the new CRD step, naming the verification source and the pinned chart version — matching gpu-operator/cert-manager, and the "re-verify on chart bump" note in the slinky-slurm-operator-crds header (lines 21-29).

recipes/checks/gatekeeper/health-check.yaml
```diff
@@ -27,6 +27,26 @@ spec:
...
+                name: constrainttemplates.templates.gatekeeper.sh
+              status:
+                (conditions[?type == 'Established']):
+                  - status: "True"
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: configs.config.gatekeeper.sh
```
njhensley, line 58: **🔵 Nitpick — "cannot function without" is an imprecise filter for the gatekeeper CRD pair**

The PR body justifies the pair as "the two CRDs gatekeeper cannot function without, regardless of feature flags". That holds for `constrainttemplates.templates.gatekeeper.sh` (the ConstraintTemplate controller generates the per-constraint CRDs and is the core admission path), but is imprecise for `configs.config.gatekeeper.sh` — Gatekeeper enforces policy perfectly well with no Config CR present; the Config kind governs optional sync / exempt-namespace / match behavior.

**Blast radius:** None functionally — the assertion is harmless and passes, since the CRD ships unconditionally in the chart's `crds/` directory (which is also why `helm show crds` surfaces it: Helm never templates `crds/`, so no values toggle can suppress it). It matters only because this rationale is what a future maintainer will reuse when deciding which CRDs to assert for the next operator, and "cannot function without" is the wrong filter to inherit.

**Fix:** Reword to something like "the two CRDs the chart installs unconditionally in crds/" — which is both accurate and the criterion that actually generalizes. Fold into the header comment the finding above already asks for.

recipes/checks/kueue/health-check.yaml
```diff
@@ -134,4 +167,4 @@ spec:
                 name: default
                 namespace: default
               status:
-                ((conditions || `[]`)[?type == 'Active' && status == 'True'] | length(@) > `0`): true
+                ((conditions || `[]`)[?type == 'Active' && status == 'True'] | length(@) > `0`): true
```
njhensley: **🔵 Nitpick — Trailing newline dropped — looks like a half-applied fix**

The last byte of the file is `0x65` ('e') rather than `0x0a`; all five sibling check files end `0x0a`. The diff renders the final line as modified plus `\ No newline at end of file`, though its content is byte-identical. CI will not catch it: `.yamllint.yaml` sets `new-line-at-end-of-file: disable` ("Many existing files lack trailing newline") and there is no `.editorconfig`, so `yamllint -c .yamllint.yaml` exits 0.

**Blast radius:** Cosmetic. It puts a spurious modified line and blame entry on the `validate-quota-crs-active` assertion, which this PR did not intend to touch. Worth noting because it appears to be half-applied rather than unnoticed: commit 594c9041 stripped the trailing newline from both kueue and slinky-slurm-operator, and the follow-up 553d2368 restored slinky's but not kueue's.

**Fix:** Restore the trailing newline so the diff for this file is purely additive.

recipes/checks/slinky-slurm-operator/health-check.yaml
```diff
@@ -18,6 +18,14 @@
 # running and healthy in the `slinky` namespace. Checks that both
 # deployments have at least one available replica and that no pods in
 # the namespace are stuck in Pending, Failed, or Unknown phases.
+#
+# No CRD Established=True block here (negative finding for #1246): the
+# slinky-slurm-operator chart is deployed with crds.enabled: false (see
```
njhensley, line 23: **🔵 Nitpick — Negative-finding comment cites the weaker of two sources**

The comment says the chart "is deployed with crds.enabled: false (see recipes/registry.yaml, slinky-slurm-operator entry)". The actual setting is at `recipes/components/slinky-slurm-operator/values.yaml:29-30`; registry.yaml:703-708 mentions the flag only inside a prose reminder to re-verify it on version bumps. Not a dead pointer, just the weaker of the two sources.

**Blast radius:** Documentation precision only. I verified the substance of the negative finding is entirely correct: `crds.enabled: false` is set, and `recipes/checks/slinky-slurm-operator-crds/health-check.yaml` does assert Established=True for all six `slinky.slurm.net` CRDs (accountings / controllers / loginsets / nodesets / restapis / tokens). Good call to document the omission rather than leave it silent.

**Fix:** Point at `recipes/components/slinky-slurm-operator/values.yaml` as the primary citation, optionally keeping registry.yaml as the secondary reference for the re-verify-on-bump note.

njhensley (commented):
## Review method

Four parallel persona reviewers (Correctness, Operability/CI-DX, Domain & Architecture, Test-coverage), then an adversarial senior meta-review that re-derived every claim from the resolved code rather than the reviewer's assertion. 13 candidate findings: 2 suppressed as out-of-scope or noise, 5 re-tiered down, 2 net-new added.

## Overall assessment

**The risk this PR flags as unresolved is closable offline, and I closed it.** The body states gatekeeper's CRD names are "not yet verified in-tree" and that no component was live-validated. I rendered each chart at its exact `recipes/registry.yaml` pin and confirmed all 12 asserted names:

| Chart (pinned version) | Result |
|---|---|
| gatekeeper 3.22.2 | `constrainttemplates.templates.gatekeeper.sh`, `configs.config.gatekeeper.sh` ✅ |
| kueue 0.18.2 | `resourceflavors` / `clusterqueues` / `localqueues.kueue.x-k8s.io` ✅ |
| kubeflow-trainer 2.2.0 | `trainjobs.trainer.kubeflow.org` ✅ |
| network-operator 26.1.1 | `nicclusterpolicies.mellanox.com` ✅ |
| k8s-nim-operator 3.1.0 | `nimservices` / `nimcaches` / `nimpipelines.apps.nvidia.com` ✅ |

Every one of the 12 lives in its chart's unconditional `crds/` directory. Helm never templates `crds/`, so no values toggle can suppress them — which closes the name-typo risk *and* the feature-gating concern without a cluster. This is stronger evidence than the in-tree greps PR #1245 accepted for its four components, and it means the gatekeeper "confirm before merge" gate is already satisfied.

**Recommend pasting the rendered-CRD output into the Testing section**, since it is otherwise the only provenance for 12 hardcoded strings.

Everything that survived is documentation hygiene. Nothing blocks merge. The most substantive item is the `clustertrainingruntimes` omission on kubeflow-trainer, inline below.

The conservative-subset instinct and the slinky-slurm-operator negative finding are both right calls, and the per-component isolation makes revert trivial as claimed.

## Confirmed non-issues (examined and refuted)

Listing these so it's clear they were considered, not overlooked:

| Claim | Verdict |
|---|---|
| Missing RBAC on `customresourcedefinitions` | **Refuted** — the validator binds its per-run ServiceAccount to the built-in `cluster-admin` ClusterRole (`pkg/validator/job/rbac.go:83`); no verb can be missing |
| A typo'd CRD name hangs for the full 5m | **Refuted** — an entirely-absent resource returns `ErrCodeNotFound`, bounded by `AbsentResourceGracePeriod` = 30s (`pkg/defaults/timeouts.go:403`); it fails fast |
| The slice-literal assert is a CLAUDE.md anti-pattern | **Refuted as actionable** — it fails *closed* in all three degenerate cases (absent CRD, absent `status.conditions`, multi-element projection), and is byte-identical to the seven blocks #1245 already merged. Normalizing to the defensive `length(@) > 0` form would leave the corpus split between two idioms; that's a single sweeping PR across all 12+ blocks, not a comment here |
| Mixed step naming is convention drift | **Refuted** — `validate-crds-established` (multi-CRD) vs `validate-<x>-crd-established` (single-CRD) matches #1245's landed convention exactly |
| `Closes #1246` will orphan the deferred verification | **Refuted** — this repo squash-merges with `squash_merge_commit_message=BLANK`, so the commit-body trailer never reaches GitHub, and the PR description carries no closing keyword. #1246 will stay open, which is the desired outcome. Worth dropping the trailer from the commit body anyway, since it misstates intent |
| The 5m assert timeout starves downstream steps | **Mechanism confirmed, impact negligible** — it *is* a whole-Test shared budget (`validators/chainsaw/inprocess.go:88`), not per-step. But a healthy CRD assert is one by-name apiserver GET resolving in milliseconds (~0.01% of 300s), and an absent CRD fast-fails the whole Test in 30s, so no downstream step is starved of budget it would have used |
| gatekeeper and kueue are unreachable from any overlay/mixin | **True, but it lowers risk rather than raising it** — neither check executes in any resolved recipe today, so the residual live-cluster risk is currently unreachable. And `TestComponentRegistry_RequiresHealthCheck` mandates a health check per registry entry, so an unreferenced component with a check is the required steady state (`agentgateway-crds`, `prometheus-operator-crds` sit the same way), not dead weight |
| kueue's new CRD step is redundant with `validate-quota-crs-active` | **True in the pass case — keep it anyway.** A CR cannot be fetched unless its CRD is Established, so the existing step does imply it. But registry.yaml:599-606 applies the quota CRs post-install in wave N+1, so the CRD step is a strictly earlier signal on a fresh install; it disambiguates "CRDs never installed" from "controller not reconciling"; and it survives an overlay that overrides `manifestFiles` with differently-named queue CRs |
| Cluster-scoped CRD GET misdirected into the component namespace | **Refuted** — `validators/chainsaw/fetch.go:53-57` consults `mapping.Scope.Name()` and omits the namespace for cluster-scoped kinds |
| Inserting a step perturbs evidence / conformance output | **Refuted** — `pkg/evidence` has no step-level shape, and no Go test asserts on step counts or names |
| The 3-assert blocks consume extra `ChainsawMaxParallel` worker slots | **Refuted** — that bounds concurrent *components*; steps within a Test run strictly sequentially |

Two items were deliberately **not** filed: the `slinky-slurm-operator-crds` header saying chart "1.1.0" while registry.yaml:687 pins 1.2.0 is real drift, but `git blame` puts it in PR #1243 in a file this PR never opens — worth a separate drive-by. And a reviewer's inability to reproduce the body's `python -m yamllint` invocation is a local-environment artifact; the repo gate (`yamllint -c .yamllint.yaml`) exits 0 on all six files, so the substantive lint claim is true.

## Minor: the PR title undercounts the negative finding

The title and summary say "6 operator-owning health checks", but five files receive assertions and the sixth (slinky-slurm-operator) correctly receives a documented negative finding instead. The body's table is honest about this, but given this repo's `squash_merge_commit_title=PR_TITLE` / `message=BLANK` settings, **the title is the only text that survives the merge**. Consider "5 components + 1 documented negative finding".

## Worth a follow-up issue (not this PR's to fix)

No test can catch a wrong CRD name or malformed JMESPath. `TestValidateTestReadOnly_RegistryContent` only gates the read-only op allowlist. `TestRunChainsawTestInProcess_RegistryCorpusParses` runs each check against an *empty* fake fetcher, so every `Fetch` returns NotFound and `checks.Check` — where the JMESPath is compiled and evaluated — is never reached. No workflow invokes `make check-health`, and KWOK ships no CRD fixtures.

This is a real gap, but it is **identical to the one #1245's seven merged blocks already have** — pre-existing, and not this contributor's to close. A concrete fix exists using machinery already in the repo: `tools/bom` renders every registry chart, and `pkg/helm/render.go:147` only needs `--include-crds` flipped to cross-check every asserted CRD name against the chart-shipped set. All five chart repos are public and unauthenticated, so it is feasible — but it would make `make test` network-dependent, so its home is an opt-in job alongside `make bom-check` / `.github/workflows/bom-refresh.yaml`, not `make qualify`.

## Summary

```
🔴 Blocker 0 | 🟠 Major 0 | 🟡 Minor 3 | 🔵 Nitpick 4     Recommendation: Approve with comments
```

Solid, well-scoped work. The one change I'd actually push for is adding `clustertrainingruntimes` to the kubeflow-trainer check.

mohityadav8:
@njhensley i think everthing is solved in this

recipes/checks/gatekeeper/health-check.yaml
```diff
@@ -27,6 +31,29 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      # Verified against chart 3.22.2 (recipes/registry.yaml): both CRDs
```
njhensley: **🟡 Minor — gatekeeper comment claims "Verified against chart 3.22.2" but PR body admits it was never verified in-tree; promised inline TODO is absent**

Header (L21) and step comment (L35-37) both state the two gatekeeper CRDs are 'verified against chart 3.22.2 ... both CRDs ship in the chart's unconditional crds/ dir, so no values toggle can suppress them' — phrased as accomplished fact. The PR body's own table row for gatekeeper says '**Not yet verified in-tree.**' and promises 'a TODO to confirm via helm template gatekeeper/gatekeeper --version 3.22.2 before this is trusted at face value.' No TODO exists in the file.

**Blast radius:** A future maintainer reading the durable comment trusts a live-verification that never happened; if either name is wrong for 3.22.2 the check false-negatives. Both names are canonical upstream gatekeeper CRDs and the assert fails closed, so no runtime defect today — this is a comment/PR-body integrity issue.

**Fix:** Either perform the `helm template gatekeeper/gatekeeper --version 3.22.2` check and keep 'Verified', or soften the wording to 'asserted-but-unverified' and add the promised inline TODO so the comment matches the confidence stated in the PR body.

recipes/checks/gatekeeper/health-check.yaml
```diff
@@ -27,6 +31,29 @@ spec:
...
+      # ship in the chart's unconditional crds/ dir, so no values toggle
+      # can suppress them. Re-verify on defaultVersion bump.
+      try:
+        - assert:
+            resource:
+              apiVersion: apiextensions.k8s.io/v1
+              kind: CustomResourceDefinition
+              metadata:
+                name: constrainttemplates.templates.gatekeeper.sh
```
njhensley, line 49: **🟡 Minor — gatekeeper's two CRD names are the only asserted names with zero in-tree evidence backing**

constrainttemplates.templates.gatekeeper.sh and configs.config.gatekeeper.sh are asserted Established=True, but a repo-wide search finds no in-tree corroboration: no conformance evidence dump, no manifest instantiating a ConstraintTemplate/Config CR, and components/gatekeeper/values.yaml never references either group. Every other check in this PR is anchored by real evidence (kueue manifests, kubeflow assert-crds.yaml + Go lifecycle code, network-operator NicClusterPolicy CR, NIM live kubectl dump).

**Blast radius:** gatekeeper health check only; a wrong plural/group fails closed on a healthy gatekeeper install (false negative). Practical risk low — both are canonical upstream names installed unconditionally by the chart's crds/ dir. This is a verification-gap/fragility finding, not a confirmed defect.

**Fix:** Verify both names against the open-policy-agent/gatekeeper 3.22.2 chart crds/ and record the corroboration (a conformance reference or an assert-crds fixture) so the assertion has the same in-tree anchor as its five siblings. Tracked with F2 in #1948.

recipes/checks/gatekeeper/health-check.yaml
```diff
@@ -27,6 +31,29 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      # Verified against chart 3.22.2 (recipes/registry.yaml): both CRDs
+      # ship in the chart's unconditional crds/ dir, so no values toggle
```
njhensley: **🟡 Minor — No automated guard couples the "verified against chart X.Y.Z" comments / asserted CRD set to registry defaultVersion (tracked in #1948)**

Each of the 6 checks hardcodes 'verified against chart X.Y.Z; re-verify on defaultVersion bump'. The only tests over these files (validators/chainsaw: TestValidateTestReadOnly_RegistryContent, TestRunChainsawTestInProcess_RegistryCorpusParses) check allowlist compliance + YAML parse only; the BOM freshness guard gates docs, not recipes/checks. So a registry defaultVersion bump triggers no failure — the 'verified against' comment silently goes stale and a renamed/removed CRD assert becomes a false negative.

**Blast radius:** Version bumps ship with stale 'verified against' comments and possibly-wrong CRD names, surfacing only as a live-cluster health failure at validate time, not at PR time. This is a pre-existing property of the already-merged #1245 pattern (gpu-operator carries the same uncoupled comment), so it is not newly broken by this PR.

**Fix:** Out of scope for this PR — tracked in #1948. A Go guard mirroring TestCommittedBOMVersionsMatchRegistry could re-parse the 'verified against chart X.Y.Z' token and assert it equals the registry defaultVersion for that component.

recipes/checks/kubeflow-trainer/health-check.yaml
```diff
@@ -26,6 +33,39 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
```
njhensley, line 36: **🟡 Minor — CRD-established step runs first and no name was live-validated — a wrong name fails the whole per-component check only after the 5m assert timeout**

validate-crds-established is the first step (before deployment/pod checks) and spec.timeouts.assert is 5m. An assert against a non-existent CRD name retries until the 5m timeout, then fails the entire chainsaw Test. Because the PR states none of the 6 names were live-validated, a typo/rename reports the component unhealthy even when the operator is fully functional.

**Blast radius:** Per-component only — each health-check.yaml is an independent chainsaw Test, so one bad file fails just its own component's gate and per-file revert is trivial. Cost is a 5m stall plus a confident false 'unhealthy' verdict for that component.

**Fix:** Live-validate the CRD names before merge (the same gate #1245 used), or lower the assert timeout for the CRD-established step so a wrong name fails fast rather than after 5m.

recipes/checks/kubeflow-trainer/health-check.yaml
```diff
@@ -26,6 +33,39 @@ spec:
...
+      # trainjobs + trainingruntimes + clustertrainingruntimes verified
+      # against chart 2.2.0 (recipes/registry.yaml). clustertrainingruntimes
+      # is the CRD platform-kubeflow.yaml actually instantiates a CR
+      # against — see registry.yaml:638-646.
```
njhensley: **🟡 Minor — Comments cite volatile line-number references (registry.yaml:638-646, values.yaml:29-30) that rot like the SHA refs CLAUDE.md discourages**

The kubeflow comment cites 'registry.yaml:638-646' and the slinky comment cites 'components/slinky-slurm-operator/values.yaml:29-30'. Both resolve correctly today (638-646 is the platform-kubeflow ClusterTrainingRuntime/manifestFiles block; 29-30 is crds:/enabled: false), but line numbers shift on any insertion above them. CLAUDE.md's rationale against SHA refs in durable artifacts ('refs get rewritten') applies identically to line numbers in durable YAML comments.

**Blast radius:** Maintainability only — a future reader follows a stale range to unrelated content. No runtime effect.

**Fix:** Reference stable anchors by symbol instead: 'the kubeflow-trainer entry's manifestsUseChartCRDs / manifestFiles in registry.yaml' and 'the crds.enabled: false key in components/slinky-slurm-operator/values.yaml'.

recipes/checks/network-operator/health-check.yaml
```diff
@@ -26,6 +29,17 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-nicclusterpolicy-crd-established
```
njhensley: **🔵 Nitpick — Inconsistent CRD-step naming: network-operator uses validate-nicclusterpolicy-crd-established vs validate-crds-established elsewhere**

gatekeeper, k8s-nim-operator, kubeflow-trainer, and kueue all name the CRD step 'validate-crds-established'; network-operator names it 'validate-nicclusterpolicy-crd-established'. Cosmetic divergence across an otherwise uniform corpus.

**Blast radius:** None functional; makes cross-file grep/reporting slightly less uniform.

**Fix:** Rename network-operator's step to 'validate-crds-established' — the single-CRD case doesn't warrant a distinct convention.

njhensley (commented):
📋 **Multi-persona review — PR #1919**

▎ **Method:** 3 persona passes (Correctness/Chainsaw-domain · Verification/Evidence-integrity · Operability/CI-DX), each finding independently confirmed or refuted by a senior meta-reviewer against the resolved code. Line links pinned to head `fd8f2fd3`.
▎ **Tier legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick · ✅ Confirmed non-issue

Clean, well-scoped backfill. The assertion pattern is verified **fail-closed** (missing CRD, missing `Established` condition, or `status: "False"` all correctly fail — no vacuous-pass path), it reuses the exact idiom the merged #1245 checks established, and 5 of the 6 components' CRD names are anchored to concrete in-tree evidence (kueue manifests, kubeflow assert-crds + Go lifecycle code, network-operator `NicClusterPolicy` CR, NIM live `kubectl` dump). The slinky negative finding is correct (`crds.enabled: false`; CRDs owned by the sibling `-crds` component). All cited chart versions match the registry.

Nothing blocks merge — 0 blockers, 0 majors. The one soft spot is **gatekeeper**: its two CRD names are the only asserted names with no in-tree corroboration, and the file comments say "Verified against chart 3.22.2" while the PR body itself admits gatekeeper was *not* verified in-tree and promises an inline TODO that isn't present (F1). Worth resolving the comment/PR-body contradiction before this is trusted at face value, but the names are canonical upstream CRDs and the assert fails closed, so it's a robustness/integrity issue, not a defect.

The two systemic findings (F2 no version-coupling guard, F3 gatekeeper evidence gap) are **pre-existing** to the #1245 pattern and are tracked in **#1948** — not merge conditions for this PR.

- **Fail-closed semantics** — `(conditions[?type == 'Established']): - status: "True"` requires exactly one `Established` condition equal True; absent CRD/condition or `"False"` all fail. The JMESPath projection (vs a raw one-element list) is precisely what dodges the exact-slice-length trap, since CRDs also carry a `NamesAccepted` condition.
- **`try` without `catch`/`error`** — correct, idiomatic chainsaw for a pure assert step; matches the merged gpu-operator reference.
- **CRD names for kueue / kubeflow / network-operator / k8s-nim** — all confirmed against shipped CRs, conformance evidence, and Go lifecycle code. The later-added `trainingruntimes`/`clustertrainingruntimes` are evidence-backed (`validators/performance/trainer_lifecycle.go` waits on the exact string; `platform-kubeflow.yaml` instantiates a `ClusterTrainingRuntime`).
- **slinky negative finding** — `crds.enabled: false` confirmed; the 6 `slinky.slurm.net` CRDs are owned by the sibling `slinky-slurm-operator-crds` check with its own `Established` block.
- **Header comments match assert blocks** in all 6 files (kubeflow correctly lists all 3 after the sync commit); cited chart versions all match the registry.
- **CRD-step-first ordering** — no correctness consequence; `Established` is a legitimate prerequisite to check first.

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major   | 0 | — |
| 🟡 Minor   | 5 | F1 gatekeeper "Verified" overstatement · F3 gatekeeper no in-tree evidence · F2 no version-coupling guard *(→ #1948)* · F4 5m false-negative on wrong name · F5 volatile line-refs |
| 🔵 Nitpick | 1 | F6 step-name inconsistency |

**Recommendation: Approve with comments.** Only F1 (a comment edit) is genuinely tied to this PR; F2/F3 are tracked in #1948; F4/F5/F6 are safe to address in-PR or defer in writing. The most reliable clear for gatekeeper (F3) is a live-cluster verification pass of the 6 checks — the same gate #1245 used.

recipes/checks/network-operator/health-check.yaml
```diff
@@ -14,10 +14,13 @@
...
 # Network Operator Health Check
 #
-# Validates that the NVIDIA Network Operator is running and healthy in the
-# nvidia-network-operator namespace. Checks that the network-operator
-# deployment has at least one available replica and that no pods in the
-# namespace are stuck in Pending, Failed, or Unknown phases.
+# Validates that the NVIDIA Network Operator is running and healthy in
+# the nvidia-network-operator namespace. Also asserts
+# nicclusterpolicies.mellanox.com is Established=True (verified against
+# chart 26.1.1; re-verify on defaultVersion bump). Checks that the
```
yuanchen8911: This provenance is already stale: the registry default became 26.4.1 (#1938) after this branch was authored, so the "re-verify on defaultVersion bump" trigger has fired before merge. I rendered both versions with `--include-crds`: `nicclusterpolicies.mellanox.com` ships in both, so the assertion itself remains correct — but 26.4.1 also adds `nicnodepolicies.mellanox.com`. Please rebase, update the cite to 26.4.1, and document the assertion-set decision for the new CRD. Suggest deliberately excluding `nicnodepolicies` (no in-tree recipe instantiates a CR against it — same load-bearing-only philosophy as the rest of this PR) rather than expanding the set. The systemic version-coupling issue stays tracked in #1948.

recipes/checks/gatekeeper/health-check.yaml
```diff
@@ -27,6 +31,31 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      # Asserted against chart 3.22.2 (recipes/registry.yaml): both CRDs
+      # are documented as shipping in the chart's unconditional crds/ dir,
+      # so no values toggle should suppress them — but this has not been
+      # live-validated. TODO: confirm via helm template before trusting
+      # at face value. See #1948.
```
yuanchen8911: This TODO can be closed: rendered `gatekeeper/gatekeeper --version 3.22.2 --include-crds` — both `constrainttemplates.templates.gatekeeper.sh` and `configs.config.gatekeeper.sh` are present (among 17 CRDs total; the feature-conditional mutation/expansion/external-data ones you excluded also render under default values, which confirms limiting to the two unconditional core CRDs was the right call). Please replace the "not live-validated / TODO" hedge with the verification note; keep the #1948 reference for the version-coupling concern.

recipes/checks/kubeflow-trainer/health-check.yaml
```diff
@@ -26,6 +33,40 @@ spec:
   timeouts:
     assert: 5m
   steps:
+    - name: validate-crds-established
+      # trainjobs + trainingruntimes + clustertrainingruntimes verified
```
yuanchen8911, line 37: Verification for the one assertion here without in-tree evidence: `trainjobs` is covered by the kind-cluster chainsaw assert and `clustertrainingruntimes` is load-bearing via platform-kubeflow's manifestFiles, but `trainingruntimes` had neither. Rendered `oci://ghcr.io/kubeflow/charts/kubeflow-trainer --version 2.2.0 --include-crds`: all three asserted CRDs ship (plus `jobsets.jobset.x-k8s.io`, correctly not asserted). With that, the PR body's "left out pending confirmation" caveat for this component can be dropped — please sync the body with the diff.

yuanchen8911 (changes requested):
All asserted CRD names verified — every one exists in its chart rendered at the pinned version (details inline; two of the PR's own open TODOs are closed by that verification). The evidence-per-assertion discipline here is exactly right, and the slinky negative finding is correct (`crds.enabled: false` confirmed, CRDs owned by the sibling `-crds` component).

Two staleness fixes requested before approval:

1. The network-operator provenance comment predates the 26.4.1 registry bump (#1938) — see inline.
2. The PR body's kubeflow-trainer row says only `TrainJob` is asserted ("TrainingRuntime/ClusterTrainingRuntime … left out pending confirmation"), but the diff asserts all three. Please update the body to match the diff — see inline for the verification that lets the caveat be dropped entirely.

recipes/checks/network-operator/health-check.yaml
yuanchen8911: Addressed in e7497aa4 — verified: the cite is now 26.4.1, the branch head carries the 26.4.1 registry pin, and the nicnodepolicies exclusion is documented with the load-bearing-only rationale. Resolving.

recipes/checks/gatekeeper/health-check.yaml
yuanchen8911: Addressed in e7497aa4 — verified: the TODO hedge is replaced with the render-verification note and the #1948 reference is kept for the version-coupling concern. Resolving.

recipes/checks/kubeflow-trainer/health-check.yaml
yuanchen8911: The YAML side of this is done — e7497aa4 closes out the file-level items and I've resolved those threads. What remains before approval is the PR body sync this comment asked for: the body is now stale in four places and contradicts the diff.

- The kubeflow-trainer evidence row still cites only the TrainJob kind-cluster assert, and the caveat line still says TrainingRuntime/ClusterTrainingRuntime are "left out pending confirmation" — the diff asserts all three, and the render above confirms all three ship in chart 2.2.0.
- The gatekeeper evidence row still says "Not yet verified in-tree" with the TODO — contradicted by the verification note e7497aa4 itself added to the YAML.
- The Risk Assessment still claims gatekeeper's two CRD names are unconfirmed against the live chart.
- The Rollout notes still ask for a pre-merge helm-template verification pass — that gate has been satisfied.

Please update the body to describe the PR's current state.

One process nit: the branch caught up with main via a merge commit (f3639313); per the repo's branch-hygiene guidance please rebase onto origin/main and squash before the final push.

mohityadav8: @yuanchen8911 done

yuanchen8911: Body sync verified, and the squash preserved the reviewed content exactly (all six health-check files byte-identical to e7497aa4).

One new blocker: the squash was built on a stale base, so the diff now reverts `.settings.yaml`'s awscli pin from 1.45.57 to 1.45.56 — undoing #1975, merged this morning. Please rebase onto origin/main (the stale hunk drops out), squash, and force-push — that also clears the merge-commit nit. Will approve once pushed.

mohityadav8: @yuanchen8911 done evrthing

njhensley (approved):

yuanchen8911 (approved):
68c378a0 verified — the stale-base regression is gone (the diff is back to exactly the six health-check files, all byte-identical to the verified e7497aa4 content), the body sync holds, and the KWOK Tier 1 failure was a cancelled run, not an assertion failure: the re-run is green. Approving.

Note for the merge: the branch carries a merge-of-main commit rather than a rebase, so please land this via squash-merge to flatten the history.

PR #1932, opened 2026-07-28
fix(skills): recover cross-review lanes, harden consensus/fallback

yuanchen8911:
## Summary

Stop `aicr-cross-review` from throwing away work it has already done. Four independent causes, all of which ended a round with `incomplete` or an unusable result while the reviewers themselves were fine:

- **Reviewer lanes** treated recoverable failures as terminal. Codex is required, CodeRabbit best-effort; both returned `unavailable` for causes that were retryable or diagnosable.
- **The persisted-review fallback** could not reliably identify a stored CodeRabbit review, so it discarded valid ones and could accept a wrong one. Identity is now file modes plus blob OIDs, not the commit pair.
- **Consensus** discarded work in several ways: it voided an entire review over one malformed integration finding, accepted findings nobody could locate or corroborate, and let two lanes reporting the same defect read as two uncorroborated ones.
- **The shell contract** was stated in two prompt blocks out of six, each added only after a zsh bug had already shipped in it.

## Motivation / Context

Across seven review rounds on a large feature branch, the two optional lanes returned `unavailable` repeatedly. Both causes turned out to be recoverable, and one was not what the lane reported.

**Codex** treated every non-completed job identically. A job rejected upstream (`Selected model is at capacity`) dies in about ten seconds without using its window; a job still running when the bounded wait expires needs more time, not another attempt. Retrying the first is cheap and usually works. Retrying the second just doubles the wall clock.

**CodeRabbit** (best-effort lane) hung at `connecting_to_review_service` until its timebox killed it, writing no log at all. This reads like a cloud outage and was initially diagnosed as per-account contention. It was neither: `~/.coderabbit` is outside the sandbox write allowlist, so the CLI could not create its log or review store. It fails by hanging rather than erroring, which is why the cause was not obvious.

**The shell contract** existed in only two prompt blocks, and each acquired it *reactively* — after a zsh-semantics bug had already shipped in it. Three other prompts hand agents `git -C` commands and carried no shell note at all. Review of this PR surfaced the pattern directly: a fix here used `shopt -s nullglob`, a bash builtin absent in zsh, where an unmatched glob is a hard error that aborts the command list rather than passing through literally.

**The store fallback and the consensus stop** were not in the original scope. Both surfaced under review of this PR and while running the skill against live PRs, and both had the same shape as the lane bugs: a recoverable condition treated as terminal. The base-commit gate discarded valid reviews whenever the local ref lagged the fetched base — the common case in the fork layout this skill supports. The consensus stop killed a run on PR #1908 in which all four lanes returned `ok` and the integration lane had produced a real, evidenced defect.

Fixes: N/A
Related: N/A

## Type of Change

- [ ] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: agent skill (`.agents/skills/aicr-cross-review/`)

## Implementation Notes

**Codex lane — split the failure modes, retry only the cheap one.** The dispatch protocol now branches on `.waitTimedOut`:

- **Fast transient failure** — retried once, and only when all three hold: status is `failed` (never `cancelled`, which the companion emits only for explicit cancellation), the job died in **under 60 seconds** by its own timestamps, and the error names a known-retryable cause. `.waitTimedOut` being false is not sufficient on its own — a failure at 8:59 also satisfies it while having burned the whole window. The threshold is a number rather than a judgment call because the observed cases sit far apart: a capacity rejection at ~10s against a genuine timeout at 10m19s.
- **Wait elapsed, job still alive** (`.waitTimedOut` true, `.job.status` still `queued`/`running`) — needs more time, not another attempt. One continuation wait on the same job id; see "A wait that elapses no longer ends the Codex lane" below.
- **Exhausted budget** — a *second* `.waitTimedOut`, or no parseable JSON because the outer timeout killed the call (a dead broker). No retry, no further wait.

Retry budget is exactly one, and only for the first case. Neither case substitutes the agent's own reading for Codex's vote.

**Why the 10-minute cap is not simply raised.** The bounded wait runs inside a `Bash` call, and that tool silently kills any foreground command at 600000 ms. The cap is a platform ceiling, not a tuning choice — exceeding it would require polling across several calls or a background dispatch. This is now recorded in the script so the next reader does not try to raise the number.

**CodeRabbit lane — bypass the sandbox for that one command, and diagnose by absence.** A stall at `connecting` *with no new file in `~/.coderabbit/logs/`* is sandbox denial: a process killed mid-run still flushes a partial log, so zero bytes means it never created one. A real cloud fault leaves a log with 429/queue lines. The docs also record that contention was ruled out, since an unsandboxed run succeeding while a sandboxed one hangs is indistinguishable from lock contention until you ask why the sandboxed one failed.

**Why bypass and not an allowlist entry.** Adding `~/.coderabbit` to the sandbox write allowlist is the narrower grant and was considered first. It was rejected because the allowlist lives in each contributor's local settings while this skill is checked into the repo: anyone who has not made the same local edit gets the silent ten-minute hang instead of a usable lane. The bypass is portable and visible at the call site. Its cost — a permission prompt on step 2 of each round — is accepted, and `SKILL.md` records the trade rather than presenting bypass as the only option. Adding the allowlist entry locally remains worthwhile for frequent users; it does not conflict.

**Bounding the worktree leak the three-call split introduces.** A single invocation could carry an EXIT trap; three calls cannot, so a lane killed between steps 2 and 3 leaves its detached worktree behind. Step 1 now reaps `cr-rabbit.*` directories older than **120 minutes** and prunes the resulting stale admin entries. Both limits are what make the **find** half safe: it matches only the prefix this lane creates, and 120 minutes is twelve times the 600000 ms (10 minute) cap a single review can occupy, so it can never reach another session's live worktree. It uses `find` rather than a glob — under zsh an unmatched glob would abort the command list.

The **prune** half is repo-wide, not prefix-scoped, and the script says so rather than letting the two limits above read as covering both. It stays near-harmless because `git worktree prune` reaps only entries whose directory is already gone; the residual case is an unrelated worktree whose directory is transiently absent (unmounted volume, an `rm` in flight), whose entry would go with it. Accepted — re-adding a worktree is cheap, and the alternative is letting this lane's own stale entries accumulate. One cosmetic rough edge is also documented: a `cr-rabbit.*` directory left by a *different* clone gets its directory removed while its admin entry lives in that other repo, collected by that repo's own next prune.

This bounds the leak; it does not clear it at the moment it happens, and `SKILL.md` says so.

**Shell contract in one place.** `SHELL_CONTRACT` is a single constant interpolated in exactly one place — `NO_EXECUTION`, which every prompt builder composes exactly once — so all six assembled prompts carry it exactly once. `integrationPrompt`, `crossPrompt` and `refutePrompt` previously had no shell note at all despite being handed `git` commands. Placement matters: `NO_EXECUTION` already embeds `PINNED_READS`, so interpolating into `PINNED_READS` or `CODEX_LEAN` as well silently doubles it in every lane, which no block-level check can detect.

**Codex classification hardened.** All companion calls pin `--cwd`, since state is keyed by workspace root and each Bash call is a fresh shell — an unpinned lookup reports a live job as unknown, exiting **1** with empty stdout and `No job found` on **stderr** (measured on companion v1.0.2). That is now distinguished from a real timeout instead of both being read as budget exhaustion. The recheck after a miss is **unconditional** — always once, with `--cwd` pinned, whether or not the missed call already carried it. Two causes produce the identical message and only one is settled by adding the flag. The other is transient: in companion v1.0.2 `saveState` writes `state.json` with a plain `fs.writeFileSync` (truncate-then-write, no temp-and-rename) while `loadState` wraps `JSON.parse` in a bare `catch` returning the **default** state, whose `jobs` list is empty. A read landing inside that write window yields a well-formed `No job found` rather than an error, and the background worker rewrites that file precisely while the status call runs — so an identical repeat need not return an identical answer. Still never re-dispatch. Retry additionally requires the failure to be demonstrably fast and transient: `cancelled` is **never** retried (the companion emits it only for explicit cancellation), elapsed time is checked against a 60-second threshold rather than inferred from `.waitTimedOut` (a failure at 8:59 also satisfies it), and the error must name a known-retryable cause.

**Persisted-store fallback.** If a run still fails, the lane scans `~/.coderabbit/reviews/*/*/reviews/*/git.json`. The scan is done in `python3`, not a shell glob: the tool shell is zsh, where an unmatched glob is a hard error that aborts the command list and would strand the log fallback in exactly the no-match case it exists for. Malformed or half-written records are skipped rather than aborting the scan.

Acceptance is `head` plus **the pinned change itself**, never `baseCommitId`. Measured against the real local store: the CLI writes `baseCommitId` from the base it resolves locally in that working directory, **not** from the `--base-commit` the lane passes. The same head `e95a4b6dc` carries two different values — `3d4f77ddb` (a main tip) and `6942d69ff` (a commit that is not on main at all) — while three other valid records carry the stale local `main` (`97e3f8b4b`) rather than the base Phase 1 fetched from the canonical repo (`3d4f77ddb`). Gating on equality discarded good reviews whenever the local ref lagged, which in the fork layout this skill supports is the common case rather than a corner. `baseCommitId` is now reported in `statusNote` and never branched on, and the superseded rule that discarded a record on a `baseCommitId` difference is **deleted** rather than left standing below its replacement — where it was the more imperative of the two and the last word on the subject, so an agent resolving the conflict would most likely have reinstated the removed gate.

What identifies the review instead is the **blob OIDs the record already stores**. Acceptance is layered, cheapest first, and only the last step is identity:

1. Same paths, from `git diff --numstat -z` over the **three-dot** range Phase 1 saved (two-dot compares the base tip to the head, so it includes everything that landed on the base after branching — on a real pair, 13 files where the valid record had 59).
2. Same per-file line counts, against the record's `linesAdded` / `linesRemoved`.
3. The committed-only `lanes` shape.
4. The **file modes and blob OIDs** the record's own patch states, equal to what `git diff --raw --no-abbrev` reports for the pinned range.

Steps 1–2 are a **prefilter, not identity**: two different patches collide on a numstat trivially — a `beta`→`BETA` edit and a `gamma`→`GAMMA` edit in the same file both report one added and one removed — and a review of this head against another base can leave exactly such a record.

**Why OIDs and not the stored patch text.** Comparing the text was the obvious move and is wrong: bare `git diff` output moves with the reader's gitconfig. Measured on a scratch repo, the patch text differs under `diff.noprefix` and `diff.context` and fails outright under `diff.external`, so a contributor with ordinary settings gets a permanent mismatch and this fallback silently never fires — the same works-on-my-machine class the sandbox decision rejected an allowlist for. Pinning flags (`--no-ext-diff`, `--no-color`, `-U3`, `--src-prefix`) fixes *our* side only; if the CLI produced the record under that same config, hardening makes the mismatch worse rather than better. Blob OIDs are content hashes: stable on both sides, and exact content identity rather than a rendering of it. Both git calls still pass `--no-ext-diff --no-color`, which the `--raw` call needs and `--numstat` is measured not to.

**An entry with no `index` line is not rejected on sight.** Git omits that line exactly when the blob is unchanged, which for a record reaching this check means a mode change (`chmod +x`) — verified on fixtures, while a rename *with* modification does carry one. Since the loop rejects the whole record on its first offending entry, and this repo ships executable scripts (`tools/cleanup`, `tools/e2e`), a single `chmod` would have permanently disabled the fallback for that PR. Such an entry is now accepted when the pinned diff agrees the blob is unchanged (`src == dst`) and rejected as `no-index-line` otherwise, so the gate stays exact. The documented cause was also wrong — it named an external diff driver, which is not the common case.

**Modes are part of identity; blob OIDs alone are not.** With a `chmod +x` in one commit and an edit in the next, a stored review of the *edit alone* carries the same path, line counts, `lanes` and **both** blob OIDs as the pinned range covering both changes. Reproduced on a fixture — the gate accepted the wrong record:

| | modes | OIDs | numstat |
|---|---|---|---|
| Pinned (chmod+edit) | `100644 → 100755` | `c07ce13..0e682b8` | `1 1` |
| Stored (edit only) | `100755 → 100755` | `c07ce13..0e682b8` | `1 1` |

Modes are now compared before OIDs. Git states them in one of four shapes (`old mode`/`new mode`, `new file mode`, `deleted file mode`, or the trailing mode on the `index` line), and the trailing mode is absent exactly when the `old`/`new` pair is present, so the cases do not overlap. A record stating no mode at all cannot prove identity and is skipped. A mode-only change still passes through the `no-index-line` branch (verified).

**Reaper failures can no longer abort setup.** `rm -rf "$STALE"` was unguarded under `set -euo pipefail`, so an unremovable stale directory — reachable where `TMPDIR` is unset and `/tmp` is shared between users — aborted STEP 1 before the prune, the `mktemp` and the `worktree add`. Verified on a fixture: the old form dies, the wrapped form reaches the worktree-add step. Only the reaper is neutralised; a failing `worktree add` stays fatal, since STEP 2 must not review the wrong tree.

**Cleanup guidance now names `${TMPDIR:-/tmp}/cr-rabbit.*`.** Setup creates the worktree under that fallback, but cleanup said `$TMPDIR/cr-rabbit.*`, which with `TMPDIR` unset points at `/cr-rabbit.*` while the leak sits in `/tmp`.

**A 100%-similarity rename also omits the line but never reaches that branch**, and the comment claiming otherwise is corrected. The reason: the CLI scopes its stored patch and counts to the post-rename path, so such a record fails the scope prefilter first. Renames remain the documented false-negative; only the explanation changed.

**A wait that elapses no longer ends the Codex lane.** `.waitTimedOut` true says only that the *inner wait* expired; the job is dispatched in the background and outlives the Bash call waiting on it. When `.job.status` is still `queued`/`running`, the job needs more **time**, not another attempt — so the protocol now grants **one continuation wait** on the *same* job id in a fresh call. Nothing is re-dispatched, no work is duplicated, and the first job's result is still the one collected. A second `.waitTimedOut` is then genuinely exhausted budget, and the lane reports `unavailable` with the job id so the result can be fetched later. A live job therefore gets two waits (~18 min), which is the "polling across several calls" the 600000 ms per-call ceiling always left open.

Found by running this skill against PR #1917 (run `wf_62cb6130-2f4`): the lane hit `.waitTimedOut` at the full 540000 ms while the job was demonstrably mid-work — its log showed pinned `git grep` commands completing with exit 0 during the consumer-search phase — and the job was **still** reported `running` long after the review had been abandoned, with its result still retrievable. Calling that exhausted budget discarded a required lane, and the entire review with it, over a job that had merely not finished.

**A malformed integration finding no longer voids the whole review.** An integration finding claims a specific consumer breaks, so one lacking `consumerPath`/`consumerLine` cannot be verified and must stay out of consensus — but it is now dropped on its own, with a `log()` naming what went, instead of failing the run.

Found by running this skill against PR #1908 (run `wf_c06599d7-3d4`): the integration lane returned several findings — one a real, evidenced stale UAT fixture — plus a stale-doc-comment finding at `pkg/recipe/metadata_store_test.go:2675` that legitimately has no consumer. The review reported `incomplete` with all four lanes `ok`, ~400k subagent tokens spent, and no report produced at all. The run still stops when **every** integration finding is unusable, which is the case the check was added for: silently dropping the lane's only finding once yielded `consensusReached: true` while a required lane had contributed nothing.

**The args parse is wrapped.** Its bare failure named neither the skill nor the offending input — `JSON Parse error: Unterminated string`, no agent spawned, run dead before Phase 1 — where the missing-arg loop beside it has always reported the field. Hit while passing `args` as a JSON-encoded string, which the Workflow tool does not want; the message now says so.

**Coordinates are validated by one shared rule, on both sides of a finding.** The response schema requires `path` and `line`, leaves `consumerPath`/`consumerLine` optional (deliberately — only integration findings carry a consumer), and constrains none of the four, so `""`, `"   "`, `0` and `-1` all satisfy it and all passed a truthiness/null check — putting an unlocatable finding into consensus and, when it was the lane's only finding, producing a clean-looking result.

`hasCoords` is now applied to a finding's own `path`/`line` inside `intake()` — **every lane, not just integration** — and to the consumer pair for integration findings. It is one helper rather than two call-site conditions on purpose: each earlier version of this guard fixed the pair it was shown and left the other, and a shared rule is what stops the next field pair from repeating that. Tightening the schema instead would fail a whole lane on one bad field, which is the all-or-nothing behavior this PR removes.

**The zero-survivor rule covers every required round-1 batch, not just integration.** Claude's and Codex's round-1 findings run through one `intakeBatch` helper that reports what survived. A required lane whose round-1 findings are *all* malformed contributed nothing, and letting them vanish silently was the same false-clean one lane over — `status: ok`, `consensusReached: true`, zero candidates. Mixed batches still proceed. CodeRabbit is exempt: a total loss there records `NONE` and raises the bar, exactly as a lane that never ran.

**It deliberately stops at round 1 — cross-review `newFindings` are exempt.** Late findings pass the same `intakeBatch` gate, so a malformed one still never enters the tally, but a total loss there is not fatal. The rule tests "this lane contributed nothing", which round 1 can assert because findings are the lane's whole output. In the cross-review round the output is the *evaluations*, and the completeness gate above has already returned `incomplete` unless the lane evaluated every presented candidate — so by that point it has provably contributed, and `newFindings` are supplementary and volunteered. `FINDING_ITEM` constrains neither `line` nor `evidence`, so a schema-valid late finding with `line: 0` or empty `evidence` yielded `returned=1, accepted=0` and aborted a finished run before the tally and verification ran — the same disproportionate total loss seen on PR #1908, one round over, and against this PR's own goal of not discarding work already done. Malformed late findings are dropped individually, each logged by `intake()` with its reason, as the integration lane already does.

Rejection reasons are also counted separately (`unlocatable` vs `unevidenced`) rather than inferred from a subtraction, which had been reporting an own-coordinate failure as "dropped for missing evidence" and pointing the reader at the wrong defect.

**Candidate dedup keeps the normalized summary in its key — location-only keying was tried and reverted.** It did merge the real duplicates (two lanes wording one defect differently; measured on a run against PR #1936, `nvidia_smi.go:130` and `validator.md:412` each appeared twice), but the evaluation schema permits exactly **one verdict per candidate id**, so a merged pair of *distinct* same-line defects has no correct verdict: confirming the real one also confirms the false one, and refuting the false one dismisses the real one. Retaining both summaries prevented data loss but not mis-adjudication — and mis-adjudication is a correctness defect where duplicate splitting is only a quality one.

Text similarity was evaluated as an alternative merge signal and rejected on evidence: a pair of separately-anchored summaries scored **higher** (0.359 Jaccard) than one of the true duplicate pairs (0.281), so no threshold separates the populations.

**Duplicates are surfaced without merging.** Candidates sharing a location — `path:line` **and** the same `consumerPath`/`consumerLine` — are flagged as possible duplicates of each other. The consumer half matters: one changed declaration breaking two callers is deliberately two candidates, and hinting they might be duplicates would push reviewers to collapse a distinction the key exists to preserve. The flag and the flag reaches both the cross-review candidate list and the refuter prompt. Reviewers decide whether the two are one defect and evaluate them consistently, so equivalence stays an explicit judgement rather than an assumption from a shared line number, and each real defect keeps its own verdict.

**Merging stops once candidates are presented.** A finding raised *during* cross-review that merged into an already-presented id would be tallied with evaluations cast before it existed: two prior AGREEs could confirm a defect nobody read, and one adversarial refuter cannot split a composite afterwards. Late findings now always become their own candidate and, being unpresented, stay contested for the human — the behavior the cross-review round already specifies for them.

**The persisted-review matcher keeps the raw status.** Dropping it collapsed a pinned rename-plus-edit onto an edit-only record: on a fixture, a pinned `R077 old.txt -> new.txt` and a stored `M new.txt` share modes, blobs and line counts once status and source path are discarded, so a review that never saw the rename could be accepted. A rename or copy in the pinned range is now rejected as `rename-unverifiable`, making the documented rename false-negative true rather than merely claimed.

**Contested findings now say why they are unsettled.** The bucket held two states that need different things from a reader: presented and evaluated but short of 2-of-3, versus raised *during* the cross-review round and so never cross-evaluated at all. Measured on a run against PR #1929, **8 of 8** contested findings were the latter — each carrying a single `AGREE` from its reporter and `NONE` elsewhere, so the count read as eight disagreements when there were none. Each entry now carries `adjudication` (`evaluated` | `raised-late`) and a matching `why`.

The change is deliberately **annotation-only**: array membership, the `resolution` values and the `consensusReached` formula are untouched, so nothing already consuming the result can regress. `consensusReached` keeps counting late findings, since a late finding is unadjudicated and letting it report consensus would be the overstatement this skill exists to avoid.

**Two smaller repairs in the same loop.** `filePath` is now established once, up front, so the identity loop cannot put a `None` into an argv list and raise a `TypeError` with no `ERROR` line — previously the guard existed on the counts comprehension but not the identity loop. The sentinel is tested with `is not None` rather than truthiness, so an empty `filePath` cannot read as a pass. And a git failure can no longer be mislabelled a content mismatch, because both git calls fail closed once at the top.

`-z` matters independently of all this: `git diff --name-only` plus `.stdout.split()` shreds a valid path containing a space (`with space.txt` → two set members) and rejected every record for such a PR. The field split is capped at `maxsplit=2`, since `-z` emits paths raw and unquoted and a filename containing a TAB would otherwise raise `ValueError` and exit 1 with a traceback and no `ERROR` line. The reader is additionally told that **any** non-zero exit means no verdict was reached, so an unanticipated record shape cannot be read as a clean negative.

Every entry must additionally prove it was **committed-only**. `lanes` is a boolean map (`{"committed": true, "uncommitted": false}`), so the check tests the values and requires that exact shape. Merely rejecting a truthy `uncommitted` was not enough — a record from an older CLI that omits `lanes`, one carrying a non-dict, and one with `committed: false` all passed as clean and would have contributed comments on code outside the pinned head. Absent proof of clean, skip. The review itself now runs with `--committed`. Findings are read only from UUID-named siblings carrying `fileName` and `comment`, since a real record also contains `git.json`, `incrementalDiff.json` and `internalState.json`.

The pinned diff that the file-set check compares against is computed **fail-closed**. An unreachable base — shallow clone, pruned ref, wrong repo path — exits non-zero with empty stdout; taking that as an empty file set would reject every stored record and then report "no stored review matches", which the prompt explicitly labels the normal case. A broken-git condition would have been laundered into a clean negative. The scan now prints `ERROR git-diff-failed` and exits 2 instead, and the prompt distinguishes that from an empty result.

Several records can also pass every check simultaneously — a re-run after a transient failure produces a twin with the same pair and the same file set, which is precisely the situation this fallback exists for. Candidates are ordered by mtime and exactly one `MATCH` is emitted, newest first, since a re-run supersedes what it replaced; the rest print `SKIP superseded-duplicate`. One known false negative is documented rather than worked around: a rename never matches. The scan uses `git diff --numstat -z` and parses the rename tokens, but the CLI scopes its stored patch and counts to the post-rename path while the pinned range reports the rename as a whole, so the record's numstat differs and the scope prefilter rejects it. Not observed in practice, and skipping is the safe direction.

## Testing

```bash
# Infra-only change: no Go, no YAML, no docs under docs/.
# Scoped checks matching what the change touches:

# 1. workflow.mjs parses (as an async function body — top-level await/return
#    are valid in the Workflow execution context, so a naive parse fails)
node -e "...new AsyncFunction(...)..."   # SYNTAX OK

# 2. SKILL.md frontmatter is well-formed and version bumped
#    keys: name, description, user-invocable, disallowed-tools, argument-hint, version
#    version: 0.3.0 -> 0.3.18

# 3. Ran the RECOVERY block verbatim in the real tool shell (zsh 5.9):
#      - no-match  -> rc=0, empty output, following statement executes
#      - truncated record present -> skipped, scan continues, valid match returned
#    (The prior bash-oriented version failed here: `shopt` -> command not found,
#     and an unmatched glob aborted the command list before the log fallback.)

# 4. Completeness check on the shell contract: every assembled prompt that hands
#    out a shell command composes SHELL_CONTRACT (6/6), single definition, no
#    hand-written copies remaining.

# 5. RECOVERY block extracted from the ASSEMBLED prompt (not a hand copy) and run
#    against the real ~/.coderabbit store, 120 records:
#      - stale-base record (baseCommitId = local main, != pinned BASE_SHA)
#          -> MATCH   (old rule REJECTED this valid review)
#      - head with two records, differing baseCommitId
#          -> one MATCH (newest) + one SKIP superseded-duplicate
#      - valid range, no stored record   -> rc=0, empty output
#      - unreachable base                -> ERROR git-diff-failed rc=128, exit 2
#    Also confirmed no raw NUL/TAB leaks through the JS template literal (chr() is
#    used precisely because a backslash escape would be consumed at assembly time).

# 5b. numstat parity: git diff --numstat -z reproduces a real record's
#     linesAdded/linesRemoved exactly. Whitespace path + rename handling checked in a
#     scratch repo: old split() -> {'with','space.txt','new','name.txt'};
#     new parser -> {'with space.txt': (1,0), 'new name.txt': (1,0)}.

# 5c. lanes predicate, old vs new, over six entry shapes:
#       clean committed-only  accept/accept | uncommitted:true   skip/skip
#       lanes absent          accept/SKIP   | lanes is a list    accept/SKIP
#       committed:false       accept/SKIP   | mixed clean+dirty  skip/skip

# 5d. numstat parser over a scratch repo containing a rename, a space path, a
#     TAB path and a binary file. Raw: -^-^b.bin|0^0^|old.txt|new.txt|1^0^tab^name.txt|
#     1^0^with space.txt|   (| = NUL, ^ = TAB)
#       bare split(TAB)  -> ValueError: too many values to unpack (expected 3, got 4)
#       split(TAB, 2)    -> {'b.bin': ('-','-'), 'new.txt': ('0','0'),
#                            'tab\tname.txt': ('1','0'), 'with space.txt': ('1','0')}

# 5e. Repo-wide grep confirms no surviving baseCommitId gate in either file.

# 5f. Blob-OID identity gate, against the real store:
#       genuine record                              -> ACCEPT
#       tampered dst blob                           -> SKIP blob-mismatch
#       record with no index line                   -> SKIP no-index-line
#       (the numstat prefilter alone accepts a same-count content change)

# 5j. Config independence — the assembled block, run against the real store under
#     each hostile gitconfig via GIT_CONFIG_COUNT. All six return MATCH:
#       diff.noprefix=true, diff.context=7, diff.algorithm=histogram,
#       diff.external=/bin/false, color.diff=always, core.abbrev=12
#     Re-run specifically for the STALE-BASE record (baseCommitId 97e3f8b4b vs
#     pinned 3d4f77ddb): MATCH under the default config and under all six.

# 5l. CRLF: raw git patch bytes for a CRLF file carry 4 CRLF sequences;
#     subprocess text=True yields 0, so a patch-TEXT comparison would reject a
#     valid record. Blob-OID identity holds on the same fixture (b5eff57/d5a6cc6
#     parsed from the stored index line == git --raw), so the CRLF hazard does
#     not reach the current design. text=True now only reads --numstat/--raw
#     output, which is ASCII.

# 5q. Live runs of the skill against two real PRs (see Motivation):
#       PR 1908 (config-change): claude/codex/coderabbit/integration all ok.
#       PR 1917 (code-change):   claude/coderabbit/integration ok; codex hit
#         .waitTimedOut at 540000 ms mid-consumer-search. Job verified still
#         "running" afterwards via `codex-companion status --cwd`, result still
#         retrievable — the evidence behind the continuation wait above.
#       NOT covered: the integration-drop path did not fire on PR 1917 (the
#       lane's single finding carried a consumer), so it remains verified only
#       on the six synthetic branches in 5p.

# 5r. Coordinate validation via the shared hasCoords helper. Unit level:
#       valid -> ok;  "" / "   " / null path -> REJECT
#       line 0 / -1 / 1.5 / null            -> REJECT
#     Integration lane end-to-end, BOTH coordinate pairs:
#       all valid            accepted=1 -> proceed
#       own path ""          accepted=0 -> INCOMPLETE
#       own line 0           accepted=0 -> INCOMPLETE
#       consumer path "   "  accepted=0 -> INCOMPLETE
#       consumer line -1     accepted=0 -> INCOMPLETE
#     hasCoords has exactly two call sites (intake + integration filter);
#     intakeBatch has three (round 1, cross-review, integration); zeroSurvivors
#     has two (round 1, integration) — the cross-review site is deliberately gone.

# 5u. Dedup / duplicate flagging, driven directly through addCandidate:
#       distinct same-line defects   -> SEPARATE ids (F1/F2), each flagged
#                                       "possible duplicate of <other>" on both sides
#       two consumers, same changed line -> SEPARATE ids, NO hint (the key's
#                                       consumer distinction is preserved)
#       same consumer, different wording -> SEPARATE ids + hint
#       no consumers (non-integration)   -> hint
#       identical wording (normalized) -> merged, sources=[claude,coderabbit], sev=major
#       late exact match after presenting -> new id (never inherits votes)

# 5v. Presented-merge guard:
#       round-1 same-location dup   -> merged, sources=[claude,codex], sev=major
#       late same-location finding  -> NEW id, unpresented (stays contested)
#       second late finding         -> its own id; 3 candidates total

# 5w. Rename fail-closed, on a rename+edit fixture:
#       pinned :100644 100644 01f84f8..45557c6 R077 old.txt new.txt, numstat 1/1
#       old matcher (status dropped) -> an edit-only record with the same blobs MATCHES
#       new matcher                  -> SKIP rename-unverifiable

# 5x. Contested annotation — regression guards, all PASS:
#       resolution values unchanged (no new value written)
#       consensusReached formula unchanged
#       contested membership still driven by contestedIds
#       no new top-level result array
#       lateIds declared at module level and populated at the one late site
#       contested entries annotated
#     Diff is +21/-4; the three removed lines are each replaced by a superset.

# 5t. Required-lane zero-survivor guard, per batch:
#       claude: 1 good                 accepted=1 -> proceed
#       claude: only blank-coords      accepted=0 -> INCOMPLETE (unlocatable=1)
#       claude: only blank-evidence    accepted=0 -> INCOMPLETE (unevidenced=1)
#       claude: good + malformed       accepted=1 -> proceed
#       codex:  only line 0            accepted=0 -> INCOMPLETE
#       coderabbit: only malformed     accepted=0 -> proceed (best-effort)
#       claude: no findings            accepted=0 -> proceed
#     Cross-review newFindings, same batches, NOT subject to the rule:
#       codex: only line 0             accepted=0 -> proceed (dropped + logged)
#       codex: only blank-evidence     accepted=0 -> proceed (dropped + logged)
#       codex: good + malformed        accepted=1 -> proceed, good one admitted

# 5s. Reaper guard now covers `git worktree prune` too, verified in zsh under
#     set -euo pipefail: with the prune outside the guard the block aborts before
#     worktree-add; inside it, execution reaches worktree-add. Motivated by an
#     observed `Operation not permitted` on .git/worktrees for a worktree created
#     after the sandbox profile was computed.

# 5p. Integration-finding acceptance, all six branches driven directly. "Accepted"
#     is measured on what survives intake(), which independently drops blank
#     evidence — gating on the coordinate check alone reported a false clean:
#       good only                        accepted=1 -> proceed
#       ONLY blank-evidence              accepted=0 -> INCOMPLETE
#       good + blank                     accepted=1 -> proceed
#       only no-consumer                 accepted=0 -> INCOMPLETE
#       blank + no-consumer              accepted=0 -> INCOMPLETE
#       no findings                      accepted=0 -> proceed
#     The PR 1908 shape is "good + blank"-like, which previously voided the run.

# 5m. Mode identity, on a chmod-then-edit fixture (C0 base, C1 chmod +x, C2 edit):
#       pinned C0...C2 modes 100644->100755, stored C1...C2 modes 100755->100755,
#       identical OIDs c07ce13..0e682b8 and identical numstat 1/1.
#       OLD gate -> MATCH (wrong record accepted).  NEW gate -> SKIP mode-mismatch.
#       mode-only change (C0...C1) -> still ACCEPT via the no-index-line branch.
#     Real ~/.coderabbit store re-checked: stale-base rescue still MATCH.

# 5n. Reaper guard, zsh 5.9 under set -euo pipefail, with an rm that fails:
#       unguarded -> aborts, never reaches worktree add
#       wrapped   -> "REACHED worktree-add step"

# 5o. Args error, driving the assembled script as an async function body:
#       malformed JSON string -> "aicr-cross-review: could not parse the \"args\"
#         input as JSON (Unterminated string ...). Pass args as a real object ..."
#       valid object, missing field -> "missing required arg: repo" (unchanged)

# 5k. no-index-line branch, on fixtures (chmod +x, pure rename, rename+modify,
#     plain edit):
#       mode.txt   no index line, src==dst -> ACCEPT
#       plain.txt / renmod*  index line present -> ACCEPT
#       negative control: index line stripped, blobs differ -> no-index-line
#     ren2.txt (R100) also has no index line in git's own output, but a stored
#     record covering a rename never reaches this branch: the CLI scopes its
#     patch and counts to the post-rename path, so the scope prefilter rejects
#     it first. Renames stay a documented false-negative, not an ACCEPT.
#     Patch-text comparison was measured to differ under three of these
#     (md5 7766a9e2 baseline vs 62a0f028 noprefix, 301c70c5 context=7,
#      and "external diff died" -> rc!=0) — hence OIDs.

# 5g. Collision reproduction: a beta->BETA edit and a gamma->GAMMA edit in one
#     file both yield numstat "1\t1\tf.txt"; patch text differs.

# 5h. Companion state read: lib/state.mjs saveState uses plain fs.writeFileSync;
#     loadState catches JSON.parse and returns defaultState() with jobs: [].
#     Confirms a pinned lookup can miss transiently.

# 5i. The chr(10) fix was found BY this harness: '\n' written in the .mjs source
#     was consumed at template assembly and landed as a literal newline inside a
#     Python string literal -> SyntaxError in all three assembled variants.

# 6. STEP 1 reaper, run verbatim in zsh 5.9 under `set -euo pipefail`:
#      - one 120-min-old and one fresh cr-rabbit.* dir -> old removed, fresh kept
#      - no matching dir at all -> rc=0, following statements execute
```

Full `make qualify` was skipped deliberately. The change touches only `.agents/skills/aicr-cross-review/**` — one `.mjs` script and one `.md`. No workflow under `.github/workflows/` references `.agents/skills` or `SKILL.md`, there is no `Makefile` target for skills, the repo has no JS linter config, and `yamllint` does not cover Markdown. Tests, e2e, and Go lint cannot regress from these files.

Both fixes were exercised end-to-end during a seven-round review: a capacity rejection at ten seconds, a genuine timeout at 10m19s, two sandboxed CodeRabbit runs that hung with zero log output, and a bypassed retry that completed with 23 findings against the pinned pair.

## Risk Assessment

- [x] **Low** — Isolated change, easy to revert

**Rollout notes:** Prompt text, an embedded Python block, and documentation — no runtime code path in `aicr` is affected. The skill is opt-in and invoked explicitly. Reverting is a single-commit revert with no migration.

## Checklist

- [ ] Linter passes (`make lint`) — N/A, no linter covers these paths (see Testing)
- [ ] I added/updated tests for new functionality — N/A, no test harness exists for skill prompts
- [x] I updated docs if user-facing behavior changed (`SKILL.md` operational notes)
- [x] Verified the new shell block by executing it in the actual tool shell, not by reasoning about its semantics

### DIFF
--- a/.agents/skills/aicr-cross-review/scripts/workflow.mjs
+++ b/.agents/skills/aicr-cross-review/scripts/workflow.mjs
@@ -29,7 +29,20 @@ export const meta = {
 // ---------- args ----------
 // Tolerate args arriving as a JSON string (observed 2026-07-21: the harness
 // delivered `args` stringified even when the tool call passed a real object).
-const parsedArgs = typeof args === 'string' ? JSON.parse(args) : (args || {})
+// The parse is wrapped because its bare failure is undiagnosable, unlike the
+// missing-arg loop below: it throws "JSON Parse error: Unterminated string" naming
+// neither this skill nor `args`, no agent ever spawns, and the run dies before Phase 1.
+// Observed while passing args as a JSON-encoded string, which the Workflow tool
+// explicitly does not want — nothing in the message pointed at that.
+let parsedArgs
+try {
+  parsedArgs = typeof args === 'string' ? JSON.parse(args) : (args || {})
+} catch (err) {
+  throw new Error(
+    `aicr-cross-review: could not parse the "args" input as JSON (${err.message}). ` +
+    'Pass args as a real object in the Workflow tool call, not a JSON-encoded string — ' +
+    'a stringified object with embedded quotes is the usual cause.')
+}
 const { pr, repo, repoPath, headSha, baseSha, diffPath, prType, changeList, repoNotes } = parsedArgs
 for (const [k, v] of Object.entries({ pr, repo, repoPath, headSha, baseSha, diffPath, prType })) {
   if (!v) throw new Error(`missing required arg: ${k}`)
@@ -122,6 +135,13 @@ Repo working copy: ${repoPath}.`
 // they go inside SINGLE quotes. Double quotes stop globbing but still expand $(...)
 // and backticks, so a filename like docs/$(id).md would execute — the exact thing this
 // skill promises not to run. Single quotes are the only literal form in sh/bash/zsh.
+// Interpolated in exactly ONE place: NO_EXECUTION, which every prompt builder composes
+// exactly once, so every assembled prompt carries this once. Do not also add it to
+// PINNED_READS or CODEX_LEAN — NO_EXECUTION already embeds PINNED_READS, so a second
+// interpolation silently doubles it in every lane (that is how the first attempt got
+// it wrong, and a block-level check cannot see it).
+const SHELL_CONTRACT = `Shell contract: the tool shell is zsh, not bash. Never name a shell variable "status" (readonly in zsh; the assignment silently fails). Do not reach for bash builtins such as "shopt" — they are simply not found. An unmatched glob is a hard error that aborts the whole command list, not a literal passthrough as in bash, so prefer doing the work in python3 whenever a pattern might match nothing.`
+
 const PINNED_READS = `  read:   git -C "${repoPath}" show '${headSha}:<path>'
   search: git -C "${repoPath}" grep -n -e '<pattern>' ${headSha} -- '<glob>'
   SINGLE quotes are required and must not be changed to double quotes: they are what
@@ -144,7 +164,8 @@ Execution rules:
 - ALLOWED: the trusted commands this prompt explicitly prescribes anywhere (including any
   detached worktree or CLI invocation it spells out), read-only "gh" queries, and the
   saved diff plus pinned reads/searches:
-${PINNED_READS}`
+${PINNED_READS}
+${SHELL_CONTRACT}`
 
 const OUTPUT_RULES = `
 Reporting rules:
@@ -167,14 +188,34 @@ const CODEX_DISPATCH = `
 Codex dispatch protocol (mandatory):
 1. comp=$(ls -t ~/.claude/plugins/cache/openai-codex/codex/*/scripts/codex-companion.mjs | head -1)
 2. Dispatch the review as a BACKGROUND job (pass --background to codex-companion task) so it returns a job id immediately. NEVER run foreground: a foreground call hangs forever if the broker dies mid-turn.
+   Pass --cwd "${repoPath}" on this call and on EVERY later status/result call. Companion state is keyed by the workspace root derived from the working directory, and each Bash call is a fresh shell that may start somewhere else — an unpinned lookup silently resolves to a different workspace and reports the job as unknown.
 3. Wait for it with the companion's own bounded wait — do NOT hand-roll a poll loop:
-     node "$comp" status <job-id> --wait --timeout-ms 600000 --poll-interval-ms 30000 --json
-   Read the returned JSON yourself. The job state is at .job.status (NOT a top-level "status" field), and .waitTimedOut is true if the 10-minute cap expired while the job was still queued/running.
-4. .job.status == "completed" → fetch the payload with: node "$comp" result <job-id> --json  (NOT status, which returns only a job summary).
-   Anything else (failed, cancelled, or .waitTimedOut) → return status:"unavailable" with the observed .job.status in statusNote. Do NOT retry and do NOT substitute your own review — an unavailable Codex lane is the expected outcome here.
-Every Codex task you dispatch must carry the execution rules verbatim in its own prompt — Codex runs in its own shell and does not inherit this one's constraints.
-Shell note: never name a shell variable "status" (readonly in zsh; the assignment silently fails).
-Sandbox: the review itself runs sandboxed. The one known exception is the companion's own job log under ~/.claude/plugins/data, which is sandbox-denied — if dispatch fails on that write, bypass for that call only. Never bypass for anything else, and never for a call that touches the working copy or GitHub.`
+     node "$comp" status <job-id> --cwd "${repoPath}" --wait --timeout-ms 540000 --poll-interval-ms 30000 --json
+   Run that Bash call with a timeout of 600000 ms. The inner wait is deliberately 60s SHORTER than the outer cap: the two must not be equal, or Bash can kill the command before it prints its JSON, leaving you with no .waitTimedOut to classify on — and an unclassifiable kill is the one case step 5 must never guess at, since retrying a genuine timeout doubles the wall clock for nothing.
+   Read the returned JSON yourself. The job state is at .job.status (NOT a top-level "status" field), and .waitTimedOut is true if the inner deadline expired while the job was still queued/running.
+   Absent JSON has TWO causes and they are not interchangeable — distinguish them before classifying:
+     - Exit status 1 with EMPTY stdout and "No job found" on stderr (measured on companion v1.0.2; the message goes to stderr even with --json, so capture both streams before deciding): this is a LOOKUP MISS. It is NOT evidence the job died — it may still be running. Never classify a lookup miss as exhausted budget; doing so abandons a live required lane. Note the exit status is a reliable discriminator here, so read it directly rather than inferring from output — and do not read it through a pipe, where $? is the last stage's status, not the companion's.
+       ALWAYS recheck EXACTLY ONCE with --cwd "${repoPath}" pinned, whether or not the call that missed already carried it. Two independent causes produce the identical message, and only one of them is settled by adding the flag:
+         - An unpinned lookup resolves to a different workspace and reports a live job as unknown. Pinning fixes it.
+         - A pinned lookup can miss TRANSIENTLY. In companion v1.0.2, saveState writes state.json with a plain fs.writeFileSync (truncate-then-write, no temp-file-and-rename), and loadState wraps JSON.parse in a bare catch that returns the DEFAULT state — whose jobs list is empty. A read landing inside that write window therefore yields a well-formed "No job found" rather than an error, and the background worker is updating that file precisely while the status call runs. So an identical repeat is NOT guaranteed to return an identical answer; a brief pause before the recheck makes it likelier to clear.
+       Once a pinned recheck has also missed, return status:"unavailable" saying the job could not be located, and stop. Do NOT re-dispatch the task: the original may still be running, and a second dispatch doubles the work while the first result becomes unreachable. Do NOT call it exhausted budget either; the distinction belongs in statusNote.
+     - No output at all because your Bash call hit its 600000 ms timeout: that IS exhausted budget. Do not retry; say so in statusNote.
+4. .job.status == "completed" → fetch the payload with: node "$comp" result <job-id> --cwd "${repoPath}" --json  (NOT status, which returns only a job summary).
+5. Otherwise classify, and retry ONLY a failure that is demonstrably both fast and transient. All three conditions must hold:
+   a. .job.status is "failed" — NOT "cancelled". The companion emits "cancelled" only for explicit cancellation, so retrying it restarts work something or someone deliberately stopped.
+   b. .waitTimedOut is false AND the job died in UNDER 60 SECONDS — compute elapsed from the job's own timestamps, not from the absence of a timeout. Use the number, not a judgment call: the two observed cases sit far apart (an upstream capacity rejection landed at ~10s, a genuine timeout at 10m19s), and leaving "quickly" undefined is how a one-retry budget turns into retry-whenever-it-feels-right. A job that fails at 8:59 also satisfies .waitTimedOut == false while having consumed nearly the whole window; that is not a transient blip and retrying it costs another full window.
+   c. The error text names a known-retryable cause: an upstream capacity rejection ("Selected model is at capacity"), a broker startup error, or a transient dispatch fault. An unrecognised error is not assumed retryable.
+   All three true → dispatch ONE new background job per step 2 and wait once more per step 3. If the second attempt also fails, return status:"unavailable" with BOTH observed .job.status values and the broker error text in statusNote.
+   Anything else — "cancelled", a late failure, or an unrecognised error — is a deliberate stop or an unexplained failure. Do NOT retry. Return status:"unavailable" with .job.status, .waitTimedOut, elapsed time, and whether the job log showed active progress.
+   Retry budget is exactly one, and only when a, b and c all hold. Never loop.
+6. .waitTimedOut true is DIFFERENT from all of the above, and is NOT the end of the lane. It means only that the inner wait elapsed — it says nothing about the job, which is dispatched in the background and outlives the Bash call that was waiting on it. Re-read .job.status:
+   - Still "queued" or "running" → the job is ALIVE and needs more TIME, not another attempt. Wait once more on the SAME job id, in a NEW Bash call, exactly as in step 3 (same 540000 ms inner wait, same 600000 ms outer timeout). This is a CONTINUATION, not a retry: no second job is dispatched, no work is duplicated, and the first job's result stays the one you collect. Then classify the second wait's outcome by these same rules — except that this continuation is granted ONCE, so a second .waitTimedOut IS exhausted budget: return status:"unavailable" saying the job was still running after both waits, and give the job id so the result can be fetched later with: node "$comp" result <job-id> --cwd "${repoPath}" --json
+   - Any terminal state ("completed", "failed", "cancelled") → it finished while you were between calls. Handle it under step 4 or step 5; do not treat the earlier timeout as the verdict.
+   Why this exists: measured on a real run, the lane hit .waitTimedOut at the full 540000 ms while the job was demonstrably mid-work (its log showed pinned git greps completing, exit 0), the job was STILL "running" long after the review was abandoned, and its result remained retrievable. Reporting "exhausted budget" there discarded a required lane, and with it the whole review, over a job that simply had not finished yet. Re-dispatching would have been wrong — waiting again was not.
+   Never substitute your own review for an unavailable Codex lane in either case — an unavailable lane is an expected outcome the workflow handles.
+   The 600000 ms outer ceiling is not tunable: the wait runs inside a Bash call, and the Bash tool silently kills any foreground command at that point. The inner wait is 540000 ms precisely so it finishes and prints its JSON before that happens — do not raise it to match. Step 6 is how a job legitimately exceeds ten minutes: not a longer wait, but a second one across a fresh call, which is the "polling across several calls" the ceiling leaves open. Total budget for a live job is therefore two waits, about eighteen minutes.
+Every Codex task you dispatch must carry the execution rules verbatim in its own prompt — Codex runs in its own shell and does not inherit this one's constraints. That includes the shell contract stated elsewhere in this prompt: it arrives via NO_EXECUTION, which every prompt builder composes exactly once, so it is deliberately not repeated here.
+Sandbox: the review itself runs sandboxed. The one known exception is the companion's own job log under ~/.claude/plugins/data, which is sandbox-denied — if dispatch fails on that write, bypass for that call only. Never bypass for anything else, and never for a call that performs a Git operation, mutates the working copy, or writes to GitHub. Reading a path is not the boundary; acting on it is.`
 
 const REVIEW_FOCUS = {
   'code-change': `Review for bugs, regressions, broken consumers, security issues, and instruction/config compliance.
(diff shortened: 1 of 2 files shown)


.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -171,7 +171,12 @@ Codex dispatch protocol (mandatory):
...
+   - EXHAUSTED BUDGET — .waitTimedOut is true. The job was still queued/running at the cap, so it needs more time, not another attempt; retrying doubles wall-clock and usually times out again. Do NOT retry. Return status:"unavailable" with .job.status, .waitTimedOut, and whether the job log showed active progress.
+   Retry budget is exactly one, and only for the fast-terminal-failure case. Never loop.
+   Never substitute your own review for an unavailable Codex lane in either case — an unavailable lane is an expected outcome the workflow handles.
+   The 10-minute cap in step 3 is not tunable: that wait runs inside a Bash call, and the Bash tool silently kills any foreground command at 600000 ms.
```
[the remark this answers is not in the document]
yuanchen8911: Confirmed and fixed in 6942d69ff.

This was the sharpest of the three. The retry decision in step 5 reads `.waitTimedOut`, so with the inner wait and the outer Bash timeout both at 600000 ms, Bash could kill the command before it printed any JSON — leaving nothing to classify on. The dangerous direction is that an unclassifiable kill looks like a fast terminal failure and gets retried, which is exactly the wasted full window the split was introduced to avoid.

Two changes:

- Inner wait lowered to `--timeout-ms 540000`, 60s below the outer cap, with the reason recorded inline so it does not get "tidied" back to matching values.
- Added an explicit rule: if the call produces no parseable JSON at all, treat it as exhausted-budget and do not retry.

The `SKILL.md` operational note was updated too, since it previously said only that the cap is not tunable and did not explain why the inner wait is deliberately lower.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -251,8 +256,23 @@ const CODERABBIT_RUN = `Run this WHOLE sequence in a single Bash invocation so t
...
-Timebox: run this Bash call with an explicit timeout of 600000 ms. The Bash tool defaults to 2 minutes and is capped at 10, so an unset or larger timeout silently kills the run. If it does not finish inside that window, check the newest file in ~/.coderabbit/logs/ for 429/queue lines and return status:"unavailable" with what you saw in statusNote — CodeRabbit is best-effort and the review continues without it.`
+Accept the run ONLY if its reported baseCommit equals ${baseSha}; otherwise discard it and return status:"unavailable" with what it reported. Ignore currentBranch, baseBranch, and workingDirectory — a detached worktree correctly reports currentBranch:"HEAD", the CLI's inferred baseBranch stays unrelated even when --base-commit is honored, and the commit pair is what identifies the reviewed context.
+Timebox: run this Bash call with an explicit timeout of 600000 ms. The Bash tool defaults to 2 minutes and is capped at 10, so an unset or larger timeout silently kills the run.
```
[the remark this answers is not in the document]
yuanchen8911: Confirmed and fixed in 6942d69ff.

Worth flagging that this thread was auto-marked "Addressed in commit 877b216". It was not — that commit only added `nullglob` to the store scan and did not touch the exit-status handling. The finding was still live when re-checked against the branch head, so it was verified against the current code rather than taken as resolved.

Two changes:

- The review invocation now captures its status (`rc=0; coderabbit review ... || rc=$?`) instead of letting `set -e` abort the block, so a non-zero exit still reaches recovery.
- The recovery step is now explicitly documented as a separate Bash call, never appended to the block above — that block may already have been killed at the timebox, and its EXIT trap has by then removed the worktree.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -251,8 +256,23 @@ const CODERABBIT_RUN = `Run this WHOLE sequence in a single Bash invocation so t
...
+Therefore run the coderabbit command with sandbox bypass (dangerouslyDisableSandbox) from the start; the observed evidence above is the justification the sandbox rules require. Bypass ONLY this command. Never bypass anything that touches the working copy or GitHub.
+Do NOT read a "connecting" stall as an outage, a 429, or contention with another session — a concurrent run on the same account was ruled out as the cause (an unsandboxed run succeeded while a sandboxed one hung, which merely looks like contention).
+
+RECOVERY — always try this before returning unavailable, and also whenever your own run stalls or is killed at the timebox. The CLI persists every completed review to ~/.coderabbit/reviews/<a>/<b>/reviews/<id>/, where git.json carries "head" and "baseCommitId". Search for a record with head == ${headSha} AND baseCommitId == ${baseSha}:
+   for g in ~/.coderabbit/reviews/*/*/reviews/*/git.json; do
+     python3 -c "import json,sys; d=json.load(open(sys.argv[1])); print(sys.argv[1]) if d.get('head')=='${headSha}' and d.get('baseCommitId')=='${baseSha}' else None" "$g"
+   done
```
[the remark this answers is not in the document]
yuanchen8911: Fixed across two commits — the finding had two halves and only one was closed initially.

The empty-store half (unmatched glob passing through literally, so `python3` opened the pattern and printed a `FileNotFoundError` traceback) was addressed in 877b216 via `shopt -s nullglob`, with `shopt -u` after the loop since `nullglob` is shell-wide rather than loop-scoped.

The malformed-record half was still open and is fixed in 6942d69ff: `json.load` on a half-written record raised `JSONDecodeError`, which under `set -e` stranded the log fallback that runs when no record matches. The scan now skips unparseable records and continues.

Verified with a directory containing one valid and one truncated record: the loop completes under `set -euo pipefail` and returns only the valid match.

yuanchen8911:
@coderabbitai full review

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -456,24 +767,93 @@ function intake(f, source, filesChecked) {
...
+}
+// Same batch rule as every other required lane; the only extra is the consumer-side
+// filter above, whose count is reported separately. Do NOT infer a reason by subtracting
+// counts: own-coordinate and evidence failures are distinct and were being reported as
+// one, which sent the reader after the wrong defect.
+const integBatch = intakeBatch(integUsable, 'integration', integ.filesChecked)
+const integReturned = (integ.findings || []).length
+if (integReturned && !integBatch.accepted) {
+  return incomplete(zeroSurvivors('integration', { ...integBatch, returned: integReturned },
+    `, and ${integDropped.length} lacked a usable consumerPath/consumerLine. An integration finding is a claim that a specific consumer breaks and cannot be verified without both ends located and evidence`))
```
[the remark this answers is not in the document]
yuanchen8911: This is a parser artifact rather than a defect, so no change was made.

The script is executed by Claude Code's Workflow tool as an async function body, where top-level `return` is valid and is how the workflow result is produced. `node --check` parses the file as a standalone ES module, which is not how it runs, so the reported error is expected there.

Evidence:

- Parsed in the actual execution context — `new (Object.getPrototypeOf(async function(){}).constructor)(...)` over the file contents — it compiles with no error.
- The same four module-level returns are present on `main` at `443559ab4`, unchanged by this PR. No enclosing function was removed, so there is none to restore.
- The file records this at lines 15-17: "runs in a custom async execution context where top-level `return` statements are valid (they return the workflow result). Standard JS linters may flag those as parse errors; that is expected."
- The script has run to completion four times against real pull requests while this branch was being developed, which would not be possible if the returns were illegal.

Adding an enclosing function would break the script: the workflow result is returned by exactly those statements.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -560,11 +1009,14 @@ if (pending.length) {
...
-    for (const f of res.newFindings || []) {
-      const c = intake(f, k, res.filesChecked)
-      if (c && !pending.includes(c.id)) pending.push(c.id)
+    // newFindings go through the same intake gate as round 1 — an unevidenced or
+    // unlocatable late finding must not enter the tally either — and the same
+    // zero-survivor rule: a required lane raising only malformed late findings has
+    // contributed nothing, and letting them vanish silently is the same false-clean.
+    const nb = intakeBatch(res.newFindings, k, res.filesChecked)
+    for (const c of nb.candidates) if (!pending.includes(c.id)) pending.push(c.id)
+    if (nb.returned && !nb.accepted) {
```
njhensley: **🟠 Major — newFindings zero-survivor aborts a completed run over one malformed late finding**

`nb = intakeBatch(res.newFindings, ...)` then `if (nb.returned && !nb.accepted) bad = bad || zeroSurvivors(...)` → `if (bad) return incomplete(bad)`. `nb` is computed only from `newFindings`; the lane's `evaluations` are consumed separately (1002–1011). `FINDING_ITEM` makes `line` an integer with no minimum and `evidence` a string with no `minLength`, so a schema-valid supplementary late finding with `line: 0` or `evidence: ""` yields `returned=1, accepted=0`. A cross-reviewer that delivers every correct evaluation plus one imprecise volunteered late finding thus returns `incomplete` before the tally and verification even run — discarding all adjudication, no report produced. That is the exact PR-1908 symptom the integration per-finding rework removed, reintroduced through the cross-review path. The round-1 rationale ("the lane contributed nothing") is false here: `newFindings` is supplementary to the evaluations the lane did deliver, and lanes are explicitly prompted to emit them.

**Blast radius:** Any cross-review run where a required lane volunteers a single line-less or evidence-less late finding → total loss of a completed, expensive run. Recoverable by re-run, but non-deterministic.

**Fix:** Drop malformed `newFindings` individually and log them — exactly as the integration lane already does at 899–902 — rather than treating a `newFindings` total-loss as fatal; or gate the stop on the lane also having returned zero *evaluations*, the only true "contributed nothing" state at this point.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -422,8 +751,33 @@ function addCandidate(f, source) {
...
+  // Same location, different wording — the case location-only keying was meant to fix.
+  // Flag it instead of merging: the flag reaches both the cross-review candidate list and
+  // the refuter prompt, so reviewers decide whether the two are one defect and evaluate
+  // them consistently. That keeps equivalence an explicit judgement rather than an
+  // assumption from a shared line number, and keeps one verdict per real defect.
+  // The hint must respect the SAME identity the key does, consumer coordinates included.
+  // One changed declaration breaking two callers is deliberately two candidates; hinting
+  // that they are possible duplicates would push reviewers to collapse a distinction the
+  // key exists to preserve. Only a genuine same-location, same-consumer pair is hinted.
+  const sameSpot = (a, b) => a.path === b.path && a.line === b.line
```
njhensley, line 768: **🔵 Nitpick — possible-duplicate flag doesn't collapse a same-defect double-confirmation**

Two lanes wording one defect differently produce distinct `normSummary` → two candidates at the same spot, each carrying only an advisory flag. Both can reach 2-of-3 AGREE and pass verification, so one real defect can appear twice in `confirmed[]`. Documented as the deliberate reversal of location-only keying; the actual comment collapse would live in `post_review.py` (outside this diff).

**Blast radius:** Cosmetic double-representation of a confirmed finding; no false/overstated consensus.

**Fix:** Optional: when both members of a flagged pair resolve to `confirmed`, emit one merged finding (union of sources/evidence).

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -456,29 +823,108 @@ function intake(f, source, filesChecked) {
...
 for (const f of integ.findings || []) {
-  if (!f.consumerPath || f.consumerLine == null) {
-    return incomplete(`Integration lane returned a finding at ${f.path}:${f.line} ("${f.summary}") without consumerPath/consumerLine. An integration finding is a claim that a specific consumer breaks and cannot be verified without both.`)
-  }
+  // Presence is not usability, and the CONSUMER side needs the same rule as the finding's
+  // own coordinates — same schema, same absent constraints, same false-clean if it slips.
+  // hasCoords is shared with intake() precisely so the two cannot drift apart again.
+  // The finding's own path/line are checked in intake(), so a finding reaching consensus
+  // has both ends locatable.
+  if (!hasCoords(f.consumerPath, f.consumerLine)) integDropped.push(f)
```
njhensley, line 896: **🔵 Nitpick — consumer-less integration finding is dropped (documented design, logged)**

An integration finding with valid own path/line/evidence but no `consumerPath/consumerLine` goes to `integDropped` (890–898), is logged (899–901), and never becomes a candidate — so a real non-consumer defect only the integration lane spots is lost while `status` stays `ok`. This is the documented consumer-pair-only contract; the drop is logged (not silent) and the general lanes review the same code.

**Blast radius:** Low: bounded to a non-consumer defect that ONLY the integration lane raises and both general lanes miss.

**Fix:** Optional: route consumer-less integration findings into the normal candidate pool as ordinary findings so they still get cross-reviewed.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -446,8 +800,21 @@ function flagUnchecked(c, f, source, filesChecked) {
 // its reviewer as a source, after which evidenceFor() hands it the co-reporter's
 // citation and it silently tips the 2-of-3 tally. Rejecting at intake means an
 // unevidenced report never becomes a vote anywhere.
+// ONE coordinate rule, used for a finding's own path/line and for an integration
+// finding's consumerPath/consumerLine. The schema requires both fields but constrains
+// neither — no minLength, no minimum — so "", "   ", 0 and -1 all satisfy it and all
+// used to pass a truthiness/null check. Every previous version of this guard fixed the
+// pair it was shown and left the other, which is why this is a shared helper rather than
+// two call-site conditions: the next field pair added should reuse it, not re-derive it.
+const hasCoords = (p, l) => typeof p === 'string' && p.trim() !== '' && Number.isInteger(l) && l > 0
```
njhensley, line 809: **🔵 Nitpick — line:0 file-level finding dropped as unlocatable; fatal if lane's only finding**

`hasCoords` requires `l > 0`, so a schema-valid whole-file finding at `line: 0` drops as unlocatable (832) and, if it's a required lane's only finding, round-1 zero-survivor (869) returns incomplete. Deliberate (the guard enumerates 0/-1 as reject values) and agents normally anchor whole-file findings at line 1. A facet of the major finding above.

**Blast radius:** Low edge case.

**Fix:** Optional: accept `line === 0` as a file-level sentinel, or don't count a lineless-but-evidenced finding as "contributed nothing".

njhensley (dismissed):
## Multi-Persona Review — Approve with comments

**Method.** Four independent persona reviewers (Correctness, Domain/Consensus-semantics, Robustness/Operability, Docs-consistency) reviewed the resolved code in parallel; every finding was then re-derived from source by an adversarial senior meta-reviewer that confirmed, refuted, or re-tiered it.

### Overall

High-quality reliability hardening of the `aicr-cross-review` workflow. The consensus/candidate JS logic is correct and unusually well-defended — I traced the candidate lifecycle end-to-end and found **no false-consensus, lost-finding, vanishing-candidate, or crash bug** in the core machinery, and the headline invariant (`consensusReached` cannot be true while a candidate silently disappears) holds. The embedded zsh reaper and the persisted-review Python block were independently validated (rename `-z` token indexing, mode-before-OID identity, fail-closed on git error, no surviving backslash-escapes); the sandbox-bypass boundary is narrow and consistent between `SKILL.md` and `workflow.mjs`; and every load-bearing doc claim matches the code.

The one finding worth acting on before merge (🟠) is notable precisely because it undercuts this PR's own goal — "stop throwing away work it has already done": one path still discards a fully-completed cross-review run over a single supplementary malformed late finding. Everything else is nitpick-level. Approving with comments.

### Inline comments
- 🟠 **A** — `workflow.mjs:1018` — `newFindings` zero-survivor aborts a completed run over one malformed late finding *(the one to consider before merge)*
- 🔵 **B** — `workflow.mjs:768` — possible-duplicate flag doesn't collapse a same-defect double-confirmation *(retiered 🟡→🔵)*
- 🔵 **C** — `workflow.mjs:896` — consumer-less integration finding dropped *(documented, logged; retiered 🟡→🔵)*
- 🔵 **D** — `workflow.mjs:809` — `line:0` file-level finding dropped as unlocatable; fatal if lane's only finding

### Additional 🔵 code-quality notes (no action required)
- **E** `workflow.mjs:771` — the possible-duplicate loop mutates `flags` on an already-adjudicated candidate; cosmetic (flags never feed `tally`). Could skip when `presentedIds.has(other.id)`.
- **F** `workflow.mjs:389` — `subprocess.run(text=True)` newline translation could corrupt a `\r` path token; unreachable in practice (`core.quotePath`) and fail-closed.
- **G** `workflow.mjs:1093` — `PINNED_READS` is emitted twice per prompt (standalone + via `NO_EXECUTION`); harmless, unlike the `SHELL_CONTRACT` doubling the code correctly guards against.

- **Consensus core:** dedupe key + `presentedIds` merge-guard, collision-free ID generation, `byKey` first-holder retention, intake accounting (`returned == accepted + unlocatable + unevidenced`), 2-of-3 tally with `verdictFor` source-fallback supplying only CodeRabbit's real round-1 vote, late findings never tallied (short-circuit at 1031), `consensusReached` invariant — all sound.
- **Embedded Python fallback:** no surviving backslash-escape, `numstat`/`rawmap` `-z` rename token indexing correct, modes-before-OIDs identity, `no-index-line`/`rename-unverifiable` handling, fail-closed `sys.exit(2)` on git error — verified.
- **zsh shell:** `read -r -d ''` reaper loop, `find -mmin +120` age gate, `set -euo pipefail`, `${TMPDIR:-/tmp}` handling — tested correct under zsh 5.9.
- **Sandbox boundary:** exactly two narrow bypass exceptions, consistent across both files.
- **Docs:** timeouts, dedupe key, sandbox exceptions, `SHELL_CONTRACT` single-interpolation, persisted-store fallback — `SKILL.md` matches `workflow.mjs` on every load-bearing claim. Version bump `0.3.0 → 0.3.18` intentional. `.claude/skills` symlinks `.agents/skills` — no un-synced mirror.

```
🔴 Blocker 0 | 🟠 Major 1 | 🟡 Minor 0 | 🔵 Nitpick 6      Recommendation: Approve with comments
```

.agents/skills/aicr-cross-review/scripts/workflow.mjs
yuanchen8911: Confirmed and fixed in 0ebd1d1db.

Both halves check out. The mechanism: `nb` is computed only from `res.newFindings`, the lane's `evaluations` are consumed separately just above, and `if (bad) return incomplete(bad)` sits before the tally and the verification round. Reachability: `FINDING_ITEM` declares `line: {type: 'integer'}` with no `minimum` and `evidence: {type: 'string'}` with no `minLength`, so `line: 0` and `evidence: ""` are both schema-valid and both land as `returned=1, accepted=0`.

The point that settled it is the one about the round-1 rationale not transferring. The completeness gate immediately above already returns `incomplete` unless the lane returned exactly one evaluation per presented candidate, so by the time control reaches that line the lane has provably contributed its entire evaluation set. "The lane contributed nothing" is false by construction there, which makes the comment justifying the guard wrong rather than just the guard being miscalibrated.

Took the first option: malformed late findings are now dropped individually. No new logging was needed since `intake()` already logs each drop with its specific reason, so this stays non-silent in the way the integration lane is. The zero-survivor rule is untouched for round 1 and for the integration lane, where the "whole output" premise does hold.

Verified against the shipped guards that a `line: 0` and a blank-evidence late finding each yield `returned=1, accepted=0` — fatal before, dropped-and-logged now — and that a mixed batch still admits the good finding.

`SKILL.md` was carrying the same wrong claim ("applies to every required batch ... and their cross-review `newFindings`"), so it now scopes the rule to round-1 batches and states why the cross-review round is deliberately excluded.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
yuanchen8911: Verified, and the fatal half is gone as of 0ebd1d1db.

You had this right as a facet of A rather than a separate defect. With late findings now dropped individually, a `line: 0` finding raised during cross-review no longer takes the run down with it — it drops and is logged.

Leaving the `hasCoords` behavior itself as is. Rejecting `0` is deliberate rather than incidental: the comment above the helper enumerates `""`, `"   "`, `0` and `-1` as exactly the values the schema permits and the guard is there to reject, and treating `0` as a file-level sentinel would reintroduce a value that has to be special-cased at every site that consumes a line number. Agents anchor whole-file findings at line 1 in practice, and the drop is logged with the offending `path`/`line` so a lane that gets it wrong is visible rather than silent.

That leaves only the round-1 case, where a lane whose sole finding is at `line: 0` still returns `incomplete`. That one I think is correct: findings are the whole output in round 1, so it genuinely is a lane that contributed nothing locatable.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
yuanchen8911: Verified, not changing it here.

The mechanism is real: two lanes wording one defect differently produce distinct `normSummary`, so `sameSpot` flags the pair but `byKey` keeps them as separate candidates, and both can clear 2-of-3 and land in `confirmed[]`.

Keeping the flag-don't-merge behavior, for the reason the code comment gives: collapsing on a shared location is what the key change was undoing, and one changed declaration breaking two consumers is deliberately two candidates. Auto-merging a flagged pair at consensus time would put that judgement back in the workflow, which is where it was getting it wrong.

Agreed the real fix is a display-side collapse, and as you note that lives in `post_review.py`, outside this diff. Leaving it out of scope here rather than growing this PR into the comment-rendering path.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
yuanchen8911: Verified, keeping the current behavior.

Accurate description of the path: a consumer-less integration finding goes to `integDropped`, is logged, and never becomes a candidate, while `status` stays `ok`.

Keeping it as the documented contract. An integration finding is a claim that a specific consumer breaks, so without both ends located there is nothing for a cross-reviewer or a refuter to check — routing it into the general pool would admit a claim that cannot be evaluated on its own terms. The drop is logged with each `path:line`, so a lane producing them is visible.

The residual exposure is the one you scoped it to: a non-consumer defect that only the integration lane sees and both general lanes miss. Narrow enough that I would rather leave it than widen what enters consensus in this PR. Worth a follow-up if it ever shows up in a real run — happy to file one if you would like it tracked.

njhensley (approved):
## Multi-Persona Re-Review — Approve ✅ (`0ebd1d1d`)

Re-review after the force-push. The PR's diff vs `main` remains exactly the two skill files (`SKILL.md`, `workflow.mjs`); the Go `inference_perf_constraint.go` change in the commit-to-commit delta is rebase noise (landed on `main` independently), not part of this PR.

### 🟠 A (from my prior review) — RESOLVED

The fatal `newFindings` zero-survivor guard in the cross-review round was removed (`workflow.mjs` ~1025). Verified sound:

- **No new false-clean.** The completeness gate above already returns `incomplete` unless the lane evaluated every presented candidate, so "the lane contributed nothing" is unreachable at this point — which is exactly why the round-1 zero-survivor rule doesn't transfer here.
- **Not silent.** Malformed late findings still pass through `intakeBatch → intake()`, which logs each drop with its reason and keeps it out of the tally — as the integration lane does.
- **`bad` still live** for the genuinely-fatal unevidenced-AGREE/DISAGREE case; `if (bad) return incomplete(bad)` remains.
- **Docs match:** `SKILL.md` now scopes the rule to "every required **round-1** batch" and documents the deliberate cross-review exemption with the PR-1908 rationale. Version `0.3.18 → 0.3.19`.

### Remaining
The other findings (B/C/D/E/F/G) were all optional 🔵 nitpicks; none were blocking and none needed changes. No new issues introduced by the fix.

```
🔴 Blocker 0 | 🟠 Major 0 | 🟡 Minor 0 | 🔵 Nitpick 6      Clean approve
```

The one substantive item from the earlier review is fully addressed with a correct, well-reasoned fix. Approving.

PR #1936, opened 2026-07-28
fix(validator): disclose cordoned GPU nodes in check-nvidia-smi

### LINKED ISSUE #1668: Validator handles cordoned GPU nodes inconsistently across checks
## Description

The validator treats cordoned (unschedulable) GPU nodes inconsistently across checks:

- **Node-scoped checks that deploy test pods** (e.g. `check-nvidia-smi`) silently skip cordoned nodes: the check enumerates schedulable GPU nodes only, reports e.g. `Found 1 GPU node(s)` on a two-GPU-node cluster, and can pass with nothing in the report indicating reduced coverage.
- **Cluster-aggregate checks** (`gpu-operator-health`, `expected-resources`) include cordoned nodes: they assert on ClusterPolicy `status.state`, which the GPU operator computes across all nodes it manages — DaemonSet operands ignore cordons — so an unhealthy cordoned node fails these checks.

A cordon therefore narrows the scope of some checks and not others. Observed on a two-GPU-node bare-metal cluster with one node cordoned for maintenance: `check-nvidia-smi` validated 1/1 (silently excluding the cordoned node) while `expected-resources` and `gpu-operator-health` failed on the same node's operand state.

## Why it matters

- **Silent scope narrowing in evidence:** a passing validation run implies the cluster was validated, but the attestation does not record that a node was excluded. This is the spuriously-passing-check failure mode the project's own review rules treat as fail-closed territory.
- **Wrong signal semantics:** a cordon means "do not schedule new pods," not "this node is out of GPU service." Keying exclusion off cordons conflates transient maintenance with deliberate exclusion.

## Proposed behavior (skip-with-disclosure, never silent)

1. Node-scoped checks enumerate **all** GPU nodes, validate schedulable ones, and report cordoned nodes explicitly as `skipped: cordoned`.
2. The phase result and emitted report/attestation record coverage (e.g. `nodesValidated: 1/2`), so a pass on reduced scope is visible and auditable.
3. Optionally, a strict flag (e.g. `--require-all-nodes`) fails the phase when any GPU node was skipped — for release-gate attestations.
4. Document the GPU operator's `nvidia.com/gpu.deploy.operands=false` node label as the supported signal for *deliberate* exclusion — it consistently removes the node from both node-scoped checks (node stops advertising GPUs) and ClusterPolicy aggregation, unlike a cordon.

rsd-darshan:
## Summary

`check-nvidia-smi` now enumerates every GPU node, validates only the schedulable ones, and explicitly discloses cordoned nodes plus a coverage line — instead of silently narrowing scope with no signal in the evidence.

## Motivation / Context

A cordon (`Node.Spec.Unschedulable`) means "do not schedule new pods," not "exclude from GPU service" — it's commonly transient maintenance, not deliberate exclusion. `check-nvidia-smi` enumerated only the schedulable GPU cohort (`helper.FindSchedulableGpuNodes`), so a cordoned GPU node silently disappeared from the reported node count (`Found 1 GPU node(s)` on a two-GPU-node cluster) with nothing indicating reduced coverage. Meanwhile cluster-aggregate checks (`gpu-operator-health`, `expected-resources`) already include cordoned nodes via ClusterPolicy aggregation, so the two check families disagreed on scope with no disclosure of the mismatch — the spuriously-passing-check failure mode this project's own conventions treat as fail-closed territory.

Fixes: #1668
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

## Implementation Notes

- Added `helper.FindGpuNodes`, which returns every node with allocatable `nvidia.com/gpu`, each tagged with its cordon state. `helper.FindSchedulableGpuNodes` is now a thin filter on top of it, so its signature and behavior for existing callers (RDMA fabric gate, performance worker sizing) are unchanged.
- `checkNvidiaSMI` partitions the result into schedulable vs. cordoned, validates only the schedulable subset, and prints `<node>: skipped (cordoned)` plus `nodesValidated: <validated>/<total>` to stdout evidence (which flows into the CTRF report per the existing `// → CTRF evidence` convention).
- If a GPU node exists but every GPU node is cordoned, the check now skips with a reason that discloses the cordoned count (`all N GPU node(s) are cordoned; nothing to verify`) rather than reusing the generic "no GPU nodes found" message, so the two distinct conditions stay distinguishable in evidence.
- Cluster-aggregate checks (`gpu-operator-health`, `expected-resources`) are unaffected — they already cover cordoned nodes since DaemonSet operands ignore cordons.
- Documented the GPU Operator's `nvidia.com/gpu.deploy.operands=false` node label in `docs/contributor/validator.md` as the supported signal for *deliberate*, durable exclusion, as distinct from a cordon.
- Out of scope (left for follow-up if wanted): an opt-in `--require-all-nodes` strict flag for release-gate attestations, mentioned as optional in the issue.

## Testing

```bash
GOFLAGS="-mod=vendor" go build ./validators/...
GOFLAGS="-mod=vendor" go vet ./validators/...
GOFLAGS="-mod=vendor" go test -race ./validators/helper/... ./validators/deployment/...
golangci-lint -c .golangci.yaml run ./validators/helper/... ./validators/deployment/...
```

- `go test -race` — full pass, `validators/helper` and `validators/deployment`.
- `golangci-lint` (pinned v2.12.2) — 0 issues on both affected packages.
- New tests: `helper.TestFindGpuNodes`, `helper.TestFindSchedulableGpuNodes` (cordon partitioning), and `TestCheckNvidiaSMI_{NoGPUNodes,AllNodesCordoned,MixedCordonAndNonGPU}` (skip-reason disclosure).
- Note: `validators/` is excluded from the repo-wide `make test-coverage` floor (`Makefile` `test` target filters `-e /validators`), so the package-coverage delta gate in CONTRIBUTING does not apply here.
- Full `make qualify` (e2e/scan/license-check) was not run locally — those need `chainsaw`/`goreleaser`/`grype`/`go-licenses`, not present in this environment; deferring to CI.

## Risk Assessment

- [x] **Low** — Isolated change to one check's node enumeration and evidence output; existing callers of `FindSchedulableGpuNodes` are behaviorally unchanged.

**Rollout notes:** N/A — evidence-format addition only, no schema or API change.

## Checklist


rsd-darshan:
Addressed all 3 actionable findings in cd806b58:

1. **Table-driven test** (`nvidia_smi_test.go`) — collapsed the three skip-scenario tests into one table-driven `TestCheckNvidiaSMI_SkipScenarios` with subtests.
2. **Double error-wrap** (`nvidia_smi.go:47`) — `checkNvidiaSMI` now returns `helper.FindGpuNodes`'s error as-is instead of re-wrapping an already-correctly-coded value.
3. **Context cancellation in loops** (`gpu.go`, `nvidia_smi.go`) — added `ctx.Done()` checks in `FindGpuNodes`, `FindSchedulableGpuNodes`, and the cordon-partition loop in `checkNvidiaSMI`, matching the existing pattern in `rdmaFabricProbe`.

`go build`/`vet`/`test -race` and `golangci-lint` (pinned v2.12.2) all pass clean on `validators/helper` and `validators/deployment`.

validators/deployment/nvidia_smi.go
```diff
@@ -98,7 +127,9 @@ func checkNvidiaSMI(ctx *validators.Context) error {
 				len(failedNodes), len(gpuNodes), failedNodes))
 	}
 
-	fmt.Printf("Successfully verified GPU on all %d nodes\n", len(gpuNodes))
+	fmt.Printf("nodesValidated: %d/%d (%d cordoned, skipped)\n",
```
yuanchen8911: The coverage disclosure only reaches check stdout, and the default evidence path strips it. `pkg/evidence/redact/redact.go` sets `tr.Stdout = nil` and `tr.Message = ""` for every CTRF test, and `pkg/evidence/attestation/emit.go` applies that redaction unless `--full` is passed, which defaults to false.

In a published, signed bundle this check therefore appears as `passed` with no message and no stdout, so a 1-of-2-node pass is indistinguishable from a 2-of-2 pass, and an all-cordoned run loses its skip reason.

This line also lacks the `RESULT: ` prefix that `pkg/validator/validator.go` uses to echo summary lines into live CLI output, so the coverage figure is invisible during the run as well.

`docs/contributor/validator.md` says the line makes a reduced-scope pass visible in the CTRF report. That holds for `aicr validate --output`, but not for the artifact a downstream consumer verifies. Carrying the counts in a structured CTRF field rather than stdout would survive redaction.

docs/contributor/validator.md
```diff
@@ -393,6 +393,41 @@ constraint, which is an explicit anti-pattern in CLAUDE.md.
...
+fewer nodes than the cluster has, with nothing in the evidence
+indicating reduced coverage. That is the spuriously-passing-check
+failure mode CLAUDE.md's anti-pattern table calls out, applied to node
+enumeration instead of a single boolean.
+
+Checks that deploy a per-node test pod (`check-nvidia-smi`) must
+instead call `helper.FindGpuNodes`, which returns every GPU node
+tagged with its cordon state, and:
+
+1. Validate only the schedulable subset (a cordoned node cannot
```
yuanchen8911: The stated rationale is not correct for this probe. `validators/deployment/testdata/nvidia-smi-verify-pod.yaml` sets `nodeName: ${NODE_NAME}`, so the pod is bound at creation and admitted directly by the kubelet. `Node.Spec.Unschedulable` is a scheduler-side predicate and does not apply, and the template also tolerates all taints. The pod would run fine on a cordoned node.

Line 419 of this same file relies on exactly that principle for DaemonSet operands, so the document currently argues both ways.

If excluding cordoned nodes is a deliberate policy choice, it would be clearer to say so. As written the check gives up coverage it can actually obtain, and future check authors may copy the reasoning.

docs/contributor/validator.md
```diff
@@ -393,6 +393,41 @@ constraint, which is an explicit anti-pattern in CLAUDE.md.
...
+   schedule the probe pod).
+2. Report cordoned nodes explicitly in the check's stdout evidence
+   (`<node>: skipped (cordoned)`), never omit them from the node count.
+3. Print a coverage line (`nodesValidated: <schedulable>/<total>`) so a
+   pass on reduced scope is visible in the CTRF report.
+
+Cluster-aggregate checks that assert on an operator's aggregate status
```
yuanchen8911: `expected-resources` is not exempt. `validators/deployment/expected_resources.go:1003` (`rdmaFabricProbe`, reached from `checkExpectedResources`) calls `helper.FindSchedulableGpuNodes` at line 1007, and the comment at line 995 states that cordoned nodes are excluded. On success it prints `RDMA fabric (%s): allocatable (uniform) on all %d RDMA GPU node(s)` at line 989, where the count is the schedulable cohort only.

That is the same silently-narrowed pass this section defines as the anti-pattern, with no cordon disclosure and no coverage line. A contributor auditing node-scoped checks against this list would skip `expected-resources` on the document's authority.

validators/deployment/nvidia_smi.go
```diff
@@ -98,7 +127,9 @@ func checkNvidiaSMI(ctx *validators.Context) error {
 				len(failedNodes), len(gpuNodes), failedNodes))
```
yuanchen8911, line 188: The coverage disclosure is emitted only on the success path. This failure return, and the busy-skip return above it, report a node count without the cordon qualifier, so a run that fails after validating a narrowed subset still describes its scope as if it were complete.

Rule 2 of the new documentation section applies to these paths too.

validators/helper/gpu.go
```diff
@@ -46,20 +46,62 @@ const AKSRdmaSharedResource = "rdma/hca_shared_devices_a"
...
 		return nil, aicrErrors.Wrap(aicrErrors.ErrCodeInternal, "failed to list nodes", err)
 	}
 
-	var gpuNodes []v1.Node
+	var gpuNodes []GpuNode
 	for _, node := range nodeList.Items {
-		if node.Spec.Unschedulable {
-			continue
+		select {
+		case <-ctx.Done():
```
yuanchen8911, line 76: Context cancellation inside this loop is wrapped as `ErrCodeTimeout`, but the existing callers of `FindSchedulableGpuNodes` re-wrap the returned error as `ErrCodeInternal`, so the distinction is lost before it reaches a caller that could act on it. Worth checking whether this should propagate as-is, per the guidance about not re-wrapping an error that already carries the right code.

Separately, both `ctx.Done()` checks run over slices already materialised in memory by the preceding `List` call, so they cannot interrupt any I/O — they only affect how a cancellation that already happened is labelled.

validators/helper/gpu_test.go
```diff
@@ -144,3 +163,64 @@ func TestCheckPodRunningOrTerminal(t *testing.T) {
 		})
 	}
 }
+
+// TestFindGpuNodes proves FindGpuNodes reports every GPU node — schedulable
+// and cordoned alike — tagged with its cordon state, so callers that must
+// disclose cordoned nodes (rather than silently drop them, issue #1668) have
+// the full set to work from.
+func TestFindGpuNodes(t *testing.T) {
```
yuanchen8911, line 171: `TestFindGpuNodes` checks membership against a map without consuming matched entries, so a duplicated result can stand in for a missing expected node. `TestFindSchedulableGpuNodes` at line 204 has the same shape: it asserts the cordoned node is absent but never that both schedulable nodes are present.

Neither fixture covers a node advertising `nvidia.com/gpu` at quantity `0`, which is a case `FindGpuNodes` has to decide on.

validators/deployment/nvidia_smi_test.go
```diff
@@ -15,12 +15,20 @@
...
 import (
+	"strings"
 	"testing"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// isSkipError reports whether err came from validators.Skip. Skip wraps the
```
yuanchen8911: `isSkipError` classifies by string-matching the rendered suffix `": skip"` because the sentinel is unexported. That is brittle for a fail-closed path: any change to how `validators.Skip` renders turns a real failure into a silent skip in these tests.

This mirrors the existing helper in `validators/conformance/dra_support_check_test.go`, so it follows precedent, but the pattern now exists in two places. Exporting the sentinel and using `errors.Is` would make it robust.

validators/deployment/nvidia_smi_test.go
```diff
@@ -109,3 +117,59 @@ func TestVerifyNvidiaSMILogs(t *testing.T) {
...
 }
+
+// TestCheckNvidiaSMI_SkipScenarios covers the enumeration/disclosure paths
+// that return before any pod is scheduled: no GPU nodes at all must give a
+// different, more generic skip reason than every GPU node being cordoned
+// (issue #1668 — a cordon narrows scope, it must never look identical to
+// "nothing to check"), and non-GPU nodes must never count toward either the
+// schedulable or cordoned tally.
+func TestCheckNvidiaSMI_SkipScenarios(t *testing.T) {
+	t.Parallel()
```
yuanchen8911, line 125: No test covers the mixed cordoned-plus-schedulable path, which is the headline behaviour of this change: the per-node `skipped (cordoned)` lines and the `nodesValidated: n/m` figure are never asserted. The scenarios here cover the all-cordoned and no-GPU-node cases only.

yuanchen8911 (commented):
A few issues from a review of this change, mostly around whether the new disclosure actually reaches the surfaces that matter. Inline comments below; three items could not be anchored inline because their files are not in this diff:

The unit tests added here never execute in CI. `Makefile:229` runs `go test` over `$(go list ./... | grep -v -e /tests/chainsaw/ -e /validators)`, which filters out both `validators/deployment` and `validators/helper`. `.github/actions/go-test/action.yml:72` is the only CI step that runs Go unit tests and it invokes `make test`. So `TestCheckNvidiaSMI_SkipScenarios`, `TestFindGpuNodes` and `TestFindSchedulableGpuNodes` are compiled by `go vet` and golangci-lint but never run, and the coverage gate cannot see the new exported `FindGpuNodes` and `GpuNode` surface. PR #1918 appears to address this exclusion.

The check catalog still overclaims coverage. `recipes/validators/catalog.yaml:63` and `recipes/validators/README.md:43` both describe `check-nvidia-smi` as verifying nvidia-smi on all GPU nodes, while the check now covers the schedulable subset only.

The same narrowing pattern exists in the performance validators. `validators/performance/nccl_all_reduce_bw_constraint.go:729` and `validators/performance/inference_perf_constraint.go:641` size their workloads from `helper.FindSchedulableGpuNodes` with no cordon disclosure and no coverage line, so the new documentation section's taxonomy does not cover them.

rsd-darshan:
Addressed both CodeRabbit findings from the 2nd pass in 977854ea:

1. **Coverage line only on success path** — refactored the disclosure into a pure `gpuNodeCoverage` struct (`partitionGpuNodes`/`enumerationLines`/`coverageLine`) so `nodesValidated` now prints on every exit: all-cordoned skip, busy skip, failed validation, and success.
2. **Route output through a configured writer** — declined. This check (and every other check in `validators/deployment`) writes evidence via direct `fmt.Printf`/`fmt.Println` per the documented convention (`// → CTRF evidence`, see `docs/contributor/validator.md`); introducing a writer abstraction for just this one check would be inconsistent with the rest of the package and isn't something this PR's scope calls for.

`go build`/`vet`/`test -race` and `golangci-lint` all pass clean.

rsd-darshan:
Thanks for the thorough review — went through each point and pushed fixes in 977854ea:

1. **Tests never run in CI** — confirmed: my branch forked from before `#1918` (merged today) fixed the `Makefile` `/validators` exclusion from `make test`. Since `#1918` is already on `main`, the merge-ref this PR's CI actually tests against will include it, so `TestCheckNvidiaSMI_SkipScenarios`/`TestFindGpuNodes`/`TestFindSchedulableGpuNodes` (and the new tests added in this round) will run. No action needed here since it's already resolved upstream — happy to double check once CI actually runs on this PR.

2. **Catalog/README overclaim coverage** — fixed. `recipes/validators/catalog.yaml` and `recipes/validators/README.md` now say "all schedulable GPU nodes (cordoned nodes are disclosed, not silently skipped)".

3. **Coverage disclosure stripped by default redaction** — confirmed, and it's a real gap: the `nodesValidated` line only reaches `TestResult.Stdout`/`.Message`, which `pkg/evidence/redact`'s default "minimal" policy strips. I corrected the doc to say this explicitly (visible for live/`--output`/`--full`, not guaranteed in a default signed bundle) rather than overclaiming, and filed #1951 to track carrying this kind of outcome data in a structured field that survives redaction — that's a cross-cutting change to the shared CTRF/evidence pipeline, so it didn't feel right to fold into this PR's scope.

4. **Cordon rationale is factually wrong for this probe** — confirmed and fixed. You're right that `nodeName` + tolerate-all bypasses the scheduler entirely, so `Unschedulable` never actually blocks this pod. Reworded the doc section to state plainly that excluding cordoned nodes here is a deliberate policy choice (respecting drain/maintenance intent), not a technical necessity, and to flag that a future check with a different tradeoff should say so explicitly rather than assume cordons are unconditionally unschedulable.

5. **`rdmaFabricProbe` has the same undisclosed narrowing** — confirmed, my "cluster-aggregate checks are unaffected" claim was too broad. Narrowed the doc to call out that `rdmaFabricProbe` is itself node-scoped with the identical gap, and filed #1952 to track applying the same disclosure pattern there (and auditing the performance workload-sizing call sites you flagged, which are a different kind of narrowing — sizing vs. "verify every node" — so I didn't want to assume they need identical treatment without a closer look).

6. **Coverage line missing on busy/failed exit paths** — fixed as part of the same refactor in point 3/1 above: `gpuNodeCoverage.coverageLine()` is now called from every return path.

7. **`ctx.Done()` checks over an already-materialized slice, and the ErrCodeTimeout distinction gets lost when re-wrapped as ErrCodeInternal by existing `FindSchedulableGpuNodes` callers** — left as-is. Agreed this is largely cosmetic for the in-memory loop itself, but it matches the existing `rdmaFabricProbe` precedent in this same file for a near-identical loop, and removing it now would mean the two loops disagree on convention. Not worth relitigating in this PR.

8. **Test correctness: map-membership check doesn't catch a duplicate masking a missing node** — fixed. `TestFindGpuNodes`/`TestFindSchedulableGpuNodes` now use a seen-set to explicitly reject duplicates and assert every expected node is present, and added a zero-allocatable-GPU node fixture.

9. **`isSkipError` string-matching is brittle** — fixed, but scoped to this PR's own files: exported `validators.IsSkip` (uses `errors.Is` against the sentinel) and switched `nvidia_smi_test.go` to it. Left `validators/conformance/dra_support_check_test.go`'s copy alone since that file isn't otherwise touched by this PR.

10. **Missing headline test: mixed cordoned+schedulable disclosure/coverage** — added `TestGpuNodeCoverage_MixedCordonedAndSchedulable`, which exercises `partitionGpuNodes`/`enumerationLines`/`coverageLine` directly against a mixed node set. I went this route rather than a full pod-lifecycle simulation (no existing check test in this repo simulates `WaitForPodSucceeded`+`GetLogs` via the fake clientset, and building that from scratch felt like a bigger, more fragile lift than the disclosure logic itself warranted) — the refactor in point 6 pulls the disclosure into pure functions specifically so this scenario is testable without needing that machinery.

`go build`/`vet`/`test -race` (full `validators/...` tree) and `golangci-lint` all pass clean.

rsd-darshan:
One sub-point I'd only partially addressed: the coverage line lacking the \`RESULT: \` prefix (\`pkg/validator/validator.go\`'s \`resultSummaryPrefix\`), so it was invisible during a live run regardless of redaction — not just stripped from the signed bundle. Fixed in cfc271c8: \`coverageLine()\` now emits \`RESULT: nodesValidated: ...\`, which the validator runtime echoes via \`slog.Info\` unconditionally after each check. Updated the doc section and test expectations to match.

To be precise about what's now covered vs. still tracked in #1951: the coverage figure is visible in live CLI output on every exit path regardless of \`--full\`/redaction; only the redacted signed-bundle artifact still lacks it (structured-field follow-up in #1951 unchanged).

`go build`/`vet`/`test -race` (full `validators/...`) and `golangci-lint` pass clean.

rsd-darshan:
Addressed the 3rd CodeRabbit pass in 74f4538:

1. **No-GPU-nodes exit missing coverage line** — fixed. \`checkNvidiaSMI\` now runs the partition and prints \`enumerationLines\`/\`coverageLine(0)\` before that skip too, so \`nodesValidated: 0/0\` appears there like every other exit — this was the one path my own "every exit path" refactor missed.
2. **\`TestIsSkip\` should be table-driven with a real wrap-chain case** — fixed. Converted to table-driven (repo convention) and added a \`fmt.Errorf("outer: %w", Skip(...))\` case so the test actually proves \`IsSkip\` classifies through a genuine Go error-wrap chain, not just a string-concatenated lookalike.
3. **Route output through a configured writer** — declined again, same reasoning as the prior pass: every check in this package writes evidence via direct \`fmt.Printf\`/\`fmt.Println\`, and introducing a writer abstraction for just this one check would be inconsistent with the rest of \`validators/deployment\`.

\`go build\`/\`vet\`/\`test -race\` (full \`validators/...\`) and \`golangci-lint\` all pass clean.

rsd-darshan:
Addressed the 4th CodeRabbit pass in d2aedc7:

1. **Empty-discovery skip path missing coverage line** — already fixed in 74f4538 (pushed just before this pass ran, so the review caught the pre-fix state).
2. **MD038 markdownlint: spaces inside \`RESULT: \` code spans** — fixed, dropped the trailing space in both isolated-prefix code spans (lines 429/437). The code span containing the full example line (\`RESULT: nodesValidated: ...\`) was already fine since the space there isn't adjacent to a backtick.

yuanchen8911 (commented):
Thanks for the PR! 

The implementation satisfies the agreed scope of #1668 — cordoned nodes are disclosed on every exit path, the `RESULT:` convention is used correctly, and the helper refactor preserves existing callers' behavior. One documentation-contract mismatch and two cosmetic nits:

**Documented coverage numerator doesn't match the implementation.** The new rule in `docs/contributor/validator.md` documents the coverage line as `RESULT: nodesValidated: <schedulable>/<total>`, but `checkNvidiaSMI` passes the *validated* count: `0` on the busy-skip path ([nvidia_smi.go#L148](https://github.com/NVIDIA/aicr/blob/d2aedc79eb73f93b66f14b12aebab6751deabb3d/validators/deployment/nvidia_smi.go#L148)) even though schedulable nodes exist there, and only the successful-node count on the failure path ([nvidia_smi.go#L179](https://github.com/NVIDIA/aicr/blob/d2aedc79eb73f93b66f14b12aebab6751deabb3d/validators/deployment/nvidia_smi.go#L179)). So `<schedulable>` documents the wrong numerator on exactly the skip and failure paths the rule explicitly covers. The implementation's semantics are the better contract — the test even asserts them deliberately ("validated tracks nodes actually confirmed, not nodes attempted") — so suggest changing the placeholder to `<validated>` and defining that term in the doc.

Nits:

- With zero GPU nodes the enumeration prints `Found 0 GPU node(s), 0 schedulable, 0 cordoned:` — a trailing colon introducing an empty list. Consider suppressing the listing when the total is 0.
- The busy-skip coverage line renders `nodesValidated: 0/2 (0 cordoned, skipped)` when nothing is cordoned — accurate but reads oddly. Consider omitting the parenthetical when the cordoned count is 0.

Rebase too.

yuanchen8911 (changes requested):
Requesting the documentation-contract fix from my earlier review comment: the coverage-line rule in `docs/contributor/validator.md` documents the numerator as `<schedulable>`, but the implementation passes the validated count (0 on the busy-skip path, successful-node count on the failure path) — the doc should say `<validated>` and define it, since the implementation's semantics are the better contract and the tests assert them deliberately. The two output-phrasing nits from the same comment are optional. Implementation itself is solid and satisfies the #1668 scope.

rsd-darshan:
Thanks — fixed in 71014dc:

**Doc-contract mismatch (blocking):** you're right, the doc said `<schedulable>` but the code passes the validated count, which diverges on the busy-skip (0, nothing attempted) and failure (successful-subset) paths — exactly the paths this rule exists for. Rather than change the code's behavior (the mixed-scenario test already asserts the validated-count semantics deliberately, and it's the more useful signal), I fixed the doc: it now says `<validated>` and defines what that means on each exit path.

**Both nits:** also applied — empty cluster now prints `Found 0 GPU node(s).` instead of a header over an empty list, and the coverage line omits `(0 cordoned, skipped)` when nothing is cordoned. Added `TestGpuNodeCoverage_EdgeCasePhrasing` covering both.

**On "Rebase too":** deliberately not doing this one. This branch was created fresh from `upstream/main` per a standing rule for this workflow — no rebasing or updating a branch after creation, always start clean per issue — so I'm keeping it as-is rather than making an exception mid-review. It doesn't block anything mechanically: the PR is `MERGEABLE` with no conflicts against `main`, and `#1918`'s CI fix (the `make test` `/validators` exclusion you flagged earlier) is already on `main`, so it'll be picked up automatically in the merge-ref this PR's CI tests against, without needing a rebase on my end.

`go build`/`vet`/`test -race` (full `validators/...`) and `golangci-lint` all pass clean.

yuanchen8911 (approved):
71014dc verified — the coverage-line rule now documents `<validated>` with a definition that matches the implementation exactly (0 on the all-cordoned and busy-skip paths, successful-node count on the failure path), and both optional phrasing nits are implemented with tests asserting the exact strings. Approving.

One staleness to fix before merge: the PR body's summary bullet still describes the coverage line as `nodesValidated: <schedulable>/<total>` — please update it to `<validated>` so the body matches the doc and implementation.

PR #1943, opened 2026-07-29
fix(api): accept versionless legacy bundle recipes

### LINKED ISSUE #1941: Align /v1/bundle OpenAPI schema with versionless legacy recipe support
## Summary

The OpenAPI contract for `POST /v1/bundle` requires `apiVersion`, even though the server intentionally accepts versionless legacy `RecipeResult` artifacts.

## Current behavior

The runtime decoder accepts a recipe with no `apiVersion`. `RecipeResult.ValidateProfileContract` treats an empty version as legacy when `metadata.selectedProfile` is absent.

The OpenAPI request body instead references `LegacyRecipeResponse`, which inherits the `RecipeResponse` requirement for `apiVersion`. OpenAPI validators and generated clients therefore reject a request the server supports. The deprecated `BundleRequest.recipe` schema has the same mismatch.

This behavior predates PR #1933; the schema on `origin/main` already required `apiVersion`.

## Expected behavior

Model `/v1/bundle` input as a legacy union that accepts:

- `aicr.run/v1alpha2` recipes.
- Versionless legacy recipes.
- No profile-bearing `aicr.run/v1alpha3` recipes.

Reuse `VersionlessLegacyRecipeResponse` or introduce a request-specific legacy union so response schemas can retain their existing requirements.

## Acceptance criteria

- `POST /v1/bundle` accepts versionless legacy recipes in the OpenAPI contract.
- Versioned `aicr.run/v1alpha2` recipes remain accepted.
- Profile-bearing recipes remain rejected on v1.
- `BundleRequest.recipe`, if retained, uses the same contract.
- OpenAPI synchronization tests cover both legacy variants.

yuanchen8911:
## Summary

Align the `/v1/bundle` OpenAPI request contract with the server's supported
legacy `RecipeResult` header shapes. Keep current recipe responses strict while
reusing the same request schema in the deprecated bundle wrapper.

## Motivation / Context

The runtime accepts legacy recipes with missing or empty header fields, but the
published request schema required canonical `apiVersion` and `kind` values.
OpenAPI validators and generated clients therefore rejected supported requests
before they reached the server.

Fixes #1941.

Fixes: #1941
Related: #1933 (overlapping schema and test surface), #1953 (runtime kind
normalization follow-up)

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Implementation Notes

- Split the shared `RecipeResult` shape from wrappers that impose header
  requirements and constraints.
- Keep `RecipeResponse` strict while allowing `/v1/bundle` requests to omit
  `apiVersion` or `kind`, or carry their legacy empty-string values.
- Keep the legacy `kind: Recipe` value valid on the request side. It was the
  value this contract published through v0.18.0, and the handler validates no
  header field, so removing it would make previously conforming clients
  spec-invalid against a request-validating gateway.
- Apply the same request schema to the deprecated `BundleRequest.recipe` field.
- Correct the resource kind to `RecipeResult`, fail closed when shared header
  properties disappear, and document both OpenAPI contract tests.
- Gate runtime acceptance separately from the spec: `TestBundleHandler_LegacyRecipeHeaders`
  posts absent, empty, single-field, and legacy `kind: Recipe` bodies to the
  handler. The spec gate alone only checks the spec against itself.
- Assert the `/v1/recipe` response `$ref` sites, and match `allOf` branches by
  content rather than position so a reordered — or grafted-on — branch is caught.

## Testing

```bash
# Full package. Run outside the agent sandbox: TestGracefulShutdown binds a
# port, which the sandbox denies with "operation not permitted".
GOFLAGS="-mod=vendor" go test -race -count=1 ./pkg/server
GOFLAGS="-mod=vendor" go test -race -count=1 ./pkg/server -run '^(TestOpenAPIV1BundleRecipeContract|TestOpenAPIEnumsMatchGoTypes|TestBundleHandler_LegacyRecipeHeaders|TestBundleHandler_EmptyComponentRefs|TestBundleHandler_IncoherentComponentRef)$'
env -u GITLAB_TOKEN make test-coverage
golangci-lint run -c .golangci.yaml ./pkg/server/...
yamllint -c .yamllint.yaml api/aicr/v1/server.yaml
./tools/check-docs-mdx
git diff origin/main...HEAD --check
```

**Control checks.** Every new gate was verified to fail when the defect it
guards against is introduced, then reverted:

- Adding a `kind` check to `HandleBundles` fails the legacy cases in
  `TestBundleHandler_LegacyRecipeHeaders` while the canonical control passes.
- Adding a paired-header check (`(kind == "") != (apiVersion == "")`) fails
  exactly the five single-field cases and none of the paired ones — the shape
  the earlier version of this test would have missed entirely.
- Removing `Recipe` from the `BundleRecipeRequest` kind enum fails
  `TestOpenAPIV1BundleRecipeContract`.
- Grafting a third strict `$ref` onto the `allOf` fails the contract gate. Before
  the fail-closed rewrite of the helper it passed silently.

Project coverage passed at 80.7% against the 80% threshold; `pkg/server` is
unchanged at 78.7% on both this branch and `origin/main`. Scoped lint reports
zero issues.

Full `make qualify` was not rerun for the focused test and doc passes: the
change touches only the OpenAPI contract, docs, and tests, with no production Go
code.

## Risk Assessment

**Rollout notes:** No migration is required. Current versioned requests remain
valid, and requests written against the v0.18.0 contract stay valid; the schema
now represents the legacy input already accepted at runtime.

The request schema is deliberately looser than the response schema because
`bundleHandler.HandleBundles` performs no header validation at all. Tightening
the runtime to enforce the documented contract is a separate question and is
not in scope here.

## Checklist

### DIFF
--- a/api/aicr/v1/server.yaml
+++ b/api/aicr/v1/server.yaml
@@ -193,7 +193,7 @@ paths:
                   summary: Basic recipe request
                   value:
                     apiVersion: aicr.run/v1alpha2
-                    kind: Recipe
+                    kind: RecipeResult
                     metadata:
                       version: v1.0.0
                       created: "2025-01-15T10:30:00Z"
@@ -861,6 +861,11 @@ paths:
         This design enables a simple workflow: pipe the output from GET /v1/recipe directly
         to POST /v1/bundle.
 
+        Current recipes carry apiVersion aicr.run/v1alpha2 and kind RecipeResult.
+        For backward compatibility, legacy artifacts may omit either header field,
+        carry it as an empty string after a decode/remarshal round trip, or carry
+        the kind Recipe published by this contract through v0.18.0.
+
         The response is a binary zip file containing all generated bundle artifacts
         organized by bundler type (e.g., gpu-operator/, network-operator/).
       parameters:
@@ -1113,13 +1118,13 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/RecipeResponse"
+              $ref: "#/components/schemas/BundleRecipeRequest"
             examples:
               simpleRecipe:
                 summary: Simple recipe with GPU operator
                 value:
                   apiVersion: aicr.run/v1alpha2
-                  kind: Recipe
+                  kind: RecipeResult
                   componentRefs:
                     - name: gpu-operator
                       type: Helm
@@ -1130,7 +1135,7 @@ paths:
                 summary: Full recipe with metadata
                 value:
                   apiVersion: aicr.run/v1alpha2
-                  kind: Recipe
+                  kind: RecipeResult
                   metadata:
                     version: v1.0.0
                     created: "2025-01-15T10:30:00Z"
@@ -1562,21 +1567,22 @@ components:
             $ref: "#/components/schemas/Subtype"
           description: List of configuration subtypes within this measurement
 
-    RecipeResponse:
+    RecipeResponseBase:
       type: object
-      description: Complete recipe response with metadata and component references
-      required: [apiVersion, kind]
+      description: >
+        Reusable RecipeResult shape. Response and bundle-request wrappers apply
+        their respective apiVersion requirements.
       properties:
         apiVersion:
           type: string
-          description: API version of the recipe format
-          enum: [aicr.run/v1alpha2]
-          example: aicr.run/v1alpha2
+          description: >
+            API version of the recipe format. Response and bundle-request
+            wrappers constrain the supported value.
         kind:
           type: string
-          description: Resource type
-          enum: [Recipe]
-          example: Recipe
+          description: >
+            Resource type. Response and bundle-request wrappers constrain the
+            supported value.
         metadata:
           type: object
           description: Recipe metadata
@@ -1761,6 +1767,61 @@ components:
           description: Deployment constraints (driver versions, etc.)
           additionalProperties: true
 
+    RecipeResponse:
+      description: Complete recipe response with metadata and component references
+      allOf:
+        - $ref: "#/components/schemas/RecipeResponseBase"
+        - type: object
+          required: [apiVersion, kind]
+          properties:
+            apiVersion:
+              type: string
+              enum: [aicr.run/v1alpha2]
+              example: aicr.run/v1alpha2
+            kind:
+              type: string
+              enum: [RecipeResult]
+              example: RecipeResult
+
+    BundleRecipeRequest:
+      description: >
+        RecipeResult accepted by POST /v1/bundle. Current aicr.run/v1alpha2
+        artifacts and legacy artifacts with absent or empty header fields are
+        accepted. The kind value Recipe is also accepted: it was the value this
+        contract published through v0.18.0, and the handler performs no header
+        validation, so clients written against that spec must keep working.
+
+        A legacy kind is echoed into the generated bundle's recipe.yaml rather
+        than normalized, and the CLI file loader accepts an absent or empty kind
+        but rejects Recipe. A bundle generated from a kind Recipe body is
+        therefore not reloadable via aicr bundle -r or aicr validate -r. Send
+        kind RecipeResult to keep the emitted artifact round-trippable.
+
+        apiVersion deliberately has no equivalent legacy window. The artifact
+        group/version bump is a hard break with no transition period (see
+        docs/design/011-artifact-apiversion-policy.md), so an artifact stamped
+        with a prior group/version should be regenerated rather than sent. That
+        is a client-contract rule rather than an enforced one: this endpoint
+        validates no header field, so a prior apiVersion is ignored rather than
+        rejected. The CLI file-load path does enforce it.
+      allOf:
+        - $ref: "#/components/schemas/RecipeResponseBase"
+        - type: object
+          properties:
+            apiVersion:
+              type: string
+              description: >
+                Current API version, or empty after a legacy struct round trip.
+                Prior group/versions are not accepted — see the schema
+                description for the hard-break policy.
+              enum: ["", aicr.run/v1alpha2]
+            kind:
+              type: string
+              description: >
+                RecipeResult, the legacy Recipe value published through
+                v0.18.0, or empty for a legacy artifact
+              enum: ["", Recipe, RecipeResult]
+
     QueryRequest:
       type: object
       description: Query request with criteria and dot-path selector
@@ -1782,12 +1843,12 @@ components:
       description: Request body for bundle generation
       deprecated: true
       x-deprecation-reason: >
-        Use RecipeResponse directly as the request body and specify bundlers
+        Use a RecipeResult directly as the request body and specify bundlers
         via the "bundlers" query parameter (comma-delimited).
       required: [recipe]
       properties:
         recipe:
-          $ref: "#/components/schemas/RecipeResponse"
+          $ref: "#/components/schemas/BundleRecipeRequest"
           description: The recipe to generate bundles from
         bundlers:
           type: array
--- a/docs/user/api-reference.md
+++ b/docs/user/api-reference.md
@@ -416,7 +416,34 @@ Generate deployment bundles from a recipe.
 
 **Request Body:**
 
-The request body is the recipe (RecipeResult) directly. No wrapper object needed.
+The request body is the recipe (`RecipeResult`) directly. No wrapper object is
+needed. Current artifacts carry `apiVersion: aicr.run/v1alpha2` and
+`kind: RecipeResult`, and new clients should send that versioned form.
+
+For backward compatibility, the endpoint also accepts:
+
+- Legacy artifacts that omit `apiVersion` or `kind`, or carry them as empty
+  strings after a decode/remarshal round trip.
+- The `kind: Recipe` value this contract published through v0.18.0.
+
+The handler does not validate these header fields, so all three shapes reach the
+bundler identically.
+
+`apiVersion` has no equivalent legacy window on purpose. An artifact
+group/version bump is a hard break with no transition period, so a recipe
+stamped with a prior group/version should be regenerated rather than sent.
+
+That is a client-contract rule, not an enforced one. As above, this endpoint
+validates no header field: a prior `apiVersion` is ignored rather than rejected,
+so such a request still succeeds. The CLI file-load path does enforce it and
+rejects an unsupported `apiVersion` outright.
+
+**Round-trip caveat for `kind: Recipe`.** The header is copied into the
+generated bundle's `recipe.yaml` rather than normalized. The CLI file loader
+tolerates an absent or empty `kind` but rejects `Recipe`, so a bundle generated
+from a `kind: Recipe` body cannot be fed back through `aicr bundle -r` or
+`aicr validate -r`. Send `kind: RecipeResult` if you need the emitted artifact
+to remain reloadable.
 
 #### Components
 
--- a/pkg/server/bundle_handler_test.go
+++ b/pkg/server/bundle_handler_test.go
@@ -30,6 +30,8 @@ import (
 	"github.com/NVIDIA/aicr/pkg/bundler/result"
 	aicr "github.com/NVIDIA/aicr/pkg/client/v1"
 	aicrerrors "github.com/NVIDIA/aicr/pkg/errors"
+	"github.com/NVIDIA/aicr/pkg/header"
+	"github.com/NVIDIA/aicr/pkg/recipe"
 )
 
 // fixtureBundleAttester returns fixed, non-nil bundle JSON from Attest so the
@@ -341,7 +343,7 @@ func TestBundleHandler_EmptyComponentRefs(t *testing.T) {
 	t.Parallel()
 	h := newTestBundleHandler(t)
 
-	body := `{"apiVersion": "aicr.run/v1alpha2", "kind": "Recipe", "componentRefs": []}`
+	body := `{"apiVersion": "aicr.run/v1alpha2", "kind": "RecipeResult", "componentRefs": []}`
 	req := httptest.NewRequest(http.MethodPost, "/v1/bundle", strings.NewReader(body))
 	req.Header.Set("Content-Type", "application/json")
 	w := httptest.NewRecorder()
@@ -359,7 +361,7 @@ func TestBundleHandler_IncoherentComponentRef(t *testing.T) {
 	t.Parallel()
 	h := newTestBundleHandler(t)
 
-	body := `{"apiVersion": "aicr.run/v1alpha2", "kind": "Recipe", "componentRefs": [` +
+	body := `{"apiVersion": "aicr.run/v1alpha2", "kind": "RecipeResult", "componentRefs": [` +
 		`{"name": "gpu-operator", "type": "Helm", "version": "v1", "tag": "v2"}]}`
 	req := httptest.NewRequest(http.MethodPost, "/v1/bundle", strings.NewReader(body))
 	req.Header.Set("Content-Type", "application/json")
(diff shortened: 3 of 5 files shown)


pkg/server/bundle_handler_test.go
```diff
@@ -341,7 +341,7 @@ func TestBundleHandler_EmptyComponentRefs(t *testing.T) {
 	t.Parallel()
 	h := newTestBundleHandler(t)
 
-	body := `{"apiVersion": "aicr.run/v1alpha2", "kind": "Recipe", "componentRefs": []}`
+	body := `{"apiVersion": "aicr.run/v1alpha2", "kind": "RecipeResult", "componentRefs": []}`
```
njhensley, line 346: **🟡 Minor — Versionless/empty-header acceptance is untested at the HTTP handler boundary**

The behavior this PR advertises — /v1/bundle accepting recipes with absent or empty apiVersion/kind — has no test that exercises the handler. `TestOpenAPIV1BundleRecipeContract` only inspects the YAML spec text (os.ReadFile+yaml.Unmarshal) and never starts a handler. The two edited handler tests both send fully-versioned bodies and assert 400 on unrelated paths. `loader_test.go`'s empty-kind coverage does not apply — the HTTP path decodes straight into `recipe.RecipeResult` and bypasses `loader.go` (decode → AdoptRecipe → PrepareAndValidate, none of which inspect the header).

**Blast radius:** Low today; a future decode/validation refactor that adds header gating would silently break versionless acceptance with no failing test.

**Fix:** Add a table case POSTing `{"componentRefs":[...]}` and empty-string-header variants, asserting non-4xx-on-header-grounds. Minor rather than Major because there is no header-gating logic on this path to regress against today.

api/aicr/v1/server.yaml
```diff
@@ -1761,6 +1766,41 @@ components:
...
+        accepted.
+      allOf:
+        - $ref: "#/components/schemas/RecipeResponseBase"
+        - type: object
+          properties:
+            apiVersion:
+              type: string
+              description: >
+                Current API version, or empty after a legacy struct round trip
+              enum: ["", aicr.run/v1alpha2]
```
njhensley, line 1817: **🟡 Minor — BundleRecipeRequest enum is stricter than the handler (permissive-direction spec/runtime drift)**

`BundleRecipeRequest` pins apiVersion to ["", aicr.run/v1alpha2] and kind to ["", RecipeResult], but the handler validates neither — `PrepareAndValidate` checks only ref coherence, and the `IsSupportedAPIVersion` gate lives on the file-load path (loader.go:102), which this endpoint bypasses. So the running server accepts any apiVersion/kind string (e.g. aicr.run/v1alpha1, Recipe), silently ignoring it. This is exactly the drift direction the parity test's own docstring warns about ("clients conforming to the spec will reject inputs the server actually accepts").

**Blast radius:** Harmless in practice and arguably a defensible allowlist-of-intent, but undocumented.

**Fix:** Either add a one-line note (schema description or api-server.md) that the enum is advisory client/codegen guidance rather than server-enforced, or fail closed on unknown non-empty apiVersions via header.IsSupportedAPIVersion so spec and handler converge. A doc note is sufficient.

pkg/server/openapi_sync_test.go
```diff
@@ -83,6 +83,122 @@ func TestOpenAPIEnumsMatchGoTypes(t *testing.T) {
...
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			if len(tt.schema.AllOf) != 2 {
+				t.Fatalf("allOf has %d entries, want 2", len(tt.schema.AllOf))
+			}
+			const want = "#/components/schemas/RecipeResponseBase"
+			if got := tt.schema.AllOf[0].Ref; got != want {
+				t.Errorf("base $ref = %q, want %q", got, want)
+			}
+			closure := tt.schema.AllOf[1]
```
njhensley: **🔵 Nitpick — Contract test indexes allOf positionally; allOf is order-independent**

The test asserts AllOf[0].Ref == RecipeResponseBase and reads required/enums from AllOf[1]. OpenAPI allOf is semantically unordered, so a spec author who swaps the two entries (a still-valid, equivalent contract) would trip a spurious failure. Downgraded from Minor to Nitpick: the failure is loud (base $ref + enum mismatch → CI red), never a silent false-pass, and the same author co-owns both entries.

**Blast radius:** None — fail-safe brittleness, not a correctness gap.

**Fix:** Optional: locate the base entry by Ref == RecipeResponseBase and the closure by Ref == "" instead of indexing [0]/[1].

pkg/server/openapi_sync_test.go
```diff
@@ -198,3 +314,11 @@ func equalStrings(a, b []string) bool {
 	}
 	return true
 }
+
+func equalStringSets(a, b []string) bool {
```
njhensley: **🔵 Nitpick — equalStringSets is a sorted-slice compare, not true set semantics**

The helper sorts copies then defers to `equalStrings` (length + element compare), so ["a","a"] vs ["a"] returns false — duplicates are not collapsed, and the name overstates it.

**Blast radius:** Harmless — every enum input is a unique literal.

**Fix:** Optional: rename to equalStringsUnordered, or dedupe both sides if true set semantics are intended.

docs/user/api-reference.md
```diff
@@ -416,7 +416,11 @@ Generate deployment bundles from a recipe.
 
 **Request Body:**
 
-The request body is the recipe (RecipeResult) directly. No wrapper object needed.
+The request body is the recipe (`RecipeResult`) directly. No wrapper object is
+needed. Current artifacts carry `apiVersion: aicr.run/v1alpha2` and `kind:
```
njhensley: **🔵 Nitpick — Inline code span split across a source line break (renders fine)**

The span `kind: RecipeResult` opens at the end of one line and closes on the next; backticks are balanced and CommonMark folds the intra-span line ending to a single space, so it renders correctly on GitHub and Fern.

**Blast radius:** None — purely cosmetic.

**Fix:** Optional: reflow so the span sits on one source line for readability.

api/aicr/v1/server.yaml
```diff
@@ -1761,6 +1766,41 @@ components:
           description: Deployment constraints (driver versions, etc.)
           additionalProperties: true
 
+    RecipeResponse:
```
njhensley, line 1770: **🔵 Nitpick — RecipeResponse flat-object → allOf composition may churn external codegen**

The RecipeResponse type name is preserved and both in-repo $refs (spec L190/L420) stay intact, and there is no generated client in this repo, so nothing breaks here. Strict OpenAPI generators may emit a differently-shaped composed type for external SDK consumers who regenerate.

**Blast radius:** Speculative, out of repo scope.

**Fix:** Worth a line in the API changelog if published SDKs exist; otherwise no action.

njhensley (dismissed):
## 🧭 Multi-Persona Review — Approve with comments

**Method:** four parallel persona reviewers (API-contract, Correctness, Test-coverage, Docs) followed by an adversarial senior meta-reviewer that independently re-derived each finding from the resolved code at head `14bf3d0`.

### Assessment

A clean, well-scoped **contract + docs + test-only** change — no production Go touched. It relaxes the `/v1/bundle` OpenAPI request schema (`RecipeResponse` → shared `RecipeResponseBase` + strict `RecipeResponse` / permissive `BundleRecipeRequest` `allOf` wrappers) to match what the server already accepts, and adds a `TestOpenAPIV1BundleRecipeContract` gate.

The central premise is **verified against the code**: the bundle handler decodes straight into `recipe.RecipeResult` (`decode → AdoptRecipe → PrepareAndValidate`) and never inspects `apiVersion`/`kind`, so versionless/empty-header legacy recipes are genuinely accepted. The PR also *silently fixes a pre-existing contract bug* — the old response enum was `[Recipe]` while the server actually stamps `RecipeResult`, so conforming clients would have rejected the real `GET /v1/recipe` body.

No blockers. The one substantive theme (the two 🟡 inline comments) is that the newly-advertised versionless-acceptance contract is real but **un-gated and untested at the HTTP boundary**, and the request enum is actually *stricter* than the fully-lenient handler — worth one behavioral test and/or a doc note, not a hold.

- **Versionless-not-required IS gated** — the test asserts the bundle-request closure has no `required` and `RecipeResponseBase.Required` is empty, so re-adding a header requirement fails CI.
- **Constant-vs-literal enum comparison is a protective gate** — a `GroupVersion` bump that misses the YAML literal fails the test.
- **`Recipe`→`RecipeResult` examples are a correct fix**, not churn — the old response enum `[Recipe]` was wrong vs the server's emitted `RecipeResult`.
- **Deprecated `BundleRequest` wrapper has no live Go decode path** — relaxing its `$ref` is documentation-only.
- **Docs are accurate, complete, and anchor-clean** — both user + contributor pages updated, no stale `kind: Recipe` examples remain.

### Summary

|:---:|:---:|:---:|:---:|
| 0 | 0 | 2 | 4 |

_Reviewed with a multi-persona panel + adversarial meta-reviewer. All six findings were independently reproduced against the code at head `14bf3d0`._

pkg/server/bundle_handler_test.go
yuanchen8911: Fixed in d46eec6bc.

Added `TestBundleHandler_LegacyRecipeHeaders` in `pkg/server/bundle_handler_test.go`: a table-driven test that posts a real resolved recipe body to `HandleBundles` in four shapes — canonical, absent `apiVersion`/`kind`, empty-string headers, and `kind: Recipe` — and asserts 200 for each.

The canonical case is included as a control, so a passing run means the body was genuinely accepted rather than the assertion being vacuous. The test also fails closed if the fixture ever stops being the canonical shape, since a "legacy" case has to differ from the baseline to mean anything.

Verified the test actually does its job: temporarily adding a `kind` check to `HandleBundles` makes the three legacy cases fail with 400 while the control still passes.

api/aicr/v1/server.yaml
yuanchen8911: Partly fixed, partly intentional.

`kind` — the enum now includes the legacy value: `["", Recipe, RecipeResult]`. That was the value this contract published from v0.14.0 through v0.18.0 (verified against each tag), and since the handler validates nothing, a client written against the released spec would otherwise have become spec-invalid at a request-validating gateway. Fixed in d46eec6bc.

`apiVersion` — stays `["", aicr.run/v1alpha2]` deliberately. `pkg/header` documents the group/version bump as a hard break with no transition window, and `pkg/recipe/loader.go` rejects a prior group/version on the file-load path. Publishing `aicr.nvidia.com/v1alpha1` as valid would contradict that policy. The schema description and `docs/user/api-reference.md` now state the asymmetry and the reason for it, which they previously did not.

The remaining direction of the drift — the handler validating no header field at all — is unchanged here. That is better closed by tightening the runtime than by loosening the spec further, and it is a behavioral change rather than a contract fix, so it belongs in its own PR.

pkg/server/openapi_sync_test.go
yuanchen8911: Fixed in d46eec6bc.

Replaced the positional indexing with an `allOfConstraint` helper that identifies the base `$ref` and the inline constraint object by content rather than position, so swapping the two `allOf` entries — an equivalent contract — no longer trips the gate.

It still fails closed: a missing base `$ref`, or any number of inline constraint objects other than exactly one, is an error rather than a silently skipped assertion.

pkg/server/openapi_sync_test.go
yuanchen8911: Fixed in d46eec6bc. Renamed to `equalStringsUnordered`, which describes what it actually does.

Kept the duplicate-sensitive behavior on purpose and documented why on the helper: a repeated enum member is itself a spec defect worth failing on, not something to collapse silently.

docs/user/api-reference.md
yuanchen8911: Already fixed — the span was reflowed onto a single source line when that paragraph was rewritten, so this thread is showing against an outdated diff.

api/aicr/v1/server.yaml
yuanchen8911: Acknowledged — accepting this risk rather than changing the composition.

Confirmed there is no in-repo generator, CI job, Fern config, or e2e fixture consuming `api/aicr/v1/server.yaml`; the only consumer is `pkg/server/openapi_sync_test.go`. So nothing in this repo changes shape.

External SDK users who regenerate may get a composed type for `RecipeResponse`. The type name is preserved and both response `$ref` sites still point at it, so the wire contract is unchanged — only the generated Go/Python/TS type shape could differ. There is no way to verify generator output here without running codegen, so flagging it rather than claiming it is a non-issue.

api/aicr/v1/server.yaml
yuanchen8911: Filed the runtime-side follow-up as #1953 — normalize the legacy `kind` on ingest so the emitted `bundle/recipe.yaml` stays reloadable, keeping `apiVersion` strict per the hard-break policy.

docs/user/api-reference.md
```diff
@@ -416,7 +416,29 @@ Generate deployment bundles from a recipe.
...
+- Legacy artifacts that omit `apiVersion` or `kind`, or carry them as empty
+  strings after a decode/remarshal round trip.
+- The `kind: Recipe` value this contract published through v0.18.0.
+
+The handler does not validate these header fields, so all three shapes reach the
+bundler identically.
+
+`apiVersion` has no equivalent legacy window on purpose. An artifact
+group/version bump is a hard break with no transition period, so a recipe
+stamped with a prior group/version must be regenerated rather than accepted.
```
njhensley: **🔵 Nitpick — Docs say a prior apiVersion "must be regenerated rather than accepted," but the handler accepts it**

The prose states an artifact stamped with a prior group/version "must be regenerated rather than accepted." But the bundle handler validates no header field (there is no IsSupportedAPIVersion check on the /v1/bundle path — that gate lives only on the CLI file-load path in loader.go), so a POST with e.g. apiVersion: aicr.run/v1alpha1 still returns 200, the field silently ignored. The new TestBundleHandler_LegacyRecipeHeaders covers only empty/absent apiVersion, not a bogus non-empty one. The word "accepted" reads as handler behavior when it actually describes the client-contract policy.

**Blast radius:** None — fail-safe, and self-disclosed five lines earlier ("The handler does not validate these header fields"). Wording-precision only.

**Fix:** Optional: soften to "should be regenerated rather than sent," or add a negative case to TestBundleHandler_LegacyRecipeHeaders documenting that the server is lenient here too (a prior apiVersion is ignored, not rejected).

njhensley (dismissed):
## 🔁 Re-review — approve stands (head `06d223d`)

Re-reviewed after the force-push. The update **directly and thoroughly addresses every finding** from my prior pass, and strengthens the PR beyond it. Verified each fix against the code at `06d223d`.

### Prior findings — all resolved

- **🟡 Missing HTTP-boundary test** → new `TestBundleHandler_LegacyRecipeHeaders`: a 10-case table (control + fail-closed fixture guard + absent/empty single & both fields + legacy `Recipe` kind) that POSTs to `HandleBundles` and asserts **200**. Exactly the behavioral gate that was missing.
- **🟡 Request enum stricter than handler** → `kind` enum widened to `["", Recipe, RecipeResult]` to match the lenient handler; schema + `api-reference.md` + `api-server.md` now document the back-compat window, the `kind: Recipe` **round-trip caveat** (verified accurate — the input recipe is copied into the bundle's `recipe.yaml` and the CLI loader rejects `Recipe`), and the deliberate `apiVersion` hard-break policy. Bonus: a new response-site guard asserts `/v1/recipe` 200s still point at the **strict** `RecipeResponse`, blocking a future silent relax to the permissive enum.
- **🔵 `allOf` positional indexing** → `allOfConstraint()` now matches branches by content (`baseRef`), not position.
- **🔵 `equalStringSets` naming** → renamed `equalStringsUnordered`.
- **🔵 Split code span** → reflowed onto one source line.

### One optional wording nit

See the single inline comment — the docs describe a prior `apiVersion` as "must be regenerated rather than accepted," which reads as handler enforcement, but the handler is actually lenient (it validates nothing and would 200). Fail-safe and already self-disclosed; non-blocking.

### Summary

|:---:|:---:|:---:|:---:|
| 0 | 0 | 0 | 1 (optional) |

Approval stands. Nice, comprehensive response to review.

docs/user/api-reference.md
yuanchen8911: Agreed — that wording was mine and it does read as handler behavior when it states a client-contract rule. Fixed in 3f0fab1bc, in both the docs prose and the `BundleRecipeRequest` schema description, which carried the same phrasing.

Now: *"should be regenerated rather than sent"*, followed by an explicit statement that this is a client-contract rule rather than an enforced one — a prior `apiVersion` is ignored, not rejected, and the CLI file-load path is where it is actually enforced.

On the alternative you offered: I did not add a negative case pinning the lenient `apiVersion` behavior. #1953 explicitly weighs making the handler *reject* an unsupported non-empty `apiVersion` the way the loader does, so a test asserting 200 for a bogus value would encode behavior we may deliberately change, and would then have to be rewritten by that fix. Happy to add it as a characterization test if you would rather have the current behavior pinned in the meantime — just say so.

njhensley (approved):
## 🔁 Re-review — approve, clean pass (head `3f0fab1`)

Re-reviewed after the latest push. The single wording nit from my prior pass is resolved and there are no remaining findings.

The docs and schema no longer imply the `/v1/bundle` handler rejects a prior `apiVersion`. Both now state the enum is a **client-contract rule, not enforced** — the endpoint validates no header field (a prior `apiVersion` is ignored, request still 200s), while the CLI file-load path (`loader.go`, `IsSupportedAPIVersion`) does reject an unsupported `apiVersion` outright. Both claims verified against the code.

### Summary of the review arc
- **Original (`14bf3d0`):** 2 🟡 + 4 🔵 — missing HTTP-boundary test, request enum stricter than handler, `allOf` positional indexing, helper naming, split code span, codegen note.
- **Re-review 1 (`06d223d`):** all six resolved (new `TestBundleHandler_LegacyRecipeHeaders`, widened `kind` enum + round-trip caveat docs, content-based `allOf` matching, rename, reflow, plus a response-site strictness guard); one new 🔵 wording nit.
- **Re-review 2 (`3f0fab1`):** wording nit resolved.

|:---:|:---:|:---:|:---:|
| 0 | 0 | 0 | 0 |

A well-run PR — thorough, responsive to review, and cleanly documented. LGTM.

PR #1944, opened 2026-07-29
fix(bundler): create argocd-helm static/ only when populated

### LINKED ISSUE #1942: argocd-helm deployer fails for OCP recipes on empty static/ directory
## Description

`aicr bundle --deployer argocd-helm` fails for every OCP recipe with:

```
[INVALID_REQUEST] bundle contains an unexpected directory: "static"
```

The `argocd-helm` generator always creates a `static/` directory, but for OCP recipes nothing is ever written into it. The bundle inventory validator then rejects the empty directory because it is not part of the expected tree, and generation aborts.

This is not specific to any one OCP overlay or to a particular flag combination. It reproduces on current `main`.

## Reproduction

```bash
aicr recipe --service ocp -o ocp.yaml

aicr bundle -r ocp.yaml -o ./out \
  --deployer argocd-helm \
  --accelerated-node-selector nvidia.com/gpu.present=true \
  --accelerated-node-toleration nvidia.com/gpu:NoSchedule \
  --system-node-selector node-role=system
```

Result: exit code 2, with the error above. The partially written output directory contains an empty `static/`:

```
001-nfd-ocp-olm  002-nfd-ocp  003-gpu-operator-ocp-olm  004-gpu-operator-ocp
005-network-operator-ocp-olm  006-network-operator-ocp
Chart.yaml  README.md  static  templates  values.schema.json  values.yaml

$ find out/static -type f | wc -l
0
```

## Scope

All three OCP overlays are affected: `ocp`, `ocp-training`, `ocp-inference`.

Only the `argocd-helm` deployer is affected. The same recipe bundles cleanly everywhere else:

| Deployer | Result |
|---|---|
| `helm` | works (38 files) |
| `argocd` | works (31 files) |
| `argocd-helm` | **fails** |
| `flux` | works (26 files) |
| `helmfile` | works (41 files) |

Conversely, `argocd-helm` works for the other 97 overlays in the embedded catalog. The failure is specific to the OCP × `argocd-helm` intersection.

## Root cause

- `pkg/bundler/deployer/argocdhelm/argocdhelm.go` — `writeStaticValuesAndBuildStubs` calls `os.MkdirAll` on `<output>/static` unconditionally, then writes a values file only for components it classifies as Helm charts.
- `pkg/bundler/checksum/inventory.go` — `validateExactTree` walks the finished tree and fails on any directory not present in `expectedDirectories`. An empty `static/` contributes no files, so it never enters that set.

For OCP recipes the per-component values land in the numbered subchart directories instead, leaving `static/` empty and tripping the validator.

## Suggested fix

Options, roughly in order of least surprise:

- Create `static/` lazily, only when the first component values file is about to be written.
- Remove the directory at the end of generation when it is still empty.
- Register `static/` as an expected directory so an empty one is tolerated.

The first two keep the emitted tree free of empty directories, which matches the validator's existing exact-tree contract.

## Notes

Found while sweeping all five deployers across the full embedded catalog. Pre-existing on `main` — reproduced at `d4f7bef46`, unrelated to any in-flight profile work.

yuanchen8911:
## Summary

Create the argocd-helm bundle's `static/` directory lazily, so a bundle whose components contribute no static values omits it entirely instead of emitting an empty directory that fails bundle inventory validation. Also remove an empty `static/` left behind by an earlier failed run, so retrying into the same output directory succeeds.

## Motivation / Context

`aicr bundle --deployer argocd-helm` failed for every OCP recipe with:

```
[INVALID_REQUEST] bundle contains an unexpected directory: "static"
```

`writeStaticValuesAndBuildStubs` created `<output>/static` unconditionally, then wrote a values file into it only for components resolving to a remote Helm chart:

```go
isHelmChart := ref.Type != recipe.ComponentTypeKustomize && ref.Source != ""
```

The OCP overlays declare every component as `type: Helm` built from `manifestFiles` with **no** upstream `source`, so `Source == ""` skips all of them and nothing is ever written. `validateExactTree` in `pkg/bundler/checksum/inventory.go` then walks the finished tree and rejects any directory absent from `expectedDirectories` — an empty directory contributes no files, so it never enters that set.

Fixes: #1942
Related: #1947

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: N/A

## Implementation Notes

The `MkdirAll` moves behind an `ensureStaticDir` closure invoked immediately before the first component values file is written. Bundles that do write static values are unaffected — the directory is created at the same point in the flow, with the same permissions, before the same write.

Lazy creation alone does not help a user who already hit the bug. Generation writes directly into `--output` and does not clear it between runs, so the failed run's empty `static/` is still on disk and a plain retry reproduces the same error. When nothing is written, the directory is therefore removed.

The path is guarded in one place, `checkStaticPath`, shared by the create and remove routes. `os.Remove` unlinks regular files as readily as it removes empty directories, so removal must not run blind; and `MkdirAll` reports a non-directory only as a generic `ENOTDIR`, so without a shared guard the same user error surfaced as `INVALID_REQUEST` for an all-local recipe but `INTERNAL` for any recipe carrying an upstream chart — the outcome decided by recipe shape rather than by the mistake. `Lstat` rather than `Stat` also rejects a symlink planted after `checksum.ValidateOutputRoot` ran, which permits pre-existing regular files, and a symlink gets its own message rather than being reported as "not a directory".

Emptiness is checked explicitly rather than inferred from an `os.Remove` errno, which keeps the intent readable and avoids platform-specific error values. A populated `static/` is kept and reported by `validateExactTree` as unexpected output — though only when checksums are enabled, since that validator runs inside the `IncludeChecksums` path. With checksums off a stale `static/` survives into the bundle, unchanged from before this fix: a reused `--output` was never cleared, so stale files of any kind already persisted.

That new `ErrCodeInvalidRequest` is why `writeValuesFiles` switches to `errors.PropagateOrWrap`: re-coding it as `INTERNAL` would present a bad `--output` path as a server fault, and over the API a 5xx withholds the cause from the response.

Of the three options in the issue, lazy creation was chosen over pruning after the fact or registering `static/` as always-expected. It keeps the emitted tree free of empty directories, preserving the inventory validator's exact-tree contract rather than widening it.

Only the `argocd-helm` deployer is touched. `helm`, `argocd`, and `helmfile` already bundled OCP recipes cleanly and are unchanged.

`flux` has the same defect on its OCI output path — `sources/` is created unconditionally and stays empty for an OCP recipe — filed as #1947 rather than fixed here. The catalog sweep above did not surface it because it bundles to a local directory, where `gitSources` is populated. Widening this PR to a second deployer would be scope creep, and the two fixes should be reviewable independently.

## Testing

```bash
GOFLAGS="-mod=vendor" go test -race ./pkg/bundler/deployer/argocdhelm/ -count=1
GOFLAGS="-mod=vendor" go test ./pkg/bundler/... -count=1
golangci-lint run -c .golangci.yaml ./pkg/bundler/deployer/argocdhelm/...
./tools/check-docs-mdx
```

The argocd-helm package passes with the race detector, affected-package lint reports 0 issues, and the MDX doc check passes.

**Regression sweep.** Binaries built from this branch and from the merge-base were run against all 100 embedded catalog overlays on the `argocd-helm` deployer, comparing whole output trees:

| Result | Count |
|---|---|
| Byte-identical to baseline | 97 |
| Newly succeeding (`ocp`, `ocp-training`, `ocp-inference`) | 3 |
| Differing output | 0 |

The three OCP overlays go from exit 2 to exit 0, producing 34 files each with `static/` correctly absent. The other 97 are unchanged, including their `static/` contents.

**Regression tests.** `TestGenerate_StaticDirCreatedOnlyWhenPopulated` is table-driven with three cases and sets `IncludeChecksums: true`, so it exercises `finalizeOutput` → `checksum.WriteChecksums` → `validateExactTree` — the check that actually rejected the empty directory — rather than asserting only the proxy that `static/` is absent. Verified against three source states:

| Source state | all-local | stale `static/` | remote chart (control) |
|---|---|---|---|
| `origin/main` | fail | fail | pass |
| lazy creation only | pass | fail | pass |
| this branch | pass | pass | pass |

Both failure modes report the real `bundle contains an unexpected directory: "static"` error, and the control case passes in every state, so the test cannot pass vacuously.

`TestGenerate_StaleStaticPathIsNotDestroyed` covers the guard's safety across both routes to the static path. A regular file is rejected with the path named, its contents intact, and the same `ErrCodeInvalidRequest` asserted for a recipe with no remote chart (reaching `removeStaleStaticDir`) and one with an upstream chart (reaching `ensureStaticDir`) — so the two routes cannot drift apart. A symlink is rejected and left in place, and a populated stale directory is preserved. The two file/symlink cases covering `ensureStaticDir` fail without the shared guard; the `removeStaleStaticDir` and populated-directory cases pass with or without it, serving as controls.

Package coverage: `pkg/bundler/deployer/argocdhelm` 82.6% → 83.7% (+1.1%).

`make qualify` was not run in full. Two `pkg/bundler/attestation` tests fail in this environment (`TestNewKeyVerificationIdentity_PEM`, `TestFetchAmbientOIDCToken`); both are pre-existing sandbox/network failures in a package this change does not touch, and both pass in CI.

## Risk Assessment

The change is confined to one deployer. Its effects are to skip creating a directory that would otherwise be empty, and to remove that directory when a previous run left it behind. Bundles that populate `static/` follow an identical code path; a non-empty `static/` is never removed, and a non-directory or symlink at that path is reported rather than deleted. Package documentation was updated alongside the code so it no longer describes `static/` as unconditional.

**Rollout notes:** No migration or flag. Bundles previously containing an empty `static/` could not be generated at all, so no existing artifact layout changes.

## Checklist

### DIFF
--- a/pkg/bundler/deployer/argocdhelm/argocdhelm.go
+++ b/pkg/bundler/deployer/argocdhelm/argocdhelm.go
@@ -40,8 +40,8 @@
 //     layer.
 //  3. Build a root values.yaml with ONLY dynamic paths (recipe defaults when
 //     available, empty strings otherwise). Static values stay in chart files.
-//  4. Write Chart.yaml + static/ + templates/ + values.yaml as a valid Helm
-//     chart.
+//  4. Write Chart.yaml + templates/ + values.yaml as a valid Helm chart,
+//     plus static/ when any component contributes a values file to it.
 //
 // This approach means changes to the Argo CD deployer (new component types,
 // sync policies, etc.) automatically flow through without duplication.
@@ -439,9 +439,14 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
 // targetRevision, and deployer defaults injected), and values.schema.json.
 // Each written file is appended to output.Files/TotalSize.
 func (g *Generator) writeValuesFiles(outputDir string, output *deployer.Output) error {
+	// PropagateOrWrap rather than Wrap: writeStaticValuesAndBuildStubs now
+	// reports a non-directory at the static path as ErrCodeInvalidRequest, and
+	// re-coding that as INTERNAL would present a bad --output path as a server
+	// fault — and over the API would withhold the cause from the response,
+	// since 5xx bodies omit it.
 	staticFiles, staticSize, dynamicOnlyValues, err := g.writeStaticValuesAndBuildStubs(outputDir)
 	if err != nil {
-		return errors.Wrap(errors.ErrCodeInternal, "failed to write static values", err)
+		return errors.PropagateOrWrap(err, errors.ErrCodeInternal, "failed to write static values")
 	}
 	output.Files = append(output.Files, staticFiles...)
 	output.TotalSize += staticSize
@@ -523,20 +528,40 @@ func (g *Generator) finalizeOutput(ctx context.Context, output *deployer.Output,
 	return nil
 }
 
-// writeStaticValuesAndBuildStubs writes each component's values to static/<name>.yaml
-// and builds the dynamic-only stubs map for the root values.yaml.
+// writeStaticValuesAndBuildStubs writes static/<name>.yaml for each component
+// that resolves to a remote Helm chart, and builds the dynamic-only stubs map
+// for the root values.yaml. When no component qualifies, static/ is not
+// created and a stale one from an earlier run is removed.
 func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string, int64, map[string]any, error) {
 	staticDir, err := deployer.SafeJoin(outputDir, "static")
 	if err != nil {
 		return nil, 0, nil, err
 	}
-	if mkdirErr := os.MkdirAll(staticDir, 0755); mkdirErr != nil {
-		return nil, 0, nil, errors.Wrap(errors.ErrCodeInternal, "failed to create static directory", mkdirErr)
-	}
 
 	var files []string
 	var totalSize int64
 	dynamicOnlyValues := make(map[string]any)
+
+	// static/ is created lazily, immediately before the first component
+	// values file is written. A recipe whose components all render as local
+	// charts (type Helm with no upstream Source, as the OCP overlays do)
+	// skips every iteration below, and an unconditional MkdirAll would then
+	// leave an empty directory that the bundle inventory validator rejects
+	// as unexpected. See issue #1942.
+	staticDirReady := false
+	ensureStaticDir := func() error {
+		if staticDirReady {
+			return nil
+		}
+		if err := checkStaticPath(staticDir); err != nil {
+			return err
+		}
+		if mkdirErr := os.MkdirAll(staticDir, 0755); mkdirErr != nil {
+			return errors.Wrap(errors.ErrCodeInternal, "failed to create static directory", mkdirErr)
+		}
+		staticDirReady = true
+		return nil
+	}
 
 	for _, ref := range g.RecipeResult.ComponentRefs {
 		// Skip non-Helm components (Kustomize, manifest-only)
@@ -584,6 +609,9 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string, 
 		}
 
 		// Write static values (dynamic paths removed)
+		if dirErr := ensureStaticDir(); dirErr != nil {
+			return nil, 0, nil, dirErr
+		}
 		staticPath, staticSize, staticErr := deployer.WriteValuesFile(staticValues, staticDir, ref.Name+".yaml")
 		if staticErr != nil {
 			return nil, 0, nil, errors.Wrap(errors.ErrCodeInternal,
@@ -593,7 +621,89 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string, 
 		totalSize += staticSize
 	}
 
+	// Generation writes directly into --output and the directory is not
+	// cleared between runs, so a failed pre-fix run leaves its empty static/
+	// behind. Without this, retrying into that same directory reproduces the
+	// original failure even though nothing new is created.
+	if !staticDirReady {
+		if err := removeStaleStaticDir(staticDir); err != nil {
+			return nil, 0, nil, err
+		}
+	}
+
 	return files, totalSize, dynamicOnlyValues, nil
+}
+
+// checkStaticPath rejects anything occupying the static path that is not a
+// plain directory. Both the create and the remove path call it, so a
+// pre-existing file or symlink produces the same ErrCodeInvalidRequest
+// regardless of whether the recipe happens to contribute static values —
+// otherwise the identical user error would surface as INVALID_REQUEST for an
+// all-local recipe and INTERNAL (from MkdirAll's ENOTDIR) for any recipe with
+// an upstream chart.
+//
+// The check matters most before removal, since os.Remove unlinks regular files
+// as readily as it removes empty directories. Lstat rather than Stat also
+// rejects a symlink planted after checksum.ValidateOutputRoot ran, which
+// happens before generation and permits pre-existing regular files.
+func checkStaticPath(staticDir string) error {
+	info, statErr := os.Lstat(staticDir)
+	switch {
+	case os.IsNotExist(statErr):
+		return nil
+	case statErr != nil:
+		return errors.Wrap(errors.ErrCodeInternal,
+			"failed to inspect static directory", statErr)
+	case info.Mode()&os.ModeSymlink != 0:
+		return errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("output path %q is a symbolic link", staticDir))
+	case !info.IsDir():
+		return errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("output path %q exists and is not a directory", staticDir))
+	}
+	return nil
+}
+
+// removeStaleStaticDir clears a static/ left behind by an earlier run when the
+// current generation contributed no files to it.
+//
+// Emptiness is checked explicitly rather than inferred from an os.Remove
+// errno, which keeps the intent readable and avoids platform-specific error
+// values. A populated static/ is kept: its contents are reported by
+// validateExactTree as unexpected output, though only when checksums are
+// enabled, since that validator runs inside the IncludeChecksums path. With
+// checksums off a stale static/ survives into the bundle, which is unchanged
+// from before this fix — a reused --output was never cleared, so stale files
+// of any kind already persisted.
+//
+// Anything that blocks removal of an empty directory fails generation rather
+// than being logged and ignored: leaving it behind either trips that same
+// validator or, when checksums are disabled, ships the bundle this fix exists
+// to prevent.
+func removeStaleStaticDir(staticDir string) error {
+	if err := checkStaticPath(staticDir); err != nil {
+		return err
+	}
+
+	entries, readErr := os.ReadDir(staticDir)
+	if readErr != nil {
+		if os.IsNotExist(readErr) {
+			return nil
+		}
+		return errors.Wrap(errors.ErrCodeInternal,
+			"failed to inspect static directory", readErr)
+	}
+	if len(entries) > 0 {
+		slog.Debug("keeping populated static directory from an earlier run",
+			"path", staticDir, "entries", len(entries))
+		return nil
+	}
+
+	if rmErr := os.Remove(staticDir); rmErr != nil && !os.IsNotExist(rmErr) {
+		return errors.Wrap(errors.ErrCodeInternal,
+			fmt.Sprintf("failed to remove empty static directory %q", staticDir), rmErr)
+	}
+	return nil
 }
 
 // valuesSchemaProperty, valuesSchemaDeployerProps, valuesSchemaDeployer,
--- a/pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
+++ b/pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
@@ -592,6 +592,273 @@ func TestGenerate_DataFiles(t *testing.T) {
 	}
 }
 
+// TestGenerate_StaticDirCreatedOnlyWhenPopulated verifies that static/ is
+// emitted only when at least one component contributes a values file.
+//
+// Components land in static/ only when they resolve to a remote Helm chart
+// (type Helm with a non-empty Source). A recipe whose components are all
+// local charts — the OCP overlays, which build from manifestFiles with no
+// upstream Source — contributes nothing, so an unconditionally created
+// static/ would remain empty. The bundle inventory validator walks the
+// finished tree and rejects any directory contributing no files, which
+// failed generation outright. See issue #1942.
+func TestGenerate_StaticDirCreatedOnlyWhenPopulated(t *testing.T) {
+	// Manifest-only components must carry post-manifests so the delegated
+	// argocd.Generator wraps them as local charts, mirroring how the OCP
+	// overlays supply manifestFiles.
+	localManifest := func(name string) map[string][]byte {
+		return map[string][]byte{
+			name + ".yaml": []byte("apiVersion: v1\nkind: ConfigMap\nmetadata:\n  name: " + name + "\n"),
+		}
+	}
+
+	localOnlyRefs := []recipe.ComponentRef{
+		{Name: "nfd-ocp-olm", Namespace: "nfd", Type: recipe.ComponentTypeHelm, Source: ""},
+		{Name: "nfd-ocp", Namespace: "nfd", Type: recipe.ComponentTypeHelm, Source: ""},
+	}
+	localOnlyValues := map[string]map[string]any{
+		"nfd-ocp-olm": {"enabled": true},
+		"nfd-ocp":     {"enabled": true},
+	}
+	localOnlyManifests := map[string]map[string][]byte{
+		"nfd-ocp-olm": localManifest("nfd-ocp-olm"),
+		"nfd-ocp":     localManifest("nfd-ocp"),
+	}
+
+	tests := []struct {
+		name           string
+		refs           []recipe.ComponentRef
+		values         map[string]map[string]any
+		postManifests  map[string]map[string][]byte
+		staleStaticDir bool
+		wantStaticDir  bool
+	}{
+		{
+			name:          "all components local charts omits static dir",
+			refs:          localOnlyRefs,
+			values:        localOnlyValues,
+			postManifests: localOnlyManifests,
+			wantStaticDir: false,
+		},
+		{
+			// A pre-fix failed run leaves an empty static/ in --output, and
+			// generation does not clear the directory between runs, so the
+			// retry must remove it rather than fail the same way again.
+			name:           "stale empty static dir from a prior run is removed",
+			refs:           localOnlyRefs,
+			values:         localOnlyValues,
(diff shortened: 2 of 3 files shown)


pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -530,14 +530,29 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
+	// skips every iteration below, and an unconditional MkdirAll would then
+	// leave an empty directory that the bundle inventory validator rejects
+	// as unexpected. See issue #1942.
+	staticDirReady := false
+	ensureStaticDir := func() error {
+		if staticDirReady {
+			return nil
+		}
+		if mkdirErr := os.MkdirAll(staticDir, 0755); mkdirErr != nil {
+			return errors.Wrap(errors.ErrCodeInternal, "failed to create static directory", mkdirErr)
```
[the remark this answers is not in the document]
yuanchen8911: The double-wrap is real, but it predates this change and does not originate at the flagged line.

This PR's call site returns the error unwrapped:

```go
if dirErr := ensureStaticDir(); dirErr != nil {
    return nil, 0, nil, dirErr
}
```

The wrap happens one level up in `writeValuesFiles`, which this PR does not touch:

```go
staticFiles, staticSize, dynamicOnlyValues, err := g.writeStaticValuesAndBuildStubs(outputDir)
if err != nil {
    return errors.Wrap(errors.ErrCodeInternal, "failed to write static values", err)
}
```

Before this change, the same function returned an identically coded error through that same path:

```go
return nil, 0, nil, errors.Wrap(errors.ErrCodeInternal, "failed to create static directory", mkdirErr)
```

So the behavior is unchanged. `ensureStaticDir` returns exactly what the inline `MkdirAll` returned.

There is a genuine defect underneath this, but it is broader than the flagged lines: `writeValuesFiles` also re-wraps the pre-existing `errors.New(errors.ErrCodeInvalidRequest, "invalid component name ...")` return, so an invalid component name surfaces to callers as `INTERNAL` rather than `INVALID_REQUEST`. Fixing that correctly means auditing every coded return from `writeStaticValuesAndBuildStubs`, not just the two call sites here.

Keeping this PR scoped to the empty-`static/` generation failure and filing the error-code masking separately.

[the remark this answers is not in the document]
yuanchen8911: Correcting my earlier reply: the "genuine defect underneath" claim did not hold up.

I filed the follow-up as #1946 and then closed it as not planned, because an end-to-end reproduction failed and the analysis showed both coded errors reaching that wrap are guarded upstream:

- `"invalid component name"` is explicitly defense-in-depth; `argocd.go:499` and `localformat/writer.go:352` validate names first.
- `resolveOverrideKey`'s `InvalidRequest` errors require `DynamicValues`, and `pkg/bundler/bundler.go:1985` rejects unknown components in a dynamic declaration with the correct `INVALID_REQUEST` before generation begins. That guard sits in the bundler core, so it covers both the CLI and the API.

The reproduction used a `--data` registry component deliberately lacking `valueOverrideKeys`, referenced from an overlay and passed via `--dynamic`, on both `argocd-helm` and `helm`. Both runs failed upstream with the correct code:

```
[INVALID_REQUEST] unknown component "testcomp" in dynamic declaration: not found in component registry
```

The three other `resolveOverrideKey` call sites in this file (1036, 1665, 1693) return the error unwrapped, so they mask nothing.

So there is no reachable defect and no demonstrated user-visible impact — a style inconsistency only, and one where most of the 30 sites are correct because they wrap uncoded stdlib errors.

The original point stands unchanged: this PR's call site returns the error unwrapped, the wrap is one level up on a line this PR does not touch, and the pre-fix code returned an identically coded error through that same path. No change to this PR.

pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
```diff
@@ -592,6 +592,273 @@ func TestGenerate_DataFiles(t *testing.T) {
...
+				Name:      "cert-manager",
+				Namespace: "cert-manager",
+				Type:      recipe.ComponentTypeHelm,
+				Chart:     "cert-manager",
+				Version:   "v1.20.2",
+				Source:    "https://charts.jetstack.io",
+			})
+			values["cert-manager"] = map[string]any{"replicaCount": 1}
+		}
+		rr := newRecipeResult("v1.0.0", refs)
```
njhensley, line 765: **🔵 Nitpick — Lstat-vs-Stat symlink rationale has no regression test**

The code comment justifies os.Lstat over os.Stat to reject a symlink planted after ValidateOutputRoot ran, but the only test hitting the `!info.IsDir()` branch seeds a regular file — non-dir under both Lstat and Stat. A future refactor to Stat would pass every test yet follow a symlink-to-dir, skip the rejection, and silently keep an unexpected tree. Shipped code is correct; this is future-refactor hardening (down-tiered from Minor).

**Blast radius:** Test-only gap; production is correct today.

**Fix:** Add a symlink-to-dir subcase asserting Generate fails with "is not a directory" and the target is untouched.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -593,9 +621,91 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
+	return nil
+}
+
+// removeStaleStaticDir clears a static/ left behind by an earlier run when the
+// current generation contributed no files to it.
+//
+// Emptiness is checked explicitly rather than inferred from an os.Remove
+// errno, which keeps the intent readable and avoids platform-specific error
+// values. A populated static/ is kept: its contents are reported by
+// validateExactTree as unexpected output, though only when checksums are
```
njhensley, line 673: **🔵 Nitpick — State-changing removal is silent; the no-op keep logs**

The keep-populated branch emits slog.Debug("keeping populated static directory..."), but the os.Remove branch that actually deletes a directory under the user's --output logs nothing. The action worth an audit trail is the one without a log line.

**Blast radius:** Diagnostics only — an operator debugging "why did my static/ disappear between runs" has no signal.

**Fix:** Add slog.Debug("removed stale empty static directory", "path", staticDir) on successful removal.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -439,9 +439,14 @@ func (g *Generator) Generate(ctx context.Context, outputDir string) (*deployer.O
...
 func (g *Generator) writeValuesFiles(outputDir string, output *deployer.Output) error {
+	// PropagateOrWrap rather than Wrap: writeStaticValuesAndBuildStubs now
+	// reports a non-directory at the static path as ErrCodeInvalidRequest, and
+	// re-coding that as INTERNAL would present a bad --output path as a server
+	// fault — and over the API would withhold the cause from the response,
+	// since 5xx bodies omit it.
 	staticFiles, staticSize, dynamicOnlyValues, err := g.writeStaticValuesAndBuildStubs(outputDir)
 	if err != nil {
-		return errors.Wrap(errors.ErrCodeInternal, "failed to write static values", err)
+		return errors.PropagateOrWrap(err, errors.ErrCodeInternal, "failed to write static values")
```
njhensley, line 449: **🔵 Nitpick — PropagateOrWrap fallback is dead code today**

Every error writeStaticValuesAndBuildStubs returns is already a *StructuredError, so PropagateOrWrap always propagates and the ErrCodeInternal fallback + message never apply.

**Blast radius:** None functionally — correct and defensive against a future non-coded return.

**Fix:** No change required; the comment at 442-446 already documents the intent.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -593,9 +621,91 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
 		totalSize += staticSize
 	}
```
njhensley, line 623: **🔵 Nitpick — No test for staticDirReady=true + stale leftover skip path**

removeStaleStaticDir only runs when the run wrote no static files. When a run does write static files but a stale foreign file also sits in static/, removal is skipped and validateExactTree is the sole (here untested) backstop.

**Blast radius:** Small; validateExactTree is the real backstop.

**Fix:** Optionally extend the remote-chart case with a pre-seeded static/leftover.yaml and assert Generate fails while the real static file is still written.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -593,9 +621,91 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
+// The check matters most before removal, since os.Remove unlinks regular files
+// as readily as it removes empty directories. Lstat rather than Stat also
+// rejects a symlink planted after checksum.ValidateOutputRoot ran, which
+// happens before generation and permits pre-existing regular files.
+func checkStaticPath(staticDir string) error {
+	info, statErr := os.Lstat(staticDir)
+	switch {
+	case os.IsNotExist(statErr):
+		return nil
+	case statErr != nil:
```
njhensley, line 654: **🔵 Nitpick — Lstat/ReadDir/Remove error branches uncovered**

The three errors.Wrap(ErrCodeInternal, ...) failure paths need filesystem fault injection (unreadable/perm-denied dirs) and are unexercised.

**Blast radius:** Minimal — standard unreachable-without-fault-injection error paths.

**Fix:** Acceptable to leave uncovered; a chmod-based subcase could cover ReadDir-error if desired.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -593,9 +621,91 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
 
+	// Generation writes directly into --output and the directory is not
+	// cleared between runs, so a failed pre-fix run leaves its empty static/
+	// behind. Without this, retrying into that same directory reproduces the
+	// original failure even though nothing new is created.
+	if !staticDirReady {
+		if err := removeStaleStaticDir(staticDir); err != nil {
+			return nil, 0, nil, err
+		}
+	}
```
njhensley, line 632: **🔵 Nitpick — Stale-dir cleanup is a per-deployer point patch, not a shared contract**

The lazy-create half matches localformat, but the removeStaleStaticDir stale-sweep is unique to argocd-helm; flux (#1947) still eager-MkdirAll's sources/ and localformat has lazy-create only — three levels of the same non-cleared-output concern with no shared helper.

**Blast radius:** Architecture/maintainability — each deployer re-derives its own output hygiene; drift increases with the flux twin (#1947).

**Fix:** Non-blocking. Follow-up: factor an ensure-dir-lazily + reap-if-stale-empty helper into pkg/bundler/deployer.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -593,9 +621,91 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
+// happens before generation and permits pre-existing regular files.
+func checkStaticPath(staticDir string) error {
+	info, statErr := os.Lstat(staticDir)
+	switch {
+	case os.IsNotExist(statErr):
+		return nil
+	case statErr != nil:
+		return errors.Wrap(errors.ErrCodeInternal,
+			"failed to inspect static directory", statErr)
+	case info.Mode()&os.ModeSymlink != 0:
```
njhensley, line 657: **🔵 Nitpick — Absolute output path echoed in a 4xx error detail**

The non-dir rejection embeds the absolute staticDir in an ErrCodeInvalidRequest, which server.WriteErrorFromErr surfaces in details["error"] for 4xx.

**Blast radius:** Minor server filesystem-layout disclosure — effectively unreachable because the API path bundles into a fresh per-request os.MkdirTemp; consistent with existing SafeJoin/ValidateOutputRoot 4xx messages.

**Fix:** None required; optionally log the path server-side and return a generic message.

pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -523,21 +528,41 @@ func (g *Generator) finalizeOutput(ctx context.Context, output *deployer.Output,
...
+	// static/ is created lazily, immediately before the first component
+	// values file is written. A recipe whose components all render as local
+	// charts (type Helm with no upstream Source, as the OCP overlays do)
+	// skips every iteration below, and an unconditional MkdirAll would then
+	// leave an empty directory that the bundle inventory validator rejects
+	// as unexpected. See issue #1942.
+	staticDirReady := false
+	ensureStaticDir := func() error {
+		if staticDirReady {
+			return nil
```
njhensley, line 554: **🔵 Nitpick — Directory-permission literal divergence across deployers**

New MkdirAll(staticDir, 0755) matches this file's convention (359/389/1466), but flux uses 0750 and there is no pkg/defaults constant for bundle dir perms.

**Blast radius:** Cosmetic/negligible — user-local bundle artifacts.

**Fix:** None for this PR; the package would benefit from a single defaults.BundleDirPerm constant.

njhensley (approved):
# Multi-persona review — Approve ✅

**Method:** 4 independent persona reviewers (Correctness · Security · Domain/Architecture · Test-coverage), then an adversarial senior meta-reviewer that re-derived every finding from the resolved code at `e682321`. 8 findings surfaced; all confirmed at nitpick tier (one down-tiered 🟡→🔵), zero refuted, no higher-severity issue found in a fresh adversarial pass.

## Overall assessment

Clean, correct, and unusually well-documented bug fix. Lazy `static/` creation (`ensureStaticDir` flips `staticDirReady` only after a successful `MkdirAll` that immediately precedes a write, so `!staticDirReady` is a precise "no static file contributed" signal), the guarded `removeStaleStaticDir` (Lstat type-check → explicit emptiness check → non-recursive `os.Remove` that fails closed on any error), and the `Wrap`→`PropagateOrWrap` switch are all sound. The meta-reviewer reproduced the crux (a hypothetical `Stat` swap would follow a symlink-to-dir; the shipped `Lstat` fails closed) and verified: no empty-dir leak, `templates/` is never empty, `validateExactTree` is genuinely exercised (`IncludeChecksums: true`), and the only concurrency race is a pre-existing whole-deployer property. Tests pass with `-race`.

Everything flagged is optional polish — none blocks merge. The two most worth a quick follow-up are the symlink test (inline on the test file) and a log line on the removal branch.

## Confirmed non-issues (examined, cleared)

- **`PropagateOrWrap` is the correct idiomatic choice** — the prior unconditional `Wrap(Internal, …)` mis-coded the new `ErrCodeInvalidRequest` non-directory case as a 5xx (which also withholds the cause over the API); the switch fixes it.
- **`os.Remove` blast radius is safe** — no-follow + non-recursive; `staticDir` is a fixed `SafeJoin` leaf; the Lstat→ReadDir→Remove TOCTOU window fails closed (`ENOTEMPTY ≠ IsNotExist`). No path traversal, no external delete.
- **No remote-chart regression** — any run writing ≥1 file keeps `static/`; the control test drives real `validateExactTree`.
- **`templates/`, `outputDir`, copy-tree mkdirs** are never empty when any component exists — the empty-dir pattern is unique to the remote-Helm-only `static/`.
- **Doc update** in `cli-reference.md` accurately matches the behavior.

## Summary

|:---:|:---:|:---:|:---:|
| 0 | 0 | 0 | 8 |

**Recommendation: Approve** — merge-ready; the eight inline nitpicks are optional polish.

PR #1950, opened 2026-07-29
test(bundler): cover argocd-helm OCP bundling and pin its error contract

yuanchen8911:
## Summary

Follow-up to #1944, which merged before this round of review fixes landed. Adds end-to-end coverage for the fix itself, makes the error-code assertion added there capable of failing, tightens a stale-output assertion, promotes one log to `Warn`, and scopes a threat-model comment to what the code actually delivers.

## Motivation / Context

#1944 changed `writeValuesFiles` to `errors.PropagateOrWrap` so a bad `--output` path reports as `INVALID_REQUEST` instead of an internal fault, and added an assertion intended to pin that.

The assertion could not fail. `stderrors.Is` walks the `Unwrap` chain and `StructuredError.Is` matches on `Code`, so under an outer `Wrap(ErrCodeInternal, ...)` the chain is `INTERNAL → INVALID_REQUEST` and the positive check still passes. `ExitCodeFromError` reads the **outermost** `StructuredError`, so reverting that one line flipped `aicr bundle` from exit 2 back to exit 8 with the suite green — verified by reverting it and watching the tests pass.

Related: #1942, #1944

## Type of Change

- [ ] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [ ] Other: N/A

## Implementation Notes

**End-to-end coverage.** No test bundled a real OCP recipe through `--deployer argocd-helm`. The chainsaw suite at `tests/chainsaw/cli/bundle-ocp/` used the default helm deployer, and `.github/workflows/kwok-recipes.yaml` skips `service==ocp` for every deployer in its matrix, so the merged fix had Go unit coverage only — the 100-overlay sweep that actually exercised it is not part of CI. Two steps are added to the existing suite: bundle the recipe it already generates with `--deployer argocd-helm`, and assert `static/` is omitted rather than emitted empty.

**Error-code assertion.** Adds the negative direction — no `ErrCodeInternal` anywhere in the chain — which is what actually pins `writeValuesFiles` to `PropagateOrWrap`. The positive assertion alone is satisfied by a buried code.

**Stale-output assertion.** The "populated static dir is preserved" subtest accepted any error from `Generate`, unlike its two siblings. It now asserts the specific message. Worth noting the validator reports the stale **directory**, not the file inside it: for an all-local recipe `static/` is not an expected directory at all, so the walk rejects it before reaching its contents. Asserting the file-level message would have failed.

**Log level.** The kept-stale-directory message moves from `Debug` to `Warn`. It reports artifact residue surviving into the bundle, which `Debug` puts below the CLI's default level.

**Threat model.** The `checkStaticPath` comment claimed `Lstat` "rejects a symlink planted after `checksum.ValidateOutputRoot` ran." That overstates it. `Lstat` rejects a symlink already present when generation reaches the path, but this is a check-then-act on a path name, so one planted between the check and the following `MkdirAll` or `Remove` is not caught.

No TOCTOU code was added. Closing that window would require `openat`-based operations throughout, and buys nothing in a supported configuration: it needs a local process with write access to `--output`, and anything with that access can already rewrite the finished bundle and its checksums. The comment is scoped to the guarantee the code delivers rather than the code being changed to model an attacker who has already won.

No behavior change beyond the log level.

## Testing

```bash
GOFLAGS="-mod=vendor" go test -race ./pkg/bundler/deployer/argocdhelm/ -count=1
golangci-lint run -c .golangci.yaml ./pkg/bundler/deployer/argocdhelm/...
yamllint -c .yamllint.yaml tests/chainsaw/cli/bundle-ocp/chainsaw-test.yaml
cd tests/chainsaw/cli/bundle-ocp && chainsaw test --test-dir . --no-cluster
```

Passes with the race detector; affected-package lint and yamllint both report 0 issues.

**Both new guards have a failing state**, which is the point — an assertion that cannot fail is not coverage.

| guard | against | result |
|---|---|---|
| error-code assertion | revert `PropagateOrWrap` → `Wrap` on current `main` | before: suite **green**; now: suite **fails** |
| chainsaw argocd-helm step | binary built from `fd26a4741^` (pre-#1944) | **fails** at `generate-argocd-helm-bundle` with `bundle contains an unexpected directory: "static"` |

The chainsaw suite was run locally both ways:

```bash
cd tests/chainsaw/cli/bundle-ocp
AICR_BIN=<fixed>   chainsaw test --test-dir . --no-cluster   # 1 passed
AICR_BIN=<pre-fix> chainsaw test --test-dir . --no-cluster   # 1 failed, at the new step
```

Of everything #1944 changed, the error-code half was the only part with no failing state at all.

## Risk Assessment

Test assertions, two added chainsaw steps, one log level, and one comment. No production code path changes; the only runtime effect is that a kept stale `static/` is now visible at the CLI's default log level.

**Rollout notes:** None.

## Checklist

### DIFF
--- a/pkg/bundler/deployer/argocdhelm/argocdhelm.go
+++ b/pkg/bundler/deployer/argocdhelm/argocdhelm.go
@@ -643,9 +643,23 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string, 
 // an upstream chart.
 //
 // The check matters most before removal, since os.Remove unlinks regular files
-// as readily as it removes empty directories. Lstat rather than Stat also
-// rejects a symlink planted after checksum.ValidateOutputRoot ran, which
-// happens before generation and permits pre-existing regular files.
+// as readily as it removes empty directories. A pre-existing regular file at
+// this path reaches generation on every route: checksum.ValidateOutputRoot
+// runs first on the DefaultBundler path but permits regular files, rejecting
+// only symlinks and special objects.
+//
+// Lstat rather than Stat additionally rejects a symlink here. On the
+// DefaultBundler path that is redundant, since ValidateOutputRoot already
+// rejects a symlink anywhere under the output root before any deployer runs;
+// it still holds for a caller that constructs this Generator directly and
+// bypasses that check.
+//
+// The guarantee stops there. This is a check-then-act on a path, so a symlink
+// planted between this call and the MkdirAll or Remove that follows is not
+// caught. Closing that would require openat-based operations throughout, and
+// it buys nothing in a supported configuration: it takes a local process with
+// write access to --output, and anything with that access can already rewrite
+// the finished bundle and its checksums.
 func checkStaticPath(staticDir string) error {
 	info, statErr := os.Lstat(staticDir)
 	switch {
@@ -694,7 +708,7 @@ func removeStaleStaticDir(staticDir string) error {
 			"failed to inspect static directory", readErr)
 	}
 	if len(entries) > 0 {
-		slog.Debug("keeping populated static directory from an earlier run",
+		slog.Warn("keeping populated static directory from an earlier run",
 			"path", staticDir, "entries", len(entries))
 		return nil
 	}
--- a/tests/chainsaw/cli/bundle-ocp/chainsaw-test.yaml
+++ b/tests/chainsaw/cli/bundle-ocp/chainsaw-test.yaml
@@ -69,10 +69,18 @@ spec:
       try:
         - script:
             content: |
+              # Each check below carries its own `|| { …; exit 1; }`, which is
+              # what makes it fail closed — `set -e` does not apply to the left
+              # side of `||`. set -eu is defence in depth for anything added
+              # later without that guard: a bare `test -f` returns 1 without
+              # terminating, and Chainsaw reads only the script's final exit
+              # status, so an unguarded check would pass silently.
+              set -eu
               WORK="/tmp/chainsaw-bundle-ocp"
-              test -f "${WORK}/bundle/deploy.sh"
-              test -f "${WORK}/bundle/README.md"
-              test -f "${WORK}/bundle/recipe.yaml"
+              for f in deploy.sh README.md recipe.yaml; do
+                test -f "${WORK}/bundle/${f}" \
+                  || { echo "missing scaffold file: ${f}" >&2; exit 1; }
+              done
               ls "${WORK}"/bundle/[0-9][0-9][0-9]-*/ >/dev/null 2>&1 \
                 || { echo "no numbered component folders found" >&2; exit 1; }
             check:
@@ -136,6 +144,57 @@ spec:
                 [ "${rdns_seq}" -lt "${cr_seq}" ] \
                   || { echo "readiness gate (${rdns_seq}) must precede CR ${cr_name} (${cr_seq})" >&2; exit 1; }
               done
+            check:
+              ($error == null): true
+
+    - name: generate-argocd-helm-bundle
+      description: >-
+        Bundle the same OCP recipe with --deployer argocd-helm. Every OCP
+        component is a local chart, so none contributes a static/ values file.
+        Emitting that directory anyway made it empty, and the bundle inventory
+        validator rejects a directory contributing no files, which failed
+        generation outright. See issue #1942.
+      try:
+        - script:
+            content: |
+              AICR_BIN="${AICR_BIN:?Set AICR_BIN to the path of the aicr binary}"
+              WORK="/tmp/chainsaw-bundle-ocp"
+              ${AICR_BIN} bundle \
+                --recipe "${WORK}/recipe.yaml" \
+                --output "${WORK}/bundle-argocd-helm" \
+                --deployer argocd-helm
+            check:
+              ($error == null): true
+
+    - name: assert-argocd-helm-omits-empty-static
+      description: >-
+        The bundle must be emitted without a static/ directory rather than with
+        an empty one, and must still carry its chart scaffold.
+      try:
+        - script:
+            content: |
+              # As above: the explicit `|| { …; exit 1; }` on each check is
+              # the safeguard, and set -eu covers anything added later without
+              # one. That matters most here, where the script ends in an `if`
+              # that yields 0 whenever static/ is absent, so an unguarded
+              # scaffold check would never fail the step.
+              set -eu
+              WORK="/tmp/chainsaw-bundle-ocp"
+              BUNDLE="${WORK}/bundle-argocd-helm"
+              for f in Chart.yaml values.yaml; do
+                test -f "${BUNDLE}/${f}" \
+                  || { echo "missing scaffold file: ${f}" >&2; exit 1; }
+              done
+              ls "${BUNDLE}"/[0-9][0-9][0-9]-*/ >/dev/null 2>&1 \
+                || { echo "no numbered component folders found" >&2; exit 1; }
+              # -e dereferences, so it reads false for a dangling symlink;
+              # -L catches that case and keeps the assertion honest about any
+              # entry at this path, not just a resolvable one.
+              if [ -e "${BUNDLE}/static" ] || [ -L "${BUNDLE}/static" ]; then
+                echo "static/ present; expected omitted for an all-local recipe" >&2
+                ls -la "${BUNDLE}/static" >&2
+                exit 1
+              fi
             check:
               ($error == null): true
       cleanup:
--- a/pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
+++ b/pkg/bundler/deployer/argocdhelm/argocdhelm_test.go
@@ -775,7 +775,27 @@ func TestGenerate_StaleStaticPathIsNotDestroyed(t *testing.T) {
 			RepoURL:                "https://github.com/example/repo.git",
 			TargetRevision:         "main",
 			ComponentPostManifests: map[string]map[string][]byte{"nfd-ocp": localManifest},
-			IncludeChecksums:       true,
+			// Reaches finalizeOutput -> checksum.WriteChecksums ->
+			// validateExactTree, the check that rejected the empty directory
+			// in #1942.
+			//
+			// This only affects the "populated static dir is preserved"
+			// subtest, the one that runs far enough to reach the validator.
+			// The file and symlink subtests fail earlier inside
+			// checkStaticPath and return before finalizeOutput, so they assert
+			// checkStaticPath's own message and DO pin the error code in both
+			// directions -- including the no-ErrCodeInternal check that keeps
+			// the CLI exit code at 2. Do not relax those.
+			//
+			// For the validator subtest only: production builds this Generator
+			// with IncludeChecksums false (the argocdhelm.Generator literal in
+			// DefaultBundler.buildDeployer) and reaches the same validator from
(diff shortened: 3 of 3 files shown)


pkg/bundler/deployer/argocdhelm/argocdhelm.go
```diff
@@ -644,8 +644,16 @@ func (g *Generator) writeStaticValuesAndBuildStubs(outputDir string) ([]string,
...
+// rejects a symlink already present at this path when generation reaches it,
+// which checksum.ValidateOutputRoot does not cover: it runs before generation
+// and permits pre-existing regular files.
+//
+// The guarantee stops there. This is a check-then-act on a path, so a symlink
+// planted between this call and the MkdirAll or Remove that follows is not
+// caught. Closing that would require openat-based operations throughout, and
+// it buys nothing in a supported configuration: it takes a local process with
+// write access to --output, and anything with that access can already rewrite
+// the finished bundle and its checksums.
```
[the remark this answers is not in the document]
yuanchen8911: The window is real, and the lines this comment is anchored to are the deliberate response to it.

`checkStaticPath` is a check-then-act on a path name, so a symlink planted between the `Lstat` and the following `MkdirAll` or `Remove` is not caught. That is stated in the comment at these exact lines rather than left implied, which is the second of the two remedies suggested here: explicitly document that the output directory is trusted and not concurrently modified.

Descriptor-relative operations were considered and rejected on reachability. Reaching the window requires a local process with write access to `--output` while generation is running. Anything holding that access can already rewrite the generated tree, the finished bundle, and `checksums.txt`, so redirecting `static/` grants an attacker nothing they do not already have. Closing it properly would mean converting the deployer's writes to `os.Root`/`openat` throughout — a substantial change to a shared write path — in exchange for no benefit in a supported configuration.

The supported flows do not expose it either. The server generates into a fresh `os.MkdirTemp` per request (`pkg/server/bundle_handler.go:168`), so there is no pre-existing tree and no second writer. The CLI writes to a directory the invoking user owns.

What the guard does deliver is the case that actually occurs: a pre-existing file or symlink at the static path, which `checksum.ValidateOutputRoot` permits because it runs before generation and only rejects symlinks and special objects at the output root. Both the create and remove routes reject it identically, and `os.Remove` is never reached for a non-directory — that is the regression this PR exists to prevent, since `os.Remove` unlinks regular files as readily as it removes empty directories.

Leaving this thread open for human reviewers rather than resolving it, since the trade-off is a judgement call rather than a factual dispute.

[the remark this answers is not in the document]
yuanchen8911: Correcting one claim in my previous reply, and withdrawing the closing line about leaving this open.

I wrote that the guard delivers "a pre-existing file or symlink at the static path, which `checksum.ValidateOutputRoot` permits." The file half is right; the symlink half is wrong.

`ValidateOutputRoot` is a recursive `filepath.WalkDir` over the whole output tree that `os.Lstat`s every entry and rejects anything with `ModeSymlink` at any depth (`inventory.go:200-216`). There is no early-out in `requireBundleRoot`, and `bundler.go:287` calls it unconditionally before any deployer runs. Only regular files pass.

Confirmed by running the CLI against a seeded output directory:

```
# symlink at <output>/static
[INVALID_REQUEST] output root contains a symlink or special object: .../static

# regular file at <output>/static
[INVALID_REQUEST] output path ".../static" exists and is not a directory
```

So a symlink is rejected by `ValidateOutputRoot` and never reaches `checkStaticPath`. The symlink branch there is reachable only for a caller constructing `argocdhelm.Generator` directly and bypassing `DefaultBundler` — the unit tests do exactly that — or inside the check-then-act window. That is what the pre-PR wording described, and my rewrite lost it.

The code comment is corrected in 64ce0ac6a: regular files are noted as reaching generation on every route, while the symlink rejection is marked redundant on the `DefaultBundler` path and load-bearing only for a direct caller.

For anything replaying this thread later: the guard's protection is rejecting a pre-existing **non-directory** at `static/`, which matters because `os.Remove` unlinks regular files as readily as it removes empty directories. Symlinks are handled upstream by `ValidateOutputRoot` on the production path.

The reachability argument against an `openat` rewrite is unaffected — reaching the residual window still requires a local process with write access to `--output`, which can already rewrite the finished bundle and its checksums.

lockwobr (dismissed):

tests/chainsaw/cli/bundle-ocp/chainsaw-test.yaml
```diff
@@ -69,10 +69,16 @@ spec:
       try:
         - script:
             content: |
+              # set -eu so a bare check cannot fail silently: without it a
```
njhensley: **🔵 Nitpick — set -eu comment attributes the safeguard to the wrong construct**

The comment says "without it a failing `test -f` … the whole guard passes with scaffold files missing." In the refactored form every check is `test -f … || { echo…; exit 1; }`, so the explicit `exit 1` — not `set -e` — is what makes the guard fail-closed (and `set -e` doesn't fire on the left of `||`). The stated mechanism describes the pre-PR bare-`test -f` form.

**Blast radius:** Cosmetic — no runtime effect. Same comment appears at line 174.

**Fix:** Keep `set -eu` (worthwhile defense-in-depth: backstops a future revert to bare checks, and `set -u` catches unset-var typos), but reword so it credits the explicit `|| exit 1` as the fail-closed mechanism rather than `set -e`.

tests/chainsaw/cli/bundle-ocp/chainsaw-test.yaml
```diff
@@ -138,6 +144,54 @@ spec:
...
+              set -eu
+              WORK="/tmp/chainsaw-bundle-ocp"
+              BUNDLE="${WORK}/bundle-argocd-helm"
+              for f in Chart.yaml values.yaml; do
+                test -f "${BUNDLE}/${f}" \
+                  || { echo "missing scaffold file: ${f}" >&2; exit 1; }
+              done
+              ls "${BUNDLE}"/[0-9][0-9][0-9]-*/ >/dev/null 2>&1 \
+                || { echo "no numbered component folders found" >&2; exit 1; }
+              if [ -e "${BUNDLE}/static" ]; then
```
njhensley: **🔵 Nitpick — [ -e static ] reads false for a dangling symlink**

`[ -e ]` dereferences, so a dangling `static` symlink would be treated as "correctly omitted."

**Blast radius:** Immaterial: checkStaticPath rejects a symlink at that path and the prior generate step (`$error == null`) would already have failed — the generator can't emit one.

**Fix:** None needed; noting for completeness. Could add `|| [ -L … ]` if belt-and-suspenders is wanted.

njhensley (dismissed):

**Method:** 3 independent persona reviewers (Correctness/Test-coverage · Operability/CI-DX · Security/threat-model) → an adversarial senior meta-reviewer re-deriving each finding from the resolved code. Line links pinned to head `64ce0ac6`.

### Overall assessment

Tight, high-quality test-hardening follow-up to #1944/#1942 — it does exactly what it claims. The headline change, the negative `!errors.Is(err, ErrCodeInternal)` assertion, was **mutation-verified**: reverting `writeValuesFiles` from `PropagateOrWrap` back to `Wrap(ErrCodeInternal, …)` fails *only* the new assertion while the pre-existing positive check stays green — reproducing the exit-2→8 regression the PR targets. The `unexpected directory: "static"` assertion matches `validateExactTree`'s real emitted string, the `IncludeChecksums:true` reasoning is correct, the new chainsaw argocd-helm steps are genuinely exercised in CI (via `.github/actions/cli-e2e`, not gated by the `kwok-recipes` `ocp` skip), cleanup is correctly re-scoped to the new last step, and the rewritten `checkStaticPath` threat-model comment is accurate and honestly scoped (TOCTOU window real, dismissal sound for supported configs; `ValidateOutputRoot` + direct-construction bypass both verified).

The two surviving items are cosmetic nitpicks; nothing blocks merge.

- Negative `!errors.Is(err, ErrCodeInternal)` — non-vacuous and load-bearing (mutation-tested).
- `unexpected directory: "static"` string matches `validateExactTree` output (`inventory.go:1244`).
- `IncludeChecksums:true` gating is correct; the file/symlink subtests still exercise `checkStaticPath`.
- `checkStaticPath` rewritten threat-model comment — accurate, not overclaimed.
- `Debug`→`Warn` log — appropriate, no sensitive leak.

### Summary

```
🔴 Blocker 0 | 🟠 Major 0 | 🟡 Minor 0 | 🔵 Nitpick 2     Recommendation: Approve
```

tests/chainsaw/cli/bundle-ocp/chainsaw-test.yaml
yuanchen8911: Correct, and your diagnosis of the cause is exact — that comment describes the pre-PR bare-`test -f` form. I wrote it against the broken version and did not update it after adding the explicit guards.

Confirmed the mechanism before rewording: with a missing file the script exits 1 both with and without `set -eu`, so the `|| { echo …; exit 1; }` is doing the work and `set -e` never fires on the left of `||`.

Reworded at both occurrences in e2120b58f. `set -eu` is kept and now described as defence in depth for anything added later without its own guard, rather than as the safeguard itself.

tests/chainsaw/cli/bundle-ocp/chainsaw-test.yaml
yuanchen8911: Your reachability analysis is right, and I should say plainly that you marked this "none needed" and I changed it anyway.

I added `|| [ -L …]` in e2120b58f. The reasoning was narrow: the assertion reads as "nothing at this path," and `[ -e ]` alone does not deliver that for a dangling link, so the check now matches what it claims regardless of whether the generator can produce one. Verified across the cases — dangling symlink goes from rc=0 to rc=1, while resolvable symlink, real directory, real file, and absent are unchanged.

That said, the cost was real: the push dismissed both approvals. Given you had explicitly waived it, folding it in with the reword was the wrong trade to make unprompted, and I would not do it that way again. Happy to revert that hunk if you would rather keep the diff to what you asked for.

njhensley (approved):
## Re-review — multi-persona + adversarial meta-review (delta)

Re-review against head `e2120b58` after a force-push that dismissed the prior approval. **Both prior 🔵 nitpicks are resolved and no new issues were introduced.**

### Prior-feedback status

| Prior finding | Disposition |
|---|---|
| 🔵 `set -eu` comment credited `set -e` for a safeguard the explicit `\|\| exit 1` provides | ✔️ Addressed — comments at `chainsaw-test.yaml:72` & `:174` reworded to credit the explicit `\|\| { …; exit 1; }` and frame `set -eu` as defense-in-depth |
| 🔵 `[ -e static ]` reads false for a dangling symlink | ✔️ Addressed — now `[ -e … ] \|\| [ -L … ]` with an explanatory comment |

### New findings

None. The two edits are correct: the rewritten comments are technically accurate, and the `\|\| [ -L … ]` addition sits in an `if` condition (where `set -e` is suppressed), so a genuinely-absent `static/` still exits 0 — no spurious failure.

### Summary

```
🔴 Blocker 0 | 🟠 Major 0 | 🟡 Minor 0 | 🔵 Nitpick 0     Recommendation: Approve
```

The base fix and its coverage were validated in the prior pass (negative `!errors.Is(err, ErrCodeInternal)` assertion mutation-tested; `unexpected directory: "static"` matches `validateExactTree`; `checkStaticPath` threat-model comment accurate; CI coverage via `.github/actions/cli-e2e`). This delta only improves the two documentation/robustness nitpicks.

Issue #1954, opened 2026-07-29
kai-scheduler chart unfetchable: registry.yaml points at `oci://ghcr.io/kai-scheduler/kai-scheduler`, which does not exist on ghcr


giladg:
## Summary

The embedded recipe registry pins kai-scheduler to an OCI location that is not
pullable — anonymously or authenticated:

```yaml
# recipes/registry.yaml (v0.14.0 through v0.18.0)
- name: kai-scheduler
  helm:
    defaultRepository: oci://ghcr.io/kai-scheduler/kai-scheduler
    defaultChart: kai-scheduler
    defaultVersion: v0.14.1
```

Since kai-scheduler is part of every recipe, this makes every resolved recipe
undeployable for consumers that fetch charts from the coordinates in
`ComponentRef.Source` (e.g. Go SDK `client/v1` integrators driving their own
Helm deployment). The failure is:

```
GET "https://ghcr.io/v2/kai-scheduler/kai-scheduler/manifests/v0.14.1":
GET "https://ghcr.io/token?scope=repository%3Akai-scheduler%2Fkai-scheduler%3Apull&service=ghcr.io":
response status code 403: denied: requested access to the resource is denied
```

## Evidence

Anonymous token request is denied (public ghcr packages grant anonymous pull
tokens, so this is not a rate-limit artifact):

```console
$ curl -s "https://ghcr.io/token?scope=repository:kai-scheduler/kai-scheduler:pull&service=ghcr.io"
{"errors":[{"code":"DENIED", ...}]}
```

With an authenticated GitHub token, a registry token is granted but the
repository is unknown — i.e. the package does not exist (or is private to an
org the token can't see):

```console
$ curl -s -H "Authorization: Bearer $TOKEN" \
    "https://ghcr.io/v2/kai-scheduler/kai-scheduler/tags/list"
{"errors":[{"code":"NAME_UNKNOWN","message":"repository name not known to registry"}]}
```

The previous public location exists and is anonymously pullable, but only up to
v0.13.0 — there is no v0.14.x there:

```console
$ curl -s -H "Authorization: Bearer $ANON_TOKEN" \
    "https://ghcr.io/v2/nvidia/kai-scheduler/kai-scheduler/tags/list?n=2000" | jq '.tags | length'
673        # newest release tags: v0.12.17, v0.13.0 — no v0.14.*
```

## Affected versions

`recipes/registry.yaml` carries these coordinates in at least v0.14.0, v0.15.0,
[...]

Issue #1927, opened 2026-07-27
aicr-triage: unresolved comment, approval, and closure-verification gaps


yuanchen8911:
## Summary

Three review findings on the `aicr-triage` skill were raised on #1911 and were still unresolved when it merged. All three concern the same surface — **what the skill posts, what the user approved before it posts, and how it verifies the result** — so they are tracked together here rather than in three places.

All three are verified against the current head of the skill. **Finding 2 is now implemented in #1956**; findings 1 and 3 remain open.

| # | Finding | Raised by | Thread | Severity |
|---|---|---|---|---|
| 1 | Comment body is posted without ever being shown for approval | @ArangoGutierrez | [3659605657](https://github.com/NVIDIA/aicr/pull/1911#discussion_r3659605657) | P2 |
| 2 | No-change verdicts leave no timeline signal | @mchmarny | [3647790325](https://github.com/NVIDIA/aicr/pull/1911#discussion_r3647790325) | P2 — **implemented in #1956** (the source's rule, in full) |
| 3 | Closure verification races the board's async Done workflow | @ArangoGutierrez | [3659605667](https://github.com/NVIDIA/aicr/pull/1911#discussion_r3659605667) | P3 |

## Related PRs

- **#1911** — the PR that shipped the skill. All three threads live there and remain unresolved.
- **#1956** — in flight, draft. Fixes the stall clock and the comment convention, plus two of @ArangoGutierrez's four findings (the unasserted `gh auth status` scope check, and `$ITEMS`/`$ACTIVE` filenames that carried no board identity). **It now also implements finding 2** — the source's rule in full, per the resolution history below.
- **`yuanchen8911:feat/aicr-triage-no-change-acks`** — the **wide** variant of finding 2, preserved. Previously stacked on #1956 and since folded in; kept as a reference so switching from narrow to wide is an edit rather than new work.

---

## 1. Comment body is posted without approval (P2)

Step 6 asks the user to approve the exact field write (`owner/repo#N — Status → Backlog, Priority → P2`). Step 7 then posts a public comment under that user's identity whose body the agent authors at apply time and never showed them.

The stated invariant is scoped to fields and closures — *"No field is edited and no issue is closed without this confirmation"* — so the comment falls outside it. Step 5's table is `issue | title | current | proposed action | reasoning`; neither it nor the Step 6 option text carries the comment body.

[...]

PR #1956, opened 2026-07-29
WIP: docs(skills): restore aicr-triage policy, harden mechanics

yuanchen8911:
## Summary

Hardens the `aicr-triage` skill — a port of @mchmarny's original `triage-issues` skill — for use on a shared org board, and **restores its triage policy to the original's** where this port had drifted from it.

## Motivation / Context

The skill is a port. Its **triage policy belongs to @mchmarny's original** and should not diverge: the promote, demote and close criteria, the comment forms, and the rule about which issues get commented. Its **mechanism** is where a shared org board demands more — guards, pagination, concurrency-safe temp paths, and verdicts a real board produces that the original did not need to name.

This PR does both halves.

### Policy restored to the original

Three divergences had crept in here, each introduced during review of this PR and each verified against @mchmarny's skill and against the board's own comment history:

| Divergence | Original | Evidence it was wrong |
|---|---|---|
| A "name the unchanged field" clause (`Triaged: P2→P1, Status stays Ready.`) | Bare transition: `Triaged: P2→P1.` | The bare form dominates the board; every comment carrying the clause was written during this PR's own review |
| Confirming form reordered to `confirmed at <Status> / <Priority>` | `confirmed at <Priority> / <Status>` | Original's template, and all four such comments on the board |
| Promote criteria overridden by a "must change what happens next" test that excluded in-review work | "A bug actively being worked on" and "In review and important to ship soon" **are** promote reasons | Original lists both explicitly |

All three are reverted. The invented clause was also the origin of several rule contradictions found across earlier review rounds; removing it removed them. The promote criteria govern outright. **Demote P1 → P2** is documented as the *later* correction when a promotion's premise expires — not a competing test at promotion time. #1668 is the worked sequence: promoted as actively-worked, demoted the same day once #1936 was in review.

### The triage rules this skill applies

Stated in full so the policy is reviewable without reading the diff. Rows marked
**@mchmarny** are his original's, verbatim in intent; rows marked *added* exist because
a real board produces states his three verdicts did not need to name.

**Verdict precedence** (highest first): Manual Review > Close > Incomplete > Demote > Promote > First-time > No change.

| Verdict | Applies when | Source |
|---|---|---|
| **Promote P2 → P1** | a bug actively being worked on; a direct unblocker for other tracked work; the cross-cutting parent goal of a set of epics that is the next priority; in review and important to ship soon | @mchmarny |
| **Demote Ready → Backlog** | blocked on upstream code, an external testbed, or future work ("once X lands"); an umbrella epic whose child issues are the actionable units; self-labeled Roadmap / RFC / Proposal; epics belong in Backlog, their children are Ready (source wording — used as a classification heuristic, verified before being asserted in a comment) | @mchmarny |
| **Close** | superseded by an architectural decision documented elsewhere; a tracking epic whose only deliverable is captured by a single child; a duplicate of a more specific issue | @mchmarny |
| **Demote P1 → P2** | something the board or repository can show has changed since the item became P1 — the work whose urgency justified it moved to Backlog or closed, the pressure behind it passed, or evidence contradicted its premise. Not applicable on a re-reading of facts that already held, and **"a fix is in review" is explicitly not a reason** — the source lists that as a *promotion* criterion | *added* |
| **Promote Backlog → Ready** | the blocker named in an earlier demotion has cleared; scope that was open is now settled; a newly filed issue meets the Ready bar and was parked only because nobody classified it | *added* |
| **Incomplete** | Status set without Priority, or Priority without Status — fill the missing field, and correct the populated one if the rules above say it is wrong | *added* |
| **First-time classification** | no Status set. Default Backlog / P2; Ready when well-scoped and urgent — **@mchmarny's direct instruction** when commissioning this (Slack, 2026-07-24). The P1/P0 refinements are port defaults, flagged as such in the file | @mchmarny + *added* |
| **Manual Review** | no exact executable write can be named — evidence unavailable, identity ambiguous, board state contradictory, or the operation unsupported on this board. Never written, never offered for confirmation | *added* |
| **No change** | correctly placed. No edit, but still commented | @mchmarny |

**Promotion criteria govern outright.** Demote P1 → P2 is a separate, later verdict gated on an observable change since the item became P1 — not a competing test applied at promotion time. Say in the promote comment what the promotion buys, so a later demote has something concrete to point at.

**Stalled** is information, never a verdict of its own: an `In progress` item whose `stallBase` exceeds 30 days. It bears on exactly one criterion, per the source's pitfall table — an item idle past the line does not satisfy "actively being worked on", so it is surfaced, never auto-promoted.

### Comment rules

| Situation | Comment | Source |
|---|---|---|
| Any field change | `Triaged: <transition>. <1–2 sentences of reasoning>` — arrow form, Unicode `→` (U+2192), no spaces | @mchmarny |
| No change | `Triaged: confirmed at <Priority> / <Status>. <1 sentence on why the placement is correct>` | @mchmarny |
| Close | `Closing: <reason>.` then a blank line, then what changed, where the work lives now, how to reopen | @mchmarny |
| Field previously unset | `Priority→P1` — the field name, never a bogus origin like `—→P1` | *added* |
| Both fields written | join the clauses with `, `, Status first | *added* |

Every triaged issue is commented, no-change included, skipped only when this run already commented on it — @mchmarny's rule. Field values are read off the board, never copied from an example. There is deliberately **no** clause naming the field that did not change: that was tried here, contradicted the original and the dominant form on the board, and was reverted.

### Mechanism added

- **Completeness.** The original defines three verdicts. A real board also has issues with no Priority, no Status or both (Incomplete, First-time), items closed but still in an active column, and ambiguous cases that must not be guessed (Manual Review). It needs the reverse directions `P1 → P2` and `Backlog → Ready`, which were being performed in practice and were unclassifiable.
- **Reliability.** Every fetch, slice and write fails closed, and every guard that can fire after the run directory exists deletes it inline before exiting. The vendored source has no abort guards at all — it verifies in prose. (No counts are quoted; hand-maintained tallies have been wrong four times in this PR's history.) A zero-byte output, an empty result, a loop that runs zero times or a swallowed exit status is a defect, not a pass.
- **Correct API behaviour.** `gh issue list --json comments` returns only the oldest 100 per issue; `gh api --jq` rejects `--arg`; `--paginate` emits one array per page; a pipeline discards the left-hand exit status; `gh auth status` prints one scopes line per account and none for fine-grained tokens; zsh does not word-split a scalar. Each is handled explicitly.
- **Cost.** Candidate reading is one API call per repository, not one per issue.
- **Concurrency.** Temp files carry board identity, so a run against a different board cannot silently read a stale dump, and they are deleted when the report prints.
- **Closure permission.** Project write and repository access are separate grants. A project editor with only `pull` on the repo could post a public `Closing:` comment and then fail at `gh issue close`, leaving a false record. Closures now verify `triage` or better first.
- **Run state is private.** Dumps hold full comment threads; with `TMPDIR` unset the `/tmp` fallback and umask 022 created them world-readable at a guessable path. They now live in a `0700` per-run directory (`$RUNDIR`), sentinel-marked so no guard can ever `rm -rf` a mistaken paste (pattern **and** sentinel checked at every fence, verified: zero unsafe delete sites). Cleanup follows the write boundary: pre-write aborts delete the directory inline; post-write (outcome-unknown) aborts keep it so Step 8 can verify against the dumps, deleting after the failure report. `run_start` is persisted inside `$RUNDIR`, read back, and the local clock is checked against GitHub's at Step 1 (abort beyond 300s skew, both directions).
- **Preflight the original lacks entirely.** No auth or scope check existed, so a read-only credential failed at the first write — after the user had already confirmed. The check now asserts the quoted `project` scope, is scoped to the active account, preserves `gh`'s exit status, and probes `viewerCanUpdate` on the target board for credentials that report no scopes.

### A bug inherited from the original

Staleness cannot be read from `updatedAt`. The skill's own `Triaged:` comments bump it, so triaging an issue makes it look freshly active and it can never age past the 30-day line. Visible on the board today: #870 reads `updatedAt` 2026-07-14 — a maintainer's `Triaged:` comment — against real activity from 2026-07-02.

`stallBase` fixes it here, derived from issue creation and non-bookkeeping comments only. **The original has the same defect**, since it mandates the same comments and reads the same field.

Fixes: N/A
Related: #1927, #1911

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Other: coding-agent skills (`.agents/skills/aicr-triage`)

## Implementation Notes

**Prerequisites — prove write capability, do not infer it.** `gh auth status` was printed, never checked. It is now scoped with `--active --hostname github.com`, because plain `gh auth status` prints one scopes line *per configured account* — an inactive account holding `project` would satisfy a naive grep while an under-scoped active account did the writing. Its own exit status is preserved, since `|| true` turned a real auth failure (expired token, no credentials) into the no-scopes branch and passed the preflight.

Where a scopes line exists it must carry the **quoted** `'project'`; a bare `grep project` is also satisfied by the read-only `read:project`. Where none exists — a fine-grained PAT or `GH_TOKEN`, which grant Projects access through read/write *permissions* rather than scopes — absence is not evidence of write capability, and read-only is a supported setting. That case is probed directly with `viewerCanUpdate` and fails closed.

**Step 2 — board identity in temp paths.** `$ITEMS`/`$ACTIVE` were fixed basenames, so a dump left by an earlier run against a *different* board could be picked up silently — and `$TMPDIR` resolves differently sandboxed vs not. Both now carry `-${owner}-${num}`, matching what the per-repo `$OPEN` path already did.

**Step 3 — `stallBase`.** The stall clock is derived from issue creation and non-bookkeeping comments only, so the skill's own output cannot reset it. Three things that derivation has to get right:

- **Comment pagination.** `gh issue list --json comments` expands to `comments(first: 100)` and paginates only the outer issue list, so an issue past 100 comments yields its *oldest* 100. Capped issues are re-fetched singly with `gh issue view`, which does paginate.
- **The stale bot.** `.github/workflows/stale.yaml` posts a notice at 90 days and exempts only by label, never by board Status, so without excluding it an untouched `In progress` item goes silent for its final 30 days before auto-close. The match is **anchored to the message opening** — an unanchored substring would also drop a contributor reply that quotes the notice.
- **Guards, including the loop that does the work.** An unguarded redirect still creates a zero-byte file, surfacing as an empty stalled table indistinguishable from "nothing is stalled". The capped-issue selector needed its own: an unguarded `jq` inside `for n in $(…)` exits nonzero, yields nothing, runs the loop zero times and still reports success — silently skipping every re-fetch. It now writes to a file read line by line, because hoisting it to a variable and iterating `$capped` is wrong in **zsh**, which does not word-split a scalar: two capped issues arrive as one iteration holding both numbers.

`updatedAt` is still fetched, for diagnostics only, and documented as never driving the verdict.

**Step 4 — verdicts the skill was performing but could not classify.** Added **Demote P1 → P2** and **Promote Backlog → Ready** with criteria and evidence requirements. Both were already happening in practice (#1668, #1266, #1954) while Step 4 defined only `Ready→Backlog` and `P2→P1`.

Also reconciled the promote rules with each other — three times, each earlier attempt over-correcting and being itself reverted; the churn is recorded in #1927. Where they landed: the source's criteria govern **as written** ("in review" carries its own qualifier, *"and important to ship soon"*; an item idle past the stall line does not satisfy "actively being worked on"). **Demote P1 → P2** and **Promote Backlog → Ready** are port additions the source does not define, flagged as such inline and in the Step 5 reasoning so the confirming human knowingly authorizes port policy; demotion requires an observable change since the item became P1, and "a fix is in review" is explicitly not a reason. The two P1→P2 comments on the board (#1668, #1266) predate that rule and would not satisfy it — they are why it exists, not precedents under it.

**Step 7 — the comment form is the source skill's.** `Triaged: <transition>. <reasoning>` for any field change, `Triaged: confirmed at <Priority> / <Status>.` for no-change, `Closing:` plus a blank line for closures. Arrow is Unicode `→` (U+2192), never ASCII. Three details the source leaves implicit are settled here because this port writes verdicts it did not: a previously-unset field takes `Priority→P1` rather than a bogus origin, a both-field verdict joins with `, ` Status-first, and values are read off the board rather than copied from an example. There is deliberately **no** clause naming the field that did not change — that was invented here, contradicted the source and the board, and is reverted.

**Comment coverage** follows the source — every approved verdict is commented, no-change included — with the four exclusions stated rather than inferred: Manual Review is never written, a declined no-change batch is not posted, a run that aborted before the comment step did not post, and a same-run repeat is skipped.

**Round-6 (Codex) fixes.** `$RUNDIR` deletion is now impossible against a mistaken paste (sentinel + pattern at every fence); post-write failures preserve run state for Step 8's verification instead of deleting the evidence the unknown-outcome contract needs; escalations resting on port-only P1/P0 rules present the sourced default as the proposed action with the escalation as a separate labeled Step 6 option; a declined ACK batch un-triages rather than silently un-commenting; duplicate-close targets resolve (and self-reference rejects) before the public `Closing:` posts; the dedupe clock is verified against GitHub's server; and a 500-open-issue repo no longer falsely aborts (501 sentinel).

**Round-5 fresh-review fixes.** Six abort guards inside `case` arms (plus the field-edit stanza, which lacked a `$RUNDIR` assignment entirely) were missing the inline cleanup the file claimed was universal — a blind spot of the earlier mechanical edit, which matched `; exit 1; }` but not `; exit 1 ;;`. Body files now carry repo identity (`comment-<owner>-<repo>-<n>.md`), closing the same cross-repo collision class the file already fixed for its dumps. A P0 whose incident has passed routes to Manual Review — the only zero-verdict state found in a full determinism walk — and the comment-content rules (one paragraph, backticked identifiers, blocker+reversal sentences) are now flagged as port additions. The "children may be Ready" softening was reverted to the source's wording, per the port's own source-wins rule.

**Comment bodies are data files, not heredocs.** Reasoning quotes issue-owned text, and a quoted line equal to a heredoc delimiter terminates it early and executes the remainder as shell — reproduced during review. Bodies are written to files under the private run directory and passed via `--body-file`; nothing issue-derived enters shell source.

**Step 8 — three-state `commented` column.** A nonzero exit from `gh issue comment` means *outcome unknown*, never "not posted". `no` is reached whenever an edit lands but Step 7 exits before the comment stanza — **not** only on multi-field verdicts: a single-field promotion can return nonzero while the write actually landed, and Step 8's re-read then confirms the field. That row reports `commented: no` and goes to Manual review — while the *field* is reported `unknown`, since a failed `item-edit` carries the same post-acceptance-timeout ambiguity and may well have landed.

## No-change comments

Every triaged issue gets a comment, including no-change verdicts, skipped only when this run already commented on the issue — the source's rule, unmodified. The port adds one consent mechanism on top, flagged as a port addition: Step 6 asks for the no-change comment batch as one all-or-nothing question. **Declining does not create comment-less triage** (which would break the source's coverage rule) — a declined batch means those issues were *reviewed but not triaged*: no ACK, not counted as triaged, reported as such so a later run picks them up.

#1927 tracks two further findings from #1911 that this PR does **not** address: comment bodies posted without being shown for approval, and closure verification racing the board's async Done workflow.

## Testing

**`make qualify` passes** — `MAKE_EXIT=0`, zero failures, "Codebase qualification completed" — run locally with `GITLAB_TOKEN` unset, from the worktree's `/private/tmp` real path, with **isolated, pre-warmed** `GOCACHE`/`GOLANGCI_LINT_CACHE`. The full-suite run required four attempts, and the failures along the way are worth recording because two are reusable lessons:

1. **Run 1** failed in `tests/releasepolicy` — one subtest, `script exceeded test deadline` at ~11s. Isolated re-runs of the same package pass on both clean `main` (23.0s) and this branch (22.0s), so this is CPU-contention flake of a ~10s per-script wall-clock deadline under full-suite parallel load, not a code failure. (Run 1 also *reported* exit 0 at first — the runner piped `make qualify | tail`, which returns tail's status. The pipeline-exit-status trap this very PR documents, reproduced in its own gate run.)
2. **Run 2** failed `lint-go` with 84 findings — every one in *other sessions'* worktrees (`../issue-1954/…`, `../issue1941…/…`), the shared-GOCACHE poisoning phantom. No finding referenced this tree.
3. **Run 3** (isolated but *cold* caches) failed `releasepolicy` again, four different subtests — the cold cache forced full-module recompilation during the test phase, starving the deadline-bound scripts harder.
4. **Run 4** (isolated caches, **pre-warmed** via `go build ./...` + test-binary compilation before the gate): **green end-to-end** — tests, lint, e2e, scan.
5. **Runs 5–8**, same recipe, one per subsequent review round's fixes, the last on the exact pushed head: all **green** (`MAKE_EXIT=0`, zero failures).

Runs 4 through 8 all passed; each of runs 5–8 ran after a review round's fixes, run 8 on the exact pushed head.

Skill-file snippet verification (all live, none executing the reviewed commit's code):
- **Scope/write preflight** — asserts quoted `'project'`, scoped `--active --hostname github.com`, preserves gh's exit status (an expired-token fixture previously slipped through `|| true`); `viewerCanUpdate` probes the target board unconditionally and returns `true` for NVIDIA/248; the org-then-user sequential form avoids the GraphQL NOT_FOUND-on-the-other-branch abort. Repo `triage` permission verified before closures.
- **`stallBase`** — reproduces the inherited bug on the live board (#870: `updatedAt` 2026-07-14 = a maintainer's `Triaged:` comment vs real activity 2026-07-02); capped-issue re-fetch measured on kubernetes/kubernetes#22368 (oldest-100 truncation: 2018 vs true 2026 newest); stale-bot anchor keeps a contributor reply that quotes the notice; guards negative-tested.
- **`run_start`/`$RUNDIR`** — persisted and read back (a stale hand-pasted timestamp reproduced `dupes=2` on #1668 before the fix); every fence assigns what it uses under the fresh-shell rule; write and read-back guarded; `mktemp -d` directory is `0700`, deletion fence verified.
- **zsh semantics** — `for n in $capped` collapses two ids into one iteration in zsh (reproduced; bash splits); the file+read-loop form yields two in both. All fenced blocks parse under `zsh -n`; frontmatter parses via `yq`.
- **`SOURCE.md`** — byte-identical to the file @mchmarny shared (verified after the provenance header; sha256 `2c25368b…af77090` matches the header's claim).

Three review rounds by an independent agent (verdict-construction across all nine verdict types, fence-by-fence variable walk, full `SOURCE.md` policy cross-check) plus two Codex reviews were applied; every confirmed finding is fixed in this head.

## Risk Assessment

**Rollout notes:** Affects only agent behavior when the skill is invoked; no runtime, build, or API surface. The confirmation gate is unchanged and still blocks every write.

## Checklist

- [x] Linter passes (`make lint`) — N/A, no Go/YAML changes
- [x] I added/updated tests for new functionality — N/A; every snippet verified against live data, results above
- [x] I updated docs if user-facing behavior changed — this PR is the doc


yuanchen8911:
cc @mchmarny — no rush on this, I know you are out. Flagging it so it is waiting when you are back, and the decision below is the only thing blocked on you.

This PR was originally the full response to your #1911 thread. I split it: the two fixes here are uncontested and independent, and the contested piece is held on yuanchen8911:feat/aicr-triage-no-change-acks, stacked on this branch and ready to open when you have weighed in.

The open question is not whether to comment on no-change verdicts — your argument for it is sound, and silence really is ambiguous from the issue's point of view. It is how wide to cast it. On the current board a routine pass changes about ten of eighty-eight issues, so commenting every no-change verdict is roughly 78 comments per run, and an issue sitting in Backlog for a year collects one confirmation per run for that whole year. I am not confident the twelfth "still correctly placed" carries information.

Two options, and I would rather you pick than have me guess:

1. As you proposed: comment on every triaged issue. The branch implements this today, and Step 6 lets a maintainer decline the whole no-change batch on any given run, so the volume is opt-out per run rather than permanent.

2. Narrower: comment only when the placement was actually investigated — the thread was read, a blocker cleared, the title and the evidence disagreed — and stay silent when it was confirmed from board fields alone. This keeps the signal you asked for, that an assignee learns someone looked, without the recurring churn on issues nobody is touching.

One thing worth knowing either way: adopting no-change comments without the stallBase fix in this PR would have silently disabled the stalled-issue detector, since every comment resets updatedAt. That is also a live bug independent of the ACK question. #870 shows it right now: updatedAt reads 2026-07-14, which is your own Triaged: comment, while the newest real activity is a contributor comment from 2026-07-02.

#1927 tracks the decision and stays open until you land on one.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -153,6 +160,33 @@ An active board item absent from `$OPEN` is closed: report it under Manual Revie
...
+OPEN="${TMPDIR:-/tmp}/aicr-triage-open-${repo//\//-}.json"
+
+# Newest human activity: issue creation, or the newest comment that is not one of
+# this skill's own "Triaged:"/"Closing:" notes.
+jq 'map({
+      number,
+      stallBase: ([ .createdAt,
+                    ( .comments[]
+                      | select((.body | test("^\\s*(Triaged|Closing):")) | not)
+                      | .createdAt ) ] | max)
```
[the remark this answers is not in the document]
yuanchen8911: Partly right, and it caught a real problem — but in the prose, not the filter. Fixed in f17caba4e.

The author-agnostic match is deliberate. The motivating case for this change, #870, is polluted by a maintainer's own hand-written Triaged: comment: updatedAt reads 2026-07-14 from that comment while the newest real activity is a contributor comment from 2026-07-02. Scoping the filter to the skill's identity would leave exactly that bug unfixed, so the suggested change would defeat the purpose of the PR.

The reachability half does not hold either. Triaged: and Closing: are reserved bookkeeping markers on this board, not free text — a substantive update never opens with them. The failure you describe requires someone to prefix a real status update with a marker that means the opposite.

What was genuinely wrong: the surrounding text described the filter as excluding "this skill's own" notes, which misdescribes the code and is presumably what pointed you at author matching. That wording now says the prefix is matched by design rather than the author, and cites #870 as the reason.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -51,6 +52,10 @@ rendered into any table, escape `|` and replace CR/LF with spaces in every cell.
 ```bash
 owner="<owner>"; num="<project-number>"   # default: NVIDIA / 248
 
+# Stamp the run start BEFORE any write — Step 7's dedupe compares comment
+# timestamps against it, so a value captured later would miss earlier comments.
+date -u +%Y-%m-%dT%H:%M:%SZ
+
```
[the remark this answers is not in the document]
yuanchen8911: Correct, and now fixed — but on the branch where this code actually lives.

This finding was raised against cfb9cc230, before the PR was split. run_start is no longer part of this PR: it supports the no-change dedupe, which moved to yuanchen8911:feat/aicr-triage-no-change-acks pending a maintainer decision on comment volume. Confirmed absent from the current diff.

The underlying point was right, so I applied it there in 0d2e12ddf. Step 1 now states that the variable does not survive into Step 7 because each block is a fresh shell, so the printed value must be pasted literally — and calls out the specific trap that re-running date in Step 7 would postdate this run's own comments and report zero duplicates every time.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -311,22 +361,86 @@ gh project item-edit \
...
+# Pipe to real jq, not `gh api --jq`: the latter rejects --arg ("unknown flag"),
+# and --paginate emits one JSON array per page, so -s flattens them ( .[][] ).
+dupes=$(gh api "repos/$repo/issues/$n/comments?per_page=100" --paginate \
+  | jq -s --arg me "$me" --arg since "$run_start" \
+      '[ .[][] | select(.user.login == $me)
+               | select(.created_at > $since)
+               | select(.body | test("^\\s*(Triaged|Closing):")) ] | length') \
+  || { echo "dedupe check failed for #$n — skip the comment and report"; exit 1; }
+
+[ "$dupes" = "0" ] || { echo "#$n already has a triage comment from this run — skipping"; }
```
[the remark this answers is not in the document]
yuanchen8911: Correct, and a real bug. Fixed in 0d2e12ddf, on the branch where the code now lives.

The guard was a bare test whose false branch only echoed, with gh issue comment following as a separate statement — so a nonzero duplicate count printed skipping and then posted the duplicate anyway, which is the exact failure the check existed to prevent.

The check and the comment are now one block with the comment inside the zero branch, and the surrounding text says why splitting them across two blocks reintroduces the bug.

Noting for the record that this was raised against cfb9cc230, before the PR was split; the dedupe path is not in this PR. It is on yuanchen8911:feat/aicr-triage-no-change-acks, held pending a maintainer decision on comment volume.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -230,11 +259,11 @@ forever.
 
 **No change** — correctly placed. Listed in the report, no comment, no edit.
```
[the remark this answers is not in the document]
yuanchen8911: Disagreeing with this one — it is reading the previous scope of the PR.

The PR was split after your earlier pass. Its objective is no longer to comment on every triaged issue; the current title and summary say the opposite, and the summary states explicitly that the PR does not change which issues the skill comments on. So "No change produces no comment" is not a contradiction, it is the unchanged existing rule that this PR deliberately leaves alone.

The presentation, confirmation, application, dedup, and reporting paths you are asking for do exist — implemented on yuanchen8911:feat/aicr-triage-no-change-acks, stacked on this branch. That work is held rather than merged because the volume tradeoff needs a maintainer's call: on the current board it is roughly 78 comments per run, and the maintainer who requested the behavior is out. #1927 tracks the decision.

The timeline visibility gap is real and acknowledged in the PR body. Closing it is deferred deliberately, not overlooked.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -153,6 +154,40 @@ An active board item absent from `$OPEN` is closed: report it under Manual Revie
...
+      stallBase: ([ .createdAt,
+                    ( .comments[]
+                      | select((.body | test("^\\s*(Triaged|Closing):")) | not)
+                      | .createdAt ) ] | max)
+    })' "$OPEN" > "${TMPDIR:-/tmp}/aicr-triage-stall-${repo//\//-}.json"
+```
+
+The prefix is a reserved marker on this board, not free text: a substantive
+update never opens with `Triaged:`. Matching on author instead would leave the
+#870 case unfixed, since the comment there is a maintainer's.
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in c6c1ce7d3.

The anchor line was 203, not 183 — line 183 is `# and #870, ...` inside the bash fence, which already has a space after the hash. Line 203 was prose that wrapped so `#870` landed at the start of a line: "would leave the / #870 case unfixed". Reflowed to "would leave / issue #870 unfixed" so the identifier is never line-initial.

Worth noting the rendered output was never wrong: CommonMark requires a space after `#` for an ATX heading, so GitHub renders `#870 case unfixed` as plain text. This is lint hygiene, not a rendering defect, and the repo has no markdownlint gate. Taking it anyway since the reflow costs nothing.

Swept the rest of the file for the same shape — a line matching `^#+[^ #]` outside a code fence — and found no others.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -379,6 +483,21 @@ Status must read Done; anything else goes to Manual Review. Print two tables:
...
+board moved with nothing on the issue to say why. Report those in the Applied table
+and *also* under Manual review, so the missing comment is visible rather than
+inferred from a blank column.
+
+**The comment column carries three states, matching the failure contract in Step 7 —
+never two.** A nonzero exit from `gh issue comment` means *outcome unknown*, not "not
+posted": the request can be accepted by GitHub and the response lost to a timeout.
+Recording that as `no` invites a second run to post a duplicate on an issue that
+already has the comment. Use `no` only when the command is known not to have posted —
+the dedupe check reported an existing comment, or the step was never attempted — and
```
njhensley: **🟡 Minor — Step 8 cites a "dedupe check" that exists nowhere in the skill**

This clause tells the agent to record `commented: no` when "the dedupe check reported an existing comment," but no comment-existence check exists anywhere in the file (`grep dedupe`/`run_start` → only this line), and per the diff this three-state paragraph is newly added, not a leftover. The only comment-dedupe machinery lives on the deferred no-change-acks branch (#1927). So a PR framed as "two self-contained fixes / does not change which issues get commented" forward-references unshipped behavior.

**Blast radius:** No wrong action — an agent cannot invoke a check that isn't there, so it falls through to the valid `unknown` / "step never attempted" paths. Impact is reader confusion + broken self-containment framing.

**Fix:** Drop the "the dedupe check reported an existing comment" clause; in this PR's scope `commented: no` is produced only by the "step was never attempted" abort path. Re-add the clause in #1927 where the check actually ships.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -379,6 +483,21 @@ Status must read Done; anything else goes to Manual Review. Print two tables:
...
+   (`yes` / `no` / `unknown`)
 2. **Manual review required** — issue | reason (fetch failed, ambiguous match,
    closed but Status is not Done, edit or comment outcome unknown)
+
+A field edit that landed while its comment failed is **not** a clean success: the
+board moved with nothing on the issue to say why. Report those in the Applied table
+and *also* under Manual review, so the missing comment is visible rather than
+inferred from a blank column.
+
+**The comment column carries three states, matching the failure contract in Step 7 —
```
njhensley: **🔵 Nitpick — "carries three states … never two" over-claims for the shipped scope**

With the timed-out-comment case now correctly routed to `unknown`, `commented: no` in the Applied column is reachable only via the phantom dedupe check (see prior comment) or a mid-Step-7 abort — whose items aren't Applied rows. So in the Applied column specifically, `no` is effectively unreachable in shipped scope, making "never two" aspirational there. Same root cause as the dedupe reference.

**Blast radius:** Report-formatting nuance only.

**Fix:** Soften to "may carry three states" and tie `commented: no` solely to the abort path until the dedupe check ships in #1927.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -318,15 +391,46 @@ expansion and command substitution.
...
+  error codes — so the comment survives being quoted elsewhere.
+- **On a blocker-based Demote, state the blocker as evidence and give the reversal
+  condition** — quote the issue's own dependency line or link the blocking issue, then
+  close with "Returns to Ready once the upstream chart ships." Without the reversal
+  sentence a demotion reads as a rejection, and nothing tells a future run what to
+  watch for.
+- **Structural demotes have neither, and must not invent them.** An umbrella epic or
+  a self-labeled Roadmap/RFC/Proposal belongs in Backlog because of what it *is*, not
+  because something blocks it — there is no blocker to cite and no event that returns
+  it to Ready. Say what carries the work instead: "Umbrella epic — the per-accelerator
```
njhensley: **🔵 Nitpick — Structural-demote worked example omits the mandatory arrow prefix**

Bullet 1 mandates every comment open in arrow form (`Triaged: Ready→Backlog.`). The structural-demote example here is prose-only ("Umbrella epic — the per-accelerator coverage work is carried by its child issues…") with no arrow prefix. It clearly illustrates only the replacement-for-reversal sentence, but an agent could read the fragment as the whole comment and omit the arrow — producing a comment worse than the base generic form.

**Blast radius:** Occasional missing arrow prefix on structural demotes.

**Fix:** Show the full form: `Triaged: Ready→Backlog. Umbrella epic — …` so the example obeys bullet 1.

.agents/skills/aicr-triage/SKILL.md
```diff
@@ -318,15 +391,46 @@ expansion and command substitution.
...
 ```
 
-Body format: first line starts with `Triaged:` (field updates) or `Closing:`
-(closures), followed by one or two sentences of reasoning. For closures, say
-what changed, where the work lives now, and how to reopen.
+**Match the board's established comment convention.** These comments are read as
+a series — a reader scanning an issue's history should see one voice, so follow
+the shape the board's maintainers already use rather than inventing a variant:
+
+- **Open with the transition in arrow form**, no spaces: `Triaged: Ready→Backlog.`,
```
njhensley: **🔵 Nitpick — "Arrow form" is not pinned to the → glyph**

The rule says "arrow form, no spaces" without naming the glyph; every example uses Unicode → (U+2192). An agent could render ASCII `->`, diverging from the board's established → comments — the exact one-voice inconsistency this section exists to prevent. Weak: the examples pin it by demonstration.

**Blast radius:** Cosmetic thread-voice drift.

**Fix:** Name the glyph explicitly: "the Unicode arrow →, not ->".

njhensley (dismissed):
## Multi-persona review — approve with comments

Reviewed at head `31b70673` via three independent persona reviewers (correctness of the embedded shell/jq/`gh`, prompt-followability, and board-domain fidelity), with every finding re-derived from the resolved code by an adversarial senior meta-reviewer.

**This is a sound, carefully-reasoned change.** The headline value is the Step 3 pagination fix: `gh issue list --json comments` silently caps at `comments(first:100)` (the *oldest* 100), which would corrupt the stall clock on long threads — the PR correctly detects capped issues and re-fetches them via `gh issue view`. The `stallBase` jq merge was independently reproduced across every edge case (`// .comments` fallback, empty `--slurpfile → []`, ISO-8601 lexical `max`, the `>=100` threshold, filename expansion, camelCase-vs-REST field casing) and holds up. This revision also added the two fail-closed guards on the terminal `jq` (`|| exit 1` + `test -s "$STALL"`), documented `updatedAt` as diagnostics-only, added the stale-bot exclusion (safe under-report-guard direction), and fixed the `Priority→P2`→`P1→P2` example — all improvements over the prior push, no regressions.

Nothing here blocks merge. Remaining notes are one self-consistency nit and some polish:

**🟡 Minor · 🔵 Nitpick — see inline comments.** The one worth a look before merge (🟡) is that Step 8 references a "dedupe check" that doesn't exist anywhere in the skill — it forward-references the comment-dedupe machinery that only ships in the deferred #1927 branch, which lightly contradicts this PR's "self-contained / does not change which issues get commented" framing. It causes no wrong action.

**🔵 Doc/body nit (not inline — about the PR description):** the Implementation Notes still say "record the reversal condition on **every** demotion," but the shipped file correctly carves out structural demotes ("Structural demotes have neither, and must not invent them"). Worth syncing the body so a future reader isn't misled — the file is the correct one.

### Confirmed non-issues (examined, not defects)
- **`test()` on a null `.body`** would throw — but `gh` emits `""` not `null` for empty comment bodies, so `"" | test(...)` returns false. Unreachable.
- **`stallBase` filtering by `Triaged:`/`Closing:` prefix instead of author** — validated as *by design*: #870's polluting comment is a human maintainer's, so an author filter would miss it; the only failure mode (a substantive comment opening `Triaged:`) over-reports staleness, which is the safe direction since stalled never produces a verdict.

|:---:|:---:|:---:|:---:|
| 0 | 0 | 1 | 4 |

Recommendation: **Approve with comments.**

.agents/skills/aicr-triage/SKILL.md
yuanchen8911: Fixed in 71ba62bbb. The clause is gone.

You were right that it forward-referenced unshipped machinery: the dedupe check lives only on the stacked no-change-ACK branch, not here. Worth noting the follow-on — with that clause removed, a later review round showed `commented: no` is still genuinely reachable in this PR's scope, just by a different path: Step 7 runs the field stanza once per changed field and stops on the first nonzero exit, so a two-field verdict can land the first edit, fail the second, and never attempt the comment. Step 8 now documents that path instead, and routes such a row to Manual review.

.agents/skills/aicr-triage/SKILL.md
yuanchen8911: Fixed in 71ba62bbb, though the conclusion moved after further review.

Your reachability analysis was right about the then-current text, and I initially rewrote the paragraph to say `no` was unreachable in the Applied column. That was wrong: a two-field verdict whose second edit fails leaves a row that is applied, uncommented, and whose comment was never attempted — a real `no`. The paragraph now leads with the distinction that matters, failed versus not posted, and documents that path. The failed field itself is reported `unknown` rather than failed, since a nonzero item-edit carries the same post-acceptance-timeout ambiguity as any other write.

.agents/skills/aicr-triage/SKILL.md
yuanchen8911: Fixed in 71ba62bbb, by restructuring rather than by editing the example.

Your point stood, and the same class of defect kept recurring: four separate times a fix to one bullet in that list broke a neighbouring bullet's rule. The shape rules are now replaced by a derived grammar — one clause per board field, joined — plus a table giving the exact comment for every Step 4 verdict. Structural demotes derive their clause like anything else and yield `Triaged: Ready→Backlog, Priority stays P2. <reasoning>`, so the example cannot drift from the rule again. All eight verdict types were checked against every remaining rule simultaneously; none is unsatisfiable.

.agents/skills/aicr-triage/SKILL.md
yuanchen8911: Fixed in 71ba62bbb. The rule now names the glyph explicitly: the Unicode arrow U+2192, never ASCII.

Checked it against the board while fixing: of 51 existing Triaged: comments, 45 use the tight arrow, one uses spaces around it, and none use ASCII. So this is established practice rather than a new constraint, and the text now says so with the counts, since a separate finding showed the section was presenting new tightenings as descriptions of existing practice.

yuanchen8911:
Closing without merge.

This started as a port of a 197-line skill and grew into an 1,100-line file across a dozen review rounds. Every fix was additive — another guard, another flag, another clarifying paragraph — and the accumulated result is too complicated to trust, execute, or maintain. That complexity is a worse defect than any of the individual ones the reviews caught.

What's worth keeping is recorded rather than merged. The source skill's rules stay authoritative as shared. The one real bug found in it — staleness read from `updatedAt`, which the skill's own `Triaged:` comments reset (see #870) — and the other open findings are tracked in #1927.

If this is attempted again, start minimal and stay close to the source. A port that needs this much hardening prose is telling you the design is wrong, not that it needs more prose.

PR #1959, opened 2026-07-29
docs(supply-chain): document monitoring your own signing identity

### LINKED ISSUE #1462: docs(supply-chain): guide users to monitor their own signing identity
## Problem

Users of `bundle --attest` and the evidence signing/collection features sign under their own identity and upload to Rekor by default. AICR documents how to sign and verify but never tells them to watch the log for their own identity (the producer-side analogue of AICR monitoring its own releases).

## Proposed approach

Add a "Monitoring your signing identity" section to `docs/integrator/supply-chain-verification.md` (with a pointer from `docs/user/artifact-verification.md`):

- Why a signer should monitor Rekor (an entry under their identity they did not produce signals compromise).
- How: the same `sigstore/rekor-monitor` reusable workflow configured with their signing identity (keyless cert SAN + OIDC issuer, or KMS key fingerprint).
- Both targets: public-good Rekor, and a private Rekor for enterprises on the `--trust-root` / private-Sigstore path (#1153).

Docs only; no code.

Part of epic #1149.

lockwobr:
## Summary

Documents how a signer monitors the transparency log for their own signing identity, the producer-side complement to the consumer-side verification the rest of these docs cover. Adds a `Monitoring Your Signing Identity` section to the integrator guide and a pointer to it from the user-facing verification page.

## Motivation / Context

Users of `aicr bundle --attest`, `aicr validate --emit-attestation --push`, and the evidence signing commands sign under their own identity and upload to a transparency log by default. AICR documented how to sign and how to verify, but never told signers to watch the log for their own identity. An entry under your identity that you did not produce is the signal that the identity was used without you, and for keyless signing it is the only signal: there is no private key on disk whose use could be audited, and the Fulcio certificate is short-lived.

Fixes: #1462
Related: #1149, #1153, #1650

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

## Implementation Notes

**The issue's proposed approach does not work as written, and the docs say so.** #1462 proposed pointing users at "the same `sigstore/rekor-monitor` reusable workflow". That workflow resolves its Rekor target from Sigstore's default signing config, which lists only Rekor v1, and Sigstore has stated the public-good instance will keep v1 as the default log for the foreseeable future. Since #1650, AICR's signing paths default to Rekor v2. So the upstream workflow cannot monitor the log that AICR users sign to by default.

The section therefore routes by log version rather than pretending one tool covers everything:

- **Rekor v2 (the default for every keyless and KMS path):** use `tools/rekor-monitor`, which resolves the v2 shard from the same signing config AICR signs against and takes the watched identity as flags. Includes a drop-in scheduled workflow derived from AICR's own `rekor-monitor.yaml`, with the checkpoint-artifact round trip and the two hardening steps called out.
- **Rekor v1, public or private:** use the upstream reusable workflow, which is what it supports. Covers both certificate identities and, for KMS, the `fingerprints` mode, with the `cosign public-key` pipeline to derive the fingerprint.

**Two coverage gaps are stated plainly rather than papered over:** a KMS key signing to the default Rekor v2 has no turnkey monitor (`tools/rekor-monitor` exposes only certificate-identity flags, though the upstream library it builds on does match v2 entries by fingerprint), and a private Rekor v2 reached via `--signing-config` is out of reach for both tools.

The per-mode identity table is grounded in the code rather than inferred: KMS entries carry a public key with no SAN or issuer to match, and `--tlog-upload=false` uploads nothing, so there is nothing to monitor.

Five upstream Sigstore references were added, each at the claim it supports: the threat model and Rekor overview for why monitoring is the detection control, the Rekor evolution post as the primary source for the v1-default statement, the `sigstore/rekor-monitor` repository, and Sigstore's rekor-monitor walkthrough for the consistency-check versus identity-search split and the fingerprint mode.

## Testing

```bash
make qualify
```

Docs-only change; no code paths touched. `tools/check-docs-mdx` passes (`OK: all doc files are MDX-safe`), including the `${{ ... }}` expressions inside the YAML fences. The lychee job in `fern-docs-ci.yaml` runs with `--offline`, so it gates local links and anchors only; the five new external URLs were each fetched and confirmed to return the expected content. The one new intra-repo anchor (`supply-chain-verification.md#monitoring-your-signing-identity`) resolves, and no duplicate heading slugs were introduced.

## Risk Assessment

**Rollout notes:** N/A. Additive documentation only; no existing headings renamed or restructured, so inbound anchor links are unaffected.

## Checklist


docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,259 @@ aicr trust update
...
+| Interactive keyless (`--attest`, browser or `--oidc-device-flow`) | Fulcio certificate: your email address in the SAN, plus the OIDC issuer that authenticated you | public-good Rekor v2 |
+| Non-interactive keyless (ambient CI OIDC, or `--identity-token`) | Fulcio certificate: the workload identity URL in the SAN; under GitHub Actions the issuer is `https://token.actions.githubusercontent.com` | public-good Rekor v2 |
+| KMS key (`--signing-key <kms-uri>`) | a public key, not a certificate: no SAN and no issuer to match, only the key itself | public-good Rekor v2 |
+| Private Sigstore (`--fulcio-url` / `--rekor-url`) | as keyless above, but issued by your own CA | your private Rekor v1 |
+| Rekor v1 opt-out (`--rekor-url https://rekor.sigstore.dev`) | as keyless above | public-good Rekor v1 |
+| Air-gapped KMS (`--tlog-upload=false`) | nothing is uploaded | none |
+
+Rekor v2 is the default for every keyless and KMS signing path in the CLI;
+`--rekor-url` and `--signing-config` are the opt-outs. See
+[Rekor v2 Signing](../contributor/rekor-v2-signing.md).
```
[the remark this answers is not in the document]
lockwobr: Confirmed and fixed in 799390ea. Verified in `pkg/cli/bundle.go`: `signingTargetFromFlags` takes only `rekorURL` and `signingConfigPath`, so `--fulcio-url` leaves `useTUFSigningConfig = true` and a private CA on its own still publishes to public-good Rekor v2. Split into two rows (private Fulcio alone, and private Fulcio with `--rekor-url`) and rewrote the following paragraph to state that `--fulcio-url` selects the CA, not the log.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,259 @@ aicr trust update
...
+          set -euo pipefail
+          id="$(gh api --paginate \
+            "repos/${GITHUB_REPOSITORY}/actions/artifacts?name=rekor-v2-checkpoint&per_page=100" \
+            | jq -rs '[.[].artifacts[] | select(.expired == false)]
+                | sort_by(.created_at) | last | .id // empty')"
+          if [ -z "${id}" ]; then
+            echo "No prior checkpoint; treating this as the first run."
+            exit 0
+          fi
+          gh api "repos/${GITHUB_REPOSITORY}/actions/artifacts/${id}/zip" > checkpoint.zip
```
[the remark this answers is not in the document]
lockwobr: Fixed in 799390ea. Agreed that a caveat after the sample is not enough when the sample is what gets copied. The `jq` selection now carries both guards inline (`head_repository_id == repository_id` and `head_branch == "main"`), lifted verbatim from AICR's own `rekor-monitor.yaml`, with a comment explaining that a poisoned checkpoint passes the consistency proof while collapsing the identity-scan window to empty. The no-checkpoint message was updated to match the narrower selection.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,259 @@ aicr trust update
...
+      file_issue: true
+      artifact_retention_days: 14
+      # Omit `url` for the public-good Rekor v1; set it to reach a private one.
+      url: https://rekor.internal.example.com
+      config: |
+        monitoredValues:
+          certIdentities:
+            - certSubject: ci@myorg\.example\.com
+              issuers:
+                - https://fulcio\.internal\.example\.com
```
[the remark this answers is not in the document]
lockwobr: Fixed in 799390ea. Also verified the premise rather than assuming it: `identity.CertMatchesPolicy` compiles `certSubject` with `regexp.Compile` and matches via `MatchString`, and does the same per entry of `issuers` (`vendor/github.com/sigstore/rekor-monitor/pkg/identity/identity.go:640-662`). `pkg/rekor/v1/identity.go` calls that same shared function, so this holds for the v1 example. Both values are now anchored, with a sentence explaining why.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,259 @@ aicr trust update
...
+   SAN and OIDC issuer, or its key fingerprint, against what you configured. An
+   unanchored regex is the usual cause of a false positive.
+3. **Treat an unexplained hit as compromise of the identity, not of the
+   artifact.** For keyless signing the OIDC account is what was taken: rotate its
+   credentials, revoke active sessions and tokens, and check the identity
+   provider's own audit log. For a KMS key, rotate the key and revoke the
+   principals allowed to sign with it.
+4. **Re-verify what you already published.** Run `aicr verify` over every bundle
+   you shipped under that identity, pinning the signer with `--require-creator`,
+   so you know which artifacts are genuine.
```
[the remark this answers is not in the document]
lockwobr: Good catch, and the most substantive finding here. Fixed in 799390ea. The original sentence claimed `--require-creator` tells you "which artifacts are genuine", which is backwards: a compromised identity satisfies that check exactly as well as the legitimate signer does, so it narrows the set without partitioning it. Rewritten to say so explicitly and to direct the reader to establish legitimacy from a record the attacker does not control, comparing digest and provenance against their own build and release history, and quarantining anything unaccounted for.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,259 @@ aicr trust update
...
+    permissions:
+      contents: read
+      actions: read   # read the prior run's checkpoint artifact
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          repository: NVIDIA/aicr
+      - uses: actions/setup-go@v7
+        with:
+          go-version-file: go.mod
```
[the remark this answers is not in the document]
lockwobr: Partially applied in 799390ea, with a deliberate divergence on the rest.

Applied: `actions/checkout` was cloning `NVIDIA/aicr` with no `ref:`, so a scheduled security job floated on our default branch and the tool being audited could change without review. Now pinned to `ref: v0.18.0` with a comment telling the reader to bump deliberately.

Not applied: baking full-length SHAs for `actions/checkout`, `actions/setup-go`, and the upstream reusable workflow into a documentation example. SHAs in prose rot at the next upstream release and readers copy stale ones, and upstream's own README documents `@main` for `reusable_monitoring.yml`, so pinning it here would diverge from the source the reader will check. Instead the existing advice was broadened from a bare "pin the actions by commit SHA" to cover every external reference in both examples (each `uses:`, the reusable workflow, and the new `ref:`), tied to the tag-versus-digest argument already made earlier in this guide. The security point is kept; the example stays readable and does not go stale.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,287 @@ aicr trust update
...
+            checkpoint_v2.txt.stall
+          retention-days: 30   # must exceed the cron interval
+          if-no-files-found: ignore
+```
+
+The examples use readable version tags so they stay legible. In production, pin
+every external reference in them by full-length commit SHA: each `uses:` action,
+the upstream reusable workflow in the next example, and the `ref:` above. A tag
+is mutable and can be repointed, which is the same argument the digest-pinning
+advice earlier in this guide makes for images.
```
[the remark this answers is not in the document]
lockwobr: Valid and consistent with the previous fix, so applied in a5335c96. The shell example was cloning `NVIDIA/aicr` at its default branch, which is the same floating-source risk as the unpinned `actions/checkout` `ref:`, just in the standalone snippet rather than the workflow. The clone is now `git clone --branch v0.18.0 --depth 1` with a comment to bump deliberately after review, and the pinning paragraph now enumerates the clone alongside each `uses:`, the reusable workflow, and the `ref:`, noting that the cloned source is the monitor being trusted.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+
+AICR hit that same wall and wrote `tools/rekor-monitor` for it. The tool resolves
+the v2 shard from the same signing config AICR signs against, then delegates the
+security-critical verification to upstream's library packages. It takes the
+watched identity as flags, so it monitors your identity as readily as AICR's:
+
+```shell
+# Pin the tool you audit. Cloning the default branch means the monitor can
+# change under you between runs; bump this ref deliberately after review.
+git clone --branch v0.18.0 --depth 1 https://github.com/NVIDIA/aicr && cd aicr
```
ArangoGutierrez: The pinned tag predates the contract described right below it: at v0.18.0 the monitor never prints a CLASSIFICATION= line, writes no .scan/.stall companions, and realMain returns only 0, 1 or 2 — never 3. Following this verbatim wires a Sigstore or network blip to exit 1, which this page defines as a security finding.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+    permissions:
+      contents: read
+      actions: read   # read the prior run's checkpoint artifact
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          repository: NVIDIA/aicr
+          # Pin the tool you audit. A scheduled security job should not float on
+          # another project's default branch; bump this deliberately.
+          ref: v0.18.0
```
ArangoGutierrez: Same version gap in the workflow example: at v0.18.0 there is no CLASSIFICATION= output to branch on, and the .scan/.stall paths uploaded below are never produced — which if-no-files-found: ignore then quietly hides. The CLASSIFICATION= hardening advice further down cannot be applied at this ref.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+name: Signer identity monitor
+on:
+  schedule:
+    - cron: "17 * * * *"
+permissions: {}
+concurrency:
+  group: signer-identity-monitor
+  cancel-in-progress: false
+env:
+  CERT_SUBJECT: '^ci@myorg\.example\.com$'
```
ArangoGutierrez: This pairs an email SAN with the GitHub Actions issuer, a combination Fulcio never issues — under that issuer the SAN is the workflow identity URL, exactly as the table above states and as AICR's own monitor is configured. Upstream ANDs subject and issuer, so a copied config matches nothing and the monitor stays green while watching an identity that cannot exist. Same pairing in the shell example above.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+      file_issue: true
+      artifact_retention_days: 14
+      # Omit `url` for the public-good Rekor v1; set it to reach a private one.
+      url: https://rekor.internal.example.com
+      config: |
+        monitoredValues:
+          certIdentities:
+            - certSubject: ^ci@myorg\.example\.com$
+              issuers:
+                - ^https://fulcio\.internal\.example\.com$
```
ArangoGutierrez: issuers is matched against the certificate's OIDC issuer extension (1.3.6.1.4.1.57264.1.8), which carries the identity provider that authenticated the signer — not the Fulcio CA that issued the cert. A private Fulcio URL here never matches; this wants the IdP issuer.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+            checkpoint_v2.txt.scan
+            checkpoint_v2.txt.stall
+          retention-days: 30   # must exceed the cron interval
+          if-no-files-found: ignore
+```
+
+The examples use readable version tags so they stay legible. In production, pin
+every external reference in them by full-length commit SHA: each `uses:` action,
+the upstream reusable workflow in the next example, the `ref:` above, and the
+`git clone --branch` in the shell example, since the cloned source is the
```
ArangoGutierrez: git clone --branch only accepts a branch or tag name, so this instruction cannot be followed as written — a SHA fails with 'fatal: Remote branch <sha> not found in upstream origin'. A clone followed by git checkout <sha> would work.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+          # from the live log head) passes the consistency proof while collapsing
+          # the identity-scan window to empty, silently skipping entries. So
+          # accept only artifacts produced by this repository on main:
+          # head_repository_id == repository_id rejects every fork run, and
+          # head_branch == "main" limits to the scheduled runs.
+          id="$(gh api --paginate \
+            "repos/${GITHUB_REPOSITORY}/actions/artifacts?name=rekor-v2-checkpoint&per_page=100" \
+            | jq -rs '[.[].artifacts[]
+                | select(.expired == false)
+                | select(.workflow_run.head_branch == "main")
```
ArangoGutierrez: Worth a one-line note on the branch assumption: this literal main rejects every artifact in a repo whose default branch is named differently, leaving the monitor permanently in first-run state — baselining and scanning nothing, on every run, without saying so.

ArangoGutierrez (commented):
Solid section — the threat model is right, and I confirmed the two stated coverage gaps and the fork-rejection jq against the code (the select clauses match the production workflow exactly). Six things worth fixing, the version pin being the big one.

1. The pinned tag predates the contract described right below it: at v0.18.0 the monitor never prints a CLASSIFICATION= line, writes no .scan/.stall companions, and realMain returns only 0, 1 or 2 — never 3. Following this verbatim wires a Sigstore or network blip to exit 1, which this page defines as a security finding. (docs/integrator/supply-chain-verification.md:544)
2. Same version gap in the workflow example: at v0.18.0 there is no CLASSIFICATION= output to branch on, and the .scan/.stall paths uploaded below are never produced — which if-no-files-found: ignore then quietly hides. The CLASSIFICATION= hardening advice further down cannot be applied at this ref. (docs/integrator/supply-chain-verification.md:592)
3. This pairs an email SAN with the GitHub Actions issuer, a combination Fulcio never issues — under that issuer the SAN is the workflow identity URL, exactly as the table above states and as AICR's own monitor is configured. Upstream ANDs subject and issuer, so a copied config matches nothing and the monitor stays green while watching an identity that cannot exist. Same pairing in the shell example above. (docs/integrator/supply-chain-verification.md:577)
4. issuers is matched against the certificate's OIDC issuer extension (1.3.6.1.4.1.57264.1.8), which carries the identity provider that authenticated the signer — not the Fulcio CA that issued the cert. A private Fulcio URL here never matches; this wants the IdP issuer. (docs/integrator/supply-chain-verification.md:689)
5. git clone --branch only accepts a branch or tag name, so this instruction cannot be followed as written — a SHA fails with 'fatal: Remote branch <sha> not found in upstream origin'. A clone followed by git checkout <sha> would work. (docs/integrator/supply-chain-verification.md:648)
6. Worth a one-line note on the branch assumption: this literal main rejects every artifact in a repo whose default branch is named differently, leaving the monitor permanently in first-run state — baselining and scanning nothing, on every run, without saying so. (docs/integrator/supply-chain-verification.md:614)

docs/integrator/supply-chain-verification.md
yuanchen8911: The v0.18.0 pin in both examples predates the behavior documented around them. At that tag, realMain exits 0/1/2 with no CLASSIFICATION= line, every runtime failure maps to exit 1, and there is no .scan/.stall resume state -- all of that landed in #1900/#1929, which no tag contains. A copied workflow can't branch alerting on CLASSIFICATION= or resume a backlog as the text describes, and wiring alerts to "exit 1 = security finding" pages on every network hiccup.

d4f7bef460dc8d1ef7ea0334a6935c0038de88e4 (the #1929 merge) is an executable pin that contains the classification, timeout, resumable-scan, and stall-tracking changes. Note `git clone --branch` resolves only branches and tags, so the shell example should be a normal clone followed by `git checkout --detach d4f7bef460dc8d1ef7ea0334a6935c0038de88e4`, and the workflow example's `ref:` should use the same SHA.

docs/integrator/supply-chain-verification.md
yuanchen8911: These two values can never match the same certificate. Matching is AND-ed across SAN and issuer (CertMatchesPolicy returns subjectMatches && issuerMatches), and a certificate issued via token.actions.githubusercontent.com carries a workflow-URL SAN, never an email -- the signing-mode table above says exactly this. As written the monitor runs green forever while matching nothing.

The Rekor v1 example below has the same shape of problem: the `issuers` field matches the certificate's OIDC-issuer extension, not the certificate authority, so `^https://fulcio\.internal\.example\.com$` there should be the private OIDC identity provider. Both examples should use a SAN/issuer pair read off one real certificate, per the "What your signature records" section.

docs/integrator/supply-chain-verification.md
yuanchen8911: Two paths silently re-baseline the monitor here, and both end with a clean exit while the prior identity window is permanently skipped (a missing zip is treated as first-run bootstrap, which baselines at the current head).

First, this workflow runs in the reader's repository, and for scheduled runs head_branch is that repository's default branch -- if it isn't `main`, this filter rejects every checkpoint the workflow itself uploads, so every run is a fresh baseline. Second, a pause longer than `retention-days: 30` expires the artifact with the same result.

A note can't make this fail closed, because the fetch step exits 0 either way. Bootstrap should be an explicit one-time action -- for example a workflow_dispatch input that permits starting without a checkpoint -- and a scheduled run that finds no artifact after that should fail as an operational error rather than silently skipping the gap. The branch also needs to be the repository's actual default branch rather than a hard-coded `main`.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+states that the public-good instance will continue using Rekor v1 as the default
+log for the foreseeable future. The workflow's inputs are `file_issue`,
+`artifact_retention_days`, `once`, `config`, and `url`; none of them selects a
+different signing config, and pointing `url` at a v2 shard falls through to its
+v1 client and fails.
+
+AICR hit that same wall and wrote `tools/rekor-monitor` for it. The tool resolves
+the v2 shard from the same signing config AICR signs against, then delegates the
+security-critical verification to upstream's library packages. It takes the
+watched identity as flags, so it monitors your identity as readily as AICR's:
```
yuanchen8911, line 550: For an actively signing identity this is not yet turnkey, and the section should say so. On any identity match the tool deliberately holds the cursor before the matching chunk so the finding re-alerts every run until triaged. That is the right default, but the only built-in suppression is `--known-tags-file`, which matches release tags carried in the SAN -- it applies to AICR's release identity and not to an email or generic CI identity. So the reader's first legitimate signature after baseline puts the monitor into alerting on every scheduled run, with no acknowledgment flag to move past it. (Manually editing the checkpoint files is not a safe workaround to suggest, since it can skip unreviewed entries.)

The accurate framing is that the tool is turnkey for identities that sign rarely enough to triage each hit, and for tag-bearing release identities via `--known-tags-file`; an actively signing generic identity needs an acknowledgment mechanism the tool doesn't have yet.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+One further hardening step from AICR's own workflow is worth copying before you
+rely on this: it branches notifications on the `CLASSIFICATION=` value, so that
+Sigstore or GitHub-API flakiness produces a quiet, self-healing failure rather
+than paging like a security event.
+
+### Monitoring Rekor v1 and a private Rekor
+
+If you opted signing out to Rekor v1 with `--rekor-url`, including a private
+Rekor v1 alongside a private Fulcio, the upstream reusable workflow is the right
+tool: v1 is what it supports. Sigstore's
```
yuanchen8911: Neither of this subsection's two cases actually works with the workflow shown.

For the public-good v1 log ("Omit `url` for the public-good Rekor v1"): this repository's own rekor-monitor.yaml records why upstream was rejected -- the v1 identity scan runs ~50x slower than the public log grows, so it can never keep up in a bounded CI job. After its initial baseline the recommended workflow provides no sustained identity coverage.

For a private Rekor v1 (`url: https://rekor.internal.example.com`): upstream's GetLogVerifier resolves the checkpoint key from Sigstore's public TUF trusted root, which does not contain a private log's key, and the reusable workflow exposes no input for private trust material -- so this fails at verifier setup, and making it work requires a fork or a different workflow, not a config change.

Both cases belong in the coverage-gaps paragraph rather than being presented as covered. What the upstream workflow does cover as shown is the fingerprint/identity config format itself, which the KMS fingerprint part of this section can keep.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+Then substitute that digest for the `config:` block in the workflow above:
+
+```yaml
+      config: |
+        monitoredValues:
+          fingerprints:
+            - <hex digest from the command above>
+```
+
+**Two coverage gaps, stated plainly.** A **KMS key signing to the default Rekor
```
yuanchen8911: There is a third gap: private Fulcio signing into the public-good v2 log. tools/rekor-monitor materializes its CA roots from Sigstore's public TUF trusted root only (ConfigureTrustedCAs with no override flag), and the upstream identity search skips any entry whose chain doesn't validate against those roots -- a stderr note and `continue`, not a match and not a failed entry. So entries under a private-CA certificate never reach the SAN/issuer match at all, and the table row "Private Fulcio alone -> public-good Rekor v2" is currently monitorable by neither tool.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+
+1. **Rule yourself out first.** Cross-check the entry's log index and integrated
+   timestamp against your own CI run history and any local signing you did. AICR
+   automates this for its release identity by correlating matches against its
+   release workflow's run history; for a personal identity the equivalent is
+   asking whether you signed anything at that moment.
+2. **Confirm the match is really your identity.** Re-read the entry's certificate
+   SAN and OIDC issuer, or its key fingerprint, against what you configured. An
+   unanchored regex is the usual cause of a false positive.
+3. **Treat an unexplained hit as compromise of the identity, not of the
```
yuanchen8911, line 900: Steps 3 and 4 assume interactive keyless signing with a public deployment bundle, but the section covers CI workload identities, KMS, private Sigstore, and recipe evidence.

For a CI workload identity there is no OIDC account session to rotate -- the SAN is a workflow URL, and containment means treating the repository as compromised: its workflow permissions, tokens, and environments.

Step 4's `aicr verify --require-creator` fits keyless bundles only. KMS bundles have no certificate creator and verify via `--key`; evidence artifacts verify through `aicr evidence verify` with `--expected-issuer` and `--expected-identity-regexp`. Worth splitting the containment and re-verification guidance by signing mode.

yuanchen8911 (changes requested):
This is a well-researched section and the signing-mode table, tool flag descriptions, upstream workflow inputs, and coverage-gap framing all check out against the code. But the two copy-paste workflows and the v1/private-log guidance have several silent false-negative paths, and the examples pin a tool version that lacks the behavior the text documents. Since silent false negatives are exactly what this section exists to prevent, these need fixing before merge. Details inline; the short list:

1. Both examples pin v0.18.0, which predates the CLASSIFICATION/exit-code/resume behavior the prose documents.
2. The v2 example's SAN/issuer pair can never match a real certificate, and the v1 example puts the Fulcio CA URL in the OIDC issuers field. Both configurations run green while matching nothing.
3. A missing or expired checkpoint artifact silently re-baselines the monitor at the current head, permanently skipping the gap.
4. The tool has no acknowledgment path for legitimate matches, so an actively signing identity wedges the monitor into permanent re-alerting.
5. The upstream reusable workflow cannot sustain identity coverage on the public-good v1 log and cannot verify a private v1 log, so the "Monitoring Rekor v1 and a private Rekor" section recommends a tool that works for neither of its two headline cases.
6. Private Fulcio into the public v2 log is a third coverage gap: tools/rekor-monitor silently skips entries whose chain doesn't validate against the public Fulcio roots.
7. The incident-response steps assume interactive keyless with a public bundle; the containment and re-verification guidance doesn't hold for CI workload identities, KMS, or evidence artifacts.

njhensley (commented):
Reviewed with a multi-persona panel (docs-accuracy, security/supply-chain, CI-DX) plus adversarial verification against the resolved code at `a5335c96`.

The standing feedback from @yuanchen8911 and @ArangoGutierrez already covers the crux issues — the `v0.18.0` pin lacking the `CLASSIFICATION=`/exit-3/`.scan`/`.stall` contract, the SAN/issuer pairs that can never match, `git clone --branch <sha>`, the hard-coded `head_branch == "main"`, and the coverage-gap framing — and I independently reproduced those against the code; **they still hold at this head.** Rather than restate them, the inline comments below are **only the net-new findings** those reviews did not raise, so this isn't piling on.

**Net-new (inline):**

- 🟠 The v1 example runs an unpinned `@main` third-party reusable workflow with `id-token: write` — the copied block can mint OIDC as the reader's repo. Production grants no such token and pins nothing to `@main`.
- 🟡 The v1 workflow's top-level `permissions: read-all` is over-broad and inconsistent with the v2 example's `permissions: {}`.
- 🟡 The signing-mode table keys its KMS/private/v1 rows on flags (`--rekor-url`, `--signing-key`, `--signing-config`, `--tlog-upload`) that exist only on `aicr bundle` — the `validate --push` / `evidence publish` / `evidence sign` paths are keyless-only, always Rekor v2 (`pkg/cli/validate_evidence.go:117`).
- 🔵 The v2 example drops production's `workflow_dispatch`, the `retention-days` rationale, and the retry/atomic-write fetch wrapper.

Nothing here is a hard blocker on its own; the merge-blocking items are the ones already in @yuanchen8911's standing review. This is really strong, carefully-researched work — the fix set is small and mostly mechanical.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+  schedule:
+    - cron: "0 * * * *"
+permissions: read-all
+jobs:
+  monitor:
+    permissions:
+      contents: read
+      issues: write
+      id-token: write
+    uses: sigstore/rekor-monitor/.github/workflows/reusable_monitoring.yml@main
```
njhensley: **🟠 Major — v1 example runs an unpinned @main reusable workflow with id-token: write**

The copied v1 block runs `sigstore/rekor-monitor/.github/workflows/reusable_monitoring.yml@main` — a mutable ref — under a job granting `id-token: write` + `issues: write`. Unpinned upstream code executes in the reader's repo context and can mint an OIDC token as their repo (authenticating to any cloud that trusts that repo's OIDC subject) and write issues. The doc's own pin-by-SHA advice sits between the two examples, but the block a reader copies verbatim is `@main`. Production (`.github/workflows/rekor-monitor.yaml`) never grants `id-token: write` and calls no third-party reusable workflow — this introduces a privilege the hardened original deliberately avoids.

**Blast radius:** Any reader who copies the v1 example gains an hourly job in which unpinned third-party code can obtain their repo's OIDC identity and issue-write — a supply-chain escalation inside a guide whose selling point is supply-chain hardening.

**Fix:** SHA-pin the `uses:` in the example itself (not only in surrounding prose), and drop `id-token: write` unless upstream `reusable_monitoring.yml` documents requiring it.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+[walkthrough of rekor-monitor](https://blog.sigstore.dev/using-rekor-monitor/)
+covers the same ground in more depth, including the split between its consistency
+check and its identity search, and the key-fingerprint mode used below.
+
+```yaml
+name: Signer identity monitor (Rekor v1)
+on:
+  schedule:
+    - cron: "0 * * * *"
+permissions: read-all
```
njhensley: **🟡 Minor — v1 top-level `permissions: read-all` is over-broad and inconsistent with the v2 example**

The v1 workflow sets `permissions: read-all` at top level, then the job re-declares `permissions:`. Job-level permissions replace workflow-level, so `read-all` is dead for the monitor job (misleading), and any second job a reader adds without its own `permissions:` block silently inherits read to every scope — violating least-privilege. The v2 example correctly models `permissions: {}`.

**Fix:** Change the top-level to `permissions: {}` to match the v2 example's fail-closed default; the job block already grants exactly what it needs.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+```shell
+# Fulcio certificate from a keyless-signed bundle. The SAN is under
+# "X509v3 Subject Alternative Name"; the OIDC issuer is extension
+# 1.3.6.1.4.1.57264.1.8.
+jq -r '.verificationMaterial.certificate.rawBytes' \
+  ./my-bundle/attestation/bundle-attestation.sigstore.json \
+  | base64 -d | openssl x509 -inform DER -noout -text
+```
+
+| Signing mode | What the log entry identifies you by | Log |
```
njhensley, line 508: **🟡 Minor — Bundle-only signing flags are presented as general across all four signing paths**

The intro lists four uploaders (`aicr bundle --attest`, `validate --emit-attestation --push`, `evidence publish`, `evidence sign`), but the flags the table keys its rows on — `--signing-key`, `--fulcio-url`, `--rekor-url`, `--signing-config`, `--tlog-upload` — are registered only on `aicr bundle`. `pkg/cli/validate_evidence.go:117` states the other three paths 'expose no Rekor-v1 override today' and are keyless-only, always targeting the default Rekor v2. So the table's KMS / private-Fulcio / private-log / v1 / air-gapped rows apply only to `aicr bundle`; a reader monitoring their `evidence publish` identity cannot use `--rekor-url` to opt into v1 or `--signing-key` for KMS as the table implies.

**Fix:** Scope the flag-keyed rows to `aicr bundle` explicitly, or add a sentence that the `validate`/`evidence` paths are keyless-only and always write to the default Rekor v2.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+`1` is a security finding (`tamper` or `identity`), and `3` is a non-security
+failure (`operational` for Sigstore, Rekor, or network trouble; `degraded` when
+the log is outpacing the per-run scan). Every completed run prints a matching
+`CLASSIFICATION=` line to branch on. Full flag and exit-code reference:
+[`tools/rekor-monitor/README.md`](https://github.com/NVIDIA/aicr/blob/main/tools/rekor-monitor/README.md).
+
+On a schedule, in your own repository:
+
+```yaml
+name: Signer identity monitor
```
njhensley, line 622: **🔵 Nitpick — v2 example omits `workflow_dispatch` (no manual seed / test / re-arm path)**

The `on:` block has only `schedule:`, dropping the production workflow's `workflow_dispatch: {}`. A schedule-only security monitor cannot be run on demand: (1) the reader can't fire a run to seed the first checkpoint or prove it works; (2) GitHub does not run a new `schedule:` workflow until it lands on the default branch, so a PR-branch test never executes; (3) GitHub auto-disables scheduled workflows after 60 days of repo inactivity, with no manual path to re-arm.

**Fix:** Add `workflow_dispatch: {}` under `on:`, matching production.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+        if: ${{ !cancelled() }}
+        with:
+          name: rekor-v2-checkpoint
+          # The cursor plus its .scan and .stall companions. All three must
+          # travel together so the next run resumes an in-progress catch-up.
+          path: |
+            checkpoint_v2.txt
+            checkpoint_v2.txt.scan
+            checkpoint_v2.txt.stall
+          retention-days: 30   # must exceed the cron interval
```
njhensley: **🔵 Nitpick — `retention-days: 30` justification comment understates the real constraint**

The comment says the value 'must exceed the cron interval', but 30 days trivially exceeds the 1-hour cron. The binding constraint is that the cursor must survive the longest gap between two successful runs (outages, held cursors, a scheduling pause, the 60-day auto-disable), itself capped by GitHub's 90-day max artifact retention. The value 30 is fine; the rationale is what's off. Production frames it correctly ('wide enough to survive a brief scheduling pause').

**Fix:** Reword to 'must comfortably exceed the longest expected gap between successful runs' (or reuse production's phrasing).

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,290 @@ aicr trust update
...
+            | jq -rs '[.[].artifacts[]
+                | select(.expired == false)
+                | select(.workflow_run.head_branch == "main")
+                | select(.workflow_run.head_repository_id == .workflow_run.repository_id)]
+                | sort_by(.created_at) | last | .id // empty')"
+          if [ -z "${id}" ]; then
+            echo "No prior checkpoint from this repository on main; treating this as the first run."
+            exit 0
+          fi
+          gh api "repos/${GITHUB_REPOSITORY}/actions/artifacts/${id}/zip" > checkpoint.zip
```
njhensley: **🔵 Nitpick — Fetch step drops production's retry/atomic-write wrapper**

The bare `gh api .../zip > checkpoint.zip` replaces production's `.github/scripts/gh-api-retry.sh`, which retries 5xx/429 and publishes via `OUTFILE.part` only on success. A single transient GitHub-API blip now reddens the whole monitor job with no retry. This is resilience, not correctness/security: `set -euo pipefail` + `gh api`'s non-zero exit on HTTP≥400 means a failed download aborts the step before the Scan step consumes the zip (the tool tolerates a missing zip as first-run), so it fails closed.

**Fix:** Optional — add a one-line note that production wraps this fetch in a retry helper, or keep as-is since it fails closed.

docs/integrator/supply-chain-verification.md
lockwobr: Correct, and this was the worst defect here. Fixed in d3fc34ee. Verified: v0.18.0 was published 2026-07-23 and contains zero occurrences of `CLASSIFICATION` in `tools/rekor-monitor/main.go`, and `git tag --contains d4f7bef4` returns nothing, so no tag carries the contract the prose describes. Both examples now pin `d4f7bef460dc8d1ef7ea0334a6935c0038de88e4` (the #1929 merge), where `git grep CLASSIFICATION` hits main.go:491,503 and the exit-code table proves operational/degraded map to 3. `git diff d4f7bef4 origin/main -- tools/rekor-monitor` is empty, so the documented flags and the .scan/.stall companions all exist at that exact ref.

docs/integrator/supply-chain-verification.md
lockwobr: Same fix, d3fc34ee. The workflow example now pins the same SHA, so the CLASSIFICATION= branching advice and the .scan/.stall upload paths are actually reachable at the documented ref rather than being silently swallowed by if-no-files-found: ignore.

docs/integrator/supply-chain-verification.md
lockwobr: Correct, and this one was self-contradicting: the table two screens above states the rule the examples then violated. Fixed in d3fc34ee. Confirmed the AND at `vendor/github.com/sigstore/rekor-monitor/pkg/identity/identity.go:666` (`return subjectMatches && issuerMatches`), so the pairing matched nothing and the monitor would have run green forever. Both examples now use a workflow-URL SAN with the Actions issuer, modelled on production's CERT_SUBJECT, and the prose states that an email SAN pairs with the interactive IdP issuer `https://oauth2.sigstore.dev/auth` (`pkg/bundler/attestation/oidc.go`).

docs/integrator/supply-chain-verification.md
lockwobr: Correct. Fixed in d3fc34ee. `CertMatchesPolicy` reads `certExtensionOIDCIssuerV2` off the certificate and never the CA, so a Fulcio URL there could not match. The example now uses a private IdP issuer and the prose names extension 1.3.6.1.4.1.57264.1.8 explicitly so the distinction is not left implicit.

docs/integrator/supply-chain-verification.md
lockwobr: Correct, the instruction was unrunnable as written. Fixed in d3fc34ee: plain `git clone` followed by `git checkout --detach d4f7bef4...`, with a comment noting that `git clone --branch` resolves only branch and tag names.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed on the problem, and taking it further than a note. A note cannot make this fail closed, since the fetch step exits 0 either way. In d3fc34ee the branch filter now comes from `github.event.repository.default_branch` rather than a literal `main`, and bootstrap is explicit: a `workflow_dispatch` input permits starting without a checkpoint, and a scheduled run that finds none fails with an `::error::` instead of silently re-baselining.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed on all points, and I took your suggested pin. Fixed in d3fc34ee: both examples use `d4f7bef460dc8d1ef7ea0334a6935c0038de88e4`, and the shell example is a plain clone plus `git checkout --detach`, since `--branch` resolves only branches and tags. Verified the SHA carries the classification, timeout, resumable-scan and stall-tracking behavior, and that `tools/rekor-monitor` at that commit is byte-identical to current main, so the documented behavior matches the pin exactly.

docs/integrator/supply-chain-verification.md
lockwobr: Correct on both. Fixed in d3fc34ee. Confirmed the AND at identity.go:666. Both examples now use a coherent SAN/issuer pair read off one real certificate, and the v1 `issuers` value is a private IdP issuer rather than the Fulcio URL, with the prose naming the OIDC issuer extension so the distinction is explicit.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed, including that a note cannot fix it. Fixed in d3fc34ee along the lines you described: bootstrap is now an explicit `workflow_dispatch` input, a scheduled run that finds no checkpoint fails with an `::error::` rather than exiting clean into a fresh baseline, and the branch filter uses `github.event.repository.default_branch` instead of a hard-coded `main`. The retention-expiry path lands in the same fail-closed branch.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed, and the section now says so. Added in d3fc34ee, verified against main.go:276-287 ("alert and hold. Do NOT persist this chunk's progress, so it is re-detected every run until a maintainer triages it") and the tag-based suppression in monitor.go. The framing is now yours: turnkey for identities that sign rarely enough to triage each hit, and for tag-bearing release identities via `--known-tags-file`; an actively signing generic identity needs an acknowledgment mechanism the tool does not have yet. No checkpoint hand-editing is suggested.

docs/integrator/supply-chain-verification.md
lockwobr: Correct on both counts, and both are now coverage gaps rather than recommendations. Fixed in d3fc34ee. The public-good v1 case cites this repository's own rationale (rekor-monitor.yaml: the v1 identity scan runs ~50x slower than the log grows). The private v1 case is verified in the upstream module: `GetLogVerifier` walks `trustedRoot.RekorLogs()` and errors with "couldn't find matching log instance with keyID", and the reusable workflow exposes no input for private trust material. As you suggested, the config-format content and the KMS `fingerprints` block are kept, since that part genuinely is what the upstream workflow illustrates.

docs/integrator/supply-chain-verification.md
lockwobr: Confirmed and added as a third gap in d3fc34ee. Verified `ConfigureTrustedCAs("", "", trustedRoot)` in monitor.go with hardcoded empty paths and no CLI override, and that upstream's v2 identity search writes a stderr note and `continue`s on a chain that does not validate, so it is neither a match nor a failed entry. The "Private Fulcio alone" table row is reconciled against the new gap so the two no longer disagree.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed. Split by signing mode in d3fc34ee. Containment now branches interactive keyless, CI workload identity (no OIDC session to rotate; treat the repository as compromised, including workflow permissions, tokens and environments), and KMS. Re-verification branches keyless bundle (`--require-creator`), KMS bundle (`--key`), evidence (`aicr evidence verify --expected-issuer / --expected-identity-regexp`), and private trust root (`--trust-root`). Every flag was checked against pkg/cli before writing it.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed, and this reverses an earlier position of mine on this PR. I had declined to put SHAs in the examples on the grounds that doc SHAs rot and readers copy stale ones. That reasoning does not survive your point: the argument here is privilege, not readability. The copied block runs mutable third-party code in a job that can mint an OIDC token as the reader's repository, which our own production workflow deliberately never grants. Fixed in d3fc34ee, pinned to `aa97a44631e9ae0f2b262a77ee5ea8edbb11e26c`, the exact commit `go.mod` already vendors, so the doc and the module agree.

On dropping `id-token: write`: I checked, and upstream requires it. Its README shows the grant in all three example workflows with the comment "Needed to detect the current reusable repository and ref", and `base_monitoring.yml` has a `detect-workflow` job that runs `slsa-github-generator`'s `detect-workflow-js`, which mints the token. A caller cannot downgrade it without breaking that job. So the grant stays, and I have made it the stated reason the SHA pin is mandatory rather than advisory.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed. Fixed in d3fc34ee: top level is now `permissions: {}`, matching the v2 example's fail-closed default, with the job block unchanged since it already grants exactly what it needs. Worth noting for the record that upstream's own README uses `read-all` here, so this is us hardening past the upstream example rather than matching it.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed, and fixed in d3fc34ee: the flag-bearing rows are now explicitly scoped to `aicr bundle`, with a paragraph stating that the `validate` and `evidence` paths are keyless-only and always write to the default Rekor v2, per the comment in `pkg/cli/validate_evidence.go`.

One refinement to the finding, in case it matters to you: `--fulcio-url`, `--rekor-url` and `--signing-config` are not registered on `aicr bundle` alone. They also exist on `aicr recipe sign-catalog` (pkg/cli/recipe_sign_catalog.go). That command is `Hidden: true` and documented as CI-only for the goreleaser after-hook, so I scoped the text to the user-facing surface rather than enumerating it. `--signing-key` and `--tlog-upload` really are bundle-only. Happy to mention sign-catalog explicitly if you would rather the text be exhaustive than simple.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed, and the 60-day auto-disable point is the one I would not have thought of. Added in d3fc34ee. It is implemented as `workflow_dispatch:` carrying the `bootstrap` input rather than a bare `{}`, because the fail-closed checkpoint handling needs that input, but it satisfies all three of your cases: seeding, provability on a branch, and re-arming.

docs/integrator/supply-chain-verification.md
lockwobr: Agreed, the rationale was the part that was wrong. Reworded in d3fc34ee to say the value must outlast the longest gap between successful runs, with the ceiling being the repository's artifact retention maximum. One note: I phrased the 90-day figure as public-repository-specific, since private and enterprise repositories allow up to 400. That is GitHub platform documentation rather than something verifiable in this repo.

docs/integrator/supply-chain-verification.md
lockwobr: Taken as the note rather than the reimplementation, since as you say it already fails closed. Added in d3fc34ee: a one-line pointer that production wraps this fetch in `.github/scripts/gh-api-retry.sh`, which retries 5xx/429 and publishes via `OUTFILE.part` only on success. The example itself stays plain so the security-relevant logic remains legible.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,418 @@ aicr trust update
...
+([`.github/scripts/gh-api-retry.sh`](https://github.com/NVIDIA/aicr/blob/main/.github/scripts/gh-api-retry.sh)),
+which recovers 5xx and 429 responses and publishes each response through
+`OUTFILE.part` so a failed attempt's error body can never be mistaken for a
+checkpoint. The example above is left plain because it already fails closed.
+
+### Monitoring Rekor v1 and a private Rekor
+
+If you opted signing out to Rekor v1 with `aicr bundle --rekor-url`, be aware
+that neither of the two obvious cases has a turnkey monitor today. Both are
+coverage gaps, not recipes.
```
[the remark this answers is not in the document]
lockwobr: Correct, and it was the same contradiction the human reviewers caught elsewhere in this section: prose declaring something unsupported while the copyable block below still recommended it. Fixed in 7f05dc28.

The `url:` line is gone. The block now carries a comment stating that it defaults to the public-good Rekor v1 and that a private log cannot be reached through this workflow at all, pointing at the direct-binary path with `--tuf-repository` / `--tuf-root-path` instead. I also added a sentence before the block labelling it as illustrating the `monitoredValues` shape rather than being a turnkey recipe, since the public-good v1 case does not give sustained coverage either.

I did not replace it with a direct-binary example. Doing that well needs a worked TUF-repository setup that serves the private log's key, which is a larger piece than this section should carry; the gap paragraph above now names the flags and the approach so a reader knows where to go.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,424 @@ aicr trust update
...
+          echo "Re-arm deliberately with a workflow_dispatch run and bootstrap=true."
+          exit 1
+
+      - name: Scan for our signing identity
+        run: |
+          GOFLAGS="-mod=vendor" go run ./tools/rekor-monitor \
+            --file checkpoint_v2.txt \
+            --restore-zip checkpoint.zip \
+            --cert-subject "${CERT_SUBJECT}" \
+            --cert-issuer "${CERT_ISSUER}"
```
yuanchen8911: This is blocking, and applies equally to the shell example at lines 558-562: both examples watch a `refs/tags/.*` release identity but never pass `--known-tags-file`. Per the limitation stated just above, on an identity match the tool holds the cursor and returns without persisting the chunk's progress, so the reader's first legitimate release after baseline leaves the monitor re-alerting every run and — the part that matters — entries in later chunks remain unscanned until the finding is triaged. The only triage path is the flag the examples omit, so a verbatim copy suspends the monitor's core function on the first ordinary use of the watched identity. The prose says a tag-bearing release identity is turnkey via `--known-tags-file`; the copyable blocks for exactly that identity should wire it.

Two constraints on the fix:

- The allowlist must come from an authoritative record of legitimately signed releases, not raw repository tags — an attacker-pushed or never-signed tag in `git tag --list` would be pre-suppressed. Production's own step is the copyable pattern: it derives `known-tags.txt` from the release workflow's completed run history (`.github/workflows/rekor-monitor.yaml`, the `runs.json` to `known-tags.txt` step), precisely because repo tags are deleted for ephemeral RCs while run history persists (#1902).

- Worth carrying over production's caveat too: an attacker re-signing an existing release tag is suppressed by this mechanism; a per-tag entry-count/provenance check is the tracked follow-up.

For the shell example, a comment showing the run-history derivation (or pointing at production's step) is enough; for the workflow example, add the fetch step and the flag.

docs/integrator/supply-chain-verification.md
```diff
@@ -463,6 +463,424 @@ aicr trust update
...
+and issuer are AND-ed, so an incoherent pair matches nothing at all and the
+monitor reports clean forever while watching an identity that cannot exist.
+
+`--cert-subject` and `--cert-issuer` are regexes matched against the
+certificate's SAN and OIDC issuer extension; anchor them, or a lookalike
+identity matches. `--file` is the cursor: the first run baselines at the current
+tree head and scans nothing, and every later run scans only the window added
+since. The tool writes two companions alongside it, `<file>.scan` and
+`<file>.stall`, holding partial progress and catch-up trend so that a large
+backlog is scanned across several bounded runs; all three must survive between
```
yuanchen8911: Small accuracy fix: "all three must survive between runs or the scan restarts" is true of `<file>` and `<file>.scan`, but not `.stall`. A missing `.scan` makes `readProgress()` return 0 and the window rescans from the start; a missing `.stall` only resets the catch-up convergence history feeding the `degraded` classification — scan progress is untouched (`checkpoint.go`: `readProgress` vs `readScanTrend`). Suggest splitting the clause: the checkpoint and `.scan` must survive or the scan restarts; `.stall` only preserves the stall-detection history.

docs/user/artifact-verification.md
```diff
@@ -353,6 +353,16 @@ case "$(jq '.exit' result.json)" in
...
 ```
 
+## Monitoring Your Signing Identity
+
+Verification is only half the picture. It proves that an artifact you hold came
+from the identity you expect, but it cannot tell you that somebody else signed
+something *as you*: an entry in the transparency log under your signing identity
+that you did not produce means the identity was used without you. Keyless signing
+leaves no local trace of a signing event, so the log is the only place that
+misuse shows up. For what each signing mode records and how to watch for it, see
```
yuanchen8911, line 363: "Means" overstates what the entry proves: an identity legitimately shared with CI automation or another approved signer produces entries you did not personally make, and the integrator guide's own "rule yourself out first" step anticipates exactly that. Suggest "may indicate the identity was used without you" (or similar), keeping the point that the log is the only durable record. The same absolute wording appears in `docs/integrator/supply-chain-verification.md` line 471 ("means your identity was used without you") — please fix both in the same pass.

yuanchen8911 (changes requested):
All previous findings were fixed and the threads are resolved. There's a blocking issue remaining, please take a look: both examples watch a `refs/tags/.*` release identity but never pass `--known-tags-file` (inline comment below). Two small same-pass fixes are also inline, and one item to defer.

On your `recipe sign-catalog` question: keep the text scoped to the user-facing surface as you have it. Documenting a `Hidden: true` CI-only command in an integrator doc adds noise without helping the reader.

One follow-up rather than a change here: neither the doc examples nor the production workflow set `persist-credentials: false` on `actions/checkout`. The example faithfully mirrors production, so I'd rather harden both together in a focused follow-up than have the doc diverge. Exposure is low (the persisted token is scoped to `contents: read` + `actions: read`).

docs/integrator/supply-chain-verification.md
lockwobr: Agreed, and this is the same failure mode this PR keeps hitting: the prose states a limitation correctly, then the copyable block below it does not account for it. The section says a tag-bearing release identity is turnkey *via* `--known-tags-file`, and then neither example passed it. Fixed in 247ceed1.

Your second constraint is the part I would have got wrong on my own, so I took production verbatim rather than improvising. The workflow example now carries the `Fetch signed release tags` step from `.github/workflows/rekor-monitor.yaml`, including its jq unchanged:

```
jq -rs '[.[].workflow_runs[] | select(.head_branch != null) | .head_branch] | unique | .[]' runs.json > known-tags.txt
```

with `RELEASE_WORKFLOW_FILE` in `env` and production's note to change it and `CERT_SUBJECT` together, plus the rationale that the allowlist must come from an authoritative record rather than `git tag --list` and that run history persists after ephemeral RC tags are deleted (#1902). The one substitution is plain `gh api --paginate` in place of `.github/scripts/gh-api-retry.sh`, since a reader's repo has no such script; the `jq -rs` slurp is kept because `--paginate` emits one object per page. The shell example gets the flag plus a comment stating the derivation and the never-from-`git tag --list` constraint.

Production's caveat is carried as its own paragraph: an attacker re-signing an existing release tag is suppressed because that tag is legitimately allowlisted, and the allowlist keys on a completed release run rather than proof the signing step succeeded. Cited #1887 as the tracked follow-up, which is what `docs/contributor/maintaining.md` names for the same gap.

docs/integrator/supply-chain-verification.md
lockwobr: Correct, thanks. Fixed in 247ceed1. Verified before rewriting: `progressPath()` is `.scan` (checkpoint.go:112) and `readProgress` returns 0 when absent, so the window rescans from the start; `stallPath()` is `.stall` (checkpoint.go:161) and `readScanTrend` returns the zero value, feeding only bestRemaining/stall/passes for the `degraded` classification. The clause is now split so `.scan` carries the must-survive requirement and `.stall` is described as costing stall detection rather than scan progress.

docs/user/artifact-verification.md
lockwobr: Agreed, and it was inconsistent with the section's own "rule yourself out first" step, which exists precisely because a shared or automated identity produces entries you did not personally make. Fixed in both files in 247ceed1: `may indicate that the identity was used without you` in the integrator guide and the matching change in `docs/user/artifact-verification.md`, keeping the point that the log is the only durable record.

yuanchen8911 (approved):
All three findings verified fixed at fb2fede78 — both monitor examples now carry --known-tags-file with the release-run-derived allowlist guidance, the .scan/.stall persistence semantics are exact, and the identity-entry claim is correctly softened. /lgtm

PR #1960, opened 2026-07-29
fix(bom): extract nested operator images

yuanchen8911:
## Summary

Extend BOM and mirror image discovery to recognize operator-style `image: {name, repository, tag}` mappings and KAI’s `scalingPodImage` descriptor. Fail closed on invalid recognized descriptors so BOM and mirror output cannot succeed while omitting a runtime-defaulted image.

## Motivation / Context

KAI Scheduler declares operator-managed operand images as mapping-valued descriptors, while the shared extractor recognized only scalar `image:` values. The committed BOM therefore listed only the operator and CRD upgrader, and air-gap outputs omitted binder, scheduler, podgrouper, admission, scalingpod, and the other rendered operands.

The KAI operator defaults a missing binder image name at runtime. A `binder.image.name=null` override previously caused AICR to omit binder silently even though the deployment still pulled it.

This is separate from the incomplete chart-reference report in #1954. It supersedes the scheduler downgrade proposed by draft #1958, which would not fix that integration. Unsupported descriptor shapes discovered during review are tracked separately in #1961.

Fixes: N/A
Related: #1954, #1961

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: BOM and mirror image discovery

## Implementation Notes

- Recognize allowlisted mapping-valued `image` and `scalingPodImage` nodes using `name`, `repository`, and `tag` fields.
- Preserve repository-only descriptors only when `name` is absent. A present null, empty, or non-scalar recognized member returns a structured invalid-request error.
- Make invalid structured descriptors fatal for `mirror list` and the BOM generator, including non-strict scheduled generation, instead of publishing a known-incomplete image set.
- Preserve cancellation while scanning manifest lists and between individual reads.
- Reuse existing image-reference combination and validation paths so scalar and mapping descriptors share filtering and deduplication behavior.
- Reject a present `registry` or `digest` member in a recognized descriptor instead of silently dropping it, so a partially-combined reference cannot resolve against the wrong registry or lose a digest pin. #1961 remains open for actually combining these members.
- Treat a null or empty `tag` (the Helm appVersion idiom) like an absent tag instead of failing the survey; a non-scalar tag still fails closed.
- Trim a trailing repository slash in the name-absent fallback, matching the name-present path.
- Blank out directive-only Helm lines instead of deleting them so descriptor-error line numbers index the original file.
- Use an override-aware KAI test renderer so the autoscaling and null-name assertions are causally tied to their `--set` values.
- Regenerate `docs/user/container-images.md`; the KAI inventory expands from 2 to 11 images and the repository-wide unique count from 87 to 96.

## Testing

```bash
make bom-docs
make bom-check
GOFLAGS="-mod=vendor" go test ./pkg/bom ./pkg/mirror ./tools/bom
GOFLAGS="-mod=vendor" go test -race ./pkg/bom/... ./pkg/mirror/... ./tools/bom/...
golangci-lint run -c .golangci.yaml ./pkg/bom/... ./pkg/mirror/... ./tools/bom/...
GOFLAGS="-mod=vendor" go test -coverprofile=<current> ./pkg/bom/...
GOFLAGS="-mod=vendor" go test -coverprofile=<current> ./pkg/mirror/...
GOFLAGS="-mod=vendor" go test -coverprofile=<current> ./tools/bom/...
GOFLAGS="-mod=vendor" go test -coverprofile=<baseline> ./pkg/bom/...
GOFLAGS="-mod=vendor" go test -coverprofile=<baseline> ./pkg/mirror/...
GOFLAGS="-mod=vendor" go test -coverprofile=<baseline> ./tools/bom/...
go run ./cmd/aicr mirror list --service eks --accelerator h100 --intent training --os ubuntu --set kaischeduler:global.clusterAutoscaling=true --format json
go run ./cmd/aicr mirror list --service eks --accelerator h100 --intent training --os ubuntu --set kaischeduler:binder.image.name=null --format json
```

All focused gates passed. Coverage changed as follows:

- `pkg/bom`: 95.8% → 96.1%
- `pkg/mirror`: 80.7% → 81.6% (rebased baseline; #1933 rewrote the mirror override machinery, so `TestSetNestedValue`/`splitPath` were dropped with the removed helper and the manifest-only test refs use Helm type per the new component-ref coherence rules)
- `tools/bom`: 83.0% → 83.1%

The autoscaling reproduction includes `scalingpod:v0.14.1`. The null-name reproduction exits 2, identifies `kai-scheduler` and the invalid `name` field, and writes no mirror output.

Full `make qualify` was run locally: test-coverage (with `-race`, 80.8% vs the 80% floor), lint, tuning-check, e2e, scan, and license-check all passed. The only local failures were `tests/releasepolicy` script-deadline timeouts that reproduce only under full-suite parallel load on the dev machine, pass in isolation on both this branch and `origin/main`, and touch a tree with zero diff from `main` that imports none of this PR's packages.

## Risk Assessment

**Rollout notes:** Regenerate mirror manifests or bundles to include the newly discovered KAI operand images. Invalid mapping-valued image overrides now fail instead of producing an incomplete mirror list; no configuration migration is required for valid descriptors.

## Checklist

### DIFF
--- a/pkg/bom/extract.go
+++ b/pkg/bom/extract.go
@@ -31,16 +31,23 @@ import (
 // YAML parsing. Files under recipes/components/*/manifests/ are sometimes
 // Helm-template-shaped (the bundler processes them as chart templates), so
 // raw YAML parsing would fail on the bare directives.
-const helmTemplatePlaceholder = "_aicr_helm_template_"
+const (
+	helmTemplatePlaceholder = "_aicr_helm_template_"
+	imageRepositoryKey      = "repository"
+	imageTagKey             = "tag"
+)
 
 var helmTemplateRE = regexp.MustCompile(`\{\{[^{}]*\}\}`)
 
 // stripHelmTemplates pre-processes a YAML document so the parser doesn't
 // choke on Go-template directives. Two passes:
-//  1. Drop any line whose non-whitespace content consists entirely of one or
-//     more Helm directives (e.g., `  {{- if foo }}`, `  {{- end }}`,
+//  1. Blank out any line whose non-whitespace content consists entirely of
+//     one or more Helm directives (e.g., `  {{- if foo }}`, `  {{- end }}`,
 //     `  {{- toYaml . | nindent 4 }}`). These are control-flow scaffolding
-//     that produces no YAML node when rendered.
+//     that produces no YAML node when rendered. The line is kept (empty)
+//     rather than deleted so yaml.Node positions — and therefore the
+//     line/column reported by descriptor errors — still index the original
+//     file.
 //  2. On surviving lines, replace inline directives with a placeholder so a
 //     value like `key: {{ .Values.x }}` becomes `key: _aicr_helm_template_`
 //     instead of breaking YAML parsing. The placeholder is filtered out by
@@ -51,6 +58,7 @@ func stripHelmTemplates(data []byte) []byte {
 	for _, l := range lines {
 		stripped := helmTemplateRE.ReplaceAll(l, nil)
 		if len(bytes.TrimSpace(stripped)) == 0 && bytes.Contains(l, []byte("{{")) {
+			out = append(out, nil)
 			continue
 		}
 		out = append(out, helmTemplateRE.ReplaceAll(l, []byte(helmTemplatePlaceholder)))
@@ -59,8 +67,15 @@ func stripHelmTemplates(data []byte) []byte {
 }
 
 // ExtractImagesFromYAML walks every YAML document in data and returns the
-// sorted, de-duplicated set of `image:` scalar values. It skips empty values,
-// `null`, and any value still containing an unrendered Go template directive.
+// sorted, de-duplicated set of container image references. It recognizes both
+// scalar `image:` values and operator-style `image: {name, repository, tag}`
+// mappings. Empty or null scalar `image:` values and values still containing an
+// unrendered Go template directive are skipped. A recognized structured
+// descriptor returns an invalid-request error when a present name or
+// repository field is null, empty, or non-scalar, when a tag field is
+// non-scalar, or when a registry or digest member is present (dropping either
+// would emit a wrong or un-pinned reference). A null or empty tag follows the
+// Helm appVersion idiom and is treated as absent.
 //
 // Helm template directives ({{ ... }}) are replaced with a placeholder before
 // parsing, so files mixing YAML with Helm templates (those under
@@ -105,12 +120,19 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
 		// `version` siblings carry the registry and tag. A sibling
 		// `containerSHA` (Skyhook Package CRD; ghcr.io/nvidia/nodewright)
 		// supplies the OCI digest and is folded in as `@<sha>`.
-		var imgScalar, repoScalar, verScalar, shaScalar string
+		var (
+			imgScalar, repoScalar, verScalar, shaScalar string
+			imgMappings                                 []*yaml.Node
+		)
 		for i := 0; i+1 < len(n.Content); i += 2 {
 			k, v := n.Content[i], n.Content[i+1]
 			target := v
 			if v.Kind == yaml.AliasNode && v.Alias != nil {
 				target = v.Alias
+			}
+			if isStructuredImageKey(k.Value) && target.Kind == yaml.MappingNode {
+				imgMappings = append(imgMappings, target)
+				continue
 			}
 			if target.Kind != yaml.ScalarNode {
 				continue
@@ -118,12 +140,21 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
 			switch k.Value {
 			case "image":
 				imgScalar = strings.TrimSpace(target.Value)
-			case "repository":
+			case imageRepositoryKey:
 				repoScalar = strings.TrimSpace(target.Value)
 			case "version":
 				verScalar = strings.TrimSpace(target.Value)
 			case "containerSHA":
 				shaScalar = strings.TrimSpace(target.Value)
+			}
+		}
+		for _, imgMapping := range imgMappings {
+			mapped, err := imageReferenceFromMapping(imgMapping)
+			if err != nil {
+				return err
+			}
+			if isLikelyImage(mapped) {
+				seen[mapped] = struct{}{}
 			}
 		}
 		if imgScalar != "" {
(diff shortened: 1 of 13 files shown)


pkg/bom/extract.go
```diff
@@ -105,27 +110,39 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
...
 				verScalar = strings.TrimSpace(target.Value)
 			case "containerSHA":
 				shaScalar = strings.TrimSpace(target.Value)
 			}
 		}
+		for _, imgMapping := range imgMappings {
+			if mapped := imageReferenceFromMapping(imgMapping); isLikelyImage(mapped) {
+				seen[mapped] = struct{}{}
+			}
+		}
```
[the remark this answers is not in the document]
yuanchen8911: No code change here. The extraction pass is synchronous and in-memory over bytes already returned by Helm; it performs no I/O. `Discover` already propagates cancellation through the Helm renderer and checks `gctx.Err()` before manifest scans. The new mapping branch adds only a scan of the rendered descriptor mapping and does not introduce a new cancellation boundary.

The reachability check also fails: no supported catalog render produces traversal latency meaningful relative to Helm rendering. Changing the exported extractor API and every BOM, mirror, and test caller would be unrelated hardening rather than a fix required by #1960, so I am keeping it out of this PR.

pkg/bom/extract.go
```diff
@@ -160,6 +177,72 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
...
+	}
+	if !namePresent {
+		name = repository
+		repository = ""
+	}
+	if name == "" {
+		return ""
+	}
+	return combineCRDTriplet(name, repository, tag)
```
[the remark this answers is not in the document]
yuanchen8911: No code change here. `repository` and `tag` are not universally required by the extractor contract: a complete reference can reside in `name`, the repository-only form can itself carry a complete reference, and registry-qualified untagged references are accepted by the same `isLikelyImage` path as scalar values. Requiring both fields would reject valid forms.

The reachability check also fails for the reported examples. The current catalog render does not produce a structured descriptor missing `repository` or `tag`; KAI v0.14.1 renders both. This PR already rejects present null, empty, or non-scalar members, and documents separate `registry`/`digest` mapping shapes as unsupported. The hypothetical missing-field forms therefore do not justify narrowing the generic extractor in #1960.

pkg/bom/extract.go
```diff
@@ -160,6 +183,104 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
...
+	}
+	if !namePresent {
+		name = repository
+		repository = ""
+	}
+	if name == "" {
+		return "", nil
+	}
+	return combineCRDTriplet(name, repository, tag), nil
+}
```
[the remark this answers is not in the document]
yuanchen8911: Thanks. I am leaving this as-is. The first switch is the allowlist, scalar validation is intentionally shared once, and the second switch only assigns the validated value. Collapsing them would duplicate validation or add pointer/helper indirection without changing behavior. Since this is cosmetic and #1960 is scoped to the KAI/BOM defect, that refactor belongs outside this PR.

pkg/mirror/discover_test.go
```diff
@@ -283,6 +288,166 @@ spec:
 	}
 }
 
+type kaiConfigRenderer struct {
+	inputs []helm.ChartInput
+}
+
+func (r *kaiConfigRenderer) Render(_ context.Context, input helm.ChartInput) ([]byte, error) {
+	r.inputs = append(r.inputs, input)
```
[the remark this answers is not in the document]
yuanchen8911: Thanks. I verified that both tests build recipes with a single KAI component, so Render has exactly one caller in these cases. Discover waits for errgroup completion before the test reads inputs, and the focused race suite passes. A mutex would guard only a hypothetical future multi-component use of this test fake, so I am leaving it out of the focused #1960 change. If the fake is later reused concurrently, that change should add synchronization with the new caller.

tools/bom/main.go
```diff
@@ -287,6 +337,13 @@ func surveyComponent(ctx context.Context, repoRoot string, c component, r helm.R
 		if len(out) > 0 {
 			imgs, parseErr := bom.ExtractImagesFromYAML(out)
 			if parseErr != nil {
+				if bom.IsInvalidStructuredImageDescriptor(parseErr) {
+					return res, wrapInvalidDescriptorSurveyError(
+						parseErr,
+						c.Name,
+						"",
+					)
+				}
```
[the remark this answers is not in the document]
yuanchen8911: Thanks. I am leaving this out of #1960. This affects only diagnostic context when a version variant itself renders an invalid structured descriptor. KAI has no derived version variant, current supported variant renders are valid, and the generator still fails closed without the version in the message. It does not affect the KAI/BOM completeness defect addressed here; broader variant diagnostics should be handled separately.

pkg/bom/extract.go
```diff
@@ -160,6 +183,104 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
...
+// scalar. Only those three fields are combined. Other fields are ignored, so
+// descriptors that rely on separate registry or digest fields are unsupported
+// and can yield an incomplete or incorrectly resolved reference.
+func imageReferenceFromMapping(n *yaml.Node) (string, error) {
+	var name, repository, tag string
+	var namePresent bool
+	for i := 0; i+1 < len(n.Content); i += 2 {
+		key, value := n.Content[i], n.Content[i+1]
+		switch key.Value {
+		case "name", imageRepositoryKey, "tag":
```
ArangoGutierrez: The allowlist drops every other member, so `{registry: nvcr.io, repository: nvidia/foo, tag: v1}` yields `nvidia/foo:v1`, which `ParseImageRef` resolves to `docker.io` — an air-gap mirror seeds from Docker Hub instead of NGC, and a sibling `digest:` is dropped the same way so a pinned reference becomes floating. Before this PR both shapes emitted nothing, so this trades a visible gap for a silently wrong entry; failing closed on a present `registry` or `digest` member would match the null-member handling just below.

pkg/bom/extract.go
```diff
@@ -160,6 +183,104 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
...
+			namePresent = true
+			name = scalar
+		case imageRepositoryKey:
+			repository = scalar
+		case "tag":
+			tag = scalar
+		}
+	}
+	if !namePresent {
+		name = repository
```
ArangoGutierrez: With `name` absent, `repository` moves into `name` and `repository` is cleared, so the `strings.TrimRight(repository, "/")` inside `combineCRDTriplet` never runs — `{repository: ghcr.io/kai-scheduler/kai-scheduler/, tag: v0.14.1}` produces `ghcr.io/kai-scheduler/kai-scheduler/:v0.14.1`. `ParseImageRef` keeps that trailing slash in `Repository`, so the malformed reference reaches the mirror list and the CycloneDX PURL; trimming before the fallback assignment fixes it.

ArangoGutierrez (commented):
Four issues, all in the new structured-descriptor path. The first two are blockers: each emits a wrong image reference where the pre-PR code emitted nothing, and that reference lands in both the air-gap mirror seed and the CycloneDX BOM. Checked by running the extractor against the PR head and against the merge base.

1. The allowlist drops every other member, so `{registry: nvcr.io, repository: nvidia/foo, tag: v1}` yields `nvidia/foo:v1`, which `ParseImageRef` resolves to `docker.io` — an air-gap mirror seeds from Docker Hub instead of NGC, and a sibling `digest:` is dropped the same way so a pinned reference becomes floating. Before this PR both shapes emitted nothing, so this trades a visible gap for a silently wrong entry; failing closed on a present `registry` or `digest` member would match the null-member handling just below. (pkg/bom/extract.go:213)
2. With `name` absent, `repository` moves into `name` and `repository` is cleared, so the `strings.TrimRight(repository, "/")` inside `combineCRDTriplet` never runs — `{repository: ghcr.io/kai-scheduler/kai-scheduler/, tag: v0.14.1}` produces `ghcr.io/kai-scheduler/kai-scheduler/:v0.14.1`. `ParseImageRef` keeps that trailing slash in `Repository`, so the malformed reference reaches the mirror list and the CycloneDX PURL; trimming before the fallback assignment fixes it. (pkg/bom/extract.go:241)
3. An empty member is fatal but an absent one is not, so `{repository: ghcr.io/foo/bar, tag: ""}` aborts the whole run while `{repository: ghcr.io/foo/bar}` is accepted and emits an untagged reference that resolves to `:latest`. `tag: ""` is the standard Helm idiom for "use appVersion" and there is no override flag, so a single chart bump that renders it turns `make bom-docs` and `aicr mirror list` into hard failures — treating empty and absent alike keeps the fail-closed intent without that cliff. (pkg/bom/extract.go:258)
4. `stripHelmTemplates` deletes directive-only lines before parsing, so `value.Line` is a position in the stripped buffer rather than in the file. A manifest with two `{{- ... }}` lines above a bad `tag` reports line 6 for a descriptor that sits on line 8, which points the operator at the wrong place in exactly the templated manifests under `recipes/components/*/manifests/` that this path exists to survey. (pkg/bom/extract.go:225)

pkg/bom/extract.go
```diff
@@ -160,6 +183,104 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
...
+
+func nonNullImageMappingScalar(n *yaml.Node) (string, bool) {
+	if n.Kind == yaml.AliasNode && n.Alias != nil {
+		n = n.Alias
+	}
+	if n.Kind != yaml.ScalarNode || n.Tag == "!!null" {
+		return "", false
+	}
+	value := strings.TrimSpace(n.Value)
+	return value, value != ""
```
ArangoGutierrez, line 290: An empty member is fatal but an absent one is not, so `{repository: ghcr.io/foo/bar, tag: ""}` aborts the whole run while `{repository: ghcr.io/foo/bar}` is accepted and emits an untagged reference that resolves to `:latest`. `tag: ""` is the standard Helm idiom for "use appVersion" and there is no override flag, so a single chart bump that renders it turns `make bom-docs` and `aicr mirror list` into hard failures — treating empty and absent alike keeps the fail-closed intent without that cliff.

pkg/bom/extract.go
```diff
@@ -160,6 +183,104 @@ func walkForImages(n *yaml.Node, seen map[string]struct{}) error {
...
+		}
+
+		scalar, ok := nonNullImageMappingScalar(value)
+		if !ok {
+			return "", errors.Wrap(
+				errors.ErrCodeInvalidRequest,
+				"invalid image descriptor member",
+				&invalidStructuredImageDescriptorError{
+					field:  key.Value,
+					line:   value.Line,
```
ArangoGutierrez, line 253: `stripHelmTemplates` deletes directive-only lines before parsing, so `value.Line` is a position in the stripped buffer rather than in the file. A manifest with two `{{- ... }}` lines above a bad `tag` reports line 6 for a descriptor that sits on line 8, which points the operator at the wrong place in exactly the templated manifests under `recipes/components/*/manifests/` that this path exists to survey.

pkg/bom/extract.go
yuanchen8911: Fixed in ef9d8eef3. A present registry or digest member in a recognized descriptor now fails closed with a structured descriptor error instead of emitting a partially-combined reference, so {registry, repository, tag} can no longer seed the mirror from docker.io and a sibling digest can no longer be silently dropped. Actually combining these members into the reference stays tracked in #1961. Added tests for both members.

pkg/bom/extract.go
yuanchen8911: Fixed in ef9d8eef3. The name-absent fallback now trims the trailing slash before repository becomes the name, matching what combineCRDTriplet already does on the name-present path, so {repository: ghcr.io/kai-scheduler/kai-scheduler/, tag: v0.14.1} yields ghcr.io/kai-scheduler/kai-scheduler:v0.14.1. Added a test covering this shape.

pkg/bom/extract.go
yuanchen8911: Fixed in ef9d8eef3. A null or empty tag is now treated like an absent tag — the Helm use-appVersion idiom — so tag: "" emits the same untagged reference as an omitted tag instead of failing the whole survey. A non-scalar tag still fails closed, and name/repository keep the strict null/empty rejection. Added tests for both tag: "" and a bare tag: null.

pkg/bom/extract.go
yuanchen8911: Fixed in ef9d8eef3. stripHelmTemplates now blanks directive-only lines out instead of deleting them, so yaml.Node positions — and the line/column reported in descriptor errors — index the original file. Added a test asserting the reported line matches the file position when directive-only lines precede the bad descriptor.

pkg/mirror/discover.go
```diff
@@ -269,25 +292,49 @@ func (l *Lister) Discover(ctx context.Context, rec *recipe.RecipeResult) (*Mirro
...
+				),
+				extractErr,
+				map[string]any{"component": compName, "manifest": mPath},
+			)
+		}
 		ci.Warnings = append(ci.Warnings,
 			fmt.Sprintf("manifest image extraction failed %s: %v", mPath, extractErr))
-		return acc
+		return acc, nil
 	}
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in fae165ec1. The non-fatal extraction-failure branch now emits the matching slog.Warn with component, path, and error before recording the warning, mirroring the read-failure branch, so the signal survives output formats that drop ComponentImages.Warnings.

njhensley (approved):
**Approving.** Reviewed via a multi-persona panel (correctness, supply-chain, operability, test-coverage) plus an adversarial meta-reviewer, then re-reviewed the delta after the revision.

The substantive change — recognizing operator-style `image: {name, repository, tag}` / `scalingPodImage` descriptors, expanding the KAI inventory 2 → 11 images (repo total 87 → 96) — is correct and well-tested, and the revision resolves every issue @ArangoGutierrez raised:

- **registry/digest members** now **fail closed** (`ErrCodeInvalidRequest`) instead of silently emitting a wrong-registry / un-pinned reference — consistent with the PR's own fail-closed principle, and tracked for real support in #1961.
- **empty/null `tag`** is now treated as absent (the Helm appVersion idiom) rather than aborting the whole `mirror list` / BOM run; `name`/`repository` present-empty stay fatal (correct).
- **name-absent + trailing-slash repository** now trims the slash, so no more `repo/:tag` / empty-name PURL.
- descriptor-error **line/column** now index the original file (`stripHelmTemplates` blanks directive-only lines instead of deleting them).

Each fix ships with a targeted test. `pkg/bom` and `tools/bom` pass at this head; BOM regenerates cleanly at 96 images (no current component trips the new fail-closed path). Verified the error wrap chain preserves `IsInvalidStructuredImageDescriptor` through `PropagateOrWrap`/`WrapWithContext`, so `mirror list` exits nonzero with an actionable component + manifest + field message.

Only residual is an optional nitpick — no test pins numeric/unquoted tag behavior (`tag: 1.28`); non-blocking. Nice work, and credit to @ArangoGutierrez for the sharp catches.

PR #1970, opened 2026-07-30
feat(slurm): add accounting ownership modes

kaynetu:
## Summary

Add explicit `disabled`, `customer-managed`, and `aicr-provided` Slurm accounting ownership modes. AICR can now provision and validate a recipe-managed MariaDB accounting stack, detect ownership conflicts from snapshot evidence, and verify persisted job records with `sacct`.

## Motivation / Context

Slurm accounting previously lacked an explicit ownership contract: generated recipes could not safely distinguish a customer database from an AICR-managed stack, and bundle-time overrides could produce incoherent deployments. This change makes ownership explicit while keeping recipe generation observational and enforcing installation safety only after effective bundle values are known.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: configuration, serialization, OpenAPI, component registry, and conformance checks

## Implementation Notes

- Adds a typed `configuration.slurm.accounting.mode` contract. Newly resolved Slurm recipes default to `disabled`; `customer-managed` enables Slurm accounting without installing a database; `aicr-provided` installs the full MariaDB stack. Configured recipes use `aicr.run/v1alpha3`.
- Keeps the accounting choice outside recipe matching criteria and preserves stable SDK entry points. New `WithOptions` variants support per-resolution accounting behavior while existing methods remain source-compatible.
- Integrates with ADR-015 profile recipes: profile-only, accounting-only, and combined recipes can share the strict `v1alpha3` artifact version without either validator assuming exclusive ownership of that version.
- Records snapshot-observed MariaDB Operator evidence in `metadata.mariaDBOperatorState`. `crs-detected` and `unknown` warn during resolution and block AICR-provided bundling; `api-detected` warns but proceeds; `absent` is silent; missing query-mode evidence warns and proceeds.
- Validates the complete effective Helm-value inventory before emitting any SDK bundle component. Protected ownership values cannot be changed through `--set` or `--set-json`, preventing partial or incoherent accounting bundles.
- Adds MariaDB Operator CRDs, MariaDB Operator, and the Slurm accounting database as ordered components. The initial database, user, and generated password are configured on the `MariaDB` CR, avoiding the unsupported `User.spec.passwordSecretKeyRef.generate` shape.
- Applies system node scheduling and storage-class overrides to MariaDB and SlurmDBD pods. MariaDB requests 6Gi, limits 8Gi, and carries the documented Slurm InnoDB tuning defaults.
- Extends deployment health checks for the CRDs, operator, database, generated user/database/grant, and Slurm accounting StatefulSet. The existing Slinky health check conditionally submits a uniquely named job and polls `sacct` for a completed persisted allocation when accounting is enabled.
- Adds ADR-016 and updates CLI, API, SDK, recipe, component, validator, deployment, and operational documentation.

## Testing

```bash
golangci-lint run -c .golangci.yaml ./pkg/recipe/... ./pkg/serializer/... ./pkg/config/... ./pkg/client/v1/... ./pkg/server/... ./pkg/cli/... ./pkg/bundler/... ./validators/conformance/...
go test -race ./pkg/recipe/... ./pkg/serializer/... ./pkg/config/... ./pkg/client/v1/... ./pkg/server/... ./pkg/cli/... ./pkg/bundler/... ./validators/conformance/...
make bom-docs
```

- `make qualify`: passed; project coverage 80.9% (80% threshold).
- Affected-package lint: passed with zero issues.
- Targeted race tests: passed. The first sandboxed run could not write Sigstore's user cache; the full unrestricted qualification run passed the same attestation tests.
- BOM regeneration: 37 components and 102 image references.
- Coverage versus `origin/main`: `pkg/bundler` 82.5% → 82.7%, `pkg/cli` 73.9% → 73.8%, `pkg/client/v1` 80.3% → 81.0%, `pkg/config` 93.0% → 92.9%, `pkg/recipe` 88.6% → 88.6%, `pkg/serializer` 75.8% → 75.6%, `pkg/server` 82.1% → 81.6%, and `validators/conformance` 52.7% → 54.0%. No package decreased by more than 0.5%.

## Risk Assessment

- [x] **High** — Breaking change, affects critical paths, or complex rollout

**Rollout notes:** Newly generated Slurm recipes carry typed configuration and therefore use `v1alpha3`; consumers that only accept `v1alpha2` must upgrade. The default mode remains operationally disabled, and customer-managed mode installs no MariaDB components. Before using AICR-provided mode, regenerate from a current snapshot when possible and resolve any blocking MariaDB CR ownership evidence. External `--data` Slurm compositions that neither inherit the updated Slurm leaves nor declare `mariadb-operator-crds`, `mariadb-operator`, and `slurm-accounting-mariadb` no longer resolve, including in `disabled` mode. A cluster left with a partially installed legacy `MariaDB` CR may require removing and recreating that CR because its initial username and database fields are immutable.

## Checklist


pkg/serializer/reader.go
```diff
@@ -548,6 +615,11 @@ func FromFileWithKubeconfigContext[T any](ctx context.Context, path, kubeconfig
 		slog.Error("reader is unexpectedly nil despite no error", "path", path, "format", fileFormat)
 		return nil, errors.New(errors.ErrCodeInternal, fmt.Sprintf("reader is nil for %q", path))
 	}
+	for _, opt := range opts {
+		if opt != nil {
+			opt(ser)
+		}
+	}
```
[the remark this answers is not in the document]
kaynetu: Addressed in 4a2167c9. The duplicate ReaderOption loop was removed from fromFileWithKubeconfigContext; options are now applied only by NewFileReaderWithContext. Serializer tests and make qualify pass.

pkg/serializer/reader.go
```diff
@@ -521,6 +577,17 @@ func ReadFileBytesWithKubeconfigContext(
...
 func FromFileWithKubeconfigContext[T any](ctx context.Context, path, kubeconfig string, opts ...ReaderOption) (*T, error) {
+	return fromFileWithKubeconfigContext[T](ctx, path, kubeconfig, opts...)
+}
+
+// FromFileWithKubeconfigContextOptions is the configurable variant of
+// FromFileWithKubeconfigContext. Reader options apply consistently to local,
+// HTTP, and ConfigMap-backed inputs.
+func FromFileWithKubeconfigContextOptions[T any](ctx context.Context, path, kubeconfig string, opts ...ReaderOption) (*T, error) {
+	return FromFileWithKubeconfigContext[T](ctx, path, kubeconfig, opts...)
+}
```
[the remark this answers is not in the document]
kaynetu: addressed

docs/user/container-images.md
```diff
@@ -19,8 +19,8 @@ A machine-readable **CycloneDX 1.6 JSON** companion to this page is produced by
 <!-- BEGIN AICR-BOM -->
 ## Summary
 
-- Components: **34**
-- Unique images: **96**
+- Components: **37**
+- Unique images: **97**
```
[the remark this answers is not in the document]
kaynetu: Verified by running `make bom-docs` on 4a2167c9; the generator produced no diff. The component table contains 94 base references, while the Version variants table contributes 8 more references, for 102 total references. Deduplicating those references yields the generated summary value of 97 unique images, so no hand edit is needed.

pkg/serializer/reader.go
```diff
@@ -371,6 +387,47 @@ func (r *Reader) Deserialize(v any) error {
...
+		}
+	case FormatTable:
+		return false, errors.New(errors.ErrCodeInvalidRequest,
+			"table format is not supported for apiVersion inspection")
+	default:
+		return false, errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("unsupported format for apiVersion inspection: %s", r.format))
+	}
+	return header.APIVersion == strictAPIVersion, nil
+}
```
kaynetu, line 429: verify that i have addressed this

docs/design/016-slurm-accounting-enablement.md
```diff
@@ -0,0 +1,545 @@
...
+2. accounting enabled against a customer-managed database, either in-cluster
+   or external; and
+3. accounting enabled against a database installation provided by AICR.
+
+Today accounting is a bundle-time chart override:
+
+```shell
+aicr bundle \
+  --recipe recipe.yaml \
+  --set slinkyslurm:accounting.enabled=true
```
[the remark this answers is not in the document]
kaynetu: No change needed here. `slinkyslurm` is an intentional registered CLI override alias under the `slinky-slurm` component's `valueOverrideKeys`, and it is used consistently by the user guide and demos. The ownership validator also maps registered aliases back to the canonical component, so these examples target the correct component.

recipes/components/slurm-accounting-mariadb/values.yaml
```diff
@@ -0,0 +1,48 @@
...
+  passwordSecretKeyRef:
+    name: mariadb-password
+    key: password
+    generate: true
+  cleanupPolicy: Skip
+  storage:
+    size: 20Gi
+  replicas: 1
+  galera:
+    enabled: false
```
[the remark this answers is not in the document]
kaynetu: We explicitly declared that airc is not responsible for operations of this cluster. Aicr just provides it

pkg/recipe/accounting.go
```diff
@@ -0,0 +1,314 @@
...
+
+const (
+	slinkySlurmComponentName            = "slinky-slurm"
+	mariaDBOperatorCRDsComponentName    = "mariadb-operator-crds"
+	mariaDBOperatorComponentName        = "mariadb-operator"
+	slurmAccountingMariaDBComponentName = "slurm-accounting-mariadb"
+	accountingEnabledOverrideKey        = "enabled"
+	componentEnabledOverrideKey         = "enabled"
+	componentInstallOverrideKey         = "install"
+)
```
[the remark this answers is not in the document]
kaynetu: No change intended. These constants deliberately distinguish two different paths that happen to end in the same key today: `accountingEnabledOverrideKey` is the nested `slinky-slurm.accounting.enabled` field, while `componentEnabledOverrideKey` is a component-level `enabled` gate used by `ComponentRef.IsEnabled`. Keeping the semantic names prevents the component gate from being confused with the accounting subfield.

pkg/client/v1/aicr.go
```diff
@@ -1222,6 +1295,50 @@ func (c *Client) BundleComponents(ctx context.Context, r *RecipeResult) ([]Compo
...
+		name   string
+		values map[string]any
+	}
+	resolved := make([]resolvedValues, len(result.Components))
+	group, groupCtx := errgroup.WithContext(ctx)
+	for index := range result.Components {
+		facade := result.Components[index]
+		if !strings.EqualFold(facade.Kind, "helm") {
+			continue
+		}
```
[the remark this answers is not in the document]
kaynetu: No change here. Each Helm read already receives `groupCtx`, so the first lookup failure cancels every in-flight read. `errgroup.Go` starts goroutines immediately rather than queueing work, and this short in-memory scheduling loop will normally finish before a worker reports an error; checking `groupCtx.Err()` in the loop would therefore add complexity without materially preventing I/O or changing correctness.

pkg/client/v1/types.go
```diff
@@ -244,6 +244,11 @@ type RecipeRequest struct {
 	// Empty applies the resolved declaration's mandatory default.
 	Profile string
 
+	// AccountingMode selects the Slurm accounting database ownership model.
+	// It is valid only when Platform is "slurm". Empty defaults to "disabled"
+	// for newly resolved Slurm recipes.
+	AccountingMode string
```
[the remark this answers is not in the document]
kaynetu: This is already handled in the current implementation: `Client.ResolveRecipe` appends `WithAccountingMode(req.AccountingMode)` only when `req.AccountingMode != ""` (`pkg/client/v1/aicr.go`). An omitted field therefore keeps the builder default, while explicit invalid values still go through `ParseAccountingMode`.

pkg/serializer/reader.go
```diff
@@ -371,6 +387,47 @@ func (r *Reader) Deserialize(v any) error {
...
```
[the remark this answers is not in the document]
kaynetu: No change for this PR. Production strict-version paths use `NewFileReaderWithContext`, which fully buffers and size-bounds the file, URL, or ConfigMap input before `resolveStrictMode` inspects it. Direct `NewReader(..., WithStrictAPIVersion(...))` currently has no production call site; introducing a context-bearing deserialize API for arbitrary live streams would be a broader serializer API change rather than an accounting-mode fix.

api/aicr/v1/server.yaml
```diff
@@ -2857,6 +2924,115 @@ components:
...
+              $ref: "#/components/schemas/RecipeResponseBase/properties/deploymentOrder"
+            constraints:
+              type: array
+              items:
+                type: object
+                required: [name, value]
+                propertyNames:
+                  enum: [name, value, severity, remediation, unit]
+            validation:
+              $ref: "#/components/schemas/RecipeResponseBase/properties/validation"
```
njhensley: **🟠 Major — Dangling $ref: ConfiguredRecipeResponse.validation points at a property RecipeResponseBase does not have**

`ConfiguredRecipeResponse.validation` uses `$ref: "#/components/schemas/RecipeResponseBase/properties/validation"`, but `RecipeResponseBase` has no `validation` property (only apiVersion, kind, metadata, criteria, configuration, componentRefs, deploymentOrder, constraints). The sibling `deploymentOrder` ref (line 3025) resolves; this one does not. `ProfileRecipeResponse` inlines its own `validation` schema (2887) instead of ref-ing it, which is why it doesn't hit this. `openapi_sync_test.go` walks the raw YAML by path and never resolves `$ref`s, so CI ships the broken reference green.

**Blast radius:** Any consumer that RESOLVES the spec breaks or silently drops the field: Fern docs rendering, client codegen, external request-validating gateways applying the /v2/bundle oneOf. The in-process Go server doesn't consume the spec, so runtime request handling is unaffected — which caps this below a blocker.

**Fix:** Inline the `validation` object schema in `ConfiguredRecipeResponse` exactly as `ProfileRecipeResponse` does (2887–2925), OR add a real `validation` property to `RecipeResponseBase` and have both wrappers `$ref` it.

pkg/recipe/accounting.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+	mode, err := ParseAccountingMode(string(*cfg.accountingMode))
+	if err != nil {
+		return nil, err
+	}
+	cfg.accountingMode = &mode
+	return cfg, nil
+}
+
+func applyBuildConfig(result *RecipeResult, cfg *buildConfig) error {
```
njhensley, line 198: **🟡 Minor — ADR-016 §8 mandates profile/accounting owned-path disjointness as a generation-time error — no such check exists**

ADR-016 §8 (docs/design/016-slurm-accounting-enablement.md:347): "When a Slurm recipe has both a selected profile and accounting configuration, their owned paths must be disjoint. Any overlap is a generation-time error; there is no precedence rule." `applyBuildConfig` runs after `SelectedProfile` is stamped and blindly `deepMergeMap`s the accounting gates onto `ref.Overrides` (accounting.go:229) with no scan of profile-owned vs accounting-owned paths. An overlapping path would be silently clobbered (order-dependent) rather than rejected.

**Blast radius:** Latent today — no shipped profile touches slinky-slurm/MariaDB paths, so the sets are disjoint by accident. Becomes a real correctness hole the moment a Slurm-family profile owns any accounting-adjacent path.

**Fix:** In `applyBuildConfig` (or a post-build validator), when `SelectedProfile != nil`, assert the accounting-owned path set and `SelectedProfile.OwnedPaths` are disjoint; return `ErrCodeInvalidRequest` on overlap. Add a table test. Reasonable as a tracked follow-up rather than a merge blocker.

pkg/bundler/validations/checks.go
```diff
@@ -1095,3 +1096,66 @@ func CheckDriverOwnershipCoherence(ctx context.Context, componentName string, re
...
+		return nil, nil
+	}
+	mode, configured := recipeResult.AccountingMode()
+	if !configured || mode != recipe.AccountingModeAICRProvided {
+		return nil, nil
+	}
+
+	state := recipeResult.Metadata.MariaDBOperatorState
+	switch state {
+	case "":
```
njhensley, line 1145: **🟡 Minor — Conflict-detection gate is skipped for criteria-only aicr-provided recipes (no snapshot → empty state → warn-only)**

`computeMariaDBOperatorState(nil)` returns "" (mariadb_operator_state.go:35). `aicr recipe --slurm-accounting-mode aicr-provided` with no `--snapshot` yields `Metadata.MariaDBOperatorState == ""`, whose `case "":` arm (checks.go:1125) returns a warning only and ALLOWS bundling. The crs-detected→block protection that exists to stop a competing MariaDB install is therefore skipped whenever there is no snapshot evidence.

**Blast radius:** A criteria-only aicr-provided bundle deployed against a cluster already running the MariaDB Operator installs a competing stack — the exact outcome crs-detected→block prevents. This is documented intended behavior (godoc: criteria-only/older-snapshot workflows stay compatible), and fails open by design, so it is a minor gap rather than a defect.

**Fix:** Confirm this is intended (it reads as deliberate). If kept, ensure docs/user/slinky-slurm-accounting.md states plainly that aicr-provided from a criteria-only recipe skips conflict detection, and consider surfacing the warning more prominently at bundle time.

pkg/bundler/bundler.go
```diff
@@ -364,6 +375,288 @@ func (b *DefaultBundler) Make(ctx context.Context, recipeResult *recipe.RecipeRe
...
+
+func validAccountingPort(value any) bool {
+	switch port := value.(type) {
+	case int:
+		return port >= 1 && port <= 65535
+	case int32:
+		return port >= 1 && port <= 65535
+	case int64:
+		return port >= 1 && port <= 65535
+	case float64:
```
njhensley, line 483: **🟡 Minor — No passing test for a whole-number float64 accounting port (JSON/--set-json path unguarded)**

`validAccountingPort` has a dedicated `case float64` accept branch (bundler.go:491, `float64(int64(port)) == port`) precisely because JSON/`--set-json` decoders deliver integers as float64. The tests cover int 3306 (pass), 3306.5 float64 (reject), and "3306" string (reject) — but no whole-number float64 (float64(3306)) expecting success. A regression that removed the float64 accept branch stays green while rejecting the port on every recipe POSTed as JSON or supplied via --set-json.

**Blast radius:** Every JSON-loaded or --set-json-supplied customer-managed/aicr-provided bundle would fail with a spurious port error; the unit suite stays green — the codebase's own documented int/float64 anti-pattern.

**Fix:** Add a table case with port = float64(3306) and wantErr: false (and ideally a --set-json typed-path variant).

pkg/bundler/validations/checks_test.go
```diff
@@ -1507,6 +1507,97 @@ func TestCheckDriverOwnershipCoherence(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			warnings, errs := CheckMariaDBOperatorOwnershipCoherence(
+				context.Background(), "mariadb-operator", tt.recipeResult, nil, nil)
+			if len(warnings) != tt.wantWarnings {
+				t.Fatalf("warnings = %d (%v), want %d", len(warnings), warnings, tt.wantWarnings)
+			}
+			if len(errs) != tt.wantErrors {
```
njhensley, line 1607: **🟡 Minor — Ownership-coherence test asserts error count + message but not the error CODE (409 vs 400 can flip unnoticed)**

The source deliberately returns `ErrCodeConflict` (HTTP 409) for CRsDetected/Unknown and `ErrCodeInvalidRequest` (HTTP 400) for an unrecognized state (checks.go:1142/1149/1156). `TestCheckMariaDBOperatorOwnershipCoherence` asserts only `len(errs)` and a message substring — never the structured code. A regression flattening all three to one code, or swapping Conflict↔InvalidRequest, would flip the server's HTTP status and go completely unnoticed.

**Blast radius:** Server/SDK consumers keying retry or HTTP-status handling off the outermost code silently regress (409↔400) with a green test — a real caller-facing contract break.

**Fix:** Add `wantCode` per case and assert via `stderrors.Is(err, aicrerrors.New(wantCode, ""))`: CRsDetected/Unknown → ErrCodeConflict, unrecognized → ErrCodeInvalidRequest.

njhensley (commented):
# 🔎 Multi-Persona Review — feat(slurm): add accounting ownership modes

**Method:** 5 independent persona reviewers (Correctness · Security · Domain & Architecture · API-contract · Test-coverage) fanned out in parallel, then every finding was adversarially re-derived from the resolved code by a senior meta-reviewer that confirmed / refuted / re-tiered it against `ed2f31ab`.

## Overall assessment

Careful, defensively-layered work. The headline claim — *ownership Helm values cannot be changed via `--set`/`--set-json`* — **holds up end-to-end** under a two-layer, fail-closed design (`enforceAccountingOwnership` pre-filter + `ValidateAccountingValues` on fully-resolved values, covering `--set`, `--set-json`, `--set-file`, `--dynamic`, and registry aliases). The evidence-gating state machine fails closed to `unknown` on malformed input, credentials are operator-generated and never logged, and the v1alpha2→v1alpha3 version gating, SDK source-compatibility, and CRDs→operator→instance ordering are all correct.

**No 🔴 blockers.** One 🟠 worth fixing before merge (a shipped-but-unresolvable OpenAPI `$ref` that CI's path-walking sync test structurally cannot catch), one 🟡 spec-compliance gap worth a follow-up (ADR-016 §8), and a cluster of accurate minor test/validation gaps. See inline comments.

## 🟡 Additional minor finding (no inline anchor — the flagged line is unchanged)

**`query` command omits `--slurm-accounting-mode` from duplicate-flag validation** — `pkg/cli/query.go:100`. `pkg/cli/recipe.go:150` added `flagSlurmAccountingMode` to its `validateSingleValueFlags` list; `query.go:100` was not updated, yet `query` registers the flag (`queryCmdFlags → recipeCmdFlags`) and consumes it (`query.go:250`). So `aicr query … --slurm-accounting-mode a --slurm-accounting-mode b` silently last-wins, while the identical `aicr recipe` invocation errors and the server rejects repeated values. **Fix:** add `flagSlurmAccountingMode` to the `validateSingleValueFlags(...)` call at `query.go:100`. (Posted in the body rather than inline because the omission is on an unchanged line, which GitHub can't anchor.)

| claim | disposition |
|-------|-------------|
| `/v2/bundle` `oneOf` not disjoint for absent-apiVersion + configuration | **Refuted** — `ConfiguredRecipeResponse` allOf-inherits `RecipeResponse`'s `required: [apiVersion, kind]`; branches are disjoint on apiVersion. |
| `/v1/bundle` honors accounting v1alpha3 but rejects profiled v1alpha3 | **Intended (🔵)** — profiles are a documented v2-only *user feature*; accounting is recipe-driven bundle behavior the endpoint honors from the resolved recipe. |
| AICR-provided contract literals duplicated Go + 2 values.yaml | **🔵** — `TestEmbeddedAICRProvidedAccountingContract` resolves the real overlay and cross-checks the Go contract; drift fails CI and fails closed at runtime. |
| sacct in-poll `ExitCode!=0` / exec-error branches lack dedicated tests | **🔵** — terminal-unsuccessful and `ExitCode:1` submission paths are tested; untested branches all funnel to the same fail-closed timeout. |
| COMPLETED + empty ExitCode → permanent fail | **🔵** — `sacct --parsable2` emits State and ExitCode atomically; a COMPLETED-but-nonzero job *should* hard-fail. Theoretical. |
| No positive control for v1alpha3 accounting file-load | **🔵** — positive controls exist in `loader_test.go` / `bundle_handler_test.go`. |
| Legacy bundle-time `accounting.enabled` override warn-and-continue | **🔵 non-issue** — unchanged pre-PR customer-managed semantics, no credential/competing-install path. |
| `%02d` hour field truncates past 99h | **No action** — `%02d` is minimum-width, never truncates. |
| CodeRabbit: `/v1/recipe` returns v1alpha3 | **Refuted** — `normalizeLegacyRecipeResult` strips `Configuration` and resets to v1alpha2. |
| CodeRabbit: `slurmAccountingMode` honored on v1 resolving endpoints | **Refuted** — `recipeResolveOptions` rejects it with `ErrCodeInvalidRequest` when `!v2`. |
| CodeRabbit: `ProfileRecipeResponse` drops sibling props under `additionalProperties:false` | **Refuted** — closed branch re-enumerates the new fields; gated by `openapi_sync_test.go`. |
| CodeRabbit: BOM 94 vs 97 image-count inconsistency | **Refuted** — 97 = unique images, 94 = sum of per-component ref counts; pre-existed the PR. |

## Summary

```
🔴 Blocker 0 | 🟠 Major 1 | 🟡 Minor 5 | 🔵 Nitpick 7      Recommendation: Approve with comments
```

**Before merge:** fix 🟠 A1 (inline the `validation` schema). **Follow-up:** ADR-016 §8 disjointness. **Optional:** close the A2/T1/T2 validation + test gaps.

<sub>Reviewed via a multi-persona panel + adversarial senior meta-review. The other "human reviewer" GitHub shows on this PR is the author self-commenting.</sub>

pkg/recipe/accounting.go
```diff
@@ -0,0 +1,314 @@
...
+				normalizedMode, mode))
+	}
+
+	expectedAccounting := mode != AccountingModeDisabled
+	expectedInstall := mode == AccountingModeAICRProvided
+	if err := requireBoolOverride(r, slinkySlurmComponentName,
+		[]string{"accounting", "enabled"}, expectedAccounting); err != nil {
+		return err
+	}
+	for _, name := range []string{
```
yuanchen8911, line 368: This validates the `install` overrides on the three MariaDB components but does not reject `overrides.enabled: false` on them. A recipe artifact carrying `mariadb-operator: {install: true, enabled: false}` in aicr-provided mode passes `ValidateCoherence`, gets filtered by `IsEnabled`, and `ValidateAccountingValues` (pkg/bundler/bundler.go:382) only requires `slinky-slurm` and `slurm-accounting-mariadb` in the effective inventory — so the bundle ships SlurmDBD plus the MariaDB CR with no operator to reconcile it. Reproduced: bundling such a recipe exits 0. Suggest requiring all four components effectively enabled in aicr-provided mode, or rejecting an explicit `enabled: false` on the accounting components here.

api/aicr/v1/server.yaml
yuanchen8911: `$ref: "#/components/schemas/RecipeResponseBase/properties/validation"` is dangling — `RecipeResponseBase` defines no `validation` property, so strict resolution fails, and code generators that resolve the reference can fail. `ProfileRecipeResponse` inlines the validation shape; either share it or inline it here the same way. A ref-integrity assertion in `openapi_sync_test.go` would keep this from recurring.

api/aicr/v1/server.yaml
```diff
@@ -2910,18 +3086,21 @@ components:
       description: >
         Version-partitioned RecipeResult accepted by POST /v2/bundle.
         Legacy input with apiVersion aicr.run/v1alpha2 or no apiVersion remains
-        additive; v1alpha3 input is closed and requires
-        metadata.selectedProfile. The branches are disjoint on apiVersion:
+        additive; v1alpha3 input is closed and requires either
+        metadata.selectedProfile or configuration.slurm.accounting. The
```
yuanchen8911, line 3131: The v2 legacy branch (`LegacyBundleRecipeV2Request`) prohibits `selectedProfile` but not `configuration`, and it inherits `RecipeResponseBase`, which now carries `configuration`. The schema therefore accepts `configuration.slurm.accounting` on an absent/empty/v1alpha2 recipe that the server rejects at runtime (`configuration.slurm.accounting requires apiVersion aicr.run/v1alpha3`). Add the analogous `not: required: [configuration]` there. The v1 `BundleRecipeRequest` has the same gap; since that schema is deliberately additive and single-branch, a conditional (if/then on apiVersion) or at minimum a documented restriction would bring it in line. Both schemas should match runtime validation.

pkg/recipe/accounting.go
yuanchen8911, line 198: This runs after profile resolution (`buildWithStore`, pkg/recipe/builder.go:232) and deep-merges into slinky-slurm overrides without consulting `metadata.selectedProfile.ownedPaths`. A profile declaration that owns `slinky-slurm: accounting.enabled` is silently overwritten while the recipe still records the profile as owning that path — two contradictory ownership claims in one artifact. Reproduced via an external `--data` profile: `selectedProfile.value: off` with effective `accounting.enabled: true`. Suggest rejecting with `ErrCodeInvalidRequest` when the accounting-owned paths intersect the selected profile's `OwnedPaths`.

pkg/recipe/builder.go
```diff
@@ -197,10 +219,23 @@ func (b *Builder) buildWithStore(ctx context.Context, c *Criteria, buildFn func(
...
+	// sequential or concurrent builds cannot contaminate each other or the
+	// cached store.
+	for i := range result.ComponentRefs {
+		result.ComponentRefs[i] = cloneComponentRef(result.ComponentRefs[i])
+	}
 
 	if b.Version != "" {
 		result.Metadata.Version = b.Version
 	}
+	if err := applyBuildConfig(result, buildCfg); err != nil {
```
yuanchen8911, line 233: `DeploymentOrder` was computed by `TopologicalSort` in `finalizeRecipeResult` before this call sets `install: false`, and nothing recomputes it afterward. Since the sort deliberately excludes disabled components (see `componentSets`), every newly generated disabled/customer-managed Slurm recipe lists all three MariaDB components in `deploymentOrder` while `IsEnabled` reports them disabled — visible via `aicr query --selector deploymentOrder`. Bundling stays correct because it re-filters, but the artifact violates the module's own invariant. Recompute or filter the order after the accounting overrides are applied.

recipes/checks/mariadb-operator/health-check.yaml
```diff
@@ -0,0 +1,85 @@
...
+  steps:
+    - try:
+        - assert:
+            resource:
+              apiVersion: apps/v1
+              kind: Deployment
+              metadata:
+                namespace: mariadb-system
+                labels:
+                  app.kubernetes.io/name: mariadb-operator
```
yuanchen8911, line 31: Chart 26.6.0 with these values renders three Deployments — `mariadb-operator`, `mariadb-operator-webhook`, and `mariadb-operator-cert-controller` — each with a distinct `app.kubernetes.io/name` label (verified by rendering the chart). Every assert here selects only `mariadb-operator`, so a crashlooping webhook (which gates MariaDB CR admission) or cert-controller reports healthy. Add the equivalent asserts for the other two labels.

pkg/cli/query.go
```diff
@@ -238,7 +243,31 @@ func buildRecipeFromCmdWithConfig(ctx context.Context, cmd *cli.Command, cfg *ap
 	}
 
 	slog.Info("building recipe from criteria", "criteria", criteria.String())
-	return client.ResolveRecipeFromCriteriaWithProfile(ctx, aicr.WrapCriteria(criteria), profile)
+	return client.ResolveRecipeFromCriteriaWithOptions(ctx, aicr.WrapCriteria(criteria), resolveOpts...)
+}
+
+func accountingResolveOptions(cmd *cli.Command, cfg *appcfg.AICRConfig) ([]aicr.RecipeResolveOption, error) {
```
yuanchen8911, line 250: The query command's `validateSingleValueFlags` call (line 100) omits `flagSlurmAccountingMode` while recipe.go includes it, so `aicr query --slurm-accounting-mode disabled --slurm-accounting-mode aicr-provided` silently uses the last value while `aicr recipe` exits 2 on the same input.

pkg/cli/query.go
```diff
@@ -238,7 +243,31 @@ func buildRecipeFromCmdWithConfig(ctx context.Context, cmd *cli.Command, cfg *ap
...
+		mode, present, err := cfg.Spec.Recipe.ResolveAccountingMode()
+		if err != nil {
+			return nil, err
+		}
+		if !present {
+			return nil, nil
+		}
+		value = string(mode)
+	}
+	if value == "" {
```
yuanchen8911: An explicitly set empty `--slurm-accounting-mode=""` is silently treated as omitted — and since this helper is shared, that holds for both `aicr recipe` and `aicr query`; config resolution likewise treats an explicitly present empty `mode:` as absent. The server rejects an empty `slurmAccountingMode` query value. Rejecting the explicit empty input on the CLI/config surfaces keeps them consistent.

pkg/bundler/bundler.go
```diff
@@ -364,6 +375,288 @@ func (b *DefaultBundler) Make(ctx context.Context, recipeResult *recipe.RecipeRe
...
+					"bundlers filter cannot omit accounting-required component %q for mode %s",
+					name, mode))
+			}
+		}
+	}
+
+	return nil
+}
+
+func (b *DefaultBundler) warnLegacyAccountingOverride(provider recipe.DataProvider) {
```
yuanchen8911: The deprecation warning checks scalar and typed overrides but not `DynamicValues`, while `enforceAccountingOwnership` checks all three sources — a legacy `accounting.enabled` toggle supplied via `--dynamic` skips the warning.

pkg/client/v1/aicr.go
```diff
@@ -1222,6 +1295,50 @@ func (c *Client) BundleComponents(ctx context.Context, r *RecipeResult) ([]Compo
...
+		group.Go(func() error {
```
yuanchen8911, line 1322: One goroutine per Helm component with no concurrency limit. Fine for in-tree catalogs; a large external `--data` catalog could exhaust file descriptors. `group.SetLimit` would bound it — fine as a follow-up.

yuanchen8911 (changes requested):
Thanks — the architecture here is right: typed mode as the single ownership selector, stable install-gated inventory, observational recipe generation with fail-closed bundling, and the recipe as the source of truth for deployed state. The layered enforcement (protected paths, effective-value contract, snapshot-evidence gate, sacct probe) is strong and broadly tested.

Requesting changes for a set of gaps in the coherence net, detailed inline: an `enabled: false` escape that ships an incoherent AICR-provided bundle, an invalid `$ref` in the new OpenAPI schema plus a version-partitioning gap for `configuration` in both bundle request schemas, the accounting mutation bypassing profile `OwnedPaths`, a stale `deploymentOrder` in disabled/customer-managed recipes, an incomplete operator health check, and a missing single-value guard on `aicr query`.

Two smaller notes inline as well (empty-flag handling, `--dynamic` gap in the deprecation warning), and two PR-body fixes: the rollout notes should mention that external Slurm compositions which neither inherit the updated leaves nor declare the three MariaDB refs will no longer resolve, and the MariaDB resource figure should read 6Gi requests (4096M is the InnoDB buffer pool), not 4 GiB.

kaynetu:
I will address the comments and look into test failures.

pkg/bundler/validations/checks.go
kaynetu: Confirmed—this behavior is intentional for criteria-only recipes and older snapshots. Without `metadata.mariaDBOperatorState`, bundling warns that target-cluster conflicts were not evaluated and proceeds for backward compatibility; users who require conflict detection should capture a current snapshot. This is now explicit in [Conflict detection requires snapshot evidence](https://github.com/NVIDIA/aicr/blob/7c07fa1caddd4a989b27c4902b2096757c0cc265/docs/user/slinky-slurm-accounting.md#conflict-detection-requires-snapshot-evidence) and the [CLI recipe guidance](https://github.com/NVIDIA/aicr/blob/7c07fa1caddd4a989b27c4902b2096757c0cc265/docs/user/cli-reference.md#snapshot-mode).

api/aicr/v1/server.yaml
kaynetu: Fixed in `7c07fa1c`: `ConfiguredRecipeResponse.validation` is now an inline schema, and the OpenAPI sync coverage guards the configured response contract.

pkg/recipe/accounting.go
kaynetu: Fixed in `7c07fa1c`: accounting-owned paths are checked against `selectedProfile.ownedPaths` before mutation. Exact, ancestor, and MariaDB-component overlap cases now return `INVALID_REQUEST` in table tests.

pkg/bundler/bundler.go
kaynetu: Fixed in `7c07fa1c`: added a passing `float64(3306)` customer-managed port case, alongside the existing fractional-float rejection coverage.

pkg/bundler/validations/checks_test.go
kaynetu: Fixed in `7c07fa1c`: the ownership-coherence table now asserts structured codes with `errors.Is`—`CONFLICT` for detected/unknown state and `INVALID_REQUEST` for an invalid state.

pkg/recipe/accounting.go
kaynetu: Fixed in `7c07fa1c`: coherence validation now requires effective enablement of `slinky-slurm` and all three MariaDB components. Loader and direct bundler regressions cover explicit `enabled: false`.

api/aicr/v1/server.yaml
kaynetu: Fixed in `7c07fa1c`: the dangling reference was removed by inlining the validation schema, with OpenAPI sync coverage for the configured response contract.

api/aicr/v1/server.yaml
kaynetu: Fixed in `7c07fa1c`: `LegacyBundleRecipeV2Request` prohibits `configuration`, and the v1 bundle schema conditionally prohibits it for absent, empty, and `v1alpha2` headers. OpenAPI contract tests cover both routes.

pkg/recipe/accounting.go
kaynetu: Fixed in `7c07fa1c`: accounting/profile owned-path intersections now fail with `INVALID_REQUEST` before accounting overrides are applied, with exact and ancestor-overlap regression cases.

pkg/recipe/builder.go
kaynetu: Fixed in `7c07fa1c`: `applyBuildConfig` recomputes `DeploymentOrder` after applying install overrides. Tests cover disabled, customer-managed, and AICR-provided modes.

recipes/checks/mariadb-operator/health-check.yaml
kaynetu: Fixed in `7c07fa1c`: the health check now covers deployment availability plus pod, container, and init-container failure states for the operator, webhook, and cert-controller.

pkg/cli/query.go
kaynetu: Fixed in `7c07fa1c`: query duplicate-value validation now includes `slurm-accounting-mode`, with a repeated-flag regression test.

pkg/cli/query.go
kaynetu: Fixed in `7c07fa1c`: explicitly empty accounting mode values are now rejected consistently by recipe/query CLI parsing and config resolution, with coverage for each surface.

pkg/bundler/bundler.go
kaynetu: Fixed in `7c07fa1c`: the legacy warning now resolves and canonicalizes `DynamicValues` as well as scalar and typed overrides. The regression table covers all three sources.

pkg/client/v1/aicr.go
kaynetu: Fixed in `7c07fa1c`: Helm value resolution now uses `SetLimit(defaults.HelmValueResolutionConcurrency)` with a limit of 8, plus a saturation test that proves no ninth read starts before release.

njhensley (dismissed):
## ✅ Approve

Re-reviewed at `d59b446c`. All prior feedback is resolved in code — verified, not just marked resolved.

**My 6 findings** (from the `ed2f31ab` review) — all ✔️ Addressed:
- 🟠 Dangling `$ref` (`ConfiguredRecipeResponse.validation`) → inlined schema, identical to `ProfileRecipeResponse`
- 🟡 ADR-016 §8 owned-path disjointness → `ValidateOwnershipDisjoint` + overlap tests
- 🟡 Criteria-only aicr-provided conflict gate → documented ("Conflict detection requires snapshot evidence")
- 🟡 `query` duplicate-flag validation → `flagSlurmAccountingMode` added + regression test
- 🟡 Whole-number `float64` port → passing test case added
- 🟡 Ownership-coherence error code → `wantCode` + `stderrors.Is` assertions (409 vs 400)

**Other reviewers** (`coderabbitai` ~40 threads, `yuanchen8911` ~11) — every thread resolved and **16/16 substantive items code-verified as genuinely fixed**, including: closed v1alpha3 schemas re-enumerate `configuration`/`mariaDBOperatorState`; contract-path duplication pinned by test; ctx-cancellation honored in the Helm fan-out; MariaDB components can't be disabled via `enabled:false`; `DeploymentOrder` recomputed after `install:false`; `oneOf` hardened with `not: required:[configuration]`; sacct `CANCELLED by <uid>` treated terminal.

One cosmetic nit (non-blocking): `recipes/checks/mariadb-operator/health-check.yaml` sets `spec.timeouts.assert` but not `.error` — the rejection assertions are all present, so it falls back to the chainsaw default. Fine to leave.

No blockers, no regressions. Nice, thorough work on the ownership contract and the fail-closed validation. 🚢

api/aicr/v1/server.yaml
```diff
@@ -2943,18 +3163,31 @@ components:
         admitted as the legacy shape.
 
         metadata.selectedProfile remains reserved for aicr.run/v1alpha3 and
-        is rejected on this route.
+        is rejected on this route. Typed configuration is accepted only when
+        apiVersion is aicr.run/v1alpha3.
```
yuanchen8911: The runtime decode for this route is now strict for v1alpha3 (`decodeRecipeResultRequest` applies `DisallowUnknownFields` when apiVersion is aicr.run/v1alpha3, pkg/server/bundle_handler.go), but this schema stays additive — no `additionalProperties: false` and no mention of the restriction. A body that validates against this spec can still get a 400 on an unknown field. Since the behavior fails closed, a sentence here documenting that v1alpha3 bodies are decoded strictly and unknown fields are rejected is sufficient; matching the v2 branch structure isn't needed.

yuanchen8911 (commented):
Thanks for the update — I verified each fix at d59b446c and all ten inline findings from my previous review are correctly addressed. Four small items remain before approval: one inline below, plus the three here.

The heading rename in docs/user/api-reference.md ("Profile-aware v2 endpoints" → "Configured v2 endpoints", line 399) left a broken inbound link: docs/integrator/automation.md line 370 still points to `../user/api-reference.md#profile-aware-v2-endpoints` with the old link text. It's the only inbound reference to the old slug repo-wide. CI didn't catch it because lychee's fragment checking is off (`.lychee.toml` doesn't enable it). Please update the fragment and the link text in this PR.

Two PR-description corrections:

1. The body says "MariaDB requests 4 GiB, limits 8 GiB", but the shipped values request 6Gi (recipes/components/slurm-accounting-mariadb/values.yaml lines 38–44); 4096M is the `innodb_buffer_pool_size`, not the pod request.

2. The rollout notes should mention the external-composition break: every Slurm resolution now defaults accounting and requires the three MariaDB componentRefs in `applyBuildConfig` (pkg/recipe/accounting.go), so external `--data` Slurm compositions that neither inherit the updated leaves nor declare those refs stop resolving — including in `disabled` mode.

api/aicr/v1/server.yaml
kaynetu: addressed in the latest commit

yuanchen8911 (approved):
Approved at 35e255c2e. Verified the four remaining items from my previous pass: the v2 bundle request schema now documents strict v1alpha3 decoding (unknown fields rejected with 400), the automation.md link points at the renamed configured-v2-endpoints anchor, and the PR body corrections (MariaDB 6Gi request figure, external-composition break in the rollout notes) are in. All earlier inline findings were verified fixed at d59b446c and carry forward. Please let CI finish on this head before merging.

Issue #1632, opened 2026-07-07
validator lifecycle: guarantee cleanup and result integrity across Job deadline, SIGTERM, and orchestrator timeout


yuanchen8911:
## Summary

Validator Jobs have no guaranteed window to run cleanup or report results when their deadline fires or the pod is terminated. The Job's `activeDeadlineSeconds` and the in-process check timeout (`AICR_CHECK_TIMEOUT`) carry the same duration, but the Job clock starts at pod creation — scheduling and image-pull time eat into the budget. A check that starts late is SIGKILLed mid-run with its Go `defer`s never executing: test resources leak, and the run's real result is replaced by a generic Job timeout. This affects every validator check.

## Background

This was surfaced and fully designed during review of #1620 (rounds addressing cleanup lifecycle), then deliberately descoped: each fix widened the blast radius (all Job deadlines, orchestrator waits, termination grace, signal semantics, result extraction) beyond that PR's GPU-allocation subject. #1620 ships honest local containment instead — the two GPU checks bound their own work (`draSupportWorkBudget`/`secureAccessWorkBudget`) so their bounded cleanup fits inside existing catalog timeouts, and leaked test namespaces carry the ownership label `aicr.run/gpu-test-run` for sweeping. The full design was implemented once and reverted; the analysis lives in #1620's review history.

## Proposed design (from the #1620 review rounds)

1. **Job-deadline headroom** — `activeDeadlineSeconds = check timeout + startup grace + cleanup grace`, keeping `AICR_CHECK_TIMEOUT` bare so the inner context expires first and deferred cleanup gets a real budget.
2. **SIGTERM-aware check context** — wrap the shared `validators.LoadContext` context with `signal.NotifyContext(..., SIGTERM)` so a Job-deadline kill cancels the running check gracefully; raise `terminationGracePeriodSeconds` above the cleanup budget so cleanup and result flushing complete before SIGKILL.
3. **Orchestrator wait-envelope coherence** — `Deployer.WaitForCompletion` must cover the extended Job deadline plus the SIGTERM window (and the client facade's `ValidationOperationTimeout` must exceed the worst-case per-check envelope, invariant-tested), so the orchestrator never classifies a run as timed out while the pod is still legitimately finishing.
4. **Fail-closed result extraction** — when the Job reports failure/deadline, treat any flushed partial result as authoritative rather than masking it with a generic timeout classification.
[...]

Issue #1192, opened 2026-06-04
inference-perf: validator times out on healthy cluster when dynamo frontend discovery bootstrap races (false negative)


yuanchen8911:
## Summary

The `inference-perf` validator intermittently fails (~50% of runs) with `[TIMEOUT] timed out waiting for inference endpoint to serve requests` even though the DynamoGraphDeployment reports `successful`, all decode workers are `Ready` (8/8), and the frontend pod is `1/1`. The frontend reaches "ready" but never actually serves `/v1/chat/completions`, so `waitForEndpointReady` exhausts its 5-minute window and the phase fails on an otherwise-healthy cluster.

This is a **false negative**: the deployment is fine; the gate fails because of a transient dynamo frontend discovery-bootstrap race.

## Symptom

```
time=...Z level=ERROR msg=FAIL message="[TIMEOUT] timed out waiting for inference endpoint to serve requests: context deadline exceeded"
[cli] phase result: phase=performance status=failed duration=13m2s
```

Observed repeatedly across runs on the RTX PRO 6000 cluster (`b40-aicr`, EKS `sdrvxbrhcc-dgxc-k8s-aws-usw2-non-prod`). Clean runs at concurrency 1024 / 1536 / 2048 all pass — when the race is won. It is intermittent (~50%), independent of concurrency, and not load-related.

## Root cause — dynamo frontend discovery stream panics and never recovers

Captured the live frontend pod logs during the endpoint-wait window of a flaked run. The frontend has **two independent discovery layers that disagree**:

| Layer | Role | State in the flaked run |
|---|---|---|
| `dynamo_runtime::discovery::kube` | Kubernetes-level instance discovery | ✅ Saw all 8 workers — `Snapshot (seq=13): 8 instances, added=["ee504b38b0550"]` (logged **once**, early) |
| `dynamo_llm` / `KVStoreDiscovery` (`v1/instances` bucket) | The KV store the HTTP router queries to find model backends | ❌ Never populated — `KVStoreDiscovery::list: bucket missing for query=AllEndpoints, prefix=v1/instances, bucket=v1/instances` logged **97×** through the whole window |

The smoking gun — the discovery stream that feeds `v1/instances` panics and the tokio worker never recovers:

```
thread 'tokio-runtime-worker' panicked at
/usr/local/cargo/registry/src/index.crates.io-.../futures-util-0.3.31/src/stream/unfold.rs:108:21:
Unfold must not be polled after it returned `Poll::Ready(None)`
[...]

PR #1971, opened 2026-07-30
fix(validator): re-establish Job watch after apiserver idle-stream closure

### LINKED ISSUE #1966: Bug(validate): Job watch not re-established after API-server idle-stream closure
### Prerequisites

## Summary

`aicr validate` waits on each validator Job with a single Kubernetes watch. kube-apiserver terminates every watch stream after `--min-request-timeout` (default 1800s, randomized ×[1,2) → 30–60 min) — routine behavior that clients are expected to survive by re-establishing the watch. aicr instead treats the closure as a fatal wait failure: the validator is torn down mid-run and reported as timed out, regardless of its catalog `timeout`.

## Environment

- aicr v0.18.0
- k0s v1.36.2, default apiserver flags (`--min-request-timeout` unset = 1800s)
- bare metal, single node, 8× H200 SXM
- Applies to any default-configured API server (kubeadm, k0s, managed clouds)

## What happened

External validator (orchestrator pattern) with catalog `timeout: 6h`, expected ~5h runtime — aborted at **34m4s**:

```
[cli] running validator: name=gpu-burnin phase=performance
[cli] validator completed: name=gpu-burnin status=other
[cli] phase completed: phase=performance status=other validators=1 passed=0 failed=0 duration=34m4.655641001s
[cli] command failed: error=[INTERNAL] validation failed: one or more phases did not pass exitCode=8
```

CTRF: exit code −1, message `timeout: validator did not complete within 6h0m0s` — after 34 minutes. The validator and its workload were healthy to the moment of teardown (its telemetry stream was clean throughout). As a side-effect, the teardown orphaned the workload Job the validator had created — it kept running unwatched (see Related issues).

## Root cause (v0.18.0)

- `pkg/k8s/pod/job.go` `WaitForJobTerminal`: opens ONE watch; when the result channel closes, it re-checks the Job once and — if not yet terminal — returns `ErrCodeUnavailable "watch channel closed before Job reached terminal state"` instead of re-watching. (Additionally, `defer watcher.Stop()` binds the first watcher value, so any re-watch variable reassignment would leak the stream.)
- `pkg/validator/validator.go` (~L429) funnels every wait error into `HandleTimeout`, and `pkg/validator/job/result.go:149` renders the *configured* `Entry.Timeout` in the message regardless of the actual cause — hence "6h0m0s" reported at minute 34.

## Impact

- No validator can reliably run longer than ~30 min through `aicr validate` on a default-configured API server; completions in the 30–60 min band are jitter luck, regardless of catalog `timeout`.
- The embedded `inference-perf` cold-start path (~53 min per the budget comments in `pkg/defaults/timeouts.go`; 65m catalog timeout) cannot complete on a default API server.
- The misleading "timeout: … within <configured>" message misdirects diagnosis toward the catalog timeout.

## Reproduction

Default cluster (`--min-request-timeout` unset), a catalog entry wrapping `sleep 2700` with `timeout: 2h`, `aicr validate --phase performance` → the check dies somewhere in the 30–60 min band with the timeout message.

## Proposed solutions (happy to submit the PR — maintainers' pick)

**A — minimal:** on channel close, re-check the Job; if not terminal, re-establish the watch from the re-check's `ResourceVersion` and continue waiting. The wait's context deadline (catalog timeout + buffer) remains the sole give-up condition. Handle "410 Gone" with a fresh Get; change the defer to closure form (`defer func() { watcher.Stop() }()`).

**B — canonical:** replace the hand-rolled loop with client-go `watchtools`/`RetryWatcher`, which implements exactly this resume semantics (including 410 handling and bookmarks).

## Note on hang detection

This fix does not weaken it. Watch closure was never correlated with workload health — it kills healthy long runs and sleeps through the first ~30 minutes of a genuinely hung one. The real timeout authorities are untouched: the wait's context deadline and both Jobs' `activeDeadlineSeconds`.

## Related issues (not duplicates)

- #1632 — in-Job lifecycle (deadline headroom, SIGTERM grace, cleanup) for embedded checks. This issue is the CLI-side wait; the workload-orphaning side-effect observed here is an instance of #1632's class.
- #1192 — inference-perf false-negative timeout from an in-validator endpoint bootstrap race (~13-min runs); different mechanism.

njhensley:
## Summary

`aicr validate` waits on each validator Job with a single Kubernetes watch. kube-apiserver terminates every watch stream after `--min-request-timeout` (30–60 min, randomized), which clients are expected to survive by re-establishing the watch. `WaitForJobCompletion`/`WaitForJobTerminal` now re-Get the Job and re-establish the watch from the observed `ResourceVersion` instead of reporting the closure as a fatal wait failure.

## Motivation / Context

On a default-configured API server no validator can reliably run longer than ~30 min through `aicr validate`; completions in the 30–60 min band are jitter luck regardless of the catalog `timeout`. The healthy long-running validator is torn down mid-run and mislabeled as timed out.

Fixes: #1966
Related: #1632, #1192

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

## Implementation Notes

- `pkg/k8s/pod/job.go` — `WaitForJobCompletion`/`WaitForJobTerminal` re-establish the watch from the observed `ResourceVersion` when the result channel closes without context cancellation; a `410 Gone` (stale ResourceVersion) triggers a fresh Get and restarts the watch from current state instead of returning `ErrCodeUnavailable`.
- The watcher `defer` is changed to closure form (`defer func() { watcher.Stop() }()`) so the re-established stream is stopped rather than leaking the original, now-superseded watcher.
- `pkg/validator/job/result.go` — `HandleTimeout` now threads the actual wait cause through, so an infrastructure failure renders `validation failed: <cause>` instead of the misleading `timeout … within <configured catalog timeout>` that misdirected operators toward the catalog knob.
- The context deadline and the Job's `activeDeadlineSeconds` remain the sole give-up authorities; hang detection is unchanged (approach A from the issue).

## Testing

```bash
make test        # -race, GREEN on pkg/k8s/pod and pkg/validator/...
make lint        # golangci-lint GREEN on affected packages
make test-coverage  # 79.6% total (threshold 75%) — PASS
```

Added tests for the re-watch/resume path (`pkg/k8s/pod/job_test.go`, `wait_termination_test.go`) and the cause-threading message (`pkg/validator/job/result_test.go`).

## Risk Assessment

**Rollout notes:** N/A — no user-facing API, flag, or documented-behavior change.

## Checklist

- [x] I updated docs if user-facing behavior changed — N/A (no user-facing behavior change)

### DIFF
--- a/pkg/k8s/pod/job.go
+++ b/pkg/k8s/pod/job.go
@@ -16,15 +16,122 @@ package pod
 
 import (
 	"context"
+	"log/slog"
 	"time"
 
+	"github.com/NVIDIA/aicr/pkg/defaults"
 	"github.com/NVIDIA/aicr/pkg/errors"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/apimachinery/pkg/watch"
 	"k8s.io/client-go/kubernetes"
 )
+
+// jobWatchResumeBackoff paces each watch re-establishment in rewatchJob. It is
+// a package var (not the constant directly) only so tests can drop it to a
+// negligible value and exercise the resume path without a real sleep;
+// production code never reassigns it.
+var jobWatchResumeBackoff = defaults.K8sJobWatchResumeBackoff
+
+// jobWatchResumeJitterFactor spreads the resume backoff by up to +20% so that
+// many validators whose watches were all closed by the same apiserver rollout
+// do not reconnect in lockstep and produce a synchronized thundering herd.
+const jobWatchResumeJitterFactor = 0.2
+
+// resumeContext builds the structured-error context shared by the resume
+// helper so a resume failure names the Job, mirroring the sibling pod-wait
+// helpers.
+func resumeContext(namespace, name string) map[string]any {
+	return map[string]any{keyNamespace: namespace, keyName: name}
+}
+
+// isRetryableWatchError reports whether a watch.Error event is a routine,
+// resumable signal (HTTP 410 Gone / ResourceExpired — the apiserver compacted
+// past the watch's ResourceVersion) rather than a fatal stream error. Callers
+// treat a retryable error like a channel close and re-establish the watch;
+// anything else aborts the wait.
+//
+// Against a real apiserver this is the *common* shape of a stale-ResourceVersion
+// rejection: the Watch call itself succeeds and the 410 arrives as the stream's
+// first ERROR event, not as a synchronous call error.
+func isRetryableWatchError(event watch.Event) bool {
+	if event.Type != watch.Error {
+		return false
+	}
+	err := apierrors.FromObject(event.Object)
+	return apierrors.IsResourceExpired(err) || apierrors.IsGone(err)
+}
+
+// resumeJobWatch reconnects a Job watch that ended — the channel closed or the
+// apiserver emitted a retryable 410 — before the Job reached a terminal state.
+//
+// It resyncs via a field-selected List rather than a Get, which is load-bearing:
+//   - The List's collection ResourceVersion (metadata.resourceVersion) is the
+//     current cluster revision, so the re-established watch always starts from a
+//     live RV. Re-watching from the Job's own object ResourceVersion is unsafe:
+//     a Job that runs untouched past the apiserver's watch-cache window keeps
+//     reporting the same aged-out object RV, so every watch from it 410s
+//     immediately and the wait degrades to paced polling for the rest of the run
+//     (the exact long-running-validator case issue #1966 targets).
+//   - The List still returns the Job, so a terminal transition that landed while
+//     the watch was down is observed here and not missed. Because the new watch
+//     starts from the List's collection RV, any transition after the List is
+//     replayed by the watch — there is no List-to-Watch gap (an empty
+//     ResourceVersion would reopen that gap and is deliberately not used).
+//
+// A jittered bounded backoff precedes every attempt so a stream that keeps
+// ending immediately (flapping LB, apiserver rollout) cannot make the loop spin,
+// and fleets sharing one apiserver do not reconnect in lockstep. The backoff
+// honors ctx cancellation, so the caller's context deadline remains the sole
+// give-up authority.
+//
+// It returns exactly one of:
+//   - terminal != nil: the Job was terminal in the resync List; the caller
+//     returns it (w is nil, nothing to continue).
+//   - w != nil: a fresh watch to continue consuming (terminal is nil).
+//   - err != nil: give up. Transient List/Watch failures are classified as
+//     ErrCodeUnavailable (vs ErrCodeTimeout when the context is done) via
+//     classifyReGetError, matching the sibling pod-wait helpers.
+func resumeJobWatch(ctx context.Context, client kubernetes.Interface, namespace, name string) (terminal *batchv1.Job, w watch.Interface, err error) {
+	select {
+	case <-time.After(wait.Jitter(jobWatchResumeBackoff, jobWatchResumeJitterFactor)):
+	case <-ctx.Done():
+		return nil, nil, errors.WrapWithContext(errors.ErrCodeTimeout,
+			"context canceled before Job watch resume", ctx.Err(), resumeContext(namespace, name))
+	}
+
+	// Resync via List: its collection ResourceVersion is current, and its result
+	// lets us catch a terminal transition that landed while the watch was down.
+	list, listErr := client.BatchV1().Jobs(namespace).List(ctx, metav1.ListOptions{
+		FieldSelector: "metadata.name=" + name,
+	})
+	if listErr != nil {
+		return nil, nil, classifyReGetError(ctx, "job watch closed and Job re-list failed", listErr)
+	}
+	if len(list.Items) == 0 {
+		// The Job no longer exists — deleted while the watch was down.
+		return nil, nil, errors.NewWithContext(errors.ErrCodeInternal,
+			"Job not found during watch resume", resumeContext(namespace, name))
+	}
+	job := &list.Items[0]
+	if isJobTerminal(job) {
+		return job, nil, nil
+	}
+
+	// Re-establish the watch from the current collection ResourceVersion so it
+	// starts from a live revision and replays any change after the List.
+	w, watchErr := client.BatchV1().Jobs(namespace).Watch(ctx, metav1.ListOptions{
+		FieldSelector:   "metadata.name=" + name,
+		ResourceVersion: list.ResourceVersion,
+	})
+	if watchErr != nil {
+		return nil, nil, classifyReGetError(ctx, "failed to re-establish Job watch after resync", watchErr)
+	}
+	return nil, w, nil
+}
 
 // WaitForJobCompletion waits for a Kubernetes Job to complete successfully or fail.
 // Returns nil if job completes successfully, error if job fails or context deadline exceeded.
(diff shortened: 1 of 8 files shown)


pkg/k8s/pod/job.go
```diff
@@ -16,16 +16,145 @@ package pod
...
+//     classifyReGetError, matching the sibling pod-wait helpers.
+func resumeJobWatch(ctx context.Context, client kubernetes.Interface, namespace, name string) (terminal *batchv1.Job, w watch.Interface, err error) {
+	recheck, recheckErr := client.BatchV1().Jobs(namespace).Get(ctx, name, metav1.GetOptions{})
+	if recheckErr != nil {
+		return nil, nil, classifyReGetError(ctx, "job watch closed and Job re-check failed", recheckErr)
+	}
+	if isJobTerminal(recheck) {
+		return recheck, nil, nil
+	}
+	w, resynced, rewatchErr := rewatchJob(ctx, client, namespace, name, recheck.ResourceVersion)
```
yuanchen8911: `recheck.ResourceVersion` is the Job's last-modified revision, not a current cluster RV. A Job running untouched for 30+ minutes returns the same stale RV on every Get, so once that RV has aged out of the watch cache, every rewatch attempt from here 410s again and the loop degrades to paced polling (Get -> Watch -> ERROR-410 -> resume) until the Job goes terminal or the deadline expires. The resync branch inside `rewatchJob` (job.go:85-92) has the same root defect: its fresh Get returns the same stale object RV, so even when reached it restarts a watch that fails identically.

Resync from a field-selected List instead and pass its `metadata.resourceVersion` — the collection RV is current, and the List result still lets you check for a terminal transition that landed in the gap.

pkg/k8s/pod/job.go
```diff
@@ -16,16 +16,145 @@ package pod
...
+		return nil, nil, errors.WrapWithContext(errors.ErrCodeTimeout,
+			"context canceled before Job watch resume", ctx.Err(),
+			rewatchContext(namespace, name, fromRV))
+	}
+
+	w, err = client.BatchV1().Jobs(namespace).Watch(ctx, metav1.ListOptions{
+		FieldSelector:   "metadata.name=" + name,
+		ResourceVersion: fromRV,
+	})
+	if err == nil {
```
yuanchen8911: Against a real apiserver this branch is taken even when `fromRV` is expired: the cacher returns the watch successfully and delivers the 410 as the stream's first ERROR event (kubernetes v1.36.3, staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L636-L641), so the call-error resync below never runs in the common case. Recommend routing both forms — the synchronous 410 caught here and the event-form 410 caught by `isRetryableWatchError` — through a single List-based recovery path, so the two shapes of the same apiserver signal can't drift apart.

pkg/k8s/pod/job_test.go
```diff
@@ -168,6 +171,374 @@ func TestWaitForJobCompletion_WatchError(t *testing.T) {
...
+			return true, second, nil
+		default:
+			return true, watch.NewFake(), nil
+		}
+	})
+
+	go func() {
+		time.Sleep(10 * time.Millisecond)
+		// Emit a 410 Gone as a watch.Error event rather than closing the channel.
+		expired := apierrors.NewResourceExpired("compacted")
```
yuanchen8911, line 445: The existing tests cover synchronous 410 errors and ERROR-event 410s separately, but they do not model the critical stale-RV behavior: in the event-form tests here and in wait_termination_test.go, the second watch succeeds even though the preceding Get still returns object RV "1". Add a regression whose watch reactor inspects the requested ResourceVersion and emits an ERROR-410 for every watch from "1"; the List reactor should return a newer `ListMeta.ResourceVersion`, and only a watch from that collection RV should become usable. Assert that the List-based resync occurred and a working watch was established, rather than merely asserting that repeated Gets eventually observed terminality.

yuanchen8911 (changes requested):
The re-establishment path means the wait no longer aborts on idle-stream closure, and `isRetryableWatchError` now decodes an ERROR-event 410 properly. One residual defect remains in the aged-RV recovery: the resume always restarts the watch from the Job's object ResourceVersion, which for an unchanged long-running Job is the same stale RV the apiserver just rejected. Liveness is preserved — each resume cycle's Get can still observe terminality — but a usable watch is never re-established in exactly the long-running-validator scenario this PR targets.

Reproduced against 2823dd8 with a fake client mirroring the real apiserver's semantics (Watch from an aged-out RV succeeds and emits a 410 Status as the first ERROR event; Get of the unchanged Job keeps returning object RV "1"): every re-established watch 410ed immediately and the wait converged only through the resume-path Gets. Once the RV has aged out, the loop settles into one Get+Watch pair every roughly 2-2.4 seconds plus request latency for the rest of the run — the backoff caps the rate but does not prevent the persistent polling.

The race-safe fix: on 410, resync via a field-selected List, evaluate the returned Job for terminal state, and start the new watch from the List's `metadata.resourceVersion` (the current collection RV). An empty ResourceVersion is not a safe substitute — a terminal transition landing in the List-to-Watch gap would never be replayed and the wait would hang to its deadline.

njhensley:
Thanks for the thorough review, @yuanchen8911 — the stale-object-RV analysis is exactly right, and the reproduction was very helpful. Pushed `be6fc881` addressing it and the CodeRabbit items.

**Core fix — stale object RV → List-based resync (job.go:76,147)**
`resumeJobWatch` no longer re-watches from the Job's object `ResourceVersion`. It now resyncs via a **field-selected List** and re-establishes the watch from the List's `metadata.resourceVersion` (the current *collection* RV), which is always live. The List result is still evaluated for a terminal transition that landed during the gap, and because the new watch starts from the collection RV, any transition after the List is replayed — no List-to-Watch gap (empty RV deliberately not used, per your note). This also **unifies both 410 shapes**: the synchronous call-error branch is gone entirely, so the ERROR-event 410 (`isRetryableWatchError`) and any residual call-level failure both funnel through the one List-based recovery path — they can't drift apart.

Added a regression in both `job_test.go` and `wait_termination_test.go` (`*_StaleObjectRVResyncsViaList`) with a watch reactor that emits an ERROR-410 for any watch from the aged-out object RV and only serves a usable watch from the collection RV; the List reactor returns a newer `ListMeta.ResourceVersion`. The test asserts the resync List occurred and a working watch was established from the collection RV — it times out if the code regresses to the object RV.

**CodeRabbit items**
- job.go:98 — transient re-establish-Watch failures now route through `classifyReGetError` → `ErrCodeUnavailable`, consistent with the re-list path.
- wait_internal_test.go — added direct `resumeJobWatch` branch tests (List-fail → Unavailable, Watch-fail → Unavailable, Job-deleted → Internal, ctx-canceled-during-backoff → Timeout).
- result.go:179 — `HandleTimeout` now renders the *effective* timeout (`ValidatorDefaultTimeout` fallback that `runPhase` applies) instead of a bare `0s`.
- result.go:232 — `boundTerminationMsg` trims back to a UTF-8 rune boundary so truncation can't split a multi-byte rune.

`go test -race`, `golangci-lint`, and `make test-coverage` (80.9%, floor 80) all green; `resumeJobWatch` is 100% covered.

pkg/k8s/pod/job.go
yuanchen8911: Addressed in be6fc881 — verified: resumeJobWatch resyncs via a field-selected List, evaluates the List result for terminal state (including Job-deleted handling), and re-establishes the watch from the List collection ResourceVersion; the empty-RV replay gap is deliberately avoided and documented. Resolving.

pkg/k8s/pod/job_test.go
yuanchen8911: Addressed in be6fc881 — verified: TestWaitForJobCompletion_StaleObjectRVResyncsViaList uses a watch reactor that 410s every watch from the stale object RV and only admits a watch from the newer List collection RV, plus resync-terminal, resumes-twice, and transient-List-failure coverage. Resolving.

pkg/k8s/pod/job.go
yuanchen8911: be6fc881 unifies the event-form 410 through the List-based recovery, which covers the common shape against a real apiserver. The synchronous form was removed rather than unified: a 410 returned by the Watch call itself (initial watch, or the resync re-watch) is now a hard failure instead of triggering recovery.

Reachability of that residual is narrow — the sync form arises when the watch cache is disabled (non-default on the managed platforms AICR targets; the cacher otherwise accepts the watch and delivers the 410 as the first ERROR event), and the resync re-watch starts from a fresh collection RV where a sync 410 is essentially impossible. So this is nonblocking and I'm approving the PR as-is.

If you want to close the corner cheaply: check apierrors.IsResourceExpired(err) || apierrors.IsGone(err) on the initial Watch call error and route it into the same resumeJobWatch loop. Otherwise fine to leave as a known limitation.

yuanchen8911 (approved):
be6fc881 verified — the aged-RV resume defect is fixed with the race-safe List-based recovery (resync via field-selected List, terminal evaluation of the List result, watch re-established from the current collection ResourceVersion, no empty-RV replay gap), and the stale-RV regression test models the real apiserver semantics exactly. Approving.

One residual noted on the open thread: the synchronous 410 form is a hard failure rather than routed through the recovery path — reachable only with the watch cache disabled, so nonblocking; a cheap IsResourceExpired/IsGone check on the Watch call error would close it if desired.

njhensley:
The remaining unresolved review threads (the `boundTerminationMsg` UTF-8 erasure + its test gap, the container-not-found timeout-vs-cause message, the `staleRVWatchReactor` test duplication, and the stale `resumeJobWatch` godoc wording) are tracked as a follow-up in #1976 so this PR can land. The `boundTerminationMsg` item is a real correctness bug and is flagged as higher-priority there.

PR #1973, opened 2026-07-30
feat(validator): carry redaction-safe CTRF extra outcome data

### LINKED ISSUE #1951: Validator evidence disclosure (e.g. node coverage lines) is stripped by default redaction
## Context

`#1936` added a `nodesValidated: <schedulable>/<total>` coverage line (plus per-node `skipped (cordoned)` disclosure) to `check-nvidia-smi`'s stdout evidence, so a pass on a reduced node set is visible rather than silent (`#1668`).

That disclosure only reaches `TestResult.Stdout`/`.Message` in the CTRF report. `pkg/evidence/redact/redact.go` documents that its "minimal" policy (the default) omits per-test `Stdout` and `Message` from a published, signed evidence bundle — only `--full` keeps them (`pkg/evidence/attestation/emit.go`).

So in the default signed-bundle path, `check-nvidia-smi` shows as `passed` with no message and no stdout: a 1-of-2-node pass is indistinguishable from a 2-of-2 pass, and an all-cordoned skip loses its reason. The disclosure is real for `aicr validate` live/`--output` runs and for `--full` bundles, but not for the artifact a downstream consumer actually verifies by default.

## Proposal

Carry structured, low-cardinality outcome data (e.g. per-check coverage counts, skip-reason codes) in a CTRF field that survives the default redaction policy, instead of overloading `Stdout`/`Message` (free-form log text, correctly redacted by default since it can leak node names, IPs, etc.).

This likely means:
- Extending `pkg/validator/ctrf.TestResult` with a small structured field (e.g. an `extra`/`properties`-style map, matching the upstream CTRF spec's `extra` object) that specific checks can populate.
- Deciding, in `pkg/evidence/redact`, what is safe to allowlist there by default (the whole point of the minimal policy is to fail closed on unknown data, so this needs its own allowlist entry, not a blanket keep-all).
- Wiring `checkNvidiaSMI` (and any other check emitting similar coverage disclosure) to populate it instead of relying on `fmt.Printf`.

## Non-goals

- Not proposing to change the default redaction policy's treatment of `Stdout`/`Message` — that's correctly conservative for free-form log text.

Found during review of `#1936`.

njhensley:
## Summary

Carry validator coverage disclosure (per-check node counts, skip-reason codes) in a structured CTRF `extra` field that survives the default "minimal" redaction policy, instead of overloading `Stdout`/`Message` (which are correctly stripped from signed evidence bundles by default).

## Motivation / Context

`#1936` added `nodesValidated: <schedulable>/<total>` (plus per-node cordoned-skip disclosure) to `check-nvidia-smi` stdout, so a pass on a reduced node set is visible rather than silent (`#1668`). But that disclosure only reaches `TestResult.Stdout`/`.Message`, which `pkg/evidence/redact`'s default "minimal" policy omits from a published, signed bundle. So in the default signed-bundle path a 1-of-2-node pass was indistinguishable from a 2-of-2 pass, and an all-cordoned skip lost its reason — for the very artifact a downstream consumer verifies by default.

Fixes: #1951
Related: #1936, #1668

## Type of Change

## Component(s) Affected

- [x] Other: `pkg/evidence/redact`, `validators/`

## Implementation Notes

- **`ctrf.TestResult.Extra` (`map[string]string`)** mirrors the CTRF spec's `extra` object. `ctrf.ExtraLinePrefix` (`##AICR-EXTRA## `) is the shared transport contract between the in-pod producer and the orchestrator consumer; it lives in `pkg/validator/ctrf` because both sides import it.
- **Producer** — `validators.EmitExtra(map[string]string)` marshals a check's outcome data to one prefixed stdout line. **Consumer** — `pkg/validator/job.ExtractResult` parses the last such line into `TestResult.Extra` and strips all prefixed lines from human-readable `Stdout`.
- **Redaction** — `redact.CTRF` rebuilds `Extra` against a fail-closed allowlist (`ctrfExtraAllowlist`: `nodesValidated`, `nodesTotal`, `skipReason`). Unknown keys (including any a future check adds) are dropped; an empty result ships no `extra: {}`. `PolicyVersion` bumps `v1` → `v2`.
- **`checkNvidiaSMI`** now populates `Extra` with its coverage counts / skip code; `validators/helper.FindAllGpuNodes` computes the candidate-node total.
- **Contract** (enforced by godoc + the allowlist): `Extra` values must be low-cardinality counts or enum codes only — never node names, IPs, or hostnames. The default treatment of free-form `Stdout`/`Message` is unchanged (a stated non-goal of the issue).

## Testing

```bash
make qualify
```

- `golangci-lint` on all changed packages — **0 issues**
- `go test -race ./pkg/validator/... ./pkg/evidence/... ./validators/...` — **PASS**
- `make test-coverage` — **80.8%** (threshold 80%) PASS
- `make qualify` — PASS end-to-end (chainsaw e2e 23/0/0; grype: only pre-existing Low/Unknown transitive advisories, non-gating)

New exported funcs covered: `EmitExtra` 75.0%, `FindAllGpuNodes` 87.5%.

## Risk Assessment

- [x] **Low** — Additive, fail-closed, well-tested, easy to revert.

**Rollout notes:** Backwards compatible. `Extra` is `omitempty`; existing checks that don't emit it are unaffected. Redaction fails closed, so no new data leaks into signed bundles. `PolicyVersion` bump (`v2`) lets verifiers tell which rules ran.

## Checklist

### DIFF
--- a/validators/deployment/nvidia_smi.go
+++ b/validators/deployment/nvidia_smi.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"strconv"
 	"strings"
 
 	"github.com/NVIDIA/aicr/pkg/defaults"
@@ -33,6 +34,17 @@ const (
 	nvidiaSMIPodTemplateFile = "testdata/nvidia-smi-verify-pod.yaml"
 	gpuCheckSuccessMsg       = "GPU_CHECK_SUCCESS"
 	nvidiaSMILogContextLines = 20
+
+	// skipReason* are the low-cardinality enum codes emitted via EmitExtra when
+	// the check skips. Unlike the human-readable RESULT/enumeration stdout lines
+	// (which the default redaction policy strips), these survive minimal
+	// redaction (see pkg/evidence/redact), so a signed bundle records WHY a check
+	// was skipped. Each must accurately describe the observed condition and be
+	// mirrored in redact.ctrfSkipReasons — a code missing from that closed set is
+	// dropped from the published bundle.
+	skipReasonNoGPUNodes            = "no-gpu-nodes"             // cluster has no GPU nodes at all
+	skipReasonNoSchedulableGPUNodes = "no-schedulable-gpu-nodes" // GPU nodes exist but all cordoned/unschedulable
+	skipReasonNodesBusy             = "nodes-busy"               // schedulable GPU nodes exist but are busy with workloads
 )
 
 // gpuNodeCoverage partitions check-nvidia-smi's discovered GPU nodes into the
@@ -126,18 +138,26 @@ func checkNvidiaSMI(ctx *validators.Context) error {
 
 	if len(allNodes) == 0 {
 		printLines(coverage.coverageLine(0))
+		emitExtraOrWarn(nvidiaSMISkipExtra(skipReasonNoGPUNodes))
 		return validators.Skip("no GPU nodes found in the cluster")
 	}
 
 	gpuNodes := coverage.schedulable
 	if len(gpuNodes) == 0 {
 		printLines(coverage.coverageLine(0))
+		// GPU nodes exist but all cordoned — a distinct, accurate code, never a
+		// false "no-gpu-nodes" in the signed evidence.
+		emitExtraOrWarn(nvidiaSMISkipExtra(skipReasonNoSchedulableGPUNodes))
 		return validators.Skip(fmt.Sprintf(
 			"all %d GPU node(s) are cordoned; nothing to verify", len(coverage.cordoned)))
 	}
 
 	// Check if any nodes are busy
+	// A probe error is treated as busy (fail-safe: don't run on a node we can't
+	// clear), but tracked separately from CONFIRMED occupancy so an all-errors
+	// skip is not signed as "nodes-busy".
 	var busyNodes []string
+	confirmedBusy := false
 	for _, node := range gpuNodes {
 		busy, busyErr := helper.IsNodeGpuBusy(ctx.Ctx, ctx.Clientset, node.Name)
 		if busyErr != nil {
@@ -147,11 +167,18 @@ func checkNvidiaSMI(ctx *validators.Context) error {
 		}
 		if busy {
 			busyNodes = append(busyNodes, node.Name)
+			confirmedBusy = true
 		}
 	}
 
 	if len(busyNodes) > 0 {
 		printLines(coverage.coverageLine(0))
+		// Only sign nodes-busy when at least one node was CONFIRMED occupied; a
+		// skip driven solely by probe errors must not masquerade as occupancy in
+		// the signed evidence (the error is already logged above).
+		if confirmedBusy {
+			emitExtraOrWarn(nvidiaSMISkipExtra(skipReasonNodesBusy))
+		}
 		return validators.Skip(fmt.Sprintf("GPU nodes busy with existing workloads: %v", busyNodes))
 	}
 
@@ -181,6 +208,11 @@ func checkNvidiaSMI(ctx *validators.Context) error {
 		validated++
 	}
 
+	// Emit the structured coverage disclosure (survives minimal redaction)
+	// alongside the human RESULT line, on BOTH the pass and fail paths, so a
+	// signed bundle records how many of the cluster's GPU nodes were verified.
+	emitExtraOrWarn(nvidiaSMICoverageExtra(validated, coverage.total))
+
 	if len(failedNodes) > 0 {
 		printLines(coverage.coverageLine(validated))
 		return errors.New(errors.ErrCodeInternal,
@@ -191,6 +223,31 @@ func checkNvidiaSMI(ctx *validators.Context) error {
 	printLines(coverage.coverageLine(validated))
 	fmt.Printf("Successfully verified GPU on all %d schedulable node(s)\n", len(gpuNodes))
 	return nil
+}
+
+// nvidiaSMICoverageExtra builds the structured coverage disclosure: how many GPU
+// nodes passed verification (validated) out of the total present incl. cordoned
+// (from the same node snapshot). Unlike the RESULT stdout line it survives
+// minimal redaction. Values are counts only — never node names or IPs.
+func nvidiaSMICoverageExtra(validated, total int) map[string]string {
+	return map[string]string{
+		"nodesValidated": strconv.Itoa(validated),
+		"nodesTotal":     strconv.Itoa(total),
+	}
+}
+
+// nvidiaSMISkipExtra builds the structured skip disclosure carrying only the
+// low-cardinality reason enum code.
+func nvidiaSMISkipExtra(reason string) map[string]string {
+	return map[string]string{"skipReason": reason}
+}
+
+// emitExtraOrWarn emits structured extra evidence, logging (never failing) on
+// error — a failed stdout write must not flip the check's verdict.
+func emitExtraOrWarn(extra map[string]string) {
+	if err := validators.EmitExtra(extra); err != nil {
+		slog.Warn("failed to emit validator extra", "error", err)
+	}
 }
 
 func verifySingleGPUNode(ctx *validators.Context, nodeName string) error {
--- a/docs/contributor/validator.md
+++ b/docs/contributor/validator.md
@@ -192,9 +192,93 @@ validators.Run(map[string]validators.CheckFunc{
 
 | Channel | Captured as |
 |---------|-------------|
-| **stdout** | CTRF `message` (human-readable evidence) — use `fmt.Printf` |
+| **stdout** | CTRF `stdout` (human-readable evidence) — use `fmt.Printf` |
 | **stderr** | Streamed live to the user — use `slog.*` |
 | `/dev/termination-log` | Failure reason (≤ 4096 bytes), written on `return error` |
+| **stdout sentinel lines** | Structured/side-channel data — see [Stdout sentinels](#stdout-sentinels) |
+
+### Stdout sentinels
+
+A check runs inside a pod; the only channels that reach the orchestrator are the
+exit code, `/dev/termination-log`, and the captured stdout (pod logs). To move
+*more* than a pass/fail verdict across that boundary, checks emit **sentinel
+lines** — stdout lines with a reserved prefix that a specific parser recognizes
+and pulls out. Plain (non-prefixed) stdout is unaffected and flows into the CTRF
+`stdout` array as before.
+
+Two sentinels exist today, with deliberately different lifecycles:
+
+| Prefix | Emitted by | Parsed by | Payload | Kept in CTRF `stdout`? | Survives minimal redaction? |
+|--------|-----------|-----------|---------|------------------------|-----------------------------|
+| `RESULT:` + one space | any check (`fmt.Printf`) | `extractResultSummaries` (`pkg/validator/validator.go`) | free-form human text (throughput, bandwidth, TTFT…) | **yes** — line stays; trailing text is *also* echoed to the live CLI at INFO | **no** — dies with `stdout` under the default policy |
+| `##AICR-EXTRA##` + one space | `validators.EmitExtra` | `parseExtraSentinels` (`pkg/validator/job/result.go`) | one JSON object → `TestResult.Extra` (counts / enum codes) | **no** — stripped as transport, not evidence | **yes** — allowlisted keys are published (see below) |
+
+Both are parsed with `strings.CutPrefix` and both parsers are pure, unit-tested
+functions. They are separate channels on purpose: `RESULT:` surfaces live
+metrics to a human watching a run, so it carries unbounded free-form text and is
+correctly redacted with the rest of `stdout`; `##AICR-EXTRA##` carries structured
+data that must *outlive* redaction, so it is low-cardinality, allowlisted, and
+stripped from the human evidence. Do not route structured outcome data through
+`RESULT:` (it would not survive publication) or human prose through
+`##AICR-EXTRA##` (it would be dropped by the allowlist or leak identifiers).
+
+### Structured evidence that survives redaction
+
+`stdout` and `message` are free-form log text: they can leak node names, IPs,
+DNS names, and secret/cert names, so the default **minimal** redaction policy
+(`pkg/evidence/redact`) strips them from a published, signed evidence bundle —
+only `--full` keeps them. A check that reports on `stdout` alone therefore shows
+as `passed` with no detail in the artifact a downstream consumer verifies by
+default: a 1-of-2-node pass is indistinguishable from 2-of-2, and a skip loses
+its reason.
+
+To carry outcome data that *does* survive minimal redaction, emit it into the
+CTRF `extra` map (`ctrf.TestResult.Extra`, mirroring the CTRF spec's `extra`
+object) via `validators.EmitExtra`:
+
+```go
+// Emit failures are non-fatal — warn, but never flip the check's verdict on a
+// failed stdout write (checkNvidiaSMI wraps this as the emitExtraOrWarn helper):
+emit := func(extra map[string]string) {
+    if err := validators.EmitExtra(extra); err != nil {
+        slog.Warn("failed to emit validator extra", "error", err)
+    }
+}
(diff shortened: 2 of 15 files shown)


njhensley:
Addressed CodeRabbit's review in the amended commit. Disposition:

**Applied (7):**
- **Validate `Extra` values, not just keys** (redact.go, types.go) — the allowlist is now key **and** value: `nodesValidated`/`nodesTotal` must be a bare non-negative decimal, `skipReason` a lowercase kebab-case enum code. An identifier smuggled under an allowed key (an IP in `nodesTotal`, a hostname in `skipReason`) is dropped at the publication boundary. Added adversarial-value test cases.
- **Keep the last _valid_ sentinel** (result.go) — `parseExtraSentinels` now parses each sentinel as encountered and retains the last valid non-empty payload, so a valid coverage line followed by a malformed one no longer clears the evidence. Added a `valid → malformed` case.
- **Defensively copy `Extra`** (builder.go) — `AddResult` clones the map; strengthened the test to mutate the source after insertion.
- **No fabricated / omitted coverage** (nvidia_smi.go) — the denominator is computed before verification, coverage is emitted on **both** pass and fail paths (counting only nodes that passed), and `nodesTotal` is omitted rather than fabricated when node discovery fails. Added `totalKnown=false` and partial-failure cases.
- **Table-driven allowlist test** (redact_test.go) — refactored into named cases covering surviving keys, dropped unknowns, invalid values under allowed keys, and full-drop-to-nil.
- **MD038** (validator.md) — removed trailing spaces from inline code spans.

**Not applied (2), with reasoning:**
- **`fmt.Printf` → `slog` on the success line** — that line is pre-existing (not introduced here), and this check deliberately uses `fmt.Printf` as its stdout **evidence** channel (the whole function does). Routing one line through `slog` would move it off the evidence transport and be inconsistent with the surrounding code.
- **`ctx.Done()` in `FindAllGpuNodes`’ filter loop** — it iterates an already-fetched, in-memory node list (the `List` call is already ctx-bounded), and its sibling `FindSchedulableGpuNodes` intentionally does the same without a per-iteration ctx check. Adding it to only one of the two is inconsistent for negligible benefit.

`make qualify` passes (lint 0 issues, `-race` tests, coverage 80.9% ≥ 80%).

njhensley:
Follow-up: fixed the doc `EmitExtra` examples to model the warn-but-preserve-verdict error handling (matching `emitExtraOrWarn` in `nvidia_smi.go`) instead of `_ = ...`. Doc-only change; amended into the commit.

njhensley:
Second review round addressed (all 4 applied):

- **Closed `skipReason` allowlist** (redact.go) — replaced the kebab-case *regex* with a closed set `ctrfSkipReasons` (`no-gpu-nodes`, `no-schedulable-gpu-nodes`, `nodes-busy`). You're right that a shape regex would still pass an arbitrary low-cardinality identifier like `customer-prod-cluster`; a well-formed but unlisted code is now dropped. Added test cases for a listed-code survivor and an unlisted-code drop. (This reverses my earlier "structural regex is enough" call — the closed set is the correct fail-closed boundary for signed evidence.)
- **Accurate skip codes** (nvidia_smi.go) — zero *schedulable* GPU nodes no longer misreports as `no-gpu-nodes` when the cluster actually has cordoned GPU nodes: it now emits `no-schedulable-gpu-nodes` (checked via `FindAllGpuNodes`) with a matching human message. On the busy path I kept `nodes-busy`: a busy-check lookup error is *already* folded into "treat as busy" as pre-existing fail-safe behavior, and the code accurately reflects the "unavailable, treated as busy" verification decision (the underlying error is logged).
- **Assert node identities** (gpu_test.go) — `FindAllGpuNodes`/`FindSchedulableGpuNodes` tests now assert the exact node-name sets (cordoned included only in the all-nodes result; cpu-only excluded from both), not just counts.
- **Doc stdout mapping** (validator.md) — corrected the channel table: pod stdout → CTRF `stdout` only (`message` comes from `/dev/termination-log`).

`make qualify` passes (lint 0 issues, `-race` tests, coverage 80.9% ≥ 80%).

njhensley:
Applied the multi-persona review. Disposition of all 9 findings:

**Applied (8):**
- **F1** (false `no-gpu-nodes` on transient List error) — the zero-schedulable branch now fails **closed**: on a `FindAllGpuNodes` error it emits **no** `skipReason` rather than a fabricated one (an absent reason is honest; a wrong one is signed misinformation). Logic extracted into a pure `nvidiaSMIZeroScheduleOutcome` helper.
- **F4** (headline branch logic untested) — new table-driven test on that helper covers all three outcomes incl. the F1 fail-closed path (100% covered). Note: the full `checkNvidiaSMI` verify path stays cluster-dependent/pre-existing 0%, but the fragile decision logic is now locked down.
- **F5** (parse-before-truncate wiring untested) — extracted `processValidatorLogs` (used by both `ExtractResult` and `HandleTimeout`, de-duplicating them) with a test that emits the sentinel *after* >`ValidatorMaxStdoutLines` of output and asserts `Extra` survives (100% covered).
- **F3** (user docs silent on surviving keys) — added the preserved-`extra` disclosure to `docs/user/artifact-verification.md` and `docs/user/validation.md`, linking the contributor contract.
- **F2 + F9** (count regex too wide / magic literal) — tightened to a named `maxCountDigits = 5` const (≤ 99,999) with a drop-test for over-long counts.
- **F6** (prefix-boundary edges) — added `parseExtraSentinels` cases for a mid-line prefix (kept) and a prefix missing its trailing space (not a sentinel).
- **F7** (EmitExtra error branch) — test points `extrasOut` at an always-erroring writer, asserting a wrapped `ErrCodeInternal` (EmitExtra 75% → 87.5%).

**Not applied (1):**
- **F8** (two `Nodes().List` calls on the happy path) — left as-is. It's explicitly optional; a combined helper would churn the public `FindSchedulableGpuNodes`/`FindAllGpuNodes` API for one saved List per run, and correctness is already guarded by the `len(allNodes) > total` between-calls check. Noted for a possible follow-up.

`make qualify` passes (lint 0 issues, `-race` tests, coverage 80.9% ≥ 80%).

validators/deployment/nvidia_smi.go
```diff
@@ -66,11 +89,26 @@ func checkNvidiaSMI(ctx *validators.Context) error {
 	}
 
 	if len(busyNodes) > 0 {
+		emitExtraOrWarn(nvidiaSMISkipExtra(skipReasonNodesBusy))
```
yuanchen8911: `busyNodes` is populated on both confirmed occupancy (`busy=true`) and Pods LIST errors (L82–84). A Pods LIST failure — for example API throttling, a transport timeout, or an apiserver connection reset — therefore signs `skipReason: nodes-busy` in the evidence bundle with no confirmed workload.

Track probe failures separately from confirmed occupancy. Call `emitExtraOrWarn(nvidiaSMISkipExtra(skipReasonNodesBusy))` only when at least one successful query returned `busy=true`. If the only evidence is error returns, either omit the skip reason or surface the probe error rather than masking it as node occupancy.

validators/deployment/nvidia_smi.go
```diff
@@ -43,7 +54,19 @@ func checkNvidiaSMI(ctx *validators.Context) error {
 	}
 
 	if len(gpuNodes) == 0 {
-		return validators.Skip("no GPU nodes found in the cluster")
+		// Zero SCHEDULABLE GPU nodes is not the same as no GPU nodes: the cluster
+		// may hold only cordoned ones. Decide the reason from the cordoned-
+		// inclusive count — and if that discovery errors, emit NO reason rather
+		// than a fabricated "no-gpu-nodes" that would become signed misinformation.
+		allNodes, allErr := helper.FindAllGpuNodes(ctx.Ctx, ctx.Clientset)
```
yuanchen8911: The zero-schedule classification is derived from two independent Node LIST snapshots (this call and `FindSchedulableGpuNodes` at L51). During a cordon/uncordon or device-plugin rollout — both supported cluster operations — the schedulable query can return 0 while this snapshot sees a node that has since become schedulable. `nvidiaSMIZeroScheduleOutcome` then signs `skipReason: no-schedulable-gpu-nodes` for a check that was runnable at the time of classification.

Enumerate all GPU nodes once with `FindAllGpuNodes`, partition that single result into schedulable and unschedulable, and drive node selection, zero-schedule classification, and the coverage denominator from the same snapshot. This eliminates the TOCTOU window and, as a side effect, eliminates the second-discovery-error branch in finding #3.

validators/deployment/nvidia_smi.go
```diff
@@ -102,6 +147,53 @@ func checkNvidiaSMI(ctx *validators.Context) error {
...
+// cordoned-inclusive GPU node count and allErr its discovery error.
+//
+// On discovery error it returns emit=false: an absent skipReason is honest,
+// whereas a guessed "no-gpu-nodes" would be signed misinformation when the
+// cluster may actually hold only cordoned GPU nodes. Otherwise it distinguishes
+// "GPU nodes exist but all cordoned" from a genuine absence of GPU nodes.
+func nvidiaSMIZeroScheduleOutcome(allCount int, allErr error) (reason string, emit bool, msg string) {
+	switch {
+	case allErr != nil:
+		return "", false, "no GPU nodes found in the cluster"
```
yuanchen8911: `emit=false` is correct: no skip reason is signed when discovery fails. However, `validators.Skip(msg)` writes `msg` to the termination log, and `ExtractResult` surfaces it as `TestResult.Message`. That message is retained in live, raw, and `--full` CTRF evidence. The claim "no GPU nodes found in the cluster" is inaccurate: the code knows only that the first snapshot contained no schedulable GPU nodes; whether cordoned GPU nodes exist is unknown because the second LIST failed.

If finding #2 is fixed (single snapshot, partition), this branch is eliminated. Otherwise change the message to something uncertainty-aware, e.g. `"no schedulable GPU nodes found; unable to determine whether cordoned GPU nodes exist"`.

yuanchen8911 (changes requested):
Three issues in `validators/deployment/nvidia_smi.go`. Two block merge (#1, #2); #3 is resolved by the #2 fix or by correcting the wording.

njhensley:
Thanks @yuanchen8911 — all three addressed in the latest push (`00c6f650`). These caught real evidence-integrity bugs the automated passes missed.

**#1 — TOCTOU between two Node LISTs (blocking).** Fixed as you prescribed: added `helper.FindGpuNodes`, which lists nodes **once** and partitions into `Schedulable`/`All` from a single snapshot. `checkNvidiaSMI` now drives node selection, the zero-schedule classification, and the coverage denominator from that one view — no second LIST, so the cordon/uncordon and device-plugin-rollout races are gone. `FindAllGpuNodes` (the second-LIST helper this PR had added) is removed; `FindSchedulableGpuNodes` stays for its other callers and is asserted to agree with the partition.

**#2 — probe errors masked as `nodes-busy` (blocking).** Now tracked separately: a `IsNodeGpuBusy` error still makes the node count as busy (fail-safe skip), but `skipReason: nodes-busy` is emitted **only** when at least one node was *confirmed* occupied (`busy=true`). A skip driven solely by probe errors signs no reason (the error is logged), consistent with the fail-closed treatment of the other skip paths.

**#3 — inaccurate "no GPU nodes found" message.** Eliminated by #1: with a single snapshot there is no second-LIST-failure branch, so the misleading message can no longer be produced. The cordoned-only case accurately reads `no schedulable GPU nodes: N GPU node(s) present but unschedulable (cordoned)`.

Side effects: `nvidiaSMIZeroScheduleOutcome` simplified (no discovery-error arg) and `nvidiaSMICoverageExtra` dropped its `totalKnown` param since the denominator is now always known from the snapshot — both remain pure and 100%-covered. `make qualify` passes (lint 0, `-race`, coverage 80.9%).

validators/deployment/nvidia_smi.go
yuanchen8911: Addressed in 00c6f650 — verified: confirmedBusy gates the nodes-busy extra, so an all-probe-errors skip no longer signs occupancy in the evidence. Optional cosmetic leftover: the human-visible Skip message still says busy with existing workloads on the all-errors path; fine to leave, or reword to mention the probe failures. Resolving.

validators/deployment/nvidia_smi.go
yuanchen8911: Addressed in 00c6f650 — verified: FindGpuNodes returns a single-snapshot GpuNodePartition and node selection, zero-schedule classification, and the coverage denominator all derive from it. The TOCTOU window is gone. Resolving.

validators/deployment/nvidia_smi.go
yuanchen8911: Resolved by the finding-2 fix as anticipated: the second-discovery-error branch is eliminated and nvidiaSMIZeroScheduleOutcome now distinguishes all-cordoned from genuinely-no-GPU-nodes accurately, with both outcomes asserted in TestNvidiaSMIZeroScheduleOutcome. Resolving.

yuanchen8911 (dismissed):
00c6f650 verified — both blockers fixed: confirmedBusy gates the nodes-busy evidence signing so probe errors no longer masquerade as occupancy, and the single-snapshot GpuNodePartition drives node selection, zero-schedule classification, and the coverage denominator, eliminating the TOCTOU window. Finding 3 dissolved with the second LIST as anticipated, and TestNvidiaSMIZeroScheduleOutcome asserts both accurate outcomes. Approving.

njhensley:
Rebased onto `main` now that **#1936 has merged** — the overlapping PR I flagged earlier. This significantly slims this PR and resolves the overlap cleanly:

- **#1936 already landed the single-snapshot, TOCTOU-safe design** (`helper.FindGpuNodes` returning `[]GpuNode{Node, Cordoned}`, partitioned by `partitionGpuNodes`, with `ctx.Done()` checks) **and** the human-readable coverage disclosure (`RESULT: nodesValidated: X/Y (Z cordoned, skipped)` + per-node `skipped (cordoned)` stdout lines). That's exactly the structure @yuanchen8911's TOCTOU finding (#1) and CodeRabbit's F9 (ctx cancellation) asked for — so I **dropped** my now-redundant `GpuNodePartition`/`FindGpuNodes` and `nvidiaSMIZeroScheduleOutcome`, and use #1936's.
- **This PR's remaining, non-overlapping contribution is the redaction-surviving structured `Extra`** — precisely the gap #1951 was filed against (#1936's disclosure is stdout-only and is stripped from the default signed bundle). I layered `EmitExtra` onto #1936's exit points: `skipReason` at each skip (`no-gpu-nodes` / `no-schedulable-gpu-nodes` / `nodes-busy`) and `nodesValidated`/`nodesTotal` on both pass and fail, all driven from #1936's single `gpuNodeCoverage` snapshot.
- **My earlier review fixes are preserved** on top of the new base: the accurate cordoned-vs-absent skip code, the confirmed-busy tracking (a probe-error-only skip no longer signs `nodes-busy`), the closed `ctrfSkipReasons` allowlist + value validation in redact, the parse-before-truncate helper, and the user-doc disclosure.

Net: `validators/helper/gpu.go` reverts entirely to #1936's version; `checkNvidiaSMI` is #1936's structure plus the `Extra` emission layer. `make qualify` passes (lint 0, `-race`, coverage 80.9%). Single signed commit, now current on `main`.

yuanchen8911 (approved):
Re-approving after the rebase onto merged #1936 (the earlier approval was auto-dismissed by the force-push). Verified at 88ee46fd:

- The plumbing files (ctrf types/builder, redact, extras) are byte-identical to the previously approved head, and the nvidia_smi adoption is rebuilt cleanly on #1936's partition with all three earlier fixes preserved (single-snapshot classification, confirmedBusy gating, accurate skip messages).
- The only new delta is the result.go sentinel transport, which is sound: Extra is parsed from the full logs before tail-truncation, malformed sentinel lines are skipped without discarding an earlier valid payload, sentinels are stripped from human stdout, and both extraction paths are covered.

On the two open CodeRabbit threads: the value-format enforcement one is already implemented — redact.go's ctrfExtraAllowlist is fail-closed on key AND value (bounded counts, closed skip-code enum) at the publication boundary, so the thread is stale against current code. The fmt.Printf one misapplies the logging rule: validator-check stdout is the CTRF evidence transport; slog would remove the line from evidence. Neither blocks.

PR #1980, opened 2026-07-31
fix(validator): resolve deferred Job watch resume review feedback

### LINKED ISSUE #1976: Follow-up (#1971): resolve deferred review feedback on Job watch resume
## Summary

Follow-up to PR #1971 (Job watch re-establishment after apiserver idle-stream closure). The PR was approved, but five reviewer threads were deferred rather than resolved in-branch. One is a genuine correctness defect introduced by the PR's termination-message bounding; the rest are correctness-adjacent and code-quality quick wins. Tracking them here so #1971 can land.

All references are file paths in `pkg/k8s/pod` and `pkg/validator/job` as of PR #1971; line numbers are hints and may drift.

### 1. `boundTerminationMsg` can erase the entire message (correctness, higher priority)

`pkg/validator/job/result.go` — the rune-boundary trim loop validates the **whole** prefix (`for len(head) > 0 && !utf8.ValidString(head) { head = head[:len(head)-1] }`). A single invalid UTF-8 byte anywhere *before* the cut point makes every prefix invalid, so the loop trims `head` to empty and the result carries only the `... [truncated N bytes]` suffix — the readable message is discarded.

- Fix: trim only a trailing partial rune (e.g. `utf8.DecodeLastRuneInString`, treating `RuneError` with size 1 as the boundary case); preserve earlier bytes.
- Container termination messages are attacker-influenceable in principle and not guaranteed valid UTF-8, so this is reachable, not purely theoretical.

### 2. Truncation tests cannot detect a wiped message (test gap; same root as #1)

`pkg/validator/job/result_test.go` — the mid-rune case asserts only that the output is valid UTF-8 and contains `... [truncated`. An implementation that discards the whole prefix satisfies both. Strengthen: assert the surviving prefix is non-empty and the dropped-byte count is small, and add a case whose input holds an invalid byte before `maxBytes`.

### 3. Container-not-found branch always blames the catalog timeout (correctness)

`pkg/validator/job/result.go` `HandleTimeout` — the `!found` (validator container absent) branch renders `"timeout: validator did not complete within %s ..."` regardless of `waitCause`. When the wait failed because the apiserver watch broke (infra/unavailable) and the container status is absent, this misdirects diagnosis to the configured timeout — the exact misreporting the PR removes in the sibling `else` branch. Route this message through `waitFailureMessage` too, keeping the container-contract detail as a suffix.

### 4. Duplicate stale-RV apiserver model in tests (maintainability)

`pkg/k8s/pod/wait_termination_test.go` re-declares `freshCollectionRV` (shadowing the file-scope constant) and re-implements the stale-RV 410 reactor already provided by `staleRVWatchReactor` in `pkg/k8s/pod/job_test.go`. Two independent models of the same apiserver 410 behavior can drift; reuse the shared helper (it already accepts the `onFresh` Job).

### 5. Stale "re-check Get" wording in `resumeJobWatch` godoc (docs)

`pkg/k8s/pod/job.go` — the `resumeJobWatch` doc block still says it "re-check Get"s the Job and returns `ErrCodeUnavailable` "when a re-check Get fails". The implementation now resyncs via a field-selected **List** and emits `"job watch closed and Job re-list failed"`. Align the wording (resync List / replacement Watch).

## Acceptance

- [ ] `boundTerminationMsg` preserves the readable prefix when invalid UTF-8 precedes the cut; regression test with an invalid byte before `maxBytes`.
- [ ] Container-not-found timeout branch reflects the actual `waitCause`.
- [ ] `wait_termination_test.go` reuses `staleRVWatchReactor` instead of a duplicated model.
- [ ] `resumeJobWatch` godoc matches the List-based implementation.
- [ ] `make test` (`-race`) and `golangci-lint` green on `pkg/k8s/pod` and `pkg/validator/job`.

njhensley:
## Summary

Resolves the five deferred reviewer threads from #1971 (now merged), tracked in #1976: one genuine correctness defect in the termination-message bounding plus four correctness-adjacent / code-quality quick wins in `pkg/validator/job` and `pkg/k8s/pod`.

## Motivation / Context

#1971 landed the Job-watch re-establishment fix with five reviewer threads deferred rather than resolved in-branch. This follow-up closes them.

Fixes: #1976
Related: #1971, #1966

## Type of Change

## Component(s) Affected

## Implementation Notes

- **`boundTerminationMsg` no longer erases the message (correctness).** The old rune-boundary trim validated the *whole* prefix (`for … !utf8.ValidString(head)`), so a single invalid UTF-8 byte anywhere before the cut trimmed `head` to empty and emitted only the `... [truncated N bytes]` suffix. Now trims just a trailing partial rune via `utf8.DecodeLastRuneInString` (RuneError + size 1 = boundary case); earlier bytes — including a pre-existing invalid byte — are preserved. Container termination messages are attacker-influenceable and not guaranteed valid UTF-8, so this is reachable.
- **Container-not-found timeout branch reflects the actual cause (correctness).** `HandleTimeout`'s `!found` branch previously always rendered the catalog-timeout wording regardless of `waitCause`. It now routes through `waitFailureMessage`, so an infra/unavailable cause is surfaced verbatim (`validation failed: …`) with the container-contract detail kept as a suffix — mirroring the sibling `else` branch #1971 already fixed.
- **Test duplication removed (maintainability).** `wait_termination_test.go` now reuses the shared `staleRVWatchReactor` helper and file-scope `freshCollectionRV` instead of re-declaring a shadowed constant and re-implementing the stale-RV 410 model.
- **Godoc / comment alignment (docs).** `resumeJobWatch` / `WaitForJobTerminal` godoc and the watch-loop comments no longer say "re-check Get"; they now describe the field-selected List resync the implementation actually performs.

## Testing

```bash
go test -race ./pkg/k8s/pod/... ./pkg/validator/job/...   # ok
golangci-lint run -c .golangci.yaml ./pkg/k8s/pod/... ./pkg/validator/job/...   # 0 issues
```

Strengthened `TestBoundTerminationMsg` (asserts non-empty surviving prefix + small partial-rune trim, adds an invalid-byte-before-`maxBytes` regression case) and added `TestHandleTimeoutContainerNotFoundReflectsCause`. Verified the regression is caught: reverting the `boundTerminationMsg` fix fails the new case (`trimmed 98 bytes back from maxBytes=100`).

## Risk Assessment

- [x] **Low** — Isolated to two packages, no public API change, fully unit-tested, easy to revert.

**Rollout notes:** N/A — backwards compatible; only the rendered termination/timeout message text changes for the affected edge cases.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A — internal messages only)


yuanchen8911 (approved):
LGTM. Verified both correctness fixes against the surrounding code: the `boundTerminationMsg` trailing-partial-rune trim behaves correctly across the edge cases (mid-rune cut, lone `0xFF` at the boundary, pre-existing invalid bytes preserved), and the `!found` branch now mirrors the sibling branch via `waitFailureMessage`, which handles a nil cause as a timeout. The new regression assertions would catch a reintroduction of the whole-prefix validation.

Two minor, non-blocking notes:

1. The PR body's Implementation Notes describe the fix as trimming "via `utf8.DecodeLastRuneInString` (RuneError + size 1 = boundary case)", but the shipped code uses a manual `utf8.RuneStart` back-scan plus `utf8.FullRuneInString` — presumably wording from an earlier revision. A one-line body update would keep the description matching the merged mechanism.

2. Degenerate edge: if `maxBytes` were smaller than one multi-byte rune (e.g., 2 with a leading `€`), the entire head is a partial rune and the output degrades to suffix-only. Unreachable with the current constant caller, so fine as-is — just noting it's the one input class the "never erase the prefix" guarantee doesn't cover.

PR #1981, opened 2026-07-31
fix(validator): disclose cordoned RDMA nodes in expected-resources

### LINKED ISSUE #1952: Other node-scoped checks silently narrow to schedulable GPU nodes without disclosure
## Context

`#1668`/`#1936` fixed `check-nvidia-smi`: it now enumerates every GPU node via `helper.FindGpuNodes`, validates only the schedulable subset, and explicitly discloses cordoned nodes plus a `nodesValidated` coverage line instead of silently narrowing scope.

The same undisclosed narrowing exists in at least one other node-scoped probe that was out of scope for `#1936`:

- **`rdmaFabricProbe`** (`validators/deployment/expected_resources.go`, reached from `checkExpectedResources`) calls `helper.FindSchedulableGpuNodes` (line ~1007) to build its RDMA-capable cohort. On success it prints `RDMA fabric (%s): allocatable (uniform) on all %d RDMA GPU node(s)` where the count is the schedulable cohort only — a cordoned RDMA-capable node is silently excluded from both the cohort and the printed count, with no `skipped (cordoned)`-style disclosure. This is the same spuriously-narrowed-pass shape `#1668` was filed against.

There may be others worth auditing at the same time:
- `validators/performance/nccl_all_reduce_bw_constraint.go` and `validators/performance/inference_perf_constraint.go` also call `helper.FindSchedulableGpuNodes` to size workload placement. These are workload-sizing decisions rather than "verify every node" checks, so the same disclosure pattern may not directly apply, but worth a deliberate look rather than assuming.

## Proposal

Apply the same disclosure pattern established in `#1936` (`docs/contributor/validator.md`, "Node-scoped checks: cordoned GPU nodes" section) to `rdmaFabricProbe` at minimum, and audit the performance call sites to decide whether they need it too.

Found during review of `#1936`, which narrowed its own doc claim ("cluster-aggregate checks... are unaffected") to not overclaim coverage of `rdmaFabricProbe`.

njhensley:
## Summary

`rdmaFabricProbe` now enumerates RDMA GPU nodes via `helper.FindGpuNodes` and explicitly discloses cordoned Mellanox RDMA nodes (reported as `skipped (cordoned)` and counted in `nodesTotal`) instead of silently narrowing scope to the schedulable subset. Coverage is surfaced through `EmitExtra` so it survives redaction.

## Motivation / Context

`#1668`/`#1936` fixed the same undisclosed narrowing in `check-nvidia-smi`. `rdmaFabricProbe` (reached from `checkExpectedResources`) had the identical spuriously-narrowed-pass shape: it built its RDMA cohort from `helper.FindSchedulableGpuNodes` and printed a count over the schedulable subset only, so a cordoned RDMA-capable node was silently excluded from both the cohort and the printed count with no disclosure. This applies the `#1936` disclosure pattern to `rdmaFabricProbe` and completes the audit called for in the issue.

Fixes: #1952
Related: #1936, #1668, #1951

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

## Implementation Notes

- **Disclosure + Extra wiring:** `rdmaFabricProbe` now enumerates every RDMA GPU node via `helper.FindGpuNodes`, validates only the schedulable subset, and discloses cordoned nodes as `skipped (cordoned)` while counting them in `nodesTotal`. The coverage is emitted via `EmitExtra` (`nodesValidated`/`nodesTotal`) so it survives redaction, following the `#1951` pattern.
- **PolicyVersion unchanged (v2):** the fix reuses the existing `nodesValidated`/`nodesTotal` allowlist keys, so no new redaction-allowlist entries are introduced and `PolicyVersion` stays at `v2`.
- **Performance call sites deliberately left unchanged:** `nccl_all_reduce_bw` and `inference_perf` also call `helper.FindSchedulableGpuNodes`, but they were audited and intentionally not modified — they assert workload placement/sizing, not node-coverage, so cordoned-node disclosure does not apply to them.

## Testing

```bash
unset GITLAB_TOKEN
golangci-lint run -c .golangci.yaml ./validators/deployment/... ./pkg/evidence/redact/...   # 0 issues
go test -race ./validators/deployment/... ./pkg/evidence/redact/...                          # ok
```

- Targeted lint (0 issues) and `go test -race` (pass) on both affected packages.
- New table-driven tests in `validators/deployment/expected_resources_test.go` cover the cordoned-RDMA-node disclosure path (`skipped (cordoned)` + `nodesTotal` accounting); `pkg/evidence/redact/redact_test.go` confirms the emitted coverage keys survive redaction.
- Full `make qualify` intentionally not run for this validator + docs change (e2e disproportionately heavy); the CI gate will run it.

## Risk Assessment

**Rollout notes:** No schema or PolicyVersion change (reuses existing v2 allowlist keys); backwards compatible. N/A migration.

## Checklist


njhensley:
Addressed both CodeRabbit findings. (1) Cancellation mid-scan now returns the accumulated coverage (cordoned nodes already seen) with schedulable set from the scanned cohort, instead of an empty partition — matching the function contract and all other error paths. (2) The FindGpuNodes error path: verified helper.FindGpuNodes returns a coded StructuredError (ErrCodeInternal on the node-List failure — the realistic path — and ErrCodeTimeout only on ctx cancel, which the in-loop listCtx.Done() branch surfaces on the accumulated coverage), so the ErrCodeInternal wrap adds gate context without clobbering the code; kept with a clarifying comment. Lint + go test -race ./validators/deployment/... green.

njhensley:
Applied CodeRabbit fix: FindGpuNodes error path now uses errors.PropagateOrWrap, so an inner ErrCodeTimeout (FindGpuNodes cancelling in its own node scan before this loop) propagates unchanged while a plain List failure still gets ErrCodeInternal. Updated TestRDMAFabricProbe_FailsClosedOnListError to assert the propagated ErrCodeInternal code + fail-closed behavior instead of the removed gate-context message string. Lint + go test -race ./validators/deployment/... green.

njhensley:
Thanks @coderabbitai — verified all three inline comments on \`bb384ad6\` against the current code. All three are non-actionable:

1. **\`expected_resources.go:1122\` — "return the existing structured error" (Major):** Not a defect. \`helper.FindGpuNodes\` returns a \`*errors.StructuredError\` on *every* error path (\`ErrCodeInternal\` on the List failure, \`ErrCodeTimeout\` on its own scan cancellation), so \`PropagateOrWrap\` preserves \`ErrCodeTimeout\` identically to a bare \`return err\`. The difference is only in the (currently unreachable) plain-error case, where \`PropagateOrWrap\` adds the gate-context message rather than leaking an uncoded error — strictly more robust. This is the deliberate resolution from the previous round; keeping it.

2. **\`expected_resources.go\` Done-branch — "cancellation discards accumulated coverage" (Major):** Already fixed / stale. The \`listCtx.Done()\` branch already returns the accumulated \`coverage\` (with \`coverage.schedulable = len(cohort)\`), not a zero \`rdmaFabricCoverage{}\`. This matches your own committable suggestion. No change needed.

3. **\`expected_resources_test.go\` — "add a cancellation mid-scan test" (nitpick):** Skipping. The \`listCtx.Done()\` branch in \`rdmaFabricProbeCoverage\` is not deterministically reachable through the public function: \`FindGpuNodes\` runs on the *same* \`listCtx\` and short-circuits first with an identical \`ctx.Done()\` guard, so any cancellation state trips there (→ \`ErrCodeTimeout\` via the path above) before the probe's own loop iterates. Exercising it directly would require adding a production cancellation seam purely for the test — unwarranted churn for a defensive \`select\`. The disclosure contract on the nil-error and fail-closed paths is already locked by \`TestRDMAFabricProbeCoverage_DisclosesCordoned\` and \`_CountsCordonedOnFailClosed\`.

yuanchen8911 (changes requested):
The fail-closed disclosure path has a gap: on a never-ready poll, the pod is killed before the terminal disclosure runs.

The catalog's 8m timeout feeds both the Job's activeDeadlineSeconds and AICR_CHECK_TIMEOUT with no margin (pkg/validator/v1/job_plan.go L229/L273/L303), and LoadContext starts its 8m context at process start (validators/context.go L100). Since GPUReadinessTimeout is also 8m, an exhausted poll returns at ~processStart+8m — after the Job deadline at jobStart+8m. With no SIGTERM handler, the process dies before printLines/EmitExtra execute, losing the enumeration and nodesValidated/nodesTotal exactly where a cordoned node narrowed coverage. Reachable in supported states: all RDMA nodes cordoned for maintenance, or a rollout slower than 8m.

Scope: the kill race predates this PR — every exhausted pollUntilStable gate already loses its timeout diagnostic this way, so the general no-margin defect should be a follow-up issue. But the every-terminal-outcome disclosure contract is new here, so the disclosure gap belongs in this PR.

Suggested fix: also emit the coverage disclosure once eagerly after the first probe observation, keeping the terminal emit. parseExtraSentinels keeps the last valid sentinel, so a clean exit's emit wins and a deadline kill still leaves the early one in the logs — no per-tick spam, no envelope changes.

njhensley:
@yuanchen8911 thanks — good catch on the mid-poll kill race. Addressed in `d57b74dc`.

**The gap:** on a never-ready poll, the terminal `emitExtraOrWarn` never runs when the Job's `activeDeadlineSeconds` SIGKILLs the process at the same no-margin 8m as the poll budget — so the cordoned-node coverage (the whole point of this gate) never reaches the signed bundle. Reachable exactly as you describe: all RDMA nodes cordoned for maintenance, or a rollout slower than the budget.

**Fix (your suggested approach):** `verifyRDMAFabricReady` now emits the structured coverage **eagerly, once**, on the first poll observation that enumerates any RDMA-candidate node, with `nodesValidated=0` (nothing is certified mid-poll), keeping the terminal emit. `parseExtraSentinels` keeps the last valid sentinel (verified in `pkg/validator/job/result.go`), so a clean exit's terminal emit wins and a deadline kill leaves the eager floor as the disclosure of record — no per-tick spam, no envelope changes.

**Scoping notes:**
- Only the **structured Extra** is emitted eagerly, not the stdout enumeration — the Extra is the piece that survives redaction into the signed bundle (#1951/#1952), and duplicating the stdout `RESULT:`/enumeration would produce divergent counts across the poll. Node names still appear only in the redacted stdout channel. Flag if you'd also like the stdout enumeration floored.
- The eager emit is wired through an injected seam (`verifyRDMAFabricReadyEmit`) so it's unit-testable without capturing the `validators`-package stdout transport — no mutable global.
- The broader no-margin kill race (every exhausted `pollUntilStable` gate loses its timeout diagnostic this way) is pre-existing and out of scope here; agree it belongs in a follow-up issue for the general `activeDeadlineSeconds` vs `AICR_CHECK_TIMEOUT` margin.

**Tests:** added `TestVerifyRDMAFabricReady_EagerDisclosureFloor` (table-driven, `-race`) — asserts the eager floor emits `{validated:0, total:2}` (cordoned node disclosed) *before* the terminal emit on both the fail-closed timeout and the ready-certifies paths. `docs/contributor/validator.md` updated to describe the eager-floor/last-wins semantics. `golangci-lint` clean, full `validators/deployment` package green under `-race`.

njhensley:
Follow-up in `03cab57e` (amended, still one commit):

**Removed the now-dead `rdmaFabricProbe` wrapper** (CodeRabbit nitpick, valid). This PR orphaned it: the poll switched from `rdmaFabricProbe` (int-returning) to `rdmaFabricProbeCoverage` (full partition), so the wrapper had no production callers left. Deleted it and migrated its 5 unit-test call sites to `rdmaFabricProbeCoverage(ctx)` reading `cov.schedulable` — matching the pattern `TestRDMAFabricProbeCoverage_DisclosesCordoned`/`_CountsCordonedOnFailClosed` already use. `golangci-lint` clean, `validators/deployment` green under `-race`.

The other three CodeRabbit comments on this commit are re-anchored duplicates of the first-round items already resolved/explained:
- **"return the existing structured error" (`PropagateOrWrap`)** — `FindGpuNodes` always returns a `*StructuredError`, so `PropagateOrWrap` preserves `ErrCodeTimeout` identically while defensively wrapping any future plain error with gate context. Deliberate; keeping it.
- **"cancellation discards accumulated coverage"** — already returns the accumulated `coverage` (CodeRabbit's own status now shows ✅ Addressed).
- **"add a cancellation mid-scan test"** — the `listCtx.Done()` branch isn't deterministically reachable through `rdmaFabricProbeCoverage`: `FindGpuNodes` shares the same `listCtx` and short-circuits first with an identical guard. Testing it would need a production cancellation seam, unwarranted for a defensive `select`.

njhensley:
Fixed the stale `rdmaFabricProbe` doc reference in `docs/contributor/validator.md` (now `rdmaFabricProbeCoverage`, matching the wrapper removal) in `60280cc2` — doc-only, still one commit. The other three CodeRabbit comments on the prior commit are the recurring first-round items (two now show ✅ Addressed; the cancellation-test nitpick remains not deterministically testable, as explained above).

yuanchen8911 (approved):
Issues resolved — the eager disclosure floor closes my finding. /lgtm

One follow-up: the comment defers the broader no-margin kill race (the catalog timeout feeding both activeDeadlineSeconds and AICR_CHECK_TIMEOUT with zero margin, which predates this gate) as "tracked separately", but I couldn't find an open issue for it. Please file one and point the comment at it.

PR #1982, opened 2026-07-31
feat(supply-chain): publish release metadata as signed OCI referrers

### LINKED ISSUE #1957: feat(supply-chain): publish all release metadata as cosign-signed OCI referrer attestations
## Problem

Release metadata is produced today by three independent mechanisms, with different signers, different formats, and different discovery paths:

| Metadata | How it ships | Signed by | Discoverable via |
|---|---|---|---|
| SLSA provenance (images) | `actions/attest-build-provenance` with `push-to-registry: true` (`.github/actions/sbom-and-attest/action.yml`) | GitHub attestation flow | OCI referrers + GitHub attestations API |
| SPDX SBOM (images) | `cosign attest --type spdxjson`, once per platform | cosign, **version unpinned** | depends on the cosign version installed |
| OpenVEX | not published at all | n/a | n/a |
| SLSA provenance (binaries) | `cosign attest-blob` post-hook in `.goreleaser.yaml` writing `aicr-attestation.sigstore.json` / `aicrd-attestation.sigstore.json` | cosign v3.1.2, Rekor v2 | GitHub release asset only |
| SPDX SBOM (binaries) | goreleaser `sboms: artifacts: binary` | **unsigned** | GitHub release asset |
| Recipe catalog | `aicr recipe sign-catalog` | Rekor v2 | loose release asset plus a copy in every archive |

A consumer who wants provenance, SBOM, and vulnerability posture for a released image has to know three retrieval methods, and one of the three has no retrieval method at all: `.openvex.json` is only ever passed to Grype as scanner input (`.github/workflows/on-tag.yaml`, `.github/workflows/vuln-scan-images.yaml`). Downstream scanners therefore re-flag CVEs that we have already triaged as not-affected, because our reachability analysis never leaves the repo.

Three concrete defects fall out of the same root cause:

1. **cosign is unpinned on the SBOM path.** `.github/actions/sbom-and-attest/action.yml` installs `sigstore/cosign-installer` with no `cosign-release` input and never calls `load-versions`, unlike `.github/workflows/on-tag.yaml`, which pins the `.settings.yaml` version (`v3.1.2`). cosign v3 defaults `--new-bundle-format=true`, which is precisely "attach a Sigstore bundle using OCI referrers API". So whether our SBOM attestations land as OCI 1.1 referrers or as legacy `.att` tags is currently decided by an installer default rather than by us. Either outcome is wrong to depend on implicitly, and it is not reproducible across runs if the installer's default moves.
2. **Per-platform SBOMs share one subject.** `.github/actions/attest-image-from-tag/action.yml` resolves the subject with `crane digest <image>:<tag>`, which returns the multi-arch index digest. Both the linux/amd64 and the linux/arm64 SPDX attestations are then attached to that same index digest, so they can only be told apart by reading the predicate body.
3. **Binary SBOMs are unsigned.** goreleaser uploads them as plain release assets with no attestation, while the binary provenance next to them is Rekor v2 signed.

## Goal

Every piece of release metadata is a cosign-signed in-toto attestation attached to its subject as an OCI 1.1 referrer, retrievable through one documented path.

## Scope

- **Publish OpenVEX.** Attach `.openvex.json` to each released image with `cosign attest --type openvex` (a natively supported predicate type, no custom URI needed). Keep the existing Grype `vex:` input as-is; this adds publication, it does not change scanning.
- **Pin cosign on the SBOM path.** Wire `load-versions` into `.github/actions/sbom-and-attest` and pass `cosign-release` from `.settings.yaml`, then set `--new-bundle-format` explicitly rather than inheriting whatever the installer defaults to.
- **Fix the per-platform subject.** Attach each platform's SBOM (and VEX, if per-platform) to that platform's manifest digest rather than to the shared index digest.
- **Decide the provenance mechanism.** Either unify image provenance onto `cosign attest` alongside the rest, or keep `actions/attest-build-provenance`. This is a real trade-off, not an obvious win: the GitHub action is what makes the provenance SLSA Build Level 3 (builder identity is the reusable `attest-images.yaml` workflow), and moving off it needs an equivalent story. Document the decision either way.
- **Sign the binary SBOMs** and decide whether binary metadata gets an OCI home (for example, pushing the binary attestation bundles and SBOMs to an OCI artifact so the same referrers query works for binaries) or stays release-asset-only by design.
- **Document the unified retrieval path** in `docs/integrator/supply-chain-verification.md` and `docs/user/artifact-verification.md`: one command per metadata kind, plus what a consumer does with the VEX document once they have it.

## Out of scope

- Changes to the in-toto predicate shapes themselves.
- The Rekor v2 signing migration (already shipped in #1650).
- Adding new metadata kinds beyond provenance, SBOM, and VEX.

## Success criteria

- `cosign download attestation` (or an equivalent referrers-API query) against a released image digest returns provenance, SBOM, and VEX, all signed and all verifiable.
- A downstream Grype or Trivy scan of a released image can consume our published VEX and suppress the same CVEs our own scan suppresses.
- cosign's version on every attestation path comes from `.settings.yaml`, and the bundle format is set explicitly rather than inherited.
- Per-platform SBOMs are attached to per-platform digests.
- Docs describe a single retrieval path per metadata kind.

Epic: #1149

lockwobr:
## Summary

Publishes every kind of release metadata as a cosign-signed attestation attached to its subject as an OCI referrer, and closes the three concrete defects that the split between mechanisms was hiding.

## Motivation / Context

Release metadata shipped via three independent mechanisms with different signers and different discovery paths, and one kind was never published at all: `.openvex.json` was only ever passed to Grype as scanner input, so downstream scanners re-flagged CVEs we had already triaged as not-affected. Three defects fell out of the same root cause:

1. `.github/actions/sbom-and-attest/action.yml` installed cosign with no version pin and never called `load-versions`, unlike `on-tag.yaml` which pins the `.settings.yaml` version. Since cosign v3 defaults `--new-bundle-format=true` (which is what attaches via the OCI referrers API), whether our SBOM attestations landed as referrers or as legacy `.att` tags was decided by an installer default rather than by us.
2. `attest-image-from-tag` resolves its subject with `crane digest` on the candidate tag, which returns the multi-arch index digest, so the linux/amd64 and linux/arm64 SPDX attestations were both attached to the same subject and could only be told apart by reading the predicate body.
3. goreleaser uploaded binary SBOMs as plain release assets with no attestation, while the binary provenance beside them was Rekor v2 signed.

Fixes: #1957
Related: #1149, #1650

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: release tooling (`.github/actions/`, `.github/scripts/`, `.goreleaser.yaml`)

## Implementation Notes

**Two design decisions were deliberately left as-is, with the reasoning recorded in the code and docs.**

*Provenance stays on `actions/attest-build-provenance`.* SLSA Build L3 requires build isolation, and GitHub grants it only when the attestation is generated from a reusable workflow, so the certificate SAN becomes `.../attest-images.yaml@refs/tags/...` rather than the caller. Moving to `cosign attest` would sign with the same job's OIDC token and lose the service-generated, caller-untamperable builder identity, dropping to L2. Since all three metadata kinds already land on the same image as referrers, the split costs a different verifier binary, not a different discovery path.

*Binary metadata stays on the GitHub release.* The binaries are release assets, so an OCI artifact would create a second discovery path for evidence about something that does not live in a registry, plus a new repository, naming contract, and GC lifecycle. The gap the issue actually identified was unsigned SBOMs, and that is closed by the sigstore bundle now sitting next to each asset, verified with the same `cosign verify-blob-attestation` flow already documented for the binary itself.

**ghcr.io does not implement the OCI 1.1 referrers API.** `GET /v2/nvidia/aicr/referrers/sha256:...` returns `404 MANIFEST_UNKNOWN`. It does serve the spec's referrers *tag fallback*, which is how this works end to end on ghcr. The current per-platform defect was confirmed the same way against a live released image: two cosign SBOM referrers, both carrying the index digest as subject.

**Pinning cosign is not a no-op.** The unpinned installer currently resolves to v3.0.6, which writes legacy `dsse` Rekor entries; the pinned v3.1.2 writes `hashedrekord`/PAE. `cosign verify-attestation` handles both, but the entry type on that path changes with this PR.

`crane digest --platform` fails closed if a platform is missing or the image is not an index. That is deliberate: the vuln-scan matrix already scans all seven images on both architectures, so a single-arch image is a regression rather than a valid state.

## Testing

```bash
make lint          # 0 golangci-lint issues, yamllint, license, docs-mdx, agents-sync, bom-pinning all green
go test ./tests/releasepolicy/...
actionlint -shellcheck=   # the exact merge-gate invocation, clean
shellcheck .github/scripts/sign-sbom.sh && bash -n .github/scripts/sign-sbom.sh
goreleaser check   # clean apart from the pre-existing brews deprecation
```

`tests/releasepolicy` fails 13 tests locally on macOS because `.github/scripts/publish-homebrew.sh` calls GNU `timeout`, which macOS does not ship. The identical 13 fail on unmodified `main` (`comm` of the two failure sets is empty), so this branch introduces no regression; CI runs on ubuntu where `timeout` exists.

Nothing in this change can execute outside a tagged release. Each shell change fails loudly and before publish if wrong: a missing platform aborts the attest job before promotion, an unknown cosign flag fails immediately, and the goreleaser `signs:` stanza fails at the sign phase rather than after upload. `sign-sbom.sh` fails closed when `SLSA_PREDICATE` is unset, which is deliberate: goreleaser registers the `${signature}` artifact even when the script writes nothing, so a silent skip would leave a dangling asset that fails at upload. CI paths are unaffected (`on-tag` always sets `SLSA_PREDICATE`; `build-attested.yaml` and `packaging.yaml` pass `--skip=sbom`).

## Risk Assessment

**Rollout notes:** Release-path only; no runtime or library surface changes. The two behaviour changes to be aware of are the Rekor entry type on the SBOM attestation path (dsse to hashedrekord/PAE, from pinning cosign) and the new goreleaser `signs:` stanza, which is the least-exercisable path here since it can only run under a real tagged release. Both fail loudly rather than silently. Revert is a straight revert of this commit.

## Checklist


.goreleaser.yaml
```diff
@@ -148,6 +148,30 @@ gomod:
...
+# packaging.yaml) pass --skip=sbom, which leaves this stanza with no matching
+# artifacts.
+signs:
+  - id: sbom
+    artifacts: sbom
+    cmd: .github/scripts/sign-sbom.sh
+    args:
+      - "${artifact}"
+      - "${signature}"
+    signature: "${artifact}.sigstore.json"
```
yuanchen8911, line 172: This stanza produces six new release assets (aicr x darwin/linux x amd64/arm64, aicrd x linux amd64/arm64, each `.sbom.json.sigstore.json`), and GoReleaser registers signature artifacts as release-uploadable ([sign pipeline](https://github.com/goreleaser/goreleaser/blob/v2.17.0/internal/pipe/sign/sign.go#L285-L294), [release upload filter](https://github.com/goreleaser/goreleaser/blob/v2.17.0/internal/pipe/release/release.go#L176-L191)); there is no `ids:` filter in the `release:` section. `expected_release_asset_names` in [.github/scripts/release-images.sh#L96-L112](https://github.com/NVIDIA/aicr/blob/35f9484c28a6f0f87ffcd6d408bc9e102bfe1ee1/.github/scripts/release-images.sh#L96-L112) is still the exact 13-name list, so `publish-release` fails its exact-mode asset validation on every tagged release -- after `promote-images` has already run, since the publish job depends on it. A rerun then fails earlier at the subset check in `release-target` ("unexpected asset") when reusing the draft.

The fixture in `tests/releasepolicy/release_scripts_test.go` does not derive from the shell allowlist; it independently duplicates the same 13 names, and neither list is reconciled with the goreleaser signing configuration -- which is why `go test ./tests/releasepolicy/...` stays green while every release would break.

Please add the six bundle names to `expected_release_asset_names` and to the test fixture. A releasepolicy check that cross-references the `signs:` output pattern against the allowlist would keep these from drifting apart again.

docs/integrator/supply-chain-verification.md
```diff
@@ -55,6 +55,109 @@ export IMAGE_API_DIGEST="${IMAGE_API}@${DIGEST_API}"
...
+attached to that platform's own manifest digest; querying the index digest for
+an SBOM returns nothing.
+
+Resolve the digests first. `crane digest` without `--platform` returns the index
+digest; with `--platform` it returns that platform's child manifest digest.
+
+```shell
+export IMAGE="ghcr.io/nvidia/aicr"
+export DIGEST=$(crane digest "${IMAGE}:${TAG}")
+export DIGEST_AMD64=$(crane digest --platform linux/amd64 "${IMAGE}:${TAG}")
```
yuanchen8911: These platform lookups re-resolve the mutable tag after `${DIGEST}` was pinned. If the tag is repointed between the two resolutions, the platform digest comes from a different index than the provenance/VEX subject, and the wildcard signer regexp (`refs/tags/.+`) still verifies that other release's validly signed SBOM -- pairing metadata with content it does not describe, which is the mutable-tag threat this guide is structured around.

Resolve from the pinned index instead, exactly as `attest-image-from-tag` itself does:

```shell
export DIGEST_AMD64=$(crane digest --platform linux/amd64 "${IMAGE}@${DIGEST}")
export DIGEST_ARM64=$(crane digest --platform linux/arm64 "${IMAGE}@${DIGEST}")
```

The same change applies to the `DIGEST_API_AMD64` export in the container-image-SBOM section, the `DIGEST_AMD64` export under Method 2, and the `DIGEST_AMD64` export in docs/user/artifact-verification.md.

docs/integrator/supply-chain-verification.md
```diff
@@ -55,6 +55,109 @@ export IMAGE_API_DIGEST="${IMAGE_API}@${DIGEST_API}"
...
+
+export AICR_ISSUER="https://token.actions.githubusercontent.com"
+export AICR_SIGNER='^https://github\.com/NVIDIA/aicr/\.github/workflows/attest-images\.yaml@refs/tags/.+$'
+```
+
+Then one command per kind:
+
+```shell
+# 1. Build provenance (index digest)
+gh attestation verify "oci://${IMAGE}@${DIGEST}" \
```
yuanchen8911, line 112: `gh attestation verify` fetches bundles from GitHub's attestations API by default; it only reads the registry referrer with `--bundle-from-oci` ([gh manual](https://cli.github.com/manual/gh_attestation_verify)). The provenance is pushed as a referrer (`push-to-registry: true` in sbom-and-attest), but as written this "unified registry retrieval" flow never exercises that referrer for provenance, and a silently failed referrer push would go undetected while the command still succeeds via the API. Verified against the v0.18.0 release image: with an intentionally invalid GitHub API token, the `--bundle-from-oci` form succeeds while this default form fails with HTTP 401, confirming the two discovery paths are distinct.

Please add `--bundle-from-oci` here and in the docs/user/artifact-verification.md quick example. The later sections of this guide that still verify via the attestations API (the Method 1/Method 2 provenance commands) should either be updated to match or explicitly labeled as the API-based alternative -- otherwise the guide claims a single OCI discovery path while demonstrating a second one.

docs/user/artifact-verification.md
```diff
@@ -24,6 +24,56 @@ verified, but it cannot reach the higher trust levels described below.
...
+```shell
+export IMAGE="ghcr.io/nvidia/aicr"
+export TAG=$(curl -s https://api.github.com/repos/NVIDIA/aicr/releases/latest | jq -r '.tag_name')
+export DIGEST=$(crane digest "${IMAGE}:${TAG}")
+export DIGEST_AMD64=$(crane digest --platform linux/amd64 "${IMAGE}:${TAG}")
+export AICR_ISSUER="https://token.actions.githubusercontent.com"
+export AICR_SIGNER='^https://github\.com/NVIDIA/aicr/\.github/workflows/attest-images\.yaml@refs/tags/.+$'
+
+# Build provenance (attached to the multi-platform index digest)
+gh attestation verify "oci://${IMAGE}@${DIGEST}" --repo NVIDIA/aicr \
```
yuanchen8911, line 54: This quick path proves "some NVIDIA/aicr release workflow signed this," not "release `${TAG}` signed this": the gh command omits `--source-ref`, and both cosign regexps end in `refs/tags/.+$`. Since every attestation is digest-bound, the practical exposure is tag-provenance ambiguity rather than wrong-content metadata (an RC re-tagged to GA carries attestations for the same bytes), so this is hardening rather than a defect -- but the fix is small: add `--source-ref "refs/tags/${TAG}"` to the gh command, and use an exact `--certificate-identity` with the `${TAG}`-qualified identity instead of a dynamically constructed regex. The integrator guide's unified example retains the same wildcard `AICR_SIGNER` and should get the same treatment.

.github/scripts/sign-sbom.sh
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
```
yuanchen8911, line 1: `TestReleaseScriptsStructure` in tests/releasepolicy enumerates release scripts explicitly (`release-images.sh`, `publish-homebrew.sh`) and enforces mode 0755, the bash shebang, the copyright header, and strict mode. This script already satisfies all four -- please add it to the list so they stay enforced. Its fail-closed env guards and retry loop are also directly testable with the existing `runScript` fixture pattern.

yuanchen8911 (changes requested):
I verified the wiring end to end: the single caller chain (attest-images.yaml -> attest-image-from-tag -> sbom-and-attest), the cosign pin flowing from .settings.yaml via load-versions, SLSA_PREDICATE/AICR_SIGNING_CONFIG being set by generate-slsa-predicate before goreleaser runs, and the --skip=sbom claims for the snapshot workflows. The subject-cardinality model (SBOM per platform manifest, VEX and provenance on the index) is right, and the fail-closed choices are consistently made.

One blocking issue (the release asset allowlist) and a few doc corrections, inline below.

Separately, the new invariants deserve locking in tests/releasepolicy, which exists precisely for this: nothing asserts the SBOM subject is the per-platform digest (the exact defect this PR fixes could regress to the index digest with all tests green), nothing asserts `--new-bundle-format=true` is present on the attest calls, and `TestReleaseCosignAttestationsAreBounded` enumerates only the two SBOM steps, so the new OpenVEX step's timeout is unenforced. These are cheap given the suite's existing helpers.

.goreleaser.yaml
lockwobr: This was the most valuable catch on either of my supply-chain PRs, and you diagnosed the root cause precisely: the fixture duplicates the allowlist rather than deriving from it, so a green suite told me nothing. Fixed in 537d72da.

I established the six filenames two independent ways rather than guessing. Statically: goreleaser's sbom pipe defaults `artifacts: binary` documents to `{{ .Binary }}_{{ .Version }}_{{ .Os }}_{{ .Arch }}.sbom.json`, the sign pipe re-expands `signature:` with `env["artifact"]` set to the SBOM *basename* and registers the result as `artifact.Signature`, and `ReleaseUploadableTypes()` includes it with no `release.ids` filter to stop it. Empirically: a minimal reproduction of the same builds/sboms/signs shape run under `goreleaser release --snapshot` emitted exactly six `Signature` artifacts. Both agree on `<sbom-name>.sigstore.json`.

I also confirmed your failure prediction rather than trusting it: with a `timeout` shim on PATH to clear this machine's pre-existing failures, reverting only the shell allowlist while leaving the release carrying the bundles made `TestReleasePublishRequiresExactDraftAssetsAndID` fail all three subtests through the real script.

Your suggested anti-drift check is in as `TestReleaseSbomSignaturesAreAllowlisted`, and it hardcodes nothing: it extracts `expected_release_asset_names()` from the shipped script and evaluates it under bash, derives the signature suffix from the `signs:` stanza (rejecting a templated suffix, which would make asset names unpredictable), asserts a bijection between `.sbom.json` documents and their signatures, and asserts the Go fixture equals the shell allowlist exactly. Mutation-verified in both directions.

docs/integrator/supply-chain-verification.md
lockwobr: Correct, and it reintroduced the exact threat this guide is built around. Fixed in 537d72da at all four sites, resolving from `${IMAGE}@${DIGEST}` and `${IMAGE_API}@${DIGEST_API}` as `attest-image-from-tag` itself does. Verified against the live registry that the pinned-index form returns the same amd64 digest as the tag form for the current release, and that a missing platform exits 1 with `no child with platform`, which is what the action's comment claims.

docs/integrator/supply-chain-verification.md
lockwobr: Confirmed, and this one also undercut a claim in my PR description, so thank you for testing it rather than reading it. Fixed in 537d72da. `--bundle-from-oci` is present in gh 2.95.0 and I verified the full form end to end against the real v0.18.0 image (exit 0). Added to both the unified example and the quick example. The remaining Method 1 and Method 2 commands are now explicitly labelled as the API-based alternative, with a paragraph explaining that the default form would not detect a silently failed referrer push, which is precisely your point.

docs/user/artifact-verification.md
lockwobr: Agreed, and I kept your framing that this is hardening rather than a defect, since every attestation is digest-bound. Fixed in 537d72da: `--source-ref "refs/tags/${TAG}"` added to the quick path (the integrator guide already carried it) and the wildcard `AICR_SIGNER` replaced with an exact tag-qualified `--certificate-identity` in both guides. Verified the exact-identity form succeeds against the real release image rather than assuming the identity string.

.github/scripts/sign-sbom.sh
lockwobr: Added in 537d72da. It satisfied all four structural checks unchanged, so registering it only locks that in. I also took the second half of your note: `TestReleaseSignSbomBehavior` exercises the fail-closed env guards and the retry loop through the existing `runScript` fixture with a mocked cosign, covering missing and partial arguments, a missing or empty SBOM, `SLSA_PREDICATE` unset, `AICR_SIGNING_CONFIG` unset, success, retry-then-succeed, and failure after the third error, asserting attempt counts and the full cosign argv.

.github/actions/sbom-and-attest/action.yml
```diff
@@ -32,20 +60,38 @@ runs:
...
+            exit 1
+          fi
+        done
+        # A multi-platform release image must yield two distinct child
+        # manifests. Equal digests mean the caller resolved the index instead
+        # of a platform, or the image lost a platform. Fail closed either way.
+        if [[ "${INPUT_AMD64_DIGEST}" == "${INPUT_ARM64_DIGEST}" ]]; then
+          echo "::error::amd64_digest and arm64_digest are identical; the image is not a multi-platform index"
           exit 1
         fi
```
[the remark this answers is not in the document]
lockwobr: Fixed in 537d72da. A fail-closed loop now rejects either platform digest when it equals `INPUT_IMAGE_DIGEST`, before the existing distinctness check, so setting one platform to the index while varying the other no longer passes. Mutation-verified: removing the guard makes the two new subtests pass when they should fail.

.github/actions/attest-image-from-tag/action.yml
```diff
@@ -103,8 +103,41 @@ runs:
...
+          echo "${arch}_digest=${platform_digest}" >> "$GITHUB_OUTPUT"
+        done
+
     - name: Generate SBOM and attestations
       uses: ./.github/actions/sbom-and-attest
       with:
         image_name: ${{ steps.validate.outputs.image_name }}
         image_digest: ${{ steps.validate.outputs.expected_digest }}
+        amd64_digest: ${{ steps.platforms.outputs.amd64_digest }}
+        arm64_digest: ${{ steps.platforms.outputs.arm64_digest }}
```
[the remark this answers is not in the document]
lockwobr: Added in 537d72da. The logic lives in composite-action YAML, but it is reachable: `release_workflow_test.go` already extracts a step's `run:` and executes it under bash. `TestReleasePlatformDigestResolution` asserts the step id, the bounded `crane` call, and the `amd64_digest`/`arm64_digest` handoff into the `sbom-and-attest` `with:` block, then runs the script against a fake `crane` for both children resolving, a missing platform, amd64 equal to the index, arm64 equal to the index, and a malformed digest. `TestReleaseSbomAttestInputValidation` covers the receiving side.

.github/scripts/sign-sbom.sh
```diff
@@ -0,0 +1,95 @@
...
+    exit 0
+  fi
+  if [[ "${n}" -lt "${ATTEMPTS}" ]]; then
+    echo "cosign attest-blob attempt ${n} failed for ${SBOM_PATH}; retrying" >&2
+    sleep $((n * 5))
+  fi
+done
+
+echo "cosign attest-blob failed after ${ATTEMPTS} attempts for ${SBOM_PATH}" >&2
+exit 1
```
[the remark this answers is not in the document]
lockwobr: Added in 537d72da as `TestReleaseSignSbomBehavior`, using the existing `runScript` fixture with a mocked cosign: missing and partial arguments, missing SBOM, empty SBOM, `SLSA_PREDICATE` unset, `AICR_SIGNING_CONFIG` unset, success, retry-then-succeed at two attempts, and failure after the third error. Attempt counts and the full cosign argv are asserted.

docs/integrator/supply-chain-verification.md
```diff
@@ -55,6 +55,109 @@ export IMAGE_API_DIGEST="${IMAGE_API}@${DIGEST_API}"
...
+an SBOM returns nothing.
+
+Resolve the digests first. `crane digest` without `--platform` returns the index
+digest; with `--platform` it returns that platform's child manifest digest.
+
+```shell
+export IMAGE="ghcr.io/nvidia/aicr"
+export DIGEST=$(crane digest "${IMAGE}:${TAG}")
+export DIGEST_AMD64=$(crane digest --platform linux/amd64 "${IMAGE}:${TAG}")
+export DIGEST_ARM64=$(crane digest --platform linux/arm64 "${IMAGE}:${TAG}")
```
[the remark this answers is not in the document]
lockwobr: Fixed in 537d72da at all four locations. yuanchen8911 raised the same point with the sharper consequence, which is worth recording here: re-resolving the mutable tag after `${DIGEST}` is pinned means a repointed tag yields a platform digest from a different index than the provenance and OpenVEX subject, and the wildcard signer regexp would still verify that other release's validly signed SBOM.

docs/user/artifact-verification.md
```diff
@@ -24,6 +24,56 @@ verified, but it cannot reach the higher trust levels described below.
...
+  --signer-workflow NVIDIA/aicr/.github/workflows/attest-images.yaml
```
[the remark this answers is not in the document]
lockwobr: Fixed in 537d72da: `--source-ref "refs/tags/${TAG}"` added, and the wildcard identity regexp replaced with an exact tag-qualified `--certificate-identity`, verified against the real release image.

docs/user/artifact-verification.md
```diff
@@ -24,6 +24,56 @@ verified, but it cannot reach the higher trust levels described below.
...
+
+```shell
+grype "${IMAGE}@${DIGEST}" --vex aicr-openvex.json --only-fixed --fail-on high
+trivy image --vex aicr-openvex.json "${IMAGE}@${DIGEST}"
+```
+
+Each statement carries a `justification` and an `impact_statement` explaining
+why the vulnerable code is not reachable. Read those before adopting a
+suppression: the analysis is specific to how AICR invokes the affected
+component. Verification of these attestations needs Cosign v3.0.1 or newer. For
```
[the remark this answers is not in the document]
lockwobr: Fixed in 537d72da. Both guides now scope the Cosign v3.0.1+ floor to the `cosign verify-attestation` commands and state that `gh attestation verify` ships with the GitHub CLI and does not use Cosign.

docs/integrator/supply-chain-verification.md
```diff
@@ -55,6 +55,130 @@ export IMAGE_API_DIGEST="${IMAGE_API}@${DIGEST_API}"
...
+  --certificate-identity "${AICR_SIGNER}" \
+  "${IMAGE}@${DIGEST_AMD64}" \
+  | jq -r '.payload' | base64 -d | jq '.predicate' > sbom-linux-amd64.spdx.json
+
+# 3. OpenVEX vulnerability triage (index digest)
+cosign verify-attestation --type openvex \
+  --certificate-oidc-issuer "${AICR_ISSUER}" \
+  --certificate-identity "${AICR_SIGNER}" \
+  "${IMAGE}@${DIGEST}" \
+  | jq -r '.payload' | base64 -d | jq '.predicate' > aicr-openvex.json
```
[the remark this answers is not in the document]
lockwobr: Valid, and testing it turned up that your suggested fix is necessary but not sufficient. Fixed in 2e8c0173.

Reproduced first: `printf '' | jq -r '.payload' | base64 -d | jq '.predicate' > out.json` exits 0 and leaves a 0-byte file. Adding `set -o pipefail` corrects the exit status but still leaves the 0-byte file, because the shell truncates the redirect target before `cosign` ever runs. So all four sites now do both: `set -o pipefail`, plus capturing the predicate and writing it only once the whole pipeline has succeeded. The one pipeline that merely prints gets pipefail alone, which is all it needs.

Verified by extracting each block and running it against a fake failing `cosign`: all four now exit 1 and create zero files, where the pre-fix shape exits 0 and creates an empty `aicr-openvex.json`. A fake succeeding `cosign` writes the correct predicate.

tests/releasepolicy/release_scripts_test.go
```diff
@@ -57,6 +58,251 @@ func TestReleaseScriptsStructure(t *testing.T) {
...
+	output, err := command.CombinedOutput()
+	if err != nil {
+		t.Fatalf("evaluate expected_release_asset_names: %v\n%s", err, output)
+	}
+	var names []string
+	if err := json.Unmarshal(output, &names); err != nil {
+		t.Fatalf("parse allowlist %q: %v", output, err)
+	}
+	return names
+}
```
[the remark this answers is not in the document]
lockwobr: Fixed in 2e8c0173, and it was my own new code, so a fair catch. `shellReleaseAssetNames` now uses `exec.CommandContext` with a 10s bounded context matching `runScript`'s existing shape in the same package, and reports a deadline overrun distinctly from a script failure.

.github/actions/sbom-and-attest/action.yml
```diff
@@ -99,35 +159,91 @@ runs:
...
+          exit 1
+        fi
+        # Fail closed on a malformed document rather than publishing a VEX
+        # attestation that no scanner can apply.
+        if ! jq -e '(.["@context"] | type == "string") and (.statements | type == "array")' \
+          "${vex_file}" > /dev/null; then
+          echo "::error::OpenVEX document is missing @context or statements"
+          exit 1
+        fi
+        echo "file=${vex_file}" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
lockwobr: Accepting the substance and declining the mechanism, with reasoning. Applied in 2e8c0173.

You are right that the guard was too weak: it checked only that `@context` was a string and `statements` an array, so both `statements: []` and `[{}]` passed and an unusable attestation could ship. That is now rejected, along with missing document metadata, an out-of-enum `status` or `justification`, a `not_affected` statement carrying neither `justification` nor `impact_statement`, and an `affected` statement with no `action_statement`.

What I did not do is pin a schema or add a validator. This step runs in the release job *after* image promotion, so a schema-fetching dependency would introduce a new network failure mode at the most expensive point in the release. Everything above is expressible in `jq`, which is already a dependency. As a local substitute for version pinning, the guard asserts `@context` equals `https://openvex.dev/ns/v0.2.0` exactly, which is what makes the enum tables beneath it legitimate and forces a future spec bump to move the guard and its tests together.

Two notes from reading the actual v0.2.0 spec while implementing. "Has a `vulnerability`" is too weak a check, since the required field is a struct whose only required member is `name`, so `"vulnerability": {}` would pass; the guard requires a non-empty `vulnerability.name`. And `products` is **not** a v0.2.0 requirement, because it can cascade from an encapsulating format. This document has no such encapsulation and scanners match on `products[].purl`, so a statement without one cannot suppress anything; the guard rejects it on that ground rather than as a spec violation, and the comment says so.

Tests are in `TestReleaseOpenVEXValidation`, 18 subtests covering the four cases you named plus a downgraded `@context`, hostile shapes (a scalar `vulnerability`, a string `products`, a scalar inside `statements`, a top-level array), `not_affected` with only an `impact_statement` (which must be *accepted*), and a case that stages the committed `.openvex.json` and asserts it survives the guard it ships through. Verified the tests actually bite: reverting the jq block to the old two-condition check fails 14 of the 18.

yuanchen8911 (approved):
All five findings verified fixed at 2e8c0173a. TestReleaseSbomSignaturesAreAllowlisted closes both halves of the asset-allowlist gap — the stanza-to-allowlist tie and the fixture now deriving from the shipped allowlist instead of duplicating it. Digest-pinned platform lookups, --bundle-from-oci with the API-path distinction, --source-ref on the quick path, and the sign-sbom.sh structure/behavior tests all check out. /lgtm

PR #1984, opened 2026-08-02
fix(recipe): align AKS inference K8s floors with DRA-GA >= 1.34 rationale

### LINKED ISSUE #1969: Align AKS inference K8s floors with DRA-GA >= 1.34 rationale
## Summary

The AKS **inference** recipe family has the same Kubernetes-floor drift that #1772 identified (and PR #1908 fixes) for the training family. `recipes/overlays/aks.yaml` pins `K8s.server.version >= 1.34` with the rationale that **all** AKS recipes carry DRA (`nvidia-dra-driver-gpu`) via `base.yaml`, and DRA graduated to GA in Kubernetes 1.34. The non-dynamo inference leaves still declare lower floors.

Because snapshot-time recipe selection evaluates only a candidate leaf's **own** `spec.constraints` (ancestors are pruned by `filterToMaximalLeaves` before evaluation), a leaf that omits the `>= 1.34` restatement resolves on a 1.32.x cluster and ships a DRA component whose `resource.k8s.io` APIs are not GA there.

## Current leaf floors

| Overlay | Floor |
|---|---|
| `aks.yaml` (service parent) | `>= 1.34` |
| `aks-inference.yaml` | `>= 1.30` |
| `h100-aks-inference.yaml` | `>= 1.32.4` |
| `h100-aks-ubuntu-inference.yaml` | `>= 1.32.4` |
| `h100-aks-ubuntu-inference-dynamo.yaml` | `>= 1.34` ✅ (already correct) |

Verified live: `aicr query --service aks --accelerator h100 --intent inference --os ubuntu` resolves a `>= 1.32.4` floor today. None of the three overlays disables the DRA component, so the family rationale applies to them unchanged.

## Proposed work

Mirror PR #1908's pattern:

- Restate `>= 1.34` on the three leaves above, with the family-standard `# DRA requires Kubernetes 1.34+ (GA)` comment.
- Extend the floor regression test in `pkg/recipe/metadata_store_test.go` with inference cases, keeping #1908's two-part assertion: (1) the resolved recipe carries the floor, and (2) the leaf overlay itself restates it.
- Sweep fixtures for stale floor assertions (the Azure cuj2 UAT fixture is dynamo-based and already asserts `>= 1.34`; #1908 already bumps the Go snapshot fixtures to `v1.34.0`, so fallout should be minimal).

Sequencing: land after #1908 merges — it extends the same test function and would otherwise conflict.

## References

- #1772 (training-family alignment; deliberately training-scoped)
- #1908 (training fix; the implementation pattern to mirror)
- #1744 (Slurm leaf bump + floor regression test)
- `recipes/overlays/aks.yaml` (DRA-GA `>= 1.34` rationale)

mohityadav8:
Bumps aks-inference, h100-aks-inference, and h100-aks-ubuntu-inference from >= 1.30 / >= 1.32.4 to >= 1.34, matching the DRA-GA rationale already stated in aks.yaml and mirroring the training-family fix (#1908). h100-aks-ubuntu-inference-dynamo was already correct.

Extends TestH100AKSUbuntuTrainingSlurmFloorNotClobbered with the inference-family leaf cases.

Fixes: #1969


pkg/recipe/metadata_store_test.go
```diff
@@ -2948,6 +2948,53 @@ func TestH100AKSUbuntuTrainingSlurmFloorNotClobbered(t *testing.T) {
 			},
 			wantK8sFloor: ">= 1.34",
 		},
+		// AKS inference family (see #1969, the inference-family sibling of
```
njhensley, line 2951: **🔵 — Test fn name + godoc scoped to "AKS training" but now also covers the AKS inference family**

These 4 new inference cases live in TestH100AKSUbuntuTrainingSlurmFloorNotClobbered, whose name and function-level godoc (~L2833-2850) describe only an AKS *training* floor guard — the inference scope is documented only in this inline comment. The name/godoc never mention inference, so a search for the #1969 inference-floor regression guard won't surface it under a "Training Slurm" name.

**Blast radius:** Discoverability only — no runtime impact; the test passes. (Retiered 🟡→🔵: the name was already a partial misnomer pre-PR, and inference scope is at least visible via the inline comment + case names.)

**Fix:** Optional: rename to a family-neutral TestAKSK8sFloorNotClobbered and fold the #1969 note into the godoc (it locks the >= 1.34 DRA-GA floor for both the training #1772/#1908 and inference #1969 families). Or split into a sibling matching the TestGB200OKEFloorNotClobbered shape.

recipes/overlays/aks-inference.yaml
```diff
@@ -31,8 +31,9 @@ spec:
 
   # Inference specific constraints for AKS workloads
   # Constraint names use fully qualified measurement paths: {type}.{subtype}.{key}
+  # DRA requires Kubernetes 1.34+ (GA)
```
njhensley, line 34: **🔵 — Inference leaves append the DRA comment instead of replacing the "specific constraints" line, unlike #1908**

This PR appends `# DRA requires Kubernetes 1.34+ (GA)` while keeping the existing `# ... specific constraints ...` line on all 3 leaves. The mirrored training fix #1908 (git show aa8733f7) *replaced* that line instead — so the two families now differ cosmetically.

**Blast radius:** Cosmetic only — comments, no behavior/resolution change. Minor asymmetry for anyone diffing training vs inference overlays.

**Fix:** Optional, for byte-for-byte symmetry with #1908: drop the pre-existing `# ... specific constraints ...` line on the 3 leaves so only the DRA rationale remains. Keeping both lines is arguably more informative — safe to leave as-is.

njhensley (approved):
## 📋 Multi-persona review — approve

**Method:** 3 independent persona reviewers (Domain/Recipe-correctness · Test-coverage · Consistency/Ops) → adversarial senior meta-reviewer re-derived every finding from the resolved code. Pinned to head `58a8449c`.

A clean, tightly-scoped fix that does exactly what it claims — raising the `K8s.server.version` floor on the three lagging AKS inference overlays to `>= 1.34`, matching the DRA-GA rationale in `aks.yaml` and faithfully mirroring the merged training-family fix #1908/#1772. Verified rather than assumed:

- **Correct floor** — DRA graduated to GA in K8s 1.34 (`resource.k8s.io` stable v1, default-on); consistent with `aks.yaml`.
- **Complete family** — all 4 `service: aks` + `intent: inference` overlays now *restate* `>= 1.34` in their own `Spec.Constraints` (the dynamo leaf was already correct); no A100 inference leaf exists. This matters because ancestors are pruned at snapshot-eval time — a leaf that only *inherited* the floor would silently escape the gate.
- **Strong test** — the 4 new cases run through both assertions (resolved floor **and** leaf self-restatement); mutation-tested (revert a floor → fail; drop a restatement → fail check 2). `./pkg/recipe/` passes, golangci-lint clean.
- **No docs drift, no BOM regen** (no chart pin touched); **PR body accurate**.

**Confirmed non-issues (examined, cleared):** floor-value correctness · family completeness (4/4, no A100 gap) · test discriminating power · docs drift · BOM regen · #1908 mirror fidelity · PR-body accuracy · operational unsatisfiability (AKS supports 1.34; same accepted risk as the merged training fix).

The two inline items below are cosmetic nitpicks — neither blocks merge.

```
🔴 Blocker 0 | 🟠 Major 0 | 🟡 Minor 0 | 🔵 Nitpick 2      Recommendation: Approve with comments
```

Issue #1649, opened 2026-07-08
validator(dra-support): behavioral MNNVL ComputeDomain IMEX allocation subtest


yuanchen8911:
## Summary

`dra-support` (conformance phase) validates ComputeDomain DRA on ComputeDomain-only clusters via driver health probes and full ResourceSlice validation, but its *behavioral* allocation subtest runs only when full-GPU DRA (`gpu.nvidia.com`) is usable. In the shipped ComputeDomain-only configuration the check records the behavioral subtest as not-applicable and relies on slice validation alone.

Add a dedicated behavioral subtest that exercises the MNNVL ComputeDomain → generated ResourceClaimTemplate → IMEX-channel allocation flow inside the check itself, so ComputeDomain-only clusters get a behavioral DRA pass/fail signal from `dra-support` directly.

## Current state

**ComputeDomain-only is the shipped default, not an edge case.** `recipes/components/nvidia-dra-driver-gpu/values.yaml` sets `resources.gpus.enabled: false` and `computeDomains.enabled: true`. Every GB200/GB300 cluster AICR deploys therefore takes the not-applicable branch at `validators/conformance/dra_support_check.go:232-237`, which records `skipped (not applicable): full-GPU DRA not enabled…` and returns a pass.

The marker is already in place at `dra_support_check.go:197`.

**The not-applicable pass is not vacuous, but it is structural.** The code is explicit that N/A is allowed only because `validateNVIDIAResourceSlices` already passed, and that validation is substantive — it requires complete, current-generation pools advertising real devices. But structural validation is not allocation: a driver can publish well-formed `compute-domain.nvidia.com` ResourceSlices and still fail to hand a pod an IMEX channel.

## Why this matters most for conformance-only runs

Behavioral coverage of this flow does exist today, through the NCCL all-reduce validator. `nccl_all_reduce_bw_constraint.go:943-949` creates the ComputeDomain CR, the DRA driver reconciles it into the `nccl-all-reduce-imex` ResourceClaimTemplate, and the worker pods reference it so the kubelet mounts `/dev/nvidia-caps-imex-channels` (documented at `validators/performance/testdata/gb200/eks/runtime-nvls.yaml:24-30`). A full `aicr validate` on GB200 that includes the performance phase does exercise the flow end to end.

[...]

PR #1985, opened 2026-08-02
GKE generated AICR GB300 recipies

rawsocket:
## Summary

Recipes for GB300.

## Motivation / Context

Updating the AICR repo with validate versions.

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [X] Refactoring (no functional changes)

## Component(s) Affected

- [X] Recipe engine / data (`pkg/recipe`)
- [X] Other: kwok/

## Implementation Notes

GB300 recipes generated with up to date component versions.

## Testing

```bash
make qualify
```

## Risk Assessment

- [X] **Low** — Isolated change, well-tested, easy to revert

## Checklist

- [X] Tests pass locally (`make test` with `-race`)
- [X] Linter passes (`make lint`)
- [X] I did not skip/disable tests to make CI green
- [X] Commits are cryptographically signed (`git commit -S`) — [GPG signing info](https://docs.github.com/en/authentication/managing-commit-signature-verification)


recipes/overlays/gb300-gke-cos-inference-dynamo.yaml
```diff
@@ -0,0 +1,94 @@
...
+  validation:
+    deployment:
+      checks:
+        - operator-health
+        - expected-resources
+        - gpu-operator-version
+        - check-nvidia-smi
+      constraints:
+        - name: Deployment.gpu-operator.version
+          value: ">= v24.6.0"
```
mchmarny, line 79: **Blocking — this lowers an inherited validation floor instead of adding to it.**

This overlay's `base` is `gb300-gke-cos-inference`, which sets the same constraint to `>= v25.10.0` (`gb300-gke-cos-inference.yaml:79`). Overlay constraints union by name with the overlay winning on collision (`mergeValidationPhase`, `pkg/recipe/validation.go:141-148`), so this line *replaces* the base floor.

Confirmed by resolving the recipe at this head:

```
$ aicr recipe --service gke --accelerator gb300 --os cos --intent inference --platform dynamo
  deployment:
    constraints:
      - name: Deployment.gpu-operator.version
        value: '>= v24.6.0'
```

Consequence: a GB300 cluster running gpu-operator v24.6–v25.9 **passes** the `dynamo` deployment gate that the non-dynamo GB300 recipe fails it on — a false PASS on the more specialized recipe. The sibling `b200-gke-cos-inference-dynamo.yaml:65` deliberately holds `>= v25.10.0` and comments why.

Fix: set `">= v25.10.0"`, or drop the constraint block to inherit the base floor. If v24.6 is genuinely correct for GB300 + Dynamo, please add a comment saying why it is lower than its own base.

recipes/overlays/gb300-gke-cos-inference.yaml
```diff
@@ -0,0 +1,96 @@
...
+    # Consequently, standalone networking installer DaemonSets (gke-nccl-roce or
+    # gke-nccl-tcpxo) are intentionally bypassed to prevent overriding GKE managed networking.
+    - name: nodewright-customizations
+      type: Helm
+      manifestFiles:
+        - components/nodewright-customizations/manifests/tuning-gke.yaml
+      overrides:
+        service: gke
+        accelerator: gb300
+        intent: multiNodeTraining
```
mchmarny, line 61: **Blocking — training tuning profile applied to an inference recipe.**

`criteria.intent` here is `inference` (L16), but the `nodewright-customizations` override sets `intent: multiNodeTraining`. This block is byte-identical to the one in `gb300-gke-cos-training.yaml:61`, which is where it belongs.

The value is not inert — it lands in the rendered NodeWright tuning ConfigMap (`recipes/components/nodewright-customizations/manifests/tuning-gke.yaml:85`, `configMap.intent: {{ $cust.intent }}`), so inference nodes receive the multi-node-training tuning profile.

Every other inference recipe passes `inference` here (`b200-gke-cos-inference.yaml:61`, `gb200-eks-inference.yaml:73`); the `*-training` overlays are the ones using `multiNodeTraining`.

Fix: `intent: inference`.

(CodeRabbit flagged this as well — I verified it independently against the head and it is real and still open.)

kwok/profiles/gke/a4x-maxgpu-4g-metal.yaml
```diff
@@ -0,0 +1,31 @@
...
+  name: a4x-maxgpu-4g-metal
+  labels:
+    provider: gke
+    nodeType: accelerated
+    accelerator: gb300
+    machineFamily: a4x-maxgpu-4g-metal
+
+spec:
+  instanceType: a4x-maxgpu-4g-metal
+  arch: arm64
```
mchmarny, line 17: **Blocking — this profile is never loaded; the GB300 KWOK lanes ran on amd64 H100 nodes.**

`get_profiles()` in `kwok/scripts/apply-nodes.sh:52` maps service+accelerator to profile paths, and the `gke` branch is still hardcoded to EKS fallbacks:

```bash
gke)
    # GKE uses same profiles for now
    case "$accelerator" in
        gb200) gpu_profile="eks/p6-gb200.yaml" ;;
        *)     gpu_profile="eks/p5-h100.yaml" ;;
    esac
```

Nothing in the tree references `kwok/profiles/gke/`, and `kwok/README.md:131` calls out updating `get_profiles()` as a required step when adding a profile.

The effect is visible in this PR's own CI. From `Tier 2 / gb300-gke-cos-training (helm)`:

```
[INFO] Recipe: gb300-gke-cos-training (service=gke, accelerator=gb300)
[INFO] Profiles: system=eks/system-m7i.yaml, gpu=eks/p5-h100.yaml
[INFO] Creating 4 GPU nodes (p5.48xlarge, 8x NVIDIA-H100-80GB-HBM3)
```

So all four green `Tier 2 / gb300-*` checks validated arm64 GB300 recipes against amd64 H100 nodes — the `arch: arm64` on this line, and the arm64 image pins in the overlays, were never exercised.

Fix: add a `gb300) gpu_profile="gke/a4x-maxgpu-4g-metal.yaml"` case and a `gke` system-profile entry pointing at `gke/system-n2.yaml`. If you would rather land the recipes first, deleting these two profile files is an equally clean resolution — the concern is shipping them in a state where they appear wired up but are not.

Separately and non-blocking, same file: `memory: 245760` (L28) — both sibling GPU profiles annotate the unit (`memory: 81920  # MiB per GPU`, `memory: 196608  # MiB per GPU (192GB)`). 245760 MiB is 240 GiB, which does not line up with GB300's published ~288 GB HBM3e per GPU. Worth re-checking against the Version Vector source and adding the unit comment.

recipes/components/gke-nccl-roce/values.yaml
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# GPUDirect RDMA over RoCE configuration for NVIDIA GPUs on GKE.
+# Derived dynamically from authoritative GKE Version Vector compatibility matrices and GCE SoT.
+
+image:
```
mchmarny, line 18: **Non-blocking — this values file is unreachable.**

`gke-nccl-roce` is not in `recipes/registry.yaml` (only `gke-nccl-tcpxo`, L176) and no overlay declares it as a `componentRef`. The only mentions anywhere are the comments in the two GB300 overlays (L52) explaining that the RoCE installer is *intentionally bypassed* under GKE Managed RDMA.

Because it is unregistered it also will not appear in `docs/user/container-images.md`, so these pins (`nccl-roce-installer:v2.28.9-arm64`, `gib-plugin:v1.1.2-arm64`) sit outside the BOM freshness gate and the weekly refresh will not catch drift on them.

Suggest dropping it from this PR and adding it together with its registry entry when the component is actually wired up.

recipes/overlays/gb300-gke-cos-inference.yaml
```diff
@@ -0,0 +1,96 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
mchmarny, line 1: **Non-blocking — missing Apache license header.**

This file and three others in the PR carry only the copyright line, without the Apache-2.0 block:

- `recipes/overlays/gb300-gke-cos-inference.yaml`
- `recipes/overlays/gb300-gke-cos-training.yaml`
- `kwok/profiles/gke/a4x-maxgpu-4g-metal.yaml`
- `kwok/profiles/gke/system-n2.yaml`

These are the only four files of their kind in the tree without it — every other overlay, kwok profile, and component `values.yaml` has it. The other two new files here (`-dynamo`, `-kubeflow`) do have it, so this looks like the generator emitting two header variants. No CI gate catches it today.

mchmarny (changes requested):
Welcome to AICR. The mechanical parts of adding an accelerator to this repo are easy to get wrong and you got them right. `coverage_golden.yaml` is correctly regenerated, `pkg/tuning/compute_test.go` and `docs/integrator/components/nodewright.md` are updated in step, and the inheritance chain (`gb300-gke-cos-{training,inference}` -> `-kubeflow`/`-dynamo`) matches how the b200 family is laid out. `go test ./pkg/recipe/... ./pkg/tuning/...` passes clean at this head.

The three blockers are all in the same category — config that ships to users and looks right but does the wrong thing — rather than anything structural. Each is a small fix.

**Scope validated:** all 10 changed files read at `be8b984`; recipe resolution run locally for the gb300 dynamo/inference/training paths; `pkg/recipe` + `pkg/tuning` tests run; `apply-nodes.sh` profile selection traced against this PR's own Tier 2 CI logs.

**CI:** `Tier 1 / oke-ol-training (argocd-git)` and the dependent `KWOK Test Summary` are red, but that is infrastructure, not your change — the log shows `curl: (35) Recv failure: Connection reset by peer` followed by `Error: repo argo not found` while adding the Argo CD Helm repo, in an OKE lane this PR does not touch. A re-run should clear it. The PR is also currently `BEHIND` main and will need a rebase.

Findings inline. Happy to re-review as soon as you have pushed.

recipes/overlays/gb300-gke-cos-inference.yaml
yuanchen8911: Changing the intent to inference is necessary but not sufficient here. I pulled the pinned package (ghcr.io/nvidia/nodewright-packages/nvidia-tuning-gke:0.1.2, the digest referenced by tuning-gke.yaml) and its profiles/ directory contains only gb200/ and h100/ — there is no gb300 profile. The prepare script (skyhook_dir/prepare_nvidia_configs.sh) validates the accelerator against those directories and exits 1 with "ERROR: unknown accelerator 'gb300'" before the intent value is ever read. So with accelerator: gb300, Skyhook tuning fails on every GB300 node for both leaves — training included — whatever the intent says.

Two ways to resolve: ship GB300 with tuning disabled (keep the component wired, pin enabled: false, so enabling later is a one-line change), or point GB300 at the existing gb200 profile — GB300 shares the Grace host architecture, so the gb200 settings may well be applicable, but that combination has not been officially validated on GB300.

@ayuskauskas as the NodeWright maintainer — would you be comfortable with the gb200 profile on GB300 nodes, or should this ship disabled until a validated gb300 profile is published? Same applies to gb300-gke-cos-training.yaml.

recipes/overlays/gb300-gke-cos-training-kubeflow.yaml
```diff
@@ -0,0 +1,58 @@
...
+      valuesFile: components/kubeflow-trainer/values.yaml
+      manifestFiles:
+        - components/kubeflow-trainer/manifests/torch-distributed-cluster-training-runtime.yaml
+      dependencyRefs:
+        - cert-manager
+        - kube-prometheus-stack
+        - gpu-operator
+      overrides:
+        runtime:
+          image: "nvcr.io/nvidia/pytorch:25.01-py3-arm64"
```
yuanchen8911, line 53: These runtime.image / runtime.env values are not consumed anywhere: the chart has no runtime key, and the ClusterTrainingRuntime manifest (components/kubeflow-trainer/manifests/torch-distributed-cluster-training-runtime.yaml) templates only acceleratedNodeSelector/acceleratedTolerations — its container image is hardcoded to a digest-pinned pytorch/pytorch:2.11.0-cuda12.8 image. I verified the resolved recipe carries these overrides as dead values.

Two registry checks make this blocking for GB300: nvcr.io returns 404 for the 25.01-py3-arm64 tag (it does not exist; 25.01-py3 does), and the pinned pytorch/pytorch digest is a single-platform amd64 manifest, so the generated runtime cannot execute on A4X Max's arm64 hosts as shipped. The fix needs to parameterize the manifest image and pin a multi-arch digest from NVIDIA's validated GB300 stack — CUDA 13 / NVIDIA PyTorch 25.08 or newer.

recipes/overlays/gb300-gke-cos-inference-dynamo.yaml
```diff
@@ -0,0 +1,94 @@
...
+      valuesFile: components/dynamo-platform/values.yaml
+      dependencyRefs:
+        - grove
+        - cert-manager
+        - kube-prometheus-stack
+        - gpu-operator
+        - kai-scheduler
+      overrides:
+        engine:
+          cudaVersion: "13.0"
```
yuanchen8911, line 67: Chart ai-dynamo 1.2.1 does not consume engine.cudaVersion or engine.arch — the component values file has no engine key and no other dynamo overlay sets one, so these land in the release values unused. The underlying requirement is real: Dynamo requires CUDA 13 for B300/GB300, and the plain vllm-runtime:1.2.1 is CUDA 12.9, so it is not a supported GB300 runtime. But runtime selection happens in workload DynamoGraphDeployments, not these chart values. Suggest removing the block and adding a comment that GB300 model-serving workloads must use the CUDA 13 runtime variants — vllm-runtime:1.2.1-cuda13 is published multi-arch on NGC.

recipes/overlays/gb300-gke-cos-training.yaml
```diff
@@ -0,0 +1,96 @@
...
+      dependencyRefs:
+        - nfd
+        - cert-manager
+        - kube-prometheus-stack
+        - nodewright-customizations
+      overrides:
+        cdi:
+          enabled: true
+        gdrcopy:
+          enabled: true
```
yuanchen8911, line 40: Three of these keys need cleanup (same block in the inference leaf):

- gdrcopy.enabled: true — the inherited values-gke-cos.yaml sets this false with the comment "not supported on COS (host-managed driver)", and GPU Operator deploys GDRCopy as a sidecar of its driver pod, so with driver.enabled: false the flag cannot provision anything. The resolved recipe ships gdrcopy: true, which at best is inert and at worst contradicts the documented COS support stance. The block comment's claim that GDRCopy provides the cross-fabric transport should go with it.
- driver.version: 580.95.05 — dead pin on a disabled driver. The value matches Google's documented A4X Max host-driver floor, so it belongs in a comment or prerequisite doc, but nothing reads it here (and nothing can validate it today: the GPU collector emits no driver-version measurement).
- driver.nvidiaDriverRoot — not a gpu-operator chart key; that name belongs to the nvidia-dra-driver-gpu chart, which already inherits the correct /home/kubernetes/bin/nvidia from the gke-cos base. The operator's COS path is already set via hostPaths.driverInstallDir in values-gke-cos.yaml.

driver.enabled: false and cdi.enabled: true merely restate the inherited values, so after cleanup this override block may not be needed at all.

recipes/overlays/gb300-gke-cos-training.yaml
```diff
@@ -0,0 +1,96 @@
...
+          nvidiaDriverRoot: /home/kubernetes/bin/nvidia
+
+    # GKE A4/A4X/A5X Managed RDMA & GDRCopy Networking Topology:
+    # On GKE Blackwell and Vera systems, high-bandwidth inter-node communication over RoCE
+    # utilizes GPU Operator with GDRCopy enabled (`gdrcopy: enabled: true`) across GKE's
+    # native multi-NIC optical fabric. When node pools operate under GKE Managed RDMA
+    # (`--gpudirect-strategy=rdma`), GKE natively automates the lifecycle and host
+    # installation of Google InfiniBand (gIB) plugins and NCCL tuners (/usr/local/gib/).
+    # Consequently, standalone networking installer DaemonSets (gke-nccl-roce or
+    # gke-nccl-tcpxo) are intentionally bypassed to prevent overriding GKE managed networking.
```
yuanchen8911, line 53: This comment overstates what GKE provisions on A4X Max. The gIB-plugin/NCCL-tuner automation it describes is the A4/A4X flow; for A4X Max, Google's documentation requires deploying the asapd-lite DaemonSet (without it "child interfaces aren't created, and the dranet driver filters out the RDMA NICs"), Managed DRANET with ResourceClaimTemplates against deviceClassName mrdma.google.com, and ComputeDomain/IMEX setup for MNNVL.

To be fair about what AICR does cover: the recipe installs nvidia-dra-driver-gpu with ComputeDomains enabled, and the dra-support conformance check validates compute-domain ResourceSlices. What remains external and unvalidated is asapd-lite, the mrdma.google.com claims, and the behavioral ComputeDomain → ResourceClaimTemplate → IMEX flow (already tracked as TODO #1649 in the validator). Suggest correcting the comment and documenting asapd-lite/DRANET as external provisioning prerequisites (docs/integrator/eks-dynamo-networking.md is the precedent).

recipes/overlays/gb300-gke-cos-training.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+  criteria:
+    service: gke
+    accelerator: gb300
+    os: cos
+    intent: training
+
+  constraints:
+    - name: K8s.server.version
+      value: ">= 1.34.3-gke.1318000"
```
yuanchen8911, line 21: FYI, no change requested: 1.34.3-gke.1318000 is the documented A4X Max floor for the 1.34 track, but AICR's version comparison parses the -gke.N suffix into Extras and never compares it (pkg/version/doc.go), so this constraint enforces only >= 1.34.3 — a cluster at 1.34.3-gke.1000000 passes, and the separate 1.35 track floor (1.35.0-gke.2745000) cannot be expressed at all. A comment noting the suffix is informational would help; we'll file a follow-up for provider-aware build-version comparison.

yuanchen8911 (commented):
Reviewed at be8b984 with recipe resolution run locally, the pinned tuning package pulled from ghcr.io, and image manifests verified against registry APIs.

## Change requests

1. **NodeWright tuning cannot work as configured**: nvidia-tuning-gke:0.1.2 contains only gb200/ and h100/ profiles, and its prepare script rejects `accelerator: gb300` before intent is read, failing Skyhook tuning on every GB300 node. Either ship with `enabled: false` or reuse the gb200 profile — maintainer question in the nodewright thread.

2. **gb300-gke-cos-inference-dynamo sets `Deployment.gpu-operator.version` to `>= v24.6.0`**, which replaces (not tightens) the parent's `>= v25.10.0` on merge. GB300 support shipped in v25.10 — drop the constraint or set `>= v25.10.0`.

3. **The new GKE KWOK profiles are never selected**: apply-nodes.sh still maps gke/gb300 to eks/p5-h100, and this PR's Tier 2 CI logs show the GB300 lanes ran on amd64 H100 nodes. Wire them in or remove them. Also set `gpu.memory` to 285696 MiB (Google lists 279 GB per GPU; 245760 encodes only 240 GB).

4. **The Kubeflow `runtime.image`/`runtime.env` overrides are consumed by nothing**, the proposed 25.01-py3-arm64 tag does not exist on nvcr.io (404), and the manifest's hardcoded pytorch/pytorch digest is amd64-only — it cannot run on A4X Max's arm64 hosts. Parameterize the manifest image and pin a multi-arch CUDA 13 digest (NVIDIA PyTorch 25.08+).

## Cleanup (config that reads as functional but is not)

- Drop `gdrcopy.enabled: true` (inert without an operator-managed driver; values-gke-cos.yaml documents gdrcopy as unsupported on COS), the dead `driver.version` pin, and `driver.nvidiaDriverRoot` (not a gpu-operator key).
- Drop the dynamo leaf's `engine.cudaVersion`/`engine.arch` (chart 1.2.1 does not consume them); document instead that GB300 requires the CUDA 13 runtime variants (vllm-runtime:1.2.1-cuda13 is published multi-arch).
- Correct the Managed RDMA comments: A4X Max additionally requires the asapd-lite DaemonSet, mrdma.google.com resource claims, and ComputeDomain/IMEX setup — document these as external provisioning prerequisites.
- Remove unreachable gke-nccl-roce/values.yaml (no registry entry, no consumer); add the Apache header to the four files missing it.

Branch is behind main with a merge commit — please rebase onto origin/main and squash to a single signed/DCO commit.

recipes/overlays/gb300-gke-cos-inference.yaml
ayuskauskas: A GB200 profile on a GB300 should be fine. I agree on the intent though that should match.

recipes/overlays/gb300-gke-cos-training.yaml
```diff
@@ -0,0 +1,96 @@
...
+    # installation of Google InfiniBand (gIB) plugins and NCCL tuners (/usr/local/gib/).
+    # Consequently, standalone networking installer DaemonSets (gke-nccl-roce or
+    # gke-nccl-tcpxo) are intentionally bypassed to prevent overriding GKE managed networking.
+    - name: nodewright-customizations
+      type: Helm
+      manifestFiles:
+        - components/nodewright-customizations/manifests/tuning-gke.yaml
+      overrides:
+        service: gke
+        accelerator: gb300
```
ayuskauskas, line 60: Change this to be gb200 instead of gb300 as the nvidia-tuning-gke doesn't support a gb300 accelerator and they would use the same tuning.

recipes/overlays/gb300-gke-cos-inference.yaml
yuanchen8911: Thanks Alex — that settles it. So the resolution for both leaves is:

- keep nodewright-customizations enabled, but set accelerator: gb200 in the overrides so the package selects the existing GB200 profile
- fix the intent per leaf: inference here, multiNodeTraining in gb300-gke-cos-training.yaml

@rawsocket one knock-on for the same push: the tuning table in docs/integrator/components/nodewright.md and the expected rows in pkg/tuning/compute_test.go are generated from the recipes' accelerator override, so the gke/gb300 row added in this PR will change once the override becomes gb200 — please re-run make tuning-docs and update the test expectation in the same commit. A short comment in the overlays noting that gb300 intentionally maps to the gb200 tuning profile (per this thread) would also help the next reader.

mchmarny:
Still blocked on three functional issues at `2d01524`:

1. Restore the Dynamo deployment constraint to `Deployment.gpu-operator.version >= v25.10.0`; the leaf currently replaces the parent floor with `>= v24.6.0`, producing a false pass. [Thread](https://github.com/NVIDIA/aicr/pull/1985#discussion_r3705665561)
2. Point both GB300 NodeWright overrides at the supported `gb200` tuning profile. Keep `intent: inference` for inference and `intent: multiNodeTraining` for training, then regenerate the tuning docs/test expectation. Package `0.1.2` has no `gb300` profile and currently exits before applying tuning. [Thread](https://github.com/NVIDIA/aicr/pull/1985#discussion_r3705665576)
3. Make the Kubeflow `runtime.image` and `runtime.env` overrides feed the emitted `ClusterTrainingRuntime`, and pin a valid arm64 or multi-arch image. The current manifest still hardcodes an amd64 image for these arm64 nodes. [Thread](https://github.com/NVIDIA/aicr/pull/1985#discussion_r3706197549)

The previous KWOK profile-selection blocker is resolved by current `main`: label-based discovery now selects `gke/system-n2.yaml` and `gke/a4x-maxgpu-4g-metal.yaml`. The unused `gke-nccl-roce` values file and missing license headers remain non-blocking cleanup.

Focused local recipe/tuning/profile tests pass. Current-head CI is now running.

rawsocket:
Will come back to it at a later time. All the changes suggested here are implemented.

PR #1999, opened 2026-08-03
fix(aiperf-bench): move runtime to NVIDIA distroless python

mchmarny:
## Summary

Move the `aiperf-bench` runtime stage from `python:3.13-slim` to `nvcr.io/nvidia/distroless/python:3.13-v4.0.8` (pinned by digest), and replace the `/bin/sh -c` benchmark invocation with a shell-free Python entrypoint since the distroless base ships no shell.

## Motivation / Context

Requested move to the NVIDIA distroless Python base for the AIPerf benchmark image. The swap is not drop-in: the base has no shell, no apt, no pip, and no `useradd`, which breaks both the builder stage and the benchmark Job's runtime contract. This PR does the work those constraints require and re-validates the vulnerability posture end to end.

Fixes: N/A
Related: N/A

## Type of Change

- [ ] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Validator (`pkg/validator`) — `validators/performance` only
- [x] Other: `aiperf-bench` image, `.openvex.json`, `.grype.yaml`, `RELEASING.md`

## Implementation Notes

**Why only the runtime stage moved.** The distroless image has no shell, apt, or pip, so it cannot host `RUN` steps. The builder stage stays on `python:3.13-slim` and still compiles the venv; only the final stage changed. Both bases put CPython 3.13 at `/usr/local/bin`, so the venv's interpreter symlinks and compiled extensions (pyzmq, uvloop, crick) resolve unchanged.

**Replacing `/bin/sh -c`.** `buildAIPerfJob` chained `aiperf`, `echo`, and `cat` in a shell to produce the sentinel framing `parseAIPerfOutput` consumes. That framing moved to `validators/performance/aiperf_entrypoint.py` (stdlib only), invoked in exec form. Division of responsibility:

- **Go owns every benchmark flag.** The wrapper only prepends `aiperf profile` and appends the framed result. This keeps flag correctness — including the `--model` flag that aiperf 0.11.0 requires — asserted in Go tests rather than in an untested Python file.
- **The model is now a discrete argv element** instead of a `"$AICR_MODEL"` shell expansion. This is strictly safer: with no shell there is nothing to re-scan a metacharacter-bearing value. Verified by running the image with model `$(touch /tmp/pwned)` and confirming the payload arrives verbatim and does not execute.
- `AICR_MODEL` is still set on the container, now informational, so `kubectl describe pod` shows what a run benchmarked.

**Non-root user.** `RUN useradd` cannot run, so the image adopts the base's built-in `nvs` user (uid 1000) instead of minting uid 10001. Still non-root; verified `uid 1000` at runtime.

**Vulnerability posture.** The distroless base ships two Debian packages *older* than `python:3.13-slim`:

| | `python:3.13-slim` (before) | distroless (after, with VEX) |
|---|---|---|
| Critical | 0 | 0 |
| **High** | **2** | **2** |
| Medium | 14 | 15 |
| Low | 4 | 4 |
| VEX-suppressed | 11 | 32 |

- The 2 remaining Highs are the pre-existing CPython `CVE-2026-11940` / `CVE-2026-11972`. **The swap does not fix them** — both bases ship CPython 3.13.14.
- **libexpat1 2.7.1-2** contributes 21 findings (4 High). Suppressed in `.openvex.json` on two independently verified grounds: (1) *nothing in the image links it* — reading ELF `DT_NEEDED` for every binary under `/usr/local/lib`, `/usr/lib`, `/lib`, `/opt/venv/lib`, `/usr/local/bin`, `/usr/bin` returns zero references to libexpat; (2) CPython routes `xml.parsers.expat` through a `pyexpat` whose `DT_NEEDED` is `['libc.so.6']` only, with expat statically bundled at 2.8.1. aiperf itself has zero XML imports across its 682 `.py` files. Suppression is scoped to the unused system library; for the 13 CVEs fixed only in 2.8.2 the impact statement explicitly states the bundled 2.8.1 is *not* itself patched rather than overclaiming.
- **liblzma5 5.8.1-1 (`CVE-2026-34743`, Medium) is deliberately left unsuppressed.** `_lzma.so` genuinely links it, so the evidence is weaker than for expat, and `python:3.13-slim` already ships the fixed `5.8.1-1+deb13u1`. Leaving it visible preserves the signal that the base image is stale. **This is the one net-new finding in this PR.**

**Recommended follow-up:** file upstream with the NVIDIA distroless team to refresh the Debian layer — `3.13-v4.0.8` was built 2026-06-11 and both stale packages have fixes available in Debian trixie.

**VEX housekeeping.** Dropped 11 `aicr-gate` statements (`GO-2026-5005/5006/5013/5014/5015/5017/5018/5019/5020/5021/5023`) confirmed *entirely absent* from the current scan after a chainsaw bump — verified individually, not inferred from a count delta. Document bumped to version 10 with a refreshed timestamp and tooling note. The remaining `aicr-gate` statements still apply (count-neutral).

**Note for reviewers:** two unsuppressed Highs remain on `aicr-gate` — `GO-2026-5970` (x/text) and `GO-2026-5942` (x/net). Those want a dependency bump, not a VEX entry, and are out of scope here.

## Testing

```bash
make qualify   # exit 0
```

- `make qualify` passes clean on the final tree.
- `golangci-lint run -c .golangci.yaml ./validators/performance/...` → 0 issues.
- `go test -race ./validators/performance/` passes; coverage 56.4% (unchanged — `validators/` is excluded from the project floor per #1752).
- Image built for **both** `linux/amd64` and `linux/arm64`; `aiperf --version` → `0.11.0` in the distroless runtime.
- Wrapper exercised end-to-end inside the real distroless image with a stub `aiperf`:
  - happy path emits `sentinel / JSON / sentinel` and parses correctly;
  - the newline guard is covered (aiperf writes the export without a trailing newline, so without it the closing sentinel would share a line with the JSON);
  - non-zero aiperf exit propagates verbatim (exit 7) with no sentinel emitted;
  - missing result file is a hard error rather than an empty sentinel block;
  - missing required env fails closed;
  - runs as uid 1000.
- New regression test ties the wrapper's exact output shape to `parseAIPerfOutput`.
- Grype re-scan used the CI-pinned v0.110.0 with the same `--only-fixed --vex .openvex.json -c .grype.yaml` flags, against an image tagged so the PURL resolves to `pkg:oci/aicr-aiperf-bench`; all 4 new High suppressions confirmed landing under `appliedIgnoreRules` with `namespace = "vex"`.

## Risk Assessment

**Rollout notes:** The benchmark Job's `Command`/`Args` shape changes, so a released `aicr` binary must run against a matching `aiperf-bench` image. The validator resolves the inner image from the same tag as the outer validator (`resolveAiperfImage` / `AICR_VALIDATOR_IMAGE_TAG`), so they move together in normal use. A pinned older `aiperf-bench` image combined with a new binary would fail at pod start on the missing `/opt/aicr/aiperf_entrypoint.py`. Revert is a straight rollback of the Dockerfile and Go change.

## Checklist

### DIFF
--- a/.openvex.json
+++ b/.openvex.json
@@ -3,9 +3,9 @@
   "@id": "https://github.com/NVIDIA/aicr/.openvex.json",
   "author": "NVIDIA AICR maintainers",
   "role": "document creator",
-  "timestamp": "2026-07-24T00:00:00Z",
-  "version": 9,
-  "tooling": "manual; aiperf-bench statements verified against aiperf v0.11.0 source (which pins pillow~=12.2.0) — the GHSA-whj4-6x5x-4v2j and GHSA-pwv6-vv43-88gr statements were dropped because their CVEs are fixed in pillow 12.2.0 and no longer present; aicr statement reachability verified by source inspection (CGO-free ko build, no libssl linkage); aicr-gate statements suppress CVEs in the embedded upstream kyverno/chainsaw binary (affected packages identified via vuln.go.dev), justified by chainsaw's ephemeral cluster-internal readiness-gate usage",
+  "timestamp": "2026-08-03T00:00:00Z",
+  "version": 10,
+  "tooling": "manual; aiperf-bench statements verified against aiperf v0.11.0 source (which pins pillow~=12.2.0); the aiperf-bench runtime base moved from python:3.13-slim to nvcr.io/nvidia/distroless/python:3.13-v4.0.8, adding 21 libexpat1 2.7.1-2 statements (4 High, 16 Medium, 1 Low) whose unreachability was verified by reading ELF DT_NEEDED across the built image (no binary links libexpat) and by confirming CPython's pyexpat statically bundles expat 2.8.1; aicr statement reachability verified by source inspection (CGO-free ko build, no libssl linkage); aicr-gate statements suppress CVEs in the embedded upstream kyverno/chainsaw binary (affected packages identified via vuln.go.dev), justified by chainsaw's ephemeral cluster-internal readiness-gate usage; v10 dropped 11 redundant aicr-gate statements (GO-2026-5005/5006/5013/5014/5015/5017/5018/5019/5020/5021/5023): the 2026-08-03 scan of ghcr.io/nvidia/aicr-gate reports each of those CVEs under its GHSA identifier instead of the go.dev one, and the GHSA-aliased statement for every one of them is retained here (e.g. GO-2026-5005 / CVE-2026-39833 is covered by GHSA-jppx-rxg9-jmrx), so no suppression coverage is lost — chainsaw is unchanged at v0.2.15 and the vulnerable golang.org/x/crypto/ssh code is still present in the embedded binary, so the retained GHSA statements must not be removed",
   "statements": [
     {
       "vulnerability": {
@@ -49,41 +49,7 @@
     },
     {
       "vulnerability": {
-        "name": "GO-2026-5005",
-        "description": "chainsaw (embedded kyverno binary): key constraints not enforced in golang.org/x/crypto/ssh/agent (CVE-2026-39833, GO-2026-5005)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
         "name": "GHSA-f5wc-c3c7-36mc",
-        "description": "chainsaw (embedded kyverno binary): agent constraints dropped when forwarding keys in golang.org/x/crypto/ssh/agent (CVE-2026-39832, GO-2026-5006)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
-        "name": "GO-2026-5006",
         "description": "chainsaw (embedded kyverno binary): agent constraints dropped when forwarding keys in golang.org/x/crypto/ssh/agent (CVE-2026-39832, GO-2026-5006)."
       },
       "products": [
@@ -117,41 +83,7 @@
     },
     {
       "vulnerability": {
-        "name": "GO-2026-5013",
-        "description": "chainsaw (embedded kyverno binary): byte arithmetic underflow and panic in golang.org/x/crypto/ssh (CVE-2026-46597, GO-2026-5013)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
         "name": "GHSA-vgwf-h737-ff37",
-        "description": "chainsaw (embedded kyverno binary): client can cause server deadlock on unexpected responses in golang.org/x/crypto/ssh (CVE-2026-39830, GO-2026-5017)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
-        "name": "GO-2026-5017",
         "description": "chainsaw (embedded kyverno binary): client can cause server deadlock on unexpected responses in golang.org/x/crypto/ssh (CVE-2026-39830, GO-2026-5017)."
       },
       "products": [
@@ -185,41 +117,7 @@
     },
     {
       "vulnerability": {
-        "name": "GO-2026-5018",
-        "description": "chainsaw (embedded kyverno binary): pathological RSA/DSA parameters may cause DoS in golang.org/x/crypto/ssh (CVE-2026-39829, GO-2026-5018)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
         "name": "GHSA-89gr-r52h-f8rx",
-        "description": "chainsaw (embedded kyverno binary): bypass of FIDO/U2F security key physical interaction in golang.org/x/crypto/ssh (CVE-2026-39831, GO-2026-5019)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
-        "name": "GO-2026-5019",
         "description": "chainsaw (embedded kyverno binary): bypass of FIDO/U2F security key physical interaction in golang.org/x/crypto/ssh (CVE-2026-39831, GO-2026-5019)."
       },
       "products": [
@@ -253,23 +151,6 @@
     },
     {
       "vulnerability": {
-        "name": "GO-2026-5020",
-        "description": "chainsaw (embedded kyverno binary): infinite loop on large channel writes in golang.org/x/crypto/ssh (CVE-2026-39834, GO-2026-5020)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
         "name": "GHSA-5cgq-3rg8-m6cv",
         "description": "chainsaw (embedded kyverno binary): auth bypass via unenforced @revoked status in golang.org/x/crypto/ssh/knownhosts (CVE-2026-42508, GO-2026-5021)."
       },
@@ -287,41 +168,7 @@
     },
     {
       "vulnerability": {
-        "name": "GO-2026-5021",
-        "description": "chainsaw (embedded kyverno binary): auth bypass via unenforced @revoked status in golang.org/x/crypto/ssh/knownhosts (CVE-2026-42508, GO-2026-5021)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
         "name": "GHSA-x527-x647-q7gg",
-        "description": "chainsaw (embedded kyverno binary): VerifiedPublicKeyCallback permissions skip enforcement in golang.org/x/crypto/ssh (CVE-2026-46595, GO-2026-5023)."
-      },
-      "products": [
-        {
-          "@id": "pkg:oci/aicr-gate",
-          "identifiers": {
-            "purl": "pkg:oci/aicr-gate"
-          }
-        }
-      ],
-      "status": "not_affected",
-      "justification": "vulnerable_code_not_in_execute_path",
-      "impact_statement": "The aicr-gate image embeds the upstream, cosign-attested kyverno/chainsaw binary (/usr/local/bin/chainsaw, CHAINSAW_VERSION in .settings.yaml) alongside the AICR gate binary; see cmd/gate/Dockerfile. The vulnerable code is present ONLY in that embedded chainsaw binary -- the AICR gate binary is clean (built with Go 1.26.5, golang.org/x/crypto v0.53.0, golang.org/x/net v0.56.0). chainsaw v0.2.15 (the latest release) was built with an older Go toolchain and older golang.org/x/{crypto,net}; kyverno has not yet shipped a rebuilt release. Within AICR, chainsaw runs exclusively as an ephemeral, cluster-internal readiness-gate Job: the bundler launches it to `chainsaw test` recipe-authored assertions, talking only to the trusted in-cluster Kubernetes API server. It exposes no network listener, accepts no untrusted ingress, and processes only operator-authored manifests and trusted API-server responses. chainsaw is a Kubernetes test runner and never establishes SSH client, server, or agent connections; the golang.org/x/crypto/ssh* code path is not reached during `chainsaw test`."
-    },
-    {
-      "vulnerability": {
-        "name": "GO-2026-5023",
         "description": "chainsaw (embedded kyverno binary): VerifiedPublicKeyCallback permissions skip enforcement in golang.org/x/crypto/ssh (CVE-2026-46595, GO-2026-5023)."
       },
       "products": [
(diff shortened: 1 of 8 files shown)


.openvex.json
```diff
@@ -3,9 +3,9 @@
...
+  "timestamp": "2026-08-03T00:00:00Z",
+  "version": 10,
+  "tooling": "manual; aiperf-bench statements verified against aiperf v0.11.0 source (which pins pillow~=12.2.0); the aiperf-bench runtime base moved from python:3.13-slim to nvcr.io/nvidia/distroless/python:3.13-v4.0.8, adding four libexpat1 2.7.1-2 statements whose unreachability was verified by reading ELF DT_NEEDED across the built image (no binary links libexpat) and by confirming CPython's pyexpat statically bundles expat 2.8.1; aicr statement reachability verified by source inspection (CGO-free ko build, no libssl linkage); aicr-gate statements suppress CVEs in the embedded upstream kyverno/chainsaw binary (affected packages identified via vuln.go.dev), justified by chainsaw's ephemeral cluster-internal readiness-gate usage; v10 dropped 11 aicr-gate statements (GO-2026-5005/5006/5013/5014/5015/5017/5018/5019/5020/5021/5023) confirmed absent from the 2026-08-03 scan of ghcr.io/nvidia/aicr-gate after a chainsaw bump",
```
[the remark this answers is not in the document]
mchmarny: Fixed in 4517c5d7. The count was written when only the four High statements existed and never updated after the medium/low ones were appended — the field now reads "adding 21 libexpat1 2.7.1-2 statements (4 High, 16 Medium, 1 Low)", which matches `jq '[.statements[] | select(.vulnerability.description | test("libexpat1"))] | length' .openvex.json` → 21.

validators/performance/aiperf_entrypoint.py
```diff
@@ -0,0 +1,108 @@
...
+
+    # No shell: argv is passed straight to execvp, so no element is re-parsed.
+    # stdout/stderr are inherited so benchmark progress reaches the pod log.
+    completed = subprocess.run(cmd, check=False)
+    if completed.returncode != 0:
+        print(
+            f"aiperf-entrypoint: {AIPERF_BIN} exited {completed.returncode}",
+            file=sys.stderr,
+        )
+        return completed.returncode
```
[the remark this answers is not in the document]
mchmarny: Fixed in 4517c5d7, with one change from the suggested diff: the handler returns **127** rather than 1.

Reproduced first — with `PATH=/nonexistent` the old code emitted a raw `FileNotFoundError: [Errno 2] No such file or directory: 'aiperf'` traceback, unlike every other path in the file.

127 is what the `/bin/sh -c` predecessor returned for "command not found", so any operator runbook or log triage keyed on that code keeps working across this change. Caught as `OSError` so a non-executable binary (`PermissionError`) is covered too.

Verified in the rebuilt image: missing binary → exit 127 with `aiperf-entrypoint: cannot execute aiperf: ...` and no traceback; happy path → exit 0 with correct sentinel framing; aiperf failure → exit 7 still propagates verbatim.

validators/performance/inference_perf_test.go
```diff
@@ -896,19 +912,55 @@ func TestBuildAIPerfJob_PrebuiltImageAndSentinel(t *testing.T) {
...
+	// would fail at pod start with "exec: /bin/sh: no such file or directory".
+	wantCommand := []string{aiperfEntrypointPython, aiperfEntrypointScript}
+	if !reflect.DeepEqual(container.Command, wantCommand) {
+		t.Errorf("container.Command = %v, want %v", container.Command, wantCommand)
+	}
+	for _, arg := range append(append([]string{}, container.Command...), container.Args...) {
+		if strings.HasSuffix(arg, "/sh") || strings.HasSuffix(arg, "/bash") || arg == "-c" {
+			t.Errorf("argv element %q implies a shell; the distroless runtime has none", arg)
+		}
+	}
```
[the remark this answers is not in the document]
mchmarny: Extracted in 4517c5d7 as `assertNoShellIndicators(t, ctr v1.Container, why string)`. It takes the container rather than a pre-joined argv so both call sites stop repeating the `append(append([]string{}, Command...), Args...)` dance, and it is marked `t.Helper()` so failures report the caller line.

mchmarny:
All four review items addressed in 4517c5d7 (force-push; `make qualify` exit 0 on the result).

| Item | Resolution |
|---|---|
| `.openvex.json` tooling count said "four" libexpat1 statements | Corrected to 21 (4 High, 16 Medium, 1 Low) — verified against the file |
| `aiperf_entrypoint.py` uncaught `FileNotFoundError` | Catches `OSError`, exits **127** (what the `/bin/sh -c` predecessor returned for command-not-found) rather than 1 |
| Duplicated shell-indicator check in tests | Extracted to `assertNoShellIndicators` |
| `.grype.yaml` stale `Reviewed:` date (outside-diff comment) | Refreshed `2026-06-01` → `2026-08-03`, and added `libexpat1` to the suppressed-package list in the same comment block |

The force-push also corrected a mistake of mine: an over-eager squash had absorbed upstream #1987 (the codeql-action bump) into this branch. The branch is now a single commit parented directly on `origin/main` touching exactly the 7 intended files.

Re-verified after the changes: wrapper exit paths in the rebuilt distroless image (missing binary → 127, happy path → 0 with correct sentinel framing, aiperf failure → 7), `golangci-lint` 0 issues, `go test -race ./validators/performance/` green.

.openvex.json
```diff
@@ -1211,6 +1024,489 @@
...
...g in the image links it: reading ELF DT_NEEDED entries for every binary under /usr/local/lib, /usr/lib, /lib, /opt/venv/lib, /usr/local/bin, and /usr/bin returns zero references to libexpat, so /usr/lib/x86_64-linux-gnu/libexpat.so.1 (2.7.1-2) is never loaded. (2) The one XML path that does execute does not use it: CPython routes xml.parsers.expat through lib-dynload/pyexpat.cpython-313-x86_64-linux-gnu.so, whose DT_NEEDED is ['libc.so.6'] only because expat is statically bundled at EXPAT_VERSION 2.8.1 \u2014 a version at or above the fixed version for all four CVEs. So the 14 venv packages that import xml.* (PIL, matplotlib, pandas, fontTools, pyarrow, datasets, ...) run patched 2.8.1 code, not 2.7.1-2. aiperf itself imports no XML module at all: scanning its 682 .py files for '^(import|from) (xml|pyexpat|xmlrpc)' returns zero hits. AICR invokes the image only as `aiperf profile <text-LLM> --url <endpoint>` (buildAIPerfJob in validators/performance/inference_perf_constraint.go), exchanging JSON over HTTP; no XML document from any source reaches a parser. The statically bundled expat is 2.8.1, above this CVE's 2.7.2 fix threshold, so even the reachable XML path is already patched."
```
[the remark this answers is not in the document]
mchmarny: Fixed in 84cbe6c5 — the phrase was written when the document carried only the four High statements and was never revisited after the medium/low ones landed.

One correction to the rationale: the split is **8 patched / 13 not**, not nine. Verified by parsing each statement’s own `before <version>` text against the bundled 2.8.1:

- bundled 2.8.1 **is** at/above the fix for 8: `CVE-2025-59375` (2.7.2), `CVE-2026-45186` (2.8.1), `CVE-2026-41080` (2.8.0), `CVE-2026-25210` (2.7.4), `CVE-2026-24515` (2.7.4), `CVE-2026-32776`/`32777`/`32778` (2.7.5)
- bundled 2.8.1 is **not** patched for the 13 fixed in 2.8.2

Rather than restate a per-CVE threshold inside the shared block, I removed the claim from it entirely — each of those four statements already ended with its own correct threshold sentence, so the suggested edit would have duplicated it. The shared text now reads:

> expat is statically bundled at EXPAT_VERSION 2.8.1. So the 14 venv packages that import xml.* (…) exercise that bundled build, not the flagged 2.7.1-2 package. Whether 2.8.1 is itself patched varies by CVE and is stated per statement below.

All 21 libexpat statements now carry an accurate per-CVE claim: 8 name their fix threshold, 13 say the bundled copy is not itself patched and rest solely on the "nothing links the flagged package" argument. Re-scanned after the edit — all 21 still suppress (`vex-suppressed: 32`, zero libexpat in `matches`), High unchanged at 2, `make qualify` exit 0.

validators/performance/aiperf_entrypoint.py
```diff
@@ -0,0 +1,119 @@
...
+            f"aiperf-entrypoint: cannot execute {AIPERF_BIN}: {err}",
+            file=sys.stderr,
+        )
+        return 127
+    if completed.returncode != 0:
+        print(
+            f"aiperf-entrypoint: {AIPERF_BIN} exited {completed.returncode}",
+            file=sys.stderr,
+        )
+        return completed.returncode
```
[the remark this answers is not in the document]
mchmarny: Fixed in 1612e034. Confirmed the regression before changing anything:

```
$ python3 -c "import subprocess,sys; print(subprocess.run([sys.executable,\"-c\",\"import sys; sys.exit(-9)\"]).returncode)"
247
$ /bin/sh -c 'set -e; /bin/sh -c "kill -KILL \$\$"'; echo $?
137
```

So the wrapper really did diverge from the `/bin/sh -c` predecessor on signal death, and it matters concretely — the benchmark is a load generator, so an OOMKilled run is a realistic case and pod-level triage keys on 137.

Negative return codes now map to `128 + signum` in a `_report_failure` helper, which also names the signal in the diagnostic. The module docstring now states the full exit-status contract rather than just "propagates verbatim".

On the regression test: I did not add one. The repo has no Python test harness — no `pytest.ini`/`tox.ini`/`pyproject.toml`/`conftest.py`, and neither the Makefile nor CI invokes pytest — so introducing one for a ~110-line wrapper is disproportionate to this PR. Instead I verified all four statuses against the real distroless image, which exercises the actual shipped artifact rather than a mock:

| case | exit | stderr |
|---|---|---|
| aiperf SIGKILLed | **137** | `aiperf killed by signal 9 (SIGKILL)` |
| aiperf exits 7 | **7** | `aiperf exited 7` |
| happy path | **0** | correct sentinel framing |
| aiperf missing | **127** | `cannot execute aiperf: [Errno 2] ...` |

`make qualify` exit 0. If you would rather have a real pytest suite for this wrapper, that is a reasonable follow-up but wants its own PR to add the harness and wire it into CI.

njhensley (commented):
## 📋 Multi-Persona Review — ✅ Approve with comments

**Method:** 4 independent persona reviewers (Correctness · Security/Supply-chain · Domain & Architecture · Test-coverage) → an adversarial senior meta-reviewer that re-derived every finding from the resolved code at `1612e034`.

This is a well-executed, honestly-documented base-image swap. The meta-reviewer independently reproduced and upheld the load-bearing claims:

- **Injection surface is genuinely eliminated** — `subprocess.run([bin, "profile", *argv], check=False)`, `shell=False`, no `os.system`; the model is a discrete argv element handed to execvp, so a metacharacter-bearing name is inert with no quoting.
- **Exit-code parity with the removed `/bin/sh -c` holds** — 127 exec-not-found, 128+N signal (OOMKilled still reports 137), verbatim non-zero, fail-closed on a missing result file. Benchmark-flag equivalence is exact.
- **The VEX work is rigorous, not hand-wavy** — the 13 libexpat CVEs fixed only in 2.8.2 explicitly state the statically-bundled 2.8.1 is *not* itself patched and fall back to the independent unreachability argument, rather than a false "already patched." **No suppression coverage is actually lost** by the 11 dropped `GO-` statements — each has a retained GHSA twin for the same CVE.
- Non-root posture preserved (uid 1000 `nvs`), digest pin correct, `RELEASING.md` updated.

**No blocker or major finding survived adjudication.** Two 🟡 items are worth addressing (a `py_compile`/test gate for the Python wrapper, and a factual correction to the `.openvex.json` tooling note); the rest is polish. Inline comments follow.

- Injection surface — genuinely eliminated.
- libexpat suppressions — not over-claimed; the 2.8.2-only CVEs correctly disclaim "already patched."
- Dropped `GO-` statements — no coverage lost; retained GHSA twins cover the same CVEs.
- Exit-code parity + fail-closed on missing result file — holds.
- Non-root (uid 1000 nvs), digest pin, benchmark-flag equivalence — all sound.

### Tier summary

| 0 | 0 | 3 | 6 | **Approve with comments** |

<sub>Reviewed with a multi-persona + adversarial meta-reviewer workflow. Findings are anchored to the diff at `1612e034`.</sub>

validators/performance/aiperf_entrypoint.py
```diff
@@ -0,0 +1,142 @@
...
+        print(
+            f"aiperf-entrypoint: {AIPERF_BIN} killed by signal {signum} ({name})",
+            file=sys.stderr,
+        )
+        return status
+    print(f"aiperf-entrypoint: {AIPERF_BIN} exited {returncode}", file=sys.stderr)
+    return returncode
+
+
+def main() -> int:
```
njhensley, line 90: **🟡 Minor — Python wrapper has no automated test or compile gate (CI is Go-only)**

The exit-code mapping (127 exec-not-found, 128+N signal, verbatim non-zero, 2 env-require, 1 missing-file), the signal-name lookup, and the sentinel framing all live in Python, and the repo has no Python test infrastructure (no pytest/conftest/pyproject) and no `py_compile`/import gate in the image build. A syntax error or logic regression builds a perfectly valid image and only fails when a benchmark Job execs the script.

**Blast radius:** A regression surfaces at the most expensive point — a live multi-GPU inference-perf run, after image build and workload deploy. `validators/` is excluded from the 80% floor (#1752) and the path fails closed, which is why this is minor rather than major.

**Fix:** Add a build-time `RUN python -m py_compile validators/performance/aiperf_entrypoint.py` in the builder stage (catches syntax regressions for free), and/or a Go test that shells out to `python3 aiperf_entrypoint.py` with a stub `aiperf` on PATH (cases: success framing byte-shape, exit 127/2/1/137) to lock the exit-code contract. (Meta-review retiered 🟠→🟡.)

.openvex.json
```diff
@@ -3,9 +3,9 @@
...
+  "timestamp": "2026-08-03T00:00:00Z",
+  "version": 10,
+  "tooling": "manual; aiperf-bench statements verified against aiperf v0.11.0 source (which pins pillow~=12.2.0); the aiperf-bench runtime base moved from python:3.13-slim to nvcr.io/nvidia/distroless/python:3.13-v4.0.8, adding 21 libexpat1 2.7.1-2 statements (4 High, 16 Medium, 1 Low) whose unreachability was verified by reading ELF DT_NEEDED across the built image (no binary links libexpat) and by confirming CPython's pyexpat statically bundles expat 2.8.1; aicr statement reachability verified by source inspection (CGO-free ko build, no libssl linkage); aicr-gate statements suppress CVEs in the embedded upstream kyverno/chainsaw binary (affected packages identified via vuln.go.dev), justified by chainsaw's ephemeral cluster-internal readiness-gate usage; v10 dropped 11 aicr-gate statements (GO-2026-5005/5006/5013/5014/5015/5017/5018/5019/5020/5021/5023) confirmed absent from the 2026-08-03 scan of ghcr.io/nvidia/aicr-gate after a chainsaw bump",
```
njhensley: **🟡 Minor — .openvex.json tooling note attributes the drop to a chainsaw bump that never happened**

The `tooling` note says the 11 dropped aicr-gate statements were "confirmed absent from the 2026-08-03 scan … after a chainsaw bump." But `.settings.yaml` is untouched by this PR and still pins `chainsaw: v0.2.15`, and the retained ssh statements still read "chainsaw v0.2.15 (the latest release) … kyverno has not yet shipped a rebuilt release" — directly contradicting a bump. The real reason the drop is safe: each dropped `GO-2026-50xx` id is a duplicate of a retained GHSA statement for the identical CVE (e.g. GO-2026-5005/CVE-2026-39833 ↔ retained GHSA-jppx-rxg9-jmrx), so Grype now matches under the GHSA id and no suppression coverage is lost.

**Blast radius:** No security regression today, but a misleading audit trail: a future maintainer reading "chainsaw bump fixed these" could delete the retained GHSA twins, unmasking real High/Medium `golang.org/x/crypto/ssh` findings in the embedded chainsaw binary (the vulnerable code is still shipped; only its scanner identifier changed).

**Fix:** Correct the rationale, e.g. "dropped 11 redundant GO-advisory-id statements; the same CVEs remain suppressed under their GHSA identifiers (the scanner reports them under GHSA, not the go.dev namespace). Chainsaw is unchanged at v0.2.15 and the vulnerable ssh code is still present in the embedded binary."

validators/performance/inference_perf_test.go
```diff
@@ -896,19 +912,51 @@ func TestBuildAIPerfJob_PrebuiltImageAndSentinel(t *testing.T) {
 		t.Errorf("aiperfBaseImage %q should be the pre-built ghcr image", aiperfBaseImage)
 	}
 
-	script := container.Args[0]
-	if strings.Contains(script, "pip install") {
-		t.Errorf("script should not pip install at runtime — aiperf is baked into the image; got:\n%s", script)
+	// The runtime image is distroless (no /bin/sh), so the Job must invoke the
+	// baked Python framing wrapper in exec form. A regression back to a shell
+	// would fail at pod start with "exec: /bin/sh: no such file or directory".
+	wantCommand := []string{aiperfEntrypointPython, aiperfEntrypointScript}
```
njhensley, line 928: **🟡 Minor — Dockerfile↔Go path-constant coupling test is tautological**

`wantCommand := []string{aiperfEntrypointPython, aiperfEntrypointScript}` compares `container.Command` against the exact constants `buildAIPerfJob` uses to build it — constant vs. constant, so it passes by construction and can never detect a mismatch. The real coupling is between those Go constants (`/opt/venv/bin/python`, `/opt/aicr/aiperf_entrypoint.py`) and the Dockerfile `COPY … /opt/aicr/aiperf_entrypoint.py` destination + venv layout, and no test bridges the two.

**Blast radius:** A future edit that moves the COPY destination or the venv path without mirroring the constant passes `make qualify` cleanly, then fails at pod start (`python: can't open file '…'`) — caught only in UAT/prod.

**Fix:** Add a small test that reads `aiperf-bench.Dockerfile` and asserts it contains a `COPY …aiperf_entrypoint.py <aiperfEntrypointScript>` line and that `/opt/venv/bin` is on PATH.

validators/performance/inference_perf_test.go
```diff
@@ -290,8 +291,23 @@ func TestParseAIPerfOutput(t *testing.T) {
 			wantTTFT:       84.1,
 		},
 		{
+			// Byte-for-byte shape emitted by aiperf_entrypoint.py: benchmark
+			// chatter, then sentinel, then the exported JSON, then sentinel.
+			// aiperf writes profile_export_aiperf.json without a trailing
+			// newline, so the wrapper inserts one — without that guard the
```
njhensley: **🔵 Nitpick — Test comment claims the trailing-newline guard is required to parse — it is not**

The comment says that without the wrapper's trailing-newline guard "the closing sentinel would share a line with the JSON's final brace and this parse would fail." `parseAIPerfOutput` locates the sentinels with `strings.Index` (substring, not line-based) and `TrimSpace`s the payload, so the glued shape `…95.3}===AIPERF-RESULT===` parses fine. The case also hand-adds the `\n` in its literal, so it never exercises the guard-absent path.

**Blast radius:** Low — false confidence only; the parser is robust either way and the wrapper's guard remains a harmless nicety.

**Fix:** Drop the claim, or move the assertion to a path that actually removes the guard and observes the difference.

validators/performance/inference_perf_test.go
```diff
@@ -290,8 +291,23 @@ func TestParseAIPerfOutput(t *testing.T) {
...
+			// closing sentinel would share a line with the JSON's final brace
+			// and this parse would fail.
+			name: "wrapper framing with no trailing newline in export",
```
njhensley, line 301: **🔵 Nitpick — Parse-shape test hand-encodes wrapper output instead of executing it**

The "wrapper framing with no trailing newline" case is a Go-authored reconstruction of what `aiperf_entrypoint.py` is assumed to emit; nothing binds it to the Python's real output. They coincide today (`}\n===AIPERF-RESULT===\n`) but the emitter and parser can drift with no failing test.

**Blast radius:** The one seam between the Python emitter and the Go parser has no end-to-end regression guard. Same root cause as the Python-untested finding, hence nitpick (meta-review retiered 🟡→🔵).

**Fix:** Derive the parser input from an actual `python3 aiperf_entrypoint.py` run so the same bytes the wrapper emits are the bytes `parseAIPerfOutput` consumes.

validators/performance/inference_perf_constraint.go
```diff
@@ -2930,44 +2948,35 @@ func buildAIPerfJob(namespace, jobName, endpoint, model string, concurrency int,
...
+		"--request-count", strconv.Itoa(requestCount),
+		"--warmup-request-count", strconv.Itoa(warmupCount),
+		"--prompt-input-tokens-mean", strconv.Itoa(inputTokensMean),
+		"--prompt-input-tokens-stddev", "0",
+		"--prompt-output-tokens-mean", strconv.Itoa(outputTokensMean),
+		"--prompt-output-tokens-stddev", "0",
+		"--num-dataset-entries", strconv.Itoa(aiperfNumDatasetEntries),
+		"--random-seed", strconv.Itoa(aiperfRandomSeed),
+		"--extra-inputs", "temperature:0",
+		"--output-artifact-dir", aiperfArtifactDir,
```
njhensley, line 2977: **🔵 Nitpick — Artifact dir double-sourced (argv + env) from the same constant, no guard**

aiperf is told where to write via the argv `--output-artifact-dir aiperfArtifactDir` (this line) and the wrapper is told where to read via `AICR_AIPERF_ARTIFACT_DIR = aiperfArtifactDir` (line 3025). Both resolve from the same Go constant, so intra-Go drift is near-impossible today.

**Blast radius:** Latent foot-gun: a future change adding a per-run subdir to only the argv would make the wrapper read a stale path and return 1 ("cannot read result file") on an otherwise-successful benchmark.

**Fix:** Optionally add a build-job assertion that the `--output-artifact-dir` argv value equals the `AICR_AIPERF_ARTIFACT_DIR` env value.

.openvex.json
njhensley: **🔵 Nitpick — "Confirmed absent from scan" is un-reproducible / namespace-fragile**

Whether Grype's VEX matcher resolves a `GO-2026-xxxx`-reported finding against a GHSA-only statement is alias-resolution behavior this review can't re-run. If a future Grype DB revision re-emits the ssh CVEs under the go.dev namespace, the dropped statements no longer match by id and the findings resurface.

**Blast radius:** Bounded — the image scan job runs `fail-build: false` as a daily Slack notification, not a PR-blocking gate, so worst case is noise. Overlaps the tooling-note finding.

**Fix:** Optionally keep one GO-namespace statement as belt-and-suspenders if alias matching is ever in doubt.

validators/performance/inference_perf_constraint.go
```diff
@@ -137,6 +137,23 @@ const (
...
+	// Python instead. Both paths are fixed by aiperf-bench.Dockerfile — keep
+	// them in sync with the COPY destination and venv layout there.
+	aiperfEntrypointPython = "/opt/venv/bin/python"
+	aiperfEntrypointScript = "/opt/aicr/aiperf_entrypoint.py"
+
+	// envAIPerfArtifactDir and envAIPerfResultMarker configure
+	// aiperf_entrypoint.py's framing. envAIPerfModel is informational only —
+	// the model reaches aiperf as an explicit --model argv element — and is
+	// kept so `kubectl describe pod` shows what a run benchmarked.
+	envAIPerfModel        = "AICR_MODEL"
```
njhensley, line 153: **🔵 Nitpick — AICR_MODEL env is now dead-but-set; generic name beside the live override**

`envAIPerfModel = "AICR_MODEL"` is no longer read by anything (the model now arrives as a `--model` argv element); it is set purely for `kubectl describe` visibility, which the comment documents. It sits next to the live override `AICR_INFERENCE_PERF_MODEL` and the wrapper's `AICR_AIPERF_*` vars, so the generic prefix invites confusion with the real knob.

**Blast radius:** Cosmetic/operability; no functional effect.

**Fix:** Rename to `AICR_AIPERF_MODEL` for prefix consistency, or leave as-is with the existing comment.

validators/performance/aiperf-bench.Dockerfile
```diff
@@ -60,30 +72,40 @@ ENV PATH="/opt/venv/bin:${PATH}"
 RUN pip install --no-cache-dir --upgrade pip \
  && pip install --no-cache-dir "aiperf==${AIPERF_VERSION}"
 
-# ---- Final stage: runtime only, no toolchain ----
-FROM python:3.13-slim
+# ---- Final stage: runtime only, no toolchain, no shell ----
+# renovate: pinned by digest. The tag is retained for readability; the digest
+# is what actually resolves, so a retag upstream cannot silently change the
+# runtime. Bump both together.
+FROM nvcr.io/nvidia/distroless/python:3.13-v4.0.8@sha256:89574a7b9524b7166ff3e0c547ad87a203437b3afe71dfb9878e279196254b76
```
njhensley, line 89: **🔵 Nitpick — Builder FROM is a floating tag while this runtime FROM is digest-pinned**

This runtime stage is digest-pinned (`…distroless/python:3.13-v4.0.8@sha256:8957…`), but the builder stage (line 55, `FROM python:3.13-slim`) is a floating tag. Two builds of the same commit can compile the venv (pyzmq/uvloop/crick native extensions) against a different CPython 3.13 patch.

**Blast radius:** Functionally sound — only the ABI-compatible venv is copied forward and CPython guarantees ABI stability within 3.13.x — but it softens the "same inputs → same outputs" principle for the image bytes.

**Fix:** Optionally digest-pin the builder `FROM` too (renovate supports `# renovate-digest:`).

validators/performance/aiperf_entrypoint.py
mchmarny: Fixed, both halves of your suggestion:

- The builder stage now `COPY`s the wrapper and runs `python -m py_compile /opt/aicr/aiperf_entrypoint.py`; the runtime stage takes it via `COPY --from=builder`, so the copy that ships is the one that compiled.
- `validators/performance/aiperf_entrypoint_test.go` executes the real file against a stub `aiperf` on a throwaway PATH. `TestAIPerfEntrypointExitContract` locks 7 verbatim / 137 SIGKILL / 127 exec-not-found / 1 missing result file / 2 unset env, and asserts a failed run emits no sentinel. `TestAIPerfEntrypointFramingFeedsParser` covers argv passthrough, inherited chatter, the exact framing bytes, and the parser round-trip.

`python3` is `t.Skip`ped locally but `t.Fatal` under `CI` — a silent skip in CI would be your failure mode wearing a green badge. Mirrors `requireHelm` in `pkg/bundler/deployer/argocdhelm`.

Mutation-checked: removing the newline guard, the 128+N mapping, the `py_compile` line, or the runtime `COPY` destination each fails the suite.

.openvex.json
mchmarny: Fixed — correct on both counts. `.settings.yaml` still pins `chainsaw: v0.2.15`, and each of the 11 IDs resolves 1:1 to a retained GHSA statement (verified: none survives as a statement name, each appears in exactly one retained GHSA description).

The `tooling` field now records the GHSA-alias rationale, states chainsaw is unchanged at v0.2.15 with the vulnerable ssh code still shipped, and says explicitly that the retained GHSA statements must not be removed — which was the deletion hazard you flagged. The commit message carried the same false claim and has been corrected too.

validators/performance/inference_perf_test.go
mchmarny: Fixed. `TestAIPerfEntrypointPathsMatchDockerfile` reads `aiperf-bench.Dockerfile` and asserts against the Go constants: the venv root derived from `aiperfEntrypointPython` is both created and copied forward, a `COPY` lands `validators/performance/aiperf_entrypoint.py` at `aiperfEntrypointScript`, the runtime `COPY --from=builder` lands at that same path, and the `py_compile` gate names it.

Mutating the destination to `/opt/aicr/entrypoint.py` fails the test. The constant-vs-constant `Command` assertion stays as a cheap shape check.

validators/performance/inference_perf_test.go
mchmarny: Fixed — the claim was wrong. `parseAIPerfOutput` uses `strings.Index` + `TrimSpace`, so `…95.3}===AIPERF-RESULT===` parses fine.

The comment now says the guard is for log readability, not parseability, and I added a `closing sentinel glued to the JSON's final brace` case so the claim is proven by a test rather than asserted in prose.

validators/performance/inference_perf_test.go
mchmarny: Fixed. `TestAIPerfEntrypointFramingFeedsParser` runs the real wrapper against a stub that writes the export without a trailing newline, then feeds the wrapper's actual stdout to `parseAIPerfOutput`. It also asserts the stub's export genuinely lacks a trailing newline, so the case can't silently stop exercising the guard.

The hand-written case in `TestParseAIPerfOutput` stays as parser coverage for environments without python3.

validators/performance/inference_perf_constraint.go
mchmarny: Added. `TestBuildAIPerfJob_PrebuiltImageAndSentinel` now reads `--output-artifact-dir` with `argvFlagValue` and asserts it equals the `AICR_AIPERF_ARTIFACT_DIR` env value.

Mutating the argv side to `aiperfArtifactDir + "/run"` fails with exactly the diagnosis you predicted.

.openvex.json
mchmarny: Addressed via the corrected `tooling` note rather than by re-adding the GO-namespace statements. The note now records that the same CVEs are suppressed under GHSA and that the vulnerable code still ships, which removes the delete-the-twins hazard; and as you note the scan is `fail-build: false`, so a DB namespace flip costs noise on a daily notification, not a blocked merge.

Say the word if you'd rather pay ~190 lines of JSON for the belt-and-suspenders version.

validators/performance/inference_perf_constraint.go
mchmarny: Leaving as-is. `AICR_MODEL` isn't unique to this Job — `model_cache.go:347` sets the same name on the model-cache Job, where it *is* read (`snapshot_download(os.environ['AICR_MODEL'])`). Renaming only the aiperf occurrence would split a name that means the same thing across both pods an operator inspects.

The distinction you're guarding is already carried by the `AICR_INFERENCE_PERF_*` prefix on the live knobs, and the comment documenting this one as informational stays.

validators/performance/aiperf-bench.Dockerfile
mchmarny: Leaving as-is for repo consistency: all five Dockerfiles here pair a floating builder tag with a digest-pinned runtime — `cmd/gate`, `validators/deployment`, `validators/performance`, and `validators/conformance` all use `golang:${GO_VERSION}-bookworm` alongside a digest-pinned `distroless/static`.

Pinning only this builder makes it the odd one out and adds a digest to maintain for a stage whose bytes never ship. Agree on the principle — it should be one change across all five, which I'm happy to file as a follow-up.

mchmarny:
All 9 review items addressed in `86fc21f6` (force-push; `make qualify` exit 0 on the result).

| Item | Resolution |
|---|---|
| 🟡 Python wrapper had no test or compile gate | `py_compile` gate in the builder stage + `aiperf_entrypoint_test.go`, which executes the real wrapper against a stub `aiperf` (exit contract: 7 / 137 / 127 / 1 / 2) and feeds its actual stdout to `parseAIPerfOutput` |
| 🟡 VEX note blamed a chainsaw bump that never happened | Corrected — chainsaw is unchanged at `v0.2.15`; the 11 dropped statements are GO-ID duplicates of retained GHSA ones (verified 1:1). The commit message carried the same claim and was corrected too |
| 🟡 Dockerfile↔Go coupling test was tautological | `TestAIPerfEntrypointPathsMatchDockerfile` reads the Dockerfile and asserts the `COPY` destination and venv layout against the Go constants |
| 🔵 Newline-guard comment claimed a parse failure | Claim was wrong (`strings.Index` + `TrimSpace`); comment corrected and a glued-sentinel case added to prove it |
| 🔵 Parse-shape test hand-encoded the wrapper output | Now derived from a real `aiperf_entrypoint.py` run |
| 🔵 Artifact dir double-sourced | Assertion that the `--output-artifact-dir` argv equals the `AICR_AIPERF_ARTIFACT_DIR` env |
| 🔵 GO-namespace fragility | Addressed via the corrected tooling note; the GO-ID duplicates stay dropped (details in-thread) |
| 🔵 Rename `AICR_MODEL` | Declined — the name is shared with the model-cache Job, where it *is* read (`model_cache.go:347`) |
| 🔵 Digest-pin the builder `FROM` | Declined here — all five Dockerfiles pair a floating builder with a digest-pinned runtime; worth doing as one sweep |

Every new assertion was mutation-checked: removing the newline guard, the 128+N signal mapping, the `py_compile` line, the runtime `COPY` destination, or the argv/env artifact-dir agreement each fails the suite.

Thanks for the multi-persona pass — the VEX audit-trail catch in particular was the one that would have quietly cost us the retained GHSA suppressions later.

validators/performance/aiperf_entrypoint_test.go
```diff
@@ -0,0 +1,416 @@
...
+				compilesSource = true
+			}
+		case "COPY":
+			src, dst := fields[len(fields)-2], fields[len(fields)-1]
+			switch {
+			case src == venvRoot && dst == venvRoot:
+				copiesVenv = true
+			case src == "validators/performance/"+aiperfEntrypointSource && dst == aiperfEntrypointScript:
+				copiesSource = true
+			case src == aiperfEntrypointScript && dst == aiperfEntrypointScript:
```
njhensley, line 390: **🔵 Nitpick — Dockerfile-parser test is stage-blind; runtime COPYs aren't required to carry --from=builder**

`TestAIPerfEntrypointPathsMatchDockerfile` flattens all Dockerfile lines ignoring `FROM` stage boundaries, and the `copiesVenv`/`copiesToRuntime` matchers check only `src == dst`, not the `--from=builder` flag. A refactor that dropped `--from=builder` (copying from the build context instead of the builder stage) or misplaced a COPY across stages would still satisfy the guard.

**Blast radius:** Narrow — the common breakages this test targets (renamed dest, deleted COPY, removed `py_compile`, moved venv path) are all still caught, and dropping `--from=builder` would usually break `docker build` outright, so this is a low-exploitability precision gap rather than a real hole.

**Fix:** Optional: assert the two runtime COPY lines contain `--from=builder`, and/or track the current `FROM` stage while scanning. Follow-up nit, not a merge blocker.

njhensley (approved):
## 🔁 Re-Review — ✅ Approve

**Delta reviewed:** `1612e034` → `86fc21f6` (fix commit). Disposition of all 9 prior findings against the resolved code + an adversarial pass on the fix surface. Delta is **test + docs only** (4 files). Validated at the fix head: `go test -race` **PASS**, `golangci-lint` **0 issues**, `py_compile` **OK**, `.openvex.json` valid.

Thanks for the thorough turnaround — the fixes aren't cosmetic: the new +416-line `aiperf_entrypoint_test.go` executes the real Python wrapper end-to-end with a stub `aiperf`, and the Dockerfile now byte-compiles the wrapper in the builder and ships the exact file it accepted.

### Prior-feedback status

| Finding | Tier | Disposition |
|---|:---:|:---|
| F1 — Python wrapper untested / no compile gate | 🟡 | ✔️ Addressed — `py_compile` gate in builder + test asserts exit 7/137/127/1/2 on the real wrapper |
| F2 — openvex note "chainsaw bump" factually wrong | 🟡 | ✔️ Addressed — reframed as GHSA-alias dedup; states chainsaw unchanged at v0.2.15 |
| F5 — Dockerfile↔Go path test tautological | 🟡 | ✔️ Addressed — `TestAIPerfEntrypointPathsMatchDockerfile` reads the real Dockerfile (residual precision gap noted inline as B1) |
| F3 — comment claims newline guard needed to parse | 🔵 | ✔️ Addressed — comment corrected + new glued-sentinel parse case |
| F4 — parse test hand-encodes wrapper output | 🔵 | ✔️ Addressed — `TestAIPerfEntrypointFramingFeedsParser` feeds real wrapper stdout to `parseAIPerfOutput` |
| F6 — artifact dir double-sourced, no guard | 🔵 | ✔️ Addressed — argv `--output-artifact-dir` == env assertion added |
| F7 — "confirmed absent from scan" un-reproducible | 🔵 | ◐ Mitigated — reworded note is now concrete (GHSA-alias mapping); inherent VEX-repro difficulty remains |
| F8 — `AICR_MODEL` dead-but-set naming | 🔵 | ✖️ Not addressed — out of delta (optional nit) |
| F9 — builder floating tag vs digest-pinned runtime | 🔵 | ✖️ Not addressed — optional nit |

### Adversarially cleared (no new issue)
- **CI gate correct** — `requirePython3` fatals when `CI` is set; GitHub Actions always exports `CI=true` and `ubuntu-latest` ships python3, so the Python contract tests genuinely execute in CI and never silent-skip.
- No `__pycache__`/.pyc shipped — runtime `COPY --from=builder` names the exact `.py`; shipped bytes byte-identical.
- SIGKILL→137 deterministic; no goroutine/race/timeout fragility; no vacuous assertions.

**One new low-severity follow-up nit (B1, inline):** the Dockerfile-parser test is stage-blind. Not a blocker.

### Verdict

| 🔴 Blocker | 🟠 Major | 🟡 Minor (open) | 🔵 Nitpick (new) | Recommendation |
| 0 | 0 | 0 | 1 (B1) | **Approve** |

<sub>Re-reviewed with a multi-persona + adversarial meta-reviewer workflow, dispositioned against the fix commit at `86fc21f6`.</sub>

PR #2000, opened 2026-08-03
validate: fail-closed GKE device-plugin ownership check

### LINKED ISSUE #1755: validate: fail-closed check for GKE device-plugin ownership conflict

yuanchen8911:
## Summary

Adds the fail-closed GKE device-plugin ownership check from #1755: a new node-set constraint form (`NodeTopology.gpu-nodes.label`) evaluated at `aicr validate` readiness, wired into the GKE-COS overlays so an unlabeled GPU node pool fails validation closed (exit 2) with an actionable diagnostic before any phase runs.

## Motivation / Context

AICR's GKE recipes ship the GPU Operator with `devicePlugin.enabled: true`, while a default-provisioned GKE GPU node pool also runs GKE's managed device plugin — two advertisers of `nvidia.com/gpu` per node, with nondeterministic ownership. Until now nothing detected the conflict: `aicr bundle` is offline, and the generic constraint syntax compares scalar readings, so "every GPU node carries `gke-no-default-nvidia-gpu-device-plugin=true`" was inexpressible. PR #1998 documented the prerequisite; this PR enforces it.

**Scope position (ADR-015 Deferred Decision 2 — marked resolved in the ADR by this PR).** The reusable node-set constraint form — including the negated direction — lands here under #1755, exactly as the ADR reads it. The follow-up GKE `gpuStack` profile recipes (#1761 adoption step) will consume the form unchanged when they land: `operator` uses the positive predicate, `csp-managed` the negated one. Nothing in this PR touches the profile mechanism, the `Constraint` schema, the OpenAPI `propertyNames` allowlists, or the v1alpha2/v1alpha3 version gate.

**Merge order / dependency:** resolved — #1998 merged first (2026-08-04) and this PR is rebased onto that main. The component-catalog paragraph now describes the readiness gate (`NodeTopology.gpu-nodes.label`, fail-closed, exit 2) instead of claiming no deterministic check exists, closing the loop the readiness remediation text points into.

**Review-round fixes folded in at head:** deprecated apimachinery label validators replaced with `content.IsLabelKey`/`content.IsLabelValue` (the vendored v0.36.3 marks the old aliases deprecated); a `ValidatePhases` readiness-preflight test pinning the default client path fail-closed (the plural entry point previously had zero coverage on that branch); an encode→decode round-trip test running real `topology.Collector` output (plain, disambiguated, truncated shapes) through the evaluator so the cross-package wire format cannot silently diverge before #2003; parse-validation table rows for a valid key with an invalid label value; `key=` (empty value) accepted as a valid positive form — Kubernetes permits empty label values, the collector already encodes them (`|<nodes>`), and the disambiguated map key `<key>.` cannot collide with a real label key (keys may not end in a dot) — pinned by table rows for the uniform and mixed shapes plus a collector round-trip leg; and a comment documenting checkReadiness's deliberate error-code flattening.

Fixes: #1755
Related: #1761, #1998 (merge first — see above)

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Collectors / snapshotter (`pkg/collector`, `pkg/snapshotter`) — exported truncation predicate only
- [x] Other: `recipes/overlays/gke-cos.yaml`, `tests/uat/gcp/cluster-config.yaml`, `pkg/constraints`

## Implementation Notes

**The form is name-dispatched, not schema-extended.** `constraints.Evaluate` dispatches on the exact name `NodeTopology.gpu-nodes.label` before the scalar path (precedent: the `nccl-benchmark-*` sentinel names). The constraint stays `name` + `value`, so both existing call sites (readiness, overlay/profile evaluation) get the form with no API or artifact-version change.

**Value grammar** is `<label-key>=<value>` (every GPU node carries it) or `!<label-key>` (no GPU node carries the key), validated with the vendored Kubernetes label validators (`content.IsLabelKey`/`content.IsLabelValue`) — so the scalar operator grammar (`>= x`, `!= x`) cannot be misread as a key/value predicate, and a key no node can legally carry (e.g. a double slash) is rejected instead of making the negated predicate pass vacuously.

**GPU-node universe** = nodes under the snapshot's `cloud.google.com/gke-accelerator*` label readings — GKE's native signal, present from pool creation before the Operator or NFD run (NFD's `nvidia.com/gpu.*` labels don't exist on the pre-deployment cluster this check validates). Per the issue's acceptance requirements, both predicate directions fail closed on: truncated node lists (`--max-nodes-per-entry` snapshots), an empty universe (no vacuous pass), missing `NodeTopology.label` readings, and mixed/disambiguated label values (`key.value` encoding). Disambiguated-shape decoding enforces `encodeLabels`' invariants — plain and disambiguated forms never coexist, and genuine disambiguation always yields at least two entries — so a distinct dotted label whose value equals its own suffix (e.g. a literal `<key>.true=true` label) cannot satisfy the predicate; an accepted disambiguated set must additionally partition its nodes — overlapping node sets prove the encoder's `<key>.<value>` map key collided with a real label of that literal name (one reading silently overwritten, #2003) and fail closed as an ambiguous reading. A single disambiguated-shape entry without the plain key (a possible collision remnant) and structurally malformed readings (missing or extra `|` separators, empty node lists, node tokens that are not canonical RFC-1123 node names) also fail closed — both would otherwise let the negated predicate pass vacuously. The residual ambiguity (collisions with identical node sets, or every disambiguated entry overwritten) is what the lossy encoding cannot express; the lossless-encoding root-cause fix is tracked in #2003, structured truncation metadata in #2002. The truncation detector is owned by the collector (`topology.IsTruncatedNodeList`, next to `formatNodeList`, with a round-trip test) so the format and its detector cannot drift apart.

**Both validator entry points are gated.** `checkReadiness` runs in `ValidatePhases` and in the per-phase `ValidatePhase` (the exported SDK path `Client.ValidateState` documentation directs per-phase callers to) — a single-phase caller cannot bypass the recipe's readiness constraints, and declared readiness constraints with a nil snapshot fail closed instead of silently skipping. Both pinned by tests.

**Wiring: readiness, not `spec.constraints`.** A top-level constraint would exclude the GKE overlays during snapshot-based generation on exactly the unlabeled cluster the diagnostic exists to fix. The check lives in the GKE-COS base overlay's `validation.readiness.constraints`; `checkReadiness` now evaluates readiness-phase constraints alongside the top-level set (they were declared, merged, and carried into `ValidationInput` but never consumed) and appends the constraint's `remediation` to the failure message — that's where the issue's required device-plugin diagnostic lives.

**UAT cluster config.** The GCP UAT GPU pool (`tests/uat/gcp/cluster-config.yaml`) predates the documented prerequisite and does not carry the label; without fixing it this PR would fail the GKE UAT lane. The label is added to the pool's `nodeConfig.labels`. Note the config actuates at cluster provision time — the nightly lifecycle (provision→CUJ→teardown) picks it up on the next run; a held daytime cluster would need re-provisioning or a manual `gcloud container node-pools update`.

**Sequencing with #1998.** #1998 merges first; this PR rebases on it and refreshes the component-catalog paragraph (added there) that states "AICR has no deterministic check for a violation today" — kept out of this PR until the rebase to avoid cross-PR conflicts.

## Testing

```bash
make qualify   # PASS (exit 0) through the collision-guard revision; on the final head the local run hit
               # pre-existing tests/releasepolicy deadline timeouts (reproduced identically on clean
               # origin/main — machine-local flake); affected packages re-verified with -race and CI is green
GOFLAGS="-mod=vendor" go test -race ./pkg/constraints/... ./pkg/validator/ ./pkg/recipe/...   # PASS
golangci-lint run -c .golangci.yaml ./pkg/constraints/... ./pkg/validator/...   # 0 issues
```

Coverage (per-package, current vs origin/main baseline):
- `pkg/constraints`: 97.9% → 97.9% (flat)
- `pkg/validator`: 48.5% → 50.9% (+2.4%)
- `pkg/collector/topology`: 84.2% → 84.4% (+0.2%)

CLI end-to-end (branch-built binary, no cluster): generated the GKE H100 training recipe with `aicr recipe` — the hydrated artifact carries the readiness constraint with its remediation — then ran `aicr validate --no-cluster` against four crafted snapshots: labeled pool → exit 0 ("all 2 GPU node(s) carry …"); unlabeled pool → exit 2 naming the offending nodes plus the full device-plugin remediation; truncated snapshot → exit 2 with the regenerate-without-`--max-nodes-per-entry` instruction; no GPU nodes → exit 2 with the empty-universe diagnostic. Live-cluster validation deliberately skipped: the gate is inline (no cluster I/O), and the first post-merge nightly UAT provisions from this PR's cluster-config and runs `aicr validate` through the gate, providing the live confirmation automatically.

New tests: table-driven coverage of both predicate directions × {pass, fail, mixed values via disambiguated keys, multiple accelerator types, non-GPU nodes carrying the label, truncated universe reading, truncated target reading (both directions), empty universe (both directions), missing label subtype/measurement/nil snapshot, prefix-collision non-misattribution, scalar-grammar rejection, malformed values}; validator-level test that readiness-phase constraints are evaluated and carry remediation into the failure.

New exported symbols: the `GPUNodesLabelConstraintName` const and `topology.IsTruncatedNodeList` (covered by the round-trip test). `make qualify` was run unsandboxed after verifying the sandbox-only failures (`cleanup_test.sh` stubs, `httptest` port binds) reproduce identically on a clean `origin/main` checkout.

## Risk Assessment

**Rollout notes:** A pre-change `aicr` binary evaluates only top-level constraints and silently ignores `validation.readiness.constraints` in a recipe generated at this commit (version skew is advisory-only) — validate recipes with a matching or newer binary.  `aicr validate` against a GKE cluster whose GPU pools lack `gke-no-default-nvidia-gpu-device-plugin=true` now fails readiness closed (exit 2) — intended and documented, but visible to existing users of unlabeled clusters; the failure message carries the exact remediation. Snapshots captured with `--max-nodes-per-entry` also fail this constraint closed with a regenerate instruction. Recipe generation and bundling are unaffected. The UAT GKE pool config is updated in this PR; held daytime UAT clusters need one re-provision (or a manual pool update) before their next validate run.

## Checklist


pkg/validator/validator.go
```diff
@@ -204,8 +225,9 @@ func (v *Validator) ValidatePhases(
 		return nil, err
 	}
 
-	// Pre-flight: evaluate top-level validation constraints against snapshot.
-	// Fails fast before deploying any Jobs if prerequisites aren't met.
+	// Pre-flight: evaluate the top-level and readiness-phase constraints
+	// against the snapshot. Fails fast before deploying any Jobs if
+	// prerequisites aren't met.
 	if err := checkReadiness(validationInput, snap); err != nil {
```
njhensley, line 236: **🟠 Major — `ValidatePhases` readiness-failure branch is untested — the default `aicr validate` path can regress to fail-open**

Coverage confirms this `return nil, err` (when `checkReadiness` fails inside `ValidatePhases`) has hit count 0. Only the singular `ValidatePhase` path is tested (`TestValidatePhaseRunsReadinessPreflight`); both `ValidatePhases` tests pass a `nil` snapshot so the gate returns early. The **default** client validate path routes through `ValidatePhases` (`pkg/client/v1/aicr.go`). A regression that reorders `checkReadiness` below the `NoCluster` short-circuit here would pass every existing test while silently fail-opening the primary path — the exact #1755 failure this PR exists to prevent. The code is correct today; the gap is that nothing pins it.

**Blast radius:** Default `aicr validate` (full-phase) could regress to fail-open on the readiness gate with no test catching it; the singular path is guarded but the primary plural path is not.

**Fix:** Add a `ValidatePhases` analog of `TestValidatePhaseRunsReadinessPreflight` — no-cluster Validator + a recipe with a failing readiness constraint + non-nil snapshot — asserting `ValidatePhases` returns `ErrCodeInvalidRequest` and no `PhaseResults`.

njhensley (commented):
## 🔎 Multi-Persona Review — fail-closed GKE device-plugin ownership check

**Method:** 4 independent persona reviewers (Correctness · Security/fail-closed · Domain & Architecture · Test-coverage) fanned out against the diff, then every finding was re-derived from the resolved code at `1aede4d5` by an adversarial senior meta-reviewer (confirm / refute / re-tier). Personas post nothing directly; only surviving findings appear below.

### Overall assessment

This is an unusually careful, defense-in-depth PR, and the two highest-stakes lenses came back clean:

- **Correctness** traced `encodeLabels` ↔ `decodeLabelEntries` and confirmed the disambiguation/collision invariants hold — every ambiguous / truncated / empty / malformed shape returns an error rather than vacuously satisfying a predicate. The `cs[:len(cs):len(cs)]` three-index slice correctly avoids mutating the caller's constraints; no double-evaluation (Readiness is excluded from `PhaseOrder`); regex, nil-safety and `go vet` clean.
- **Security / fail-closed** confirmed both entry points (`ValidatePhases` + `ValidatePhase`) gate *before* the `--no-cluster` short-circuit, nil snapshot fails closed, the truncation regex is not defeatable by crafted node names / label values, and value-grammar validation closes the vacuous-negated-pass hole in both directions.

No blockers. The single 🟠 is a **test-coverage gap, not a code defect** — the code is correct today but nothing pins the default `aicr validate` path's fail-closed behavior. The 🟡s are a merge-order doc dependency, an architecture-coupling smell, and a missing round-trip guard.

**Recommendation: ✅ Approve with comments.**

### Confirmed non-issues (examined and refuted)

- No fail-open path in either predicate direction — empty universe, truncated universe/target, missing readings, malformed encodings, non-canonical node tokens, single-dotted ambiguous shape, and collision-remnant all return errors.
- Truncation regex `\(\+\d+ more\)$` is not forgeable — RFC-1123 node names and label values cannot contain the pattern or `|`.
- No constraint-slice aliasing — `cs[:len(cs):len(cs)]` is the correct full three-index expression.
- No double-evaluation of readiness constraints — Readiness is excluded from `PhaseOrder`; phase execution consumes different constraint sets.
- Both entry points gated, `--no-cluster` still evaluates the gate, nil snapshot fails closed, no SDK bypass via `pkg/client/v1`.

### Summary

| Tier | Count |
|------|-------|
| 🔴 Blocker | 0 |
| 🟠 Major | 1 |
| 🟡 Minor | 3 |
| 🔵 Nitpick | 2 |

Inline comments follow.

<sub>Reviewed with a multi-persona + adversarial meta-review workflow. Findings were re-derived from the resolved code before posting; the reviewed commit's code was not executed beyond `go test`/`go vet`.</sub>

pkg/constraints/gpu_nodes.go
```diff
@@ -0,0 +1,376 @@
...
+package constraints
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
+
+	"github.com/NVIDIA/aicr/pkg/collector/topology"
```
njhensley, line 25: **🟡 Minor — I/O-free constraints eval package now transitively imports client-go for a 3-line regexp helper**

`pkg/constraints` imports `pkg/collector/topology` solely for the pure `IsTruncatedNodeList` regexp, but `topology.go` imports `k8s.io/client-go/kubernetes` + `pkg/k8s/client`, so this foundational shared eval package (used by recipe/validator/client) now transitively drags in the k8s client graph. Verified this is coupling, not a defect: `topology` does not import `constraints` (no import cycle), and `doc.go`'s runtime claim ("never performs network or cluster I/O") still holds — the helper is pure. Compiles, tests pass.

**Fix:** Optional/directional: hoist the truncation marker + `IsTruncatedNodeList` (ideally the whole `value|nodes` codec) into a client-go-free leaf that both the collector and `pkg/constraints` depend on. Reasonable but heavy for one helper — fine to fold into the #2003 lossless-encoding work.

pkg/constraints/gpu_nodes.go
```diff
@@ -0,0 +1,376 @@
...
+// membership list can falsely satisfy both predicate directions — #1755
+// acceptance requirement 1), and a structurally malformed reading is
+// rejected rather than decoded: no "|" separator, more than one "|", an
+// empty node list, or a node token that is not a canonical Kubernetes node
+// name (RFC 1123 subdomain — rejects whitespace, empties, and embedded
+// separators). A malformed token would otherwise never equal a real
+// universe member, letting the negated predicate pass vacuously. Truncated
+// readings skip token validation — their "(+N more)" tail is not a node
+// name by design — and keep their distinct truncation diagnostic.
+func decodeLabelEntries(data map[string]measurement.Reading, key string) ([]labelNodeSet, error) {
```
njhensley, line 217: **🟡 Minor — Cross-package wire-format contract has no encode→decode round-trip test**

`decodeLabelEntries`/`cutLabelEncoding`/`splitNodes` independently re-implement `topology.encodeLabels`' wire format — the `|` separator, comma-joined node list, and the `<key>.<value>` disambiguation rule. Only `IsTruncatedNodeList` has a shared round-trip guard; the decoder tests hand-build encoded strings rather than feeding real `encodeLabels()` output through `decodeLabelEntries`. A future collector-side format change (separator swap, different disambiguation trigger) could silently break decode — or make it fail-closed on healthy snapshots — with zero failing tests.

**Fix:** Add a round-trip test that runs `encodeLabels` output (plain, disambiguated, truncated shapes) through `decodeLabelEntries`, so the format and its parser cannot diverge. The #2003 lossless-encoding fix is the durable resolution.

recipes/overlays/gke-cos.yaml
```diff
@@ -78,6 +78,28 @@ spec:
...
+          value: gke-no-default-nvidia-gpu-device-plugin=true
+          remediation: >-
+            GKE's managed device plugin must be disabled on AICR-managed GPU
+            node pools via the node label
+            gke-no-default-nvidia-gpu-device-plugin=true; otherwise it
+            conflicts with the GPU Operator's device plugin
+            (devicePlugin.enabled: true). Add the label to the GPU node pool
+            (gcloud container node-pools update ... --node-labels=...);
+            note --node-labels REPLACES the pool's full label set, so pass
+            every existing label plus the new one (see the Component
```
njhensley, line 98: **🟡 Minor — Failure remediation cites a Component Catalog section that does not exist on this tree**

The readiness remediation tells operators to "see the Component Catalog's GKE Device-Plugin Ownership section for the discovery-then-update procedure," but `docs/user/component-catalog.md` has no such section, and the `gcloud --node-labels` discovery-then-update procedure appears nowhere in `docs/`. A user who hits exit 2 today follows a dead pointer. The PR body notes this is gated on #1998 (merge-first) adding the paragraph, refreshed here on rebase — which mitigates but leaves a fragile merge-order dependency.

**Fix:** Ensure #1998 lands first (or add the section here) and re-verify the reference resolves before merge. If timing slips, point the remediation at `docs/user/validation.md`, which does cover the behavior.

pkg/validator/validator.go
```diff
@@ -47,16 +47,34 @@ import (
...
+	}
 
-	slog.Info("readiness pre-flight", "constraints", len(validationInput.Constraints))
+	slog.Info("readiness pre-flight", "constraints", len(cs))
 
-	for _, c := range validationInput.Constraints {
+	for _, c := range cs {
 		result := constraints.Evaluate(c, snap)
 		if result.Error != nil {
 			return errors.WrapWithContext(errors.ErrCodeInvalidRequest,
```
njhensley, line 85: **🔵 Nitpick — checkReadiness flattens the evaluator's deliberate ErrCodeNotFound to ErrCodeInvalidRequest**

This unconditional re-wrap collapses the `ErrCodeNotFound` that `gpu_nodes.go` returns for empty-universe / missing-readings, unlike `PropagateOrWrap` used elsewhere in the file. Defensible on the readiness path — the documented contract is a uniform fail-closed exit 2, and "constraint can't be evaluated" reads as invalid-request — so no change is required. Flagging only for awareness; a one-line comment noting the deliberate flattening would preempt the next reader's double-take.

pkg/constraints/gpu_nodes.go
```diff
@@ -0,0 +1,376 @@
...
+		}
+		return key, "", true, nil
+	}
+	key, want, ok := strings.Cut(v, "=")
+	if !ok || want == "" {
+		return "", "", false, errors.NewWithContext(errors.ErrCodeInvalidRequest,
+			"invalid node-set constraint value: expected \"<label-key>=<value>\" or \"!<label-key>\"",
+			map[string]any{ctxValue: raw})
+	}
+	if errs := append(validation.IsQualifiedName(key), validation.IsValidLabelValue(want)...); len(errs) > 0 {
```
njhensley: **🔵 Nitpick — IsValidLabelValue(want) is never the sole rejecting predicate in tests**

The positive-form validation `append(IsQualifiedName(key), IsValidLabelValue(want)…)` is only exercised with an invalid key (where `IsQualifiedName` already fails). No case supplies a valid key with an invalid label value (e.g. `foo=bad value` with a space), so `IsValidLabelValue(want)`'s independent contribution is unpinned. Working guard, no current defect.

**Fix:** Add a table row with a valid key and an invalid label value expecting `ErrCodeInvalidRequest`. (Companion nit: no case supplies `key=` with an empty value — the load-bearing `want == ""` guard at line 138 is similarly unpinned, though its misparse direction fails closed.)

recipes/overlays/gke-cos.yaml
mchmarny, line 98: **Blocking — refresh the referenced catalog after #1998 merged.** The section now exists, but it still says AICR has “no deterministic check for a violation today” ([current `main`](https://github.com/NVIDIA/aicr/blob/f82256053827648045cab93597c0c74e007bf220/docs/user/component-catalog.md#L171)). Merging this head would ship the readiness gate while its remediation sends operators to documentation that denies the gate exists. Rebase onto current `main` and update that paragraph to describe the `NodeTopology.gpu-nodes.label` readiness behavior, as the PR body promises.

mchmarny (changes requested):
REQUEST_CHANGES — 1 documentation-truth blocker on exact head `1aede4d5`. Focused race tests, affected-package lint, YAML/diff checks, and CLI recipe hydration passed; live CI has zero failures. Other unresolved comments are non-blocking.

recipes/overlays/gke-cos.yaml
yuanchen8911: Fixed in 02a3e805b. Rebased onto current main and replaced the stale paragraph: it now states aicr validate enforces the prerequisite deterministically at readiness (NodeTopology.gpu-nodes.label, fail-closed on missing/mixed labels, an empty GPU-node set, and readings --max-nodes-per-entry actually truncated — a non-truncating cap validates normally — exit 2 before any check Jobs deploy), links docs/user/validation.md for the gate mechanics, and keeps the explanation of why bundle/operator-health/check-nvidia-smi don't catch the conflict as secondary context. The remediation pointer now resolves to a section that describes the gate it belongs to. Full make qualify passes locally at this head.

pkg/validator/validator.go
yuanchen8911: Fixed in 02a3e805b: added TestValidatePhasesRunsReadinessPreflight — no-cluster Validator, a failing top-level recipe constraint, non-nil snapshot — asserting ValidatePhases returns ErrCodeInvalidRequest and nil PhaseResults. Note the fixture uses a top-level constraint rather than validation.readiness.constraints; checkReadiness merges both into one gate, so the test pins the branch-ordering risk you identified (checkReadiness reordered below the NoCluster short-circuit) on the plural path, the same way the existing singular-path test does.

pkg/constraints/gpu_nodes.go
yuanchen8911: Agreed it's coupling, not a defect, and taking your suggested disposition: deferring the hoist to #2003, where the whole value|nodes codec moves anyway — extracting a leaf package for one pure helper now would be churned again by that rework. No change in this PR.

pkg/constraints/gpu_nodes.go
yuanchen8911: Fixed in 02a3e805b: TestGPUNodesLabelRoundTripsCollectorEncoding runs real topology.Collector output — via the fake clientset, not hand-built strings — through the full Evaluate path for the plain, disambiguated, truncated, and empty-value shapes (truncated asserted fail-closed in both predicate directions). A collector-side format change now fails this test instead of silently breaking the decoder. #2003 remains the durable resolution.

recipes/overlays/gke-cos.yaml
yuanchen8911: Resolved by events plus the rebase in 02a3e805b: #1998 merged first as planned, this branch is rebased onto that main, and the referenced GKE Device-Plugin Ownership section (including the discovery-then-update procedure) exists on this tree. Also refreshed the section's detection paragraph so it describes this gate rather than denying it exists.

pkg/validator/validator.go
yuanchen8911: Took the one-line-comment option in 02a3e805b: the re-wrap now carries a comment stating the flattening is deliberate — the readiness contract is one uniform fail-closed exit, and "constraint cannot be evaluated" is an invalid request at that boundary. Behavior unchanged.

pkg/constraints/gpu_nodes.go
yuanchen8911: Fixed in 02a3e805b: added a table row with a valid key and an invalid value (gke-no-default-nvidia-gpu-device-plugin=bad value) expecting ErrCodeInvalidRequest, so IsLabelValue is the sole rejecting predicate there. Your companion nit's key= case went the other way on verification: Kubernetes permits empty label values and the collector already encodes them, so key= is now accepted as a valid positive form, pinned by uniform and mixed table rows plus a collector round-trip leg. Note the validators changed in the same commit (deprecated aliases replaced by content.IsLabelKey/content.IsLabelValue); the new rows pin the replacements.

pkg/constraints/gpu_nodes.go
```diff
@@ -0,0 +1,304 @@
...
+// limitations under the License.
+
+package constraints
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/validation"
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 02a3e805b, with one precision on the suggestion: content.IsQualifiedName exists in the vendored v0.36.3 but is itself deprecated in favor of content.IsLabelKey (which the old validation.IsQualifiedName aliases directly), so the replacements used are content.IsLabelKey and content.IsLabelValue. validation.IsDNS1123Subdomain (node-name checks) is not deprecated and stays.

pkg/constraints/gpu_nodes.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+		return key, "", true, nil
+	}
+	key, want, ok := strings.Cut(v, "=")
+	if !ok || want == "" {
+		return "", "", false, errors.NewWithContext(errors.ErrCodeInvalidRequest,
+			"invalid node-set constraint value: expected \"<label-key>=<value>\" or \"!<label-key>\"",
+			map[string]any{ctxValue: raw})
+	}
+	if errs := append(validation.IsQualifiedName(key), validation.IsValidLabelValue(want)...); len(errs) > 0 {
```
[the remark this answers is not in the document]
yuanchen8911: Taking this after all — fixed in 02a3e805b. My earlier pushback's mechanism was wrong on verification: the collector already encodes empty label values ("|<nodes>", with a dedicated collector test), the decoder handles them, and the disambiguated map key "<key>." cannot collide with a real label key because qualified names may not end in a dot — so there is no encoding ambiguity to fail closed on. "key=" is now a valid positive form (every GPU node carries the key with an empty value), pinned by table rows for the uniform and mixed shapes plus a collector round-trip leg.

mchmarny (approved):

njhensley (approved):

PR #2004, opened 2026-08-04
feat(notices): add Python dependency licenses to third-party notices

lockwobr:
## Summary

`THIRD_PARTY_NOTICES.md` covered only Go modules, so the third-party Python packages redistributed inside the released `aiperf-bench` image carried no attribution. This adds them under the same freshness enforcement as the Go half, and fixes two pre-existing defects that were silently corrupting the released file.

## Motivation / Context

`ghcr.io/nvidia/aicr-validators/aiperf-bench` is one of the seven canonical released images (asserted in `attest-images.yaml`) and ships AIPerf plus its full 130-package transitive closure. That is redistribution, and MIT/BSD/Apache all attach notice obligations to it, but none of those packages appeared in the notices file.

The closure is fetched rather than vendored, so it cannot be enumerated offline the way `vendor/` can. Three MPL-2.0 packages are present (`certifi`, `orjson`, `tqdm`), which is worth an OSRB look.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Validator (`pkg/validator`) — the `aiperf-bench` image under `validators/performance/`
- [x] Other: notices tooling (`tools/`), merge gate

## Implementation Notes

**Pipeline.** `validators/performance/requirements.txt` is the single input, shared by the image build and the collector, so the attribution record and the artifact cannot describe different closures. `make python-licenses` resolves it with `pip download` and reads each wheel's `dist-info/METADATA`, rendering a committed Markdown fragment. `make notices` appends that fragment verbatim.

**Why a committed fragment.** `make notices-check` regenerates and git-diffs in the merge gate, so generation must stay hermetic and byte-deterministic. Isolating the network step into a separate, deliberate command preserves that. The fragment records the sha256 of `requirements.txt` and `make notices` fails closed when it no longer matches, so editing requirements without refreshing licenses cannot merge. No new CI job, no docker or network in the gate.

**No docker.** aiperf declares `requires-python <3.14`, so it cannot be installed by a venv on a 3.14 host. `pip download --python-version` fetches *for* a target interpreter while running on any interpreter. `pip-licenses` and `syft` were both evaluated against the same wheels: syft carries no source URL for Python packages and no license text; pip-licenses reports Trove classifier strings where `METADATA` holds clean SPDX, and via `PYTHONPATH` it either drops `setuptools` or picks up its own venv. Reading `METADATA` is exact by construction and adds no dependency.

**One platform, guarded.** The closure resolves on `linux/amd64` only. A wheel's architecture does not change its license, and the only way a package can differ between architectures is an environment marker, so `check_arch_markers` fails the run if any dependency is gated to a single machine architecture. Collecting both is also impossible with `--only-binary`: `crick` publishes no linux/arm64 wheel.

### Two pre-existing Go-side fixes

Both corrupted the released file in ways invisible in a diff:

1. **Non-license files were emitted.** `go-licenses save` copies whole-module source for restricted licenses, and every saved file was rendered. The notices carried **143 non-license files** (`README.md`, `Makefile`, `.gitignore`, `.travis.yml`, `*.go`) across 20 hashicorp packages. Now filtered to license-bearing filenames; `NOTICE` and dual-license variants still pass.
2. **License text escaped its code block.** A license containing a Markdown fence closed the wrapping ```` ```text ```` block early, inverting the rendering of everything after it. hashicorp READMEs and pillow's `LICENSE` both trigger this. Fences are now sized to exceed the longest backtick run in the content.

Net: 204 Go modules plus 130 Python packages, and the file is **21266 lines smaller** than before despite the added section.

Also portability-only: `mktemp` now passes an explicit template (macOS ignores `$TMPDIR` without one), and the hash check accepts `sha256sum` or `shasum`.

## Testing

```bash
make lint                          # pass
go test ./tests/notices/...        # 10 tests pass
golangci-lint run ./tests/notices/...  # 0 issues
make notices                       # byte-identical across two runs
```

New `tests/notices/` package drives both generators, following the `tests/releasepolicy` pattern for testing non-Go tooling. Fixtures are synthetic wheels built in-test with `archive/zip`, so no network. Each pins a real failure: undeclared `License-File`, `LICENCE` spelling, fence-containing license, prose in the `License` field, no license file, PEP 503 normalization.

**Mutation-verified rather than trusted green.** Reverting the dynamic-fence fix makes `FencesBalance` fail (`closing fence with no opener`); reverting the license-file fallback makes `RecoversUndeclaredText` fail (`fallback notice count = 5, want 1`). Two tests need no tooling at all and guard the committed artifact directly: the requirements/fragment hash match, and whole-file fence balance.

**`make qualify` was not fully green locally.** Three failures, all environmental, none touched by this change, each reproducing on clean `main` on the same machine: `test-shell` (sandboxed `mktemp`), `tests/releasepolicy` (`timeout` is GNU coreutils, absent on macOS), `pkg/oci` (local helm 4.2.0 vs pinned 4.2.3). Everything else passes. Relying on CI for those three.

## Risk Assessment

**Rollout notes:** No runtime behavior changes. The `aiperf-bench` image installs the same packages via `requirements.txt` instead of an inline pin, with a build-time guard so `AIPERF_VERSION` cannot disagree with what is installed. One judgment call worth review: filtering out the MPL source that `go-licenses` deliberately saves. A Markdown notices file is not how source availability is discharged and these modules are redistributed unmodified, but it is a deliberate narrowing of what ships.

## Checklist


validators/performance/requirements.txt
```diff
@@ -0,0 +1,11 @@
...
+#
+# This file is the single input shared by two consumers, so they cannot drift:
+#   - aiperf-bench.Dockerfile installs from it (`pip install -r`).
+#   - tools/generate-python-licenses collects the license of every package the
+#     resulting environment contains, for THIRD_PARTY_NOTICES.md.
+#
+# Bumping anything here requires re-running 'make python-licenses' and
+# 'make notices'; the notices generator fails closed if this file changes
+# without a corresponding license refresh.
+aiperf==0.11.0
```
njhensley, line 11: **🟠 Major — Freshness gate binds requirements.txt bytes, not the resolved closure**

requirements.txt pins only `aiperf==0.11.0`; the ~130-package transitive closure is unpinned. The image build (`pip install -r`, aiperf-bench.Dockerfile) and the collector (`pip download`, tools/generate-python-licenses) resolve this file independently, against mutable PyPI, at different times — so they can pick different transitive versions, or a different package set if a dep adds/drops a sub-dependency. The only staleness gate hashes `sha256(requirements.txt)` (generate-notices), and those bytes never change while `aiperf==0.11.0` holds. When any transitive dep publishes a new release the shipped image drifts from the committed attribution and CI stays green — the notices can under-report or misversion packages actually redistributed in the released image. The commit's 'single input … cannot describe different closures' conflates a shared input *file* with a reproducible *resolution*; it holds only for the one pinned line.

**Blast radius:** THIRD_PARTY_NOTICES.md (a per-release legal attribution artifact) can silently diverge from the shipped aiperf-bench image, with no CI signal. `make notices-check` structurally can't catch it — it regenerates from the always-fresh committed fragment.

**Fix:** Drive both the build and the collector from a fully-pinned + hashed lock (`pip-compile --generate-hashes` / `uv pip compile`), `pip install --require-hashes`, and hash the *lock* in the gate. That makes the recorded versions provably equal the shipped ones, makes the image reproducible/dependency-confusion-resistant, and collapses the F2 marker-guard concern. (The unpinned live-PyPI install is pre-existing, not a regression, but this PR is the natural place to close it since accuracy is the whole point.)

tools/generate-python-licenses
```diff
@@ -0,0 +1,425 @@
...
+_LICENSE_FILE = re.compile(
+    r"^(LICEN[CS]E|COPYING|COPYRIGHT|NOTICE|AUTHORS|PATENTS)([-._].*)?$",
+    re.IGNORECASE,
+)
+
+# Matches a marker pinning a specific machine architecture, e.g.
+# `platform_machine == "aarch64"`. A marker that merely *lists* architectures
+# inclusively (greenlet, hf-xet) resolves true on both targets and is fine; it
+# is the equality against a single value that would split the closure.
+_ARCH_MARKER = re.compile(
```
njhensley, line 113: **🟡 Minor — arch-marker guard only matches `platform_machine ==`**

`_ARCH_MARKER` matches only the equality form. A dep gated `platform_machine != 'x86_64'` (effectively arm64-only), `... in 'aarch64 arm64'`/`not in`, or reversed `'aarch64' == platform_machine` is both excluded from the amd64-only download AND not flagged by `check_arch_markers` (the `if not found & _ARCH_TARGETS: continue` path) — defeating the single-platform-collection soundness argument for exactly the case it exists to catch.

**Blast radius:** Latent hole in a defense-in-depth guard; no current dependency triggers it, so it's not an active mis-attribution today, but `!=`/`in` are common marker forms.

**Fix:** Flag any `platform_machine` marker via any operator/operand order, require a provable both-arch-inclusive result to pass, and fail closed on unrecognized marker syntax. Pinning + hashing the closure (F1) removes the need for this reasoning entirely.

tools/generate-notices
```diff
@@ -220,9 +307,23 @@ EOF
...
+        # files into this document, which is not what a notices file is for and
+        # is not how source availability is actually discharged (we redistribute
+        # these modules unmodified; the upstream repository is the source).
+        # The glob still admits NOTICE and the dual-license variants
+        # (LICENSE-MIT / LICENSE-APACHE) that the emit loop exists to catch.
         license_files=()
         if [[ -d "${LICENSES_DIR}/${pkg}" ]]; then
             while IFS= read -r -d '' f; do
+                if ! printf '%s' "$(basename "${f}")" \
+                    | grep -qiE '^(licen[cs]e|notice|copying|copyright|authors|patents)([-._].*)?$'; then
```
njhensley, line 324: **🟡 Minor — license-file filter drops UNLICENSE/MIT-LICENSE, admits license.go**

The start-anchored `^(licen[cs]e|notice|copying|copyright|authors|patents)([-._].*)?$` drops suffix-style names (`UNLICENSE`, `MIT-LICENSE`, `COPYING3`) while `license.go` slips through via `[-._].*`. The Go side has no declaration fallback, so a module whose only license file is `UNLICENSE` renders as 'License text unavailable' — the exact attribution-loss class this PR fixes — and a go-licenses-saved `license.go` could be emitted as license text. The same regex on the Python side (line 105) is shielded by the License-File declaration path.

**Blast radius:** Silent loss of verbatim license text for modules using suffix/Unlicense filenames; possible emission of source files as licenses. Not verified against the current closure.

**Fix:** Allow the keyword as a suffix (and add `unlicen[cs]e`); constrain the trailing part to doc extensions (`.txt|.md|.rst|.html|` empty) so `.go` is excluded. Apply symmetrically in both generators.

tests/notices/staleness_test.go
```diff
@@ -0,0 +1,226 @@
...
+}
+
+// TestNoticesRejectsStaleFragment drives the generator's fail-closed path: a
+// fragment whose recorded hash does not match the requirements must abort the
+// run rather than emit a notices file describing the wrong dependency set.
+func TestNoticesRejectsStaleFragment(t *testing.T) {
+	// The generator checks its own tool prerequisites before reaching the
+	// staleness gate, so skip rather than fail where those are absent. The
+	// notices-freshness merge-gate job has them.
+	if _, err := exec.LookPath("go-licenses"); err != nil {
```
njhensley, line 152: **🟡 Minor — Generator fail-closed tests skip in every CI job**

`TestNoticesRejectsStaleFragment` / `TestNoticesRejectsMissingFragment` `t.Skip` when go-licenses is absent. The `tests` job (qualification.yaml) never installs go-licenses, so `make test` hits the skip; the `notices-freshness` job runs `make notices-check`, not `go test`. So the tests that specifically cover generate-notices' fail-closed branches execute in no CI job.

**Blast radius:** A regression removing the sha256 staleness gate or the missing-fragment guard would pass all gates; the fail-closed code runs in prod but its regression tests are dead.

**Fix:** Install go-licenses in the test job, or run `go test ./tests/notices/...` inside notices-freshness.

tools/generate-notices
```diff
@@ -59,6 +74,69 @@ if ! command -v yq >/dev/null 2>&1; then
...
 
+# Fail closed on a missing/empty Python fragment rather than silently emitting
+# a Go-only notices file: quietly dropping the attributions for everything
+# redistributed inside the aiperf-bench image is the failure mode this gate
+# exists to prevent.
+if [[ ! -s "${PYTHON_NOTICES}" ]]; then
+    echo "ERROR: ${PYTHON_NOTICES} is missing or empty." >&2
+    echo "The aiperf-bench image redistributes third-party Python packages whose" >&2
+    echo "licenses must appear in ${OUTPUT}. Regenerate the committed fragment" >&2
+    echo "with 'make python-licenses' (requires docker + network) and commit it." >&2
```
njhensley, line 85: **🟡 Minor — Stale 'requires docker' guidance — tool no longer uses docker**

The tool was rewritten to use `pip download` (its own docstring: 'no container'), but generate-notices still says 'requires docker + network' / 'needs docker and PyPI' at lines 53/56/85/91/119 (two are user-facing error messages) and merge-gate.yaml:130 says 'docker + PyPI'.

**Blast radius:** A maintainer hitting the failure is pointed at the wrong remediation (installing docker won't help).

**Fix:** Replace 'docker + network' / 'docker + PyPI' with 'network access to PyPI' in generate-notices and merge-gate.yaml.

tools/generate-python-licenses
```diff
@@ -0,0 +1,425 @@
...
+        add(f"\n### {pkg['key']}\n")
+        add(f"* Version: {pkg['version']}")
+        add(f"* License: {pkg['license']}")
+        add(f"* Source: {pkg['url']}\n")
+
+        if pkg["texts"]:
+            for filename in sorted(pkg["texts"]):
+                body = pkg["texts"][filename].rstrip("\n")
+                fence = fence_for(body)
+                add(f"#### {filename}\n")
```
njhensley, line 356: **🔵 Nitpick — Filename emitted as raw Markdown heading**

`add(f"#### {filename}")` emits the license filename unsanitized; for License-File-declared names the value comes straight from wheel METADATA and never passes the `_LICENSE_FILE` regex, and `fence_for` wraps only the body. Given the unpinned PyPI closure (see the requirements.txt finding), a hostile/typosquatted wheel could inject Markdown (links/images/headings) into the published notices.

**Blast radius:** Low — dev-time, reviewed, non-executed document; worst case is defaced Markdown / phishing-link surface in the release artifact. No code execution.

**Fix:** Strip control chars / newlines and neutralize leading `#`/backtick runs before emitting the heading, and apply `_LICENSE_FILE` validation to License-File-declared names too.

tools/generate-notices
```diff
@@ -59,6 +74,69 @@ if ! command -v yq >/dev/null 2>&1; then
...
+if command -v sha256sum >/dev/null 2>&1; then
+    req_sha=$(sha256sum "${PYTHON_REQUIREMENTS}" | cut -d' ' -f1)
+elif command -v shasum >/dev/null 2>&1; then
+    req_sha=$(shasum -a 256 "${PYTHON_REQUIREMENTS}" | cut -d' ' -f1)
+else
+    echo "ERROR: neither sha256sum nor shasum is available to verify" >&2
+    echo "${PYTHON_NOTICES} against ${PYTHON_REQUIREMENTS}." >&2
+    exit 1
+fi
+if ! grep -qF "${req_sha}" "${PYTHON_NOTICES}"; then
```
njhensley, line 115: **🔵 Nitpick — sha256 staleness grep is unanchored over the whole fragment**

`grep -qF "${req_sha}"` accepts the digest appearing anywhere in the fragment (which embeds arbitrary license text) rather than binding it to the `Generated by … (sha256 \`…\`)` provenance line.

**Blast radius:** A 64-hex collision in license text is cryptographically impossible, so no practical risk — but the check never verifies the provenance line itself and is broader than intended.

**Fix:** Anchor to the rendered line, e.g. `grep -qF "(sha256 \`${req_sha}\`)"`.

tools/generate-notices
```diff
@@ -234,19 +335,33 @@ EOF
...
+    cat "${PYTHON_NOTICES}"
 } > "${OUTPUT}"
 
-ENTRIES=$(wc -l < "${INDEX_FILE}" | tr -d ' ')
-echo "Wrote ${OUTPUT} (${ENTRIES} dependencies)" >&2
+GO_ENTRIES=$(wc -l < "${INDEX_FILE}" | tr -d ' ')
+# Count the index table rows, not '### ' headings: license texts are embedded
+# verbatim and some contain Markdown headings of their own, which inflated the
+# count (Shakespeare/Getty/tiktoken data notices in this closure).
+PY_ENTRIES=$(grep -c '^| `' "${PYTHON_NOTICES}" || true)
```
njhensley, line 366: **🔵 Nitpick — PY_ENTRIES can count embedded license-text table rows**

`grep -c '^| \`'` counts index-table rows but would also count any `| \``-prefixed line inside an embedded license text (e.g. a Markdown table in a license).

**Blast radius:** Cosmetic — feeds only the informational stderr summary echo, gates nothing.

**Fix:** Restrict the count to the index table (between the header row and the next blank line / `## ` heading).

njhensley (commented):
# 🔍 Multi-Persona Review — Python dependency licenses

**Method:** 4 independent persona reviewers (Correctness · Supply-chain/License · CI-DX/Operability · Security), each finding adversarially re-derived by a senior meta-reviewer against the resolved code at `2a016017`.

## Overall assessment

Careful, unusually well-tested tooling — synthetic-wheel fixtures, mutation-verified tests, byte-deterministic offline regeneration, and thoughtful fail-closed gates. The Go-side fixes (dynamic fence sizing, non-license-file filtering) address real, diff-invisible corruption and are well pinned by tests. **No blockers.**

The one substantive issue is an accuracy gap between what the notices *render* and what the image *ships* (see the inline comment on `requirements.txt`), which undercuts the PR's "single input … cannot describe different closures" claim. Everything else is polish or latent hardening.

## Inline findings

9 inline comments are attached to the diff, each tagged with its tier. The crux is the 🟠 **Major** on `validators/performance/requirements.txt`: the transitive closure is unpinned, so the image build and the license collector resolve it independently against mutable PyPI, and the only staleness gate hashes `requirements.txt` bytes that never change on upstream drift — so the attribution record can silently diverge from the shipped image with no CI signal. Recommended fix: a fully-pinned + hashed lock shared by the Dockerfile install and the collector, hashed by the gate.

## Additional findings (couldn't anchor inline)

The committed fragment `validators/performance/licenses/python-notices.md` is too large for GitHub to expose a diff patch, so these two anchor to it by line:

- 🟡 **Minor — License column carries garbage / prose / inconsistent SPDX** (`python-notices.md`, e.g. L66 `kiwisolver`). `clean_license()` passes any ≤60-char first line lacking permission/copyright, so the fragment records `kiwisolver` = `=========================` (a rule line), `matplotlib` = `License agreement for matplotlib versions 1.3.0 and later` (prose), `choreographer` = `# MIT License` (raw Markdown), plus pervasive spelling drift (`Apache-2.0`/`Apache 2.0`/`Apache License 2.0`; `BSD-3-Clause`/bare `BSD`). Presentational — verbatim text below each entry is correct and the table stays well-formed — but `=========` as a license identifier is indefensible if this artifact is cited. Prefer the PEP 639 `License-Expression` and reject non-SPDX-shaped legacy values to the pointer instead of emitting them.
- 🟡 **Minor — `tokenizers` entry contradicts itself** (`python-notices.md`, L130). The index row says License = "See license text below", but the text section renders "License text unavailable in the distributed package" (no license file shipped) — a circular pointer for an Apache-2.0 package with attribution obligations. When `texts` is empty, override the pointer label so label and body agree.

Two more couldn't anchor (a context line / an out-of-diff line):

- 🔵 **Nitpick — `LICENSES_DIR` rm -rf guard is a denylist** (`tools/generate-notices` ~L142). Rejects only `""|"/"|"."|".."`; passes `//`, `~`, `$HOME`, and arbitrary absolute paths. Operator-scoped env override, self-inflicted, unchanged by this PR. Optional: require the resolved path to be a descendant of the repo working tree before deleting.
- 🔵 **Nitpick — `yq` relied on but not explicitly installed** (`.github/workflows/merge-gate.yaml` notices-freshness job, ~L499). `tools/generate-notices` hard-requires `yq` and exits without it, but the job installs only `go-licenses` and relies on the `ubuntu-latest` image bundling `yq`. It currently does; if the runner image drops it, the required gate breaks with an opaque error. Install/verify `yq` explicitly alongside `go-licenses`.

## ✅ Confirmed non-issues (examined, cleared)

- **Go MPL-2.0 source-file filtering** (dropping the 143 non-license files `go-licenses save` copied) — the judgment call flagged in the PR body. **Cleared:** `THIRD_PARTY_NOTICES.md` is an attribution document, not a source-distribution channel; MPL-2.0 §3.2 source availability is discharged by the retained Source URL plus unmodified vendored redistribution, and the Python MPL packages (`certifi`, `orjson`, `tqdm`) get the identical license-text-only treatment. No obligation broken.
- **`make notices` determinism, `fence_for` sizing, `clean_license`/`normalize`, mktemp templates, the `pip download` subprocess construction (list form, no shell), and the collect-tags list** — all re-derived and correct.

## Summary

|:---:|:---:|:---:|:---:|
| 0 | 1 | 6 | 6 |

**Recommendation: Approve with comments.** The crux is the `requirements.txt` major — pin + hash the closure so the attribution provably matches the shipped image (or explicitly accept the drift window and soften the "cannot diverge" language). The `platform_machine` guard, the license-file regex, and the dead fail-closed tests are cheap hardening in the same spirit; the fragment license-column cleanups and the stale "docker" guidance are good follow-up polish.

<sub>Generated by a multi-persona + adversarial-meta-review pass. Personas surface; the meta-reviewer confirms/refutes against the code; only survivors are posted.</sub>

validators/performance/aiperf-bench.Dockerfile
```diff
@@ -57,8 +57,22 @@ RUN apt-get update \
...
+# so that this image and the third-party license record in
+# THIRD_PARTY_NOTICES.md are driven by ONE input. tools/generate-python-licenses
+# installs the same file to enumerate what ships here; without a shared input it
+# would have to reconstruct this environment and could silently diverge from it.
+COPY validators/performance/requirements.txt /tmp/requirements.txt
+
+# AIPERF_VERSION now feeds only the io.aicr.aiperf.version label below, so it
+# can drift from what is actually installed. Fail the build closed instead of
+# publishing an image whose label misreports its own contents.
+RUN grep -qE "^aiperf==${AIPERF_VERSION}([[:space:]]|$)" /tmp/requirements.txt \
```
njhensley, line 83: **🔵 Nitpick — AIPERF_VERSION interpolated unquoted into an ERE**

`grep -qE "^aiperf==${AIPERF_VERSION}([[:space:]]|$)"` interpolates the build ARG unescaped, so ERE metacharacters in the value would alter the pattern.

**Blast radius:** Not attacker-reachable — AIPERF_VERSION is a build ARG (default 0.11.0), no workflow passes it, and it fails closed correctly on mismatch today. Hardening only.

**Fix:** Escape dots / anchor the pattern, or do a literal string comparison; optionally validate the ARG against a semver shape.

mchmarny (approved):

PR #2015, opened 2026-08-04
constraints: reject bare bang expressions

### LINKED ISSUE #2001: constraints: bare '!'-prefixed scalar value parses as exact match that can never match
## Problem

In `pkg/constraints`, a constraint value beginning with a bare `!` token (e.g. `value: "!some-label-key"`) on an ordinary scalar constraint parses cleanly as `OperatorExact` with the literal value `"!some-label-key"` — an exact string match that can never match any reading. The constraint silently never passes (or, for a negative-testing author expecting negation semantics, silently never expresses what they meant), with no parse error.

`!=` exists as a real operator prefix (`value: "!= foo"`), so a `!`-prefixed bare token is plausibly an author's attempt at negation and is always a mistake — but today it fails silently rather than loudly.

## Context

Found while implementing the node-set constraint form for #1755 (PR #2000), whose value grammar legitimately uses `!<label-key>` for its negated predicate — but only under the name-dispatched `NodeTopology.gpu-nodes.label` form. The scalar path's acceptance of the same shape as an exact match invites confusion between the two grammars.

## Proposed fix

In `ParseConstraintExpression`, reject a value matching `^!\S` that is not `!=` with `ErrCodeInvalidRequest` ("did you mean `!=`, or the node-set form?"), or at minimum log a warning. Fail-closed is the right direction: a constraint that can never match is a validation gate that never gates.

## Scope note

Deliberately excluded from PR #2000 (pre-existing behavior, not introduced there) per the bounded-scope rule.

krtadev:
## Summary

Reject bare `!` prefixes in scalar constraint expressions instead of treating them as exact-match values. Valid `!=` expressions continue to work unchanged.

## Motivation / Context

A scalar expression such as `!some-label-key` was previously accepted as a literal exact-match value, even though bare `!` syntax is only valid in the separate node-set form.

Fixes: #2001  
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: Constraint parsing (`pkg/constraints`)

## Implementation Notes

- Reject expressions matching a bare `!` followed immediately by a non-whitespace character.
- Preserve valid not-equal expressions such as `!= ubuntu` and `!=ubuntu`.
- Return `ErrCodeInvalidRequest` with guidance to use `!=` or the separate node-set form.
- Add regression tests for `!some-label-key` and `!ubuntu`, including error-code verification.

## Testing

```bash
go test -race ./pkg/constraints/...
make qualify
```

Both commands passed locally. The behavior was also verified through `aicr validate --no-cluster` using a recipe containing `value: "!ubuntu"`; validation rejected it with the expected `INVALID_REQUEST` error.

## Risk Assessment

**Rollout notes:** N/A. This only rejects previously accepted invalid scalar syntax.

## Checklist

- [x] I updated docs if user-facing behavior changed (N/A: no documentation change required)


mchmarny (approved):
Reviewed exact head 7470370a. No code blockers found.

PR #2033, opened 2026-08-04
fix(kwok): fail closed on unmapped profiles; skip in batch

### LINKED ISSUE #1997: kwok: fail closed when no node profile matches the recipe
## Problem

`get_profiles()` in `kwok/scripts/apply-nodes.sh` hardcodes service+accelerator →
profile mappings and silently falls back to `eks/p5-h100.yaml` (amd64 H100) for
any unknown combination.

PR #1985 demonstrated the failure mode: its four GB300 GKE lanes ran green while
the Tier 2 CI logs show the simulated nodes were amd64 H100 (`Profiles:
system=eks/system-m7i.yaml, gpu=eks/p5-h100.yaml`). The arm64 GB300 profiles the
PR added were never exercised, so the green checks were misleading.

## Proposal

- Fail closed: error out when no exact service+accelerator profile mapping
  exists, instead of defaulting.
- Before applying nodes, verify the selected profile's `arch` and accelerator
  labels agree with the recipe criteria; mismatch is an error.

This matches the repo's fail-closed rule for presence/negative checks
(CLAUDE.md anti-patterns).

## Context

Surfaced in the PR #1985 (GKE GB300) review.

framsouza:
Replace `apply-nodes.sh`'s hardcoded profile mapping with label-driven
discovery so unmapped `(service, accelerator)` combinations can no longer
silently fall back to `eks/p5-h100.yaml`.

## Motivation / Context

`get_profiles()` in `kwok/scripts/apply-nodes.sh` hardcoded a small
service+accelerator → profile map and defaulted everything else to
amd64 H100. #1985 hit the failure mode: four GB300 GKE lanes reported
green while the CI logs showed the simulated nodes were amd64 H100 —
the arm64 GB300 profiles the PR added were never exercised. The tests
said "your recipe schedules correctly" when they had actually tested
completely different hardware.

Fixes: #1997
Related: #1985

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: KWOK simulation scaffolding (`kwok/scripts/`, `kwok/README.md`, `.github/workflows/kwok-recipes.yaml`)

## Implementation Notes

Profile selection moves to `kwok/scripts/lib/profile-select.sh` and is
driven by profile-declared labels rather than a hardcoded case
statement. `select_profiles(service, accelerator, profiles_root)` walks
`kwok/profiles/<service>/` and matches:

| Role   | Match rule |
|--------|------------|
| system | `provider == <service>` and `nodeType == system` |
| gpu    | `provider == <service>` and `nodeType == accelerated` and `accelerator == <accelerator>` |

Exactly one match per role is required. Zero or multiple matches — or
an unknown service/accelerator — is an error, with a diagnostic that
enumerates what is on disk. Adding a profile now requires only a YAML
file with the right labels; no bash edit.

Direct-vs-batch semantics are intentionally different:

- **Direct** (`apply-nodes.sh <recipe>`, `make kwok-e2e RECIPE=…`) —
fails closed with the full diagnostic. If you explicitly ask about a
recipe, you get the honest answer.
- **Batch** (`run-all-recipes.sh`, `make kwok-test-all`, CI) — skips
unmapped recipes with a `WARN` and returns `rc=0`. Matrix coverage
stays green while profiles are backfilled in follow-up PRs; a recipe
that can't be simulated is not the same as a recipe that failed.

`resolve_recipe_criteria` is extracted alongside `select_profiles` so
`apply-nodes.sh` and `run-all-recipes.sh` share one defaulting policy
(`null`/`any` → `eks`/`h100`) and cannot drift.

Exposed by this change (currently silently passing, will `SKIP` after
merge until profiles are added): `a100-eks-*`, all `gke-*`, all `aks-*`,
all `oke-*`, all `bcm-*` recipes. Each is a candidate for a follow-up
profile PR.

## Testing

```bash
bash kwok/scripts/lib/profile-select_test.sh   # 14/14 pass
bash kwok/scripts/lib/sync-budget_test.sh      #  9/9 pass (unaffected)
shellcheck -x kwok/scripts/lib/profile-select*.sh kwok/scripts/apply-nodes.sh kwok/scripts/run-all-recipes.sh
yamllint .github/workflows/kwok-recipes.yaml
bash -n kwok/scripts/{apply-nodes.sh,run-all-recipes.sh,lib/profile-select.sh,lib/profile-select_test.sh}
```

New unit tests cover: happy paths (eks/h100, eks/gb200), unmapped
service, unmapped accelerator, missing system profile, ambiguous
system/gpu profiles, mislabeled provider in wrong directory, argument
validation, nonexistent roots, and resolve_recipe_criteria
explicit/default/any/missing-overlay paths. End-to-end skip logic
verified against the real overlay tree — 4 mapped recipes proceed,
7 unmapped ones skip cleanly.

Risk Assessment

- [x] Low — Isolated to KWOK scripts; no Go code touched, no
production code paths affected, purely CI/dev-loop scaffolding.

Rollout notes: No user-visible change. Recipes without matching
profiles will move from silent false-passes to explicit SKIP lines in
the CI logs; follow-up PRs add the missing profiles per provider/GPU.

Checklist

- [x] Tests pass locally (bash kwok/scripts/lib/profile-select_test.sh; no Go changes so make test is unchanged)
- [x] Linter passes (shellcheck, yamllint, bash -n)
- [x] I updated docs if user-facing behavior changed (kwok/README.md)
- [x] Changes follow existing patterns in the codebase (mirrors lib/sync-budget.sh + sync-budget_test.sh)

### DIFF
--- a/.github/workflows/kwok-recipes.yaml
+++ b/.github/workflows/kwok-recipes.yaml
@@ -92,7 +92,14 @@ jobs:
           fetch-depth: 0
 
       - name: Script unit tests (kwok/scripts/lib)
-        run: bash kwok/scripts/lib/sync-budget_test.sh
+        run: |
+          # yq is required by profile-select_test.sh; fail closed here so a
+          # runner image without yq cannot silently skip the assertions and
+          # let the CI light lie about coverage (#1997).
+          command -v yq >/dev/null || { echo "::error::yq is required for kwok/scripts/lib tests"; exit 1; }
+          bash kwok/scripts/lib/sync-budget_test.sh
+          bash kwok/scripts/lib/profile-select_test.sh
+          bash kwok/scripts/run-all-recipes_test.sh
 
       - name: Classify recipes into tiers
         id: classify
@@ -101,6 +108,64 @@ jobs:
           DISPATCH_RECIPE: ${{ github.event.inputs.recipe }}
         run: |
           set -euo pipefail
+
+          # Shared profile-selection helpers. Discovery filters out any
+          # recipe whose (service, accelerator) has no matching KWOK profile
+          # on disk, so the matrix never dispatches a job that would silently
+          # "SKIP" and return green without validating anything (#1997).
+          # shellcheck source=kwok/scripts/lib/profile-select.sh
+          source kwok/scripts/lib/profile-select.sh
+          PROFILES_ROOT="${GITHUB_WORKSPACE}/kwok/profiles"
+
+          # read_overlay_field <overlay> <yq_expr> — prints the resolved
+          # value on stdout (or "" when the field is absent/null). yq
+          # evaluation failures — a malformed overlay YAML — surface as
+          # a ::error:: annotation and return 1 rather than being
+          # swallowed by `|| true` and collapsed to an empty string,
+          # which the downstream filter would misread as "non-testable,
+          # skip". Same silent-swallow class the rest of the PR is
+          # closing.
+          #
+          # Diagnostics go to stderr so they don't get captured into the
+          # caller's variable via $(...) and become an annotation only in
+          # the step log. Caller MUST use `|| exit 1` — an inner exit
+          # inside command substitution only kills the subshell.
+          read_overlay_field() {
+            local overlay="$1" expr="$2" value
+            if ! value=$(yq eval "${expr} // \"\"" "${overlay}" 2>/dev/null); then
+              echo "::error file=${overlay}::yq failed to parse overlay (reading ${expr})" >&2
+              yq eval "${expr}" "${overlay}" >&2 || true
+              return 1
+            fi
+            echo "${value}"
+          }
+
+          # profile_status <overlay> — prints 0 to stdout if a profile exists
+          # for the recipe's (service, accelerator), PROFILE_SELECT_RC_NO_MATCH
+          # (2) if no profile exists (recipe is not in scope for this
+          # checkout), or exits 1 with an ::error:: annotation on any other
+          # selector failure (malformed criteria, malformed profile,
+          # ambiguous match — the tree is broken and must not be masked by
+          # dropping the recipe from the matrix).
+          profile_status() {
+            local overlay="$1" criteria svc accel rc=0
+            if ! criteria=$(resolve_recipe_criteria "${overlay}" 2>/dev/null); then
+              echo "::error file=${overlay}::resolve_recipe_criteria failed"
+              resolve_recipe_criteria "${overlay}" >&2 || true
+              exit 1
+            fi
+            read -r svc accel <<< "${criteria}"
+            select_profiles "${svc}" "${accel}" "${PROFILES_ROOT}" >/dev/null 2>&1 || rc=$?
+            if (( rc == 0 || rc == PROFILE_SELECT_RC_NO_MATCH )); then
+              echo "${rc}"
+              return 0
+            fi
+            echo "::error file=${overlay}::select_profiles failed (rc=${rc}) for service=${svc} accelerator=${accel} — see stderr"
+            select_profiles "${svc}" "${accel}" "${PROFILES_ROOT}" >&2 || true
+            exit 1
+          }
+
+          dropped="[]"   # recipes filtered out of the matrix (no profile)
 
           # --- Deployer list: single source of truth (#1172) ---
           # Defined once, here (before dispatch, since dispatch also needs it).
@@ -112,6 +177,25 @@ jobs:
 
           # --- workflow_dispatch: test exactly the requested recipe ---
           if [[ -n "${DISPATCH_RECIPE}" ]]; then
+            dispatch_overlay="recipes/overlays/${DISPATCH_RECIPE}.yaml"
+            if [[ ! -f "${dispatch_overlay}" ]]; then
+              echo "::error::Dispatch recipe '${DISPATCH_RECIPE}' has no overlay at ${dispatch_overlay}"
+              exit 1
+            fi
+            # workflow_dispatch is an explicit ask; an unmapped recipe here
+            # must fail (not silently produce an empty matrix) so the
+            # operator knows their request cannot be validated.
+            # Capture rc separately: an exit 1 inside profile_status only
+            # kills the $(...) subshell, so a fatal path (malformed
+            # profile / ambiguous match) would otherwise be swallowed
+            # and misread as the unmapped-skip case below.
+            status=$(profile_status "${dispatch_overlay}") || exit 1
+            if [[ "${status}" == "${PROFILE_SELECT_RC_NO_MATCH}" ]]; then
+              echo "::error::Dispatch recipe '${DISPATCH_RECIPE}' has no KWOK profile for its"
+              echo "::error::(service, accelerator). Add one under kwok/profiles/<service>/ or"
+              echo "::error::dispatch a different recipe."
+              exit 1
+            fi
             single=$(jq -cn --arg r "${DISPATCH_RECIPE}" --argjson deployers "$deployers" \
               '[ $deployers[] as $d | {recipe: $r, deployer: $d} ]')
             echo "tier1_pairs=${single}" >> "$GITHUB_OUTPUT"
@@ -127,16 +211,30 @@ jobs:
 
           for overlay in recipes/overlays/*.yaml; do
             name=$(basename "$overlay" .yaml)
-            service=$(yq eval '.spec.criteria.service // ""' "$overlay" 2>/dev/null || true)
+            service=$(read_overlay_field "$overlay" '.spec.criteria.service') || exit 1
 
             # Skip non-testable overlays (no service, or OCP — needs OpenShift operators)
             if [[ -z "$service" || "$service" == "null" || "$service" == "any" || "$service" == "ocp" ]]; then
               continue
             fi
 
+            # Drop recipes with no matching KWOK profile so the matrix cell
+            # can't report green without running anything. Fatal selector
+            # errors (ambiguous match, malformed profile) inside
+            # profile_status call exit 1, but that only kills the $(...)
+            # subshell — the outer script has to check the substitution's
+            # rc explicitly and propagate it, otherwise a broken tree gets
+            # silently reclassified as an unmapped-skip and hidden.
+            status=$(profile_status "${overlay}") || exit 1
+            if [[ "${status}" == "${PROFILE_SELECT_RC_NO_MATCH}" ]]; then
+              dropped=$(echo "$dropped" | jq -c --arg r "$name" '. + [$r]')
+              echo "::notice file=${overlay}::${name} filtered from KWOK matrix — no profile for its (service, accelerator)"
+              continue
+            fi
+
             all=$(echo "$all" | jq -c --arg r "$name" '. + [$r]')
 
-            accel=$(yq eval '.spec.criteria.accelerator // ""' "$overlay" 2>/dev/null || true)
+            accel=$(read_overlay_field "$overlay" '.spec.criteria.accelerator') || exit 1
             if [[ -z "$accel" || "$accel" == "null" || "$accel" == "any" ]]; then
               tier1=$(echo "$tier1" | jq -c --arg r "$name" '. + [$r]')
             fi
@@ -165,14 +263,25 @@ jobs:
             # Build set of affected accelerator-specific overlays
             for overlay in recipes/overlays/*.yaml; do
               name=$(basename "$overlay" .yaml)
-              service=$(yq eval '.spec.criteria.service // ""' "$overlay" 2>/dev/null || true)
-              accel=$(yq eval '.spec.criteria.accelerator // ""' "$overlay" 2>/dev/null || true)
+              service=$(read_overlay_field "$overlay" '.spec.criteria.service') || exit 1
+              accel=$(read_overlay_field "$overlay" '.spec.criteria.accelerator') || exit 1
 
               # Only accelerator-specific, KWOK-testable overlays belong in Tier 2
               if [[ -z "$service" || "$service" == "null" || "$service" == "any" || "$service" == "ocp" ]]; then
                 continue
               fi
               if [[ -z "$accel" || "$accel" == "null" || "$accel" == "any" ]]; then
+                continue
+              fi
+
+              # Same profile-availability filter as the main loop — an
+              # unmapped accelerator recipe must not enter the matrix, or
+              # its Tier 2 cell reports green after skipping (#1997).
+              # Same subshell-exit trap as above: capture rc explicitly
+              # so fatal selector errors propagate instead of being
+              # silently reclassified as unmapped-skips.
+              status=$(profile_status "${overlay}") || exit 1
+              if [[ "${status}" == "${PROFILE_SELECT_RC_NO_MATCH}" ]]; then
                 continue
               fi
 
@@ -302,6 +411,10 @@ jobs:
           echo "Tier 1 (generic):      $(echo "$tier1" | jq 'length') recipe(s) × $(echo "$deployers" | jq 'length') deployer(s) = ${tier1_count} pair(s)"
           echo "Tier 2 (diff-aware):   $(echo "$tier2" | jq 'length') recipe(s) × 1 deployer (helm) = ${tier2_count} pair(s)"
           echo "Tier 3 (full matrix):  $(echo "$tier3" | jq 'length') recipe(s) × $(echo "$deployers" | jq 'length') deployer(s) = ${tier3_pairs} pair(s) in ${tier3_batch_count} batch(es)"
+          dropped_count=$(echo "$dropped" | jq 'length')
+          if (( dropped_count > 0 )); then
+            echo "Filtered from matrix: ${dropped_count} recipe(s) with no matching KWOK profile — $(echo "$dropped" | jq -r 'join(", ")')"
+          fi
   # Thin caller of the shared kwok-test-run.yaml reusable workflow (#1172).
   # tier1_pairs is well under 256 entries, so it's passed in a single call —
   # no batching needed.
--- a/kwok/README.md
+++ b/kwok/README.md
@@ -52,15 +52,55 @@ flowchart LR
 | Scripts | `kwok/scripts/` | Create nodes, validate scheduling |
 | CI Workflow | `.github/workflows/kwok-recipes.yaml` | Auto-discover and test recipes |
 
-## Profile Mapping
+## Profile Selection
 
-`apply-nodes.sh` reads recipe criteria and selects matching profiles:
+Selection happens in two steps: normalize the recipe criteria, then match
+profiles against the normalized values.
 
-| Service | Accelerator | GPU Profile |
-|---------|-------------|-------------|
-| eks | h100 (default) | `eks/p5-h100.yaml` |
-| eks | gb200 | `eks/p6-gb200.yaml` |
-| other | any | `eks/p5-h100.yaml` (fallback) |
+**1. Criteria normalization.** `resolve_recipe_criteria` in
+`kwok/scripts/lib/profile-select.sh` reads
+`spec.criteria.{service,accelerator}` from the overlay and applies:
+
+| Input value | Normalized to |
+|-------------|---------------|
+| missing / `null` — `service` | `eks` |
+| `any` — `service` | `eks` |
+| missing / `null` — `accelerator` | `h100` |
+| `any` — `accelerator` | `h100` |
+| any other value | passed through verbatim |
+
+Both `apply-nodes.sh` (direct path) and `run-all-recipes.sh` (batch path)
+call this same resolver, so the two entry points cannot drift on how
+placeholder values collapse.
+
+**2. Profile matching.** The normalized `(service, accelerator)` pair is
+then looked up under `kwok/profiles/<service>/` by matching each
+candidate's `metadata.labels`:
+
+| Role | Match rule |
+|------|------------|
+| system | `provider == <service>` and `nodeType == system` |
+| gpu | `provider == <service>` and `nodeType == accelerated` and `accelerator == <accelerator>` |
+
+Exactly one match per role is required. Zero or multiple matches — or an
+unknown service or accelerator — is a hard error; there is no silent
+fallback to another provider (see #1997). Selection is implemented in
+`kwok/scripts/lib/profile-select.sh` and unit-tested by the sibling
+`profile-select_test.sh` (wired into the `kwok-recipes.yaml` discover job).
+
+**Direct vs batch semantics.** `apply-nodes.sh <recipe>` (and
+`make kwok-e2e RECIPE=...`) fails closed with the full diagnostic when
+no profile matches. `run-all-recipes.sh` (and `make kwok-test-all`,
+plus the CI matrix) instead **skips** unmapped recipes with a WARN so
+batch coverage stays green while profiles are backfilled — a recipe
+you can't test isn't the same as a recipe that failed.
+
+Currently on disk:
+
+| Service | Accelerator | System profile | GPU profile |
+|---------|-------------|----------------|-------------|
+| eks | h100 | `eks/system-m7i.yaml` | `eks/p5-h100.yaml` |
+| eks | gb200 | `eks/system-m7i.yaml` | `eks/p6-gb200.yaml` |
 
 **Cluster defaults:** 2 system nodes, 4 GPU nodes (32 GPUs), Kubernetes v1.33.5, region `us-east-1`.
 
@@ -128,13 +168,27 @@ Test it: `unset GITLAB_TOKEN && make build && make kwok-e2e RECIPE=your-recipe-n
 
 ## Adding Node Profiles or Cloud Providers
 
-Copy an existing profile and modify, then update the mapping in `kwok/scripts/apply-nodes.sh` (`get_profiles()` around line 52):
+Profiles are discovered by label — no script edit is needed. Copy an
+existing profile, update `metadata.labels.{provider,nodeType,accelerator}`
+and `spec.*`, and drop it under `kwok/profiles/<provider>/`:
 
 ```bash
 cp kwok/profiles/eks/p5-h100.yaml kwok/profiles/eks/p5-a100.yaml
+# then edit metadata.labels.accelerator and spec.gpu.* to match the new instance type
 ```
 
-For new cloud providers, copy the `kwok/profiles/eks/` directory structure and update `apply-nodes.sh`.
+For a new cloud provider, create the `kwok/profiles/<provider>/` directory
+and add:
+
+- **exactly one** system profile with `metadata.labels.{provider: <provider>, nodeType: system}`, and
+- **exactly one** accelerated profile *per supported accelerator* with `metadata.labels.{provider: <provider>, nodeType: accelerated, accelerator: <accelerator>}`.
+
+`select_profiles` requires a unique match in each role and returns a
+fatal error (not a skip) on ambiguity, so a duplicate `nodeType: system`
+or two profiles carrying the same accelerator label will break both
+direct invocations and batch CI. The accelerator label is required —
+without it, a profile is silently skipped for GPU-role lookups.
+`apply-nodes.sh` picks up the new profiles on the next run.
 
 ## CI Integration
 
--- a/kwok/scripts/lib/profile-select.sh
+++ b/kwok/scripts/lib/profile-select.sh
@@ -1,0 +1,231 @@
+#!/usr/bin/env bash
+# shellcheck shell=bash
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+
+# Fail-closed KWOK profile selection driven by profile-declared labels.
+#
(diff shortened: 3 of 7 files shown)


.github/workflows/kwok-recipes.yaml
```diff
@@ -92,7 +92,9 @@ jobs:
           fetch-depth: 0
 
       - name: Script unit tests (kwok/scripts/lib)
-        run: bash kwok/scripts/lib/sync-budget_test.sh
+        run: |
+          bash kwok/scripts/lib/sync-budget_test.sh
+          bash kwok/scripts/lib/profile-select_test.sh
```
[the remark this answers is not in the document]
framsouza: Fix on https://github.com/NVIDIA/aicr/pull/2033/changes/b27f262b6b535946c2f2573a163ad5a838095b42

kwok/README.md
```diff
@@ -52,15 +52,36 @@ flowchart LR
...
+| Role | Match rule |
+|------|------------|
+| system | `provider == <service>` and `nodeType == system` |
+| gpu | `provider == <service>` and `nodeType == accelerated` and `accelerator == <accelerator>` |
+
+Exactly one match per role is required. Zero or multiple matches — or an
+unknown service or accelerator — is a hard error; there is no silent
+fallback to another provider (see #1997). Selection is implemented in
+`kwok/scripts/lib/profile-select.sh` and unit-tested by the sibling
+`profile-select_test.sh` (wired into the `kwok-recipes.yaml` discover job).
```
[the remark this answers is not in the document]
framsouza: Fix on https://github.com/NVIDIA/aicr/pull/2033/changes/d7f75afc1bc803630fe8e44c502e909372e2d25c

kwok/README.md
```diff
@@ -128,13 +149,19 @@ Test it: `unset GITLAB_TOKEN && make build && make kwok-e2e RECIPE=your-recipe-n
...
 ```bash
 cp kwok/profiles/eks/p5-h100.yaml kwok/profiles/eks/p5-a100.yaml
+# then edit metadata.labels.accelerator and spec.gpu.* to match the new instance type
 ```
 
-For new cloud providers, copy the `kwok/profiles/eks/` directory structure and update `apply-nodes.sh`.
+For a new cloud provider, create the `kwok/profiles/<provider>/` directory
+and add at least one `nodeType: system` and one `nodeType: accelerated`
+profile with `metadata.labels.provider: <provider>`. `apply-nodes.sh` will
+pick them up on the next run.
```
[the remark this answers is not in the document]
framsouza: Fix on https://github.com/NVIDIA/aicr/pull/2033/changes/a1d785b3cd5d45f8f8a84464201292936e448e31

kwok/scripts/lib/profile-select.sh
```diff
@@ -0,0 +1,146 @@
...
+# Prints "<service> <accelerator>" to stdout after applying the same
+# defaulting apply-nodes.sh uses: missing/null → eks/h100, and the
+# "any" placeholder collapses to the same defaults so KWOK has a
+# concrete pair to look up. Kept here so run-all-recipes.sh and
+# apply-nodes.sh cannot drift on this policy. Requires yq.
+resolve_recipe_criteria() {
+    local overlay_file="$1"
+    if [[ ! -f "${overlay_file}" ]]; then
+        echo "[ERROR] Recipe overlay not found: ${overlay_file}" >&2
+        return 1
```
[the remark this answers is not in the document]
framsouza: Fix on https://github.com/NVIDIA/aicr/pull/2033/changes/c65e6e70944e9d09a3920f7eea08cd7f91f2a7fa

kwok/scripts/lib/profile-select.sh
```diff
@@ -0,0 +1,146 @@
...
+    fi
+
+    local svc accel
+    svc=$(yq eval '.spec.criteria.service // "eks"' "${overlay_file}")
+    accel=$(yq eval '.spec.criteria.accelerator // "h100"' "${overlay_file}")
+    [[ "${svc}" == "any" || "${svc}" == "null" ]] && svc="eks"
+    [[ "${accel}" == "any" || "${accel}" == "null" ]] && accel="h100"
```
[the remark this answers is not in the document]
framsouza: Fix https://github.com/NVIDIA/aicr/pull/2033/changes/ea60d2e89f44891f0afd452e31f9a514befabeed

kwok/scripts/lib/profile-select.sh
```diff
@@ -0,0 +1,146 @@
...
+                available="${d##*/}"
+            else
+                available="${available},${d##*/}"
+            fi
+        done
+        if [[ -n "${available}" ]]; then
+            echo "[ERROR] Services with profiles on disk: ${available}" >&2
+        fi
+        return 1
+    fi
```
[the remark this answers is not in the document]
framsouza: Fix https://github.com/NVIDIA/aicr/pull/2033/changes/7b9aade8f88445f81858c5c85d4bf8795861d09c

kwok/README.md
```diff
@@ -128,13 +168,24 @@ Test it: `unset GITLAB_TOKEN && make build && make kwok-e2e RECIPE=your-recipe-n
...
 cp kwok/profiles/eks/p5-h100.yaml kwok/profiles/eks/p5-a100.yaml
+# then edit metadata.labels.accelerator and spec.gpu.* to match the new instance type
 ```
 
-For new cloud providers, copy the `kwok/profiles/eks/` directory structure and update `apply-nodes.sh`.
+For a new cloud provider, create the `kwok/profiles/<provider>/` directory
+and add at least:
+
+- one system profile with `metadata.labels.{provider: <provider>, nodeType: system}`, and
+- one accelerated profile with `metadata.labels.{provider: <provider>, nodeType: accelerated, accelerator: <accelerator>}`.
```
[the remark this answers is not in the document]
framsouza: Fix https://github.com/NVIDIA/aicr/pull/2033/changes/a0dfa88456059b02025f53446280816485565644

kwok/scripts/lib/profile-select_test.sh
```diff
@@ -0,0 +1,262 @@
...
+write_overlay "${OVERLAY_DIR}/missing.yaml"
+got_out=$(resolve_recipe_criteria "${OVERLAY_DIR}/missing.yaml" 2>/dev/null)
+got_rc=$?
+check "resolve-missing-defaults-to-eks-h100" 0 "eks h100" "" "${got_rc}" "${got_out}" ""
+
+# 14. 'any' placeholder collapses to the same defaults.
+write_overlay "${OVERLAY_DIR}/any.yaml" any any
+got_out=$(resolve_recipe_criteria "${OVERLAY_DIR}/any.yaml" 2>/dev/null)
+got_rc=$?
+check "resolve-any-collapses-to-defaults" 0 "eks h100" "" "${got_rc}" "${got_out}" ""
```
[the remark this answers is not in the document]
framsouza: Fix https://github.com/NVIDIA/aicr/pull/2033/changes/05a497032ffc8e57391160aa3077862e77327b2e

framsouza:
@coderabbitai resume

kwok/scripts/run-all-recipes.sh
```diff
@@ -242,6 +248,36 @@ run_recipe_test() {
...
+    local overlay_file="${OVERLAYS_DIR}/${recipe}.yaml"
+    if [[ -f "${overlay_file}" ]]; then
+        local criteria svc accel
+        if criteria=$(resolve_recipe_criteria "${overlay_file}" 2>/dev/null); then
+            read -r svc accel <<< "${criteria}"
+            local sel_err select_rc=0
+            sel_err=$(select_profiles "${svc}" "${accel}" "${KWOK_DIR}/profiles" 2>&1 >/dev/null) || select_rc=$?
+            if (( select_rc == PROFILE_SELECT_RC_NO_MATCH )); then
+                log_warn "SKIP ${recipe}: no KWOK profile for service=${svc} accelerator=${accel} (add one under kwok/profiles/${svc}/ — see #1997)"
+                return 0
```
mchmarny, line 284: Blocking: could we keep an unmapped profile from becoming a successful CI job? Each matrix cell calls run-all-recipes.sh with one explicit recipe, but this returns 0 and main records it in passed. On this head, 72 of 90 service-scoped overlays resolve to PROFILE_SELECT_RC_NO_MATCH, so those jobs can remain green without bundle or scheduling validation - the false-success direction #1997 is meant to close. Please make explicit or CI recipes propagate a nonzero result or gate no-coverage separately, and add a regression showing that a single unmapped recipe cannot be reported as passed.

mchmarny (changes requested):
Requesting changes: 1 remaining merge blocker on 2af2b69f.

Welcome to AICR, @framsouza, and thanks for taking on this false-positive gap. The label-driven selection and focused regression suite are a strong direction. I verified the latest malformed-profile fix locally (22/22 tests pass), so CodeRabbit's earlier blocker is addressed. The remaining inline issue is the CI no-match path: explicit matrix jobs can still report success without running bundle or scheduling validation. Once that path cannot produce a green false-success, happy to re-review.

Current required CI has not run yet because external-contributor validation is still pending.

framsouza:
@mchmarny thanks for catching that. I verified end-to-end locally (branch head f8242bfd). `kwok/scripts/run-all-recipes.sh gb200-oke-training`, the exact invocation `.github/actions/kwok-test/action.yml:178` uses per matrix cell exits 1 with `[ERROR] Recipe gb200-oke-training was explicitly requested but has no KWOK profile...`. Both the discovery filter (c0695738) and the script-layer defense (f8242bfd) are in place; regression at `kwok/scripts/run-all-recipes_test.sh` proves both sides of the asymmetry. Pls let me know if i'm missing something

kwok/scripts/run-all-recipes.sh
framsouza: Thanks for the specific reproduction, the 72/90 number pushed me to close both layers on branch head f8242bfd:

- Non-zero on explicit invocation: `kwok/scripts/run-all-recipes.sh gb200-oke-training` (exact form used at `.github/actions/kwok-test/action.yml:178`) exits 1 with a clear diagnostic. main records it in `failed[]`, not `passed[]`. Implicit batch mode still SKIPs so `make kwok-test-all` locally isn't disrupted.
- No-coverage gated at CI: c0695738 filters unmapped recipes at the classify step, local sim reproduces your 18-in / 72-dropped split.
- Regression: `kwok/scripts/run-all-recipes_test.sh` asserts `explicit-unmapped-recipe-is-not-reported-as-passed` (rc must be non-zero). Passes on this head.

.github/workflows/kwok-recipes.yaml
```diff
@@ -134,6 +189,16 @@ jobs:
...
 
+            # Drop recipes with no matching KWOK profile so the matrix cell
+            # can't report green without running anything. Fatal selector
+            # errors (ambiguous match, malformed profile) already exited 1
+            # inside profile_status.
+            if [[ "$(profile_status "${overlay}")" != "0" ]]; then
+              dropped=$(echo "$dropped" | jq -c --arg r "$name" '. + [$r]')
+              echo "::notice file=${overlay}::${name} filtered from KWOK matrix — no profile for its (service, accelerator)"
+              continue
+            fi
```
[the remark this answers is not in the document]
framsouza: Fix https://github.com/NVIDIA/aicr/pull/2033/changes/00c37cc0a02355f4b77f5503467af924bd5c2c67

.github/workflows/kwok-recipes.yaml
```diff
@@ -134,6 +195,20 @@ jobs:
...
             fi
 
+            # Drop recipes with no matching KWOK profile so the matrix cell
+            # can't report green without running anything. Fatal selector
+            # errors (ambiguous match, malformed profile) inside
+            # profile_status call exit 1, but that only kills the $(...)
+            # subshell — the outer script has to check the substitution's
+            # rc explicitly and propagate it, otherwise a broken tree gets
+            # silently reclassified as an unmapped-skip and hidden.
+            status=$(profile_status "${overlay}") || exit 1
```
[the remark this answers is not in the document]
framsouza: Fix https://github.com/NVIDIA/aicr/pull/2033/commits/f9946dd4f44774825141cc19f5e6559e1f302ddf

mchmarny:
@framsouza also please squash commits

mchmarny (approved):
APPROVE — prior blocker resolved on 1275249f. Explicit unmapped recipes now fail instead of being reported as passed, and the regression test covers the CI invocation shape. Focused shell tests pass; live CI has no failures and remaining lanes are still running.

.github/workflows/kwok-recipes.yaml
```diff
@@ -102,6 +109,64 @@ jobs:
...
+          # swallowed by `|| true` and collapsed to an empty string,
+          # which the downstream filter would misread as "non-testable,
+          # skip". Same silent-swallow class the rest of the PR is
+          # closing.
+          #
+          # Diagnostics go to stderr so they don't get captured into the
+          # caller's variable via $(...) and become an annotation only in
+          # the step log. Caller MUST use `|| exit 1` — an inner exit
+          # inside command substitution only kills the subshell.
+          read_overlay_field() {
```
njhensley, line 133: **🟡 Minor — CI read_overlay_field reuses the `// ""` falsy-swallow the library reader abandoned**

The library's `_read_criteria_field` was rewritten to reject `service: false` by YAML tag, but classify still reads `service` via `yq '... // ""'` here and does `-z`→`continue` (~L217) *before* the strict `resolve_recipe_criteria` runs. So `service: false` collapses to "" and is silently dropped as non-testable instead of erroring like the direct path does. Contrived input, but it's an inconsistency with the fix's own stated guarantee.

**Blast radius:** Coverage-exclusion of a malformed overlay (not a wrong-hardware false-pass); the direct run_recipe_test path still fails the same input strictly.

**Fix:** Route the service field through the strict reader in classify (or add a tag check) so a malformed-typed service errors instead of being silently excluded.

.github/workflows/kwok-recipes.yaml
```diff
@@ -102,6 +109,64 @@ jobs:
...
+          # for the recipe's (service, accelerator), PROFILE_SELECT_RC_NO_MATCH
+          # (2) if no profile exists (recipe is not in scope for this
+          # checkout), or exits 1 with an ::error:: annotation on any other
+          # selector failure (malformed criteria, malformed profile,
+          # ambiguous match — the tree is broken and must not be masked by
+          # dropping the recipe from the matrix).
+          profile_status() {
+            local overlay="$1" criteria svc accel rc=0
+            if ! criteria=$(resolve_recipe_criteria "${overlay}" 2>/dev/null); then
+              echo "::error file=${overlay}::resolve_recipe_criteria failed"
```
njhensley, line 153: **🔵 Nitpick — profile_status writes ::error:: to stdout inside $(...), so the annotation is captured and discarded**

Both `::error::` lines here (L153 and L163) lack `>&2`, unlike the sibling read_overlay_field (L136) whose header documents the rule. profile_status is always called as `status=$(profile_status ...)`, so these lines are captured into the discarded var and never render as GitHub annotations. Both branches `exit 1` (caught by `|| exit 1` at every call site), so the job still fails closed and the raw stderr re-emit (L154/L164) still shows — only the file-pinned annotation is lost.

**Blast radius:** Lost operator-facing annotation; no misclassification, no false green.

**Fix:** Add `>&2` to the two `echo "::error..."` lines to match read_overlay_field.

kwok/scripts/run-all-recipes.sh
```diff
@@ -242,6 +248,49 @@ run_recipe_test() {
...
+            sel_err=$(select_profiles "${svc}" "${accel}" "${KWOK_DIR}/profiles" 2>&1 >/dev/null) || select_rc=$?
+            if (( select_rc == PROFILE_SELECT_RC_NO_MATCH )); then
+                if is_explicit_recipe "${recipe}"; then
+                    log_error "Recipe ${recipe} was explicitly requested but has no KWOK profile for service=${svc} accelerator=${accel}. Add a profile under kwok/profiles/${svc}/ or drop this recipe from the invocation (see #1997)."
+                    return 1
+                fi
+                log_warn "SKIP ${recipe}: no KWOK profile for service=${svc} accelerator=${accel} (add one under kwok/profiles/${svc}/ — see #1997)"
+                return 0
+            fi
+            if (( select_rc != 0 )); then
```
njhensley, line 286: **🔵 Nitpick — FATAL rc=1 propagation in run_recipe_test is untested at the integration layer**

Both harness cases drive gb200-oke-training (rc=2 no-match), exercising the explicit-fail and SKIP branches but never this `select_rc != 0 → return select_rc` path — the dup/broken-profile-must-not-be-swallowed case central to mchmarny's concern. `select_profiles` returning 1 is unit-covered (profile-select_test tests 6/7), and this propagation line is trivially correct, so it's a coverage gap rather than a defect.

**Blast radius:** A future edit that reordered these branches to swallow rc=1 as a skip would have no regression guard.

**Fix:** Add a run-all-recipes_test.sh case with a temp KWOK_DIR fixture containing duplicate system/GPU profiles + matching overlay; assert run_recipe_test returns 1 in both explicit and implicit modes.

kwok/README.md
```diff
@@ -52,15 +52,55 @@ flowchart LR
...
+Exactly one match per role is required. Zero or multiple matches — or an
+unknown service or accelerator — is a hard error; there is no silent
+fallback to another provider (see #1997). Selection is implemented in
+`kwok/scripts/lib/profile-select.sh` and unit-tested by the sibling
+`profile-select_test.sh` (wired into the `kwok-recipes.yaml` discover job).
+
+**Direct vs batch semantics.** `apply-nodes.sh <recipe>` (and
+`make kwok-e2e RECIPE=...`) fails closed with the full diagnostic when
+no profile matches. `run-all-recipes.sh` (and `make kwok-test-all`,
+plus the CI matrix) instead **skips** unmapped recipes with a WARN so
```
njhensley, line 94: **🟡 Minor — README direct-vs-batch says the CI matrix 'skips unmapped recipes with a WARN' — CI actually DROPS at classify and explicit-fails**

Accurate only for implicit `make kwok-test-all`. In CI, discover DROPS the recipe at classify time (::notice, never dispatched); and an explicitly-named recipe reaching run-all-recipes.sh via action.yml:178 fails rc=1, not skips. This paragraph is the canonical description of the new fail-closed contract, so the drift is worth fixing.

**Blast radius:** Doc-only, but a future contributor could 'fix' the explicit-fail path believing CI is meant to skip, reopening the false-green hole.

**Fix:** Document three distinct modes — direct → fail closed; implicit make kwok-test-all → SKIP-with-WARN; CI → DROP at classify + explicit-fail — and qualify 'Zero matches is a hard error' as direct-path only.

kwok/scripts/lib/profile-select.sh
```diff
@@ -0,0 +1,231 @@
...
+        local provider nodeType accel relpath
+        # yq failures inside the loop propagate — a malformed profile is a
+        # broken tree, not a silent no-match. Rc=1 (fatal), not the
+        # skippable no-match code.
+        provider=$(_read_profile_label "${profile}" provider "${profiles_root}") || return 1
+        nodeType=$(_read_profile_label "${profile}" nodeType "${profiles_root}") || return 1
+        relpath="${profile#"${profiles_root}"/}"
+        # Skip anything whose provider label disagrees with its directory —
+        # protects against a mis-copied file living under the wrong service.
+        if [[ "${provider}" != "${service}" ]]; then
```
njhensley, line 190: **🔵 Nitpick — provider != directory mismatch is silently continue'd, not a hard error**

A profile whose provider label disagrees with its parent directory is skipped. For a stray file this is the intended defense (documented at L188-189). But if the mislabeled file is the SOLE system/GPU profile for that role, the mislabel degrades to an invisible rc=2 no-match (batch SKIP / CI drop) rather than surfacing the tree-integrity fault.

**Blast radius:** A single-character typo in provider on the sole eks profile silently zeroes coverage for every eks recipe, routed through the skippable no-match path.

**Fix:** Optional: return rc=1 with a diagnostic when a file under <service>/ declares provider=<other>, rather than continue.

kwok/scripts/run-all-recipes_test.sh
```diff
@@ -0,0 +1,106 @@
...
+}
+
+# 1. Explicit invocation — the operator named this recipe on the command
+# line, matching what every CI matrix cell does. An unmapped profile
+# must fail (rc != 0) rather than be reported as passed.
+# shellcheck disable=SC2034  # read by sourced run_recipe_test via is_explicit_recipe
+EXPLICIT_RECIPES="${UNMAPPED_RECIPE}"
+rc=0
+run_recipe_test "${UNMAPPED_RECIPE}" >/dev/null 2>&1 || rc=$?
+check "explicit-unmapped-recipe-is-not-reported-as-passed" ne 0 "${rc}"
```
njhensley, line 90: **🔵 Nitpick — Regression test 1 asserts only `rc != 0` with output discarded — vacuous-pass risk**

apply-nodes.sh is not stubbed (only cleanup_between_tests is). If the explicit-fail branch were removed, execution would fall through to apply-nodes.sh and fail there (no cluster) → rc!=0, and the test would still PASS — for a downstream failure, not the explicit-unmapped guard it claims to protect. `ne 0` can't distinguish 'failed for the right reason' from 'failed for any reason'.

**Blast radius:** The test proving explicit-unmapped fails closed can go green even after that logic is removed.

**Fix:** Assert rc == 1 specifically and capture the log to assert it contains the explicit-branch diagnostic ('explicitly requested but has no KWOK profile').

kwok/scripts/run-all-recipes_test.sh
```diff
@@ -0,0 +1,106 @@
...
+
+# Source run-all-recipes.sh without letting main() run. The script's
+# tail is the literal line `main "$@"`; strip it so we can source and
+# then invoke individual functions with stubs in place.
+SCRIPT_UNDER_TEST="${SCRIPT_DIR}/run-all-recipes.sh"
+# Keep the temp copy alongside the original so its ${SCRIPT_DIR} / lib
+# resolution (which uses BASH_SOURCE) points at the real lib directory.
+TMP_SOURCE=$(mktemp "${SCRIPT_DIR}/.run-all-recipes.test.XXXXXX.sh")
+trap 'rm -f "${TMP_SOURCE}"' EXIT
+grep -v '^main "\$@"$' "${SCRIPT_UNDER_TEST}" > "${TMP_SOURCE}"
```
njhensley, line 36: **🔵 Nitpick — Harness strips `main "$@"` with an exact-anchored grep; a reformat silently no-ops the strip**

`grep -v '^main "\$@"$'` removes the tail call by exact-line match before sourcing. If run-all-recipes.sh's tail line ever changes (trailing whitespace, `exec main`, a comment), the pattern stops matching, `main "$@"` stays in the temp copy, and sourcing executes main() with the test's args — driving into real kind/kubectl ops. No assertion that a line was removed.

**Blast radius:** Currently benign; would fail loudly (cluster ops) rather than pass silently, so future-fragility only.

**Fix:** Assert exactly one line was removed (or that `main "$@"` is absent from TMP_SOURCE) before sourcing; optionally define a no-op main() after sourcing.

kwok/scripts/lib/profile-select_test.sh
```diff
@@ -0,0 +1,317 @@
...
+run_select eks h100 "${BAD_ROOT}"
+check "malformed-profile-fails-fatal" 1 "" "Malformed KWOK profile eks/broken.yaml" \
+    "${got_rc}" "${got_out}" "${got_err}"
+rm -rf "${BAD_ROOT}"
+
+if (( fails > 0 )); then
+    echo "${fails} test(s) failed"
+    exit 1
+fi
+echo "All 22 tests passed"
```
njhensley, line 317: **🔵 Nitpick — 'All 22 tests passed' is a hardcoded literal, not a derived counter**

The success banner prints a literal count rather than a per-check counter. Counts match today, but if a contributor adds a case and forgets to bump the string — or an early exit skips trailing checks without failing — the harness can print 'All 22 tests passed' while a different number ran, and CI stays green.

**Blast radius:** A silently short-circuited or under-counted run reports success.

**Fix:** Increment a `ran` counter inside check() and print/assert it (`echo "All ${ran} tests passed"`).

kwok/scripts/lib/profile-select_test.sh
```diff
@@ -0,0 +1,317 @@
...
+write_overlay "${OVERLAY_DIR}/explicit.yaml" gke gb200
+got_out=$(resolve_recipe_criteria "${OVERLAY_DIR}/explicit.yaml" 2>/dev/null)
+got_rc=$?
+check "resolve-explicit-criteria" 0 "gke gb200" "" "${got_rc}" "${got_out}" ""
+
+# 13. Missing fields default to eks/h100 (matches apply-nodes.sh policy).
+write_overlay "${OVERLAY_DIR}/missing.yaml"
+got_out=$(resolve_recipe_criteria "${OVERLAY_DIR}/missing.yaml" 2>/dev/null)
+got_rc=$?
+check "resolve-missing-defaults-to-eks-h100" 0 "eks h100" "" "${got_rc}" "${got_out}" ""
```
njhensley, line 204: **🔵 Nitpick — Untested select_profiles / criteria branches**

No case covers: an unknown nodeType (the case statement has no default arm, so a typo'd nodeType silently drops the profile → skippable no-match), an accelerated profile missing its accelerator label (joins available_accels as ""), a mislabeled SYSTEM profile (test 8 covers only the accelerated role), or mixed explicit-service + defaulted-accelerator criteria.

**Blast radius:** A one-character typo in nodeType/provider on a real profile turns a should-run recipe into a silent skip — the same coverage-lie the PR targets, via a different field.

**Fix:** Add the four cases; consider a default nodeType arm that warns on unrecognized values and test it.

njhensley (approved):
## 🔬 Multi-persona review — approve with comments

**Method:** four independent persona reviewers (bash-correctness · CI/DX-operability · test-coverage · domain/fail-closed) whose findings were each re-derived from the resolved code by an adversarial senior meta-reviewer. Unit tests run locally (`profile-select_test` 22/22, `run-all-recipes_test` 2/2); `shellcheck -x` clean on the changed scripts.

### Overall
Strong PR. It replaces the silent-fallback profile map (root cause of the #1985 false-green — arm64 GB300 GKE lanes validated on amd64 H100 nodes) with label-driven, fail-closed selection, a well-reasoned direct-vs-batch asymmetry, and a thorough unit suite. The bash is careful: subshell-exit propagation is handled at every `$(...)` call site, the skippable-`NO_MATCH` vs fatal-`rc=1` split is disciplined, and the `yq`-required gate fails closed without adding an unpinned install.

**@mchmarny's CHANGES_REQUESTED blocker is resolved on this head**, three independent ways:
1. The classify-step **drop filter** removes unmapped recipes before dispatch — no matrix cell is ever created.
2. **Defense-in-depth:** the cell invokes `run-all-recipes.sh --deployer X <recipe>` positionally (`action.yml:178`) → `EXPLICIT_RECIPES` → `run_recipe_test` returns **rc=1**, not a skip.
3. A **regression test** proves an explicitly-named unmapped recipe cannot be reported as passed.

No blockers survived meta-review. Approving; the findings below are Minor/Nitpick. The one worth landing is **F1** — a *new adjacent* fail-open the drop-filter introduces.

### Findings not anchorable to the diff (pre-existing lines)

**🟡 F1 — Empty-tier fail-open: no minimum-coverage floor.** The new drop-filter can empty a whole tier. Tier jobs gate on `tierN_pairs != '[]'`, so an empty tier is **skipped**, and the (pre-existing) `summary` job at `.github/workflows/kwok-recipes.yaml` L497-512 treats a skipped tier as *pass* (it only fails on `failure`/`cancelled`). Because only `kwok/profiles/eks/` ships today, deleting/renaming that directory makes every recipe resolve to the *skippable* `PROFILE_SELECT_RC_NO_MATCH`, empties all three tiers, and the run prints **"All recipe validations passed"** with zero bundle/scheduling coverage. Downgraded to Minor because reaching zero coverage requires removing the whole profile tree (a conspicuous diff); the realistic single-profile case is safe exclusion. **Fix:** floor guard in `discover` — fail when the push/nightly tier3 (or tier1) pair count is 0 (`::error::KWOK matrix is empty — profile tree likely broken`). Only `workflow_dispatch` no-match should legitimately produce no cells.

**🟡 F1b — `summary` needs the tier jobs but not `discover`** (`kwok-recipes.yaml` L473). This PR widens `discover`'s fatal-exit surface; if `discover` fails, the tiers skip and `summary` (`if: always()`) sees them skipped → exit 0. The overall workflow *run* still goes red, so this only bites if **`KWOK Test Summary`** specifically is the required branch-protection check. **Fix:** add `discover` to `summary`'s `needs` and fail when `needs.discover.result != 'success'`.

**🔵 F11 — pre-existing / out of scope:** `kwok/scripts/apply-nodes.sh:152` hardcodes `DEFAULT_REGION`/`DEFAULT_ZONES` to `us-east-1` for every provider. Untouched by this PR and inert today (only eks profiles), but latent once gke/aks profiles land. Follow-up, not a merge concern.

### Confirmed non-issues (examined & cleared)
- ✅ **@mchmarny's blocker** — resolved at both discovery (drop) and in-runner (explicit-fail).
- ✅ **`yq`-required gate** — fails closed, no unpinned install (matches the repo convention that `ubuntu-latest` preinstalls yq/jq).
- ✅ **Silent `eks/p5-h100` fallback** — fully removed from all changed scripts.
- ✅ **`workflow_dispatch`** fails loud on an unmapped recipe; **local `make kwok-test-all`** still skips → no dev-loop regression.

### Summary
`🔴 0 · 🟠 0 · 🟡 3 · 🔵 8` — Approve with comments. The single follow-up worth landing is **F1's coverage-floor guard**; everything else is polish.

<sub>Reviewed with a multi-persona + adversarial-meta-review pass. Inline comments follow.</sub>

PR #2037, opened 2026-08-04
fix: reject recipes where both gpu-operator and gpu-operator-ocp are enabled

### LINKED ISSUE #1685: Reject recipes enabling both GPU Operator variants during allocation-policy resolution
## Problem

`ResolveGPUAllocationPolicy` (`pkg/validator/v1/allocation_policy.go`) resolves `devicePlugin.enabled` from only one GPU operator component. When a recipe enables **both** `gpu-operator` and `gpu-operator-ocp`, it prefers `gpu-operator`, emits a `slog.Warn`, and ignores the OCP component's `devicePlugin.enabled` entirely.

This leaves an orientation gap: a custom recipe with `nvidia-dra-driver-gpu` `resources.gpus.enabled=true` + `gpuResourcesEnabledOverride=true`, `gpu-operator` `devicePlugin.enabled=false`, and `gpu-operator-ocp` `devicePlugin.enabled=true` resolves to `dra-resource-claim` and passes policy resolution — while the OCP ClusterPolicy still renders an enabled device plugin (`recipes/components/gpu-operator-ocp/manifests/clusterpolicy.yaml`), i.e. the dual-advertisement state #1327 rejects.

## Scope

- **Unreachable from the stock catalog**: the OCP overlay swaps components (disables `gpu-operator` and `nvidia-dra-driver-gpu`), and the catalog sweep (`TestStockCatalogResolvesDevicePluginPolicy`) proves no stock overlay enables both. Two GPU operators on one cluster also collide at the operand level regardless of allocation policy.
- Pre-existing #1664 behavior (the warn-and-prefer branch); #1671 did not change it. Surfaced during #1671 cross-review.

## Proposed fix

Reject recipes enabling both GPU Operator variants at allocation-policy resolution time with `ErrCodeInvalidRequest` (fail closed instead of silently preferring one), with table tests covering: standard-only, OCP-only, both (rejected), and neither.

## Related

#1327 (allocation-policy contract), #1664 (introduced the preference branch), #1671 (cross-review origin)

gat786:
## Summary

ResolveGPUAllocationPolicy would previously resolve incorrect policy when both `gpu-operator` and `gpu-operator-ocp` were enabled. Having two gpu operator is also an invalid recipe and should not be enabled at the same time. This PR now rejects when a recipe has both of the operators enabled.

## Motivation / Context

Fixes: #1685 
Related: #1327 #1664 #1671

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Testing

```bash
make qualify
```

## Risk Assessment

**Rollout notes:** 

## Checklist

### DIFF
--- a/pkg/validator/v1/allocation_policy.go
+++ b/pkg/validator/v1/allocation_policy.go
@@ -93,26 +93,40 @@ const (
 //
 // The device-plugin advertiser is read from an enabled gpu-operator OR
 // gpu-operator-ocp componentRef (OCP recipes disable the former and carry the
-// latter); if both are enabled, gpu-operator wins with a warning. When
-// NEITHER is present and enabled, there is no device-plugin advertiser at all
-// — INFERRING an externally managed advertiser from that absence is an
-// explicit #1327 non-goal. An advertiser declared explicitly via
-// metadata.selectedProfile.advertiser: external (ADR-015 GKE amendment) IS
-// supported and handled by the external-advertiser branch below.
+// latter); if both are enabled then ErrCodeInvalidRequest (#1685) — two GPU
+// operators collide at the operand level regardless of allocation policy,
+// and failing closed beats silently preferring one (a divergent
+// devicePlugin.enabled across the two would otherwise slip through).
+//
+// When NEITHER is present and enabled, there is
+// no device-plugin advertiser at all — INFERRING an externally managed
+// advertiser from that absence is an explicit #1327 non-goal. An advertiser
+// declared explicitly via metadata.selectedProfile.advertiser: external
+// (ADR-015 GKE amendment) IS supported and handled by the external-advertiser
+// branch below.
 //
 // Validity gates (ErrCodeInvalidRequest, fail closed at resolution time):
+//
+//   - an ENABLED gpu-operator and an ENABLED gpu-operator-ocp causes
+//     automatic rejection. Both GPU operators enabled: reject (#1685) — two
+//     operators collide at the operand level.
+//
 //   - an ENABLED nvidia-dra-driver-gpu component with resources.gpus.enabled
 //     ABSENT: the chart's declared default (true) would diverge from any
 //     silent resolution — the switch must be explicitly true or false.
+//
 //   - gpus.enabled=true with gpuResourcesEnabledOverride=false: the upstream
 //     chart install guard rejects this combination.
+//
 //   - no whole-GPU advertiser: gpus.enabled explicitly false (or the DRA
 //     component absent/disabled) AND no usable device-plugin advertiser
 //     (devicePlugin.enabled=false, or no enabled GPU operator component).
+//
 //   - gpus.enabled=true with devicePlugin.enabled=true: dual advertisement —
 //     both mechanisms advertising whole GPUs on the same nodes risks GPU
 //     over-admission; exactly one advertiser is required. (Transitional
 //     warning until the production-default flip; an error since.)
+//
 //   - gpus.enabled=false with gpuResourcesEnabledOverride=true: the inert
 //     waiver disarms the chart-guard tripwire that protects the
 //     device-plugin default. (Transitional warning until the
@@ -204,23 +218,23 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult) 
 	// gpu-operator-ocp, and error guidance must name the component actually
 	// carrying devicePlugin.enabled.
 	operatorName := gpuOperatorComponentName
-	// A declared external advertiser takes the dedicated branch below, which
-	// aggregates devicePlugin.enabled across BOTH operator components with OR
-	// semantics — nothing is "ignored" there, so the warn-and-prefer
-	// diagnostic below would be misleading; suppress it on that path.
+	// A declared external advertiser takes the dedicated branch below.
+	// The #1685 rejection above guarantees at most one operator component
+	// is enabled by the time either path reads devicePlugin.enabled.
 	externalAdvertiser := r.Metadata.SelectedProfile != nil &&
 		r.Metadata.SelectedProfile.Advertiser == allocpolicy.AdvertiserExternal
 	opRef := enabledComponentRef(r, gpuOperatorComponentName)
-	if ocpRef := enabledComponentRef(r, gpuOperatorOCPComponentName); ocpRef != nil {
+	ocpRef := enabledComponentRef(r, gpuOperatorOCPComponentName)
+	if ocpRef != nil {
 		if opRef != nil {
-			if !externalAdvertiser {
-				slog.Warn("both gpu-operator and gpu-operator-ocp are enabled in the recipe; resolving devicePlugin.enabled from gpu-operator",
-					"preferred", gpuOperatorComponentName, "ignored", gpuOperatorOCPComponentName)
-			}
-		} else {
-			opRef = ocpRef
-			operatorName = gpuOperatorOCPComponentName
+			// Both GPU operators enabled: reject (#1685) — two operators collide at the operand level.
+			return "", errors.New(errors.ErrCodeInvalidRequest, fmt.Sprintf(
+				"invalid GPU allocation configuration: components %q and %q are both enabled — two GPU operators collide at the operand level; enable exactly one (OpenShift recipes disable %q and carry %q) (issue #1685)",
+				gpuOperatorComponentName, gpuOperatorOCPComponentName,
+				gpuOperatorComponentName, gpuOperatorOCPComponentName))
 		}
+		opRef = ocpRef
+		operatorName = gpuOperatorOCPComponentName
 	}
 	if opRef != nil {
 		values, err := r.GetValuesForComponentWithContext(ctx, opRef.Name)
@@ -244,32 +258,13 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult) 
 	// only metadata.selectedProfile.advertiser produces this branch.
 	if externalAdvertiser {
 		// Under a declared external advertiser EVERY enabled operator
-		// component is a potential second advertiser, so the reading
-		// aggregates across gpu-operator AND gpu-operator-ocp with OR
-		// semantics (mirroring checkAdvertiserCoherence in
-		// pkg/recipe/profile.go) instead of the diagnostic warn-and-prefer
-		// selection above — preferring one component would let the other
-		// component's live plugin slip past the dual-advertisement gate.
-		externalDevicePluginEnabled := false
-		for _, name := range []string{gpuOperatorComponentName, gpuOperatorOCPComponentName} {
-			ref := enabledComponentRef(r, name)
-			if ref == nil {
-				continue
-			}
-			values, err := r.GetValuesForComponentWithContext(ctx, ref.Name)
-			if err != nil {
-				return "", err
-			}
-			// Key absent → true: the upstream chart default, same as the
-			// preferred-component reading above.
-			enabled, err := lookupBoolValue(values, ref.Name, valuePathDevicePluginEnabled, true)
-			if err != nil {
-				return "", err
-			}
-			if enabled {
-				externalDevicePluginEnabled = true
-			}
-		}
+		// component is a potential second advertiser, so the reading is
+		// the devicePlugin.enabled of any enabled operator component. The
+		// #1685 reject above guarantees at most one operator component is
+		// enabled here, so devicePluginEnabled — read from the surviving
+		// opRef above — already carries the aggregation the pre-#1685
+		// code performed across both components.
+		externalDevicePluginEnabled := devicePluginEnabled
 
 		observation := allocpolicy.Observation{
 			Advertiser:          allocpolicy.AdvertiserExternal,
@@ -288,9 +283,11 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult) 
 	// Verdicts (chart guard, dual advertisement, no advertiser, inert
 	// waiver) come from the single shared #1327 tuple evaluator, so the
 	// hydrating artifact gate (pkg/recipe checkAdvertiserCoherence) and
-	// this resolver cannot drift: an artifact the gate emits is exactly an
-	// artifact this resolver accepts (ADR-015 gate/resolver symmetry).
-	// Only the policy SELECTION below stays here.
+	// this resolver apply the same #1327 tuple verdicts — an artifact the
+	// gate emits passes the same tuple rows this resolver accepts
+	// (ADR-015 gate/resolver symmetry over the shared tuple rows; the
+	// #1685 dual-operator rejection above is resolver-side only and not
+	// mirrored at the gate). Only the policy SELECTION below stays here.
 	observation := allocpolicy.Observation{
 		DevicePluginEnabled:  &devicePluginEnabled,
 		GPUOperatorComponent: operatorName,
--- a/pkg/validator/v1/allocation_policy_test.go
+++ b/pkg/validator/v1/allocation_policy_test.go
@@ -243,8 +243,12 @@ func TestResolveGPUAllocationPolicy(t *testing.T) {
 			want: GPUAllocationPolicyDevicePluginExtendedResource,
 		},
 		{
-			// Both operator components enabled: gpu-operator wins (warn).
-			name: "both gpu-operator and gpu-operator-ocp enabled: gpu-operator resolves the advertiser",
+			// Two GPU operators on one cluster collide at the operand level
+			// regardless of devicePlugin enablement at any operator level or external
+			// advertisers presence, so enabling both fails closed.
+			// (ErrCodeInvalidRequest) instead of silently preferring one (#1685
+			// and #1327).
+			name: "both gpu-operator and gpu-operator-ocp enabled: invalid (rejected)",
 			recipe: &recipe.RecipeResult{ComponentRefs: []recipe.ComponentRef{
 				gpuOperatorRef(true),
 				{
@@ -254,7 +258,31 @@ func TestResolveGPUAllocationPolicy(t *testing.T) {
 					},
 				},
 			}},
-			want: GPUAllocationPolicyDevicePluginExtendedResource,
+			wantErr: true,
+			wantMsg: "invalid GPU allocation configuration: components \"gpu-operator\" and \"gpu-operator-ocp\" are both enabled",
+		},
+		{
+			// Regression for the #1685 orientation gap: a custom recipe with
+			// gpu-operator devicePlugin.enabled=false and gpu-operator-ocp
+			// devicePlugin.enabled=true previously resolved dra-resource-claim
+			// (gpu-operator won) while the OCP ClusterPolicy still rendered an
+			// enabled device plugin — the dual-advertisement state #1327
+			// rejects. Enabling both operators must fail closed before any
+			// policy is derived, regardless of which side carries the
+			// divergent pin.
+			name: "both operators enabled with divergent devicePlugin pins: invalid (orientation gap)",
+			recipe: &recipe.RecipeResult{ComponentRefs: []recipe.ComponentRef{
+				draDriverRef(true, true),
+				gpuOperatorRef(false),
+				{
+					Name: "gpu-operator-ocp",
+					Overrides: map[string]any{
+						"devicePlugin": map[string]any{"enabled": true},
+					},
+				},
+			}},
+			wantErr: true,
+			wantMsg: "invalid GPU allocation configuration: components \"gpu-operator\" and \"gpu-operator-ocp\" are both enabled",
 		},
 		{
 			// OCP recipes carry the DRA driver under its -ocp alias; the
@@ -556,13 +584,12 @@ func TestResolveGPUAllocationPolicyExternalAdvertiser(t *testing.T) {
 			wantMsg: "devicePlugin.enabled=true",
 		},
 		{
-			// Under a declared external advertiser the reading aggregates
-			// across EVERY enabled operator component (OR semantics,
-			// mirroring the recipe-side checkAdvertiserCoherence): the
-			// warn-and-prefer selection used for diagnostics elsewhere must
-			// not let gpu-operator-ocp's live plugin slip past the gate
-			// behind a disabled gpu-operator.
-			name: "external advertiser with gpu-operator plugin off but gpu-operator-ocp plugin on is dual advertisement",
+			// Two GPU operators on one cluster collide at the operand level
+			// regardless of devicePlugin enablement on any operator level or external
+			// advertisers presence, so enabling both fails closed.
+			// (ErrCodeInvalidRequest) instead of silently preferring one (#1685
+			// and #1327).
+			name: "external advertiser with gpu-operator plugin off but gpu-operator-ocp plugin on is dual advertisement (rejected)",
 			recipe: externalProfileRecipe(
 				draDriverRef(false, false),
 				gpuOperatorRef(false),
@@ -574,13 +601,15 @@ func TestResolveGPUAllocationPolicyExternalAdvertiser(t *testing.T) {
 				},
 			),
 			wantErr: true,
-			wantMsg: "devicePlugin.enabled=true",
+			wantMsg: "invalid GPU allocation configuration: components \"gpu-operator\" and \"gpu-operator-ocp\" are both enabled",
 		},
 		{
-			// Both operator components enabled with the plugin off on each:
-			// the OR-aggregate stays false and external remains the sole
-			// advertiser.
-			name: "external advertiser with both operator components' plugins off resolves",
+			// Two GPU operators on one cluster collide at the operand level
+			// regardless of devicePlugin enablement on any operator level or external
+			// advertisers presence, so enabling both fails closed.
+			// (ErrCodeInvalidRequest) instead of silently preferring one (#1685
+			// and #1327).
+			name: "external advertiser with both operator components' plugins off (rejected)",
 			recipe: externalProfileRecipe(
 				draDriverRef(false, false),
 				gpuOperatorRef(false),
@@ -591,19 +620,23 @@ func TestResolveGPUAllocationPolicyExternalAdvertiser(t *testing.T) {
 					},
 				},
 			),
-			want: GPUAllocationPolicyDevicePluginExtendedResource,
+			wantErr: true,
+			wantMsg: "invalid GPU allocation configuration: components \"gpu-operator\" and \"gpu-operator-ocp\" are both enabled",
 		},
 		{
-			// An enabled operator component with the key ABSENT follows the
-			// upstream chart default (true) in the aggregate too.
-			name: "external advertiser with gpu-operator-ocp missing the devicePlugin key is dual advertisement",
+			// Two GPU operators on one cluster collide at the operand level
+			// regardless of devicePlugin enablement on any operator level or external
+			// advertisers presence, so enabling both fails closed.
+			// (ErrCodeInvalidRequest) instead of silently preferring one (#1685
+			// and #1327).
+			name: "external advertiser with gpu-operator-ocp missing the devicePlugin key is dual advertisement (rejected)",
 			recipe: externalProfileRecipe(
 				draDriverRef(false, false),
 				gpuOperatorRef(false),
 				recipe.ComponentRef{Name: "gpu-operator-ocp"},
 			),
 			wantErr: true,
-			wantMsg: "devicePlugin.enabled=true",
+			wantMsg: "invalid GPU allocation configuration: components \"gpu-operator\" and \"gpu-operator-ocp\" are both enabled",
 		},
 		{
 			name:    "external advertiser with DRA gpus enabled is dual advertisement",
--- a/docs/design/015-recipe-configuration-profiles.md
+++ b/docs/design/015-recipe-configuration-profiles.md
@@ -890,6 +890,15 @@ any check runs. This ADR therefore **amends the #1327 model**:
   `devicePlugin.enabled=false`, preserving #1327's fail-closed posture for
   recipes without the declaration (the non-goal narrows from "externally
   managed advertisers" to "*inferring* externally managed advertisers").
+- The device-plugin advertiser selection itself was tightened (#1685): a
+  recipe with both `gpu-operator` and `gpu-operator-ocp` enabled is
+  rejected (`ErrCodeInvalidRequest`) at resolution — profile or not,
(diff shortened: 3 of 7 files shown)


mchmarny (dismissed):
APPROVE — reviewed the fail-closed GPU allocation-policy change at b5a5518; 0 blockers. Focused race tests and package lint pass. Merge Gate remains queued behind contributor validation.

pkg/validator/v1/allocation_policy.go
```diff
@@ -190,14 +192,18 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult)
...
-			slog.Warn("both gpu-operator and gpu-operator-ocp are enabled in the recipe; resolving devicePlugin.enabled from gpu-operator",
-				"preferred", gpuOperatorComponentName, "ignored", gpuOperatorOCPComponentName)
-		} else {
-			opRef = ocpRef
-			operatorName = gpuOperatorOCPComponentName
+			// both gpu-operator and gpu-operator-ocp are enabled, reject during resolve GPUAllocationPolicy
+			// #1685
+			return "", errors.New(
+				errors.ErrCodeInvalidRequest,
+				"both gpu-operator and gpu-operator-ocp are enabled in the recipe; cannot resolve GPUAllocationPolicy in this state",
```
njhensley: **🟡 Minor — error message breaks this function's contract**

Every other `ErrCodeInvalidRequest` in `ResolveGPUAllocationPolicy` (lines 172-174, 223-234, 240-244, 248-253) opens with `invalid GPU allocation configuration:` and closes with concrete remediation + `(issue #1327)`. This new message has none of the three: no fix guidance (the fix is "disable one; OCP recipes disable gpu-operator"), no issue ref in the text (#1685 is comment-only), and it leaks the internal symbol `GPUAllocationPolicy` into user-facing text that reaches the CLI and HTTP 400 `details["error"]`.

**Fix:** Match the siblings, e.g.:

```go
return "", errors.New(errors.ErrCodeInvalidRequest, fmt.Sprintf(
    "invalid GPU allocation configuration: components %q and %q are both enabled — two GPU operators collide at the operand level; enable exactly one (OpenShift recipes disable %q and carry %q) (issue #1685)",
    gpuOperatorComponentName, gpuOperatorOCPComponentName,
    gpuOperatorComponentName, gpuOperatorOCPComponentName))
```

The existing `wantMsg` substring (`both gpu-operator and gpu-operator-ocp are enabled`) survives that rewrite.

pkg/validator/v1/allocation_policy.go
```diff
@@ -190,14 +192,18 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult)
...
-	if ocpRef := enabledComponentRef(r, gpuOperatorOCPComponentName); ocpRef != nil {
+	ocpRef := enabledComponentRef(r, gpuOperatorOCPComponentName)
+	if ocpRef != nil {
 		if opRef != nil {
-			slog.Warn("both gpu-operator and gpu-operator-ocp are enabled in the recipe; resolving devicePlugin.enabled from gpu-operator",
-				"preferred", gpuOperatorComponentName, "ignored", gpuOperatorOCPComponentName)
-		} else {
-			opRef = ocpRef
-			operatorName = gpuOperatorOCPComponentName
+			// both gpu-operator and gpu-operator-ocp are enabled, reject during resolve GPUAllocationPolicy
```
njhensley: **🔵 Nitpick — comment grammar**

`// both gpu-operator and gpu-operator-ocp are enabled, reject during resolve GPUAllocationPolicy` is a comma splice with an awkward "reject during resolve".

**Fix:** `// Both GPU operators enabled: reject (#1685) — two operators collide at the operand level.`

njhensley (dismissed):
## Multi-persona review — Approve with comments

**Method:** 3 independent persona reviewers (Correctness · Domain/Architecture · Docs+Test-coverage) ran in parallel, then every candidate finding was re-derived from the resolved code by an adversarial senior meta-reviewer (CONFIRMED / REFUTED / RETIERED). Line links pin to head `b5a5518a`.

> This review concurs with @mchmarny's prior approval and adds detail — no finding contradicts that decision.

### Assessment
Correct, tightly-scoped fix. The both-operators case now fails closed (`ErrCodeInvalidRequest`) instead of `slog.Warn` + silently preferring `gpu-operator` — squarely consistent with the #1327 fail-closed philosophy already governing this function. The meta-reviewer independently confirmed: control flow is equivalent for every non-both case, the rejection doesn't over-fire on disabled operators (test `allocation_policy_test.go:207-227`), the DRA gate that runs first can't create a fail-open path (both return `ErrCodeInvalidRequest`), and the removed `slog` import leaves no dangling reference. The divergent-pin "orientation gap" test is a genuine regression test that would have caught the old bug. Blast radius is contained — `recipes/overlays/ocp.yaml` disables `gpu-operator`, so no stock recipe trips this; only hand-authored custom/SDK recipes reach it (breaking change acknowledged).

### Additional findings (files not in this diff)

🟡 **Minor — doc catalog omits the new rejection** · `docs/user/validation.md:553-569`
This section enumerates every resolution-time `ErrCodeInvalidRequest` state (it lists five) and even names `gpu-operator-ocp` a line above — but the new both-operators rejection isn't added. The PR checklist ticks "docs updated"; per `CLAUDE.md` this user-visible behavior change should amend the doc in the same PR. Change "Two further states are likewise rejected" → "Three…" and add the both-operators clause.

🔵 **Informational (follow-up, not this PR)** · `allocation_policy.go:196-207`
The collision is a policy-independent recipe-validity invariant but is enforced only on the validate/bundle path (+`--cncf-submission`), not at `aicr recipe` generation — so an invalid recipe can still be generated and only fails later. Pre-existing architecture (every gate here has the same reach), not a regression this PR introduces. Possible future work: lift the mutual-exclusion into recipe resolution so generation fails closed too.

🔵 **Optional test** · `allocation_policy_test.go`
The non-over-fire negative case only covers *gpu-operator disabled + ocp enabled*; a symmetric *gpu-operator enabled + ocp present-but-disabled* case would close the loop. `enabledComponentRef` is symmetric so risk is near-zero — drop if not worth it.

### Confirmed non-issues (examined & cleared)
Control-flow equivalence · no over-fire on disabled operators · DRA-gate evaluation order (no fail-open) · clean `slog` removal · correct `ErrCodeInvalidRequest`→HTTP 400 mapping · ADRs 014/015 not made stale (old "prefer gpu-operator" was Go-only, never documented).

### Tally
🔴 Blocker 0 | 🟠 Major 0 | 🟡 Minor 2 | 🔵 Nitpick 3 — **Approve with comments.**

gat786:
Lemme recheck the nitpicks and minor comments and fix them.

pkg/validator/v1/allocation_policy.go
gat786: Thanks for the detailed fix along with whats going wrong message as well.

pkg/validator/v1/allocation_policy.go
gat786: Again thanks for the fix as well, I have patched.

mchmarny:
Looking at that conflict, this will require manual resolution. Do not select “Accept both.” PR head is 6889be5; current main is 01932ea.
  
  - Preserve current main’s new allocpolicy and external-advertiser handling.
  - Replace its warn-and-prefer block with the PR’s fail-closed dual-operator rejection.
  - Remove the now-unused log/slog import.
  - Update the auto-merged external-advertiser test that currently allows both operators; it must now expect rejection.
  - Fix the PR test/error mismatch: the test expects both gpu-operator and gpu-operator-ocp are enabled, but the implementation currently emits quoted component names in a different
    order.

gat786:
fixing the things, on it.

gat786:
@mchmarny 

As far as I understand, when both operators are enabled, we should reject the recipe because they collide at operand level. I believe this introduction of externalAdvertisor is making me confused a bit. Since there are 3 levels of conditions now that our code will have to check. Previously we would reject if our code reached the second level i.e. if both operators were enabled. Now since we add externalAdvertisors as the final condition check, if it is not enabled and we have both operators enabled we reject the recipe.

what happens in the case where there is an externalAdvertisor along with both the operators enabled?

mchmarny:
Best way to think of these as two independent invariants:

1. If both GPU Operator variants are enabled, reject immediately because their operands collide, regardless of the advertiser or `devicePlugin.enabled` values (#1685).

2. Then, only after that check passes, evaluate advertiser coherence. With an external advertiser, any remaining operator device plugin and DRA whole-GPU advertisement must be disabled (#1327).

Therefore, external advertiser + both operators is still rejected. Make the dual-operator check unconditional and place it before the external-advertiser branch. Any external-advertiser tests containing both enabled operators should expect that same dual-operator error.

@yuanchen8911 did I get this right?

gat786:
thanks for the description.

pkg/validator/v1/allocation_policy_test.go
```diff
@@ -245,7 +248,31 @@ func TestResolveGPUAllocationPolicy(t *testing.T) {
 					},
 				},
 			}},
-			want: GPUAllocationPolicyDevicePluginExtendedResource,
+			wantErr: true,
+			wantMsg: "both gpu-operator and gpu-operator-ocp are enabled",
```
mchmarny: **Blocking — this test fails at the current head.**

`wantMsg` is matched as a substring, and this string does not appear in the error the implementation emits. The message is `components "gpu-operator" and "gpu-operator-ocp" are both enabled — …` (quoted names, different word order), so the assertion can never match:

```
allocation_policy_test.go:379: error "[INVALID_REQUEST] invalid GPU allocation configuration:
components \"gpu-operator\" and \"gpu-operator-ocp\" are both enabled — two GPU operators collide
at the operand level; enable exactly one …"
does not contain "both gpu-operator and gpu-operator-ocp are enabled"
```

Same problem on line 275 for the new orientation-gap case — one fix covers both.

You could not have seen this from the PR page, and that is not your fault: the `tests / Test` job never ran on this PR (details in the review summary), so the checks went green without the unit suite executing.

**Minimum correction:** pick one side and align the other. Either match a substring that actually occurs (e.g. `are both enabled`), or simplify the production message. Verify locally with:

```bash
go test ./pkg/validator/v1/ -run TestResolveGPUAllocationPolicy
```

pkg/validator/v1/allocation_policy.go
```diff
@@ -202,16 +216,19 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult)
 	externalAdvertiser := r.Metadata.SelectedProfile != nil &&
 		r.Metadata.SelectedProfile.Advertiser == allocpolicy.AdvertiserExternal
 	opRef := enabledComponentRef(r, gpuOperatorComponentName)
-	if ocpRef := enabledComponentRef(r, gpuOperatorOCPComponentName); ocpRef != nil {
+	ocpRef := enabledComponentRef(r, gpuOperatorOCPComponentName)
+	if ocpRef != nil {
 		if opRef != nil {
 			if !externalAdvertiser {
```
mchmarny: **Blocking — the dual-operator check needs to be unconditional.**

Guarding the rejection behind `!externalAdvertiser` means a recipe that declares an external advertiser **and** enables both GPU operators is still accepted. That is the one case invariant #1 is meant to catch: the two operators' operands collide on the cluster whether or not a third-party advertiser is declared, so the advertiser value cannot make the collision safe.

Concretely, `pkg/validator/v1/allocation_policy_test.go` line 557 (`external advertiser with both operator components' plugins off resolves`) has both operators enabled and asserts success — it passes today only because of this condition. Under the invariant it should expect the dual-operator error instead.

**Minimum correction:** drop the `if !externalAdvertiser` wrapper so the check is `if opRef != nil && ocpRef != nil { return error }`, and hoist it above the external-advertiser branch so it runs before any advertiser reasoning. Then update that test case (and any other external-advertiser case with both operators enabled) to expect the rejection.

To restate the ordering from the thread above, since it is the part that is easy to lose:

1. Both operator variants enabled → reject immediately, regardless of advertiser or `devicePlugin.enabled`.
2. Only once that passes, evaluate advertiser coherence.

Worth saying: the rest of this function reads correctly. Hoisting `opRef = ocpRef` out of the `else` is safe — I traced it, and the external-advertiser branch returns early using its own OR-aggregated reading, so it never consumes the value.

README.md
```diff
@@ -4,6 +4,8 @@
 [![On Tag Release](https://github.com/NVIDIA/aicr/actions/workflows/on-tag.yaml/badge.svg)](https://github.com/NVIDIA/aicr/actions/workflows/on-tag.yaml)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](LICENSE)
 
+This is a dummy commit
```
mchmarny: **Blocking — scratch text on the project front page.**

This line arrived with the head commit `af935e27 feat: dummy` and would land directly under the badges on the repo landing page and on docs.nvidia.com/aicr.

Easy to leave behind by accident when testing a push, so no harm done — it just needs to come out before merge, along with the commit itself so it does not sit in `main`'s history:

```bash
git rebase -i origin/main   # drop the "feat: dummy" commit
git push --force-with-lease origin fix/gpu-allocation-policy-rejection
```

If you would rather not rebase interactively, `git reset --soft HEAD~1` then re-commit only the validator changes works too.

pkg/validator/v1/allocation_policy.go
```diff
@@ -92,26 +91,41 @@ const (
...
 // The device-plugin advertiser is read from an enabled gpu-operator OR
 // gpu-operator-ocp componentRef (OCP recipes disable the former and carry the
-// latter); if both are enabled, gpu-operator wins with a warning. When
+// latter); if both are enabled, and there is no externalAdvertiser then
+// ErrCodeInvalidRequest (#1685) — two GPU operators collide at the operand
+// level regardless of allocation policy, and failing closed beats silently
+// preferring one (a divergent devicePlugin.enabled across the two would
+// otherwise slip through). When NEITHER is present and enabled, there is
+// no device-plugin advertiser at all — externally managed
+// advertisers are an explicit #1327 non-goal. When
```
mchmarny: **Non-blocking** — the sentence *"When NEITHER is present and enabled, there is no device-plugin advertiser at all"* now appears twice, once here and again on the next line.

The duplicate also softens the original point. It used to say that **inferring** an external advertiser from that absence is the #1327 non-goal; the copy here reads as though external advertisers are the non-goal, which is a different (and wrong) claim now that the external-advertiser branch exists.

Suggest deleting the first copy on lines 98–100 and keeping the original wording that follows. Since this block is the reference other contributors read before touching allocation policy, it is worth having exactly one statement of each rule.

mchmarny (changes requested):
Thanks for sticking with this one, @gat786 — and for working through the invariant discussion above. The core change is the right shape: rejecting the dual-operator case instead of silently preferring one operator is exactly what #1685 needs, and the orientation-gap test you added is a genuinely good catch that goes beyond the original report.

**Requesting changes: 4 blockers**, reviewed at head `af935e27`. Three are quick; one is a real semantic gap. Details are inline.

### Please read this part first — the green checks are misleading

The unit tests **fail at this head**, and the PR page does not show it. `tests / Test`, `tests / E2E`, and `analyze` have no result at all — they never ran, because `copy-pr-bot` gates workflows on NVIDIA runners until a vetter approves the run for an external contributor. Every green check you see is either a `-skip` sentinel or Lint / Security-Scan / CLI-E2E.

So please do not read green as passing here. Until the runners are unblocked, `go test ./pkg/validator/v1/...` locally is your source of truth. I ran it against your head and both new subtests fail on an assertion mismatch (first inline comment).

### The four blockers

1. **Failing tests** — `wantMsg` never matches the emitted error, on both new cases.
2. **Dual-operator check is still conditional** — external advertiser + both operators is still accepted, which is the case invariant #1 exists to catch.
3. **`feat: dummy` commit** — adds scratch text to `README.md`.
4. *(non-blocking)* A duplicated sentence in the doc block.

### Commit signing

Both commits are currently unsigned:

```
6fc36901  verified=false  reason=unsigned
af935e27  verified=false  reason=unsigned
```

To be clear about what is and is not missing: your **DCO sign-off is correct** — both commits carry a proper `Signed-off-by` line, so that half is already done. What is missing is the `-S` cryptographic signature. The two are independent, and AICR requires both on every commit; a branch ruleset rejects unsigned commits at push time, so this will block the merge regardless of the code review.

- One-time setup (git config + registering a key): [CONTRIBUTING.md → Developer Certificate of Origin](https://github.com/NVIDIA/aicr/blob/main/CONTRIBUTING.md#one-time-setup)
- Fixing commits you have already pushed: [CONTRIBUTING.md → Forgot to Sign or Sign Off?](https://github.com/NVIDIA/aicr/blob/main/CONTRIBUTING.md#forgot-to-sign-or-sign-off)
- Generating a GPG or SSH signing key and adding it to your account: [GitHub docs](https://docs.github.com/en/authentication/managing-commit-signature-verification)

Since you will be rebasing anyway to drop the dummy commit, this re-signs the whole branch in one pass:

```bash
git rebase --exec 'git commit --amend -s -S --no-edit' origin/main
git push --force-with-lease origin fix/gpu-allocation-policy-rejection
```

Setting `git config commit.gpgsign true` once means you only need `-s` from then on.

### Housekeeping, not your doing

- **Prior approvals no longer apply.** Mine and @njhensley's were pinned to `b5a5518` and were dismissed automatically when the head moved. They will need refreshing after the fixes — that is normal, not a withdrawal of support.
- **The `needs-rebase` label looks stale.** `git merge-tree` against current `main` reports a clean merge, so the earlier conflict appears already resolved. Ignore the bot's rebase nag unless it recurs; I will get the label sorted.

### Suggested order

Fix the assertion (1) and make the check unconditional (2) together, since (2) changes which cases expect errors. Then drop the dummy commit and re-sign the branch in the same rebase. Run `make test` before pushing — with the runners gated, local is the only signal you will get.

Happy to look again as soon as you push, and glad to pair on the invariant ordering if any of it is still ambiguous.

README.md
gat786: Hey Mark, sorry my work machine is misbehaving and not signing things properly so I made this commit for just testing that, my bad, i will reset according to your message.

gat786: I removed, fixing the others.

gat786: suddenly out of nowhere my signingkeys were not being correct forwarded to the dev-machine and hence this messup.

pkg/validator/v1/allocation_policy_test.go
gat786: Thanks for explanation on through out this PR, Along with your explanations @njhensley ofcourse you as well, I even asked an LLM about things i did not understand and was able to complete the things. It is indeed helpful at times.

pkg/validator/v1/allocation_policy.go
gat786: yeah I came across this practice when I contributed to an another repo, there they followed something like `effective go` practices. what it means if you have a `return` in an `if` case, you can drop the `else` conditions and it should work just fine.

pkg/validator/v1/allocation_policy.go
gat786: I removed the duplicated lines.

gat786:
again an issue with signing.

gat786:
I will make sure that commits are signed before pushing here and causing unnecessary emails and notifs.

pkg/validator/v1/allocation_policy.go
```diff
@@ -202,16 +215,17 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult)
...
+	if ocpRef != nil {
 		if opRef != nil {
-			if !externalAdvertiser {
-				slog.Warn("both gpu-operator and gpu-operator-ocp are enabled in the recipe; resolving devicePlugin.enabled from gpu-operator",
-					"preferred", gpuOperatorComponentName, "ignored", gpuOperatorOCPComponentName)
-			}
-		} else {
-			opRef = ocpRef
-			operatorName = gpuOperatorOCPComponentName
+			// Both GPU operators enabled: reject (#1685) — two operators collide at the operand level.
```
njhensley, line 230: **🔵 — Record this breaking change in ADR-015 / PR notes**

This rejection now fires even under a declared external advertiser, removing the previously-resolvable 'external advertiser + both operator components enabled' shape (ADR-015 aggregation). It's a defensible call — two GPU operators collide at the operand level regardless of the advertiser, and no shipping recipe pairs gpu-operator-ocp with a GKE external advertiser.

**Blast radius:** No known real recipe hits this combination; the three external-advertiser tests were updated to expect the rejection.

**Fix:** Note the behavior removal in the ADR-015 GKE amendment and/or the PR's breaking-change notes so the intent is captured where the next reader looks. Related: the external-advertiser aggregation loop below (258-277) is now vestigial since both-enabled can't reach it.

njhensley (dismissed):
## Multi-persona review (round 2) — Approve with comments

**Method:** independent persona reviewers (Correctness · Domain/Architecture · Docs+Test-coverage) scoped to the delta, then every finding re-derived from the resolved code by an adversarial senior meta-reviewer (CONFIRMED / REFUTED / RETIERED). Verified at head `45d3c7c2`: `go test -race ./pkg/validator/v1/` = **ok**, `golangci-lint` = **0 issues**. Line links pin to head `45d3c7c2`.

> Concurs with @mchmarny's approval. This round confirms the two blockers from my earlier pass are resolved.

### Assessment
Both prior blockers are cleared. (1) The new tests' `wantMsg` substrings now match the reworded error, and (2) the blanket both-operators rejection — which had broken three pre-existing `TestResolveGPUAllocationPolicyExternalAdvertiser` cases — is now a **deliberate** design decision, with those tests flipped to expect the rejection. That's a coherent call: two GPU operators collide at the operand level (two ClusterPolicies / two operand sets) regardless of who advertises `nvidia.com/gpu`, and no shipping recipe pairs the OpenShift-only `gpu-operator-ocp` with a GKE external advertiser, so nothing real regresses. The package passes tests and lint at head. Remaining items are all non-blocking quality/doc cleanups.

### Findings

🟡 **Minor — stale "warn-and-prefer" inline comments** · [`allocation_policy.go:211-214`](https://github.com/NVIDIA/aicr/blob/45d3c7c2b99e0e5387e00b3ca992570d6d105394/pkg/validator/v1/allocation_policy.go#L211) & [`:251-257`](https://github.com/NVIDIA/aicr/blob/45d3c7c2b99e0e5387e00b3ca992570d6d105394/pkg/validator/v1/allocation_policy.go#L251)
The `chore: comments fixed` commit reconciled the **godoc** (91-105) but missed these two. 211-214 still says `externalAdvertiser` exists so "the warn-and-prefer diagnostic below would be misleading; suppress it on that path" — there's no warn to suppress and the rejection fires regardless of `externalAdvertiser`. 251-257 still contrasts the loop with "the diagnostic warn-and-prefer selection above," which no longer exists. Reword both to match the current reject-or-select flow.

🟡 **Minor — external-advertiser aggregation loop is now vestigial** · [`allocation_policy.go:258-277`](https://github.com/NVIDIA/aicr/blob/45d3c7c2b99e0e5387e00b3ca992570d6d105394/pkg/validator/v1/allocation_policy.go#L258)
Because both-enabled is rejected at line 220, at most one operator is enabled when this branch runs — so the "OR across `gpu-operator` AND `gpu-operator-ocp`" loop can only ever see one, recomputing the value already in `devicePluginEnabled` (line 237) with a redundant second hydration of that component. Not a bug, but the loop's stated purpose ("preferring one would let the other slip past") is now unreachable. Reuse `devicePluginEnabled`, or annotate that both-enabled can't reach here.

🟡 **Minor — doc catalog still omits the rejection** · `docs/user/validation.md` (resolution-time rejection catalog)
Still enumerates five `ErrCodeInvalidRequest` states with no both-operators entry. Per `CLAUDE.md`, a user-visible rejection state should be documented in the same PR.

🔵 **Nitpick — record the breaking change** · [`allocation_policy.go:220`](https://github.com/NVIDIA/aicr/blob/45d3c7c2b99e0e5387e00b3ca992570d6d105394/pkg/validator/v1/allocation_policy.go#L220)
The rejection removes the previously-resolvable "external advertiser + both operators" shape (ADR-015 aggregation). Worth a line in the ADR-015 GKE amendment / PR breaking-change notes so the intent is captured for the next reader.

### Confirmed non-issues (examined & cleared)
Function godoc (91-105) now accurate · error message matches its `ErrCodeInvalidRequest` siblings · exactly-one-operator no-over-fire cases pass · `externalAdvertiser` not dead (used at line 250) · package passes `go test -race` and `golangci-lint`.

### Tally
🔴 Blocker 0 | 🟠 Major 0 | 🟡 Minor 3 | 🔵 Nitpick 1 — **Approve with comments.**

pkg/validator/v1/allocation_policy.go
```diff
@@ -202,16 +215,17 @@ func ResolveGPUAllocationPolicy(parent context.Context, r *recipe.RecipeResult)
 	externalAdvertiser := r.Metadata.SelectedProfile != nil &&
```
yuanchen8911, line 224: The comment block just above (L211-214) still says the warn-and-prefer diagnostic "below" would be misleading and is suppressed on the external path. The `slog.Warn` it refers to was removed by this PR, and the rejection is unconditional, so nothing is suppressed anymore. Suggest rewriting to state only why `externalAdvertiser` is computed here (it selects the dedicated aggregation branch below).

yuanchen8911 (commented):
The fix itself looks right. The rejection uses `ErrCodeInvalidRequest` via `pkg/errors`, fires before any policy derivation, and the new orientation-gap regression test covers exactly the divergent devicePlugin-pin state from #1685. It is enforced on both the CLI validation path and the SDK path (`Client.ValidateState` via `ToValidationInputWithContext`). Targeted race tests on the package pass, and no stock recipe or overlay enables both operators, so supported recipes are unaffected.

The remaining items are all comment/doc drift created by this change — no logic changes needed. Two of them concur with findings already posted in earlier reviews; the last two are new.

1. `docs/user/validation.md` (L553-566): the "Configured GPU allocation policy" section enumerates the five configurations rejected at resolution time. This PR adds a sixth (both `gpu-operator` and `gpu-operator-ocp` enabled), which is a user-visible breaking change for custom/SDK recipes, so it needs a sentence in that list in this PR. The checklist currently has "I updated docs" checked, but the diff contains no docs change.

2. `pkg/validator/v1/allocation_policy.go` (L251-257): the comment inside the external-advertiser branch still contrasts the loop with "the diagnostic warn-and-prefer selection above", which this PR removed. And with the both-enabled case now rejected before this branch, at most one enabled operator can ever reach it, so the OR-aggregation across both operator names (L258-277) is degenerate — harmless, but the comment should say the dual-operator gate above guarantees at most one enabled operator reaches advertiser coherence. (Same theme as the inline comment below on L211-214.)

3. `pkg/recipe/profile.go` (L818-821): the gate-side comment still says "under an empty advertiser the resolver reads the preferred component only (gpu-operator wins over gpu-operator-ocp when both are enabled)". This PR removed that resolver behavior — the resolver now rejects instead of preferring — so the comment describes semantics that no longer exist.

4. `pkg/recipe/profile.go` (L793-805) and `pkg/allocpolicy/allocpolicy.go` (L158-187): the absolute symmetry prose ("an artifact this gate emits is exactly an artifact validation accepts") is no longer true for the dual-operator case — the gate still accepts a both-enabled recipe that the resolver now rejects. Gate-side enforcement is outside #1685's explicit scope (the issue scopes the fix to allocation-policy resolution time), so the right fix here is to narrow that prose to the shared #1327 tuple verdicts only, not to expand the gate or `CheckCoherence` in this PR. If bundle-time rejection is wanted later, that is a separate focused follow-up covering raw, REST, SDK, and direct-bundler recipes.

Process: please squash the two commits into one before the next push (both are already signed and DCO'd).

pkg/validator/v1/allocation_policy.go
gat786: Thanks for the detailed review, I have tried to tackle all of the points mentioned and i think we are now closer to completing the issue goals.

pkg/validator/v1/allocation_policy.go
gat786: Thanks for the review, I went ahead and did fix all of the issues that you have mentioned, they all make sense to me. I think this should do it.

yuanchen8911: Reopening this one — the block at L211-214 is unchanged at 39ae66c68 and is now stale twice over: the warn-and-prefer diagnostic it says to suppress was removed by this PR, and after your rework the external-advertiser branch no longer aggregates across both operator components either (it reuses the single reading from the surviving opRef — nice simplification, by the way). Suggest replacing the block with something like:

```go
// A declared external advertiser takes the dedicated branch below.
// The #1685 rejection above guarantees at most one operator component
// is enabled by the time either path reads devicePlugin.enabled.
```

Everything else from my review is addressed — this comment block and the validation.md entry (separate comment) are the only two items left.

yuanchen8911:
One doc item from my review is still open at 39ae66c68: `docs/user/validation.md`'s "Configured GPU allocation policy" section still enumerates five rejected configurations and doesn't mention the new dual-operator rejection. Since this is a user-visible breaking change for custom/SDK recipes, it needs a sentence in that list in this PR (the checklist has "I updated docs" checked). Something like:

> A recipe enabling both `gpu-operator` and `gpu-operator-ocp` is also rejected at resolution time — two GPU operators collide at the operand level; enable exactly one (OpenShift recipes disable `gpu-operator` and carry `gpu-operator-ocp`).

The ADR-015 note you added covers the design record; this is just the user-facing half.

gat786:
I think, I ticked all the boxes, also rebased so that commit history looks cleaner.

mchmarny:
Remaining blocker at `ae1166e8`:

- [`pkg/validator/v1/allocation_policy.go:173`](https://github.com/NVIDIA/aicr/blob/ae1166e81b9bad971dd828a3a2c8301ecb961cd5/pkg/validator/v1/allocation_policy.go#L173) still calls `slog.Warn`, but this PR removes the `log/slog` import. The package does not compile (`undefined: slog`), so Merge Gate and KWOK fail before exercising the change. Restore the import (or refactor the remaining call), then rerun the required checks.

I reproduced this on the exact head. With only the import restored in an isolated clone, the affected race tests pass and affected-package lint reports 0 issues; I found no second code blocker.

Mechanical state: the branch is mergeable and all review threads are resolved. After CI is green, the existing `CHANGES_REQUESTED` review state still needs a fresh disposition.

gat786:
I think it changed after main merge.

gat786:
i will fix it.

mchmarny (approved):

Issue #1281, opened 2026-06-09
DC8 — Daytime human-access deployment


njhensley:
**Parent:** #1264 · **Labels:** `enhancement`, `area/ci` (auto, `.github/**`),
`area/tests` (auto, `tests/**`), `area/docs` (auto) · **Size:** M · **Depends on:**
DC1 (the day/night scheduler + daytime lease) and DC2 (provision-and-hold lifecycle +
enforced cleanup boundary)

**Goal.** At the morning handoff, stand up **one long-lived deployment per cloud** that
humans interact with **outside CI** through the working day, then tear it down before
the nightly batch. The two flavors are split across the two clouds — **AWS hosts a
training cluster, GCP hosts an inference cluster** (≈1 training + 1 inference total) —
so the day's scarce reservation time is divided between human use (DC8) and the nightly
version matrix (DC5). The inference deployment exposes a served (OpenAI-compatible)
endpoint humans can hit; the training deployment lets humans submit jobs / demo. This
is **not a CI lane**: nothing here emits a TestGrid column, and access is distributed
out-of-band.

**Scope.**
- **Morning-handoff provisioning.** After DC1's nightly batch drains a reservation,
  acquire the daytime lease (DC1) and use DC2's **provision-and-hold** mode to stand up
  the daytime cluster sized for its intent, deploy the workload (training stack on AWS,
  Dynamo inference on GCP), and **leave it up** — no per-run teardown. Use a stable,
  lease-tagged cluster name (not `aicr-uat-${{ github.run_id }}`, `uat-aws.yaml:53`) so
  the evening teardown and DC2's pre-batch guard can find it.
- **Inference served endpoint.** The GCP inference deployment exposes its
  OpenAI-compatible endpoint (the same Dynamo `DynamoGraphDeployment` DC3 deploys for
  `phase_serve`) for humans to send completions to; the AWS training deployment exposes
  a way to submit Kubeflow `TrainJob`s / run a demo. Reuse the DC3 serve assets and the
  DC2 inference config — DC8 does not invent a new serving stack.
- **Out-of-band access.** Distribute the kubeconfig / endpoint URL **out-of-band** (a
  team channel / shared secret store), **never** through the CI path, the evidence
  bundle, or the TestGrid dashboard. The daytime cluster produces **no** TestGrid
  column and emits **no** evidence bundle — it is human-facing, not a UAT cell.
- **Evening teardown before the batch.** Tear the daytime deployment down and release
[...]

PR #2042, opened 2026-08-04
fix(uat): ephemeral daytime clusters, consistent cross-cloud names

njhensley:
## Summary

Two related daytime-UAT changes: (1) make the daytime human-access clusters **ephemerally named** (unique per provision) instead of reusing one stable name per reservation, unifying the prefix to `aicr-uat-` across AWS/GCP/Azure; and (2) drop the morning `daytime-up` cron so standing a cluster up is **on-demand**, keeping only the evening `daytime-down` cron as a safety-net teardown.

## Motivation / Context

The daytime lifecycles named the cluster `aicr-uat-day-<reservation>` (AWS) / `aicr-day-<reservation>` (GCP/Azure) — the **same name every day**. So the morning `daytime-up` re-provisioned the exact name the previous evening's `daytime-down` had deleted, colliding with cloud deletion tombstones and stale terraform-state locks (the Azure blob-lease lock especially). This is the same failure class that motivated run-scoped clusters to be uniquely named. Separately, auto-provisioning every daytime cluster on a morning cron burned scarce GPU capacity whether or not anyone wanted it — human use is better served on demand.

Fixes: N/A
Related: #1275, #1281

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT pipelines + scheduler (`.github/workflows/uat-{aws,gcp,azure,daytime}.yaml`)

## Implementation Notes

**Ephemeral naming + discovery**
- `daytime-up` appends the up-run's id: `aicr-uat-day-<reservation>-<run_id>`, unique per provision.
- Because the name is no longer reconstructable from the reservation alone, `daytime-down` and the pre-batch guard find the held cluster by scanning the stable, reservation-scoped **prefix** (`list-clusters` / `clusters list` / `az aks list` + match) — the same mechanism already used for run-scoped clusters. A new "Resolve daytime cluster to tear down" step resolves the concrete name and overrides `DEPLOYMENT_ID`; the actuator derives the cluster name, resource group, and TF-state key from `.deployment.id`, so the discovered name is all teardown needs. **0 matches** = benign no-op (teardown skips via `DAYTIME_FOUND`), **>1** = fail loud.
- The guard now also runs for `daytime-up` and **refuses** (naming the exact `lifecycle=daytime-down` reclaim command) if a prior day's cluster leaked — with unique names, a leaked cluster would otherwise let two stack on a reservation that can host one. Refuse, not auto-reap.
- Prefix match anchored to the numeric `run_id` (`^<prefix>[0-9]`) so a reservation name that is a prefix of another cannot false-match.
- Consistent naming: GCP/Azure had drifted to `aicr-<run_id>` / `aicr-day-<reservation>`, diverging from their own committed `id: aicr-uat` in `cluster-config.yaml`. All three now use `aicr-uat-<run_id>` and `aicr-uat-day-<reservation>-<run_id>`. Nothing keys off the prefix (the other `aicr-uat-*` references are evidence labels and a fixed identity RG).

**On-demand daytime-up (`uat-daytime.yaml`)**
- Removed the morning `0 15 * * *` cron (and the `MORNING_UP_CRON` env + its Resolve-action branch). `daytime-up` is now manual only — dispatch `uat-daytime.yaml -f action=up` (whole rotation) or `uat-run.yaml -f lifecycle=daytime-up` (single reservation).
- The evening `daytime-down` cron stays as an **unconditional** safety net: it releases the reservation ~2h before the nightly batch whether or not a cluster was stood up, so a manually-held cluster never leaks into the nightly window.
- Docs describe how to run daytime-up by hand.

## Testing

```bash
actionlint .github/workflows/uat-{aws,gcp,azure,daytime}.yaml   # clean (only a pre-existing SC2016 info in an unrelated Summary step)
yamllint  .github/workflows/uat-{aws,gcp,azure,daytime}.yaml    # clean
```

No Go changed. Not hardware-tested — correctness backstops are the idempotent no-op and the fail-closed guard.

## Risk Assessment

- [x] **Medium** — Touches all per-cloud UAT pipelines + the daytime scheduler; CI/infra only, no product code.

**Rollout notes:** Tear down any daytime cluster still standing under the *old* name (`aicr-uat-day-<reservation>` / `aicr-day-<reservation>`) before this lands — the new prefix scan won't discover an old-scheme name. Independent of the in-flight single-cluster-reuse work (separate branch); the two touch adjacent guard code and will have a small resolvable overlap on whichever merges second.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — N/A, no Go changed
- [x] Linter passes (`make lint`) — actionlint + yamllint clean
- [ ] I added/updated tests for new functionality — N/A (workflow YAML)
- [x] I updated docs if user-facing behavior changed (`docs/contributor/uat.md`)


mchmarny (approved):
**APPROVE** - 0 blockers at head `1784c75`.

Reviewed the ephemeral daytime naming, the prefix-scan guard and resolver on all three clouds, the `DAYTIME_FOUND` teardown gates, the `uat-daytime.yaml` scheduler change, and the docs.

The load-bearing mechanic is `echo "DEPLOYMENT_ID=..." >> "$GITHUB_ENV"` overriding a job-level `env.DEPLOYMENT_ID`, so I checked it against the runner rather than assuming. `FileCommandManager` writes `context.Global.EnvironmentVariables[name]`, and `StepsRunner` rebuilds each step's `env` context from `Global.EnvironmentVariables` before evaluating that step's `if:`, layering only *step-level* `env:` on top. Job-level `env` is seeded into the same dictionary at job init, so the later `GITHUB_ENV` write wins for every subsequent step. `env.DAYTIME_FOUND` works for the same reason and evaluates false when unset.

Other checks:

- Permissions for the new list calls exist: `eks:*` (`infra/uat-aws-account/federation.tf`), `roles/container.admin` (`infra/uat-gcp-account/iam.tf`), subscription Contributor (`infra/uat-azure-account/federation.tf`). No fail-closed guard will trip on a missing IAM action.
- `mapfile -t held < <(... || true)` yields a defined empty array under `set -u` on bash 5, so `${#held[@]}` is safe on the zero-match path.
- Longest GKE name is `aicr-uat-day-gcp-h100-<run_id>` at 33 chars, inside the 40-char cap.
- No stale `aicr-day-` or unsuffixed `aicr-uat-day-<reservation>` references remain outside the files this PR rewrites.

One consequence worth naming rather than blocking: for `daytime-down`, a cancelled or failed Resolve leaves `DAYTIME_FOUND` unset, so every teardown step skips and the cluster stays up. That is the deliberate design (the step says so loudly and prints the re-dispatch command), and the pre-batch guard catches it on the next run.

PR #2043, opened 2026-08-04
feat(ci): add registry-inventory supply-chain egress gate

njhensley:
## Summary

Adds `tools/registry-inventory`, a static analyzer that inventories every package/container registry AICR's builds and CI reach out to, a low-churn allowlist drift gate so a newly introduced registry can't slip in unnoticed, and a committed host-level docs page (`docs/contributor/registry-egress.md`) as the browsable, diffable companion.

## Motivation / Context

We had no single, checkable answer to "what registries does the build/CI touch?" — the transitive chart→image surface is covered by `tools/bom`, but the *direct* egress (chart repos, base images, ko push targets, GitHub Actions, dev-tool installers) was only knowable by hand. This makes that surface explicit, enforces review when it grows, and publishes it as a docs page — mirroring the `tools/bom` gate + committed-doc pattern.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Docs/examples (`docs/contributor/registry-egress.md` + nav)
- [x] Other: `tools/registry-inventory` (new), Makefile targets, `.github/CODEOWNERS`, `.gitignore`

## Implementation Notes

- **Sources parsed (structured, high-reliability):** `recipes/registry.yaml` (chart repos), `.settings.yaml`, `.goreleaser.yaml`/`.ko.yaml` (base images + ko push repos), Dockerfiles (`FROM`), `.github` Actions `uses:`, and `tools/setup-tools` (best-effort shell installers). Normalized to `(host, packageType, direction, consumer, pinType, detail, source)` records; emitted as YAML + Markdown via `make registry-inventory`.
- **Three gates under `make test`:**
  - `TestRegistryHostsWithinAllowlist` — fails on a host absent from the CODEOWNER-gated `registry-allowlist.yaml`, **and** on a required source that failed to parse (fail-closed against silent under-detection).
  - `TestExternalActionsArePinned` — SHA-pinned third-party actions.
  - `TestCommittedRegistryEgressDocFresh` — the committed docs page is a byte projection of the live sources. It's **host-level only**, so it drifts (and the test fails) just when the egress *surface* changes — low-churn, matching the allowlist gate's sensitivity, the same discipline as the BOM. Per-record pins/versions (which churn on every bump) live only in the `dist/` output.
  - `make registry-check` is the opt-in standalone CLI (mirrors `bom-check`).
- **Deliberately out of scope, reported as `uncoveredSources` rather than silently dropped:** rendered chart→image pulls (owned by `tools/bom`), shell-download URLs built from variables, Go Sigstore endpoint constants, and per-*owner* GitHub Action trust.
- The allowlist is CODEOWNER-gated to `@nvidia/aicr-maintainer` — a trust root, so adding a host requires maintainer sign-off.
- Built via `go run` (no committed binary); a stray build output is gitignored.

The design and this implementation were hardened through an internal multi-persona + adversarial review pass (correctness, security/supply-chain, operability, test-coverage).

## Testing

```bash
GOFLAGS=-mod=vendor go test -race -cover ./tools/registry-inventory/...   # ok, 85.5% coverage
golangci-lint run -c .golangci.yaml ./tools/registry-inventory/...        # 0 issues
yamllint / check-docs-filenames / check-docs-mdx                          # clean (doc is MDX-safe)
make registry-check                                                        # OK, 25 hosts all allowlisted
make registry-docs && git diff --stat                                      # byte-stable (no diff)
```

Negative paths verified: missing required sources make `-check` fail closed; a stale committed doc fails `TestCommittedRegistryEgressDocFresh`.

## Risk Assessment

- [x] **Low** — Additive, self-contained tool package + one generated docs page; no product code paths touched. Always-on enforcement is the three tests under `make test`; `make registry-check` is opt-in.

**Rollout notes:** N/A. `registry-check` is intentionally not wired into `make qualify`/merge-gate (opt-in, like `bom-check`); the drift, source-error, action-pinning, and doc-freshness gates run via `make test`.

## Checklist

- [x] I updated docs if user-facing behavior changed (new contributor docs page + tool README)
- [x] Changes follow existing patterns in the codebase (`tools/bom` gate + committed-doc shape)


.github/CODEOWNERS
```diff
@@ -33,6 +33,13 @@ recipes/** @nvidia/aicr-maintainer
 recipes/evidence/** @nvidia/aicr-maintainer
 recipes/evidence/allowlist.yaml @nvidia/aicr-maintainer
 
+# Registry-egress allowlist (tools/registry-inventory). Like the evidence
+# allowlist above, this file is a trust root: the registry-check gate passes
+# only for hosts listed here, so adding one widens the build/CI supply-chain
+# surface. Called out explicitly so a new host never rides in on a broader
+# tools/ approval — a maintainer must sign off.
+tools/registry-inventory/registry-allowlist.yaml @nvidia/aicr-maintainer
```
mchmarny: [blocking] Protect the detector, not only the allowlist. `tools/registry-inventory/**` still falls under the broader `@nvidia/aicr-write` ownership, so a PR can remove an extractor, regenerate the host doc, and leave all three gates green without maintainer approval. I reproduced this by removing `extractGitHubActions`; `github-actions` became merely “unused” and every gate passed. Make the detector path maintainer-owned so weakening detection requires the same sign-off as widening the allowlist.

tools/registry-inventory/sources.go
```diff
@@ -0,0 +1,466 @@
...
+		}
+		if ref == "scratch" {
+			continue
+		}
+		if _, isStage := stages[strings.ToLower(ref)]; isStage {
+			continue // FROM builder — an earlier stage, not a registry pull
+		}
+		pt, pin := imagePin(ref)
+		recs = append(recs, Record{
+			Host: imageHost(ref), PackageType: PkgContainerImage, Direction: DirPull,
```
mchmarny, line 418: [blocking] Fail closed on interpolated `FROM` values. With `ARG BASE=evil.example/image:v1` followed by `FROM ${BASE}`, this passes `${BASE}` to `imageHost`, records `docker.io`, and never surfaces `evil.example` to the allowlist. Parse global `ARG` defaults or return a source error for unresolved `FROM` refs, and add that case as a regression test.

mchmarny (changes requested):
Request changes — 3 fail-open blockers on `cac85686`. Two new findings are inline; the existing required-source failure thread (https://github.com/NVIDIA/aicr/pull/2043#discussion_r3715103819) is also merge-blocking. Focused Go tests, lint, registry check, and docs checks pass, but adversarial probes reproduce the bypasses. Live CI has no failures; remaining lanes are still running.

njhensley:
Thanks for the adversarial probes, @mchmarny — all three fail-open blockers are fixed on the latest push, plus the CodeRabbit items. Details:

### Blockers (fail-closed)
1. **Detector not CODEOWNER-protected** → `tools/registry-inventory/**` is now maintainer-owned (was just the allowlist). Weakening a detector now needs the same sign-off as widening the allowlist. Reproduced your `extractGitHubActions`-removal scenario; it's the review gate that has to catch this, and now does.
2. **Interpolated `FROM`** → Dockerfile parsing now collects global `ARG NAME=default` values and substitutes them, so `ARG BASE=evil.example/image:v1` + `FROM ${BASE}` surfaces `evil.example` to the allowlist. A `FROM` whose **host segment** is still unresolved (`FROM ${UNDECLARED}`) returns a `sourceError` (fail closed) instead of being mislabeled `docker.io`. An interpolated *tag* with a concrete host (`golang:${GO_VERSION}`) is unaffected. Regression tests added (`TestDockerfileInterpolatedFrom`).
3. **Required-source failures classed as warnings** (the r3715103819 thread) → `extractDockerfiles` and `extractGitHubActions` now return read/walk failures as `sourceErrors`, and a missing/unreadable `.github` fails closed. The gate already escalates on `sourceErrors`, so a dropped source can no longer stay green. Tests added.

### CodeRabbit items addressed
- `isFirstPartyAction` no longer treats an empty/malformed identity as first-party (fail-closed; test added).
- Shell logger skip is now **per-segment** (split on `&&`/`||`/`|`), so `log_info "x" && curl evil` still surfaces the curl — a logger prefix no longer masks the rest of the line.
- `loadAllowlist` maps only `fs.ErrNotExist` → `ErrCodeNotFound` (else `ErrCodeInternal`); caller uses `PropagateOrWrap` to preserve the code.
- `isSHA40` accepts uppercase hex; `mdEscape` flattens CR/LF; https-under-any-key uses a generic `http` package type; `sum.golang.org` documented on the `proxy.golang.org` entry; test helpers replaced with `strings.Contains`.

### Respectfully declining two suggestions (happy to revisit)
- **Move business logic into `pkg/*`** — this is a self-contained build/CI helper with no external Go consumers, matching `tools/bom` (which keeps its logic in `package main` over a thin `pkg/bom`). Extracting a package here adds surface without a caller. If you'd prefer the split for consistency, I'll do it.
- **`context.Context` propagation / errgroup** — the tool does zero network or subprocess I/O; it's pure local file reads, fully deterministic, with nothing to time out or cancel (`tools/bom` takes a context only because it shells out to `helm`). I left it out deliberately, but will add a bounded context if maintainers want uniformity.

**Verified:** `go test -race` (86% coverage), `golangci-lint` 0 issues, `make registry-check` OK (25 hosts), `make registry-docs` byte-stable, `check-docs-mdx`/yamllint clean. Negative paths confirmed: unresolved `FROM` host, missing `.github`, and missing required sources all fail `-check` closed.

njhensley:
Second CodeRabbit round addressed on the latest push:

**Fixed (both Major + quick-wins):**
- **`substituteArgs` ARG-prefix bug** — switched to a regex that captures the whole `\w+` identifier, so a shorter `$REG` can no longer consume the prefix of `$REGISTRY`. Regression test added (overlapping ARG names).
- **Shell scanner failures → `sourceErrors`** — an oversized line past the 1 MiB token limit now fails closed (consistent with the Dockerfile scanner), extracted into a testable `scanShellData` with an oversized-line regression test.
- **`sum.golang.org`** — the `go install` extractor now emits it alongside `proxy.golang.org`, and it's in the allowlist (26 hosts). It's real egress, so it's gated, not just documented.
- **README** — now lists all three `make test` gates (incl. `TestCommittedRegistryEgressDocFresh`), adds Dockerfiles/`.github` to the required-source list with the `sourceErrors` note, and adds `http` to the packageType values.
- **Uppercase-SHA test case** added to `TestActionPin`.

**Respectfully skipping (with reason):**
- **Thread `context.Context` through `loadAllowlist`/`diffAllowlist`** — same rationale as before: the tool does only local, non-cancellable file reads with no network/subprocess I/O, so there's nothing to bound. Happy to add if maintainers want uniformity.
- **Convert the new subtests to table-driven** — `TestDiffAllowlist`, `TestLoadAllowlistErrors`, and `TestDockerfileInterpolatedFrom` have genuinely *different assertions per case* (record-host vs sourceError-presence vs error-code), which reads more clearly as named subtests than a uniform table. The uniform input→output tests already are table-driven.

Verified: `go test -race` 86.3%, lint 0, `registry-check` OK (26 hosts), docs byte-stable, mdx/yaml clean.

njhensley:
Third CodeRabbit round — all 5 addressed (all valid):

- **`koPushRepos` missed same-indent list items** (Major) — YAML allows `- item` aligned with the `repositories:` key; entry matching now uses `>= baseIndent`. Same-indent regression test added.
- **`goInstallPin` classified bare `$VAR` as a tag** — now any `$` in the version (`@$TOOL_VERSION`, `@${V}`, `@$(cmd)`) is a var pin. Test added.
- **Dockerfile walk `werr` passed silently** — now recorded as a `sourceError`, matching `extractGitHubActions` (the two passes no longer disagree; an unreadable dir hiding a Dockerfile fails closed).
- **`go install` checksum-db record untested** — added a test asserting both `proxy.golang.org` and `sum.golang.org` records.
- **Vacuous sort assertion** in the markdown test (both-absent passed) — now asserts each host is present before comparing positions.
- Also added the slashless `host:port` case for `imagePin` (the F9 branch).

Verified: `go test -race` 86.3%, lint 0, `registry-check` OK (26 hosts), docs byte-stable.

njhensley:
Round 4 (1 Major) addressed: **ARG defaults now apply positionally**. Dockerfile parsing is a single in-order pass — an `ARG` only affects the `FROM`s that follow it, and a later re-declaration can't retroactively change an earlier `FROM` (the previous two-pass approach let the last `ARG` win globally). A bare `ARG NAME` (no default) stays unresolved and fails closed. Regression test added asserting `FROM ${BASE}` before/after a re-declare resolves to the two distinct hosts. Verified: `go test -race`, lint 0, `registry-check` OK (26 hosts).

mchmarny (approved):

PR #2046, opened 2026-08-04
feat(cli): support aicr verify via AICRConfig (spec.verify)

### LINKED ISSUE #1567: Support aicr verify via AICRConfig (spec.verify)
## Summary

Extend `AICRConfig` (`--config`) to cover `aicr verify`. Today the schema is scoped to the **producer pipeline** — `pkg/config/doc.go` states the config is accepted "on the snapshot, recipe, bundle, and validate commands." `verify` is the consumer/verifier side and has no `--config` support, so verification policy can't be captured in the same version-controlled document as the rest of an end-to-end workflow.

## Motivation

A `--config`-driven workflow can build and validate reproducibly, but the verification gate (trust floor, pinned creator, CLI-version constraint, private trust root, KMS key) still has to be passed as flags every time. A `spec.verify` section would let teams commit their verification policy alongside the recipe/bundle/validate config — useful for both end users and for testing the supply-chain surface.

## Proposed change (for discussion)

Add an optional `spec.verify` section mapping the `aicr verify` flags that are durable (non-secret) inputs:

- `minTrustLevel` (`--min-trust-level`)
- `requireCreator` (`--require-creator`)
- `cliVersionConstraint` (`--cli-version-constraint`)
- `certificateIdentityRegexp` (`--certificate-identity-regexp`)
- `key` (`--key`, a KMS URI or local PEM path)
- `trustRoot` (`--trust-root`, a path)

All of these are references/policy, not secrets, so none of the identity-token concerns apply. CLI flags would win over config, consistent with the existing commands.

## Open question / scope

This is a **scope expansion**: the config was intentionally limited to the producer pipeline, and `verify` introduces a consumer-side actor to the schema. Filing under the epic for the team to decide whether consumer-side verification policy belongs in `AICRConfig`. **Close as out-of-scope if the team prefers to keep the config producer-only.**

Epic: #1149
Related: #1152 (`verify --key`), #1153 (`verify --trust-root`)

lockwobr:
## Summary

Adds an optional `spec.verify` section to `AICRConfig` and a `--config` flag to
`aicr verify`, so verification policy (trust floor, pinned creator, CLI-version
constraint, private trust root, KMS key) can be committed alongside the rest of
an end-to-end workflow instead of retyped as flags on every invocation.

## Motivation / Context

`AICRConfig` was scoped to the producer pipeline — `pkg/config/doc.go` stated it
was accepted "on the snapshot, recipe, bundle, and validate commands." `verify`
is the consumer side and had no `--config` support, so a `--config`-driven
workflow could build and validate reproducibly but still had to pass the
verification gate by hand every time.

This matters most where the verify step runs as a separate CI stage from the
build: that stage's trust floor now lives in version control next to the recipe.

Fixes: #1567
Related: #1149, #1152, #1153

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

**Schema shape.** The issue sketched six flat fields. They are grouped into two
sub-sections instead, mirroring how `aicr verify` consumes them: `policy` feeds
`verifier.Policy` (assertions checked after verification runs) and `trust` feeds
`verifier.VerifyOptions` (the material verification runs against). This also
matches the nesting every other `spec.*` section already uses, and disambiguates
otherwise-bare names — `trust.key` reads clearly where a top-level `key:` would
not. Field names themselves are unchanged from the issue.

**Deliberate exclusions.** Three `verify` flags are not in the schema: the bundle
directory (a positional argument), `--format` (presentation, not policy), and
`--insecure-ignore-tlog`. The last is a policy *weakener* that drops the
transparency-log requirement, so keeping it command-line-only means a committed
file can never silently disable that check. It still composes with a
config-supplied `trust.key`, and the precondition guard now reads the resolved
key rather than the raw flag.

**Shared constraint parser.** `verifier.ParseVersionConstraint` is extracted and
exported so the config layer and `CheckPolicy` validate `cliVersionConstraint`
through one grammar rather than two that can drift. `CheckPolicy` now calls it.

**Spec-path attribution.** Validation happens at load time so a typo fails as
`invalid spec.verify.policy.minTrustLevel` rather than after a full verification
run. This initially did not work: `errors.PropagateOrWrap` passes the inner error
through untouched when it already carries `ErrCodeInvalidRequest`, which silently
discarded the spec path. Switched to plain `errors.Wrap`, matching the sibling
resolvers in the same file.

**A validation limit, documented rather than papered over.** The shared parser
splits off a leading comparison operator and rejects an empty remainder, but does
not check the remainder is version-shaped. So `">="` is rejected at load time
while `">= not-a-version"` is accepted and fails later at evaluate time. Pinned
by `TestVerifyResolve_VersionConstraintValueNotVersionChecked` and noted in the
godoc and the CLI reference. Tightening it would mean a stricter grammar in
config than `aicr verify` itself accepts, which this package explicitly avoids.

**`key` and `trustRoot` are passed through unvalidated.** They are references
whose resolution (KMS reachability, file contents) belongs to the verifier, which
reports far better errors than a syntactic check here could.

## Testing

```bash
go test ./pkg/config/... ./pkg/bundler/verifier/... ./pkg/cli/...   # all ok
golangci-lint -c .golangci.yaml run                                 # 0 issues
```

The CLI wiring was written test-first: the six `bundle_verify_config_test.go`
cases were added before the resolved values were applied in
`runBundleVerifyCmd`, five failed, and the implementation followed. That is what
surfaced the `PropagateOrWrap` defect above, which a tests-after pass would have
locked in as correct.

Coverage on changed packages: `pkg/config` 93.6%, `pkg/bundler/verifier` 76.3%,
`pkg/cli` 74.2%. Every new exported symbol (`ParseVersionConstraint`,
`VerifySpec.Resolve`, `AICRConfig.Verification`) is covered.

Note on `make qualify`: it does not pass on the contributor workstation used
here, for two pre-existing environmental reasons unrelated to this branch — local
`helm` is v4.2.0 against a v4.2.3 exact pin, and GNU `timeout` is absent on
macOS so the `.github/scripts` release tests abort. Both should pass on CI, which
provisions the pinned toolchain.

## Risk Assessment

- [x] **Low** — Additive only. The new `spec.verify` section is optional, every
  existing config document keeps validating, and with no `--config` the verify
  command behaves exactly as before (all pre-existing `bundle_verify` tests pass
  unchanged).

**Rollout notes:** N/A — no migration, no behavior change for existing callers.

## Checklist

### DIFF
--- a/docs/user/cli-config.md
+++ b/docs/user/cli-config.md
@@ -314,8 +314,11 @@ an attested bundle whose binary attestation is absent, or one carrying external
 
 What no policy value can wave through: checksum failures, and attestations that
 are present but fail verification. Those are rejected regardless of the floor.
-`aicr verify` logs at INFO whenever config sets the floor to anything other than
-`max`.
+
+`aicr verify` logs the floor at INFO when config is what supplies it: that is,
+when `--min-trust-level` is absent and the configured value is anything other
+than `max`. An explicit flag takes precedence instead, and that override is
+logged separately by the flag-precedence path.
 
 Every field is a durable, non-secret reference or policy value; no private key
 material is part of the schema. Three `aicr verify` flags are deliberately
--- a/docs/user/cli-config.md
+++ b/docs/user/cli-config.md
@@ -301,6 +301,15 @@ checked after verification runs, `trust` holds the material it verifies against.
 | `trust.key` | string | KMS key URI (`awskms://` \| `gcpkms://` \| `azurekms://` \| `hashivault://`) or local PEM public-key path; the verify counterpart to `spec.bundle.attestation.signingKey` |
 | `trust.trustRoot` | string | Path to a private Sigstore `trusted_root.json`, additive to the built-in public-good root |
 
+`policy.minTrustLevel` sets **operator policy, not an org-enforced guardrail**.
+A committed value lowers the effective floor as readily as it raises it
+(`unknown` makes the trust check a no-op, since every level meets it), so treat
+it as a reviewable choice rather than a control that cannot be relaxed. Its
+blast radius is bounded: verification still rejects checksum-failed and
+invalidly-signed bundles regardless of policy, so only checksum-valid *unsigned*
+bundles pass a lowered floor. `aicr verify` logs at INFO whenever config sets the
+floor to anything other than `max`.
+
 Every field is a durable, non-secret reference or policy value; no private key
 material is part of the schema. Three `aicr verify` flags are deliberately
 excluded: the bundle directory (a positional argument), `--format`
--- a/docs/user/cli-config.md
+++ b/docs/user/cli-config.md
@@ -304,11 +304,18 @@ checked after verification runs, `trust` holds the material it verifies against.
 `policy.minTrustLevel` sets **operator policy, not an org-enforced guardrail**.
 A committed value lowers the effective floor as readily as it raises it
 (`unknown` makes the trust check a no-op, since every level meets it), so treat
-it as a reviewable choice rather than a control that cannot be relaxed. Its
-blast radius is bounded: verification still rejects checksum-failed and
-invalidly-signed bundles regardless of policy, so only checksum-valid *unsigned*
-bundles pass a lowered floor. `aicr verify` logs at INFO whenever config sets the
-floor to anything other than `max`.
+it as a reviewable choice rather than a control that cannot be relaxed.
+
+A lowered floor admits any bundle whose actual trust level reaches it, which is
+broader than unsigned bundles. It also covers chains that legitimately degraded:
+an attested bundle whose binary attestation is absent, or one carrying external
+`--data`, both report `attested` against a `verified` maximum, so the default
+`max` rejects them while a lowered floor does not.
+
+What no policy value can wave through: checksum failures, and attestations that
+are present but fail verification. Those are rejected regardless of the floor.
+`aicr verify` logs at INFO whenever config sets the floor to anything other than
+`max`.
 
 Every field is a durable, non-secret reference or policy value; no private key
 material is part of the schema. Three `aicr verify` flags are deliberately
--- a/docs/user/cli-config.md
+++ b/docs/user/cli-config.md
@@ -1,24 +1,30 @@
 # CLI Configuration File (AICRConfig)
 
 `AICRConfig` is a Kubernetes-style YAML/JSON document that captures the inputs
-to the four workflow commands — `aicr snapshot`, `aicr recipe`, `aicr bundle`,
-and `aicr validate` — so an end-to-end run version-controls as a single file
-instead of a shell script full of flags. Each command accepts it through the
-same `--config` flag:
+to the five workflow commands — `aicr snapshot`, `aicr recipe`, `aicr bundle`,
+`aicr validate`, and `aicr verify` — so an end-to-end run version-controls as a
+single file instead of a shell script full of flags. Each command accepts it
+through the same `--config` flag:
 
 ```shell
 aicr snapshot --config aicr-config.yaml
 aicr recipe   --config aicr-config.yaml
 aicr bundle   --config aicr-config.yaml
 aicr validate --config aicr-config.yaml
+aicr verify   --config aicr-config.yaml
 ```
+
+The first four are the producer pipeline; `spec.verify` is the consumer side, so
+one document can carry both how an artifact is built and the trust floor a
+downstream consumer enforces against it.
 
 This page documents the complete document schema in one place. The
 [CLI Reference](cli-reference.md) shows per-command usage in its
 [Snapshot](cli-reference.md#snapshot-config-file-mode),
 [Recipe](cli-reference.md#config-file-mode-recommended),
-[Validate](cli-reference.md#validate-config-file-mode), and
-[Bundle](cli-reference.md#bundle-config-file-mode) config-file-mode sections.
+[Validate](cli-reference.md#validate-config-file-mode),
+[Bundle](cli-reference.md#bundle-config-file-mode), and
+[Verify](cli-reference.md#verify-config-file-mode) config-file-mode sections.
 The schema's source of truth is
 [`pkg/config`](https://github.com/NVIDIA/aicr/tree/main/pkg/config).
 
@@ -34,11 +40,12 @@ spec:
   recipe: {}                   # at least ONE must be present
   bundle: {}
   validate: {}
+  verify: {}
 ```
 
 Each `spec.*` section is optional and each command reads only its own section,
 so a file may carry just one section or any combination. A document with none
-of the four sections is rejected.
+of the five sections is rejected.
 
 ## Loading, Precedence, and Secrets
 
@@ -183,6 +190,15 @@ spec:
         push: ""                     # OCI ref to push the signed bundle
         plainHTTP: false
         insecureTLS: false
+  verify:                            # consumer side: policy for `aicr verify`
+    policy:                          # assertions checked after verification
+      minTrustLevel: verified        # unknown | unverified | attested | verified | max
+      requireCreator: ci@myorg.example.com
+      cliVersionConstraint: ">= 0.16.0"
+    trust:                           # material verification runs against
+      certificateIdentityRegexp: ""  # must contain NVIDIA/aicr when set
+      key: ""                        # KMS URI or local PEM public-key path
+      trustRoot: ""                  # private Sigstore trusted_root.json
 ```
 
 ## Field Reference
@@ -270,9 +286,31 @@ Inputs to `aicr validate`.
 | `evidence.attestation.bom` / `.push` | string | BOM input; OCI ref for the signed bundle push |
 | `evidence.attestation.plainHTTP` / `.insecureTLS` | bool (tri-state) | Push transport options |
 
+### spec.verify
+
+Verification policy for `aicr verify`, the one consumer-side section. The two
+sub-sections mirror how the command consumes them: `policy` holds assertions
+checked after verification runs, `trust` holds the material it verifies against.
+
+| Field | Type | Notes |
+|-------|------|-------|
+| `policy.minTrustLevel` | string | `unknown` \| `unverified` \| `attested` \| `verified`, or `max` (the CLI default) to auto-detect the highest level the bundle can reach |
+| `policy.requireCreator` | string | Pins the OIDC identity in the bundle attestation's signing certificate |
+| `policy.cliVersionConstraint` | string | Constrains the `aicr` version in the attestation predicate; supports `>=`, `>`, `<=`, `<`, `==`, `!=`, and a bare version means `>=` |
+| `trust.certificateIdentityRegexp` | string | Certificate identity pattern for binary attestation verification; must contain `NVIDIA/aicr` |
+| `trust.key` | string | KMS key URI (`awskms://` \| `gcpkms://` \| `azurekms://` \| `hashivault://`) or local PEM public-key path; the verify counterpart to `spec.bundle.attestation.signingKey` |
+| `trust.trustRoot` | string | Path to a private Sigstore `trusted_root.json`, additive to the built-in public-good root |
+
+Every field is a durable, non-secret reference or policy value; no private key
+material is part of the schema. Three `aicr verify` flags are deliberately
+excluded: the bundle directory (a positional argument), `--format`
+(presentation, not policy), and `--insecure-ignore-tlog`, which weakens the
+trust floor and so stays command-line-only rather than something a committed
+file can silently enable. It still composes with a config-supplied `trust.key`.
+
 ## Cross-Section Rules
 
-- At least one of the four `spec.*` sections must be present.
+- At least one of the five `spec.*` sections must be present.
 - `spec.recipe.criteria` and `spec.recipe.input.snapshot` are mutually
   exclusive.
 - When **both** `spec.recipe.output.path` and `spec.bundle.input.recipe` are
--- a/docs/user/cli-config.md
+++ b/docs/user/cli-config.md
@@ -11,7 +11,7 @@ aicr snapshot --config aicr-config.yaml
 aicr recipe   --config aicr-config.yaml
 aicr bundle   --config aicr-config.yaml
 aicr validate --config aicr-config.yaml
-aicr verify   --config aicr-config.yaml
+aicr verify   ./my-bundle --config aicr-config.yaml   # bundle dir is positional
 ```
 
 The first four are the producer pipeline; `spec.verify` is the consumer side, so
--- a/pkg/bundler/verifier/trust.go
+++ b/pkg/bundler/verifier/trust.go
@@ -236,7 +236,11 @@ func (r *VerifyResult) MaxAchievableTrustLevel() TrustLevel {
 // here and fails later at Evaluate.
 func ParseVersionConstraint(expr string) (*constraints.ParsedConstraint, error) {
 	expr = strings.TrimSpace(expr)
-	if !hasOperatorPrefix(expr) {
+	// A leading "!" is never a bare version. The shared parser rejects a bare
+	// "!" prefix outright (it means the user wanted "!=" or the node-set
+	// form), so prepending ">=" here would hide invalid syntax from that
+	// check and let it through as ">= !1.2.3".
+	if !hasOperatorPrefix(expr) && !strings.HasPrefix(expr, "!") {
 		expr = ">= " + expr
 	}
 	constraint, err := constraints.ParseConstraintExpression(expr)
--- a/pkg/bundler/verifier/trust.go
+++ b/pkg/bundler/verifier/trust.go
@@ -179,13 +179,9 @@ func (r *VerifyResult) CheckPolicy(p Policy) (string, error) {
 		if r.ToolVersion == "" {
 			return "tool version not available in attestation (bundle may be unattested)", nil
 		}
-		expr := strings.TrimSpace(p.VersionConstraint)
-		if !hasOperatorPrefix(expr) {
-			expr = ">= " + expr
-		}
-		constraint, err := constraints.ParseConstraintExpression(expr)
+		constraint, err := ParseVersionConstraint(p.VersionConstraint)
 		if err != nil {
-			return "", errors.Wrap(errors.ErrCodeInvalidRequest, "invalid tool version constraint", err)
+			return "", err
 		}
 		passed, err := constraint.Evaluate(r.ToolVersion)
 		if err != nil {
@@ -223,6 +219,31 @@ func (r *VerifyResult) MaxAchievableTrustLevel() TrustLevel {
 		return TrustAttested
 	}
 	return TrustVerified
+}
+
+// ParseVersionConstraint parses a CLI-version constraint expression using the
+// same grammar CheckPolicy enforces at verify time, applying the bare-version
+// default (e.g. "0.8.0" means ">= 0.8.0").
+//
+// Exported so configuration layers can reject a malformed constraint when the
+// document is loaded rather than after a full verification run. Sharing the
+// parser is what keeps the two entry points from drifting: a value that parses
+// at load time is guaranteed to parse when Policy is evaluated.
+//
+// Note the check is operator-level only. The underlying parser splits off a
+// leading comparison operator and rejects an empty remainder, but does not
+// verify that the remainder is version-shaped, so ">= not-a-version" parses
+// here and fails later at Evaluate.
+func ParseVersionConstraint(expr string) (*constraints.ParsedConstraint, error) {
+	expr = strings.TrimSpace(expr)
+	if !hasOperatorPrefix(expr) {
+		expr = ">= " + expr
+	}
+	constraint, err := constraints.ParseConstraintExpression(expr)
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid tool version constraint", err)
+	}
+	return constraint, nil
 }
 
 // hasOperatorPrefix returns true if the expression starts with a comparison operator.
--- a/pkg/bundler/verifier/trust_test.go
+++ b/pkg/bundler/verifier/trust_test.go
@@ -265,6 +265,10 @@ func TestParseVersionConstraint(t *testing.T) {
 		{name: "surrounding whitespace tolerated", expr: "  0.8.0  ", satisfied: "0.9.0", rejected: "0.7.0"},
 		{name: "empty expression rejected", expr: "", wantErr: true},
 		{name: "operator with no version rejected", expr: ">=", wantErr: true},
+		// A bare "!" is invalid syntax that the shared parser rejects
+		// outright. Prepending ">=" would hide it from that check, so the
+		// bare-version default must not apply to "!"-prefixed input.
+		{name: "bare ! prefix rejected, not coerced to >=", expr: "!0.8.0", wantErr: true},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
--- a/pkg/bundler/verifier/trust_test.go
+++ b/pkg/bundler/verifier/trust_test.go
@@ -262,6 +262,8 @@ func TestParseVersionConstraint(t *testing.T) {
 		{name: "explicit ==", expr: "== 0.8.0", satisfied: "0.8.0", rejected: "0.8.1"},
 		{name: "explicit !=", expr: "!= 0.8.0", satisfied: "0.8.1", rejected: "0.8.0"},
 		{name: "explicit <", expr: "< 0.8.0", satisfied: "0.7.0", rejected: "0.8.0"},
+		{name: "explicit >", expr: "> 0.8.0", satisfied: "0.8.1", rejected: "0.8.0"},
+		{name: "explicit <=", expr: "<= 0.8.0", satisfied: "0.8.0", rejected: "0.9.0"},
 		{name: "surrounding whitespace tolerated", expr: "  0.8.0  ", satisfied: "0.9.0", rejected: "0.7.0"},
 		{name: "empty expression rejected", expr: "", wantErr: true},
 		{name: "operator with no version rejected", expr: ">=", wantErr: true},
--- a/pkg/bundler/verifier/trust_test.go
+++ b/pkg/bundler/verifier/trust_test.go
@@ -243,3 +243,57 @@ func TestGetTrustLevels(t *testing.T) {
 		}
 	}
 }
+
+// TestParseVersionConstraint covers the grammar shared by the CLI flag and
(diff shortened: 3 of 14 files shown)


docs/user/cli-config.md
```diff
@@ -1,24 +1,30 @@
...
+single file instead of a shell script full of flags. Each command accepts it
+through the same `--config` flag:
 
 ```shell
 aicr snapshot --config aicr-config.yaml
 aicr recipe   --config aicr-config.yaml
 aicr bundle   --config aicr-config.yaml
 aicr validate --config aicr-config.yaml
+aicr verify   --config aicr-config.yaml
 ```
```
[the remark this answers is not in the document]
lockwobr: Valid, fixed in 270890ea. `aicr verify` requires the bundle directory as a positional argument and errors with `bundle directory is required` before it ever reads `--config`, so that line was a copy-paste trap. Now reads `aicr verify ./my-bundle --config aicr-config.yaml` with a comment noting the positional.

pkg/config/resolve.go
```diff
@@ -496,6 +497,93 @@ func (v *ValidateSpec) Resolve() (*ValidateResolved, error) {
...
+		out.RequireCreator = v.Policy.RequireCreator
+		out.VersionConstraint = v.Policy.CLIVersionConstraint
+	}
+
+	if v.Trust != nil {
+		if v.Trust.CertificateIdentityRegexp != "" {
+			if err := verifier.ValidateIdentityPattern(v.Trust.CertificateIdentityRegexp); err != nil {
+				return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+					"invalid spec.verify.trust.certificateIdentityRegexp", err)
+			}
```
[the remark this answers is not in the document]
lockwobr: Valid, fixed in 270890ea. `ValidateIdentityPattern` only substring-checked for the required repository anchor, so `https://github.com/NVIDIA/aicr/[` passed and the failure surfaced later from the identity matcher instead of at load time with its spec path. That defeats the point of validating in `Resolve`.

Fixed in the shared validator rather than in `pkg/config`, so the `--certificate-identity-regexp` CLI flag path gets the same protection; it had the identical gap. This matches how `ParseVersionConstraint` is shared in this PR: one validator, two entry points, no drift.

Added `TestVerifyResolve_InvalidValues/identity_regexp_that_is_not_a_compilable_regexp` and `TestValidateIdentityPattern_RejectsUncompilablePattern`. Both were confirmed failing before the fix.

pkg/bundler/verifier/trust.go
```diff
@@ -225,6 +221,31 @@ func (r *VerifyResult) MaxAchievableTrustLevel() TrustLevel {
...
+// verify that the remainder is version-shaped, so ">= not-a-version" parses
+// here and fails later at Evaluate.
+func ParseVersionConstraint(expr string) (*constraints.ParsedConstraint, error) {
+	expr = strings.TrimSpace(expr)
+	if !hasOperatorPrefix(expr) {
+		expr = ">= " + expr
+	}
+	constraint, err := constraints.ParseConstraintExpression(expr)
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeInvalidRequest, "invalid tool version constraint", err)
```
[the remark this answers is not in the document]
lockwobr: Not applying: the premise does not hold. There is no bare-`!` rejection in the shared parser to bypass.

`ParseConstraintExpression` tries the operator prefixes `>=`, `<=`, `!=`, `==`, `>`, `<`. `!0.8.0` matches none of them (`!=` does not match `!0`), so it falls through to the exact-match branch with `Value: "!0.8.0"`, which is non-empty and therefore **accepted**. The only rejections in that function are an empty expression and an empty value after an operator.

Verified by probe rather than by reading alone:

```
ParseVersionConstraint("!0.8.0")   -> err=<nil>  constraint=">= !0.8.0"
ParseVersionConstraint(">= !0.8.0") -> err=<nil> constraint=">= !0.8.0"
evaluate "!0.8.0" against 0.9.0    -> ok=false err=non-numeric component: "!0"
```

So `!0.8.0` is accepted at load time with or without the `>=` prefix; dropping the prepend would change it from a `>=` comparison to an exact match, and it would still not be rejected. Removing the prepend therefore fixes nothing and would break the documented bare-version default (`"0.8.0"` means `">= 0.8.0"`).

The underlying observation — that a malformed version passes load validation and fails at evaluate time — is real, but it is already known, documented, and pinned by a test in this PR: see `TestVerifyResolve_VersionConstraintValueNotVersionChecked`, the `ParseVersionConstraint` godoc ("the check is operator-level only"), and the CLI reference note. Tightening it would mean enforcing a stricter grammar in config than `aicr verify` itself accepts, which `pkg/config` deliberately avoids ("there is no parallel parser to maintain").

pkg/cli/bundle_verify.go
```diff
@@ -134,19 +138,32 @@ func runBundleVerifyCmd(ctx context.Context, cmd *cli.Command) error {
...
 	if identityRegexp != "" {
 		if validErr := verifier.ValidateIdentityPattern(identityRegexp); validErr != nil {
 			return validErr
 		}
 		verifyOpts.CertificateIdentityRegexp = identityRegexp
 	}
-	verifyOpts.Key = cmd.String("key")
-	verifyOpts.TrustRoot = cmd.String("trust-root")
+	verifyOpts.Key = stringFlagOrConfig(cmd, "key", resolved.Key)
+	verifyOpts.TrustRoot = stringFlagOrConfig(cmd, "trust-root", resolved.TrustRoot)
```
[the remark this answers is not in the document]
lockwobr: The underlying concern is real, but it is not specific to `verify` and is not introduced by this PR, so I am not fixing it here.

`config.Load` has accepted `http://` and `https://` since the `--config` flag shipped (`pkg/config/loader.go`, the `strings.HasPrefix(source, "http://")` branch). Every `--config`-aware command shares it. `spec.bundle.attestation` already carries `signingKey`, `fulcioURL`, and `rekorURL` over that same path, and a tampered `signingKey` pointing at an attacker-controlled KMS is at least as damaging as a lowered `minTrustLevel`. `spec.validate.evidence.attestation.push` is in the same position.

Hardening only `aicr verify` would leave the larger hole open while creating an inconsistency where one command rejects a source the other four accept, which is the kind of asymmetry that gets rediscovered as a bug later.

The right fix is at the loader, for all commands at once: reject `http://` outright (or gate it behind an explicit opt-in), and set a `CheckRedirect` that refuses HTTPS-to-HTTP downgrades. That is a deliberate behavior change to a shared, already-documented surface, and it deserves its own PR with its own tests rather than riding along in a schema addition.

Happy to file that as a follow-up issue and link it here. Flagging one nuance for whoever picks it up: `readSource` gates the scheme, but the redirect policy is the default one, so an `https://` source that 302s to `http://` is currently followed. That is the sharper half of this finding and worth calling out explicitly in the issue.

pkg/bundler/verifier/trust.go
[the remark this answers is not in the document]
lockwobr: You are right and I was wrong. Retracting my previous reply and fixing this in 586356d5.

The bare-`!` rejection is real, at `pkg/constraints/constraint.go:75-78`. Two mistakes on my side:

1. I read `constraint.go` **before** rebasing this branch. The rebase pulled in four new `main` commits, one of which added that check. My claim that "no such rejection exists" described a version of the file that this branch no longer sits on.
2. My probe was methodologically useless for this question. I exercised `ParseVersionConstraint("!0.8.0")`, which prepends `>=` as its first act, so it could never reach the bare-`!` branch. I should have called `constraints.ParseConstraintExpression("!0.8.0")` directly. Verifying through the wrapper I was asking about was circular.

Your original point stands exactly as written: prepending `>=` hides invalid syntax from the shared parser, so `!0.8.0` passed load validation and failed only at evaluate time.

Fix skips the bare-version default for `!`-prefixed input so it reaches the shared parser and is rejected there:

```go
if !hasOperatorPrefix(expr) && !strings.HasPrefix(expr, "!") {
    expr = ">= " + expr
}
```

Added the regression case you asked for, `bare ! prefix rejected, not coerced to >=`, confirmed failing before the fix and passing after. `!= 0.8.0` still parses, and the bare-version default is unaffected.

pkg/config/doc.go
```diff
@@ -13,25 +13,30 @@
...
 // Secrets (notably the cosign identity token) are not part of the schema;
 // they must be supplied via environment variables or dedicated CLI flags.
 // Durable, non-secret references are in scope by contrast: the private
-// Sigstore endpoints (spec.bundle.attestation.fulcioURL / rekorURL) and the
-// KMS signing-key reference (spec.bundle.attestation.signingKey) all belong
-// in version-controlled config even though they configure signing.
+// Sigstore endpoints (spec.bundle.attestation.fulcioURL / rekorURL), the
+// KMS signing-key reference (spec.bundle.attestation.signingKey), and their
+// verify-side counterparts (spec.verify.trust.key / trustRoot) all belong in
+// version-controlled config even though they configure signing.
```
njhensley: **🔵 Nitpick — doc.go: "even though they configure signing" is inaccurate for the verify-side references**

The sentence lumps `spec.verify.trust.key / trustRoot` into the list closing with "...all belong in version-controlled config even though they configure signing." The verify-side key/trustRoot configure signature *verification*, not signing.

**Fix:** End the clause with "...even though they configure signing or its verification." (or split the clause). Godoc-only, no runtime impact.

pkg/cli/bundle_verify.go
```diff
@@ -167,9 +184,9 @@ func runBundleVerifyCmd(ctx context.Context, cmd *cli.Command) error {
 
 	// Check policy requirements
 	policy := verifier.Policy{
-		MinTrustLevel:     cmd.String("min-trust-level"),
-		RequireCreator:    cmd.String("require-creator"),
-		VersionConstraint: cmd.String("cli-version-constraint"),
+		MinTrustLevel:     stringFlagOrConfig(cmd, "min-trust-level", resolved.MinTrustLevel),
```
njhensley, line 187: **🔵 Nitpick — A one-line doc note that minTrustLevel is operator policy, not an enforced floor**

A committed `spec.verify.policy.minTrustLevel: unknown` silently lowers the trust floor from the `max` default (CheckPolicy's `else if p.MinTrustLevel != ""` branch + TrustUnknown ordering make the floor a no-op), and stringFlagOrConfig emits no INFO log when *config* (vs a flag) lowers it. Blast radius is bounded: the independent `result.Errors` gate (bundle_verify.go:212-214, ErrCodeUnauthorized) still rejects checksum-failed / invalidly-signed bundles regardless of policy, so only checksum-valid *unsigned* bundles slip through. This is a deliberate, reviewable, version-controlled operator choice.

**Fix:** Optional: add a sentence to the VerifyPolicySpec.MinTrustLevel godoc / cli-config.md noting minTrustLevel sets operator policy and can lower the effective floor (it is not an org-enforced guardrail), and consider an INFO log when config lowers the floor below `max`. Not blocking.

pkg/bundler/verifier/trust_test.go
```diff
@@ -243,3 +243,61 @@ func TestGetTrustLevels(t *testing.T) {
 		}
 	}
 }
+
+// TestParseVersionConstraint covers the grammar shared by the CLI flag and
+// spec.verify.policy.cliVersionConstraint. The bare-version default is the
+// behavior worth pinning: "0.8.0" must mean ">= 0.8.0", not "== 0.8.0".
+func TestParseVersionConstraint(t *testing.T) {
```
njhensley, line 250: **🔵 Nitpick — TestParseVersionConstraint table omits the `>` and `<=` operators**

The dedicated table added by this PR to pin the shared grammar covers >=, ==, !=, <, bare, whitespace, empty, `>=`-no-version, and `!`-prefix, but not `>` or `<=` (both handled by hasOperatorPrefix). They are only exercised transitively elsewhere.

**Fix:** Add two rows, e.g. `{expr: "> 0.8.0", satisfied: "0.8.1", rejected: "0.8.0"}` and `{expr: "<= 0.8.0", satisfied: "0.8.0", rejected: "0.9.0"}`.

pkg/cli/bundle_verify_config_test.go
```diff
@@ -0,0 +1,155 @@
+// Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
njhensley, line 1: **🔵 Nitpick — No command-level test for --min-trust-level overriding config (regression-guard gap)**

The override branch of stringFlagOrConfig is command-tested only for `--key`; there is no verify-command test that `--min-trust-level` overrides a config value, nor that a config with no policy still enforces the `max` default. If the wiring ever regressed to pass `""`, CheckPolicy would skip the trust check entirely (fail-open). Not a live defect (the flag has a compile-time `Value:"max"` default and stringFlagOrConfig is unit-tested), just an untested regression guard.

**Fix:** Add: config `minTrustLevel: verified` + `--min-trust-level unknown` → passes (flag wins); and a policy-less config against writeVerifiableKeyFixture asserting the `max` floor fails an unsigned bundle.

pkg/config/config.go
```diff
@@ -342,3 +347,69 @@ type ValidateExecutionSpec struct {
...
+// *policy* — a trust floor, a pinned creator identity, a KMS key reference,
+// a trusted-root path — so it belongs in the same version-controlled
+// document as the producer sections rather than being retyped on each
+// invocation. See issue #1567.
+//
+// The two sub-sections mirror how `aicr verify` consumes them: Policy feeds
+// verifier.Policy (assertions checked after verification runs) and Trust
+// feeds verifier.VerifyOptions (the material verification runs against).
+//
+// Three `aicr verify` flags are deliberately absent:
```
njhensley, line 362: **🔵 Nitpick — tlog-exclusion invariant has no dedicated test (relies on generic strict-decode)**

The security-relevant claim that a config file can never disable the transparency-log check rests on strict decoding rejecting an unknown `trust.ignoreTlog` field. That is genuinely enforced (loader DisallowUnknownFields / KnownFields(true)) and covered by the generic TestLoad_RejectsUnknownYAMLField tests — so this is belt-and-suspenders, not a missing safety net.

**Fix:** Optional: a named test asserting `spec.verify.trust.ignoreTlog: true` fails to load, to lock the flag-only design against a future schema addition.

pkg/cli/bundle_verify_config_test.go
```diff
@@ -0,0 +1,155 @@
...
+		combined += " " + err.Error()
+	}
+	return combined
+}
+
+// TestBundleVerifyCmd_ConfigKeyReachesKeyVerification proves
+// spec.verify.trust.key is plumbed into VerifyOptions.Key. It reuses the
+// --key plumbing signal: a nonexistent PEM path drives the key branch to a
+// public-key resolution error that the keyless path does not produce.
+func TestBundleVerifyCmd_ConfigKeyReachesKeyVerification(t *testing.T) {
```
njhensley, line 62: **🔵 Nitpick — Config-key test uses a weaker discriminator than its sibling**

TestBundleVerifyCmd_ConfigKeyReachesKeyVerification asserts output contains "public key" to prove the config key routed to the key path; the sibling override test uses the stronger path-name discriminator (from-config.pem vs from-flag.pem). If the keyless path ever emits "public key", this test passes vacuously.

**Fix:** Assert the config-supplied path name (`from-config.pem`) reaches the error, matching the override test's stronger pattern.

njhensley (dismissed):
## 📋 Multi-Persona Review

**Method:** 5 independent persona reviewers (Correctness · Security/Supply-chain · API-contract · Test-coverage · Docs) fanned out in parallel, then every candidate finding was re-derived from the resolved code by an adversarial senior meta-reviewer. Pinned to head `586356d5`.

### Overall assessment

Clean, high-quality, additive PR — approving. The `spec.verify` schema is well-designed, the threat model is deliberate and correctly enforced, and the tests are unusually rigorous (test-first, assertions chosen to avoid vacuous passes, the `!`-coercion fix directly pinned). Nothing survived meta-review above 🔵 Nitpick.

The security story holds end-to-end:
- **`--insecure-ignore-tlog` is genuinely config-unreachable** — read only from `cmd.Bool` (`bundle_verify.go:167`), no schema field, and strict decoding would *reject* a smuggled `ignoreTlog` key. Fail-closed at both CLI and library layers.
- **Malformed policy fails closed** — `config.Load → Validate → Resolve` aborts before any verification work, with spec-path attribution. No warn-and-continue.
- **Trust material is scoped** — `key` restricted to whitelisted KMS schemes / local PEM, `trustRoot` a bounded local read, config sources reject `cm://` and `file://`.

One correction worth recording: a lowered `minTrustLevel` cannot admit *corrupt* bundles — the independent `result.Errors` gate (`bundle_verify.go:212`) rejects checksum/attestation failures regardless of policy, so it admits only checksum-valid *unsigned* bundles.

### Confirmed non-issues (examined, refuted)

- **`certificateIdentityRegexp` permissiveness** — a permissive config regexp still requires the GitHub-Actions issuer pin (`NewShortCertificateIdentity(TrustedOIDCIssuer,…)`) + public-good root; identical to the pre-existing `--certificate-identity-regexp` flag, not net-new risk.
- **Precedence / nil-safety / `!`-coercion / double-resolve** — all traced clean.
- **Docs** — schema/enum/operator accuracy, the `#verify-config-file-mode` anchor, and the "four→five" count audit across the tree all verified correct.
- **OpenAPI** — `AICRConfig` is intentionally outside the HTTP contract; `spec.verify` correctly needs no `server.yaml` entry.
- **Tests pass:** `pkg/config` 93.5%, `pkg/bundler/verifier` 76.5%, `pkg/cli` 74.8%; every new exported symbol at 100% per-func coverage.

### Summary

|-----------|---------|---------|-----------|
| 0 | 0 | 0 | 6 |

**Recommendation: ✅ Approve with comments.** None blocking — the doc-wording nit and the two missing parser-table rows are the only ones I'd bother with before merge; the rest are optional hardening.

mchmarny (dismissed):
**APPROVE** - 0 blockers at head `586356d`.

Reviewed the `spec.verify` schema, its `Resolve`/`validate` wiring, the CLI plumbing in `bundle_verify.go`, and both doc pages. Full CI is green at this head (tests, lint, Tier 1 KWOK matrix, KMS and Sigstore E2E lanes).

What I checked beyond the diff:

- `stringFlagOrConfig` gates on `cmd.IsSet`, so the `--min-trust-level` flag default (`max`) does not shadow a config-supplied level. The command-level tests confirm it end to end.
- The `--insecure-ignore-tlog` precondition reads the *resolved* key, so a config-supplied `trust.key` satisfies it instead of being rejected as if `--key` were missing.
- Adding `regexp.Compile` to `ValidateIdentityPattern` also tightens `pkg/server/signing.go`, `pkg/recipe/catalog/verify.go`, and `pkg/bundler/bundler.go`. All three fail closed on input that would have failed later anyway, so it is strictly better.
- No other surface enumerates the section set. The "four sections" wording in `pkg/config/config.go`, `pkg/config/doc.go`, and `docs/user/cli-config.md` is updated everywhere it appears.
- `pkg/config` importing `pkg/bundler/verifier` does not cycle, and the verifier's Sigstore setup stays lazy.

Leaving @njhensley's nitpicks as-is; none of them block.

pkg/config/doc.go
lockwobr: Good catch, fixed in 77a554eb. The clause now ends "...even though they configure signing or its verification." You are right that `trust.key` / `trustRoot` configure verification, not signing; I extended that sentence when adding the verify-side references and did not revisit its tail.

pkg/bundler/verifier/trust_test.go
lockwobr: Fixed in 77a554eb. Added both rows as suggested:

```go
{name: "explicit >",  expr: "> 0.8.0",  satisfied: "0.8.1", rejected: "0.8.0"},
{name: "explicit <=", expr: "<= 0.8.0", satisfied: "0.8.0", rejected: "0.9.0"},
```

Worth pinning explicitly given the point of that table is to lock the shared grammar; leaving two of the six operators to transitive coverage undercut it.

pkg/cli/bundle_verify_config_test.go
lockwobr: Fixed in 77a554eb. Now asserts `from-config.pem` reaches the error, matching the override test's discriminator. You are right that the weaker form could pass vacuously if the keyless path ever emitted "public key" — the assertion would then no longer prove the config key routed anywhere.

pkg/cli/bundle_verify.go
lockwobr: Took both halves of this in 77a554eb, and thank you for the precise blast-radius analysis — it is exactly right and I have folded it into the docs rather than leaving it in the review thread.

**Docs.** `VerifyPolicySpec.MinTrustLevel` godoc and `docs/user/cli-config.md` now state that this is operator policy rather than an org-enforced guardrail, that a committed value lowers the floor as readily as it raises it (`unknown` makes the check a no-op since every level meets it), and that the bounded blast radius comes from the independent `result.Errors` gate still rejecting checksum-failed and invalidly-signed bundles, so only checksum-valid *unsigned* bundles pass a lowered floor.

**The log.** Added, because your observation about `stringFlagOrConfig` is the crux: it only logs when a *flag* overrides config, so config silently lowering the floor was the one path with no audit line at all.

```go
if !cmd.IsSet("min-trust-level") && resolved.MinTrustLevel != "" && resolved.MinTrustLevel != "max" {
    slog.Info("trust floor set by config", "minTrustLevel", resolved.MinTrustLevel,
        "source", "spec.verify.policy.minTrustLevel")
}
```

Gated on `!= "max"` so restating the default stays quiet and only a genuine deviation is surfaced.

pkg/cli/bundle_verify_config_test.go
lockwobr: Fixed in 77a554eb. Added both cases you specified:

- `TestBundleVerifyCmd_MinTrustLevelFlagOverridesConfig` — config `verified` + `--min-trust-level unknown` against the unsigned fixture; the flag must win, so the run must not report `does not meet minimum "verified"`.
- `TestBundleVerifyCmd_PolicylessConfigKeepsMaxDefault` — a config with a `trust` section but no `policy` must leave the flag's compile-time `max` default intact, asserted via the `maximum achievable` message the unsigned fixture triggers.

Your framing of the risk is what made this worth adding: it is not a live defect, but the failure mode is **fail-open**. If the wiring ever regressed to passing `""`, `CheckPolicy` would skip the trust check silently and every test asserting a *failure* would still pass. The second test is the one that catches that, and it is the case I would not have thought to write.

pkg/config/config.go
lockwobr: Added in 77a554eb as `TestVerifyTrustRejectsIgnoreTlogField`, which writes a document carrying `spec.verify.trust.ignoreTlog: true` and asserts `config.Load` rejects it by name.

Agreed it is belt-and-suspenders today — strict decoding genuinely enforces it and the generic unknown-field tests cover the mechanism. The reason I took it anyway is the one you named: the invariant is a *design* promise ("a committed file can never disable transparency-log verification") that currently holds only as a side effect of the field's absence. A future schema addition would not trip any existing test. A named test states the intent, so anyone who later adds that field has to consciously delete an assertion that says why it must not exist.

docs/user/cli-config.md
```diff
@@ -270,9 +286,40 @@ Inputs to `aicr validate`.
...
+| `trust.key` | string | KMS key URI (`awskms://` \| `gcpkms://` \| `azurekms://` \| `hashivault://`) or local PEM public-key path; the verify counterpart to `spec.bundle.attestation.signingKey` |
+| `trust.trustRoot` | string | Path to a private Sigstore `trusted_root.json`, additive to the built-in public-good root |
+
+`policy.minTrustLevel` sets **operator policy, not an org-enforced guardrail**.
+A committed value lowers the effective floor as readily as it raises it
+(`unknown` makes the trust check a no-op, since every level meets it), so treat
+it as a reviewable choice rather than a control that cannot be relaxed. Its
+blast radius is bounded: verification still rejects checksum-failed and
+invalidly-signed bundles regardless of policy, so only checksum-valid *unsigned*
+bundles pass a lowered floor. `aicr verify` logs at INFO whenever config sets the
```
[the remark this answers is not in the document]
lockwobr: Valid, and this is the good kind of catch because my wording erred toward *understating* risk in a security-relevant doc. Fixed in c535d8db.

I traced it rather than taking it on faith, and the mechanism is concrete. `verifier.go:463-465` sets `TrustAttested` with reason "bundle attested but binary attestation not found (incomplete chain)", while `MaxAchievableTrustLevel` returns `TrustVerified` for an attested bundle with no external data. So that bundle is **signed**, reports `attested` against a `verified` maximum, records nothing in `result.Errors`, and is therefore rejected by the `max` default but admitted by any lowered floor. External `--data` produces the same shape at `verifier.go:435-437`.

So "only checksum-valid unsigned bundles" was wrong on two counts: the admitted set is not limited to unsigned bundles, and its most interesting members are signed ones whose chain legitimately degraded.

Both the `VerifyPolicySpec.MinTrustLevel` godoc and `docs/user/cli-config.md` now say a lowered floor admits any bundle whose actual level reaches it, name the degraded-chain cases explicitly, and separate that from what no policy value can wave through (checksum failures and attestations present but failing verification, which are gated independently via `result.Errors`).

`@njhensley` — flagging since this refines the blast-radius framing from your review comment, which I had propagated into the docs verbatim.

docs/user/cli-config.md
```diff
@@ -266,13 +282,51 @@ Inputs to `aicr validate`.
...
+A lowered floor admits any bundle whose actual trust level reaches it, which is
+broader than unsigned bundles. It also covers chains that legitimately degraded:
+an attested bundle whose binary attestation is absent, or one carrying external
+`--data`, both report `attested` against a `verified` maximum, so the default
+`max` rejects them while a lowered floor does not.
+
+What no policy value can wave through: checksum failures, and attestations that
+are present but fail verification. Those are rejected regardless of the floor.
+`aicr verify` logs at INFO whenever config sets the floor to anything other than
+`max`.
```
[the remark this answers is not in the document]
lockwobr: Valid, fixed in 6e3e5f76. The guard is:

```go
if !cmd.IsSet("min-trust-level") && resolved.MinTrustLevel != "" && resolved.MinTrustLevel != "max" {
```

so "logs at INFO whenever config sets the floor" overstated it — an explicit `--min-trust-level` suppresses that path entirely.

Both the godoc and `docs/user/cli-config.md` now say the floor is logged when *config is what supplies it*, meaning `--min-trust-level` is absent and the configured value is anything other than `max`. I also added the half that makes it actionable rather than just narrower: when a flag does override, that override is still logged, by the flag-precedence path in `stringFlagOrConfig`. So there is no case where the effective floor goes unlogged; the two paths just report it differently, and the docs now say which is which.

mchmarny (approved):

njhensley (approved):

Issue #2041, opened 2026-08-04
chainsaw: positive list-match loop lacks the negative loop's ghost filter


mchmarny:
## Summary

`evaluateAssert`'s list-and-match loop returns success on the first matching item with no `isTerminatingOrLost` filter, while `evaluateError`'s loop explicitly skips such ghosts, citing the orphan-pod trap. The asymmetry means a positive assertion can be satisfied by a pod that is Terminating or on a lost node.

Reported by @yuanchen8911 in https://github.com/NVIDIA/aicr/pull/2032#issuecomment-5182383693 (item 2), scoped there as investigate-then-fix.

## Current behavior

Positive path, `pkg/chainsaw/inprocess.go:373-389` — iterates items, returns nil on the first shape match.

Negative path, `pkg/chainsaw/inprocess.go:445-458`:

```go
// Skip ghosts (Terminating or NodeLost): a node replaced or lost
// mid-run leaves its DaemonSet/workload pod behind for minutes until
// pod GC, while a healthy replacement already runs on the live node.
if isTerminatingOrLost(actual) {
    continue
}
```

The mechanism is real — this exact trap turned a healthy `ebs-csi-node` DaemonSet into a false failure — but on the positive side it fails in the *unsafe* direction: a ghost that still carries a Ready condition would let the check report Pass while the live replacement is not ready.

## Scope

1. Establish whether a supported flow can reach it. Current positive list assertions target four Deployments, and validation runs post-deploy rather than mid-teardown, so reachability is unproven.
2. If reachable — or if we decide symmetry is worth having regardless, given a false PASS is the dangerous direction — add the filter. Note the semantics that must come with it: when filtering leaves zero items, the result has to be the `ErrCodeNotFound` "no resources found" path, not a pass, or the fix trades one false PASS for another.

This becomes more load-bearing once the readiness gate runs through the same executor (#2038): `pkg/chainsaw` stops being validator-only and becomes the single assertion engine for both callers.

## Acceptance criteria

- A documented reachability verdict, recorded in the PR or here.
- If the filter lands: an all-ghosts item set fails as not-found rather than passing, covered by a test.

Issue #2040, opened 2026-08-04
chainsaw: IsChainsawTest substring detection fails in both directions


mchmarny:
## Summary

`IsChainsawTest` decides whether content dispatches to the allowlist-guarded in-process Test executor or to the raw-resource assertion path, using two substring checks. It is wrong in both directions, and the false-positive direction produces a vacuous PASS.

Reported by @yuanchen8911 in https://github.com/NVIDIA/aicr/pull/2032#issuecomment-5182383693 (item 3).

## Current behavior

`pkg/chainsaw/runner.go:133-135`:

```go
func IsChainsawTest(raw string) bool {
    return strings.Contains(raw, "chainsaw.kyverno.io") && strings.Contains(raw, "kind: Test")
}
```

**False negative.** A quoted `kind: "Test"` is not recognized and falls through to the raw-resource path. Fails closed, and all 37 current registry checks use the unquoted form.

**False positive — this is the dangerous one.** Raw Kubernetes YAML that happens to carry both strings (in a comment, an annotation, or ConfigMap data) is dispatched as a Test. `runChainsawTestInProcess` unmarshals it into `v1alpha1.Test`, gets an empty `Spec.Steps`, runs zero assertions, and falls through to `result.Passed = true` (`pkg/chainsaw/inprocess.go:96-120`). `ValidateTestReadOnly` does not catch it either — an empty step list has no operations to reject.

So a health check that asserts nothing reports healthy.

## Proposed change

- Parse the document's `apiVersion` and `kind` rather than substring-matching, so detection follows YAML semantics.
- Reject a dispatched Test that carries no executable `assert` / `error` operation, with `ErrCodeInvalidRequest`. A check that evaluates nothing must never report Pass.

## Acceptance criteria

- `kind: "Test"` (quoted) is detected as Test format.
- Raw K8s YAML mentioning `chainsaw.kyverno.io` and `kind: Test` in a comment or in data is **not** dispatched as a Test.
- A Test with zero steps, or with steps carrying no assert/error op, fails with `ErrCodeInvalidRequest` instead of passing.

mchmarny:
Triaged: P2→P1. This bug is actively being fixed in #2045; substring dispatch can turn raw YAML into an empty Chainsaw Test and report a vacuous PASS.

Issue #2039, opened 2026-08-04
chainsaw: List collapses all API errors to ErrCodeInternal, mis-driving the absent-resource grace


mchmarny:
## Summary

`clusterFetcher.List` wraps every `resource.List` failure as `ErrCodeInternal`, while `Fetch` carefully distinguishes NotFound (`ErrCodeNotFound`) from any other API failure (`ErrCodeUnavailable`). `resourceObservedErr` treats `ErrCodeInternal` as proof the resource was observed, so a transient API failure during a label-selector List latches the absent-resource grace off for list-based asserts.

Reported by @yuanchen8911 in https://github.com/NVIDIA/aicr/pull/2032#issuecomment-5182383693 (item 1).

## Current behavior

`pkg/chainsaw/fetch.go:115-119`:

```go
list, err := resource.List(ctx, opts)
if err != nil {
    return nil, errors.Wrap(errors.ErrCodeInternal,
        fmt.Sprintf("failed to list %s in namespace %q", gvk, namespace), err)
}
```

`pkg/chainsaw/inprocess.go:294`:

```go
func resourceObservedErr(err error) bool {
    return stderrors.Is(err, errors.New(errors.ErrCodeInternal, ""))
}
```

`Fetch` already documents and implements the intended mapping (`fetch.go:60-74`), and both `evaluateAssert` and `assertSingleDocument` propagate fetcher codes as-is precisely so the grace latch can tell "exists but not ready" from "absent / transient".

## Failure scenario

A list-based assert targets a kind whose resources do not exist yet. The apiserver returns a 503 on the first List. The error carries `ErrCodeInternal`, so `sawResource` latches true, `notFoundGraceDeadline` returns the full deadline instead of `AbsentResourceGracePeriod`, and the check holds a worker slot for the full readiness budget rather than failing fast — under exactly the flaky conditions the grace was added to bound.

Effect is a mis-driven retry deadline, not a wrong verdict.

## Proposed change

Mirror `Fetch`'s mapping in `List`: `apierrors.IsNotFound` → `ErrCodeNotFound`, everything else → `ErrCodeUnavailable`. Reserve `ErrCodeInternal` for a genuine shape mismatch, which is the contract the retry loops already assume.

## Acceptance criteria

- A stubbed transient List failure does not set `sawResource`, and the absent-resource grace still bounds the retry.
- Table test covers NotFound, transient (503), and success.

mchmarny:
[...]

Issue #2038, opened 2026-08-04
gate: run readiness checks in-process; drop the embedded chainsaw binary


mchmarny:
## Summary

The `aicr-gate` image still ships and shells out to the upstream `kyverno/chainsaw` binary, even though AICR has executed the same Chainsaw `Test` format in-process since #1236. That binary is the sole source of the two HIGH findings the weekly image scan reports on `aicr-gate` and of all 47 `pkg:oci/aicr-gate` OpenVEX statements (of 81 total in `.openvex.json`).

#2032 promoted `validators/chainsaw` to `pkg/chainsaw` so the gate binary can depend on the in-process executor without depending on validator code. This issue is the follow-up that PR set up.

## Current state

- `pkg/chainsawgate/runner/runner.go:195` — `exec.CommandContext(tctx, "chainsaw", args...)`.
- `pkg/chainsawgate/runner/runner.go:140` — `Evaluate` materializes each component's test YAML into a temp dir purely to hand a path to the binary.
- `cmd/gate/Dockerfile:31,52` — `FROM ghcr.io/kyverno/chainsaw:${CHAINSAW_VERSION}` plus `COPY --from=chainsaw /ko-app/chainsaw /usr/local/bin/chainsaw`.
- `cmd/gate/Dockerfile:57` — bakes `cmd/gate/chainsaw-config.yaml` to `/etc/chainsaw/config.yaml`, surfaced as the `--chainsaw-config` flag (`cmd/gate/main.go:63,89`). It exists to pin `cleanup.skipDelete` so the binary cannot delete resources — a constraint the in-process executor satisfies structurally, since its allowlist admits only `assert` and `error`.

Upstream cannot fix the CVEs: chainsaw v0.2.15 is the latest release, and chainsaw `main` still sits at x/net 0.53.0 / x/text 0.36.0, below the 0.56.0 / 0.39.0 fixes.

## Why this is safe

Every operation in `recipes/checks/` is `assert:` (111) or `error:` (129) — zero `apply`, `script`, `wait`, or `proxy`. That is exactly the subset `pkg/chainsaw`'s in-process executor implements and its read-only allowlist (`ValidateTestReadOnly`) permits.

## Proposed change

- Point `runner.RunComponent` at `chainsaw.Run`; drop the temp-dir materialization and the `ConfigPath` option.
[...]

PR #2052, opened 2026-08-05
fix(gate): close residual fail-closed and execution-parity gaps

### LINKED ISSUE #2051: gate/validator: close residual fail-closed and execution-parity gaps after #2045
## Problem

Since #2045, the readiness gate and the deployment validator share the in-process assertion executor in `pkg/chainsaw`. This is a **tracking issue** for the remaining assertion-completeness, discovery-classification, and execution-parity gaps found in a post-merge multi-reviewer review of that PR (head `3171b594`, merged as `b7c83bbb`). It may be completed by multiple PRs.

For assertion evaluation, the invariant is:

- The executor must not report Pass unless every authored document and operation was evaluated.
- Ambiguous cluster state must remain Unavailable during evaluation and produce Unknown if the evaluation budget expires. It must never be classified as NotFound or Pass.

The embedded catalog does not currently exercise the false-pass paths: its 134 `error:` operations are unnamed, list-form `v1/Pod` checks, and it contains none of the malformed document shapes below. Items 1–7 remain reachable through supported integrator-authored or local gate inputs. Item 8 can also affect embedded checks during an API or discovery outage. Item 9 independently affects documented local gate execution.

## Fail-open: discovery classification

- [ ] **1. Partial discovery failures classify as NotFound.** `GetAPIGroupResources` (vendored client-go `restmapper/discovery.go:147-152`) drops `ErrGroupDiscoveryFailed` whenever partial results exist, so a kind in the failed group surfaces from the deferred mapper as a bare `NoKindMatchError`. `isGenuineNoMatch` (`pkg/chainsaw/fetch.go:187-194`) then classifies it `ErrCodeNotFound` — the pass path for a named negative assertion. The existing unit test constructs an error shape the production mapper never produces. Fix: classify as `ErrCodeUnavailable`; test partial discovery through the real `DeferredDiscoveryRESTMapper` + memcache client, not a hand-wrapped error.
- [ ] **2. Cooldown-denied refresh accepts a stale no-match.** `resolveMapping` (`pkg/chainsaw/fetch.go:159-176`) retries after a denied refresh against the same un-reset delegate, so a CRD installed within `DiscoveryRefreshCooldown` (60s) reads as absent while the bundler-emitted `--stability-window=30s` can latch Ready before rediscovery. Fix: a cooldown-denied unresolved mapping is `ErrCodeUnavailable`, not NotFound. Acceptance criteria: a stale-delegate test that heals only after `Reset()`, a race-safe mapper test double, and concurrency coverage of the single-flight cooldown (`ChainsawMaxParallel` goroutines share one fetcher).

## Fail-open: malformed authoring (reject, fail closed)

- [ ] **3. An operation carrying both `assert:` and `error:` silently skips the error check.** The allowlist (`pkg/chainsaw/allowlist.go:180`) accepts either field non-nil; the executor's switch evaluates `Assert` first and never reaches `Error`. Chainsaw defines one action per operation — reject the combined shape.
- [ ] **4. Mixed Test/non-Test YAML streams silently drop the non-Test documents.** `decodeTests` (`pkg/chainsaw/inprocess.go:229-236`) skips documents failing the header decode or kind match with no error; `IsChainsawTest` routes the whole stream in-process when any document is a Test, so the dropped documents are evaluated by no path. Reject mixed streams.
- [ ] **5. Empty or whitespace-only bundle entries pass vacuously.** Content with no recognizable Test routes to `assertRawResources` (`pkg/chainsaw/runner.go:184-206`), which returns `Passed=true` for zero documents — bypassing the #2040 no-op guard, which covers only decoded Test documents. Reject empty entries (the `aicr/no-op-check` annotation remains the sanctioned way to declare an intentional no-op).

## Execution parity and consistency

- [ ] **6. Named/list asymmetry on NotFound for negative assertions.** The named branch of `evaluateError` treats `ErrCodeNotFound` as pass; the list branch propagates it into the retry loop until max-wait. Desired contract, stated explicitly: a genuine `ErrCodeNotFound` from either `Fetch` or `List` satisfies a negative assertion; `ErrCodeUnavailable` never does. (The embedded catalog does not normally reach the collection-NotFound path — all list-form negatives are core `v1/Pod`.)
- [ ] **7. `spec.timeouts.assert` replaces the gate `--timeout` instead of being capped by it.** The removed exec path wrapped the chainsaw subprocess in an independent outer `context.WithTimeout`; in-process (`pkg/chainsaw/inprocess.go:131-140`) the authored timeout fully substitutes the caller budget, and `--max-wait` is only checked between evaluations. Fix: effective timeout = min(authored assertion timeout, gate's remaining outer budget).
- [ ] **8. Budget exhaustion on persistent Unavailable reports Fail instead of Unknown.** `isBudgetExhausted` (`pkg/chainsawgate/runner/runner.go:233-236`) matches only context sentinels, but the retry loop returns the last substantive error at deadline, so a discovery/API outage that exhausts the budget labels a never-observed component Fail. Exit codes are unaffected (Fail and Unknown gate identically); this misdirects operator triage — and can affect embedded checks during an outage.
- [ ] **9. Multi-file `KUBECONFIG` regression for documented local gate execution.** `newClusterFetcher` → `GetKubeClient` → `ResolveKubeconfigPath` returns `""` for colon/semicolon-separated values, and `BuildKubeClient` then goes in-cluster, so a local run with `KUBECONFIG=/a:/b` exits 2. The replaced chainsaw binary used clientcmd's merging loading rules; `cmd/gate/main.go:16` documents local KUBECONFIG execution and there is no `--kubeconfig` override.

## Non-blocking cleanup (batch into any of the above PRs)

- [ ] Stale "embedded chainsaw" comments: `pkg/bundler/readiness.go:36`, `.github/workflows/on-tag.yaml:35`.
- [ ] Misleading comment in `validators/deployment/expected_resources.go` (`getDynamicClient`): production `ctx.DynamicClient` is pre-populated unbounded by `LoadContext`, so the bounded constructor is test-only.
- [ ] 50ms assertion budgets in `TestEvaluate_AppliesDefaultNamespace` can flake under load/`-race`.
- [ ] `assertRunningPod` closure in `pkg/chainsaw/inprocess_test.go` captures the parent `*testing.T` and calls `t.Fatalf` from parallel subtests.

## References

- #2045 (source change), #2038 #2039 #2040 #2041 (issues it fixed)
- Review pinned at `3171b594c92549005d660390ce7a1e7eec0163ab`; all finding sites verified byte-identical at the merge commit `b7c83bbb8f9f2284433333559456a8de6983d8e3`.

mchmarny:
## Summary

Closes the nine gaps tracked in #2051 from the post-merge review of #2045, where the readiness gate and the deployment validator started sharing the in-process assertion executor in `pkg/chainsaw`. Each gap let an assertion report `Pass` without having established what it claims to.

## Motivation / Context

#2045 removed the embedded `chainsaw` binary (its only source of unfixable HIGH CVEs) and moved readiness evaluation in-process. The multi-reviewer pass over that PR found residual assertion-completeness, discovery-classification, and execution-parity gaps — none critical, none merge-blocking, all reachable through supported integrator-authored or local gate inputs, and two (items 8 and 9) reachable for embedded checks too.

The invariant the issue states is the one this PR restores:

> The executor must not report Pass unless every authored document and operation was evaluated. Ambiguous cluster state must remain Unavailable during evaluation and produce Unknown if the evaluation budget expires. It must never be classified as NotFound or Pass.

Fixes: #2051
Related: #2045, #2038, #2039, #2040, #2041

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `pkg/chainsaw`, `pkg/chainsawgate`, `cmd/gate`

## Implementation Notes

**Discovery classification (items 1–2).** Reaching `ErrCodeNotFound` from `resolveMapping` now requires two conditions, because that code is the immediate-pass path for a negative `error:` assertion:

1. *The retry ran against discovery performed after the lookup began.* A `refreshDiscovery` denied by the 60s `DiscoveryRefreshCooldown` re-read the same un-invalidated delegate, so a CRD installed inside that window read as absent while the bundler-emitted `--stability-window=30s` could latch Ready first. A reset **generation counter** replaces a bare boolean: a caller denied its own refresh still observes the generation advanced by whichever goroutine won, so it does not stall unnecessarily. This is what makes the single-flight cooldown safe across the `ChainsawMaxParallel` goroutines sharing one fetcher.
2. *The kind's own API group was fully enumerated.* `restmapper.GetAPIGroupResources` drops `ErrGroupDiscoveryFailed` whenever partial results exist, so a kind in a failed group arrives as a **bare** `NoKindMatchError` — indistinguishable from a genuine no-match by the error alone. The fetcher now consults the same cached discovery client the mapper resolves through (one `MemCacheClient` feeds both, so a `Reset()` cannot leave the probe reading a stale view). The probe is **scoped to the one group** on purpose: real clusters routinely carry a broken aggregated APIService, and treating any partial failure as global would strand every negative assertion on a permanently degraded cluster.

`NewClusterFetcherWithClient` is the new wiring point; the deployment validator moved onto it so both consumers get identical classification. `NewClusterFetcher` keeps its variadic-compatible signature as the test/injection seam, and its inability to see partial discovery is asserted by a test so removing that wiring is a visible change rather than a silent one.

**Authoring shapes (items 3–5).** Three shapes that evaluated nothing now fail closed. The `assert`+`error` case is rejected in both the allowlist (with step attribution) and the executor (defense in depth), because the executor's switch reaches `Assert` first — the *negative* check is the half that was silently dropped. Genuinely empty YAML documents (trailing `---`, comment-only, explicit `null`) stay punctuation and are skipped; only content-bearing non-Test documents are rejected.

**Execution parity (items 6–8).** The timeout cap restores what the removed exec path enforced by wrapping the subprocess in an independent outer `context.WithTimeout(stepTimeout)`. **No in-tree check changes behavior** — every authored value is at or below its caller's budget — but a readiness test copied from a validator check (`assert: 5m`) is now capped at the gate's 2m per evaluation rather than overrunning it. The `Unknown` verdict keys on `ErrCodeUnavailable`, which is never terminal, so `pkg/chainsaw` retries it for the whole budget and can only surface it once that budget runs out; an `Unavailable` arriving at the gate *is* the "budget expired while still blind" case, however it is spelled.

**Item 9** is a plain regression fix in `pkg/k8s/client`: a multi-file `KUBECONFIG` is a clientcmd loading-rules value, so `ResolveKubeconfigPath` correctly reports no path — but `BuildKubeClient` then fell through to `rest.InClusterConfig()`, failing every local `KUBECONFIG=/a:/b` run with exit 2. It now hands that case to clientcmd's default loading rules, and classifies its failures as caller input rather than internal.

## Testing

```bash
make qualify   # exit 0
```

`make qualify` passes end to end: full `-race` suite, coverage 81.8% (threshold 80%), `go vet`, `golangci-lint` 0 issues, yamllint, e2e, CLI chainsaw suite, security scan, docs/agents-sync checks.

**Every fix is guarded by a test verified to fail without it.** Each source file was reverted to `HEAD` in isolation (or the specific branch stubbed out, where the change also alters an exported signature) and the new tests re-run:

| Item | Guard | Without the fix |
|---|---|---|
| 1 partial discovery | `TestClusterFetcher_PartialDiscoveryClassification` | FAIL |
| 2 cooldown-denied no-match | `TestClusterFetcher_CooldownDeniedNoMatchIsUnavailable` | FAIL |
| 3 `assert`+`error` | `TestCombinedAssertAndErrorIsRejected` (both layers) | FAIL |
| 4 mixed streams | `TestDecodeTests_RejectsMixedStreams` | FAIL |
| 5 empty content | `TestAssertRawResources_EmptyContentIsRejected` | FAIL |
| 6 list NotFound parity | `TestEvaluateError_ListNotFoundSatisfiesNegativeAssertion` | FAIL |
| 7 timeout cap | `TestEffectiveTimeoutIsCappedByCallerBudget` | FAIL (ran 30s against a 300ms budget) |
| 8 Unknown verdict | `TestToComponentResult`, `TestEvaluate_UnreadableClusterIsUnknown` | FAIL |
| 9 multi-file KUBECONFIG | `TestBuildKubeClient_MultiFileKUBECONFIG` | FAIL |

Per the issue's acceptance criteria, the discovery tests drive the **real** `DeferredDiscoveryRESTMapper` over a real `MemCacheClient` instead of a hand-wrapped error the production mapper never produces. `TestPartialDiscovery_ProducesBareNoKindMatch` pins the client-go behavior being defended against, so the guard fails loudly (rather than passing for a new reason) if upstream ever starts propagating `ErrGroupDiscoveryFailed`. The concurrency subtest drives `ChainsawMaxParallel` goroutines through one fetcher with a race-safe mapper double.

Coverage on changed packages:

- `pkg/chainsaw`: 78.7% → 81.1% (+2.4%)
- `pkg/chainsawgate/runner`: 97.9% → 98.0% (+0.1%)
- `pkg/k8s/client`: 85.7% → 90.6% (+4.9%)

All new exported functions are covered (`WithGroupDiscovery` 100%, `NewClusterFetcherWithClient` 75%, `NewClusterFetcher` 100%).

## Risk Assessment

Medium because it changes verdict classification on the readiness gate and the deployment validator, both of which gate deployments. Every change moves in the fail-closed direction, so the failure mode is a check that waits or reports a problem rather than one that passes when it should not.

Two behavior changes worth naming explicitly:

- **A negative assertion on a kind the cluster does not serve may now wait up to one cooldown** (60s, bounded, well inside every budget: 2m gate, 6m validator) when its lookup lands inside a window it did not itself refresh. The first evaluation always refreshes, so the common case is unaffected. The embedded catalog is unaffected either way — all 134 `error:` operations target core `v1/Pod`, which always resolves.
- **An authored `spec.timeouts.assert` larger than the caller budget is now capped.** No in-tree check is affected (all are at or below their caller's budget), but an integrator readiness test copied from a validator check would previously overrun the gate's `--timeout`.

**Rollout notes:** No migration. No API, flag, or manifest changes. `NewClusterFetcher` keeps a source-compatible signature (variadic options), and `NewRESTMapperForConfig` is retained.

## Checklist


yuanchen8911 (commented):
Reviewed at 7938b372d. The intended fixes for all nine #2051 items are present and their mechanisms check out against the base code, but item 9 remains incomplete for normalized multi-file input. Two issues, plus minor notes — posted inline.

Approve once the two items (the multi-file KUBECONFIG whitespace divergence and the integrator-doc caller-budget contradiction) are addressed.

pkg/k8s/client/client.go
```diff
@@ -169,28 +196,59 @@ func BuildKubeClient(kubeconfig string) (*kubernetes.Clientset, *rest.Config, er
...
+	switch {
+	case kubeconfig != "":
 		config, err = clientcmd.BuildConfigFromFlags("", kubeconfig)
 		if err != nil {
 			return nil, nil, errors.WrapWithContext(errors.ErrCodeInvalidRequest, "failed to build kube config", err, map[string]interface{}{
-				"kubeconfig": kubeconfig,
+				kubeconfigErrContextKey: kubeconfig,
 			})
 		}
+	case multiPathKubeconfigEnv() != "":
```
yuanchen8911: Multi-file KUBECONFIG: whitespace handling diverges between AICR's classification and clientcmd's loading.

`multiPathKubeconfigEnv()` trims the KUBECONFIG value before classifying it, and `ResolveKubeconfigPath` trims too, so outer whitespace is supported input elsewhere in this package. But this branch hands clientcmd `NewDefaultClientConfigLoadingRules()`, whose precedence is built from the raw env value, split without trimming.

The failure mode: with one-sided whitespace (e.g. `KUBECONFIG="/a:/b "` or `" /a:/b"`), one filename is mangled, and clientcmd's loading rules skip missing files non-fatally — the merge silently drops that file. If the surviving file supplies a usable context, the run proceeds against the wrong context; otherwise loading fails. With whitespace on both ends, both paths are mangled and config load fails outright. Either way, AICR's own trimmed classification claims this branch handles the value.

Suggested fix: capture the normalized value once and override the loading rules' precedence with it (`loadingRules.Precedence = filepath.SplitList(merged)`), so the raw value no longer controls which filenames are loaded, and reuse the same normalized value for the error context so the recorded kubeconfig source matches what was actually loaded. The existing test passes because it sets a clean value; a subtest with outer whitespace fails on the current head.

pkg/k8s/client/client.go
```diff
@@ -169,28 +196,59 @@ func BuildKubeClient(kubeconfig string) (*kubernetes.Clientset, *rest.Config, er
...
+		// KUBECONFIG names several files, so ResolveKubeconfigPath reports
+		// no single path. Merging them is clientcmd's job: hand it the
+		// default loading rules, which read and merge KUBECONFIG natively.
+		// Without this the empty path fell through to InClusterConfig()
+		// below, so any local run with KUBECONFIG=/a:/b failed with "unable
+		// to load in-cluster configuration" — the `gate` CLI documents
+		// exactly that local-KUBECONFIG invocation, and the chainsaw binary
+		// it replaced used these same loading rules.
+		merged := multiPathKubeconfigEnv()
```
yuanchen8911: Minor: `multiPathKubeconfigEnv()` is read three times in this function (the case guard, here, and the `NewForConfig` error handler). Capturing it once above the switch is cleaner — and is absorbed by the normalization fix suggested above.

pkg/k8s/client/client_test.go
```diff
@@ -375,3 +375,123 @@ func TestGetKubeClient_CallsOnce(t *testing.T) {
...
+}
+
+// TestBuildKubeClient_MultiFileKUBECONFIG covers the local-execution regression
+// behind the `gate` CLI's documented KUBECONFIG invocation. ResolveKubeconfigPath
+// deliberately reports no path for a ':'-separated KUBECONFIG (it is a clientcmd
+// loading-rules value, not a path), and BuildKubeClient then fell through to
+// rest.InClusterConfig() — so any local run with KUBECONFIG=/a:/b failed with
+// "unable to load in-cluster configuration". The chainsaw binary the gate
+// replaced used clientcmd's merging loading rules, so this was a regression.
+func TestBuildKubeClient_MultiFileKUBECONFIG(t *testing.T) {
```
yuanchen8911, line 415: Test hardening for the merge proof: this fixture passes even if only the first file loads, since the first file alone carries a complete context. Making the first file's `current-context` resolvable only through the second file's context definition (with conflicting cluster/user values across the files) would prove both files loaded and that KUBECONFIG precedence held.

docs/integrator/recipe-development.md
```diff
@@ -841,10 +841,12 @@ spec:
...
+When `--readiness-hooks` is set, the bundler wraps this test into a `NNN-<name>-readiness/` folder containing a `Job` that runs the `gate` CLI (`ghcr.io/nvidia/aicr-gate`). The deploy blocks on that Job — via `helm --wait` for the helm deployer (the gate Job is a `post-install,post-upgrade` hook, and `--wait` blocks on hook completion regardless of `--wait-for-jobs`), or via Argo CD's built-in `batch/Job` health on the next sync-wave for the `argocd`/`argocd-helm` deployers. Keep `spec.timeouts.assert` shorter than the gate's per-test timeout so a single poll can't outlast one gate iteration. This is now enforced rather than advisory: the effective budget is the **smaller** of the authored `spec.timeouts.assert` and the caller's per-component budget (the gate's `--timeout`, the validator's catalog timeout), so an authored value larger than the caller allows is capped rather than honored. A shorter authored value still shortens the budget as before. See [Readiness Gates](../user/cli-reference.md#readiness-gates) for the deploy-time behavior.
```
yuanchen8911: This names "the validator's catalog timeout" as the caller budget for the new timeout cap, but `docs/contributor/validator.md` in this same PR correctly states the validator's per-check caller budget is `defaults.ChainsawAssertTimeout`, with the 8m catalog timeout as the outer envelope. Align this page to the contributor page.

pkg/chainsaw/failclosed_test.go
```diff
@@ -0,0 +1,426 @@
...
+	return nil, f.err
+}
+
+// TestEffectiveTimeoutIsCappedByCallerBudget covers the budget-substitution
+// regression. The removed exec path wrapped the chainsaw subprocess in an
+// independent outer context.WithTimeout(stepTimeout), so an authored
+// spec.timeouts.assert larger than the caller's budget was cut off at the
+// caller's bound. In-process the authored value replaced the budget outright,
+// letting authored content overrun the gate's --timeout by any factor.
+func TestEffectiveTimeoutIsCappedByCallerBudget(t *testing.T) {
```
yuanchen8911, line 372: Minor: this asserts wall-clock elapsed under 3s against a 300ms budget. The 10x margin makes flake unlikely; just worth remembering if it ever flakes under `-race` on loaded CI.

pkg/chainsaw/inprocess.go
```diff
@@ -227,13 +252,21 @@ func decodeTests(yamlContent string) ([]v1alpha1.Test, error) {
...
 		if err := yaml.Unmarshal(buf, &header); err != nil {
-			continue // not a Kubernetes-shaped document
+			return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("document %d is not a Kubernetes-shaped mapping; a stream carrying a chainsaw "+
+					"Test may hold only Test documents, because nothing else evaluates the rest", docIdx), err)
 		}
 		if header.Kind != chainsawTestKind {
-			continue
+			return nil, errors.New(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("document %d has kind %q, not %q; a stream carrying a chainsaw Test may hold "+
```
yuanchen8911: Minor: a mapping document with no `kind` alongside a Test is rejected as `document N has kind ""` — correct fail-closed behavior, just slightly opaque wording; fine to leave.

pkg/k8s/client/client.go
mchmarny: Confirmed and fixed. I reproduced it before changing anything: with `KUBECONFIG=" /a:/b"` the run resolved to the **second** file's context (`second.example`) instead of the first — silently targeting a different cluster than the value names, exactly as you described.

Mechanism confirmed in vendor: `NewDefaultClientConfigLoadingRules` splits `os.Getenv(RecommendedConfigPathEnvVar)` verbatim (`loader.go:164-168`), and `Load()` adds a missing file to `missingList` and `continue`s (`loader.go:228-233`), so the mangled entry is dropped non-fatally.

Fix is your suggestion plus per-entry trimming: `loadingRules.Precedence = kubeconfigPrecedence(merged)`, where `kubeconfigPrecedence` splits the trimmed value, trims each entry, and drops empties. Per-entry trimming keeps this branch consistent with `ResolveKubeconfigPath`, which fully trims a single path — so whatever this package classifies as handled is exactly what clientcmd loads. The same normalized value feeds the error context.

`TestBuildKubeClient_MultiFileKUBECONFIGWhitespace` covers no/leading/trailing/both-ends/per-entry whitespace; 3 of the 5 cases fail without the override. `TestKubeconfigPrecedence` pins the normalization.

pkg/k8s/client/client.go
mchmarny: Fixed — read once into `merged` above the switch, used by the branch guard, the `Precedence` override, and the `NewForConfig` error classification.

pkg/k8s/client/client_test.go
mchmarny: Fixed. The fixture is now split so neither file suffices alone: the first names `current-context: merged-ctx` but never defines it (and carries a decoy cluster), the second defines that context plus the cluster/user it points at. The assertion is `merged.example`, reachable only if both files loaded and precedence held.

Verified the guard by pinning `Precedence` to the first file only — the test fails.

docs/integrator/recipe-development.md
mchmarny: Fixed. The page now names `defaults.ReadinessGateExecTimeout` (the gate's `--timeout`, 2m) and `defaults.ChainsawAssertTimeout` (`aicr validate --phase deployment`) as the caller budgets, and states that the `expected-resources` catalog timeout is the outer Job envelope rather than a per-component assertion budget.

Checked against the code rather than the prose: `validators/deployment/expected_resources.go:297` passes `defaults.ChainsawAssertTimeout` to `chainsaw.Run`. The contributor page was already correct, so this was a self-contradiction inside the PR.

pkg/chainsaw/failclosed_test.go
mchmarny: No change. The wall-clock assertion is the only way to observe the cap from outside the executor, and the margin is wider than it looks: without the fix this case does not miss the 3s bound narrowly, it runs the full `AbsentResourceGracePeriod` (30s) against a 300ms budget — 100x, not 10x. The 3s bound sits between the two with an order of magnitude either side.

Noted for triage if it ever does flake under `-race` on loaded CI.

pkg/chainsaw/inprocess.go
mchmarny: Changed anyway — one-liner. A mapping with no `kind` now reports `document 1 has no kind field, not kind "Test"` instead of `has kind ""`. Covered by a new case in `TestDecodeTests_RejectsMixedStreams`.

docs/integrator/recipe-development.md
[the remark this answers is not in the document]
mchmarny: Fixed. The page now names `defaults.ReadinessGateExecTimeout` (gate `--timeout`, 2m) and `defaults.ChainsawAssertTimeout` (deployment validator) as the caller budgets, and states that the `expected-resources` catalog timeout is the outer Job envelope, not a per-component assertion budget.

Confirmed against `validators/deployment/expected_resources.go:297`, which passes `defaults.ChainsawAssertTimeout` to `chainsaw.Run`.

pkg/chainsaw/fetch.go
```diff
@@ -171,9 +252,57 @@ func (f *clusterFetcher) resolveMapping(gvk schema.GroupVersionKind) (*meta.REST
...
+// negative assertion on a permanently degraded cluster. Only a kind whose OWN
+// group could not be enumerated is ambiguous.
+func (f *clusterFetcher) groupDiscoveryFailure(group string) error {
+	if f.discovery == nil {
+		return nil
+	}
+	_, _, err := f.discovery.ServerGroupsAndResources()
+	if err == nil {
+		return nil
+	}
```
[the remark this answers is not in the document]
mchmarny: Threaded. `resolveMapping` now takes `ctx`, `Fetch`/`List` pass their own, and `groupDiscoveryFailure` returns early when the context is already done — as `ErrCodeUnavailable`, so an expired context can never clear a no-match to `NotFound`.

One correction to the rationale: the probe does not normally fan out. It is only reached when the reset generation advanced, which means the mapper's retry just rebuilt the delegate through `GetAPIGroupResources` → `ServerGroupsAndResources`, repopulating the memcache — so the probe reads a warm cache and issues no request. The per-group-version fan-out is a cold-cache path, not the common one.

The bound is still worth having, for the reason your last line gives rather than the throughput one: `DiscoveryInterface` is context-free, so nothing tied that cold path to the assertion deadline. Covered by `TestClusterFetcher_ProbeHonorsContext`, verified to fail without the guard.

pkg/chainsaw/fetch.go
```diff
@@ -171,9 +252,57 @@ func (f *clusterFetcher) resolveMapping(gvk schema.GroupVersionKind) (*meta.REST
...
+		// Discovery failed outright rather than per-group. The kind cannot
+		// be declared absent on the strength of that.
+		return err
+	}
+	for gv, gvErr := range groupErr.Groups {
+		if gv.Group == group {
+			return fmt.Errorf("%s: %w", gv, gvErr)
+		}
+	}
+	return nil
```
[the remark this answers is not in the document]
mchmarny: Fixed — `errors.Wrap(errors.ErrCodeUnavailable, gv.String(), gvErr)`.

The call site still wraps with the same code plus the GVK context. Matching codes is not the double-wrap anti-pattern the guidelines warn about; that one is about overwriting a correct inner code with a wrong outer one.

pkg/k8s/client/client_test.go
```diff
@@ -375,3 +375,123 @@ func TestGetKubeClient_CallsOnce(t *testing.T) {
...
+		name  string
+		value string
+		want  bool
+	}{
+		{"single path", "/home/u/.kube/config", false},
+		{"empty", "", false},
+		{"unix separated", "/a:/b", true},
+		{"windows separated", `C:\a;C:\b`, true},
+		{"trailing separator still names a list", "/a:", true},
+	}
```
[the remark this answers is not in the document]
mchmarny: Fixed. `C:\a;C:\b` contains `:`, so the case passed on the Unix separator alone and proved nothing about `;`. Now `\a;\b`.

pkg/chainsaw/fetch_discovery_test.go
```diff
@@ -0,0 +1,468 @@
...
+	// ConfigMap is absent from the stub's healthy core group, so a live probe
+	// would clear it to NotFound. With the context already done it must not.
+	_, err := f.Fetch(ctx, "v1", "ConfigMap", "ns", "x")
+	var se *errors.StructuredError
+	if !stderrors.As(err, &se) {
+		t.Fatalf("expected StructuredError, got %T %v", err, err)
+	}
+	if se.Code != errors.ErrCodeUnavailable {
+		t.Errorf("code = %q, want %q — an expired context cannot prove a group was enumerated (err=%v)",
+			se.Code, errors.ErrCodeUnavailable, err)
```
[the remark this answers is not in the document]
mchmarny: Fixed in `fd73eff8`. `stderrors.Is(err, errors.New(errors.ErrCodeUnavailable, ""))`, with `errors.ErrCodeUnavailable` used directly in the message — confirmed there is no `errors.Code` helper.

Also applied to the `isUnavailable` helper in the same file, which had the identical shape.

Not swept across the table-driven cases in this file and `fetch_test.go`: those carry the expected code in a `wantCode` field and use it to build the failure message, and the `As` form is the established shape there. Converting them would be unrelated churn to pre-existing tests.

Checked the assertion is still non-vacuous rather than assuming it: removing the context guard in `groupDiscoveryFailure` makes `TestClusterFetcher_ProbeHonorsContext` fail.

pkg/chainsaw/fetch.go
```diff
@@ -171,9 +252,69 @@ func (f *clusterFetcher) resolveMapping(gvk schema.GroupVersionKind) (*meta.REST
...
+		return nil, errors.Wrap(errors.ErrCodeUnavailable,
+			fmt.Sprintf("no REST mapping for %s and discovery could not be refreshed "+
+				"within the %s cooldown; treating as unresolved rather than absent",
+				gvk, defaults.DiscoveryRefreshCooldown), err)
+	}
+
+	if groupErr := f.groupDiscoveryFailure(ctx, gvk.Group); groupErr != nil {
+		return nil, errors.Wrap(errors.ErrCodeUnavailable,
+			fmt.Sprintf("no REST mapping for %s, but discovery for API group %q is incomplete",
+				gvk, gvk.Group), groupErr)
```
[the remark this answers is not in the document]
mchmarny: Not taking this one — the unconditional wrap is deliberate, and I tested the suggestion before declining it.

This is a fail-closed classification boundary. `groupDiscoveryFailure` returning non-nil means exactly one thing: *the group could not be proven enumerated*. The verdict at that point is `ErrCodeUnavailable` **by construction**, and it must not depend on what code the probe's cause happens to carry. `PropagateOrWrap` makes it data-dependent.

That matters because `groupDiscoverer` is an exported injection seam (`WithGroupDiscovery`), so an implementation can return any structured code. I applied your diff and had the probe return a `NotFound`-coded error:

```
err = [NOT_FOUND] context expired before API group discovery could be verified: context canceled
```

`NotFound` is the immediate-pass path for a negative `error:` assertion — that is precisely the fail-open class this PR exists to close. `TestClusterFetcher_ProbeHonorsContext` catches it, which is the test doing its job.

Second cost, visible in that same output: `PropagateOrWrap` returns the inner error as-is for structured causes, so it discards the message at this layer. The GVK and group name — `no REST mapping for %s, but discovery for API group %q is incomplete` — disappear from the two structured paths, which are the common ones. That is the most useful context an operator gets for "why did this assertion not resolve".

On the guidelines: the anti-double-wrap rule is about re-wrapping *and changing the code* (its example overwrites `ErrCodeNotFound` with `ErrCodeInternal`). Here the outer code matches what the structured causes already carry, so nothing is clobbered, and `errors.Wrap` keeps the cause in the `Unwrap` chain — no diagnostic is lost by wrapping, only by propagating.

`PropagateOrWrap` is right where a callee's code should win. It is wrong at a boundary whose whole job is to force one verdict.

lalitadithya (approved):

Issue #746, opened 2026-05-05
[Feature]: Complete OLM component setup, overlays and validations


kaponco:
### Prerequisites

- [x] I searched existing issues

### Feature Summary

Once having the basic OLM support. We want to complete the following:
- Add more components to match the base overlay
- Introduce overlay for CRs
- Add more validators for OCP - exiting or new (if needed)

### Problem/Use Case

The current support for OCP in [PR #726](https://github.com/NVIDIA/aicr/pull/726) is basic. We need to add more OCP features to level it up with rest of the services.

### Proposed Solution

- OCP should match the base overlay as much as possible. Generally what will be omitted is the monitoring part which is part of openshift.
- Setting up an overlay mechanism by allowing to point to an overlay CR yaml. The logic should reuse the current overlay as much as possible
- Check if OLM can support all current command flags. If not fail loudly 

### Success Criteria

- All exiting relevant deployment validators should pass testing that components are deployed correctly
- Test overlay mechanism (for example "training" overlay ) 
- Unit test that the command fails if unsupported flag is given for OLM flow (e.g. bundle)
- All other service must not be impacted by the change

### Component

Recipe engine / data

### Priority

Important (would improve my workflow)

### Are you willing to contribute?

Yes, I can open a PR

kaponco:
Need to add - handling node-selector and tolerations in the CR using templates (for OCP only)

kaponco:
Closing - opened a new and updated issue - #1818

PR #2057, opened 2026-08-05
recipes: add OCP-compatible components for cert-manager, prometheus-adapter, nvidia-dra-driver-gpu, and k8s-nim-operator

### LINKED ISSUE #1818: [Feature]: Add OCP compatible components
### Prerequisites

### Feature Summary

Integrating components to OCP recipes (on top of those added in the first phase)
- cert-manager
- Prometheus adapter 
- Nvidia-dra-driver-gpu
- k8s-nim-operator

### Problem/Use Case

OCP service does not cover enough functionality for dealing with true workloads. Need to expand and support more features both operational, and applicative. So closing the gap where possible - considering:
- Small but meaningful scope of this change
- Openshift requirements (not all components are adapted to openshift)

### Proposed Solution

Register/Overlaying components so they can be deployed and verified in openshift. 
There are few integration options:
- For a matching ocp operator use in-tree solution (as done previously - e.g. nfd)
- Inherit the generic helm repo, given the base image and RBAC rules match the OCP requirements
- Adapter only (e.g. prometheus adapter) using simple manifest (e.g. for OCP given monitoring) 

### Success Criteria

- Support Recipe, Bundling, validation
- Components passed full deploy process and deployment validation in practice
- Not breaking exiting services, validation etc
- Maintain same API
- Automated testing

### Alternatives Considered

_No response_

### Component

Recipe engine / data

### Priority

None

Maintain existing reporting patterns

Yes, I can open a PR

mohityadav8:
## Summary

Adds OpenShift (OCP) support for four components — cert-manager, prometheus-adapter, nvidia-dra-driver-gpu, and k8s-nim-operator — completing phase 2 of OCP coverage.

## Motivation / Context

#746 shipped OCP support for nfd, gpu-operator, and network-operator via an OLM installer + CR pattern. This PR extends that coverage to the remaining four components requested for OCP recipes.

Fixes: #1818
Related: #746

## Type of Change

## Component(s) Affected

## Implementation Notes

Two integration patterns, chosen per component based on OLM/OperatorHub availability:

1. **OLM installer + CR** (`cert-manager-ocp-olm` + `cert-manager-ocp`) — Red Hat publishes a certified cert-manager Operator, so this mirrors the nfd/gpu-operator/network-operator pattern from #746. The CertManager CR is cluster-scoped; operand Deployments always land in a hardcoded `cert-manager` namespace regardless of the operator's own namespace (see comments in `components/cert-manager-ocp/values.yaml`).
2. **Helm-reuse** (`prometheus-adapter-ocp`, `nvidia-dra-driver-gpu-ocp`, `k8s-nim-operator-ocp`) — no OCP operator exists for these, so each reuses the exact same upstream chart/version as its non-OCP counterpart, with OCP-specific manifests (SCC RoleBindings, RBAC, CA bundle injection) layered on top. `nvidia-dra-driver-gpu-ocp` and `k8s-nim-operator-ocp` pin `fullnameOverride` to match their base component's rendered name, so they intentionally reuse the base component's existing `health-check.yaml` instead of duplicating it.
3. `prometheus-adapter-ocp` points at OCP's built-in Thanos Querier instead of deploying `kube-prometheus-stack`, since OCP already runs cluster monitoring — this also means `cert-manager`, `prometheus-adapter`, and `nvidia-dra-driver-gpu` are no longer force-disabled on the `ocp` recipe (their `-ocp` variants replace them).

New `overlays/ocp-inference-nim.yaml` wires `k8s-nim-operator-ocp` with `dependencyRefs: [cert-manager-ocp, gpu-operator-ocp]`, mirroring `h100-eks-ubuntu-inference-nim.yaml` but pointed at the OCP-native dependencies.

**Open items flagged in code comments, not yet verified on a real cluster:**
- `nvidia-dra-driver-gpu-ocp/manifests/scc-rolebinding.yaml` — exact ServiceAccount names (`nvidia-dra-driver-gpu-controller`/`-kubelet-plugin`) inferred from chart naming convention, not confirmed against the actual chart.
- `k8s-nim-operator-ocp/manifests/scc-rolebinding.yaml` — assumes `nonroot-v2` SCC is sufficient for NIM operand pods; may need a broader SCC depending on the specific NIM image.
- `cert-manager-ocp-olm/readiness.yaml` — uses chainsaw's `script`/`outputs` binding to scope the CSV assertion to the Subscription's `installedCSV`; this repo has no existing example of that binding pattern to copy, so the exact syntax needs a `chainsaw test` run to confirm.
- `ocp-inference-nim.yaml`'s `K8s.server.version: ">= 1.34"` constraint (for DRA) needs checking against which OCP minor version actually ships that Kubernetes version.

## Testing

```bash
yamllint -c .yamllint.yaml recipes/registry.yaml recipes/overlays/ocp.yaml recipes/overlays/ocp-inference-nim.yaml recipes/components/cert-manager-ocp-olm recipes/components/cert-manager-ocp recipes/components/prometheus-adapter-ocp recipes/components/nvidia-dra-driver-gpu-ocp recipes/components/k8s-nim-operator-ocp recipes/checks/cert-manager-ocp-olm recipes/checks/cert-manager-ocp recipes/checks/prometheus-adapter-ocp
make lint-yaml
make bom-docs
```

All pass clean. **Not yet run:** `make qualify` / `make test` / `make component-test` against a real cluster — the OLM components (`cert-manager-ocp-olm`, `cert-manager-ocp`) specifically need a real OpenShift cluster or OLM manually bootstrapped into Kind, since `make component-test`'s Kind harness has no OperatorHub catalog by default. Marking this PR as draft until that's done.

## Risk Assessment

- [x] **Medium** — New components only, gated behind `service: ocp` criteria in the recipe registry; doesn't touch any existing non-OCP recipe or component. Risk is concentrated in the four open items above, not in blast radius.

**Rollout notes:** Draft — moving to ready-for-review once the four open items are verified against a real or OLM-bootstrapped cluster.

## Checklist

- [ ] Tests pass locally (`make test` with `-race`) — not run, no cluster available yet
- [ ] Linter passes (`make lint`) — `make lint-yaml` passes; full `make lint` currently fails on a pre-existing, unrelated `goconst` issue in `validators/conformance/slinky_slurm_accounting_check.go` (not touched by this PR)
- [x] I added/updated tests for new functionality (health-check.yaml / readiness.yaml chainsaw tests for all 5 new components)
- [x] I updated docs if user-facing behavior changed (`docs/user/container-images.md` via `make bom-docs`)
- [ ] Commits are cryptographically signed (`git commit -S`)


recipes/checks/cert-manager-ocp-olm/health-check.yaml
```diff
@@ -0,0 +1,26 @@
...
+# OLM operator readiness is verified at deploy time by the bundler's
+# --readiness-hooks gate (components/cert-manager-ocp-olm/readiness.yaml).
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: cert-manager-ocp-olm-health-check
+spec:
+  timeouts:
+    assert: 5m
+  steps: []
```
yuanchen8911, line 28: This check has `steps: []` but is missing the `metadata.annotations` block with `aicr/no-op-check: "true"`. TestRegistryNoOpChecksDeclareIntent (pkg/chainsaw/inprocess_test.go) fails any check that declares no assert/error operations without that annotation, and the in-process executor (pkg/chainsaw/inprocess.go) rejects it at runtime — so both `make test` and the operator-health deployment check fail for every OCP recipe. The three sibling OCP OLM checks (gpu-operator-ocp-olm, network-operator-ocp-olm, nfd-ocp-olm) all carry the annotation at line 26; copy that pattern.

recipes/overlays/ocp.yaml
```diff
@@ -84,6 +90,49 @@ spec:
...
+      manifestFiles:
+        - components/prometheus-adapter-ocp/manifests/ca-bundle-configmap.yaml
+        - components/prometheus-adapter-ocp/manifests/auth-config-configmap.yaml
+        - components/prometheus-adapter-ocp/manifests/rbac.yaml
+
+    # nvidia-dra-driver-gpu (issue #1818) — "inherit the generic helm
+    # repo" pattern, reuses the exact upstream chart with an added SCC
+    # RoleBinding. dependencyRefs mirrors overlays/base.yaml's
+    # `dependencyRefs: [gpu-operator]` but pointed at the OCP variant.
+    - name: nvidia-dra-driver-gpu-ocp
```
yuanchen8911, line 133: Removing the `enabled: false` overrides for cert-manager, prometheus-adapter, and nvidia-dra-driver-gpu (further down in this file) re-enables the inherited base refs rather than replacing them. This overlay has no `base:` field, so it inherits the root base.yaml (pkg/recipe/metadata.go:685), which ships cert-manager and nvidia-dra-driver-gpu enabled; prometheus-adapter arrives via the monitoring-hpa overlay (`intent: any`). componentRefs merge by name (mergeComponentRef, pkg/recipe/metadata.go:1273), and the `-ocp` refs have different names, so resolution now enables both members of each pair.

For DRA this is a hard conflict: components/nvidia-dra-driver-gpu-ocp/values.yaml pins nameOverride/fullnameOverride to the base component's names, so the two releases render the same object identities in the same namespace and fight over ownership. For prometheus-adapter, both releases register the same cluster-scoped v1beta1.custom.metrics.k8s.io APIService, and the base adapter points at a kube-prometheus-stack that stays disabled on OCP. For cert-manager, the upstream jetstack chart and the Red Hat operator both manage the same operand Deployments in the hardcoded cert-manager namespace.

Please restore the three disables; the `-ocp` variants coexist with them by name. (Comments elsewhere in this PR — ocp-inference-nim.yaml and the DRA values file — still say these are force-disabled.)

recipes/overlays/ocp.yaml
```diff
@@ -84,6 +90,49 @@ spec:
...
+        - cert-manager-ocp-olm
+
+    # Prometheus Adapter (issue #1818) — "adapter only" pattern, reuses
+    # the same chart as the base component but points it at OCP's
+    # built-in Thanos Querier instead of kube-prometheus-stack (which
+    # stays disabled below).
+    - name: prometheus-adapter-ocp
+      type: Helm
+      valuesFile: components/prometheus-adapter-ocp/values.yaml
+      manifestFiles:
```
yuanchen8911: These two ConfigMaps need to exist before the chart's Deployment starts: values.yaml mounts them via extraVolumes. But `manifestFiles` on an upstream-chart component are emitted as a separate `<name>-post` release installed after the primary release (pkg/bundler/deployer/localformat/doc.go), and the deploy script installs releases serially with `--wait` — so the primary release fails at the wait timeout, exhausts its retries, and the `-post` release that would create the ConfigMaps is never reached. Move ca-bundle-configmap.yaml and auth-config-configmap.yaml to `preManifestFiles`.

recipes/components/nvidia-dra-driver-gpu-ocp/manifests/scc-rolebinding.yaml
```diff
@@ -0,0 +1,55 @@
...
+    helm.sh/hook-weight: "-5"
+    helm.sh/hook-delete-policy: before-hook-creation
+  labels:
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: system:openshift:scc:privileged
+subjects:
+  - kind: ServiceAccount
```
yuanchen8911, line 46: Two problems here, and fixing only one still leaves the install broken.

First, ordering: this RoleBinding is listed under `manifestFiles`, which deploys in the `-post` release after the upstream chart — but the kubelet-plugin DaemonSet pods can't pass OCP SCC admission without it, so the serial `--wait` install of the main chart fails after its retries and the `-post` release never runs. The `helm.sh/hook: pre-install` annotation only orders resources within the `-post` release, not against the main chart. It belongs in `preManifestFiles`.

Second, the subjects don't exist. Chart 0.4.1 rendered with this PR's fullnameOverride creates ServiceAccounts `nvidia-dra-driver-gpu-service-account-controller` and `nvidia-dra-driver-gpu-service-account-kubeletplugin` (verified via `helm template`), not the `nvidia-dra-driver-gpu-controller` / `nvidia-dra-driver-gpu-kubelet-plugin` named here — the file's own header flags the names as inferred. The privileged DaemonSet runs as `...-service-account-kubeletplugin`; binding that account is sufficient (the controller does not render privileged or hostPath access).

recipes/components/prometheus-adapter-ocp/values.yaml
```diff
@@ -0,0 +1,165 @@
...
+
+extraVolumeMounts:
+  - name: prometheus-auth-config
+    mountPath: /etc/prometheus-adapter-ocp-auth
+    readOnly: true
+  - name: prometheus-ca-bundle
+    mountPath: /etc/prometheus-adapter-ocp-ca
+    readOnly: true
+
+serviceAccount:
```
yuanchen8911: Chart 5.3.0 defaults `runAsUser: 10001` and `fsGroup: 10001`, and this values file doesn't override them. Under OCP's default restricted-v2 SCC, UID uses MustRunAsRange and FSGroup uses MustRunAs against the namespace-allocated range, so a fixed 10001 outside that range is rejected at admission and the adapter pod never schedules. The rbac.yaml manifest here only binds cluster-monitoring-view — it grants no SCC. Either null out the fixed IDs so OCP assigns them from the namespace range, or add an SCC RoleBinding for the prometheus-adapter-ocp ServiceAccount like the other two Helm-reuse components do.

recipes/registry.yaml
```diff
@@ -934,3 +943,96 @@ components:
...
+    healthCheck:
+      # Intentionally reuses the base component's health check: OCP
+      # values.yaml pins nameOverride/fullnameOverride to the same
+      # strings as base, so rendered resource names are identical — see
+      # components/nvidia-dra-driver-gpu-ocp/values.yaml for why.
+      assertFile: checks/nvidia-dra-driver-gpu/health-check.yaml
+    helm:
+      defaultRepository: oci://registry.k8s.io/dra-driver-nvidia/charts
+      defaultChart: dra-driver-nvidia-gpu
+      defaultVersion: "0.4.1"
```
yuanchen8911, line 1013: Chart dra-driver-nvidia-gpu 0.4.1 declares `kubeVersion: '>=1.32.0-0'`, but the OCP overlays accept `K8s.server.version >= 1.29` — and OCP 4.16/4.17/4.18 ship Kubernetes 1.29/1.30/1.31. A recipe that passes constraint validation still fails at helm install, before rendering. The fix has to cover the leaf overlays too: ocp-training.yaml:29 and ocp-inference.yaml:29 restate `>= 1.29` independently of ocp.yaml. Either raise the floor to 1.32 everywhere (documenting the OCP-minor implication), or gate the DRA component behind a version constraint.

recipes/registry.yaml
```diff
@@ -934,3 +943,96 @@ components:
...
+      defaultVersion: 5.3.0
+      defaultNamespace: monitoring
+    nodeScheduling:
+      system:
+        nodeSelectorPaths:
+          - nodeSelector
+        tolerationPaths:
+          - tolerations
+
+  - name: nvidia-dra-driver-gpu-ocp
```
yuanchen8911, line 1000: DRA-aware code recognizes the driver by the literal component name `nvidia-dra-driver-gpu`: pkg/allocpolicy/allocpolicy.go:63, pkg/bundler/bundler.go:2615, pkg/bundler/validations/checks.go:319. Once the inherited base ref is disabled on OCP (see the ocp.yaml comment), OCP recipes contain only this new name, so GPU allocation-policy resolution ignores resources.gpus.enabled, dra-support/expected-resource validation skips the driver, and driver-root coherence checks miss it. A DRA opt-in on OCP would then pass validation as device-plugin mode — including the dual-advertisement case validation exists to reject.

Reusing the canonical DRA name alone is not a complete fix, because two separate stale-NVML mechanisms are involved: annotation injection requires both canonical names, gpu-operator and nvidia-dra-driver-gpu (pkg/bundler/bundler.go:2610-2679), and the Helm post-install migration wait/restart matches only nvidia-dra-driver-gpu (pkg/bundler/deployer/helm/templates/deploy.sh.tmpl:414-477) — OCP uses gpu-operator-ocp, so both protections go dark either way. The fix needs to cover both identities or centralize component aliases.

recipes/components/cert-manager-ocp-olm/readiness.yaml
```diff
@@ -0,0 +1,62 @@
...
+kind: Test
+metadata:
+  name: cert-manager-ocp-olm-readiness
+spec:
+  timeouts:
+    assert: 30s
+  steps:
+    - name: get-installed-csv
+      try:
+        - script:
```
yuanchen8911: This gate can't execute as written, though for a different reason than the VERIFY BEFORE MERGE comment anticipates: readiness gates don't run upstream chainsaw at all. The gate Job runs the in-process engine (cmd/gate/main.go — assert/error only; the chainsaw binary was removed in #2038), and pkg/chainsaw's read-only allowlist rejects `script` as a side-effecting operation, so this Test is refused before any output-binding semantics come into play — the gate Job fails every poll until max-wait and stalls the tier. The fallback already sketched in the comment at line 30 is the right fix: assert `status.installedCSV` non-empty on the Subscription directly, then assert the CSV phase. Asserts retry within timeouts.assert, which also covers the window where OLM hasn't populated installedCSV yet.

recipes/overlays/ocp-inference-nim.yaml
```diff
@@ -0,0 +1,77 @@
...
+  # Mirrors h100-eks-ubuntu-inference-nim.yaml, but layered on ocp-inference
+  # and pointed at the OCP-native dependencies (cert-manager-ocp,
+  # gpu-operator-ocp) instead of the base Helm componentRefs those don't
+  # exist as on OCP (both are force-disabled in overlays/ocp.yaml).
+  base: ocp-inference
+
+  criteria:
+    service: ocp
+    intent: inference
+    platform: nim
```
yuanchen8911, line 30: The new criteria combination adds `service=ocp + platform=nim` and `service=ocp + intent=inference + platform=nim` projections to the coverage matrix, but pkg/recipe/testdata/coverage_golden.yaml wasn't regenerated — TestCoverageGoldenMatrix fails under `make test` ("projection missing from golden — regenerate deliberately"). Regenerate with AICR_UPDATE_GOLDEN=1 and review the new projections.

docs/user/container-images.md
```diff
@@ -76,49 +81,59 @@ from the component's registry default above.
...
 |-----------|-----------------|-------------|--------|
-| kube-prometheus-stack | 83.7.0 | aks | 8 |
+| kube-prometheus-stack | 83.7.0 | aks | 0 |
 
 ## Images by component
 
 ### agentgateway
 
-- `cr.agentgateway.dev/controller:v1.3.1`
+> Warning: [NOT_FOUND] helm binary not found on PATH; install helm to discover images from chart templates
```
yuanchen8911: This regeneration ran without helm on PATH: 33 "NOT_FOUND: helm binary not found" warnings were committed, and the unique-image count dropped from 97 to 14 — the Helm-backed components' image lists were erased, not updated. Air-gap and security consumers rely on this file as the canonical inventory, and TestCommittedBOMVersionsMatchRegistry doesn't catch this (it gates only the version column and component set). Please rerun `make bom-docs` with helm installed and commit the full render — zero NOT_FOUND warnings, image count back at ~97+, growing with the new OCP components.

yuanchen8911 (commented):
Review of 04d67ac. Eleven blocking findings (several sharpen open items already acknowledged in the description), plus one doc gap and minor items noted at the end. Three themes:

**1. Recipe resolution now installs conflicting duplicates.** Overlays without a `base:` implicitly inherit base.yaml (pkg/recipe/metadata.go:685), and componentRefs merge by name — so the new `-ocp` refs do not replace the inherited cert-manager, prometheus-adapter, and nvidia-dra-driver-gpu refs; removing their `enabled: false` overrides re-enables them alongside the `-ocp` variants. The DRA pair is the worst case: the `-ocp` values pin nameOverride/fullnameOverride to the base names, so the two releases render the same Kubernetes object identities and Helm ownership conflict is certain.

**2. Several new components cannot come up on a supported OCP cluster.** The prometheus-adapter Deployment and the DRA kubelet-plugin DaemonSet each depend on resources emitted in their `-post` release (the serial `--wait` install fails after its retries before that release is ever reached); the DRA SCC RoleBinding additionally binds ServiceAccounts the chart does not create; the adapter pod is inadmissible under restricted-v2 (chart pins runAsUser/fsGroup 10001, no SCC granted); the DRA chart 0.4.1 declares `kubeVersion: '>=1.32.0-0'` while the ocp overlays accept `>= 1.29`; and the cert-manager readiness gate uses a `script` op that the gate CLI's in-process engine (assert/error only) rejects outright.

**3. Repo gates fail.** The empty cert-manager-ocp-olm health check lacks the `aicr/no-op-check` annotation (TestRegistryNoOpChecksDeclareIntent fails, and the runtime executor rejects it); the new ocp/nim criteria projections are missing from pkg/recipe/testdata/coverage_golden.yaml (TestCoverageGoldenMatrix fails); and the committed container-images.md was regenerated without helm on PATH — 33 NOT_FOUND warnings, unique images down from 97 to 14.

There is also a design question worth settling before the pattern spreads: DRA-aware code recognizes components by literal name (`nvidia-dra-driver-gpu`, and separately `gpu-operator`), so on OCP both the allocation-policy/validation paths and the stale-NVML protections go dark. Details inline.

**Not inline (file not in the diff):** none of the five new registry components is documented in docs/user/component-catalog.md, and its OCP summary paragraph (line 58) still says OCP replaces only gpu-operator/nfd/network-operator. The repo's documentation rule puts catalog updates in the same PR as the registry change.

**Minor:** recipes/overlays/ocp.yaml now ends without a trailing newline, and five committed ⚠ VERIFY BEFORE MERGE markers (readiness.yaml, both scc-rolebinding.yaml files, prometheus-adapter-ocp/values.yaml, ocp-inference-nim.yaml) should be resolved or converted to tracked follow-ups before merge.

yuanchen8911 (changes requested):
Requesting changes per the detailed review above (https://github.com/NVIDIA/aicr/pull/2057#pullrequestreview-4878610398): the duplicate component resolution on OCP recipes, the failing test gates (no-op annotation, coverage golden), and the BOM regression need to be addressed before merge.

mohityadav8:
cc @yuanchen8911 can u check it once i have done aprox all suggested things

pkg/bundler/validations/checks.go
```diff
@@ -874,15 +897,15 @@ func draLockstepViolations(ctx context.Context, recipeResult *recipe.RecipeResul
 			"%s: the operator-managed driver installs to hostPaths.driverInstallDir=%s "+
 				"but the DRA kubelet plugin reads nvidiaDriverRoot=%s; CDI spec generation "+
 				"will fail and DRA-allocated pods will stall in ContainerCreating. Fix with "+
-				"--set dradriver:nvidiaDriverRoot=%s (or, on a cluster with a "+
+				"--set %s:nvidiaDriverRoot=%s (or, on a cluster with a "+
 				"platform-preinstalled driver, the full preinstalled-driver profile).",
-			componentName, installDir, root, installDir))
+			componentName, installDir, draKeys[0], root, installDir))
```
yuanchen8911: The format arguments are out of order here. The verbs expect `(componentName, installDir, root, draKeys[0], installDir)` but the call passes `(componentName, installDir, draKeys[0], root, installDir)` — so the message renders "reads nvidiaDriverRoot=nvidia-dra-driver-gpu" and, with root `/`, "Fix with --set /:nvidiaDriverRoot=...". This is what's failing the two existing message-asserting unit tests. The second case below got the same rewrite right; only this one is swapped.

yuanchen8911 (changes requested):
I verified every open thread against head 45539744a. Four are resolved. The items below are what remains.

Two threads stay open (details in the threads): the 1.32 version floor needs an explicit compatibility decision and documentation in `docs/integrator/openshift.md`, and the DRA/GPU-operator `-ocp` alias coverage is still incomplete across several code paths.

Two new issues introduced by the latest commits:

1. Swapped format arguments in `pkg/bundler/validations/checks.go` (inline comment below). This breaks two existing unit tests in `go test -race ./pkg/bundler/...` and renders nonsense remedy text like `--set /:nvidiaDriverRoot=...`.

2. `tests/chainsaw/cli/bundle-ocp/chainsaw-test.yaml` (`assert-argocd-helm-omits-empty-static`, L153-180) still asserts the OCP bundle emits no `static/` directory, on the premise that every OCP component is a local chart. This PR intentionally adds two upstream Helm components, so the assertion will fail once CI runs on this head. Please update it to assert that `nvidia-dra-driver-gpu-ocp.yaml` and `prometheus-adapter-ocp.yaml` exist under `static/`, that their base equivalents do not, and update the doc statement that all OpenShift components are local charts.

Process items before the next push:

- The branch is behind main and carries 8 commits including 3 merge commits — please rebase onto origin/main and squash to a single signed, DCO'd commit (no merge commits).
- Testing evidence: please run the final automated checks (`make qualify`) at the squashed head and record the results in the PR body's Testing section, and either complete #1818's practical OCP deployment-validation criterion or explicitly revise that issue's acceptance scope. Updating the body text alone isn't sufficient.

recipes/registry.yaml
yuanchen8911: The floor is raised to `>= 1.32` in all three overlays, which resolves the mechanical mismatch. Keeping this open for the remaining half: OCP 4.16/4.17/4.18 ship Kubernetes 1.29/1.30/1.31 and 4.19 ships 1.32, so the raised floor drops 4.16-4.18 — while #1818 states "Compatibility / Breaking Changes: None". This needs an explicit decision recorded in the PR body (and #1818 if the answer is "OCP 4.19+ only"), plus a user-facing note in `docs/integrator/openshift.md`. The alternative from the original comment — gating only the DRA component behind a version constraint and keeping the overlay floor at 1.29 — is still available if 4.16-4.18 support matters.

recipes/registry.yaml
yuanchen8911: Partial progress at 45539744a: driver-root coherence now resolves the `-ocp` name via `resolveDRAComponentRef` — that site works. The rest of the alias surface is still uncovered, and most of it traces to one root cause — allocpolicy's canonical-name constants and `Descriptor()` have no `-ocp` entries:

1. Validator allocation-policy resolution: the lookup misses the `-ocp` ref entirely, so a DRA opt-in on OCP resolves as device-plugin mode and skips every DRA gate, including the dual-advertisement rejection that `docs/user/component-catalog.md` promises.

2. Bundle-time allocation-policy override enforcement (`pkg/bundler/allocation_policy.go`) iterates `allocpolicy.Descriptor()`, so `-ocp` components aren't covered there either.

3. Profile closure/coherence in `pkg/recipe/profile.go` iterates the same `Descriptor()` — same gap.

4. `deploy.sh.tmpl` L414-418 matches only the literal name `nvidia-dra-driver-gpu`, so the stale-NVML migration wait/restart never runs on OCP bundles.

5. Annotation injection recognizes the alias syntactically, but is structurally inert on OCP: `gpu-operator-ocp` carries no chart version, so it always hits the empty-version warn-and-skip.

6. `driverAbsentRemedy`'s shared override-set constant hardcodes `--set gpuoperator:...` / `--set dradriver:...` (checks.go and its `pkg/client/v1` sibling), so an OCP driver-absent recipe gets remedy keys that don't exist in it.

The default stays safe from dual whole-GPU advertisement (the `-ocp` values ship `gpus.enabled: false`), but `resources.computeDomains.enabled` is true, so the rollout/migration protections in items 4-5 still matter, and DRA opt-in is in #1818's supported scope. Fixing the Descriptor/constants root cause (items 1-3) belongs in this PR — the alias machinery already exists. Items 4-6 could be a documented follow-up issue if you'd rather scope them out; if so, please note the limitation in the component docs.

mohityadav8:
@yuanchen8911 pushed the format-args fix and the chainsaw-test/doc updates for the two new issues. sstill working on the three remaining items: the 1.32 compatibility decision + docs/integrator/openshift.md, the allocpolicy -ocp alias coverage (items 1-3 from your list), and the rebase/squash + make qualify run. Will follow up on each.

mohityadav8:
Ran make qualify at the squashed head (commit 1a2d50fd). All tests in packages this PR modifies pass cleanly:

pkg/allocpolicy: PASS, 100% coverage
pkg/bundler, pkg/bundler/validations: PASS
pkg/recipe: PASS
pkg/chainsaw: PASS

Two unrelated pre-existing failures in this local environment, neither touched by this PR:

pkg/oci: local helm binary is v3.21.3, one test expects an exact v4.2.3 pin; other pkg/oci failures are simulated-network-fault test fixtures unrelated to this diff.
tests/releasepolicy: jq not installed locally, plus file-mode differences from this checkout's permission bits - release-tooling tests, not exercised by this PR.

make component-test for the OLM components (cert-manager-ocp-olm, cert-manager-ocp) still requires a real OpenShift cluster or an OLM-bootstrapped Kind cluster -not available in this environment, consistent with the original PR description. Requesting guidance on whether that gap blocks merge or can be tracked as a follow-up against #1818's acceptance scope.

recipes/registry.yaml
yuanchen8911: Verified at 1a2d50fdd — good progress: `ComponentDRADriverOCP` and the two `Descriptor()` entries cover items 2-3 (bundle-time override enforcement and profile closure now see the -ocp components).

Two things remain on this thread:

1. Item 1 is one line away: the validator's own lookup at `pkg/validator/v1/allocation_policy.go:156` still queries only the canonical name (`enabledComponentRef(r, draDriverGPUComponentName)`), so an OCP DRA opt-in still bypasses every DRA gate — the new constant exists but this call site wasn't updated. Please extend the lookup to try `allocpolicy.ComponentDRADriverOCP` as well (same fallback pattern as your `resolveDRAComponentRef`).

2. Items 4-6 (the deploy.sh name match, the annotation injection that always hits the empty-version skip on OCP, and the hardcoded `gpuoperator:`/`dradriver:` remedy keys) are fine to defer — but please file the follow-up issue and add a short limitation note in the component docs so the deferral is recorded, then link the issue here.

With those two, this thread can close.

recipes/checks/cert-manager-ocp-olm/health-check.yaml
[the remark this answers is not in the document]
yuanchen8911: The no-op is deliberate and now properly declared: the check carries `aicr/no-op-check: "true"` and the header documents that OLM operator readiness is verified at deploy time by the bundler's readiness-hooks gate (`components/cert-manager-ocp-olm/readiness.yaml`). This matches the established pattern of the three sibling OCP OLM checks (gpu-operator-ocp-olm, network-operator-ocp-olm, nfd-ocp-olm), which are no-ops for the same reason — the OLM Subscription/CSV is a deploy-time concern, and the operand health is asserted by the cert-manager-ocp check, which now includes per-Pod phase assertions. Keeping the sibling pattern consistent; resolving.

recipes/components/prometheus-adapter-ocp/manifests/auth-config-configmap.yaml
```diff
@@ -0,0 +1,53 @@
...
+    kind: Config
+    clusters:
+      - name: thanos-querier
+        cluster:
+          certificate-authority: /etc/prometheus-adapter-ocp-ca/service-ca.crt
+          server: https://thanos-querier.openshift-monitoring.svc.cluster.local:9091
+    users:
+      - name: prometheus-adapter-ocp
+        user:
+          tokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
```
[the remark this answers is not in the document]
yuanchen8911: The redirect behavior lives in the upstream prometheus-adapter binary's HTTP client, not in anything this PR can configure — the auth config selects the token file and target URL only, so `CheckRedirect` isn't reachable from a values/manifest change. On the reachability side, the target is the fixed in-cluster Thanos Querier service endpoint in openshift-monitoring, which serves the query API directly over TLS and is not a redirecting endpoint; producing a redirect there would require control of the cluster-monitoring stack, which is outside this threat model (and per the finding itself, reachability is internal). Happy to note it as an upstream hardening suggestion, but it's not actionable in this PR. Resolving.

yuanchen8911 (changes requested):
Re-verified everything at 1a2d50fd — this round addressed nearly all of it: the format-args fix, the reworked chainsaw static/ assertions, the openshift.md compatibility decision (OCP 4.19+), the allocpolicy Descriptor entries, the rebase to a single signed commit, and the recorded make qualify run. All corresponding threads are resolved, along with the two review-bot threads that are addressed or not actionable in this PR (rationale in each).

Two items remain, both in the open thread on recipes/registry.yaml:

1. The validator's DRA lookup (pkg/validator/v1/allocation_policy.go:156) still queries only the canonical component name, so an OCP DRA opt-in bypasses the DRA gates. One-line fix now that ComponentDRADriverOCP exists.

2. The deferred alias items (deploy.sh name match, OCP annotation-injection skip, hardcoded remedy override keys) need the deferral recorded: a filed follow-up issue plus a short limitation note in the component docs, linked in the thread.

With those two, this is ready for my approval.

mohityadav8:
@yuanchen8911 can u check this now can u run workflow for this so if any test is failing i can fix that

docs/user/component-catalog.md
```diff
@@ -49,13 +49,18 @@ The source of truth is [`recipes/registry.yaml`](https://github.com/NVIDIA/aicr/
...
+| **cert-manager-ocp** | cert-manager CertManager CR for OpenShift. The operand Deployments (controller, cainjector, webhook) land in a hardcoded `cert-manager` namespace regardless of the operator's own namespace. Deployed after `cert-manager-ocp-olm`. OCP-specific. | [cert-manager](https://github.com/cert-manager/cert-manager) |
+| **prometheus-adapter-ocp** | Prometheus Adapter for OpenShift. Reuses the same upstream chart as `prometheus-adapter`, pointed at OCP's built-in Thanos Querier instead of kube-prometheus-stack (which stays disabled on OCP). No certified OCP operator exists for this component. OCP-specific. | [prometheus-adapter](https://github.com/kubernetes-sigs/prometheus-adapter) |
+| **nvidia-dra-driver-gpu-ocp** | NVIDIA DRA GPU driver for OpenShift. Reuses the same upstream chart as `nvidia-dra-driver-gpu`, with an added SCC RoleBinding granting the kubelet-plugin DaemonSet the host device access OCP's default restricted-v2 SCC forbids. No certified OCP operator exists for this component. OCP-specific. | [NVIDIA DRA Driver](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu) |
```
yuanchen8911: Committable version of the limitation note requested in the registry.yaml thread, recording the deferral to #2136:

```suggestion
| **nvidia-dra-driver-gpu-ocp** | NVIDIA DRA GPU driver for OpenShift. Reuses the same upstream chart as `nvidia-dra-driver-gpu`, with an added SCC RoleBinding granting the kubelet-plugin DaemonSet the host device access OCP's default restricted-v2 SCC forbids. No certified OCP operator exists for this component. OCP-specific. Known limitation: some GPU-driver rollout protections and remedy hints do not yet cover the OCP aliases (`gpu-operator-ocp`, `nvidia-dra-driver-gpu-ocp`) — the deployer's stale-NVML migration wait/restart, driver-version annotation injection, and the driver-absent remedy's `gpuoperator:`/`dradriver:` override keys; tracked in [#2136](https://github.com/NVIDIA/aicr/issues/2136). | [NVIDIA DRA Driver](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu) |
```

recipes/registry.yaml
yuanchen8911: Filed the follow-up for items 4-6: #2136 (on the AICR board, P2). It records the deploy.sh name match, the annotation-injection empty-version skip on OCP, and the hardcoded `gpuoperator:`/`dradriver:` remedy keys, with the file references from this thread.

For the limitation note, I posted a committable suggestion on the `nvidia-dra-driver-gpu-ocp` row in `docs/user/component-catalog.md` (https://github.com/NVIDIA/aicr/pull/2057#discussion_r3741402524) — applying it satisfies the doc-note half of item 2.

On item 1: the alias fallback in 0c681c6 addresses it. One remaining ask there — `allocation_policy_test.go` has `gpu-operator-ocp` cases but nothing exercising the new `nvidia-dra-driver-gpu-ocp` fallback; please add a case for the OCP DRA opt-in shape (and ideally the both-enabled preference) so the fix is covered.

With the suggestion applied and the test added, this thread can close.

yuanchen8911 (approved):
Verified at c824f5f86: the alias fallback fix (0c681c6), the applied limitation note pointing at #2136, and test coverage for the -ocp fallback (opt-in resolution, dual-advertisement rejection through the alias, canonical preference when both are enabled). All threads resolved. Approving.

mohityadav8:
@yuanchen8911 thanku very much for helping me in this pr this was hard for me 😁🤍

Issue #2019, opened 2026-08-04
sdk: reconcile facade alias state and semver contract language


mchmarny:
## Summary

Two records about the facade's contract are out of date, and they disagree with each other and with reality. Both are cheap to fix and both currently mislead integrators.

## A — Stale alias tracking

Issue #1078 ("feat(aicr): wrap facade alias types as facade-owned structs") is closed as **COMPLETED** (2026-05-30), but the aliases it describes are all still transparent passthroughs:

- `BundleConfig = config.Config`
- `BundleAttester = attestation.Attester`
- `BundleArtifact = *result.Output`
- `CriteriaRegistry = recipe.CriteriaRegistry`
- `OIDCResolveOptions = bundleattest.ResolveOptions`

`pkg/client/v1/bundle.go` still carries comments reading "Transparent alias of `pkg/bundler/config.Config` (the alias is tracked by #1078)" and "(#1078 wraps it)". Either the issue was closed in error or the comments are wrong; today a reader cannot tell which.

Note that `CriteriaRegistry` is a *deliberate* alias with a documented rationale in `types.go` (the registry is behavior-rich and carries per-DataProvider mutable state, so wrapping buys no semver protection). That reasoning should survive whatever we decide; it is not the same case as `BundleConfig`.

## B — The semver promise overstates what v0 can deliver

The package godoc says:

> This package's exported API follows semver.

The module is `github.com/NVIDIA/aicr` at v0.x. Under Go module semantics, v0 minors may break freely, and `pkg/client/v1` shares the root module's version — so the promise is not one the project can currently keep. `docs/integrator/go-library.md` hedges this correctly ("Today AICR is pre-1.0. **Pin a patch version**"), but a consumer reading `go doc` never sees that hedge.

## C — API-diff does not inspect alias targets

The compatibility gate added by #2060 compares declarations exported by
`pkg/client/v1` against the latest stable release. For named types outside the
compared package, `apidiff` matches package path and type name but does not
recursively compare the target definition. The transparent aliases listed above
therefore expose a blind spot: removing a `CriteriaRegistry` method, changing
the `BundleAttester` interface, or removing an exported field from
`BundleArtifact` can break SDK consumers while `make api-diff` remains green.

Until the aliases are wrapped or covered by scoped checks, changes to their
target definitions require manual compatibility review. PR #2060 documents this
interim limitation; this issue tracks the durable fix.

## Proposed change

[...]

Issue #2016, opened 2026-08-04
[Epic]: Go SDK — make pkg/client/v1 the complete integration surface


mchmarny:
## Goal

Make `pkg/client/v1` the complete and defensibly stable integration surface for the core AICR workflow.

An integrator must be able to complete:

`snapshot → criteria → recipe → validate → bundle → verify`

without importing another AICR `pkg/*` package.

## v1 closure scope

| Issue | Status | Required outcome |
|---|---|---|
| #2120 | open | Concurrent snapshot collection satisfies the facade's published isolation contract. |
| #2245 | open | Bundle, validation, and snapshot configuration no longer require `Config.Unwrap()` or `pkg/config`. |
| #2437 | **landed** | The facade derives recipe criteria from a snapshot without exposing `pkg/fingerprint` or `pkg/recipe`. |
| #2028 | open | A merge-blocking architecture test prevents CLI/server business-logic bypasses around the facade. |
| #2029 | open | The integrator guide and compile-only example demonstrate the complete workflow using only `pkg/client/v1`. |

## Acceptance criteria

- [ ] The complete workflow compiles using only `pkg/client/v1`, standard-library packages, and explicitly stable third-party types.
- [ ] Snapshot collection is safe for concurrent SDK consumers.
- [ ] All configuration needed by the workflow is available through the facade.
- [ ] The exported-surface stability test and released-baseline API-diff gate remain green.
- [ ] The CLI/server architecture gate rejects new unstable-package bypasses.
- [ ] Documentation contains no unstable-package escape hatch for the core workflow.

## Post-v1 follow-up

These open native children remain valuable but are not required by the v1 exit gate:

- #2025 — mirror inventory.

#2225, #2227, and #2256 have since closed.

## Native sub-issue inventory

The native graph contains 20 children: 15 closed and 5 open. Four open children are v1 blockers (#2120, #2245, #2028, #2029); one is post-v1 (#2025).

#2025 does not gate #2028 or #2029. Mirror inventory is outside the `snapshot → criteria → recipe → validate → bundle → verify` workflow this epic scopes, so until it closes the CLI's mirror/trust use is a reasoned allowlist exception naming #2025 rather than an unfinished workflow silently accepted.

## Out of scope

- Making every CLI capability available through the SDK.
- Publishing trust-cache maintenance before its underlying operation supports reliable cancellation.
- General component upgrade-lifecycle safety.
- Tagging `v1.0.0`; this epic is a prerequisite, not the release action.

[...]

Issue #1974, opened 2026-07-31
tests/releasepolicy: 10s per-script deadline flakes under full-suite load on dev machines


yuanchen8911:
## Summary

`tests/releasepolicy`'s shell-script tests enforce a hardcoded 10-second wall-clock deadline per script invocation (`runScript`, `release_scripts_test.go:1425`). Under full-suite `make test` parallelism (`-race`, `-covermode=atomic`, ~96 packages concurrently), individual script runs starve past that margin on developer machines and fail with `script exceeded test deadline: context deadline exceeded` — a false alarm that intermittently breaks the documented pre-push `make qualify` gate.

## Evidence (2026-07-30, Apple Silicon dev machine)

- Failed 3 of 7 full local `make qualify` runs in one day — with a **different subset of tests each time** (e.g. `TestReleasePromotionPrereleaseWritesOnlyVersionAliases`, `TestReleasePreflightPaginatesAndPeelsAnnotatedTags` in one run; `TestReleaseTargetRerunPolicy`, `TestReleaseStablePromotionConvergesAndRerunsWithoutMutation` in another). Every failure is the deadline sentinel, never a functional assertion.
- The package **passes in isolation** on the same machine with the gate's exact flags and env (`AICR_CRITERIA_STRICT=1 go test -short -count=1 -race`), ~153s total across all 31 script invocations.
- Clean `origin/main` in-suite on the same machine: passed, but at 190s package time — visibly near the margin even without any in-flight change.
- The triggering PR branch (#1967) touches nothing the package executes; its only contribution is more concurrent test load elsewhere in the suite.
- Never observed on CI runners (checks green throughout).

## Root cause

A fixed wall-clock deadline measures the machine and its concurrent load, not the script's correctness. The deadline exists to catch a *hung* script (infinite loop, stuck prompt) — and hangs are separated from slow-but-working runs by orders of magnitude, so a tight constant buys no detection value while making the gate environment-sensitive.

## Proposed fix

Derive the per-script budget from the test binary's own deadline instead of a constant, with a hang-detection-sized cap:

```go
budget := 90 * time.Second
if d, ok := t.Deadline(); ok {
    if remaining := time.Until(d) - 30*time.Second; remaining < budget {
        budget = remaining
    }
}
ctx, cancel := context.WithTimeout(context.Background(), budget)
```

[...]

PR #2060, opened 2026-08-05
ci(sdk): gate client API compatibility

### LINKED ISSUE #2018: sdk: add a mechanical API-diff gate for pkg/client/v1
## Summary

No mechanical API-compatibility check runs against `pkg/client/v1`. A dropped method, a changed signature, or a removed struct field that nobody remembered to pin in `stability_test.go` ships silently. For a package whose entire purpose is a stability contract with out-of-tree integrators, hand-maintained assertions are not a sufficient gate.

## Current state

- No `apidiff`, `gorelease`, or equivalent anywhere in `.github/workflows/` or `Makefile`.
- `make qualify` — the documented CI-equivalent gate — does not check API compatibility.
- The only defense is `stability_test.go`, whose gaps are tracked in issue (1) of the epic.

## Proposed change

Add an API-diff gate scoped to `pkg/client/v1`, run from `make qualify` so it holds locally and in the merge gate (per the repo's "local development equals CI" principle).

Design notes for whoever picks this up:

- **Tool.** `golang.org/x/exp/cmd/apidiff` compares two package versions and classifies changes as compatible / incompatible. `gorelease` is more opinionated (it wants to tell you the next version) and is module-scoped, which is awkward here because we only want to gate one package inside the module. `apidiff` is the better fit; pin its version in `.settings.yaml` like every other tool.
- **Baseline.** Diff against the most recent release tag, not `origin/main` — the contract is with released consumers.
- **Verdict.** An incompatible change should **fail** the gate, with an explicit override mechanism (a label, or a file listing acknowledged breaks) so an intentional pre-1.0 break can land with a deliberate decision rather than silence. Whichever form it takes, the override must leave a written record — that record is the raw material for the release notes' breaking-change section.
- Report additive changes as informational output, not a failure.

## Acceptance criteria

- `make qualify` fails on an incompatible change to `pkg/client/v1` relative to the last release tag.
- The tool version is pinned in `.settings.yaml` and installed by `make tools-setup`.
- An intentional break can be acknowledged explicitly, and the acknowledgement is discoverable when writing release notes.
- A deliberately-broken test case (removing an exported method) demonstrably fails the gate.

Parent: #2016

tjrasche:
## Summary

Adds a release-baseline API compatibility gate for `pkg/client/v1` to local
qualification and CI, with fail-closed acknowledgements for intentional breaks.

## Motivation / Context

The public SDK facade is consumed by out-of-tree Go callers, but the existing
hand-maintained stability test cannot reliably detect every removed export or
incompatible signature change. This adds a mechanical comparison against the
latest stable release.

Fixes: #2018
Related: #1891, #2016, #2019

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: SDK compatibility CI and developer tooling

## Implementation Notes

- Pins and verifies `apidiff`, then compares `pkg/client/v1` with the latest
  reachable stable tag in a temporary detached worktree.
- Reports compatible additions and blocks incompatible changes unless exactly
  acknowledged for the active baseline. Malformed, duplicate, mismatched, and
  stale acknowledgements fail closed.
- Keeps the reusable `go-test` action backward-compatible by making the API-diff
  input optional; AICR qualification always supplies the pin and runs the gate.
- Documents the transparent-alias limitation and manual review requirement
  tracked by #2019.

## Testing

```bash
unset GITLAB_TOKEN
make qualify
```

Passed. This includes race-enabled tests, coverage, lint, e2e, vulnerability
scan, license checks, and the new API-diff decision and tool-version tests.

## Risk Assessment

**Rollout notes:** Contributors should rerun `make tools-setup` and use a clone
with full tag history. Existing cross-repository `go-test` consumers remain
compatible.

## Checklist

### DIFF
--- a/.github/actions/go-test/action.yml
+++ b/.github/actions/go-test/action.yml
@@ -30,6 +30,10 @@ inputs:
   helm_version:
     description: 'Helm version pinned in .settings.yaml (testing_tools.helm, via load-versions) — required by the argocd-helm live-render tests, which fail (not skip) in CI when helm is missing'
     required: true
+  apidiff_version:
+    description: 'Optional apidiff version pinned in .settings.yaml (linting.apidiff, via load-versions); leave empty to skip the SDK API compatibility check'
+    required: false
+    default: ''
 
 runs:
   using: 'composite'
@@ -71,9 +75,23 @@ runs:
       shell: bash
       run: make test
 
+    - name: Install API-diff tool
+      if: inputs.apidiff_version != ''
+      shell: bash
+      env:
+        APIDIFF_VERSION: ${{ inputs.apidiff_version }}
+      run: |
+        set -euo pipefail
+        GOFLAGS= go install "golang.org/x/exp/cmd/apidiff@${APIDIFF_VERSION}"
+
     - name: Coverage Report
       if: inputs.coverage_report == 'true'
       uses: ./.github/actions/go-coverage
       with:
         coverage_file: ./coverage.out
         threshold: ${{ inputs.coverage_threshold }}
+
+    - name: Check SDK API compatibility
+      if: inputs.apidiff_version != ''
+      shell: bash
+      run: make api-diff
--- a/docs/contributor/maintaining.md
+++ b/docs/contributor/maintaining.md
@@ -27,6 +27,48 @@ The short form:
 
 Bi-weekly cadence; hotfix between cycles when a fix is critical.
 
+### SDK API Compatibility
+
+`make api-diff` compares the exported `pkg/client/v1` surface to the latest
+stable release tag. It runs through `make qualify` and the qualification
+workflow; additive changes are reported, while removals and incompatible type
+changes fail the gate.
+
+Local runs require the repository-pinned `apidiff` and `yq`; install them with
+`make tools-setup`. They also require full tag history and a stable release tag
+reachable from `HEAD`. The gate checks out that baseline in a temporary
+detached worktree, so it leaves the current working tree unchanged but adds
+checkout and filesystem I/O cost to `make qualify`.
+
+The gate compares declarations exported by `pkg/client/v1`. For external named
+types reached through transparent aliases, `apidiff` matches the target's
+package path and type name but does not recursively compare its definition. A
+change to such a target can therefore alter the facade contract without failing
+the gate. Reviewers must manually assess changes to aliased target types as
+changes to the stable facade until
+[#2019](https://github.com/NVIDIA/aicr/issues/2019) resolves this gap.
+
+To acknowledge an intentional break, first run `make api-diff`. Add a
+baseline-scoped entry to `pkg/client/v1/api-diff-exceptions.yaml` containing the
+reported baseline plus non-empty `issue`, `summary`, and `rationale` fields.
+Exactly one acknowledgement entry is allowed per baseline. Copy every reported
+incompatible line that begins with `- ` into `incompatible_changes`, omitting
+the `- ` prefix. The `incompatible_changes` list must exactly and completely
+match the command output; omissions and extras both fail the gate.
+
+The acknowledgement authorizes a break only for the active baseline, so keep it
+in place through the release that ships that break — it is what keeps the gate
+green until the release tag lands. Once the tag advances the baseline, the entry
+is obsolete and should be pruned. The gate is deliberately asymmetric about
+this. When the diff against the new baseline is clean there is nothing for a
+stale entry to authorize, so the gate only warns that the entry is prunable and
+still exits successfully; this is what lets the release pipeline and open pull
+requests stay green while the cleanup lands. When the diff is *not* clean, a
+stale entry is a hard failure: an acknowledgement scoped to an older baseline
+must never be accepted for a break against the current one. Prune the entry in a
+follow-up change after the release; Git history retains the release-notes record
+of the breaking change.
+
 ### Common Release Breakages
 
 **`goreleaser` fails with auth conflict.** `goreleaser` panics if both
--- a/tools/api-diff
+++ b/tools/api-diff
@@ -1,0 +1,183 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# api-diff compares pkg/client/v1 with the most recent stable release. The
+# current working tree is never modified: the release source is materialized in
+# a temporary detached worktree and apidiff consumes its export data.
+
+set -euo pipefail
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+# shellcheck source=tools/common
+. "${DIR}/common"
+
+if command -v go >/dev/null 2>&1; then
+    if ! go_path="$(go env GOPATH)" || [[ -z "${go_path}" ]]; then
+        err "Could not determine GOPATH with 'go env GOPATH'."
+    fi
+    export PATH="${go_path}/bin:${PATH}"
+fi
+
+has_tools apidiff git yq
+
+# apidiff resolves the new side from its working directory. Anchor it to this
+# checkout so callers outside the module cannot accidentally compare releases
+# from the module cache instead of the current source tree.
+cd "${REPO_ROOT}"
+
+readonly PACKAGE='github.com/NVIDIA/aicr/pkg/client/v1'
+readonly VERSIONS_FILE="${REPO_ROOT}/.settings.yaml"
+readonly EXCEPTIONS_FILE="${API_DIFF_EXCEPTIONS_FILE:-${REPO_ROOT}/pkg/client/v1/api-diff-exceptions.yaml}"
+readonly EXIT_NO_BASELINE=10
+readonly EXIT_EXCEPTIONS_FILE_MISSING=11
+readonly EXIT_EXCEPTIONS_EVALUATION_FAILED=12
+readonly EXIT_EXCEPTIONS_MALFORMED=13
+readonly EXIT_STALE_ACKNOWLEDGEMENT=14
+readonly EXIT_DUPLICATE_ACKNOWLEDGEMENT=15
+readonly EXIT_INCOMPATIBLE_CHANGES=16
+readonly EXIT_APIDIFF_VERSION_MISMATCH=17
+
+# These statuses are part of the decision-flow test contract. They prevent a
+# failure in an earlier branch from satisfying an assertion for a later one.
+api_diff_err() {
+    local status=$1
+    local message=$2
+
+    printf "${COLOR_RED}[ERR]${COLOR_RESET} %s\n" "${message}" >&2
+    exit "${status}"
+}
+
+if ! required_apidiff_version="$(yq eval -r '.linting.apidiff' "${VERSIONS_FILE}")"; then
+    api_diff_err "${EXIT_APIDIFF_VERSION_MISMATCH}" "Could not read the pinned apidiff version from ${VERSIONS_FILE}."
+fi
+if [[ -z "${required_apidiff_version}" || "${required_apidiff_version}" == "null" ]]; then
+    api_diff_err "${EXIT_APIDIFF_VERSION_MISMATCH}" "No apidiff version is pinned in ${VERSIONS_FILE}."
+fi
+
+if ! installed_apidiff_version="$(go_binary_module_version "$(command -v apidiff)" golang.org/x/exp)"; then
+    api_diff_err "${EXIT_APIDIFF_VERSION_MISMATCH}" "Could not verify the module version of apidiff; run make tools-setup to install ${required_apidiff_version}."
+fi
+if [[ "${installed_apidiff_version}" != "${required_apidiff_version}" ]]; then
+    api_diff_err "${EXIT_APIDIFF_VERSION_MISMATCH}" "apidiff version mismatch: installed ${installed_apidiff_version}, required ${required_apidiff_version}; run make tools-setup."
+fi
+
+baseline="${API_DIFF_BASELINE:-}"
+if [[ -z "${baseline}" ]]; then
+    tags="$(git tag --merged HEAD -l 'v*' --sort=-v:refname)"
+    while IFS= read -r tag; do
+        if [[ "${tag}" != *-* ]]; then
+            baseline="${tag}"
+            break
+        fi
+    done <<< "${tags}"
+fi
+if [[ -z "${baseline}" ]]; then
+    api_diff_err "${EXIT_NO_BASELINE}" "No stable release tag reachable from HEAD; cannot check ${PACKAGE} compatibility."
+fi
+
+if [[ ! -f "${EXCEPTIONS_FILE}" ]]; then
+    api_diff_err "${EXIT_EXCEPTIONS_FILE_MISSING}" "API-diff exceptions file not found: ${EXCEPTIONS_FILE}"
+fi
+
+temp_dir="$(mktemp -d "${TMPDIR:-/tmp}/aicr-api-diff.XXXXXX")"
+release_worktree="${temp_dir}/release"
+release_export="${temp_dir}/release.export"
+
+cleanup() {
+    git worktree remove "${release_worktree}" >/dev/null 2>&1 || true
+    rm -rf "${temp_dir}"
+}
+trap cleanup EXIT
+
+msg "Comparing ${PACKAGE} against ${baseline}"
+git worktree prune
+git worktree add --detach "${release_worktree}" "${baseline}" >/dev/null
+(
+    cd "${release_worktree}"
+    apidiff -w "${release_export}" "${PACKAGE}"
+)
+
+# Additions are informational and intentionally do not fail the gate. Capture
+# the complete report once, then extract its incompatible section for
+# acknowledgement decisions. A second apidiff invocation would reload the
+# working-tree package and its dependency graph.
+report="$(apidiff "${release_export}" "${PACKAGE}")"
+incompatible="$(printf '%s\n' "${report}" | awk '
+    $0 == "Incompatible changes:" { in_incompatible = 1; next }
+    $0 == "Compatible changes:" { in_incompatible = 0 }
+    in_incompatible { print }
+')"
+if [[ -n "${report}" ]]; then
+    printf '%s\n' "${report}"
+fi
+
+if ! acknowledgements_valid="$(yq eval -r '
+    (.acknowledgements | tag == "!!seq") and
+    (.acknowledgements | all_c(
+        (tag == "!!map") and
+        (.baseline | tag == "!!str" and length > 0) and
+        (.issue | tag == "!!str" and length > 0) and
+        (.summary | tag == "!!str" and length > 0) and
+        (.rationale | tag == "!!str" and length > 0) and
+        (.incompatible_changes | tag == "!!seq" and length > 0) and
+        (.incompatible_changes | all_c(tag == "!!str" and length > 0))
+    ))
+' "${EXCEPTIONS_FILE}")"; then
+    api_diff_err "${EXIT_EXCEPTIONS_EVALUATION_FAILED}" "Could not validate API-diff acknowledgements in ${EXCEPTIONS_FILE}."
+fi
+if [[ "${acknowledgements_valid}" != "true" ]]; then
+    api_diff_err "${EXIT_EXCEPTIONS_MALFORMED}" "Malformed API-diff acknowledgements in ${EXCEPTIONS_FILE}: acknowledgements must be an array of entries with non-empty baseline, issue, summary, and rationale strings plus a non-empty incompatible_changes string array."
+fi
+
+stale_baselines="$(BASELINE="${baseline}" yq eval -r '[.acknowledgements[] | select(.baseline != strenv(BASELINE)) | .baseline] | unique | sort | join(", ")' "${EXCEPTIONS_FILE}")"
+acknowledgement_count="$(BASELINE="${baseline}" yq eval -r '[.acknowledgements[] | select(.baseline == strenv(BASELINE))] | length' "${EXCEPTIONS_FILE}")"
+if (( acknowledgement_count > 1 )); then
+    api_diff_err "${EXIT_DUPLICATE_ACKNOWLEDGEMENT}" "Duplicate API-diff acknowledgements for baseline ${baseline} in ${EXCEPTIONS_FILE}; exactly one entry is allowed."
+fi
+
+# A clean diff has nothing to adjudicate, so a stale acknowledgement cannot
+# authorize anything: it is merely obsolete bookkeeping. Failing here would
+# deadlock releases, because an entry that is required before the release tag
+# exists becomes stale the moment that tag advances the baseline. Warn instead
+# so the maintainer can prune it on a green build.
+if [[ -z "${incompatible}" ]]; then
+    msg "No incompatible ${PACKAGE} changes since ${baseline}."
+    if [[ -n "${stale_baselines}" ]]; then
+        log_warning "Prunable API-diff acknowledgement baseline(s) in ${EXCEPTIONS_FILE}: ${stale_baselines}; active baseline is ${baseline} and its diff is clean. Remove the acknowledgement(s) for non-active baselines."
+    fi
+    exit 0
+fi
+
+# There are incompatible changes to adjudicate: a stale acknowledgement must
+# never stand in for one scoped to the active baseline.
+if [[ -n "${stale_baselines}" ]]; then
+    api_diff_err "${EXIT_STALE_ACKNOWLEDGEMENT}" "Stale API-diff acknowledgement baseline(s) in ${EXCEPTIONS_FILE}: ${stale_baselines}; active baseline is ${baseline}. Remove acknowledgements for non-active baselines."
+fi
+
+actual_changes="$(printf '%s\n' "${incompatible}" | sed -n 's/^- //p' | sort)"
+
+if (( acknowledgement_count == 1 )); then
+    issue="$(BASELINE="${baseline}" yq eval -r '.acknowledgements[] | select(.baseline == strenv(BASELINE)) | .issue' "${EXCEPTIONS_FILE}")"
+    summary="$(BASELINE="${baseline}" yq eval -r '.acknowledgements[] | select(.baseline == strenv(BASELINE)) | .summary' "${EXCEPTIONS_FILE}")"
+    acknowledged_changes="$(BASELINE="${baseline}" yq eval -r '.acknowledgements[] | select(.baseline == strenv(BASELINE)) | .incompatible_changes[]' "${EXCEPTIONS_FILE}" | sort)"
+    if [[ "${actual_changes}" == "$(printf '%s\n' "${acknowledged_changes}" | sort)" ]]; then
+        log_warning "Allowing acknowledged incompatible API change(s) for ${baseline}: ${issue}: ${summary}"
+        exit 0
+    fi
+    log_error "The incompatible API changes do not exactly match the acknowledgement for ${baseline}."
+    log_error "Acknowledged changes: ${acknowledged_changes}"
+fi
+
+api_diff_err "${EXIT_INCOMPATIBLE_CHANGES}" "Incompatible ${PACKAGE} changes since ${baseline}. Add a baseline-scoped acknowledgement to ${EXCEPTIONS_FILE} only for an intentional breaking change."
--- a/tools/check-tools
+++ b/tools/check-tools
@@ -50,6 +50,7 @@ fi
 # version_cmd: Command to extract version
 # match_strategy: How to compare versions
 #   - exact: Full version must match
+#   - strict_exact: Full version must match; missing/unverifiable/mismatch fails the check
 #   - major: Only major version must match
 #   - major_minor: Major.minor must match
 #   - presence: Only check if installed (no version comparison)
@@ -83,6 +84,9 @@ get_tool_metadata() {
         # Linting
         golangci_lint)
             echo "golangci-lint:::golangci-lint version --short 2>/dev/null:::major"
+            ;;
+        apidiff)
+            echo "apidiff:::go_binary_module_version \"\$(command -v apidiff)\" golang.org/x/exp:::strict_exact"
             ;;
         yamllint)
             echo "yamllint:::yamllint --version 2>/dev/null | awk '{print \$2}':::major"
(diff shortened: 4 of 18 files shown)


.github/actions/go-test/action.yml
```diff
@@ -30,6 +30,9 @@ inputs:
   helm_version:
     description: 'Helm version pinned in .settings.yaml (testing_tools.helm, via load-versions) — required by the argocd-helm live-render tests, which fail (not skip) in CI when helm is missing'
     required: true
+  apidiff_version:
```
mchmarny, line 33: Non-blocking, but it breaks copy-paste callers: this is a new **required** input, and `.github/actions/README.md` — the documented contract for this composite action — was not updated. The input list at [L193](https://github.com/NVIDIA/aicr/blob/6f4e3e0024a2d2c739136ded0676f3a8dfeb30b2/.github/actions/README.md#L193) still enumerates only `helm_version`/`coverage_report`, and all four usage examples (L336, L357, and the cross-repo form `NVIDIA/aicr/.github/actions/go-test@main` at L423) omit `apidiff_version`. Anyone following the README now fails at input validation.

Second-order: the examples also need `fetch-depth: 0` on their `actions/checkout`, since `make api-diff` resolves its baseline with `git tag --merged HEAD`. `qualification.yaml` got that in this PR; the documented examples did not.

Minimum correction: add `apidiff_version` to the README's input list and to the four examples, and note the `fetch-depth: 0` requirement alongside them.

tools/api-diff
```diff
@@ -0,0 +1,126 @@
...
+if command -v go >/dev/null 2>&1; then
+    export PATH="$(go env GOPATH)/bin:${PATH}"
+fi
+
+has_tools apidiff git
+
+readonly PACKAGE='github.com/NVIDIA/aicr/pkg/client/v1'
+readonly EXCEPTIONS_FILE="${REPO_ROOT}/pkg/client/v1/api-diff-exceptions.yaml"
+
+baseline="${API_DIFF_BASELINE:-$(git tag --merged HEAD -l 'v*' --sort=-v:refname | grep -v -- '-' | head -1)}"
```
mchmarny: Non-blocking: the friendly error on the next two lines is unreachable. Under `set -euo pipefail`, when no tag matches, `grep` exits 1, `pipefail` propagates it through `head`, and that becomes the status of this assignment — so bash exits here and `if [[ -z "${baseline}" ]]` never runs.

Reproduced with the exact construct:

```bash
$ /bin/bash -c 'set -euo pipefail; b="${API_DIFF_BASELINE:-$(printf "" | grep -v -- "-" | head -1)}"; if [[ -z "$b" ]]; then echo REACHED; fi'
$ echo $?
1
```

The fail direction is safe — the gate still fails — but the operator gets a bare exit 1 with no diagnostic instead of "No stable release tag reachable from HEAD". `head -1` also leaves a latent SIGPIPE/`pipefail` interaction on larger tag lists.

Minimum correction: run the discovery outside the conditional so its status doesn't abort the script, e.g. `baseline="${API_DIFF_BASELINE:-}"` then `if [[ -z "${baseline}" ]]; then baseline="$(git tag ... | grep ... | head -1 || true)"; fi`.

For the record, the rest of the gate holds up — I ran it end to end at this head: it allows the acknowledged `v0.18.0` break, fails with exit 1 when I rescoped that acknowledgement to `v0.17.0`, and the awk parser still resolves the active entry when I appended a trailing second entry.

tests/releasepolicy/release_scripts_test.go
```diff
@@ -28,6 +28,8 @@ import (
 	"time"
 )
 
+const scriptDeadline = 30 * time.Second
```
mchmarny: Non-blocking, but needs coordination before either lands: #2063 adds a `scriptDeadline` **function** (`func scriptDeadline(t *testing.T) time.Duration`, deriving the budget from `t.Deadline()` with a 90s cap) to this same file for the same symptom, under #1974 rather than #2018. Same package, same identifier, two different shapes — whichever merges second conflicts, and a clean auto-merge would be a duplicate declaration.

The branches are also on different bases of this file: `runScript` sits at L1424 here and L1682 there, so expect a textual conflict on top of the symbol clash.

Suggestion: drop the harness change from this PR and let #2063 own it — an adaptive budget subsumes a fixed 30s constant, and it keeps this PR scoped to the API gate its title describes. If instead this one should win, #2063 should be closed in its favor. Either way the two shouldn't both merge as-is.

mchmarny (commented):
**COMMENT** — 0 merge blockers on `6f4e3e0`, 3 non-blocking findings.

**Validated by running the gate, not just reading it.** In a detached worktree at head, with the pinned `apidiff` installed:

- `bash tools/api-diff` resolves baseline `v0.18.0`, reports the additive surface, and exits 0 with the acknowledgement warning for the `BundleOptions` comparability break — the intended happy path.
- Rescoping that acknowledgement to `baseline: v0.17.0` makes it exit **1** with the "Add a baseline-scoped acknowledgement" error. The exception is genuinely baseline-bound; it cannot silently cover a future break.
- Appending a second, unrelated entry after the active one still resolves correctly, so the awk parser survives the file growing — which it will.

That covers the direction that matters for a gate: it fails closed, and the acknowledgement is an exact-match on the change list rather than a wildcard. Wiring is consistent too — `apidiff` is pinned in `.settings.yaml` with a Renovate annotation, surfaced through `load-versions`, installed by `setup-tools`, presence-checked by `check-tools`, and `qualification.yaml` picked up the `fetch-depth: 0` the baseline lookup needs. `qualification.yaml` is the only workflow consuming the `go-test` action, so the new required input doesn't strand an existing caller.

CI on this head is 30 pass / 15 skipping, 0 failing. `BEHIND main` — rebase before merge.

Findings are inline. One smaller note not worth its own thread: the two consecutive `if [[ -n "${acknowledgement}" ]]` guards near the end of `tools/api-diff` can collapse into a single block.

One thing to weigh before merge: `make qualify` now pays a full ~10.6k-file worktree checkout of the baseline tag on every local run. That's the right trade for a mechanical compatibility backstop, but it is a noticeable new cost on the documented pre-push gate — worth a line in `docs/contributor/maintaining.md` so contributors aren't surprised by it.

tools/api-diff
njhensley: **🟡 Minor — F1 — Unreachable "no baseline" diagnostic — silent set -e abort**

The guard on lines 36-38 is dead code. `baseline="${API_DIFF_BASELINE:-$(git tag --merged HEAD -l 'v*' --sort=-v:refname | grep -v -- '-' | head -1)}"` runs the pipeline inside a `${VAR:-...}` default. When no stable tag is reachable, `grep -v` matches nothing and returns 1; under `set -o pipefail` the whole pipeline is non-zero, the command substitution fails, and under `set -e` the assignment aborts the script right here — *before* `if [[ -z "${baseline}" ]]; then err "No stable release tag reachable…"` can run. Reproduced: `set -euo pipefail; x="${U:-$(printf '' | grep -v -- '-' | head -1)}"; echo reached` exits 1 and never prints `reached`. It still fails closed (exit 1), but a contributor on a shallow/tag-less clone gets a bare `set -e` abort with no message — the worst DX path. Fix: compute the baseline in an explicit assignment (append `|| true` to the substitution, or resolve inside an `if`) so the diagnostic on 36-38 becomes reachable.

docs/contributor/maintaining.md
```diff
@@ -27,6 +27,18 @@ The short form:
 
 Bi-weekly cadence; hotfix between cycles when a fix is critical.
 
+### SDK API Compatibility
```
njhensley, line 30: **🟡 Minor — F2 — make qualify gains undocumented local preconditions**

`api-diff` is now part of `make qualify` (Makefile:295), the documented pre-PR gate. Locally it hard-requires two things this section doesn't mention: (1) `apidiff` on PATH — `has_tools apidiff git` exits with `err` if absent, so a contributor whose env predates this PR and hasn't re-run `make tools-setup` fails an unrelated `qualify`; (2) full history + reachable stable tags — a shallow clone can't resolve a baseline (see F1) and `git worktree add <tag>` needs the objects. CI is shielded (qualification.yaml checkout has `fetch-depth: 0`, apidiff installed fresh), so the impact is bounded to local/shallow/stale environments. Add a line to this section (or DEVELOPMENT.md) noting `make qualify` now needs `apidiff` (`make tools-setup`) plus full tag history.

docs/contributor/maintaining.md
```diff
@@ -27,6 +27,18 @@ The short form:
...
+
+`make api-diff` compares the exported `pkg/client/v1` surface to the latest
+stable release tag. It runs through `make qualify` and the qualification
+workflow; additive changes are reported, while removals and incompatible type
+changes fail the gate.
+
+An intentional break must have a baseline-scoped entry in
```
njhensley: **🟡 Minor — F4 — maintaining.md omits how to author an exception**

The gate does an exact, exhaustive set-match: the exception must list *every* incompatible change, each string byte-identical to apidiff's output, or it fails closed and re-blocks. This section says an entry needs "the tracking issue, a short summary, and a rationale" but never mentions the load-bearing `incompatible_changes` field at all, never tells the maintainer to run `make api-diff` and copy each reported `- ` line verbatim (dropping the `- `), and never says the list must be complete + exact. A maintainer landing a legitimate break has to reverse-engineer the format from a stderr dump. Add 2-3 lines: "Run `make api-diff`; copy every line from the reported incompatible-changes block (dropping the leading `- `) into `incompatible_changes` — the gate requires an exact, complete match."

tools/api-diff
```diff
@@ -0,0 +1,126 @@
...
+
+incompatible="$(apidiff -incompatible "${release_export}" "${PACKAGE}")"
+if [[ -z "${incompatible}" ]]; then
+    msg "No incompatible ${PACKAGE} changes since ${baseline}."
+    exit 0
+fi
+
+echo "${incompatible}" >&2
+actual_changes="$(printf '%s\n' "${incompatible}" | sed -n 's/^- //p' | sort)"
+acknowledgement="$(awk -v baseline="${baseline}" '
```
njhensley: **🟡 Minor — F6 — No automated test for the api-diff exception parser**

The awk state machine (72-107) that extracts the baseline-matching acknowledgement and drives the allow/deny decision is the actual compatibility gate, yet it has no test. `make test-shell` exists (globs `tools/*_test.sh`) and the repo tests sibling release scripts exhaustively in `tests/releasepolicy/release_scripts_test.go`. Most standalone bash tools are untested (repo norm → this would be 🔵), but this is the one bash tool whose parser bug could wave an SDK break through, so it earns a flag. Suggest a `tools/api-diff_test.sh` (à la `tools/cleanup_test.sh`) driving the awk with fixture exception files + stubbed apidiff output: exact-match → exit 0; partial/mismatched/wrong-baseline/malformed → non-zero (fail closed). This would also lock in the fail-closed guarantee the meta-review confirmed.

tools/check-tools
```diff
@@ -84,6 +84,9 @@ get_tool_metadata() {
         golangci_lint)
             echo "golangci-lint:::golangci-lint version --short 2>/dev/null:::major"
             ;;
+        apidiff)
+            echo "apidiff:::echo installed:::presence"
```
njhensley: **🔵 Nitpick — F3 — apidiff pin not enforced locally — local/CI drift**

apidiff has no version subcommand, so check-tools uses `presence` (`echo installed`, always passes) and setup-tools reinstalls only when `! command_exists apidiff` (no version compare unless `--upgrade`). A locally-stale apidiff passes `make tools-check` and is never bumped to the pinned pseudo-version, while CI installs the pin fresh each run — so results can diverge purely from apidiff version skew. This is the identical known-class limitation of every `presence`-classed tool (addlicense, go-licenses, grype), not materially worse, so it's a nitpick — but apidiff is more version-sensitive. Optional: have setup-tools always `go install apidiff@${APIDIFF_VERSION}` (idempotent; module cache makes it cheap).

tools/api-diff
njhensley: **🔵 Nitpick — F5 — Bespoke awk YAML parser where yq is already a dep**

The exceptions file is parsed by hand-rolled awk keyed on exact indentation (`^  - baseline:`, `^    issue:`, `^      - `). `yq` is already a required repo tool (load-versions, setup-tools, check-tools all use it), and the awk can't handle flow style, block scalars, comments mid-block, or re-indentation — a future reformat silently breaks matching (fails closed, but confusingly). The "minimize deps" argument doesn't hold when yq is already mandatory. Consider `yq '.acknowledgements[] | select(.baseline == strenv(BASELINE))'`. Works for the committed file today, so nitpick.

njhensley (commented):
## 📋 Multi-Persona Review — `ci(sdk): gate client API compatibility`

**Method:** 4 independent persona reviewers (Correctness/shell · CI-DX · Domain & Architecture · Test-coverage), then every finding re-derived from the resolved code by an adversarial senior meta-reviewer (crux claims reproduced, not assumed).

### Overall assessment

Solid, well-scoped, and — most importantly for a compatibility gate — it **fails closed**. The meta-reviewer independently tried and could *not* construct any exceptions-file / apidiff-output pair that wrongly waves an unacknowledged break through: the allow path requires an exact, sorted, whole-set match, so any extra incompatible change falls through to a hard failure. The load-bearing assumptions all check out — `apidiff -incompatible` lines really are `- `-prefixed (so the comparison isn't a vacuous `empty==empty`), the recorded `v0.18.0` `BundleOptions` exception byte-matches the real non-comparability break (`BinaryAttestation []byte`), CI wiring is complete (`fetch-depth: 0` on the one job that runs the gate; apidiff installed with `GOFLAGS=` so the tool install skips vendor mode while the diff keeps it), and the 30s test-harness bump doesn't weaken the independent sub-4s network-timeout assertions.

No blockers or majors. The four 🟡 minors are a real dead-code branch plus documentation/test gaps that bite precisely when a maintainer next needs to land an intentional break.

- **No fail-open exists** — the allow branch requires `actual_changes == sorted(acknowledged_changes)`; any extra incompatible change makes `actual` a strict superset → fails closed.
- **Compare is non-vacuous** — `apidiff -incompatible` output verified `- `-prefixed, so `actual_changes` is populated.
- **The `v0.18.0` `BundleOptions` exception is genuine and exact** — `BinaryAttestation []byte` (`pkg/client/v1/bundle.go:78`, from #1891) makes the struct non-comparable.
- **30s deadline is safe** — the sub-4s network-timeout tests use their own independent `time.Since`; a hung script still fails.
- **CI wiring complete** — `fetch-depth: 0` on the sole gate-running job; `GOFLAGS=` correctly scoped to the tool install; vendor mode intact for the diff.
- **Scope correct** — `pkg/client/v1` is the sole public SDK surface; the gate complements the compile-time `stability_test.go`.

### Summary

| 0 | 0 | 4 | 6 | 🟢 **Approve with comments** |

Highest-value follow-ups: **F1** (make the dead "no baseline" diagnostic reachable) and the **F4 / F2** docs, so the next maintainer landing an intentional break isn't reverse-engineering the exception format. Per-finding detail is in the inline comments.

tools/api-diff
```diff
@@ -0,0 +1,126 @@
...
+    remainder="${acknowledgement#*$'\t'}"
+    summary="${remainder%%$'\t'*}"
+    acknowledged_changes="${remainder#*$'\t'}"
+    if [[ "${actual_changes}" == "$(printf '%s\n' "${acknowledged_changes}" | sort)" ]]; then
+        log_warning "Allowing acknowledged incompatible API change(s) for ${baseline}: ${issue}: ${summary}"
+        exit 0
+    fi
+fi
+
+if [[ -n "${acknowledgement}" ]]; then
```
njhensley: **🔵 Nitpick — F7 — Redundant consecutive acknowledgement guards**

Lines 119-124 are two consecutive `if [[ -n "${acknowledgement}" ]]` blocks, each emitting one `log_error`. Merge into a single guard with both `log_error` calls.

tools/api-diff
```diff
@@ -0,0 +1,126 @@
...
+        gsub(/^['\''"]|['\''"]$/, "", issue)
+        next
+    }
+    active && /^    summary: / {
+        summary = $0
+        sub(/^    summary: /, "", summary)
+        gsub(/^['\''"]|['\''"]$/, "", summary)
+        next
+    }
+    active && /^      - / {
```
njhensley: **🔵 Nitpick — F8 — awk change-collector not scoped to incompatible_changes:**

`active && /^      - /` appends *any* 6-space-indented list item while the baseline block is active, regardless of which key it sits under. If the exception schema later grows another nested 6-space list, those items would pollute `changes`. Fails closed (set mismatch), so safe, but fragile. Track an `in_changes` flag set on the `    incompatible_changes:` line and cleared on any other 4-space key.

tools/api-diff
```diff
@@ -0,0 +1,126 @@
...
+
+if [[ ! -f "${EXCEPTIONS_FILE}" ]]; then
+    err "API-diff exceptions file not found: ${EXCEPTIONS_FILE}"
+fi
+
+temp_dir="$(mktemp -d "${TMPDIR:-/tmp}/aicr-api-diff.XXXXXX")"
+release_worktree="${temp_dir}/release"
+release_export="${temp_dir}/release.export"
+
+cleanup() {
```
njhensley, line 98: **🔵 Nitpick — F9 — Killed runs leak .git/worktrees registrations**

Cleanup is `trap cleanup EXIT` only. A SIGKILL (or `make -k` abort) bypasses it, leaving a registered worktree under `.git/worktrees/release.*` plus a TMPDIR dir. `mktemp -d …XXXXXX` gives unique names so the next run still works (fail-safe), but stale entries accumulate on dev machines and long-lived self-hosted runners until someone runs `git worktree prune`. Add `git worktree prune >/dev/null 2>&1 || true` before `git worktree add`.

tools/api-diff
njhensley: **🔵 Nitpick — F10 — Stale exceptions never swept; rationale not enforced**

Two small gaps: (1) Once the baseline advances (v0.18.0 → next release), the awk's `active = ($3 == baseline)` stops matching the old entry, which is then silently ignored — the intended "record for release notes" behavior, but nothing ever flags or removes dead entries, so they accumulate. (2) The parser requires `issue`, `summary`, and `incompatible_changes` to be non-empty, but does NOT require `rationale`, even though maintaining.md lists rationale as required — a doc/enforcement mismatch. Consider a lightweight check flagging acknowledgements older than the current baseline, and align the required-field set with the doc.

tools/api-diff
```diff
@@ -0,0 +1,126 @@
...
+# shellcheck source=tools/common
+. "${DIR}/common"
+
+if command -v go >/dev/null 2>&1; then
+    export PATH="$(go env GOPATH)/bin:${PATH}"
+fi
+
+has_tools apidiff git
+
+readonly PACKAGE='github.com/NVIDIA/aicr/pkg/client/v1'
```
yuanchen8911, line 40: apidiff compares only this one package, and the pinned apidiff implementation ([golang/exp@b88d891fe743, apidiff/correspondence.go#L190-L209](https://github.com/golang/exp/blob/b88d891fe743/apidiff/correspondence.go#L190-L209)) matches named types outside the compared package by package path and name only, without comparing their definitions. The facade's exported surface is largely transparent aliases into evolving packages — `CriteriaRegistry` (`pkg/client/v1/types.go:199`), `BundleAttester = attestation.Attester` and `BundleArtifact = *result.Output` (`pkg/client/v1/bundle.go`) — all documented in the stable tier in `docs/integrator/public-api.md`.

Concretely: removing a `CriteriaRegistry` method, changing the `BundleAttester` interface, or removing an exported field from `BundleArtifact`'s target breaks ordinary SDK consumers, but the alias's path and name are unchanged, so this gate stays green and an unacknowledged breaking release can ship.

Gating the whole target packages would be wrong too — that would also freeze unrelated evolving APIs in those packages. Either add scoped checks for the alias-target types, or document the limitation and extend #2019 (which already tracks reconciling the facade alias state) to cover it.

.github/actions/go-test/action.yml
yuanchen8911, line 33: This new required input plus the unconditional `make api-diff` step break the action's documented consumers. The three usage examples in `.github/actions/README.md` (lines 333, 354, and the cross-repo example at 423) still pass only `go_version`/`helm_version`/`coverage_report`. Per GitHub's metadata docs, `required: true` does not reject missing action inputs, so a copied example runs `go install golang.org/x/exp/cmd/apidiff@` with an empty version and fails; and `actions/checkout` fetches a single commit by default, so even with the input supplied the examples have no reachable release tag.

The one real caller (`qualification.yaml`) was updated, so live CI is fine — but please update the in-repo README examples (`apidiff_version` + `fetch-depth: 0`) in this PR, and make the api-diff step skip when the input is empty: the cross-repo example's repository cannot provide AICR's `make api-diff` target, so backward compatibility is the only way that advertised usage keeps working.

tools/api-diff
```diff
@@ -0,0 +1,126 @@
...
+if [[ -z "${incompatible}" ]]; then
+    msg "No incompatible ${PACKAGE} changes since ${baseline}."
+    exit 0
+fi
+
+echo "${incompatible}" >&2
+actual_changes="$(printf '%s\n' "${incompatible}" | sed -n 's/^- //p' | sort)"
+acknowledgement="$(awk -v baseline="${baseline}" '
+    /^  - baseline: / {
+        if (active && issue != "" && summary != "" && changes != "") {
```
yuanchen8911: The parser accepts incomplete acknowledgement entries — a fail-open waiver defect, verified by running this awk block against a test file:

- `rationale` is never read — no branch matches a `rationale:` line, so an entry without one passes, while `docs/contributor/maintaining.md` (added in this PR) requires the tracking issue, a summary, and a rationale.
- YAML `null` passes as content: an entry with `issue: null` and `summary: null` is accepted because awk sees the literal string "null", which is non-empty. yamllint accepts such a file, so nothing upstream catches it.

Since this file is the mechanism for waiving the gate, it should be fail-closed on malformed metadata: parse the YAML structurally (`yq` is already a repo dependency) with typed, non-null validation of `issue`, `summary`, and `rationale`.

Relatedly, this parsing and matching logic has no test coverage — nothing in `tests/` references api-diff, and the gate decision depends on exact string equality of sorted change lists, so a formatting variation in `api-diff-exceptions.yaml` silently stops an acknowledgement from firing. `tests/releasepolicy` is the established pattern for testing release shell scripts (this PR already touches it); please add table-driven cases with the parser fix — valid entry, wrong baseline, null fields, duplicate baseline entries, re-indented file.

tools/api-diff
yuanchen8911: Under `set -euo pipefail`, when no stable `v*` tag is reachable, `grep -v` exits 1, pipefail propagates that through the command substitution, and the bare assignment's exit status aborts the script — so the `err "No stable release tag reachable from HEAD..."` guard on the next lines is dead code. Reproduced: a tagless clone gets a bare exit 1 with no output.

Make the no-match case nonfatal so the empty-baseline check can print the intended instructions — but scope it to the grep (e.g. `grep -v -- '-' || true` inside the pipeline) rather than blanket-suppressing the substitution, so a genuine git failure still surfaces.

tools/setup-tools
```diff
@@ -966,6 +968,17 @@ if [[ "${SKIP_TOOLS}" == "false" ]] && command_exists go; then
         fi
     fi
 
+    # apidiff
+    if command_exists apidiff && [[ "${UPGRADE}" != "true" ]]; then
```
yuanchen8911, line 1002: The presence-only skip here (and the matching `presence` strategy for apidiff in `tools/check-tools`) reuses any pre-existing apidiff binary without ever comparing it to the `.settings.yaml` pin. Reproduced end-to-end with a September 2024 apidiff built under Go 1.23: tools-setup and tools-check both report success, then `make api-diff` fails with "package requires newer Go version go1.26" — local qualification diverging from CI, which installs the exact pin fresh each run.

Local-equals-CI is this PR's stated purpose, so this belongs in the PR rather than a follow-up, even though other tools use the same presence-only pattern. The module version is recoverable via `go version -m "$(command -v apidiff)"` — compare it to the pin and reinstall on mismatch.

tools/api-diff
```diff
@@ -0,0 +1,126 @@
...
+incompatible="$(apidiff -incompatible "${release_export}" "${PACKAGE}")"
+if [[ -z "${incompatible}" ]]; then
+    msg "No incompatible ${PACKAGE} changes since ${baseline}."
+    exit 0
+fi
+
+echo "${incompatible}" >&2
+actual_changes="$(printf '%s\n' "${incompatible}" | sed -n 's/^- //p' | sort)"
+acknowledgement="$(awk -v baseline="${baseline}" '
+    /^  - baseline: / {
```
yuanchen8911: The parser prints the first complete entry for the active baseline and exits; the END block emits at most one entry. Two entries with the same baseline — say two intentional breaks tracked by two issues in one release cycle — contribute only the first entry's `incompatible_changes`, and the exact-match comparison then fails with a misleading "do not exactly match the acknowledgement" error even though the file textually acknowledges everything.

This fails closed, so it's a usability/schema-ambiguity issue rather than a bypass — lower priority, but silently ignoring duplicates is the wrong behavior either way, and it's cheap to settle while replacing the parser: either union complete entries per baseline, or enforce and document exactly one entry per baseline (error on duplicates).

yuanchen8911 (dismissed):
Reviewed at head 6f4e3e00. The gate works on the primary CI path (all checks green at this head, including the new api-diff step in `tests / Test`), and several parts are done well: the detached-worktree comparison never mutates the working tree, exceptions are baseline-scoped so they expire at the next release, the exact-match rule prevents one acknowledgement from masking additional breaks, and the version wiring (`.settings.yaml` → load-versions → go-test → qualification.yaml → `make qualify`) keeps local and CI in sync.

Seven issues, in priority order. Details in the inline comments (item 6 below has no inline anchor because the file is not part of this diff).

1. The gate misses breaking changes behind the facade's transparent type aliases — apidiff compares only `pkg/client/v1` and matches external named types by path and name without comparing their definitions. Needs either a scoped fix or an explicitly tracked limitation (`tools/api-diff`).
2. The three documented go-test invocations in `.github/actions/README.md` were not updated for the new required `apidiff_version` input and would fail if copied; the cross-repo example additionally needs the action to stay backward-compatible (`.github/actions/go-test/action.yml`).
3. The acknowledgement parser accepts incomplete entries — `rationale` is never read, and `issue: null` / `summary: null` pass as valid — a fail-open waiver defect. The fix should land with tests for the parser and matching logic, which currently have none (`tools/api-diff`).
4. With no reachable stable tag, the script exits silently before the intended "No stable release tag" diagnostic can print (`tools/api-diff`).
5. setup-tools and check-tools accept any pre-existing apidiff binary without comparing it to the pin, so local qualification can silently diverge from CI — directly against this PR's local-equals-CI purpose (`tools/setup-tools`).
6. `docs/contributor/tests.md` (the canonical "`make qualify` Gate" section, around line 479) still documents `make qualify` as five steps — test-coverage, lint, e2e, scan, license-check — and omits api-diff. Since qualify now requires the pinned apidiff tool and reachable release-tag history, contributors reading this page will hit unexplained local failures. Please add the api-diff step there alongside the maintaining.md section.
7. Only the first acknowledgement entry per baseline is used; a duplicate baseline entry is silently ignored. Fails closed, so lower priority, but cheap to settle while fixing the parser (`tools/api-diff`).

Items 2–6 should be fixed in this PR, plus either a scoped alias check or a documented limitation with follow-up tracking for item 1. Item 7 is lower priority.

.github/actions/go-test/action.yml
tjrasche: Resolved: `apidiff_version` is optional and both API-diff steps are conditional. In-repo examples provide the pin with `fetch-depth: 0`; the cross-repo example omits it to preserve compatibility.

tools/api-diff
tjrasche: Resolved: stable tags are now discovered explicitly, so no match produces the intended diagnostic while genuine `git tag` failures retain their status.

tjrasche: Resolved: tag discovery no longer relies on the failing `grep | head` substitution, so tagless clones reach the clear no-baseline error.

docs/contributor/maintaining.md
tjrasche: Resolved: the contributor docs now call out `make tools-setup`, full tag history, a reachable stable tag, and the temporary-worktree cost.

docs/contributor/maintaining.md
tjrasche: Resolved: the docs now require running `make api-diff` and copying the complete, exact incompatible-change set without its leading list markers.

tools/api-diff
tjrasche: Resolved: `tools/api-diff_test.sh` now covers valid, mismatched, malformed, null, duplicate, re-indented, wrong-baseline, and tag-discovery cases.

tools/check-tools
tjrasche: Resolved: the installed module version is read from Go build metadata and compared exactly with the repository pin.

tools/api-diff
tjrasche: Resolved: acknowledgement parsing now uses structural, typed `yq` validation instead of indentation-sensitive awk parsing.

tools/api-diff
tjrasche: Resolved: the consecutive guards were removed; baseline counting and exact change matching now use one decision path.

tools/api-diff
tjrasche: Resolved: `git worktree prune` now runs before creating the baseline worktree, cleaning registrations left by interrupted runs.

tools/api-diff
tjrasche: Resolved: `rationale` is required and non-empty; stale baselines warn on a clean diff and fail when incompatibilities exist.

.github/actions/go-test/action.yml
tjrasche: Resolved: API-diff inputs and steps are optional, local examples use full history and the pin, and cross-repository consumers retain the original behavior.

tools/api-diff
tjrasche: Resolved: typed `yq` validation rejects null or incomplete metadata, and decision-flow tests cover malformed, duplicate, and re-indented files.

tools/api-diff
tjrasche: Resolved: explicit tag iteration handles the no-match case without suppressing genuine Git failures.

tools/setup-tools
tjrasche: Resolved: setup and verification now extract the exact apidiff module version; mismatches trigger reinstall or failure instead of silently passing.

tools/api-diff
tjrasche: Resolved: exactly one acknowledgement per baseline is now enforced with a dedicated diagnostic and regression tests.

tools/api-diff
tjrasche: Resolved via the documented-and-tracked option: the limitation is documented in this PR, and #2019 now explicitly tracks the alias-target `apidiff` blind spot, scoped compatibility checks or alias removal/wrapping, and regression-test acceptance criteria.

tests/releasepolicy/release_scripts_test.go
tjrasche: Obsolete: this PR no longer changes `tests/releasepolicy/release_scripts_test.go`, and #2063 was closed without merging, so the symbol/conflict risk no longer exists.

tools/api-diff
tjrasche: Obsolete: the indentation-sensitive awk acknowledgement collector was replaced with structural, typed `yq` parsing, so this scoping concern no longer applies.

mchmarny (approved):
Approved at 875d31fe: prior feedback is addressed, all review threads are resolved, focused API-diff decision/version/setup tests pass, and current CI is fully green (28 pass, 0 fail, 0 pending). No blockers found.

PR #2061, opened 2026-08-05
fix(bundler): tracked -post release for vendored mixed components

### LINKED ISSUE #1835: Vendored bundles: recipe-side manifests are install-only Helm hooks
## Summary

Under `--vendor-charts`, every recipe-side manifest of a mixed Helm component is injected into the vendored wrapper chart as a `helm.sh/hook: post-install` resource. Helm does not manage hook resources as part of the release: they are not re-applied on `helm upgrade` (unless the manifest itself declares a `post-upgrade` hook), and they are left behind by `helm uninstall`. Vendored bundles therefore ship recipe-side manifests with an install-only lifecycle.

## Mechanism

- [`writeMixedManifests`](https://github.com/NVIDIA/aicr/blob/8b72e5119928fd61de5e9cab8a4943064f5a279f/pkg/bundler/deployer/localformat/vendor_folder.go#L236) runs each rendered recipe-side manifest through [`injectPostInstallHooks`](https://github.com/NVIDIA/aicr/blob/8b72e5119928fd61de5e9cab8a4943064f5a279f/pkg/bundler/deployer/localformat/vendor_wrapper.go#L159) ([call site](https://github.com/NVIDIA/aicr/blob/8b72e5119928fd61de5e9cab8a4943064f5a279f/pkg/bundler/deployer/localformat/vendor_folder.go#L285)), which adds `helm.sh/hook: post-install` plus `helm.sh/hook-delete-policy: before-hook-creation` to any doc without author-declared hook annotations.
- Post-install hooks fire on install only. Upgrades skip them, and uninstall does not delete them; `before-hook-creation` only replaces the resource the next time the hook fires.

## Impact

The motivating case is the AKS `nvidia-peermem-reloader` DaemonSet added in #1756: privileged, mounts the host root, and re-asserts `modprobe nvidia-peermem` via chroot. In a vendored bundle it is never updated on `helm upgrade` and keeps running after `helm uninstall`. Every mixed component's recipe-side manifests (for example the network-operator NicClusterPolicy) share the same lifecycle.

The default (non-vendored) path is unaffected — the folder deployer wraps the same manifests in a normal `<component>-post` Helm release with full upgrade and uninstall lifecycle.

## Current disposition

Documented as a vendored-mode trade-off in `docs/user/bundling.md` (added in #1756), including the manual cleanup path after uninstall: `kubectl delete -f <bundle>/<NNN>-<component>/templates/`.

## Possible directions

- Emit recipe-side manifests of vendored mixed components as a separate managed `<component>-post` wrapper release, mirroring the non-vendored path, instead of injecting hooks.
- Or, as a smaller step, inject `post-install,post-upgrade` so upgrades re-apply the manifests, and document only the uninstall residue.
- Either way, continue honoring author-declared hook annotations (for example the CRD-ordering hooks on `nic-cluster-policy-aks.yaml`).

Related: #1756 (docs note and motivating manifest).

rorajani:
## Summary
Replace `--vendor-charts` Helm-hook injection for mixed-component recipe-side manifests with the same tracked `<component>-post` Helm release the non-vendored path already uses.

This closes upgrade drift, uninstall residue, and Argo CD PostSync silent-drop **without** the delete-then-recreate hazard of adding `post-upgrade` to `before-hook-creation` hooks (blocking review on the earlier approach).

## Motivation / Context
Fixes: #1835
Related: review on #2061 (mchmarny, njhensley) — Option B / structural fix

## Type of Change

## Component(s) Affected

## Implementation Notes
- `localformat.Write` now injects `-post` under `VendorCharts` (same as upstream-Helm path).
- Removed `injectPostInstallHooks` / `writeMixedManifests`; vendored primary wraps only the tarball.
- Author-declared `helm.sh/hook` annotations on recipe manifests are **stripped** on the `-post` path via existing `stripHelmHooks` — matching non-vendored. Ordering is preserved by `NNN-` folder sequencing.
- Docs rewritten; broken `#kueue` anchor removed.

## Testing
```bash
make qualify   # pass (api-diff after installing pinned apidiff)
# local smoke: aicr bundle --vendor-charts → primary + -post, no helm.sh/hook
```
Coverage: `pkg/bundler/deployer/localformat` 75.7% → 75.4% (-0.3%)

## Risk Assessment
- [x] **Medium** — layout change for existing vendored mixed bundles (extra `-post` folder); operators with old hook-based installs may need one-time `kubectl delete` of orphaned hook resources (documented).

## Checklist


pkg/bundler/deployer/localformat/vendor_wrapper.go
```diff
@@ -225,7 +237,7 @@ func injectHookOnDoc(doc map[string]any, weight int) bool {
 		// post-upgrade, etc.). Don't overwrite, don't consume a weight.
 		return false
 	}
-	annos["helm.sh/hook"] = "post-install"
+	annos["helm.sh/hook"] = "post-install,post-upgrade"
```
mchmarny: **Blocking — data loss on `helm upgrade` for cascade-delete kinds.**

This function still stamps `helm.sh/hook-delete-policy: before-hook-creation` a few lines down ([L246-247](https://github.com/NVIDIA/aicr/blob/0920e5acc859bf92b35e1f1fffdcfd7f355e9259/pkg/bundler/deployer/localformat/vendor_wrapper.go#L246-L247)). Adding `post-upgrade` therefore doesn't "re-apply" the manifest — it makes Helm **delete then recreate** it on every upgrade. `docs/user/component-catalog.md:37` already documents that exact semantic for the Kueue quota CRs; this change extends it to every hookless recipe-side manifest, including kinds whose deletion cascades.

Reproduced at this head — `aicr bundle -r <h100-eks-inference> --vendor-charts` emits:

```yaml
# 001-agentgateway-crds/templates/gateway-api-crds.yaml
kind: CustomResourceDefinition   # gatewayclasses, gateways, grpcroutes, httproutes, referencegrants
metadata:
  annotations:
    helm.sh/hook: post-install,post-upgrade
    helm.sh/hook-delete-policy: before-hook-creation
```

`001-agentgateway-crds/install.sh` runs `helm upgrade --install agentgateway-crds ./`, so a second `./deploy.sh` deletes those five Gateway API CRDs; the apiserver garbage-collects every user-authored `Gateway`/`HTTPRoute`/`GRPCRoute`/`InferencePool`, and the CRDs are recreated empty. Not recoverable. Same shape for [`talos-namespace.yaml`](https://github.com/NVIDIA/aicr/blob/0920e5acc859bf92b35e1f1fffdcfd7f355e9259/recipes/components/network-operator/manifests/talos-namespace.yaml) (`kind: Namespace`, cascades to everything in it) and [`compute-domain.yaml`](https://github.com/NVIDIA/aicr/blob/0920e5acc859bf92b35e1f1fffdcfd7f355e9259/recipes/components/slinky-slurm/manifests/compute-domain.yaml). These are hookless because `TestManifestHelmHooksRequired` only requires hooks for custom apiVersions.

CI can't catch this: `tests/chainsaw/cli/bundle-vendor-charts` inspects the emitted files but never performs a second `helm upgrade`.

There's also no way for an author to opt out via the delete-policy escape at L243-247 — Helm parses only `before-hook-creation`/`hook-succeeded`/`hook-failed` and defaults to `before-hook-creation` when nothing matches, so a value like `keep` is silently ignored.

Minimum correction: keep `post-install` only for kinds whose deletion cascades (`CustomResourceDefinition`, `Namespace`), injecting `post-install,post-upgrade` for the rest — plus a regression test asserting a CRD doc and a Namespace doc do **not** come back with `post-upgrade` alongside `before-hook-creation`.

docs/user/bundling.md
```diff
@@ -154,10 +154,10 @@ aicr bundle --recipe recipe.yaml --vendor-charts --output ./bundles
 > manifests shipped alongside a vendored upstream chart — for example the
 > network-operator NicClusterPolicy or the AKS `nvidia-peermem-reloader`
 > DaemonSet) are injected into the vendored chart as `helm.sh/hook:
-> post-install` resources. Helm does not manage hook resources as part of the
-> release: they are not re-applied on `helm upgrade` (unless the manifest
-> declares a `post-upgrade` hook itself) and are left behind by
-> `helm uninstall`. Remove them manually after uninstalling a vendored bundle
+> post-install,post-upgrade` resources. They apply on install and are
+> re-applied on `helm upgrade`, but Helm does not track hook resources as
```
mchmarny: Non-blocking: "re-applied on `helm upgrade`" reads as an in-place apply, but with the injected `helm.sh/hook-delete-policy: before-hook-creation` it's a delete followed by a recreate. Worth stating explicitly and mirroring the operator warning the Kueue entry already carries in `docs/user/component-catalog.md` ("quiesce before upgrading… upgrades reset to shipped defaults"), so operators know an upgrade churns the resource rather than patching it.

mchmarny (changes requested):
**REQUEST_CHANGES** — 1 blocker, 1 non-blocking.

Validated at `0920e5a` by building `aicr` at head, generating a real `--vendor-charts` bundle for the `h100-eks-inference` recipe, and inspecting the emitted `templates/`. `go test ./pkg/bundler/deployer/localformat/...` passes.

CI is green but is not evidence for this change: the only `--vendor-charts` e2e (`tests/chainsaw/cli/bundle-vendor-charts`) bundles and file-inspects, and never runs a second `helm upgrade` — which is exactly the path this PR newly activates.

The blocker is the interaction with the `helm.sh/hook-delete-policy: before-hook-creation` that the same function stamps: adding `post-upgrade` turns each upgrade into a delete-then-recreate, which for the hookless CRD and Namespace manifests means cascading deletion of user data. Details inline.

On @coderabbitai's delete-policy thread: confirmed as a real defect rather than a verification ask, so raising the severity accordingly.

pkg/bundler/deployer/localformat/vendor_wrapper.go
njhensley: **🔴 Blocker — post-install,post-upgrade + before-hook-creation deletes-then-recreates recipe-side manifests on every helm upgrade; cascades to data loss for CRD/Namespace kinds**

This line sets `helm.sh/hook: post-install,post-upgrade`, and L246-247 also stamps `helm.sh/hook-delete-policy: before-hook-creation` on every hookless manifest. `vendor_folder.go:285` applies this kind-blind to all `--vendor-charts` mixed-component manifests (NOT stripped, unlike the non-vendored path at `local_helm.go:168`). `install.sh` deploys via `helm upgrade --install`, so adding `post-upgrade` fires the hook on every upgrade and `before-hook-creation` deletes the existing resource before recreating it. Pre-PR (post-install only) the hook never fired on upgrade, so this destructive path is newly introduced. This is the same defect @mchmarny flagged, independently re-derived.

**Blast radius:** `agentgateway-crds` is a mixed component (helm chart `oci://cr.agentgateway.dev/charts` + `gateway-api-crds.yaml`/`inference-extension-crds.yaml`, wired in `recipes/mixins/platform-inference.yaml`) — deleting a CustomResourceDefinition GCs every Gateway/HTTPRoute/GRPCRoute/InferencePool cluster-wide. `network-operator/manifests/talos-namespace.yaml` (Namespace) cascades to all contents; `slinky-slurm/manifests/compute-domain.yaml` tears down the IMEX domain. Not recoverable.

**Fix:** Make injection kind-aware: for cascade-owning/cluster-scoped kinds (CustomResourceDefinition, Namespace, and arguably CRs/PVCs) keep `post-install` only, drop `before-hook-creation`, or emit them as ordinary wrapper-chart templates ordered by Helm's kind InstallOrder so upgrades patch in place. Ship with a regression test asserting a CRD and a Namespace doc do NOT get `post-upgrade` alongside `before-hook-creation`.

pkg/bundler/deployer/localformat/vendor_wrapper.go
```diff
@@ -225,7 +237,7 @@ func injectHookOnDoc(doc map[string]any, weight int) bool {
...
 	annos["helm.sh/hook-weight"] = strconv.Itoa(weight)
 	// Honor an author-declared delete-policy for the same reason we
```
njhensley: **🟡 Minor — Comment documents `keep` as a helm.sh/hook-delete-policy escape hatch, but `keep` is not a valid value for that annotation**

The comment here (through L245) tells authors they can pre-set `helm.sh/hook-delete-policy: keep` to preserve install-time state across upgrades. But `keep` is not a valid hook-delete-policy value (valid: before-hook-creation, hook-succeeded, hook-failed); it is the value of a DIFFERENT annotation, `helm.sh/resource-policy`. Helm ignores the unrecognized value and defaults to before-hook-creation, so the documented opt-out does nothing. (Pre-existing comment, but it now sits next to the newly-added upgrade semantics.)

**Blast radius:** An author following the comment to protect a stateful manifest gets silent delete-recreate anyway. The real opt-out is declaring `helm.sh/hook` explicitly (the L235 alreadyHooked short-circuit), which the comment doesn't mention.

**Fix:** Drop the `keep` reference, or redirect to `helm.sh/resource-policy: keep` on a non-hook template; mention the working author-set-`helm.sh/hook` opt-out.

pkg/bundler/deployer/localformat/vendor_wrapper.go
```diff
@@ -132,10 +135,19 @@ func InjectPostInstallHooks(in []byte) ([]byte, error) {
...
+// `helm.sh/hook: post-install,post-upgrade` and a stable per-document
+// hook-weight on every top-level resource that doesn't already declare
+// a hook. Used in mixed-component vendored wrappers so the recipe-side
+// raw manifests apply AFTER the vendored subchart's resources on both
+// `helm install` and `helm upgrade`.
+//
+// Why both phases: a post-install-only hook fires on the initial
+// install but Helm skips it on subsequent `helm upgrade` runs, so a
+// version bump on the recipe-side manifest never reaches the cluster.
+// Adding post-upgrade makes upgrades re-apply the manifest. Uninstall
```
njhensley: **🟡 Minor — godoc/header say upgrades "re-apply" the manifest — actually delete-then-recreate**

The header comment (L32-33) and this `injectPostInstallHooks` godoc block, both edited by this PR, describe the upgrade as apply/"re-apply the manifest" with no mention that `before-hook-creation` (L246-247) makes each upgrade a destructive delete-then-recreate. Not raised by the other reviewers (their wording note was about the user docs, not these Go comments).

**Blast radius:** Maintainers reading these godocs to reason about hook semantics under-estimate the upgrade blast radius (transient teardown, owned-object cascade).

**Fix:** State explicitly that, because delete-policy is before-hook-creation, each upgrade deletes and recreates the resource; cross-reference the operator churn warning.

docs/user/bundling.md
```diff
@@ -154,10 +154,10 @@ aicr bundle --recipe recipe.yaml --vendor-charts --output ./bundles
 > manifests shipped alongside a vendored upstream chart — for example the
 > network-operator NicClusterPolicy or the AKS `nvidia-peermem-reloader`
 > DaemonSet) are injected into the vendored chart as `helm.sh/hook:
-> post-install` resources. Helm does not manage hook resources as part of the
-> release: they are not re-applied on `helm upgrade` (unless the manifest
-> declares a `post-upgrade` hook itself) and are left behind by
-> `helm uninstall`. Remove them manually after uninstalling a vendored bundle
+> post-install,post-upgrade` resources. They apply on install and are
```
njhensley: **🟡 Minor — bundling.md "re-applied on helm upgrade" hides destructive delete-then-recreate churn/reset**

The changed note says manifests apply on install and are 're-applied on helm upgrade', omitting that the injected `before-hook-creation` delete policy tears the resource down and resets it to shipped defaults on every upgrade. The repo's own `component-catalog.md` Kueue entry documents this identical mechanism correctly (quiesce before upgrading, upgrades reset to shipped defaults). (Echoes @mchmarny's non-blocking note.)

**Blast radius:** Operators get no warning to quiesce/drain before upgrading and no notice that hand-tuned state is reset; inconsistent with the project's established phrasing for the same behavior.

**Fix:** Mirror the Kueue warning: re-created (delete-then-recreate) on every helm upgrade, transient teardown of the resource, post-install drift reset to the shipped manifest.

pkg/bundler/deployer/localformat/vendor_wrapper_test.go
```diff
@@ -102,8 +102,12 @@ data:
 		t.Fatalf("re-parse: %v\n%s", err, got)
 	}
 	annos := node["metadata"].(map[string]any)["annotations"].(map[string]any)
-	if annos["helm.sh/hook"] != "post-install" {
-		t.Errorf("hook = %v, want post-install", annos["helm.sh/hook"])
+	// Both phases: post-install for the initial deploy AND post-upgrade
+	// so `helm upgrade` re-applies the recipe-side manifest. A single
+	// post-install phase leaves the manifest stale across upgrades
```
njhensley: **🟡 Minor — No regression test for cascade-delete kinds (CRD/Namespace) through the hook-injection path**

This assertion (and every test driving injectHookOnDoc: TestInjectPostInstallHooks_*, TestWrite_VendorCharts_Mixed) uses only ConfigMap/ClusterRole — neither cascades. Namespace manifests elsewhere in the suite flow through the stripped pre-manifest path, never injectHookOnDoc. The PR only flips expected strings from post-install to post-install,post-upgrade; nothing pins the intended (safe) treatment of a CRD/Namespace routed through injection.

**Blast radius:** The exact real-world manifests this fix targets have no test guarding against the blocker's data-loss path. Any kind-aware fix MUST ship with such a test.

**Fix:** Add a table-driven test feeding a CRD and a Namespace doc through injectPostInstallHooks and asserting the delete-policy/hook phase differs from the ConfigMap case (or, if uniform is intentional, pin that as a conscious contract).

pkg/bundler/deployer/localformat/vendor_wrapper.go
```diff
@@ -132,10 +135,19 @@ func InjectPostInstallHooks(in []byte) ([]byte, error) {
 }
 
 // injectPostInstallHooks rewrites a multi-document YAML stream to add
```
njhensley: **🟡 Minor — Vendored path uses untracked Helm hooks where the non-vendored path deploys tracked release members; divergence widened**

Non-vendored (local_helm.go:168 stripHelmHooks) makes recipe-side manifests ordinary members of the `<component>-post` release (three-way-merge patch on upgrade, deleted on uninstall). Vendored makes them Helm hooks: not tracked as release members, survive `helm uninstall`, and only upgrade-repeatable via before-hook-creation (delete-recreate). This PR widens the divergence — the uninstall-leaking residue is now re-created on every upgrade, not once. (Not raised by the other reviewers.)

**Blast radius:** Operational surprise across all vendored-mixed components: orphaned hook resources after uninstall + per-upgrade churn.

**Fix:** Prefer emitting recipe-side manifests as regular wrapper-chart templates ordered by Helm kind InstallOrder so the vendored path matches the non-vendored tracked lifecycle; or wire the manual uninstall-cleanup path into undeploy.sh.

njhensley: **🔵 Nitpick — Rename injectPostInstallHooks / InjectPostInstallHooks — it now injects post-upgrade too**

The helper injects post-install AND post-upgrade; the PostInstall name is stale. (Also raised by CodeRabbit.)

**Blast radius:** Reader confusion only.

**Fix:** Rename to a lifecycle-neutral name (e.g. injectPostLifecycleHooks) and the TestInjectPostInstallHooks_* tests to match. Optional / defer.

njhensley (commented):
📋 **Multi-persona review — head `0920e5a`**

▎ Method: 4 persona passes (Correctness, Domain & Architecture, Docs, Test-coverage), each finding independently confirmed or refuted by a senior meta-reviewer against the resolved code. Line links pinned to head `0920e5a`.

> ⚠️ **@mchmarny already filed CHANGES_REQUESTED** on this PR. My 🔴 blocker is the **same defect**, independently re-derived from the code — I credit it rather than claim it. My additive signal over the existing reviews is narrow: the `keep`-comment defect (🟡), the misleading Go doc-comments this PR added (🟡), the tracked-vs-untracked lifecycle divergence (🟡), and two refutations of CodeRabbit findings (below). The blocker itself is corroboration, and I widen its blast radius to **CustomResourceDefinitions** (`agentgateway-crds` is a mixed component, so its CRD manifests get the same annotations).

## Overall assessment — Comment (defer to @mchmarny's standing Request-changes)

Adding `post-upgrade` correctly makes `helm upgrade` re-fire the hook (fixing #1835's stale DaemonSet). But the same function also stamps `helm.sh/hook-delete-policy: before-hook-creation`, so re-firing is a **delete-then-recreate**, and injection is kind-blind — for the hookless cascade-delete manifests on the `--vendor-charts` path (`agentgateway-crds` CRDs, `talos-namespace` Namespace, slinky `compute-domain` CR) every `./deploy.sh` re-run cascade-deletes user data. That's the blocker. Everything else is doc/comment/test polish subordinate to fixing it.

Inline findings follow. One finding can't be inline (its file isn't in this diff):

**🟡 Minor — the `--vendor-charts` e2e never runs a second `helm upgrade`.** `tests/chainsaw/cli/bundle-vendor-charts/chainsaw-test.yaml` uses a fake helm (`version`/`pull` only), never installs, never upgrades — verification is a static grep of emitted annotations. The behavior this PR fixes has zero execution-level coverage, which is why the blocker passed green CI. Acceptable as a hermetic bundling test, but note the blind spot; ideally add a KWOK/kind lane that installs then upgrades a vendored mixed bundle and asserts re-apply (and non-destruction of a CRD/Namespace variant).

---

- **CodeRabbit — "`leaves them behind by helm uninstall`" is obsolete: REFUTED.** That sentence is in the `--vendor-charts` block, which *injects* hooks and does **not** call `stripHelmHooks`. Injected hook resources genuinely aren't tracked as release members and do survive `helm uninstall`, so the statement is correct. CodeRabbit conflated it with the non-vendored (stripped) `local_helm.go` path.
- **CodeRabbit — Argo CD "PostSync never fires" for vendored mixed: REFUTED / out of scope.** The vendored folder is consumed by the localformat helm `install.sh`, not Argo CD; the wrapper always carries the subchart's real (non-hook) resources so a sync fires; and `post-install` already mapped to PostSync pre-PR, so adding `post-upgrade` introduces no new Argo behavior.
- **Delete-policy stamping branches are covered** — `TestInjectPostInstallHooks_AddsAnnotations` / `_HonorsAuthorDeletePolicy` / `_PreservesAuthorHook`. The gap is the cascade-kind/upgrade interaction, not the branches.
- **`docs/user/cli-reference.md`** change (`post-install` → `post-install,post-upgrade`) is accurate.

---
## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 1 | delete-then-recreate cascade data loss on upgrade (CRD/Namespace/CR) — same as @mchmarny |
| 🟠 Major   | 0 | — |
| 🟡 Minor   | 6 | invalid `keep` escape-hatch comment · godoc "re-apply" wording · bundling.md churn warning · missing cascade-kind test · e2e no-upgrade gap · hook-vs-tracked lifecycle divergence |
| 🔵 Nitpick | 1 | rename `injectPostInstallHooks` |

**Recommendation:** the fix is kind-aware injection (skip `post-upgrade`/`before-hook-creation` for cascade-owning kinds) **plus** a regression test; the doc/comment minors then fall out naturally. Posting as a Comment since @mchmarny's Request-changes already gates the merge.

docs/user/bundling.md
```diff
@@ -152,15 +152,41 @@ aicr bundle --recipe recipe.yaml --vendor-charts --output ./bundles
...
-> `helm uninstall`. Remove them manually after uninstalling a vendored bundle
-> (`kubectl delete -f <bundle>/<NNN>-<component>/templates/`). The default
-> (non-vendored) path wraps the same manifests in a normal `<component>-post`
-> Helm release with full upgrade and uninstall lifecycle.
+> post-install` resources with `helm.sh/hook-delete-policy:
+> before-hook-creation`. Because Helm treats hook resources as fire-and-forget
+> rather than tracked release members, this shape carries three operator-facing
+> limitations that the default (non-vendored) path does not have:
+>
+> - **Upgrade drift.** A `post-install`-only hook does not fire on
```
mchmarny: **Blocking - this bullet drops the accuracy carve-out the previous text carried, and its own named example contradicts it.**

The sentence this replaces read "not re-applied on `helm upgrade` (unless the manifest declares a `post-upgrade` hook itself)". Without that carve-out the bullet is false for most shipped manifests: `injectHookOnDoc` short-circuits on `alreadyHooked` and leaves author-declared hooks untouched ([vendor_wrapper.go#L223-L226](https://github.com/NVIDIA/aicr/blob/eb1bb225d69a80f28548962de67eb78d509a8841/pkg/bundler/deployer/localformat/vendor_wrapper.go#L223-L226)), pinned by `TestInjectPostInstallHooks_PreservesAuthorHook`.

Of the 45 manifests under `recipes/components/*/manifests/`, 27 declare their own hook and 21 declare `post-install,post-upgrade`. That includes this paragraph's first example, the network-operator `NicClusterPolicy` at `recipes/components/network-operator/manifests/nic-cluster-policy-aks.yaml:37`, which does fire on `helm upgrade`. Only the second example (`nvidia-peermem-reloader.yaml`) is hookless and actually exhibits the drift, so the two manifests named side by side, both from the AKS network-operator component, behave oppositely.

Minimum fix: restore the "unless the manifest declares a `post-upgrade` hook itself" qualifier and scope the bullet to hookless manifests.

docs/user/bundling.md
```diff
@@ -152,15 +152,41 @@ aicr bundle --recipe recipe.yaml --vendor-charts --output ./bundles
...
+>   Remove them manually after uninstalling a vendored bundle:
+>   `kubectl delete -f <bundle>/<NNN>-<component>/templates/`.
+> - **Argo CD deploy silently drops the manifests.** Argo maps
+>   `helm.sh/hook: post-install` to `argocd.argoproj.io/hook: PostSync`.
+>   Under `syncPolicy.automated`, a chart whose only resources are
+>   hook-annotated appears trivially "Synced" and the PostSync hook never
+>   fires — recipe-side manifests are silently never applied.
+>
+> Note that `helm.sh/hook-delete-policy: before-hook-creation` also means any
+> firing of the hook (e.g., re-installing the bundle or a hypothetical future
```
mchmarny: **Blocking - `post-upgrade` delete-then-recreate is live today, not hypothetical.**

These manifests already ship `helm.sh/hook: post-install,post-upgrade` plus `helm.sh/hook-delete-policy: before-hook-creation`, and pass through the vendored `templates/` unmodified:

- `recipes/components/kueue/manifests/resource-flavor.yaml:36-38` (also `cluster-queue.yaml`, `local-queue.yaml`)
- `recipes/components/agentgateway/manifests/inference-gateway.yaml:32-34`
- `recipes/components/kubeflow-trainer/manifests/torch-distributed-cluster-training-runtime.yaml:22-24`

`helm upgrade` deletes and recreates them on every run right now. `docs/user/component-catalog.md:37` documents exactly that for the Kueue quota CRs ("quiesce queues before upgrading ... upgrades reset them to the shipped defaults"), so as written this page tells operators the opposite of the catalog, and withholds the quiesce warning from precisely the operators who need it.

Minimum fix: describe the delete-then-recreate as current behavior for author-hooked manifests and cross-link the Kueue quiesce guidance, instead of framing it as a future variant.

docs/user/bundling.md
```diff
@@ -152,15 +152,41 @@ aicr bundle --recipe recipe.yaml --vendor-charts --output ./bundles
...
+> `post-upgrade` variant) *deletes* the resource before recreating it. For
+> stateful or ownership-bearing kinds (CRDs whose deletion cascades to CRs,
+> Namespaces with owned objects, resources with operator-tuned state) this is
+> destructive; the Kueue component
+> [documents this exact pattern](component-catalog.md#kueue) and instructs
```
mchmarny: **Non-blocking - broken anchor.** `component-catalog.md#kueue` does not resolve: there is no `Kueue` heading in that file, the content is a table row at `docs/user/component-catalog.md:37`. GitHub and Fern generate anchors from headings only, so the link lands at the top of the page. CI will not catch it, since lychee runs `--offline` without `--include-fragments`. Suggest linking to `component-catalog.md#components` and naming the row in prose.

mchmarny (commented):
**COMMENT — 3 blockers, 2 non-blocking.**

Re-reviewed at head `eb1bb22`. The PR is now docs-only (three-dot diff vs `main` is `docs/user/bundling.md`, +34/-8); the Go change that drew my earlier `CHANGES_REQUESTED` is gone, so that blocker is resolved by removal. Every finding from the prior round on `vendor_wrapper.go` is moot.

Validated by reading the code the paragraph describes, not the diff alone: `injectHookOnDoc`'s `alreadyHooked` short-circuit, `stripHelmHooks`'s single caller (`local_helm.go:168`, non-vendored only, so the Argo CD bullet is accurate and reachable), and `go test -run TestInjectPostInstallHooks ./pkg/bundler/deployer/localformat/` (8/8 pass). CI is green but is not evidence for this change: no check validates a doc statement against the annotations actually emitted, and lychee runs `--offline` without `--include-fragments`.

Two blockers are inline. The third has no changed-line anchor:

**Blocking - `Fixes #1835` will auto-close an issue this PR explicitly defers.** #1835 is open, and the text this PR adds says the structural fix is "tracked as follow-up work; see issue #1835" ([L189](https://github.com/NVIDIA/aicr/blob/eb1bb225d69a80f28548962de67eb78d509a8841/docs/user/bundling.md#L189)). Merging as written ships a doc pointing readers at an issue the same merge closes. Change the PR body to `Related: #1835`.

**Non-blocking - stale title and body.** The title is still `fix(bundler): re-apply vendored mixed-component manifests on helm upg…` for a change that no longer touches `pkg/bundler`; it lands verbatim in the squash commit and the generated changelog. Suggest `docs(bundling): document vendored mixed-component hook limitations`. The body is the previous code-change commit message followed by an unfilled template.

Separately: `mergeStateStatus` is `BEHIND` (main is at `b52aefb`), and my earlier `CHANGES_REQUESTED` still gates merge. I will clear it once the above is addressed.

CodeRabbit's Aug 10 comment overlaps the two inline blockers. Its stated rationale conflates the PR objective with the removed code change, but its observation about `nfd-network-rule.yaml` is correct.

pkg/bundler/deployer/localformat/vendor_wrapper.go
rorajani: Thanks — agreed. Dropped the `post-upgrade` + `before-hook-creation` approach.

This PR now takes Option B / structural fix (#1835): vendored mixed components emit a tracked `<component>-post` Helm release (same as non-vendored) instead of injecting `helm.sh/hook` resources. Upgrades three-way-merge in place; uninstall removes the resources; no delete-then-recreate on CRD/Namespace kinds.

Regression coverage: `TestWrite_VendorCharts_MixedCascadeDeleteKinds` (hookless CRD, author-hooked CRD, Namespace). See `0d86236c`.

rorajani: Thanks and agreed. Dropped the `post-upgrade` + `before-hook-creation` approach.

This PR now takes Option B / structural fix (#1835): vendored mixed components emit a tracked `<component>-post` Helm release (same as non-vendored) instead of injecting `helm.sh/hook` resources. Upgrades three-way-merge in place; uninstall removes the resources; no delete-then-recreate on CRD/Namespace kinds.

Regression coverage: `TestWrite_VendorCharts_MixedCascadeDeleteKinds` (hookless CRD, author-hooked CRD, Namespace). See `0d86236c`.

docs/user/bundling.md
rorajani: Thanks — agreed. Dropped the `post-upgrade` + `before-hook-creation` approach.

This PR now takes Option B / structural fix (#1835): vendored mixed components emit a tracked `<component>-post` Helm release (same as non-vendored) instead of injecting `helm.sh/hook` resources. Upgrades three-way-merge in place; uninstall removes the resources; no delete-then-recreate on CRD/Namespace kinds.

Regression coverage: `TestWrite_VendorCharts_MixedCascadeDeleteKinds` (hookless CRD, author-hooked CRD, Namespace). See `0d86236c`.

docs/user/bundling.md
rorajani: Thanks — agreed. Dropped the `post-upgrade` + `before-hook-creation` approach.

This PR now takes Option B / structural fix (#1835): vendored mixed components emit a tracked `<component>-post` Helm release (same as non-vendored) instead of injecting `helm.sh/hook` resources. Upgrades three-way-merge in place; uninstall removes the resources; no delete-then-recreate on CRD/Namespace kinds.

Regression coverage: `TestWrite_VendorCharts_MixedCascadeDeleteKinds` (hookless CRD, author-hooked CRD, Namespace). See `0d86236c`.

docs/user/bundling.md
rorajani: Docs rewritten for the structural fix: no longer document hook re-apply / delete-recreate as the supported vendored path. Author-declared hooks on recipe manifests are stripped into ordinary `-post` templates (same as non-vendored). Broken `#kueue` anchor removed with that section.

docs/user/bundling.md
rorajani: Docs rewritten for the structural fix: no longer document hook re-apply / delete-recreate as the supported vendored path. Author-declared hooks on recipe manifests are stripped into ordinary `-post` templates (same as non-vendored). Broken `#kueue` anchor removed with that section.

docs/user/bundling.md
rorajani: Docs rewritten for the structural fix: no longer document hook re-apply / delete-recreate as the supported vendored path. Author-declared hooks on recipe manifests are stripped into ordinary `-post` templates (same as non-vendored). Broken `#kueue` anchor removed with that section.

pkg/bundler/deployer/localformat/vendor_wrapper.go
rorajani: Agreed — this is the path taken. Vendored and non-vendored now share the tracked `-post` release shape; hook injection helpers are removed.

pkg/bundler/deployer/localformat/vendor_wrapper_test.go
rorajani: Added `TestWrite_VendorCharts_MixedCascadeDeleteKinds` covering CRD (hookless + author-hooked) and Namespace through the vendored mixed path; asserts no `helm.sh/hook` / `before-hook-creation`.

pkg/bundler/deployer/localformat/vendor_wrapper.go
rorajani: N/A after Option B — `injectPostInstallHooks` / related helpers and comments were deleted rather than renamed or reworded.

pkg/bundler/deployer/localformat/vendor_wrapper.go
rorajani: N/A after Option B — `injectPostInstallHooks` / related helpers and comments were deleted rather than renamed or reworded.

pkg/bundler/deployer/localformat/vendor_wrapper.go
rorajani: N/A after Option B — `injectPostInstallHooks` / related helpers and comments were deleted rather than renamed or reworded.

pkg/bundler/deployer/localformat/vendor_wrapper.go
```diff
@@ -132,10 +135,19 @@ func InjectPostInstallHooks(in []byte) ([]byte, error) {
...
+// raw manifests apply AFTER the vendored subchart's resources on both
+// `helm install` and `helm upgrade`.
+//
+// Why both phases: a post-install-only hook fires on the initial
+// install but Helm skips it on subsequent `helm upgrade` runs, so a
+// version bump on the recipe-side manifest never reaches the cluster.
+// Adding post-upgrade makes upgrades re-apply the manifest. Uninstall
+// residue is a separate Helm limitation — hook resources are not
+// tracked as release members and survive `helm uninstall`; the
+// vendored bundling docs describe the manual cleanup path.
```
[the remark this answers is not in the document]
rorajani: N/A after Option B — `injectPostInstallHooks` / related helpers and comments were deleted rather than renamed or reworded.

docs/user/bundling.md
```diff
@@ -150,17 +150,32 @@ aicr bundle --recipe recipe.yaml --vendor-charts --output ./bundles
...
+releases, so any `helm.sh/hook` annotation a recipe manifest declares is
+stripped when the `-post` chart is written — the same treatment the
+non-vendored path has always applied.
+
+> Earlier releases injected these manifests into the vendored wrapper chart as
+> `helm.sh/hook: post-install` resources, which Helm never re-applied on
+> upgrade, left behind on uninstall, and Argo CD silently skipped as a PostSync
+> hook. Bundles generated before this change may still have those orphaned
+> resources in the cluster; remove them with
+> `kubectl delete -f <bundle>/<NNN>-<component>/templates/` before redeploying.
```
mchmarny: **Blocking — this migration note prescribes a command that destroys cluster data, and understates when operators have to run it.**

**1. The cleanup is mandatory, not housekeeping.** Resources Helm created as hooks are not release members of `<component>-post`. On redeploy, `helm upgrade --install <component>-post ./` aborts with `exists and cannot be imported into the current release: invalid ownership metadata` — the live object carries either no `meta.helm.sh/release-name` or the *primary* release's. "may still have those orphaned resources" reads as optional tidying; it is a hard precondition, and the note never says the redeploy fails without it.

**2. For CRD-bearing components the prescribed command is the cascade-delete this PR exists to prevent.** [`recipes/mixins/platform-inference.yaml#L21-L27`](https://github.com/NVIDIA/aicr/blob/c412628ec16b9374bda4d1507d6f0c729f95a800/recipes/mixins/platform-inference.yaml#L21-L27) makes `agentgateway-crds` a Helm component (`source: oci://cr.agentgateway.dev/charts`) whose `manifestFiles` are five `CustomResourceDefinition`s. `kubectl delete -f <bundle>/<NNN>-agentgateway-crds/templates/gateway-api-crds.yaml` on a live cluster garbage-collects every Gateway / HTTPRoute / GRPCRoute / InferencePool cluster-wide.

Verified reachable at this head: built `aicr` and ran `bundle --vendor-charts` for `--service eks --accelerator h100 --os ubuntu --intent inference`, which emits exactly `001-agentgateway-crds/` (wrapper + tarball, no `templates/`) and `002-agentgateway-crds-post/templates/{gateway-api-crds,inference-extension-crds}.yaml`.

**Minimum fix:** lead with the non-destructive Helm adoption path, say plainly that redeploy fails until the resources are adopted or removed, and reserve `kubectl delete` for kinds where cascade is acceptable with an explicit CRD/Namespace warning.

```bash
kubectl label    <kind>/<name> app.kubernetes.io/managed-by=Helm --overwrite
kubectl annotate <kind>/<name> meta.helm.sh/release-name=<component>-post \
                               meta.helm.sh/release-namespace=<ns> --overwrite
```

tests/chainsaw/cli/bundle-vendor-charts/chainsaw-test.yaml
```diff
@@ -154,6 +154,18 @@ spec:
...
+                # #1835: recipe-side manifests of mixed components ship as a
+                # tracked <name>-post release, never as helm.sh/hook resources
+                # inside the vendored wrapper. A hook annotation anywhere in a
+                # vendored bundle means the collapsed-wrapper shape regressed:
+                # helm upgrade would skip the resource, helm uninstall would
+                # orphan it, and Argo CD would treat it as a PostSync hook that
+                # never fires under syncPolicy.automated.
+                if [ -d "${dir}templates" ] && \
+                   grep -rq 'helm.sh/hook' "${dir}templates" 2>/dev/null; then
+                  echo "${name}: templates/ must not carry helm.sh/hook annotations" >&2
```
mchmarny, line 168: Non-blocking — **this guard is vacuous for the recipe the test bundles, so it cannot detect the regression it describes.**

The fixture recipe at [L59-L61](https://github.com/NVIDIA/aicr/blob/c412628ec16b9374bda4d1507d6f0c729f95a800/tests/chainsaw/cli/bundle-vendor-charts/chainsaw-test.yaml#L59-L61) (`--service eks --accelerator gb200 --os ubuntu --intent training`) has no component that is both vendorable and mixed. Its only `manifestFiles` carrier is `nodewright-customizations`, whose `source: ""` makes `shouldVendor` return false, so it never routes through the vendored path.

Ran it at this head: the bundle emits no `-post` folder at all, and exactly two `templates/` dirs — `006-nodewright-customizations/` (manifest-only local chart) and `009-gpu-operator-pre/` (pre-injection). Both already flow through `stripHelmHooks` on `main`, so this grep passes identically before and after the fix. Under the pre-#1835 collapsed shape this same recipe also produced zero hook-annotated templates, because `writeMixedManifests` was only reachable when `shouldVendor(c) && len(manifests) > 0`.

The Go tests (`TestWrite_VendorCharts_Mixed`, `TestWrite_VendorCharts_MixedCascadeDeleteKinds`) do cover the behavior, hence non-blocking — but as written this assertion contributes no e2e signal.

**Fix:** bundle an inference recipe in the same step (`--intent inference` on eks pulls in `agentgateway-crds`, a real vendored mixed component), and additionally assert that a `NNN-<name>-post/templates/` folder *exists*. Absence-only assertions still pass if the `-post` injection silently stops emitting the manifests altogether.

mchmarny (changes requested):
**REQUEST_CHANGES** — 1 blocker, 2 non-blocking.

Re-reviewed at head `c412628e` (previous round was `eb1bb22`, docs-only). Option B is the right structural fix and my earlier `CHANGES_REQUESTED` on `vendor_wrapper.go` is resolved by removal — every finding from that round is moot.

Validated by building `aicr` at head and generating a real `--vendor-charts` bundle for `--service eks --accelerator h100 --os ubuntu --intent inference`: `001-agentgateway-crds/` (wrapper + tarball, no `templates/`) + `002-agentgateway-crds-post/templates/*`, and zero `helm.sh/hook` anywhere in the emitted bundle. `go build ./...` and `go test ./pkg/bundler/deployer/localformat/... ./pkg/bundler/deployer/helmfile/...` pass. Argo CD (`waveForFolder`, name-suffix keyed) and helmfile (`CarriesPostManifests`) both route the new folder correctly, and flux already had this shape.

CI is green and — unlike the previous round — the primary/`-post` split is genuinely exercised by unit tests. The one *e2e* assertion this PR adds is not; see the inline comment on `chainsaw-test.yaml`.

The blocker is the migration note in `bundling.md`, inline.

---

**Non-blocking — four sites still document the removed "collapsed vendored folder" as current behavior.** One root cause; the PR updated `folder.go`, `doc.go`, `vendor_folder.go`, `vendor_wrapper.go`, `bundling.md` and `cli-reference.md` but missed these. No changed-line anchor, so listing here:

- [`docs/contributor/recipe.md#L104`](https://github.com/NVIDIA/aicr/blob/c412628ec16b9374bda4d1507d6f0c729f95a800/docs/contributor/recipe.md#L104) — published page: "the injected `-post` wrapper, or the collapsed vendored folder under `--vendor-charts`"
- [`pkg/bundler/deployer/localformat/writer.go#L101-L106`](https://github.com/NVIDIA/aicr/blob/c412628ec16b9374bda4d1507d6f0c729f95a800/pkg/bundler/deployer/localformat/writer.go#L101-L106) — `Options.VendorCharts` godoc, in a file this PR edits: "(for mixed components) post-install-hook templates. Mixed components no longer split into primary + -post."
- [`pkg/bundler/deployer/helmfile/releases.go#L261-L264`](https://github.com/NVIDIA/aicr/blob/c412628ec16b9374bda4d1507d6f0c729f95a800/pkg/bundler/deployer/helmfile/releases.go#L261-L264)
- [`pkg/recipe/components.go#L139-L141`](https://github.com/NVIDIA/aicr/blob/c412628ec16b9374bda4d1507d6f0c729f95a800/pkg/recipe/components.go#L139-L141)

---

**Checked and deliberately not filed:** CodeRabbit's "reject a post-phase `Namespace` matching `Component.Namespace`" — no shipped recipe puts a `Namespace` in `manifestFiles`; all five `talos-namespace.yaml` files attach via `preManifestFiles` in `recipes/mixins/os-talos.yaml`. The scenario is equally present on the non-vendored path, so it is pre-existing and unreachable, not something this PR introduces. Removing exported `InjectPostInstallHooks` is outside `make api-diff` scope (`pkg/client/v1` closure) and has no remaining callers. The `folderCount` pre-pass and the emission conditions still agree for every vendored / kustomize / manifest-only shape.

docs/user/bundling.md
rorajani: Addressed in `eb9d4fb1`.

Rewrote the migration note to:
1. State that redeploy fails with Helm ownership conflicts until the live hook-era objects are adopted or removed (hard precondition, not optional cleanup).
2. Lead with the non-destructive Helm adoption path (`kubectl label` / `kubectl annotate` onto `<component>-post`).
3. Reserve `kubectl delete` for kinds where cascade is acceptable, with an explicit **do not** delete CRD/Namespace warning — `agentgateway-crds` Gateway API / Inference Extension CRDs called out as the concrete risk.

Also moved the adoption commands out of the blockquote so the MDX safety gate accepts the `<kind>` / `<name>` placeholders.

tests/chainsaw/cli/bundle-vendor-charts/chainsaw-test.yaml
rorajani: Addressed in `eb9d4fb1`.

Switched the fixture to `--service eks --accelerator h100 --os ubuntu --intent inference` so the bundle includes vendored mixed `agentgateway-crds`, and added presence checks for:
- `NNN-agentgateway-crds-post/templates/` with at least one manifest
- vendored primary `NNN-agentgateway-crds/` has `charts/*.tgz` and **no** `templates/`

Validated in CI: `cli-bundle-vendor-charts` passed, and KWOK Tier 1 (including `eks-inference` variants) is green after re-running the GoReleaser download flake.

docs/user/bundling.md
```diff
@@ -150,17 +150,53 @@ aicr bundle --recipe recipe.yaml --vendor-charts --output ./bundles
...
+> each resource is adopted or removed.
+
+Adoption is the default path — it is non-destructive and required before
+`helm upgrade --install` of the `-post` chart. For each previously
+hook-injected resource (name and kind from the new
+`<NNN>-<component>-post/templates/` files, or from the old primary
+`templates/` if you still have that bundle):
+
+```bash
+kubectl label <kind>/<name> app.kubernetes.io/managed-by=Helm --overwrite
```
mchmarny, line 188: Non-blocking — the adoption commands need `-n <ns>` for namespaced resources.

`kubectl label <kind>/<name> …` resolves against the operator's current context namespace, so for the namespaced half of the affected manifests it errors `NotFound` and the migration stalls partway through. Not hypothetical for the default inference bundle: `007-agentgateway-post/` ships a `Gateway` and an `AgentgatewayParameters` into `agentgateway-system` (its generated `install.sh` runs `--namespace agentgateway-system --create-namespace`). Across `recipes/components/*/manifests/` roughly half the kinds are namespaced (ConfigMap, DaemonSet, Subscription, OperatorGroup, RoleBinding, PersistentVolumeClaim, LocalQueue, Gateway); the rest are cluster-scoped (CRD, ClusterQueue, ResourceFlavor, ClusterTrainingRuntime, NicClusterPolicy).

Minimum fix: add `-n <ns>` to both commands — the same `<ns>` already passed to `meta.helm.sh/release-namespace` — with a half-line noting it is omitted for cluster-scoped kinds.

mchmarny (approved):
**APPROVE** — 0 blockers.

Re-reviewed at head `eb9d4fb1` (previous round was `c412628e`). Both findings from that round are resolved:

- **Blocker — the migration note prescribed a destructive `kubectl delete -f`.** Fixed. `docs/user/bundling.md` now states the redeploy *fails* with Helm ownership conflicts until each hook-era object is adopted or removed (hard precondition, not optional cleanup), leads with the non-destructive `kubectl label`/`annotate` adoption path, and carries an explicit **do not** delete for CRD/Namespace kinds naming `agentgateway-crds` as the concrete risk.
- **Non-blocking — the chainsaw guard was vacuous for its fixture recipe.** Fixed, and verified non-vacuous below.
- **Non-blocking — four sites documented the removed "collapsed vendored folder".** All four updated (`docs/contributor/recipe.md`, the `Options.VendorCharts` godoc, `helmfile/releases.go`, `pkg/recipe/components.go`); a repo-wide grep finds no remaining reference.

Validated at this head by building `aicr` and generating a real `--vendor-charts` bundle for the test's own recipe (`--service eks --accelerator h100 --os ubuntu --intent inference`) against its fake `helm`:

- Emits `001-agentgateway-crds/` + `002-agentgateway-crds-post/` and `006-agentgateway/` + `007-agentgateway-post/`. Primaries hold `charts/*.tgz` and no `templates/`; both `-post` folders hold the recipe manifests; `helm.sh/hook` appears nowhere in the bundle; all four folders are covered by `checksums.txt`.
- Counterfactual on `main`: the same recipe collapses to `001-agentgateway-crds/templates/*` and `005-agentgateway/templates/*` **with** `helm.sh/hook`, and no `-post` folder exists — so all three new chainsaw assertions fail on `main`. The e2e guard now detects the regression it describes.
- `go test -race ./pkg/bundler/... ./pkg/recipe/...` and `golangci-lint run -c .golangci.yaml` on the changed packages are clean.
- Flux already emitted a separate `-post` chart for mixed components regardless of vendoring (`flux.go` L511-537), so the two paths are now consistent rather than divergent, and `InjectPostInstallHooks` has no remaining callers anywhere in the tree.

CI is green at this exact SHA, including `tests / CLI E2E` (which runs the updated `cli-bundle-vendor-charts`) and the KWOK `eks-inference` Tier 1 lanes.

Branch is `BEHIND` `main` — mechanical only; update before merge.

njhensley (approved):

docs/user/bundling.md
rorajani: Follow-up: https://github.com/NVIDIA/aicr/pull/2190 — adds `-n <ns>` to the adoption examples (same ns as `meta.helm.sh/release-namespace`) and notes to omit `-n` for cluster-scoped kinds.

PR #2063, opened 2026-08-05
fix(releasepolicy): derive per-script deadline from t.Deadline()

### LINKED ISSUE #1974: tests/releasepolicy: 10s per-script deadline flakes under full-suite load on dev machines

framsouza:
## Summary

Replace the hardcoded 10s per-script wall-clock deadline in `tests/releasepolicy` with a `t.Deadline()`-derived budget (90s cap, 30s margin off the outer test-binary deadline) so shelled-out release-script tests survive CPU starvation under full-suite `make qualify` load without losing hang detection.

## Motivation / Context

Under full-suite `make qualify` runs (`-race`, coverage, ~96 packages concurrent) the reporter observed a ~50% flake rate in `tests/releasepolicy`, always the `script exceeded test deadline: context deadline exceeded` sentinel, always a different subtest each time. In-isolation timings: p50=0.49s, p90=4.52s, max=5.54s per script; under contention wall times hit 12–30s against the 10s constant. Never observed on CI.

A tight constant conflates *hangs* (minutes) with *slow-but-progressing* runs (single-digit seconds under contention) — differ by orders of magnitude, so the 10s bound catches no additional hangs while producing false positives that erode trust in the documented pre-push gate.

Fixes: #1974
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tests/releasepolicy` (release-script test helpers)

## Implementation Notes

- New helper `scriptDeadline(t)` in `release_scripts_test.go`:
  - Returns `scriptDeadlineCap` (90s) when the outer test binary has no deadline (`-timeout 0`) or plenty of headroom.
  - Otherwise shrinks to `t.Deadline() - scriptDeadlineMargin` (30s) so `t.Fatalf` can report cleanly before Go reaps the run.
  - Positive-remaining guard prevents a degenerate near-expired deadline from returning a zero/negative budget.
- Applied to both call sites in the package: `runScript` (`release_scripts_test.go:1682`) and `runOpenVEXGuard` (`release_workflow_test.go:592`) — same hardcoded 10s pattern, same package, same starvation vulnerability class.
- Perf assertions using `time.Since(started) > 4*time.Second` at lines 1346/1388 are **intentionally left alone**: they assert that the script's own `AICR_NETWORK_TIMEOUT_SECONDS=1` bound fires — a real regression signal, not a harness deadline.

**Alternatives considered:**
- Raise the constant to 60–90s: works, leaves a constant to re-tune per-environment.
- Serialize `tests/releasepolicy` into its own `-p` slot in `make test`: addresses root-cause contention but requires Makefile restructure, slows total wall time, and doesn't protect future subprocess-heavy tests in the same package. Adaptive deadline is a superset.

## Testing

```bash
# Package in isolation with the gate's exact flags — zero deadline sentinels
unset GITLAB_TOKEN && AICR_CRITERIA_STRICT=1 go test -short -count=1 -race ./tests/releasepolicy/...

# Mandatory Go lint gate on affected package
GOFLAGS="-mod=vendor" golangci-lint run -c .golangci.yaml ./tests/releasepolicy/...
```

Results on the development machine:
- `go build ./tests/releasepolicy/...` — clean
- `go vet ./tests/releasepolicy/...` — clean
- `golangci-lint run` on affected package — **0 issues**
- `go test -race -short ./tests/releasepolicy/...` — **zero `script exceeded test deadline` occurrences** (the sentinel this fix targets). Some subtests fail on this specific Mac because `timeout` (GNU coreutils) is not installed and system bash is 3.2 — pre-existing environment issues, orthogonal to this change, would fail identically on `origin/main`.
- Full `make test`: only failing package was `validators/performance` (`env: bash: No such file or directory` shebang issue) — also unrelated to this change.

## Risk Assessment

**Rollout notes:** N/A. Test-file-only change. Reverting is a single-file diff.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — affected package's target sentinel no longer fires; unrelated pre-existing env failures noted above
- [x] Linter passes (`make lint`) — `golangci-lint run` on affected package is clean; `yamllint` N/A (no YAML changes)
- [ ] I added/updated tests for new functionality — N/A, this is a test-harness fix; the subject-under-test is the test-suite behavior itself
- [ ] I updated docs if user-facing behavior changed — N/A, no user-visible behavior change

### DIFF
--- a/tests/releasepolicy/release_scripts_test.go
+++ b/tests/releasepolicy/release_scripts_test.go
@@ -165,7 +165,7 @@ func TestReleaseSbomSignaturesAreAllowlisted(t *testing.T) {
 // allowlist rather than a copy of it.
 func shellReleaseAssetNames(t *testing.T, tag string) []string {
 	t.Helper()
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), scriptDeadline(t))
 	defer cancel()
 	text := string(readFile(t, ".github/scripts/release-images.sh"))
 	const marker = "expected_release_asset_names() {"
@@ -1679,7 +1679,7 @@ type scriptResult struct {
 
 func runScript(t *testing.T, environment []string, relative string, args ...string) scriptResult {
 	t.Helper()
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), scriptDeadline(t))
 	defer cancel()
 	command := exec.CommandContext(ctx, repositoryPath(t, relative), args...)
 	command.Env = environment
@@ -1689,6 +1689,31 @@ func runScript(t *testing.T, environment []string, relative string, args ...stri
 	}
 	return scriptResult{output: string(output), err: err}
 }
+
+// scriptDeadline returns the per-script wall-clock budget for shelled-out
+// helpers in this package. The value guards against genuine hangs (infinite
+// loops, stuck prompts) while tolerating CPU starvation when the full test
+// suite runs concurrently with -race + coverage.
+//
+// Hangs and slow-but-progressing runs differ by orders of magnitude, so a
+// tight constant catches no additional hangs while producing false positives
+// under contention (see issue #1974). We cap at scriptDeadlineCap and shrink
+// only if the outer test binary's own deadline is closer, leaving
+// scriptDeadlineMargin so t.Fatalf can report cleanly before Go reaps the run.
+func scriptDeadline(t *testing.T) time.Duration {
+	t.Helper()
+	if d, ok := t.Deadline(); ok {
+		if remaining := time.Until(d) - scriptDeadlineMargin; remaining > 0 && remaining < scriptDeadlineCap {
+			return remaining
+		}
+	}
+	return scriptDeadlineCap
+}
+
+const (
+	scriptDeadlineCap    = 90 * time.Second
+	scriptDeadlineMargin = 30 * time.Second
+)
 
 func sortedImages() []string {
 	images := make([]string, 0, len(releaseImages))
--- a/tests/releasepolicy/release_workflow_test.go
+++ b/tests/releasepolicy/release_workflow_test.go
@@ -24,7 +24,6 @@ import (
 	"sort"
 	"strings"
 	"testing"
-	"time"
 
 	"gopkg.in/yaml.v3"
 )
@@ -589,7 +588,7 @@ func TestReleaseOpenVEXValidation(t *testing.T) {
 // annotations and the GITHUB_OUTPUT contents) alongside the exit status.
 func runOpenVEXGuard(t *testing.T, script, workspace string) (string, error) {
 	t.Helper()
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), scriptDeadline(t))
 	defer cancel()
 	outputs := filepath.Join(t.TempDir(), "outputs")
 	command := exec.CommandContext(ctx, "bash", "-c", script)


tests/releasepolicy/release_scripts_test.go
```diff
@@ -1690,6 +1690,31 @@ func runScript(t *testing.T, environment []string, relative string, args ...stri
...
+//
+// Hangs and slow-but-progressing runs differ by orders of magnitude, so a
+// tight constant catches no additional hangs while producing false positives
+// under contention (see issue #1974). We cap at scriptDeadlineCap and shrink
+// only if the outer test binary's own deadline is closer, leaving
+// scriptDeadlineMargin so t.Fatalf can report cleanly before Go reaps the run.
+func scriptDeadline(t *testing.T) time.Duration {
+	t.Helper()
+	if d, ok := t.Deadline(); ok {
+		if remaining := time.Until(d) - scriptDeadlineMargin; remaining > 0 && remaining < scriptDeadlineCap {
```
mchmarny, line 1706: Non-blocking: when this guard fails because `remaining <= 0`, control falls through to `return scriptDeadlineCap` — so within the last 30s of the test binary's budget a script is handed **90s**, longer than the time actually left. That inverts the doc comment's own stated intent two lines up ("shrink only if the outer test binary's own deadline is closer, leaving `scriptDeadlineMargin` so `t.Fatalf` can report cleanly before Go reaps the run").

Consequence is diagnostics-only and the window is narrow (10m per-binary budget from `.settings.yaml`; the package runs ~113–148s locally), which is why this isn't blocking: instead of the clean `script exceeded test deadline` sentinel you get a whole-binary `panic: test timed out` dump — in a run that was already going to be reaped.

Minimum correction is to clamp rather than fall through, e.g. subtract the margin only when it leaves a positive budget and otherwise use whatever remains:

```go
if d, ok := t.Deadline(); ok {
    remaining := time.Until(d)
    if remaining <= 0 {
        return time.Nanosecond // already expired; let the context fire immediately
    }
    if remaining > scriptDeadlineMargin {
        remaining -= scriptDeadlineMargin
    }
    if remaining < scriptDeadlineCap {
        return remaining
    }
}
```

Note @coderabbitai flagged this same spot but proposed returning **zero**, which would be worse: `context.WithTimeout(ctx, 0)` is already expired, so every remaining script would be killed instantly and fail. Clamping to the remainder keeps the fail direction without manufacturing failures.

mchmarny (commented):
**COMMENT** — 0 merge blockers on `cde40ec`, 2 non-blocking findings.

**Validated:** detached worktree at head. `go vet` clean; `golangci-lint run -c .golangci.yaml ./tests/releasepolicy/...` 0 issues; `go test -short -count=1 -race -timeout=10m ./tests/releasepolicy/...` passes at head in 147.6s vs 112.9s on `main` — so the 9× larger cap does not translate into a materially longer package run, which was my main concern about trading hang-detection latency for tolerance. CI is 21 pass / 11 skipping, 0 failing. `BEHIND main`, so a rebase is needed before merge.

The premise holds: hangs and contention-slowed runs differ by orders of magnitude, so deriving the budget from `t.Deadline()` is the right shape. One correctness note on the fallthrough is inline.

**Second finding (not anchorable — the lines are unchanged):** the two `elapsed > 4*time.Second` assertions at [L1346](https://github.com/NVIDIA/aicr/blob/cde40ecc5e11b88b0f6d80adcbeb0992bd5e4576/tests/releasepolicy/release_scripts_test.go#L1346) and [L1388](https://github.com/NVIDIA/aicr/blob/cde40ecc5e11b88b0f6d80adcbeb0992bd5e4576/tests/releasepolicy/release_scripts_test.go#L1388) are in the same flake class this PR is closing. The stated rationale for leaving them — they assert the script's own `AICR_NETWORK_TIMEOUT_SECONDS=1` bound, a real regression signal — is right about the *signal*, but the bound is measured in absolute wall clock via `time.Since(started)`. By this PR's own contention data (p50 0.49s scripts stretching to 12–30s), a script whose internal budget is ~1s will exceed a 4s wall-clock ceiling under the same starvation, so #1974 can reopen from those two `t.Errorf`s. Worth scaling them off the same helper, or asserting relative to a measured baseline.

**Two smaller notes:**
- The description says the helper was applied to "both call sites," but three were converted — `shellReleaseAssetNames` at L168 as well. Worth correcting so the diff and the description agree.
- This collides with #2060, which adds `const scriptDeadline = 30 * time.Second` to this same file for the same symptom (under #2018 rather than #1974). Same package, same identifier, different shapes — whichever lands second conflicts, and a clean auto-merge would be a duplicate declaration. The two branches are also on different bases of this file (`runScript` at 1424 there vs 1682 here). Worth settling on one approach across the two PRs; this adaptive one strictly supersedes a fixed 30s constant.

framsouza:
Superseded by #2088, which was merged overnight with the same `t.Deadline()`-derived approach (authored by the issue reporter). Closing to avoid churn.

Issue #2065, opened 2026-08-05
GKE actuator node-pool create timeout (30m) too short for slow GPU pools → orphaned clusters


njhensley:
### Summary

The GKE UAT actuator (`ghcr.io/mchmarny/cluster/gke`) hardcodes a **30-minute create timeout** on `google_container_node_pool` (`terraform/compute.tf`, `timeouts { create = "30m" }`), which is too short for slow GPU pool provisioning. Terraform then fails the whole `apply` even though GKE finishes and reports the pool `RUNNING` shortly after.

### Evidence

On UAT run [30965268896](https://github.com/NVIDIA/aicr/actions/runs/30965268896) the `cpu-worker` pool took **~35 min** and briefly reported `ERROR`; GKE recovered it, but Terraform had already timed out and failed the run. Because that was a manual/held run, teardown was skipped and the healthy cluster + its VPC network group (1 cluster VPC + 8 `gpu-nic` VPCs + Cloud Router + NAT + firewall + subnets) leaked — contributing to the global `NETWORKS` quota (50) exhaustion that blocked new runs.

### Ask (upstream)

Bump the node-pool `create`/`update` timeout `30m → 60m` in the GKE actuator, mirroring what `mchmarny/cluster#31` did for the AKS actuator's worker-pool timeouts, then bump the pinned `GKE_ACTUATOR_IMAGE` in `.github/workflows/uat-gcp.yaml`.

### In-repo mitigation (separate PR)

A bounded **Bringup retry** in `uat-gcp.yaml` — a second (idempotent) `terraform apply` refreshes state, finds the self-healed pool `RUNNING`, and converges — reduces the failure rate but does not remove the need for the upstream timeout bump. Tracked alongside the janitor/teardown-reliability work.

### Type
Bug — causes orphaned clusters/networks and quota exhaustion.

mchmarny:
Upstream ask is done — released as **`ghcr.io/mchmarny/cluster/gke:v0.5.17`**.

**1. Node-pool timeouts (`terraform/compute.tf`)** — `create`/`update` `30m → 60m`, mirroring what mchmarny/cluster#31 did for AKS. `delete` stays at `30m`; pool deletion is not capacity-bound.

[...]

PR #2066, opened 2026-08-05
fix(ci): retry GKE UAT bringup on transient node-pool timeout

njhensley:
## Summary

Wrap the GCP UAT **Bringup Infra** step in a bounded 2-attempt retry so a transient node-pool provisioning timeout no longer fails the run (and strands a healthy cluster).

## Motivation / Context

The GKE actuator (`ghcr.io/mchmarny/cluster/gke`) hardcodes a **30-minute `create` timeout** on `google_container_node_pool` (`terraform/compute.tf`). A slow GPU pool can take **~35 min**: Terraform times out and fails the `apply` even though GKE finishes and reports the pool `RUNNING` moments later. On run [30965268896](https://github.com/NVIDIA/aicr/actions/runs/30965268896) the `cpu-worker` pool did exactly this — and because it was a held/manual run, teardown was skipped and the healthy cluster + its full VPC network group (1 cluster VPC + 8 `gpu-nic` VPCs + router + NAT + firewall + subnets) leaked, contributing to the `NETWORKS` quota (50) exhaustion that blocked later runs.

`terraform apply` is idempotent, so a second attempt refreshes state, waits out the still-completing (or already-healthy) pool, and converges. Bounded to 2 attempts so a genuinely-stuck provision still fails fast and surfaces (and, on nightly, teardown still runs).

Fixes: N/A (mitigates)
Related: #2065 (upstream node-pool timeout bump 30m→60m — the durable fix)

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT GCP workflow (`.github/workflows/uat-gcp.yaml`)

## Implementation Notes

- Mirrors the existing `Destroy Cluster` retry pattern (`docker run` inside an `if`, which does not trip `set -e`, plus an explicit post-loop guard so a real failure still `exit 1`s).
- Scoped to GCP, where the 30m timeout and the evidence are. AKS already carries 60m worker-pool timeouts (mchmarny/cluster#31); EKS parity can follow if the same symptom appears there.
- This reduces the orphan *rate* for one mode; the general backstop (a scheduled janitor that reaps orphaned `aicr-<run_id>` groups with no live workflow run) is tracked separately.

## Testing

```bash
yamllint .github/workflows/uat-gcp.yaml    # clean
actionlint .github/workflows/uat-gcp.yaml  # no new findings (3 pre-existing info-level SC2016 in an unrelated step)
```

## Risk Assessment

- [x] **Low** — Adds a bounded retry around an idempotent apply; worst case is one extra apply attempt (~within the 280m job budget). Easy to revert.

**Rollout notes:** N/A — CI-only.

## Checklist

- [x] Linter passes (`make lint` — yamllint/actionlint clean on the edited step)

### DIFF
--- a/.github/workflows/uat-gcp.yaml
+++ b/.github/workflows/uat-gcp.yaml
@@ -454,11 +454,43 @@ jobs:
         if: inputs.lifecycle != 'daytime-down'
         run: |
           set -euo pipefail
-          /usr/bin/docker run \
-            -e CONFIG_CONTENT="$(base64 < "$CLUSTER_CONFIG")" \
-            -e AUTO_APPROVE=true \
-            -e KEY_CONTENT="$(base64 < ${{ steps.auth.outputs.credentials_file_path }})" \
-            ${{ env.GKE_ACTUATOR_IMAGE }} apply
+          # Retry the actuator apply once on a transient provisioning failure.
+          # The GKE actuator hardcodes a 30m node-pool create timeout
+          # (terraform/compute.tf), but a slow GPU pool can take ~35m; Terraform
+          # then fails the create even though GKE finishes and reports the pool
+          # RUNNING moments later (observed on run 30965268896, #2065). On a
+          # nightly run that false failure is recoverable (Destroy still runs),
+          # but on daytime-up / skip_delete runs it strands a healthy cluster and
+          # its whole VPC network group, leaking the GPU node and NETWORKS quota.
+          # `terraform apply` is idempotent: a second attempt refreshes state,
+          # waits out the still-completing (or already-healthy) pool, and
+          # converges. Bounded to 2 attempts so a genuinely-stuck provision still
+          # fails fast and surfaces. The durable fix is an upstream node-pool
+          # timeout bump 30m->60m (mirroring the AKS actuator's
+          # mchmarny/cluster#31) — tracked in #2065.
+          brought_up=false
+          for attempt in 1 2; do
+            echo "Bringup attempt ${attempt}..."
+            if /usr/bin/docker run \
+              -e CONFIG_CONTENT="$(base64 < "$CLUSTER_CONFIG")" \
+              -e AUTO_APPROVE=true \
+              -e KEY_CONTENT="$(base64 < ${{ steps.auth.outputs.credentials_file_path }})" \
+              ${{ env.GKE_ACTUATOR_IMAGE }} apply; then
+              brought_up=true
+              break
+            fi
+            if [ "${attempt}" -lt 2 ]; then
+              echo "Bringup attempt ${attempt} failed; retrying in 60s (a transient node-pool state often self-heals, and apply is idempotent)..."
+              sleep 60
+            fi
+          done
+          # Fail the step if every attempt failed: `docker run` in an `if` does
+          # not trip `set -e`, so without this guard a real bringup failure would
+          # exit 0 and the run would proceed against a cluster that never came up.
+          if [ "$brought_up" != true ]; then
+            echo "::error::Bringup failed after 2 attempts"
+            exit 1
+          fi
 
       # Connect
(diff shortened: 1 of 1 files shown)


mchmarny (approved):

yuanchen8911:
Two follow-up defects on main from this PR.

The budget comment is stale and the timeout cap is insufficient against the now-pinned v0.5.17 actuator. v0.5.17 widens the Terraform node-pool create/update timeout from 30m to 60m, but the retry comment still describes the old 30m cap and refers to a "pending upstream timeout bump" that has already landed. On the fast-recovery path (first apply observed at ~49m, second apply quick), the retry pushes the total to approximately 285m — already past the unchanged 280m cap. If the second apply approaches the new 60m timeout, the total reaches ~353m, well past the proposed 315m. The budget comment and `timeout-minutes` need to be recalculated against v0.5.17's actual timeouts, and the stale 30m/pending-fix language removed. The robust fix is independently budgeted teardown; if the single-job design stays, the budget must be recomputed.

The retry also has a false-green path on a retained ERROR node pool. Terraform provider Google v7.35 records the pool ID before creation and retains it when the pool reaches ERROR state; a subsequent apply produces a no-change plan and exits 0 while the pool is broken. `brought_up=true` is then set and `kubectl get nodes` passes (cluster API reachable, healthy nodes visible from other pools), so the run proceeds or holds without detecting the broken pool. On a `skip_tests=true` or `daytime-up` run this holds an unusable cluster indefinitely — the `failure()` teardown added by #2067 does not fire on a zero exit. Fix: verify every configured GKE node pool reports RUNNING after the actuator succeeds before setting `brought_up=true`.

yuanchen8911:
[uat-gcp.yaml:457–470](https://github.com/NVIDIA/aicr/blob/00cb0519e9f59b6c8874e06f526822436ed000c9/.github/workflows/uat-gcp.yaml#L457-L470) (retry comment) · [uat-gcp.yaml:72–82](https://github.com/NVIDIA/aicr/blob/00cb0519e9f59b6c8874e06f526822436ed000c9/.github/workflows/uat-gcp.yaml#L72-L82) (budget comment) · [uat-gcp.yaml:472–491](https://github.com/NVIDIA/aicr/blob/00cb0519e9f59b6c8874e06f526822436ed000c9/.github/workflows/uat-gcp.yaml#L472-L491) (retry loop)

**Stale retry comment (lines 457–470).** The comment describes the old v0.5.16 actuator behavior: "30m node-pool create timeout" and "durable fix is an upstream node-pool timeout bump 30m→60m ... tracked in #2065." Both are stale. Main now pins v0.5.17, which already widens the Terraform node-pool create/update timeout to 60m. The "pending upstream fix" has landed.

**Stale budget (lines 72–82).** The budget comment sizes the job for a single ~30m bringup attempt. With v0.5.17's 60m timeouts, the fast-recovery retry path reaches ~285m (first apply observed at ~49m, plus sleep, plus quick second apply, plus full UAT), exceeding the 280m cap. A worst-case second apply approaching 60m pushes the total to ~353m. The teardown MUST fit in this budget — a job-level timeout cancels pending `always()` steps — so the budget needs to be recalculated or teardown moved to an independent workflow.

**False-green on retained ERROR pool (lines 472–491).** `brought_up=true` is set on any exit 0 from the actuator. The Connect step runs `kubectl get nodes`, which verifies only that the cluster API is reachable — not that every node pool is healthy. Terraform provider Google v7.35 retains a node pool in ERROR state and produces a no-change plan (exit 0) on a subsequent apply. On a `skip_tests=true` or `daytime-up` run, the step sets `brought_up=true` and the cluster is held with a broken pool. `failure()` is false, so #2067's teardown gate does not fire. Fix: after the actuator exits 0, verify every configured GKE node pool reports RUNNING before setting `brought_up=true`.

PR #2067, opened 2026-08-05
fix(ci): tear down daytime-up UAT cluster on failed/cancelled run

njhensley:
## Summary

Make `daytime-up` UAT runs tear down their cluster when the run **failed or was cancelled**, instead of unconditionally holding it. A clean `daytime-up` still holds for handoff.

## Motivation / Context

`daytime-up` (provision + deploy + HOLD) deliberately skips teardown so a human can use the cluster during the day. But that skip is **unconditional**: a `daytime-up` run that fails at Bringup/deploy — or is cancelled — leaves a broken, unusable cluster held indefinitely, leaking its GPU node and full VPC network group (1 cluster VPC + 8 `gpu-nic` VPCs + Cloud Router + NAT + firewall + subnets). This is one of the sources of the `NETWORKS` quota (50) exhaustion that recently blocked all GCP UAT runs.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT workflows (`.github/workflows/uat-{gcp,azure,aws}.yaml`)

## Implementation Notes

- Adds a `daytime-up` branch — `(inputs.lifecycle == 'daytime-up' && (failure() || cancelled()) && steps.infra.outcome != 'skipped')` — to every teardown-gating step: **Re-authenticate for teardown**, **Destroy Cluster**, and (AWS) **Delete LoadBalancer Services (pre-teardown)**, across all three clouds.
- **Holds on success:** on a clean `daytime-up` all three branches are false, so teardown is skipped and the cluster is held — unchanged behavior.
- **No-ops a pre-provision cancel:** `steps.infra.outcome != 'skipped'` means a cancel before Bringup (nothing created) does nothing.
- **Targets the right cluster:** each Destroy step sets `.deployment.id` from this run's own `DEPLOYMENT_ID` (`aicr-uat-day-<reservation>-<run_id>`); the daytime-*down* runtime name-resolution is gated on `daytime-down` and does not run here.
- **Composes with #2049:** a `daytime-up` cancelled mid-Bringup now triggers Destroy, which relies on that PR's orphaned-lock break to actually succeed.

## Testing

```bash
yamllint  .github/workflows/uat-{gcp,azure,aws}.yaml   # clean
actionlint .github/workflows/uat-{gcp,azure,aws}.yaml  # no expression errors; only pre-existing info-level SC2016 in unrelated steps
```

## Risk Assessment

[...]

Issue #2064, opened 2026-08-05
UAT GCP Bringup fails: #2042 names overflow SA account_id cap


yuanchen8911:
## Summary

Every GCP UAT provisioning run is broken on current main. The GKE actuator derives GCP service-account ids as `<deployment.id>-system-nodes` / `<deployment.id>-worker-nodes` (`iam.tf` lines 6/47 in `ghcr.io/mchmarny/cluster/gke:v0.5.16`), and GCP caps `account_id` at 30 characters. The UAT cluster names introduced by #2042 push both derived ids past that cap, so `terraform plan` fails validation before anything is provisioned.

First observed on run [31021150393](https://github.com/NVIDIA/aicr/actions/runs/31021150393) (`uat-run.yaml`, reservation `gcp-h100`, lifecycle `daytime-up`, main @ 01932ea85):

```
Error: "account_id" ("aicr-uat-day-gcp-h100-31021150393-system-nodes") must be between 6 and 30 characters long
  with google_service_account.system_nodes, on iam.tf line 6
Error: "account_id" ("aicr-uat-day-gcp-h100-31021150393-worker-nodes") must be between 6 and 30 characters long
  with google_service_account.worker_nodes, on iam.tf line 47
```

## Mechanism

`uat-gcp.yaml` (post-#2042) sets:

- nightly: `DEPLOYMENT_ID = aicr-uat-<run_id>` — 20 chars (run ids are 11 digits), so `account_id` = 33 chars > 30.
- daytime-up: `DEPLOYMENT_ID = aicr-uat-day-<reservation>-<run_id>` — 33 chars for `gcp-h100`, so `account_id` = 46 chars > 30.

The budget is `deployment.id` <= 17 chars (`-system-nodes` / `-worker-nodes` are 13). The pre-#2042 nightly name `aicr-<run_id>` was 16 chars (29-char account_id) — inside the cap, which is why nightlies passed until now. #2042's naming comment checked GKE's 40-char cluster-name cap but not the GCP SA `account_id` cap.

## Reachability

- Deterministic, not transient: every GCP cell (nightly, daytime-up; main and release cells alike — the names are lifecycle-derived, not version-derived) fails at `Bringup Infra`.
[...]

PR #2068, opened 2026-08-05
feat(ci): add UAT orphan janitor (dry-run-first, all clouds)

njhensley:
## Summary

Adds an hourly, all-cloud backstop that finds orphaned UAT deployments and reaps them through the existing cloud actuator. Scheduled runs are report-only; deletion requires an explicit `workflow_dispatch` with `enforce=true` on `main`.

## Motivation / Context

The in-run `Destroy Cluster` step cannot cover teardown misses caused by hard-cancelled runners, failed daytime teardown, abandoned `skip_delete` runs, or false-negative bringup failures. Those misses can leave clusters, networks, resource groups, and Terraform state behind until cloud quotas are exhausted.

This janitor complements the in-run teardown and the focused lifecycle fixes. It is the final reconciliation backstop, not a replacement for teardown at the end of each UAT run.

Fixes: N/A
Related: #2049, #2066, #2067, #2081, #2132

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: UAT workflows and shell tooling

## Implementation Notes

**Reconciliation.** The hourly workflow discovers UAT clusters and cluster-gone network/resource-group remnants in GCP, both AWS accounts (including GB200), and Azure. Candidate discovery is deliberately broad, but deletion eligibility is controlled by an anchored allowlist for the nightly and daytime deployment-name schemas. The daytime expression accepts both the active ADR-017 slug/slot names and legacy reservation names during the migration window.

**Safety gates.** A candidate must contain a numeric run ID, have a completed or definitively purged owning run, and be at least 24 hours old. Unknown run statuses, malformed API responses, non-404 API errors, invalid timestamps, bad configuration values, and discovery failures all fail closed. A daytime candidate is additionally deferred while a UAT lifecycle run is active; the residual check-to-reap TOCTOU window is tracked in #2132 for resolution before scheduled enforcement.

**Destruction.** Reaping uses the same pinned actuator images and `destroy` path as in-run teardown. It breaks stale state locks first, preserves cloud-specific resource ordering, performs the AWS load-balancer cleanup required before VPC deletion, reports per-candidate results, and fails the job if any enforced reap fails.

**Dry-run-first rollout.** The schedule always sets `DRY_RUN=true`. Only an explicit `enforce=true` dispatch on `refs/heads/main` can delete resources, and invalid `DRY_RUN`, age-floor, or cloud values abort before discovery.

## Testing

```bash
bash tools/uat-janitor_test.sh
# PASS: 77/77 decision-logic assertions

shellcheck -x .github/scripts/uat-janitor.sh tools/uat-janitor_test.sh
yamllint -c .yamllint.yaml .github/workflows/uat-janitor.yaml .github/workflows/uat-run.yaml
actionlint -color -shellcheck= .github/workflows/uat-janitor.yaml .github/workflows/uat-run.yaml
git diff --check origin/main...HEAD
# All passed on the rebased head
```

Full `make qualify` was not repeated locally because this change is limited to workflows, shell tooling, and documentation. The scoped executable tests and linters above cover the changed surfaces. The final rebased head also passed the complete GitHub Merge Gate, including Test, Lint, CLI E2E, E2E, Security Scan, CodeQL analysis, actionlint, and malware scanning.

## Risk Assessment

**Rollout notes:** Keep scheduled runs in dry-run mode while reviewing decisions and confirming OIDC/WIF access in every cloud account, including the separate AWS GB200 account. Enforced deletion remains a deliberate manual dispatch. Revisit the cross-workflow lease in #2132 before enabling scheduled enforcement.

## Checklist

### DIFF
--- a/.github/scripts/uat-janitor.sh
+++ b/.github/scripts/uat-janitor.sh
@@ -1,0 +1,569 @@
+#!/usr/bin/env bash
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# UAT orphan janitor — the backstop for the in-run teardown (uat-{gcp,aws,azure}
+# .yaml "Destroy Cluster"). A UAT run can leave its cluster + full resource group
+# (GPU node, VPC network group, router/NAT, resource group, TF state) behind when
+# teardown never runs: a runner hard-killed mid-cancel, a daytime-up held cluster
+# whose evening daytime-down never fired, a manual skip_delete run abandoned, or a
+# Bringup "failure" that stranded a healthy cluster. Those orphans accumulate until
+# they exhaust a cloud quota (in Aug 2026 the GCP NETWORKS quota, 50, hit 48/50).
+#
+# This reaps them. For every cloud resource whose name carries the UAT deployment
+# id (aicr-uat-<run_id> nightly and aicr-uat-day-<reservation-or-slug>-<run_id>
+# daytime — see the ALLOWLIST below for both forms), it
+# asks GitHub whether the OWNING run <run_id> is still active — and only reaps once
+# the run is finished and old enough. It reaps via the SAME actuator `destroy` the
+# in-run teardown uses, so Terraform handles resource ordering (router/NAT before
+# VPC, the GCP default-route quirk, the Azure resource group) correctly.
+#
+# Safety model (this deletes cloud infrastructure — read before editing):
+#   * ALLOWLIST — a candidate must match one of the two deployment-id schemas,
+#     anchored end to end:
+#       ^aicr-uat-[0-9]+$                 nightly, all clouds
+#       ^aicr-uat-day-[a-z0-9-]+-[0-9]+$  daytime, all clouds. The middle segment
+#                                         is deliberately permissive so it matches
+#                                         BOTH the ADR-017 name now used on main —
+#                                         aicr-uat-day-<slug>-<slot>-<run_id>
+#                                         (for example gh1-0) — AND the legacy
+#                                         aicr-uat-day-<reservation>-<run_id>
+#                                         form (for example gcp-h100) during the
+#                                         transition. Tighten to the slug/slot
+#                                         shape once no legacy-held clusters remain.
+#                                         Still anchored: the aicr-uat-day- prefix
+#                                         and a trailing numeric run_id are required.
+#     (GCP nightly stays aicr-uat-<run_id> too: the GKE actuator now bounds the
+#     derived node-SA account_id — mchmarny/cluster#36, image >= v0.5.17.)
+#     A prefix test would admit aicr-uat-unrelated-7, which no workflow produces;
+#     the anchors reject it. Persistent infra (aicr-testgrid-*, anything else)
+#     matches neither schema and is never in scope. Per-cloud discovery is
+#     deliberately coarser — it only nominates CANDIDATES; classify() is the
+#     authoritative gate, and run_id_of/is_daytime remain defense in depth.
+#   * RUN-ID REQUIRED — a name with no trailing numeric run_id is skipped. We
+#     never guess.
+#   * NOT-ACTIVE — the owning run must be `completed` (or purged/404). This is an
+#     allowlist: queued / in_progress / waiting / pending / requested are skipped
+#     because the cluster is live, and ANY other status — including one GitHub
+#     adds after this was written — is skipped as indeterminate.
+#   * FAIL CLOSED — any GitHub API error that is not a definitive run-level 404
+#     skips the candidate this cycle. We never delete on ambiguous liveness.
+#     "Definitive" is narrow on purpose: the runs endpoint must first be proven
+#     reachable (else a repo-level 404 would mark every candidate an orphan), the
+#     error must be an `HTTP 404` and not merely text containing "404", and a
+#     200 whose body lacks `status`/`updated_at` is ambiguous, not a green light.
+#     An unparsable `updated_at` skips rather than reading as infinitely old.
+#   * AGE FLOOR — keyed off the RUN's updated_at (uniform across clouds), not the
+#     cloud resource's create time. Nightly/ad-hoc: NIGHTLY_MIN_AGE_HOURS (default
+#     24h). Well above the 280-min job cap, so a still-finishing run's just-created
+#     cluster is never touched — and, deliberately, long enough that a nightly
+#     `skip_delete` run (a supported debugging hold, uat-{aws,azure,gcp}.yaml) can
+#     survive a full working day before the janitor reclaims it. There is NO API
+#     signal that distinguishes a skip_delete hold from a genuine orphan, so the
+#     floor is the only lever: a real mid-cancel orphan therefore also persists up
+#     to ~a day, which is acceptable — quota pressure builds over days, and the
+#     hourly janitor reaps within the hour once the floor clears. (Lower it only if
+#     a cloud's quota headroom demands faster reclamation.) Daytime:
+#     DAYTIME_MIN_AGE_HOURS (default 24h). daytime-up is on-demand (workflow_dispatch), so its run is
+#     already `completed` while the cluster is legitimately HELD for the working
+#     day; the sole daytime schedule is the evening daytime-down (uat-daytime.yaml,
+#     cron '0 2 * * *' UTC), which reaps the held cluster each night. A legitimate
+#     hold therefore always lasts < 24h (up just after one 02:00 down, torn down at
+#     the next), so a daytime cluster older than 24h has provably survived an
+#     evening daytime-down = an orphan the nightly teardown missed. (Do NOT lower
+#     this to the nightly floor: a completed daytime-up run does not mean the
+#     cluster should be gone.)
+#   * DRY_RUN — defaults to true. It reports every decision and never deletes until
+#     a caller sets DRY_RUN=false (the workflow's `enforce` dispatch input). Only
+#     the exact strings `true`/`false` are accepted; anything else aborts before
+#     discovery rather than being resolved toward enforce. In dry-run the summary
+#     counts "would-reap", never "reaped" — a report must not claim a destroy
+#     that did not happen.
+#
+# Usage:   DRY_RUN=true ./uat-janitor.sh <gcp|aws|azure>
+# Requires: gh (GH_TOKEN in env, actions:read), jq, yq, docker, and the cloud CLI
+#           + credentials for <cloud> already configured by the caller.
+
+set -euo pipefail
+
+CLOUD="${1:?usage: uat-janitor.sh <gcp|aws|azure>}"
+DRY_RUN="${DRY_RUN:-true}"
+REPO="${GITHUB_REPOSITORY:-NVIDIA/aicr}"
+# AWS/Azure discovery only. GCP matches a second shape too (see ALLOWLIST above),
+# so it spells its patterns out inline rather than using this.
+NAME_PREFIX="aicr-uat-"
+NIGHTLY_MIN_AGE_HOURS="${NIGHTLY_MIN_AGE_HOURS:-24}"
+DAYTIME_MIN_AGE_HOURS="${DAYTIME_MIN_AGE_HOURS:-24}"
+
+# Directory of this script, so sibling helpers (uat-aws-cleanup-lb.sh) resolve
+# regardless of the caller's CWD.
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Probed once by main(). A path-level 404 (repo renamed, GITHUB_REPOSITORY wrong,
+# token missing actions:read) is indistinguishable from a purged-run 404 when
+# looking at a single candidate — and would classify the ENTIRE fleet as orphaned.
+# So a 404 is only ever read as "run purged" once the runs endpoint itself is
+# proven reachable. Default `no` keeps a sourced/never-probed caller fail-closed.
+REPO_REACHABLE="${REPO_REACHABLE:-no}"
+
+log() { echo "[janitor:${CLOUD}] $*"; }
+
+# --- pure decision helpers -------------------------------------------------
+
+# Trailing all-digit segment (the run id). Fails (non-zero) if the last
+# hyphen-delimited segment is not purely numeric — we then skip the candidate.
+run_id_of() {
+  local last="${1##*-}"
+  case "$last" in
+    '' | *[!0-9]*) return 1 ;;
+  esac
+  printf '%s' "$last"
+}
+
+is_daytime() { case "$1" in aicr-uat-day-*) return 0 ;; *) return 1 ;; esac; }
+
+# DRY_RUN gates every destroy, and only the exact string `true` holds it back.
+# Anything else would select the enforce path, so a typo (DRY_RUN=flase) or a
+# shell-truthy value (1, yes, TRUE) must be rejected outright rather than read as
+# permission to delete. Only `true`/`false` are accepted.
+valid_dry_run() { case "$1" in true | false) return 0 ;; *) return 1 ;; esac; }
+
+# The age floors gate every destroy on the run's age. A non-integer floor (e.g.
+# "24h") makes `[ "$age" -lt "$min" ]` error under an `if` — read as false, i.e.
+# REAP — and a negative floor clears for every age. The shipped workflow never
+# sets these, but the script accepts them as configuration, so validate them up
+# front like DRY_RUN. `*[!0-9]*` rejects empty, non-numeric, and negative alike;
+# the length cap rejects a value big enough to overflow bash arithmetic (past
+# ~2^63 the same `-lt` test errors with status 2 and fails open to REAP). Five
+# digits (<=99999h ~= 11y) is far more than any real floor and well under overflow.
+valid_age_floor() {
+  case "$1" in '' | *[!0-9]*) return 1 ;; esac
+  [ "${#1}" -le 5 ]
+}
+
+# Whole hours between an ISO-8601 timestamp and now (GNU date; runner is ubuntu).
+# Returns non-zero on an unparsable timestamp rather than substituting an epoch:
+# a 0 would read as an infinitely-old deployment and clear every age floor.
+age_hours_since() {
+  local t
+  t="$(date -u -d "$1" +%s 2>/dev/null)" || return 1
+  [ -n "$t" ] || return 1
+  echo $(( ( $(date -u +%s) - t ) / 3600 ))
+}
+
+# Is a UAT lifecycle run currently in flight? A daytime cluster's NAME carries the
+# daytime-UP run id, so classify()'s run-status gate cannot see an evening
+# daytime-DOWN tearing down the SAME reservation (a different run). The teardown
+# always executes as a uat-run.yaml run — whether the evening scheduler dispatched
+# it (uat-daytime.yaml's drive job dispatches uat-run.yaml as its own top-level
+# run) or an operator ran `gh workflow run uat-run.yaml -f lifecycle=daytime-down`
+# directly — so THAT is the workflow we watch, not the scheduler wrapper (whose
+# watcher can time out before the child). Coarse but safe: ANY in-flight uat-run
+# run defers ALL daytime reaps this cycle (the janitor is hourly, so the cost is a
+# one-cycle delay).
+#
+# This is a liveness PROBE, not a mutex: it collapses the common race (a
+# scheduled/dispatched down already running) to near zero, but a residual TOCTOU
+# window remains — a down that STARTS in the ~1h between this check and the reap.
+# The 24h floor and DRY_RUN-default backstop that window; a shared cross-workflow
+# lease (#2132) is the tracked follow-up for a hard "never races" guarantee.
+#
+# Fails CLOSED: any API error, or a 200 whose body is not a well-formed run list
+# (`{"message":...}`, `{"workflow_runs":null}`), reads as "active" (defer) rather
+# than idle. Uses `gh` raw + external `jq` (like classify()) so the harness drives it.
+LIFECYCLE_WORKFLOW="uat-run.yaml"
+uat_lifecycle_active() {
+  local json count
+  # Filter to in_progress SERVER-SIDE (?status=in_progress) rather than reading the
+  # newest N mixed-status runs and filtering here — that removes any dependence on
+  # dispatch volume fitting in one page. in_progress is the state of a run actually
+  # executing a teardown; a run still queued / about to start is the documented
+  # TOCTOU residual, backstopped by the 24h floor + DRY_RUN default. per_page=100
+  # is far above the number of concurrently-executing uat-run runs (the lease
+  # serializes per reservation), so the count is effectively unbounded for our need.
+  json="$(gh api "repos/${REPO}/actions/workflows/${LIFECYCLE_WORKFLOW}/runs?status=in_progress&per_page=100" 2>/dev/null)" || return 0
+  # Fail CLOSED on a malformed 200: require an actual array, else emit a non-numeric
+  # marker the case below reads as "active" (defer). `.workflow_runs | length` on a
+  # missing/null field would otherwise be 0 = idle.
+  count="$(jq -r 'if (.workflow_runs | type) == "array" then (.workflow_runs | length) else "malformed" end' <<<"$json" 2>/dev/null)" || return 0
+  case "$count" in '' | *[!0-9]*) return 0 ;; esac
+  [ "$count" -gt 0 ]
+}
+
+# Echo "REAP" or "SKIP:<reason>" for a deployment id. Consults the GitHub run for
+# both liveness and age so the decision is identical across clouds.
+classify() {
+  local id="$1" run_id json status upd age min
+  # Exactly the two supported deployment-id schemas (see the ALLOWLIST bullet in
+  # the header) — fully anchored, so merely starting with `aicr-` is not enough and
+  # a shape we do not generate (aicr-uat-unrelated-7, aicr-testgrid-vpc) can never
+  # reach the actuator. discover_* is deliberately coarser; THIS is the gate.
+  if [[ ! "$id" =~ ^aicr-uat-[0-9]+$ ]] &&
+     [[ ! "$id" =~ ^aicr-uat-day-[a-z0-9-]+-[0-9]+$ ]]; then
+    echo "SKIP:not-allowlisted"; return
+  fi
+  run_id="$(run_id_of "$id")" || { echo "SKIP:no-run-id"; return; }
+
+  local errf; errf="$(mktemp)"
+  if json="$(gh api "repos/${REPO}/actions/runs/${run_id}" 2>"$errf")"; then
+    rm -f "$errf"
+    status="$(jq -r '.status // empty' <<<"$json")"
+    upd="$(jq -r '.updated_at // empty' <<<"$json")"
+    if [ -z "$status" ] || [ -z "$upd" ]; then
+      echo "SKIP:incomplete-run-json(run ${run_id})"; return   # fail closed
+    fi
+  elif [ "$REPO_REACHABLE" = yes ] && grep -q 'HTTP 404' "$errf"; then
+    rm -f "$errf"
+    # Run purged from history (GitHub retains ~months) => long dead => orphan.
+    status="completed"; upd=""
+  else
+    rm -f "$errf"
+    echo "SKIP:gh-api-error(run ${run_id})"; return   # fail closed
+  fi
+
+  # Allowlist, not denylist: ONLY a confirmed `completed` run may be reaped. The
+  # active statuses are named separately because they are a routine skip an
+  # operator should not have to think about; anything else is a status GitHub did
+  # not have when this was written, and fails closed rather than falling through.
+  case "$status" in
+    completed) ;;
+    queued | in_progress | waiting | pending | requested)
+      echo "SKIP:run-active(${status},run ${run_id})"; return ;;
+    *)
+      echo "SKIP:run-status-unsupported(${status},run ${run_id})"; return ;;
+  esac
+
+  if [ -n "$upd" ]; then
+    age="$(age_hours_since "$upd")" || { echo "SKIP:bad-updated-at(${upd})"; return; }
+  else
+    age=999999   # confirmed run-level 404: purged from history, past every floor
+  fi
+  if is_daytime "$id"; then min="$DAYTIME_MIN_AGE_HOURS"; else min="$NIGHTLY_MIN_AGE_HOURS"; fi
+  if [ "$age" -lt "$min" ]; then
+    echo "SKIP:too-young(${age}h < ${min}h)"; return
+  fi
+  # Daytime last-mile race guard (see the concurrency note in uat-janitor.yaml):
+  # only daytime needs it — a nightly cluster's owning run IS the one the gate
+  # above already checked, but a daytime name points at the daytime-UP run, not
+  # the daytime-DOWN run (a uat-run.yaml run) that tears it down. Defer if any
+  # uat-run lifecycle run is in flight. Reduces, does not eliminate, the race
+  # (residual TOCTOU per uat_lifecycle_active); fails closed.
+  if is_daytime "$id" && uat_lifecycle_active; then
+    echo "SKIP:lifecycle-active(run ${run_id})"; return
+  fi
+  echo "REAP"
+}
+
+# --- reaping (actuator destroy) --------------------------------------------
+
+# The Azure reap stages a copy of the runner's az credentials for the actuator
+# container. This is module-level (not a `local`) so main() can arm an EXIT trap
+# for it BEFORE any credential is written, which a `trap ... RETURN` inside reap
+# cannot do safely: a RETURN trap set in a function is not cleared when that
+# function returns, so it fires again on the caller's return with the variable
+# out of scope — under `set -e` that aborts the whole janitor mid-sweep.
+AZ_MOUNT=""
+cleanup_az_mount() {
+  [ -n "$AZ_MOUNT" ] || return 0
+  rm -rf "$AZ_MOUNT"
+  AZ_MOUNT=""
+}
+
+# Break a stale Terraform state lock before destroying. Orphans from a run
+# cancelled mid-bringup carry an orphaned lock that would fail every destroy
+# attempt (see uat-*.yaml teardown). Best-effort; a genuinely-absent lock is the
+# common case. AWS has no state lock (S3 backend without dynamodb/use_lockfile).
+break_state_lock() {
+  local id="$1" loc
+  case "$CLOUD" in
+    gcp)
+      loc="$(yq -r '.deployment.location' "$JANITOR_CONFIG")"
+      gcloud storage rm \
+        "gs://cluster-state-${GCP_PROJECT_ID}/deployments/${loc}/${id}/default.tflock" \
+        2>/dev/null && log "cleared stale GCS state lock for ${id}" || true
+      ;;
+    azure)
+      loc="$(yq -r '.deployment.location' "$JANITOR_CONFIG")"
+      local rg="cluster-state-rg" sa key blob="deployments/${loc}/${id}/terraform.tfstate"
+      sa="$(az storage account list -g "$rg" --subscription "$AZURE_SUBSCRIPTION_ID" \
+        --query "[?starts_with(name, 'clst')].name | [0]" -o tsv 2>/dev/null || true)"
+      if [ -n "${sa:-}" ]; then
+        key="$(az storage account keys list -g "$rg" -n "$sa" \
+          --subscription "$AZURE_SUBSCRIPTION_ID" --query '[0].value' -o tsv 2>/dev/null || true)"
+        # Hand the account key to `az` through AZURE_STORAGE_KEY, not --account-key,
+        # so the full-access tfstate key never appears in argv (world-readable via
+        # /proc/<pid>/cmdline). Mirrors the proven pattern in uat-azure.yaml teardown.
+        AZURE_STORAGE_ACCOUNT="$sa" AZURE_STORAGE_KEY="$key" \
+          az storage blob lease break --account-name "$sa" \
+          --container-name tfstate --blob-name "$blob" >/dev/null 2>&1 \
+          && log "broke stale state-blob lease for ${id}" || true
+      fi
+      ;;
+  esac
+}
+
+# One EKS actuator destroy pass for the staged config <cfg>. Factored out so the
+# AWS reap can run it, sweep leaked LB resources, and run it again.
+aws_actuator_destroy() {
+  docker run --rm \
+    -e CONFIG_CONTENT="$(base64 <"$1" | tr -d '\n')" \
+    -e AUTO_APPROVE=true \
+    -e AWS_ACCESS_KEY_ID="$AWS_ACCESS_KEY_ID" \
+    -e AWS_SECRET_ACCESS_KEY="$AWS_SECRET_ACCESS_KEY" \
+    -e AWS_SESSION_TOKEN="${AWS_SESSION_TOKEN:-}" \
+    "$JANITOR_ACTUATOR_IMAGE" apply
+}
+
+# Destroy one orphaned deployment via the actuator (destroy == apply w/ destroy=true).
+# Returns the actuator's exit status so the caller counts only destroys that
+# actually happened — the rollout asks operators to trust these logs before
+# enabling enforce, so a reported reap must be a real one.
+reap() {
+  local id="$1" cfg rc=0
+  # `reap` is invoked as `if reap "$id"`, which disables errexit for this whole
+  # body — a failure while staging the destroy config does NOT abort here. So
+  # check each step explicitly and bail BEFORE docker: an unmodified or
+  # half-written config still carries the committed placeholder deployment.id and
+  # destroy=false, which in ENFORCE mode would make the actuator APPLY
+  # (provision/update) a cluster while this logged a reap. Fail closed instead.
+  if ! cfg="$(mktemp --suffix=.yaml)"; then
+    echo "::warning::janitor could not create temp config for ${id}"
+    return 1
+  fi
+  if ! yq '.' "$JANITOR_CONFIG" >"$cfg" ||
+     ! DEPLOY_ID="$id" yq -i '.deployment.id = strenv(DEPLOY_ID) | .deployment.destroy = true' "$cfg"; then
+    echo "::warning::janitor could not stage destroy config for ${id} (destroy NOT attempted)"
+    rm -f "$cfg"
+    return 1
+  fi
+
+  if [ "$DRY_RUN" = "true" ]; then
+    log "DRY-RUN would reap: ${id}"
+    rm -f "$cfg"; return 0
+  fi
+
+  log "reaping: ${id}"
+  break_state_lock "$id"
+  case "$CLOUD" in
+    gcp)
+      docker run --rm \
+        -e CONFIG_CONTENT="$(base64 <"$cfg" | tr -d '\n')" \
+        -e AUTO_APPROVE=true \
+        -e KEY_CONTENT="$(base64 <"$GOOGLE_GHA_CREDS_PATH" | tr -d '\n')" \
+        "$JANITOR_ACTUATOR_IMAGE" apply || rc=$?
+      ;;
+    aws)
+      # A LoadBalancer Service (AWS inference UAT deploys agentgateway) leaves a
+      # classic ELB + k8s-elb-* SG OUTSIDE Terraform state; `destroy` then fails
+      # at DeleteVpc with DependencyViolation and the VPC leaks — the exact orphan
+      # class this janitor exists to reap, failing identically every cycle without
+      # this. Mirror uat-aws.yaml teardown: delete the LB Services gracefully while
+      # the cluster still answers (best-effort — no-op once the cluster is gone),
+      # destroy, and if that fails, sweep the leaked ELB/SG and retry once so a
+      # VPC-only remnant still tears down. Both helper modes are scoped to this
+      # deployment id's ownership tag and always exit 0.
+      "$SCRIPT_DIR/uat-aws-cleanup-lb.sh" graceful "$id" || true
+      aws_actuator_destroy "$cfg" || rc=$?
+      if [ "$rc" -ne 0 ]; then
+        log "aws destroy failed for ${id} (rc=${rc}); sweeping leaked LB resources and retrying once"
+        "$SCRIPT_DIR/uat-aws-cleanup-lb.sh" sweep "$id" || true
+        rc=0
+        aws_actuator_destroy "$cfg" || rc=$?
+      fi
+      ;;
+    azure)
+      # The actuator container runs as uid `builder` (not the runner's uid) and
+      # writes to its az config dir on startup, so it needs a writable copy —
+      # `chmod -R a+rwX` on a throwaway mirrors the proven Bringup/Destroy steps
+      # in uat-azure.yaml. Do not tighten the mode without confirming the image's
+      # uid, or every Azure reap fails. mktemp -d is 0700, so the copy is private
+      # until the chmod, and the runner is ephemeral and single-tenant.
+      #
+      # Cleanup is already armed: AZ_MOUNT is module-level and main() installs an
+      # EXIT trap for it before any reap runs, so the credentials cannot survive
+      # even a `set -e` abort between mktemp and the copy. The inline call below
+      # removes them at the earliest possible moment rather than at exit.
+      AZ_MOUNT="$(mktemp -d)"
+      if cp -a "$HOME/.azure/." "$AZ_MOUNT/" && chmod -R a+rwX "$AZ_MOUNT"; then
+        docker run --rm \
+          -e CONFIG_CONTENT="$(base64 <"$cfg" | tr -d '\n')" \
+          -e AUTO_APPROVE=true \
+          -v "$AZ_MOUNT:/home/builder/.azure" \
+          "$JANITOR_ACTUATOR_IMAGE" apply || rc=$?
+      else
+        rc=1
+        echo "::warning::janitor could not stage az config for ${id}"
+      fi
+      cleanup_az_mount
+      ;;
+  esac
+  rm -f "$cfg"
+
+  if [ "$rc" -eq 0 ]; then
+    log "reaped: ${id}"
+  else
+    echo "::warning::janitor reap failed for ${id} (rc=${rc}; will retry next run)"
+  fi
+  return "$rc"
+}
+
+# --- per-cloud discovery ---------------------------------------------------
+# Each emits candidate deployment ids (one per line, matching the header's
+# ALLOWLIST), deduped by the caller. Discovery is intentionally broad (clusters
+# AND their networks /
+# resource groups) so a cluster that was already deleted but left its network
+# group / resource group behind is still caught.
+
+# All GCP deployment ids (nightly aicr-uat-<run_id> and daytime aicr-uat-day-*)
+# share the aicr-uat- prefix, same as AWS/Azure.
+# This is a coarse CANDIDATE filter; classify() is the authoritative, schema-anchored
+# gate.
+# Run one cloud list command and echo its stdout. A NON-ZERO exit (missing CLI,
+# expired creds, a per-API authorization gap the auth step did not catch, a
+# throttle) is surfaced as a ::warning:: rather than swallowed — otherwise a
+# failed listing looks identical to an empty fleet and the run reports "nothing
+# to reap" while orphans persist. Discovery stays best-effort (the run continues
+# and reaps whatever DID list), but the failure is now visible in the log.
+run_discovery() {
+  local label="$1"; shift
+  local out errf rc
+  errf="$(mktemp)"
+  out="$("$@" 2>"$errf")" && rc=0 || rc=$?
+  if [ "$rc" -ne 0 ]; then
+    echo "::warning::${CLOUD} discovery '${label}' failed (rc=${rc}); its orphans are NOT reconciled this run: $(tr '\n' ' ' <"$errf")" >&2
+  fi
+  rm -f "$errf"
+  printf '%s' "$out"
+}
+
+discover_gcp() {
+  run_discovery "GKE clusters" gcloud container clusters list \
+    --project "$GCP_PROJECT_ID" --format="value(name)" | grep -E "^${NAME_PREFIX}" || true
+  # A deployment's VPC is named <id>-vpc; recover <id> for clusters already gone.
+  # The ^ anchor means a foreign name merely ending in -vpc cannot be captured.
+  run_discovery "GCP networks" gcloud compute networks list \
+    --project "$GCP_PROJECT_ID" --format="value(name)" | sed -nE "s/^(${NAME_PREFIX}.*)-vpc$/\1/p" || true
+}
+
+discover_aws() {
+  run_discovery "EKS clusters" aws eks list-clusters \
+    --region "$AWS_REGION" --query 'clusters[]' --output text | tr '\t' '\n' | grep -E "^${NAME_PREFIX}" || true
+  # The EKS actuator tags the VPC Name=<id>-vpc (network.tf); recover <id> for a
+  # cluster already deleted but whose VPC group still leaks (mirrors the GCP -vpc
+  # and Azure -rg recovery so AWS is not the one path that misses network-only
+  # orphans). --output text tab-joins Name-tag values within a VPC.
+  run_discovery "EC2 VPCs" aws ec2 describe-vpcs --region "$AWS_REGION" \
+    --query "Vpcs[].Tags[?Key=='Name'].Value | [] | [?starts_with(@, '${NAME_PREFIX}')]" --output text \
+    | tr '\t' '\n' | sed -nE "s/^(${NAME_PREFIX}.*)-vpc$/\1/p" || true
+}
+
+discover_azure() {
+  run_discovery "AKS clusters" az aks list \
+    --subscription "$AZURE_SUBSCRIPTION_ID" --query '[].name' -o tsv | grep -E "^${NAME_PREFIX}" || true
+  # The actuator's resource group is <id>-rg; recover <id> for clusters already gone.
+  run_discovery "Azure resource groups" az group list \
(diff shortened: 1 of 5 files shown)


.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,270 @@
...
+}
+
+is_daytime() { case "$1" in aicr-uat-day-*) return 0 ;; *) return 1 ;; esac; }
+
+# Whole hours between an ISO-8601 timestamp and now (GNU date; runner is ubuntu).
+age_hours_since() {
+  local t
+  t="$(date -u -d "$1" +%s 2>/dev/null || echo 0)"
+  echo $(( ( $(date -u +%s) - t ) / 3600 ))
+}
```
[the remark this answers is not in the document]
njhensley: Fixed. `age_hours_since` now returns non-zero on an unparsable timestamp instead of substituting epoch 0, and `classify` skips with `SKIP:bad-updated-at`.

Tightened the sibling path too: `.status`/`.updated_at` are read with `// empty`, and a 200 whose body lacks either field now yields `SKIP:incomplete-run-json` rather than falling through the `case` as "not active". Regression-locked in `tools/uat-janitor_test.sh`.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,270 @@
...
+    status="$(jq -r '.status' <<<"$json")"
+    upd="$(jq -r '.updated_at' <<<"$json")"
+  elif grep -q '404' "$errf"; then
+    rm -f "$errf"
+    # Run purged from history (GitHub retains ~months) => long dead => orphan.
+    status="completed"; upd=""
+  else
+    rm -f "$errf"
+    echo "SKIP:gh-api-error(run ${run_id})"; return   # fail closed
+  fi
```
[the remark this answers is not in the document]
njhensley: Fixed, as proposed. The runs endpoint is probed once in `main()`, and `classify` only reads a 404 as "run purged" when `REPO_REACHABLE=yes` **and** the error matches `HTTP 404` (not a bare `404` substring).

The global defaults to `no`, so a sourced or never-probed caller stays fail-closed. Two assertions cover it: a 404 with the repo unproven must SKIP, and only a proven-reachable repo may REAP.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,270 @@
...
+      log "REAP  ${id}"
+      reap "$id"
+      reaped=$((reaped + 1))
+    else
+      log "skip  ${id} -> ${verdict#SKIP:}"
+      skipped=$((skipped + 1))
+    fi
+  done <<<"$candidates"
+
+  log "summary: ${reaped} reaped, ${skipped} skipped$([ "$DRY_RUN" = true ] && echo ' (dry-run: nothing deleted)')"
```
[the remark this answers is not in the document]
njhensley: Fixed. `reap` captures the actuator status per cloud branch into `rc` and returns it; `main` increments `reaped` only when `reap` succeeds and tracks `failed` separately. The summary and the new job-summary table both report reaped/failed/skipped.

.github/workflows/uat-janitor.yaml
```diff
@@ -0,0 +1,158 @@
...
+      - name: Authenticate to Azure
+        uses: azure/login@532459ea530d8321f2fb9bb10d1e0bcf23869a43 # v3.0.0
+        with:
+          client-id: "047f0427-5dcf-4e7c-97c3-9c71d92c5abb"
+          tenant-id: "43083d15-7273-40c1-b7db-39efd9ccc17a"
+          subscription-id: ${{ env.AZURE_SUBSCRIPTION_ID }}
+      - name: Reconcile Azure orphans
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: ./.github/scripts/uat-janitor.sh azure
```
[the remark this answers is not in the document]
njhensley: Fixed. The script writes a reaped/failed/skipped table to `GITHUB_STEP_SUMMARY` and `main` exits non-zero when `failed > 0`, so a persistent stale lock or permissions gap fails the job instead of staying buried in a `::warning::`.

Dry-run cannot fail the job — `reap` short-circuits before the actuator, so `failed` stays 0.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,270 @@
...
+    --format="value(name)" 2>/dev/null | grep -E "^${NAME_PREFIX}" || true
+  # A deployment's VPC is named <id>-vpc; recover <id> for clusters already gone.
+  gcloud compute networks list --project "$GCP_PROJECT_ID" \
+    --format="value(name)" 2>/dev/null | sed -n "s/^\(${NAME_PREFIX}.*\)-vpc$/\1/p" || true
+}
+
+discover_aws() {
+  aws eks list-clusters --region "$AWS_REGION" --query 'clusters[]' --output text 2>/dev/null \
+    | tr '\t' '\n' | grep -E "^${NAME_PREFIX}" || true
+}
```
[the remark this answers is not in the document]
njhensley: Holding this one for the author rather than implementing it.

Adding VPC discovery *widens* the set of resources this destructive job will consider, and — as you note — the authoritative AWS ownership tag/naming convention lives in the actuator, not in this repo, so it can't be confirmed from here. Expanding the reap surface on an unverified naming assumption is the one change worth being slow about. Left as-is pending the author confirming the actuator's tagging.

Note the GCP and Azure discovery paths already recover IDs from leftover networks/resource groups, so this is an AWS coverage gap, not a correctness bug.

.github/workflows/uat-janitor.yaml
```diff
@@ -0,0 +1,158 @@
...
+        default: false
+
+permissions:
+  contents: read
+
+# Never let two janitor runs reconcile the same cloud state concurrently; let an
+# in-flight reap finish rather than cancelling it mid-destroy.
+concurrency:
+  group: uat-janitor
+  cancel-in-progress: false
```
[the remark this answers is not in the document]
njhensley: Real concern, but holding it for the author — it changes the concurrency design of the destructive path, which isn't a mechanical fix.

Two things make it non-trivial: GitHub `concurrency` groups are static per workflow run and evaluated at dispatch, so a janitor job can't join `uat-<reservation>` per-candidate the way this suggests; it would need a different exclusion mechanism. And the age floors already cover the common case (a daytime cluster must be >= 24h past its run's `updated_at`, a nightly one >= 6h), so the residual race is narrow.

Flagging rather than fixing so the author can decide on the mechanism.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,331 @@
...
+    status="completed"; upd=""
+  else
+    rm -f "$errf"
+    echo "SKIP:gh-api-error(run ${run_id})"; return   # fail closed
+  fi
+
+  case "$status" in
+    queued | in_progress | waiting | pending | requested)
+      echo "SKIP:run-active(${status},run ${run_id})"; return ;;
+  esac
```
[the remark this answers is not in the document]
njhensley: Fixed. The `case` is now an allowlist — `completed` falls through to the age check, the five active statuses keep their distinct `SKIP:run-active` message (a routine skip an operator shouldn't have to think about), and everything else returns `SKIP:run-status-unsupported`.

Regression cases added for three unrecognized statuses. The header's NOT-ACTIVE bullet now states the allowlist explicitly.

tools/uat-janitor_test.sh
```diff
@@ -0,0 +1,153 @@
...
+GH_STUB="ok:completed:$(ago 99)"
+check "rejects non-allowlisted name" "SKIP:not-allowlisted" "$(classify aicr-testgrid-123)"
+check "rejects bare prefix match"    "SKIP:not-allowlisted" "$(classify some-aicr-uat-123)"
+check "rejects missing run id"       "SKIP:no-run-id"       "$(classify aicr-uat-manual)"
+
+echo "classify (liveness gate):"
+for st in queued in_progress waiting pending requested; do
+    GH_STUB="ok:${st}:$(ago 99)"
+    check "skips active run (${st})" "SKIP:run-active(${st},run 7)" "$(classify aicr-uat-7)"
+done
```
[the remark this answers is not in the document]
njhensley: Fixed. The `case` is now an allowlist — `completed` falls through to the age check, the five active statuses keep their distinct `SKIP:run-active` message (a routine skip an operator shouldn't have to think about), and everything else returns `SKIP:run-status-unsupported`.

Regression cases added for three unrecognized statuses. The header's NOT-ACTIVE bullet now states the allowlist explicitly.

tools/uat-janitor_test.sh
```diff
@@ -0,0 +1,153 @@
...
+age_hours_since "not-a-timestamp" >/dev/null 2>&1
+check "fails closed on garbage" "1" "$?"
+age_hours_since "null" >/dev/null 2>&1
+check "fails closed on literal null" "1" "$?"
+
+echo "classify (allowlist + run id):"
+GH_STUB="ok:completed:$(ago 99)"
+check "rejects non-allowlisted name" "SKIP:not-allowlisted" "$(classify aicr-testgrid-123)"
+check "rejects bare prefix match"    "SKIP:not-allowlisted" "$(classify some-aicr-uat-123)"
+check "rejects missing run id"       "SKIP:no-run-id"       "$(classify aicr-uat-manual)"
```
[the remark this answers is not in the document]
njhensley: Holding this one for the author — the `^aicr-uat-` allowlist is on the explicit do-not-change-autonomously list for this PR, since it's the outermost guard on a job that destroys cloud infrastructure.

For the record, the finding looks correct and the repo backs it. All three clouds build the id identically (`uat-{aws,gcp,azure}.yaml`):

```
daytime: format('aicr-uat-day-{0}-{1}', inputs.reservation, github.run_id)
nightly: format('aicr-uat-{0}', github.run_id)
```

so `aicr-uat-unrelated-7` matches neither schema yet currently classifies as REAP. Tightening to `^aicr-uat-[0-9]+$` and `^aicr-uat-day-[a-z0-9-]+-[0-9]+$` would close it.

The one reason it needs a human rather than a bot: it's strictly *narrowing*, so the risk isn't a bad deletion — it's that a hand-rolled actuator deployment with a custom id would silently stop being reaped. That trade-off is the author's call.

(The committed placeholder id `aicr-uat` is already safe either way — `run_id_of` rejects a non-numeric tail.)

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,340 @@
...
+#     a caller sets DRY_RUN=false (the workflow's `enforce` dispatch input).
+#
+# Usage:   DRY_RUN=true ./uat-janitor.sh <gcp|aws|azure>
+# Requires: gh (GH_TOKEN in env, actions:read), jq, yq, docker, and the cloud CLI
+#           + credentials for <cloud> already configured by the caller.
+
+set -euo pipefail
+
+CLOUD="${1:?usage: uat-janitor.sh <gcp|aws|azure>}"
+DRY_RUN="${DRY_RUN:-true}"
```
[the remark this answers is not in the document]
njhensley: Fixed. Added a `valid_dry_run` guard that accepts only the exact strings `true`/`false`; `main` calls it before any discovery and returns 2 otherwise, so an unrecognized value aborts instead of resolving toward enforce.

Regression cases cover `flase`, `TRUE`, `True`, `1`, `yes`, `on`, empty, whitespace, and a trailing space — all must reject. The header's DRY_RUN bullet documents it.

(The workflow's `${{ inputs.enforce == true && 'false' || 'true' }}` already yields only those two literals, so this is defense for local and ad-hoc invocation.)

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,340 @@
...
+
+  local mode; mode="$([ "$DRY_RUN" = true ] && echo dry-run || echo enforce)"
+  log "summary: ${reaped} reaped, ${failed} failed, ${skipped} skipped (${mode})"
+  if [ -n "${GITHUB_STEP_SUMMARY:-}" ]; then
+    {
+      echo "### UAT janitor — ${CLOUD} (${mode})"
+      echo
+      echo "| reaped | failed | skipped |"
+      echo "| ---: | ---: | ---: |"
+      echo "| ${reaped} | ${failed} | ${skipped} |"
```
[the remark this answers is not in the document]
njhensley: Fixed. The summary and job-summary column are now labelled by mode — `would-reap` in dry-run, `reaped` only in enforce — so a dry-run report can never claim a destroy that did not happen. Given the rollout asks operators to trust these logs before flipping to enforce, that mislabel was worth closing.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,361 @@
...
+
+# Echo "REAP" or "SKIP:<reason>" for a deployment id. Consults the GitHub run for
+# both liveness and age so the decision is identical across clouds.
+classify() {
+  local id="$1" run_id json status upd age min
+  case "$id" in
+    ${NAME_PREFIX}*) ;;
+    *) echo "SKIP:not-allowlisted"; return ;;
+  esac
+  run_id="$(run_id_of "$id")" || { echo "SKIP:no-run-id"; return; }
```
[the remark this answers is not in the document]
njhensley: This is the same finding as the thread on `tools/uat-janitor_test.sh` (#discussion_r3722669165), where you already agreed to leave it for the author — same disposition here.

Recapping so it isn't lost: the finding looks correct, and the naming is unambiguous in-repo (all three of `uat-{aws,gcp,azure}.yaml` build the id as `format('aicr-uat-day-{0}-{1}', inputs.reservation, github.run_id)` or `format('aicr-uat-{0}', github.run_id)`), so `aicr-uat-unrelated-123` matches neither schema yet currently classifies as REAP.

Not applying it autonomously because the `^aicr-uat-` allowlist is the outermost guard on a job that destroys cloud infrastructure and is explicitly the author's call for this PR. Flagged for them with the concrete patch.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,270 @@
...
+      cp -a "$HOME/.azure/." "$az_mount/"
+      chmod -R a+rwX "$az_mount"
+      docker run --rm \
+        -e CONFIG_CONTENT="$(base64 <"$cfg" | tr -d '\n')" \
+        -e AUTO_APPROVE=true \
+        -v "$az_mount:/home/builder/.azure" \
+        "$JANITOR_ACTUATOR_IMAGE" apply \
+        && log "reaped: ${id}" \
+        || echo "::warning::janitor reap failed for ${id} (will retry next run)"
+      rm -rf "$az_mount"
```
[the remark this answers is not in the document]
njhensley: Partially addressed — taking the cleanup half, holding the permission half.

**Done — guaranteed cleanup.** Staging is now inside an `if`, so `set -e` cannot abort between `mktemp -d` and the copy, and the `rm -rf "$az_mount"` is reached on every path including a failed `docker run` (which previously could not fail the branch, but now sets `rc`). A staging failure sets `rc=1` and warns rather than leaving the mount behind.

**Not done — tightening the mode.** `chmod -R a+rwX` is copied verbatim from the two proven actuator invocations in `uat-azure.yaml` (Bringup L494-496, Destroy L1044-1046). The actuator image's uid isn't asserted anywhere in this repo — the `builder` name comes from the mount path — so narrowing the mode to a specific owner is a guess that would fail *closed on every Azure reap*, and it isn't testable from here. I'd rather leave the janitor consistent with the path that's known to work than diverge on an unverified uid.

Mitigating the exposure in the meantime: `mktemp -d` creates the directory `0700`, so the copy is private up to the `chmod`, the runner is ephemeral and single-tenant, and the credentials no longer outlive the reap. If the mode is worth tightening it should be tightened in all three places at once, against a confirmed uid — a better fit for its own change than for this PR. Added a comment at the call site recording exactly that.

.github/scripts/uat-janitor.sh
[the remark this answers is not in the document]
njhensley: Fixed — author authorized the tightening (it's the narrowing direction).

`classify` now requires a full anchored match against the three deployment-id schemas the project actually generates:

```
^aicr-uat-[0-9]+$                 AWS/Azure nightly
^aicr-[0-9]+$                     GCP nightly (#2064/#2070 — GCP node-SA account_id caps at 30 chars)
^aicr-uat-day-[a-z0-9-]+-[0-9]+$  daytime, all clouds
```

`aicr-uat-unrelated-7` now returns `SKIP:not-allowlisted`, as do `aicr-uat-`, `aicr-testgrid-vpc`, `aicr-uat-day--7` (empty reservation), `aicr-uat-day-h100-` (empty run id), `aicr-uat-123-vpc`, and `some-aicr-uat-123` — 13 rejection cases plus 4 acceptance cases, including a hyphenated reservation (`aicr-uat-day-gcp-h100-<run_id>`; real names are `aws-h100`, `gcp-h100`, `azure-h100`, `aws-gb200`, `kind-h100`).

`discover_*` stays coarser by design — it only nominates candidates; `classify` is the authoritative gate. `run_id_of`/`is_daytime` unchanged as defense in depth. Header documents all three schemas.

tools/uat-janitor_test.sh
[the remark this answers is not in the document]
njhensley: Fixed — author authorized the tightening (it's the narrowing direction).

`classify` now requires a full anchored match against the three deployment-id schemas the project actually generates:

```
^aicr-uat-[0-9]+$                 AWS/Azure nightly
^aicr-[0-9]+$                     GCP nightly (#2064/#2070 — GCP node-SA account_id caps at 30 chars)
^aicr-uat-day-[a-z0-9-]+-[0-9]+$  daytime, all clouds
```

`aicr-uat-unrelated-7` now returns `SKIP:not-allowlisted`, as do `aicr-uat-`, `aicr-testgrid-vpc`, `aicr-uat-day--7` (empty reservation), `aicr-uat-day-h100-` (empty run id), `aicr-uat-123-vpc`, and `some-aicr-uat-123` — 13 rejection cases plus 4 acceptance cases, including a hyphenated reservation (`aicr-uat-day-gcp-h100-<run_id>`; real names are `aws-h100`, `gcp-h100`, `azure-h100`, `aws-gb200`, `kind-h100`).

`discover_*` stays coarser by design — it only nominates candidates; `classify` is the authoritative gate. `run_id_of`/`is_daytime` unchanged as defense in depth. Header documents all three schemas.

.github/workflows/uat-janitor.yaml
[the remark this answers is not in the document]
njhensley: Author decision: **document, don't build a cross-workflow mutex.** No behavior change; rationale is now a comment above the `concurrency` block.

The exclusion a per-reservation mutex would buy is already provided by the design:

- Janitor candidates are **run-scoped unique deployment ids**. A live lifecycle run reports `queued`/`in_progress` on the GitHub API, so `classify` skips it outright — its cluster is never a candidate.
- A lifecycle run touches only its **own** run-scoped name and Terraform state prefix, so it shares no resource with the older orphan the janitor targets. There's no collision to serialize.
- Actuator destroys are idempotent, and GCP/Azure additionally serialize on the Terraform state lock.

With `DRY_RUN` defaulting true and the nightly>=6h / daytime>=24h floors on top, a mutex spanning both workflows isn't warranted for a dry-run backstop. Tracked as a possible future enhancement if enforce mode ever shows contention in practice.

.github/scripts/uat-janitor.sh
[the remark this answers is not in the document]
njhensley: Addressed — cleanup is now armed *before* any credential is written, and it survives a `set -e` abort.

I did **not** use `trap 'rm -rf "$az_mount"' RETURN`, because I tested it and it breaks the janitor. A RETURN trap set inside a function is not cleared when that function returns, so it fires again on the *caller's* return with the variable out of scope:

```
inner body
TRAP fired, m=/var/folders/.../tmp.5FMAYyn6UJ
outer after inner
TRAP fired, m=UNSET
line 8: m: parameter null or not set
exit=1
```

Under `set -e` that aborts the whole sweep mid-run. Instead `AZ_MOUNT` is module-level with a `cleanup_az_mount` helper, and `main()` installs `trap cleanup_az_mount EXIT` before any reap — so cleanup is armed earlier than a RETURN trap could manage. The trap is installed in `main()` rather than at module scope so sourcing the file (the unit harness) doesn't clobber the caller's own EXIT trap. Verified: creds are removed inline on the normal path, and by the EXIT trap when `set -e` kills the script between `mktemp -d` and the copy.

On the permissions half: `chmod -R a+rwX` stays, now with a one-line rationale at the call site. It mirrors the production Bringup/Destroy pattern in `uat-azure.yaml` — the actuator container runs as uid `builder`, which differs from the runner uid, and it writes to `~/.azure` on startup. `mktemp -d` is `0700` so the copy is private until the chmod, the runner is ephemeral and single-tenant, and the credentials no longer outlive the reap.

njhensley:
Aligned the janitor with **#2081 (ADR-017)**: the daytime deployment-id schema is now `aicr-uat-day-<slug>-<slot>-<run_id>` (was `…-<reservation>-<run_id>`).

- Tightened the classify allowlist regex `^aicr-uat-day-[a-z0-9-]+-[0-9]+$` → `^aicr-uat-day-[a-z][a-z0-9]{1,3}-[0-9]+-[0-9]+$` (slug = `^[a-z][a-z0-9]{1,3}$`, slot = digits, run_id = digits). It deliberately **no longer matches** the pre-ADR-017 hyphenated `<reservation>` form (e.g. `aicr-uat-day-gcp-h100-<id>`), which is now a regression-tested reject.
- Updated header/discovery comments to the ADR-017 shape; nightly schemas (`aicr-uat-<run_id>`, GCP `aicr-<run_id>` per #2064/#2070) and discovery are unchanged.
- Test updated: daytime cases use `aicr-uat-day-gh1-0-<id>`, added slug `ag2` + multi-digit-slot acceptance and old-reservation-form + malformed-slug rejections. `tools/uat-janitor_test.sh` passes; shellcheck clean.

Transitional note: while the janitor is dry-run/draft it won't enforce; by the time it does, #2081 will have landed and daytime names will be the new form. Any pre-#2081 daytime orphan (old reservation form) is a one-time manual cleanup.

njhensley:
Follow-up: #2070 was closed and the GCP node-SA cap was fixed **upstream in the actuator** (mchmarny/cluster#36; main now pins GKE v0.5.17). So GCP nightly stays `aicr-uat-<run_id>` like the other clouds — the `aicr-<run_id>` short form no longer exists. Realigned the janitor:

- Allowlist back to **two** schemas: `^aicr-uat-[0-9]+$` (nightly, all clouds) + `^aicr-uat-day-[a-z][a-z0-9]{1,3}-[0-9]+-[0-9]+$` (daytime, ADR-017). Dropped `^aicr-[0-9]+$`; `aicr-<digits>` is now a regression-tested reject.
- `discover_gcp` reverted to `^aicr-uat-` (matches AWS/Azure).
- Bumped the janitor's GKE actuator pin v0.5.16 → **v0.5.17** to match main.

Test passes; shellcheck + yamllint + actionlint clean.

njhensley:
Applied fixes from a multi-reviewer cross-review (Claude + Codex + integration; CodeRabbit lane was unavailable):

- **[major] Daytime allowlist matched nothing real.** The regex had been tightened to the ADR-017 `<slug>-<slot>` form, but #2081 (ADR-017) is not merged — main still emits `aicr-uat-day-<reservation>-<run_id>` (e.g. `aicr-uat-day-gcp-h100-<id>`), so every real daytime orphan was `SKIP:not-allowlisted` and the daytime backstop reaped nothing. Widened to `^aicr-uat-day-[a-z0-9-]+-[0-9]+$`, which matches **both** the current reservation form **and** the ADR-017 slug/slot form during the transition (tighten once #2081 lands). Tests now assert both forms are accepted; the reservation form drives the age-floor cases.
- **[medium] AWS discovery couldn't recover a cluster-gone/VPC-remains orphan.** Added `aws ec2 describe-vpcs` Name-tag recovery (`<id>-vpc`, mirroring the GCP `-vpc` / Azure `-rg` paths) so AWS is no longer the one path that misses network-only orphans.
- **[minor] Discovery CLI failures were swallowed as 'nothing to reap'.** New `run_discovery` helper emits a `::warning::` on any non-zero list (missing CLI, expired creds, a per-API authorization gap the auth step didn't catch, throttling) instead of reporting a clean empty fleet. Still best-effort.
- **[major, contested] No lane for cross-account `aws-gb200`.** Added an `aws-gb200` job assuming the role in tenancy 897722687756 with `cluster-config-gb200.yaml` (documented OIDC-trust prerequisite).
- **[medium, contested] daytime-down liveness race.** Documented in the concurrency rationale: a daytime name encodes the daytime-**up** run id so the liveness gate can't see an active daytime-**down**; the 24h floor covers it (a cluster old enough for the janitor is one daytime-down already failed to reap), and break_state_lock then clears only an unheld lock.

shellcheck + yamllint + actionlint clean; `tools/uat-janitor_test.sh` passes. Still a dry-run draft.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,440 @@
...
+
+# Destroy one orphaned deployment via the actuator (destroy == apply w/ destroy=true).
+# Returns the actuator's exit status so the caller counts only destroys that
+# actually happened — the rollout asks operators to trust these logs before
+# enabling enforce, so a reported reap must be a real one.
+reap() {
+  local id="$1" cfg rc=0
+  cfg="$(mktemp --suffix=.yaml)"
+  yq '.' "$JANITOR_CONFIG" >"$cfg"
+  DEPLOY_ID="$id" yq -i '.deployment.id = strenv(DEPLOY_ID) | .deployment.destroy = true' "$cfg"
```
yuanchen8911: reap() is invoked as `if reap "$id"` (line 406), which disables errexit for the whole function body. If this yq rewrite fails (or the mktemp/copy above it), the function falls through to `docker run ... apply` with the unmodified committed config — deployment.id still the placeholder and destroy still false — so an enforce-mode reap can provision or update a cluster while logging "reaped". This reproduces: with the second yq forced to fail, docker receives destroy: false and the function reports success.

Please check the copy and the yq -i explicitly and return non-zero before docker.

.github/workflows/uat-janitor.yaml
```diff
@@ -0,0 +1,217 @@
...
+#   * Janitor candidates are RUN-SCOPED unique deployment ids. A live lifecycle
+#     run reports `queued`/`in_progress` on the GitHub API, so classify() skips
+#     it outright — its cluster is never a candidate in the first place.
+#   * A lifecycle run touches only its OWN run-scoped name and Terraform state
+#     prefix, so it shares no resource with whatever older orphan the janitor is
+#     targeting. There is no collision to serialize.
+#   * Actuator destroys are idempotent, and GCP/Azure additionally serialize on
+#     the Terraform state lock.
+#   * Daytime nuance: a daytime name encodes the daytime-UP run id, so the
+#     liveness gate cannot see an active daytime-DOWN teardown of the same cluster
```
yuanchen8911: This "Daytime nuance" argument doesn't hold. A cluster old enough for the janitor is claimed to be "one daytime-down already FAILED to reap, not one it is mid-way through" — but daytime-down fires every night at 02:00 UTC (uat-daytime.yaml) and resolves its target by reservation prefix (uat-gcp.yaml "Resolve daytime cluster"), so the next scheduled down retries the old orphan. The janitor runs hourly at :17; at ~02:17 in enforce mode it can classify that same cluster REAP (its daytime-up run is long completed), break the down run's live TF state lock via break_state_lock, and start a concurrent destroy. AWS has no state lock at all.

Manual enforce dispatches can overlap this today, and the planned schedule flip would make it routine — it needs either the mutex this comment argues away, or a liveness check that can see an active daytime-down.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,440 @@
...
+    rm -f "$errf"
+    echo "SKIP:gh-api-error(run ${run_id})"; return   # fail closed
+  fi
+
+  # Allowlist, not denylist: ONLY a confirmed `completed` run may be reaped. The
+  # active statuses are named separately because they are a routine skip an
+  # operator should not have to think about; anything else is a status GitHub did
+  # not have when this was written, and fails closed rather than falling through.
+  case "$status" in
+    completed) ;;
```
yuanchen8911, line 240: A cluster deliberately retained via skip_delete (a supported debugging input in uat-run.yaml; uat-gcp.yaml intentionally suppresses teardown when set) is `completed` + >6h here, so it classifies REAP. Enforcement is cloud-wide with no per-deployment exclusion, so an enforce=true dispatch deletes a cluster someone may be actively debugging — and while such a cluster exists, reaping the real orphans in that cloud isn't safe.

This needs an explicit policy decision rather than shipping the ambiguity silently: either document a hard six-hour retention limit for skip_delete clusters prominently, or add an exclusion/lease mechanism.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,440 @@
...
+  case "$CLOUD" in
+    gcp)
+      docker run --rm \
+        -e CONFIG_CONTENT="$(base64 <"$cfg" | tr -d '\n')" \
+        -e AUTO_APPROVE=true \
+        -e KEY_CONTENT="$(base64 <"$GOOGLE_GHA_CREDS_PATH" | tr -d '\n')" \
+        "$JANITOR_ACTUATOR_IMAGE" apply || rc=$?
+      ;;
+    aws)
+      docker run --rm \
```
yuanchen8911: The AWS reap goes straight to the actuator, but the normal teardown runs uat-aws-cleanup-lb.sh first (and sweeps between retries) because a LoadBalancer Service leaves a classic ELB + k8s-elb-* SG outside Terraform state, and destroy then fails at DeleteVpc with DependencyViolation. AWS inference UAT deploys agentgateway, which creates exactly such a Service — so the orphan class this janitor exists for fails to reap, identically, every cycle.

A complete fix needs the graceful cleanup while the EKS cluster still exists, plus the sweep/retry fallback for the case where only the VPC remains.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,440 @@
...
+      echo "SKIP:run-status-unsupported(${status},run ${run_id})"; return ;;
+  esac
+
+  if [ -n "$upd" ]; then
+    age="$(age_hours_since "$upd")" || { echo "SKIP:bad-updated-at(${upd})"; return; }
+  else
+    age=999999   # confirmed run-level 404: purged from history, past every floor
+  fi
+  if is_daytime "$id"; then min="$DAYTIME_MIN_AGE_HOURS"; else min="$NIGHTLY_MIN_AGE_HOURS"; fi
+  if [ "$age" -lt "$min" ]; then
```
yuanchen8911, line 253: If either age-floor variable is overridden with a non-integer (NIGHTLY_MIN_AGE_HOURS=24h) or a negative value (-1), the guard is bypassed: the non-integer makes `[ "$age" -lt "$min" ]` error inside an `if` (reads as false → REAP), and a negative floor passes every age. Both reproduce as REAP for a one-hour-old run.

The shipped workflow never sets these, but the script accepts them as configuration and already validates DRY_RUN against exactly this typo class — validate both floors as non-negative integers up front, alongside valid_dry_run.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,440 @@
...
+      ;;
+    azure)
+      loc="$(yq -r '.deployment.location' "$JANITOR_CONFIG")"
+      local rg="cluster-state-rg" sa key blob="deployments/${loc}/${id}/terraform.tfstate"
+      sa="$(az storage account list -g "$rg" --subscription "$AZURE_SUBSCRIPTION_ID" \
+        --query "[?starts_with(name, 'clst')].name | [0]" -o tsv 2>/dev/null || true)"
+      if [ -n "${sa:-}" ]; then
+        key="$(az storage account keys list -g "$rg" -n "$sa" \
+          --subscription "$AZURE_SUBSCRIPTION_ID" --query '[0].value' -o tsv 2>/dev/null || true)"
+        az storage blob lease break --account-name "$sa" --account-key "$key" \
```
yuanchen8911: --account-key puts the full-access tfstate account key in argv, world-readable via /proc/<pid>/cmdline. The existing teardown in uat-azure.yaml passes the same key through AZURE_STORAGE_KEY specifically to avoid this (its comment spells out the argv exposure). Please match that pattern here.

.github/workflows/uat-janitor.yaml
```diff
@@ -0,0 +1,217 @@
...
+  cancel-in-progress: false
+
+env:
+  # Scheduled runs are always dry-run; only an explicit enforce=true dispatch
+  # deletes. inputs.enforce is null for the schedule event, so this is 'true' there.
+  DRY_RUN: ${{ inputs.enforce == true && 'false' || 'true' }}
+
+jobs:
+  gcp:
+    if: >-
```
yuanchen8911, line 102: The job gates check repository and cloud but not github.ref, while the header comment (line 33) claims "scheduled + dispatch runs execute on main". GitHub lets a dispatch select any branch, and the GCP/Azure trusts also admit test/uat, so a non-main dispatch runs that branch's copy of the janitor with delete authority. Requires write access, so hardening rather than an exploit — either add `github.ref == 'refs/heads/main'` to the job conditions or correct the comment.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,440 @@
...
+#     counts "would-reap", never "reaped" — a report must not claim a destroy
+#     that did not happen.
+#
+# Usage:   DRY_RUN=true ./uat-janitor.sh <gcp|aws|azure>
+# Requires: gh (GH_TOKEN in env, actions:read), jq, yq, docker, and the cloud CLI
+#           + credentials for <cloud> already configured by the caller.
+
+set -euo pipefail
+
+CLOUD="${1:?usage: uat-janitor.sh <gcp|aws|azure>}"
```
yuanchen8911, line 100: Operator-facing nit: CLOUD is never validated against gcp|aws|azure; a typo reaches main() and dies with `discover_<x>: command not found` instead of the usage message. A case guard next to the existing `${1:?usage...}` would keep manual invocation friendly. (It can't widen deletion scope — fails closed.)

tools/uat-janitor_test.sh
```diff
@@ -0,0 +1,191 @@
...
+GH_STUB=404
+REPO_REACHABLE=no
+check "skips 404 when repo unproven" "SKIP:gh-api-error(run 7)" "$(classify aicr-uat-7)"
+REPO_REACHABLE=yes
+check "reaps 404 when repo proven" "REAP" "$(classify aicr-uat-7)"
+
+echo "classify (age floors):"
+REPO_REACHABLE=yes
+GH_STUB="ok:completed:$(ago 2)"
+check "nightly under floor" "SKIP:too-young(2h < 6h)" "$(classify aicr-uat-7)"
```
yuanchen8911, line 238: These age-floor assertions expect the default 6h/24h literals, but the harness never pins NIGHTLY_MIN_AGE_HOURS / DAYTIME_MIN_AGE_HOURS before sourcing the janitor — an environment that exports either makes the hermetic test fail (this reproduces). Set both explicitly next to the GH_STUB/REPO_REACHABLE exports.

tools/uat-janitor_test.sh
```diff
@@ -0,0 +1,191 @@
...
+echo "classify (fail closed on ambiguity):"
+GH_STUB=500
+check "skips on API error" "SKIP:gh-api-error(run 7)" "$(classify aicr-uat-7)"
+GH_STUB=nofield
+check "skips on incomplete run JSON" "SKIP:incomplete-run-json(run 7)" "$(classify aicr-uat-7)"
+GH_STUB="ok:completed:not-a-timestamp"
+check "skips on unparsable updated_at" "SKIP:bad-updated-at(not-a-timestamp)" "$(classify aicr-uat-7)"
+
+echo "classify (404 requires a reachable repo):"
+GH_STUB=404
```
yuanchen8911, line 219: The 404 stubs only exercise the exact "gh: Not Found (HTTP 404)" string. Nothing pins that an error merely containing "404" (without the HTTP 404 token) stays SKIP:gh-api-error, so a future loosening of the grep would pass this suite while widening the purged-run reap path. One extra stub case would lock the narrowness the header advertises.

yuanchen8911 (commented):
The classification logic and top-level deletion gate are solid — anchored allowlist regexes, fail-closed API handling, typo-safe DRY_RUN — and the naming and state-backend layout assumptions check out against the existing uat-* workflows. CI is green and the harness runs via make test.

The inline comments cover five enforce-path fixes (reap fail-open, daytime teardown race, AWS LB cleanup, age-floor validation, Azure key exposure) plus a skip_delete policy decision that should land before enforcement is trusted, and four lower-priority items (ref gating, CLOUD validation, test env pinning, a 404 negative test).

- The PR body and squashed commit message are stale — the commit message describes the dropped GCP short schema (`^aicr-[0-9]+$`) as shipped, the body says cross-account GB200 isn't reconciled while the workflow adds an aws-gb200 job, and "15/15" understates the ~55-assertion harness. Please refresh both before the next push.

- Please document the janitor (dry-run/enforce semantics, age floors) in docs/contributor/uat.md in this PR.

Minor, no action required: the 60-minute serial-destroy budget means a multi-orphan backlog drains across several dispatches (bounded progress — worth a header note), and the actuator-image pins are now hand-synced across four workflows (a parity test would guard future bumps).

.github/scripts/uat-janitor.sh
njhensley: Fixed. `reap()` now checks `mktemp` and both `yq` steps explicitly and returns non-zero **before** the actuator call, so a staging failure can no longer fall through to `docker … apply` with the committed placeholder `deployment.id` + `destroy=false`. Added a comment noting the `if reap "$id"` errexit-disable is exactly why the explicit checks (rather than relying on `set -e`) are required.

.github/workflows/uat-janitor.yaml
njhensley: Agreed — the "daytime nuance" argument was wrong, so I replaced it with an actual guard rather than re-arguing it. `classify()` now calls `daytime_run_active()` and defers (`SKIP:daytime-lifecycle-active`) whenever any `uat-daytime` lifecycle run is in flight, so an enforce reap never races a live evening daytime-down — breaking a Terraform lock the down still holds (GCP/Azure), or a concurrent destroy on AWS (no state lock). It fails safe: any API error or unparsable body reads as "active" (defer). The concurrency comment is rewritten to describe the guard, and both `daytime_run_active` and the `classify()` deferral are unit-tested (including the fail-safe path). A cross-workflow mutex is left as a tracked future option if enforce ever shows contention.

.github/scripts/uat-janitor.sh
njhensley: Addressed by raising the nightly floor 6h → 24h (default), so a `skip_delete` hold survives a full working day before reclamation. As you note, a past run's dispatch inputs aren't retrievable via the API, so there's no signal that distinguishes a `skip_delete` hold from a genuine orphan — the floor is the only lever, and I documented that tradeoff prominently in the AGE FLOOR header (including that a genuine mid-cancel orphan therefore also persists ~a day, which is acceptable since quota pressure builds over days and the hourly janitor reaps within the hour once the floor clears).

I went with the time-bounded grace over a tag/lease to keep the change self-contained. If you'd prefer the hard guarantee, I'm happy to add an explicit retain-tag the UAT run stamps and the janitor honors — just a larger change touching `uat-{aws,azure,gcp}.yaml`.

.github/scripts/uat-janitor.sh
njhensley: Fixed. The AWS reap now runs `uat-aws-cleanup-lb.sh graceful "$id"` before the actuator destroy, and on failure runs `sweep "$id"` + retries the destroy once — mirroring the in-run teardown, so both the cluster-still-exists (graceful) and VPC-only-remnant (sweep) cases are covered. The actuator call is factored into `aws_actuator_destroy` for the retry, and the helper is resolved via a new `SCRIPT_DIR` so it works regardless of CWD.

.github/scripts/uat-janitor.sh
njhensley: Fixed. Added `valid_age_floor` (`*[!0-9]*` rejects empty, non-numeric, and negative alike) and validate both `NIGHTLY_MIN_AGE_HOURS` and `DAYTIME_MIN_AGE_HOURS` in `main()` up front, right alongside `valid_dry_run` — a bad floor now aborts with a FATAL usage message instead of erroring inside the `if` and reading as REAP.

.github/scripts/uat-janitor.sh
njhensley: Fixed. The lease-break now passes the key via `AZURE_STORAGE_ACCOUNT`/`AZURE_STORAGE_KEY` env and drops `--account-key`, matching the `uat-azure.yaml` teardown so the full-access tfstate key never appears in argv (`/proc/<pid>/cmdline`).

.github/workflows/uat-janitor.yaml
njhensley: Fixed. Added `github.ref == 'refs/heads/main'` to all four job `if:` conditions, so a `workflow_dispatch` that selects `test/uat` (or any non-main branch) is filtered out and can't run that branch's janitor with delete authority. Updated the header comment to state the gate *enforces* main-only execution rather than merely asserting it.

.github/scripts/uat-janitor.sh
njhensley: Done. `main()` now validates `CLOUD` against `gcp|aws|azure` and aborts with the usage message on anything else, before any discovery. As you noted it can only narrow scope (fails closed), but it keeps manual invocation friendly instead of dying with `discover_<x>: command not found`.

tools/uat-janitor_test.sh
njhensley: Fixed. The harness now `export`s `NIGHTLY_MIN_AGE_HOURS=6` / `DAYTIME_MIN_AGE_HOURS=24` before sourcing the janitor, so a host or CI shell that exports either can't break the hermetic run. The assertions keep the 6h/24h literals — now pinned rather than relying on defaults, which also decouples them from the nightly default moving to 24h.

tools/uat-janitor_test.sh
njhensley: Fixed. Added a `404plain` stub case (`gh: gateway error code 404 upstream` — contains `404` but not the `HTTP 404` token) and an assertion that it stays `SKIP:gh-api-error` **even with `REPO_REACHABLE=yes`**, locking the narrowness so a future loosening of `classify()`'s grep to a bare `404` would fail the suite.

.github/scripts/uat-janitor.sh
yuanchen8911: Verified fixed. mktemp and both yq steps are checked with an explicit return before the actuator call, and the errexit-suppression comment documents the trap for future editors. A forced staging failure returns non-zero with zero docker invocations.

.github/workflows/uat-janitor.yaml
yuanchen8911: The guard is real for the normal scheduled path — the daytime scheduler's drive job watches the dispatched run to completion, so daytime_run_active() sees it — but four gaps remain, and the updated comment now claims "the janitor never races a live daytime-DOWN destroy," which the code doesn't deliver:

1. The direct-dispatch path is invisible: docs/contributor/uat.md tells operators to run `gh workflow run uat-run.yaml ... -f lifecycle=daytime-down` for a single reservation. That never creates a uat-daytime run, so the guard reports idle during that live destroy.

2. The watcher can die before the child: drive has timeout-minutes: 200 while the dispatched uat-run.yaml allows 360 (uat-gcp.yaml itself 300). A slow or cancelled teardown outlives the watched scheduler run and the guard goes idle mid-destroy.

3. The fail-safe contract doesn't hold on malformed 200s: through `[.workflow_runs[]? ...] | length`, both `{"message":"partial response"}` and `{"workflow_runs":null}` return 0 — read as idle, not "defer" as the function comment promises. The `[]?` swallows the missing field; check the field exists before counting, else return active.

4. A liveness probe is not mutual exclusion — a down can start right after the check, during a 10–25 min destroy.

Suggest either a shared lease/mutex if the "never races" guarantee is wanted, or: widen the guard to defer daytime reaps while any uat-run.yaml run is in flight (coarse, one-cycle cost), fix the jq to fail closed, and soften the comment to state the remaining TOCTOU window plainly.

.github/scripts/uat-janitor.sh
yuanchen8911: The 24h floor is a reasonable policy choice and the header rationale is good, but the ask was a prominently documented retention limit and this isn't visible where operators will look: the skip_delete input description in uat-run.yaml and docs/contributor/uat.md say nothing about automatic reclamation, and the PR body still advertises 6h. Please add one line to the input description ("cluster is reclaimed by the janitor after ~24h") and cover it in the uat.md janitor section.

.github/scripts/uat-janitor.sh
yuanchen8911: Verified fixed. graceful runs before the destroy, a failed destroy triggers sweep plus exactly one retry (rc correctly reset), both helper modes are scoped to the deployment id's ownership tag and always exit 0, AWS_REGION is set in both AWS job envs, and SCRIPT_DIR makes the sibling call CWD-independent.

.github/scripts/uat-janitor.sh
yuanchen8911: Better, but one fail-open range case remains — reproduced: valid_age_floor accepts any digit string, and with NIGHTLY_MIN_AGE_HOURS=99999999999999999999 bash's test builtin errors ("integer expected", status 2) inside the `if` and falls through to REAP for a one-hour-old run — the same inversion this fix closed for "24h" and "-1", one input class over. Bound the accepted range (e.g. reject floors longer than 5 digits) and add the regression case to the harness.

.github/scripts/uat-janitor.sh
yuanchen8911: Verified fixed. The lease break passes the key via AZURE_STORAGE_ACCOUNT/AZURE_STORAGE_KEY and --account-key is gone, matching the uat-azure.yaml teardown pattern.

.github/workflows/uat-janitor.yaml
yuanchen8911: Verified fixed. github.ref == 'refs/heads/main' is on all four job conditions and the header comment now describes the gate as enforced rather than assumed.

.github/scripts/uat-janitor.sh
yuanchen8911: Verified fixed. Unknown clouds abort with the usage message in main() before any discovery.

tools/uat-janitor_test.sh
yuanchen8911: The isolation fix is right, but the pin is now stale against the shipped default: the harness exports NIGHTLY_MIN_AGE_HOURS=6 while production defaults to 24, so the new retention policy is never asserted and a regression flipping the default back to 6 would pass the suite. Pin 24 and adjust the two nightly-floor assertions (or keep 6 for the boundary cases and add one assertion on the sourced default).

tools/uat-janitor_test.sh
yuanchen8911: Verified fixed. The 404plain stub asserts a bare-404 error stays SKIP:gh-api-error even with REPO_REACHABLE=yes — exactly the narrowness guard asked for.

yuanchen8911:
Round-2 verification: the config-staging, AWS LB, Azure key, ref-gating, CLOUD validation, hermetic-env, and 404-narrowness fixes all check out (those threads are resolved), and the squashed commit message is corrected and validly signed. Four residual items are on their threads: the daytime guard's remaining gaps (direct uat-run.yaml dispatch, watcher-vs-child timeout, jq fail-open on malformed 200s, and the overclaiming "never races" comment), the age-floor integer-range case, skip_delete retention visibility, and the stale 6h test pin.

Still open from round 1: the PR body (nightly floor is now 24h not 6h, "15/15" understates the 62-assertion harness, and the v1-scope paragraph still says cross-account GB200 isn't reconciled while the aws-gb200 job covers it), and the docs/contributor/uat.md janitor section.

.github/workflows/uat-janitor.yaml
njhensley: Good analysis — all four gaps are real, and you're right the comment overclaimed. Took the refined liveness path you outlined rather than a mutex:

1 & 2 (invisible direct-dispatch / watcher dying before child): the guard now watches **`uat-run.yaml`** in-flight runs, not the `uat-daytime` scheduler wrapper. The teardown always executes as a `uat-run.yaml` run — whether the evening scheduler dispatched it as its own top-level run, or an operator ran `gh workflow run uat-run.yaml -f lifecycle=daytime-down` directly — so both paths are now visible and we no longer depend on the drive job's `timeout-minutes`.

3 (jq fails open on malformed 200s): rewritten to `if (.workflow_runs | type) == "array" then … else "malformed" end` — a missing field, `{"workflow_runs":null}`, or `{"message":…}` now emits a non-numeric marker that the `case` reads as **active** (defer). Added `malformed` and `nullruns` stub cases + assertions to lock it.

4 (TOCTOU — probe ≠ mutex): comment softened to state the residual window plainly — a down that *starts* in the ~1h between check and reap — backstopped by the 24h floor + DRY_RUN default. A shared cross-workflow lease is noted as the tracked follow-up if a hard "never races" guarantee is ever needed. I did not add the lease now, consistent with keeping this a dry-run-first backstop; happy to if you'd prefer it before the enforce flip.

.github/scripts/uat-janitor.sh
njhensley: Documented. Added to the `skip_delete` input description in `uat-run.yaml` ("The UAT janitor reclaims a retained cluster ~24h after the run finishes.") and a new **orphan janitor** section in `docs/contributor/uat.md` with a "Retention limit for `skip_delete`" note (one-working-day loan; re-provision if you need it longer). Also corrected the PR body, which still said 6h.

.github/scripts/uat-janitor.sh
njhensley: Fixed. `valid_age_floor` now also rejects anything longer than 5 digits (`[ "${#1}" -le 5 ]`) — `99999h` (~11y) is far beyond any real floor and well under the ~2^63 point where `[ "$age" -lt "$min" ]` errors with status 2 and falls through to REAP. Added a `valid_age_floor` block to the harness including the 20-digit overflow case and the `100000` (6-digit) boundary.

tools/uat-janitor_test.sh
njhensley: Fixed via your second option. The harness now `unset`s both floors before sourcing, so the **shipped defaults** are what's exercised — with an explicit assertion that they're 24h (`nightly default is 24h` / `daytime default is 24h`), so a regression flipping the default back to 6 now fails the suite. The two boundary cases keep their explicit 6h/24h thresholds via a plain local assignment (`classify()` reads the global at call time), so the arithmetic is still exercised at a known threshold, decoupled from the default. Hermetic and regression-catching both.

.github/workflows/uat-janitor.yaml
yuanchen8911: Verified fixed. The probe now watches uat-run.yaml, which covers both the direct-dispatch path and the scheduler-watcher timeout (the teardown run itself is what's observed); the jq requires an actual workflow_runs array and reads both malformed-200 shapes as active/defer (verified); and the comment now states probe-not-mutex with the residual TOCTOU window plainly and identifies the cross-workflow lease as the follow-up. The residual is non-blocking as scoped.

.github/scripts/uat-janitor.sh
yuanchen8911: Verified fixed. The retention limit is now on the skip_delete input description and in a dedicated uat.md section — visible where operators will look. One precision nit on the wording noted separately (non-blocking).

.github/scripts/uat-janitor.sh
yuanchen8911: Verified fixed. The 5-digit length cap closes the overflow fall-through (length test, so the validation itself can't overflow), and the regression cases — including the 20-digit reproducer — are in the harness.

tools/uat-janitor_test.sh
yuanchen8911: Verified fixed. The harness unsets both floors before sourcing and asserts the shipped 24h defaults directly, with the boundary cases on an explicit local floor — a default regression now fails the suite.

.github/workflows/uat-janitor.yaml
```diff
@@ -0,0 +1,232 @@
...
+#   * Actuator destroys are idempotent, and GCP/Azure additionally serialize on
+#     the Terraform state lock.
+#   * Daytime last-mile: a daytime name encodes the daytime-UP run id, so the
+#     run-status liveness gate cannot see an active daytime-DOWN teardown of the
+#     same cluster (a different run). The teardown always executes as a
+#     uat-run.yaml run — the evening scheduler dispatches uat-run.yaml as its own
+#     top-level run, and a direct `gh workflow run uat-run.yaml -f
+#     lifecycle=daytime-down` is a uat-run.yaml run too — so classify() calls
+#     uat_lifecycle_active() and, if ANY uat-run run is in flight, defers
+#     (SKIP:lifecycle-active) to the next hourly cycle. This is a liveness PROBE,
```
yuanchen8911, line 80: Non-blocking: the probe closes the already-active-teardown race, but a daytime-down that starts between the check and the reap can still collide with an enforce-mode destroy. This comment documents that honestly and identifies a cross-workflow lease as the follow-up — fine as scoped; consider filing the follow-up issue so it doesn't live only in a comment.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,567 @@
...
+# The 24h floor and DRY_RUN-default backstop that window; a shared cross-workflow
+# lease is the tracked follow-up if a hard "never races" guarantee is ever needed.
+#
+# Fails CLOSED: any API error, or a 200 whose body is not a well-formed run list
+# (`{"message":...}`, `{"workflow_runs":null}`), reads as "active" (defer) rather
+# than idle. Uses `gh` raw + external `jq` (like classify()) so the harness drives it.
+LIFECYCLE_WORKFLOW="uat-run.yaml"
+uat_lifecycle_active() {
+  local json active
+  json="$(gh api "repos/${REPO}/actions/workflows/${LIFECYCLE_WORKFLOW}/runs?per_page=30" 2>/dev/null)" || return 0
```
yuanchen8911: Non-blocking hardening: this reads the newest 30 mixed-status runs and filters client-side, so an active run buried under an abnormal dispatch burst could be missed. At current volume (~19 runs/day) 30 covers ~27h, well beyond a 300–360-minute execution — querying the active statuses directly (?status=in_progress etc.) would remove the volume assumption entirely.

docs/contributor/uat.md
```diff
@@ -146,6 +146,12 @@ You never *have* to tear a cluster down by hand — the evening safety-net cron
...
 
+### The orphan janitor (backstop teardown)
+
+`uat-janitor.yaml` (hourly, all three clouds) is the last-resort backstop for whatever the in-run teardown misses — a runner hard-killed mid-cancel, a daytime hold whose evening `daytime-down` never fired, an abandoned `skip_delete` run, or a Bringup false-failure that stranded a healthy cluster. It reconciles every `aicr-uat-<run_id>` / `aicr-uat-day-<reservation>-<run_id>` deployment against its owning GitHub run and reaps only the ones whose run is finished and past an age floor, via the same actuator `destroy` the in-run teardown uses. It is **dry-run by default**: scheduled runs only report; an actual reap requires a deliberate `workflow_dispatch` with `enforce=true`. See `.github/scripts/uat-janitor.sh` for the full safety model.
+
+**Retention limit for `skip_delete`.** A cluster kept alive with `skip_delete=true` (manual debugging) is reclaimed by the janitor **~24h after its run finishes** — there is no API signal that distinguishes an intentional hold from a genuine orphan, so the age floor is the only lever. Treat a `skip_delete` cluster as a one-working-day loan; if you need it longer, re-provision.
```
yuanchen8911: Non-blocking precision nit: "is reclaimed by the janitor ~24h after its run finishes" overstates the shipped behavior — scheduled runs are dry-run, so at 24h the cluster becomes eligible; reclamation happens on an enforce dispatch (automatic only once scheduled enforcement is enabled). The paragraph above gives the context; "becomes eligible for enforced reclamation" would be exact.

.github/scripts/uat-janitor.sh
```diff
@@ -0,0 +1,567 @@
...
+
+# One EKS actuator destroy pass for the staged config <cfg>. Factored out so the
+# AWS reap can run it, sweep leaked LB resources, and run it again.
+aws_actuator_destroy() {
+  docker run --rm \
+    -e CONFIG_CONTENT="$(base64 <"$1" | tr -d '\n')" \
+    -e AUTO_APPROVE=true \
+    -e AWS_ACCESS_KEY_ID="$AWS_ACCESS_KEY_ID" \
+    -e AWS_SECRET_ACCESS_KEY="$AWS_SECRET_ACCESS_KEY" \
+    -e AWS_SESSION_TOKEN="$AWS_SESSION_TOKEN" \
```
yuanchen8911: Non-blocking manual-path nit: this expands "$AWS_SESSION_TOKEN" under set -u, so running the script by hand with static AWS keys (no session token) aborts. The OIDC jobs always provide one, so CI is unaffected — "${AWS_SESSION_TOKEN:-}" fixes the advertised manual path.

yuanchen8911 (commented):
All four residual items are fixed as described — the lifecycle probe now watches uat-run.yaml (covering direct dispatches and the scheduler-watcher timeout), the jq fails closed on malformed responses, the age-floor length cap closes the overflow case, the skip_delete retention is documented on the input and in uat.md, and the harness asserts the shipped 24h defaults. No blocking code findings remain; four non-blocking notes are inline.

Before merge (hygiene, not code): the branch is behind main (GitHub reports mergeStateStatus BEHIND, and repo policy requires rebasing before push), so a rebase + CI rerun is needed; and the PR body has drifted from the code — the daytime scheduler is no longer dormant (it fired Aug 6 and 7), AWS discovery has included tag-based VPC recovery since the first commit, the "Draft until…" note and "15/15" test count are stale, and the v1-scope paragraph still says cross-account GB200 isn't reconciled while the aws-gb200 job covers it.

.github/workflows/uat-janitor.yaml
njhensley: Filed as #2132 (Enhancement, `theme/ci-dx`/`area/ci`) — the cross-workflow lease follow-up, to revisit before scheduled enforcement is enabled. The `concurrency` comment here (and the `uat_lifecycle_active` comment) now point at it, so it no longer lives only in a comment.

.github/scripts/uat-janitor.sh
njhensley: Done — switched to a server-side `?status=in_progress&per_page=100` filter, so the newest-N window (and the ~19-runs/day volume assumption) is gone: the executing-teardown run is matched directly regardless of dispatch burst. A still-queued/about-to-start run is the documented TOCTOU residual (backstopped by the 24h floor + DRY_RUN default). Kept the fail-closed array guard; the stub's idle case now returns an empty array to match the filtered response.

docs/contributor/uat.md
njhensley: Fixed. `uat.md` now reads "**becomes eligible for enforced reclamation ~24h after its run finishes**" with a parenthetical that eligibility only becomes a reclaim on an `enforce=true` dispatch (or automatically once scheduled enforcement is on). Applied the same precision to the `skip_delete` input description ("becomes eligible for janitor reclamation").

.github/scripts/uat-janitor.sh
njhensley: Fixed — `-e AWS_SESSION_TOKEN="${AWS_SESSION_TOKEN:-}"`, so a manual run with static keys (no session token) no longer aborts under `set -u`. The OIDC jobs still pass the real token; access-key/secret stay unguarded since they're mandatory.

yuanchen8911 (approved):

Issue #1460, opened 2026-06-24
ci(supply-chain): adopt sigstore/rekor-monitor for release identity + consistency


lockwobr:
## Problem

AICR signs releases and logs them to Rekor, but nothing watches the log for entries under AICR's release identity (potential OIDC/key compromise) or verifies the log stays append-only.

## Proposed approach

Add `.github/workflows/rekor-monitor.yaml` calling the `sigstore/rekor-monitor` reusable workflow on an hourly cron:

- Identity monitoring of issuer `https://token.actions.githubusercontent.com` and cert-identity regexp `https://github.com/NVIDIA/aicr/.github/workflows/on-tag.yaml@refs/tags/.*`.
- Consistency monitoring (checkpoint stored as a workflow artifact).
- `file_issue: true`; `artifact_retention_days` well above the hourly interval.

Zero Go. Pin the action to a released tag. This protects the shared trust root every AICR user consumes.

Part of epic #1149.

PR #2069, opened 2026-08-05
ci(supply-chain): daily re-verification of the latest release

### LINKED ISSUE #1461: ci(supply-chain): daily re-verification of latest release Rekor entries
## Problem

An artifact AICR released could lack a live, retrievable Rekor entry (a signing-side upload that silently failed or was skipped). This is distinct from log soundness and is not covered by rekor-monitor.

## Proposed approach

A daily scheduled workflow that resolves the latest release, downloads its artifacts and bundles/signatures, and runs the shipped verify commands against live public-good Rekor (`aicr` bundle attestation verification, `aicr recipe verify-catalog`, `cosign verify-blob-attestation`). A failure (entry gone) files an issue. Near-zero new code.

Part of epic #1149.

lockwobr:
## Summary

Adds a daily scheduled workflow that re-runs AICR's shipped verification commands against the latest release, so an artifact that shipped without a live, retrievable Rekor entry is detected rather than discovered by a user.

## Motivation / Context

A signing-side upload can silently fail or be skipped, leaving a released artifact whose attestation cannot be verified. That is distinct from transparency-log soundness and is not covered by `rekor-monitor`, which watches the log rather than our published assets.

Fixes: #1461
Related: #1149, #1460, #1897

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: CI (`.github/workflows/`, `tests/releasepolicy`)

## Implementation Notes

**Almost entirely assembly, per the issue's "near-zero new code".** `.github/actions/install-aicr-release` already downloads the release archive, checks it against `aicr_checksums.txt`, extracts `aicr-attestation.sigstore.json`, and runs `cosign verify-blob-attestation` pinned to `on-tag.yaml@refs/tags/<exact tag>` with retries, failing closed when the bundle is missing. It is reused directly and leaves a verified `./aicr` that the catalog check then uses. Release resolution reuses `.github/scripts/gh-api-retry.sh`, the cosign pin comes from `load-versions`, and the issue-filing, de-duplication, streak and auto-close steps are taken from `rekor-monitor.yaml`. The only genuinely new logic is the classifier.

**What is verified:** the binary provenance bundle; `recipe-catalog.sigstore.json` via `aicr recipe verify-catalog`, run with the binary just verified and against the loose release asset (the copy `cli-reference.md` tells users to curl); and each `*_linux_amd64.sbom.json` against its `.sigstore.json` sibling.

**What is deliberately not verified**, with the reason recorded in the workflow header and in `maintaining.md`:
- Bundle attestations. No bundle is a release asset; bundles are produced on demand by `aicr bundle --attest`, so there is nothing released to check and no step was fabricated.
- `aicrd`. Confirmed not archived (`archives.ids: [aicr]`), so no `aicrd` tarball or attestation is a release asset. Its SBOM is, and is covered.
- Image OCI referrers from #1982. Different retention and GC model from GitHub Releases, and 7 images times 3 predicate kinds is roughly 21 registry round-trips per run, which would dilute the one signal this job keeps crisp. Images are already pulled weekly by `vuln-scan-images.yaml`. Recorded as a follow-up, with the requirement that any registry-side check must use `gh attestation verify --bundle-from-oci`, since the default form reads GitHub's attestations API and proves nothing about the registry copy.

**Infra noise must not page.** This is the central design constraint, and it reuses `rekor-monitor`'s vocabulary rather than inventing a parallel one: `clean`/0, `tamper`/1 (security), `operational`/3. The security gate is a strict allowlist (`== 'tamper'`) and the degraded gate its exact inverse, so an empty classification (any failure before the classifier runs) lands on the operational side by construction. `tamper` is asserted only on positive evidence: an artifact absent from the release's own asset inventory (an API read that either succeeded or aborted the step), or a cryptographic failure that survives two independent guards, a live TUF reachability probe and a transport-signature check on the captured failure text. Both guards are demote-only, so the worst case is a real finding reported as operational, which is still a red job that escalates after three consecutive failures and re-fires the next day. Never the reverse.

**One latent false-page found and fixed while building this:** the attestation presence check was `tar -tzf … | grep -q`, which under `pipefail` can SIGPIPE `tar` and turn a *present* attestation into a non-zero pipeline. It now lists to a file and greps the file, and an unlistable tarball classifies as `operational` rather than as a finding.

**SBOM signature checks are gated on a `v0.18.0` floor.** SBOM signing landed in #1957, after that release, and the current latest ships SBOMs with no signatures. Requiring them unconditionally would false-page today; from the next release a missing bundle is a finding. Both sides of the floor are covered by tests.

## Testing

```bash
go test ./tests/releasepolicy/...
make lint
actionlint -shellcheck=
```

22 subtests across `TestReleaseReverifyWorkflowShape`, `TestReleaseReverifyClassification`, and `TestReleaseReverifyGuardsAreLoadBearing`, executing the step's real `run:` under bash with fakes for `gh`, `cosign`, `curl`, `aicr` and `sleep`, and a real `tar -czf` archive. Coverage includes a clean pass, each missing-artifact case, Sigstore and Rekor outages, a TUF failure, a network timeout, and the case where a genuine finding coincides with an outage (which must still page).

`TestReleaseReverifyGuardsAreLoadBearing` mutates the extracted script in memory (each guard to `if false`) and asserts the intact script classifies `operational` while the mutated one classifies `tamper`, so the suite cannot pass if a guard silently stops being load-bearing.

Test delta: `tests/releasepolicy` fails the same 13 tests as unmodified `main` (`comm` of the two failure sets is empty in both directions), all from the missing GNU `timeout` on macOS. With a passthrough `timeout` shim the full package passes. `golangci-lint` reports 0 issues; `actionlint -shellcheck=` is clean.

**Also run against live services, not only fakes.** `workflow_dispatch` cannot be triggered before merge (GitHub requires the workflow file on the default branch), so both `run:` blocks were extracted verbatim and executed locally against the real GitHub API, the real `v0.18.0` release, live Rekor and live Sigstore:

- **Resolve step:** resolved `v0.18.0`, 13 assets, correct step summary.
- **Happy path:** asset-presence checks passed, `binary provenance verified for v0.18.0`, `catalog verified` with identity `on-tag.yaml@refs/tags/v0.18.0`, SBOM checks correctly skipped at the floor. `CLASSIFICATION=clean`, exit 0.
- **Missing artifact:** with `recipe-catalog.sigstore.json` removed from the inventory, `::error::release v0.18.0 does not publish recipe-catalog.sigstore.json`, `CLASSIFICATION=tamper`, exit 1.
- **Infrastructure failure:** with Sigstore unreachable and the download failing, `::warning::the released archive never downloaded; treating as infrastructure`, `CLASSIFICATION=operational`, exit 3.

The last two are the properties this job lives or dies on: a genuinely missing artifact pages, an outage does not. Both are now confirmed against real tooling rather than inferred from fake-based tests.

Separately, the live run confirms the `SBOM_SIGNING_FLOOR` is load-bearing rather than defensive: `v0.18.0` publishes six SPDX documents and zero `.sigstore.json` siblings, so without the floor this job would have paged on its first scheduled run.

## Risk Assessment

**Rollout notes:** Additive, observe-only. It reads releases and verifies published artifacts; it publishes nothing and changes no release path. Worst case on a bug is a spurious issue, and the classification design biases toward under-reporting rather than false-paging. The verification logic and both classification outcomes have been run against live GitHub, Rekor and Sigstore (see Testing). What remains unexercised is only the Actions plumbing around them: the composite-action invocation, the `steps.*.outcome` wiring, and the issue create/close/streak steps (same code as `rekor-monitor`, asserted structurally here), plus the three-consecutive-failure escalation, which needs three real runs. Running `gh workflow run release-reverify.yaml` immediately after merge exercises the first three.

## Checklist

### DIFF
--- a/.github/workflows/release-reverify.yaml
+++ b/.github/workflows/release-reverify.yaml
@@ -158,6 +158,14 @@ env:
   SIGSTORE_PROBE_URLS: >-
     https://tuf-repo-cdn.sigstore.dev/1.root.json
     https://fulcio.sigstore.dev/api/v2/trustBundle
+  # The binaries whose linux/amd64 SPDX SBOM every post-floor release must
+  # publish, each with a sibling .sigstore.json bundle. The expected set is
+  # derived from the TAG, never from the release's own inventory: deriving it
+  # from the artifact under test would make only a zero-SBOM release a finding,
+  # so deleting one of the two would leave the loop verifying the survivor and
+  # reporting clean. Authoritative list: expected_release_asset_names() in
+  # .github/scripts/release-images.sh -- keep this linux/amd64 subset in sync.
+  EXPECTED_SBOM_BINARIES: "aicr aicrd"
   # Workspace for downloads and captured verification logs, kept out of ./_rel
   # (which belongs to the install-aicr-release action).
   WORK_DIR: ./_reverify
@@ -258,10 +266,17 @@ jobs:
           RELEASE_ASSETS: ${{ steps.release.outputs.assets_file }}
           INSTALL_OUTCOME: ${{ steps.install.outcome }}
         run: |
-          # Deliberately not `set -e`: this step must run every check, aggregate
-          # the findings, and exit with the classified status. Each check owns
-          # its own failure handling.
+          # This step must run EVERY check, aggregate the findings, and exit with
+          # the classified status, so errexit has to be off. It is not off by
+          # default: with no `shell:` key GitHub invokes this block as
+          # `bash -e {0}`, and `set -uo pipefail` does not clear an inherited
+          # -e. Turn it off explicitly, as rekor-monitor.yaml does. Without
+          # this, any unguarded top-level failure aborts the step before
+          # `finish` runs, and a `tamper` already recorded earlier in the same
+          # run never reaches GITHUB_OUTPUT: the gates then read the empty
+          # classification as operational and the finding never pages.
           set -uo pipefail
+          set +e
 
           mkdir -p "${WORK_DIR}"
           classification=clean
@@ -434,17 +449,18 @@ jobs:
           elif ! command -v cosign > /dev/null 2>&1; then
             operational "cosign was never installed; treating as infrastructure"
           else
-            # Re-run the verification the action already performed, this time
-            # capturing its output so guard #2 has something to read. The action
-            # retried three times; this is the classification pass, and a pass
-            # here means the action's failure was transient -- its own log stays
-            # authoritative for WHICH of its sub-checks (checksum, extract,
-            # cosign) failed.
-            # Guarded: the step runs without `set -e`, so an unguarded failure
-            # here would fall through to cosign, which then fails on a missing
-            # bundle with a message no infra_shaped pattern matches, producing a
-            # security page for what is only a full disk or a permissions fault
-            # on WORK_DIR. Every sibling branch demotes, and so must this one.
+            # Re-run BOTH of the action's sub-checks, capturing their output so
+            # the demotion tests have something to read. Nothing here parses the
+            # action's own log, so reproducing its checks is the only way to know
+            # WHICH one failed; presuming "cosign" would misread a checksum
+            # mismatch (see the checksum re-check below). The action retried
+            # three times, so a pass here means its failure was transient.
+            # Guarded: errexit is off in this step (see `set +e` above), so an
+            # unguarded failure here would fall through to cosign, which then
+            # fails on a missing bundle with a message no infra_shaped pattern
+            # matches, producing a security page for what is only a full disk or
+            # a permissions fault on WORK_DIR. Every sibling branch demotes, and
+            # so must this one.
             if ! tar -xzf "${REL_DIR}/${archive}" -C "${WORK_DIR}" \
                 aicr aicr-attestation.sigstore.json 2>/dev/null; then
               operational "the archive listed but could not be extracted; treating as infrastructure"
@@ -454,7 +470,30 @@ jobs:
                 --certificate-oidc-issuer https://token.actions.githubusercontent.com \
                 --certificate-identity-regexp "${identity_re}" \
                 "${WORK_DIR}/aicr" > "${WORK_DIR}/binary-verify.log" 2>&1; then
-              operational "binary provenance verified on re-run after the install action failed; treating as a transient failure"
+              # Provenance passing does NOT clear the checksums manifest: the
+              # attestation binds the BINARY, while aicr_checksums.txt covers the
+              # ARCHIVE. The action fails at its own `sha256sum -c` before it
+              # ever reaches cosign, so without this re-check a manifest that
+              # stopped matching an otherwise-valid archive would report
+              # "transient" every single day, while every consumer following the
+              # documented checksum flow fails every single time.
+              #
+              # Ordered after cosign deliberately: a truncated or corrupt
+              # download also fails a checksum, but such an archive would have
+              # failed to list, or produced a binary the attestation rejects,
+              # further up this chain. Reaching here means the bytes are intact,
+              # so a mismatch is the published manifest disagreeing with the
+              # published archive.
+              if [ ! -r "${REL_DIR}/aicr_checksums.txt" ]; then
+                operational "the checksums manifest is not on disk to re-check; treating as infrastructure"
+              elif (cd "${REL_DIR}" && grep " ${archive}\$" aicr_checksums.txt | sha256sum -c -) \
+                  > "${WORK_DIR}/checksum-verify.log" 2>&1; then
+                operational "binary provenance and checksum verified on re-run after the install action failed; treating as a transient failure"
+              else
+                status=$?
+                cat "${WORK_DIR}/checksum-verify.log" >&2 || true
+                classify_failure "checksum verification of ${archive} for ${TAG}" "${WORK_DIR}/checksum-verify.log" "${status}"
+              fi
             else
               # $? is the elif condition's status and MUST be captured before any
               # other command runs, so timeout's 124 (or a 137 SIGKILL) survives
@@ -500,14 +539,23 @@ jobs:
           elif [ "${TAG}" = "${SBOM_SIGNING_FLOOR}" ] || [ "${newest}" != "${TAG}" ]; then
             echo "release ${TAG} is at or before the SBOM signing floor ${SBOM_SIGNING_FLOOR}; SBOM bundles not expected"
           else
-            sboms="$(grep -E '_linux_amd64\.sbom\.json$' "${RELEASE_ASSETS}" || true)"
-            if [ -z "${sboms}" ]; then
-              security "release ${TAG} publishes no linux/amd64 SBOM assets"
+            # Derived from the tag, so a mandatory SBOM that was never uploaded
+            # is a finding rather than an entry that simply is not iterated.
+            read -r -a sbom_binaries <<< "${EXPECTED_SBOM_BINARIES}"
+            expected_sboms=""
+            for binary in "${sbom_binaries[@]}"; do
+              expected_sboms+="${binary}_${version}_linux_amd64.sbom.json"$'\n'
+            done
+            expected_sboms="${expected_sboms%$'\n'}"
+            if [ -z "${expected_sboms}" ]; then
+              operational "EXPECTED_SBOM_BINARIES is empty; no SBOM subjects to check"
             fi
             while IFS= read -r sbom; do
               [ -n "${sbom}" ] || continue
               bundle="${sbom}.sigstore.json"
-              if ! have_asset "${bundle}"; then
+              if ! have_asset "${sbom}"; then
+                security "release ${TAG} does not publish the mandatory SBOM ${sbom}"
+              elif ! have_asset "${bundle}"; then
                 security "release ${TAG} ships ${sbom} with no attestation bundle (${bundle})"
               elif ! command -v cosign > /dev/null 2>&1; then
                 operational "cosign is unavailable; cannot verify ${sbom}"
@@ -530,7 +578,7 @@ jobs:
                 cat "${WORK_DIR}/${sbom}.verify.log" >&2 || true
                 classify_failure "SBOM attestation verification for ${sbom}" "${WORK_DIR}/${sbom}.verify.log" "${status}"
               fi
-            done <<< "${sboms}"
+            done <<< "${expected_sboms}"
           fi
 
           finish
--- a/.github/workflows/release-reverify.yaml
+++ b/.github/workflows/release-reverify.yaml
@@ -1,0 +1,525 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Daily re-verification of the latest published release's Sigstore artifacts
+# (NVIDIA/aicr#1461, part of the supply-chain epic #1149).
+#
+# WHAT THIS COVERS THAT rekor-monitor.yaml DOES NOT. The Rekor monitor proves the
+# transparency LOG is sound (append-only, no unexplained entries under the release
+# identity). It says nothing about whether a given release actually SHIPPED the
+# artifacts that chain to that log. A signing-side upload that silently failed or
+# was skipped -- goreleaser's post-hook exiting 0 without writing a bundle, an
+# asset that never made it onto the GitHub Release -- leaves a published release
+# whose provenance cannot be reconstructed by a consumer, and nothing in the log
+# is wrong. This workflow is the consumer's-eye check: every day it resolves the
+# latest release and runs the SHIPPED verification commands against it, exactly as
+# documented in docs/integrator/supply-chain-verification.md.
+#
+# WHAT IT VERIFIES (all release assets of the latest non-draft, non-prerelease
+# GitHub Release):
+#   - aicr_<version>_linux_amd64.tar.gz against aicr_checksums.txt, then the
+#     binary's SLSA provenance bundle (aicr-attestation.sigstore.json, shipped
+#     inside the archive) with `cosign verify-blob-attestation` pinned to
+#     on-tag.yaml@refs/tags/<exact tag>. This is delegated wholesale to the
+#     .github/actions/install-aicr-release composite, which already does the
+#     download, checksum, extract, exact-tag identity pinning, 3-attempt retry
+#     and fail-closed-on-missing-attestation.
+#   - recipe-catalog.sigstore.json (a loose release asset, deliberately NOT
+#     covered by aicr_checksums.txt) via `aicr recipe verify-catalog`, run with
+#     the released binary the step above just verified. This is the only check
+#     that proves the SHIPPED binary's embedded registry.yaml +
+#     validators/catalog.yaml still digest to what the release signed.
+#   - every aicr*/aicrd* linux/amd64 SPDX SBOM asset and its sibling
+#     <asset>.sigstore.json attestation bundle, again with
+#     `cosign verify-blob-attestation`. Gated behind SBOM_SIGNING_FLOOR: releases
+#     at or before that tag predate SBOM signing (NVIDIA/aicr#1957) and
+#     legitimately ship unsigned SBOMs; every release after it MUST ship a bundle
+#     per SBOM, and a missing one is a finding.
+#
+# WHAT IT DELIBERATELY DOES NOT VERIFY:
+#   - `aicr verify` (bundle attestation). A deployment bundle is produced on
+#     demand by a user (`aicr bundle --attest`); no bundle is a release asset, so
+#     there is nothing released to re-verify. Generating one here would test the
+#     signing path, not the retrievability of a published artifact.
+#   - The container images' OCI referrer attestations (SBOM / OpenVEX / SLSA
+#     provenance, NVIDIA/aicr#1982). Those live in ghcr.io's referrer store, a
+#     different system with a different retention and GC model from GitHub
+#     Releases, and re-verifying all seven images x three predicate kinds would
+#     add ~21 registry round-trips per run -- multiplying this job's operational
+#     noise against the one signal it exists to keep crisp. The images are
+#     already pulled and scanned weekly by vuln-scan-images.yaml. A sibling
+#     registry-side re-verification (which would need `gh attestation verify
+#     --bundle-from-oci` so it reads the REGISTRY copy rather than GitHub's
+#     attestations API) is tracked separately; see docs/contributor/maintaining.md.
+#   - Rekor entry liveness by log index. `cosign verify-blob-attestation` verifies
+#     a self-contained bundle: the inclusion proof and RFC3161 timestamp travel
+#     inside it and are checked against the live Sigstore TUF trust root, so a
+#     pass proves the bundle is retrievable and cryptographically sound, not that
+#     Rekor would still serve that entry by index. Re-fetching by index is a
+#     documented residual, also in maintaining.md.
+#
+# CLASSIFICATION (same vocabulary and exit codes as rekor-monitor.yaml, so the
+# two workflows triage identically):
+#   - clean      -> exit 0
+#   - tamper     -> exit 1, security: a released artifact is MISSING, or it fails
+#                   verification against a Sigstore that answered. Opens a
+#                   security tracking issue mentioning the maintainers.
+#   - operational-> exit 3, infrastructure: Sigstore/GitHub-API/network trouble.
+#                   Pages no one. The job goes red; only three consecutive failed
+#                   scheduled runs open a calm `area/ci` degraded issue.
+#
+# WHY AN INFRA FAILURE CANNOT MASQUERADE AS A MISSING ENTRY. `tamper` is asserted
+# on POSITIVE evidence only, never as a fallback:
+#   1. "Missing" means the asset name is absent from the release's own asset
+#      inventory (a GitHub API read that either succeeded or aborted the step), or
+#      the attestation is absent from an archive we successfully downloaded and
+#      checksummed. Neither can be produced by a network failure -- a failed read
+#      aborts before any comparison.
+#   2. A failed cryptographic verification is promoted to `tamper` only when BOTH
+#      independent guards agree: Sigstore's TUF CDN answered a live probe, AND the
+#      captured failure output carries no transport/outage signature. Either guard
+#      alone demotes to `operational`. The guards are demote-only, so the failure
+#      mode is a real finding reported as operational (still a red job, still a
+#      degraded issue after three days, and it re-fires tomorrow) -- never the
+#      reverse.
+#   3. Any failure BEFORE the classifying step runs (checkout, release
+#      resolution, load-versions) leaves the classification output empty, which
+#      the issue gates treat as operational by construction.
+#
+# TRIAGE:
+#   - Security issue (tamper): confirm by hand with the commands in
+#     docs/integrator/supply-chain-verification.md against the named tag. A
+#     genuinely missing asset means the release must be re-cut or the asset
+#     re-uploaded and re-signed; a verification failure on a present asset means
+#     the artifact does not match what the release signed -- treat as tampering
+#     and begin incident response.
+#   - Degraded issue (operational, 3+ consecutive days): Sigstore or the GitHub
+#     API has been unreachable. No action unless it persists past an upstream
+#     incident window.
+
+name: Release Re-Verification
+
+on:
+  schedule:
+    - cron: "41 13 * * *"  # daily, offset away from the release and UAT crons
+  # Also manual: GitHub disables scheduled workflows after 60 days of repository
+  # inactivity, and a dispatch re-arms the schedule as well as allowing an
+  # on-demand run after a release.
+  workflow_dispatch: {}
+
+permissions: {}
+
+concurrency:
+  group: release-reverify
+  cancel-in-progress: false
+
+env:
+  ALERT_TITLE: "Release re-verification: a released artifact is missing or does not verify"
+  DEGRADED_TITLE: "Release re-verification: degraded"
+  # Maintainers group notified on an alert (GitHub team; issues cannot be
+  # assigned to a team, so it is @-mentioned in the issue body instead).
+  MAINTAINERS_TEAM: "@nvidia/aicr-maintainer"
+  # Releases at or before this tag predate signed SBOM assets (NVIDIA/aicr#1957)
+  # and ship SPDX documents with no sibling .sigstore.json. From the next release
+  # onward a missing bundle is a finding. Raise this ONLY to correct history.
+  SBOM_SIGNING_FLOOR: v0.18.0
+  # Liveness probe for the Sigstore trust root every verification depends on.
+  # This is the outer guard that keeps an outage from reading as a finding.
+  SIGSTORE_PROBE_URL: https://tuf-repo-cdn.sigstore.dev/1.root.json
+  # Workspace for downloads and captured verification logs, kept out of ./_rel
+  # (which belongs to the install-aicr-release action).
+  WORK_DIR: ./_reverify
+  # Where install-aicr-release leaves its download and its installed binary.
+  REL_DIR: ./_rel
+  AICR_BIN: ./aicr
+
+jobs:
+  reverify:
+    name: Re-verify the latest release's Sigstore artifacts
+    runs-on: ubuntu-latest
+    # Bounds the worst case: a release download, one cosign verification per
+    # artifact (each internally retried), and the issue bookkeeping. Every
+    # network call is individually bounded by `timeout --foreground`, so this is
+    # a backstop, not the primary bound.
+    timeout-minutes: 30
+    permissions:
+      contents: read  # checkout, plus `gh release download` for the assets
+      actions: read   # read this workflow's run history for the failure streak
+      issues: write   # open/close the alert and degraded issues
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Resolve the latest release
+        id: release
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          # /releases/latest excludes drafts and prereleases, so this is the
+          # release a `curl .../releases/latest` consumer would actually install
+          # -- the same resolution documented in supply-chain-verification.md.
+          # gh-api-retry.sh publishes release.json only on success and removes it
+          # on total failure, so a GitHub API outage aborts this step (and, with
+          # no classification emitted, is treated as operational downstream)
+          # rather than falling through to an empty asset inventory that would
+          # read as "every asset is missing".
+          .github/scripts/gh-api-retry.sh release.json "repos/${GITHUB_REPOSITORY}/releases/latest"
+          tag="$(jq -r '.tag_name // empty' release.json)"
+          # Validate before the tag reaches a regexp or a filename. Same shape
+          # install-aicr-release enforces on its input.
+          if [[ ! "${tag}" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z][0-9A-Za-z.]*)?$ ]]; then
+            echo "::error::/releases/latest returned '${tag}', which is not a release tag like v1.2.3." >&2
+            exit 1
+          fi
+          jq -r '.assets[].name' release.json > release-assets.txt
+          {
+            echo "tag=${tag}"
+            echo "assets_file=${PWD}/release-assets.txt"
+          } >> "$GITHUB_OUTPUT"
+          echo "Resolved latest release ${tag} with $(wc -l < release-assets.txt) assets."
+          {
+            echo "### Release re-verification"
+            echo
+            echo "Target: [\`${tag}\`](${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/releases/tag/${tag})"
+          } >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Load versions
+        id: versions
+        uses: ./.github/actions/load-versions
+
+      # The authoritative binary check. Reused rather than reimplemented: this
+      # action already downloads the archive and checksums, verifies the tarball
+      # against the published checksum, extracts the SLSA provenance bundle,
+      # fails closed when the bundle is absent, and runs
+      # `cosign verify-blob-attestation` pinned to on-tag.yaml at the EXACT tag
+      # with a 3-attempt retry. continue-on-error hands its outcome to the
+      # classifying step below instead of ending the job with an unclassified
+      # red, which would page nobody but also file nothing.
+      - name: Verify the released aicr binary provenance
+        id: install
+        continue-on-error: true
+        uses: ./.github/actions/install-aicr-release
+        with:
+          aicr-version: ${{ steps.release.outputs.tag }}
+          cosign_version: ${{ steps.versions.outputs.cosign }}
+
+      - name: Re-verify release artifacts and classify
+        id: verify
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          TAG: ${{ steps.release.outputs.tag }}
+          RELEASE_ASSETS: ${{ steps.release.outputs.assets_file }}
+          INSTALL_OUTCOME: ${{ steps.install.outcome }}
+        run: |
+          # Deliberately not `set -e`: this step must run every check, aggregate
+          # the findings, and exit with the classified status. Each check owns
+          # its own failure handling.
+          set -uo pipefail
+
+          mkdir -p "${WORK_DIR}"
+          classification=clean
+
+          # security() records POSITIVE evidence that a released artifact is
+          # missing or does not verify. It is the only path to a page, and it is
+          # never reachable from a failed network call.
+          security() {
+            echo "::error::$1" >&2
+            classification=tamper
+          }
+
+          # operational() records infrastructure trouble. It never overwrites a
+          # security finding: a real finding that coincides with an outage still
+          # pages.
+          operational() {
+            echo "::warning::$1" >&2
+            [ "${classification}" = "tamper" ] || classification=operational
+          }
+
+          # sigstore_reachable is guard #1. Every verification here depends on
+          # the Sigstore TUF trust root; if the CDN does not answer, a failed
+          # verification says nothing about the artifact.
+          sigstore_reachable() {
+            local attempt
+            for attempt in 1 2 3; do
+              if timeout --foreground 30s curl -fsS -o /dev/null "${SIGSTORE_PROBE_URL}"; then
+                return 0
+              fi
+              [ "${attempt}" -eq 3 ] || sleep $((attempt * 5))
+            done
+            return 1
+          }
+
+          # infra_shaped is guard #2: does the captured failure output carry a
+          # transport or upstream-outage signature? Demote-only by construction
+          # -- a match can turn a would-be finding into operational, never the
+          # reverse -- so a broad pattern set is the safe direction.
+          infra_shaped() {
+            local patterns
+            patterns='timed? ?out|deadline exceeded|connection (refused|reset)|no such host'
+            patterns="${patterns}|i/o timeout|dial tcp|network is unreachable|tls handshake"
+            patterns="${patterns}|unexpected eof|temporary failure|too many requests|rate limit"
+            patterns="${patterns}|status (429|5[0-9][0-9])|service unavailable|bad gateway"
+            patterns="${patterns}|gateway time-?out|trusted root|tuf|error fetching"
+            patterns="${patterns}|could not (fetch|download|reach)|failed to (fetch|download)"
+            grep -qiE "${patterns}" "$1"
+          }
+
+          # classify_failure turns a failed cryptographic verification into a
+          # classification. tamper requires BOTH guards to agree; anything else
+          # is operational.
+          classify_failure() {
+            if ! sigstore_reachable; then
+              operational "$1 failed while Sigstore was unreachable; treating as infrastructure, not a finding"
+              return
+            fi
+            if infra_shaped "$2"; then
+              operational "$1 failed with a transport/outage signature in its output; treating as infrastructure, not a finding"
+              return
+            fi
+            security "$1 failed against a reachable Sigstore: the released artifact does not verify"
+          }
+
+          have_asset() { grep -Fxq "$1" "${RELEASE_ASSETS}"; }
+
+          version="${TAG#v}"
+          archive="aicr_${version}_linux_amd64.tar.gz"
+
+          # Pin provenance to the release workflow AND the exact tag, matching
+          # install-aicr-release: a bundle attested for one release must not
+          # satisfy another. Escape the tag's dots so they match literally.
+          identity_tag="${TAG//./\\.}"
+          identity_re="^https://github.com/${REPO}/.github/workflows/on-tag\\.yaml@refs/tags/${identity_tag}$"
+
+          # (1) Inventory. Absence from the release's own asset list is the
+          # cleanest possible "the upload silently failed" signal, and it is
+          # decided from an API read that already succeeded.
+          for required in "${archive}" "aicr_checksums.txt" "recipe-catalog.sigstore.json"; do
+            if have_asset "${required}"; then
+              echo "asset present: ${required}"
+            else
+              security "release ${TAG} does not publish ${required}; a consumer cannot verify this release"
+            fi
+          done
+
+          # (2) Binary provenance. install-aicr-release is authoritative; this
+          # only classifies its outcome.
+          if [ "${INSTALL_OUTCOME}" = "success" ]; then
+            echo "binary provenance verified for ${TAG}"
+          elif [ ! -f "${REL_DIR}/${archive}" ]; then
+            operational "the released archive never downloaded; treating as infrastructure"
+          elif ! tar -tzf "${REL_DIR}/${archive}" > "${WORK_DIR}/archive-members.txt" 2>/dev/null; then
+            # Listed into a file rather than piped into `grep -q`: under
+            # `pipefail`, grep's early exit can SIGPIPE tar and turn a PRESENT
+            # attestation into a non-zero pipeline, i.e. a false finding.
+            # A tarball that will not list at all is a truncated or corrupt
+            # download, not a publishing failure.
+            operational "the downloaded archive could not be listed; treating as infrastructure"
+          elif ! grep -qx 'aicr-attestation.sigstore.json' "${WORK_DIR}/archive-members.txt"; then
+            security "release ${TAG} ships no binary attestation (aicr-attestation.sigstore.json) inside ${archive}"
+          elif ! command -v cosign > /dev/null 2>&1; then
+            operational "cosign was never installed; treating as infrastructure"
+          else
+            # Re-run the verification the action already performed, this time
+            # capturing its output so guard #2 has something to read. The action
+            # retried three times; this is the classification pass, and a pass
+            # here means the action's failure was transient -- its own log stays
+            # authoritative for WHICH of its sub-checks (checksum, extract,
+            # cosign) failed.
+            tar -xzf "${REL_DIR}/${archive}" -C "${WORK_DIR}" aicr aicr-attestation.sigstore.json
+            if timeout --foreground 300s cosign verify-blob-attestation \
+                --bundle "${WORK_DIR}/aicr-attestation.sigstore.json" \
+                --type https://slsa.dev/provenance/v1 \
+                --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+                --certificate-identity-regexp "${identity_re}" \
+                "${WORK_DIR}/aicr" > "${WORK_DIR}/binary-verify.log" 2>&1; then
+              operational "binary provenance verified on re-run after the install action failed; treating as a transient failure"
+            else
+              cat "${WORK_DIR}/binary-verify.log" >&2
+              classify_failure "binary provenance verification for ${TAG}" "${WORK_DIR}/binary-verify.log"
+            fi
+          fi
+
+          # (3) Signed recipe catalog. The loose asset is the path
+          # cli-reference.md tells users to curl, so that is the copy verified
+          # here; the archive ships an identical bundle.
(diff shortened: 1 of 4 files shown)


.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,525 @@
...
+                  --type https://slsa.dev/provenance/v1 \
+                  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+                  --certificate-identity-regexp "${identity_re}" \
+                  "${WORK_DIR}/${sbom}" > "${WORK_DIR}/${sbom}.verify.log" 2>&1; then
+                echo "SBOM attestation verified: ${sbom}"
+              else
+                cat "${WORK_DIR}/${sbom}.verify.log" >&2
+                classify_failure "SBOM attestation verification for ${sbom}" "${WORK_DIR}/${sbom}.verify.log"
+              fi
+            done <<< "${sboms}"
```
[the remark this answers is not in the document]
lockwobr: Valid, and worth more than its Trivial rating, so I fixed it and put it under test. Applied in 6cefa015.

Both `gh release download` and `cosign verify-blob-attestation` now take `< /dev/null`, with a comment recording why. Neither binary reads stdin today, so this was latent rather than live, but the failure mode is the one this job exists to catch: SBOMs skipped with no `::error::`, no `::warning::` and a `clean` classification. A check that silently verifies nothing is worse than one that fails.

Added `TestReleaseReverifySBOMLoopIsolatesChildStdin`, which models a stdin-consuming child in each fake independently. It pins the redirect count at 2 (so adding a third stdin-reading child to the loop has to revisit the test rather than silently neuter it), asserts the intact script verifies all three subjects in order, and asserts a script with the redirects stripped verifies exactly the first one. It also asserts the mutated run still reports `clean` and exit 0, which puts the *silence* under test rather than just the skip.

Confirmed against a real on-disk mutation rather than only the in-memory one: stripping the redirects skips `aicrd_0.19.0_linux_amd64.sbom.json` and `aicr-gate_0.19.0_linux_amd64.sbom.json` with a clean classification. Three subjects rather than two, because with two an early-terminating loop is indistinguishable from one that merely drops the last entry.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+          elif ! command -v cosign > /dev/null 2>&1; then
+            operational "cosign was never installed; treating as infrastructure"
+          else
+            # Re-run the verification the action already performed, this time
+            # capturing its output so guard #2 has something to read. The action
+            # retried three times; this is the classification pass, and a pass
+            # here means the action's failure was transient -- its own log stays
+            # authoritative for WHICH of its sub-checks (checksum, extract,
+            # cosign) failed.
+            tar -xzf "${REL_DIR}/${archive}" -C "${WORK_DIR}" aicr aicr-attestation.sigstore.json
```
mchmarny: Non-blocking: this `tar -xzf` is the one unguarded command in the chain, and a failure here manufactures a page.

The step runs under `set -uo pipefail` with `-e` deliberately off, so a failed extraction doesn't stop anything — control falls straight into `cosign`, which then fails on a missing bundle. `classify_failure` runs both guards and both pass it through: Sigstore is reachable, and cosign's `no such file or directory` matches none of the `infra_shaped` patterns (the closest are `could not (fetch|download|reach)` and `failed to (fetch|download)`). Result: `security` → exit 1 → an `area/security` issue that @-mentions `@nvidia/aicr-maintainer`, for a full disk.

It's narrow — `tar -tzf` must have listed successfully and the member grep must have matched, so only an environmental failure (disk full, permissions on `${WORK_DIR}`) reaches it. But every sibling branch in this same chain already demotes to `operational`, and it's the one spot that contradicts the header's stated invariant that `tamper` is asserted on positive evidence only and "can never be produced by" an infrastructure fault.

Minimum correction, matching the shape of the branches directly above:

```bash
if ! tar -xzf "${REL_DIR}/${archive}" -C "${WORK_DIR}" aicr aicr-attestation.sigstore.json 2>/dev/null; then
  operational "the archive listed but could not be extracted; treating as infrastructure"
elif timeout --foreground 300s cosign verify-blob-attestation \
    ...
```

A regression row in `TestReleaseReverifyClassification` with an unwritable `WORK_DIR` (or a fake `tar` that exits non-zero on extract while the earlier list succeeds) would pin it — the suite already proves it can distinguish these directions.

mchmarny (commented):
**COMMENT** — 0 merge blockers on `6cefa01`, 1 non-blocking finding.

I exercised the classifier rather than reasoning about it from the source, because the thing that would actually hurt here is a day-one false page on a healthy release:

- **Ran this workflow's exact binary check against the real latest release.** `cosign verify-blob-attestation --type https://slsa.dev/provenance/v1 --certificate-identity-regexp '^https://github.com/NVIDIA/aicr/.github/workflows/on-tag\.yaml@refs/tags/v0\.18\.0$'` against the v0.18.0 tarball → **Verified OK**.
- **Checked every day-one precondition that could page.** Latest is `v0.18.0`, exactly `SBOM_SIGNING_FLOOR`, so the SBOM leg is correctly exempt (that release does ship unsigned SBOMs). The v0.18.0 archive really does contain `aicr-attestation.sigstore.json`, `recipe-catalog.sigstore.json` and `aicr`, so the `tar -xzf` member list is right; `recipe verify-catalog` has existed since v0.15.0, so the *released* binary can run it; and all three required inventory assets are present.
- **Chased down the SBOM predicate type**, which was my main blocker hypothesis — `--type https://slsa.dev/provenance/v1` looks wrong for an SBOM. It isn't: `sign-sbom.sh` attests the SBOM *as subject* with the run's SLSA provenance as predicate, and its header documents the identical verify command. (The `spdxjson`/`openvex` types in the docs belong to the container-image attestations — a different system.)
- **Checked for a trust inversion in the catalog check.** It runs `./aicr`, so I traced `install-aicr-release`: the download/checksum/extract/cosign block is a subshell under `set -euo pipefail`, and `install -m 0755 ./_rel/aicr ./aicr` only runs after it succeeds. So `[ -x "${AICR_BIN}" ]` genuinely means "provenance-verified binary" — there's no path where an unverified binary vouches for the catalog.
- **Mutation-tested the suite in the dangerous direction.** Changing `security()` to set `classification=operational` — real findings never page — fails 10 subtests across `TestReleaseReverifyClassification` and `TestReleaseReverifyGuardsAreLoadBearing`. The tests are load-bearing, not decorative. `go test -run TestReleaseReverify` passes intact; they ship their own passthrough `timeout` shim, so they avoid the missing-GNU-`timeout` failures the rest of the package has on macOS.

The stdin-isolation test around `release_reverify_test.go:530` is the best thing here: it pins the redirect count, strips the redirects, and asserts the mutated loop verifies only the first SBOM **and still reports clean**. Pinning the false-PASS shape with a counterfactual is exactly right for a job whose whole purpose is catching silent skips.

The identity regexp leaves the dots in `github.com` and `.github` unescaped, but that is character-for-character what `install-aicr-release` already builds and the comment says the match is deliberate — not a finding.

CI is 27 pass / 14 skipping, 0 failing. `BEHIND main`, so a rebase is needed before merge. One nit not worth its own thread: the streak query counts any non-`success` conclusion, so a `cancelled` or `skipped` scheduled run inflates the count toward the degraded issue — calm `area/ci` path only, never a page.

njhensley (commented):
## 🔍 Multi-persona review

**Method:** 4 independent persona reviewers (correctness, security/supply-chain, CI-DX/operability, test-scope) → an adversarial senior meta-reviewer that re-derived every finding from the resolved shell at `6cefa015`. Legend: 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick.

**Overall:** A genuinely well-built, well-tested workflow. The core design — positive-evidence-only `tamper`, demote-only guards, empty-classification→operational routing, reuse of the hardened `install-aicr-release` composite — is sound and unusually well covered (mutation-tested guards, a counterfactual stdin-isolation test). **No blockers.** Every surviving finding sits on the edge of the same two invariants the workflow exists to guarantee: *never false-page*, and *never drop a real signal*. Two of those (the two false-page findings and the silently-closed-tamper finding) are worth closing before this becomes load-bearing.

**Recommendation: Approve with comments.** Additive/observe-only, biases toward under-reporting, easy to revert.

### Prior reviewers (duplicate-work note)
- **@mchmarny** — 0 merge blockers on `6cefa01`; his non-blocking `tar -xzf` finding is confirmed and deepened here (now with the note that it has no regression test), and his regexp nit is included as a nitpick.
- **@lockwobr** — CodeRabbit's stdin finding was already fixed (`< /dev/null`) and is **not** re-raised.
- The empty-log false-page, the non-tag-scoped auto-close, and the missing-Slack-page findings are net-new beyond the prior reviews.

> ℹ️ For whoever triages: a three-dot `git diff origin/main...HEAD` in a stale checkout misleadingly attributes ~400 lines of unrelated `sbom-and-attest`/`attest-image-from-tag` tests to this PR — those are already-merged #1982. The true PR (`git diff 9b180ac5b..HEAD`) is **4 files**; `release_workflow_test.go` changes by **+1 line**. No scope issue. The PR is **behind main** — rebase before merge.

### ✅ Confirmed non-issues (examined, cleared)
- **Trust inversion in the catalog check** — `./aicr` is written only *after* cosign verify succeeds in `install-aicr-release`; `[ -x AICR_BIN ]` can't be true with an unverified binary.
- **SBOM `--type https://slsa.dev/provenance/v1`** — matches `sign-sbom.sh`; correct predicate + subject binding.
- **`${{ }}` injection into `run:`** — none; all values arrive via `env:`, enforced by the shape test.
- **Hardening** — `permissions: {}`, least-privilege job perms, `persist-credentials:false`, all actions SHA-pinned.
- **`sort -V` SBOM floor, `have_asset` `grep -Fxq`, SBOM regex excluding `.sigstore.json`, `< /dev/null` stdin isolation, empty→operational routing** — correct and test-locked.
- **`gh-api-retry.sh` in resolve** — fail-closed; an API outage aborts rather than reading an empty inventory as "all assets missing."

### Tier table
|:---:|:---:|:---:|:---:|
| 0 | 4 | 4 | 2 |

Major: empty-log false page · unguarded `tar -xzf` · silently-closed tamper issue · no Slack page. Details inline.

<sub>Reviewed with a multi-persona + adversarial meta-review pass. Findings verified against the resolved code at the head SHA.</sub>

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+
+          # classify_failure turns a failed cryptographic verification into a
+          # classification. tamper requires BOTH guards to agree; anything else
+          # is operational.
+          classify_failure() {
+            if ! sigstore_reachable; then
+              operational "$1 failed while Sigstore was unreachable; treating as infrastructure, not a finding"
+              return
+            fi
+            if infra_shaped "$2"; then
```
njhensley: **🟠 Major — infra_shaped false-negatives on an empty log → false tamper page**

`classify_failure` reaches `security()` (tamper → pages `@nvidia/aicr-maintainer`) whenever `sigstore_reachable` passes AND `infra_shaped "$2"` is false. On an **empty/near-empty** captured log, `grep -qiE` returns non-zero → `infra_shaped` false → tamper. A `timeout --foreground 300s` kill (L343/365/399) exits 124 having written nothing; an OOM/SIGKILL of cosign likewise leaves empty stderr. All are infrastructure, but if the TUF CDN answers the probe (a *different* endpoint than whatever hung) they classify as `tamper` — directly violating the header's L82–95 "positive evidence only" invariant.

**Blast radius:** False security alert @-mentioning the maintainers; defeats the workflow's single reason to exist.

**Fix:** In `classify_failure`, demote before the pattern test — `[ -s "$2" ] || { operational "..."; return; }` — and treat timeout exit codes (124/137) as operational. Same fix family as the `tar -xzf` finding below.

.github/workflows/release-reverify.yaml
njhensley: **🟠 Major — Unguarded `tar -xzf` turns an environmental extract failure into a false tamper page (mchmarny-flagged; untested)**

Under `set -uo pipefail` (no `-e`), this extract is unchecked. L324 `tar -tzf` already proved the archive lists and L331 proved the member is present, so a failure *here* is environmental (disk-full, `WORK_DIR` perms, ENOSPC mid-write). Control falls through to `cosign` at L343 against a now-missing `${WORK_DIR}/aicr` → `open …: no such file` → matches no `infra_shaped` pattern → with TUF reachable → `tamper`. Confirming and deepening @mchmarny's inline finding; adding that no test covers list-succeeds/extract-fails (`corruptArchive` fails at `-tzf`, `omitAttestation` fails at L331).

**Blast radius:** False security page on a local/infra hiccup — a direct violation of the workflow's core "infra never masquerades as a finding" property.

**Fix:** `tar -xzf … || { operational "the archive listed but could not be extracted; treating as infrastructure"; }` and add a regression row (unwritable WORK_DIR, or a `tar` fake that fails extract but not list).

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+            --label "area/ci" \
+            --body-file degraded-body.md
+
+      - name: Close open issues on success
+        if: ${{ success() }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          for title in "${ALERT_TITLE}" "${DEGRADED_TITLE}"; do
```
njhensley: **🟠 Major — Non-tag-scoped auto-close silently closes an unresolved tamper issue**

`ALERT_TITLE` (L128) carries no tag, and this Close-on-success loop closes **every** open issue matching that exact title on **any** clean run, while the job only ever verifies `/releases/latest` (L179). Sequence: release `vX` pages a tamper issue → maintainers begin incident response → `vX+1` ships (unrelated) → the next daily run verifies `vX+1` clean (exit 0) → it closes the `vX` tamper issue with "completed cleanly," though the clean run proved nothing about `vX`, which is never re-checked again (no longer latest). The body's "auto-closes on the next clean run" (L455) is misleading for this path. Distinct from rekor-monitor, whose alert tracks the release-agnostic *log*, so a clean run there genuinely clears it.

**Blast radius:** An unremediated supply-chain tamper finding is auto-resolved by an unrelated newer release; affected consumers of vX stay exposed and vX is never re-checked.

**Fix:** Embed the resolved `${TAG}` in the alert title (or a per-tag label/marker) and scope the close to the tag the clean run actually verified.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+          # Machine-readable line for the log, mirroring tools/rekor-monitor.
+          echo "CLASSIFICATION=${classification}"
+          echo "classification=${classification}" >> "${GITHUB_OUTPUT}"
+          case "${classification}" in
+            clean) exit 0 ;;
+            tamper) exit 1 ;;
+            *) exit 3 ;;
+          esac
+
+      - name: Open security alert issue
```
njhensley, line 586: **🟠 Major — No Slack page for a tamper finding; diverges from rekor-monitor's alerting**

The sibling this PR mirrors, `rekor-monitor.yaml`, posts a Slack alert on every security classification ("Post Slack alert on security finding" via `SLACK_SERVICE`). This workflow has **no Slack step** — a `tamper` (which the alert body at L451 says to "treat as tampering and begin incident response") produces only a GitHub issue @-mentioning `@nvidia/aicr-maintainer`, whose delivery depends entirely on each maintainer's team-mention notification settings. The header's "triage identically" (L72) holds only for vocabulary/exit codes, not paging reliability.

**Blast radius:** An incident-grade supply-chain finding can sit unseen; compounds the no-retry issue-create path below (the GitHub issue is the *only* channel).

**Fix:** Add the Slack step (reuse `SLACK_SERVICE`), or document in the header why this security signal is intentionally lower-urgency than rekor's. Maintainer call.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          # Persistence via run history rather than a state file: escalate only
+          # when this run plus the two prior SCHEDULED runs all failed. The API
+          # returns completed runs most-recent-first and excludes this in-progress
+          # run, so index("success") is the count of consecutive leading
+          # failures; `// length` handles an all-failures window. This is a true
+          # streak, not a failure count: [fail,success,fail] yields 1.
+          streak="$(gh api "repos/${GITHUB_REPOSITORY}/actions/workflows/release-reverify.yaml/runs?event=schedule&status=completed&per_page=3" \
```
njhensley: **🟡 Minor — Failure streak counts cancelled/skipped/timed_out runs as failures**

`[.workflow_runs[].conclusion] | (index("success") // length)` over `status=completed` counts every non-`success` leading conclusion (cancelled, skipped, startup_failure, timed_out) toward the 3-consecutive threshold. A manually-cancelled or timed-out scheduled run inflates the streak and can open the calm `area/ci` degraded issue before three genuine operational failures. Direction-safe (only over-escalates the no-page path; identical to rekor-monitor L366-367).

**Blast radius:** Occasional spurious (calm, deduped, auto-closed) degraded issue. Never a page.

**Fix:** Optional: count only `"failure"` conclusions, e.g. `map(. == "failure") | index(false) // length`.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+            --body-file degraded-body.md
+
+      - name: Close open issues on success
+        if: ${{ success() }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+          for title in "${ALERT_TITLE}" "${DEGRADED_TITLE}"; do
+            gh issue list --repo "${GITHUB_REPOSITORY}" --state open \
```
njhensley: **🟡 Minor — Close-on-success gh issue list is unguarded → a transient list error flips a clean run red**

The per-issue `gh issue close` is `|| echo ::warning`-guarded (L525-527), but this outer `gh issue list | jq | while` pipeline (L517-520) is unguarded under `set -euo pipefail`. A transient 5xx/rate-limit on `gh issue list` makes the pipeline non-zero → `set -e` aborts the step → the job goes **failed on a clean verification**, that run then counts as non-success in the next streak query, and any open issue isn't closed this cycle. Mirrors the same latent gap in rekor-monitor L418-429.

**Blast radius:** Misleading red on a clean run + streak inflation (no-page side). No dropped genuine finding.

**Fix:** Capture the list with `|| true` (or `mapfile`) so a list blip can't fail the success path.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+            echo
+            echo "The run log names the exact artifact. Confirm by hand against \`${TAG}\`, then:"
+            echo "- **asset absent** — a signing/upload step silently skipped. Re-upload and re-sign the asset, or re-cut the release."
+            echo "- **asset present but fails verification** — the published bytes do not match what the release signed. Treat as tampering and begin incident response."
+            echo
+            echo "Sigstore reachability and the failure output were both checked before this was raised, so an outage alone cannot have produced it."
+            echo
+            echo "This issue auto-closes on the next clean run."
+          } > alert-body.md
+          url="$(gh issue create --repo "${GITHUB_REPOSITORY}" \
```
njhensley: **🟡 Minor — Alert issue create/list has no retry → a genuine tamper page can be dropped**

The alert step runs `gh issue list` (L435) and `gh issue create --label "area/security,theme/supply-chain"` (L457) under `set -e` with **no retry** — unlike release-resolution, which uses `gh-api-retry.sh`. A transient API 5xx, or a renamed/deleted label (create rejects an unknown label), aborts the step → no issue opens, no @-mention, and the degraded fallback is gated out (`classification == 'tamper'`). The only remaining signal is a red job — for the one event this workflow exists to surface loudly.

**Blast radius:** Low probability, high impact — a real tamper finding with no notification. Compounds the missing-Slack finding.

**Fix:** Wrap create/list in bounded retry (reuse `gh-api-retry.sh`) and/or tolerate a missing label.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+
+jobs:
+  reverify:
+    name: Re-verify the latest release's Sigstore artifacts
+    runs-on: ubuntu-latest
+    # Bounds the worst case: a release download, one cosign verification per
+    # artifact (each internally retried), and the issue bookkeeping. Every
+    # network call is individually bounded by `timeout --foreground`, so this is
+    # a backstop, not the primary bound.
+    timeout-minutes: 30
```
njhensley: **🟡 Minor — timeout-minutes:30 exceedable by stacked per-call bounds → cancelled run opens no degraded issue**

The per-call `timeout --foreground` bounds are sequential and stack: binary cosign 300s + catalog dl 120s + catalog verify 300s + N×(SBOM dl 120s + SBOM cosign 300s) + install-action retries + sigstore probes. A real release ships ~3 SBOMs (aicr/aicrd/aicr-gate), so worst case ≈ 3×420 + 420 + overhead > 30 min. If a slow-but-reachable Sigstore drives each cosign toward its ceiling, the job hits `timeout-minutes` and is **cancelled** — cancelled skips `if: failure()`, so the degraded step never runs → a red run with no tracking issue, still counted in the streak.

**Blast radius:** Atypical trigger (sustained ~300s Sigstore latency); stays on the no-page side (red run, no page, no dropped finding).

**Fix:** Raise the job timeout with explicit worst-case sizing (as rekor-monitor did), or shrink/bound the per-call cosign timeouts and SBOM count so the stacked worst case provably fits.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+          # pages.
+          operational() {
+            echo "::warning::$1" >&2
+            [ "${classification}" = "tamper" ] || classification=operational
+          }
+
+          # sigstore_reachable is guard #1. Every verification here depends on
+          # the Sigstore TUF trust root; if the CDN does not answer, a failed
+          # verification says nothing about the artifact.
+          sigstore_reachable() {
```
njhensley, line 326: **🔵 Nitpick — sigstore_reachable probes the TUF CDN only, not Rekor/Fulcio**

`SIGSTORE_PROBE_URL` (L139) targets `tuf-repo-cdn.sigstore.dev` only, so guard #1 stays green during a Rekor/Fulcio-only outage. Not material: `verify-blob-attestation` checks a self-contained bundle offline against the TUF trust root, so the TUF CDN is the real network dependency, and residual Rekor/Fulcio error text is caught by guard #2's `status 5xx|service unavailable|tuf|trusted root` patterns.

**Blast radius:** None in practice; noted for completeness of the two-guard independence claim.

**Fix:** Optional: add a Rekor/Fulcio liveness probe, or document that guard #2 is the intended backstop for non-TUF outages.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,529 @@
...
+          have_asset() { grep -Fxq "$1" "${RELEASE_ASSETS}"; }
+
+          version="${TAG#v}"
+          archive="aicr_${version}_linux_amd64.tar.gz"
+
+          # Pin provenance to the release workflow AND the exact tag, matching
+          # install-aicr-release: a bundle attested for one release must not
+          # satisfy another. Escape the tag's dots so they match literally.
+          identity_tag="${TAG//./\\.}"
+          identity_re="^https://github.com/${REPO}/.github/workflows/on-tag\\.yaml@refs/tags/${identity_tag}$"
```
njhensley: **🔵 Nitpick — Identity regexp leaves github.com / .github dots unescaped (not exploitable)**

Only the tag's dots are escaped (`identity_tag="${TAG//./\\.}"`); the literal dots in `github.com` and `/.github/` remain, so the anchored regexp reads them as "any char." Not reachable: `REPO` is pinned to `NVIDIA/aicr`, the OIDC issuer is pinned, and the Fulcio SAN structure is fixed. Same looseness exists in `install-aicr-release/action.yml:80`; inconsistent with `sign-sbom.sh`'s fully-escaped canonical form. @mchmarny noted the same and called it not-a-finding.

**Blast radius:** None today; a latent widening / consistency nit.

**Fix:** Optional: escape those dots on both sides to match the signer-side canonical pattern.

.github/workflows/release-reverify.yaml
lockwobr: Correct, and thank you for the precision: the trace, the exact reason both guards pass it through, and the structural argument that every sibling branch already demotes. Fixed in d28752c7 with the shape you suggested.

I reproduced it on disk before fixing, and it is verbatim what you described: `::error::binary provenance verification for v0.18.0 failed against a reachable Sigstore`, `CLASSIFICATION=tamper`, for a failed extract. Pinned two ways, per your suggestion: a row in `TestReleaseReverifyClassification` driven by a fake `tar` that succeeds on `-tzf` and fails on `-xzf` (chosen over an unwritable WORK_DIR so it does not depend on runner privileges), and a mutation row in `TestReleaseReverifyGuardsAreLoadBearing` that restores the unguarded form and asserts it flips to `tamper`.

Your finding also prompted an audit of the whole step for the same shape, which found three more paths that bypass the guards rather than misjudge them. Those are fixed in the same commit and detailed on njhensley's threads: `classify_failure` trusting an unreadable or empty log, an unreadable asset inventory reporting every asset missing, and an unguarded `sort -V` failing open. The general lesson I have written into the header is that the guards are demote-only, so they cannot rescue a path that never reaches them.

.github/workflows/release-reverify.yaml
lockwobr: Correct, and it turned out to have a second trigger you did not name. Fixed in d28752c7.

Yours: an empty log makes `grep -qiE` return non-zero, so `infra_shaped` is false and control reaches `security`. A `timeout` kill exits 124 having written nothing, an OOM leaves empty stderr.

The second: `grep` exits **2** on an unreadable or absent file, which is also indistinguishable from no-match. That is reachable when the `> ${WORK_DIR}/<name>.log` redirect itself fails, which is precisely the disk fault from the `tar` finding.

`classify_failure` now runs four demotions before `security`: exit 124/137, empty log, Sigstore unreachable, then the pattern test, with `[ -r "$1" ] || return 0` at the top of `infra_shaped`. The failed command's exit status is captured as the first statement in each `else` branch, before `cat`, at all three call sites; a comment records that ordering requirement because it is easy to break.

The unreadable-log test is driven by a genuinely reachable path rather than a synthetic one: it pre-seeds the log mode 0000 so the step's own redirect fails, and skips under uid 0 since root bypasses the mode. Mutation proof: reverting all four demotions fails 5 classification rows and 4 load-bearing rows.

.github/workflows/release-reverify.yaml
lockwobr: Confirmed and fixed in d28752c7, and you were right that no test covered list-succeeds/extract-fails. There is now a row driven by a fake `tar` that succeeds on `-tzf` and fails on `-xzf`, plus a mutation row restoring the unguarded form.

One detail worth recording: the fake `cosign` had to be made filesystem-aware first. It previously ignored the bundle path entirely and returned a scripted exit code, so a failed extract would have "passed" and the regression test would have been decorative. It now emits real cosign's `opening bundle: no such file or directory` when the bundle is absent.

.github/workflows/release-reverify.yaml
lockwobr: Agreed, and the distinction you drew from rekor-monitor is the crux: its alert tracks the release-agnostic log, so a clean run genuinely clears it, whereas this job only ever verifies `/releases/latest`. Fixed in d28752c7.

The alert title now carries the tag, and the close loop iterates `${ALERT_TITLE} [${TAG}]` plus the degraded title, so a clean run can only close the alert for the tag it actually verified. The degraded issue keeps global scope deliberately, since it tracks the checker rather than a release.

I also fixed the misleading body wording you flagged. It now says the issue auto-closes only while that tag is still latest, and that once a newer release ships the tag is no longer re-checked and the issue must be closed by hand after remediation.

`TestReleaseReverifyAlertCloseIsTagScoped` pins it: a clean run on v0.19.0 with an open v0.18.0 alert closes only the degraded issue, a same-tag alert closes, and a title merely containing the alert text does not. The mutation subtest strips the tag from the loop and asserts the unscoped form closes the other tag's alert, i.e. reproduces exactly the sequence you described.

.github/workflows/release-reverify.yaml
lockwobr: Maintainer call made: adding Slack. Applied in d28752c7, mirrored from `rekor-monitor.yaml` rather than invented — same `secrets.SLACK_SERVICE` wiring, same `failure() && classification == 'tamper'` gate so degraded never pages, same message shape with the run link and tracking issue, and the same unset-secret no-op with a warning. One deviation: `timeout --foreground 30s` on the `curl`, per this repo's bounded-network-call convention.

Your point about delivery depending on per-maintainer mention settings is the reason I did not take the document-the-divergence option. It also compounds with your L457 finding, since the GitHub issue was the only channel and its create path had no retry; both are fixed in the same commit.

.github/workflows/release-reverify.yaml
lockwobr: Agreed, and the low-probability/high-impact framing is right: this is the one notification the job exists to send. Fixed in d28752c7. Create is a 3-attempt loop with backoff, then a single unlabelled fallback create with a warning, so a renamed or deleted label costs the labels rather than the notification. Total failure emits an error but exits 0 so the Slack step still fires.

While rewriting it I also closed a de-dup gap you did not raise: the step used the search API, whose indexing lag can miss an issue this workflow opened minutes earlier and open a duplicate. It now reads `repos/{repo}/issues?state=open` through `gh-api-retry.sh --paginate` with an exact-title filter and `select(.pull_request == null)`.

.github/workflows/release-reverify.yaml
lockwobr: Agreed, fixed in d28752c7: the list is `|| echo '[]'`, the jq extraction goes into a variable, and the loop is fed from a here-string, so neither a list blip nor a jq failure can abort the success path. I also added `< /dev/null` to the list and to `gh issue close`, since the stdin-isolation problem CodeRabbit found in the SBOM loop applies to this loop too.

.github/workflows/release-reverify.yaml
lockwobr: Fixed in d28752c7: `map(. == "failure") | (index(false) // length)`. I also guarded the `gh api` read itself (`|| true`, streak defaults to 0), since the read failing is part of the same outage being reported.

Recording your own calibration for the reviewer record, because I agree with it: this was never capable of producing a false page. It only over-escalates the calm, deduped, auto-closed no-page path.

.github/workflows/release-reverify.yaml
lockwobr: Valid, and the cancelled-run consequence is the part that matters: a cancelled job skips `if: failure()`, so no degraded issue opens. Fixed in d28752c7 by doing both halves.

Per-call cosign bounds cut 300s to 120s (a bundle verification plus a TUF fetch is seconds; 120s is still roughly 100x headroom, and the tight side degrades to `operational`). Job budget raised to 60 minutes with the arithmetic on the field:

```
binary cosign re-run            120s
catalog download                120s
catalog verify                  120s
4 SBOMs x (120s dl + 120s)      960s
6 x classify_failure probes     840s
subtotal                       2160s = 36m
install-aicr-release            ~10m
checkout + load-versions         ~2m
issue/Slack bookkeeping          ~3m
total                           ~51m
```

A shape test pins `timeout-minutes >= 45` so a casual reduction has to revisit the arithmetic. Residual, flagged rather than fixed here: `install-aicr-release` has no per-call `timeout`, so its ~10m is an estimate rather than a bound. Changing a shared action felt out of scope for this PR.

.github/workflows/release-reverify.yaml
lockwobr: Implemented in d28752c7, though I want to record that I agree with your own assessment that it is not material. `SIGSTORE_PROBE_URL` became `SIGSTORE_PROBE_URLS` and now includes Fulcio; all probes must answer, so a Fulcio-only outage demotes.

Fulcio rather than Rekor, deliberately: Rekor v2 shard hostnames rotate (log2025-1, log2026-1), so a fixed shard URL would start failing permanently after a rotation and, because probes are demote-only, would **silently disable paging**. That is the one risk this change actually introduces, so each probe failure is logged by name and the comment records the hazard. A row using a fake `curl` that fails one URL only pins the behavior.

.github/workflows/release-reverify.yaml
lockwobr: Fixed in d28752c7 to match `sign-sbom.sh`'s canonical fully-escaped form, on the maintainer's call, while noting both you and mchmarny correctly assessed it as not exploitable: `REPO` is pinned, the issuer is pinned, and the Fulcio SAN structure is fixed.

Flagging the consequence you identified: `.github/actions/install-aicr-release/action.yml:80` still has the loose form and is now inconsistent with this workflow. I did not change a shared action in this PR; I will file a follow-up so the two converge.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,732 @@
...
+          # That direction is fail-OPEN (silent under-verification), the opposite
+          # of a false page but still wrong, so it is demoted explicitly.
+          newest=""
+          if ! newest="$(printf '%s\n%s\n' "${SBOM_SIGNING_FLOOR}" "${TAG}" | sort -V | tail -n1)" \
+              || [ -z "${newest}" ]; then
+            operational "could not order ${TAG} against the SBOM signing floor ${SBOM_SIGNING_FLOOR}; treating as infrastructure"
+          elif [ "${TAG}" = "${SBOM_SIGNING_FLOOR}" ] || [ "${newest}" != "${TAG}" ]; then
+            echo "release ${TAG} is at or before the SBOM signing floor ${SBOM_SIGNING_FLOOR}; SBOM bundles not expected"
+          else
+            sboms="$(grep -E '_linux_amd64\.sbom\.json$' "${RELEASE_ASSETS}" || true)"
```
yuanchen8911: The SBOM presence check derives its expected set from the release's own asset inventory: `sboms="$(grep -E '_linux_amd64\.sbom\.json$' "${RELEASE_ASSETS}" || true)"`, and only the zero-SBOMs case is a finding. But `expected_release_asset_names()` in `.github/scripts/release-images.sh` mandates both `aicr_<version>_linux_amd64.sbom.json` and `aicrd_<version>_linux_amd64.sbom.json`. Delete the `aicrd` one (with its bundle) from the release and this loop verifies the surviving `aicr` pair and the run stays clean — a silently-missing mandatory artifact, which is exactly the failure class this job exists to catch. Requiring the two expected names (derived from the tag, as the inventory check in section 1 already does for the archive) closes it.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,732 @@
...
+            if ! tar -xzf "${REL_DIR}/${archive}" -C "${WORK_DIR}" \
+                aicr aicr-attestation.sigstore.json 2>/dev/null; then
+              operational "the archive listed but could not be extracted; treating as infrastructure"
+            elif timeout --foreground 120s cosign verify-blob-attestation \
+                --bundle "${WORK_DIR}/aicr-attestation.sigstore.json" \
+                --type https://slsa.dev/provenance/v1 \
+                --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+                --certificate-identity-regexp "${identity_re}" \
+                "${WORK_DIR}/aicr" > "${WORK_DIR}/binary-verify.log" 2>&1; then
+              operational "binary provenance verified on re-run after the install action failed; treating as a transient failure"
```
yuanchen8911: If `aicr_checksums.txt` stops matching an otherwise valid, attested archive, the composite fails at its `sha256sum -c` (action.yml line 91) with the archive left in `_rel`. This fallback then re-runs only the cosign verification, which passes — the attestation binds the binary, not the checksums manifest — and the run lands here on "treating as a transient failure". Since `./aicr` was never installed, the catalog check also demotes. So a persistent, consumer-visible integrity failure (users following the documented checksum verification fail every time) classifies operational every day, and the degraded issue it eventually opens asserts "no released artifact was shown to be missing or unverifiable", which would be false here. The comment above this branch says the action's log stays authoritative for which sub-check failed — but nothing reads it. Suggest re-running the checksum in this fallback (a few lines against files already on disk) so a mismatch reaches `classify_failure` rather than being presumed transient.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,732 @@
...
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          TAG: ${{ steps.release.outputs.tag }}
+          RELEASE_ASSETS: ${{ steps.release.outputs.assets_file }}
+          INSTALL_OUTCOME: ${{ steps.install.outcome }}
+        run: |
+          # Deliberately not `set -e`: this step must run every check, aggregate
+          # the findings, and exit with the classified status. Each check owns
+          # its own failure handling.
+          set -uo pipefail
```
yuanchen8911, line 278: The comment says "Deliberately not `set -e`", but this step runs under errexit anyway: there is no `shell:` key anywhere in the file, so GitHub invokes the block as `bash -e {0}`, and `set -uo pipefail` does not clear the inherited `-e`. The extraction guard's comment further down asserts "the step runs without `set -e`" on the same false premise. Meanwhile the tests execute the extracted script under plain `bash -c` (no `-e`), so the semantics the classification subtests and the mutation tests prove are not the semantics production runs. With every branch now individually guarded the immediately reachable consequence is narrow (an unguarded top-level failure such as `mkdir` aborts the step, and a `tamper` recorded earlier in the same run is demoted to operational because the classification never reaches `GITHUB_OUTPUT`) — but any future unguarded command added here will pass the tests and abort in production the same way. One line fixes it: an explicit `set +e` after `set -uo pipefail` (as rekor-monitor.yaml line 233 does), which makes the comment true and aligns production with the tested semantics.

.github/workflows/release-reverify.yaml
```diff
@@ -0,0 +1,732 @@
...
+      # against the published checksum, extracts the SLSA provenance bundle,
+      # fails closed when the bundle is absent, and runs
+      # `cosign verify-blob-attestation` pinned to on-tag.yaml at the EXACT tag
+      # with a 3-attempt retry. continue-on-error hands its outcome to the
+      # classifying step below instead of ending the job with an unclassified
+      # red, which would page nobody but also file nothing.
+      - name: Verify the released aicr binary provenance
+        id: install
+        continue-on-error: true
+        uses: ./.github/actions/install-aicr-release
```
yuanchen8911, line 255: The sizing comment now documents this composite's `gh`/`cosign` calls as unbounded with a ~10m estimate — but an estimate is not a bound. A wedged `gh release download` (action.yml line 66) can still run to the 60-minute job timeout, and a cancelled job skips every `if: failure()` step: no empty-classification handling, no issue bookkeeping, and under the new failure-only streak rule the run leaves no trace toward the degraded streak either. Repeated stalls would produce red cancelled runs with no degraded issue ever. Since the composite is shared with the UAT release cells, suggest a follow-up issue to add command-level timeouts inside the action (per the repo's I/O timeout rule) with its own UAT validation, rather than fixing it in this PR.

docs/contributor/maintaining.md
```diff
@@ -268,6 +268,113 @@ monitoring-coverage gap, not a verification gap. A follow-up may add a one-time
...
+  attestations API and proves nothing about the registry copy's retrievability.
+- **Rekor entry liveness by log index.** `cosign verify-blob-attestation`
+  verifies a self-contained bundle: the inclusion proof and RFC3161 timestamp
+  travel inside it and are checked against the live Sigstore trust root. A pass
+  proves the bundle is retrievable and cryptographically sound, not that Rekor
+  would still serve that entry by index.
+
+Triage is in the workflow file's header comment. In short: a security issue names
+the exact artifact — an absent asset means re-upload and re-sign (or re-cut the
+release), while an asset that is present but fails verification means the
```
yuanchen8911: "the published bytes do not match what the release signed, and is an incident" overstates what a present-asset verification failure proves: `cosign verify-blob-attestation` also fails on a certificate-identity or predicate-type mismatch with fully intact bytes (for example an asset re-signed under a different workflow identity). Suggest directing the responder to the verifier's reported failure reason first, before treating it as byte tampering.

docs/contributor/maintaining.md
```diff
@@ -268,6 +268,113 @@ monitoring-coverage gap, not a verification gap. A follow-up may add a one-time
 new-shard backfill; until then, treat a rotation log line as a prompt to
 spot-check releases made around the rotation boundary.
 
+### Daily release re-verification
```
yuanchen8911, line 271: This section describes only the daily schedule, but the workflow also declares `workflow_dispatch` — and the workflow header explains why a maintainer would use it (re-arming the schedule after 60 days of repository inactivity, and an on-demand run right after a release). Worth a sentence here, since this is the page a maintainer will actually read.

tests/releasepolicy/release_reverify_test.go
```diff
@@ -0,0 +1,1473 @@
...
+			code: 0,
+		},
+
+		// --- security: positive evidence that a published artifact is absent ---
+		{
+			name: "a missing catalog signature is a security finding",
+			opts: reverifyOptions{
+				tag: reverifySBOMFloor,
+				assets: []string{
+					"aicr_0.18.0_linux_amd64.tar.gz",
```
yuanchen8911: The two clean cases now derive their asset names from `reverifySBOMFloor`, but the security and operational fixtures still hardcode `0.18.0` (here and in the `assetsFor("0.18.0")` cases below) while their tag comes from the constant. When the floor is raised the two desynchronize — and the failure mode is uneven in the wrong direction: clean and operational cases fail loudly (the derived tag's archive is now "missing"), but because `tamper` is sticky, security cases keep passing on that same missing-archive finding rather than on the condition they were written to test. Deriving all fixtures from the constant keeps every case exercising its intended path. Low-severity test maintenance, no workflow change implied.

tests/releasepolicy/release_reverify_test.go
```diff
@@ -0,0 +1,1473 @@
...
+		name string
+		opts reverifyOptions
+		want string
+		code int
+	}{
+		{
+			name: "a complete release verifies clean",
+			opts: reverifyOptions{
+				tag:            reverifySBOMFloor,
+				assets:         assetsFor(strings.TrimPrefix(reverifySBOMFloor, "v")),
```
yuanchen8911: This at-floor clean case ships no SBOM assets at all, so it proves the version gate fires but not the gate's behavior on the actual supported release shape: the real `v0.18.0` ships unsigned `aicr` and `aicrd` linux/amd64 SBOMs with no `.sigstore.json` siblings. Every case that does include an unsigned `.sbom.json` is above-floor and expects `tamper`, so a regression that made the exemption stop working for releases that actually ship unsigned SBOMs would not be caught. Adding both SBOM subjects (without bundles) to this case covers it. Test coverage only — not evidence the current floor logic is broken.

yuanchen8911 (changes requested):
Review at d28752c. The classification hardening in the latest revision — exit-status capture with 124/137 demotion, the empty/unreadable-log guard, the widened probe set, the guarded extraction, and the honest job-timeout sizing — resolves the largest false-page risks, and the demote-only design with mutation-tested guards remains genuinely solid work. Three issues remain that undercut guarantees this PR states as its own design constraints, so requesting changes rather than approving.

Blocking:

1. The SBOM presence check derives its expected set from the release's own asset inventory (comment at the `sboms=` line). `expected_release_asset_names()` mandates both `aicr` and `aicrd` linux/amd64 SBOMs; delete the `aicrd` one and the run stays clean. A silently-missing mandatory artifact is the exact failure class #1461 asks this job to catch.

2. A persistent `aicr_checksums.txt` mismatch is permanently classified operational (comment on the fallback branch). The fallback re-runs only cosign, which passes — the attestation binds the binary, not the checksums manifest — so a consumer-visible integrity failure never reaches the security gate, and the degraded issue's body would assert something false.

3. The classify step actually runs under errexit, contradicting its own comment and its tests (comment at `set -uo pipefail`). No `shell:` key exists, so GitHub invokes the block as `bash -e {0}`; the tests execute the extracted script under plain `bash -c`, so the semantics the 30+ subtests and the mutation tests prove are not the semantics production runs. The fix is one line.

Suggested as a follow-up issue rather than this PR: the reused `install-aicr-release` composite's network calls are unbounded (the sizing comment now says so honestly, but its ~10m figure is an estimate, not a bound), so a wedged download can still run to the 60-minute cancellation, which skips every `if: failure()` step and — under the new failure-only streak rule — leaves no trace toward the degraded streak either. Real gap, but the composite is shared with the UAT release cells, so the fix wants its own validation there.

Also inline, take or leave: a triage overstatement and the missing `workflow_dispatch` mention in maintaining.md, hardcoded `0.18.0` asset names in several test fixtures, and the floor exemption being untested for the actual v0.18.0 release shape (unsigned SBOMs present).

Open questions, non-blocking: `infra_shaped`'s bare `tuf` / `trusted root` substrings are broad enough that routine cosign trust-root chatter in a genuine mismatch log could demote a real finding (demote-direction only, but worth tightening while in `classify_failure`); and the maintaining.md follow-up note on `gh attestation verify --bundle-from-oci` may want a minimum `gh` version once known.

.github/workflows/release-reverify.yaml
lockwobr: You are right on every point, and this is the most valuable comment on the PR: it invalidated the semantics every prior mutation proof rested on. Fixed in 02b1dad0.

Confirmed before applying:
```
bash -e -c 'set -uo pipefail; false; echo REACHED'          -> rc=1, REACHED not printed
bash -e -c 'set -uo pipefail; set +e; false; echo REACHED'  -> REACHED, rc=0
grep -c 'shell:' release-reverify.yaml                      -> 0
```
And `rekor-monitor.yaml:231-232` does exactly `set -uo pipefail` then `set +e`.

Applied the one-line fix and corrected both comments that repeated the false premise. More importantly I closed the harness gap so this cannot recur: the tests now write the extracted block to a file and run `bash -e <file>`, which is GitHub's documented default, rather than `bash -c` with no `-e`. The script's own `set +e` is now what disables errexit in both places, so the harness proves the step does that rather than assuming it.

Rows whose result changed once the two matched: **zero**. I want to be plain that this was luck rather than design, which is exactly your point. Proof the gap is closed: removing `set +e` and adding one unguarded `grep -q nonexistent /dev/null` now fails with `classification = "" (exit 1), want "clean"` — the precise production failure you described, where the step aborts, nothing reaches GITHUB_OUTPUT, and a finding recorded earlier in the run silently demotes. That same code passed under the old harness. I also strengthened the shape assertion from the bare string `set +e` to the two-line sequence, after finding the bare string matched a comment that merely mentioned it.

.github/workflows/release-reverify.yaml
lockwobr: Correct, and it is the same derive-from-the-artifact-under-test flaw that shipped the release-asset allowlist bug in #1982, so I treated it as a real defect rather than a nit. Fixed in 02b1dad0.

New workflow env `EXPECTED_SBOM_BINARIES: "aicr aicrd"`, with a comment naming `expected_release_asset_names()` as authoritative. The loop builds `<binary>_<version>_linux_amd64.sbom.json` from the tag and adds a missing-subject branch ahead of the missing-bundle one.

I kept the here-string `while` loop rather than switching to a `for`, deliberately: a `for` loop would make CodeRabbit's earlier `< /dev/null` fix and its stdin-isolation test vacuous.

Mutation proof in `TestReleaseReverifyMandatorySBOMSetIsDerivedFromTheTag`: reverting to the inventory grep makes a release missing `aicrd`'s SBOM report `clean`/exit 0 without mentioning the dropped subject; intact it reports `tamper`/exit 1 naming it. This also caught a pre-existing fixture that shipped only the `aicr` pair and had been asserting something untrue; corrected rather than accommodated.

.github/workflows/release-reverify.yaml
lockwobr: Agreed, and the detail that the degraded issue would assert something false in that state is what made me treat this as more than a classification nit. Fixed in 02b1dad0: the fallback now re-runs both sub-checks.

Ordered after cosign deliberately, and the comment says why: a truncated download would have failed to list, or produced a binary the attestation rejects, further up the chain, so reaching the checksum re-check means the bytes are intact and a mismatch is the manifest disagreeing with the archive.

On the stale comment you flagged: I made the comment true rather than making the code read the action's log, and said so explicitly in it. Nothing here parses that log, so reproducing the checks is the only way to know which one failed.

The mutation proof needed an inverted direction, since removing this guard hides a finding rather than manufacturing one, so `TestReleaseReverifyGuardsAreLoadBearing` gained a `wantIntactTamper` field: intact `tamper`, mutated `operational`. The tests also needed a `sha256sum` fake, since macOS ships only `shasum -a 256` and without it a missing binary would have read as a checksum failure on this host.

.github/workflows/release-reverify.yaml
lockwobr: Agreed, and I have taken the follow-up route you suggest rather than changing a shared action from a consumer's PR: [#2084](https://github.com/NVIDIA/aicr/issues/2084), covering both the missing command-level timeouts and the looser identity regexp at `action.yml:80` that njhensley raised.

Your framing is sharper than mine was, and I have carried it into the issue: a wedged `gh release download` runs to the job timeout, a cancelled job skips every `if: failure()` step, and under the new failure-only streak rule the run leaves no trace toward the degraded streak either. So repeated stalls would produce red cancelled runs with no degraded issue ever, which is the invisible-failure case rather than merely a slow one.

docs/contributor/maintaining.md
lockwobr: Correct, and the distinction matters for the responder's first move. Fixed in 02b1dad0. The runbook now directs them to the verifier's reported reason first, and splits the two cases: a digest mismatch is the incident path, while an identity or predicate mismatch with intact bytes is a signing-path problem with different remediation, for example an asset re-signed under a different workflow identity.

docs/contributor/maintaining.md
lockwobr: Added in 02b1dad0. The section now covers `workflow_dispatch` and the two reasons a maintainer reaches for it: re-arming after GitHub's 60-day inactivity disable, and getting an on-demand run straight after cutting a release rather than waiting for the next cron. You are right that this is the page a maintainer actually reads.

tests/releasepolicy/release_reverify_test.go
lockwobr: Agreed, and the uneven failure direction is the part that would have bitten. Fixed in 02b1dad0: every fixture now derives from `reverifySBOMFloor`, including the shared setup in `TestReleaseReverifyGuardsAreLoadBearing`, and `sbomSubjects()` reads `EXPECTED_SBOM_BINARIES` from the workflow so fixtures cannot drift from what the workflow mandates either. No hardcoded `0.18.0` remains.

tests/releasepolicy/release_reverify_test.go
lockwobr: Agreed, and it is a real coverage gap rather than a cosmetic one: every case carrying an unsigned `.sbom.json` was above-floor and expected `tamper`, so a regression breaking the exemption for releases that actually ship unsigned SBOMs would have gone uncaught. Fixed in 02b1dad0. The case is renamed `a release at the floor with unsigned SBOMs verifies clean` and now ships `aicr` and `aicrd` linux/amd64 `.sbom.json` with no `.sigstore.json` siblings, which is the shape real v0.18.0 has.

yuanchen8911 (approved):
all addressed, thanks. /lgtm

PR #2071, opened 2026-08-05
fix(evidence): bound on-disk bundle reads with FileReadTimeout ctx

### LINKED ISSUE #2054: evidence publish: bound loadOnDiskBundle file reads with FileReadTimeout context
## Summary

`aicr evidence publish` reads a user-supplied on-disk bundle directory with no context/timeout bound on the file I/O, so a bundle path on a hung network mount (NFS/FUSE) blocks the command indefinitely before any sign/push timeout applies.

## Mechanism

`loadOnDiskBundle` in `pkg/evidence/attestation/publish.go` reads bundle files through three helpers — `readBundleRecipeProfile`, `readBundlePredicate`, and `HasBundleMarkers` — using bare `os.Open` / `io.ReadAll` / `os.Stat` with no context or deadline. The command's context is not consulted until `signAndPush`, after all local reads complete.

This deviates from the repo's I/O rule (timeout-bounded contexts on all I/O) and the established file-read pattern: `defaults.FileReadTimeout`, as used by `pkg/recipe/adapter.go` and similar readers.

## What it is not

- Not a size/memory issue: both `io.ReadAll` calls are already capped via `io.LimitReader` at `MaxRecipePOSTBytes`.
- Not introduced by PR #2044: all three readers exist on `main` with identical I/O structure (verified by diff against `origin/main` during that PR's review); #2044 only changed return values and added the coherence check after the reads.

## Proposed fix

Wrap the bundle reads in the `FileReadTimeout` context pattern (`context.WithTimeout(ctx, defaults.FileReadTimeout)` around the read, per `pkg/recipe/adapter.go`), so a dead mount surfaces as a timeout error instead of an indefinite hang. Plumb the caller's `ctx` into `loadOnDiskBundle` and its helpers.

## Origin

Surfaced by cross-review during PR #2044; classified pre-existing (follow-up, not in-PR).

njhensley:
## Summary

Bounds `aicr evidence publish`/`sign` on-disk summary-bundle reads with the caller context plus `defaults.FileReadTimeout`, so a hung NFS/FUSE mount surfaces as `ErrCodeTimeout` instead of hanging the command forever.

## Motivation / Context

The evidence publish/sign path read the on-disk summary bundle with unbounded blocking syscalls. On a wedged network/FUSE mount those reads block indefinitely, and the caller's context deadline had no effect — the command simply hung with no error. This bounds every on-disk summary-bundle read so the operation fails closed with a timeout error.

Fixes: #2054
Related: #2083 (follow-up: bound the residual reads on these paths)

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: Evidence attestation + verifier (`pkg/evidence/attestation`, `pkg/evidence/verifier`)

## Implementation Notes

- **Bound the real read, not a preflight probe.** We wrap each blocking read (marker detection and the statement/recipe file reads in `publish`, `sign`, and the verifier `fetch` path) with the caller context and `defaults.FileReadTimeout`, rather than adding a `stat` preflight. A `stat` can succeed while the subsequent `read` wedges on the same hung mount, so only bounding the actual read is correct.
- **Goroutine + `select` mechanism.** Each blocking read runs in a goroutine whose result is delivered on a buffered channel consumed via `select` against `ctx.Done()`. When the context deadline fires first, the call returns `ErrCodeTimeout`. The tradeoff is one short-lived goroutine per read that may outlive the timeout on a truly stuck mount; it is detached and harmless (no shared mutable state, result discarded), and this is the standard pattern for making an otherwise-uninterruptible syscall respect a context deadline. Both human reviewers confirmed this is fine for a short-lived CLI leaf command.
- **Signature change:** `HasBundleMarkers(dir string) bool` becomes `HasBundleMarkers(ctx context.Context, dir string) (bool, error)` so marker detection is itself bounded and reports a **timeout** (not a bad-input answer) when the mount is wedged. All six call sites were updated to thread the context and check `err` before `ok`. Non-timeout stat errors still read as "not a marker" (fail-soft, unchanged from before this PR).

## Testing

```bash
make qualify
golangci-lint run -c .golangci.yaml ./pkg/evidence/attestation/... ./pkg/evidence/verifier/...
go test -race ./pkg/evidence/attestation/... ./pkg/evidence/verifier/...
```

`make qualify` passes; lint clean on both touched packages. Table-driven timeout tests in `publish_test.go`, `sign_test.go`, and `fetch_test.go` cover the canceled-context path, plus a dedicated `TestWithFileReadTimeout_InFlightTimeout` that exercises the goroutine+`select` arm deterministically (fn blocks past an in-flight cancel via a started/release/finished handshake, then is joined so no worker leaks). Coverage is healthy on all touched functions.

## Known Limitations (tracked in #2083)

This PR bounds the on-disk **summary-bundle** reads, which is the scope of #2054. Cross-review surfaced adjacent reads on the same code paths that remain unbounded or mis-diagnosed and are deferred to #2083 to keep this change minimal:

- `verify`'s `DetectInputForm` stats the input dir *before* the bounded probe, and several post-materialize verifier reads are still unbounded.
- The OCI push-staging walk (`pkg/oci`) is only ctx-checked between entries, so `publish` can still hang one step later on a fully dead mount.
- `verify`'s `classifyFailure` has no transient branch, so a timeout currently classifies identically to a genuine "invalid attestation".
- Non-`ENOENT` stat errors (EIO/ESTALE) read as "not a marker" rather than surfacing the storage fault distinctly.

The doc comments and this description are scoped to what is actually bounded here; the follow-up will close the residual end-to-end.

## Risk Assessment

**Rollout notes:** No user-facing surface change. `HasBundleMarkers` is an internal helper; its signature change is contained to `pkg/evidence` and all call sites are updated in this PR. Backwards compatible for CLI users — the only behavioral change is that a previously-hanging summary-bundle read now returns a timeout error.

## Checklist

### DIFF
--- a/pkg/evidence/attestation/publish.go
+++ b/pkg/evidence/attestation/publish.go
@@ -93,7 +93,7 @@ func Publish(ctx context.Context, opts PublishOptions) error {
 		return errors.Wrap(errors.ErrCodeInvalidRequest, "invalid push reference", err)
 	}
 
-	bundle, outDir, err := loadOnDiskBundle(opts.BundleDir)
+	bundle, outDir, err := loadOnDiskBundle(ctx, opts.BundleDir)
 	if err != nil {
 		return err
 	}
@@ -143,15 +143,15 @@ func Publish(ctx context.Context, opts PublishOptions) error {
 // unsigned in-toto Statement. The predicate is trusted as-is — Publish
 // signs the pre-built bundle bytes verbatim; integrity auditing is the
 // job of `aicr evidence verify`.
-func loadOnDiskBundle(dir string) (*Bundle, string, error) {
+func loadOnDiskBundle(ctx context.Context, dir string) (*Bundle, string, error) {
 	if dir == "" {
 		return nil, "", errors.New(errors.ErrCodeInvalidRequest, "bundle directory is required")
 	}
-	summaryDir, outDir, err := resolveSummaryDir(dir)
+	summaryDir, outDir, err := resolveSummaryDir(ctx, dir)
 	if err != nil {
 		return nil, "", err
 	}
-	pred, stmt, err := readBundlePredicate(summaryDir)
+	pred, stmt, err := readBundlePredicate(ctx, summaryDir)
 	if err != nil {
 		return nil, "", err
 	}
@@ -159,7 +159,7 @@ func loadOnDiskBundle(dir string) (*Bundle, string, error) {
 		return nil, "", errors.New(errors.ErrCodeInvalidRequest,
 			"bundle statement predicate is missing recipe.{name,digest}")
 	}
-	profile, advertiser, descriptorIdentity, err := readBundleRecipeProfile(summaryDir)
+	profile, advertiser, descriptorIdentity, err := readBundleRecipeProfile(ctx, summaryDir)
 	if err != nil {
 		return nil, "", err
 	}
@@ -184,6 +184,40 @@ func loadOnDiskBundle(dir string) (*Bundle, string, error) {
 	return bundle, outDir, nil
 }
 
+// withFileReadTimeout runs fn — a blocking local-filesystem syscall (os.Open /
+// io.ReadAll / os.Stat) — under a FileReadTimeout-bounded derivative of ctx,
+// returning fn's own error, or an ErrCodeTimeout error if the bound elapses
+// first. It exists because a bare os.Open/os.Stat against a hung NFS/FUSE mount
+// blocks in the syscall indefinitely: no downstream sign/push timeout can
+// unblock it, so the read itself must be time-bounded to surface a dead mount
+// as a timeout rather than an indefinite hang.
+//
+// fn runs in a goroutine so the caller unblocks the instant ctx expires. The
+// accepted tradeoff: on timeout the parked goroutine stays blocked in the
+// syscall until the kernel returns or the process exits — fine for a
+// short-lived CLI leaf command (evidence publish/sign), which is the only
+// caller of these bundle readers.
+func withFileReadTimeout(ctx context.Context, what string, fn func() error) error {
+	ctx, cancel := context.WithTimeout(ctx, defaults.FileReadTimeout)
+	defer cancel()
+	// Fail closed and deterministically if the caller's context is already
+	// canceled/expired, before launching the goroutine — otherwise a fast
+	// syscall could win the select race against an already-done context.
+	if err := ctx.Err(); err != nil {
+		return errors.WrapWithContext(errors.ErrCodeTimeout, "timed out reading "+what, err,
+			map[string]interface{}{"timeout": defaults.FileReadTimeout.String()})
+	}
+	done := make(chan error, 1) // buffered so the goroutine never leaks on timeout
+	go func() { done <- fn() }()
+	select {
+	case <-ctx.Done():
+		return errors.WrapWithContext(errors.ErrCodeTimeout, "timed out reading "+what, ctx.Err(),
+			map[string]interface{}{"timeout": defaults.FileReadTimeout.String()})
+	case err := <-done:
+		return err
+	}
+}
+
 // readBundleRecipeProfile decodes the summary bundle's recipe.yaml and
 // returns the canonical name=value selection it carries, its declared
 // advertiser, and the recipe-scoped policy-descriptor identity recomputed
@@ -194,17 +228,25 @@ func loadOnDiskBundle(dir string) (*Bundle, string, error) {
 // carries only the (suffixed) recipe name. The read is size-bounded like
 // readBundlePredicate: the publish target may be an attacker-influenced
 // bundle root where os.ReadFile would allocate the whole file before any
-// size check.
-func readBundleRecipeProfile(summaryDir string) (profile, advertiser, descriptorIdentity string, err error) {
+// size check. It is also time-bounded via withFileReadTimeout so the read
+// cannot hang indefinitely on a dead NFS/FUSE mount.
+func readBundleRecipeProfile(ctx context.Context, summaryDir string) (profile, advertiser, descriptorIdentity string, err error) {
 	path := filepath.Join(summaryDir, RecipeFilename)
-	f, err := os.Open(path) //nolint:gosec // bundle-local path resolved by resolveSummaryDir
-	if err != nil {
-		return "", "", "", errors.Wrap(errors.ErrCodeNotFound, "failed to read bundle recipe.yaml", err)
-	}
-	defer func() { _ = f.Close() }()
-	body, err := io.ReadAll(io.LimitReader(f, defaults.MaxRecipePOSTBytes+1))
-	if err != nil {
-		return "", "", "", errors.Wrap(errors.ErrCodeInternal, "failed to read bundle recipe.yaml", err)
+	var body []byte
+	if rerr := withFileReadTimeout(ctx, "bundle recipe.yaml", func() error {
+		f, oErr := os.Open(path) //nolint:gosec // bundle-local path resolved by resolveSummaryDir
+		if oErr != nil {
+			return errors.Wrap(errors.ErrCodeNotFound, "failed to read bundle recipe.yaml", oErr)
+		}
+		defer func() { _ = f.Close() }()
+		b, rErr := io.ReadAll(io.LimitReader(f, defaults.MaxRecipePOSTBytes+1))
+		if rErr != nil {
+			return errors.Wrap(errors.ErrCodeInternal, "failed to read bundle recipe.yaml", rErr)
+		}
+		body = b
+		return nil
+	}); rerr != nil {
+		return "", "", "", rerr
 	}
 	if int64(len(body)) > defaults.MaxRecipePOSTBytes {
 		return "", "", "", errors.New(errors.ErrCodeInvalidRequest,
@@ -226,13 +268,21 @@ func readBundleRecipeProfile(summaryDir string) (profile, advertiser, descriptor
 // pointer lands in its parent so the on-disk layout matches the one-shot
 // `validate --emit-attestation --push` output (pointer.yaml beside
 // summary-bundle/).
-func resolveSummaryDir(dir string) (summaryDir, outDir string, err error) {
+func resolveSummaryDir(ctx context.Context, dir string) (summaryDir, outDir string, err error) {
 	clean := filepath.Clean(dir)
-	if HasBundleMarkers(clean) {
+	ok, err := HasBundleMarkers(ctx, clean)
+	if err != nil {
+		return "", "", err
+	}
+	if ok {
 		return clean, filepath.Dir(clean), nil
 	}
 	candidate := filepath.Join(clean, SummaryBundleDirName)
-	if HasBundleMarkers(candidate) {
+	ok, err = HasBundleMarkers(ctx, candidate)
+	if err != nil {
+		return "", "", err
+	}
+	if ok {
 		return candidate, clean, nil
 	}
 	return "", "", errors.New(errors.ErrCodeInvalidRequest,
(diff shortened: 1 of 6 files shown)


pkg/evidence/verifier/fetch.go
```diff
@@ -114,12 +114,22 @@ func MaterializeBundle(
 // materializeDir accepts either the summary-bundle root or a parent
 // containing it. Bundles are recognized by recipe.yaml + manifest.json
 // at the candidate root.
-func materializeDir(input string) (*MaterializedBundle, error) {
-	if attestation.HasBundleMarkers(input) {
+func materializeDir(ctx context.Context, input string) (*MaterializedBundle, error) {
+	ok, err := attestation.HasBundleMarkers(ctx, input)
```
lockwobr, line 118: The verifier side of this change does not deliver the hang-immunity it reads as delivering: an unbounded stat on the same operator-supplied path runs *before* this probe.

`Verify` calls `DetectInputForm(opts.Input)` at `verify.go:56`, and `DetectInputForm` does a bare, context-free `os.Stat(input)` at `input.go:41` — on the exact directory `HasBundleMarkers` stats here at `verify.go:76`. So `aicr evidence verify --input /mnt/hung/bundle` still blocks indefinitely in `input.go:41` and never reaches the bounded probe.

The new `TestBundleMarkerProbe_CanceledCtxPropagatesTimeout` calls `materializeDir`/`resolveBundleDir` directly, bypassing `DetectInputForm`, which is why the gap is invisible to it. The publish path is genuinely covered (`runEvidencePublishCmd` touches nothing before `attestation.Publish`); this is verifier-only.

Post-materialize reads of the same directory are also still unbounded: `bounded_read.go:31`, `inventory.go:178`, `signature.go:72`. Either bound `DetectInputForm` too, or narrow the claim in the `fetch_test.go` doc comment that "a hung NFS/FUSE mount must fail closed" — end-to-end for `verify`, it currently does not.

pkg/evidence/attestation/publish.go
```diff
@@ -246,33 +296,59 @@ func resolveSummaryDir(dir string) (summaryDir, outDir string, err error) {
...
-		info, statErr := os.Stat(filepath.Join(dir, f))
-		if statErr != nil || info.IsDir() {
-			return false
+		path := filepath.Join(dir, f)
+		var present bool
+		if err := withFileReadTimeout(ctx, "bundle marker "+f, func() error {
+			info, statErr := os.Stat(path)
+			// Any stat error (absent, permission, is-a-dir) means "not a
+			// marker"; only the timeout wrapper injects a non-nil error.
+			present = statErr == nil && !info.IsDir()
```
lockwobr, line 320: The godoc immediately above (lines 302-305) says the `(bool, error)` shape keeps a genuine "not a bundle" answer "distinct from a ctx timeout / **unexpected failure** (→ `(false, err)`)". This line does the opposite: `statErr` is discarded entirely and the closure unconditionally returns `nil`, so EACCES, EIO, ESTALE and ELOOP all yield `(false, nil)`. Only the timeout wrapper can produce a non-nil error — which is exactly what the inline comment at 312-313 says, so the two comments contradict each other.

Practical effect: a soft-mounted NFS or a container-runtime mount that returns ESTALE/EIO rather than hanging is reported to the operator as `INVALID_REQUEST "directory X does not look like a summary bundle"` from `resolveSummaryDir` (and the same from `materializeDir`). That is a user-error diagnostic for an I/O fault, and it is precisely the conflation the godoc claims the new signature prevents.

It fails *closed* — no bad bundle is accepted — so this is diagnostic quality plus a doc that overpromises, not a security hole. Two ways out: return `(false, err)` for stat errors that are not `fs.ErrNotExist`, or correct the godoc to match the deliberate fail-soft behavior.

pkg/evidence/verifier/fetch.go
```diff
@@ -114,12 +114,22 @@ func MaterializeBundle(
...
+	if err != nil {
+		return nil, errors.PropagateOrWrap(err, errors.ErrCodeInternal,
```
lockwobr, line 120: This `PropagateOrWrap` message is unreachable. `HasBundleMarkers` returns either `(false, nil)` or the error from `withFileReadTimeout`, which is always `errors.WrapWithContext(...)` — i.e. always a `*StructuredError`. `PropagateOrWrap` (`pkg/errors/errors.go:154-157`) returns `err` verbatim whenever `stderrors.As` finds one, so `"failed to probe bundle markers at "+input` is never attached — here or at lines 129, 409 and 418.

The message the operator actually sees comes from `publish.go:310` and names only the bare filename, with no directory at all (see the comment there). Since these four sites probe two different candidate directories, there is no way to tell which path stalled.

The cheapest fix is at `publish.go:310` — include `path` in the label. If you want the caller-side context too, it needs `errors.WrapWithContext` rather than `PropagateOrWrap`, at the cost of re-coding the error.

pkg/evidence/attestation/publish.go
```diff
@@ -246,33 +296,59 @@ func resolveSummaryDir(dir string) (summaryDir, outDir string, err error) {
...
+// a ctx timeout / unexpected failure (→ (false, err)): conflating the two
+// would let a hung mount masquerade as "not a bundle" and fail open.
+func HasBundleMarkers(ctx context.Context, dir string) (bool, error) {
 	for _, f := range []string{RecipeFilename, ManifestFilename} {
-		info, statErr := os.Stat(filepath.Join(dir, f))
-		if statErr != nil || info.IsDir() {
-			return false
+		path := filepath.Join(dir, f)
+		var present bool
+		if err := withFileReadTimeout(ctx, "bundle marker "+f, func() error {
```
lockwobr: The label drops the directory: `path` is computed one line above at 308, but only `f` (the bare filename) reaches the message. On a hung mount the operator gets:

```
[TIMEOUT] timed out reading bundle marker recipe.yaml: context deadline exceeded
```

with no indication of *which* directory stalled — and the four probe sites in `verifier/fetch.go` each try two candidates (the input dir and its nested `summary-bundle/`). The `Context` map does not close the gap either: `StructuredError.Error()` renders only Code/Message/Cause, never Context, and Context here holds only `{"timeout": "30s"}`.

`"bundle marker "+path` would fix it. This is also the only surviving message, because the caller-side `"failed to probe bundle markers at "+dir` in `verifier/fetch.go` is unreachable — see the comment there.

pkg/evidence/attestation/publish.go
```diff
@@ -184,6 +184,40 @@ func loadOnDiskBundle(dir string) (*Bundle, string, error) {
 	return bundle, outDir, nil
 }
 
+// withFileReadTimeout runs fn — a blocking local-filesystem syscall (os.Open /
```
lockwobr, line 187: This doc comment's rationale — "no downstream sign/push timeout can unblock" a syscall wedged on a hung mount — is right, and it is also the reason the mitigation is incomplete: it stops at `loadOnDiskBundle`, and the rest of `evidence publish` re-reads the same directory with exactly the pattern this comment argues is insufficient.

After `loadOnDiskBundle` returns, `Publish` continues into `signAndPush` → `Push(pushCtx, PushOptions{SourceDir: bundle.SummaryDir})` (`emit.go:306`) → `oci.PackageAndPush` → `prepareSource`, which walks that same directory (`pkg/oci/source_files.go:955`, `fs.WalkDir`) and opens and stats every file (`source_files.go:1065`, `:1076`). Those are plain blocking syscalls; the only protection is `contextError(ctx, ...)` **between** entries at `:1048` and `:1062`.

So for #2054's scenario — bundle dir on a dead mount — `aicr evidence publish` no longer hangs in the predicate/recipe read but still hangs indefinitely in the push-staging walk of the same directory. Operators get the same unbounded hang, one step later. This is an incomplete fix rather than a regression, and it may be fine to scope out, but the PR reads as closing the hang for `publish` and it does not. `sign` is unaffected (it reads from a pulled temp dir).

Worth either extending the bound to the staging walk or noting the residual hang explicitly in the issue.

pkg/evidence/attestation/publish.go
```diff
@@ -184,6 +184,40 @@ func loadOnDiskBundle(dir string) (*Bundle, string, error) {
...
+// short-lived CLI leaf command (evidence publish/sign), which is the only
+// caller of these bundle readers.
+func withFileReadTimeout(ctx context.Context, what string, fn func() error) error {
+	ctx, cancel := context.WithTimeout(ctx, defaults.FileReadTimeout)
+	defer cancel()
+	// Fail closed and deterministically if the caller's context is already
+	// canceled/expired, before launching the goroutine — otherwise a fast
+	// syscall could win the select race against an already-done context.
+	if err := ctx.Err(); err != nil {
+		return errors.WrapWithContext(errors.ErrCodeTimeout, "timed out reading "+what, err,
```
lockwobr, line 207: This wraps whatever `ctx.Err()` happens to be — including `context.Canceled` — as `ErrCodeTimeout` with `"timed out reading "+what` and a `timeout: 30s` context value that never elapsed. Same at line 214.

Two consequences when an operator hits Ctrl-C during `evidence publish`:

- They see `[TIMEOUT] timed out reading in-toto Statement: context canceled`, which reads as a wedged mount rather than their own interrupt.
- `errors.IsTransient` (`pkg/errors/errors.go:139-145`) returns true for `ErrCodeTimeout`, so any caller retrying on transience would retry an operation the operator deliberately aborted. No such retry loop is reachable from here today, but the classification is wrong regardless.

Branching on `stderrors.Is(ctx.Err(), context.Canceled)` and propagating a cancellation-shaped error fixes both, and would also stop the `timeout` context value from asserting a bound that did not fire.

pkg/evidence/attestation/publish_test.go
```diff
@@ -65,6 +65,106 @@ func wantInvalidRequest(t *testing.T, err error) {
...
+			return err
+		}},
+		{"loadOnDiskBundle", func(ctx context.Context) error {
+			_, _, err := loadOnDiskBundle(ctx, dir)
+			return err
+		}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			ctx, cancel := context.WithCancel(context.Background())
```
lockwobr, line 115: This construction means the goroutine-plus-`select` mechanism — the part that actually delivers hang-immunity — is never executed by any test in this PR.

`context.WithTimeout(parent, d)` cancels the child synchronously when the parent is already canceled, so an immediately-canceled context always short-circuits at the `ctx.Err()` pre-check (`publish.go:206-209`) and returns before `go func() { done <- fn() }()` at `:211` and the `select` at `:212-218`. Every new test uses this same `WithCancel` + immediate `cancel()` shape — all five subtests here, `TestSignExisting_CanceledContextSurfacesTimeout`, and both subtests of `TestBundleMarkerProbe_CanceledCtxPropagatesTimeout`. A grep for `WithTimeout` across these test files finds nothing.

So `publish.go:213-215` has zero coverage, while the doc comment above claims these tests "prove each on-disk reader path is bounded by the caller's context". Concretely: a regression that reverted `withFileReadTimeout` to a direct inline `fn()` call would still pass every test added here.

The gap is cheap to close without a real hung mount — a context with a short but *live* deadline (say 20ms) plus an `fn` that blocks on a channel past it exercises the real path deterministically.

pkg/evidence/verifier/fetch_test.go
```diff
@@ -46,6 +49,73 @@ func TestMaterializeBundle_DirRejectsNonBundle(t *testing.T) {
...
+			cancel()
+
+			err := tt.call(ctx)
+			if err == nil {
+				t.Fatalf("%s: expected error on canceled ctx, got nil", tt.name)
+			}
+			if !stderrors.Is(err, errors.New(errors.ErrCodeTimeout, "")) {
+				t.Errorf("%s: got %v, want ErrCodeTimeout (must not flatten to \"not a bundle\")", tt.name, err)
+			}
+			if stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, "")) {
```
lockwobr, line 101: This assertion holds at the unit level but is erased at the command boundary, so the property it guards does not survive to the operator.

`Verify` records a `MaterializeBundle` error, passes it to `setFailureCause`, then sets `r.Exit = ExitInvalid` and returns `nil` (`verify.go:76-81`) — the coded error is discarded, never returned. `classifyFailure` (`failure.go:43-57`) has no timeout or transient branch, so a `stepMaterialize` error with no registry status falls through to `CauseUnknown`. The CLI's `default:` arm then maps `ExitInvalid` to `errors.New(errors.ErrCodeInvalidRequest, ...)` at `pkg/cli/evidence_verify.go:151-159` — the exact outcome this line asserts against.

So `aicr evidence verify --input /mnt/hung/bundle` exits 2 with `failureCause.class = "unknown"`, byte-identical in exit code and class to a genuine "this attestation is invalid" verdict. A CI gate or policy engine keying on either cannot distinguish a dead mount from an attestation-integrity failure. The timeout text does survive in the step detail and JSON, so this is a misclassification rather than a total loss.

`errors.IsTransient` already exists for exactly this bucketing and is not consulted anywhere on this path; a transient branch in `classifyFailure` plus a distinct exit code or cause class would make the unit guarantee hold end to end.

lockwobr (commented):
Cross-review of `85702abb`. The publish path — the actual scope of #2054 — is genuinely covered: `runEvidencePublishCmd` does no filesystem access on the operator-supplied dir before `attestation.Publish`, so the bounded readers really are the first touch. All six `HasBundleMarkers` call sites were updated, every one checks `err` before `ok`, and the `ctx.Err()` pre-check before the goroutine launch is a real correctness detail (without it a fast local stat could win the select race against an already-canceled context).

The inline comments cover three things that read as fixed but are not, plus two doc/diagnostic mismatches and a test-coverage gap.

### Open questions

1. **Cumulative bound.** On a fully hung mount `loadOnDiskBundle` serially burns up to 6 × `FileReadTimeout` (four marker stats across two candidate dirs, plus the statement and recipe reads) ≈ 180s, with no overall deadline on the command. Is per-syscall the right granularity, or should the bound be per-call-tree?
2. **The PR description mentions bounding a "directory walk"**, but the diff contains no walk change — and there does not appear to be an *unbounded* walk left on the publish/sign path either (`pkg/oci/source_files.go:955` already checks `ctx.Err()` between entries; `attestation/manifest.go:52` and `verifier/inventory.go:336` are likewise already ctx-aware and untouched). Is the description describing prior work, or was an intended change dropped?
3. **Remaining unbounded reads in the same code paths** — `verifier/input.go:41`, `verifier/bounded_read.go:31`, `verifier/inventory.go:178`, `verifier/pointer.go:38`, `verifier/signature.go:72`, `attestation/pointer.go:365/369/380`, `attestation/bom.go:51`. A tracking issue would make the partial coverage explicit rather than implied.
4. **Was fail-open-on-stat-error deliberate?** The inline comment at `publish.go:312-313` suggests yes — in which case the godoc above the function is what needs correcting, not the code.
5. **Three-way duplication.** `materializeDir`, `resolveBundleDir` and `resolveSummaryDir` are now near-identical two-candidate probes. Under `dupl`'s threshold, but a shared helper would keep them from drifting.

Nothing here blocks on CI: all non-skipped checks are green on this commit.

pkg/evidence/attestation/publish.go
yuanchen8911, line 320: The doc comment above says the (bool, error) shape keeps "a ctx timeout / unexpected failure" distinct from a genuine "not a bundle" answer, but this closure collapses every stat error to present=false and unconditionally returns nil — only the timeout wrapper can produce a non-nil error. A permission-denied or EIO stat on the marker path therefore flows through resolveSummaryDir / materializeDir as ErrCodeInvalidRequest "does not look like a summary bundle", pointing the operator at their input instead of the backing storage.

This preserves pre-PR behavior (the old code also returned false on any stat error), and propagating stat errors would be a behavior change beyond #2054's timeout scope — so rather than expanding behavior here, narrow the contract to match the code: the doc comment should say only timeouts surface as errors and all stat failures (including permission errors) read as "not a marker". The PR description's "reports timeout/IO errors instead of silently returning false" claim should be corrected the same way. Surfacing non-ENOENT stat errors is a reasonable follow-up if there's appetite for it.

pkg/evidence/attestation/publish_test.go
```diff
@@ -65,6 +65,106 @@ func wantInvalidRequest(t *testing.T, err error) {
...
+			cancel()
```
yuanchen8911, line 116: Every new timeout test (here, sign_test.go, fetch_test.go) cancels the context before invoking the reader, so all of them return at withFileReadTimeout's early ctx.Err() check — coverage shows the <-ctx.Done() select arm at publish.go:212-218 has zero executions. That path is the actual fix for #2054: a regression in it (removing the buffered channel, breaking the select) would pass this suite while reintroducing the indefinite hang.

Strong non-blocking recommendation: add a deterministic cancel-after-start test. withFileReadTimeout takes fn as a closure, so a direct unit test can use a started/release/finished handshake — the fn stub signals started then blocks on release; the test waits for started, cancels the context explicitly, asserts ErrCodeTimeout, then closes release and joins the stub via finished. Deterministic, fast, and leaves no goroutine behind.

yuanchen8911 (dismissed):
Reviewed at 85702ab. The approach is sound: bounding the actual read (not a stat preflight) is the right call, the ctx.Err() pre-check makes the canceled-context path deterministic, error codes are preserved through the verifier (PropagateOrWrap keeps ErrCodeTimeout), and all call sites of the signature change are updated with no stragglers repo-wide. CI is green at this commit.

Approving with two non-blocking comments; the first is a strong recommendation:

1. Missing coverage of the in-flight timeout path (detailed inline on publish_test.go). Every new timeout test pre-cancels the context, so the goroutine + select mechanism — the actual fix for #2054 — has zero test executions. The implementation is correct, but a regression there would pass the suite while reintroducing the hang. A deterministic cancel-after-start test is cheap since withFileReadTimeout takes fn as a closure.

2. HasBundleMarkers' contract overstates what the error return conveys (detailed inline on publish.go). The doc comment and the PR description both say marker detection reports "timeout/IO errors", but the closure collapses every non-timeout stat error (EACCES, EIO, stale NFS handle) to (false, nil). Propagating stat errors would be a behavior change outside #2054's timeout scope, so the right fix is to narrow the doc comment — and the PR body's "timeout/IO errors" claim — to timeouts only, and leave stat-error propagation to a follow-up if wanted.

No change requested on two documented tradeoffs: context.Canceled reported as ErrCodeTimeout with "timed out reading ..." wording (no consumer distinguishes cancel from deadline today), and one timeout goroutine per stat (fine for a CLI leaf command, revisit only if these readers move into a long-lived server path).

Housekeeping: the branch is behind main — please rebase onto origin/main before the next push.

njhensley:
Thanks for the thorough cross-review, @yuanchen8911 @lockwobr (and CodeRabbit). Addressed in the force-pushed `1b5bd8d`:

**Fixed in this PR:**
- **In-flight timeout coverage** (all three reviewers, the strong rec): added `TestWithFileReadTimeout_InFlightTimeout` — an fn that signals `started`, blocks past an in-flight `cancel()`, then is released and joined via a `finished` handshake. This exercises the `goroutine + select` arm (`publish.go`'s `<-ctx.Done()` case) directly; the pre-existing canceled-ctx tests only hit the `ctx.Err()` pre-check, so that arm previously had zero executions. No leaked worker.
- **`HasBundleMarkers` godoc overclaim** (@yuanchen8911, @lockwobr): narrowed the contract to match the code — only the timeout wrapper yields `(false, err)`; every stat error (incl. EACCES/EIO/ESTALE) reads as "not a marker" `(false, nil)`, fail-soft, unchanged. The PR description's "timeout/IO errors" wording is corrected the same way.
- **Timeout message drops the directory** (@lockwobr): the marker read now labels with the full `path`, not the bare filename, so a timeout names which candidate dir stalled.
- **`fetch_test` doc overclaim**: scoped the "must fail closed" comment to the marker probe only, with an explicit note that `verify` is not yet end-to-end hang-immune.
- Commit message corrected: dropped the inaccurate "directory walk" phrasing (@lockwobr — no walk change in the diff).

**Deferred to #2083** (filed to make the partial coverage explicit): the residual unbounded/mis-diagnosed reads on these paths — `DetectInputForm`'s pre-probe stat, post-materialize verifier reads, the OCI push-staging walk (so `publish` can still hang one step later on a fully dead mount), `classifyFailure`'s missing transient branch, and surfacing non-`ENOENT` stat faults distinctly. Keeping this PR scoped to the summary-bundle reads that #2054 calls out.

**Deliberately not changed:** the per-stat goroutine (both of you agreed it's fine for a CLI leaf; a concurrency limiter would be over-engineering here), and `context.Canceled` → `ErrCodeTimeout` (documented tradeoff — no consumer distinguishes cancel from deadline today; folded into #2083 if we want it).

Rebased onto latest `origin/main`; CI re-running.

yuanchen8911 (approved):
Re-approved at 1b5bd8d. Both review asks landed and were verified against the delta: the in-flight timeout test exercises the goroutine+select arm with a started/release/finished handshake and joins the worker, and the HasBundleMarkers godoc plus PR description are narrowed to match the fail-soft stat behavior, with non-ENOENT stat surfacing and the residual unbounded reads tracked in #2083. Rebase onto main confirmed.

PR #2073, opened 2026-08-05
feat(supply-chain): publish aiperf-bench third-party source as OCI referrer

lockwobr:
## Summary

Publishes the source of every third-party Python package installed into the `aiperf-bench` image, attached to that image as an OCI 1.1 referrer, and documents retrieval for all seven released images.

## Motivation / Context

NVIDIA's container policy requires the source of every third-party open-source component added on top of an approved base image to be retrievable by whoever pulls the image, **regardless of license**. That is a separate obligation from attribution and is not satisfied by `THIRD_PARTY_NOTICES.md`.

Six of the seven released images already comply. Their dependencies are vendored, so the GitHub release source archive for the matching tag contains `vendor/` — the complete source of every module compiled in, since builds run with `-mod=vendor`. Only `aiperf-bench` installs from wheels (built distributions), leaving nothing under NVIDIA's control that holds its source.

OSRB confirmed the referrers API is an acceptable mechanism provided retrieval is documented, since the artifact is invisible to `docker pull`.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Validator (`pkg/validator`) — the `aiperf-bench` image
- [x] Docs/examples (`docs/`)
- [x] Other: release workflow, tooling

## Implementation Notes

**It does not gate the release.** `attach-source` runs *after* `publish` and is deliberately excluded from `release-check`. Attaching depends on PyPI serving 129 sdists — an uncontrolled third party on the critical path. Gating on it would turn a PyPI blip or a yanked version into a failed release and a release-day fire drill. Nothing is lost by attaching afterwards: a referrer binds to the image digest, and promotion does not change the digest, so the artifact lands on the same subject either way. A failure is a re-runnable job rather than a recut tag.

**The package list is read from the committed notices fragment**, not resolved independently, so the source archive and the attribution record cannot describe different closures. It also takes sdist URLs from the PyPI JSON API, turning a 30-minute `pip download` into ~47 seconds.

**The archive is byte-reproducible.** gzip stamps the current time into its header, so a naive `tarfile.open(mode="w:gz")` produced a different referrer digest on every run.

**Retry is transient-only.** 4 attempts with exponential backoff on both the metadata lookups and the downloads. `404` is excluded on purpose: it is the definitive answer that no sdist exists, and it feeds the fail-closed path that aborts on any package whose source cannot be published. A failed *download* aborts rather than being skipped, since skipping would produce an archive silently missing a package.

**`oras` receives a bare filename with `cwd` set.** It rejects absolute paths because the file argument becomes the layer's `org.opencontainers.image.title`; passing an absolute path would publish the builder's local directory layout. `--disable-path-validation` would have suppressed the check and kept the leak.

**No maintainer contacts ship in the artifact.** aiperf's PyPI metadata carries a maintainer distribution list; it is deliberately not repeated in the README that ships inside a public archive.

## Testing

```bash
make lint                          # pass
go test ./tests/notices/...        # 14 tests pass
golangci-lint run ./tests/notices/...  # 0 issues
```

**Verified end to end against a real registry (ttl.sh)**, not just locally: attach, `oras discover`, `oras pull`, and a byte-identical SHA256 on the retrieved 253 MB archive with all 130 entries intact. That test found the absolute-path bug, which a localhost prototype had missed.

New tests pin the properties that would otherwise rot silently: that the closure is read from the committed notices fragment, that missing source fails closed, that no `@nvidia.com` address is embedded, that `oras` gets a relative path, and the retry policy including 404 handling. The oras-path assertion is mutation-verified — reverting the fix fails it with the intended message.

**`make qualify` is not fully green on this machine**, from three pre-existing environmental failures unrelated to this change, each reproducing on a clean `main` here: `test-shell` (sandboxed `mktemp`), `tests/releasepolicy` (`timeout` is GNU coreutils, absent on macOS), `pkg/oci` (local helm 4.2.0 vs pinned 4.2.3). Every other package passes.

## Risk Assessment

**Rollout notes:** No image contents change and no runtime behaviour changes. The new release job runs after publish and cannot block a release. `oras` is a new pinned tool dependency (`v1.3.0`), needed because cosign cannot attach arbitrary artifacts. If the `aiperf-bench` image is replaced, this deletes cleanly: one job, one script, one docs section.

## Checklist

### DIFF
--- a/.github/workflows/on-tag.yaml
+++ b/.github/workflows/on-tag.yaml
@@ -811,6 +811,173 @@ jobs:
   # =============================================================================
   # Notify: Post release announcement to Slack
   # =============================================================================
+  # Source: attach the aiperf-bench third-party source as an OCI 1.1 referrer
+  # =============================================================================
+  #
+  # NVIDIA's container policy requires the source of every third-party
+  # open-source component added on top of an approved base image to be
+  # retrievable by whoever pulls that image, regardless of license. This is a
+  # separate obligation from attribution (THIRD_PARTY_NOTICES.md) and is not
+  # satisfied by it.
+  #
+  # Only aiperf-bench needs this. The six Go images vendor their dependencies,
+  # so the GitHub release source archive for this tag already carries the
+  # complete source of every module compiled into them.
+  #
+  # A referrer rather than a layer in the image: the sdists are ~250 MB, which
+  # would otherwise be paid on every benchmark pod pull forever for data nothing
+  # reads at runtime. OSRB confirmed the referrers API is acceptable provided
+  # retrieval is documented (docs/integrator/supply-chain-verification.md).
+  #
+  # Runs against the resolved candidate digest, not a tag: a referrer binds to
+  # one digest, so attaching to a moving tag would leave later builds
+  # unreferenced.
+  #
+  # Deliberately NOT part of release-check, and deliberately after publish.
+  # Attaching depends on PyPI serving 129 sdists, which is an uncontrolled
+  # third party on the critical path of a release. Gating on it converts a PyPI
+  # blip, or a yanked package version, into a failed release and a release-day
+  # fire drill. Nothing is lost by attaching afterwards: a referrer binds to the
+  # image digest, and promotion does not change the digest, so the artifact
+  # lands on exactly the same subject either way. A failure here is re-runnable
+  # as a single job rather than requiring the tag to be recut.
+  #
+  # Gated on resolve-candidates and publish succeeding: there is no image to
+  # attach to otherwise. The job itself fails loudly if the attach, the
+  # signature, or the discoverability check does not succeed.
+  attach-source:
+    name: Attach Third-Party Source
+    runs-on: ubuntu-latest
+    needs: [detect, resolve-candidates, publish]
+    if: needs.resolve-candidates.result == 'success' && needs.publish.result == 'success'
+    timeout-minutes: 20
+    permissions:
+      contents: read
+      # packages: write to push the referrer to ghcr.io alongside the image.
+      packages: write
+      # id-token: write for keyless cosign signing of the referrer. oras attach
+      # produces an unsigned artifact, and anyone with push access could attach a
+      # second one carrying the same artifactType; signing lets a consumer tell
+      # ours apart rather than trusting whatever the type filter returns.
+      id-token: write
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Load versions
+        id: versions
+        uses: ./.github/actions/load-versions
+
+      - name: Authenticate to registry
+        uses: ./.github/actions/ghcr-login
+
+      - name: Setup ORAS
+        uses: oras-project/setup-oras@1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d  # v2.0.1
+        with:
+          version: ${{ steps.versions.outputs.oras }}
+
+      - name: Install Cosign
+        uses: sigstore/cosign-installer@6f9f17788090df1f26f669e9d70d6ae9567deba6  # v4.1.2
+        with:
+          cosign-release: ${{ steps.versions.outputs.cosign }}
+
+      - name: Build and attach source archive
+        id: attach
+        env:
+          DIGEST_MAP: ${{ needs.resolve-candidates.outputs.digest_map }}
+        run: |
+          set -euo pipefail
+          digest="$(jq -er '."aiperf-bench"' <<<"${DIGEST_MAP}")"
+          image="ghcr.io/nvidia/aicr-validators/aiperf-bench@${digest}"
+          # Skip if a source referrer is already attached to this digest.
+          # `oras attach` stamps org.opencontainers.image.created into the
+          # manifest, so re-running would push a NEW manifest digest and leave
+          # two referrers of the same artifactType on one image. A consumer
+          # selecting by type could then not tell which is ours -- reintroducing,
+          # on the deliberately re-runnable path, exactly the ambiguity the
+          # signature exists to remove.
+          # Reuse an existing referrer rather than attaching a second one:
+          # `oras attach` stamps org.opencontainers.image.created, so a re-run
+          # would push a new manifest digest and leave two referrers of the same
+          # artifactType on one image, which a consumer could not disambiguate.
+          #
+          # Crucially this does NOT skip signing. If a previous run attached and
+          # then failed before signing, the digest carries an unsigned referrer;
+          # skipping here would make every re-run skip the signature too and
+          # fail verification forever, contradicting this job's re-runnable
+          # recovery. Reusing the digest lets a re-run finish the interrupted
+          # attach-then-sign sequence instead of wedging on it.
+          existing="$(oras discover --format json "${image}" \
+            | jq -r '[.referrers[]? | select(.artifactType
+                     == "application/vnd.nvidia.aicr.source.v1+tar")
+                     | .digest] | first // ""')"
+          if [[ -n "${existing}" ]]; then
+            echo "Source referrer already attached (${existing}); reusing it and re-signing."
+            echo "referrer=${existing}" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          echo "Attaching third-party source to ${image}"
+          # The tool prints the referrer manifest digest on stdout; everything
+          # else goes to stderr.
+          referrer="$(python3 tools/generate-aiperf-source --attach "${image}")"
+          if [[ ! "${referrer}" =~ ^sha256:[0-9a-f]{64}$ ]]; then
+            echo "::error::did not receive a referrer digest from the attach step"
+            exit 1
+          fi
+          echo "referrer=${referrer}" >> "$GITHUB_OUTPUT"
+
+      - name: Sign the source referrer
+        env:
+          REFERRER: ${{ steps.attach.outputs.referrer }}
+        # Signed for the same reason every other referrer AICR publishes is:
+        # the artifact is otherwise unauthenticated, and a consumer selecting by
+        # artifactType alone cannot distinguish ours from one an attacker with
+        # push access attached beside it (CWE-345). Signing the referrer's own
+        # digest is what makes the documented retrieval verifiable.
+        run: |
+          set -euo pipefail
+          cosign sign --yes \
+            "ghcr.io/nvidia/aicr-validators/aiperf-bench@${REFERRER}"
+
+      - name: Verify the referrer is retrievable
+        env:
+          DIGEST_MAP: ${{ needs.resolve-candidates.outputs.digest_map }}
+        # Attaching is not the same as being findable. ghcr.io has no referrers
+        # API and serves these through the spec's tag fallback, so confirm a
+        # client can actually discover the artifact before the release proceeds.
+        run: |
+          set -euo pipefail
+          digest="$(jq -er '."aiperf-bench"' <<<"${DIGEST_MAP}")"
+          image="ghcr.io/nvidia/aicr-validators/aiperf-bench@${digest}"
+          found="$(oras discover --format json "${image}" \
+            | jq -r '[.referrers[]? | select(.artifactType
+                     == "application/vnd.nvidia.aicr.source.v1+tar")] | length')"
+          # Exactly one, not at least one. Two referrers of the same type leave
+          # a consumer unable to tell which is authoritative, and `>= 1` would
+          # hide that rather than catch it.
+          if [[ "${found}" -ne 1 ]]; then
+            echo "::error::expected exactly 1 source referrer on ${image}, found ${found}"
+            exit 1
+          fi
+
+          # Discoverable is not the same as verifiable. Confirm the signature we
+          # just created actually resolves, so a silently-failed sign cannot ship
+          # an artifact the documented retrieval will then reject.
+          referrer="$(oras discover --format json "${image}" \
+            | jq -r '.referrers[] | select(.artifactType
+                     == "application/vnd.nvidia.aicr.source.v1+tar") | .digest')"
+          cosign verify \
+            --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+            --certificate-identity \
+              "https://github.com/${GITHUB_REPOSITORY}/.github/workflows/on-tag.yaml@${GITHUB_REF}" \
+            "ghcr.io/nvidia/aicr-validators/aiperf-bench@${referrer}" >/dev/null
+
+          echo "Source referrer discoverable and signature verified on ${image}"
+
+  # =============================================================================
 
   notify:
     name: Slack Notification
@@ -846,7 +1013,23 @@ jobs:
   summary:
     name: Release Summary
     runs-on: ubuntu-latest
-    needs: [detect, tests, build-ko, build-docker, build-gate, docker-manifest, resolve-candidates, image-vuln-scan, attest, release-check, promote-images, publish, publish-homebrew, deploy, notify]
+    needs:
+      - detect
+      - tests
+      - build-ko
+      - build-docker
+      - build-gate
+      - docker-manifest
+      - resolve-candidates
+      - image-vuln-scan
+      - attest
+      - attach-source
+      - release-check
+      - promote-images
+      - publish
+      - publish-homebrew
+      - deploy
+      - notify
     if: always()
     timeout-minutes: 5
     permissions:
@@ -868,6 +1051,11 @@ jobs:
           HOMEBREW: ${{ needs.publish-homebrew.result }}
           DEPLOY: ${{ needs.deploy.result }}
           NOTIFY: ${{ needs.notify.result }}
+          # Non-gating by design (see the attach-source job), so a failure here
+          # does not red the release. It still reds its own job, but the summary
+          # is where the release outcome is read as a whole, and a missing row
+          # would leave "did the source publish?" unanswered there.
+          ATTACH_SOURCE: ${{ needs.attach-source.result }}
           TAG: ${{ github.ref_name }}
           COMMIT: ${{ github.sha }}
           IS_PRERELEASE: ${{ needs.detect.outputs.is_prerelease }}
@@ -892,6 +1080,7 @@ jobs:
             echo "| Release Check | ${RELEASE_CHECK} |"
             echo "| Image Promotion | ${PROMOTE} |"
             echo "| Publish | ${PUBLISH} |"
+            echo "| Third-Party Source | ${ATTACH_SOURCE} |"
             echo "| Homebrew | ${HOMEBREW} |"
             echo "| Deploy | ${DEPLOY} |"
             echo "| Slack Notification | ${NOTIFY} |"
--- a/.settings.yaml
+++ b/.settings.yaml
@@ -46,6 +46,15 @@ security_tools:
   cosign: 'v3.1.2'
   # renovate: datasource=github-releases depName=anchore/syft depType=security_tools
   syft: 'v1.50.0'
+  # renovate: datasource=github-releases depName=oras-project/oras depType=security_tools
+  # Attaches the aiperf-bench third-party source archive as an OCI 1.1 referrer
+  # (tools/generate-aiperf-source). Cosign cannot attach arbitrary artifacts, so
+  # this is a separate dependency from the signing/attestation path.
+  #
+  # Stored WITHOUT the leading 'v', unlike cosign/syft/grype above:
+  # oras-project/setup-oras looks the version up against bare release keys and
+  # throws on 'v1.3.0'. tools/setup-tools re-adds the prefix for the download URL.
+  oras: '1.3.0'
 
 # E2E Testing Tools
 testing_tools:
--- a/docs/integrator/supply-chain-verification.md
+++ b/docs/integrator/supply-chain-verification.md
@@ -13,8 +13,9 @@ top-level [`SECURITY.md`](../../SECURITY.md).
 
 Verification uses [Cosign](https://docs.sigstore.dev/cosign/system_config/installation/),
 the [GitHub CLI](https://cli.github.com/) (`gh`), `crane` (recommended; `docker inspect` resolves a digest only after a local pull),
-`jq`, and — for in-cluster enforcement — `kubectl`. Binary and bundle
-verification (`aicr verify`) need only the `aicr` binary.
+`jq`, [ORAS](https://oras.land/docs/installation) (only for retrieving the
+third-party source archive), and — for in-cluster enforcement — `kubectl`.
+Binary and bundle verification (`aicr verify`) need only the `aicr` binary.
 
 **Cosign version.** AICR release signatures (the binary attestation and the
 signed recipe catalog) are recorded in **Rekor v2** as of the v2 cutover (see the
@@ -57,15 +58,19 @@ docker login ghcr.io
 
 ## Unified Metadata Retrieval
 
-Every piece of release metadata AICR publishes for a container image is a
-signed in-toto attestation attached to that image as an OCI referrer. There is
-one retrieval command per metadata kind, and each kind has a fixed subject:
+Release metadata AICR publishes for a container image is attached to that image
+as an OCI referrer. The first three kinds below are signed in-toto attestations.
+The source archive is a plain OCI artifact rather than an attestation, but it is
+Sigstore-signed like the rest, so it is verified with `cosign verify` instead of
+`cosign verify-attestation`. There is one retrieval command per kind, and each
+has a fixed subject:
 
 | Metadata | Predicate type | Subject | Retrieved with |
 |----------|----------------|---------|----------------|
 | SLSA build provenance | `https://slsa.dev/provenance/v1` | multi-platform index digest | `gh attestation verify` |
 | SPDX SBOM | `https://spdx.dev/Document` | per-platform manifest digest | `cosign verify-attestation --type spdxjson` |
 | OpenVEX | `https://openvex.dev/ns` | multi-platform index digest | `cosign verify-attestation --type openvex` |
+| Third-party source (`aiperf-bench` only) | `application/vnd.nvidia.aicr.source.v1+tar` | multi-platform index digest | `oras pull` — see [Third-Party Source Code](#third-party-source-code) |
 
 The subjects differ because the claims differ. Provenance describes the build
 that produced the whole release image, and the VEX document is a single,
@@ -153,6 +158,87 @@ Cosign, so the floor does not apply to it. GHCR does not implement the OCI 1.1
 specification's referrers *tag* schema (a `sha256-{hex}` tag holding an index of
 referrers). Cosign and ORAS do this transparently, which is why the commands
 above are the documented path rather than a raw referrers API call.
+
+## Third-Party Source Code
+
+Source for every third-party open-source component AICR adds on top of its base
+images is published, regardless of license. Where it lives depends on how the
+component reaches the image.
+
+| Image | Third-party components | Where the source is |
+|-------|------------------------|---------------------|
+| `aicr`, `aicrd`, `aicr-gate`, `aicr-validators/{conformance,deployment,performance}` | Go modules compiled into the binary | The GitHub release source archive for the matching tag. Builds run with `-mod=vendor`, so `vendor/` in that archive is exactly the source that ships. |
+| `aicr-validators/aiperf-bench` | Python packages installed from wheels | An OCI referrer on the image, retrieved with the commands below. |
+| all | base image contents | Provided by NVIDIA with `nvcr.io/nvidia/distroless/*` under that image's own approval. |
+
+License texts for all of the above are in `THIRD_PARTY_NOTICES.md`, published as
+an asset on each release. That file discharges attribution; this section covers
+source availability, which is a separate obligation.
+
+### Retrieving the aiperf-bench source
+
+The archive is attached as an OCI referrer, so it is **not** fetched by
+`docker pull` and will not appear in the image's layers. Retrieve it explicitly:
+
+```bash
+IMAGE="ghcr.io/nvidia/aicr-validators/aiperf-bench:v0.19.0"
+
+# 1. Resolve the image to its digest and list what is attached to it.
+DIGEST=$(crane digest "${IMAGE}")
+oras discover --format tree "${IMAGE%:*}@${DIGEST}"
+
+# 2. Select the source referrer. Fail loudly on anything but exactly one:
+#    artifactType is not a unique key, and anyone with push access to the
+#    repository can attach another referrer carrying the same type. Taking the
+#    first match would silently pick theirs.
+mapfile -t MATCHES < <(oras discover --format json "${IMAGE%:*}@${DIGEST}" \
+  | jq -r '.referrers[]
+           | select(.artifactType == "application/vnd.nvidia.aicr.source.v1+tar")
+           | .digest')
+if [ "${#MATCHES[@]}" -ne 1 ]; then
+  echo "expected exactly 1 source referrer, found ${#MATCHES[@]}" >&2
+  exit 1
+fi
+SOURCE="${MATCHES[0]}"
+
+# 3. Verify the signature before trusting the bytes. Pin the exact signing
+#    workflow and tag; a regexp over the repository would accept a signature
+#    from any workflow in it.
+cosign verify \
+  --certificate-oidc-issuer "${AICR_ISSUER}" \
+  --certificate-identity \
+    "https://github.com/NVIDIA/aicr/.github/workflows/on-tag.yaml@refs/tags/${TAG}" \
+  "${IMAGE%:*}@${SOURCE}"
+
+# 4. Pull the verified archive.
+oras pull -o ./aiperf-source "${IMAGE%:*}@${SOURCE}"
+
+# 5. The archive holds one sdist per installed package, plus a README.
+tar tzf ./aiperf-source/aiperf-bench-python-source.tar.gz | head
+```
+
+The archive contains the source distribution of every Python package installed
+into the image **that publishes one upstream**, at the exact version installed.
+One package is intentionally absent: `aiperf` itself publishes no source
+distribution to PyPI. It is an
+NVIDIA package and its source is at [ai-dynamo/aiperf](https://github.com/ai-dynamo/aiperf); the
+archive's `README.txt` records this too.
+
+Only `aiperf-bench` carries this referrer. The other six images need none: their
+dependencies are vendored, so the release source archive already contains them.
+
+**Scope of the correspondence.** The archive is resolved from the same
+`requirements.txt` the image installs, so both derive from one input. Two gaps
+remain and are tracked rather than claimed away:
+
+* `pip` and `wheel` are provided by `python -m venv` rather than declared in
+  `requirements.txt`, so they ship in the image without source in this archive.
+* The image build layer-caches its dependency resolution, so a cached build can
+  carry an older closure than a freshly resolved archive (tracked in
+  [#2086](https://github.com/NVIDIA/aicr/issues/2086)).
+
+A referrer binds to one image digest, so resolve the tag to a digest first as
+shown above rather than assuming a tag keeps the same attachment across builds.
 
 ### Using the published VEX document
 
--- a/tools/generate-aiperf-source
+++ b/tools/generate-aiperf-source
@@ -1,0 +1,457 @@
+#!/usr/bin/env python3
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Publish the third-party Python source for the aiperf-bench image.
+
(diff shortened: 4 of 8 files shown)


.github/workflows/uat-gcp.yaml
```diff
@@ -454,11 +454,43 @@ jobs:
...
+              -e AUTO_APPROVE=true \
+              -e KEY_CONTENT="$(base64 < ${{ steps.auth.outputs.credentials_file_path }})" \
+              ${{ env.GKE_ACTUATOR_IMAGE }} apply; then
+              brought_up=true
+              break
+            fi
+            if [ "${attempt}" -lt 2 ]; then
+              echo "Bringup attempt ${attempt} failed; retrying in 60s (a transient node-pool state often self-heals, and apply is idempotent)..."
+              sleep 60
+            fi
```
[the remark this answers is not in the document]
lockwobr: Going to revert this change, does not belong in this PR.

njhensley (commented):
## Multi-persona review — `feat(supply-chain): publish aiperf-bench third-party source as OCI referrer`

**Method:** 4 independent persona reviewers (Supply-chain/Security · CI-DX/Operability · Correctness · Domain & Docs) fanned out in parallel, then every finding was adjudicated by an adversarial senior meta-reviewer that re-derived each claim from the resolved code. The two Major findings were independently reproduced before surviving. Reviewed at head `f25c3e4f`.

### Overall assessment

This is a carefully-reasoned, high-quality PR. The **producer side is solid**: PyPI sha256 verification with fail-closed on mismatch, fail-closed on any non-allowlisted missing sdist, byte-reproducible tar, the absolute-path-leak fix done the right way (bare filename + `cwd`, not `--disable-path-validation`), maintainer-DL leak avoidance, a least-privilege SHA-pinned workflow, and non-gating placement off the release critical path. The tests execute the real Python code rather than asserting on its source text. No blockers.

Both **Major** findings concern the **consumer-facing verification story**, and both are doc-only:

- **F1** — the documented retrieval command (`SOURCE=$(... jq -r ... .digest)`) breaks when more than one same-`artifactType` referrer exists, which is reachable by (a) the push-access attacker the design explicitly defends against, or (b) a benign re-run of the deliberately re-runnable `attach-source` job. It verifies a pre-selected set rather than selecting *by* verification, so it can't fall through to the legitimate signed referrer — it just errors.
- **F2** — the consumer `cosign verify` identity regexp (`^https://github.com/NVIDIA/aicr/`) is looser than the tight `on-tag.yaml@refs/tags/.+$` pin this same file uses everywhere else, and looser than the standard it argues for at lines 98–101.

Neither breaks a clean first release, but both undercut the security guarantee the design markets, and both are cheap to fix. The Minor items (F3 signature not verified at release time · F4 non-idempotent re-run duplicate · F5 metadata-error misclassification) and the nitpicks are polish.

**Recommendation: Approve with comments** — address F1 & F2 before/with merge.

```
🔴 Blocker 0  |  🟠 Major 2  |  🟡 Minor 3  |  🔵 Nitpick 6
```

PyPI sha256 fail-closed integrity · missing-sdist fail-closed abort · retry policy (404 correctly not retried) · absolute-path leak avoided via bare filename + `cwd` · maintainer-DL leak avoided · SystemExit propagates through `ThreadPoolExecutor` (checksum mismatch truly aborts) · reproducible tar · `read_closure` regex verified against the real 130-row notices file · no `KeyError` in `KNOWN_NO_SDIST` lookups · last-`Digest:`-line parsing · index-vs-platform digest consistency doc↔code · artifactType string exact-match doc↔code · anchor `#third-party-source-code` resolves · least-privilege permissions + SHA-pinned actions · non-gating-after-publish placement.

<sub>Generated by a multi-persona + adversarial-meta-review pass. `lockwobr`'s "COMMENTED" entry is the author self-commenting; CodeRabbit's earlier suggestions are already largely incorporated into this head and were not re-raised.</sub>

docs/integrator/supply-chain-verification.md
```diff
@@ -154,6 +158,68 @@ specification's referrers *tag* schema (a `sha256-{hex}` tag holding an index of
...
+
+```bash
+IMAGE="ghcr.io/nvidia/aicr-validators/aiperf-bench:v0.19.0"
+
+# 1. Resolve the image to its digest and list what is attached to it.
+DIGEST=$(crane digest "${IMAGE}")
+oras discover --format tree "${IMAGE%:*}@${DIGEST}"
+
+# 2. Select the source referrer.
+SOURCE=$(oras discover --format json "${IMAGE%:*}@${DIGEST}" \
```
njhensley: **🟠 Major — Documented consumer retrieval breaks with >1 same-artifactType referrer — the CWE-345 case it defends against**

Step 2 does `SOURCE=$(oras discover ... | jq -r '.referrers[] | select(.artifactType=="application/vnd.nvidia.aicr.source.v1+tar") | .digest')`. `jq -r` prints one line per match, so if two referrers of that artifactType exist, `SOURCE` becomes `sha256:aaa\nsha256:bbb`. Step 3's `cosign verify "${IMAGE%:*}@${SOURCE}"` and step 4's `oras pull ... "@${SOURCE}"` then get a malformed newline-bearing reference and error out.

Two ways to reach ≥2 referrers:
- **(a) attacker** — a push-access actor attaches a second same-type referrer. This is the exact CWE-345 threat the `attach()` docstring and the workflow comments cite. The command's promise ("the signature is what identifies ours") is *not* delivered: it pre-selects `SOURCE` by `artifactType` and verifies that fixed set — it never selects **by** verification, so it can't fall through to the legitimate signed referrer; it just dies.
- **(b) benign re-run** — see F4 (oras stamps a per-run `created` timestamp, so a re-run of the re-runnable `attach-source` job yields a second referrer).

It fails closed (an attacker's artifact is never trusted), which is why this is Major and not a blocker — but the documented path denies retrieval to legitimate consumers precisely when the defense should engage.

**Fix:** select by verification — loop the candidate digests and keep the one whose signature verifies, then pull that one:
```bash
for d in $(oras discover --format json "${IMAGE%:*}@${DIGEST}" | jq -r '.referrers[] | select(.artifactType=="application/vnd.nvidia.aicr.source.v1+tar") | .digest'); do
  cosign verify --certificate-oidc-issuer ... --certificate-identity-regexp ... "${IMAGE%:*}@${d}" && SOURCE="$d" && break
done
oras pull -o ./aiperf-source "${IMAGE%:*}@${SOURCE}"
```
Also worth making `attach-source` idempotent (F4) so benign re-runs don't create the duplicate in the first place.

docs/integrator/supply-chain-verification.md
```diff
@@ -154,6 +158,68 @@ specification's referrers *tag* schema (a `sha256-{hex}` tag holding an index of
...
+  | jq -r '.referrers[]
+           | select(.artifactType == "application/vnd.nvidia.aicr.source.v1+tar")
+           | .digest')
+
+# 3. Verify it before trusting it. Selecting by artifactType alone is not
+#    sufficient: anyone with push access to the repository can attach another
+#    referrer carrying the same type. The signature is what identifies ours.
+cosign verify \
+  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
+  --certificate-identity-regexp '^https://github.com/NVIDIA/aicr/' \
```
njhensley: **🟠 Major — Consumer cosign verify identity regexp is far looser than the doc's own standard**

The referrer is signed **inline** by the `attach-source` job in `on-tag.yaml` (`cosign sign`), so the real keyless SAN is `https://github.com/NVIDIA/aicr/.github/workflows/on-tag.yaml@refs/tags/<TAG>`. But the documented verify uses:
```
--certificate-identity-regexp '^https://github.com/NVIDIA/aicr/'
```
That is prefix-only with an unanchored tail — it accepts a signature from **any** workflow on **any** ref (`refs/heads/*`, `refs/pull/*`, other reusable workflows holding `id-token: write`). This directly contradicts this document's own guidance at lines 98–101 (even `refs/tags/.+` is called too loose), and every other verify block in the file pins the tight form, e.g. line 555:
```
--certificate-identity-regexp '^https://github\.com/NVIDIA/aicr/\.github/workflows/on-tag\.yaml@refs/tags/.+$'
```
Against the PR's stated CWE-345 threat (a repo-write actor), the loose regexp accepts a signature that actor could mint from a branch-triggered workflow, whereas the tight pin would reject it. The dots are also unescaped (`github.com` matches `githubXcom`), cosmetic next to the pinning gap.

**Fix:** use the same tight, dot-escaped, `@refs/tags/`-anchored pattern the rest of the file uses:
```
--certificate-identity-regexp '^https://github\.com/NVIDIA/aicr/\.github/workflows/on-tag\.yaml@refs/tags/.+$'
```

.github/workflows/on-tag.yaml
```diff
@@ -811,6 +811,129 @@ jobs:
...
+        # the artifact is otherwise unauthenticated, and a consumer selecting by
+        # artifactType alone cannot distinguish ours from one an attacker with
+        # push access attached beside it (CWE-345). Signing the referrer's own
+        # digest is what makes the documented retrieval verifiable.
+        run: |
+          set -euo pipefail
+          cosign sign --yes \
+            "ghcr.io/nvidia/aicr-validators/aiperf-bench@${REFERRER}"
+
+      - name: Verify the referrer is retrievable
```
njhensley, line 945: **🟡 Minor — Release-time verify step checks discoverability only, not the signature it just created**

The "Verify the referrer is retrievable" step counts referrers by `artifactType` (`length >= 1`) and stops there. A `cosign sign` that succeeded but produced a signature a consumer can't verify (an identity/format regression — the class of the legacy-dsse/Rekor-v2 bugs this repo has hit before) would ship and only be caught by a downstream integrator following the docs. The `set -euo pipefail` sign step catches a *missing* signature, but not an *unverifiable* one.

**Fix:** after signing, run the documented consumer path against the just-signed digest and fail the job on mismatch — this also keeps the workflow and the doc command honest with each other:
```bash
cosign verify --certificate-oidc-issuer https://token.actions.githubusercontent.com \
  --certificate-identity-regexp '<the tight on-tag.yaml@refs/tags pattern from F2>' \
  "ghcr.io/nvidia/aicr-validators/aiperf-bench@${REFERRER}"
```

tools/generate-aiperf-source
```diff
@@ -0,0 +1,409 @@
...
+    # becomes the layer's org.opencontainers.image.title, so an absolute path
+    # would publish the builder's local directory layout into the artifact.
+    # Passing the basename keeps that annotation to just the archive name, which
+    # is also what a consumer wants to see. Preferred over
+    # --disable-path-validation, which would suppress the check and keep the leak.
+    result = subprocess.run(
+        [
+            "oras", "attach",
+            "--artifact-type", ARTIFACT_TYPE,
+            "--annotation", "org.opencontainers.image.title=Third-party Python source",
```
njhensley, line 397: **🟡 Minor — Re-running attach-source creates a duplicate referrer the gate can't detect**

The tar is deterministic (`reproducible_tar` + `_reset`), giving a stable *blob* digest. But the *referrer manifest* digest is not pinned: `oras attach` stamps `org.opencontainers.image.created` (current time) into the manifest by default, and the tool passes only `--annotation org.opencontainers.image.title=...`. So re-running the (deliberately re-runnable) `attach-source` job over identical input pushes a manifest with a **new** digest → a **second** referrer of the same `artifactType`. The verify step's `length >= 1` still passes, so the duplicate is invisible to the gate — and it feeds F1's break for benign re-runs, self-inflicting the very ambiguity the signing is meant to resolve.

Confidence caveat: this hinges on oras v1.3.0 emitting a per-run `created` annotation (long-standing default oras behavior, but not verified live here). F1 case (a) does not depend on it.

**Fix:** pin the manifest timestamp for idempotency, e.g. add `--annotation org.opencontainers.image.created=1970-01-01T00:00:00Z` to the `oras attach` args (or assert the discovered signed-referrer count is exactly 1 in the verify step).

tools/generate-aiperf-source
```diff
@@ -0,0 +1,409 @@
...
+    """
+    def lookup() -> dict:
+        with urllib.request.urlopen(
+            PYPI_JSON.format(name=name, version=version), timeout=30
+        ) as response:
+            return json.load(response)
+
+    try:
+        data = with_retry(lookup, f"{name}=={version} metadata")
+    except (urllib.error.URLError, TimeoutError, ConnectionError, json.JSONDecodeError):
```
njhensley: **🟡 Minor — Metadata-fetch failures are reclassified as "no sdist published", yielding a misleading abort and a harmful suggested remediation**

`sdist_for` catches `(urllib.error.URLError, TimeoutError, ConnectionError, json.JSONDecodeError)` and returns `None`. Because `urllib.error.HTTPError` **is a subclass of `URLError`**, a persistent retryable 5xx (re-raised by `with_retry` after `MAX_ATTEMPTS`) *and* an unretried 403/401 both collapse to `None` — identical to a genuine "no sdist exists." That package lands in `unavailable`, and if not allowlisted, `build_archive` aborts (line 260) with `no source distribution published for: X ... Add the package to KNOWN_NO_SDIST ... or remove the dependency`.

For a package whose sdist really exists, that message is wrong and its remediation is actively harmful: adding a genuinely source-bearing package to `KNOWN_NO_SDIST` permanently suppresses the source-publication obligation this tool exists to enforce. Note `download()` (the second phase) deliberately does *not* catch — so the asymmetry is real. It does stay fail-*closed* (aborts, ships nothing), which is why this is Minor, not Major.

**Fix:** narrow the catch to what actually means "no sdist" — treat only `HTTPError` with `code == 404` (and arguably `JSONDecodeError`) as `None`; re-raise every other exhausted-retry error so the run fails with the true network/server cause. Split the `except` so `HTTPError` is inspected before the broad `URLError` clause.

.github/workflows/on-tag.yaml
```diff
@@ -811,6 +811,129 @@ jobs:
...
+  # Deliberately NOT part of release-check, and deliberately after publish.
+  # Attaching depends on PyPI serving 129 sdists, which is an uncontrolled
+  # third party on the critical path of a release. Gating on it converts a PyPI
+  # blip, or a yanked package version, into a failed release and a release-day
+  # fire drill. Nothing is lost by attaching afterwards: a referrer binds to the
+  # image digest, and promotion does not change the digest, so the artifact
+  # lands on exactly the same subject either way. A failure here is re-runnable
+  # as a single job rather than requiring the tag to be recut.
+  #
+  # `if: always()` so an earlier non-blocking failure elsewhere still lets the
```
njhensley: **🔵 Nitpick — if: always() comment contradicts the actual success-gated if:**

The header comment (lines 845–847) says "`if: always()` so an earlier non-blocking failure elsewhere still lets the source attach," but the job's actual condition (line 852) is `if: needs.resolve-candidates.result == 'success' && needs.publish.result == 'success'` — no status function, so the implicit all-`needs`-succeed gate also applies. No functional bug (the success gate is the safer choice, and only `summary` (`if: always()`) depends on `attach-source`, so it can't block the release). But a future maintainer who trusts the "always()" comment and adds a non-blocking job to `needs` would find the job silently skipped. Update the comment to describe the real gating.

.github/workflows/on-tag.yaml
```diff
@@ -868,6 +1007,10 @@ jobs:
           HOMEBREW: ${{ needs.publish-homebrew.result }}
           DEPLOY: ${{ needs.deploy.result }}
           NOTIFY: ${{ needs.notify.result }}
+          # Non-gating by design (see the attach-source job), so the summary
+          # is the only place a failure surfaces. Omitting it would make a
```
njhensley: **🔵 Nitpick — Summary comment claims it is "the only place a failure surfaces" — a failed job also reddens the whole run**

The comment (line 1011) says the summary row "is the only place a failure surfaces." But `attach-source` runs after `publish` succeeds and fails loudly (`exit 1`) on an attach/discoverability failure, so its own job — and thus the overall workflow run — goes red on an already-published release. A failure therefore surfaces in two places (red job/run + the summary row), which is the intended "fails loudly" behavior but not what the comment says. Consider softening the comment (and, optionally, whether a red run on a fully-published release is the signal you want vs. a step-summary warning + separate alert).

.settings.yaml
```diff
@@ -46,6 +46,11 @@ security_tools:
   cosign: 'v3.1.2'
   # renovate: datasource=github-releases depName=anchore/syft depType=security_tools
   syft: 'v1.50.0'
+  # renovate: datasource=github-releases depName=oras-project/oras depType=security_tools
+  # Attaches the aiperf-bench third-party source archive as an OCI 1.1 referrer
+  # (tools/generate-aiperf-source). Cosign cannot attach arbitrary artifacts, so
+  # this is a separate dependency from the signing/attestation path.
+  oras: 'v1.3.0'
```
njhensley: **🔵 Nitpick — oras pinned in .settings.yaml + load-versions but not installed by tools/setup-tools**

`oras: 'v1.3.0'` is a declared `security_tools` pin (exposed via `load-versions`), but `tools/setup-tools` has no oras install stanza and `tools/check-tools` has no metadata for it (it is silently skipped, exactly like `cosign`, and `make tools-check` is non-fatal). So there's no local-dev or gate breakage — CI installs oras via the SHA-pinned `setup-oras` action, and the tool errors clearly via `shutil.which("oras")` if run locally. For parity with grype/syft/crane, either add an oras stanza to `tools/setup-tools` or note it's CI-only.

docs/integrator/supply-chain-verification.md
```diff
@@ -154,6 +158,68 @@ specification's referrers *tag* schema (a `sha256-{hex}` tag holding an index of
...
+  "${IMAGE%:*}@${SOURCE}"
+
+# 4. Pull the verified archive.
+oras pull -o ./aiperf-source "${IMAGE%:*}@${SOURCE}"
```
njhensley, line 214: **🔵 Nitpick — oras missing from the docs Prerequisites tool list**

this comment is on the `oras pull` line, but the gap is in **Prerequisites** (docs lines 14–16). That list names Cosign, `gh`, `crane`, `jq`, `kubectl` — `crane` was correctly added, but `oras` (used here by `oras discover`/`oras pull`) is not. Add `oras` ([oras.land](https://oras.land/docs/installation)) to the Prerequisites sentence.

tools/generate-aiperf-source
```diff
@@ -0,0 +1,409 @@
...
+        got_sha256 = hashlib.sha256(payload).hexdigest()
+        if got_sha256 != want_sha256:
+            raise SystemExit(
+                f"ERROR: {name} sdist failed integrity check.\n"
+                f"  expected sha256 {want_sha256}\n"
+                f"  received sha256 {got_sha256}\n"
+                "Refusing to publish source that does not match what the index declares."
+            )
+
+        (staging / filename).write_bytes(payload)
```
njhensley, line 337: **🔵 Nitpick — sdist written under index-supplied filename without basename normalization**

`(staging / filename).write_bytes(payload)` uses `entry["filename"]` from the PyPI JSON verbatim as a path component. A filename containing path separators or `..` would resolve outside `staging`. The bytes are sha256-verified and PyPI enforces strict `{name}-{version}` filename shapes over TLS, so there's no realistic trigger — but the write trusts the field's structure, not just its bytes. Cheap hardening: `staging / pathlib.PurePosixPath(filename).name`, or assert `filename == os.path.basename(filename)` before writing.

.github/workflows/on-tag.yaml
```diff
@@ -811,6 +811,129 @@ jobs:
   # =============================================================================
   # Notify: Post release announcement to Slack
```
njhensley, line 812: **🔵 Nitpick — Scrambled section banners around the inserted attach-source job**

The `attach-source` job was inserted directly beneath the `# Notify: Post release announcement to Slack` banner (line 812), so that banner now sits above `attach-source` while the real `notify` job (far below) is preceded only by a bare `====` rule. Move the "Notify" banner down to immediately precede the `notify` job and give `attach-source` its own titled banner.

docs/integrator/supply-chain-verification.md
lockwobr: Fixed in `559f4271`. You're right that the documented retrieval was the weak link — `SOURCE=$(...)` captures multiple lines when there are two matches, so the very case the signature defends against is the one that breaks the commands.

Now uses `mapfile` and fails unless there is exactly one match, with a comment explaining that artifactType is not a unique key.

docs/integrator/supply-chain-verification.md
lockwobr: Fixed in `559f4271`. Agreed, and it was inconsistent with the doc's own standard three sections up — the SBOM and VEX commands pin `--certificate-identity "${AICR_SIGNER}"` exactly.

Now pins `https://github.com/NVIDIA/aicr/.github/workflows/on-tag.yaml@refs/tags/${TAG}` (this artifact is signed by `on-tag.yaml`, not `attest-images.yaml`) instead of a regexp that would have accepted any workflow in the repo.

.github/workflows/on-tag.yaml
lockwobr: Fixed in `559f4271`. Good catch — signing and then only checking discoverability meant a silently-failed sign would still pass the gate and ship an artifact the documented retrieval would reject.

The verify step now runs `cosign verify` against the referrer digest with the same pinned identity.

tools/generate-aiperf-source
lockwobr: Fixed in `559f4271`, and thank you — this one was self-inflicted on the path I'd deliberately designed as the recovery story.

Your read on `org.opencontainers.image.created` is right: the tarball is byte-reproducible but the manifest is not, so a re-run pushes a second referrer of the same type and `length >= 1` cannot see it.

Two changes: the attach step now discovers first and skips if a source referrer is already present (so a re-run is idempotent), and the gate asserts **exactly one** rather than at least one.

tools/generate-aiperf-source
lockwobr: Fixed in `559f4271`. This was the one I was most glad you caught — the misleading message was bad, but the remediation it suggested was actively harmful. Adding a package that genuinely publishes source to `KNOWN_NO_SDIST` would permanently suppress publishing it.

`sdist_for` no longer catches the network errors. Since `HTTPError` subclasses `URLError`, that `except` was collapsing a persistent 5xx and an unretried 403 into the same `None` a wheel-only package returns. Transport failures now propagate and abort with the actual error; `None` means only that the index lists no sdist. Verified both paths.

.github/workflows/on-tag.yaml
lockwobr: Fixed in `559f4271` — the comment was left over from an earlier revision where the job was `if: always()`. It now describes the actual gate: resolve-candidates and publish succeeding, since there is no image to attach to otherwise.

.github/workflows/on-tag.yaml
lockwobr: Fixed in `559f4271`. Fair — the job does red on its own. Reworded to say the summary is where the release outcome is read as a whole, so a missing row leaves "did the source publish?" unanswered there, rather than claiming it is the only signal.

.settings.yaml
lockwobr: Fixed in `559f4271`. Added to `tools/setup-tools` (brew on darwin, release tarball on linux, mirroring the syft block) and a `case` arm in `tools/check-tools`. `check-tools` enumerates `.settings.yaml` keys dynamically, so the `security_tools.oras` entry registers it automatically.

docs/integrator/supply-chain-verification.md
lockwobr: Fixed in `559f4271` — added to the Prerequisites list. The retrieval steps ask users to run `oras` and `cosign`, so it should have been there.

lockwobr: Correction to my previous reply: I said this was done, but it was not — the Prerequisites list still omitted ORAS when I wrote that. Now actually added in `ea2b1a03`, alongside `crane` and `jq`, noted as needed only for retrieving the source archive.

Apologies for the premature claim.

.settings.yaml
yuanchen8911: setup-oras at the pinned SHA (v2.0.1) performs an exact lookup of the version input against its bundled release table, whose keys are bare versions ('1.0.0'…'1.3.3', default '1.3.1') with no v-prefix normalization (https://github.com/oras-project/setup-oras/blob/1d808f7d7f6995cc68b7bf507bfe5c5446e1dc9d/src/lib/release.ts#L43-L50). Passing 'v1.3.0' throws "official ORAS CLI releases does not contain version v1.3.0", so the Setup ORAS step fails on every tag run and the referrer is never attached. The job failure is visible in the release summary, but the release has already published, so the obligation goes undischarged until a manual fix. Suggest pinning bare '1.3.0' and having oras join the existing extractVersion renovate rule.

Adjacent problem on this key: the renovate annotation above is currently dead. The custom manager in .github/renovate.json5 requires the value line immediately after the annotation line, and the three explanatory comment lines in between break the match. Moving the prose above the annotation line fixes it.

tools/generate-aiperf-source
```diff
@@ -0,0 +1,416 @@
...
+    for line in PYTHON_NOTICES.read_text(encoding="utf-8").splitlines():
+        if line.startswith("## Python Package Index"):
+            in_index = True
+            continue
+        if in_index and line.startswith("## "):
+            break
+        if in_index:
+            match = INDEX_ROW.match(line)
+            if match:
+                packages.append((match.group(1), match.group(2)))
```
yuanchen8911: The closure is read from python-notices.md, which has no pip row — but pip ships in the image: the builder creates the venv (which includes pip), upgrades it (`pip install --upgrade pip`), and the entire venv is copied into the final stage on both released architectures (validators/performance/aiperf-bench.Dockerfile L67-L87, L107-L112). The archive therefore omits the source of a third-party package the image distributes, which is the obligation this PR implements. Either remove pip from the shipped venv (nothing needs it at runtime, and it trims CVE surface) or get pip's exact installed version into the shared closure via the notices generator.

tools/generate-aiperf-source
```diff
@@ -0,0 +1,416 @@
...
+    walks directories in sorted order and _reset zeroes per-file metadata.
+    """
+    with open(path, "wb") as raw:
+        with gzip.GzipFile(filename="", mode="wb", fileobj=raw, mtime=0) as gz:
+            with tarfile.open(fileobj=gz, mode="w") as tar:
+                yield tar
+
+
+def build_archive(out_dir: pathlib.Path) -> tuple[pathlib.Path, int, list[str]]:
+    packages = read_closure()
```
yuanchen8911, line 283: The archive and the image resolve their package sets at different times. requirements.txt pins only aiperf==0.11.0, and aiperf itself declares many ranged dependencies, so the Dockerfile's `pip install -r` resolves the transitive closure at image build time. The build does use a persistent BuildKit cache (`cache-from/to: type=gha`), but an ordinary cache miss or invalidation — a base-image digest bump, a Dockerfile edit, cache eviction — is enough for pip to resolve a different closure than the committed notices snapshot this tool reads. TestCommittedFragmentMatchesRequirements only hashes the unchanged top-level requirements file, so it cannot see transitive drift. The result is an archive that may not correspond to the image digest it is attached to — the exact correspondence the referrer binding is supposed to guarantee. A complete lock file (or an inventory taken from the built image) as the single input for both the install and this tool would close it. If deferred, it needs a follow-up issue and a note in the docs section, since the same drift already affects the attribution record.

tools/generate-aiperf-source
```diff
@@ -0,0 +1,416 @@
...
+
+    print(f"Attaching to {image}...", file=sys.stderr)
+    # Run from the archive's directory and pass the bare filename. `oras attach`
+    # rejects absolute paths outright, and for good reason: the file argument
+    # becomes the layer's org.opencontainers.image.title, so an absolute path
+    # would publish the builder's local directory layout into the artifact.
+    # Passing the basename keeps that annotation to just the archive name, which
+    # is also what a consumer wants to see. Preferred over
+    # --disable-path-validation, which would suppress the check and keep the leak.
+    result = subprocess.run(
```
yuanchen8911, line 393: `subprocess.run` here has no timeout, so the documented local `--attach` invocation can hang indefinitely on a registry connection that stalls after connecting; the 20-minute Actions deadline only bounds CI. A named timeout with TimeoutExpired handled would match the repo's bounded-I/O rule. Small fix that directly supports the new publication path — worth doing in this PR.

.github/workflows/on-tag.yaml
```diff
@@ -811,6 +811,162 @@ jobs:
...
+
+      - name: Build and attach source archive
+        id: attach
+        env:
+          DIGEST_MAP: ${{ needs.resolve-candidates.outputs.digest_map }}
+        run: |
+          set -euo pipefail
+          digest="$(jq -er '."aiperf-bench"' <<<"${DIGEST_MAP}")"
+          image="ghcr.io/nvidia/aicr-validators/aiperf-bench@${digest}"
+          # Skip if a source referrer is already attached to this digest.
```
yuanchen8911, line 894: This skip closes the duplicate-referrer ambiguity, but it opens a recovery trap in the window it exists to serve. If a run attaches successfully and then fails at the sign step (or the job dies between attach and sign), the digest carries an unsigned referrer. Every re-run then takes this branch (a referrer of the type exists), sets skipped=true, skips signing, and fails the verify step's cosign check on the unsigned referrer — permanently, until someone manually deletes it. That contradicts the job comment above, which still says "A failure here is re-runnable as a single job rather than requiring the tag to be recut." One way out: when exactly one referrer of the type exists but its signature does not verify, sign that existing digest instead of skipping, so a re-run completes the interrupted attach-then-sign sequence rather than wedging on it.

.github/workflows/on-tag.yaml
```diff
@@ -811,6 +811,162 @@ jobs:
   # =============================================================================
   # Notify: Post release announcement to Slack
```
yuanchen8911, line 812: The new block is inserted between the "Notify: Post release announcement to Slack" banner and the notify job, so the banner now heads attach-source and notify sits under a bare separator. Moving the new block after notify (or moving the banner down with its job) keeps the section headers navigable.

docs/integrator/supply-chain-verification.md
```diff
@@ -154,6 +159,77 @@ specification's referrers *tag* schema (a `sha256-{hex}` tag holding an index of
 referrers). Cosign and ORAS do this transparently, which is why the commands
 above are the documented path rather than a raw referrers API call.
 
+## Third-Party Source Code
```
yuanchen8911, line 162: Inserting this H2 above the pre-existing "### Using the published VEX document" and "### Why provenance uses a different signer" re-parents both under Third-Party Source Code in the rendered outline, where they don't belong (they were children of Unified Metadata Retrieval). Moving the new section below them fixes the hierarchy; anchor slugs are unchanged so no links break.

yuanchen8911 (changes requested):
The design here is solid — reproducible archive, fail-closed source resolution, checksum verification against the index's own declarations, signing the referrer digest, and the non-gating placement is well argued. The latest push (ea2b1a0) already resolved several review items: the documented verification now pins the exact signing identity, the discoverability gate requires exactly one referrer and verifies its signature, sdist_for no longer conflates transport failures with a missing sdist, the stale always() comment is reworded, and oras is covered in prerequisites, check-tools, and setup-tools.

Three blocking issues remain, and the new skip logic introduces one more:

1. The attach-source job still fails on every release: .settings.yaml pins `oras: 'v1.3.0'`, and setup-oras at the pinned SHA performs an exact lookup against bare version keys, so `v1.3.0` throws at the Setup ORAS step. The failure is visible in the release summary, but the release has already published by then, so the compliance artifact is absent until a manual fix — and as pinned, a re-run fails the same way.

2. The source closure omits pip, which ships in the image: the builder upgrades pip inside the venv and the entire venv is copied into the final stage on both architectures, while the notices closure the archive derives from has no pip entry.

3. The archive is not locked to the subject image. requirements.txt pins only aiperf, which itself declares many ranged dependencies; a BuildKit cache miss or invalidation resolves a potentially different closure at image build than the committed notices snapshot the archive is built from, and the freshness test only hashes the unchanged top-level file. That breaks the per-digest correspondence the referrer binding exists to guarantee.

4. New in this push: the skip-if-already-attached logic breaks the job's own re-runnable recovery claim for the attach-succeeded-but-sign-failed window — every re-run then skips attach and sign and fails verification on the unsigned referrer, permanently, until someone deletes it manually. Details inline.

Suggested disposition: block on the setup-oras fix, a single locked closure (including or removing runtime pip), and the re-run recovery gap. The subprocess timeout is small and directly supports the new publication path, so it is worth fixing here too. The remaining items (renovate annotation placement, section banner, doc heading hierarchy) are quick cleanups.

Items 2 and 3 are arguably inherited from the notices pipeline rather than introduced here, but this PR is what turns them into a per-digest correspondence claim, so they need settling in scope or an explicit deferral issue.

tools/generate-aiperf-source
lockwobr: This needs its own issue, this also has an effect on cves, so going open an issue for this.

lockwobr:
Pushed `ae7e6997`. All four blocking items addressed, with one deliberately deferred to an issue rather than fixed here.

| # | Item | Disposition |
|---|---|---|
| 1 | `setup-oras` rejects `v1.3.0` | **Fixed.** Confirmed against the action: it defaults to a bare `1.3.1` and its README uses `version: 1.0.0`, so the `v` prefix would have thrown at Setup ORAS on every release. `.settings.yaml` now pins `1.3.0` bare, with a comment explaining why it differs from the `v`-prefixed cosign/syft/grype entries beside it. `tools/setup-tools` re-adds the prefix for the download URL. |
| 2 | `pip` ships but is unsourced | **Documented, not fixed.** `pip` and `wheel` come from `python -m venv`, so no resolution of `requirements.txt` will surface them, and pinning them was explicitly rejected to keep CVE fixes arriving automatically. Named as a known gap in the docs and the tool's module docstring rather than left silent. Happy to take a different disposition if you think it should block. |
| 3 | Archive not locked to the subject image | **Fixed differently than suggested.** Rather than pinning, the closure is now resolved from the same `requirements.txt` the Dockerfile installs, via `pip install --dry-run --report` (~90s, no downloads). Both artifacts derive from one input and the dependencies stay floating. |
| 4 | Skip logic breaks re-run recovery | **Fixed**, exactly as you suggested. A re-run now reuses the existing referrer digest and re-signs it instead of skipping, so an attach-succeeded-then-sign-failed run completes the interrupted sequence. |

**On item 3, one thing your review made me look at that turned out to be worse than the drift itself:** the image build layer-caches its dependency resolution. `RUN pip install -r requirements.txt` is keyed on the base image plus the one-line `requirements.txt`, and all three workflows that build this image use `cache-from`/`cache-to: type=gha`. I verified empirically that a `RUN` layer is replayed rather than re-executed.

So while the cache hits, the image is not picking up CVE fixes at all — which is the entire reason the deps are unpinned — and a cached image can carry an older closure than a freshly resolved archive.

I had a fix for that in this branch and backed it out: it changes build behaviour across three workflows including the weekly vulnerability scan, adds several minutes per architecture per build, and is pre-existing rather than introduced here. Filed as #2086 with the evidence and the proposed patch.

Also fixed from the inline comments: sdist selection is now sorted rather than first-match, and the doc's retrieval rejects anything but exactly one referrer.

Thank you for the review — items 2 and 4 in particular were things I would not have found. Item 4 was a regression I introduced while fixing the duplicate-referrer problem, and the `KNOWN_NO_SDIST` point on `sdist_for` was the one where the misleading message came with actively harmful remediation.

.github/workflows/on-tag.yaml
lockwobr: Fixed in `ae7e6997`, exactly as you suggested. The attach step now reuses an existing referrer's digest and falls through to signing, rather than setting `skipped=true` and short-circuiting.

Traced it: run 1 attaches R and fails at sign; run 2 discovers R, reuses that digest, signs it, verification passes. The job comment's re-runnable claim holds again.

tools/generate-aiperf-source
lockwobr: Here is the issue https://github.com/NVIDIA/aicr/issues/2086

yuanchen8911 (approved):
Verified the fixes in ae7e6997 against the earlier review; approving. CI is green on this commit.

What was verified:

- The oras pin is now bare ('1.3.0'), matching setup-oras's release-key lookup, with the rationale recorded in .settings.yaml and setup-tools re-adding the prefix for the download URL. The release-breaking failure is gone.
- read_closure now resolves requirements.txt directly with the builder's Python version and platform tags, so the archive and the image derive from one input. The residual layer-cache gap is documented in the docs and tracked in #2086.
- The attach step now reuses an existing referrer's digest and signing runs unconditionally, so an interrupted attach-then-sign sequence completes on re-run instead of wedging on an unsigned referrer.
- The pip/wheel gap is now stated explicitly in the docs rather than implied away.

Non-blocking follow-ups, fine to land as-is or fold into a small cleanup:

- The renovate annotation on the oras key still does not match: the custom manager requires the value line immediately after the annotation line, and the comment block sits in between. With the bare value, oras also needs to join the extractVersion packageRule (upstream tags are v-prefixed).
- The pip/wheel bullet in the docs has no tracking issue, unlike the caching bullet next to it — worth filing one so "tracked rather than claimed away" holds for both.
- subprocess.run for oras attach (and the new pip resolve) still has no timeout; the open review threads on the section banner and the doc heading hierarchy also remain.

PR #2079, opened 2026-08-05
feat(slinky): add shared RWX storage support for home and data

kaynetu:
## Summary

Adds opt-in shared RWX storage for Slinky Slurm home and data directories. Generated bundles can now create the two PVCs before the Slinky release and mount them consistently into configured LoginSets and NodeSets.

## Motivation / Context

Slinky Slurm login and worker pods need shared `/home` and `/scratch/fsw` filesystems, but AICR previously required operators to create PVCs and patch chart values manually. This change provides a validated recipe and CLI contract while leaving existing bundles unchanged unless shared storage is explicitly enabled.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: Slinky component values, recipe overlays, and Chainsaw coverage

## Implementation Notes

- Adds an opt-in `storage` contract with explicit `home` and `data` PVC definitions. PVC names must be valid and distinct, storage quantities must be positive, unsupported keys are rejected, and resolved string values are normalized before rendering.
- Adds `--shared-storage-class` and the matching config field as a common default while preserving per-volume `storageClassName` overrides. CLI and config values receive equivalent trimming and blank-value validation, and repeated scalar flags are rejected.
- Emits the PVC resources through an ordered `slinky-slurm-pre` bundle component so they can be installed before the Slinky release. Empty pre-install output is omitted. Helm retention preserves the claims during release upgrade/removal; Argo CD retention preserves them during Application deletion and automated sync pruning.
- Injects `/home` and `/scratch/fsw` mounts into every configured LoginSet and NodeSet. This intentionally includes sets with `enabled: false` so later Helm enablement preserves the shared-storage contract.
- Requests `ReadWriteMany`; the selected StorageClass and backing storage system must support RWX access.
- Adds user, CLI, component, and contributor documentation plus unit and Chainsaw regression coverage.

## Testing

```bash
env -u GITLAB_TOKEN make qualify
golangci-lint run -c .golangci.yaml ./pkg/bundler/...
```

- `make qualify`: passed locally, including the race-enabled Go suite, 81.7% project coverage (80% threshold), full lint, 24/24 Chainsaw tests, vulnerability scan, license checks, and repository checks.
- Affected-package race tests: passed.
- Affected-package lint: passed with zero issues.

## Risk Assessment

**Rollout notes:** Shared storage is disabled by default, so existing generated bundles are unchanged. When enabled, operators must select a StorageClass and backend that support `ReadWriteMany`. The generated claims are retained across Helm release removal, Argo CD Application deletion, and Argo CD automated pruning, but explicit PVC or namespace deletion can still delete backing data depending on the PersistentVolume reclaim policy.

## Checklist


recipes/components/slinky-slurm/manifests/shared-storage-pvcs.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+{{- $component := index .Values "slinky-slurm" -}}
+{{- with (index $component "storage") }}
+{{- if (index . "enabled") }}
+{{- range $key, $pvc := . }}
+{{- if ne $key "enabled" }}
+---
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
```
njhensley, line 23: **🟠 Major — Shared PVCs are deleted by the documented `helm uninstall` teardown → home/scratch data loss**

These are standalone `PersistentVolumeClaim` objects inside the injected `NNN-slinky-slurm-pre` local-helm chart, with no `helm.sh/resource-policy: keep` annotation. The shipped uninstall doc (`pkg/bundler/deployer/helm/templates/README.md.tmpl`) instructs `helm uninstall` for every folder including `*-pre` auxiliaries — and `helm uninstall` deletes standalone (non-StatefulSet) PVCs, destroying `/home` and `/scratch/fsw`. `docs/user/cli-reference.md:2647` further reassures that "Helm intentionally does not delete PVCs (StatefulSet-managed volumes are preserved)" — true only for StatefulSet PVCs, so it actively misleads for this feature, and the PR summary's "not dynamically removed" claim is contradicted by the packaging.

**Blast radius:** Any Slurm deploy with shared storage enabled that is later uninstalled per the shipped docs loses all user home + scratch data (and, under a Delete reclaimPolicy such as EFS access-point provisioning, the backing data).

**Fix:** Add `annotations: {helm.sh/resource-policy: keep}` to both PVCs, and add a lifecycle/data-loss section to `slinky-slurm-storage.md` (undeploy retains PVCs; re-enable reuses claims/data; deleting a PVC may delete backing data depending on reclaimPolicy).

pkg/bundler/slinky_storage.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	if !enabled {
+		return nil
+	}
+	if !supported {
+		return errors.New(errors.ErrCodeInvalidRequest,
+			"slinky-slurm shared storage is not supported by this recipe")
+	}
+
+	volumes := make([]slinkySharedVolume, 0, 2)
```
njhensley, line 173: **🟡 Minor — Duplicate home/data PVC `name` is not rejected**

Each `storage.<k>.name` is DNS1123-validated, but nothing checks `storage.home.name != storage.data.name`. `claimName` is user-settable while `volumeName` is hardcoded (`shared-home`/`shared-data`), so `appendNamedObject`'s dedup (keyed on volumeName) never fires. Setting both names equal renders two PVCs with the same `metadata.name`/namespace — the bundle succeeds, then fails at apply (helm duplicate-resource) or silently collapses `/home` and `/scratch/fsw` onto one claim. (Independently flagged by 3 reviewers; downgraded from major because defaults are distinct, so it only bites deliberate name overrides.)

**Blast radius:** Slinky bundles where an operator overrides both per-volume names to a colliding value; broken/ambiguous bundle surfaces at cluster apply, not at bundle time.

**Fix:** Reject equal claimNames after the volume-parse loop (ErrCodeInvalidRequest), and add a `TestMaterializeSlinkySharedStorage` case.

pkg/bundler/slinky_storage.go
```diff
@@ -0,0 +1,333 @@
...
+	if err != nil {
+		return slinkySharedVolume{}, err
+	}
+	if validationErrors := validation.IsDNS1123Subdomain(claimName); len(validationErrors) > 0 {
+		return slinkySharedVolume{}, errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("slinky-slurm storage.%s.name %q is invalid: %s",
+				key, claimName, strings.Join(validationErrors, "; ")))
+	}
+
+	storageClassName, err := requiredString(config, "storageClassName", "storage."+key)
```
njhensley: **🟡 Minor — `storageClassName` presence-checked but not format-validated; trimmed value discarded**

`name` fails closed via `IsDNS1123Subdomain` and `size` via `ParseQuantity`, but `storageClassName` is only presence-checked and its trimmed return is discarded (`if _, err = requiredString(...)`). The raw, untrimmed value renders unquoted into the PVC YAML (`storageClassName: {{ index $pvc "storageClassName" }}`), so a malformed class (trailing space, colon, embedded newline) corrupts or injects into the manifest — a fail-open asymmetry versus the sibling fields.

**Blast radius:** Operator-supplied value, so the blast radius is the operator's own bundle — defense-in-depth rather than cross-tenant. A typo (trailing space/colon) silently ships a broken PVC that fails at helm install.

**Fix:** Validate the resolved class with `IsDNS1123Subdomain` (mirroring `name`) and use the trimmed value; optionally quote it in the template.

pkg/bundler/slinky_storage.go
```diff
@@ -0,0 +1,333 @@
...
+				slinkyMountTarget{
+					path:   setPath + "." + definition.containerKey + ".volumeMounts",
+					parent: container,
+					key:    "volumeMounts",
+					makeEntry: func(volume slinkySharedVolume) map[string]any {
+						return map[string]any{slinkyNameKey: volume.volumeName, "mountPath": volume.mountPath}
+					},
+				},
+				slinkyMountTarget{
+					path:   setPath + ".podSpec.volumes",
```
njhensley, line 105: **🟡 Minor — No deployed/UAT proof that injected `podSpec.volumes` merge (not replace) chart volumes**

materialize appends to `podSpec.volumes` / container `volumeMounts`, assuming the Slinky chart treats podSpec as passthrough-merge (precedent: nodeSelector/tolerations/resourceClaims injected the same way). Coverage is rendered-value unit + `--no-cluster` chainsaw only — nothing proves a LoginSet/NodeSet pod actually comes up with BOTH the shared mounts and the chart's own config/munge/spool volumes.

**Blast radius:** If merge semantics differ from the assumption, enabling shared storage could drop essential Slurm mounts on a real cluster; opt-in + default-off, so it would surface in UAT, not unit/chainsaw CI.

**Fix:** Add (or cite) a deployed/UAT assertion that a LoginSet/NodeSet pod has both the shared mounts and the chart's own volumes.

tests/chainsaw/cli/bundle-slinky-storage/chainsaw-test.yaml
```diff
@@ -0,0 +1,134 @@
...
+    - name: disabled-storage-renders-no-pvcs
+      try:
+        - script:
+            content: |
+              AICR_BIN="${AICR_BIN:?Set AICR_BIN to the path of the aicr binary}"
+              WORK="/tmp/chainsaw-bundle-slinky-storage"
+              ${AICR_BIN} bundle \
+                --recipe "${WORK}/recipe.yaml" \
+                --output "${WORK}/disabled"
+              ! ls -d "${WORK}"/disabled/[0-9][0-9][0-9]-slinky-slurm-pre >/dev/null 2>&1
```
njhensley, line 52: **🟡 Minor — Chainsaw `disabled-storage-renders-no-pvcs` is a fail-open negative check**

The step has no `set -e`; its exit status is the trailing `! ls -d .../slinky-slurm-pre`. If the `bundle` command itself fails, the pre-folder is never created, so `! ls` exits 0 and the `($error == null): true` check passes vacuously — a bundle regression for the default-off recipe would slip through. The sibling `enabled-storage-requires-class` step correctly captures `rc`.

**Blast radius:** Test-only: a regression breaking `aicr bundle` for the default-off Slurm recipe would not be caught by this step.

**Fix:** Add `set -e` after the `AICR_BIN` assignment (or capture and assert the bundle rc == 0) before the `! ls`.

pkg/bundler/slinky_storage.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	return append([]any(nil), entries...), nil
+}
+
+func appendNamedObject(entries []any, entry map[string]any, path string) ([]any, error) {
+	name := entry[slinkyNameKey]
+	for _, existing := range entries {
+		existingMap, ok := existing.(map[string]any)
+		if !ok {
+			return nil, errors.New(errors.ErrCodeInvalidRequest,
```
njhensley, line 334: **🟡 Minor — `appendNamedObject` idempotent-skip branch untested**

Only the reject path (existing `shared-home` with a different mountPath) is tested; the `reflect.DeepEqual(existingMap, entry)` idempotent-skip branch — the stated re-run / installer-pre-declared-mount contract — has no test. A regression flipping it to append-a-duplicate or error would go uncaught.

**Blast radius:** Idempotency/re-run safety of the mount-append logic is unverified.

**Fix:** Add a case pre-seeding `loginsets.slinky.login.volumeMounts` with the exact `{name: shared-home, mountPath: /home}` entry and assert len stays 2 (not 3).

pkg/bundler/slinky_storage.go
```diff
@@ -0,0 +1,333 @@
...
+// materializeSlinkySharedStorage validates enabled shared-storage values and
+// appends the matching volumes and mounts to Slinky's LoginSet and NodeSet
+// values. The PVC objects themselves are rendered by the component's
+// shared-storage-pvcs pre-manifest.
+func materializeSlinkySharedStorage(values map[string]any, supported bool) error {
+	storageValue, exists := values["storage"]
+	if !exists {
+		return nil
+	}
+	storage, ok := storageValue.(map[string]any)
```
njhensley, line 130: **🟡 Minor — Defensive error branches in `slinky_storage.go` untested**

Uncovered branches: `storage.enabled` non-bool, `storage` present but `enabled` absent (no-op), storage-not-object inside materialize, `objectField` non-object (container/podSpec), `namedObjectList` non-list, `appendNamedObject` non-object entries. Each returns `ErrCodeInvalidRequest` but none is asserted, so a message/code regression on malformed values ships uncaught.

**Blast radius:** Error-path robustness for malformed Helm values is unverified; low individually, broad in aggregate.

**Fix:** Add table rows mutating `validSlinkyStorageValues` to hit each branch (`enabled: "yes"`; drop `enabled`; `storage.slinky.login = "x"`; `volumeMounts = "x"`; `volumeMounts = ["x"]`).

recipes/components/slinky-slurm/manifests/shared-storage-pvcs.yaml
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+{{- $component := index .Values "slinky-slurm" -}}
+{{- with (index $component "storage") }}
+{{- if (index . "enabled") }}
+{{- range $key, $pvc := . }}
+{{- if ne $key "enabled" }}
+---
+apiVersion: v1
```
njhensley, line 21: **🔵 Nitpick — Template hardcodes `index .Values "slinky-slurm"` with no binding test**

The template keys values via the string literal `index .Values "slinky-slurm"` (relies on the writer setting `RenderInput.ComponentName = component.Name`), while Go gates on `const slinkySlurmComponentName`. Nothing binds the template literal to the registry name, so a future rename would silently drop the render (empty → pre-folder omitted).

**Blast radius:** Refactor-safety only; no current defect.

**Fix:** Add a small test asserting the template's `index .Values "..."` literal matches the registry's `slinky-slurm` component name / const.

pkg/bundler/slinky_storage_test.go
```diff
@@ -0,0 +1,346 @@
...
+		options  []config.Option
+		wantHome string
+		wantData string
+	}{
+		{
+			name: "generic class does not affect shared paths",
+			options: []config.Option{
+				config.WithStorageClass("gp3"),
+			},
+		},
```
njhensley, line 360: **🔵 Nitpick — No end-to-end test of a mixed per-volume storage class**

Override logic (`TestApplySharedStorageClassOverride` 'per-volume scalar override wins') and rendering (`TestSlinkySharedStorageManifest`) are covered independently, but no single test drives `--shared-storage-class X` + a per-volume `storageClassName Y` through materialize→template asserting home renders Y while data renders X.

**Blast radius:** The most operationally interesting mixed-class config is only half-covered end-to-end.

**Fix:** Add a manifest-render case with differing home/data classes, or extend the chainsaw test with a `--set slinkyslurm:storage.home.storageClassName=home-sc` alongside `--shared-storage-class`.

njhensley (commented):
## Multi-Persona Review

**Method:** 4 independent persona reviewers (Correctness/Go · Domain & Architecture · Security/Operability · Test-coverage) fanned out in parallel, then every finding was re-derived from the resolved code by an adversarial senior meta-reviewer that confirmed / refuted / re-tiered each. Anchored to head `0b8dab482`.

### Overall assessment

Well-built, cohesive feature. The core contract is genuinely fail-closed: `--shared-storage-class` never falls back to the possibly-RWO `--storage-class`, an enabled-but-classless config fails the bundle, shared vs. generic `storageClassPaths` are proven non-overlapping, and the PVC set can't desync from the injected volume references (both gate on `storage.enabled`). The gb200 dual-pre-manifest case is handled correctly, values are keyed properly under `.Values["slinky-slurm"]`, and the test suite (unit + chainsaw + registry pins) is solid.

One finding is worth addressing before merge — **🟠 the shared PVCs are destroyed by the documented `helm uninstall` teardown path (data loss), and the cli-reference PVC note misleads for this feature.** Everything else is polish.

### 🟡 Additional minor finding (not inline — `recipe.md` isn't in this diff)

**`sharedStorageClassPaths` missing from the `recipe.md` ComponentConfig table.** The PR documents the new field in `docs/contributor/component.md`, but the parallel integrator-facing ComponentConfig field table in [`docs/contributor/recipe.md` (~L98, next to `storageClassPaths`)](https://github.com/NVIDIA/aicr/blob/0b8dab482e4d9d8816bd0170906be174bd9021f1/docs/contributor/recipe.md#L98) still lists only `storageClassPaths`. Add the row to keep the mirror in sync (CLAUDE.md doc rule: field additions reflected across all enumerating surfaces).

### Confirmed non-issues (examined, not flagged)

- **`make bom-docs` regen not required** — this branch adds only 3 registry lines (no `defaultVersion` bump, no new component/image), so the BOM version+component-set gate is unaffected. (A `main…HEAD` three-dot diff shows unrelated version bumps that live on `main`, not this PR.)
- **Fail-closed RWX contract holds** — no fallback to `--storage-class`; enabled-without-class fails the bundle (verified in code and the chainsaw `enabled-storage-requires-class` step, which correctly captures `rc` + greps).
- **Non-overlap** of shared vs. generic paths is tested at the registry level *and* behaviorally.
- **PVC/volume can't desync** — both gate on `storage.enabled`; unsupported `storage.*` keys are rejected even when injected via `--set-json` (materialize runs last).
- **gb200 dual pre-manifest** — the empty PVC render is filtered out while the always-rendering ComputeDomain keeps the pre-folder.

### Summary

|:---------:|:-------:|:-------:|:---------:|
| 0 | 1 | 7 | 2 |

*Review method: parallel multi-persona panel + adversarial senior meta-reviewer. Not a substitute for CI; no code from this PR was executed.*

tests/chainsaw/cli/bundle-slinky-storage/chainsaw-test.yaml
```diff
@@ -0,0 +1,134 @@
...
+              set -e
+              WORK="/tmp/chainsaw-bundle-slinky-storage"
+              PVC_MANIFEST=$(ls "${WORK}"/enabled/[0-9][0-9][0-9]-slinky-slurm-pre/templates/shared-storage-pvcs.yaml)
+              awk -v work="${WORK}" '
+                BEGIN { document = 0 }
+                /^---$/ { document++; next }
+                { print > (work "/pvc-" document ".yaml") }
+              ' "${PVC_MANIFEST}"
+              test -s "${WORK}/pvc-0.yaml"
+              test -s "${WORK}/pvc-1.yaml"
```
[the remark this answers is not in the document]
kaynetu: addressed in new commit

pkg/bundler/slinky_storage.go
kaynetu: Verified against the exact upstream slurm-operator v1.2.0 implementation. The Helm chart passes podSpec.volumes and the login/slurmd volumeMounts into the LoginSet/NodeSet CRs. The operator then constructs its built-in pod volumes and container mounts as the base and applies the CR fields through Kubernetes StrategicMergePatch; volumes merge by name and volumeMounts merge by mountPath. Therefore the shared entries are added without replacing Slinky’s config, logfile, sackd, SSH, or SSSD mounts. I also validated the resulting deployment on the EKS GB200 cluster, so the source-level behavior has runtime confirmation.

pkg/bundler/slinky_storage_test.go
```diff
@@ -0,0 +1,403 @@
...
+				}),
+			},
+			wantHome: "home-sc",
+			wantData: "efs-sc",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			cfg := config.NewConfig(tt.options...)
```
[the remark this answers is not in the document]
kaynetu: Verified against the current code. This test imports `github.com/NVIDIA/aicr/pkg/bundler/config`; that package has no `WithNoCluster` option, so the suggested change does not compile (`undefined: config.WithNoCluster`). `bundler.New` only constructs its local config/attester (plus optional local attestation-file checks), and this test then calls `applySharedStorageClassOverride`, which reads the embedded component registry and mutates an in-memory values map. Neither path constructs a Kubernetes client or reads cluster credentials. Leaving this unchanged.

tests/chainsaw/cli/bundle-slinky-storage/chainsaw-test.yaml
```diff
@@ -0,0 +1,145 @@
...
+              AICR_BIN="${AICR_BIN:?Set AICR_BIN to the path of the aicr binary}"
+              WORK="/tmp/chainsaw-bundle-slinky-storage"
+              rm -rf "${WORK}" && mkdir -p "${WORK}"
+              ${AICR_BIN} recipe \
+                --service eks \
+                --accelerator h100 \
+                --os ubuntu \
+                --intent training \
+                --platform slurm \
+                --output "${WORK}/recipe.yaml"
```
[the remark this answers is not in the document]
kaynetu: Verified both supported flag positions with the built CLI: `aicr --no-cluster recipe --help`, `aicr recipe --no-cluster --help`, and the equivalent `bundle` commands all exit 1 with `flag provided but not defined: -no-cluster`. Source defines `--no-cluster` only for `validate` (`pkg/cli/validate.go`). The E2E harness already invokes `chainsaw test --no-cluster` in `tools/e2e`; the full `make qualify` run logged `NoCluster true`, and `cli-bundle-slinky-storage` passed. Adding the flag to these recipe/bundle commands would break the test, so I am leaving them unchanged.

pkg/bundler/bundler.go
```diff
@@ -1099,6 +1102,13 @@ func (b *DefaultBundler) extractComponentValues(ctx context.Context, recipeResul
 			}
 		}
 
+		if ref.Name == slinkySlurmComponentName {
+			supported := hasSlinkySharedStoragePreManifest(ref.PreManifestFiles)
+			if materializeErr := materializeSlinkySharedStorage(values, supported); materializeErr != nil {
+				return nil, materializeErr
+			}
+		}
```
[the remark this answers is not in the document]
kaynetu: Verified against the current call path. `extractComponentValues` already checks `ctx.Err()` at the start of every component iteration, immediately before resolving and materializing that component (`pkg/bundler/bundler.go:1016-1019`). The shared-storage materializer performs no I/O, waits, retries, or blocking operations: its storage-definition loop is fixed at 2 entries (home/data), its set-kind loop is fixed at 2 entries (LoginSets/NodeSets), and each mount-target update adds exactly 2 volumes. The only variable collections are the chart's configured set names and pre-existing mount entries; under the Slinky values contract these are small in-memory maps/slices, not a long-running operation.

Threading `ctx` through every helper and fixed-size loop would not make the potentially larger operations (`component.DeepCopyMap` and `sort.Strings`) interruptible, so it would add API churn without providing the cancellation guarantee described by the finding. Cancellation is already observed at the component boundary before this synchronous, bounded transformation. Leaving this unchanged.

recipes/components/slinky-slurm/manifests/shared-storage-pvcs.yaml
kaynetu: Addressed in `522ef1b4`. The generated home and data PVCs now carry `helm.sh/resource-policy: keep` plus `argocd.argoproj.io/sync-options: Delete=false,Prune=false`: Helm/Helmfile/Flux retain them during release upgrade/removal, `Delete=false` protects Argo CD Application deletion, and `Prune=false` protects manual or automated sync pruning when storage is disabled. `TestSlinkySharedStorageManifest` asserts both annotations, and the Chainsaw home/data PVC assertions verify them in the generated bundle. `docs/user/slinky-slurm-storage.md` now documents retention, claim reuse, Helm ownership/adoption, Argo `OutOfSync` behavior, explicit PVC/namespace deletion, and reclaim-policy data-loss risk; the generic uninstall guidance in `docs/user/cli-reference.md` was corrected as well. Full `make qualify` passed on the rebased commit with 81.7% coverage and 24/24 Chainsaw tests.

pkg/bundler/slinky_storage.go
kaynetu: Fixed in `522ef1b4`. StorageClass values are trimmed, DNS-1123 validated, normalized into rendered values, and quoted; tests cover invalid and whitespace cases.

tests/chainsaw/cli/bundle-slinky-storage/chainsaw-test.yaml
kaynetu: Fixed in `522ef1b4`. `set -e` makes a bundle failure fail the negative check; the Chainsaw scenario passes.

pkg/bundler/slinky_storage.go
kaynetu: Fixed in `522ef1b4`. Equal normalized home/data claim names now return `INVALID_REQUEST`, with a collision test.

recipes/components/slinky-slurm/manifests/shared-storage-pvcs.yaml
kaynetu: Fixed in `522ef1b4`. Retention annotations and lifecycle tests/docs now cover Helm removal, Argo CD deletion/pruning, claim reuse, and reclaim-policy risk.

pkg/bundler/slinky_storage.go
kaynetu: Fixed in `522ef1b4`. The exact-existing-mount case proves the idempotent skip and keeps the final list at two entries.

pkg/bundler/slinky_storage_test.go
kaynetu: Fixed in `522ef1b4`. The mixed-class test drives override through materialization and rendering, then verifies distinct home and data StorageClasses.

pkg/bundler/slinky_storage.go
kaynetu: Fixed in `522ef1b4`. Tests cover missing/non-boolean enablement, malformed objects/lists, and non-object entries while asserting `INVALID_REQUEST`.

pkg/bundler/slinky_storage.go
kaynetu: Verified. Slinky v1.2.0 uses StrategicMergePatch for volumes and mounts, and EKS GB200 UAT confirmed shared and built-in mounts coexist.

recipes/components/slinky-slurm/manifests/shared-storage-pvcs.yaml
kaynetu: Fixed in `522ef1b4`. The binding test checks the registry component name against the template literal and asserts that two PVCs render.

kaynetu:
Also addressed the non-inline documentation note in `522ef1b4`: `docs/contributor/recipe.md` now lists `sharedStorageClassPaths` beside `storageClassPaths`.

pkg/bundler/slinky_storage.go
```diff
@@ -0,0 +1,347 @@
...
+			}
+			container, err := objectField(set, definition.containerKey, setPath)
+			if err != nil {
+				return nil, err
+			}
+			podSpec, err := objectField(set, "podSpec", setPath)
+			if err != nil {
+				return nil, err
+			}
+			targets = append(targets,
```
njhensley, line 95: **🔵 Nitpick — No in-repo guard that the Slinky chart merges (not replaces) injected volumes**

The injected podSpec.volumes / container.volumeMounts land as Helm *values*; every test asserts only that the values render, not that the upstream Slinky chart strategic-merges them alongside its own chart-native volumes. If a future chart bump changes list-merge semantics, the PVCs are created but never mounted — a silent empty /home and /scratch/fsw. Un-guardable in CI (the chart isn't vendored) and opt-in / default-off, so it surfaces only in UAT. This is the residual of prior finding #4 (correctly left ◐ Partially addressed — the merge behavior is closed by your cited Slinky v1.2.0 StrategicMergePatch + EKS GB200 UAT evidence, not by an in-repo assertion).

**Blast radius:** All login + compute pods across the 4 shipped Slurm recipes, only when shared storage is opted in.

**Fix:** Add a UAT / rendered-CR assertion that a LoginSet/NodeSet pod actually carries both the shared mounts and the chart's own volumes, or pin the exact upstream values.schema.json keys relied on.

docs/user/slinky-slurm-storage.md
```diff
@@ -0,0 +1,115 @@
...
+`helm.sh/resource-policy: keep` when a release is upgraded or removed. Argo CD
+treats that annotation like `Delete=false` during Application deletion; the
+explicit `argocd.argoproj.io/sync-options: Delete=false,Prune=false` annotation
+also prevents sync pruning, whether manual or automated, when shared storage is
+disabled but the Application remains. Argo CD may report such a retained,
+no-longer-desired PVC as `OutOfSync` until it is explicitly deleted or restored
+to the desired manifests.
+
+Re-enabling shared storage with the same namespace and claim names reuses the
+retained claims and their data, provided the existing immutable PVC fields are
```
njhensley, line 91: **🔵 Nitpick — Doc omits redeploy immutability hard-fail when changing class/size on an enabled bundle**

storageClassName is immutable and spec.resources.requests.storage only expands (and only with allowVolumeExpansion). Changing --shared-storage-class, or shrinking size, on a redeploy of an already-enabled bundle hard-fails `helm upgrade` / Argo CD sync on PVC immutable-field validation — the `keep` resource-policy governs deletion, not updates. This line softly alludes to it ("provided the existing immutable PVC fields are compatible") but never states the redeploy failure mode explicitly.

**Fix:** Add one sentence: changing the shared StorageClass or shrinking size after the claims exist requires deleting the PVCs first (with the data-loss caveat already documented).

pkg/bundler/slinky_storage.go
```diff
@@ -0,0 +1,347 @@
...
+}
+
+func parseSlinkySharedVolume(
+	storage map[string]any,
+	key string,
+	volumeName string,
+	mountPath string,
+) (slinkySharedVolume, error) {
+
+	value, exists := storage[key]
```
njhensley, line 235: **🔵 Nitpick — parseSlinkySharedVolume missing / non-object home|data branches untested**

The `storage.<key> is required when shared storage is enabled` and `storage.<key> must be an object` branches (lines 235-244) have no covering test — existing cases only mutate sub-fields, never delete `home`/`data` or set them to a non-object. e.g. --set-json 'slinkyslurm:storage={"enabled":true,"home":"x"}' is unpinned by Go tests (chainsaw doesn't hit it either).

**Fix:** Add two table rows: drop `home` entirely, and set `data` to a scalar, each asserting the INVALID_REQUEST message.

pkg/bundler/bundler.go
```diff
@@ -1656,6 +1666,49 @@ func (b *DefaultBundler) applyNodeSchedulingOverrides(componentName string, valu
...
+	values map[string]any,
+	provider recipe.DataProvider,
+) error {
+
+	if b.Config == nil || b.Config.SharedStorageClass() == "" {
+		return nil
+	}
+	registry, err := recipe.GetComponentRegistryFor(provider)
+	if err != nil {
+		return errors.WrapWithContext(errors.ErrCodeInternal,
```
njhensley, line 1683: **🔵 Nitpick — applySharedStorageClassOverride error / no-path branches untested**

The registry-load-error branch (GetComponentRegistryFor err → ErrCodeInternal) and the `comp == nil || len(paths)==0` early return (when SharedStorageClass is set) are both unexercised — the generic-class test exits at the earlier SharedStorageClass()=="" guard, not the no-paths return. Registry-load-error is near-unreachable with the embedded FS, so this is purely defensive.

**Fix:** Optional: a table row with SharedStorageClass set against a component lacking sharedStorageClassPaths, asserting a clean no-op.

docs/user/slinky-slurm-storage.md
```diff
@@ -0,0 +1,115 @@
+# Slurm Shared Storage
+
+AICR can add persistent shared filesystems mounted at `/home` and `/scratch/fsw`
+to Slinky Slurm login and compute pods. The feature is off by default and
+requires a pre-existing Kubernetes StorageClass that supports `ReadWriteMany`
+(RWX).
```
njhensley, line 6: **🔵 Nitpick — Doc omits multi-tenant shared-/home authorized_keys caveat**

A single shared /home is visible to every login and compute pod, so a user's ~/.ssh/authorized_keys placed on it propagates cluster-wide; isolation reduces to the backend's POSIX uid/gid permissions. Given the feature ships an intentionally SSH-locked login pod by default (empty rootSshAuthorizedKeys + PasswordAuthentication off), the shared-home posture change is worth calling out.

**Blast radius:** Multi-tenant Slurm deployments where operators assume per-pod isolation.

**Fix:** Add one sentence noting that isolation on the shared filesystem is enforced only by backend POSIX permissions and that user-placed authorized_keys on /home apply cluster-wide.

njhensley (approved):
## 🔁 Re-review — all prior feedback addressed ✅

**Method:** 4 independent persona reviewers (Correctness · Domain/Storage-Lifecycle · Test-Coverage · Security/Docs) → adversarial senior meta-reviewer re-deriving each claim from the resolved code at `522ef1b4`. Delta re-review of my prior 9 findings, each verified against the code/tests rather than the "fixed" replies.

**Legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick  |  ✔️ Addressed · ◐ Partially · ✖️ Not addressed

### Prior-feedback status

|---|---------------|-------------|
| 🟠 | Shared PVCs deleted by documented `helm uninstall` → data loss | ✔️ **Addressed** — `keep` + `argocd.argoproj.io/sync-options: Delete=false,Prune=false` on both PVCs; `cli-reference.md` rescoped (StatefulSet-claim vs standalone-PVC); new "Lifecycle and deletion" data-loss section |
| 🟡 | Duplicate home/data claimName not rejected | ✔️ **Addressed** — equal-claimName reject + test |
| 🟡 | `storageClassName` presence-only, trimmed value discarded | ✔️ **Addressed** — `IsDNS1123Subdomain` + trimmed write-back + template `\| quote` |
| 🟡 | No deployed/UAT proof volumes **merge** not **replace** | ◐ **Partially** — closed by cited Slinky v1.2.0 StrategicMergePatch + EKS GB200 UAT; no in-repo render guard (un-guardable, chart not vendored) — see inline N1 |
| 🟡 | Chainsaw `disabled-storage` fail-open negative check | ✔️ **Addressed** — `set -e` before the `! ls` |
| 🟡 | `appendNamedObject` idempotent-skip branch untested | ✔️ **Addressed** — integration `len==2` case + unit idempotent case |
| 🟡 | Defensive error branches untested | ✔️ **Addressed** — table cases for each malformed-value branch |
| 🔵 | Template `index .Values "slinky-slurm"` literal has no binding test | ✔️ **Addressed** — component-binding test |
| 🔵 | No e2e mixed per-volume storage class test | ✔️ **Addressed** — mixed-class materialize→render test |

**8 of 9 fully addressed; #4 partial by nature (un-guardable in-repo).** Package tests pass at head; CodeRabbit's two items (loop `ctx.Err()`, preserve renderer error code) are also both addressed.

### New (all 🔵, non-blocking)

Five nitpicks inline — **N2 and N5 are one-paragraph doc additions worth folding in**; N1 a nice-to-have UAT assertion; N3/N4 low-value defensive test rows.

### Summary

```
Prior:  ✔️ 8 addressed   ◐ 1 partial (un-guardable)   ✖️ 0 not addressed
New:    🔴 0   🟠 0   🟡 0   🔵 5
```

Clean work — approving. None of the new items block; treat them as optional fast-follows.

PR #2081, opened 2026-08-05
feat(uat): slot-aware daytime cluster-name convention (ADR-017)

njhensley:
## Summary

Introduce a slot-aware UAT daytime cluster-name convention — `aicr-uat-day-<slug>-<slot>-<run_id>` — replacing `aicr-uat-day-<reservation>-<run_id>`, so the cross-run teardown/guard discovery key is scoped per `(reservation, slot)` and stays within GKE's 40-char cap with durable headroom (design: ADR-017).

## Motivation / Context

The daytime cluster name is a cross-run discovery key: `daytime-down` and the pre-batch guard find a held cluster by scanning a stable prefix (the exact name isn't reconstructable from a different run). The old `aicr-uat-day-<reservation>-` key was **cloud-scoped** (collides once an account holds >1 reservation), had **no slot axis** for the multi-slot end state, and burned GKE's 40-char budget on the verbose reservation name (~7 chars headroom, eroding as run_ids grow and reservation names lengthen).

The new scheme uses a short registry `slug` (2–4 chars) plus a `slot` index, giving a per-`(reservation, slot)` discovery key that survives to the "multiple reservations per account, each multi-slot" end state, with ≥6 chars GKE headroom permanently at a 13-digit run_id.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Build/CI/tooling
- [x] Documentation update

## Component(s) Affected

- [x] Other: UAT CI workflows (`.github/workflows/uat-*.yaml`), reservation registry + broker (`infra/uat/reservations.yaml`, `pkg/uatbroker`, `tools/uat-broker`)

## Implementation Notes

- **New convention:** nightly `aicr-uat-<run_id>` (unchanged); daytime `aicr-uat-day-<slug>-<slot>-<run_id>` (slot `0` today).
- **Registry `slug`** (new field, validated at load time: non-empty, unique, `^[a-z][a-z0-9]{1,3}$`): `aws-h100→ah1`, `gcp-h100→gh1`, `azure-h100→zh1`, `aws-gb200→ag2`, `kind-h100→kh1`. Emitted by `uat-broker reservations --name` → threaded through `uat-run.yaml` to the cloud lanes, same pattern as `accelerator`.
- **`slot`** is a runtime `uat-run.yaml` input (default `0`), forwarded to aws/gcp/azure (not the kind lane, which provisions no cloud cluster) — forward-compatible with multi-slot without another rename.
- **Guard + teardown** scan the new `(slug, slot)` prefix; fail-closed behavior preserved (list-error → block, bounded retry, `>1`-match → fail loud).
- **Length budget:** GKE (40) is the only binding cap; worst case 34 chars → ≥6 headroom at a 13-digit run_id. AKS (63, incl. `<name>-rg`) and EKS (100) have ample room.

## Testing

```bash
make qualify   # CI-equivalent gate — run in CI on this PR
```

The implementation ran `go test -race ./pkg/uatbroker/... ./tools/uat-broker/...`, `golangci-lint`, the broker `slug=` emit check, and workflow lint before committing. Table-driven tests added for slug validation (valid / empty / duplicate / bad-charset) and the `slug=` output. Note: `make qualify` was not run locally before push — CI is the first full gate.

## Risk Assessment

- [x] **Medium** — Touches the UAT provision/teardown critical path across three cloud workflows plus the broker.

**Rollout notes:** The guard and teardown carry a **dual-prefix migration shim** — they match both the new `aicr-uat-day-<slug>-<slot>-` and the legacy `aicr-uat-day-<reservation>-` prefixes, so any live old-named daytime cluster is still discovered and torn down by the next `daytime-down` (no manual pre-merge teardown needed). Once a cycle has passed with no old-named clusters remaining, drop the legacy leg (marked in-code as a transitional shim).

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — broker/tool packages; full `make qualify` deferred to CI
- [x] Linter passes (`make lint`) — changed Go packages + workflows


.github/workflows/uat-aws.yaml
```diff
@@ -384,7 +419,8 @@ jobs:
             echo "::error::could not list clusters to determine daytime-cluster state for reservation ${RESERVATION}: ${OUT}"
             exit 1
           fi
-          HELD=$(printf '%s\n' "${OUT}" | tr '\t' '\n' | grep "^${PREFIX}[0-9]" || true)
+          # Broad matcher (anything prefix-ish blocks a batch), both prefixes.
+          HELD=$(printf '%s\n' "${OUT}" | tr '\t' '\n' | grep -E "^(${PREFIX}|${LEGACY_PREFIX})[0-9]" || true)
```
lockwobr, line 431: The `SLUG_RE`/`SLOT_RE` gates added in *Validate inputs* exist because these values become `grep -E` prefixes, and the comment there states the failure precisely: an invalid pattern makes `grep` exit non-zero, `|| true` swallows it, and the scan reads as "no match".

`${RESERVATION}` is interpolated into this same alternation via `LEGACY_PREFIX` (line 411) but gets no equivalent check. `Registry.Validate` (`pkg/uatbroker/registry.go:105-111`) constrains the reservation `name` only for non-emptiness and uniqueness; there is no charset rule, unlike `slugPattern` for `slug`.

This change also widens the exposure. The previous scan used plain `grep` (BRE), where `(`, `)`, `|`, `+`, `?` are literal. Under `-E` they are operators.

So a future registry row named e.g. `aws-h100(x)` or `aws-h100+` passes registry validation, then makes `grep -E` exit 2 here. `HELD` ends up empty and the guard prints "no daytime cluster ... proceeding", provisioning the nightly batch into a reservation that is still held. The same shape at line 492 makes `daytime-down` record `DAYTIME_FOUND=false` and skip teardown entirely.

All five committed reservation names are `[a-z0-9-]` only, so nothing breaks today. Two ways to close it: add a charset rule for `name` in `Registry.Validate` next to the slug one, or drop `|| true` and distinguish `grep` rc=1 (no match) from rc>=2 (error).

Same pattern at `uat-gcp.yaml:364` / `uat-azure.yaml:380` and in all three teardown resolvers.

.github/workflows/uat-gcp.yaml
```diff
@@ -323,7 +360,8 @@ jobs:
             echo "::error::could not list clusters to determine daytime-cluster state for reservation ${RESERVATION}: ${OUT}"
             exit 1
           fi
-          HELD=$(printf '%s\n' "${OUT}" | grep "^${PREFIX}[0-9]" || true)
+          # Broad matcher (anything prefix-ish blocks a batch), both prefixes.
+          HELD=$(printf '%s\n' "${OUT}" | grep -E "^(${PREFIX}|${LEGACY_PREFIX})[0-9]" || true)
```
lockwobr, line 399: Same as the note on `uat-aws.yaml:423`: `${RESERVATION}` reaches this `grep -E` alternation through `LEGACY_PREFIX` (line 351) with no charset validation anywhere, while `slug` and `slot` are both gated in *Validate inputs*. With the scan now using ERE rather than BRE, a reservation name containing `(`, `|`, `+` or `?` makes the pattern invalid, `|| true` turns that into an empty `HELD`, and the guard proceeds as if no daytime cluster were held.

.github/workflows/uat-azure.yaml
```diff
@@ -340,7 +376,8 @@ jobs:
             echo "::error::could not list clusters to determine daytime-cluster state for reservation ${RESERVATION}: ${OUT}"
             exit 1
           fi
-          HELD=$(printf '%s\n' "${OUT}" | grep "^${PREFIX}[0-9]" || true)
+          # Broad matcher (anything prefix-ish blocks a batch), both prefixes.
+          HELD=$(printf '%s\n' "${OUT}" | grep -E "^(${PREFIX}|${LEGACY_PREFIX})[0-9]" || true)
```
lockwobr, line 388: Same as the note on `uat-aws.yaml:423`: `${RESERVATION}` reaches this `grep -E` alternation through `LEGACY_PREFIX` (line 368) unvalidated, while `slug` and `slot` are both gated in *Validate inputs*. An ERE metacharacter in a future reservation name makes the pattern invalid, and `|| true` converts that error into a silent "no match".

.github/workflows/uat-aws.yaml
```diff
@@ -444,7 +488,8 @@ jobs:
           # End-anchored: only an exact <prefix><run-id> name is a teardown
           # target. The guard's matcher is deliberately broader — anything
           # prefix-ish should BLOCK a batch, but must not be destroyed here.
-          mapfile -t held < <(printf '%s\n' "${OUT}" | tr '\t' '\n' | grep -E "^${PREFIX}[0-9]+$" || true)
+          # Matches the new (slug, slot) prefix OR the legacy reservation prefix.
+          mapfile -t held < <(printf '%s\n' "${OUT}" | tr '\t' '\n' | grep -E "^(${PREFIX}|${LEGACY_PREFIX})[0-9]+$" || true)
           if [[ "${#held[@]}" -eq 0 ]]; then
             echo "no ${PREFIX}* cluster found; nothing to tear down (already gone or never provisioned)."
```
lockwobr: The scan above now covers `${PREFIX}` and `${LEGACY_PREFIX}`, but this zero-match message names only `${PREFIX}`. During the migration window an operator triaging a suspected leak reads "no aicr-uat-day-ah1-0-* cluster found" and cannot tell from the log whether the legacy `aicr-uat-day-<reservation>-*` prefix was searched too.

Diagnostic only: `DAYTIME_FOUND=false` and the teardown skip are both correct. Same at `uat-gcp.yaml:433` and `uat-azure.yaml:449`.

docs/contributor/uat.md
```diff
@@ -91,10 +91,10 @@ The `lifecycle` input selects one of three cluster lifecycles, all sharing the r
...
... deletion tombstones and stale terraform-state locks make same-name reuse flaky). The `slug` is a short 2–4 char registry field (account-stable and unique across reservations, where the reservation name is only cloud-unique), and `slot` is a runtime input (default `0`, slot-ready for a future multi-cluster-per-day end state); together they keep the daytime name inside GKE's 40-char cap. Because the name is no longer reconstructable from the reservation/slug alone, the evening `daytime-down` and the pre-batch guard find the held cluster by **scanning the prefix** (`list-clusters`/`az aks list`/`clusters list` + match), the same name-free discovery the guard already uses for run-scoped clusters. During the rename the scans carry a second, **legacy** leg matching the old `aicr-uat-day-<reservation>-` prefix so an in-flight old-named daytime cluster is still discovered and torn down; it is a transitional shim removed once none remain. The actuator derives the cluster name, resource group, and terraform-state key all from `.deployment.id`, so recovering the name recovers everything teardown needs. `skip_delete` is a nightly-only debugging escape and is ignored by the daytime lifecycles.
```
lockwobr: "the same name-free discovery the guard already uses for run-scoped clusters" does not hold: there is no scan-based discovery path for run-scoped (nightly) clusters. Every cluster-listing call in the tree matches a daytime prefix only (`uat-aws.yaml:413`/`471`, `uat-gcp.yaml:353`/`410`, `uat-azure.yaml:370`/`426`), and nightly names are derived deterministically from `github.run_id` (`uat-aws.yaml:132`), so nothing ever needs to discover them.

The clause predates this PR, but it sits on a line being rewritten here, so it is in scope for the same edit.

docs/contributor/uat.md
```diff
@@ -155,14 +155,14 @@ Access is **out-of-band by design**: nothing here routes a kubeconfig or endpoin
 # and more than one means a leak the pre-batch guard should have caught.
```
lockwobr, line 155: Line 151, immediately above this block, still reads "first **discover** it by its reservation-scoped prefix". Discovery moved to the `(slug, slot)` key in this PR, and the snippet below already uses `aicr-uat-day-ah1-0-`.

Every other occurrence in this file (lines 10, 14, 94-97, 185) was rewritten to `(slug, slot)`-scoped; this sentence was missed. An operator following the prose rather than the snippet scans `aicr-uat-day-<reservation>-*`, which matches no newly provisioned cluster, and concludes none is held.

docs/contributor/uat.md
```diff
@@ -155,14 +155,14 @@ Access is **out-of-band by design**: nothing here routes a kubeconfig or endpoin
...
 one() { [ "$(printf '%s' "$1" | grep -c .)" -eq 1 ] || { echo "expected exactly one daytime cluster, got: ${1:-<none>}" >&2; return 1; }; }
 
-# AWS — training cluster (prefix aicr-uat-day-aws-h100-)
+# AWS — training cluster (prefix aicr-uat-day-ah1-0-; slug ah1, slot 0)
 name=$(aws eks list-clusters --region us-east-1 --query "clusters[]" --output text \
-  | tr '\t' '\n' | grep '^aicr-uat-day-aws-h100-')
+  | tr '\t' '\n' | grep '^aicr-uat-day-ah1-0-')
```
lockwobr: These operator snippets match the new prefix only, but the guard and teardown in this same PR carry a dual-prefix shim (`uat-aws.yaml:411`/`423`) precisely because an old-named daytime cluster may still be held at cut-over.

Inside that window, `one()` on line 156 turns the zero match into `expected exactly one daytime cluster, got: <none>`, so the operator concludes nothing is held while the workflows would still discover and tear that cluster down. Adding the legacy alternative for the life of the shim would keep the doc and the workflows consistent, e.g. `grep -E '^aicr-uat-day-(ah1-0|aws-h100)-'`.

Self-resolving once a full daytime cycle has run under the new names. Line 97 does mention the legacy leg in prose, just not in the copy-paste block.

tools/uat-broker/main.go
```diff
@@ -182,6 +182,10 @@ func runReservations(args []string, stdout, stderr io.Writer) error {
 			return lookupErr
 		}
 		// GITHUB_OUTPUT-style key=value lines; every value is single-line.
+		// slug is surfaced so uat-run.yaml's resolve step exposes
+		// needs.resolve.outputs.slug (the daytime cluster name's discovery key,
+		// ADR-017), the same way cloud/accelerator are threaded to the pipelines.
+		fmt.Fprintf(&b, "slug=%s\n", res.Slug)
```
lockwobr, line 188: `slug=` is now emitted as the first key, but `tools/uat-broker/README.md:23-31` still documents this command's output as starting at `cloud=`, and the word `slug` appears nowhere in that file.

No runtime impact: both in-repo consumers parse by key, not position (`uat-run.yaml:190` appends the whole block to `$GITHUB_OUTPUT`, `uat-nightly-batch.yaml:194-199` matches `^nightly-intents=`). The cost is that the README is the documented contract for this subcommand, so a future consumer written from it will not know `slug` is available.

Worth noting that ADR-017's consumer-inventory table lists `tools/uat-broker/main.go` but not its README, which is how this was missed.

lockwobr (changes requested):
Cross-review of `e837ed45`, with each finding independently re-checked against the pinned tree before being reported. Details are inline.

Nothing here is a merge blocker: the one finding with an operational failure mode is latent (no current registry row can trigger it), and the rest is documentation drift, most of it inside the migration window this PR opens.

**Open questions**

- Once the legacy `aicr-uat-day-<reservation>-` leg is dropped, the guard's discovery key becomes purely `(slug, slot)`-scoped, so a `nightly` run at the default `slot=0` would no longer be blocked by a leaked daytime cluster in a *different* slot on the same reservation. A nightly batch consumes the whole reservation, not one slot. Unreachable today since `uat-run.yaml` rejects `slot != 0`, but the follow-up that removes the shim should decide the guard's slot scope deliberately rather than inheriting it.
- Is the shim removal tracked anywhere? "Drop the legacy leg once none remain" appears in six places with no issue reference, and there is no TODO-with-issue or CI check, so nothing schedules the cleanup or flags it once the window closes.
- `docs/contributor/uat.md` names ADR-017 in prose at lines 10, 94-97 and 185 but never links to `../design/017-uat-cluster-name-convention.md`, unlike how ADR-016 is referenced from `docs/user/cli-reference.md` and `docs/integrator/recipe-development.md`. Stylistic only.
- GitHub applying the `workflow_dispatch` default when a dispatcher omits `-f slot=` was reasoned from the workflow definitions, not executed. If defaults were not applied, `inputs.slot` would be empty and the per-cloud `SLOT_RE` preflight fails closed, so the failure direction is safe either way.
- The GKE 40-char / AKS 63 / EKS 100 caps are taken from the ADR text and were not confirmed against upstream cloud docs. The ADR's length-budget arithmetic is internally consistent.

lockwobr (approved):
Re-reviewed at `78d5fbb0`. All eight findings from the previous round are addressed, and I've resolved those threads.

Verified against the current tree:

- The unvalidated reservation name in the `grep -E` alternation is closed at the data source with `namePattern` (`^[a-z]([a-z0-9-]*[a-z0-9])?$`) in `Registry.Validate`, rather than per-workflow. That covers all six scan sites by construction, and the lookup path fails closed before any cloud lane runs. The four added charset cases (ERE metacharacter, dot, uppercase, trailing hyphen) lock the invariant.
- The daytime-down no-match log now names both the new and legacy prefixes in all three pipelines.
- `docs/contributor/uat.md`: the "run-scoped clusters" discovery claim is corrected, the discovery prose is `(slug, slot)`-scoped, and the operator snippets carry the legacy alternative for the life of the shim.
- `tools/uat-broker/README.md` documents the `slug=` key.

The ADR-017 prose links raised as an open question were added too.

CI is green on this commit.

PR #2082, opened 2026-08-05
fix(testgrid-publish): reject null and zero-timestamp predicates

### LINKED ISSUE #2053: testgrid-publish: reject 'predicate: null' statements before publication
## Summary

`tools/testgrid-publish` accepts an in-toto statement whose `predicate` is the JSON literal `null` and publishes TestGrid results with a fabricated year-1 timestamp and blank metadata — defeating the loader's own documented fail-closed intent for broken predicates.

## Mechanism

In `tools/testgrid-publish/bundle.go` (`loadPredicate`):

- `predicate: null` decodes into `json.RawMessage("null")`, which is **non-nil**, so the `stmt.Predicate == nil` presence check passes.
- `json.Unmarshal([]byte("null"), &pred)` is a **no-op** for a struct target (no error), leaving a zero-value `attestation.Predicate`.
- A zero predicate carries no profile block, so `ValidatePredicateTypeCoherence` passes for `recipe-evidence/v1`.

The caller in `tools/testgrid-publish/main.go` then proceeds with the zero predicate: `started/finished.Timestamp = pred.AttestedAt.Unix()` (year-1 epoch), blank `aicr_version` / `k8s_version` metadata, and a build-id derived from the zero time — despite its comment stating "a corrupt or unparseable predicate means the bundle is broken; fail closed rather than fabricating an attestation timestamp."

## Reachability

`.github/workflows/testgrid-publish.yml` (manual community/backfill dispatch) states: "bundle_ref is validated for format (ghcr.io/<org>/...@sha256:<64hex>) but not provenance/signature before the tool pulls and parses it", and offers `source_class: community`. An externally supplied digest-pinned bundle with `predicate: null` therefore publishes.

## Proposed fix

Unmarshal into a pointer (`var pred *attestation.Predicate`) and reject nil before coherence validation, plus a null-predicate regression test.

## Origin

Surfaced during PR #2044 review; pre-existing on `main` (#2044 only added the type/profile coherence check after the same decode), so tracked here as a follow-up rather than fixed in that PR.

yuanchen8911:
## Summary

`tools/testgrid-publish` now rejects an in-toto statement whose `predicate` is the JSON literal `null`, and one whose `attestedAt` is zero (e.g. an empty `{}` predicate), instead of publishing TestGrid results with a fabricated year-1 attestation timestamp and blank metadata.

## Motivation / Context

`loadPredicate` decoded the raw predicate into a value struct, so `json.Unmarshal([]byte("null"), &pred)` was a silent no-op: the zero-value predicate passed the presence and coherence checks and reached the publisher, defeating the loader's documented fail-closed intent. An empty predicate object reaches the same fabricated-timestamp state through a different door — it decodes to a non-nil zero value. Both are reachable via the manual `testgrid-publish.yml` dispatch, which accepts an externally supplied digest-pinned bundle without provenance validation, and neither can be produced by a supported producer (`attestation.Build` defaults a zero `attestedAt` to the current time before the statement is built).

Fixes: #2053
Related: #2044 (review round that surfaced the defect)

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tools/testgrid-publish`

## Implementation Notes

Decode into `var pred *attestation.Predicate`: a JSON-literal `null` unmarshals to a nil pointer (whereas a struct target silently stays zero-valued), and a nil guard rejects it with `ErrCodeInvalidRequest` before `ValidatePredicateTypeCoherence`. A second guard rejects `pred.AttestedAt.IsZero()` for the same reason — a zero timestamp means the bundle is broken, and would otherwise flow into the published `started.Timestamp` and the derived build ID. Both rejection cases are covered by a table-driven regression test alongside the existing `loadPredicate` subtests.

## Testing

```bash
go test ./tools/testgrid-publish/...            # ok
go test -race ./tools/testgrid-publish/...      # ok
golangci-lint run -c .golangci.yaml ./tools/testgrid-publish/...  # 0 issues
```

Coverage: `tools/testgrid-publish`: 77.8% → 78.1% (+0.3%). Full `make qualify` will be run before this PR leaves draft.

## Risk Assessment

**Rollout notes:** N/A — tightens input validation in a CI publishing tool; no user-facing surface.

## Checklist

### DIFF
--- a/tools/testgrid-publish/bundle.go
+++ b/tools/testgrid-publish/bundle.go
@@ -119,8 +119,9 @@ func parseCriteria(bundleDir string) (RecipeCriteria, error) {
 }
 
 // loadPredicate reads the in-toto Statement from the bundle and returns
-// the decoded Predicate. Returns an error when the file is absent or
-// malformed — callers should fall back to sensible defaults.
+// the decoded Predicate. Returns ErrCodeNotFound when the file is absent —
+// the only case callers may fall back on — and ErrCodeInvalidRequest for a
+// malformed predicate, which must fail closed.
 func loadPredicate(bundleDir string) (*attestation.Predicate, error) {
 	path := filepath.Join(bundleDir, attestation.StatementFilename)
 	data, err := readBoundedFile(path, defaults.MaxBundlePOSTBytes)
@@ -144,18 +145,31 @@ func loadPredicate(bundleDir string) (*attestation.Predicate, error) {
 		return nil, errors.New(errors.ErrCodeInvalidRequest, "statement has no predicate field")
 	}
 
-	var pred attestation.Predicate
+	// Decode into a pointer: a JSON-literal `null` predicate unmarshals to a
+	// nil pointer (a struct target would silently stay zero-valued and later
+	// fabricate a year-1 attestation timestamp), so reject it here.
+	var pred *attestation.Predicate
 	if err := json.Unmarshal(stmt.Predicate, &pred); err != nil {
 		return nil, errors.Wrap(errors.ErrCodeInvalidRequest,
 			"failed to decode predicate", err)
+	}
+	if pred == nil {
+		return nil, errors.New(errors.ErrCodeInvalidRequest, "statement predicate is null")
+	}
+	// Same fail-closed rule for an empty predicate object: every supported
+	// producer stamps attestedAt (attestation.Build defaults it to now), so
+	// a zero value means the bundle is broken and would otherwise publish a
+	// fabricated year-1 timestamp and build ID.
+	if pred.AttestedAt.IsZero() {
+		return nil, errors.New(errors.ErrCodeInvalidRequest, "statement predicate has no attestedAt")
 	}
 	// The shared bidirectional contract, not just a type allowlist: v1 must
 	// not carry a profile block, v2 must carry a well-formed one, anything
 	// else (including an absent predicateType) is unknown. Without it a v1
 	// statement smuggling a profile block — evidence every other consumer
 	// rejects — would publish to TestGrid.
-	if err := attestation.ValidatePredicateTypeCoherence(stmt.PredicateType, &pred); err != nil {
+	if err := attestation.ValidatePredicateTypeCoherence(stmt.PredicateType, pred); err != nil {
 		return nil, err
 	}
-	return &pred, nil
+	return pred, nil
 }
--- a/tools/testgrid-publish/bundle_test.go
+++ b/tools/testgrid-publish/bundle_test.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"encoding/json"
+	stderrors "errors"
 	"os"
 	"path/filepath"
 	"strings"
@@ -23,6 +24,7 @@ import (
 	"time"
 
 	"github.com/NVIDIA/aicr/pkg/defaults"
+	"github.com/NVIDIA/aicr/pkg/errors"
 	"github.com/NVIDIA/aicr/pkg/evidence/attestation"
 )
 
@@ -261,6 +263,42 @@ func TestLoadPredicate(t *testing.T) {
 			t.Fatal("expected error for invalid JSON")
 		}
 	})
+
+	invalidPredicates := []struct {
+		name            string
+		predicate       any
+		wantErrContains string
+	}{
+		{"null predicate returns error", nil, "statement predicate is null"},
+		{"empty predicate object returns error", map[string]any{}, "statement predicate has no attestedAt"},
+	}
+	for _, tt := range invalidPredicates {
+		t.Run(tt.name, func(t *testing.T) {
+			dir := t.TempDir()
+			stmt := map[string]any{
+				"_type":         "https://in-toto.io/Statement/v0.1",
+				"predicateType": attestation.PredicateTypeV1,
+				"predicate":     tt.predicate,
+			}
+			stmtBytes, _ := json.Marshal(stmt)
+			if err := os.WriteFile(
+				filepath.Join(dir, attestation.StatementFilename),
+				stmtBytes, 0o600,
+			); err != nil {
+				t.Fatal(err)
+			}
+			_, err := loadPredicate(dir)
+			if err == nil {
+				t.Fatal("expected error for invalid predicate")
+			}
+			if !stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, "")) {
+				t.Errorf("error code = %v, want ErrCodeInvalidRequest", err)
+			}
+			if !strings.Contains(err.Error(), tt.wantErrContains) {
+				t.Errorf("error = %v, want message containing %q", err, tt.wantErrContains)
+			}
+		})
+	}
 
 	t.Run("statement with no predicate field returns error", func(t *testing.T) {
 		dir := t.TempDir()


tools/testgrid-publish/bundle_test.go
```diff
@@ -262,6 +264,29 @@ func TestLoadPredicate(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		_, err := loadPredicate(dir)
+		if err == nil {
+			t.Fatal("expected error when predicate is JSON null")
+		}
+		if !stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, "")) {
+			t.Errorf("error code = %v, want ErrCodeInvalidRequest", err)
+		}
+	})
```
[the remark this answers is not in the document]
yuanchen8911: The two invalid-predicate cases this PR introduces are now a single table-driven loop (`invalidPredicates` in `TestLoadPredicate`) as of e98b81dd1, retained at the current head c6e3afde1.

The three pre-existing subtests (`missing statement`, `invalid json`, `no predicate field`) are left in their established sequential form: they pre-date this PR, their setups are heterogeneous (no file vs. raw non-JSON bytes vs. a structured statement without the key), and restructuring them is unrelated refactoring outside this fix's scope.

tools/testgrid-publish/bundle.go
```diff
@@ -144,18 +144,31 @@ func loadPredicate(bundleDir string) (*attestation.Predicate, error) {
...
 	// not carry a profile block, v2 must carry a well-formed one, anything
 	// else (including an absent predicateType) is unknown. Without it a v1
 	// statement smuggling a profile block — evidence every other consumer
 	// rejects — would publish to TestGrid.
-	if err := attestation.ValidatePredicateTypeCoherence(stmt.PredicateType, &pred); err != nil {
+	if err := attestation.ValidatePredicateTypeCoherence(stmt.PredicateType, pred); err != nil {
 		return nil, err
 	}
-	return &pred, nil
+	return pred, nil
```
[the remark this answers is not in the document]
yuanchen8911: Declining this one for the current PR. The layering rule this cites governs `pkg/cli` and `pkg/server` — user-interaction layers that must share functional packages with other consumers. `tools/testgrid-publish` is a standalone CI publishing tool with no second consumer of these guards, and the shared, reusable part of predicate validation (`ValidatePredicateTypeCoherence`) already lives in `pkg/evidence/attestation`. What this PR adds is a few lines of input rejection specific to this tool's fail-closed publishing policy (#2053).

Extracting a shared API without a consumer that needs it would broaden the fix beyond its issue. If a second consumer of bundle-statement loading appears, extraction is the right move at that point.

tools/testgrid-publish/bundle.go
```diff
@@ -144,18 +145,31 @@ func loadPredicate(bundleDir string) (*attestation.Predicate, error) {
...
 			"failed to decode predicate", err)
 	}
+	if pred == nil {
+		return nil, errors.New(errors.ErrCodeInvalidRequest, "statement predicate is null")
+	}
+	// Same fail-closed rule for an empty predicate object: every supported
+	// producer stamps attestedAt (attestation.Build defaults it to now), so
+	// a zero value means the bundle is broken and would otherwise publish a
+	// fabricated year-1 timestamp and build ID.
+	if pred.AttestedAt.IsZero() {
```
njhensley, line 163: 🔵 **Nitpick / scope note (not a blocker).** This guard correctly rejects a *zero* `attestedAt`, but a bundle carrying an arbitrary *non-zero* timestamp still publishes it verbatim into `started.json`/`finished.json` and the `buildID` (`{unix}-{digest8}`), which drives TestGrid column ordering. That's acceptable under the documented `workflow_dispatch` + digest-pinned trust model (the digest still partitions distinct content, so a spoofed timestamp can misorder a column but not overwrite unrelated evidence). Flagging only so the fail-closed boundary is explicit — no change requested.

tools/testgrid-publish/bundle_test.go
```diff
@@ -262,6 +264,38 @@ func TestLoadPredicate(t *testing.T) {
...
+				filepath.Join(dir, attestation.StatementFilename),
+				stmtBytes, 0o600,
+			); err != nil {
+				t.Fatal(err)
+			}
+			_, err := loadPredicate(dir)
+			if err == nil {
+				t.Fatal("expected error for invalid predicate")
+			}
+			if !stderrors.Is(err, errors.New(errors.ErrCodeInvalidRequest, "")) {
```
njhensley, line 294: 🔵 **Nitpick (optional tightening).** All three fail-closed guards return `ErrCodeInvalidRequest`, so asserting only the code can't distinguish which branch each input took — the `null` case and the empty-object case could silently swap branches under a future refactor and this test would still pass. Consider a `wantErrContains` column asserting the distinct messages (`"statement predicate is null"` vs `"statement predicate has no attestedAt"`) to pin each input to its guard. Non-blocking — the current cases already exercise both new branches correctly.

njhensley (dismissed):

**Method:** 3 independent persona reviewers (Correctness · Supply-chain/Security · Test-coverage & Architecture), each adjudicated against the resolved code; the crux was independently reproduced.

### Overall assessment — ✅ Approve

Clean, correctly-scoped fail-closed fix. Decoding the predicate into `var pred *attestation.Predicate` is the right mechanism: a JSON-literal `null` keeps its literal bytes in the `json.RawMessage` field (so the earlier `stmt.Predicate == nil` check does *not* catch it) and only unmarshals to a nil pointer — which the new `pred == nil` guard rejects. An empty `{}` object is caught by the `AttestedAt.IsZero()` guard. Both previously flowed through as a zero-value struct and published a fabricated year-1 timestamp + blank metadata to TestGrid.

Verified sound:
- **No false-rejection of real bundles** — `attestation.Build` defaults a zero `AttestedAt` to `now()` before `BuildPredicate`, so every legitimate producer stamps a non-zero timestamp; `IsZero()` is true only for the year-1 fabrication being fixed.
- **Fail-closed routing** — `main.go` falls back to `time.Now()` *only* on `ErrCodeNotFound` (missing/unsigned bundle); `ErrCodeInvalidRequest` propagates and exits non-zero, failing the publish.
- **Regression test** exercises *both* new branches distinctly (`null` → nil-pointer path; `{}` → zero-`attestedAt` path) and asserts `ErrCodeInvalidRequest` idiomatically via `stderrors.Is`.
- `go test ./tools/testgrid-publish/...` passes.

All findings are 🔵 nitpicks — none block merge. Folded below; also left inline.

### Findings (all 🔵, non-blocking)

- 🔵 **attestedAt-only guard** (`bundle.go:163`) — an arbitrary *non-zero* timestamp still publishes and drives the build-id/column ordering. Accepted under the digest-pinned `workflow_dispatch` trust model; noted only to make the fail-closed boundary explicit.
- 🔵 **Test asserts shared error code, not distinct branch** (`bundle_test.go:293`) — consider a `wantErrContains` column to pin each input to its specific guard message. Optional.
- 🔵 **CodeRabbit's `pkg/*` relocation suggestion is not warranted** (`bundle.go:125`) — the real logic (`ValidatePredicateTypeCoherence`, `Predicate`, `PredicateTypeV1/V2`) already lives in `pkg/evidence/attestation`; what remains is appropriate tool-local glue.

### Confirmed non-issues (examined, not defects)

- `json null` → nil-pointer is genuinely the operative guard (verified empirically).
- Absent-predicate `time.Now()` fallback is by design (distinguishes "broken evidence" → fail-closed from "no evidence asserted" → local/unsigned fallback).
- Blank `AICRVersion`/signer fields are honest emptiness (surface as `(missing)`), not the plausible-but-false fabrication this PR targets.

**Duplicate-work note:** `coderabbitai[bot]` posted 2 suggestions (fully table-drive the test; relocate to `pkg/*`) — both non-blocking; addressed above. The other listed reviewer is the PR author (self).

| 0 | 0 | 0 | 3 | **Approve** |

tools/testgrid-publish/bundle.go
yuanchen8911: Agreed, and thanks for making the boundary explicit. A forged non-zero `attestedAt` is inside the documented trust model for `workflow_dispatch` digest-pinned bundles: the digest partitions content, so a spoofed timestamp can misorder a TestGrid column but cannot overwrite or impersonate other evidence. Guarding it further would require provenance/signature verification at ingest, which is deliberately out of #2053's scope. Keeping this PR at the two fail-closed guards (null and zero timestamp); no change made.

tools/testgrid-publish/bundle_test.go
yuanchen8911: Good catch — applied in c820e176e. The table now carries a `wantErrContains` column ("statement predicate is null" vs "statement predicate has no attestedAt") asserted via `strings.Contains` alongside the code check, so each input is pinned to its guard and a branch swap under a future refactor fails the test.

One housekeeping note: the branch was rebased onto the same main tip your branch update brought in, replacing the merge commit with the repo's rebase/squash convention — content is identical plus this test change.

njhensley (approved):

PR #2087, opened 2026-08-06
ci(supply-chain): build aiperf-bench without the BuildKit layer cache

### LINKED ISSUE #2086: aiperf-bench dependency resolution is layer-cached, so the image stops picking up CVE fixes
## Problem

The `aiperf-bench` image intentionally leaves its Python dependencies unpinned (`requirements.txt` contains only `aiperf==0.11.0`, and 44 of aiperf's 45 core deps are ranged) so that rebuilds pick up patched transitive versions.

That does not happen in practice, because the resolution is layer-cached.

`RUN pip install --no-cache-dir -r /tmp/requirements.txt` is keyed on the base image plus the preceding `COPY requirements.txt`. Neither changes between releases, and every workflow that builds the image uses `cache-from: type=gha` with `cache-to: type=gha,mode=max`:

- `.github/workflows/on-tag.yaml` (release)
- `.github/workflows/on-push.yaml` (`:edge`)
- `.github/workflows/vuln-scan-images.yaml` (weekly scan)

BuildKit cannot see that `pip` reached the network, so on a cache hit it replays the old layer verbatim.

Verified empirically with a minimal Dockerfile whose `RUN` records a timestamp: two builds produced `---> Using cache` and an identical timestamp, confirming the layer is replayed rather than re-executed.

## Impact

1. **The image stops picking up CVE fixes.** A cache hit ships a months-old dependency closure, which is the opposite of what leaving the deps unpinned is for.
2. **The weekly vulnerability scan is affected too**, since it builds the same way. It can scan a stale closure and report on packages a freshly built image would not contain.
3. **It is nondeterministic.** GitHub evicts unused cache entries after ~7 days and caps a repo at 10 GB, and a base-image or Dockerfile change busts the key. So a given release may or may not have resolved fresh, and there is no signal either way.

## Why it is filed rather than fixed in #2073

#2073 publishes the third-party source for this image and resolves the closure from the same `requirements.txt` the Dockerfile installs. While the build is cached, a cached image can carry an older closure than a freshly resolved archive, so the two can disagree.

That is a pre-existing condition rather than something #2073 introduces, and fixing it changes build behaviour for three workflows and adds several minutes per architecture per build. #2073 documents the gap instead of claiming it away.

## Possible fix

Exclude `aiperf-bench` from the layer cache so the resolution runs every build. The Go validators have no resolution step and should keep their cache:

```yaml
cache-from: ${{ matrix.phase == 'aiperf-bench' && '' || 'type=gha' }}
cache-to: ${{ matrix.phase == 'aiperf-bench' && '' || 'type=gha,mode=max' }}
```

Alternatively, a cache-busting `ARG` on that layer alone.

**Cost:** a real resolve and install on every build, several minutes per architecture, currently hidden by the cache. On arm64 that includes any source builds.

## Alternative considered

Pinning the closure with a hash-locked requirements file would make the image reproducible and the correspondence exact, but it stops CVE fixes arriving automatically; they would come via Renovate PRs instead, as they already do for Go modules, cosign, and syft. That trade-off was explicitly rejected in favour of keeping the deps floating.

lockwobr:
## Summary

Stop importing the BuildKit layer cache on the `aiperf-bench` matrix leg, so its
Python dependency resolution actually re-runs on every build instead of being
replayed from a months-old cached layer.

## Motivation / Context

`aiperf-bench` leaves its Python closure unpinned on purpose (`requirements.txt`
declares only `aiperf==0.11.0`; 44 of aiperf's 45 core deps are ranged) so that
rebuilds absorb patched transitive versions. That never happened, because the
resolution was layer-cached.

`RUN pip install -r /tmp/requirements.txt` is keyed on the base image plus the
preceding `COPY requirements.txt`. Neither changes between releases, and all
three workflows that build the image passed `cache-from: type=gha`. BuildKit
cannot see that `pip` reached the network, so on a cache hit it replays the old
layer verbatim. The image stopped picking up CVE fixes, and the weekly
vulnerability scan inherited the defect: it could report on a closure a freshly
built image would not contain.

Fixes: #2086
Related: #2073

## Type of Change

- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: CI workflows (`on-tag`, `on-push`, `vuln-scan-images`), `aiperf-bench` image, `docs/integrator`

## Implementation Notes

**The condition is negated deliberately.** The obvious spelling is a silent
no-op:

```yaml
# WRONG - every phase still gets type=gha
cache-from: ${{ matrix.phase == 'aiperf-bench' && '' || 'type=gha' }}
```

GitHub's `a && b || c` returns `c` whenever `b` is falsy, and the empty string
is falsy. So the true branch collapses and the cache value is handed back to
every phase, including `aiperf-bench`. The shipped form puts the non-empty value
in the true branch:

```yaml
cache-from: ${{ matrix.phase != 'aiperf-bench' && 'type=gha' || '' }}
```

`TestAIPerfBenchBuildsWithoutLayerCache` detects that specific mis-spelling
separately from an outright missing gate, because the failure is invisible: the
workflow reads correctly and the build emits no warning.

**Why a test rather than review vigilance.** Restoring `cache-from` here
produces no build-time signal at all; the regression only surfaces later as a
stale CVE report. The gate discovers its targets by scanning every workflow for
`docker/build-push-action` steps that build `aiperf-bench.Dockerfile`, so a new
workflow is covered the day it lands, and it fails closed if the three known
workflows stop matching (a discovery-based test that finds nothing would
otherwise pass vacuously). Plain `docker build` steps (UAT workflows,
`.github/actions/e2e`, the Makefile) import no cache and are correctly excluded.
This follows the existing pattern in `tests/releasepolicy`, which already
asserts over workflow YAML across multiple files.

**Cost.** A real resolve on every build. Recent `vuln-scan-images` runs show a
cache miss on this leg already costs 3m19s (amd64) and 4m14s (arm64) against a
15 minute job timeout, versus ~25s on a hit. Dropping `cache-to` also stops the
image's full venv from consuming the repo's shared 10 GB GHA cache quota.

**Alternative considered.** A cache-busting `ARG` on the `pip` layer alone would
keep the `apt-get`/venv layers cached (~40s), but it adds a build-arg to every
matrix leg and leaves the image consuming cache quota. Not worth the extra
moving part at this margin.

**Docs.** #2073 landed a bullet in `docs/integrator/supply-chain-verification.md`
tracking this exact gap as a caveat on the source-archive correspondence. Since
this PR closes it, that bullet is rewritten to state the guarantee instead.

**Incidental change.** `tools/registry-inventory/registry-allowlist.yaml` gains the
standard 13-line Apache header. `make lint` added it while verifying this branch (the
file is not in `LICENSE_IGNORES`). It is unrelated to #2086 and no gate forces it, so
it is called out here rather than split into its own commit.

## Testing

```bash
make lint          # 0 issues
make test          # no new failures
golangci-lint run -c .golangci.yaml ./tests/releasepolicy/...   # 0 issues
go test -race -run TestAIPerfBenchBuildsWithoutLayerCache ./tests/releasepolicy/
go test -race ./tests/notices/   # #2073's suite, unaffected by the doc edit
```

The new gate was mutation-tested against all three regressions it must catch:

| Mutation | Result |
|---|---|
| `cache-from: type=gha` restored | FAIL, "reuses the BuildKit layer cache on the aiperf-bench leg" |
| falsy-ternary spelling (`== 'aiperf-bench' && '' \|\| ...`) | FAIL, "looks like a gate but is a no-op" |
| `aiperf-bench` matrix leg removed from a workflow | FAIL, vacuous-pass guard fires |

No coverage gate applies: the only Go change is a new test file plus a package
doc comment.

Pre-existing local failures, identical on clean `main` (14 on both), are shell-script
tests that shell out to GNU `timeout`, which macOS does not ship. `make qualify`'s
e2e and scan legs were not run locally for the same environment reason; neither is
implicated by a change to CI YAML and a test file.

## Risk Assessment

**Rollout notes:** The `aiperf-bench` build gets slower by the amount shown
above, well inside the existing 15 minute timeout. Every other image is
untouched and keeps its cache. Reverting is a two-line change per workflow.

## Checklist


docs/integrator/supply-chain-verification.md
```diff
@@ -228,14 +228,15 @@ Only `aiperf-bench` carries this referrer. The other six images need none: their
...
 
 * `pip` and `wheel` are provided by `python -m venv` rather than declared in
   `requirements.txt`, so they ship in the image without source in this archive.
-* The image build layer-caches its dependency resolution, so a cached build can
-  carry an older closure than a freshly resolved archive (tracked in
-  [#2086](https://github.com/NVIDIA/aicr/issues/2086)).
+
+The image build no longer imports a BuildKit layer cache, so its dependency
+resolution re-runs on every build and cannot carry an older closure than the
+archive resolved alongside it ([#2086](https://github.com/NVIDIA/aicr/issues/2086)).
```
njhensley: 🟡 **Minor — the rewrite replaces a tracked gap with a guarantee that isn't quite true.** "...its dependency resolution re-runs on every build and **cannot carry an older closure than the archive resolved alongside it**." The image closure is baked by `RUN pip install` in the early `build-docker` matrix; the third-party source **archive** (#2073) is resolved *independently* later in the `attach-source` job (`tools/generate-aiperf-source` → its own `pip install --dry-run --report`, `needs: [detect, resolve-candidates, publish]`). Two independent, floating resolutions at different times: with ranged deps, a transitive release landing in the build→attach window makes the *archive* resolve **newer**, so the image genuinely *can* carry an older closure than the archive — the exact case "cannot" rules out.

Removing the cache fixed the real #2086 defect (unbounded, months-old staleness) and shrank the residual to a few-minute window — worth stating as a *narrowed, still-tracked* gap rather than an absolute guarantee. Suggest: keep it in the "gaps tracked rather than claimed away" spirit — the image now resolves fresh at build time, but the archive is resolved separately afterward, so the two closures can still differ by that window; correspondence is by shared input (`requirements.txt`), not by identical resolved closure. (An exact guarantee would require resolving the archive from the built image's own `pip freeze`/a shared lockfile.)

tests/releasepolicy/image_cache_policy_test.go
```diff
@@ -0,0 +1,148 @@
...
+		if uses, _ := step["uses"].(string); !strings.Contains(uses, buildPushAction) {
+			continue
+		}
+		with, ok := step["with"].(map[string]any)
+		if !ok {
+			continue
+		}
+		found = true
+		for _, key := range []string{"cache-from", "cache-to"} {
+			value, _ := with[key].(string)
```
njhensley: 🟡 **Minor — the gate fails *open* if a re-enabled cache is written as a YAML list.** `value, _ := with[key].(string)` — if a future edit writes `cache-from:` as a sequence (`cache-from:\n  - type=gha`), yaml.v3 unmarshals it to `[]interface{}`, the assertion drops to `""`, and it lands in the `case value == ""` OK branch: a silent pass that would let the #2086 regression ship green (reproduced empirically). A block scalar `|` stays a string and is correctly caught — only a genuine list escapes.

Trigger is non-idiomatic (GHA `with:` inputs are strings), so 🟡 not 🟠, but this is precisely the repo's documented "negative check that passes on an ambiguous condition → fail closed" anti-pattern. Fix: when the key is present but not a string, `t.Errorf` instead of treating it as empty — e.g. `if raw, present := with[key]; present { if s, ok := raw.(string); ok { value = strings.TrimSpace(s) } else { t.Errorf("%s job %s: %s is %T, not a scalar the gate can inspect", workflow, jobName, key, raw) } }` — or normalize a list to a joined string before the regex checks.

tools/registry-inventory/registry-allowlist.yaml
```diff
@@ -1,3 +1,17 @@
+# Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES.  All rights reserved.
```
njhensley, line 1: 🔵 **Nitpick — unrelated scope, bundled.** The 13-line Apache header is legitimate hygiene (the file isn't in `LICENSE_IGNORES`, so `make license` would generate exactly this), but it's unrelated to the #2086 cache fix and no gate forces it here — CI's `make license` runs `addlicense` in *add* mode with no following `git diff --exit-code`, so a missing header wouldn't fail. Harmless, but per the repo's "separate commits for distinct changes" norm it'd be cleaner as its own header-hygiene commit (or just call it out as incidental in the PR description). No change required.

njhensley (dismissed):
## Multi-persona review — PR #2087

**Method.** Four independent persona reviewers (correctness, CI-DX/operability, supply-chain/security, test-coverage & docs) fanned out over the resolved code, then a senior meta-reviewer adversarially re-derived every finding. Personas post nothing; only surviving findings appear here.

**Legend.** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick

**Assessment.** Solid, well-scoped supply-chain fix. The mechanism is correct — `${{ matrix.phase != 'aiperf-bench' && 'type=gha' || '' }}` yields `type=gha` for the Go legs and an empty (no-cache) value for aiperf-bench, and the deliberate negation avoids the `a && b || c` falsy-collapse footgun. The new `TestAIPerfBenchBuildsWithoutLayerCache` gate is genuinely load-bearing: it catches cache-restored, the falsy spelling, and a dropped/renamed leg (fail-closed), and discovers targets by Dockerfile path so a new workflow is covered on landing. `--no-cache-dir` + a fresh venv means no other closure-freeze mechanism survives, and `provenance: false` makes the `cache-to` removal attestation-neutral.

Two 🟡 and two 🔵 remain — none blocking. The most substantive is a docs sentence that upgrades a still-real (if now much smaller) gap into an absolute guarantee.

| Tier | Count |
|---|---|
| 🔴 Blocker | 0 |
| 🟠 Major | 0 |
| 🟡 Minor | 2 |
| 🔵 Nitpick | 2 |

### Confirmed non-issues (examined, refuted)
- Empty-string `cache-from`/`cache-to` breaking the build — build-push-action treats empty as no-cache.
- `cache-to` removal affecting other legs' caches or provenance — each leg builds independently; `provenance: false` already set.
- Another freeze mechanism (`--mount=type=cache`, wheel cache, base deps) — `--no-cache-dir` + fresh venv, none present.
- UAT workflows missed — they use plain `docker build` (no cache import), correctly out of scope.
- Broken doc anchors / GHA template injection — none; `matrix.phase` is a static matrix value in a `with:` input.

### Out-of-diff nitpick
🔵 **`tools/generate-aiperf-source` `read_closure()` docstring (line 157)** — still lists *"the image build layer-caches its own resolution (#2086), so a cached build can ship an older closure"* as a live gap. This PR removes that cache and updates the two parallel prose spots (the `.md` caveat and the new Dockerfile comment), but leaves the archive resolver's own docstring citing #2086 as open. Worth updating in the same PR so the three #2086 references stay consistent; the correct residual gap to describe here is the independent-resolution window (see the docs comment above), not the removed cache.

docs/integrator/supply-chain-verification.md
lockwobr: Confirmed and fixed in 2bd16dfb — you are right, and the direction matters: because `attach-source` needs `[detect, resolve-candidates, publish]`, the archive resolves *after* `build-docker`, so with ranged deps the archive is the one that can come out newer. That is exactly the case my "cannot carry an older closure" ruled out. I checked `tools/generate-aiperf-source` and it does run its own `pip install --dry-run --report` against `requirements.txt`, so these really are two independent floating resolutions.

Restored as a narrowed, still-tracked gap rather than a guarantee: the bullet now states that the two resolve independently and can differ by the build-to-attach window, that correspondence is by shared input rather than identical resolved closure, and that an exact guarantee would need the archive resolved from the image's own `pip freeze` or a shared lockfile. A separate sentence records what #2086 actually changed: the divergence was unbounded (a cached build could replay a months-old closure) and is now bounded by that window.

tests/releasepolicy/image_cache_policy_test.go
lockwobr: Good catch, fixed in 2bd16dfb. Reproduced it before fixing: rewriting `cache-from` as a sequence made the gate report `ok`, because `[]interface{}` failed the type assertion, fell through to `""`, and landed in the no-cache-is-fine branch.

Fixed by splitting the two cases that `""` was conflating:

- **absent** → `continue`. No cache input at all means nothing is imported, so the invariant holds by construction.
- **present but not a string** → `t.Errorf`. Fails closed with the observed type in the message.

Same mutation now yields: `cache-from is []interface {}, not a scalar this gate can read; rewrite it as a string or teach the gate that shape before relying on it (#2086)`.

You are right that this was the repo's own documented anti-pattern. It is a negative check, so an ambiguous condition had to fail closed, not open.

tools/registry-inventory/registry-allowlist.yaml
lockwobr: Fair, and your read of the mechanism is right: `make license` runs `addlicense` in add mode with no `git diff --exit-code`, so nothing forces the header here.

It rode along because `make lint` added it while I was verifying this branch, and I flagged it before opening the PR rather than discovering it later. Taking your second option and calling it out as incidental in the PR description instead of splitting it, since it is header-only and the branch already has review on it. Happy to pull it into its own commit if you would rather keep the split strict.

njhensley (approved):
## Re-review — delta since my (dismissed) review

My prior `APPROVE` was dismissed by the force-push. New head is the fix commit + a merge of `main`; the PR's own scope is unchanged (same 8 files). Thanks for the detailed replies on all three inline comments.

### Prior-feedback status

|---|---|---|
| 🟡 F1 | Docs over-claim: image "cannot carry an older closure than the archive resolved alongside it" | **✔️ Addressed** — rewritten as a narrowed, still-tracked gap: two independent resolutions (image in `build-docker`, archive later in `attach-source`), correspondence "by shared input, not identical resolved closure," the `pip freeze`/shared-lockfile alternative, and #2086 reframed as bounding a previously-unbounded divergence. |
| 🟡 F3 | Gate test fails *open* if the cache value is a YAML list | **✔️ Addressed + strengthened** — absent (`continue`) vs present-but-non-string (`t.Errorf`, fails closed with the observed `%T`) is exactly the fix; and the new `aiperfBenchCacheValues` per-key check now also catches a degraded `cache-to` missing `mode=max`. Test passes. |
| 🔵 F4 | Unrelated license header bundled | **✔️ Resolved (by agreement)** — PR body now flags it under **Incidental change.** as unrelated to #2086 with no gate forcing it. |
| 🔵 F2 | Stale #2086 language in `generate-aiperf-source` docstring | **✖️ Not addressed (optional)** — `read_closure()`'s docstring still lists *"the image build layer-caches its own resolution (#2086), so a cached build can ship an older closure"* as a live gap. It's now doubly stale: the cache is removed, and its two-gap list no longer matches the `.md`'s (pip/wheel + independent-resolution window). Out-of-diff nitpick — a one-line touch-up would keep the three #2086 references consistent, but it doesn't block. |

### New findings on the fix-commit surface
None. Re-checked the net-new `aiperfBenchCacheValues` logic adversarially — regexes are correct (trailing `scope=` allowed, missing `mode=max` caught) and absent/empty inputs stay legal on purpose. The value-pin couples the test to the `type=gha` backend, but that's a documented, intended policy choice.

**Both 🟡 minors resolved (F3 beyond the ask), F4 handled, only the optional out-of-diff F2 nit remains. Approving.**

Issue #2056, opened 2026-08-05
Convert literal test timeouts to named constants (post-#2044 sweep)


yuanchen8911:
## Summary

PR #2044 (review round) established a pattern for test subprocess deadlines: a file-local named constant with a one-line comment instead of a bare literal, per the AGENTS.md Constants Rule. Two files were converted in that PR:

- `pkg/bundler/deployer/argocdhelm/argocdhelm_test.go` — `helmTemplateTimeout = 30 * time.Second` (12 sites)
- `validators/conformance/secure_access_check_test.go` — `prologueBashTimeout = 10 * time.Second` (1 site)

The rest of the repo still carries ~26 literal `context.WithTimeout(..., N*time.Second/Minute)` sites in test files. This issue tracks the repo-wide sweep, deliberately deferred from #2044 to keep that PR scoped.

## Affected files

```
3  pkg/k8s/pod/wait_termination_test.go
3  pkg/chainsaw/inprocess_test.go
3  pkg/bundler/bundler_dra_annotation_parity_test.go
2  validators/performance/inference_perf_test.go
2  tests/releasepolicy/release_scripts_test.go
2  pkg/trust/trust_test.go
2  pkg/recipe/builder_test.go
1  validators/performance/nccl_trainjob_retry_test.go
1  tests/releasepolicy/release_workflow_test.go
1  pkg/oci/helm_checkpoint_c_test.go
1  pkg/k8s/agent/job_watch_test.go
1  pkg/evidence/verifier/crosscheck_test.go
1  pkg/collector/k8s/server_test.go
1  pkg/collector/k8s/discovery_test.go
1  pkg/collector/gpu/gpu_test.go
1  pkg/bundler/deployer/localformat/vendor_write_test.go
1  pkg/bundler/checksum/checksum_test.go
```

(Found with `grep -rn 'WithTimeout([^,]*, [0-9][0-9]*\*time\.' --include='*_test.go' pkg/ validators/ tools/ tests/ | grep -v 'defaults\.'`.)

## Guidance

- Prefer **file-local constants** named for what they bound (e.g. `helmTemplateTimeout`, `scriptRunTimeout`), not a shared test-timeout constant in `pkg/defaults` — that package is production configuration.
- Group only **semantically identical** deadlines under one constant. Elapsed-time assertions (wall-clock upper bounds proving short-circuit or bounded-backoff behavior) are a different kind of value from subprocess/context deadlines and should not be forced under the same name.
[...]

PR #2088, opened 2026-08-06
fix(tests): derive releasepolicy script deadlines from test timeout

### LINKED ISSUE #1974: tests/releasepolicy: 10s per-script deadline flakes under full-suite load on dev machines

yuanchen8911:
## Summary

Replace the hardcoded 10-second per-script deadlines in `tests/releasepolicy` with a budget derived from the test binary's own deadline (`t.Deadline()`), capped at a hang-detection-sized 90 seconds.

## Motivation / Context

The fixed 10s wall-clock deadline measures the machine and its concurrent load, not the script's correctness. Under full-suite `make test` parallelism (`-race`, `-covermode=atomic`), individual script runs starve past that margin on developer machines and fail with `script exceeded test deadline` — intermittently breaking the documented pre-push `make qualify` gate. The deadline exists to catch a *hung* script, and hangs sit orders of magnitude above slow-but-working runs, so a tight constant buys no detection value.

Fixes: #1974
Related: #2056

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tests/releasepolicy` (release script shell tests)

## Implementation Notes

- `scriptBudget(t)` grants `max(remaining − 30s margin, remaining/2)`, capped at a 90s hang-detection ceiling (`scriptHangCap`). The margin (`deadlineReportMargin`) is reserved so a killed script still fails via `t.Fatalf` with its captured output instead of the test runner's deadline panic — but never more than half the remaining time is taken as margin, so a short targeted `-timeout` (e.g. 25s) still yields a usable budget instead of collapsing to a false failure.
- The budget never exceeds the real remaining time, and once the deadline has passed it goes non-positive: the context expires immediately, failing fast with the script-timeout message instead of starting doomed work.
- The calculation is a pure helper, `scriptBudgetFor(remaining)`, with `scriptBudget(t)` keeping only the `t.Deadline()` handling — pinned by a table-driven boundary test (`TestScriptBudgetFor`: hang cap, cap boundary, full margin, half-margin cap, short `-timeout`, nearly exhausted, expired).
- All three subprocess helpers in the package share the derived budget: `runScript` (~31 invocations), `shellReleaseAssetNames`, and `runOpenVEXGuard`. No fixed-deadline subprocess sites remain in the package.
- The budget inherits whatever `-timeout` the environment (gate, CI, dev machine) was provisioned with — no per-machine constant to re-tune. Timeouts use named constants per the #2056 convention.

## Testing

```bash
AICR_CRITERIA_STRICT=1 go test -short -count=1 -race ./tests/releasepolicy/   # ok, ~170s
golangci-lint run -c .golangci.yaml ./tests/releasepolicy/...                 # 0 issues
# short targeted -timeout must not collapse the budget (review-round repro):
go test -count=1 -run '^TestReleaseStablePromotionConvergesAndRerunsWithoutMutation$' \
  -timeout=25s ./tests/releasepolicy                                          # ok, ~12s
make qualify
```

Coverage gate: test-only change in a test-only package (`_test.go`); no exported production code affected, no coverage delta.

## Risk Assessment

**Rollout notes:** N/A — test-infrastructure only; no user-facing behavior.

## Checklist

### DIFF
--- a/tests/releasepolicy/release_scripts_test.go
+++ b/tests/releasepolicy/release_scripts_test.go
@@ -165,7 +165,8 @@ func TestReleaseSbomSignaturesAreAllowlisted(t *testing.T) {
 // allowlist rather than a copy of it.
 func shellReleaseAssetNames(t *testing.T, tag string) []string {
 	t.Helper()
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	budget := scriptBudget(t)
+	ctx, cancel := context.WithTimeout(context.Background(), budget)
 	defer cancel()
 	text := string(readFile(t, ".github/scripts/release-images.sh"))
 	const marker = "expected_release_asset_names() {"
@@ -183,7 +184,7 @@ func shellReleaseAssetNames(t *testing.T, tag string) []string {
 	command.Env = append(os.Environ(), "RELEASE_TAG="+tag)
 	output, err := command.CombinedOutput()
 	if ctx.Err() != nil {
-		t.Fatalf("evaluate expected_release_asset_names exceeded test deadline: %v\n%s", ctx.Err(), output)
+		t.Fatalf("evaluate expected_release_asset_names exceeded %v budget (derived from -timeout): %v\n%s", budget, ctx.Err(), output)
 	}
 	if err != nil {
 		t.Fatalf("evaluate expected_release_asset_names: %v\n%s", err, output)
@@ -230,8 +231,9 @@ func TestReleaseSignSbomBehavior(t *testing.T) {
 				t.Fatalf("create fake bin: %v", err)
 			}
 			writeExecutable(t, filepath.Join(bin, "cosign"), fakeCosign)
-			// The retry loop backs off 5s then 10s, which would outrun the
-			// harness deadline; a PATH shim keeps the waits instantaneous.
+			// The retry loop backs off 5s then 10s; a PATH shim keeps the
+			// waits instantaneous so the test stays fast regardless of the
+			// per-script budget.
 			writeExecutable(t, filepath.Join(bin, "sleep"), noopSleep)
 
 			sbomPath := filepath.Join(dir, "aicr_1.2.3_linux_amd64.sbom.json")
@@ -1677,15 +1679,82 @@ type scriptResult struct {
 	err    error
 }
 
+// scriptHangCap bounds a single script invocation. Genuine hangs (stuck
+// prompt, infinite loop) exceed this by orders of magnitude, while
+// slow-but-correct runs under full-suite -race load stay well inside it
+// (see issue #1974).
+const scriptHangCap = 90 * time.Second
+
+// deadlineReportMargin is reserved from the test binary's -timeout so a
+// killed script still fails via t.Fatalf with its captured output instead
+// of being torn down by the test runner's deadline panic.
+const deadlineReportMargin = 30 * time.Second
+
+// scriptBudget derives the per-script deadline from the test binary's own
+// deadline so the budget inherits whatever -timeout the environment (gate,
+// CI, dev machine) was provisioned with, capped at hang-detection scale.
+func scriptBudget(t *testing.T) time.Duration {
+	t.Helper()
+	deadline, ok := t.Deadline()
+	if !ok {
+		return scriptHangCap
+	}
+	return scriptBudgetFor(time.Until(deadline))
+}
+
+// scriptBudgetFor computes the budget for the given time remaining before the
+// binary deadline. It reserves deadlineReportMargin for t.Fatalf to report
+// before the runner's deadline panic, but never takes more than half the
+// remaining time as margin — a short -timeout (e.g. 25s on a targeted run)
+// must still yield a usable budget rather than collapse it. Once the deadline
+// has passed the result goes non-positive and the context expires
+// immediately, failing fast with the script-timeout message instead of
+// starting doomed work.
+func scriptBudgetFor(remaining time.Duration) time.Duration {
+	budget := remaining - deadlineReportMargin
+	if half := remaining / 2; budget < half {
+		budget = half
+	}
+	if budget > scriptHangCap {
+		budget = scriptHangCap
+	}
+	return budget
+}
+
+func TestScriptBudgetFor(t *testing.T) {
+	tests := []struct {
+		name      string
+		remaining time.Duration
+		expected  time.Duration
+	}{
+		{"ample time hits hang cap", 10 * time.Minute, scriptHangCap},
+		{"cap boundary", scriptHangCap + deadlineReportMargin, scriptHangCap},
+		{"full margin reserved", 100 * time.Second, 70 * time.Second},
+		{"full-margin/half tie at 2x margin", 2 * deadlineReportMargin, deadlineReportMargin},
+		{"margin capped at half", 50 * time.Second, 25 * time.Second},
+		{"short targeted -timeout", 25 * time.Second, 12500 * time.Millisecond},
+		{"nearly exhausted", time.Second, 500 * time.Millisecond},
+		{"expired deadline fails fast", -10 * time.Second, -5 * time.Second},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := scriptBudgetFor(tt.remaining); got != tt.expected {
+				t.Errorf("scriptBudgetFor(%v) = %v, want %v", tt.remaining, got, tt.expected)
+			}
+		})
+	}
+}
+
 func runScript(t *testing.T, environment []string, relative string, args ...string) scriptResult {
 	t.Helper()
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	budget := scriptBudget(t)
+	ctx, cancel := context.WithTimeout(context.Background(), budget)
 	defer cancel()
 	command := exec.CommandContext(ctx, repositoryPath(t, relative), args...)
 	command.Env = environment
 	output, err := command.CombinedOutput()
 	if ctx.Err() != nil {
-		t.Fatalf("script exceeded test deadline: %v\n%s", ctx.Err(), output)
+		t.Fatalf("script exceeded %v budget (derived from -timeout): %v\n%s", budget, ctx.Err(), output)
 	}
 	return scriptResult{output: string(output), err: err}
 }
--- a/tests/releasepolicy/release_workflow_test.go
+++ b/tests/releasepolicy/release_workflow_test.go
@@ -24,7 +24,6 @@ import (
 	"sort"
 	"strings"
 	"testing"
-	"time"
 
 	"gopkg.in/yaml.v3"
 )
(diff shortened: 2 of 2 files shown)


tests/releasepolicy/release_scripts_test.go
```diff
@@ -1677,9 +1678,46 @@ type scriptResult struct {
...
+	remaining := time.Until(deadline)
+	budget := remaining - deadlineReportMargin
+	if half := remaining / 2; budget < half {
+		budget = half
+	}
+	if budget > scriptHangCap {
+		budget = scriptHangCap
+	}
+	return budget
+}
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in 6038fce27: the calculation is extracted into a pure helper `scriptBudgetFor(remaining time.Duration)` (with `scriptBudget(t)` keeping only the `t.Deadline()` handling), pinned by the table-driven `TestScriptBudgetFor` covering ample time → 90s hang cap, the exact cap boundary, full 30s margin reserved, margin capped at half the remaining time, a short targeted `-timeout` (25s → 12.5s), a nearly exhausted deadline (1s → 500ms), and an expired deadline yielding a non-positive budget for immediate fail-fast.

tests/releasepolicy/release_scripts_test.go
```diff
@@ -1677,9 +1678,74 @@ type scriptResult struct {
...
+func TestScriptBudgetFor(t *testing.T) {
+	tests := []struct {
+		name      string
+		remaining time.Duration
+		expected  time.Duration
+	}{
+		{"ample time hits hang cap", 10 * time.Minute, scriptHangCap},
+		{"cap boundary", scriptHangCap + deadlineReportMargin, scriptHangCap},
+		{"full margin reserved", 100 * time.Second, 70 * time.Second},
+		{"margin capped at half", 2 * deadlineReportMargin, deadlineReportMargin},
```
njhensley: **🔵 Nitpick — "margin capped at half" case doesn't exercise the half-bump branch**

At remaining=60s (2×deadlineReportMargin), budget = 60-30 = 30s and half = 30s are equal, so `budget < half` is false — the half-bump branch does not fire; the 30s result comes from the full-margin path. The case name implies it validates the half-cap regime, but that regime is actually exercised only by the 25s and 1s rows. Not a coverage gap (the branch is covered), just a naming/clarity mismatch.

**Blast radius:** Test readability only.

**Fix:** Optional: add an interior half-regime row, e.g. remaining=50s → budget=20 < half=25 → bumped to 25s, so the half-bump is the load-bearing assertion; or rename this case to 'full-margin/half tie at 2×margin'.

tests/releasepolicy/release_scripts_test.go
```diff
@@ -1677,9 +1678,74 @@ type scriptResult struct {
...
+				t.Errorf("scriptBudgetFor(%v) = %v, want %v", tt.remaining, got, tt.expected)
+			}
+		})
+	}
+}
+
 func runScript(t *testing.T, environment []string, relative string, args ...string) scriptResult {
 	t.Helper()
-	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), scriptBudget(t))
```
njhensley: **🔵 Nitpick — script-kill message omits the derived budget**

When a script is killed, the failure reports ctx.Err() ('context deadline exceeded') but not the budget in effect. At the default 10m -timeout the budget is always 90s, so a kill means a real hang and the message is actionable. If the package is ever run under a short -timeout, scriptBudgetFor shrinks the budget below 90s and the same message can no longer tell an operator whether the script hung (~90s) or was starved by a tight budget.

**Blast radius:** Cosmetic under current config; a diagnosis aid only in the (currently non-existent) short-timeout scenario.

**Fix:** Optional: compute `budget := scriptBudget(t)` once, pass it to both WithTimeout and the message, e.g. t.Fatalf("script exceeded %s budget (from -timeout): %v\n%s", budget, ctx.Err(), output).

tests/releasepolicy/release_scripts_test.go
```diff
@@ -1677,9 +1678,74 @@ type scriptResult struct {
...
+// deadlineReportMargin is reserved from the test binary's -timeout so a
+// killed script still fails via t.Fatalf with its captured output instead
+// of being torn down by the test runner's deadline panic.
+const deadlineReportMargin = 30 * time.Second
+
+// scriptBudget derives the per-script deadline from the test binary's own
+// deadline so the budget inherits whatever -timeout the environment (gate,
+// CI, dev machine) was provisioned with, capped at hang-detection scale.
+func scriptBudget(t *testing.T) time.Duration {
+	t.Helper()
```
njhensley, line 1697: **🔵 Nitpick — t.Helper() in scriptBudget is a no-op**

scriptBudget never calls t.Fatal/t.Error, so t.Helper() has no effect on failure-line attribution here. Harmless and conventional (matches other *testing.T helpers in the file).

**Blast radius:** None.

**Fix:** None required; leave as-is for consistency, or drop the t.Helper() call.

njhensley (dismissed):
## Multi-persona review — recommendation: ✅ Approve with comments

**Method:** independent persona reviewers (correctness / test-design, CI-DX / operability, Go idioms) → adversarial re-derivation against the resolved code. The budget math was re-verified by hand across all seven boundary cases; `TestScriptBudgetFor` passes, `go vet` and `golangci-lint -c .golangci.yaml ./tests/releasepolicy/...` are clean.

This is a clean, well-factored, test-only fix for #1974. The pure/impure split (`scriptBudgetFor` testable, `scriptBudget` isolating `t.Deadline()`) is idiomatic; naming and comments are accurate; the removed `time` import is genuinely unused; test-local constants are the right call (the package imports `pkg/defaults` nowhere and these describe test-runner deadline arithmetic, not a production default). Under `make test`/`make qualify` (`test_timeout: 10m`) every one of the ~31 `runScript` calls receives the full 90s `scriptHangCap` — 9× the old fixed 10s — so #1974's dev-machine starvation is no longer reachable, with no CI regression.

The core safety property holds: for every positive `remaining`, the derived budget is strictly `< remaining`, so the context timeout always fires before the test runner's own deadline panic, leaving margin for `t.Fatalf` to report captured output. Expired/negative deadlines fail fast correctly.

### 🟡 Minor — PR-body completeness claim is inaccurate; two same-class sites remain unconverted

The description states *"All three subprocess helpers in the package share the derived budget … No fixed-deadline subprocess sites remain in the package."* Two more subprocess helpers in the same package still use hardcoded `context.WithTimeout` deadlines on structurally identical `exec.CommandContext(ctx, "bash", …)` step-runners:

- [`release_reverify_test.go:1256`](https://github.com/NVIDIA/aicr/blob/6038fce27da1335ff00109b4218996b17fef2c6b/tests/releasepolicy/release_reverify_test.go#L1256) — `runReverifyCloseStep`, `30*time.Second`
- [`release_reverify_test.go:1486`](https://github.com/NVIDIA/aicr/blob/6038fce27da1335ff00109b4218996b17fef2c6b/tests/releasepolicy/release_reverify_test.go#L1486) — `runReverifyClassifier`, `60*time.Second`

These are the *same class* of flake #1974 describes. Their larger 30s/60s margins make recurrence materially less likely, and #1974 did not demonstrate a flake at these sites, so this is **non-blocking** — but the "no fixed-deadline sites remain" defense-in-depth the body claims isn't actually true. Either convert both to `scriptBudget(t)` for a uniform fix (one-line swap per site), or narrow the PR-body claim to the three script/guard helpers and note these two as out of scope. Not a merge blocker.

### Inline nitpicks (🔵)
Three optional polish items posted inline: a test-table case whose name over-promises (half-bump branch is a tie, not exercised, at 60s — the branch is covered elsewhere), the kill message omitting the derived budget (cosmetic at the 10m timeout), and a no-op `t.Helper()`. None require changes.

- Budget exceeding real remaining time → impossible for positive input (proven per-branch).
- Half-margin reserving too little for `t.Fatalf` → microsecond operation; 500ms is ample.
- `ctx.Err()` post-`CombinedOutput` timeout detection → canonical Go idiom, **pre-existing** (not introduced here).
- Hand-computed expected values in the test table → intentional independent oracle.
- Cosign shim comment rewrite → accurate; correctly generalizes the old fixed-10s claim.

### Notes
- The author (`yuanchen8911`) shows as `COMMENTED` on this PR — that is a self-comment, not an independent reviewer; no other human review to duplicate.
- CodeRabbit's earlier `<1s`-fallback concern is already resolved by the half-margin clamp (marked addressed in `a2fce81`).

|-----------|---------|---------|-----------|----------------|
| 0 | 0 | 1 | 3 | ✅ Approve with comments |

tests/releasepolicy/release_scripts_test.go
yuanchen8911: Fixed in 999d68b5d: renamed the 60s case to "full-margin/half tie at 2x margin" and added an interior half-regime row (50s -> 25s, where budget = 20s < half = 25s bumps), so the half-bump branch is now the load-bearing assertion of its own named case.

tests/releasepolicy/release_scripts_test.go
yuanchen8911: Fixed in 999d68b5d: the budget is computed once and included in the kill message — "script exceeded 90s budget (derived from -timeout): ..." — in runScript, and the same treatment applied to shellReleaseAssetNames and runOpenVEXGuard so all three subprocess kill sites report the budget in effect.

tests/releasepolicy/release_scripts_test.go
yuanchen8911: Agreed it's a no-op — keeping t.Helper() as-is for consistency with the other *testing.T helpers in the file, per the "leave as-is" option.

njhensley (approved):

PR #2093, opened 2026-08-06
fix(collector): emit lossless label/taint readings

### LINKED ISSUE #2003: collector/topology: label/taint disambiguation map keys can collide with real dotted label names, silently dropping readings
## Problem

`encodeLabels` (`pkg/collector/topology/topology.go`) disambiguates a label key carrying multiple distinct values as `<key>.<value>`. That synthesized map key can collide with a *different* label literally named `<key>.<value>`: both write to the same entry in the output map, and one silently overwrites the other, nondeterministically by Go map iteration order.

Concrete case (reproduced): real label `k` is `true` on gpu-a and `false` on gpu-b (→ disambiguated entries `k.true`, `k.false`), while a distinct label named `k.true` with value `true` exists on both nodes (→ plain entry keyed `k.true`). The two `k.true` writes race; either a genuine reading or the distinct label's reading is lost from the snapshot.

This is silent data loss in the collector output, independent of any consumer. `encodeTaints` has the same shape (`<key>.<effect>` disambiguation).

## Downstream impact

The node-set constraint form (#1755, PR #2000) fails closed on the detectable signature (overlapping node sets across an accepted disambiguated set violate the one-value-per-node partition), but detection-side heuristics cannot recover the lost reading, and collisions with identical node sets are undetectable in principle.

## Proposed fix

Make the encoding lossless: separate the key and value structurally instead of joining them with `.` — e.g. reserve a separator that cannot appear in label keys, or move label readings to `Subtype.Items` with explicit `key`/`value`/`nodes` fields (a snapshot-format change, so version-gated). Related: #2002 tracks structured truncation metadata; a combined structured encoding could resolve both.

## Context

Found during Codex review of PR #2000; the evaluator-side fail-closed guard shipped there, this issue tracks the collector-side root cause.

varmesh:
## Summary

The topology collector encodes a label's key and value into a single snapshot key, so two different readings can produce the same key and one is silently dropped. Adds an `items` encoding that keeps each reading's fields separate; the existing `data` map is unchanged.

## Motivation / Context

`encodeLabels` renders a multi-valued label as `<key>.<value>`, which collides with a label literally named that. `encodeTaints` has a second defect: it counts entries per key but disambiguates with effect, so two taints sharing both collapse into one entry. 

`items` carries key, value, effect and node membership as separate fields. Nothing is manufactured, so nothing can collide.

Fixes: #2003
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `pkg/constraints`, `pkg/fingerprint`, `pkg/client/v1`, `pkg/evidence/redact`

## Implementation Notes

**Additive, not a migration.** `data` is byte-identical to before — verified by capturing the same cluster with the released binary and this build and diffing both subtypes. Consumers that are not migrated cannot regress.

**No `apiVersion` bump.** Adding an optional field beside an existing one is additive-only under [ADR-011](docs/design/011-artifact-apiversion-policy.md)  matching how `da21e227` introduced `Subtype.Items`.

**One decode path.** `topology.LabelReadings` / `TaintReadings` prefer `items` and fall back to `data`, so the branch exists once rather than in each consumer. `pkg/constraints`, `pkg/fingerprint`, `pkg/snapshotter`, `pkg/client/v1` and the
shipped template all read through it. Every legacy path is untouched and still exercised by tests, so pre-existing snapshots evaluate exactly as they did.

## Testing

```bash
make qualify        # exit 0, all six stages
```

Also verified against a live 3-node Kind cluster: `data` output byte-identical between the released binary and this build, `items` present and correct, and `--max-nodes-per-entry` still capping the rendered list.

New coverage includes collision determinism (50 and 200 iterations, since the failure being guarded is a race on map ordering), legacy `data`-only compatibility, and the first rendering test for the shipped snapshot template.

## Risk Assessment

## Checklist

- [x] Commits are cryptographically signed (`git commit -S`) —

### DIFF
--- a/examples/templates/snapshot-template.md.tmpl
+++ b/examples/templates/snapshot-template.md.tmpl
@@ -103,15 +103,21 @@
 {{- end }}
 
 ## Node Topology
+{{- /* Prefer the lossless Items encoding; fall back to the folded Data map
+       for snapshots captured before it existed. Data cannot represent a label
+       whose name collides with a synthesized "<key>.<value>", so the two
+       renderings can differ in row count on such a cluster. */}}
 {{- $topoSummary := dict }}
+{{- $topoTaintItems := list }}
+{{- $topoLabelItems := list }}
 {{- $topoTaints := dict }}
 {{- $topoLabels := dict }}
 {{- range .Measurements }}
   {{- if eq .Type.String "NodeTopology" }}
     {{- range .Subtypes }}
       {{- if eq .Name "summary" }}{{ $topoSummary = .Data }}{{ end }}
-      {{- if eq .Name "taint" }}{{ $topoTaints = .Data }}{{ end }}
-      {{- if eq .Name "label" }}{{ $topoLabels = .Data }}{{ end }}
+      {{- if eq .Name "taint" }}{{ $topoTaints = .Data }}{{ $topoTaintItems = .Items }}{{ end }}
+      {{- if eq .Name "label" }}{{ $topoLabels = .Data }}{{ $topoLabelItems = .Items }}{{ end }}
     {{- end }}
   {{- end }}
 {{- end }}
@@ -126,18 +132,34 @@
 
 | Taint Key | Effect | Value | Nodes |
 |-----------|--------|-------|-------|
+{{- if $topoTaintItems }}
+{{- range $topoTaintItems }}
+| {{ index .Context "key" }} | {{ index .Context "effect" }} | {{ index .Context "value" }} | {{ (index .Data "node-list").String }} |
+{{- end }}
+{{- else }}
 {{- range $key, $reading := $topoTaints }}
 {{- $parts := splitList "|" $reading.String }}
+{{- if eq (len $parts) 3 }}
 | {{ $key }} | {{ index $parts 0 }} | {{ index $parts 1 }} | {{ index $parts 2 }} |
+{{- else }}
+| {{ $key }} | | {{ index $parts 0 }} | {{ index $parts 1 }} |
+{{- end }}
+{{- end }}
 {{- end }}
 
 ### Labels
 
 | Label Key | Value | Nodes |
 |-----------|-------|-------|
+{{- if $topoLabelItems }}
+{{- range $topoLabelItems }}
+| {{ index .Context "key" }} | {{ index .Context "value" }} | {{ (index .Data "node-list").String }} |
+{{- end }}
+{{- else }}
 {{- range $key, $reading := $topoLabels }}
 {{- $parts := splitList "|" $reading.String }}
 | {{ $key }} | {{ index $parts 0 }} | {{ index $parts 1 }} |
+{{- end }}
 {{- end }}
 
 ---
--- a/examples/templates/snapshot-template.md.tmpl
+++ b/examples/templates/snapshot-template.md.tmpl
@@ -142,7 +142,10 @@
 {{- if eq (len $parts) 3 }}
 | {{ $key }} | {{ index $parts 0 }} | {{ index $parts 1 }} | {{ index $parts 2 }} |
 {{- else }}
-| {{ $key }} | | {{ index $parts 0 }} | {{ index $parts 1 }} |
+{{- /* Two fields means encodeTaints moved the effect into the key to keep two
+       effects of one key apart, so the last dotted segment is the effect. */}}
+{{- $segments := splitList "." $key }}
+| {{ join "." (initial $segments) }} | {{ last $segments }} | {{ index $parts 0 }} | {{ index $parts 1 }} |
 {{- end }}
 {{- end }}
 {{- end }}
--- a/examples/templates/snapshot-template.md.tmpl
+++ b/examples/templates/snapshot-template.md.tmpl
@@ -1,3 +1,14 @@
+{{- /* An item carries its node names inline only when its folded key is
+       shared; otherwise it references the data entry holding them. The node
+       list is the final field of every folded encoding. */}}
+{{- define "nodeList" }}
+{{-   $item := index . 0 }}{{ $folded := index . 1 }}
+{{-   if index $item.Data "node-list" }}{{ (index $item.Data "node-list").String }}
+{{-   else if index $item.Data "node-list-ref" }}
+{{-     $entry := index $folded (index $item.Data "node-list-ref").String }}
+{{-     if $entry }}{{ last (splitList "|" $entry.String) }}{{ end }}
+{{-   end }}
+{{- end }}
 # Cluster Snapshot Report
 
 {{- /* Extract data from measurements */ -}}
@@ -134,7 +145,7 @@
 |-----------|--------|-------|-------|
 {{- if $topoTaintItems }}
 {{- range $topoTaintItems }}
-| {{ index .Context "key" }} | {{ index .Context "effect" }} | {{ index .Context "value" }} | {{ (index .Data "node-list").String }} |
+| {{ index .Context "key" }} | {{ index .Context "effect" }} | {{ index .Context "value" }} | {{ template "nodeList" (list . $topoTaints) }} |
 {{- end }}
 {{- else }}
 {{- range $key, $reading := $topoTaints }}
@@ -156,7 +167,7 @@
 |-----------|-------|-------|
 {{- if $topoLabelItems }}
 {{- range $topoLabelItems }}
-| {{ index .Context "key" }} | {{ index .Context "value" }} | {{ (index .Data "node-list").String }} |
+| {{ index .Context "key" }} | {{ index .Context "value" }} | {{ template "nodeList" (list . $topoLabels) }} |
 {{- end }}
 {{- else }}
 {{- range $key, $reading := $topoLabels }}
--- a/pkg/collector/topology/readings.go
+++ b/pkg/collector/topology/readings.go
@@ -66,9 +66,9 @@ func LabelReadings(st *measurement.Subtype) ([]LabelReading, error) {
 		return nil, nil
 	}
 	if len(st.Items) > 0 {
-		return labelReadingsFromItems(st.Items)
+		return labelReadingsFromItems(st)
 	}
-	return labelReadingsFromData(st.Data), nil
+	return labelReadingsFromData(st.Data)
 }
 
 // TaintReadings returns the readings in a NodeTopology "taint" subtype. See
@@ -82,9 +82,9 @@ func TaintReadings(st *measurement.Subtype) ([]TaintReading, error) {
 		return nil, nil
 	}
 	if len(st.Items) > 0 {
-		return taintReadingsFromItems(st.Items)
+		return taintReadingsFromItems(st)
 	}
-	return taintReadingsFromData(st.Data), nil
+	return taintReadingsFromData(st.Data)
 }
 
 // HasLosslessReadings reports whether a subtype carries the Items form.
@@ -94,31 +94,58 @@ func HasLosslessReadings(st *measurement.Subtype) bool {
 	return st != nil && len(st.Items) > 0
 }
 
-func labelReadingsFromItems(items []measurement.ItemEntry) ([]LabelReading, error) {
+func labelReadingsFromItems(st *measurement.Subtype) ([]LabelReading, error) {
+	items := st.Items
 	out := make([]LabelReading, 0, len(items))
 	for i := range items {
 		key, err := itemContext(items[i], itemCtxKey, i)
 		if err != nil {
 			return nil, err
 		}
-		nodes, count, truncated, err := itemNodes(items[i], i)
-		if err != nil {
-			return nil, err
-		}
 		out = append(out, LabelReading{
 			// An empty label value is legal, so a missing value key and an
 			// empty one both decode to "".
-			Key:       key,
-			Value:     items[i].Context[itemCtxValue],
-			Nodes:     nodes,
-			NodeCount: count,
-			Truncated: truncated,
-			RawKey:    key,
+			Key:    key,
+			Value:  items[i].Context[itemCtxValue],
+			RawKey: key,
 		})
 	}
+	// RawKey is the data entry this reading folds onto, and membership can
+	// only be resolved once the whole set is known — folding depends on how
+	// many values a key carries.
 	applyLabelRawKeys(out)
+	folds := foldCounts(rawKeysOf(out))
+	for i := range items {
+		nodes, count, truncated, err := itemNodes(items[i], i, st.Data, out[i].RawKey, folds)
+		if err != nil {
+			return nil, err
+		}
+		out[i].Nodes, out[i].NodeCount, out[i].Truncated = nodes, count, truncated
+	}
 	return out, nil
 }
+
+// rawKeysOf and foldCounts report how many readings fold onto each data entry.
+// An entry claimed by more than one reading describes only one of them, so no
+// reading may reference it.
+func rawKeysOf[T interface{ raw() string }](readings []T) []string {
+	out := make([]string, 0, len(readings))
+	for _, r := range readings {
+		out = append(out, r.raw())
+	}
+	return out
+}
+
+func foldCounts(rawKeys []string) map[string]int {
+	out := make(map[string]int, len(rawKeys))
+	for _, k := range rawKeys {
+		out[k]++
+	}
+	return out
+}
+
+func (r LabelReading) raw() string { return r.RawKey }
+func (r TaintReading) raw() string { return r.RawKey }
 
 // applyLabelRawKeys replays encodeLabels' rule over a decoded set: a key
 // carrying more than one value renders as "<key>.<value>". Disambiguation
@@ -135,14 +162,11 @@ func applyLabelRawKeys(readings []LabelReading) {
 	}
 }
 
-func taintReadingsFromItems(items []measurement.ItemEntry) ([]TaintReading, error) {
+func taintReadingsFromItems(st *measurement.Subtype) ([]TaintReading, error) {
+	items := st.Items
 	out := make([]TaintReading, 0, len(items))
 	for i := range items {
 		key, err := itemContext(items[i], itemCtxKey, i)
-		if err != nil {
-			return nil, err
-		}
-		nodes, count, truncated, err := itemNodes(items[i], i)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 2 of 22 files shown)


pkg/collector/topology/readings.go
```diff
@@ -0,0 +1,384 @@
...
+				map[bool]string{true: "carries", false: "does not carry"}[marker]))
+	case truncated && count <= len(nodes):
+		return nil, 0, false, errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("topology item %d: %q is truncated but %q (%d) does not exceed the %d nodes named",
+				idx, itemDataNodeList, itemDataNodeCount, count, len(nodes)))
+	case !truncated && count != len(nodes):
+		return nil, 0, false, errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("topology item %d: %q is complete but %q is %d against %d nodes named",
+				idx, itemDataNodeList, itemDataNodeCount, count, len(nodes)))
+	}
```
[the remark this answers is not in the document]
varmesh: The marker is space-joined, not comma-joined, so it stays inside the last element rather than becoming one

pkg/collector/topology/readings.go
```diff
@@ -0,0 +1,329 @@
...
+			return nil, err
+		}
+		nodes, count, truncated, err := itemNodes(items[i], i)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, TaintReading{
+			Key:       key,
+			Effect:    items[i].Context[itemCtxEffect],
+			Value:     items[i].Context[itemCtxValue],
```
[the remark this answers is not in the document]
varmesh: effect is now required — missing or empty is rejected, same as key.  Left out the enum check on purpose: an unknown effect matches nothing, so it's harmless.

pkg/constraints/gpu_nodes.go
```diff
@@ -180,11 +180,64 @@ func findLabelSubtype(snap *snapshotter.Snapshot) *measurement.Subtype {
...
+					"node list; regenerate the snapshot without --max-nodes-per-entry", r.RawKey),
+				map[string]any{ctxConstraint: GPUNodesLabelConstraintName, ctxReading: r.RawKey})
+		}
+		if !validNodeNames(r.Nodes) {
+			return nil, errors.NewWithContext(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("label reading %q is malformed (nodes %v) — expected canonical node names; "+
+					"regenerate the snapshot with a current aicr build", r.RawKey, r.Nodes),
+				map[string]any{ctxConstraint: GPUNodesLabelConstraintName, ctxReading: r.RawKey})
+		}
+		entries = append(entries, labelNodeSet{
```
mchmarny, line 232: Blocking: this Items path never enforces the Kubernetes invariant that a node has only one value for a given label key. A structurally valid snapshot can list `gpu-a` under both `key=true` and `key=false`; `evaluateEveryGPUNodeHasValue` ignores the false entry and returns PASS because the true entry covers the universe. The legacy Data path rejects overlapping value sets, so making Items authoritative weakens the failure direction. Reject duplicate node membership within and across readings for the same key, and add a regression asserting `ErrCodeInvalidRequest` for conflicting Items.

pkg/collector/topology/readings.go
```diff
@@ -0,0 +1,384 @@
...
+	// unreadable rather than imprecise. Both count directions matter: too low
+	// understates the cluster, too high lets a consumer read a partial list as
+	// a complete one.
+	switch {
+	case marker != truncated:
+		return nil, 0, false, errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("topology item %d: %q is %v but %q %s the truncation marker",
+				idx, itemDataTruncated, truncated, itemDataNodeList,
+				map[bool]string{true: "carries", false: "does not carry"}[marker]))
+	case truncated && count <= len(nodes):
```
mchmarny: Blocking: truncated Items only require `node-count > len(nodes)`, so `node-list: "n1,n2 (+3 more)"`, `node-count: 3`, `truncated: true` is accepted even though the suffix implies five nodes. `countGPUNodesFromItems` then records 3 while the rendered list says 5, violating the documented true-total contract. Parse the `+N` suffix and require `node-count == visible names + N`; cover both lower and higher mismatches.

mchmarny (dismissed):
Requesting changes: 2 correctness blockers in the lossless topology Items path. I reviewed exact head `7be2ea3d0b70ba8b9cb6988352a488d0325e1d35` across collector encoding/decoding, the GPU-node constraint, fingerprint consumers, the shipped template, the legacy fallback, and live CI. Checks are green, but the inline issues can accept contradictory or internally inconsistent topology data.

Coverage passes overall at 82.2%; the reported package deltas for `pkg/constraints` (-0.77%) and `pkg/fingerprint` (-1.93%) also exceed the repository 0.5% flag threshold.

pkg/collector/topology/topology.go
```diff
@@ -119,22 +119,26 @@ func (c *Collector) Collect(ctx context.Context) (*measurement.Measurement, erro
...
 			measurement.NewSubtypeBuilder("summary").
 				Set("node-count", measurement.Int(nodeCount)).
-				Set("taint-count", measurement.Int(len(taintData))).
-				Set("label-count", measurement.Int(len(labelData))),
+				Set("taint-count", measurement.Int(len(taintItems))).
+				Set("label-count", measurement.Int(len(labelItems))),
 		).
-		WithSubtype(measurement.Subtype{Name: "taint", Data: taintData}).
-		WithSubtype(measurement.Subtype{Name: "label", Data: labelData}).
+		WithSubtype(measurement.Subtype{Name: "taint", Data: taintData, Items: taintItems}).
```
yuanchen8911, line 135: Emitting items makes every pre-upgrade baseline report drift against a post-upgrade snapshot of the same unchanged cluster. compareItems (pkg/diff/diff.go:220) runs unconditionally, so a nil-items baseline yields an items.length change plus one added-change per item field, and `aicr diff --fail-on-drift` returns a conflict — the exact CI pattern documented in docs/integrator/automation.md.

Suggested rule for pkg/diff: for the label/taint subtypes, treat items as authoritative and skip data when both snapshots carry items; otherwise compare data and suppress the one-sided items representation. Mixed-vintage comparisons must also normalize the representation-sensitive summary label/taint counts — a pre-upgrade collision snapshot has undercounted summary.label-count/taint-count, so the corrected counts alone would still trip the gate. This removes upgrade-only drift; ignoring folded data when both sides carry items also stops same-version collision flapping (encodeTaints/encodeLabels iterate unordered maps with last-writer-wins on folded-key collisions), while legacy-vs-legacy comparisons retain the pre-existing limitation.

pkg/collector/topology/topology.go
```diff
@@ -119,22 +119,26 @@ func (c *Collector) Collect(ctx context.Context) (*measurement.Measurement, erro
...
+		WithSubtype(measurement.Subtype{Name: "label", Data: labelData, Items: labelItems}).
```
yuanchen8911, line 136: The items entries duplicate the full node-list strings already carried in data, roughly doubling the topology subtype. With the default `--max-nodes-per-entry=0` this is unbounded, agent mode always hands the snapshot through a ConfigMap (pkg/snapshotter/agent.go:660), and the ConfigMap read path caps at 1 MiB (pkg/serializer/reader.go:719). A 100-node/150-label probe measured the topology YAML growing from 548,124 to 1,115,885 bytes — past the limit before any other measurement is included. Snapshots that fit before this change can now fail at the internal handoff; the payload needs to stay bounded (e.g. don't carry the same membership strings twice).

pkg/constraints/gpu_nodes.go
yuanchen8911, line 232: decodeLabelEntriesFromItems appends every key-matching entry with no overlap check, while decodeLabelEntriesFromData requires accepted entries to partition their nodes (fail-closed, per its own comment). An externally produced snapshot listing one GPU node under both value=true and value=false lets evaluateEveryGPUNodeHasValue pass on the true entry alone. Since measurement-api.md is a documented producer contract, this ambiguous shape should return ErrCodeInvalidRequest like the legacy path — the overlap check ports over cheaply.

pkg/collector/topology/readings.go
```diff
@@ -0,0 +1,384 @@
...
+		})
+	}
+	sortLabelReadings(out)
+	return out
+}
+
+// taintReadingsFromData decodes the legacy taint encoding, whose two shapes
+// are told apart by field count: "<effect>|<value>|<nodes>" for a plain key,
+// "<value>|<nodes>" for a disambiguated one where the effect is the key suffix.
+func taintReadingsFromData(data map[string]measurement.Reading) []TaintReading {
```
yuanchen8911: In the two-field legacy fallback, taintReadingsFromData extracts Effect from the key.Effect suffix but leaves the suffix attached to Key, so legacy decoding returns Key "node.kubernetes.io/unreachable.NoExecute" where the items path returns "node.kubernetes.io/unreachable" — the two encodings disagree on the logical key for routine multi-effect taints.

The two-field encoding itself proves the final suffix was synthesized by encodeTaints, and the decoder already relies on that suffix to populate Effect — so it should remove the same suffix from Key, with RawKey preserving the original folded map key. Unlike the label case (where the verbatim-Key limitation is documented and genuinely ambiguous), no closed-set assumption is needed here. No in-repo consumer reads TaintReadings yet, but this is a newly exported contract.

pkg/collector/topology/readings.go
yuanchen8911: The truncation consistency check requires count > len(visible) but never parses N from the "(+N more)" marker, so node-list "n1,n2 (+3 more)" with node-count 3 is accepted although the list itself claims 5 total. countGPUNodesFromItems then fingerprints from the inconsistent count. The marker format is fixed — parse N and require count == visible + N.

pkg/collector/topology/readings.go
```diff
@@ -0,0 +1,384 @@
...
+)
+
+// LabelReading is one aggregated label reading.
+type LabelReading struct {
+	Key   string
+	Value string
+	// Nodes carry Key=Value. When Truncated, the list is incomplete and its
+	// last element holds the "(+N more)" marker.
+	Nodes []string
+	// NodeCount is the true total, including nodes omitted by truncation.
```
yuanchen8911, line 33: The struct doc says NodeCount is "the true total, including nodes omitted by truncation", but on the data fallback path (readings.go:306, :345) it is len(nodes) from the rendered list — the visible count only. No current consumer reads NodeCount from a legacy-data subtype, so this is latent, but the exported field should not have encoding-dependent semantics: prefer recovering the true total on the fallback path using the same parsed "(+N more)" marker, rather than only documenting that consumers must check Truncated first.

examples/templates/snapshot-template.md.tmpl
```diff
@@ -126,19 +132,35 @@
...
+{{- range $topoTaintItems }}
+| {{ index .Context "key" }} | {{ index .Context "effect" }} | {{ index .Context "value" }} | {{ (index .Data "node-list").String }} |
+{{- end }}
+{{- else }}
 {{- range $key, $reading := $topoTaints }}
 {{- $parts := splitList "|" $reading.String }}
+{{- if eq (len $parts) 3 }}
 | {{ $key }} | {{ index $parts 0 }} | {{ index $parts 1 }} | {{ index $parts 2 }} |
+{{- else }}
+| {{ $key }} | | {{ index $parts 0 }} | {{ index $parts 1 }} |
```
yuanchen8911: For a legacy data-only snapshot where one taint key carries multiple effects (the routine node.kubernetes.io/unreachable NoExecute/NoSchedule pair), the 2-part branch renders the synthesized "key.Effect" map key in the Taint Key column and leaves Effect blank. This branch is an improvement — the previous template hard-errored on 2-part readings — but the 2-part shape itself proves the trailing suffix is an encoded effect, so split the encoded effect suffix and render both columns correctly.

pkg/collector/topology/readings_test.go
```diff
@@ -0,0 +1,487 @@
...
+	if len(readings) != len(want) {
+		t.Fatalf("got %d readings, want %d: %+v", len(readings), len(want), readings)
+	}
+	for i, w := range want {
+		got := readings[i]
+		if got.Key != w.Key || got.Value != w.Value || got.NodeCount != w.NodeCount {
+			t.Errorf("reading %d = {%q %q n=%d}, want {%q %q n=%d}",
+				i, got.Key, got.Value, got.NodeCount, w.Key, w.Value, w.NodeCount)
+		}
+		if len(got.Nodes) != len(w.Nodes) {
```
yuanchen8911: TestLabelReadingsRecoversEveryReading compares only len(got.Nodes) against len(w.Nodes); the exact node names in the want fixtures are never asserted (the taint test does compare contents). A decode bug attributing the right number of wrong nodes would pass.

pkg/collector/topology/readings_test.go
```diff
@@ -0,0 +1,487 @@
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			item := validItem()
+			tt.mutate(&item)
+			st := &measurement.Subtype{Name: "label", Items: []measurement.ItemEntry{item}}
+			if _, err := LabelReadings(st); err == nil {
+				t.Error("LabelReadings() error = nil, want a decode error")
```
yuanchen8911: The malformed-item cases assert only err != nil, not that the error carries ErrCodeInvalidRequest via stderrors.Is. Callers are insulated today (pkg/constraints re-wraps decoder failures with its own ErrCodeInvalidRequest), but the exported decoder's error-code contract should be pinned by its own tests.

pkg/collector/topology/template_test.go
```diff
@@ -0,0 +1,108 @@
...
+		var sb strings.Builder
+		if err := serializer.NewTemplateWriter(shippedTemplate, &sb).Serialize(context.Background(), snap); err != nil {
+			t.Fatalf("render failed: %v", err)
+		}
+		return sb.String()
+	}
+
+	out := render(t)
+	for _, want := range []string{
+		"NVIDIA-H100-80GB-HBM3",
```
yuanchen8911, line 89: The fixture puts NVIDIA-B200 on gpu-b (line 62) but the rendered-output want list only checks NVIDIA-H100-80GB-HBM3 — the second value of the same label key is never asserted, which is the exact multi-value case the items encoding exists for.

yuanchen8911 (dismissed):
The lossless items encoding is well built where it counts: the legacy taint/label Data encoders are unchanged (non-collision output verified byte-identical; summary counts now derive from items, so collision-cluster summaries correct their previous undercount), the shared topology.LabelReadings/TaintReadings decoder keeps the fallback branch in one place for Go consumers (the shipped template implements its own fallback), and the collision-determinism and legacy-compat coverage is thorough. Verified at 7be2ea3, CI green.

There are two blocking issues, both at the boundaries rather than in the core logic.

First, the upgrade path breaks the documented drift gate: a baseline captured before this release has no items, pkg/diff compares items unconditionally, and `aicr diff --fail-on-drift` then exits with a conflict on an unchanged cluster — hundreds of false adds. For the label/taint subtypes, treat items as authoritative and skip data when both snapshots carry items; otherwise compare data and suppress the one-sided items representation. Mixed-vintage comparisons must also normalize the representation-sensitive summary label/taint counts, which otherwise still trip the gate on collision clusters. This removes upgrade-only drift; when both snapshots carry items, ignoring folded data also stops collision flapping, while legacy-vs-legacy comparisons retain the pre-existing limitation.

Second, emitting the full node list in both data and items roughly doubles the topology payload; with the default `--max-nodes-per-entry=0` and agent mode's mandatory ConfigMap handoff (plus the reader's own 1 MiB cap), a large-cluster snapshot that fit before this change can now fail at capture — an independent 100-node/150-label probe measured the topology YAML alone growing from ~548 KB to ~1.12 MB.

Two smaller boundary-validation issues follow the repo's fail-closed rule for externally produced snapshots (measurement-api.md is a documented integrator contract): the items decoder accepts a node listed under two values of the same label key, where the legacy decoder rejects overlapping node sets; and the truncation check never parses the "(+N more)" count, so an item whose node-count contradicts its own marker is accepted.

Two additional consistency gaps affect the legacy fallback: NodeCount is not the documented true total when truncated, and legacy taint decoding leaves the synthesized .Effect suffix attached to the logical Key.

The rest is test-depth and rendering polish, inline.

pkg/collector/topology/topology.go
yuanchen8911: I agree with items cross-referencing data for v1alpha2.

Slimming data is not safe: pre-PR binaries read it directly. Their node-set evaluator rejects truncated lists, while fingerprinting does not recover the (+N more) count. An older same-version consumer could therefore fail validation or undercount GPU nodes. ADR-011 requires the existing data encoding and semantics to remain unchanged.

For the cross-reference:

- Omit node-list only when the folded key uniquely identifies one item; collisions keep the inline list.
- Use an explicit reference to the data entry. An omitted node-list without that reference must fail decoding.
- Centralize hydration and validation in LabelReadings and TaintReadings, checking uniqueness, identity, count, and truncation.
- In pkg/diff, compare hydrated items when both snapshots have them; retain data normalization for mixed-vintage comparisons.

Agreed on removing data in the next snapshot apiVersion. New emissions can then use self-contained items, while the v1alpha2 compatibility decoder remains through its acceptance window. Worth documenting that transition in measurement-api.md.

yuanchen8911:
Verified all open review threads against the current head (4135d987).

My earlier comments are all addressed — thanks for the thorough follow-up.

One remaining item from the latest review round looks valid: in readings.go (~line 445), a legacy two-field taint entry whose map key has no dot suffix decodes to a reading with Effect == "" instead of erroring. The encoder always produces a dotted key for the two-field shape, so a suffix-less key is malformed input and should fail closed with ErrCodeInvalidRequest, same as the items path rejecting an empty effect. Should be a small fix plus one test case.

LGTM once that's addressed.

yuanchen8911 (dismissed):
All my previous comments are addressed, and Mark's two blockers are fixed with the requested regression tests. One non-blocking item remains (taint decode fail-closed on a suffix-less two-field key, readings.go ~445) — fine as a follow-up.

yuanchen8911 (approved):
All review findings are addressed at head 776e0e61. The last remaining item — fail-closed taint decode on a suffix-less two-field key (readings.go ~445) — is fixed with regression tests covering both malformed shapes, plus multi-node coverage for hasHeterogeneousGPUPool on both encodings. Verified the diff and green CI on the completed checks.

Issue #1873, opened 2026-07-23
Harden CheckDriverOwnershipCoherence against post-validation drift/bypass


yuanchen8911:
## Context

Follow-up from #1819 (fixes #1757), which added the bundle-time `CheckDriverOwnershipCoherence` gate. The multi-reviewer cross-review of #1819 surfaced two ways the gate's verdict can be undermined **after** bundle-time validation. Both were deliberately split out of #1819 to keep it focused; this issue tracks closing them.

Reference implementations for both exist in #1819's pre-trim history (commit `25bd52fda`).

## A — Values TOCTOU closure (exact-value reuse)

The gate runs on both bundle paths — `DefaultBundler.Make` (the CLI and the server's `MakeBundle`) and the public `Client.BundleComponents` SDK path — but each currently **self-resolves** component values independently of the read used to emit the bundle. With a mutable provider (`LayeredDataProvider` re-reads external `--data` files on every call, `pkg/recipe/provider.go`), the values the gate validates and the values written into the bundle can differ within a single operation.

**Goal:** resolve component values once and validate the exact maps that are emitted, on both bundle paths.

**Acceptance criteria:**
- Bundle generation resolves each component's effective values once and reuses that exact map for both the coherence check and the emitted bundle.
- Covers both `DefaultBundler.Make` and `Client.BundleComponents`.
- A `FilesystemSource` / `LayeredDataProvider` mutation test proves the gate examines exactly the values emitted (mutating the external file between the two reads cannot slip a different value past the gate).

## B — Argo CD Helm install-time ownership guards

The `argocd-helm` deployer merges every value under a component's override key over the static bundle values at **install time**, so an operator can `--set` a guarded driver-ownership leaf (`gpuoperator.driver.enabled`, `toolkit.enabled`, `hostPaths.driverInstallDir`, `dradriver.nvidiaDriverRoot`) and undo the bundle-time verdict.

**Goal:** the generated `argocd-helm` chart rejects install-time overrides of the guarded ownership leaves — including scalar/null parent replacement that would drop a guarded leaf — while leaving unrelated siblings overridable.

**Acceptance criteria:**
- Helm-render-time `fail` guards on the guarded leaves in the generated `argocd-helm` chart.
- Unrelated overrides (e.g. `gpuoperator.driver.version`) still render successfully.
- Regression + live `helm template` tests.

## Notes

[...]

PR #2095, opened 2026-08-06
fix(sdk): context-aware selector, route snapshot Jobs through facade

### LINKED ISSUE #2020: sdk: SelectFromRecipe is not context-aware and both callers bypass it
## Summary

`aicr.SelectFromRecipe` — the facade's entry point for `aicr query`-style dot-path selection — takes no `context.Context`, and both in-tree callers bypass it entirely. It is the one exported function in `pkg/client/v1` that cannot be canceled, and it has already drifted from the code paths that are actually exercised.

## Current behavior

`pkg/client/v1/query.go`:

```go
func SelectFromRecipe(r *RecipeResult, selector string) (any, error) {
	...
	hydrated, err := recipe.HydrateResult(internal)
```

`recipe.HydrateResult` manufactures its own context:

```go
func HydrateResult(result *RecipeResult) (map[string]any, error) {
	ctx, cancel := context.WithTimeout(context.Background(), defaults.FileReadTimeout)
	defer cancel()
	return HydrateResultWithContext(ctx, result)
}
```

Hydration performs values-file reads through the data provider, so this is I/O under a self-created `context.Background()` — contrary to the repo's own rule that all I/O operations take a caller-supplied context with timeout.

Meanwhile both real callers use the context-aware variant directly and never touch the facade function:

- `pkg/cli/query.go:135` — `recipe.HydrateResultWithContext(ctx, result.Resolved())` then `recipe.Select(...)`
- `pkg/server/recipe_handler.go:492` — same pair, with a comment explaining it avoids `aicr.SelectFromRecipe` "to preserve the legacy handler's distinct error" shape

So an integrator calling the documented facade function gets *different* cancellation behavior from what the CLI and server do, and the server has already written down that the facade version does not meet its needs.

## Proposed change

Add a context-aware form and make it the one implementation:

```go
func SelectFromRecipeWithContext(ctx context.Context, r *RecipeResult, selector string) (any, error)
```

Then route both in-tree callers through it. If the server's distinct error shaping is still required, that is a thin mapping at the handler rather than a reason to maintain a second code path — capture what it needs (structured codes on the hydrate vs select failure) in the facade function's contract.

Keep the existing `SelectFromRecipe` as a documented wrapper over the new form for source compatibility, or remove it if the API-diff gate (#2018) is already in place to record the break. Whichever way, it must not remain the only spelling.

## Acceptance criteria

- The facade exposes a context-accepting selector entry point, and hydration honors the caller's context and cancellation.
- `pkg/cli/query.go` and `pkg/server/recipe_handler.go` call the facade rather than `recipe.HydrateResultWithContext` + `recipe.Select` directly.
- The server's error-shape requirement is met without a parallel implementation, or is documented as a handler-level mapping.
- A test proves a canceled context aborts hydration.

Parent: #2016

### LINKED ISSUE #2021: sdk: route aicr snapshot through CollectSnapshot; close BundleComponents drift
## Summary

`Client.CollectSnapshot` and `Client.BundleComponents` are exported, documented, and pinned by the stability test — and have **zero production callers in this repository**. They are exercised only by their own unit tests. Code that no in-tree consumer runs is code that drifts from the paths we actually maintain, and integrators are the ones who discover it.

## Current state

**`CollectSnapshot`** — `aicr snapshot` does not use it. `pkg/cli/snapshot.go` does not import `pkg/client/v1` at all; it builds a `collector.NewDefaultFactory(...)`, assembles a `snapshotter.AgentConfig` itself, and drives the snapshotter directly. The facade method wraps `snapshotter.DeployAndGetSnapshot` with a facade-owned `AgentConfig` that is a hand-maintained field-for-field mirror (`translate.go:toInternalAgentConfig`). Every field added to `snapshotter.AgentConfig` must be manually mirrored in two more places, with nothing failing if it is not — the mirror's own doc comment acknowledges this: *"a future field added to either side stays at its zero value until plumbed through."*

`pkg/cli/snapshot.go` also carries behavior the facade has no equivalent for: the in-pod agent mode (`AICR_AGENT_MODE`), network-collector options (`--cluster-config`, `--discover-network`), and resource request/limit parsing.

**`BundleComponents`** — no caller. The CLI and the server both use `MakeBundle`. It is a genuinely different operation (in-memory per-component values and manifests rather than a written bundle tree), and it is plausibly what an integrator embedding AICR wants most — which makes its lack of exercise the more concerning, not less. #1873 item A already identifies a values-TOCTOU on this exact path: it self-resolves component values independently of what it emits, so with a mutable `LayeredDataProvider` the values a validation gate examines can differ from the values returned.

## Proposed change

Two independent halves; the first is the important one.

**A — Route `aicr snapshot` through `Client.CollectSnapshot`.** Extend the facade `AgentConfig` and `CollectSnapshot` to cover what the CLI needs (collector factory options, agent mode, resource lists), then make the CLI a thin adapter over it. This is the pattern `aicr validate` already follows successfully. Outcome: the mirror is exercised on every snapshot run, and an unplumbed field becomes a test failure instead of a silent zero value.

If some part genuinely cannot move — the in-pod agent path is a plausible candidate, since it does not deploy a Job at all — say so explicitly in the facade godoc rather than leaving the divergence undocumented.

**B — Give `BundleComponents` a real consumer or real coverage.** Preferred: fix #1873 item A here by resolving each component's effective values **once** and reusing that exact map for both the coherence check and the returned bundle, then add an integration test that mutates an external `--data` file between reads and proves the gate sees exactly what is emitted. If no in-tree consumer is plausible, that test is the substitute for one and should be treated as load-bearing.

## Acceptance criteria

- `aicr snapshot` produces byte-identical output before and after routing through the facade (covered by existing snapshot e2e/coverage tests).
- Any snapshot capability deliberately left outside the facade is named in the `CollectSnapshot` godoc with the reason.
- `BundleComponents` resolves values once and validates the exact maps it returns, with a `LayeredDataProvider` mutation test proving it.
- No exported facade method remains with neither an in-tree caller nor integration-level coverage.

Parent: #2016
Related: #1873

mchmarny:
## Summary

Close two pre-stability drift items on the `pkg/client/v1` facade: make the
dot-path selector context-aware and route every in-tree snapshot Job through
`Client.CollectSnapshot`, with the facade↔internal `AgentConfig` mirror now
enforced by a test.

## Motivation / Context

Both issues are children of #2016 (facade parity before the surface is
declared stable) and overlap in `pkg/client/v1` + `pkg/cli`, so they land
together.

While implementing, the drift turned out to be worse than the issues
described:

- The facade `AgentConfig` was **already missing** `ClusterConfigPath` and
  `DiscoverNetwork` — exactly the silent-zero-value failure the mirror's own
  doc comment predicted.
- `AgentConfig.Output` was a **dead field**: `DeployAndGetSnapshot` computed
  its own ConfigMap target and ignored it.
- A **third** hand-rolled mirror existed at `pkg/cli/validate.go:161`, which
  built a `snapshotter.AgentConfig` inline and bypassed `CollectSnapshot` too.
- For the TOCTOU, `BundleComponents` already resolved values once for
  *emission*; the divergent read is in the gate
  (`validations.effectiveComponentValues`), which re-resolves independently.

Fixes: #2020
Fixes: #2021
Related: #2016, #1873

## Type of Change

- [x] Documentation update

## Component(s) Affected

- [x] Other: SDK facade (`pkg/client/v1`)

### #2020 — context-aware selector

`SelectFromRecipeWithContext(ctx, *RecipeResult, selector)` is now the single
implementation; `SelectFromRecipe` remains as a documented wrapper that derives
a `defaults.FileReadTimeout`-bounded context, for source compatibility.

`pkg/cli/query.go` and `pkg/server/recipe_handler.go` both call it. The
server's 404-vs-5xx requirement is met as a **handler-level mapping** over a
documented contract rather than a second code path: selector-path failures
carry `ErrCodeNotFound` as the **outermost** structured code, hydration
failures never do (`HydrateResultWithContext` hard-wraps every per-component
failure as `ErrCodeInternal`).

The handler matches with `stderrors.As` on the outermost node, **not**
`stderrors.Is` — `Is` walks the wrap chain and would misread a hydration
failure whose cause happens to carry `ErrCodeNotFound` (a missing values file)
as a missing selector path. This is a deliberate deviation from the repo's
usual "prefer `errors.Is` for code matching" rule, commented at the call site.

`WrapResolved(*recipe.RecipeResult) *RecipeResult` is new: the query handler
projects a legacy (`/v1`) view of the resolved recipe before selecting, and
needs to re-enter the facade with it. The wrapped result is **queryable only**
— no owning Client, so `MakeBundle` / `BundleComponents` / `ValidateState`
reject it.

### #2021 item A — snapshot routing

`pkg/snapshotter` splits Job-mode collection into two exported steps:

- `DeployAndCollect(ctx, cfg) (*Snapshot, []byte, error)` — deploy, wait,
  parse, and return **both** the typed snapshot and the raw agent bytes.
- `DeliverSnapshot(ctx, data, templatePath, output) error` — template /
  stdout / ConfigMap / file delivery.

`NodeSnapshotter.measureWithAgent` is now the composition of the two, so the
CLI and facade paths emit the same bytes **by construction**. `Snapshot.Raw`
carries the agent's exact document; delivering it (rather than re-serializing
the parsed struct) is what keeps `aicr snapshot` output byte-identical when a
newer agent image emits fields this binary does not model.

`aicr snapshot` (Job mode) and `aicr validate` both now go through
`Client.CollectSnapshot`. `Snapshot.Unwrap()` (inverse of `WrapSnapshot`, analog
of `RecipeResult.Resolved()`) gives validate the measurement-level payload it
needs.

**Deliberately left outside the facade**, named in the `CollectSnapshot` godoc
with reasons, per the issue's acceptance criteria:

- Local (`AICR_AGENT_MODE`) collection — deploys no Job; needs a
  `collector.Factory` and `serializer.Serializer`, types the semver-stable
  facade does not expose. It projects config down via the now-exported
  `ToInternalAgentConfig` rather than assembling a second copy.
- `ClusterConfigPath` in Job mode — the path must resolve inside the pod and
  the Job does not mount it; rejected with `ErrCodeInvalidRequest` as before.

`TestAgentConfigMirrorsInternal` makes the mirror an invariant with two
assertions: **shape** (same-named, same-typed fields both ways) and **value**
(every field set to a distinct non-zero value must survive the projection —
catching a field present on both sides but missing from
`ToInternalAgentConfig`). A new field of an unmodeled type fails the test
rather than being skipped. Both halves were verified to fail against
deliberately introduced drift before being kept.

### #2021 item B — values TOCTOU (`#1873` item A, SDK path)

`BundleComponents` now resolves the Helm-value inventory **once**, before any
gate, and pins it onto the recipe result the gates see via the new
`RecipeResult.WithResolvedValues` (a shallow copy serving deep copies, so a
gate that layers `--set` overrides onto what it reads cannot corrupt the
emitted map). Scoped to the SDK path as #2021 specifies; `DefaultBundler.Make`
remains open under #1873.

### Review-round fixes

Three hazards surfaced in review, all closed in this branch:

1. **Malformed `cm://` output accepted by prefix only.** `agentConfigMapTarget`
   matched the scheme and passed the string straight to the Job, so
   `cm://aicr-snapshot` (name, no namespace) only failed inside the pod —
   after RBAC and the Job existed, and with `Cleanup` false (the zero value
   SDK callers get) they persist, including a cluster-admin binding. The URI
   is now fully parsed with `pod.ParseConfigMapURI`, and both it and a
   Job-mode `ClusterConfigPath` are rejected **before the Kubernetes client is
   built** — so a rejection is never masked by a kubeconfig error and never
   mutates the cluster. `TestDeployAndCollectRejectsBeforeClusterAccess` and
   `TestCollectSnapshot_RejectsBeforeClusterAccess` assert the *ordering* by
   passing an unusable kubeconfig and requiring the input error to win.
   (The duplicate `ClusterConfigPath` check inside `deployAndWaitForResult`
   was removed — it now has a single caller.)

2. **`DeliverSnapshot`'s `cm://` branch logged success without writing.** That
   is only correct when the same URI was `AgentConfig.Output` at collection
   time; an integrator who collected to the default internal ConfigMap and
   then delivered to `cm://ns/name` got no artifact **and** no error, while
   the integrator guide claimed delivery writes ConfigMaps. Delivery now
   performs the write (idempotent when the Job already staged those bytes),
   making the destination set total. Parameters moved to a `SnapshotDelivery`
   struct to carry the kubeconfig that requires. The old
   `rewriteSnapshotConfigMap` became `writeSnapshotConfigMap`, shared by
   delivery and the AKS-pool merge.

3. **CLI truncated the output file before collecting** (CodeRabbit,
   outside-diff). `createSnapshotSerializer` opens the destination with
   `os.Create`, and it was built above the mode branch — so on the Job path,
   which never uses it, a failed collection left the previous capture erased.
   Construction moved into the local-collection branch that actually consumes
   it. Job-mode destinations are validated without side effects instead: the
   template by `parseSnapshotCmdOptions`, the `cm://` URI by
   `DeployAndCollect`.

A second review round surfaced three more, also fixed:

4. **Empty `Namespace` reached the cluster before failing.** The `cm://`
   branch was validated, but the default branch built `cm://<Namespace>/aicr-snapshot`
   unchecked — an empty namespace yields `cm:///aicr-snapshot`, rejected only
   when finally parsed, after RBAC and the Job. Guarded in `DeployAndCollect`
   rather than in `agentConfigMapTarget`, because `Namespace` is required in
   **both** output modes: the Job and its RBAC are created there regardless of
   whether `Output` is a `cm://` URI, so guarding only the default branch would
   still let a `cm://ns/name` caller with an empty namespace mutate the cluster.
   The CLI is insulated (both commands default the flag); this is the
   SDK-caller path.

5. **Orphaned godoc fragment on `DeployAndCollect`.** Not a typo — since Go
   1.19 `gofmt` treats `//nolint:...` as a directive and moves it to the end of
   the doc block, so rationale written as continuation lines under it got
   hoisted above and orphaned mid-sentence. Rewriting in place would just be
   re-broken by the next `gofmt`; the rationale is now a normal paragraph
   before a single-line directive, with a note so it is not reintroduced.

6. **`ToInternalAgentConfig` un-exported.** It was exported on the stated
   grounds that the CLI's local-collection path needed it — which turned out to
   be false (that path builds a `collector.Factory` and `serializer.Serializer`
   from `opts`, and *cannot* take an `AgentConfig`, since setting
   `NodeSnapshotter.AgentConfig` is what selects Job mode). That left an
   exported facade symbol with no caller outside the package: the exact drift
   this PR exists to remove. Back to unexported, stability pin dropped; the
   mirror test is an internal test, so enforcement is unaffected.

**Declined:** a suggestion to swap `context.Background()` for `context.TODO()`
in `SelectFromRecipe`. `AGENTS.md`'s canonical GOOD example for a context-less
wrapper is `context.WithTimeout(context.Background(), ...)` — this pattern
verbatim — and the repo uses it in eight places including `recipe.HydrateResult`,
the very function this wrapper calls. `context.TODO()` means "unclear which
context, fill in later"; here the absence is deliberate and permanent, which is
why `SelectFromRecipeWithContext` exists and is documented as preferred.

Also from CodeRabbit: `TestDeliverSnapshot_ConfigMapIsNoOp` asserted on a
`t.TempDir()` it never entered, so `os.ReadDir` was vacuously empty. It is
replaced by `TestDeliverSnapshot_ConfigMapRejectsMalformedURI` (real
assertions, no cluster needed) and `TestDeliverSnapshot_ConfigMapWritesNoLocalFile`
(uses `t.Chdir`, and pins that the ConfigMap branch never falls through to
file output).

### Behavior changes (2, both intentional)

1. **`CollectSnapshot` timeout.** Bounded by `AgentConfig.Timeout +
   defaults.SnapshotOperationGrace` (new, 1m). `AgentConfig.Timeout` budgets
   Job *completion* only — deploy, `--aks-gpu-pools` projection, and ConfigMap
   retrieval sit outside it — so a bare cap would have silently shrunk the
   completion budget by however long deployment took, failing Jobs on slow
   clusters that previously succeeded.
2. **Unparseable snapshot on `aicr snapshot`.** Previously this path parsed
   best-effort (skipping only the GPU-placement warning) and delivered the
   bytes anyway; the SDK path already failed. Converged on strict: a document
   `aicr snapshot` cannot parse is one `aicr validate` and
   `aicr recipe --snapshot` would reject too, so writing it out and exiting 0
   reports a success the artifact cannot back.

### Not included

An unrelated missing license header on
`tools/registry-inventory/registry-allowlist.yaml` (added by `make lint`'s
`license` fixer during qualification) was reverted to keep this PR scoped —
`make license-check` passes without it.

## Testing

```bash
make qualify
```

Full `make qualify` passes. Two stages fail **only** under the agent sandbox
and were re-run without it to confirm:

- `check-agents-sync` — its `diff` uses process substitution (`/dev/fd/63`),
  which the sandbox blocks; fails identically on `main`. Unsandboxed:
  `OK: AGENTS.md is in sync with .claude/CLAUDE.md`.
- `./tools/e2e` — chainsaw fixtures write to `/tmp/chainsaw-*`, outside the
  sandbox write allowlist. Unsandboxed: **24 passed, 0 failed, 0 skipped**.

Other stages: `test-coverage` 82.3% (threshold 80%), `golangci-lint ./pkg/...`
0 issues, `go test -race ./pkg/... ./cmd/...` clean, `make scan` clean.

**New tests**

- `TestSelectFromRecipeWithContextCancellation` — a canceled context aborts
  hydration, and the failure is classified as hydration (never `NotFound`).
- `TestSelectFromRecipeNotFoundIsOutermostCode` — pins the code contract the
  handler maps on.
- `TestWrapResolvedIsQueryable` — selector parity with the resolved result;
  rejected by the bundle path.
- `TestAgentConfigMirrorsInternal` — shape + value mirror invariant.
- `TestBundleComponents_ResolvesValuesOnce` — a provider whose document
  changes on every read proves one resolution feeds both gate and emission.
  **Verified to fail pre-fix** (`values file read 2 times, want exactly 1`).
- `TestBundleComponents_PinnedValuesSurviveGateMutation` — the pin hands out
  deep copies, so a gate's `--set` layering cannot leak into the bundle.
- `TestWithResolvedValues` — pin semantics (nil receiver, empty snapshot,
  pinned vs unpinned components, deep copies, cancellation, `DeepCopy` drops
  the pin).
- `TestDeliverSnapshot_*` — file delivery is byte-identical including a field
  the local `Snapshot` type does not model; `cm://` rejects malformed URIs and
  never falls through to file output; template rendering; template mode
  rejects an unparseable document.
- `TestAgentConfigMapTargetRejectsMalformedURI`,
  `TestDeployAndCollectRejectsBeforeClusterAccess`,
  `TestCollectSnapshot_RejectsBeforeClusterAccess` — fail-before-mutate for a
  malformed `cm://` URI, an empty/whitespace `Namespace`, and a Job-mode
  `ClusterConfigPath`. Each passes an unusable kubeconfig so the assertion
  proves the *ordering* relative to building the Kubernetes client, not just
  the rejection.
- `TestSnapshotUnwrapRoundTrips` — `WrapSnapshot`/`Unwrap` returns the same
  pointer; facade-constructed snapshots rebuild minimally; `Raw` stays empty
  outside `CollectSnapshot`.
- `TestAgentOutputURILogic` was converted from a *simulation* of
  `measureWithAgent`'s logic into a real call to `agentConfigMapTarget`, and
  now also asserts the `deliverViaConfigMap` flag.

**Coverage deltas** (per-package, branch vs `origin/main`)

| Package | Before | After | Δ |
|---|---|---|---|
| `pkg/client/v1` | 81.8% | 81.9% | +0.1% |
| `pkg/snapshotter` | 54.2% | 55.9% | +1.7% |
| `pkg/cli` | 74.9% | 74.8% | -0.1% |
| `pkg/server` | 82.6% | 82.6% | — |
| `pkg/recipe` | 100.0% | 100.0% | — |

No new exported function is left at 0% coverage.

## Risk Assessment

Spans the snapshot collection path (`pkg/snapshotter`), the SDK facade, both
CLI commands that capture snapshots, and the REST query handler. Mitigations:
`measureWithAgent` is now a composition of the same two functions the facade
calls, so byte-identity holds by construction rather than by assertion; the
full 24-test e2e suite passes; and the mirror invariant is verified to fail
against injected drift.

**Rollout notes:** No migration. `ToInternalAgentConfig` was exported earlier in
this PR and is now unexported again — it never shipped. `snapshotter.DeliverSnapshot` changed shape
during review (positional `templatePath, output` → a `SnapshotDelivery`
struct); it was introduced in this same PR, so nothing external depends on the
earlier form. Both facade signature changes are additive —
`SelectFromRecipe` is retained as a wrapper, and `CollectSnapshot` keeps its
signature. `snapshotter.DeployAndGetSnapshot` is replaced by
`DeployAndCollect`; `pkg/snapshotter` is not covered by the facade's semver
guarantee (see `docs/integrator/public-api.md`), and it had no external
consumers in tree.

## Checklist

### DIFF
--- a/pkg/cli/snapshot.go
+++ b/pkg/cli/snapshot.go
@@ -24,6 +24,7 @@ import (
 	"github.com/urfave/cli/v3"
 	corev1 "k8s.io/api/core/v1"
 
+	aicr "github.com/NVIDIA/aicr/pkg/client/v1"
 	"github.com/NVIDIA/aicr/pkg/collector"
 	"github.com/NVIDIA/aicr/pkg/config"
 	"github.com/NVIDIA/aicr/pkg/defaults"
@@ -120,10 +121,15 @@ type snapshotCmdOptions struct {
 	tmplOpts        *snapshotTemplateOptions
 }
 
-// toAgentConfig converts the resolved options into the snapshotter.AgentConfig
-// the snapshotter deploy path expects.
-func (o *snapshotCmdOptions) toAgentConfig() *snapshotter.AgentConfig {
-	return &snapshotter.AgentConfig{
+// toAgentConfig converts the resolved options into the facade AgentConfig that
+// Client.CollectSnapshot consumes. The facade type — not
+// snapshotter.AgentConfig — is the single spelling the CLI builds, so the
+// facade/internal mirror is exercised on every snapshot run instead of only
+// by SDK consumers. Job mode is its only consumer: local (in-pod) collection
+// deploys no Job and builds a collector.Factory and serializer.Serializer from
+// opts directly, so it never needs an AgentConfig.
+func (o *snapshotCmdOptions) toAgentConfig() *aicr.AgentConfig {
+	return &aicr.AgentConfig{
 		Kubeconfig:         o.kubeconfig,
 		Namespace:          o.namespace,
 		Image:              o.image,
@@ -507,59 +513,88 @@ See examples/templates/snapshot-template.md.tmpl for a sample template.
 				return err
 			}
 
-			// Create factory. Network-collector options come from the
-			// flags we just parsed; their env-var Sources mean a Job
-			// running our binary inside the cluster picks up
-			// AICR_CLUSTER_CONFIG_PATH / AICR_DISCOVER_NETWORK
-			// straight into opts before we get here, so this single
-			// builder serves both the local-mode bypass and the
-			// inside-the-Job rerun without a separate rebuild path.
-			factory := collector.NewDefaultFactory(
-				collector.WithMaxNodesPerEntry(opts.maxNodesPerEntry),
-				collector.WithOS(opts.os),
-				collector.WithClusterConfigPath(opts.clusterConfigPath),
-				collector.WithDiscoverNetwork(opts.discoverNetwork),
-				collector.WithKubeconfigPath(opts.kubeconfig),
-			)
-
-			// Create output serializer
-			ser, err := createSnapshotSerializer(opts.tmplOpts, opts.kubeconfig)
-			if err != nil {
-				return errors.Wrap(errors.ErrCodeInternal, "failed to create output serializer", err)
-			}
-			if c, ok := ser.(serializer.Closer); ok {
-				defer func() {
-					if closeErr := c.Close(); closeErr != nil {
-						slog.Warn("failed to close snapshot serializer", "error", closeErr)
-					}
-				}()
-			}
-
-			// Build snapshotter configuration
-			ns := snapshotter.NodeSnapshotter{
-				Version:         version,
-				Factory:         factory,
-				Serializer:      ser,
-				RequireGPU:      opts.requireGPU,
-				AKSGPUPoolsPath: opts.aksGPUPoolsPath,
-			}
+			agentCfg := opts.toAgentConfig()
 
 			// When running inside an agent Job, collect locally instead of
-			// deploying another agent (prevents infinite nesting). The
-			// already-built `factory` above carries every CLI-resolved
-			// option — opts.clusterConfigPath, opts.discoverNetwork,
-			// opts.kubeconfig, opts.os, opts.maxNodesPerEntry — so we
-			// reuse it directly. The flags' cli.EnvVars Sources have
-			// already populated opts from the Job-set env vars before
-			// we reach this point, which keeps the dev-bypass case
-			// (`AICR_AGENT_MODE=true aicr snapshot --cluster-config
-			// <path>`) consistent with the in-pod path.
+			// deploying another agent (prevents infinite nesting). This path
+			// deploys nothing, so it stays on pkg/snapshotter directly — it
+			// needs a collector.Factory and a serializer.Serializer, which the
+			// semver-stable facade deliberately does not expose (see the
+			// Client.CollectSnapshot godoc).
+			//
+			// The factory carries every CLI-resolved option —
+			// opts.clusterConfigPath, opts.discoverNetwork, opts.kubeconfig,
+			// opts.os, opts.maxNodesPerEntry. The flags' cli.EnvVars Sources
+			// have already populated opts from the Job-set env vars before we
+			// reach this point, which keeps the dev-bypass case
+			// (`AICR_AGENT_MODE=true aicr snapshot --cluster-config <path>`)
+			// consistent with the in-pod path.
+			//
+			// Both the factory and the serializer are built HERE rather than
+			// above the branch: createSnapshotSerializer opens the destination
+			// with os.Create, which truncates it. Building it on the Job path —
+			// which never uses it, delivering raw agent bytes instead — would
+			// erase an existing snapshot file before the Job had produced a
+			// replacement, so a failed collection would destroy the previous
+			// capture. Job-mode destinations are instead validated without
+			// side effects: parseSnapshotCmdOptions checks the template, and
+			// DeployAndCollect parses a cm:// URI before touching the cluster.
 			if os.Getenv("AICR_AGENT_MODE") == "true" {
+				factory := collector.NewDefaultFactory(
+					collector.WithMaxNodesPerEntry(opts.maxNodesPerEntry),
+					collector.WithOS(opts.os),
+					collector.WithClusterConfigPath(opts.clusterConfigPath),
+					collector.WithDiscoverNetwork(opts.discoverNetwork),
+					collector.WithKubeconfigPath(opts.kubeconfig),
+				)
+
+				ser, serErr := createSnapshotSerializer(opts.tmplOpts, opts.kubeconfig)
+				if serErr != nil {
+					return errors.Wrap(errors.ErrCodeInternal, "failed to create output serializer", serErr)
+				}
+				if c, ok := ser.(serializer.Closer); ok {
+					defer func() {
+						if closeErr := c.Close(); closeErr != nil {
+							slog.Warn("failed to close snapshot serializer", "error", closeErr)
+						}
+					}()
+				}
+
+				ns := snapshotter.NodeSnapshotter{
+					Version:         version,
+					Factory:         factory,
+					Serializer:      ser,
+					RequireGPU:      opts.requireGPU,
+					AKSGPUPoolsPath: opts.aksGPUPoolsPath,
+				}
 				return ns.Measure(ctx)
 			}
 
-			ns.AgentConfig = opts.toAgentConfig()
-			return ns.Measure(ctx)
+			// Job mode goes through the facade so `aicr snapshot`, `aicr
+			// validate`, and SDK consumers all deploy the agent the same way.
+			// The recipe source is irrelevant here — CollectSnapshot does not
+			// consult the Client's DataProvider — so the embedded source keeps
+			// construction free of cluster- or flag-dependent setup.
+			client, err := aicr.NewClient(aicr.WithRecipeSource(aicr.EmbeddedSource()))
+			if err != nil {
+				return err
+			}
+			defer func() { _ = client.Close() }()
+
+			snap, err := client.CollectSnapshot(ctx, agentCfg)
+			if err != nil {
+				return err
+			}
+
+			// Deliver the RAW agent bytes, never a re-serialization of the
+			// parsed snapshot: a newer agent image can emit fields this
+			// binary's Snapshot type does not model, and a typed round trip
+			// would silently drop them.
+			return snapshotter.DeliverSnapshot(ctx, snap.Raw, snapshotter.SnapshotDelivery{
+				Output:       agentCfg.Output,
+				TemplatePath: agentCfg.TemplatePath,
+				Kubeconfig:   agentCfg.Kubeconfig,
+			})
 		},
 	}
 }
--- a/pkg/client/v1/aicr.go
+++ b/pkg/client/v1/aicr.go
@@ -1119,6 +1119,18 @@ func (c *Client) LoadRecipe(ctx context.Context, path, kubeconfig string) (*Reci
 // and component registry were already per-Client at the time and
 // stayed correct throughout, so ResolveRecipe results never drifted.
 //
+// # Read-once value coherence
+//
+// Each component's effective Helm values are resolved exactly once per
+// call, and that same snapshot feeds the accounting check, the
+// registry-declared component validations (including the gpu-operator
+// driver-ownership coherence gate), and the returned ComponentBundle. A
+// DataProvider need not be stable — LayeredDataProvider re-reads external
+// --data files on every call — so a gate that resolved values
+// independently could validate one set of values while a different set was
+// returned. Pinning the snapshot removes that window: what the gates
+// examined is what you get back (issue #1873 item A).
+//
 // # Synchronization
 //
 // Read-locks Client.mu so a concurrent Close can't race the values
@@ -1177,23 +1189,6 @@ func (c *Client) BundleComponents(ctx context.Context, r *RecipeResult) ([]Compo
 		return nil, errors.Wrap(errors.ErrCodeTimeout, "context cancelled before bundling", err)
 	}
 
-	// Component preflight: run the registry-declared component validations
-	// before handing back component values. DefaultBundler.Make runs the
-	// same gate (runComponentValidations) before writing a bundle; without
-	// it here the SDK path would return values the bundle path refuses to
-	// render — e.g. the gpu-operator driver-ownership coherence check
-	// (severity: error) on a recipe resolved from a snapshot that observed
-	// no NVIDIA kernel driver. This path has no bundle-time --set
-	// overrides, so the bundler config is nil; validations that act solely
-	// on bundle-time flags no-op on a nil config.
-	preflightWarnings, preflightErr := validations.RunComponentValidations(ctx, r.internal, nil)
-	for _, warning := range preflightWarnings {
-		slog.Warn(warning, "source", "component-validation")
-	}
-	if preflightErr != nil {
-		return nil, preflightErr
-	}
-
 	// Agree with DefaultBundler.Make (bundler.go), which fails closed with
 	// ErrCodeInvalidRequest whenever there is nothing to deploy — whether
 	// that's every component disabled or a zero-componentRef recipe. An
@@ -1204,15 +1199,42 @@ func (c *Client) BundleComponents(ctx context.Context, r *RecipeResult) ([]Compo
 			"recipe has no enabled components")
 	}
 
-	// Resolve the complete Helm-value inventory before emitting any component.
-	// Accounting ownership spans multiple charts, so validating one chart at a
-	// time could return a partial SDK bundle that DefaultBundler.Make rejects.
+	// Resolve the complete Helm-value inventory ONCE, before any gate runs and
+	// before emitting any component. Two properties depend on this ordering:
+	//
+	//  1. Accounting ownership spans multiple charts, so validating one chart
+	//     at a time could return a partial SDK bundle that
+	//     DefaultBundler.Make rejects.
+	//  2. Read-once coherence (#1873 item A). A LayeredDataProvider re-reads
+	//     external --data files on every call, so a gate that resolves values
+	//     independently can validate one set and let a different set be
+	//     emitted. Pinning the resolved maps onto the recipe result the gates
+	//     see makes the values they examine, by construction, the values
+	//     returned below.
 	helmValues, err := resolveHelmComponentValues(ctx, r)
 	if err != nil {
 		return nil, err
 	}
-	if err := bundler.ValidateAccountingValues(r.internal, helmValues); err != nil {
+	pinned := r.internal.WithResolvedValues(helmValues)
+	if err := bundler.ValidateAccountingValues(pinned, helmValues); err != nil {
 		return nil, err
+	}
+
+	// Component preflight: run the registry-declared component validations
+	// before handing back component values. DefaultBundler.Make runs the
+	// same gate (runComponentValidations) before writing a bundle; without
+	// it here the SDK path would return values the bundle path refuses to
+	// render — e.g. the gpu-operator driver-ownership coherence check
+	// (severity: error) on a recipe resolved from a snapshot that observed
+	// no NVIDIA kernel driver. This path has no bundle-time --set
+	// overrides, so the bundler config is nil; validations that act solely
+	// on bundle-time flags no-op on a nil config.
+	preflightWarnings, preflightErr := validations.RunComponentValidations(ctx, pinned, nil)
+	for _, warning := range preflightWarnings {
+		slog.Warn(warning, "source", "component-validation")
+	}
+	if preflightErr != nil {
+		return nil, preflightErr
 	}
 
 	bundles := make([]ComponentBundle, 0, len(r.Components))
@@ -1345,6 +1367,10 @@ func resolveHelmComponentValues(
 // CollectSnapshot deploys the snapshotter Job to the cluster identified
 // by cfg.Kubeconfig and returns the captured Snapshot.
 //
+// This is the single Job-mode collection path in the tree: `aicr snapshot`
+// and `aicr validate` both run it, so the facade AgentConfig mirror is
+// exercised on every snapshot AICR takes.
+//
 // CollectSnapshot does NOT consult the Client's recipe data provider —
 // the Client is required only to keep the facade surface uniform
 // (every public operation goes through a Client) and to leave room
@@ -1355,6 +1381,54 @@ func resolveHelmComponentValues(
 // cfg.Kubeconfig is the path (or empty for in-cluster). cfg.Namespace,
 // cfg.Image, cfg.ServiceAccountName must be set; other fields fall
 // back to package defaults documented on snapshotter.AgentConfig.
+//
+// # Output and delivery
+//
+// The returned Snapshot carries both the parsed form and, in Snapshot.Raw,
+// the exact bytes the agent emitted. CollectSnapshot does not write them
+// anywhere except when cfg.Output names a ConfigMap (cm://namespace/name),
+// which the Job writes directly. Persisting to a file, stdout, or a Go
+// template is the caller's step — pass Snapshot.Raw to
+// snapshotter.DeliverSnapshot, as `aicr snapshot` does. Delivering Raw rather
+// than re-serializing the parsed snapshot is what keeps the output
+// byte-identical to the agent's when a newer agent image emits fields this
+// binary does not model.
+//
+// # Fail-before-mutate
+//
+// Inputs that can be rejected without contacting the cluster are checked
+// before the Kubernetes client is built, so a rejection never leaves RBAC or
+// a Job behind (with cfg.Cleanup false — the zero value — they would
+// persist). That covers a malformed cfg.Output ConfigMap URI and a non-empty
+// cfg.ClusterConfigPath, both ErrCodeInvalidRequest.
+//
+// # Deliberately outside this method
+//
+// Two snapshot capabilities are NOT reachable through CollectSnapshot, by
+// design, because neither deploys a Job:
+//
+//   - Local (in-pod) collection — the mode the agent container itself runs
+//     under AICR_AGENT_MODE=true, and the dev bypass of the same name. It
+//     runs collectors in-process against the local node instead of deploying
+//     an agent, so it needs a collector.Factory and a serializer.Serializer,
+//     types the semver-stable facade deliberately does not expose. Use
+//     snapshotter.NodeSnapshotter directly, as pkg/cli/snapshot.go does. It
+//     takes no AgentConfig, so leaving it out costs no coverage of the field
+//     mirror: every deployed Job still projects through this method.
+//   - cfg.ClusterConfigPath — an l8k cluster-config.yaml ingested by the
+//     in-pod network collector. The path must resolve inside the pod and the
+//     Job does not mount it, so Job mode rejects a non-empty value with
+//     ErrCodeInvalidRequest. Use cfg.DiscoverNetwork for live discovery from
+//     a Job, or the local mode above to read a host file.
+//
+// # Timeout
+//
+// The operation is bounded by cfg.Timeout + defaults.SnapshotOperationGrace
+// (or defaults.SnapshotOperationTimeout + grace when cfg.Timeout is unset),
+// so a caller passing context.Background() still gets a bounded run. The
+// grace exists because cfg.Timeout budgets Job completion only — deployment
+// and result retrieval sit outside it, and a bare cap would silently shrink
+// the completion budget. A tighter caller deadline always wins.
 //
 // Errors:
 //   - ErrCodeInvalidRequest when the Client is nil, cfg is nil, or
@@ -1392,20 +1466,26 @@ func (c *Client) CollectSnapshot(ctx context.Context, cfg *AgentConfig) (*Snapsh
 	// still gets a bounded operation. Preference order:
 	//   1. cfg.Timeout — caller-controlled, wins when set.
 	//   2. SnapshotOperationTimeout — package default (matches CLISnapshotTimeout).
+	// SnapshotOperationGrace is added on top because the chosen value budgets
+	// Job COMPLETION only: deploy, pool projection, and ConfigMap retrieval
+	// happen outside it, so a bare cap would quietly shrink the caller's
+	// completion budget by however long deployment took.
 	// context.WithTimeout honors the smaller of the parent deadline and
 	// the value supplied here, so callers with a tighter context keep it.
-	cap := cfg.Timeout
-	if cap <= 0 {
-		cap = defaults.SnapshotOperationTimeout
+	budget := cfg.Timeout
+	if budget <= 0 {
+		budget = defaults.SnapshotOperationTimeout
 	}
-	ctx, cancel := context.WithTimeout(ctx, cap)
+	ctx, cancel := context.WithTimeout(ctx, budget+defaults.SnapshotOperationGrace)
 	defer cancel()
 
-	snap, err := snapshotter.DeployAndGetSnapshot(ctx, toInternalAgentConfig(cfg))
+	snap, raw, err := snapshotter.DeployAndCollect(ctx, toInternalAgentConfig(cfg))
 	if err != nil {
 		return nil, err
 	}
-	return fromInternalSnapshot(snap), nil
+	out := fromInternalSnapshot(snap)
+	out.Raw = raw
+	return out, nil
 }
 
 // ValidateState evaluates a resolved recipe against an observed cluster
--- a/pkg/client/v1/aicr_internal_test.go
+++ b/pkg/client/v1/aicr_internal_test.go
@@ -66,6 +66,42 @@ func (b *blockingDataProvider) Source(path string) string {
 	return b.underlying.Source(path)
 }
 
+// mutatingValuesProvider models a LayeredDataProvider whose backing file
+// changes underfoot: reads of watchPath return contents[0], contents[1], ...
+// in order (the last entry repeating once exhausted), and every other path
+// delegates to the embedded FS so the component registry still loads.
+//
+// It exists to make the read-once guarantee observable. An operation that
+// resolves a component's values twice — once to validate, once to emit —
+// sees two DIFFERENT value sets here, which is exactly the window issue
+// #1873 item A describes.
+type mutatingValuesProvider struct {
+	underlying recipe.DataProvider
+	watchPath  string
+	contents   [][]byte
+	reads      atomic.Int64
+}
+
+func (p *mutatingValuesProvider) ReadFile(ctx context.Context, path string) ([]byte, error) {
+	if path != p.watchPath {
+		return p.underlying.ReadFile(ctx, path)
+	}
+	n := p.reads.Add(1)
+	index := int(n) - 1
+	if index >= len(p.contents) {
+		index = len(p.contents) - 1
+	}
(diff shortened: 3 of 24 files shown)


pkg/client/v1/aicr.go
```diff
@@ -1356,6 +1382,45 @@ func resolveHelmComponentValues(
...
+//     an agent, so it needs a collector.Factory and a serializer.Serializer,
+//     types the semver-stable facade deliberately does not expose. Use
+//     snapshotter.NodeSnapshotter directly (as pkg/cli/snapshot.go does),
+//     projecting config with ToInternalAgentConfig so the field mirror is
+//     still the single source.
+//   - cfg.ClusterConfigPath — an l8k cluster-config.yaml ingested by the
+//     in-pod network collector. The path must resolve inside the pod and the
+//     Job does not mount it, so Job mode rejects a non-empty value with
+//     ErrCodeInvalidRequest. Use cfg.DiscoverNetwork for live discovery from
+//     a Job, or the local mode above to read a host file.
```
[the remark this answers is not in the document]
mchmarny: Addressed in 80ae568 — the guard landed one layer down, in `snapshotter.DeployAndCollect`, rather than in `CollectSnapshot`.

The functional concern is real and closed: a non-empty `ClusterConfigPath` is now rejected with `ErrCodeInvalidRequest` as the **first** thing `DeployAndCollect` does, before `getKubeClient` and before any cluster access. `CollectSnapshot` reaches it via `ToInternalAgentConfig(cfg)`, so the documented facade contract holds and the caller gets the error rather than a Job that silently collects less.

Placing it there instead of in the facade covers `NodeSnapshotter` too and keeps a single spelling of the check — the duplicate that used to live in `deployAndWaitForResult` was removed, since it now has one caller.

Covered by `TestCollectSnapshot_RejectsBeforeClusterAccess` and `TestDeployAndCollectRejectsBeforeClusterAccess`. Both assert the *ordering*, not just the rejection: they pass an unusable kubeconfig and require the input error to win, which is what proves nothing was mutated first (with `Cleanup` false — the zero value — RBAC and the Job would otherwise persist).

pkg/cli/snapshot.go
```diff
@@ -120,10 +121,14 @@ type snapshotCmdOptions struct {
...
-func (o *snapshotCmdOptions) toAgentConfig() *snapshotter.AgentConfig {
-	return &snapshotter.AgentConfig{
+// toAgentConfig converts the resolved options into the facade AgentConfig that
+// Client.CollectSnapshot consumes. The facade type — not
+// snapshotter.AgentConfig — is the single spelling the CLI builds, so the
+// facade/internal mirror is exercised on every snapshot run instead of only
+// by SDK consumers. The local-collection path projects this back down via
+// aicr.ToInternalAgentConfig rather than assembling a second copy.
+func (o *snapshotCmdOptions) toAgentConfig() *aicr.AgentConfig {
+	return &aicr.AgentConfig{
```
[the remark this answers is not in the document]
mchmarny: Fixed in daf2a280 — the comment was stale and, worse, it was justifying an export that should not exist.

You are right that the local branch never calls `aicr.ToInternalAgentConfig`: it builds a `collector.Factory` and `serializer.Serializer` from `opts` and ignores `agentCfg`. It also *cannot* use an `AgentConfig` — setting `NodeSnapshotter.AgentConfig` is precisely what selects Job mode.

That made the export itself the real problem: `ToInternalAgentConfig` was exported on the stated grounds that the local path needed it, and it had no caller outside `pkg/client/v1`. An exported facade symbol with no in-tree consumer is exactly the drift #2021 exists to remove, so it is back to unexported `toInternalAgentConfig`, and the stability pin is gone with it. The mirror test lives in the internal `translate_test.go`, so enforcement is unaffected.

The comment now says what the code does: Job mode is the only consumer.

pkg/client/v1/query.go
```diff
@@ -15,30 +15,73 @@
...
+// SelectFromRecipe is the context-less form of SelectFromRecipeWithContext,
+// kept for source compatibility. It derives a defaults.FileReadTimeout-bounded
+// context so hydration's values reads stay bounded, but the caller cannot
+// cancel them.
+//
+// Prefer SelectFromRecipeWithContext wherever a context.Context is available.
+func SelectFromRecipe(r *RecipeResult, selector string) (any, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), defaults.FileReadTimeout)
+	defer cancel()
+	return SelectFromRecipeWithContext(ctx, r, selector)
```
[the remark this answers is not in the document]
mchmarny: Skipping this one — `context.Background()` is correct here.

The guideline forbids `context.Background()` in I/O methods, but its stated remedy is a timeout-bounded context, and the canonical GOOD example in `AGENTS.md` is this exact shape:

```go
func (r *Reader) Read(url string) ([]byte, error) {
    ctx, cancel := context.WithTimeout(context.Background(), r.TotalTimeout)
    defer cancel()
    return r.ReadWithContext(ctx, url)
}
```

`SelectFromRecipe` is that pattern verbatim: a context-less compatibility wrapper that derives a bounded context and delegates to the context-aware form. The repo does the same in eight places, including `recipe.HydrateResult` — the very function this wrapper ends up calling — plus `GetValuesForComponent` and `GetComponentValues`.

`context.TODO()` would also be the wrong signal. It means "the right context is unclear, fill this in later." Here the absence is deliberate and permanent: the API takes no caller context by design, which is why `SelectFromRecipeWithContext` exists and is documented as preferred. `context.TODO()` appears nowhere in the repo as a pattern.

pkg/snapshotter/agent.go
```diff
@@ -413,37 +399,187 @@ func getKubeClient(kubeconfig string) (k8sclient.Interface, error) {
...
+//
+// Every input that can be rejected without contacting the cluster is checked
+// up front — before the Kubernetes client is even built, so a rejection is
+// never masked by a kubeconfig error and never leaves RBAC or a Job behind
+// (with Cleanup false, the zero value, they would persist). That covers a
+// malformed cm:// Output and a Job-mode ClusterConfigPath.
+// (measureWithAgent only delivers bytes), but it is the value
+// aicr.Client.CollectSnapshot hands to SDK callers and to `aicr validate`.
+//
+//nolint:unparam // The *Snapshot return has no intra-package consumer
```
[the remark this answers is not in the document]
mchmarny: Fixed in daf2a280, and the root cause is worth recording.

The fragment was not a typo. Since Go 1.19 `gofmt` treats a directive-shaped comment (`//word:...`, which `//nolint:unparam` matches) as a directive and moves it to the END of the doc block. My rationale was written as continuation lines under the directive, so gofmt hoisted them above it and orphaned them mid-sentence — silently, on the `gofmt -w` I ran to fix an unrelated formatting complaint.

Rewriting the sentence in place would just have been re-broken by the next `gofmt`. So the rationale now lives as a normal doc paragraph before the directive, with the directive reduced to a single self-contained line, and the paragraph notes the gofmt behavior so the next person does not reintroduce it. `gofmt -l` is clean and stays clean.

pkg/snapshotter/agent.go
```diff
@@ -413,37 +399,187 @@ func getKubeClient(kubeconfig string) (k8sclient.Interface, error) {
...
+			// error says "invalid configmap URI", which does not tell the
+			// caller WHICH input was wrong. Naming the field is the point.
+			return "", false, errors.Wrap(errors.ErrCodeInvalidRequest,
+				fmt.Sprintf("invalid ConfigMap output URI %q (expected cm://namespace/name)", config.Output),
+				parseErr)
+		}
+		return config.Output, true, nil
+	}
+	return fmt.Sprintf("%s%s/aicr-snapshot", serializer.ConfigMapURIScheme, config.Namespace), false, nil
+}
```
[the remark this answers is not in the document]
mchmarny: Good catch — fixed in daf2a280, with the guard placed one level up.

The analysis is right: an empty `Namespace` builds `cm:///aicr-snapshot`, which `pod.ParseConfigMapURI` rejects only once it is finally parsed — after `getKubeClient`, RBAC, and the Job. With `Cleanup` false (the SDK default) those persist, cluster-admin binding included. The CLI is insulated because both commands default the flag (`default` for snapshot, `aicr-validation` for validate), so this is squarely the SDK-caller path.

I put the check in `DeployAndCollect` rather than in `agentConfigMapTarget`, because `Namespace` is required in BOTH output modes, not just the internal-ConfigMap one: `deployAndWaitForResult` creates the Job and its RBAC in `config.Namespace` regardless of whether `Output` is a cm:// URI. Guarding only the default branch would still let a `cm://ns/name` caller with an empty `Namespace` reach cluster mutation. It now sits with the other fail-before-mutate checks, which is also where the godoc promises it.

Used `strings.TrimSpace` as suggested. Covered by `TestDeployAndCollectRejectsBeforeClusterAccess` (empty and whitespace-only cases) and `TestCollectSnapshot_RejectsBeforeClusterAccess`, both of which pass an unusable kubeconfig so the assertion proves the ordering, not just the rejection.

lalitadithya (approved):

pkg/snapshotter/agent.go
```diff
@@ -413,37 +399,203 @@ func getKubeClient(kubeconfig string) (k8sclient.Interface, error) {
...
+// agentConfigMapTarget resolves where the agent Job stages its result and
+// whether that ConfigMap is the user's delivery vehicle.
+//
+// The Job always writes to a ConfigMap. When config.Output is a cm:// URI the
+// user asked for that exact ConfigMap, so the Job targets it directly and
+// deliverViaConfigMap is true — which makes a failed AKS-pool-merge rewrite
+// fatal rather than a warning, because the bytes the user will read live
+// there. Any other Output (file, stdout, template, or unset) stages to an
+// internal ConfigMap in config.Namespace that the caller never sees.
+//
```
yuanchen8911, line 510: `agentConfigMapTarget` accepts an explicit `cm://<ns>/<name>` whose namespace differs from `AgentConfig.Namespace`, but the Job's Role and RoleBinding grant ConfigMap access only in `AgentConfig.Namespace` (`pkg/k8s/agent/rbac.go:96-152`), so the agent gets a 403 after the namespace, RBAC, and Job already exist — the same late-failure-with-residue outcome this guard exists to prevent. The test table currently encodes the mismatch as valid (`agent_test.go:497-501`: agent namespace `default`, output `cm://custom-ns/my-snapshot`). Either reject a namespace mismatch at this pre-flight point, or deliberately extend the RBAC to the output namespace — and update the test to pin whichever contract is chosen.

pkg/snapshotter/agent.go
```diff
@@ -413,37 +399,203 @@ func getKubeClient(kubeconfig string) (k8sclient.Interface, error) {
...
-		return nil, errors.Wrap(errors.ErrCodeInternal, "failed to parse snapshot data", err)
+		return nil, nil, errors.Wrap(errors.ErrCodeInternal, "failed to parse snapshot data", err)
 	}
 
 	warnOnGPUPlacementMismatch(&snap)
 
-	return &snap, nil
+	return &snap, snapshotData, nil
+}
+
```
yuanchen8911, line 500: The pre-flight accepts ConfigMap destinations Kubernetes will reject: `pod.ParseConfigMapURI` (`pkg/k8s/pod/configmap.go:27-53`) splits once and only checks non-emptiness, so `cm://default/bad/name` passes with name `bad/name`. The failure then happens inside the agent pod — after the namespace, ServiceAccount, RBAC (including the cluster binding), and Job exist — and with `Cleanup` false (the SDK zero value) they persist, which is the outcome the comment above says this guard prevents. Validating the parsed namespace/name against DNS-1123 (`IsDNS1123Subdomain`/`IsDNS1123Label`) here would close it.

pkg/snapshotter/agent.go
```diff
@@ -733,7 +842,11 @@ func rewriteMergedSnapshotConfigMap(ctx context.Context, uri, kubeconfig string,
 	return nil
 }
 
-func rewriteSnapshotConfigMap(ctx context.Context, uri, kubeconfig string, snapshotData []byte) error {
+// writeSnapshotConfigMap applies snapshot bytes to a cm://namespace/name
+// destination, replacing whatever is there. Shared by DeliverSnapshot (the
+// caller's chosen destination) and the AKS-pool merge rewrite (replacing the
+// pre-merge content the agent Job stored).
```
yuanchen8911, line 848: This `errors.Wrap(errors.ErrCodeInternal, ...)` overwrites structured codes the writer already returns — an invalid kubeconfig surfaces as `ErrCodeInvalidRequest` from `ConfigMapWriter.Serialize` (`pkg/serializer/configmap.go:81-87`) and reaches callers of the exported `DeliverSnapshot` as `ErrCodeInternal`. Per the error-wrapping rule, propagate when the inner error already carries the right code (`errors.PropagateOrWrap` fits).

pkg/snapshotter/agent.go
```diff
@@ -413,37 +399,203 @@ func getKubeClient(kubeconfig string) (k8sclient.Interface, error) {
...
+	// every other destination.
+	Kubeconfig string
+}
+
+// DeliverSnapshot writes captured snapshot bytes to the user's destination:
+// a Go template render when TemplatePath is set, otherwise stdout (Output
+// empty, "-", or the stdout URI), a ConfigMap (cm://namespace/name), or a
+// file.
+//
+// data must be the RAW bytes from DeployAndCollect, not a re-serialization of
```
yuanchen8911, line 557: Same class inside `deliverWithTemplate` (below, pre-existing lines): `TemplateWriter.Serialize` returns `ErrCodeNotFound` for a missing template and `ErrCodeInvalidRequest` for a malformed one (`pkg/serializer/template.go:80-95`), and every wrap in that function replaces them with `ErrCodeInternal`. On this newly exported path a caller's bad template looks like an internal failure.

pkg/snapshotter/agent.go
```diff
@@ -413,37 +399,203 @@ func getKubeClient(kubeconfig string) (k8sclient.Interface, error) {
...
+// Mirrors the delivery-relevant subset of AgentConfig so callers that already
+// hold one can forward it, and callers that hold only bytes (an SDK consumer
+// with a Snapshot.Raw) can construct one directly.
+type SnapshotDelivery struct {
+	// Output is the destination: empty, "-", or the stdout URI for stdout;
+	// a cm://namespace/name URI for a ConfigMap; any other value is a file
+	// path.
+	Output string
+
+	// TemplatePath, when set, renders the snapshot through a Go template
```
yuanchen8911, line 541: The byte-identical contract holds only for file delivery: the stdout branch appends a trailing newline after the raw bytes, and the `cm://` branch parses the document into `map[string]any` and re-serializes it, so formatting and bytes can differ from `Snapshot.Raw` (unknown fields do survive). Digests or byte comparisons over the advertised exact bytes won't match stdout or ConfigMap content. Either scope the prose to file delivery or carry the raw bytes through the other branches.

pkg/snapshotter/agent.go
```diff
@@ -413,37 +399,203 @@ func getKubeClient(kubeconfig string) (k8sclient.Interface, error) {
...
+// config.Namespace and the caller delivers the returned bytes.
+//
+// # Fail-before-mutate
+//
+// Every input that can be rejected without contacting the cluster is checked
+// up front — before the Kubernetes client is even built, so a rejection is
+// never masked by a kubeconfig error and never leaves RBAC or a Job behind
+// (with Cleanup false, the zero value, they would persist). That covers a
+// malformed cm:// Output, an empty Namespace, and a Job-mode
+// ClusterConfigPath.
```
yuanchen8911, line 431: Two doc issues in this block: (1) the sentence starting "(measureWithAgent only delivers bytes), but it is the value..." lost its lead-in — the fragment has no antecedent. (2) The fail-before-mutate claim reads as covering agent inputs generally, but only the `cm://` output and Job-mode `ClusterConfigPath` are pre-flighted — an invalid `Image`, `JobName`, or `ServiceAccountName` still fails after namespace/RBAC work. That part is pre-existing SDK hardening rather than a defect here; scope the sentence to the two checked inputs or track broader input validation separately.

pkg/cli/snapshot.go
```diff
@@ -120,10 +121,15 @@ type snapshotCmdOptions struct {
...
-// toAgentConfig converts the resolved options into the snapshotter.AgentConfig
-// the snapshotter deploy path expects.
-func (o *snapshotCmdOptions) toAgentConfig() *snapshotter.AgentConfig {
-	return &snapshotter.AgentConfig{
+// toAgentConfig converts the resolved options into the facade AgentConfig that
+// Client.CollectSnapshot consumes. The facade type — not
+// snapshotter.AgentConfig — is the single spelling the CLI builds, so the
+// facade/internal mirror is exercised on every snapshot run instead of only
+// by SDK consumers. Job mode is its only consumer: local (in-pod) collection
+// deploys no Job and builds a collector.Factory and serializer.Serializer from
```
yuanchen8911, line 129: This comment — and the same claim in `Client.CollectSnapshot`'s godoc and `translate.go:28-40` — says the local-collection path projects config via `aicr.ToInternalAgentConfig`, but the local branch builds the collector factory and `NodeSnapshotter` directly from `opts` fields and never constructs a facade `AgentConfig`. The only non-test caller of the export is `CollectSnapshot` itself, so the stated rationale for exporting it is currently unrealized — either wire the local path through the projection or correct the three comments.

pkg/client/v1/types.go
```diff
@@ -77,6 +77,16 @@ type Snapshot struct {
 	Kind       string
 	CapturedAt time.Time
 
+	// Raw is the exact YAML document the collection agent emitted, set by
+	// Client.CollectSnapshot and empty on snapshots obtained any other way
+	// (WrapSnapshot, a hand-constructed Snapshot).
+	//
+	// Persist THESE bytes rather than re-serializing the parsed snapshot: a
+	// newer agent image can emit fields this binary's Snapshot type does not
```
yuanchen8911, line 85: Adding `Raw` here (with `ClusterConfigPath` and `DiscoverNetwork` on `AgentConfig`) leaves `docs/integrator/public-api.md:59-63` stating the opposite — that the full payload is unexported and the network fields are not surfaced — and it now contradicts `go-library.md`, which this PR updated. Since `pkg/client/v1` is the documented-stable surface, those matrix rows should move in this PR per the same-PR documentation rule.

pkg/client/v1/aicr_internal_test.go
```diff
@@ -521,6 +557,113 @@ func TestBundleComponents_RejectsZeroComponentRefs(t *testing.T) {
...
+	}
+}
+
+// TestBundleComponents_PinnedValuesSurviveGateMutation proves the pinned
+// snapshot is not corrupted by a gate that mutates the map it receives.
+// CheckDriverOwnershipCoherence layers bundle-time --set overrides onto the
+// values map it resolves, so serving the gate a shared reference would let
+// those overrides leak into the emitted bundle. The pin hands out deep
+// copies; a second BundleComponents call must therefore see the same values.
+func TestBundleComponents_PinnedValuesSurviveGateMutation(t *testing.T) {
```
yuanchen8911, line 634: This test cannot detect the leak it documents, twice over: it compares two identically-constructed `BundleComponents` runs to each other, so a deterministic gate mutation leaking into both emitted `HelmValues` passes the equality check; and the mutating branch never executes anyway, because `BundleComponents` passes a nil bundler config and `effectiveComponentValues` returns before the override layering when the config is nil (`pkg/bundler/validations/checks.go:562-584`). A pre-gate baseline capture plus a non-nil config exercising the override path would make it real. (The production getter does return deep copies — this test just doesn't prove it.)

pkg/client/v1/aicr.go
```diff
@@ -1392,20 +1466,26 @@ func (c *Client) CollectSnapshot(ctx context.Context, cfg *AgentConfig) (*Snapsh
...
 	// the value supplied here, so callers with a tighter context keep it.
-	cap := cfg.Timeout
-	if cap <= 0 {
-		cap = defaults.SnapshotOperationTimeout
+	budget := cfg.Timeout
+	if budget <= 0 {
+		budget = defaults.SnapshotOperationTimeout
 	}
-	ctx, cancel := context.WithTimeout(ctx, cap)
+	ctx, cancel := context.WithTimeout(ctx, budget+defaults.SnapshotOperationGrace)
```
yuanchen8911, line 1479: `budget + defaults.SnapshotOperationGrace` overflows for a caller-supplied `Timeout` near `math.MaxInt64`, producing a negative `WithTimeout` and an instantly-expired context — an SDK caller intending "effectively unbounded" gets an immediate timeout. Extreme edge case; a clamp before the addition covers it.

pkg/cli/snapshot.go
```diff
@@ -507,59 +513,88 @@ See examples/templates/snapshot-template.md.tmpl for a sample template.
...
+
+			snap, err := client.CollectSnapshot(ctx, agentCfg)
+			if err != nil {
+				return err
+			}
+
+			// Deliver the RAW agent bytes, never a re-serialization of the
+			// parsed snapshot: a newer agent image can emit fields this
+			// binary's Snapshot type does not model, and a typed round trip
+			// would silently drop them.
```
yuanchen8911, line 592: Non-blocking: stdout delivery goes through `DeliverSnapshot`, which writes `os.Stdout` directly with no injection point — pre-existing behavior, but now on an exported API; a writer field on `SnapshotDelivery` would restore the `cmd.Root().Writer` testability convention.

yuanchen8911 (changes requested):
Review against head daf2a2808. Four items below should be fixed before merge (the cross-namespace ConfigMap output, invalid ConfigMap names passing pre-flight, and error-code overwrites in both non-file delivery branches), plus a stable-API doc sync. The rest are non-blocking doc/test notes. The core mechanics verified clean: the pinned values getter hands out deep copies, the outermost-code contract behind the 404 mapping holds, both existing pre-flight checks precede Kubernetes client construction, and the mirror test's shape assertions catch added/dropped/retyped fields.

PR #2097, opened 2026-08-06
fix(evidence): make gang and webhook conformance tests behavioral

### LINKED ISSUE #1529: Conformance evidence: 3 tests pass without exercising the capability they prove
## Summary

Three CNCF AI Conformance evidence artifacts capture tests that **pass but do not actually exercise the capability they claim to prove**. The evidence is internally consistent and was generated from real cluster runs, but the test design is too weak to be conformance-grade. Each needs a stronger test scenario and a re-run to regenerate the snapshot.

These were surfaced during a documentation audit of `docs/conformance/cncf/v1.35/nim-eks/evidence/`. The `.md` files are generated snapshots (`**Generated:** …`), so the fix lives in the test manifests/scripts under `pkg/evidence/scripts/`, not in the docs.

### 1. Cluster autoscaling — fixed-size ASG cannot demonstrate scaling
`docs/conformance/cncf/v1.35/nim-eks/evidence/cluster-autoscaling.md`

The GPU Auto Scaling Group is captured with `MinSize: 2, MaxSize: 2, DesiredCapacity: 2`. With min = max = desired the group is pinned at 2 nodes — it physically cannot scale up (already at max) or down (already at min). The evidence proves an ASG *exists*, not that GPU-aware autoscaling *works*.

**Fix:** provision the GPU ASG with `min < desired < max` (e.g. min 0–1 / max 4). Submit a pending GPU workload that does not fit current capacity → observe a new node join (scale-up); delete it → observe a node removed (scale-down). Capture the node-count transitions.

> **Disposition (2026-08-08): intentional — out of scope.** The fixed-size ASG is deliberate cluster infrastructure configuration, not a test defect; changing it is an infra decision outside this issue. No script change will be made for this finding.

### 2. Gang scheduling — ample capacity never triggers the all-or-nothing barrier
`docs/conformance/cncf/v1.35/nim-eks/evidence/gang-scheduling.md`

A `PodGroup{minMember: 2}` is launched while the cluster has ample free GPUs, so both pods schedule regardless of whether gang semantics exist — an ordinary scheduler would behave identically. The all-or-nothing barrier is never exercised.

**Fix:** add a negative case. With fewer free GPUs than the group requires, confirm **neither** pod schedules (both `Pending`, not one `Running`/one `Pending`); then free capacity and confirm **both** transition to `Running` together. Assert both pods share the same PodGroup. The partial-capacity `Pending` state is the actual proof.

### 3. Robust operator — schema rejection mistaken for webhook validation
`docs/conformance/cncf/v1.35/nim-eks/evidence/robust-operator.md`

The "webhook rejection" test submits `spec: {}` and is rejected with `spec.authSecret: Required value` / `spec.image: Required value`. That is **CRD OpenAPI schema** validation enforced by the apiserver — it would reject `spec: {}` even with no validating webhook installed. It does not prove the validating admission webhook is operational.

**Fix:** submit a CR that is *schema-valid* (all required fields present) but violates a rule only the **validating webhook** enforces (a business-logic constraint), and show it is rejected with a webhook-specific message.

## How to fix

1. Strengthen the test artifacts under `pkg/evidence/scripts/` (gang-scheduling manifest, the autoscaling scenario, the operator webhook test case).
2. Re-run `pkg/evidence/scripts/collect-evidence.sh` against a real EKS/H100 conformance cluster to regenerate the three evidence snapshots.

> **Step 2 disposition (2026-08-08):** the strengthened tests were validated live on a GB200 EKS conformance cluster (two full-suite runs, 8 PASS / 1 SKIP / 0 FAIL), and the regenerated evidence is committed as the new `docs/conformance/cncf/v1.35/trainer-eks-gb200/` set in PR #2097. The live run also exposed and fixed a latent PodGroup-association defect (the pod-group labels never bound pods to the group; the `pod-group-name` annotation is load-bearing). The original `nim-eks` snapshots are **deliberately untouched** — that set is already submitted upstream to the CNCF AI Conformance repo, and overwriting a submitted evidence set is its own problem. An in-place `nim-eks` regeneration (which will exercise the NIM webhook probe live — today's cluster runs Kubeflow Trainer, and the Dynamo probe needs its own run since the operator dispatch selects one operator per invocation, see #2137) requires a cluster with the NIM + agentgateway stack and is tracked as follow-up in #2137; until then the old `nim-eks` gang/robust-operator snapshots reflect the pre-fix weak tests, as documented in the PR.

## Context

Identified alongside the documentation correctness sweep in #1528 (follow-up note: https://github.com/NVIDIA/aicr/pull/1528#issuecomment-4835150636). The docs PR intentionally left these untouched because editing captured evidence snapshots would falsify recorded results.

yuanchen8911:
## Summary

Makes two CNCF conformance evidence tests behavioral: the gang-scheduling test now exercises the all-or-nothing barrier under constrained capacity, and the operator webhook tests submit schema-valid CRs that only the validating admission webhook rejects, with webhook-attributed verdicts.

## Scope

This PR was rescoped on 2026-08-08 to exactly the original findings of #1529. An earlier revision of this PR had grown into a broad rework across the evidence script, Go validators, `tools/cleanup`, CLI, and docs; that content was removed (review comments before the rescope refer to it and no longer apply to this diff).

**In scope — two behavioral test changes, five files, plus the regenerated evidence set:**

- `pkg/evidence/cncf/scripts/collect-evidence.sh` — the gang-scheduling section (two-phase barrier test) and the three operator webhook probes/verdicts (webhook-attributed rejection), including removing the sections' pre-declared "Result: PASS" summary lines.
- `pkg/evidence/cncf/scripts/manifests/gang-scheduling-test.yaml` — node-pinning affinity (`GANG_TEST_NODE` placeholder) for the two-phase test.
- `pkg/evidence/cncf/result_propagation_test.go` — the Dynamo propagation lanes updated to the new webhook protocol, plus regression lanes pinning: webhook-before-SKIP (no operator-attributed denial + zero DGDs → FAIL), the foreign-webhook branch (a policy webhook's denial → FAIL), the schema-shaped-rejection branch (`Required value` → FAIL), the Kubeflow dispatch branch (exact-name grep + HA `readyReplicas >= 1` readiness → PASS), and a full two-phase gang happy path through the real `collect_gang` (stateful occupancy stub, both gates, the two-part gang-evaluation gate, joint completion → PASS).
- `validators/conformance/helpers_test.go` — asserts the load-bearing `pod-group-name` annotation on the Go check's generated pods (a revert to labels-only fails the test).
- `docs/conformance/cncf/index.md` — the gang and webhook rows describe the new behavioral tests, the idle ≥2-GPU-node prerequisite, the load-bearing PodGroup association, and the affirmative scheduler-evaluation gate (docs-in-same-PR rule).
- `validators/conformance/gang_scheduling_check.go` — two-line fix: the Go gang check's pods gain the same load-bearing `pod-group-name` annotation (its labels-only association had the identical latent defect, proven by the live run; the check remains capacity-unconstrained — that deeper limitation is unchanged and noted in the code comment).
- `docs/conformance/cncf/v1.35/trainer-eks-gb200/` — the regenerated evidence set from the full-suite live run (README + 9 artifacts). Deliberately a NEW sibling profile: the upstream-submitted `nim-eks` H100 set is untouched, and this training-profile set carries no PRODUCT.yaml (it is #1529's regenerated evidence, not a second CNCF submission — the training stack deploys no inference gateway, so `ai_inference` records a fail-closed SKIP).

**Explicitly NOT in scope:**

- **#1529 finding 1 (fixed-size GPU ASG):** dispositioned in the issue as intentional infrastructure configuration, not a test defect — no script change.
- **In-place `nim-eks` snapshot regeneration:** the acceptance validation (step 2) ran live on the GB200 trainer cluster and its evidence is committed in this PR as `trainer-eks-gb200/`; regenerating the upstream-submitted `nim-eks` set in place (which will exercise the NIM webhook probe live; the Dynamo probe needs its own run — the operator dispatch selects one operator per invocation, Dynamo > NIM, see #2137's routing constraint) needs a NIM + agentgateway cluster and is follow-up work tracked in #2137.

**Live validation (2026-08-08, GB200 EKS conformance cluster, KAI + kubeflow-trainer):** both sections were run live from this branch. The gang run first **caught a real latent defect** — the manifest's `pod-group.scheduling.run.ai/*` labels never associated the pods to the pre-created PodGroup (KAI's pod-grouper ignores them and auto-creates per-pod groups; the workers scheduled *individually*, sequentially through the one free GPU — the exact #1529 false-pass mechanism, invisible under ample capacity). Fixed with the load-bearing `pod-group-name` annotation (KAI's `PodGroupAnnotationForPod`, verified in v0.14.1 source). With the association fixed: barrier held (both workers unbound, with KAI's explicitly gang-attributed refusal "Resources were found for 1 pods while 2 are required for gang scheduling"), release phase completed both together — **PASS**. The robust-operator section **PASSed** with the exact pinned-name denial (`validator.trainjob.trainer.kubeflow.org`), which also confirms server-side dry-run support and validates the name-first classification ordering (the live denial text contains "Invalid value", which a schema-shape-first check would have misclassified).

The stable evaluation signal was selected from the run's artifacts and is now **verdict-bearing and gang-specific**: barrier credit requires both workers `PodScheduled=False/Unschedulable` AND the named PodGroup's scheduling status carrying KAI's one-of-two gang refusal ("Resources were found for 1 pods while 2 are required for gang scheduling") — closing both the scheduler-outage window and the generic-refusal window (queue quota or transient fit errors clearing at release time). Re-validated live after the two-part gate landed; the committed gang evidence is regenerated from that run.

**Full-suite live run (2026-08-08, same GB200 EKS cluster, all nine sections, this branch @ the validator-fix head):**

| Section | Result |
|---|---|
| DRA Support | PASS |
| Gang Scheduling | **PASS** (barrier held with the affirmative gang-attributed refusal; both workers completed together after release) |
| Secure Accelerator Access | PASS |
| Accelerator Metrics | PASS |
| AI Service Metrics | PASS (trainer path) |
| Inference Gateway | SKIP (agentgateway confirmed absent — training-stack cluster) |
| Robust AI Operator | **PASS** (webhook-attributed rejection: `validator.trainjob.trainer.kubeflow.org`) |
| Pod Autoscaling (HPA) | PASS |
| Cluster Autoscaling | PASS |

Total: 9 — 8 PASS, 1 SKIP, 0 FAIL — **reproduced identically on four full-suite runs**, the latest at the final reviewed head (post review-round fixes: effective-usage occupancy, schedulable-node pick, VIOLATED branch, activeDeadlineSeconds bounds, no-idle evidence capture), including the affirmative gang-attributed barrier refusal on every run. The committed evidence set is regenerated from the latest run. The rewritten sections pass legitimately (barrier exercised, webhook attributed), the untouched sections are unaffected by this PR, and the cluster was restored afterward (standing slurm workload scaled back, no test namespaces left behind).

**Snapshot publication (dispositioned):** the regenerated evidence is committed as the new `v1.35/trainer-eks-gb200/` profile; the upstream-submitted `v1.35/nim-eks/` H100 set is deliberately untouched — its in-place regeneration with the strengthened tests (which will exercise the NIM webhook probe live; the Dynamo probe needs a separate run per #2137's routing constraint) is tracked in #2137. All undraft blockers for THIS PR are cleared.
- **Cleanup/artifact-truthfulness defects** found during this PR's earlier review cycle (Dynamo destructive cleanup, HPA leak-under-PASS, validator delete handling, remaining hardcoded summary): batched in #2133 for a separate small PR.
- The other six evidence sections, `tools/cleanup`, `pkg/cli`, and `pkg/evidence/cncf/collector.go` — untouched. (The Go gang validator and the conformance docs page, originally listed here, were later brought in scope with explicit maintainer approval: the two-line association fix after the live run proved the shared defect, and the docs row under the docs-in-same-PR rule.)

## Motivation / Context

Two evidence artifacts captured tests that pass without exercising the capability they claim to prove (#1529):

- The gang test launched a `PodGroup{minMember: 2}` into ample free capacity — both pods schedule under any scheduler, so the barrier was never exercised.
- The NIM "webhook rejection" probe submitted `spec: {}`, which the CRD OpenAPI schema rejects with no webhook installed at all; the verdicts also accepted any failure text (the NIM grep matched `error`) or PASSed outright with no webhook rejection at all.

Fixes: #1529 — step 1 in full; step 2's acceptance validation done live on the GB200 trainer cluster with the regenerated `trainer-eks-gb200` set committed. The in-place `nim-eks` regeneration (needs a NIM + agentgateway cluster; exercises the NIM webhook probe live — the Dynamo probe needs a separate run per #2137's routing constraint) is follow-up work tracked in #2137 — the upstream-submitted `nim-eks` set is deliberately not overwritten
Related: #2133

## Type of Change

## Component(s) Affected

- [x] Other: CNCF conformance evidence collector (`pkg/evidence/cncf/scripts/`)

## Implementation Notes

**Gang scheduling (two-phase barrier test).** Both workers are pinned to one chosen GPU node via a `GANG_TEST_NODE` placeholder, and associate to the PodGroup through the load-bearing `pod-group-name` annotation. Phase 1 stages a blocker occupying all but one GPU so the two-pod group cannot fit; phase 2 deletes the blocker and requires both pods to complete together. PASS requires both phases.

Barrier credit requires four things, all fail-closed (a failed read is UNOBSERVABLE, never "held"):

- **Neither worker bound** — checked via `spec.nodeName`, not pod phase (Pending includes bound-but-pulling pods).
- **The scheduler's affirmative gang decision** — both workers `PodScheduled=False/Unschedulable` AND the named PodGroup's scheduling status carrying KAI's one-of-two gang refusal ("Resources were found for 1 pods while 2 are required for gang scheduling"). Both signals were selected from the live GB200 run and are verdict-bearing; a down/backlogged scheduler (no condition) or a generic refusal (queue quota, transient fit error) is NOT CREDITED.
- **The exactly-one-free invariant at both ends of the window** — an occupancy gate after the blocker is Running, and a post-window re-check, both computing effective per-pod GPU usage (max of app-container sum and largest init container; DRA ResourceClaim-held GPUs are an accepted, documented limitation).
- **A validly staged node** — the pick takes an idle, Ready, schedulable node with ≥ 2 allocatable GPUs (occupancy checked per candidate; cordoned/NotReady nodes are excluded), after `ensure_fresh_namespace` so stale leftovers cannot inflate occupancy.

The evidence classifies every non-credit outcome distinctly (VIOLATED for a bound member, NOT CREDITED for occupancy drift or a missing gang decision, UNOBSERVABLE for failed reads), and the final verdict references that same classification instead of re-deriving it.

Failure diagnosis and time budget: a blocker that cannot reach Running captures its pod description and names both likely causes (occupied GPUs, cold image pull); the blocker dies instantly on delete (`terminationGracePeriodSeconds: 0`, bounded `--timeout=30s`); release-phase waits draw from one elapsed-aware budget (ending by section-start + 270s, ≤ 90s per pod, warm image cache from the blocker's pull) so a slow success is never killed before its PASS and a failure keeps headroom for diagnostics; worker-1 is read once instead of waited on when worker-0 already failed.

**Operator webhooks (webhook-attributed rejection).** Each probe is schema-valid but violates a rule only the validating webhook enforces: NIM sets `metrics.enabled: true` without a `serviceMonitor` (all CRD-required fields present); Kubeflow references a non-existent runtime; Dynamo keeps its empty spec deliberately — at the pinned CRD version no spec fields are required, so the empty spec reaches admission and only the webhook rejects it. Probes are submitted with `--dry-run=server` (nothing persists, no cleanup). The verdict requires a denial attributed to the **operator's own** webhook by its pinned name (`validator.trainjob.trainer.kubeflow.org`, `vnimservice-v<version>.kb.io`, `vdynamographdeployment.kb.io`) — a denial from a different admission webhook (e.g. a cluster policy webhook such as Gatekeeper, itself an AICR-supported component) is called out and does not count; schema-shaped rejections (`Required value`) are explicitly called out as NOT demonstrating the webhook, and the previous PASS-without-rejection verdict branches are now FAIL. The Dynamo verdict gates on the webhook BEFORE its no-workload SKIP branch — the probe needs only the operator, so an absent DynamoGraphDeployment (the normal state on a stock inference cluster) cannot convert an observed webhook failure into a non-failing SKIP. The Kubeflow controller-readiness read uses `readyReplicas >= 1` instead of a literal `1/1` column match, so an HA-scaled controller is not misread as unready under the now-stricter verdict.

## Testing

```bash
bash -n pkg/evidence/cncf/scripts/collect-evidence.sh   # clean
yamllint pkg/evidence/cncf/scripts/manifests/gang-scheduling-test.yaml   # clean
GOFLAGS=-mod=vendor go test -race ./pkg/evidence/cncf/ -count=1          # ok
```

Go changes: the propagation-test lanes plus the two-line production fix in `validators/conformance/gang_scheduling_check.go` (the load-bearing PodGroup annotation, race-tested and linted). `golangci-lint run -c .golangci.yaml ./pkg/evidence/cncf/...` — 0 issues. Full `make qualify` was not run for this script/manifest/test-lane change; CI runs the complete gate. The behavioral acceptance test (#1529 step 2) has been RUN: full suite live on a GB200 EKS conformance cluster — 8 PASS / 1 SKIP / 0 FAIL (see the Live validation section), with the regenerated evidence committed under `docs/conformance/cncf/v1.35/trainer-eks-gb200/`.

## Risk Assessment

**Rollout notes:** Committed evidence snapshots are unaffected until regenerated. The strengthened gang test requires a GPU node with ≥ 2 allocatable GPUs and briefly stages a blocker pod on it; the webhook probes are server-side dry-run only and persist nothing. Clusters whose operators genuinely lack a working validating webhook will now FAIL the robust-operator section instead of passing — the intended fail-closed behavior.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — scoped to the affected package, see Testing
- [x] Linter passes (`make lint`) — yamllint on the changed manifest; golangci-lint 0 issues on the changed Go test file's package
- [x] I added/updated tests for new functionality — propagation lanes updated to the new webhook protocol plus a regression lane pinning the webhook-before-SKIP rule; the live regeneration run validates the behavioral scenarios end to end
- [x] I updated docs if user-facing behavior changed — no user-facing surface changed


pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -780,7 +784,40 @@ EOF
...
+gpu_node_free() {
+    local node="$1" alloc used free
+    alloc=$(kubectl get node "${node}" -o jsonpath='{.status.allocatable.nvidia\.com/gpu}' 2>/dev/null)
+    used=$(kubectl describe node "${node}" 2>/dev/null | awk '$1=="nvidia.com/gpu"{print $2}' | tail -1)
+    case "${alloc}" in ''|*[!0-9]*) alloc=0 ;; esac
+    case "${used}" in ''|*[!0-9]*) used=0 ;; esac
+    free=$((alloc - used))
+    [ "${free}" -lt 0 ] && free=0
+    echo "${free}"
+}
```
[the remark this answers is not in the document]
yuanchen8911: The fail-closed half of this is done as of 27bd4e434: gpu_node_free/gpu_total_free return an unreadable sentinel on any kubectl failure (never 0), and every consumer fails the phase closed with an API-failure diagnosis; the node listing itself is also rc-checked.

Deriving the used count from Pod API data instead of the describe table is a deliberate trade-off documented at the helper: the allocated-resources rows have been format-stable for years, and summing per-pod requests via the API is substantially more code for the same number. Happy to revisit in a follow-up if the describe format ever moves.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -817,10 +860,162 @@ EOF
...
+    elif [ -n "${node0}" ] && [ -n "${node1}" ]; then
+        echo "WARNING: both pods were scheduled during the constrained phase — capacity was not actually constrained (free-GPU accounting mismatch); the negative case is inconclusive." >> "${EVIDENCE_FILE}"
+    elif [ -n "${node0}" ] || [ -n "${node1}" ]; then
+        echo "WARNING: exactly one pod was scheduled during the constrained phase — the all-or-nothing barrier was violated." >> "${EVIDENCE_FILE}"
+    elif [ "${free_end}" -ne 1 ]; then
+        echo "WARNING: ${free_end} GPU(s) free at the end of the window (expected 1) — a concurrent workload changed capacity mid-observation; the negative case is inconclusive." >> "${EVIDENCE_FILE}"
+    else
+        negative_ok=true
+        echo "Neither pod was scheduled while exactly one GPU stayed free — insufficient for the whole group, so the all-or-nothing barrier held." >> "${EVIDENCE_FILE}"
+    fi
```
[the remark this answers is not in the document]
yuanchen8911: Declining this one: the 60s hold is an internal proof window, not an operational wait — a shorter window weakens the negative proof and a longer one only delays the run. The scale-test waits got env knobs because node boot times genuinely vary per cluster, and even those needed clamping machinery to keep them inside the section budget; adding a knob here would buy configurability nobody has asked for at the cost of another budget interaction.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2801,12 +3039,200 @@ EOF
...
+            elapsed=$((elapsed + 30))
+        done
+        capture "GPU nodes after scale-down window" kubectl get nodes -l nvidia.com/gpu.present=true
+        echo "" >> "${EVIDENCE_FILE}"
+        if [ "${scale_down_ok}" = "true" ]; then
+            echo "Scale-down observed after ${elapsed}s: GPU node count returned to baseline (${baseline_count})." >> "${EVIDENCE_FILE}"
+        else
+            echo "GPU node count did not return to baseline within ${SCALE_DOWN_TIMEOUT}s." >> "${EVIDENCE_FILE}"
+        fi
+    fi
```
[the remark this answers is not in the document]
yuanchen8911: Addressed differently as of 27bd4e434: scale-down success now requires BOTH the GPU node count and the mechanism's capacity metric (ASG DesiredCapacity, or the Karpenter NodeClaim count) to return to baseline, with every polled read rc-checked. Pinning the specific new node was considered and rejected: consolidation may legitimately remove a different (older) node than the one that joined, which would false-FAIL a healthy scale-down. The metric+count pair proves the capacity left without assuming which node carries it.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -817,10 +830,137 @@ EOF
...
+    cat >> "${EVIDENCE_FILE}" <<'EOF'
+
+## Phase 2 — Positive Case (Capacity Freed)
+
+The blocker pods are deleted. With capacity for the whole group available, both
+gang pods must now be scheduled together and complete.
+EOF
+    log_info "Deleting capacity blockers..."
+    capture "Delete capacity blockers" kubectl delete pods -n gang-scheduling-test -l app=gang-capacity-blocker --wait=true --timeout=60s
+
```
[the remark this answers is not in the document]
yuanchen8911: Addressed as of fe719f95c: the result-propagation harness (which executes the shipped script in a subprocess with stubbed cluster functions) now covers the new evidence gates directly — gateway controller present-but-unready FAILs, gateway Deployment read failure FAILs (never SKIP), a foreign hpa-test namespace is refused fail-closed, and a GKE autoscaler not covering the GPU node pool FAILs. These join the existing lanes for operator discovery (absent vs probe-failed), the DGD query failure, provider-read failure, and the tools/cleanup suite's fifteen ownership/exit-code lanes. The live-wait phases (gang hold window, autoscale scale-up/down) need a time-compressed harness to be testable offline; they get their first end-to-end exercise at evidence regeneration on the conformance cluster (issue 1529 step 2).

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2354,7 +2512,8 @@ INVALID_CR
     local running_pods
     running_pods=$(kubectl get pods -n "${nim_ns}" -l app.kubernetes.io/managed-by=k8s-nim-operator --no-headers 2>/dev/null | grep -c "Running" || true)
     local webhook_ok
-    webhook_ok=$(echo "${webhook_result}" | grep -ci "denied\|forbidden\|invalid\|error" || true)
+    # Only webhook-attributable rejections count (see Webhook Rejection Test above).
+    webhook_ok=$(echo "${webhook_result}" | grep -ci "admission webhook.*denied the request" || true)
```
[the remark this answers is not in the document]
yuanchen8911: Done as of fe719f95c: all three operator verdicts (Dynamo, NIM, Kubeflow) require a webhook-attributable denial for PASS — a run whose webhook is absent or not demonstrated FAILs. Attribution uses the shared classify_webhook_denial ladder keyed on exit status first, and the denial must name the operator's own webhook (exact names from the pinned charts; a denial from an unrelated policy webhook is reported and does not certify).

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2462,10 +2621,13 @@ INVALID_CR
...
-    # Check if webhook rejected it
+    # Check if the rejection is attributable to the admission webhook — CRD
+    # schema or RBAC rejections must not masquerade as webhook validation (#1529).
     echo "" >> "${EVIDENCE_FILE}"
-    if echo "${webhook_result}" | grep -qi "denied\|forbidden\|invalid\|error"; then
-        echo "Webhook correctly rejected the invalid resource." >> "${EVIDENCE_FILE}"
+    if echo "${webhook_result}" | grep -qi "admission webhook.*denied the request"; then
+        echo "Validating webhook rejected the invalid resource — webhook is operational." >> "${EVIDENCE_FILE}"
+    elif echo "${webhook_result}" | grep -qi "denied\|forbidden\|invalid\|error"; then
+        echo "WARNING: request was rejected, but not by the admission webhook (schema/RBAC-level rejection) — webhook operation not demonstrated." >> "${EVIDENCE_FILE}"
```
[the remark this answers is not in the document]
yuanchen8911: Verified against the pinned dynamo-operator 1.2.1 chart and refuted: the v1beta1 DynamoGraphDeployment CRD declares no required spec fields and no x-kubernetes-validations rules, so spec: {} passes CRD validation and reaches admission, where the operator's validating webhook (vdynamographdeployment.kb.io, default matchPolicy Equivalent) rejects it with 'spec.services must have at least one service'. The probe is therefore webhook-only as required. The verification rationale is documented in the script and the emitted evidence text as of fe719f95c.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2485,14 +2721,15 @@ INVALID_CR
...
+    # Only webhook-attributable rejections count (see Webhook Rejection Test above).
+    webhook_ok=$(echo "${webhook_result}" | grep -ci "admission webhook.*denied the request" || true)
 
     if [ "${dgd_query_ok}" != "true" ]; then
         echo "**Result: FAIL** — DynamoGraphDeployment query failed (fail closed); rerun against a reachable API server." >> "${EVIDENCE_FILE}"
     elif [ "${dgd_count}" -gt 0 ] && [ "${running_pods}" -gt 0 ] && [ "${webhook_ok}" -gt 0 ]; then
         echo "**Result: PASS** — Dynamo operator running, webhooks operational (rejection verified), CRDs registered, DynamoGraphDeployment reconciled with ${running_pods} healthy workload pod(s)." >> "${EVIDENCE_FILE}"
     elif [ "${dgd_count}" -gt 0 ] && [ "${running_pods}" -gt 0 ]; then
-        echo "**Result: PASS** — Dynamo operator running, CRDs registered, DynamoGraphDeployment reconciled with ${running_pods} healthy workload pod(s)." >> "${EVIDENCE_FILE}"
+        echo "**Result: FAIL** — Dynamo operator running and DynamoGraphDeployment reconciled, but a webhook-attributable rejection was not demonstrated; operational webhooks are required (fail closed)." >> "${EVIDENCE_FILE}"
```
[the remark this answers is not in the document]
yuanchen8911: Same as the earlier thread on this probe: verified against the pinned dynamo-operator 1.2.1 chart, the v1beta1 CRD has no required spec fields and no CEL rules, so the empty spec reaches admission and is rejected only by the operator's validating webhook — the classification lands on operator-denied, not generic. Rationale documented in-script as of fe719f95c.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2801,10 +2979,14 @@ EOF
...
+    # MaxSize). A pinned group (min == max == desired) cannot scale in either
+    # direction, so it must not certify autoscaling (#1529).
     echo "" >> "${EVIDENCE_FILE}"
-    if [ "${asg_verified}" = "true" ]; then
-        echo "**Result: PASS** — EKS cluster with GPU nodes managed by Auto Scaling Group, ASG configuration verified via AWS API. Evidence is configuration-level; a live scale event is not triggered to avoid disrupting the cluster." >> "${EVIDENCE_FILE}"
+    if [ "${asg_verified}" = "true" ] && [ -n "${asg_min}" ] && [ -n "${asg_max}" ] && [ "${asg_max}" -gt "${asg_min}" ]; then
+        echo "**Result: PASS** — EKS cluster with GPU nodes managed by Auto Scaling Group with scale headroom (MinSize=${asg_min} < MaxSize=${asg_max}), verified via AWS API. Evidence is configuration-level; a live scale event is not triggered to avoid disrupting the cluster." >> "${EVIDENCE_FILE}"
+    elif [ "${asg_verified}" = "true" ]; then
+        echo "**Result: FAIL** — GPU Auto Scaling Group is pinned (MinSize=${asg_min:-unknown}, MaxSize=${asg_max:-unknown}) and cannot scale up or down. Provision the GPU node group with MinSize < MaxSize and re-run." >> "${EVIDENCE_FILE}"
```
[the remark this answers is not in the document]
yuanchen8911: Done as of fe719f95c: MinSize < MaxSize is only a pre-gate. PASS requires the live attributed transition — blocker pods pin every free GPU, the trigger pod must be observed Unschedulable, the mechanism must add capacity with the trigger scheduled on a non-baseline node bound to it (the verified ASG's DesiredCapacity rise + instance-to-ASG ownership for Cluster Autoscaler; a post-baseline NodeClaim owning the trigger's node for Karpenter), and the added capacity must drain back to baseline after the load is deleted.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -817,10 +860,162 @@ EOF
...
+    capture "Capacity blocker pods" kubectl get pods -n gang-scheduling-test -l app=gang-capacity-blocker -o wide
+
+    if ! kubectl wait --for=condition=Ready pod -l app=gang-capacity-blocker \
+        -n gang-scheduling-test --timeout="${DEPLOY_TIMEOUT}s" &>/dev/null; then
+        capture "Blocker pod status (not Ready)" kubectl get pods -n gang-scheduling-test -o wide
+        echo "" >> "${EVIDENCE_FILE}"
+        echo "**Result: FAIL** — capacity blocker pods did not become Ready; free-GPU capacity could not be constrained (fail closed). Check for pre-existing GPU workloads and re-run." >> "${EVIDENCE_FILE}"
+        cleanup_ns gang-scheduling-test
+        return
+    fi
```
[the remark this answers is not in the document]
yuanchen8911: Done as of fe719f95c: every verdict-feeding kubectl apply runs through the rc-checked capture helper and fails the section closed with the apply error recorded in the evidence — gang blockers and test manifest, the autoscale load, and the HPA test manifest.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2801,12 +3039,200 @@ EOF
...
+        autoscaler="cluster-autoscaler"
+    elif kubectl get deploy -n karpenter karpenter &>/dev/null; then
+        autoscaler="karpenter"
+    fi
+    echo "" >> "${EVIDENCE_FILE}"
+    if [ -z "${autoscaler}" ]; then
+        echo "**Result: FAIL** — scale headroom exists (${headroom_desc}) but no autoscaler controller (Cluster Autoscaler or Karpenter) was found, so nothing scales the group on pending pods. Install an autoscaler and re-run." >> "${EVIDENCE_FILE}"
+        return
+    fi
+    echo "Autoscaler controller detected: ${autoscaler}" >> "${EVIDENCE_FILE}"
```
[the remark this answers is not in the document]
yuanchen8911: Done as of fe719f95c: controller detection is label-based across all namespaces (app.kubernetes.io/name=karpenter, with app=karpenter as legacy fallback), mirroring the Go validator's finder; the name substring is only a final fallback. Detection is tri-state — a failed Deployment or CRD listing is UNVERIFIABLE and fails closed, never read as absence or a fall-through to the Cluster Autoscaler path.

docs/conformance/cncf/index.md
```diff
@@ -112,9 +112,9 @@ recorded section failure is no longer masked as overall success.
...
+| **Gang scheduling test** | Deploys PodGroup, verifies co-scheduling | Two-phase: under constrained capacity neither pod may schedule (all-or-nothing barrier), then capacity is freed and both must run together + worker logs |
 | **HPA autoscaling** | Metrics API + scale-up validation | CUDA GPU stress test + scale-up |
 | **Metrics** | Custom metrics API data-path verification | DCGM exporter + Prometheus queries |
 | **Gateway** | Condition verification (Accepted, Programmed) | Same |
-| **Webhook test** | Rejection test with invalid CR | Same |
-| **Cluster autoscaling** | Cloud node group validation | Cloud-provider autoscaler API |
+| **Webhook test** | Rejection test with invalid CR | PASS requires a rejection attributable to the admission webhook in every operator path; the NIM probe is a schema-valid CR violating webhook-only rules |
+| **Cluster autoscaling** | Cloud node group validation | EKS: scale-headroom gate (pinned or unverifiable min/max fails) + live scale test — pending GPU pods must trigger an observed node join and removal (extends runtime by up to ~45 min); GKE: node-pool autoscaler configuration evidence |
```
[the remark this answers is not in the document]
yuanchen8911: Done as of fe719f95c: the comparison table's Speed row states the ~5-10 minutes plus the up-to-~60-minute EKS live-scale extension (node boot, scale-down window, rerun settle wait, load staging), matching the note above the table.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2801,12 +3060,208 @@ EOF
...
+    echo "" >> "${EVIDENCE_FILE}"
+    if [ "${scale_up_ok}" = "true" ]; then
+        echo "Scale-up observed after ${elapsed}s: new GPU node \`${new_node}\` joined (baseline ${baseline_count} nodes)." >> "${EVIDENCE_FILE}"
+    else
+        echo "No new GPU node joined within ${SCALE_UP_TIMEOUT}s." >> "${EVIDENCE_FILE}"
+    fi
+
+    # Scale-down: delete the load; the autoscaler must drop back to baseline.
+    log_info "Deleting autoscale load and waiting for scale-down (up to ${SCALE_DOWN_TIMEOUT}s)..."
+    capture "Delete autoscale load" kubectl delete namespace "${ns}" --ignore-not-found --timeout=120s
```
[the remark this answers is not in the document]
yuanchen8911: Addressed by design as of fe719f95c: the load pods are deleted first (a test step that runs even under NO_CLEANUP) and scale-down is observed against per-mechanism criteria; the namespace itself is removed only after a full PASS. On any non-PASS outcome it is deliberately kept as the rerun marker that arms the next run's capacity-settle window.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -782,6 +782,14 @@ EOF
 
 # --- Section 2: Gang Scheduling ---
 collect_gang() {
```
njhensley, line 784: 🟠 **Major** — The entire two-phase `collect_gang()` rewrite (barrier gate, occupancy gates, `gang_budget_remaining`, the two-part gang-evaluation gate, and the VIOLATED/NOT-CREDITED/occupancy-changed/no-gang-decision/UNOBSERVABLE classification + its mirror in the final verdict) has no automated coverage. `result_propagation_test.go` has **no `gang` SECTION mode** (grep: 0 hits), and `collector_test.go` only stubs `runSectionFn` with a canned error — the shell is never sourced/run. This is the PR's headline behavioral logic, and it's fail-closed safety code resting on a single manual GB200 run. The same harness that drives the operator webhook lanes via a stub `kubectl()` could cover the classification table cheaply (VIOLATED when a worker binds, occupancy-changed, generic-vs-gang-attributed no-gang-decision, UNOBSERVABLE on a failed read, and the barrier-held PASS). A future edit that flips a gate polarity or mis-parses the awk would ship a false PASS with nothing to catch it.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2210,34 +2591,39 @@ INVALID_CR
...
-        echo "Webhook correctly rejected the invalid resource." >> "${EVIDENCE_FILE}"
-    elif echo "${webhook_result}" | grep -qi "cannot create resource\|unauthorized"; then
-        echo "WARNING: Rejection was from RBAC, not the admission webhook." >> "${EVIDENCE_FILE}"
-    elif echo "${webhook_result}" | grep -qi "denied\|forbidden\|invalid"; then
-        echo "Webhook rejected the invalid resource (unconfirmed source)." >> "${EVIDENCE_FILE}"
+    # webhook_ok=1 ONLY for a webhook-attributed denial. A schema-shaped
+    # rejection ("Required value"/"Invalid value") or any other failure
+    # (RBAC, transport) does not demonstrate the webhook.
+    local webhook_ok=0
+    if echo "${webhook_result}" | grep -qE 'admission webhook "validator\.trainjob\.trainer\.kubeflow\.org" denied the request'; then
```
njhensley, line 2612: 🟠 **Major** — All three operator probes got the same rewrite (exact-name webhook grep, foreign-webhook discrimination, `Required value|Invalid value` schema branch, PASS→FAIL flip). But `collect_operator` selects Dynamo first (this line), and every new `operator-*` test lane stubs the Dynamo deploy present, so all three route to `collect_operator_dynamo`. NIM's `vnimservice-v[a-z0-9]+\.kb\.io` grep (:2745), Kubeflow's `validator.trainjob.trainer.kubeflow.org` grep (:2598) **and its new `readyReplicas >= 1` controller-ready logic** (:2619), plus both foreign-webhook branches, are entirely untested. A typo in the NIM/Kubeflow regex flips every healthy such cluster to a false FAIL; a loosened match silently restores the false-PASS bug this PR exists to fix — the Dynamo lanes stay green either way. Add `operator-nim-*` / `operator-kubeflow-*` lanes whose stub returns the matching deploy so the router picks that collector.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -803,44 +811,411 @@ EOF
...
+    barrier_sched0=$(kubectl get pod gang-worker-0 -n gang-scheduling-test -o jsonpath='{range .status.conditions[?(@.type=="PodScheduled")]}{.status}/{.reason}{end}' 2>/dev/null) || barrier_observed=false
+    barrier_sched1=$(kubectl get pod gang-worker-1 -n gang-scheduling-test -o jsonpath='{range .status.conditions[?(@.type=="PodScheduled")]}{.status}/{.reason}{end}' 2>/dev/null) || barrier_observed=false
+    barrier_group_refusal=$(kubectl get podgroup gang-test-group -n gang-scheduling-test -o jsonpath='{range .status.schedulingConditions[*]}{.message}{"\n"}{end}' 2>/dev/null) || barrier_observed=false
+    # One match boolean, shared by the gate and the diagnostic below: the
+    # pods and the PodGroup are read sequentially (no consistent snapshot),
+    # so a gang-attributed group refusal alongside an incomplete worker
+    # condition is a real state — the diagnostic must not relabel it as
+    # "not gang-attributed".
+    local barrier_group_matched=false
+    if printf '%s' "${barrier_group_refusal}" | grep -q "for 1 pods while 2 are required for gang scheduling"; then
```
njhensley, line 1090: 🟡 **Minor** — Crediting the barrier requires the literal substring `for 1 pods while 2 are required for gang scheduling` (this line) AND both workers' `PodScheduled` exactly `"False/Unschedulable"` (:1079). Both are KAI-version-specific (validated only against KAI v0.14.x). A wording or reason-string change in a future KAI release makes the main behavioral gang section FAIL on every otherwise-healthy run. This is the acceptable fail-closed direction (never a false PASS), but it's high operational-surprise after a dependency bump. Consider also accepting the structured group `schedulingConditions[].reason`/`reasons[]` as a version-resilient secondary signal (still ANDed with the numeric check), or emit a distinct "unrecognized KAI message format" diagnostic so an upgrade is diagnosable.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2320,46 +2707,64 @@ EOF
...
     echo '```' >> "${EVIDENCE_FILE}"
 
     echo "" >> "${EVIDENCE_FILE}"
-    if echo "${webhook_result}" | grep -qi "denied\|forbidden\|invalid\|error"; then
-        echo "Webhook correctly rejected the invalid resource." >> "${EVIDENCE_FILE}"
+    # webhook_ok=1 ONLY for a webhook-attributed denial: the old
+    # `denied|forbidden|invalid|error` grep certified the webhook on ANY
+    # failure output (schema, RBAC, transport).
+    local webhook_ok=0
+    if echo "${webhook_result}" | grep -qE 'admission webhook "vnimservice-v[a-z0-9]+\.kb\.io" denied the request'; then
```
njhensley, line 2759: 🟡 **Minor** — The NIM (`vnimservice-v<ver>.kb.io`, this line) and Dynamo (`vdynamographdeployment.kb.io`, :2881) webhook-name greps and their probe premises (NIM: `metrics.enabled: true` without `serviceMonitor` is webhook-only; Dynamo: an empty `spec: {}` passes the pinned v1beta1 CRD schema and reaches admission) were **not run live** — only the Kubeflow-trainer probe was (disclosed in the PR body; follow-up #2137). If any premise is wrong at the pinned operator/CRD version (e.g. the CRD later declares required fields → `spec: {}` rejected by schema, or the webhook entry name differs), the section FAILs a genuinely-healthy operator. Direction is fail-closed (the exact-name grep can't be satisfied by a schema rejection, so no false PASS), and no misleading NIM/Dynamo evidence is committed. Treat both verdicts as provisional until validated on live NIM + Dynamo clusters.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2210,34 +2591,39 @@ INVALID_CR
...
+    # webhook_ok=1 ONLY for a webhook-attributed denial. A schema-shaped
+    # rejection ("Required value"/"Invalid value") or any other failure
+    # (RBAC, transport) does not demonstrate the webhook.
+    local webhook_ok=0
+    if echo "${webhook_result}" | grep -qE 'admission webhook "validator\.trainjob\.trainer\.kubeflow\.org" denied the request'; then
+        webhook_ok=1
+        echo "The operator's validating admission webhook (validator.trainjob.trainer.kubeflow.org) rejected the resource." >> "${EVIDENCE_FILE}"
+    elif echo "${webhook_result}" | grep -qi "admission webhook" && echo "${webhook_result}" | grep -qi "denied the request"; then
+        echo "WARNING: the rejection came from a DIFFERENT admission webhook (e.g. a cluster policy webhook), not the operator's (validator.trainjob.trainer.kubeflow.org) — this does not demonstrate the operator's webhook." >> "${EVIDENCE_FILE}"
+    elif echo "${webhook_result}" | grep -q "Required value\|Invalid value"; then
```
njhensley, line 2617: 🟡 **Minor** — The `elif ... grep -q "Required value\|Invalid value"` branch (Kubeflow here; NIM :2750; Dynamo :2886) is the whole rationale for changing the NIM probe payload (from `spec: {}`, which the CRD schema rejects, to a schema-valid spec only the webhook rejects). No harness lane feeds a schema-shaped message to confirm it lands in this WARNING/FAIL branch rather than being mistaken for a webhook denial. One `operator-dynamo-schema-reject` lane echoing a `... Required value` message on `--dry-run=server` and expecting FAIL would cover the exact regression the payload change guards against.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -803,44 +811,411 @@ EOF
...
+  terminationGracePeriodSeconds: 0
+  securityContext:
+    runAsNonRoot: false
+    seccompProfile:
+      type: RuntimeDefault
+  tolerations:
+    - operator: Exists
+  containers:
+    - name: blocker
+      image: nvidia/cuda:12.9.0-base-ubuntu24.04
```
njhensley, line 982: 🟡 **Minor** — The blocker pulls `nvidia/cuda:12.9.0-base-ubuntu24.04` (a mutable tag, not digest-pinned or mirrored) from Docker Hub cold on the chosen node, bounded by the 24×5s Running wait. Docker Hub anonymous rate-limits or a slow ~200MB pull can exceed 120s → blocker never Running → the FAIL path fires (it does clean up and names cold-pull as a cause, which is good). Workers correctly reuse the warm cache. Pin by digest and/or pull through an internal registry mirror; optionally make the blocker wait elapsed-budgeted.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -803,44 +811,411 @@ EOF
...
+            gang_pick_read_err=true
+            log_info "Gang node candidate ${_node}: occupancy unverifiable (pod list failed) — skipping"
+            continue
+        fi
+        if [ "${_used}" -eq 0 ]; then
+            gang_node="${_node}"
+            gang_gpus="${_gpus}"
+            break
+        fi
+        log_info "Gang node candidate ${_node}: ${_used} GPU(s) already in use — skipping"
```
njhensley, line 924: 🟡 **Minor** — The per-candidate skip reasons (cordoned/NotReady/occupied, :905/:909/:916/:924) and the "no idle node" verdict use `log_info` → stdout only; the durable evidence `.md` (what an operator reads without cluster access) gets just the generic FAIL line, with no capture of which nodes were rejected or their occupancy. Contrast the blocker-not-Running and occupancy-mismatch paths, which do `capture` a describe/pod list. Before the no-idle-node return, `capture` a node/GPU-usage summary (or `kubectl get pods -A --field-selector spec.nodeName=<candidate>`) for the rejected candidates.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -803,44 +811,411 @@ EOF
...
+    # SUMS them, so a GPU-requesting sidecar would undercount; no supported
+    # component ships one.)
+    gang_node_gpu_used() {
+        local out
+        if ! out=$(kubectl get pods -A --field-selector "spec.nodeName=$1" -o jsonpath='{range .items[*]}{.status.phase}{" A "}{range .spec.containers[*]}{.resources.limits.nvidia\.com/gpu}{" "}{end}{"I "}{range .spec.initContainers[*]}{.resources.limits.nvidia\.com/gpu}{" "}{end}{"\n"}{end}' 2>/dev/null); then
+            echo "error"
+            return
+        fi
+        printf '%s\n' "${out}" | awk '
+            $1 != "Succeeded" && $1 != "Failed" {
```
njhensley, line 854: 🔵 **Nitpick (informational)** — In `gang_node_gpu_used`, an empty `.status.phase` makes the `"A"` sentinel land in `$1` while the field loop starts at `i=2`, so app-container GPUs on that pod are skipped (reproduced: `' A 3 I '` → 0, `'Pending A 3 I '` → 3). In isolation this could undercount occupancy toward a vacuous barrier, but it's **unreachable** (the apiserver always populates `.status.phase` ≥ Pending) and **doubly-guarded** downstream (a zero-free barrier makes KAI emit "0 pods" not "1 pods", failing the exact-message gate at :1076). Harmless today; tidy by starting the loop at `i=1` or treating the first `A` as the app-mode marker regardless of position.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -803,44 +811,411 @@ EOF
...
+        cleanup_ns gang-scheduling-test
+        return
+    fi
+
+    log_info "Deploying gang test (group needs 2 GPUs; 1 free)..."
+    echo "" >> "${EVIDENCE_FILE}"
+    echo "**Apply test manifest (node-pinned)**" >> "${EVIDENCE_FILE}"
+    echo '```' >> "${EVIDENCE_FILE}"
+    echo "\$ sed \"s/GANG_TEST_NODE/${gang_node}/\" manifests/gang-scheduling-test.yaml | kubectl apply -f -" >> "${EVIDENCE_FILE}"
+    sed "s/GANG_TEST_NODE/${gang_node}/" "${SCRIPT_DIR}/manifests/gang-scheduling-test.yaml" | kubectl apply -f - >> "${EVIDENCE_FILE}" 2>&1
```
njhensley, line 1032: 🔵 **Nitpick** — `sed ... | kubectl apply -f - >> "${EVIDENCE_FILE}" 2>&1` ignores the apply status; a failed apply (e.g. PodGroup CRD missing) surfaces only later as "barrier UNOBSERVABLE" when the worker reads fail. Fail-closed (still FAIL), just a less precise diagnostic. Capturing the pipe rc and emitting an explicit "apply failed" verdict would triage faster.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -2320,46 +2707,64 @@ EOF
...
     fi
 
-    # Verdict
+    # Verdict — the webhook-attributed rejection is REQUIRED (this
+    # section's claim is "webhooks operational").
     echo "" >> "${EVIDENCE_FILE}"
     local crd_count
     crd_count=$(kubectl get crds 2>/dev/null | grep -c "apps\.nvidia\.com" || true)
     local running_pods
     running_pods=$(kubectl get pods -n "${nim_ns}" -l app.kubernetes.io/managed-by=k8s-nim-operator --no-headers 2>/dev/null | grep -c "Running" || true)
```
njhensley, line 2776: 🔵 **Nitpick** — `grep -c "Running"` on `kubectl get pods` output counts any line containing the token, not the STATUS column. Safe today only because RFC1123 pod names can't contain a capital `Running`. `awk '$3=="Running"'` (or the right column) is more robust against future column/name changes.

pkg/evidence/cncf/scripts/collect-evidence.sh
```diff
@@ -803,44 +811,411 @@ EOF
...
+            continue
+        fi
+        gang_candidates=$((gang_candidates + 1))
+        _used=$(gang_node_gpu_used "${_node}")
+        if [ "${_used}" = "error" ]; then
+            gang_pick_read_err=true
+            log_info "Gang node candidate ${_node}: occupancy unverifiable (pod list failed) — skipping"
+            continue
+        fi
+        if [ "${_used}" -eq 0 ]; then
```
njhensley, line 919: 🔵 **Nitpick (by design)** — The idle pick demands `_used -eq 0`, so on a busy shared conformance cluster where every >=2-GPU node carries any GPU pod, the section FAILs with a precise "free a GPU node and re-run" diagnosis. This is intentional and well-documented in-code (a partially-used node can't stage the exactly-one-free invariant) — noting for awareness, not a defect.

pkg/evidence/cncf/scripts/manifests/gang-scheduling-test.yaml
```diff
@@ -38,11 +43,29 @@ kind: Pod
...
+  # pod-group-name is the LOAD-BEARING association: KAI's pod-grouper skips a
+  # bare pod carrying this annotation (no auto-created per-pod PodGroup) and
+  # the scheduler joins it to the named PodGroup (verified against KAI
+  # v0.14.1: PodGroupAnnotationForPod in pod_controller.go/pod_info.go). The
+  # pod-group labels alone do NOT associate — the pod-grouper ignores them
+  # and creates per-pod groups, silently degrading the test to individual
+  # scheduling (observed live on the gb200 conformance cluster).
+  annotations:
+    pod-group-name: gang-test-group
   labels:
```
njhensley, line 57: 🔵 **Nitpick** — The manifest's own comment states these labels "do NOT associate" and were the latent #1529 defect; keeping them beside the load-bearing `pod-group-name` annotation is harmless (arguably forward-compat) but a reader could mistake them for contributing. Either drop them or extend the comment to say they're retained deliberately and inert at KAI v0.14.1.

njhensley (commented):
## 🔍 Multi-persona review — `fix(evidence): make gang and webhook conformance tests behavioral`

**Method:** 4 independent persona reviewers (Correctness · Domain/Conformance · Test-coverage · Operability/CI-DX), each finding adversarially confirmed/refuted/re-tiered by a senior meta-reviewer against the resolved code at `4c294189`. **Legend:** 🔴 Blocker · 🟠 Major · 🟡 Minor · 🔵 Nitpick.

### Overall assessment

This PR does exactly what #1529 demands: it converts two conformance tests that "passed" **without exercising the capability they certify** into tests that stage the discriminating condition and gate on affirmative, capability-attributed signals. The gang test now forces the all-or-nothing barrier under constrained capacity and requires KAI's own gang-attributed refusal; the webhook probes are schema-valid-but-webhook-invalid and require a denial named to the operator's own webhook.

**No surviving false-PASS path** was found in gang or any of the three webhook probes. The two false-PASS *candidates* (an awk undercount, a loosened webhook match) are each blocked by an independent downstream guard. Every surviving finding fails in the **safe (fail-closed) direction** — false-FAIL brittleness or a compound, self-healing resource leak — none corrupts a conformance verdict toward over-certification. The committed GB200 evidence is genuine (real node, GB200 `nvidia-smi`, KAI PodGroup with `uid`/`resourceVersion`).

**Recommendation: ✅ Approve with comments.** The two 🟠s are test-coverage gaps, not defects. Suggested gate-before-merge: **T1** (add a gang-barrier harness lane) + **T4** (the one-line annotation assertion) — both cheap in the existing stub-`kubectl` harness. The rest are good follow-ups.

> **Prior-review note:** the `yuanchen8911` "review" comments on this PR are the **author's own** self-annotations, not an independent reviewer. CodeRabbit's review is informational.

### Findings that couldn't be anchored inline (unchanged files / out-of-hunk lines)

**🟡 O1 — Section-timeout SIGKILL can leak the GPU-holding blocker pod** — `pkg/evidence/cncf/collector.go:368`
`exec.CommandContext` sets no `Cancel`/`WaitDelay`/`Setpgid`, so the 300s `EvidenceSectionTimeout` is an **uncatchable SIGKILL of the bash process only**. The blocker pod (`collect-evidence.sh:950`) has `terminationGracePeriodSeconds: 0` but **no `activeDeadlineSeconds`**. Bounded-but-compounding pre-delete waits (stuck-terminating ns ≤90s + cold-pull blocker wait ≤120s + 60s barrier + captures) can cross 300s and leave `gang-capacity-blocker` holding `gang_gpus-1` GPUs (3 on a 4-GPU node) on a shared cluster until the next gang run reclaims it. A bash `trap` can't help (SIGKILL is uncatchable). **Fix:** set `spec.activeDeadlineSeconds` on the blocker/workers so the kubelet reclaims them independent of bash liveness. *(Merges O4; downgraded from 🟠 — compound trigger + self-healing.)*

**🟡 T4 — The new load-bearing annotation is added but not asserted** — `validators/conformance/helpers_test.go:526`
`gang_scheduling_check.go:505` adds the `pod-group-name` annotation — "the LOAD-BEARING association" per its own comment, without which KAI degrades the test to per-pod scheduling (the exact #1529 defect) — but `TestBuildGangTestPodUsesCPUOnlyWorkload` asserts only the **label**. A revert to label-only passes green. Add: `if got := pod.Annotations["pod-group-name"]; got != run.groupName { t.Errorf(...) }`.

**🔵 D5 — Committed evidence header says `Platform: linux/amd64` on an arm64 GB200 cluster** — `collect-evidence.sh:532`
`detect_cluster_info` reads `.items[0].status.nodeInfo.architecture` — an amd64 control-plane/CPU node, not the GB200 (arm64) worker — so `gang-scheduling.md`/`robust-operator.md` headers read `linux/amd64`. Cosmetic; derive Platform from the GPU node (already selected via `-l nvidia.com/gpu.present=true` elsewhere).

### ✅ Confirmed non-issues (examined, refuted / verified safe)

- **No surviving false-PASS path** in gang or any webhook probe. The awk undercount (C1) is blocked by the exact "1 pods/2 required" gate; a loosened webhook match is blocked by the quote-anchored name grep.
- **`--dry-run=server` webhook probes persist nothing** — removing the old `kubectl delete ... --ignore-not-found` cleanup is correct; no CR-leak path remains.
- **Budget math is sound** (`gang_section_start+270` vs 300s; 90s per-pod cap; single-observation floor at remainder 0). Blocker sizing `gang_gpus-1` makes "exactly 1 free" invariant, so the hardcoded "1 pods/2 required" numbers are stable.
- **Exactly-one `**Result:` invariant** holds on every gang path; the parser isn't fooled by `**Barrier phase:**` or dumped YAML.
- **Dynamo webhook-before-SKIP reorder** is correct and regression-tested (`operator-dynamo-no-dgd-no-webhook` → FAIL).

### Summary

|:---:|:---:|:---:|:---:|
| 0 | 2 | 7 | 6 |

*Generated by a multi-persona + adversarial meta-review pass. Personas propose; the meta-reviewer independently re-derives and gates.*

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Addressed at 7b40f221d: result_propagation_test.go now has a gang-barrier-pass lane that drives the real collect_gang end to end through the stub harness - stateful occupancy reads (0 at pick, 3 after the blocker), both occupancy gates, the binding checks, the two-part gang-evaluation gate (PodScheduled False/Unschedulable plus the PodGroup message), release waits, and joint completion, landing in PASS. The happy path and the gate polarities are now regression-covered; lanes for the remaining classification branches (VIOLATED, occupancy-changed, UNOBSERVABLE) are a reasonable follow-up on the same stub pattern.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Partially addressed at 7b40f221d: an operator-kubeflow-webhook-pass lane stubs the Kubeflow deploy present so the router selects the Kubeflow collector - it exercises the validator.trainjob.trainer.kubeflow.org exact-name grep and the readyReplicas >= 1 readiness logic against a 2-replica controller. The schema-rejection branch is covered by a new operator-dynamo-schema-reject lane. NIM lanes are deferred: the NIM probe premises (payload shape, webhook naming) have never been validated against a live NIM cluster, so a stub lane would lock in possibly-wrong expectations; #2137's nim-eks regeneration run settles the premises first, and the lanes follow it.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Agreed on the trade-off - the strings are accepted fail-closed by design, and a maintenance comment now sits beside them tying them to the KAI version pin so a dependency bump updates both together. The structured schedulingConditions reason as a version-resilient secondary signal is a good idea but adds machinery this PR is trying not to grow; noted for follow-up if a KAI bump actually bites.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Disclosed and tracked: these probe premises cannot be validated on this cluster (no NIM deployed), so they are called out in the PR body and owned by #2137, whose nim-eks regeneration run is the only place they can be settled.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Taken at 7b40f221d: an operator-dynamo-schema-reject lane feeds a Required value message through the dry-run path and asserts it lands in the schema-rejection FAIL branch rather than being mistaken for a webhook denial - exactly the regression the payload change guards against.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Declined for this PR: mutable version tags are the convention across every manifest in the repo (recipes, health checks, the other conformance manifests), so a one-off digest pin here would be inconsistent without buying much. If digest pinning is worth doing, it should be a repo-wide policy change - happy to file a follow-up for that.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Taken at 7b40f221d: the no-idle-node FAIL now captures the full per-candidate table (allocatable GPUs, occupancy, cordon/ready state) and the cluster-wide GPU pod placement into the evidence file, so the skip reasons are durable for post-mortem review.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Taken at 7b40f221d: the loop now starts at i=1 so an empty phase cannot shift the sentinel.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Declined as diagnostics-only: if the apply fails the pods never appear and the section lands in the fail-closed UNOBSERVABLE branch, so the verdict cannot be wrong - capturing rc would only enrich the diagnostic message. Keeping the diff minimal here.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Pre-existing line this PR does not touch, and the same defect family as #2133 item 5 where it is already tracked - keeping it out of this diff.

pkg/evidence/cncf/scripts/collect-evidence.sh
yuanchen8911: Agreed, intentional - a partially-used node cannot stage the exactly-one-free invariant. Related improvement at 7b40f221d: when no idle node qualifies, the FAIL now captures the candidate table and GPU pod placement so the free-a-node diagnosis is self-contained in the evidence.

pkg/evidence/cncf/scripts/manifests/gang-scheduling-test.yaml
yuanchen8911: Taken at 7b40f221d: the comment now states the labels are retained deliberately and are inert at KAI v0.14.1 (reporting / forward-compat only), with the annotation identified as the load-bearing association - which is also now asserted in the validator unit tests.

njhensley (approved):
## ✅ Approve — re-review of `de70c65d`

Re-review after the amend (my prior review was against `4c294189`). The author directly addressed the feedback: **7 findings fully addressed, 2 partial/documented**, no net-new issues in the delta. I re-read every changed line and ran `go test ./pkg/evidence/cncf/` (green, incl. the three new lanes).

### Prior-feedback status

| Prior finding | Tier | Disposition | In `de70c65d` |
|---|:---:|:---:|---|
| T1 — `collect_gang()` had zero automated coverage | 🟠 | ✔️ Addressed | `gang-barrier-pass` lane drives the real `collect_gang` end-to-end (idle pick, both occupancy gates, two-part gang gate, release, verdict) → PASS |
| T2 — NIM & Kubeflow webhook paths untested | 🟠 | ◐ Partial | `operator-kubeflow-webhook-pass` covers the Kubeflow grep + `readyReplicas>=1` (HA 2/2); NIM's distinct `vnimservice-v…kb.io` regex still has no lane |
| O1 — section-timeout SIGKILL could leak the GPU blocker | 🟡 | ✔️ Addressed | `activeDeadlineSeconds: 600` on blocker + both workers (kubelet-enforced) |
| T3 — schema-rejection branch untested | 🟡 | ✔️ Addressed | `operator-dynamo-schema-reject` lane → FAIL |
| T4 — load-bearing annotation not asserted | 🟡 | ✔️ Addressed | `helpers_test.go` now asserts `pod.Annotations["pod-group-name"]` |
| O3 — no-idle-node FAIL wrote no evidence diagnostics | 🟡 | ✔️ Addressed | Captures candidate list + `kubectl get pods -A -o wide` before the FAIL |
| C1 — awk undercount on empty `.status.phase` | 🔵 | ✔️ Addressed | Loop starts at `i=1` |
| D4 — non-associating labels looked contributing | 🔵 | ✔️ Addressed | Comment clarifies deliberate retention (inert at v0.14.1) |
| C2 — gang PASS coupled to exact KAI strings | 🟡 | ◐ Documented | Comment: version-coupled, fails closed, update strings with the KAI pin |
| C3 — NIM/Dynamo premises not live-validated | 🟡 | Deferred | Tracked in #2137 (disclosed) |
| O2 · C4 · C5 · D5 | 🟡/🔵 | Open | Minor/nitpick — deferrable |
| D3 — requires a fully idle node | 🔵 | ⊘ N/A | By design |

### Net-new from the delta: none
Verified the awk `i=1` change (no regression), `activeDeadlineSeconds: 600` (safely above the 300s section timeout; workers finish in seconds), and that the new lanes are faithful and isolated (`SCRIPT_DIR = t.TempDir()`, so the stateful `occ-count` can't leak across subtests).

### One non-blocking follow-up
**T2 + C3 together** — the NIM operator webhook path is the only one with *both* zero test coverage *and* no live validation, so a NIM-cluster false-FAIL wouldn't surface until a live run. An `operator-nim-*` lane would close the test half cheaply; the live half is already tracked in #2137. Everything else open is minor/nitpick or tracked.

Both Majors are resolved (T1 fully, T2's Kubeflow path covered). LGTM — approving.

*Multi-persona + adversarial meta-review; delta re-review dispositioned against the resolved code at `de70c65d`.*

PR #2109, opened 2026-08-07
fix(kwok): strict criteria reader in CI classify; document fail-closed contract

framsouza:
## Summary

Address the three unresolved review findings on #2033 (@njhensley): the CI classify step read `.spec.criteria.{service,accelerator}` via a lenient `yq eval '... // \"\"'` helper (so `service: false` silently dropped instead of erroring like the direct path); `profile_status`'s `::error::` echoes lacked `>&2` and were captured & discarded by `$(...)`; and the README's fail-closed-contract paragraph misdescribed CI behavior.

## Motivation / Context

Follow-up to #2033. The parent PR replaced KWOK's hardcoded profile mapping with strict, tag-validated selection so unmapped `(service, accelerator)` combos can't silently fall back to `eks/p5-h100.yaml`. The three findings here are all cases where the *same* silent-swallow class the PR closed was still reachable through side channels — a different field, a lost annotation, and drifted docs that would let a future contributor "fix" the explicit-fail path back into a false green.

Fixes: N/A (unresolved review comments, not an issue)
Related: #2033, #1997

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: KWOK CI classify (`.github/workflows/kwok-recipes.yaml`), shared selector library (`kwok/scripts/lib/profile-select.sh`), user docs (`kwok/README.md`)

## Implementation Notes

**Finding #1 — strict criteria reader in classify.**
Rename the library helper `_read_criteria_field` → public `read_criteria_field` (same tag-based validation: rejects non-`!!str`/`!!null` tags, normalizes `null`/`""`/`"any"` to a caller-chosen default). Add a thin `read_criteria` wrapper in the workflow that calls it with an *empty* default — missing/`any`/null still short-circuit as before, but `service: false` now returns rc=1 with a `::error::` annotation instead of collapsing to `""` and being silently dropped by the downstream `-z` filter. Delete the local `read_overlay_field` helper (unused after the swap) and the now-redundant `"null"`/`"any"` string checks in the two classify loops.

**Finding #2 — README direct/batch/CI mode split.**
The prior "Direct vs batch semantics" paragraph said the CI matrix "skips unmapped recipes with a WARN" — accurate only for implicit `make kwok-test-all`. CI actually DROPs at classify (`::notice`, never dispatched), and an explicit `workflow_dispatch` recipe reaching `run-all-recipes.sh` fails rc=1. Rewrite as two failure classes (no-match rc=2 vs ambiguous/invalid rc=1) × three entry points (direct fail-closed, implicit batch SKIP+WARN, CI DROP + explicit-fail) so the contract is unambiguous. A future contributor "fixing" the explicit-fail path believing CI is meant to skip would reopen the false-green hole this PR chain closed.

**Finding #3 — `profile_status` annotations captured & lost.**
Both `::error::` lines in `profile_status` (workflow L153, L163) lacked `>&2`. The function is always called as `status=\$(profile_status ...)`, so the annotations were captured into the discarded var and never rendered in the GitHub UI — only the raw stderr re-emit survived. Add `>&2` to match the sibling helpers.

Nothing in this PR changes the direct path's behavior or the selector semantics — just wires the same strict contract through the CI entry point and documents the three modes explicitly.

## Testing

```bash
# All three kwok script test suites — 33 assertions, unchanged
$ bash kwok/scripts/lib/sync-budget_test.sh          # 9/9 PASS
$ bash kwok/scripts/lib/profile-select_test.sh       # 22/22 PASS
$ bash kwok/scripts/run-all-recipes_test.sh          # 2/2 PASS

# Lint
$ yamllint .github/workflows/kwok-recipes.yaml       # clean
$ shellcheck kwok/scripts/lib/profile-select.sh      # clean
```

No Go code changed — `make lint` / `make test` unaffected on this diff.

## Risk Assessment

- [x] **Low** — Isolated to KWOK CI scaffolding + doc. The strict reader was already the direct-path contract; this PR just makes the CI path consistent. All existing tests pass unchanged.

**Rollout notes:** N/A. No new flags, no user-visible CLI change. Any overlay with `spec.criteria.{service,accelerator}` set to a non-string type will now fail the discover job with an actionable annotation instead of being silently excluded — desirable.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — N/A for this diff (no Go); all 33 script tests pass
- [x] Linter passes (`make lint`) — yamllint + shellcheck clean
- [x] I added/updated tests for new functionality — N/A (behavior-preserving where existing tests already cover; PR B follow-up will add coverage for the finder-side branches @njhensley flagged)
- [x] I updated docs if user-facing behavior changed (README rewritten per finding #2)


kwok/scripts/lib/profile-select.sh
```diff
@@ -229,3 +242,83 @@ select_profiles() {
...
+# and the standard `|| exit 1` / `|| return 1` guard fires.
+profile_status() {
+    local overlay="${1:-}" profiles_root="${2:-}" svc accel rc=0
+    if [[ -z "${overlay}" || -z "${profiles_root}" ]]; then
+        echo "[ERROR] profile_status: overlay and profiles_root are required" >&2
+        return 1
+    fi
+    svc=$(read_criteria "${overlay}" service) || return 1
+    accel=$(read_criteria "${overlay}" accelerator) || return 1
+    if [[ -z "${svc}" || -z "${accel}" ]]; then
```
mchmarny: **Blocking:** This drops every generic Tier-1 overlay before it can enter Tier 1. These overlays intentionally set `service` but omit `accelerator`; [classify calls `profile_status` before the Tier-1 append](https://github.com/NVIDIA/aicr/blob/c1ead440494ea27ec6283356caeb21e5a3155e4e/.github/workflows/kwok-recipes.yaml#L190-L200). On this head, `resolve_recipe_criteria recipes/overlays/eks-inference.yaml` returns `eks h100` and `select_profiles` finds `eks/system-m7i.yaml:eks/p5-h100.yaml`, but `profile_status` prints `2`, so the workflow drops valid coverage. Treat an empty service as non-testable, but preserve the `h100` default for an empty accelerator (or move this filter after generic classification), and add an `eks-inference`-style regression test.

mchmarny (changes requested):
Requesting changes: 1 merge blocker on c1ead440. All five pre-existing review threads are addressed, and the targeted script tests pass; however, the new placeholder handling drops valid generic Tier-1 recipes, as noted inline. GitHub repository workflows have not run beyond labeling and review checks.

kwok/scripts/lib/profile-select.sh
framsouza: Fixed, thanks for the catch @mchmarny. Reworked `profile_status` so only service is treated as the testability signal; if service is set, we hand off to `resolve_recipe_criteria` so the h100 default still applies for an omitted accelerator

mchmarny (approved):

njhensley (approved):

PR #2127, opened 2026-08-07
fix(docs): fail-closed MDX check for bare '<' + MDX-safe content

### LINKED ISSUE #2050: check-docs-mdx passes on docs Fern cannot parse
## Summary

`make qualify`'s `check-docs-mdx` step reports **"OK: all doc files are MDX-safe"** on a file the Fern docs build refuses to parse. The two checks disagree, so the local gate gives false assurance and the breakage is only discovered after merge — where it currently fails on **every push to main**.

## Reproduction

```console
$ ./tools/check-docs-mdx docs/integrator/
OK: all doc files are MDX-safe
$ echo $?
0
```

Meanwhile `Preview Fern Docs: Comment` fails on the same tree:

```
Failed to parse ../docs/integrator/aks-gpu-setup.md:
Unexpected character `=` (U+003D) before name, expected a character that can start a name
```

The trigger is `docs/integrator/aks-gpu-setup.md:312`:

```markdown
(gate <= 2,000) — Qwen/Qwen3-8B at 256 concurrency per GPU via
```

MDX reads the bare `<` as the start of a JSX tag and then finds `=`, which cannot begin a tag name. (The `>= 50,000` on the line above is harmless — no opening `<`.)

## Why the local check misses it

`tools/check-docs-mdx` implements three pattern families:

- `MDX: non-self-closing void element`
- `MDX: bare <...>`
- `MDX: autolink outside code fence`

None covers a bare `<` followed by a character that cannot start a tag name (`<=`, `<3`, `< 5`). Each rule was added in response to a specific escape:

- #619 — non-self-closing `img` tags
- #759 — bare angle-bracket URL in `cli-reference.md`
- #1147 / #1148 — check did not cover contributor docs

This is the fourth instance of the same failure mode. The pattern list is a denylist of MDX syntax we have already been burned by, so it will keep lagging the parser — which is the repo's documented anti-pattern (a security/correctness gate shaped as a denylist should be inverted to fail closed on the unknown).

## Proposed change

Prefer validating with the real parser over adding rule #4:

1. **Run the actual MDX parser locally.** Parse each doc with the same MDX/remark version Fern uses and fail on any parse error. This closes the class rather than one instance, and eliminates the disagreement by construction. Cost: a Node dependency in the docs lane, so it likely belongs in `make lint`'s docs path rather than in the Go-only fast path.
2. **If a Node dependency is unacceptable**, at minimum add a rule for a bare `<` not followed by `[A-Za-z/!]` outside code fences/spans, and record in `tools/check-docs-mdx` that the check is best-effort — so a green run is not read as "Fern will accept this".

Either way the docs CI failure should be **visible**: `Preview Fern Docs: Comment` has failed on every main SHA for some time (`b7c83bbb`, `be4580a9`, `9a0f6563`, …), which is long enough that an always-red check no longer signals anything.

## Acceptance criteria

- A doc that Fern cannot parse fails the local gate — demonstrated with the `aks-gpu-setup.md` case above (fails before the fix to that file, passes after).
- `Preview Fern Docs` is green on `main`.
- If the check remains pattern-based, its limitations are stated in the script so a passing run is not mistaken for parser-level validation.

## Out of scope

Fixing the one offending line in `aks-gpu-setup.md` — that unblocks the docs build but leaves the detection gap, which is the point of this issue. Worth doing in the same PR, not instead of it.

njhensley:
## Summary

Invert the `<`-family MDX rule in `tools/check-docs-mdx` into a fail-closed allowlist (check 6) so bare `<` sequences Fern's MDX parser rejects (e.g. `<= 2,000`) are caught, and fix the four offending doc lines the checker previously reported as OK.

## Motivation / Context

`check-docs-mdx` reported OK on content Fern's MDX parser actually rejects: a bare `<` followed by a non-name-start character reads as the start of a JSX tag and fails with `Unexpected character = before name`. The three existing pattern families were a reactive denylist and never covered this class — a false-OK gap.

Fixes: #2050
Related: N/A

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tools/check-docs-mdx` (docs lint)

## Implementation Notes

Rather than add a Node/remark MDX parser to the `make lint` fast path, the `<`-family rule is inverted from a denylist into a fail-closed allowlist: flag any bare `<` not followed by a letter, `/`, `!`, or `>`, evaluated outside code fences and inline spans. This closes the `<`-follow hazard by construction instead of enumerating one more escape each time a new offender surfaces.

The checker **remains an approximation** of the actual MDX parser (now stated in-script) and can still drift on hazard classes outside the `<`-follow family. This PR does not claim full MDX parity — it closes one structurally-defined class.

## Testing

```bash
./tools/check-docs-mdx        # OK
./tools/check-docs-mdx_test.sh # passes; provably fails if check 6 is removed
make lint                      # PASS
make test-shell                # PASS (runs the new hermetic regression test)
make qualify                   # all phases PASS (e2e 24/24) except a final
                               # environmental api-diff (missing apidiff binary
                               # in sandbox — unrelated; no Go API surface touched)
```

The new `tools/check-docs-mdx_test.sh` is hermetic and is provably load-bearing: it fails when check 6 is removed.

## Risk Assessment

**Rollout notes:** N/A — isolated docs + tooling change, easily reverted.

## Checklist


tools/check-docs-mdx
```diff
@@ -132,6 +146,20 @@ for dir in "${SEARCH_DIRS[@]}"; do
...
+      # at end of line — is a parse error (e.g. `(gate <= 2,000)` yields
+      # "Unexpected character = (U+003D) before name"). This complements
+      # checks 4 and 5 rather than replacing them: they cover the cases where
+      # `<` IS followed by a name-start (autolinks, placeholders), while this
+      # covers the cases where it is not. Inverting to an allowlist means any
+      # future `<`-follow hazard fails closed by default.
+      line ~ /<([^A-Za-z!\/>]|$)/ {
+        printf "MDX: bare < not starting a valid tag: %s:%d: %s\n", FILENAME, NR, $0
+        errors++
+      }
```
njhensley, line 212: Good catch — fixed in c6c869ea3. The fence tracker now handles both ```` ``` ```` and `~~~` fences with CommonMark closing-length semantics, and inline-span stripping now matches by backtick-run length (so `` `` ``…`` `` `` double-backtick spans are preserved). Added regression fixtures for tilde-fenced, double-backtick, and fence-length cases; a negative control confirms a bare `<=` outside any span/fence still fails, so detection is not weakened.

mchmarny (approved):
Verified against current `main` (`602c8db1`):

- Diff applies clean; with the tool change alone the new check-6 rule flags exactly the four offenders (`aks-gpu-setup.md:312`, `validation.md:533`, `component-catalog.md:73`, `api-server.md:125`) and nothing else — no false positives across `docs/user`, `docs/integrator`, `docs/contributor`.
- With the doc fixes applied: `OK: all doc files are MDX-safe`.
- `tools/check-docs-mdx_test.sh`: 11/11 pass; removing check 6 makes `bare-lt-exits-nonzero` fail, so the rule is load-bearing.

This unblocks `Publish Fern Docs`, which has been failing on `docs/integrator/aks-gpu-setup.md:312` (`Unexpected character = (U+003D) before name`) — most recently on the `v0.18.0` tag push (run 30029834883). Merging and re-dispatching the publish.

The open CodeRabbit nits (duplicate diagnostic for `<3` across checks 5/6, the header inline-span claim for check 1, indented/multi-line fence edges) are refinements to a checker that is explicitly documented as an approximation — tracking as a follow-up rather than holding the publish fix.

PR #2138, opened 2026-08-09
fix(coverage): resolve UAT wiring from the reservation registry

### LINKED ISSUE #1977: tools/coverage wired-runner scanner blind to UAT lanes after reservation-registry rewrite
## Summary

`make coverage-docs` (the `tools/coverage` generator behind `docs/user/coverage-matrix.md`) can no longer see the wired UAT lanes. Regenerating the page today silently downgrades every UAT-derived row — `uat, GPU (H100, real), nightly` becomes `chainsaw, simulated, per-PR`, and `evidence verify` becomes `not-yet-covered` — which would falsely claim nightly UAT coverage vanished.

## Cause

Commit a363ce4c8 (UAT reservation registry, #1274/#1559) restructured the UAT workflows:

- `TEST_CONFIG` is now composed from `${{ inputs.… }}` expressions, so `tools/coverage/wiring.go`'s `uatConfigRef` regex no longer matches any literal `tests/uat/<cloud>/tests/*-config.yaml` string in the workflow files.
- Several scanned commands moved into `tests/uat/lib/phases.sh`, which the wired-runner scanner does not read.

Two secondary staleness items in the same generator:

- The version axis still reports `main only`, but the nightly lanes run with `previous_n=1`.
- `make coverage-check` is opt-in (not wired into `make qualify` or the merge gate), so the page will read as stale/wrong until the scanner is fixed, with nothing flagging it on PRs.

## Impact

- The committed `docs/user/coverage-matrix.md` is currently correct only because contributors know not to commit regenerated output; anyone who runs `make coverage-docs` and commits the result regresses the page.
- Discovered during PR #1967's review: a one-sentence preamble fix had to be hand-applied instead of regenerated for exactly this reason.

## Proposed fix

- Teach `wiring.go` to resolve `TEST_CONFIG` through the reservation-registry inputs (or scan the composite/lib layer, including `tests/uat/lib/phases.sh`) instead of grepping workflow files for literal config paths.
- Update the version axis to reflect `previous_n`.
- Consider wiring `make coverage-check` into `make qualify` once the scanner is trustworthy again, so drift is caught at PR time.

Inherited from main (pre-dates #1967); not a blocker for that PR.

Dhirenderchoudhary:
## Summary

The coverage-matrix generator went blind to the nightly UAT lanes after the reservation-registry rewrite, so `make coverage-docs` silently downgraded every UAT row. Rewires the scanner to read the registry and the per-cloud pipelines directly, and gates matrix freshness at PR time.

## Motivation / Context

Regenerating `docs/user/coverage-matrix.md` on current `main` drops `uat` from 6 rows and demotes `evidence verify` from `covered`/`nightly` to `not-yet-covered`. The committed page was only accurate because nobody regenerated it — #1967 hand-edited it rather than running the generator.

Two independent causes:

- `uatConfigRef` matched a literal `tests/uat/<cloud>/tests/<name>.yaml` in the workflows. None remain — the pipelines build `TEST_CONFIG` from inputs resolved out of the reservation registry, so `wired.intents` came back empty.
- `tests/uat/*/run` are shims that source `tests/uat/lib/phases.sh`. Scanning only the shims found no `aicr` invocations at all. The one row that kept `uat` (`validate`) was matching prose inside a comment in `aws/run`, not a real invocation.

Fixes: #1977
Related: #1967, #1644

## Type of Change

- [x] Documentation update
- [x] Build/CI/tooling

## Component(s) Affected

- [x] Other: `tools/coverage`, `.github/workflows/merge-gate.yaml`, `Makefile`

## Implementation Notes

**Each fact now comes from the artifact that owns it.** `infra/uat/reservations.yaml` (via `pkg/uatbroker`) answers *which lanes and intents* the nightly batch dispatches; the per-cloud `uat-<cloud>.yaml` pipelines answer *which runner phases are enabled steps*. `tests/uat/lib` becomes a UAT signal root, gated on there being at least one enrolled lane so an inert library can't read as live coverage.

**Both scans fail closed.** Returning an empty result on a moved registry or a renamed input is indistinguishable from "nothing is wired" — that's the exact silent downgrade this generator exists to prevent, and the old fail-open behaviour is why this rotted unnoticed.

**Journey coverage requires both halves.** A reservation enrolled for the intent *and* its pipeline actually executing the journey's workload phase. This is the one judgment call worth review: once the scanner reads the registry, `cuj2-inference-dynamo` wants to flip to `covered`, because inference *is* enrolled nightly on aws/gcp/azure/gb200. But `serve` is commented out in all four pipelines (#1644) — the stack is deployed and validated, the journey never runs. Marking it `covered` would trade this issue's false negative for a false positive and then lock that into the merge gate, so it stays `stubbed` with a note pointing at the disabled step instead of the current one citing DC3/DC6, both closed.

Decoding the pipelines as YAML rather than grepping is what makes that readable: a commented-out step is absent from the parsed document, where a raw-text scan matches its invocation verbatim.

**Version axis** is derived from `uat-nightly-batch.yaml`'s `previous_n` instead of the hardcoded "the live axis today is `main`" claim, which stopped being true when the multi-version matrix landed.

**Freshness gating**: `coverage-check` joins `make qualify`, and a `coverage-freshness`/`-skip` job pair joins the merge gate (mirroring `bom-freshness`), watching every signal source. The generator reads only in-repo signals — no network, no cluster — so unlike `bom-check` it's cheap enough to gate directly.

Net effect on the committed matrix: the 6 UAT rows are restored and now reproducible (the CLI-verb table regenerates byte-identical to what was committed), the version-axis paragraph is corrected, and `cuj2-inference` keeps an accurate note.

## Testing

```bash
make coverage-docs   # regenerated; byte-stable across consecutive runs
GOFLAGS=-mod=vendor go test ./tools/coverage/... -count=1 -race
GOFLAGS=-mod=vendor go test ./pkg/uatbroker/... ./tools/uat-broker/... ./tests/uat/...
GOFLAGS=-mod=vendor golangci-lint run -c .golangci.yaml ./...
GOFLAGS=-mod=vendor go vet ./...
make lint-yaml check-docs-mdx check-docs-filenames license check-agents-sync
actionlint .github/workflows/merge-gate.yaml
```

All green. `golangci-lint` reports 0 issues module-wide.

Coverage delta — `tools/coverage: 82.2% → 85.5% (+3.3%)`.

I mutation-tested the new guards to confirm they aren't vacuous (each mutation reverted after):

| Mutation | Caught by |
|---|---|
| Remove the `tests/uat/lib` scan target (the original bug) | `TestBuildMatrixStatusFromSignals`, `TestUATVerbsResolveThroughSharedPhaseLibrary`, `TestRepoUATLanesAreVisible` |
| Drop the journey-phase gate (intent-only coverage) | `TestJourneyStepDisabledIsNotCovered` |
| Make the registry load fail-open | `TestBuildMatrixFailsClosed` |

`TestRepoUATLanesAreVisible` runs the real scanner against the real repo as a standing anti-rot guard — if the registry moves or the phase commands migrate out of `tests/uat/lib` again, it fails there instead of silently downgrading the next regeneration.

Two caveats on the full suite: I could not run `make qualify` end to end (`e2e` and `scan` need Docker + registry access I don't have locally), and full `make test` has pre-existing failures in `pkg/oci` (`TestHelmV4_2_3ExplicitVersionPull`) and `tests/releasepolicy` (needs GNU `timeout`, absent on macOS). I confirmed the identical failure set on unmodified `origin/main`, so neither is from this change. Leaving those to CI.

## Risk Assessment

The generator change itself is isolated and revertible, but this adds a new blocking merge-gate job, so a false positive there would block unrelated PRs.

**Rollout notes:** `coverage-freshness` only fires on PRs touching the matrix inputs (the doc, `tools/coverage`, `pkg/cli`, `pkg/uatbroker`, `infra/uat/reservations.yaml`, the `uat-*.yaml` pipelines, `tests/uat`, `tests/chainsaw`, `demos`, `Makefile`); everything else hits the `-skip` companion. The failure message names the fix (`make coverage-docs` and commit). No migration, no flags, backwards compatible — revert the workflow hunk to disable the gate without touching the generator.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — see the caveat above on two pre-existing env failures


tools/coverage/wiring.go
```diff
@@ -15,80 +15,270 @@
...
+					phases[m[1]] = true
+				}
 			}
 		}
 	}
-	for r := range runnerSet {
-		w.runners = append(w.runners, r)
+	if len(phases) == 0 {
+		return nil, errors.New(errors.ErrCodeNotFound,
+			rel+" declares no enabled UAT runner step, but its cloud is enrolled in the nightly batch")
```
njhensley: **🟡 Minor — Fail-closed on a fully-commented enrolled pipeline breaks every local `make qualify`**

When an enrolled cloud (non-empty `nightly-intents` in reservations.yaml) has ALL runner steps commented out in its `uat-<cloud>.yaml`, `scanLanePhases` returns this error → `BuildMatrix` fails → `make coverage-check` exits 1. Because `coverage-check` is wired into `qualify` UNCONDITIONALLY (Makefile:296) — unlike the path-filtered merge-gate job — this breaks *every* developer's `make qualify` regardless of what they changed, until the state is reconciled.

**Blast radius:** The realistic trigger is contrived (the real disabled-`serve` case keeps `train` enabled), so this is design tension rather than a bug — but the opt-out that fixes it (`nightly-intents: []` on that cloud's reservation) is documented only in reservations.yaml, not at this error site or the Makefile.

**Fix:** Name the opt-out in the error string (or the coverage-check Makefile comment) so a pipeline maintenance window doesn't turn into a confusing toolchain-wide `make qualify` failure. Optionally degrade a single otherwise-reachable lane to a stubbed row rather than a hard BuildMatrix error.

tools/coverage/wiring.go
```diff
@@ -15,80 +15,270 @@
...
+	phaseRef := regexp.MustCompile(`tests/uat/` + regexp.QuoteMeta(cloud) + `/run[ \t]+([a-z][a-z0-9_-]*)`)
+	phases := map[string]bool{}
+	for _, job := range wf.Jobs {
+		for _, step := range job.Steps {
+			// Line by line, skipping shell comments. Decoding the YAML drops a
+			// commented-out *step*, but a `run:` block is an opaque scalar, so a
+			// commented line inside one survives into step.Run and would otherwise
+			// read as an executed phase.
+			for _, line := range strings.Split(step.Run, "\n") {
+				if strings.HasPrefix(strings.TrimSpace(line), "#") {
```
njhensley: **🟡 Minor — scanLanePhases only strips full-line comments — a trailing/echo-embedded runner ref is a false phase match**

This filter skips a line only when its trimmed prefix is `#` (full-line comment), then runs `phaseRef` against the whole surviving line. A trailing inline comment or echo-embedded reference — e.g. `echo done  # ./tests/uat/aws/run serve "${TEST_CONFIG}"` — registers `serve` as executed (reproduced independently by two reviewers as `map[serve:true]`), flipping `cuj2-inference-dynamo` from `stubbed` to `covered` and claiming real GPU-nightly coverage that never ran. This is exactly the comment-block false-credit the YAML-decode design exists to prevent — the fix went halfway.

**Blast radius:** Latent, not live: today's disabled `serve` steps are full YAML comment blocks (stripped by `yaml.Unmarshal` before reaching `step.Run`) and enabled steps are bare invocations, so nothing triggers it in-repo. A future contributor disabling a phase with an inline comment would silently over-report the journey.

**Fix:** Cut each line at its first unquoted `#` before matching (or anchor the match to the line's first command token), and add a trailing-comment-only case to `TestShellCommentedPhaseIsNotExecuted`.

tools/coverage/wiring.go
```diff
@@ -15,80 +15,270 @@
...
+	var wf nightlyBatchWorkflow
+	if err = yaml.Unmarshal(data, &wf); err != nil {
+		return VersionAxis{}, errors.Wrap(errors.ErrCodeInvalidRequest, "parse "+nightlyBatchRelPath, err)
+	}
+	input, ok := wf.On.WorkflowDispatch.Inputs[previousNInput]
+	if !ok {
+		return VersionAxis{}, errors.New(errors.ErrCodeNotFound,
+			nightlyBatchRelPath+" declares no "+previousNInput+" input; the version axis cannot be derived")
+	}
+	n, err := strconv.Atoi(input.Default)
```
njhensley: **🔵 Nitpick — Version axis reads the workflow_dispatch input default, not the cron's `|| '1'` fallback (comment misstates GHA semantics)**

`scanVersionAxis` reads `on.workflow_dispatch.inputs.previous_n.default`, and the comment above claims the cron uses that default. It does not: on a `schedule` event the `inputs` context is empty and workflow_dispatch defaults are NOT applied — which is precisely why uat-nightly-batch.yaml uses a separate literal `PREVIOUS_N: ${{ inputs.previous_n || '1' }}`. The true scheduled value is the `|| '1'` fallback, not the input default.

**Blast radius:** Both literals are `'1'` today so the rendered prose is correct, and the freshness gate reads the same default so CI stays green even under drift — only a human reader is misled if someone bumps one literal and not the other. Cosmetic doc prose.

**Fix:** Source the axis from the `|| 'N'` fallback the cron actually uses (or assert the two literals agree and fail closed on mismatch), and correct the comment about GitHub schedule/inputs semantics.

njhensley (commented):

**Method:** four independent persona reviewers (Correctness · CI/Operability · Domain & Architecture · Test-coverage), then an adversarial senior meta-reviewer that re-derived every finding against the resolved code at `34f506f6`.

### Overall — Approve with comments

Well-reasoned, well-tested fix for a real silent-downgrade bug (#1977). The old generator regex-scraped `uat-*.yaml` for literal `tests/uat/<cloud>/tests/<name>.yaml` config paths that no longer exist after the reservation-registry rewrite, so it saw zero wired lanes and quietly demoted every UAT row. The rewrite sources each fact from the artifact that owns it — *which lanes/intents* from `infra/uat/reservations.yaml` (via `pkg/uatbroker`), *which phases* from the per-cloud pipelines decoded as YAML — fails closed on unresolvable inputs, and gates freshness in both `make qualify` and the merge gate.

Verified the honesty contract against the real files: all four pipelines (aws/gcp/azure/kind) enable `train` and comment out `serve`, so cuj1→covered and cuj2→stubbed with the accurate journey-step-disabled note. The test suite is strong (86.2%; all three advertised mutations genuinely caught; the fail-closed table's error substrings all match real errors). **No blockers, no majors.**

Two 🟡 (both latent, not live in the current tree) and three 🔵 are inline / below.

### Additional nitpicks (anchor to unchanged lines, so noted here)

- 🔵 **Verb-level UAT coverage doesn't honor disabled phases** — [`tools/coverage/scan.go#L193`](https://github.com/NVIDIA/aicr/blob/34f506f6407bc82edd8bb30eb5fdfda6e308ddf5/tools/coverage/scan.go#L193). `scanVerbs` reads `phases.sh` as one blob, so a verb reachable only from a disabled phase function would still be UAT-covered — asymmetric with the phase-gated CUJ rows. Harmless today (`phase_serve` has no `aicr` invocation); latent if a serve-only verb is added later.
- 🔵 **`looksLikeText` branch has 0% coverage** — [`tools/coverage/scan.go#L181`](https://github.com/NVIDIA/aicr/blob/34f506f6407bc82edd8bb30eb5fdfda6e308ddf5/tools/coverage/scan.go#L181). Extensionless text-sniff path is never walked; a one-line extensionless-script fixture under `tests/chainsaw/` would close it.
- 🔵 **`kind` single-GPU nvkind collapses into the coarse "GPU (H100, real)" class** — [`tools/coverage/build.go`](https://github.com/NVIDIA/aicr/blob/34f506f6407bc82edd8bb30eb5fdfda6e308ddf5/tools/coverage/build.go). The hardware/cadence columns union all lanes, so today enrolling kind flips no verdict; latent overclaim only if a future intent runs *only* on the kind lane. Not introduced by this PR.

- **Refuted — "axis prose overclaims by ignoring `nightly-intent-min-versions`":** the premise was wrong — v0.18.0 and v0.19.0 are both shipped, so the `>= v0.18.0` gate is satisfied by the previous-release cell; and the generator uses `NightlyIntentsOrDefault`, never `EligibleNightlyIntents`, so the min-version gate never touches the matrix. Prose is accurate.
- `readBoundedFile` bounded-read pattern is correct (`LimitReader(f, max+1)`, `>max` reject; read-only `Close` ignored per repo rules).
- `yaml:"on"` tag decodes correctly under yaml.v3 (no YAML 1.1 boolean-alias issue).
- Determinism holds (sorted clouds, membership-only maps). All-opted-out registry → zero lanes with no error is correct-by-design, not fail-open.
- `coverage-freshness`/`-skip` are mutually exclusive over the boolean paths-filter output; the aggregate gate can't deadlock; pinned SHAs; no network/cluster/tool deps ("in-repo signals only" confirmed — no `os/exec`, `net/http`, or `go build`).
- `pkg/uatbroker` coupling is the correct dependency direction and eliminates schema drift; the paths-filter watches every signal the generator reads.

### Summary

🔴 Blocker 0 · 🟠 Major 0 · 🟡 Minor 2 · 🔵 Nitpick 3 — **Approve with comments.** Neither 🟡 is live in the current tree; both are cheap hardening opportunities, not merge blockers.

Dhirenderchoudhary:
@njhensley

Thanks for the thorough review really helpful.

Addressed both yellows and the version-axis nit in `e597f0cd`:
- empty enrolled pipeline error now names the `nightly-intents: []` opt-out
- phase scan strips trailing/unquoted `#` before matching (with a trailing-comment test)
- version axis requires the schedule `|| 'N'` fallback to agree with the dispatch default

Also merged `main` and cleared the `merge-gate.yaml` conflict. Happy to follow up on anything else.

njhensley:
@Dhirenderchoudhary, please ensure your commits are `signed and verified`. Thanks!

Dhirenderchoudhary:
@njhensley Verified and signed

njhensley:
@Dhirenderchoudhary, please address linting issues. Should be good to go after that.

Dhirenderchoudhary:
FIxed lint CI failures

njhensley (approved):
All good on my end — approving! ✅

Everything from the review is now buttoned up:

- Fail-closed opt-out names nightly-intents: [] in both the error string and the Makefile comment — no more mystery make qualify failures during pipeline maintenance. 👍
- Comment stripping got a proper quote/backslash-aware stripShellComment with its own TestStripShellComment plus a trailing-comment case in TestShellCommentedPhaseIsNotExecuted. Closes the false-credit gap cleanly.
- Version axis now reads the real || 'N' schedule fallback and asserts it agrees with the dispatch default (fails closed on drift), with the GHA semantics fixed in the comment and the model.go godoc.
- gofmt — confirmed fixed in fc823779; lint is clean now.

Re-verified against the real repo: go test ./tools/coverage/... -race passes (including TestRepoUATLanesAreVisible), and coverage-matrix.md regenerates byte-identical and stable across runs. The filepath.IsLocal guard and merge-gate self-gating hunk are both correct too.

Really thorough follow-through on every point — thanks for that. 🚀

PR #2155, opened 2026-08-11
fix(recipe): demote nodes criteria dimension to metadata-only

### LINKED ISSUE #1781: recipe: demote nodes criteria dimension to metadata-only

Kevin-Hawkins:
## Summary

Remove `nodes` from `Criteria.Matches()` — it is metadata-only advisory information, not an overlay selection dimension. `nodes` is retained in `Specificity()` so that nodes-only CLI queries pass the minimum-specificity guard.

## Motivation / Context

No overlay in the embedded catalog gates on `nodes`. Despite this, the field participated in overlay selection:

- `Matches()`: the removed branches were already inert for the embedded catalog — both required `c.Nodes != 0` on the overlay receiver, and no embedded overlay sets `nodes`, so queries always matched. The removal is a correctness fix and cleanup rather than a behavior change for embedded-catalog users.
- `Specificity()`: the only observable behavioral change — `--nodes N` now correctly scores Specificity()==1 so nodes-only queries pass the CLI guard, while tiebreaking is unaffected (no overlay sets the field).

Audit performed across `recipes/overlays/`, `recipes/mixins/`, `pkg/recipe/testdata/`, `tests/`, and `examples/` — zero overlays set `criteria.nodes:`.

The `Criteria.Nodes` field is retained: it is still parsed from CLI/API/snapshots and used downstream by the bundler (GKE priority quota calculation) and fingerprint matching.

Fixes: #1781
Related: #1542

## Type of Change

## Component(s) Affected

## Implementation Notes

**`nodes` retained in `Specificity()` but removed from `Matches()`:** This asymmetry is intentional — `Specificity()` must count `nodes` so that a `--nodes 8` query passes the `Specificity()==0` guard in the CLI and server. Since no overlay sets `criteria.nodes`, the score never influences tiebreaking.

**New server-side rejection path:** `criteriaValid()` now guards both `/recipe` and `/query` endpoints. Requests with zero-specificity criteria (all dimensions unset) return `400 "no criteria provided: specify at least one of service, accelerator, intent, os, platform, nodes"`. This is a behaviour change: previously, the server would resolve and return the base recipe for an empty-criteria request.

**Hard error for external catalogs with `criteria.nodes`:** External `--data` overlays with `criteria.nodes != 0` are now rejected at catalog load time (`ErrCodeInvalidRequest`). Such overlays would silently match every query post-demotion. Operators must remove or zero `criteria.nodes` from their external catalog before upgrading.

## Testing

```bash
make qualify
```

All tests pass. The only failure in `make qualify` is a pre-existing `pkg/oci` test pinning Helm `v4.2.3` while `v4.1.3` is installed locally — unrelated to this change.

Coverage: `pkg/recipe` 89.1% (+0.1%).

## Risk Assessment

- [x] **Low** — Embedded catalog users: no overlay selection behaviour changes. External `--data` catalog operators with `criteria.nodes` set will receive a hard error at startup and must audit before upgrading.

**Rollout notes:** Operators using `--data` catalogs should audit for `criteria.nodes` before upgrading. The load-time error provides an explicit signal rather than silent misbehaviour.

## Checklist


pkg/recipe/criteria.go
```diff
@@ -557,9 +553,7 @@ func (c *Criteria) Specificity() int {
 	if c.Platform != CriteriaPlatformAny && c.Platform != "" {
 		score++
 	}
-	if c.Nodes != 0 {
-		score++
-	}
+	// Nodes is metadata-only and does not contribute to specificity; see #1781.
```
mchmarny: **Blocking.** Removing `Nodes` from `Specificity()` breaks `aicr recipe --nodes 8`, which is a documented invocation (`docs/user/cli-reference.md` line 462).

`Specificity()` has two consumers, and only one is inert for `nodes`:
- `metadata_store.go:632` — overlay tiebreak sort. Inert, since no overlay sets `Nodes` (audit confirmed).
- `cli/query.go:238` — the "user stated no criteria" guard. **Not inert.** `--nodes 8` now scores 0 and trips it.

Reproduced against `main` and this head:
```
main    $ aicr recipe --nodes 8   → recipe emitted (overlays: base, monitoring-hpa), exit 0
21b9cd2 $ aicr recipe --nodes 8   → [INVALID_REQUEST] no criteria provided: specify at least
                                     one of --service, --accelerator, --intent, --os,
                                     --platform, --nodes, --config, ...        exit 2
```
The message names `--nodes` as a remedy for the guard `--nodes` can no longer satisfy. Same wording at `query.go:209` on the snapshot path. `pkg/server` has no equivalent guard, so `GET /recipe?nodes=8` still succeeds — CLI and API now disagree.

Two ways to close this; either is fine:
1. **Keep `Nodes` in `Specificity()`** and scope the PR to `Matches()`. I verified this restores the CLI path with the `Matches()` change fully intact, and it costs nothing for the stated goal since the tiebreak is already inert.
2. **Keep the removal** and drop `--nodes` from both messages (`query.go:209`, `:240`), declare the behavior change in the PR body and `docs/user/cli-reference.md`, and add a CLI-level regression test asserting `--nodes`-only now errors.

Either way this needs a regression test at the `buildRecipe` level — `recipe_test.go:141` only covers flag→`Criteria` parsing, which is why `make qualify` stayed green.

mchmarny (changes requested):
**REQUEST_CHANGES** — 1 blocker, validated against head `21b9cd25`. CI was still running; it does not cover the affected path.

The `Matches()` demotion is sound and the catalog audit checks out — I re-ran it with `yq` across every `recipes/**` and `pkg/recipe/testdata/**` file and confirmed zero overlays set `spec.criteria.nodes`. Both removed `Matches()` branches require `c.Nodes != 0`, so the embedded catalog is genuinely unaffected. Issue #1781's `completionTuple` sub-item is resolved by construction. The blocker is the `Specificity()` half only — see the inline comment.

Non-blocking: [`docs/contributor/recipe.md` line 464](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/docs/contributor/recipe.md#L464) still states nodes *"remains a matching dimension (present in `Criteria.Matches`)"* — the same sentence corrected in `coverage.go`, but the doc mirror was missed.

pkg/recipe/criteria.go
njhensley: **🔴 Blocker — Demoting nodes from Specificity() breaks nodes-only `aicr recipe --nodes N`**

Removing the `if c.Nodes != 0 { score++ }` block here means a nodes-only Criteria scores Specificity()==0 (reproduced). That trips the CLI guard `if criteria.Specificity()==0` at pkg/cli/query.go:238, so `aicr recipe --nodes 8` (no other flags) now hard-errors with ErrCodeInvalidRequest — pre-PR it scored 1 and emitted the generic recipe. This is a real behavior change, contradicting the PR's "no functional change" label. Worse, the guard message (query.go:240 non-snapshot, query.go:209 snapshot path) still lists `--nodes` among the flags that satisfy the guard — it rejects the exact flag it tells the user to supply. @mchmarny filed this same blocker independently.

**Blast radius:** Any operator/script invoking `aicr recipe --nodes N` alone. `--nodes N` combined with any real dimension still works. The server has no equivalent guard, so `GET /v1/recipe?nodes=8` still returns 200 — CLI and API now disagree.

**Fix:** Simplest path that also satisfies #1781: keep Nodes in Specificity() and scope the demotion to Matches() only (the overlay tiebreak at metadata_store.go:632 is already inert). Alternatively, drop `--nodes` from both guard messages and relabel the PR as a behavior change.

pkg/recipe/criteria.go
```diff
@@ -439,17 +439,13 @@ func (c *Criteria) Matches(other *Criteria) bool {
...
-	// Recipe 0 (any) → match any query value
-	if other.Nodes == 0 && c.Nodes != 0 {
-		// Query is generic but recipe is specific - no match
-		return false
-	}
-	if other.Nodes != 0 && c.Nodes != 0 && c.Nodes != other.Nodes {
-		// Both specific but different values - no match
-		return false
-	}
+	// Nodes is metadata-only: no overlay in the embedded catalog gates on
```
njhensley, line 442: **🔵 Nitpick — PR body's Matches() motivation is backwards; external-catalog note is fine**

For the embedded catalog the removed Matches() branches were already inert: both required `c.Nodes != 0` (the overlay/receiver), and no embedded overlay sets nodes, so a nodes-specific query ALWAYS matched pre-PR (reproduced: `overlay{}.Matches(query{Nodes:8})==true`). So the Matches() removal changes nothing for the embedded catalog — the only observable behavioral change is in Specificity() (see blocker). Worth correcting the PR narrative so that isn't obscured.

**Blast radius:** Documentation/rationale only. Separately (optional): an external --data overlay with criteria.nodes set is now silently ignored for matching/specificity/coverage — documented here, but a one-time slog.Warn at overlay-load would make the shift operator-visible ("never silently upgrade").

**Fix:** Correct the PR body to state the Matches() change is inert for the embedded catalog; optionally add a load-time warning for external overlays that declare criteria.nodes.

pkg/recipe/criteria_test.go
```diff
@@ -292,6 +292,44 @@ func TestCriteriaMatches(t *testing.T) {
 			},
 			want: true,
 		},
+		// Nodes is metadata-only (#1781): it does not participate in overlay
+		// selection, so differing Nodes values must never prevent a match.
+		{
```
njhensley, line 312: **🔵 Nitpick — New Matches nodes cases only vary Service+Nodes**

The three new 'nodes ignored' cases all hold Service==EKS and vary only Nodes. None combines a nodes difference with a real mismatch on another dimension, which would confirm the nodes removal doesn't mask a legitimate mismatch.

**Blast radius:** Defensive only — existing cases already cover single-dimension mismatches independently.

**Fix:** Add one case: {Service:EKS,Nodes:4} vs {Service:GKE,Nodes:8} asserting want:false (nodes ignored, service mismatch still wins).

njhensley (commented):
## 🧭 Multi-persona review — `recipe: demote nodes criteria dimension to metadata-only`

**Method:** 4 independent persona reviewers (Correctness · Domain/Architecture · Test-coverage · Docs) → adversarial senior meta-reviewer that re-derived every finding from the resolved code and reproduced the crux by execution. Anchored to head `21b9cd25`.

> **Note:** @mchmarny already has an open **CHANGES_REQUESTED** on this PR, and the blocker below (F1) is their finding — this review **independently corroborates** it (reproduced + 3 personas), it is not a new discovery. CodeRabbit's two comments (external-catalog wording, Matches regression test) were already addressed in the diff.

### Overall assessment

The refactor's intent is sound and the embedded-catalog audit holds: no overlay sets `criteria.nodes`, so removing nodes from `Matches()` is genuinely inert there, and the promised downstream consumers survive — `fingerprint.Match` and the bundler GKE-quota path read `Criteria.Nodes` directly, independent of `Matches`/`Specificity` (no fingerprint/determinism impact).

The one part that changes behavior is removing nodes from **`Specificity()`**, and it breaks a documented CLI invocation while emitting a self-contradictory error message — so the "Refactoring (no functional changes)" label is inaccurate. The fix is small.

**Recommendation: Request changes** (1 Blocker). Inline comments are on the changed lines; findings on unchanged files (CLI guard, server, docs, tests) are listed below since the API can't anchor them to this diff.

### Findings (full list)

| id | tier | finding | location |
|----|------|---------|----------|
| F1 | 🔴 Blocker | Nodes-only `Specificity()==0` trips the CLI guard; message lists `--nodes` as the remedy it rejects; "no functional change" label wrong | [query.go#L238](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/cli/query.go#L238) · [msg L240](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/cli/query.go#L240) · [snapshot L209](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/cli/query.go#L209) · root cause [criteria.go#L556](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/recipe/criteria.go#L556) *(inline)* |
| F2 | 🟡 Minor | CLI/API divergence: server recipe handler has no `Specificity` guard, so nodes-only is accepted via API but rejected via CLI | [recipe_handler.go#L213](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/server/recipe_handler.go#L213) |
| F3 | 🟡 Minor | Stale doc: nodes "remains a matching dimension (present in `Criteria.Matches`)" | [recipe.md#L464](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/docs/contributor/recipe.md#L464) |
| F4 | 🟡 Minor | Stale doc: Specificity() "counts six fields … nodes" (now five) | [recipe.md#L174](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/docs/contributor/recipe.md#L174) |
| F5 | 🟡 Minor | No CLI-boundary test for `aicr recipe --nodes N` (why F1 slipped through) | [query_test.go](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/cli/query_test.go) |
| F6 | 🟡 Minor | No nodes-only → `Specificity()==0` explicit unit case; add `{&Criteria{Nodes:8}, want:0}` | [criteria_test.go#L344](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/recipe/criteria_test.go#L344) |
| F7 | 🔵 Nit | External `--data` overlay with `criteria.nodes` silently ignored — documented; a load-time warn would help | [criteria.go#L442](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/recipe/criteria.go#L442) *(folded into inline)* |
| F8 | 🔵 Nit | PR body's `Matches()` motivation is backwards; real change is `Specificity()` | *inline* |
| F9 | 🔵 Nit | api-reference nodes "(0 = any)" implies matching semantics that no longer exist | [api-reference.md#L136](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/docs/user/api-reference.md#L136) |
| F10 | 🔵 Nit | Glossary/CLI-ref list nodes among criteria that "select/resolve" a recipe | [README.md#L93](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/docs/README.md#L93) · [cli-reference.md#L322](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/docs/user/cli-reference.md#L322) |
| F11 | 🔵 Nit | New Matches tests only vary Service+Nodes | [criteria_test.go#L297](https://github.com/NVIDIA/aicr/blob/21b9cd2577770dcee41c8146904157877372a03d/pkg/recipe/criteria_test.go#L297) *(inline)* |

- **Downstream `Nodes` consumers unaffected** — `fingerprint.Match` and bundler GKE quota (`bundler.go:2517`) read `Criteria.Nodes` directly; no effect on fingerprint identity, determinism, or evidence reproducibility.
- **`Matches()` removal is inert for the embedded catalog** — audit confirmed zero overlays set `criteria.nodes`; removed branches required `c.Nodes != 0`.

### Summary

|:---:|:---:|:---:|:---:|
| **1** | 0 | 5 | 4 |

Cleanest fix: keep `Nodes` in `Specificity()` and scope the demotion to `Matches()` — the only `Specificity` consumer that matters for nodes is the CLI guard (the overlay tiebreak is already inert), so #1781's goal is met without the regression.

pkg/recipe/criteria.go
Kevin-Hawkins: Addressed. The PR description has been corrected: the Matches() removal is now described as a correctness fix/cleanup that was already inert for the embedded catalog (both removed branches required c.Nodes != 0 on the receiver, and no embedded overlay sets nodes). The observable behavioral change is correctly attributed to Specificity().
On the external-catalog warning: went further than slog.Warn — external overlays with criteria.nodes set are now rejected at load time with ErrCodeInvalidRequest, making the shift operator-visible as a hard error rather than a silent one.

pkg/recipe/criteria_test.go
Kevin-Hawkins: Addressed — added "nodes ignored: service mismatch still rejects despite differing nodes": {Service:EKS, Nodes:4} vs {Service:GKE, Nodes:8} asserting want:false, confirming nodes removal doesn't mask a real dimension mismatch.

pkg/recipe/criteria_test.go
```diff
@@ -292,6 +292,95 @@ func TestCriteriaMatches(t *testing.T) {
...
+			criteria: &Criteria{
+				Service: CriteriaServiceEKS,
+				Nodes:   4,
+			},
+			other: &Criteria{
+				Service: CriteriaServiceGKE,
+				Nodes:   8,
+			},
+			want: false,
+		},
```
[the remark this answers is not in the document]
Kevin-Hawkins: Fixed — renamed to "despite differing nodes".

pkg/server/recipe_handler_test.go
```diff
@@ -1060,6 +1060,54 @@ func TestHandleQuery_SelectorPresence(t *testing.T) {
...
+
+			if w.Code != http.StatusBadRequest {
+				t.Fatalf("status = %d, want %d; body: %s", w.Code, http.StatusBadRequest, w.Body.String())
+			}
+			if !strings.Contains(w.Body.String(), "no criteria provided") {
+				t.Errorf("body = %q, want it to contain %q", w.Body.String(), "no criteria provided")
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
Kevin-Hawkins: Added TestHandleRecipes_EmptyCriteriaRejected with table-driven GET and POST cases asserting HTTP 400 and "no criteria provided" for zero-specificity /v1/recipe requests.

mchmarny:
**Re-review at `5b58096`.** The `Specificity()` blocker is fixed, verified by execution rather than inspection: `aicr recipe --nodes 8` emits a recipe (exit 0, overlays `base`, `monitoring-hpa`), and `GET /v1/recipe?nodes=8` returns 200, which also closes the CLI/API divergence (F2). The doc findings are all addressed. One blocker remains, plus one should-fix.

### 🔴 Blocker: the nodes regression test is inert

[`pkg/cli/recipe_test.go#L707`](https://github.com/NVIDIA/aicr/blob/5b580962b9c08cda982f19415b3f57d4ab3d91f3/pkg/cli/recipe_test.go#L707) passes `--no-cluster` to `recipe`, which has no such flag (only `validate` does). The command aborts at flag parsing before it ever reaches the guard:

```
WITH    --no-cluster: err = "flag provided but not defined: -no-cluster"
WITHOUT --no-cluster: err = <nil>
```

The assertion is `err != nil && strings.Contains(err.Error(), "no criteria provided")`, so a flag-parse error satisfies it trivially. Mutation check: deleting the `Nodes` block from `Specificity()` and re-running leaves `pkg/cli` green. The test does not detect the regression it is named for, and its comment ("If nodes were removed from `Specificity()`, this assertion would fail with exactly that message") is not true today.

This is the same gap that let the original defect through: `make qualify` stayed green because nothing exercised `buildRecipe`. The `Specificity()` unit case in `pkg/recipe` does catch a direct removal, but the CLI-boundary guard, the one that would catch a change on the `pkg/cli/query.go` side, catches nothing.

Fix: drop `"--no-cluster",` from the args. The criteria path is already offline ("generating recipe offline - reads inputs only; does not deploy to or modify any cluster"), so test isolation is not at risk here; the flag simply does not exist on this command. The [suggestion that introduced it](https://github.com/NVIDIA/aicr/pull/2155#discussion_r3772356051) was aimed at commands that do have it.

### 🟠 Should fix: OpenAPI spec not updated for the new 400

This PR adds an empty-criteria rejection to `GET` and `POST` on `/v1/recipe`, `/v2/recipe`, `/v1/query`, and `/v2/query`, all previously 200. `docs/user/api-reference.md` documents it; [`api/aicr/v1/server.yaml`](https://github.com/NVIDIA/aicr/blob/5b580962b9c08cda982f19415b3f57d4ab3d91f3/api/aicr/v1/server.yaml#L120) does not. Every criteria parameter there is still `required: false` with `default: any` and "If omitted, treated as `any` (wildcard)", which is now false for the all-omitted case.

Nothing in CI catches this (Fern does not consume the spec), so anyone generating a client from the contract keeps the stale behaviour. Worth adding the 400 to those four operations in the same PR.

---

Reviewed at `5b58096` in a detached worktree. `pkg/cli`, `pkg/recipe`, and `pkg/server` tests green; `golangci-lint -c .golangci.yaml` on all three reports 0 issues.

pkg/server/recipe_handler.go
```diff
@@ -700,6 +696,24 @@ func bodyHasTopLevelProfile(data []byte, format serializer.Format) (bool, error)
...
+// matching the CLI guard in pkg/cli/query.go. Applied to both /recipe and
+// /query so that empty-criteria resolution is consistently rejected. Returns
+// true when the request may proceed.
+func criteriaValid(w http.ResponseWriter, r *http.Request, criteria *recipe.Criteria) bool {
+	if criteria == nil {
+		WriteError(w, r, http.StatusBadRequest, aicrerrors.ErrCodeInvalidRequest,
+			"Recipe criteria cannot be empty", false, nil)
+		return false
+	}
+	if criteria.Specificity() == 0 {
```
njhensley, line 709: **🟡 Minor — empty-criteria → 400 guard is a v1+v2 behavior change broader than the nodes demotion**

`criteriaValid()` now rejects any `Specificity()==0` request with `400 "no criteria provided"` on **v1 and v2**, for **both `/recipe` and `/query`** — the prior code only rejected `criteria == nil`. This is broader than demoting `nodes` and changes a stable v1 contract: `GET /v1/recipe` with no params (the old *Minimal request* example, now removed from api-reference.md) previously returned the base recipe, and `GET /v1/query?selector=...` with no criteria previously resolved against the base recipe. It also runs *before* profile handling in `handleRecipes`, so a v2 profile-only request with an empty criteria object is rejected too (`validateV2Criteria` returns nil for empty, so this is the sole gate).

**Blast radius:** API consumers of stable v1 that relied on empty→base resolution, and any v2 profile-only flow. It's documented (server.yaml + api-reference.md), tested (`TestHandleRecipes_EmptyCriteriaRejected`, `TestHandleQuery_EmptyCriteriaRejected`), resolves prior finding F2 (CLI/API divergence), and matches the repo's fail-closed philosophy — so this is a confirm-intent flag, not a defect.

**Fix:** No code change required. Get an explicit maintainer sign-off that (a) dropping empty→base from stable v1 is acceptable and (b) profile-only v2 resolution was never intended to be valid, and consider a one-line note in the PR body that this guard is an intentional API behavior change beyond the nodes demotion.

njhensley (approved):
## 🧭 Re-review — `recipe: demote nodes criteria dimension to metadata-only`

Re-review of my prior **COMMENTED** pass (2026-08-12, `21b9cd25`). **7 commits** landed since; delta reviewed against head `60ea73c0`. **Recommendation: Approve with comments** — the blocker is fixed, all lower-tier findings are resolved, and the one net-new item below is a confirm-intent flag, not a defect.

**Legend:** ✔️ Addressed · ◐ Partial · ✖️ Not addressed · ⊘ No longer applicable

### Overall assessment

The author took **Option 1** from the prior review — kept `Nodes` in `Specificity()`, demoted it only from `Matches()` — which resolves the 🔴 blocker: `aicr recipe --nodes 8` now scores `Specificity()==1`, passes the CLI guard, and the guard message no longer names the flag it was rejecting. All 11 prior findings are addressed (F4 became moot — keeping `nodes` in `Specificity()` makes the "six fields incl. nodes" doc correct again). The external-catalog concern (F7) went further than requested — a `--data` overlay with `criteria.nodes` is now a hard load-time `ErrCodeInvalidRequest`, and I confirmed `LayeredDataProvider.Source()` returns `"external"` for `--data` files, so that guard fires in production, not just the test fake. `go test ./pkg/recipe/ ./pkg/server/ ./pkg/cli/` passes.

### Prior-feedback status

| Disposition | Prior finding | Where it landed |
|-------------|---------------|-----------------|
| ✔️ Addressed | **F1 🔴** `Specificity()` demotion breaks nodes-only CLI | `nodes` restored to `Specificity()` (`criteria.go` L557), commit `b03a58a0` + regression test |
| ✔️ Addressed | **F2 🟡** CLI/API divergence (server had no guard) | `criteriaValid()` guards `/recipe` + `/query` (`recipe_handler.go`) — see the inline note |
| ✔️ Addressed | **F3 🟡** recipe.md "remains a matching dimension" (also @mchmarny's non-blocking note) | `docs/contributor/recipe.md` L466 rewritten |
| ⊘ No longer applicable | **F4 🟡** recipe.md "six fields incl. nodes → now five" | `nodes` kept in `Specificity()`, so "six incl. nodes" is now accurate; doc adds the asymmetry note |
| ✔️ Addressed | **F5 🟡** no CLI-boundary test | `TestRecipeCmd_NodesCriteriaPassesGuard` (`pkg/cli/recipe_test.go`) |
| ✔️ Addressed | **F6 🟡** no explicit nodes-only `Specificity()` case | Added, asserting `==1` under the new design (`criteria_test.go`) |
| ✔️ Addressed | **F7 🔵** external overlay nodes silently ignored | Hard load-time error (`metadata_store.go` L345) + two tests |
| ✔️ Addressed | **F8 🔵** PR body Matches() motivation backwards | PR body rewritten; observable change now attributed to `Specificity()` |
| ✔️ Addressed | **F9 🔵** api-reference `nodes "(0 = any)"` | Now "advisory metadata — does not select or filter overlays" |
| ✔️ Addressed | **F10 🔵** glossary/cli-ref list nodes as selecting | `docs/README.md` + `docs/user/cli-reference.md` corrected |
| ✔️ Addressed | **F11 🔵** Matches tests only vary Service+Nodes | Added service-mismatch (`EKS/4` vs `GKE/8` → false) + full-criteria-differing-nodes cases |

### New findings since last review

One net-new 🟡 Minor (inline on `recipe_handler.go`) — a confirm-intent flag on the empty-criteria guard. No code change required.

### ⚠️ Process note

The PR is currently `reviewDecision: CHANGES_REQUESTED`: **@mchmarny's blocker (same as F1) is resolved in `b03a58a0` but their review is still standing and was not auto-dismissed by the pushes.** @mchmarny — could you re-review / dismiss? The nodes-only CLI regression you (and I) flagged is fixed by keeping `Nodes` in `Specificity()`.

### Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major | 0 | — |
| 🟡 Minor | 1 | empty-criteria→400 is a v1+v2 behavior change (confirm-intent, inline) |
| 🔵 Nitpick | 0 | — |

Prior 🔴 and all 10 lower-tier findings resolved. Approving with one confirm-intent comment; merge is still gated on @mchmarny's standing review.

mchmarny (approved):

Issue #1531, opened 2026-06-29
POST /v1/bundle: documented 'bundlers' query param is silently ignored


yuanchen8911:
## Summary

`POST /v1/bundle` documents a `bundlers` query parameter ("Comma-delimited list of bundler types to execute"), but the server never reads it — every bundler runs regardless. The parameter is silently ignored.

## Evidence

- Documented: `docs/user/api-reference.md` (POST /v1/bundle → Query Parameters table, `bundlers` row).
- `pkg/bundler/handler.go`: the `bundleParams` struct has no `bundlers`/bundler-filter field, and `parseQueryParams` reads `set`, `dynamic`, `deployer`, node-selector/toleration params, etc. — but never `query["bundlers"]`.

So a request like `POST /v1/bundle?bundlers=gpu-operator` bundles **every** component in the recipe, contrary to the docs.

## Fix options

1. Implement the filter: parse `bundlers`, restrict generated output to the named bundler types.
2. Or remove `bundlers` from the API until implemented.

Docs (#1528) caveat the parameter as "not currently honored" pending this issue.

Issue #928, opened 2026-05-15
harden(recipes): direct dependencyRefs from monitoring-CR creators to prometheus-operator-crds


mchmarny:
## Summary

After #926 split `prometheus-operator-crds` out as its own release, several components that create `monitoring.coreos.com/v1` CRs (PodMonitor / ServiceMonitor / etc.) rely on a multi-hop transitive dependency path rather than declaring a direct `dependencyRefs` edge. This is brittle.

## Concrete example

`nvsentinel`'s registry comment at `recipes/registry.yaml:258-264` documents the chain:

> nvsentinel depends on gpu-operator, which depends on kube-prometheus-stack, which depends on prometheus-operator-crds

If anyone removes the `gpu-operator → kube-prometheus-stack` edge in `recipes/overlays/base.yaml` — plausible since gpu-operator doesn't strictly *need* kube-prometheus-stack to install, only to scrape metrics — the helm-diff race on nvsentinel's PodMonitor CRD silently returns. No CI test catches it; the failure surfaces only on a fresh-cluster `helmfile apply`.

## Why this matters

The whole thesis of #926 is "let `dependencyRefs` encode the ordering" — applying that consistently means every component whose templates create monitoring CRs should declare `prometheus-operator-crds` as a direct `dependencyRefs` entry. The DAG-flattening makes the extra edge a no-op when the transitive path already exists, but it survives refactors of intermediate nodes.

## Acceptance Criteria

- [ ] Audit `recipes/overlays/`, `recipes/mixins/`, and `recipes/components/*/values.yaml` for charts whose templates emit `monitoring.coreos.com/v1` resources (PodMonitor, ServiceMonitor, Alertmanager, AlertmanagerConfig, Prometheus, PrometheusRule, Probe, ThanosRuler).
- [ ] Add direct `dependencyRefs: [prometheus-operator-crds]` to each such component in `recipes/overlays/base.yaml` (in addition to any existing edges).
- [ ] Update the `nvsentinel` registry comment in `recipes/registry.yaml` to remove the "transitively satisfied" wording once the direct edge is in place.
- [ ] Consider a chainsaw or Go test that asserts every component declaring a `monitoring.coreos.com/v1` CR (detectable by grepping rendered chart output) has `prometheus-operator-crds` reachable in its `dependencyRefs` closure.

## Related

- PR #926 (introduced the split)
- Issue #914 (original race)
- Review comment: https://github.com/NVIDIA/aicr/pull/926#discussion_r3250799459

PR #2168, opened 2026-08-11
docs: ADR-018 class-partitioned bundles

yuanchen8911:
## Summary

Adds ADR-018: class-partitioned bundles — a design for generating a runtime-core bundle and an ops bundle that co-exist in the same cluster, from a single resolved recipe.

## Motivation / Context

Runtime core and ops components have different owners, change-approval processes, and release cadences; the first consuming integration needs a runtime/ops split. The naive approach (two recipes, or two resolutions with filters) allows the same component to ship with different arguments in each bundle. The ADR makes that collision impossible by construction: one resolution, a registry-only `class` per component, and an atomic `--split` that partitions the resolved componentRefs into one complete bundle set.

Key decisions: class is registry-only and stable (not a criteria axis — no recipe-shape or test-matrix growth); `ops` may depend on `core`, never the reverse, enforced by a resolved-graph guard; a plan/validate-once/render-subsets pipeline preserves ADR-015 profile locks for profiled recipes; the bundle set is the v1 provenance unit with create-only atomic staged publication (net-new for local output); deployer/output support is an explicit fail-closed matrix (Helm-local in v1); reconciliation-based app-of-apps deployment with no enforced ordering between bundles is supported, given a failed-sync retry policy so racing syncs converge (self-heal alone is not sufficient).

Fixes: N/A
Related: internal design discussion (2-bundle idea) — no tracking issue yet

## Type of Change

- [x] Documentation update

## Component(s) Affected

## Implementation Notes

Design-only PR; no code changes. The ADR includes a catalog prerequisite (repointing 38 `kube-prometheus-stack` dependency declarations across `gpu-operator`, `dynamo-platform`, `kubeflow-trainer` to `prometheus-operator-crds`) and an open question on the class taxonomy (two classes vs a third workload-specific class) to settle with the first consumer before the registry classification lands. Review discussion settled the working rubric — ops must be GPU-free and verifiable on CPU-only clusters — and classified nvsentinel as core under it.

## Testing

```bash
# Doc-only change (one new Markdown file under docs/design/, no code).
GOCACHE=<isolated> make lint   # PASS: golangci + yamllint + license headers,
                               # agents-sync, doc filename/MDX gates,
                               # chart-pin (bom-pinning) check
# make test / e2e / scan skipped per doc-only verification policy: they
# cannot regress from a new docs/design Markdown file. PR CI is green;
# the test/scan/GPU jobs are path-filtered and skip by design on
# docs-only changes.
```

## Risk Assessment

**Rollout notes:** N/A (design document; implementation phased per the ADR's plan)

## Checklist

### DIFF
--- a/docs/design/018-class-partitioned-bundles.md
+++ b/docs/design/018-class-partitioned-bundles.md
@@ -1,0 +1,646 @@
+# ADR-018: Class-Partitioned Bundles
+
+## Status
+
+**Proposed** — 2026-08-11.
+
+Builds on the resolved-recipe invariants from
+[ADR-015](015-recipe-configuration-profiles.md) (profile locks, strict
+boundary) and the component subset filter from
+[#1531](https://github.com/NVIDIA/aicr/issues/1531). It does not change the
+recipe resolution algorithm.
+
+## Problem
+
+AICR today emits exactly one bundle per resolved recipe. Every component —
+GPU runtime, networking, scheduling, and observability — deploys as a single
+unit with a single deployment order.
+
+Operators increasingly want to split that unit. The runtime core (components
+whose absence prevents workloads from being scheduled or run) and the ops
+stack (components whose absence mainly reduces observability, administration,
+or operational convenience) have different owners, different change-approval
+processes, and different release cadences. The need for co-existing bundles
+has surfaced in several contexts; the first concrete consumer needs a
+runtime bundle and an ops bundle that co-exist in the same cluster and are
+deployed and operated by different owners.
+
+The naive approach — author two recipes, or resolve the same recipe twice
+with different filters — reintroduces the exact failure the split must
+prevent: the two bundles can carry the *same component with different
+arguments*, and the collision surfaces only at deployment time, in the
+cluster, as a Helm release conflict or a silently diverged configuration.
+
+Two gaps block a safe split today:
+
+- Components have no classification. `recipes/registry.yaml` entries carry
+  identity, sourcing, scheduling, and validation fields, but nothing that
+  says which bundle a component belongs to.
+- The only subsetting primitive, the `bundlers` name filter (#1531), is an
+  ad hoc list on the SDK/REST surface. It has no CLI flag, no notion of a
+  named group, and no check that a set of subsets forms a complete,
+  non-overlapping partition of the recipe.
+
+## Goals
+
+- Classify every registry component into a named class, with `core` and
+  `ops` as the initial classes.
+- Generate one atomic **bundle set** — one separately deployable
+  sub-bundle (rendered as a class directory) per class, subject to
+  cross-class ordering prerequisites — from a **single resolved recipe**,
+  via a new `aicr bundle --split` flag.
+- Make same-component-different-arguments collisions impossible by
+  construction, not detected after the fact.
+- Preserve deployment-order correctness across sub-bundles: the per-class
+  order is the induced subsequence of the recipe's global topological order.
+- Make the bundle set auditable: it records the resolved recipe it came
+  from and the class of every component.
+- Keep ordinary generation unchanged: no `--split` still produces one
+  ordinary bundle, byte-identical relative to the post-prerequisite
+  catalog baseline (see Class dependency direction).
+
+## Non-Goals
+
+- Deploying more than one instance of the same component (for example two
+  Prometheus stacks). Component identity remains unique across the bundle
+  set; multi-instance support is future work.
+- Standalone per-class bundle artifacts that are independently verified,
+  attested, or published. In v1 the bundle set is the provenance and
+  verification unit; class directories can be invoked separately after
+  predecessor-class requirements are satisfied, but are not independently
+  *published*. A standalone class artifact with an externally-satisfied
+  profile-lock contract is follow-up work.
+- Cross-version compatibility between sub-bundles. AICR emits the set and
+  its ordering contract; verifying that an ops bundle from one generation
+  is compatible with a core bundle deployed from another is
+  operator-managed, and a deploy-time compatibility protocol is explicit
+  follow-up work.
+- OCI output for split generation. Multi-artifact naming and
+  partial-publish recovery are undefined; `--split` with OCI output is
+  rejected in v1.
+- Split generation for OCP recipes. Supported OCP recipes declare the
+  initial `ops` components but disable them all, so a split has nothing
+  to offer OCP until the `-ocp` monitoring variants are classified;
+  `--split` rejects all-disabled classes (see Partition validation).
+- Reassigning a component's class per recipe shape. Class is registry-only
+  and stable; moving a component between classes is a coordinated
+  migration, not a recipe-authoring decision.
+- Changing recipe resolution, overlay matching, or mixin composition. The
+  class is read at bundle time only.
+- A generic dependency solver or per-class release tooling.
+- A registry-level mutual-exclusion (`conflictsWith`) declaration. Today's
+  scattered checks (for example gpu-operator vs gpu-operator-ocp,
+  [#1685](https://github.com/NVIDIA/aicr/issues/1685))
+  would benefit from one, but that is a separate follow-up.
+- More than two shipped classes. The mechanism is N-ary by design (an
+  `apps` class for workload platform operators such as Kueue or Dynamo is
+  a plausible future), but this ADR ships `core` and `ops` only.
+
+## Decision
+
+### Resolve once, partition atomically
+
+The load-bearing invariant: a recipe is resolved exactly once — one criteria
+match, one overlay/mixin merge, one `RecipeResult` — and the split is a
+**partition of the resolved componentRefs**, applied at bundle time. Each
+enabled component belongs to exactly one class, and one sub-bundle
+therefore owns **everything emitted for that component** — its single
+instance, its RBAC objects, and all of its Helm releases (including
+synthetic `-pre`/`-post`/`-readiness` wrapper releases) in the
+component's target namespace — with the one merged value set the
+resolution produced. Target namespaces are shared surfaces, not owned:
+the initial taxonomy itself puts core's `prometheus-operator-crds` and
+the three ops components in `monitoring`.
+
+This is why collision detection is not needed for the case that motivated
+the design: with a single resolution there is no representation in which two
+bundles hold the same component with different arguments. If two classes
+both need a component, one class owns it and the other depends on it across
+the bundle boundary.
+
+The partition is **atomic**: `--split` always generates the complete bundle
+set, never a single class in isolation. Partial generation would let two
+invocations — at different catalog versions, or before and after a
+classification change — disagree about ownership of a component, recreating
+the collision across calls that the single resolution prevents within one.
+An operator who wants only the ops components still deploys only the `ops/`
+directory of a complete, verified set, after its `core/` prerequisites are
+installed.
+
+### Component class in the registry
+
+Each entry in `recipes/registry.yaml` gains an optional `class` field:
+
+```yaml
+- name: kube-prometheus-stack
+  displayName: Kube Prometheus Stack
+  class: ops
+  ...
+```
+
+- Allowed values are the known class names; the initial set is `core` and
+  `ops`. The class set is a **closed enumeration in Go**, not
+  registry-defined strings: an unenumerated class cannot be pre-validated
+  against the catalog, so new classes are added only in code, at a
+  release — cheap to extend, but the constraints are kept.
+- Omitted means `core`. This default is **availability-safe, not
+  classification-safe**: an unclassified component lands in the bundle
+  that must always be installed, but a component that *should* be `ops`
+  and omits the field is misclassified into `core` with no mechanical
+  signal beyond the external-replacement warning below.
+- The field is named `class`, not `type`, because `componentRefs[].type`
+  already means Helm-vs-Kustomize.
+- Class is **registry-only and stable**. Overlays cannot reassign it, and
+  it is not recorded in the resolved `RecipeResult` — so `recipe.yaml`,
+  checksums, and attestations of ordinary bundles are unchanged. Moving a
+  component between classes is a registry change treated as a coordinated
+  migration for clusters with a deployed split.
+- The default applies uniformly to **external registry entries** too: an
+  external `registry.yaml` entry replaces the embedded entry wholesale, so
+  a replacement for an `ops` component that omits `class` lands in `core`.
+  A replacement that intends `ops` must restate it; registry merging emits
+  a merge-time warning when an external entry replaces an embedded `ops`
+  component without declaring `class`.
+- External entries carry `class` like embedded ones, so an internal or
+  third-party catalog can classify its own components into the existing
+  classes **without any public change**. Only a brand-new class *name*
+  requires extending the Go enum — a one-value, release-time addition
+  made when a concrete consumer needs it.
+
+Initial classification: `kube-prometheus-stack`, `prometheus-adapter`, and
+`k8s-ephemeral-storage-metrics` are `ops`; everything else, including
+`nvsentinel` and `prometheus-operator-crds`, is `core`. The working
+rubric, sharpened by DGXC operational experience: **`ops` must be GPU-free
+and verifiable on CPU-only clusters**. `nvsentinel` is `core` under that
+test — its node remediation is part of the GPU runtime and must be tested
+with GPUs, and its dependencies (`cert-manager`, `gpu-operator`,
+`prometheus-operator-crds`) are already `core`.
+
+**`class` is not a criteria axis.** Criteria (`service`, `accelerator`,
+`intent`, `os`, `platform`) drive overlay matching, and every new axis
+multiplies the recipe-shape and test matrix. `class` deliberately does not
+participate in matching: it is a label read at bundle time, after
+resolution. The recipe count, the overlay files, and the resolved
+`RecipeResult` are identical whether or not `--split` is ever used, so the
+split adds surface linearly per class (bundle output and partition checks),
+not multiplicatively per recipe × class. The blast radius of introducing
+the field is correspondingly small: with `core` as the default, existing
+registry entries need no edits beyond the handful classified as `ops`, and
+generation without `--split` remains byte-identical relative to the
+post-prerequisite catalog baseline.
+
+### Class dependency direction
+
+Cross-class dependency edges must point one way: **`ops` may depend on
+`core`; `core` must not depend on `ops`.** The class graph is built from
+**cross-class edges only** — same-class dependencies are ignored at class
+level (they would otherwise collapse into self-loops) and remain governed
+by the existing component-level cycle validation. That cross-class graph
+must be acyclic. This keeps the deployment contract trivial (install core
+first, then ops) and gives partition validation a crisp rule to enforce.
+
+Direction is the floor, not the goal: classification should also
+**minimize cross-class edges** in the first place, because operating
+models exist where no ordering between bundles can be enforced at all. In
+the DGXC Runtime, for example, the runtime and operations stacks deploy as
+separate Argo CD app-of-apps with no enforceable ordering between them.
+The install-core-first contract therefore binds only the sequential helm
+`deploy.sh` path. In an externally managed, reconciliation-based model,
+each class converges independently and cut edges are not enforced — but
+convergence is not free: a sync that races a missing prerequisite (such
+as CRDs) fails rather than waits, so the operating model must configure a
+**failed-sync retry policy** with a retry window long enough for
+prerequisites to converge (in Argo CD, `syncPolicy.retry` — automated
+sync alone does not reattempt a failed sync of the same revision, and
+self-heal only repairs live drift, so it is a complement, not a
+substitute). Separate app-of-apps with no enforced ordering is a
+supported operating model **given such a retry policy**; note
+this describes externally managed deployments — v1's own generated
+output rejects the Argo/Flux deployers until their class-derived
+identities land (see the support matrix). The v1 bridge for such models:
+operators consume the helm-local bundle set with their own per-class
+Application/Kustomization wrappers; AICR-generated Argo/Flux split output
+is the deferred follow-up.
+
+Retry also has limits: it covers prerequisites whose absence makes
+reconciliation *fail* (such as missing CRDs); **fail-open prerequisites**
+— ops resources that apply successfully despite a missing non-CRD
+dependency — are not verified in v1, nor is cross-bundle version
+compatibility; a live set-identity or preflight protocol remains
+deferred.
+
+The existing catalog violates the rule, and fixing it is a prerequisite.
+Three core components declare dependencyRefs on `kube-prometheus-stack`
+(declared to guarantee the ServiceMonitor CRDs exist before their monitors
+are applied; the prerequisite PR verifies each edge's actual purpose
+before repointing it): `gpu-operator` in 26 declarations (including
+base), `dynamo-platform` in 8, and `kubeflow-trainer` in 4 (3 overlays
+plus 1 mixin, `recipes/mixins/platform-kubeflow.yaml`) — 38 declarations
+in total, **across overlays and mixins**. Because `dependencyRefs` merge
+additively across the composition chain, repointing only
+`recipes/overlays/base.yaml` fixes nothing; every declaration must be
+repointed to `prometheus-operator-crds`, which stays in `core`. CRDs are
+cluster API surface needed by both sides; keeping shared CRD packages in
+`core` is the general pattern.
+
+Dependency edges are **install-time requirements only**; runtime data
+flows need no edge and naturally run in the allowed direction. The
+Prometheus split is the canonical case: core components (such as
+`nvsentinel`) install monitor CRs, which requires only the
+`monitoring.coreos.com` CRDs — `prometheus-operator-crds`, a `core`
+component (`nvsentinel` already declares that direct edge; see
+[#928](https://github.com/NVIDIA/aicr/issues/928)) —
+while the ops Prometheus *scrapes* those monitors at runtime, an
+ops-consumes-core flow with no install edge at all. If ops is absent,
+core runs fully; its metrics are simply not collected yet, which is the
+definition of `ops`.
+
+The repoint prerequisite changes ordinary bundles **once**, independent of
+`--split`: `dependencyRefs` serialize into each affected recipe's emitted
+`recipe.yaml`, so the repoint is a one-time `recipe.yaml` and checksum
+change for every recipe that carried the old edge. All byte-identity
+claims in this ADR are relative to that post-prerequisite catalog
+baseline; the class field and split plumbing themselves change nothing in
+ordinary output.
+
+To keep the boundary fixed, a standing guard test validates the
+**effective dependency graph of every resolved catalog recipe** — not the
+raw overlay/mixin YAML — and fails on any `core` → `ops` edge or
+class-level cycle, in the spirit of the existing deployment-order guard
+tests.
+
+### Bundle generation with `--split`
+
+`aicr bundle` gains a boolean `--split` flag:
+
+```shell
+aicr bundle --recipe recipe.yaml --split --output ./bundles
+```
+
+- No `--split` keeps today's behavior: one bundle, all enabled
+  components.
+- `--split` takes no class arguments: it renders every class with
+  **declared components** in the resolved recipe, one directory per
+  class. A known class with no declared components is **skipped** (no
+  directory) — distinct from the rejected all-disabled case below — so
+  the flag is not hardcoded to two classes: adding a future class (for
+  example `apps` for workload platform operators) changes the output
+  layout of recipes that declare such components, and nothing else. A
+  class-subset selector is deliberately absent: emitting less
+  than the complete partition reintroduces the cross-call ownership
+  collisions this design exists to prevent (see the atomicity rationale
+  above), so per-class emission stays coupled to the deferred
+  standalone-class-artifact follow-up.
+- `--split` generates the complete bundle set in one invocation. The
+  pipeline is **plan, validate once, render subsets**: the complete
+  candidate is loaded, bundle-time overrides are applied, and all
+  validation — coherence, component validations, and the ADR-015 profile
+  lock — runs once against the full component set, exactly as today.
+  Only after validation passes is the partition rendered, one directory
+  per class, reusing the #1531 subset filter for rendering only.
+- This ordering is required, not incidental: the current pipeline filters
+  before `ValidateProfileLock`, so rendering a class through today's
+  pipeline would fail whenever another class holds a profile-locked
+  component. Validating the union first preserves ADR-015 for profiled
+  recipes (all AKS and GKE recipes) without weakening the lock.
+- Dependency edges cut by the partition use the existing
+  satisfied-externally semantics (the same path used when a declared
+  component is disabled).
+- Per-class deployment order is the induced subsequence of the recipe's
+  global `deploymentOrder`, so relative ordering is consistent across
+  sub-bundles by construction.
+
+### Partition validation
+
+Before writing any output, `--split` rejects:
+
+- a component assigned to an unknown class (assignment is total and
+  single-valued by construction: one registry field with a default);
+- a class-level dependency cycle, or any `core` → `ops` edge in the
+  resolved recipe;
+- a partition that separates components required to travel together. In
+  v1 the only such cohesion set is the AICR-provided Slurm accounting set
+  ([ADR-016](016-slurm-accounting-enablement.md)), known directly to
+  partition validation; a generic cohesion
+  declaration in the registry is deferred;
+- a class whose declared components are **all disabled** after recipe
+  disables and bundle-time overrides are applied (scalar
+  `--set <component>:enabled=false` is the supported bundle-time disable
+  mechanism; a class with no declared components at all is skipped, not
+  rejected — see above);
+- a caller-supplied `bundlers` component filter (#1531, SDK/REST). The
+  filter removes components before validation and would break the
+  complete-partition invariant or fail profiled recipes; in split mode the
+  class partition is the only supported subsetting, and the #1531 filter
+  is reused internally for rendering only;
+- disabled checksums — the set would be unverifiable before publication
+  (see Bundle set layout and provenance); and
+- an unsupported deployer or output combination (see below).
+
+The all-disabled rule matters because the existing subset filter treats an
+empty component list as "no filter" — naive reuse would render *every*
+enabled component into the empty class's directory. It also sets v1 scope
+honestly: supported OCP recipes declare the initial `ops` components but
+disable them all (`recipes/overlays/ocp.yaml`), so `--split` on OCP
+recipes — and any bundle-time override combination that disables a whole
+class — is rejected in v1. OCP split support is follow-up work that
+starts with classifying the `-ocp` monitoring variants.
+
+### Bundle set layout and provenance
+
+The **bundle set is the v1 provenance and verification unit.** The set root
+holds the full `recipe.yaml`, `split.yaml`, `checksums.txt`, and the
+attestation when enabled — covering the whole set. Class directories can be
+deployed separately after their predecessor-class requirements are satisfied,
+but are not independently verified or published artifacts.
+
+The set reuses the existing bundle checksum and attestation contract
+unchanged, applied at the set root: the same file inventory and exclusion
+rules as today's single bundle (with `split.yaml` and the class
+directories as ordinary payload), and the same attestation subject. For
+ordinary bundles, disabling checksums keeps its existing semantics; in
+**split mode, disabling checksums is rejected** — the bundle set is the
+provenance unit and pre-publication verification requires the root
+manifest, so a checksum-less split set would be unverifiable by
+construction.
+
+Whole-set verification begins at the root `checksums.txt`; when present,
+the attestation authenticates that manifest. `split.yaml` is never a
+trust root. Verification is a whole-set operation: verify the set, keep
+it intact, and run `<class>/deploy.sh` from within it. A class directory
+copied out of the set carries **no standalone provenance** — the
+exact-tree manifest cannot verify a detached subtree.
+
+**Publication is atomic on disk, not just in semantics, and create-only
+in v1.** The `--split` contract:
+
+- An explicit output path is required (`--output`, config, or a
+  non-empty SDK `OutputDir`) and the exact destination must be absent;
+  its parent may exist. A pre-existing destination is rejected before
+  any work — replacing a populated set is deliberately out of v1 scope.
+- The entire set — every class directory, checksums, and attestation —
+  renders and verifies in private staging on the **same filesystem** as
+  the destination (expected shape: a private sibling under the
+  destination's retained parent), then publishes with one atomic move to
+  the absent destination.
+- Concurrent attempts have at most one winner; losers fail without
+  altering the published set. A pre-commit failure leaves the
+  destination absent, with staging cleaned up; once publication has
+  committed, cleanup diagnostics cannot un-publish the set.
+- The guarantee holds at the functional API boundary (`pkg/bundler` and
+  the `pkg/client/v1` facade), not only in the CLI. The REST client
+  contract is unchanged — it already streams a complete ZIP or an error
+  — but the handler must pass a non-existent child of its private temp
+  directory to satisfy the create-only precondition, so the
+  implementation touches `pkg/server` as well.
+- Returned result paths reference the final published locations.
+
+This is **net-new work**: the existing closed-world staging machinery
+from [#1758](https://github.com/NVIDIA/aicr/issues/1758) is the pattern,
+but it runs only in the CLI and only for OCI output (for local output,
+`opts.ociRef == nil`, generation writes directly to the destination) —
+the mode `--split` rejects. Relocating staged publication into the
+functional API for local output is part of this design's scope, not an
+extension of existing behavior. Failure-path tests are specified with
+the implementation.
+
+`split.yaml` is **audit metadata only**: it records the set's
+`recipeDigest` — computed over the emitted set-root `recipe.yaml`, using
+the same canonicalization as `evidence/attestation.SubjectDigest` — the
+class of every component, and the class ordering contract (core before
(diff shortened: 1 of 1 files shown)


docs/design/018-composable-class-bundles.md
```diff
@@ -0,0 +1,412 @@
...
+
+**Publication is atomic on disk, not just in semantics.** Split generation
+renders the entire set — every class directory, checksums, and attestation
+— into a private staging workspace and exposes it at the output path only
+after the whole set is complete and verified. A failure while rendering a
+later class must not leave a deployable-looking partial set at the output
+path. This guarantee holds at the functional API boundary (`pkg/bundler`
+and the `pkg/client/v1` facade), not only in the CLI, so SDK and REST
+callers get the same contract. (Non-split generation is unchanged; today
+only OCI output stages.)
```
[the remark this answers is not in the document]
yuanchen8911: Partially addressed in 87007f416. The ADR now pins the contract's observable guarantees: atomicity holds at the functional API boundary (pkg/bundler and the pkg/client/v1 facade, so SDK and REST callers get the same behavior), and on failure staging is cleaned up and any pre-existing content at the output path is left untouched.

The remaining mechanics — staging location, destination-replacement semantics, and the failure-path test matrix — are deliberately deferred to the implementation PR. This ADR intentionally stays at requirement level for its first version; fully specifying rename primitives and test cases here would duplicate what the implementation PR must define and review anyway.

docs/design/018-composable-class-bundles.md
```diff
@@ -0,0 +1,412 @@
...
+    └── <component folders>
+```
+
+`nvsentinel`'s dependency on `prometheus-operator-crds` crosses the boundary
+in the allowed direction (`ops` → `core`) and is recorded as satisfied
+externally in the ops directory. The runtime owner deploys `core/`; the
+observability owner deploys `ops/` after it. Each directory is deployable
+and operable on its own from the generated set; taking a *newer* set's
+`ops/` against an older deployed `core/` is an operator decision that
+`split.yaml` makes auditable but v1 does not mechanically verify.
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 87007f416. The example now states the prerequisite explicitly: each class directory is separately deployable once its cross-class prerequisites are satisfied, and ops/ requires core/ to be installed first. It also answers the check-vs-assume question: in v1, ops/deploy.sh and the README state the prerequisite but assume it rather than querying the cluster — consistent with split.yaml being audit-only.

docs/design/018-composable-class-bundles.md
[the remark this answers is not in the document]
yuanchen8911: Agreed on both mechanism and scope. The current revision moves the guarantee to `DefaultBundler.Make` / `pkg/client/v1` and says failures preserve existing output, but it still explicitly defers staging location, destination replacement, cleanup, and failure-path tests to implementation. That leaves the contract incomplete despite the thread being auto-resolved as outdated. The follow-up ADR edit should specify reject-vs-replace behavior for an existing destination, same-filesystem sibling staging with a no-replace commit, cleanup/error semantics, and the SDK/REST success boundary and tests.

docs/design/018-composable-class-bundles.md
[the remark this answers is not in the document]
yuanchen8911: Agreed and addressed in `87007f4`. The example now states that the runtime owner deploys `core/`, the observability owner invokes `ops/deploy.sh` only after the prerequisites exist, and v1 does not preflight them—it assumes the operator follows the core-before-ops contract. No further change is needed for this comment.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,418 @@
...
+registry entries need no edits beyond the handful classified as `ops`, and
+generation without `--split` remains byte-identical to today.
+
+### Class dependency direction
+
+Cross-class dependency edges must point one way: **`ops` may depend on
+`core`; `core` must not depend on `ops`.** Collapsing component edges to
+class edges must yield an acyclic graph. This keeps the deployment contract
+trivial (install core first, then ops) and gives partition validation a
+crisp rule to enforce.
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 65c8922c2. The class graph is now defined from cross-class edges only; same-class dependencies are ignored at class level and remain governed by the existing component-level cycle validation.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,418 @@
...
+  `ops`.
+- Omitted means `core`. This is the fail-safe direction: an unclassified
+  component lands in the bundle that must always be installed.
+- The field is named `class`, not `type`, because `componentRefs[].type`
+  already means Helm-vs-Kustomize.
+- Class is **registry-only and stable**. Overlays cannot reassign it, and
+  it is not recorded in the resolved `RecipeResult` — so `recipe.yaml`,
+  checksums, and attestations of ordinary bundles are unchanged. Moving a
+  component between classes is a registry change treated as a coordinated
+  migration for clusters with a deployed split.
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in 65c8922c2. split.yaml's per-component class map is now explicitly the disambiguating record: two sets are equivalent only when both recipeDigest and the class map match, and a set generated before a class migration stays auditable through its recorded map. A separate registry digest adds no information beyond that map for this purpose, so we kept the map as the provenance.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,439 @@
...
+are implementation detail, specified with the implementation.)
+
+`split.yaml` is **audit metadata only**: it records the resolved recipe's
+`recipeDigest` — `sha256` of the canonicalized recipe, matching the recipe
+digest already used by bundle attestation (`pkg/evidence/attestation`) —
+the class of every component, and the class ordering contract (core before
+ops). The term "fingerprint" is avoided: AICR already uses it for cluster
+identity. `split.yaml` lets an operator answer "which recipe and partition
+produced what is deployed here" — it is not a deploy-time enforcement
+mechanism, and deployment does not reject on digest comparison.
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in 65c8922c2. recipeDigest is now specified as computed over the emitted set-root recipe.yaml (the post-override resolved recipe) using the same canonicalization as evidence/attestation.SubjectDigest, and the classification-drift point is covered by making the split.yaml class map the disambiguating record (see the provenance paragraph).

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,439 @@
...
+  class: ops
+  ...
+```
+
+- Allowed values are the known class names; the initial set is `core` and
+  `ops`.
+- Omitted means `core`. This is the fail-safe direction: an unclassified
+  component lands in the bundle that must always be installed.
+- The field is named `class`, not `type`, because `componentRefs[].type`
+  already means Helm-vs-Kustomize.
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in 65c8922c2. The registry section now states the default applies uniformly to external registry entries: a wholesale replacement of an ops component that omits class lands in core (the fail-safe direction), and a replacement intending ops must restate it.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,439 @@
...
+### Partition validation
+
+Before writing any output, `--split` rejects:
+
+- a component assigned to an unknown class (assignment is total and
+  single-valued by construction: one registry field with a default);
+- a class-level dependency cycle, or any `core` → `ops` edge in the
+  resolved recipe;
+- a partition that separates components required to travel together (for
+  example the AICR-provided Slurm accounting set from ADR-016);
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in 65c8922c2. v1's only cohesion set is the AICR-provided Slurm accounting set (ADR-016), known directly to partition validation; a generic cohesion declaration in the registry is explicitly deferred.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,441 @@
...
+— into a private staging workspace and exposes it at the output path only
+after the whole set is complete and verified. A failure while rendering a
+later class must not leave a deployable-looking partial set at the output
+path. This guarantee holds at the functional API boundary (`pkg/bundler`
+and the `pkg/client/v1` facade), not only in the CLI, so SDK and REST
+callers get the same contract. On failure, staging is cleaned up and any
+pre-existing content at the output path is left untouched. (Non-split
+generation is unchanged; today only OCI output stages. The remaining
+mechanics — staging location, destination replacement, failure-path tests —
+are implementation detail, specified with the implementation.)
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in 10f79bbfc. The failure guarantee is now scoped: a failure before publication leaves pre-existing content untouched with staging cleaned up, while post-commit cleanup diagnostics cannot un-publish the set — matching the semantics of the existing closed-world publication machinery from #1758, which the ADR now cites as the precedent the split extends to the API boundary. Returned result paths are specified to reference the final published locations.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+— into a private staging workspace and exposes it at the output path only
+after the whole set is complete and verified. A failure while rendering a
+later class must not leave a deployable-looking partial set at the output
+path. This guarantee holds at the functional API boundary (`pkg/bundler`
+and the `pkg/client/v1` facade), not only in the CLI, so SDK and REST
+callers get the same contract. A failure **before publication** leaves any
+pre-existing content at the output path untouched, with staging cleaned
+up; once publication has committed, cleanup diagnostics cannot un-publish
+the set. Returned result paths reference the final published locations, as
+today. (The CLI's existing closed-world generation machinery from #1758
```
njhensley: **🟠 Major — Atomic staged publication understates net-new cross-layer work**

The atomicity section cites #1758 as "already stages and publishes single bundles this way; the split extends that behavior to the API boundary." But that staged-publication path lives **only in `pkg/cli/bundle.go` and is gated on OCI output** — `pkg/bundler.Make` writes *directly* to the destination for local output, and `--split` **rejects** OCI. So the cited precedent covers exactly the mode split forbids; relocating atomic staged publication into `pkg/bundler` + `pkg/client/v1` for local directories is genuinely net-new, not an extension. Separately, atomically replacing a *populated multi-directory* destination is not a single `rename(2)` (non-empty-dir + cross-filesystem hazards), so deferring the swap to "implementation detail" hand-waves the load-bearing safety property.

**Fix:** Reframe as a functional-layer relocation (call it out in Implementation-Plan step 3) and state the atomicity preconditions — staging co-located on the same filesystem as the output; an explicit policy for a pre-existing populated destination — rather than deferring them.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+
+`aicr bundle` gains a boolean `--split` flag:
+
+```shell
+aicr bundle --recipe recipe.yaml --split --output ./bundles
+```
+
+- No `--split` keeps today's behavior: one bundle, all enabled
+  components.
+- `--split` takes no class arguments: it always renders every class
```
njhensley: **🟡 Minor — Empty-class semantics underspecified**

"renders every class **present**" (L249-250) sits awkwardly next to "reject a class with **zero enabled components**" (L282-283, L300-303). Reconcilable if "present" means "known & non-empty," but the doc never says which classes are subject to the empty-class rule — which muddies both the legitimate non-OCP *core-only recipe* case and the "N-ary by design — adding `apps` changes layout not contract" promise (L94-95, L251-252): a fixed known-class set + reject-empty would make a future `apps` class reject most recipes.

**Fix:** Pin the model once: are absent known classes skipped, or must all known classes be non-empty?

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+enabled component into the empty class's directory. It also sets v1 scope
+honestly: supported OCP recipes disable the entire initial `ops` set
+(`recipes/overlays/ocp.yaml`), so `--split` on OCP recipes — and any
+bundle-time override combination that empties a class — is rejected in v1.
+OCP split support is follow-up work that starts with classifying the
+`-ocp` monitoring variants.
+
+### Bundle set layout and provenance
+
+The **bundle set is the v1 provenance and verification unit.** The set root
```
njhensley, line 351: **🟡 Minor — Detached class subtree has no verifiable provenance — tie the limitation to the mechanism**

Root `checksums.txt` verification is **exact-tree** (confirmed against `validateExactTree` — fails on any missing/extra file), so an `ops/` dir copied out of the set cannot be verified against the set-root manifest even when byte-correct. The ADR defers standalone verified artifacts correctly and states class dirs "are not independently verified or published" (L313, L66-71) — but frames this as a *feature* non-goal rather than a *verification limitation* of the multi-owner / multi-host workflow that motivates the feature (L22-26, L429-431).

**Fix:** Add a sentence: verification is a whole-set operation — verify the set → keep it intact → run `<class>/deploy.sh` from within it; a detached class dir carries no provenance and must not be treated as verified.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+- name: kube-prometheus-stack
+  displayName: Kube Prometheus Stack
+  class: ops
+  ...
+```
+
+- Allowed values are the known class names; the initial set is `core` and
+  `ops`.
+- Omitted means `core`. This is the fail-safe direction: an unclassified
+  component lands in the bundle that must always be installed.
```
njhensley: **🟡 Minor — "Fail-safe to core" is availability-safe, not classification-safe**

Omitted-→-`core` (L138) + wholesale external-registry replacement (L146-150) is safe for *deployment availability* but silent for *classification integrity*: a GPU-free component that should be `ops` but omits `class` lands in `core` with no mechanical signal — the standing guard fires only on an actual `core → ops` edge (L234-237), and empty-class rejection won't fire if other ops components remain. The external-registry replacement path is exactly where an author most easily drops the field.

**Fix:** Soften the claim to "availability-safe, not classification-safe," and consider a positive assertion for external entries that replace a known-ops component.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+The install-core-first contract therefore binds only the sequential helm
+`deploy.sh` path. In an externally managed, reconciliation-based model,
+each class converges independently and cut edges are not enforced — but
+convergence is not free: a sync that races a missing prerequisite (such
+as CRDs) fails rather than waits, so the operating model must configure a
+**failed-sync retry policy** with a retry window long enough for
+prerequisites to converge (in Argo CD, `syncPolicy.retry` — automated
+sync alone does not reattempt a failed sync of the same revision, and
+self-heal only repairs live drift, so it is a complement, not a
+substitute). Separate app-of-apps with no enforced ordering is a
```
njhensley, line 216: **🟡 Minor — Argo app-of-apps "supported" vs deployer "Rejected" reads contradictory**

Separate app-of-apps is called "a supported operating model **given** a `syncPolicy.retry`" (L198-199) while the matrix **Rejects** `argocd`/`argocd-helm`/`flux` for `--split` in v1 (L388-390). The doc does reconcile it (L199-202: the model "describes externally managed deployments — v1's own generated output rejects the Argo/Flux deployers"), but the motivating DGXC consumer (L22-26) then has no first-class v1 *generated* path — it must hand-wire the `helm`-local `core/`/`ops/` set into its own Applications.

**Fix:** Add one sentence naming the bridge: the externally-managed model consumes the v1 helm-local set with operator-authored Application/Kustomization wrappers; AICR-generated Argo/Flux split output is the deferred follow-up.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+  deliberately barred from changing component presence ("component presence
+  is not affected by reselection"), and a profile expresses one selected
+  variant, not a partition into co-existing sets.
+- **Caller-defined partitions (bare `bundlers` name lists or arbitrary
+  per-invocation cuts), no class field.** Rejected: callers would
+  hand-maintain component lists that drift as recipes evolve; nothing
+  would check completeness, overlap, or dependency direction; and two
+  invocations could cut the DAG differently, producing co-existing
+  bundles that claim the same component. A registry-fixed cut also has a
+  property no per-invocation cut can: CI validates the entire catalog
```
njhensley, line 579: **🟡 Minor — "CI validates empty classes before any user runs bundle" overstates**

The Alternatives section credits the registry-fixed cut with letting "CI validate the entire catalog … (direction, cohesion, **empty classes**) before any user runs `bundle`" (L493-494). But empty-class is defined as zero-enabled "after recipe disables **and bundle-time overrides**" (L282-283, L302-303). Catalog CI has no per-invocation `--set`/override context, so it can only catch recipe-shape empties (OCP), never override-induced ones — which the ADR itself pushes to bundle-time partition validation (L289).

**Fix:** Qualify to "direction, cohesion, and *recipe-shape* empty classes; override-induced empty classes are enforced at bundle time."

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+supported operating model **given such a retry policy**; note
+this describes externally managed deployments — v1's own generated
+output rejects the Argo/Flux deployers until their class-derived
+identities land (see the support matrix).
+
+The existing catalog violates the rule, and fixing it is a prerequisite.
+Three core components declare dependencyRefs on `kube-prometheus-stack`
+(to guarantee the ServiceMonitor CRDs exist before their monitors are
+applied): `gpu-operator` in 26 overlay declarations
+(including base), `dynamo-platform` in 8, and `kubeflow-trainer` in 4 — 38
```
njhensley: **🟡 Minor — 4th kubeflow edge is in a mixin — "overlay declarations" misleads the repoint**

Verified: the 38 core→`kube-prometheus-stack` edges are gpu-operator 26 + dynamo-platform 8 + kubeflow-trainer 4, and the 4th kubeflow edge lives in `recipes/mixins/platform-kubeflow.yaml`, **not** `recipes/overlays/`. The ADR says "overlay declarations" (L208-209), "raw overlay YAML" (L235), and "repoint all 38" (L527) — an implementer scoping the prerequisite PR to `recipes/overlays/` finds 37 and leaves one `core → ops` mixin edge (the resolved-graph guard would eventually catch it, so not a blocker).

**Fix:** Reword to "declarations across overlays **and mixins**," and note the kubeflow split (3 overlay + 1 mixin).

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+stays in `core`. CRDs are cluster API surface needed by both sides; keeping
+shared CRD packages in `core` is the general pattern.
+
+Dependency edges are **install-time requirements only**; runtime data
+flows need no edge and naturally run in the allowed direction. The
+Prometheus split is the canonical case: core components (such as
+`nvsentinel`) install monitor CRs, which requires only the
+`monitoring.coreos.com` CRDs — `prometheus-operator-crds`, a `core`
+component (`nvsentinel` already declares that direct edge; see #928) —
+while the ops Prometheus *scrapes* those monitors at runtime, an
```
njhensley, line 253: **🔵 Nitpick — Bare issue/ADR refs don't hyperlink**

`#1531` and `[ADR-015]` hyperlink, but `#928` (L221), `#1758` (L337), `#1685` (L92), and `(ADR-016)` (L288) stay bare — siblings 015/016 hyperlink first mention.

**Fix:** Hyperlink the first mention of each so readers can navigate.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+self-heal only repairs live drift, so it is a complement, not a
+substitute). Separate app-of-apps with no enforced ordering is a
+supported operating model **given such a retry policy**; note
+this describes externally managed deployments — v1's own generated
+output rejects the Argo/Flux deployers until their class-derived
+identities land (see the support matrix).
+
+The existing catalog violates the rule, and fixing it is a prerequisite.
+Three core components declare dependencyRefs on `kube-prometheus-stack`
+(to guarantee the ServiceMonitor CRDs exist before their monitors are
```
njhensley: **🔵 Nitpick — Repoint assumes all 38 edges are CRD-ordering edges**

The repoint asserts all 38 edges guarantee ServiceMonitor CRDs (L206-208), but only `nvsentinel` and `k8s-ephemeral-storage-metrics` carry the documented #928 CRD-race rationale in `base.yaml`; the `gpu-operator`/`dynamo`/`kubeflow` edges are uncommented. The CRD-only pattern is almost certainly correct (a ServiceMonitor CR needs only the CRD to apply).

**Fix:** Add a note to confirm each edge's intent during the prerequisite PR / smoke-test ordering post-repoint.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+bundles hold the same component with different arguments. If two classes
+both need a component, one class owns it and the other depends on it across
+the bundle boundary.
+
+The partition is **atomic**: `--split` always generates the complete bundle
+set, never a single class in isolation. Partial generation would let two
+invocations — at different catalog versions, or before and after a
+classification change — disagree about ownership of a component, recreating
+the collision across calls that the single resolution prevents within one.
+An operator who wants only the ops components still deploys only the `ops/`
```
njhensley, line 126: **🔵 Nitpick — "deploys only the ops/ directory" reads standalone**

L121-122 reads as if `ops/` were independently deployable; core-first is clarified only later at L429-431.

**Fix:** Add "…against a cluster where `core` is already installed."

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+pre-existing content at the output path untouched, with staging cleaned
+up; once publication has committed, cleanup diagnostics cannot un-publish
+the set. Returned result paths reference the final published locations, as
+today. (The CLI's existing closed-world generation machinery from #1758
+already stages and publishes single bundles this way; the split extends
+that behavior to the API boundary. The remaining mechanics — staging
+location, destination replacement, failure-path tests — are implementation
+detail, specified with the implementation.)
+
+`split.yaml` is **audit metadata only**: it records the set's
```
njhensley, line 407: **🔵 Nitpick — recipeDigest naming invites same-digest⇒same-partition misread**

The ADR pre-empts this well (L352-356), but `recipeDigest` on an identity field inside `split.yaml` invites an auditor comparing it in isolation to conclude "same digest ⇒ same partition."

**Fix:** Optional schema-doc note that `recipeDigest` identifies the resolved recipe only; partition identity requires the class map.

njhensley (commented):
## Multi-persona review — ADR-018 class-partitioned bundles

**Method:** 4 independent persona reviewers (Domain/Architecture · Correctness/Consistency · Supply-chain/Provenance · Docs) fanned out, then a senior meta-reviewer re-derived every finding against the resolved doc **and code** (`pkg/bundler`, `pkg/cli/bundle.go`, `pkg/bundler/checksum`, the overlays/mixins). Findings anchored to head `9bf22aa`.

### Overall assessment

Strong, unusually well-researched **Proposed** ADR whose load-bearing invariants hold up against the actual code: the *filter-before-`ValidateProfileLock`* ordering that makes "validate-once, render-subsets" **required** (not incidental) is real; the induced-subsequence per-class ordering falls out of the existing filter; the full-set synthetic `-pre`/`-post` identity check closes a real cross-class collision hole; and the Prometheus repoint scope (**38** = 26 gpu-operator + 8 dynamo-platform + 4 kubeflow-trainer, ops→ops edges correctly excluded) reproduces exactly. The major v1 punts (OCI, standalone verified artifacts, cross-version compat, `conflictsWith`) are correctly deferred and mostly flagged honestly.

**No blockers.** One Major (the atomic-publication framing) is worth fixing before the implementation PRs land, plus six Minor clarity/accuracy items and a sweep of nitpicks — all inline below.

### Confirmed non-issues (examined, not flagged)

- **"byte-identical" is not stated once** — it's qualified "relative to the post-prerequisite catalog baseline" at L58-59, L171-172, L230, L519. Well-anchored; a persona finding here was **refuted**.
- **Validate-once ordering** (L261-269): correct against `bundler.go` (filter precedes `ValidateProfileLock`).
- **Class never enters `RecipeResult`/`recipe.yaml`** → no leak into ordinary output; ordinary bundles' digests/attestations genuinely unchanged.
- **`split.yaml` integrity**: transitively covered by `checksums.txt` — not a provenance hole.
- **Full-set synthetic-release identity check**: addresses a real cross-class collision (verified in the Helm writer).

### Tier table

| 🔴 Blocker | 🟠 Major | 🟡 Minor | 🔵 Nitpick | Refuted |
|---|---|---|---|---|
| 0 | 1 | 6 | 9 | 1 |

**Recommendation: 🟢 Approve with comments** — a solid design doc; tighten the atomic-publication framing (C2) and the "overlay-only" repoint phrasing (C7) before the implementation PRs, and sweep the nitpicks opportunistically.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+
+### Bundle set layout and provenance
+
+The **bundle set is the v1 provenance and verification unit.** The set root
+holds the full `recipe.yaml`, `split.yaml`, `checksums.txt`, and the
+attestation when enabled — covering the whole set. Class directories can be
+deployed separately after their predecessor-class requirements are satisfied,
+but are not independently verified or published artifacts.
+
+The set reuses the existing bundle checksum and attestation contract
```
njhensley, line 357: **🔵 Nitpick — split.yaml is not a standalone trust root**

`split.yaml` (and its embedded `recipeDigest`) is integrity-protected transitively via `checksums.txt`, but is not an independent trust root — a `split.yaml`-first verification shortcut would be fooled by a consistently rewritten recipe+split+tree triple.

**Fix:** One sentence: provenance verification always goes through `checksums.txt` + attestation, never `split.yaml` in isolation.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+  validation — coherence, component validations, and the ADR-015 profile
+  lock — runs once against the full component set, exactly as today.
+  Only after validation passes is the partition rendered, one directory
+  per class, reusing the #1531 subset filter for rendering only.
+- This ordering is required, not incidental: the current pipeline filters
+  before `ValidateProfileLock`, so rendering a class through today's
+  pipeline would fail whenever another class holds a profile-locked
+  component. Validating the union first preserves ADR-015 for profiled
+  recipes (all AKS and GKE recipes) without weakening the lock.
+- Dependency edges cut by the partition use the existing
```
njhensley, line 306: **🔵 Nitpick — satisfied-externally cut edges fail-open on non-CRD prereqs**

Cut cross-class edges + no v1 deploy-time preflight (L432-436) means `ops/` trusts an unverified, unbound `core/` — honestly flagged. CRD-shaped prereqs fail-closed at the apiserver, but non-CRD prereqs and version skew don't.

**Fix:** When describing the deferred compatibility protocol, consider naming the missing primitive: a shared set-identity token a live `core/` could present and a deploying `ops/` require.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+  materialize, are follow-up work with an explicit migration story.
+
+## Consequences
+
+- Adding a component now includes deciding its class; omission safely
+  defaults to `core`.
+- The `core`/`ops` boundary becomes a checked contract: the resolved-graph
+  guard fails CI on a new `core` → `ops` dependency instead of letting it
+  break split deployments.
+- 38 existing dependency declarations (`gpu-operator`, `dynamo-platform`,
```
njhensley: **🔵 Nitpick — Hard-coded "38" in three places is a drift hazard**

The exact counts appear at L208, L512-513, and L527 and drive a concrete step; every new GPU-model overlay redeclaring the edge silently makes the prose wrong (the authoritative check is the guard test).

**Fix:** Keep one "as of 2026-08-11" snapshot and phrase the step around "every `kube-prometheus-stack` dependency edge (currently 38)."

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+- The only subsetting primitive, the `bundlers` name filter (#1531), is an
+  ad hoc list on the SDK/REST surface. It has no CLI flag, no notion of a
+  named group, and no check that a set of subsets forms a complete,
+  non-overlapping partition of the recipe.
+
+## Goals
+
+- Classify every registry component into a named class, with `core` and
+  `ops` as the initial classes.
+- Generate one atomic **bundle set** — one separately deployable sub-bundle
```
njhensley: **🔵 Nitpick — Three terms for the per-class unit**

"sub-bundle" (L48/65/273/366), "class directory" (L68/313), and "class tree" (L320/347) all name the per-class artifact.

**Fix:** Pick one on-disk term (e.g. "class directory") and reserve "sub-bundle" for the deployable-unit concept; optionally add a one-line vocabulary note.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,551 @@
...
+claims in this ADR are relative to that post-prerequisite catalog
+baseline; the class field and split plumbing themselves change nothing in
+ordinary output.
+
+To keep the boundary fixed, a standing guard test validates the
+**effective dependency graph of every resolved catalog recipe** — not the
+raw overlay YAML — and fails on any `core` → `ops` edge or class-level
+cycle, in the spirit of the existing deployment-order guard tests.
+
+### Bundle generation: `--split`
```
njhensley: **🔵 Nitpick — Heading slug triple-hyphen**

`### Bundle generation: \`--split\`` (L239) yields a triple-hyphen slug (`bundle-generation---split`) — the exact pattern the repo's doc-style rules warn against.

**Fix:** Use "### Bundle generation with --split" or "### The --split flag."

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Confirmed and fixed in c61ee6b72 — you're right on both counts. The #1758 staging machinery is gated on OCI output (bundleTarget is nil for local output, where generation writes directly to the destination), so it was cited as precedent for exactly the mode --split rejects. The ADR now cites it as the pattern only, states the relocation to pkg/bundler/pkg/client/v1 is net-new scope, and defuses the populated-multi-directory replace hazard by decision: v1 --split is create-only — explicit output path required (--output, config, or non-empty SDK OutputDir), exact destination must be absent (parent may exist), same-filesystem staging, one atomic move, at most one winner under concurrency, pre-commit failure leaves the destination absent. The REST client contract is unchanged (already complete-ZIP-or-error); the handler passes a non-existent child of its temp dir to satisfy the precondition, so pkg/server is in the implementation scope.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — the model is now pinned: a known class with zero declared components in the resolved recipe is skipped (no directory), which keeps the N-ary promise (a future apps class doesn't reject non-apps recipes); a class whose declared components are all disabled (recipe disables or bundle-time overrides) is rejected — which is the OCP case, and the OCP paragraph now says 'declares but disables' rather than 'disables the set'.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — added to the provenance section: whole-set verification begins at the root checksums.txt (when present, the attestation authenticates that manifest); verification is a whole-set operation — verify the set, keep it intact, and run <class>/deploy.sh from within it; a class directory copied out of the set carries no standalone provenance since the exact-tree manifest cannot verify a detached subtree.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — the default is now described as availability-safe, not classification-safe, with the silent-misclassification risk stated. For the external-registry path (where the field is most easily dropped), registry merging emits a merge-time warning when an external entry replaces an embedded ops component without declaring class — merge-time rather than a surfaced validation warning, since mergeRegistries has both entry sets in hand and needs no new diagnostic plumbing.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — added the bridge sentence: the externally-managed model consumes the v1 helm-local bundle set with operator-authored per-class Application/Kustomization wrappers; AICR-generated Argo/Flux split output is the deferred follow-up. So the DGXC-style consumer has a named v1 path, and 'supported operating model' vs the rejected generated deployers no longer reads contradictory.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — qualified as suggested: CI validates direction, cohesion, and recipe-shape all-disabled classes; override-induced cases are enforced at bundle time (catalog CI has no per-invocation override context).

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — reworded to 'across overlays and mixins' with the kubeflow-trainer split called out (3 overlays + 1 mixin, recipes/mixins/platform-kubeflow.yaml), and the guard-test sentence now says raw overlay/mixin YAML. Good catch — an implementer scoping to recipes/overlays/ would have found 37.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — #928, #1685, #1758, and ADR-016 are now full links like #1531 and ADR-015.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Addressed in c61ee6b72 — the parenthetical now says the edges were declared for ServiceMonitor-CRD ordering and that the prerequisite PR verifies each edge's actual purpose before repointing it, rather than asserting all 38 share one purpose.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — now reads 'deploys only the ops/ directory of a complete, verified set, after its core/ prerequisites are installed.'

docs/design/018-class-partitioned-bundles.md
yuanchen8911: No change — agreed the misread is plausible, but the ADR pre-empts it explicitly (recipeDigest + class map identify the resolved recipe and partition, not the rendered payload; two invocations can share a recipeDigest yet render different class trees), and the name matches the existing evidence/attestation.SubjectDigest recipe-identity contract. Renaming would decouple it from that contract for little gain.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — stated directly: whole-set verification begins at the root checksums.txt; when present, the attestation authenticates that manifest; split.yaml is never a trust root.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Addressed in c61ee6b72 with sharper wording than my first attempt: retry covers prerequisites whose absence makes reconciliation fail (missing CRDs); fail-open prerequisites — ops resources that apply successfully despite a missing non-CRD dependency — are not verified in v1, nor is cross-bundle version compatibility; a live set-identity/preflight protocol remains deferred.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — the count and per-component breakdown now live only in Class dependency direction; Consequences and the implementation plan say 'all' and reference that section.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Addressed in c61ee6b72 — defined once at first use: 'sub-bundle (rendered as a class directory)'. Kept the register distinction in running text (sub-bundle for the logical artifact, class directory for the filesystem form) since full unification makes several sentences read worse.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Fixed in c61ee6b72 — heading renamed to 'Bundle generation with --split', avoiding the triple-hyphen slug.

njhensley (dismissed):
## Re-review — prior feedback addressed ✅

Follow-up to my earlier multi-persona review (16 findings against `9bf22aad`). The revision (`e85ab17f`) addresses **14 of 16** findings, with the remaining 2 being cosmetic nitpicks I would not block on.

**Highlights**
- **Atomic publication (the Major):** reworked into an explicit *create-only* contract — destination-must-be-absent, same-filesystem private staging, one atomic move, concurrent-winner semantics, `pkg/server` accommodation — and now states plainly that relocating staged publication into `pkg/bundler`/`pkg/client/v1` for local output is **net-new work, not an extension** of the CLI/OCI-only #1758 path. Stronger than requested.
- **Empty-class semantics:** cleanly split into *no declared components → skipped (no directory)* vs *declared-but-all-disabled → rejected*, which also repairs the N-ary extensibility claim.
- **Provenance:** added the whole-set verification statement (a detached class dir carries no standalone provenance under the exact-tree manifest), the merge-time warning for external `ops` replacements, and a named deferred set-identity/preflight protocol for fail-open prerequisites.
- **Repoint scope:** corrected to "38 declarations **across overlays and mixins**" (the 4th kubeflow edge lives in `recipes/mixins/platform-kubeflow.yaml`), and the count is no longer hard-coded in three places.

**Remaining (cosmetic, non-blocking):** an optional `recipeDigest` schema note (risk already well pre-empted), and "class tree" still coexisting with "class directory."

No net-new issues introduced by the revision. **Approving.**

<sub>Re-reviewed with a multi-persona + adversarial meta-review pass.</sub>

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Follow-up: took your optional suggestion after all in bc82265e2 — the ADR now states the split.yaml schema (defined with the implementation) restates the digest-scope note alongside the recipeDigest field.

docs/design/018-class-partitioned-bundles.md
yuanchen8911: Follow-up: the last straggler ('class trees' in the provenance section) is now 'class directories' in bc82265e2 — terminology fully unified.

yuanchen8911:
@coderabbitai review

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,630 @@
...
+- an unsupported deployer or output combination (see below).
+
+The all-disabled rule matters because the existing subset filter treats an
+empty component list as "no filter" — naive reuse would render *every*
+enabled component into the empty class's directory. It also sets v1 scope
+honestly: supported OCP recipes declare the initial `ops` components but
+disable them all (`recipes/overlays/ocp.yaml`), so `--split` on OCP
+recipes — and any bundle-time override combination that disables a whole
+class — is rejected in v1. OCP split support is follow-up work that
+starts with classifying the `-ocp` monitoring variants.
```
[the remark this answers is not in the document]
yuanchen8911: The premise doesn't hold: only typed overrides targeting enabled are rejected. Scalar --set <component>:enabled=false is the supported bundle-time disable mechanism (see docs/user/bundling.md, 'Enable or disable components' — disabling is allowed, re-enabling a recipe-disabled component is rejected). So the all-disabled-via-overrides state is reachable through supported input and the validation case is not unreachable. Clarified in 7fd761609 by naming the scalar mechanism in the partition-validation bullet.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,630 @@
...
+attestation when enabled — covering the whole set. Class directories can be
+deployed separately after their predecessor-class requirements are satisfied,
+but are not independently verified or published artifacts.
+
+The set reuses the existing bundle checksum and attestation contract
+unchanged, applied at the set root: the same file inventory and exclusion
+rules as today's single bundle (with `split.yaml` and the class
+directories as ordinary payload), and the same attestation subject. Disabling checksums
+remains allowed with the existing option semantics and simply yields a set
+without provenance, exactly as for a single bundle.
```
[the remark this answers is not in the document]
yuanchen8911: Fixed in 7fd761609 — real contradiction. Split mode now rejects disabled checksums: the bundle set is the provenance unit and pre-publication verification requires the root manifest, so a checksum-less split set would be unverifiable by construction. Ordinary (non-split) bundles keep the existing option semantics.

docs/design/018-class-partitioned-bundles.md
```diff
@@ -0,0 +1,630 @@
...
+
+Component-level identities need one guard beyond unique component names:
+*synthetic* release names. The Helm writer injects `<name>-pre` /
+`<name>-post` wrapper releases and today detects collisions against the
+complete component list, so per-class rendering could miss a collision
+split across classes (a core component's injected wrapper vs an ops
+component's name — reachable with external component registries). Emitted
+identity validation (namespace, release name, including synthetic wrappers
+and reserved suffixes) therefore runs against the **full set** before
+subsets are rendered, consistent with the validate-once pipeline.
```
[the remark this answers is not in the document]
yuanchen8911: Clarified in 7fd761609: split identity validation covers deployer-level identities only (namespace, release names including synthetic wrappers and reserved suffixes). Collisions between rendered Kubernetes objects are pre-existing single-bundle behavior, unchanged by partitioning, and deliberately out of this ADR's scope.

mchmarny (approved):
Approved — 0 blockers. Re-reviewed the exact head: all 16 prior findings are addressed. The ADR preserves one resolved recipe, opt-in/non-split parity, whole-set provenance, and fail-closed v1 scope; the remaining class-taxonomy question is correctly held as a pre-implementation decision while the ADR remains Proposed. Focused `make lint` and current CI pass; all review threads are resolved.

PR #2169, opened 2026-08-11
feat(slinky-slurm): add configurable Enroot settings and update Slinky CUJ with epilog/prolog examples

kaynetu:
## Summary

Adds AICR-managed, overridable Enroot configuration and environment defaults for Slinky Slurm login and compute containers. Adds canonical operator documentation for bundle-time Enroot overrides, job-level environment precedence, safe rollout on existing clusters, and `--set-file` usage for Slurm prolog and epilog scripts.

## Motivation / Context

Slinky Pyxis workloads need consistent Enroot runtime configuration on both login and `slurmd` containers. Previously, operators had to rely on image defaults or manually create and mount configuration resources; this change provides a recipe-native baseline while preserving bundle-time site overrides.

Fixes: N/A
Related: N/A

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: Slinky component values, Slurm overlays, and Chainsaw coverage

## Implementation Notes

- Adds provider-neutral `enroot.config` and `enroot.env` maps to the Slinky component values. The shipped Enroot configuration is limited to the intentional cluster policy defaults `ENROOT_MOUNT_HOME=yes` and `ENROOT_REMAP_ROOT=yes`; other Enroot paths retain their image defaults and sites can add overrides at bundle time.
- Sets `NCCL_DEBUG=WARN` as a cluster-wide Pyxis workload default. An `enroot.env` value takes precedence over an ordinary job export of the same name; a job override must export the value and name the variable with `srun --container-env=<VAR>`.
- Renders the maps into a pre-install ConfigMap as `/etc/enroot/enroot.conf` and `/etc/enroot/environ.d/99-aicr-defaults.env`, then mounts both files into the default LoginSet and NodeSet containers.
- Registers the Enroot pre-manifest on every Slurm leaf. The pre-folder now always exists because Enroot always renders; conditional shared-storage PVC templates remain omitted while storage is disabled, and the GB200 leaf retains its IMEX ComputeDomain in the same pre-folder.
- Adds a canonical Slinky Slurm Enroot guide and links it from the user index and component catalog. The guide documents the `slinkyslurm:enroot.env.*` and `slinkyslurm:enroot.config.*` override surfaces, requires keys and values to be single-line, and recommends keeping fabric- or workload-specific NCCL settings job-scoped. It also documents NodeSet and `slurmctld` prolog/epilog maps through `--set-file`.
- Documents that the generated `slinky-slurm-pre` release owns the Enroot ConfigMap consumed by the main `slinky-slurm` release, so the two releases must be deployed, upgraded, rolled back, and removed together as one generated bundle.
- Regenerates the committed image inventory, including current upstream BusyBox digest drift discovered by `make bom-docs`.

## Testing

```bash
unset GITLAB_TOKEN && GOMAXPROCS=4 make qualify
GOFLAGS="-mod=vendor" go test -coverprofile="$TMPDIR/aicr-recipe-pr-current.out" ./pkg/recipe/...
make bom-docs
```

- Post-rebase `make qualify`: **PASS**, including race-enabled tests, full lint, tuning checks, E2E, vulnerability scan, license checks, and API compatibility.
- Project coverage: **82.6%** (80% threshold).
- Chainsaw E2E: **24 passed / 0 failed / 0 skipped**.
- `pkg/recipe/...` coverage: **89.0%**.
- Affected-package lint: **0 issues**.
- Enroot tests cover every Slurm leaf, default config/environment values, exactly one LoginSet and NodeSet mount per Enroot path with the expected volume name and subPath, and rendered ConfigMap identity/content. The Slinky storage Chainsaw scenario exercises the complete bundle-time override path through `aicr bundle --set`, including `slinkyslurm` alias resolution and nested Enroot values.
- Manual deployment checks were run on both a local Kind cluster and an EKS GB200 cluster, covering generated Enroot configuration and environment delivery to Slinky workloads.

## Risk Assessment

**Rollout notes:** Non-Slurm recipes are unaffected. Existing Slurm bundles gain an ordered Enroot ConfigMap pre-release plus two read-only `subPath` mounts on login and compute containers. Kubernetes does not refresh `subPath` mounts in place, and a ConfigMap-only change does not alter the LoginSet or NodeSet pod templates. After changing the ConfigMap, drain all Slurm nodes and wait for active jobs to finish before deleting worker pods; replace workers one at a time, then notify users and replace login pods. Return the workers to service only after all replacements are healthy and registered with Slurm. Deploy and manage the generated `slinky-slurm-pre` and `slinky-slurm` releases together; do not prune or uninstall the pre-release while the main release still consumes its ConfigMap. Sites should review the home-mount and root-remap defaults and override them where local security policy differs.

## Checklist


recipes/components/slinky-slurm/values.yaml
```diff
@@ -142,6 +142,25 @@ storage:
...
+    ENROOT_RUNTIME_PATH: /run/enroot/user-$(id -u)
+    ENROOT_CACHE_PATH: ${HOME}/.cache/enroot
+    ENROOT_DATA_PATH: ${HOME}/.local/share/enroot
+    ENROOT_TEMP_PATH: /tmp
+    ENROOT_CONFIG_PATH: ${HOME}/.config/enroot
+    ENROOT_MOUNT_HOME: "yes"
+    ENROOT_REMAP_ROOT: "yes"
+    ENROOT_HOOKS_PATH: /etc/enroot/hooks.d
+  env:
+    NCCL_DEBUG: WARN
```
mchmarny, line 159: **Blocking.** Shipping `NCCL_DEBUG` here makes it impossible to override from a job script — a behavior regression versus the current images.

Anything in `enroot.env` is written to `/etc/enroot/environ.d/99-networking.env`, which enroot folds into the *container's own* environment, and Pyxis then pushes container-side values back over the Slurm job environment. From Pyxis's own help text for `--container-env`:

> By default, all exported host environment variables are set in the container after the entrypoint is run, **but their existing values in the image take precedence**; the variables specified with this flag are preserved.

Mechanism:
- enroot [`src/runtime.sh:19,44-48`](https://github.com/NVIDIA/enroot/blob/master/src/runtime.sh#L19) — `environ_dirs=("${ENROOT_SYSCONF_PATH}/environ.d" …)`, every `*.env` appended into the container envfile.
- pyxis `pyxis_slurmstepd.c` → `spank_import_container_env()` calls `spank_setenv(sp, key, value, overwrite=1)` for every container variable **not** named in `--container-env`.

**Failure scenario:** user runs `export NCCL_DEBUG=INFO; srun --container-image=… app`. Before this PR no `environ.d/*.env` existed, so `INFO` reached the application. After this PR the container carries `NCCL_DEBUG=WARN`, which wins, and the user gets `WARN` with no indication why. The only escape hatch is `--container-env=NCCL_DEBUG`, which this PR does not document.

This also makes the guidance added at [`demos/cuj1-slinky-slurm.md#L173-L177`](https://github.com/NVIDIA/aicr/blob/135682c7f197feedd36767703bc60584435101eb/demos/cuj1-slinky-slurm.md#L173-L177) incorrect — "exported by the job script or passed with `srun --export`" does not override an `enroot.env` entry.

**Minimum correction:** either ship `enroot.env` empty (the mount stays in place for sites to populate), or keep `NCCL_DEBUG: WARN` and state in the doc section that `enroot.env` entries take precedence over job-script exports and require `--container-env=<VAR>` to override. Either way the `srun --export` sentence needs fixing. If you go with the empty map, please confirm the template still emits valid YAML for `99-networking.env: |` with no body.

recipes/components/slinky-slurm/values.yaml
```diff
@@ -142,6 +142,25 @@ storage:
...
+    ENROOT_LIBRARY_PATH: /usr/lib/enroot
+    ENROOT_SYSCONF_PATH: /etc/enroot
+    ENROOT_RUNTIME_PATH: /run/enroot/user-$(id -u)
+    ENROOT_CACHE_PATH: ${HOME}/.cache/enroot
+    ENROOT_DATA_PATH: ${HOME}/.local/share/enroot
+    ENROOT_TEMP_PATH: /tmp
+    ENROOT_CONFIG_PATH: ${HOME}/.config/enroot
+    ENROOT_MOUNT_HOME: "yes"
+    ENROOT_REMAP_ROOT: "yes"
+    ENROOT_HOOKS_PATH: /etc/enroot/hooks.d
```
mchmarny: Non-blocking: `ENROOT_HOOKS_PATH` is not an enroot configuration variable — this key is inert.

The string appears nowhere in enroot: not in `enroot.in`, `src/common.sh`, `src/runtime.sh`, `src/docker.sh`, `src/bundle.sh`, or `conf/enroot.conf.in`. Hook directories are hardcoded at [`src/runtime.sh:17`](https://github.com/NVIDIA/enroot/blob/master/src/runtime.sh#L17):

```sh
readonly hook_dirs=("${ENROOT_SYSCONF_PATH}/hooks.d" "${ENROOT_CONFIG_PATH}/hooks.d")
```

Because enroot's parser accepts any `ENROOT_[A-Z_]+` key, this exports a variable nothing reads rather than erroring — so it fails silently. The PR description's claim that "hook defaults remain explicitly overridable" does not hold: overriding this key has no effect.

**Correction:** drop the key. If site-supplied hooks were the actual goal, that needs a separate mount into `/etc/enroot/hooks.d` (note the image already populates that directory with enroot's `50-slurm-pmi.sh` / `50-sharp.sh` / etc., so a `subPath`-less ConfigMap mount there would shadow them).

recipes/components/slinky-slurm/values.yaml
```diff
@@ -142,6 +142,25 @@ storage:
...
+# leaf overlays and bundle-time overrides can add site-specific settings.
+enroot:
+  config:
+    ENROOT_LIBRARY_PATH: /usr/lib/enroot
+    ENROOT_SYSCONF_PATH: /etc/enroot
+    ENROOT_RUNTIME_PATH: /run/enroot/user-$(id -u)
+    ENROOT_CACHE_PATH: ${HOME}/.cache/enroot
+    ENROOT_DATA_PATH: ${HOME}/.local/share/enroot
+    ENROOT_TEMP_PATH: /tmp
+    ENROOT_CONFIG_PATH: ${HOME}/.config/enroot
```
mchmarny: Non-blocking, but `ENROOT_RUNTIME_PATH` breaks enroot for any non-root uid.

The Slinky image creates the parent as **mode 0755, root-owned** (`Dockerfile.pyxis`: `mkdir -m 755 -p /run/enroot/`). enroot then runs, under `set -euo pipefail`:

```sh
(umask 077 && mkdir -p "${ENROOT_CACHE_PATH}" "${ENROOT_DATA_PATH}" "${ENROOT_RUNTIME_PATH}")
```

So a non-root user gets `EACCES` on `mkdir /run/enroot/user-<uid>` and enroot exits non-zero before the container ever starts. Root-only clusters are unaffected — which is why the Kind and GB200 checks passed — but the login pod's SSSD users are exactly the multi-user case the `user-$(id -u)` convention exists for.

Worth noting this adopts only half of upstream's multi-user recipe: the image provisions `/usr/share/enroot/enroot-data` at **0777** for the paired `ENROOT_DATA_PATH=/usr/share/enroot/enroot-data/user-$(id -u)`, which this PR does not use (line 155 points at `${HOME}` instead).

**Suggestion:** either keep enroot's default runtime path, or pair the change with provisioning `/run/enroot` as `1777` and adopt the matching `ENROOT_DATA_PATH`.

Separately, on the surrounding keys in this block — six of the ten are inert today:
- `ENROOT_LIBRARY_PATH` / `ENROOT_SYSCONF_PATH` are enroot's compile-time `@libdir@` / `@sysconfdir@`. They match the deb-installed image, so pinning them buys nothing while converting any future image change into a hard failure (`source "${ENROOT_LIBRARY_PATH}/common.sh"` at the top of `runtime.sh`).
- `ENROOT_CACHE_PATH`, `ENROOT_DATA_PATH`, `ENROOT_CONFIG_PATH`, and `ENROOT_TEMP_PATH` are byte-identical to enroot's built-in defaults.

Trimming to just the values that actually differ from upstream would make the intent of this block much clearer.

demos/cuj1-slinky-slurm.md
```diff
@@ -100,6 +100,82 @@ AICR injects placement from bundle flags using each component's registry paths:
...
+
+```shell
+aicr bundle \
+  --recipe recipe.yaml \
+  --set slinkyslurm:enroot.env.NCCL_DEBUG=INFO \
+  --set slinkyslurm:enroot.env.NCCL_DEBUG_SUBSYS=INIT,NET \
+  --output bundle
+```
+
+Verify the environment through the Slurm path:
```
mchmarny: Non-blocking: on an existing cluster this verification step reports the **old** value, with nothing to indicate the change did not land.

I confirmed by bundling twice from the same recipe. `--set slinkyslurm:enroot.env.NCCL_DEBUG=INFO` changes `017-slinky-slurm-pre/templates/enroot-config.yaml` and the chart's `values.yaml`, but `enroot` is an AICR-only key that the Slinky chart never reads, so the rendered `LoginSet` / `NodeSet` CRs are byte-identical:

```
$ diff out-default/018-slinky-slurm/values.yaml out-override/018-slinky-slurm/values.yaml
66c66,67
<     NCCL_DEBUG: WARN
---
>     NCCL_DEBUG: INFO
>     NCCL_DEBUG_SUBSYS: INIT,NET
```

Since the CRs do not change, the operator rolls no pods; combined with `subPath` mounts not refreshing in place, the running containers keep the previous `/etc/enroot/environ.d/99-networking.env`.

The rollout note in the PR description covers this, but this doc page is where an operator will actually follow the flow.

**Suggestion:** add the restart step here (e.g. `kubectl rollout restart` / delete the LoginSet and NodeSet pods) before the `srun` verification. A durable alternative is stamping a checksum of the `enroot` map into the LoginSet/NodeSet `podSpec` annotations so the operator rolls automatically on change.

mchmarny (changes requested):
Reviewed `135682c7f197feedd36767703bc60584435101eb`. **1 blocker, 3 non-blocking findings**, all inline.

The shape of the change is good — rendering the maps through a pre-manifest, mounting by `subPath`, and always registering the pre-folder so it no longer depends on shared storage all look right, and the pre/main ordering (`017-slinky-slurm-pre` → `018-slinky-slurm`) is correct. The blocker is about the Enroot/Pyxis environment precedence rule, not the plumbing.

**Validated at this SHA:**

- Built `cmd/aicr` and ran the doc's exact commands end-to-end. Rendered ConfigMap identity (`slurm/slinky-slurm-enroot-config`) and deterministic sorted-key output both check out. `--set …NCCL_DEBUG_SUBSYS=INIT,NET` survives intact — `DisableSliceFlagSeparator: true` (`pkg/cli/bundle.go:626`) means the comma is not split, so that example is fine.
- Traced enroot's config parser: `IFS=$' \t='` means the `KEY=VALUE` form parses correctly despite differing from upstream's whitespace-separated sample, `$(id -u)` / `${HOME}` are `eval`-expanded, and `"yes"` survives `config::fini`.
- Confirmed Slinky chart v1.2.0 passes `login.volumeMounts` and `podSpec.volumes` through verbatim (`slurm.format-container` and `slurm.format-podSpec` both `toYaml`), and that the chart ships no `values.schema.json`, so the extra top-level `enroot` key is ignored rather than rejected.
- Checked the new test for shared-state pollution: `GetValuesForComponentWithContext` deep-copies, so mutating the returned maps in place is safe against the `sync.Once`-cached store.
- Chainsaw step has `set -e`, so the `PRE=$(ls -d …)` guard fails closed — no false PASS.

**CI:** all checks green, including Tier 2 `h100-eks-ubuntu-training-slurm (helm)` and `gb200-eks-ubuntu-training-slurm (helm)`. No CI signal contradicts these findings — none of those lanes launch a Pyxis container, so the environment-precedence behavior is not exercised anywhere in the suite.

**Caveat on evidence:** the blocker and the `ENROOT_RUNTIME_PATH` finding are established from enroot/Pyxis source and upstream docs, not executed against a live Slurm cluster. If your GB200 run covered a job-script `export NCCL_DEBUG=…` overriding the shipped default, I would want to see that, because it would contradict the Pyxis `--container-env` semantics.

Branch is `BEHIND` and needs a rebase, but that is mechanical and not part of the verdict.

recipes/components/slinky-slurm/values.yaml
njhensley, line 159: **🟠 Major — NCCL_DEBUG: WARN in enroot.env is unoverridable from a job script; doc guidance is wrong**

enroot writes enroot.env into /etc/enroot/environ.d/99-networking.env, folded into the container's own env; pyxis spank_import_container_env() calls spank_setenv(..., overwrite=1) for every container var not named in --container-env, so the container-side NCCL_DEBUG=WARN wins over a job-script `export NCCL_DEBUG=INFO` / `srun --export`. Stock images shipped no environ.d/*.env, so this is a regression.

**Blast radius:** Anyone debugging NCCL at runtime; cluster-wide default silently clamps per-job verbosity. The PR's own doc (cuj1:173-177) advising `srun --export` for such vars is incorrect for any key also in enroot.env.

**Fix:** Ship enroot.env empty (mount stays for sites to populate), OR keep NCCL_DEBUG=WARN and document that enroot.env entries win over job-script exports and require `--container-env=NCCL_DEBUG` to override. Fix the `srun --export` sentence either way.

recipes/components/slinky-slurm/values.yaml
```diff
@@ -142,6 +142,25 @@ storage:
...
 
+# Enroot settings shared by the login and compute containers. The matching
+# pre-manifest renders these maps as /etc/enroot/enroot.conf and
+# /etc/enroot/environ.d/99-networking.env. Keep values provider-neutral here;
+# leaf overlays and bundle-time overrides can add site-specific settings.
+enroot:
+  config:
+    ENROOT_LIBRARY_PATH: /usr/lib/enroot
+    ENROOT_SYSCONF_PATH: /etc/enroot
+    ENROOT_RUNTIME_PATH: /run/enroot/user-$(id -u)
```
njhensley: **🟡 Minor — ENROOT_RUNTIME_PATH=/run/enroot/user-$(id -u) breaks enroot for non-root uids**

Image creates /run/enroot as 0755 root-owned; enroot runs `(umask 077 && mkdir -p ... ${ENROOT_RUNTIME_PATH})` so a non-root user gets EACCES on mkdir /run/enroot/user-<uid> and enroot exits non-zero before the container starts. Root-only clusters (Kind/GB200 CI) are unaffected, masking it.

**Blast radius:** login-pod SSSD users and any REMAP_ROOT job step running as the submitting user — the exact multi-user case the user-$(id -u) convention exists for.

**Fix:** Keep enroot's default ${XDG_RUNTIME_DIR}/enroot, or provision /run/enroot as 1777 and adopt the paired ENROOT_DATA_PATH the image ships at 0777.

recipes/components/slinky-slurm/values.yaml
```diff
@@ -142,6 +142,25 @@ storage:
     storageClassName: ""
     size: 100Gi
 
+# Enroot settings shared by the login and compute containers. The matching
+# pre-manifest renders these maps as /etc/enroot/enroot.conf and
+# /etc/enroot/environ.d/99-networking.env. Keep values provider-neutral here;
+# leaf overlays and bundle-time overrides can add site-specific settings.
+enroot:
```
njhensley, line 149: **🟡 Minor — New enroot override surface documented only in demos/, not the canonical docs/ tree**

PR adds a user-visible override surface (slinkyslurm:enroot.env.*, slinkyslurm:enroot.config.*) and a new env-var deployment pattern. CLAUDE.md requires docs/ updates for new env-var/deployment patterns; component-catalog.md's slinky-slurm row is unchanged and no docs/ page mentions enroot. (--set-file itself is already documented in cli-reference.md.)

**Blast radius:** Discoverability only; flag mechanics already documented.

**Fix:** Add a pointer/section in docs/user/component-catalog.md (or a short slinky-slurm-enroot.md) for the enroot env/config knobs, or confirm the CUJ is the intended canonical home.

recipes/components/slinky-slurm/values.yaml
njhensley: **🔵 Nitpick — ENROOT_HOOKS_PATH is not a real enroot config variable — inert**

Hook dirs are hardcoded in enroot runtime.sh: hook_dirs=(${ENROOT_SYSCONF_PATH}/hooks.d ${ENROOT_CONFIG_PATH}/hooks.d). ENROOT_HOOKS_PATH is never read; enroot's parser accepts any ENROOT_[A-Z_]+ key so it exports a var nothing reads (also equals the default anyway).

**Blast radius:** None functionally; misleads operators into thinking hook path is configurable (PR description's 'hook defaults remain overridable' does not hold).

**Fix:** Drop the key.

recipes/components/slinky-slurm/values.yaml
```diff
@@ -142,6 +142,25 @@ storage:
...
+  config:
+    ENROOT_LIBRARY_PATH: /usr/lib/enroot
```
njhensley: **🔵 Nitpick — Several ENROOT_* keys merely restate enroot's compile-time defaults**

ENROOT_LIBRARY_PATH (@libdir@), ENROOT_SYSCONF_PATH (@sysconfdir@), ENROOT_TEMP_PATH (/tmp) equal enroot's built-in defaults; pinning LIBRARY/SYSCONF converts any future image path change into a hard failure (runtime.sh sources ${ENROOT_LIBRARY_PATH}/common.sh). CACHE/DATA/CONFIG/RUNTIME are genuine container overrides and should stay.

**Blast radius:** Adds surface area / brittleness; no runtime effect today.

**Fix:** Trim to only the keys that differ from upstream (MOUNT_HOME, REMAP_ROOT, and a corrected RUNTIME/DATA), with a comment per intentional delta.

demos/cuj1-slinky-slurm.md
njhensley: **🔵 Nitpick — srun verify step omits the subPath stale-mount / pod-restart caveat**

On an existing cluster, changing enroot.env + re-bundle updates only the ConfigMap; the LoginSet/NodeSet CRs are byte-identical (enroot is an AICR-only key the chart never reads), so no pod rolls, and subPath mounts never refresh in place. The `srun ... env | grep NCCL_DEBUG` verify then reports the OLD value with no signal the change did not land.

**Blast radius:** First-time reader iterating on an existing cluster is misled. Fresh installs are correct.

**Fix:** Add a `kubectl rollout restart` / pod-delete step before the srun verify, or stamp a checksum of the enroot map into the podSpec annotations so pods roll automatically on change.

recipes/components/slinky-slurm/manifests/enroot-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  labels:
+    app.kubernetes.io/name: slinky-slurm
+    app.kubernetes.io/component: enroot-config
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+data:
+  enroot.conf: |
+    {{- range $key, $value := index $enroot "config" }}
+    {{ $key }}={{ $value }}
+    {{- end }}
+  99-networking.env: |
```
njhensley: **🔵 Nitpick — 99-networking.env filename is a misnomer (carries NCCL_DEBUG, a logging knob)**

The 99- prefix (last-sorted, highest precedence) is a correct choice, but the file name implies networking config while it currently holds only a logging verbosity default.

**Blast radius:** Cosmetic / operator confusion.

**Fix:** Rename to a contents-accurate name (e.g. 99-aicr-defaults.env).

recipes/components/slinky-slurm/values.yaml
```diff
@@ -156,11 +175,25 @@ loginsets:
...
+          subPath: 99-networking.env
+          readOnly: true
     # Pin the initconf sidecar to an immutable Alpine tag (chart default
     # is :latest).
     initconf:
       image:
         tag: "3.23.3"
+    podSpec:
+      volumes:
+        - name: enroot-config
```
njhensley, line 191: **🔵 Nitpick — ConfigMap lives in the -pre helm release but is consumed by the main release**

The ConfigMap is rendered in NNN-slinky-slurm-pre; the pods that mount it are in NNN-slinky-slurm. `helm uninstall <name>-pre` or a GitOps prune removes the ConfigMap while pods still reference it; helm upgrade of the main release alone never touches it. Same pattern as the existing shared-storage PVCs.

**Blast radius:** Operational footgun on partial upgrade/rollback/prune; latent until a pod reschedules.

**Fix:** Document the coupling; ensure GitOps prune policies retain the pre release's ConfigMap.

recipes/components/slinky-slurm/manifests/enroot-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: slinky-slurm-enroot-config
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: slinky-slurm
+    app.kubernetes.io/component: enroot-config
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+data:
+  enroot.conf: |
+    {{- range $key, $value := index $enroot "config" }}
+    {{ $key }}={{ $value }}
```
njhensley, line 29: **🔵 Nitpick — A multi-line operator override value would break the | block scalar**

`{{ $key }}={{ $value }}` emits each pair on one line under a 4-space-indented | literal block scalar. A site override whose value contains a newline emits a second physical line with no indent, terminating the block scalar early → invalid/mis-parsed YAML. All committed defaults are single-line, so no current recipe is affected.

**Blast radius:** Operator-supplied override only; latent.

**Fix:** Optional hardening: document that enroot values must be single-line, or nindent/quote-guard the value.

tests/chainsaw/cli/bundle-slinky-storage/chainsaw-test.yaml
```diff
@@ -49,7 +49,16 @@ spec:
...
                 --output "${WORK}/disabled"
-              ! ls -d "${WORK}"/disabled/[0-9][0-9][0-9]-slinky-slurm-pre >/dev/null 2>&1
+              PRE=$(ls -d "${WORK}"/disabled/[0-9][0-9][0-9]-slinky-slurm-pre)
+              test -f "${PRE}/templates/enroot-config.yaml"
+              test ! -e "${PRE}/templates/shared-storage-pvcs.yaml"
+              yq -e '
+                .apiVersion == "v1" and
+                .kind == "ConfigMap" and
+                .metadata.name == "slinky-slurm-enroot-config" and
+                (.data."enroot.conf" | contains("ENROOT_RUNTIME_PATH=")) and
```
njhensley: **🔵 Nitpick — Enroot pre-manifest asserted only in the disabled-storage branch**

The test description claims it validates the always-on enroot manifest across default-off AND enabled paths, but only disabled-storage-renders-no-pvcs asserts enroot-config.yaml. The enabled branch checks only PVCs/slurm values. Enroot is storage-independent so a regression would still be caught by the disabled branch + Go tests.

**Blast radius:** Low — completeness gap vs the stated description.

**Fix:** Add a `test -f .../enabled/...-pre/templates/enroot-config.yaml` to the enabled path, or narrow the description.

njhensley (commented):
📋 Multi-persona review — head `135682c7`

▎ Method: 4 persona passes (Domain/enroot-Pyxis, Kubernetes/Operability, Correctness/template, Test-coverage & Docs), each finding independently confirmed or refuted by a senior meta-reviewer against the resolved code + enroot upstream source. Line links pinned to head `135682c7`.

> ⚠️ **@mchmarny already has a CHANGES_REQUESTED review on this PR.** This pass **confirms all four of his findings** with independent verification and adds net-new ones; overlapping inline comments are marked `= mchmarny #N`. One correction: my Domain persona initially raised a 🔴 "malformed `enroot.conf`" blocker, but the meta-reviewer **refuted** it — @mchmarny was right that `KEY=VALUE` parses fine (details below), so that theory should not carry weight.

## Overall assessment — Comment (no hard blocker survives independent verification)

The plumbing is sound: rendering the maps through a pre-manifest, mounting by read-only `subPath`, always registering the pre-folder, and the `017-…-pre → 018-…` ordering all check out; the new Go/chainsaw tests genuinely exercise render + override propagation and are fail-closed. **Nothing that reproduces breaks the feature or CI.** The one substantive standing issue is behavioral/documentation: `NCCL_DEBUG: WARN` shipped in `enroot.env` cannot be overridden from a job script (Pyxis `overwrite=1`), and the doc telling users to `srun --export` such vars is wrong for any key that also lives in `enroot.env`. That plus the inert/redundant/non-root-unsafe enroot keys and the docs-location gap are worth addressing before merge.

- **"`enroot.conf` `KEY=VALUE` is malformed → feature is a no-op" — REFUTED.** enroot parses `enroot.conf` with `IFS=$' \t='` (`config::load` in `enroot.in`), so `KEY=VALUE` tokenizes identically to `KEY VALUE`; the `bash -n` gate accepts assignment lines and `${HOME}`/`$(id -u)` are `eval`-expanded. `ENROOT_MOUNT_HOME`/`ENROOT_REMAP_ROOT` **do** apply.
- **Hardcoded ConfigMap-name drift is guarded** — `TestSlinkySlurmEnrootDefaultsRenderConfigMap` pins the literal for both the rendered name and both volume `configMap.name` refs, so a one-sided rename fails CI.
- **Deterministic render** — `text/template` `range` sorts string map keys; output is byte-stable and emits `KEY=VALUE` text directly (no randomized map marshaling).
- **Fail-closed everywhere it matters** — pre-folder `017→018` ordering under `set -e`; ConfigMap and pods co-namespaced via `.Release.Namespace`; missing-ConfigMap volume is *required* so pods fail-closed rather than start misconfigured; chainsaw `PRE=$(ls -d …)` fails closed.
- **No shared-state pollution** in the new tests — `GetValuesForComponentWithContext` returns a deep-copied/fresh map.
- **subPath-replaces-image-`enroot.conf`** and **volume-name collision** — unverifiable (chart/images not vendored) and likely benign; flagged for the author to confirm, not asserted.
- **BusyBox digest drift** in `container-images.md` is a benign `bom-docs` regen, gated by `TestCommittedBOMVersionsMatchRegistry`.

## Summary

| Tier | Count | Items |
|------|-------|-------|
| 🔴 Blocker | 0 | — |
| 🟠 Major | 1 | NCCL_DEBUG unoverridable + wrong `srun --export` doc guidance |
| 🟡 Minor | 2 | non-root `ENROOT_RUNTIME_PATH` EACCES · enroot override surface missing from `docs/` |
| 🔵 Nitpick | 6 | inert `ENROOT_HOOKS_PATH` · redundant default keys · stale-mount doc caveat · `99-networking.env` name · cross-release ConfigMap ownership · multi-line-override block scalar · chainsaw enabled-branch assert |

Recommendation: **Comment** — no hard blocker survives independent verification, but the 🟠 and the two 🟡s should be addressed. Aligns with @mchmarny's CHANGES_REQUESTED on the narrow grounds of override-precedence behavior + docs, **not** the malformed-config theory.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. We kept `NCCL_DEBUG=WARN` as the cluster-wide default and corrected both the CUJ and the new canonical Enroot guide: an `enroot.env` entry wins over an ordinary job export, and a job override must export the value and pass `--container-env=NCCL_DEBUG` (or a comma-separated list). The guide also explicitly states that `srun --export` alone does not override the container value.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. Removed `ENROOT_HOOKS_PATH` entirely and removed the associated hook/default claims from the PR description. Hook injection is not part of this change.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. Removed `ENROOT_RUNTIME_PATH` and all path values that merely restated Enroot or image defaults. The shipped `enroot.config` now contains only the intentional policy deltas `ENROOT_MOUNT_HOME=yes` and `ENROOT_REMAP_ROOT=yes`. Tests explicitly assert that `ENROOT_RUNTIME_PATH` is absent, so non-root users retain Enroot's XDG runtime default.

demos/cuj1-slinky-slurm.md
kaynetu: Addressed in ce9f0b04. The CUJ now links to the canonical Enroot guide for existing-cluster rollout. The guide explains `subPath` staleness, requires draining all Slurm nodes and waiting for active jobs to finish, replaces worker pods one at a time, replaces login pods after notifying users, and verifies only after replacement. We document pod deletion rather than `kubectl rollout restart`, because these operator-managed pods are not refreshed by restarting the LoginSet Deployment.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. We kept `NCCL_DEBUG=WARN` as the recommended cluster-wide baseline and documented the required escape hatch in both the CUJ and canonical guide: export the job value and pass its name through `srun --container-env`. The guide explicitly notes that an ordinary export or `srun --export` alone does not override an `enroot.env` entry.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. Removed the `ENROOT_RUNTIME_PATH` override, so non-root users retain Enroot's built-in XDG runtime path. The rendering tests also assert that `ENROOT_RUNTIME_PATH` is absent.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. Added `docs/user/slinky-slurm-enroot.md` as the canonical guide and linked it from the user index, docs navigation, and the Slinky Slurm row in the component catalog. The CUJ remains concise and links to that guide for the complete override and rollout workflow.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. Removed `ENROOT_HOOKS_PATH`; the final defaults and PR description no longer claim that hook paths are configurable through this surface.

kaynetu: Addressed in ce9f0b04. Removed `ENROOT_HOOKS_PATH`; the final defaults and PR description no longer claim that hook paths are configurable through this surface.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. Removed the compile-time and built-in path defaults, including `ENROOT_LIBRARY_PATH`, `ENROOT_SYSCONF_PATH`, `ENROOT_CACHE_PATH`, `ENROOT_DATA_PATH`, `ENROOT_CONFIG_PATH`, `ENROOT_RUNTIME_PATH`, and `ENROOT_TEMP_PATH`. The shipped config now contains only the two intentional cluster-policy overrides: `ENROOT_MOUNT_HOME=yes` and `ENROOT_REMAP_ROOT=yes`.

demos/cuj1-slinky-slurm.md
kaynetu: Addressed in ce9f0b04. The CUJ links directly to the canonical guide, which explains that ConfigMap-backed `subPath` mounts remain stale, requires draining and waiting for active jobs before any deletion, and provides the ordered worker/login pod replacement and post-replacement verification flow.

recipes/components/slinky-slurm/manifests/enroot-config.yaml
kaynetu: Addressed in ce9f0b04. Renamed the file and every mount/test/reference to `/etc/enroot/environ.d/99-aicr-defaults.env`.

recipes/components/slinky-slurm/values.yaml
kaynetu: Addressed in ce9f0b04. The canonical guide now documents that `slinky-slurm-pre` owns the ConfigMap consumed by `slinky-slurm`; both releases must be deployed, upgraded, rolled back, and removed as one generated bundle. It explicitly says not to prune or uninstall the pre-release while the main release exists, and gives the reverse teardown order.

recipes/components/slinky-slurm/manifests/enroot-config.yaml
kaynetu: Addressed in ce9f0b04 using the documentation option. The canonical guide states that keys and values in both Enroot maps must be single-line because newlines can produce an invalid ConfigMap or malformed Enroot configuration.

tests/chainsaw/cli/bundle-slinky-storage/chainsaw-test.yaml
kaynetu: Addressed in ce9f0b04. Added the missing `test -f` assertion for the enabled-storage bundle's `slinky-slurm-pre/templates/enroot-config.yaml`, matching the test description and the disabled-storage branch.

docs/user/slinky-slurm-enroot.md
```diff
@@ -0,0 +1,150 @@
...
+containers so interactive and batch jobs receive consistent defaults.
+
+## Shipped defaults
+
+The Slinky Slurm component ships these provider-neutral defaults:
+
+| Component value | Default | Purpose |
+| --- | --- | --- |
+| `enroot.config.ENROOT_MOUNT_HOME` | `yes` | Mount the user's home directory in the container |
+| `enroot.config.ENROOT_REMAP_ROOT` | `yes` | Remap the container user to root inside the container |
```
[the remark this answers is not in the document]
kaynetu: Addressed in 5486b73e. Updated the table to state that `ENROOT_REMAP_ROOT=yes` remaps the current user to root inside the container, equivalent to Enroot's `--root` option.

pkg/recipe/metadata_store_test.go
```diff
@@ -780,6 +780,171 @@ func TestSlurmLeavesIncludeSharedStoragePreManifest(t *testing.T) {
...
+			if len(want) != 0 {
+				t.Errorf("%v missing Enroot mounts: %v", tt.path, want)
+			}
+		})
+	}
+
+	// Bundle-time map overrides feed the pre-manifest renderer, allowing sites
+	// to replace defaults or add provider-specific Enroot and networking keys.
+	config["ENROOT_CACHE_PATH"] = "/custom/cache"
+	env["NCCL_DEBUG"] = "INFO"
```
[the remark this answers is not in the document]
kaynetu: Addressed in 5486b73e. Removed the direct map mutations from the `pkg/recipe` test, which now verifies default resolution and rendering only. Added a `bundle-time-enroot-overrides` Chainsaw step that runs the real `aicr bundle --set` path with the `slinkyslurm` alias and asserts both overridden values in the rendered ConfigMap. The focused Go and Chainsaw tests and full `make qualify` pass.

pkg/recipe/metadata_store_test.go
```diff
@@ -780,6 +780,168 @@ func TestSlurmLeavesIncludeSharedStoragePreManifest(t *testing.T) {
...
+						t.Errorf("mount %q subPath = %v, want %q", mountPath, got, subPath)
+					}
+					delete(want, mountPath)
+				}
+			}
+			if len(want) != 0 {
+				t.Errorf("%v missing Enroot mounts: %v", tt.path, want)
+			}
+		})
+	}
```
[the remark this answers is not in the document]
kaynetu: Addressed in 779ca0d4. Each expected login and compute Enroot mount now asserts `name: enroot-config` in addition to its `mountPath` and `subPath`, so a mount referencing another volume cannot satisfy the test. I also fixed the whitespace lint finding by splitting the multiline condition into named booleans. The focused Go test, package `golangci-lint` (0 issues), and full `make qualify` pass.

pkg/recipe/metadata_store_test.go
```diff
@@ -780,6 +780,173 @@ func TestSlurmLeavesIncludeSharedStoragePreManifest(t *testing.T) {
...
+				mountPath, _ := mount["mountPath"].(string)
+				if subPath, found := want[mountPath]; found {
+					if got := mount["name"]; got != "enroot-config" {
+						t.Errorf("mount %q name = %v, want enroot-config", mountPath, got)
+					}
+					if got := mount["subPath"]; got != subPath {
+						t.Errorf("mount %q subPath = %v, want %q", mountPath, got, subPath)
+					}
+					delete(want, mountPath)
+				}
```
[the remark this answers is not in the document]
kaynetu: Addressed in 1afa746e. The test now counts every match for each expected Enroot `mountPath` and requires the count to equal exactly one, while preserving the volume-name and `subPath` assertions. This rejects both missing and duplicate mounts. The focused Go test, package `golangci-lint` (0 issues), and full `make qualify` pass.

pkg/recipe/metadata_store_test.go
```diff
@@ -780,6 +780,176 @@ func TestSlurmLeavesIncludeSharedStoragePreManifest(t *testing.T) {
...
+			configMap.Metadata.Namespace, configMap.Metadata.Name)
+	}
+	enrootConfig := configMap.Data["enroot.conf"]
+	hasMountHome := strings.Contains(enrootConfig, "ENROOT_MOUNT_HOME=yes")
+	hasRemapRoot := strings.Contains(enrootConfig, "ENROOT_REMAP_ROOT=yes")
+	if !hasMountHome || !hasRemapRoot {
+		t.Errorf("rendered enroot.conf missing defaults:\n%s", enrootConfig)
+	}
+	if got := configMap.Data["99-aicr-defaults.env"]; !strings.Contains(got, "NCCL_DEBUG=WARN") {
+		t.Errorf("rendered 99-aicr-defaults.env missing default:\n%s", got)
```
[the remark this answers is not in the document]
kaynetu: Thanks. I agree exact assignment parsing would make this test stricter, but the current test’s scope is a rendering smoke check, not validation of Enroot file syntax. The source-value assertions above already verify the exact default values, while this section confirms those values appear in the rendered ConfigMap. I’m leaving the additional parser-based test hardening out of scope for this PR.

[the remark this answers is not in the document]
kaynetu: No issue needed. This is optional test hardening with low risk, and the existing source-value assertions plus rendered ConfigMap smoke checks provide sufficient coverage for this PR. Resolving this thread.

mchmarny (approved):
Approving `1afa746e`. All four of my findings on `135682c7` are addressed, verified against the live head rather than the replies.

**Blocker (NCCL_DEBUG override) - resolved** via the second option I offered: `NCCL_DEBUG=WARN` stays, and both the CUJ and the new `docs/user/slinky-slurm-enroot.md` state that an `enroot.env` entry beats an ordinary job export and require `srun --container-env=<VAR>`. The incorrect `srun --export` sentence is gone.

**`ENROOT_HOOKS_PATH` - removed.** No occurrence anywhere in the tree.

**`ENROOT_RUNTIME_PATH` and the inert path keys - removed.** `enroot.config` is now exactly `ENROOT_MOUNT_HOME` and `ENROOT_REMAP_ROOT`, pinned by a `len(config) == 2` assertion in Go plus a fail-closed `contains("ENROOT_RUNTIME_PATH=") | not` in Chainsaw, which is stronger than naming the key.

**Stale `subPath` verification - resolved.** The canonical guide carries a full drain-and-replace runbook and the CUJ links to it. I checked the `SlurmNodeStateDrain` / `SlurmNodeStateIdle` pod conditions used in the `kubectl wait` steps against the operator's condition table; both are real.

Validated beyond the reply text, on a bundle generated from this head:

- The rendered pre-release ConfigMap carries the expected two-key `enroot.conf` and `NCCL_DEBUG=WARN`; `017-slinky-slurm-pre` sorts before `018-slinky-slurm` in the generic install loop, and the GB200 leaf keeps its ComputeDomain in the same pre-folder.
- `--set slinkyslurm:enroot.env.NCCL_DEBUG=INFO --set slinkyslurm:enroot.config.ENROOT_MOUNT_HOME=no` propagates through alias resolution into the rendered ConfigMap.
- The mounts are actually consumed: chart v1.2.0 has no default `podSpec.volumes` or `volumeMounts` for either LoginSet or NodeSet, so these lists clobber nothing, and `slurm.format-container` / `slurm.format-podSpec` are plain `toYaml` pass-throughs into the CRs. There is no `values.schema.json`, so the AICR-only `enroot` key is inert to the chart.
- Both pinned images (`login-pyxis` and `slurmd-pyxis`, at the pinned digests) ship a fully commented-out `/etc/enroot/enroot.conf`, so replacing that file via `subPath` drops nothing. `environ.d/10-terminal.env` survives because `subPath` adds a file rather than shadowing the directory, and `99-` sorts after it.
- `pkg/recipe` 89.0%, `golangci-lint` 0 issues, `yamllint` clean on every changed YAML, and the Chainsaw `yq` assertions replay green against the real rendered artifact.

Branch is `BEHIND`; the only movement on main touching these paths is `nodewright-customizations/manifests/tuning.yaml`, so the rebase is mechanical.

njhensley (approved):

PR #2170, opened 2026-08-12
fix(server): harden bundle vendor path against SSRF and resource exhaustion

### LINKED ISSUE #2118: Harden server chart vendoring against SSRF and resource exhaustion
## Bug description

A network-reachable `aicrd` can be instructed to vendor a Helm chart from an arbitrary HTTP(S) or OCI source. The default server address binds all interfaces, the middleware chain has no authentication or authorization layer, and `POST /v1/bundle` / `POST /v2/bundle` accept caller-supplied component sources.

With `vendor-charts=true`, the server passes that source to `helm pull`. Validation checks only the URL scheme; it does not reject loopback, link-local, private, or metadata-service destinations, nor does it constrain DNS rebinding or redirects. The downloaded `.tgz` is then loaded with `os.ReadFile` without a byte cap.

## Impact

A caller that can reach `aicrd` can trigger server-side requests into protected networks and consume unbounded temporary-disk and process memory within the request timeout. A Helm-compatible internal endpoint may also have its chart payload returned inside the generated bundle.

## Evidence

- Default all-interface bind: https://github.com/NVIDIA/aicr/blob/6cb3ab9445073c875007bfe95ca54aa2a609843c/pkg/server/config.go#L52-L64
- Middleware chain without authentication: https://github.com/NVIDIA/aicr/blob/6cb3ab9445073c875007bfe95ca54aa2a609843c/pkg/server/middleware.go#L29-L60
- Caller-supplied recipe adoption: https://github.com/NVIDIA/aicr/blob/6cb3ab9445073c875007bfe95ca54aa2a609843c/pkg/server/bundle_handler.go#L161-L218
- `helm pull` and unbounded `os.ReadFile`: https://github.com/NVIDIA/aicr/blob/6cb3ab9445073c875007bfe95ca54aa2a609843c/pkg/bundler/deployer/localformat/vendor.go#L121-L188
- Scheme-only repository validation: https://github.com/NVIDIA/aicr/blob/6cb3ab9445073c875007bfe95ca54aa2a609843c/pkg/bundler/deployer/localformat/vendor.go#L295-L365

## Reproduction

1. Start `aicrd` with its default listen address.
2. Submit a valid `RecipeResult` whose Helm component source points to an HTTP service reachable only from the server.
3. Request `/v1/bundle?vendor-charts=true`.
4. Observe the server invoke Helm against that destination. A sufficiently large chart archive is loaded entirely into memory.

## Expected behavior

The server boundary must not provide unrestricted network egress or accept an unbounded chart artifact from an untrusted request.

## Acceptance criteria

- [ ] Bundle generation is authenticated/authorized, or server-side chart vendoring is disabled unless explicitly enabled by the operator.
- [ ] HTTP(S), OCI, DNS, and redirect targets are checked against an egress policy that rejects loopback, link-local, metadata-service, and private destinations by default.
- [ ] Downloaded chart archives have hard compressed and extracted-size limits and are streamed instead of read unbounded into memory.
- [ ] Tests cover direct IPs, DNS resolution, redirects, OCI registries, oversized archives, and valid public repositories.
- [ ] Security-sensitive defaults and deployment guidance are documented.

framsouza:
## Summary

Hardens `POST /v1/bundle?vendor-charts=true` on `aicrd` with three layered defenses: an egress-policy pre-check on the repository host, a hard cap on the vendored artifact size, and an operator opt-in that keeps the vendor path off by default. Also flips the server's default bind from all-interfaces to loopback.

## Motivation / Context

Issue #2118 documents a composed weakness on the API server surface: default bind on all interfaces + no authentication + scheme-only URL validation + unbounded `os.ReadFile` on the pulled `.tgz`. Together these let any TCP peer reachable to `aicrd` drive server-side `helm pull` at arbitrary hosts (SSRF into cloud-metadata / internal services, with the response body round-tripping into the returned bundle) and consume unbounded memory.

Fixes: #2118
Related: N/A

## Type of Change

- [x] Documentation update

The default-bind change (`""` → `127.0.0.1`) and the vendor-charts opt-in are behavior changes for existing operators — see **Rollout notes** below.

## Component(s) Affected

- [x] Bundlers (`pkg/bundler`, `pkg/component/*`) — `pkg/bundler/deployer/localformat`
- [x] Core libraries (`pkg/errors`, `pkg/k8s`) — `pkg/defaults` (new constants)

## Implementation Notes

**Three layered defenses**, each independently useful, together closing the composed weakness:

1. **Egress policy on the vendor path** — `checkEgressPolicy(ctx, repoURL)` runs inside `CLIChartPuller.Pull` before `helm pull` and rejects repository hosts that resolve to loopback, link-local, RFC1918 / RFC6598 CGNAT / RFC4193 ULA private, multicast, unspecified, or the well-known cloud-metadata IPs (`169.254.169.254`, `100.100.100.200`, `fd00:ec2::254`, `fe80::a9fe:a9fe`). Cloud-metadata check runs *first* so operator logs get the specific class (an SRE reading `link-local` won't immediately think "someone tried to steal my IAM creds"). Injectable resolver hook (`lookupIP`) keeps every test hermetic; production uses `net.DefaultResolver.LookupIPAddr` bound to the request context. Applies to `http`, `https`, and `oci://` uniformly.

2. **Artifact size bound** — new `defaults.HelmChartArtifactLimit = 64 MiB` (well above real charts, low enough to bound server memory). `readBoundedTgz` replaces the unbounded `os.ReadFile` with `Stat` + `io.LimitReader(f, limit+1)`. The `+1` catches a Stat/Open race where the file grows between calls.

3. **Server-level opt-in + bind default** — `POST /v1/bundle?vendor-charts=true` returns `400 vendor-charts is not enabled on this server` unless the operator sets `AICR_ALLOW_VENDOR_CHARTS=true`. Startup emits an INFO log when the opt-in is on, so an unexpected enablement is grep-able. Default bind flipped from `""` (all interfaces) to `127.0.0.1`; operator opts into external exposure via `AICR_SERVER_ADDRESS`. Setting the env to the empty string preserves the legacy all-interface behavior exactly for operators who need it.

**Design non-goals** (each worth its own issue/PR):
- **Full DNS-rebinding closure.** Between the pre-check and helm's re-resolution there's still a rebinding window. Full closure needs an in-process fetch that pins the resolved IP into the outbound HTTP request. Documented as residual risk in `docs/integrator/kubernetes-deployment.md`.
- **Real authn on `POST /v1/bundle`.** Token vs. mTLS vs. OIDC is a design conversation, not a hardening PR. The opt-in gate + loopback bind get the acute exposure down without waiting on that.

## Testing

```bash
make lint             # 0 issues (golangci-lint + yamllint + MDX + license + BOM + agents-sync)
go test -race ./pkg/bundler/deployer/localformat/... ./pkg/server/... ./pkg/defaults/...
# and adjacent packages: ./pkg/bundler/... ./pkg/client/... — all pass
```

**Coverage delta on touched packages:**
- `pkg/bundler/deployer/localformat`: 75.7% → 75.9% (+0.2%)
- `pkg/server`: 82.3% → 82.4% (+0.1%)
- `pkg/defaults`: 100.0% (unchanged)

**New test coverage:**
- `TestCheckEgressPolicy_LiteralIP` — 18-case matrix (loopback v4/v6, cloud-metadata across clouds, link-local, RFC1918, CGNAT 100.64/10, ULA fc00::/7, unspecified, multicast, `oci://` variants, plus positive cases for public IPs)
- `TestCheckEgressPolicy_ResolvedHost` — 6 mock-resolved cases including the split-horizon rejection (public+private both returned)
- `TestCheckEgressPolicy_MalformedURL` — 3 cases
- `TestReadBoundedTgz` — 5 cases (under-limit, exactly-at, one-over, far-over, empty)
- `TestReadBoundedTgz_Missing` — one case verifying missing-file surfaces as `Internal`, not `InvalidRequest`
- `TestCLIChartPuller_SSRFFilterWired` — end-to-end proof the filter runs inside `Pull()`
- `TestBundleHandler_VendorChartsDisabled` — 400 when the server has not opted in (fires BEFORE body decode)
- `TestBundleHandler_VendorChartsFalseAllowed` — `vendor-charts=false` passes the gate

**Live end-to-end verification** with a freshly built `aicrd` binary:
- Default bind is loopback-only — external interface refuses connection.
- `POST /v1/bundle?vendor-charts=true` returns 400 `vendor-charts is not enabled on this server` by default.
- `vendor-charts=false` passes the gate and hits the empty-components error downstream (proves the gate doesn't over-reject).
- `AICR_ALLOW_VENDOR_CHARTS=true` opens the gate + emits the INFO log warning.
- `AICR_SERVER_ADDRESS=0.0.0.0` opts into external exposure.
- All 6 chart hostnames in the embedded catalog (charts.jetstack.io, helm.ngc.nvidia.com, aws.github.io, kubernetes-sigs.github.io, ghcr.io, registry.k8s.io) resolve to public IPs — filter is a no-op for real recipes.

## Risk Assessment

**Rollout notes:**

Two behavior changes require operator awareness:

1. **Default bind flip** — deployments that relied on the previous all-interface default without setting `AICR_SERVER_ADDRESS` will no longer be reachable off-loopback. The stock in-tree K8s Deployment manifest already exposes via a Service+Ingress in front of the pod-local server, so the majority of prod deployments are unaffected. Operators exposing `aicrd` directly on a pod IP must set `AICR_SERVER_ADDRESS=0.0.0.0`.
2. **Vendor-charts default off** — any client that relied on `vendor-charts=true` working against an unauthenticated `aicrd` will get 400 until the operator sets `AICR_ALLOW_VENDOR_CHARTS=true`. This is intentional: the previous default performed server-side network egress on behalf of any reachable caller.

Both are opt-in-to-restore via env vars, so the migration is one env line in the Deployment for operators who want the old behavior. Both are documented in `docs/integrator/kubernetes-deployment.md` and the vendor-charts note in `docs/user/api-reference.md`.

## Checklist


pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -131,6 +133,17 @@ func (p *CLIChartPuller) Pull(ctx context.Context, c Component) ([]byte, VendorR
...
+	// SSRF defense: reject repository hosts that resolve to loopback,
+	// link-local, private, multicast, unspecified, or cloud-metadata IPs
+	// BEFORE invoking `helm pull`. A pre-check does not close the DNS-
+	// rebinding window on its own (helm re-resolves at fetch time), but it
+	// blocks the trivial cases and gives operators a coded 400 for the
+	// obvious probe traffic. Full closure requires an in-process fetch that
+	// pins the resolved IP into the HTTP request.
+	if err := checkEgressPolicy(ctx, c.Repository); err != nil {
+		return nil, VendorRecord{}, "", err
+	}
```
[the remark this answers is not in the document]
framsouza: Fixed it

pkg/defaults/timeouts.go
```diff
@@ -917,6 +917,43 @@ const (
...
 	// overrides ServerShutdownTimeout (value parsed as seconds).
 	EnvServerShutdownTimeoutSeconds = "SHUTDOWN_TIMEOUT_SECONDS"
+
+	// ServerDefaultBindAddress is the default listen address for aicrd.
+	// Loopback-only unless an operator explicitly opts into external
+	// exposure via EnvServerAddress. The prior all-interface default
+	// combined with the un-authenticated middleware chain to make every
+	// endpoint (including the vendor-charts SSRF surface — issue #2118)
+	// reachable from any network peer the container was routable to.
+	ServerDefaultBindAddress = "127.0.0.1"
```
njhensley: **🔴 Blocker — Loopback-only default bind breaks kubelet probes, Service routing, and Cloud Run**

The server binds the single mux to `127.0.0.1:8080` (this default flows to config.go:63 → server.go:116 `httpServer.Addr`). In a standard K8s pod, kubelet `httpGet` liveness/readiness probes and kube-proxy Service DNAT both target the **pod IP**, not loopback → connection refused → liveness restart loop + readiness never passes → zero Service endpoints. Cloud Run additionally requires the container to listen on `0.0.0.0:$PORT`. Verified three in-tree surfaces break and none set `AICR_SERVER_ADDRESS`: the documented prod manifest in this PR (`docs/integrator/kubernetes-deployment.md` L108-131, httpGet probes), `tilt/k8s/deployment.yaml` (`make dev-env`), and `.github/workflows/deploy.yaml` (Cloud Run deploy of `ghcr.io/nvidia/aicrd`).

**Blast radius:** Every default-manifest deployment (documented prod manifest, Tilt dev-env, Cloud Run) breaks on upgrade — CrashLoop / no Service endpoints / failed Cloud Run rollout. The PR body's rationale that a Service+Ingress makes it fine does not hold: a Service forwards to podIP:targetPort, which a loopback listener refuses.

**Fix:** Pick one and do it completely: (a) keep the secure default but set `AICR_SERVER_ADDRESS=0.0.0.0` in all three in-tree surfaces in this PR and pair with the already-recommended NetworkPolicy (0.0.0.0 + ClusterIP + NetworkPolicy is the correct in-cluster posture); or (b) keep the 0.0.0.0 bind default and rely on the opt-in vendor gate + NetworkPolicy/ingress. The 'live-verified: external interface refuses' check confirmed the security property but was never run inside a K8s Deployment, so it never exercised the probe/Service path this breaks.

docs/integrator/kubernetes-deployment.md
```diff
@@ -236,6 +236,8 @@ duplicated here.
 | Variable | Default | Description |
 |----------|---------|-------------|
 | `PORT` | 8080 | HTTP server port |
+| `AICR_SERVER_ADDRESS` | `127.0.0.1` | Listen address. Loopback-only by default; set to `0.0.0.0` (or a specific interface) to expose beyond the pod, and front with an authenticated ingress or firewalled network segment. Empty string preserves the legacy all-interface bind. |
```
njhensley: **🟡 Minor — No breaking-change / migration note for the bind flip; docs internally inconsistent**

This env-table row documents the new loopback default, but the example Deployment above it (L108-131) still omits `AICR_SERVER_ADDRESS`, so a reader following the page top-to-bottom deploys the manifest the new default breaks. The commit is typed `fix(server):`, so `tools/changelog` files a behavior-breaking default change as a bugfix with no operator-facing migration note.

**Blast radius:** Operators upgrading in place get a broken rollout with no forewarning; new operators copy a manifest that can't work.

**Fix:** Add `AICR_SERVER_ADDRESS: "0.0.0.0"` to the documented and Tilt manifests (folds into the F1 fix), and add an explicit 'Breaking change: default bind is now loopback' note near the env-var table + release notes.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+var cgnat100 = net.IPNet{IP: net.IPv4(100, 64, 0, 0), Mask: net.CIDRMask(10, 32)}
+
+// disallowedIPReason returns a non-empty reason string when ip is in a
+// range aicrd must not initiate egress to from a request-driven code path.
+// The empty string means "allowed."
+//
+// Cloud-metadata IPs are checked FIRST — 169.254.169.254 also matches
+// IsLinkLocalUnicast, but the metadata-specific reason is what an
+// operator needs to see in logs to recognize a credential-theft attempt.
+func disallowedIPReason(ip net.IP) string {
```
njhensley, line 522: **🟡 Minor — IP egress classification gaps: NAT64/6to4 transition addresses + reserved/martian IPv4**

`disallowedIPReason` relies on Go's `net.IP` classifiers + the explicit cloud-metadata list. Reproduced gaps: (1) NAT64 `64:ff9b::10.0.0.1` and 6to4 `2002:0a00:0001::` embedding private/metadata IPv4 are classified *allowed* — reachable where a NAT64/DNS64 gateway translates non-global v4 (a real but non-default GKE/EKS topology). (2) Reserved/martian IPv4 `0.0.0.0/8` (a `0.x` host routes to localhost on Linux), `240.0.0.0/4`, and `255.255.255.255` fall through as allowed; only `0.0.0.0` itself is caught via `IsUnspecified`. (Note: IPv4-mapped IPv6 like `::ffff:169.254.169.254` IS correctly caught — verified.)

**Blast radius:** SSRF to metadata/private ranges in NAT64/DNS64 environments; the `0.x→localhost` case defeats the loopback protection on Linux. Preconditions bound real-world reach, hence Minor not Major.

**Fix:** Decode the `64:ff9b::/96` and `2002::/16` embedded v4 and re-run the check (or reject non-global-unicast IPv6 outright), and add explicit CIDR checks for `0.0.0.0/8`, `240.0.0.0/4`, and the broadcast address alongside the existing cgnat100 check.

pkg/server/config.go
```diff
@@ -86,5 +98,27 @@ func parseConfig() *config {
 		}
 	}
 
+	// Opt-in vendor-charts. Same helper used by Serve() when wiring the
+	// bundle handler, so both surfaces read exactly one env value.
+	cfg.AllowVendorCharts = allowVendorChartsFromEnv()
```
njhensley, line 103: **🟡 Minor — config.AllowVendorCharts is a dead field; env read twice**

`parseConfig()` sets `cfg.AllowVendorCharts = allowVendorChartsFromEnv()`, but that field is read only by config_test.go — production wires the handler from an independent second `allowVendorChartsFromEnv()` call in serve.go:157. Two sources of truth for one knob (contrast `cfg.Address`, which is genuinely consumed at server.go:116).

**Blast radius:** Latent trap: a future maintainer wiring behavior off `cfg.AllowVendorCharts` (the documented config-extension pattern) would find it inert; an unparseable value warns twice.

**Fix:** Pick one source: thread `s.config.AllowVendorCharts` into `newBundleHandler` and drop the direct env read in serve.go, or delete the struct field.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+// private ranges, or a well-known cloud-metadata IP. Applies to http,
+// https, and oci:// schemes uniformly.
+//
+// The check is a fail-safe pre-filter, not a complete SSRF mitigation.
+// DNS rebinding between this resolution and `helm pull`'s re-resolution
+// remains an open surface until the fetch is moved in-process and the
+// resolved IP is pinned into the request. Operators exposing aicrd
+// beyond a trusted network should also gate the vendor path behind an
+// explicit opt-in (see server config: AllowVendorCharts).
+func checkEgressPolicy(ctx context.Context, repoURL string) error {
```
njhensley, line 366: **🟡 Minor — Egress-policy rejections are not logged or metered server-side**

`checkEgressPolicy` returns a well-detailed error but never `slog`s it; it surfaces as `ErrCodeInvalidRequest` → 400, and `WriteErrorFromErr` embeds the cause (incl. the `cloud-metadata` reason class) into the **client** 4xx response body without logging. So the reason class the disallowedIPReason comment says an operator 'needs to see in logs to recognize a credential-theft attempt' is delivered to the caller, not any operator log; there is also no rejection counter.

**Blast radius:** SSRF/metadata-theft probing against an opted-in server is invisible operator-side — the commit's stated detection story isn't delivered.

**Fix:** Emit `slog.Warn("vendor-charts egress rejected", "host",…, "ip",…, "reason",…, "requestID",…)` at the rejection point and add a rejection counter to the metrics.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -131,6 +136,35 @@ func (p *CLIChartPuller) Pull(ctx context.Context, c Component) ([]byte, VendorR
...
+	//   - DNS rebinding between the pre-check and helm's own re-resolution
+	//   - helm's HTTP redirects during tarball fetch (subprocess is not
+	//     intercepted)
+	//   - OCI registry redirects and blob-GETs (protocol not intercepted)
+	//
+	// A follow-up will move the tarball fetch in-process to close all three.
+	if err := checkEgressPolicy(ctx, c.Repository); err != nil {
+		return nil, VendorRecord{}, "", err
+	}
+	if err := resolveAndValidateHTTPIndex(ctx, c); err != nil {
```
njhensley, line 189: **🟡 Minor — Headline SSRF vector (public index → private tarball) not tested end-to-end through Pull()**

The 'public index.yaml points at a private/metadata tarball → reject' scenario the PR headlines is covered only at the `resolveAndValidateHTTPIndex` unit level (via `withMockIndexFetcher`). No test drives `(*CLIChartPuller).Pull()` with a public `Repository` + a mock index returning a private/metadata tarball URL and asserts `InvalidRequest`. `SSRFFilterWired` rejects at layer 1 (loopback repo) and `NoBinary` exercises the layer-2 pass path only.

**Blast radius:** The 'key vector closed' claim isn't proven at the entrypoint; the 3-line wiring could regress undetected.

**Fix:** Add one Pull() case: `Repository: https://93.184.216.34/charts`, mock index fetcher returning a `169.254.169.254` tarball URL, assert `ErrCodeInvalidRequest` + 'cloud-metadata'.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+		// callers can distinguish "someone tried to smuggle us into a
+		// private range" from "upstream network is flaky." Anything else
+		// is a genuine transport/dial failure → Unavailable.
+		return nil, errors.PropagateOrWrap(err, errors.ErrCodeUnavailable,
+			fmt.Sprintf("vendor-charts: index pre-check GET %s failed", indexURL))
+	}
+	defer resp.Body.Close() //nolint:errcheck // response body is read-only; close error is not actionable.
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New(errors.ErrCodeUnavailable,
+			fmt.Sprintf("vendor-charts: index pre-check GET %s returned HTTP %d", indexURL, resp.StatusCode))
```
njhensley: **🟡 Minor — Every non-200 index pre-check status is classified Unavailable**

`defaultFetchIndexYAML` maps every non-200 (403, 404, 500, …) to `ErrCodeUnavailable`. Since the pre-check runs before `helm pull`, this is terminal and is what the client sees: an index 404 (wrong URL / chart-less repo) becomes a retryable 503, and a 403 isn't `Unauthorized`. This contradicts `classifyHelmCLIError` on the fallback path, which maps repo-not-found → `ErrCodeNotFound`.

**Blast radius:** Clients/automation branching on retryability get contradictory signals depending on which layer trips first.

**Fix:** Split status handling: 404 → NotFound, 401/403 → Unauthorized, 5xx/other → Unavailable.

pkg/defaults/timeouts.go
```diff
@@ -917,6 +917,43 @@ const (
...
+
+	// EnvServerAddress overrides ServerDefaultBindAddress. Set to "0.0.0.0"
```
njhensley: **🔵 Nitpick — Constant docstring conflates 'empty' with 'unset'**

`os.LookupEnv` returns ok=true for a present-but-empty value, so `AICR_SERVER_ADDRESS=""` binds all interfaces — which the doc table and PR body correctly document as an intentional escape hatch. But this constant's docstring says 'Empty (unset) preserves loopback-only', conflating the two: only *unset* preserves loopback.

**Blast radius:** Wording only; a reader trusting the docstring could mis-model the empty-value case.

**Fix:** Reword to distinguish unset (loopback) from empty-string (all interfaces).

pkg/bundler/deployer/localformat/vendor_test.go
```diff
@@ -152,3 +197,829 @@ func TestCLIChartPuller_NoBinary(t *testing.T) {
...
+// with realistic YAML payloads without depending on an httptest.Server
+// (which would bind loopback and get rejected by our own egress policy).
+func withMockIndexFetcher(t *testing.T, fn func(ctx context.Context, indexURL string) ([]byte, error)) {
+	t.Helper()
+	orig := fetchIndexYAML
+	fetchIndexYAML = fn
+	t.Cleanup(func() { fetchIndexYAML = orig })
+}
+
+func TestCheckEgressPolicy_LiteralIP(t *testing.T) {
```
njhensley, line 348: **🔵 Nitpick — IPv4-mapped IPv6 absent from the egress test matrix**

`disallowedIPReason` correctly catches `::ffff:127.0.0.1`/`::ffff:169.254.169.254` via Go's To4()/ip.Equal normalization, but no row in `TestCheckEgressPolicy_LiteralIP` pins it — exactly the kind of mapped-address handling an SSRF filter should assert rather than assume.

**Blast radius:** None today; a future net-stdlib change could silently regress unnoticed.

**Fix:** Add `https://[::ffff:127.0.0.1]/x` (expect loopback) and `http://[::ffff:169.254.169.254]/` (expect cloud-metadata) rows.

pkg/bundler/deployer/localformat/vendor_test.go
```diff
@@ -152,3 +197,829 @@ func TestCLIChartPuller_NoBinary(t *testing.T) {
 		t.Errorf("Code = %v, want %v", se.Code, errors.ErrCodeUnavailable)
 	}
 }
+
+// withMockResolver swaps the package-level lookupIP for the duration of a
```
njhensley, line 259: **🔵 Nitpick — Test-hook docstrings claim an unimplemented panic-on-double-install**

`withMockResolver`/`withMockIndexFetcher` docstrings state they 'panic on double-install', but the bodies only save/restore via `t.Cleanup`. The package-level `var lookupIP`/`var fetchIndexYAML` hooks are process globals; the pattern is safe only because no test calls `t.Parallel()`.

**Blast radius:** Comment drift; marking one of these table tests parallel later would race the shared global.

**Fix:** Correct the comment (or implement the claimed guard) and note these tests must not run in parallel.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+// versionSpec is not a valid pinned semver, treats it as a Helm-
+// compatible semver constraint (^1.2, ~1.2, 1.2.x, >=1.2 <2, etc.) and
+// resolves to the highest matching entry version, matching helm's own
+// resolution behavior for `helm pull --version <constraint>`.
+//
+// Returns the URL list, the resolved concrete version string (equal to
+// versionSpec for exact matches; the index entry's version for
+// constraints — useful for downstream error messages), and an error.
+// Errors carry a StructuredError code so callers can PropagateOrWrap.
+func selectChartURLs(entries []helmIndexEntry, chartName, versionSpec string) ([]string, string, error) {
```
njhensley, line 852: **🔵 Nitpick — selectChartURLs resolver-divergence not in the documented residual list; constraint slow-path largely unreachable**

`selectChartURLs` re-implements helm's version resolution; helm re-fetches the index and re-resolves independently, so the URL egress-validated may differ from the one helm pulls — a residual not in the Pull() residual comment. Separately, `validateForPull`'s `safeChartVersionRE` (leading char must be alphanumeric) rejects `^1.2`/`~1.2`/`>=1.2 <2` before they reach the constraint slow-path, so only `N.N.x`-style wildcards actually reach it — the doc comment overstates constraint parity with helm, and the pre-check is stricter than the tool it guards.

**Blast radius:** Defense-in-depth that can bit-rot as helm evolves; a user supplying `^1.2` with --vendor-charts gets a hard 400.

**Fix:** Add resolver-divergence to the documented residuals, and narrow the selectChartURLs doc comment to the reachable `N.N.x` subset (or widen the regex to admit constraint operators if constraint pulls are a requirement).

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+// attachHelmBasicAuth sets HELM_REPOSITORY_USERNAME/PASSWORD on req only
+// when: (a) the operator has opted in by setting AICR_HELM_REPOSITORY_HOST,
+// (b) the request scheme is HTTPS (never leak creds in cleartext), and
+// (c) the request host equals the configured allowed host. Any mismatch
+// suppresses the credentials silently — a WARN log would be too chatty on
+// the redirect path where suppression is expected. Go's stdlib http.Client
+// separately strips Authorization on cross-origin redirects; this function
+// is the initial-URL gate that stops the leak at the source.
+func attachHelmBasicAuth(req *http.Request) {
+	user := os.Getenv("HELM_REPOSITORY_USERNAME")
```
njhensley: **🔵 Nitpick — helm env var names used as raw literals while sibling uses a named constant**

`attachHelmBasicAuth` reads `HELM_REPOSITORY_USERNAME`/`HELM_REPOSITORY_PASSWORD` as bare literals while the AICR-owned companion uses `defaults.EnvHelmRepositoryHost`.

**Blast radius:** Minor inconsistency with the named-constants rule (mitigated: these are helm-native names).

**Fix:** Add `defaults.EnvHelmRepositoryUsername/Password` for symmetry, or a one-line comment noting they are helm-native names spelled verbatim.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+var fetchIndexYAML = defaultFetchIndexYAML
+
+// defaultFetchIndexYAML performs the production HTTP GET. Every redirect
+// hop is passed back through checkEgressPolicy so a public repo cannot
+// redirect the index-fetch itself into the private range. Response body
+// is capped at defaults.HelmChartIndexBodyLimit.
+func defaultFetchIndexYAML(ctx context.Context, indexURL string) ([]byte, error) {
+	client := &http.Client{
+		Transport: defaults.NewHTTPTransport(),
+		Timeout:   defaults.HelmChartPullTimeout,
```
njhensley: **🔵 Nitpick — Index pre-check reuses the 5m HelmChartPullTimeout for a small GET**

`defaultFetchIndexYAML` sets `Timeout: HelmChartPullTimeout` (5m) and also runs under Pull's already-5m ctx, so a slow index GET can eat most of helm's budget. Harmless in practice (transport-level timeouts bound the connection phase) but semantically off.

**Blast radius:** Negligible; naming/observability clarity.

**Fix:** Introduce a smaller `HelmChartIndexFetchTimeout`, or comment that the index fetch deliberately shares the pull budget.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			if len(via) >= defaults.HelmChartIndexMaxRedirects {
```
njhensley, line 601: **🔵 Nitpick — defaultFetchIndexYAML redirect-cap branch is untested**

The `len(via) >= HelmChartIndexMaxRedirects` guard is never exercised; the redirect-to-private-IP subtest trips `checkEgressPolicy` first, never the count cap.

**Blast radius:** Low (helm re-does its own fetch anyway).

**Fix:** Add an httptest handler emitting N+1 self-redirects and assert InvalidRequest / 'exceeded … redirects'.

pkg/bundler/deployer/localformat/vendor_test.go
```diff
@@ -134,11 +141,49 @@ func TestClassifyHelmCLIError(t *testing.T) {
...
 
+// TestCLIChartPuller_SSRFFilterWired is the end-to-end proof that
+// checkEgressPolicy runs INSIDE Pull() — not just as a standalone
+// function that could accidentally be bypassed. A loopback URL must be
+// rejected as ErrCodeInvalidRequest before helm-binary discovery, so
+// even a machine with helm on PATH cannot be steered at the local
+// network from the vendor path.
+func TestCLIChartPuller_SSRFFilterWired(t *testing.T) {
+	// HelmBin points at /bin/true so if the SSRF filter DIDN'T run, we'd
+	// see ErrCodeNotFound (helm succeeds, no .tgz produced) — a distinct
```
njhensley: **🔵 Nitpick — SSRFFilterWired comment misstates the bypass signal (assertion is sound)**

The comment claims that if the filter didn't run we'd see `ErrCodeNotFound` (helm via /bin/true produces no .tgz). Actually `readSingleTgz` on an empty dir returns `ErrCodeInternal`, and with the filter removed the index pre-check would dial 127.0.0.1/index.yaml → Unavailable. The assertion (`!= InvalidRequest`, mentions 'loopback') is still correct.

**Blast radius:** Comment accuracy only.

**Fix:** Correct the comment or drop the specific-code claim.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+	matched, resolvedVersion, err := selectChartURLs(entries, chartName, c.Version)
+	if err != nil {
+		return err
+	}
+	base, err := url.Parse(c.Repository)
+	if err != nil {
+		return errors.Wrap(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("vendor-charts: cannot parse repository URL %q", c.Repository), err)
+	}
+	for _, raw := range matched {
```
njhensley, line 787: **🔵 Nitpick — Matched index entry with empty urls passes egress validation vacuously**

If the selected index entry has no `urls`, the `for _, raw := range matched` loop runs zero times and `resolveAndValidateHTTPIndex` returns nil — the pre-check contributes no defense for that entry. Not exploitable (helm has nothing to fetch and errors downstream).

**Blast radius:** None (security posture holds because helm fails, not because the check caught anything).

**Fix:** Treat an empty URL list for a matched version as NotFound/InvalidRequest for a clearer failure.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +233,432 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+// resolution behavior for `helm pull --version <constraint>`.
+//
+// Returns the URL list, the resolved concrete version string (equal to
+// versionSpec for exact matches; the index entry's version for
+// constraints — useful for downstream error messages), and an error.
+// Errors carry a StructuredError code so callers can PropagateOrWrap.
+func selectChartURLs(entries []helmIndexEntry, chartName, versionSpec string) ([]string, string, error) {
+	// Fast path 1: exact string equality. Some indexes carry non-semver
+	// aliases (e.g., "latest") that must not be silently mapped elsewhere.
+	for _, e := range entries {
```
njhensley, line 824: **🔵 Nitpick — Duplicate index version entries: pre-check validates first exact match, helm may resolve another**

Fast-path 1 returns the first entry whose Version string-equals versionSpec. A malformed/attacker-controlled index with duplicate version entries (public URL first, private second) would have only the first validated while helm's independent resolution could select the other — the same TOCTOU class already documented as a residual, and requiring an attacker who already controls the (egress-validated public) repo host.

**Blast radius:** Very low real-world exploitability.

**Fix:** Validate all entries matching the resolved version, not just the first exact-string match.

njhensley (changes requested):
## 🔍 Multi-Persona Review — SSRF / vendor-charts hardening

**Method:** 5 independent persona reviewers (Security · Correctness · Test-coverage · Operability/CI-DX · Domain/Architecture) fanned out against the diff, then an adversarial senior meta-reviewer re-derived every finding from the resolved code at `00ceb70b`. Two persona findings were refuted (see below) and are not raised as issues.

|:---:|:---:|:---:|:---:|
| 1 | 0 | 8 | 11 |

### Overall assessment

The SSRF / resource-exhaustion hardening itself is **well-constructed and unusually honest about its own residuals** — the layered egress policy, the `+1` bounded reads, the opt-in gate ordering (fires before body decode, both v1+v2, fails closed), the HTTPS+host-match credential-suppression gate, and the split-horizon fail-closed logic all held up under adversarial re-derivation. Test coverage of the new logic is genuinely strong.

**However, one confirmed 🔴 Blocker is unrelated to SSRF:** the bundled default-bind flip to `127.0.0.1` breaks every in-tree deployment surface via kubelet `httpGet` probes and kube-proxy, which dial the pod IP — not loopback — and no in-tree manifest sets `AICR_SERVER_ADDRESS`. That's why this is **Request changes**; the remaining findings are Minor/Nitpick and several could be follow-ups.

Inline comments follow, each tagged with its tier. Two findings without a diff-anchorable line are folded in here:

- 🟡 **Minor — Bound interface logged only at Debug** (`pkg/server/server.go` startup): after a change whose whole point is *which interface binds*, there is no INFO line naming the bound address. When a probe/Service fails post-upgrade, the operator has no log signal pointing at the bind address. Promote the startup line to `slog.Info(\"server listening\", \"address\", …)` — the vendor-charts-enabled INFO log sets the precedent.
- 🔵 **Nitpick — Docstring wrongly claims `helm pull` authenticates via `HELM_REPOSITORY_USERNAME/PASSWORD` env vars** (`pkg/bundler/deployer/localformat/vendor.go:104-108`, and the auth-error hint at `:729`): empirically false — `helm pull --repo` with those set sends **no** `Authorization` header (verified against helm v4.2.3 on both the index and tarball GETs). Security-wise this is fail-safe (the subprocess never transmits them). But it's a real functional gap: an authed private HTTP repo will pass the authenticated pre-check yet fail `helm pull` with 401. Fix the comment, or pass `--username/--password` to the subprocess gated on the same `AICR_HELM_REPOSITORY_HOST` match.

- **"helm-pull subprocess exfiltrates `HELM_REPOSITORY_USERNAME/PASSWORD` to a caller-supplied repo"** — **refuted by empirical test**: helm v4.2.3 `helm pull --repo` ignores those env vars and sends no `Authorization` header, so there is no exfiltration path. The only real residual is the docstring inaccuracy noted above.
- **IPv4-mapped IPv6 bypass** — refuted: `::ffff:10.0.0.1` / `::ffff:127.0.0.1` / `::ffff:169.254.169.254` are all correctly caught via Go's `To4()`/`ip.Equal` normalization.
- **`checkEgressPolicy` drift across its 3 call sites** — single source of truth, no drift.
- **`readBoundedTgz` boundary math & `PropagateOrWrap` code preservation** — verified correct.

---
*Reviewed with a multi-persona + adversarial-meta-review workflow. Findings were confirmed against the resolved code, not taken from a single reviewer's claim.*

framsouza:
Multi-persona + CodeRabbit review pass 3 addressed in `98e7854a` (rebased onto latest upstream/main). All 🔴/🟠 findings fixed; nearly all 🟡/🔵 fixed; the two knowingly-deferred ones are called out below.

### 🔴 Blocker — reverted default bind

`ServerDefaultBindAddress` back to `""` (all interfaces). The loopback default broke kubelet livenessProbe/readinessProbe and kube-proxy Service routing because both dial the pod IP directly, not loopback. Operators who want a tighter bind on a sidecar or bare-host deployment set `AICR_SERVER_ADDRESS=127.0.0.1` explicitly. The SSRF hardening never depended on the bind address — the opt-in gate + egress policy + index pre-check + artifact cap are the acute controls. Deployment doc updated. Live-verified pod-IP + loopback both reachable.

### 🟡/🔵 fixes

| Finding | Fix |
|---|---|
| **Docstring on `helm pull` env-var auth is empirically false** (v4.2.3 sends no `Authorization` header) | Docstring on `CLIChartPuller.Pull` rewritten to state helm CLI does NOT read the env vars for `helm pull --repo`; the vendor path's own pre-check is the only consumer, gated by `AICR_HELM_REPOSITORY_HOST`. |
| **helmIndexEntry declared twice** (anonymous struct in `helmIndex` + alias) | One named struct type, referenced from both `helmIndex.Entries` and `selectChartURLs`. |
| **Trailing slash on base URL** — `https://host/charts` (no `/`) causes RFC 3986 to drop the last segment during `ResolveReference` | Force trailing slash before parsing base; matches helm's `repo.ResolveReferenceURL`. New `TestResolveAndValidateHTTPIndex_RelativeURLResolution`. |
| **Two-segment version `1.2` returned NotFound** — Masterminds `NewVersion` coerces to `1.2.0`; helm treats as constraint `>=1.2.0 <1.3.0` | Switched to `StrictNewVersion` with v-prefix stripping so partial specs fall through to the constraint path. New test cases: `"1.2"` → `1.2.7`, `"1"` → `1.5.0`. |
| **Every non-200 index response → Unavailable** | 4xx → `InvalidRequest` (client-error class), 5xx → `Unavailable` (retryable). New `TestDefaultFetchIndexYAML_4xx` with 5 client + 2 server cases. |
| **`config.AllowVendorCharts` dead field / env read twice** | `Serve()` now reads via the parsed `cfg.AllowVendorCharts`; direct `allowVendorChartsFromEnv()` call removed. `parseConfig()` remains the single site consulting the environment. |
| **`withMockResolver` promises panic on double-install; body doesn't check** | Mutex-guarded sentinel; nested install calls `t.Fatal` loudly. |
| **IPv6 cloud-metadata + IPv4-mapped test rows** | Added: `fd00:ec2::254`, `fe80::a9fe:a9fe`, and `::ffff:{127.0.0.1,10.0.0.1,169.254.169.254}` — pins that `To4()` normalization catches mapped disallowed classes. |
| **Egress-policy rejections silent server-side** | `slog.WarnContext` at both the outer repository-URL rejection and the per-tarball-URL rejection — component, host, and reason logged. |
| **Reject empty urls list vacuously passing egress** | Matched entry with no URLs → `NotFound` up-front with a clear message. New test case. |
| **Index pre-check reuses 5m HelmChartPullTimeout for small GET** | New `defaults.HelmChartIndexPreCheckTimeout = 30s`. |
| **Helm env-var literals repeated** | New `defaults.EnvHelmRepositoryUsername` / `EnvHelmRepositoryPassword` alongside `EnvHelmRepositoryHost`. `TestAttachHelmBasicAuth_ConstantsUsed` pins values. |
| **Stale test-symbol reference in config_test.go comment** | Removed the dead `TestParseConfig_AllowVendorCharts` reference. |
| **api-reference vendor-charts row missing `AICR_HELM_REPOSITORY_HOST`** | Row rewritten to require the host gate + HTTPS + host match, and to state helm CLI does not honor the env vars itself. |
| **Deployment env-block table** | Updated for the reverted bind default; added `AICR_HELM_REPOSITORY_HOST` and honest annotation on the username/password vars. |

### 🔵 Knowingly deferred (with reasons)

- **Bound-interface startup log promoted to Info** — sensible but the existing Debug-level startup log carries the value; adding a redundant Info line for a policy-neutral field crowds an already-noisy startup log path. Deferred to a follow-up that reworks the startup-log surface holistically.
- **Duplicate index version entries** (pre-check picks first exact match, helm may resolve another) — pre-check aims to catch obvious SSRF, not to be a byte-exact simulator of helm's resolver. This is a real divergence but the failure mode (pre-check allows a URL helm actually fetches from a different mirror) is safe; the reverse (pre-check rejects a URL helm never fetches) would be the risk case and does not apply. Deferred as the in-process-tarball-fetch follow-up will make this moot.

### Validation

- `make lint` clean.
- `go test -race` on 19 touched + adjacent packages passes.
- Coverage: `pkg/bundler/deployer/localformat` 77.8% → 78.1%; `pkg/server` 82.4% (unchanged); `pkg/defaults` 100% (unchanged).
- Live-verified: default bind reaches both loopback and pod IP; vendor-charts gate still fires; `AICR_ALLOW_VENDOR_CHARTS=yes` correctly stays off with WARN log.

framsouza:
Both new findings addressed in `f19ce08f`.

### Finding 1 — resolved-URL shape assertion

**Verified valid.** The previous `TestResolveAndValidateHTTPIndex_RelativeURLResolution` only asserted "no error" as a proxy for "trailing-slash fixup worked" — a regression that dropped the repo directory segment could still land on a public IP and pass egress, silently passing the test.

**Fix.** New package-level `checkResolvedChartURL` hook alongside `lookupIP` and `fetchIndexYAML`; defaults to `checkEgressPolicy`, swapped in tests via a new `withMockResolvedURLChecker` helper. `resolveAndValidateHTTPIndex` now calls the hook instead of `checkEgressPolicy` directly. The updated test captures the exact URL and asserts by string equality:

```go
want := "https://93.184.216.34/nested/path/charts/victim-chart-1.2.3.tgz"
if len(captured) != 1 || captured[0] != want { ... }
```

Added a sibling `TestResolveAndValidateHTTPIndex_AbsoluteURLKeepsHost` that pins absolute URLs are passed through unchanged (so the trailing-slash fixup does not accidentally reshape entries whose `urls:` list carries a full absolute URL).

### Finding 2 — redact credential-bearing URLs

**Verified valid.** A caller can submit `Repository: "https://user:secret@repo.example/charts"` via `POST /v1/bundle`, and multiple call sites embedded the raw string in slog Warn lines and in `%q`-formatted error messages returned to the client.

**Fix.** New `redactURL(raw string) string` helper wraps `net/url.URL.Redacted()` — parses the URL and returns the string with password replaced by `xxxxx`; returns the raw input unchanged when parsing fails (a malformed string that reaches this helper is unlikely to be well-formed userinfo). Applied to every URL-embedding site:

| Site | Redacted field |
|---|---|
| `Pull()` slog.Warn (repo egress rejection) | `c.Repository` |
| `resolveAndValidateHTTPIndex` slog.Warn (per-tarball rejection) | `c.Repository`, resolved URL |
| `checkEgressPolicy` parse-fail + no-host errors | `repoURL` |
| `defaultFetchIndexYAML` — build-req, Do-fail, non-200, body-read, size-cap | `indexURL` |
| `resolveAndValidateHTTPIndex` — YAML error, missing entry, base parse, malformed URL, tarball rejection | `indexURL`, `c.Repository`, `raw`, `resolved` |
| `classifyHelmCLIError` — not-found, unauth, timeout, unreachable, generic | `c.Repository` |
| `validateForPull` — IsOCI mismatch, invalid URL, wrong scheme | `c.Repository` |

Also updated the classifyHelmCLIError unauthorized-hint message to point at "chart repository authentication in the deployment docs" instead of naming `HELM_REPOSITORY_USERNAME`/`HELM_REPOSITORY_PASSWORD` — those env vars are consumed only by the pre-check's own gated attachment, not by the helm CLI itself (verified against v4.2.3), so the previous hint was misleading.

**Test coverage:**
- `TestRedactURL` — 6-case pure-function contract (empty, no-userinfo, password-masked, user-only, oci-scheme, resolved-tarball URL). Belt-and-braces: asserts the literal string `"secret"` never survives the redactor.
- `TestVendorErrorPathsRedactCredentials` — 5-case boundary test that drives real error paths (`checkEgressPolicy` no-host, `validateForPull` IsOCI mismatch, `resolveAndValidateHTTPIndex` missing-entry + malformed-YAML, `classifyHelmCLIError` chart-not-found) with a sentinel secret and asserts it never appears in the surfaced error. Rather than list every call site (which drifts), this locks the boundary invariant.

**Scope note:** `URL.Redacted()` masks the userinfo portion (`user:xxxxx@`) but not query-string tokens (`?token=X`). That's consistent with the reviewer's named helper and with what a chart repository URL is typically shaped like. Query-string auth is a distinct attack surface — if a follow-up wants to cover it, a separate helper that strips known auth params (`token`, `access_token`, `sig`) would be the right fit.

### Validation

- `make lint` clean.
- `go test -race` on all touched + adjacent packages passes.
- Coverage on `pkg/bundler/deployer/localformat` 78.1% → 78.3%; `pkg/server` and `pkg/defaults` unchanged.
- **Live-verified with a real credential-bearing URL:** `POST /v1/bundle?vendor-charts=true` with `Repository: "https://user:SECRET_TOKEN_XYZ@127.0.0.1/repo"` (opt-in enabled) returns 400 (loopback rejection); both response body and server log show `"https://user:xxxxx@127.0.0.1/repo"` with `SECRET_TOKEN_XYZ` nowhere present.

framsouza:
Finding addressed in `283a9602` — plus a coupled leak I spotted while implementing.

### Finding — redactURL parse-failure placeholder

**Verified valid.** My previous implementation returned `raw` when `url.Parse` failed. Go's `url.Parse` is lenient but *does* reject: control characters in the host (`\x00`, `\x7f`), unclosed IPv6 brackets, invalid percent escapes, missing scheme, and a few other real inputs. All of these can carry `user:secret@` and reach the helper.

**Fix.** `redactURL` now returns a fixed `unparseableURLPlaceholder = "[redacted: unparseable url]"` on parse failure. Test coverage extended with three parse-failing credential-bearing cases (control char, unclosed bracket, invalid escape). The generic `strings.Contains(got, "secret")` assertion stays on every case so a future regression can't sneak past.

### Adjacent leak I fixed at the same time

While adding the boundary test for the parse-failure path, I noticed `redactURL` alone was NOT sufficient. `net/url.Parse` returns a `*url.Error` whose `.Error()` embeds the raw URL verbatim:

```
parse "https://user:secret@repo\x00.example/x": net/url: invalid control character in URL
```

At four wrap sites (`checkEgressPolicy`, `validateForPull`, and two in `resolveAndValidateHTTPIndex`) my code passed that error as `errors.Wrap(..., err)`. `StructuredError.Error()` re-emits `.Cause` via `%v`, so the secret leaks through the cause chain even after `redactURL` masks the visible instance. `redactURL` alone would give a false sense of safety.

**Fix.** At each of those four sites, dropped the parse-error cause via `errors.New` instead of `errors.Wrap`. The parse error's detail was "invalid URL syntax" — no information the message text doesn't already convey. All four sites now carry a comment pointing back to the reason.

Extended `TestVendorErrorPathsRedactCredentials` with two boundary cases (`checkEgressPolicy` and `validateForPull` parse-failure paths) that would have failed without the drop-cause fix. Verified by running the boundary test with only the `redactURL` change first — the two new cases failed, confirming the coupled leak was real.

### Scope note

The reviewer asked for a `redactURL` change + one test case. I extended to also fix the four cause-chain leaks because the boundary test they're implicit in (the parse-failure path only makes sense as a defense if the cause is also dropped) doesn't hold otherwise. Both changes together form the actual defense; either alone is incomplete.

### Validation

- `make lint` clean.
- `go test -race` on all touched + adjacent packages passes.
- Coverage on `pkg/bundler/deployer/localformat` 78.3% → 78.4%.
- `TestRedactURL` — 9/9 subtests pass (3 new parse-fail cases + 6 pre-existing).
- `TestVendorErrorPathsRedactCredentials` — 7/7 subtests pass (2 new parse-fail cases + 5 pre-existing).

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +252,571 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+	// case on Linux) an alias for localhost that would sidestep the
+	// loopback check above. Only matches IPv4 forms via To4().
+	if v4 := ip.To4(); v4 != nil {
+		for i := range martianIPv4Ranges {
+			if martianIPv4Ranges[i].Contains(v4) {
+				return "martian"
+			}
+		}
+	}
+	return ""
```
njhensley, line 571: **🟡 Minor — F3 (partial) — NAT64/6to4 IPv6 embedding still bypasses egress filter**

Martian IPv4 (0.0.0.0/8, 240.0.0.0/4, broadcast) is now correctly rejected — the martian half of the prior finding is closed. But NAT64 `64:ff9b::/96` (e.g. `64:ff9b::a9fe:a9fe` = 169.254.169.254) and 6to4 `2002::/16` are global-unicast IPv6, so `ip.To4()` returns nil and they fall through to the `return ""` allow. On a dual-stack cluster with a NAT64/DNS64 gateway (a real GKE/EKS topology) these reach metadata/private endpoints.

**Blast radius:** SSRF to metadata/private ranges where NAT64/6to4 translation infrastructure is present; bounded by that precondition, hence Minor.

**Fix:** Decode the embedded v4 (64:ff9b:: low 32 bits; 2002:: bytes 2-5) and recurse disallowedIPReason on it, or reject these documented-translation IPv6 prefixes outright.

docs/user/api-reference.md
```diff
@@ -546,7 +546,7 @@ Generate deployment bundles from a recipe.
...
...k, link-local, private, or cloud-metadata IPs are rejected with `400 INVALID_REQUEST`, and vendored artifacts are capped at 64 MiB. **Private HTTP(S) repository credentials:** the aicrd pre-check sends `HELM_REPOSITORY_USERNAME`/`HELM_REPOSITORY_PASSWORD` (as HTTP Basic auth) ONLY when `AICR_HELM_REPOSITORY_HOST` is set to that repository's exact host, the request scheme is `https`, and the request host matches (case-insensitive). All three conditions must hold — an operator setting only the username/password env vars will get no credentials attached, preventing a caller-supplied `Repository` URL from harvesting the operator's helm credentials. (Note: the upstream `helm pull --repo` subprocess does not itself read these env vars — private HTTP repos require a prior `helm repo add --username --password` in the aicrd image or an SDK-based puller.) OCI credentials flow through the standard docker config (`~/.docker/config.json` or `$DOCKER_CONFIG`), exactly like `helm pull oci://...`. If prerequisites are missing the request fails with a structured error code (`SERVICE_UNAVAILABLE` / HTTP 503 for missing helm; a 4xx from the index pre-check surfaces as `INVALID_REQUEST` / HTTP 400). |
```
njhensley: **🟡 Minor — N1 (new) — api-reference contradicts the new pre-check status split**

The vendor-charts row still says "a 4xx from the index pre-check surfaces as `INVALID_REQUEST` / HTTP 400". But the index-status fix now returns NOT_FOUND/404 for a missing repo and UNAUTHORIZED for 401/403 — pinned by TestDefaultFetchIndexYAML_4xx. A client writing status-code branching against the endpoint is misled.

**Blast radius:** Operator-facing API doc drift introduced by this rework.

**Fix:** Restate the mapping by class: 404→NOT_FOUND/404, 401/403→UNAUTHORIZED, other 4xx→INVALID_REQUEST/400, 5xx→SERVICE_UNAVAILABLE/503.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -131,6 +147,43 @@ func (p *CLIChartPuller) Pull(ctx context.Context, c Component) ([]byte, VendorR
...
+	//   - Resolver divergence: selectChartURLs re-implements helm's semver
+	//     constraint resolution. helm re-fetches the index and re-resolves
+	//     independently at pull time, so the URL egress-validated here may
+	//     differ from the one helm actually pulls if the index changes
+	//     between calls or if the two resolvers pick differently under
+	//     ambiguous inputs (defense-in-depth that can bit-rot).
+	//
+	// A follow-up will move the tarball fetch in-process to close all four.
+	if err := checkEgressPolicy(ctx, c.Repository); err != nil {
+		slog.WarnContext(ctx, "vendor-charts: egress policy rejected repository URL",
```
njhensley, line 185: **🔵 Nitpick — F5 (partial) — no rejection counter; index redirect-layer rejection unlogged**

slog.WarnContext now fires at the repository (L179) and resolved-tarball (L721) rejections — the core "never slog'd" gap is closed. Residual: a redirect-hop rejection inside defaultFetchIndexYAML's CheckRedirect propagates up unlogged; there is no aggregate rejection counter/metric for alerting; and the reason class is still embedded in the client-visible 4xx body via WriteErrorFromErr.

**Blast radius:** Partial operator visibility into SSRF probing; blind spot on the redirect-fetch layer.

**Fix:** Log the redirect-layer rejection too, and add a reason-keyed rejection counter to the metrics.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -131,6 +147,43 @@ func (p *CLIChartPuller) Pull(ctx context.Context, c Component) ([]byte, VendorR
...
+	//
+	// This is NOT complete SSRF closure. The following residuals remain and
+	// are the operator's responsibility to close with a Kubernetes
+	// NetworkPolicy or an egress firewall — see the deployment docs:
+	//
+	//   - DNS rebinding between the pre-check and helm's own re-resolution
+	//   - helm's HTTP redirects during tarball fetch (subprocess is not
+	//     intercepted)
+	//   - OCI registry redirects and blob-GETs (protocol not intercepted)
+	//   - Resolver divergence: selectChartURLs re-implements helm's semver
```
njhensley, line 170: **🔵 Nitpick — F11 (partial) — constraint slow-path is production-dead but undocumented**

Resolver divergence is now a documented residual. But safeChartVersionRE rejects `^1.2`/`~1.2`/`>=1.2 <2`/`*` before Pull() reaches selectChartURLs, so the caret/tilde/range constraint branch is unreachable from the production request path — only partial-pins like `1.2`/`1.2.x` actually diverge. The residual comment overstates the reachable divergence surface.

**Blast radius:** Maintenance trap: a future maintainer loosening the regex without re-auditing egress; no exploit today.

**Fix:** Note in the residual comment that safeChartVersionRE currently gates out constraint syntax, so only partial-pin forms reach the constraint resolver.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +252,571 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+// resolution behavior for `helm pull --version <constraint>`.
+//
+// Returns the URL list, the resolved concrete version string (equal to
+// versionSpec for exact matches; the index entry's version for
+// constraints — useful for downstream error messages), and an error.
+// Errors carry a StructuredError code so callers can PropagateOrWrap.
+func selectChartURLs(entries []helmIndexEntry, chartName, versionSpec string) ([]string, string, error) {
+	// Fast path 1: exact string equality. Some indexes carry non-semver
+	// aliases (e.g., "latest") that must not be silently mapped elsewhere.
+	for _, e := range entries {
```
njhensley, line 824: **🔵 Nitpick — F17 (not addressed) — duplicate same-version index entries: only the first is validated**

Both fast-paths in selectChartURLs `return e.URLs` on the FIRST entry whose version matches (string-equal, then semver-equal). If an index carries two same-version entries with different urls lists (public first, private second), only the first is egress-validated while helm's independent re-resolution could select the other. No test covers a duplicate-version index.

**Blast radius:** Narrow TOCTOU requiring a poisoned/duplicate index; overlaps the documented resolver-divergence residual.

**Fix:** Return the URLs of ALL entries matching the resolved version (so every candidate is egress-checked), or reject an index with duplicate entries for that version.

njhensley (commented):
## 🔁 Re-review — SSRF / resource-exhaustion hardening of the bundle vendor path

**Method:** delta re-review. Three persona reviewers (Security/SSRF, Correctness/API-server, Test-coverage/Docs) ran in parallel against the resolved code, each dispositioning the prior 17 findings and hunting net-new issues; a senior meta-reviewer independently reproduced the crux claims from the code. Anchored to head `9739d207`.

**Disposition legend:** ✔️ Addressed · ◐ Partially addressed · ✖️ Not addressed · ⊘ No longer applicable

### Overall assessment

The PR was reworked into a fresh squashed commit (`6ca6fd7c`; the two later commits are `main` merges). **The prior 🔴 blocker is fully resolved** and 12 of 17 prior findings are cleanly addressed. Rather than patching the three in-tree manifests, the default bind was **reverted to `""` (all interfaces)** with loopback re-documented as opt-in — the correct call, since the SSRF posture never depended on the bind (the opt-in gate + egress policy + index pre-check + artifact cap are the controls). Tests pass on all three touched packages with `-race`.

Remaining items are **Minor/Nitpick residuals** — none blocking.

**Recommendation: Approve with comments.** 🔴 0 · 🟠 0 · 🟡 2 · 🔵 6

### Prior-feedback status (17)

| # | Tier | Finding | Disposition |
|---|------|---------|-------------|
| F1 | 🔴 | Loopback default bind breaks probes/Service/Cloud Run | ✔️ Addressed — default reverted to `""` |
| F2 | 🟡 | No migration note / docs inconsistent for bind flip | ⊘ No longer applicable |
| F3 | 🟡 | IP classification gaps (NAT64/6to4 + martian v4) | ◐ Partial — martian closed; NAT64/6to4 IPv6 open (inline) |
| F4 | 🟡 | `config.AllowVendorCharts` dead field | ✔️ Addressed — threaded into `newBundleHandler` |
| F5 | 🟡 | Egress rejections not logged/metered | ◐ Partial — `slog` added; no counter, redirect-layer unlogged (inline) |
| F6 | 🟡 | Headline SSRF vector not tested via `Pull()` | ✔️ Addressed |
| F7 | 🟡 | Every non-200 index pre-check → Unavailable | ✔️ Addressed — split 404/401-403/4xx/5xx |
| F8 | 🔵 | Docstring conflates empty vs unset | ✔️ Addressed |
| F9 | 🔵 | IPv4-mapped IPv6 absent from test matrix | ✔️ Addressed |
| F10 | 🔵 | Test-hook docstrings claim unimplemented panic | ✔️ Addressed |
| F11 | 🔵 | `selectChartURLs` resolver-divergence residual | ◐ Partial — regex dead-path undocumented (inline) |
| F12 | 🔵 | helm env-var raw literals | ✔️ Addressed |
| F13 | 🔵 | Index pre-check reuses 5m pull timeout | ✔️ Addressed — new 30s constant |
| F14 | 🔵 | redirect-cap branch untested | ✔️ Addressed |
| F15 | 🔵 | SSRFFilterWired comment misstates bypass signal | ✔️ Addressed |
| F16 | 🔵 | Matched entry with empty urls passes vacuously | ✔️ Addressed |
| F17 | 🔵 | Duplicate index version entries: only first validated | ✖️ Not addressed (inline) |

### Fully resolved / confirmed non-issues

F1, F4, F6, F7, F8, F9, F10, F12, F13, F14, F15, F16 — all independently re-verified as addressed against the current code. F2 dissolved by the default-bind revert. Minor cosmetic note on F10: the double-install guard is now real, but the docstrings say "panic" while `t.Fatal` uses `runtime.Goexit`.

### Summary

```
Prior:     ✔️ 12 addressed · ◐ 3 partial · ✖️ 1 not addressed · ⊘ 1 n/a
Surviving: 🔴 0 | 🟠 0 | 🟡 2 (F3, N1) | 🔵 6 (F5·F11·F17·N2·N3·N4)     Recommendation: Approve with comments
```

The blocker that drove the prior `CHANGES_REQUESTED` is gone and the fix is clean. The two Minor items (NAT64/6to4 gap; the api-reference status-code drift) are the ones worth addressing before merge; the Nitpicks are optional polish. See inline comments for detail.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +252,571 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+// helm's own basic-auth env vars if set. Tests inject canned bytes so the
+// URL-validation logic can exercise realistic index.yaml payloads without
+// depending on the developer's system resolver or network reachability.
+var fetchIndexYAML = defaultFetchIndexYAML
+
+// defaultFetchIndexYAML performs the production HTTP GET. Every redirect
+// hop is passed back through checkEgressPolicy so a public repo cannot
+// redirect the index-fetch itself into the private range. Response body
+// is capped at defaults.HelmChartIndexBodyLimit.
+func defaultFetchIndexYAML(ctx context.Context, indexURL string) ([]byte, error) {
```
njhensley, line 585: **🔵 Nitpick — N4 (new) — index fetch trusts caller-side egress check**

defaultFetchIndexYAML validates every redirect hop via checkRedirectTargetURL but never runs checkEgressPolicy on its own initial indexURL. Safe today only because Pull() checks c.Repository ~30 lines earlier and indexURL is built from the same host. A refactor that reorders Pull(), or a new caller of resolveAndValidateHTTPIndex, would fetch an unvalidated URL.

**Blast radius:** Latent SSRF-precondition coupling; no current exploit (single caller enforces ordering).

**Fix:** Add an idempotent checkEgressPolicy(ctx, indexURL) at the top of defaultFetchIndexYAML (or resolveAndValidateHTTPIndex) so the fetch fails closed regardless of caller ordering.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +252,571 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+		//   - 401 / 403       -> Unauthorized      (caller must fix creds; retry same request won't help)
+		//   - other 4xx       -> InvalidRequest    (caller-shaped problem)
+		//   - 5xx / everything else -> Unavailable (transient upstream)
+		code := errors.ErrCodeUnavailable
+		switch {
+		case resp.StatusCode == http.StatusNotFound:
+			code = errors.ErrCodeNotFound
+		case resp.StatusCode == http.StatusUnauthorized || resp.StatusCode == http.StatusForbidden:
+			code = errors.ErrCodeUnauthorized
+		case resp.StatusCode >= http.StatusBadRequest && resp.StatusCode < http.StatusInternalServerError:
```
njhensley, line 656: **🔵 Nitpick — N2 (new) — retryable 429/408 classified as non-retryable INVALID_REQUEST**

The generic 4xx arm maps every 4xx except 404/401/403 to ErrCodeInvalidRequest, so 429 (rate-limited) and 408 (request timeout) — the classic retryable 4xx — become non-retryable client errors. An orchestrator branching on the structured code won't retry a transient upstream rate-limit.

**Blast radius:** Low: only affects callers branching on the code for retry decisions against a private repo behind a rate limiter; opt-in path, no security impact.

**Fix:** Add `case 429 || 408: code = ErrCodeUnavailable` before the generic 4xx arm, mirroring the 5xx handling.

pkg/bundler/deployer/localformat/vendor.go
```diff
@@ -200,6 +252,571 @@ func (p *CLIChartPuller) detectVersion(ctx context.Context) string {
...
+	// is caught (we'd read exactly limit+1 bytes and reject) rather than
+	// silently truncated.
+	buf, err := io.ReadAll(io.LimitReader(f, limit+1))
+	if err != nil {
+		return nil, errors.Wrap(errors.ErrCodeInternal,
+			fmt.Sprintf("vendor-charts: read pulled chart %s", path), err)
+	}
+	if int64(len(buf)) > limit {
+		return nil, errors.New(errors.ErrCodeInvalidRequest,
+			fmt.Sprintf("vendor-charts: pulled chart %s exceeded limit of %d bytes during read",
```
njhensley, line 303: **🔵 Nitpick — N3 (new) — readBoundedTgz during-read race branch unreachable from tests**

Every over-cap case in TestReadBoundedTgz is caught by the Stat pre-check, so the belt-and-braces post-read "exceeded limit … during read" branch (which defends the Stat/Open grow race) has no covering test. The existing assertions are sound — this is purely a coverage gap on defense-in-depth code.

**Blast radius:** Test coverage only; not a correctness bug.

**Fix:** Annotate the branch as untestable, or add a seam that forces the file to grow between Stat and read.

njhensley (approved):
## ✅ Approve — SSRF / resource-exhaustion hardening (round-3 re-review)

**Method:** delta re-review of the new commit `df702f5e` since the prior COMMENTED review at `9739d207`. Verified each change against the resolved code, confirmed the new tests assert what they claim, and ran the package tests (`-race`, all green on `pkg/bundler/deployer/localformat`, `pkg/server`, `pkg/defaults`).

### Prior-feedback status — all 8 addressed ✔️

| # | Tier | Finding | Disposition |
|---|------|---------|-------------|
| F3 | 🟡 | NAT64/6to4 IPv6 embedding bypasses egress filter | ✔️ `decodeIPv6TransitionV4` decodes `64:ff9b::/96` + `2002::/16` and recurses `disallowedIPReason`; 6 test rows |
| N1 | 🟡 | api-reference contradicted the pre-check status split | ✔️ Row now spells the by-class mapping incl. 408/429→503 retryable |
| F5 | 🔵 | redirect-layer rejection unlogged | ✔️ `slog.WarnContext` on redirect-target rejection (metrics counter deferred — optional observability follow-up) |
| F11 | 🔵 | constraint slow-path dead but undocumented | ✔️ Residual comment notes `safeChartVersionRE` gates out constraint syntax |
| F17 | 🔵 | duplicate same-version entries: only first validated | ✔️ `selectChartURLs` returns the **union** of all matching entries' URLs; new duplicate-union test |
| N4 | 🔵 | index fetch trusted caller-side egress check | ✔️ `checkFetchTargetURL` fail-closed on the initial `indexURL` + test |
| N2 | 🔵 | retryable 429/408 → INVALID_REQUEST | ✔️ 408/429 → `Unavailable` case added + doc |
| N3 | 🔵 | during-read race branch untested | ✔️ Documented as intentional, currently-untestable Stat/Open-TOCTOU defense-in-depth |

### Assessment

A complete, high-quality response across three rounds: the round-1 🔴 blocker (loopback default bind) was resolved by reverting the default to all-interfaces with loopback re-documented as opt-in; round-3 closed every remaining Minor and Nitpick, each with a targeted test. The substantive fixes (F3 transition-address decoding; F17 same-version URL union) are implemented correctly — `decodeIPv6TransitionV4` guards against mapped-IPv4 and is depth-bounded, and `selectChartURLs` aggregates every mirror so all candidate URLs reach the egress check.

```
Cumulative: 🔴 blocker resolved · all Minor/Nitpick residuals resolved · 0 surviving findings
```

The only deferral is the optional Prometheus rejection counter (F5) — an observability enhancement, not a hardening gap. LGTM. 🚀

PR #1983, opened 2026-08-02
chore(dynamo-platform): coordinate NATS removal with Dynamo 1.4+ bump

mohityadav8:
## Summary

Coordinate the NATS removal with the Dynamo 1.4+ bump: pin `dynamo-platform` to 1.4.2, drop `global.nats.install` and the NATS storage-class/node-scheduling paths, bump Grove to v0.1.0-alpha.12, and update NATS-specific test/doc/conformance references — Dynamo 1.4+ defaults the request plane to TCP and the KV event plane to ZMQ.

## Motivation / Context

Upstream Dynamo 1.4+ disables bundled NATS by default (ai-dynamo/dynamo#11951) because the request plane now defaults to TCP and the KV event plane to ZMQ. `dynamo-platform` 1.4.2 additionally fixes a NIXL loader-path bug in the Frontend and SGLang Runtime images.

Fixes: #1836
Related: N/A

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update

## Component(s) Affected

- [x] Recipe engine / data (`pkg/recipe`)
- [x] Docs/examples (`docs/`, `examples/`)
- [x] Other: `tests/chainsaw/ai-conformance/*`, `tests/manifests/dynamo-vllm-smoke-test.yaml`, `validators/performance/testdata/inference/*`

## Implementation Notes

- `recipes/registry.yaml`: `dynamo-platform` pinned to `1.4.2`; `grove` pinned to `v0.1.0-alpha.12` (up from `v0.1.0-alpha.8`), with the required CRD migration documented in `docs/user/component-catalog.md` ("Upgrade Notes" > grove) — a generated bundle has no local chart checkout, so the executable path is `helm show crds oci://ghcr.io/ai-dynamo/grove/grove-charts --version v0.1.0-alpha.12 | sed -n '/^---$/,$p' | kubectl apply --server-side --force-conflicts -f -` (the `sed` filter is required: `helm show crds` writes `Pulled:`/`Digest:` lines to stdout for an OCI chart, which `kubectl` otherwise rejects), not a `<chart>/crds/` path. Verified live end-to-end on a real EKS cluster: dynamo-operator, grove-operator, and kai-scheduler gang-scheduling all confirmed healthy together post-migration.
- `recipes/components/dynamo-platform/values.yaml`: removed `global.nats.install: true` and the `nats.config.jetstream` PVC block; `global.nats.install` left unset so the chart's new default (`false`) applies.
[...]

PR #2172, opened 2026-08-12
fix(skills): cross-review sandbox probe and companion resolution

yuanchen8911:
## Summary

Two fixes to the `aicr-cross-review` skill, both from failures observed in a single live review round: the CodeRabbit lane's sandbox probe missed network denial and hung for a full 10-minute timebox, and the Codex lane's recovery note left the orchestrator without a way to resolve the companion path.

## Motivation / Context

**CodeRabbit lane — the probe tested the wrong dimension.** STEP 1 decided whether STEP 2 needs a sandbox bypass by writing a probe file under `~/.coderabbit`. STEP 2 then read that result as proof there was "nothing to bypass."

Writability does not imply the `coderabbit.ai` hosts are reachable. The sandbox can deny the CLI two independent ways, and both present identically — a stall at `connecting_to_review_service` until the timebox kills the call:

- **Filesystem denial** — the CLI cannot create its log or review store, and writes no log at all.
- **Network denial** — the CLI logs `403 Forbidden` on `GET https://cli.coderabbit.ai/public-configs.json` with `"data":"Connection blocked by network allowlist"`, then retries `wss://ide.coderabbit.ai/ws` every 30 s.

The broken combination is the one this skill's own guidance produces. Its SANDBOX section recommends adding `~/.coderabbit` to `filesystem.allowWrite` to remove per-round approval prompts, and never mentions the network allowlist. A machine in that state probes "sandboxed is fine" and then loses a full timebox on every run.

**Codex lane — `$comp` never reaches the orchestrator.** The recovery note tells the orchestrator to poll a live job with the companion status command, and the lane's own failure message quotes `node "$comp" result <job-id> …`. But `comp` is defined only inside the lane prompt and cannot cross Bash calls, so the orchestrator has to invent the resolution. Inventing it without `-t` sorts by version *name* and can select a stale cached companion; polling a job with a different companion version than dispatched it returns misleading results — `status` finding a job that `result` then reports as unknown — which reads exactly like a dead job and is not one.

Fixes: N/A
Related: N/A

## Type of Change

- [ ] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [x] Other: `.agents/skills/aicr-cross-review/` (agent skill definition; no shipped code)

## Implementation Notes

**Both sandbox dimensions are probed and ANDed.** The reachability half tests both hosts the CLI needs: the `cli.coderabbit.ai` config endpoint it fetches on startup, and `ide.coderabbit.ai`, the host its review session's WebSocket then connects to. The allowlist is per-host, so an entry naming only the config host — the obvious edit for a reader who saw it in the probe — would pass a single-host check and still hang STEP 2 on the WebSocket connect. A plain HTTPS GET tests the wss host because the allowlist rule is host-scoped, not protocol-scoped: the site answers `200` at `/` when reachable, and a sandbox block answers a proxy `403` that `curl -f` turns into a failure.

**`curl -f` is load-bearing.** Without it, a `403 Connection blocked by network allowlist` still exits 0, and the probe would report a blocked host as reachable — reintroducing the bug it fixes. Both host checks carry it.

**The result is fail-safe in the cheap direction.** A false negative costs one approval prompt; a false positive costs a full 10-minute timebox. So either check failing sends STEP 2 down the bypass path.

**Diagnosis guidance corrected.** The previous text said a stall with no log file means sandbox denial, implying the presence of a log rules the sandbox out. That is wrong in the network case: absence of a log separates *filesystem* denial from *network* denial, not sandbox denial from a cloud problem.

**Sandbox-bypass surface is unchanged.** STEP 2 is still the only call that may bypass, and it is still a lone `coderabbit review` command with no git or working-copy operation.

**The companion fix is documentation only.** It gives the orchestrator the same `ls -t` form the lane already uses at `workflow.mjs:224` and `:242`; no lane behavior changes.

## Testing

```bash
make check-agents-sync license check-docs-filenames check-docs-mdx check-docs-mdx-parse
```

All five pass. `make qualify` was not run and `lint-go` is not applicable: this PR changes two files under `.agents/skills/`, with no Go, YAML, recipe, chart, or `docs/` content, so the Go, e2e, coverage, scan, and api-diff lanes have nothing to exercise. `check-agents-sync` is included because it is the gate that governs agent-instruction files; it compares `AGENTS.md` to `.claude/CLAUDE.md`, neither of which this PR touches.

Verified behaviorally rather than by unit test, since the lane is prompt text executed by an agent:

- Reproduced the failure state on a machine with `~/.coderabbit` allowlisted for writes and `coderabbit.ai` blocked. The old probe reports writable, which is what made the lane choose the sandboxed path.
- The new probe returns `CR_SANDBOXED=no` in that same state, which is the value that routes STEP 2 through the bypass instead of the hang.
- Validated end-to-end on a live review (PR #1983) run from this branch's skill copy, on a machine in exactly the target state (`~/.coderabbit` writable, `coderabbit.ai` hosts network-blocked): STEP 1 returned `CR_SANDBOXED=no` from the network half, STEP 2 ran with the bypass and completed a real review with the pinned `baseCommit`, and all four lanes finished `ok`.
- Confirmed `-f` is required: without it `curl` exits 0 against the blocked endpoint in the reported 403 case.
- Confirmed the `ide.coderabbit.ai` check is meaningful and cheap: unblocked, `GET /` returns `200`; sandbox-blocked, the proxy returns `403`, so `curl -fsS` cleanly separates the two.
- Confirmed `ls -t` resolves to the companion the plugin system records as active (`installed_plugins.json` → `installPath` = `codex/1.0.6`), while a name-sorted `ls` selects the stale `1.0.2` copy still present in the cache.
- `workflow.mjs` parses in its execution shape (module `export` stripped, body wrapped in an async function) identically to `origin/main`. Note that a bare `node --check` fails on both this branch and `main`, because the script legitimately uses top-level `return`.
- The added prose contains no backticks or `${`, so it cannot break the surrounding template literals.

## Risk Assessment

No shipped code, binary, image, or recipe is affected. The change alters only the instructions given to the review lanes, and only in the direction of choosing the bypass path more often and resolving the companion the way the lane already does.

**Rollout notes:** None. Machines that already allowlist both the path and the hosts keep running fully sandboxed with no approval prompt, exactly as before.

## Checklist

- [x] Tests pass locally (`make test` with `-race`) — N/A, no Go changes; the gates listed under Testing were run instead
- [x] Linter passes (`make lint`) — the applicable non-Go gates were run; `lint-go` has no changed Go files
- [x] I added/updated tests for new functionality — N/A, the change is agent prompt and documentation text; verification is the behavioral reproduction above
- [x] I updated docs if user-facing behavior changed — `SKILL.md` updated in the same commit


.agents/skills/aicr-cross-review/SKILL.md
```diff
@@ -492,12 +507,13 @@ the consensus mechanics):
...
-  **probes** in step 1 whether `~/.coderabbit` is sandbox-writable, and runs the
-  `coderabbit` command (and only that command) with sandbox bypass exactly when the probe
-  says it is not.
+- **A sandboxed CodeRabbit run hangs instead of failing.** The sandbox can deny the CLI
+  two independent ways, and both stall at `connecting_to_review_service` until the
+  timebox kills it: `~/.coderabbit` outside the write allowlist (the CLI cannot create
+  its log or review store), or the `coderabbit.ai` hosts outside the allowed-hosts list.
+  The lane therefore **probes both** in step 1 — writability of `~/.coderabbit` *and*
+  reachability of `cli.coderabbit.ai` — and runs the `coderabbit` command (and only that
+  command) with sandbox bypass when either check fails.
```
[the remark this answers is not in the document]
yuanchen8911: Addressed in 199b7853e. The probe now tests both hosts: a third ANDed leg does `curl -fsS -o /dev/null -m 10 https://ide.coderabbit.ai/ 2>/dev/null`, and the SKILL.md wording names both `cli.coderabbit.ai` and `ide.coderabbit.ai` at the two spots that previously named only the config host.

Verified the check works before adopting it: unblocked, `GET https://ide.coderabbit.ai/` returns `200`; sandbox-blocked, the network proxy answers `403`, which `-f` turns into a probe failure. A plain HTTPS GET is sufficient for the `wss://` endpoint because the sandbox allowlist rule is host-scoped, not protocol-scoped. The result stays fail-safe in the cheap direction — if the site ever stops serving `200` at `/`, the probe degrades to a false `no`, which costs one approval prompt rather than a timebox.

.agents/skills/aicr-cross-review/SKILL.md
```diff
@@ -784,11 +811,13 @@ posting it) so no finding is left as a bare one-liner.
...
-    sandboxed CLI cannot create its log or review store and hangs at
-    `connecting_to_review_service` until the timebox kills it. Step 1 of the three-step
-    CodeRabbit protocol probes whether `~/.coderabbit` is sandbox-writable; when it is
-    not, bypass **step 2 only**, which is a lone `coderabbit review` command. When the
+  - **CodeRabbit review** — `~/.coderabbit` is outside the default write allowlist and
+    the `coderabbit.ai` hosts are outside the default network allowlist; under either
+    denial a sandboxed CLI hangs at `connecting_to_review_service` until the timebox
+    kills it. Step 1 of the three-step CodeRabbit protocol probes writability of
+    `~/.coderabbit` and reachability of both `cli.coderabbit.ai` and
+    `ide.coderabbit.ai` (the WebSocket host); when any check fails, bypass
```
njhensley, line 819: **🟡 Minor — Umbrella "Rules" sentence omits the network dimension this PR added**

This bullet correctly makes bypass fire "when any check fails" across writability *and* both-host reachability. But the umbrella sentence introducing the two exceptions — up at L806, `both conditional on the sandbox actually denying **the write**` — was left un-updated and now omits the network dimension it then describes here, contradicting its own child bullet. (This is the exact spot CodeRabbit flagged and marked "✅ Addressed in commit d7a8e6c", but that commit is not an ancestor of the squashed head and the L806 phrase is unchanged.) Doc self-consistency only — the operative probe in workflow.mjs is correct.

**Fix:** At L806, reword to cover both dimensions, e.g. "each conditional on the corresponding access probe failing" (or "…denying the write or network access").

.agents/skills/aicr-cross-review/SKILL.md
```diff
@@ -508,17 +527,25 @@ the consensus mechanics):
...
-  `~/.coderabbit/logs/`* is sandbox denial — a process killed mid-run still flushes a
-  partial log, so zero bytes means it never created one. A real cloud problem leaves a log
-  with 429/queue lines. Do not read this stall as an outage or as contention with another
-  session: an unsandboxed run succeeding while a sandboxed one hangs looks exactly like
-  contention and is not.
+  the call site. If you run this often, add **both** grants — `~/.coderabbit` (and
+  `~/.claude/plugins/data`, for the Codex companion's job log) to your local sandbox
+  `filesystem.allowWrite`, *and* the `coderabbit.ai` hosts to the network allowlist —
+  since that pair is what removes the per-round approval prompts. The skill must not
+  depend on either. Granting only the filesystem half is worse than granting neither:
```
njhensley, line 534: **🟡 Minor — "Filesystem half is worse than granting neither" is now false**

Present-tense "Granting only the filesystem half is worse than granting neither" no longer holds under the new both-dimension probe: filesystem-only (write passes, cli.coderabbit.ai curl fails → CRSBX=no) and neither (write fails → CRSBX=no) both bypass with one prompt — neither hangs. The "worse = 10-minute hang" outcome was true only of the OLD writability-only probe, which this sentence's own tail and L537 ("The probe now tests both for exactly this reason") admit is neutralized. workflow.mjs:463 already states this in correct past tense.

**Fix:** Reword to past tense / "buys nothing": e.g. "Granting only the filesystem half buys nothing — it satisfies the write probe while the network stays blocked, so the lane still bypasses. An earlier writability-only probe instead read that state as sandbox-clean and hung for a full timebox, which is why the probe now tests both."

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -428,17 +428,20 @@ STEP 1 — setup (SANDBOXED, normal Bash call):
    CRROOT=$(mktemp -d "\${TMPDIR:-/tmp}/cr-rabbit.XXXXXX")
    git -C "${repoPath}" worktree add --detach "$CRROOT/head" ${headSha}
    CRSBX=no
-   if { mkdir -p ~/.coderabbit && date +%s > ~/.coderabbit/.aicr-sandbox-probe; } 2>/dev/null; then CRSBX=yes; fi
+   if { mkdir -p ~/.coderabbit && date +%s > ~/.coderabbit/.aicr-sandbox-probe; } 2>/dev/null \
+      && curl -fsS -o /dev/null -m 10 https://cli.coderabbit.ai/public-configs.json 2>/dev/null \
+      && curl -fsS -o /dev/null -m 10 https://ide.coderabbit.ai/ 2>/dev/null; then CRSBX=yes; fi
```
njhensley, line 433: **🔵 Nitpick — curl -f conflates "host reachable" with "HTTP <400 at path"**

`curl -f` fails on any status ≥400, so it cannot distinguish a sandbox proxy 403 from an origin-returned 4xx/redirect at the probed path on a network-*allowed* machine. If CodeRabbit ever auth-gates `GET https://ide.coderabbit.ai/`, moves `public-configs.json`, or the WS host stops answering 200 on a plain GET, a fully-allowlisted machine flips to CRSBX=no and silently pays one approval prompt every run — no hang to signal it. This is the cheap/fail-safe direction and the 200-vs-403 assumption is already documented at L436, so it is accept-as-designed; flagging it as the single linchpin the fail-safe rests on.

**Fix:** Optional: a one-line breadcrumb near the probe — "if these endpoints stop answering 200 when reachable, re-point the probe" — so a future maintainer isn't left guessing why an allowlisted machine started prompting.

.agents/skills/aicr-cross-review/SKILL.md
```diff
@@ -508,17 +527,25 @@ the consensus mechanics):
   timebox per wrong guess. The step-1 probe settles it in milliseconds: machines with
```
njhensley, line 527: **🔵 Nitpick — "settles it in milliseconds" overstated for a blackhole sandbox**

The two added `curl -m 10` calls mean a filesystem-only machine whose sandbox *drops packets* (rather than returning a fast proxy 403) runs the probe up to ~20s, so "settles it in milliseconds" is overstated for that case. Still far under STEP 1's 120s default Bash timeout, so "Steps 1 and 3 are fast and need no explicit timeout" (workflow.mjs:446) stays true.

**Fix:** Optional: "…in milliseconds (up to ~20s if the sandbox blackholes rather than returning a fast 403)."

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -451,11 +454,13 @@ STEP 3 carries no "set -euo pipefail": every command in it is already \`|| true\
...
+until the timebox kills it.
+   FILESYSTEM denial — ~/.coderabbit is outside the write allowlist, so the CLI cannot create its log or review store. It writes NO log file at all.
+   NETWORK denial — the coderabbit.ai hosts are outside the allowed-hosts list. The CLI DOES write a log, naming the cause verbatim: 403 Forbidden on GET https://cli.coderabbit.ai/public-configs.json with "data":"Connection blocked by network allowlist", followed by a WebSocket retry against wss://ide.coderabbit.ai/ws every 30 s until the kill.
```
njhensley, line 461: **🔵 Nitpick — Network-denial log signature omits the ide-only-blocked sub-case**

The network-denial signature is written as "403 on cli.coderabbit.ai/public-configs.json … then wss://ide retries." But the per-host asymmetry the probe rationale itself calls out (an allowlist entry naming only the config host) produces the opposite log: the cli fetch succeeds, then ONLY the wss://ide retries appear with no cli 403. A reader matching strictly on the 403 line could misfile that sub-case. The probe still catches the config (the ide curl fails → bypass), so this is diagnosis-prose only.

**Fix:** Optional: note the two host blocks can appear independently — a cli 403, or (if only ide is blocked) the wss://ide retries alone.

.agents/skills/aicr-cross-review/scripts/workflow.mjs
```diff
@@ -428,17 +428,20 @@ STEP 1 — setup (SANDBOXED, normal Bash call):
    CRROOT=$(mktemp -d "\${TMPDIR:-/tmp}/cr-rabbit.XXXXXX")
    git -C "${repoPath}" worktree add --detach "$CRROOT/head" ${headSha}
    CRSBX=no
-   if { mkdir -p ~/.coderabbit && date +%s > ~/.coderabbit/.aicr-sandbox-probe; } 2>/dev/null; then CRSBX=yes; fi
+   if { mkdir -p ~/.coderabbit && date +%s > ~/.coderabbit/.aicr-sandbox-probe; } 2>/dev/null \
+      && curl -fsS -o /dev/null -m 10 https://cli.coderabbit.ai/public-configs.json 2>/dev/null \
```
njhensley, line 432: **🔵 Nitpick — -S flag is inert with stderr redirected to /dev/null**

Both probe curls are `-fsS … 2>/dev/null`. `-S` re-enables the error message that `-s` suppresses, but stderr is discarded, so `-S` never does anything.

**Fix:** Harmless — drop `-S`, or keep for symmetry.

njhensley (approved):
## Multi-persona review — Approve with comments ✅

**Method:** three independent persona reviewers (Correctness · Operability/Sandbox-portability · Prompt/Doc-consistency) → adversarial senior meta-reviewer re-derived every finding from the resolved code at `881a0ff5`.

### Overall

A tight, genuinely well-tested prompt/doc fix — no shipped code. The two load-bearing mechanics both check out:

- **The probe is correct and fail-safe.** The write probe and the two new `curl -fsS -m 10` host checks are `&&`-chained, so `CR_SANDBOXED=yes` requires all three to pass; any failure → `no` → STEP 2 bypasses. Walked all four machine configs (both-allowlisted / filesystem-only / neither / network-only) — each lands on the intended outcome. The curls run **inside** STEP 1's sandbox, which is what earns them the proxy `403` that makes them a valid probe; the bypass surface is not widened (STEP 2 remains the lone `coderabbit review`).
- **The two subtle traps are clean.** The `\`-continued probe lines sit inside the `CODERABBIT_RUN` template literal, so JS collapses them to one valid shell line (verified it parses under `bash -n` and `zsh -n`). The Codex-companion `ls -t … | head -1` + `node "$comp" status <job> --cwd … --json` block matches the lane's own D1/W1/W3 invocations verbatim.

Every surviving finding is doc self-consistency; there are **no blockers and no majors**. All six are inline. The only 🟡 that carries a bit of history is **F1** — it's the CodeRabbit "exception-wording" item that was marked fixed in `d7a8e6c`, but that commit isn't in the squashed head and the L806 wording is unchanged, so it's still live (worth the one-line touch-up). The other human "reviewer" the tooling flagged is the PR author self-commenting to answer CodeRabbit — not an independent review.

- Template-literal `\`-continuation renders to a single valid shell line; parses under bash & zsh.
- Companion `ls -t` resolution matches the lane's D1/W1/W3 usage verbatim; bare `status … --json` is the correct terminal-poll form.
- Fail-safe direction correct across all machine configs.
- Bypass surface unchanged — STEP 1 & 3 stay sandboxed; probe curls run inside the sandbox.
- `.claude/skills` is a symlink → `.agents/skills`, so there's no mirror copy to drift; `check-agents-sync` (CLAUDE.md ↔ AGENTS.md) is correctly untouched.
- zsh unmatched-glob on the companion glob is pre-existing lane behavior gated by the Phase-0 presence check — not introduced here.
- CodeRabbit's `ide.coderabbit.ai` WebSocket-host probe is already implemented at head (the third curl leg).

### Summary

| Tier | Count |
|------|-------|
| 🔴 Blocker | 0 |
| 🟠 Major | 0 |
| 🟡 Minor | 2 |
| 🔵 Nitpick | 4 |

Recommendation: **Approve with comments.** F1 and F2 are worth a one-line prose fix each; the rest are optional polish.

PR #2174, opened 2026-08-12
docs: add ADR-019 for k8s-aibom

mchmarny:
## Summary

Adds ADR-019, proposing k8s-aibom as an optional registry-only runtime AI inventory component after explicit upstream packaging, security, lifecycle, and qualification gates.

## Motivation / Context

Defines ownership, validation, supply-chain, RBAC, privacy, mirror, and evidence boundaries before implementation. Stock-recipe adoption and runtime-observation evidence remain separate future decisions.

Fixes: N/A
Related: https://github.com/GoogleCloudPlatform/k8s-aibom/issues/8

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Implementation Notes

The ADR approves registry-only adoption. It rejects bundle-time component disablement as a stock-selection contract because that does not change recipe health expectations.

## Testing

```bash
make check-docs-filenames check-docs-mdx check-docs-mdx-parse
node tools/mdx/check-docs-mdx-parse.mjs docs/design/019-k8s-aibom-runtime-inventory.md
lychee --config /dev/null --offline --no-progress docs/design/019-k8s-aibom-runtime-inventory.md
```

All focused documentation checks passed.

## Risk Assessment

**Rollout notes:** Documentation-only proposal; no runtime behavior changes.

## Checklist


docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,490 @@
...
+The qualified AICR values must preserve these defaults:
+
+- namespace discovery requires the explicit
+  `aibom.k8saibom.dev/enabled=true` label;
+- AICR does not label customer namespaces;
+- no GCS, webhook, GUAC, or other external sink is configured;
+- no cloud identity, credential Secret, token, certificate, or endpoint is
+  created;
+- generated BOMs remain in AIBOM status unless the operator explicitly
+  configures a sink;
```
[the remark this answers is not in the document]
mchmarny: Addressed in 03777e49. The ADR now fixes the default at 262144 bytes, defines inline storage at or below that threshold, and requires truncation with summary, hash, size, and an accurate reason when no configured sink succeeds.

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,490 @@
...
+
+### Operational safety
+
+- Kubernetes and external-sink operations are context-bounded.
+- Retries are bounded and limited to transient failures.
+- Invalid configuration and last-known-good behavior are visible.
+- Inline status and delivery queues cannot grow without bounds.
+- Readiness detects controller/configuration failure rather than only a live
+  process.
+- Resource use is measured at representative workload counts.
```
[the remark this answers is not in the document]
mchmarny: Addressed in 03777e49. The operational gate now requires numeric deadlines, cancellation, bounded retry attempts/elapsed time/backoff, fixed queue capacity and overflow behavior, plus explicit Ready/Synced/SinkFailed outcomes and freshness handling.

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,511 @@
...
+- Vendored-chart output is tested wherever the deployer supports it.
+- Scheduling values land on the controller Deployment.
+- `aicr mirror list` discovers the exact image and mirror/copy qualification
+  preserves its digest identity.
+- `make component-test COMPONENT=k8s-aibom` proves installation and the generic
+  health contract.
+- A separate, named Kind integration test opts in a fixture namespace, creates
+  a lightweight digest-pinned workload, and proves AIBOM generation,
+  reconciliation after a relevant workload change, hash consistency,
+  CycloneDX 1.6 validity, and owner-reference cleanup. The implementation PR
```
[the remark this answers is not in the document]
mchmarny: Agreed, and fixed in `46705e70`. I verified the field semantics against `GoogleCloudPlatform/k8s-aibom` at `e752beb1` — the exact commit this ADR pins — rather than accepting them on description, and upstream's own doc comment makes the point better than I could:

```go
// Distinct from InputHash: BOMHash captures the OUTPUT bytes
// (which include the BOM's generation timestamp), whereas
// InputHash captures the INPUT scrape result (which is time-
// independent). Two reconciles of an unchanged workload produce
// different BOMHash values but the same InputHash.
```

So "hash consistency" was not merely vague, it was actively misleading: the obvious reading — assert `bomHash` is stable across reconciles — yields a test that fails every single time.

The text now states both assertions that do hold and calls out the exclusion, so the implementation PR does not have to rediscover this from the CRD:

> Its hash assertions are that `status.bomDocument.sha256` matches the canonical BOM bytes, and that a reconcile of an unchanged workload preserves `status.inputHash`. It must not assert that `status.bomHash` is stable: that field covers the output bytes, which carry the BOM's generation timestamp, so it changes on every regeneration by design.

`status.bomDocument.sha256` is likewise confirmed upstream as "the hex-encoded sha256 of the canonical BOM JSON ... regardless of where it is stored", which is what makes it the right anchor for the inline-vs-external split this ADR describes.

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,516 @@
...
+## Adoption Gates
+
+Registry implementation begins only after one exact upstream release passes all
+of the following gates.
+
+### Release and supply chain
+
+- Tagged source, published chart, multi-architecture image, digest, SBOM, and
+  provenance form one coherent release.
+- Provenance and SBOM subjects bind the selected image digest.
```
njhensley, line 327: 🟡 **Minor** — The "Release and supply chain" gate is _"Provenance and SBOM subjects bind the selected image digest"_, backed by Decision 4(7) _"machine-verifiable build provenance bound to the same digest and source commit"_. Both assert only that the attestation's **subject** names the right digest/commit — neither requires verifying **who** produced the provenance (expected upstream OIDC issuer + repo) nor a minimum SLSA build level. A well-formed but self-/attacker-issued attestation whose subject correctly names a malicious digest satisfies the gate as literally written. Subject-binding defends against provenance↔image *mismatch*; it does not establish a *trusted builder*. Low blast radius today (upstream publishes none of this yet; qualification is a one-time human review of a well-known GCP repo), but this is the one gate that would become fail-open in the SLSA sense if later automated verbatim. Consider adding: _provenance is verified against the expected upstream builder identity and meets a stated minimum SLSA level; subject-digest binding alone is insufficient._ (Note this deliberately exceeds ADR-006, which defers signature verification — so state it as an ADR-019-specific requirement.)

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,516 @@
...
+release.
+
+### 2. Keep adoption registry-only
+
+The implementation adds the registry entry, values, health check, tests, and
+documentation without adding `k8s-aibom` to `recipes/overlays/base.yaml`, a leaf
+overlay, or a mixin used by stock recipes.
+
+Registry presence does not enable a component. A custom or external recipe must
+declare a `ComponentRef` explicitly. Once declared and hydrated, the component's
```
njhensley: 🟡 **Minor** — Decision 2 frames opt-in as per-recipe: _"A custom or external recipe must declare a `ComponentRef` explicitly."_ But AICR resolution is criteria-match across **all** overlays, so opt-in scope is actually governed by the declaring overlay's `criteria`. `recipes/overlays/monitoring-hpa.yaml` (`criteria: {intent: any}`) is a live example that injects `prometheus-adapter` into every matching recipe. An adopter following this ADR could author a broad-criteria custom overlay declaring `k8s-aibom` and unexpectedly attach the controller — **and its failing health check** — to unrelated recipes. The decision itself is sound (line 384 _"Stock recipe resolution and bytes remain unchanged"_ correctly covers AICR-shipped overlays); only the opt-in *granularity* description is imprecise. Suggest adding a half-sentence (here or in the adopter-doc deliverable) that opt-in scope follows the declaring overlay's `criteria` and adopters should scope narrowly, avoiding `intent: any` unless universal injection is intended.

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,516 @@
...
+- Invalid configuration and last-known-good behavior are visible.
+- Inline status remains bounded by the configured threshold.
+- Readiness detects controller/configuration failure rather than only a live
+  process.
+- Resource use is measured at representative workload counts.
+
+### AICR qualification
+
+- The component renders consistently through `helm`, `helmfile`, `argocd`,
+  `argocd-helm`, `flux`, and `localformat`.
```
njhensley: 🔵 **Nitpick** — The AICR-qualification gate lists render targets `helm, helmfile, argocd, argocd-helm, flux, localformat` as six peers. The user-selectable deployer set is only five: `allDeployerTypes` in `pkg/bundler/config/config.go` omits `localformat`, and `ParseDeployerType("localformat")` returns `ErrCodeInvalidRequest`. `localformat` is the shared internal writer the Helm/Helmfile/vendored-chart paths build on, not a `--deployer` value. This is soft because the ADR mirrors the repo's own godoc (`pkg/bundler/doc.go:27` uses the identical 6-item "Deployers" list) — so it follows internal precedent. Worth a half-sentence so an implementation-PR author doesn't write a gate/test invoking `--deployer localformat` and hit a hard parse error: note `localformat` is exercised at the writer / vendored-chart level, not via a `--deployer` flag.

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,516 @@
...
+
+An AICR recipe records desired platform configuration: selected components,
+chart or source pins, values, constraints, deployment order, and validation
+expectations. AICR evidence can bind a resolved recipe, snapshot, validation
+results, and component BOM at one point in time. It does not continuously
+observe workloads after deployment or prove that later live state still matches
+the validated recipe.
+
+AICR components are declarative metadata. A normal Helm component consists of a
+registry entry, minimal AICR values, a pinned chart version, a read-only health
```
njhensley: 🔵 **Nitpick** — _"A normal Helm component consists of a registry entry, minimal AICR values, a pinned chart version, a read-only health check, and explicit recipe references."_ A registry component entry contains no back-reference to recipes; recipes/overlays reference components via `componentRefs`. The reference lives on the recipe side, not in the component definition (Decision 2 states this correctly). Reword to _"…and is referenced explicitly by recipes"_ or drop that clause.

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,516 @@
...
+
+- workload and pod APIs are read-only;
+- AIBOM writes are limited to the controller-owned API group and required
+  status paths;
+- configuration writes are limited to required status behavior;
+- Event writes are justified and bounded;
+- Secret access is absent when sinks are disabled, or restricted to the
+  controller namespace and minimum verbs when enabled;
+- no privileged pod, host mount, host namespace, node-agent, workload exec, or
+  Secret-value discovery permission exists;
```
njhensley, line 263: 🔵 **Nitpick** — Decision 6 enumerates forbidden capabilities (_"no privileged pod, host mount, host namespace, node-agent, workload exec, or Secret-value discovery permission"_) — a denylist. A future broad grant not on the list (e.g. a ClusterRole `aggregationRule` pulling the SA into admin/edit/view, an `impersonate` verb, or cluster-wide ConfigMap/Secret read) wouldn't be named. This is largely mitigated already — L245-248 positively bound the *allowed* writes and L260 (_"unexpectedly broad … rendered resources block admission"_) is an allowlist catch-all — so it's a durability nit, not a hole. Consider explicitly adding aggregation labels and cluster-wide non-AIBOM reads to the forbidden set, per AICR's allowlist-gate convention.

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,516 @@
...
+- Default configuration has deterministic install, update, rollback, and
+  uninstall ownership.
+- CRD conversion, migration, and retention behavior is documented for the
+  selected API.
+
+### Security and privacy
+
+- Rendered RBAC and pod security satisfy Decision 5 and Decision 6.
+- No AIBOM is produced for a namespace without an explicit selector match;
+  underlying informer visibility remains documented as cluster-wide.
```
njhensley: 🔵 **Nitpick** — The ADR honestly discloses (L75-76, L331) that namespace opt-in gates *AIBOM production* but not the controller's cluster-wide informer cache; the gate only requires that fact be _"documented as cluster-wide."_ So pod specs from every namespace (including inline env values, args, image refs) are read and cached in controller memory even for namespaces never opted in. Data does not leave the cluster by default (no sink), so this is in-memory read-scope, not egress — but it undercuts a naive "namespace opt-in = the controller only sees opted-in namespaces" reading. Consider either requiring a namespace/label-scoped informer as a qualification *preference*, or stating explicitly that cluster-wide read into controller memory is an accepted residual, so adopters price it in.

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,516 @@
...
+This proposal was reviewed on 2026-08-12 against upstream commit
+[`e752beb15c8eb0179bba4f3066c7b989c84da33e`](https://github.com/GoogleCloudPlatform/k8s-aibom/commit/e752beb15c8eb0179bba4f3066c7b989c84da33e).
+At that commit:
+
+- the public API is `aibom.k8saibom.dev/v1alpha1` and explicitly experimental;
+- the repository has no source tag or GitHub release;
+- Google publishes no controller image or Helm repository for the project;
+- the chart accepts an image repository and tag, but not a first-class digest;
+- chart, app, and binary-reported versions are inconsistent;
+- the chart renders a namespace on the cluster-scoped default configuration and
```
njhensley: 🔵 **Nitpick** — _"the chart renders a namespace on the cluster-scoped default configuration and creates it as a `pre-install` hook"_ (L92-93) can be misread as "renders a `kind: Namespace`," which the chart does **not** do. At the pinned commit, `templates/default-cr-hook.yaml` sets `metadata.namespace: {{ .Release.Namespace }}` on the cluster-scoped `AIBOMControllerConfig` (with a `pre-install` hook) — a meaningless namespace on a `scope=Cluster` object, which is what Decision 6 L254 (_"the cluster-scoped default configuration has no namespace"_) targets. Reword to e.g. _"sets a `metadata.namespace` on the cluster-scoped default configuration and creates it via a `pre-install` hook."_

docs/design/019-k8s-aibom-runtime-inventory.md
```diff
@@ -0,0 +1,516 @@
...
+of the following gates.
+
+### Release and supply chain
+
+- Tagged source, published chart, multi-architecture image, digest, SBOM, and
+  provenance form one coherent release.
+- Provenance and SBOM subjects bind the selected image digest.
+- No floating tag, branch, `latest`, or locally built artifact appears in the
+  component definition.
+- The chart renders only expected images and supports digest selection.
```
njhensley: 🔵 **Nitpick** _(contingent)_ — Decision 4(5) / the gate speak of _"the selected image"_ (singular controller) and _"the chart renders only expected images and supports digest selection."_ If the qualified chart renders any sidecar (e.g. a kube-rbac-proxy, the common controller-runtime pattern), that image is a chart-default which ADR-006 Layer 3 leaves tag-only pending admission-time verification — so the single-image reproducibility framing could be over-read. Purely a clarity nit and contingent on a sidecar actually existing in the qualified release; if the chart is expected to be single-image, a one-line statement to that effect resolves it.

njhensley (dismissed):
## Multi-persona review — ADR-019 (k8s-aibom)

Reviewed with four independent persona lenses (Domain & Architecture, Supply-chain/Security, Technical Accuracy, Docs), each finding adjudicated by an adversarial senior meta-reviewer against the resolved repo and against upstream `GoogleCloudPlatform/k8s-aibom@e752beb1`.

**This is a strong, carefully-scoped ADR.** Every falsifiable claim I checked holds:

- **Upstream (at the pinned commit):** API group `aibom.k8saibom.dev/v1alpha1`, the `aibom.k8saibom.dev/enabled=true` label, status fields (`bomDocument.inline/.sha256/.truncated`, `inputHash`, `bomHash`), `inlineThresholdBytes=262144`, `AIBOMControllerConfig/default` + `Ready`/`Synced`/`SinkFailed` conditions, no source tag/release (0 tags, 0 releases confirmed), and the `pre-install`-hook namespace on the cluster-scoped config — **all accurate**.
- **AICR-side:** component shape, the `{helm, helmfile, argocd, argocd-helm, flux, localformat}` render set, `--set <component>:enabled=false` semantics, fail-closed health/validation, and ADR-006/007 consistency — all verified.
- **Default security posture is airtight** — no sink, no credential, inline-in-etcd, `truncated=true` never a silent full-BOM emission; nothing leaves the cluster by default.
- Docs gates (`check-docs-filenames`, `check-docs-mdx`, real `@mdx-js/mdx` parse) all pass; links resolve; header/status format matches sibling ADRs.

CodeRabbit's three prior findings are already addressed and resolved, so they are not re-raised here.

Nothing blocks. The inline comments below are wording-precision and gate-durability suggestions — all optional. The two 🟡 Minors (provenance builder-identity verification at L312; opt-in-granularity wording at L151) are the two worth weighing.

| Tier | Count |
|------|-------|
| 🔴 Blocker | 0 |
| 🟠 Major | 0 |
| 🟡 Minor | 2 |
| 🔵 Nitpick | 6 |

Recommendation: **Approve with comments.**

docs/design/019-k8s-aibom-runtime-inventory.md
mchmarny: Correct, and this is the one I'd have regretted. Verified ADR-006 does defer this — its non-goals list "Mandating signature verification on upstream artifacts" at that policy level — so stating it here as an ADR-019-specific requirement is the right place.

Fixed in `a3390836`. Decision 4(7) now requires provenance "attributable to the expected upstream builder identity", and the release gate adds:

> Provenance verifies against the expected upstream builder identity — OIDC issuer, source repository, and workflow ref — and meets a stated minimum SLSA build level. Subject-digest binding alone is insufficient: it detects a provenance/image mismatch, not a well-formed attacker-issued attestation whose subject correctly names a malicious digest. This requirement is specific to ADR-019 and deliberately exceeds ADR-006, which defers signature verification to admission time.

docs/design/019-k8s-aibom-runtime-inventory.md
mchmarny: Correct. Confirmed `recipes/overlays/monitoring-hpa.yaml` carries `criteria: {intent: any}` and injects `prometheus-adapter`, and it is the only overlay in tree with `intent: any`.

Fixed in `a3390836`, in Decision 2 rather than deferring to the adopter doc, since the imprecision is in the decision text:

> Opt-in scope follows the declaring overlay's `criteria`, not a single recipe. AICR resolution is criteria-match across every overlay, so a broad-criteria overlay attaches its components — and their health checks — to every matching recipe; `recipes/overlays/monitoring-hpa.yaml` (`criteria: intent: any`) is the in-tree example. An adopter declaring `k8s-aibom` must scope the declaring overlay narrowly and avoid `intent: any` unless universal injection is intended.

docs/design/019-k8s-aibom-runtime-inventory.md
mchmarny: Correct. `allDeployerTypes` (`pkg/bundler/config/config.go:59-65`) lists five, and `ParseDeployerType` returns `ErrCodeInvalidRequest` for anything outside it. Keeping the six-item list to match `pkg/bundler/doc.go:27`, with the caveat inline so an implementation-PR author does not write the failing gate:

> Only the first five are `--deployer` values; `localformat` is the shared writer those paths build on, so exercise it at the writer and vendored-chart level rather than by passing `--deployer localformat`, which fails to parse.

Fixed in `a3390836`.

docs/design/019-k8s-aibom-runtime-inventory.md
mchmarny: Correct — `recipes/registry.yaml` entries carry name, displayName, valueOverrideKeys, healthCheck, helm/kustomize, testTier, and nodeScheduling; nothing points back at a recipe. Fixed in `a3390836`, and made the direction explicit rather than just dropping the clause:

> Recipes reference it explicitly through `componentRefs`; the registry entry itself carries no back-reference to any recipe.

docs/design/019-k8s-aibom-runtime-inventory.md
mchmarny: Taking the concern, implementing it narrower than proposed.

Added in `a3390836`:

> no ClusterRole `aggregationRule` or aggregation label pulls the ServiceAccount into `admin`, `edit`, or `view`, and no `impersonate`, `escalate`, or `bind` verb and no wildcard (`*`) apiGroup, resource, or verb grant appears. The cluster-wide reads the controller does require stay bounded by the positive limits above, which are the governing allowlist;

I did not add "cluster-wide non-AIBOM reads" to the forbidden set. The rendered ClusterRole at `e752beb1` needs exactly that: cluster-wide get/list/watch on `namespaces`, `pods`, `jobs`, `daemonsets`/`deployments`/`replicasets`/`statefulsets`, and `serving.kserve.io/inferenceservices`. Forbidding them would contradict L245 ("workload and pod APIs are read-only") and make the gate unsatisfiable by any conforming release. Cluster-wide `ConfigMap`/`Secret` read — the case you actually named — is already covered by the existing Secret bullet plus the wildcard clause above.

docs/design/019-k8s-aibom-runtime-inventory.md
mchmarny: Took the second option. The ADR disclosed the fact but never said whether it was acceptable, which left the decision implicit. Now explicit, in `a3390836`:

> No AIBOM is produced for a namespace without an explicit selector match. The controller's informer cache stays cluster-wide regardless of opt-in, so pod and workload specs from every namespace — including inline environment values, args, and image references — are read into controller memory even for namespaces never opted in. No data leaves the cluster by default because no sink is configured, so this is in-memory read scope rather than egress. It is an accepted residual of this decision, not an oversight, and must be documented so adopters price it in. A namespace- or label-scoped informer is a qualification preference, not a gate.

Confirmed cluster-wide at the pinned commit: `cmd/manager/main.go:103` builds `ctrl.NewManager` with no `Cache.DefaultNamespaces` or `ByObject` restriction, so controller-runtime defaults to a cluster-wide cache.

docs/design/019-k8s-aibom-runtime-inventory.md
mchmarny: Correct, and the misreading is the likelier one. Confirmed at `e752beb1`: `charts/k8s-aibom/templates/default-cr-hook.yaml` renders `kind: AIBOMControllerConfig` with `metadata.namespace: {{ .Release.Namespace }}` and `helm.sh/hook: pre-install` — there is no `kind: Namespace` anywhere in the chart. Took your wording in `a3390836`:

> the chart sets a `metadata.namespace` on the cluster-scoped default configuration and creates it via a `pre-install` hook;

docs/design/019-k8s-aibom-runtime-inventory.md
mchmarny: Resolved the contingency, but not by asserting single-image.

At `e752beb1` the chart is single-container: `charts/k8s-aibom/templates/deployment.yaml` renders one `manager` container, no sidecar and no init container. But the gates apply to a future qualified release, not to the reviewed commit, so "the chart is expected to be single-image" is a claim this ADR cannot make. Made the requirement hold either way instead, in `a3390836`:

> The chart renders only expected images, and every rendered image — including any sidecar or init container the qualified release introduces — is selectable by digest through public values. A chart-default sub-image left tag-only would fall into ADR-006 Layer 3, outside this ADR's pinning guarantee. At the reviewed commit the chart renders a single `manager` container.

That closes the Layer 3 gap you identified without betting on the future chart's shape.

yuanchen8911 (approved):
/lgtm

glenmessenger:
Upstream update: v1.0.0 is released and covers the adoption blockers and release requirements in this ADR — signed multi-arch image, attested OCI chart, digest-pinned install.yaml. Details on the k8s-aibom https://github.com/GoogleCloudPlatform/k8s-aibom/issues/8#issuecomment-5321497332

Issue #2178, opened 2026-08-13
revisit nvsentinel assumeDriverInstalled scope once NVSentinel#1583 is decided


yuanchen8911:
> **UPDATE — the trigger has fired.** NVSentinel maintainers answered NVIDIA/NVSentinel#1583 with **documentation only**: `assumeDriverInstalled` is the recommended mechanism for host-installed drivers, and no detection fallback will be added ("folks using host installed drivers … may also not use GFD or any number of other GPU operator operands; I'd like to keep the labeler simpler"). That is the third outcome row below: **the configuration is permanent**, and the over-broad scope must now be re-decided explicitly — accept it on the record, or take the profile-scoped form, which requires #2181 (the sole remaining gate, since the upstream escape hatches are off the table).

## Summary

Revisit the `labeler.assumeDriverInstalled` configuration in the AICR recipes once [NVIDIA/NVSentinel#1583](https://github.com/NVIDIA/NVSentinel/issues/1583) is **decided** — not only if it ships a fix.

**Why the trigger is a decision, not a release.** [NVIDIA/NVSentinel#1583](https://github.com/NVIDIA/NVSentinel/issues/1583) has been reframed upstream from a bug report into a design question: NVSentinel design 018 already considered this deployment shape, assigned it to a documented Manual Labeling Procedure, and explicitly rejected GPU Feature Discovery label detection. Automatic detection would reverse that decision, so the maintainers may reasonably answer in any of four ways, and only one of them ends in deletion:

| Upstream outcome | What this issue becomes |
|---|---|
| Automatic detection ships (opt-in or default) | Delete everything — the checklist below applies as written |
| Node-scoped `assumeDriverInstalled` ships | Narrow the configuration to a selector instead of deleting it |
| Documentation only — the flag is the sanctioned answer | **The configuration is permanent.** Re-examine the over-broad scope below, since the justification for it assumed a known end date |

The third outcome is the one that needs attention rather than closure. AICR applies the flag per overlay rather than per `gpuStack` profile value, so AKS `operator-managed` and GKE `driver-installer` receive it too, even though a driver pod genuinely manages those nodes. That over-broad scope was accepted because breaking a documented user-facing override on three platforms is the worse trade. If this configuration turns out to be permanent, the cost stops being time-boxed and the trade should be re-decided explicitly — either accept it on the record, or take the profile-scoped form and its component-presence lock.

[...]

Issue #2176, opened 2026-08-12
nvsentinel metadata-collector pods rejected on AKS azure-managed (recipes don't propagate operator.runtimeClass)


yuanchen8911:
## Summary

On AKS with the default `gpuStack=azure-managed` profile, NVSentinel's `metadata-collector` DaemonSet cannot create pods because it requests RuntimeClass `nvidia`, while the profile configures the primary NVIDIA RuntimeClass as `nvidia-container-runtime`.

```text
Warning  FailedCreate  daemonset/metadata-collector
Error creating: pods "metadata-collector-" is forbidden:
pod rejected: RuntimeClass "nvidia" not found
```

#2181 now defines the resolution: configure NVSentinel's RuntimeClass from the selected AKS `gpuStack` profile, while keeping NVSentinel mandatory. This requires recipe-data changes only—no `requirePresence` field, profile-schema change, or new ownership model.

## Root cause

The GPU Operator's ClusterPolicy controller names the primary RuntimeClass after `operator.runtimeClass`. The AKS profiles select different names:

| AKS profile | `operator.runtimeClass` | RuntimeClass created | NVSentinel value required |
|---|---|---|---|
| `azure-managed` | `nvidia-container-runtime` | `nvidia-container-runtime` | `nvidia-container-runtime` |
| `operator-managed` | `nvidia` | `nvidia` | `nvidia` |

The NVSentinel chart exposes the required value directly:

```yaml
metadata-collector:
  runtimeClassName: nvidia
```

An earlier revision described the RuntimeClass as hardcoded after considering only the chart's top-level values. That was incorrect: the metadata-collector subchart exposes the value above, so no post-render patch or upstream chart change is needed.

The current recipe does not propagate the profile-selected RuntimeClass to NVSentinel. Under `azure-managed`, admission therefore rejects every metadata-collector pod before a pod object is created. The only cluster-side signal is the DaemonSet's `FailedCreate` event.

## Scope and impact

The broken configuration is AKS `azure-managed`, which is the AKS default. It prevents GPU metadata collection, including NVLink topology, NVSwitch inventory, and the GPU-to-NIC topology matrix.

AKS `operator-managed`, both GKE-COS profile values, OKE, and EKS use RuntimeClass `nvidia` and already match the chart default. Both AKS profile values must nevertheless assign the NVSentinel path explicitly so it remains profile-owned and satisfies union totality.

[...]

PR #2179, opened 2026-08-13
fix(recipes): label GPU nodes where the driver is provider-installed

### LINKED ISSUE #2175: recipes omit labeler.assumeDriverInstalled on provider-installed-driver platforms (AKS/GKE-COS/OKE)
## Summary

On platforms where the GPU driver is installed into the node image, NVSentinel's labeler cannot infer driver presence from a driver pod. It therefore does not apply `nvsentinel.dgxc.nvidia.com/driver.installed=true`, and the NVSentinel DaemonSets selected by that label remain at zero desired pods.

#2181 now defines the resolution: configure `labeler.assumeDriverInstalled` from the selected `gpuStack` profile, while keeping NVSentinel mandatory. This requires recipe-data changes only—no `requirePresence` field, profile-schema change, or new ownership model.

## Observed behavior

On AKS with the default `gpuStack=azure-managed` profile:

```text
metadata-collector             0/0
syslog-health-monitor-regular  0/0
syslog-health-monitor-kata     0/0
gpu-health-monitor-dcgm-4.x    2/2
```

AKS supplies the driver and toolkit in the node image, so the GPU Operator driver pod that the labeler normally observes does not exist. The same failure mechanism applies to GKE-COS `gke-default` and OKE. It does not apply when a persistent driver pod exists.

The metadata collector also needs the RuntimeClass fix tracked by #2176. Both fixes are required for it to become ready on AKS `azure-managed`.

## Required configuration

| Platform and profile | `labeler.assumeDriverInstalled` | Reason |
|---|---:|---|
| AKS `azure-managed` | `true` | The driver is installed in the node image; no driver pod exists. |
| AKS `operator-managed` | `false` | The GPU Operator driver pod must remain the health signal. |
| GKE-COS `gke-default` | `true` | The driver is installed at node-pool creation; no persistent driver pod exists. |
| GKE-COS `driver-installer` | `false` | Google's driver-installer DaemonSet supplies a driver pod that NVSentinel detects. |
| OKE | `true` at overlay level | The host-installed driver has no driver pod, and OKE has no `gpuStack` profile. |
| EKS | Unchanged (`false` default) | Existing driver-pod detection is correct. |

The chart exposes this configuration directly:

```yaml
labeler:
  assumeDriverInstalled: false
```

An earlier revision proposed applying the node label from an AICR manifest after considering only the chart's top-level values. That was incorrect: the labeler subchart exposes the supported flag above, which also handles newly added nodes without rerunning a labeling manifest.

When enabled, the labeler applies the driver-installed label to GPU nodes without waiting for a driver pod. NVIDIA/NVSentinel#1583 confirms this is the supported mechanism; no automatic detection fallback is planned.

## Implementation

- Add an NVSentinel `componentRef` to both values of the AKS and GKE-COS `gpuStack` profiles.
- Assign `labeler.assumeDriverInstalled` explicitly in every value, including `false` in the driver-pod-managed variants. This preserves union totality and prevents an unsafe hybrid configuration.
- Keep the OKE value at overlay level because OKE has no `gpuStack` profile.
- Retain ADR-015's existing synthetic `enabled` ownership and presence lock. NVSentinel is required for the profiled deployments, so disabling or filtering it must remain rejected.
- Do not introduce `requirePresence`, registry-level skippability, or component-specific engine logic.

PR #2179 contains the existing implementation branch. Reopen it after #2177 and rework it to move the AKS and GKE-COS values from family-wide overlays into the profile values. Replace its disableability expectation with tests for mandatory presence and the exact profile matrix above.

## Acceptance criteria

- AKS `azure-managed` and GKE-COS `gke-default` resolve `assumeDriverInstalled=true`.
- AKS `operator-managed` and GKE-COS `driver-installer` resolve `assumeDriverInstalled=false`.
- OKE retains `assumeDriverInstalled=true`; EKS remains unchanged.
- Generated bundles need no normal `--set nv-sentinel:labeler.assumeDriverInstalled=...` remedy.
- Divergent bundle-time, dynamic, and supported install-time values are rejected by the existing profile lock.
- Disabling or filtering NVSentinel from profiled AKS/GKE-COS recipes is rejected.
- Catalog integration tests pin the complete matrix.

## Closure

#2181 is the coordinating implementation issue. Close this defect once the #2181 recipe changes and acceptance tests land.

Until then, the verified bundle-time remedy for an affected profile is:

```bash
aicr bundle -r recipe.yaml \
  --set nv-sentinel:labeler.assumeDriverInstalled=true ...
```

Do not use that override for AKS `operator-managed`, GKE-COS `driver-installer`, or EKS. Manually applying the node label is not durable because the labeler can remove it during reconciliation.

## Environment

- AICR at `main`
- AKS 1.35.6, Ubuntu 24.04, 2× H100
- NVSentinel v1.9.0, GPU Operator v26.3.3, driver 580.159.04 (Azure preinstalled)

yuanchen8911:
## Summary

Set `labeler.assumeDriverInstalled=true` for NVSentinel on the three platform families whose default cluster shape has no NVIDIA driver pod — AKS, GKE-COS, and OKE — so the NVSentinel labeler applies `nvsentinel.dgxc.nvidia.com/driver.installed` and its DaemonSets actually schedule.

Upstream documents this deployment shape as one its labeler cannot auto-detect, and tells admins to label the nodes by hand. `labeler.assumeDriverInstalled` is the chart-level automation of that documented remedy — so setting it from the recipe is using the upstream-provided mechanism, not working around one.

## Motivation / Context

NVSentinel's labeler infers driver presence by watching for a driver pod. Where the driver ships in the node image or is preloaded at node-pool creation, no such pod ever exists, so the label is never applied and every DaemonSet selecting on it reports 0 desired pods — silently, with no error and no event:

```
metadata-collector             0/0
syslog-health-monitor-regular  0/0
syslog-health-monitor-kata     0/0
gpu-health-monitor-dcgm-4.x    2/2   (unaffected — selects on the DCGM label)
```

Because the failure is partial and `gpu-health-monitor` stays healthy, it is easy to miss. On AKS this affects every deployment out of the box, since `gpuStack=azure-managed` is the default.

The knob is a documented upstream value whose stated purpose is exactly this case. It is declarative, needs no extra RBAC or DaemonSet, and applies to new nodes automatically — a node-labeling manifest would have to re-run on every scale-out.

Fixes: #2175
Related: NVIDIA/NVSentinel#1583 (the root defect this works around — this change is removable once it ships), #2177, #2178

#2177 fixes an unrelated `runtimeClassName` mismatch on the same component. The two bugs are additive: both fixes are needed before `metadata-collector` reaches Ready.

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

- [ ] Other: ____________

## Implementation Notes

A single shared values file, `recipes/components/nvsentinel/values-preinstalled-driver.yaml`, is referenced by the `aks`, `gke-cos`, and `oke-ol` overlays. The chart path was verified against the pulled chart rather than assumed: the umbrella `Chart.yaml` declares the dependency under its own name `labeler` with no alias, so the top-level `labeler.assumeDriverInstalled` key is correct, and `helm template` confirms it renders the `--assume-driver-installed` arg.

**Why this is the upstream-provided remedy.** NVSentinel design 018 (`docs/designs/018-syslog-monitor-preinstalled-driver-support.md`, "Syslog Health Monitor Support for Pre-installed Drivers") specifies two-tier driver-pod detection — tier 1 the GPU Operator driver DaemonSet, tier 2 GKE's `nvidia-driver-installer` — and is explicit that it does not cover this case. It lists "custom machine images with pre-baked drivers (no installer pods)" and "GPU Operator with all components disabled" under **Scenarios requiring manual labeling**, names the gap in its own Cons, and includes a **Manual Labeling Procedure** telling admins to `kubectl label nodes ... driver.installed=true`.

`labeler.assumeDriverInstalled` is the chart-level automation of exactly that documented remedy. To be precise: design 018 predates the flag and does not mention it — the point is that the deployment shape and its remedy are upstream-documented, not that the flag is blessed by that document.

**Scope: overlay-level, deliberately.** The value is set on each overlay's `componentRefs`, so it applies to the whole family rather than only the affected `gpuStack` value. The tighter fit would be to gate it per profile value, since the two non-default values do have a driver pod (AKS `operator-managed`: the GPU Operator's own; GKE-COS `driver-installer`: Google's standalone `nvidia-driver-installer` DaemonSet, which the labeler already detects).

That form was prototyped and rejected on evidence. Naming a component in a profile fragment also makes its **presence** profile-owned via the synthetic `enabled` path, and the profile lock then rejects any output where the component is absent or disabled. The prototype produced exact per-value scoping (`true` for `gke-default`, `false` for `driver-installer`) but turned the documented `--set <component>:enabled=false` capability into a hard failure:

```
[INVALID_REQUEST] profile-owned component "nvsentinel" is absent or disabled in the output
```

Trading a working, documented user-facing override on three platforms for a tighter scope is the wrong deal on its own terms, independently of how long this configuration lives, so the overlay form was kept. A regression test pins that NVSentinel stays disableable.

The residual cost is smaller than it first looks. The two non-default, opt-in profile values also skip driver-pod detection, which can label a GPU node while its driver is still loading — but:

- The `driver.installed` label gates only NVSentinel's own monitors (`metadata-collector`, `syslog-health-monitor`). It never gates GPU scheduling, device-plugin advertisement, or workload placement.
- Design 018 deliberately keys the label on driver *installation* rather than driver *functionality*, because it wants the syslog monitor to run "even if drivers are broken, to capture error logs" — so labeling a node whose driver is still loading is closer to that monitor's intent than harmful.
- This is the supported configuration for these platforms, not a stopgap: NVSentinel design 018 lists this deployment shape under "Scenarios requiring manual labeling" and gives a Manual Labeling Procedure, and `labeler.assumeDriverInstalled` is the chart-level automation of that documented remedy. It may become redundant if NVIDIA/NVSentinel#1583 is answered with automatic detection — one of four possible outcomes there — which #2178 tracks.

## Testing

```bash
make qualify
```

Unit tests added in `pkg/recipe/nvsentinel_driver_label_test.go` pin which resolved recipes carry the value (AKS, GKE-COS, OKE) and that EKS does not, plus that NVSentinel stays disableable under both profiles. Both include explicit non-vacuity guards.

Verified end to end on a live AKS cluster (AKS 1.35.6, 2x H100, NVSentinel v1.9.0, Azure-preinstalled driver 580.159.04). The pre-existing hand-patched flag was reverted first so the check could not pass vacuously:

| Step | GPU node label | `metadata-collector` | `syslog-health-monitor-regular` |
|---|---|---|---|
| Flag removed | absent | 0/0 | 0/0 |
| Recipe-generated config applied | `true` | 2/2 | 2/2 |

The applied manifest was rendered from the values file this PR generates (`aicr bundle` output fed to `helm template`), and `kubectl diff` confirmed the only change against the live object was the added `--assume-driver-installed` arg.

`make bom-docs` was run as required. The BOM gate covers the version column and the component set; this change touches neither, and the only delta was pre-existing upstream `busybox` digest drift already present on `main` — rendered-image drift, which is not PR-gated and is owned by the weekly BOM-refresh workflow. `docs/user/container-images.md` is therefore intentionally not part of this PR.

## Risk Assessment

**Rollout notes:** Values-only change to three overlays; no API, CLI, or schema change. Reverting is deleting the values file and its three `valuesFile` references. #2178 tracks revisiting this once NVIDIA/NVSentinel#1583 is decided: if that answer is automatic detection, the configuration is deleted; if it is documentation or a node-scoped flag, it stays and its scope is re-examined.

## Checklist


yuanchen8911:
Parking this pending two decisions it depends on, per the discussion on the linked issues.

The branch is complete and green (`make qualify` exit 0 on the committed tree): the shared values file, the three overlay references, tests, and docs. The final amend (`0ca522233`) also folds in the OKE operator-managed override tuple update — `gpuOperatorManagedOverrideSet` now disables `labeler.assumeDriverInstalled`, so the documented flip to GPU-Operator-managed mode on OKE restores pod-based driver detection.

Why parked rather than merged:

- The flag is applied per overlay, so the pod-managed profile values (AKS `operator-managed`, GKE-COS `driver-installer`) receive it too, where it can label a node before its driver loads — `metadata-collector` then fails NVML init and CrashLoopBackOffs, failing the nvsentinel health check and deployment-phase validation. Scoping it per `gpuStack` value is the correct shape, and that requires either #2181 (own a value without locking component presence) or the upstream direction in NVIDIA/NVSentinel#1583 — both open design decisions.
- This PR's `TestNVSentinelStaysDisableableUnderProfiles` also conflicts with #2177 on a merged tree (that PR makes nvsentinel profile-owned on AKS); whichever lands second must rebase and reconcile that test.

Interim state for users: the remedy is documented and verified — `--set nv-sentinel:labeler.assumeDriverInstalled=true` at bundle generation — and a separate PR adds bundle-time detection, a non-vacuous health check, and the documentation for both nvsentinel gaps (Refs #2175, #2176).

Reopen when #2181 or NVIDIA/NVSentinel#1583 is decided; #2178 tracks the revisit.

PR #2187, opened 2026-08-13
fix(validator): report deadline-killed validator as failed, not "pod not found"

### LINKED ISSUE #2186: validate: DeadlineExceeded validator reports 'pod not found' status=other, hiding real assert failures
## Summary

When a deployment-phase validator Job exceeds its `activeDeadlineSeconds`, the Job controller kills and deletes its pod. The CLI then reports `pod not found for Job aicr-expected-resources-<hash>: [NOT_FOUND] pod for job not found` with validator status `other` — an infrastructure-looking error with the pod's logs gone. If what actually happened is that a health-check assert correctly refused to pass, the detection is buried: the operator sees a flaky-looking `NOT_FOUND`, not "these DaemonSets are not rolled out".

The phase still fails closed (`status=other` → `aicr validate` exits 8), so this is a diagnostics defect, not a fail-open.

## Reproduction (observed live, twice, on aicr-uat-31034819439)

1. NVSentinel deployed with `metadata-collector` and `syslog-health-monitor-regular` at 0 desired pods (the #2175 condition).
2. Recipe carries DaemonSet-readiness asserts (`numberReady == desiredNumberScheduled && desiredNumberScheduled > 0`) in the nvsentinel health check — the #2183 hardening.
3. `aicr validate --phase deployment`:

```
events:  Warning DeadlineExceeded  job/aicr-expected-resources-863edfd9  Job was active longer than specified deadline
         Normal  SuccessfulDelete  job/aicr-expected-resources-863edfd9  Deleted pod: ...
result:  expected-resources  status=other
         message: pod not found for Job aicr-expected-resources-863edfd9: [NOT_FOUND] pod for job not found
```

Healthy cluster: same validator completes in seconds and passes. Broken: 8m20s (the deadline), both runs.

## Two contributing pieces

1. **Chainsaw asserts wait.** An assert that will never become true holds the validator until the Job deadline rather than failing with the assert diff. #2183 is bounding its own asserts so they fail fast inside the deadline — that fixes the nvsentinel case at the check level.
2. **The CLI loses the failure's identity (this issue).** `DeadlineExceeded` → pod deleted → the pod lookup finds nothing → `NOT_FOUND` → `other`. The Job object still records the `DeadlineExceeded` condition at that point, so the CLI could report "validator exceeded its deadline while asserts were unsatisfied" instead of "pod not found". Two candidate improvements:
   - On pod-not-found, consult the Job's conditions and surface `DeadlineExceeded` as the message, status `failed` rather than `other`.
   - Consider capturing the pod's last logs before the Job controller deletes it (the validator had been printing assert progress), or set `ttlSecondsAfterFinished`/pod-failure policy such that the failed pod survives long enough to read.

Per-check assert timeouts (#2183) reduce how often this path is hit but do not remove it — any future check whose assert wait exceeds the deadline lands here again.

## Scope

`status=other` handling and the pod-lookup path in the deployment validator plumbing (`pkg/validator/v1`, the `pod not found for Job` message). Pre-existing on `main`; surfaced by #2183's live verification, not caused by it.

## References

- #2183 — the check hardening whose live verification surfaced this; bounds its own asserts as the check-level mitigation
- #2175 — the underlying silent condition used to reproduce

yuanchen8911:
## Summary

A validator Job killed on `activeDeadlineSeconds` is now reported as a failed check naming its deadline, instead of an inconclusive `pod not found for Job <name>` with status `other` — in every case where the orchestrator observes the Job reach a terminal state, which is the reproduction in #2186. One narrow exception is documented rather than fixed here (see Known limitation).

## Motivation / Context

When a validator Job exceeds its `activeDeadlineSeconds`, the Job controller kills **and deletes** the pod. `ExtractResult`'s pod lookup then legitimately finds nothing, and the result was reported as exit code `-1` — CTRF `StatusOther` — with the message `pod not found for Job <name>: [NOT_FOUND] pod for job not found`.

That mislabels a real, reproducible check failure as an infrastructure hiccup. The operator sees a flaky-looking `NOT_FOUND` rather than "this validator ran and its assertions never became true", and the pod's logs are already gone. Observed live twice on `aicr-uat-31034819439`, where DaemonSet-readiness asserts correctly refused to pass and the run took the full 8m20s deadline both times.

The phase already failed closed (`other` is blocking), so this is a diagnostics defect, not a fail-open.

Fixes: #2186
Related: #2183, #2175, #2189

## Type of Change

- [x] Documentation update
- [ ] Build/CI/tooling

## Component(s) Affected

## Implementation Notes

The Job object outlives its pod and still records why it failed, so the pod-not-found path in `ExtractResult` now consults the Job's terminal `Failed` condition before deciding what to report:

- **`Failed/DeadlineExceeded`** — a definitive verdict: the validator **Job** did not succeed within its own deadline (this says nothing about whether the check body ever ran — see Known limitation). Reported as exit code `1` (`StatusFailed`), naming the deadline and noting that no pod remains, so its logs are unavailable. The message states that observed fact rather than naming a cause — `ErrCodeNotFound` also covers a pod never created (quota/webhook rejection) and one removed externally, for which asserting controller deletion would name a false root cause. The reason is matched against `batchv1.JobReasonDeadlineExceeded` from the pinned API.
- **Any other `Failed` reason** — the reason and message are surfaced for diagnosis, but the exit code stays `-1` (`StatusOther`). With the pod gone, the container's own verdict is genuinely unknown.
- **No terminal `Failed` condition, or the Job itself can no longer be read** — the original `pod not found for Job ...` wording is preserved verbatim. Diagnosis is never blocked on the extra Job `Get` succeeding; a failure there is logged at DEBUG and falls through.

Job condition text is capped with `boundTerminationMsg` at `defaults.ValidatorMaxTerminationMsgBytes`, matching the pod-status branch. The result flows into ConfigMaps and rendered reports, so an unbounded condition message would have been the one way around that cap.

Only a **confirmed** absence licenses that diagnosis. `GetPodForJob` distinguishes "every candidate is gone" (`ErrCodeNotFound`) from "the List itself failed" (`ErrCodeInternal`); an apiserver hiccup is the latter and establishes nothing about the pod, so it stays inconclusive rather than being promoted to a verdict claiming the controller deleted a pod that was never confirmed gone.

The diagnostic Job read is bounded by `defaults.K8sCleanupTimeout` (30s). `aicr validate` opts out of a facade-level deadline (`WithValidationTimeout(0)`), so the inherited context may carry none at all, and an unbounded `Get` would hang result extraction on a stalled apiserver instead of emitting the diagnosis it exists to produce.

`docs/user/validation.md` needed four corrections in its exit-code block, two of them to statements this change made wrong and two that were already wrong and became self-contradictory next to the new text.

- **Deadline mapping.** The exit-code table and the gating-subtleties note both attributed an exceeded `activeDeadlineSeconds` to status `other`; both now describe it as `failed` — qualified rather than absolute. The diagnosis lives in `ExtractResult`, which runs only once `WaitForCompletion` sees the Job go terminal; if the client-side wait expires first the run routes to `HandleTimeout` and still reports `other`. That budget is `activeDeadlineSeconds` plus a 30s buffer exactly equal to the pod's termination grace, so the race has zero margin and a validator using its full grace can land on the old path. The note describes that exception; routing `HandleTimeout` through the same helper is tracked in #2189.
- **Sibling reference page.** `docs/user/cli-reference.md` carried the same pre-change mapping in its status and exit-code tables; all three rows now match the shipped behaviour and link to the fuller treatment in `validation.md`.
- **What `other` means.** It was described as a crash or an OOM. The non-deadline branch added here also yields `other` — a Job that failed for a reason such as `BackoffLimitExceeded` with its pod already gone proves neither — so `other` is now defined as an indeterminate outcome, with crash and OOM as examples.
- **Exit 5 scope (pre-existing).** The table and note claimed a per-validator wait deadline retains `ErrCodeTimeout` and exits `5`. It does not: `runPhase` hands that error to `HandleTimeout` and records a check result, so it exits `8`. Exit `5` is now stated as a rule rather than a list of sources — any `ErrCodeTimeout` reaching the top-level CLI. The sources are not enumerable in a doc line: the snapshot-agent wait; a `cm://` ConfigMap GET (`classifyConfigMapGetError`); a local recipe read on a stalled NFS/FUSE mount, which `LoadRecipe`'s 30s `RecipeOperationTimeout` and the file reader classify as `ErrCodeTimeout`; and evidence signing under `--emit-attestation --push`, which runs *after* the phase checks. A local snapshot read is deliberately absent from that list — `LoadFromFileWithKubeconfig` adds no deadline of its own, so a stalled read there hangs rather than classifying.

The last item was pre-existing, but the new client-wait note sits one line above it, so leaving it would have shipped two adjacent sentences contradicting each other.

Both branches were already blocking (`ctrf.IsFailingStatus` covers `failed` and `other`), so this changes diagnostic fidelity only — never whether a phase fails or what the process exit code is.

The reported deadline is the one Kubernetes actually enforced. It applies the same fallback the orchestrator uses when a catalog entry pins no timeout (so it never renders a bare `0s`), and the same whole-second truncation `BuildJobPlan` applies when assigning `activeDeadlineSeconds` — a `1500ms` entry yields a `1s` Job deadline, and the message says `1s`, not `1.5s`.

Deliberately **out of scope** (the issue's second candidate improvement): capturing the pod's last logs before the controller deletes it, or adjusting `ttlSecondsAfterFinished` / pod-failure policy so the failed pod survives long enough to read. That changes Job lifecycle and log-capture plumbing; this PR only restores the failure's identity. Also out of scope: the same pod-not-found branch in `HandleTimeout`, which already threads a meaningful `waitCause` (#1966) and is not the path the reproduction takes.

**Known limitation.** A pod that never leaves `Pending` — for example a `required` dependency-affinity term whose dependency isn't ready, which `pkg/validator/v1/affinity.go` documents as parking the orchestrator pod until `activeDeadlineSeconds` fires — reaches the same `Failed/DeadlineExceeded` condition and is likewise reported `failed`, even though no check body ran. Distinguishing never-started from ran-and-hung is not possible from the Job status alone once the pod is deleted: `Ready`, `Active`, `Failed`, and `StartTime` all read identically in the two cases, so the discriminator would have to be captured before deletion or reconstructed from Events. That is the same pre-deletion state capture listed as out of scope above, and is tracked as #2189 — along with the `HandleTimeout` bypass described under the doc change — rather than widened into this PR. The rendered message stays honest either way — it says the validator did not complete within its deadline, not that a check ran and failed.

## Testing

```bash
KUBEBUILDER_ASSETS="$(setup-envtest use -p path)" go test -race -count=1 ./pkg/validator/...
golangci-lint run -c .golangci.yaml ./pkg/validator/...
make qualify
```

Five test functions added in `pkg/validator/job/result_test.go`, one of them table-driven over four deadline variants (eight cases total). The envtest ones stamp the terminal condition the Job controller would set (envtest runs no controller-manager); the List-failure case uses a fake clientset reactor:

- `TestExtractResultPodNotFoundDeadlineExceeded` — table-driven over four cases: the message names the deadline, reason and controller message with the bare pod-not-found wording gone; an entry with no pinned timeout reports the default rather than `0s`; an empty controller message falls back to the reason; and a sub-second catalog timeout reports the truncated deadline the Job enforced. Each subtest gets its own namespace and Job, since the catalog timeout is baked into the deployed Job.
- `TestExtractResultPodNotFoundOtherJobFailure` — `BackoffLimitExceeded` surfaces its reason while staying `StatusOther`.
- `TestExtractResultPodNotFoundJobAlreadyReaped` — Job gone: original wording and `StatusOther` preserved.
- `TestExtractResultPodListFailsStaysInconclusive` — a failed pod List with a terminal `DeadlineExceeded` Job stays `-1` / `StatusOther`. Verified non-vacuous: with the `ErrCodeNotFound` gate removed it fails with `ExitCode = 1`, `CTRFStatus = "failed"`, and a message claiming the controller deleted the pod.
- `TestExtractResultPodNotFoundBoundsConditionText` — an oversized condition message stays near the cap with a truncation suffix. Verified non-vacuous: unbounded it produces 8350 bytes against the 4096 cap.

The pre-existing `TestExtractResultPodNotFound` was tightened to assert the original wording survives when the Job carries no `Failed` condition, so the fallback branch is pinned rather than merely non-empty.

Coverage: `pkg/validator/job: 88.9% → 90.0% (+1.1%)`. Both new functions are at 100%.

`make qualify` was run on the Go-equivalent revision of this change and reached the vulnerability scan with every preceding stage green — tests with `-race`, lint, tuning-check, e2e and license check all passed. The `scan` stage failed locally on 60 Critical / 141 High findings that came entirely from `tools/mdx/node_modules/`, a gitignored npm tree the MDX docs gate installs into the working directory during the run; a pristine `main` checkout scans clean with the same command and config. Revisions after that point are documentation-only plus one test-only guard, verified with scoped checks — `golangci-lint`, race-enabled `pkg/validator/job` tests, and the MDX doc checks — rather than a full re-run. This change touches no `go.mod`, `vendor/` or workflow pin, and CI scans a fresh checkout, which is the authoritative gate.

**Verified end-to-end on a live AKS cluster**, bracketing the behaviour change:

| Step | Result |
|---|---|
| Break — remove `labeler --assume-driver-installed` | labels stripped, consumers drop to 0 desired in 15s |
| Validate the broken cluster | `expected-resources` reports **`failed`** with `validator Job did not complete within its 8m0s deadline … (Job condition Failed/DeadlineExceeded)`, where `main` reports `pod not found` / `other` |
| Restore the argument | labels return, both DaemonSets 2/2 in 15s |
| Re-validate | **passed 4/4 in 3m26s** (runID `20260813-125049`) |

The three unaffected validators stayed correctly isolated and passing throughout, so the deadline reclassification does not leak into neighbouring checks. The cluster was left as found.

Not covered by local tests: the 30s bound on the diagnostic `Get`. client-go's fake `Action` does not carry the request context, so there is nothing observable to assert short of a stalled-apiserver harness.

## Risk Assessment

**Rollout notes:** No migration. One behavior change worth naming: a deadline-killed validator's CTRF status moves from `other` to `failed` on the path where the orchestrator sees the Job go terminal. It stays `other` when the client-side wait expires first (the zero-margin full-grace race described above, tracked in #2189), so consumers should not treat the new mapping as absolute. Both are blocking, the phase outcome and process exit code are unchanged, and any consumer keying on `other` for this case was reading it as inconclusive when it was in fact a verdict.

## Checklist

- [x] I updated docs if user-facing behavior changed — four corrections to the `docs/user/validation.md` exit-code block (see Implementation Notes)


pkg/validator/job/result.go
```diff
@@ -60,9 +73,12 @@ func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
...
 	if err != nil {
-		// Pod was never created or was deleted externally
-		result.ExitCode = -1
-		result.TerminationMsg = fmt.Sprintf("pod not found for Job %s: %v", d.jobName, err)
+		// Pod was never created, was deleted externally, or — the common
+		// case — the Job controller killed and deleted it on
+		// activeDeadlineSeconds expiry. The Job object still records why,
+		// so classify from its Failed condition instead of reporting a bare
+		// "pod not found" (see issue #2186).
+		result.ExitCode, result.TerminationMsg = d.missingPodOutcome(ctx, err)
```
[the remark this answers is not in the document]
yuanchen8911: Agreed — fixed.

`GetPodForJob` returns `ErrCodeInternal` when the List fails (`pkg/k8s/pod/find.go:55`) and `ErrCodeNotFound` only on a confirmed absence (`find.go:84`), and the diagnosis now gates on the latter via `stderrors.As` plus a code check — the same discrimination `deployer.go:342` already uses. Anything else falls through to the original wording at exit `-1` / `StatusOther`.

Regression test added: `TestExtractResultPodListFailsStaysInconclusive` drives a fake clientset whose pod List returns an internal error while the Job carries a terminal `Failed/DeadlineExceeded` condition. Confirmed non-vacuous — with the gate removed it fails with `ExitCode = 1`, `CTRFStatus = "failed"`, and a message claiming the controller deleted the pod.

pkg/validator/job/result.go
```diff
@@ -108,6 +124,70 @@ func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
...
+	}
+
+	if cond.Reason == jobReasonDeadlineExceeded {
+		return validatorExitFailed, fmt.Sprintf(
+			"validator did not complete within its %s deadline; the Job controller deleted the pod, so its logs are unavailable (Job condition Failed/%s: %s)",
+			effectiveTimeout(d.config.Entry.Timeout), cond.Reason, detail)
+	}
+
+	return -1, fmt.Sprintf("Job %s failed (%s): %s; pod not found: %v",
+		d.jobName, cond.Reason, detail, findErr)
```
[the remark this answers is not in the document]
yuanchen8911: Agreed — fixed.

Both Job-condition-derived messages now pass through `boundTerminationMsg(..., defaults.ValidatorMaxTerminationMsgBytes)`. The constant's own rationale is exactly this case: the result flows into ConfigMaps and rendered reports, so it is bounded at the source regardless of what upstream caps. The pod-status branch already did this, so this path was the one way around it.

Test added: `TestExtractResultPodNotFoundBoundsConditionText` stamps a condition message of `2 * ValidatorMaxTerminationMsgBytes` and asserts the result stays near the cap with a truncation suffix. Without the bound it produces 8350 bytes against a 4096 cap.

pkg/validator/job/result.go
```diff
@@ -108,6 +124,70 @@ func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
...
+// jobFailedCondition returns the validator Job's terminal Failed condition, if
+// the Job can still be read and carries one. A Get failure (including the
+// NotFound left by an already-reaped Job) is reported at DEBUG and yields
+// ok=false so the caller falls back to its original message — diagnosis must
+// never be blocked on a second API call succeeding.
+func (d *Deployer) jobFailedCondition(ctx context.Context) (batchv1.JobCondition, bool) {
+	job, err := d.config.Clientset.BatchV1().Jobs(d.config.Namespace).Get(ctx, d.jobName, metav1.GetOptions{})
+	if err != nil {
+		slog.Debug("failed to read Job while diagnosing missing pod", "job", d.jobName, "error", err)
+		return batchv1.JobCondition{}, false
```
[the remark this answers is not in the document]
yuanchen8911: Agreed — fixed.

The read is now wrapped in `context.WithTimeout(ctx, defaults.K8sCleanupTimeout)` (30s), the same short allowance the orchestrator already grants the sibling post-mortem capture around `HandleTimeout` in `validator.go`.

Worth recording why this mattered specifically here: `pkg/cli/validate.go:290` passes `WithValidationTimeout(0)`, documented at `pkg/client/v1/aicr.go:1597` as imposing no facade cap, so the inherited context can carry no deadline at all. The new sequence — pod lookup returns not-found promptly, then the Job Get stalls — was a hang window that did not exist before this change.

No unit test for the bound: client-go's fake `Action` does not carry the request context, so there is nothing observable to assert short of a stalled-apiserver harness.

docs/user/validation.md
```diff
@@ -855,11 +855,11 @@ error codes (see [`pkg/errors/exitcode.go`](https://github.com/NVIDIA/aicr/blob/
...
 
 > **Important:** two subtleties to be aware of when gating a pipeline on exit code:
 >
-> 1. Both `failed` and `other` are blocking. A phase whose status is `other` (check crashed, pod OOM, `activeDeadlineSeconds` exceeded) drives the same non-zero exit as `failed` — an inconclusive check fails closed rather than passing silently. `--fail-on-error=false` suppresses **both** result-driven exits (the run reports the outcomes in the CTRF report and exits 0); it does not distinguish `failed` from `other`.
+> 1. Both `failed` and `other` are blocking. A phase whose status is `other` (check crashed, pod OOM) drives the same non-zero exit as `failed` — an inconclusive check fails closed rather than passing silently. `--fail-on-error=false` suppresses **both** result-driven exits (the run reports the outcomes in the CTRF report and exits 0); it does not distinguish `failed` from `other`. A validator Job killed on `activeDeadlineSeconds` reports `failed`, not `other`: the Job's terminal `Failed/DeadlineExceeded` condition is a verdict, so the result names the deadline instead of the deleted pod.
```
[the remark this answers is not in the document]
yuanchen8911: Checked this against the code and I don't think it holds — `other` is the correct classification for an OOM kill here.

`ExitCodeToCTRFStatus` in `pkg/validator/ctrf/builder.go` maps only the exact value `1` to `failed`:

```go
case 0: return StatusPassed
case 1: return StatusFailed
case 2: return StatusSkipped
default: return StatusOther
```

An OOM kill is SIGKILL, so the container exits `137`, which falls through to `default` and yields `StatusOther`. "Non-zero" and "`failed`" are not equivalent on this path — only `1` is `failed`.

`TestExtractResultOOMKilled` in `pkg/validator/job/result_test.go` already pins exactly this: it stamps `ExitCode: 137` and asserts `CTRFStatus() == ctrf.StatusOther`. The OOM branch at `result.go:90-91` only rewrites `TerminationMsg` to `"Container OOMKilled"`; it never alters the exit code.

Worth noting the "pod OOM" wording under `other` is pre-existing text, not part of this change — the edit here only removed `activeDeadlineSeconds` from that list, since a deadline-killed Job now reports `failed`. Making the suggested change would put the doc in conflict with a passing test.

Leaving the line as-is.

yuanchen8911:
@coderabbitai review

docs/user/cli-reference.md
```diff
@@ -1297,15 +1297,15 @@ Results are output in CTRF (Common Test Report Format) — an industry-standard
 | `passed` | Check or constraint passed |
 | `failed` | Check or constraint failed |
 | `skipped` | Check could not be evaluated (missing data, no-cluster mode) |
-| `other` | Unexpected outcome (crash, OOM, timeout) |
+| `other` | Indeterminate outcome — the check produced no usable verdict (crash, OOM, or a validator Job that failed for a non-deadline reason with no inspectable pod) |
 
 **Exit Codes:**
 | Code | Description |
```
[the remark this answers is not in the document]
yuanchen8911: Leaving this as-is — it's pre-existing formatting rather than something this change introduces, and no gate in this repo enforces it.

The `**Exit Codes:**` label and the table that follows it are both untouched by this PR. The edits here are to three table *rows* — the `other` status description, and the exit `5` and exit `8` entries — not to the surrounding layout. The missing blank line predates the change and would remain if these rows were reverted.

There is also no markdownlint in the project: no `.markdownlint*` configuration at the repo root, and no reference to it in the `Makefile` or any workflow under `.github/workflows/`. MD058 is not part of the docs gate, which is the MDX check plus lychee for links.

Happy to see it fixed as part of a docs-formatting pass across the file, but folding a pre-existing cosmetic nit into a behavior fix widens the diff without making the fix more correct.

pkg/validator/job/result_test.go
```diff
@@ -280,6 +285,261 @@ func TestExtractResultPodNotFound(t *testing.T) {
...
+			Conditions: []batchv1.JobCondition{{
+				Type:    batchv1.JobFailed,
+				Status:  corev1.ConditionTrue,
+				Reason:  batchv1.JobReasonDeadlineExceeded,
+				Message: "Job was active longer than specified deadline",
+			}},
+		},
+	}
+	//nolint:staticcheck // SA1019: fake.NewSimpleClientset is sufficient for tests
+	client := fake.NewSimpleClientset(deadlineJob)
```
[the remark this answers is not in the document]
yuanchen8911: Keeping the directive as written — the suggested split would diverge from the established pattern in this very package.

The exact form used here appears four times in the sibling test file, at `pkg/validator/job/deployer_test.go` lines 641, 663, 679 and 697:

```go
//nolint:staticcheck // SA1019: fake.NewSimpleClientset is sufficient for tests
```

That is also golangci-lint's canonical directive shape — `//nolint:<linter> // <explanation>` — where the trailing comment *is* the explanation `nolintlint` looks for, not an unrelated remark that happens to share the line. Moving it to a preceding comment would leave a bare `//nolint:staticcheck` whose justification is no longer attached to the directive.

`.golangci.yaml` pins no `nolintlint` settings, so nothing in the configured lint set asks for the separated form, and `golangci-lint run -c .golangci.yaml ./pkg/validator/...` is clean as written.

If the separated style is preferred, it is worth applying to all five sites together rather than making this one line the outlier.

pkg/validator/job/result.go
```diff
@@ -108,6 +118,112 @@ func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
...
+	// precisely because the result flows into ConfigMaps and rendered reports,
+	// and the pod-status branch above already bounds its own message the same
+	// way — an unbounded Job condition would be the one way around it.
+	if cond.Reason == batchv1.JobReasonDeadlineExceeded {
+		// States the observed fact — no pod remains — rather than naming a
+		// cause. ErrCodeNotFound covers a pod deleted by the Job controller on
+		// deadline expiry (the dominant case), one never created at all
+		// (quota or webhook rejection), and one removed externally; asserting
+		// controller deletion would name a false root cause for the other two.
+		return validatorExitFailed, boundTerminationMsg(fmt.Sprintf(
```
njhensley: **🟡 Minor — DeadlineExceeded→failed also labels a never-ran check as a product `failed` verdict**

Confirmed, but an acknowledged design tradeoff — not a blocker. `activeDeadlineSeconds` expiry stamps `Failed/DeadlineExceeded` regardless of whether the container body ever ran. A pod that stayed `Pending` for the whole deadline (the `required` dependency-affinity case documented in `pkg/validator/v1/affinity.go:53`) reaches the identical condition and is now reported exit 1 → `StatusFailed`; semantically a check whose body never executed has an *unknown* verdict, which is what `StatusOther` exists to express, so that subcase is arguably `other`.

Why it's fine to merge anyway: (a) both statuses are blocking (`ctrf.IsFailingStatus`), so gating is identical; (b) the rendered message is deliberately scoped to the Job ("did not complete within its deadline… no pod remains"), making no false claim about a check body; (c) once the pod is deleted the two cases are genuinely indistinguishable from Job status. The one residual blast radius is triage fidelity — a CTRF consumer that treats `failed` (product) differently from `other` (infra/inconclusive), e.g. the UAT product-vs-infra dashboard classification, will now bucket a "dependency never came ready" event as a product failure. Already captured as a Known limitation and tracked in #2189; flagging only so the tradeoff is visible in review.

pkg/validator/job/result.go
```diff
@@ -224,6 +340,17 @@ func effectiveTimeout(configured time.Duration) time.Duration {
...
 }
 
+// enforcedDeadline is the deadline Kubernetes actually applied, as opposed to
+// the duration the catalog asked for. BuildJobPlan converts the catalog timeout
+// with int64(timeout.Seconds()) before assigning it to ActiveDeadlineSeconds
+// (pkg/validator/v1/job_plan.go), so any sub-second component is truncated
+// away: a 1500ms entry yields a 1s Job deadline. Reporting the raw duration
+// would name a deadline that never existed, and the whole point of this message
+// is to describe the Kubernetes state that produced the failure.
+func enforcedDeadline(configured time.Duration) time.Duration {
```
njhensley: **🟡 Minor — enforcedDeadline re-derives the deadline instead of reading the already-fetched Job's spec.activeDeadlineSeconds**

Latent drift / duplication. `jobFailedCondition` already fetches the live Job (`result.go:214`) and then discards everything except the condition — yet `job.Spec.ActiveDeadlineSeconds` is the literal value Kubernetes enforced, and the message's whole stated purpose is to describe that state. Instead the deadline is reconstructed by re-running the same `int64(seconds)` truncation `BuildJobPlan` performs at `pkg/validator/v1/job_plan.go:273`, with no test coupling the two copies. If that conversion ever gains a floor/round or the default-timeout fallback diverges, `enforcedDeadline` silently names a deadline that was never applied.

Suggestion: have `jobFailedCondition` return the `*batchv1.Job` (or the condition plus `*job.Spec.ActiveDeadlineSeconds`), format the deadline from that field (nil-guarded), and delete `enforcedDeadline`. Non-blocking — the two are currently identical.

pkg/validator/job/result.go
```diff
@@ -108,6 +118,112 @@ func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
...
+
+	// Only a CONFIRMED absence licenses a Job-condition diagnosis.
+	// GetPodForJob distinguishes "every candidate is gone" (ErrCodeNotFound)
+	// from "the List itself failed" (ErrCodeInternal); an apiserver hiccup is
+	// the latter, and it establishes nothing about the pod. Promoting it to a
+	// definitive verdict on the strength of a Job condition alone would flatten
+	// an ambiguous lookup into a claim the controller deleted the pod, which is
+	// exactly the failure mode the repo's fail-closed lookup rule forbids. An
+	// unestablished absence stays inconclusive.
+	var sErr *errors.StructuredError
```
njhensley: **🔵 Nitpick — Use stderrors.Is for the ErrCodeNotFound gate to match the repo idiom and the sibling in this file**

Idiom consistency. The gate uses `if !stderrors.As(findErr, &sErr) || sErr.Code != errors.ErrCodeNotFound`. CLAUDE.md's code-based-matching rule prefers `stderrors.Is(err, errors.New(errors.ErrCodeNotFound, ""))`, and `*StructuredError.Is` matches on `Code`, so `!stderrors.Is(findErr, errors.New(errors.ErrCodeNotFound, ""))` has exactly this gate's semantics (false for both "not a StructuredError" and "wrong code"). `isDeadlineCause` in this same file (`result.go:377`) already uses the `.Is` form. Purely stylistic; behavior is equivalent.

pkg/validator/job/result_test.go
```diff
@@ -280,6 +285,261 @@ func TestExtractResultPodNotFound(t *testing.T) {
...
 		t.Error("TerminationMsg should contain pod not found message")
 	}
+	if !strings.Contains(result.TerminationMsg, "pod not found for Job") {
+		t.Errorf("TerminationMsg = %q, want the original pod-not-found wording when the Job carries no Failed condition", result.TerminationMsg)
+	}
+}
+
+// setJobFailedCondition stamps a terminal Failed condition on the deployed Job,
+// standing in for the Job controller (envtest runs no controller-manager).
+func setJobFailedCondition(t *testing.T, ns, jobName, reason, message string) {
```
njhensley: **🔵 Nitpick — No negative test for a JobFailed condition with Status != True**

Test gap. The `c.Status == corev1.ConditionTrue` guard at `result.go:220` is never exercised — `setJobFailedCondition` always stamps `ConditionTrue`, so dropping the status check would go uncaught. This is the "presence check that passes on an ambiguous condition" class the repo anti-patterns table warns about (a `JobFailed/False` shouldn't be treated as a failure). One subtest stamping `JobFailed` with `Status=False` and asserting the original `pod not found` fallback (exit -1 / `other`) would pin it. Uncommon state in practice, so optional.

pkg/validator/job/result_test.go
```diff
@@ -280,6 +285,261 @@ func TestExtractResultPodNotFound(t *testing.T) {
...
+
+// TestExtractResultPodNotFoundDeadlineExceeded covers issue #2186: a validator
+// Job killed on activeDeadlineSeconds has its pod deleted by the Job
+// controller, so the pod lookup finds nothing. Every case must be reported as a
+// failed check naming the deadline, not an inconclusive "pod not found".
+//
+// Each subtest creates its own namespace and Job rather than sharing a fixture:
+// the catalog timeout is a per-case input baked into the deployed Job, and the
+// condition is stamped on that specific Job's status.
+func TestExtractResultPodNotFoundDeadlineExceeded(t *testing.T) {
```
njhensley: **🔵 Nitpick — No case for a sub-second timeout producing a degenerate "0s" enforced deadline**

Test gap (low value). The smallest table case is `90500ms → 1m30s`. A catalog `entry.Timeout` in (0, 1s) is *not* defaulted (`job_plan.go:229` only defaults `== 0`), so `enforcedDeadline` computes `int64(0.5)*time.Second == 0` and the message reads "within 0s deadline." This is consistent-by-design (`BuildJobPlan` also sets `activeDeadlineSeconds=0`, which in K8s means *no* deadline, so `DeadlineExceeded` couldn't fire from a real sub-second entry) and catalog timeouts are authored in minutes — effectively unreachable. Noting for completeness; a `500ms` row or a floor on the rendered deadline would remove the degenerate string if you want belt-and-suspenders.

pkg/validator/job/result.go
```diff
@@ -108,6 +118,112 @@ func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
...
+	if !stderrors.As(findErr, &sErr) || sErr.Code != errors.ErrCodeNotFound {
+		return -1, notFound
+	}
+
+	cond, ok := d.jobFailedCondition(ctx)
+	if !ok {
+		return -1, notFound
+	}
+
+	detail := strings.TrimSpace(cond.Message)
```
njhensley: **🔵 Nitpick — Whitespace-only cond.Message (load-bearing TrimSpace) is untested**

Test gap (low value). The `detail` fallback uses `strings.TrimSpace(cond.Message)`, but the "empty controller message" subtest passes `""`, which would fall back to the reason even without the `TrimSpace`. A `"   "` (whitespace-only) message would prove the trim is load-bearing. Trivial.

docs/user/validation.md
```diff
@@ -854,13 +854,13 @@ error codes (see [`pkg/errors/exitcode.go`](https://github.com/NVIDIA/aicr/blob/
...
+| `5` | A structured `ErrCodeTimeout` reached the top-level CLI — for example while loading the recipe or snapshot, waiting on the snapshot-agent Job, or signing evidence under `--emit-attestation --push` (which runs after the checks). A per-validator wait deadline does **not** reach this code: `runPhase` converts that error into a check result, so it surfaces as exit `8` (see note 2) |
+| `8` | One or more phases reported status `failed` (a check that returned a failure verdict — exit code `1` — or, in most cases, a validator Job killed on `activeDeadlineSeconds` — see the note below) **or** `other` (an indeterminate outcome — the check produced no usable verdict; a crash or OOM are the common examples, as is a validator Job that failed for a non-deadline reason with no inspectable pod), including per-check internal timeouts (e.g., DynamoGraphDeployment not ready within `InferenceWorkloadReadyTimeout`). Suppressed by `--fail-on-error=false` |
 
 > **Important:** two subtleties to be aware of when gating a pipeline on exit code:
 >
```
njhensley, line 861: **🔵 Nitpick — Doc's "buffer equals termination grace" (zero-margin race) isn't test-enforced**

Doc fragility. The note's "zero-margin" exception rests on `ValidatorWaitBuffer` (30s) *equaling* `ValidatorTerminationGracePeriod` (30s). But `pkg/defaults/timeouts_test.go:265` only asserts grace **≤** buffer, not equality — a future buffer bump would silently make the wording stale (the exception would become rarer/impossible, not more common, so it's conservative rather than a correctness risk). If you want the doc to stay honest automatically, an equality assertion (or wording like "at least the grace period") would couple them. Optional.

njhensley (commented):
## Multi-persona review — `fix(validator): report a deadline-killed validator as failed, not "pod not found"`

**Method:** three parallel persona reviewers (Correctness · Kubernetes-Domain/Architecture · Test-coverage+Docs), then adversarial adjudication against the resolved code at `6a94a4ed`. Every crux fact was independently reproduced.

### Overall assessment

A careful, well-scoped diagnostics fix with an unusually thorough writeup and test suite. A validator Job killed on `activeDeadlineSeconds` has its pod deleted by the Job controller, so the pod lookup legitimately finds nothing; the old code buried that as an inconclusive `pod not found` / `other`. It now consults the Job's terminal `Failed` condition and reports `Failed/DeadlineExceeded` as `failed` naming the deadline, while keeping every ambiguous case fail-closed. **No blockers or majors.** The load-bearing invariants hold up:

- **Fail-closed gate is real** — diagnosis runs only on a *confirmed* absence (`ErrCodeNotFound`); an `ErrCodeInternal` List failure stays inconclusive. Proven non-vacuously by `TestExtractResultPodListFailsStaysInconclusive`.
- **Routing is correct** — the deadline case reaches `ExtractResult` (not `HandleTimeout`), since `WaitForCompletion` returns nil on a terminal Failed Job and `runPhase` calls `ExtractResult` only on `waitErr==nil`.
- **`enforcedDeadline` names the real k8s deadline** — `BuildJobPlan` applies the identical `ValidatorDefaultTimeout` fallback + whole-second truncation.
- **ctrf mapping** — exit 1→`failed`, -1/137→`other`; both blocking. The doc's `other` redefinition and the exit-5-vs-8 correction are accurate, and the `#cicd-integration` anchor resolves.
- **The diagnostic Job Get is bounded** (`K8sCleanupTimeout`, 30s) and both interpolated condition messages route through `boundTerminationMsg`.

**Recommendation: Approve with comments.** The one actionable code nit is the `enforcedDeadline` re-derivation (read the already-fetched `job.Spec.ActiveDeadlineSeconds` instead); the `DeadlineExceeded→failed` semantics for a never-ran check is an acknowledged tradeoff already tracked in #2189.

> **Prior activity:** CodeRabbit's substantive points (bound the diagnostic Job Get, gate on `ErrCodeNotFound`, cap the condition text) are **already addressed in this PR** and covered by tests; its OOM-should-be-`failed`, markdownlint MD058, and split-`//nolint` nits were **correctly rebutted** — verified here (`exit 137 → other` is pinned by `TestExtractResultOOMKilled`; no markdownlint gate exists; the inline `//nolint` form matches four sibling sites).

### Confirmed non-issues (examined, refuted or already-addressed)

- **OOM/exit-137 should be `failed`** → refuted; ctrf maps 137→`other`.
- **markdownlint MD058 blank line** → refuted; no markdownlint gate in the repo.
- **Split the `//nolint:staticcheck` directive** → refuted; matches sibling sites and canonical nolintlint shape.
- **Bound the Job Get / gate on `ErrCodeNotFound` / cap condition text** → already fixed in this PR.

### Summary

| 0 | 0 | 2 | 5 | **Approve with comments** |

pkg/validator/job/result.go
yuanchen8911: Addressed in 7cc69b6a.

`jobFailedCondition` now returns the `*batchv1.Job` alongside the condition, and `enforcedDeadline` reads `job.Spec.ActiveDeadlineSeconds` — the value Kubernetes actually enforced. The `int64(seconds)` re-derivation survives only as a nil-guarded fallback for when the Job could not be read or the field is unset, so the two copies can no longer disagree about a deadline that was applied.

pkg/validator/job/result.go
yuanchen8911: Addressed in 7cc69b6a. Converted to `!stderrors.Is(findErr, errors.New(errors.ErrCodeNotFound, ""))`, matching `isDeadlineCause` in the same file and the code-based-matching rule.

pkg/validator/job/result_test.go
yuanchen8911: Addressed in 7cc69b6a.

Added `TestExtractResultPodNotFoundJobFailedConditionFalse`, which stamps `JobFailed` with `Status=False` and asserts the original pod-not-found fallback (exit -1 / `other`).

Verified non-vacuous: with the `c.Status == corev1.ConditionTrue` guard removed the test fails with `ExitCode = 1, want -1` and `CTRFStatus = "failed", want "other"`.

pkg/validator/job/result.go
yuanchen8911: Addressed in 7cc69b6a.

Added a `"   "` row to the `TestExtractResultPodNotFoundDeadlineExceeded` table. Verified non-vacuous: with `strings.TrimSpace` dropped the case fails, rendering `Failed/DeadlineExceeded:    ` instead of falling back to the reason.

pkg/validator/job/result.go
yuanchen8911: Agreed, and left as-is deliberately. The `Pending`-for-the-whole-deadline subcase is genuinely indistinguishable from Job status once the pod is deleted, both statuses are blocking, and the rendered message is scoped to the Job rather than the check body. Captured as a Known limitation and tracked in #2189.

pkg/validator/job/result_test.go
yuanchen8911: Not adding a case — the state is unreachable rather than merely rare.

`BuildJobPlan` sets `activeDeadlineSeconds = int64(timeout.Seconds())`, so a sub-second entry yields `0`. In Kubernetes `activeDeadlineSeconds: 0` means no deadline is enforced at all, so the Job controller can never stamp `Failed/DeadlineExceeded` for such an entry and `enforcedDeadline` is never reached with it. A `500ms` row would assert a rendering for a condition that cannot occur.

The whole-second truncation itself stays covered by the existing `90500ms → 1m30s` row.

docs/user/validation.md
yuanchen8911: Leaving this one out of this PR. The change it asks for lands in `pkg/defaults/timeouts_test.go`, which this diff does not touch, and by your own reading the drift direction is conservative — a buffer bump makes the exception rarer, not the wording wrong. Better as a follow-up than as scope added to a validator diagnostics fix.

njhensley (dismissed):
## Re-review — ✅ Approve (`7cc69b6a`)

Re-reviewed the delta since my prior review (`6a94a4ed`). Every one of the seven prior findings is resolved or soundly dispositioned; the two 🟡 minors are gone, no new issues were introduced by the fix commit, and `golangci-lint`/`go vet`/`go build` plus the fake-clientset test are green locally (envtest-backed subtests run in CI; the author verified each non-vacuously).

### Prior-feedback status

|---|---------------|-------------|
| 🟡 F2 | `enforcedDeadline` re-derived the deadline instead of reading the fetched Job | ✔️ **Addressed** — `jobFailedCondition` now returns the `*batchv1.Job`; `enforcedDeadline` reads `job.Spec.ActiveDeadlineSeconds` (the value K8s enforced), with the `int64(seconds)` re-derivation kept only as a nil-guarded fallback (`result.go:357-361`). The two copies can no longer disagree. |
| 🔵 F3 | `stderrors.As`+field compare instead of the repo `.Is` idiom | ✔️ **Addressed** — now `!stderrors.Is(findErr, errors.New(errors.ErrCodeNotFound, ""))` (`result.go:161`), matching `isDeadlineCause` in the same file. |
| 🔵 F4 | No negative test for a `JobFailed` condition with `Status != True` | ✔️ **Addressed** — `TestExtractResultPodNotFoundJobFailedConditionFalse` stamps `Status=False` and asserts the pod-not-found fallback; verified non-vacuous. |
| 🔵 F6 | Whitespace-only `cond.Message` (load-bearing `TrimSpace`) untested | ✔️ **Addressed** — a `"   "` row added to the deadline table; verified non-vacuous (drops to `Failed/DeadlineExceeded:    ` without the trim). |
| 🟡 F1 | `DeadlineExceeded→failed` also labels a never-ran (Pending-whole-deadline) check as `failed` | ⊘ **By design, acknowledged** — the never-ran subcase is genuinely indistinguishable from Job status once the pod is deleted, both statuses are blocking, and the message is scoped to the Job. Documented as a Known limitation and tracked in **#2189**. My finding already flagged this as acceptable. |
| 🔵 F5 | No sub-second-timeout "0s deadline" case | ⊘ **No longer applicable / correctly declined** — a sub-second entry yields `activeDeadlineSeconds=0`, which in K8s means *no* deadline, so `Failed/DeadlineExceeded` can never fire and `enforcedDeadline` is never reached with it. The state is unreachable, not merely rare; the whole-second truncation stays covered by the `90500ms → 1m30s` row. |
| 🔵 F7 | Doc "buffer equals grace" not test-enforced | ✖️ **Deferred (reasonable)** — the coupling assertion lands in `pkg/defaults/timeouts_test.go`, outside this diff, and the drift direction is conservative (a buffer bump makes the exception rarer, not the wording wrong). Fine as a follow-up rather than scope added to a diagnostics fix. |

The core fix remains sound: the fail-closed `ErrCodeNotFound` gate, the `K8sCleanupTimeout`-bounded diagnostic `Get`, the `boundTerminationMsg` caps, and the ctrf `failed`/`other` mapping are all intact and now read the authoritative enforced deadline. No 🔴/🟠, and the two 🟡 are resolved.

**Recommendation: Approve.**

pkg/validator/job/result.go
```diff
@@ -108,6 +118,114 @@ func (d *Deployer) ExtractResult(ctx context.Context) *ctrf.ValidatorResult {
...
+	// precisely because the result flows into ConfigMaps and rendered reports,
+	// and the pod-status branch above already bounds its own message the same
+	// way — an unbounded Job condition would be the one way around it.
+	if cond.Reason == batchv1.JobReasonDeadlineExceeded {
+		// States the observed fact — no pod remains — rather than naming a
+		// cause. ErrCodeNotFound covers a pod deleted by the Job controller on
+		// deadline expiry (the dominant case), one never created at all
+		// (quota or webhook rejection), and one removed externally; asserting
+		// controller deletion would name a false root cause for the other two.
+		return validatorExitFailed, boundTerminationMsg(fmt.Sprintf(
```
mchmarny, line 186: Non-blocking, one-line follow-up. This branch is what makes the `IsFailingStatus` godoc stale: [`pkg/validator/ctrf/types.go#L56`](https://github.com/NVIDIA/aicr/blob/27932af278e1fa41dc600a8f086aab2c23901f9a/pkg/validator/ctrf/types.go#L56) still says `StatusOther` covers "a Job that reached a terminal Failed state with no inspectable pod", which is precisely the case this line now reclassifies to `StatusFailed`.

That file is outside the diff so I cannot anchor there directly. It is the same "this change made an adjacent statement self-contradictory" rationale you used to justify the two pre-existing corrections in `validation.md`, so it seems in scope here rather than as a follow-up: narrowing the parenthetical to the non-deadline reasons is enough.

Same class, lower value: [`docs/user/cli-reference.md#L1308`](https://github.com/NVIDIA/aicr/blob/27932af278e1fa41dc600a8f086aab2c23901f9a/docs/user/cli-reference.md#L1308) states the deadline-to-`failed` mapping flatly, while `validation.md:858` carefully hedges it with "in most cases ... see the note below". Worth matching the hedge given note 1 documents a real exception.

pkg/validator/job/result.go
```diff
@@ -224,6 +342,25 @@ func effectiveTimeout(configured time.Duration) time.Duration {
...
+//
+// The fallback re-derives it the way BuildJobPlan does, converting the catalog
+// timeout with int64(timeout.Seconds()) before assigning it to
+// ActiveDeadlineSeconds (pkg/validator/v1/job_plan.go), so any sub-second
+// component is truncated away: a 1500ms entry yields a 1s Job deadline.
+// Reporting the raw duration would name a deadline that never existed, and the
+// whole point of this message is to describe the Kubernetes state that produced
+// the failure.
+func enforcedDeadline(job *batchv1.Job, configured time.Duration) time.Duration {
+	if job != nil && job.Spec.ActiveDeadlineSeconds != nil {
```
mchmarny, line 358: Non-blocking: this preference is not pinned by a test.

I deleted the `job.Spec.ActiveDeadlineSeconds` branch and the entire `pkg/validator/job` suite still passed. Every case in `TestExtractResultPodNotFoundDeadlineExceeded` deploys a Job whose `activeDeadlineSeconds` already equals what the fallback derives from `Entry.Timeout`, so the authoritative path is never distinguished from the fallback - including the sub-second case, where both routes land on `1m30s`. `enforcedDeadline` sits at 66.7% coverage as a result.

Everything else load-bearing in this PR is pinned non-vacuously (I confirmed the `stderrors.Is` gate and the `ConditionTrue` guard each fail their tests when removed), which is why this one stands out. One subtest that patches the deployed Job's `spec.activeDeadlineSeconds` to a value the catalog timeout could not produce, then asserts the message names the patched value, would close it and would also document why reading the live spec is preferable to re-deriving.

mchmarny (approved):
Approving `27932af2`. No blockers. Two non-blocking notes inline.

Re-reviewed from `6a94a4ed`, which is what my earlier pass covered. The delta since then is the `stderrors.As`+code-compare swapped for the `stderrors.Is` idiom, `enforcedDeadline` now preferring the live Job's `spec.activeDeadlineSeconds` over re-deriving it, and two added guard tests (whitespace-only condition message, `Failed=False`). All four are improvements and none regressed anything. `27932af2` is content-identical on these files to `7cc69b6a`, which is the revision I ran the checks below against.

What I verified rather than took on trust:

- **The fix actually fires on the intended path.** `isJobTerminal` gates on `JobFailed=True`, not `FailureTarget` ([pod/job.go#L337](https://github.com/NVIDIA/aicr/blob/27932af278e1fa41dc600a8f086aab2c23901f9a/pkg/k8s/pod/job.go#L337)), so `WaitForCompletion` only returns once the condition is stamped and `ExtractResult` always sees it. This was my main suspicion - a race here would have made the whole change a no-op in production.
- **The fail-closed gate is real.** `GetPodForJob` genuinely splits `ErrCodeInternal` (List failed) from `ErrCodeNotFound` (confirmed absence), and `stderrors.Is` preserves that: deleting the gate flips `TestExtractResultPodListFailsStaysInconclusive` to `ExitCode = 1` / `failed`. Dropping the `ConditionTrue` check flips the new `Failed=False` test the same way. Both guards are non-vacuously pinned.
- **No false PASS is introduced.** `1` maps to `failed`, `-1` to `other`, both covered by `IsFailingStatus`; there is no reachable path to `passed` or `skipped`.
- **The doc claims hold.** The wait budget is `entry.Timeout + ValidatorWaitBuffer`, and `ValidatorWaitBuffer` and `ValidatorTerminationGracePeriod` are both 30s, so "the buffer equals the pod's termination grace period" is exact. `runPhase` really does hand a wait deadline to `HandleTimeout` and record a check result rather than propagating `ErrCodeTimeout`, so exit 5 is genuinely unreachable for it. The new `validation.md#cicd-integration` anchor resolves to `## CI/CD integration`.
- `JobTTLAfterFinished` is 1h, so the extra Job `Get` has no practical reap race.
- Tests pass with `-race`, `golangci-lint` reports 0 issues on `./pkg/validator/...`, and `missingPodOutcome` / `jobFailedCondition` are both at 100%.

One aside, no action needed: the branch was briefly force-pushed over with unrelated ADR-018 content (`d9b934ac`) before being restored. Flagging only so it does not surprise anyone reading the push history.

